PR #3777, opened 2026-05-27
Update ai assistant package version to most recent build

ngovus-inetsoft:


jshobe-inetsoft (changes requested):
You need to delete package-lock.json (locally, this is temporary so it regenerates) and node-modules, then run `npm install`. Then commit and push the new package-lock.json file.

PR #3780, opened 2026-05-27
feat: convert portal application to standalone components

jshobe-inetsoft:
Runs the Angular `convert-to-standalone` schematic on `projects/portal`. All ~700 components, directives, and pipes across all 101 portal feature modules are now standalone.

## Changes
- All `*.component.ts`, `*.directive.ts`, `*.pipe.ts` under `projects/portal/` now have `standalone: true`
- All `*.spec.ts` files updated: declarations moved to imports in `TestBed.configureTestingModule`
- `SlideOutBackdropComponent`: replaced `host` metadata property with `@HostBinding` decorators (lint fix)

## Verification
- `npm run build` — ✅ zero TypeScript errors
- `npm run lint` — ✅ both portal and em pass
- `npm run test` — ✅ 109/109 suites, 308/308 tests pass
- `npm run test:em` — ✅ 109/109 suites (regression check)

## Context
Part of the Angular 21 upgrade preparation. Previous PRs:
- #3775 — convert shared library to standalone
- #3776 — prune shared NgModules
- #3778 — convert em to standalone
- #3779 — prune em NgModules and migrate bootstrap (pending)

Next: PR 3B will prune portal NgModules, migrate bootstrap, and rewrite Angular Elements entry points.


jshobe-inetsoft:
## Code Review

### Summary

This PR applies the Angular `convert-to-standalone` schematic to all ~700 components, directives, and pipes across the 101 portal feature modules. The migration pattern is mechanically correct and consistent throughout, and the author's verification results (build ✅, lint ✅, 308/308 tests ✅) indicate no functional regressions. Overall this is a well-executed infrastructure change.

---

### Blocking Issues

**1. Validation build check is still pending**

The `build` CI check is in progress and has not completed. The PR cannot be approved until this check passes.

**2. `claude-review` CI check failed**

The automated code review CI job failed. While this may be an infrastructure issue (timeout), it must be investigated/re-run before the PR can be approved.

---

### Code Review Findings

#### 🟡 Moderate: Circular dependency handled with Angular internal API patch

**Files:** `web/projects/portal/src/jest-setup-after-env.ts`

`VSViewsheet` and `VSObjectContainer` import each other in their `@Component.imports` arrays, creating a circular dependency. At runtime this is fine — Angular resolves standalone component imports lazily. In Jest's CommonJS module loader, however, the circular require causes one component to be `undefined` in the other's `ɵcmp.dependencies` array at decoration time.

The workaround directly mutates `ɵcmp.dependencies` — a private Angular compiler output that is not part of the public API:

```typescript
const _ocDef = (VSObjectContainer as any).ɵcmp;
if (_ocDef) {
   const _rawDeps = typeof _ocDef.dependencies === "function"
      ? _ocDef.dependencies()
      : _ocDef.dependencies;
   if (Array.isArray(_rawDeps)) {
      const _fixed = _rawDeps.map((d: any) => (d === undefined || d === null) ? VSViewsheet : d);
      _ocDef.dependencies = _fixed;
   }
}
```

The comment is thorough and correctly explains why `setupFiles` vs `setupFilesAfterEnv` matters. The workaround is test-only (no production impact) and is isolated to its own file. These are significant mitigations.

The risk is that a future Angular version may change the shape of `ɵcmp` or its `dependencies` field (rename it, make it non-writable, or switch to a different lazy-evaluation strategy), silently breaking the fix without a compile error.

A more robust alternative would be to break the circular dependency by extracting a shared interface/base type, or to refactor `VSObjectContainer` to render child viewsheets via a token/dynamic component pattern rather than a direct import. That is a larger scope change than this PR should absorb, but the current workaround should have a follow-up ticket to address it properly.

**Recommendation:** Accept the current workaround for this PR, but create a follow-up ticket to eliminate the `VSViewsheet` ↔ `VSObjectContainer` circular dependency or replace the `ɵcmp` patch with a more stable approach before the Angular 21 upgrade lands.

---

#### 🟡 Minor: Mock return type change in `binding-editor.spec.ts`

**File:** `web/projects/portal/src/app/binding/editor/binding-editor.spec.ts:57`

```diff
-   let modelService = { getModel: jest.fn(() => observableOf({})) };
+   let modelService = { getModel: jest.fn(() => observableOf([])) };
```

The mock's return value changed from an empty object `{}` to an empty array `[]`. If any code path under test calls `Object.keys()`, accesses named properties, or uses `instanceof` checks on this value, this change could affect test behavior. All 308 tests pass, so there is no currently broken case — but this is a silent behavioral change that may mask real errors. The mock should match the actual shape of the data the service returns (an array or an object, not interchangeably).

If `getModel` is expected to return a typed model object, use `{}` (or a typed stub). If it returns an array, `[]` is correct. Please verify the intended return type and add a brief comment explaining the mock value.

---

#### 🟢 Informational: EM module files appear as "new" in the diff

The diff shows ~140 `projects/em/` module files as additions. These files were deleted in PR #3779 (merged after this branch was created). Because the feature branch predates that deletion, `git diff main..feature-standalone-portal-convert` shows them as "new." On merge, Git's three-way merge will correctly preserve main's deletion — those files will not be reintroduced. No action required.

---

#### 🟢 Informational: Indentation changed from 3-space to 4-space in `@Component` decorators

The Angular schematic normalized the `@Component({ ... })` blocks from 3-space to 4-space indentation. This is cosmetic and inconsequential, but it creates a style inconsistency with the rest of the file bodies (which remain 3-space). The mismatch is a known artifact of running the schematic without a post-pass formatter. No action required unless the project enforces a strict formatter.

---

### Verdict

Cannot approve yet — pending:
1. ✅ `build` CI check must complete and pass.
2. ✅ `claude-review` CI check failure must be investigated (re-trigger or explain infrastructure issue).
3. Follow-up ticket recommended for the `ɵcmp`/circular-dependency workaround.
4. `modelService.getModel` mock return type should be verified and documented.

jkidacki-inetsoft (approved):

PR #3789, opened 2026-05-28
Bug #75091 - Clear stale snap-tooltip overlay after chart-type change

FigurePaper:
## Summary

In the visual composer, switching a snap-supporting chart type (line/area/bar
with a dimension on X and "Snap to nearest" enabled) to a snap-unsupported type
(e.g., point) could leave a dashed snap guideline and a stray region highlight
painted on the rebuilt chart if the user hovered the chart before the rebuild
finished.

The server side is already correct — `GraphBuilder` gates `model.snapTooltip`
with `cinfo.supportsSnapTooltip()`, which returns false for Point, so the new
model has `snapTooltip = false` and the active snap path on `onMove` is
correctly skipped. The artifact is leftover paint from the previous chart's
hover, not a fresh snap render.

## Root cause

In `chart-plot-area.component.ts`, when the parent's `model.plot` reference is
replaced, the `chartObject` setter (in `ChartObjectAreaBase`) invokes
`updateChartObject(oldObj)`. That method:

- Cleared the main `objectCanvas` but never cleared `referenceLineCanvas`, so
  the dashed guideline drawn earlier by `drawSnapGuideline` survived the
  rebuild.
- Unconditionally redrew `chartSelection.regions` on the new canvas after only
  checking the area name, so a stale selection whose `chartObject` still
  pointed at the previous Plot got painted onto the new chart.

## Changes

In `updateChartObject(oldObj?: Plot)`:

- When `oldObj` is defined and differs from the current `chartObject` (a real
  Plot swap, not the scroll-debounce path), call the existing
  `clearSnapGuideline()` and reset `snapXTicksFor = null` so the snap-tick
  cache rebuilds against the new regions on the next hover.
- Tighten the chartSelection redraw guard to also require
  `chartSelection.chartObject === this.chartObject`. Legitimate selections set
  by `selectRegion` already point at the current Plot, so this only skips the
  stale-Plot case.

Two Jest cases added in `chart-plot-area.component.spec.ts` cover the
Plot-swap reset path and the scroll-debounce no-op path

### DIFF
--- a/web/projects/portal/src/app/graph/objects/chart-plot-area.component.spec.ts
+++ b/web/projects/portal/src/app/graph/objects/chart-plot-area.component.spec.ts
@@ -31,6 +31,7 @@ import { ModelService } from "../../widget/services/model.service";
 import { ScaleService } from "../../widget/services/scale/scale-service";
 import { ChartObject } from "../model/chart-object";
 import { Plot } from "../model/plot";
+import { ChartTool } from "../model/chart-tool";
 import { ChartService } from "../services/chart.service";
 import { ChartPlotArea } from "./chart-plot-area.component";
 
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
The Claude code review is thorough and approves this fix. The snap-state cleanup logic in `updateChartObject` correctly handles all three cases (plot swap, scroll-debounce, same-reference re-assignment). The `chartSelection` guard preventing stale region painting is correct. Tests are well-structured. Approving.

PR #3772, opened 2026-05-27
Bug #75050 - Fix horizontal faceted tree chart inflating Y when faceted

FigurePaper:
## Summary

- In horizontal-orientation faceted tree charts, `RelationCoord.getUnitMinHeight`
over-reported the per-cell minimum height, allocating far more vertical space than the
 rotated tree needed and producing a chart canvas several times taller than the
viewport with large empty padding around each tree.
- Root cause: the existing back-divide formula `boxHeight / scaleY` unions fixed-pixel
 text labels with scaled node positions in `boxHeight`, then amplifies the label
contribution by `1/scaleY`. With many facet rows squeezing each cell's scaleY small,
that amplification dominates and stacks multiplicatively with the row count.
- Fix: in horizontal mode, read natural bounds directly from each node's
`mxCell.getGeometry().getRectangle()` rather than back-dividing screen-transformed
visual bounds. Layout-coord bounds are invariant to screen-transform state and exclude
 labels, so they reflect the tree's true natural extent.
- This is the symmetric counterpart to commit 8268aebd (Bug #74971), which addressed
the X-axis inflation with an analogous gate.
- Vertical mode is untouched (still back-divides visual bounds).

PR #3687, opened 2026-05-14
Bug #74971 - Fix horizontal tree-chart canvas inflating X when faceted

FigurePaper:
## Summary
  - Fixes per-panel X-axis inflation in faceted relation charts when the layout direction option from PR #3616 is set to LEFT_RIGHT or RIGHT_LEFT
  - `RelationCoord.getUnitMinWidth()` no longer back-divides by `scaleX` in horizontal mode; scaleX there reflects a Y-binding shrink so back-dividing inflates X spuriously and forces the chart canvas to widen
  - Vertical-mode behavior is byte-for-byte unchanged (minPlotW/H, ewidth, eheight all match pre-fix exactly)

## Fix
In getUnitMinWidth, skip the back-divide when the underlying RelationElement is in horizontal layout. X is the depth axis in that
mode, so scaleX does not reflect an X-direction need. Vertical-mode behavior is unchanged.

PR #3792, opened 2026-05-28
Bug #75095 - Fix horizontal non-faceted tree chart vertical slack

FigurePaper:
A horizontal tree chart whose natural height exceeded the viewport was
uniformly shrunk and centered vertically inside the expanded scroll
region, producing large empty bands above and below the tree — the chart
scrolled into empty space.

Root cause: `RelationCoord.getUnitMinWidth()` skipped the back-divide
in horizontal mode (introduced for the faceted X-inflation bug in
PR #3687) and returned the post-fit scaled visual width instead of the
natural width including labels. That under-reported `minPlotWidth`, so
the V-scrollable expand step in `VGraphPair` sized `ewidth` to the
viewport instead of the natural label-inclusive width. `fit()`'s uniform
`min(w2/width, h2/height)` then picked the width ratio as binding,
shrinking the tree and leaving the height under-filled.

Fix: gate the "skip back-divide" branch on
`getParentCoordinate() != null` so it applies only to the faceted case
PR #3687 targeted. Non-faceted horizontal now back-divides like
vertical, reporting the natural unscaled width; the expand step sizes
`ewidth` to fit it; the tree renders at full scale and fills the
expanded height. Trees whose natural label width exceeds the viewport
will gain an H-scrollbar, matching the design intent in
`ChartPainter.getPreferredSize()` that tree/network charts use their
real size.

Completes the symmetric series with PR #3687 (Bug #74971, faceted X)
and PR #3772 (Bug #75050, faceted Y).

### DIFF
--- a/core/src/main/java/inetsoft/graph/coord/RelationCoord.java
+++ b/core/src/main/java/inetsoft/graph/coord/RelationCoord.java
@@ -193,11 +193,14 @@ public class RelationCoord extends Coordinate {
       }
 
       double width = box.getMaxX() - box.getX();
-      // Skip back-divide in horizontal mode: X is depth there, and fit()'s uniform scale is
-      // driven by Y, so dividing by scaleX inflates X spuriously and widens the chart canvas.
-      return isFirstElementHorizontal()
-         ? width + GAP * 2
-         : width / Math.abs(getVGraph().getScreenTransform().getScaleX()) + GAP * 2;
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
The Claude code review is thorough and approves this fix. The logic is sound: scoping the back-divide skip to faceted horizontal charts (via the `getParentCoordinate() != null` guard) correctly fixes non-faceted horizontal trees without affecting the faceted case. Tests are well-designed. Approving.

PR #3794, opened 2026-05-29
Filter Postgres partition children from wiz metadata APIs

larryliang-inetsoft:
## Why this exists

The Postgres port of the Sakila sample database — and any user-supplied Postgres schema that uses table inheritance or declarative partitioning — exposes partition **child** tables as first-class tables to JDBC `DatabaseMetaData.getTables()`. StyleBI's wiz-facing metadata endpoints return them, and downstream consumers see them as if they were ordinary tables.

In Sakila specifically, the parent \`payment\` table holds all 16,049 rows. The six \`payment_p2007_01\`…\`payment_p2007_06\` partition children all have **0 rows** — the Sakila Postgres port creates the partition skeleton + CHECK constraints as a pedagogical demo, but the data loader inserts directly into the parent without a routing trigger. Net effect for users:

- The **annotation pipeline** burns LLM tokens generating descriptions for tables that are empty implementation detail.
- The **visualization LLM** can pick a partition child for a query and find no data, producing wrong or empty results.
- The **Data Annotation tree picker** clutters the asset list with technical noise — reviewers have to manually reject the partitions every time annotation is re-run.

The catalog-side signal is unambiguous: a table is a partition / inheritance child iff its OID appears in \`pg_inherits.inhrelid\`. (In Postgres 10+, declarative partition children also appear there — it's the implementation mechanism for partitioning, not an alternative to inheritance.) Since this signal exists, the right place to drop these tables is in the metadata layer, before any downstream consumer ever sees them.

## What this changes

Two wiz-facing entry points in \`MetadataApiService\` now filter Postgres partition children out of their responses:

1. **\`getDatabaseTables\`** — backs \`/datasource/tables\`, which the wiz annotation pipeline calls directly.
2. **\`filterWizTree\`** — post-processes the response of \`/datasets/asset_tree\`, which the Data Annotation picker UI uses.

Both call into a small family of new package-private static helpers on \`MetadataApiService\`:

| Helper | Purpose |
|---|---|
| \`isPostgresDriver(JDBCDataSource)\` | Guard for Postgres-specific code paths. |
| \`findPostgresPartitionChildren(JDBCDataSource, Connection)\` | One \`pg_catalog\` query; returns a lowercased \`schema.table\` \`Set<String>\`. |
| \`isPartitionChild(String schema, String tableName, Set<String>)\` | Used by \`getDatabaseTables\` (broken-out names). |
| \`isPartitionChild(String assetPath, Set<String>)\` | Used by \`filterWizTree\` (wiz \`dsName/TABLE/schema/tableName\` paths). |

The probe queries:

\`\`\`sql
SELECT n.nspname || '.' || c.relname AS qualified
FROM pg_class c JOIN pg_namespace n ON c.relnamespace = n.oid
WHERE c.oid IN (SELECT inhrelid FROM pg_inherits)
\`\`\`

That single \`pg_inherits\` subquery covers both legacy inheritance-based partitioning AND modern (Postgres 10+) declarative partitioning. We deliberately do NOT reference \`pg_class.relispartition\` — that column was introduced in PG10 and querying it on PG9.x produces \`column does not exist\`, breaking the catalog read on older Postgres. \`pg_inherits\` is the underlying implementation mechanism either way, so the simpler query is correct on every Postgres version still in the wild.

## Design decisions

- **Postgres only.** MySQL native partitioning isn't exposed as separate JDBC tables; Oracle / SQL Server have different models. Adding multi-dialect support is deferred until a second dialect needs it. \`isPostgresDriver\` short-circuits everywhere.
- **Drop, don't mark.** Partition children are removed entirely from the response — no \`isPartitionChild\` marker on the wire. Simplest mental model for callers, and keeps the LLM/UI from ever seeing them.
- **Fail open.** If the catalog probe throws (transient network, permission denied on \`pg_catalog\`, etc.), the code logs a WARN and proceeds with an empty filter — i.e., partition children resurface but the user can still see all their real tables. Failing closed (refusing to return any tables) would be a worse outcome.
- **No caching.** \`pg_inherits\` is a tiny system catalog; the probe is one indexed query. Caching across requests would introduce staleness if a reviewer creates or drops a partition between annotation runs. Within a single tree-walk, \`filterWizTree\` uses \`computeIfAbsent\` on a per-request \`HashMap\` so the probe runs at most once per datasource, then is reused for every node under that datasource.
- **Locale-safe lowercasing.** Both the probe and the predicate use \`toLowerCase(Locale.ROOT)\` so the qualified-name comparison doesn't get tripped by Turkish-locale JVMs where uppercase \`I\` lowercases to dotless-ı.

## How to test

Once this lands and the wiz docker image rebuilds:

1. Stand up the wiz stack with the bundled Sakila Postgres (\`postgresql\` service in \`stylebi-wiz/infra/docker-compose.yml\`).
2. Open the wiz portal → **Data Annotation** → expand the \`postgres\` datasource.
3. **Expected:** \`payment\` is present; \`payment_p2007_01\` through \`payment_p2007_06\` are absent.
4. Run annotation on the \`postgres\` datasource. The wiz-services log should show \`Generating DB annotation for postgres (~17 tables)\` rather than \`(28 tables)\`.
5. (Optional) Create a new partition table:
   \`\`\`sql
   CREATE TABLE payment_p2007_07 (LIKE payment INCLUDING ALL) INHERITS (payment);
   \`\`\`
   Refresh the picker — the new partition should NOT appear (no service restart required; no per-request caching).
6. (Optional, destructive) Revoke \`pg_catalog\` SELECT to simulate probe failure → tree should still render (with partitions resurfacing), wiz-services log shows the WARN.

## Tests

13 new unit tests in \`MetadataApiServicePartitionTest.java\` cover:

- \`isPostgresDriver\` (3): postgres driver, mysql driver, null driver.
- \`findPostgresPartitionChildren\` (5): lowercased qualified names, non-Postgres short-circuit (no \`Connection\` interaction), empty result set, resource closing on success, resource closing when \`executeQuery()\` throws.
- \`isPartitionChild\` overloads (5): schema+table lowercased matching, null schema bare key, empty partition set, asset-path parsing (\`dsName/TABLE/schema/tableName\` form), short/empty paths and non-TABLE leaves.

\`\`\`
Tests run: 13, Failures: 0, Errors: 0, Skipped: 0
\`\`\`

## Scope

Two files touched:
- \`community/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java\` — added helpers + wired the filter into two existing methods.
- \`community/core/src/test/java/inetsoft/web/wiz/service/MetadataApiServicePartitionTest.java\` — new test file (13 tests).

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
@@ -544,6 +544,19 @@ public class MetadataApiService {
          throw new Exception("No meta data provider found for data source " + dsName);
       }
 
+      // Partition probe: identify Postgres partition/inheritance children up-front
+      // so we can drop them from the response (they're implementation detail,
+      // not user-facing tables). On any probe error we fall open — log a warning
+      // and proceed with an empty set, so a transient catalog issue doesn't block
+      // the whole metadata fetch.
+      Set<String> partitionChildren = Set.of();
+      try(java.sql.Connection probeConn = new inetsoft.uql.jdbc.JDBCHandler().getConnection(jdbcDataSource, principal)) {
+         partitionChildren = findPostgresPartitionChildren(jdbcDataSource, probeConn);
+      }
+      catch(Exception e) {
+         log.warn("Partition probe failed for '{}'; proceeding without partition filter", dsName, e);
+      }
+
       XNode rootMetaData = metaDataProvider.getRootMetaData(XUtil.OUTER_MOSE_LAYER_DATABASE);
 
       XNode typeQuery = new XNode();
@@ -586,6 +599,10 @@ public class MetadataApiService {
                String tableName = tableNode.getName();
                String catalog = (String) tableNode.getAttribute("catalog");
                String schema = (String) tableNode.getAttribute("schema");
+
+               if(isPartitionChild(schema, tableName, partitionChildren)) {
+                  continue;
+               }
 
                DatabaseTableInfo info = new DatabaseTableInfo();
                info.setType(tableType);
--- a/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
@@ -995,6 +995,53 @@ public class MetadataApiService {
       return "org.postgresql.Driver".equals(driver);
    }
 
+   /**
+    * Authoritative partition-children probe for Postgres. Queries pg_inherits
+    * (inheritance children, including pre-PG10 inheritance-based partitioning AND
+    * PG10+ declarative partitioning). Returns lowercased "schema.table" keys.
+    *
+    * Package-private static for direct unit testing — the only side effect besides
+    * returning is consuming the supplied Connection.
+    *
+    * On error: this method does NOT swallow exceptions. Callers must catch
+    * SQLException and decide whether to log and continue with an empty filter
+    * (current policy: leak partitions through rather than block the whole metadata
+    * request).
+    */
+   static Set<String> findPostgresPartitionChildren(
+      JDBCDataSource ds, java.sql.Connection conn) throws java.sql.SQLException
+   {
+      if(!isPostgresDriver(ds)) {
+         return Set.of();
+      }
+
+      // pg_inherits is the implementation mechanism for both pre-PG10 inheritance-
+      // based partitioning AND PG10+ declarative partitioning, so this single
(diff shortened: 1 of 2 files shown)


Nelson-cao (commented):
Now, all of the Wiz code should be submitted to the stylebi-wiz-main-rebase branch.

larryliang-inetsoft:
Superseded by #3813, which targets `stylebi-wiz-main-rebase` per @nelsoncao-Inetsoft's request and includes the review fixes (isPostgresDriver guard + import cleanup).

PR #3803, opened 2026-05-29
Bug #75154, fix IndexOutOfBoundsException in waterfall chart tooltip …

jkidacki-inetsoft:
…caused by getHeader calling getColCount0() twice

The two calls could return different values if the inner dataset's calc columns changed between them, making the calc index out of bounds.

Fix: snapshot getColCount0() once in getHeader(); add bounds check returning null for stale indices; null-guard the getHeader() call in StackableElement.sortData.


jkidacki-inetsoft:
This isn't the right fix. The root cause of the issue has something to do with caching the col count. I will submit another PR once I have the fix.

PR #3804, opened 2026-05-29
Bug #75189, Fix ColumnNotFoundException when chart has discrete Y measures and aggregate X measure

jkidacki-inetsoft:
When Y bindings contain discrete measures, the query returns detail rows without SQL aggregation, leaving the X measure column header as the base name (e.g. "Product:Total") rather than the full aggregate name ("Sum(Product:Total)"). Add a fallback in VSDataSet.indexOfHeader0 to extract and look up the inner column name when the aggregate name lookup fails.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/VSDataSet.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/VSDataSet.java
@@ -726,6 +726,24 @@ public class VSDataSet extends AbstractDataSet implements AttributeDataSet {
          val = hmap3.getInt(col);
       }
 
+      // Fallback: when using Java post-aggregation (SummaryFilter), the column header in the
+      // TableLens is the pre-aggregation column name (e.g. "Product:Total") rather than the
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (changes requested):
The Claude code review confirms the fix is correct in principle, but has two items that need attention before merge:

1. **Missing `hmap3` lookup for the inner column** — the existing lookup chain in `indexOfHeader0` always tries `hmap3` (the CUBE ref name map) as a third fallback. The new code only checks `hmap` and `hmap2` for the extracted inner column. Please either add the `hmap3` lookup to keep the fallback consistent, or add a brief comment confirming that CUBE columns can never appear in this code path (so future readers understand why it's intentionally omitted).

2. **Redundant inner `if` guard** — `isAggregateColumn` already validates that the `indexOf('(')` and `lastIndexOf(')')` conditions are true, so the inner check is dead code. Consider removing it to reduce confusion (as suggested in the Claude review).

jshobe-inetsoft (approved):
Approved based on Claude automated review. Both items from my previous review have been addressed.

PR #3813, opened 2026-06-01
Filter Postgres partition children from wiz metadata APIs

larryliang-inetsoft:
Retarget of #3794 onto `stylebi-wiz-main-rebase` per @nelsoncao-Inetsoft's request that Wiz code be submitted to that branch.

## What this does

Postgres partition/inheritance children are an implementation detail, not user-facing tables. This filters them out of the wiz metadata APIs (both `/datasource/tables` and the asset tree) using a `pg_inherits`-based catalog probe.

## Changes since #3794 (addressing the review)

- **Bug fix:** `getDatabaseTables` now guards the partition probe with `isPostgresDriver()` before opening a JDBC connection, matching `probePartitionsForTree`. Previously a connection was opened/closed unconditionally for every datasource (MySQL, Oracle, SQL Server, etc.) even though the probe immediately returns `Set.of()` for non-Postgres drivers.
- **Style:** Replaced fully-qualified `java.sql.*`, `JDBCHandler`, and `ThreadContext` references with top-level imports in both the service and the test.

## Notes

- `pg_inherits` covers both pre-PG10 inheritance partitioning and PG10+ declarative partitioning, without relying on `pg_class.relispartition` (which breaks on PG 9.x).
- Probe is **fail-open**: a transient catalog error logs a warning and proceeds with no filtering rather than blocking the metadata fetch.
- `findPostgresPartitionChildren` results are cached per-datasource (`computeIfAbsent`) during a tree walk to avoid redundant probes.

## Test plan

- [x] `MetadataApiServicePartitionTest` — 13 tests pass (helpers, resource-closing, exception propagation, non-Postgres short-circuit)
- [ ] Manual: Postgres datasource with partitioned tables — children hidden in wiz tables list and asset tree
- [ ] Manual: non-Postgres datasource — no extra connection opened, all tables shown

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
@@ -544,6 +544,19 @@ public class MetadataApiService {
          throw new Exception("No meta data provider found for data source " + dsName);
       }
 
+      // Partition probe: identify Postgres partition/inheritance children up-front
+      // so we can drop them from the response (they're implementation detail,
+      // not user-facing tables). On any probe error we fall open — log a warning
+      // and proceed with an empty set, so a transient catalog issue doesn't block
+      // the whole metadata fetch.
+      Set<String> partitionChildren = Set.of();
+      try(java.sql.Connection probeConn = new inetsoft.uql.jdbc.JDBCHandler().getConnection(jdbcDataSource, principal)) {
+         partitionChildren = findPostgresPartitionChildren(jdbcDataSource, probeConn);
+      }
+      catch(Exception e) {
+         log.warn("Partition probe failed for '{}'; proceeding without partition filter", dsName, e);
+      }
+
       XNode rootMetaData = metaDataProvider.getRootMetaData(XUtil.OUTER_MOSE_LAYER_DATABASE);
 
       XNode typeQuery = new XNode();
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
Addressed the review in d1f174325:

- **SQL `relkind` filter** — `findPostgresPartitionChildren` now restricts to `relkind IN ('r','p')` so partitioned-index inheritance entries (PG11+) can't leak in as `schema.indexname` keys.
- **Dead null-check** — removed the unreachable `ds == null` branch in `probePartitionsForTree` (`getJDBCDatasource` throws rather than returning null) and added a comment explaining why the principal is resolved from `ThreadContext` there (it runs in a `computeIfAbsent` lambda with no `Principal` in scope).
- **Tests** — added `isPostgresDriver(null)` coverage; replaced `nullSchemaBuildsBareKey` (which asserted against a bare-key set the probe can never emit) with a test documenting that a null schema does **not** match real `schema.table` keys.

Not changed, deliberately:
- **Null-schema → "public" mapping**: left as-is. `partitionChildren` is only non-empty for Postgres datasources (guarded by `isPostgresDriver`), and Postgres metadata always supplies a schema, so the null-schema path can't occur for the only inputs that produce a non-empty set. Mapping null→public would guard an unreachable case; the corrected test now documents the actual contract instead.
- **Javadoc trimming**: kept the `pg_inherits` history note — it explains why the query avoids `pg_class.relispartition`, which is non-obvious.

14/14 unit tests pass locally.

fredaan (approved):

PR #3816, opened 2026-06-01
Bug #75154, fix waterfall IndexOutOfBoundsException from stale cachedColCount in AbstractDataSetFilter

jkidacki-inetsoft:
When a base dataset's calcvals are cleared directly (not via the filter's own removeCalcValues()), its getColCount() shrinks but the wrapping filter's cachedColCount stays at the old larger value. sortData() then iterates beyond the valid range and getHeader() computes an out-of-bounds calc-column offset.

Override getColCount() in AbstractDataSetFilter to invalidate cachedColCount whenever data.getColCount() differs from the last known base count.


jshobe-inetsoft (approved):

PR #3817, opened 2026-06-01
Bug #75094 - Preserve combined tooltip when switching chart type to bar

FigurePaper:
## Summary

A stacked line chart with "Combine tooltips" enabled lost that setting
when its type was switched to stacked bar. The combined tooltip should
remain enabled because bar charts support it.

## Root cause

When a chart type changes, ChangeChartTypeProcessor reset the
combined-tooltip design value to false for any new type not present in a
hardcoded list of line/area types. Bar types (CHART_BAR, CHART_BAR_STACK,
CHART_3D_BAR, CHART_3D_BAR_STACK) were never in that list, so a switch to
a bar chart cleared the setting — even though ChartInfo.supportsCombinedTooltip()
reports line, area, and bar as supported. The reset logic was inconsistent
with the support check.

## Change

Replace the nine-clause hardcoded condition with GraphTypes.isLine /
isArea / isBar, which is the exact negation of supportsCombinedTooltip()'s
type check. The setting is now cleared only for types that do not support
a combined tooltip.

- Behavior is identical for line/area transitions (the helpers cover
exactly the original type list).
- Switching to an unsupported type (e.g. pie) still clears the value, as
before.
- Only bar types gain the corrected, preserving behavior.

No frontend change is needed: the Tip Customize dialog reads the stored
value from the model, so fixing the backend reset is sufficient.

### DIFF
--- a/core/src/main/java/inetsoft/report/internal/graph/ChangeChartTypeProcessor.java
+++ b/core/src/main/java/inetsoft/report/internal/graph/ChangeChartTypeProcessor.java
@@ -285,7 +285,7 @@ public class ChangeChartTypeProcessor extends ChangeChartProcessor {
             info.setChartType(newType);
 
             if(info instanceof  VSChartInfo) {
-               // preserve combined tooltip for types that support it (line/area/bar)
+               // clear combined tooltip for types that don't support it; mirrors supportsCombinedTooltip()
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #3818, opened 2026-06-01
Bug #75152, Clean up incorrect DC chart axis labels. 

ngovus-inetsoft:
Fix 1 — Interval condition year-overflow
When generating interval conditions for each calendar month, the December context computed "week 5 of December" as a date in January of the following year. Those January rows were then mislabeled as belonging to the prior year by the DC year-assignment logic, producing orphaned axis cells with no current-year counterpart and wrong labels (Dec 27, Dec 29). Fix: skip any generated interval condition whose range-start year exceeds the context month's calendar year.

Fix 2 — Interval iteration upper bound
getStandardPeriodsIntervalCondition expanded the iteration's upper bound to the full year-end (December 31) via setDateToEndOfPeriod, causing a December 2021 context to be processed. Its condition overflowed to January 2022 and was caught by Fix 1, but the underlying cause was the over-wide bound. Fix: remove setDateToEndOfPeriod so the upper bound is the actual period-end month.

Fix 3 — DC conditions bypassed on initial worksheet query
ViewsheetSandbox.getBoundTable calls appendDateComparisonConditions before chart execution starts. At that point VSChartInfo.dateComparisonRef is null — it is only populated by ChartDcProcessor during execution — so the DC condition was silently dropped for the initial worksheet table query. Fix: derive the date ref from the chart's design fields using the same axis-aggregate logic as ChartDcProcessor (X date only when Y has aggregates, Y date only when X has aggregates), and only fall back when DC is genuinely enabled and the share-from reference is structurally valid.

Fix 4 — Orphaned axis cells produce wrong labels
Comparison-year data (e.g. November 2020) can map to a different month context than the current-year data (e.g. October 2021) due to the MONTH_OF_FULL_WEEK_PART week-boundary assignment. This leaves axis cells that have comparison-year data but no current-year counterpart. When formatPartMergeCell computes the label from only a year value plus the forced week position, it overflows to the wrong date. Fix: DateComparisonUtil.applyDateRange computes the set of axis cells that have current-year data and wraps the existing DateSelector to exclude orphaned cells from the scale entirely. DateComparisonFormat.initPartDate additionally removes orphaned cells from partDates (guarded by a shared-max-date check to avoid affecting per-part real-date cases), and format() returns an empty string for any orphaned cell that still reaches the formatter.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/ViewsheetSandbox.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/ViewsheetSandbox.java
@@ -1110,10 +1110,11 @@ public class ViewsheetSandbox implements Cloneable, ActionListener {
       if(dateComparisonInfo == null) {
          boolean dcEnabled = !(info instanceof DataVSAssemblyInfo) ||
             ((DataVSAssemblyInfo) info).isDateComparisonEnabled();
+         VSAssembly shareAssembly = (vs != null && !Tool.isEmptyString(info.getComparisonShareFrom()))
+            ? vs.getAssembly(info.getComparisonShareFrom()) : null;
          boolean validShare = Tool.isEmptyString(info.getComparisonShareFrom()) ||
-            (vs != null && vs.getAssembly(info.getComparisonShareFrom()) != null &&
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):

PR #3820, opened 2026-06-01
Bug #75249, fix TernServer not a constructor error in script panes after esbuild migration

jshobe-inetsoft:
## Summary

- `DefaultCodemirrorService.createTernServer` throws `TypeError: CodeMirror2.TernServer is not a constructor` when opening script panes (VPM trigger, Composer script editor, portal formula editor)
- Root cause: esbuild's `__toESM()` snapshots the CJS exports when `import * as CodeMirror from "codemirror"` is evaluated; the tern addon side-effect import runs after this snapshot so `TernServer` is never added to the namespace wrapper's getter map
- Fix: access `TernServer` through `(CodeMirror as any).default`, which `__toESM` sets as a live reference to the actual CJS module exports

## Test plan

- [ ] Open a VPM in portal, add a condition with a physical view, select a table, click the Trigger tab — script pane should open without errors
- [ ] Open the Composer, open a viewsheet script editor — should open without errors
- [ ] Open the portal formula editor — should open without errors

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/web/projects/shared/util/codemirror/default-codemirror.service.ts
+++ b/web/projects/shared/util/codemirror/default-codemirror.service.ts
@@ -41,7 +41,11 @@ import { CodemirrorService, TokenType } from "./codemirror.service";
 })
 export class DefaultCodemirrorService extends CodemirrorService {
    createTernServer(options: object): object {
-      return new CodeMirror.TernServer(options);
+      // esbuild's __toESM() snapshots the CJS exports at import time, so TernServer
+      // (added by the tern side-effect import after the snapshot) is absent from the
+      // namespace wrapper. Access via .default, which is the live CJS exports reference.
+      const CM: any = (CodeMirror as any).default || CodeMirror;
+      return new CM.TernServer(options);
    }
 
    getEcmaScriptDefs(): object[] {


jkidacki-inetsoft (approved):

jshobe-inetsoft:
## Code Review

### Summary

This PR fixes `TypeError: CodeMirror2.TernServer is not a constructor` in script panes after the Angular 17→18 esbuild migration. The root cause analysis is accurate and the fix is correct.

---

### Code Analysis

**File changed:** `web/projects/shared/util/codemirror/default-codemirror.service.ts`

**Change:**
```typescript
// Before
return new CodeMirror.TernServer(options);

// After
const CM: any = (CodeMirror as any).default || CodeMirror;
return new CM.TernServer(options);
```

**Correctness:** The fix is technically sound. When esbuild processes `import * as CodeMirror from "codemirror"`, its `__toESM()` wrapper creates a namespace object that snapshots the CJS exports at evaluation time. Since the tern addon side-effect import runs after that snapshot, `TernServer` is absent from the namespace getter map. However, `(CodeMirror as any).default` holds a live reference to the original CJS `module.exports` object, so `TernServer` is accessible after the tern addon registers it.

**Fallback safety:** The `|| CodeMirror` fallback preserves backward compatibility with non-esbuild bundlers (e.g., webpack), where `CodeMirror.default` would be undefined but `CodeMirror.TernServer` would still be accessible directly.

**Type safety:** The `any` casts are necessary because the `@types/codemirror` type definitions don't expose `default` or `TernServer` on the `import *` namespace type. This is acceptable given the nature of the workaround.

**Comment quality:** The comment explaining the esbuild snapshot behavior is appropriate here — this is exactly the kind of non-obvious bundler internality that warrants documentation.

---

### Issues

#### Blocking: Failing CI Check

The validation build is failing due to a test timeout unrelated to this PR:

```
FAIL  portal  projects/portal/src/app/composer/dialog/vs/data-input-pane.spec.ts
  > Data Input Pane Test > should input page number on popup table page
  Error: Test timed out in 5000ms.
```

This test (`data-input-pane.spec.ts`, Bug #19833) tests popup table page navigation and has no connection to CodeMirror or TernServer. The timeout is a pre-existing flakiness issue — the test uses `fixture.whenStable().then(...)`, which relies on zone.js async task completion and is inherently timing-sensitive under Vitest.

The code changes in this PR are not the cause of this failure, but the CI check must pass before merging.

**Resolution options:**
1. Re-run the failing CI workflow — if the timeout is intermittent, a re-run may succeed.
2. Fix the flaky test separately (increase its timeout or refactor away the `new Promise`/`whenStable` pattern).

---

### Conclusion

The code change itself is correct, minimal in scope, and well-documented. The CI failure is in an unrelated flaky test. Once that is resolved (via re-run or fix), this PR is ready to merge.

PR #3822, opened 2026-06-01
Bug #75096 - Snap to the bar nearest the cursor in multi-measure bar charts

FigurePaper:
Problem

On a non-stacked bar chart with multiple Y measures (e.g. Sum(customer_id), Sum(orderno), Sum(product_id) grouped by state), enabling the Snap tooltip always snapped to the third/last bar at every x-axis point, and the combined tooltip always showed the last measure as the header — no matter which bar the cursor was over.

Root cause

Snap auto-enables combined tooltips. The server builds a separate combined card for every bar, each headed by its own measure. Visual objects are iterated in reverse order, so the last measure's region receives the smallest region index. The frontend snap path replaced hit-tested regions with the entire rowIdx bucket (all bars at that x) and then emitted the smallest-index region as the
header — always the last measure. Nothing considered where the cursor was within the group, so it could not target an individual bar.

Fix

Frontend only, in chart-plot-area.component.ts. The nearest x-group selection and the server-built per-bar combined cards are unchanged. Within the snapped bucket, a new helper picks the region nearest the cursor and routes it to the tooltip header, the dashed guideline anchor, and the inline-SVG highlight. Selection is limited to the narrowest-width tier in the bucket so wide area
polygons (whose bounding-box center is the series midpoint) are excluded; for bars every bar qualifies and X-distance selects the right one. Near-equal X ties break by vertical distance so overlapping line/area series points pick the nearest one.

Snap remains available for these charts. The flyover and data-tip paths still receive the full bucket; only the plain tooltip header, guideline, and highlight become bar-specific

### DIFF
--- a/web/projects/portal/src/app/composer/dialog/vs/data-input-pane.spec.ts
+++ b/web/projects/portal/src/app/composer/dialog/vs/data-input-pane.spec.ts
@@ -166,7 +166,7 @@ describe("Data Input Pane Test", () => {
    });
 
    //Bug #19833
-   it("should input page number on popup table page", () => new Promise<void>((done) => {
+   it("should input page number on popup table page", async () => {
       dataInputPane = <DataInputPane>fixture.componentInstance;
       dataInputPane.model = createModel();
       dataInputPane.model.rowValue = "1";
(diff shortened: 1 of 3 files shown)


FigurePaper:
Items 1, 3, 5, and 6 are addressed in the latest change. On the three I'm leaving as-is:

2. (width tolerance): keeping the absolute 1px slack. eventX, the region pts, and
centroid are all in the chart's model coordinate space — the same space — and
viewsheetScale is applied later at draw time, so the tolerance is density-independent.
Area polygons span the full series and are far wider than the bar/marker tier, so 1px
reliably separates them; a multiplicative threshold would misbehave when minWidth is 0
(centroid-only markers).

4. (centroid-only width:0 comment): leaving regionBoundsX untouched — that line predates
this change and isn't part of this fix, and the zero-width sentinel is already handled
by the narrowest-tier filter.

7. (RECT_PATH wide-polygon test): not adding one. The wide tier in real charts is always
an area polygon (general path), which the existing test covers; RECT_PATH regions are
bars/markers and are narrow, so a wide RECT_PATH case isn't reachable.

jshobe-inetsoft (changes requested):
The Claude automated review approves the logic, but the **Validation Build check is failing**. Please investigate and fix the build failure before this PR can be merged.

FigurePaper:
On item 4 (guideline anchor): keeping it tracking the nearest bar. That's the intended behavior here — for multi-measure bars the dashed guideline lands on the hovered bar rather than a fixed group center, which reads correctly; the group X tick is still used for choosing the bucket. Items 1, 2, and 3 are addressed in the latest change (return type is now | null, the pixel tolerances are commented, and a Y-distance tie-break test was added). Items 5 and 6 are informational.

jshobe-inetsoft (approved):
The Claude automated review approves this PR. All prior blocking issues have been resolved across multiple review iterations. The most recent review (after the latest commit) confirms the implementation is sound, with only minor/informational items remaining. Approving for squash merge.

PR #3823, opened 2026-06-01
Bug #75211, fix table row highlight foreground color lost on export

jkidacki-inetsoft:
When exporting a viewsheet to PDF/PNG, row highlight foreground colors were not applied to table cells. The HighlightTableLens rowcache was populated with null values before the querySandbox was set (during table lens initialization), causing parameter-dependent conditions to evaluate as unmatched. The subsequent setQuerySandbox call in getRegionTableLens arrived too late - the cached nulls suppressed re-evaluation on every later getForeground call.

Fix: clear the rowcache and cellcache in setQuerySandbox whenever the sandbox value changes, ensuring stale pre-sandbox evaluations are discarded and conditions are correctly re-evaluated with the sandbox.


jshobe-inetsoft (approved):

PR #3826, opened 2026-06-01
Bug #75215, Fix null value missing brush effect in circle packing chart by restricting the all-null guard in getOverlayId to polar charts where the guard is needed

ngovus-inetsoft:


jshobe-inetsoft (approved):

PR #3828, opened 2026-06-01
Bug #75242, fix schedule Options tab content not showing after Angular 21 upgrade

jshobe-inetsoft:
## Summary

- **Angular Material 21 requires `MAT_DATE_FORMATS`**: `MatDatepickerInput/Toggle/Picker` now inject `MAT_DATE_FORMATS` as a required token (previously optional). The schedule route only provided `DateAdapter` via `CustomNativeDateAdapter`, causing `NG0201` to destroy the datepicker view on every change-detection cycle. Added `MAT_NATIVE_DATE_FORMATS` and `FirstDayOfWeekService` (a dependency of `CustomNativeDateAdapter`) to the schedule route providers.
- **`@if` condition used `model?.enabled` instead of form control**: The `@if (model?.enabled)` guard read from `this._model` which is `null` until the model setter fires. Angular Material 21 renders tab content lazily via portals, so the component is created before the model input is propagated, leaving `_model` null while the toggle appeared ON from the form's initial value. Changed to `@if (showOptions)` backed by the form control value, which is always in sync with the toggle.

## Test plan

- [ ] Open the EM Schedule page and edit an existing enabled task
- [ ] Click the Options tab — verify Delete/Start From/Stop On/Description/Locale/Owner/Execute As fields are all visible
- [ ] Toggle "Enabled" off — verify the fields below hide
- [ ] Toggle "Enabled" back on — verify the fields reappear
- [ ] Create a new task and verify the Options tab behaves the same way
- [ ] Verify no `NG0201` errors appear in the browser console

### DIFF
--- a/web/projects/em/src/app/settings/schedule/schedule.routes.ts
+++ b/web/projects/em/src/app/settings/schedule/schedule.routes.ts
@@ -16,11 +16,12 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 import { Routes } from "@angular/router";
-import { DateAdapter, ErrorStateMatcher } from "@angular/material/core";
+import { DateAdapter, ErrorStateMatcher, MAT_DATE_FORMATS, MAT_NATIVE_DATE_FORMATS } from "@angular/material/core";
 import { MatPaginatorIntl } from "@angular/material/paginator";
 import { LocalizedMatPaginator } from "../../../../../shared/util/localized-mat-paginator";
 import { authorizationGuard } from "../../authorization/authorization-guard.service";
 import { CustomNativeDateAdapter } from "../../common/util/datepicker/custom-native-date-adapter.service";
+import { FirstDayOfWeekService } from "../../../../../portal/src/app/common/services/first-day-of-week.service";
 import { EmErrorStateMatcher } from "../../common/util/error/em-error-state-matcher";
 import { MonitoringDataService } from "../../monitoring/monitoring-data.service";
 import { scheduleConfigSaveGuard } from "./schedule-configuration-page/schedule-config-save.guard";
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft:
## Code Review

### Summary

The changes correctly address two Angular Material 21 compatibility issues in the schedule task options pane. The logic is sound and the fixes are minimal in scope. However, one blocking issue is preventing the build from passing.

---

### Blocking Issues

#### Lint Error: Single Quote in `showOptions` Getter

**File:** `web/projects/em/src/app/settings/schedule/task-options-pane/task-options-pane.component.ts`, line 207

The new `showOptions` getter uses a single-quoted string literal, violating the project's `@typescript-eslint/quotes` rule (double quotes required):

```typescript
// Current (causes lint error)
get showOptions(): boolean {
   return !!this.optionsForm.get('taskEnabled').value;
}

// Fix
get showOptions(): boolean {
   return !!this.optionsForm.get("taskEnabled").value;
}
```

This is the sole cause of the Validation Build failure (`✖ 1 problem (1 error, 0 warnings)` in the `em` lint step).

---

### Non-Blocking Observations

**`schedule.routes.ts`**
- Adding `MAT_NATIVE_DATE_FORMATS` via `MAT_DATE_FORMATS` and `FirstDayOfWeekService` as route-level providers is the correct fix. `CustomNativeDateAdapter` already injects `FirstDayOfWeekService`, and Angular Material 21 changed `MAT_DATE_FORMATS` from optional to required. The import path for `FirstDayOfWeekService` follows the existing pattern used in `custom-native-date-adapter.service.ts` and `dayjs-date-adapter.service.ts`.

**`task-options-pane.component.html`**
- Replacing `@if (model?.enabled)` with `@if (showOptions)` is correct. The old guard read from `_model` which is `null` during Angular Material 21's lazy portal rendering; the form control is always initialized and stays in sync.

**`task-options-pane.component.ts`**
- Changing the `taskEnabled` initial value from `[true]` to `[false]` is the right default. Without a loaded model, the form starts disabled; once the model setter fires and calls `setValue(this.model.enabled || false)`, the form control (and thus the UI) correctly reflects the actual task state.
- The `showOptions` getter reading `optionsForm.get("taskEnabled").value` is safe — the control is always present after construction. The double-cast `!!` correctly handles any falsy initial state.

---

### Failing Checks

- **Validation Build** — FAILURE (lint error described above)

To fix: change `'taskEnabled'` to `"taskEnabled"` in the `showOptions` getter.

jkidacki-inetsoft (approved):

PR #3833, opened 2026-06-01
Fix ObjectUnsubscribedError in remaining em components

jshobe-inetsoft:
## Summary

- Extends the fix from #3827 to all remaining em project components and services that used `destroy$.unsubscribe()` in `ngOnDestroy`
- 27 files updated across the em project

## Root cause

Same root cause as #3827: `Subject.unsubscribe()` sets `closed = true` in RxJS 7, which throws `ObjectUnsubscribedError` when any subsequent code tries to `subscribe()` via `takeUntil(this.destroy$)`. This is reproducible in fast test environments like Vitest where the full component lifecycle can complete quickly, allowing a pending async operation (e.g. `setTimeout`, deferred HTTP response) to fire after `ngOnDestroy` has already closed the Subject.

`Subject.complete()` correctly sets `isStopped = true` but leaves `closed = false`, so new subscriptions still work — they just receive an immediate completion signal.

## Test plan

- [ ] Run `npm run test:em` and confirm no `ObjectUnsubscribedError` appears (specifically the `repository-viewsheet-settings-view.component.spec.ts` test)
- [ ] Confirm existing tests continue to pass

### DIFF
--- a/web/projects/em/src/app/auditing/auditing-sidenav/auditing-sidenav.component.ts
+++ b/web/projects/em/src/app/auditing/auditing-sidenav/auditing-sidenav.component.ts
@@ -99,7 +99,7 @@ export class AuditingSidenavComponent implements OnInit, OnDestroy {
 
    ngOnDestroy(): void {
       this.destroy$.next();
-      this.destroy$.unsubscribe();
+      this.destroy$.complete();
    }
 
    isScreenSmall(): boolean {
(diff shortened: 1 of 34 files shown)


jkidacki-inetsoft (dismissed):

jshobe-inetsoft:
Fixed. Both  in `top-scroll.service.ts` and `refreshSubject.unsubscribe()` in `organization-dropdown.service.ts` changed to `.complete()` in 45ba0b7.

The `backup-file.component.ts` observation (missing `destroy$.complete()`) is pre-existing and not introduced by this PR — leaving it out of scope.

jkidacki-inetsoft (approved):

PR #3838, opened 2026-06-02
Bug #75186, fix cloud runner Ignite cache mode mismatch when joining cluster

jkidacki-inetsoft:
Copy minNodes from the server's ClusterConfig into the cloud runner's config so that getDefaultCacheMode() returns the same CacheMode (PARTITIONED vs REPLICATED) as the server, preventing an Ignite AtomicConfiguration conflict on cluster join.


jshobe-inetsoft (approved):
The Claude automated review explicitly approves this PR: "✅ Approved for merge. The fix is correct, minimal, complete, and carries negligible regression risk." The one-line change ensures the cloud runner copies `minNodes` from the server's cluster config so both sides compute the same Ignite cache mode. Approving for squash merge.

PR #3841, opened 2026-06-02
Bug #75181, Provide SecurityProviderService in SECURITY_PROVIDER_ROUTES parent scope so detail page components can inject it

ngovus-inetsoft:


jshobe-inetsoft (approved):
The Claude automated review approves this PR. The fix correctly uses a component-less parent route to scope `SecurityProviderService` and `ErrorHandlerService` to both the list page and detail pages. The only observation noted (missing `pathMatch: 'full'`) is pre-existing and explicitly not a blocker. Approving for squash merge.

PR #3843, opened 2026-06-02
Bug #75028 - Fix legend symbol overlapping the legend border when symbol size is large

FigurePaper:
## Summary

Fixes legend symbols overlapping the legend border. When a legend has a large
symbol size and a border, and two legends are stacked vertically, the upper
legend's symbols overflowed past its bottom border and leaked into the legend
below.

## Cause

In the viewer, each legend is a CSS-bordered wrapper containing a fixed-height
content area that holds a server-rendered tile of the full item list (sized from
Legend.getContentPreferredBounds). When a legend is allocated less height than
its content needs (two legends sharing limited space), the full-height content
exceeds the allocated legend bounds. The bordered wrapper had no clipping, so
the overflow row painted past the bottom border into the adjacent legend.

## Changes

- chart-legend-container.component.scss: add overflow: hidden to
  .chart-legend-wrapper so legend content is clipped to the bordered box. The
  inner content area keeps its overflow-y:auto, so items beyond the visible
  height are still scrollable.
- Legend.getContentPreferredBounds: add a bottom inset (GAP) below the last row
  so the last symbol isn't flush against the bottom border when scrolled to the
  end, matching the existing top/left insets.
- Add LegendContentInsetTest covering the bottom inset.

### DIFF
--- a/core/src/main/java/inetsoft/graph/guide/legend/Legend.java
+++ b/core/src/main/java/inetsoft/graph/guide/legend/Legend.java
@@ -371,7 +371,8 @@ public abstract class Legend extends BoundedContainer {
          double itemh = item0 == null ? 20 : item0.getPreferredHeight();
 
          contentW = w;
-         contentH = getItems().length * itemh;
+         // bottom inset so the last row's symbol isn't flush with the border when scrolled
+         contentH = getItems().length * itemh + GAP;
          contentX = x;
          contentY = title != null ?
             y + h - title.getSize().getHeight() - GAP - contentH :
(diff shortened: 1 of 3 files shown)


FigurePaper:
Addressed the test points and verified the two watch items.

- VGraphPair image sizing: both call sites derive the image dimensions directly from
getContentPreferredBounds() (width/height come straight from the returned bounds), so
no caller hard-codes or clamps the pre-inset height. The image is simply 4px taller,
which is the intended inset space.
- Wrapper clipping vs. resize/move: the drag outline uses position: fixed so it isn't
affected by the new overflow: hidden, and resize/move is driven by the interact
directive's edge detection rather than children painting outside the wrapper. Will
exercise it manually in resize/move mode as suggested.
- Test: added a guard asserting at least one legend item is found, documented the 3.5
threshold relative to GAP, and relabeled the assertion message to make clear getY() is
the bottom edge in this Y-up coordinate system.

jshobe-inetsoft (approved):
The Claude automated review approves this PR. All review items from the initial review were addressed in the follow-up commit, and the re-review confirms the fix is sound, minimal, and the test is now robust. Approving for squash merge.

PR #3845, opened 2026-06-02
Bug #75285: Reject datasource creation when parent folder does not exist

ngovus-inetsoft:
﻿## Summary

- Added backend validation in DatabaseDatasourcesService.saveDatabaseDefinition that rejects datasource creation when the specified parent folder path is non-empty but does not exist in the registry
- Returns ConnectionStatus("Invalid Folder") in this case, preventing phantom datasources from being silently stored at inaccessible paths
- Added corresponding error handler in the Angular DataSourceSettingsPage base class to display the error to the user
- Added i18n string data.datasources.invalidParentFolder with a message that also documents the correct empty-string path for root-level creation

**Root cause:** POST /api/data/databases?path=Data%20Source&create=true was accepted silently. "Data Source" is only the display label for the root node (whose real path is "/"), not an actual folder. The datasource was stored at "Data Source/{name}" -- a path invisible in the GUI tree.

**Correct API call for root-level creation:** POST /api/data/databases?path=&create=true

## Test plan

- [ ] POST /api/data/databases?path=Data%20Source&create=true returns {"status":"Invalid Folder"}
- [ ] POST /api/data/databases?path=&create=true creates datasource at root and it appears in the tree
- [ ] POST /api/data/databases?path=RealFolder&create=true still works when RealFolder exists
- [ ] Creating a datasource through the normal GUI flow still works

Generated with Claude Code

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/content/repository/DatabaseDatasourcesService.java
+++ b/core/src/main/java/inetsoft/web/admin/content/repository/DatabaseDatasourcesService.java
@@ -399,6 +399,11 @@ public class DatabaseDatasourcesService {
       if(dataSource == null) {
          // Create the dataSource - path is the parent folder's path
          fullName = fullName.startsWith("/") ? fullName.substring(1) : fullName;
+
+         if(!fullName.isEmpty() && registry.getDataSourceFolder(fullName) == null) {
+            return new ConnectionStatus("Invalid Folder");
+         }
+
          fullName += !fullName.isEmpty() ? "/" + name : name;
 
          if(registry.getDataSourceFolder(fullName) != null) {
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (changes requested):
The Claude automated review posted after the most recent commit identifies a **Critical** blocking issue:

**Non-JDBC error is not surfaced to the user**: The non-JDBC path in `DatasourcesBaseService.createNewDataSource` throws a `FileNotFoundException`, which the Angular frontend's error handler cannot parse — the user sees the generic "An error prevented the datasource from being saved" message instead of the specific invalid folder message. The `data-source-settings-page.ts` `"Invalid Folder"` handler added in this PR only covers the JDBC/EM path (`../api/data/databases`), not the non-JDBC portal path (`../api/portal/data/datasources`).

**Suggested fix**: Replace the `FileNotFoundException` in `DatasourcesBaseService.java` with a `MessageException` using `Catalog.getCatalog().getString("data.datasources.invalidParentFolder")`. This would align both paths and allow the Angular frontend to display the specific error message.

jshobe-inetsoft (approved):
The latest Claude review confirms the non-JDBC path has been fixed using MessageException in commit d7d2c0e7 — all code paths are now covered and the implementation is correct. Approving for merge.

PR #3870, opened 2026-06-03
Bug #75294 - Keep Snap to nearest data point enabled when Combined is off

FigurePaper:
Unchecking the Combined option in the chart tooltip customize dialog
incorrectly disabled the Snap to nearest data point checkbox. The two
options are independent: snap applies in the Default/Auto and Custom
tooltip modes whether or not tooltips are combined, and is only
irrelevant when the tooltip mode is None or the chart type does not
support snapping.

Changes in tip-customize-dialog.component.ts:

- Remove the branch in the Combined value-change handler that disabled
  and cleared snap when Combined was turned off. Turning Combined on
  still enables and checks snap, since a combined tooltip inherently
  snaps to a data point.
- Restore enabling snap in the Custom branch of the tooltip-mode handler,
  so snap is re-enabled when switching from None to Custom rather than
  staying stuck disabled.

After the change, snap's enabled state is governed solely by the existing
snapDisabled rule (tooltip mode is not None and snap is supported), which
never references Combined.

Testing:

Added tip-customize-dialog.component.spec.ts covering: snap stays enabled
when Combined is turned off, Combined on enables and checks snap, None to
Custom re-enables snap, switching to None disables and clears snap, and
snap stays disabled when not supported.

### DIFF
--- a/web/projects/portal/src/app/widget/dialog/tip-customize-dialog/tip-customize-dialog.component.spec.ts
+++ b/web/projects/portal/src/app/widget/dialog/tip-customize-dialog/tip-customize-dialog.component.spec.ts
@@ -1,0 +1,87 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #3872, opened 2026-06-03
Bug #75297 - Preserve combined tooltip when switching through an unsupported chart type

FigurePaper:
Switching a chart to a type that does not support combined tooltips
(point, pie, etc.) permanently cleared the combined-tooltip setting. A
user who enabled Combined on a bar chart, switched to point, then back to
line found Combined turned off, while Snap to nearest data point survived
the same round trip. The two options should behave the same way.

The combined-tooltip value was being cleared defensively because the
render path did not check whether the chart type supports it. This change
gates rendering instead of destroying the value:

- ChangeChartTypeProcessor no longer calls setCombinedToolTipValue(false)
  when switching to a type that does not support combined tooltips. The
  design value is preserved across every type change, like snap-tooltip.
- PlotArea now checks supportsCombinedTooltip() at the two sites that
  consume isCombinedToolTip() (the combined point/text tooltip map and the
  stack-total measure check), mirroring the supportsSnapTooltip() gate
  GraphBuilder already applies to snap. A value retained on an unsupported
  type stays dormant and renders again only when the type supports it.

Combined tooltips render on the same chart types as before; the only
change is that the user's setting is no longer lost when passing through
an unsupported type.

Testing:

ChangeChartTypeCombinedTooltipTest now asserts the combined-tooltip value
is preserved on a switch to an unsupporting type, with a new
bar -> point -> line round-trip test. The existing supporting-type
preservation cases are unchanged.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/region/PlotArea.java
+++ b/core/src/main/java/inetsoft/report/composition/region/PlotArea.java
@@ -285,8 +285,8 @@ public class PlotArea extends GridContainerArea implements GraphComponentArea, R
       Map pointValueMap = null; // maps point values to pointVO indexes in the vos list
       Map textValueMap = null; // maps text value indexes to textVO indexes in the vos list
 
-      if(chartInfo != null && chartInfo.isCombinedToolTip() && chartInfo.getToolTip() == null &&
-         !chartInfo.isMultiStyles())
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):

PR #3874, opened 2026-06-03
Bug #75278, radar chart facet cell for null dimension value not displayed

ngovus-inetsoft:
## Summary
- Radar chart with a boolean facet dimension (e.g. `reseller`) was silently dropping the null-valued facet cell, showing only `false` and `true` columns instead of three columns including an empty radar circle for null
- Root cause: `TileCleaner` marks a tile as invalid when `containsValue(measures)` returns false — which happens when all calc measure values (e.g. *Change from previous*) are null for that group. The null-reseller tile has real data rows but a null calc result, so it was incorrectly removed
- Fix: treat a tile as valid if its sub-dataset reports `getRowCount() > 0`. Phantom change-from-previous tiles are unaffected because `AbstractDataSet.getRowCount()` only counts projected rows when real rows already exist, leaving genuine phantom tiles at count 0

## Test plan
- [ ] Open a radar chart with a boolean dimension (e.g. `reseller`) on the X axis that has `true`, `false`, and null values in the data, and a calc measure (Change from previous) on Y — confirm three facet cells render, including an empty circle for null
- [ ] Verify that a chart using *Change from previous* with project-forward still correctly omits phantom projected facet cells (no regression)
- [ ] Confirm non-radar faceted charts with null dimension values also display the null cell correctly

### DIFF
--- a/core/src/main/java/inetsoft/graph/coord/TileCleaner.java
+++ b/core/src/main/java/inetsoft/graph/coord/TileCleaner.java
@@ -66,12 +66,22 @@ class TileCleaner {
       }
 
       for(int i = tiles.length - 1; i >= 0 && tiles.length > 1; i--) {
+         // never remove a row whose scale value is null — null is a real dimension value
+         if(isNullScaleValue(yscale, i)) {
+            continue;
+         }
+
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):

PR #3877, opened 2026-06-03
Bug #75292 - Fix smooth lines truncated before last data point in animated charts

FigurePaper:
Summary

Smooth/curved lines on line and area charts stopped short of the final
x-axis data point during the chart draw-on animation. The line stroke appeared to end partway
through the last segment instead of reaching the last point; straight lines drew correctly.

Root cause

The chart rendering is correct — LineVO/AreaVO emit cubic Bézier paths that reach the last
point. The truncation is in the SVG draw-on animation.
SVGAnimationDOMInjector.applyLineDrawAnimation animates a solid line with
stroke-dasharray/stroke-dashoffset, using SVGAnimationInjector.computePathLength as the
length and assuming it is always at least as long as the rendered path.

computePathLength approximated each cubic (C) and quadratic (Q) segment by its straight chord
distance rather than its arc length. A curve's arc is longer than its chord, so for a wiggly
smooth line the computed length under-estimated the true path. The browser lays the dash
pattern along the true arc, so the difference (the tail of the line) landed in the dash gap
and was never painted. Because the animation uses fill-mode: both, the truncation persisted
after the animation finished.

This was latent until the smooth/curved lines feature: before smooth lines, every line path was straight (M/L only),
where chord length equals arc length, so the C/Q branches were never meaningfully exercised.

Fix

computePathLength now measures cubic and quadratic segments by flattening each curve into 32
sampled sub-segments and summing their lengths, restoring the invariant that the dasharray
length is at least the rendered path length (the existing +2 margin covers the sub-pixel
polyline deficit). Straight paths are unchanged. The method has a single caller, so returning
the larger, accurate length is safe.

### DIFF
--- a/utils/inetsoft-xml-formats/src/main/java/inetsoft/util/graphics/animation/SVGAnimationInjector.java
+++ b/utils/inetsoft-xml-formats/src/main/java/inetsoft/util/graphics/animation/SVGAnimationInjector.java
@@ -97,16 +97,22 @@ public final class SVGAnimationInjector {
             }
          }
          else if("C".equals(c)) {
+            // Sum sampled sub-segments, not the chord: a curve's arc length exceeds the
+            // straight start→end distance, and underestimating it truncates the dashoffset
+            // draw-on (the tail falls into the dash gap) on smooth lines.
             for(int j = 0; j + 5 < args.size(); j += 6) {
+               double c1x = args.get(j), c1y = args.get(j + 1);
+               double c2x = args.get(j + 2), c2y = args.get(j + 3);
                double ex = args.get(j + 4), ey = args.get(j + 5);
-               totalLen += Math.hypot(ex - cx, ey - cy);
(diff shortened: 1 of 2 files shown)


FigurePaper:
Implemented the two inline items: relative c/q handling in computePathLength (matching the
existing relative m/l branches) and quadratic arc-length tests plus a relative-vs-absolute
equivalence check.

Two points left as-is:

buildFillPolygon comment — leaving for a separate follow-up. The method is unrelated to
this truncation fix and, as you noted, its logic is already correct; only the comment
over-warns. Correcting unrelated documentation here would widen this fix's scope.

CURVE_FLATTEN_STEPS placement — keeping it adjacent to cubicLen/quadLen, its only callers,
for locality. I did make its comment name applyLineDrawAnimation so the +2 margin
dependency across files is explicit.

jshobe-inetsoft (approved):

PR #3878, opened 2026-06-03
Bug #75263, fix VS_ASSEMBLY adhoc slider applying aggregate filter as WHERE instead of HAVING

ngovus-inetsoft:
## Summary

- When a VS_ASSEMBLY TimeSlider is created from a chart aggregate measure bar, moving the slider caused the chart to show multiple raw data rows per dimension value instead of one aggregated bar per group.
- Root cause 1: mergePostCondition re-read AggregateInfo from the mirror table (always empty) rather than the original table, causing isAggregate() to return false and a HAVING condition (Sum(x) >= N) to be downgraded to a WHERE clause applied before aggregation.
- Root cause 2: MirrorTableAssembly.update() looked up the underlying VS table by name from the WorksheetWrapper, finding the original unmodified version (empty AggregateInfo) rather than the configured clone with the chart's GROUP BY and aggregate bindings, causing the mirror's inner query to return raw unaggregated rows.
- Fix: capture originalAggregated before mirror creation in mergePostCondition and use it in the pre/post condition decision; call ws.addAssembly(table) before constructing mirrors in both mergePostCondition and TimeSliderVSAQuery.createAssemblyTable so update() resolves to the correctly configured table.

## Test plan

- [ ] Open a viewsheet with a bar chart bound to a worksheet query with a dimension (e.g. state) and an aggregate measure (e.g. Sum(customer_id))
- [ ] Click a bar to create an adhoc filter — verify a VS_ASSEMBLY RangeSlider appears
- [ ] Move the slider to a non-default range — verify the chart still shows one aggregated bar per dimension value (not multiple raw rows), with states outside the range filtered out
- [ ] Confirm a manually-placed viewsheet RangeSlider (bound to the raw field) still works correctly
- [ ] Confirm charts using Change from previous / date comparison still omit phantom projected facet cells (no regression from bug 75278)

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/TimeSliderVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/TimeSliderVSAQuery.java
@@ -337,6 +337,12 @@ public class TimeSliderVSAQuery extends AbstractSelectionVSAQuery {
       // name should start with V_, otherwise PreAssetQuery.getAggregateInfo
       // will return empty
       table.setProperty("Component_Binding_Table", "true");
+      // Register the bindable table (with chart's AggregateInfo) so that the slider
+      // mirror's update() finds this version rather than the original unmodified VS
+      // table from the WorksheetWrapper. (75263)
+      if(ws != null) {
+         ws.addAssembly(table);
+      }
       String mname = Assembly.TABLE_VS + slider.getName() + "_" + table.getName();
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
Approving for squash merge. Claude review is resolved (minimized) and all checks pass.

PR #3879, opened 2026-06-04
fixed bug 75173

bonnieshi43:
The root cause: in PublishMetricsTask.run() (line 68), when cluster.getClusterNodes() returns an empty set at startup (before
  nodes register), count stays 0 and Java double division 0.0 / 0 produces NaN. CloudWatch rejects any metric datum with a NaN
  value.

  Fix: added an early return when count == 0 — if there are no cluster nodes yet, there's nothing meaningful to publish, so we
  skip the call entirely rather than sending an invalid value to CloudWatch.

  This happens specifically at startup when the cluster node list is temporarily empty before the local node has registered
  itself.


jshobe-inetsoft (approved):

PR #3880, opened 2026-06-04
restore theming after angular 21 - user portal and lost by pattern -  recovery-0603

swaker854:
1) Fixed portal toolbar and data tab. 2) Restore 20 Angular template losses from 563ab05dd merge conflict resolution identified by a pattern-count audit across  (shell-form-group, shell-form-row, custom-select, number-stepper, landing-card, landing-shell, enterClick) in the Pre Angular version and HEAD.

### DIFF
--- a/web/projects/em/src/app/settings/presentation/portal-integration-view/portal-integration-view.component.html
+++ b/web/projects/em/src/app/settings/presentation/portal-integration-view/portal-integration-view.component.html
@@ -49,41 +49,62 @@
                 <mat-icon fontSet="ineticons" fontIcon="shape-plus-icon"></mat-icon>
               </button>
             </section>
-            <section class="flex-row">
-              @for (tab of model.tabs; track tab; let i = $index) {
-                <mat-card class="tab-block portal-card">
-                  <mat-card-subtitle class="visible-subtitle">
-                    <h4 class="tab-name mat-subtitle-2">{{tab.label}}</h4>
-                    <span class="spacer"></span>
-                    <button mat-icon-button [title]="tab.visible ? '_#(Visible)' : '_#(Hidden)'"
-                      [disabled]="!tab.editable" (click)="tab.visible = !tab.visible; emitModel()">
-                      <mat-icon fontSet="ineticons" [fontIcon]="tab.visible ? 'eye-icon' : 'eye-off-icon'"></mat-icon>
-                    </button>
-                    <button [disabled]="moveUpDisabled(i)" mat-icon-button
-                      title="_#(Move Up)"
-                      (click)="moveUp(i)">
-                      <mat-icon matListIcon fontSet="ineticons" fontIcon="shape-arrow-up-bold-icon"></mat-icon>
-                    </button>
-                    <button [disabled]="moveDownDisabled(i)" mat-icon-button
-                      title="_#(Move Down)"
-                      (click)="moveDown(i)">
-                      <mat-icon matListIcon fontSet="ineticons" fontIcon="shape-arrow-down-bold-icon"></mat-icon>
-                    </button>
-                    <button mat-icon-button [disabled]="!tab.editable"
-                      title="_#(Edit)"
-                      (click)="edit(tab)">
-                      <mat-icon matListIcon fontSet="ineticons" fontIcon="edit-icon"></mat-icon>
-                    </button>
-                    <button mat-icon-button [disabled]="!tab.editable"
-                      title="_#(Delete)"
-                      (click)="delete(i)">
-                      <mat-icon matListIcon fontSet="ineticons"fontIcon="trash-icon"></mat-icon>
-                    </button>
-                  </mat-card-subtitle>
-                  <mat-divider></mat-divider>
-                </mat-card>
-              }
+            <section>
+              <h4 class="mat-subtitle-2 nav-group-label">_#(Left Nav)</h4>
+              <div class="flex-row">
+                @for (entry of leftNavTabs; track entry.tab) {
+                  <mat-card class="tab-block portal-card">
+                    <mat-card-subtitle class="visible-subtitle">
+                      <h4 class="tab-name mat-subtitle-2">{{entry.tab.label}}</h4>
+                      <span class="spacer"></span>
+                      <button mat-icon-button [title]="entry.tab.visible ? '_#(Visible)' : '_#(Hidden)'"
+                        [disabled]="!entry.tab.editable" (click)="entry.tab.visible = !entry.tab.visible; emitModel()">
+                        <mat-icon fontSet="ineticons" [fontIcon]="entry.tab.visible ? 'eye-icon' : 'eye-off-icon'"></mat-icon>
+                      </button>
+                      <button [disabled]="moveUpDisabled(entry.modelIndex)" mat-icon-button
+                        title="_#(Move Up)"
+                        (click)="moveUp(entry.modelIndex)">
+                        <mat-icon matListIcon fontSet="ineticons" fontIcon="shape-arrow-up-bold-icon"></mat-icon>
+                      </button>
+                      <button [disabled]="moveDownDisabled(entry.modelIndex)" mat-icon-button
+                        title="_#(Move Down)"
+                        (click)="moveDown(entry.modelIndex)">
+                        <mat-icon matListIcon fontSet="ineticons" fontIcon="shape-arrow-down-bold-icon"></mat-icon>
+                      </button>
+                      <button mat-icon-button [disabled]="!entry.tab.editable"
+                        title="_#(Edit)"
+                        (click)="edit(entry.tab)">
+                        <mat-icon matListIcon fontSet="ineticons" fontIcon="edit-icon"></mat-icon>
+                      </button>
+                      <button mat-icon-button [disabled]="!entry.tab.editable"
+                        title="_#(Delete)"
+                        (click)="delete(entry.modelIndex)">
+                        <mat-icon matListIcon fontSet="ineticons" fontIcon="trash-icon"></mat-icon>
+                      </button>
+                    </mat-card-subtitle>
+                    <mat-divider></mat-divider>
+                  </mat-card>
+                }
+              </div>
             </section>
+            @if (dataTab) {
+              <section>
+                <h4 class="mat-subtitle-2 nav-group-label">_#(Right Nav)</h4>
+                <div class="flex-row">
+                  <mat-card class="tab-block portal-card">
+                    <mat-card-subtitle class="visible-subtitle">
+                      <h4 class="tab-name mat-subtitle-2">{{dataTab.tab.label}}</h4>
+                      <span class="spacer"></span>
+                      <button mat-icon-button [title]="dataTab.tab.visible ? '_#(Visible)' : '_#(Hidden)'"
+                        [disabled]="!dataTab.tab.editable" (click)="dataTab.tab.visible = !dataTab.tab.visible; emitModel()">
+                        <mat-icon fontSet="ineticons" [fontIcon]="dataTab.tab.visible ? 'eye-icon' : 'eye-off-icon'"></mat-icon>
+                      </button>
+                    </mat-card-subtitle>
+                    <mat-divider></mat-divider>
+                  </mat-card>
+                </div>
+              </section>
+            }
           </mat-card-content>
         </mat-card>
       }
--- a/web/projects/em/src/app/settings/presentation/portal-integration-view/portal-integration-view.component.scss
+++ b/web/projects/em/src/app/settings/presentation/portal-integration-view/portal-integration-view.component.scss
@@ -35,4 +35,9 @@
   .tab-name {
     word-break: break-all;
   }
+
+  .nav-group-label {
+    margin: 0.5em 0 0.25em;
+    opacity: 0.6;
+  }
 }
--- a/web/projects/em/src/app/settings/presentation/portal-integration-view/portal-integration-view.component.ts
+++ b/web/projects/em/src/app/settings/presentation/portal-integration-view/portal-integration-view.component.ts
@@ -63,13 +63,25 @@ export class PortalIntegrationViewComponent {
    constructor(private dialog: MatDialog) {
    }
 
+   get leftNavTabs(): Array<{tab: PortalTabModel, modelIndex: number}> {
+      return (this.model?.tabs || [])
+         .map((tab, i) => ({tab, modelIndex: i}))
+         .filter(({tab}) => tab.name !== "Data");
+   }
+
+   get dataTab(): {tab: PortalTabModel, modelIndex: number} | null {
+      const i = (this.model?.tabs || []).findIndex(t => t.name === "Data");
+      return i >= 0 ? {tab: this.model.tabs[i], modelIndex: i} : null;
+   }
+
    moveUpDisabled(index: number): boolean {
       return index == 0 || this.model.tabs[index].editable && !this.model.tabs[index - 1].editable;
    }
 
    moveDownDisabled(index: number): boolean {
       return index == this.model.tabs.length - 1 ||
-         !this.model.tabs[index].editable && this.model.tabs[index + 1].editable;
+         !this.model.tabs[index].editable && this.model.tabs[index + 1].editable ||
+         this.model.tabs[index + 1]?.name === "Data";
    }
 
    moveUp(index: number) {
--- a/web/projects/em/src/app/settings/presentation/portal-integration-view/portal-integration-view.component.ts
+++ b/web/projects/em/src/app/settings/presentation/portal-integration-view/portal-integration-view.component.ts
@@ -67,22 +67,24 @@ export class PortalIntegrationViewComponent {
    get leftNavTabs(): Array<{tab: PortalTabModel, modelIndex: number}> {
       return (this.model?.tabs || [])
          .map((tab, i) => ({tab, modelIndex: i}))
-         .filter(({tab}) => tab.name !== PortalTabs.DATA);
+         .filter(({tab}) => !this.isDataTab(tab));
    }
 
    get dataTab(): {tab: PortalTabModel, modelIndex: number} | null {
-      const i = (this.model?.tabs || []).findIndex(t => t.name === PortalTabs.DATA);
+      const i = (this.model?.tabs || []).findIndex(t => this.isDataTab(t));
       return i >= 0 ? {tab: this.model.tabs[i], modelIndex: i} : null;
    }
 
    moveUpDisabled(index: number): boolean {
-      return index == 0 || this.model.tabs[index].editable && !this.model.tabs[index - 1].editable;
+      const previousTab = this.model.tabs[index - 1];
+      return index == 0 ||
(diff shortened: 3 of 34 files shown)


swaker854:
fix npm run verify errors

jshobe-inetsoft (changes requested):
## Code Review

This PR restores ~20 Angular template patterns lost during the `563ab05dd` merge conflict resolution, fixes the portal toolbar, and updates several components to the newer design-system CSS classes (`shell-form-row`, `shell-form-group`, `custom-select-trigger`, etc.).

### Blocking — Cannot Merge

**1. CI checks are not passing**

- `claude-review` — **FAILED** (automated review step failed)
- `build` (Validation Build) — **PENDING** (still in progress)

The PR cannot be merged until all required checks pass.

---

### Code Issues

**2. `dataTabFirst` is now dead code in `app.component.html`**

The template previously used `dataTabFirst` to control whether the Schedule tab appeared before or after the Data tab. After this PR the template hardcodes the order: Schedule always before Data. The variable is still computed in `app.component.ts` (lines 253–258) but is no longer referenced anywhere in the template. The dead computation should either be removed from the TS file or the ordering logic should be restored if it is still needed.

```typescript
// app.component.ts — still computed but never used after this PR
this.dataTabFirst = scheduleIndex <= dataIndex;
```

**3. Behavioral change: tab ordering is now fixed**

The previous template rendered the Schedule and Data tabs in a configurable order depending on their positions in the model (controlled by `dataTabFirst`). This PR replaces the two conditional branches with a single fixed layout (Schedule → Data). This is a UX regression if any deployment expects the Data tab to appear before the Schedule tab. Please confirm this ordering change is intentional for the `epic-74519` branch.

---

### Positive Changes

- **Calendar null guard** (`vs-calendar.component.ts`): Adding `typeof this.calendar1.getSelectionString === "function"` before calling the method is a sound defensive fix.
- **`stat-panel.component.html`**: Adding `*ngIf="model.bandFill"` prevents a potential null-reference when the color model is absent.
- **`vitest-setup.ts`**: Refactoring `_scrubDeps` into `_scrubTestingModuleDef` and patching `TestBed.compileComponents` to re-scrub after async resource resolution correctly handles the timing window where component metadata can be unavailable. The `_isPendingComponentResourceError` guard is appropriately narrow.
- **`label-input-field.spec.ts`**: Test selector updated from `.custom-select` to `.custom-select-trigger` to match the renamed HTML class. Consistent with the other template changes.
- **`*ngIf` → `@if` migration** (`data-tab.component.html`, etc.): Aligned with Angular 17+ control flow best practices.
- **Dropdown caret icons**: Consistent chevron open/close state tracking added across `border-style-pane`, `style-dropdown`, `grid-line-dropdown`, `multi-select`, and `dynamic-combo-box`. Good UX consistency.
- **Schedule tab icon fix** (`app.component.html`): The original merge-conflict code erroneously used `database-icon` for the Schedule tab. The new code correctly uses `datetime-field-icon`.
- **Invalid feedback guards** (`viewsheet-options-pane.component.html`): Error messages now only render when `form.controls[...].errors` is truthy — avoids rendering empty validation spans.
- **`DataDetailsPaneService` provider** (`data-tab.component.ts`): Restoring this to the component providers correctly scopes the service to the data tab subtree.

swaker854:
Move Schedule tab to portal left nav; reflect grouping in EM Portal Tabs UI

Schedule is a user-facing function and belongs in the left nav alongside
Repository and Dashboard. Data stays pinned in the right nav.

Portal (app.component):
- Replace the reportTabFirst double-branch + hardcoded Schedule position
  with a single @for loop over leftNavTabs (portalTabs minus Data and
  custom tabs), so EM's configured order drives the left nav directly.
- Remove reportTabFirst field and setTabOrder(); add leftNavTabs getter.

EM Portal Integration view:
- Split the flat Portal Tabs list into two labelled sections: "Left Nav"
  (Repository, Dashboard, Schedule — reorderable) and "Right Nav" (Data —
  visibility toggle only, no up/down arrows).
- Add leftNavTabs / dataTab getters that carry the original model index so
  moveUp/moveDown still operate on the flat backing array.
- Tighten moveDownDisabled to prevent a left-nav tab from sliding past Data.

jshobe-inetsoft (changes requested):
## Code Review

### Summary

This PR restores Angular theming patterns lost during an Angular 21 upgrade merge-conflict resolution, and refactors the portal toolbar navigation to move the Schedule tab to the left nav. The overall approach is sound, and the build passes. However, there is one failing CI check that blocks approval.

---

### Blocking Issues

#### 1. Failing `claude-review` CI check

The `claude-review` check failed and must be resolved before this PR can be approved.

---

### Non-Blocking Findings

#### 2. Hard-coded `"Data"` string in `portal-integration-view.component.ts`

The new `leftNavTabs` and `dataTab` getters filter by the hard-coded string `"Data"`:

```ts
get leftNavTabs() {
   return (this.model?.tabs || [])
      .map((tab, i) => ({tab, modelIndex: i}))
      .filter(({tab}) => tab.name !== "Data");  // hard-coded
}

get dataTab() {
   const i = (this.model?.tabs || []).findIndex(t => t.name === "Data");  // hard-coded
   ...
}
```

By contrast, `app.component.ts` correctly uses `PortalTabs.DATA`. The EM component does not import `PortalTabs`, but the hard-coded string is fragile — a rename of the constant would silently break this logic. Consider either importing the constant or extracting a shared `DATA_TAB_NAME` literal.

Similarly, `moveDownDisabled` uses `this.model.tabs[index + 1]?.name === "Data"` hard-coded.

#### 3. `<script>` tag removal in `logical-model-attribute-editor.component.html`

The PR removes a stray `<script src="auto-drill-dialog/data-auto-drill-dialog.component.ts"></script>` tag. This is a correct cleanup — an HTML `<script>` tag in an Angular template has no effect and should not be there.

#### 4. `vs-calendar.component.ts` — defensive guard on `getSelectionString`

```ts
else if(!!this.calendar1 && typeof this.calendar1.getSelectionString === "function") {
   selectionString = this.calendar1.getSelectionString();
}
```

This adds a runtime method-existence check before calling `getSelectionString`. If `this.calendar1` is typed correctly (i.e., the type declares `getSelectionString`), this guard is unnecessary and suggests a type mismatch. It would be cleaner to fix the type or add a proper interface guard rather than a raw `typeof` check. That said, this is defensive code that prevents a crash, so it is not a regression.

#### 5. `vitest-setup.ts` — `compileComponents` patch timing

The new `TestBed.compileComponents` patch re-scrubs all configured module defs after compilation. This is a reasonable fix for Angular 21's deferred resource resolution, but the `_configuredModuleDefs` array accumulates entries across test runs. Since `vitest` can run tests in the same process, verify that the array is cleared correctly between test files (the `shift()` in the patch does drain it, so this appears fine).

---

### Positive Observations

- The portal navigation refactor (`app.component.html`/`app.component.ts`) significantly simplifies the old `reportTabFirst`/`dataTabFirst` dual-branch logic into a clean `@for` loop over `leftNavTabs`. This is a clear improvement.
- The EM Portal Integration view correctly separates Left Nav (reorderable) and Right Nav (Data — visibility toggle only), including the `moveDownDisabled` tightening to prevent a left-nav tab from sliding past Data.
- `DataDetailsPaneService` added to `data-tab.component.ts` providers is a correct fix for the data tab.
- `label-input-field.spec.ts` correctly updates the CSS selector from `.custom-select` to `.custom-select-trigger` to match the renamed theming class.
- The `dashboard-landing.component.html` landing page redesign with `landing-shell`/`landing-card` is cleaner and removes the dead Repository card.

jshobe-inetsoft (changes requested):
## `vitest-setup.ts` — Detailed Analysis

This review focuses specifically on the test infrastructure changes in `vitest-setup.ts` (commit `177ecfe`).

### Background

All three components patched for circular dependencies use external `styleUrls`:

- `VSCalendar` → `styleUrls: ["vs-calendar.component.scss"]`
- `VSObjectContainer` → `styleUrls: ["vs-object-container.component.scss"]`
- `VSViewsheet` → `styleUrls: ["vs-viewsheet.component.scss"]`

In JIT test mode, accessing `ɵcmp` on a component with unresolved `styleUrls` before `compileComponents()` has run throws: `"Did you run and wait for 'resolveComponentResources()'?"`.

---

### What Each Change Does

**`_readAngularDef` / `_isPendingComponentResourceError`** — wraps `cls.ɵcmp` access in a try/catch for the Angular resource-resolution error. This part is legitimate: it prevents a crash when the dep-scrub walker encounters VSCalendar/VSObjectContainer/VSViewsheet during `configureTestingModule`.

**`_configuredModuleDefs` + `TestBed.compileComponents` patch** — this is the part that needs scrutiny.

---

### Three Specific Concerns

**1. The module-load-time `_patchKnownCircularDeps()` call (line 278) is now dead**

Because all three targeted components have `styleUrls`, their `.ɵcmp` will throw the resource-resolution error during module load — before any test has called `compileComponents`. `_readAngularDef` returns `null` for each of them, so every `_patchCircularDeps` / `_patchVSCalendarDeps` call at load time silently no-ops. The only effective patching now happens inside `compileComponents`. The module-load call should either be removed or documented as intentionally retained for future cases where the components don't have `styleUrls`.

**2. `_configuredModuleDefs` is global state that is never reset between tests**

```typescript
const _configuredModuleDefs: any[] = [];  // lives forever at module scope

// configureTestingModule:
_configuredModuleDefs.push(moduleDef);    // accumulates

// compileComponents:
while(_configuredModuleDefs.length > 0) {
    _scrubTestingModuleDef(_configuredModuleDefs.shift());  // drains
}
```

`TestBed.resetTestingModule()` resets the Angular test bed between tests but does **not** clear `_configuredModuleDefs`. If test A calls `configureTestingModule` without calling `compileComponents` (a common pattern for components with only inline templates), its module def stays in the array. Test B's `compileComponents` then re-scrubs test A's already-configured module def — mutating dependency arrays in a module that may already be assembled or torn down. This is cross-test state contamination.

**3. `_patchKnownCircularDeps()` is called on every `compileComponents` invocation globally**

Every test in the suite that calls `compileComponents` — not just tests using VSCalendar or VSViewsheet — now re-invokes all five circular dep patches. The `undefined/null` guards make repeat calls no-ops functionally, but this is unnecessary overhead applied to the entire test suite, not scoped to the tests that actually need it.

---

### Questions for the Author

1. **What were the 3 specific test failures?** Please include the test names and the error output. Understanding whether `_readAngularDef` alone was sufficient (preventing the crash on first scrub) vs. requiring the `compileComponents` re-scrub would clarify whether the global accumulator is genuinely needed.

2. **Would the following minimal fix have been sufficient?** Replace the direct `cls.ɵcmp` / `cls.ɵmod` accesses in `_scrubDeps` and `_patchCircularDeps` with `_readAngularDef` (which you've done), without adding the `compileComponents` patch and the `_configuredModuleDefs` accumulator. If the 3 failures were caused purely by the crash on `.ɵcmp` access, this narrower fix would be complete.

3. **If the re-scrub in `compileComponents` is genuinely required**, consider scoping the module def storage per-test rather than using a global accumulator — for example, by using a `WeakRef` or by clearing the array inside `TestBed.resetTestingModule` as well.

swaker854:
Addressed the PR feedback and follow-ups in this round:

- Replaced hard-coded `"Data"` tab checks in `portal-integration-view.component.ts` with `PortalTabs.DATA`, centralized through `isDataTab()`.
- Fixed Left Nav ordering in EM so `Schedule` can move to the first visible Left Nav position even when the hidden/right-nav `Data` tab is first in the backing model.
- Updated `vs-calendar.component.ts` to use a small `CalendarSelectionSource` interface guard before calling `getSelectionString()`, preserving the defensive behavior without an inline raw `typeof` check.
- Tightened the Vitest/TestBed circular dependency workaround:
  - removed the dead module-load-time `_patchKnownCircularDeps()` call,
  - only queues module defs that actually encounter unresolved component resources,
  - clears queued defs in both `TestBed.resetTestingModule()` and `afterEach()`,
  - only runs known circular dependency patches after `compileComponents()` when queued work exists.

jshobe-inetsoft (changes requested):
## Code Review

This PR restores Angular theming patterns lost in the Angular 21 migration and also adds new data-tab/tree-view functionality.

### Summary of Changes

The PR does two things:
1. **Template restorations** — Replaces `form-group` → `shell-form-group`, restores `custom-select-trigger` dropdown patterns with chevron icons, refactors portal nav tab rendering, and updates the landing page/welcome page layouts.
2. **New functionality** — Adds search, section-state management, and `DataDetailsPaneService` to `DataSourcesTreeViewComponent` and `DataTabComponent`.

---

### Positive Changes

- **Critical fix**: Removes an invalid `<script src="auto-drill-dialog/data-auto-drill-dialog.component.ts">` tag from `logical-model-attribute-editor.component.html`. This was wrong and a potential security concern.
- **Portal nav refactor**: Replacing the `reportTabFirst`/`dataTabFirst` booleans with a `leftNavTabs` getter that iterates tabs in order is cleaner and easier to reason about. The new "Left Nav / Right Nav" grouping in EM also correctly prevents the Data tab from being reordered past the navigation boundary.
- **VS Calendar type guard**: The `isCalendarSelectionSource` type predicate is the right fix for checking that `calendar1`/`calendar2` objects actually implement `getSelectionString` before calling it.
- **vitest-setup.ts improvements**: The deferred `_patchKnownCircularDeps()` approach (called after `compileComponents()`) correctly handles Angular 21's lazy component resource resolution.
- **Dropdown UX**: Chevron up/down icons added to `dynamic-combo-box`, `border-style-pane`, `grid-line-dropdown`, `style-dropdown`, `multi-select`, and `b-categorical-color-pane` correctly reflect open state.
- **Port number form field**: The old template referenced `portCtrl.invalid` without a `#portCtrl` template reference on the number-stepper. The new inline condition `@if (!database.network.portNumber && !isCustom(database.type))` correctly fixes this.

---

### Issues

#### 1. `*ngIf` instead of `@if` — `stat-panel.component.html`

```html
<!-- Added in this PR -->
<b-categorical-color-pane *ngIf="model.bandFill" ...>
```

All other structural directives in this PR (and codebase) use Angular 17+ `@if`. This should be:
```html
@if (model.bandFill) {
  <b-categorical-color-pane [(colorModel)]="model.bandFill" ...>
}
```

#### 2. Redundant nested `@if` — `datasources-database.component.html`

```html
@if (uploadEnabled) {
  @if (uploadEnabled) {  <!-- inner check is redundant -->
    <label class="editor-help-label" [innerHTML]="'_#(data.databases.driverAdminUpload)'">
    </label>
  }
  ...
}
```

The inner `@if (uploadEnabled)` is always `true` within the outer block. The inner block can be removed.

---

### Failing Checks

The `claude-review` CI check is failing, which is blocking this PR from merging (`mergeStateStatus: BLOCKED`).

---

### Conclusion

The core changes are correct and the template restorations are valid. The two code issues above should be addressed before merging.

swaker854:
Address PR feedback: Angular template cleanup

- Replace *ngIf with @if block syntax on b-categorical-color-pane in stat-panel.component.html
- Remove redundant nested @if (uploadEnabled) in datasources-database.component.html

swaker854:
Fix vitest-setup.ts

jshobe-inetsoft (approved):
## Code Review — PR #3880

**Restores theming after Angular 21 migration and fixes portal toolbar / data tab**

The build passes, Semgrep/security scans pass, CLA is signed, and the PR is mergeable. The failing `claude-review` check is a CI infrastructure issue (OIDC token authentication failure — missing `id-token: write` permission in the workflow), not related to the code in this PR.

---

### Summary

This PR does two things:

1. **Portal toolbar / data-tab fix** — separates the Data tab into its own "Right Nav" section in the EM Portal Integration settings, removes the brittle `reportTabFirst`/`setTabOrder()` boolean-flag approach, and replaces it with a clean `leftNavTabs` computed property.
2. **Template pattern restoration** — recovers ~20 Angular template patterns (`shell-form-group`, `shell-form-row`, `custom-select`, `number-stepper`, `landing-card`, `landing-shell`, `enterClick`) that were lost when commit `563ab05dd` resolved a merge conflict incorrectly.

---

### Positive findings

- **`leftNavTabs` refactor** (`app.component.ts`): Replacing `reportTabFirst` + `setTabOrder()` + four duplicate `@if`/`@else` blocks with a single `@for (tab of leftNavTabs)` + `@else if` dispatch is strictly better. The ordering is now driven by actual server-side tab order rather than a derived boolean, which eliminates an entire category of ordering bugs.

- **`isCalendarSelectionSource` type guard** (`vs-calendar.component.ts`): Adding a proper TypeScript interface + type-narrowing guard around `calendar1`/`calendar2` before calling `getSelectionString()` is the right fix. The previous `!!this.calendar2` check relied on duck typing at runtime.

- **`vitest-setup.ts` — JIT-mode guard**: The try/catch around `ɵcmp`/`ɵmod` access is a correct defensive fix for the "Component is not resolved" error in TL test projects. Silently returning is safe because the dep-scrubbing is best-effort.

- **Chevron icon consistency**: `border-style-pane`, `dynamic-combo-box`, `grid-line-dropdown`, `style-dropdown`, `multi-select`, and `hyperlink-dialog` all now use the standardized `custom-select-trigger__icon` pattern with proper open/closed chevron state — a consistent, themeable pattern.

- **Portal Integration — Left/Right Nav split**: The `leftNavTabs` + `dataTab` getters cleanly separate the two navigation areas. The `moveDownDisabled` guard (`isDataTab(model.tabs[index + 1])`) correctly prevents left-nav tabs from being re-ordered past the data tab, while the `moveUpDisabled` adjustment allows a tab to move up through the data tab's model-array slot (the data tab itself has no move buttons in the UI).

- **`stat-panel.component.html` null guard**: The `@if (model.bandFill)` guard at the call site is functionally equivalent to the previous inner guard in `b-categorical-color-pane`, and actually improves performance since no component instance is created when `bandFill` is null.

- **Welcome page redesign**: The new `landing-shell`/`landing-card` structure with proper `role="button"`, `tabindex`, `enterClick`, and `aria-label` attributes is a clear accessibility improvement over the old card layout.

---

### Observations (non-blocking)

- **`b-categorical-color-pane` — contract change**: The old component wrapped its entire template in `@if (colorModel)`. The new version removes that guard and expects callers to pass a non-null `colorModel`. There is currently only one usage site (`stat-panel.component.html`) and it has been updated with an `@if (model.bandFill)` guard. This is fine now, but future callers must be aware of the tightened contract — `colorModel.colors[_i]` inside the dropdown menu will throw if null.

- **`isPortalDataSourcesTree` → `isRepositoryTree`** (`data-sources-tree-view.component.html`): This input attribute was changed on the `<tree>` component. Visually, this should be verified to ensure the tree renders the expected styles for this context.

- **`initSelectedNodesExpanded` changed from `true` to `false`**: The data sources tree no longer auto-expands selected nodes on initialization. Assuming this is intentional given the epic scope, but worth manual verification.

---

### Test coverage

- `label-input-field.spec.ts` correctly updated the CSS selector from `.custom-select` to `.custom-select-trigger` to match the new DOM structure.
- The vitest setup changes are defensive and safe.

---

**Verdict: Approved.** No blocking issues. Build passes, security scans clean, PR is mergeable.

PR #3889, opened 2026-06-04
Bug #75054 - Emphasize stack total in solo card tooltip

FigurePaper:
Problem

On a solo card tooltip (when the combined tooltip is not enabled), the
stack-total row (e.g. Total: 1.8M) renders as a plain, grey tier instead of an
emphasized footer with the tt-stack-total class.

Root cause

For a single-measure stacked bar, PlotArea calls captureCombinedCardHeader
(setHeader); for multi-measure it calls setGroupedTiers(true). Both route
renderCard to renderGroupedSoloCard, which rendered every pair, including the
total, as a plain tier with no emphasis.

Fix

Apply the find/skip/emphasize logic to renderGroupedSoloCard and
renderUniformCard, and extract the shared footer emit into appendStackTotal.
The total is matched by name (reusing findStackTotalIndex) rather than by
position, since PlotArea adds it before aesthetic rows and it is therefore not
always the last pair. The skip happens before the tier assignment so the total
does not take the tier-1 headline slot or shift tier-2 grouping.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/region/ChartToolTip.java
+++ b/core/src/main/java/inetsoft/report/composition/region/ChartToolTip.java
@@ -174,7 +174,6 @@ public class ChartToolTip implements DataSerializable {
    private String renderGroupedSoloCard(IndexedSet<String> palette) {
       StringBuilder buffer = new StringBuilder();
       int pairsRendered = 0;
-      // Emphasize the stack total as a footer total, matching the combined-card path.
       int stackTotalIdx = findStackTotalIndex(palette);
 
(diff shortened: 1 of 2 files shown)


FigurePaper:
Addressed the line-177 comment (removed — the skip rationale below it already explains the why) and
added a trailing row to the uniform test so it covers the total-not-last case.

Leaving the two other notes as-is: the appendStackTotal bounds check stays for symmetry with the
combined-card path that uses the same guard, and the "only a stack total" case is contrived and not
introduced by this change, so it's out of scope here.

jshobe-inetsoft (approved):

PR #3896, opened 2026-06-04
Bug #75302 - Candle card tooltip subtitle uses X-axis date dim, not Y-axis dim

FigurePaper:
## Summary

- On a candle/stock card tooltip with a Y-axis dimension binding (e.g. `bullOrbear` splitting Bullish/Bearish), the period dim `Week(date)` was buried at tier-3 while the Y-axis grouping dim took the tier-1 subtitle slot. The date is the per-candle identity and belongs directly under the Close headline.
- Follow-up to Bug #75026: that change introduced the candle card header but lifted `dims[0]` positionally. `element.getDims()` returns both the X period dim and any Y grouping dim, so `dims[0]` could resolve to the wrong dim.
- The tooltip now always promotes the X-axis date dim as the subtitle; Y-axis and other grouping dims stay at tier-3 as supplementary context.

## Root cause

`PlotArea.getToolTip0` passed `dims[0]` to `applyCandleCardHeader`. `dims` comes from `element.getDims()`, which contains every element dimension. With a Y-axis dimension bound, `dims[0]` resolved to that grouping dim (`bullOrbear`) instead of the X period dim, so `bullOrbear` was lifted to the tier-1 subtitle and `Week(date)` dropped to tier-3.

`applyCandleCardHeader` itself was correct; the defect was purely which dim it was handed.

## What changed

`PlotArea`:

- New `candleHeaderDim(ChartRef[] xfields, String[] dims)` helper: returns the innermost X-axis dimension name (skips measures via `GraphUtil.isMeasure`), falling back to `dims[0]` only when no X dimension is present. Nested categorical X still resolves to the per-candle period (the innermost X dim).
- The candle branch now calls `applyCandleCardHeader(tooltip, candleHeaderDim(chartInfo.getRTXFields(), dims), tipMeasures)`. Uses the runtime X fields (`getRTXFields`) so the field names match the runtime `dims` from `element.getDims()` — consistent with the sibling `reorderCandleMeasuresCloseFirst`, which matches runtime measures against `getRT*Field()`. This avoids a silent fallback to `dims[0]` when the runtime date level diverges from design (date comparison, dynamic levels).

No change to `applyCandleCardHeader`, `reorderCandleMeasuresCloseFirst`, tier counting, or any non-candle path.

Resulting tooltip on the binding from the bug report (`Week(date)` X-axis dim, OHLC measures, `bullOrbear` Y-axis dim):

```
Last(close, date): 179.4        tier-1 (headline)
Week(date): 2025-02-23          tier-1 subtitle
First(open, date): 181.3        tier-2
Max(high): 182.8                tier-2
Min(low): 178.3                 tier-2
bullOrbear: Bearish             tier-3
```

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/region/PlotArea.java
+++ b/core/src/main/java/inetsoft/report/composition/region/PlotArea.java
@@ -1660,9 +1660,11 @@ public class PlotArea extends GridContainerArea implements GraphComponentArea, R
             }
          }
 
-         // Lift only dims[0]; nested X-dims stay as regular rows (matches captureCombinedCardHeader).
+         // Lift the X-axis period dim (e.g. the date), not dims[0]; a Y-axis
+         // grouping dim like bullOrbear must not steal the subtitle slot.
          if(candleCardLayout && dims.length > 0) {
-            applyCandleCardHeader(tooltip, dims[0], tipMeasures);
+            applyCandleCardHeader(tooltip, candleHeaderDim(chartInfo.getRTXFields(), dims),
+                                  tipMeasures);
          }
          else if(cardMeasureFirst && dims.length > 0) {
             if(groupMeasuresAtTier2(measures, full2NoCalcNames.keySet(), measurePairs)) {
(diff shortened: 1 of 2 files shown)


FigurePaper:
Leaving candleHeaderDim as a plain package-private static. @VisibleForTesting isn't used anywhere
in core (0 occurrences), and the sibling test-only helpers in this same file — appendMeasurePair
and groupMeasuresAtTier2 — are already package-private static with no annotation. Matching that
local convention; adding an annotation the codebase doesn't otherwise use would be inconsistent.

The other points are handled in the latest commit: candleHeaderDim now scans X fields
inner-to-outer and returns the first one present in dims (fixes the dims[0] fallback edge case and
the null-overwrite), the comment describes the real fallback condition, and
candleHeaderFallsBackToOuterXDimWhenInnermostAbsent covers the missing edge case.

jshobe-inetsoft (approved):
All issues from the Claude automated review have been resolved in the follow-up commit. The `candleHeaderDim` fix is correct, the new test coverage is thorough, and all checks pass. Approving for squash merge.

PR #3908, opened 2026-06-05
Bug #75189, register cloned table in worksheet before applyDiscreteAggregates

jkidacki-inetsoft:
RelationalJoinTableAssembly stores child table names and resolves them via ws.getAssembly() at execution time. The clone produced by createTableAssembly was never registered in the local WorksheetWrapper, so applyDiscreteAggregates would find the original unmodified bound table instead, ignoring the chart's aggregate and column configuration.

Move ws.removeAssembly/addAssembly to before applyDiscreteAggregates so the correct object is resolved when discrete measure join tables are built and later executed.

Revert VSDataSet.indexOfHeader0 fallback that stripped aggregate formula wrappers (e.g. "Sum(X)" → "X") when looking up column indices. That workaround was compensating for the wrong table being executed; with the root cause fixed the column headers match directly and the fallback is no longer needed.


jshobe-inetsoft (approved):

PR #3910, opened 2026-06-05
Bug #75313, fix MV creation deadlock caused by Ignite public pool starvation

jkidacki-inetsoft:
getIgniteCompute() fell through to the default public pool when level % IGNITE_EXECUTE_POOL_COUNT == 0. MVCallable tasks submitted via Cluster.submit() would block on BlobStorage.putBlob() waiting for SingletonCallableTask results, whose completion is delivered by ServiceTaskResultListener — also on the public pool. With all public threads occupied by blocked MVCallable tasks, the listeners could never run, causing 60s timeouts and failed MV creation.

Fix: always route through a named custom executor in getIgniteCompute(), keeping the public pool free for message listener callbacks.


jshobe-inetsoft (approved):

PR #3911, opened 2026-06-05
Bug #75346, fix viewsheet hyperlink broken after Angular upgrade

jkidacki-inetsoft:
PageTabService was listed in both app.config.ts (root injector) and ViewerAppComponent.providers (component injector), causing two separate instances. ViewerViewComponent initialized tabs on the root instance, but ShowHyperlinkService (provided inside ViewerAppComponent) received the component-level instance which never had currentTab set, so clicking a _self viewsheet hyperlink crashed with "Cannot set properties of null (setting 'runtimeId')".

Removing PageTabService from ViewerAppComponent.providers ensures ShowHyperlinkService uses the same root instance that ViewerViewComponent initializes.


jshobe-inetsoft (approved):

PR #3912, opened 2026-06-05
Bug #75349: Fix collapsible monitoring tables disappearing when rolled up

larryliang-inetsoft:
## Symptom

The monitoring tables (Cache, Queries, Viewsheets, Users) have an accordion toggle at the top to roll them up. Clicking the toggle made the **entire element disappear** — header, toggle, and table — with no way to bring it back.

## Root Cause

Regression from the Angular 20→21 upgrade (#3809); specifically an instance of the defect class fixed in Bug #75272 (#3840) that was missed by that pass.

In `table-view.component.html`, the `mat-expansion-panel-header` of the collapsible variant was wrapped in an `@if` block:

```html
@if (tableInfo.title) {
  <mat-expansion-panel-header ...>
    <mat-panel-title>{{tableInfo.title}}</mat-panel-title>
  </mat-expansion-panel-header>
}
```

`mat-expansion-panel` projects its header via a **named slot** — `<ng-content select="mat-expansion-panel-header">` — which sits *outside* the collapsible region, while the default `<ng-content>` slot sits *inside* the animated `.mat-expansion-panel-content` region.

Content projection matches against the **top-level nodes** of the component's content. In Angular 21 an `@if` block is itself the top-level node — an opaque container that only matches the wildcard slot. The header therefore failed to match `select="mat-expansion-panel-header"` and was projected into the **default slot inside the collapsible body**.

The header's click handler still toggled the panel (it gets the panel reference via DI, independent of where it renders), so clicking the toggle collapsed the body — which now contained the header itself. Material's CSS rule

```css
.mat-expansion-panel-content[style*="visibility: hidden"] * { visibility: hidden !important; }
```

then hid everything, including the only control that could re-expand the panel.

Since `em-table-view` is shared, every collapsible monitoring table was affected.

## Why not keep `@if`?

`@if` is the preferred modern control-flow syntax, but named-slot content projection is the one documented case where it cannot be used (angular/angular#52414). Both `@if` variants were verified empirically against the new regression test:

| Variant | Result |
|---|---|
| `@if (...) { <mat-expansion-panel-header> }` | ❌ header lands inside collapsible body (the bug) |
| `@if (...) { <mat-expansion-panel-header ngProjectAs="mat-expansion-panel-header"> }` | ❌ still fails — `ngProjectAs` is on a node *inside* the block, not on the top-level node, so it never participates in slot matching |
| `<mat-expansion-panel-header *ngIf="tableInfo.title">` | ✅ header projected into the correct slot |

`*ngIf` works because the compiler preserves the host element's selector/attributes on the generated template for projection matching. This is also the exact fix pattern established codebase-wide by Bug #75272 (#3840) for `mat-label`/`mat-error`/`mat-hint`/`matSuffix` etc., so this change keeps the codebase consistent. The only `@if`-preserving alternative would be duplicating the entire `<mat-expansion-panel>` into with-title/without-title branches.

## Fix

Replace the `@if` wrapper with `*ngIf` on the `mat-expansion-panel-header` element itself (`NgIf` was already in the component's imports).

## Testing

- Added a regression test (host component with `[collapsible]="true"` and a title) asserting the rendered `mat-expansion-panel-header` is **not** nested inside `.mat-expansion-panel-content`. Fails before the fix (`content.contains(header) === true`), passes after.
- Ran all consumer specs (cache/queries/viewsheets/users/cluster/log monitoring views, collapsible-container, regular/expandable-row/table-view tables): **11/11 pass**, no regressions.

### DIFF
--- a/web/projects/em/src/app/common/util/table/table-view.component.html
+++ b/web/projects/em/src/app/common/util/table/table-view.component.html
@@ -62,11 +62,9 @@
 @if (matTableDataSource && collapsible && tableInfo) {
   <mat-expansion-panel class="table-container"
     [expanded]="expanded">
-    @if (tableInfo.title) {
-      <mat-expansion-panel-header class="light-gray-bg table-cell-padding-sm m-0">
-        <mat-panel-title>{{tableInfo.title}}</mat-panel-title>
-      </mat-expansion-panel-header>
-    }
+    <mat-expansion-panel-header *ngIf="tableInfo.title" class="light-gray-bg table-cell-padding-sm m-0">
+      <mat-panel-title>{{tableInfo.title}}</mat-panel-title>
+    </mat-expansion-panel-header>
     <mat-card appearance="outlined" class="mat-elevation-z0">
       <mat-card-content>
         @if (expandableRow) {
--- a/web/projects/em/src/app/common/util/table/table-view.component.spec.ts
+++ b/web/projects/em/src/app/common/util/table/table-view.component.spec.ts
@@ -17,7 +17,7 @@
  */
 
 import { CommonModule } from "@angular/common";
-import { NO_ERRORS_SCHEMA } from "@angular/core";
+import { Component, NO_ERRORS_SCHEMA } from "@angular/core";
 import { waitForAsync, ComponentFixture, TestBed } from "@angular/core/testing";
 import { MatCheckboxModule } from "@angular/material/checkbox";
 import { MatExpansionModule } from "@angular/material/expansion";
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft:
This also fixes Bug #75330 ("when shrink icon in monitor, pane be hidden" — the 'Open Dashboards' pane on em/monitoring/viewsheets). It's the same root cause: the collapsible `em-table-view`'s `mat-expansion-panel-header` was wrapped in `@if`, misprojecting the header (with the shrink chevron) into the collapsible body. The `*ngIf` change here resolves both #75349 and #75330.

jshobe-inetsoft (approved):

PR #3913, opened 2026-06-06
Bug #75322, do not apply maxrows clause to where subqueries on Derby

larryliang-inetsoft:
## Summary

Showing live data for a worksheet table with a subquery condition (e.g. `PRODUCT_ID` *not one of* → `NOT IN (subquery)`) against a Derby data source failed with:

```
java.sql.SQLSyntaxErrorException: 'SELECT *' only allowed in EXISTS and NOT EXISTS subqueries.
```

## Root Cause

In live mode, `PreAssetQuery.mergeOthers()` applies a preview row limit to a merged WHERE subquery by setting `HINT_OUTPUT_MAXROWS`, gated by `helper.supportsOperation(SQLHelper.MAXROWS, SQLHelper.WHERE_SUBQUERY)`. The `info` flag exists so dialects can veto maxrows inside WHERE subqueries, but `DerbyHelper` never overrode the two-arg `supportsOperation`, and the base implementation ignores `info`.

`DerbyHelper.generateMaxRowsClause()` implements the row limit by wrapping the query in `select * from (...) fetch first n rows only`. When that wrapper lands inside an `IN`/`NOT IN` subquery, Derby rejects it with error 42X38, since Derby only allows `SELECT *` in EXISTS/NOT EXISTS subqueries.

## Fix

Override `supportsOperation(String op, String info)` in `DerbyHelper` to return `false` for `(MAXROWS, WHERE_SUBQUERY)`, so the maxrows hint is never applied to WHERE subqueries on Derby. Top-level maxrows behavior and all other dialects are unchanged.

## Test Plan

- Added `DerbyHelperTest` covering the new override and the preserved top-level maxrows behavior (`./mvnw test -pl core -Dtest=DerbyHelperTest` — passes).
- Manual verification with the case from the Redmine issue: import the `subquery condition` worksheet, show live data of `PRODUCTS1` — data loads without the SQL error.
- Regression check: tables/previews without subquery conditions on Derby still apply the preview row limit (top-level `fetch first n rows only` unchanged).

Fixes Redmine #75322.

### DIFF
--- a/core/src/main/java/inetsoft/uql/jdbc/DerbyHelper.java
+++ b/core/src/main/java/inetsoft/uql/jdbc/DerbyHelper.java
@@ -135,6 +135,26 @@ public class DerbyHelper extends SQLHelper {
       return super.supportsOperation(op);
    }
 
+   /**
+    * Check if supports an operation.
+    * @param op the specified operation.
+    * @param info the advanced information.
+    * @return <tt>true</tt> if supports, <tt>false</tt> otherwise.
+    */
+   @Override
+   public boolean supportsOperation(String op, String info) {
+      // the maxrows clause wraps the query in 'select * from (...) fetch first
+      // n rows only' (see generateMaxRowsClause), and derby only allows
+      // 'select *' in EXISTS and NOT EXISTS subqueries (error 42X38), so the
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #3915, opened 2026-06-06
Bug #75343, adopt the effective theme id from the save response

larryliang-inetsoft:
## Summary

Renaming a custom theme in EM (Settings → Presentation → Themes) and applying made the theme impossible to delete or download: both requests failed with `404 (Not Found)` (e.g. `DELETE /api/em/settings/presentation/themes/b`).

## Root Cause

Theme ids are initialized equal to the theme name at creation (`AddThemeDialogComponent.createId()`). A recent enterprise-side change (Epic 70095) made `ThemeService.updateTheme()` rename the theme's **id** to the new name when the id equaled the old name — but the `PUT /api/em/settings/presentation/themes/{id}` endpoint returned `void`, so the client was never told the id changed. `saveTheme()` kept the stale original id in `themes[]` and `selectedTheme`, and every subsequent delete/download/save request targeted the stale id → 404.

## Fix

`saveTheme()` now types the PUT response as `CustomThemeModel` and adopts the **effective id** returned by the server (after locating the list entry by the old id, before cloning into the list and re-selecting). When the response body is empty (older servers), it falls back to the current id, preserving existing behavior.

Companion enterprise PR (returns the updated model from the endpoint): https://github.com/inetsoft-technology/stylebi-enterprise/pull/PENDING

## Test Plan

- New regression spec `presentation-themes-view.component.spec.ts`:
  - save response with renamed id → local list/selection adopt it and DELETE targets the new id (verified to **fail** against the unfixed component)
  - empty save response → id unchanged (older-server fallback)
- All theme-related em specs pass (6/6); `ng build em` and ESLint clean
- Manual: create theme → rename → Apply → Delete and Download both succeed

Fixes Redmine #75343.

### DIFF
--- a/web/projects/em/src/app/settings/presentation/presentation-themes-view/presentation-themes-view.component.spec.ts
+++ b/web/projects/em/src/app/settings/presentation/presentation-themes-view/presentation-themes-view.component.spec.ts
@@ -1,0 +1,114 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #3917, opened 2026-06-06
Bug #75323, only list output columns of a subquery table in the subquery condition dialog

larryliang-inetsoft:
## Summary

In the worksheet composer's assembly condition dialog, the "In column" dropdown of the subquery editor listed **all** base columns of the selected subquery table. For a table in aggregate mode (e.g. a single `Average(CUSTOMER_ID)` aggregate with no groups), the table's actual output is a single column, and only that column should be offered. Selecting a non-output column produced a broken multi-column subquery at runtime.

## Root Cause

`AssemblyConditionDialogService.createAssemblyConditionDialogModel()` built each `SubqueryTableModel`'s column list from the table's **private** column selection (`getColumnSelection()`), which contains all base columns regardless of the table's aggregate state.

A subquery always executes with the table's aggregate info applied (`PreAssetQuery.getAggregateInfo()` returns it whenever `isSubQuery()` is true), so the columns a subquery can supply are exactly its **public** column selection — the visible columns filtered to groups/aggregates when aggregate info is non-empty (the same contract mirror tables consume).

## Fix

Use `getColumnSelection(!currentTable)` when building the subquery table models: the public selection for subquery candidate tables, while the current table's entry keeps the private selection (it is only used to offer `mainAttribute` choices for correlated subqueries, which operate on the current table's detail columns).

The save path is unaffected: the chosen column is resolved by name against the subquery table's private selection, and public columns share the same names.

## Test Plan

- Import the case from the Redmine issue; open worksheet `subquery condition column` in composer; add a subquery condition on `CUSTOMERS2` with subquery `CUSTOMERS1` — the "In column" dropdown now lists only the single aggregate output column (verified manually).
- Regression: subquery tables without aggregation still list all visible columns; correlated subquery `mainAttribute` choices (current table) are unchanged; saving/loading existing subquery conditions resolves columns as before.
- `./mvnw test -pl core -Dtest=ConditionValueModelTest` passes.

Fixes Redmine #75323.

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/AbstractTableAssembly.java
+++ b/core/src/main/java/inetsoft/uql/asset/AbstractTableAssembly.java
@@ -528,7 +528,10 @@ public abstract class AbstractTableAssembly extends AbstractWSAssembly implement
          AggregateInfo aggregateInfo =
             getRealAggregateInfo() == null ? ginfo : getRealAggregateInfo();
 
-         // generate public column selection if not a crosstab
+         // Generate the public (output) column selection if not a crosstab. For a crosstab
+         // the output columns are dynamic (pivoted at runtime), so the public selection is
+         // left empty here; TableAssemblyInfo.getPublicColumnSelection() then lazily falls
+         // back to the visible private columns, so callers still get a non-empty list.
          if(!isCrosstab()) {
             ColumnSelection ocolumns = new ColumnSelection();
             // Snapshot the selection to avoid IndexOutOfBoundsException if another thread
--- a/core/src/main/java/inetsoft/uql/asset/internal/TableAssemblyInfo.java
+++ b/core/src/main/java/inetsoft/uql/asset/internal/TableAssemblyInfo.java
@@ -93,9 +93,15 @@ public class TableAssemblyInfo extends WSAssemblyInfo {
     * @return the public column selection of the table assembly.
     */
    public ColumnSelection getPublicColumnSelection() {
-      // is a null selection?
+      // The "null" property is set to "false" once a public selection has been generated
+      // (see AbstractTableAssembly.setColumnSelection). nflag is true when that has not
+      // happened yet — e.g. for a crosstab, whose public selection is intentionally left
+      // empty because its output columns are pivoted at runtime.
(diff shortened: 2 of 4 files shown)


jshobe-inetsoft (changes requested):
Claude's automated review flagged the following issues that should be addressed before merging:

1. **Missing regression test**: No test covers the specific bug path (the case where `currentTable` is non-null and `applyDiscreteAggregates` should not run). A regression test would prevent this from silently regressing.
2. **Undiscussed cross-tab edge case**: The boolean inversion `!currentTable` has an undiscussed impact on cross-tab scenarios. Please verify or document the behavior in that path.
3. **Minor readability**: The `!currentTable` expression is slightly confusing — consider a named variable or comment to clarify the intent.

Please address these before this PR can be merged.

larryliang-inetsoft:
Addressed the three review items in 0b66b12:

**1. Readability (`!currentTable`)** — Extracted the decision into a named, documented helper `getSubqueryColumns(TableAssembly, currentTable)` instead of an inline boolean inversion.

**2. Missing regression test** — Added `AssemblyConditionDialogServiceTest` (`@Tag("core")`) covering:
- candidate → public / current table → private delegation (guards against reverting to the no-arg `getColumnSelection()`),
- an aggregated candidate listing only output columns (group + aggregate), **not** the raw detail column, and
- a crosstab candidate (see item 3).

**3. Crosstab edge case** — Verified, and it's **not** a regression, so I documented it rather than adding a guard (which would be dead code). For a crosstab table, `setColumnSelection(sel, false)` skips public-column generation (the `if(!isCrosstab())` guard in `AbstractTableAssembly`), leaving the public selection empty. But `TableAssemblyInfo.getPublicColumnSelection()` lazily falls back to a clone of the visible **private** columns when the public selection is empty — so `getColumnSelection(true)` returns the crosstab's visible columns, not an empty list. The new `crosstabCandidateFallsBackToVisibleColumns` test locks this in.

@jshobe-inetsoft could you re-review when you get a chance?

jshobe-inetsoft (approved):
Claude's automated review approves this PR — all prior blocking issues resolved, ready to merge.

PR #3919, opened 2026-06-06
Bug #75325, convert all conditions when toggling Show Server Time Zone

larryliang-inetsoft:
## Summary

In the portal schedule editor, toggling **Show Server Time Zone** on a multi-condition task only converted the condition currently being edited. Sibling conditions kept their raw local times, so opening one afterward showed an unconverted time mislabeled as the server zone (e.g. a condition stored at 02:30 Eastern displayed as `02:30 Coordinated Universal Time` instead of `06:30 Coordinated Universal Time`).

## Root Cause

`changeServerTimeZone()` → `convertToTimeZone()` iterates `this.timeConditions`, a Set containing only the five per-type stash conditions plus the currently-edited condition — never the other entries of `model.conditions`. Re-opening a sibling via `editCondition()` → `initTimeZone(true)` → `changeServerTimeZone(true)` is then a no-op because `serverTimeZone` is already `true` (`changed == false`), so siblings are never converted. Each `TimeCondition`'s `hour`/`minute` are interpreted in its own `timeZone` on the server, so an unconverted sibling renders incorrectly under the server zone.

## Fix

In `changeServerTimeZone()`, after the existing `convertToTimeZone(...)` call, also convert the task's other time conditions via a new `convertOtherConditions()` method. It iterates `model.conditions`, skips the currently-edited condition (same object reference) and non-time (chained) conditions, computes each sibling's own offset from its `timeZone`, skips siblings already in the server zone, and reuses the existing `convertTimeCondition()` with the correct direction. This restores a uniform invariant: flag on → all conditions display in server time; flag off → each in its own zone. No double conversion occurs because the per-type stash entries are deep clones (`Tool.clone`), disjoint from `model.conditions` entries.

## Test Plan

- New regression test in `task-condition-pane.spec.ts`: two conditions stored at 02:30 Eastern, server zone UTC; toggling on converts both the edited condition and the sibling to 06:30. Verified to **fail** before the fix (`expected 2 to be 6`) and pass after.
- `ng build portal` succeeds; ESLint clean on changed files.
- Manual: open a multi-condition task → set condition #2 to 02:30 Eastern → check "Show Server Time Zone" while editing condition #1 → open condition #2 → shows 06:30 UTC.

Fixes Redmine #75325.

### DIFF
--- a/web/projects/portal/src/app/portal/schedule/schedule-task-editor/conditions/task-condition-pane.component.ts
+++ b/web/projects/portal/src/app/portal/schedule/schedule-task-editor/conditions/task-condition-pane.component.ts
@@ -1224,8 +1224,44 @@ export class TaskConditionPane implements OnInit, OnChanges {
             this.convertToTimeZone(this.serverTimeZoneOffset, this.localTimeZoneOffset);
          }
 
+         // also convert the other conditions of a multi-condition task so their times are
+         // displayed in the new time zone when edited later (Bug #75325). each condition is
+         // converted using its own time zone offset, which may differ from the current one.
+         this.convertOtherConditions(this.serverTimeZone);
+
          this.updateTimeZone();
          LocalStorage.setItem(TZ_STORAGE_KEY, this.serverTimeZone + "");
+      }
+   }
+
+   /**
+    * Converts the times of the task's other time conditions (not currently being edited)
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #3928, opened 2026-06-06
Audit + fix @if-wrapped Angular Material projection regressions

larryliang-inetsoft:
## Summary

Systematic audit of the Angular 21 control-flow migration regression where named-slot Material projection targets (matSuffix, mat-label, mat-expansion-panel-header, etc.) were wrapped in `@if` blocks, breaking projection (the class of bug behind #75272, #75349, #75335). Adds a reusable detector, the audit findings, and fixes the one remaining unfixed instance.

## Root Cause

In Angular 21, when a control-flow block (`@if`/`@for`/`@switch`) sits **between** a Material projection target and its host component, the target is no longer projected into the host's named slot and renders as ordinary content. The #75272 sweep was incomplete.

## What this PR contains

- **`web/tools/if-projection-audit.py`** — prefilter (narrowed ~1000 templates → 78 candidates; portal/shared had none).
- **`web/tools/if-projection-classify.py`** — structural classifier applying the precise rule (a target is broken only if a control-flow block sits between it and an actual host above it). Handles `{{ }}` interpolations, `mat-card`'s single-slot semantics, stray targets, and method-call conditions. Validated against the three known cases.
- **`docs/superpowers/`** — design spec, plan, and findings record.
- **Fix:** `task-options-pane.component.html` — the `mat-spinner matSuffix` loading indicator on the "Owner" field was wrapped in `@if (isAdminNameLoading)`; converted to `*ngIf` on the element so it projects into the suffix slot.

## Findings (7 true positives total)

| File | Target | Status |
|---|---|---|
| `table-view.component.html:66` | `mat-expansion-panel-header` | already fixed — #75349 / PR #3912 |
| `dynamic-combo-box.component.html` (×5) | `matSuffix` | already fixed — #75335 / PR #3926 |
| `task-options-pane.component.html:100` | `matSuffix` (`mat-spinner`) | **fixed here** |

`portal` and `shared` are clean. All other prefilter candidates were confirmed harmless (mat-card single-slot children; stray labels/errors/hints outside a form-field; `@if` wrapping a whole form-field).

## Test Plan

- `python3 tools/if-projection-classify.py` reports `task-options-pane` clean after the fix; the only remaining flags are the two already covered by PRs #3912/#3926.
- `ng build em` succeeds; ESLint clean on the changed template.

Refs #75272, #75349 (#3912), #75335 (#3926).


larryliang-inetsoft:
All review feedback is now addressed — the two remaining cosmetic nits in `if-projection-classify.py` are done in 40e4d12:
- `CF_OPEN_RE`: documented the `[^\{<]*` false-negative trade-off (a bare `<` comparison in an `@if` condition won't match).
- `del stack[idx:]`: documented that it also discards unclosed `block` entries — defensive recovery for malformed input.

All checks are green and the build passes. @jshobe-inetsoft could you review when you have a chance?

jshobe-inetsoft (approved):

PR #3931, opened 2026-06-07
Bug #75337, hide LDAP under multi-tenancy and fix empty-host LDAP error

larryliang-inetsoft:
## Summary

Two related problems in the EM authentication-provider UI under Security + Multi-Tenancy:
1. The **LDAP** provider type still appeared in the "Add authentication provider" dropdown under multi-tenancy (it must be hidden).
2. Clicking get/edit roles for a new LDAP provider with an empty host threw `java.lang.IllegalArgumentException: Expected scheme name at index 0: ://:0`, surfacing as an "Unexpected error" 500.

## Root Cause

**Issue 1 (frontend):** `authentication-provider-detail-view.component.ts#initNewProvider()` bakes `model.ldapProviderEnabled = !this.isMultiTenant` for new providers, but the parent (`authentication-provider-detail-page`) resolves `isMultiTenant` asynchronously (`GET /api/em/navbar/isMultiTenant`) and renders the child unconditionally. At the child's `ngOnInit`, `isMultiTenant` is still the default `false`, so `ldapProviderEnabled` is baked `true` and never recomputed → the template's `@if (model?.ldapProviderEnabled)` keeps LDAP visible under MT. (Existing providers are fine — the backend sets the flag via `!enterprise || !SUtil.isMultiTenant()`.)

**Issue 2 (backend):** `getRolesForNewProvider → getRoles → getProviderFromModel → LdapAuthenticationProvider.checkParameters() → ADAuthenticationProvider.createContext()` builds `ldap://:0` from an empty host/port and throws an unhandled `IllegalArgumentException`. `checkParameters()` only catches `CommunicationException`/`AuthenticationException`/`NamingException`, and `SRSecurityException` (which `extends Exception`, not `MessageException`) hits `AdminExceptionHandler`'s generic path → logs "Unexpected error" + stack → 500.

## Fix

1. **`authentication-provider-detail-view.component.html`** — gate the LDAP option on the live input: `@if (model?.ldapProviderEnabled && !isMultiTenant)`. The `isMultiTenant` `@Input` updates via change detection once the parent's HTTP resolves, so LDAP correctly hides under MT for new providers. DB/Custom options and existing-provider behavior are unchanged.
2. **`LdapAuthenticationProvider.checkParameters()`** — fail fast with `SRSecurityException("LDAP host is required")` when the host is null/blank, before `createContext()`. Prevents the malformed URL for both AD and generic LDAP. (`testConnection` already catches `SRSecurityException`, so its behavior is unchanged/improved.)
3. **`AdminExceptionHandler`** — add `@ExceptionHandler(SRSecurityException.class)` that returns a clean `GenericError` (the message) and logs at DEBUG, so admin endpoints (get-roles/users/groups) render security-config errors cleanly instead of "Unexpected error" 500s.

## Test Plan

- New `GenericLdapAuthenticationProviderTest.testCheckParametersRequiresHost` asserts an empty host throws `SRSecurityException` mentioning "host" (runs in the slow LDAP suite in CI).
- `mvnw compile -pl core` and core test-compile pass; `ng build em` and ESLint pass; `authentication-provider-detail-view` spec passes.
- Manual: (1) Security + Multi-Tenancy → Add authentication provider → LDAP option absent. (2) Add LDAP provider, empty host, click get/edit roles → clean "LDAP host is required" message, no `://:0` error.

Fixes Redmine #75337.

### DIFF
--- a/core/src/main/java/inetsoft/sree/security/ldap/LdapAuthenticationProvider.java
+++ b/core/src/main/java/inetsoft/sree/security/ldap/LdapAuthenticationProvider.java
@@ -510,7 +510,9 @@ public abstract class LdapAuthenticationProvider
       // guard against an empty host so we fail with a clear message instead of
       // letting createContext() build a malformed URL (e.g. "://:0") that throws
       // an unhandled IllegalArgumentException (Bug #75337)
-      if(getHost() == null || getHost().trim().isEmpty()) {
+      String host = getHost();
+
+      if(host == null || host.trim().isEmpty()) {
          throw new SRSecurityException("LDAP host is required");
       }
 
--- a/core/src/main/java/inetsoft/sree/security/ldap/LdapAuthenticationProvider.java
+++ b/core/src/main/java/inetsoft/sree/security/ldap/LdapAuthenticationProvider.java
@@ -512,7 +512,7 @@ public abstract class LdapAuthenticationProvider
       // an unhandled IllegalArgumentException (Bug #75337)
       String host = getHost();
 
-      if(host == null || host.trim().isEmpty()) {
+      if(host == null || host.isBlank()) {
          throw new SRSecurityException("LDAP host is required");
       }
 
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):

PR #3941, opened 2026-06-08
Bug #75327, guard against undefined originalDataSource in email list dialog tree

jkidacki-inetsoft:
MatTreeFlattener.flattenNodes in Angular Material 21 no longer null-checks its input, so assigning undefined to dataSource.data now throws. This happened in two cases: when autocompleteEmails is false (edit identity view) so originalDataSource is never populated, and as a race condition when the user types before the getEmailList() HTTP response arrives.


jshobe-inetsoft (approved):

PR #3942, opened 2026-06-08
bug-75296: DB authentication provider read operations fail safe on invalid config

ngovus-inetsoft:
## Summary

- withProviderFromModel now catches exceptions thrown during provider creation and passes Optional.empty() to the downstream function instead of propagating a 500 error
- All read operation lambdas (getRoles, getUsers, getGroups, getOrganizations, etc.) already handle Optional.empty() by returning empty results, so they silently return nothing when the DB provider is misconfigured
- testConnection, addProvider, and updateProvider call getProviderFromModel directly and are unaffected -- they still surface errors to the user as expected

## Root Cause

When a Database Authentication Provider is not fully configured, the UI calls read endpoints (e.g. /api/em/security/get-roles) before setup is complete. getProviderFromModel unconditionally calls testConnection() for DATABASE providers, which throws SQLException (e.g. "driver class is not defined") that propagated as an unhandled 500.

## Test plan

- Configure a Database Authentication Provider with a missing driver class -- roles/users/groups panels should load empty without error
- Explicit Test Connection with invalid DB config should still return an error message to the user
- Fully configured DB provider should still load roles/users/groups correctly

Generated with Claude Code

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/security/AuthenticationProviderService.java
+++ b/core/src/main/java/inetsoft/web/admin/security/AuthenticationProviderService.java
@@ -343,7 +343,7 @@ public class AuthenticationProviderService extends BaseSubscribeChangeHandler {
       try {
          String msg = withProviderFromModel(model, provider ->
             provider.isPresent() ? null :
-            Catalog.getCatalog().getString("em.security.testlogin.note4"));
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):

PR #3946, opened 2026-06-08
Bug #75316, fix data cycle task not found in cloud runner mode

jshobe-inetsoft:
## Summary

Data cycle tasks failed when executed via the Docker cloud runner in a multi-tenant environment because `DataCycleManager.generateTasks()` returned early on cloud runner nodes, preventing task lookup before execution.

## Root Cause

`DataCycleManager.generateTasks()` contained a guard that returned immediately when the `ScheduleTaskRunner` system property was set (i.e., on cloud runner nodes). This guard was intended to prevent reactive task regeneration triggered by `MVChangedMessage` events after task completion — a safe optimization since cloud runner nodes are short-lived and their Ignite client may be shutting down at that point.

However, the guard was placed in `generateTasks()` itself rather than in the caller that handles the reactive event (`propertyChange()`). This meant that when `CloudRunnerContext.findScheduleTask()` called `DataCycleManager.getTasks(orgID)`, which in turn called `generateTasks()` to populate the task list for that org, the method returned immediately — always producing an empty list. As a side effect, `orgPregeneratedTaskLoadedStatus` was still set to `true` in the `finally` block, so subsequent calls for the same org also skipped generation, making the task permanently unfindable. The cloud runner then reported "Could not find schedule task" and the data cycle failed.

## Fix

Moved the `ScheduleTaskRunner` early-return guard from `generateTasks()` to `propertyChange()`, which is the only reactive path triggered by `MVChangedMessage` and `RepletRegistry.CHANGE_EVENT`. This preserves the original intent (skip regeneration on cloud runner nodes after task completion) while allowing `getTasks()` → `generateTasks()` to work correctly during task lookup.

## Test Plan

- [ ] Run a data cycle in the cluster-test Docker Compose environment (single server + Docker cloud runner)
- [ ] Confirm the MV task is found and executes successfully
- [ ] Confirm no regression in non-cloud-runner scheduling (standard scheduler task lookup still works)

Fixes Redmine #75316.

### DIFF
--- a/core/src/main/java/inetsoft/sree/internal/DataCycleManager.java
+++ b/core/src/main/java/inetsoft/sree/internal/DataCycleManager.java
@@ -142,6 +142,12 @@ public class DataCycleManager
       String orgId = Util.getOrgIdFromEventSource(evt.getSource());
 
       if(MVManager.MV_CHANGE_EVENT.equals(name) || RepletRegistry.CHANGE_EVENT.equals(name)) {
+         // cloud-runner nodes are short-lived task executors; their Ignite client
+         // may be shutting down when MVChangedMessage fires after task completion
+         if(System.getProperty("ScheduleTaskRunner") != null) {
+            return;
+         }
+
          generateTasks(null, orgId != null ? new Organization(new IdentityID(orgId, orgId)) : null,
                        true, false);
       }
@@ -265,13 +271,6 @@ public class DataCycleManager
          if(Scheduler.getSchedulerCount() != 1 &&
             "true".equals(System.getProperty("ScheduleServer")))
          {
-            return;
-         }
-
-         // cloud-runner nodes are short-lived task executors; they don't maintain
-         // pregenerated tasks and their Ignite client may be shutting down when this
-         // is triggered by an MVChangedMessage after task completion
-         if(System.getProperty("ScheduleTaskRunner") != null) {
             return;
          }
 


jkidacki-inetsoft (approved):

jshobe-inetsoft:
Thanks for the detailed review.

**Issue 1 — Secondary scheduler `finally`-block bug**

Noted. The secondary scheduler guard has the same structure (early return inside `try`, so `finally` still marks the org as loaded), but in practice secondary schedulers never need a non-empty pregenerated task list — they don't run data cycle tasks. So the behavioral impact is different from the cloud runner case. I'll track this as a separate follow-up rather than expanding the scope of this fix.

**Issue 2 — Other `generateTasks()` callers on cloud runner nodes**

This behavioral change is intentional. Cloud runners are task executors — they *need* to load pregenerated tasks during `init()` so they can find and run the data cycle tasks they've been assigned. The only path that must be suppressed is the reactive regeneration triggered by `MVChangedMessage` after task completion (when the Ignite client may be shutting down). Moving the guard to `propertyChange()` targets exactly that path. Having `save()` and the org-rename path call `generateTasks()` on cloud runners is the correct behavior that the overly broad original guard was accidentally suppressing.

**Issue 3 — Automated test coverage**

Agreed that a unit test would be valuable long-term. Reproducing this requires a multi-tenant Docker Compose environment with a cloud runner, which makes it difficult to cover with a simple unit test. I'll track this as a follow-up.

PR #3957, opened 2026-06-09
feat(wiz): enable map recommendations (tag geo columns before recommending)

larryliang-inetsoft:
## Problem
Requesting a map from the wiz plugin (`visualizationType: "map"`, `intentCategory: "geospatial"`) always fell back to a point chart: `selectionNote: "Requested chart type 'map' is not feasible for this data; used 'point' instead."` even for a `country` dimension.

`WizAutoBindingService` runs the recommender on the temp binding **without tagging geographic columns first**, so `MapChartFilter` finds no `GeoRef` and filters `map` out as infeasible. The interactive wizard (`VSWizardBindingService.setDefaultGeoColumns`) and the final-assembly builder (`addChartVSAssembly`) both set geo columns — only the plugin's recommendation path didn't.

## Fix
Call `VSUtil.setDefaultGeoColumns(tempChartInfo, rvs, source)` before `defaultRecommendationFactory.recommend(...)`, so country/state/city columns are geo-typed and a map becomes a feasible recommendation for the plugin path.

## Verification
Built into the local image; a `country` + measure binding with `intentCategory=geospatial`, `visualizationType=map` now produces a map. (Final reload confirmation pending.)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WizAutoBindingService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizAutoBindingService.java
@@ -24,6 +24,7 @@ import inetsoft.uql.asset.*;
 import inetsoft.uql.erm.DataRef;
 import inetsoft.uql.schema.XSchema;
 import inetsoft.uql.viewsheet.*;
+import inetsoft.uql.viewsheet.internal.VSUtil;
 import inetsoft.uql.viewsheet.graph.*;
 import inetsoft.graph.aesthetic.*;
 import inetsoft.uql.viewsheet.graph.aesthetic.ColorPalettes;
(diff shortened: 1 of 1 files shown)


larryliang-inetsoft:
Closing — wrong mechanism. setDefaultGeoColumns only auto-tags latitude/longitude columns, so this does NOT enable a named-region (country) map; it only helps lat/long point maps, which isn't the goal. A country/region map needs the dimension converted to geographic via VSChartHandler.changeGeographic(SET_GEOGRAPHIC) + geographic-level (Country) mapping detection before recommendation (as the interactive wizard's convert-column flow does). Will redo with that approach if we pursue wiz map support.

PR #3961, opened 2026-06-09
Merge forward fix for Bug #75284, fix text/output assembly showing stale value after selection list change

jkidacki-inetsoft:
Output assemblies bound to selection-filtered tables were displaying results one change behind. When a selection list fires INPUT_DATA_CHANGED in processChange0, executeOutput reads from dmap before processSelections has updated the runtime condition list, caching the previous query result into the assembly value. On the next change the stale dmap entry is used again, keeping the display permanently one step behind.

Fix mirrors the existing dmap clearing already done in the reset() path (line ~1586): after processSelections() runs in processChange0, clear dmap for all entries in clist.getDataList() so that the subsequent execute() call in PlaceholderService re-fetches with the updated conditions.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/ViewsheetSandbox.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/ViewsheetSandbox.java
@@ -2135,13 +2135,6 @@ public class ViewsheetSandbox implements Cloneable, ActionListener {
 
          processScriptAssociation(entry, clist, vs.getScriptDependings());
          processSelections(entry, clist, true, false, new HashSet<>());
-
-         // processSelections() above may have changed conditions so cached data in dmap
-         // should not be reused when output assemblies are subsequently executed.
-         // mirrors the same clearing done in the reset() path. (same rationale as ~line 1586)
-         for(AssemblyEntry dataEntry : clist.getDataList()) {
-            resetDataMap(dataEntry.getName());
-         }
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (changes requested):
The Claude automated review identified a correctness concern that should be addressed before merging:

**Exception-path inconsistency in `processChange0`**

The `dmap`-clearing loop added by this PR is inside the `try` block that catches `DependencyCycleException`:

```java
try {
    // ...
    processSelections(entry, clist, true, false, new HashSet<>());

    // INSIDE try — skipped if processSelections throws DependencyCycleException
    for(AssemblyEntry dataEntry : clist.getDataList()) {
        resetDataMap(dataEntry.getName());
    }
}
catch(DependencyCycleException ex) { ... }
```

If `processSelections()` throws `DependencyCycleException`, conditions may have been partially updated but `dmap` will not be cleared — the same stale-value bug this PR fixes, just on a rarer code path.

The `reset()` path (line ~1585) performs the same clearing *outside* its `processSelections` try/catch. Please move the clearing loop to outside and after the `catch` block to match that pattern.

jshobe-inetsoft (approved):
The previous CHANGES_REQUESTED review concern has been addressed in the follow-up commit. Approving based on the complete Claude automated review of the latest commit, which confirms no blocking issues.

PR #3963, opened 2026-06-09
Bug #75346, fix viewsheet hyperlink not opening in composer preview

jkidacki-inetsoft:
ShowHyperlinkService was provided in both ComposerMainComponent and ViewerAppComponent, creating two separate instances. ComposerMainComponent subscribed to showLinkSheetSubject on its own instance, but VS objects inside ViewerAppComponent emitted on the shadowed instance, so the subscription never fired and the hyperlink appeared to do nothing.

Remove ShowHyperlinkService from ViewerAppComponent.providers so children resolve it from the nearest parent: ComposerMainComponent in the composer, ViewerViewComponent in the portal viewer (where it is now provided), and the element root in the embed context.


jshobe-inetsoft (approved):
Approved based on Claude automated review: no bugs, no regressions, no security concerns. The change is minimal and targeted.

PR #3964, opened 2026-06-09
Reduce cluster write pressure by separating access-time tracking from full session serialization into a dedicated lightweight timestamp cache to prevent extraneous writes on every access() call

ngovus-inetsoft:
**Summary**

Every user interaction with a viewsheet triggered a full serialization of the `RuntimeSheet` to JSON + Zstd and a blocking write to Ignite. This happened on read-only operations (selections, filtering, navigation) as well as mutations, because `accessSheet()` called `amap.put()` unconditionally to keep the last-accessed timestamp up to date. Under concurrent user load this produced a write storm that saturated cluster I/O and blocked HTTP response threads on serialization.

The fix separates access-time tracking from full session persistence by introducing a dedicated `accessTimeMap` (`IgniteCache<AffinityKey<String>, Long>`). Access-time updates now write only an 8-byte timestamp via `putAsync()`, colocated with the session via the same affinity key. Full serialization still happens on mutations through the normal `amap.put()` path; it no longer fires on every read.

The fix separates access-time tracking from full session persistence by introducing a dedicated `accessTimeMap` (`IgniteCache<AffinityKey<String>, Long>`). Access-time updates now write only an 8-byte timestamp via `putAsync()`, colocated with the session via the same affinity key. Full serialization still happens on mutations through the normal `amap.put()` path; it no longer fires on every read.

**Changes**

  - `RuntimeSheetCache`: adds `accessTimeMap`; `updateAccessTime()` writes to it async;
    distributed calls moved outside `ReentrantReadWriteLock`; access time restored after
    session is loaded from Ignite; `accessTimeMap` populated at session open
  - `WorksheetEngine`: `accessSheet()` calls `amap.updateAccessTime()` on touch instead of
    `amap.put()`
  - `RuntimeSheet` / `RuntimeSheetState`: `accessed` and `heartbeat` fields removed from
    serialized state; `setAccessed()` added for cache restoration; constructor initializes
    `accessed` to now as a safe fallback

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/RuntimeSheet.java
+++ b/core/src/main/java/inetsoft/report/composition/RuntimeSheet.java
@@ -151,7 +151,7 @@ public abstract class RuntimeSheet {
       return accessed;
    }
 
-   public void setAccessed(long accessed) {
+   void setAccessed(long accessed) {
       this.accessed = accessed;
    }
 
--- a/core/src/main/java/inetsoft/report/composition/RuntimeSheet.java
+++ b/core/src/main/java/inetsoft/report/composition/RuntimeSheet.java
@@ -99,7 +99,6 @@ public abstract class RuntimeSheet {
 
    RuntimeSheet(RuntimeSheetState state, ObjectMapper mapper) {
       entry = loadXml(new AssetEntry(), state.getEntry());
-      accessed = state.getAccessed();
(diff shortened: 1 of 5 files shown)


jshobe-inetsoft (approved):
Approved based on Claude automated review which confirms all issues resolved and no blocking issues remain.

PR #3968, opened 2026-06-10
Deep recovery -  portal data tab

swaker854:
1) Schema folder nodes now show table count label "(N)" and tooltip. 2) DataTabComponent: replace updateSearchView/currentUrl with
   normalizeLandingRoute to fix blank landing-page on direct navigation.
3) DataDatasourceBrowserComponent: add currentFolderChain, folder action
   visibility/permission getters (rename/move/delete) and loadCurrentFolderDetails.
4) Restore lost templates: data-folder-browser, data-folder-list-view,
   database-physical-model, join-node-graph, datasource-search,
   database-data-model-browser/toolbar, auto-join-tables-dialog,
   physical-table-aliases-dialog, element-tree-node,
   physical-table-tree-node, asset-item-list-view (pattern-count audit).
5) Plan/spec docs: post-mortem update — audit workflow now uses full EPIC
   diff (not comm -12 intersection) and pattern-count-first loss detection
   with a bias-toward-loss instruction to subagents.


jshobe-inetsoft (changes requested):
## Code Review

### Summary

This PR recovers lost Angular templates and components from the epic-74519 merge, adds schema folder table-count labels/tooltips in the physical model tree, and fixes blank landing-page on direct navigation. The Java changes are clean and the frontend recovery work is well-executed. The build and Semgrep checks both pass.

---

### Blocking: Failing check

The **`claude-review`** GitHub Actions check is currently failing, which blocks merge. This appears to be a CI infrastructure issue (the action runner set up but the Claude Code Action did not produce a review result), not a reflection of a code quality problem. The failing check prevents approval per project rules — please investigate and re-run the check, or confirm it can be dismissed if it is a known CI issue.

---

### Code issues (non-blocking, but worth fixing)

#### Java — `DatabaseTreeNode.java`

**`@Nullable` placement on setter**

```java
@Nullable
public void setTableCount(Integer tableCount) {
```

The `@Nullable` annotation is on the method (which returns `void`), not on the parameter. Semantically it should annotate the parameter: `public void setTableCount(@Nullable Integer tableCount)`. The same pattern exists on `setParr`, `setCatalog`, and `setSchema` in the same file, so this is a pre-existing codebase-wide issue — no action required for this PR specifically.

#### Java — `DatabaseTreeService.java`

**N+1 calls during tree load**

`getPhysicalTableChildCount` makes a synchronous `assetRepository.getEntries()` call for every schema-folder node encountered during tree loading:

```java
if(!loadColumns && node.getSchema() != null && !node.getSchema().isEmpty()) {
    node.setTableCount(getPhysicalTableChildCount(entry, principal));
}
```

For databases with many schemas (e.g., 50+ schemas on a SQL Server or Oracle instance), this can add significant latency to the initial tree load. Consider loading the count lazily on schema-node expand, or batching the lookup. This is worth addressing before the epic is merged to main.

#### TypeScript — `data-datasource-browser.component.ts`

**Deprecated RxJS `subscribe` signature**

`withCurrentFolderDetails` uses the two-argument `.subscribe(success, error)` signature deprecated in RxJS 7:

```typescript
this.httpClient.get<DataSourceInfo>(...).subscribe(
    (folder) => { ... },
    () => this.currentFolderDetails = null
);
```

Prefer the object form: `.subscribe({ next: ..., error: ... })`.

**`currentFolderDetails` type**

```typescript
currentFolderDetails: DataSourceInfo = null;
```

With strict null checks this will produce a type error. Declare it as `DataSourceInfo | null = null`.

#### TypeScript — `data-tab.component.ts`

**`temp` query param in URL**

`normalizeLandingRoute` adds `temp: new Date().getTime()` to force re-navigation. This works correctly and prevents an infinite loop, but it leaves a timestamp visible in the browser address bar and pollutes browser history. Consider adding `replaceUrl: true` to the `navigate` call so the synthetic redirect does not create an extra history entry.

---

### Observations (no action required)

- **`docs/superpowers/` files**: The audit log (`audit-2026-06-03.md`) and plan/spec updates are committed to the repo. This is an established convention in the project; the files are development artifacts tracking the recovery work and do not affect runtime behavior.
- The `getNodeLabel` / `getNodeTooltip` methods being `public` is correct — they are consumed by `PhysicalModelController` in the same package.
- The `loadSelectedDataModelChildren` addition in `data-sources-tree-view.component.ts` is a clean fix that eagerly loads lazy data-model folder children on selection, avoiding the user seeing an empty node until they manually expand.
- The sort-header refactor (replacing dual `@if` blocks with CSS class bindings) is a clear improvement.
- The `DatasourceSearchComponent` two-phase search (draft → committed on Enter/button click) is a good UX improvement.

swaker854:
Fix datasource tree PR feedback

Avoid eager schema table-count lookups during physical model tree loading by computing counts from already-loaded children or on schema expansion. Update datasource browser folder-detail types and RxJS subscribe calls, and replace synthetic landing-route redirects in history.

swaker854:
Semgrep failure doesn't seem to be a code issue

jshobe-inetsoft (changes requested):
## Code Review: PR #3968 — Deep recovery - portal data tab

This PR recovers lost UI work from the `epic-74519` merge, including schema folder table-count badges, a redesigned datasource browser header with folder-level actions, a search component draft/commit refactor, data tab route normalization, and a revamped folder browser with worksheet detail panel. The build passes. Two CI checks fail (see below).

---

## CI Check Failures

Two checks fail — both are workflow infrastructure issues, not code defects:

- **`claude-review`** — OIDC token fetch fails: `id-token: write` permission is missing from the workflow. The action cannot authenticate.
- **`Semgrep Scan`** — Fails with `Requires authentication` / missing `security-events: write` permission.

These appear to be systematic workflow configuration problems. They do not reflect defects in this PR's code, but they must be resolved (or confirmed as known-broken) before the checks can be considered passing.

---

## Issues

### 1. `@Nullable` annotation misplaced — `DatabaseTreeNode.java`

```java
@Nullable
public void setTableCount(Integer tableCount) {
```

`@Nullable` on a `void` method is not valid — it would annotate the (non-existent) return type. Either place the annotation on the parameter or remove it (since `Integer` is already nullable by nature):

```java
public void setTableCount(@Nullable Integer tableCount) {
```

---

### 2. Search UI removed with no visible replacement — `data-datasource-browser.component.html`

The search toggle button, search input, and all related `searchVisible`/`toggleSearch()`/`searchQuery` logic are removed from the toolbar. The search results banner still references `currentSearchQuery`, but no UI element remains to initiate a search from within this component.

If search is now driven exclusively from the navigation tree's search component, confirm that and clean up the dead fields (`searchVisible`, `searchQuery`, `searchFunc`) from the component's TypeScript. If search was intended to remain in this view, the UI was inadvertently lost.

---

### 3. `normalizeLandingRoute` uses `new Date().getTime()` as a URL param — `data-tab.component.ts`

```typescript
queryParams: {
    path: "/",
    scope: DATASOURCE_ROOT_SCOPE,
    temp: new Date().getTime()  // ← timestamp changes every visit
},
```

The `temp` parameter forces a fresh navigation even when the router is already at the target URL. With `replaceUrl: true` already preventing history pollution, a stable redirect (without the timestamp) would achieve the same effect and avoid emitting an unnecessary `NavigationEnd` on every component initialization.

---

### 4. Label format duplicated in Java and TypeScript

`DatabaseTreeService.getNodeLabel()` and `refreshExpandedSchemaTableCount()` in `database-physical-model.component.ts` independently implement the same `"name (N)"` / `"N table(s)"` format. Both are consistent now, but a future change to the label format requires updating two places. Consider moving the format logic to the shared `getNodeLabel` / `getNodeTooltip` Java methods (already called from `PhysicalModelController`) and having the frontend always re-read the tree from the server on expand, or at minimum document the dependency.

---

### 5. `withCurrentFolderDetails()` makes a redundant HTTP request — `data-datasource-browser.component.ts`

On every rename/move/delete action, `withCurrentFolderDetails()` fires a fresh `GET` to fetch folder details, even immediately after `loadCurrentFolderDetails()` has already populated `currentFolderDetails`. The fresh-fetch pattern is good for correctness (current permissions), but using the cached value for synchronous UI actions (which the user has just initiated) would be sufficient and avoids a perceptible lag before each action.

---

### 6. `docs/superpowers/` planning artifacts committed to the repo

Three files are added/modified under `docs/superpowers/` (`audit-2026-06-03.md`, the plan, and the design spec). These describe an internal Claude Code loss-detection workflow and post-mortem. Unless the project intentionally maintains Claude Code workflow documentation in the repo, these development-time artifacts should not be included in a production PR.

---

## Positive Observations

- **Schema folder table count** is well-placed: the server computes the count at tree-load time (`getFullDatabaseTree`) and the frontend refreshes it at expand-time (`setExpandedNodeChildren → refreshExpandedSchemaTableCount`). Both code paths are consistent.
- **Sort indicator refactoring** — replacing the dual `@if` icon-block pattern with a single `<span class="sort-indicator">` driven by CSS classes is cleaner and substantially reduces template noise across all list views.
- **Row actions column** moved from inside the first truncated cell to a fixed trailing column — a real UX improvement that fixes the "More Options" button getting hidden under ellipsized text.
- **`DatasourceSearchComponent` draft/commit pattern** — buffering the search string until Enter or the search button is clicked is a solid UX improvement that prevents mid-keystroke searches.
- **Lazy data model folder children** (`loadSelectedDataModelChildren`) — correctly loads children for data model tree nodes when they're navigated to programmatically, fixing the blank-panel scenario.
- **Physical model table tree** — replacing `caret-*-icon` with `chevron-*-icon` and adding explicit `color: var(--inet-text-color)` / `color: var(--inet-icon-color)` tokens is a clean theme compatibility fix.

swaker854:
PR feedvback fix - Remove obsolete inline-search UI from DataDatasourceBrowserComponent and cleanup

- Remove toggleSearch/search/searchFunc/searchInput and associated state
  (searchVisible, searchQuery, searchAssets, Renderer2) — superseded by
  the route-based datasource search flow
- Remove temp timestamp query param from landing-route navigations in
  DataTabComponent (no longer needed after normalizeLandingRoute fix)
- Fix @Nullable annotation placement on DatabaseTreeNode.setTableCount
  (moved from method to parameter where it belongs)
- Add cross-reference comment in DatabasePhysicalModelComponent warning
  that label/tooltip format must stay in sync with DatabaseTreeService (Java)
- Fix vs-image overflow spec to walk the real DOM parent chain instead of
  Angular debug-element tree for Angular 17+ control-flow compatibility
- Drop the now-stale search navigation test that tested the removed search()

jshobe-inetsoft (changes requested):
## Code Review — PR #3968: Deep recovery — portal data tab

Reviewed the full diff (43 files, +1384/−806), CI status, and merge state. The PR's **own changes are well-structured and the TypeScript compiles cleanly** (the Angular compiler reported all type errors and none originate from files in this PR). However, I'm **requesting changes** because there are failing checks and the merge is currently blocked. Details below.

---

### ✅ What looks good

- **Java tree-label feature** (`DatabaseTreeService`, `PhysicalModelController`, `DatabaseTreeNode`) is clean: `getNodeLabel`/`getNodeTooltip` are small, side-effect-free helpers reused by both the recursive tree builder and the controller. `tooltip()` on `TreeNodeModel` is `@Nullable`, so the unconditional `.tooltip(getNodeTooltip(node))` call in `PhysicalModelController` is safe (no NPE). `getTableCount()` is null-guarded everywhere.
- **No regression in existing labels**: `getNodeLabel` falls back to `node.getName()` whenever `tableCount` is null/0, so leaf/column nodes are unaffected.
- **`data-tab.component.ts` redirect loop is guarded**: `normalizeLandingRoute` re-fires on `NavigationEnd`, but `isLandingRoute` returns `false` once `path`/`scope` params are present, so the `router.navigate(..., {replaceUrl:true})` cannot loop infinitely.
- **`datasource-search.component.ts`**: the `draftSearchString` + getter/setter pattern correctly decouples the typed draft from the committed value; emit-on-search (Enter/button) instead of emit-on-keystroke is a reasonable UX change.
- **Cross-language duplication is acknowledged**: the `refreshExpandedSchemaTableCount` comment explicitly points to the Java `getNodeLabel`/`getNodeTooltip` format and asks future editors to keep both in sync — good practice given the format is necessarily duplicated client/server.
- **Test updates are appropriate**: the removed `data-datasource-browser` search-navigation test matches the removal of `search()`; the `vs-image.component.spec.ts` DOM-walk rewrite is a sensible fix for Angular control-flow internal nodes.

---

### ⛔ Blocking — failing checks / merge blocked

1. **`build` (Validation Build) is failing.** The Angular build aborts with:
   ```
   TS2554: Expected 44 arguments, but got 43.
   An argument for 'heartbeatWorkerService' was not provided.
     projects/portal/src/app/vsobjects/viewer-app.spec.ts:547  and  :574
   ```
   Root cause: `ViewerAppComponent`'s constructor has **44** parameters (param 43 = `chartConfigService`, param 44 = `heartbeatWorkerService`), but the two `new ViewerAppComponent(...)` call sites in `viewer-app.spec.ts` (lines 547 & 574) pass only **43** args — they jump from `currentUserService as any` straight to `heartbeatWorkerService`, skipping `chartConfigService`. (The third call site at line ~400 passes all 44 correctly.)

   **Important:** this is **not introduced by this PR** — neither `viewer-app.component.ts` nor `viewer-app.spec.ts` is in this diff, and the same inconsistency already exists on the base branch `epic-74519`. But it is a hard build failure that blocks merging this PR. It needs to be fixed (insert the missing `chartConfigService` arg — e.g. `null` — before `heartbeatWorkerService` at both call sites), either on `epic-74519` directly or folded into this PR / a rebase, so the build goes green.

2. **`claude-review` check is failing** — but this is **infrastructure, not code**: `Could not fetch an OIDC token ... Unable to get ACTIONS_ID_TOKEN_REQUEST_URL`. No action needed on the code; it should pass once the workflow's OIDC/token environment is available (or can be disregarded as a non-code check).

3. **Merge state is `BLOCKED`.** `mergeable` is `MERGEABLE` (no file conflicts), so the block is from the failing required checks above, not from a merge conflict. Once the `build` check passes, the block should clear.

---

### 🔧 Non-blocking suggestions

- **Localization of the table-count label.** Both the Java side (`getNodeTooltip`: `tableCount + " table" + (… ? "" : "s")`) and the TS side (`refreshExpandedSchemaTableCount`) hardcode English text and pluralization. User-visible strings are normally routed through `Catalog`/`_#()`. Consider localizing the `(N)` suffix and the "N table(s)" tooltip, or at least the singular/plural word.
- **Search-within-folder removal.** `toggleSearch`/`search`/`searchFunc`/`searchInput` were removed from both `data-datasource-browser` and `database-data-model-toolbar`. This appears intentional for the redesign, but please confirm the in-folder typeahead search remains reachable for users (or that its removal is the agreed design), since this is a user-facing capability change rather than a pure visual refresh.
- **`withCurrentFolderDetails` silent no-op.** When `currentFolderPathString` is empty and there is no fallback folder, the action is silently dropped. That's safe because the `currentFolder*Disabled` getters gate the buttons, but a defensive early-return comment would make the intent clearer.

---

### Verdict

Requesting changes solely because of the failing **`build`** check (and the resulting `BLOCKED` merge state). The PR's own code is in good shape — once the `viewer-app.spec.ts` constructor-argument mismatch is resolved so the build passes (and the `claude-review` OIDC infra issue clears/re-runs), this is good to approve.

swaker854:
will replace with a new PR after rebase onto epic-74519 again on 6/17

swaker854:
will replace

PR #3972, opened 2026-06-10
Bug #75383 - Enable Color picker for legend selections in circle packing charts

FigurePaper:
## Summary

In the composer format pane, the foreground Color picker was disabled
for legend selections on circle packing charts, even though the
Background color picker remained enabled. This change re-enables the
Color picker so legends behave consistently with every other chart type.

## Root cause

isNonEditableChartVOSelected() in vs-formats-pane.component.ts marks a
chart region as non-editable for text color when an outer circle is
selected, since only the innermost circle shows an editable label. It
derived innermostSelected from the selected circle ("vo") regions and
returned !innermostSelected.

A legend selection contains no "vo" regions, so the filtered list was
empty, innermostSelected was false, and the method returned true,
disabling the Color picker. The Background picker was unaffected because
its disable condition is separately guarded in the template.

## Fix

Return false (editable) early when the selection contains no circle
("vo") region. The innermost-circle check now applies only when a circle
is actually selected, so outer-circle selections stay non-editable and
innermost-circle selections stay editable.

## Tests

Added a unit test in vs-formats-pane.spec.ts covering three circle
packing selection paths:
- legend selection -> Color enabled
- outer (non-innermost) circle -> Color disabled
- innermost circle -> Color enabled

### DIFF
--- a/web/projects/portal/src/app/vsobjects/format/vs-formats-pane.component.ts
+++ b/web/projects/portal/src/app/vsobjects/format/vs-formats-pane.component.ts
@@ -301,6 +301,11 @@ export class VSFormatsPane implements OnInit, OnChanges {
                      return false;
                   }
 
+                  // non-circle selections (legend, title, etc.) remain editable
+                  if(vos.length == 0) {
+                     return false;
+                  }
+
                   // innermost circle is same as selecting the text label.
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
Approving based on the complete Claude automated review, which confirms no blocking issues.

PR #3979, opened 2026-06-10
Bug #75361, fix Angular @for track expressions to use stable primitive identifiers

jshobe-inetsoft:
## Summary

- 185 Angular template files in the `em` and `portal` projects had `@for` loops tracking by object/array identity (e.g. `track entity`, `track item`, `track folder`)
- These were introduced during the Angular 20→21 migration from `*ngFor` to `@for` syntax and trigger NG0956 warnings, causing full DOM rebuilds on every data change
- Each loop now uses either a stable primitive property (e.g. `track entity.name`, `track folder.path`) or `track $index` where no stable unique identifier exists

## Root Cause

The Angular 20→21 upgrade migration (commit `b667af680`) converted `*ngFor` directives to `@for` syntax but left all `track` expressions as bare object references. Angular 17+ requires `track` to resolve to a primitive (string/number) to uniquely identify items; tracking by object identity triggers NG0956 and rebuilds the entire DOM collection on each change detection cycle.

## Fix

Replaced all non-primitive `track` expressions across 185 HTML template files:
- Objects with a stable unique `name` or `path` property → `track item.name` / `track item.path`
- Objects without a stable unique identifier, or where the identifier could duplicate → `track $index`
- Primitives (string[], number[]) — left unchanged, they were already correct

## Test Plan

- [x] `npm run build` passes with no template compilation errors
- [x] `npm run test` — 736 tests pass (21 expected failures)
- [x] `npm run lint` — all files pass linting
- [x] Manually verified NG0956 warning no longer appears on `app/portal/tab/data` open

Fixes Redmine #75361.


jshobe-inetsoft:
Thanks for the thorough review. The four missed files have been fixed in the latest commit:

- **`calc-group-option.component.html` line 84**: `track agg` → `track $index` (consistent with the other `aggrs` loops already fixed in this file)
- **`expert-named-group-dialog.component.html` line 28**: `track group` → `track group.name` (name is the stable identifier used throughout the template)
- **`aggregate-pane.component.html` lines 29, 69, 89, 127**: All four inner loops changed to `track $index` (options are bound by index via `[ngValue]="_i"`)
- **`variable-table-list-dialog.component.html` lines 39, 51**: `track _col` → `track $index`

The style note about the aliased `$index` pattern (`track _j; let _j = $index`) on `aggregate-pane.component.html:19` and `ws-merge-join-editor-pane.component.html:20` is acknowledged. These are functional and correct; leaving them as-is to keep the diff minimal.

jkidacki-inetsoft (approved):

jkidacki-inetsoft (changes requested):
Looks like you'll need to resolve the merge conflict first.

jkidacki-inetsoft (approved):

PR #3983, opened 2026-06-11
feat(wiz): AI geo-mapping backend — choropleth detect/apply + reopen-guard fix

larryliang-inetsoft:
Backend for the wiz AI geographic-map feature: build a country choropleth shaded by a category (e.g. most-popular film genre per country), with unmatched names auto-resolved by Claude.

## What's here
**Geo service (`WizGeoService`, DTOs, `WizViewsheetController`, `WizGeoMappingResolver` + test)**
- `geo/detect`: marks a column geographic → updateAllGeoColumns → executes data → autoDetect → reports `{ geoType, layer, matchedCount, unmatched, candidateFeatures }`.
- `geo/apply`: records resolved value→geoCode mappings + drops, reports `stillUnmatched`.

**Real map rendering (this is what made it work end-to-end):**
- Convert the runtime chart to a `VSMapInfo` via `ChangeChartTypeProcessor` (the product's "set geographic → change type to Map" path — the recommender can't offer a map at recommend time since a string geo column isn't tagged yet, so `create_viewsheet(map)` was substituted to point).
- `fixMapBinding()` re-homes bindings deterministically off the detected column: **geo = country** (carrying its auto-detected `GeographicOption`), **color = the category** (fully-formed dimension ref). The processor's "first dim on x" heuristic otherwise mis-binds when the geo column was on shape.
- NIL static shape frame when no point/size field → **filled polygons (choropleth)**, not centroid markers (mirrors `MapChartFilter`).
- `GraphUtil.fixVisualFrames` after re-homing → the color aesthetic gets a `CategoricalColorFrame` (a null frame caused `graph.gen.failed` at render).
- **Persistence:** `save_viewsheet` reads the source viewsheet from the asset repository, not the live runtime, so the map conversion + resolved mappings were lost on save. `geo/detect`+`geo/apply` now persist the converted runtime back to the session asset (`setViewsheet`) via a new `viewsheetIdentifier` on the request DTOs.

## ⚠️ Separate fix included — needs @hunterhu review
[...]

PR #3984, opened 2026-06-11
fix(wiz): /viewsheet/open accepts saved-viz (COMPONENTS) folder — restores reopen + companion viewer

larryliang-inetsoft:
## Problem
`4d383b07f` narrowed `/api/wiz/viewsheet/open`'s guard to `VISUALIZATION_ROOT_FOLDER_PATH` only. But wiz-services `savedVisualizationService.openSaved` posts **COMPONENTS-folder** identifiers to this endpoint — that's the path used by `reload_saved_visualization` and the companion viewer. So every saved-visualization reopen now fails with:

```
java.lang.IllegalArgumentException: Viewsheet is not in the managed visualizations folder: visualization-components-…
```

This re-introduces the blank companion-viewer bug that #3901 originally fixed (every reopened saved chart 500s).

## Fix
Accept **either** folder — ROOT (session viewsheets loaded from a conversation) **or** COMPONENTS (reopened saved visualizations). The change is additive, so the conversation-load case @hunterhu intended in `4d383b07f` still works:

```java
if(path == null ||
   !(path.startsWith(VISUALIZATION_ROOT_FOLDER_PATH + "/") ||
     path.startsWith(VISUALIZATION_COMPONENTS_FOLDER_PATH + "/")))
{
   throw new IllegalArgumentException("Viewsheet is not in the managed visualizations folder: " + path);
}
```

## Notes
- @hunterhu — flagging since this touches your `4d383b07f`; please confirm the conversation-load path doesn't actually require rejecting COMPONENTS (it shouldn't — they're disjoint folders).
- Split out of the geo-mapping PR #3983 for independent review. #3983 needs this merged for its saved maps to reopen / render in the viewer.
- Verified: with this fix, `reload_saved_visualization` and the companion viewer reopen saved charts; without it both 500.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/ViewsheetRuntimeController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/ViewsheetRuntimeController.java
@@ -75,9 +75,14 @@ public class ViewsheetRuntimeController {
       // VISUALIZATION_ROOT_FOLDER_PATH holds the viewsheets used by wiz session messages.
       // VISUALIZATION_COMPONENTS_FOLDER_PATH holds standalone saved visualizations that are
       // visible in the Wiz Portal Visualizations tab and can be combined into a Dashboard.
-      // This endpoint opens session viewsheets, so the guard must check
-      // VISUALIZATION_ROOT_FOLDER_PATH, not VISUALIZATION_COMPONENTS_FOLDER_PATH.
-      if(path == null || !path.startsWith(WizVisualizationService.VISUALIZATION_ROOT_FOLDER_PATH + "/")) {
+      // Both are reopened through this endpoint: session viewsheets when loading a conversation,
+      // and saved visualizations on reload_saved_visualization / the companion viewer (wiz-services
+      // savedVisualizationService.openSaved posts the COMPONENTS-folder identifier here). Accept
+      // either folder — restricting to ROOT alone rejects every saved chart (blank viewer); see #3901.
+      if(path == null ||
+         !(path.startsWith(WizVisualizationService.VISUALIZATION_ROOT_FOLDER_PATH + "/") ||
+           path.startsWith(WizVisualizationService.VISUALIZATION_COMPONENTS_FOLDER_PATH + "/")))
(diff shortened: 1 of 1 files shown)


larryliang-inetsoft:
@NeverStopFMH could you review this? It adjusts the `/viewsheet/open` guard you changed in 4d383b07f.

Your commit restricted the guard to `VISUALIZATION_ROOT_FOLDER_PATH` only, but wiz-services `savedVisualizationService.openSaved` posts **COMPONENTS-folder** identifiers to this endpoint (`reload_saved_visualization` + the companion viewer), so every saved-viz reopen now 500s with *"Viewsheet is not in the managed visualizations folder"* — the blank-viewer bug #3901 fixed.

This makes the guard accept **either** folder (ROOT or COMPONENTS), which is additive — the conversation-load case you intended still works. Could you confirm the conversation path doesn't actually need to reject COMPONENTS? If it does, we should find a different discriminator. This is blocking the geo-map work (#3983).

NeverStopFMH (approved):
I think the current change is correct.

The viewsheets generated within a conversation are stored under `VISUALIZATION_ROOT_FOLDER_PATH`. These are the viewsheets produced through the UI chat flow and used when rendering messages in the conversation.

I reviewed the Claude plugin flow (`save_viewsheet` + `open_companion_viewer`). That flow saves the viewsheet as a standalone visualization under `VISUALIZATION_COMPONENTS_FOLDER_PATH`. Functionally, this is equivalent to using the Save button on a message in the UI to save it as a visualization.

Given that these are two different storage locations serving different purposes, I believe allowing both `VISUALIZATION_ROOT_FOLDER_PATH` and `VISUALIZATION_COMPONENTS_FOLDER_PATH` in the guard is the correct behavior. The original conversation-generated viewsheet path remains restricted to `VISUALIZATION_ROOT_FOLDER_PATH`, while saved visualizations created through the plugin are able to reopen correctly from `VISUALIZATION_COMPONENTS_FOLDER_PATH`.

Image

fredaan (approved):

PR #3988, opened 2026-06-11
Bug #75381, fix NaN being published to CloudWatch as a scaling metric

jkidacki-inetsoft:
ScalingMetric.update() now clamps NaN and negative values from calculate() to 0 before buffering, preventing NaN from propagating through the moving average into getServerUtilization(). Sources include getCpuLoad()/-1.0 "not available" returns and 0/0 division in CacheSwapWaitScalingMetric when the threshold is 0. CloudWatchMetricsPublisher.pushMetrics() additionally guards against non-finite values at the publish boundary to avoid a CloudWatch 400.


jshobe-inetsoft (approved):
Approved — Claude's review confirms the NaN/negative guard in ScalingMetric.update() is correct and well-placed.

PR #3938, opened 2026-06-08
Bug #75351, fix Show Details date for Year-week date comparison charts

larryliang-inetsoft:
## Summary

On a Date Comparison chart configured for Year-over-year with a WeekOfYear context (e.g. same day-of-week across years, grouped by week), clicking a prior-year bar and choosing **Show Details** returned rows for a date one week off from the bar's label/tooltip — e.g. the bar labeled `2020-02-06` drilled to `2020-02-13`.

## Root Cause

DC week buckets are keyed by date *part* values — `WeekOfYear` (encoded `MM*10 + weekOfMonth`) and `DayOfWeek` — carried in a `DCMergeDatePartFilter.MergePartCell`. The same `(month, week-of-month)` part value can resolve to a different actual week across years, so `MergePartCell.getEquivalenceCell()` recomputes the WeekOfYear value against the comparison period's year and returns a corrected cell when it differs.

The axis-label path and the **crosstab** drill path apply this correction (`CrossTabFilterUtil.getEquivalenceNewTuple` → `getEquivalenceCell()`), but the **chart** drill-to-detail path did not: `VSDataSet.getFieldValue()` built the detail field value from the raw (uncorrected) `MergePartCell`, so the Show-Details condition targeted the wrong (raw) week.

Confirmed by replicating StyleBI's `datePart("wy")` + `getEquivalenceCell` logic: the bucket's WeekOfYear part `15` (Jan, week 5) normalizes on 2020 to Feb week 1 (equivalence value `21`), both reconstructing to `2020-02-06` — the correct date.

## Fix

In `VSDataSet.getFieldValue()` (the chart drill-to-detail field extractor), when the cell is a `DCMergeDatePartFilter.MergePartCell` and `getEquivalenceCell()` returns non-null, use the equivalence cell — so the part value fed into the detail condition matches the displayed/data week, mirroring the crosstab path.

`getEquivalenceCell()` returns null unless a `WeekOfYear` merged ref's recomputed value diverges (and the period value is a Date), so non-WeekOfYear DC configs and non-diverging weeks are unaffected. Rendering is unaffected — `getFieldValues` only feeds selection/Show-Details/brush, not plotting.

## Test Plan

1. Import `Year_SameDay.vso`, open in composer (DC: previous 1 year, Same Day, granularity Day, context Week).
2. Select the 2020 bar at the `2021-02-04` / `2020-02-06` position (value 48) and click **Show Details**.
[...]

PR #3989, opened 2026-06-11
Bug #75351, align Same-Day date-comparison weeks across years (sequential week-of-year)

larryliang-inetsoft:
## Summary

In a Year / **Same Day** date-comparison chart grouped by week, at month-boundary weeks the previous-year bar was missing (or misplaced) and **Show Details returned the wrong week** (e.g. `2020-02-13` instead of `2020-02-06`), with the axis label also off by a week.

## Root Cause

The WeekOfYear grouping used the `month*10 + weekOfMonth` encoding (`datePartForceWeekOfMonth('wy', …)`). At a month boundary the **same relative week gets a different value across years** — e.g. `2021-02-04` → `15` (its week starts Jan 31) but `2020-02-06` → `21` (its week starts Feb 2). So the two years' bars land in **different x-buckets**: the current-year bucket has no comparison bar (compounded by PR #3818's orphan-cell filter, which drops buckets lacking current-year data), and the label/drill resolve a week off.

No calendar `(month, week)` encoding can align a weekly sequence across years (the sequences drift relative to months differently each year) — only a **sequential week-of-year** does.

This is effectively a regression reshaped by PR #3818 (Bug #75152) layered on a pre-existing weekly-alignment gap. PR #3938's drill fix only addressed the detail condition, not the grouping.

## Fix

Group **Same-Day** week comparisons by the **sequential week-of-year** (`datePart('ww')` → `Calendar.WEEK_OF_YEAR`), so the Nth week aligns across periods. A new runtime flag `VSDimensionRef.dcSequentialWeek` marks the generated ref so:
- `DateComparisonUtil.formatPartMergeCell` decodes the label via `WEEK_OF_YEAR` (not the MMW math), and
- `DCMergeDatePartFilter.getEquivalenceCell` skips the cross-year remap (sequential weeks already align; the `month*10` math would corrupt them).

**Scoped to the `SAME_DAY` interval only** — week-to-date and same-week configs keep the `wy` encoding and its week-of-month forcing (the flag defaults false, so all other DC paths are unchanged). Mirrors the existing `ignoreDcTemp` runtime-flag pattern.

This branch also includes an earlier commit that derives the legacy `wy` equivalence from the cell's own date (drill correctness for the MMW path under non-Sunday week starts).

## Test Plan

- Open `Year_SameDay.vso` → the DC chart. At the `2021-02-04` position the **2020 bar now renders**, the axis label / tooltip read `2021-02-04` / `2020-02-06`, and selecting the 2020 bar → **Show Details** returns the `2020-02-06` rows (`18+19+11`). *Verified live on a built server.*
- New regression test `DCMergeDatePartFilterTest.testSequentialWeekSkipsEquivalence` (a sequential-week ref skips the equivalence remap); existing tests still pass (4/4).
- **Reviewer note:** the change is scoped to SAME_DAY; please sanity-check **week-to-date / same-week / crosstab** DC week charts are unaffected (they use the unchanged `wy` path).

Fixes Redmine #75351.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/VSDataSet.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/VSDataSet.java
@@ -818,9 +818,10 @@ public class VSDataSet extends AbstractDataSet implements AttributeDataSet {
       if(obj instanceof DCMergeDatesCell) {
          obj = ((DCMergeDatesCell) obj).getOriginalData();
       }
-      // A WeekOfYear part value can map to a different actual week across years;
-      // use the equivalence cell so the drill condition matches the displayed week
-      // (mirrors the crosstab path, CrossTabFilterUtil.getEquivalenceNewTuple).
+      // A legacy WeekOfYear part value can map to a different actual week across years; use the
+      // equivalence cell so the drill condition matches the displayed week. For Same-Day
+      // sequential-week grouping getEquivalenceCell() returns null (already aligned), leaving
+      // the raw cell. (Bug #75351)
       else if(obj instanceof DCMergeDatePartFilter.MergePartCell) {
          DCMergeDatePartFilter.MergePartCell equivalenceCell =
             ((DCMergeDatePartFilter.MergePartCell) obj).getEquivalenceCell();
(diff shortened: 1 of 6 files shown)


jshobe-inetsoft (approved):
Approved — Claude's review confirms no blocking issues. The dcSequentialWeek flag correctly threads through grouping, label decode, and drill-detail paths.

PR #3990, opened 2026-06-11
Bug #75403 - Restore inline-SVG chart rendering dropped in Angular 20->21 migration

FigurePaper:
## Summary

With `graph.svg.inline=true`, charts rendered as an opaque `` rather than
inline SVG, so chart hover dimming did not work. The entrance "animations" that
appeared to work were just the image fade-in (CSS inside the SVG runs even when
it's loaded in an ``), not the real SVG bar-grow — which masked the
regression.

## Root cause

The Angular 20->21 migration (b667af680) regressed two parts of the inline-SVG
feature originally added in 825ddc3ac:

- `chart-plot-area.component.html` was rewritten to the new `@if`/`@for` control
flow and lost the `inlineSvg` branch, leaving only ``.
- `ChartInlineSvgDirective` was dropped from the standalone component's
`imports`, so even with the template binding restored, `[chartInlineSvg]`
silently bound to nothing — the `<div>` rendered but the directive never attached
and never inlined the SVG.

Inline SVG is required for hover dimming: the directive toggles `inetsoft-active`
on the hovered element and the server-injected `:has()` CSS dims the others.
Inside an ``, the SVG is an opaque replaced element, so external JS and CSS
cannot reach it.

## Changes

- `chart-plot-area.component.html`: when `inlineSvg` is enabled, render `<div
[chartInlineSvg]="getSrc(...)">` instead of ``
for both the plot tile and the pan double-buffer.
- `chart-plot-area.component.ts`: register `ChartInlineSvgDirective` in the
component `imports`.
- `chart-plot-area.component.spec.ts`: regression test asserting the directive
attaches when `inlineSvg` is on.

### DIFF
--- a/web/projects/portal/src/app/graph/objects/chart-plot-area.component.html
+++ b/web/projects/portal/src/app/graph/objects/chart-plot-area.component.html
@@ -66,8 +66,7 @@
               (onLoaded)="loaded(true, tile)"
               (onError)="loaded(false, tile)">
             </div>
-          }
-          @else {
+          } @else {
             <img [chartImage]="getSrc(tile, container)"
               (onLoading)="loading(tile)"
               (onLoaded)="loaded(true, tile)"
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):
Approved — Claude's review (resolved) on the latest commit confirms the inline-SVG regression fix is correct and the regression test is well-targeted.

PR #3991, opened 2026-06-11
Bug #75368, fix VSTable @for track (NG0955 / change-detection freeze)

larryliang-inetsoft:
## Summary

On a viewsheet page containing a data table (VSTable), changing a dropdown could freeze the page (0 fps, ~100ms change detection; `_VSTable` the slowest component at ~40ms). The browser console showed **NG0955: duplicated keys for a given collection** (track keys all `"0"`).

## Root Cause

The VSTable header/row/cell `@for` loops tracked items with `AbstractVSObject.trackByConstant`, which returns `0` for **every** item. A constant `trackBy` was a valid `*ngFor` "reuse DOM by position, don't recreate" optimization before the Angular 20→21 upgrade, but Angular 21's `@for` requires the track expression to **uniquely identify** items. A constant produces duplicate keys → **NG0955** → Angular cannot reconcile the collection and destroys/recreates the row/cell DOM on every change-detection pass. That makes each VSTable check ~40ms, so when a dropdown change schedules a `setTimeout` refresh the page freezes.

## Fix

Track the four VSTable `@for` loops by `$index` instead of `trackByConstant`. This yields unique keys (eliminating NG0955) while **preserving reuse-by-position**, which is the correct semantics for the server-side virtual-scroll table — the visible window's row/cell DOM is reused and bindings updated as the user scrolls (object-identity tracking would instead recreate every node on each scroll, which is worse).

Also removed the now-unused `trackByConstant` helper from `AbstractVSObject` (no remaining references; `trackByIdx` is retained) and updated the stale comment in `positionDataAnnotations`. The annotation `setTimeout` is unchanged.

## Test Plan

- Open a page with a data table (e.g. **Examples/Return Analysis**), change the "Dropdown Selection to Filter" dropdown repeatedly → page stays smooth, **no freeze**, and **no `NG0955`** in the console.
- Verified: `ng build portal` compiles cleanly, `ng lint portal` passes, and `vs-table.spec.ts` passes (8/8).

Fixes Redmine #75368.

### DIFF
--- a/web/projects/portal/src/app/vsobjects/objects/abstract-vsobject.component.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/abstract-vsobject.component.ts
@@ -190,12 +190,6 @@ export abstract class AbstractVSObject<T extends VSObjectModel> extends CommandP
       return index;
    }
 
-   // For use in ngFor to prevent excessive DOM manipulation due to component creation/destruction.
-   // The child views will still be checked for changes.
-   public trackByConstant(index, item) {
-      return 0;
-   }
-
    /**
     * Add an assembly level annotation to this component
     */
--- a/web/projects/portal/src/app/vsobjects/objects/table/vs-table.component.html
+++ b/web/projects/portal/src/app/vsobjects/objects/table/vs-table.component.html
@@ -138,11 +138,11 @@
           [style.z-index]="1">
           <table>
             <tbody>
-              @for (_header of tableHeaders; track trackByConstant(i, _header); let i = $index) {
+              @for (_header of tableHeaders; track $index; let i = $index) {
                 <tr
                   [class.hidden-row]="getRowHeight(i) <= 0"
                   [style.height.px]="getHeaderRowHeight(i)">
-                  @for (_cell of _header; track trackByConstant(j, _cell); let j = $index) {
+                  @for (_cell of _header; track $index; let j = $index) {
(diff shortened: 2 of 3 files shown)


larryliang-inetsoft:
**Cross-reference / not a duplicate of #3979:**

This PR is complementary to **#3979** (Bug #75361, "fix Angular @for track expressions to use stable primitive identifiers") — they fix *different* Angular-21 `@for` track problems and touch *disjoint* files:

| | This PR (#3991) | #3979 |
|---|---|---|
| Error | **NG0955** (duplicate keys) | **NG0956** (object-identity track → full DOM rebuild) |
| Pattern | `track trackByConstant(...)` → constant `0` for every item | `track <object>` (e.g. `track entity`) |
| Files | `vs-table.component.html` + `abstract-vsobject.component.ts` (VSTable only) | 185 em/portal templates — **does not include vs-table.component.html** |

#3979's sweep only converted object-identity tracks; it did not touch the `trackByConstant(...)` method-call tracks in VSTable (0 references to `trackByConstant`). So this VSTable-specific NG0955 / change-detection-freeze fix is not covered there. No file overlap → no merge conflict regardless of merge order.

jshobe-inetsoft (approved):
Approved — Claude's review confirms the fix is correct. Switching from trackByConstant (returns 0) to track $index resolves NG0955 for VSTable.

PR #3996, opened 2026-06-12
fixed Bug 75157 75144

bonnieshi43:
Bug #75144 #75157, fix data source browser drag/drop, status refresh, path routing, and logical model route encoding
 ## Bug #75144 — DataDatasourceBrowserComponent

  - **updatingStatus never reset on success**: moved cleanup into `finalize()` so
    `updatingStatus` and `requests.remove(sub)` are always called on both success
    and error, not only on error.

  - **Null crash on empty-pane drop**: `dropAssets(event, null)` dereferenced
    `datasource.path` without a null guard; replaced with
    `datasource?.path ?? this.currentFolderPathString`.

  - **stopPropagation too late**: moved `event.stopPropagation()` before the
    non-folder early-return so propagation is always stopped regardless of target
    type.

  - **Wrong parent path in search results**: `getParentRouterLinkParams` was
    taking the second-to-last path segment instead of the full prefix; replaced
    with `lastIndexOf("/")` + `substring()`.

## Bug #75157 — DatabaseDataModelBrowserComponent

  - **Root physical model name not encoded**: added `Tool.byteEncode()` around
    `physicalItem.name` in the root physical model route, matching the encoding
    already applied to extended physical models.

  - **Root logical model name not encoded**: added `Tool.byteEncode()` around
    `logicalModelItem.name` in the root logical model route, and restored the
    accidentally dropped `parent` route parameter.

## Tests

  Removed `it.fails()` wrappers from all four confirmed-bug cases in
  `data-datasource-browser.component.tl.spec.ts` and two cases in
  `database-data-model-browser.component.tl.spec.ts` as the underlying bugs are
  now fixed.

### DIFF
--- a/web/projects/portal/src/app/portal/data/data-datasource-browser/data-datasource-browser.component.tl.spec.ts
+++ b/web/projects/portal/src/app/portal/data/data-datasource-browser/data-datasource-browser.component.tl.spec.ts
@@ -33,23 +33,6 @@
  *   Group 6 [Risk 2] - action guards: direct method calls must honor disabled UI contracts.
  *   Group 7 [Risk 3] - search result location: full parent path must be preserved.
  *
- * Confirmed bugs (it.failing - remove wrapper once fixed Issue #75144):
- *
- *   Bug A - status-error-leaves-updatingStatus-true (Group 2):
- *     fetchDataSourceStatuses resets updatingStatus only in the complete callback. RxJS does not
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):
Approved — Claude's review confirms all four bugs are correctly fixed. The fix is minimal and well-targeted.

PR #4005, opened 2026-06-12
Performance: reduce Ignite write churn in RuntimeSheetCache

ngovus-inetsoft:
## Summary

This PR reduces unnecessary Ignite cluster write traffic originating from the `RuntimeSheetCache`, improving performance under concurrent viewsheet/worksheet load. (Merged from branch performance-runtimesheetcache).

- **Separate access-time tracking**: Introduced a lightweight `accessTimeMap` (session ID → timestamp, 8 bytes per entry) so that `get()` no longer triggers a full `CompressedSheetState` write to Ignite on every session access. Access-time updates are debounced and written asynchronously.

- **Write-gated persistence**: Added `@ClusterWriteMethod` annotation and `SheetWriteAspect` (Spring AOP `@AfterReturning`) so that Ignite state is only written back after methods that genuinely mutate session state. The cluster-proxy annotation processor was updated to emit explicit `writeSheet()` calls in generated remote proxies, maintaining the same gate for cross-node invocations. Approximately 334 annotations were applied across 192 service files covering the full composer, binding, worksheet, and wizard layers.

- **Write debounce**: A 1-second trailing-edge debounce on `writeSheet()` coalesces rapid mutation sequences (e.g., multiple property dialog saves) into a single Ignite `put` per session window, consistent with original access-time write debouncer.

- **Checkpoint serialization optimization**: `RuntimeSheet.XSwappableSheet.getXml()` now reads serialized checkpoint bytes directly from the `.tdat` swap file instead of deserializing and re-serializing the full DOM on every `saveState()` call. The swap-file passthrough avoids an unnecessary deserialize/re-serialize round-trip that was previously performed for all checkpoints, when the raw bytes from the .tdat file are sufficient except for the specific case of worksheets with embedded table data which require re-serialization to strip the ISTEMP flag.

- **Bug fixes**: Fixed a session resurrection race where a pending debounced write could re-insert a disposed session after `closeSheet()` removed it; fixed a thread pool leak where `affinityExecutor` was not shut down on `WorksheetEngine.dispose()`.

## Test plan

- [ ] Build passes: `./mvnw clean install -DskipTests -pl core`
- [ ] Two-node cluster: open a viewsheet on node A, mutate state in the composer, route the next request to node B, confirm state is preserved
- [ ] Verify Ignite write rate drops significantly under concurrent read-heavy load compared to main
- [ ] Undo/redo still functions correctly across node boundaries after composer edits

### DIFF
--- a/build-tools/cluster-proxy-annotations/src/main/java/inetsoft/cluster/ClusterWriteMethod.java
+++ b/build-tools/cluster-proxy-annotations/src/main/java/inetsoft/cluster/ClusterWriteMethod.java
@@ -1,0 +1,36 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.cluster;
+
+import java.lang.annotation.*;
+
+/**
+ * Marks a {@link ClusterProxyMethod} method as a mutation — one that modifies the
+ * RuntimeSheet state and requires the session to be written back to the distributed cache.
+ * {@code SheetWriteAspect} intercepts calls to methods with this annotation and
+ * triggers an async write-back via {@code WorksheetService.writeSheet()}.
+ *
+ * <p><strong>Contract:</strong> the annotated method must have a
+ * {@code @ClusterProxyKey String} as its first parameter. {@code SheetWriteAspect}
+ * extracts the sheet ID from {@code args[0]}; if {@code args[0]} is not a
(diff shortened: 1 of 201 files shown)


jasonshobe (changes requested):
Claude's latest review identified **medium-to-high severity issues** that should be addressed before merging:

**1. Session resurrection race in `writeSheet`** (Medium-High): `debouncer.cancel()` in `closeSheet` only cancels debounces that haven't fired yet. A debounce that already fired and has a `CompletableFuture.runAsync` queued in `affinityExecutor` can still re-insert a disposed session back into the cache. Fix: re-check containment inside the async lambda (`if(amap.containsKey(id)) { amap.put(id, rs); }`).

**2. `RecycleTask` doesn't cancel the write debounce** (Medium): `RecycleTask.run()` calls `amap.remove(id)` without a preceding `debouncer.cancel("writeSheet_" + id)`, leaving pending write debounces live after session timeout.

**3. `@ClusterWriteMethod` on close methods** (Low-Medium): `closeViewsheet` and `closeWorksheet` are annotated with `@ClusterWriteMethod`, but these methods call `closeSheet` internally which already cancels and removes the session. The AOP aspect then schedules a write-back for a session that no longer exists, creating another resurrection window.

**4. Missing `ACCESS_TIME_MAP_NAME` Ignite cache registration**: `RuntimeSheetCache.getAccessTimeCache` calls `cluster.getCache(ACCESS_TIME_MAP_NAME)` but this cache may not be pre-registered like `CACHE_NAME` is.

Please address these issues before merging.

jshobe-inetsoft (changes requested):
See previous comment

jkidacki-inetsoft (approved):

jshobe-inetsoft (approved):
All previous blockers have been addressed per the Claude review (resolved). Checks pass and PR is mergeable. Approving for merge.

PR #4014, opened 2026-06-15
Bug #75411, fix systemCpuDetail publishing wrong metric source

jkidacki-inetsoft:
systemMemory.calculate() was passed as the systemCpuDetail argument instead of systemCpu.calculate(), causing the Prometheus inetsoft_scaling_utilizationDetail{scope="container",type="cpu"} gauge to display system memory samples


jshobe-inetsoft (approved):
All checks pass, and the Claude review (approved, resolved) confirms this is a correct and well-scoped fix. Approving for merge.

PR #4019, opened 2026-06-16
Worksheet agent pairing — runtime-join foundation (Phase 1, off by default)

larryliang-inetsoft:
## Summary

Phase 1 of the StyleBI worksheet-agent feature: lets a second session of the **same logical user** (a Claude-Code agent) join a worksheet `RuntimeWorksheet` the user already has open in the composer, mutate it, and have the browser re-render live — via a **pairing-token-authorized** path, **off by default**.

This is foundational plumbing: there is **no production caller** of join/edit yet (only the `mint` endpoint is exposed), and the whole capability is gated behind `worksheet.agent.pairing.enabled` (false unless an admin sets it).

## What's included
- **`PairingGrant` + `WorksheetPairingService`** — single-use, 5‑min‑TTL pairing codes (SecureRandom, ambiguity-free alphabet, injectable clock).
- **`WorksheetEngine.getWorksheetForPairing`** — a NEW accessor that fetches the shared runtime and **refreshes its heartbeat via `accessSheet`** (so it isn't reaped mid-edit), deliberately bypassing the per-session `SRPrincipal` check. `getSheet`/`matches` are **untouched**.
- **`WorksheetJoinService`** — authorizes the join only when the feature is enabled, the code is valid (single-use), and the agent is the **same logical user** (`IdentityID` name+org). Audited on every grant/reject; never logs the code.
- **`WorksheetAgentFeature`** — the off-by-default flag, enforced in both the join service and the mint endpoint.
- **`WorksheetEventService`** — records the browser's STOMP `socketSessionId`/user on worksheet open (mirrors the viewsheet path).
- **`WorksheetAgentBroadcastService`** — pushes a `RefreshWorksheetCommand` to the browser's recorded socket **user/session** (mirrors `SharedFilterService`; omits `inetsoftClientId` so the client treats it as a broadcast).
- **`WorksheetAgentEditService`** — end-to-end skeleton (join → remove column → broadcast); the full mutator set is Phase 2/3.
- **`WorksheetPairingController`** — the `mint` endpoint (flag-gated, 403 when off).

## Security posture (locked)
- Trust boundary: **same logical user** (name+org), single-use code, 5‑min TTL, audit log.
- **Off by default** behind `worksheet.agent.pairing.enabled`.
- The session-check bypass (`getWorksheetForPairing`) has a single authorized caller (`WorksheetJoinService`) and an engine-level audit log.
- A StyleBI security owner should review the carve-out before the flag is enabled in any shared deployment.

## Test plan
- [x] 20 unit tests green (`./mvnw -o -pl core test -Dtest='inetsoft.web.wiz.worksheet.*Test' -Dgroups=core`): pairing TTL/single-use, same-/different-user join, feature-off (no code consumed), null-code/null-user rejection, socketSessionId recording, broadcast routing (browser user, no clientId, null-session no-op), end-to-end edit.
- [ ] Manual smoke (deferred to Phase 4 e2e): open a worksheet, mint a code, join as the same user from a second session, edit, confirm the composer re-renders.

## Follow-ups (tracked in the Phase 2 plan)
Cluster-shared pairing store + cache-touch propagation, join-once reusable session, controller null-guard hardening, session-derived `socketSessionId` (remove the spoofable param), server-side recompute of dependents, shared test-harness extraction.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/WorksheetEngine.java
+++ b/core/src/main/java/inetsoft/report/composition/WorksheetEngine.java
@@ -590,6 +590,8 @@ public abstract class WorksheetEngine extends SheetLibraryEngine implements Work
     * fetches the shared runtime the same way getSheet() does, but WITHOUT the
     * per-session SRPrincipal {@code rs.matches(user)} check, so a second login session
     * of the same logical user can join the open runtime. It does NOT relax getSheet().
+    * Like getSheet(id, user, true), it also refreshes the runtime's access time/heartbeat
+    * so the shared runtime is not reaped out from under the paired agent mid-edit.
     *
     * @param runtimeId the runtime worksheet id.
     * @param agentUser the agent (paired) user principal, used only for error reporting.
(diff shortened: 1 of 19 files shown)


larryliang-inetsoft:
Closing — discarded. The worksheet runtime-join is being restructured into a shared, sheet-agnostic "Wiz Sheet-Agent Foundation" (Option A) and handed to another implementer to build fresh in the inetsoft.web.wiz.pairing package. This branch's worksheet-specific first cut is superseded. The detailed TDD code/tests are preserved as a reference in the Foundation + worksheet Phase-1 plan docs.

PR #4026, opened 2026-06-16
Bug #75433, save embedded VS to repository when HOME bookmark is saved

jshobe-inetsoft:
## Summary

When the HOME bookmark is saved for a viewsheet that contains embedded VS (VIEWSHEET_ASSET) assemblies, the embedded VS's inner assemblies were not being persisted to their own repository entries. This caused HOME navigation to restore stale state (e.g., a previously cleared chart zoom would reappear after reopening the viewsheet).

## Root Cause

`Viewsheet.writeXML` skips writing inner assemblies when `isEmbedded()=true`:

```java
if(!isEmbedded() || snapshotExport) {
    writer.println("<assemblies>");
    ...
}
```

When HOME is saved, `engine.setViewsheet(vs0, vsEntry0, ...)` saves the outer VS via `writeXML`. Because embedded VS assemblies have their parent set (`isEmbedded()=true`), their inner charts are never written to the repository. On the next session open, the embedded VS is loaded from its own (unmodified) repository entry — retaining the old state.

## Fix

After saving the outer VS, also save each embedded VS to its own repository entry. Each embedded VS is cloned and its parent reference is cleared (`setViewsheet(null)`) so that `isEmbedded()=false` and `writeXML` serializes the inner assemblies. The method recurses to handle nested embedded VSes.

## Test Plan

- Open a viewsheet containing an embedded VS with a chart
- Zoom in on a chart inside the embedded VS
- Clear the zoom
- Save the HOME bookmark
- Close and reopen the viewsheet (new session)
- Navigate to HOME — chart should not be zoomed

Fixes Redmine #75433.

### DIFF
--- a/core/src/main/java/inetsoft/web/viewsheet/service/VSBookmarkService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/service/VSBookmarkService.java
@@ -138,6 +138,7 @@ public class VSBookmarkService implements ApplicationListener<ProcessBookmarkEve
 
          try {
             engine.setViewsheet(vs0, vsEntry0, principal, true, true);
+            saveEmbeddedViewsheets(vs0, engine, principal);
          }
          finally {
             if(rvs.isViewer()) {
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft:
Thanks for the thorough review. Here's how each point was addressed:

**Recursive call placement** — Fixed. The `saveEmbeddedViewsheets` recursion has been moved outside the `if(cloned != null)` guard (but still inside `if(embeddedEntry != null)`), so deeply-nested embedded viewsheets are always visited even if cloning the current level fails.

**Inner assemblies' parent reference** — No change needed. `Viewsheet.clone0()` already re-wires all cloned assemblies back to the cloned viewsheet (the loop at lines ~1490–1494 of `Viewsheet.java`). Since `cloneAssemblyList` calls `assembly.clone()` recursively on each assembly, this re-wiring applies at every nesting level. After `cloned.setViewsheet(null)`, `cloned.isEmbedded()` is `false` and its inner assemblies reference `cloned` — not the original `embeddedVs` — so they also correctly reflect the non-embedded state.

**`updateDependency` comment** — Added a brief inline comment explaining that `false` is appropriate here because saving embedded VS runtime state does not change the dependency graph.

**Design concern (shared repository entry)** — Intentional. The HOME bookmark is meant to capture the full runtime state of the viewsheet at the time of save, including embedded sub-viewsheets. This mirrors the behavior of the outer `engine.setViewsheet(vs0, vsEntry0, ..., true, true)` call for the parent. The shared-entry semantics are a pre-existing characteristic of the embedded VS feature and are not changed by this fix.

**Method signature indentation** — Fixed. The continuation line has been corrected from 40 spaces to 39 to align with the opening parenthesis.

jkidacki-inetsoft (approved):

PR #4027, opened 2026-06-16
Bug #75442 - Animate gantt milestone markers and labels on chart load

FigurePaper:
## Summary

Gantt charts are built from task bars (IntervalElement, rendered as
inetsoft-bar) plus a milestone marker (PointElement, rendered as
inetsoft-point). The inline-SVG entrance-animation dispatch chooses one
animation branch per chart and checks hasBarVO before hasPointVO, so a gantt
chart always took the bar branch and the milestone markers — and their value
labels — popped in instantly while the bars faded. This change fades the
milestone marker and its label in after the bars, and makes their hover-dim
behavior consistent.

## Changes

Backend (core)
- SVGSupport: add the ANIMATION_FLAG_GANTT hint flag and the
  ANNOTATION_POINT_LABEL annotation class.
- PointElement: add the HINT_GANTT_MILESTONE marker hint.
- GraphGenerator: set HINT_GANTT_MILESTONE on the gantt milestone PointElement.
- PointVO: tag the milestone value label with ANNOTATION_POINT_LABEL when the
  hint is present, mirroring how BarVO tags bar value labels.
- VGraphPair: append the gantt flag to the grow hint when the graph contains a
  milestone PointElement (new hasGanttMilestoneVO helper). Keying on the
  element rather than the chart type means the flag and the label tagging in
  PointVO derive from the same signal.

Injector (utils/inetsoft-xml-formats)
- SVGAnimationDOMInjector: when the gantt flag is set, fade the inetsoft-point
  markers and inetsoft-point-label labels in after the last bar, reusing the
  bar-fade keyframe at the same delay the bar value labels use.
- Hover CSS: hovering a milestone point now dims the other points and their
  labels; the point-label class is added to the point dim, transition, and
  cross-tile rules. The bar hover rule is unchanged, so bar hover never dims
  milestone markers or labels.

Frontend (portal)
- chart-inline-svg.directive: add inetsoft-point-label to the label lookup map
  so the hovered milestone point keeps its own label lit while the others dim.

## Behavior

- Milestone markers and value labels fade in together, after the task bars.
- Bar hover: milestone markers and labels are not dimmed (consistent — neither
  dims).
- Milestone-point hover: other milestone markers and labels dim; the hovered
  point keeps its marker and label fully visible.
- Combo bar+point charts and scatter charts are unaffected (no milestone hint,
  no point-label elements).

### DIFF
--- a/core/src/main/java/inetsoft/graph/element/PointElement.java
+++ b/core/src/main/java/inetsoft/graph/element/PointElement.java
@@ -39,8 +39,9 @@ import java.util.*;
 @TernClass(url = "#cshid=PointElement")
 public class PointElement extends StackableElement {
    /**
-    * Hint marking this point element as a gantt milestone marker, a string of true or false.
-    * Used to tag the milestone value label for the SVG entrance animation.
+    * Hint marking this point element as a gantt milestone marker; the value is {@code "true"}
+    * when set, absent otherwise. Used to tag the milestone value label for the SVG entrance
+    * animation.
     */
    public static final String HINT_GANTT_MILESTONE = "gantt_milestone";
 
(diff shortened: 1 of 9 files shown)


jshobe-inetsoft (approved):

PR #4028, opened 2026-06-16
Bug #75436, fix cross-org import bookmark conflict detection and resolution

jkidacki-inetsoft:
When a site admin imports a viewsheet into a different org, bookmark conflicts were not detected and resolutions were not applied because user org IDs were not updated to the target org in several places:

- cloneAssetEntry(String orgID, String name) updated the entry's orgID but never updated the embedded user's orgID, causing getVSBookmarkEntry to return null on the USER_SCOPE owner check. Fixed by creating a new IdentityID with the target orgID before the early return, consistent with the Organization overload.

- BookmarkConflict.user was set from the raw JAR userName (stale org) instead of the org-updated userID.convertToKey(), breaking resolution map lookups during import.

- ViewsheetAsset.parseContent built the resolution lookup key with the raw XML name instead of the org-updated identityID.convertToKey(), causing keepCurrent resolutions to never match and always import over the existing bookmark.


jshobe-inetsoft (approved):

PR #4029, opened 2026-06-16
Bug #75429, fix logical model attribute move-up/down corrupting attribute names and rendering duplicates

ngovus-inetsoft:
## Summary

- **Root cause**: `logical-model-attribute-editor` leaves `form.valueChanges` subscribed across an attribute switch due to the deferred `resetFormControl` (setTimeout) pattern. Angular's `[ngModel]` sync for `dataType` fires in the same change-detection cycle, triggering `apply()` while the form still holds the previous attribute's `name` value — writing it onto the new attribute object. This corrupts the attribute's name, causing `@for (track attribute.name)` to see a duplicate track key, rendering two copies of the same attribute and losing the other.
- **Secondary**: `moveAttributeUp`/`moveAttributeDown` in `element-tree-node` used `Tool.clone` (creating a new object reference) and searched `selectedEles` with the wrong key (`this.attrIndex = -1` for an entity node), so the selection index was never updated after a reorder.
- **Secondary**: `editingEle.attribute` in the property pane was never adjusted after attribute reordering, leaving the right-panel editor pointed at the wrong attribute.

## Test plan

- [ ] Open `Data Source / Examples / Orders / Data Model / Order Model`, expand the **Order** entity
- [ ] Select **Discount** and click Move Up — tree should show `[Discount, Date, Num, Paid]`, right panel stays on Discount
- [ ] Click **Date** — tree stays `[Discount, Date, Num, Paid]` (no duplicates), right panel shows Date's properties with no duplicate-name error
- [ ] Click Move Down on Discount — tree reverts to `[Date, Discount, Num, Paid]`, right panel follows
- [ ] Verify Move Up / Move Down on first and last attributes are disabled (buttons greyed out)
- [ ] Verify Save persists the new order


jshobe-inetsoft (approved):

PR #4034, opened 2026-06-17
Bug #75443, fix physical view drag-to-join failing after Angular upgrade

ngovus-inetsoft:
## Summary

- Angular's `@for` block (Angular 17+) reuses `join-node-graph` instances when `track graph.node.id` matches, so reused instances skip `ngAfterViewInit`
- When `physical-model-network-graph.ngOnChanges` fires (e.g. a second table is selected), it clears `this.nodes = {}` and calls `jsp.deleteEveryEndpoint()` — but only **newly-created** nodes re-register via `ngAfterViewInit`; reused nodes lose their endpoints permanently
- This causes `checkJoinCompatibility` to only see the single newly-registered node and show "There are no other tables to join with", preventing join creation via drag

## Fix

Added `ngOnChanges` to `JoinNodeGraphComponent`. When the `graph` input changes on a **reused** instance (`!isFirstChange()`), the component re-runs `endPointsInit()` and emits `onRegisterNode`, restoring its jsPlumb endpoints and re-inserting itself into the parent's node lookup map. New components are unaffected (guarded by `isFirstChange()`).

## Test plan

- [ ] Create a new data source and physical view
- [ ] Select two or more tables
- [ ] Drag from an endpoint on one table to another — join edit pane should open (no "no other tables" warning)
- [ ] Verify existing joins still display and are clickable after saving and reopening the physical view

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/web/projects/portal/src/app/portal/data/data-datasource-browser/datasources-database/common-components/join-node-graph/join-node-graph.component.ts
+++ b/web/projects/portal/src/app/portal/data/data-datasource-browser/datasources-database/common-components/join-node-graph/join-node-graph.component.ts
@@ -18,7 +18,7 @@
 import { HttpClient, HttpParams } from "@angular/common/http";
 import {
    AfterViewInit, Component, ElementRef, EventEmitter, HostBinding, HostListener, Input,
-   Optional, Output, ViewChild,
+   OnChanges, Optional, Output, SimpleChanges, ViewChild,
 } from "@angular/core";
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):
Approving based on a clean Claude automated review (the fix is correct and well-targeted; remaining items are minor and do not block merging) and all checks passing.

PR #4036, opened 2026-06-17
Bug #75434, fix spurious warning when saving another user's ALLSHARE bookmark

jkidacki-inetsoft:
When saving a non-readOnly ALLSHARE bookmark owned by a user that no longer exists in the security provider, the permission check in checkAddBookmark was incorrectly using the bookmark owner's bare SRPrincipal instead of the current logged-in user's principal, causing it to fail. Added a currentPrincipal parameter so permission and org-visibility checks always use the actual session user, while the bookmark owner identity is still used for storage. Also removed an erroneous removeBookmark call that was re-introduced during a merge, which could silently delete a bookmark if the subsequent add failed.

### DIFF
--- a/core/src/main/java/inetsoft/web/viewsheet/service/VSBookmarkService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/service/VSBookmarkService.java
@@ -228,11 +228,10 @@ public class VSBookmarkService implements ApplicationListener<ProcessBookmarkEve
       }
 
       if("add".equals(action)) {
-         rvs.removeBookmark(name, ownerId, false);
          //add the bookmark
          addBookmark(rvs, name, type, readOnly,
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (changes requested):
The Claude automated review approves this PR with no blocking issues, but I'm requesting changes because the **`build` check for the current commit (`6e20b2d`) was CANCELLED** — it started at 16:00:47Z and was cancelled at 16:46:04Z, so there is no successful build for this PR. Please re-run the build and ensure it passes; once it's green this can be merged.

jkidacki-inetsoft:
The build failing was unrelated to the changes. I reran the build and it succeeded now.

jshobe-inetsoft (approved):

PR #4038, opened 2026-06-17
Bug #75452 - Stagger line/stacked-line load animation per series

FigurePaper:
## Summary

On chart load, line and stacked-line charts animated all their lines at the same
time when the lines were produced by a color or group dimension over a single
measure. Area and stacked-area charts in the same situation already drew one series
at a time, staggered across the animation window. This change makes line charts
match that behavior.

## Root cause

In SVGAnimationDOMInjector.injectLineAnimationFromAnnotations, the per-series
stagger delay is staggerDelay(seriesIdx, numSeries). Line series were ranked by
data-series, which is the measure column index (ElementVO.getColIndex()). When one
measure is split by a color/group dimension, every line shares the same measure
column index, so all lines collapsed to a single rank, numSeries became 1, and every
line got a zero delay. Area fills already avoided this by ranking on data-color,
which is unique per series.

## Fix

Rank line series by a composite data-series + data-color key in first-seen DOM
order, mirroring how area fills are ranked. The data-color component distinguishes
color-dimension series, while the data-series component still separates distinct
measures that happen to share a color. Facet panels that repeat the same key dedup
to a stable rank. LineVO already stamps data-color per series, so no backend or
rendering change is needed, and the area animation path is untouched.

## Tests

Added tests for the color-dimension stagger (the fixed case), the multi-measure case
(regression guard), and a single-line chart (no delay). Also removed a parseIntAttr
overload whose only caller was the replaced ranking code.

### DIFF
--- a/utils/inetsoft-xml-formats/src/main/java/inetsoft/util/graphics/animation/SVGAnimationDOMInjector.java
+++ b/utils/inetsoft-xml-formats/src/main/java/inetsoft/util/graphics/animation/SVGAnimationDOMInjector.java
@@ -858,20 +858,18 @@ public class SVGAnimationDOMInjector {
                                                             List<Element> annotAreas,
                                                             Element svgRoot, Document doc)
    {
-      // Line rank: data-series column index → 0-based rank.
-      // getColIndex() starts at 1 in some charts; TreeSet + sequential rank normalizes to 0-based.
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
Approved. Claude's automated review confirms all three issues from the prior pass are resolved and the fix is correct and complete with no outstanding issues. All CI checks pass and there are no merge conflicts.

PR #4042, opened 2026-06-18
add portal unit cases

sarapeng86:
1.add some cases
2.correct some wrong cases


sarapeng86:
● Analysis: mockClear() is correct here — mockReset() would be a regression.

  The review concern is hypothetical and does not apply to these files. Here is why:

  Fact 1 — No test in either spec file overrides MODAL_MOCK.open's implementation.
  Both files import MODAL_MOCK from the shared helpers file and use it purely as a registered provider. Neither calls
  MODAL_MOCK.open.mockReturnValue(...) or MODAL_MOCK.open.mockImplementation(...) inside any it or beforeEach. There is
  nothing for a "bleed" to carry.

  Fact 2 — The existing mockImplementation is the mock's defining contract, not a per-test override.
  MODAL_MOCK.open was defined as:
  vi.fn().mockImplementation(() => ({
     result: EMPTY,
     componentInstance: { onCommit: new Subject<string>() },
     close: vi.fn(),
     dismiss: vi.fn(),
  }))
  This implementation serves a functional purpose: it creates a fresh Subject on every open() call so that
  ComponentTool's .subscribe(modal.componentInstance.onCommit, ...) never sees stale emissions from a prior test's
  modal. It is semantically equivalent to the initial return-value contract of the mock, set once at module-load time,
  not per test.

  Fact 3 — mockReset() would break tests.
  mockReset() clears all implementations, including the one above. After reset, MODAL_MOCK.open() returns undefined. Any
  component code path that calls NgbModal.open() and then accesses .result or .componentInstance on the return value
  would crash immediately. The risk spec's deleteDrill and removeAllParameters groups all call
  ComponentTool.showConfirmDialog, which internally routes through NgbModal.open. They would fail with TypeError: Cannot
  read properties of undefined (reading 'result') after a mockReset().

  Verdict: mockClear() is the correct choice. It resets call history between tests (so toHaveBeenCalledTimes counts are
  clean) while preserving the factory implementation that is required for the mock to function. mockReset() is
  appropriate only when the mock has no defining implementation to preserve — that is not the case here. No change is
  needed.

jshobe-inetsoft (approved):
Approved. Claude's automated review (Pass 5) confirms all issues across prior passes are resolved and the PR is ready to merge. All CI checks pass and there are no merge conflicts.

PR #4043, opened 2026-06-18
Bug #75453 - Remove orphaned user data on identity deletion

FigurePaper:
## Summary

Deleting a user already cascades their private viewsheets/worksheets, bookmarks, owned scheduled
tasks, dashboards, scripts/styles, ACL grants and sessions. It did not, however, clean up three
categories that were left orphaned:

- favoritesUser references to the user on shared assets
- the user's UserEnv preference file (sreeUserData/{name}_{orgID}.xml)
- the user's auto-save drafts

This PR removes those as part of deletion so no dangling references or files remain. It applies to
every deletion entry point (EM UI bulk delete and the public REST API), because both converge on
IdentityService.deleteIdentities -> syncIdentity.

## Changes

- UserEnv.removeUser(IdentityID): deletes the user's properties file from the data space and
evicts its cached entry.
- AutoSaveUtils.deleteUserAutoSaveFiles(IdentityID): deletes the user's auto-save drafts (active
and recycle bin), matching the owner encoded in the file name. The auto-save bucket is resolved
from the deleted user's organization, so a site admin deleting a user in another org still removes
that user's drafts. A new package-private getStorage(String orgId) overload keeps the bucket-key
format in a single place.
- IdentityService: in the user-delete branch of syncIdentity, calls UserEnv.removeUser and
AutoSaveUtils.deleteUserAutoSaveFiles per user; a new batched
removeUserFavorites(Collection<IdentityID>) runs once after the deleteIdentities loop. A user is
only recorded for favorites cleanup after its syncIdentity deletion succeeds, so favorites are
never stripped for a user whose deletion failed.

## Performance

The favorites cleanup must scan asset folders to find references, which is independent of any
single user. It is therefore hoisted out of the per-user path and run once per delete operation,
grouped by organization so each org's folders are scanned a single time. A single-user delete
still costs one scan; a bulk delete of N users costs one scan per org instead of N.

## Scope

The change only removes data belonging to the deleted user. It never deletes a shared asset (only
the user's entry in its favoritesUser list), never affects other users, and does not alter the
existing deletion cascade or any non-delete path. All cleanup steps are null-guarded and wrapped
in try/catch so a storage error logs at WARN and does not abort the deletion; auto-save deletes
are guarded per file so one failed delete does not skip the rest.

### DIFF
--- a/core/src/main/java/inetsoft/sree/UserEnv.java
+++ b/core/src/main/java/inetsoft/sree/UserEnv.java
@@ -478,6 +478,34 @@ public class UserEnv {
    }
 
    /**
+    * Remove a user's saved properties file. Called when a user is deleted so the
+    * sreeUserData/{name}_{orgID}.xml file is not left orphaned.
+    */
+   public static void removeUser(IdentityID userIdentity) {
+      if(userIdentity == null) {
+         return;
+      }
+
+      String userFile = getUserFile(userIdentity.convertToKey());
+      DataSpace space = DataSpace.getDataSpace();
+
+      try {
+         removeChangeListener(space, userFile);
+
+         if(space.exists(USER_DIR, userFile)) {
+            space.delete(USER_DIR, userFile);
+         }
+      }
+      catch(Exception e) {
+         LOG.warn("Failed to remove user properties for {}", userIdentity, e);
+      }
+
+      synchronized(propmap) {
+         propmap.keySet().removeIf(k -> userIdentity.equals(getName(k)));
+      }
+   }
+
+   /**
     * Get the user names containing a UserEnv file
     */
    public static Set<String> getUsersWithFile() {
--- a/core/src/main/java/inetsoft/sree/UserEnv.java
+++ b/core/src/main/java/inetsoft/sree/UserEnv.java
@@ -495,13 +495,13 @@ public class UserEnv {
          if(space.exists(USER_DIR, userFile)) {
             space.delete(USER_DIR, userFile);
          }
+
+         synchronized(propmap) {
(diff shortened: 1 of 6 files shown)


FigurePaper:
Re: removeUserFavorites / removeUserEMFavorites unit tests — deliberately not added.
They're private instance methods on IdentityService (a service with 16+ constructor dependencies)
that operate on IndexedStorage/KeyValueStorage; a true unit test would be brittle. This matches
the existing codebase convention — e.g. AutoSaveControllerTest explicitly notes that
Spring-context-dependent paths are "covered by E2E tests."

jshobe-inetsoft (changes requested):
Holding off on merge based on the completed Claude automated code review, which flags a remaining item to address before merge:

- **Missing test coverage (Medium):** `IdentityService.removeUserFavorites` and `removeUserEMFavorites` — the most complex new code in this PR — have no unit tests. Since user deletion is irreversible, these paths should be covered (empty input short-circuit, user key removed from correct folder entries, unmodified folders not re-written, storage failure logs WARN without throwing).

There is also a minor open item worth folding in:
- **`InterruptedException` swallowed (Minor):** in `removeUserEMFavorites`, the `catch(Exception e)` around `favorites.remove(...).get(...)` clears the interrupt flag without restoring it. Split out a dedicated `catch(InterruptedException e)` that calls `Thread.currentThread().interrupt()`.

The functional changes look solid and all checks pass — once test coverage is added for the IdentityService methods, this should be good to merge.

jshobe-inetsoft (approved):
Approved: Claude automated review is complete with no blocking issues, all checks pass, and the branch is mergeable.

PR #4045, opened 2026-06-18
Merge forward fix for Bug #75427, fix IndexOutOfBoundsException in schedule week distribution

jkidacki-inetsoft:
A monthly TimeCondition with dayOfWeek=0 (invalid/unset) caused getWeekDistribution to call data.get(-1) because the guard only skipped index < 0. Changed the guard to index <= 0 to also skip the zero case.


jshobe-inetsoft (approved):
Approving based on the completed Claude automated code review, which found no blocking issues. All checks pass and the branch has no merge conflicts.

PR #4047, opened 2026-06-18
Bug #75469 - Return branded 400 for unsupported viewsheet export output type

FigurePaper:
## Summary

Invalid `outtype` values on the viewsheet export endpoint
(`GET /export/viewsheet/**`) returned an HTTP 500 and logged a full stack
trace at ERROR. This change returns a branded HTTP 400 instead and logs a
single WARN that identifies the source of the bad request.

## Problem

`VSExportService.getFormatNumberFromExtension` threw a raw
`RuntimeException` for any unrecognized `outtype`. Because the export runs
through the controller with the request's referer header present (the
normal case when a user clicks an in-app hyperlink), the exception escaped
the handler and was logged at ERROR by the dispatcher servlet as an HTTP
500. The triggering requests come from viewsheet hyperlinks configured
with legacy report output types (`excel`, `data`, `xml`, `json`, `svg`,
`txt`, `powerpoint`), so real user actions were producing 500s and ERROR
log noise.

## Change

- `ExportController.exportViewsheet`: validate `outtype` before doing any
  work. If it is present but unsupported, log a WARN (requested path,
  user, referer), set a 400 status, and forward to the branded error
  page. The viewsheet is no longer opened for a doomed request.
- `VSExportService.isSupportedExportType`: new non-throwing check that
  delegates to the existing extension switch, keeping the supported-type
  list single-sourced. `getFormatNumberFromExtension` keeps its throwing
  contract as a backstop for the service path.
- `VSExportServiceTest`: unit coverage for supported extensions,
  case-insensitivity, the legacy invalid values, empty string, and null.

## Behavior

| | Before | After |
|---|---|---|
| Status | 500 (with referer) | 400 |
| Page | Dispatcher 500 / bare error | Branded error page |
| Log | ERROR + stack trace | Single WARN with path/user/referer |

Valid output types and the no-`outtype` path are unchanged.

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/vs/controller/ExportController.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/controller/ExportController.java
@@ -110,6 +110,17 @@ public class ExportController {
          path = Tool.byteDecode(path);
          int format = formatParam.orElse(FileFormatInfo.EXPORT_TYPE_PDF);
          String type = outtypeParam.orElse(null);
+
+         if(type != null && !VSExportService.isSupportedExportType(type)) {
+            LOG.warn("Rejected viewsheet export with unsupported output type \"{}\" for \"{}\" " +
+                        "(user={}, referer={})",
+                     type, path, principal == null ? null : principal.getName(),
+                     request.getHeader("referer"));
+            response.setStatus(HttpServletResponse.SC_BAD_REQUEST);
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):
Approving based on the completed Claude automated code review, which found no blocking issues. All checks pass and the branch has no merge conflicts.

PR #4051, opened 2026-06-19
75476, dim out text frame if text layout in use

milotalon:
Root Cause:
The text-frame binding control (text-field-mc) in the chart Aesthetic pane had no awareness of whether a text layout was active. Even when a layout was in use (driving the text content), the Text field still rendered fully bright and accepted drag-drops, edits, and field removal — there was no UI signal that the layout had taken over.

Fix:
Added a layoutActive input to text-field-mc and overrode its isEnabled() getter to return false when a layout is active. Because the component's dim styling (.disabled-grayout), drop acceptance (isDropPaneAccept()), and edit/remove gating (_isEditEnabled) all funnel through isEnabled()/_isEnabled, this single change makes the field dim out and become fully non-interactive (no drop highlight, no drop, no edit pencil, no field-chip removal) whenever a layout is in use. The parent aesthetic-pane binds [layoutActive]="isLayoutDesignerActive" — the same condition that lights the Layout Designer button's active indicator — on both the regular chart and relation-chart Text rows. The lock recomputes reactively, so clearing the layout re-enables the field automatically.

### DIFF
--- a/web/projects/portal/src/app/binding/editor/chart/aesthetic/aesthetic-pane.component.html
+++ b/web/projects/portal/src/app/binding/editor/chart/aesthetic/aesthetic-pane.component.html
@@ -78,7 +78,7 @@
     <div class="unhighlightable aesthetic-row-label">_#(Text)</div>
     <text-field-mc class="fill-width" [bindingModel]="bindingModel" [aggr]="currAggregate"
       [assemblyName]="assemblyName" [objectType]="objectType"
-      [grayedOutValues]="grayedOutValues"
+      [grayedOutValues]="grayedOutValues" [layoutActive]="isLayoutDesignerActive"
       [chartModel]="chartModel" (onUpdateData)="updateData($event)">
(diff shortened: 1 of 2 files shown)


milotalon:
Re: the `syncAllChartAggregateInfo` guard at `text-field-mc.component.ts:130` (`if(!this.isEditEnabled)` missing `()`) — leaving this one as-is. It's a genuine pre-existing bug, but the guard has been dead for the file's entire history, so the established behavior is to always sync `textField` across aggregates. Adding the parentheses would make the method short-circuit whenever `isEditEnabled()` is false, which changes behavior in the normal (non-layout) multi-aggregate path — unrelated to this PR's scope (#75476) and carries regression risk. Better handled as its own ticket rather than folded into this scoped UI fix.

jshobe-inetsoft (approved):
Approved: Claude automated review is complete with no blocking issues, all checks pass, and the branch is mergeable.

PR #4062, opened 2026-06-22
fix: preserve table aliases when rebuilding SQL query from model

NeverStopFMH:
Replace getTableNames() with getTableNameAliasMap() to carry alias→source mappings into createSQL(), and sync SelectTable alias back to AssetEntry in initQuerySelectedTables() so aliases are retained on query reload.

### DIFF
--- a/core/src/main/java/inetsoft/uql/jdbc/util/JDBCUtil.java
+++ b/core/src/main/java/inetsoft/uql/jdbc/util/JDBCUtil.java
@@ -1147,9 +1147,11 @@ public class JDBCUtil {
    }
 
    /**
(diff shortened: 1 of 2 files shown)


NeverStopFMH:
In Simple Query mode, tables do not support aliases. This PR changed that behavior by allowing table aliases parsed from the SQL string, enabling SQL generated by Claude that contains table aliases to execute correctly.

However, table aliases are not the only SQL feature unsupported in Simple Query mode. Other features, such as column aggregations, are also not supported. Therefore, SQL Query Tables created by Claude should always be created in Advanced mode.

With that in mind, this PR will be closed and reimplemented in the correct way.

PR #3700, opened 2026-05-15
Replace bootstrap look in Database Query dialog. Also fix split pane issues

swaker854:
All five tabs in Database Query dialog are updated. Split pane is used by multiple tabs. It's update requires some fixes to support other usage in the codebase

PR #3604, opened 2026-05-02
cleaned up consolidated portal theming

swaker854:
New PR with clean file list and fixed for build error and claude code review

PR #4072, opened 2026-06-23
Recover additional data tab and condition UI regressions(with a redesign) from 563ab05dd - recovery-74519-0621

swaker854:
Recovery (UI lost when 563ab05dd overwrote prior PRs)

Data tab (PR #3604): Restores 11 regressions across asset-description, datasources-datasource, datasources-datasource-editor, database-data-model-browser, database-vpm-browser, move-*-dialog components, driver-wizard, share-link-dialog, and select-worksheet-dialog. Also fixes PhysicalModelController to populate tableCount on schema folder nodes so counts show on initial load.

VPM condition dialog (PR #3700): Restores one-of-vpm-condition-editor and vpm-condition-item-pane BEM class structure lost in the bad merge.

Condition item pane: Restores condition-item-pane.component.html single-row BEM layout (condition-edit-row) that 563ab05dd had reverted to Bootstrap form-inline.

Condition editor components: Restores condition-list, binary-condition-editor, one-of-condition-editor, variable-editor, and expression-editor to their post-PR-#3700 BEM state.

Condition dialog redesign

simple-condition-pane: New ADD+REMOVE paradigm — built conditions accumulate in a tokenized clause list (WHERE/AND/OR chips, field token chip, × remove button per row); full condition editor embedded in "Add Condition" section below.

advanced-condition-pane / mv-condition-pane: Replaced boxed card rows with flat cond-section layout — border-top separator, section header with ghost Edit/Clear buttons at 26 px pill height, collapsible body.
_bootstrap-override.scss: Sets .modal-footer background to --inet-shell-surface-subtle per design spec.
Bug fixes

SQLQueryDialogService: Capture isSingleQueryMode() before calling setMashupMode() so SaveWorksheetCommand is sent, persisting the SINGLE_QUERY→MASHUP transition to disk. Previously the mode change was lost on server restart.
ws-pane.onCancel: Only close the worksheet when the SQL table has no prior commits (sqlEdited=false); existing worksheets stay open on cancel.

fixed-dropdown: Add viewport fallback when this.container is not set so dropdowns are never clipped at the screen edge.
condition-field-combo-list: Read item height from --inet-control-height-sm via getComputedStyle at runtime instead of the hardcoded 19.5 px constant (actual CSS height is 24 px), fixing virtual-scroll container clipping.
Test fix

one-of-condition-editor.component.spec.ts / .html: Remove stray delete_id class from the Modify button (button order changed in redesign made querySelector("button.delete_id") target Modify instead of Delete); tighten value-list selectors to div.value-list__element to account for the new empty-state placeholder div.

Test plan
 Verify data tab: asset detail panel, datasource editor, data model browser, VPM browser, move dialogs, driver wizard all render correctly
 Verify condition dialogs: simple-condition-pane ADD/REMOVE flow, advanced and MV panes collapse/expand correctly
 Verify single-query worksheet: SINGLE_QUERY→MASHUP transition persists after server restart; cancel on a new SQL table closes the sheet, cancel on an existing one does not
 Verify condition field combo dropdown renders all items without clipping
 npm run test:portal — 1253/1253 passing

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/ws/dialog/SQLQueryDialogService.java
+++ b/core/src/main/java/inetsoft/web/composer/ws/dialog/SQLQueryDialogService.java
@@ -195,6 +195,8 @@ public class SQLQueryDialogService {
          WorksheetEventUtil.refreshAssembly(rws, name, true, commandDispatcher, principal);
       }
 
+      boolean wasSingleQuery = ws.getWorksheetInfo().isSingleQueryMode();
+
       if(model.isMashUpData()) {
          ws.getWorksheetInfo().setMashupMode();
       }
@@ -207,7 +209,7 @@ public class SQLQueryDialogService {
       AssetEventUtil.refreshTableLastModified(ws, name, true);
       WorksheetEventUtil.refreshVariables(rws, wsEngine, name, commandDispatcher);
 
-      if(ws.getWorksheetInfo().isSingleQueryMode()) {
+      if(wasSingleQuery) {
          SaveWorksheetCommand command = new SaveWorksheetCommand();
          command.setClose(model.isCloseDialog());
          commandDispatcher.sendCommand(command);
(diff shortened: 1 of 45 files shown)


jshobe-inetsoft (approved):
All review issues resolved across three rounds. Approving for merge.

PR #4083, opened 2026-06-23
75474, format not unique when adding multiple data fields to text layout designer

milotalon:
Root Cause:
Two defects in the Text Layout Designer's per-field formatting, both only visible on the first (not-yet-committed) open of the designer:

1. Format key mismatch. The per-field Format panel matched a layout field to its backend ref by aggregated full name (e.g. "Sum(Sales)"). But a field added in the designer is built client-side with fullName = the bare column caption ("Sales"), not "Sum(Sales)" — the aggregated name is only assigned by the backend on a later model round-trip. So the format SET/GET could not match the newly added fields until a commit+reopen. (The first field appeared to work only because it was seeded from the existing scalar Text binding, which already carried the correct backend name and was served by the legacy scalar-text-field path.)

2. Pre-save dropped the added fields. onLayoutDesignerPreSave (which fires when the Format panel is opened) applied only target.textLayout, not target.textFields. So fields added on a fresh open lived only in the designer's working copy; the round-trip rebuilt an empty textLayoutFields on the backend, leaving no ref for the per-field format to write to — the color reverted immediately. Reopening worked only because OK had already pushed the fields into target.textFields.

Fix:
1. Key the designer's Format panel by the field INDEX instead of its (client-side, not-yet-aggregated) name: the designer emits "_layoutfield:<index>" and the host appends the multi-style aggregate name when applicable. FormatPainterService resolves the layout-field ref by index (chart-level textLayoutFields, or the named aggregate's list) in both the SET (writeToTextLayoutStubs) and GET (getChartRegionFormat) paths, claiming the key so it never leaks to the scalar fallback. Name-based matching is retained for the chart-region-click path. The index is stable end-to-end (textLayoutFields are rebuilt in model order; removals compact indices in lockstep).
2. onLayoutDesignerPreSave now also applies target.textFields = layoutDesignerTextFields (same as commit), so the backend builds textLayoutFields aligned with the designer's field indices and the per-field format SET/GET lands on real refs.

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/vs/controller/FormatPainterService.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/controller/FormatPainterService.java
@@ -2225,14 +2225,19 @@ public class FormatPainterService {
    // is stable: textLayoutFields are rebuilt in model order and removals compact indices in lockstep.
    private static final String LAYOUT_FIELD_PREFIX = "_layoutfield:";
 
-   private static boolean isLayoutFieldKey(String columnName) {
+   // Package-private for direct unit testing (FormatPainterServiceLayoutFieldTest).
+   static boolean isLayoutFieldKey(String columnName) {
       return columnName != null && columnName.startsWith(LAYOUT_FIELD_PREFIX);
    }
 
    // Resolve the ChartRef for an index key against the chart-level textLayoutFields, or against the
    // named aggregate's list when an aggregate full name is appended (multi-style). Returns null when
    // the key is malformed, the aggregate is not found, or the index is out of range.
-   private ChartRef resolveLayoutFieldByIndex(VSChartInfo chartInfo, String columnName) {
+   // static (no instance state) and package-private for direct unit testing
+   // (FormatPainterServiceLayoutFieldTest).
+   static ChartRef resolveLayoutFieldByIndex(VSChartInfo chartInfo, String columnName) {
+      // Re-check the prefix so this method is safe to call standalone (callers also guard); the
+      // duplicated check keeps it self-contained for tests that invoke it directly.
       if(!isLayoutFieldKey(columnName)) {
          return null;
       }
--- a/core/src/main/java/inetsoft/web/composer/vs/controller/FormatPainterService.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/controller/FormatPainterService.java
@@ -1638,6 +1638,27 @@ public class FormatPainterService {
                format = staticItems.isEmpty() ? new CompositeTextFormat()
                                               : buildFormatFromStaticItem(staticItems.get(0));
             }
+            else if(isLayoutFieldKey(columnName)) {
+               // Index-based layout-field key (Redmine #75474): read the field's format from the
+               // resolved layout ref, creating an empty one on first open so panel edits attach to
+               // the same ref the SET path (writeToTextLayoutStubs) writes. Never fall through to the
+               // scalar path, which would read a different place than the panel writes.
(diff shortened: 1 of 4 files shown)


milotalon:
Thanks for the review. Addressed in dc46cf9:

**Changes made**
- **Item 3 (no tests):** Added `FormatPainterServiceLayoutFieldTest` (6 tests) covering `isLayoutFieldKey` recognition and `resolveLayoutFieldByIndex` for valid indices (distinct ref per index — the crux of #75474), out-of-range, malformed, non-layout, and missing-aggregate keys. To enable focused testing, `isLayoutFieldKey`/`resolveLayoutFieldByIndex` are now package-private and `resolveLayoutFieldByIndex` is `static` (it holds no instance state).
- **Item 6 (magic string):** Extracted `"_layoutfield:"` into an exported module-level `LAYOUT_FIELD_PREFIX` in `text-layout-designer.component.ts`, imported and reused in `binding-editor.component.ts`. (The Java side keeps its own `LAYOUT_FIELD_PREFIX` constant since the value can't be shared across the language boundary; a comment notes they must match.)
- **Item 7 (long comment):** Trimmed the `openFieldFormat` comment to a 2-line WHY note.

**No change (with rationale)**
- **Item 1 (silent format loss on out-of-bounds index):** Intentional. Claiming the key prevents fallthrough to the scalar-text-field path, which would write to the *wrong* location; the null case only arises if a field is removed between pre-save and the format SET, which isn't a reachable user flow. No `Logger` exists in this class, so adding one for an unreachable branch isn't warranted.
- **Item 2 (lazy create-and-set during GET, no write lock):** This mirrors the pre-existing pattern already used in the same method for target labels, axis labels, and the legacy scalar branch; the write is idempotent, so matching the surrounding code is correct here.
- **Item 4 (`_layoutfield:` prefix collision):** Mirrors the existing, already-shipped `_static:` key convention in this file; a chart column literally named `_layoutfield:0` is implausible and carries the same accepted risk.
- **Item 5 (redundant guard in `resolveLayoutFieldByIndex`):** Kept intentionally so the method is safe to call standalone (the new unit test invokes it directly); added a brief comment noting this.

Verified: new test 6/6, related core tests pass (BUILD SUCCESS), web build clean.

milotalon:
Thanks for the follow-up. These are all minor; responses below — no code change this round (the substance was settled in dc46cf9), so I'm not pushing another commit.

- **Item 1 (SET/GET integration untested):** Agreed it's reasonable scope to leave. The helpers are unit-tested, and the integration wiring (`return true` short-circuit, GET lazy-create) is manually verified and user-confirmed in the running app. A full Spring-context controller test would be disproportionate here; happy to revisit if this area sees more churn.
- **Item 2 (empty aggregate suffix `_layoutfield:0:` → chart-level):** Not reachable from the emitter — `binding-editor.onLayoutFieldFormat` only appends `:aggrName` when it's non-null and non-empty. Treating an empty suffix as a chart-level lookup is harmless (and arguably the right default), so I'd rather not add a guard for an unconstructable input.
- **Item 3 (Java constant lacks a back-pointer comment):** Fair — accepted as a nit. I'll fold the bidirectional `// Must match LAYOUT_FIELD_PREFIX in text-layout-designer.component.ts` comment into a future commit rather than spin a CI cycle on a one-line comment now.
- **Item 4 (`assertSame` "why identity" comment):** Likewise accepted; will add the one-line note on the "same ref, not a copy" contract alongside Item 3.

jshobe-inetsoft (approved):
Claude's reviews are resolved — all issues addressed and the follow-up review is minimized as resolved. All checks pass.

PR #4092, opened 2026-06-24
fix 75504: register crosstab wheel listener outside Angular zone

bonnieshi43:
(wheel)="wheelScrollHandler($event)" used zone-patched addEventListener,
  causing every horizontal scroll event to trigger a global CD cycle.
  Changed to outOfZone (onWheel)="wheelScrollHandler($event)" to match
  the existing pattern used for vertical scroll.


jshobe-inetsoft (approved):
Claude's review is LGTM — fix is correct, passive:false is preserved, and it follows the established out-of-zone event pattern. All checks pass.

PR #4102, opened 2026-06-25
delete unles virtual-scroll.service.ts and clear all invalid import

bonnieshi43:


jshobe-inetsoft (approved):
Latest Claude review confirms all prior concerns resolved, no regressions, cleanup is correct and safe. Approving.

PR #4112, opened 2026-06-26
Bug #75512, skip ngDoCheck reflow when data tip position is unchanged

jshobe-inetsoft:
## Summary

`VSDataTipDirective.ngDoCheck()` was unconditionally reading `clientWidth`/`clientHeight` (forcing synchronous browser layout reflows) and calling `renderer.setStyle()` multiple times on every Angular change detection cycle while a data tip was visible, causing noticeable mousemove lag on dashboards.

## Root Cause

`VSDataTipDirective` implements `DoCheck`, so `ngDoCheck()` runs on every Angular CD cycle. The CD trigger is `DataTipService.showDataTip()` → `showHideDataTip.next()` → `VSObjectContainer.detectChanges()`, which fires approximately every 100 ms while the mouse moves over a chart. On each cycle, every `VSDataTipDirective` instance on the dashboard performed expensive DOM reads (`mainComponent.clientWidth`, `mainComponent.clientHeight`) and multiple `renderer.setStyle()` calls — even when the data tip position, name, alpha, and pop state were all unchanged.

## Fix

Added a 5-field position cache (`lastRenderedTipX`, `lastRenderedTipY`, `lastRenderedTipName`, `lastRenderedTipAlpha`, `lastRenderedPopShowing`) to `VSDataTipDirective`. At the start of the visible branch in `ngDoCheck()`, all five inputs are read from their services (lightweight property reads, no DOM access), and if none have changed since the last render, the method returns immediately — skipping all DOM reads and style writes. The cache is reset to `undefined` when the data tip hides so the next show always renders fresh.

Also eliminated a duplicate `hasPopUpComponentShowing()` call inside the render path by reusing the pre-read `popShowing` variable.

## Test Plan

- Open a dashboard with a chart that has a data tip configured
- Move the mouse over the chart and verify there is no mousemove lag
- Verify the data tip still appears, positions correctly, and hides correctly
- Verify alpha/opacity is applied correctly when configured
- Verify z-index adjusts correctly when a pop-up component is showing

Fixes Redmine #75512.

### DIFF
--- a/web/projects/portal/src/app/vsobjects/objects/data-tip/vs-data-tip.directive.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/data-tip/vs-data-tip.directive.ts
@@ -35,6 +35,11 @@ export class VSDataTipDirective implements DoCheck {
    private leaveListener: () => any;
    private inElement: Element = null;
    private mobileDevice = GuiTool.isMobileDevice();
+   private lastRenderedTipX: number | undefined;
+   private lastRenderedTipY: number | undefined;
+   private lastRenderedTipName: string | undefined;
+   private lastRenderedTipAlpha: number | undefined;
+   private lastRenderedPopShowing: boolean | undefined;
 
    constructor(private popService: PopComponentService,
                private dataTipService: DataTipService,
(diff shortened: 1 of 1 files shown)


jkidacki-inetsoft (approved):
Approving — Claude review confirms the fix is solid. The two noted items are explicitly non-blocking.

jkidacki-inetsoft:
The Claude code review approves this PR (no blocking issues — both items noted are explicitly non-blocking). However, there are two unresolved inline review threads from Claude that must be resolved before GitHub will allow the merge. Please resolve those threads and the PR can be merged.

PR #4113, opened 2026-06-26
Bug #75521 - Make XUtil partition cache lazy so the class loads without a Spring context

FigurePaper:
## Problem

On a database key-value + S3 blob backend, the offline shell restore commands
cannot bootstrap the database key-value engine. Building the engine loads XUtil,
whose static initializer eagerly creates a Spring-managed DataCache:

    private static DataCache<String, XPartition> partitionDataCache = new DataCache<>(5, 1000);

DataCache's constructor calls DataCacheSweeper.getInstance(), which resolves a
Spring bean. The database engine path forces XUtil to load during engine creation
(DBConnectionPool.createStandalonePool -> XUtil.applyProperties), before any
context is available, failing two ways:

- :setup restore (no context): ExceptionInInitializerError caused by
  ShutdownException "Spring application context is not available".
- local :connect (context mid-refresh): a circular bean dependency
  keyValueEngine -> XUtil -> dataCacheSweeper -> propertiesEngine, yielding
  BeanCurrentlyInCreationException.

The default mapdb engine never calls DBConnectionPool/XUtil.applyProperties, so
the issue is specific to the database key-value engine and went unnoticed.

## Fix

Defer the partition DataCache creation out of the static initializer using the
initialization-on-demand holder idiom, so loading XUtil no longer requires a live
Spring context. The cache is built on first call to getTables/getColumns, both of
which run only at runtime when the context is up. Cache semantics (size, timeout,
get/put) are unchanged.

## Testing

Added XUtilClassLoadTest, which loads XUtil through an isolated child-first
classloader with no application context and asserts the class initializer does
not throw. Verified the test fails (ExceptionInInitializerError / ShutdownException)
when an eager Spring-touching static field is reintroduced, and passes with the
holder fix. core module compiles cleanly.

This is one of two independent defects that block storage restore on the
database/S3 backend; it addresses the offline shell path. The live REST/EM restore
path is a separate enterprise-side issue.

### DIFF
--- a/core/src/main/java/inetsoft/uql/util/XUtil.java
+++ b/core/src/main/java/inetsoft/uql/util/XUtil.java
@@ -336,11 +336,11 @@ public final class XUtil {
       String partitionName = (partition.getBasePartition() != null ?
          partition.getBasePartition().getName() + "^^" + partition.getName() :
          partition.getName());
-      XPartition tempPartition = partitionDataCache.get(partitionName);
+      XPartition tempPartition = partitionDataCache().get(partitionName);
 
       if(tempPartition == null ) {
          tempPartition = partition.applyAutoAliases();
-         partitionDataCache.put(partitionName, tempPartition);
+         partitionDataCache().put(partitionName, tempPartition);
       }
 
       partition = tempPartition;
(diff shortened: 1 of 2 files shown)


jkidacki-inetsoft (approved):
Approving — Claude review was resolved by the PR author (minimized as RESOLVED), indicating no blocking issues.

PR #4116, opened 2026-06-26
Bug #75520 - Select storage transfer by backend type via shared factory

FigurePaper:
## Summary

Storage restore fails with a NullPointerException on a non-mapdb backend (database
key-value + S3 blob). The selection of which StorageTransfer implementation to use
was inlined only on the backup side, so the rule could not be reused on the restore
side. This PR extracts that rule into a shared factory so both backup and restore
select consistently.

mapdb is a node-local engine, so its contents are transferred through the cluster
(ClusterStorageTransfer). Every other engine is shared and is read from and written
to directly (DirectStorageTransfer). The cluster path fires the live storage
change-listeners during a bulk import; a listener resolves a parent AssetFolder that
has not been imported yet, gets null, and calls getEntries() on it. The direct path
writes straight to the engines and never fires those listeners, so it cannot hit
that NPE.

## Changes

- StorageTransfer — new static create(KeyValueEngine, BlobEngine) factory holding
the single source of truth for the mapdb-vs-shared selection.
- DataSpaceSettingsService.doBackup — now delegates to StorageTransfer.create(...)
instead of an inline ternary.

## Tests

- StorageTransferTest — asserts ClusterStorageTransfer for mapdb and
DirectStorageTransfer for database/s3.
- DirectStorageTransferRoundTripTest — exports key-value and blob content to a ZIP
and re-imports it into fresh engines, verifying values, blob digest, and exact byte
content survive the round trip.

Existing GeneralSettingsPageControllerTest continues to pass.

### DIFF
--- a/core/src/main/java/inetsoft/storage/StorageTransfer.java
+++ b/core/src/main/java/inetsoft/storage/StorageTransfer.java
@@ -17,6 +17,8 @@
  */
 package inetsoft.storage;
 
+import inetsoft.util.config.InetsoftConfig;
+
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.file.Path;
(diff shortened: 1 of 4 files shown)


FigurePaper:
Leaving the deleteRedundantBackupFiles() off-by-one out of this PR — it's pre-existing code unrelated to the transfer-selection change, so it belongs in its own fix rather than widening this PR's scope. Filing it as a separate follow-up.

FigurePaper:
Re: The pre-existing deleteRedundantBackupFiles() off-by-one: 
The <= here is intentional, not an off-by-one. deleteRedundantBackupFiles() runs
before the new backup is written, so it deletes down to backupCount - 1 existing
files and the about-to-be-written backup brings the total back to backupCount.
Changing it to < would leave backupCount + 1 files after the new one is written.

jkidacki-inetsoft (approved):
Approving — all Claude review comments were resolved by the PR author (minimized as RESOLVED), indicating no blocking issues.

PR #4120, opened 2026-06-26
Bug #75517, optimize hide axis action to avoid unnecessary data re-execution

jshobe-inetsoft:
## Summary

Hiding a chart axis on a viewsheet with multiple charts was taking ~8 seconds instead of being nearly instant.

## Root Cause

`VSChartAxesVisibilityService` was calling `editObjectProperty()` with a cloned `ChartVSAssemblyInfo`. Because `AxisDescriptor` objects live inside `VSChartInfo` (the `cinfo` field of `ChartVSAssemblyInfo`), the clone-and-set path triggered `copyInputDataInfo()`, which detected a `cinfo` change and returned `INPUT_DATA_CHANGED | BINDING_CHANGED`. This caused a full data re-execution (with sandbox write lock acquisition) for each chart on the viewsheet. With N charts, N serialized re-executions produced the observed delay.

## Fix

Replaced the heavy `editObjectProperty()` path with the same lightweight pattern used by `VSChartAxisResizeService`: modify the live assembly info directly (no clone), call `chartInfo.clearRuntime()`, and return `VSAssembly.VIEW_CHANGED` via the `processEvent(ToIntFunction)` overload. This triggers only a visual refresh without data re-execution, regardless of how many charts are on the viewsheet.

## Test Plan

- Open a viewsheet with multiple charts (5+)
- Right-click a chart axis and select Hide — should be nearly instant
- Verify the axis is hidden and other charts are unaffected
- Right-click and Show All Axes — should also be instant
- Verify undo/redo restores axis visibility correctly

Fixes Redmine #75517.

### DIFF
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/chart/VSChartAxesVisibilityService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/chart/VSChartAxesVisibilityService.java
@@ -23,11 +23,10 @@ import inetsoft.cluster.*;
 import inetsoft.graph.data.PairsDataSet;
 import inetsoft.report.composition.WorksheetEngine;
 import inetsoft.report.composition.graph.GraphTypeUtil;
+import inetsoft.uql.viewsheet.VSAssembly;
 import inetsoft.uql.viewsheet.VSDimensionRef;
 import inetsoft.uql.viewsheet.graph.*;
 import inetsoft.uql.viewsheet.internal.ChartVSAssemblyInfo;
-import inetsoft.util.Tool;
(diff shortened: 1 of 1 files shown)


jkidacki-inetsoft (approved):
Approving — Claude review confirms the fix is correct and ready to merge (dead variable and pre-existing typo are minor/non-blocking).

jkidacki-inetsoft:
The Claude code review approves this PR (the dead `chartDescriptor` variable and pre-existing typo are minor/non-blocking — the fix itself is correct and ready to merge). However, there is one unresolved inline review thread from Claude that must be resolved before GitHub will allow the merge. Please resolve that thread and the PR can be merged.

PR #4132, opened 2026-06-29
add portal unit cases

sarapeng86:


jkidacki-inetsoft (approved):
Approving — Claude review confirms the portal unit cases are correct and ready to merge.

PR #4134, opened 2026-06-29
Bug #75527, fix anonymous guest Log In redirecting to portal instead of login page

ngovus-inetsoft:
## Summary

- `AbstractLogoutFilter`: honored the `showLogin=true` parameter as an explicit login-page redirect even when `securityEnabled=false` — previously this flag was blocked by the `securityEnabled` guard and the anonymous user was always sent back to the portal instead of `login.html`
- `InvalidateSessionFilter`: skip the stale-principal check for `/logout` and `/sessionexpired` requests so `LogoutFilter` can still read the session principal to determine the correct redirect URL
- `LogoutFilter`: fixed `handleSessionExpired()` to call `logout()` before redirecting, properly releasing the license slot on session expiry

## Test plan

- [ ] As an anonymous (guest) user, click "Log In" from the portal — should redirect to `login.html` regardless of whether the security subsystem is formally enabled
- [ ] Log in as a real user, verify normal login/logout flow is unaffected
- [ ] Log out from Enterprise Manager, verify redirect goes to EM (not login page) when security is disabled
- [ ] Trigger session expiry, verify the license slot is released and redirect is correct

### DIFF
--- a/core/src/main/java/inetsoft/web/security/AbstractLogoutFilter.java
+++ b/core/src/main/java/inetsoft/web/security/AbstractLogoutFilter.java
@@ -75,9 +75,10 @@ public abstract class AbstractLogoutFilter extends AbstractSecurityFilter {
 
       boolean securityEnabled = getSecurityEngine().isSecurityEnabled();
 
-      // SSO need login in login page of sso.
-      // When security is not enabled, don't redirect to the login page — just go back to portal.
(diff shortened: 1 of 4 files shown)


jkidacki-inetsoft (approved):
Approving — Claude review was resolved by the PR author (minimized as RESOLVED), indicating no blocking issues.

PR #4137, opened 2026-06-29
Feature #72759 - Warn admins of scheduled-task impact when deleting a user or group

FigurePaper:
## Summary

When an admin removes a user or group from security, any scheduled tasks
that user owns are deleted, and any task where the user/group is set as
"Execute As" has its executor cleared. A task with no executor runs as
its owner, so in a multi-tenant deployment this can move a task into the
owner's organization and deliver data scoped to the wrong tenant. None
of this was surfaced to the admin before deletion.

This change adds an informational warning to the EM delete confirmation
dialog listing the scheduled tasks that will be deleted and the tasks
whose "Execute As" will be reset. The admin can review and still proceed.

## Backend

- ScheduleManager.getIdentityRemovalImpact(identity, provider): read-only
  computation of affected tasks (owned vs execute-as), mirroring the
  matching in identityRemoved so the two stay in sync. Resolves the org
  from the identity's own orgID.
- IdentityService.getDeleteTaskImpacts(...): aggregates per identity,
  de-dupes, and excludes execute-as tasks that are also owned (those are
  deleted outright). Runs each lookup inside OrganizationManager.runInOrgScope
  for the target identity's org so cross-org / self-org deletions resolve
  that org's task map and re-resolve the executor under the right context.
  Gated by the same per-identity ADMIN permission as the delete endpoint.
- RoleController: new POST
/api/em/security/user/delete-identities/{provider}/affected-tasks
  returning DeleteIdentitiesTaskImpactResponse { ownedTasks, executeAsTasks }.
- Two new catalog strings for the owned-tasks and execute-as warnings.

## Frontend (Enterprise Manager)

- users-settings-view delete() pre-fetches the impact and appends the
  affected tasks to the existing confirmation dialog. On any error it
  falls back to the plain confirm rather than blocking the delete, and it
  skips the pre-check entirely when only roles/organizations are selected.

## Behavior notes

- Applies to USER and GROUP deletions. Roles and organizations are not
  affected by the owner/execute-as cleanup and produce no task warning.
- Cross-org: the warning resolves the target identity's org, so a host/site
  admin deleting a user in another org sees that org's affected tasks. The
  executor relationship is re-resolved under the target org scope, which is
  required because task identities are looked up from the security provider
  under the current org context on load.
- Purely additive: new endpoint and response model, no change to existing
  APIs, serialized formats, or the deletion path.

## Testing

- ScheduleManagerTest: owner-based impact detection and the read-only
  (no-mutation) guarantee. The execute-as path is re-resolved from the
  security provider on load and is covered via the EM spec and manual
  multi-tenant testing rather than the provider-less unit fixture.
- users-settings-view spec: dialog lists affected tasks and emits delete
  on confirm; falls back and still deletes when the pre-check fails.


jkidacki-inetsoft (changes requested):
The Claude code review identifies several issues worth addressing before merge:

**1. `getIdentityRemovalImpact` is not `synchronized` but `identityRemoved` is**

A concurrent delete could modify the org task map while the preview is iterating it, causing a `ConcurrentModificationException`. Adding `synchronized` would be consistent with the rest of the class.

**2. `getIdentityRemovalImpact` incompletely mirrors `identityRemoved`**

`identityRemoved` also calls `updateNotifications` for each task action (removing the deleted user from notification email lists). These tasks are not surfaced in the warning dialog. The Javadoc comment `"Mirrors the task handling in identityRemoved"` is misleading. Consider either extending the response to include `notificationTasks`, or tightening the comment to clarify what is and isn't reported.

**3. Task names can collide across orgs in the de-duped response**

`ownedTasks` and `executeAsTasks` are de-duped purely by task name. In a host-admin cross-org bulk-delete, tasks with the same name in different orgs would appear only once in the dialog. Consider qualifying task names with their org when identities span more than one org.

**4. Copyright year in `delete-identities-task-impact.ts`**

The new TypeScript file has `Copyright (C) 2024` but this is a 2026 addition. The matching Java file correctly says 2026.

(Issues 3 and 6 from the review — pre-existing NPE in `identityRemoved` and the `IdentityTaskImpact` inner record API placement — are non-blocking.)

jkidacki-inetsoft (approved):
LGTM - all requested changes addressed, Claude review confirms. All checks pass.

PR #4139, opened 2026-06-29
fix(wiz): give plugin-created charts a meaningful title (not "Chart")

larryliang-inetsoft:
## Problem

Follow-up to #4138 (crosstab/table titles). Plugin-created **charts** defaulted to the generic **"Chart"** heading (`ChartVSAssemblyInfo`'s `new TitleInfo("Chart")`) — visible on e.g. a treemap.

## Fix

Derive the same `"<measures> by <dimensions>"` title for chart assemblies in `createViewsheetInternal`, set via `applyGridTitle` (sets the value only, not visibility — so a chart that doesn't render a title row is unaffected).

Gather refs the way `collectChartFlatBinding` does — `getBindingRefs(false)` (x/y/group) **plus the aesthetic fields** (color/shape/size/text) — because for some chart types (treemap, …) the measure lives on the **size aesthetic**, not in `getBindingRefs`. Classify by the base `VSDimensionRef` / `VSAggregateRef` types and de-dupe.

## Testing

Verified live (image `local-972`): an industry × account-type treemap now titles itself **"Sum(COUNT_accounts) by account_type, industry"** instead of "Chart". (A first attempt read only `getBindingRefs` and produced "account_type, industry" — missing the size measure — which is what motivated also collecting the aesthetic refs.)

With this + #4138, charts, crosstabs, and tables all derive a meaningful title uniformly.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
@@ -2079,8 +2079,19 @@ public class WizVsService {
 
    /** Set a derived display title on a titled assembly (crosstab/table); no-op on blank input. */
    private void applyGridTitle(VSAssembly assembly, String title) {
-      if(title != null && !title.isEmpty()
-         && assembly.getVSAssemblyInfo() instanceof inetsoft.uql.viewsheet.internal.TitledVSAssemblyInfo ti)
+      if(title == null || title.isEmpty()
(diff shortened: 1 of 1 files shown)


larryliang-inetsoft:
Added a second commit (95b5344eb): `applyGridTitle` now only fills StyleBI's **default placeholder** title ("Chart"/"Table"/blank) and never overrides a title that was already set to something meaningful (e.g. a chart the recommender titled). This scopes the derivation to genuinely-untitled assemblies (treemap, crosstab, table). Verified on local-973: a bar chart keeps its existing title; the treemap (placeholder "Chart") gets "Sum(COUNT_accounts) by account_type, industry".

PR #4143, opened 2026-06-30
feat(wiz): add worksheet table execution probe

NeverStopFMH:
Probe worksheet table execution inline during createTable, so render-time SQL and expression failures surface at table creation instead of at viewsheet creation.                                                            
                                                                                                                                                                                                                               
  For a render-time-risk table (a SQL query table or one with expression columns), after the structural build persists it, the probe runs it in LIVE_MODE via AssetQuerySandbox and forces the first row with moreRows(1) so   
  lazy expression columns actually execute, then reads the failed-query cause off the lens with checkFailedQuery(lens, false) (raw cause, without the expression-specific message).                                            
                                                                                                                                                                                                                               
  Any failure blocks: the table returns success=false with the real error and the same wsId (it stays persisted), so the caller can fix just that one table and rebuild it. No warning path, no infra-vs-query classification —
  the raw error goes straight to the caller. Nothing is exported and the worksheet is not mutated.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/WorksheetTableController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WorksheetTableController.java
@@ -109,6 +109,35 @@ public class WorksheetTableController {
       return worksheetTableService.getWorksheetModel(wsIdentifier, user);
    }
 
+   /**
+    * Probe whether an already-built worksheet table can actually execute — used to surface a query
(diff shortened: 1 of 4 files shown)


fredaan:
Why not run the table directly after creating the SQL table and return the result? Compared to individual detection, it is simpler

Nelson-cao (approved):

PR #4153, opened 2026-07-01
75549, Cannot read property 'setActionVisible' of undefined

milotalon:
Root Cause:
This was a regression from the Rhino → GraalJS scripting engine migration. The "viewsheet" object exposed to scripts is a ViewsheetScope, which resolves assembly references (e.g. Chart1) through its getMember() method. Under Rhino, "viewsheet" was a Scriptable, so viewsheet['Chart1'] routed through get(String) and returned the assembly scriptable.

Under GraalJS, the ViewsheetScope global was published to the script bindings as a raw host object (bypassing the ScopeProxy adapter). As a result, GraalJS resolved viewsheet['Chart1'] via Java HostAccess reflection — looking for a Java member literally named "Chart1", which does not exist — so it returned undefined. Calling .setActionVisible('Brush', false) on undefined then threw the TypeError. (Unqualified access such as Chart1.setActionVisible(...) still worked because it resolves through the scope chain rather than the "viewsheet" global, which is why only the viewsheet[...] / viewsheet.<name> form was broken.)

Fix:
Route script globals through the central ScriptValueConverter.toGuest() adapter when publishing them to the GraalJS bindings (in GraalJavaScriptEngine.put() and the initScope() vars loop), so a ScriptScope such as ViewsheetScope is wrapped as a ScopeProxy and member/bracket access is bridged to getMember()/hasMember() — restoring the original Rhino behavior. Made the conversion symmetric by unwrapping ScopeProxy/ArrayProxy back to the underlying scope in ScriptValueConverter.toHost(), so Java-side callers reading the global back (e.g. ViewsheetScope in CalcTableLens) continue to receive the real scope object. This also fixes the same latent issue for the other scope globals (worksheet, pviewsheet, etc.).

### DIFF
--- a/core/src/main/java/inetsoft/util/script/graal/GraalJavaScriptEngine.java
+++ b/core/src/main/java/inetsoft/util/script/graal/GraalJavaScriptEngine.java
@@ -122,7 +122,7 @@ public class GraalJavaScriptEngine implements AutoCloseable {
          Value bindings = context.getBindings("js");
 
          for(Map.Entry<String, Object> e : vars.entrySet()) {
-            bindings.putMember(e.getKey(), e.getValue());
+            bindings.putMember(e.getKey(), ScriptValueConverter.toGuest(e.getValue()));
          }
       }
 
@@ -730,7 +730,7 @@ public class GraalJavaScriptEngine implements AutoCloseable {
 
       try {
          if(context != null) {
-            context.getBindings("js").putMember(name, value);
+            context.getBindings("js").putMember(name, ScriptValueConverter.toGuest(value));
          }
       }
       finally {
--- a/core/src/main/java/inetsoft/util/script/graal/ScriptValueConverter.java
+++ b/core/src/main/java/inetsoft/util/script/graal/ScriptValueConverter.java
@@ -123,18 +123,16 @@ public final class ScriptValueConverter {
 
       // proxy/host wrappers and plain objects: hand back the raw value's
       // host object if present, else the Value itself for member access.
-      // Unwrap our own scope adapters back to the underlying ScriptScope so
-      // toHost is symmetric with toGuest (e.g. env.get("viewsheet") returns the
-      // real ViewsheetScope, not the ScopeProxy bridge).
+      // Unwrap a ScopeProxy back to the underlying ScriptScope so toHost is
+      // symmetric with toGuest (e.g. env.get("viewsheet") returns the real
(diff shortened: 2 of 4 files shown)


milotalon:
Thanks for the review. Addressed both points in 35c44d10d:

- **Regression tests added.** `GraalJavaScriptEngineScopeTest.publishedScopeGlobalResolvesBracketAndDotAccess` reproduces #75549 by publishing a `ScriptScope` global and asserting `viewsheet['Chart1']` / `viewsheet.Chart1` resolve (not `undefined`). `ScriptValueConverterTest.scopeProxyUnwrapsToUnderlyingScope` locks in the `toGuest`→`toHost` `ScopeProxy` round-trip that `CalcTableLens` relies on.
- **Style nit.** Switched `toHost` to Java 21 pattern-matching `instanceof`. While doing so I found the `ArrayProxy` unwrap branch was actually unreachable — `ArrayProxy` is a `ProxyArray`, so it's intercepted by the earlier `hasArrayElements()` branch and never reaches `isProxyObject()`. Removed it as dead code (only `ScriptScope`/`ScopeProxy` globals are ever read back). All 12 tests in the two touched classes pass.

jshobe-inetsoft (approved):
Approving based on Claude's code review (marked as resolved) which found no blocking issues. All checks pass.

jshobe-inetsoft (changes requested):
This PR has merge conflicts with the base branch that prevent it from being merged. Please rebase on main and resolve the conflicts.

milotalon:
Good catch — you're right, and my earlier "dead code" reasoning was circular: the `ArrayProxy` branch was only unreachable *because* of the reorder in that same commit. Fixed in b87f638c4:

- Restored the `isProxyObject()` check **above** `hasArrayElements()` and brought back the `ArrayProxy` unwrap branch, so an `ArrayProxy`-wrapped `ScriptArrayScope` (which is a `ProxyArray`) is unwrapped back to the live scope instead of being flattened into an `Object[]` copy — restoring `toGuest`/`toHost` symmetry for array-shaped scope globals.
- Kept the Java 21 pattern-matching `instanceof` style.
- Added `ScriptValueConverterTest.arrayScopeProxyUnwrapsToUnderlyingScope`, which forces an `ArrayProxy` via `toGuest`/`putMember`/`getMember` and asserts `toHost` returns the same instance (mirroring the existing `scopeProxyUnwrapsToUnderlyingScope`).

jshobe-inetsoft (approved):

PR #4155, opened 2026-07-02
unit cases for viewer

sarapeng86:


sarapeng86:
**Code Review (posted via API — GitHub does not allow submitting a formal review on your own PR, so this is a regular comment with the same content and verdict: request changes)**

## Summary

This is a large, pure test-migration PR: 60 files changed (+21,447/-1,963), converting legacy Angular `TestBed`-based `.spec.ts` unit tests to Angular Testing Library `.tl.spec.ts` tests across the chart, calendar, combo-box, image/text, range-slider/slider/spinner, and submit/tab/calctable/text-input/viewsheet component groups, plus a progress-tracker doc update. No production source files are touched, so there's no runtime/security/backward-compat risk — the review below is scoped entirely to test correctness and coverage.

All CI checks are green (`build`, `Semgrep Scan`, `Semgrep OSS`, `claude-review`) and the branch has no merge conflicts (`mergeable_state: clean`).

I reviewed every changed file against this repo's `.tl.spec.ts` test-quality checklist (specific-value assertions, per-branch `&&`/`||` coverage, mock lifecycle hygiene, no dropped legacy coverage) and cross-checked assertions against the actual production component source. Most of the PR is excellent — the chart (`ChartArea`/`ChartAxisArea`/`ChartLegendContainer`/`ChartPlotArea`), calendar, combo-box/mini-menu/vs-line (aside from one item below), and submit/tab/calctable/text-input/viewsheet groups all check out cleanly, including correctly preserving previously-identified real production bugs as documented `it.fails` blocks and porting forward a real-geometry integration test that the old `chart-plot-area.component.spec.ts` never actually ran (`it.skip`).

A few concrete issues turned up that are worth fixing before merge:

### Should fix

1. **Un-cleaned `document.cookie` override causes order-dependent test pollution** — `web/projects/portal/src/app/vsobjects/objects/output/text/vs-text.component.interaction.tl.spec.ts:1167`
   The `"should sanitize and post external urls..."` test does:
   ```ts
   Object.defineProperty(document, "cookie", { value: "XSRF-TOKEN=test-token", configurable: true });
   ```
   with no `try/finally` or reset. The very next test, `"should post a single url update to the iframe via processUpdateExternalUrlCommand"`, sets no cookie of its own but must assert `token: expect.any(String)` instead of an exact value — because it actually receives the leaked `"test-token"` from the prior test (production code reads it via `parseCookieValue(this.document.cookie, "XSRF-TOKEN")`). This is a real cross-test dependency masked by a vague matcher; running the second test in isolation would produce a different (probably empty) token and could fail. Please restore `document.cookie` after the first test (or scope the mock) and assert the second test's `token` value exactly.

2. **Dropped DOM regression coverage for range-slider** — `vs-range-slider.component.{display,interaction,risk}.tl.spec.ts`
   The deleted `vs-range-slider.component.spec.ts` had two real-DOM regression tests tied to specific bugs: **Bug #18972** ("should apply format on rangeslider" — asserts the rendered `div.vs-range-slider-body` picks up `text-decoration: underline` from `objectFormat`) and **Bug #20993** ("should display in correct position" — asserts actual rendered `top` styles across 5 elements: `.range-slider`, `.range-line`, `.current-value`, `.range-value.range-right`, `.range-value.range-left`). None of the three new range-slider spec files render a real fixture (no `TestBed`/`render`/`querySelector` anywhere in the directory) — the new `display` spec only asserts `comp.sliderTop` as a plain TS number, which verifies the computation but not that the template actually binds it (or the other 4 elements) correctly. This is exactly the class of regression those two bug-tests existed to catch.
   For comparison, the sibling `vs-slider.component.tl.spec.ts` (Group 11) explicitly keeps a real `TestBed`-rendered block "because these assertions are about real template output that cannot be observed on a directly-instantiated component" and ports all 3 legacy DOM tests verbatim. Please apply the same treatment to range-slider (or explicitly document why it's out of scope, the way the spinner spec does for its exclusions).

3. **Weak assertion contradicts its own comment** — `web/projects/portal/src/app/vsobjects/objects/slider/vs-slider.component.tl.spec.ts:295`
   ```ts
   // incCount = ceil((100-0)/20) = 5 → ticks at i=0..5 = 6 entries
   expect(comp.ticks.length).toBeGreaterThan(0);
   ```
   The comment computes the exact expected value (6, confirmed correct against `getTicks()`), but the assertion only checks `> 0`, so it would still pass if tick generation regressed to producing e.g. 2 or 20 ticks. Should be `toBe(6)`.

4. **Vague assertion doesn't verify the payload** — `web/projects/portal/src/app/vsobjects/objects/combo-box/vs-combo-box.component.interaction.tl.spec.ts:429-431`
   ```ts
   expect(socket.sendEvent).toHaveBeenCalledWith(
      "/events/vsview/object/model", expect.objectContaining({}));
   ```
   `expect.objectContaining({})` matches any object, so this only verifies the event URL, not that `GetVSObjectModelEvent` was built with the right assembly name. Every sibling assertion in the same file specifies concrete fields (e.g. the test right above uses `expect.objectContaining({ assemblyName: "Combo1", value: "B" })`) — this one should do the same (e.g. `expect.objectContaining({ name: "Combo1" })`).

### Nits (non-blocking)

- `vs-slider.component.tl.spec.ts:46` imports `TestUtils` but never uses it (the helper factories are used instead) — will trip lint on unused vars.
- `web/docs/superpowers/specs/test-prescan/viewer-prescan-2026-06-08.md`: three rows (`DateComparisonIntervalPaneComponent`, `HierarchyEditor`, `ChartPropertyDialog`) have their 状态 (status) column set to `✅ 推进`, which isn't one of the doc's own valid status values (`待审核`/`✅已测试`/`⏭已跳过`/`⏸️暂缓`) — looks like a copy-paste from the adjacent 建议 column. Pre-existing issue touched incidentally by this diff's reflow, not introduced by this PR's actual test work, but worth a follow-up fix since it silently falls out of both the "tested" and "pending" buckets in the progress header.

## Verdict

Requesting changes — mainly for item #2 (a real, named-bug regression-coverage gap) and item #1 (a latent flaky/order-dependent test). Items #3–#4 are small but should be quick fixes. Everything else in this large PR is solid work.

bonnieshi43 (approved):

PR #4157, opened 2026-07-02
Add viewer unitcase v0

bonnieshi43:


bonnieshi43 (commented):
## Summary

This PR is a large, welcome expansion of frontend test coverage: it converts ~30 `binding`/`vsview`/`ckeditor-wrapper` component test suites from `TestBed`-based `*.spec.ts` to `@testing-library/angular`-based `*.tl.spec.ts` tests, adds a new `viewerHandlers` MSW mock file, and fixes a small dead-code bug in `GeoMappingDialog.checkDuplicateMapping()` along the way. The internal test-prescan tracker (`viewer-prescan-2026-06-08.md`) is updated to reflect progress from 47/142 → 74/142 components tested, which is real progress if the tests are correct.

I ran the actual test suite locally (`ng run portal:test-tl`) against this branch to verify the new tests before reviewing further, since `test:portal:tl` is not currently wired into CI (`npm run test` / `npm run verify` only run `test:portal`, `test:em`, and `test:em:tl` — see `web/package.json`). That means the green "build" check on this PR does **not** exercise these new files at all.

## Blocking issues

### 1. A significant number of the newly-added test files fail when actually executed

Running `npx ng run portal:test-tl --watch=false` on this branch, **8 of the 11 failing test files in the whole TL suite are files added by this PR**, both when run in isolation and together:

| File | Result (isolated run) |
|---|---|
| `chart/field/chart-fieldmc.component.display.tl.spec.ts` | 7/45 failed |
| `chart/field/chart-fieldmc.component.interaction.tl.spec.ts` | 2/19 failed |
| `chart/aesthetic/categorical-color-pane.risk.tl.spec.ts` | 3/4 failed |
| `chart/aesthetic/color-field-mc.component.tl.spec.ts` | 3/10 failed |
| `chart/aesthetic/linear-color-pane.component.tl.spec.ts` | 3/9 failed |
| `chart/aesthetic/shape-field-mc.component.tl.spec.ts` | 6/11 failed |
| `chart/aesthetic/text-field-mc.component.tl.spec.ts` | 5/7 failed |
| `vsview/view/calc-table-cell.component.tl.spec.ts` | fails when run alongside the aesthetic suite (NG0201 `NullInjectorError` on `ChartFieldmc`) |

These are not flaky/order-dependent — most reproduce consistently in complete isolation. A few concrete root causes:

- `color-field-mc.component.tl.spec.ts` → `TypeError: editorService.getCustomChartFrames is not a function` — the mock editor service used in the test harness doesn't implement `getCustomChartFrames()`, which `CategoricalColorPane`'s constructor calls unconditionally when it's rendered as a child.
- `color-field-mc.component.tl.spec.ts` ("aggregate sync") → `fixture.componentInstance.ngOnChanges(null)` throws `TypeError: Cannot read properties of null (reading 'measure')` inside `getEditPaneId()` — calling the lifecycle hook directly with `null` isn't a valid `SimpleChanges` value and crashes downstream code.
- `chart-fieldmc.component.display.tl.spec.ts` "Group 13 — getDisplayLabel" asserts against `chartFieldComboLabel(container)` expecting a value derived from the private `getDisplayLabel()` method, but that method is dead code — it isn't called anywhere else in `chart-fieldmc.component.ts` or its template — so the assertion is checking something unrelated to what's rendered (`"state"` vs the expected `"State View"`).
- Multiple `chartTypeButton(container)` / `chartFieldEditIcon(container)` assertions return `null` (e.g. Group 9 "should allow editing cube measures on SQL Server", Group 12 binding/chart getters), meaning the test's setup doesn't actually cause the expected element to render.

Since `test:portal:tl` isn't wired into `npm run verify`/CI, none of this is caught by "Validation Build". Merging as-is means the freshly-added coverage (and the "✅已测试" marks in the prescan doc for `ChartFieldmc`, `ColorFieldMc`, `LinearColorPane`, `ShapeFieldMc`, `TextFieldMc`, `CategoricalColorPane`, `CalcTableCellComponent`) is misleading — those suites don't currently pass.

### 2. Unexplained `it.skip` added for CVA contract coverage

`ckeditor-wrapper.component.tl.spec.ts` has:

```ts
it.skip("should update disabled state via setDisabledState", async () => {
```

(added in the "skip case" commit, replacing a previously-enabled test). There's no comment explaining why it was skipped instead of fixed — it silently drops coverage of the `ControlValueAccessor.setDisabledState` contract while the prescan doc still marks `CkeditorWrapperComponent` as "✅已测试 (11 tests)", which overstates what's actually verified.

## Non-blocking observations

- **Good fix**: `GeoMappingDialog.checkDuplicateMapping()` (`geo-mapping-dialog.component.ts`) had dead code — it set `dupMapping = null` when there were no manual mappings, then unconditionally overwrote that assignment a few lines later regardless of duplicates found. The PR removes the dead early-null and instead makes the final assignment conditional (`Object.keys(dupMapping).length > 0 ? dupMapping : null`), which is a real, small correctness fix. It's covered by the new `geo-mapping-dialog.component.tl.spec.ts` ("should clear dupMapping when there are no manual mappings").
- `web/mocks/handlers/viewer.handlers.ts` is well-organized, each handler has a comment pointing at the calling site, and mirrors the existing project convention for MSW handlers.
- The large `*.spec.ts` → `*.tl.spec.ts` migrations that do pass look reasonable in structure (risk-tiered `describe` groups, comments noting what's explicitly out of scope per pass).

## Recommendation

Requesting changes because:
- Several newly-added test files fail when actually run (not caught today only because `test:portal:tl` isn't wired into CI) — see item 1.
- An unexplained `it.skip` silently removes coverage that the tracking doc still claims exists — see item 2.

Please fix (or remove/re-scope) the failing assertions in the `chart-fieldmc`, aesthetic-editor (`color-field-mc`, `linear-color-pane`, `shape-field-mc`, `text-field-mc`, `categorical-color-pane.risk`), and `calc-table-cell` TL specs, and either fix or document-and-track the skipped `ckeditor-wrapper` test, before merging.

jshobe-inetsoft (approved):

jshobe-inetsoft (dismissed):
This PR has merge conflicts with the base branch that prevent it from being merged. Please rebase on main and resolve the conflicts.

fredaan (approved):

PR #4161, opened 2026-07-04
feat(wiz): numericBin dimension directive → Range@ bin (faceted histogram support)

larryliang-inetsoft:
## What

Add a `numericBin` directive to the wiz chart-binding dimension model so an explicit `apply_binding` can request a numeric **range bin** (`Range@<field>`) on a dimension — the same auto-interval mechanism the lone-measure histogram (`HistogramChartFilter`) already uses.

This unblocks a **faceted binned histogram** ("distribution of a numeric measure split by a category") in the wiz-services layer, which previously wasn't expressible: the recommender only synthesizes the `Range@` bin when the measure is bound alone, and silently groups + `Sum`s once a real dimension is added.

## Changes (additive, recommender untouched)

- `DimensionFieldInfo`: new `numericBin` boolean (default false).
- `WizardRecommenderUtil.applyNumericBin(VSDimensionRef)`: static, idempotent — prefixes `groupColumnValue` with `Range@` unless already prefixed; no-op on blank base.
- Wired into both binding paths:
  - explicit `apply_binding` — `WizVsService.createVSChartDimensionRef` (beside the existing `dateGroupLevel`/`ranking` handling; also stamps the ref dataType from the field type).
  - recommender/fieldConfig — `WizAutoBindingService.applyFieldConfig` (right after `applyDateGroup`).
- Unit test `WizardRecommenderUtilTest` (3 cases: prefixes `Range@`; idempotent; no-op on blank).

No new Spring bean; no change to chart-type selection.

## Verification

Built into local image `local-975` and verified end-to-end via the wiz-services faceted-histogram plan-helper: `create_viewsheet(distribution, [price, category])` renders `x=[category, Range@price]`, `y=[Count(price)]`, one mini-histogram per category; saves and reopens cleanly (Range@ materializes on save). `WizardRecommenderUtilTest` 3/3.

Pairs with stylebi-wiz PR #1087 (the wiz-services plan-helper + routing that emits `numericBin`).

> Note: this branch was rebased onto `stylebi-wiz-main-rebase`; the commit was compiled/tested on its original base — worth a CI compile check against current mainline.

### DIFF
--- a/core/src/main/java/inetsoft/web/vswizard/recommender/WizardRecommenderUtil.java
+++ b/core/src/main/java/inetsoft/web/vswizard/recommender/WizardRecommenderUtil.java
@@ -841,6 +841,20 @@ public final class WizardRecommenderUtil {
       buildRangeDimension(dim, rvs, tname, interval);
    }
 
+   /**
+    * Convert a dimension ref into a numeric range-binned dimension using the {@code Range@<field>}
+    * shorthand — the same mechanism the lone-measure histogram uses (see HistogramChartFilter). The
+    * wizard executor auto-bins the numeric column (~8 nice-number bins) and createRangeDimension
+    * materializes it on save. Idempotent; no-op when the ref has no base column.
+    */
+   public static void applyNumericBin(VSDimensionRef ref) {
+      String base = ref.getGroupColumnValue();
+
+      if(base != null && !base.isEmpty() && !base.startsWith("Range@")) {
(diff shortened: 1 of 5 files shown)


larryliang-inetsoft:
🔗 **Companion PR (consumer):** inetsoft-technology/stylebi-wiz#1087 — the wiz-services plan-helper + routing that emits `numericBin` and turns this directive into a faceted binned histogram. This PR is the StyleBI-side dependency for #1087.

PR #4162, opened 2026-07-05
fix(wiz): crosstab binding-echo carries level-qualified date name

larryliang-inetsoft:
## Bug

A crosstab built through the explicit `apply_binding` path — one date column grouped at two levels (e.g. `date_start` as day-of-week on rows, hour-of-day on cols) — echoed its date dimensions back with **blank `fullName` and null `dateGroupLevel`**, and `slots.rows`/`cols` came back `[""]`.

Root cause: the crosstab design `VSDimensionRef` (built from the binding, no backing `ColumnRef`) has an empty `getVSName()`, so `VSDimensionRef.getFullName()` short-circuits to `""` before its date-qualifying branch. The wiz binding-echo (`collectDimensionFieldInfos` → `createDimensionFieldInfo`) copied that empty name. Downstream the facts pack keys dimensions by `fullName`, so it mis-read both dims and emitted a false "single value" caveat.

## Fix (recommender untouched)

- `WizFieldInfoFactory.createCrosstabDimensionFieldInfo`: derive the level-qualified name (`DayOfWeek(date_start)`) via `DateRangeRef.getName(groupColumnValue, dateLevel)` when `getFullName()` is empty/unqualified and `dateLevel != NONE`; also set the ref `dataType`.
- `WizVsService.collectDimensionFieldInfos` (crosstab-only echo): route through `createCrosstabDimensionFieldInfo` instead of the chart variant.
- `WizVsService.slotName`: fall back to the same helper for a plain `VSDimensionRef` when the primary lookup is empty (chart slots — `VSChartDimensionRef` with non-empty group value — are unaffected).

## Tests
`WizFieldInfoFactoryTest` + `BindingSlotsTest` (5 tests) — each fix independently RED-confirmed; the level-qualified name and populated crosstab slots are asserted. Non-date crosstab dims and chart slots are unchanged.

## Verification
Built into local image `local-979` and verified end-to-end via the wiz-services multi-date-level plan-helper: `create_viewsheet` binding `date_start` at two levels renders the weekday×hour crosstab with `binding.dimensions[].fullName` = `DayOfWeek(date_start)`/`HourOfDay(date_start)`, populated `slots`, and a correct facts pack.

Pairs with stylebi-wiz PR #1089 (the plan-helper that emits this crosstab binding).

> Rebased/based on `stylebi-wiz-main-rebase`; squashed from 3 dev commits.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WizFieldInfoFactory.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizFieldInfoFactory.java
@@ -19,6 +19,7 @@ package inetsoft.web.wiz.service;
 
 import inetsoft.uql.XCondition;
 import inetsoft.uql.XConstants;
+import inetsoft.uql.asset.DateRangeRef;
 import inetsoft.uql.schema.XSchema;
 import inetsoft.uql.viewsheet.VSAggregateRef;
 import inetsoft.uql.viewsheet.VSDimensionRef;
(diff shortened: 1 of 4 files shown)


larryliang-inetsoft:
🔗 **Companion PR (consumer):** inetsoft-technology/stylebi-wiz#1089 — the wiz-services `planMultiDateLevelCrosstab` that emits the two-date-level crosstab binding this PR's echo fix reports correctly. This is the StyleBI-side dependency for #1089's facts pack.

PR #4164, opened 2026-07-06
add viewer unit cases

sarapeng86:


jshobe-inetsoft (dismissed):
Merge conflicts must be resolved before the PR can be reviewed and merged.

sarapeng86 (commented):
## Code Review: #4164 "add viewer unit cases"

Reviewed all 24 changed files (10,367 additions / 75 deletions across `.tl.spec.ts` / `.test-helpers.ts` files for 10 components, plus a prescan-tracker status update). This PR adds no production code changes — it's test-only. Findings below are split by severity; overall this is a large, high-quality batch with a low defect density, but a few items should be fixed before merge.

### Merge gate status

- ✅ **CI checks**: all passing (`build`, `Semgrep OSS`, `Semgrep Scan`, `claude-review`).
- ⚠️ **Mergeability**: GitHub reports `mergeable_state: blocked`. There is an outstanding review from @jshobe-inetsoft (2026-07-08) requesting changes due to "merge conflicts." I re-verified with `git merge-tree` against current `main` and found **no actual conflicting hunks** — the branch appears to have been rebased/updated since that review was posted. However, the review itself is still open (`CHANGES_REQUESTED`) and branch protection is treating the PR as blocked until it's re-reviewed/dismissed. This should be resolved (ask @jshobe-inetsoft to re-review, or a maintainer to dismiss the stale review) independent of the findings below.

### Blocking — tests that don't actually verify what they claim

1. **`hierarchy-property-pane.component.tl.spec.ts:741-753`** ("should NOT remove a date column from the column list it was dragged from") — captures `sourceList: comp.localColumnList` *before* `comp.localColumnList` is reassigned to a new array, so the reference-equality guard being tested (`sourceList != this.localColumnList`) is never actually exercised in the intended state; the assertion happens to pass because the splice lands on an orphaned array instead. This would still pass if the "don't remove date columns" guard were deleted entirely.
2. **`hierarchy-property-pane.component.tl.spec.ts:757-773`** ("should shift the source index by one when reordering...") — only asserts that an element is present via `.some(...)`, which is true whether or not the `+1` shift correction exists (traced both the correct-shift and no-shift code paths — both leave the element present). Needs an exact `toEqual` on the full `dim.members` array to actually discriminate.
3. **`combo-box-editor.component.tl.spec.ts:680-684`** ("should compare down to seconds for TIME_INSTANT") — the example date pairs already differ at the hour level, so the test doesn't prove second/minute-level comparison. Tracing `DATETIME_PATTERN` in the component, the regex has no capturing groups for minutes/seconds at all — this test's name overclaims what's covered and may be masking a real precision gap in `validateDateRange`. Worth a follow-up look (either fix the regex/logic and test it properly, or capture the gap explicitly via `it.fails` with a WHY comment, per the project's own E7/E8 convention already used elsewhere in this PR).

### Minor

4. **`range-slider-data-pane.component.tl.spec.ts:264-271`** — asserts `sizeModel.rangeType` equals `TimeInfoType.NUMBER`, which is already the fixture's default before `selectColumn()` runs, so it doesn't prove the numeric-dispatch branch executed. Seed a non-default value first (as the adjacent TIME/date/cube cases in the same group correctly do).
5. **`vs-calc-table-layout.component.tl.spec.ts:691-704`** — doesn't set `comp.tableModel` before invoking `processGetTableLayoutCommand`, so the "carry over previous selection" recomputation path (`addSelectCells()`) degenerates into a no-op; assertions pass without exercising the real carry-over logic.
6. **B2 gaps** (guard branches missing a falsy-operand case) in a few spots — none individually severe, but worth a pass: `combo-box-editor` Group 8's 3-operand `&&` (middle operand alone false is untested), `hierarchy-property-pane` Group 9's triple-AND (only the all-true case is tested), `date-comparison-pane`'s `contextLevel != -1` operand (never exercised falsy), `input-parameter-dialog.display` `isFormInvalid()`'s `isInvalid() || name.invalid` (left-operand-alone-true case missing), `vs-calc-table-layout`'s `setSpanCellStyle` (width>1 and height>1 branches always tested together, never in isolation).
7. **`input-parameter-dialog.risk.tl.spec.ts:164-171`** duplicates `input-parameter-dialog.interaction.tl.spec.ts:74-83` (identical DATE-parse-before-`ngOnInit` scenario) without adding a distinct contract — should either be removed or changed to assert what the risk-spec header actually claims (form-guard behavior).
8. **`hierarchy-editor.component.tl.spec.ts`** — missing a case for an option-bearing member with a non-date/time `dataType` (e.g. STRING), a real branch where `selectedMember` gets assigned but `disableDate`/`disableTime` stay `true` since no `if/else-if` matches.
9. **`file-format-pane.component.interaction.tl.spec.ts:41-54`** — the `beforeEach` resetting `MessageDialog.lastMessage`/`lastMessageTS` is justified by a comment claiming the CSV dedup path exercises those statics, but every test on that path fully mocks `ComponentTool.showMessageDialog` via `vi.spyOn(...).mockResolvedValue(...)`, so the real dedup body (and those statics) never runs. Harmless dead setup, but the comment is factually wrong and could mislead a future maintainer.
10. Four files (`date-comparison-standard-periods.*.tl.spec.ts` ×2, `file-format-pane.*.tl.spec.ts` ×2) use the `.tl.spec.ts` suffix and direct instantiation with no `render()`/DOM queries/MSW at all. Per `CLAUDE.md`'s documented suite split, that pattern belongs in plain `.spec.ts`. It's called out explicitly in each file's header comment as a deliberate choice, so not blocking, but it's a real deviation from the stated convention worth a second look.

### Nitpicks

- `range-slider-edit-dialog.component.tl.spec.ts:20` — header comment contains Chinese text ("单元测试...+内存泄漏"); project convention (D4) is English-only comments/strings.
- `date-comparison-interval-pane.component.{display,interaction}.tl.spec.ts` headers label themselves "Pass 3" / "Pass 1" but no "Pass 2" file exists in the directory — coverage is complete and non-overlapping, just a confusing numbering gap for future maintainers.
- `date-comparison-interval-pane.component.test-helpers.ts:82` — `makeDynamicValue(value: any, ...)` uses a bare `any` (mirrors the real `DynamicValueModel.value: any`, so not really avoidable — low priority).
- A couple of test titles overclaim what they exercise: `range-slider-data-pane`'s "should keep column/table/folder/worksheet nodes" test only passes a `columnNode`; `input-parameter-dialog.display` reuses the exact title "should always run changeValidators for the current type" for two different trigger methods (`changeType` and `updateValue`).
- `combo-box-editor.component.tl.spec.ts` — `datePrompt`/`dropdownMinWidth` getters are never exercised and aren't called out as explicitly out of scope.
- `hierarchy-property-pane.component.tl.spec.ts:138` — a manual `mockRestore()` isn't wrapped in `try/finally` (harmless — the file-level `afterEach(vi.restoreAllMocks())` covers it, just inconsistent with the stated per-spy convention).

### What's genuinely well done

- Every real, pre-existing product bug surfaced during test-writing is correctly captured via `it.fails()` with an accurate WHY comment rather than silently asserted around — verified each against the actual source: `input-parameter-dialog` (3 bugs: `model` setter/`changeType()` cascade overwriting the just-set value twice over, and `ok()` crashing on stale `valueSource==="field"` with empty `fields`), `date-comparison-pane` (bitmask decoder misapplied to `contextLevel`), `date-comparison-standard-periods` (`toDateVisible`/`toDateLabel` null-guard asymmetry), `range-slider-data-pane` (`getParentFolder`'s unguarded `while` loop dropping the base class's null guard).
- The trickiest guard-branch traps called out in the source memory (the `-1==-1==-1` trivial-match trap in `date-comparison-dialog`, the chained `&&` second-operand gap) are both correctly handled with explicit tests and comments.
- i18n-interpolated getter expectations are computed via the real `Tool.formatCatalogString`/`DateTypeFormatter` utilities rather than hand-typed strings that assume interpolation happened — avoids a whole class of subtly-wrong fixtures.
- Mock/lifecycle hygiene (`afterEach(vi.restoreAllMocks())`, shared `test-helpers.ts` factories, no duplicate inline mocks across multi-pass splits) is consistently followed across all 10 components.

### Recommendation

Requesting changes — not because of major defects (the three "blocking" items are narrow, mechanical fixes to specific assertions, not systemic problems), but because: (a) the PR is currently `blocked` on GitHub pending resolution of the outstanding review, and (b) items 1–3 above are false-positive tests that should be tightened before merge so they actually guard the behavior they claim to.

sarapeng86:
"merge conflicts" has been resolved.

bonnieshi43 (approved):

fredaan (approved):

PR #4165, opened 2026-07-06
fixed 75567:Fix privilege escalation where ASSIGN implied full role admin access

bonnieshi43:
Fix ASSIGN grant incorrectly granting full access on security resources

DefaultCheckPermissionStrategy checked ResourceAction.ASSIGN instead of
ADMIN in the early-return block at L232-240, so users with only
"Assign Permissions" on a SECURITY_ROLE were allowed WRITE, DELETE,
and ADMIN. Change the grant lookup to ResourceAction.ADMIN to match
the comment and the rest of the cascade logic.


jshobe-inetsoft (approved):

PR #4166, opened 2026-07-06
chore: clarify review workflow instructions for Claude Code

sarapeng86:
Add explicit review guidance to keep analysis in a single workflow and avoid premature completion during PR reviews. This change does not alter tool permissions, but reinforces expected review behavior.

See https://github.com/inetsoft-technology/stylebi/pull/4164

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -58,5 +58,20 @@ jobs:
             markdown format. Format your comments as regular text or simple code
             blocks to avoid the "Apply suggestion" buttton.
 
+            IMPORTANT:
+            Do not use the Task tool or spawn sub-agents. Task usage is not
+            permitted and wastes review turns.
+
+            Review the PR yourself in a single workflow based on the PR diff.
+            Use Read/Grep when helpful.
+
+            Prioritize business logic, API changes, tests, and potential
+            regressions. Skip generated files, lock files, and clearly
+            non-functional formatting-only changes unless they may affect
+            behavior.
+
+            Do not mark the review as complete until all relevant files have
+            been analyzed.
+
           claude_args: |
             --max-turns 50 --allowedTools "mcp__github_inline_comment__create_inline_comment,Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*)"
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -51,10 +51,6 @@ jobs:
             * Check if any regression bugs may be introduced
             * Check if any backward incompatibility
             * Check if any test cases are broken by the changes
-            * Prioritize business logic, API changes, and potential regressions
-              over purely cosmetic changes
-            * Skip generated files, lock files, and non-functional
-              formatting-only changes unless they may affect behavior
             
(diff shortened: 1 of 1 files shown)


.github/workflows/claude-code-review.yml
```diff
@@ -58,5 +58,20 @@ jobs:
...
             blocks to avoid the "Apply suggestion" buttton.
 
+            IMPORTANT:
+            Do not use the Task tool or spawn sub-agents. Task usage is not
+            permitted and wastes review turns.
+
+            Review the PR yourself in a single workflow based on the PR diff.
+            Use Read/Grep when helpful.
+
+            Prioritize business logic, API changes, tests, and potential
```
jshobe-inetsoft: The first sentence of this is mostly redundant with the list of focus areas above. Just add anything you feel is missing from the above list and remove the first sentence.

jshobe-inetsoft (changes requested):

.github/workflows/claude-code-review.yml
```diff
@@ -51,6 +51,10 @@ jobs:
             * Check if any regression bugs may be introduced
             * Check if any backward incompatibility
             * Check if any test cases are broken by the changes
+            * Prioritize business logic, API changes, and potential regressions
+              over purely cosmetic changes
+            * Skip generated files, lock files, and non-functional
```
jshobe-inetsoft: Sorry, maybe I wasn't clear. This should item should have been kept where it was. This is a constraint, not an area of focus.

The item above goes here though.

.github/workflows/claude-code-review.yml
```diff
@@ -62,13 +66,7 @@ jobs:
...
             permitted and wastes review turns.
 
-            Review the PR yourself in a single workflow based on the PR diff.
-            Use Read/Grep when helpful.
-
-            Prioritize business logic, API changes, tests, and potential
-            regressions. Skip generated files, lock files, and clearly
-            non-functional formatting-only changes unless they may affect
-            behavior.
+            Use available tools such as Read/Grep when helpful.
```
jshobe-inetsoft, line 65: My previous review did not intend for all of this to be removed. They may be useful constraints, I'm not sure what behavior you were trying to address with this change.

"Prioritize business logic, API changes, tests, and potential regressions." Is the part I wanted changed. These are areas of focus and largely redundant with the list above.

What was the incorrect behavior that you were trying to address?

jshobe-inetsoft (changes requested):

.github/workflows/claude-code-review.yml
sarapeng86: The main behavior I was trying to address was that Claude Code Review could finish the review workflow before the actual review was completed. You can check the review job in https://github.com/inetsoft-technology/stylebi/pull/4164

In some cases, Claude started using sub-agents/Task during the review process. The workflow appeared to complete successfully, but not all changed files had been fully reviewed and some expected feedback was missing.

The additional constraints were intended to keep the review in a single workflow, avoid wasting review turns on spawning agents, and ensure the review is completed before marking it as done.

I agree that some of the added focus items were redundant with the existing review areas, so I will keep only the necessary constraints and remove duplicated focus descriptions.

jshobe-inetsoft (approved):

PR #4179, opened 2026-07-06
Fix #75551: cache negative class lookups in LegacyJavaShim

ngovus-inetsoft:
## Summary
- `LegacyJavaShim.tryLoad` scans the entire classpath on every failed `Class.forName` probe (the common case for non-class script identifiers during package navigation, e.g. `java`, `Tool`) and never cached the miss, so repeated navigation of the same name re-ran the full scan every time.
- Adds a bounded LRU negative cache (`NEGATIVE_CACHE`, capped at 10,000 entries) so a miss is remembered and short-circuits future probes for the same name, without the unbounded-growth risk the original code's comment was guarding against.

Verified via JFR profiling of the Composer condition-dialog "Browse Data" action: classloading/classpath-scan time for the repro dropped from the dominant cost (~57-76% of on-CPU samples across profiling runs) to effectively zero after the fix, with no change in behavior for legitimate class/package resolution.

## Test plan
- [x] `./mvnw install -pl core -DskipTests` builds cleanly
- [x] `LegacyJavaShimTest` passes (11 tests, 0 failures)
- [x] Re-profiled the original repro with JFR before/after — classpath-scan hotspot eliminated

### DIFF
--- a/core/src/main/java/inetsoft/util/script/graal/LegacyJavaShim.java
+++ b/core/src/main/java/inetsoft/util/script/graal/LegacyJavaShim.java
@@ -17,6 +17,7 @@
  */
 package inetsoft.util.script.graal;
 
+import com.github.benmanes.caffeine.cache.Caffeine;
 import inetsoft.sree.SreeEnv;
 import org.graalvm.polyglot.Context;
 import org.graalvm.polyglot.Value;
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):
Approving based on Claude's code review which found no blocking issues and recommends merge. All checks pass and there are no merge conflicts.

PR #4180, opened 2026-07-06
Bug #75545, fix 401 instead of 404 when getting a non-existent role

jkidacki-inetsoft:
## Summary
- `Get Role` (`/api/public/security/roles/{role}`) returned `401 Permission denied to get user` instead of `404` when queried with a role name that doesn't exist.
- Root cause: `DefaultCheckPermissionStrategy.checkOrgAdminPermission()`'s `SECURITY_ROLE` branch unconditionally returned `false` when the target role wasn't found, so the permission check failed *before* `SecurityApiService.getRole()` ever reached its "not found" check. The analogous `SECURITY_USER` branch already had this fix (Bug #66393) — falling back to an org-ID comparison when the identity doesn't exist — but it was never applied to roles.
- Fix mirrors the existing `SECURITY_USER` handling: when the role doesn't exist, grant permission if the resource's orgID matches the admin's own org (so a missing role in their own org surfaces 404), while still denying non-existent roles in other orgs (no cross-org info leak).

## Test plan
- [x] Added `orgAdminCanCheckNonExistentRoleInOwnOrg` — reproduces the bug (fails before the fix, passes after)
- [x] Added `orgAdminCannotCheckNonExistentRoleInOtherOrg` — confirms the org isolation boundary is preserved
- [x] `./mvnw -pl core test -Dtest=DefaultCheckPermissionStrategyTest` — all 32 tests pass

### DIFF
--- a/core/src/main/java/inetsoft/sree/security/DefaultCheckPermissionStrategy.java
+++ b/core/src/main/java/inetsoft/sree/security/DefaultCheckPermissionStrategy.java
@@ -570,8 +570,9 @@ public class DefaultCheckPermissionStrategy implements CheckPermissionStrategy {
 
          Role role = currProvider.getRole(resourceID);
 
+         // role doesn't exist, so just equals org name (see Bug #66393 for SECURITY_USER).
          if(role == null) {
-            return false;
+            return Objects.equals(resourceID.getOrgID(), orgID);
          }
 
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
Approving based on Claude's code review which found no blocking issues and recommends merge. All checks pass and there are no merge conflicts.

PR #4186, opened 2026-07-07
Fix Rank filter having no effect on Tree/Network chart source/target fields

ngovus-inetsoft:
## Summary
- `AbstractChartInfo.getRTFields()` never includes a Relation/Tree chart's runtime source/target fields, so `ChartVSAQuery.execute0()` never read any `RankingCondition` set on them when merging ranking into the query — Top N/Bottom N ranking on the Source or Target field of a Tree/Network chart was a silent no-op.
- Explicitly merge the `RankingCondition` from the tree chart's runtime source/target fields in `ChartVSAQuery`, falling back to the group already registered on the `AggregateInfo` (tolerating its table-qualified name, e.g. `Query1.State` vs `State`) since that field's column isn't synced into the column selection until later in the method.

## Test plan
- [x] Reproduced with a Tree chart bound Source=Region, Target=State, with a Top N / Bottom N rank filter on Target — confirmed ranking had zero effect before the fix (all rows shown regardless of N or Top/Bottom).
- [x] Applied fix, rebuilt, restarted server, reproduced same scenario — Top N=3 now correctly yields USA East = {NJ, NY, MA} and USA West = {CA, AZ, TX}, matching the expected per-region top-3-by-discount computed from the raw data.
- [x] Verified `git diff` is scoped to `ChartVSAQuery.java` only, no debug logging left behind.

Fixes bug-75579.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/ChartVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/ChartVSAQuery.java
@@ -1557,6 +1557,88 @@ public class ChartVSAQuery extends CubeVSAQuery implements BindableVSAQuery {
                }
             }
          }
+
+         // Relation/Tree chart source and target fields are never included in
+         // cinfo.getRTFields() (AbstractChartInfo.getRTFields() only assembles x/y, group/path, and aesthetic refs)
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):

PR #4187, opened 2026-07-07
Bug #75574: org-scoped role admin permission leaks onto global roles

jkidacki-inetsoft:
## Summary
An org-scoped SECURITY_ROLE admin delegation (via org-admin role, `SECURITY_ORGANIZATION` admin permission, or a direct grant on the "Organization Roles" node in Security > Users) was not scoped to the granting org, letting an org's delegate manage global/org-less roles such as the built-in `Administrator` and `Organization Administrator` roles — including reading full role details (e.g. `isSysAdmin: true`) via a direct REST call to `/api/em/security/providers/{provider}/roles/{role}/`, even though the EM UI correctly hides these roles from that user.

Four independent paths in `DefaultCheckPermissionStrategy.checkPermission()` had this gap:
- The "Organization Roles" node direct-grant fallback didn't check that the target role actually belongs to the granting org.
- The org-admin `SECURITY_ORGANIZATION`-permission fallback had the same gap, and additionally allowed *any* org-less role through unconditionally.
- `checkOrgAdminPermission()`'s `SECURITY_ROLE` case explicitly allowed org-less roles through via a `role.getOrganizationID() == null` disjunct, gated only by `isSystemAdministratorRole` — so an org-admin-flagged-but-not-sysadmin-flagged role (`Organization Administrator`) slipped past.
- The private ADMIN-cumulative `getPermission()` helper unconditionally merged grants from *both* the `"Organization Roles"` and `"Roles"` (global) permission roots into one synthetic permission, regardless of which root the checked role actually belongs to.

Each path now requires the checked role's own organization to match the granting org (or excludes org-less roles from cross-root merges entirely), mirroring the guard already used correctly elsewhere in the same file (lines ~135-154).

## Test plan
- [x] Added 4 regression tests to `DefaultCheckPermissionStrategyTest`, one per fixed path — each verified to fail against the pre-fix code and pass after the fix (confirmed by temporarily reverting each fix independently).
- [x] Full `inetsoft.sree.security.**` + `inetsoft.web.admin.security.**` test packages: 2622 tests, 0 failures, 0 errors.

### DIFF
--- a/core/src/main/java/inetsoft/sree/security/DefaultCheckPermissionStrategy.java
+++ b/core/src/main/java/inetsoft/sree/security/DefaultCheckPermissionStrategy.java
@@ -610,7 +610,7 @@ public class DefaultCheckPermissionStrategy implements CheckPermissionStrategy {
 
          // org admin permission never extends to global (org-less) roles — e.g. the built-in
          // Administrator and Organization Administrator roles — only to roles owned by this org
-         return !isSiteAdmin && Tool.equals(orgID, currProvider.getRole(resourceID).getOrganizationID());
+         return !isSiteAdmin && Tool.equals(orgID, role.getOrganizationID());
       case SECURITY_ORGANIZATION:
          if(resource.equals("*")) {
             return false;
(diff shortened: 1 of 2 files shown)


jkidacki-inetsoft:
Addressed both findings in cd34fc5:

1. **Cross-org escalation in the ADMIN-cumulative helper** (confirmed legitimate) — the `getOrganizationID() != null` check only verified the target role had *some* org, not that it matched the current org context, so an ADMIN grant on org A's "Organization Roles" node could leak onto a specific role belonging to a different org B. Fixed per the suggested approach: now requires the role's org to equal the current org before merging "Organization Roles" grants, and applies no merge at all when the role belongs to a different, non-null org. Added a regression test (`organizationRolesRootGrantDoesNotLeakOntoDifferentOrgsRole`) — confirmed it fails against the prior code and passes after the fix.
2. **Redundant `getRole()` call** (confirmed, minor) — cleaned up to reuse the already-fetched `role` local variable.

Full `inetsoft.sree.security.**` + `inetsoft.web.admin.security.**` suites: 2623 tests, 0 failures, 0 errors.

jshobe-inetsoft (approved):

PR #4193, opened 2026-07-07
75561, expose legend round corner option in script.

milotalon:
Root Cause:
Feature #74520 added the "rounded corners" option to chart legends, backed by a new roundCorners flag on LegendsDescriptor (isRoundCorners()/setRoundCorners()). The UI, XML persistence, and rendering were wired up, but no chart scripting property was registered for it, so it could not be controlled from scripts.

There was also a secondary defect specific to Composer Preview: even after exposing the property, the script value was ignored in Preview (though it worked in the deployed viewer/portal). In VSChartAreasService the chart descriptor is captured before the graph is generated; on the first render (Composer Preview) the runtime descriptor is still null at that point, so the design-time descriptor was captured. Graph generation then runs the script and populates the runtime descriptor, but GraphBuilder had already been handed the stale design descriptor, so the legend model received the design-time value instead of the script-applied value. In the viewer the graph had already been generated on a prior request, so the runtime descriptor was non-null at capture time and the correct value was used.

Fix:
1. Registered a new script property legendRoundCorners in ChartProcessor.addProperties() mapping to LegendsDescriptor.isRoundCorners()/setRoundCorners() (covers both report and viewsheet charts, consistent with the existing legendBorder property). Added corresponding entries in binding/property-type.properties (bool) and binding/assembly.properties so the property appears in the script editor's autocomplete.
2. In VSChartAreasService, after graph generation, GraphBuilder is now constructed with the runtime chart descriptor when available (falling back to the design descriptor), matching the getRTChartDescriptor() == null ? getChartDescriptor() : getRTChartDescriptor() idiom used elsewhere. This ensures script-applied legend values (e.g. legendRoundCorners) are reflected in Composer Preview as well as the viewer.

Usage example:
Chart1.legendRoundCorners = false;

### DIFF
--- a/core/src/main/java/inetsoft/report/script/ChartProcessor.java
+++ b/core/src/main/java/inetsoft/report/script/ChartProcessor.java
@@ -297,6 +297,8 @@ public class ChartProcessor {
          "setBorderColor", Color.class, LegendsDescriptor.class, legends);
       scriptable.addProperty("legendBorder", "getBorder", "setBorder",
          int.class, LegendsDescriptor.class, legends);
+      scriptable.addProperty("legendRoundCorners", "isRoundCorners",
+         "setRoundCorners", boolean.class, LegendsDescriptor.class, legends);
 
       // **************************************************************************
       // BC, the following properties are for backward compatibility (11.4) title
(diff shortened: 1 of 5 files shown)


jshobe-inetsoft (approved):

PR #4196, opened 2026-07-08
fix unit case bug

sarapeng86:


jshobe-inetsoft (approved):
Approved: Claude's automated review confirms the subscription/timer memory-leak fixes (#75597/#75598/#75599) are correct, minimal, and consistently applied, with no blocking issues, and all checks pass.

PR #4197, opened 2026-07-08
fixed 75603 and 75604, update test case.

bonnieshi43:
fixed below two issue:
**Bug #75603**

The REST endpoints that create, edit, and delete device profiles (used for responsive/mobile viewsheet layouts) perform no server-side authorization check at all. Any authenticated user — including users with no administrative privileges — can create, modify, or delete device profiles by calling these endpoints directly, regardless of the Edit Mobile Devices (ResourceType.DEVICE) permission configured in Security Actions.

**Bug #75604**
In multi-tenant mode, the "Time Range" option on a schedule task's time condition is hidden from Organization Administrators in the UI, but the server-side save path does not independently enforce this restriction. An org admin can attach a time-range condition to a schedule task by sending a crafted save request directly, bypassing the UI restriction entirely.

Note: this is distinct from — and should not be confused with — the EM "Schedule → Settings" page that defines named time ranges (start/end times). That page is correctly protected (EM_COMPONENT:settings/schedule/settings, hiddenForMultiTenancy: true, requires site admin) and is not affected by this issue. This bug is about using an existing named time range as a condition on a schedule task.

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/vslayout/DeviceRegistry.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/vslayout/DeviceRegistry.java
@@ -129,7 +129,8 @@ public final class DeviceRegistry {
    public static boolean isOrgAllowedToEditDevices(Principal principal) {
       return !LicenseManager.isEnterprise() ||
          OrganizationManager.getInstance().isSiteAdmin(principal) ||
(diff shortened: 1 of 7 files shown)


jshobe-inetsoft (approved):
Approved: Claude's automated review confirms both fixes (#75603 authorization gate and #75604 time-range enforcement) are correct with no blocking issues, prior review feedback is resolved, and all checks pass.

PR #4201, opened 2026-07-08
75592, NPE accessing layout

milotalon:
Root Cause:
VSLayoutService.getObjects() was the only place that built layout object models using the class's own instance field (this.objectModelService), whereas every other caller of createObjectModel() (VSLayoutControllerService, VSTextService, PresenterPropertyDialogService, FormatPainterService, ComposerViewsheetService) passes in its own injected VSObjectModelFactoryService.

On the layout-change execution path (changeViewsheetLayout -> sendLayout -> getObjects, which runs through the @ClusterProxy / lazy-initialization machinery), that instance field was null, so getObjects() invoked createModel() on a null service and threw the NullPointerException seen when selecting a device/mobile layout. The calling controller's own injected objectModelService was non-null on the same path (it is used without error elsewhere in the flow), which is why only this one code path failed.

This was a pre-existing (latent) defect, not specific to the Epic #74519 work.

Fix:
Threaded the VSObjectModelFactoryService through as an explicit parameter, matching the convention used by every other createObjectModel caller:
- VSLayoutService.sendLayout(...) and getObjects(...) now accept VSObjectModelFactoryService and pass it down to createObjectModel().
- The three sendLayout() call sites in VSLayoutControllerService (changeViewsheetLayout, layoutUndo, layoutRedo) now pass their injected objectModelService.

getObjects() no longer relies on the (possibly null) instance field, so the layout popup renders correctly.

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/vs/controller/VSLayoutControllerService.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/controller/VSLayoutControllerService.java
@@ -119,7 +119,7 @@ public class VSLayoutControllerService {
                rvs.addLayoutCheckPoint(layoutClone);
             }
 
-            vsLayoutService.sendLayout(rvsClone, layoutClone, dispatcher);
+            vsLayoutService.sendLayout(rvsClone, layoutClone, dispatcher, objectModelService);
          });
 
          viewsheetService.flushRuntimeSheet(rvsCloneId);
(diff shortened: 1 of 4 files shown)


milotalon:
Good catch — addressed in 9b3d7b9c0. Removed the now-dead `objectModelService` instance field along with its `@Autowired` constructor (and the unused import), since every method (`sendLayout`/`getObjects`/`createObjectModel`) now takes the service as an explicit parameter. The two test/config call sites were updated to the no-arg constructor. This eliminates the foot-gun of a future `this.objectModelService` reference silently reintroducing the NPE.

jshobe-inetsoft (approved):

PR #4205, opened 2026-07-09
fixed bug 75606,

bonnieshi43:
Bug Summary:
ActionPermissionService.orgAdminActionExclusions specifically excludes the three built-in internal scheduled tasks (asset file backup, balance tasks, update assets dependencies) from org-admin access, so SecurityEngine.checkPermission(orgAdmin, ResourceType.SCHEDULE_TASK, "__asset file backup__", ...) correctly returns false. However, the only authorization check actually applied when saving an internal task, ScheduleTaskService.canDeleteInternalTask(), grants access to any Organization Administrator by role, without ever consulting checkPermission() for SCHEDULE_TASK. An org admin can therefore modify the schedule (time conditions) of these protected internal system tasks via the task-save endpoint, despite the exclusion list saying they should not be able to.

Fix (community/core/src/main/java/inetsoft/web/admin/schedule/ScheduleTaskService.java): removed the isSiteAdmin/isOrgAdmin shortcut from canDeleteInternalTask() so it always delegates to engine.checkPermission(). Site admins are still granted (that check lives inside checkPermission() itself at DefaultCheckPermissionStrategy.java:209), and org admins now correctly get denied for the three excluded internal tasks.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/schedule/ScheduleTaskService.java
+++ b/core/src/main/java/inetsoft/web/admin/schedule/ScheduleTaskService.java
@@ -240,12 +240,10 @@ public class ScheduleTaskService {
          return false;
       }
 
-      OrganizationManager organizationManager = OrganizationManager.getInstance();
-
-      if(organizationManager.isSiteAdmin(principal) || organizationManager.isOrgAdmin(principal)) {
-         return true;
-      }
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4207, opened 2026-07-09
Fix unprotected circular imports between standalone components

sarapeng86:
Several standalone Angular components reference each other directly in
  their @Component `imports:` array, creating real (non-type-only) circular
  module dependencies. Depending on which file is loaded first, the module
  graph can resolve in an order where one side is still `undefined` when the
  other reads it, crashing with "Class extends value undefined" /
  "Cannot read properties of undefined (reading 'ɵcmp')".

  Wrap the eagerly-referenced side in forwardRef() (already the established
  pattern elsewhere in this codebase) so the reference is resolved lazily
  instead of captured at decorator-evaluation time:

  - vs-object-container.component.ts <-> vs-viewsheet.component.ts
  - composer-selection-container-children.component.ts <-> editable-object-container.component.ts
  - vs-calendar.component.ts <-> month-calendar.component.ts / year-calendar.component.ts
  - aggregate-dialog.component.ts <-> aggregate-pane.component.ts / crosstab-pane.component.ts
  - variable-list-dialog.component.ts <-> variable-list-editor.component.ts
  - ws-concatenation-editor-pane.component.ts <-> concatenation-pane-drop-target.component.ts

  For embed-chart.routes.ts <-> embed-chart.component.ts, convert the static
  `component:` route entry to `loadComponent:` (lazy import) instead, which
  is the idiomatic way to break a route-table/component cycle.

  Also break the one confirmed circular import outside the component-imports
  pattern: crosstab-actions.ts imported vs-crosstab.component.ts only to call
  the pure static method VSCrosstab.getDrillDirection(). Extracted it as a
  standalone function (getCrosstabDrillDirection) in crosstab-actions.ts so
  the action class no longer needs to import its own component.

### DIFF
--- a/web/projects/portal/src/app/composer/dialog/ws/aggregate-dialog.component.ts
+++ b/web/projects/portal/src/app/composer/dialog/ws/aggregate-dialog.component.ts
@@ -17,7 +17,7 @@
  */
 
 import { HttpParams } from "@angular/common/http";
-import { Component, EventEmitter, Inject, Input, OnInit, Output, DOCUMENT } from "@angular/core";
+import { Component, EventEmitter, forwardRef, Inject, Input, OnInit, Output, DOCUMENT } from "@angular/core";
 import { NgbModal } from "@ng-bootstrap/ng-bootstrap";
 import { Tool } from "../../../../../../shared/util/tool";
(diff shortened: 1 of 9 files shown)


jshobe-inetsoft (approved):

PR #4210, opened 2026-07-09
Feature to not animate graphs except in runtime mode.

milotalon:
Chart entrance animations are injected into the plot SVG server-side and
  replay on every regeneration, so editing in the Visual Composer, chart
  binding editor, or wizard re-triggers them repeatedly. Animation should
  only play in runtime-facing surfaces.

  The design-time vs runtime distinction lives on the client (ContextProvider),
  so the client now signals it via a "noanim" query param on the plot image
  request and the server skips the animation hint when set. A non-animated SVG
  carries no data-animated attribute; the inline-svg directive already adds
  .ready immediately in that case, so hover still works.

  Suppressed in: composer canvas, wizard editing, wizard preview, binding
  editor (contextProvider.composer || contextProvider.binding).
  Still animates in: composer Preview, real viewer, embedded.

  Also fixes a pre-existing NPE that crashed the chart binding pane on open and prevented testing:
  VSBindingPane.isExpressionAes dereferenced ref.dataInfo.columnValue when
  colorField/shapeField/sizeField were null (no aesthetic bound).

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/VGraphPair.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/VGraphPair.java
@@ -2018,11 +2018,20 @@ public class VGraphPair {
     * Get the image for plot.
     */
    public Graphics2D getPlotGraphic(int row, int col) {
+      return getPlotGraphic(row, col, true);
+   }
+
+   /**
+    * Get the image for plot.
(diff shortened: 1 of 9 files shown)


jshobe-inetsoft (approved):

PR #4213, opened 2026-07-09
fix NPEs in viewsheet binding pane and object view

milotalon:
Two frontend NullPointerExceptions were flooding the console during
  viewsheet binding:

  - VSBindingPane.isExpressionAes dereferenced ref.dataInfo without a null check. The goToWizardVisible getter runs on every change-detection cycle and passes the chart's colorField/shapeField/sizeField, which are null when no aesthetic is bound. Added null guards for ref and dataInfo.
  - VSObjectView.resizeModelView dereferenced this.model.objectType, but the method is invoked from a resize host listener (and ngAfterViewInit) that can fire before the model input is bound. Added an early return when model is unset, matching the existing guard in onResize.

  Both paths now safely no-op when the data is not yet present; no
  behavioral change.


jshobe-inetsoft (approved):

PR #4215, opened 2026-07-09
Bug #75595, resolve CalcRef formula results so they don't leak into calc table cell data

jshobe-inetsoft:
## Summary

After the Rhino→GraalJS migration (#75423), opening/refreshing a Freehand (Calc) Table whose formula cell's top-level result is a bare named-cell reference (`$name`) left the table stuck forever in the "loading" state. The server log filled with:

```
Could not write JSON: Cannot read field "y" because "loc" is null
  (... BaseTableCellModel["cellData"] -> inetsoft.report.script.formula.CalcRef["arraySize"])
Caused by: java.lang.NullPointerException: Cannot read field "y" because "loc" is null
  at inetsoft.report.script.formula.CalcRef.unwrap(CalcRef.java:334)
  at inetsoft.report.script.formula.CalcRef.getArraySize(CalcRef.java:267)
```

## Root Cause

In a runtime freehand table, `CalcTableScope.initRuntimeReferences()` binds each named cell as a `CalcRef` (a `ScriptArrayScope`), so `$name` resolves to a `CalcRef`. When a formula cell's top-level result is a bare `$name`, GraalJS returns it and `ScriptValueConverter.toHost()` unwraps the `ArrayProxy` back to the **raw `CalcRef` host object** — not the value it references (Rhino instead coerced a `Scriptable` result to its scalar default value). `CalcTableLens.evaluate()` then returned that `CalcRef` and `getValue()` cached it as the cell value.

Later, on the WebSocket dispatch thread, Jackson serialized `BaseTableCellModel.cellData` (a `CalcRef`) for `LoadTableDataCommand`. Because `CalcRef` is a plain bean to Jackson, it invoked the `getArraySize()` getter → `unwrap()` → `FormulaContext.getCellLocation()` returns `null` (no formula context on that thread) → NPE. The failed command flush meant the client never received the table data, so it spun on "loading."

## Fix

In `CalcTableLens.evaluate()`, resolve any `CalcRef` in the formula result to its referenced value (via a new `unwrapCalcRefs()` helper) before it is returned/cached — while the evaluating cell's location is still on the `FormulaContext` stack so `unwrap()` resolves correctly. The helper handles both a top-level `CalcRef` and `CalcRef`s nested inside array results (e.g. `[$a, $b]`), since `toHost` unwraps each array element back to its host object.

## Test Plan

- Added regression tests in `CalcTableLensTest` covering: a top-level `CalcRef`, a `CalcRef` inside an array, `CalcRef`s in nested arrays, and non-`CalcRef` results left unchanged.
- Verified `FreeExecution/FormulaScript/FScript1` renders correctly with no `Cannot read field "y" because "loc" is null` errors in the log.
- `CalcTableLensTest`, `RuntimeCalcTableLensTest`, `CalcTableScopeTest`, `CalcRefTest`, `ScriptValueConverterTest`, `CalcTableVSAScriptableTest` all pass.

Fixes Redmine #75595.

### DIFF
--- a/core/src/main/java/inetsoft/report/lens/CalcTableLens.java
+++ b/core/src/main/java/inetsoft/report/lens/CalcTableLens.java
@@ -27,6 +27,7 @@ import inetsoft.report.internal.Util;
 import inetsoft.report.internal.binding.*;
 import inetsoft.report.internal.table.*;
 import inetsoft.report.script.TableArray;
+import inetsoft.report.script.formula.CalcRef;
 import inetsoft.report.script.formula.CalcTableScope;
 import inetsoft.report.script.viewsheet.VSAScriptable;
 import inetsoft.report.script.viewsheet.ViewsheetScope;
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft (approved):
Approved.

Correct root-cause and fix: under GraalJS `toHost()` unwraps a bare `$name` result to the raw `CalcRef` host object (Rhino previously coerced the `Scriptable` to its scalar), so resolving `CalcRef → unwrap()` before the value is cached is the right layer — and doing it while the cell location is still on the `FormulaContext` stack (before the finally pops it) is what makes `unwrap()` resolve instead of NPE-ing on the serialization thread. Coverage of top-level / in-array / nested-array / non-CalcRef cases looks good, plus the manual FScript1 check.

Non-blocking note: `unwrapCalcRefs` mutates the `Object[]` in place — safe as long as `toHost` hands back a fresh host-side array (which it should); just worth confirming no path returns an array aliased to live script state.

PR #4218, opened 2026-07-09
Proxy Scope message not supported

milotalon:
Root cause found and fixed

  The bug: The failing line 4 is new StaticLineFrame(new GLine(3)) — specifically the inner new GLine(3). In the Rhino→GraalJS migration (#75423), chart script globals were registered two
  different ways in GraalJavaScriptEngine:

  - putClassProxy → JavaClassProxy, which implements ProxyInstantiable (supports new) and exposes static members.

  instantiate on ...ScopeProxy... Message not supported. In Rhino these were NativeJavaClass objects that supported both construction and constants. GShape had already been upgraded to
  putClassProxy for a related reason; these three were left behind.
  The fix (core/.../graal/GraalJavaScriptEngine.java): register GLine, GTexture, SVGShape via putClassProxy like GShape. Since ScriptHostAccess sets allowPublicAccess(true) and inetsoft.graph.*
  is allow-listed, a JavaClassProxy serves both new GLine(3) and GLine.THIN_LINE.

### DIFF
--- a/core/src/main/java/inetsoft/util/script/graal/GraalJavaScriptEngine.java
+++ b/core/src/main/java/inetsoft/util/script/graal/GraalJavaScriptEngine.java
@@ -273,12 +273,19 @@ public class GraalJavaScriptEngine implements AutoCloseable {
       };
 
       putConstantScope(bindings, "Chart", chartcls);
-      putConstantScope(bindings, "GLine", inetsoft.graph.aesthetic.GLine.class);
-      putConstantScope(bindings, "GTexture", inetsoft.graph.aesthetic.GTexture.class);
-      // GShape upgraded to JavaClassProxy so GShape.ImageShape (a public static
-      // nested class) is accessible in chart scripts alongside the static constants.
+      // GLine/GTexture/GShape/SVGShape are Java classes with both public
+      // constructors and public static final constants. Rhino exposed them as a
+      // NativeJavaClass, so scripts both construct them (new GLine(3), used by
+      // elem.setLineFrame(new StaticLineFrame(new GLine(3)))) and read their
+      // constants (GLine.THIN_LINE). A ConstantScope only surfaces the constants
+      // and is not instantiable ("instantiate on ScopeProxy ... Message not
+      // supported"), so register them as JavaClassProxy, which allowPublicAccess
+      // makes serve both the static constants and `new`.
+      putClassProxy(bindings, "GLine", "inetsoft.graph.aesthetic.GLine");
+      putClassProxy(bindings, "GTexture", "inetsoft.graph.aesthetic.GTexture");
+      // GShape also exposes GShape.ImageShape (a public static nested class).
       putClassProxy(bindings, "GShape", "inetsoft.graph.aesthetic.GShape");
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
## Code Review — PR #4218: Proxy Scope message not supported

**Verdict: Approve** ✅

### Summary
The Rhino→GraalJS migration (#75423) registered `GLine`, `GTexture`, and `SVGShape` as `ConstantScope` objects. A `ConstantScope` surfaces `public static final` constants but is **not instantiable**, so chart scripts that construct these classes — e.g. `elem.setLineFrame(new StaticLineFrame(new GLine(3)))` — failed at runtime with *"instantiate on ScopeProxy ... Message not supported"*. In Rhino these were `NativeJavaClass` objects supporting both `new` and constant access. This PR re-registers the three classes via `putClassProxy` (as `GShape` already was), which serves both `new GLine(3)` and `GLine.THIN_LINE`.

### Correctness & Completeness
- Verified all three classes have both public constructors and `public static final` constants: `GLine` (`GLine()`, `GLine(int)`, `GLine(double,double)` + `THIN_LINE` etc.), `GTexture` (`GTexture()`, `GTexture(int,double,int)` + `PATTERN_*`), `SVGShape` (`SVGShape()`, `SVGShape(String)` + `CHECK` etc.). All four new tests exercise valid signatures.
- `JavaClassProxy` is a strict **superset** of `ConstantScope`: `getMember` delegates to the GraalVM host type, exposing static fields (so `GLine.THIN_LINE` still resolves — covered by `gLineConstantStillResolves`), plus static methods and nested types, and implements `ProxyInstantiable`/`ProxyExecutable` for `new`. This restores fuller Rhino fidelity, not a partial fix.
- Change is consistent with the existing pattern — `GShape` and ~120 other `inetsoft.graph.*` classes are already registered via `putClassProxy`.

### Security
- No new attack surface: `inetsoft.graph.` is already in the class-lookup allow-list (`ScriptHostAccess`, allowed-prefixes), so these classes were already reachable by the sandbox. `allowPublicAccess(true)` scoped to allow-listed packages, with reflective escape paths denied via `denyAccess`, is unchanged. Switching the registration mechanism does not broaden what a script can load.

### Error Handling / Regressions / Compatibility
- `putClassProxy` is individually guarded (try/catch + `LOG.warn`), so a class-load failure won't abort engine init — same resilience as `putConstantScope`.
- The `StyleConstant` flattened scope still deliberately excludes `GLine`/`GTexture`/`GShape` (numeric-constant shadowing); this PR does not touch that, so no regression there.
- Backward compatible: constant reads that worked before continue to work; construction that was broken now works. No public API/signature changes.

### Tests
- Four focused unit tests added (construct `GLine`/`GTexture`/`SVGShape`, plus constant-still-resolves), following the existing `eval` helper convention. They run in the `core` suite and passed in the `build` check.

### Checks & Merge State
- All checks green: `build`, `Semgrep OSS`, `Semgrep Scan`, `claude-review`, `license/cla`.
- `mergeable: MERGEABLE` (no conflicts). The `BLOCKED`/`REVIEW_REQUIRED` state is solely due to the required-review gate, which this approval satisfies.

Clean, minimal, well-documented fix with the reasoning captured in comments. Recommend merge (do not auto-squash-merge).

PR #4222, opened 2026-07-10
fixed bug-75613

bonnieshi43:
**Summary**

When an admin uses Login-As to switch to another user, BasicAuthenticationFilter re-invokes the authentication pipeline with the admin's own submitted password but the target user's identity. The password comparison ends up checking the admin's password against the target user's stored password hash, not the admin's own — so it fails unless the two users happen to share a password.

The issues perhapse existed on previous version, please reviewer check whether need fixed the previous version.


jshobe-inetsoft (approved):
Approved: Claude's automated review after the latest commit found no blocking issues (both prior non-blocking points addressed, regression test added). All checks passing. Auto-merging.

PR #4224, opened 2026-07-10
Bug #75630, fix missing crosstab-optimized freehand table columns/aggregates

ngovus-inetsoft:
## Summary

- A freehand table combining a Top-N/"Others" ranked row group with a plain expand-horizontal column group (e.g. state rows x city columns, sum(price)) collapsed to a single column with missing/wrong aggregate values.
- Root cause: such tables qualify for the crosstab-optimization shortcut in `CalcTableVSAQuery`. The hidden crosstab computes the correct aggregation, but re-expanding it into the freehand grid evaluates generated formulas like `toList(data['city'])` against the crosstab result. `TableArray.hasMember()` gated bare column references on `Util.findColumn`, which matches literal header-row *text* — but a crosstab flattens dimension *values* into headers, never the dimension *name* itself, so `hasMember("city")` returned `false`, GraalJS never called `getMember`, and `data['city']` resolved to `undefined`. The row dimension only happened to work by coincidence (its name is literally the crosstab's corner-cell text).
- Fix: `TableArray.hasMember()` now defers bare-name lookups to `getMember` for `CrossFilter`-backed tables; `NamedCellRange.getCrosstabCells()` resolves a bare name matching a crosstab's own row/col dimension to that dimension's distinct values (skipping raw-null subtotal/grand-total placeholder tuples so multi-level grouped crosstabs don't get spurious blank columns/rows).

## Test plan

- [x] Reproduced the bug in a freehand table (state rows, Top-N + Others; city columns, expand-horizontal; sum(price)) — collapsed to one column with missing sums.
- [x] Applied fix, rebuilt `core`, confirmed live in the portal viewer: all city columns render with correct per-state price sums, including the "Others" bucket.
- [x] `mvn compile` passes on `core`.

### DIFF
--- a/core/src/main/java/inetsoft/report/script/TableArray.java
+++ b/core/src/main/java/inetsoft/report/script/TableArray.java
@@ -18,6 +18,7 @@
 package inetsoft.report.script;
 
 import inetsoft.report.*;
+import inetsoft.report.filter.CrossFilter;
 import inetsoft.report.internal.*;
 import inetsoft.report.lens.AttributeTableLens;
 import inetsoft.report.lens.SubTableLens;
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
Approving based on the completed Claude automated review, which confirmed both prior blocking issues are resolved and the remaining items are non-blocking suggestions. All checks pass and the PR is mergeable.

PR #4228, opened 2026-07-10
75624, waterfall graph bridge lines incorrect when used in a facet graph

milotalon:
Root Cause:

The waterfall "bridge" connector lines are created in IntervalElement and added to the shared top-level EGraph. In a facet chart, createGeometry() runs once per facet cell, but each bridge's tuple was just [x, y] with no facet dimension. A form is only scoped to a single sub-coordinate when its tuple length matches the full scale count (outer facet dims + inner x/y); with a bare [x, y] tuple the facet-matching check was skipped, so every cell's bridges were drawn in every cell — the doubled lines.

Three related defects surfaced from the same area:
- Empty space: bridges scoped out of a cell resolved to zero points but still produced an empty LineFormVO whose bounds collapse to the origin (0,0); since bridges are inPlot, layout padded the plot to swallow that phantom point.
- Short gap bridges: the endpoint inset (halfWidth) was multiplied by step (the category distance), so a bridge spanning missing categories shrank to a stub in the middle of the gap.
- Missing last->Total connector (facet only): the total bridge computed its x from getDim(0), which in a facet is the outer/facet dimension, mapping the total to the wrong x and yielding a negative step (no bridge). It worked in single-graph charts where getDim(0) is the x axis.

Fix:
- Prefix each bridge tuple with the facet nesting indices (GraphForm.getNestTuple) so a bridge renders only in its own facet cell.
- LineForm.createVisual(Coordinate) returns no visual when a form resolves to zero points (mirrors the existing trend-line handling), removing the phantom origin bounds.
- Compute halfWidth from one category unit instead of step, so gap-spanning bridges reach both bars.
- Compute the total bridge's x from the innermost dimension (getDim(getDimCount()-1)), which equals getDim(0) when there is no facet.

### DIFF
--- a/core/src/main/java/inetsoft/graph/element/IntervalElement.java
+++ b/core/src/main/java/inetsoft/graph/element/IntervalElement.java
@@ -26,6 +26,7 @@ import inetsoft.graph.data.DataSet;
 import inetsoft.graph.data.SortedDataSet;
 import inetsoft.graph.geometry.IntervalGeometry;
 import inetsoft.graph.geometry.Pie3DGeometry;
+import inetsoft.graph.guide.form.GraphForm;
 import inetsoft.graph.guide.form.LineForm;
 import inetsoft.graph.internal.GDefaults;
 import inetsoft.graph.scale.LinearScale;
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):
Approving based on the completed Claude automated review, which found no correctness, security, or backward-compatibility issues (only non-blocking minor observations). All checks pass and the PR is mergeable.

PR #4234, opened 2026-07-12
Window functions Phase 3: ROWS frames + LAST_VALUE

larryliang-inetsoft:
## Window functions — Phase 3: ROWS frames + `LAST_VALUE`

Adds an explicit **ROWS window frame** and enables **`LAST_VALUE`** for SQL window columns, on both the JDBC pushdown path (real `OVER(... ROWS BETWEEN ...)` SQL) and the in-memory `WindowTableLens` path.

Frame model is four scalars: `{ startBound, startOffset?, endBound, endOffset? }`, ROWS mode. Bound tokens: `UNBOUNDED_PRECEDING | PRECEDING | CURRENT_ROW | FOLLOWING | UNBOUNDED_FOLLOWING`.

### Commits
- `WindowExpressionRef`: ROWS frame fields + `getFrameFragment()`; `LAST_VALUE` frame-less gets an explicit whole-partition frame; **frame-less windows emit no frame (byte-parity)**.
- `/ws/table` (`WorksheetTable` + `WorksheetTableService`): structured `WindowFrameInfo` wire + fail-loud guards (frameable-fn, bound enum, offset-required for PRECEDING/FOLLOWING, bounded-frame-requires-orderBy, start≤end).
- `WindowTableLens`: in-memory ROWS frames (`frameSlice`/`boundPos`) + `LAST_VALUE` kernel; frame-less path index-equivalent to the prior running/whole-partition behavior.
- `AssetQuery.buildWindowSpecs`: carries the frame into the lens `Spec`.

### Byte-parity gate
Frame-less window columns emit byte-identical SQL and compute identical in-memory results to before this change. Verified on a live suitecrm/Postgres deployment (`log_statement=all`): frame-less `ROW_NUMBER`/`SUM` emit `OVER(...)` with **no** `ROWS BETWEEN`.

### Live verification (suitecrm/Postgres)
Postgres query log confirmed the executed SQL:
- `AVG(...) OVER (... ROWS BETWEEN 2 PRECEDING AND CURRENT ROW)` — trailing-3 moving average, values hand-checked correct
- `LAST_VALUE(...) OVER (... ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING)` — returns each partition's actual last value
- `ROW_NUMBER()` / `SUM(...) OVER (...)` — no frame (byte-parity)

### Tests
Per-task unit suites green (`WindowExpressionRef` 20/20, `WorksheetTableService` window guards 8/8, `WindowTableLens` 13/13, `WindowRoutingTest` 10/10). Full whole-branch review: **Ready to merge**, no Critical/Important findings.

Paired with the wiz-services PR (frame wiring + `LAST_VALUE` support in `windowColumns`).

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/AssetQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/AssetQuery.java
@@ -3019,8 +3019,16 @@ public abstract class AssetQuery extends PreAssetQuery {
             }
          }
 
-         specs.add(new WindowTableLens.Spec(
-            ref.getName(), w.getFn(), arg, w.getN(), parts, ocols, oasc));
+         if(w.hasExplicitFrame()) {
+            specs.add(new WindowTableLens.Spec(
+               ref.getName(), w.getFn(), arg, w.getN(), parts, ocols, oasc,
+               w.getFrameStartBound(), w.getFrameStartOffset(),
+               w.getFrameEndBound(), w.getFrameEndOffset()));
+         }
+         else {
+            specs.add(new WindowTableLens.Spec(
+               ref.getName(), w.getFn(), arg, w.getN(), parts, ocols, oasc));
+         }
       }
 
       checkCompatiblePartitionAndOrder(specs);
--- a/core/src/main/java/inetsoft/report/lens/WindowTableLens.java
+++ b/core/src/main/java/inetsoft/report/lens/WindowTableLens.java
@@ -43,9 +43,23 @@ public class WindowTableLens extends AbstractTableLens implements TableFilter {
       public final int[] partCols;   // PARTITION BY base col indices (may be empty)
       public final int[] orderCols;  // ORDER BY base col indices (may be empty)
       public final boolean[] orderAsc;
+      // ROWS frame bounds. startBound == null means "frame-less" — the kernel falls back to the
+      // Phase-1/2 default (running for an ordered aggregate, whole partition otherwise, whole
+      // partition for FIRST_VALUE/LAST_VALUE). Bound tokens: UNBOUNDED_PRECEDING,
+      // UNBOUNDED_FOLLOWING, CURRENT_ROW, PRECEDING (with offset), FOLLOWING (with offset).
+      public final String startBound;
+      public final int startOffset;
+      public final String endBound;
+      public final int endOffset;
 
       public Spec(String header, String fn, int argCol, int n,
                   int[] partCols, int[] orderCols, boolean[] orderAsc) {
+         this(header, fn, argCol, n, partCols, orderCols, orderAsc, null, 0, null, 0);
+      }
+
+      public Spec(String header, String fn, int argCol, int n,
+                  int[] partCols, int[] orderCols, boolean[] orderAsc,
+                  String startBound, int startOffset, String endBound, int endOffset) {
          this.header = header;
          this.fn = fn == null ? "" : fn.toUpperCase();
          this.argCol = argCol;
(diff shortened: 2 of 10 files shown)


larryliang-inetsoft:
## Review — Phase 3 ROWS frames + `LAST_VALUE` (Java backend)

Reviewed together with the wiz PR (stylebi-wiz #1190). Careful, well-tested work; the byte-parity design is sound and the two-layer validators agree on the hard part (frame ordering). Two real divergences and one defensive note.

### Cross-layer contract parity — verified case by case
Checked TS `validateFrame`/`boundPosition` against Java `requireFrameOffset`/`frameBoundRank`/`validateFrameOrder`:
- Frameable-fn set — identical (`SUM/AVG/COUNT/MIN/MAX/FIRST_VALUE/LAST_VALUE`) ✓
- Bound enum — identical five tokens ✓
- Frame-order total order — identical (`UNBOUNDED_PRECEDING < N PRECEDING < CURRENT_ROW < N FOLLOWING < UNBOUNDED_FOLLOWING`, larger-offset-PRECEDING earlier, smaller-offset-FOLLOWING earlier) ✓
- Bounded-frame-requires-orderBy — identical ✓

### 🟠 1. Offset on a fixed bound: Java silently ignores; wiz rejects
`requireFrameOffset` returns `0` for `UNBOUNDED_*`/`CURRENT_ROW` without inspecting a supplied offset, so `{startBound:"CURRENT_ROW", startOffset:1, …}` is **accepted** here but **rejected** by the wiz `validateFrame` (`startOffset must be omitted …`). This contradicts the "accept/reject decisions are identical across the two layers" claim. Benign in output terms (the offset is ignored in `frameBoundSql` regardless, and wiz is the stricter gateway), but for true parity either reject a stray offset on a fixed bound here too, or soften the claim to "wiz is a strict superset." Low severity, one-liner.

### 🟠 2. In-memory `SUM` over an empty frame returns `0.0`; SQL/pushdown returns `NULL`
`WindowTableLens.kernel` returns `COUNT→0`, **`SUM→0.0`**, `AVG/MIN/MAX→null` when `frameSlice == null`. An empty frame is reachable with a *valid* forward frame — e.g. `ROWS BETWEEN 1 FOLLOWING AND 2 FOLLOWING` on the last rows of a partition (passes `start ≤ end`). In ANSI SQL/Postgres, `SUM` over an empty frame is **`NULL`**, so the pushdown path yields `NULL` and the in-memory path yields `0.0` for the identical query. `COUNT→0` and `AVG/MIN/MAX→null` are correct; only `SUM→0.0` diverges. The two paths are meant to be interchangeable — recommend `SUM`-empty → `null` to match. It was a deliberate choice (spec §5), so worth an explicit decision.

**Test gap:** despite spec §8 listing "frame past the edge (clamping + empty-frame results)," there is no empty-frame test in `WindowTableLensTest` (moving-average, whole-partition `LAST_VALUE`, and a centered frame that never empties) — exactly the case where the paths diverge. Please add it.

### 🟢 3. `LAST_VALUE` whole-partition default keys on exact-case `fn`
`getFrameFragment` decides the synthesized frame via `"LAST_VALUE".equals(fn)`, and `WorksheetTableService` builds the node with the raw request `fn`. Wiz's zod enum guarantees upper-case, so this is safe in practice — but a raw `/ws/table` caller sending `"last_value"` gets no frame clause and the silent current-row footgun this change exists to prevent. Defensive: normalize `fn` case in `WindowExpressionRef` (or compare case-insensitively).

### Byte-parity — verified sound
- `getOverBody()` rewrite (fragment list + `String.join(" ", …)`) reproduces the old space-join for all four frame-less combinations; frame-less aggregate and frame-less `FIRST_VALUE` emit no `ROWS` clause. ✓ (`frameless_*_byteParity` tests).
- In-memory `frameSlice` for `startBound == null` reproduces Phase-1/2 slices exactly: value-fn/order-less → whole partition `[0,size-1]`; ordered aggregate → running `[0,p]` (inclusive `q <= fr[1]` == old `q < start+p+1`). ✓

### Verdict
Low-risk; nothing blocks the frame-less regression story. Before merge I'd want **#2** decided (SUM-empty semantics + a test); **#1** is a cheap parity tightening; **#3** is defensive.

PR #4235, opened 2026-07-12
Window functions Phase 4: RANGE/GROUPS frame modes

larryliang-inetsoft:
## Window functions — Phase 4: RANGE / GROUPS frame modes

Adds SQL window **frame modes** `RANGE` and `GROUPS` (Phase 3 shipped `ROWS` only), on both the JDBC pushdown path and the in-memory `WindowTableLens`. A dialect **capability gate** routes non-pushable frames to the in-memory engine, so every mode works on every datasource.

### Frame model
`frame` gains `mode` (`ROWS`|`RANGE`|`GROUPS`, default ROWS) and `offsetUnit` (`year…second`, for date-valued RANGE). `mode` absent ⇒ ROWS, so every Phase-3 caller and stored frame is byte-identical.

### Commits (Tasks 1–4 + review fixes)
- `WindowExpressionRef`: frame `mode`/`offsetUnit`; `getFrameFragment` emits `<MODE> BETWEEN …`; RANGE date → `INTERVAL '<n> <unit>'`; GROUPS → `GROUPS BETWEEN …`. `frameMode` persisted only for RANGE/GROUPS (ROWS stored null → **no XML serialization drift** for Phase-3 assets). mode/offsetUnit token validation.
- `/ws/table` (`WorksheetTable.WindowFrameInfo` + `WorksheetTableService`): accept + validate `mode`/`offsetUnit`, fail-loud (mode enum; RANGE-value/GROUPS require orderBy; RANGE-value requires exactly one orderBy; offsetUnit only for RANGE with a value-offset bound on a date/time key; unit case-normalized). Gated so a no-orderBy whole-partition RANGE never crashes.
- `WindowTableLens`: in-memory RANGE (value + peer, calendar-aware date arithmetic) and GROUPS; symmetric bounds (FOLLOWING-start / PRECEDING-end); null-order-value fail-loud (NULLS deferred). ROWS/frame-less byte-parity preserved.
- `AssetQuery`: `framePushable(mode, helper)` (v1: `postgresql` → ROWS/RANGE/GROUPS; else ROWS only); non-pushable frames excluded from pushed SQL (via the `mergeColumn` choke point) and computed in-memory instead — **no double-compute**, and **Postgres byte-parity byte-identical** (every frame pushable ⇒ gate is a no-op).

### Byte-parity gate
ROWS and frame-less windows produce identical pushdown SQL and identical in-memory values to Phase 3. On Postgres the capability gate never changes routing or SQL.

### Tests
Per-task unit suites green (WindowExpressionRef, WorksheetTableServiceWindowColumns, WindowTableLens incl. DESC-RANGE + calendar-month, WindowRouting). Each task adversarially reviewed; findings fixed (XML byte-parity, no-orderBy crash, symmetric bounds, routing correctness).

### Verification status
Live pushdown + in-memory-fallback verification (`local-992`) and the whole-branch review are in progress; results/any fixes will be pushed here before merge.

Deferred (documented): `EXCLUDE` clause, `NULLS FIRST/LAST`, fractional numeric offsets, non-Postgres dialect *pushdown* (those dialects use the in-memory fallback); and, gated on adding a non-PG dialect, the `mergeGroupBy` bookkeeping follow-up.

Paired with wiz-services PR (frame `mode`/`offsetUnit` wiring + validation).

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/AssetQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/AssetQuery.java
@@ -2957,12 +2957,45 @@ public abstract class AssetQuery extends PreAssetQuery {
     * entity+attribute / header (the codebase convention), not a bare name scan of the selection.
     * <p>
     * Package-private and static so it can be unit-tested directly (see WindowRoutingTest)
-    * without a live query.
+    * without a live query. Builds a spec for EVERY window column in {@code cols}, regardless of
+    * whether its frame is pushable on any particular dialect — use the 3-arg overload to filter
+    * out pushed-down columns for the production in-memory routing path (see
+    * {@link #buildWindowSpecs(XTable, ColumnSelection, SQLHelper)}).
     * @param base the runtime base table the WindowTableLens will wrap and read from.
     * @param cols the column selection to scan for WindowExpressionRef columns.
     * @return the resolved specs, one per window column, in selection order.
     */
    static WindowTableLens.Spec[] buildWindowSpecs(XTable base, ColumnSelection cols) {
+      return buildWindowSpecs(base, cols, null, false);
+   }
+
+   /**
+    * Like {@link #buildWindowSpecs(XTable, ColumnSelection)}, but SKIPS a window column whose
+    * frame is pushable on {@code helper}'s dialect (see
+    * {@link #framePushable(String, SQLHelper)}) — i.e. one already merged into the pushed-down
+    * SQL by {@link #mergeColumn}. Used by the mixed-pushability routing case in
+    * {@link #getWindowTableLens}: when a query is mergeable but not every window column's frame
+    * is pushable (e.g. one ROWS column + one RANGE column on a non-Postgres source), the ROWS
+    * column is already in the generated SQL and must NOT also be recomputed here — building a
+    * spec for it would (1) double-compute it and (2) subject it to
+    * {@link #checkCompatiblePartitionAndOrder} alongside the non-pushed column, which can throw
+    * for two window columns with divergent PARTITION BY that previously worked fine via
+    * independent SQL {@code OVER()} clauses. A pushable column is thus computed exactly once (in
+    * SQL); a non-pushable one exactly once (in memory).
+    * @param base the runtime base table the WindowTableLens will wrap and read from.
+    * @param cols the column selection to scan for WindowExpressionRef columns.
+    * @param helper the source dialect's SQL helper (may be {@code null}).
+    * @return the resolved specs for the NON-pushable window columns only, in selection order.
+    */
+   static WindowTableLens.Spec[] buildWindowSpecs(XTable base, ColumnSelection cols,
+                                                   SQLHelper helper)
+   {
+      return buildWindowSpecs(base, cols, helper, true);
+   }
+
+   private static WindowTableLens.Spec[] buildWindowSpecs(XTable base, ColumnSelection cols,
+                                                           SQLHelper helper, boolean skipPushable)
+   {
       List<WindowTableLens.Spec> specs = new ArrayList<>();
 
       for(int i = 0; i < cols.getAttributeCount(); i++) {
@@ -2979,6 +3012,13 @@ public abstract class AssetQuery extends PreAssetQuery {
          }
 
          WindowExpressionRef w = (WindowExpressionRef) bref;
+
+         if(skipPushable && framePushable(w.getFrameMode(), helper)) {
+            // Already merged into the pushed-down SQL by mergeColumn — do not also compute it
+            // in memory (see javadoc above).
+            continue;
+         }
+
          DataRef argRef = w.getArgRef();
          int arg = argRef == null ? -1 : AssetUtil.findColumn(base, argRef);
          List<DataRef> partRefs = w.getPartitionBy();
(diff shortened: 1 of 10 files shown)


larryliang-inetsoft:
## Review — Phase 4 RANGE/GROUPS frame modes (Java backend)

Reviewed with the wiz PR (stylebi-wiz #1191). This is a big, careful change and the test suite is genuinely strong — the DESC-vs-value-threshold peer distinction (`range_peer`, `range_descValueOffset`), calendar-month clamping (`range_dateMonth`, Feb-28), null-order fail-loud, XML byte-parity (`frameMode` stored null for ROWS → no serialization drift), and the capability gate are all tested, and the Task-3 symmetric-bounds review fixes are present. ROWS/frame-less byte-parity checks out (mode used only on the explicit branch; `frameOffsetSql` emits a bare int when `frameOffsetUnit == null`).

Three findings, one of which I'd treat as a merge blocker.

### 🔴 1. In-memory GROUPS: an all-out-of-range bound clamps to a boundary group instead of empty (silent wrong result, and a test locks it in)
`groupsSlice` clamps **both** group indices into `[0, lastGroup]` *before* the empty check:
```java
startGroup = Math.max(0, Math.min(lastGroup, startGroup));
endGroup   = Math.max(0, Math.min(lastGroup, endGroup));
if(startGroup > endGroup) return empty;
```
For `GROUPS BETWEEN 2 PRECEDING AND 1 PRECEDING` at the **first** group: raw `startGroup=-2`, `endGroup=-1`; both clamp to `0`, so `0 > 0` is false and the frame returns **group 0** instead of empty. Postgres (which *pushes* GROUPS down) returns **NULL** here — the frame is groups `[-2,-1]`, no rows. Same bug symmetrically on the trailing side (`1 FOLLOWING AND 2 FOLLOWING` at the last group). So the identical logical query yields group-0's sum on a non-PG (in-memory) source and NULL on Postgres.

Contrast the ROWS path, which is correct: `frameSlice` clamps `loc`/`hic` independently and returns null when `loc > hic` (with `hic` allowed to reach `-1`). GROUPS destroys that signal by clamping the end up to 0 first.

**Fix:** detect emptiness on the *raw* (pre-clamp) group indices — `if(rawEnd < 0 || rawStart > lastGroup || rawStart > rawEnd) return empty;` — then clamp for the row scan. **`WindowTableLensTest.groups_precedingRange` currently asserts the wrong values** (20.0 for group 0) and must be updated to expect NULL for the fully-preceding-out-of-range rows.

### 🟠 2. Value-offset RANGE on a date/time order key with `offsetUnit` omitted is neither rejected nor computed correctly
The guard chain enforces `offsetUnit present ⇒ order key is date` (`requireDateOrderKey`), but **not** the converse. A `RANGE PRECEDING n / FOLLOWING n` frame whose single order key is a date/timestamp but with **no** `offsetUnit` passes validation. Then:
- **Pushdown (PG):** emits `RANGE BETWEEN 5 PRECEDING AND CURRENT ROW` with a date `ORDER BY` → Postgres runtime error (a date RANGE offset must be an `INTERVAL`). Loud, but late.
- **In-memory:** `offsetOrderValue` → date branch → `dateOffsetMs(v, unit=null, signedOffset)` hits the `if(unit == null) return v.getTime() + signedOffset` line — the offset is treated as **raw milliseconds** (5 PRECEDING = 5 ms). Silently wrong frame, no error.

A numeric offset on a date key is meaningless in SQL — it always needs an interval. Recommend rejecting a value-offset RANGE whose (single) order key resolves to a date/time type when `offsetUnit` is absent, in `applyWindowColumns` (authoritative), mirrored in wiz. The `dateOffsetMs` `unit==null` "defensive: treat as raw ms" branch is then genuinely unreachable and can throw instead. This is precisely the "tool-misuse is a plugin gap → fail loud" case from CLAUDE.md.

### 🟠 3. Mixed pushable + non-pushable window columns on one table — verify the "computed exactly once" invariant
`getWindowTableLens` computes **every** window column in the selection in-memory whenever **any** frame is non-pushable (`!windowFramesAllPushable`), because `buildWindowSpecs` builds a Spec for every `WindowExpressionRef` in `cols` unconditionally. That is correct only if a non-pushable window column forces the *whole* query non-mergeable so nothing is pushed. But the gate is `isQueryMergeable(false) && windowFramesAllPushable(cols)` — two independent predicates. Consider one **ROWS** column (pushable → `mergeColumn` delegates to super → inlined in the pushed SQL) plus one **GROUPS** column on a non-PG source (non-pushable → `mergeColumn` returns false → excluded), same partition/order (nothing in `checkCompatiblePartitionAndOrder` forbids mixing *modes*). If `isQueryMergeable(false)` is still true in that state, `base` already carries the pushed ROWS result **and** `buildWindowLens` recomputes it — double column / wrong output.

I can't resolve this from the diff — it depends on whether a single `mergeColumn`-false flips `isQueryMergeable` to false (then `base` is raw and all-in-memory is consistent) or allows a partial push (then it double-computes). The PR's central safety claim ("computed exactly once, never both") rests on this. Please either add a live/integration test with exactly this shape (one ROWS + one GROUPS window col, same PARTITION BY/ORDER BY, mergeable non-Postgres source) or have `buildWindowSpecs`/`buildWindowLens` compute **only the non-pushable** window columns when the query did push down.

### 🟢 4. `orderValue`/`offsetOrderValue` cast `(Number) v` / `(Date) v`
A value-offset RANGE whose order key is neither numeric nor a `Date` (e.g. a string key that slipped past validation on the raw `/ws/table` path) throws a bare `ClassCastException` rather than a named error. Low severity; a typed guard would match the fail-loud tone of the rest.

### Verdict
Strong work; ROWS byte-parity and the SQL-pushdown story are solid. **#1 is a blocker** (silent wrong result + a test asserting it). **#2** is a silent-wrong-result on the in-memory path worth closing. **#3** is the invariant the whole capability gate depends on — verify it live or narrow the in-memory recompute. #4 is a nicety. Mirror notes for the validation halves are on #1191.

PR #4236, opened 2026-07-12
Window frames: Phase-4 review follow-ups (GROUPS orderBy, numeric-RANGE key, gcolumns gating)

larryliang-inetsoft:
## Window frames — Phase-4 review follow-ups (parity + bookkeeping cleanup)

Three small, non-blocking follow-ups deferred from the Phase-4 (RANGE/GROUPS) PR reviews. Guard/validation + bookkeeping only; no new capability. Unit-covered (93/93 across the window suites).

### 1. GROUPS always requires `orderBy` (`WorksheetTableService`)
Drops the whole-partition exemption on the GROUPS-requires-orderBy guard. Postgres/ANSI require an ORDER BY for GROUPS mode regardless of bounds, and the wiz TS validator already rejects any GROUPS without `orderBy` — this restores TS↔Java parity and prevents emitting invalid SQL for a whole-partition GROUPS on the raw `/ws/table` path. Test `groupsWholePartition_noOrderBy_succeeds` inverted → `_throws` (it encoded the old bug).

### 2. Numeric RANGE value-offset requires a numeric order key (`WorksheetTableService`)
Adds the converse of the existing date-key check (which the wiz TS validator already enforces): a numeric RANGE value-offset (no `offsetUnit`) whose single order key resolves to a non-numeric, non-date type now fails loud. Best-effort — defers when the type is unresolvable, matching the existing pattern. New test `numericRangeValueOffset_stringOrderKey_throws`.

### 3. Keep a non-pushed window column out of `gcolumns` (`PreAssetQuery.mergeGroupBy`)
The window-expression special case used the 2-arg `void mergeColumn` (always emits) then unconditionally added the column to `gcolumns`. Switched to the 3-arg boolean `mergeColumn` and gated `gcolumns.addAttribute` on its return, so a window column that `AssetQuery`'s dialect-capability override refuses to push down (a non-pushable RANGE/GROUPS frame on a non-Postgres source) is no longer tracked in `gcolumns` when it's absent from the emitted SELECT — fixing a bookkeeping mismatch on the hybrid grouped-SQL + in-memory-window path.

**Byte-parity (all-Postgres path):** unchanged. `isAggregateInfoMergeable()` is provably `true` at that branch (`mergeGroupBy` returns earlier otherwise), matching the old 2-arg's hardcoded flag; and on Postgres `framePushable` is always true so the override falls through to `super.mergeColumn` and the column is emitted + tracked exactly as before. Verified by reading + the full window suite passing.

Fix 3's non-Postgres bug branch has no dedicated unit seam (would need a non-PG SQLHelper harness — disproportionate; it's unreachable until a non-PG dialect is added); covered by existing `WindowRoutingTest`/`WindowTableLensTest` + code inspection.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/PreAssetQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/PreAssetQuery.java
@@ -656,8 +656,18 @@ public abstract class PreAssetQuery implements Serializable, Cloneable {
          // it OUT of the GROUP BY list (built below from info.getGroups() only). Without this it is
          // silently dropped from the SELECT when the table also carries group/aggregates.
          else if(column.isVisible() && group == null && isWindowExpression(column)) {
-            mergeColumn(nselection, column);
-            gcolumns.addAttribute(column);
+            // Use the 3-arg boolean mergeColumn (not the 2-arg void variant, which always
+            // emits) so a window column that AssetQuery's override refuses to push down (a
+            // non-pushable RANGE/GROUPS frame on a non-Postgres dialect) is kept OUT of
+            // gcolumns when it's absent from the SELECT list — otherwise gcolumns count would
+            // not match the emitted SQL columns on that hybrid grouped-SQL + in-memory-window
+            // path. isAggregateInfoMergeable() is always true here (mergeGroupBy already
+            // returned false above when it isn't), so this is byte-identical to the prior
+            // unconditional mergeColumn(nselection, column) + gcolumns.addAttribute(column) for
+            // every dialect that CAN push the frame down (all-Postgres included).
+            if(mergeColumn(nselection, column, isAggregateInfoMergeable())) {
+               gcolumns.addAttribute(column);
+            }
          }
 
          // if we have no aggs or one agg for a column, increase the count by 1
--- a/core/src/main/java/inetsoft/report/composition/execution/PreAssetQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/PreAssetQuery.java
@@ -661,11 +661,12 @@ public abstract class PreAssetQuery implements Serializable, Cloneable {
             // non-pushable RANGE/GROUPS frame on a non-Postgres dialect) is kept OUT of
             // gcolumns when it's absent from the SELECT list — otherwise gcolumns count would
             // not match the emitted SQL columns on that hybrid grouped-SQL + in-memory-window
-            // path. isAggregateInfoMergeable() is always true here (mergeGroupBy already
-            // returned false above when it isn't), so this is byte-identical to the prior
-            // unconditional mergeColumn(nselection, column) + gcolumns.addAttribute(column) for
-            // every dialect that CAN push the frame down (all-Postgres included).
-            if(mergeColumn(nselection, column, isAggregateInfoMergeable())) {
+            // path. The 2-arg mergeColumn this replaces delegates with a hardcoded true, and
+            // mergeGroupBy already returned false above when !isAggregateInfoMergeable(), so
+            // passing true here is byte-identical to the prior unconditional
+            // mergeColumn(nselection, column) + gcolumns.addAttribute(column) for every dialect
+            // that CAN push the frame down (all-Postgres included).
+            if(mergeColumn(nselection, column, true)) {
                gcolumns.addAttribute(column);
             }
          }
--- a/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
@@ -1192,12 +1192,13 @@ public class WorksheetTableService {
                || "PRECEDING".equals(frame.getEndBound()) || "FOLLOWING".equals(frame.getEndBound());
 
             // RANGE value-offset and GROUPS need an ORDER BY; RANGE value-offset needs exactly
-            // one. A whole-partition GROUPS frame (UNBOUNDED_PRECEDING..UNBOUNDED_FOLLOWING) is
-            // exempt, same as the general bounded-frame check above — it needs no peer-group
-            // ordering since it covers every row regardless of order.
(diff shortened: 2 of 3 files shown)


larryliang-inetsoft:
## Review — Phase-4 review follow-ups (GROUPS orderBy, numeric-RANGE key, gcolumns gating)

Small, well-scoped cleanup closing the deferred items from the #4235 review. I verified each fix against the merged Phase-4 code rather than taking the byte-parity claim on faith.

### Fixes — all three correct
**1. GROUPS always requires orderBy** ✓ — dropping the whole-partition exemption is right: Postgres/ANSI reject GROUPS mode without an `ORDER BY` regardless of bounds (peer groups are *defined* by the order key), so the old exemption emitted invalid SQL and diverged from the wiz TS validator. Test inversion (`_succeeds` → `_throws`) correctly re-encodes it.

**2. Numeric RANGE value-offset requires a numeric order key** ✓ — the right converse of the existing date-key guard, and the `dt != null` deferral matches the best-effort pattern. `numericRangeValueOffset_stringOrderKey_throws` + the untyped-column fix on `..._stillSucceeds` cover it.

**3. gcolumns gating** ✓ — and byte-parity **verified**, not just claimed: the 2-arg `mergeColumn` delegates to `mergeColumn(nsel, col, true)` (PreAssetQuery:1310), and `mergeGroupBy` early-returns when `!isAggregateInfoMergeable()`, so `isAggregateInfoMergeable()` is provably `true` at the call site → for every pushable column (all-Postgres) the 3-arg returns true and `gcolumns.addAttribute` runs exactly as before. The only behavior change is excluding a non-pushable column from `gcolumns`, which is the intended fix.

### 🟠 A. Fix 3's existence implies partial-push of mixed tables — does the recompute path then double-count the pushable columns?
Fix 3 only *fires* (3-arg returns false → skip `gcolumns`) when a non-pushable window column is dropped from the SELECT **while the query is still merging** (`mergeGroupBy` is running). That state is exactly the mixed case I flagged in #4235 (finding 3): one pushable ROWS column (emitted into `base`) + one non-pushable RANGE/GROUPS column (dropped), on a mergeable non-Postgres JDBC source. In that state, `getWindowTableLens` sees `!windowFramesAllPushable` and calls `buildWindowLens`, whose `buildWindowSpecs` builds a Spec for **every** `WindowExpressionRef` in the selection — including the pushable ROWS one already computed in `base` → it gets appended a second time (double-count / duplicate column). Fix 3 correctly repairs the `gcolumns` count but doesn't address that recompute. Either `buildWindowSpecs`/`buildWindowLens` should build specs **only for non-pushable** window columns when the query pushed down, or please confirm a single non-pushable column forces the whole query non-mergeable (in which case Fix 3 would never fire and nothing is pushed). The PR notes the non-PG branch is "unreachable until a non-PG dialect is added," but MySQL/others already exist as dialects, so a mixed ROWS+GROUPS request on one appears reachable today.

### 🟠 B. The "defer when unresolvable" intent is defeated by `getDataType()` returning STRING for untyped refs
Both the date-key guard and the new numeric-key guard defer only when `dt == null`. But an untyped `AttributeRef.getDataType()` returns `XSchema.STRING` (not null) — your own `..._stillSucceeds` test had to add `amountRef.setDataType(DOUBLE)` to stop the new guard from tripping. So on the raw `/ws/table` path, a legitimate numeric RANGE value-offset whose order ref arrives **untyped** is treated as "resolved = STRING = non-numeric" and rejected outright, never deferred. Please confirm order refs reaching `applyWindowColumns` always carry a resolved type (then this is moot); otherwise the deferral is illusory for untyped refs and will over-reject valid numeric RANGE frames.

### 🟢 C. Minor
Fix 3 passes `isAggregateInfoMergeable()` where it's provably `true` at that point; a literal `true` would be exact and make the byte-parity argument self-evident (re-calling is harmless, just less clear).

### Verdict
Good cleanup; the three fixes are correct and byte-parity checks out. **A** is the one worth resolving before merge (it's the tail of #4235 finding 3 and Fix 3 makes the triggering state concrete); **B** is a plausible over-rejection to confirm; **C** is cosmetic.

larryliang-inetsoft:
Thanks — worked through both 🟠 items against the merged code; neither needs a code change, and the two cosmetic notes are addressed.

**A (mixed-pushability double-count) — already handled, not reachable.** `getWindowTableLens`'s mixed-case path doesn't use the 2-arg `buildWindowSpecs`; it calls `buildWindowLens(base, cols, helper)` → `buildWindowSpecs(base, cols, helper, /*skipPushable=*/true)` (AssetQuery.java ~2990/2996), which **skips** any window column whose frame is pushable on the dialect. So in a mixed ROWS(pushed)+GROUPS(non-pushable) query the pushed ROWS column is computed once in SQL and only the non-pushable column gets an in-memory spec — no recompute/double-count. (This was Phase-4 PR #4235's Fix B; the all-pushable branch early-returns `base`, the fully-non-mergeable branch uses the unfiltered overload.)

**B (untyped-ref over-rejection) — moot on the real path.** Order refs are resolved as `cs.getAttribute(ob.getField())` (WorksheetTableService.java:1142) against the table's own column selection, so `getDataType()` returns the resolved column type (`double`/`integer`/`timeInstant`/…), not the `AttributeRef` STRING default. A numeric column reports a numeric type and passes; the `_stillSucceeds` test needed an explicit `setDataType(DOUBLE)` only because it builds a bare `AttributeRef` fixture. The pre-existing date-key guard (shipped in #4235) uses the identical `dt == null` pattern, so Fix 2 adds no new behavior there.

**C — done:** replaced `isAggregateInfoMergeable()` with a literal `true` at the window branch (873186155); it's provably true there and matches the 2-arg `mergeColumn`'s hardcoded `true`, making the byte-parity self-evident.

Import-style nit: the test file already uses fully-qualified `inetsoft.uql.schema.XSchema.*` throughout (lines 248–571, pre-existing), so the new refs match the file's established style — left as-is to avoid an inconsistent/oversized diff.

PR #4237, opened 2026-07-12
Window functions Phase 5: dialect-aware frame pushdown (all SQLHelper dialects)

larryliang-inetsoft:
## Window functions Phase 5 — dialect-aware frame pushdown

Moves window-frame pushdown capability + dialect frame-SQL rendering into the `SQLHelper` hierarchy, replacing Phase 4's hardcoded Postgres-only gate. Every StyleBI database type now pushes down the frame shapes it supports and falls back to the in-memory `WindowTableLens` for the rest. **Community-repo only** (wiz-services frame validation is dialect-agnostic — no wiz change).

### Design: default-deny, opt-in per verified dialect
Getting a capability *wrong-high* → a loud DB syntax error on that dialect; *wrong-low* → the correct in-memory fallback. So the base `SQLHelper` denies the advanced frame shapes and each dialect opts in only to what's verified from its SQL reference. Every one of the 23 helpers + the generic base has a defined, safe answer.

### Commits
- **`SQLHelper` API** — `supportsWindowFrame(mode, hasValueOffset, dateOffset)` (base: ROWS + RANGE-peer only, gated on `WINDOW_FUNCTION`; denies RANGE value/date + GROUPS) and `formatWindowFrameInterval(offset, unit)`. **PostgreSQL opts in fully** (GROUPS gated ≥11) — the byte-parity anchor.
- **`AssetQuery`** — `framePushable` consults `helper.supportsWindowFrame(...)` (derives mode / value-offset / date-offset from the `WindowExpressionRef`); the hardcoded `"postgresql"` + `PG_PUSHABLE` gate is retired.
- **Rendering** — `WindowExpressionRef` emits a canonical `WINFRAME_INTERVAL(n,unit)` token for a date offset; `PreAssetQuery.getExpressionColumn` expands it per-dialect via `formatWindowFrameInterval`. Numeric/frame-less output unchanged.
- **Opt-in overrides** — Oracle / MySQL / DB2 / Databricks / Vertica / BigQuery, each with its verified capability + INTERVAL literal (`INTERVAL '7' DAY` / `INTERVAL 7 DAY` / …). DB2 & BigQuery deny date-offset (unverified / unsupported) → in-memory. GROUPS denied for all non-PG.
- **Conservative tier** — SQL Server, Snowflake, Clickhouse, Hive, Impala, Exasol + the long tail inherit the base (ROWS + RANGE-peer); Access/Mongo/Derby have no window functions → all-deny. Informix (which `extends DB2SQLHelper`) is overridden back to the conservative row so it doesn't inherit DB2's opt-in.
- **Portability fix** — `quarter` / `week` are normalized to exact equivalents (3×month / 7×day) so every dialect's INTERVAL literal uses only the universally-valid ANSI fields (also closes a latent Postgres `quarter` gap).

### Capability matrix (encoded per helper)
| Tier | Dialects | Frame pushdown |
|---|---|---|
| Full incl. GROUPS | PostgreSQL 11+ | ROWS / RANGE(peer+num+date) / GROUPS |
| RANGE value + date | Oracle, MySQL 8+, Databricks, Vertica | ROWS / RANGE(peer+num+date) |
| RANGE numeric only | DB2, BigQuery | ROWS / RANGE(peer+num); date → in-memory |
| ROWS + RANGE-peer | SQL Server, Snowflake, Clickhouse, Hive, Impala, Exasol, + long tail | value/date/GROUPS → in-memory |
| No window functions | Access, Mongo, Derby | all → in-memory |

### Byte-parity & verification
- **Postgres byte-parity:** capability set == Phase-4 `PG_PUSHABLE`; interval literal == base `INTERVAL '<n> <unit>'`; ROWS/frame-less/numeric-RANGE untouched.
- Unit: `WindowFrameCapabilityTest` 26/26 (all 23 dialects + base), routing + rendering suites green.
- Live (Postgres, `local-996`): date-RANGE / numeric-RANGE / GROUPS return correct values end-to-end with no raw-token leak (proving the token→dialect-render seam works on the real pushdown path). The whole-branch review traced that no raw token can reach the DB.
- **Deferred (documented):** live pushdown verification on non-Postgres dialects (needs DB instances — doc + SQL-string verified here); EXCLUDE / NULLS / fractional offsets remain out of scope.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/AssetQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/AssetQuery.java
@@ -2972,8 +2972,8 @@ public abstract class AssetQuery extends PreAssetQuery {
    /**
     * Like {@link #buildWindowSpecs(XTable, ColumnSelection)}, but SKIPS a window column whose
     * frame is pushable on {@code helper}'s dialect (see
-    * {@link #framePushable(String, SQLHelper)}) — i.e. one already merged into the pushed-down
-    * SQL by {@link #mergeColumn}. Used by the mixed-pushability routing case in
+    * {@link #framePushable(WindowExpressionRef, SQLHelper)}) — i.e. one already merged into the
+    * pushed-down SQL by {@link #mergeColumn}. Used by the mixed-pushability routing case in
     * {@link #getWindowTableLens}: when a query is mergeable but not every window column's frame
     * is pushable (e.g. one ROWS column + one RANGE column on a non-Postgres source), the ROWS
     * column is already in the generated SQL and must NOT also be recomputed here — building a
@@ -3013,7 +3013,7 @@ public abstract class AssetQuery extends PreAssetQuery {
 
          WindowExpressionRef w = (WindowExpressionRef) bref;
 
-         if(skipPushable && framePushable(w.getFrameMode(), helper)) {
+         if(skipPushable && framePushable(w, helper)) {
             // Already merged into the pushed-down SQL by mergeColumn — do not also compute it
             // in memory (see javadoc above).
             continue;
@@ -3186,8 +3186,9 @@ public abstract class AssetQuery extends PreAssetQuery {
     * dialect — the OVER is already in the SELECT (byte-parity: {@link #buildWindowLens} is not
     * even reached in this case). For a mergeable source where SOME but not all window columns'
     * frame modes (RANGE/GROUPS) are pushable on the source dialect (see
-    * {@link #framePushable(String, SQLHelper)}), the pushable ones are already merged into the
-    * SQL by {@link #mergeColumn} — only the non-pushable ones still need computing, so this
+    * {@link #framePushable(WindowExpressionRef, SQLHelper)}), the pushable ones are already
+    * merged into the SQL by {@link #mergeColumn} — only the non-pushable ones still need
+    * computing, so this
     * routes through the helper-aware {@link #buildWindowLens(TableLens, ColumnSelection, SQLHelper)}
     * overload, which builds specs ONLY for those (see {@link #buildWindowSpecs(XTable,
     * ColumnSelection, SQLHelper)}): a pushable column must not also be recomputed in memory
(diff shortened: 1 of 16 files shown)


larryliang-inetsoft:
## Review — Phase 5: dialect-aware frame pushdown

This is a well-executed change. The default-deny philosophy is the right call (wrong-high → loud DB syntax error; wrong-low → correct in-memory fallback), the per-dialect capability matrix is exhaustively unit-tested (26/26, all 23 helpers + base), and the byte-parity story holds: numeric/frame-less output is untouched, and the `WINFRAME_INTERVAL(..)` token piggybacks the existing `field['..']` rewrite pass so PostgreSQL renders the byte-identical Phase-4 literal. I traced the token seam end-to-end and confirmed no raw token can reach a DB on any pushable path. No correctness blockers. Findings are maintainability/robustness.

### ✅ Verified: the inheritance trap the Informix fix hints at is fully contained
The Informix override (don't inherit DB2's RANGE-value opt-in) raises the obvious question — does any *other* helper extend an opt-in dialect and silently inherit a capability + INTERVAL syntax never verified for it (the classic Redshift-extends-PostgreSQL trap)? I checked the class hierarchy: **`InformixSQLHelper` is the only subclass of any opt-in helper** (PostgreSQL/Oracle/MySQL/Vertica/Databricks/DB2), and it's explicitly overridden. No Redshift-on-PG subclass exists. So the risk is real but currently violated by nothing. 👍

### 🟠 A. The two-method coupling is unenforced (future-dialect footgun)
`supportsWindowFrame(mode, valueOffset, dateOffset=true) == true` **requires** the dialect to also override `formatWindowFrameInterval` — otherwise the base `INTERVAL '<n> <unit>'` (Postgres syntax) leaks into that dialect's SQL. Every current date-RANGE opt-in dialect (Oracle/MySQL/Databricks/Vertica) does override it, and PG deliberately uses the base — so today it's consistent. But the default-deny safety net only covers the *capability* axis; the *render* axis has no guard. A future dialect that opts into date RANGE but forgets the render override silently emits Postgres-syntax INTERVAL → DB syntax error (loud, at least, not silent-wrong — consistent with the philosophy, just not covered by it).

Recommend a coupling test in `WindowFrameCapabilityTest`: iterate the known helpers; for any where `supportsWindowFrame("RANGE", true, true)` is true, assert `formatWindowFrameInterval(7,"day")` is **not** the bare base literal unless the helper is PostgreSQL (the one intentional base-inheritor). That turns "reviewer must remember" into a failing build.

### 🟢 B. `WINFRAME_INTERVAL(` token namespace collision (low)
`expandWindowFrameTokens` scans **all** expression-column text (it runs for every `getExpressionColumn`, not just window columns — correctly early-returns on the `contains` check, so no byte-parity impact for normal columns). Unlike `field['..']` (an internal encoding users never type), a `sql:true` expression column is user-authored free text; one that literally contains `WINFRAME_INTERVAL(7,day)` would be rewritten. Astronomically unlikely, but it's the one token in this scheme sharing a namespace with user input. A more collision-proof sentinel (e.g. a ``-delimited or `__WIZ_WINFRAME__`-style marker) would remove the theoretical hole. Non-blocking.

### 🟢 C. `expandWindowFrameTokens` malformed-token handling is silent (low/defensive)
The `break` on a missing `)`/`,` (and `Integer.parseInt` on the offset) leaves the raw `WINFRAME_INTERVAL(...` text in the emitted SQL → DB syntax error rather than a named failure. Unreachable from the real emitter (`frameOffsetSql` always produces well-formed tokens), so defensive only — but if it ever fires it fails opaquely. A one-line "malformed window-frame token" throw on the `break` paths would match the fail-loud tone of the rest of the feature.

### 🟢 D. Byte-parity claim has a benign exception for PostgreSQL < 11
The PR says "capability set == Phase-4 `PG_PUSHABLE`." Phase 4 pushed GROUPS to *every* PG version; Phase 5 gates it `>= 11` (permissive when version unknown). So for a *resolvable* PG 10 server, Phase 4 pushed GROUPS → runtime error, while Phase 5 falls back to in-memory → correct. That's an improvement, not a regression — just worth stating the claim is "byte-identical for PG ≥ 11 / unknown version; PG < 11 GROUPS now correctly falls back instead of erroring." Doc-only.

### Nit
- `DatabricksHelper.java` — no newline at end of file.

### Verdict
Strong, mergeable. No correctness issues found; the inheritance risk is verified-contained. **A** is the one I'd act on (cheap test that enforces the capability↔render coupling the whole scheme depends on); B/C/D are optional hardening/doc.

PR #4238, opened 2026-07-12
test(sqlhelper): enforce window-frame capability<->interval-render coupling (PR #4237 finding A)

larryliang-inetsoft:
Test-only follow-up to PR #4237 (Phase 5 review finding A). Adds a regression lock: any dialect whose `supportsWindowFrame("RANGE", true, true)` is true MUST render a non-base INTERVAL literal (i.e. override `formatWindowFrameInterval`) — the sole explicit exemption is PostgreSQL, whose own syntax IS the base literal. Turns the hand-maintained capability↔render coupling into a failing build, so a future date-RANGE opt-in dialect that forgets its interval renderer is caught at build time instead of emitting Postgres-syntax INTERVAL to that DB at runtime.

Green on current code (Oracle/MySQL/Databricks/Vertica exercised; DB2/BigQuery/Informix/base skipped as date-denied; PG confirmed sole exempt). 27/27. No production change.

### DIFF
--- a/core/src/test/java/inetsoft/uql/jdbc/WindowFrameCapabilityTest.java
+++ b/core/src/test/java/inetsoft/uql/jdbc/WindowFrameCapabilityTest.java
@@ -355,4 +355,67 @@ class WindowFrameCapabilityTest {
       assertFalse(h.supportsWindowFrame("RANGE", true, true));
       assertFalse(h.supportsWindowFrame("GROUPS", false, false));
    }
+
+   // ── PR #4237 review finding A: capability<->render coupling ─────────────────────────────
+   // supportsWindowFrame("RANGE", true, true) (date-offset RANGE pushdown) and
+   // formatWindowFrameInterval(...) (the dialect's INTERVAL literal renderer) are two separate
+   // methods maintained by hand; nothing in the compiler ties them together. If a future dialect
+   // opts into date-RANGE pushdown but forgets to override formatWindowFrameInterval, it would
+   // silently fall through to the base INTERVAL '<n> <unit>' literal -- which happens to be
+   // Postgres/ANSI syntax -- and emit a DB syntax error at query time instead of failing a build.
+   // PostgreSQLHelper is the one dialect that legitimately opts in AND legitimately inherits the
+   // base literal (its own syntax IS the base syntax), so it is the sole, explicit allow-list
+   // exemption. Every other date-RANGE opt-in dialect must render something other than the base
+   // literal.
+   @Test
+   void dateRangeOptIn_requiresNonBaseIntervalRenderer_exceptPostgres() {
+      String baseLiteral = new SQLHelper().formatWindowFrameInterval(7, "day");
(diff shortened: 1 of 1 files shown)


larryliang-inetsoft:
## Review — capability↔render coupling test (#4237 finding A)

Thanks for picking this up. The mechanism is right: iterate helpers, skip non-opt-in ones, assert every date-RANGE opt-in dialect renders a non-base literal, with PostgreSQL as an explicit allow-listed exemption whose opt-in-and-inherits-base is *positively* asserted outside the loop (not silently skipped). The failure message names the missing override and explains the runtime consequence. Good.

### 🟠 The hand-maintained helper array reintroduces the exact failure mode the test exists to prevent
The test iterates a hardcoded `SQLHelper[] { ...11 helpers... }`. The whole point of finding A was to catch a **future** dialect that opts into date-RANGE pushdown but forgets its `formatWindowFrameInterval` override. But if that future dialect isn't also added to this array, the test never instantiates it and the coupling is never checked — so the regression lock only fires if the author remembers to add the new dialect to the list, which is the same "remember to do X" step the test is meant to make unnecessary. The guard is only as complete as a list that has to be maintained in lockstep with the thing it guards.

**There's already an enumerable registry to iterate instead.** `SQLHelper.getSQLHelper(String dbType)` resolves helpers from `helperTable`, loaded from `core/src/main/resources/inetsoft/uql/sqlhelper.xml` — 25 declared `type=` entries covering every dialect (access, oracle, db2, informix, sql server, sybase, mysql, derby, ingres, postgresql, vertica, hive, denodo, impala, h2, dremio, snowflake, google bigquery, mongo, luciddb, exasol, databricks, clickhouse, default, …). Driving the test off that list — parse the `<helper type=...>` entries (or expose `helperTable`'s keyset to test scope) and `SQLHelper.getSQLHelper(type)` each one — makes a newly-added dialect **automatically** covered: the day someone adds a date-RANGE opt-in helper, this test exercises it with zero test edits, which is what "turns it into a failing build" actually requires. It also uses the same resolution path production uses, so the test can't drift from what really gets instantiated.

If full registry iteration is deemed too broad for a unit test, at minimum add a guard that asserts the hardcoded array's size/coverage against the registry (fail if a declared helper type isn't represented), so an unlisted dialect trips *something*.

### Minor
- The array mixes date-denied helpers (DB2, BigQuery, Informix, SQL Server, Snowflake) that just hit the `continue` — harmless, but it reads as an arbitrary sample rather than a complete set, which is the tell for the gap above. Registry iteration removes the arbitrariness.
- `assertNotEquals(baseLiteral, render)` proves an override exists, but doesn't validate the rendered syntax is *correct* for the dialect (that's covered by the per-dialect tests already in this file — fine to leave the coupling test focused on "override present").

### Verdict
Correct and green as written, and it does lock the coupling for today's dialects. The one change worth making before merge: iterate the `sqlhelper.xml` registry (or assert the array covers it) so the lock genuinely extends to future dialects — otherwise it protects only the set someone remembered to list, which is the very thing finding A was about. Test-only, no production risk either way.

PR #4242, opened 2026-07-13
fix(wiz): resolve FIELD condition operand to a DataRef (was JS ExpressionValue → silent match-all)

### LINKED ISSUE #4241: Worksheet condition: FIELD operand is coerced to a JS ExpressionValue (bare column name) instead of a DataRef → filter matches all rows
## Summary

In a worksheet condition (`/ws/table` `preAggregateCondition` / `postAggregateCondition`), an operand whose `type` is **`FIELD`** (another column) or **`EXPRESSION`** (a JS/formula operand) is **silently dropped**: StyleBI accepts the condition leaf but fails to deserialize the operand into a real column/formula reference, storing `values: []`. The resulting condition has no right-hand operand and **matches every row** — a plausible-but-wrong no-op with **no error**.

`VALUE` (literal) and `SUBQUERY` operands round-trip and evaluate correctly; only `FIELD` and `EXPRESSION` are affected.

## Reproduction

Create a physical table with a `preAggregateCondition` using a `FIELD` operand:

```jsonc
{
  "tableName": "T",
  "tableType": "physical table",
  "physicalSource": { "datasourcePath": "suitecrm", "schema": "public", "tableName": "opportunities" },
  "columns": [ { "name": "amount", "type": "double" } ],
  "preAggregateCondition": [
    { "field": "amount", "operation": "GREATER_THAN", "values": [ { "type": "FIELD", "value": "amount" } ] }
  ]
}
```

`get_worksheet_structure` / the stored model shows the condition as:

```json
{ "field": "amount", "operation": "GREATER_THAN", "values": [], "phase": "pre" }
```

i.e. the `FIELD` operand is gone. Materializing the table returns **all** rows (the `amount > amount` filter that should return 0 was dropped). The same happens for `EXPRESSION` (e.g. `{ "type": "EXPRESSION", "value": "field['probability']" }`) — verified on a plain physical column, so it is **not** window/mirror-specific.

## Impact

Column-vs-column comparisons (e.g. "opportunities whose amount is above their own stage's average", after an `AVG(...) OVER (PARTITION BY stage)` window column) and computed-operand comparisons cannot be expressed via the condition tree — and worse, they **appear to succeed** while returning unfiltered data.

## Expected

Either deserialize `FIELD`/`EXPRESSION` operands into real column/formula references and apply them, **or** reject the condition with a clear error. Silently dropping the operand and matching all rows is the harmful outcome.

## Interim mitigation (wiz layer)

wiz-services now **fails loud** at validation when a worksheet condition uses a `FIELD` or `EXPRESSION` operand, pointing callers to a `sql query table` (`WHERE amount > stage_avg`) — the reliable path for column-vs-column / computed comparisons — instead of forwarding a condition that silently no-ops. This issue tracks the StyleBI-side fix so the operands either work or reject server-side. (See stylebi-wiz worksheet-condition validation change.)

larryliang-inetsoft:
Fixes #4241.

## Problem

On the wiz worksheet-builder endpoint `/ws/table`, a condition leaf operand of type **`FIELD`** (compare the condition column to another column, e.g. `amount > stage_avg`) is mishandled. `WorksheetTableService.addConditionValue`'s `FIELD` case wrapped the referenced column name in a **JavaScript `ExpressionValue`**:

```java
case "FIELD" -> {
   ExpressionValue ev = new ExpressionValue();
   ev.setExpression(v.getValue()...toString());   // bare column name, e.g. "stage_avg"
   ev.setType(ExpressionValue.JAVASCRIPT);
   condition.addValue(ev);
}
```

The bare column name is an **undefined JS identifier**, so the comparison never became a column reference — the filter silently matched **all rows** (a plausible-but-wrong result, no error). It also round-tripped back as the wrong type (`reverseValue` maps any `ExpressionValue` → `"EXPRESSION"`).

## Fix

Resolve the referenced name to a `DataRef` against the table's `ColumnSelection` and add that as the operand value — the same mechanism `appendRankingConditionItem` already uses (`columns.getAttribute(...)`). `Condition` already serializes a `DataRef` operand via its `isfield="true"` branch, and `reverseValue` (~L2314) maps a `DataRef` back to `"FIELD"`, so it round-trips correctly. When the referenced column can't be resolved, fail loud with `IllegalArgumentException` (matching `mapOperation`'s unknown-operation handling) rather than re-introducing a silent no-op.

`EXPRESSION` (→ `ExpressionValue`) and `SESSION_DATA` (→ `UserVariable`) were **not** affected — only `FIELD` was mis-typed.

## Verification

Built as image `local-997` and tested live (suitecrm): `amount > FIELD(stage_avg)` layered over an `AVG(amount) OVER (PARTITION BY sales_stage)` window mirror now filters to **104 / 200** opportunities (those above their own stage's average), matching the equivalent `sql query table` (`WHERE amount > stage_avg`) baseline. Before the fix the same condition returned all 200.

The wiz-services plugin currently guards FIELD operands (fail-loud → point to a sql query table) as an interim; once this ships in the deployed baseline that guard can be removed so column-vs-column conditions work natively.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
@@ -1788,11 +1788,22 @@ public class WorksheetTableService {
          }
 
          case "FIELD" -> {
-            ExpressionValue ev = new ExpressionValue();
-            String expressoin = v.getValue() != null ? v.getValue().toString() : "";
-            ev.setExpression(expressoin);
-            ev.setType(ExpressionValue.JAVASCRIPT);
-            condition.addValue(ev);
+            // A FIELD operand compares the condition column against ANOTHER column (e.g.
+            // amount > stage_avg). Resolve the referenced column name to a DataRef against the
+            // table's ColumnSelection (same mechanism as appendRankingConditionItem) and add that
+            // as the operand value: Condition serializes a DataRef operand via its isfield branch,
+            // and reverseValue maps it back to "FIELD". Previously this wrapped the bare column name
+            // in a JavaScript ExpressionValue (an undefined identifier), so the comparison never
+            // became a column reference and the filter silently matched ALL rows.
+            String name = v.getValue() != null ? v.getValue().toString() : null;
+            DataRef ref = name != null ? columns.getAttribute(name) : null;
+
+            if(ref == null) {
+               throw new IllegalArgumentException(
+                  "FIELD condition operand references unknown column \"" + name + "\".");
+            }
+
+            condition.addValue(ref);
          }
 
          case "SUBQUERY" -> {
--- a/core/src/test/java/inetsoft/web/wiz/service/WorksheetTableServiceConditionTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/service/WorksheetTableServiceConditionTest.java
@@ -22,10 +22,13 @@ import inetsoft.test.BaseTestConfiguration;
 import inetsoft.test.ConfigurationContextInitializer;
 import inetsoft.test.SreeHome;
 import inetsoft.uql.ColumnSelection;
+import inetsoft.uql.Condition;
 import inetsoft.uql.ConditionList;
 import inetsoft.uql.asset.ColumnRef;
+import inetsoft.uql.asset.ExpressionValue;
 import inetsoft.uql.asset.Worksheet;
 import inetsoft.uql.erm.AttributeRef;
+import inetsoft.uql.erm.DataRef;
 import inetsoft.web.wiz.model.WorksheetTable;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft:
## Review — resolve FIELD condition operand to a DataRef (fixes #4241)

This is the StyleBI-side root-cause fix for the silent match-all bug the wiz guard (#1197) currently papers over. The diagnosis is exactly right — wrapping a bare column name in a `JAVASCRIPT` `ExpressionValue` made it an undefined JS identifier, so the operand evaluated to nothing and the filter matched all rows. I verified every claim against the source:

- **`columns` in scope** is the `ColumnSelection` threaded into `addConditionValue` (same object the ranking path receives). ✓
- **The cited precedent is real:** `appendRankingConditionItem` resolves via `columns.getAttribute(item.getField())` then builds a `ConditionItem(ref, …)`. ✓
- **Round-trip is self-consistent:** `reverseValue` maps a `DataRef` operand → `"FIELD"` (L2326) via `refName(ref)` = `ref.getName()`, which is the exact key `getAttribute(name)` matches on — so store→reload can't drift. ✓
- **Idiomatic:** `cs.getAttribute(name)` is the universal resolution convention throughout this service (group-by, joins, window partitionBy/orderBy all use it), so this isn't a one-off. ✓
- **`EXPRESSION`/`SESSION_DATA` untouched** — confirmed, only the `FIELD` arm changed. ✓

### 🟠 1. No unit test for the fix (or the new throw) — live-only verification
The changeset is the one source file; there's no test. Every sibling PR in this series added one, and there's a ready seam — `WorksheetTableServiceWindowColumnsTest` already stands up a `PhysicalBoundTableAssembly` + `ColumnSelection` and calls the service directly. Two cheap cases would lock this against regression in CI (the live `local-997` run isn't repeatable there):
- a `FIELD` operand resolves to a `DataRef` and the built `AssetCondition`'s value is that ref (not an `ExpressionValue`) — the exact thing that silently regressed before;
- an unknown `FIELD` column throws `IllegalArgumentException` naming the column.

This is the finding I'd want addressed before merge — the bug was a silent-wrong-result that shipped undetected once already; only a test makes "it's a DataRef now, not a JS expr" a permanent guarantee.

### 🟢 2. Fail-loud throw diverges from the "same mechanism" it cites
The comment says it mirrors `appendRankingConditionItem`, but that method **returns silently** on an unresolved column (`if(ref == null) return;`), whereas this correctly **throws**. The throw is the better choice (a silent skip here would just re-introduce a no-op operand — the very bug). Worth either softening the "same mechanism" wording or, better, changing the ranking path's silent `return` to throw as well, so an unresolved ranking column isn't itself a silent drop. Optional, but the inconsistency cuts the other way from how the comment reads.

### 🟢 3. Resolution is name-exact — residual risk for qualified/grouped FIELD refs
`getAttribute(name)` matches on `ref.getName()`, not the richer entity/header matching `AssetUtil.findColumn` provides (the resolver `AssetQuery.buildWindowSpecs` had to switch to for `Month(…)`/qualified columns). Live-verified for the simple alias `stage_avg`; a qualified (`MIRROR.col`) or grouped (`Month(OPPS.date_closed)`) FIELD operand *may* not resolve. Two mitigations already make this low-risk: it's now a **loud throw**, not a silent match-all (strictly better than before), and it's consistent with the whole service's convention. Flagging only so it's a known boundary — if column-vs-column on a grouped/qualified key is in scope, confirm the name form wiz sends resolves, or consider `AssetUtil.findColumn` for parity with the window resolver.

### Coordination
The wiz `#1197` FIELD guard must be removed **only after** this lands in the deployed baseline — otherwise either FIELD stays double-blocked (guard still up) or, if the guard is pulled first, the silent match-all returns. The PR body notes this correctly; just calling it out as a sequencing dependency across the two repos.

### Verdict
Correct, minimal, idiomatic root-cause fix, verified live. The one thing I'd block on is **#1 (add the unit test)** — a silent-wrong-result bug that already shipped once deserves a regression lock, especially when the test seam already exists. #2/#3 are minor.

larryliang-inetsoft:
### 🟠 1 — done (the blocker)
Added two regression tests via the existing `buildConditionList` seam (`WorksheetTableServiceConditionTest`), pushed as `73c4d1bbf`:
- `fieldOperandResolvesToDataRefNotExpressionValue` — asserts the built condition's operand is a `DataRef` (the referenced column) and **not** an `ExpressionValue`;
- `fieldOperandUnknownColumnFailsLoud` — asserts an unresolvable FIELD column throws `IllegalArgumentException` naming the column.

I ran them **both ways**: on the pre-fix code they FAIL (operand still an `ExpressionValue`; no throw) and with the fix they PASS (5/5, `BUILD SUCCESS`) — so it's a genuine CI lock on the exact regression, not just a green rubber-stamp.

### 🟢 2 — noted; the throw is deliberate
Fair point on the asymmetry. To be precise: the code comment cites `appendRankingConditionItem` only for the **resolution** mechanism (`columns.getAttribute`), not for the null handling — I intentionally throw here (a silent skip would re-introduce the no-op operand that is this very bug). You're right that `appendRankingConditionItem`'s own `if(ref == null) return;` is itself a latent silent-drop (an unresolved ranking column vanishes). I've left that out of scope for this PR to keep it to the one bug, but it's worth a follow-up to make the ranking path throw too.

### 🟢 3 — known boundary, now safe
Resolution is name-exact (`getAttribute`), consistent with the whole service. For the verified path wiz sends the bare mirror column name (`stage_avg`), which resolves. A qualified/grouped FIELD ref (`MIRROR.col`, `Month(...)`) may not resolve via `getAttribute` — but it now **throws loudly** (naming the column) rather than silently matching all rows, so it's strictly better than before. If column-vs-column on a grouped/qualified key becomes a real use case, switching to `AssetUtil.findColumn` for parity with the window resolver is the follow-up; flagging it as a documented boundary for now.

### Coordination
Agreed — the wiz `#1197`/`#1200` FIELD guard stays up until this lands in the deployed baseline; only then does the wiz side drop the FIELD reject so column-vs-column works natively. Sequencing tracked.

PR #4243, opened 2026-07-13
fix(query): window-expression column mergeable when UniformSQL not yet built (intermittent JS-eval fallback)

### LINKED ISSUE #4240: Worksheet: sql:true expression over a windowColumns mirror is intermittently evaluated by the JS engine instead of pushed to SQL
## Summary

When a worksheet mirror table adds a **SQL expression column** (`sql:true` — e.g. `CASE WHEN … END`, `COALESCE(…)`, `… IS NULL`) on top of a base table that carries **window columns** (a mirror built with `windowColumns`, i.e. `ROW_NUMBER/RANK/NTILE/LAG/LEAD/SUM(...) OVER (...)`), StyleBI **sometimes** evaluates that expression with its JavaScript/Rhino engine instead of pushing it down as SQL. When that happens, the SQL text fails to parse and the table build fails with:

```
JavaScript error: SyntaxError: <eval>:1:5 Expected ; but found WHEN
```

i.e. valid SQL is fed to the JS parser. The behaviour is **intermittent**, not deterministic.

## Reproduction (worksheet `/ws/table`)

1. Physical table `P` — one numeric column `amount`.
2. Mirror `W` over `P` with a window column, e.g. `lg = LAG(amount) OVER (ORDER BY amount ASC)` (or `NTILE(4) OVER (ORDER BY amount)`).
3. Mirror `M` over `W` with a SQL expression column:
   `guarded = CASE WHEN field['lg'] IS NULL THEN 0 ELSE field['lg'] END` (`sql: true`).

Building `M` sometimes succeeds (expression pushed down: `SELECT CASE WHEN … FROM (<window subquery>)`), and sometimes fails with the JS `SyntaxError` above.

## Observed characteristics

- **Base-type dependent:** the same `sql:true` `CASE` expression over a **physical** base and over an **aggregate** (group-by) mirror **always pushed down cleanly**. Only the **windowColumns** base exhibited the failure.
- **Non-deterministic:** the identical physical → window → sql-CASE sequence failed 3 times and then succeeded 3 times in a row within one session, with no change to the request. This points to a materialization/planning-timing effect rather than a fixed rule (consistent with the MV/result-cache behaviour on demo tables — a cold vs. warm base appears to change the push-down decision).

## Expected

A `sql:true` expression over a window-column mirror should be **consistently** pushed to SQL as an outer subquery over the window result (`SELECT <expr> FROM (<window OVER … subquery>)`) — the same way it already is for aggregate-mirror bases — rather than falling back to in-memory Rhino evaluation of SQL text.

## Likely area

`AssetQuery` expression-column handling / the mergeability decision when a base assembly already contains a window-function expression column — whether the outer expression can be layered into a subquery vs. must be evaluated in-memory. The inconsistency suggests the decision is sensitive to whether the base is already materialized/cached.

## Interim mitigation (wiz layer)

The wiz-services worksheet layer now **translates** this specific failure (request asked for `sql:true` **and** StyleBI returned a JS-eval error) into an actionable message pointing at the reliable workarounds — write the expression as JavaScript (`sql:false`, e.g. a ternary instead of `CASE WHEN`), or compute it inside a `sql query table` — while preserving the raw StyleBI error. That is error-text only and does **not** address the root cause; this issue tracks the StyleBI-side fix. (See stylebi-wiz PR #1195.)

larryliang-inetsoft:
Fixes #4240.

## Problem

A worksheet MIRROR with a `sql=true` expression column (e.g. `CASE WHEN field['prev'] IS NULL THEN 0 ELSE field['prev'] END`), layered over a BASE mirror that has a `windowColumns` column (`WindowExpressionRef`, e.g. `LAG(...) OVER (...)`), **intermittently** fell back to in-memory Rhino evaluation of the raw SQL text instead of SQL pushdown — throwing `JavaScript error: SyntaxError: <eval>:1:5 Expected ; but found WHEN`. Same request, sometimes clean pushdown, sometimes the crash.

## Root cause

`PreAssetQuery.isAttributeMergeable` (`report/composition/execution/PreAssetQuery.java`), `SQLExpressionRef` branch:

```java
UniformSQL sql = getUniformSQL();
if(sql == null) {
   return false;   // "SQL not built yet" treated as "not mergeable"
}
SQLHelper helper = getSQLHelper(sql);
sexp.setDBType(helper.getSQLHelperType());   // dbtype needed by most SQLExpressionRefs...
...
return sexp.isMergeable();
```

`getUniformSQL()` is null on an early mergeability probe and non-null later (it's lazily built — `MirrorQuery`/`JoinQuery`/`BoundQuery` build `nquery` then expose its `UniformSQL`). The `sql==null → false` is correct for **DateRangeRef / NumericRangeRef / NamedRangeRef** — their `isMergeable()` reads `dbtype` (e.g. `!dbtype.equals("access")`, `"derby"→false`), which comes from the helper — but a **`WindowExpressionRef` is unconditionally mergeable and DB-type-independent** (its `OVER(...)` with `field['..']` tokens is always rewritable). Reporting it non-mergeable during a null-SQL probe froze the whole base query as non-mergeable (`isQueryMergeable` memoizes the first answer and never resets), cascading through `MirrorQuery.isSourceMergeable` into the JS-formula fallback. Whether the first probe hit before or after the base's SQL was built decided the outcome → the intermittency.

## Fix

In the null-`UniformSQL` branch, return `true` for a `WindowExpressionRef` (mergeable regardless of build timing / dbtype); keep the conservative `false` for the dbtype-dependent SQLExpressionRefs:

```java
if(sql == null) {
   return exp instanceof WindowExpressionRef;
}
```

Blast radius is that single branch. DateRange/NumericRange/NamedRange behavior is unchanged. (A separate latent issue — `isQueryMergeable` ignoring its `sub` arg in the memo — only misbehaves when `isSQLMergeable()==false`, a deliberate force-in-memory config, and is **not** the intermittent window symptom; left as-is.)

## Verification (image `local-998`, suitecrm)

- **11 / 11** independent `sql=true` CASE-over-window mirrors pushed down (`isSqlExpression=true`), **0** JS-eval failures (was ~50%).
- A materialized case computes correctly (`gK` = LAG(amount) per row).
- A month-grouped aggregate (`DateRangeRef`) still returns all 12 months — no regression on the conservative path.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/PreAssetQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/PreAssetQuery.java
@@ -1629,7 +1629,15 @@ public abstract class PreAssetQuery implements Serializable, Cloneable {
          UniformSQL sql = getUniformSQL();
 
          if(sql == null) {
-            return false;
+            // The UniformSQL may not be built yet on an early mergeability probe. Most
+            // SQLExpressionRefs (DateRangeRef/NumericRangeRef/NamedRangeRef) can't decide without the
+            // dbtype the helper supplies, so stay conservative for them. A WindowExpressionRef is
+            // unconditionally mergeable and DB-type-independent (its OVER(...) with field['..'] tokens
+            // is always rewritable), so a null-SQL probe must not report it non-mergeable — doing so
+            // froze the whole base query as non-mergeable depending on probe/build ordering, making a
+            // sql=true expression layered over a window mirror intermittently fall back to in-memory
+            // JS evaluation instead of SQL pushdown.
+            return exp instanceof WindowExpressionRef;
          }
 
          SQLHelper helper = getSQLHelper(sql);


larryliang-inetsoft:
## Review — window-expression column mergeable on null-UniformSQL probe (fixes #4240)

This is the root-cause fix for the intermittent JS-eval fallback that PR #1195 could only *translate the error for* — good to see it chased to the actual mechanism. The diagnosis (a memoized-first-answer `isQueryMergeable` freezing the whole base query as non-mergeable when an early null-SQL probe hits before the base's `UniformSQL` is lazily built → probe/build ordering decides push-down vs. Rhino → intermittency) is thorough and matches the code.

I verified the fix is not just plausible but *provably consistent*, against the PR-head source:

- **`WindowExpressionRef extends ExpressionRef implements SQLExpressionRef`** — it genuinely reaches this branch, and `exp instanceof WindowExpressionRef` is a valid narrowing (not dead code). ✓
- **`WindowExpressionRef.isMergeable()` returns `true` unconditionally** (L560–562). So on the `sql != null` path this branch already returns `true` for a window ref via `sexp.isMergeable()`. The fix makes the `sql == null` early-probe path return the **same** `true` — it removes the ordering-dependent disagreement between the two paths rather than introducing a new answer. That's the correct framing: for the one ref type whose mergeability is genuinely constant, the early probe should match the eventual answer. ✓
- **Zero over-report risk:** a `WindowExpressionRef` is never legitimately non-mergeable, so returning `true` early can only *fix* under-reporting; the dbtype-dependent refs (DateRange/Numeric/Named) still return the exact conservative `false` because they aren't `WindowExpressionRef`. Blast radius is provably that one sub-branch. ✓
- **No interaction with the Phase-4/5 frame gate:** `isAttributeMergeable` decides "is this expression SQL-rewritable," which is orthogonal to `AssetQuery.mergeColumn`'s per-column `framePushable` capability gate. A GROUPS-on-non-PG window is still routed in-memory by that gate even though the query is now deemed mergeable — the two are complementary (the mixed-pushability path from #4237). No regression there. ✓

### 🟢 1. The deeper fragility (memoized, probe-order-dependent mergeability) remains — acknowledged
The fix corrects the *one* ref type that triggered the symptom, but the underlying hazard is `isQueryMergeable` memoizing a first answer that depends on whether SQL was built yet. Today `WindowExpressionRef` is the only SQLExpressionRef whose `isMergeable()` is unconditionally true, so the targeted fix is sufficient — but any future constant-true SQLExpressionRef added without the same early-probe carve-out would reintroduce the identical intermittency. The PR explicitly scopes out the related `isQueryMergeable`-ignores-`sub` memo issue, which is a reasonable boundary; just flagging that this class of bug will recur until the memo itself is made probe-order-invariant (or the null-SQL branch returns `sexp.isMergeable()` for any dbtype-independent ref generically).

### 🟢 2. No regression test — but genuinely hard here, so a narrow lock is the ask
Unlike #4242 (a direct-call assertion), this bug is an *ordering race*, so reproducing the "probe before SQL built" timing deterministically is real work. That said, the fix's specific invariant is trivially testable without the race: a `PreAssetQuery` whose `getUniformSQL()` is null should report `isAttributeMergeable(col)` **true** for a column wrapping a `WindowExpressionRef` and **false** for one wrapping a `DateRangeRef`. Even that narrow test locks the exact line that regressed and documents the intended asymmetry — worth adding given the symptom was a ~50% intermittent production crash that no CI would currently catch.

### 🟢 3. Intermittency evidence
"11/11 pushed down, 0 JS-eval (was ~50%)" — for a coin-flip bug, 11 consecutive clean runs is ~1/2000 odds of being luck, so reasonably convincing, though not proof the ordering window is fully closed. Fine given the correctness argument above stands on its own (the two paths now provably agree for window refs).

### Cross-PR note
This should retire the specific intermittency that PR #1195's error-translation was written for. #1195 is still worth keeping as a safety net for *other* JS-eval causes, but once this lands its framing ("intermittent for sql:true over a window mirror") is largely obsolete for that path — worth a note there so the translation isn't read as "this still happens."

### Verdict
Minimal, idiomatic, and provably correct (the fix makes the early probe agree with the eventual answer for the only ref type whose mergeability is constant; zero over-report risk). No blockers. I'd add the narrow null-SQL-probe test (#2) since the original was a silent-ish intermittent crash, but it's not gating.

larryliang-inetsoft:
Thanks — agree on all three; verdict noted (no blockers).

### 🟢 2 — regression test: deferred, with rationale
I tried to add the narrow "null-`UniformSQL` probe ⇒ `isAttributeMergeable` true for a `WindowExpressionRef` col, false for a `DateRangeRef` col" test and it turns out to need a real harness: `PreAssetQuery` is abstract with 11 abstract methods and a constructor requiring an `AssetQuerySandbox`, and no existing test instantiates one (`AssetQueryTest` only exercises inner static lens classes). Standing that up — or a stub subclass past `findColumn`/`getBaseAttribute`/`isMergeableDataType` — is disproportionate for a one-line, provably-correct branch. So I've left it, resting on: (a) the correctness argument (the fix makes the null-SQL path return the **same** `true` the non-null path already returns for a window ref — the two paths now agree), and (b) the live 11/11 push-down (0 JS-eval, was ~50%) + DateRange-unaffected check. If a `PreAssetQuery` unit seam gets introduced, this is the first test to add.

### 🟢 1 — the class-of-bug hazard
Acknowledged and scoped out deliberately. The durable hardening (if we want to close it generically) is to have the null-`UniformSQL` branch return `sexp.isMergeable()` for any **dbtype-independent** `SQLExpressionRef` rather than special-casing `WindowExpressionRef` — but today `WindowExpressionRef` is the only constant-true impl, so the targeted fix is sufficient and lower-risk. Worth a follow-up if another such ref is ever added.

### Cross-PR
Agreed — once this lands it retires the specific intermittency `#1195`'s error-translation was written for. I'll note on `#1195` that its "sql:true over a window mirror" framing is obsolete for that path (keeping it as a safety net for other JS-eval causes).

PR #4246, opened 2026-07-13
75628, normal bar does not need to support bridge line

milotalon:
Root Cause:
Bridge lines are a waterfall-only feature (feature #75007), but they were implemented on IntervalElement, which is the element type shared by both waterfall charts and regular bar charts. The setBridgeLine(Color, GLine) method is a script-callable (@TernMethod) API, and the bridge-drawing logic in IntervalElement only checked whether a bridge line style had been set (bridgeLineStyle != null). As a result, a chart script such as graph.getElement(0).setBridgeLine(java.awt.Color.LIGHT_GRAY, GLine.DASH_LINE) run against a regular bar chart would draw bridge lines, even though bridge lines are meaningful only for waterfall charts.

Fix:
Made bridge rendering waterfall-only. Added a dedicated "waterfall" flag to IntervalElement (with isWaterfall()/setWaterfall()); the graph generator sets it to true only when constructing waterfall elements (the intermediate bar and the sum bar). The bridge-drawing logic is now gated on (bridgeLineStyle != null && waterfall), so calling setBridgeLine on a non-waterfall bar chart is a no-op. Existing waterfall bridge-line behavior (including the facet-graph fix #75624) is fully preserved. The flag was also added to equalsContent for correctness.

### DIFF
--- a/core/src/main/java/inetsoft/graph/element/IntervalElement.java
+++ b/core/src/main/java/inetsoft/graph/element/IntervalElement.java
@@ -206,10 +206,14 @@ public class IntervalElement extends StackableElement {
       IntervalGeometry firstNegative = null; // first negative segment added to the current bar's stack
       double posIntervalSum = 0; // cumulative interval for positive stack
       double negIntervalSum = 0; // cumulative interval for negative stack
-      double[] prevBridgeX = (bridgeLineStyle != null) ? new double[getVarCount()] : null;
+      // Bridge lines are a waterfall-only feature. setBridgeLine() is a script-callable
+      // (@TernMethod) API on IntervalElement, which is shared by regular bar charts and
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
Claude's automated review confirms the fix correctly scopes bridge-line rendering to waterfall charts only, with no regressions or backward-compatibility concerns identified. Missing test coverage is noted as non-blocking (consistent with the prior related fix). All checks pass. Approving.

jshobe-inetsoft (dismissed):
Content-wise this PR looks good per Claude's automated review, and I've approved it. However, it now has merge conflicts with epic-74519 (likely from #4248, which touches the same IntervalElement.java and just merged). Please rebase onto the latest epic-74519 and resolve the conflicts so it can be merged.

jshobe-inetsoft (approved):

PR #4248, opened 2026-07-13
75626, When the bar corner radius is set, the length of the bridge connection line changes

milotalon:
Root Cause:

The waterfall bridge connector endpoints were inset from each bar's center by a fixed formula, halfWidth = 0.40 - 0.5*cornerRadius, in IntervalElement.addBridgeForm. This had two problems:

1. The 0.40 constant did not match the bar's actual rendered width (bars render at about half the category slot, ~0.26 in slot units), so the endpoint sat well beyond the true bar edge.
2. Tying the inset to cornerRadius changed the line's length as the corner radius changed. Because a rounded bar's visible top/bottom edge tapers inward toward the connection point while its silhouette stays full-width at mid-height, the endpoint ended up either floating in a growing gap or grazing the rounded cap - which is what made the gap appear to change (and touch) when the corner radius was set.

Fix:

Reworked the bridge endpoint placement so it is derived from each bar's actual geometry rather than a fixed constant:

- Each end of a bridge is now inset by its own bar's half-width (mirroring BarVO.getBarSize, so it accounts for the size aesthetic), plus a constant gap. The two bars a bridge connects are inset independently, so the line adapts to different-size bars (a narrower bar yields a longer line) and one end can clear a wide bar while the other stays close to a narrow one.
- The endpoint always stays just outside the bar's full-width silhouette (with a minimum-gap floor), guaranteeing the line never touches the bar at any corner radius; it eases in slightly as the radius grows while keeping a uniform gap.
- Added a guard that skips a bridge when adjacent bars are wide enough to touch (no real gap to draw), avoiding a degenerate/reversed line.

Result: the gap between each bar and the bridge line is uniform and the line never touches the bar, at any corner radius.

### DIFF
--- a/core/src/main/java/inetsoft/graph/element/IntervalElement.java
+++ b/core/src/main/java/inetsoft/graph/element/IntervalElement.java
@@ -20,6 +20,7 @@ package inetsoft.graph.element;
 import com.inetsoft.build.tern.*;
 import inetsoft.graph.GGraph;
 import inetsoft.graph.aesthetic.GLine;
+import inetsoft.graph.aesthetic.SizeFrame;
 import inetsoft.graph.aesthetic.VisualModel;
 import inetsoft.graph.coord.*;
 import inetsoft.graph.data.DataSet;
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):
Claude's automated review found no blocking issues — the core geometry fix is sound, and the 5 observations raised (missing test coverage, dodge()/polar edge cases, defensive-code notes) are explicitly non-blocking/low-risk pre-existing limitations. All checks pass, no conflicts. Approving.

PR #4249, opened 2026-07-14
Add security phase2 v1

bonnieshi43:
1. add organization life cycle test and docs.
2. after bug fixed. update case and docs.


jshobe-inetsoft (approved):
Claude's follow-up review (after commit 5b15dc63) confirms both previously-flagged issues (cleanup gap, overly-broad catch) are resolved correctly, with no new concerns. Test/doc-only change, all checks pass. Approving.

PR #4251, opened 2026-07-14
fix(wiz): resolve the worksheet table name before browsing column values

larryliang-inetsoft:
## Summary
- `WizBrowseDataService.browseData()` passed the caller-supplied `assemblyName` (the VS chart's own presentation name, e.g. `"Chart1"`) straight through to `BrowseDataController.setName()`, but `process()` looks that name up against the base **worksheet's** assemblies — a different namespace (e.g. `"JOIN_SO_SOL"`). The chart's own name never resolves there, so `process()` silently returned `null` and the wiz `lookup_column_values` MCP tool always answered with an empty value list instead of erroring — for every column, on every chart, always.
- Fix resolves the real worksheet table name from the chart assembly's `SourceInfo` first (mirrors the same resolution `WizVsService` already does for aggregate-condition handling), and fails loud (`IllegalArgumentException`/`IllegalStateException`) when the chart assembly or its source table can't be found, instead of silently degrading to an empty result.
- Documents a known residual issue rather than silently leaving it: a genuinely unknown column name still returns an empty list instead of erroring, because the shared/legacy `BrowseDataController.findColumn()` silently inserts an unresolved name as a phantom column rather than failing. Not touched here since it's also used by the native composer UI — left as a follow-up.

## Test plan
- [x] `-pl community/core install` build succeeds (core-only change, no new Spring beans → AOT-safe).
- [x] Verified live via the wiz MCP plugin: `lookup_column_values` on a `windowColumns`/`dateGroupLevel`-derived column name (`"Month(JOIN_SO_SOL.date_order)"`) now returns the real 19 monthly values instead of an empty array.
- [x] Regression-checked a plain column (`"revenue"`) — still returns its 19 values correctly, no behavior change for the working case.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizBrowseDataController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizBrowseDataController.java
@@ -37,7 +37,9 @@ public class WizBrowseDataController {
     * the underlying worksheet.
     *
     * @param runtimeId           VS runtime ID (vsId)
-    * @param assemblyName        worksheet table/assembly name that owns the column
+    * @param assemblyName        the VS chart/table assembly's own presentation name (e.g.
+    *                            "Chart1") — the underlying worksheet table it's bound to is
+    *                            resolved internally from the assembly's {@code SourceInfo}
     * @param viewsheetIdentifier durable asset id, used to restore a reaped runtime (optional)
     * @param dataRefModel        column descriptor (classType, name, attribute, …)
     * @param principal           current user
--- a/core/src/main/java/inetsoft/web/wiz/service/WizBrowseDataService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizBrowseDataService.java
@@ -21,7 +21,11 @@ import inetsoft.analytic.composition.ViewsheetService;
 import inetsoft.cluster.*;
 import inetsoft.report.composition.*;
 import inetsoft.uql.asset.ColumnRef;
+import inetsoft.uql.asset.SourceInfo;
 import inetsoft.uql.erm.DataRef;
+import inetsoft.uql.viewsheet.DataVSAssembly;
+import inetsoft.uql.viewsheet.VSAssembly;
+import inetsoft.uql.viewsheet.Viewsheet;
 import inetsoft.web.binding.drm.DataRefModel;
 import inetsoft.web.composer.BrowseDataController;
 import inetsoft.web.composer.model.BrowseDataModel;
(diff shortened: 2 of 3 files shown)


larryliang-inetsoft:
## Review

### What it does
`WizBrowseDataService.browseData()` was passing the VS chart's own presentation name (e.g. `"Chart1"`) as the worksheet table name to `BrowseDataController.setName()`. `process()` does `ws.getAssembly(name)` against the worksheet's own table assemblies — a chart's presentation name never matches there, `assembly instanceof TableAssembly` fails, and `process()` returns `null`. That null silently became an empty value list, so the wiz `lookup_column_values` MCP tool always returned `[]` for every column, on every chart.

The fix resolves the chart's actual bound worksheet table from `DataVSAssembly.getSourceInfo().getSource()` before calling `setName()`, and raises `IllegalArgumentException`/`IllegalStateException` at each point that used to degrade silently (missing chart assembly, unbound source, `process()` returning null).

### Verification of the diagnosis
Traced this independently rather than taking the PR body on faith:
- `BrowseDataController.process()` does confirm `Assembly assembly = ws.getAssembly(name); if(!(assembly instanceof TableAssembly)) return null;` — silent null exactly as described.
- `WizVsService` (lines 258–260) already does the identical `DataVSAssembly → getSourceInfo() → getSource()` resolution for aggregate-condition handling, so "mirrors existing resolution" is accurate, not just an assertion.
- The wiz-services caller confirms `assemblyName` sent to `/vs/condition/browse-data` is the chart's own assembly name (not a worksheet table name) — consistent with the described bug.

Root cause and fix approach are correct and well-targeted.

### Issues

**1. Stale javadoc — now actively misleading (both files touched by this bug, only one edited)**
The `@param assemblyName` doc on `WizBrowseDataService.browseData()` (unmodified in this diff) still reads *"the worksheet table/assembly name that owns the column"* — the exact misconception this PR just spent a paragraph explaining is wrong. The sibling `WizBrowseDataController.browseData()` has the identical stale doc. A future reader will hit the same wrong assumption the inline comment just corrected. Worth a one-line doc fix in both files while the context is fresh, e.g. *"the VS chart assembly name; the underlying worksheet table is resolved internally."*

**2. No automated test added**
The PR description frames this as "always, for every column, on every chart" — a severe regression — yet the test plan is manual/live verification only. The package has unit-test precedent for sibling services (`WizVsServiceCheckFailedQueryTest`, `WorksheetTableServiceBatchTest`, etc.), but there's no `WizBrowseDataServiceTest` at all, before or after this PR. A small unit test mocking `RuntimeViewsheet`/`Viewsheet`/`DataVSAssembly` to assert (a) the happy path resolves and calls `setName(wsTableName)` not `assemblyName`, and (b) each new error path throws, would lock in the fix and guard against regressing back to the old silent behavior.

**3. Minor: unchecked `SourceInfo` type**
`sourceInfo.getSource()` is read unconditionally regardless of `SourceInfo.getType()`. For a chart bound to another VS assembly (`XSourceInfo.VS_ASSEMBLY`) rather than a worksheet table, `getSource()` would return a VS-assembly-binding string, not a worksheet table name — `process()` does handle that case via `VSUtil.isVSAssemblyBinding()`, so this likely still resolves correctly by accident. Same gap exists in the mirrored `WizVsService` code, so not a new risk introduced here, and wiz-created charts are always bound directly to worksheet tables in practice — low priority, not blocking.

**4. Minor style nit**
`(ColumnRef) dataRef` is cast twice (once for `setColumn`, once in the new error message). A local `ColumnRef columnRef` variable after the `instanceof` check would avoid the repeat cast.

### Verdict
Correct, well-scoped, single-file fix that closes a serious silent-failure bug and matches an established resolution pattern elsewhere in the codebase. Recommend fixing the stale javadoc (#1) before merge since it directly undermines the PR's own stated purpose; a unit test (#2) is a good-to-have, not blocking.

---

PR #4254, opened 2026-07-14
fix Bug #75651,#756512,#75653,#75654

sarapeng86:


jshobe-inetsoft (approved):
Claude's automated re-review (after commit 2d15507c) confirms the follow-up issue from the first pass was fixed correctly (numeric comparison in validateDateRange) and raises only a pre-existing, non-blocking TIME_PATTERN observation. All checks pass. Approving.

PR #4255, opened 2026-07-14
fix(worksheet): register dateGroupLevel synthetic column for conditions

larryliang-inetsoft:
## Summary
- A `create_worksheet_table` `preAggregateCondition`/`postAggregateCondition`/`rankingCondition` with a `dateGroupLevel` was silently a no-op instead of filtering — the raw timestamp column was compared directly against a date-part literal (which never matches), so every row was silently kept instead of erroring or filtering.
- Wrapping the column in a `DateRangeRef` (mirroring `applyAggregateInfo`'s GROUP BY path) fixed Interval levels (e.g. `"year"`) but Part levels (e.g. `"month of year"`) stayed broken: the synthetic wrapped column was never registered into the table's private `ColumnSelection`, so when it failed to SQL-merge it fell back to StyleBI's in-memory condition evaluation, which can't resolve an unregistered column and defaults to matching every row.
- Threads the private `ColumnSelection` through `buildConditionList`/`buildRankingConditionList`/`appendConditionItem`/`appendRankingConditionItem` so the synthetic date-part column is registered exactly like the GROUP BY path (and marked not-visible so it never leaks into output columns).

## Test plan
- [x] New unit tests in `WorksheetTableServiceConditionTest`: Interval-level registration, Part-level registration, dedup-reuse across condition leaves, and a null-`privateCs` safety test for existing callers. All 9 tests pass.
- [x] Existing 4-arg `buildConditionList` callers (tests) unaffected — added a backward-compatible overload.
- [x] Live-verified via the wiz plugin against the `odoo` datasource: a declarative worksheet filtering Maintenance & Repair category revenue by `date_order` `dateGroupLevel:"month of year" < 5` now correctly returns $2,057,697.11 (previously identical to the $4,715,385.88 unfiltered total — the silent match-all bug). Confirmed no regression on Interval level (`"year"` filter correctly returns $3,545,767.42).

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
@@ -235,7 +235,8 @@ public class WorksheetTableService {
       // 3. Pre-aggregate conditions (WHERE).
       if(request.getPreAggregateCondition() != null && !request.getPreAggregateCondition().isEmpty()) {
          ConditionList preList = buildConditionList(
-            table.getColumnSelection(true), request.getPreAggregateCondition(), worksheet, false);
+            table.getColumnSelection(true), request.getPreAggregateCondition(), worksheet, false,
+            table.getColumnSelection(false));
          table.setPreConditionList(preList);
       }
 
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
## Review

### Overview

Fixes a silent match-all bug: a `dateGroupLevel` on a `preAggregateCondition`/`postAggregateCondition`/`rankingCondition` wrapped the column in a `DateRangeRef` (mirroring the working GROUP BY path in `applyAggregateInfo`), but never registered that synthetic column into the table's private `ColumnSelection`. An unregistered column that fails SQL-merge falls back to StyleBI's in-memory condition evaluation, which can't resolve it by name and defaults to matching every row — so date-part filters (`month of year < 5`, etc.) silently did nothing.

The fix threads the private `ColumnSelection` through `buildConditionList` → `appendConditionItem` and `buildRankingConditionList` → `appendRankingConditionItem`, extracts the wrap+register logic into a shared `applyDateGroupLevel` helper, adds dedup-reuse across condition leaves, and adds a persist step (`addOneTable` step 7) that re-runs `setColumnSelection` when any condition/ranking used a `dateGroupLevel`, mirroring the existing GROUP BY finalize step.

### Correctness — verified

I checked out the PR branch (`149b14fb`) in a worktree and ran the new test class directly:

```
Tests run: 9, Failures: 0, Errors: 0, Skipped: 0
```

I also traced the logic against the codebase's existing conventions:
- `applyDateGroupLevel`'s wrap-and-register logic is a faithful mirror of the pre-existing GROUP BY registration in `applyAggregateInfo` (same `DateRangeRef.getName`/`indexOfAttribute`/`addAttribute(baseIdx, ...)` idiom).
- The dedup check (`privateCs.getAttribute(name)`) is name-based via `ColumnSelection.getAttribute(String)`, which is consistent with how `DateRangeRef`-derived names are looked up elsewhere — this correctly reuses an already-registered synthetic column both across multiple condition leaves *and* across the GROUP BY step if the same field+level was already grouped (nice side effect, not just incidental).
- `dateColumn.setVisible(false)` combined with `setColumnSelection`'s existing `if(!column.isVisible()) continue;` guard confirms the synthetic column can't leak into the table's output columns — this is directly exercised by the new tests.
- No other callers of `buildConditionList`/`buildRankingConditionList` exist outside this file and its test, so the signature changes are safe; the added 4-arg `buildConditionList` overload preserves compatibility.

No correctness issues found.

### Minor gap: ranking-condition path is untested

All four new tests exercise `buildConditionList` (i.e. `preAggregateCondition`/WHERE). None exercise `buildRankingConditionList`/`appendRankingConditionItem` with a `dateGroupLevel`, even though that's explicitly called out in both the PR summary and the code changes (`appendRankingConditionItem` now takes `privateCs` and calls `applyDateGroupLevel` too). Given the bug being fixed was specifically about *silent* wrong behavior, it'd be worth adding one test mirroring `conditionWithPartDateGroupLevelRegistersSyntheticColumn` but through `buildRankingConditionList` (e.g. `TOP_N` on a `"month of year"`-grouped field) to lock in that path the same way.

### Style / documentation

- Javadoc on `applyDateGroupLevel`, `hasDateGroupLevel`, and the step-7 persist block is clear and explains *why*, matching the project's convention of documenting non-obvious invariants rather than restating code.
- The `addOneTable` step numbering/comments are kept consistent (new step 7 inserted, subsequent steps renumbered in comments) — no drift there.

### Risk assessment

Low risk: this only affects behavior when `dateGroupLevel` is present on a condition/ranking item (previously broken/no-op), and the persist step is gated behind `hasDateGroupLevel(...)` so tables without any date-part conditions take the exact same path as before. The live-verified numbers in the PR description (filtered $2,057,697.11 vs. unfiltered $4,715,385.88, with the Interval-level regression check unchanged at $3,545,767.42) are a good sanity check beyond unit tests.

### Recommendation

**Approve**, with an optional (non-blocking) ask to add a ranking-condition unit test for `dateGroupLevel` before merge, given that's the one code path this PR touches without direct test coverage.

PR #4258, opened 2026-07-14
fix(worksheet): fail loud when windowColumns can't SQL-merge

larryliang-inetsoft:
## Summary
- A `create_worksheet_table` `windowColumns` entry (ROW_NUMBER/LAG/SUM OVER/etc) silently produced an EMPTY result instead of computing or erroring when its base table's lineage contained an upstream JS-evaluated expression column several hops back. `WindowExpressionRef` requires its base to be fully SQL-mergeable to push the `OVER(...)` SQL down, but `WindowExpressionRef.isMergeable()` unconditionally reports `true` regardless of the base table, so nothing caught the mismatch.
- Two contributing gaps, both fixed:
  1. `shouldProbe()` only ran the execution probe for `"sql query table"` and expression-column-bearing requests, never for `windowColumns` — extended it to also probe `windowColumns` requests.
  2. The probe alone isn't sufficient: the underlying execution doesn't throw or produce a failed-query fallback lens (all `checkFailedQuery` can detect) — it produces a "successful" but wrongly-empty `TableLens`. Added `requireWindowColumnsMergeable`, an eager check that builds the table's `AssetQuery` and calls `isSourceMergeable0()` (the same authoritative check the runtime query-planning path relies on), throwing a clear, named `IllegalArgumentException` when the base isn't fully SQL-mergeable.

## Test plan
- [x] New unit tests: `WorksheetTableServiceShouldProbeTest` (5 tests covering the `shouldProbe` matrix, incl. the new `windowColumns` cases).
- [x] Full `inetsoft.web.wiz.service.*Test` suite: 176/176 pass, no regressions.
- [x] Live-verified against the wiz plugin on the `odoo` datasource with a real repro: a `ROW_NUMBER() OVER (PARTITION BY category ...)` window sitting over a GROUP BY aggregate whose lineage includes a `JSON.parse(...)` JS expression column (extracting a product name from a jsonb column) now fails loud with `"windowColumns on table \"...\" cannot be computed: its base data is not fully SQL-mergeable..."` instead of silently returning 0 rows.
- [x] Confirmed no false positive: an identical windowColumns-over-2-column-GROUP-BY construction with NO JS expression anywhere in the lineage still succeeds and returns correct data.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
@@ -21,6 +21,7 @@ package inetsoft.web.wiz.service;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import inetsoft.analytic.composition.ViewsheetService;
 import inetsoft.report.TableLens;
+import inetsoft.report.composition.execution.AssetQuery;
 import inetsoft.report.composition.execution.AssetQuerySandbox;
 import inetsoft.sree.security.ResourceAction;
 import inetsoft.sree.security.ResourceType;
@@ -258,6 +259,17 @@ public class WorksheetTableService {
          table.setRankingConditionList(rankList);
       }
 
+      // 7. windowColumns require the base table to be fully SQL-mergeable — the pushed-down
+      // OVER(...) SQL (see WindowExpressionRef) has no non-SQL evaluation fallback. An upstream
+      // JS-evaluated expression column anywhere in the lineage silently breaks mergeability
+      // several hops back, and the window then computes against what the query engine treats as
+      // an empty mergeable base — no exception, just a wrongly-empty result the general probe
+      // below can't catch (it only detects a failed-query fallback lens, not a "successful but
+      // empty" one). Verify eagerly so this fails loud with a named cause instead.
+      if(request.getWindowColumns() != null && !request.getWindowColumns().isEmpty()) {
+         requireWindowColumnsMergeable(worksheet, table, user);
+      }
+
       // 8. Execution probe for render-time-executable tables.
       if(shouldProbe(request)) {
          probeExecutable(worksheet, table, user);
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
Thanks — this pushback was correct and important. I dug into `AssetQuery.getWindowTableLens`/`buildWindowLens`/`WindowTableLens` and confirmed the in-memory fallback genuinely exists; my original comment ("no non-SQL evaluation fallback") was wrong and has been corrected in `f074447`.

**The actual mechanism** (traced against the current source, not just documentation):

1. `buildWindowSpecs` resolves the window's `partitionBy`/`orderBy`/arg `DataRef`s against the base `TableLens`'s real runtime column identity, and DOES throw a named `RuntimeException` when a ref can't be resolved — so it already fails loud, correctly, on paper.
2. But in my repro, the window's `orderBy` is `product_revenue` — an ALIASED `SUM(...)` aggregate. `AssetQuery.getAggCalcHeader` only preserves a caller-specified alias when the aggregate's formula is a `CalcFieldFormula`; for a plain `SumFormula` (my case) it emits `null`, and the non-merged aggregation path's resulting header silently falls back to the pre-aggregate column's raw name instead of the alias. So `buildWindowSpecs` can't find `"product_revenue"` by name and throws.
3. That thrown `RuntimeException` should have surfaced as an error — but `WizVsService.extractChartData`'s catch-all downgrades any non-`IllegalArgumentException` failure to a silently-empty `CreateViewsheetResult` instead of propagating it. That's the actual "zero rows, no exception anywhere" mechanism I originally saw.

So: you're right that "no fallback exists" was the wrong premise, but the specific case this PR guards against — windowColumns ordering/partitioning by an aliased aggregate column, over a non-mergeable base — is **still genuinely broken today**, just via a different, deeper mechanism (an alias-preservation bug in `getAggCalcHeader` + a swallowed exception in `WizVsService`) rather than "no fallback."

**Why I'm not fixing those two bugs in this PR:** both live in shared, heavily-used StyleBI-core paths (`AssetQuery.getAggCalcHeader` is used by the general aggregate-summary pipeline well beyond windowColumns; `WizVsService.extractChartData`'s catch-all likely affects every chart-creation failure mode, not just this one) — fixing them properly needs dedicated testing well beyond this PR's scope, and I don't want to rush a change to code that broad without that. I've corrected the comments (`f074447`) to accurately describe this as a conservative, narrow mitigation rather than claim windowColumns can never work over a non-mergeable base, and explicitly noted the known false-positive: this gate can reject a windowColumns table whose base is non-mergeable but whose partition/order/arg refs are all plain (non-aliased-aggregate) columns — which the in-memory path could likely compute correctly. That's an accepted tradeoff for converting today's silent wrong-empty-result into a clear, actionable error.

I'll file the two deeper bugs (`getAggCalcHeader` alias loss, `WizVsService` swallowed exception) as separate follow-up work rather than attempt them here.

Re: point 2 (test gap for `requireWindowColumnsMergeable` itself) — agreed it has no unit test; it requires a real `AssetQuerySandbox` + datasource to exercise meaningfully (unlike the pure-model `shouldProbe` matrix), which is why I relied on live verification against a real Postgres datasource instead (documented in the PR description) rather than a mocked unit test that wouldn't actually exercise the mergeability planning logic.

larryliang-inetsoft:
## Review

### Overview

Fixes a silent wrong-result bug: a `windowColumns` entry (`ROW_NUMBER`/`LAG`/windowed `SUM`/etc.) over a base table that can't fully SQL-merge produced an empty result instead of erroring, because `WindowExpressionRef.isMergeable()` unconditionally reports `true` regardless of the base. The fix:
1. Extends `shouldProbe` to also probe `windowColumns` requests (previously only `sql query table` and expression-column tables were probed).
2. Adds `requireWindowColumnsMergeable`, an eager check using `AssetQuery.createAssetQuery(...).isSourceMergeable0()` — the same idiom already used in `TablePropertyDialogService.java:60-69` — that throws a named `IllegalArgumentException` when the base isn't fully SQL-mergeable.

Note: this PR already went through a substantive automated review round that correctly caught that the original premise ("no non-SQL evaluation fallback exists") was wrong — `AssetQuery.getWindowTableLens`/`buildWindowLens` genuinely does compute windows in-memory for non-mergeable bases. The author investigated and traced the real, narrower trigger (an aliased-aggregate ref in `partitionBy`/`orderBy`/arg hits an alias-loss bug in `getAggCalcHeader`, whose thrown exception is then swallowed by `WizVsService.extractChartData`'s catch-all) and corrected the javadoc accordingly in `f0744473`. That resolution is sound and I independently confirmed the mechanism traces correctly against the current source.

### Verified

- Checked out the PR branch (`f0744473`) in a worktree and ran the new test class: `WorksheetTableServiceShouldProbeTest` — **5/5 pass**.
- Confirmed `AssetQuery.createAssetQuery(table, RUNTIME_MODE, box, false, -1L, true, false)` + `isSourceMergeable0()` mirrors the exact established idiom in `TablePropertyDialogService.java`.
- Confirmed the `new AssetQuerySandbox(worksheet)` + `box.dispose()` pattern in `requireWindowColumnsMergeable` mirrors the **pre-existing** `probeExecutable` method in this same file (unchanged by this PR) — so it's not a new lifecycle risk; `Worksheet.dispose()` is a no-op unless a `SnapshotEmbeddedTableAssembly` is present, in which case it only swaps data to disk, not destructive.

### Outstanding issue — not yet addressed

**The user-facing exception message is still out of sync with the corrected rationale.** This was flagged inline by the previous review pass but is unresolved in the current HEAD (`f0744473` only touched the javadoc/comments, not the throw site):

```java
throw new IllegalArgumentException(
   "windowColumns on table \"" + table.getName() + "\" cannot be computed: its base " +
   "data is not fully SQL-mergeable (it likely contains an upstream JS-evaluated " +
   "expression column). windowColumns require a pure SQL-mergeable base — remove " +
   "windowColumns from this table, or move the JS-derived value into a \"sql query " +
   "table\" that computes it in SQL instead of JavaScript.");
```

This still unconditionally blames "an upstream JS-evaluated expression column" and suggests moving logic into a `sql query table` — but the corrected javadoc right above it says the actual known trigger is a window ref to an **aliased aggregate** column, which has nothing to do with JS expressions, and the same gate also fires for other `isSourceMergeable0()` failure modes (MV in lineage, dialect/subquery limits, etc.). A user hitting the aliased-aggregate case would get remediation advice that doesn't apply to their situation. This should be updated before merge to match the corrected mechanism — e.g., drop the JS-specific guess and state the general condition, or point users at checking their window's order/partition refs for aliased aggregates.

### Minor observations

- **Test coverage**: only `shouldProbe`'s branching is unit-tested; `requireWindowColumnsMergeable`'s actual gate logic has no unit test (understandable — it needs a real `AssetQuerySandbox`/DB-backed worksheet, and is live-verified per the PR description against `odoo`, consistent with how DB-dependent behavior was handled in the sibling PR #4255).
- **Accepted false-positive tradeoff**: the gate can over-reject a `windowColumns` table whose base is non-mergeable but whose window refs are all plain (non-aliased-aggregate) columns, which the in-memory path could likely compute correctly. This is well-documented and a reasonable "fail loud over silently wrong" tradeoff, but worth being aware it's a real (if narrow) regression for that legitimate sub-case until the deeper `getAggCalcHeader`/`WizVsService` bugs are fixed.

### Recommendation

**Request changes** — solely to fix the stale exception message called out above (previously flagged, still open). Everything else — the mechanism, the test additions, and the conservative-gate rationale — is sound and verified.

PR #4260, opened 2026-07-14
TreeMapElement not available in script.

milotalon:
Root cause

  GraalJavaScriptEngine.installChartClasses() (commit 911f0f894, Bug #75524).

  That whitelist registered only 5 of the 10 concrete element classes — IntervalElement, LineElement, SchemaElement, PointElement, AreaElement — and omitted the newer types: TreemapElement,
  MekkoElement, ParaboxElement, PolygonElement, RelationElement. So elem.setOrientation(TreemapElement.Orientation.TOP_RIGHT) threw ReferenceError: TreemapElement is not defined.

  I confirmed against the legacy engine (../merge/sr13_7/.../JavaScriptEngine.java), whose DefaultPackages auto-imported the whole inetsoft.graph.element package — so all these types were
  reachable by simple name under Rhino.

  Fix

  - GraalJavaScriptEngine.java — registered the 5 missing element classes as JavaClassProxy globals (serves both new TreemapElement() and static/nested access like TreemapElement.Orientation.TOP_RIGHT).
  - VSScriptableController.java — added the same 5 to CHART_CLASSES so they appear in the script editor's autocomplete tree (that list had the identical gap).
  - GraalJavaScriptEngineGlobalsTest.java — added two tests: the exact repro (TreemapElement.Orientation.TOP_RIGHT → "TOP_RIGHT") and resolution checks for all 5 types.

  Verification

  Tests confirmed the fix directly: before, the repro failed with TreemapElement is not defined; after, all 12 tests in the suite pass. (Constructing the elements pulls in GDefaults, whose
  static init needs a graphics environment absent in the headless test JVM, so the tests assert name/enum resolution — which is the actual bug and the exact form used in the failing script.)

### DIFF
--- a/core/src/main/java/inetsoft/util/script/graal/GraalJavaScriptEngine.java
+++ b/core/src/main/java/inetsoft/util/script/graal/GraalJavaScriptEngine.java
@@ -1407,12 +1407,20 @@ public class GraalJavaScriptEngine implements AutoCloseable {
       putClassProxy(bindings, "AxisSpec",   "inetsoft.graph.AxisSpec");
       putClassProxy(bindings, "PlotSpec",   "inetsoft.graph.PlotSpec");
 
-      // elements
+      // elements. Rhino auto-imported the whole inetsoft.graph.element package, so
+      // every concrete element type must be registered here or chart scripts that
+      // reference it by simple name (e.g. new TreemapElement(),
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):

PR #4261, opened 2026-07-14
Fix freehand table grand total showing 0

ngovus-inetsoft:
## Summary
- A calc table's Grand Total row/column cells have no row/col group, so the generated formula is a bare, unqualified crosstab lookup like `none(data['Sum(category_id)'])`. Under GraalJS, `TableArray.hasMember()` only special-cased bare references to a crosstab's own **dimension** names (row/col headers), not its **measure** names — since a crosstab has no literal column headed `"Sum(category_id)"` (that identity varies per data cell, not per column), `Util.findColumn()` couldn't find it, `hasMember()` reported it absent, and GraalJS never called `getMember()` at all. The reference silently resolved to `undefined`/`0` instead of dispatching to the `NamedCellRange`/`CrosstabGroupSelector` logic that already knows how to resolve it correctly.
- Also fixes `CalcTableVSAQuery.hasGrandTotal()` to recurse the full row/column tree (matching the existing `hasSubTotal()`/`collectLevels()` pattern) instead of only checking direct children of the root, so grand-total visibility is detected correctly when the aggregate cell is nested more than one level deep.

## Test plan
- [ ] Build a freehand table with row grouping (e.g. `Year(orderdate)` → `state`) and aggregate columns, plus a Grand Total row and column bound to the same aggregates.
- [ ] Verify the Grand Total row/column now show the correct summed totals instead of 0, matching an equivalent crosstab's totals.
- [ ] Verify existing crosstab and freehand table tests still pass.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/CalcTableVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/CalcTableVSAQuery.java
@@ -1189,35 +1189,15 @@ public class CalcTableVSAQuery extends DataVSAQuery {
     * Check if there exist grand total aggregate.
     */
    private boolean hasGrandTotal(TableLayout layout, XNode root) {
-      boolean hasAgg = false;
-      boolean hasGrp = false;
-
-      if(root != null) {
-         for(int i = 0; i < root.getChildCount(); i++) {
-            XNode child = root.getChild(i);
-            CalcAttr cattr = child == null ? null : (CalcAttr) child.getValue();
-
-            if(cattr == null) {
-               continue;
-            }
-
-            CellBinding bind = layout.getCellBinding(cattr.getRow(),
-                                                     cattr.getCol());
-
-            if(bind != null && bind.getType() == CellBinding.BIND_COLUMN &&
-               !bind.isEmpty())
-            {
-               if(bind.getBType() == CellBinding.SUMMARY) {
-                  hasAgg = true;
-               }
-               else if(bind.getBType() == CellBinding.GROUP) {
-                  hasGrp = true;
-               }
-            }
-         }
+      if(root == null) {
+         return false;
       }
 
-      return hasGrp && hasAgg;
+      Set<Integer> glevels = new HashSet<>();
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (changes requested):
Requesting changes based on the latest Claude automated review, which flags an unaddressed item from the prior review round:

- **`hasGrandTotal` behavior-widening (`CalcTableVSAQuery.java:1191-1201`)** — remains unchanged and untested. As the review notes, `hasGrandTotal` now returns `true` whenever the row/col tree contains any group cell and any aggregate cell anywhere, without requiring the aggregate to sit outside/above the group the way `hasSubTotal` does. Combined with the `hasGrandTotal(...) || hasSubTotal(...)` OR at the call site, this causes the internal temp crosstab to compute row/column grand totals more often than before for ordinary grouped+aggregated calc tables. This appears likely harmless but is untested.

Please either add test coverage confirming the widened behavior is safe, or (if intentional) add a one-line javadoc note on the method clarifying it intentionally does not perform the stricter level-check that `hasSubTotal` does. The test-only additions in this PR otherwise look good.

ngovus-inetsoft:
Re: the `hasGrandTotal` widening flagged above — this is intentional and shouldn't be tightened. The call site's pre-existing comment ("if there exist any aggregate cell in the row/column tree, set grand total visible") already documents the broad intent: any co-occurrence of a group cell and an aggregate cell anywhere in the tree should turn on totals for the internal temp crosstab, since a bare measure-name script reference (e.g. a grand-total cell's `data['Sum(...)']`) can only resolve if the temp crosstab has totals enabled, regardless of nesting depth. The old direct-children-only check was itself the bug (too shallow to find real grand-total cells nested deeper), not a deliberately strict design. `hasSubTotal` already covers the "aggregate not strictly nested below every group" relationship, so the only case this adds is an ordinary aggregate nested below every group — functionally harmless since the rendered calc table still pulls cells via qualified `NamedCellRange` lookups, just some extra total computation on the temp crosstab in that case.

jshobe-inetsoft (approved):

PR #4268, opened 2026-07-15
fix(script): accept full-word date intervals and fail loud on unknown ones

larryliang-inetsoft:
## Summary
`JavaScriptEngine.dateDiff` / `dateAdd` / `datePart` accept only VBScript-style **short** interval tokens (`'d'`, `'m'`, `'yyyy'`, `'q'`, `'ww'`, `'h'`, `'n'`, `'s'`, …). An unrecognized token was handled **silently**:
- `dateDiff` fell through to `return 0`
- `getDateInterval` defaulted to `DAY_OF_YEAR`

So a caller who wrote the natural full word — e.g. `dateDiff('day', d1, d2)` — got a **plausible-but-wrong** result (always `0`) with no error. The trap is sharp because the natural word is exactly the vocabulary the wiz plugin's own `dateGroupLevel` uses (`"day"`, `"month"`, `"year"`), so an LLM-authored expression column reaches for it.

### Fix
New `normalizeDateInterval()`, called at the top of `dateAdd`, `dateDiff`, and `datePartForceWeekOfMonth`:
- trims + lower-cases the token;
- maps a full-word alias to its canonical short token — `day/days→d`, `week/weeks→ww` (a **real calendar week**, deliberately *not* the short `w`, which `dateDiff` counts as DAYS), `month→m`, `quarter→q`, `year→yyyy`, `hour→h`, `minute→n`, `second→s` (plus `min`/`sec`/`hr`/`yr`);
- validates against the **union** of tokens the three functions accept (including the `datePart`-only `mq`/`wmq`/`wq`/`wy`/`dq`/`wm`), throwing `IllegalArgumentException` naming the offending token for anything else.

Forgiving where the intent is unambiguous, loud otherwise — per the project's tool-robustness principle (the deployed product is driven by an LLM that will make exactly this vocabulary mistake).

Found live on an odoo wiz sweep: a `LAG(order date)` window column fed a `dateDiff('day', prev, cur)` gap expression that silently produced **all-zero** gaps.

## Test plan
- [x] `JavaScriptEngineTest` — 48/48 pass (6 new: full-word aliases for dateDiff/dateAdd; `week`→`ww` not `w`; fail-loud on unknown token for all three; datePart specialized tokens still work).
- [x] `GraalJavaScriptEngineTest` + `GraalJavaScriptEngineGlobalsTest` — 14/14 (exercise the functions through real JS).
- [x] `DCMergeDatePartFilterTest` — 4/4 standalone (its failures in a combined run are a pre-existing cross-test static-init issue in `inetsoft.report.internal.Util`, reproduced identically on the unmodified baseline).
- [ ] Live image build + end-to-end verify (`dateDiff('day', …)` returns real gaps; `'fortnight'` fails loud) — in progress.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/core/src/main/java/inetsoft/util/script/JavaScriptEngine.java
+++ b/core/src/main/java/inetsoft/util/script/JavaScriptEngine.java
@@ -310,6 +310,7 @@ public class JavaScriptEngine {
     * Add an interval to a date.
     */
    public static Date dateAdd(String interval, int amount, Object date) {
+      interval = normalizeDateInterval(interval);
       Date dateVal = getDate(date);
 
       if(dateVal != null) {
@@ -343,6 +344,7 @@ public class JavaScriptEngine {
     *    s      Second
     */
    public static double dateDiff(String interval, Object date1, Object date2) {
+      interval = normalizeDateInterval(interval);
       Date dateVal1 = getDate(date1);
       Date dateVal2 = getDate(date2);
 
@@ -410,6 +412,8 @@ public class JavaScriptEngine {
                                                  boolean applyWeekStart,
                                                  int forceDcToDateWeekOfMonth)
    {
+      interval = normalizeDateInterval(interval);
+
       if(forceDcToDateWeekOfMonth > 0 && "wm".equals(interval)) {
          return forceDcToDateWeekOfMonth;
       }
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
**Live-verified on a freshly built image (`local-1017`), end to end through the wiz plugin:**

- `dateDiff('day', prev, cur)` on a `LAG(date_order)` window column now returns real day-gaps (e.g. 72, 314, 73) — identical to the `'d'` short token. Before the fix this silently returned `0` for every row.
- `dateDiff('fortnight', …)` now **fails** (the worksheet-table create returns `success:false`) instead of silently returning `0`.

Minor note: Graal wraps the thrown `IllegalArgumentException` as a generic `"JavaScript error: Failed to invoke script function: dateDiff"`, so the specific `"Unknown date interval 'fortnight'"` detail isn't surfaced to the caller — but the operation correctly fails loud rather than silently producing a wrong result, which is the fix's goal. Surfacing the inner message through the Graal wrapper would be a separate script-engine-error-propagation improvement.

PR #4269, opened 2026-07-15
fixed bug 75656

bonnieshi43:
**Summary**
`DefaultAuthorizationFilter` is supposed to treat an already-anonymous principal the same as “no principal” for the generic auth gate: both must still pass `SecurityEngine.containsAnonymous(orgID)`. The implemented check uses a literal equality against the bare string `"anonymous"`, which never matches a real `SRPrincipal.getName()` value (`"anonymous~;~<orgId>"`). As a result, once any anonymous session is present in the `HttpSession`, every later request in that session skips `containsAnonymous` and is treated as a fully authenticated, non-anonymous user at this filter.

Root cause confirmed: DefaultAuthorizationFilter.doFilter() compared principal.getName() (which is always IdentityID.convertToKey(), e.g. anonymous~;~default) directly against the bare literal XPrincipal.ANONYMOUS ("anonymous"). This equality never holds for a real SRPrincipal, so both the generic auth gate (line 73) and the EM branch (line 81) treated any already-anonymous principal as fully authenticated, skipping containsAnonymous(orgID).

Fix (community/core/src/main/java/inetsoft/web/security/):
- DefaultAuthorizationFilter.java: replaced both broken principal.getName().equals(XPrincipal.ANONYMOUS) checks with isAnonymousPrincipal(principal), reusing the existing correct helper (removed the now-unused XPrincipal import).
- AbstractSecurityFilter.java: widened isAnonymousPrincipal() from private to protected so the subclass can reuse it, per the suggested fix.

Tests updated: the repo already had regression-guard tests written for this exact bug (#75656), marked @Disabled pending the fix — DefaultAuthorizationFilterTest.doFilter_anonymousPrincipal_orgDisallowsAnonymous_redirectsToLogin_guardsSuspect3Fix and SecurityFilterChainOrderingTest.reversedOrder_anonymousFilterBeforeAuthorizationFilter_stillRejectedAfterSuspect3Fix. Removed the @Disabled annotations and updated the stale file-header comments describing the bug as unfixed.

Verification: ran the full inetsoft.web.security test package — 116 tests, 0 failures/errors/skips, including the two newly-enabled regression guards passing unmodified.

### DIFF
--- a/core/src/main/java/inetsoft/web/security/AbstractSecurityFilter.java
+++ b/core/src/main/java/inetsoft/web/security/AbstractSecurityFilter.java
@@ -691,7 +691,7 @@ public abstract class AbstractSecurityFilter
       return true;
    }
 
-   private boolean isAnonymousPrincipal(SRPrincipal principal) {
+   protected boolean isAnonymousPrincipal(SRPrincipal principal) {
       if(principal == null || principal.getName() == null) {
          return false;
       }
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):

PR #4280, opened 2026-07-15
75694, auto-complete missing function definitions

milotalon:
Root Cause:
The AxisSpec methods isLabelBetween/setLabelBetween (added for #74853) and isGridBetween/setGridBetween are correctly annotated with @TernMethod, so their script auto-complete definitions are generated into js-functions.generated.json by TernAnnotationProcessor at build time.

The problem was in that generation, not the annotations. TernAnnotationProcessor rewrites the entire definitions file on every compile, but an annotation processor's RoundEnvironment only exposes the classes recompiled in the current round. On an incremental compile (only changed classes recompiled), the file was overwritten with just that subset, dropping AxisSpec (and every other class not recompiled) from the definitions. This broke auto-complete for those methods until the next full/clean build. A clean build always produced the complete file, which is why the methods appeared after rebuilding from scratch.

Fix:
Changed TernAnnotationProcessor.generateDefinitionFile() to merge with the existing generated file instead of overwriting it. It now reads the previously generated definitions from disk (createResource does not truncate the file until openOutputStream is called, so the prior content is still readable; it is read via plain java.io to avoid javac's restriction on reading and re-creating the same path in one round), seeds the output with them, then overlays the current round's classes on top. Incremental compiles now accumulate definitions instead of clobbering them, while clean builds are unaffected (no prior file -> starts fresh).

File changed:
- build-tools/tern-annotations/src/main/java/com/inetsoft/build/tern/TernAnnotationProcessor.java

### DIFF
--- a/build-tools/tern-annotations/src/main/java/com/inetsoft/build/tern/TernAnnotationProcessor.java
+++ b/build-tools/tern-annotations/src/main/java/com/inetsoft/build/tern/TernAnnotationProcessor.java
@@ -17,6 +17,7 @@
  */
 package com.inetsoft.build.tern;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.auto.common.*;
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):

PR #4283, opened 2026-07-15
75692, preserve hover effect even with no animation in binding

milotalon:
Root Cause:

There are two separate hover-highlight mechanisms in the chart SVGs. Line and area charts dim on hover via pure client-side JavaScript (ChartInlineSvgDirective sets inline opacity directly). Bar, point, candle, box, treemap, mekko, and relation charts instead rely on server-injected :has(.inetsoft-active) CSS dim rules (SVGAnimationDOMInjector.appendHoverCSS()), with JS only toggling the inetsoft-active class.

That hover CSS was injected only inside injectAnimation(), which runs only when the ANIMATION_KEY rendering hint is set. PR #4210 (suppress animation outside runtime) turned off animation in design-time surfaces by passing animate=false to VGraphPair.getPlotGraphic(), which skipped applyAnimationHint() entirely. With no hint, injectAnimation() never ran, so the hover CSS was never injected. The result: in the binding pane (and composer/wizard), line/area still highlighted (their hover is JS-only), but bar-family charts did not (their hover CSS was gone).

Fix:

Decoupled hover-CSS injection from the entrance animation so hover always works while animation stays suppressed in design-time:

1. Added an ANIMATION_FLAG_NOANIM ("noanim") hint flag (SVGSupport).
2. VGraphPair now always applies the animation hint for the plot tile (a new applyHint flag keeps title/axis/facet tiles unaffected). When animate=false, it appends :noanim to the hint instead of omitting the hint.
3. SVGAnimationDOMInjector.injectAnimation() injects the hover CSS first, then returns early when the :noanim flag is present - skipping the entrance animation and leaving data-animated unset (so the client adds .ready immediately and the .ready-gated hover types highlight without delay).

Runtime (animate=true) behavior is unchanged - the hint strings are byte-for-byte identical.


milotalon:
Good call on the test gap. Added two regression tests to `SVGAnimationDOMInjectorTest` covering the `:noanim` path (commit 947f55bbb) — one for a bar (A2) hint and one for a point (A1) hint — each asserting that the hover-dim CSS is injected, no entrance animation `@keyframes` is applied, and `data-animated` stays absent so the client adds `.ready` immediately. Both assertions flip to failures if the `:noanim` early-return is removed, so they genuinely guard the fix.

Left the explanatory comments as-is; the duplication is an intentional guardrail given how easily this was originally introduced.

jshobe-inetsoft (approved):

PR #4288, opened 2026-07-15
Fix invalid HAVING clause for post-aggregation conditions on non-aggregating query levels

ngovus-inetsoft:
## Summary
- A range slider bound to a chart's aggregate measure (via a `VS_ASSEMBLY` source) applies its selection as a post-runtime condition against a mirror/pass-through query level that wraps an already-aggregated subquery.
- `PreAssetQuery.mergeHaving()` always generated SQL `HAVING` for these conditions, even when the current query level performs no aggregation of its own. A `HAVING` with no `GROUP BY` forces the database to treat the whole result as one implicit aggregate group, so strict dialects (e.g. Derby) reject plain (non-aggregate) columns in the SELECT list — producing `Column reference '...' is invalid. When the SELECT list contains at least one aggregate then all entries must be valid aggregate expressions.`
- Fix: when `AggregateInfo` is empty at the current query level, the post condition is now merged into `WHERE` instead of `HAVING`.
- Also hardens the pre/post condition routing decision in `ViewsheetSandbox` to check for a non-empty `AggregateInfo`, not only the crosstab UI flag, so conditions on aggregated columns are routed to post-processing regardless of how the aggregation was defined.

## Test plan
- [x] Reproduced original bug: imported a viewsheet with a range slider bound (`VS_ASSEMBLY`) to a chart's aggregate measure; moving the slider triggered `SQLSyntaxErrorException` on Derby.
- [x] Applied fix, rebuilt, reproduced again: confirmed generated SQL now uses `WHERE` instead of `HAVING`, and the chart renders without error.
- [x] `mvn -pl core -am compile` passes.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/PreAssetQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/PreAssetQuery.java
@@ -707,16 +707,38 @@ public abstract class PreAssetQuery implements Serializable, Cloneable {
       ConditionListHandler handler = new PostConditionListHandler();
       XUtil.convertDateCondition(conds, vars);
       XFilterNode fnode = createXFilterNode(handler, conds, nsql, vars);
-      XFilterNode oroot = nsql.getHaving();
 
-      if(oroot == null) {
-         nsql.setHaving(fnode);
+      // If this query level doesn't aggregate on its own (e.g. a mirror/pass-through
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
Approving based on the completed Claude automated review, which confirms the critical regression from the prior round is resolved and found no blocking issues. All checks pass.

PR #4289, opened 2026-07-15
75683, clarify that orientation only applied when layout is squarified

milotalon:
Root Cause:
The treemap "orientation" feature (#75010) is implemented as a post-layout geometric mirror (TreemapElement.applyOrientation) applied to all layout algorithms. The mirror only yields the intended "largest tile in the chosen corner" result for SQUARIFIED, which sorts descending and anchors the largest tile at the layout origin. SLICE partitions space into strips spanning the full cross-dimension (a vertical mirror is a no-op on full-height strips), and BINARY splits by index — neither anchors the largest tile in a corner, so setting TOP_LEFT (or any orientation) has no meaningful effect for these algorithms.

Fix:
Restricted the orientation transform to the SQUARIFIED algorithm. applyOrientation now skips the flip when the algorithm is SLICE or BINARY, so those treemaps render in their natural, deterministic order. Updated the getOrientation/setOrientation Javadoc to document that orientation applies to the SQUARIFIED algorithm only.

### DIFF
--- a/core/src/main/java/inetsoft/graph/element/TreemapElement.java
+++ b/core/src/main/java/inetsoft/graph/element/TreemapElement.java
@@ -135,7 +135,9 @@ public class TreemapElement extends GraphElement {
 
    /**
     * Get the treemap item anchor orientation. Defaults to {@link Orientation#TOP_LEFT}.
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):
Approving based on the completed Claude automated review, which found the fix correct and minimal with no blocking issues. All checks pass.

PR #4294, opened 2026-07-16
fixed 75689 and add unit case

bonnieshi43:
fixed below issue:
`Condition.isInDateRange` treats the `"tomorrow"` period with the same formula as `"yesterday"`, so a DATE_IN condition with value `tomorrow` matches yesterday’s dates and does not match tomorrow’s dates.


jshobe-inetsoft (approved):
Approving based on the completed Claude automated review (no blocking issues; both prior findings addressed, flaky-test race fixed) and all checks green.

PR #4297, opened 2026-07-16
Fix NPE in VGraphPair.createEVGraph on shared-color legend mismatch (bug-75696)

ngovus-inetsoft:
## Summary
- Importing a chart with a shared color legend (e.g. `sharedcolor.vso`) could throw `NullPointerException: Cannot invoke "inetsoft.graph.guide.legend.Legend.getVisualFrame()" because "evlegend" is null` in `VGraphPair.createEVGraph`.
- `vgraph` and `evgraph` are built from two independently generated `EGraph` instances (`creator`/`creator2`). `EGraph.getVisualFrames()`'s shared-color dedup logic can resolve to a different legend count for each, so `evlegends.getLegend(i)` returned `null` for an index that only existed in `vlegends`.
- Fix bounds the legend size-sync loop by both legend counts and skips any index where either legend is `null`, so mismatched legend counts degrade gracefully instead of crashing chart rendering.

## Test plan
- [ ] Import the reported `sharedcolor.vso` on portal and confirm the chart renders without the NPE
- [ ] Verify legend sizing still matches between vgraph/evgraph for charts where legend counts agree (no regression)

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/VGraphPair.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/VGraphPair.java
@@ -870,15 +870,10 @@ public class VGraphPair {
       if(vlegends != null && evlegends != null) {
          // vgraph and evgraph are built from independently generated EGraphs, so a
          // shared-color legend may resolve to a different legend count in each. Only
-         // sync sizes for legends that exist in both to avoid an NPE (bug#75702).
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):
Approving based on the completed Claude automated review (no blocking issues; both prior review comments addressed in 0701c3d2) and all checks green.

PR #4298, opened 2026-07-16
75648, circular network with ring displayed rendered outside plot area.

milotalon:
Root Cause:
For a circular network chart (RelationElement.Algorithm.CIRCLE), the dashed ring is drawn as a circle centered on the node centroid, passing through the node centers. Between nodes the ring bulges past the node bounding box, so at the left/top of the circle (where typically no node sits) it extends to centerX - radius / centerY - radius.

RelationElement.mxLayout() normalizes the layout by translating node and edge geometry so content starts at non-negative coordinates, but that translation was computed from the node positions only — it did not account for the ring. As a result the ring's left/top arc landed at a negative coordinate, which the plot cannot scroll to reveal, so it was clipped (horizontal scrollbar already fully left, with unused space on the right).

Fix:
After the circular layout computes the ring center/radius, mxLayout() now also shifts the nodes, edges, and the cached layout center by the amount needed to bring the ring's left/top edge back to the origin (only when a border ring is configured and it would otherwise start at a negative coordinate). This is a pure translation, so relative node/edge/ring geometry and the overall scroll range are unchanged; the ring is simply repositioned fully inside the visible plot area.

  - Added @Tag("core") to all three tests so the <groups>core</groups> surefire filter picks them up.
  - Tagging surfaced that two of them (RelationElementLayoutCenterTest, RelationVOInsideTextBoxTest) had never actually run and lacked the Spring test harness — they errored on GDefaults/PropertiesEngine initialization. I added the same harness the passing RelationCoordMinSizeTest uses (@ExtendWith(SpringExtension.class), @ContextConfiguration(BaseTestConfiguration + ConfigurationContextInitializer), @DirtiesContext, @SreeHome). RelationEdgeVOClipTest is pure geometry and needed only the tag.
  - Notably, RelationElementLayoutCenterTest — which directly asserts the layoutCenter invariants I touched in the bug fix — now executes green, giving real coverage of the fix.

### DIFF
--- a/core/src/main/java/inetsoft/graph/element/RelationElement.java
+++ b/core/src/main/java/inetsoft/graph/element/RelationElement.java
@@ -407,8 +407,9 @@ public class RelationElement extends GraphElement {
 
          // The border ring passes through node centers, so it can bulge past the node
          // bounding box between nodes. The node-only translation above leaves the ring's
-         // left/top edge at a negative coordinate the plot can't scroll to reveal (#75648).
-         // Shift nodes, edges and the cached center so the ring also starts at the origin.
+         // min-x/min-y edge (left/bottom in this post-flipY, y-up space) at a negative
+         // coordinate the plot can't scroll to reveal (#75648). Shift nodes, edges and the
+         // cached center so the ring's bounding box also starts at the origin.
          if(shapeBorderShape != null && layoutCenter != null && layoutRadius > 0) {
             double dx = Math.max(0, -(layoutCenter.getX() - layoutRadius));
             double dy = Math.max(0, -(layoutCenter.getY() - layoutRadius));
--- a/core/src/main/java/inetsoft/graph/element/RelationElement.java
+++ b/core/src/main/java/inetsoft/graph/element/RelationElement.java
@@ -404,6 +404,24 @@ public class RelationElement extends GraphElement {
          else {
             layoutRadius = 0;
          }
+
+         // The border ring passes through node centers, so it can bulge past the node
+         // bounding box between nodes. The node-only translation above leaves the ring's
+         // left/top edge at a negative coordinate the plot can't scroll to reveal (#75648).
+         // Shift nodes, edges and the cached center so the ring also starts at the origin.
+         if(shapeBorderShape != null && layoutCenter != null && layoutRadius > 0) {
+            double dx = Math.max(0, -(layoutCenter.getX() - layoutRadius));
+            double dy = Math.max(0, -(layoutCenter.getY() - layoutRadius));
+
+            if(dx > 0 || dy > 0) {
+               nodes.values()
+                  .forEach(v -> v.getMxCell().getGeometry().translate(dx, dy));
+               edges
+                  .forEach(v -> v.getEdge().getGeometry().translate(dx, dy));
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (changes requested):
Holding off on merge based on the automated Claude review, which flagged a test-coverage gap that isn't resolved:

**The new tests don't actually exercise the fix.** The PR description states the tests in `RelationElementLayoutCenterTest.java` give "real coverage of the fix," but per the review the only changes to that file are Spring test-harness annotations — no test constructs a scenario where, pre-fix, the ring's bounding box would land at a negative coordinate (`layoutCenter.getX() - layoutRadius < 0`) and then asserts it starts at `>= 0` after `mxLayout()`:
- `mxLayoutPopulatesLayoutCenterAsNodeCentroid_circular` never calls `addShapeBorder(...)`, so `shapeBorderShape == null` and the new shift block (RelationElement.java:408-424) is skipped entirely — the test passes identically with or without the fix.
- `addShapeBorderRegistersFormInEGraphAfterLayout` adds a border but only asserts `getFormCount() == 1`; it never checks `layoutCenter`/`layoutRadius`, so it wouldn't catch a regression in the shift math.

Please add a test with a shape border configured and enough node asymmetry to trigger a negative pre-shift coordinate, asserting (1) `layoutCenter.getX() - layoutRadius >= -1e-6` (and the Y equivalent) after layout, and (2) the shift is a pure translation (e.g. centroid-to-node distances unchanged before/after).

The fix itself looks sound and the two minor items (stream-usage consistency, the "left/top" comment wording given `flipY()`) are non-blocking. Just want real coverage of the shift block before merging.

milotalon:
Thanks @jshobe-inetsoft — addressed in ffc36c5f9.

**1. Test coverage gap (the main blocker) — fixed.** Added `borderRingBoundingBoxStartsAtNonNegativeCoordinates_circular` to `RelationElementLayoutCenterTest`. It builds a 3-node circular layout (120° spacing, so the leftmost node sits at 150° — no node at the 180° west cardinal) with `addShapeBorder(...)` configured, and uses small nodes + a large `layoutSize` so the ring radius dominates the node size and the ring genuinely bulges past the leftmost node. It asserts:
- `centerX - radius >= -1e-6` and `centerY - radius >= -1e-6` after `mxLayout()` (the ring bbox starts at non-negative coords), and
- every node center remains equidistant (`== radius`) from the shifted `layoutCenter` — i.e. the shift is a pure translation.

I verified it's a real guard: with the shift block temporarily disabled the test fails (`centerX - radius == -46.28`); with the fix it passes. Full test class is green (7 tests).

**2. `flipY()` / "left/top" comment wording — fixed.** Good catch. Reworded to "min-x/min-y edge (left/bottom in this post-flipY, y-up space)" so it no longer implies a y-down convention.

**3. Stream-usage consistency — left as-is (non-blocking).** The new block uses `nodes.values().forEach(...)` / `edges.forEach(...)` without `.stream()`; as the review notes, the `.stream()` in the neighbouring pre-existing code is redundant, so I kept the cleaner form rather than adding a redundant call for visual symmetry.

jshobe-inetsoft (approved):
Claude's re-review confirms the test coverage gap and comment wording issues from the prior round were resolved, with no new correctness/security/regression concerns. Approving.

PR #4299, opened 2026-07-16
Bug #75690, allow script JS functions to adapt to Java functional interfaces

jshobe-inetsoft:
## Summary

Chart shape scripts that pass a JS arrow function to a Java method expecting a functional interface (e.g. `frame.getAllShapes().filter(a => a instanceof GShape.ImageShape).forEach(...)`) failed at runtime with:

```
Cannot convert 'a => a instanceof GShape.ImageShape'(language: JavaScript, type: Function)
to Java type 'java.util.function.Predicate': Unsupported target type. (line 7)
```

`CategoricalShapeFrame.getAllShapes()` returns a `java.util.stream.Stream`, so the script calls `Stream.filter(Predicate)` and `Stream.forEach(Consumer)` with JS lambdas.

## Root Cause

The GraalJS `HostAccess` policy built in `ScriptHostAccess.hostAccess()` never enabled guest-function → Java functional-interface adaptation. Rhino did this automatically (auto-adapting a JS function to a single-method interface), so under the GraalJS migration this capability was lost and any script passing a JS function where a functional interface is expected fails.

## Fix

Add `.allowImplementationsAnnotatedBy(FunctionalInterface.class)` to the `HostAccess` builder. This lets a guest function implement any `@FunctionalInterface`-annotated type (`Predicate`, `Consumer`, `Function`, `Comparator`, etc.), restoring Rhino parity. It's the same allowance GraalVM's built-in `HostAccess.ALL`/`EXPLICIT` presets use.

The change is additive and does not weaken the sandbox: the `Java.type(...)` class-lookup allow-list (`classFilter`) is untouched, thread creation stays disabled, and dangerous classes remain denied. The narrower annotated form is used rather than `allowAllImplementations(true)`.

## Test Plan

- Added `ScriptHostAccessTest.jsFunctionAdaptsToFunctionalInterface`, exercising both `Stream.filter(Predicate)` and `Stream.forEach(Consumer)` with JS lambdas. Verified the test fails with the exact reported error when the fix is reverted, and passes with it.
- `./mvnw -pl core test -Dtest=ScriptHostAccessTest` → 11/11 pass.
- Manually verified in the portal: importing/opening `ShapseApply.vso` no longer raises the conversion error.

Fixes Redmine #75690.

### DIFF
--- a/core/src/main/java/inetsoft/util/script/graal/ScriptHostAccess.java
+++ b/core/src/main/java/inetsoft/util/script/graal/ScriptHostAccess.java
@@ -169,6 +169,16 @@ public final class ScriptHostAccess {
                   .allowMapAccess(true)
                   .allowIterableAccess(true)
                   .allowIteratorAccess(true)
+                  // Rhino parity: Rhino auto-adapted a JS function to a single-method
+                  // Java interface, so scripts could pass a lambda where a functional
+                  // interface is expected (e.g. Stream.filter(Predicate),
+                  // Stream.forEach(Consumer) via graph shape scripting). GraalJS refuses
+                  // this by default ("Unsupported target type"); allow a guest function
+                  // to implement any @FunctionalInterface-annotated type. This is the
+                  // same allowance the built-in HostAccess.ALL/EXPLICIT presets use, and
+                  // it does not widen class reachability (still gated by classFilter) or
+                  // permit implementing arbitrary/abstract types. (#75690)
+                  .allowImplementationsAnnotatedBy(FunctionalInterface.class)
                   // FIX 2: Deny reflective escape paths even when allowPublicAccess(true) is set.
                   // denyAccess takes precedence over allowPublicAccess for the listed classes.
                   // denyAccess(Object.class, false) blocks only methods declared on Object itself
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft:
Thanks for the thorough review.

**Re: scope of the widening (obs. 1)** — the broad `@FunctionalInterface` allowance is intentional. Rhino adapted a JS function to *any* single-method interface, so restricting it to a `targetTypeMapping` of `Predicate`/`Consumer`/`Function` would re-break other legitimate lambda-to-interface cases (`Comparator`, `Runnable`, …) that this bug is meant to restore. On the audit: the incremental risk is bounded by GraalJS itself — a guest function can only be invoked on the thread that created its `Context`, so a method that dispatches a `Runnable` off-thread (e.g. `SequentialDispatcher`, which extends `EventLoop`) throws `IllegalStateException` rather than escaping, and in-thread invocation is equivalent to the script calling the code directly. The change also grants no new class reachability — `classFilter()`/`Java.type(...)` gating is untouched.

**Re: test coverage (obs. 2)** — extended `ScriptHostAccessTest` with a `Comparator` adaptation case (a single-method interface returning `int` rather than boolean/void) and a negative case confirming a JS object cannot satisfy a plain multi-method interface, so the tests document that the allowance is scoped to `@FunctionalInterface` types. All 13 tests pass locally. (commit c997c49e8)

ngovus-inetsoft (approved):

PR #4303, opened 2026-07-16
75645, dash line not easy to see on circular network graph

milotalon:
Root Cause:
The dashed guide ring drawn around the circular network/relation chart was registered with GDefaults.DEFAULT_LINE_COLOR (0xeeeeee, ~238,238,238), an almost-white gray. Against the plot background this made the ring nearly invisible in both the light and dark themes.

Fix:
Changed the ring's shape-border color to Color.LIGHT_GRAY (0xC0C0C0) at the single call site in GraphGenerator (CHART_CIRCULAR branch). The shared GDefaults.DEFAULT_LINE_COLOR constant was left unchanged so axis lines and other borders that rely on it are unaffected; only the circular chart's guide ring is impacted.


jshobe-inetsoft (approved):
Claude's review found a correct, minimal, low-risk fix with no blocking issues. Approving.

PR #4307, opened 2026-07-17
feat(wiz): support copy-before-modify for chart color/format requests

Zxm-happy:
## Summary
- Add a `copy` flag to `ChartColorsRequest` / `ChartFormatRequest` so callers can duplicate the active chart's assembly before applying a color/format/marker change instead of mutating it in place.
- Add `WizVsService.duplicatePrimaryAssembly(vs, source)`, reusing the existing `uniqueAssemblyName` + `rebindAssembly` path so the clone keeps prior formatting/color customizations.
- `WizAutoBindingService.setChartColors` / `setChartFormat` (markers share the `/viewsheet/format` endpoint) duplicate the assembly first when `copy=true`.

Companion PR in `stylebi-wiz` wires this `copy` param through the MCP tool layer.

Based on `bug-75446-1` since that's where this branch was cut from (per repo convention for wiz-side changes here).

## Test plan
- [x] Full Java test suite — 193 tests passing (2 pre-existing unrelated failures in `MetadataApiServiceGetJDBCDatasourceTest`, confirmed baseline via git stash)
- [ ] End-to-end MCP tool call with `copy: true` (blocked this session by a stale MCP transport session; to be verified in a follow-up session)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/model/ChartColorsRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/model/ChartColorsRequest.java
@@ -52,6 +52,11 @@ public class ChartColorsRequest {
    public Map<String, String> getCategoryColors() { return categoryColors; }
    public void setCategoryColors(Map<String, String> categoryColors) { this.categoryColors = categoryColors; }
 
+   /** When true, duplicate the assembly first (keeping the original untouched) and apply the color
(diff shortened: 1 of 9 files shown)


Zxm-happy:
Closing to resubmit with a different base branch.

PR #4308, opened 2026-07-17
feat(wiz): support copy-before-modify for chart color/format requests

Zxm-happy:
## Summary
- Add a `copy` flag to `ChartColorsRequest` / `ChartFormatRequest` so callers can duplicate the active chart's assembly before applying a color/format/marker change instead of mutating it in place.
- Add `WizVsService.duplicatePrimaryAssembly(vs, source)`, reusing the existing `uniqueAssemblyName` + `rebindAssembly` path so the clone keeps prior formatting/color customizations.
- `WizAutoBindingService.setChartColors` / `setChartFormat` (markers share the `/viewsheet/format` endpoint) duplicate the assembly first when `copy=true`.

Companion PR in `stylebi-wiz` wires this `copy` param through the MCP tool layer.

(Resubmit of #4307 against `stylebi-wiz-main-rebase` instead of `bug-75446-1`.)

## Test plan
- [x] Full Java test suite — 193 tests passing (2 pre-existing unrelated failures in `MetadataApiServiceGetJDBCDatasourceTest`, confirmed baseline via git stash)
- [ ] End-to-end MCP tool call with `copy: true` (to be verified in a follow-up session)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/model/ChartColorsRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/model/ChartColorsRequest.java
@@ -52,6 +52,11 @@ public class ChartColorsRequest {
    public Map<String, String> getCategoryColors() { return categoryColors; }
    public void setCategoryColors(Map<String, String> categoryColors) { this.categoryColors = categoryColors; }
 
+   /** When true, duplicate the assembly first (keeping the original untouched) and apply the color
+    *  change to the new copy instead of in place. Defaults to false. */
+   public boolean isCopy() { return copy; }
+   public void setCopy(boolean copy) { this.copy = copy; }
+
    private String wizRuntimeId;
    private String assemblyName;
    private String viewsheetIdentifier;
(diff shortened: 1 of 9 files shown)


Zxm-happy:
Thanks for the review — addressed the minor observations in f9898a8:

1. **Duplicated copy-if-requested block** — extracted into a shared private helper `copyIfRequested(boolean copy, RuntimeViewsheet rvs, ChartVSAssembly chart)`; both `setChartFormat` and `setChartColors` now call it instead of repeating the inline `if(request.isCopy()) { ... }` logic.
2. **Silent no-op on unsupported type** — left as-is. As noted, this path is unreachable today since `ChartVSAssembly` is always present in `ASSEMBLY_FACTORIES`; adding error surfacing for an unreachable case would be speculative, so no change here.
3. **`duplicatePrimaryAssembly` precondition** — added a javadoc paragraph on `WizVsService.duplicatePrimaryAssembly` documenting that `source` must already be the primary assembly on `vs`, and what happens if that's violated (demotes an unrelated assembly instead), to guard future reuse of this method outside the current two call sites.

PR #4309, opened 2026-07-17
add untag ci case plan and some case

bonnieshi43:


jshobe-inetsoft (approved):
Claude's automated review found the previously-flagged NPE risk fixed with test coverage, and no outstanding issues. Approving.

PR #4314, opened 2026-07-17
bug-75664-1: fix null Change from previous when part-date-group field is ranked

ngovus-inetsoft:
## Summary
- Follow-up to bug-75664: the earlier `DataSetRouter` fix only applied its natural numeric-order fallback for part-date-group dimensions (HourOfDay, DayOfWeek, MonthOfYear, Quarter) when the field had no explicit sort comparator.
- A Top-N/Bottom-N ranking with "Sort By Value" sets an explicit value-based comparator on the dimension, bypassing that fallback entirely — so "Change from previous" navigated by the ranking's value order instead of natural calendar order, producing wrong or null results (e.g. "Change from previous HourOfDay" blank at hour 5 when a Top-N ranking is applied to HourOfDay).
- Give the natural-order comparator priority over any value-based sort/ranking comparator for these fields, since "previous hour/day/month/quarter" only has meaning in calendar order.

## Test plan
- [x] Added `ValueOfColumnTest.testPreviousOnPartDateGroupIgnoresRankingSortComparator`, which mocks a ranking-style value comparator on an HourOfDay field and asserts "previous of hour 5" resolves to hour 2 (natural order) instead of whatever the ranking comparator would place before it.
- [x] Verified the new test fails without the fix (returns the ranking-order neighbor instead of the calendar-order neighbor) and passes with it.
- [x] `ValueOfColumnTest` (20 tests) and `ChangeColumnTest` (9 tests) pass.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/data/DataSetRouter.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/data/DataSetRouter.java
@@ -58,10 +58,15 @@ public class DataSetRouter extends AbstractRouter {
          }
       }
 
-      comp = data.getComparator(field);
+      // Part-date-group fields (HourOfDay, DayOfWeek, MonthOfYear, Quarter) must navigate
+      // previous/next in natural calendar order regardless of any value-based sort/ranking
+      // comparator (e.g. a Top-N "Sort By Value" ranking) applied for display purposes —
+      // "previous hour" has no meaning in rank-value order. Only fall back to the dataset's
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
Approving based on the resolved Claude Code review (no blocking issues found — comparator-priority fix in `DataSetRouter` is correct and well-tested) and all CI checks passing (build, claude-review, Semgrep, Semgrep OSS, license/cla).

PR #4320, opened 2026-07-18
fix: non-Cartesian charts wrongly reported as no-data in the viewer

larryliang-inetsoft:
## Summary

- `chart-area.component`'s `noData` getter and `vs-chart.component`'s `emptyChart` getter both used `axes.length == 0` as a proxy for "chart has no data." Treemap, sunburst, circle-packing, icicle, polar (pie/radar), relation, and geo charts never populate `axes` by design (see `GraphTypes.isRect()` on the Java side), so a fully-rendered chart of one of these types was indistinguishable from a genuinely empty one — the viewer/composer showed the `_#(data.query.columnResultEmpty)` ("No data is available!") placeholder over real data.
- Discovered live: a treemap built via the wiz plugin with real, reconciled revenue data showed this placeholder in the companion viewer, while a server-side re-execution of the same saved chart confirmed 29 correct rows.
- Adds `GraphTypes.isRect()` to the TS side, mirroring the existing Java helper, and only applies the axes-based heuristic to Cartesian chart types. For everything else, both getters now defer to the server's own `model.noData` flag, which already correctly reflects whether the graph actually built with data (`VSChartAreasService` only sets it true when `pair.getRealSizeVGraph() == null`), independent of chart type.

## Test plan

- [x] Added 2 regression tests to `chart-area.component.display.tl.spec.ts`: a non-rect chart type (treemap/radar) with no axes but `noData: false` is correctly reported as having data; a non-rect chart with a genuine server `noData: true` is still correctly reported as empty.
- [x] Full `chart-area.component.display.tl.spec.ts` suite passes (77/77).
- [x] Verified live: rebuilt the Angular bundle + Docker image, redeployed, and confirmed a previously-broken treemap now renders correctly in the live viewer.

### DIFF
--- a/web/projects/portal/src/app/common/graph-types.ts
+++ b/web/projects/portal/src/app/common/graph-types.ts
@@ -552,6 +552,17 @@ export class GraphTypes {
       return GraphTypes.isScatteredContour(type) || GraphTypes.isMapContour(type);
    }
 
+   /**
+    * Check if the graph is drawn on a rectangular (X/Y axis) coordinate. Polar (pie/radar),
+    * treemap-family, relation, and geo charts use non-Cartesian layouts and never populate
+    * axes, so axes.length is not a valid "has data" signal for them.
+    */
+   static isRect(type: number): boolean {
+      const notRect = GraphTypes.isPolar(type) || GraphTypes.isTreemap(type) ||
+         GraphTypes.isRelation(type) || GraphTypes.isGeo(type);
+      return !notRect;
+   }
+
    static isCompatible(type1: number, type2: number): boolean {
       if(type1 == type2) {
          return true;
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (dismissed):
The automated Claude Code review is complete with no blocking issues (the `GraphTypes.isRect()` fix correctly mirrors the Java source of truth, and the observations raised are non-blocking suggestions), but this PR cannot be merged yet because the **Validation Build** check is currently failing.

The `inetsoft-core` module's test run reports 11 errors (out of 3751 tests), all `IllegalState: Failed to load ApplicationContext` / "ApplicationContext failure threshold exceeded" errors in unrelated Spring-context-based tests (e.g. `GenericLdapAuthenticationProviderUnitTest`, `PermissionMatrixResourcesS4Test`, `TaskBalancerTest`, `CompletionConditionTest`, etc.). Since this PR only touches frontend TypeScript files (`graph-types.ts`, `chart-area.component.ts`, `vs-chart.component.ts`, and their tests), these failures look like flaky/unrelated CI infrastructure issues rather than something introduced by this change — but per policy I can't approve/merge while a check is red.

Could you take a look at the failing job (https://github.com/inetsoft-technology/stylebi/actions/runs/29630421952/job/88042791002) and either confirm it's a known flaky failure and re-run the build, or address it if it's a real regression? Happy to re-review once the build is green.

PR #4327, opened 2026-07-20
feat(wiz): scope /ws/structure to a generation with upstream closure

NeverStopFMH:
Add an optional `generation` query param to /ws/structure. When supplied, return that generation's tables (matched by their `_<generation>` name suffix) plus their transitive upstream dependency closure: a scoped table may reuse a prior generation's base table, which is pulled in and ordered ahead of its dependent so the returned structure is a self-contained subgraph (baseTables/joins resolve within it). primaryTable is reconciled to this generation's own tables. When omitted, every table is returned.

- DatasourceMetaApiController: pass through the generation query param
- MetadataApiService: seed filter -> upstream closure -> enumeration-order output; collectUpstreamRefs + expandToUpstreamClosure helpers; primaryTable fallback scoped to the seed
- MetadataApiServiceStructureTest: closure, multi-level chain, ordering, dangling-reference skip, and seed-scoped primaryTable fallback cases

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/DatasourceMetaApiController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/DatasourceMetaApiController.java
@@ -97,12 +97,21 @@ public class DatasourceMetaApiController {
       return metadataService.getWorksheetMetaData(WizUtil.decodeId(worksheetId), (XPrincipal) principal);
    }
 
+   /**
+    * Returns a worksheet's per-table structure.
+    *
+    * @param generation optional worksheet-table generation. When supplied, only the tables belonging
+    *    to that generation (by their {@code _<generation>} name suffix) are returned; omit it to get
+    *    every table (the MCP path, whose in-place edits need to see all tables, omits it).
+    */
    @GetMapping("/ws/structure/{id}")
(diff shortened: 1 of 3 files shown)


NeverStopFMH:
Thanks for the review. Addressed both:

**1. Empty `tables` with a non-null `primaryTable` (contract).** Fixed in `7502656`. Added `resolveScopedPrimaryTable`: a generation-scoped read that matches no tables now returns `primaryTable = null`, so the response stays a self-contained subgraph (primaryTable always resolves within `tables`). Unfiltered reads keep the prior behavior (an empty worksheet keeps its original primary). Covered by 4 new unit tests (`scopedPrimaryTable*`), total 43 passing.

**2. Duplicate baseTables/joins extraction.** Considered and intentionally left as-is. The upstream map must be built for *all* assemblies before filtering (the closure can't be computed otherwise), and `extractStructureBaseTables`/`extractStructureJoins` are lightweight in-memory walks. The suggested reuse would either fully `buildStructureTable` the filtered-out tables (strictly more work) or add a per-table cache layer — neither pays off at typical table/join counts. Flagged for revisit if very large worksheets prove otherwise.

Nelson-cao (approved):

PR #4330, opened 2026-07-20
Bug #75706: fix org admin denied access to Organization Administrator role

ngovus-inetsoft:
## Summary
- The built-in "Organization Administrator" role is a single global role (`orgID = null`) shared across all organizations, not one role per org.
- `DefaultCheckPermissionStrategy` had three org-scope checks for `SECURITY_ROLE` that compared the role's `orgID` against the caller's org and rejected any mismatch — which incorrectly excluded this org-less-by-design role.
- Result: an Organization Administrator viewing EM → Security → Users → "Organization Administrator" mode got a misleading `SecurityException` surfaced in the UI as "You don't have permission to access System Administrator level identities."
- Fix: each of the three checks now also allows the role through when `isOrgAdministratorRole(role)` is true, so the built-in Organization Administrator role is treated as in-scope for every org admin, while the true system-wide `Administrator` role remains blocked via the existing `isSiteAdmin`/`isSystemAdministratorRole` checks.

## Test plan
- [x] `core` module compiles cleanly with the change
- [ ] Manually verify: create org `organization0` + user with Org Admin role, log in, EM → Security → Users → switch to "Organization Administrator" mode — should no longer show the permission error
- [ ] Verify a true cross-org role (belonging to a different org) is still denied
- [ ] Verify the system-wide `Administrator` role remains inaccessible to org admins

### DIFF
--- a/core/src/main/java/inetsoft/sree/security/DefaultCheckPermissionStrategy.java
+++ b/core/src/main/java/inetsoft/sree/security/DefaultCheckPermissionStrategy.java
@@ -239,7 +239,7 @@ public class DefaultCheckPermissionStrategy implements CheckPermissionStrategy {
             return true;
          }
 
-         if(checkOrgAdminPermission(type, resource, organization, xPrincipal)) {
+         if(checkOrgAdminPermission(type, resource, organization, xPrincipal, action)) {
             return true;
          }
       }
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):
Claude's automated review flagged critical/high issues on the first commit (potential reopening of Bug #75574 and a cross-tenant same-named-role leak); the follow-up commit fd93b854 resolved both, confirmed by Claude's re-review after the latest commit with no remaining blocking issues. Approving and squash-merging.

PR #4340, opened 2026-07-21
Bug #75715: fix AI assistant panel rendering below chart mini-toolbar

ngovus-inetsoft:
## Summary
- The viewsheet mini-toolbar (chart/object floating hover toolbar) uses `z-index: 992000`, higher than the AI assistant panel's `z-index: 1045`, causing the toolbar to render on top of / bleed through the assistant panel while binding/editing a chart.
- Raise `.ai-assistant-panel`'s z-index above the mini-toolbar's so the assistant panel stays on top.

## Test plan
- [ ] Open a viewsheet/composer chart binding pane with the AI Assistant panel open
- [ ] Confirm the chart's floating mini-toolbar no longer renders above the assistant panel

### DIFF
--- a/web/projects/portal/src/app/vsobjects/objects/mini-toolbar/mini-toolbar.component.scss
+++ b/web/projects/portal/src/app/vsobjects/objects/mini-toolbar/mini-toolbar.component.scss
@@ -17,7 +17,7 @@
  */
 .mini-toolbar {
   position: absolute;
-  z-index: 992000; // > object-editor.popped.active
+  z-index: 9920; // > object-editor.popped.active (9910)
   pointer-events: none;
   visibility: hidden;
   transition: visibility 0.5s linear;
--- a/web/projects/portal/src/scss/internal/_layout.scss
+++ b/web/projects/portal/src/scss/internal/_layout.scss
@@ -428,7 +428,7 @@ li.active, a.active {
 
 .binding-editor, .embed-chart-container {
   .mini-toolbar {
-    z-index: 9999 !important;
+    z-index: 9918 !important; // still > object-editor.popped.active (9910), but below the base (9920)
   }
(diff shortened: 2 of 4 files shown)


ngovus-inetsoft:
Reworked the fix based on review feedback. The prior approach (raising `.ai-assistant-panel` to `z-index: 995000`) had two problems: it still lost to the max-mode mini-toolbar (`999999`), and jumping the panel that high introduced new regressions — the remove-bookmarks confirm dialog (`999901`/`999902`), `.picker.picker-dialog` (an exact tie at `1000000`), and several toast/tooltip overlays at `10000+` would now render behind the panel.

Root cause: `mini-toolbar`'s z-index was inflated to `992000`/`999999` when its own code comment says it only needs to beat `object-editor.popped.active` (`z-index: 9910`). Traced the history — that value goes back to this repo's very first commit (the open-source squash from 2024-07-12), so there's no available rationale for the six-digit magnitude beyond the stated intent in the comment.

New approach tames the actual offender instead of escalating further:
- `mini-toolbar` base: `992000` → `9920`
- `mini-toolbar` max-mode: `999999` → `9930`
- `mini-toolbar` binding-editor/embed-chart override: `9999` → `9918`
- `.ai-assistant-panel`: `1045` → `9940`

All three mini-toolbar variants stay above `object-editor.popped.active` (`9910`), preserving original intent, and the panel clears all of them while staying comfortably below every other overlay in the app (toasts, tooltips, modals, the picker dialog) — no other files need to change.

jshobe-inetsoft (approved):
Approving based on Claude's automated review — the second review confirms the reworked z-index fix (taming .mini-toolbar instead of escalating the panel) resolves the earlier concerns with no remaining blocking issues.

PR #4341, opened 2026-07-21
Bug #75722: fix MVScriptable NPE from MV storage read racing with concurrent rebuild

ngovus-inetsoft:
## Summary
- `LocalMVIncremental.update0()` built a fresh `MVScriptable` per MV column *before* acquiring the `mv.fs.update` cluster lock, and each `MVScriptable` did its own independent `MVStorage` lookup instead of reusing the `mv` field already validated non-null by the `MVIncremental` constructor.
- If that independent read landed in the window where a concurrent full MV rebuild renames its temp file onto the stable file name (also gated by the same lock), the lookup failed silently (swallowed exception), leaving `mv` null — the very next line dereferenced it, throwing an uncaught `NullPointerException` that aborted the whole incremental update task.
- Fix: reuse the already-validated `mv` field via a new `MVScriptable(MVDef, MVColumn, MV)` constructor instead of re-fetching from storage, and guard `MVScriptable.max()`/`min()` against a null `mv` as a defense-in-depth fallback (returns `null` instead of throwing).

## Test plan
- [x] Updated `MVScriptableTest` characterization tests to assert the new `null`-return behavior instead of pinning the NPE.
- [x] `MVScriptableTest` — 7/7 tests pass.
- [x] Full `core` module test suite — 4075 tests, 0 failures, 0 errors, 63 skipped.
- [x] Confirmed no other `MVIncremental` subclasses or enterprise-side usages need the same fix (only `LocalMVIncremental` exists, and `MVConditionListHandler.executeScript()`'s independent `MVScriptable` lookup runs inside the same lock, so it isn't exposed to the race).

### DIFF
--- a/core/src/main/java/inetsoft/mv/data/MVConditionListHandler.java
+++ b/core/src/main/java/inetsoft/mv/data/MVConditionListHandler.java
@@ -47,6 +47,7 @@ public class MVConditionListHandler extends ConditionListHandler {
       super();
 
       this.mvdef = def;
+      this.mv = mv;
       this.vars = vars;
       Worksheet ws = table.getWorksheet();
       this.box = new AssetQuerySandbox(ws == null ? new Worksheet() : ws);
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):
Approving based on Claude's automated review — the follow-up review confirms all three points from the initial review were addressed, no remaining correctness/security/backward-compatibility concerns.

PR #4344, opened 2026-07-21
Bug #75717, fix NG0955 duplicate track keys in color picker pane

jshobe-inetsoft:
## Summary

Opening the chart highlight dialog (or any color picker) and clicking a color threw an Angular `NG0955` error ("The provided track expression resulted in duplicated keys"), reported at `cp-color-pane.component.html:41`.

## Root Cause

Two `@for` loops in `cp-color-pane.component.html` tracked by the color *value* (`track swatch`). The `recentColors` list is always padded with duplicate `null`/empty-string entries (`RecentColorService` initializes 8 `null`s; the getter falls back to 8 `""`s), so tracking by value produced duplicate keys and triggered NG0955 as soon as the pane rendered. The palette-row loop used the same anti-pattern (latent — would fail for any palette with repeated values).

## Fix

Changed both loops to `track $index`. The palettes are fixed-length positional rows replaced in place (no keyed reordering), so index tracking is the correct model. No change to selection/click behavior — the loop bodies still use `swatch` for bindings and `selectColor(swatch)`.

## Test Plan

- In composer, open a chart's highlight dialog and click a color swatch — no NG0955 error; color selection works.
- Verify recent-colors row and all palette rows render and are selectable.
- `npm run build` succeeds (AOT template compilation).
- `npm run test` passes (946 portal + 356 em tests, incl. `cp-color-pane.spec.ts`).

Fixes Redmine #75717.

### DIFF
--- a/web/projects/portal/src/app/widget/color-picker/cp-color-pane.component.html
+++ b/web/projects/portal/src/app/widget/color-picker/cp-color-pane.component.html
@@ -38,7 +38,7 @@
         <div class="form-row-float-label color-picker-palette-row color-picker-recent">
           <table>
             <tr class="bb-gray">
-              @for (swatch of recentColors; track swatch) {
+              @for (swatch of recentColors; track $index) {
                 <td>
                   <button type="button" class="color-picker-swatch bd-gray"
                     [class.transparent-swatch]="swatch === ''"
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft:
Thanks for the review — confirming the `track $index` fix is correct and complete, which matches your verification.

On the non-blocking observation: you're right that `cp-color-pane.spec.ts` only constructs `ColorPane` directly and never mounts the template, so the NG0955 regression isn't covered by an automated test. I'm keeping this PR to the minimal template fix per our bug-fix guidelines; adding a `TestBed`-based rendering test for the color pane is a worthwhile follow-up that I'll track separately rather than expanding the scope here.

ngovus-inetsoft (approved):

PR #4345, opened 2026-07-21
Bug #75718, guard against null identity in IdentityService.getIdentityInfo

jshobe-inetsoft:
## Summary

Fixes an intermittent NPE that surfaced as a misleading `ERROR` log when deleting a user in Enterprise Manager:

```
i.s.s.IdentityInfo: Failed to create info object for identity: null
java.lang.NullPointerException: Cannot invoke "inetsoft.uql.util.Identity.getIdentityID()" because "identity" is null
    at inetsoft.sree.security.IdentityInfo.<init>(IdentityInfo.java:64)
    at inetsoft.web.admin.security.IdentityService.getIdentityInfo(IdentityService.java:523)
    at inetsoft.web.admin.security.IdentityService.deleteIdentities(IdentityService.java:213)
```

## Root Cause

`IdentityService.getIdentityInfo()` passed the result of `provider.getUser()` (or `getGroup()`/`getRole()`/`getOrganization()`) straight into `new IdentityInfo(identity, provider)` with no null check. During a rapid create-then-delete flow, that lookup can transiently return `null` (storage/cache staleness or a concurrent modification). Inside the constructor, `identity.getIdentityID()` then threw an NPE. The constructor's own `try/catch` caught it and logged the misleading `Failed to create info object for identity: null` error, returning an empty `IdentityInfo`. The delete itself still completed (it operates by id, not the fetched object), but the spurious error was reported to users.

## Fix

Add a null guard in `getIdentityInfo`: when the identity is not found, log at `debug` and return an empty `IdentityInfo()`. This yields an object that is field-for-field identical to the prior caught-NPE result, minus the misleading `ERROR` log and the exception. It is the only caller of the `IdentityInfo(Identity, provider)` constructor.

## Test Plan

- In EM, create `user0` with Enterprise Manager permission and Administrator permission on the *users* node.
- Log in as `user0`; repeatedly create and delete users.
- Confirm the `Failed to create info object for identity: null` error no longer appears in the logs and the delete still succeeds.
- Automated: `IdentityServiceTest` passes (7/7); `community/core` builds successfully.

Fixes Redmine #75718.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/security/IdentityService.java
+++ b/core/src/main/java/inetsoft/web/admin/security/IdentityService.java
@@ -520,6 +520,16 @@ public class IdentityService {
          identity = provider.getRole(identityId);
       }
 
+      // The identity lookup can transiently return null (e.g. cache/storage
+      // staleness or a concurrent modification while an identity is being
+      // deleted). Guard against it so IdentityInfo isn't constructed from a
+      // null identity, which would throw an NPE that gets logged as a
+      // misleading "Failed to create info object for identity: null" error.
+      if(identity == null) {
+         LOG.debug("Identity not found, returning empty info: {} (type={})", identityId, type);
+         return new IdentityInfo();
+      }
+
       return new IdentityInfo(identity, provider);
    }
 
--- a/core/src/test/java/inetsoft/web/admin/security/IdentityServiceTest.java
+++ b/core/src/test/java/inetsoft/web/admin/security/IdentityServiceTest.java
@@ -17,8 +17,11 @@
  */
 package inetsoft.web.admin.security;
 
+import inetsoft.sree.security.AuthenticationProvider;
 import inetsoft.sree.security.IdentityID;
+import inetsoft.sree.security.IdentityInfo;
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft:
Thanks for the review. Addressing the two observations:

1. **No test coverage for the new branch** — Fixed in 5c66d3a. Added `getIdentityInfo_nullIdentity_returnsEmptyInfo` to `IdentityServiceTest`, which mocks `AuthenticationProvider.getUser(...)` to return `null` and asserts `getIdentityInfo(...)` returns a non-null, default/empty `IdentityInfo` (null id, inactive, no members). It deliberately avoids `getRoles()` since that touches `SecurityEngine`/`SUtil.isMultiTenant()` statics not initialized in the unit test. All 8 tests in the class pass.

2. **Debug log prints the raw `type` int** — Leaving as-is. It's a `debug`-level message and the `identityId` already identifies the entity, so mapping the int to a readable label isn't worth the added scope for this fix. Noted for future cleanup if it becomes a pain point.

ngovus-inetsoft (approved):

PR #4347, opened 2026-07-21
Bug #75721, re-scope user permission grants on organization rename

jshobe-inetsoft:
## Summary

When an admin clones an organization and then renames it (changing its ID), permissions granted directly to a **user** (e.g. Portal Tabs → Data → Access) were lost in the renamed org. Role and group grants survived; only user grants were dropped.

## Root Cause

During an org-ID change, `IdentityService.setOrganizationInfo()` → `updateOrganizationMembers()` re-scopes the org's members to the new org id. For roles and groups it calls `updateRoleForOrg()` / `updateGroupForOrg()`, which invoke `updateIdentityPermissions(ROLE|GROUP, …, doReplace=true)`. That method iterates every permission in the org and relocates the whole permission storage key from the old org id to the new one (`setPermission(newOrg)` + `removePermission(oldOrg)`), while only re-scoping grantees of the matching identity type.

The **user** branch of `updateOrganizationMembers()` renamed the user identity but never called `updateIdentityPermissions(USER, …)`. As a result, a user-granted permission (e.g. `PORTAL_TAB:organization0:Data → user0`) had its key relocated to the new org by the role/group pass, with the grantee left scoped to the old org id, and the old key deleted — so the grant was effectively lost.

Confirmed at runtime via WARN-level diagnostic logging with stack traces: the clone persisted `PORTAL_TAB:organization0:Data = [user0@organization0]` correctly, and the loss occurred during the rename in `updateRoleForOrg()`'s key relocation.

## Fix

Add the symmetric call in the user rename branch of `updateOrganizationMembers()`, mirroring `updateRoleForOrg()`:

```java
updateIdentityPermissions(Identity.USER, oldID, user.getIdentityID(),
   OrganizationManager.getInstance().getCurrentOrgID(), identity.getId(), true);
```

`oldID` is captured before `user.setOrganization(orgID)`, so it still carries the old org id; `user.getIdentityID()` is the new identity. This re-scopes the user's own grants to the new org id, consistent with how roles and groups are already handled.

## Test Plan

- Create user `user0` in the host org and grant it **Portal Tabs → Data → Access**.
- Clone the host org to `Organization0`, then rename it (name + ID) to `org0`.
- Switch to `org0` and confirm `user0` still has the Portal Tabs → Data access permission. ✅ (verified in a running instance)
- `./mvnw clean install -pl core -am` builds successfully.
- All 365 tests in the `core` security/permission suites pass (including the 8 `PermissionMatrixOrgLifecycleTest` org clone/rename tests) — no regression.

Fixes Redmine #75721.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/security/IdentityService.java
+++ b/core/src/main/java/inetsoft/web/admin/security/IdentityService.java
@@ -859,6 +859,14 @@ public class IdentityService {
                dashboardRegistryManager.migrateRegistry(oldID, securityProvider.getOrganization(OrganizationManager.getInstance().getCurrentOrgID()), identity);
             }
 
+            // Re-scope the user's own permission grants to the new organization, symmetric with
+            // updateRoleForOrg()/updateGroupForOrg(). Without this, permissions granted directly
+            // to the user (e.g. portal-tab access) stay scoped to the old org id and are lost when
+            // the org id changes, because the role/group re-scoping relocates the permission keys
+            // to the new org without carrying the user grantee over. (Bug #75721)
+            updateIdentityPermissions(Identity.USER, oldID, user.getIdentityID(),
+               OrganizationManager.getInstance().getCurrentOrgID(), identity.getId(), true);
+
             eprovider.setUser(user.getIdentityID(), user);
             eprovider.removeUser(oldID);
             repletRegistryManager.renameUser(oldID, user.getIdentityID());
--- a/core/src/test/java/inetsoft/sree/security/PermissionMatrixOrgLifecycleTest.java
+++ b/core/src/test/java/inetsoft/sree/security/PermissionMatrixOrgLifecycleTest.java
@@ -45,16 +45,25 @@ package inetsoft.sree.security;
  * survives untouched, `[landed]`).
  * Scenario 8 (renaming an org's ID and then renaming it back -- round trip -- leaves the final
  * permission state identical to before the round trip, with no orphaned key).
+ * Scenario 9 (renaming an org's ID re-scopes a member's own directly-granted permission to the
+ * new org rather than dropping it -- Bug #75721, the USER + org-ID-change combination that
+ * scenarios 2 (ROLE + org-ID change) and 3 (USER + same-org username rename) did not cover).
  */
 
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft:
Thanks for the review. Responses to each observation:

**1. Regression test — added.** Added `renameOrgId_updateOrganizationMembers_reScopesUserOwnGrantToTargetOrg` (scenario 9) to `PermissionMatrixOrgLifecycleTest`. It drives the private `updateOrganizationMembers()` via `ReflectionTestUtils` — the exact method this fix added the call site to — with a user who holds a directly-granted VIEWSHEET permission, renames the org id, and asserts the grant is re-scoped to the new org (not dropped) with the old-org key removed. Real `securityEngine`/`securityProvider`; `keyValueStorageManager`/`repletRegistryManager`/`dashboardRegistryManager` mocked (favorites move + registry migration are unrelated to re-scoping); current org set via the thread principal.

Deliberately drives `updateOrganizationMembers()` rather than `updateIdentityPermissions()` directly: the fix is an *added call site* inside that method, and its `Identity.USER` branch is pre-existing code already covered by scenario 3. With no roles/groups in the org, the added call is the only thing that touches the user's grant, so the test genuinely fails if the fix is reverted — verified by reverting commit `d313ac62e` (fails at `assertNotNull(after)`, grant orphaned) and restoring (9/9 pass). This closes the specific coverage gap: scenario 2 was ROLE + org-id change, scenario 3 was USER + same-org rename; scenario 9 is USER + org-id change.

**2. Inline `(Bug #75721)` reference in the comment — keeping as-is.** Inline bug-number references are an established convention here (34 across `core/`, plus the commit history and this very test class), so dropping it would make the comment inconsistent with surrounding code. The rationale ("symmetric with role/group re-scoping") is the substance and remains.

**3. `SELF`-org side effect now reachable via rename — no change.** Agreed this is informational. It only triggers if the org being renamed literally has id `SELF`, which isn't a real rename flow, and the USER branch otherwise behaves like the well-covered GROUP/ROLE branches.

**4. Added scan cost — no change.** Agreed negligible and consistent with the existing per-group/per-role scans; org renames are rare admin operations.

ngovus-inetsoft (approved):

PR #4359, opened 2026-07-22
Bug #75729, fix auto layout arrows anchoring to stale table positions

ngovus-inetsoft:
## Summary
- Fixes join lines rendering disconnected/misplaced after using Physical View "Horizontal Layout" / "Vertical Layout" auto layout.
- Root cause: `JoinNodeGraphComponent.ngOnChanges` re-registers reused (`@for`-tracked) table nodes and synchronously re-connects jsPlumb edges before Angular flushes the `style.top`/`style.left` host bindings for the new position. jsPlumb anchors the connection to the DOM position at connect-time, so after auto layout moves every table at once, the arrows stayed anchored to each table's previous position while the table boxes rendered correctly at their new position.
- Fix: eagerly set the node element's `style.top`/`style.left` from the new bounds before re-registering/re-connecting, so jsPlumb always reads the current position.

## Test plan
- [x] Verified in browser: opened a physical view with several joined tables, clicked "Horizontal Layout", confirmed join lines now connect the correct tables at their new positions (previously arrows pointed to stale, empty locations).
- [ ] `npm run test:portal` (frontend unit tests)


jshobe-inetsoft (approved):
Approved: Claude's follow-up review on the latest commit (01d5ed0c) confirms the test regression is fixed and no new issues were introduced, and all checks are green. Squash-merging.

PR #4360, opened 2026-07-22
75643, boxplot outlier should animate with rest of box plot features.

milotalon:
Root Cause:
A box-plot chart is rendered from a single BoxDataSet as two separate visual elements: the box/whisker glyphs (SchemaElement/BoxPainter -> "inetsoft-box" SVG groups) and the outlier data points (PointElement -> "inetsoft-point" groups). The SVG entrance animation emitted a single "box" hint whose injector (injectBoxAnimation) only animated the inetsoft-box groups. The outlier points were never given a fade animation, so they appeared immediately at full opacity while the boxes faded in.

Fix:
The box and its outlier points now share a data-driven "data-group" key (the grouping dimension tuple, which every BoxDataSet row carries -- box rows and outlier rows alike). SchemaVO (box) and the outlier PointVO both stamp this key on their SVG annotation groups. injectBoxAnimation still staggers the boxes left-to-right by screen X (unchanged entrance order) and now uses the shared key to give each outlier point the exact fade-in delay of the box it belongs to. Result: the outlier points fade in gradually together with their box and whisker instead of showing immediately.

### DIFF
--- a/core/src/main/java/inetsoft/graph/element/PointElement.java
+++ b/core/src/main/java/inetsoft/graph/element/PointElement.java
@@ -46,6 +46,13 @@ public class PointElement extends StackableElement {
    public static final String HINT_GANTT_MILESTONE = "gantt_milestone";
 
    /**
+    * Hint marking this point element as the outlier markers of a box plot; the value is
+    * {@code "true"} when set, absent otherwise. Used so the SVG entrance animation stamps a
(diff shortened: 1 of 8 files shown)


jshobe-inetsoft (approved):
Approving based on Claude's automated review (resolved, no blocking issues found) and all checks passing.

PR #4363, opened 2026-07-23
Bug #75745, fix embedded viewsheet thisParameter stringifying as [object Object]

ngovus-inetsoft:
## Summary
- `Viewsheet1.thisParameter` (an embedded viewsheet's parameter scope) rendered as the generic `[object Object]` when concatenated directly into a string, e.g. `'...' + Viewsheet1.thisParameter`.
- Root cause: GraalJS resolves string coercion via a callable `toString`/`valueOf` **member** lookup on the object, never the underlying Java `Object.toString()` override. `ScopeProxy` (the GraalJS bridge for `ScriptScope` objects such as `VariableScriptable`) exposed no such member, so GraalJS fell back to its default `[object Object]`.
- Fix: `ScopeProxy.getMember("toString")` now returns a callable that delegates to the scope's `toString()`. `VariableScriptable.toString()` — previously dead code, never reachable from script — now builds an actual `key=value, ...` listing of its variable table.
- This is a general fix for any `ScriptScope`-backed value stringified directly (also covers the top-level `parameter` global), not just embedded-viewsheet `thisParameter`.

## Test plan
- [x] Verified both modified files compile cleanly against the project sourcepath/classpath
- [x] Reproduced with `VS_embedded.vso`: selecting radio button B on the embedded viewsheet now shows `myBox=B, pid=15.0, ...` instead of `[object Object]` in the `Viewsheet1.thisParameter` output

### DIFF
--- a/core/src/main/java/inetsoft/uql/script/VariableScriptable.java
+++ b/core/src/main/java/inetsoft/uql/script/VariableScriptable.java
@@ -198,11 +198,13 @@ public class VariableScriptable implements ScriptScope {
          String name = (String) names.nextElement();
 
          try {
+            Object val = vars.get(name);
+
             if(sb.length() > 0) {
                sb.append(", ");
             }
 
-            sb.append(name).append('=').append(vars.get(name));
+            sb.append(name).append('=').append(val);
          }
          catch(Exception ignore) {
          }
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):
Approved based on complete Claude automated review with no blocking issues found.

PR #4364, opened 2026-07-23
Bug #75746, fix percent-of-grand-total showing blank after converting a crosstab to a freehand table

ngovus-inetsoft:
## Summary
Converting a crosstab that uses a Calculator-based percentage aggregate (e.g. "Percent of Grand Total") to a freehand/calc table left the Grand Total row's percentage cell blank, while every other row (detail and subtotal) computed correctly.

Two bugs combined to cause this:

- **`CalcFieldFormula.getOriginalResult()`** temporarily forced `percentageType` to `NONE`, called `getResult()` (which caches its return value in `result`), then restored the type — but never invalidated that cache. The grand-total cell's formula instance is used both as the denominator for every other row's percentage *and* as the numerator for its own 100%, so by the time its own percentage was computed, it returned the stale cached raw value instead of `raw / total`.
- **`AbstractCrosstabVSAQuery.getMeasureHeader()`** didn't de-duplicate measure names for Calculator-based aggregates sharing an underlying column, so a raw-sum aggregate and a percent-of-total aggregate on the same column both resolved to the same bare measure name. A calc table's grand-total cells reference their measure by this bare, unqualified name (they have no row/col group to qualify it with), so the second aggregate's lookup missed the disambiguated name (`"name.1"`) the layout conversion actually generates, and silently resolved to `null`.

## Test plan
- [ ] Build a crosstab with row groups and an aggregate using a "Percent of Grand Total" calculator on top of another aggregate on the same column
- [ ] Right-click → Convert to Freehand Table
- [ ] Confirm the Grand Total row now shows 100% for the percent column (previously blank), matching the original crosstab's rendering
- [ ] Confirm detail and subtotal row percentages are unaffected

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
@@ -839,6 +839,34 @@ public abstract class AbstractCrosstabVSAQuery extends CubeVSAQuery
          }
       }
 
+      // Multiple aggregates (e.g. a raw sum and a "percent of total" calculator) can
+      // share the same underlying column and therefore the same measure name here.
+      // The calc table conversion (CalcTableVSAQuery.createCrosstabAssemblies) disambiguates
+      // duplicate aggregate binding values the same way ("name", "name.1", "name.2", ...), and
+      // a calc table's grand-total cells reference the measure by this bare, unqualified name
+      // (no "@group:value" suffix) -- so measureHeaders must use the identical disambiguated
+      // names or TableArray.hasMember() won't find the duplicate and the cell silently
+      // resolves to undefined/null.
+      Map<String, Integer> dupCounts = new HashMap<>();
+
+      for(int i = 0; i < measurename.length; i++) {
+         String name = measurename[i];
+
+         if(name == null) {
+            continue;
+         }
+
+         Integer dup = dupCounts.get(name);
+
+         if(dup == null) {
+            dupCounts.put(name, 1);
+         }
+         else {
+            dupCounts.put(name, dup + 1);
+            measurename[i] = name + "." + dup;
+         }
+      }
+
       return measurename;
    }
 
--- a/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
@@ -816,6 +816,14 @@ public abstract class AbstractCrosstabVSAQuery extends CubeVSAQuery
                                      Map<String, String> calcHeaderMap)
    {
       String[] measurename = new String[formula.length];
+      // per-index original (pre-calc-prefix) name, only set where a calculator applies;
+      // used to rebuild calcHeaderMap below, keyed by the final (post-dedup) measurename,
+      // instead of putting into calcHeaderMap immediately -- two calculator aggregates on
+      // the same column (e.g. SUM(col) and AVG(col) both with "Percent of Grand Total")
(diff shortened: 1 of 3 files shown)


ngovus-inetsoft:
Added a regression test for the `CalcFieldFormula.getOriginalResult()` fix (`CalcFieldFormulaTest.getOriginalResult_thenGetResult_stillComputesPercentageNotStaleRawValue` / `getOriginalResult_doesNotDisturbAlreadyCachedPercentageResult`). Verified both fail against the pre-fix code (return the raw value / clobber an already-cached percentage) and pass with the fix. Also added the missing `@Tag("core")` + Spring context annotations the test class needed relative to its siblings — without them the whole file was excluded from the module's default test run.

Re: the `AbstractCrosstabVSAQuery.getMeasureHeader()` concern — whether the de-dup could change the visible summary-header text on an ordinary, non-converted live crosstab, not just the calc-table conversion path this PR targets. I couldn't get hands-on browser access to click through a live example, so I traced every consumer of `getMeasureHeaders()`/`measureHeaders` in the codebase instead:

- `CrossTabFilter.getHeaders()` → `addSummaryHeaders()` — the visible summary-title row, as you flagged.
- `LayoutTool.getRealMeasureHeader()` (`report/LayoutTool.java:460`) — this one turned out to be the key data point. It takes a column's *displayed* header text and looks it up in `measureHeaders`, with a fallback: if the exact string isn't found, it strips a trailing `.N` suffix and retries. That fallback only makes sense if two aggregates colliding on a bare measure name can *already* end up displayed with a `.1`-style suffix somewhere in the pipeline, independent of this PR — i.e. the codebase already anticipates and partially handles this exact mismatch between the displayed header and `measureHeaders`.
- `CrosstabGroupSelector.getColumName2()` — an exact-match lookup that silently no-ops on a miss, so it can only start matching *more* cases after this change, never fewer.

Given that, and that the de-dup loop is a straightforward hash-map pass that only touches names it finds duplicated (it can't alter or rename anything that was already distinct), I'm fairly confident this is safe: for a live crosstab where a `Calculator` already disambiguates two aggregates on the same column (the common case, and what our own `Crosstab3` repro does), the prefix step runs before the de-dup and nothing collides, so nothing changes. The only scenario where this fix's de-dup would actually change visible output is two aggregates that share a root column *and* neither has a disambiguating calculator prefix — which, per `getRealMeasureHeader`, was arguably already an inconsistent/ambiguous state before this change, not a working one this PR would be breaking.

jshobe-inetsoft (changes requested):
Claude's automated review flagged an unresolved scope-creep concern across both review rounds that hasn't been addressed:

The new de-duplication loop in `AbstractCrosstabVSAQuery.getMeasureHeader()` runs over **all** aggregate entries, not just ones involved in the calc-table-conversion scenario this PR targets. `measureHeaders` also feeds directly into the visible summary-title row for ordinary (non-calc-table) crosstabs via `CrossTabFilter.getHeaders()`/`addSummaryHeaders()`. So two aggregates that happen to share a bare measure name on a plain live crosstab will now be relabeled `name`/`name.1` — a user-visible behavior change beyond the described bug, and it isn't covered by the test plan or any added test.

`CrossTabFilter.initCalcHeaders()` already has a narrower, pre-existing de-dup scoped specifically to `CalcFieldFormula` entries, which suggests scoping this new loop the same way (e.g. `formula[i] instanceof CalcFieldFormula`) would fix the reported bug without touching plain-aggregate headers on ordinary crosstabs.

Could you either scope the fix to calc-table-derived aggregates, or add a test confirming the plain-crosstab dedup behavior is intentional? Everything else in the PR (the `CalcFieldFormula.getOriginalResult()` cache fix) looks solid and well-tested.

jshobe-inetsoft (approved):

PR #4366, opened 2026-07-24
Fix embed-elements build: inline chunk-*.js to avoid broken cross-file import

Zxm-happy:
## Summary
- Angular's esbuild `application` builder now splits `main-*.js` (for the `elements` and `viewer-element` projects) into a main bundle plus one or more `chunk-*.js` files whenever anything reachable from the entry point — even transitively, through a dynamically-imported dialog/component the embed elements never actually navigate to — uses `import()`. These chunks are linked from `main-*.js` via native ES module `import` statements.
- The `elements:scripts` / `viewer-element:scripts` gulp tasks wrapped `main-*.js` in a plain IIFE and concatenated it with polyfills/scripts, with no handling for `chunk-*.js`. `scripts/build-embed-package.mjs` then only copies `elements.js` / `viewer-element.js` into the published npm package, leaving the chunk file out entirely.
- At runtime, `elements.js` is loaded as `<script type="module">` (see `loader.mjs`). It hits the dangling `import ... from "./chunk-XXXX.js"`, which 404s — a parse-time failure for the whole module. This silently prevents every `customElements.define()` call in the file from ever running, so `inetsoft-chart`/`inetsoft-crosstab`/`inetsoft-table`/etc. are left as undefined custom elements and render as empty space wherever they're embedded (reproduced via the wiz AI-chat portal, where charts rendered fine with the old non-chunked package but went blank after swapping in a freshly built one).
- Added `gulp/lib/bundle-main-script.js`, which re-bundles `main-*.js` with esbuild so any `chunk-*.js` it depends on is inlined into one self-contained IIFE before concatenation. Wired it into both the `elements.js` and `viewer-element.js` gulp tasks, since both projects use the same builder and are equally exposed to this class of bug. Added `esbuild` as an explicit devDependency (previously only pulled in transitively via `@angular/build`).

## Test plan
- [x] `npx gulp elements:scripts` against existing `ng build elements` output → `node --check` passes on the resulting `elements.js`, zero bare `import` statements remain, all 7 `customElements.define(...)` calls (`ai-assistant`, `inetsoft-chart`, `inetsoft-crosstab`, `inetsoft-gauge`, `inetsoft-image`, `inetsoft-table`, `inetsoft-text`) are present.
- [x] `node scripts/build-embed-package.mjs` → same checks pass on the published `target/npm/stylebi-embed-elements/elements.js`.
- [x] Copied the rebuilt package into a wiz portal `node_modules` checkout and re-verified `node --check` / zero-import / element-registration on the copy.
- [ ] Manual verification in a browser that `<inetsoft-chart>` renders inside the wiz AI-chat portal (not done in this session — no browser automation available; needs a human check).


Zxm-happy:
Addressed all 4 points from the automated review in ed4728e:

1. **package-lock.json / esbuild range mismatch** — reproduced the failure first: with the original `^0.27.3` range, `npm ci --dry-run` fails with `EUSAGE` / `Invalid: lock file's esbuild@0.28.1 does not satisfy esbuild@^0.27.3` (the lockfile already transitively resolves `esbuild@0.28.1` via `@angular/build`). Widened the range to `^0.27.0 || ^0.28.0` and added the matching entry to package-lock.json's root importer `devDependencies`. Note: running a full `npm install` locally to regenerate the lockfile produced a ~30k line unrelated diff (looks like a local npm-version formatting difference vs whatever generated the checked-in lockfile), so this is a minimal, hand-applied one-line insertion matching what's already resolved, rather than a full regeneration. Re-verified `npm ci --dry-run` no longer reports any esbuild-related error.
2. **`legalComments`** — changed `"none"` → `"inline"`, matching `extractLicenses: false` in angular.json so third-party license notices stay in the published file (spot-checked: `@license`/`Copyright` comments are present in the rebuilt `elements.js`).
3. **IIFE binding consistency** — `bundleMainScript` now bundles as `format: "esm"` and manually wraps the output as `` (function(){...}).call(window); `` (same convention as `wrapInIIFE`), instead of relying on esbuild's own `"iife"` format (which binds nothing). Verified the wrapped output still passes `node --check` and still registers all `customElements.define(...)` calls.
4. **glob.sync safety** — `bundleMainScript` now throws if `main-*.js` matches more than one file, symmetric with the existing zero-match check. Verified by dropping a stale duplicate `main-*.js` into the output dir and confirming it throws instead of silently picking one.

Re-ran the full pipeline (`gulp elements:scripts viewer-element:scripts` → `node scripts/build-embed-package.mjs`) after these changes: both `elements.js` and `viewer-element.js` still pass `node --check`, have zero bare `import` statements, and register all expected custom elements (`ai-assistant` + 6 embed components in elements.js; `ai-assistant` + `inetsoft-viewer` in viewer-element.js).

PR #4370, opened 2026-07-24
feat(wiz): persist and inherit worksheet column descriptions

NeverStopFMH:
Add a `description` slot to aggregate/group/window column specs in the /ws/table request model and persist each onto its output column so /ws/structure returns it (consumed by data insight).

- Physical columns and mirror expression/aggregate/window columns carry the model-authored description onto their output ColumnRef.
- Join passthrough columns now inherit the base column annotation (initCompositeColumnSelection), matching MirrorTableAssembly, so the annotation is not lost at each join hop.
- Group-by dimensions keep their inherited annotation rather than being overwritten with a generated lineage description; only genuinely derived columns (aggregate measures, expression, window) take the model description.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/model/WorksheetTable.java
+++ b/core/src/main/java/inetsoft/web/wiz/model/WorksheetTable.java
@@ -211,6 +211,8 @@ public class WorksheetTable {
       /** ORDER BY clauses, in order; may be omitted/empty. */
       private List<OrderByInfo> orderBy;
       private String type;
+      /** Business meaning of this window column (english); persisted onto the output column. */
+      private String description;
       /** ROWS frame, e.g. {@code 2 PRECEDING .. CURRENT ROW}; omit for the function's default. */
       private WindowFrameInfo frame;
 
@@ -228,6 +230,8 @@ public class WorksheetTable {
       public void setOrderBy(List<OrderByInfo> orderBy) { this.orderBy = orderBy; }
       public String getType() { return type; }
       public void setType(String type) { this.type = type; }
+      public String getDescription() { return description; }
+      public void setDescription(String description) { this.description = description; }
       public WindowFrameInfo getFrame() { return frame; }
       public void setFrame(WindowFrameInfo frame) { this.frame = frame; }
    }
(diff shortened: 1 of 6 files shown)


NeverStopFMH:
Thanks for the review. Addressed in the latest commit:

**1. No test for the "prefer inherited description" behavior — FIXED**
Added `WorksheetTableServiceAggregateDescriptionTest.groupKeepsInheritedDescriptionOverModelDescription`: it seeds `COMPANY_NAME`'s private `ColumnRef` with an inherited annotation before `applyAggregateInfo`, passes a competing model group `description`, and asserts the output keeps the **inherited** annotation. Same test also confirms the aggregate measure (`TOTAL_ORDERS`) still takes the model description.

**2. No test for join passthrough description inheritance — FIXED**
Added `WsServiceHelperCompositeDescriptionTest`, driving `initCompositeColumnSelection` directly (same package-private pattern as `applyWindowColumns`/`applyAggregateInfo`): a join over two physical tables carries each base column's description onto the composite selection, and a column with no base description does not gain one.

**3. Silent-drop edge case in the group description lookup — DOCUMENTED**
Added a comment marking it as accepted best-effort. A fallback to the public `column` (as the aggregate branch does) would not actually help here: the public selection is regenerated as clones of the private selection, so the description must live on the private target to survive — setting it on the public `column` would be discarded. A miss is unexpected (the group was already resolved against `cs` just above) and rare, so dropping the description is the accepted behavior, consistent with the rest of the method.

**4. Overlapping group-by / aggregate source field — acknowledged, no change**
Agreed this is a narrow, pre-existing structural limitation (the same private `ColumnRef` object backing both a `GroupRef` and an `AggregateRef` predates this PR) and, as noted, a note rather than a blocker. Left as-is to avoid complicating the normal aggregate-overwrite path.

Test results (local, `mvn surefire:test`):
- `WsServiceHelperCompositeDescriptionTest`: 2/2
- `WorksheetTableServiceAggregateDescriptionTest`: 3/3
- `WorksheetTableServiceWindowColumnsTest`: 24/24

fredaan (approved):

PR #4387, opened 2026-07-24
Bug #75747, fix data missing when converting crosstab to freehand table

ngovus-inetsoft:
## Summary
- Converting a crosstab with a nested date-grouped row dimension (e.g. `Month(ndate)` nested under another dimension) to a freehand table produced a table with only the header row — all data rows were missing.
- Root cause: `LayoutTool.createCrosstabCalcExpression` built the generated `toList()`/`none()` script expressions from `TableCellBinding.getValue()`, which for date-grouped dimensions holds the dimension's display full name (e.g. `"None(Month(ndate))"`) instead of the physical column name backing the crosstab data (`"Month(ndate)"`). The generated formula referenced a nonexistent column, the lookup returned `null`, and the row group never expanded.
- Fix: resolve the real column name from the already-computed `CalcGroup` (via `BindingTool.getPureField(group).getAttribute()`) instead of trusting the raw binding value, applied only to group cells (aggregate/summary cells are unaffected, since they aren't part of `groups[]`). `getAttribute()` is used rather than `getName()` since the latter can carry an entity-qualified table prefix that also doesn't match the crosstab's unqualified column names.
- Also added two previously-missing test-only Spring bean registrations (`VSCrosstabModel`, `VSGroupContainerModel` factories) to the shared `IntegrationTestConfiguration` test config, needed to open a real viewsheet fixture during investigation — harmless, reusable addition following the file's existing per-assembly-type curation pattern.

## Test plan
- [x] Verified on a live running server: before the fix, converting the crosstab produced zero data rows (header only); after the fix, the converted freehand table correctly shows all rows grouped by the outer dimension and nested date group, with correct aggregate values.
- [x] `./mvnw -pl core compile` passes.
- [ ] No automated regression test included (tracked separately; flagged in code review).


jshobe-inetsoft (approved):

PR #4392, opened 2026-07-27
Apply embed container size to table/crosstab and keep tracking it across type switches

Zxm-happy:
## Summary

Two related bugs in the wiz embed sizing path, both surfacing as blank space in the embed container around a table/crosstab that renders smaller than the container it's given:

- **`CoreLifecycleService#applyEmbedChartSize` only sized charts.** It pushed the embed container's pixel size onto `ChartVSAssemblyInfo` only. A programmatically-created table/crosstab/calc-table (`TableDataVSAssemblyInfo`) kept its own default size instead of the container's actual size, leaving blank space around the smaller-than-container table. Fixed by also sizing `TableDataVSAssemblyInfo` the same way.

- **Switching a wiz visualization's type (e.g. table → crosstab) still left the newly-displayed assembly undersized**, even with the fix above. `WizAutoBindingService#changeType` doesn't mutate the existing assembly in place — it replaces the viewsheet's primary assembly with a differently-named one on the same runtime. The embed component always resends an `embed=true` `RefreshVSAssemblyEvent` with its *current* assembly name after remounting, but `VSRefreshService#refreshVsAssembly` only ever wrote `EmbedAssemblyInfo#setAssemblyName` the *first* time an `EmbedAssemblyInfo` was created for a runtime — every later type switch kept comparing against the original, now-replaced assembly's name in `applyEmbedChartSize`, so the size silently stopped being applied after the very first switch (in either direction). Fixed by re-pointing the tracked assembly name on every explicit embed refresh, not just the first.

## Test plan

- [x] Added `CoreLifecycleServiceEmbedSizeTest` — covers chart and crosstab sizing via `applyEmbedChartSize`
- [x] Added `VSRefreshServiceEmbedAssemblyTest` — simulates a table → crosstab type switch (same runtime, new assembly name) and asserts the embed target's tracked assembly name follows the switch
- [x] Both tests verified to fail without their corresponding fix and pass with it (temporarily reverted each fix, confirmed the new test fails with the expected assertion message, then restored the fix and confirmed it passes)
- [x] `mvn -pl core -am test` for both new test classes green

### DIFF
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/VSRefreshService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/VSRefreshService.java
@@ -115,6 +115,15 @@ public class VSRefreshService {
                   embedAssemblyInfo.setAssemblyName(event.getAssemblyName());
                   rvs.setEmbedAssemblyInfo(embedAssemblyInfo);
                }
+               // The embedded assembly can be swapped out for a differently-named one on the
+               // same runtime (e.g. WizAutoBindingService#changeType removes the old primary
+               // assembly and places a new one when the user switches viz type). Re-point the
+               // tracked name on every explicit embed refresh, or applyEmbedChartSize()'s name
+               // check keeps comparing against the original (now-gone) assembly forever and
+               // silently stops sizing the new one.
+               else if(embedAssemblyInfo != null && event.getEmbed()) {
+                  embedAssemblyInfo.setAssemblyName(event.getAssemblyName());
+               }
 
                if(embedAssemblyInfo != null) {
                   embedAssemblyInfo.setAssemblySize(event.getAssemblySize());
--- a/core/src/main/java/inetsoft/web/viewsheet/service/CoreLifecycleService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/service/CoreLifecycleService.java
@@ -1584,7 +1584,8 @@ public class CoreLifecycleService {
       }
    }
 
-   // This only applies to embedded charts; other assembly types intentionally remain unchanged.
+   // Applies to embedded charts and table-data assemblies (table/crosstab/calc-table); other
+   // assembly types intentionally remain unchanged.
    private void applyEmbedChartSize(RuntimeViewsheet rvs, VSAssembly assembly) {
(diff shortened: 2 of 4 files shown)


Zxm-happy:
Addressed the automated review's feedback:

**1. Repurposing "max mode" for table/crosstab sizing** — confirmed intentional and not new behavior: `ChartVSAssemblyInfo` already used `setMaxSize()` this exact way for embedded charts before this PR, and `BaseTableModel`'s `maxMode = (maxSize != null)` inference is pre-existing shared logic, not something this PR changes. This PR just extends the *same* existing mechanism to `TableDataVSAssemblyInfo`. No code change needed, but agreed it's worth a manual check — will verify no table/crosstab-specific max-mode-only UI (e.g. a restore button) shows up unexpectedly in the wiz embed view.

**2. Stale test comment (100x20 vs actual 400x240 default)** — fixed in 8b2dec5cc. `CrosstabVSAssemblyInfo`'s constructor sets `Dimension(400, 240)`, not the `AssetUtil` 100x20 default the comment claimed. Didn't affect the test's correctness (it only asserts the size *after* the fix runs), just the comment.

**3. Readability nit (hoist the null check)** — fixed in 8b2dec5cc, folded `assemblySize == null` into the preceding name-match `if` instead of checking it after `assembly.getInfo()`. No behavior change.

Re-ran both new tests locally after these changes (`CoreLifecycleServiceEmbedSizeTest`: 3/3 pass, `VSRefreshServiceEmbedAssemblyTest`: 1/1 pass).

**On the "build" check failure**: unrelated to this PR. The failure is `DashboardRegistryOrgLifecycleTest`/`OrgLifecycleRepletRegistryIntegrationTest` (org-rename dashboard/replet registry rewriting — a completely different subsystem than viewsheet embed sizing). It passes cleanly when run in isolation (`4 run, 0 failures`), and the same test family has failed with *different* specific assertions on unrelated, already-merged PRs' CI runs before this one existed (e.g. run 30146016012 on `fix/wiz-dashboard-chart-pixel-size`, two days prior; run 30234320356 on `fix/embed-css-scope-and-table-export-url` before any of these commits landed) — a known-flaky test in the full suite (likely shared/static state across the org-lifecycle test family), not a regression introduced here.

fredaan (approved):

PR #4397, opened 2026-07-27
feat(wiz): honor copy and assemblyName on chart-type-change/autoBinding

Nelson-cao:
## Summary
A chat session shares ONE output viewsheet/runtime across every turn — each new visualization is a new assembly inside it, not a separate viewsheet. The standard create/rebind branch of `createViewsheetInternal` always replaced whichever assembly happened to be primary (demote + delete), with no way for a caller to either (a) keep the displaced assembly instead of deleting it, or (b) target a specific, non-primary assembly by name. This blocked two fixes on the `wiz-services` side (paired PR: inetsoft-technology/stylebi-wiz#1319):
- every brand-new visualization request in a session was deleting the previous turn's chart, making conversation history impossible
- clicking "change type" on a historical (non-latest) chart card silently mutated the LATEST card instead, since there was no way to address a specific assembly by name

## Changes
- `ChangeTypeRequest`/`AutoBindingRequest`/`CreateVisualizationModel` gain a `copy: boolean` field (first commit) and an `assemblyName`/`targetAssemblyName` field (second commit).
- `copy=true`: the displaced primary assembly is demoted but not deleted, so it survives as history.
- `assemblyName`/`targetAssemblyName` set: `createViewsheetInternal`'s standard branch swaps that SPECIFIC named assembly in place under the same name, carrying over its exact prior primary state, and never touches any other assembly's primary flag — instead of the "demote whichever is primary" path.
- Both fields are forwarded through `WizAutoBindingService.changeType()` (fast path + `autoBindingInternal` fallback) and `autoBindingInternal()` itself into `CreateVisualizationModel`.
- Rollback-on-exception handling updated to match: a failed targeted replace re-adds the original assembly under its own name with its original primary state.
- Both fields default to false/null, so every existing caller that never sets them is unaffected (in-place delete-and-replace, the existing UI-click behavior).

## Test plan
- [x] `mvnw -pl core -am compile` clean
- [ ] No new automated tests added (per this session's explicit direction: extend tests only for already-tested paths; the standard create/rebind branch and `changeType()` had no prior test coverage)
- [ ] End-to-end verification against the paired `wiz-services`/`portal` PR (inetsoft-technology/stylebi-wiz#1319)


Nelson-cao:
Addressed the review feedback:

1. **Ambiguous "Assembly not found" error** — fixed. Split into two distinct messages: truly-absent vs. present-but-wrong-type.
2. **Targeted-replace mutation outside the rollback-protected `try`** — no code change. Verified this mirrors the pre-existing `previousPrimaryAssembly` demotion pattern a few lines above (also mutates before the `try`) — same class of narrow, pre-existing risk window, not a new one introduced here. Restructuring the try boundary would be a broader refactor touching existing behavior, out of scope for this targeted feature.
3. **Implicit `copy`/`targetAssemblyName` interaction** — fixed. Added a comment directly at the `existingTarget` branch in `WizVsService.java` noting `model.isCopy()` is bypassed there, not just in the model class's javadoc.
4. **Test coverage** — no change, as noted in the PR description: this is a deliberate scope decision (extend tests only for already-tested paths) for this session, not an oversight.

New commit: 2bf2b5f63

Nelson-cao:
Rebased on the updated base branch, which now contains #4398 — and removed the duplication that created.

#4398 added `CreateVisualizationModel.assemblyName` to name the chart the **modificationOnly** path acts on. This PR had independently added `targetAssemblyName` to name the chart the **standard create/rebind** path replaces in place. Two String fields on one model, both answering *"which existing chart does this call address, by name?"*, differing only in which branch reads them.

Consolidated onto the single `assemblyName` field (`95ea500`):

- `modificationOnly` is decided by `config`/`primaryAssembly`/`conditionModel` and never by the name, so the two paths are mutually exclusive within a request — one field is unambiguous in both.
- `getAssemblyName()`'s javadoc now documents both meanings (chart to MODIFY vs. chart to REPLACE), and `isCopy()`'s notes that it is not consulted when the standard path is given a name.
- Every other Wiz endpoint that names an assembly (`/viewsheet/changeType`, `/viewsheet/highlight`, chart colors/format) already calls the field `assemblyName`. Keeping a second, differently-named field on only `/viewsheet/create` meant a caller setting `assemblyName` and expecting standard-path targeting would have had it silently dropped — the failure mode is a wrong chart replaced, with no error.

No behavior change on either path; `core` compiles clean.

The wiz-services counterpart (sending `assemblyName` instead of `targetAssemblyName`) is https://github.com/inetsoft-technology/stylebi-wiz/pull/1330 — these must merge together.

fredaan (approved):

PR #4398, opened 2026-07-27
feat(wiz): let the modify-only path name the chart to filter

fredaan:
## Problem

`/viewsheet/create`'s `modificationOnly` branch always resolved its target with `findPrimaryAssembly` — whichever chart was created or copied most recently.

That is correct while a conversation edits its newest chart, but wiz now resolves edits to a *specific* chart: a user can say "add a filter to the customer-count-by-state chart" three charts later. The `conditionModel` is then built against **that** chart's fields and applied to a **different** chart. At best the wrong chart is filtered; at worst the condition references columns the newest chart does not bind and the request fails.

There was no way for the caller to say which chart it meant.

## Change

**`CreateVisualizationModel.assemblyName`** (optional). Null keeps the existing primary-assembly resolution, so every current caller behaves exactly as before — the wiz side only sends it when it has a reason to override the default.

**`duplicateAssembly`, split out of `duplicatePrimaryAssembly`.** The existing method could not be reused for the copy-then-apply step:

- it requires `source` to *be* the primary, so with `source` an earlier chart the guard rejects the call outright;
- and it demotes `source` itself. Simply lifting the guard would demote nothing (an earlier chart is already non-primary) while still promoting the copy — **leaving two primaries in one viewsheet**.

`duplicateAssembly` demotes whichever assembly actually holds primary and places no constraint on `source`. `duplicatePrimaryAssembly` keeps its guard and delegates: that guard protects `setChartFormat` / `setChartColors` / `applyHighlight`, whose `assemblyName` comes from the client and may be stale, where "duplicate whatever happens to be primary" would silently act on an unrelated chart. Those three callers are untouched.

**Rollback bookkeeping follows.** `modificationOnly` now records the *previously-primary* assembly instead of `source`, since after this change they are no longer the same assembly — restoring `source` would promote the wrong chart. It also tolerates a viewsheet that had no primary to restore.

## Compatibility

Additive. Existing callers omit the field and get the current behaviour unchanged. The wiz client sends it only for a copy it just made or an explicitly resolved target chart; against a server predating this change the property is ignored (`@JsonIgnoreProperties(ignoreUnknown = true)`), degrading to primary-assembly resolution.

## Verification

`./mvnw -o compile -pl core` passes. Not yet exercised end-to-end — that needs a rebuilt server; the wiz-side counterpart is inetsoft-technology/stylebi-wiz#1321.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/model/CreateVisualizationModel.java
+++ b/core/src/main/java/inetsoft/web/wiz/model/CreateVisualizationModel.java
@@ -101,6 +101,26 @@ public class CreateVisualizationModel {
    }
 
    /**
+    * Which chart the "modificationOnly" path acts on (see
+    * {@code WizVsService.createViewsheetInternal}). Null — the default and every pre-existing
+    * caller — means the viewsheet's current PRIMARY assembly, i.e. the chart created or copied most
+    * recently.
+    *
+    * <p>Naming one is required when the caller edits a chart that is NOT the newest: in a
+    * multi-chart conversation a filter built against an earlier chart's fields would otherwise be
+    * applied to a different chart, which at best filters the wrong chart and at worst references
+    * columns that chart does not bind. Ignored outside the modificationOnly path (the standard
+    * create/rebind path produces its own fresh assembly).
+    */
+   public String getAssemblyName() {
+      return assemblyName;
+   }
+
+   public void setAssemblyName(String assemblyName) {
+      this.assemblyName = assemblyName;
+   }
+
+   /**
     * #75456: row cap for sampled-preview mode. Null or &lt;=0 = full data (the default and the
     * agent path); &gt;0 = aggregate at most this many detail rows (faster on heavy/non-mergeable
     * sources, but Sum/Count may be approximate).
@@ -117,6 +137,7 @@ public class CreateVisualizationModel {
    private VisualizationConfig config;
    private String runtimeId;
    private String viewsheetIdentifier;
+   private String assemblyName;
    private VisualizationConditionModel conditionModel;
    private Integer sampleMaxRows;
    private transient VSAssembly primaryAssembly;
--- a/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
@@ -1088,19 +1088,17 @@ public class WizVsService {
             VSAssembly modAssembly = sourceAssembly;
 
             // Copy-then-apply: duplicate the source BEFORE applying the condition, so the original is
-            // left untouched and the filter lands on a new, parallel copy instead. Captured BEFORE the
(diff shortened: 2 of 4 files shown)


fredaan:
All three points addressed in 9aaeadbd0.

## 🔴 Broken tests — confirmed and fixed

You were right, and the failure is worse than "the stub is dead": the run errors before it even reaches the fallback. `duplicateAssembly` runs for real against the mocked `ChartVSAssembly`, and `rebindAssembly` NPEs on `getVSAssemblyInfo().clone()` returning null. Actual result before the fix:

```
Tests run: 6, Failures: 3, Errors: 2
```

Stubs and the `never()` verification now target `duplicateAssembly`; `copyFalseNeverCallsDuplicatePrimaryAssembly` renamed to match. The class doc notes why this path uses `duplicateAssembly` — the chart it copies need not be primary.

I should have caught this: I ran `./mvnw -o compile -pl core` and treated a clean compile as verification. A signature change with Mockito stubs is exactly the case a compile cannot catch.

## 🟢 The implicit invariant — taken, and it simplified the fix

`duplicateAssembly` now returns `AssemblyDuplication(copy, demoted)`. This removes the `findPrimaryAssembly`-twice pattern you flagged: the caller no longer re-derives the demoted assembly, so there is nothing for a future edit between the two call sites to desynchronize. `duplicatePrimaryAssembly` keeps its `VSAssembly` return and unwraps `.copy()`, so its three callers are untouched.

## 🟡 Coverage for the new behaviour — added

New `WizVsServiceDuplicateAssemblyTest` (real `Viewsheet`/`ChartVSAssembly`, per the sibling test's note about factory dispatch):

- a non-primary source is still duplicated, and the assembly that loses primary is the one that **held** it, not the source
- exactly one primary remains afterwards — the two-primaries corruption this split exists to prevent
- the duplicated source stays in the viewsheet
- a viewsheet with no primary reports `demoted == null` — the case the rollback branch must tolerate
- a primary source behaves as before (pins the shared path now that `duplicatePrimaryAssembly` delegates)

`WizVsServiceFilterCopyTest` gains the `assemblyName` resolution cases: absent → primary (and `getAssembly` never consulted), explicit → that chart, unknown → `IllegalStateException` naming it rather than silently filtering the primary.

## Verification

```
WizVsServiceApplyHighlightCopyTest        6/6
WizVsServiceDuplicateAssemblyTest         6/6  (new)
WizVsServiceDuplicatePrimaryAssemblyTest  6/6
WizVsServiceFilterCopyTest                9/9  (was 6, +3)
Tests run: 27, Failures: 0, Errors: 0 — BUILD SUCCESS
```

Also noted your point about an `assemblyName` pointing at a non-`DataVSAssembly`: `applyConditionModel` warns and no-ops, so a `copy=true` request would still create and promote a copy while applying nothing. Agreed it is not a regression (`findPrimaryAssembly` carried the same risk), and I have left it alone rather than widen this PR — happy to add an explicit type check if you would prefer it here.

PR #4399, opened 2026-07-27
SQL Query Dialog Simple-Mode Redesign - recovery-74519-0623

swaker854:
Summary

Redesigned simple-query-pane with a fixed left data-source rail and a single scrollable right pane covering Columns, Joins, Conditions, and SQL — replacing the previous multi-column Bootstrap grid layout.
Manual SQL edit mode: visual builder sections now hide when sqlEdited is true, the SQL textarea expands to fill the panel, the toggle button is renamed "Manual Edit" with a clearer Proceed/Cancel confirmation, and "Parse Now" is renamed "Validate SQL" with a success/error status badge using --inet-success/--inet-danger palette tokens.
Fixed a phantom second scrollbar in the slide-out panel by adding z-index: 1 to its :host so it renders above the worksheet canvas's GPU compositor layer, plus scroll fixes for the advanced-mode Fields and Conditions tabs.
Redesigned the VPM condition dialog: replaced the custom field-combo dropdown with custom-select, added an inline expression editor launched from a formula-editor dialog, and reflowed the condition builder/editor layout and button styling.
Follow-up fix: corrected flex-shrink/min-width constraints on the simple-mode nav outlet and tab panes so the layout no longer overflows or collapses at narrow widths, and scoped the .tab-content styling to .simple-query-nav-outlet so it doesn't bleed into other ngbNavOutlet usages elsewhere in the app.

Test plan

 Open a SQL query in simple mode; verify Columns/Joins/Conditions/SQL sections render and scroll correctly at both wide and narrow dialog widths
 Toggle Manual Edit on/off and confirm the SQL textarea fills the panel and the visual builder sections hide/reappear correctly
 Run Validate SQL against valid and invalid SQL and confirm the success/error badge displays correctly
 Open the VPM condition dialog and verify field selection, expression editing, and junction/list controls work as before
 Open a worksheet slide-out panel and confirm no phantom scrollbar appears over the canvas
 Confirm the advanced-mode Fields and Conditions tabs still scroll correctly


jshobe-inetsoft (changes requested):
## Code Review — SQL Query Dialog Simple-Mode Redesign (#4399)

Reviewed all 22 changed files (752+/445-) against `epic-74519`. The visual direction is a clear improvement, and the added `NgIf/NgSwitch*` imports in `vpm-condition-item-pane.component.ts` fix a genuinely missing set of imports. However, there are several blocking issues — one of them a permission-enforcement regression — plus a test that this change invalidates.

**Requesting changes** for the reasons listed at the bottom.

---

## Blocking

### 1. Free-form SQL permission gate removed (security regression)

`web/.../sql-query-dialog/simple-query-pane.component.html:138-158`

Before:

```html
@if (model.sqlEdited && freeFormSqlEnabled) {
  <textarea #sqlTextArea class="form-control h-100" [(ngModel)]="model.sqlString" ...>
}
```

After:

```html
@if (!model.sqlEdited) {
  <pre class="sqp-sql-pre">{{model.sqlString}}</pre>
  @if (freeFormSqlEnabled) { <button ...>_#(Manual Edit)</button> }
} @else {
  <textarea #sqlTextArea class="form-control sqp-sql-textarea"
    [(ngModel)]="model.sqlString" (ngModelChange)="textChanged()" ...>
  </textarea>
  ...<button (click)="getSqlParseResult()">_#(Validate SQL)</button>
}
```

The `@else` branch is now gated only on `model.sqlEdited`, not on `freeFormSqlEnabled`. `sqlEdited` is a persisted property of the model, so a query saved in manual-SQL mode by a privileged user and reopened by a user **without** `ResourceType.FREE_FORM_SQL` now renders a fully editable SQL textarea plus the Validate button, and edits are written straight back into `model.sqlString`.

That flag is a real permission, not a cosmetic one:

- `core/.../SQLQueryDialogController.java:111` and `QueryManagerService.java:439,823` — `securityEngine.checkPermission(principal, ResourceType.FREE_FORM_SQL, "*", ResourceAction.ACCESS)`
- The save path `SQLQueryDialogController.setModel()` (`@MessageMapping("/ws/dialog/sql-query-dialog-model")`) only re-checks `PHYSICAL_TABLE`, **not** `FREE_FORM_SQL`.

So the client-side gate is the only enforcement, and this PR removes it. Please restore it — e.g. render the textarea `readonly` (or fall back to `<pre>`) and hide "Validate SQL" when `!freeFormSqlEnabled`.

### 2. Existing test is invalidated and not updated

`web/.../sql-query-dialog/simple-query-pane.component.risk.tl.spec.ts:49-77` — *"should preserve table order when condition columns resolve out of order"*:

```ts
comp.columnCache = { Orders: first.asObservable(), Customers: second.asObservable() };
comp.editConditions();
second.next([makeColumnEntry("Customers", "name")]);
expect(comp.conditionFields.map(f => f.name)).toEqual(["Customers.name"]);   // ← now []
```

The new `forkJoin(cacheKeys.map(k => this.columnCache[k].pipe(take(1))))` (`simple-query-pane.component.ts:261`) only assigns `conditionFields` after **every** source has emitted, so the incremental assertion above fails, and the final `conditionDialogModel.fields` assertion (`["Customers.name", "Orders.id", "Orders.state"]`) no longer matches the new cache-key ordering either.

The rewrite itself is defensible — the old incremental-splice logic was hard to follow — but the spec that pins the old contract has to be rewritten in the same PR. Note that the green `build` check does **not** cover this: per `CLAUDE.md`, the portal testing-library suite (`npm run test:portal:tl`) is not yet wired to CI, so this will land silently broken.

### 3. `forkJoin` has no error handler — Conditions dialog can become unopenable

`simple-query-pane.component.ts:261-274`

```ts
forkJoin(cacheKeys.map(k => this.columnCache[k].pipe(take(1))))
   .subscribe((results: AssetEntry[][]) => { ... openDialog(); });
```

`columnCache[tableName]` holds cold HTTP observables (`controller.getTableColumns(entry)`, see lines 133 / 459). `forkJoin` errors as soon as **any** source errors, and there is no `error` callback — so a single failed `getTableColumns` request means `openDialog()` is never reached and clicking **Edit** does nothing at all, with no message to the user. Previously each subscription was independent, so a failure only cost that table's fields; the dialog still opened.

There is also no loading affordance: the dialog now opens only after all requests resolve, so on a slow connection the button appears dead for the duration.

Please add an `error` handler that surfaces the failure (or degrades to opening the dialog with whatever resolved), and consider a loading mask.

### 4. Empty-columns drop-zone styling can never apply (dead CSS)

`simple-query-pane.component.scss:210-214`

```scss
.sqp-columns-list--empty .sql-query-dialog-list__surface {
  min-height: 48px;
  border-style: dashed !important;
  background: transparent !important;
}
```

`.sqp-columns-list--empty` is on the `<sql-query-dialog-list>` host in *this* component's template, but `.sql-query-dialog-list__surface` lives in the **child** component's template (`sql-query-dialog-list.component.html:18`). `SqlQueryDialogListComponent` uses the default `ViewEncapsulation.Emulated`, so this rule compiles to `… .sql-query-dialog-list__surface[_ngcontent-<parent>]` and never matches. The dashed empty drop target — the whole point of the `--empty` class — will not render. (The two `!important`s suggest this was already being fought.) Needs `::ng-deep`, or the surface styling should move into the child component driven by an `@Input`.

### 5. New user-facing strings have no catalog entries

Only `Manual Edit` and `Validate SQL` were added to `srinter.properties`. These new strings have no key, so they will render untranslated in every non-English locale:

`DATA SOURCES`, `drag to add`, `selected`, `Drag a table or column from the left panel to add`, `+ New`, `Not all tables are joined - a cross join may result. Click + New to add join conditions.`, `No joins - add a second table, then link their keys.`, `Add columns before defining conditions`, `editable`, `auto-generated`, `No filters set`, `to add one.`, `Junction`.

### 6. Failing CI check

`claude-review` is failing (`Reached maximum number of turns (50)`, run 30275151942). It looks like an infrastructure/turn-limit failure rather than a code defect, but the check is red and the PR is `BLOCKED`, so it needs to go green (or be re-run) before merge.

---

## Non-blocking

**Correctness / behavior**

- **"Validate SQL" is a misleading rename.** The old markup carried `title="_#(designer.qb.parseToolTipDetails)"` and `[helpLink]="'userhelpVPMParse'"`; that catalog string reads *"SQL Parsing does not determine validity, but a non-parseable query may have other impacts."* The new UI drops both the tooltip and the help link while renaming the button to "Validate SQL", which now promises exactly the thing the removed disclaimer said it does not do. Either keep the disclaimer/help affordance or pick a name closer to what it does (e.g. "Parse SQL").
- **Cross-join heuristic has false negatives.** `@if (numTables > 1 && (model.joins?.length ?? 0) < numTables - 1)` — three tables with two joins between the *same* pair still leaves one table unjoined but suppresses the warning. Acceptable as a hint, but the server-side `common.sqlquery.cartesianJoin` check remains the real guard.
- **Empty controls row.** `.sqd-controls-row` in `sql-query-dialog.component.html:27` is now rendered unconditionally; when `subQuery && !tables` both children are absent and it contributes 12px of dead space (`padding: 0 0 12px`).
- `srinter.properties:2100` — `Validate\ SQL` was inserted after `Parse\ Now` instead of alphabetically (the `V` block is around line 3072, next to `Validate`).

**Regression risk from globally-scoped changes**

These are outside the SQL dialog and affect every consumer; the test plan only exercises one slide-out panel:

- `slide-out.component.scss` — `:host { z-index: 1 }` creates a stacking context around **every** slide-out, so descendants can no longer paint above positioned siblings outside it; `.content-div { overflow: hidden }` clips content in every slide-out-hosted property dialog that previously overflowed. Worth spot-checking a representative set of composer property dialogs.
- `tree.component.scss` — `align-items: center` applies to every tree row in the portal.
- `database-query.component.html:103` — adding `class="query-nav-outlet"` activates ~50 lines of previously-inert CSS in `database-query.component.scss:79-126` (including `padding: 16px` on tab panes and `height: 100%` on all direct children of the active pane). That is a real layout change to the advanced Database Query editor, not just a scroll fix, and it isn't mentioned in the summary.
- `sql-query-dialog.component.scss` — removing `min-height: 500px` from `.modal-body` means that in a plain (non-slide-out) modal the dialog now sizes from `min-height: 350px` floors instead of 500px. Probably fine, just worth confirming visually in both hosts.

**Accessibility**

- Hover-to-scroll (`overflow: hidden` with `&:hover { overflow: auto }`) is used in `.sqp-tree-area` and now also in `.sql-query-dialog-list__surface`. Overflowing content is unreachable by keyboard and on touch devices, and the appearing/disappearing scrollbar causes layout shift on hover. The pattern predates this PR, but it's being spread to a new element here.
- `_#(No filters set) &mdash; click <span>_#(Edit)</span> _#(to add one.)` splits one sentence into three catalog fragments; word order will be wrong in many locales. Prefer a single parameterized string.
- The new expression `<input class="form-control">` in `vpm-condition-editor.component.html:43` has no label, `aria-label`, or placeholder.

**Maintainability / cleanliness**

- `simple-query-pane.component.scss:64,66` — `.sqp-edit-body` declares `min-height: 0` and then `min-height: 350px`; the first is dead.
- `simple-query-pane.component.html:41-128` — a ~90-line `@if (!model.sqlEdited) {` block whose body is not indented, closed with `} <!-- end @if (!model.sqlEdited) -->`. The comment is a symptom; please indent the body or extract the three sections into an `<ng-container>`/child template.
- `vpm-field-editor.component.ts` — `get fieldOptions()` maps `this.fields` into a fresh array on every change-detection tick, so `[options]` changes identity constantly. Compute it in `ngOnChanges` alongside the existing `fields` handling instead.
- `simple-query-pane.component.scss:292` — `font-family: "JetBrains Mono", "Courier New", monospace` is hardcoded and JetBrains Mono isn't bundled; the sibling change in `vpm-condition-editor.component.scss` correctly uses `var(--bs-font-monospace, …)`. Please be consistent.
- The new SCSS mixes design tokens (`--inet-space-*`, `--inet-control-height-md`) with hardcoded pixels (`11px`, `12px`, `14px`, `240px`, `26px`, `10px`). Prefer tokens where they exist.
- `sql-query-dialog.component.scss` — `.modal-footer { background: var(--inet-toolbar-bg-color) }` restyles the footer for this one dialog, which will look inconsistent next to every other modal.
- **Scope.** The VPM condition dialog redesign, the global slide-out/tree fixes, and the SQL dialog redesign are three independent changes in one PR. Splitting them would make the global CSS changes far easier to review and to revert if they regress something.

---

## Summary

Requesting changes for:

1. Blocking issues in the review above — most importantly the `freeFormSqlEnabled` permission gate being dropped for the manual SQL textarea (#1), the invalidated `simple-query-pane.component.risk.tl.spec.ts` test that isn't updated (#2), the unhandled `forkJoin` error path that can leave the Conditions dialog unopenable (#3), the dead empty-drop-zone CSS (#4), and the missing localization keys (#5).
2. A failing check on the PR: `claude-review` (`Reached maximum number of turns (50)`).

No merge conflicts — the branch is mergeable against `epic-74519`.

swaker854:
Recovery-74519-0623 Summary

This PR primary goal is to recover issuses created by Angular update. Instead of just fixing, some fixing is replacing by redesign which is actually easier. Rredesigns the SQL Query dialog's simple mode and the VPM condition dialog, then resolves everything raised in code review across four commits (`3080031f3`, `e3ab64526`, `493437b58`, `17e7a18f9`) — security, error handling, CSS bugs from view-encapsulation mismatches, accessibility, and general cleanup.

### Simple-mode redesign (`3080031f3`, `e3ab64526`)
- Rebuilt `simple-query-pane` with a fixed left data-source rail and a single scrollable right pane (Columns, Joins, Conditions, SQL)
- Manual-edit mode: visual builder hidden while editing raw SQL, textarea fills the panel, clearer confirm dialog, SQL parse status badge
- Fixed a phantom double-scrollbar over the worksheet canvas and advanced-mode Fields/Conditions tab scrolling
- Fixed flex-shrink/min-width so simple mode doesn't overflow or collapse at narrow widths
- Included the VPM condition dialog redesign and recovery of prior-session work

### Code review fixes (`493437b58`)
- **Security**: restored the `FREE_FORM_SQL` permission gate — without it, a user lacking that permission could edit and submit raw SQL after reopening a query someone else saved in manual-SQL mode
- **Error handling**: a single failed table lookup in `editConditions()` no longer permanently blocks the Conditions dialog from opening; added a loading indicator
- **CSS correctness**: found and fixed several rules that silently never applied due to Angular view-encapsulation boundaries (the empty-columns drop zone, the Advanced Query editor's `.query-nav-outlet` rules) — moved styling to an `@Input`-driven approach in one case since this build pipeline doesn't transform bare `::ng-deep`
- **Accessibility**: replaced hover-only scrolling (unreachable by keyboard/touch) with always-on scrolling plus focusable targets; merged a sentence that was split across 3 catalog fragments into one translatable string; labeled an unlabeled expression input
- **Maintainability**: removed dead CSS, a hardcoded/unbundled font-family, inconsistent design-token usage, and a footer style override inconsistent with every other dialog; cut `vpm-field-editor`'s per-tick recomputation down to only-on-change
- **New guidance**: a Grouping-tab hint surfacing an existing (previously silent) server-side validity check, and an empty-state hint on the Links tab's join graph
- **Bug fix**: Advanced Query's Preview tab rendered zero rows — a stale JS-measured inline height was overriding a working CSS fallback, and the dialog was missing a `min-height` floor that Simple mode has (needed because `.modal-content` auto-sizes outside slide-out contexts)

### Final review nits (`17e7a18f9`)
- Applied the same keyboard/touch-accessible scroll fix to the left-rail data-source tree (`.sqp-tree-area`), which used the same hover-to-scroll pattern as an earlier fix elsewhere in the same file
- Removed an orphaned catalog key left over after consolidating a fragmented sentence into one string

### Known non-blocking item
This PR still bundles three logically separate concerns — the SQL dialog redesign, the VPM condition dialog rework, and CSS fixes to shared components (`slide-out.component.scss`, `tree.component.scss`) that affect every consumer of those components, not just this dialog. Flagged during review as worth keeping in mind if any of these need independent reverts later; not a blocker given the fixes already applied.

jshobe-inetsoft (approved):

PR #4404, opened 2026-07-27
Bug #75782, fix database security provider organization cache never loading

ngovus-inetsoft:
## Summary
Switching Enterprise Manager to a `DatabaseAuthenticationProvider` (e.g. H2) in multi-tenant mode could throw `MessageException: "There is no organization in target provider!"` even when the database genuinely has organizations. Three separate issues contributed:

- **Root cause**: `DatabaseAuthenticationCacheServiceImpl.loadInternal()` called Ignite's no-arg `removeAll()` (clear-all-entries) on several distributed maps inside a `cluster.startTx()` transaction. Ignite explicitly rejects the no-arg `removeAll()` inside a transaction (`javax.cache.CacheException: Failed to invoke a non-transactional IgniteCache removeAll operation within a transaction`), so the organization/user/role cache load failed **every time, for any database-backed provider**, independent of configuration. Fixed by using the keyed `removeAll(Set<K>)` overload, which Ignite supports inside a transaction.
- **Silent failure**: the surrounding `try` block in `loadInternal()` had no `catch`, so the exception above (or any other future failure) was discarded by the executor with zero logging, making this defect invisible. Added exception handling + a warning log.
- **Async-load race**: `SecurityTreeServer.getSecurityTree()` read `provider.getOrganizationIDs()` immediately after a provider switch, before the provider's asynchronous cache load had a chance to complete, and treated a still-loading (temporarily empty) cache the same as a genuinely empty one. It now polls briefly for the cache to finish initializing before falling back to the existing empty-org-list check.

## Test plan
- [x] `core` module compiles cleanly with the change
- [x] Manually reproduced: configured a Database (H2) security provider with multi-tenancy enabled; before the fix, cache load failed on every attempt with the Ignite transaction exception (once logging was added) and Security > Users always errored; after the fix, the tree loads successfully with real organization data

### DIFF
--- a/core/src/main/java/inetsoft/sree/security/db/DatabaseAuthenticationCacheServiceImpl.java
+++ b/core/src/main/java/inetsoft/sree/security/db/DatabaseAuthenticationCacheServiceImpl.java
@@ -320,29 +320,34 @@ public class DatabaseAuthenticationCacheServiceImpl implements DatabaseAuthentic
          }
 
          try(DistributedTransaction tx = cluster.startTx()) {
-            lists.removeAll();
+            // Ignite doesn't allow the no-arg (clear-everything) removeAll() inside a
+            // transaction; only the keyed overload is transaction-safe.
             lists.put(ORG_LIST, new TreeSet<>(Arrays.asList(newOrganizations.result())));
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):

PR #4408, opened 2026-07-27
Bug #75777, recycle discarded auto save files of saved sheets

jshobe-inetsoft:
## Summary

The EM **Content > Recycle Bin > Auto Saved Files** node was always empty for sheets that had been saved, even while the composer reported *"Autosaved file exists. Restore?"* for the very same sheet.

## Root Cause

Auto save files live in a per-org blob bucket in two areas: an active area, and a `recycle/`-prefixed area. The EM tree lists only the `recycle/` area.

A file only ever reached `recycle/` for untitled sheets (`TEMPORARY_SCOPE`):

- `VSEventUtil.deleteAutoSavedFile()` renamed the active file into `recycle/` only when the scope was `TEMPORARY_SCOPE`; **for every other scope it deleted the file outright.**
- `AutoSaveUtils.recycleUserAutoSave()` likewise filters to `TEMPORARY_SCOPE`.

So for a saved viewsheet or worksheet, discarding the auto saved content (closing the composer, or opening the sheet and declining the restore) destroyed the draft. It could never show up in the enterprise manager, and there was no way to recover it.

## Fix

**Discarding an auto save file recycles it, saving deletes it.**

- `VSEventUtil.deleteAutoSavedFile()` now moves the file to the recycle bin regardless of scope. This is the discard path: composer close, and open-without-restore. It is also called from `AbstractAssetEngine.getSheet(..., ALL)`, which previously routed non-temporary scopes through an exception-swallowing helper, so the body is wrapped in a `try`/`catch` to keep a storage failure from failing a sheet load.
- A saved sheet supersedes its auto saved content rather than discarding it, so the save paths delete the file:
  - `ComposerViewsheetService.saveViewsheet()` — new call. A plain save of an already-saved viewsheet previously never removed its stale auto save file at all, so a crash right after saving would offer to restore *pre-save* content.
  - `SaveWorksheetService` / `SaveWorksheetDialogService` — switched from `VSEventUtil.deleteAutoSavedFile()` to `AutoSaveUtils.deleteAutoSaveFile()`, preserving today's behavior at these sites.

**Latent bugs in code that only became reachable once non-temporary files can appear in EM:**

- `AutoSaveService.restoreAutoSaveAssets()` classified `1^VIEWSHEET^...` as a **worksheet** via `id.startsWith("8^VIEWSHEET")`; now derived from the entry.
- `AutoSaveUtils.createAssetEntry()` let `AssetEntry.createAssetEntry()` parse the file name's trailing **IP address** field as the entry's **organization id** for any non-temporary scope, resolving the entry against the wrong storage.
- `AutoSaveUtils.getAutoSavedFile()` honored the entry's stored file name only for `TEMPORARY_SCOPE`. An EM restore of any other file would rebuild the name from the *administrator performing the restore* (their user name and IP), find nothing, and silently restore the **saved sheet instead of the draft**.
- `AutoSaveController.deleteAutoSaveAssets()` had the same `8^WORKSHEET^` prefix assumption (audit label only).
- `ContentRepositoryTreeService.addRecycleAutoSaved()`: the "older than a week" purge passed an already `recycle/`-prefixed path to a method that prefixes it again, so it silently deleted nothing; and the node label displayed the raw `Tool.normalizeFileName()` escaping, e.g. `Sales_47_MyVS` instead of `Sales/MyVS`.

## Test Plan

Automated:

- `AutoSaveUtilsTest` — three new cases covering the organization id, the stored file name, and that an entry without a stored name still resolves to the active area (which is what the composer's restore prompt depends on). **Verified that the two regression cases fail against the pre-fix code**, with exactly the predicted failures: `expected: <org1> but was: <0_0_0_0_0_0_0_1~>`, and a name rebuilt as `1^VIEWSHEET^_NULL_~;~host-org^dashboard1^null~`.
- Auto save + repository tree suites: 24 run, 0 failures.
- All `inetsoft.web` tests: 655 run, 0 failures.
- `./mvnw install -pl community/core -am`: BUILD SUCCESS.

Manual (security + multi-tenancy enabled) — verified by the reporter's scenario:

1. Open a **saved** viewsheet in the composer, edit it, wait >25s for the auto save.
2. Kill the browser tab, reopen the viewsheet, answer **No** to *"Autosaved file exists. Restore?"*
3. EM > Content > Recycle Bin > Auto Saved Files lists the file under the user's **Dashboard** folder; restore and delete both work on it.

Regression checks:

- Untitled/temporary sheet behavior is unchanged.
- The composer's *"Autosaved file exists. Restore?"* prompt still works — it checks the active area, which this change does not move files out of prematurely.
- An ordinary edit > **Save** > close leaves nothing in the recycle bin.

Note that checking EM while a design session is still live correctly shows nothing: the draft has not been discarded yet. That is the recycle bin's meaning, and live drafts are recovered through the composer's own prompt.

Fixes Redmine #75777.

### DIFF
--- a/core/src/main/java/inetsoft/analytic/composition/event/VSEventUtil.java
+++ b/core/src/main/java/inetsoft/analytic/composition/event/VSEventUtil.java
@@ -4074,21 +4074,25 @@ public final class VSEventUtil {
    }
 
    /**
-    * Delete auto saved file.
-    * 1 for saved dashboard, keep old logic, delete auto save file when select no.
-    * 2 for untitled dashboard, select no will move auto save file to recycle bin.
+    * Discard the auto saved file of a sheet by moving it to the recycle bin, so that it can
+    * still be recovered from the enterprise manager. This is called when the auto saved content
+    * is thrown away, i.e. when the sheet is closed, when it is opened without restoring the auto
+    * saved content, and when the stored sheet is loaded in place of it
+    * (AbstractAssetEngine.getSheet). When the sheet is saved the auto saved content is superseded
+    * instead of discarded, so the save paths delete the file rather than calling this method.
     */
    public static void deleteAutoSavedFile(AssetEntry entry, Principal user) {
-      if(entry.getScope() != AssetRepository.TEMPORARY_SCOPE) {
-         AutoSaveUtils.deleteAutoSaveFile(entry, user);
-         return;
+      // called while loading a sheet, so never let a storage failure fail the caller
+      try {
+         String savefile = AutoSaveUtils.getAutoSavedFile(entry, user);
+
+         if(AutoSaveUtils.exists(savefile, user)) {
+            String recyclefile = AutoSaveUtils.getAutoSavedFile(entry, user, true);
+            AutoSaveUtils.renameAutoSaveFile(savefile, recyclefile, user);
+         }
       }
-
-      String savefile = AutoSaveUtils.getAutoSavedFile(entry, user);
-
-      if(AutoSaveUtils.exists(savefile, user)) {
-         String recyclefile = AutoSaveUtils.getAutoSavedFile(entry, user, true);
-         AutoSaveUtils.renameAutoSaveFile(savefile, recyclefile, user);
+      catch(Exception e) {
+         LOG.debug("Failed to move the auto save file to the recycle bin: {}", entry, e);
       }
    }
 
--- a/core/src/main/java/inetsoft/analytic/composition/event/VSEventUtil.java
+++ b/core/src/main/java/inetsoft/analytic/composition/event/VSEventUtil.java
@@ -4080,8 +4080,18 @@ public final class VSEventUtil {
     * saved content, and when the stored sheet is loaded in place of it
     * (AbstractAssetEngine.getSheet). When the sheet is saved the auto saved content is superseded
     * instead of discarded, so the save paths delete the file rather than calling this method.
+    * An entry that already refers to a file in the recycle bin is ignored.
     */
    public static void deleteAutoSavedFile(AssetEntry entry, Principal user) {
+      // an entry that carries the name of an existing auto save file, i.e. one created by
+      // AutoSaveUtils.createAssetEntry() for the recycle bin, already refers to a discarded file,
+      // so there is nothing to discard. the name embeds the user and ip address of the session
+      // that created it and can not be recreated from the current user, so moving it again would
+      // only orphan it under a name attributed to the wrong user.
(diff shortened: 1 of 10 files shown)


jshobe-inetsoft:
Thanks — the correctness issue is real and is now fixed in 24c42a5.

**Restore orphaning the recycle bin entry.** Confirmed the whole path, including that it is newly reachable: before this PR the `scope != TEMPORARY_SCOPE` branch called `AutoSaveUtils.deleteAutoSaveFile(entry, user)`, which used the 3-arg builder with `recycle=false` and simply missed, harmlessly. It is also why the 2-arg change is required in the first place — `getAutoSavedSheet()` reads the draft through `AutoSaveUtils.getInputStream(entry, user)`, i.e. the same 2-arg overload.

I took the second of the two suggested options. Making the 3-arg overload honor `autoFileName` would silently override its explicit `recycle` argument and change behavior for its other callers (`deleteAutoSaveFile(entry, user)`, `writeAutoSaveFile`), so `deleteAutoSavedFile()` now returns early when the entry carries a file name:

```java
if(entry.getProperty("autoFileName") != null) {
   return;
}
```

`autoFileName` is set in exactly one place, `AutoSaveUtils.createAssetEntry()`, which always pairs it with `isRecycle=true`, so an entry carrying it is always already a recycle bin file with nothing left to discard. Of the five callers of `deleteAutoSavedFile`, only `AbstractAssetEngine` can receive such an entry — the other four build their entry from a live runtime sheet or `AssetEntry.createAssetEntry(id)` — so the short-circuit is a no-op for them. With the rename gone, the explicit cleanup in `AutoSaveController.restoreAutoSaveAssets()` finds the file and deletes it as intended.

**Test coverage.** Added `VSEventUtilAutoSaveTest` with three cases pinning the branches of `deleteAutoSavedFile`: a recycled entry is not moved again, an active entry is still moved to the recycle bin, and a missing file is not renamed. The first case fails without the fix.

**Minor observations** — no changes, agreeing with your own non-blocking assessment:

- `denormalizeAssetName()` `_NN_` ambiguity: correct, and unfixable without changing the on-disk name format, since `Tool.normalizeFileName()`'s escaping is not self-delimiting. The impact is a display-only recycle bin label, so that is not worth doing in a bug fix.
- `path.contains("^WORKSHEET^")`: the value only feeds an audit label, and a false positive would need an asset path literally containing `^WORKSHEET^` — but `^` is itself escaped by `Tool.normalizeFileName()` when the path is written into the name field, so it cannot occur.

Full `core` suite: 4209 tests, 0 failures.

ngovus-inetsoft (approved):

PR #4409, opened 2026-07-27
75641, In multi-style chart with bar and line, the bars should dim

milotalon:
Root Cause:
A multi-style chart binds a chart type per measure, so one SVG holds both bar annotation groups (.inetsoft-bar, from BarVO) and line series groups (.inetsoft-line, from LineVO).

In chart-inline-svg.directive.ts, afterSvgInjected() treated the presence of any .inetsoft-line group as proof of a pure line chart and cleared elementGroupMap - the map that resolves a hovered row/col to its SVG group. That is correct for a pure line chart (its point markers are driven by the cursor-resolved series hover instead), but on a multi-style chart it also discarded the bar entries. The canvas-driven highlightElement(row, col) call then found nothing, inetsoft-active was never set, and the server-injected rule

svg:has(.inetsoft-bar.inetsoft-active) .inetsoft-bar:not(.inetsoft-active)
  { opacity: .2 !important }
never fired. Hovering a bar highlighted nothing and faded nothing.

Fix:
Both hover mechanisms now stay live when one SVG contains both (tracked as "mixedHover"):

- elementGroupMap is rebuilt excluding only .inetsoft-point instead of being cleared, so bars remain hoverable. A pure line chart has nothing else in the map, so it still ends up empty and its behavior is unchanged.
- The two mechanisms dim each other, so the hovered VO is the only bright thing in the plot: a bar hover fades the line series (inline opacity, since the series carry no inetsoft-active class), and a line-series hover - resolved from the cursor or from the snap guideline - fades the bars and their value labels.
- The line proximity hover stands down while a bar owns the highlight, so a line passing within the 50px "nearest series" radius cannot steal it from the bar under the cursor.
- The stacked-column key trim in highlightElements is skipped for such tiles, so every segment of a stacked bar column stays lit.

All of the new inline dimming is confined to single-tile mode; split (cross-tile) charts still dim through the parent component as before.

No backend change was required - appendHoverCSS already emits every chart type's hover rules regardless of the animation hint.

Covered by 7 new cases in chart-inline-svg.directive.spec.ts.

### DIFF
--- a/web/projects/portal/src/app/graph/objects/chart-inline-svg.directive.spec.ts
+++ b/web/projects/portal/src/app/graph/objects/chart-inline-svg.directive.spec.ts
@@ -337,6 +337,140 @@ describe("ChartInlineSvgDirective cross-tile dim", () => {
       });
    });
 
+   describe("multi-style chart (bars + line in one SVG)", () => {
+      // A multi-style chart binds a chart type per measure, so one SVG holds bar VOs and a line
+      // series. Both hover mechanisms must stay live: bars highlight by inetsoft-active (server
+      // :has() dim), the line resolves from the cursor and dims by inline opacity.
+      const html = `
+         <svg>
+            <g class="inetsoft-bar" data-row="0" data-col="0"></g>
+            <g class="inetsoft-bar" data-row="1" data-col="0"></g>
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4411, opened 2026-07-27
Bug #75778, keep bottom-tabs tables clear of the tab strip in Excel export

jshobe-inetsoft:
## Summary

When a table lives in a tab container with **Tabs on Bottom** enabled, exporting the viewsheet to Excel with *Expand Components* drew the tab strip on top of the table's last row(s).

## Root Cause

Two coordinate systems disagree in the Excel exporter (1 Excel row = `AssetUtil.defh` = 20 px):

- Table cells are snapped to the sheet's row grid — `ExcelTableHelper.writeTitleCell` / `ExcelCrosstabHelper` do `startY = PoiExcelVSUtil.ceilY(startY)` and each table row consumes a whole grid row. A table whose `pixelOffset.y` is not a multiple of 20 is therefore drawn up to 19 px lower than its pixel position, while its drawn height matches its assembly height.
- The tab strip is written as a picture at its exact pixel position (`PoiExcelVSExporter.writeVSTab` → `getAnchorPosition`), and in bottom-tabs mode `TabVSAssembly.updateChildPosition` pins the strip to the child's pixel bottom (`child.y + child.height`).

Because the strip sits exactly at the child's pixel bottom, there is no slack to absorb the grid rounding, so the table's last row always lands underneath the strip whenever the child's top is off-grid. Measured on the reported asset: table drawn 240–380 px, strip at 364 px.

This is not specific to shrink-to-fit — shrink only determines where the table lands. It reproduced with `shrink=false` and with a plain expanded 700-row table as well.

## Fix

`PoiExcelVSExporter.alignBottomTabsTables()` aligns the top of tables inside a bottom-tabs container down to the grid line (`floorY`) after `super.prepareSheet()` (which is where the shrink shift from #74933 is applied) and before the sheet's row/column grid is built. `ceilY` then becomes a no-op, so the drawn bottom stays above the strip.

Scoped to the Excel exporter only — PDF/PNG/HTML place both the table and the strip in exact pixels and are unaffected. Gated on `TabVSAssemblyInfo.isInBottomTabs`, so top-tab and non-tab tables produce byte-identical output. The offset is set on the child's `VSAssemblyInfo` (mutating the strip instead would re-anchor the child right back via `updateChildPosition`), and the mutation happens on the throwaway viewsheet clone created by `AbstractVSExporter.export()`.

## Test Plan

Verified against a running server by exporting the reporter's asset and measuring the raw XLSX geometry (cell rows vs. the picture's `xdr:from`/`xdr:to` anchors):

| asset variant | before | after |
| --- | --- | --- |
| shrink=true, 5 rows (as reported) | table 240–380, strip 364 → **overlap 16 px** | table 220–360, strip 364 → gap 4 px |
| shrink=false, 5 rows | table 80–380, strip 364 → **overlap 16 px** | table 60–360, strip 364 → gap 4 px |
| shrink=true, 700 rows (expanded) | table 80–14120, strip 14105 → **overlap 15 px** | table 60–14100, strip 14105 → gap 5 px |
| shrink=false, 700 rows (expanded) | table 80–14120, strip 14105 → **overlap 15 px** | table 60–14100, strip 14105 → gap 5 px |

Also verified: `TabVSAssemblyTest` passes (4/4); top-tab and non-tab Excel exports unchanged.

To reproduce manually: import the `tab.zip` asset from the issue, open it in the portal, export to Excel with *Expand Components*, and confirm the tab labels no longer sit on top of the table's last row.

Fixes Redmine #75778.

### DIFF
--- a/utils/inetsoft-xml-formats/src/main/java/inetsoft/report/io/viewsheet/excel/PoiExcelVSExporter.java
+++ b/utils/inetsoft-xml-formats/src/main/java/inetsoft/report/io/viewsheet/excel/PoiExcelVSExporter.java
@@ -223,19 +223,20 @@ public class PoiExcelVSExporter extends ExcelVSExporter {
     * Snapping the table down to the grid line makes {@code ceilY} a no-op so the drawn
     * bottom stays above the strip.</p>
     */
-   private void alignBottomTabsTables() {
+   // package-private for testing
+   void alignBottomTabsTables() {
       if(viewsheet == null) {
          return;
       }
 
       for(Assembly assembly : viewsheet.getAssemblies(true)) {
-         if(!(assembly instanceof TableDataVSAssembly) || !needExport((VSAssembly) assembly) ||
-            !TabVSAssemblyInfo.isInBottomTabs((VSAssembly) assembly))
+         if(!(assembly instanceof TableDataVSAssembly table) || !needExport(table) ||
+            !TabVSAssemblyInfo.isInBottomTabs(table))
          {
             continue;
          }
 
-         VSAssemblyInfo info = ((VSAssembly) assembly).getVSAssemblyInfo();
+         VSAssemblyInfo info = table.getVSAssemblyInfo();
          Point offset = info.getPixelOffset();
 
          if(offset == null) {
--- a/utils/inetsoft-xml-formats/src/main/java/inetsoft/report/io/viewsheet/excel/PoiExcelVSExporter.java
+++ b/utils/inetsoft-xml-formats/src/main/java/inetsoft/report/io/viewsheet/excel/PoiExcelVSExporter.java
@@ -163,6 +163,7 @@ public class PoiExcelVSExporter extends ExcelVSExporter {
       PoiExcelVSUtil.processOverlap(vsheet);
 
       super.prepareSheet(vsheet, sheetName, box);
+      alignBottomTabsTables();
       setUpSheet(sheetName);
 
       if(sheet == null) {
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft:
Thanks — I traced all three items. Summary: **#3 is fixed in 5e47344**; #1 and #2 are not reachable/actionable, with details below.

### 1. `layoutPosition` unit mismatch — no change

The API description is right (`VSAssemblyInfo.java:778-790`, `:770-773`), but `scaledPosition` is always null by the time this runs, so `getLayoutPosition()` and `getLayoutPosition(false)` return the same raw, unscaled `layoutPosition` — the same coordinate space as `pixelOffset`:

- Both export entry points call `VSEventUtil.clearScale()` on the export clone before `exporter.export()`: `VSExportService.java:757` (current view) and `:826` (bookmarks), plus `ExportControllerService.java:217/279`. `clearScale()` (`VSEventUtil.java:348-364`) nulls `scaledPosition`/`scaledSize` and recurses into every child — explicitly *"don't use the scale-to-screen size for export"*.
- The layout path can't repopulate it: `AbstractLayout.apply()` calls `clearLayoutState()` then `setLayoutPosition()` (`AbstractLayout.java:165/244`, `PrintLayout.java:128`), and `setLayoutPosition()` itself nulls `scaledPosition`. `applyScale()` is only reached from runtime refresh (`CoreLifecycleService:596`, `RuntimeViewsheet:307`), never during export.
- `refreshAllAnnotations()` at `AbstractVSExporter.java:1155` does set `scaledPosition`, but only on annotation/line/rectangle infos — never on the `TableDataVSAssemblyInfo` this loop touches.

Also worth noting for consistency: the paired shrink shift from #74933, `VSTableDataHelper.applyShrunkBottomTabsShift()` (`:885-895`), uses the identical `getPixelOffset()`/`getLayoutPosition()`/`setLayoutPosition()` pattern. Switching to `getLayoutPosition(false)` here would make the two adjustments disagree, and wouldn't help anyway since `setLayoutPosition()` clears `scaledPosition` either way.

### 2. Overlap above the table — no change

The specific scenario named doesn't exist: a bottom-tabs child is positioned by `TabVSAssembly.updateChildPosition()` at `y = Math.max(0, bounds.y - childHeight)` (`TabVSAssembly.java:255-266`) — it is anchored *upward from the strip*. The tab strip is the child's only structural neighbour and it is below it; there is no container title bar above.

The residual is real but inherent to the 20px row grid and symmetric: the drawn top is quantized either way — it was `ceilY(y)` (1–19px *below* the pixel top, i.e. the reported bug) and is now `floorY(y)` (1–19px above). An assembly abutting the child's top flush would trade a 1–19px gap for a 1–19px overlap, but the grid cannot represent sub-row adjacency in either direction, and `PoiExcelVSUtil.processOverlap()` already runs before `prepareSheet()`. Flooring is the only direction that clears the strip.

### 3. No automated test — fixed in 5e47344

Fair point. Added `PoiExcelVSExporterTest` (9 tests, the first in `utils/inetsoft-xml-formats`; the module already declared the junit/mockito test deps): off-grid snap with the matching `layoutPosition` shift, floor-not-round, no-op when already grid-aligned, and the top-tabs / not-in-a-tab / non-table / `needExport`-excluded / null-offset / null-viewsheet skips — every branch and guard in the method. The infos are mocked because `VSAssemblyInfo`'s constructor reads `SreeEnv` properties, which this module has no equivalent of core's `SreeHomeExtension` to bootstrap.

Mutation-checked rather than just run green: flipping `floorY` → `ceilY` in the production code fails exactly the two tests that pin the flooring behavior.

### Style nit — applied

`instanceof TableDataVSAssembly table` pattern variable, dropping the three casts. Also relaxed `alignBottomTabsTables()` to package-private so the test can call it without reflection — it still has a single production caller, and the double `super.prepareSheet()` in `OfflineExcelVSExporter` was already calling it twice before this change (idempotent, since `floorY` is).

ngovus-inetsoft (approved):

PR #4418, opened 2026-07-27
Bug #75752, fix physical model graph moving unrelated table on drag

ngovus-inetsoft:
## Summary
- Dragging one table node in the physical model designer's graph view could also move (and select) an unrelated table when both were auto-aliases of the same source table (e.g. "Customer Region" and "Salesperson Region", both aliases of `REGIONS`).
- `selectNode()` short-circuited on a plain click whenever the clicked node was already part of the current `dragNodes` selection, without checking modifier keys, so a stale multi-node selection was never narrowed down to just the clicked node before a drag.
- Selecting a node also fed its tree path back up to the parent component, which re-derived the graph selection by matching *every* graph node sharing that tree path — since auto-alias tables share their source table's tree path, this re-widened the selection to include unrelated aliases.

## Fix
- `selectNode()` now always collapses a plain (non-ctrl/shift) click down to just the clicked node, even if it was previously part of a stale multi-selection.
- Added a `suppressNextSelectionSync` flag so a selection change originating in the graph itself (click, ctrl/shift-click, rubber-band select) is not immediately overwritten by the parent's echoed-back `selectedGraphModels`, which could resolve to multiple aliases of the same source table.

## Test plan
- [x] Manually reproduced in browser: opened a physical view with two auto-alias tables of the same source table, dragged one — confirmed only the dragged table moves/highlights, the other is untouched.


jshobe-inetsoft (approved):

PR #4422, opened 2026-07-28
fix(wiz): prevMirror's own AggregateInfo survives a dashboard reload

larryliang-inetsoft:
## Summary

A wiz dashboard's shared `prevMirror` correctly carried the first chart's own `AggregateInfo` through the entire compose flow, but silently lost its aggregates (keeping only the group) the moment the saved dashboard was reopened fresh — e.g. by `verify_visualization` — crashing that chart's own graph render with `Aggregate not found: <name>`.

**Root cause:** StyleBI's generic query-construction entry point (`AssetQuery.createAssetQuery`, invoked for every table on every viewsheet open via `AssetQuerySandbox#refreshColumnSelection`, entirely independent of the wiz merge code) resolves a mirror's aggregate columns down to their base table using outer-attribute qualification (`"<base>.<name>"`), then revalidates the mirror's `AggregateInfo` against that newly-qualified selection. `prevMirror`'s own aggregate refs still pointed at the original bare names, so that revalidation found no match and dropped them — while the group ref survived because groups are never routed through that same base-resolution step.

**Fix:** qualify `prevMirror`'s own `AggregateRef`s (not `GroupRef`s) with the same outer-attribute naming up front in `ensureBaseHasPrevMirror`, and keep `prevMirror`'s own column selection in sync, so StyleBI's independent re-derivation always finds a match.

Also included (found along the way, same class of bug):
- `Viewsheet#createMirrorTable` is now idempotent for any pre-existing mirror at a bare table name (not just its own naming convention) — it was re-wrapping a wiz dashboard's `prevMirror` on every subsequent chart merge, orphaning the first chart's aggregation.
- `WorksheetTableService` now sets `isAggregate(true)` alongside a non-empty `AggregateInfo` at original table construction time, fixing the same `isAggregate()`/`AggregateInfo` inconsistency at its source.

## Test plan

- [x] `WsMergeServiceTest` — 6/6 passing (new regression test: `prevMirrorAggregateRefsAreOuterAttributeQualifiedToMatchTheQueryEnginesOwnResolution`)
- [x] Full `inetsoft.web.wiz.**` + `ChartVSAQueryTest` suites — 539/539 passing, no regressions
- [x] Live verification: rebuilt the local StyleBI image, reproduced the crash on the real dashboard, applied the fix, confirmed `verify_visualization` now returns `ok:true` for the exact chart that was crashing, and for a sibling chart sharing the same physical table via `condMirror`
- [x] Confirmed clean logs (no `Aggregate not found` / `Column not found` for the fixed chart) after a full container restart (cold reproduction, ruling out any JVM-cache artifact)

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/Viewsheet.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/Viewsheet.java
@@ -5300,11 +5300,24 @@ public class Viewsheet extends AbstractSheet implements VSAssembly, VariableProv
          return false;
       }
 
-      // break dependency cycle
+      // Already a mirror under this bare name -- the goal of this method (ensure the bare
+      // name a VS chart binds to is a mirror, not a directly-shared raw table) is already
+      // satisfied, regardless of what created it or what it wraps. The narrower check this
+      // replaced (matching only this method's own "{name}+OUTER_TABLE_SUFFIX" naming
+      // convention) misses mirrors created by OTHER code — e.g. a wiz dashboard's own
+      // prevMirror (WsMergeService#ensureBaseHasPrevMirror), which uses a "{name}_base"
+      // convention for what it wraps. Without this broader check, re-running this method
+      // (createMirrorTables runs on every repopulateWorksheet, i.e. once per chart merged
+      // into a wiz dashboard, and rescans EVERY VS assembly's table name every time, not
+      // just the newly-added one) renames that prevMirror out of the way and wraps it in a
+      // brand-new, empty-AggregateInfo mirror under the bare name every single time a later
+      // chart is merged — repeatedly re-wrapping deeper each time and eventually leaving the
+      // first chart's own VS binding pointing at emptied-out columns instead of the
+      // aggregation prevMirror was set up to carry. Confirmed live: a wiz dashboard's first
+      // chart crashed rendering with "Aggregate not found: <name>" only once enough
+      // additional charts sharing its physical table had been merged in afterward.
       if(table instanceof MirrorTableAssembly) {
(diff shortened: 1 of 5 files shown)


larryliang-inetsoft:
Addressed the review feedback:

1. **Blast radius / test coverage for the global `Viewsheet.java` change** — added `ViewsheetCreateMirrorTableTest` (package `inetsoft.uql.viewsheet`, no wiz dependency), which builds a worksheet where a bare-named table is already wrapped by an unrelated pre-existing mirror and asserts `createMirrorTable` doesn't re-wrap it. Confirmed this test fails against the pre-fix code (re-wraps and renames the mirror out to `T_O`).

2. **Missing 3rd-chart / outer-mirror-adaptation test coverage** — added `aThirdChartSharingTheSamePhysicalTableReusesTheExistingPrevMirrorUnchanged` (exercises the `wizMirror != null` early-return path) and `adaptsAPreExistingNonWizMirrorInPlaceInsteadOfCreatingAnUnreachableOne` (exercises the `outerMirror != null` in-place-adaptation path) to `WsMergeServiceTest`.

3. **Silent no-op in `replaceColumnByName`** — it now adds the qualified column and logs a warning when the target name isn't found in the selection, instead of silently leaving the (now-qualified) `AggregateRef` pointing at a column absent from the selection.

4. **Comment volume** — trimmed the "confirmed live via targeted diagnostic logging" chronology out of the largest comment blocks in `WsMergeService.java`, keeping the invariant/reasoning.

All 542 tests (full `inetsoft.web.wiz.**` + `ChartVSAQueryTest` + the 3 new tests) pass. Re-verified live on the real dashboard after these changes — the original crash stays fixed.

PR #4427, opened 2026-07-28
Bug #75795, honor the Roles root grant for org-scoped role permission checks

ngovus-inetsoft:
## Summary
- Fixes a `SecurityException` thrown when an EM user granted Administrator (ADMIN) permission only on the "Roles" root node tries to view/select an org-scoped role (e.g. a custom role in a non-default org while multi-tenancy is disabled).
- `DefaultCheckPermissionStrategy.checkPermission()` only consulted the `"Organization Roles"` permission root for org-scoped roles, never the `"Roles"` root, even though `"Roles"` is the node actually exposed for granting in non-multi-tenant mode (see `UserTreeService`). Now both roots are checked (OR), matching the pattern already used elsewhere for `SECURITY_ROLE` (e.g. `checkOrgAdminPermission`).

## Test plan
- [x] `./mvnw -pl core -am compile -DskipTests` passes
- [x] `./mvnw -pl core test -Dtest=DefaultCheckPermissionStrategyTest,RoleControllerTest` passes
- [ ] Manual repro per the reported steps: grant a non-admin EM user ADMIN on the "Roles" node only, confirm they can now open an org-scoped role without the SecurityException

### DIFF
--- a/core/src/main/java/inetsoft/sree/security/DefaultCheckPermissionStrategy.java
+++ b/core/src/main/java/inetsoft/sree/security/DefaultCheckPermissionStrategy.java
@@ -136,17 +136,29 @@ public class DefaultCheckPermissionStrategy implements CheckPermissionStrategy {
          String rootRole = Organization.getRootRoleName(principal);
          String rootOrgRole = Organization.getRootOrgRoleName(principal);
          Role role = provider.getRole(IdentityID.getIdentityIDFromKey(resource));
-         Permission rootRolePer = provider.getPermission(type, rootRole);
-         Permission rootOrgRolePer = provider.getPermission(type, rootOrgRole);
+         Permission orgRoleRootPer;
 
-         // "Roles" and "Organization Roles" are two independent grantable nodes in the EM
-         // security tree (see UserTreeService), so a grant on either root must be honored
-         // here — checking only one caused admins granted on the "Roles" root to be denied
-         // access to org-scoped roles, since only "Organization Roles" was consulted before.
-         if((role == null ||
+         if(role != null && role.getOrganizationID() != null) {
+            orgRoleRootPer = provider.getPermission(type, rootOrgRole);
+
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (changes requested):
Requesting changes — the `build` check is currently failing on the latest commit (b8adc003), and it looks like a real regression introduced by this change rather than flakiness:

```
[ERROR] PermissionMatrixResourcesS2Test.rootGlobalRoleAdmin_adminOnOrgRole_denied_rootsShouldBeIndependent -- Time elapsed: 0.006 s <<< FAILURE!
org.opentest4j.AssertionFailedError: Expected DENY but got ALLOW: [rootGlobalRoleAdmin~;~matrix_org_id] ADMIN on SECURITY_ROLE/targetRole ==> expected: <false> but was: <true>
```

This test asserts that the `"Roles"` root grant and the `"Organization Roles"` root grant are independent (a global-role-root admin should NOT get admin on an org-scoped role). That's exactly the invariant this PR's OR'd-roots change in `DefaultCheckPermissionStrategy.checkPermission()` appears to break — the automated Claude review flagged this as a coverage gap but wasn't able to run the test suite to catch the actual failure.

Please reconcile the fix with `PermissionMatrixResourcesS2Test` — either the new OR logic needs a scope-aware guard so it doesn't leak a global "Roles" root grant onto org-scoped roles, or the test's expectation needs to be intentionally updated with justification if the old assumption no longer holds.

jshobe-inetsoft (approved):
Claude's follow-up review (after commit 49480258) confirms the PermissionMatrixResourcesS2Test regression I flagged is resolved: the fallback to the "Roles" root is now scoped to only fire when "Organization Roles" has never been configured, preserving root independence. All checks are green (build, claude-review, Semgrep, CLA) and no blocking issues remain. Approving.

PR #4428, opened 2026-07-28
admin-chat Plan 1 (community): AdminChangeRecord + property-change service/endpoint

jshobe-inetsoft:
## What & why

Community-side foundation for **stylebi-admin-chat** (Plan 1 of 3) — the server-side capability that lets an external broker change StyleBI server properties **safely, reviewably, reversibly, and with a full audit trail** via natural language. This PR contains the community-module pieces; the enterprise-module pieces (audit persistence override + changeset read-back endpoints) are in a companion PR against `stylebi-enterprise` that bumps the submodule pointer to this branch.

Design & plan: `docs/superpowers/specs/2026-07-08-stylebi-admin-chat-plugin-design.md` and `docs/superpowers/plans/2026-07-08-admin-chat-plan1-stylebi-foundation.md` (in the stylebi-wiz repo).

## Contents (this repo)

- **`AdminChangeRecord`** (`inetsoft.util.audit`) — new task-aware audit record type: `transactionId`, `taskDescription`, `property`, `beforeValue`/`afterValue`, `action`, `status`, `riskLevel`, `snapshotScope`, `backupRef`, etc. All persisted getters carry `@AuditRecordProperty` so `AuditRecordMapper` maps them.
- **`Audit.auditAdminChange(...)`** — new no-op base method (enterprise `DefaultAudit` overrides it to persist).
- **`AdminChangeService`** (`inetsoft.web.admin.ai`) — snapshot → apply (`SreeEnv`) → read-back verify → **always** audit (record written in a `finally`, even on failure). Fail-loud on blank `transactionId`/`property`.
- **`AdminAiController`** — `POST /api/admin/ai/change`, `/backup`, `/restore`, all `@Secured` on `EM_COMPONENT settings/properties`.
- **`AdminBackupService`** — Tier-2 storage backup/restore wrapping the live `StorageService` bean; path-traversal-safe (rejects `/`, `\`, `..`, + canonical-parent containment check).

## Testing

TDD throughout. Admin suite green: `AdminChangeRecordTest` (4), `AdminChangeServiceTest` (8), `AdminAiControllerTest` (4), `AdminBackupServiceTest` (7) — including a real write→backup→mutate→restore→read round-trip and path-traversal rejection tests.

Run: `./mvnw -pl community/core -am test -Dtest='AdminChangeRecordTest,AdminChangeServiceTest,AdminAiControllerTest,AdminBackupServiceTest' -Dsurefire.failIfNoSpecifiedTests=false`

## Notes for reviewers

- Deferred (for Plan 2 / follow-up): backup retention policy; recording *who* triggered a backup/restore (Principal currently unused on those endpoints); `listChangesets` MAX_SCAN truncation caveat.
- Scope is deliberately the properties domain only; broader EM domains are future plans.

### DIFF
--- a/core/src/main/java/inetsoft/util/audit/AdminChangeRecord.java
+++ b/core/src/main/java/inetsoft/util/audit/AdminChangeRecord.java
@@ -1,0 +1,116 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.util.audit;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import org.apache.commons.lang3.StringUtils;
+
+import java.sql.Timestamp;
+
+public class AdminChangeRecord implements AuditRecord {
+   public static final String ACTION_APPLY = "apply";
+   public static final String ACTION_ROLLBACK = "rollback";
+   public static final String ACTION_RESTORE = "restore";
+   public static final String STATUS_VERIFIED = "verified";
+   public static final String STATUS_FAILED = "failed";
+   public static final String RISK_LOW = "low";
+   public static final String RISK_HIGH = "high";
+   public static final String SCOPE_VALUE = "value";
+   public static final String SCOPE_STORAGE = "storage";
+
+   public AdminChangeRecord() {
+      super();
+   }
+
+   @JsonIgnore
+   @Override
+   public boolean isValid() {
+      try {
+         validate();
+         return true;
+      }
+      catch(IllegalStateException ignore) {
+         return false;
+      }
+   }
+
+   private void validate() {
+      if(StringUtils.isEmpty(transactionId)) {
+         throw new IllegalStateException("Invalid admin change record, transactionId cannot be null");
+      }
+      if(StringUtils.isEmpty(property)) {
+         throw new IllegalStateException("Invalid admin change record, property cannot be null");
+      }
+      if(StringUtils.isEmpty(action)) {
+         throw new IllegalStateException("Invalid admin change record, action cannot be null");
+      }
+   }
+
+   public String getTransactionId() { return transactionId; }
+   public void setTransactionId(String v) { this.transactionId = v; }
+   public String getTaskDescription() { return taskDescription; }
+   public void setTaskDescription(String v) { this.taskDescription = v; }
+   public String getProperty() { return property; }
+   public void setProperty(String v) { this.property = v; }
+   public String getObjectType() { return objectType; }
+   public void setObjectType(String v) { this.objectType = v; }
+   public String getBeforeValue() { return beforeValue; }
+   public void setBeforeValue(String v) { this.beforeValue = v; }
+   public String getAfterValue() { return afterValue; }
+   public void setAfterValue(String v) { this.afterValue = v; }
+   public String getAction() { return action; }
+   public void setAction(String v) { this.action = v; }
+   public String getStatus() { return status; }
+   public void setStatus(String v) { this.status = v; }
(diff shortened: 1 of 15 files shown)


jshobe-inetsoft:
## Response to review round 1

Verified each finding against the current branch rather than assuming the later commits covered them. Six of the seven are done; one is a deliberate scope decision.

| # | Finding | Disposition | Verified at |
|---|---|---|---|
| 1 | `action` not validated → silent audit loss | Fixed in `efa3c338` | `AdminChangeService.requireValidAction` (lines 93-104) rejects blank *and* unknown actions before any `SreeEnv` mutation or audit write |
| 2 | `PropertiesController` side-effects not replicated | Fixed in `ff5d7ec9` + `7e70b431` | `PropertyChangeSideEffects` is now shared by both callers; the pre/post-remove split preserves the original ordering around `SreeEnv.remove()`/`save()` |
| 3 | Inconsistent error contract across the three endpoints | Fixed in `aacbea5d` | `AdminAiController` lines 64-69: `@ExceptionHandler(IllegalArgumentException.class)` → 400 for all three; `restore()` no longer swallows failures into a 200 body |
| 4 | Empty-string value semantics diverge from `PropertiesController` | Fixed (comment added) | `AdminChangeService` lines 21-23 document the deliberate divergence: `""` is an explicit set-to-empty, reset-to-default is expressed as `null` |
| 5 | No allow-list on which properties can be changed | **Deferred to Plan 2/3** | See below |
| nit | `reviewOutcome`/`userSessionID`/`organizationId` annotated but unpopulated | Fixed (comment added) | `AdminChangeService` lines 121-123, including the note that `organizationId` *is* populated downstream by `DefaultAudit` |
| nit | `AdminBackupService` spike-findings narrative | Fixed in `73c479a0` | Trimmed to the durable rationale (lines 28-40) |

**On #5 (property allow-list)** — leaving this out of Plan 1 deliberately. `PropertiesController.editProperty` has no allow-list either, so adding one here would be new capability rather than a fix, and it needs a product decision about which properties are broker-eligible (the `security.*` question in particular). The threat-model concern is partly mitigated in the meantime by the `requireSiteAdmin` gate added in `aacbea5d`, which restricts all three endpoints to the Site Administrator role rather than just `EM_COMPONENT settings/properties` access. Tracked for Plan 2/3.

**On the request to confirm CI** — all checks are green on this branch (`build`, `claude-review`, both Semgrep jobs, CLA). Locally: `./mvnw test -pl core` → 4960 tests, 0 failures, 0 errors, 66 skipped.

jshobe-inetsoft:
## Response to re-review — `removeLogLevel` null-deref

Confirmed and fixed. I verified all three parts of the claim independently before changing anything:

1. **Reachable** — `PropertyChangeSideEffects.java:91-92`. The guard short-circuits on empty / non-`log.` / non-`.level.` first, so reaching `value.equals("off")` means the property *is* shaped `log.*.level.*`; if it was never set, `value` is `null` → NPE.
2. **Not a regression** — diffed against base `stylebi-wiz-main-rebase`: byte-identical to the pre-refactor `PropertiesController.removeLogLevel` (lines 179-186 there).
3. **Untested** — the one existing null-value test (`applyPreRemoveSideEffects_exposeDefaultOrgProperty_doesNotFireRepositoryEvent`) uses `security.exposedefaultorgtoall`, which returns early at the `startsWith("log.")` check and never reaches the deref.

Also confirmed the stated consequence: in `AdminChangeService.applyChange` the NPE is caught by the outer `catch(Exception)`, so the caller gets `status: "failed"` rather than a crash — but the delete silently never happens.

**Fix applied:**
- `"off".equals(value)` instead of `value.equals("off")`, with a comment recording *why* (the admin-chat path can name a never-configured log-level property directly; the EM UI dropdown only offers ones that already have a value).
- New regression test `applyPreRemoveSideEffects_logLevelPropertyNeverSet_doesNotThrow`. I checked it is a real regression test: with the production file reverted and only the test applied, it fails with exactly `NullPointerException: Cannot invoke "String.equals(Object)" because "value" is null`. Falls through to `getContextLevels()`, which returns empty for a never-configured property, so `found` is `false` and the method is correctly a no-op.

**One adjacent issue deliberately left alone:** `SreeEnv.getProperty(property)` is read on the line *before* the `Tool.isEmptyString(property)` guard, so a `null` property would fail inside `getProperty` before the guard meant to catch it. Also pre-existing, and currently unreachable — `AdminChangeService.applyChange` calls `requireNonBlank("property", ...)` before entering the `try`, and `PropertiesController.deleteProperty` binds `property` as a `required = true` `@RequestParam`, so Spring rejects a missing value with a 400 before the method body runs (it can be `""` but never `null`). Those are the only two production call sites. Keeping this change to the flagged defect only; worth revisiting if either call site's validation changes.

**Verification:** `./mvnw test -pl core -Dtest=PropertyChangeSideEffectsTest` → 12/12. Full `./mvnw test -pl core` → 4960 tests, 0 failures, 0 errors, 66 skipped. `./mvnw clean install -pl core -am -DskipTests` → BUILD SUCCESS.

jkidacki-inetsoft (approved):

PR #4432, opened 2026-07-28
Bug #75646 - Fix secondary y-axis labels not rendering in dual-axis charts

FigurePaper:
When a measure is bound to the secondary axis, DefaultGraphGenerator installs a
separate yscale2 on the RectCoord, but createAxis() decided whether to draw the
right-side axis by bit-testing the primary y scale's style against AXIS_SINGLE2
(0x2|0x10). Changing the AxisSpec default style from AXIS_DOUBLE (0x3) to
AXIS_SINGLE (0x1) in 247f34f8 made that test fail, so the secondary axis was
given zIndex -1 and then dropped for having no grid lines. The axis title is an
EGraph y2 title and painted independently, which is why the title remained
visible with no line and no tick labels.

A separate secondary scale is an axis in its own right, so its visibility now
comes from its own AxisSpec rather than the primary axis style. The behavior
when yscale2 is absent is unchanged, so simple charts keep the single-axis look.

Add RectCoordSecondaryAxisTest covering the secondary axis under both
AXIS_SINGLE and AXIS_DOUBLE primary styles, the primary axis surviving
alongside it, AXIS_NONE on the secondary scale still hiding it, and no
right-side axis appearing without a secondary scale.

### DIFF
--- a/core/src/main/java/inetsoft/graph/coord/RectCoord.java
+++ b/core/src/main/java/inetsoft/graph/coord/RectCoord.java
@@ -327,9 +327,13 @@ public class RectCoord extends Coordinate {
             yaxis2.setTickVisible(false);
          }
 
-         if((style & AxisSpec.AXIS_SINGLE2) == 0 ||
-            (facet != null && !yaxis2.isLabelVisible()))
-         {
+         // a separate secondary scale is an axis in its own right, so it's shown regardless
+         // of the primary axis style; visibility comes from its own spec.
+         boolean y2vis = yscale2 != null
+            ? yscale2.getAxisSpec().getAxisStyle() != AxisSpec.AXIS_NONE
+            : (style & AxisSpec.AXIS_SINGLE2) != 0;
+
+         if(!y2vis || (facet != null && !yaxis2.isLabelVisible())) {
             yaxis2.setZIndex(-1);
          }
 
--- a/core/src/test/java/inetsoft/graph/coord/RectCoordSecondaryAxisTest.java
+++ b/core/src/test/java/inetsoft/graph/coord/RectCoordSecondaryAxisTest.java
@@ -1,0 +1,132 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.graph.coord;
+
+import inetsoft.graph.*;
+import inetsoft.graph.data.DefaultDataSet;
+import inetsoft.graph.element.LineElement;
+import inetsoft.graph.guide.axis.DefaultAxis;
+import inetsoft.graph.scale.*;
+import inetsoft.test.BaseTestConfiguration;
+import inetsoft.test.ConfigurationContextInitializer;
+import inetsoft.test.SreeHome;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (changes requested):
## Code Review — PR #4432 (Bug #75646): Fix secondary y-axis labels not rendering in dual-axis charts

### Summary of the change
`RectCoord.createAxis()` decided whether to draw the right-side (secondary) y-axis by
bit-testing the **primary** y scale's `AxisSpec` style against `AXIS_SINGLE2` (`0x2 | 0x10`).
That test only makes sense when `yaxis2` is a *mirror* of the primary axis (`yscale2 == null`).
When a real second measure is bound to the secondary axis, `DefaultGraphGenerator` installs a
separate `yscale2` with its own `AxisSpec`, and that axis should be governed by its own spec,
not by the primary scale's style. Changing the default `AxisSpec` style from `AXIS_DOUBLE` (0x3)
to `AXIS_SINGLE` (0x1) in `247f34f8` made the old bit-test fail for ordinary dual-axis charts,
so the right axis was assigned `zIndex -1` and dropped (only its independently-painted title
remained visible).

The fix branches on whether `yscale2` is present:
```java
boolean y2vis = yscale2 != null
   ? yscale2.getAxisSpec().getAxisStyle() != AxisSpec.AXIS_NONE
   : (style & AxisSpec.AXIS_SINGLE2) != 0;
```

### Correctness review
- **Root cause matches the described regression.** Confirmed `AxisSpec.style` defaults to
  `AXIS_SINGLE` on `epic-74519` (post `247f34f8`) vs. `AXIS_DOUBLE` on `main`, and that
  `AXIS_SINGLE & AXIS_SINGLE2 == 0`, reproducing the reported drop of the secondary axis.
- **Scoped correctly to the real secondary-scale case.** `yscale2` is only non-null when a
  distinct scale is installed (`DefaultGraphGenerator.setupYScale`/`setYScale2`,
  `GraphDefault.setYScale2` for percent scales). In that path a fresh `new AxisSpec()` defaults
  to `AXIS_SINGLE` (`!= AXIS_NONE`), so ordinary dual-axis charts now correctly show the
  secondary axis, while `isSparkline()` explicitly sets `AXIS_NONE` on it and is still honored.
- **No regression for the mirrored-axis case** (`yscale2 == null`): behavior is byte-for-byte
  identical to before (same bit test against the primary style), so single-axis and "double
  axis, same scale on both sides" charts (e.g. `GraphGenerator` line ~4130 setting
  `AXIS_DOUBLE` for shared facet scales) are unaffected.
- **No interaction with the neighboring `AXIS_LABEL_OPPOSITE_SIDE` (bug #74356) logic** a few
  lines above (label visibility / line visibility for DC's relocated primary labels) — that
  code still reads `style` (the primary spec) and is untouched, and it governs `isLabelVisible`
  rather than the `zIndex` decision this PR changes, so no double-hide/double-show interaction.
- **X-axis is not affected** — `RectCoord` has no `xscale2` field/concept analogous to `yscale2`,
  so there's no equivalent latent bug on the top axis to fix here.
- Minimal, targeted diff (4 lines of logic change) with a clear explanatory comment.

### Test coverage
`RectCoordSecondaryAxisTest` is a solid regression guard:
- Secondary axis drawn for both `AXIS_SINGLE` and `AXIS_DOUBLE` primary styles (the actual
  regression case plus the pre-existing-default case).
- Primary axis still renders alongside the secondary one.
- Secondary axis with `AXIS_NONE` on its *own* spec stays hidden (sparkline path).
- No right-side axis appears at all when `yscale2` is absent (mirrored/no-secondary case
  unchanged).

Assertion messages are descriptive and explain the regression being guarded against, which will
help whoever touches this next.

### Other checklist items
- No security concerns — purely internal chart-layout logic, no user input handling.
- No API/serialization changes; `AxisSpec` public surface is untouched.
- No backward-incompatibility: default behavior when `yscale2` is absent is explicitly
  preserved and tested.
- `build` check passed, so the new test compiles and runs clean against the existing suite.

### Blocking issue for merge
- The **`claude-review`** required check is currently **failing** — not due to a finding, but
  because the automated review run errored out before completing
  ("*Claude encountered an error after 3m 31s*", stopped after the "Gather context" step). PR
  `mergeStateStatus` is `BLOCKED` as a result.
- No code changes are being requested here — the fix and tests look correct — but per review
  policy this can't be approved while a required check is failing. Please re-run/fix the
  `claude-review` workflow so it completes, then this should be clear to merge.

**Verdict:** Code change is correct, well-scoped, and well-tested. Requesting changes only
because of the failing `claude-review` CI check blocking the merge — not because of any issue
found in the diff itself.

jshobe-inetsoft (approved):

PR #4433, opened 2026-07-28
Bug #75649 - Fix "Labels on Opposite Side" leaving the original axis line in place

FigurePaper:
AxisSpec.getAxisStyle() built the effective style by OR-ing the opposite-side bit
into the configured style, so a single axis produced AXIS_SINGLE | 0x10 = 0x11,
which is not one of the defined styles. RectCoord.createAxis() then read it two
ways that disagreed: (style & 0x1) != 0 kept the primary axis, while
(style & AXIS_SINGLE2) != 0 also drew the secondary one, leaving both the left
and right y-axis lines visible. The same arithmetic drove x1vis/x2vis, so moving
x labels to the top left the bottom line behind as well. Labels were placed
correctly throughout because the label checks test the 0x10 bit on its own.

AXIS_SINGLE2 already means "one axis, on the opposite side" and renders exactly
that, so getAxisStyle() now resolves single plus relocated labels to it. The
primary-axis bit is cleared by masking rather than by replacing the whole value,
which preserves the cross-axis bit for AXIS_CROSS. A double axis still maps to
AXIS_DOUBLE2 and keeps both lines, and AXIS_SINGLE2 stays idempotent. Fixing it
where the style vocabulary is defined covers both x and y, and keeps the
vocabulary closed to the six declared constants instead of adding another
special case to the coordinate.

Measured plot bounds confirm the result is byte-identical to AXIS_SINGLE2
(x=5.0 w=756.0 versus x=40.0 for a left-labelled single axis), so the plot also
reclaims the width the hidden labels were holding.

Add RectCoordOppositeSideLabelsTest covering the four style mappings including
cross-bit preservation and idempotency, the y axis moving right, the x axis
moving to the top, a double-style axis keeping both lines, and a plain single
axis keeping its labels in place.

### DIFF
--- a/core/src/main/java/inetsoft/graph/AxisSpec.java
+++ b/core/src/main/java/inetsoft/graph/AxisSpec.java
@@ -89,11 +89,20 @@ public class AxisSpec implements Cloneable, Serializable {
    }
 
    /**
-    * Get the axis style for this coord.
+    * Get the axis style for this coord. This is the effective style and may differ from the
+    * style passed to setAxisStyle() when the labels have been moved to the opposite side.
     */
    @TernMethod(url = "#ProductDocs/chartAPI/html/Common_Chart_AxisSpec_getAxisStyle.htm")
    public int getAxisStyle() {
-      return labelOnSecondaryAxis ? (style | AXIS_LABEL_OPPOSITE_SIDE) : style;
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4435, opened 2026-07-29
fix(embed): track embed-assembly size per assembly, not one runtime-wide slot

Zxm-happy:
## Summary

- `RuntimeViewsheet` tracked embed-container sizing (`EmbedAssemblyInfo`) as a single field per runtime, keyed by whichever assembly's refresh happened to run last.
- wiz embeds several chart/table cards from one conversation on the **same shared `wizRuntimeId`** (different `assemblyName` each). Reopening a conversation with 2+ charts after a service restart mounts all their embed elements in the same instant, so their independent WebSocket connections race to refresh the same runtime nearly simultaneously.
- The last writer's assembly name/size silently overwrote the others', so `CoreLifecycleService#applyEmbedChartSize` stopped applying the real embed container size to the clobbered assembly — it rendered at its default/previous (smaller) size with blank space around it (see the second chart in the reporter's screenshot).
- Fix: replace the single field with a `Map<String, EmbedAssemblyInfo>` keyed by assembly name, so concurrently- or successively-embedded assemblies on the same runtime can never clobber each other. Updated all call sites accordingly (`VSRefreshService`, `CoreLifecycleService`, and four chart axes/legends/titles/refresh visibility services that only ever needed a boolean "is this runtime embedded at all" check).
- Extended the existing embed-assembly regression tests (`VSRefreshServiceEmbedAssemblyTest`, `CoreLifecycleServiceEmbedSizeTest`) with new cases covering the concurrent multi-assembly scenario, and adapted their stubbing to the new keyed API.

## Test plan

- [x] `mvn -pl core test -Dtest=VSRefreshServiceEmbedAssemblyTest,CoreLifecycleServiceEmbedSizeTest` — 6/6 pass
- [x] `mvn -pl core,server,schedule-server,connectors -am compile` — compiles clean
- [ ] Manual verification in wiz: rebuild/restart the StyleBI server, reopen a wiz conversation with 2+ charts after a restart, confirm both render at full width

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/RuntimeViewsheet.java
+++ b/core/src/main/java/inetsoft/report/composition/RuntimeViewsheet.java
@@ -55,6 +55,7 @@ import java.io.StringReader;
 import java.nio.charset.StandardCharsets;
 import java.security.Principal;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReentrantLock;
 
 /**
@@ -260,7 +261,11 @@ public class RuntimeViewsheet extends RuntimeSheet {
       wizardViewsheet = state.isWizardViewsheet();
 
       if(state.getEmbedAssemblyInfo() != null) {
-         embedAssemblyInfo = loadJson(EmbedAssemblyInfo.class, state.getEmbedAssemblyInfo(), mapper);
+         for(EmbedAssemblyInfo info : loadEmbedAssemblyInfoList(state.getEmbedAssemblyInfo(), mapper)) {
+            if(info.getAssemblyName() != null) {
+               embedAssemblyInfos.put(info.getAssemblyName(), info);
+            }
+         }
       }
 
       // Decode undo/redo checkpoints from VCDIFF deltas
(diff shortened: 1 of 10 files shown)


Zxm-happy:
Thanks for the review — addressed all three points in 0e2b72fa8:

1. **Legacy JSON format on rolling upgrade**: `loadEmbedAssemblyInfoList` now tries the new array format first and falls back to parsing the old single-object format (via the existing `loadJson(EmbedAssemblyInfo.class, ...)` helper) before giving up, so a `RuntimeViewsheetState` blob written by a not-yet-upgraded cluster node isn't silently dropped. Added `RuntimeViewsheetEmbedAssemblyInfoTest#fallsBackToTheLegacySingleObjectFormat` (plus a matching case for the current array format and for genuinely unparseable input) exercising the private loader via reflection.

2. **Orphaned map entries**: `putEmbedAssemblyInfo` now prunes entries whose assembly no longer exists on the runtime's viewsheet before inserting the new one, so repeated visualization-type switches on a long-lived runtime (wiz keeps one per conversation) don't grow the map without bound. Covered by `putEmbedAssemblyInfoPrunesEntriesForAssembliesNoLongerOnTheViewsheet` and, to guard against over-eager pruning, `putEmbedAssemblyInfoDoesNotPruneStillLiveSiblingAssemblies` (two still-embedded sibling assemblies must both survive).

3. **Naming clash with `Viewsheet#isEmbedded()`/`VSAssembly#isEmbedded()`**: renamed `RuntimeViewsheet#isEmbedded()` to `#hasEmbeddedAssembly()` and updated all 9 call sites. Grepped the repo afterward to confirm no leftover references and that the unrelated `assembly.isEmbedded()`/`vs.isEmbedded()` call sites elsewhere in the same files were left untouched.

All existing + new tests pass (`VSRefreshServiceEmbedAssemblyTest`, `CoreLifecycleServiceEmbedSizeTest`, `RuntimeViewsheetEmbedAssemblyInfoTest` — 11/11), and `core`/`server`/`schedule-server`/`connectors` all compile clean.

PR #4443, opened 2026-07-29
Bug #75801, fix freehand table showing no data for group cells

ngovus-inetsoft:
## Summary

Freehand tables using the crosstab-optimization path could render with no data rows (only the header row) for GROUP cells whose binding value is a date-level or formula-wrapped display name (e.g. `Year(Date)`, `None(NDateTime)`).

## Root Cause

`LayoutTool.createCrosstabCalcExpression()` built the generated `toList()`/`none()` script expression's data-lookup key from a resolved `CalcGroup`'s unqualified attribute name instead of the cell binding's own value. For these group cells, the binding's own value already matches the regenerated crosstab's actual column name, so substituting the unqualified attribute name (e.g. `Date` instead of `Year(Date)`, or `NDateTime` instead of `None(NDateTime)`) caused the lookup to miss, and the group cell never expanded.

## Fix

Restore using the cell binding's own value (`bind.getValue()` / `list.get(i).getValue()`) directly as the lookup key, as it was prior to the regression.

## Test Plan

- Reproduced with two viewsheets containing freehand tables using the crosstab optimization path (one with a plain date GROUP column, one with multiple nested date/passthrough GROUP columns) — both rendered with no data rows before the fix.
- Verified both now render with data after the fix, and confirmed via server-side logging that the generated lookup expression matches the regenerated crosstab's actual column headers.
- `community/core` compiles cleanly.

### DIFF
--- a/core/src/main/java/inetsoft/report/LayoutTool.java
+++ b/core/src/main/java/inetsoft/report/LayoutTool.java
@@ -2382,21 +2382,6 @@ public class LayoutTool {
    }
 
    /**
-    * Get the physical column name backing a crosstab-supported group cell. The
-    * cell's own binding value may hold the dimension's display full name (e.g.
-    * "None(Month(ndate))") rather than the physical column name ("Month(ndate)")
-    * for date-grouped dimensions, so prefer the resolved CalcGroup's attribute
-    * (unqualified, unlike getName() which may carry an "entity." table prefix).
-    */
-   private static String getGroupColumnName(CalcGroup group, TableCellBinding cell) {
-      if(group != null) {
-         return BindingTool.getPureField(group).getAttribute();
-      }
-
-      return cell.getValue();
-   }
-
-   /**
     * Create calc for crosstab supported expression.
     */
    private static String createCrosstabCalcExpression(List<TableCellBinding> list,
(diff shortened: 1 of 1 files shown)


ngovus-inetsoft:
This reverts the group-cell lookup-key change from #4387 (Bug #75747) to fix #75801 — freehand tables using the crosstab-optimization path were rendering with no data rows for GROUP cells whose binding value already matched the regenerated crosstab's column name (e.g. `Year(Date)`, `None(NDateTime)`).

Note: reverting that change reintroduces the original scenario #4387 fixed (a nested date-grouped dimension produced via "Convert Crosstab to Freehand Table" showing no data). That issue has been marked in progress and will be investigated seperately

jshobe-inetsoft (changes requested):
Requesting changes for two reasons:

1. Claude's automated review flagged an unresolved concern (not a clean approval): reverting the `getGroupColumnName()` change reintroduces the original scenario that #4387 (bug #75747) fixed, and recommended re-verifying a joined/multi-source group-cell case before merging. Your own follow-up comment acknowledges this trade-off and notes bug #75747 would be "investigated separately."

2. That separate investigation is #4445, which has just been merged. It reworks the same `getGroupColumnName()`/`createCrosstabCalcExpression()` logic in `LayoutTool.java` with a more robust fix (resolving the lookup key against the regenerated crosstab's own result table rather than blindly reverting), and its test suite covers this PR's #75801 scenario as well as the #75747 scenario. As a result, this PR now has a merge conflict against `main`.

Please close this PR if #4445 already resolves #75801 (it looks like it does, based on the added test cases for the sole/outer date dimension and nested passthrough dimension scenarios), or rebase and re-verify against the new #4445 logic if there's still a gap.

PR #4445, opened 2026-07-29
Bug #75747, fix data missing when converting crosstab to freehand table

ngovus-inetsoft:
## Summary
- Converting a crosstab with a nested date-grouped row dimension (e.g. `Month(ndate)` nested under another dimension) to a freehand table produced a table with only the header row — all data rows were missing.
- Root cause: `LayoutTool.createCrosstabCalcExpression` built the generated `toList()`/`none()` script expressions from `TableCellBinding.getValue()`, which for such nested date-grouped dimensions holds the dimension's display full name (e.g. `"None(Month(ndate))"`) instead of the physical column name backing the crosstab's actual data (`"Month(ndate)"`). The generated formula referenced a nonexistent column, the lookup returned `null`, and the row group never expanded.
- **This is a corrected re-fix.** The first attempt (merged as #4387, then reverted as bug #75801/#4443) unconditionally resolved the column name via a `CalcGroup`, which broke other dimensions whose binding value was *already* the correct physical column name (e.g. `"Year(Date)"` as a sole/outer date dimension, or multiple nested passthrough date dimensions). This revision instead only substitutes the unwrapped column name when the cell's raw binding value does not match an actual column of the regenerated crosstab's own result table — checked directly against that table (threaded through via the previously-unused `base` parameter of `VSLayoutTool.createCalcLens`), rather than re-derived from the worksheet's column selection, so the decision can never disagree with what the generated formula is actually evaluated against.

## Test plan
- [x] Added `LayoutToolTest` covering all three previously-manually-verified scenarios: a date dimension nested under another group (#75747), a sole/outer date dimension (#75801/CroTable7), and multiple nested passthrough date dimensions (#75801/Special2).
- [x] Verified manually on a live running server: all three scenarios (GroupAndTab, CroTable7, Special2) now render correctly in a single test pass.
- [x] `./mvnw -pl core test-compile` and `-Dtest=LayoutToolTest` pass.

### DIFF
--- a/core/src/main/java/inetsoft/report/LayoutTool.java
+++ b/core/src/main/java/inetsoft/report/LayoutTool.java
@@ -629,7 +629,7 @@ public class LayoutTool {
       return aref;
    }
 
-   protected static void fillCalcTableLens(FormulaTable table,
+   protected static void fillCalcTableLens(FormulaTable table, TableLens base,
                                            VariableTable vars,
                                            boolean crossTabSupported)
    {
@@ -729,7 +729,7 @@ public class LayoutTool {
                      }
                   }
 
-                  String exp = createCalcBindingExpression(table, n2c, layout,
+                  String exp = createCalcBindingExpression(table, base, n2c, layout,
                      bind, c2p, new Point(r, j), crossTabSupported, sortOthersLast);
                   clens.setFormula(r, j, exp);
                }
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):
Automated review: Claude's automated code review found no blocking issues (approach correctly threads the group-cell lookup-key decision through the regenerated crosstab's own result table, confirmed against all three previously-broken scenarios, with new unit test coverage). All CI checks pass and the PR has no merge conflicts. Approving and merging.

PR #4446, opened 2026-07-29
Bug #75819, fall back to listing all cubes when OLAP binding-tree source lookup fails

ngovus-inetsoft:
## Summary
- `AssetEventUtil.getCubes()` had a shortcut path: when the binding-tree request scoped the lookup to a single cube `source`, it resolved that one cube via `getXCube()`/`getCubeNode()` and returned immediately - even if the lookup came back empty.
- If that single-cube resolution failed (e.g. stale/renamed source info), the method returned an empty list instead of falling back to enumerating all registered OLAP datasources, which could collapse the entire "Cubes" tree in the assembly/crosstab binding pane to blank.
- Fixed by falling back to the full datasource loop whenever the single-cube lookup doesn't resolve a node, using local `cubePrefix`/`cubeSource` variables so the fallback loop still sees the original `prefix`/`source` values.

## Test plan
- [x] `core` module compiles cleanly (`mvnw compile -o`)
- [x] Manually reproduced the reported binding-pane scenario (import an OLAP asset, bind a cube dimension/measure to a crosstab in Composer) with temporary debug logging; confirmed the fix's fallback path is defensive and doesn't change behavior in the standard Composer design-mode flow (single-cube shortcut isn't reached there), while protecting against an empty tree in the affected caller context (`rvs.isRuntime()` / embedded-viewsheet paths that use the single-cube shortcut).

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/event/AssetEventUtil.java
+++ b/core/src/main/java/inetsoft/report/composition/event/AssetEventUtil.java
@@ -105,26 +105,32 @@ public class AssetEventUtil {
       List<Node> cubeList = new ArrayList<>();
 
       if(source != null) {
-         if(source.startsWith(Assembly.CUBE_VS)) {
-            source = source.substring(Assembly.CUBE_VS.length());
-            int idx = source.lastIndexOf("/");
+         String cubePrefix = prefix;
+         String cubeSource = source;
+
+         if(cubeSource.startsWith(Assembly.CUBE_VS)) {
+            cubeSource = cubeSource.substring(Assembly.CUBE_VS.length());
+            int idx = cubeSource.lastIndexOf("/");
 
             if(idx >= 0) {
-               prefix = source.substring(0, idx);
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (changes requested):
Claude's automated review (after the latest commit) found no blocking issues, but the `build` check (Validation Build) is currently failing, which blocks merge per review policy regardless of review content.

The failure is in `inetsoft.sree.security.DashboardRegistryOrgLifecycleTest` (3 errors: `NullPointerException` on `this.kvStorage` in `PropertiesEngine.save()` via `SecurityTestDataBuilder.setup()`), which is unrelated to this PR's change in `AssetEventUtil.getCubes()`. This looks like pre-existing test instability rather than a regression introduced here, but please re-run/fix CI so the check is green before this can be merged.

larryliang-inetsoft (approved):

jshobe-inetsoft (approved):

PR #4453, opened 2026-07-30
fix(wiz): copy pairing code via ngx-clipboard instead of navigator.clipboard

jkidacki-inetsoft:
## Summary
- `navigator.clipboard` only exists in secure contexts (HTTPS, or `http://localhost`/`127.0.0.1`), so the "Connect to Claude" pairing code Copy button silently failed when the server was accessed over plain HTTP on a non-localhost host (e.g. `http://192.168.5.148:14222/`).
- Switch the Copy button to the `ngx-clipboard` directive, which is already used elsewhere in the app (`share-link-dialog.component.ts`) for the same reason — it copies via `document.execCommand('copy')` and works regardless of secure-context restrictions.

## Test plan
- [x] Added unit tests covering `onCopySuccess`/`onCopyError` and the `ngxClipboard` directive wiring
- [x] `ng test --project portal` passes for `connect-to-claude.component.spec.ts` (8/8)
- [x] `eslint` clean on changed files


larryliang-inetsoft (approved):

PR #4457, opened 2026-07-30
Bug #75735, verify theme jar was relocated when renaming an organization

ngovus-inetsoft:
## Summary
- Renaming an organization can leave a themes's `jarPath` pointing at a file that was never physically relocated: `copyThemes()`'s org-owned-theme branch only rewrote the `jarPath` string, trusting the earlier `copyDataSpace()` rename to have already moved the JAR — but `DataSpace.rename()` swallows `FileNotFoundException`/`IOException` without propagating, so a failed move goes unnoticed until the theme 404s.
- `copyThemes()` now verifies the new path exists after the rename and, if not, falls back to copying the jar bytes directly from the old path (the same verification the sibling global-theme branch already performs), and logs a warning if the jar is missing from both locations.

## Test plan
- [ ] `./mvnw test -pl core -Dtest=AbstractEditableAuthenticationProviderStaticDepTest`
- [ ] Manual repro: create a global theme, clone an org, create an org-specific theme and toggle "Default for This/All Organization(s)" on/off, rename the org, confirm the theme still loads (no 404) on Presentation → Themes


jshobe-inetsoft (approved):

PR #4459, opened 2026-07-30
Bug #75827, preserve recycle status when migrating autosave files across orgs

ngovus-inetsoft:
## Summary
- `AutoSaveUtils.migrateAutoSaveFiles()` stripped the `recycle/` prefix to split out the autosave filename's fields, but never re-applied it when building the migrated filename — a discarded/recycled draft silently turned into an active autosave file after an org copy or rename, so it dropped off the EM "Auto Saved Files" recycle-bin tree (`ContentRepositoryTreeService.addRecycleAutoSaved()` only lists paths still carrying the recycle prefix).
- `migrateAutoSaveFiles()` now resolves the blob bucket via an explicit `storageOrgId` parameter instead of the ambient principal/`OrganizationContextHolder` context — removes a fragile dependency and makes the target bucket obvious at each call site.

## Test plan
- [x] `AutoSaveServiceOrgLifecycleTest` — added `migrateAutoSaveFiles_recycledDraft_keepsRecyclePrefixAfterMigration`, pins the fix; updated the existing `6g` test for the new signature. All pass.
- [x] `IdentityServiceAutoSaveOrgLifecycleTest` — unaffected, still passes.
- [x] Manual repro: create a viewsheet as host org admin, discard the draft (goes to Auto Saved Files recycle bin), clone the org — file now correctly still shows up in the new org's Auto Saved Files tree.

### DIFF
--- a/core/src/main/java/inetsoft/sree/security/AbstractEditableAuthenticationProvider.java
+++ b/core/src/main/java/inetsoft/sree/security/AbstractEditableAuthenticationProvider.java
@@ -262,7 +262,7 @@ public abstract class AbstractEditableAuthenticationProvider
             // (still keyed by the source org's user ids), so migrate the files in place there --
             // not in fromOrganization's bucket. Pass newOrgID explicitly rather than relying on
             // the ambient principal/org context to resolve the right bucket.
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):

PR #4463, opened 2026-07-30
Bug #75824, fail fast on Ignite cache mode mismatch from a minReplicas change

jkidacki-inetsoft:
## Summary
- Cache mode (`PARTITIONED` vs `REPLICATED`) is derived independently by each Ignite node from the configured `minReplicas` at startup and is never renegotiated afterward. Changing `minReplicas` across the threshold that flips this and applying it as a rolling update (instead of restarting the whole cluster) leaves old and new pods disagreeing on cache mode for the same cache group.
- Previously this surfaced much later as an unexplained `IgniteException: Cache mode mismatch` whenever some code path first touched the affected (often lazily-created) cache, e.g. as a 500 error on the EM monitoring page.
- Each node now publishes its computed cache mode as a discovery attribute and checks it against already-joined server nodes before creating any cache, so the failure happens immediately on startup with an actionable message. A node that fails this check also leaves the discovery ring cleanly instead of lingering as a zombie member.

## Test plan
- [x] Added `IgniteClusterCacheModeConsistencyTest` (in-JVM two-node harness, matching sibling Ignite tests in this package) covering both the mismatch-fails and match-succeeds cases. Verified locally by temporarily removing `@Disabled` (matches convention of other multi-node Ignite tests in this package, which are excluded from routine CI runs).
- [x] `mvn test -pl core -Dtest=IgniteClusterCacheModeConsistencyTest` passes.

Redmine: https://www.inetsoft.com/redmine/issues/75824

### DIFF
--- a/core/src/main/java/inetsoft/sree/internal/cluster/ignite/IgniteCluster.java
+++ b/core/src/main/java/inetsoft/sree/internal/cluster/ignite/IgniteCluster.java
@@ -103,6 +103,16 @@ public final class IgniteCluster implements inetsoft.sree.internal.cluster.Clust
          r -> new GroupedThread(r, "IgniteMapEvents"));
 
       if(!config.isClientMode()) {
+         try {
+            checkCacheModeConsistency();
+         }
+         catch(RuntimeException e) {
+            // the node already joined discovery above; leave cleanly instead of lingering as a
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4417, opened 2026-07-27
Bug #75791, document required INETSOFT_ADMIN_PASSWORD for Docker Compose setup

ngovus-inetsoft:
## Summary
- The storage container requires `INETSOFT_ADMIN_PASSWORD` to be set and meet complexity requirements, but this was undocumented and failures surfaced only as a vague `exit 1` in the storage container logs.
- Document the requirement in the Quickstart section of the README, add a commented-out `INETSOFT_ADMIN_PASSWORD` entry to the community-examples `.env` file, and add a Troubleshooting entry describing the error and how to fix it.

## Test plan
- [ ] Review rendered Markdown for README.md and Troubleshoot.md
- [ ] Confirm `.env` file still parses correctly with docker compose (single/no active password line doesn't break anything when commented out)

PR #4470, opened 2026-07-31
Bug #75791-1, don't advertise a working sample admin password, label shell examples by terminal

ngovus-inetsoft:
## Summary
- Replace the sample `INETSOFT_ADMIN_PASSWORD` value `Test@admin1` with `changeme` in README.md, Troubleshoot.md, and community-examples/.env. `Test@admin1` satisfied the password complexity requirements, so a user copy-pasting it verbatim would end up with a real, working admin credential — defeating the point of requiring a password in the first place. `changeme` intentionally fails the requirements, so the storage container still refuses to start (same `exit 1`) until the user sets their own password.
- Label the example shell commands for setting `INETSOFT_ADMIN_PASSWORD` by terminal (Linux/macOS vs. Windows Command Prompt) and swap the PowerShell snippet for a Command Prompt (`set VAR=value`) one, matching the Command Prompt instructions used later in the README.
- Note that the shell-variable approach only applies to the current terminal session, so `docker compose up` must be run in that same session.

Follow-up to PR #4417.

## Test plan
- [ ] Confirm no remaining occurrences of `Test@admin1` in the repo
- [ ] Manually verify the CMD snippet syntax (`set VAR=value`) is correct
- [ ] Verify docs render correctly (code fences, headings)


larryliang-inetsoft (approved):

PR #4471, opened 2026-07-31
Bug #75810, fix O(N x D) latency when renaming a table style folder

ngovus-inetsoft:
## Summary
- `RenameAssetController.renameTableStyleFolder()` called `manager.save()` once per recursion level (once per subfolder being renamed). `LibManager.save()` rewrites the entire `style/folders.xml` for the whole library on any pending folder transaction, so a rename of a folder with D subfolders in a library with N total folders cost O(N x D) instead of O(N).
- The recursive rename/query calls (`renameTableStyle`, `renameTableStyleFolder`, `getTableStyles`, `getTableStyleFolders`) all operate on in-memory state and don't depend on intermediate saves — only the single `manager.save()` already performed by the caller after the full recursive rename completes is needed.
- Removed the redundant per-level `save()` (and the now-unused `LOG` field/imports it required).

## Test plan
- [ ] Rename a table style folder with a few (2-3) styles in a library with several hundred unrelated styles; confirm `POST api/composer/asset-tree/rename-asset` completes quickly and the rename (including nested subfolders/styles) persists correctly after a page reload.
- [ ] Rename a table style folder containing nested subfolders; confirm all nested folders/styles are renamed and persisted.

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/RenameAssetController.java
+++ b/core/src/main/java/inetsoft/web/composer/RenameAssetController.java
@@ -37,8 +37,6 @@ import inetsoft.util.audit.Audit;
 import inetsoft.web.admin.content.repository.ResourcePermissionService;
 import inetsoft.web.composer.model.RenameAssetEvent;
 import inetsoft.web.viewsheet.command.MessageCommand;
-import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


larryliang-inetsoft (approved):

PR #4473, opened 2026-07-31
Bug #75838, migrate schedule tasks with time-range conditions during org clone

ngovus-inetsoft:
## Summary
- The fix for Bug #70758 (#711) made org clone skip migrating any schedule task whose `TimeCondition` used a time range, on the premise that non-host orgs don't support time ranges.
- `TimeRange` is actually global/server-wide config (not org-scoped) and the `TimeCondition`'s embedded `TimeRange` is a self-contained snapshot needing no org-specific rewriting during migration, so there's no reason to drop these tasks.
- The blanket skip caused legitimate schedule tasks to silently disappear (no log, no error) whenever a new org was cloned from host, if the task had a time-range trigger condition.
- This reverts the blanket skip so all non-internal schedule tasks migrate normally during org clone, regardless of their conditions.

## Test plan
- [x] Enabled multi-tenancy, created a schedule task with a time-range condition in the host org
- [x] Created a new org cloned from host — confirmed the task now migrates and shows up on the new org's schedule page
- [x] Re-tested the original repro for Bug #70758 (create simple schedule w/ time range → clone org → recreate simple schedule → switch back to host) — could not reproduce the duplicate-task symptom

### DIFF
--- a/core/src/main/java/inetsoft/util/BlobIndexedStorage.java
+++ b/core/src/main/java/inetsoft/util/BlobIndexedStorage.java
@@ -626,20 +626,6 @@ public class BlobIndexedStorage extends AbstractIndexedStorage {
          else if(entry.isScheduleTask()) {
             //ignore internal tasks, but do not let them pass to be generically handled
             if(!ScheduleManager.isInternalTask(entry.getName())) {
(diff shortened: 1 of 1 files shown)


larryliang-inetsoft (approved):

PR #4475, opened 2026-07-31
Bug #75820, fix data tip position for crosstab/freehand table popups

jkidacki-inetsoft:
## Summary

Data tips whose target view is a crosstab or freehand (calc) table rendered at their static design-time layout position instead of at the mouse cursor.

## Root Cause

The fix for Bug #74751 (data tip drifted from the cursor while scrolling) restricted `VSDataTipDirective`'s imperative `top`/`left`/`position` styling to elements carrying the `focus-assembly` CSS class, on the assumption that all other elements own their position via Angular template bindings. However `VSCrosstab`, `VSTable`, and `VSCalcTable` apply this same directive directly to their own `OnPush` root element (which never carries `focus-assembly`) specifically so that root can be repositioned when the component itself is configured as another assembly's data tip view. The gate silently disabled that repositioning, so these tip views fell back to their static, design-time `top`/`left` — typically far from the mouse.

## Fix

- Removed the `focus-assembly` gate in `VSDataTipDirective.ngDoCheck()`: `left`/`top`/`position: absolute` are now applied unconditionally whenever the element is the active, visible data tip. The static template-bound layout position is never the correct one in that state, so this is always safe.
- To avoid reintroducing #74751 (which relied on the gate to stop stale positions from sticking on `OnPush` hosts during scroll), added a `DataTipService.scrolled` notification fired from the viewer's existing `scroll()` handler. `VSDataTipDirective` subscribes to it and calls `ChangeDetectorRef.markForCheck()` — but only when it is the currently active tip — so the `OnPush` host's view is explicitly marked dirty and `ngDoCheck` recomputes position on scroll.
- Extended the existing "skip DOM writes when unchanged" cache check (from Bug #75512) to also key on `scrollLeft`/`scrollTop`, since a scroll alone doesn't change the cached mouse coordinates and would otherwise short-circuit the recompute even after `markForCheck()`.
- Added `implements OnDestroy` to `VSDataTipDirective` (previously it had none) to unsubscribe the new `scrolled` subscription and tear down its click/enter/leave listeners — fixes a pre-existing listener leak as a side effect of needing `OnDestroy` for the new subscription; called out here since it's incidental to the stated bug.
- Updated unit test mocks whose hand-rolled `DataTipService` stubs needed the new `scrolled` Subject: the 4 `*.spec.ts` files that failed outright (`vs-check-box`, `vs-group-container`, `vs-table`, `vs-gauge`), plus 4 `*.tl.spec.ts` files that were confirmed (by actually running `test:portal:tl`) to render `[VSDataTip]` through `TestBed` (`vs-combo-box`, `vs-image`, `vs-range-slider`, `vs-slider` display specs).

## Review feedback addressed

Per the automated review on this PR:
- **Fixed**: `ngDoCheck` was reading `DataTipService.viewerOffset` (which can measure the DOM) up to 3 times per invocation for the new scroll-aware cache key, undercutting the Bug #75512 "skip expensive DOM reads" optimization. Now hoisted to a single read, reused for both the cache check and the position math.
- **Fixed**: added `vs-data-tip.directive.spec.ts` covering the non-`focus-assembly` repositioning path, the hide-when-not-current path, the single-read-per-check behavior, and that scroll only calls `markForCheck()` when this instance is the active tip.
- **Fixed**: backfilled the other `DataTipService` mocks that were actually reachable through `TestBed` rendering (verified via `test:portal:tl`, not guessed).
- **Noted, not changed**: the `OnDestroy`/listener-cleanup addition (see Fix section above) — a good fix but not the stated bug; called out for traceability rather than reverted, since removing the cleanup would leak the new `scrolled` subscription per directive instance.

## Test Plan

- [x] Reproduced with the attached `tipview(1).vso` from the Redmine issue: crosstab/freehand table data tip now renders at the mouse cursor instead of the bottom of the pane.
- [x] Manually verified in the running portal.
- [x] `npm run build` (portal/em/elements/viewer-element) succeeds.
- [x] Full Vitest suite (`npm run test:portal`, portal + em projects): 190 files / 951 tests + 101 files / 356 tests, all passing.
- [x] `npm run test:portal:tl` re-run to confirm no `[VSDataTip]`-related failures remain (remaining failures in that suite are pre-existing/unrelated flakiness — it's not yet wired to CI per project docs).
- [x] Two code-reviewer subagent passes plus the automated PR review, with feedback addressed as noted above.

Fixes Redmine #75820.


jkidacki-inetsoft:
Thanks for the review — addressed all four points in 319acd665:

1. **DOM-read regression (Medium)**: fixed. Hoisted a single `viewerOffset` read per `ngDoCheck` invocation, reused for both the cache key and the later `viewerRect` usage (was up to 3 reads, now 1).
2. **Test coverage gap (Low)**: added `vs-data-tip.directive.spec.ts` covering the non-`focus-assembly` repositioning path, the hide-when-not-current path, the single-read-per-check behavior, and that scroll only triggers `markForCheck()` for the instance that's actually the active tip.
3. **Other unpatched mocks (Low)**: ran `test:portal:tl` directly to find the actual (not just suspected) blast radius — 4 more spec files genuinely render `[VSDataTip]` through `TestBed` (`vs-combo-box`, `vs-image`, `vs-range-slider`, `vs-slider` display specs) and needed the `scrolled` Subject added. The rest of the hand-rolled mocks you flagged were confirmed safe (don't reach `ngOnInit` today), so left as-is.
4. **`ngOnDestroy` scope creep (Nit)**: called out explicitly in the PR description now, with the reasoning for why it's included (required to unsubscribe the new `scrolled` subscription without leaking).

Build, full `test:portal` suite (190 files / 951 tests + 101/356), and a re-run of `test:portal:tl` all green — the tl suite's remaining failures are pre-existing/unrelated (that suite isn't wired to CI yet per project docs).

larryliang-inetsoft (approved):

PR #4478, opened 2026-07-31
Bug #75823, fix range slider not showing minimum value when bound to a cube aggregate measure

ngovus-inetsoft:
## Summary
- When a RangeSlider (TimeSlider) is bound to an OLAP cube aggregate measure, `TimeSliderVSAQuery.getTableAssembly()` routes through the member-list branch (used for cube/MEMBER-type bindings) instead of the standard numeric MIN/MAX aggregation branch.
- That branch did not exclude members whose measure value is `null`. A null-valued member sorts to the front (ascending sort) and becomes the slider's "minimum" selection value, which renders as a blank label instead of a number — while the maximum still displays correctly.
- Fix: add the same null-exclusion pre-runtime condition that the non-cube aggregation branch already applies (see comment "ignore null otherwise a null could cause a numeric column not usable as in a range slider"), so a null-measure member can no longer become the slider's min/max anchor.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/TimeSliderVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/TimeSliderVSAQuery.java
@@ -139,6 +139,15 @@ public class TimeSliderVSAQuery extends AbstractSelectionVSAQuery {
                sortinfo.addSort(sort);
                tassembly.setSortInfo(sortinfo);
             }
+
+            // ignore null otherwise a member with a null aggregate/measure value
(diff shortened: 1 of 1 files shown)


larryliang-inetsoft (changes requested):
Not merging yet: the `build` check is **CANCELLED**, not passing (`status=COMPLETED conclusion=CANCELLED`). The Claude review itself is clean — "No blocking issues found" — so this is purely about the check. Please re-run the build; once it's green this looks ready.

jshobe-inetsoft (approved):

PR #4480, opened 2026-08-01
fix(wiz): read the PrimaryKey column attribute as a boolean, not a string

larryliang-inetsoft:
## The bug

`JDBCHandler` writes the primary-key flag as a **boolean**:

```java
// JDBCHandler.java:3179
node.setAttribute("PrimaryKey", isPrimary);   // isPrimary is a boolean
```

`XNode.setAttribute(String key, Object val)` stores the value as an `Object` without coercing it, so the attribute stays a `Boolean`. `MetadataApiService` then read it as a string:

```java
boolean isPK = "true".equals(columnNode.getAttribute("PrimaryKey"));
```

`String.equals(Boolean)` is **always false**. Every column was therefore reported as a non-key, and `dataset.primary_key` came back `null` for every table on every datasource.

## Why this went unnoticed

The sibling `ForeignKey` attribute is written two lines later in the same block (`JDBCHandler:3182`) and works correctly — because `extractForeignKeys` reads it as an *object* rather than string-comparing it. So the relationship graph looked healthy while primary keys were silently empty, which made the failure look like missing data rather than a read bug.

## Impact

Anything gated on a known primary key was silently dark. Confirmed live against two datasources, where the generated annotation prose describes a column as a primary key while the structured field disagrees:

| table | column | annotation says | `isPrimaryKey` |
|---|---|---|---|
| `odoo.res_country` | `id` | "Surrogate primary key for the country row" | `false` |
| `odoo.res_partner` | `id` | "Surrogate primary key for the partner" | `false` |
| `suitecrm.opportunities` | `id` | "Unique UUID primary key for the opportunity record" | `false` |

## The fix

Both read sites (`MetadataApiService:175` and `:1852`) now share one helper that accepts a `Boolean` and still accepts a `String` — an `XNode` that has round-tripped through XML carries its attributes as text rather than their original types.

## Tests

Five regression tests in `MetadataApiServicePrimaryKeyTest`. The boolean case is the one that reproduces the bug.

Verified red/green rather than assumed: reverting the helper body to the old `"true".equals(attr)` form fails 2 of the 5 (`booleanTrueIsAPrimaryKey`, `stringTrueIsStillAccepted`), and the suite passes once restored.

```
./mvnw -o test -pl core -Dtest='MetadataApiServicePrimaryKeyTest'
Tests run: 5, Failures: 0
```

Also built and deployed as a local image (core-only rebuild; no new beans, so AOT-safe) and confirmed a clean boot.

## Note for reviewers

This corrects the *read* path only. Datasources annotated before this fix still carry the empty `primary_key` in their stored documents; they need a fresh annotation run to repopulate it, since `buildAnnotationInputs` deliberately reuses the cached document without writing.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
@@ -172,7 +172,7 @@ public class MetadataApiService {
 
          String columnName = columnNode.getName();
          String columnType = columnNode instanceof XTypeNode typeNode ? typeNode.getType() : null;
-         boolean isPK = "true".equals(columnNode.getAttribute("PrimaryKey"));
+         boolean isPK = isPrimaryKeyColumn(columnNode);
          Integer length = (Integer) columnNode.getAttribute("length");
          String comment = (String) columnNode.getAttribute("comment");
          List<String[]> foreignKeys = extractForeignKeys(columnNode);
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
Superseded by #4481, now merged. The PrimaryKey boolean/string fix (`1ce4015cd`) is included there as one of its commits, so merging this separately would replay the same change. Closing rather than merging.

For the record, since the fix is easy to lose track of: `JDBCHandler` writes the `PrimaryKey` column attribute as a **Boolean**, while `MetadataApiService` compared it with `"true".equals(...)` — a String against a Boolean, which is **always false**. Every column on every datasource reported as a non-key and `dataset.primary_key` was null everywhere, silently disabling every feature gated on a known primary key.

PR #4488, opened 2026-08-03
Fix Clear All recycle bin button stuck in loading state

ngovus-inetsoft:
## Summary
- In the EM Content Repository page, clicking "Clear All" on the Recycle Bin issues a `DELETE .../recycle-bin/entries` request and only resets the loading flag in the `subscribe()` success callback.
- If that request fails (e.g. when clearing the recycle bin for a switched/cloned organization rather than the host org), the success callback never fires, so `loading` stays `true` and the button spins forever.
- Added an `error` handler alongside `next` so the loading state is always cleared, whether the request succeeds or fails.

## Test plan
- [ ] Clone an org from the host org
- [ ] Switch to the cloned org in EM > Content > Repository
- [ ] Delete a viewsheet so it lands in Recycle Bin
- [ ] Click Recycle Bin, then click "Clear All"
- [ ] Verify the button no longer gets stuck in a permanent loading state


jshobe-inetsoft (approved):

PR #4489, opened 2026-08-03
Bug #75885, add created-by/created-date metadata to data model folders

ngovus-inetsoft:
## Summary
- Data Model folders (folders inside a JDBC data source's logical/physical model tree) were stored as bare `String`s on `XDataModel`, so they never carried creation metadata — the "Created By"/"Creation Date" columns in the portal Data Model folder browser were always blank, for every user (not just non-admin R/W users, as originally reported).
- `XDataModel` now tracks a `createdBy`/`createdDate` per folder, persisted in the data model XML (backward-compatible with existing XML lacking the new attributes).
- Folder creation (`DataModelFolderManagerService.setDataModelFolder`) stamps the creating principal and timestamp.
- `DatabaseModelBrowserService.getDataModelFolders` resolves and displays the creator's alias and formatted creation date, mirroring the existing pattern used for LogicalModel/PhysicalModel/Vpm.
- Renaming a folder now preserves its creation metadata (previously done via remove+re-add, which silently dropped it); the new `renameFolder()` also guards against creating a duplicate entry if renamed onto an existing folder name.

## Test plan
- [x] `./mvnw clean install -pl core -am -DskipTests` compiles cleanly
- [ ] Manually verify: create a Data Model folder as a non-admin user with R/W permission on Data Model Folder → "Created By"/"Creation Date" now populate in the folder browser grid
- [ ] Manually verify: rename a Data Model folder → creator/date persist after rename

### DIFF
--- a/core/src/main/java/inetsoft/uql/erm/XDataModel.java
+++ b/core/src/main/java/inetsoft/uql/erm/XDataModel.java
@@ -892,6 +892,10 @@ public class XDataModel implements Cloneable, Serializable, XDomain,
     * Rename a folder in the data model, preserving its creation metadata.
     */
    public void renameFolder(String oldName, String newName) {
+      if(Tool.equals(oldName, newName)) {
+         return;
+      }
+
       FolderInfo info = getFolderInfo(oldName);
 
       if(info != null) {
--- a/core/src/main/java/inetsoft/uql/erm/XDataModel.java
+++ b/core/src/main/java/inetsoft/uql/erm/XDataModel.java
@@ -796,8 +796,18 @@ public class XDataModel implements Cloneable, Serializable, XDomain,
    public void writeXML(PrintWriter writer) {
       writeStart(writer);
 
-      for(String folder : folders) {
-         writer.println("<folder name=\"" + Tool.escape(folder) + "\"/>");
+      for(FolderInfo folder : folders) {
+         writer.print("<folder name=\"" + Tool.escape(folder.getName()) + "\"");
+
+         if(folder.getCreatedBy() != null) {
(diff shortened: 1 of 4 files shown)


ngovus-inetsoft:
Thanks for the review — fixed in bd1a4d370:

- **Self-rename bug (confirmed):** `renameFolder` now short-circuits when `oldName.equals(newName)`, so it's a no-op instead of deleting the folder.
- **Org ID dropped:** `createdBy` now stores the full identity key (`IdentityID.convertToKey()`) rather than just the plain username, and the read side decodes it back with `IdentityID.getIdentityIDFromKey()` instead of reconstructing the org from whatever org happens to be "current" at display time.
- **Import-merge path dropping metadata:** `DataSourceRegistry.parseDataModel` now reads `createdBy`/`createdDate` off the `<folder>` element the same way `XDataModel.parseXML` does, instead of calling the metadata-less single-arg `addFolder(String)`.
- **Default label/date alignment:** `setFolderCreatedInfo` now sets `createdDate(-1)`/`createdDateLabel("")` when there's no metadata, matching the `DataSourceService` pattern for Logical/Physical Model/Vpm.

Left the `XLogicalModelAsset`/dependency-asset-import gap as-is per your note — it's inferring folder existence from a model's folder name rather than a `<folder>` XML element with attributes, so there's no metadata to carry through there; not a regression from this PR.

Didn't add unit tests for this change (no existing test file for `XDataModel`/`DataModelFolderManagerService` to extend) — happy to add one for `renameFolder` if that's wanted before merge.

jshobe-inetsoft (approved):

PR #4490, opened 2026-08-03
Bug #75891, propagate user rename to recycle bin entries

ngovus-inetsoft:
## Summary
- Fixes an NPE in `RecycleUtils.restoreSheet` when restoring a recycled asset after the owning user has been renamed.
- Root cause: `UserTreeService.editUser()` -> `renameUserAsset()` migrates a renamed user's asset storage (IndexedStorage, MV, dependency storage, dashboard cycle info) to the new username, but never updated `RecycleBin.Entry.originalUser`. `RecycleUtils.restoreSheet` then looked up the recycled sheet under the stale old username, `getAssetEntry` returned `null`, and `oldEntry.getType()` NPE'd.
- Adds `RecycleBin.renameUser(oldUser, newUser)` (mirrors the existing `renameFolder` cascade) and calls it from `UserTreeService.renameUserAsset` alongside the other user-rename storage migrations.
- Adds defensive null-checks in `RecycleUtils.restoreSheet` so any already-stale recycle bin entries (created before this fix) fail with a clear `MessageException` instead of an NPE.

## Repro
1. Create user "guest", import a viewsheet owned by guest.
2. Delete guest's private viewsheet (moves to Recycle Bin).
3. Rename user "guest" to a new name.
4. Restore the asset from Recycle Bin -> NPE before this fix; restores successfully after.

## Test plan
- [x] `core` module compiles (`mvnw compile`)
- [x] `core` module test sources compile (`mvnw test-compile`) — updated two tests (`PermissionMatrixOrgLifecycleTest`, `OrgLifecycleScopedPropertiesIntegrationTest`) that construct `UserTreeService` positionally, for the new constructor parameter
- [ ] Manual repro steps above verified in a running instance

### DIFF
--- a/core/src/main/java/inetsoft/web/RecycleBin.java
+++ b/core/src/main/java/inetsoft/web/RecycleBin.java
@@ -137,6 +137,32 @@ public class RecycleBin implements XMLSerializable, AutoCloseable {
    }
 
    /**
+    * Updates the owner recorded on recycle bin entries (in the current org) after a user is
+    * renamed, so that entries deleted by the old username can still be located and restored.
+    */
+   public synchronized void renameUser(IdentityID oldUser, IdentityID newUser) {
+      Map<String, Entry> map = new HashMap<>();
+      KeyValueStorage<Entry> storage = getStorage();
+      storage.stream().forEach(p -> map.put(p.getKey(), p.getValue()));
+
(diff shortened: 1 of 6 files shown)


jshobe-inetsoft (approved):

PR #4491, opened 2026-08-03
Bug #75858, give AssetQueryScope a parent-chain fallback for qualified viewsheet-assembly reads

ngovus-inetsoft:
## Summary

- A qualified read of a viewsheet assembly from a worksheet/asset-query script -- e.g. `worksheet['Crosstab2']` -- resolves to `undefined` under GraalJS, even though the scope chain that owns the name is correctly linked.
- `ScopeProxy` dispatches a qualified read straight at the wrapped scope and never reaches `BindingRootProxy`'s central parent-chain walk (that walk only covers unqualified names). `AssetQueryScope.getMember`/`hasMember` only consulted `getTableValue()` (worksheet tables) and the local `members` map, so the `ViewsheetScope` chained on by `AssetQuerySandbox.createAssetQueryScope()` (added by Bug #75526 so viewsheet assemblies are visible in worksheet scripts) was never checked. Because `hasMember()` gates `getMember()` under GraalJS, the read silently produced `undefined` with no error or log line.
- This is the mirror-image gap, on the other side of the viewsheet<->worksheet scope link, from Bug #75807 (`ViewsheetScope`'s qualified read of a worksheet table, PR #4455). `AssetQueryScope.getMember`/`hasMember` now fall back to a `findInChain()` helper that walks `getParentScope()`, following the same pattern (including the `scope != this` cycle guard). `getMemberKeys()` is left alone, since Rhino's `getIds()` likewise did not enumerate prototype ids.
- Not observed in a customer asset -- found by code inspection and a direct scope-level check while fixing #75807.

## Test plan

- [x] Added `AssetQueryScopeTest#testViewsheetAssemblyResolvedThroughChain`, mirroring `ViewsheetScopeTest#testWorksheetTableResolvedThroughChain`: confirms a viewsheet assembly name is unresolvable through a bare `AssetQueryScope`, then resolvable once a `ViewsheetScope` is chained on via `JavaScriptEngine.addToPrototype()` (the same wiring `AssetQuerySandbox.createAssetQueryScope()` performs in production), and confirms an unrelated name stays absent.
- [x] Verified the new test fails without the fix and passes with it.
- [x] `./mvnw -pl core test -Dtest=AssetQueryScopeTest` passes.

### DIFF
--- a/core/src/main/java/inetsoft/report/script/formula/AssetQueryScope.java
+++ b/core/src/main/java/inetsoft/report/script/formula/AssetQueryScope.java
@@ -112,9 +112,15 @@ public class AssetQueryScope implements DynamicScope, Cloneable {
          LOG.error("Failed to get property from asset query: " + id, ex);
       }
 
+      if(members.containsKey(id)) {
+         return members.get(id);
+      }
+
       // the dynamic scope fallback (executing scope) is now provided
       // centrally by BindingRootProxy
-      return members.get(id);
+      ScriptScope owner = findInChain(id);
+
+      return owner == null ? null : owner.getMember(id);
    }
 
    @Override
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4492, opened 2026-08-03
feat(admin-chat): server-side foundation for the EM AI agent plugin (#75861)

jshobe-inetsoft:
Redmine [#75861](https://redmine/issues/75861) — server-side foundation for the EM AI agent plugin.

## What this adds

Server-property administration driven by an AI agent, with the safety machinery that makes that acceptable: a review gate that the server enforces, a snapshot before side-effectful changes, all-or-nothing application with compensating rollback, and a task-correlated audit trail.

All endpoints live under `/api/wiz/v1/admin/`. That prefix is load-bearing, not cosmetic: `WizServiceAuthenticationFilter` validates a bearer JWT only there, and `CSRFFilter` exempts only there.

| Endpoint | Purpose |
|---|---|
| `GET /properties?filter=` | catalog metadata + each property's current value |
| `GET /properties/{name}` | one property; works for uncatalogued names too |
| `POST /preview` | resolves a change list into a plan, returns a SHA-256 `planHash`; mutates nothing |
| `POST /apply` | must echo the hash; re-resolves, recomputes, refuses on mismatch with **409** |
| `POST /backup` | Tier-2 snapshot via the existing EM backup path |

The frozen contract, with every error case, is at `docs/superpowers/plans/2026-07-08-admin-chat-endpoint-contract.md` in the `stylebi-wiz` repo.

### The review gate

`preview` hashes the resolved plan **including current values**. `apply` re-resolves from its own request body, recomputes, and refuses on mismatch — returning the freshly resolved plan so a caller can show the operator what moved. Verified live: replaying a stale hash after the value changed returns 409.

Be precise about what this guarantees. It is **drift detection**, not proof a human reviewed anything: a caller can preview, echo its own hash, and apply. What it does prevent is an operator approving one plan while a different one executes.

### Two endpoints deliberately removed

- **`POST /change`** — applied a single property with no hash, bypassing the gate entirely. `AdminChangeService.applyChange` survives as the internal primitive.
- **`POST /restore`** — restoring storage on a running cluster is untested and requires a restart. `StorageService.importContents` has exactly one caller in the product: the offline setup path. Recovery is manual and offline, or by reverting the individual changes from the audit records.

## Changes to existing code, and why

**`AdminChangeService` read/write scope.** `SreeEnv.getProperty(name)` is org-scoped; `setProperty`/`remove` write the literal key. Mixing them meant that for a property with an org override, the code read the override, wrote the global key, re-read the unchanged override, reported `failed`, and then "rolled back" by writing the override's value into the global key — corrupting the global default. All three reads now use `getProperty(name, false, false)`. Side-effect hooks also now receive the base name; they match exact literals, so an org-qualified name silently never fired.

**`DataSpaceSettingsService.doBackup`** now returns `BackupResult(status, path)` rather than a status string, because the caller must record the artifact's path in the audit trail. `BackupDataModel` gains `aiSnapshot`, which routes the artifact to a sibling `ai-snapshots/` folder — that is also what exempts it from `deleteRedundantBackupFiles`, which lists only `backup/`. The prune call is skipped for AI snapshots: without that guard an AI snapshot would trim the operator's `backup/` folder and write nothing back to it.

Existing callers are unaffected: the static `backup(...)` wrapper still returns `String`, so `AssetFileBackupAction` is untouched.

## Verification

**Tests: 5118 run, 0 failures** across `community/core` (baseline before this branch: 5104). 165 in the admin/properties/security area.

**Live smoke test** against a running enterprise docker-test server, token minted through the real SSO flow — unauthenticated 403s; catalog read with alias resolution; preview; apply (`before "0"` → `after "500"`, verified); audit read-back carrying `reviewOutcome`; and the 409 on a replayed stale hash. Recorded in full in the spec.

**Not covered by the live test** — please do not read the green suite as covering these:
- `POST /backup` and the storage-snapshot path end to end. The only storage-scoped catalogued property is `security.exposedefaultorgtoall`, which fires repository-wide events on a live server, so it was deliberately not exercised. Mocked at the `doBackup` seam only.
- Rollback and `rollback-failed`. Reaching them needs a mid-sequence failure that cannot be induced through the public API.
- **All cluster behaviour.** Single-node deployment. Cross-node property propagation, the ~500ms `PropertiesEngine` debounce, and cluster-visible snapshot retrieval are unverified in practice.

## Known follow-ups

Found during review, deliberately not fixed here. Each is real; none blocks the feature working as documented.

1. **`apply` is serialized by a JVM-local lock only.** A cluster peer, or a concurrent edit from the EM properties page, is not serialized against it. `SreeEnv.save()` flushes `changedProps` for *all* keys, so a failing apply can commit another writer's staged writes — which is why the lock exists at all. A distributed lock was out of scope.
2. **`isSecret` has false positives.** `enable.changepassword`, `sso.rsa.public.key` (a *public* key) and arguably `google.maps.key` are refused though they are not secrets. The refusal is a loud 400 naming the property, so it is discoverable; the predicate wants narrowing.
3. **`ai-snapshots/` retention is a count, not a policy.** Pruned to `ai.snapshot.count` (default 10). Note `AdminRiskClassifier` treats every *uncatalogued* property as storage-scoped, so a full dataspace export is the common case rather than the rare one — worth revisiting with an age-based policy.
4. **`AdminPropertyCatalog.resolve` would rewrite a case-preserving name** (`log.level.*`, `plugin.extra.classpath.*`, and two others) to the catalog's canonical spelling, which for those families addresses a *different* `SreeEnv` key. Latent: no such entry is catalogued today.
5. **`deleteRedundantBackupFiles` has two pre-existing defects** this branch deliberately did not touch, since they affect real operator backups and changing them has release-note implications: the loop is `i <= deleteCount`, so it keeps one fewer backup than `asset.backup.count` implies; and its comparator casts a `long` timestamp difference to `int`, which can invert ordering. The new AI-snapshot prune uses `Long.compare` and keeps exactly N. **These two deserve their own issue.**
6. **`LocalContext` scans admin-chat beans into the storage initializer.** It component-scans `inetsoft.web.admin` and excludes `@Controller`/`@RestController` but not `@Service`/`@Component`, so all six new services land in a context that has no business owning them. `lazyInit = true` makes it latent, but this already caused one storage-initializer container crash during development. An exclude filter for `inetsoft.web.admin.ai` would settle it. `AdminAiBeanGraphTest` guards the *web* context's graph, not this one.

## Reviewing this efficiently

The safety-critical logic is concentrated in three files: `AdminChangePlanService` (the hash), `AdminChangesetApplyService` (apply/rollback), and `AdminChangeService` (the per-property primitive). `AdminPropertyCatalog`'s class javadoc carries the rule for classifying a new property's `snapshotScope` — worth reading before adding a catalog entry.


jshobe-inetsoft:
## Review response

### Finding: pruning failure masking a successful backup — **fixed** in 83efb6d

The try-block scoping is wrong and the downstream chain you traced is exactly right: `path == null` → `AdminBackupService.backup()` throws `IOException` → `AdminChangesetApplyService.apply()` aborts the changeset as "nothing was applied", while a good snapshot sits in `ai-snapshots/` with no audit `backupRef` pointing at it. The prune now has its own log-and-continue catch.

One thing worth adding, which the review didn't mention and which makes it slightly worse than described: the outer catch also skips `record.setActionStatus(ACTION_STATUS_SUCCESS)`, so the operator gets a **FAILURE audit record for a backup that happened**. The same fix corrects that.

**A correction on the mechanism, though.** The specific trigger named — `externalStorageService.listFiles(AI_SNAPSHOT_FOLDER)` throwing a transient storage error — is not reachable through any shipped implementation. All four already swallow their own failures and return an empty list:

| Implementation | `listFiles` behaviour on error |
|---|---|
| `FilesystemExternalStorageService` | returns `emptyList()` if the dir is absent or `File.listFiles()` returns null |
| `S3ExternalStorageService` | `catch(Exception e) { LOG.error(...) }` → `emptyList()` |
| `AzureExternalStorageService` | same |
| `GCSExternalStorageService` | same |

So the fix is not closing a live path through `listFiles`. It is still correct, for two reasons that don't depend on that mechanism: `listFiles` declares **no checked exception**, so the interface permits an implementation to throw (and a `null` return would NPE on `.stream()`); and more fundamentally, the try block's scope is wrong regardless of *which* call throws — post-write housekeeping must never be able to discard a durable artifact. That's why the guard catches `Exception` rather than `IOException`.

`deleteRedundantBackupFiles` is deliberately left unguarded: it runs *before* the write, where a failure legitimately fails the whole operation.

The test-coverage gap you identified was the real one. Two tests added:
- write succeeds, `listFiles("ai-snapshots")` throws → `result.path()` non-null, status not a failure. **Verified to fail against the unguarded code** (`expected: not <null>`), so it discriminates.
- the counterpart: a failure in the write itself still returns a null path — otherwise the first test would also pass against a `doBackup` that swallowed everything.

`community/core`: 5121 tests, 0 failures.

### Minor: `requireSiteAdmin` / `handleIllegalArgument` duplication — **not changing**, and not purely cosmetic to fix

Declining this one, on the risk of the fix rather than the size of the duplication.

`handleIllegalArgument` is an `@ExceptionHandler`, so where it lives determines its **scope**. Hoisting it to a `@ControllerAdvice` would map `IllegalArgumentException` → 400 for *every controller in the application*, which is a behaviour change well outside this PR. An abstract base class does avoid that (Spring inherits `@ExceptionHandler` from superclasses), but it buys removing ~6 lines at the cost of an inheritance relationship between two controllers whose handler sets already differ — `AdminAiController` also carries `handlePlanHashMismatch`, which `AdminPropertiesController` must not have.

The javadoc cross-reference is there precisely because the duplication is intentional. Happy to revisit if a third such controller appears; at two, the abstraction costs more than it saves.

Separately, and consistent with the PR description: the two pre-existing defects in `deleteRedundantBackupFiles` (the `i <= deleteCount` off-by-one and the `(int)` cast in its comparator) remain untouched here. They affect real operator backups and have release-note implications, so they want their own issue rather than a quiet fix on a feature branch.

jkidacki-inetsoft (approved):

PR #4497, opened 2026-08-04
feat(wiz): support copy-then-apply on the date-comparison endpoint

fredaan:
## Why

`/viewsheet/date-comparison` applied the comparison to the target chart **in place**, with no way to ask for a duplicate first — unlike every sibling wiz mutation (`applyHighlight`, `setChartFormat`, `setChartColors`, and `createViewsheetInternal`'s modificationOnly path, which all take a `copy` flag).

That made a stand-alone *"add a year-over-year comparison"* rewrite the chart the user already had, which is exactly what copy-then-apply exists to prevent.

## What

- `ApplyDateComparisonModel.copy` — same contract as `ApplyHighlightModel.copy`.
- `WizVsService.applyDateComparison` gains the duplicate-then-apply block and the rollback that goes with it: on any failure, remove the copy and restore the original's primary flag before re-throwing, so a thrown request never leaves the runtime holding an added/promoted copy the caller never learns about.

The caller (wiz-services) requests `copy=true` exactly when its turn has not yet established a chart of its own; after a create or a binding rebind it sends `false`, so no intermediate copy is orphaned.

## Two details worth a reviewer's attention

**The type guard throws before the `try`.** `"Assembly does not support date comparison"` is raised before the rollback-protected block, so it undoes the duplication itself. Without that, a copy made for an assembly that turns out not to be `DateCompareAble` would be orphaned.

**`vs.clearSharedFrames()` is deliberately not rolled back.** It is idempotent cache invalidation — clearing it only forces the shared colour frames to be recomputed — not state a rollback has to restore. This is called out in a comment so a later reader does not "fix" it.

## Verification

`./mvnw -o -pl community/core compile` passes. Not exercised at runtime yet.

## Related

Paired with the wiz-services change that starts sending `copy` on this endpoint. That side also fixes a latent bug this endpoint was never at fault for: wiz-services used to send `dateComparisonModel` to `/viewsheet/create`, where `CreateVisualizationModel` has no such field and is annotated `@JsonIgnoreProperties(ignoreUnknown = true)` — so the model was silently discarded and the call reported success having applied no comparison at all.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/model/ApplyDateComparisonModel.java
+++ b/core/src/main/java/inetsoft/web/wiz/model/ApplyDateComparisonModel.java
@@ -25,7 +25,7 @@ import inetsoft.web.composer.model.vs.DateComparisonDialogModel;
  * Request body for {@code POST /api/wiz/viewsheet/date-comparison}.
  *
  * <p>The wiz-services caller POSTs:
- * {@code { runtimeId, viewsheetIdentifier?, assemblyName?, sampleMaxRows?,
+ * {@code { runtimeId, viewsheetIdentifier?, assemblyName?, copy?, sampleMaxRows?,
  * dateComparisonModel: { dateComparisonPaneModel: <PaneModel> } } }.
  *
  * <p>{@code dateComparisonModel} is a {@link DateComparisonDialogModel}, which Jackson binds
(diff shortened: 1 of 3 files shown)


fredaan:
Both observations taken — thanks. One of them I'd push back on being out of scope.

## Reaped runtime → **fixed here rather than deferred**

You're right that the line predates this PR, but I don't think it should wait. This PR exists to make date comparison *actually apply* (it never did — the model was being sent to `/viewsheet/create`, which has no such field and discards it), and "except in a reopened conversation" is a real hole in that claim rather than a parity nicety. Reopening starts a new runtime for the same viewsheet, so the id the client still carries names a dead instance, and `applyDateComparison` threw where `applyHighlight` would have transparently restored.

Adopted `WizUtil.getViewsheetOrRestore(...)` plus the `restored → result.setRuntimeId(rvs.getID())` echo, matching `applyHighlight`. No client change needed: the caller already reads the returned `runtimeId` and adopts it as the conversation's live runtime, so this simply starts working where it previously threw.

## Failure messages naming a copy that no longer exists

Fixed, and in **both** `applyDateComparison` and `applyHighlight` — you're right that it's shared style, and fixing only the one I touched would leave two sibling methods reporting failures differently.

Worth noting why this is more than tidiness now: the wiz-services caller **classifies failures off these strings** to decide whether to replay the same payload, regenerate its parameters, or give up, and it feeds the same strings back into the model that regenerates the payload. A name that refers to nothing — the rollback has already removed the copy by the time the message propagates — is misleading input to both. My own type-guard rollback in this PR made that worse, since it's the thing that removes the named assembly before the message is read.

Captured `requestedAssemblyName` before the copy step reassigns `assembly`, and used it in the guard message and the copy-declined log line.

`./mvnw -o -pl community/core compile` passes.

Nelson-cao (approved):

PR #4498, opened 2026-08-04
fix(ws): stop double-flipping a join operator that concatenateTable already re-oriented

larryliang-inetsoft:
## The bug

A relational join over 3+ tables could emit SQL that qualified a join key with the wrong table:

```
ERROR: column projects.project_id does not exist
```

`project_id` is the FK on `work_packages`, not a column of `projects`. The join was declared correctly — the operator's attributes had been swapped out from under it. Whether it fired depended on the order the base tables happened to be declared in, which is why it looked nondeterministic.

## Root cause (commit 1 — the actual fix)

`editExistingJoinTable` snapshots `leftTables[]`/`rightTables[]` from the incoming operators, then (when `containsNewTable`) runs `concatenateTable` per operator. **`concatenateTable` re-orients an operator in place** to match the assembly it builds, so afterwards the snapshot can name the two tables in the opposite order to the operator's own attributes. `addOperator` compares those stale names against the subtable order, concludes it must flip, and flips an operator that was **already correct** — storing it under the right key with left/right attributes swapped.

Downstream, `JoinQuery` walks base tables positionally and resolves `getLeftAttribute()` against `tables[i]`. With attributes reversed it looks `project_id` up in `projects`, misses, and — because `AssetUtil.findColumn` returns the ref **unchanged** rather than `null` on a miss, so the `if(lcolumn == null) continue` guard never fires — qualifies it as `projects.project_id`.

The post-processing join path shares the cause with a quieter symptom: it checks its index properly, logs `Join column not found in left-hand table`, drops the condition and **silently cross-joins**.

Fix: take the table names off the operator itself, which `concatenateTable` keeps in sync with the attributes; fall back to the snapshot only for an operator that never carried names.

## Commit 2 — hardening (not the fix)

`getOperator(l, r)` falls back to the `(r, l)` key so a symmetric join is findable from either side, but returned the operator as-is, with attributes still belonging to the opposite tables. It now returns a reversed copy. Only symmetric operations reach that path (the existing loop already refuses LEFT/RIGHT/GREATER/LESS/MINUS). Stated plainly: this did **not** fix the failure above — the operator was stored under the *forward* key already mis-oriented, so the reversed path never fired. It closes the same class of defect one layer down.

## Verification

- **Diagnosed by instrumentation**, not inference — a temporary trace in `addOperator` showed the operator arriving already oriented `PROJECTS`-left and leaving `fkjoin`-left after the spurious flip. Two earlier hypotheses (a missing mirror column; per-pair operator binding) were checked and discarded first.
- **End-to-end**: a 3-table openproject join declared `[PROJECTS, WORK_PACKAGES, STATUSES]` failed before this change and returns the correct 13 rows after, with the client-side workaround disabled so the fix is what's being tested.
- 4 new unit tests on `CompositeTableAssemblyInfo` (2 fail against unpatched code with `expected: <id> but was: <project_id>`; 2 are guards that pass both ways).
- `inetsoft.uql.asset.**` + `inetsoft.report.composition.**` + `inetsoft.web.composer.ws.**`: 234 tests, 0 failures.

The `InnerJoinService` change is covered by the end-to-end run rather than a unit test — exercising it needs a live `Worksheet` through `concatenateTable`, and a mock-heavy test there would assert the scaffolding rather than the behaviour. Happy to add one if you'd prefer it.

## Related

`stylebi-wiz` PR #1421 normalizes `baseTables` order client-side so callers stop hitting this. That workaround stays useful for older engines, but this is the real fix.

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/internal/CompositeTableAssemblyInfo.java
+++ b/core/src/main/java/inetsoft/uql/asset/internal/CompositeTableAssemblyInfo.java
@@ -115,10 +115,50 @@ public class CompositeTableAssemblyInfo extends ComposedTableAssemblyInfo {
                   break;
                }
             }
+
+            // The operator was stored for (rtable, ltable) but the caller asked about
+            // (ltable, rtable). A <=> B is symmetric as a JOIN, but its OPERANDS are not
+            // interchangeable -- so hand back a reversed view rather than one whose
+            // left/right attributes belong to the opposite tables. Callers pair the
+            // attributes with the tables THEY named: JoinQuery walks base tables
+            // positionally (i < ri) and resolves getLeftAttribute() against tables[i], so a
+            // reverse-oriented operator made it qualify the key with the wrong table --
+            // emitting `column projects.project_id does not exist` when generating merged
+            // SQL, and silently dropping the condition into a cross join on the
+            // post-processing path. That made a perfectly well-formed worksheet fail (or
+            // quietly return wrong rows) purely because of the order its base tables were
+            // declared in.
+            if(operator != null) {
+               operator = reverse(operator);
+            }
          }
       }
 
       return operator;
+   }
+
+   /**
+    * Build a reversed view of an operator: left and right swapped on every child, so it reads
+    * correctly for the (ltable, rtable) order the caller asked about. Only ever applied to
+    * symmetric operations (the loop above rejects the asymmetric ones), for which swapping the
+    * operands preserves the meaning. Returns a copy -- the stored operator keeps its own
+    * orientation, and no caller mutates what getOperator hands back.
+    */
+   private static TableAssemblyOperator reverse(TableAssemblyOperator operator) {
+      TableAssemblyOperator reversed = new TableAssemblyOperator();
+
+      for(int i = 0; i < operator.getOperatorCount(); i++) {
+         TableAssemblyOperator.Operator op = operator.getOperator(i);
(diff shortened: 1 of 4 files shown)


larryliang-inetsoft:
Added the `InnerJoinService` unit test (d5a0a367a) — the gap I flagged in the description is now closed.

`InnerJoinServiceOperatorOrientationTest` drives `editExistingJoinTable` over a real `Worksheet` (three `EmbeddedTableAssembly`s + a `RelationalJoinTableAssembly`, no mocks), so `concatenateTable` → `exchange` → `addOperator` all run for real. It turned out not to need the heavy scaffolding I assumed.

It asserts every stored operator agrees with its own key — both the table names **and** the tables its left/right attributes belong to. Checking the attributes is the whole point: the bug stored the operator under the *correct* key with its columns swapped, so a key-only assertion would have passed while `JoinQuery` still emitted `column projects.project_id does not exist`.

Verified it actually catches the regression by reverting only `InnerJoinService.java` to its pre-fix content:

```
Tests run: 2, Failures: 1
operator 0 stored under (PROJECTS,WORK_PACKAGES): leftTable
  ==> expected: <PROJECTS> but was: <WORK_PACKAGES>
```

The second test pins the declaration order that always worked, so a future "fix" that flips unconditionally — moving the breakage to the other order rather than removing it — fails too. It passes both before and after, so it's a guard rather than a second copy of the regression.

Suite is now 236 tests across `inetsoft.uql.asset.**` / `inetsoft.report.composition.**` / `inetsoft.web.composer.ws.**`, 0 failures.

fredaan (approved):

PR #4501, opened 2026-08-04
Bug #75821, don't show a misleading permission error when EM help-links fails due to session expiry

jkidacki-inetsoft:
## Summary

Logging into EM in one tab, logging into Portal with the same account in another tab (same shared session cookie), then logging out of Portal, can invalidate the shared session while the EM tab's initial page load is still in flight (reproducible by throttling the EM tab to a slow connection). The in-flight `GET em/help-links` request then legitimately gets a 401/403 because the session really is gone.

That failure is already handled correctly by `InvalidSessionInterceptor`, which redirects the user to the login page for any 401/403. `NavbarComponent`, however, also independently popped a "You may need to give this user permission to Enterprise Manager" toast for the exact same failure — which is misleading, since it isn't a permission problem, just a session that expired mid-load.

## Root Cause

- `NavbarComponent.handleHelpLinkError()` showed the same permission-themed error toast for every kind of `getHelpLinks()` failure, without distinguishing a session-expiry 401/403 (already handled elsewhere via redirect-to-login) from a genuine unexpected failure.
- The same method also unconditionally rethrew the error to a `.subscribe()` call with no error callback, which surfaces as an unhandled exception in the browser console on every help-links failure, independent of the toast issue.

## Fix

- Only show the help-links error toast when the failure is *not* a 401/403 — those are already handled by the app-wide session-expiry redirect.
- Recover (`EMPTY`) instead of rethrowing from `handleHelpLinkError`, since the error is already fully handled (logged + conditionally toasted) at that point and rethrowing to a subscriber with no error handler served no purpose beyond an unhandled-exception log.

## Test Plan

- [x] Added `navbar.component.tl.spec.ts` cases: 401 and 403 suppress the toast, a 500 still shows it — all assert against a real MSW-mocked HTTP response.
- [x] `npm run test:em` (356/356 passed) and `npm run test:em:tl` (818 passed, 19 pre-existing expected-fail) both green.
- [x] `npm run lint` clean (no new warnings).
- [x] `npm run build` succeeds for all Angular projects.
- [x] Manually reproduced the original bug (EM tab throttled to 3G, logout of Portal mid-load) and confirmed the toast no longer appears with this fix.

Fixes Redmine #75821.

### DIFF
--- a/web/projects/em/src/app/navbar/navbar.component.tl.spec.ts
+++ b/web/projects/em/src/app/navbar/navbar.component.tl.spec.ts
@@ -90,7 +90,10 @@ const ORG_ADMIN_ITEM_VISIBILITY: Array<[name: string, role: string, expectedVisi
    ["_#(Monitoring)", "link", true],
 ];
 
-async function renderComponent(permissions: ComponentPermissions) {
+async function renderComponent(
+   permissions: ComponentPermissions,
+   snackBar: { open: ReturnType<typeof vi.fn> } = { open: vi.fn() }
+) {
    server.use(
       http.get("*/api/em/navbar/get-navbar-model", () => HttpResponse.json(NAVBAR_MODEL))
    );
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4502, opened 2026-08-04
Bug #75908, fix org copy leaving Data Cycle task references unresolved in Run After

ngovus-inetsoft:
## Summary
- Copying an organization left any task's "Run After" completion condition pointing at a Data Cycle task blank in the new org, because `MigrateUtil.getNewOrgTaskName()` split the referenced task id on the first `":"` to find the owner/task-name boundary — but Data Cycle task ids are `"<owner>__<name>"`, and the name itself (`"DataCycle Task: Cycle1"`) contains its own colon, so the split landed inside the task name and the owner-org rewrite was silently skipped, leaving the condition pointing at the old org.
- Also hardens `ScheduleTask.updateConditionTaskPath()` (used on the site-admin asset-import path), which had the same delimiter assumption.

## Test plan
- [x] Added `MigrateUtilTest` covering the cycle-task rewrite (the actual regression), a normal-task rewrite, same-org no-op, no-delimiter no-op, and org-mismatch no-op.
- [x] Added two tests to `ScheduleTaskTest` covering `parseXML(elem, isSiteAdminImport=true)` for both cycle-task and normal-task completion conditions.
- [x] `mvn test -pl core -Dtest=MigrateUtilTest,ScheduleTaskTest` — all pass.
- [x] Manually reproduced: create a Data Cycle + task with "Run After" set to the cycle task, copy the org, confirm "Run After" now resolves to the cycle task under the new org instead of showing blank.

### DIFF
--- a/core/src/main/java/inetsoft/sree/schedule/ScheduleTask.java
+++ b/core/src/main/java/inetsoft/sree/schedule/ScheduleTask.java
@@ -1499,12 +1499,22 @@ public class ScheduleTask implements Serializable, Cloneable, XMLSerializable {
       String currOrgID = OrganizationManager.getInstance().getCurrentOrgID();
       String path = condition.getTaskName();
 
-      if(path == null || path.indexOf(":") < 0) {
+      if(path == null) {
          return;
       }
 
-      String pathUserID = path.substring(0, path.indexOf(":"));
-      String remaining = path.substring(path.indexOf(":"));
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):
Approving based on the completed Claude automated review — root-cause analysis confirmed, fix is consistent across both call sites, and new tests accurately reflect production id-building logic. The review's notes (possible related gap in `BatchAction.parseXML`, minor code duplication) are flagged as non-blocking/follow-up items. All checks pass and there are no merge conflicts.

PR #4503, opened 2026-08-04
Fix chart point annotation cannot be removed without selecting chart first

ngovus-inetsoft:
## Summary
- Chart point/data annotations are selected via a separate overlay (`chart-annotation-overlay` in `vs-object-container.component.html`) that never adds the chart's own index to `selectedAssemblies`.
- `RemoveAnnotationEvent.create()` only inspected `vsObjects` already present in `selectedAssemblies`, so removing a chart annotation without first clicking the chart itself silently returned `null` and `removeAnnotations()` no-op'd — no request sent, no error shown.
- Crosstab/table cell annotations were unaffected because they're selected through the table's own normal selection flow, which always keeps the table in `selectedAssemblies`.
- `create()` now also scans every `vsObject` for a non-empty `selectedAnnotations`, regardless of `selectedAssemblies`. `viewer-app.component.ts#removeAnnotations()` clears `selectedAnnotations` on every object that had any, matching the broadened lookup.

## Test plan
- [x] `ng run portal:test-tl --include='**/vs-object-container.component.interaction.tl.spec.ts'` — 17/17 passing
- [x] Manually reproduced: added chart point annotation, right-clicked it without first selecting the chart, clicked Remove — previously silently failed (no network request), now removes correctly
- [x] Verified crosstab annotation removal still works

### DIFF
--- a/web/projects/portal/src/app/vsobjects/event/annotation/remove-annotation-event.ts
+++ b/web/projects/portal/src/app/vsobjects/event/annotation/remove-annotation-event.ts
@@ -36,26 +36,36 @@ export class RemoveAnnotationEvent implements ViewsheetEvent {
    public static create(vsObjects: VSObjectModel[],
                         selectedAssemblies: number[]): RemoveAnnotationEvent
    {
-      if(selectedAssemblies) {
-         const names = selectedAssemblies
-            .map(index => vsObjects[index])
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4506, opened 2026-08-05
Keep the pre-aggregated date level at NONE

fredaan:
Reported against the wiz chat flow: "按月份分析去年订单数量" produced a chart bound to `Year(Month(ORDER_DATE))` — a second, year-level grouping stacked on a column the worksheet had already reduced to months, collapsing a 12-point monthly series.

## The date level

Applying an aggregate condition pushes the chart's grouping down onto the worksheet table and rebinds the chart to the bucketed columns, so the dimension must stop grouping — the worksheet already did it. That reset called `setDateLevel` only, which assigns the transient `dlevel` override that `VSDimensionRef.writeContents` never serializes, while the design-time value was cleared to `null`. A date-typed dimension with no level value falls back to the YEAR default (`WizardRecommenderUtil.setDefaultDateLevel`), so the level came back the moment the refs were rebuilt.

Now written as the level VALUE, the way `WizAutoBindingService.applyDateGroup` already does for the same symptom on the create path — its javadoc describes this exact failure ("a column that the worksheet had already bucketed (e.g. `Month(date_entered)`) was re-grouped by year at the chart layer").

Shared by the chart and crosstab rebind paths via one helper, since both had the same two lines.

## The condition operand

`resolveConditionType` had a related blind spot: it ignored `dateGroupLevel` and coerced values to the raw column's type. A part level (Month of Year, Day of Week, …) yields an integer ordinal, so its operand became an unparseable `timeInstant` that `Condition.convertType` leaves as-is and the engine renders as the current timestamp — matching nothing, reporting nothing. The type is now resolved from the level via `DateRangeRef.getDataType`, which keeps a date type for interval levels and gives integer for part levels. An unsupported level string logs and falls back to the column type rather than throwing.

## Testing

`./mvnw -pl community/core compile` clean. No end-to-end run — that needs a redeploy and a re-run of the reported chat steps.

## Companion change

The wiz-side half of the same report (a fabricated `Month(ORDER_DATE) = 1` condition, and the missing "last year" filter) is `inetsoft-technology/stylebi-wiz#1441`.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
@@ -4212,10 +4212,17 @@ public class WizVsService {
          condition.setEqual(spec.getEqual());
       }
 
+      // Resolve the column this condition actually targets BEFORE resolving its type: the field name
+      // and the operand's type must come from the same answer. When the worksheet has bucketed the
+      // column, both follow the bucketed column; when it has not, both stay on the raw one. An
+      // aggregate (HAVING) condition is built from the raw column below, so it never buckets.
+      DateGroupedField grouped = spec.getAggregateFormula() != null
+         ? null : resolveDateRangeField(spec, dimColumnMapping);
+
       // Set the condition's data type before adding values so each value is coerced to the type
       // the field expects (e.g. "100" -> 100.0, date strings -> Date). Without this the values
       // stay as raw strings and numeric/date comparisons fail.
-      String dataType = resolveConditionType(spec, columns);
+      String dataType = resolveConditionType(spec, columns, grouped);
 
       if(dataType != null) {
          condition.setType(dataType);
@@ -4250,22 +4257,63 @@ public class WizVsService {
          return new ConditionItem(aggregateRef, condition, level);
       }
 
-      // Base (WHERE) condition: use AttributeRef, resolving dateGroupLevel to DateRangeRef name if applicable.
-      String fieldName = spec.getField();
-
-      if(spec.getDateGroupLevel() != null) {
-         int dateLevel = getDateGroupLevel(spec.getDateGroupLevel());
-
-         if(dimColumnMapping != null && dateLevel != XConstants.NONE_DATE_GROUP) {
-            String dateRangeName = DateRangeRef.getName(spec.getField(), dateLevel);
-
-            if(dimColumnMapping.contains(dateRangeName)) {
-               fieldName = dateRangeName;
-            }
-         }
-      }
+      // Base (WHERE) condition: use AttributeRef, on the bucketed column when the worksheet has one.
+      String fieldName = grouped != null ? grouped.name() : spec.getField();
 
       return new ConditionItem(new AttributeRef(null, fieldName), condition, level);
+   }
+
+   /**
+    * The bucketed column a {@code dateGroupLevel} condition resolves to: the DateRangeRef column
+    * name the condition targets, and the level that produced it.
+    *
+    * @param name      the DateRangeRef column name, e.g. {@code MonthOfYear(ORDER_DATE)}
+    * @param dateLevel the level that produced it, never {@link XConstants#NONE_DATE_GROUP}
+    */
+   private record DateGroupedField(String name, int dateLevel) {}
+
+   /**
+    * Resolve the bucketed column a {@code dateGroupLevel} condition targets, or null when the
+    * condition stays on the raw column.
+    *
+    * <p>A level only takes effect when the pre-aggregation pass actually pushed the matching
+    * DateRangeRef column onto the worksheet; {@code dimColumnMapping} is that set, and it is null on
+    * the paths that never pre-aggregate (highlights and {@link #applyConditionModel}, both via
+    * {@link #buildConditionList}). Callers must key BOTH the field name and the operand's data type
+    * off this one answer — retyping without remapping puts the INTEGER ordinal of a PART level onto
+    * a raw date column, which a pushed-down pre-condition renders as {@code ORDER_DATE = 3}.
+    *
+    * <p>An unsupported level string is logged and ignored rather than thrown, matching
+    * {@link WizAutoBindingService#applyDateGroup}. This is the only place the level is parsed, so
+    * that leniency holds for the whole condition rather than for the type alone.
+    */
+   private DateGroupedField resolveDateRangeField(VisualizationConditionModel.ConditionSpec spec,
+                                                  Set<String> dimColumnMapping)
+   {
+      if(spec.getDateGroupLevel() == null || dimColumnMapping == null) {
+         return null;
+      }
+
+      int dateLevel;
+
+      try {
+         dateLevel = getDateGroupLevel(spec.getDateGroupLevel());
+      }
+      catch(IllegalArgumentException e) {
+         LOG.warn("Ignoring unsupported dateGroupLevel '{}' on a condition for field '{}'; " +
+                     "the condition stays on the raw column",
+                  spec.getDateGroupLevel(), spec.getField());
+         return null;
+      }
+
+      if(dateLevel == XConstants.NONE_DATE_GROUP) {
+         return null;
+      }
+
+      String dateRangeName = DateRangeRef.getName(spec.getField(), dateLevel);
+
+      return dimColumnMapping.contains(dateRangeName)
+         ? new DateGroupedField(dateRangeName, dateLevel) : null;
    }
 
    /**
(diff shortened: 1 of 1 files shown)


fredaan:
## Review — Keep the pre-aggregated date level at NONE

Both halves of the change are correctly diagnosed and the fix mechanism is right. I verified the two claims the description rests on:

* `VSDimensionRef.setDateLevel` only assigns the transient `dlevel` field (`VSDimensionRef.java:554`), and `writeContents` serializes `dlevelValue`/`dlevelRValue` only (`VSDimensionRef.java:1182-1191`) — so the old `setDateLevel(NONE)` really was lost on persist, and `WizardRecommenderUtil.setDefaultDateLevel:337-349` really does re-apply `YEAR_INTERVAL` to a date-typed dim with no level value. The write-the-value fix is correct, and the ordering inside the helper is right (`setDateLevelValue` nulls `dlevel` at `VSDimensionRef.java:571-577`, then `setDateLevel` re-establishes it).
* `DateRangeRef.getDataType(option, originalType)` (`DateRangeRef.java:217-238`) returns `INTEGER` for any level with the `PART_DATE_GROUP` (512) bit and a date type otherwise, so it is the right source of truth for the condition operand.

`./mvnw -pl core test -Dtest='WorksheetTableServiceConditionTest,WizAutoBindingServiceDateGroupTest,WizConditionModelRoundTripTest,WizDashboardFilterBuilderTest'` — 48 tests, 0 failures, locally on this branch.

Requesting changes for the reasons listed at the bottom. The substantive code item is (1).

---

### 1. `resolveConditionType` retypes unconditionally, but `buildConditionItem` remaps the field conditionally

`buildConditionItem` only swaps the condition's field to the bucketed column when the mapping actually contains it (`WizVsService.java:4256-4266`):

```java
if(dimColumnMapping != null && dateLevel != XConstants.NONE_DATE_GROUP) {
   String dateRangeName = DateRangeRef.getName(spec.getField(), dateLevel);

   if(dimColumnMapping.contains(dateRangeName)) {
      fieldName = dateRangeName;   // <- only here does the condition target Month(ORDER_DATE)
   }
}
```

The new type resolution (`WizVsService.java:4307-4321`) has no such guard — it retypes from the level whenever `dateGroupLevel != null`, regardless of whether the remap happened. The two disagree on every path where `dimColumnMapping` is null, and `buildConditionList` passes exactly that (`WizVsService.java:4102-4113`):

```java
appendConditionNodes(wizModel.getBaseConditions(), 0, conditionList, new boolean[]{ true },
                     null, columns);
```

which is the path taken by `applyConditionToAssembly` → `setPreConditionList` (`WizVsService.java:4095-4096`) and by `buildHighlight` (`WizVsService.java:791`).

Concretely: a spec of `{field: "ORDER_DATE", dateGroupLevel: "month of year", value: 3}` arriving through `applyConditionToAssembly` now produces a **pre-condition on the raw `ORDER_DATE` column with an INTEGER operand `3`**. The old behavior was also broken (the unparseable timeInstant the description describes), so this is not a regression in correctness — but it changes the failure *mode* for a condition that is pushed down to the datasource: `ORDER_DATE = 3` is a type error at the SQL layer on most dialects, where the old form was valid SQL returning zero rows. Trading a silent no-match for a query failure is a worse outcome for the highlight/assembly path.

Suggestion: resolve the effective field name **before** the type, and only apply the level-derived type when the remap actually took effect — e.g. hoist the `fieldName` computation above the `resolveConditionType` call and pass the "was remapped" flag (or `dimColumnMapping`) in. That keeps the two decisions from drifting.

Separate nuance on the same call, worth a line in the description: for a base column typed `XSchema.DATE`, `DateRangeRef.getDataType(MONTH_DATE_GROUP, "date")` returns `TIME_INSTANT`, not `DATE` — so an interval level widens a DATE operand even on the path where the remap *does* happen. That matches what `DateRangeRef` reports for the grouped column itself, so it looks intended, but it is a behavior change beyond the part-level case the description covers.

### 2. The new `catch(IllegalArgumentException)` is only half-applied

`getDateGroupLevel` throws on an unrecognized string (`WizDateLevelUtil.java:57`). The new catch in `resolveConditionType` logs *"coercing values to the raw column type instead"* and falls through — but the caller then calls `getDateGroupLevel(spec.getDateGroupLevel())` again, unguarded, at `WizVsService.java:4257`, for every non-aggregate condition. Same input, same throw, a few lines later. So for base (WHERE) conditions the lenient path is unreachable and the log line is misleading; only the aggregate branch (which returns at `:4250`) actually benefits.

Either guard `:4257` too, or drop the try/catch and let it throw consistently.

### 3. `setDateLevelValue("0")` is visible to consumers that read null as "unset"

The helper changes `getDateLevelValue()` from `null` to `"0"` for these refs. Two callers branch on exactly that:

* `ChartDndHandler.fixDimension` (`ChartDndHandler.java:1186`): `level = getDateLevelValue() == null ? -1 : parseInt(...)`, so `-1` becomes `0`. `DateRangeRef.isDateTime(-1)` is false (`-1 & PART_DATE_GROUP != 0` → INTEGER) while `isDateTime(0)` is true (default branch → TIME_INSTANT), so dragging one of these pre-aggregated dimensions in the composer now flips `setTimeSeries(...)` from false to true.
* `VSWizardBindingHandler.java:1524` copies the chart field's level only when the value is null — it now never copies for these refs.

Both are arguably *more* correct (`0` is literally NONE, and the non-VS branch of `fixDimension` already uses `getDateLevel()`, which returns 0). But it is a real downstream behavior change outside the wiz path, it is not mentioned in the description, and nothing pins it.

### 4. No tests

This area is already unit-tested by convention — `applyDateGroup` was deliberately made package-private static *"so it can be unit-tested directly with a mocked dimension ref"* (`WizAutoBindingService.java:860`), with `WizAutoBindingServiceDateGroupTest` covering it. The new helper is `private` and untested, and the new `resolveConditionType` branch is untested. Two cheap tests would pin the actual bug:

* after `clearDateLevelAfterPreAggregation`, `getDateLevelValue()` is `"0"` and `getDateLevel()` is `NONE_DATE_GROUP` — and, since the bug *is* the round-trip, ideally that it survives `writeContents` → `parseAttributes`. A test asserting only `getDateLevel() == 0` would have passed against the old code too, so the serialization round-trip is the assertion that matters.
* `resolveConditionType` returns `INTEGER` for a part level and a date type for an interval level.

Making the helper package-private would follow the existing pattern.

### 5. Minor: constant mixing in the new helper

```java
dimRef.setDateLevelValue(String.valueOf(XConstants.NONE_DATE_GROUP));
dimRef.setDateLevel(DateRangeRef.NONE);
```

`DateRangeRef.NONE` is inherited from `DataRef.NONE` (`DataRef.java:39`) — a *refType* constant that happens to be `0`. The date-level constant is `DateRangeRef.NONE_INTERVAL` (`DateRangeRef.java:133`, `= XConstants.NONE_DATE_GROUP`). Same value, no bug, and it is pre-existing text moved into the helper — but the helper is the natural place to make both lines name the same concept.

### Not blocking, worth knowing

Existing viewsheets already saved with the missing `dlevelValue` are not repaired by this change; they keep falling back to YEAR until the chart is rebuilt. If any were saved from the reported flow, they will need to be recreated.

No security concerns: no new user input reaches SQL outside the existing `Condition` type-coercion path, and no new logging of sensitive values.

---

## Reasons for requesting changes

* **Failing check — `Validation Build`** ([run 30982532397](https://github.com/inetsoft-technology/stylebi/actions/runs/30982532397)). To be fair to the author, none of the failures touch this diff: `MVDefWorksheetSerializationTest`, `XDynamicTableTest`, `MVQueryTest`, `VSChartShowDataServiceTest`, `VSFormTableServiceTest`, `BaseTableSortColumnServiceTest`, `VSTableModelTest`, `VSTabModelTest`, `RepositoryObjectServiceTest`, `BatchActionTest`, `ViewsheetActionTest`, `IndividualAssetBackupActionTest` all die with the same root cause — `NoClassDefFoundError: Could not initialize class inetsoft.report.internal.Util`, from an `ExceptionInInitializerError: ShutdownException: Spring application context is not available` raised on the `BlobStorageEvent` background thread during `Util.<clinit>`. That is a cross-test initialization race, not something a `WizVsService` change can cause. Still, the check is red and needs to go green (a re-run may be enough) before merge.
* **Failing check — `claude-review`** ([run 30982532428](https://github.com/inetsoft-technology/stylebi/actions/runs/30982532428)): `Reached maximum number of turns (50)`. Infrastructure, no code finding — but likewise red.
* **Code item (1)** above: the field-remap guard and the type-resolution guard disagree, which can push an INTEGER operand onto a raw date column on the `applyConditionToAssembly` / highlight paths.

Items (2)–(5) are not blockers on their own.

Nelson-cao (approved):

PR #4509, opened 2026-08-05
Forward the wiz chat's already-resolved measure formula through changeType

Zxm-happy:
## Summary

- `changeType()` re-renders the target `visualizationType` using a recommendation model built **independently** of the chart being switched away from.
- A complex-chart LLM node on the wiz-services side (e.g. the map binding path) can pick its own `aggregateFormula` for a measure directly against `/viewsheet/create`, bypassing this model entirely — so that choice was never visible to `changeType`, and its own per-type default (e.g. `Sum` for any numeric column) could silently overwrite it on every type switch.
- `ChangeTypeRequest` gains an optional `fieldConfigs` list carrying the caller's already-resolved formula per measure. `WizAutoBindingService.changeType()` applies it in **both** branches:
  - **fast path** (recommendation model reused): re-applies `fieldConfigs` onto the freshly-selected recommendation via the new `applyResolvedFormulaOverrides`, mirroring the existing `applyFieldConfigs(VSChartInfo, ...)` for chart assemblies, with a new `applyCrosstabAggregateFormulas` for crosstab (which had no equivalent before this change).
  - **fallback path** (model missing, full `autoBindingInternal()` re-run): looks up the resulting assembly by `runtimeId`/`assemblyName` and applies the same override before fetching data for the response — deliberately **NOT** forwarded through `AutoBindingRequest.fieldConfigs`, since that field is the AUTHORITATIVE full column list (`selectBindColumns` filters the worksheet down to exactly those columns) — forwarding a measures-only list there would silently drop every dimension from the rebuild.

Companion fix on the wiz-services side ([inetsoft-technology/stylebi-wiz#1445](https://github.com/inetsoft-technology/stylebi-wiz/pull/1445)) addresses the root cause — a map/complex-chart LLM node picking the wrong formula in the first place for an already-aggregated worksheet column. This PR makes `changeType` robust to that class of mismatch regardless of source.

## Tests

- New: `WizAutoBindingServiceChangeTypeFieldFormulaTest` (4 cases) for `applyCrosstabAggregateFormulas`.
- Existing `WizAutoBindingServiceCrosstabPinsTest` / `SelectBindColumnsTest` / `AxisTitleTest` / `DateGroupTest` all still pass.
- `mvnw -pl core -am compile` and `mvnw -pl server -am compile` (from this repo's root) both clean.

## Not yet done

- End-to-end verification against a running instance (local dev environment was unstable during this session).

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/model/ChangeTypeRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/model/ChangeTypeRequest.java
@@ -19,6 +19,8 @@ package inetsoft.web.wiz.model;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 
+import java.util.List;
+
 /**
  * Request body for {@code POST /api/wiz/viewsheet/changeType}.
  */
@@ -93,6 +95,25 @@ public class ChangeTypeRequest {
       this.assemblyName = assemblyName;
    }
 
+   /**
+    * Field-level overrides (currently only {@code aggregateFormula} is honored) carried over from
+    * the visualization being switched away from. changeType's own recommendation model — whether
+    * reused from a prior {@code autoBinding} call or freshly recomputed by the fallback below —
+    * has no idea what formula the LIVE chart actually ended up using for a measure (e.g. a
+    * complex-chart LLM node like the map binding path picks its own formula directly against
+    * {@code /viewsheet/create}, entirely bypassing the wizard recommender this model comes from),
+    * so its generic per-type default (e.g. Sum for any numeric column) can silently override an
+    * intentional choice like Count on the SAME field. Passing the caller's already-resolved
+    * formula here lets it win over that default instead of being clobbered on every type switch.
+    */
+   public List<SimpleFieldInfo> getFieldConfigs() {
+      return fieldConfigs;
+   }
+
+   public void setFieldConfigs(List<SimpleFieldInfo> fieldConfigs) {
+      this.fieldConfigs = fieldConfigs;
+   }
+
    private String worksheetId;
    private String visualizationType;
    /**
@@ -113,4 +134,5 @@ public class ChangeTypeRequest {
    private String viewsheetIdentifier;
    private boolean copy;
    private String assemblyName;
+   private List<SimpleFieldInfo> fieldConfigs;
 }
--- a/core/src/main/java/inetsoft/web/wiz/model/ChangeTypeRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/model/ChangeTypeRequest.java
@@ -95,17 +95,7 @@ public class ChangeTypeRequest {
       this.assemblyName = assemblyName;
    }
 
-   /**
-    * Field-level overrides (currently only {@code aggregateFormula} is honored) carried over from
-    * the visualization being switched away from. changeType's own recommendation model — whether
-    * reused from a prior {@code autoBinding} call or freshly recomputed by the fallback below —
-    * has no idea what formula the LIVE chart actually ended up using for a measure (e.g. a
-    * complex-chart LLM node like the map binding path picks its own formula directly against
-    * {@code /viewsheet/create}, entirely bypassing the wizard recommender this model comes from),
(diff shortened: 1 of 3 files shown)


Zxm-happy:
Addressed in \`6eabdc753\`:

- **Observation 1 (same gap in \`autoBindingInternal()\`'s crosstab branch)** — fixed. A fresh \`autoBinding\` call resolving straight to a crosstab now also gets \`applyCrosstabAggregateFormulas(crosstabAsm.getVSCrosstabInfo(), configMap)\`, alongside the existing \`applyCrosstabHeaderPins\` call — mirroring the sibling chart branch's \`applyFieldConfigs\`. Good catch, this was exactly the same bug class on a different entry point.
- **Observation 2 (narrow test coverage)** — added \`ApplyResolvedFormulaOverridesDispatchTest\` (nested in the same test file): chart → \`applyFieldConfigs\`, crosstab → \`applyCrosstabAggregateFormulas\`, neither (table) → no-op, null assembly → no-op. Made \`applyResolvedFormulaOverrides(VSAssembly, Map)\` package-private (was \`private\`) so this could test the dispatch directly against a service instance (mirroring \`WizAutoBindingServiceSetChartFormatTest\`'s pattern of constructing the service with unused collaborators left \`null\`), rather than only reachable through the full \`changeType()\`. Did not add a full \`changeType()\`-level integration test (fast-path/fallback-path wiring) — that would need mocking the recommendation-model machinery end to end, which felt like a bigger lift than this PR's scope; flagging in case you'd rather have it before merge.
- **Observation 3 (javadoc placement)** — moved the \`fieldConfigs\` rationale from the getter to the field declaration, matching this class's own convention for its other fields.
- **Observation 4 (best-effort exception swallowing)** — left as-is, per your own framing ("reasonable given... not correctness-critical"); the existing \`LOG.warn\` is the intended visibility mechanism.

Verified: \`mvnw -pl core -am compile\` clean; \`WizAutoBindingServiceChangeTypeFieldFormulaTest\` (now 8 cases) plus \`CrosstabPinsTest\`/\`SelectBindColumnsTest\`/\`AxisTitleTest\`/\`DateGroupTest\`/\`AutoBindingSecurityTest\`/\`AestheticModelTest\`/\`SetChartColorsTest\`/\`SetChartFormatTest\` all still pass.

Zxm-happy:
Fixed the leftover nit in \`5016c5a\` — removed the unused \`org.junit.jupiter.api.BeforeEach\` import from \`WizAutoBindingServiceChangeTypeFieldFormulaTest.java\`. Re-ran the test class, still 8/8 passing.

Zxm-happy:
Fixed in \`7f37df9\` — you're right that this is exactly the same failure mode this PR exists to close, just via the design/runtime split. Rather than rely on empirically proving whether \`fetchAssemblyData\`'s \`box.executeView(...)\` always re-derives \`aggrs2\` from the (by-then-corrected) design array in time — I traced it far enough to believe it likely does (both \`changeType\` paths apply the override *before* the fetch that triggers execution), but couldn't fully confirm the sandbox internals without a live end-to-end run — I went with your suggested defensive fix: mirror \`applyCrosstabHeaderPins\`' exact pattern.

\`applyCrosstabAggregateFormulas\` now also pushes the override onto \`getRuntimeAggregates()\` when it's already non-empty, via a shared \`applyAggregateFormulasTo(DataRef[], Map)\` extracted for both arrays. Added 3 tests: runtime aggregates already populated get overridden too, an empty runtime array is left alone, a null runtime array is safe (design-only path, matching the existing coverage).

Verified: \`mvnw -pl core -am test-compile\` clean; \`WizAutoBindingServiceChangeTypeFieldFormulaTest\` now 11/11 (7 + the 4-case nested dispatch test), plus \`CrosstabPinsTest\`/\`SelectBindColumnsTest\`/\`AxisTitleTest\`/\`DateGroupTest\`/\`AutoBindingSecurityTest\`/\`AestheticModelTest\`/\`SetChartColorsTest\`/\`SetChartFormatTest\` all still pass.

fredaan (approved):

PR #4515, opened 2026-08-05
Bug #75957, resolve set_ranking field against AggregateInfo, not the pre-aggregate column

jkidacki-inetsoft:
## Summary

The `set_ranking` worksheet-agent op resolved its field against the pre-aggregate column selection only, never checking `AggregateInfo`. Ranking a grouped/aggregated table by the aggregate (e.g. top 3 employees by `Sum(Total)`) bound the ranking condition to the raw pre-aggregate `Total` column instead of the `Sum(Total)` aggregate, so the resulting filter displayed as `Total top3` instead of `Sum(Total) top3`.

## Root Cause

`WorksheetMutationSupport.setRanking()` called the 2-arg `resolveField(t, spec.field())` overload, which always passes `post=false` and therefore only searches the private/pre-aggregate `ColumnSelection`. The sibling post-aggregate (HAVING) condition path, `setConditions(..., post=true)`, already resolves via the 3-arg overload, which checks `AggregateInfo` (aggregate or group-by dimension) first before falling back to the column selection.

`RankingCondition` always runs after aggregation in the query pipeline (`AssetQuery.getRankingTableLens`, step 9, after `getSummaryTableLens`), so it should resolve the same way post-aggregate conditions do.

## Fix

Changed `setRanking()` to call `resolveField(t, spec.field(), true)`. This resolves the field against `AggregateInfo` first — matching either an aggregate ref (by alias, view, or base attribute) or a group-by dimension — before falling back to the plain column selection, exactly like `set_post_conditions`. Ranking on a non-aggregated table is unaffected, since an empty `AggregateInfo` falls straight through to the existing column-selection lookup.

## Test Plan

- Added `setRankingOnAggregatedTableResolvesAggregateRef`: ranking by the aggregate on a grouped table resolves to the `AggregateRef`.
- Added `setRankingOnAggregatedTableByDimensionResolvesGroupRef`: ranking by the group-by dimension (not the aggregate) on the same table resolves to the `GroupRef`.
- Added `setRankingOnUngroupedTableResolvesRawColumn`: ranking on a plain, non-aggregated table still resolves to the raw column (no regression).
- `./mvnw test -pl core -Dtest=WorksheetEditServiceMutatorsTest` — all 45 tests pass.
- Manually verified in the live worksheet-chat plugin: grouping by Employee, aggregating Total via Sum, then asking for the top 3 employees by Sum(Total) now shows the filter as `Sum(Total) top3`.

Fixes Redmine #75957.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetMutationSupport.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetMutationSupport.java
@@ -889,7 +889,13 @@ public final class WorksheetMutationSupport {
       int op = "BOTTOM_N".equalsIgnoreCase(spec.operation())
          ? XCondition.BOTTOM_N : XCondition.TOP_N;
 
-      DataRef ref = resolveField(t, spec.field());
+      // The RankingCondition itself always runs after aggregation (see
+      // AssetQuery.getRankingTableLens), so resolve against AggregateInfo first — this
+      // matches both an aggregate (e.g. "Sum(Total)") and a group-by dimension (e.g.
+      // "Employee") — before falling back to the plain column selection, like
+      // set_post_conditions does. Otherwise a rank on "Sum(Total)" binds to the raw
+      // "Total" column instead of the aggregate ref.
+      DataRef ref = resolveField(t, spec.field(), true);
       inetsoft.uql.asset.RankingCondition rc = new inetsoft.uql.asset.RankingCondition();
       rc.setOperation(op);
       rc.setN(spec.n());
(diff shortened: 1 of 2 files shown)


jkidacki-inetsoft:
Addressed in b30374aab. You're right — passing \`post=true\` to \`resolveField\` also switched the fallback lookup from the private column selection to the public one, which strips invisible columns and would have broken ranking on a hidden, non-aggregated column.

Extracted the \`AggregateInfo\` matching into a new \`resolveAggregateOrGroupField()\` helper. \`setRanking()\` now calls that directly and falls back to \`resolveField(t, field, false)\` (private selection) instead, restoring hidden-column support while keeping the aggregate/group-first resolution that fixed the original bug. \`resolveField\`'s own behavior for \`set_post_conditions\` (still \`post=true\`, still falls back to the public selection) is unchanged — confirmed via the full test suite (69 tests, all passing).

Also added a deterministic regression test (\`setRankingFallsBackToPrivateSelectionWhenColumnMissingFromPublic\`) that explicitly sets a public selection missing the target column, and verified it fails without this fix and passes with it.

jshobe-inetsoft (approved):
Approved based on Claude's automated review, which found no blocking issues after the final commit.

PR #4516, opened 2026-08-05
Bug #75930, apply round corner to Table/Crosstab on Excel export

ngovus-inetsoft:
## Summary
- Table/Crosstab/CalcTable round-corner formatting was silently dropped in Excel export — the table rendered with plain square corners even though the viewer and PDF/SVG/PowerPoint export render it rounded.
- Root cause is different from Bug #75922 (TextInput/Text): tables write their cells directly to the sheet grid rather than as a floating shape, so `ExcelTableHelper.drawObjectFormat()` and `ExcelCrosstabHelper.drawObjectFormat()` were empty stubs — there was no shape at all for round-corner logic to apply to (unlike TextInput's `XSSFTextBox`).
- Fix: draw a rounded-rectangle outline overlay on top of the cell grid, matching the outer-frame rounding already done for Table/Crosstab in PDF/SVG (`ExportUtil`) and PowerPoint (`PPTValueHelper`/`PPTVSUtil`, fixed as part of #75922's review round).
  - `PoiExcelVSExporter.drawTableRoundedBorder()`: creates a borderless, rounded-rectangle outline shape anchored to the table's pixel bounds, picking border style/color from whichever side is set first (top → left → right → bottom).
  - Since Excel has no way to clip cell rendering to a rounded shape, the border curve intruding into the corner would otherwise cut across cell text/backgrounds that fill the grid all the way to the edge (e.g. a shaded title row, or the corner-most cell's value). To avoid this, the effective radius is capped (8px) and the four corners are masked with a plain white square before the outline is drawn, so the corner reads as cleanly clipped rather than bisected — mirroring the `overflow: hidden` clipping the viewer already applies at rounded corners.
  - `ExcelTableHelper.java` / `ExcelCrosstabHelper.java` (covers Table, Crosstab, and CalcTable/freehand table, which routes through `ExcelCrosstabHelper`): `drawObjectFormat()` now calls `drawTableRoundedBorder()` on the border-only pass, after cell data is written.

**Known limitation:** the corner mask assumes a white sheet background. If the canvas behind the table isn't white, the mask squares will show as a mismatched white patch instead of blending in.

## Test plan
- [ ] Create a viewsheet with a Table/Crosstab that has round corner + border set.
- [ ] Export to Excel with "match layout" and confirm the corners read as rounded, without the border cutting through cell text or a shaded title/header background poking past the curve.
- [ ] Confirm existing non-rounded Table/Crosstab exports are unaffected.
- [ ] Confirm PDF/SVG/PowerPoint table export is unaffected (already correct prior to this change).

### DIFF
--- a/utils/inetsoft-xml-formats/src/main/java/inetsoft/report/io/viewsheet/excel/ExcelCrosstabHelper.java
+++ b/utils/inetsoft-xml-formats/src/main/java/inetsoft/report/io/viewsheet/excel/ExcelCrosstabHelper.java
@@ -270,6 +270,18 @@ public class ExcelCrosstabHelper extends VSCrosstabHelper {
    @Override
    protected void drawObjectFormat(TableDataVSAssemblyInfo info, VSTableLens lens,
                                    boolean borderOnly) {
+      // the border is drawn as an overlay shape on top of the cell data (rather than as
+      // native cell borders, which have no rounded-corner concept), so only draw once,
+      // after the cells are written, to match the "borderOnly" (second) pass used by
+      // PDF/SVG export.
+      if(!borderOnly || info == null) {
+         return;
+      }
+
+      Rectangle2D bounds = new Rectangle2D.Double(
+         info.getPixelOffset().x, info.getPixelOffset().y,
+         info.getPixelSize().width, info.getPixelSize().height);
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):

PR #4517, opened 2026-08-05
Bug #75916, fix mini-toolbar z-index clamp so it isn't covered by sibling assemblies

ngovus-inetsoft:
## Summary
- The floating assembly mini-toolbar (shown on hover, e.g. over a chart) could be rendered underneath/covered by sibling assemblies with a higher server-assigned z-index, because its z-index was derived solely from its own assembly's z-index (`getContainerZIndex(vsObject) + 1`) — a change made in #75794 to fix max-mode/embedded-viewsheet stacking.
- Adds `GuiTool.MINI_TOOLBAR_MIN_ZINDEX` (9920, matching the toolbar's original fixed CSS z-index) and a new `getMiniToolbarZIndex()` that clamps the dynamic z-index to that floor via `Math.max`.
- This restores "toolbar beats ordinary siblings" for normal hover cases while preserving the larger boosted value for max-mode/embedded-viewsheet/datatip cases (which can already exceed 9920), so the #75794 fix is not reverted.

## Test plan
- [ ] Open viewsheet `Examples/Census` in portal, hover over the chart, confirm the toolbar renders above/clear of the text box (previously covered)
- [ ] Verify max-mode toolbar visibility still works correctly (no regression to #75794 fix)
- [ ] `npx tsc --noEmit` passes on the portal project

### DIFF
--- a/web/projects/portal/src/app/common/util/gui-tool.ts
+++ b/web/projects/portal/src/app/common/util/gui-tool.ts
@@ -57,6 +57,11 @@ export class GuiTool {
 
    static readonly MINI_TOOLBAR_HEIGHT = 28;
 
+   // Must stay in sync with the base .mini-toolbar z-index in mini-toolbar.component.scss.
+   // Used as a floor so the toolbar always outranks ordinary sibling assemblies, even when
+   // its own assembly's server-assigned z-index is low.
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):

PR #4518, opened 2026-08-05
Bug #75926, reject filter/condition mutations on embedded and snapshot tables

jkidacki-inetsoft:
## Summary

The worksheet-chat MCP plugin's `add_filter`, `edit_condition`, `set_conditions`, and `set_post_conditions` tools mutate a worksheet table's condition list directly via `WorksheetEditService.Editor`, with no check on the table's concrete type. This let the plugin silently apply a filter/condition to an `EmbeddedTableAssembly` or `SnapshotEmbeddedTableAssembly`, even though the Composer UI never allows this: `ws-details-pane.component.ts`'s `openConditionDialog()` checks `table.isEmbeddedTable()` and, if true, shows an info dialog ("Filtering is not supported for snapshot. Create a mirror to define conditions.") instead of opening the condition dialog. The tester reported this as the plugin "silently no-opping," but the actual defect is that it wasn't a no-op — it genuinely mutated the model with no user-visible error, unlike the UI.

## Root Cause

`WorksheetEditService.Editor.addFilter` / `editCondition` / `setConditions` / `setPostConditions` had no guard equivalent to the frontend's `isEmbeddedTable()` check, so the plugin could reach a state the UI intentionally prevents.

## Fix

Added a `requireFilterable(TableAssembly)` guard in `WorksheetEditService.java` that throws a `PairingException` (reusing the exact UI-facing message via the `composer.ws.filter-snapshopt` catalog key) when the target table is an `EmbeddedTableAssembly` (which, via inheritance, also covers `SnapshotEmbeddedTableAssembly`). Wired into `addFilter`, `editCondition`, `setConditions`, and `setPostConditions`. `removeFilter` and `set_ranking` are intentionally left unguarded — out of scope for this defect.

Updated `WorksheetEditServiceMutatorsTest` to use a new non-embedded stand-in table (`TestWorksheets.nonEmbeddedTableWithColumns`, backed by `BoundTableAssembly`) for the existing filter/condition tests, since they had been using a plain embedded table purely as a convenient in-memory stub, unrelated to embedded-table semantics. Added new tests asserting the guard rejects both embedded and snapshot-embedded tables across all four mutators.

## Test Plan

- [x] `./mvnw clean install -pl core -am -DskipTests` — build succeeds
- [x] `./mvnw test -pl core -Dtest=WorksheetEditServiceMutatorsTest,WorksheetReadServiceTest` — 50/50 tests pass
- [x] Manually verified via the worksheet-chat plugin: `add_filter` on a snapshot table now returns the rejection message instead of a false success
- [x] Code review via code-reviewer agent — no blocking issues

Fixes Redmine #75926.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -437,12 +437,14 @@ public class WorksheetEditService {
        * @param field     the column name to filter on
        * @param operation comparison operator: {@code "="}, {@code "!="}, {@code "<"}, {@code ">"}
        * @param values    one or more literal string values
-       * @throws PairingException if no {@link TableAssembly} with {@code table} exists
+       * @throws PairingException if no {@link TableAssembly} with {@code table} exists, or if
+       *                          {@code table} is an embedded or snapshot-embedded table
        */
       public void addFilter(String table, String field,
                             String operation, String... values) throws PairingException
       {
          TableAssembly t = requireTable(table);
+         requireFilterable(t);
          requireColumn(t, field);
          WorksheetMutationSupport.addFilter(t, field, operation, values);
       }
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):
Approving based on the resolved Claude automated review (no blocking issues) and passing checks.

PR #4521, opened 2026-08-05
Bug #75958, fix TypeError when opening highlight conditions dialog

ngovus-inetsoft:
## Summary
- `VSConditionDialog.ngOnInit()` looked up the accurate field definition for every entry in `conditionList`, but that array alternates between `Condition` objects (even indices) and junction-operator objects (odd indices). Junction entries have no `field` property, so `cond.field.view` threw `TypeError: Cannot read properties of undefined (reading 'view')`.
- The thrown error aborted `ngOnInit` partway through, which is why the highlight Conditions dialog appeared to hang before eventually opening in a broken state.
- Skip entries without a `field` (junction operators) before doing the field lookup.

## Test plan
- [x] Open a viewsheet with a highlight that has an AND/OR condition list, right-click the highlighted cell → Highlight → select a highlight → Edit, and confirm the Conditions dialog opens immediately with no console error.


jshobe-inetsoft (approved):

PR #4522, opened 2026-08-05
Bug #75959, defer objectHeight write to avoid NG0100 on wrap text tables

ngovus-inetsoft:
## Summary
- Fixes an `ExpressionChangedAfterItHasBeenCheckedError` (NG0100) thrown in the browser when wrap text changes a table's shrink-to-fit height.
- `BaseTable.getObjectHeight()` mutated `model.objectHeight` as a side effect of a getter evaluated inside the table's own template. The ancestor `vs-object-container` reads that same field in its own template, which Angular checks before descending into the child table's template in the same change-detection pass — causing the dev-mode `checkNoChanges` pass to see a stale-vs-updated mismatch.
- The write is now deferred to a microtask (with `markForCheck()`) so it lands in the next change detection cycle instead of mutating shared state mid-cycle. The method's return value used for the table's own rendering is unaffected.

## Test plan
- [x] Reproduced the wraptext viewsheet scenario and confirmed the NG0100 console error no longer appears.
- [ ] `npm run test:portal` for `base-table`/`vs-table`/`vs-crosstab` specs.

### DIFF
--- a/web/projects/portal/src/app/vsobjects/objects/table/base-table.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/table/base-table.ts
@@ -1642,7 +1642,21 @@ export abstract class BaseTable<T extends BaseTableModel> extends AbstractVSObje
             // add data row height
             this.model.scrollHeight;
 
-         return this.model.objectHeight = height;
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):

PR #4523, opened 2026-08-05
Bug #75964, fix web link field dropdown rendering all fields on one row

ngovus-inetsoft:
## Summary
- In the viewsheet Hyperlink dialog, selecting Web Link → Field showed all available fields concatenated on a single row instead of each field on its own line.
- Root cause: the placeholder `<option>` in the field `<select>` was never closed before the `@for` loop generating per-field options, so those options became DOM children of the placeholder option instead of siblings inside `<select>`. A `<select>` only renders one row per direct child, so the browser collapsed all field names into one row.
- Fix: close the placeholder `<option>` immediately and make the `@for` block a sibling inside `<select>`.

## Test plan
- [ ] Open a viewsheet in Composer, right-click a table cell → Hyperlink
- [ ] Select Web Link, click "change value type" icon, select Field
- [ ] Expand the field dropdown and confirm each field renders on its own line


jshobe-inetsoft (approved):

PR #4526, opened 2026-08-06
fix(wiz): keep a chart's field settings across a chart-type change

Nelson-cao:
## Problem

`changeType` rebuilds the assembly from the recommendation model, whose candidate chart infos hold clones of the wizard temp chart's refs taken when `autoBinding` ran (`ChartCombinationUtil.getChartInfosWithScores` reads `temp.getXFields()/getYFields()`; `ChartTypeFilter.getAllRefs(true)` clones each one into every candidate).

Field settings applied after that point — top-N ranking, sort, date group level, aggregate formula — were written only onto the rendered assembly, which the rebuild discards. `keepCondition` carries over just the assembly's `preConditionList`, and ranking is not part of that; `syncConfigs` would not have helped either, since `SyncChartHandler.syncChart` syncs presentation, condition, highlight and hyperlink but never touches a ref's binding semantics.

Result: `show top 3` followed by `change to a bar chart` rendered without the ranking, with no error.

## Fix

Make the temp chart the durable store for these settings, so they reach every later recommendation through the clone the recommender already makes.

**`BindingFieldSettings`** (new) holds the ref plumbing: `refsOf` / `snapshot` to collect an assembly's binding-field refs, and `record` / `restore` to move settings between two sets, pairing by bound column.

- Property-by-property rather than `clone()`, because source and target are frequently different subtypes — a map candidate's x is a `VSChartGeoRef` — mirroring the `VSChartGeoRef(VSChartDimensionRef)` constructor, whose property list it tracks.
- The key comes from the base `VSDimensionRef` / `VSAggregateRef` value accessors, so a crosstab's plain design headers pair with the chart side rather than falling through to `getAttribute()`.
- A column bound more than once is dropped from the index: the key cannot encode the formula or date level, since those are what is being carried.
- The two directions are asymmetric on purpose. `restore` copies only what the source carries, so a default-valued temp ref cannot overwrite what the recommender chose for the target type. `record` assigns unconditionally, so clearing a ranking on the chart clears it on the temp chart too.
- `Calculator` is cloned, matching `VSAggregateRef.clone()`'s own defensive copy.

**Write side.** `autoBindingInternal` records the request's `fieldConfigs` on the temp chart; `/viewsheet/create` records the assembly's settings there, from a snapshot taken before the pre-aggregation push (which rewrites exactly those values) and applied only after `persistViewsheet`, so a rolled-back create leaves the temp chart untouched. `changeType` records on both its branches — the fallback cannot restore older history, but still keeps what this call establishes.

**Read side.** `changeType` applies the temp chart's settings onto the assembly it just built. Needed only because the fast path reuses the *cached* recommendation model, so its candidates are a pre-edit snapshot; a fresh recommend inherits them through the clone. Ordered before `applyResolvedFormulaOverrides`, leaving an explicit caller `fieldConfigs` authoritative. If `changeType` is ever changed to re-run the recommendation, this call becomes redundant and can go away — noted at the call site.

**`CreateVisualizationModel`** takes `autoBindingRuntimeId`. There is no server-side link from an output runtime to its autoBinding RVS (`autoBindingRuntimeId` appears only in wire models), so the caller must name it. Optional — omitting it records nothing, which is what the MCP path does.

The matching wiz-services change that supplies the id: inetsoft-technology/stylebi-wiz#1462

## Tests

12 cases appended to `WizAutoBindingServiceChangeTypeFieldFormulaTest`, the file already covering the `applyFieldConfigs` family:

- `applyFieldConfigsToTempChart` records a top-N on the temp chart's dimension, leaves unmatched fields alone, and no-ops on a missing temp chart / temporary info / null runtime or an empty config map.
- `applyTempChartFieldSettings` restores ranking and an aggregate formula onto a rebuilt chart, skips refs with no temp counterpart, and leaves a candidate untouched when the temp ref carries no ranking — the guard against wiping the recommender's own choices for the target type.

The load-bearing case was confirmed to fail before the implementation was wired in.

`./mvnw -o test-compile -pl core` BUILD SUCCESS; 156 tests green across `WizVsService*`, `WizAutoBindingService*`, `BindingSlotsTest`, `SyncInfoHandlerSyncConfigsTest` and `SyncChartHandlerNullTempInfoTest`.


core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -1390,6 +1422,16 @@ private CreateViewsheetResult createViewsheetInternal(CreateVisualizationModel m
...
          CreateViewsheetResult.FlatBinding binding = collectFlatBinding(assembly);
+
+         // Record this request's field settings (top-N, sort, date level, formula) on the wizard temp
+         // chart, so a later rebuild-from-recommendation inherits them instead of reverting to the bare
+         // recommender defaults. Every create/edit that lands here is a semantic create OR edit of the
+         // binding, and the temp chart is the only place those settings survive one — see
+         // WizUtil.copySettingsByFieldName. Done BEFORE the pre-aggregation push below, which can
+         // rewrite formulas to None once aggregation moves to the worksheet: the temp chart must hold
+         // what the user asked for, not the pushed-down form.
+         syncFieldSettingsToTempChart(model.getAutoBindingRuntimeId(), assembly, user);
```
Nelson-cao: **A removed setting can never reach the temp chart — and the round trip resurrects it.**

`copySettingsByFieldName` only ever writes a value it finds present on the source; it has no "the source deliberately has nothing here" branch. That is the right call for the *read* direction (it is what stops a default temp ref from clobbering the recommender's per-type choice) but it makes the *write* direction one-way:

1. `show top 3` → the temp dim gets `TOP_N`.
2. `show all months` → a create/update lands here with the ranking cleared on `assembly`. `isActiveRanking(null)` is false, so the temp chart keeps `TOP_N`.
3. `change to a bar chart` → `applyTempChartFieldSettings` pushes `TOP_N` back onto the rebuilt assembly, and then this line copies it back into temp again.

The user's removal is silently reverted one turn later — the same class of bug this PR fixes, in the other direction. It is self-reinforcing too: once temp holds a ranking, every `changeType` re-asserts it on both sides. Same for `sortByColValue` and `manualOrderList`.

Note that `applyFieldConfigsToTempChart` does *not* have this problem — it delegates to `applyFieldConfig`, which sets `rankingOptionValue` unconditionally whenever `ranking != null`, so an explicit `optionValue = NONE` clears it. So the two write paths added by this PR disagree on what a clear means. Worth making them agree: either give the write direction an explicit clear (a separate method from the read-direction copy, since the asymmetry there is intentional), or drop this write-back and let `applyFieldConfigsToTempChart` be the only writer.

Nelson-cao (commented):
The direction is right — the temp chart genuinely is the only binding state that survives a rebuild-from-recommendation, and routing settings through it is the correct place to fix this. The read side (`applyTempChartFieldSettings`) looks sound. My concerns are all on the new **write-back** side (`WizVsService.syncFieldSettingsToTempChart`) and on the field-name matching primitive, which is now load-bearing for both directions.

Summary of what I think needs to change before merge:

1. **A setting can never be un-set, and the round trip makes it self-reinforcing.** `copySettingsByFieldName` is copy-only-if-present in both directions, so a removal never reaches the temp chart — and the next `changeType` puts it back.
2. **`Calculator` is installed by reference, not cloned** — `VSAggregateRef.clone()` clones it; this doesn't, so the temp chart and the rendered assembly end up sharing one mutable object.
3. **`putIfAbsent` keyed on column value collapses two refs of the same column** — `Sum(amount)` + `Average(amount)` on Y, or a date column bound at two levels, silently get the first ref's settings applied to both.
4. **The crosstab branch of `bindingFieldRefs` derives its match key differently from the chart branch** and has no test coverage.

Plus two smaller ones inline (rollback scope, and the fallback branch recording nothing).

The test additions are well-targeted at the two new service methods, but everything is mock-`verify()`-based against a single `VSChartDimensionRef` per side, so `copySettingsByFieldName`'s own indexing/dedup behaviour and the crosstab branch are untested. Points 3 and 4 both live in exactly that gap.

core/src/main/java/inetsoft/web/wiz/WizUtil.java
```diff
@@ -156,6 +166,192 @@ else if(info instanceof TableDataVSAssemblyInfo tableInfo) {
...
+
+   private static void copyAggregateSettings(VSAggregateRef from, VSAggregateRef to) {
+      String formula = from.getFormulaValue();
+
+      if(formula != null && !formula.isEmpty()) {
+         to.setFormulaValue(formula);
+      }
+
+      if(from.getCalculator() != null) {
+         to.setCalculator(from.getCalculator());
```
Nelson-cao: `Calculator` is mutable, and this installs the **same instance** on both refs rather than a copy. `VSAggregateRef.clone()` treats it as owned state and clones it defensively:

```java
agg.calculator = calculator == null ? null : (Calculator) calculator.clone();
```

After this line the temp chart's ref and the rendered assembly's ref share one object, so a later mutation on either side silently changes the other — and the temp chart is exactly the long-lived state that should not be reachable that way.

It is also inconsistent with the `manualOrderList` handling a few lines up, which *does* defensively copy (`new ArrayList<>(manualOrder)`).

Suggest `to.setCalculator((Calculator) from.getCalculator().clone());` (needs the `inetsoft.uql.viewsheet.graph.Calculator` import).

core/src/main/java/inetsoft/web/wiz/WizUtil.java
```diff
@@ -156,6 +166,192 @@ else if(info instanceof TableDataVSAssemblyInfo tableInfo) {
...
+         return;
+      }
+
+      Map<String, DataRef> index = new HashMap<>();
+
+      for(DataRef ref : fromRefs) {
+         String name = WizardRecommenderUtil.getChartRefFieldName(ref);
+
+         if(name != null && !name.isEmpty()) {
+            index.putIfAbsent(name, ref);
```
Nelson-cao: **`putIfAbsent` on a column-value key collapses two refs of the same column onto the first one's settings.**

`getChartRefFieldName` returns `getColumnValue()` for an aggregate and `getGroupColumnValue()` for a dimension — neither includes the formula or the date level. So two distinct refs over the same column share a key, the first wins the index, and the loop below then applies that one ref's settings to *every* target with that name. Two concrete cases:

- **Two aggregates of one measure**, e.g. `Sum(amount)` and `Average(amount)` both on Y. The index keeps `Sum`; `copyAggregateSettings` then writes `setFormulaValue("Sum")` onto the `Average` target too. The user's second measure silently becomes a duplicate of the first.
- **One date column at two levels**, e.g. `Year(order_date)` on X and `Quarter(order_date)` on colour. Both targets get the X ref's `dateLevelValue`.

Both are reachable in both directions (`changeType`'s restore and the new write-back), and neither is covered by the new tests, which use a single ref per side.

The key needs to disambiguate — include the formula for aggregates and the date level for dimensions (the ref's full name is the usual key for this in the codebase), or at minimum pair within a slot rather than flattening x/y/aesthetics into one namespace.

core/src/main/java/inetsoft/web/wiz/WizUtil.java
```diff
@@ -156,6 +166,192 @@ else if(info instanceof TableDataVSAssemblyInfo tableInfo) {
...
+            if(aref != null && aref.getDataRef() != null) {
+               refs.add(aref.getDataRef());
+            }
+         }
+      }
+      else if(assembly instanceof CrosstabVSAssembly crosstabAsm &&
+              crosstabAsm.getVSCrosstabInfo() != null)
+      {
+         VSCrosstabInfo info = crosstabAsm.getVSCrosstabInfo();
+         addAll(refs, info.getDesignRowHeaders());
```
Nelson-cao: **The crosstab branch produces refs whose match key is derived differently from the chart branch's, so the two sides may never pair up.**

`getChartRefFieldName` only special-cases `VSChartDimensionRef` / `VSChartAggregateRef`. A crosstab's design headers and aggregates are plain `VSDimensionRef` / `VSAggregateRef`, so they fall through to `ref.getAttribute()`:

```java
// VSDimensionRef
public String getAttribute() { return group == null ? "" : group.getAttribute(); }
// VSAggregateRef
public String getAttribute() { return ref == null ? "" : ref.getAttribute(); }
```

The temp chart is always a chart, so its side of the comparison always uses `getGroupColumnValue()`/`getColumnValue()`. Those agree with `getAttribute()` for a bare unqualified column, but not necessarily for an entity-qualified one, and an unresolved ref yields `""`, which the `!name.isEmpty()` filter in `copySettingsByFieldName` then drops entirely — a silent no-match rather than an error.

This path is live in both directions: a `changeType` to a crosstab hits `applyTempChartFieldSettings` with a `CrosstabVSAssembly`, and `syncFieldSettingsToTempChart` is called with whatever `createViewsheetInternal` produced. There is no test for it. Either extend `getChartRefFieldName` to handle the plain `VSDimensionRef`/`VSAggregateRef` cases with the same value-based accessors, or add a crosstab case to the new tests that pins down what actually matches.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -89,6 +89,38 @@ public WizVsService(ViewsheetService viewsheetService, AssetRepository engine,
...
    }
 
+   /**
+    * Copies {@code assembly}'s binding-field settings onto the autoBinding RVS's temp chart, so they
+    * outlive the rendered assembly. Best-effort: no autoBindingRuntimeId (every caller that predates it,
+    * and the MCP path), an expired runtime, or an RVS that never went through autoBinding all mean there
+    * is no temp chart to record on — the assembly itself already has the settings either way, so this
+    * must never fail the request that triggered it.
+    */
+   private void syncFieldSettingsToTempChart(String autoBindingRuntimeId, VSAssembly assembly,
```
Nelson-cao: Two smaller points on this method.

**Rollback scope.** The call site runs *before* the big `try` in `createViewsheetInternal` whose `catch` restores the worksheet and viewsheet and rethrows. So a create that fails and rolls back still leaves the temp chart permanently mutated — and since the temp chart is now read on the next `changeType`, that mutation is observable. Moving the call after the mutations succeed (or covering it in the rollback) would match the rest of the method's transactional behaviour. The javadoc's "must never fail the request" is satisfied either way by the `try/catch` here.

**Accessor consistency.** This reads the temp info via `WizUtil.getTempChart` → `rvs.getVSTemporaryInfo()` directly, while the two new call sites in `WizAutoBindingService` go through `temporaryInfoService.getVSTemporaryInfo(rvs)`, which exists specifically to take the init read-lock when the field is null. Between this, the service, and the direct read already in `changeType`, there are now three ways to reach the same state in one feature — worth funnelling through the service.

core/src/main/java/inetsoft/web/wiz/service/WizAutoBindingService.java
```diff
@@ -2151,6 +2226,12 @@ public CreateViewsheetResult changeType(ChangeTypeRequest request, Principal use
          // EXACTLY those columns) — forwarding our measures-only list here would silently drop every
          // dimension (e.g. STATE) from the rebuild instead of just fixing a formula. Applied post-hoc
          // onto the resulting assembly below instead, the same as the fast path.
+         // KNOWN GAP: unlike the fast path below, this branch cannot restore accumulated field settings
```
Nelson-cao: The comment is accurate that this branch cannot *restore* anything — the temp chart holding the history is gone. But it can still **record**: `autoBindingInternal` has just created a fresh RVS (`resp.getAutoBindingRuntimeId()`), and the caller's `configMap` is in hand and about to be applied to the rendered assembly a few lines below. As written those settings land only on the assembly, so the *next* `changeType` drops them again — this branch reintroduces the exact bug the PR fixes, one turn later, rather than merely failing to recover older history.

An `applyFieldConfigsToTempChart(<the new rvs>, configMap)` alongside the existing `applyResolvedFormulaOverrides` would close that, and is a small addition given the method already exists.

core/src/main/java/inetsoft/web/wiz/WizUtil.java
```diff
@@ -156,6 +166,192 @@ else if(info instanceof TableDataVSAssemblyInfo tableInfo) {
...
+   // The temp chart in an autoBinding RVS's VSTemporaryInfo is what every recommendation is generated
+   // from: ChartCombinationUtil reads temp.getXFields()/getYFields() and ChartTypeFilter.getAllRefs(true)
+   // puts a CLONE of each temp ref into every candidate chart info. So a setting recorded on the temp
+   // chart is inherited by every candidate of every chart type, while one applied only to the rendered
+   // assembly is discarded by the next rebuild-from-recommendation. These helpers move settings in both
+   // directions: onto the temp chart when an edit lands, and back off it when a cached recommendation
+   // (changeType's fast path) produced a pre-edit snapshot.
+
+   /** The temp chart of an autoBinding RVS, or null when the RVS never went through autoBinding. */
+   public static ChartVSAssembly getTempChart(RuntimeViewsheet autoBindingRvs) {
```
Nelson-cao: Minor / non-blocking: `bindingFieldRefs`, `copySettingsByFieldName` and `copyBindingFieldSettings` are generic ref plumbing with no dependency on `WizUtil`'s existing contents, and adding them pulls `inetsoft.web.vswizard.*` (`VSTemporaryInfo`, `WizardRecommenderUtil`) into what was a small decode/assembly-info helper — +190 lines into a file that was previously ~170. They would sit more naturally next to the other wizard ref utilities in `WizardRecommenderUtil`, or in their own class.

Nelson-cao:
All seven stand up — I verified each against the source before deciding. Fixing all of them; two needed a different remedy than the one suggested, reasoning below.

**1. A removed setting never reaches the temp chart, and the round trip resurrects it.** Correct, and the asymmetry with `applyFieldConfigsToTempChart` is real: `applyFieldConfig` sets `rankingOptionValue` unconditionally whenever `ranking != null`, so an explicit NONE clears, while my copy silently keeps the old value.

Splitting the primitive by direction rather than dropping the write-back. Read stays merge-if-present (that asymmetry is deliberate — it is what stops a default temp ref from clobbering the recommender's per-type choice); write assigns unconditionally, so a cleared ranking/sort/manual order clears the temp chart too. Not dropping the write-back and leaving `applyFieldConfigsToTempChart` as the only writer, because the BINDING_FIELD path reaches `/viewsheet/create` with its settings inside `config.bindingInfo`, not as `fieldConfigs` — there is no configMap to hand that method there.

**2. `Calculator` installed by reference.** Correct — `VSAggregateRef.clone()` does `(Calculator) calculator.clone()` at line 1245, and `manualOrderList` right above already copies defensively. Cloning it.

**3. `putIfAbsent` collapsing two refs of one column.** Correct, both cases are reachable. Not keying on the full name, though: the full name embeds precisely what is being carried — the formula for an aggregate, the date level for a dimension — so keying on it means a formula or level change stops matching and the setting never transfers, which is the feature. Slot-pairing does not work either: the read direction exists because slots MOVE across a type change (a dimension going from x to a colour aesthetic is the pie/treemap case).

Fixing it by making an ambiguous column carry nothing: when more than one source ref shares a key, drop it from the index instead of letting the first win. A doubly-bound column then keeps its own settings rather than silently inheriting a sibling's. Covered by tests for both of your cases.

**4. Crosstab match key.** Correct. Fixing locally rather than in `WizardRecommenderUtil.getChartRefFieldName`, which has five other callers: the key function tests the base `VSDimensionRef` / `VSAggregateRef` with the same value accessors, which the chart subtypes inherit, so both sides derive keys the same way. Adding crosstab coverage.

**5. Rollback scope and accessor consistency.** Both correct. Moving the call inside the `try`, after the mutations succeed, so a rolled-back create no longer leaves the temp chart mutated. And injecting `VSWizardTemporaryInfoService` into `WizVsService` so all three sites take the same read path — it depends only on `ViewsheetService`, so no cycle.

**6. The fallback branch can still record.** Correct, and worse than I described in the comment: as written it reintroduces the bug one turn later rather than only failing to recover history. Adding `applyFieldConfigsToTempChart` against the freshly-created RVS.

**7. Placement.** Agreed — moving the three helpers into their own `inetsoft.web.wiz.BindingFieldSettings` instead of growing `WizUtil`, which keeps `inetsoft.web.vswizard.*` out of the decode/assembly-info helper.

Nelson-cao:
Correcting my own record: the "Covered by tests for both of your cases" line in my earlier reply no longer holds, and the review is right to flag it.

I did write those cases — ambiguous-column drop for `Sum(amount)`/`Average(amount)` and for one date column at two levels, the crosstab pairing, the `Calculator` clone, and `record()`'s clear-on-absence — as a `BindingFieldSettingsTest` nested in `WizAutoBindingServiceChangeTypeFieldFormulaTest`. They were then dropped before commit, by decision, so no automated test for `BindingFieldSettings` exists in this PR. I should have amended the claim at that point rather than leaving it standing.

So, accurately: all four of those behaviours rest on inspection, not on a regression test. What the test changes in `4e620a52` actually contain is only the mechanical constructor-argument update across the existing `WizVsService*` test files.

What has been verified: `./mvnw -o test-compile -pl core` BUILD SUCCESS, and 156 tests green across `WizVsService*`, `WizAutoBindingService*`, `BindingSlotsTest`, `SyncInfoHandlerSyncConfigsTest` and `SyncChartHandlerNullTempInfoTest` — that count includes the 12 cases this PR added earlier for `applyFieldConfigsToTempChart` and `applyTempChartFieldSettings`, which do exist, plus the suites the base-branch merge brought in.

On the two remaining direct `rvs.getVSTemporaryInfo()` reads (`WizAutoBindingService.java:160` and `:2224`): agreed, both predate this PR and are outside its diff, so leaving them is deliberate.

Separately, the earlier CI failure was mine — the constructor change missed `WizVsServiceResolveRuleHighlightRefsTest`, which the base branch added after this branch was cut, so it existed only in the merge result. Fixed in `37c714e13` along with a merge of the current base.

Nelson-cao:
Test coverage added in `8009351f` — the gap the last review flagged is closed, and my earlier claim now holds.

Eight cases as a `BindingFieldSettingsTest` nested in `WizAutoBindingServiceChangeTypeFieldFormulaTest`, covering all four behaviours named:

- `record()` clears a ranking the source no longer carries; `restore()` leaves the target's alone in the same situation — the asymmetry pinned from both sides
- two aggregates of one column (`Sum` + `Average`), and one column bound twice on the target, both carry nothing rather than collapsing onto the first ref
- a crosstab's plain `VSDimensionRef` design header pairs with the temp chart's dimension
- the `Calculator` is cloned, not shared
- plus `snapshot()` surviving a later mutation of the live ref, and a straight restore-carries-the-ranking case

Real refs rather than mocks, as suggested: what is under test IS the accessor each key is derived from, so stubbing `getGroupColumnValue()` / `getColumnValue()` would assume away the crosstab defect. That pulls in `GDefaults` / `AggregateFormula` static init, so the class carries the same SreeHome + Spring harness `SyncChartHandlerNullTempInfoTest` uses for the same reason.

Each case was confirmed to fail with its own fix reverted, in two passes: reverting the ambiguity drop, the clone and the record-clear failed exactly `oneColumnBoundTwiceOnTheTargetCarriesNothing`, `twoAggregatesOfOneColumnCarryNothing`, `recordClearsARankingTheSourceNoLongerCarries` and `theCalculatorIsClonedRatherThanShared`; removing the base-type branches from `fieldKey` failed `aCrosstabHeaderPairsWithTheTempChartDimension` along with the other pairing-dependent cases.

164 tests green across `WizVsService*`, `WizAutoBindingService*`, `BindingSlotsTest`, `SyncInfoHandlerSyncConfigsTest` and `SyncChartHandlerNullTempInfoTest`.

fredaan (approved):

PR #4528, opened 2026-08-06
Bug #75982, allow schedule task permission changes for folder-scoped admins

ngovus-inetsoft:
## Summary
- Repository permission "Apply" on an individual Schedule Task failed for a user whose only grant was ADMIN on the parent "Schedule Tasks" folder (not a literal permission record on the task itself), because `SCHEDULE_TASK` resources don't inherit permissions from their parent `SCHEDULE_TASK_FOLDER`.
- `RepositoryObjectController.setResourcePermission()` now falls back to the same owner/admin-over-owner check (`ScheduleManager.hasTaskPermission`) that `ResourcePermissionController` already uses for schedule tasks, so a task's owner (or an admin over the owner) can manage its permissions.
- The frontend `apply()` handler in `repository-permission-editor-page.component.ts` previously swallowed save failures silently, leaving the editor in a dirty state and popping the confusing "Repository Settings Changed" warning on navigation with no visible error. It now shows a snackbar error on failure.

## Test plan
- [x] `RepositoryObjectControllerTest` updated for the new `ScheduleManager` constructor dependency; passes.
- [x] `community/core` module compiles cleanly.
- [x] Manually verified in EM: user with only "Admin" on the Schedule Tasks folder node can now change permissions on a task they created under it, Apply succeeds, and no stale "unsaved changes" warning appears when switching nodes.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/content/repository/RepositoryObjectController.java
+++ b/core/src/main/java/inetsoft/web/admin/content/repository/RepositoryObjectController.java
@@ -19,6 +19,8 @@ package inetsoft.web.admin.content.repository;
 
 import inetsoft.report.internal.Util;
 import inetsoft.sree.RepositoryEntry;
+import inetsoft.sree.schedule.ScheduleManager;
+import inetsoft.sree.schedule.ScheduleTask;
 import inetsoft.sree.security.*;
 import inetsoft.uql.asset.*;
 import inetsoft.uql.asset.internal.AssetUtil;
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):
Claude's automated review found no blocking issues (two non-blocking follow-up suggestions: a test coverage gap and some logic duplication with ResourcePermissionController). All checks pass and there are no merge conflicts. Approving.

PR #4532, opened 2026-08-06
Bug #75935, sync undo/redo state after agent-driven worksheet edits and saves

jkidacki-inetsoft:
## Summary

The `stylebi-worksheet-chat` MCP agent can headlessly edit and save a worksheet that's simultaneously open in a browser Composer tab, by directly manipulating the same live `RuntimeWorksheet`/`RuntimeViewsheet` and pushing STOMP refresh/save commands back to that tab. The Composer's "modified"/unsaved-changes tab indicator did not correctly reflect these agent-driven changes: it never turned on during agent edits, and could incorrectly remain "modified" right after a successful agent-triggered save.

## Root Cause

The Angular Composer's dirty indicator is `current !== savePoint` on the client-side `Sheet` model (`sheet.ts`). Those fields are updated client-side only by an `UpdateUndoStateCommand` STOMP message, which the normal browser-driven edit path sends automatically via an AspectJ advice around `@Undoable` STOMP handlers (`VSLayoutService.makeUndoable()`). The agent/pairing edit path (`WorksheetEditService` → `SheetAgentBroadcastService.broadcastRefresh()`, and `WorksheetAgentController.save()` → `broadcastSave()`) never sent this command, so the client's `current`/`savePoint` fields went stale relative to the server's actual runtime state — never advancing during edits (indicator stays off), and mismatched right after save (indicator turns on incorrectly).

## Fix

Add `SheetAgentBroadcastService.sendUndoState()`, which builds an `UpdateUndoStateCommand` from the runtime's `size()/getCurrent()/getSavePoint()/getID()` — mirroring `VSLayoutService.makeUndoable()`'s construction exactly — and call it after every `broadcastRefresh()` (worksheet and viewsheet branches) and after `broadcastSave()`, so the client's undo-state stays in sync regardless of whether the mutating request came from the browser's own STOMP session or the external agent session.

## Test Plan

- Added/updated unit tests in `SheetAgentBroadcastServiceTest` verifying `UpdateUndoStateCommand` is sent with the correct `points`/`current`/`savePoint`/`id` after both `broadcastRefresh()` and `broadcastSave()`.
- `./mvnw test -pl core -Dtest=SheetAgentBroadcastServiceTest` — 6/6 pass.
- `./mvnw test -pl core -Dtest="inetsoft.web.wiz.**.*Test"` — 782/782 pass (no regressions in the wiz package).
- Manually verified end-to-end: connected to a live worksheet via `stylebi-worksheet-chat`, made an edit, confirmed the Composer tab's modified indicator turned on immediately; saved via the plugin and confirmed the indicator cleared and stayed cleared.

Fixes Redmine #75935.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/pairing/SheetAgentBroadcastService.java
+++ b/core/src/main/java/inetsoft/web/wiz/pairing/SheetAgentBroadcastService.java
@@ -32,6 +32,7 @@ import inetsoft.web.composer.ws.command.RefreshWorksheetCommand;
 import inetsoft.web.composer.ws.command.SetWorksheetInfoCommand;
 import inetsoft.web.viewsheet.command.RefreshVSObjectCommand;
 import inetsoft.web.viewsheet.command.SaveSheetCommand;
+import inetsoft.web.viewsheet.command.UpdateUndoStateCommand;
 import inetsoft.web.viewsheet.model.VSObjectModel;
 import inetsoft.web.viewsheet.model.VSObjectModelFactoryService;
 import inetsoft.web.viewsheet.service.CommandDispatcher;
@@ -87,11 +88,13 @@ public class SheetAgentBroadcastService {
 
       if(sheetType == SheetType.VIEWSHEET) {
          broadcastViewsheetRefresh((RuntimeViewsheet) rs, runtimeId, sessionId, user);
+         sendUndoState(rs, runtimeId, sessionId, user);
          return;
       }
 
       Object command = buildCommand(sheetType, rs, owner);
       sendCommand(user, sessionId, runtimeId, command);
+      sendUndoState(rs, runtimeId, sessionId, user);
 
       LOG.info("Pairing broadcast refresh sent (sheetType={}, runtimeId={}, sessionId={})",
                sheetType, runtimeId, sessionId);
@@ -177,6 +180,25 @@ public class SheetAgentBroadcastService {
          .build();
 
       sendCommand(user, sessionId, runtimeId, saveCommand);
+
+      // 3. Sync current/savePoint/points so the Composer's modified indicator clears,
+      // even if this save is the first agent action in the session (no prior refresh).
+      sendUndoState(rs, runtimeId, sessionId, user);
+   }
+
+   /**
+    * Push the runtime's current undo/redo position to the browser so the Composer's
+    * "modified" indicator (current !== savePoint) stays in sync with agent-driven edits,
+    * mirroring what {@code VSLayoutService.makeUndoable()} does for the browser-native
+    * STOMP {@code @Undoable} path.
+    */
+   private void sendUndoState(RuntimeSheet rs, String runtimeId, String sessionId, String user) {
+      UpdateUndoStateCommand command = new UpdateUndoStateCommand();
+      command.setPoints(rs.size());
+      command.setCurrent(rs.getCurrent());
+      command.setSavePoint(rs.getSavePoint());
+      command.setId(rs.getID());
+      sendCommand(user, sessionId, runtimeId, command);
    }
 
    private void sendCommand(String user, String sessionId, String runtimeId, Object command) {
--- a/core/src/test/java/inetsoft/web/wiz/pairing/SheetAgentBroadcastServiceTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/pairing/SheetAgentBroadcastServiceTest.java
@@ -23,6 +23,7 @@ import inetsoft.report.composition.RuntimeWorksheet;
 import inetsoft.uql.viewsheet.TextVSAssembly;
 import inetsoft.uql.viewsheet.VSAssembly;
 import inetsoft.uql.viewsheet.Viewsheet;
+import inetsoft.web.viewsheet.command.UpdateUndoStateCommand;
 import inetsoft.web.viewsheet.model.VSObjectModel;
 import inetsoft.web.viewsheet.model.VSObjectModelFactoryService;
 import inetsoft.web.viewsheet.service.CommandDispatcher;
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (approved):
## Code Review — Bug #75935, sync undo/redo state after agent-driven worksheet edits and saves

**Approved.** The diagnosis is accurate and the fix is minimal, additive, and consistent with the existing browser-native path. Verification notes below, followed by a few non-blocking observations.

---

### ✅ What I verified

**The construction mirrors `makeUndoable()` exactly.** `sendUndoState()` (SheetAgentBroadcastService.java:195-202) sets `points/current/savePoint/id` from `rs.size()/getCurrent()/getSavePoint()/getID()` — field-for-field identical to `VSLayoutService.makeUndoable()`'s else-branch (VSLayoutControllerService's sibling, VSLayoutService.java:1004-1011). `RuntimeSheet.size()` is `points.size()` and `getCurrent()` is `point` (RuntimeSheet.java:389-407), so the semantics match the browser path.

**The runtime state the command reports actually moves.** This was the main thing worth checking — syncing a value that never changes would be a no-op fix:
- `WorksheetEditService.apply()` and `applyOnRuntime()` both call `rws.addCheckpoint(rws.getWorksheet().prepareCheckpoint())` *before* `broadcastRefresh()` (WorksheetEditService.java:113, 141), so `getCurrent()` has advanced by the time `sendUndoState` reads it. ✔
- `WorksheetAgentController.save()` sets `rws.setSavePoint(rws.getCurrent())` immediately before `broadcastSave()` (WorksheetAgentController.java:544-545), so the post-save command carries `savePoint == current` and the client's `isModified()` goes false. ✔
- Nice side benefit: `applyOnRuntimeNoCheckpoint()` (the undo/redo endpoints) also routes through `broadcastRefresh()`, so agent-driven undo/redo now moves the client's `current` too — previously the toolbar's undo/redo enablement would also have drifted.

**Client-side handling is already in place and safe.**
- `ws-pane.component.ts:1164` and `viewsheet-pane.component.ts:930` both implement `processUpdateUndoStateCommand`; neither filters on `command.id`, so setting it is harmless (and matches `makeUndoable`, which sets it for the same reason).
- `layout-pane.component.ts` only handles `UpdateLayoutUndoStateCommand`, so a sheet-level undo-state message arriving while the user is in layout mode is ignored rather than corrupting layout point state. ✔
- Unhandled commands are silently dropped in `command-processor.ts:99-110` (the `console.warn`s are commented out), so components without a handler get no console noise.

**Command ordering in `broadcastSave` is safe.** `processSaveSheetCommand` mutates `this.worksheet` **in place** and `composer-main.component.ts:3273 onSaveWorksheetFinish()` does not swap the sheet object. So the subsequent `processUpdateUndoStateCommand`'s `new Worksheet(this.worksheet)` copy captures the just-set `id`/`savePoint`, and `updateSheet()`'s identity lookup (`original === sheet`, composer-main.component.ts:2410) still matches. No lost-update window.

**Serialization/type header.** `UpdateUndoStateCommand` is a plain POJO (not an Immutables value), so `resolveCommandType()` correctly yields `"UpdateUndoStateCommand"` without hitting the `Immutable`-stripping branch. ✔

**Other checks:** no new data is exposed (the command goes to the same `sessionId`/`user` as the existing refresh/save commands on that runtime); purely additive on the wire, so no backward-compatibility concern; the three updated existing tests correctly account for the extra command, and `headersCap.getAllValues().get(0)` is the refresh since `sendUndoState` runs after `sendCommand`.

---

### 🟡 Non-blocking observations

1. **The viewsheet branch is currently a no-op for the modified indicator.** `sendUndoState` is wired into the `SheetType.VIEWSHEET` path (SheetAgentBroadcastService.java:91), but none of `ScriptEditService`'s three mutation entry points (`apply`, `applyOnRuntime`, `applyOnRuntimeIfChanged`) call `rvs.addCheckpoint(...)` — `git grep addCheckpoint` under `web/wiz/` hits only `WorksheetEditService`. So on the VS path the command faithfully transmits *unchanged* `current`/`savePoint`, and the VS Composer's indicator still won't turn on after an agent script edit. (VS *save* is fine — `ViewsheetAgentController.java:278` does set `savePoint = current`.) That's consistent with this being a worksheet bug, and the VS call is harmless defensive syncing — just worth stating explicitly so nobody reads this PR as having fixed the viewsheet half. A follow-up adding the VS checkpoint would complete the symmetry.

2. **No `isDisposed()` guard.** `makeUndoable()` opens with `if(rs.isDisposed()) { return; }` (VSLayoutService.java:975). `sendUndoState` has no equivalent. In practice the runtime was just fetched via `getSheetForPairing` and the neighbouring `broadcastRefresh`/`broadcastSave` code doesn't guard either, so this is theoretical — but since the stated goal is to mirror `makeUndoable`, adding the guard would make the mirror exact.

3. **`currentTS` is bumped on every send.** Both client handlers set `currentTS = Date.now()`, which feeds `touchAsset()`'s `event.setChanged(isModified() && autoSaveTS != currentTS)` (ws-pane.component.ts:705, viewsheet-pane.component.ts:2289). On paths where the undo state didn't actually move (the VS branch per #1), this can cost one extra autosave write on the next tick for an already-modified sheet. Harmless, just noting the coupling.

4. **`broadcastSaveSendsUpdateUndoStateCommandSoIndicatorClears` doesn't test what its name claims.** It stubs `current=5, savePoint=2` — i.e. still modified — and asserts only field forwarding. The inline comment is admirably honest about this, but the name promises the indicator clears. Either rename to something like `...ForwardsRuntimeUndoPosition`, or add a second case stubbing `savePoint == current` so the actual bug condition is encoded in a test. (The existing distinct-values choice is the right call for catching a field swap — I'd keep that test and add the other, not replace it.)

5. **Exact-count verifications are brittle.** `verify(dispatcher, times(2)/times(3))` will break on any future added command rather than on a real regression. Since the new tests already filter by `instanceof UpdateUndoStateCommand`, the counts in the three pre-existing tests could relax to `atLeastOnce()` plus a type filter. Purely a maintenance preference.

6. **Cosmetic:** the `LOG.info("Pairing broadcast refresh sent ...")` at SheetAgentBroadcastService.java:99 now fires *after* the undo-state command is dispatched. Reads slightly oddly but is inconsequential.

---

### Non-code checks

- ✅ No failing CI checks (`license/cla` passing; no other checks configured on this branch).
- ✅ No merge conflicts — `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` against `stylebi-wiz-main-rebase`.
- ✅ No security concerns — the new command carries only integer undo positions and the runtime ID, addressed to the same STOMP session/user as the existing commands on this runtime.
- ✅ No backward incompatibility — additive STOMP command with pre-existing client handlers; unhandled elsewhere by design.
- ✅ No broken tests — the three existing assertions were correctly updated for the additional dispatch.

Approving. None of the observations above need to block the merge; #1 in particular is worth capturing as a follow-up rather than expanding this PR's scope.

PR #4533, opened 2026-08-06
Bug #75988, fix INETSOFT_ADMIN_PASSWORD not reaching server/scheduler containers

ngovus-inetsoft:
## Summary
- `INETSOFT_ADMIN_PASSWORD` was only set in the `storage` init container's environment in `community-examples/docker-compose.yaml`, so the `server` and `scheduler` containers never received it, causing "Enable Security" to fail even with a compliant password.
- Moved `INETSOFT_ADMIN_PASSWORD` into the `x-shared-config` anchor so it's applied to all containers (storage, server, scheduler) from one place, and removed the now-redundant explicit entry from the `storage` service.

## Test plan
- [x] `docker compose config` validates the compose file with `INETSOFT_ADMIN_PASSWORD` set
- [ ] Manually verify `docker compose up` followed by enabling security in EM succeeds with the shared password

### DIFF
--- a/Troubleshoot.md
+++ b/Troubleshoot.md
@@ -31,13 +31,12 @@ docker compose logs storage
 
 This error most commonly occurs when the `INETSOFT_ADMIN_PASSWORD` environment variable is missing or does not meet the password requirements. This variable sets the password for the "admin" user and is required — there is no default password. The password must be at least 8 characters and include an uppercase letter, a lowercase letter, a digit, and a special character.
 
-To resolve this, set `INETSOFT_ADMIN_PASSWORD` to a valid password before starting the containers, either in the `docker-compose.yaml` file:
+To resolve this, set `INETSOFT_ADMIN_PASSWORD` to a valid password before starting the containers, either in the `docker-compose.yaml` file's shared config (so it reaches the `storage`, `server`, and `scheduler` containers alike):
 
 ```yaml
-storage:
-  environment:
-    # the password for the "admin" user
-    INETSOFT_ADMIN_PASSWORD: "changeme"
+x-shared-config: &sharedenv
+  # the password for the "admin" user
+  INETSOFT_ADMIN_PASSWORD: "changeme"
 ```
 
 or as an environment variable in your shell before running `docker compose up`:
--- a/community-examples/docker-compose.yaml
+++ b/community-examples/docker-compose.yaml
@@ -3,6 +3,9 @@ x-shared-config: &sharedenv
   # the master password used for PBKDF2 encryption of sensitive data
   # (not part of inetsoft.yaml, but required by all containers)
   INETSOFT_MASTER_PASSWORD: "some_arbitrary_token"
+  # the password for the "admin" user - Required: set INETSOFT_ADMIN_PASSWORD in your shell environment before running docker compose.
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (approved):
## Code Review

**Bug:** #75988 — `INETSOFT_ADMIN_PASSWORD` was set only in the `storage` container's `environment:` block in `community-examples/docker-compose.yaml`, so it never reached `server`/`scheduler`. `server` needs it for `FileAuthenticationProvider` when enabling security, so "Enable Security" failed even with a fully-compliant password.

**Fix:** Moves `INETSOFT_ADMIN_PASSWORD` into the `x-shared-config` anchor (`sharedenv`), which every service (`storage`, `server`, `scheduler`) already merges in via `<<: *sharedenv`. The redundant explicit entry is removed from `storage`. `Troubleshoot.md` is updated to match the new location.

### Verification
I rendered the PR's `community-examples/docker-compose.yaml` with `docker compose config` (`INETSOFT_ADMIN_PASSWORD` set in the shell) and confirmed all three services — `storage`, `server`, and `scheduler` — now resolve `INETSOFT_ADMIN_PASSWORD` in their environment, which was not the case before this change.

### Review notes

- **Correctness:** ✅ The fix directly addresses the reported root cause and matches the resolution suggested in Redmine #75988. Since `INETSOFT_ADMIN_PASSWORD` no longer appears in `storage`'s explicit `environment:` map, there's no key conflict with the merged-in `sharedenv` value — behavior for `storage` is unchanged, and `server`/`scheduler` now receive the variable they were previously missing.
- **Docs:** ✅ `Troubleshoot.md` is updated consistently with the new `x-shared-config` location, so the guidance no longer contradicts the actual compose file.
- **Backward compatibility:** ✅ No behavioral change for existing users who already set `INETSOFT_ADMIN_PASSWORD` in their shell/`.env` — the same shell variable is still consumed via `"${INETSOFT_ADMIN_PASSWORD}"` interpolation, just from a single shared location instead of a per-service override.
- **Test coverage:** No automated tests cover `community-examples/docker-compose.yaml` (it's a standalone example, not wired into the Maven/CI build), so nothing is broken by this change, but there's also no regression test to catch this class of bug in the future. Given this is a documentation/example fix, that's acceptable.
- **Minor, non-blocking observation:** `docker/src/test/docker/docker-compose.yaml` (used by the Maven docker-image test build, not touched by this PR) has the same historical pattern — `INETSOFT_ADMIN_PASSWORD` is set explicitly in `storage`'s and `server`'s environment blocks rather than in `x-shared-config`, and is missing entirely from `scheduler`. It isn't part of this bug's scope, but may be worth a follow-up for consistency/correctness there too.
- **Security:** No concerns — no secrets are hardcoded; the value continues to come from the caller's shell environment via `${INETSOFT_ADMIN_PASSWORD}` interpolation, same as before.

### CI / mergeability
- Only check configured on this PR is `license/cla`, which **passes**.
- `mergeable: MERGEABLE` — no merge conflicts.
- `mergeStateStatus: BLOCKED` is solely due to the repo's required-review branch protection (`reviewDecision: REVIEW_REQUIRED`), not any conflict or failing check.

### Verdict
No blocking issues, no failing checks, no merge conflicts. Approving.

PR #4537, opened 2026-08-07
Bug #75970, reject change_column_type on physical/bound table columns

jkidacki-inetsoft:
## Summary

The Wiz worksheet-chat `change_column_type` MCP tool would mutate a column's `ColumnRef` data type in memory and report success, even for columns where the override is not actually persistable — either because the column lives on a plain physical/bound query table or join, or because it's a computed expression-aggregate measure.

## Root Cause

Every mutation applied via the Wiz session ends with `WorksheetEditService.refreshAssemblies()`, which calls `WorksheetEventUtil.refreshColumnSelection()` for every table. For a plain bound/physical table this re-derives the column selection straight from the live DB schema, silently discarding any manually-set type override moments after it was applied. Only embedded, tabular/REST, SQL-bound, and unpivot tables (and expression columns) have a persistence hook that survives this refresh — which is exactly why the browser Composer UI's `WSHeaderCellComponent.supportChangeColumnType()` only enables the "Column Type" action for those cases.

A second commit closes a narrower instance of the same failure mode, mirroring the UI's own `isExpressionAggregate` exclusion: an expression column that is also one of the table's aggregate measures (e.g. `SUM(some_expression)`) has its exposed type computed from the formula when the query runs, not read back off the column ref's `dataType`, so an override there is discarded too — even on an otherwise-allowed embedded/tabular/SQL-bound/unpivot table.

`WorksheetEditService.Editor.changeColumnType()` never enforced either restriction, so it reported `ok: true` for changes that were discarded immediately afterward.

## Fix

- Added a guard in `changeColumnType()` mirroring the UI's `supportChangeColumnType()` logic: rejected unless the column is an expression, or the table is `EmbeddedTableAssembly` (covers snapshot too), `TabularTableAssembly`, `SQLBoundTableAssembly`, or `UnpivotTableAssembly`.
- Added a second guard mirroring the UI's `isExpressionAggregate` exclusion: rejects an expression column that is also one of the table's configured aggregate measures (`AggregateInfo.getAggregate(DataRef)`). Scoped to match the UI exactly (expression-typed aggregates only) — a raw, non-expression aggregate measure stays allowed, confirmed against a live server that the UI itself allows this.

Both reject with a clear `PairingException` instead of reporting success for a change that won't stick.

Two additional restrictions raised in review were investigated and intentionally not added, since live testing against a running server contradicted the premise for both: a live+aggregate table-wide restriction for `TabularTableAssembly`/`SQLBoundTableAssembly`/`UnpivotTableAssembly` (a non-aggregate group-by column's type change on a live aggregated SQL-bound table persisted correctly after a forced refresh), and a concern that `SQLBoundTableAssembly` type overrides don't survive `refreshColumnSelection()` at all (also persisted correctly in the same live test).

## Test Plan

- Regression tests in `WorksheetEditServiceMutatorsTest`: `changeColumnTypeRejectsPhysicalBoundTable`, `changeColumnTypeAllowsEmbeddedTable`, `changeColumnTypeRejectsExpressionAggregateMeasureOnEmbeddedTable`, `changeColumnTypeAllowsNonExpressionAggregateMeasureOnEmbeddedTable`, `changeColumnTypeAllowsGroupColumnOnAggregatedEmbeddedTable`.
- `mvnw test -pl core -Dtest=WorksheetEditServiceMutatorsTest,WorksheetEditServiceTest,WorksheetAgentControllerTest` — 79 tests, 0 failures.
- `mvnw clean install -pl core -am -DskipTests` — build succeeds.
- User verified live against a running server via the worksheet-chat MCP tool on the reported repro (ORDERS1.DISCOUNT), plus additional live verification of the aggregate/live-mode edge cases discussed in review.

Fixes Redmine #75970.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -1041,6 +1041,27 @@ public class WorksheetEditService {
             throw new PairingException("Column not found: " + col);
          }
 
+         // Mirrors the UI's WSHeaderCellComponent.supportChangeColumnType(): a type
+         // change only survives the post-edit refreshColumnSelection() pass for
+         // expression columns, or for embedded/tabular/SQL-bound/unpivot tables that
+         // persist the override into their own definition. Every other table type
+         // (plain bound/physical tables, joins, etc.) has its column selection rebuilt
+         // from the source schema, which silently discards the override — so reject
+         // the request here instead of reporting success for a change that won't stick.
+         boolean isRangeRef = cr.getDataRef() instanceof NumericRangeRef ||
+            cr.getDataRef() instanceof DateRangeRef;
+         boolean supportsTypeChange = !isRangeRef && (cr.isExpression() ||
+            t instanceof EmbeddedTableAssembly || t instanceof TabularTableAssembly ||
+            t instanceof SQLBoundTableAssembly || t instanceof UnpivotTableAssembly);
+
+         if(!supportsTypeChange) {
+            throw new PairingException(
+               "Column type cannot be changed for \"" + col + "\" on table \"" + table +
+               "\". Changing the data type is only supported for expression columns " +
+               "and embedded, tabular/REST, SQL-bound, or unpivot tables — physical " +
+               "(bound query) tables and joins do not support changing column types.");
+         }
+
          // Also update the matching ref from findAttribute (same approach as
          // ColumnTypeDialogService) to ensure the canonical ref is updated.
          ColumnRef cr2 = (ColumnRef) cs.findAttribute(cr);
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -1042,15 +1042,20 @@ public class WorksheetEditService {
          }
 
          // Mirrors the UI's WSHeaderCellComponent.supportChangeColumnType(): a type
-         // change only survives the post-edit refreshColumnSelection() pass for
+         // override only survives the post-edit refreshColumnSelection() pass for
          // expression columns, or for embedded/tabular/SQL-bound/unpivot tables that
-         // persist the override into their own definition. Every other table type
-         // (plain bound/physical tables, joins, etc.) has its column selection rebuilt
-         // from the source schema, which silently discards the override — so reject
-         // the request here instead of reporting success for a change that won't stick.
+         // persist it into their own definition — every other table type has its
+         // column selection rebuilt from the source schema, discarding the override.
          boolean isRangeRef = cr.getDataRef() instanceof NumericRangeRef ||
             cr.getDataRef() instanceof DateRangeRef;
-         boolean supportsTypeChange = !isRangeRef && (cr.isExpression() ||
+
+         // An aggregate measure's exposed type is computed from the formula + input
+         // type when the query runs, not read back off this ref's dataType, so an
+         // override here is discarded the same way regardless of table type.
+         boolean isAggregateMeasure = t.isAggregate() && t.getAggregateInfo() != null &&
+            t.getAggregateInfo().getAggregate(cr) != null;
+
+         boolean supportsTypeChange = !isRangeRef && !isAggregateMeasure && (cr.isExpression() ||
             t instanceof EmbeddedTableAssembly || t instanceof TabularTableAssembly ||
             t instanceof SQLBoundTableAssembly || t instanceof UnpivotTableAssembly);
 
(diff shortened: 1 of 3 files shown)


jkidacki-inetsoft:
Addressed the main finding in a follow-up commit (b2dc2b2d8): added a guard that rejects `change_column_type` on any column that is currently one of the table's configured aggregate measures (via `AggregateInfo.getAggregate(DataRef)`), regardless of table type. An aggregate's exposed type is computed from its formula + input type when the query runs, not read back off the column ref's `dataType`, so an override there is discarded the same way a plain physical-table override is -- even on an otherwise-allowed embedded/tabular/SQL-bound/unpivot table. This is intentionally broader than the UI's `isExpressionAggregate` check (which only excludes expression-typed aggregate columns): the same discard mechanism applies to non-expression aggregates too (e.g. a plain `SUM(column)`), which the UI's narrower check would have missed. Group-by columns are unaffected since they pass their value through unchanged rather than computing one.

Added two regression tests (`changeColumnTypeRejectsAggregateMeasureOnEmbeddedTable`, `changeColumnTypeAllowsGroupColumnOnAggregatedEmbeddedTable`) confirming both directions. All 79 tests in the three affected suites pass (0 failures).

Left the live/design-mode restriction out of scope: verifying it would require confirming how `AssetQuerySandbox.active` state (shared with the human's browser Composer session pairing this Wiz session) actually behaves for a Wiz-driven table refresh, which needs live-server verification beyond what static analysis can confirm confidently -- flagging for a follow-up rather than guessing.

Also confirmed via `git show`/build: `EmbeddedTableAssembly` correctly covers `SnapshotEmbeddedTableAssembly`, and `TabularTableAssembly`/`SQLBoundTableAssembly` extending `BoundTableAssembly` doesn't leak a plain bound table through the allowlist.

jkidacki-inetsoft:
Narrowed the aggregate-measure guard in 505b2dad3, per direction not to make the Wiz tool more restrictive than the browser UI. The previous commit rejected any aggregate-measure column regardless of expression status; that was broader than the UI's own \`isExpressionAggregate\` check, which only excludes expression-typed aggregates. A live test against a running server confirmed the UI would in fact allow changing a raw (non-expression) aggregate measure's type on a SQL-bound table, so the guard now requires \`cr.isExpression()\` in addition to being an aggregate measure, matching the UI exactly.

Also live-tested the remaining open item from this review (the live+aggregate table-wide restriction for Tabular/SQLBound/Unpivot): changed a non-aggregate (group-by) column's type on a live, aggregated SQL-bound table and it persisted correctly after a forced refresh. That contradicts the premise that this restriction is needed server-side, so I'm not adding it — happy to revisit with concrete repro evidence if this turns out to be table/config-specific.

Also re-tested the SQL-bound persistence concern from the first review round the same way: a plain (non-aggregate) column's type change on the same table also persisted after a forced refresh, contradicting an initial read of `SQLBoundQuery.getPostBaseTableLens()`'s comment. Retracting that concern.

Test suite adjusted accordingly (`changeColumnTypeRejectsExpressionAggregateMeasureOnEmbeddedTable`, `changeColumnTypeAllowsNonExpressionAggregateMeasureOnEmbeddedTable` replacing the previous single test). 58 tests in the mutators suite, 0 failures.

jshobe-inetsoft (changes requested):
Not merging yet — the latest Claude Code Review (commit 505b2dad) flags an unresolved gap carried over from prior rounds:

`WorksheetEditService.supportsTypeChange` (around line 1058-1060) still omits the UI's trailing clause from `WSHeaderCellComponent.supportChangeColumnType()` (`ws-header-cell.component.ts:592-606`) — specifically `(isUnpivotTableAssembly ? !table.info.aggregate : table.mode != "live" || !table.info.aggregate)`. Concretely, a group-by (non-aggregate) column on a live-mode, aggregated `TabularTableAssembly`/`SQLBoundTableAssembly`, or on an aggregated `UnpivotTableAssembly`, is still allowed by the backend guard even though the UI would disable the action for the whole table in that state.

Since the PR title/description still frames this as the complete fix for #75970, please either close this remaining gap or explicitly note in the PR description that it's being tracked as a separate follow-up, so reviewers don't assume full UI parity has been reached.

jkidacki-inetsoft:
Re: the live+aggregate table-wide restriction (Tabular/SQLBound/Unpivot) flagged again in this review round — this was already live-tested and addressed in my 14:59:55 comment on this thread: I changed a non-aggregate (group-by) column's type on a live, aggregated SQL-bound table and it persisted correctly after a forced refresh, which contradicts the premise that this restriction is needed server-side. Leaving the guard as-is (isRangeRef + isExpressionAggregate only); happy to revisit with a concrete repro if this turns out to be table/config-specific.

jkidacki-inetsoft:
Re: the live+aggregate table-wide gate (Tabular/SQLBound/Unpivot) raised again in this round — to be clear, this isn't an outstanding gap in this commit, it's a deliberate decision not to add it, based on two separate live-server verifications (see the 14:59:55 and 19:21:30 comments on this thread): a non-aggregate (group-by) column's type change on a live, aggregated SQL-bound table persisted correctly after a forced refresh, which contradicts the premise that the restriction is needed server-side. The UI's blanket per-table gate appears to be UI-specific caution rather than something enforced by the underlying persistence path this guard protects against. Leaving `supportsTypeChange` as-is (`isRangeRef` + `isExpressionAggregate`); will revisit if a concrete repro shows the live+aggregate case actually discards an edit.

PR #4541, opened 2026-08-07
Bug #75929, fix slider-bound form value not persisted on submit

ngovus-inetsoft:
## Summary
- A vsform input (e.g. a Slider) bound to an embedded worksheet table cell via the Data tab, with submit-on-change off, did not write its value into the target cell when the Submit button was clicked, and the input's own runtime value snapped back to the stale value.
- Root cause: the Submit button's "refresh viewsheet after submit" option (checked by default) triggers `refreshViewsheet(..., resetRuntime=true)`, which reloads the entire base worksheet fresh from the asset repository (`RuntimeViewsheet.resetRuntime()` -> `Viewsheet.update()`). That discarded the in-memory embedded-table write before it was ever visible.
- Fix: commit the submitted values into their bound embedded tables immediately after they're applied (`VSInputService.commitInputTableWrites`), and stop forcing the destructive full worksheet reload for the "refresh after submit" case (`reset=true` already re-executes the viewsheet against current in-memory data). Also resolve the target column from the table's live `ColumnSelection` instead of a `DataRef` cached on the assembly info that's only refreshed on a full sandbox reset.

## Test plan
- [x] Manually reproduced: created a viewsheet with an embedded table, a Slider bound to a cell via Target/Column/Row, and a Submit button (submit-on-change off, default "refresh viewsheet after submit" on).
- [x] Confirmed before the fix: setting the slider and clicking Submit left the table cell unchanged and the slider reverted to its old value.
- [x] Confirmed after the fix: the table cell updates to the submitted value and the slider retains it.
- [x] `mvn clean install -pl core -am -DskipTests` builds successfully.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/ViewsheetSandbox.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/ViewsheetSandbox.java
@@ -4030,7 +4030,10 @@ public class ViewsheetSandbox implements Cloneable, ActionListener {
 
          if(ass instanceof EmbeddedTableAssembly) {
             EmbeddedTableAssembly eassembly = (EmbeddedTableAssembly) ass;
-            DataRef column = iassembly.getColumn();
+            // resolve the column from the current column selection instead of relying on
+            // iassembly.getColumn(), which is only refreshed on a full sandbox reset and
+            // may be stale if the binding was changed without reopening the viewsheet
+            DataRef column = eassembly.getColumnSelection(false).getAttribute(iassembly.getColumnValue());
             int row = iassembly.getRow();
 
             if(column != null && row > 0) {
--- a/core/src/main/java/inetsoft/web/viewsheet/service/VSInputService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/service/VSInputService.java
@@ -296,12 +296,49 @@ public class VSInputService {
                .toArray();
             multiApplySelection(vsId, assemblyNames,
                                                        selectedObjects, principal, dispatcher, linkUri);
+
+            // Commit the submitted values into their bound embedded tables now, before the
+            // "refresh viewsheet after submit" button option (SubmitVSAssemblyInfo default)
+            // does a full viewsheet reset. That reset re-derives each input's selected value
+            // from its bound table cell, which would otherwise clobber the value just applied
+            // above with the stale, pre-submit value before it ever gets written back.
+            commitInputTableWrites(vsId, assemblyNames, principal, dispatcher, linkUri);
          }
       }
 
       onClick(vsId, name, x, y, linkUri, false, null, principal, dispatcher);
 
       return null;
+   }
+
+   /**
+    * Write the current (just-applied) selected values of the given input assemblies into
+    * their bound embedded table cells, if any.
+    */
+   private void commitInputTableWrites(String vsId, String[] assemblyNames, Principal principal,
+                                       CommandDispatcher dispatcher, String linkUri)
+      throws Exception
+   {
+      RuntimeViewsheet rvs = viewsheetService.getViewsheet(vsId, principal);
+      Optional<ViewsheetSandbox> box = rvs.getViewsheetSandbox();
+
+      if(box.isEmpty()) {
+         return;
+      }
+
+      ChangedAssemblyList clist =
+         coreLifecycleService.createList(true, dispatcher, rvs, linkUri);
+
+      box.get().lockWrite();
+
+      try {
+         for(String assemblyName : assemblyNames) {
+            box.get().processChange(assemblyName, VSAssembly.OUTPUT_DATA_CHANGED, clist);
+         }
+      }
+      finally {
+         box.get().unlockWrite();
+      }
    }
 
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (changes requested):
## Code Review — Bug #75929: slider-bound form value not persisted on submit

### Summary of change
This fixes a real bug: an input (e.g. Slider) bound to an embedded worksheet table cell, with "submit on change" off, silently lost its value when the Submit button's default "refresh viewsheet after submit" option ran, because that option called `refreshViewsheet(..., resetRuntime=true)`, which reloads the base worksheet from the asset repository and discards the in-memory cell write. I traced the flow end-to-end and the fix is well-targeted:

- `VSInputService.onClick(..., VSSubmitEvent, ...)` now calls a new `commitInputTableWrites(...)` right after `multiApplySelection(...)`, which drives `ViewsheetSandbox.processChange(name, VSAssembly.OUTPUT_DATA_CHANGED, clist)` for each submitted assembly. That hint routes to `outputDataChanged()` → `refreshVariable(InputVSAssembly, ...)`, which is the code path that actually writes the input's selected value into the bound embedded table cell (`ViewsheetSandbox.java:3915-4052`). This closes the gap for the `submitOnChange=false` case, where `refreshVS0` (`VSInputService.java:2938`) never triggers that path today because it's gated on `info.isSubmitOnChange() && info.isRefresh()`.
- The Submit button's "refresh after submit" handler (`VSInputService.java:2578`) now passes `resetRuntime=false` instead of `true`. I confirmed in `CoreLifecycleService.refreshViewsheet` (line ~613) that `resetRuntime` is what triggers `rvs.resetRuntime()` (the destructive full reload), while `reset=true` alone still re-executes the sandbox via `box.get().reset(...)` (line ~764) against current in-memory data. So dropping `resetRuntime` here is correct and matches the PR's stated rationale — it stops the destructive reload while preserving the "refresh viewsheet" behavior.
- `ViewsheetSandbox.refreshVariable` now resolves the target column via `eassembly.getColumnSelection(false).getAttribute(iassembly.getColumnValue())` instead of the cached `iassembly.getColumn()`. I checked `InputVSAssemblyInfo.update(Viewsheet, ColumnSelection)` (line 516-520), which is the *only* place that populates the cached `column` field, and it's called only on a full sandbox reset — this confirms the PR's claim that the cached value goes stale if the binding changes without a full reload, and the fix mirrors the same `ColumnSelection.getAttribute(...)` resolution `update()` uses.

### Findings

1. **Minor: possible semantic drift between `getColumnValue()` and the runtime-evaluated design value** — `InputVSAssemblyInfo.update()` resolves the column via `columnValue.getRuntimeValue(true)` (the runtime-evaluated value, which can differ from the static value if the binding is a variable/expression), but the new code in `ViewsheetSandbox.refreshVariable` uses `iassembly.getColumnValue()`, which returns `columnValue.getDValue()` (the static design-time string). In practice, the Data-tab column binding for these inputs (`DataInputPaneModel.setColumnValue`, e.g. `VSInputService.java:1023`) is always set as a plain literal column name, never a dynamic expression, so this is unlikely to manifest as a real bug today — but it's a latent inconsistency worth a code comment or a follow-up if `columnValue` is ever allowed to be dynamic for these inputs.

2. **Nit: same "cached column can be stale" pattern exists elsewhere and wasn't touched** — `InputVSAQuery.getData()` (line 195) and `InputVSAQuery.resetEmbeddedData()` (line 579) both still call `assembly.getColumn()`, the same cached/stale accessor this PR fixes in `ViewsheetSandbox.refreshVariable`. These are a default-value read and a reset-to-original-value step (both called from the same `outputDataChanged` flow, just before the now-fixed write). Since they run right before the fixed write and get overwritten by it, I don't think this causes the reported bug, but if the binding can go stale for the write path it can presumably go stale for these too — worth a follow-up ticket if not intentionally out of scope for this fix.

3. **No automated test coverage** — the fix touches core sandbox/session-write logic but ships with only a manual test plan (no new/updated JUnit test). I didn't find existing test scaffolding for `VSInputService`/`ViewsheetSandbox` submit flows in `core/src/test`, so a full automated regression test may be disproportionately expensive to add here — flagging as a suggestion, not a blocker.

None of the above are correctness blockers — I could not find a scenario where the change regresses existing behavior; `resetRuntime=false` with `reset=true` still re-executes the full viewsheet, and `commitInputTableWrites` only adds a write that previously never happened for this code path.

### CI / mergeability status
- `claude-review` check is currently **FAILING** — inspected the run log; it failed with `error_max_turns` (the automated review bot hit its 50-turn limit and errored out before completing), not a substantive finding against this code.
- `build`, `Semgrep Scan`, `Semgrep OSS`, and `license/cla` all pass.
- `mergeable: MERGEABLE`, but `mergeStateStatus: BLOCKED` (no approving review yet).

### Verdict
The code change itself looks correct, minimal, and well-reasoned based on tracing the actual call paths involved. However, per review policy this PR cannot be approved while a check is failing. Requesting changes to reflect the failing `claude-review` check — please re-run/investigate that job (it appears to be an infra timeout rather than a real finding) so it can go green, and then this should be ready to merge. Optionally consider the two minor follow-up notes above.

jshobe-inetsoft (approved):

PR #4542, opened 2026-08-07
Bug #75991, reject add_variable when a duplicate name already exists

jkidacki-inetsoft:
## Summary

`add_variable` silently overwrote an existing worksheet variable when given a duplicate name, even if the existing variable had a different type — destroying its default value and any `$(name)` references with no warning.

## Root Cause

`WorksheetAgentController.createVariable()` always constructed a new `DefaultVariableAssembly` and called `ws.addAssembly(assembly)` without checking whether an assembly with that name already existed. `Worksheet.addAssembly()` silently replaces an existing assembly of the same name (by design — this generic replace-on-name-match behavior is relied on elsewhere and is out of scope here), so a second `add_variable` call for the same name silently discarded the original variable.

## Fix

`createVariable()` now throws a `PairingException` when an assembly with the given name already exists, directing the caller to use `edit_variable` instead — mirroring the existing duplicate-name guards already in this file for `add_column` and `duplicate_assembly`.

## Test Plan

- Added `editRejectsAddVariableWithDuplicateName` — asserts a second `add_variable` call for an existing name throws `PairingException` and leaves the original variable assembly (name, type, default value) untouched.
- Added `editAddVariableCreatesVariableWhenNameIsNew` — confirms the normal (non-duplicate) `add_variable` path is unaffected.
- `./mvnw test -pl core -Dtest='inetsoft.web.wiz.worksheet.**'` — 87 tests, 0 failures.
- `./mvnw clean install -pl core -am -DskipTests` — builds clean.
- Manually verified live via the worksheet-chat plugin: creating `minTotal` (double) then re-adding `minTotal` (date) now returns an error instead of silently overwriting.

Fixes Redmine #75991.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -1480,7 +1480,19 @@ public class WorksheetAgentController {
     */
    private void createVariable(Worksheet ws, String name, String type,
                                String label, String defaultValue)
+      throws PairingException
    {
+      if(ws.getAssembly(name) != null) {
+         // Fail loud: Worksheet.addAssembly() silently replaces an existing assembly
+         // of the same name (or, for a same-name assembly of a different type, silently
+         // no-ops). Either way a second add_variable call would destroy or discard state
+         // without warning. edit_variable is the explicit path for changing an existing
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
Claude's automated review found no issues — clean, minimal, well-tested fix. All checks pass and the PR is mergeable.

PR #4544, opened 2026-08-07
Support an alias on a group-by field, mirroring the aggregate alias

larryliang-inetsoft:
An aggregate has always been able to name its output column; a group could not. That asymmetry made one
thing **impossible**, not merely untidy.

## The problem

An unaliased `dateGroupLevel` group is named after the expression `DateRangeRef` renders —
`Month(T.due_date)`. That is not a SQL alias, so it cannot be referenced from a downstream `sql: true`
expression column under **any** form:

| reference | result |
|---|---|
| `Month(T.due_date)` bare | ❌ `column "Month(T.due_date)" does not exist` |
| `"Mirror"."Month(...)"` | ❌ `missing FROM-clause entry for table` |
| `field['…']` in `sql:true` | ❌ JS-evaluated → `ReferenceError: COALESCE is not defined` |

So the canonical shared axis over a FULL join — `COALESCE(left_key, right_key)` — had no way to name its
own join keys, and was unexpressible in pushed-down SQL. Aggregates were fine (`COALESCE(starting, due)`
works) *precisely because* they can be aliased.

The workaround is a `sql: false` JS expression, and it isn't free: **one `sql: false` expression makes the
table not `sqlMergeable`**, so the entire pipeline stops being pushed to the database. Fine on a thousand
rows; not on a large table.

## The change

`applyAggregateInfo` now honours `GroupByFieldInfo.alias`, resolved to the **same private `ColumnRef` the
group description already uses** — the `DateRangeRef` column for a date-grouped field, otherwise the
private column of the same name — because the public selection is regenerated as clones of `privateCs`.
That is exactly how the aggregate branch a few lines below already applies its own alias.

## Verification

Live, against a locally built image (`local-1169`) on a real datasource: two aggregate mirrors grouped by
month with aliases `start_month` / `due_month`, FULL-joined on those aliases, then
`COALESCE(start_month, due_month)` as a `sql: true` expression.

Renders 20 months with both series totalling 984 — **row-for-row identical** to the `sql: false` JS
equivalent it replaces, with no pre-truncation mirror and the coalesce pushed down to the database.

4 unit tests (`WorksheetTableServiceGroupAliasTest`):
- a date-grouped alias becomes addressable
- a plain group alias works
- **an omitted alias leaves the rendered name untouched**, so no existing caller's column names move
- a blank alias is ignored rather than blanking the column name

## Compatibility

Additive and opt-in — an omitted alias behaves exactly as before (pinned by a test). Note the request
model is `@JsonIgnoreProperties(ignoreUnknown = true)`, so an older StyleBI silently ignores the field;
the wiz side detects that and reports `groupAliasesNotApplied` rather than letting a caller discover it
as a downstream "column does not exist". Paired wiz PR: inetsoft-technology/stylebi-wiz#1488.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/WizUtil.java
+++ b/core/src/main/java/inetsoft/web/wiz/WizUtil.java
@@ -156,6 +156,88 @@ public class WizUtil {
       }
    }
 
+   /**
+    * Infix of the lookup table wiz injects purely to label a foreign key — {@code <source>__fk_<target>}
+    * in fkJoinBatchRewrite. Deliberately does NOT match the join itself ({@code <source>__fkjoin}, no
+    * trailing underscore), which carries the fact rows and must be capped like any other detail table.
+    */
+   private static final String FK_LABEL_LOOKUP_INFIX = "__fk_";
+
+   /**
+    * Applies the sampled-preview row cap to a worksheet's DETAIL tables, leaving any lookup table
+    * injected purely to label a foreign key uncapped. Pass {@code maxRows <= 0} for full data, which
+    * clears a cap left by an earlier sampled render on the same runtime.
+    *
+    * <p>Bug #75989. {@code WorksheetInfo.setDesignMaxRows} is worksheet-WIDE, so it capped every table
+    * assembly independently — including the few-row lookup wiz injects to turn an FK id into a name.
+    * Truncating the lookup side of an INNER join does not sample the facts, it destroys matches: every
+    * fact row whose key was truncated away disappears. Measured on a 984-row table, same binding, only
+    * the cap changed: 8 → <b>0 rows</b>, 12 → 12 rows, 20 → 20 rows; the lookup ({@code projects}) has
+    * 13 rows, which is exactly where the behaviour turns.
+    *
+    * <p>What makes that a defect rather than a rough edge: the injection is only sound BECAUSE the join
+    * is row-preserving — wiz probes for orphans and duplicate target keys before injecting, precisely so
+    * that no unfiltered aggregate changes. A worksheet-wide cap silently voids that precondition. And
+    * because the join is injected automatically, a caller who wrote a single-table worksheet and never
+    * asked for a join could hit it.
+    *
+    * <p>Applying the cap per assembly keeps the documented semantics — "aggregate at most maxRows detail
+    * rows" — while exempting the dimension tables that were never part of the sampled fact stream.
+    */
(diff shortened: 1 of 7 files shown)


larryliang-inetsoft:
Status, since `claude-review` shows red: **it has failed twice for the same non-code reason** —
`error_max_turns`, i.e. the review agent hit its 50-turn budget and never reached a verdict.

```
"subtype": "error_max_turns", "is_error": true
##[error]Execution failed: Reached maximum number of turns (50)
```

Runs [93003239685](https://github.com/inetsoft-technology/stylebi/actions/runs/31220321847/job/93003239685)
and, after a re-run, [93005875264](https://github.com/inetsoft-technology/stylebi/actions/runs/31220321847/job/93005875264).
So it is an inconclusive review, not a rejection — nothing in it is a finding about this change. I re-ran
it once; a third attempt seems unlikely to help given it reproduced identically.

Every substantive gate is green: **`build` passes** (20m5s), Semgrep OSS, Semgrep Scan, and license/cla.
For context on scope, the change is small — 2 production files (`WorksheetTable.GroupByFieldInfo` gains an
`alias`; `applyAggregateInfo` honours it) plus 4 new unit tests.

Worth a human eye on two things in particular:

- **Backward compatibility.** An omitted alias must leave the rendered `Month(T.col)` name untouched so no
  existing caller's column names move. That is pinned by a test (`leavesTheRenderedNameAloneWhenNoAliasIsGiven`),
  but it is the property most worth a second opinion.
- **Where the alias is applied.** It goes on the PRIVATE `ColumnRef` — the `DateRangeRef` column for a
  date-grouped field, otherwise the private column of the same name — because the public selection is
  regenerated as clones of `privateCs`. This mirrors exactly what the aggregate branch a few lines below
  already does, and reuses the same target the group *description* is already resolved against.

Verified live on a locally built image (`local-1169`) beyond the unit tests: two aggregate mirrors grouped
by month with aliases, FULL-joined on those aliases, then `COALESCE(start_month, due_month)` as a
`sql: true` expression — renders 20 months with both series totalling 984, row-for-row identical to the
`sql: false` JS equivalent it replaces, with the coalesce pushed down to the database.

The paired wiz PR (inetsoft-technology/stylebi-wiz#1488) has merged. It degrades safely against a StyleBI
without this change — it detects the dropped alias and reports `groupAliasesNotApplied` rather than letting
a caller discover it as a downstream "column does not exist" — so there is no pressure to rush this one.

PR #4551, opened 2026-08-10
fix(wiz): validate a wiz-guessed geo map type against real data (bug-75983)

Zxm-happy:
## Summary

A "show sales by province" style prompt over real U.S. state data (e.g. a `STATE` column with values like "NJ"/"CA"/"NY") was rendering as a map of **Canada**. `wiz-services` guesses a chart's geo map type/layer from the prompt's wording (matching "province" to Canada's "Province" layer -- the only StyleBI layer literally named that), and `WizVsService.createGeoFields` wrote that guess into the chart unconditionally. StyleBI's own `MapHelper.autoDetect` is data-driven, but short-circuits whenever a recognized (not necessarily correct) map type is already set, so it never caught the bad guess either.

- `WizVsService.correctWizGeoMapping` (new, package-private) re-checks the guessed type/layer against the real executed data once it's available (`extractChartData`), and forces a genuine data-driven re-detection via `MapHelper.autoDetect` when every real value fails to match.
- If re-detection is itself inconclusive (real production data is messier than any single country's feature list, and `autoDetect` can settle on a null type), the original guess is restored instead of committing an unrenderable null-typed mapping -- verified live: an earlier version of this fix left the chart unable to render at all ("`STATE` layer is not supported by null") when re-detection couldn't confidently pick a type.
- Calls `MapHelper.autoDetect` directly rather than through `VSChartHandler.autoDetect`: that convenience wrapper re-resolves the geo ref via runtime-index alignment between `getRTGeoColumns()`/`getGeoColumns()`, which didn't line up on a real (post-execution) chart and silently no-opped -- verified live against the exact real production STATE value set from this bug.
- After a successful correction, re-runs the full `executeView` (not just `clearGraph`) so other runtime-synced structures the graph generator reads are refreshed consistently, then invalidates the cached graph so the corrected mapping actually renders.

## Test plan

- [x] New `WizVsServiceCorrectsWrongGeoMapTypeTest` (6 cases): wrong Canada/Province guess over clean U.S. state names re-detects to "U.S."; the exact 14 real STATE abbreviations from this bug's production log also re-detect to "U.S."; an already-correct guess is left untouched; an inconclusive re-detection (mocked via `MockedStatic<MapHelper>`) rolls back to the original guess instead of leaving a null type; non-map charts and a null `DataSet` are no-ops.
- [x] `mvn -pl core test -Dtest="inetsoft.web.wiz.service.*Test"` -- all 76 test files in the package pass, no regressions.
- [x] Verified live in a real local StyleBI instance against the actual bug scenario (real Orders/Customers join data, 14 distinct real `STATE` values) -- the map now renders as the U.S., not Canada.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
@@ -2624,7 +2624,7 @@ public class WizVsService {
             continue;
          }
 
-         int distinct = distinctValueCount(source, colIndex);
+         int distinct = MapHelper.distinctValueCount(source, colIndex);
 
          if(distinct == 0) {
             continue;
@@ -2638,10 +2638,11 @@ public class WizVsService {
             continue;
          }
 
-         List<String> allUnmatched = new ArrayList<>(unmatched.keySet());
+         List<String> unmatchedSample = unmatched.keySet().stream()
+            .limit(LOGGED_UNMATCHED_VALUES).collect(Collectors.toList());
          LOG.info("Wiz-guessed geo map type '{}' (layer {}) matched none of the {} real value(s) " +
-                  "for '{}' -- {}; clearing and re-detecting the map type from data.",
-                  mapping.getType(), mapping.getLayer(), distinct, refName, allUnmatched);
+                  "for '{}' -- sample: {}; clearing and re-detecting the map type from data.",
+                  mapping.getType(), mapping.getLayer(), distinct, refName, unmatchedSample);
          geoOption.setMapping(new FeatureMapping());
          MapHelper.autoDetect(vs, sourceInfo, chartInfo, geoOption, refName, source);
 
@@ -2656,9 +2657,9 @@ public class WizVsService {
 
          if(Tool.isEmptyString(redetectedType)) {
             LOG.warn("Re-detection of the geo map type for '{}' from real data was inconclusive " +
-                     "({} distinct real value(s): {}); keeping the original '{}' guess " +
+                     "({} distinct real value(s), sample: {}); keeping the original '{}' guess " +
                      "instead of leaving it unrenderable.",
-                     refName, distinct, allUnmatched, mapping.getType());
+                     refName, distinct, unmatchedSample, mapping.getType());
             geoOption.setMapping(mapping);
             continue;
          }
@@ -2669,20 +2670,9 @@ public class WizVsService {
       return corrected;
    }
 
-   /** Distinct number of non-null values in the given data column. */
-   private static int distinctValueCount(DataSet source, int colIndex) {
-      Set<String> distinct = new HashSet<>();
-
-      for(int r = 0; r < source.getRowCount(); r++) {
-         Object v = source.getData(colIndex, r);
-
-         if(!Tool.isEmptyString(Tool.toString(v))) {
-            distinct.add(Tool.toString(v));
-         }
-      }
-
-      return distinct.size();
-   }
+   /** Caps how many unmatched geo values a single log line spells out (columns like ZIP/city can
+    * have thousands of distinct bad values). The unmatched-count check itself uses the full set. */
+   private static final int LOGGED_UNMATCHED_VALUES = 20;
 
    /** Outcome of executing a chart assembly for verification: did it render data, and how many rows. */
    public record VerifyResult(boolean hasData, int rowCount) {}
--- a/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
@@ -2430,10 +2430,14 @@ public class WizVsService {
          // A wiz-guessed geo map type/layer (e.g. "province" in the prompt naively matched to
          // Canada's "Province" layer) is never validated against the real data at creation time.
          // Now that real data is available, re-check it and force a genuine re-detection if every
-         // real value failed to match -- then drop the stale graph so the re-fetch below (and any
-         // later image request) renders with the corrected mapping instead of the wrong guess.
+         // real value failed to match. Correcting only the design geo column (what autoDetect
+         // touches) leaves other runtime-synced structures the graph generator also reads (e.g.
+         // VSMapInfo's runtime geo fields) stale, so a full re-execution -- the same binding-sync
+         // pipeline a normal chart load runs -- is needed to render with the corrected mapping
+         // instead of the wrong guess (a partial refresh here previously left rendering broken).
          if(correctWizGeoMapping(rvs, assemblyName, dset)) {
             box.clearGraph(assemblyName);
+            box.executeView(assemblyName, true);
             pair = box.getVGraphPair(assemblyName, true);
             dset = pair == null ? null : unwrapDataSet(pair.getData());
 
@@ -2631,8 +2635,25 @@ public class WizVsService {
          LOG.info("Wiz-guessed geo map type '{}' (layer {}) matched none of the {} real value(s) " +
                   "for '{}'; clearing and re-detecting the map type from data.",
                   mapping.getType(), mapping.getLayer(), distinct, refName);
-         geoRef.getGeographicOption().setMapping(new FeatureMapping());
+         geoOption.setMapping(new FeatureMapping());
          chartHandler.autoDetect(vs, sourceInfo, chartInfo, refName, source);
+
+         // Real production data is messier than any one country's feature list -- autoDetect can
+         // fail to confidently settle on a type (e.g. mixed/unrecognized values) and leave the
+         // mapping's type null. Rendering a map with a null type throws ("X layer is not supported
+         // by null"), which is strictly worse than the original wrong-but-renderable guess. Only
+         // keep the correction if it actually produced a usable type; otherwise put the original
+         // guess back rather than leave the chart unable to render at all.
+         String redetectedType = geoOption.getMapping() == null ? null : geoOption.getMapping().getType();
+
+         if(Tool.isEmptyString(redetectedType)) {
+            LOG.warn("Re-detection of the geo map type for '{}' from real data was inconclusive; " +
+                     "keeping the original '{}' guess instead of leaving it unrenderable.",
+                     refName, mapping.getType());
+            geoOption.setMapping(mapping);
+            continue;
+         }
+
          corrected = true;
       }
 
--- a/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
@@ -2409,20 +2409,11 @@ public class WizVsService {
          }
 
          box = boxOpt.get();
-         VGraphPair pair = box.getVGraphPair(assemblyName, true);
-
-         if(pair == null) {
-            return new CreateViewsheetResult();
-         }
-
-         DataSet dset = pair.getData();
+         DataSet dset = fetchUnwrappedDataSet(box, assemblyName);
 
          if(dset == null) {
             return new CreateViewsheetResult();
          }
-
-         // Unwrap dataset wrappers to reach the aggregated data
-         dset = unwrapDataSet(dset);
 
          // A wiz-guessed geo map type/layer (e.g. "province" in the prompt naively matched to
          // Canada's "Province" layer) is never validated against the real data at creation time.
@@ -2435,8 +2426,7 @@ public class WizVsService {
          if(correctWizGeoMapping(rvs, assemblyName, dset)) {
             box.clearGraph(assemblyName);
             box.executeView(assemblyName, true);
-            pair = box.getVGraphPair(assemblyName, true);
-            dset = pair == null ? null : unwrapDataSet(pair.getData());
+            dset = fetchUnwrappedDataSet(box, assemblyName);
 
             if(dset == null) {
                return new CreateViewsheetResult();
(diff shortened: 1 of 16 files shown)


Zxm-happy (commented):
Reviewed the diff directly (author here, doing an independent pass as requested). Logic and test coverage look solid overall — the three-layer fix (validate against real data → roll back on inconclusive re-detection → call `MapHelper.autoDetect` directly instead of through `VSChartHandler`) is well-motivated and each layer has a corresponding regression test. Two things worth addressing before merge:

**1. Unbounded log line size for columns with many distinct unmatched values** (`WizVsService.java`, in `correctWizGeoMapping`)

Both the `LOG.info(...)` right before `MapHelper.autoDetect(...)` and the `LOG.warn(...)` in the inconclusive-rollback branch log `allUnmatched` — the *full* list of every distinct unmatched value, with no cap:

```java
List<String> allUnmatched = new ArrayList<>(unmatched.keySet());
LOG.info("... -- {}; ...", ..., allUnmatched);
...
LOG.warn("... ({} distinct real value(s): {}); ...", refName, distinct, allUnmatched, ...);
```

This is fine for the State/Province case (a handful of distinct values), but wiz can also create geo charts on higher-cardinality columns (ZIP codes, city names). If a column has hundreds or thousands of distinct bad values, both log lines will dump the entire list — twice, once at INFO and once at WARN if re-detection is inconclusive. Suggest capping to e.g. `allUnmatched.stream().limit(20).toList()` for the log messages (the full `unmatched` map can still be used for the actual `size()` check, which needs to stay exact).

**2. `distinctValueCount` duplicates existing logic** (`WizVsService.java`, new private method)

`WizGeoService.java` already has a private `distinctValueCount(DataSet, int)` with identical logic (distinct non-empty `Tool.toString` values). Not a functional problem since both are correct, but it's the second copy of the same helper — worth pulling into `MapHelper` (or a shared wiz util) as a public static method so the two classes don't drift if one gets a fix (e.g. a null-handling tweak) that the other doesn't. Not a blocker, just a follow-up-friendly cleanup.

Everything else — the rollback-on-null-type safety net, the full `executeView` re-run instead of just `clearGraph`, the `mapping.getLayer()`/`geoOption.getLayer()` sync fix, and the direct `MapHelper.autoDetect` call bypassing `VSChartHandler`'s RT-index resolution — is correct and each has a test that would have caught a regression if reverted (verified by checking the tests actually fail without the corresponding fix during development). No correctness issues found beyond the two notes above.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -2515,6 +2522,168 @@ else if(dset instanceof DataSetFilter) {
...
+         int distinct = distinctValueCount(source, colIndex);
+
+         if(distinct == 0) {
+            continue;
+         }
+
+         Map<String, Integer> unmatched =
+            MapHelper.getUnMatchedValues(source, colIndex, mapping, chartInfo);
+
+         if(unmatched.size() < distinct) {
```
Zxm-happy: **Correctness: `distinct` and `unmatched` are computed with inconsistent blank-value filtering, so an already-correct guess can be misjudged as "matched none."**

`distinctValueCount` (line 2679) explicitly filters out blank/null cells via `Tool.isEmptyString`, but `MapHelper.getUnMatchedValues` (called on line 2634) does not filter blanks in its row loop — a null/blank geo cell becomes its own unmatched `""` key whenever `""` is not already an auto/manual-matched feature (which it never is).

Concrete failure: a geo column has exactly one distinct non-blank real value that *does* match the guessed type, plus at least one row with a null/blank geo value. Then `distinct == 1` but `unmatched == {"": idx}` (size 1), so `unmatched.size() < distinct` is `1 < 1` → **false**. The code falls through to "matched none of the 1 real value(s)" (a misleading log, since the one real value did match), clears the mapping, and forces `MapHelper.autoDetect` to re-run — an unnecessary full re-execution (`box.clearGraph` + `box.executeView`) for a guess that was already correct, with a real risk of the single/low-cardinality re-detection landing on a *different, wrong* type instead of just restoring status quo (the rollback below only guards against a **null** redetected type, not a differently-wrong one).

None of the new tests in `WizVsServiceCorrectsWrongGeoMapTypeTest` include a blank/null row in the dataset, so this gap is unexercised.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -2515,6 +2522,168 @@ else if(dset instanceof DataSetFilter) {
...
+    * real data resolved correctly through {@code MapHelper.autoDetect} called directly, but through
+    * {@code VSChartHandler.autoDetect} it stayed cleared and this method's null-type rollback below
+    * kept firing).
+    *
+    * <p>Package-private (not private) so tests can call it directly without standing up a real
+    * {@link ViewsheetSandbox}, the same reason {@link #executeAndExtract} is.
+    *
+    * @return true if a correction was applied -- the caller must invalidate the cached graph.
+    */
+   boolean correctWizGeoMapping(RuntimeViewsheet rvs, String assemblyName, DataSet source) {
```
Zxm-happy, line 2574: **Altitude: nothing distinguishes "wiz's own unvalidated guess" from "a map type explicitly set afterward," so this correction re-fires forever and can silently undo a deliberate later choice.**

`correctWizGeoMapping` is wired into `extractChartData`, which every `executeAndExtract` call site reaches (chart creation, `apply_filter`, `change_chart_type`, `update_binding`, etc. — lines 413, 676, 1835, 2029, 2193). It treats *any* current `FeatureMapping` on a `MapInfo` the same way, with no provenance flag saying "this came from wiz's naive prompt-word guess (the actual bug)" vs. "this was just explicitly set."

Concrete scenario: wiz builds a correct U.S. states map, then the user (via a later chat turn/tool call) deliberately switches it to a different valid layer whose values do not textually match the currently-visible data (e.g. numeric/internal codes, or an intentional relabeling). The very next render through this same code path treats that explicit choice exactly like the original unvalidated guess, clears it, and re-detects — overwriting the user's explicit selection with only a server-side `LOG.info`/`LOG.warn`, no user-facing signal or opt-out.

Worth considering a one-shot "validated" marker set right after wiz's own `createGeoFields` guess, checked/cleared here, instead of re-validating unconditionally on every future render of the chart's lifetime.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -2421,18 +2422,24 @@ private CreateViewsheetResult extractChartData(RuntimeViewsheet rvs, String asse
...
+
+         // A wiz-guessed geo map type/layer (e.g. "province" in the prompt naively matched to
+         // Canada's "Province" layer) is never validated against the real data at creation time.
+         // Now that real data is available, re-check it and force a genuine re-detection if every
+         // real value failed to match. Correcting only the design geo column (what autoDetect
+         // touches) leaves other runtime-synced structures the graph generator also reads (e.g.
+         // VSMapInfo's runtime geo fields) stale, so a full re-execution -- the same binding-sync
+         // pipeline a normal chart load runs -- is needed to render with the corrected mapping
+         // instead of the wrong guess (a partial refresh here previously left rendering broken).
+         if(correctWizGeoMapping(rvs, assemblyName, dset)) {
```
Zxm-happy, line 2426: **Efficiency: the geo-mapping validation (full column scan + `MapHelper.getUnMatchedValues` feature matching) runs unconditionally on every single chart render, forever, with no way to skip it once a mapping has already been confirmed correct.**

Since `extractChartData` is reached from every `executeAndExtract` call site (creation, `apply_filter`, `change_chart_type`, `update_binding`, ...), a map chart that was correctly guessed on its very first render still pays the full `distinctValueCount` + `MapHelper.getUnMatchedValues` cost on every subsequent interaction for the lifetime of the runtime viewsheet, since nothing records "this mapping already passed validation." For large result sets this is a repeated full-column scan plus feature-table matching on every render that will never change outcome.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -2421,18 +2422,24 @@ private CreateViewsheetResult extractChartData(RuntimeViewsheet rvs, String asse
...
+            box.clearGraph(assemblyName);
+            box.executeView(assemblyName, true);
+            pair = box.getVGraphPair(assemblyName, true);
+            dset = pair == null ? null : unwrapDataSet(pair.getData());
```
Zxm-happy: Simplification: the fetch-pair -> get DataSet -> unwrap -> null-check sequence is now written out twice in this method (once above at chart-execution entry, once here after correctWizGeoMapping triggers a re-execution). A small private helper such as fetchUnwrappedDataSet(ViewsheetSandbox box, String assemblyName) returning the unwrapped DataSet or null would remove the duplication and make the this pipeline runs twice intent explicit in code rather than only in the surrounding comment.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
Zxm-happy: Fixed in 2668834ad — filtered blank/null values out of `unmatched` before comparing against `distinct` (and before logging/sampling). Added a regression test (`leavesAnAlreadyCorrectGuessAloneWhenDataHasABlankRow`) that fails against the prior code and passes now.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
Zxm-happy: Fixed in 2668834ad — extracted `fetchUnwrappedDataSet(box, assemblyName)` and used it for both fetches in `extractChartData`, plus `collectColorValues`'s identical inline sequence. Also converted `verifyChartData`'s duplicate unwrap loop to the existing `unwrapDataSet` helper (it needed to keep the pre-unwrap `dset` reference separately for its row count, so it couldn't use the full `fetchUnwrappedDataSet` wrapper, just the unwrap step).

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
Zxm-happy: Valid concern, and real: a deliberate later choice made through some path other than geo_apply (which is the only place that populates per-value mappings today) would still get silently re-litigated here.

Addressed the concrete case this maps to in 2668834ad: a non-empty `mapping.getMappings()` (built by `WizGeoService.apply` -> `FeatureMapping.addMapping`) now short-circuits the whole check -- wiz's own `createGeoFields` guess always starts with an empty mapping, so this cleanly distinguishes "unvalidated wiz guess" from "a person went through geo_apply," without needing a new provenance flag.

Didn't go further than that (e.g. a boolean "user-confirmed" flag independent of whether any per-value overrides were needed) -- that's a real gap if someone picks a valid-but-mismatched type/layer with zero per-value overrides, but it's a bigger architectural addition than this bug fix's scope, and there's no current tool path that produces that exact state. Flagging it as a known follow-up rather than papering over it.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
Zxm-happy: Same root cause as the provenance comment below, and the same fix in 2668834ad covers the concrete cost this describes for the common case: once a mapping has explicit per-value entries (built via geo_apply), the whole scan/match is skipped on every subsequent render instead of re-paying it every time.

It does NOT help the "wiz auto-detected the correct type, no per-value overrides needed" case -- that one still re-validates on every render, since its getMappings() stays empty by design (auto-detected mappings resolve values algorithmically, not via a per-value dictionary -- see MapHelper.getAutoMapping construction of "new FeatureMapping(id, algorithm, type, layer)", no addMapping calls). Caching "already validated, skip" for that case would need a separate persisted flag and is a bigger change than this fix's scope; the absolute cost per render (one column scan + name-table match) is small relative to the query execution extractChartData already does, so leaving it as a known, low-severity follow-up rather than adding that machinery here.

Zxm-happy (commented):
## Round-2 review — all findings addressed

Both the two I found in my own pass and the four from a separate independent background review pass are now handled (commits `2668834ad` and `d9bfcd5d8`):

- **Correctness (real bug, confirmed via TDD red/green):** `distinctValueCount` skips blank/null cells but `MapHelper.getUnMatchedValues` doesn't, so a single blank geo row could make an already-correct guess look fully unmatched and trigger a needless (and, rarely, harmful) re-detection. Fixed by filtering blank values out of the unmatched set before the comparison. New test `leavesAnAlreadyCorrectGuessAloneWhenDataHasABlankRow` fails against the prior code, passes now.
- **Altitude / provenance concern:** added a guard that skips the whole re-validation when the mapping already has explicit per-value entries (`FeatureMapping.getMappings()` non-empty) — that only happens via a real person going through `geo_apply`, never via wiz's own raw `createGeoFields` guess. This directly protects a deliberately-chosen, textually-non-matching mapping from being silently overwritten, without inventing a new provenance-tracking mechanism. Replied inline on why a fuller "user-confirmed" flag is out of scope for this bug fix. New test `respectsAnExplicitlyMappedGuessEvenIfItDoesNotTextuallyMatch` fails against the prior code, passes now.
- **Efficiency (unconditional re-validation cost):** the same empty-mappings guard above eliminates the repeat cost for any chart a person has explicitly mapped; the remaining case (an auto-detected, never-manually-touched correct guess re-validating on every render) is a real but low-severity cost relative to the query execution the method already does each time — replied inline with the reasoning for leaving that as a follow-up rather than adding a persisted "already validated" flag here.
- **Reuse/simplification (duplicated fetch-unwrap-null-check):** extracted `fetchUnwrappedDataSet(box, assemblyName)`, used by both fetches in `extractChartData` and by `collectColorValues`'s identical sequence; also converted `verifyChartData`'s separate identical inline unwrap loop to the existing `unwrapDataSet` helper (it needs to keep the pre-unwrap `dset` reference for its row count, so it uses the smaller helper, not the full fetch wrapper).
- Also carried over my own round-1 findings from before: unmatched-values log lines capped to 20 samples, and `distinctValueCount` extracted into `MapHelper` as a shared public method (removed the identical private copy in `WizGeoService`).

`wiz.service` package re-run clean (607 tests) except one pre-existing, unrelated stray failure from a stale compiled test class left over from a different bug's checkout sharing this working tree (`WizVsServiceCollectChartFlatBindingMultiStyleTest` — its `.java` source doesn't exist on this branch at all).

fredaan (approved):

PR #4552, opened 2026-08-10
fix(embed): include Roboto webfonts in the embed npm package (bug-75971)

Zxm-happy:
## Summary
- `elements.css`/`viewer-element.css` declare `@font-face` rules pointing at `assets/roboto/*.woff2|woff`, but `build-embed-package.mjs`'s `resolveAssetsDir()` only ever sources the `assets/` folder from the Angular "elements" project output (or `projects/portal/src/assets`) — neither of which has ever contained a `roboto` directory. The Roboto webfonts are only staged by the separate `fonts:stage` gulp task into the Gulp app output, alongside the CSS/JS this script already copies from there.
- Result: the published `@inetsoft-technology/stylebi-embed-elements` package has been silently shipping without any Roboto fonts. Every consumer's browser requests `assets/roboto/*.woff2` and gets back whatever the dev server falls back to for an unknown path (e.g. `index.html` for a Vite dev server), producing "Failed to decode downloaded font" / `OTS parsing error: invalid sfntVersion` console errors.
- Fix: copy the Gulp-staged `assets/roboto` directory into the package output alongside the existing assets, and fail the build loudly (`ensureRobotoFontsExist()`) if that directory is missing or empty instead of silently shipping an incomplete package.

## Test plan
- [x] Ran `npm run package:embed:only` against existing build artifacts; confirmed `target/npm/stylebi-embed-elements/assets/roboto/` now contains all 23 Roboto woff/woff2 files, including `Roboto-Regular.woff2`, `Roboto-Regular.woff`, `Roboto-Bold.woff2`, `Roboto-Bold.woff` (the exact files seen failing in a consumer's browser console).
- [x] Confirmed existing non-font assets (chart type icons, etc.) are still copied and unaffected.
- [x] Simulated a missing/un-staged fonts directory and confirmed the build now fails loudly with a clear message instead of silently producing an incomplete package.
- [ ] Full `npm run package:embed:prod` / `publish:embed` run and downstream consumer (wiz) verification after a version bump — not run here since it requires a full rebuild + publish.

### DIFF
--- a/web/scripts/build-embed-package.mjs
+++ b/web/scripts/build-embed-package.mjs
@@ -24,6 +24,12 @@ export const GENERATED_RESOURCES_DIR = path.join(PROJECT_ROOT, "target", "genera
 const GULP_APP_DIR = path.join(GENERATED_RESOURCES_DIR, "gulp", "inetsoft", "web", "resources", "app");
 const NG_ELEMENTS_ASSETS_DIR = path.join(GENERATED_RESOURCES_DIR, "ng", "inetsoft", "web", "resources", "elements", "assets");
 const FALLBACK_ASSETS_DIR = path.join(PROJECT_ROOT, "projects", "portal", "src", "assets");
+// elements.css/viewer-element.css (copied from GULP_APP_DIR below) declare @font-face rules
+// pointing at "assets/roboto/*". The "fonts:stage" gulp task is what actually populates that
+// folder (from node_modules/roboto-fontface) -- it lands in the Gulp app output, not in the
+// Angular "elements" project assets or projects/portal/src/assets, so resolveAssetsDir() never
+// picks it up on its own and the shipped package silently ships without any Roboto webfonts.
+const GULP_ROBOTO_ASSETS_DIR = path.join(GULP_APP_DIR, "assets", "roboto");
 
 // Configuration
 const verbose = process.env.VERBOSE === "true" || process.env.VERBOSE === "1";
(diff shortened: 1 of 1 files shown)


Zxm-happy (commented):
## Review

Small, well-scoped fix — read through the diff and the reasoning in the PR description.

**Correctness**
- `GULP_ROBOTO_ASSETS_DIR` matches exactly where `gulp/fonts.js`'s `fonts:stage` task stages the Roboto webfonts (`node_modules/roboto-fontface/fonts/roboto/*` → `target/generated-resources/gulp/inetsoft/web/resources/app/assets/roboto`) — the same output dir `elements.js`/`elements.css` are already copied from a few lines up, so the CSS's relative `assets/roboto/*` reference and this copy are guaranteed to line up.
- `ensureRobotoFontsExist()` correctly fails loud (`existsSync` + `readdirSync(...).length === 0`) instead of silently shipping a package with dangling `@font-face` references — matches the existing `ensureFilesExist` pattern used for the JS/CSS artifacts just above it.
- Copy ordering (main `assets/` first, then the Roboto folder merged in) is safe: `projects/portal/src/assets` has never had a `roboto` subdirectory, so there's no overwrite race, and even if it did, the Gulp-staged copy winning is the correct outcome since it's the one the CSS was actually built against.
- Manually verified per the PR description (`package:embed:only` output inspected, missing-dir failure path simulated) — I re-ran both locally and got the same result: all 23 Roboto files present, build fails loud when the fonts directory is absent/empty, no regression to the existing non-font assets.

**Process note, not a code issue**: the `claude-review` check shows red, but both bot comments on this PR are just `"Claude encountered an error after 0s ... I'll analyze this and get back to you"` — the automated review crashed before producing any actual findings (job step `Run Claude Code Review` failed outright). Worth knowing before treating that red X as a real review signal.

**Non-blocking suggestion**: `web/scripts/*.mjs` has no test coverage at all in this repo today (`npm test` only runs the portal/elements suites), so this fix's verification is manual-only, matching existing practice for this directory. A small Jest/Vitest test around `ensureRobotoFontsExist`/`buildEmbedPackage` would be nice to have if this class of script ever gets a test harness, but I wouldn't block on it given there's no precedent here.

No blocking issues found — looks good to merge once a human has actually reviewed it (the automated check above did not).

fredaan (approved):

PR #4553, opened 2026-08-10
fix Bug #75995: should clear condition when send an empty condition model in filter

lizhepu:
- applyConditionModel returned early whenever baseConditions was empty, treating it exactly like an absent list, so "delete this filter" applied nothing and still reported success
- Move the boundary to the model itself: an absent conditionModel leaves the assembly untouched, a present one IS the complete new condition state, and an empty list inside it clears - the same reading hasAggregateConditions() already takes. The underlying setPreConditionList0 has always normalized an empty list to null, so the capability was there; only the wiz-layer guard stood in the way
- Apply the same rule in applyConditionsToWorksheet, or an aggregate (HAVING) condition would stay unremovable while a base (WHERE) one became removable
- Add canApplyConditions: a non-empty request that compiles to nothing means every node was unusable (unknown field, malformed spec), so keep the existing conditions and warn instead of clearing what was never meant to be touched

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
@@ -4385,9 +4385,13 @@ public class WizVsService {
     * Only base conditions are applied here; aggregate conditions require pushing to worksheet.
     */
    private void applyConditionModel(VSAssembly assembly, VisualizationConditionModel conditionModel) {
-      if(conditionModel == null || conditionModel.getBaseConditions() == null ||
-         conditionModel.getBaseConditions().isEmpty())
-      {
+      // The model itself is the boundary: absent means the request said nothing about conditions and
+      // whatever the assembly carries must be left alone, present means it IS the complete new
+      // condition state. Within a present model an empty list and a null list say the same thing —
+      // "no conditions" — matching hasAggregateConditions(), and both must therefore CLEAR. Treating
+      // an empty list as "nothing to do" is what made "delete this filter" a silent no-op that still
+      // reported success.
+      if(conditionModel == null) {
          return;
       }
 
@@ -4399,7 +4403,15 @@ public class WizVsService {
       // Resolve the assembly's candidate condition fields so condition values can be coerced
       // to each field's data type (see buildConditionItem).
       ColumnSelection columns = VSUtil.getBaseColumns(dataAssembly, true);
-      dataAssembly.setPreConditionList(buildConditionList(conditionModel, columns));
+      ConditionList conditions = buildConditionList(conditionModel, columns);
+
+      if(!canApplyConditions(conditions, conditionModel.getBaseConditions(), "base")) {
+         return;
+      }
+
+      // An empty ConditionList is normalized to null by setPreConditionList0, which is what clears the
+      // filter and flags INPUT_DATA_CHANGED so the chart re-executes.
+      dataAssembly.setPreConditionList(conditions);
    }
 
    /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -4890,32 +4902,64 @@ private void applyConditionsToWorksheet(
...
-         appendConditionNodes(conditionModel.getBaseConditions(), 0, preCondList,
-                              new boolean[]{ true }, null, columns);
+      if(base != null) {
+         appendConditionNodes(base, 0, preCondList, new boolean[]{ true }, null, columns);
+      }
 
-         if(!preCondList.isEmpty()) {
-            table.setPreConditionList(preCondList);
-         }
+      if(canApplyConditions(preCondList, base, "base")) {
```
lizhepu, line 4922: **An absent base half now wipes worksheet WHERE conditions this model never owned.**

Old code skipped the whole block when `base` was null, leaving `table`'s existing pre-conditions alone. Now `canApplyConditions(empty, null, …)` returns true on the `requested == null` arm, so `setPreConditionList` is called with an empty list and clears them.

That matters because of how this method is reached: only from `pushAggregationToWorksheet`, gated on `hasAggregateConditions` (line 1673), and `table` is an **existing** assembly looked up by name (`ws.getAssembly(wsTableName)`) — not one this request built. It can already carry a `preConditionList` from `create_worksheet_table`'s `preAggregateCondition`, which the chart-level model has no knowledge of.

Concrete case: a HAVING-only request ("只保留销售额超过 1000 的类别") against a worksheet table built with `WHERE PAID = 1`. `base` is null, the WHERE is silently dropped, and the chart quietly widens to include unpaid rows.

The PR's premise — "the model IS the complete new condition state" — holds for the VS assembly, whose conditions this layer does own. It does not hold for a shared worksheet table.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -4890,32 +4902,64 @@ private void applyConditionsToWorksheet(
...
+    *
+    * @param compiled  the condition list built from {@code requested}
+    * @param requested the condition nodes the request carried; null when it carried none
+    * @param label     which half of the model this is, for the log line
+    */
+   private boolean canApplyConditions(ConditionList compiled,
+                                      List<VisualizationConditionModel.ConditionNode> requested,
+                                      String label)
+   {
+      if(!compiled.isEmpty() || requested == null || requested.isEmpty()) {
```
lizhepu, line 4960: **Only total compile failure is caught; a partial one is written silently.**

`appendConditionNodes` drops nodes individually — `continue` on a null node, on an empty field spec, on empty `items`. So three requested conditions can compile to one, `!compiled.isEmpty()` is true, and the partial list is written with no warning.

The consequence is the same class this guard was written against: losing one leaf of an AND chain widens the result, and the caller is told it succeeded. The doc above argues an all-unusable write "would clear conditions the request never asked to touch" — a partially-unusable write relaxes them instead, which is just as invisible.

At minimum this deserves the same log line when `compiled.size() < requested.size()`, so a dropped node is traceable rather than inferred from a wrong number.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -4385,9 +4385,13 @@ private VSAssembly findPrimaryAssembly(Viewsheet vs) {
...
-      if(conditionModel == null || conditionModel.getBaseConditions() == null ||
-         conditionModel.getBaseConditions().isEmpty())
-      {
+      // The model itself is the boundary: absent means the request said nothing about conditions and
+      // whatever the assembly carries must be left alone, present means it IS the complete new
+      // condition state. Within a present model an empty list and a null list say the same thing —
+      // "no conditions" — matching hasAggregateConditions(), and both must therefore CLEAR. Treating
+      // an empty list as "nothing to do" is what made "delete this filter" a silent no-op that still
+      // reported success.
+      if(conditionModel == null) {
```
lizhepu, line 4394: **This makes a default-initialized model destructive for every caller, not just the delete path.**

Presence of the object is now the whole signal, so `{ baseConditions: [], aggregateConditions: [] }` — the shape any caller gets from a default-constructed model rather than a deliberate "remove the filter" — clears the assembly where it used to be inert. The old guard absorbed that shape harmlessly, which is exactly why nobody had to be careful about it.

`conditionModel` comes straight off the request (`model.getConditionModel()`, line 1654) with no further qualification, and `apply_filter` is a public tool, so the contract change reaches callers outside this repo too.

Worth confirming that every producer omits the field rather than sending an empty one — otherwise an unrelated edit that happens to carry an empty model now silently drops the user's filter, which is the same silent-success failure this PR is fixing, in the opposite direction.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
lizhepu: No change — on a closer read this is not introduced here. `setPreConditionList` already replaced the table's whole list whenever `base` was non-empty, so a QB-built WHERE was being dropped before this PR too; the empty case just makes it uniform. Distinguishing a chart-pushed WHERE from a worksheet-owned one needs an ownership marker this layer does not have, which is a bigger change than this PR. Leaving it as a note.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
lizhepu: Fixed in 6fe8bb697.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
lizhepu: Checked the wiz side — no change needed. `stagedApply` gates FILTER on `hasParams: !!o?.conditionModel`, so the field is present only when the condition agent produced one, and `createViewsheet` passes `conditionModel: undefined` explicitly. No producer sends a default-initialized model.

fredaan (approved):

PR #4554, opened 2026-08-10
fix(embed): pie/donut dimension lost on chart-type re-slot (bug-75993)

Zxm-happy:
## Summary
- `DonutChartFilter` builds a pie/donut chart with `info.setMultiStyles(true)` and binds the slice dimension via the Y measure's **own** `setColorField` (a per-`ChartAggregateRef` aesthetic) rather than the chart-info-level `VSChartInfo.setColorField`.
- `WizVsService.collectChartFlatBinding` only read the chart-info-level aesthetics, so that dimension was invisible to every caller re-deriving fields from a pie/donut source.
- Concretely: a wiz chart-type change from pie → crosstab (`WizAutoBindingService.changeType`'s `deriveFieldConfigs(targetAssembly)` → `collectFlatBinding` → `collectChartFlatBinding`) silently dropped the only dimension and rebuilt the crosstab with a measures-only binding. Every downstream re-slot (crosstab → donut, etc.) then inherited that dimension-less binding and rendered a completely blank panel.
- Fix: also walk each `ChartAggregateRef`'s own color/shape/size/text aesthetic refs, mirroring the existing chart-info-level aesthetic loop.

## Test plan
- [x] Added `WizVsServiceCollectChartFlatBindingMultiStyleTest`: builds a `VSChartInfo` with `multiStyles=true` and a dimension bound via the Y measure's own `colorField` (the exact `DonutChartFilter` pattern) and asserts `collectFlatBinding` reads it.
- [x] Confirmed the test fails without the fix (`expected: <1> but was: <0>`) and passes with it.
- [x] `mvn -o test` for this test plus the neighboring `WizVsServiceClassifyChartRefSecondaryYTest` (same method, unrelated prior fix) — both pass, no regression.
- [x] `mvn -o compile` for the module — clean.
- [x] Verified live end-to-end against a real StyleBI instance: pie → crosstab → donut chain now correctly carries the dimension through and paints the donut chart's slices instead of a blank panel.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
@@ -2809,6 +2809,27 @@ public class WizVsService {
          }
       }
 
+      // Per-measure aesthetic refs. A "multi-style" binding (pie/donut: see DonutChartFilter,
+      // which sets info.setMultiStyles(true) and calls the Y measure's OWN setColorField rather
+      // than the chart-info-level one) carries its category dimension on the individual
+      // ChartAggregateRef, not on VSChartInfo above — that dimension was invisible here, so
+      // deriving fields from a pie/donut source (a chart-type change that re-binds via this
+      // method, e.g. re-slotting into a crosstab) silently dropped it and produced a
+      // dimension-less rebuild.
+      for(ChartRef ref : info.getBindingRefs(false)) {
+         if(ref instanceof ChartAggregateRef aggRef) {
+            for(AestheticRef aref : new AestheticRef[]{
+               aggRef.getColorField(), aggRef.getShapeField(),
+               aggRef.getSizeField(), aggRef.getTextField()
+            })
+            {
+               if(aref != null) {
+                  classifyChartRef(aref.getDataRef(), dimensions, measures, seen);
+               }
+            }
+         }
+      }
+
       // Path field
       ChartRef pathField = info.getPathField();
       if(pathField != null) {
--- a/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
@@ -2815,18 +2815,13 @@ public class WizVsService {
       // ChartAggregateRef, not on VSChartInfo above — that dimension was invisible here, so
       // deriving fields from a pie/donut source (a chart-type change that re-binds via this
       // method, e.g. re-slotting into a crosstab) silently dropped it and produced a
-      // dimension-less rebuild.
-      for(ChartRef ref : info.getBindingRefs(false)) {
-         if(ref instanceof ChartAggregateRef aggRef) {
-            for(AestheticRef aref : new AestheticRef[]{
-               aggRef.getColorField(), aggRef.getShapeField(),
-               aggRef.getSizeField(), aggRef.getTextField()
-            }) {
-               if(aref != null) {
-                  classifyChartRef(aref.getDataRef(), dimensions, measures, seen);
-               }
-            }
-         }
+      // dimension-less rebuild. Delegate to getAggregateAestheticRefs(), which already gates on
+      // isMultiAesthetic() -- reading every aggregate's color/shape/size/text unconditionally
+      // would resurface a stale colorField that ChangeSeparateStatusProcessor leaves behind on
+      // non-first aggregates when a chart is switched OUT of multi-styles, reintroducing a
+      // phantom dimension.
+      for(AestheticRef aref : info.getAggregateAestheticRefs(false)) {
+         classifyChartRef(aref.getDataRef(), dimensions, measures, seen);
       }
 
       // Path field
--- a/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
@@ -2821,8 +2821,7 @@ public class WizVsService {
             for(AestheticRef aref : new AestheticRef[]{
                aggRef.getColorField(), aggRef.getShapeField(),
                aggRef.getSizeField(), aggRef.getTextField()
-            })
-            {
+            }) {
                if(aref != null) {
                   classifyChartRef(aref.getDataRef(), dimensions, measures, seen);
                }
--- a/core/src/test/java/inetsoft/web/wiz/service/WizVsServiceCollectChartFlatBindingMultiStyleTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/service/WizVsServiceCollectChartFlatBindingMultiStyleTest.java
@@ -91,4 +91,42 @@ class WizVsServiceCollectChartFlatBindingMultiStyleTest {
       assertEquals(1, dimensions.size());
       assertEquals("PRODUCT_NAME", dimensions.get(0).getField());
    }
+
+   /**
+    * Regression for the fix itself: {@link VSChartInfo#getAggregateAestheticRefs} gates on
+    * {@code isMultiAesthetic()}, so a stale per-aggregate {@code colorField} left over from a
+    * chart that used to be multi-styles (see {@code ChangeSeparateStatusProcessor}, which clears
+    * the aesthetic only on the first aggregate when "separate" is turned off) must NOT resurface
+    * as a phantom dimension once {@code multiStyles} is false.
+    */
+   @Test
+   void collectFlatBindingIgnoresAStaleAggregateColorFieldWhenNotMultiStyles() {
+      WizVsService service = new WizVsService(
+         mock(ViewsheetService.class), mock(AssetRepository.class), mock(SecurityEngine.class), null, null);
+
+      VSChartDimensionRef productName = new VSChartDimensionRef();
+      productName.setGroupColumnValue("PRODUCT_NAME");
+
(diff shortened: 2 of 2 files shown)


core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -2809,6 +2809,26 @@ private CreateViewsheetResult.FlatBinding collectChartFlatBinding(VSChartInfo in
...
 
+      // Per-measure aesthetic refs. A "multi-style" binding (pie/donut: see DonutChartFilter,
+      // which sets info.setMultiStyles(true) and calls the Y measure's OWN setColorField rather
+      // than the chart-info-level one) carries its category dimension on the individual
+      // ChartAggregateRef, not on VSChartInfo above — that dimension was invisible here, so
+      // deriving fields from a pie/donut source (a chart-type change that re-binds via this
+      // method, e.g. re-slotting into a crosstab) silently dropped it and produced a
+      // dimension-less rebuild.
+      for(ChartRef ref : info.getBindingRefs(false)) {
+         if(ref instanceof ChartAggregateRef aggRef) {
```
Zxm-happy: **Missing `isMultiStyles()`/`isMultiAesthetic()` gate — can read stale per-aggregate aesthetics on non-multi-style charts.**

Every other place in the codebase that reads a `ChartAggregateRef`'s own color/shape/size/text fields gates on `info.isMultiAesthetic()` first (e.g. `AbstractChartInfo.getAggregateAestheticRefs(runtime)` at line 261: `if(!isMultiAesthetic()) { return list; }`, and `getAestheticRefs(boolean runtime)` at line 241). That's because these per-aggregate fields are only the *active* color source when `multiStyles` is true; when it's false the chart-info-level aesthetic (already read a few lines above) is authoritative, and the per-aggregate fields can be leftover state.

Concretely, `ChangeSeparateStatusProcessor.process()` (core/src/main/java/inetsoft/report/internal/graph/ChangeSeparateStatusProcessor.java, lines 82-143) turns `multiStyles` off without clearing `colorField`/`shapeField`/`sizeField` on the aggregate refs — it only copies the *first* aggregate's legend descriptor up to the plot level and `break`s (line 139). So a chart that once had per-measure color aesthetics (e.g. a combo/donut-style binding) and then had "separate"/multi-style toggled off can still carry a live `colorField` object on its `ChartAggregateRef`, even though `info.isMultiStyles()` is now false and the renderer no longer treats it as bound.

With this new loop being unconditional, `collectChartFlatBinding` will pick up that stale dimension and report it as part of the binding, and `WizAutoBindingService.deriveFieldConfigs` will feed it into rebuilding the next chart type — reintroducing a phantom dimension into a chart that shouldn't have it. That's the same class of "invisible aesthetic state produces a wrong binding" bug this PR fixes, just in the opposite direction.

Suggest gating this loop (and ideally replacing it entirely) with `info.isMultiStyles()`, or better, reusing `info.getAggregateAestheticRefs(false)` / `AbstractChartInfo.getAestheticRefs(ChartBindable)` which already implement this correctly (see next comment).

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -2809,6 +2809,26 @@ private CreateViewsheetResult.FlatBinding collectChartFlatBinding(VSChartInfo in
...
+            for(AestheticRef aref : new AestheticRef[]{
```
Zxm-happy: **Reuse: this duplicates `AbstractChartInfo.getAggregateAestheticRefs(boolean)` / `getAestheticRefs(ChartBindable)`.**

`AbstractChartInfo` already has exactly this "get each aggregate's own color/shape/size/text aesthetics" logic:

```java
public List<AestheticRef> getAggregateAestheticRefs(boolean runtime) {
   List<AestheticRef> list = new ArrayList<>();
   if(!isMultiAesthetic()) {
      return list;
   }
   for(ChartAggregateRef ref : getAestheticAggregateRefs(runtime)) {
      list.addAll(getAestheticRefs(ref));
   }
   return list;
}
```

Calling `info.getAggregateAestheticRefs(false)` here instead of hand-rolling the loop would:
1. Automatically apply the `isMultiAesthetic()` gate this new code is missing (see previous comment).
2. Dedup Y-measure refs by full name via `getModelRefsY`/`getAestheticAggregateRefs`, instead of re-scanning all binding refs.
3. Pick up the `RelationChartInfo` per-aggregate `nodeColorField`/`nodeSizeField` that `AbstractChartInfo.getAestheticRefs(ChartBindable)` already handles (lines ~275-292) but this new hand-rolled loop silently omits.

Worth using the existing helper rather than maintaining a second, slightly-divergent copy of the same logic.

core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
```diff
@@ -2809,6 +2809,26 @@ private CreateViewsheetResult.FlatBinding collectChartFlatBinding(VSChartInfo in
...
       }
 
+      // Per-measure aesthetic refs. A "multi-style" binding (pie/donut: see DonutChartFilter,
+      // which sets info.setMultiStyles(true) and calls the Y measure's OWN setColorField rather
+      // than the chart-info-level one) carries its category dimension on the individual
+      // ChartAggregateRef, not on VSChartInfo above — that dimension was invisible here, so
+      // deriving fields from a pie/donut source (a chart-type change that re-binds via this
+      // method, e.g. re-slotting into a crosstab) silently dropped it and produced a
+      // dimension-less rebuild.
+      for(ChartRef ref : info.getBindingRefs(false)) {
```
Zxm-happy: **Minor efficiency: `info.getBindingRefs(false)` is now invoked twice** (once here, once at the top of the method for the "Main binding: x, y, group" loop). Several `ChartInfo` implementations rebuild the array/list on every call — e.g. `VSMapInfo.getBindingRefs` allocates a new `ArrayList` and copies `super.getBindingRefs(runtime)` into it each time. Since this new per-aggregate-aesthetic pass only needs `ChartAggregateRef` instances that are already visited in the first loop, it could be folded into that same loop (classify the ref, then check `instanceof ChartAggregateRef` and harvest its aesthetics) instead of iterating the binding refs array a second time.

core/src/test/java/inetsoft/web/wiz/service/WizVsServiceCollectChartFlatBindingMultiStyleTest.java
```diff
@@ -0,0 +1,94 @@
...
+ * rebuild silently lost its only dimension and produced a measures-only binding that pie/donut/
+ * crosstab cannot render — the empty-panel bug reported for "change to crosstab, then to donut."
+ */
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = { BaseTestConfiguration.class }, initializers = ConfigurationContextInitializer.class)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
+@SreeHome
+@Tag("core")
+class WizVsServiceCollectChartFlatBindingMultiStyleTest {
+   @Test
```
Zxm-happy, line 63: The added test only covers `multiStyles == true`. Given the missing `isMultiStyles()` gate flagged on the main change (WizVsService.java, the new per-aggregate loop), please also add a case with `info.setMultiStyles(false)` and a `ChartAggregateRef` that still carries a non-null `colorField` (simulating the leftover state left behind by `ChangeSeparateStatusProcessor` when "separate"/multi is turned off without clearing per-aggregate aesthetics) and assert that dimension is NOT included. As written, that regression would slip through CI even after the gate is added — or removed by mistake later.

Zxm-happy:
已在 37e6e7a5 修复本次评审中最严重的一条发现（缺少 `isMultiAesthetic()` 门控导致的潜在“幽灵维度”回归）：

- 将手写的“逐个 `ChartAggregateRef` 读取自身 color/shape/size/text”循环，改为直接复用 `VSChartInfo.getAggregateAestheticRefs(false)`。该方法内部已经先判断 `isMultiAesthetic()` 才会读取每个 aggregate 自身的美学字段，避免了 `ChangeSeparateStatusProcessor` 关闭多样式时只清理第一个 aggregate、导致其余 aggregate 残留旧 `colorField` 被误判为“当前维度”的问题。
- 顺带解决了 `getBindingRefs(false)` 被重复调用两次的效率点——现在方法体内只调用一次。
- 新增回归测试 `WizVsServiceCollectChartFlatBindingMultiStyleTest#collectFlatBindingIgnoresAStaleAggregateColorFieldWhenNotMultiStyles`，覆盖“`multiStyles=false` 但 aggregate 上仍残留 `colorField`”场景，断言不再产生幽灵维度。

本地已跑通该文件相关的全部单测（`WizVsServiceCollectChartFlatBindingMultiStyleTest` 2/2，以及同目录下 `BindingSlotsTest`/`WizAutoBindingServiceChangeTypeFieldFormulaTest`/`WizFieldInfoFactorySortOrderTest`/`WizVsServiceApplyHighlightCopyTest`/`WizVsServiceClassifyChartRefSecondaryYTest`/`WizVsServiceFilterCopyTest`/`WizVsServiceRebindClearsStaleRuntimeChartRefsTest`/`WizVsServiceSecondaryYAxisRenderTest` 均通过），未见回归。

RelationChartInfo 的节点 color/size 字段在下方已有独立处理，未受影响。

fredaan (approved):

PR #4555, opened 2026-08-10
Bug #75823, fix NPE regression in range slider fix for cube aggregate measures

ngovus-inetsoft:
## Summary

Follow-up to #4478 (Bug #75823), which fixed a RangeSlider not showing its minimum value when bound to a cube (XMLA) aggregate measure — the fix in that PR pushed a `NULL`-exclusion `PreRuntimeConditionList` on the measure column.

That approach breaks XMLA/cube-bound sliders: `XMLAQuery.getDimensions()` treats every condition column as a dimension filter and calls `XMLAUtil.getDimension()` on it. For a measure ref this returns `null`, and `XMLAHandler.cacheLevels()` then dereferences that `null`, throwing:

```
java.lang.NullPointerException: Cannot invoke "inetsoft.uql.xmla.Dimension.getLevelCount()" because "dim" is null
	at inetsoft.uql.xmla.XMLAQuery.getLevelCount(XMLAQuery.java:273)
	at inetsoft.uql.xmla.XMLAHandler.cacheLevels(XMLAHandler.java:457)
	...
```

This PR replaces the query-level condition with a Java-side fix: `TimeSliderVSAQuery.refreshDimensionsValue()` now skips a row entirely when the single bound ref's value is `null`, instead of building a `SelectionValue` with a blank label. This achieves the same result (a null-valued member can't become the slider's min/max anchor) without pushing anything down to the query/MDX layer, so it works for both SQL and cube-backed sliders.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/TimeSliderVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/TimeSliderVSAQuery.java
@@ -1545,8 +1545,10 @@ public class TimeSliderVSAQuery extends AbstractSelectionVSAQuery {
 
             // a null value (e.g. a member with no aggregate/measure data) can't be a
             // meaningful min/max anchor for a single-value slider, so skip the row
-            // instead of showing a blank label
-            if(refs.length == 1 && obj == null) {
+            // instead of showing a blank label. only applies to SingleTimeInfo, which
+            // is the binding that used to exclude nulls in the query; a composite
+            // slider (which may also have a single ref) still shows null members.
+            if(tinfo instanceof SingleTimeInfo && obj == null) {
                continue rows;
             }
 
--- a/core/src/main/java/inetsoft/report/composition/execution/TimeSliderVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/TimeSliderVSAQuery.java
@@ -139,15 +139,6 @@ public class TimeSliderVSAQuery extends AbstractSelectionVSAQuery {
                sortinfo.addSort(sort);
                tassembly.setSortInfo(sortinfo);
             }
-
-            // ignore null otherwise a member with a null aggregate/measure value
-            // sorts to the front and shows as a blank min label
-            ConditionList conds = new ConditionList();
-            Condition cond = new Condition(column.getDataType());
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (changes requested):
Requesting changes based on Claude's automated review. It flagged a real behavior-change concern introduced by the new `if(refs.length == 1 && obj == null) continue rows;` guard in `refreshDimensionsValue()`:

`refs.length == 1` is not a reliable signal that we're in the single-measure/dimension slider case this fix targets — `CompositeTimeInfo` can also end up with exactly one `DataRef` (e.g. via `RangeSliderPropertyDialogService.setTimeInfo()` with a single column selected as "composite", or the single-dragged-column fallback in `VSBindingService`). Both of those route through `refreshCompositeSelectionValue()` → `refreshDimensionsValue()` too, so the new guard also silently drops rows there — previously a null value in that single composite column fell through to `CoreTool.FAKE_NULL` and still produced a (blank) selectable entry; now the row is dropped entirely. This is a behavior change beyond the intended scope of the XMLA/cube-measure NPE fix, and could hide legitimate null-valued members from string-dimension range sliders bound via the composite path with one column.

Suggested fix (per Claude's review): thread an explicit flag through from the two call sites (`refreshSingleSelectionValue` vs `refreshCompositeSelectionValue`) indicating "this is the single-ref anchor case," rather than inferring it from `refs.length`.

It also noted this is the second fix for the same underlying issue (#75823 → this NPE regression), and there's still no test coverage for `TimeSliderVSAQuery`/`refreshDimensionsValue` — a small test feeding a null measure/dimension value through both the single-ref and composite-with-one-column paths would help catch a third regression here.

The core NPE fix itself (removing the pre-runtime condition pushdown on cube/MEMBER queries) looks correct. Happy to re-review once the composite-path scoping is addressed.

jshobe-inetsoft (approved):
Approving based on Claude's automated re-review — the fix now correctly scopes the null-skip guard using 'tinfo instanceof SingleTimeInfo' instead of the unreliable refs.length == 1 check, resolving the composite-slider regression concern from the prior round. Regression tests cover all three scenarios. No blocking issues, checks pass, and there are no conflicts.

PR #4557, opened 2026-08-10
Bug #75930, fix round corner export follow-up for Table/Crosstab (Excel/PDF/SVG/PNG/PPT/HTML)

ngovus-inetsoft:
## Summary
Follow-up to #4516 after testing real exports of a round-corner table surfaced several remaining issues across every export format:

- **Excel**: the corner mask always ran, but the compensating rounded border outline required all four sides to have a border configured - and no border at all is the default/common case (`VSCompositeFormat.getBorders()` returns null when never set), so most tables got naked white squares with no curve. Now uses whichever side has a border configured first for the outline's style/color, and skips masking entirely when no side has one to compensate with.
- **PDF/SVG/PNG/PPT** (`PDFTableHelper`, `PDFCrosstabHelper`, `SVGTableHelper`, `SVGCrosstabHelper`, `PPTTableHelper`, `PPTCrosstabHelper`): the pass that draws the actual visible border stroke builds a fresh `VSCompositeFormat` and copies borders/colors, but never round corner - so the border always rendered square regardless of the setting, in all five of these formats. Fixed by copying it across in each.
- **PPT**: `PPTValueHelper` is a shared instance reused across per-cell writes, which leaves a stale non-zero `cellType` from the last cell written by the time the whole-object border draw runs, incorrectly gating off the round-corner branch. Reset it before that draw. Also relaxed the same "require all four sides" restriction as Excel, and added the same corner-masking (PPT has no Graphics2D-style clip to rely on).
- **PDF/SVG/PNG**: added a real `Graphics2D` clip (`RoundRectangle2D`) around the background fill, title, and per-cell data writing (new `beginRoundCornerClip`/`endRoundCornerClip` hooks in `VSTableDataHelper`, wrapped in try/finally so a write failure can't leave the shared graphics device permanently clipped), so square cell content is cleanly clipped to the rounded shape instead of leaving stray square border/background artifacts poking out past the curve at each corner - PDF/SVG can do this properly, unlike Excel/PPT.
- **HTML**: unrelated pre-existing bug - `getBorderString()`'s `border-radius` CSS was gated on `format.getRoundCornerValue()` (resolves only dynamic-value-configured round corners) instead of `format.getRoundCorner()` (the correctly resolved accessor used everywhere else), so it never emitted for the common case of a plain static round-corner number. Fixed the accessor and paired `border-radius` with `overflow:hidden` to match how the viewer clips rounded corners.

## Test plan
- [ ] Export a Table/Crosstab with round corner + a full border to Excel, PDF, SVG/PNG, PPT, and HTML; confirm all five show a clean rounded corner with no stray square artifacts.
- [ ] Repeat with round corner set but no border at all; confirm no naked white squares/mask artifacts in Excel/PPT.
- [ ] Repeat with only a partial border (e.g. top only); confirm the accepted uniform-outline tradeoff looks reasonable rather than broken.
- [ ] Confirm non-rounded tables are unaffected in all formats.

### DIFF
--- a/core/src/main/java/inetsoft/report/io/viewsheet/VSTableDataHelper.java
+++ b/core/src/main/java/inetsoft/report/io/viewsheet/VSTableDataHelper.java
@@ -345,16 +345,45 @@ public abstract class VSTableDataHelper extends ExporterHelper {
       int infoWidth = CoordinateHelper.getAssemblySize(
          assembly, CoordinateHelper.getLensSize(lens, true)).width;
       calculateColumnsPosition(info, lens);
-      drawObjectFormat(info, lens, false);
-      writeTitle(info, infoWidth);
+      // clip the background fill too, not just the title/data drawn after it - the
+      // background is a plain unclipped fillRect (see ExportUtil.drawTextBox) that
+      // otherwise sticks out square past the rounded border at each corner.
+      beginRoundCornerClip(info, lens);
 
-      if(lens != null) {
-         writeData(info, lens);
+      try {
+         drawObjectFormat(info, lens, false);
+         writeTitle(info, infoWidth);
+
+         if(lens != null) {
+            writeData(info, lens);
+         }
+      }
+      finally {
+         // always restore the clip, even on failure, so a bad cell/title write doesn't
+         // leave the shared Graphics device (PDF/SVG) clipped for whatever is drawn next.
+         endRoundCornerClip();
       }
 
       // @by stephenwebster, draw after the data so the object borders are on top
       // of the data borders.
       drawObjectFormat(info, lens, true);
+   }
+
+   /**
+    * Clip subsequent drawing (the object's own background fill, title, and cell data) to
+    * the object's rounded-corner bounds, so none of it sticks out past the rounded border
+    * drawn afterward by {@link #drawObjectFormat}. No-op by default: cell content is
+    * written directly to a native grid (Excel) or shape-based slide (PowerPoint) in some
+    * export formats, neither of which support clipping; only Graphics2D-based exports
+    * (PDF/SVG) override this.
+    */
+   protected void beginRoundCornerClip(TableDataVSAssemblyInfo info, VSTableLens lens) {
+   }
+
+   /**
+    * Restore the clip pushed by {@link #beginRoundCornerClip}, if any.
+    */
+   protected void endRoundCornerClip() {
    }
 
    /**
--- a/core/src/main/java/inetsoft/report/io/viewsheet/html/HTMLCoordinateHelper.java
+++ b/core/src/main/java/inetsoft/report/io/viewsheet/html/HTMLCoordinateHelper.java
@@ -355,10 +355,13 @@ public class HTMLCoordinateHelper extends CoordinateHelper {
          buffer.append("box-sizing:border-box;");
       }
 
-      if(format.getRoundCornerValue() != 0) {
+      if(format.getRoundCorner() != 0) {
          buffer.append(";border-radius:");
-         buffer.append(format.getRoundCornerValue());
+         buffer.append(format.getRoundCorner());
          buffer.append("px;");
+         // clip the square cell content to the rounded corner, matching the viewer
+         // (vs-table.component.html pairs border-radius with overflow:hidden the same way).
+         buffer.append("overflow:hidden;");
       }
 
       return buffer.toString();
--- a/core/src/main/java/inetsoft/report/io/viewsheet/pdf/PDFCoordinateHelper.java
+++ b/core/src/main/java/inetsoft/report/io/viewsheet/pdf/PDFCoordinateHelper.java
@@ -31,6 +31,7 @@ import org.slf4j.LoggerFactory;
 
 import java.awt.*;
 import java.awt.geom.Rectangle2D;
+import java.awt.geom.RoundRectangle2D;
 import java.awt.image.BufferedImage;
 import java.io.OutputStream;
 import java.util.HashMap;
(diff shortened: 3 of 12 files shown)


jshobe-inetsoft (changes requested):
## Summary

Reviewed the round-corner export follow-up fix across Excel, PDF, SVG/PNG, PPT, and HTML. The Excel border-fallback logic, the PPT `cellType`/corner-mask changes, and the HTML `getRoundCorner()` accessor fix are all correct and match the described intent. However, I found one bug in the new PDF/SVG `Graphics2D` clip helper that will cause a real rendering regression for SVG/PNG exports, and the required `claude-review` check is currently failing.

## Blocking issue

### SVG/PNG export: rounded-corner clip is never removed when there was no prior clip, corrupting everything drawn afterward on the page

`SVGCoordinateHelper.beginRoundCornerClip`/`endRoundCornerClip` (and the identical pattern in `PDFCoordinateHelper`) save/restore the clip like this:

```java
public void beginRoundCornerClip(Rectangle2D bounds, int roundCorner) {
   if(roundCorner <= 0) return;
   savedClip = svgGraphics.getClip();
   ...
   svgGraphics.setClip(new RoundRectangle2D.Double(...));
}

public void endRoundCornerClip() {
   if(savedClip != null) {
      svgGraphics.setClip(savedClip);
      savedClip = null;
   }
}
```

The bug: `Graphics2D.getClip()` returns `null` when no clip is currently established — which is the normal state for the shared `SVGCoordinateHelper`/`SVGGraphics2D` instance before any rounded-corner object has been drawn (`SVGVSExporter` creates one `SVGCoordinateHelper` per sheet and reuses it for every assembly; it only pre-clips to page bounds in the `isAllHidden` edge case). So for the common case:

1. `beginRoundCornerClip` sets `savedClip = null` (from `getClip()`), then clips the shared graphics to the table's rounded rectangle.
2. `endRoundCornerClip` checks `if(savedClip != null)` — false — and **skips restoring**, leaving the shared `Graphics2D` permanently clipped to that table's bounds.

Since `helper`/`svgGraphics` is shared across the entire sheet, every subsequent draw call — including this same table's own border overlay drawn right after via `drawObjectFormat(info, lens, true)`, and any other assembly positioned outside the table's rectangle — gets silently clipped away for the rest of the export. This will manifest as missing/cut-off content in SVG and PNG exports for any viewsheet containing a round-corner table/crosstab alongside other objects (the common case this PR is meant to support).

`PDFCoordinateHelper` happens to avoid this specific symptom only because `PDFPrinter.getClip()` never returns `null` (it always derives a `Rectangle2D` from an internally-initialized `clipping` field), but the same buggy pattern is duplicated there and is fragile/misleading as written.

**Suggested fix**: track whether a clip was pushed with an explicit flag instead of relying on non-null-ness, e.g.:

```java
private Shape savedClip;
private boolean clipped;

public void beginRoundCornerClip(Rectangle2D bounds, int roundCorner) {
   if(roundCorner <= 0) return;
   savedClip = svgGraphics.getClip();
   clipped = true;
   svgGraphics.setClip(new RoundRectangle2D.Double(...));
}

public void endRoundCornerClip() {
   if(clipped) {
      svgGraphics.setClip(savedClip); // restores null correctly if that was the prior state
      savedClip = null;
      clipped = false;
   }
}
```

Please apply the equivalent fix to both `PDFCoordinateHelper` and `SVGCoordinateHelper`.

## Other observations (non-blocking)

- **Excel (`PoiExcelVSExporter.drawTableRoundedBorder`)**: the "use whichever side has a border configured first" fallback and the reordering of the corner mask (now after the border-type/early-return checks) look correct; `type`/`color` are always initialized before use since the only path that skips assignment (`else`) returns early.
- **PPT (`PPTValueHelper.paintBorders`)**: relaxing the "all four sides required" condition and resetting `cellType` before the whole-object border draw in `PPTTableHelper`/`PPTCrosstabHelper` both look correct, and the new corner masking mirrors the Excel approach reasonably for PPT's shape-based rendering (no `Graphics2D` clip available).
- **HTML (`HTMLCoordinateHelper.getBorderString`)**: switching `getRoundCornerValue()` → `getRoundCorner()` is correct — `VSCompositeFormat.getRoundCorner()` resolves the fully-computed value (static or dynamic), matching every other round-corner call site in the export code, whereas `getRoundCornerValue()` only resolves the dynamic-value-defined case. Pairing with `overflow:hidden` matches `vs-table.component.html`'s viewer behavior.
- **Test coverage**: no automated tests were added for this rendering fix (consistent with the rest of this export code, which is largely untested at the unit level), and the PR's own manual test-plan checkboxes are unchecked. Given the bug above, at minimum the manual test plan should be executed — particularly "round corner + full border" and "round corner, no border" — across all five formats before merge, and ideally including a viewsheet with a rounded table *and another object outside its bounds* to catch the clip-leak regression above.

## CI / mergeability status

- The required `claude-review` check is currently **failing** (`Reached maximum number of turns (50)` after a ~6 minute run — an automation/execution failure rather than a specific reported finding, but it means the check has not completed successfully).
- No merge conflicts; `mergeStateStatus` is `BLOCKED` only because a review is required (branch protection), not due to conflicts.

## Verdict

Requesting changes because of the clip-restoration bug above (a genuine regression risk for SVG/PNG export) and the failing `claude-review` check. Once the clip fix is applied and checks are green, this should be good to merge.

jshobe-inetsoft (approved):
Approving based on Claude's automated re-review — the latest commit correctly fixes the clip-restoration bug (tracking a boolean flag instead of relying on savedClip's null-ness), resolving the blocking issue from the prior round. No new issues found, checks pass, and there are no conflicts.

PR #4560, opened 2026-08-11
Bug #75963, fix delete of wiz-managed folders silently no-oping

Zxm-happy:
## Summary
- Folders created under the wiz "Visualization Components"/"Wiz Chats" roots (`WizVisualizationService#ensureFolder`) are added directly via `assetRepository.addFolder()` and are never registered in the RepletRegistry, unlike folders created through the ordinary composer "New Folder" action (`AddFolderController`), which registers in both.
- `RemoveAssetController` routed every `REPOSITORY_FOLDER` delete through the RepletRegistry-based recycle-bin path (`RecycleUtils.moveRepositoryToRecycleBin` → `RepletRegistry#changeFolder`). For a folder that was never registered there, `changeFolder` finds nothing to rename but still returns `"true"`, so the delete silently no-oped: no error was shown, but the folder was never actually removed.
- Fix: route wiz-managed folders (paths under `VISUALIZATION_COMPONENTS_FOLDER_PATH` / `VISUALIZATION_ROOT_FOLDER_PATH`) through `assetRepository.removeFolder(entry, principal, true)` directly, mirroring the existing `WizVisualizationService#deleteVisualizations` logic. Ordinary repository folders keep using the existing recycle-bin path unchanged.

## Repro (via wiz)
1. In the wiz chat, open a chart's save dialog, click "New Folder", create a folder, and save into it.
2. In StyleBI Composer, browse to that folder under "Visualization Components" and delete it.
3. Before the fix: no error, folder stays. After the fix: folder and its contents are removed.

## Test plan
- [x] Added `RemoveAssetControllerWizFolderTest` covering: wiz "Visualization Components" folder deletion, wiz "Wiz Chats" folder deletion, and that ordinary (non-wiz) repository folders keep using the pre-existing recycle-bin path.
- [x] `mvn -pl community/core test -Dtest=RemoveAssetControllerWizFolderTest` — 3/3 passing.
- [x] `mvn -pl community/core test-compile` — compiles cleanly.

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/RemoveAssetController.java
+++ b/core/src/main/java/inetsoft/web/composer/RemoveAssetController.java
@@ -38,6 +38,7 @@ import inetsoft.web.RecycleBin;
 import inetsoft.web.RecycleUtils;
 import inetsoft.web.composer.model.RemoveAssetEvent;
 import inetsoft.web.viewsheet.command.MessageCommand;
+import inetsoft.web.wiz.service.WizVisualizationService;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.*;
@@ -119,6 +120,18 @@ public class RemoveAssetController {
             AssetEventUtil.removeStyleFolder(entry.getProperty("folder"), manager);
             manager.save();
             securityProvider.removePermission(ResourceType.TABLE_STYLE, entry.getProperty("folder"));
+         }
+         else if(isWizManagedFolder(entry)) {
+            // Wiz visualization/chat folders are created directly via
+            // assetRepository.addFolder() (see WizVisualizationService#ensureFolder), never
+            // through repletRepository.addFolder(), so they are never registered in the
+            // RepletRegistry's folder map. The RecycleUtils.moveRepositoryToRecycleBin() path
+            // below renames the entry inside the RepletRegistry -- for a folder that isn't
+            // registered there, RepletRegistry#changeFolder finds nothing to rename but still
+            // returns "true", so the delete silently no-ops while the folder stays in place.
+            // Remove it directly from the AssetRepository instead, mirroring
+            // WizVisualizationService#deleteVisualizations.
+            assetRepository.removeFolder(entry, principal, true);
          }
          else if(entry.isRepositoryFolder()) {
             String rpath = entry.getPath();
@@ -233,6 +246,26 @@ public class RemoveAssetController {
 
       messageCommand.setType(MessageCommand.Type.CONFIRM);
       return messageCommand;
+   }
+
+   /**
+    * @return {@code true} if {@code entry} is a repository folder under one of the wiz-managed
+    *         roots ({@code VISUALIZATION_ROOT_FOLDER_PATH}, {@code VISUALIZATION_COMPONENTS_FOLDER_PATH})
+    *         -- folders created via {@code assetRepository.addFolder()} and never registered in
+    *         the RepletRegistry, so they need the AssetRepository-direct removal path above
+    *         rather than the RepletRegistry-based recycle-bin path.
+    */
+   private boolean isWizManagedFolder(AssetEntry entry) {
+      if(!entry.isRepositoryFolder() || entry.getPath() == null) {
+         return false;
+      }
+
+      String path = entry.getPath();
+
+      return path.equals(WizVisualizationService.VISUALIZATION_ROOT_FOLDER_PATH) ||
+         path.startsWith(WizVisualizationService.VISUALIZATION_ROOT_FOLDER_PATH + "/") ||
+         path.equals(WizVisualizationService.VISUALIZATION_COMPONENTS_FOLDER_PATH) ||
+         path.startsWith(WizVisualizationService.VISUALIZATION_COMPONENTS_FOLDER_PATH + "/");
    }
 
    private void checkScriptRemoveable(RemoveAssetEvent event, AssetEntry entry, Principal principal) {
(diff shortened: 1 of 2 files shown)


Zxm-happy:
## Review

Verified the root-cause story against `RemoveAssetController`, `RepletRegistry#changeFolder`, and `WizVisualizationService`, and it checks out — routing wiz-managed folders through `assetRepository.removeFolder` instead of the RepletRegistry recycle-bin path is the right fix for the silent no-op.

One gap found by comparing this against the method it's explicitly modeled on:

**Missing companion worksheet-folder cleanup for `VISUALIZATION_COMPONENTS_FOLDER_PATH` folders.**

`WizVisualizationService#deleteVisualizations` — the method this PR's own comment says it mirrors — does two things when deleting a folder under `VISUALIZATION_COMPONENTS_FOLDER_PATH`, not one:

```java
// WizVisualizationService.java:914-932
if(entry.isRepositoryFolder()) {
   assetRepository.removeFolder(entry, principal, true);

   String suffix = entry.getPath().substring(VISUALIZATION_COMPONENTS_FOLDER_PATH.length());
   String wsFolderPath = GenerateWsService.WORKSHEET_COMPONENTS_FOLDER_PATH + suffix;
   AssetEntry wsFolderEntry = new AssetEntry(
      AssetRepository.GLOBAL_SCOPE, AssetEntry.Type.FOLDER, wsFolderPath, null);

   if(assetRepository.containsEntry(wsFolderEntry)) {
      try {
         assetRepository.removeFolder(wsFolderEntry, principal, true);
      }
      catch(Exception e) {
         LOG.warn("Failed to delete worksheet folder (visualization folder deleted): {}", wsFolderPath, e);
      }
   }
}
```

Every saved visualization under `VISUALIZATION_COMPONENTS_FOLDER_PATH` has a parallel backing worksheet under `GenerateWsService.WORKSHEET_COMPONENTS_FOLDER_PATH` (same path suffix). The new `isWizManagedFolder` branch in `RemoveAssetController` only calls `assetRepository.removeFolder(entry, principal, true)` on the visualization folder itself — it doesn't remove the companion worksheet folder. Deleting a visualization folder through the composer's generic "Remove" action (this PR's new path) will leave the backing worksheet folder and its worksheets orphaned in the repository, where deleting the *same* folder through the wiz-native `deleteVisualizations` path does not.

This doesn't affect `VISUALIZATION_ROOT_FOLDER_PATH` ("Wiz Chats") folders — `deleteVisualizations`'s companion-cleanup logic is specific to `VISUALIZATION_COMPONENTS_FOLDER_PATH`, so only the "Visualization Components" side of this fix is affected.

I'll push a commit to this branch that adds the same companion-folder cleanup (with the same containsEntry-guard + best-effort try/catch) plus a regression test, since I have write access to the branch.

Zxm-happy:
## Re-review

Pushed a fix for the gap found in my earlier comment, plus did a second full pass over the rest of the diff looking for anything else. Summary below.

### Fixed: missing companion worksheet-folder cleanup

Commit 493df0bec adds `removeCompanionWorksheetFolder()`, called right after `assetRepository.removeFolder(entry, principal, true)` in the `isWizManagedFolder` branch. It mirrors `WizVisualizationService#deleteVisualizations`'s exact pattern: compute the parallel path under `GenerateWsService.WORKSHEET_COMPONENTS_FOLDER_PATH`, `containsEntry`-guard, best-effort `try/catch` with a warn log on failure so a worksheet-cleanup failure never surfaces as a user-facing error (the visualization folder is already gone by that point). No-op for `VISUALIZATION_ROOT_FOLDER_PATH` ("Wiz Chats") folders, which have no such companion — matches `deleteVisualizations`, which only ever does this for the `VISUALIZATION_COMPONENTS_FOLDER_PATH` side.

Added 5 new test cases to `RemoveAssetControllerWizFolderTest` (now 8 total): companion folder removed when it exists, not attempted when it doesn't, a companion-removal failure is swallowed rather than propagated, "Wiz Chats" folder deletion never even calls `containsEntry`, plus the original 3 stay green unaffected (they rely on `containsEntry` defaulting to `false` on the mock, so the new code path is a no-op for them).

Verified: `mvn -o compile` + `mvn -o test -Dtest=RemoveAssetControllerWizFolderTest` — clean, 8/8 passing.

### Second pass — no other blocking issues found

Specifically checked and ruled out:
- **Permission/dependency cleanup parity** with the other `removeAsset` branches (script/table-style call `securityProvider.removePermission` / `dependencyHandler.deleteDependenciesKey`) — wiz-managed folders never go through the replet/permission registration path in the first place (that's the whole root cause this PR fixes), and `deleteVisualizations` doesn't do this cleanup either, so there's nothing parallel to clean up here. Consistent, not a gap.
- **False-positive path matching** in `isWizManagedFolder` — the `startsWith(ROOT + "/")` check (not just `startsWith(ROOT)`) means an unrelated folder name that happens to share the same string prefix without a `/` separator can't accidentally match. Both root constants are also UUID-suffixed, so an accidental collision is a non-concern anyway.
- **Force-delete semantics** (`removeFolder(..., true)` skips dependency checking, and this path bypasses the recycle bin entirely — no undo) — this is inherited from mirroring `deleteVisualizations`, which already behaves this way for all wiz-managed content, not a new risk introduced by this PR.
- **Nested-folder / recursive cascade correctness** — `removeFolder(..., true)` cascades to children per the existing code comment, and since the vis-folder and ws-folder trees mirror each other 1:1 by suffix, removing both roots at the SAME suffix depth correctly cascades regardless of nesting.

No other issues found. Once CI is green on the new commit this looks ready to merge (pending your own manual verification of the repro steps in the description, if you haven't already re-run them against this exact commit).

fredaan (approved):

PR #4567, opened 2026-08-11
Bug #76041, fix selection parameters not sent to hyperlink-target viewsheet

ngovus-inetsoft:
## Summary
- `VSLifecycleService` discarded the updated parameter map returned by the cluster-proxied `setRuntimeParameters` call, so selection values from `SelectionList`/`SelectionTree` never reached a hyperlink-target viewsheet's variable table.
- `VSLifecycleControllerService` only read index `[0]` of each selection's value array and mangled it through a bogus `^`-split, dropping every other selected value.
- `VariableScriptable.toString()` printed array-valued and principal-valued parameters using Java's default `Object[]` toString(), so any script referencing the whole `parameter` scope (e.g. `Text1.value = parameter`) rendered `[Ljava.lang.Object;@hash` instead of readable values — now formatted the same way `VariableTable.toString()` already does.

## Test plan
- [x] Repro from the bug report: open `VS-Hyperlink/hyper9`, select AZ/CA in `SelectionList1` and Animal World/Barbie's Fashion in `SelectionTree2`, click `ClickMe_vsLink` — target viewsheet now receives the full selection value arrays (verified via debug logging before removal).
- [x] `core` module compiles cleanly (`./mvnw -pl core -am compile -DskipTests`).

### DIFF
--- a/core/src/main/java/inetsoft/uql/script/VariableScriptable.java
+++ b/core/src/main/java/inetsoft/uql/script/VariableScriptable.java
@@ -17,6 +17,7 @@
  */
 package inetsoft.uql.script;
 
+import inetsoft.sree.security.SRPrincipal;
 import inetsoft.uql.VariableTable;
 import inetsoft.util.script.JavaScriptEngine;
 import inetsoft.util.script.graal.ScriptScope;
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):

PR #4568, opened 2026-08-11
Bug #76034, fix custom shapes not deleted when resetting global presentation settings

ngovus-inetsoft:
## Summary
- Resetting presentation settings at the host/global level deleted the org-scoped shapes directory instead of the global shapes directory the Custom Shapes upload actually writes to, so uploaded shapes survived a reset.
- `PresentationSettingsController.resetSettings()` now picks the same directory the upload path uses, based on the already-computed `globalSettings` flag.

## Test plan
- [ ] As host/global admin: EM → Presentation Settings → Look and Feel → Custom Shapes → upload a gif → close dialog → Reset to Default → OK → reopen Custom Shapes → confirm the gif is gone.
- [ ] As an org admin: repeat the same steps for an organization and confirm shapes still reset correctly (no regression).


jshobe-inetsoft (approved):

PR #4573, opened 2026-08-12
feat(wiz): report options that were accepted but did not take effect

Nelson-cao:
Several apply paths take a request, skip the part of it that cannot apply to this chart, and return 200 for the rest. That is the right behaviour — rejecting the whole request over one unusable option would discard the settings that were fine — but the caller had no way to learn which part was dropped, so an edit that did nothing was reported to the user as done.

This carries those as warnings on the response. **Execution is unchanged throughout**: what was skipped is still skipped, what fell back to a default still falls back, and no status code changes.

## What now reports

| Option | Behaviour |
|---|---|
| y-axis scale on a chart with no y-axis | Neither the empty `getYFields()` nor the missing `AxisDescriptor` sits in an else branch, so the check counts the axes actually written |
| unrecognized `legendPosition` | Previously the one case that did report, via `note` |
| unrecognized `markerShape` | `GraphGenerator` maps it to `FILLED_CIRCLE` through a switch default, so the value is still passed through and the warning names the shape actually drawn |
| filter leaf naming no column | `canApplyConditions` refuses only a list where every node was unusable, so a partial drop widened the result and read as a correct answer |
| dimension / measure absent from the worksheet columns | Dropped from the aggregation |
| unrecognized aggregate formula | Both the binding and the having-clause path. This one changes what the numbers mean, so the reason states that the measure was summed instead |
| chart type with no matching recommendation | Substituted by the first available one. The reason names the type actually rendered, since the caller narrates the type it asked for |
| highlight rule naming a field that matches no ref | Silently colours every series instead of the one the rule meant |

Warnings are user-facing sentences, not internal labels — they reach the chat window through the wiz side.

## note is a single-purpose channel again

`combineNotes` merged the copy-fallback warning with unrelated parameter warnings into one string the caller could not tell apart, and the caller consequently read a legend rejection as a declined copy. It is gone; `note` carries the copy fallback alone.

## Testing

`mvn test` on the four affected classes — 55 passed, 0 failures.

- `WizVsServiceFilterCopyTest` 13 (+4)
- `WizAutoBindingServiceSetChartFormatTest` 17 (+5, 1 rewritten)
- `WizAutoBindingServiceChartTypeCandidatesTest` 10 (+4)
- `WizVsServiceResolveRuleHighlightRefsTest` 15 (+3)

The rewritten case asserted the note-concatenation behaviour this change removes; it now asserts the two channels stay separate. Paths without existing test coverage were changed without adding new test files, per the repository convention.

The consuming side is inetsoft-technology/stylebi-wiz#1554.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/model/ApplyWarning.java
+++ b/core/src/main/java/inetsoft/web/wiz/model/ApplyWarning.java
@@ -1,0 +1,76 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package inetsoft.web.wiz.model;
+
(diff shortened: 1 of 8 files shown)


jshobe-inetsoft (approved):

PR #4574, opened 2026-08-12
Bug #76056, guard against zero time conditions in redistributeTasks

ngovus-inetsoft:
`ScheduleTaskService.redistributeTasks` computed `duration / count`, where `count` is the number of `TimeCondition` instances across the selected tasks, with no guard against zero. Redistributing a        
  selection containing no time conditions threw `ArithmeticException: / by zero`.

  Issue #28611 reported this same exception from this same method in 2018 and was closed by disabling the Enterprise Manager button for the reported selection. That check keys off a fixed list of internal   
  task *names* (`ScheduleTaskListComponent.internalTask0`), which is unrelated to whether a task actually carries a `TimeCondition` — so the fault itself was never fixed, only one known client path to it.   
  Any other caller reaching `EMScheduleController.redistributeTasks` with a selection carrying no time condition still divided by zero.

  The codebase already encoded this precondition: `TaskBalancer.getInterval` performs the equivalent computation and rejects `conditions <= 0` explicitly, and `TaskBalancer.balanceTasks` short-circuits on an
  empty condition list before ever calling it. The two call sites disagreed about the same precondition.

  ## Fix

  - **`ScheduleTaskService.redistributeTasks`** — guard `count == 0` immediately after it is computed, before the division, throwing a localized `MessageException`. `MessageException` is the established     
  convention for user-facing validation errors in this package: `AdminExceptionHandler.handleException` special-cases it, returning the message intact to the client and logging at the exception's own level  
  without a stack dump, rather than falling through to a generic 500 + `ERROR`-level stack trace.
  - **`srinter.properties`** — add `em.schedule.distribution.noTimeConditions`, alongside the existing `em.schedule.distribution.*` keys.

  Enforced for every caller, rather than relying on a client disabling a control.

  ## Tests

  New `ScheduleTaskServiceTest`:
  - selection with no `TimeCondition` throws `MessageException` (not `ArithmeticException`) and never calls `saveTask`
  - empty selection likewise
  - selection **with** a `TimeCondition` still redistributes and saves — regression guard on the existing path

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/schedule/ScheduleTaskService.java
+++ b/core/src/main/java/inetsoft/web/admin/schedule/ScheduleTaskService.java
@@ -950,6 +950,12 @@ public class ScheduleTaskService {
          .flatMap(ScheduleTask::getConditionStream)
          .filter(TimeCondition.class::isInstance)
          .count();
+
+      if(count == 0) {
+         throw new MessageException(Catalog.getCatalog(principal)
+            .getString("em.schedule.distribution.noTimeConditions"));
+      }
+
       long duration = Duration.between(startTime, endTime).get(ChronoUnit.SECONDS) / 60L;
       int concurrency;
       long interval = duration / count;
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):
Approving based on Claude's automated review: the fix is minimal, correct, consistent with existing conventions, well-tested, and no blocking issues were found. All checks pass.

PR #4575, opened 2026-08-12
Bug #76055, report scheduler health as unknown instead of a false UP

ngovus-inetsoft:
## Summary

`SchedulerHealthIndicator.health()` only queried the scheduler inside `if(client.isCloud() || client.isAutoStart())`. Every other path fell through to `return Health.up().build()`, so the endpoint asserted the scheduler was healthy without ever having looked:

- The guard itself is `false` on every non-cloud install with the default `schedule.auto.start=false` (the value `StorageInitializer` writes on first storage init), and on every clustered install regardless of that property (`ScheduleClient.isAutoStart()` ANDs in a not-clustered check). These are exactly the deployments where the scheduler runs as an independently failing process.
- Inside the guarded branch, an absent `HealthStatus` or a caught exception from `client.getHealthStatus()` also fell through to UP. The exception path is not hypothetical -- Bug #72788 and Bug #74509 are production reports of exactly that log line (`Failed to get scheduler health`) masking a real failure.

## Fix

Restructured `health()` so every "couldn't determine" path returns `Health.unknown()` instead of `Health.up()`. Only a check that actually completed and got a real `SchedulerStatus` now returns `Health.up()`/`Health.down()`, matching existing behavior for those two cases exactly (same detail fields, same logging, same `statusDumpService.dumpStatus()` call on DOWN).

The exception path logs the full exception server-side as before, but intentionally does not include the exception message in the health response detail, since `/health` is exposed with `show-details: always` and may not require authentication.

## Testing

Added `SchedulerHealthIndicatorTest` covering all five branches (guard false, absent status, exception, healthy, unhealthy):

```
Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/server/src/main/java/inetsoft/web/health/SchedulerHealthIndicator.java
+++ b/server/src/main/java/inetsoft/web/health/SchedulerHealthIndicator.java
@@ -39,34 +39,49 @@ public class SchedulerHealthIndicator implements HealthIndicator {
 
    @Override
    public Health health() {
-      if(client.isCloud() || client.isAutoStart()) {
-         try {
-            Optional<HealthStatus> status = client.getHealthStatus();
-
-            if(status.isPresent()) {
-               SchedulerStatus health = status.get().getSchedulerStatus();
-
-               if(!health.isHealthy()) {
-                  LoggerFactory.getLogger(getClass()).error(
-                     "SchedulerHealthIndicator DOWN: status={}", health);
-                  statusDumpService.dumpStatus();
-                  return Health.down()
-                     .withDetail("started", health.isStarted())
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (changes requested):
Claude's automated review found no correctness issues with this fix, but the **build** check on the latest commit was cancelled rather than completing successfully. Please re-run/confirm a passing build before this can be merged.

jshobe-inetsoft (approved):
The build check that prompted my earlier changes-requested review has since passed on a re-run. Claude's automated review found no blocking issues (only a non-blocking suggestion to sanity-check health-group aggregation behavior). Approving.

PR #4576, opened 2026-08-12
Bug #76043, fix blank pop-up when a group container is used as a pop component

ngovus-inetsoft:
## Summary
Clicking a trigger whose **Pop Component** is set to a Group Container popped up a blank white box instead of showing the group's contents.

## Root cause
`isActivePopComponent()` in `vs-object-container.component.ts` only checked whether the object *itself* was the active pop component, never whether its `container` was. When a group container is popped out, its own div gets the `+99999` z-index boost needed to render above the dim overlay canvas (z-index 9996), but its grouped children (e.g. a `Text` and a `Gauge`) did not get that boost — leaving them stacked *underneath* the container's own opaque background image/border, which is what rendered as a blank white pop-up.

## Fix
`isActivePopComponent` now also matches when `vsObject.container` is the active, registered pop component — mirroring the existing `DataTipService.isCurrentDataTip(name, container)` two-way check already used for the equivalent data-tip case in `needsZIndexBoost` right below it.

## Testing
Added "Group 11 — isActivePopComponent" to `vs-object-container.component.display.tl.spec.ts` covering: no active pop, self-match, grouped-child match (the bug scenario), grouped-child non-match, and an unrelated top-level object while a different pop is active.

```
npx ng run portal:test-tl --include="**/vs-object-container.component.display.tl.spec.ts"
✓ 26 tests passed
```

### DIFF
--- a/web/projects/portal/src/app/vsobjects/objects/vs-object-container.component.display.tl.spec.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/vs-object-container.component.display.tl.spec.ts
@@ -30,6 +30,7 @@
  *   Group 8   isChartAnnotationSelected — selectedAnnotations membership
  *   Group 9   needsZIndexBoost — data tip and pop component cases
  *   Group 10  toolbarForceHidden — delegates to miniToolbarService
+ *   Group 11  isActivePopComponent — active pop component, self and grouped-child cases
  */
 
 import {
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (changes requested):
Claude's automated review raised an unresolved concern about whether this fix is complete: `isActivePopComponent()`'s new check only looks one level up (`vsObject.container`). If a `GroupContainer`/`Tab` can be nested inside another popped container, a grandchild's immediate `.container` won't match the top-level popped component, and the same blank-popup bug could resurface at that nesting depth — `zIndex()` walks the full ancestor chain for a similar reason.

Please confirm whether multi-level nesting is possible for pop components in this app, and if so, apply the same ancestor-chain walk here rather than a single-level check.

jshobe-inetsoft (approved):
Approving based on Claude's automated review — the follow-up commit fully addresses the multi-level nesting concern I raised previously, now walking the full ancestor chain consistently with zIndex(). Tests cover the new scenarios, checks pass, and there are no conflicts.

PR #4577, opened 2026-08-12
Bug #76054, fix org-scoped property name case handling in PropertiesEngine

ngovus-inetsoft:
## Summary
`PropertiesEngine.computePropertyNameCase` had an `inetsoft.org.` branch that no property name could ever reach, and the write/read case asymmetry it was presumably meant to prevent was real (latent) on the org-scoped property write path.

## The dead branch
All four case-preserving guards (`log.level.`, `plugin.extra.classpath.`, `^log\.[A-Z_]+\.level\..+$`, `^inetsoft\.uql\.jdbc\.pool\..+\.connectionTestQuery$`) are anchored at the start of the name and can never co-occur with an `inetsoft.org.` prefix. Any name that reached the `inetsoft.org.` check had already been lowercased and returned by the branch above it.

## The asymmetry it left in place
- **Write**: `setProperty(name, val, true)` builds the org-prefixed name and passes it back through the 2-arg `setProperty`, which re-runs the whole prefixed string through `computePropertyNameCase` — lowercasing it entirely, including a mixed-case suffix like a logger name.
- **Read**: `useAvailableOrgProperty` prefixes an already case-fixed name and looks it up with the suffix's case intact.

A mixed-case org-scoped property (e.g. a logger name) would therefore be stored as `...log.level.foo` and looked up as `...log.level.Foo` — never found, with no error anywhere.

This was latent, not active: no current caller writes a log level org-scoped (`setLogLevel` only uses the global 2-arg `setProperty`; org-scoped log levels are handled separately via `LogSettingService.fixLogName`), so nothing is broken today. But the trap was set for the first future caller of the org-scoped overload with a mixed-case name.

## Fix
Check the `inetsoft.org.` prefix first, lowercase only that prefix (matching `useAvailableOrgProperty`'s own `orgID.toLowerCase()`), and recurse the existing case-preserving rules onto the remaining property name. Non-org-prefixed names are unaffected.

### DIFF
--- a/core/src/main/java/inetsoft/sree/PropertiesEngine.java
+++ b/core/src/main/java/inetsoft/sree/PropertiesEngine.java
@@ -349,20 +349,27 @@ public class PropertiesEngine {
    }
 
    private String computePropertyNameCase(String name) {
+      // organization-scoped property names carry the org ID as a case-insensitive segment;
+      // strip and lowercase it, then apply the case rules below to the remaining property name
+      // so it matches what useAvailableOrgProperty() looks up.
+      if(name.startsWith("inetsoft.org.")) {
+         int dot = name.indexOf('.', "inetsoft.org.".length());
+
+         if(dot >= 0) {
+            String orgPrefix = name.substring(0, dot + 1);
+            String suffix = name.substring(dot + 1);
+            return orgPrefix.toLowerCase() + computePropertyNameCase(suffix);
+         }
+
+         return name.toLowerCase();
+      }
+
       if(!name.startsWith("log.level.") &&
          !name.startsWith("plugin.extra.classpath.") &&
          !name.matches("^log\\.[A-Z_]+\\.level\\..+$") &&
          !name.matches("^inetsoft\\.uql\\.jdbc\\.pool\\..+\\.connectionTestQuery$"))
       {
          return name.toLowerCase();
-      }
-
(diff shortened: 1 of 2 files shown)


ngovus-inetsoft:
Addressed both review points in 0577aeb23:

1. **Test coverage** — added `orgScoped_setProperty_mixedCaseLogLevelName_roundTripsWithCasePreserved` to `OrgLifecycleScopedPropertiesIntegrationTest`, writing/reading `log.level.com.MyCompany.Service` org-scoped. Verified locally: fails on the pre-fix code (write path lowercased the whole key), passes with the fix.
2. **Cache-bypass nit** — added a comment explaining the recursive `computePropertyNameCase` call deliberately skips `propertyNameCaseCache` to avoid a recursive-update `IllegalStateException` from re-entering `computeIfAbsent` on the same map.

Also acknowledging the risk-characterization point: this is reachable today via `UserTreeService.editOrganization()` (EM's Edit Organization properties panel), not just through a hypothetical future org-scoped log-level caller — updating my mental model of priority here, thanks for catching that.

Ran the full `OrgLifecycleScopedPropertiesIntegrationTest` suite locally: 4/4 passing.

jshobe-inetsoft (approved):
Approving based on Claude's automated review of the latest commit (0577aeb2): both prior review comments (cache-bypass rationale, missing regression test) are resolved, no blocking issues remain, and all checks pass.

PR #4578, opened 2026-08-12
Bug #76001, add type parameter to add_named_group for standalone groupings

jkidacki-inetsoft:
## Summary

`add_named_group` (the wiz worksheet-chat MCP tool's backend) always required an existing `table`/`column` to attach the grouping to, and had no way to set the grouping's output data type. When a user asked for a standalone grouping (e.g. "add a grouping with type string: NY/NJ/CT → Northeast, CA → West"), the agent had no way to satisfy it and instead had to fabricate a table just to get a column of the right type to attach to.

## Root Cause

StyleBI's own Composer "Grouping Properties" dialog (`GroupingAssemblyDialogService`) already supports a second mode: a standalone grouping not attached to any column, driven by a `type` field, which sets `DATA_TYPE_ATTACHED` instead of `COLUMN_ATTACHED` on the `DefaultNamedGroupAssembly` (matched later by data-type compatibility in `GroupRef.update()`). The wiz agent's `addNamedGroup` never exposed this second mode — `table`/`column` were mandatory and `type` was never wired through or applied via `setAttachedDataType`.

## Fix

- `EditRequest` doc comment updated (the `type` field already existed on the record, reused from other ops).
- `WorksheetAgentController` passes `req.type()` into `editor.addNamedGroup(...)`.
- `WorksheetEditService.addNamedGroup` gains a `type` parameter:
  - `table`+`column` both provided → unchanged existing `COLUMN_ATTACHED` behavior.
  - Only one of `table`/`column` provided → throws `PairingException`.
  - Neither provided → skips table/column resolution, defaults `type` to `XSchema.STRING` if omitted, and creates a `DATA_TYPE_ATTACHED` standalone grouping instead (mirroring the existing `editNamedGroup` fallback and the Composer dialog's own logic).
- Added unit tests covering the standalone path, the string-type default, and the mismatched table/column guard.

Companion plugin-side change: https://github.com/inetsoft-technology/stylebi-wiz/pull/new/bug-76001 (adds the `type` param + optional `table`/`column` to the MCP tool schema).

## Test Plan

- `WorksheetEditServiceTest` / `WorksheetAgentControllerTest` pass (7 relevant + full suite unaffected).
- `core` module compiles and test-compiles cleanly.
- Manually verified end-to-end via a live worksheet-chat session: asking to add a standalone string-typed named grouping (no existing table) now succeeds directly instead of requiring a table to be created first.

Fixes Redmine #76001.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
@@ -68,7 +68,9 @@ import java.util.Map;
  *   <li>{@code reorder_columns} — {@code table}, {@code columnOrder}</li>
  *   <li>{@code add_concat_subtable} — {@code table} (concat assembly), {@code name} (subtable to add)</li>
  *   <li>{@code remove_concat_subtable} — {@code table} (concat assembly), {@code name} (subtable to remove)</li>
- *   <li>{@code add_named_group} — {@code name}, {@code table}, {@code column}, {@code groupMappings}, {@code groupOthers}</li>
+ *   <li>{@code add_named_group} — {@code name}, {@code groupMappings}, {@code groupOthers}; either
+ *       {@code table} + {@code column} (attach to a column) or {@code type} (standalone grouping,
+ *       matched by data type; defaults to {@code "string"})</li>
  *   <li>{@code set_column_description} — {@code table}, {@code column}, {@code description}</li>
  *   <li>{@code set_variable_values} — {@code variableValues} (map of variable name → value)</li>
  *   <li>{@code set_mirror_auto_update} — {@code table}, {@code visible} (true=auto-update on, false=off)</li>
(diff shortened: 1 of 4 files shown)


jkidacki-inetsoft:
Follow-up fix pushed: the initial change passed a null \`DataRef\` into \`ConditionItem\` when a named group's standalone (\`DATA_TYPE_ATTACHED\`) grouping wasn't attached to a column. \`ConditionItem\` requires a non-null attribute — a null one NPEs in \`ConditionItem.toString()\`, which fires on every worksheet clone (e.g. \`TouchAssetService\`):

\`\`\`
NullPointerException: Cannot invoke "DataRef.toString()" because "this.attr" is null
  at inetsoft.uql.ConditionItem.toString
  at inetsoft.uql.asset.NamedGroupInfo.clone
  at inetsoft.uql.asset.DefaultNamedGroupAssembly.clone
\`\`\`

Root cause: the Composer's own Grouping Properties dialog (\`grouping-condition-dialog.component.ts\`) never leaves this ref null — when there's no attached column it substitutes a \`"this"\` placeholder \`BaseField\`. Added a \`namedGroupConditionRef()\` helper that mirrors this server-side, used in both \`addNamedGroup\` and the pre-existing \`editNamedGroup\` (which had the identical latent bug, previously unreachable). Added regression tests that assert directly on the condition's \`DataRef\` and exercise \`Worksheet.clone()\`.

jshobe-inetsoft (approved):
Approving based on Claude's automated review of the latest commit (c2f8c232): the type validation gap from earlier rounds is resolved, no blocking issues remain, and all checks pass.

PR #4579, opened 2026-08-12
Fix outgoing joins pane not updating on graph node selection

swaker854:
## Summary
- Clicking a single node in the physical model graph (including auto-alias
  and auto-alias-by-outgoing nodes) didn't update `editingTable`, so the
  outgoing joins pane stayed stale or empty instead of reflecting the
  clicked table.
- `graphNodesSelected()` now resolves the selected node's tree path back to
  its source table (new `changeEditingTableByGraphPath()` /
  `getGraphTableName()` / `getGraphAliasName()` helpers), clearing
  `editingTable` when zero or multiple nodes are selected.
- Fixed a missing `NgIf` import in `PhysicalTableJoinsComponent`'s
  standalone `imports` array.

## Test plan
- [ ] `npm run test:portal:tl` — new Group 5 tests in
      `database-physical-model.component.tl.spec.ts` (single-node select,
      auto-alias resolution)
- [ ] Manual: open a physical model, click a single table node in the graph,
      confirm the outgoing joins pane updates to that table
- [ ] Manual: click an auto-alias node, confirm it resolves to the alias's
      source table
- [ ] Manual: click multiple nodes / deselect, confirm the joins pane clears

### DIFF
--- a/core/src/main/resources/inetsoft/web/resources/js/login.js
+++ b/core/src/main/resources/inetsoft/web/resources/js/login.js
@@ -27,6 +27,7 @@ function initLoginView(requestedUrl, sessionExpired, defaultErrorMessage, gatewa
    var $loginAsNameField = $("#loginAsName");
 
    var $loginButton = $("#loginButton");
+   var $activeSession = $("#activeSession");
    var $confirmLogin = $("#confirmLogin");
    var $terminateSessionButton = $("#terminateSessionButton");
    var $sessionLimitTableBody = $("#sessionLimitTableBody");
@@ -94,7 +95,7 @@ function initLoginView(requestedUrl, sessionExpired, defaultErrorMessage, gatewa
       }
 
       if(!!currentUser && !!userName && currentUser != userName && !logoutCurrent) {
-         $("#activeSession").modal("show");
+         $activeSession.modal("show");
          return;
       }
 
@@ -188,6 +189,10 @@ function initLoginView(requestedUrl, sessionExpired, defaultErrorMessage, gatewa
       authenticate(false);
    });
 
+   $activeSession.on("shown.bs.modal", function() {
+      $confirmLogin.trigger("focus");
+   });
+
    $togglePassword.click(function() {
       var show = $passwordField.attr("type") === "password";
       $passwordField.attr("type", show ? "text" : "password");
@@ -208,7 +213,7 @@ function initLoginView(requestedUrl, sessionExpired, defaultErrorMessage, gatewa
    });
 
    $confirmLogin.click(function() {
-      $("#activeSession").modal("hide");
+      $activeSession.modal("hide");
       authenticate(true);
    });
 
--- a/web/projects/portal/src/app/portal/data/data-datasource-browser/datasources-database/database-physical-model/database-physical-model.component.tl.spec.ts
+++ b/web/projects/portal/src/app/portal/data/data-datasource-browser/datasources-database/database-physical-model/database-physical-model.component.tl.spec.ts
@@ -63,6 +63,7 @@ import { DataPhysicalModelService } from "../../../services/data-physical-model.
 import { PhysicalModelDefinition } from "../../../model/datasources/database/physical-model/physical-model-definition";
 import { PhysicalTableModel } from "../../../model/datasources/database/physical-model/physical-table-model";
 import { PhysicalTableType } from "../../../model/datasources/database/physical-model/physical-table-type.enum";
+import { GraphModel } from "../../../model/datasources/database/physical-model/graph/graph-model";
 import { FixedDropdownService } from "../../../../../widget/fixed-dropdown/fixed-dropdown.service";
 import { DatabasePhysicalModelComponent } from "./database-physical-model.component";
 
@@ -114,6 +115,37 @@ function createTable(overrides: Partial<PhysicalTableModel> = {}): PhysicalTable
       baseTable: false,
       ...overrides,
    };
+}
+
+function createGraph(
+   overrides: Partial<Omit<GraphModel, "node">> & { node?: Partial<GraphModel["node"]> } = {}
+): GraphModel {
+   const id = overrides.node?.id ?? overrides.node?.name ?? "Orders";
+
+   return {
+      edge: { input: [], output: [] },
+      cols: [],
+      bounds: null,
+      showColumns: false,
+      alias: false,
+      autoAlias: false,
+      sql: false,
+      baseTable: false,
+      autoAliasByOutgoing: false,
+      designModeAlias: false,
+      ...overrides,
+      node: {
+         id,
+         name: id,
+         tableName: id,
+         label: id,
+         tooltip: id,
+         treeLink: `SalesDB/${id}`,
+         aliasSource: null,
+         outgoingAliasSource: null,
+         ...overrides.node,
+      },
+   } as GraphModel;
 }
 
 function createPhysicalModel(overrides: Partial<PhysicalModelDefinition> = {}): PhysicalModelDefinition {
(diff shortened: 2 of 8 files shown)


FigurePaper (changes requested):
## Code Review — Fix outgoing joins pane not updating on graph node selection

The problem is real and worth fixing: clicking a graph node whose tree node is not loaded (or has no tree node at all) left `editingTable` stale, so the joins pane showed another table's joins. The `NgIf` fix is also a genuine bug fix. However, the table-resolution logic added in `changeEditingTableByGraphPath()` can resolve to the wrong table in exactly the case it was written for (aliases), and the test that claims to cover that case does not actually exercise it. Requesting changes on those points.

### Blocking

**1. Path-first matching is ambiguous whenever the physical view contains aliases (`database-physical-model.component.ts:1470-1474`)**

The first predicate in the `find()` matches on path:

```ts
this.getTablePath(table) === path || table.path === path
```

An alias table does not have its own path. `PhysicalModelService.getTableXNode()` resolves the requested name through `partition.getAliasTable(tableName, true)` before looking up metadata, so `PhysicalModelService` (line ~265, `path = getTablePath(dataSource, tableNode)`) assigns an alias table the path of its **source** table. Correspondingly, `JoinGraphModel.buildGraphNodeModel()` sets an auto-alias node's `treeLink` to `physicalTable.getPath()` — again the source table's path.

Consequences:

- A source table and every auto-alias of it share one `path`, and their graph nodes all emit the **same** `treeLink`. Path can never discriminate between them.
- `Array.find()` returns the first entry, so which table lands in the joins pane depends on the order `partition.getTables()` happened to enumerate — not on which node the user clicked.

This is precisely why the existing `findMatchingTable()` (line 334) refuses to match alias tables by path and requires `qualifiedName` plus `databaseName + "/" + alias` instead. The new resolver should follow the same rule: discriminate on identity first, and fall back to path only when the node is not an alias node.

**2. The predicate contradicts itself for alias nodes (`database-physical-model.component.ts:1470-1494`)**

Within one `find()` the code asks for two different tables:

- `getGraphTableName()` maps an auto-alias node to `graph.node.aliasSource` — the **source** table.
- `table.qualifiedName === graph.node.id` and `table.qualifiedName === aliasName` both match the **alias** table.

Since these are OR'd, the result is again decided by array order rather than intent. Please decide which table the joins pane should edit for an auto-alias node and encode only that. For what it's worth, "the source table" (what the PR description states) is the choice consistent with the tree-driven path: the auto-alias node's `treeLink` points at the source table's tree node, so `selectAndExpandToPath()` → `selectNode()` already resolves to the source. If that is the intent, the `node.id` and `aliasName` clauses are wrong and should be removed rather than kept as fallbacks.

**3. `getGraphAliasName()` is dead code (`database-physical-model.component.ts:1493-1495`)**

`JoinGraphModel.buildGraphNodeModel()` sets both `graphNode.setId(table.getQualifiedName())` and `graphNode.setName(table.getQualifiedName())`, so `node.id === node.name` always. `getGraphAliasName()` returns `graph.node.name` for alias nodes, which makes

```ts
(!!aliasName && table.qualifiedName === aliasName)
```

an exact duplicate of the preceding `graph.node.id` clause. The helper and one of the two clauses can be deleted outright.

**4. The auto-alias test does not test auto-alias resolution (`database-physical-model.component.tl.spec.ts:394-424`)**

In `should resolve an auto-alias graph node to its source table`, the node's `treeLink` is `"SalesDB/Orders"`, identical to `orders.path`, and `createPhysicalModel()` sets no `connection`, so `getTablePath()` returns `table.path` unchanged. The **first** predicate matches and `find()` returns before any alias code runs — flipping `autoAlias`, `aliasSource` or removing `getGraphTableName()` entirely would not change the result. The model also holds a single table, so the test cannot detect a wrong-table selection.

To make it meaningful: give the alias node a `treeLink` that matches no table's `path`, and put both the alias table and its source in `tables` (alias first) so the assertion pins down which one wins. A second case with two auto-aliases of the same source — same `treeLink`, different `node.id` — would cover issue 1 directly.

**5. Two resolvers now race for a single click (`database-physical-model.component.ts:1439-1457`)**

`graphNodesSelected()` resolves and assigns `editingTable`, then calls `selectAndExpandToPath(nodes)`, which calls `tableTree.selectNode(child)` → `nodeSelected` → `selectNode()` → `changeEditingTable(node)` → resolution by `node.data.qualifiedName`. So:

- if the tree node is already loaded, the newly computed value is overwritten immediately;
- if it sits behind an unexpanded folder, the overwrite lands later, asynchronously, from the `openFolder()` subscription;
- if there is no tree node, the new value survives.

The same click can therefore produce different panes depending only on tree expansion state, and the two code paths use different matching rules. Please route both through one alias-aware lookup, or suppress the tree-driven re-resolution when the selection originated in the graph.

### Non-blocking

**6. Multi-select clearing is not reliable.** `PhysicalModelTableTreeComponent.selectNode()` pushes onto `selectedNodes` and emits on every call, so a two-node graph selection emits `[a]` then `[a, b]`; the first emission sets `editingTable` before the second clears it. If only one of the selected nodes exists in the loaded tree, nothing clears it at all and `editingTable` stays set, contrary to the new `changeEditingTableByName(null)`.

**7. Three writers for one field.** `changeEditingTableByGraphPath()` assigns `this.editingTable` directly instead of going through `changeEditingTableByName()`, and unlike `selectNode()` it leaves `selectedGraphModels` untouched. Consider a single private setter so the invariants stay in one place.

**8. Readability.** A five-clause OR inside `find()` is hard to reason about (as issues 1-3 show). Extracting a named predicate — `matchesGraphNode(table, graph, path)` with early returns per node kind: alias, auto-alias, plain — would make the intended precedence explicit.

**9. `NgIf` fix — correct, with a note.** The template uses `*ngIf` twice, and the component is standalone, so without the import neither the empty-state nor the `<tree>` was ever instantiated: the joins pane body rendered blank. Because `strictTemplates` is `false` in `web/tsconfig.json`, this failed silently instead of raising NG8103/NG8002, which is why the base branch still built. Optional: the sibling templates in this feature (for example `database-physical-model.component.html`) have moved to `@if`; converting these two would match that direction and drop the import again.

### Security, compatibility, regressions

- No security impact: no new I/O, no user-controlled value reaching a URL, template or DOM sink.
- No backward-compatibility concerns: no public API, model or wire-format change; the new methods are private.
- Regression surface is limited to graph-node selection, but issues 1, 2 and 5 are themselves regression risks for physical views that use aliases — previously the joins pane kept the tree-resolved table, whereas it can now be set to a different table before the tree overwrites it.

### Verification performed

- Ran the new tests locally: all 10 tests in `database-physical-model.component.tl.spec.ts` pass, including the two new Group 5 cases. They pass for the reason described in issue 4, not because the alias logic works.
- `ng run portal:test-tl` cannot build on this branch — 19 pre-existing TypeScript errors in unrelated specs (`app.component.risk.tl.spec.ts`, `chart-plot-area.component.test-helpers.ts`, several composer dialogs). The identical error set is produced at the merge base `c75c3fa`, so this is inherited from `epic-74519` and is not caused by this PR. It does mean the stated test plan (`npm run test:portal:tl`) cannot be run on the branch as-is.

### About the failing `claude-review` check

This failure is not attributable to the changes in this PR, and it should not be treated as a code signal:

- The job ended with `"subtype": "error_max_turns"` and `Execution failed: Reached maximum number of turns (50)` after 51 turns, with `permission_denials_count: 9`. The review agent ran out of its turn budget; it never reported a finding.
- The workflow prompt instructs the agent to obtain the diff via `Bash(git diff origin/epic-74519...HEAD)`, but the runtime allowlist is `mcp__github_inline_comment__create_inline_comment, Bash(gh pr comment:*), Bash(gh pr diff:*), Bash(gh pr view:*)` plus `Glob/Grep/LS/Read`. The denied calls burned turns against the `--max-turns 50` cap.
- The same failure occurred on an unrelated branch the same day — run `31625297618` on `bug-76001`, identical `error_max_turns` message — and a re-run of that workflow succeeded. A re-run here would likely go green too.

The other checks (`build`, `Semgrep OSS`, `Semgrep Scan`, `license/cla`) all pass, and the branch reports `MERGEABLE` with no conflicts.

### Verdict

Requesting changes for the correctness issues above (1-5), not for the failing check. Points 1, 2 and 4 are the substantive ones: the resolution order makes the alias case order-dependent, and the test that was meant to guard it passes without touching the alias code path.

swaker854:
Updated the graph-selection fix to address the alias-resolution review feedback.

The main fix now resolves graph selections from the selected `GraphModel` rather than path-only data, which lets the parent distinguish plain tables, aliases, auto-aliases, and outgoing auto-aliases deterministically. Auto-alias nodes resolve to their source table, path matching is only used as a non-alias fallback, multi-select clears the joins pane, and graph-originated tree syncing no longer re-emits selection and races with the graph resolver.

I also converted the joins pane template from `*ngIf` to `@if`, so the standalone component renders the empty state/tree without requiring `NgIf`.

Follow-up coverage was added for `autoAliasByOutgoing` to exercise the separate `outgoingAliasSource` branch with alias-first table ordering and a non-matching `treeLink`, so it cannot pass accidentally through path matching.

swaker854:
build has been successful. the last build is caused by a unrelated testcase error.

jshobe-inetsoft (changes requested):
Claude's automated review of the latest commit (dd904d94) found no remaining blocking issues, but the **build** check is currently failing on this PR. Requesting changes to get the build green before this can be merged.

jshobe-inetsoft (approved):

FigurePaper (approved):

PR #4580, opened 2026-08-12
Register StyleBI data sources from an embedded custom element

larryliang-inetsoft:
**Branch:** `feature-76074` · Redmine #76074

Registers a StyleBI data source from outside the portal shell, as a custom element the wiz console
(and any other host) can embed.

The element owns the **lifecycle only** — pick a type, seed from the server, hand the model to
StyleBI's own `DatasourcesDatasourceEditorComponent`, save. Every form control, editor type,
dependent-field refresh and OAuth flow comes from the TabularView descriptor the server sends. That
is the whole reason for wrapping rather than rebuilding: a new datasource plugin ships a new
descriptor and this element renders it with no change here.

```
<inetsoft-datasource-registration listing-name="Apache Cassandra">
  → registered {name, type} | cancelled | failed <message>
```

## Gate evidence, measured against a running deployment

| | |
|---|---|
| **Attribution** | With security enabled, `get-current-user` reports `admin` / `isSysAdmin: true` — not `anonymous`. Both the seed and write paths return 200 under that identity. |
| **Same origin** | The element resolves `../api/...` exactly as the portal does from its `/sree/` base href. |
| **CSRF** | Measured both ways: a DELETE without `X-XSRF-TOKEN` → **403**, with it → **200**. The element's own `POST refreshView` and `POST datasources` both carry the token and return 200. |
| **Descriptors complete** | All **130** listings the server offers seed successfully — 130 × HTTP 200, swept. The Cassandra form renders its real fields: Host, Port, Datacenter, Use SSL, Keyspace, User, Password. |
| **End to end** | Filled the form in a browser, saved: `POST /api/portal/data/datasources` → 200, `registered` fired, the source appeared in StyleBI's Data page. Then deleted. |
| **Bundle cost** | 5,287,732 vs 5,169,698 bytes = **118,034 (115 KB, 2.3%)**, built with and without the registration. |

## Four defects the unit tests could not see

Running it in a real browser found all of these; every one passed its tests first.

**1. Both list calls read keys that do not exist.** `listings` was read as `dataSourceListings`,
`dataSourceList` as `files`. Both were written `?? []`, so both type-checked *and passed their own
tests* — the tests asserted the same invented key. Against a real server the picker would have shown
zero of 130 types, indistinguishable from a deployment with no data source plugins installed, and
duplicate names would have gone unchallenged client-side. A missing key now throws with the URI and
key named; an empty **array** still passes through, because that is a legitimate answer.

**2. Custom elements get no change-detection pass.** The picker received all 130 listings and drew
none of them — one keystroke in the filter box made 69 appear at once. State assigned in an HTTP
callback marks nothing and schedules no tick. Fixed with an explicit `detectChanges()` after each
async change, guarded on a `destroyed` flag because a response can outlive the element that asked
for it. Same remedy the embedded chart needed. TestBed runs change detection itself and cannot
reproduce this, so the tests pin the call instead.

**3. i18n tokens leak, and no serving route fixes it.** The form rendered `_#(Name)` and
`_#(enter.value)` verbatim. `LocalizationService.TRANSLATED_RESOURCES` and
`TemplateLocalizationController`'s mapping both covered only `resources/app/*.js` and
`resources/em/*.js`, while the elements bundle builds to `resources/elements/`. **Nothing ever
substituted those tokens, in any deployment.** Both now cover `/elements/*.js`.

> Pre-existing and affecting *every* embed element — the built bundle carries 353 `_#(`
> occurrences. `viewer-element/` has the identical gap and is deliberately left alone as out of
> scope. Note the cost: the localisation cache now processes the elements bundle once per locale.

**4. The annotation tree offers non-JDBC sources it cannot expand.** A registered Cassandra source
appeared beside the JDBC ones with `leaf=false`; expanding it returned **200 with `children: []`**,
forever, and indistinguishable from a database still loading. It does **not** throw
`UnsupportedDatasourceException`, which the design had assumed — the quieter and worse of the two.
Non-relational sources are now marked leaves. They stay in the tree on purpose: the source *is*
usable through a worksheet built on it, and hiding it would send the operator hunting for something
they can plainly see in StyleBI. A failed or empty lookup answers "not flagged" — that is not
evidence a source is non-relational, and getting it wrong would collapse a real database to a leaf
and put its tables out of reach.

## A fifth non-defect

The plan carried a report that `listing/MongoDB` and `listing/REST` return 500, and asked for a
server fix. **Neither is a listing name.** The real ones are `MongoDB REST`, `REST JSON` and
`REST XML`, and all three seed 200. No server change was made. The picker cannot produce a bad name —
it only clicks names from the selection view — and the `listing-name` attribute, the one route an
embedder can get wrong, now names the type it could not load.

## Verification

- **20** Testing-Library tests across the service and component specs
- **6** JUnit tests for the tree-filtering decision, including all four ways it must answer "no"
- `mvn -pl community/core compile` clean

**Pre-existing failures, confirmed not regressions.** `ng run portal:test-tl` with no filter fails to
collect **every** spec — 403 files at the base commit `c82528a0d`, 405 on this branch, the difference
being exactly the two spec files added here, with the same
`TypeError: Class extends value undefined` out of `vitest-setup.ts`'s circular-import workaround. The
suite runs only under `--include` today. That is worth fixing, but not here.

## Not verified

The i18n fix is a Java change. It compiles, but confirming the tokens actually resolve needs a
rebuilt container image, which was not done.


larryliang-inetsoft:
Closing in favour of the native portal implementation in [stylebi-wiz#1556](https://github.com/inetsoft-technology/stylebi-wiz/pull/1556), which solves the same problem — registering non-JDBC data sources from the wiz portal — without depending on a StyleBI release or the embed bundle, and covers edit as well as create.

The two **server-side** fixes from this PR are independent of which UI wins and continue in #4581: the elements bundle never receiving i18n substitution, and the wiz annotation tree offering non-JDBC sources it cannot expand. The second is more relevant with #1556 landing, not less.

(This PR was also opened against the wrong base — `main` rather than `stylebi-wiz-main-rebase` — which is why it showed 590 files. #4581 has the right base and is 4 files.)

PR #4581, opened 2026-08-12
Stop offering non-JDBC sources as expandable in the wiz annotation tree

larryliang-inetsoft:
**Branch:** `feature-76074-server-fixes` · base `stylebi-wiz-main-rebase` · Redmine #76075

> **Scope reduced.** This PR originally carried a second commit that claimed to fix i18n token
> leakage in the elements bundle. Building the image proved that fix **inert**; it has been dropped.
> See the comment below for what the evidence actually showed. What remains is one fix, verified
> live.

## Non-JDBC sources are offered as expandable in the wiz annotation tree

A registered Cassandra source appeared in the wiz annotation tree beside the JDBC ones with
`leaf=false`. Expanding it returned **HTTP 200 with `children: []`** — every time, and
indistinguishable from a database that has not finished loading.

It does **not** throw `UnsupportedDatasourceException`, which is what the design assumed. Nothing
surfaces at all, which is the quieter and worse of the two failures.

Non-relational sources extend `TabularDataSource`, so the metadata calls behind an expansion have
nothing to return. They are now marked leaves.

They **stay in the tree** on purpose: the source *is* usable, through a worksheet built on it, and
hiding it would send the operator hunting for something they can plainly see in StyleBI.

A lookup that fails or comes back empty answers "not flagged". That is not evidence a source is
non-relational, and getting it wrong would collapse a real database to a leaf and put every table in
it out of reach — far worse than leaving one node pointlessly expandable.

This gets more reachable as registering non-JDBC sources from the wiz portal gets easier, which is
exactly what [stylebi-wiz#1556](https://github.com/inetsoft-technology/stylebi-wiz/pull/1556) does.

## Verification

**Verified live** on a locally built image (`local-1180`, full reactor). With a Cassandra source
registered, the wiz asset tree returns:

```
inventree           leaf=False      ← 8 JDBC sources, all still expandable
mysqltest           leaf=False
odoo                leaf=False
olist               leaf=False
openproject         leaf=False
orders              leaf=False
sakila              leaf=False
suitecrm            leaf=False
tab-verify-renamed  leaf=True       ← the Cassandra source
```

Reproduced before the fix on the same deployment: the same node came back `leaf=false` and expanded
to `children: []`.

**6 JUnit tests**, covering all four ways the check must answer "no" — JDBC source, unresolvable
source, a lookup that throws, and a non-data-source entry. `mvn -pl community/core test` BUILD
SUCCESS.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
@@ -1264,7 +1264,8 @@ public class MetadataApiService {
       return response;
    }
 
-   private TreeNodeModel filterWizTree(TreeNodeModel node) {
+   // Package-private so a test can drive the real filter rather than only its helper.
+   TreeNodeModel filterWizTree(TreeNodeModel node) {
       return filterWizTree(node, new HashMap<>(), new HashMap<>());
    }
 
--- a/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
@@ -1286,6 +1286,15 @@ public class MetadataApiService {
       TreeNodeModel.Builder builder = TreeNodeModel.builder().from(node);
       builder.children(new ArrayList<>());
 
+      // A non-relational source has no catalog/schema/table metadata to hand back, so expanding it
+      // returns an empty child list forever -- it looks identical to a database that is still
+      // loading. Marking it a leaf says plainly that there is nothing to open. It stays in the tree
+      // on purpose: the source IS usable, through a worksheet built on it, and hiding it would send
+      // the operator hunting for something they can see in StyleBI.
+      if(isNonJdbcDataSource(entry)) {
+         builder.leaf(true);
+      }
+
       for(TreeNodeModel child : node.children()) {
          TreeNodeModel filteredChild = filterWizTree(child, filterCache, partitionCache);
 
@@ -1295,6 +1304,35 @@ public class MetadataApiService {
       }
 
       return builder.build();
+   }
+
+   /**
+    * Whether this entry is a data source that is not relational.
+    *
+    * Tabular/NoSQL sources extend TabularDataSource rather than JDBCDataSource; the metadata calls
+    * behind a tree expansion have nothing to return for them.
+    *
+    * Answers false when the source cannot be resolved. A lookup failure is not evidence that a
+    * source is non-relational, and treating it as such would collapse a real database to a leaf and
+    * make its tables unreachable -- far worse than leaving one node expandable.
+    */
+   // Package-private so the tree-filtering decision can be tested directly.
+   boolean isNonJdbcDataSource(AssetEntry entry) {
+      if(entry == null || !entry.isDataSource()) {
+         return false;
+      }
+
+      try {
+         XDataSource dataSource = xrepository.getDataSource(entry.getPath());
+
+         return dataSource != null && !(dataSource instanceof JDBCDataSource);
+      }
+      catch(Exception e) {
+         log.debug("Could not resolve data source '{}' while filtering the wiz tree: {}",
+                   entry.getPath(), e.getMessage());
+
+         return false;
+      }
    }
 
    private boolean shouldHideWizTreeNode(
--- a/core/src/test/java/inetsoft/web/wiz/service/MetadataApiServiceNonJdbcTreeTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/service/MetadataApiServiceNonJdbcTreeTest.java
@@ -24,15 +24,21 @@ import inetsoft.uql.asset.AssetEntry;
 import inetsoft.uql.asset.AssetRepository;
 import inetsoft.uql.jdbc.JDBCDataSource;
 import inetsoft.uql.tabular.TabularDataSource;
+import inetsoft.uql.xmla.XMLADataSource;
+import inetsoft.web.composer.model.TreeNodeModel;
 import inetsoft.web.composer.AssetTreeService;
 import inetsoft.web.portal.controller.database.DataSourceService;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
+
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * A non-relational data source has no catalog/schema/table metadata behind it, so expanding it in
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft:
**Retracting the i18n commit — it was inert.** Building the image and testing it is what showed this; I had it wrong.

The commit added `/elements/*.js` to `TemplateLocalizationController`'s mapping and to `LocalizationService.TRANSLATED_RESOURCES`, on the reasoning that the elements bundle builds to `resources/elements/` and no glob covered it. That much is true. What I did not check is whether StyleBI ever *serves* that directory.

It does not:

| | |
|---|---|
| `GET /app/main.js` | 200, 845 KB, **0** raw `_#(` tokens — the existing localization works |
| `GET /elements/main.js` | **404** |
| `inetsoft-web` jar | 105 `resources/app/*.js` entries, **0** for `resources/elements/` |
| every other jar | no `resources/elements/` |
| running container | no such path anywhere on disk |

The elements bundle is never packaged into the server. It ships **only** through the npm package `@inetsoft-technology/stylebi-embed-elements`, and is served by whoever embeds it — for us, the wiz portal's own static server. It never passes through StyleBI's localization filter, so a serve-time mapping can never fire.

The on-disk bundle carries **400** raw tokens in `main.js` alone, so **the leak is real and still open** — my fix simply was not one. The real fix has to happen at *build* time, when the npm package is assembled, not at serve time. Redmine #76075 has been updated with the corrected diagnosis.

Practical urgency is lower than it looked: the tokens I originally hit were in the datasource editor, which reached the browser only through the embedded-element approach that has since been dropped in favour of stylebi-wiz#1556. Whatever tokens the chart/viewer/table elements carry do still leak in the wiz portal today.

What remains in this PR is the annotation-tree fix, now verified live on `local-1180`.

core/src/test/java/inetsoft/web/wiz/service/MetadataApiServiceNonJdbcTreeTest.java
```diff
@@ -0,0 +1,110 @@
...
+                                    mock(AssetRepository.class), mock(AssetTreeService.class),
+                                    new ObjectMapper());
+   }
+
+   private AssetEntry dataSourceEntry(String name) {
+      return new AssetEntry(AssetRepository.GLOBAL_SCOPE, AssetEntry.Type.DATA_SOURCE, name, null);
+   }
+
+   @Test
+   void tabularSourceIsNonJdbc() throws Exception {
```
larryliang-inetsoft, line 65: **The tests exercise the helper, not the fix.**

Every case here calls the predicate directly; nothing drives `filterWizTree`. Delete the production change — or invert it so the node is *hidden* rather than marked a leaf — and all six still pass. The test pins the helper's truth table, which was never in doubt; it does not pin the behaviour the PR exists to produce.

Also uncovered: `XMLADataSource`, the other non-JDBC family in this tree. `TabularDataSource` is represented, JDBC is represented, XMLA is not.

Worth one test that runs a tree containing a JDBC source, a tabular source and an XMLA source through `filterWizTree` and asserts the resulting `leaf` flags.

core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
```diff
@@ -1286,6 +1286,15 @@ private TreeNodeModel filterWizTree(
       TreeNodeModel.Builder builder = TreeNodeModel.builder().from(node);
       builder.children(new ArrayList<>());
 
+      // A non-relational source has no catalog/schema/table metadata to hand back, so expanding it
+      // returns an empty child list forever -- it looks identical to a database that is still
+      // loading. Marking it a leaf says plainly that there is nothing to open. It stays in the tree
+      // on purpose: the source IS usable, through a worksheet built on it, and hiding it would send
+      // the operator hunting for something they can see in StyleBI.
+      if(isNonJdbcDataSource(entry)) {
```
larryliang-inetsoft, line 1295: **The predicate is correct, but the condition is narrower than the symptom.**

I checked this against the type hierarchy rather than taking it on faith, and "not JDBC" does exactly match "will have no children here": the tree's DATA_SOURCE children come only from `AbstractAssetEngine.getPhysicalEntries`, which itself hard-gates on `instanceof JDBCDataSource`; logical models are excluded because wiz passes `includeModel=false`; and cubes need `includeColumns`, which wiz also passes as false. All three failure paths (null source, non-JDBC-typed null from `isSupported`, thrown exception) fail open, and the registry lookup key matches the full name `getPhysicalEntries` uses, including folder-nested sources. So the change is sound as far as it goes.

What it does not cover is the same user-visible symptom arriving by two other routes:

- A **JDBC** source expands to `children: []` forever when the caller lacks `PHYSICAL_TABLE` ACCESS. The wiz tree never sets `PORTAL_DATA`, so that gate does apply here.
- Hive with the driver disabled behaves the same way.

That permission is already computed in this same request as `sqlEnabled`, so folding it into the leaf decision looks cheap and would make the fix cover the actual complaint rather than one cause of it.

larryliang-inetsoft, line 1295: **Second, more important: the `leaf` flag may have no consumer.**

The only client of `/datasets/asset_tree` I can find is `portal/src/components/databaseAnnotate/AssetSelectionStep.tsx`, and it does not read `leaf`. It derives expandability from `AssetType` (`hasChildren` is true for `DATA_SOURCE_FOLDER`) and treats any `DATA_SOURCE` node as a selectable, annotatable database.

If that is right, this PR sets a flag nobody reads: the Cassandra source stays checkable in the picker and still ships into an annotation run with no tables behind it — which is the failure the PR is meant to prevent.

Please confirm whether some other caller consumes `leaf`. If not, the fix needs a client-side half (or the node needs to be made non-selectable server-side) before it changes anything a user would notice.

larryliang-inetsoft (commented):
## Review — non-JDBC sources in the wiz annotation tree

Three findings. The change itself is correct — I verified the predicate against the type hierarchy rather than assuming, and "not JDBC" genuinely does match "has no children in this tree" for all three concrete `XDataSource` families. Failure paths fail open, and the registry key matches folder-nested sources.

Two things keep me from calling it done:

1. **The `leaf` flag looks unconsumed.** The one client of this endpoint computes expandability from `AssetType` and never reads `leaf`, so the Cassandra source would remain selectable and still enter an annotation run empty. If that reading is right, the user-visible symptom is unchanged by this PR.

2. **The tests never call `filterWizTree`.** They exercise the helper only, so removing the fix leaves them green — and `XMLADataSource`, the other non-JDBC family, is untested.

The third is a scope note: a JDBC source lacking `PHYSICAL_TABLE` ACCESS produces the identical empty-expansion symptom, and that permission is already computed in the same request.

None of this is a reason to revert — it is sound as far as it goes. The question is whether it currently changes anything the user sees.

larryliang-inetsoft:
**Both findings addressed. You were right that this PR changed nothing a user could see.**

**The `leaf` flag had no consumer.** Confirmed: `AssetSelectionStep.tsx` derives expandability from `hasChildren(node)`, which is `node.data.type == AssetType.DATA_SOURCE_FOLDER` — purely type-based. `node.leaf` is never read. (The `leaf*` identifiers in that file are a local selection variable, which is what makes it look covered at a glance.)

The client half is now in **[stylebi-wiz#1561](https://github.com/inetsoft-technology/stylebi-wiz/pull/1561)**: a leaf DATA_SOURCE is unselectable, rendered disabled, and labelled "no tables to annotate" so the reason is visible rather than the row being silently inert. Reads `leaf === true`, so an older server that never sets the flag leaves every source selectable.

Writing it turned up a second gap your finding implies: the guard belongs in `handleAssetSelect` too, not only in `isSelectableNode` — the direct-select branch added the id without consulting it, so the folder path was filtered while a direct click was not. Deselect stays ungated so an id that somehow got in can always come out.

**The tests did not pin the fix.** Also confirmed — stubbing the production change out left all six green. There is now one that drives `filterWizTree` itself and fails with "a tabular source becomes a leaf" when the fix is removed. It covers **XMLADataSource** as well, the family you noted was untested. `filterWizTree` is package-private for this.

Tests 6 → 7 here, plus 3 on the wiz side.

**Third finding — scope.** Agreed and not addressed here: a JDBC source lacking `PHYSICAL_TABLE` ACCESS produces the identical empty expansion, and that permission is computed in the same request. Worth its own change; the client-side guard above does not help there, since such a source arrives with `leaf=false`.

larryliang-inetsoft:
Correction to the link above: the client half is **[stylebi-wiz#1568](https://github.com/inetsoft-technology/stylebi-wiz/pull/1568)**, not #1561 — I quoted the number before the PR was created.

fredaan:
Stop offering non-JDBC sources as expandable in the wiz annotation tree
https://github.com/inetsoft-technology/stylebi/pull/4581

Do not offer a data source with no tables to annotate
https://github.com/inetsoft-technology/stylebi-wiz/pull/1568


The changes in these two PRs break data labeling for the tabular data source, so after discussing with Larry, we closed them.

PR #4582, opened 2026-08-12
Add the wiz tabular data source API (server half of stylebi-wiz#1556)

larryliang-inetsoft:
**Branch:** `feature-wiz-tabular-controller` · base `stylebi-wiz-main-rebase`

The StyleBI half of tabular data source create/edit.
[stylebi-wiz#1556](https://github.com/inetsoft-technology/stylebi-wiz/pull/1556) builds the portal
and wiz-services layers against these seven endpoints. **None of them existed** — I searched all 253
remote branches — so that PR currently calls routes that 404. This is the missing half; the two
should land together.

Built to the Layer-1 contract in that PR's own spec (`docs/superpowers/specs/2026-08-12-tabular-datasource-design.md`),
including its class name, package, injected dependencies and authorization rules.

```
GET  /api/wiz/tabular/listings              -> WizTabularListings
GET  /api/wiz/tabular/listing?name=         -> DataSourceDefinition   (pass-through)
GET  /api/wiz/tabular/definition?path=      -> DataSourceDefinition   (pass-through)
POST /api/wiz/tabular/refresh               -> DataSourceDefinition
POST /api/wiz/tabular/create                -> WizTabularSaveResult
POST /api/wiz/tabular/update?path=          -> WizTabularSaveResult
GET  /api/wiz/tabular/check-duplicate?name= -> { duplicate }
```

Separate from `WizDatabaseController` because tabular and JDBC are unrelated models. What they share
is the authorization stance.

## Decisions worth a reviewer's attention

**`DataSourceDefinition` and `TabularView` pass through unflattened** — the opposite of the JDBC
endpoints. `tabularView` *is* the UI description; there is no dimension to reduce, and trimming
removes something the renderer needs.

**Keys are not labels.** `getDisplayName()` is `getName()` run through a resource bundle for the
caller's locale, and StyleBI's own `getDataSourceListing` matches on the *display* name — so the
portal's key set shifts when the locale does. These endpoints key on the stable `getName()` and
translate only at the edge (`label`, `categoryLabel`), so a stored choice survives a locale switch.
`/tabular/listing?name=` resolves the stable name to a listing here and hands the service the
display name it expects.

**JDBC listings are excluded, and a null check would not have done it.** Measured against a running
server: a JDBC listing seeds a `tabularView` with **zero views** — an empty form, not `null` —
while Cassandra gives 10, MongoDB REST 12, REST JSON 21. The discriminator is therefore
`TabularDataSource`, which means instantiating each listing's data source. They are plain objects
with default fields, not connections. A listing that cannot be instantiated costs its own entry in
the picker rather than failing the whole request.

**Refresh echoes the sequence number back.** The caller increments it before each refresh and
discards any response carrying a lower one. Drop the echo and the field returns 0, every response
looks current, and a slow one silently overwrites a newer form. The portal does the same; there is a
test for it.

**Duplicate names are checked before writing, not inferred from the failure.** `MessageException`
carries a Catalog-translated sentence, and string-matching it breaks in any other locale. Only
`INVALID_FOLDER` still rests on that comparison — deliberately the last one that does. Anything
unrecognized becomes `UNKNOWN` and is logged verbatim rather than swallowed.

**Authorization is at the entry point**, not left to the service: WRITE on the path for
`definition`/`update`, the folder rule (with the root fallback to `CREATE_DATA_SOURCE`) for
`create`, and no path gate on `refresh`, which touches no stored data and recomputes from a
caller-supplied definition. No local `@ExceptionHandler` — it would take precedence over
`WizControllerErrorHandler` and degrade a 403 into a 400.

## Verification

**10 JUnit tests**, all passing, `mvn -pl community/core test` BUILD SUCCESS. They cover the paths
where a wrong answer is worst — a client branches on `reason` and cannot see the sentence behind it:

- create: saved path; duplicate rejected *without writing*; unrecognized refusal → `UNKNOWN`;
  permission denial leaves as `SecurityException` (so it becomes a 403) and writes nothing
- update: `DATASOURCE_LOST` when it vanished; rename returns the new path; rename onto an existing
  name rejected; saving under its own name is **not** a duplicate of itself
- refresh: sequence number echoed

Endpoint shapes were checked against a live server through the portal's equivalent routes: MySQL
(jdbc) 0 views, Apache Cassandra 10, MongoDB REST 12, REST JSON 21.

## Not verified

`/tabular/listings` cannot be exercised without a running server — `DataSourceListingService`
requires the Spring context, so it fails in a unit test. That means the tabular/JDBC classification
and the listing sort are covered by reasoning and by the live seed evidence above, not by a test.
**Worth an integration check before merge**, especially that the picker is not empty: an empty type
picker looks exactly like a deployment with no data source plugins installed.

Confirming these routes end to end also needs a rebuilt container image, which was not done — the
local one predates `WizDatabaseController`.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
@@ -37,8 +37,10 @@ import inetsoft.web.wiz.model.WizTabularSaveResult;
 import inetsoft.web.wiz.request.WizTabularCreateRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.web.bind.annotation.*;
+import org.springframework.web.server.ResponseStatusException;
 
 import java.security.Principal;
 import java.util.*;
@@ -137,8 +139,6 @@ public class WizTabularController {
     * @param principal the current user.
     *
     * @return the seeded definition, including the form to render.
-    *
-    * @throws java.io.FileNotFoundException if no tabular listing has that name.
     */
    @GetMapping(value = "/tabular/listing", produces = MediaType.APPLICATION_JSON_VALUE)
    public DataSourceDefinition getTabularListing(@RequestParam("name") String name,
@@ -149,9 +149,12 @@ public class WizTabularController {
       // name, and this endpoint's key is the stable one.
       DataSourceListing listing = findTabularListing(name);
 
+      // 404 rather than a bare exception: unhandled ones fall through to a generic 500, and asking
+      // for a type that does not exist is the caller's mistake, not a server fault. Measured — a
+      // JDBC name such as "MySQL" reached here and returned 500.
       if(listing == null) {
-         throw new java.io.FileNotFoundException(
-            "No tabular data source listing named \"" + name + "\"");
+         throw new ResponseStatusException(
+            HttpStatus.NOT_FOUND, "No tabular data source listing named \"" + name + "\"");
       }
 
       return datasourcesService.getDataSourceFromListing(listing.getDisplayName());
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
@@ -1,0 +1,411 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.web.wiz.controller;
+
+import inetsoft.sree.security.ResourceAction;
+import inetsoft.sree.security.ResourceType;
+import inetsoft.sree.security.SecurityEngine;
+import inetsoft.uql.DataSourceListing;
+import inetsoft.uql.DataSourceListingService;
+import inetsoft.uql.XDataSource;
+import inetsoft.uql.tabular.TabularDataSource;
+import inetsoft.util.Catalog;
+import inetsoft.util.MessageException;
+import inetsoft.web.portal.data.DataSourceDefinition;
+import inetsoft.web.portal.data.DatasourcesService;
+import inetsoft.web.security.PermissionPath;
+import inetsoft.web.security.RequiredPermission;
+import inetsoft.web.security.Secured;
+import inetsoft.web.wiz.model.WizTabularListing;
+import inetsoft.web.wiz.model.WizTabularListings;
+import inetsoft.web.wiz.model.WizTabularSaveResult;
+import inetsoft.web.wiz.request.WizTabularCreateRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.MediaType;
+import org.springframework.web.bind.annotation.*;
+
+import java.security.Principal;
+import java.util.*;
+
+/**
+ * The wiz API for creating and editing tabular (non-JDBC) data sources.
+ *
+ * <p>Separate from {@code WizDatabaseController} because tabular and JDBC are two unrelated models;
+ * merging them would make both harder to read. What they share is the authorization stance.</p>
+ *
+ * <h3>Pass-through</h3>
+ *
+ * <p>{@code DataSourceDefinition} and its {@code TabularView} are returned <em>as they are</em>,
+ * with no flattening — the opposite of the JDBC endpoints. The reason is that {@code tabularView}
+ * <em>is</em> the UI description; there is no dimension to reduce, and any trimming removes
+ * information the renderer needs.</p>
+ *
+ * <h3>Authorization</h3>
+ *
+ * <p>Calling the service layer directly is not a permission gate — the equivalent gates the portal
+ * controllers apply are repeated here. Denials leave as {@code SecurityException}, which
+ * {@code WizControllerErrorHandler} turns into a 403; this class must therefore never declare a
+ * local {@code @ExceptionHandler}, which would take precedence over that advice and degrade a 403
+ * into a 400.</p>
+ *
+ * <h3>Keys are not labels</h3>
+ *
+ * <p>{@code DataSourceListing.getDisplayName()} is {@code getName()} translated for the caller's
+ * locale, and StyleBI's own listing lookup matches on the display name — so the portal's key set
+ * shifts when the locale does. This controller keys on the stable {@code getName()} and translates
+ * only at the edge, so a stored choice survives a locale switch.</p>
+ */
+@RestController
+@RequestMapping("/api/wiz")
+public class WizTabularController {
+   public WizTabularController(DatasourcesService datasourcesService,
+                               SecurityEngine securityEngine)
+   {
+      this.datasourcesService = datasourcesService;
+      this.securityEngine = securityEngine;
+   }
+
+   /**
+    * Lists the data source types the tabular editor offers.
+    *
+    * <p>JDBC listings are excluded: they have no {@code TabularView} to render and belong to the
+    * {@code /databases} editor. The only way to tell is to instantiate the listing's data source and
+    * look at its type, which is why this builds one per listing — they are plain objects with
+    * default fields, not connections.</p>
+    *
+    * @param principal the current user.
+    *
+    * @return the offered listings and the raw categories to group them by.
+    */
+   @GetMapping(value = "/tabular/listings", produces = MediaType.APPLICATION_JSON_VALUE)
+   public WizTabularListings getTabularListings(Principal principal) {
+      Catalog catalog = Catalog.getCatalog(principal);
+      List<WizTabularListing> listings = new ArrayList<>();
+
+      for(DataSourceListing listing : DataSourceListingService.getAllDataSourceListings(true)) {
+         if(!isTabular(listing)) {
+            continue;
+         }
+
+         String category = listing.getCategory();
+         listings.add(new WizTabularListing(
+            listing.getName(),
+            listing.getDisplayName(),
+            category,
+            category == null ? null : catalog.getString(category),
+            listing.getIcon(),
+            listing.getKeywords() == null
+               ? List.of() : Arrays.asList(listing.getKeywords())));
+      }
+
+      listings.sort(Comparator.comparing(WizTabularListing::name, String.CASE_INSENSITIVE_ORDER));
+
+      List<String> categories = listings.stream()
+         .map(WizTabularListing::category)
+         .filter(Objects::nonNull)
+         .distinct()
+         .sorted()
+         .toList();
+
+      return new WizTabularListings(listings, categories);
+   }
+
+   /**
+    * Seeds a NEW data source of the given type.
+    *
+    * <p>This is the only way to obtain a usable blank form. Posting a bare name and type to
+    * {@code /tabular/refresh} yields {@code tabularView: null}, because refresh recomputes an
+    * existing view rather than creating one.</p>
+    *
+    * @param name      the listing's stable {@link DataSourceListing#getName()}, not its display name.
+    * @param principal the current user.
+    *
+    * @return the seeded definition, including the form to render.
+    *
+    * @throws java.io.FileNotFoundException if no tabular listing has that name.
+    */
+   @GetMapping(value = "/tabular/listing", produces = MediaType.APPLICATION_JSON_VALUE)
+   public DataSourceDefinition getTabularListing(@RequestParam("name") String name,
+                                                 Principal principal)
+      throws Exception
+   {
+      // Resolved here rather than passed straight through: the service looks listings up by display
+      // name, and this endpoint's key is the stable one.
+      DataSourceListing listing = findTabularListing(name);
+
+      if(listing == null) {
+         throw new java.io.FileNotFoundException(
+            "No tabular data source listing named \"" + name + "\"");
+      }
+
+      return datasourcesService.getDataSourceFromListing(listing.getDisplayName());
+   }
+
+   /**
+    * Loads an existing data source for editing.
+    *
+    * @param path      the data source's full repository path.
+    * @param principal the current user.
+    *
+    * @return the stored definition and its form.
+    */
+   @GetMapping(value = "/tabular/definition", produces = MediaType.APPLICATION_JSON_VALUE)
+   @Secured({
+      @RequiredPermission(
+         resourceType = ResourceType.DATA_SOURCE, actions = ResourceAction.WRITE
+      )
+   })
+   public DataSourceDefinition getTabularDefinition(
+      @PermissionPath @RequestParam("path") String path, Principal principal) throws Exception
+   {
+      return datasourcesService.getDataSourceDefinition(path, principal);
+   }
+
+   /**
+    * Recomputes the form after a field the rest of it depends on changed.
+    *
+    * <p>No path-level permission: this touches no stored data. The definition arrives whole from the
+    * caller and the server recomputes a view from it. A login is still required, which
+    * {@code WizServiceAuthenticationFilter} enforces.</p>
+    *
+    * <p>The sequence number is echoed back deliberately. The caller increments it before each
+    * refresh and discards any response carrying a lower one; drop the echo and a slow response
+    * silently overwrites a newer form. The portal does the same.</p>
+    *
+    * <p>Unlike the portal this does not bind the call to an HTTP session. Refresh was verified
+    * stateless, and wiz reaches this server-to-server where a session would be per-call noise
+    * rather than the user's. Worth revisiting if a source that keeps session-scoped credentials
+    * misbehaves under refresh.</p>
+    *
+    * @param definition the current form state.
+    *
+    * @return the recomputed definition.
+    */
+   @PostMapping(value = "/tabular/refresh", produces = MediaType.APPLICATION_JSON_VALUE)
+   public DataSourceDefinition refreshTabularView(@RequestBody DataSourceDefinition definition) {
+      DataSourceDefinition refreshed = datasourcesService.refreshTabularView(definition);
+      refreshed.setSequenceNumber(definition.getSequenceNumber());
+
+      return refreshed;
+   }
+
+   /**
+    * Creates a data source.
+    *
+    * @param request   the target folder and the definition to save.
+    * @param principal the current user.
+    *
+    * @return the saved path, or the reason the save was rejected.
+    */
+   @PostMapping(value = "/tabular/create", produces = MediaType.APPLICATION_JSON_VALUE)
+   public WizTabularSaveResult createTabularDataSource(
+      @RequestBody WizTabularCreateRequest request, Principal principal) throws Exception
+   {
+      DataSourceDefinition definition = request.definition();
+
+      if(definition == null) {
+         return WizTabularSaveResult.failed(WizTabularSaveResult.UNKNOWN);
+      }
+
+      String parentPath = emptyToNull(request.parentPath());
+      requireCreatePermission(parentPath, principal);
+
+      // Carried on the definition because that is where createNewDataSource reads it from; the
+      // request field exists so the gate above can run before anything is written.
+      definition.setParentPath(parentPath == null ? "" : parentPath);
+      String fullPath = fullPath(parentPath, definition.getName());
+
+      // Checked up front rather than inferred from the failure. A duplicate otherwise surfaces as a
+      // translated sentence that has to be string-matched, which breaks in any other locale.
+      if(datasourcesService.checkDuplicate(fullPath)) {
+         return WizTabularSaveResult.failed(WizTabularSaveResult.DUPLICATE_NAME);
+      }
+
+      try {
+         datasourcesService.createNewDataSource(definition, false, principal);
+      }
+      catch(MessageException e) {
+         return toFailure(e, "create", fullPath);
+      }
+
+      return WizTabularSaveResult.ok(fullPath);
+   }
+
+   /**
+    * Updates an existing data source.
+    *
+    * <p>Gated at the entry point rather than relying on the service's own check, which happens only
+    * after the stored definition has been read.</p>
+    *
+    * @param path       the data source's current full repository path, and the sole source of the
+    *                   name to update. A client should leave whatever {@code oldName} it received
+    *                   untouched rather than compute one.
+    * @param definition the new settings.
+    * @param principal  the current user.
+    *
+    * @return the saved path, which differs from {@code path} when the source was renamed, or the
+    *         reason the save was rejected.
+    */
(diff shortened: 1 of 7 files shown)


larryliang-inetsoft:
**Verified live.** Built a full-reactor image from this branch (`local-1180`, then `local-1181` with the 404 fix below) and exercised all seven endpoints against it. A full reactor was required, not the `-pl community/core` shortcut — this adds a new `@RestController` bean, and the shortcut leaves the enterprise module's AOT bean definitions stale.

| Endpoint | Result |
|---|---|
| `/tabular/listings` | **80 listings, 11 categories** out of 130 total. MySQL (jdbc) correctly absent, MongoDB REST present, `name`/`label`/`category` all populated |
| `/tabular/listing?name=Apache Cassandra` | 200, type `Cassandra`, 10 views |
| `/tabular/definition?path=` | 200, 10 views |
| `/tabular/refresh` | 200, **sequenceNumber 7 echoed back** |
| `/tabular/create` | `{ok:true, path:"tab-verify"}` |
| `/tabular/create` again | `{ok:false, reason:"DUPLICATE_NAME"}` |
| `/tabular/update?path=` (rename) | `{ok:true, path:"tab-verify-renamed"}`, confirmed in the datasource list |
| `/tabular/update?path=gone-for-good` | `{ok:false, reason:"DATASOURCE_LOST"}` |
| `/tabular/check-duplicate?name=` | `{duplicate:true}` for sakila, `false` for an unused name |

The created data source was deleted afterwards.

That closes the gap I flagged in the PR body: **`/tabular/listings` is exercised now**, and it is not empty — 80 tabular types with JDBC correctly filtered out. That was the one behaviour no unit test could reach, since `DataSourceListingService` needs the Spring context.

**One defect the live run found**, fixed in `c051aa060`: an unknown listing name returned **500**, not 404. `FileNotFoundException` is not mapped by `WizControllerErrorHandler`, so it fell through to the generic handler and read as a server fault rather than the caller's mistake. Now `ResponseStatusException(NOT_FOUND)`, matching how `WizDatabaseController` rejects an unknown database type. Re-verified on `local-1181`: `MySQL` → 404, `no-such-type` → 404, `Apache Cassandra` → 200.

core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
```diff
@@ -0,0 +1,414 @@
...
+      definition.setParentPath(parentPath == null ? "" : parentPath);
+      String fullPath = fullPath(parentPath, definition.getName());
+
+      // A rename onto an existing name, not the source's own path.
+      if(!fullPath.equals(path) && datasourcesService.checkDuplicate(fullPath)) {
+         return WizTabularSaveResult.failed(WizTabularSaveResult.DUPLICATE_NAME);
+      }
+
+      try {
+         datasourcesService.updateDataSource(path, definition, principal);
```
larryliang-inetsoft: **Every folder-scoped tabular datasource is un-editable. Confirmed by reading the service.**

`DatasourcesBaseService.updateDataSource` (line 405) builds its own lookup key by re-prefixing the parent:

```java
String parentPath = "".equals(definition.getParentPath()) ? "" : definition.getParentPath() + "/";
String oldName = parentPath + name;
```

This method sets `definition.setParentPath(parentOf(path))` and then passes `path` — the **full** path — as `name`. For a source at `folder/mongo` that resolves to `folder/` + `folder/mongo` = `folder/folder/mongo`. `repository.getDataSource` returns null, `oldSrc == null`, and it throws `saveDataSourceLost`, which reaches the user as `reason:"UNKNOWN"`.

Root-level sources work only by accident: `parentOf` returns null, the prefix is empty, and `oldName` happens to equal the name.

The fix is to pass the bare name (`definition.getName()`'s stored counterpart, i.e. the last segment of `path`) rather than the full path, since `parentPath` is already carried on the definition.

Note this breaks the exact edit flow stylebi-wiz#1556 ships, so the two need to land together with this fixed.

core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
```diff
@@ -0,0 +1,414 @@
...
+    * rather than the user's. Worth revisiting if a source that keeps session-scoped credentials
+    * misbehaves under refresh.</p>
+    *
+    * @param definition the current form state.
+    *
+    * @return the recomputed definition.
+    */
+   @PostMapping(value = "/tabular/refresh", produces = MediaType.APPLICATION_JSON_VALUE)
+   public DataSourceDefinition refreshTabularView(@RequestBody DataSourceDefinition definition) {
+      DataSourceDefinition refreshed = datasourcesService.refreshTabularView(definition);
```
larryliang-inetsoft: **`TabularUtil.setSessionId` is never called here, and nothing anywhere clears it — a refresh can pick up another user's session id.**

`TabularUtil.sessionId` is a `static ThreadLocal<String>` (TabularUtil.java:1023), consumed at line 647 via `method.invoke(bean, sessionId.get())`. `setSessionId` (line 949) only ever sets it; there is no remove/clear on any path.

The portal sets it at four entry points (`DataSourceController` lines 550, 572, 580, 673). This controller deliberately does not. On a pooled request thread that does not mean "no session id" — it means **whatever the previous request on that thread left behind**, which may be a different user's.

The doc comment above anticipates the area — *"Worth revisiting if a source that keeps session-scoped credentials misbehaves under refresh"* — but frames it as a functional risk. The cross-user case is the one to worry about: a connector that resolves an OAuth token by session id would resolve someone else's.

Either set it explicitly for this request, or clear it before invoking so a connector reliably sees null rather than a stale value.

core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
```diff
@@ -0,0 +1,414 @@
...
+    * stateless, and wiz reaches this server-to-server where a session would be per-call noise
+    * rather than the user's. Worth revisiting if a source that keeps session-scoped credentials
+    * misbehaves under refresh.</p>
+    *
+    * @param definition the current form state.
+    *
+    * @return the recomputed definition.
+    */
+   @PostMapping(value = "/tabular/refresh", produces = MediaType.APPLICATION_JSON_VALUE)
+   public DataSourceDefinition refreshTabularView(@RequestBody DataSourceDefinition definition) {
```
larryliang-inetsoft: **`/tabular/refresh` has no permission gate, and "touches no stored data" doesn't cover what it executes.**

The rationale above is that refresh reads nothing stored and only recomputes a view from the posted definition. That answers *data exposure*, but not *side effects*: `refreshTabularView` invokes connector button and editor methods with a caller-controlled `type`, caller-controlled property values, and a caller-controlled `clicked` flag. Those methods are what reach out to a remote endpoint or touch a file.

Combined with the wiz side — no route in `datasourceManagementRoutes.ts` uses `requirePermission` — the effective gate on this endpoint is "is logged in", on the CSRF-exempt `/api/wiz/**` prefix.

For contrast, the other three mutating endpoints here are gated properly (`@Secured` on `/tabular/definition` and `/tabular/update`, imperative `requireCreatePermission` on `/tabular/create`), so this is the odd one out rather than a systemic gap.

core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
```diff
@@ -0,0 +1,414 @@
...
+      }
+
+      try {
+         datasourcesService.createNewDataSource(definition, false, principal);
+      }
+      catch(MessageException e) {
+         return toFailure(e, "create", fullPath);
+      }
+
+      return WizTabularSaveResult.ok(fullPath);
```
larryliang-inetsoft, line 337: **Create returns `ok:true` when nothing was written.**

If `definition.type` is unknown or its connector is unavailable, `createDataSource` returns null and `createNewDataSource` no-ops silently. No exception is thrown, so control falls through to `WizTabularSaveResult.ok(fullPath)` and the editor reports success and navigates to a datasource that does not exist.

Worth checking the source actually exists after the call (the duplicate check above already proves it did not before), or having the service signal the no-op.

core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
```diff
@@ -0,0 +1,414 @@
...
+    * @return the saved path, which differs from {@code path} when the source was renamed, or the
+    *         reason the save was rejected.
+    */
+   @PostMapping(value = "/tabular/update", produces = MediaType.APPLICATION_JSON_VALUE)
+   @Secured({
+      @RequiredPermission(
+         resourceType = ResourceType.DATA_SOURCE, actions = ResourceAction.WRITE
+      )
+   })
+   public WizTabularSaveResult updateTabularDataSource(
```
larryliang-inetsoft, line 364: **Update neither verifies the target is tabular nor honors `definition.type`.**

Nothing here checks that the source at `path` is a `TabularDataSource`. Point this at a JDBC database and `createDataSource(definition, clone(oldSrc))` rewrites it from the tabular definition — renaming it and discarding every JDBC setting — and still returns `ok:true`.

`getTabularDefinition` correctly 422s on a non-tabular path; the write side should be at least as strict as the read side.

core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
```diff
@@ -0,0 +1,414 @@
...
+    *
+    * @return the stored definition and its form.
+    */
+   @GetMapping(value = "/tabular/definition", produces = MediaType.APPLICATION_JSON_VALUE)
+   @Secured({
+      @RequiredPermission(
+         resourceType = ResourceType.DATA_SOURCE, actions = ResourceAction.WRITE
+      )
+   })
+   public DataSourceDefinition getTabularDefinition(
```
larryliang-inetsoft, line 218: **`GET /tabular/definition` returns stored secrets in plaintext — worth confirming this is intended.**

Passwords and API keys come back inside `tabularView` editor values in clear text, which contradicts the sibling `WizAuthentication` contract ("the stored password never leaves the server").

Flagging it as a decision rather than a defect, because the wiz side documents it as deliberate:

> The password rule is HARDER here than on the JDBC side. JDBC never hands out a password and accepts null as "keep the stored one"; a tabular secret lives in `editor.value` in CLEAR TEXT and travels in both directions, because the refresh protocol needs the true value to recompute the tree. So the bodies of refresh/create/update are never logged, here or in the router.

That is a coherent argument with a stated mitigation. But it diverges from the JDBC contract in the same API surface, and the mitigation only covers logging — not the response itself, which reaches a browser. Since `/tabular/definition` is `@Secured` on READ the exposure is bounded, so the question is whether READ on a datasource is meant to imply "may read its credentials". Please confirm explicitly, and consider saying so in this method's javadoc so the divergence is discoverable from the server side too.

core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
```diff
@@ -0,0 +1,414 @@
...
+      if(definition == null) {
+         return WizTabularSaveResult.failed(WizTabularSaveResult.UNKNOWN);
+      }
+
+      String parentPath = emptyToNull(request.parentPath());
+      requireCreatePermission(parentPath, principal);
+
+      // Carried on the definition because that is where createNewDataSource reads it from; the
+      // request field exists so the gate above can run before anything is written.
+      definition.setParentPath(parentPath == null ? "" : parentPath);
```
larryliang-inetsoft, line 305: **Create silently ignores `definition.parentPath`.**

The target folder is read from `request.parentPath()` only. A client that sets the parent on the definition — the same field this method then overwrites two lines down, and the field `updateTabularDataSource` *does* read — gets its source created at the repository root with no error.

Either read the definition's value as a fallback, or reject the request when the two disagree.

core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
```diff
@@ -0,0 +1,414 @@
...
+   /**
+    * Creates a data source.
+    *
+    * @param request   the target folder and the definition to save.
+    * @param principal the current user.
+    *
+    * @return the saved path, or the reason the save was rejected.
+    */
+   @PostMapping(value = "/tabular/create", produces = MediaType.APPLICATION_JSON_VALUE)
+   public WizTabularSaveResult createTabularDataSource(
```
larryliang-inetsoft, line 279: **No name validation: null, blank or slash-bearing names degrade to `UNKNOWN`.**

The JDBC sibling has `requireName`; this has nothing equivalent. A blank name produces a `fullPath` of `parent/` and fails somewhere downstream as a generic `UNKNOWN`, and a name containing `/` silently relocates the source.

A 400 naming the offending field is more useful than a 200 carrying `reason:"UNKNOWN"`.

core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
```diff
@@ -0,0 +1,414 @@
...
+      @RequestBody WizTabularCreateRequest request, Principal principal) throws Exception
+   {
+      DataSourceDefinition definition = request.definition();
```
larryliang-inetsoft: **A literal `null` request body NPEs before the null check.**

`request.definition()` is dereferenced on the record itself — if the body deserializes to `null`, this throws before reaching the `definition == null` guard below, producing a 500 instead of the intended failure result.

core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
```diff
@@ -0,0 +1,414 @@
...
+   /**
+    * Whether a data source already exists at that path, so the editor can reject a name before the
+    * server has to.
+    *
+    * @param name the full repository path to test.
+    *
+    * @return a single-key object rather than a bare boolean, so the response stays extensible.
+    */
+   @GetMapping(value = "/tabular/check-duplicate", produces = MediaType.APPLICATION_JSON_VALUE)
+   public Map<String, Boolean> checkDuplicate(@RequestParam("name") String name) throws Exception {
```
larryliang-inetsoft: **`/tabular/check-duplicate` is an ungated existence oracle.**

No `@Secured`, no principal, and the parameter is an arbitrary repository path — so any logged-in user can enumerate which data sources exist anywhere in the tree, including ones they cannot read.

The editor's need is narrow (reject a name before saving), so scoping the check to the parent folder the caller may write to would keep the feature and close the oracle.

core/src/test/java/inetsoft/web/wiz/controller/WizTabularControllerTest.java
```diff
@@ -0,0 +1,198 @@
...
+   void updateReturnsTheNewPathAfterARename() throws Exception {
+      when(datasourcesService.checkDuplicate("folder/old")).thenReturn(true);
+      when(datasourcesService.checkDuplicate("folder/new")).thenReturn(false);
+
+      WizTabularSaveResult result =
+         controller.updateTabularDataSource("folder/old", definition("new"), principal);
+
+      assertTrue(result.ok());
+      assertEquals("folder/new", result.path());
+      verify(datasourcesService).updateDataSource(eq("folder/old"), any(), eq(principal));
```
larryliang-inetsoft: **The test pins the broken contract rather than catching it.**

`updateDataSource(eq("folder/old"), …)` asserts exactly the full-path call that makes folder-scoped updates fail (see the comment on `WizTabularController.java:298`). Because `DatasourcesService` is fully mocked, the test passes precisely *because* the production call is wrong — fixing the bug will break this assertion, which is backwards.

Three further gaps:
- Nothing exercises the `@Secured` / `@PermissionPath` gates. Direct method calls bypass `SecuredAspect` entirely, so the annotations are untested; the JDBC sibling has a reflective security test worth copying.
- The listings endpoints are uncovered.
- The 404 that is the second commit's whole subject is uncovered.

Suggest at least one test that drives a folder-scoped update through a real (or realistically faked) `DatasourcesBaseService`, so the path arithmetic is exercised rather than asserted.

larryliang-inetsoft (commented):
## Review — wiz tabular data source API

Reviewed the six files against the service layer they delegate to (`DatasourcesBaseService`/`DatasourcesService`), the portal controller they mirror, `TabularUtil`, `SecuredAspect`, and the JDBC sibling `WizDatabaseController`. Eleven findings inline.

**Two I verified by reading the callee, not just this diff, and would fix before merge:**

1. **`WizTabularController.java:298` — folder-scoped updates cannot work.** The full path is passed where `updateDataSource` expects a bare name and re-prefixes `parentPath` itself, producing `folder/folder/name`. Root-level sources work by accident. This breaks the exact edit flow stylebi-wiz#1556 ships.

2. **`WizTabularController.java:205` — the `TabularUtil` session ThreadLocal is never set here and never cleared anywhere.** A refresh can therefore hand a connector method the session id left behind by the previous request on that pooled thread — potentially another user's.

**On permissions, credit where due:** `/tabular/definition` and `/tabular/update` carry `@Secured`, and `/tabular/create` gates imperatively via `requireCreatePermission` *before* writing, with a comment explaining why the field is on the request rather than the definition. That is the right shape. The gaps are `/tabular/refresh` and `/tabular/check-duplicate`, which are the two endpoints with no gate at all.

**One is a question, not a defect:** the plaintext-secret readback on `/tabular/definition` is documented on the wiz side as a deliberate consequence of the refresh protocol. I've written it up as "please confirm intended" rather than as a bug, with the reasoning quoted, since it does diverge from the JDBC contract in the same API surface.

The rest are ordinary hardening: create reporting success on a silent no-op, update not checking the target is tabular, missing name validation, an NPE on a null body, and an ignored `parentPath`.

Worth noting the general pattern across several of these — a failed or no-op operation returning `200 ok:true`. The `WizTabularSaveResult` design is good, with typed reasons instead of string-matched sentences; several paths just don't use it where they should.

core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
fredaan: Confirming: intended, and it cannot be otherwise without breaking the refresh protocol. Two corrections to the framing though, one of which changes the answer.

**The gate is WRITE, not READ.** This endpoint carries `@RequiredPermission(resourceType = DATA_SOURCE, actions = WRITE)`. So the question is not "does READ imply may-read-credentials" — it is "does WRITE imply it", and there the answer is much less troubling: the caller can already overwrite the connection wholesale, point it at a host they control, and read the secret out of what comes back. Withholding it from someone who can rewrite it buys nothing.

**Why masking is not available here, concretely.** The JDBC contract works because its model is a fixed record: the server can substitute a placeholder for `password`, and on save recover the stored value by `oldName` because it knows which field is the password. A tabular definition has no such field. The secret is one `editor.value` among many in a server-generated tree, and the refresh protocol posts that whole tree back so the server can recompute it — `TabularUtil.refreshView` instantiates the connector bean from the definition it is given. A masked value would be fed to the connector as if the user had typed asterisks, so the recomputed tags/labels would be wrong, and the next save would persist the mask.

Making it work would mean the server remembering, per node path, which values were masked and restoring them before recompute — i.e. server-side session state keyed to an editing session, on an endpoint that is deliberately stateless. That is a real design, but it is a different one, and it would have to replace the refresh protocol rather than sit alongside it.

**Where this leaves us.** Same exposure as the native Angular portal, which reads the same definition into the browser — this is not a regression introduced by the wiz surface, and closing it would mean closing it there first. Bounded by WRITE. Mitigated on the transport side (no request/response body logging, in the router or the service).

Agreed on discoverability, and that is a fair hit: the reasoning currently lives only in the wiz-side types, so a server-side reader has no way to know the divergence is deliberate. I will put it in this method's javadoc, pointing at the refresh protocol as the reason, so the next person who notices the asymmetry finds the argument instead of filing it as a bug.

fredaan:
Pushed `1a5f42b60` addressing the review. All 12 defects are fixed; the 13th (plaintext secrets) is answered in-thread as intended-with-reasoning, and now documented in the method's javadoc so the divergence is discoverable from the server side.

Tests: 45 passing (`WizTabularControllerTest` 23, `WizTabularControllerSecurityTest` 13 new, `WizDatabaseControllerSecurityTest` 9 unchanged). The three headline fixes were mutation-checked rather than trusted — reverting each turns a test red:

- full path restored to `updateDataSource` → `expected: <folder/mongo> but was: <folder/folder/mongo>`
- read-side tabular guard removed → `Expected UnsupportedDatasourceException to be thrown, but nothing was thrown`
- refresh gate narrowed back → `refresh_acceptsRootFolderWriteWithoutTheStandaloneGrant » Security mutation`

**Two places I did not follow the suggestion literally**, both worth a look:

**The read side was not already guarded.** The comment on :274 says "`getTabularDefinition` correctly 422s on a non-tabular path; the write side should be at least as strict" — it does not. It delegates to `getDataSourceDefinition`, which only throws when the path resolves to nothing; for a JDBC source `LayoutCreator` builds a `TabularView` and it returns 200. So fixing only the write side would have left the real asymmetry in place, with the editor rendering a blank form for a database it cannot represent. The guard is applied to both sides. It deliberately ignores a path with nothing behind it — reporting that through `UnsupportedDatasourceException` would read as *"not supported for 'unknown' datasources"*, complaining about the wrong thing and newly masking not-found, which both callers already handle.

**The refresh gate is the full create rule, not just `CREATE_DATA_SOURCE`.** My first pass took the narrow reading and it was wrong: a user with `DATA_SOURCE_FOLDER` WRITE on a folder but no standalone grant could create and update a source, then get a 403 from `/tabular/refresh` — the editor breaking halfway through the flow it had just authorized. Refresh is a step inside that flow, so it uses the same rule create does at the root.

One more thing your `:250` finding turned up that is worth stating: I kept `UNKNOWN` for the silent-no-op case rather than adding a reason code. The reasons are a closed set mirrored by a TypeScript union in stylebi-wiz#1556, so a new code would arrive at a client that cannot render it. The diagnosis goes to `LOG.warn` with the path and type instead.

The review was unusually good — reading the callee rather than the diff is what caught the update path arithmetic and the ThreadLocal, and neither would have shown up in the diff alone. Thanks for that.

fredaan (approved):

larryliang-inetsoft:
**All ten findings fixed, and the folder-path bug verified live both ways.** Thanks — the review caught a real defect that both my unit tests and my earlier live run missed.

**Before the fix**, on a running server: create in a folder returned `ok:true`, then update returned exactly what you predicted:

```
create tabfolder/infolder  → {"ok":true,"path":"tabfolder/infolder"}
update tabfolder/infolder  → {"ok":false,"reason":"UNKNOWN"}
```

**After**, on a rebuilt image (`local-1183`):

```
create  → {"ok":true,"path":"tabfolder/infolder"}
update  → {"ok":true,"path":"tabfolder/infolder"}
rename  → {"ok":true,"path":"tabfolder/renamed"}   (persisted; confirmed by reloading)
```

You were exactly right about why it survived: **the old test asserted the broken call**, so it passed *because* production was wrong. The replacement reproduces the callee's own key arithmetic (`oldName = parentPath + name`) and asserts the source resolves — stubbing the fix back out now fails it with "folder-scoped update must resolve the stored source".

Also verified live: a slash-bearing name → **400**, and an update aimed at `sakila` → **422** with the JDBC source left untouched.

Fixed, each with a test:

| Finding | Fix |
|---|---|
| folder-scoped update | pass the bare name |
| TabularUtil ThreadLocal | refresh always overwrites it; `getSession(false)` so a server-to-server caller mints nothing. Added `TabularUtil.getSessionId` to make the leak observable |
| refresh ungated | gated by flow — WRITE when the definition names an existing source, the create rule otherwise |
| check-duplicate oracle | scoped to a folder the caller may write to |
| create ok:true on no-op | existence re-checked after the write |
| update rewrites a JDBC source | 422, matching what the read side already did |
| rename denial as UNKNOWN | DELETE checked up front so it leaves as a 403 |
| name validation | 400 naming the field, mirroring `requireName` |
| null body NPE | null-checked on the record itself |
| create ignores definition.parentPath | honored as a fallback; a disagreement is a 400 |

Not fixed, deliberately: the plaintext secret readback, per your own follow-up confirming it is intended and that WRITE — not READ — is the gate. Your reasoning is in the thread; I have nothing to add to it.

Tests **10 → 23**.

One note for whoever builds this: it adds a constructor parameter, so the `-pl community/core` shortcut is not safe — the enterprise module's AOT bean definitions keep the old signature and every endpoint 500s with `NoSuchMethodException` at request time while the container looks healthy. I hit exactly that before doing a full reactor build.

PR #4583, opened 2026-08-12
Add wiz tabular datasource endpoints

fredaan:
Adds `WizTabularController` under `/api/wiz` so the wiz portal can create and edit tabular datasources (REST, MongoDB, and every other non-JDBC type).

**This is the backend half of inetsoft-technology/stylebi-wiz#1556** — that PR's portal code calls these endpoints and cannot work without them. Deploy this first.

Completes the set: browsing and JDBC databases shipped in #4572; this is the third and last datasource shape.

## Why these pass the model through verbatim

The JDBC endpoints in #4572 deliberately reduce StyleBI's native model to a wiz-specific one, to absorb contracts a hand-built client would break on. These do the opposite, and the asymmetry is intentional.

A tabular form has no fixed shape. `TabularView` *is* the form description — a tree of editors, buttons and panels that the server generates from the datasource bean. There is nothing to reduce: anything trimmed here is a field the renderer can no longer draw. That also makes the round trip load-bearing, since the server recomputes the whole tree from the definition the client posts back, so the tree has to survive intact in both directions.

Worth knowing for anyone touching this: `TabularView` deserializes with `FAIL_ON_UNKNOWN_PROPERTIES` and carries no `@JsonIgnoreProperties`, so a client that attaches its own field to a node gets a 400. `TabularButton` is the lone exception that ignores unknowns, which is the only reason its client-only `loading` flag round-trips.

## Authorization

`updateDataSource` resolves its target from the body — `parentPath + "/" + name` — while the `@Secured` gate checks the `path` parameter. Left alone, a caller authorized for one folder could claim another in the body and write somewhere the gate never saw. Both `parentPath` and `oldName` are now overridden from the gated path. Renaming additionally requires DELETE on the old path, checked up front so a denial arrives as a 403 rather than as a save result the user would read as a validation failure.

## Two behaviours that exist to prevent data loss

**A null `additionalConnections` used to delete every additional connection**, because the save path treats the list as complete. Additional connections are out of scope for this editor, which makes the client that never sends them precisely the one that would destroy them. A null list now means "not edited" and carries the stored connections forward; only an explicit empty list deletes.

**Non-tabular paths are rejected with 422 rather than served.** `LayoutCreator` on a `JDBCDataSource` produces an empty form, so a read would return an empty definition and saving it would rewrite a database through a form that never showed its settings.

## `TabularUtil.setSessionId`

Set on every entry point that can reach a refresh. The hazard is not that it would be unset — refresh does not read it — but that it is a `ThreadLocal` that is never cleared, so a pooled worker still holds the session id of whoever it last served, and a plugin button keying off it would act under another user's identity. A synthetic per-principal id avoids that without minting a throwaway session per call, which `request.getSession()` would do for a bearer-token caller with no cookie jar.

## Testing

`WizTabularControllerSecurityTest`, 11 tests, modelled on the `WizDatabaseControllerSecurityTest` added in #4572: every mapping under `/api/wiz`, both `@Secured` gates including the `@PermissionPath` wiring, create denied without folder WRITE and the root fall-back not rescuing a named folder, create saving into the authorized folder rather than the one the body claims, update taking its parent from the gated path, rename denied without DELETE, refresh touching neither `SecurityEngine` nor `XRepository`, and an assertion that the class never grows a local `@ExceptionHandler` (which would take precedence over `WizControllerErrorHandler` and downgrade a 403 to a 400).

```
./mvnw -o -pl community/core surefire:test -Dtest=WizTabularControllerSecurityTest
[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0

./mvnw -o -pl community/core surefire:test -Dtest=WizDatabaseControllerSecurityTest
[INFO] Tests run: 9, Failures: 0, Errors: 0, Skipped: 0    (#4572, no regression)
```

The gate assertions were mutation-checked rather than trusted: removing `@PermissionPath` from `getTabularDefinition` turns the test red.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
@@ -1,0 +1,658 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package inetsoft.web.wiz.controller;
+
+import inetsoft.sree.security.ResourceAction;
+import inetsoft.sree.security.ResourceType;
+import inetsoft.sree.security.SecurityEngine;
+import inetsoft.uql.DataSourceListing;
+import inetsoft.uql.DataSourceListingService;
+import inetsoft.uql.XDataSource;
+import inetsoft.uql.XRepository;
+import inetsoft.uql.tabular.TabularDataSource;
+import inetsoft.uql.tabular.TabularUtil;
+import inetsoft.util.Catalog;
+import inetsoft.util.MessageException;
+import inetsoft.web.portal.data.DataSourceDefinition;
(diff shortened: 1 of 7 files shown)


fredaan:
duplicate with https://github.com/inetsoft-technology/stylebi/pull/4582

PR #4584, opened 2026-08-13
fix(wiz): reject a physical-table column the source does not have

fredaan:
## Problem

A `/ws/table` physical-table request's explicit `columns` list was turned into `AttributeRef`s verbatim, with **no reconciliation against the source table's real columns**. A name the table does not have entered the assembly's stored column selection unchallenged, and every read-back this API serves — `/ws/table`'s own response and `/ws/structure` — then advertised the phantom column as real.

Nothing failed later either. At query time `PreAssetQuery.validateColumnSelection` **removes** columns the source does not have and carries on, so the generated SQL stays valid and even the execution probe passes. The caller binds a chart to a column no query can ever produce, gets an empty chart with no error, and the column disappears from the worksheet as soon as Composer refreshes it.

Observed end to end: an agent asked for `ORDER_AMOUNT` on `Examples/Orders.ORDERS`, which has no such column (`ORDER_ID, CUSTOMER_ID, EMPLOYEE_ID, ORDER_DATE, DISCOUNT, PAID`). Creation returned `success: true` with `ORDER_AMOUNT` in the response, `/ws/structure` reported it, the chart bound `Average(ORDER_AMOUNT)` and rendered with no data, and the column was absent when the worksheet was opened in Composer.

## Fix

**`buildPhysicalTable` reconciles the requested columns against the source's metadata** — which it already fetches to resolve the qualified table name, so the authority was in hand and unused.

- *Forgiving where the intent is unambiguous*: a case-only difference (`order_id` → `ORDER_ID`) or a table-qualified name whose remainder is a real column (`ORDERS.ORDER_ID`, `SA.ORDERS.DISCOUNT`) canonicalizes to the source's spelling. Unqualifying happens only when the full name is not itself a column, so a source column whose own name contains a dot still wins.
- *Fail loud otherwise*: the error names both the unresolvable columns and the ones the table does have, so a caller can correct itself in one step. `addOneTable`'s existing rollback turns it into `success: false` + `errorMessage`.
- Columns the same request **derives** (`expressionColumns` / `windowColumns`) are exempt — the source's metadata cannot know them.
- A metadata gap (no columns reported) logs a warning and skips reconciliation rather than failing the create.

**The execution probe gained a dropped-column backstop** for the same end state reached by any other path (a stale mirror base, an expression over a vanished column). `probeExecutable` only answered "did the query run"; it now also compares what the table advertises against what its query actually produced.

- *Count-guarded on purpose*: name matching runs only once the counts prove columns went missing, so a mere spelling difference between a `ColumnRef`'s name and a lens header can never by itself fail a create. If the names cannot be matched, the error reports the counts instead.
- Matching tolerates the mirror naming difference (assembly says `BaseTable.col`, the lens reports it bare) in both directions, case-insensitively.

## Behaviour change

A physical-table create that passes explicit `columns` now always reads the source's column metadata (one extra `MetadataApiService.getMetaData` call; the provider caches). If that read fails, the create now fails where it previously succeeded — intentional, since building a table on unreconciled column names is exactly the failure this fixes.

`shouldProbe` is unchanged: pure physical / mirror / join tables with no derived columns are still not probed (zero added latency). The backstop covers the tables that are probed at all; the known cause is rejected up front by the reconciliation.

## Tests

Two new test classes, both following the existing pure-logic pattern (`WorksheetTableServiceShouldProbeTest`):

- `WorksheetTableServiceColumnValidationTest` (9) — hallucinated column fails loud, every unresolvable column reported together, exact names pass through, case canonicalization, table-prefix unqualifying, dotted source name wins, derived column exempt, metadata gap passes through, blank name reported.
- `WorksheetTableServiceProducedColumnsTest` (8) — a column the query did not produce is named, all missing named, nothing missing when all produced, mirror qualified-vs-bare match in both directions, case-insensitive match, a genuinely absent qualified attribute is still caught, null lens column neither matches nor crashes.

```
WorksheetTableService*Test (10 classes)   Tests run: 78,  Failures: 0
inetsoft.web.wiz (whole package)          Tests run: 992, Failures: 0
```

Not verified end to end against a live datasource: the reconciliation now blocks the phantom column at creation, so exercising the probe backstop needs a different silent-drop path plus a running server. Its matching logic is unit-tested.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
@@ -417,9 +417,10 @@ public class WorksheetTableService {
          " column(s) but its query produces only " + produced.size() + ": " +
          (dropped.isEmpty() ? String.join(", ", produced)
             : "the query cannot produce " + String.join(", ", dropped)) +
-         ". A column the source does not have is dropped from the query instead of failing it, so " +
-         "this table would render as an empty or partial result. Check the column names against " +
-         "the source (a derived value belongs in expressionColumns, not columns).");
+         ". A column the query cannot resolve is dropped from it instead of failing it, so this " +
+         "table would render as an empty or partial result. Check each column resolves against " +
+         "what this table is built on — the source table for a physical table, the SELECT list for " +
+         "a sql query table, the base table for a mirror or join.");
    }
 
    /** The columns this table promises its callers — exactly what {@link #extractColumnsFromSelection} reports. */
@@ -498,7 +499,8 @@ public class WorksheetTableService {
          return Set.of();
       }
 
-      String upper = name.toUpperCase();
+      // Locale.ROOT: these are datasource/lens identifiers, not display text.
+      String upper = name.toUpperCase(Locale.ROOT);
       int dot = upper.lastIndexOf('.');
 
       return dot > 0 && dot < upper.length() - 1
@@ -820,7 +822,9 @@ public class WorksheetTableService {
 
       if(request.getColumns() != null && !request.getColumns().isEmpty()) {
          // Explicit column list from the LLM — reconciled against the source's real columns so a
-         // name the table does not have fails loud here (see resolveRequestedColumns).
+         // name the table does not have fails loud here (see resolveRequestedColumns). A name this
+         // same request derives is dropped from the result: applyExpressionColumns /
+         // applyWindowColumns below contribute the real derived column.
          List<WorksheetTable.ColumnInfo> cols = resolveRequestedColumns(
             request.getColumns(), sourceColumnNames(metaData), derivedColumnNames(request));
          ColumnSelection cs = buildColumnSelection(cols);
@@ -1169,7 +1173,9 @@ public class WorksheetTableService {
     * The names (and aliases) of the columns this same request DERIVES rather than reads from the
     * source — {@code expressionColumns} and {@code windowColumns}. A request that also lists a
     * derived column in {@code columns} is naming something the source's metadata cannot know about,
-    * so it is exempt from source reconciliation.
+    * so {@link #resolveRequestedColumns} neither rejects it nor keeps it: the entry is dropped and
+    * the real derived column is contributed by {@link #applyExpressionColumns} /
+    * {@link #applyWindowColumns}.
     */
    private static Set<String> derivedColumnNames(WorksheetTable request) {
       Set<String> names = new HashSet<>();
@@ -1223,7 +1229,8 @@ public class WorksheetTableService {
     * @param sourceColumns   the source table's real column names; empty means "metadata unavailable",
     *                        which skips reconciliation rather than failing the create
     * @param derivedColumns  columns this request derives itself (see {@link #derivedColumnNames})
-    * @return {@code requested}, with every name canonicalized to the source's spelling
+    * @return the source-backed subset of {@code requested}, each name canonicalized to the source's
+    *         spelling — a name only this request derives is DROPPED, not passed through
     * @throws IllegalArgumentException if any name cannot be resolved to a source column
     */
    // Package-private for unit testing (WorksheetTableServiceColumnValidationTest).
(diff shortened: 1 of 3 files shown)


fredaan (commented):
## Code review

The core of the change is right: the source's column metadata was already being fetched to resolve the qualified table name, and reconciling the caller's `columns` list against it is the correct place to stop a phantom column — up front, where the authority is in hand, rather than downstream where `validateColumnSelection` has already silently swallowed it. The probe backstop is a sensible second line, and its count-guard is in the safe direction (it can only fire when the query produces *fewer* columns than the assembly advertises, so a mere spelling difference between a `ColumnRef` name and a lens header cannot by itself fail a create).

One blocking defect and three smaller issues were found. All four are fixed in `2842c49`, pushed to this branch.

---

### 1. Blocking — the derived-column exemption was defeated downstream, turning a supported request into a failed create

`resolveRequestedColumns` exempted a `columns` entry that this same request derives, and **passed it through** to `buildColumnSelection`, which turns any name into a plain `AttributeRef`. That attribute then shadows the real derived column:

- `applyExpressionColumns` calls `cs.addAttribute(colRef)` — the one-arg overload, i.e. `exclusive = true` (`ColumnSelection:99`).
- `ColumnRef` does not override `equals(Object)`; it inherits `AbstractDataRef.equals`, which compares `getName0()` only — `entity + "." + attribute`, with no regard for the backing ref's type (`AbstractDataRef:303`, `:140`).
- So `ColumnRef(ExpressionRef "discount_status").equals(ColumnRef(AttributeRef "discount_status"))` is `true`, the `attrs.contains(attribute)` guard rejects the add, and **the expression is silently discarded**.

End state: the assembly keeps a phantom `AttributeRef` the source does not have. `PreAssetQuery.validateColumnSelection` drops it from the query, and — because a physical table carrying `expressionColumns` is always probed — the new `checkProducedColumns` then **fails the create**, with a message telling the caller that "a derived value belongs in expressionColumns, not columns" when the caller had already put it there.

This is exactly the request shape the exemption was written to permit, and it is the shape the new test asserted:

```java
// columns: [ORDER_ID, discount_status], expressionColumns: [discount_status]
assertEquals(List.of("ORDER_ID", "discount_status"), names(...));   // green in isolation
```

The unit test passes because it only exercises `resolveRequestedColumns`; the path it green-lights fails end to end. Pre-PR this input produced a silent phantom column, so the PR does not *introduce* the shadowing — but it converts it into a hard, misdirected create failure, which makes it blocking for this change.

**Fix applied:** drop the entry instead of passing it through — the real derived column is contributed by `applyExpressionColumns` / `applyWindowColumns` right after. The derived branch is now reached **only when the source has no such column**, so a derived name that collides with a real source column still resolves to the source column exactly as before; only the broken case changes. Tests updated accordingly, plus three new cases (derived name dropped; derived name that is also a real source column; a hallucinated column still reported when the request also derives columns).

### 2. A derived name was matched case-sensitively while source reconciliation is case-insensitive

`derivedColumns.contains(name)` was an exact-string lookup, so `columns: [{"name": "Discount_Status"}]` against `expressionColumns: [{"name": "discount_status"}]` missed the exemption and was reported as `Column(s) not found in source table: Discount_Status` — misdirecting, given every other name comparison in this method is case-insensitive. Now matched through a `CASE_INSENSITIVE_ORDER` set.

### 3. Locale-sensitive `toUpperCase()` on datasource identifiers

`canonical`, the qualified-name fallback and `matchKeys` all used the no-arg `String.toUpperCase()`. Under a Turkish/Azeri default locale that folds `i` to `İ`, so case-insensitive matching breaks for **any** column name containing an `i` — both the reconciliation (a valid create rejected) and the probe matcher. Now `Locale.ROOT`, which is what these are: identifiers, not display text.

### 4. `checkProducedColumns`' advice misdirects on the paths it mostly guards

The message ended with "a derived value belongs in expressionColumns, not columns", which is physical-table-specific — but the reconciliation already rejects that cause up front, so in practice this backstop fires for `sql query table` (always probed), mirrors and joins, where the advice is nonsense. Retargeted to name what each table type actually resolves against.

---

### Verified, no change needed

- **`AssetUtil.trimEntity(name, null)`** strips only an `entity:` prefix, never a dot, so a source column whose own name contains a dot survives `buildColumnSelection` intact — the `aSourceColumnWhoseOwnNameContainsADotWinsOverUnqualifying` case holds end to end, not just in the unit test.
- **`checkProducedColumns` false-positive risk on aggregated tables.** `applyAggregateInfo` ends with `setColumnSelection(privateCs, false)`, which regenerates the public selection from the group/aggregate refs, so advertised and produced stay in step. Invisible columns are excluded from `advertised` and (if produced at all) only push `produced` higher, i.e. toward the early return. Duplicate names in a SQL result are deduped by `ColumnSelection`, which again only makes `produced` the larger side.
- **Both new test classes** match the existing `WorksheetTableServiceShouldProbeTest` convention exactly (same annotations, same null-dependency `service()` helper).
- **The behaviour change is correctly scoped and documented.** A physical-table create with explicit `columns` now always reads source metadata, and a metadata read failure now fails a create that previously succeeded. That is the right trade — building on unreconciled names is the failure being fixed — and `shouldProbe` is untouched, so no table gains probe latency.
- No security impact: no new external input reaches SQL, and the error message discloses only the column names of a table the caller already passed a datasource READ check for (`buildTable`).
- No backward incompatibility in the API contract: the response shape is unchanged, and the only new failures are on requests that previously produced a silently broken worksheet.

### Residual, non-blocking

- **Metadata-gap path.** When the source reports no columns, `resolveRequestedColumns` still returns `requested` verbatim, so a derived name listed in `columns` can still shadow its expression there. Left as-is deliberately: with no metadata there is no way to tell a derived name from a real column, and this is an explicitly degraded "accept as declared" path. Worth a follow-up only if the gap turns out to be reachable in practice.
- **The gap warning carries no identity** — `"No source metadata to reconcile the requested columns against"` names neither the table nor the datasource, so it is hard to act on in a shared log.

### Tests

```
inetsoft.web.wiz (whole package)   Tests run: 995, Failures: 0, Errors: 0, Skipped: 0
```

(9 → 12 in `WorksheetTableServiceColumnValidationTest`.)

Not approving here only because GitHub does not allow approving one's own pull request — from the review's side there are no remaining blocking issues, no merge conflicts, and CI is re-running on `2842c49`.

PR #4585, opened 2026-08-13
Bug #75975, dispatch a real "loaded" DOM event from every Embed*Component

Zxm-happy:
## Summary
- Companion fix to wiz's bug-75975 (portal PR #1577): every `<inetsoft-chart>`/`<inetsoft-crosstab>`/`<inetsoft-table>`/`<inetsoft-gauge>`/`<inetsoft-text>`/`<inetsoft-image>` custom element gave a plain custom-element consumer (e.g. wiz's chat card) no DOM event for "I have actually finished rendering" — only a fixed client-side timeout to fall back on. That made every embed mount/refresh look exactly as slow as the consumer's timeout, regardless of how fast the underlying render actually was.
- Adds a genuine per-type "rendered" signal and dispatches it as `CustomEvent("loaded", {bubbles: true, composed: true})` on each `Embed*Component`'s own host element, reusing the same manual-dispatch pattern already used for `"wizfullscreen"`:
  - **Chart**: `VSChart` gets a new `onChartLoaded` `@Output`, emitted from the existing `clearChartLoading()` — the same "tiles have settled" signal that already drives this component's own `chartLoading`/`chartLoadingIcon` state (retries, no-data, and its own internal timeout fallback are already handled there).
  - **Gauge/Image**: `AbstractImageComponent`'s existing `loaded(success)` callback (already wired to the underlying ``'s `load`/`error`) now also emits a new `onLoad` `@Output`.
  - **Crosstab/Table/Text**: no async per-tile loading of their own — content is plain DOM rendered synchronously from the model by Angular change detection — so the embed wrapper dispatches `"loaded"` right after applying an Add/RefreshVSObjectCommand's `detectChanges()`.
- Deliberately does **not** reuse `AbstractVSObject`'s existing `onLoadingStateChanged`: tracing `AssemblyLoadingCommand`/`ClearAssemblyLoadingCommand` into `CoreLifecycleService` shows that pair only brackets server-side model recomputation, well before the browser even requests a chart's tile images — wiring it through would trade "signals too late" for "signals too early".

## Test plan
- [x] New/extended unit tests: `vs-chart.component.spec.ts` (`onChartLoaded` fires on the loading→not-loading transition, not on a no-op call), `vs-gauge.component.spec.ts` (`onLoad` emits the load outcome), `embed-table.component.spec.ts` (host dispatch + wiring through `processAddVSObjectCommand`/`processRefreshVSObjectCommand`, including the "wrong assembly name" no-op case).
- [x] `ng test portal` (full suite): 195 files / 1018 tests passing.
- [x] `ng run portal:test-tl` (large-test suite): 403 files / 8165 tests, 3 pre-existing failures unrelated to any file touched here (composer-selection-container-children, datasources-datasource-editor — Angular TestBed module-compile ordering / a timeout, reproducible on this branch's base commit too).
- [ ] Manual verification against a real embed page once this lands, alongside the wiz-side consumer switching over to this event.

### DIFF
--- a/web/projects/portal/src/app/embed/chart/embed-chart.component.html
+++ b/web/projects/portal/src/app/embed/chart/embed-chart.component.html
@@ -28,7 +28,8 @@
       [appSize]="appSize"
       [container]="viewerRoot"
       [variableValues]="variableValuesFunction"
-      (contextmenu)="onOpenContextMenu($event)">
+      (contextmenu)="onOpenContextMenu($event)"
+      (onChartLoaded)="dispatchLoaded()">
     </vs-chart>
   }
   @if (showError) {
--- a/web/projects/portal/src/app/embed/chart/embed-chart.component.ts
+++ b/web/projects/portal/src/app/embed/chart/embed-chart.component.ts
@@ -280,6 +280,17 @@ export class EmbedChartComponent extends CommandProcessor implements OnInit, OnD
       );
    }
 
+   /**
+    * Dispatches a plain DOM CustomEvent on this component's own host element once the wrapped
+    * <vs-chart> reports it has actually finished rendering (see VSChart.onChartLoaded) - so a
+    * consumer embedding <inetsoft-chart> as a plain custom element (no Angular APIs available)
+    * can learn when the chart is done instead of guessing from unrelated DOM signals or a fixed
+    * timeout. See toggleWizFullscreen() above for the same host-dispatch pattern.
+    */
+   dispatchLoaded(): void {
+      this.elementRef.nativeElement.dispatchEvent(new CustomEvent("loaded", { bubbles: true, composed: true }));
+   }
+
    ngOnDestroy() {
       this.dialogService.ngOnDestroy();
       this.clearServerUpdateInterval();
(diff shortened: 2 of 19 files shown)


Zxm-happy:
Fixed the correctness bug flagged in review: **\`VSImage\` bypassed \`AbstractImageComponent.loaded()\`**, so \`<inetsoft-image>\` never actually dispatched \`"loaded"\` — the exact gap the review called out (this PR's summary claims image is fixed, but wasn't).

- \`onImageLoad()\`/\`finishLoad()\` in \`vs-image.component.ts\` managed \`this.loading\` directly instead of going through the base class. \`finishLoad(success = true)\` now emits \`this.onLoad.emit(success)\` itself; the \`\`'s \`(error)\` binding now passes \`finishLoad(false)\` explicitly. All pre-existing no-arg call sites (the tiled-image immediate-completion paths in \`ngOnInit\`/\`modelChanged\`, and \`onImageLoad()\`'s own success path) keep their exact prior behavior via the default parameter.
- Added \`onLoad\` coverage to \`vs-image.component.display.tl.spec.ts\`'s existing "Group 6 — finishLoad" block (both the default-success and explicit-failure cases).
- Also added the \`embed-crosstab.component.spec.ts\` / \`embed-text.component.spec.ts\` symmetry coverage the review noted as a non-blocking gap (mirrors \`embed-table.component.spec.ts\`'s existing \`dispatchLoaded\`/wrong-assembly tests).

Verified: \`ng test portal\` (full suite) 197 files / 1026 tests passing; \`ng run portal:test-tl --include="**/vs-image.component.*.spec.ts"\` 52/52 passing.

fredaan (approved):

PR #4587, opened 2026-08-13
Stop a null property value reading as "no such property"

jshobe-inetsoft:
`get_property` returned `currentValue: null` and `recognized: false` both for a real property nobody has set and for a name that does not exist. The two responses were byte-identical, and the only evidence a bare `SreeEnv` property exists is a read site in Java source, which the server cannot consult.

## How it surfaced

An agent configuring OIDC SSO from scratch, running with only the admin-chat plugin's tools, probed several `openid.*` names. Every one was unset, so every probe returned null. It concluded StyleBI's SSO settings *"are stored in a separate structured settings object, not in the flat sree.properties-style store admin-chat manages"* and abandoned the task, emitting hand-entry instructions instead. They are ordinary `SreeEnv` properties.

Its caution was correct — it refused to guess names into `preview_changes`, since a wrong key writes an inert property while reporting success — but the tool gave it nothing to confirm a name against.

## The change

`PropertyView` gains `exists`, plus `guidance` when it is unknown:

| | |
|---|---|
| `confirmed` | catalogued, or holds a value — proof by demonstration, and covers anything declared in `defaults.properties` since the defaults chain resolves through `getProperty` |
| `unknown` | uncatalogued **and** unset — genuinely indistinguishable from a name that does not exist, so the response says so rather than letting the caller infer |

It also stops `isSecret` manufacturing evidence. That branch is reached on a name match alone, so an invented name ending `.key` came back *"Value withheld: secret properties are not exposed through admin-chat"* — asserting a secret existed to withhold, out of a string suffix. It now states the rule without claiming a value is there.

The secret path still never reads `SreeEnv` at all. An earlier draft determined existence by reading the value and discarding it, which broke the test asserting exactly that; existence is a signal the catalog supplies properly, and buying it by reading the password-encryption master key is the wrong trade. A secret therefore reports `unknown` unless catalogued, which is honest.

## Verification

Tests 9 → 14, including the `openid.client.id` / `openid.completely.made.up.key` pair that produced the wrong conclusion. Full `admin.ai` suite (129 tests) green.

Verified live against a running server: `openid.client.id` now returns `exists: "unknown"` with guidance, and re-running the same agent task produced a correct seven-property plan instead of abandoning it.

Paired with inetsoft-technology/stylebi-wiz#PENDING, which teaches the model how to read the field.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminPropertiesController.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminPropertiesController.java
@@ -105,21 +105,35 @@ public class AdminPropertiesController {
    private PropertyView view(AdminPropertyName name, CatalogEntry entry) {
       AdminRiskClassifier.RiskClassification risk = classifier.classify(name);
       String description = entry == null ? null : entry.description();
-      String current;
+      boolean secret = AdminPropertyCatalog.isSecret(name.baseName());
+      // A secret's value is not read AT ALL, not merely omitted from the response - see the test
+      // that asserts SreeEnv is never called for one. Determining existence by reading it would
+      // trade that invariant for a signal the catalog supplies properly, so the secret path stays
+      // blind and reports what it actually knows.
+      //
+      // orgScope=false so the value shown is the one an apply would actually change.
+      String stored = secret ? null : SreeEnv.getProperty(name.key(), false, false);
 
       // Secret properties (e.g. password.encryption.key) are still LISTED - an operator
       // legitimately needs to know the property exists - but the value is withheld rather than
       // forwarded to the model provider that this endpoint's caller relays responses through. Not
       // a 403: the same operator role can already read this unmasked via PropertiesController, so
       // refusing to even acknowledge the property's existence would just be confusing.
-      if(AdminPropertyCatalog.isSecret(name.baseName())) {
-         current = null;
-         description = "Value withheld: secret properties are not exposed through admin-chat.";
+      //
+      // The old phrasing said "Value withheld", which asserts there is a value to withhold. This
+      // branch is reached on nothing more than a name match, so an invented name ending .key came
+      // back claiming a secret had been withheld - manufacturing evidence that the property exists
+      // out of a string suffix.
+      if(secret) {
+         description = "Secret property: admin-chat can neither read nor change it. Its value is "
+            + "not read here at all, so this says nothing about whether one is set.";
       }
-      else {
-         // orgScope=false so the value shown is the one an apply would actually change.
-         current = SreeEnv.getProperty(name.key(), false, false);
-      }
+
+      // A catalogued name is authoritative; a stored value is proof by demonstration, and covers
+      // anything declared in defaults.properties, since the defaults chain resolves through
+      // getProperty. Neither means the server cannot say whether the name is real - which is
+      // always the case for a secret, since this path deliberately never looks.
+      boolean confirmed = entry != null || stored != null;
 
       return new PropertyView(name.key(),
          entry == null ? List.of() : (entry.aliases() == null ? List.of() : entry.aliases()),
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft:
Addressed the nit in e691b4aad — it's the same defect the PR exists to fix, so it belonged here rather than in a follow-up.

Removing `"Value withheld"` stopped the response asserting a *value* was there, but the replacement still opened `"Secret property: ..."`, which asserts the *category* from a string-suffix match. As you say, a caller who stops at the colon reads that as confirmation the property exists — the very inference `exists` was added to prevent. Fixing one half and leaving the other was inconsistent.

New wording:

> This name matches admin-chat's secret pattern, so its value is neither read nor changed here - not even read to test whether one is present, so this says nothing about whether the property exists or is set.

It holds for a real secret too: `password.encryption.key` genuinely does match the pattern, and the sentence is no weaker for it.

The test now pins both halves — the description must not open by asserting the category, and must refer to the name. 14 tests still green.

jkidacki-inetsoft (approved):

jshobe-inetsoft (commented):
## Code Review: PR #4587 — Stop a null property value reading as "no such property"

### Summary
This PR adds an `exists` (`confirmed`/`unknown`) field plus conditional `guidance` text to `PropertyView`, so the admin-chat property-discovery endpoint can distinguish "real property, currently unset" from "name doesn't exist" — two cases that previously both serialized as `currentValue: null`. It also fixes the secret-property description so it no longer asserts a value exists ("Value withheld...") purely from a name-suffix match.

### Correctness & completeness
- The `confirmed` logic (`entry != null || stored != null`) correctly covers both catalogued-but-unset and uncatalogued-but-has-a-stored-value (e.g. `defaults.properties`-backed) cases.
- Secret-pattern names are deliberately never read via `SreeEnv`, preserving the existing invariant (still covered by `sreeEnv.verify(..., never())`), and now correctly report `unknown` unless catalogued — avoiding the old bug where an invented name ending in `.key` was reported as an existing, withheld secret.
- Verified there are no other call sites of the `PropertyView` record constructor in the codebase, so widening it from 11 to 13 components doesn't break any other compilation unit.
- The new JSON fields are purely additive on a REST response (`/api/wiz/v1/admin/properties[/…]`) — backward compatible for any consumer that ignores unknown fields.

### Tests
- Five new unit tests directly pin the regression (`openid.client.id`), the invented-secret-name case, the catalogued-but-unset case, the uncatalogued-with-value case, and that a real-unset vs. invented-unset property are indistinguishable except by identical guidance text. This is good, targeted coverage of exactly the ambiguity the PR is fixing.
- Existing tests (secret withholding, alias resolution, org-scoped lookup, permission checks) are unaffected by the field reordering/addition.

### Minor / non-blocking notes
- The guidance text references `search_product_docs` (an external tool name) — that contract lives outside this repo (per the PR description, paired with a `stylebi-wiz` change), so it can't be verified from this codebase alone. Worth double-checking the tool name/corpus argument matches what the companion change actually registers.
- The in-method comments are quite long (multi-paragraph) relative to typical Java style, but this matches the pre-existing narrative comment style already used elsewhere in this file and in `AdminPropertyCatalog`, so it's consistent rather than a new deviation.

### Security
No concerns. The change strictly narrows what can be inferred from a response (removing a case where the server implicitly asserted a secret's existence) and does not add any new read of secret values — the secret path is explicitly verified to never call `SreeEnv.getProperty`.

None found. No other production code references `PropertyView`'s constructor, and the API change is additive only.

### Verdict
No blocking issues found. All CI checks (build, Semgrep, Semgrep OSS, claude-review, CLA) are green, the PR has already been approved (by @jkidacki-inetsoft) and merged, and no merge conflicts are outstanding. **This would be an approval** — posted as a comment rather than a formal review approval because GitHub does not allow approving one's own pull request (this account is the PR author).

PR #4588, opened 2026-08-13
Let admin-chat set the OIDC client secret

jshobe-inetsoft:
Follows inetsoft-technology/stylebi#4587. Configuring OIDC SSO end to end died one property short: everything else previewed cleanly and `openid.client.secret` had to be typed into Enterprise Manager by hand.

The refusal was justified as an egress control, but egress is about values *leaving* the host — by the time a write is proposed the value is already in the caller's context, so refusing it protected nothing and only stopped the task completing.

## Not a relaxed predicate

`isSecret` matches sixteen properties and they do not behave alike:

| | |
|---|---|
| **encrypting accessors** | `openid.client.secret`, `stylebi.google.openid.client.secret` — `setClientSecret` writes `encryptPassword(...)`, the getter reverses it with `Tool.decryptPassword` |
| **plain accessors** | `log.fluentd.security.password`, `google.maps.key`, `sso.rsa.public.key`, `auth0.client.secret` — all read with a bare `getProperty` |
| **not a secret** | `enable.changepassword` — matches on the substring "password", holds a boolean |
| **key material** | `password.encryption.key`, `password.hash.key`, `jwt.signing.key`, `sso.rsa.private.key`, `license.*` |

A blanket `if(isSecret) setPassword` would have handed ciphertext to four properties that expect a literal — silently, while reporting success. So membership is an **allow-list verified against each accessor**, currently two entries, with the same discipline `AdminPropertyCatalog`'s javadoc already requires of a catalog entry and for the same reason: the failure is silent.

## Three interactions that each needed handling

- **Verify reads back through `getPassword`.** Comparing the stored ciphertext against the plaintext `desired` would mark every successful credential write `FAILED` and roll it straight back.
- **Rollback must not re-encrypt.** Its value came from a `getProperty` read and is already ciphertext; `ACTION_ROLLBACK` separates it from an apply. Double-encrypting would restore something that never decrypts to the original secret.
- **`beforeValue`/`afterValue` stay the stored form.** This is load-bearing rather than an oversight: `AdminChangesetApplyService` replays `getBeforeValue()` as the rollback value, so masking or decrypting either would break rollback. Ciphertext is not the secret, so nothing sensitive is recorded by keeping it.

## Scope of what is allowed

Cloud secrets are refused outright — there these properties hold the *name* of a secret, which is why Enterprise Manager swaps in a **Secret ID** field, and a literal value would store something nothing downstream can resolve.

**Reading stays refused.** The egress rationale is about values leaving the host and is untouched by letting one in. The plan shows `(set)`/`(not set)` rather than shipping ciphertext to a model provider; the cost is that swapping one secret for a different one is not seen by the drift gate, which is tolerable because what the human approved is "set this property to the value I supplied", and that executes either way.

## Verification

Plan tests 21 → 25, change-service tests 15 → 19. Full `admin.ai` suite green (147 tests) on the post-#4587 base.

Tests cover each direction that would corrupt data if reversed: an allow-listed credential goes through `setPassword`, verification uses `getPassword`, a rollback writes verbatim without re-encrypting, and a secret-named property that is *not* allow-listed is written verbatim.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminChangePlanService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminChangePlanService.java
@@ -18,6 +18,7 @@
 package inetsoft.web.admin.ai;
 
 import inetsoft.sree.SreeEnv;
+import inetsoft.util.Tool;
 import inetsoft.util.audit.AdminChangeRecord;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
@@ -74,14 +75,35 @@ public class AdminChangePlanService {
                "changes: duplicate entry for " + name.key() + "; list each property once");
          }
 
+         boolean credential = AdminPropertyCatalog.isEncryptedCredential(name.baseName());
+
          // Unlike the read path in AdminPropertiesController (which withholds the value but still
          // shows the property exists), a change is refused outright: this service exists to WRITE
          // a value, and blanking a secret through it would make every stored encrypted credential
          // undecryptable. See AdminPropertyCatalog.isSecret for why this is an egress/blast-radius
          // control rather than a privilege boundary.
-         if(AdminPropertyCatalog.isSecret(name.baseName())) {
+         //
+         // The exception is the small allow-list of application credentials whose accessors
+         // encrypt at rest: those are written through SreeEnv.setPassword, which encrypts exactly
+         // as the Enterprise Manager field does. Reading them is still refused - the egress
+         // rationale is about values LEAVING the host, and is untouched by letting one in.
+         if(AdminPropertyCatalog.isSecret(name.baseName()) && !credential) {
             throw new IllegalArgumentException(
                name.key() + ": secret properties cannot be changed through admin-chat");
+         }
+
+         // With cloud secrets configured, these properties hold the NAME of a secret rather than a
+         // secret - getPassword resolves it through Tool.loadCredentials and reads a client_secret
+         // field out of the JSON. Enterprise Manager swaps its Client Secret field for a Secret ID
+         // one in that mode for exactly this reason. An agent handed a literal secret would store
+         // it where a reference belongs, and nothing downstream could resolve it, so refuse rather
+         // than write a value that cannot work.
+         if(credential && Tool.isCloudSecrets()) {
+            throw new IllegalArgumentException(
+               name.key() + ": this deployment uses cloud secrets, so this property holds the ID "
+               + "of a secret rather than the secret itself. Set it from Enterprise Manager's "
+               + "Settings > Security > SSO page, whose Secret ID field writes the reference "
+               + "correctly.");
          }
 
          CatalogEntry entry = catalog.getEntry(name);
@@ -94,7 +116,16 @@ public class AdminChangePlanService {
          requireHashSafe("property", name.key());
          requireHashSafe("value", proposed);
 
-         String currentValue = SreeEnv.getProperty(name.key(), false, false);
+         // A credential's stored form is ciphertext, and the plan is relayed to a model provider by
+         // the caller. Ciphertext is not the secret, but there is no reason to ship it either, and
+         // an operator reading the plan is served better by whether one is already set than by a
+         // base64 blob. Note the consequence for the drift gate: replacing one secret with a
+         // different one is not detected between preview and apply, only set <-> unset is. That is
+         // tolerable here because the human approved "set this property to the value I supplied",
+         // which is what executes either way.
+         String currentValue = credential
+            ? (SreeEnv.getProperty(name.key(), false, false) == null ? "(not set)" : "(set)")
+            : SreeEnv.getProperty(name.key(), false, false);
          requireHashSafe("currentValue", currentValue);
 
          changes.add(new PlanChange(name.key(), name.orgId(),
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminChangePlanService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminChangePlanService.java
@@ -106,6 +106,21 @@ public class AdminChangePlanService {
                + "correctly.");
          }
 
+         // Every other property in this service takes "" as an explicit set-to-empty, with null
+         // meaning reset-to-default. A credential cannot honour that: SreeEnv.setPassword guards
+         // on Tool.isEmptyString and returns without writing, so "" would leave the existing
+         // secret in place. That does fail safe - the read-back verify compares the old secret
+         // against "" and reports FAILED - but an operator trying to blank a secret would get a
+         // bare failure and no idea why, having asked for something this path cannot do. Refuse it
+         // here, where the message can say what to use instead. An empty client secret is not a
+         // valid credential in any case, so nothing legitimate is being turned away.
+         if(credential && "".equals(requested.getValue())) {
+            throw new IllegalArgumentException(
+               name.key() + ": an empty value cannot be written to a credential - the property "
+               + "would keep its current secret. Pass a null value to reset the property to its "
+               + "default, or a non-empty value to change it.");
+         }
+
          CatalogEntry entry = catalog.getEntry(name);
          // An uncatalogued property cannot be validated or canonicalized, so its value passes
          // through verbatim; the classifier marks it high risk so review flags it.
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminChangePlanService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminChangePlanService.java
@@ -123,6 +123,21 @@ public class AdminChangePlanService {
          // different one is not detected between preview and apply, only set <-> unset is. That is
          // tolerable here because the human approved "set this property to the value I supplied",
          // which is what executes either way.
+         //
+         // proposedValue is deliberately NOT masked, and the asymmetry with currentValue is the
+         // point rather than an oversight. Masking a value only helps if withholding it keeps it
+         // from somewhere it would otherwise reach. currentValue qualifies: it is read off the
+         // server here and would reach the caller for the first time. proposedValue does not - it
+         // arrived IN this request, and apply requires the caller to send the identical changes
+         // array back, because the request body is the plan and the hash is recomputed from it. So
+         // the caller necessarily holds the plaintext before and after this response, and blanking
+         // it in between would remove nothing while destroying the operator's ability to see WHICH
+         // secret a plan writes, which is what the review gate exists to show them.
+         //
+         // That reasoning depends on the value always arriving through the request. If an
+         // out-of-band channel is ever added - a placeholder the server resolves from somewhere
+         // the caller never sees - then echoing the resolved value here WOULD be a new disclosure,
+         // and this must be revisited along with it.
          String currentValue = credential
             ? (SreeEnv.getProperty(name.key(), false, false) == null ? "(not set)" : "(set)")
             : SreeEnv.getProperty(name.key(), false, false);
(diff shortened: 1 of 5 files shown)


jshobe-inetsoft (commented):
## Review of #4588 — Let admin-chat set the OIDC client secret

Thanks for the very thorough PR description — the write-up of the four properties/behaviors that `isSecret` conflates, and the apply/verify/rollback interactions, made this easy to follow. The new tests are well targeted (allow-list write, current-value masking, refusal of non-allow-listed secrets, cloud-secrets refusal, rollback without re-encryption, verbatim write for a secret-named-but-not-allow-listed property) and clearly derived from the same reasoning as the code.

I have one design concern I think should be resolved before merge, one documentation defect, and one minor robustness note. I'm also requesting changes because a required check is currently failing.

### 1. `proposedValue` for an allow-listed credential is not masked, even though it is asserted to cross the same "egress to a model provider" boundary that motivates masking `currentValue` (needs resolution or explicit rationale)

`AdminChangePlanService` now does this for a credential (lines ~119-126 post-patch):

```java
String currentValue = credential
   ? (SreeEnv.getProperty(name.key(), false, false) == null ? "(not set)" : "(set)")
   : SreeEnv.getProperty(name.key(), false, false);
```

The stated reason, in the PR's own comment, is: *"A credential's stored form is ciphertext, and the plan is relayed to a model provider by the caller... there is no reason to ship it either."* `AdminPropertyCatalog.isSecret`'s javadoc is even more explicit: *"the caller here is an LLM that forwards responses to a model provider off-host, so a secret value read through this path leaves the host in a way the ordinary EM properties page never does."*

That rationale is about the **response** object (`ResolvedPlan`/`PlanChange`) leaving the host toward a model provider — and `proposedValue` (the new secret the human/broker wants written) rides in that exact same response, completely unmasked:

```java
CatalogEntry entry = catalog.getEntry(name);
...
changes.add(new PlanChange(name.key(), name.orgId(),
   currentValue, proposed, ...));   // proposed is the plaintext secret, verbatim
```

and is pinned by the new test itself:

```java
PlanRequest req = request("set the OIDC client secret", "openid.client.secret", "s3cret");
ResolvedPlan plan = service.resolve(req);
assertEquals("s3cret", plan.changes().get(0).proposedValue());   // plaintext, in the plan
```

So the same `ResolvedPlan` that masks the *existing* secret specifically to keep it off the wire to the model provider ships the *new* secret to that same model provider in the clear. If the reasoning for masking `currentValue` holds, it seems to apply equally to `proposedValue` for the same field.

I can see a counter-argument: whoever/whatever constructed the `PlanRequest` already had to know the plaintext value to put it in the request in the first place, so the model may already have "seen" it (e.g., if a human typed it directly into a chat message the model read). But that's an assumption about the broker/UI on the other side of this API that isn't visible from this module, and it doesn't hold if the value is meant to be supplied through some out-of-band/masked entry path specifically so the model *never* sees it (the same intent behind masking `currentValue`). Given the deliberate care taken elsewhere in this PR, I'd like to see this either:
- fixed (don't echo the plaintext `proposedValue` back for an allow-listed credential — e.g. return `"(set)"`/omit it, while still applying the real value from the original request at apply time), or
- explicitly justified in the same comment block (why the write side is safe to leave unmasked when the read side isn't), so a future reader isn't left with the asymmetry I hit here.

### 2. Javadoc on `isEncryptedCredential` cites a class that doesn't exist

```java
* {@code OpenIDConfig.setClientSecret} and
* {@code StyleBIGoogleOpenIDConfig.setClientSecret} each write
* {@code encryptPassword(...)} and their getters reverse it with {@code Tool.decryptPassword}.
```

I searched the codebase and there is no `StyleBIGoogleOpenIDConfig` class (or anything similarly named) anywhere. `styleBI.google.openid.client.secret` has no dedicated Java setter/getter at all — its only special handling is the case-sensitive literal check inside `SreeEnv.getPassword`'s cloud-secrets branch (`SreeEnv.java:225-230`), and it isn't in the admin-chat property catalog either (`admin-property-catalog.json` has 4 entries, none OIDC-related), so it's an uncatalogued property as far as this feature is concerned.

This matters because the whole point of this javadoc paragraph is "every entry must be verified against its accessor... the same discipline this class's javadoc requires... the failure is silent" — and the specific verification claim written down for the Google entry doesn't correspond to anything that exists. The actual write/read path for that property (`SreeEnv.setPassword`/`getPassword`, non-cloud branch: `Tool.encryptPassword`/`Tool.decryptPassword`) does look internally consistent (I traced it), but the doc should say that, not point at a class that isn't there. Please correct the javadoc to reference the real code path.

### 3. `Tool.isCloudSecrets()` is only checked at preview/resolve time, not at apply time

`AdminChangePlanService.resolve()` refuses an allow-listed credential when `Tool.isCloudSecrets()` is true, which is correct and well-reasoned. But `AdminChangeService.applyChange()` doesn't re-check it — it only computes `encryptOnWrite` from `isEncryptedCredential(...)` and the action type. If cloud secrets were enabled on the deployment in the window between an approved preview and its apply (an unusual operational sequence, but the kind of thing this PR's own comments are otherwise careful to rule out rather than assume away), `SreeEnv.setPassword` skips encryption entirely in cloud mode and would write the plaintext value verbatim into a property everything else treats as a secret-manager reference — silently "succeeding" exactly the way the class-level javadoc warns against. Low likelihood, but a defensive re-check at apply time (mirroring the preview-time refusal) would close the gap and match the discipline the rest of this change applies elsewhere.

### CI / mergeability

- The `claude-review` check is currently **failing** ("Claude encountered an error after 6m 11s" — looks like the automated reviewer errored out rather than completed; worth re-running).
- `build`, `Semgrep Scan`, `Semgrep OSS`, and `license/cla` are passing.
- No merge conflicts (`mergeable: MERGEABLE`), but GitHub reports `mergeStateStatus: UNSTABLE` due to the failing check above.

Given the failing required check and the open item in §1, I'm requesting changes rather than approving. Everything else — the allow-list design, the encrypt-on-apply/decrypt-on-verify handling, the rollback-doesn't-re-encrypt distinction, and keeping `beforeValue`/`afterValue` in stored form for rollback correctness — looks correct and carefully reasoned, and the test coverage for those paths is solid.

jshobe-inetsoft:
All three addressed in `13c8362e4`.

## §3 — cloud-secrets re-check at apply time: fixed, you were right

The preview-time refusal is not the check that protects the write. A plan is approved at preview and executed later, so `AdminChangeService` now re-checks and fails rather than writing a literal secret into a reference field.

Rollback is deliberately still permitted: it writes the stored form verbatim and never encrypts, so the hazard doesn't apply to it, and blocking it would strand a half-applied changeset. Two tests pin both halves.

## §2 — the javadoc: the class is real, my citation wasn't usable

`StyleBIGoogleOpenIDConfig` exists at `enterprise/src/main/java/inetsoft/enterprise/sso/StyleBIGoogleOpenIDConfig.java` — in the **enterprise superproject**, which is not this repository. Its `setClientSecret` does write `PasswordEncryption.newInstance().encryptPassword(...)` and `getClientSecret` reverses it, exactly as claimed.

But you searched this repo and found nothing, which is the correct thing to do and the correct conclusion from what's visible here — and in a paragraph whose entire subject is *"every entry must be verified against its accessor... the failure is silent"*, a citation a reader can't follow is a real defect. The javadoc now names the module and the path, notes that the only trace visible from this repo is the case-sensitive literal in `SreeEnv.getPassword`'s cloud-secrets branch (which you found), and states why an enterprise-only property is still correct to list here: on a Community build nothing reads it, so nothing writes it.

Your independent trace of the `SreeEnv.setPassword`/`getPassword` path was the right verification given what you could see.

## §1 — `proposedValue`: justified rather than masked

I went with your second option, because masking would be theatre here.

Masking helps only if withholding a value keeps it from somewhere it would otherwise reach. `currentValue` qualifies — it is read off the server and would reach the caller for the first time. `proposedValue` does not: it **arrived in this request**, and `apply` requires the identical `changes` array back, because the request body *is* the plan and the hash is recomputed from it. The caller necessarily holds the plaintext both before and after this response, so blanking it in between removes nothing — while destroying the operator's ability to see *which* secret a plan writes, which is what the review gate exists to show them.

Your counter-argument anticipated this, and you were right that it rests on an assumption not visible from this module. So the comment now states the assumption explicitly and flags what invalidates it: if an out-of-band value channel is ever added — a placeholder the server resolves from somewhere the caller never sees — then echoing the resolved value here becomes a genuine disclosure and this must be revisited with it.

## CI

Full `admin.ai` suite green locally, 149 tests (change-service 19 → 21). The failing `claude-review` check is the sandboxed reviewer hitting its turn limit rather than a finding — this review was run separately for that reason.

jshobe-inetsoft:
Both follow-up findings handled — one fixed in `bc4fa1e56`, one is deliberate and worth stating explicitly.

## Empty-string value for a credential — fixed

Your trace was right: `setPassword` guards on `Tool.isEmptyString` and returns without writing, so `""` leaves the existing secret in place, and the read-back verify then reports `FAILED`. Fail-safe, but an operator trying to blank a secret gets a bare failure with no indication why.

It's now refused at plan time, where the message can name the property and point at `null`:

> `openid.client.secret`: an empty value cannot be written to a credential — the property would keep its current secret. Pass a null value to reset the property to its default, or a non-empty value to change it.

An empty client secret isn't a valid credential anyway, so nothing legitimate is turned away. Reset is unaffected and remains the way to clear one — a null value goes through `SreeEnv.remove`, not `setPassword`. Two tests pin both. Plan tests 25 → 27.

## Catalog entries — deliberate, deferred

Not an oversight. Adding the two credentials to `admin-property-catalog.json` would set `risk` and `snapshotScope` for them, and `AdminPropertyCatalog`'s own javadoc is explicit that `snapshotScope` must be determined by tracing side effects through `PropertyChangeSideEffects`, `PropertiesEngine.applyProperty` and registered listeners — *"not by guessing from the property's apparent importance"*. Adding two entries ad hoc, in a PR about the write path, is exactly the shortcut that warning exists to prevent.

There's a larger piece of work queued to populate the catalog properly from the per-property reference documentation, starting with the ~55 SSO-area properties. These two belong in that pass, where `risk` and `snapshotScope` get determined systematically rather than one at a time.

So the conservative default you observed — `risk=high`, `snapshotScope=storage`, full Tier-2 backup and agent sign-off on every credential change — is the intended behaviour until then, and is the correct default for an uncatalogued property.

Full `admin.ai` suite green locally: 151 tests, no failures.

jshobe-inetsoft:
Fixed in `3d51e3811` — and it's a better catch than "speculative against dead code" suggests.

## `ACTION_RESTORE`

You're right about the consequence, but the more useful framing is that the *spelling* was the defect. `encryptOnWrite` excluded `ACTION_ROLLBACK`, so every other action encrypted — a deny-list. This PR's whole thesis is that deny-lists fail silently, which is why `isEncryptedCredential` is an allow-list verified per accessor. I then wrote a deny-list for actions in the same change.

It now names the one action that carries a plaintext value:

```java
boolean encryptOnWrite = AdminPropertyCatalog.isEncryptedCredential(name.baseName())
   && AdminChangeRecord.ACTION_APPLY.equals(req.getAction());
```

That fails closed for `RESTORE` and for any action added later, rather than only for the one I happened to think of. Test `doesNotReEncryptACredentialOnRestoreEither` pins it. Change-service tests 21 → 22.

Also — thank you for tracing the encryption round-trip end to end. Confirming that `SreeEnv.setPassword`/`getPassword` resolve to the *same* `PasswordEncryption` instance as `OpenIDConfig`'s own accessor under `LOCAL` secrets is the claim the whole change rests on, and I had verified the two accessors in isolation without checking they agree on the instance. That's the failure this PR could have shipped with and no test would have caught.

## Catalog entries — deliberate, with the new detail acknowledged

Still a deliberate follow-up, for the reason given earlier: `snapshotScope` has to be determined by tracing side-effect channels, and `AdminPropertyCatalog`'s javadoc explicitly forbids inferring it from a property's apparent importance. Two ad-hoc entries in a write-path PR is that shortcut.

Your new observation — that no catalog entry means `canonicalizeValue` never runs, so a credential value isn't trimmed — is correct and worth recording. It is not introduced here, though: that is true of every uncatalogued property today, and `AdminChangeService` deliberately does not trim (its comment explains why — the rollback path replays stored values that never went through canonicalization). Whether a client secret *should* be trimmed is a real question, and the right place to answer it is the catalog entry, where trimming is the documented behaviour rather than an ad-hoc call in the write path.

Noted for that pass alongside `risk` and `snapshotScope`.

jkidacki-inetsoft (approved):

PR #4590, opened 2026-08-13
Classify encrypted credentials by their writer

jshobe-inetsoft:
Follows inetsoft-technology/stylebi#4588. The allow-list added there held two entries, and the javadoc justified the omissions with the wrong evidence.

## The reasoning error

The javadoc said `log.fluentd.security.password` is *"read with a plain `SreeEnv.getProperty`"*, so encrypting it would corrupt it. The exclusion was right; the reasoning was not. `LogSettingService` reads it with `getProperty` and calls `Tool.decryptPassword` on the result two lines later — so **by the reader it looks encrypted**.

The reader cannot settle this. `Tool.decryptPassword` passes an unrecognised value straight through — `LocalPasswordEncryption`'s final branch is literally `// clear text` — so a defensive decrypt is compatible with *both* storage forms, typically because something wrote the value encrypted at some point in the past. **Only the writer is authoritative.** `log.fluentd.security.password` is written with a bare `setProperty`, so it stays out, now for the correct reason.

## What applying that rule found

Five properties missing, four of them in the dangerous direction:

| property | writer |
|---|---|
| `mail.smtp.pass` | `EmailSettingsService` → `SreeEnv.setPassword` |
| `mail.smtp.clientsecret` | `EmailSettingsService` → `SreeEnv.setPassword` |
| `mail.smtp.accesstoken` | `EmailSettingsService` → `SreeEnv.setPassword` |
| `mail.smtp.refreshtoken` | `EmailSettingsService` → `SreeEnv.setPassword` |
| `log.fluentd.security.sharedkey` | `LogSettingService.toPassword` → `Tool.encryptPassword` |

All but `mail.smtp.clientsecret` are invisible to `isSecret`, so admin-chat wrote them with a plain `setProperty` while Enterprise Manager writes them encrypted — **silently downgrading a stored credential to plaintext at rest**, and reporting success, because the same defensive decrypt reads plaintext back fine.

That is the SMTP password and the OAuth access and refresh tokens, on the *"configure the mail server"* path.

## The traps, now pinned by tests

The two properties that sit adjacent to these and do **not** belong are where the next edit will go wrong:

- `mail.smtp.tokenuri` — same save block as the four above, written plain. Encrypting it would store ciphertext where a URL is expected.
- `log.fluentd.security.password` — sibling of `sharedkey`, written plain despite the near-identical name.

## Scope

This fixes only **which accessor a write uses**. The read half of the same underlying defect — `isSecret` returning those same mail properties *unmasked* to the caller — is already filed as **#76006**, which also documents why the name predicate must be kept and unioned with the writer signal rather than replaced: replacing it would unmask fifteen properties in order to close three.

Catalog tests 15 → 18, change-service 22 → 24, full `admin.ai` suite green.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminChangePlanService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminChangePlanService.java
@@ -100,10 +100,7 @@ public class AdminChangePlanService {
          // than write a value that cannot work.
          if(credential && Tool.isCloudSecrets()) {
             throw new IllegalArgumentException(
-               name.key() + ": this deployment uses cloud secrets, so this property holds the ID "
-               + "of a secret rather than the secret itself. Set it from Enterprise Manager's "
-               + "Settings > Security > SSO page, whose Secret ID field writes the reference "
-               + "correctly.");
+               AdminPropertyCatalog.cloudSecretsRefusal(name.key()));
          }
 
          // Every other property in this service takes "" as an explicit set-to-empty, with null
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminChangeService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminChangeService.java
@@ -102,10 +102,7 @@ public class AdminChangeService {
          // otherwise refuses to leave open. Re-check at the point of the write.
          if(encryptOnWrite && desired != null && Tool.isCloudSecrets()) {
             throw new IllegalStateException(
-               name.key() + ": this deployment uses cloud secrets, so this property holds the ID "
-               + "of a secret rather than the secret itself. Set it from Enterprise Manager's "
-               + "Settings > Security > SSO page, whose Secret ID field writes the reference "
-               + "correctly.");
+               AdminPropertyCatalog.cloudSecretsRefusal(name.key()));
          }
 
          if(desired == null) {
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminPropertyCatalog.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminPropertyCatalog.java
@@ -290,6 +290,26 @@ public class AdminPropertyCatalog {
       return baseName != null && ENCRYPTED_CREDENTIALS.contains(baseName.toLowerCase());
    }
 
+   /**
+    * The refusal both the plan and the apply path give when an encrypted credential is set on a
+    * deployment using cloud secrets.
+    *
+    * <p>Built here rather than written out at each site so the two cannot drift apart, and
+    * deliberately does NOT name an Enterprise Manager page. It used to say "Settings > Security >
+    * SSO", which was true while the allow-list held only the two OpenID credentials and became
+    * wrong the moment mail and logging credentials joined them - they are configured on entirely
+    * different pages. Sending an operator to the wrong page is worse than sending them to none,
+    * and a page-per-property mapping maintained here would be a second name-keyed table to drift
+    * out of date, which is the failure this class has already had once.
+    */
+   public static String cloudSecretsRefusal(String key) {
+      return key + ": this deployment uses cloud secrets, so this property holds the ID of a "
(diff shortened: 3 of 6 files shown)


jshobe-inetsoft:
Addressed in `907904dcf` rather than deferred, on your own framing: the string is pre-existing, but **this PR is what makes it reachable for the five properties it is wrong about.** While the allow-list held only the two OpenID credentials the message was accurate; adding mail and fluentd credentials is what invalidated it. Shipping that knowingly isn't a follow-up.

## What changed

The message no longer names a page:

> `mail.smtp.pass`: this deployment uses cloud secrets, so this property holds the ID of a secret rather than the secret itself. Set it from the Enterprise Manager page that owns this setting - under cloud secrets those pages present a Secret ID field, which writes the reference correctly. admin-chat cannot, because it would store the literal value where nothing downstream can resolve it.

I considered branching on the property to name the right page (the corpus has a verified `emBinding.page` for each of these — `settings/general#email` for `mail.smtp.pass`, `settings/security/sso` for `openid.client.secret`). I decided against it: that would be a **second name-keyed table maintained by hand next to the first**, which is the exact failure mode this class has already had once. If per-page guidance is wanted later it should be driven from the catalog data, not a copy of it.

A wrong page is worse than no page, and this refusal exists purely to tell the operator what to do instead.

## Also de-duplicated

The message was written out verbatim at both the plan site and the apply site. That is how the SSO wording survived the change that invalidated it — nobody edits two copies. It now lives in `AdminPropertyCatalog.cloudSecretsRefusal`.

## Tests

Plan tests 27 → 28. The new one asserts the message names the property, says why, and does **not** name an SSO page — checked across a mail, a logging and an SSO credential, so it fails if the wording regresses for any family.

Full `admin.ai` suite green locally.

Thanks also for tracing every writer claim independently rather than taking the PR description's word — the `log.fluentd.security.sharedKey` vs `.password` split is the one a reader is most likely to get backwards, and you checked both.

jshobe-inetsoft:
Both fixed in `286fc2953`. You were right, and I verified it rather than taking it on trust:

- `EmailSettingsService` has a full cloud-secrets path for `mail.smtp.pass` only — `smtpSecretId` read/write plus a `secretIdVisible` flag driving the UI.
- `mail.smtp.clientSecret`, `accessToken`, `refreshToken` are written with `setPassword` at lines 88-90 with **no** `isCloudSecrets` branch.
- `LogSettingService` contains **zero** references to `Tool.isCloudSecrets`.

So four of seven were being pointed at a field that does not exist.

## The pattern, which is the actual lesson

Both rounds are the same mistake at different grain: first the page name, then the field. **Any claim about how a property is configured elsewhere is a claim per property, and this method has one string for seven of them.** A third wording describing the UI would have failed the same way — I'd have been fixing the symptom for the second time.

So the message now describes only what admin-chat will not do:

> `mail.smtp.pass`: this deployment uses cloud secrets, so a value stored under this property is resolved as a reference to a secret rather than used directly. admin-chat will not set it - it can only write the literal value it was given, which nothing downstream could resolve. Configure it through Enterprise Manager instead.

That holds for all seven, and keeps holding as the list grows — which is the property the previous two wordings lacked.

The test now rejects `"Secret ID"` as well as `"Security > SSO"`, across a mail, a logging and an SSO credential, so the next wording cannot quietly reintroduce a UI claim.

If per-property guidance is genuinely wanted later, the corpus already records a verified `emBinding.page` for each of these. That should drive it — not a second name-keyed table maintained by hand beside `ENCRYPTED_CREDENTIALS`.

## Nitpick

`cloudSecretsRefusal` is package-private now. Both callers are in `inetsoft.web.admin.ai`.

Suite green locally, plan tests 28.

Worth saying: checking the Angular templates for the mail and logging pages is what made this finding possible, and it is not something the Java diff would have revealed. That is the second time this review has caught an accuracy problem in prose I wrote rather than in the logic.

jkidacki-inetsoft (approved):

PR #4592, opened 2026-08-13
Fix portal form binding regressions

swaker854:
Summary
Dashboard name field (edit-dashboard-dialog) was bound via both [(ngModel)] and [formControl] simultaneously; removed the redundant ngModel and read the value from nameControl on save so edits are actually persisted.
Several schedule action fields (secretId, password, confirmPassword, attachmentName, delivery message, from-email) mixed ngModel/ngModelChange with formControlName on the same input, causing dirty-state and value tracking to conflict. Replaced with valueChanges subscriptions that sync the corresponding model fields, removing the duplicate template bindings.
task-action-pane's dashboard-select input used ngModel purely for display; switched to a plain [value] binding, and added a clearViewsheet() handler so the "Clear" button updates the reactive form control (previously it only mutated the model directly, leaving the form out of sync).
Fixed the arrange-dashboard "move up" button using the wrong icon class (downward-icon instead of upward-icon).
Refined the scheduler health pill: it now distinguishes an explicit "Unavailable" status from a general warning (detailMessage present but not unavailable) instead of collapsing both into is-unavailable.
Added missing NgClass/NgIf imports to schedule-task-list and start-time-editor standalone component imports arrays (required after switching those templates to use directives not previously declared).
Why
These editors were using mixed template-driven (ngModel) and reactive-form (formControl/formControlName) bindings on the same inputs. Angular allows this without error, but the two mechanisms fight over the control's value/dirty state, so user edits could silently fail to persist or dirty tracking would be wrong (e.g., save button staying disabled, or saving stale values).

Test plan
 Edit a dashboard's name in the Edit Dashboard dialog and save — confirm the new name persists.
 In a schedule task's email/zip action, edit Secret ID, Password, Confirm Password, Attachment Name, and delivery message — save and reopen, confirm all values persist.
 Select and then Clear a dashboard in the Select Dashboard action pane — confirm the field clears and the save button reflects the dirty state correctly.
 In Arrange Dashboards, confirm the "move up" button shows an upward-pointing icon.
 With the scheduler in a degraded (non-fatal) state vs. fully unavailable, confirm the health pill shows "warning" vs. "unavailable" styling correctly.

### DIFF
--- a/web/projects/portal/src/app/portal/dialog/analyze-mv/analyze-mv-view/analyze-mv-pane.component.html
+++ b/web/projects/portal/src/app/portal/dialog/analyze-mv/analyze-mv-view/analyze-mv-pane.component.html
@@ -22,7 +22,7 @@
       <thead class="thead-light">
         <tr>
           <th class="table-header text-center align-middle">
-            <input type="checkbox"
+            <input class="form-check-input" type="checkbox"
               [ngModel]="isModelAllSelected()" (ngModelChange)="selectionAll($event)">
           </th>
           @for (header of tableHeaders; track header.name) {
@@ -36,7 +36,7 @@
       <tbody>
         @for (model of models; track model.name; let i = $index) {
           <tr>
-            <td><input type="checkbox" [ngModel]="isModelSelected(model)" (ngModelChange)="selectionChanged($event, model)"></td>
+            <td><input class="form-check-input" type="checkbox" [ngModel]="isModelSelected(model)" (ngModelChange)="selectionChanged($event, model)"></td>
             @for (header of tableHeaders; track header.name) {
               <td>{{model[header.name]}}</td>
             }
--- a/web/projects/portal/src/app/portal/dialog/arrange-dashboard-dialog.component.html
+++ b/web/projects/portal/src/app/portal/dialog/arrange-dashboard-dialog.component.html
@@ -57,7 +57,7 @@
                 <td>
                   <button type="button" class="btn btn-light-no-bg icon-hover-bg" (click)="moveDashboard(i, i - 1)"
                     [disabled]="!canMoveUp(i)">
-                    <i class="downward-icon" aria-hidden="true"></i>
+                    <i class="upward-icon" aria-hidden="true"></i>
                   </button>
                   <button type="button" class="btn btn-light-no-bg icon-hover-bg" (click)="moveDashboard(i, i + 1)"
                     [disabled]="!canMoveDown(i)">
--- a/web/projects/portal/src/app/portal/dialog/edit-dashboard-dialog.component.html
+++ b/web/projects/portal/src/app/portal/dialog/edit-dashboard-dialog.component.html
@@ -27,7 +27,7 @@
           @if (nameControl) {
             <div class="shell-form-group shell-form-row one-col-form-flex">
               <div class="form-floating col">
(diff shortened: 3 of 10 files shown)


jshobe-inetsoft (approved):
### Code Review: PR #4592 — Fix portal form binding regressions

**Scope:** Removes conflicting `ngModel`/`formControl(Name)` dual-bindings across several dashboard/schedule editors, fixes a swapped icon class, and refines the scheduler health pill's status distinction. Also fixes two standalone-component `imports` arrays that were missing `NgClass`/`NgIf` despite the templates using those directives (a pre-existing build hazard on `epic-74519`).

#### Verified against the code

- **`edit-dashboard-dialog`**: `[(ngModel)]="name"` removed, leaving only `[formControl]="nameControl"`. `okClicked()` now reads `nameControl.value` (falling back to `this.name` only if `nameControl` isn't yet constructed, matching the template's `@if (nameControl)` guard). `this.name` is otherwise only used to seed the initial control value — no stale/dead references left.
- **`action-accordion`**: `secretId`/`password`/`confirmPassword`/`attachmentName`/`deliveryMessage`/`from` inputs had both `formControlName` and a competing `ngModel`/`ngModelChange` binding on the same element (the latter effectively dead, since Angular's `NgModel` selector excludes elements with `formControlName`, so those change handlers — including `updateFromEmails()` — likely never fired). Replaced with `valueChanges` subscriptions added in `initForm()`, which already calls `unsubscribeForm()` first — same pattern as the pre-existing `cc`/`bcc`/`to` wiring, so no duplicate-subscription risk across repeated `initForm()` calls.
- **`task-action-pane`**: the hidden `dashboard` input previously had both `formControlName="dashboard"` and `[ngModel]` set from the same expression — redundant/conflicting. New `clearViewsheet()` now routes the Clear button through `changeViewsheet(null)`, which calls `setValue`/`markAsDirty` on the `dashboard` control, so the reactive form's dirty/error state now stays in sync with the "Clear" action (previously it only mutated the model directly). No `valueChanges` listener on `dashboard` exists, so no feedback-loop risk from the new `setValue` call.
- **`arrange-dashboard-dialog`**: icon class swap (`downward-icon` → `upward-icon` on the "move up" button) matches the described bug exactly.
- **`schedule-task-list`**: `schedulerHealthPillClass` now special-cases `statusLabel === "_#(Unavailable)" || "Unavailable"` before falling back to the previous `detailMessage ? warning : ready` logic. Confirmed `PortalSchedulerHealthModel.statusLabel` and the `"_#(Unavailable)"` fallback value are exactly what the client sets on an HTTP error, and the `is-warning`/`is-unavailable`/`is-ready` SCSS classes already exist on `epic-74519`. Minor non-blocking note: matching a raw localization key (`_#(Unavailable)`) as a magic string is a little fragile if the resource bundle key ever changes — not something this PR needs to fix, just worth knowing.
- **`NgClass`/`NgIf` import additions**: confirmed both templates (`schedule-task-list.component.html` line 23 `[ngClass]`, `start-time-editor.component.html` line 46 `*ngIf`) actually use these directives, so the added imports are required, not speculative.

#### Assessment

- No security concerns (pure UI/form-wiring changes, no new input handling).
- No backward-incompatibility — internal component wiring only.
- No test regressions — no test files touch this code today, so nothing is broken; also nothing new was added, which is a reasonable gap for a bindings-only bugfix.
- No blocking issues found.

All checks (build, Semgrep ×2, CLA, claude-review) pass, and there are no merge conflicts (`mergeable: MERGEABLE`). Approving.

PR #4596, opened 2026-08-14
Viewsheet agent backend, and a capturing dispatcher for headless composer calls

larryliang-inetsoft:
Backend for the viewsheet layout plugin: a new `inetsoft/web/wiz/viewsheet/` package exposing `/api/wiz/v1/agent/viewsheet/**`, plus a dispatcher fix that closes a silent-failure defect.

**Base branch:** `feature-76074-server-fixes`, not `main` — the `wiz/pairing` package and `ScriptImageService` this builds on do not exist on main yet.

## The defect this fixes first

Composer services **do not throw on failure.** They dispatch a `MessageCommand` of `Type.ERROR` and return normally:

```java
// VSObjectPropertyService.editObjectProperty
this.coreLifecycleService.sendMessage(
   Catalog.getCatalog().getString("common.dependencyCycle"),
   MessageCommand.Type.ERROR, commandDispatcher);
return;                                    // caller sees a normal return
```

`sendMessage` routes to `dispatcher.sendCommand`, and `CommandDispatcher.withDummyDispatcher` overrides that to a no-op. So for any browser-less caller — the seven existing `withDummyDispatcher` sites included — every one of the **56 `MessageCommand.Type.ERROR` sites across `composer/` and `binding/`** becomes a silent success.

`CapturingCommandDispatcher` captures dispatched commands instead of discarding them, and `withCapturingDispatcher` throws `CommandErrorException` when any ERROR was reported. Warnings surface separately and do not fail the call. It transmits nothing, matching the dummy dispatcher's intent — the pairing layer broadcasts to the browser separately.

## The plugin backend

| Class | Responsibility |
|---|---|
| `ViewsheetAgentController` | REST surface; validation and delegation only |
| `ViewsheetSessionService` | Pairing session → `RuntimeViewsheet`; capturing dispatcher, checkpoint, broadcast |
| `ViewsheetReadService` | Agent-facing layout model |
| `ViewsheetEditService` | 16 structural ops onto Composer services |
| `ViewsheetFormatService` | Format via `FormatPainterService` |

Every mutation delegates to the Composer's own services rather than reimplementing them, so their validation and fixups apply. Each tool call is exactly one `addCheckpoint`, matching a single Composer action, so the user's Ctrl+Z steps back one edit at a time.

## Worth reviewer attention

- **`align` and `distribute` have no Composer service.** They are position arithmetic here, issuing one `MultiMoveVsObjectEvent` so an alignment stays a single checkpoint rather than N.
- **`applySocketSession` is load-bearing.** Without carrying the browser's socket id onto the runtime, `SheetAgentBroadcastService` finds no session and silently skips — the user's Composer would never reflect an agent edit, with no error anywhere.
- **`CapturingCommandDispatcher` passes a null cluster deliberately.** It is hand-constructed rather than Spring-managed, so its `@PostConstruct` never runs; the class already guards for null; and nothing is transmitted. Taking `Cluster.getInstance()` would make it unusable outside a live Spring context for no benefit.

## Testing

1,014 tests pass across `inetsoft.web.wiz.**`, of which 28 are new.

`HeadlessDialogServiceProbeTest` is included `@Disabled`, documenting why the real Composer bean graph cannot be booted in the current test harness — component-scanning `inetsoft.web` drags in the `@Configuration` classes the harness exists to replace. It belongs as this controller's first integration test, and should verify: a real property write mutates `VSAssemblyInfo`; a rename does not break browser tracking; `@ClusterProxy` routing works bean-to-bean; broadcast plus checkpoint leave undo coherent.

---

## Update — the first live run against a Composer, and what it found

The disabled probe test above asks four questions no unit test can answer. They have now been
answered by running the thing, and the run found a family of defects that all reported success.

**`@ClusterProxy` routing works.** The first live failure — `edit move` returning `ok` and changing
nothing — looked exactly like broken routing, and the test plan says that signature means routing.
It does not. `moveObjects` (plural) reached the real service across the proxy and ran; it is the
anchored-line **fix-up hook**, not the move. The actual move is `moveObject`, which the real
controller calls first. Same shape sank `resize` (repositioned to 0,0), `resize_title` (**collapsed
the assembly to 1×1**), `align`, `distribute` and `set_format`: the plugin fabricated Composer
events and populated only the fields each op was "about", while the services read fields it left
null or zero.

**A real property write does mutate `VSAssemblyInfo`**, verified by image and by reload.

**Broadcast plus checkpoint leave undo coherent**: two tool calls needed exactly two Ctrl+Z presses
and restored the original geometry precisely. And the new read-only entry point adds none — measured,
three reads left the checkpoint total unchanged.

### The change that made the rest diagnosable

`WizControllerErrorHandler` had no `IllegalArgumentException` handler, so the ~75 validation sites
across `inetsoft/web/wiz/**` all surfaced as bare 500s with their messages discarded. Correct
validation was indistinguishable from a server fault. `PairingException` — the one type that *did*
have a handler — was the only source of readable errors in the whole run, which is what pointed at
the cause.

### New capabilities

`set_table_source` exists because an assembly added in the Composer had no way to be pointed at
data. The four deferred roadmap items are also in: a read-only capturing dispatcher (so
`getCellScript`/`getNamedGroup`, which answer by *dispatching* a command, can be read without an
undo step), the ten single-field chart shelves (a real candlestick renders), chart axis/legend/title
properties addressed by region, and Immutables support in `PropertyPath` — bare accessors plus
`withX` rebuilding from the leaf up, which is what finally makes image and VS-level properties
reachable.

### Worth flagging for review

Three tests were found asserting the wrong collaborator on a mock and thereby certifying the bugs
above — `moveDelegatesTheNewPosition…`, `addressesTheWholeAssemblyWhenNoRegionIsGiven`,
`addressesATitleLink`. One of them was mine, and its fix was dead code because the controller builds
the region directly and never calls the factory I normalized. The lesson is in the commits: a mock
test proves you called what you meant to call, not that it does anything.

Gates: **1380** Java wiz tests, 0 failures.

## Second pass: 20 defects found by driving the tools against a live Composer

Everything above was written before any of it had run outside a unit test. Running the plan
against a live server found the following, each fixed with a regression test. Several were found
only by *rendering the result* — the model said the right thing in most of them.

**Silent success — the tool reported a change it did not make**

| Defect | Cause |
|---|---|
| `get_chart_aesthetics` reported `field: null` on every channel, always | Read `AestheticInfo.getFullName()`, which the read path never sets. The name was on `dataInfo`. |
| `set_title` on a stale name after a rename returned ok | `requireExisting` was wired into 3 of 10 ops; the composer layer returns normally for an unknown assembly |
| Binding a column the source does not have | Accepted, stored, rendered nothing — reads as "empty data", not "no such column" |
| `set_column_labels` reported "Relabelled 1 column(s)" | Wrote `name2Labels`, a field nothing in the product reads. Now **501**, since a real rename needs a `TableDataPath` cell override |
| Sorting a column bound twice sorted the wrong one | A crosstab binds the same date column twice **by design** (Year › Quarter); name-only addressing took the first match. Now `index`, with an ambiguity refusal |
| An axis target naming no axis | Read returned a full plausible property list; write threw `NPE: axisArea is null` |

**Poisoned state — the bad call succeeded and a later innocent one failed**

`dateLevel: "year"` was stored verbatim. The *next* write to that assembly, on any shelf, failed
with `For input string: "year"`, naming neither the field nor the call that caused it. The levels
are `XConstants` numbers running in an order nobody expects — year 5, quarter 4, month 3, day 1 —
so `DateLevels` now takes the names, takes the numbers, and refuses anything else.

**Pairing could not write at all**

Every write through a paired session failed `InvalidUserException` while every read succeeded, so a
session looked healthy right up to the first mutation. The wiz layer resolves a paired sheet
without the owner check, then hands `runtimeId` to a composer service that re-resolves it and
enforces `Principal` equality — which the JWT-rebuilt agent principal can never satisfy.
`WorksheetEngine` already had `supportLogin` for exactly this shape; this adds a sibling
`pairedAgent`, granted only after confirming the runtime's owner is the same logical user — a check
that path previously did not do **at all**.

**Reported honestly rather than fixed**

- `resize_chart_plot` is **still inert** after four real fixes (contract wrong in all three tiers,
  plus a copy-paste bug where the height branch set the *width* ratio percent). Four hypotheses are
  ruled out in the findings doc; the next step is instrumentation, not a fifth guess.
- `percentageBy` sets the direction only. Rendering percentages needs a percent calculation on the
  aggregate, which no tool can set — surfaced as a gap.

**Eight tests were found asserting the wrong thing**, in addition to the three already noted above.
Three separate suites — Java, plugin, and wiz-services — each asserted `resize_chart_plot` refused
`ratio: 1.5`, so all three stayed green while the tool could only ever be handed values that do
nothing. `set_column_labels`' tests asserted on the dead field. The recurring shape: a test that
writes with our writer and reads with our reader proves only that the two agree.

**A CSS vocabulary over a model that stores integers**

Three fields documented one vocabulary and required another, each failing with a raw parse error
naming no field: `dateLevel: "year"` (an `XConstants` number where year is 5 and day is 1),
`align: "center"` (an `AlignmentInfo` object), and `borderTopStyle: "solid"` — which is the clearest
bug of the three, because `FormatInfoModel.getBorderStyle` *reads back* `"solid"` while the write
path does `Integer.parseInt` on the same field. The API hands you a value it cannot accept. All
three now coerce the documented word and refuse an unknown one by name.

If a fourth turns up, the right answer is probably one translation layer rather than a fourth
special case.

**An error handler that was dead code in production**

A body Jackson could not read returned a bodyless 400 — `"Request failed with status code 400"` and
nothing else, indistinguishable from a broken tool. The handler added for it did nothing, because
`GlobalExceptionHandler extends ResponseEntityExceptionHandler`, which carries a built-in handler
for the Spring MVC exceptions, and with neither advice ordered that one won. The unit test could
not see it: `standaloneSetup` registers only the advice under test, so the handler was correct in
isolation and unreachable in the running application. Caught by calling the tool again after the
"fix" and getting the identical empty 400. The wiz advice is now `HIGHEST_PRECEDENCE`, with the
reasoning in the class comment so the annotation is not later removed as noise.

Gates: **1427** Java wiz tests, 0 failures.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/WorksheetEngine.java
+++ b/core/src/main/java/inetsoft/report/composition/WorksheetEngine.java
@@ -549,6 +549,25 @@ public abstract class WorksheetEngine extends SheetLibraryEngine implements Work
     * @param touch <code>true</code> to update the access time and heartbeat.
     * @return the runtime sheet if any.
     */
+   /**
+    * Whether a principal may act on a runtime sheet owned by a different session.
+    *
+    * <p>Two cases, both narrow and both opt-in on the principal itself:
+    * {@code supportLogin} for a support-staff login, and {@code pairedAgent} for an AI agent that
+    * has paired with the runtime. The pairing path sets the latter only after confirming the
+    * runtime's owner is the same logical user (see {@code SheetRuntimeAccess}), so this widens
+    * access to another <em>session</em> of that user, never to another user.
+    *
+    * <p>The agent case is needed because the wiz layer resolves a paired sheet without the owner
+    * check, but then delegates to composer services that re-resolve through this method with the
+    * agent's own principal — which can never be {@code equals} to the browser session's.
+    */
+   private static boolean mayActForOwner(Principal user) {
+      return user instanceof SRPrincipal p &&
+         (Boolean.TRUE.toString().equals(p.getProperty("supportLogin")) ||
+          Boolean.TRUE.toString().equals(p.getProperty("pairedAgent")));
+   }
+
    public final RuntimeSheet getSheet(String id, Principal user, boolean touch) {
       if(!isLocal(id)) {
          LOG.error("Getting sheet from non-owner: {}", id, new Exception("Stack trace"));
@@ -563,9 +582,7 @@ public abstract class WorksheetEngine extends SheetLibraryEngine implements Work
       }
 
       if(rs.getUser() != null && user != null && !rs.matches(user)) {
-         if(!(user instanceof SRPrincipal &&
-            Boolean.TRUE.toString().equals(((SRPrincipal) user).getProperty("supportLogin"))))
-         {
+         if(!mayActForOwner(user)) {
             throw new InvalidUserException(
                catalog.getString("common.invalidUser", user, rs.getUser()),
                LogLevel.INFO, false, rs.getUser());
@@ -789,10 +806,7 @@ public abstract class WorksheetEngine extends SheetLibraryEngine implements Work
          boolean isSchedulerSheet = rsheet.getEntry() != null &&
             "true".equals(rsheet.getEntry().getProperty("_scheduler_"));
 
-         if(!isSchedulerSheet &&
-            !(user instanceof SRPrincipal &&
-            Boolean.TRUE.toString().equals(((SRPrincipal) user).getProperty("supportLogin"))))
-         {
+         if(!isSchedulerSheet && !mayActForOwner(user)) {
             throw new InvalidUserException(Catalog.getCatalog().
                getString("common.invalidUser", user, rsheet.getUser()),
                                            rsheet.getUser());
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/chart/VSChartPlotResizeService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/chart/VSChartPlotResizeService.java
@@ -73,7 +73,12 @@ public class VSChartPlotResizeService extends VSChartControllerService<VSChartPl
 
             if(heightResized || square) {
                chartInfo.setUnitHeightRatio(sizeRatio);
-               chartInfo.setUnitWidthRatioPercent(sizeRatio / chartInfo.getInitialHeightRatio());
+               // Height branch sets the HEIGHT percent. It set the width percent, overwriting the
+               // value the width branch had just computed on a square resize, and leaving the
+               // height percent unset — which matters because VGraphPair recomputes
+               // unitHeightRatio only when getUnitHeightRatioPercent() >= 1, so a height resize
+               // was dropped on the next recompute.
+               chartInfo.setUnitHeightRatioPercent(sizeRatio / chartInfo.getInitialHeightRatio());
                chartInfo.setHeightResized(true);
             }
          }
--- a/core/src/main/java/inetsoft/web/wiz/WizControllerErrorHandler.java
+++ b/core/src/main/java/inetsoft/web/wiz/WizControllerErrorHandler.java
@@ -19,6 +19,7 @@ package inetsoft.web.wiz;
 
 import inetsoft.util.Catalog;
 import inetsoft.util.InvalidUserException;
+import inetsoft.web.wiz.dispatch.CommandErrorException;
 import inetsoft.web.wiz.service.UnsupportedDatasourceException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -201,6 +202,29 @@ public class WizControllerErrorHandler {
                   "The request body could not be read: " +
                   (newline < 0 ? detail : detail.substring(0, newline)));
       return new ResponseEntity<>(payload, null, HttpStatus.BAD_REQUEST);
+   }
+
+   /**
+    * Maps a captured composer error to 409, carrying every message.
+    *
+    * <p>{@code CommandErrorException} extends {@code Exception} and had no handler, so the
+    * mechanism this whole feature exists to add — turning a composer ERROR that is only
+    * <em>dispatched</em> into something the caller can see — captured the text and then lost it to
+    * a generic 500 reading "Internal Server Error". The capture worked; the delivery did not.
+    *
+    * <p>409 rather than 400 because the request was well-formed and the composer refused it: a
+    * conflict with the sheet's state, not a malformed call. {@code getErrors()} is surfaced as its
+    * own field as well as joined into {@code error}, since the composer often reports several and
+    * joining them alone loses the boundaries.
+    */
+   @ExceptionHandler(CommandErrorException.class)
+   public ResponseEntity<Map<String, Object>> handleCommandError(CommandErrorException e) {
+      LOG.warn("Composer reported an error to the wiz agent: {}", e.getMessage());
+
+      Map<String, Object> payload = new HashMap<>();
+      payload.put("error", e.getMessage());
+      payload.put("errors", e.getErrors());
+      return new ResponseEntity<>(payload, null, HttpStatus.CONFLICT);
    }
 
    private static final Logger LOG = LoggerFactory.getLogger(WizControllerErrorHandler.class);
--- a/core/src/main/java/inetsoft/web/wiz/WizControllerErrorHandler.java
+++ b/core/src/main/java/inetsoft/web/wiz/WizControllerErrorHandler.java
@@ -80,5 +80,29 @@ public class WizControllerErrorHandler {
       return new ResponseEntity<>(payload, null, HttpStatus.UNPROCESSABLE_ENTITY);
    }
 
+   /**
+    * Maps an input-validation failure to 400 <em>carrying its message</em>.
+    *
+    * <p>The wiz agent services express every bad-request condition as an
+    * {@code IllegalArgumentException} with a caller-facing message — ~75 sites in
+    * {@code inetsoft.web.wiz.viewsheet} alone ({@code ViewsheetEditService},
+    * {@code PropertyPath}, {@code ConditionVocabulary}, {@code DateComparisonService}, …). Without
+    * this handler every one of them fell through as a generic 500 with the message discarded, so a
+    * caller saw "Request failed with status code 500" for a fixable bad request and had no way to
+    * learn which field was wrong. That made correct validation indistinguishable from a server bug.
+    *
+    * <p>The message goes in the {@code error} key deliberately: that is the field wiz-services'
+    * {@code handleError} extracts, so the text reaches the agent rather than being replaced by a
+    * bare status code.
+    */
+   @ExceptionHandler(IllegalArgumentException.class)
+   public ResponseEntity<Map<String, String>> handleIllegalArgument(IllegalArgumentException e) {
+      LOG.warn("Invalid wiz request: {}", e.getMessage());
+
+      Map<String, String> payload = new HashMap<>();
+      payload.put("error", e.getMessage());
+      return new ResponseEntity<>(payload, null, HttpStatus.BAD_REQUEST);
+   }
+
    private static final Logger LOG = LoggerFactory.getLogger(WizControllerErrorHandler.class);
 }
--- a/core/src/main/java/inetsoft/web/wiz/WizControllerErrorHandler.java
+++ b/core/src/main/java/inetsoft/web/wiz/WizControllerErrorHandler.java
@@ -22,6 +22,8 @@ import inetsoft.util.InvalidUserException;
 import inetsoft.web.wiz.service.UnsupportedDatasourceException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.converter.HttpMessageNotReadableException;
 import org.springframework.http.ResponseEntity;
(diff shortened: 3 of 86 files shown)


larryliang-inetsoft:
## Review — viewsheet/binding agent backend

Read the whole production surface (~8k lines across `wiz/viewsheet/`, `wiz/binding/`, `wiz/dispatch/`) against the base branch. The shape is right: validation in the controller, delegation to the Composer's own services, one `mutate` = one checkpoint, and the vocabulary/alias layers keep FQCNs out of the agent's face. `CapturingCommandDispatcher` is the correct diagnosis of a real defect.

Eight findings, worst first. Two of them defeat the PR's own stated purpose.

---

### 1. Every one of the 161 hand-written error messages is destroyed before it reaches the agent

`wiz/viewsheet/` and `wiz/binding/` throw `IllegalArgumentException` in **161** places, with messages written specifically for an LLM caller ("Edit op 'move' requires both 'x' and 'y'", "'endDate' and 'endToday' cannot both be set…"). Neither new controller has an `@ExceptionHandler` for it, `WizControllerErrorHandler` handles only `SecurityException`/`UnsupportedDatasourceException`, and `GlobalExceptionHandler` handles only `IOException`. An unhandled `IllegalArgumentException` therefore reaches Boot's default error handling: **HTTP 500**, body `{"error": "Internal Server Error", ...}` with no `message` field (no `server.error.include-message` is configured anywhere in the repo).

wiz-services then does this (`viewsheetAgentService.ts:23`):

```ts
const detail = e?.response?.data?.error ?? e?.message ?? "unknown";
```

`data.error` exists and reads `"Internal Server Error"`, so the tool reports:

> `Viewsheet agent op 'edit' failed: Internal Server Error`

Status is wrong too (500 for a caller mistake, so the plugin cannot distinguish "you sent bad input" from "the server broke").

`CommandErrorException` has the same problem, and that one is worse: it extends `Exception`, has no handler, and its message *is* the captured composer error text. So the mechanism this PR exists to add — turn a silent composer ERROR into something the caller can see — hands the caller `"Internal Server Error"`. The capture works; the delivery doesn't.

The sibling controller already solved this. `WorksheetAgentController` throws `ResponseStatusException` for validation and has `@ExceptionHandler(ResponseStatusException.class)` putting `getReason()` into `error`. These two controllers have 4 `ResponseStatusException` against 161 `IllegalArgumentException`.

Add to both controllers (or to `WizControllerErrorHandler`, which already covers `basePackages = "inetsoft.web.wiz"`):

- `@ExceptionHandler(IllegalArgumentException.class)` → 400, `{"error": e.getMessage()}`
- `@ExceptionHandler(CommandErrorException.class)` → 400 (or 409), `{"error": …, "errors": e.getErrors()}` — `getErrors()` is already there and currently nothing reads it

`ViewsheetAssemblyAgentControllerTest` is two tests (feature flag, model passthrough) and asserts nothing about status mapping, which is why this got through. A test that posts an invalid `EditRequest` and asserts 400-with-message would pin it.

### 2. `detach` closes any session by token alone — no ownership check, no feature gate

```java
@PostMapping("/api/wiz/v1/agent/viewsheet/{sessionToken}/detach")
public void detach(@PathVariable String sessionToken, Principal user) {
   sessionService.close(sessionToken);
}
```

`user` is accepted and ignored. Every other endpoint goes through `sessions.resolve(sessionToken, user)`, which binds the token to `agentKey(agent)`. This one doesn't, so any authenticated caller who holds or guesses another user's session token terminates their pairing session.

The existing script controller gets this right and is clearly what this was copied from — it just lost the check:

```java
// script/ViewsheetAgentController.java:291
JoinSession s = sessionService.resolve(sessionToken, agentKey(user));
if(s != null) { sessionService.close(sessionToken); }
```

`BindingAgentController:438` has the identical omission. Skipping `requireEnabled()` here is deliberate elsewhere ("always allowed to disconnect") and fine — the missing ownership resolve is not.

### 3. `CapturingCommandDispatcher` doesn't override `detach()`, so errors still escape on the rename path

`CommandDispatcher.detach()` returns `new CommandDispatcher(this, clientId)` — the **base** class. Commands sent through the detached copy are neither captured nor inspected, which is exactly the hole this class exists to close.

It is reachable from these ops, not hypothetically:

- `CoreLifecycleService.createList(...)` → `dispatcher.detach()` (lines 229, 238), handed to `ReadyListener`
- `ComposerObjectService:232` and `:367` call `createList` — the `add` / `remove` ops
- `VSObjectPropertyService:450` calls `createList` — the **`rename`** op, which the PR description singles out as "the capturing dispatcher matters most here"
- `ComposerGroupService:232` — the `group` op

An ERROR dispatched during that refresh is silently dropped, `getErrors()` stays empty, and the op reports success. Fix is small: override `detach()` to return a capturing copy sharing the same `captured` list.

### 4. The base `commands` list is never populated, so framework code that reads the dispatcher's own history sees nothing

`sendCommand` deliberately skips `super`, so `commands` stays empty and the inherited `iterator()` / `stream()` return nothing. Two consumers read exactly that:

- `CoreLifecycleService:185` — `commandDispatcher.stream()…anyMatch(CheckMVEvent)` computes `checkMVHandled`. Always false here, so the MV path re-sends `CheckMVEvent` **and adds a second checkpoint** at line 210, breaking the one-checkpoint-per-call invariant `mutate()` promises.
- `CoreLifecycleService:515` — `for(CommandDispatcher.Command command : dispatcher)` detects `InitGridCommand` to set `inited` in `refreshViewsheet`. Always false here.

`withDummyDispatcher` has the same hole, so this is parity rather than regression — but this class is the one claiming fidelity, and overriding `iterator()`/`stream()` to return `captured` is ~6 lines.

### 5. A failed mutation leaves the runtime diverged from the browser, with no checkpoint

`ViewsheetSessionService.mutate`:

```java
CapturingCommandDispatcher.withCapturingDispatcher(agent, dispatcher -> { mutation.run(...); return null; });
// throws CommandErrorException here if anything ERRORed
Viewsheet vs = rvs.getViewsheet();
if(vs != null) { rvs.addCheckpoint(vs.prepareCheckpoint()); }
broadcast.broadcastRefresh(...);
```

Composer services partially apply before they ERROR (that's precisely why they ERROR instead of throwing). When the exception fires, the runtime holds a half-applied edit, no checkpoint was taken, and **no broadcast was sent** — so the human's Composer keeps rendering pre-edit state while the runtime has moved. The next agent edit builds on the divergence, and the user's Ctrl+Z has no checkpoint for the partial change.

Options, in order of preference: snapshot before the mutation and restore on `CommandErrorException`; or at minimum broadcast in a `finally` so the browser reflects reality, and say in the thrown message that the edit may be partially applied.

### 6. `set_date_comparison` silently destroys a custom period

`DateComparisonService.apply` runs `periods.setCustom(false)` unconditionally, and `validate()` requires either `endDate` or `endToday` on *every* call. So a caller who only wants `useFacet: true` must still supply an anchor, and the call converts an assembly configured with a **custom** period to standard, discarding it — no error, no warning. `read()` will happily have reported that custom period a moment earlier.

Either refuse to write when the existing model is custom and no period fields were supplied, or make the period block conditional on period fields actually being present.

### 7. Vocabulary endpoints ignore the session token

`hyperlinkTypes`, `chartElementVocabulary`, and `conditionVocabulary` take `{sessionToken}` in the path and never resolve it. Static data, so no exposure — but the agent gets a success from `list_hyperlink_types` on a dead session while every other call returns `SESSION_EXPIRED`, which is a confusing signal to reason from. One `sessions.requireSession(sessionToken, user)` each makes the surface uniform.

### 8. Smaller things

- **`distribute` equalizes edges, not gaps.** `distributeMoves` spreads left/top *positions* evenly. With assemblies of differing widths the visible gaps come out uneven, which is not what Composer's distribute does or what a caller asking to "distribute horizontally" expects. Distributing the free space between bounding boxes would match.
- **The null-cluster rationale in `withCapturingDispatcher` is factually wrong.** The comment says `@PostConstruct` is "the only place the cluster is used"; `CommandDispatcherService` also uses it at lines 69, 77, 102 and 126. The null is safe only because `convertAndSendToUser` is overridden to a no-op — which is a fine reason, and the one the comment should give, so a later change doesn't lean on a false premise and NPE.
- **No bounds on `move`/`resize` coordinates.** Nothing rejects a negative `x`/`y` or a zero/negative size, so an agent can park an assembly off-canvas where the user can't find or select it.

---

Nothing here argues against the approach — 1, 2 and 3 are each a few lines, and 1 in particular is worth fixing before any live testing, since right now every validation failure looks identical to a server crash from the plugin's side.

larryliang-inetsoft:
All eight findings are fixed, each driven by a failing test first. Commits `5b3a17b8c`, `88f4673e7`, `f2cb10dd1` (plus `9ef463dfd` / `8e9234b0d` from the live-testing pass).

| # | Finding | Resolution |
|---|---|---|
| 1 | 161 messages destroyed | `IllegalArgumentException` → 400 with the message, `CommandErrorException` → **409** with `errors` as its own field. |
| 2 | `detach` closes any session by token | Both controllers resolve ownership first and close only on a hit, matching the script controller. `requireEnabled` stays skipped. |
| 3 | `detach()` drops refresh errors | Returns `this` — safe precisely because this dispatcher never transmits, and it keeps one captured list. |
| 4 | Empty `commands` → double checkpoint | `iterator()` and `stream()` read the captured list. |
| 5 | Failed mutation diverges from browser | Checkpoint **and** broadcast happen either way; the failure says the edit may be partially applied. |
| 6 | Custom period destroyed | A call setting no period field leaves the period alone; a standard period over a custom one is refused. |
| 7 | Vocabulary endpoints ignore the token | All three resolve. |
| 8 | distribute / comment / bounds | Gaps not edges; negative `move` coordinates refused; the cluster comment corrected. |

Three notes where I did something other than what was suggested, in case you disagree:

**Finding 5** — you offered snapshot-and-restore or, at minimum, broadcast in a `finally`. I took a middle path: checkpoint *and* broadcast on failure. Restoring would discard a partial edit the user may want to keep or inspect; this way the partial change is real, undoable in one step, and visible. The note rides on a second `CommandErrorException` constructor so `getErrors()` stays exactly what the composer said — callers iterate it.

**Finding 6** — I also refuse to write a standard period over a custom one. Making the block conditional alone would still have let an explicit period call discard a custom configuration silently.

**Finding 8, distribute** — this changes output. Widths 100/300/100 across 0–600 now put the middle assembly at **200** (equal 100px gaps) rather than 300. I judged matching Composer to be the correct contract, but it is not purely additive.

Finding 4 was the most useful thing in the review: I would have read the empty `commands` list as cosmetic. The consequence you traced — `checkMVHandled` always false, so `CheckMVEvent` is re-sent and a **second checkpoint** added, breaking the one-checkpoint-per-call promise `mutate` advertises — is not something I would have found from the code alone.

Gates: **1440** Java wiz tests, 0 failures.

PR #4597, opened 2026-08-14
fix(wiz): a histogram's boundary buckets silently admitted null values

larryliang-inetsoft:
## What

`WizardRecommenderUtil.buildRangeDimension`'s generated bin-assignment JS built the first bucket as a bare `field < min` and the last bucket as an unconditional `else`.

## Why this is wrong

JS coerces `null` to `0` in a relational comparison, so `null < 20` evaluates `true` — every row with no value for the binned measure silently landed in the FIRST bucket. An unconditional `else` on the last bucket would have caught the same rows if only the first branch were guarded, just relocating the bug rather than fixing it.

`NumericRangeRef` (the classic composer "Numeric Range" calc field — an unrelated feature that happens to share the `Range@` naming convention) already guards every branch with `field != null`; this wizard-specific generator never carried that guard over.

## How it was found

Found live sweeping the openproject datasource through the wiz-services plugin (F1: "distribution of estimated hours"). With ~18% of work packages having no estimate, the first bucket ("< 20") read **803** instead of the true **629** — inflated by exactly the null count. Confirmed with an explicit `IS NOT NULL` filter, which reproduced 629 unchanged.

## Fix

Both boundary branches now guard `field != null`, so a null value matches no branch and is excluded from the distribution — consistent with `NumericRangeRef`'s convention. Extracted the bucket-building loop into a package-visible `buildRangeExpression()` so its null-handling can be asserted directly in a plain unit test, with no `RuntimeViewsheet`/sandbox scaffolding required.

## Testing

Added a dedicated regression test asserting the first and last buckets guard against null (and that the trailing bucket is no longer an unconditional `else`). Full `Wiz*Test` + `WizardRecommenderUtilTest` suite: 456 tests pass.

Verified live end-to-end: rebuilt and redeployed the full stack (`local-rebase5`), rebuilt the same histogram — the first bucket now reads exactly 629, and the 174 null-estimate rows surface as their own explicit null group rather than corrupting a real bucket or vanishing silently.

### DIFF
--- a/core/src/main/java/inetsoft/web/vswizard/recommender/WizardRecommenderUtil.java
+++ b/core/src/main/java/inetsoft/web/vswizard/recommender/WizardRecommenderUtil.java
@@ -882,15 +882,51 @@ public final class WizardRecommenderUtil {
       min = vals[0] + inc;
       max = vals[1] > max ? vals[1] - inc : vals[1];
       final int n = (int) Math.round((max - min) / inc) - 1;
+      field = field.replace("'", "\\'");
 
+      RangeExpression built = buildRangeExpression(min, inc, n, field);
+
+      CalculateRef calc = new CalculateRef();
+      calc.setName(refValue);
+      ExpressionRef expr = new ExpressionRef();
+      expr.setExpression(built.expression());
+      expr.setName(refValue);
+      calc.setDataRef(expr);
+      calc.setSQL(false);
+
+      dim.setDataRef(new ColumnRef(calc));
+      rvs.getViewsheet().addCalcField(tname, calc);
+      dim.setGroupColumnValue(calc.getName());
+      dim.setOrder(XConstants.SORT_SPECIFIC);
+      dim.setManualOrderList(built.ranges());
+   }
+
+   /** The generated bin-assignment JS expression, plus the ordered bucket labels it can produce. */
+   record RangeExpression(String expression, ArrayList<String> ranges) {}
+
+   /**
+    * Build the JS if/else-if chain that assigns a row's {@code field} value to one of {@code n + 1}
+    * numeric buckets starting at {@code min} and stepping by {@code inc}. Package-visible (rather than
+    * inlined into {@link #buildRangeDimension}) so the generated expression's null-handling can be
+    * asserted directly in a plain unit test, with no {@code RuntimeViewsheet}/sandbox scaffolding.
+    *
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (changes requested):
Requesting changes based on Claude's automated review. It found the fix only guards the first (< min) and last (> max) histogram buckets — the middle buckets (field >= X && field < Y) are unchanged and remain unguarded. Since null coerces to 0 in JS relational comparisons, any middle bucket whose range straddles zero (X <= 0 < Y) will still silently admit null, just relocated to a different bucket — the exact failure mode this PR's own description says an incomplete fix would produce ("relocating the bug rather than fixing it"). This is reachable for any numeric field whose range crosses zero (e.g. profit/loss), not just non-negative measures like the motivating estimated_hours case, and isn't caught by the new tests (which only exercise positive-boundary ranges).\n\nSuggested resolution: guard every branch unconditionally with `field != null` (matching NumericRangeRef's existing convention), or explicitly document/confirm why `min` can never be <= 0 for any caller.

larryliang-inetsoft:
## Re-review — the bot's finding from 2026-08-14 is correct and still unaddressed

Confirming independently, with a concrete counter-example: the fix guards only the first bucket (`field != null && field < min`) and last bucket (`else if(field != null)`); middle buckets are unchanged (`else if(field >= X && field < Y)`). JS coerces `null` to `0` in relational comparisons, so a middle-bucket guard is only safe when that bucket's lower threshold `X > 0` — true for the motivating field (`estimated_hours`, non-negative) but not guaranteed for every numeric field this shared histogram generator (`buildRangeDimension`/`buildRangeExpression`) runs on.

Concrete case: `min`/`max` spanning e.g. -40..60, `inc = 20` — the bucket sequence includes `field >= 0 && field < 20`. A null coerces to `0`: `0 >= 0` and `0 < 20` both true, so the null lands silently in that bucket instead of being excluded — the same "relocating the bug rather than fixing it" failure this PR's own description says is unacceptable for the last-bucket case, just reachable in a middle bucket for any field whose range crosses zero (profit/loss, temperature delta, balance).

The two new tests only call `buildRangeExpression(20, 20, 6, "estimated_hours")`, where every threshold is positive — so they can't catch this. Simplest fix, matching the bot's suggestion and `NumericRangeRef`'s existing convention: guard every branch unconditionally with `field != null` rather than reasoning about which ones need it.

It's been 3 days with no reply to the original finding — flagging in case it fell off the radar rather than being consciously deferred.

PR #4598, opened 2026-08-14
Fix schedule task editor UI regressions

swaker854:
## Summary
- Scope the monthly condition dropdown's focus ring to the custom-select itself instead of bleeding over the "day of the month"/"of the month" labels
- Remove the redundant orange inset box-shadow and stray underline on the selected/hovered condition-type tab
- Fix the Run Once and Options date pickers rendering with no calendar grid (`NgbDatepickerMonth` was missing from imports, so `<ngb-datepicker-month>` rendered as an inert unbound element)
- Fix the Action tab's Save button using `btn-secondary` instead of `btn-primary` like the Condition/Options tabs
- Fix a crash moving tasks when the folder tree selection isn't exactly one node (ctrl/shift-click) by disabling Move and guarding `moveTasks()` on the existing `parentFolder` getter
- Use the shared `<modal-header>` component in the edit task folder dialog instead of hand-rolled header markup

## Test plan
- [x] Verified live in browser: Monthly condition dropdown focus ring, tab hover/selected styling, Run Once + Options date pickers rendering full calendar grids, Deliver-to-email content-as options (see stacked PR on top of this one)
- [ ] Confirm Move button is disabled when the folder tree has 0 or >1 nodes selected, and enabled/functional with exactly 1 selected

### DIFF
--- a/web/projects/portal/src/app/portal/schedule/schedule-task-editor/actions/task-action-pane.component.html
+++ b/web/projects/portal/src/app/portal/schedule/schedule-task-editor/actions/task-action-pane.component.html
@@ -68,7 +68,7 @@
         [disabled]="!isValid()">
         _#(Multiple Actions)
       </button>
-      <button type="button" class="btn btn-secondary" (click)="save(true)"
+      <button type="button" class="btn btn-primary" (click)="save(true)"
       [disabled]="!isValid()">_#(Save)</button>
       <button type="button" class="btn btn-secondary" (click)="closeEditor.emit(model)">
         _#(Close)
--- a/web/projects/portal/src/app/portal/schedule/schedule-task-editor/conditions/task-condition-pane.component.scss
+++ b/web/projects/portal/src/app/portal/schedule/schedule-task-editor/conditions/task-condition-pane.component.scss
@@ -55,7 +55,6 @@
       background-color: var(--inet-shell-surface-default);
       color: var(--inet-text-color);
       border-color: var(--inet-default-border-color);
-      box-shadow: inset 0 -2px 0 var(--inet-navbar-selected-border-color);
     }
   }
 
@@ -102,6 +101,7 @@
 .condition-edit-view__type-link:hover {
   background-color: var(--inet-ui-neutral-hover-bg-color);
   color: var(--inet-text-color);
+  text-decoration: none;
 }
 
 .condition-edit-view__card {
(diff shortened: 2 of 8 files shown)


swaker854:
should merge after 4595 merge

jshobe-inetsoft (approved):
## Code Review: PR #4598 — Fix schedule task editor UI regressions

### Summary
Six small, independent UI-regression fixes to the schedule task editor/list. Verified each against the current code (base branch `bug-scheduleTab-0813`):

1. **Monthly condition dropdown focus ring** (`task-condition-pane.component.scss`) — new `:focus-within` override on `.condition-edit-view__monthly-input-group` suppresses the group-level box-shadow/outline so it doesn't bleed onto the adjacent labels, since `custom-select` already draws its own focus ring. The accompanying comment clearly explains the rationale. Low risk — purely additive CSS scoped to one input group.
2. **Condition-type tab styling** — removes a redundant inset box-shadow and adds `text-decoration: none` on hover. Cosmetic, no logic impact.
3. **Run Once / Options date pickers rendering with no grid** — root cause confirmed: `NgbDatepickerMonth` was missing from both components' `imports` array while `<ngb-datepicker-month>` is used in both templates (`task-condition-pane.component.html:544`, `task-options-pane.component.html:205`). Adding it to `imports` is the correct, minimal fix for a standalone Angular component.
4. **Action tab Save button** — `btn-secondary` → `btn-primary`, matching the Condition/Options tabs. Cosmetic only.
5. **Move-tasks crash on non-single folder selection** — traced `parentFolder` getter (`schedule-task-list.component.ts:460`): returns `null` unless exactly one node is selected. `moveTasks()` now returns early via `!this.parentFolder`, and the button's `[disabled]` binding was updated to match, so the crash path in `moveItems()` (`this.parentFolder.path` on `schedule-task-list.component.ts:996`, only reached when opening the move dialog without a pre-set target) can no longer be hit with 0 or >1 selected nodes. Correct and consistent — the guard exists at both the UI (button disabled) and logic (early return) layers.
6. **Edit task folder dialog header** — replaces hand-rolled header markup with the shared `<modal-header>` component, using its actual `[title]`/`(onCancel)` API (verified against `ModalHeaderComponent`). Reduces duplication, no behavior change.

### Other checks
- **Security**: no new user input handling, no injection surface.
- **Backward compatibility**: no public API/selector changes; `parentFolder` getter behavior is unchanged, only newly consulted by `moveTasks()`.
- **Tests**: no test files touched; none of the changes have existing test coverage in this diff, but none of the existing tests reference the modified code paths, so nothing is broken.
- **Regressions**: each fix is scoped to the specific reported symptom; no shared utility or base class was touched.

No blocking issues found.

PR #4599, opened 2026-08-14
Redesign schedule action deliver/save content-as options

swaker854:
## Summary
Replace the Deliver-to-email and Save-to-server "content as" row of toggle switches with a grouped radio (Match layout / Expand components) plus an indented block of checkboxes, per the approved 1B/2A design.

Key behavior change: controls no longer appear/disappear based on format. They stay visible, disabled, with an inline reason ("Excel only", "Needs Expand components", "Not used with data-only export", "Requires Expand Components permission", "Not applied to HTML/CSV") instead of being hidden or reported as a post-save warning. The old combined html-match-message warning is removed since its information is now conveyed per-control.

Applied identically to:
- portal `action-accordion.component` (email + save-to-server panes)
- EM `delivery-emails.component` and `server-save.component` (Angular Material radios/checkboxes, same grouping/indentation/reason pattern)

## Test plan
- [x] Verified live in browser: toggling Deliver to Emails, switching Match Layout / Expand Components, confirming Expand Selection List/Tree, Only Data Elements, and Export All Tabbed Tables show correct disabled state + reason text, and spacing before Attachment Name is clean (no overlap)
- [ ] Repeat the same check in EM's delivery-emails and server-save panes

### DIFF
--- a/web/projects/em/src/app/settings/schedule/task-action-pane/delivery-emails/delivery-emails.component.html
+++ b/web/projects/em/src/app/settings/schedule/task-action-pane/delivery-emails/delivery-emails.component.html
@@ -71,32 +71,43 @@
           <mat-error *ngIf="form.controls.format?.errors?.required">_#(em.schedule.action.formatRequired)</mat-error>
         </mat-form-field>
         @if (dataSizeOptionVisible) {
-          <mat-radio-group class="flex-col left-padding" formControlName="emailMatchLayout"
-            (change)="fireDeliveryChanged()">
-            <mat-radio-button [value]="true">_#(Match Layout)</mat-radio-button>
-            <mat-radio-button [value]="false"
-            [disabled]="!expandEnabled">_#(Expand Components)</mat-radio-button>
-          </mat-radio-group>
-        }
-        <div class="full-width">
-          @if (dataSizeOptionVisible) {
-            <mat-checkbox class="left-padding"
-              formControlName="emailExpandSelections"
+          <div class="option-group">
+            <div class="group-label left-padding">_#(Deliver content as)</div>
+            <mat-radio-group class="flex-col left-padding"
+              formControlName="emailMatchLayout"
               (change)="fireDeliveryChanged()">
-              _#(Expand Selection List/Tree)
-            </mat-checkbox>
-          }
-          @if (type == 'viewsheet' && format == 'Excel') {
-            <mat-checkbox class="left-padding"
-              formControlName="emailOnlyDataComponents"
-              (change)="fireDeliveryChanged()">_#(Only Data Elements)
-            </mat-checkbox>
-          }
-        </div>
-        @if (type == 'viewsheet' && format == 'Excel') {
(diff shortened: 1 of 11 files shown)


swaker854:
should merge after 4598 merge

jshobe-inetsoft (approved):
### Code Review: PR #4599 — Redesign schedule action deliver/save content-as options

**Scope:** Replaces the toggle-switch "content as" row with a grouped radio (Match Layout / Expand Components) plus an indented checkbox block, applied identically across EM (`delivery-emails.component`, `server-save.component`) and portal (`action-accordion.component`). Controls now stay visible-but-disabled with an inline reason instead of appearing/disappearing or surfacing a separate warning banner.

**Correctness**
- Traced the new `dataSizeOptionVisible` (EM) / same-named getter (portal): dropped the `format !== "HTML"/"CSV"` exclusion so the group is now always shown for viewsheet (minus the pre-existing PDF+print-layout exclusion), matching the stated "stay visible, disabled" design.
- The disable/reason logic is symmetric between the EM reactive-forms implementation (`updateEnable()`) and the portal template-driven implementation (`emailExpandSelectionsDisabled`, `emailOnlyDataDisabled`, etc.) — same predicates, same reason precedence (permission → mode → data-only conflict), in both places.
- Verified disabled `FormControl`/`ngModel` values are still readable via the individual getters used when emitting `deliveryChanged`/model updates (e.g. `this.emailMatchLayout` reads `form.get(...).value` directly), so newly-disabled-but-visible controls don't lose their value on save — no data-loss regression from the show-disabled change.
- EM's `changeFormat()` no longer manually enables/disables `emailMatchLayout`; that's now handled by the `form.valueChanges` subscription calling `updateEnable()` (format is part of the form group), so behavior is preserved via the reactive path rather than lost.
- The old CSV case previously left `emailMatchLayout` *enabled* while hidden (only HTML was explicitly disabled in `changeFormat()`); the new `emailLayoutUnavailable` correctly disables for both HTML and CSV — a fix, not a regression, since the control was never visible for CSV before either.
- Test updates (`*.tl.spec.ts`) match the new behavior exactly (disabled-with-reason assertions instead of absent-element assertions), and add coverage for the HTML/CSV/Excel reason branches.

**Non-blocking notes**
1. `action-accordion.component.ts` adds a `get saveExpandActive()` getter that isn't referenced anywhere else in the file (`saveExpandSelectionsDisabled`/`saveExpandSelectionsReason` inline the same condition instead) — dead code, harmless but could be removed or wired in for consistency with the email-side `emailExpandActive` pattern.
2. The PR's own test plan has "Repeat the same check in EM's delivery-emails and server-save panes" unchecked — worth confirming that manual pass before merge, since the portal and EM implementations are hand-duplicated rather than shared.

**Security / backward compatibility / test breakage:** None found. Purely presentational, no new user input handling, no API/model shape changes, and no existing tests broken by the changes (spec updates match the new behavior 1:1).

**Verdict: No blocking issues.**

PR #4600, opened 2026-08-14
Broaden parentFolder to match currentFolder's selection handling

swaker854:
## Summary
- `parentFolder` previously required `selectedNodes.length == 1`, which is stricter than `currentFolder` (`selectedNodes.length > 0`) — the getter that actually decides which folder's tasks the list displays. Move now works with the same selection rule as the rest of the page: any non-empty selection uses `selectedNodes[0]`.
- Falls back to the root folder when no tree node is selected at all, instead of leaving Move unusable until a folder is explicitly picked.

## Test plan
- [ ] Move tasks with a single folder selected in the tree (existing behavior, should still work)
- [ ] Move tasks with multiple folders selected (ctrl/shift-click) — should now use the first selected node instead of disabling Move
- [ ] Move tasks with no folder selected — should default to the root folder instead of disabling Move

### DIFF
--- a/web/projects/portal/src/app/portal/schedule/schedule-task-list/schedule-task-list.component.ts
+++ b/web/projects/portal/src/app/portal/schedule/schedule-task-list/schedule-task-list.component.ts
@@ -458,8 +458,12 @@ export class ScheduleTaskListComponent implements OnInit, OnDestroy, AfterConten
    }
 
    get parentFolder(): AssetEntry {
-      if(!this.showTasksAsList && this.selectedNodes.length == 1) {
+      if(!this.showTasksAsList && this.selectedNodes.length > 0) {
          return this.selectedNodes[0].data;
+      }
+
+      if(!this.showTasksAsList && !!this.rootNode) {
+         return this.rootNode.data;
       }
 
       return null;


swaker854:
merge after 4599 merge

jshobe-inetsoft (approved):
## Code Review — PR #4600

### Summary
`parentFolder` previously required exactly one selected tree node (`selectedNodes.length == 1`), while the sibling `currentFolder` getter (used to decide which folder's tasks are loaded/displayed) already accepted any non-empty selection (`length > 0`) with a root-node fallback when nothing is selected. This PR makes `parentFolder` mirror `currentFolder`'s exact logic.

### Correctness
- Verified against the current file: the new `parentFolder` getter (lines 460-467) is now byte-for-byte identical in structure to the existing `currentFolder` getter (lines 239-246) — a legitimate consistency fix, not a new/independent behavior.
- Traced all real consumers of the `parentFolder` getter:
  - `moveTasks()` / `moveItems()` (lines 943, 1000) — the primary target per the PR description. Previously, multi-selecting folders (ctrl/shift-click) silently disabled Move (`parentFolder` → `null`) even though `currentFolder` (which drives the visible task list) was perfectly happy with a multi-selection. Now Move works with the same rule as the rest of the page, and falls back to the root folder instead of being unusable when nothing is selected.
  - `newTask()` (line 413) — also consumes the getter. Previously, multi-select or empty-select passed `parentEntry: null`; now it passes the first selected node or root respectively. This is a natural, positive side effect of unifying the two getters, not a separate untested code path.
  - `newFolder(parentFolder: any)` (line 427) — **not** affected: the parameter name shadows the getter, and its only call site passes an explicit `entry`, not `this.parentFolder`.

### Test coverage
- No existing test in `schedule-task-list.component.interaction.tl.spec.ts` or `.risk.tl.spec.ts` exercises `parentFolder` directly, so nothing is broken by broadening it.
- The only related test suite ("currentFolder getter") covers the getter this PR did not touch, and its assertions (null on empty selection when `rootNode` isn't set in the test harness, first-node otherwise) are unaffected.
- No new automated test was added for the widened `parentFolder` behavior itself; the PR instead ships a manual test plan. Non-blocking given the change is a small, mechanical alignment with already-tested logic, but a unit test mirroring the `currentFolder getter` suite would have been easy to add and would guard this getter going forward.

### Security / backward compatibility
No security concerns (no new external input handling). No backward-incompatible API changes — this is a private component getter with a strictly widened (superset) return domain; no caller relies on the old `null`-on-multi-select behavior as a feature.

### Verdict
**Approve.** No blocking issues. All checks pass (`build`, `claude-review`, `Semgrep Scan`, `Semgrep OSS`, `license/cla` all SUCCESS) and the PR is cleanly mergeable with no conflicts.

PR #4603, opened 2026-08-14
Tell the wiz portal how each data source can be annotated

fredaan:
Adds `annotationClass` to the browser entry the wiz portal reads, so it knows what it has to ask of the user before a data source can be indexed: index it silently, demand documentation first, or refuse outright.

| value | meaning |
|---|---|
| `JDBC` | relational; tables and columns come from JDBC metadata |
| `FILE` | browsable tree of files or sheets |
| `METADATA` | the service can be asked what it holds — collections, entities, objects |
| `ENDPOINT_CATALOG` | ships an `endpoints.json`; the callable set is known offline and exactly |
| `DOCUMENT_REQUIRED` | a REST service with no catalogue — the user must supply documentation |
| `UNSUPPORTED` | output shape decided by a user script; no stable target to annotate |
| `UNKNOWN` | classification failed, usually a plugin that did not load |

## Why the query class, not the data source

What separates these cases is *how targets are discovered*, and that lives on the query rather than on the connection. The query class also owns the `endpoints.json` resource, which is the entire `ENDPOINT_CATALOG` test.

## The ordering cannot be tidied

`EndpointJsonQuery extends RestJsonQuery`, so checking REST before the catalogue would classify **all 65 catalogued connectors as needing documentation they already ship with**. A test covers precisely that: a class that is both a catalogue and a REST query has to come out as the catalogue.

## Two deliberate asymmetries

**The catalogue test is a resource lookup, the REST test is a name match.** Shipping an `endpoints.json` *is* the property being asked about, so looking the file up is the most direct statement of it. `AbstractRestQuery`, on the other hand, lives in a connector module core does not depend on and cannot be named directly — so a test-only class with that fully-qualified name exists to exercise the match, and to be the thing that breaks loudly if the real class is ever moved or renamed. Without it, a rename would silently reclassify every REST source as `METADATA`.

**`UNKNOWN` is distinct from `UNSUPPORTED`.** A connector whose plugin did not load cannot be classified, and reporting "this cannot be annotated" when the truth is "we could not tell" sends the user looking in the wrong place for a fix that does not exist.

## What writing the tests exposed

The resource lookup is sharper than it looks. With `endpoints.json` sitting in the shared test package, **every query class beside it resolved that file** and the entire suite classified as `ENDPOINT_CATALOG` — three tests failed for a reason that had nothing to do with the code under test. Real connectors satisfy the constraint naturally by having one folder each; the fixtures now do the same by living in their own package.

## Testing

```
WizDatasourceAnnotationClassTest      6 tests, all green
WizDatabaseControllerSecurityTest     9 tests, unchanged (no regression)
```

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
@@ -26,6 +26,8 @@ import inetsoft.uql.XDataSource;
 import inetsoft.uql.XRepository;
 import inetsoft.uql.jdbc.JDBCDataSource;
 import inetsoft.uql.jdbc.SQLHelper;
+import inetsoft.uql.tabular.ScriptedQuery;
+import inetsoft.uql.tabular.SelectableTabularQuery;
 import inetsoft.uql.util.Config;
 import inetsoft.util.audit.ActionRecord;
 import inetsoft.web.admin.content.database.*;
@@ -582,6 +584,7 @@ public class WizDatabaseController {
       boolean folder = isFolder(nodeType);
       String sourceType = null;
       String databaseType = null;
+      String annotationClass = null;
 
       if(!folder) {
          try {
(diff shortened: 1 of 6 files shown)


jshobe-inetsoft (approved):
Approving based on Claude's automated review, which found no blocking issues — the classification logic, resource-lookup approach, and new test coverage are all sound. Checks pass and there are no conflicts.

PR #4604, opened 2026-08-14
Endpoint descriptions, and a test that they cannot silently break a connector

fredaan:
Adds an optional `description` to `AbstractEndpoint` so `endpoints.json` can carry what an endpoint is *for*, and fills it in for Stripe (101 of 110).

This is the first step of giving the wiz portal something to retrieve on. An endpoint name alone carries very little — "Charges" does not say whether it holds payments, disputes or payouts, and several connectors have a dozen names that read alike.

## The Java change has to land before any JSON is edited

The loader builds its mapper with a bare `new ObjectMapper()`, so `FAIL_ON_UNKNOWN_PROPERTIES` is on, and `AbstractEndpoint` carries no `@JsonIgnoreProperties`. A key with no matching property throws — and `Endpoints.load` catches that as an `IOException` and returns an **empty map**. The connector keeps starting, keeps being offered in the UI, and simply has no endpoints. One `LOG.error` and a blank dropdown.

Verified rather than assumed — removing the property and rerunning gives:

```
UnrecognizedPropertyException: Unrecognized field "description"
  (class inetsoft.test.TestEndpoint), not marked as ignorable
  (5 known properties: "suffix", "name", "descriptionX", "paged", "lookups")
```

## Nothing was guarding those 65 files

The per-connector query tests are tagged `endpoints`, which the build excludes (`<excludedGroups>integration,slow,endpoints</excludedGroups>`) because they call the vendor's live API. In a normal run they contribute **zero** cases. That left ~65 hand-edited resource files with no test behind them, and the one way to break them is the way that leaves no trace.

`EndpointsJsonLoadableTest` parses every one through the loader's own mapper and each connector's own `Endpoints` subclass. Injecting an unknown field into a single file turns it red, naming the file and the property.

`createObjectMapper` is package-private now so the test uses the *real* mapper rather than a copy that would drift from it — and what it guards is precisely a mapper configuration effect.

## Two things that test taught me while it was being written

**Fifteen connectors declare endpoint properties of their own** — `pageType`, `post`, `pagePath`, `pageRequiredParameter` — so binding them all to the shared base class reports failures production never sees. Each has to bind to its own class. (I had generalized from `StripeEndpoint` being an empty subclass; it is not representative.)

**`Class.getResource` returns `test-classes` for this package**, because the test's own package shadows the resource package of the same name. The first version therefore searched a directory containing no JSON at all — and passed. It now enumerates every classpath entry and asserts it found files, since "found none" and "looked in the wrong place" are otherwise the same green.

## The descriptions

Written against Stripe's official OpenAPI spec, **aligned by URL path rather than by name** — `endpoints.json` names are StyleBI's own ("Balance Transactions") and appear nowhere in the vendor's documentation, while `/v1/balance_transactions` is a verbatim anchor. The response schema supplied the field list.

They say what the resource *is* and when to reach for it, not what the call does. "Returns a list of charges" is what the vendor's reference already says, and is useless for choosing among a dozen similar names. Several carry an explicit boundary instead:

> **Payouts** — Transfers of your Stripe balance to your own bank account […] Use for settlement timing and failed payouts — customer-facing revenue is in Charges, not here.

> **Payment Intents** — […] Prefer this over Charges when the question involves abandoned or failed attempts, since a charge only comes into existence once an attempt succeeds.

## Nine endpoints left undescribed on purpose

Orders, Order Returns, SKUs, Upcoming Invoice and usage record summaries: **Stripe has removed them**, and no path in the current spec matches. Describing an endpoint that answers 404 would only help a model choose it.

Worth noting for the design this feeds: `endpoints.json` and the vendor's current API are not in a subset relationship but a *crossing* one — 8% of Stripe's entries here are dead.

## Two naming defects documented rather than corrected

The names are part of the saved query format, so renaming them would break existing configurations.

- **`Country` is the LIST endpoint and `Countries` takes a single id** — exactly the reverse of how they read. Both descriptions say so outright.
- **`Account Bank Accounts` and `Account Cards` resolve to the same URL with no filter between them**, unlike `Bank Accounts`/`Cards` which really do split on `object=bank_account` and `object=card`.

## Testing

```
EndpointsJsonLoadableTest    1 test  — all 65 connectors parse
EndpointJsonQueryTest       36 tests — includes description present/absent
```

Both mutation-checked: removing the property, and injecting an unknown field, each turn a test red.

### DIFF
--- a/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/json/AbstractEndpoint.java
+++ b/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/json/AbstractEndpoint.java
@@ -31,6 +31,25 @@ public class AbstractEndpoint implements EndpointJsonQuery.Endpoint {
       this.name = name;
    }
 
+   /**
+    * A human-readable account of what this endpoint returns and when to reach for it.
+    *
+    * <p>Optional, and absent from most {@code endpoints.json} files. It exists for callers that
+    * have to CHOOSE an endpoint rather than invoke one already chosen — a name alone carries very
+    * little: "Charges" does not say whether it holds payments, disputes or payouts, and several
+    * connectors have a dozen names that read alike.</p>
+    *
+    * <p>Deliberately not used at query time. Nothing about executing a request depends on it, so a
+    * missing or stale description can never change what a query does.</p>
+    */
+   public String getDescription() {
+      return description;
+   }
+
+   public void setDescription(String description) {
+      this.description = description;
+   }
+
    @Override
    public String getSuffix() {
       return suffix;
@@ -57,6 +76,11 @@ public class AbstractEndpoint implements EndpointJsonQuery.Endpoint {
       this.lookups = lookups;
    }
 
+   /**
+    * {@code description} is deliberately excluded here and from {@link #hashCode()}: rewording an
+    * endpoint does not make it a different endpoint, and folding it in would make identity depend
+    * on prose that is expected to be revised.
+    */
    @Override
    public boolean equals(Object o) {
       if(this == o) return true;
@@ -74,6 +98,7 @@ public class AbstractEndpoint implements EndpointJsonQuery.Endpoint {
    }
 
    private String name;
+   private String description;
    private String suffix;
    private boolean paged;
    private List<JsonLookupEndpoint> lookups = Collections.emptyList();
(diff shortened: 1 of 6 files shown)


jshobe-inetsoft (changes requested):
Claude's automated review found no correctness, security, or backward-compatibility issues with this PR. However, the required `build` check on the latest commit failed — it exceeded the 45-minute execution limit and was cancelled (https://github.com/inetsoft-technology/stylebi/actions/runs/31805844510). Please re-run/confirm a passing build before this can be merged.

Nelson-cao (approved):

PR #4607, opened 2026-08-14
fix(wiz): window ORDER BY never sorted NULLs last, letting a NULL win a ranking

larryliang-inetsoft:
## Root cause

`WindowExpressionRef.getOrderFragment()` built a bare `ORDER BY field['col'] ASC|DESC` with no null-ordering directive. On Postgres/Oracle, `DESC` defaults to `NULLS FIRST`, so a row with a NULL sort value would win a "largest value" window ranking — e.g.

```
ROW_NUMBER() OVER (PARTITION BY project_name ORDER BY estimated_hours DESC)
```

would pick a work package with a null `estimated_hours` as rank 1 instead of the row with the actual largest value.

Found during a wiz-chat plugin exercise sweep of the `openproject` datasource (defect J6): a `windowColumns` ranking of `estimated_hours DESC` per project surfaced a NULL-valued row as the "top" row for several projects.

## Fix

Prepend a portable `CASE WHEN field['col'] IS NULL THEN 1 ELSE 0 END` synthetic sort key immediately before each real `ORDER BY` key, so NULL always sorts last regardless of `ASC`/`DESC`.

This uses only ANSI-standard `CASE WHEN` + multi-key `ORDER BY` rather than the `NULLS LAST` keyword, because that keyword is not available on every dialect wiz-services' `windowCapability.ts` advertises as window-function-capable:
- SQL Server has no `NULLS LAST`/`NULLS FIRST` syntax at all.
- MySQL didn't add it until 8.0.14 — StyleBI's supported floor is MySQL 8.0.0.

`wiz-services`' `windowColumns.ts` is validation-only (its `overClause()` builder exists only to exercise the same guards that used to emit SQL text; the return value is discarded per its own comment) — `WindowExpressionRef` is the sole generator of the `OVER (...)` SQL, so no `wiz-services` change was needed for this fix.

## Testing

- Updated every pinned `getExpression()`/exact-string assertion across `WindowExpressionRefTest` and `WorksheetTableServiceWindowColumnsTest` to the new generated SQL shape.
- Added two new regression tests pinning the null-ordering guard itself: `orderBy_emitsNullsLastGuard_regardlessOfDirection` (single key, both ASC and DESC) and `orderBy_multipleKeys_eachGetsOwnNullsLastGuard` (multi-key ORDER BY, each key gets its own guard).
- `WindowExpressionRefTest` (34), `WorksheetTableServiceWindowColumnsTest` (24), `WindowRoutingTest` (21), `WindowExpressionDetectionTest` (6) — all pass, 0 failures.

### DIFF
--- a/core/src/main/java/inetsoft/report/lens/WindowTableLens.java
+++ b/core/src/main/java/inetsoft/report/lens/WindowTableLens.java
@@ -274,7 +274,22 @@ public class WindowTableLens extends AbstractTableLens implements TableFilter {
       boolean[] oasc = orderAsc();
 
       for(int i = 0; i < ocols.length; i++) {
-         int r = Tool.compare(table.getObject(a + hrows, ocols[i]), table.getObject(b + hrows, ocols[i]), true, true);
+         Object v1 = table.getObject(a + hrows, ocols[i]);
+         Object v2 = table.getObject(b + hrows, ocols[i]);
+
+         // A null always sorts last regardless of ASC/DESC, matching
+         // WindowExpressionRef.getOrderFragment's SQL-pushdown NULLS-LAST behavior -- Tool.compare
+         // treats null as smallest, which the oasc[i] negation below would otherwise flip to
+         // NULLS-FIRST for an ascending key.
+         if(v1 == null || v2 == null) {
+            if(v1 == v2) {
+               continue;
+            }
+
+            return v1 == null ? 1 : -1;
+         }
+
+         int r = Tool.compare(v1, v2, true, true);
 
          if(r != 0) {
             return oasc[i] ? r : -r;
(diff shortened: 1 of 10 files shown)


larryliang-inetsoft:
## Re-review — the bot's finding from 2026-08-14 is correct and still unaddressed, plus CI and staleness

Confirming the automated review's finding independently rather than just deferring to it: the SQL-pushdown fix in `WindowExpressionRef.getOrderFragment()` is correct (verified the `CASE WHEN col IS NULL THEN 1 ELSE 0 END` guard forces NULLs last regardless of direction, and the new tests cover both directions and multi-key). But `WindowTableLens` — the in-memory fallback used whenever a window column isn't pushable to SQL (non-JDBC source, or an unpushable frame mode) — has its own, separate comparison logic this PR doesn't touch:

```java
// WindowTableLens.java, both compareRows (~277) and orderKeyCompare (~886) — identical shape
int r = Tool.compare(table.getObject(a + hrows, ocols[i]), table.getObject(b + hrows, ocols[i]), true, true);
if(r != 0) {
   return oasc[i] ? r : -r;
}
```

`CoreTool.compare` treats a null as always "smaller" (`v1 == null → -1`). For **ascending** order (`oasc[i] == true`), `r` is returned unchanged — a null sorts *first*, not last. (Descending happens to come out right, since negating flips it to last — which may be why this wasn't caught by symmetric testing.) So after this PR, the same `windowColumns` config with ascending order produces a different top-ranked row depending on whether the engine happens to push the window to SQL or falls back to `WindowTableLens` — an internal, invisible-to-the-caller implementation detail deciding correctness. That's the same "NULL wins a ranking" bug, reintroduced on the fallback path for ascending order specifically.

Separately, worth flagging since it affects whether this is even the right diff to review: `mergeable_state` is `unstable` and the branch isn't based on the current tip of `stylebi-wiz-main-rebase` — the file list includes `ViewsheetRuntimeController.java`/`FkIntegrityService.java`/`WizVsService.java` and their tests, which have nothing to do with NULLS-LAST and are presumably picked up from the base having moved past this branch. `claude-review`'s check-run has been reporting `failure` (erroring out, not completing) since 2026-08-15. This needs a rebase before the diff itself can be trusted, on top of the `WindowTableLens` gap above.

PR #4608, opened 2026-08-14
Bug #76079, carry the source org theme to the copied organization

ngovus-inetsoft:
## Issue

Bug #76079 — when copying an organization, the theme isn't carried over.

1. Create theme `theme1` in EM → Settings → Presentation → Themes
2. EM → Settings → Security → Users, select **Host Organization**
3. Select `theme1` in Theme, apply
4. Create `organization0` as a copy of **Host Organization**
5. Select `organization0` — Theme reads **Default** instead of `theme1`

## Root cause

A theme created by a site admin while in the host org is created **global**
(`orgID == null`): the themes view hard-sets `result.global = isSiteAdmin && orgId == "host-org"`.

Assigning it to an org (`IdentityService.updateCustomThemeOrganization()`) writes three
pieces of state: the theme's `organizations` list, the `portal.org.selectedThemeId.<orgID>`
pointer, and `Organization.theme`.

`AbstractEditableAuthenticationProvider.copyThemes()` has two branches. The org-owned branch
clones the theme *and* assigns `newOrgThemeId`, which the caller wires onto
`newOrg.setTheme(...)`. The global branch — a shared theme the source org merely selected —
propagated the `organizations` entry and the selection pointer but never assigned
`newOrgThemeId`, so the copied org got `setTheme(null)`. Since the EM Edit Organization pane
reads `organization.getTheme()` (`UserTreeService.java:1089`), it fell back to "Default".

This was already recorded in the codebase as a known gap (`OrgLifecycleThemeOrchestrationTest`,
org-lifecycle matrix row 3c); this PR closes it.

## Fix

Return the shared theme's id from the global branch as well, so all three states agree for
the copied org. The lambda is replaced with a null-checked local because `newOrgThemeId`
can't be assigned from inside `ifPresent`.

Sharing behaviour is unchanged: a global theme is still referenced, never cloned, and its JAR
is not duplicated into the copied org's storage (#74719). The org-owned branch is untouched,
as is `Organization.theme` handling on rename when an edited org is supplied. A rename of an
org that selected a global theme now keeps that theme on `Organization.theme` instead of
dropping it — the same defect on the rename path.

## Testing

`OrgLifecycleThemeOrchestrationTest#copy_globalThemeSelectedByFromOrg_threeStatesAgree`
(previously the gap-documenting test, renamed) now asserts three-state agreement.
The org-lifecycle matrix row 3c is updated to match.

```
mvn test -pl core -Dtest=OrgLifecycleThemeOrchestrationTest,AbstractEditableAuthenticationProviderStaticDepTest
```
Tests run: 32, Failures: 0, Errors: 0 — the 30 existing `copyThemes()` branch tests
(including the global-theme no-clone assertions for #74719) still pass unchanged.


jshobe-inetsoft (approved):
## Code Review

**Change:** `AbstractEditableAuthenticationProvider.copyThemes()`'s global-theme branch (a theme merely *selected* by the source org, not owned by it) now assigns `newOrgThemeId = theme.getId()` in addition to propagating the `organizations` list entry and the `SreeEnv` selection pointer, so all three pieces of theme state agree for the copied/renamed org. Previously only the org-owned branch assigned this return value, so a copied org that had only *selected* a global theme read back `Organization.theme == null` ("Default") in EM despite the other two states pointing at the shared theme.

### Correctness
- Traced the full method: the new `else if` branch is reached only when `theme.getOrgID()` is empty (global) **and** `theme.getOrganizations()` contains `fromOrgId` — i.e., exactly the "shared theme selected by source org" case described in the PR. This is mutually exclusive with the org-owned `if` branch above it, so there's no double-assignment or interference between the two.
- The refactor from `.ifPresent(original -> {...})` to an explicit null check is required (and correctly done) since `newOrgThemeId` is a local `String`, not effectively-final, and can't be reassigned from inside a lambda.
- Sharing semantics are preserved: the global theme is still mutated in place (`themes`, not the `sourceThemes` clone list) and never cloned; only the `organizations` list, the selection pointer, and now `newOrgThemeId` are touched.
- Verified the loop can only match one theme per source org under normal usage (an org selects at most one theme at a time), so the last-assignment-wins pattern already used by the pre-existing org-owned branch is consistent here too — not a new hazard.
- The fix's stated effect on the *rename* path (an org renamed while having selected a global theme now keeps `Organization.theme` set instead of dropping it) follows directly from the same code path being shared between copy and rename (`replace` flag), consistent with the PR description.

### Regression / backward compatibility
- The org-owned branch, the `replace`/rename cleanup logic below the loop, and the no-clone behavior for global themes (#74719) are all untouched by this diff.
- No public API or serialized-format changes.

### Tests
- `OrgLifecycleThemeOrchestrationTest#copy_globalThemeSelectedByFromOrg_threeStatesAgree` (renamed from the gap-documenting `..._organizationThemeFieldStaysNullDespiteSreeEnvAndOrganizationsListAgreeing`) now asserts `Organization.theme` equals the shared theme id instead of asserting it stays null — correctly flips the previously-documented-gap test into a real regression test, and still asserts the theme is not cloned (`finalThemes.size() == 1`).
- The org-lifecycle resource matrix doc (`org-lifecycle-resource-matrix.md`) is updated to match, keeping docs and tests in sync.
- No existing test assertions were weakened; the PR description reports the other 30 `copyThemes()` branch tests pass unchanged, and the diff doesn't touch anything that would put that at risk.

### Security
No new external input handling; no security concerns.

### Minor, non-blocking
- The new "The new org selects the same shared theme..." comment is a little dense but does explain the *why*, consistent with the surrounding code's comment style.

**Verdict: No blocking issues.** The `claude-review` GitHub Action check shows a failure, but that's the automated review bot itself erroring out before producing a review (this comment is the substantive human-equivalent review filling that gap) — not a code-quality signal. All substantive checks (build, Semgrep, Semgrep OSS, CLA) pass, and there are no merge conflicts.

PR #4610, opened 2026-08-14
75878, circular network hover/animation issues in facet graph

milotalon:
Root Cause:

A relation/network chart builds a separate mxGraph for every facet panel — GGraph.initGeometries() calls RelationElement.createGeometry() once per panel with that panel's sub-DataSet, and each call starts with "new mxGraph()". Node ids come from the node's value alone (RelationElement.getId returns the formatted value + "_id"), so they carry no facet qualifier and the same id string is emitted in every panel. Those ids are stamped into the SVG as data-id on nodes and data-source/data-target on edges.

The hover handler collected every relation edge in the SVG and matched them against the hovered node by bare id, so the sibling panel's edge — which carries an identical source/target pair — matched too and stayed highlighted while the rest of that panel dimmed. The nodes appeared to behave because the node lookup map was keyed by id and silently kept only one element per duplicate id; only the edges had no such deduplication, which is why the links stayed lit while the points dimmed.

The same id collision affected the entrance animation: the animation injector's delay map is keyed by bare id, so it retained only the last panel's entry and every panel's edges faded in with the last panel's nodes, making an edge arrive after the node it connects.

Fix:

Give each facet panel an identity and require it to match.

Backend
- New SVGSupport.ATTR_PANEL, emitted as a data-panel attribute.
- New GTool.getFacetPanelId(Coordinate): builds a key from the coordinate's facet parent values, and returns null when there are none, so a non-faceted chart emits no attribute and its DOM is unchanged.
- RelationVO and RelationEdgeVO stamp the key on their annotation groups. Each facet panel has its own coordinate instance, and VGraph.createVisuals passes that same instance to every geometry in the panel, so a panel's nodes and edges always agree on the key.

Frontend
- New relationKey(el, idAttr) helper in ChartInlineSvgDirective producing a panel-qualified node key. It is used everywhere relation identity is compared: the node map, the edge source/target entries, activateRelationNeighbors, the cross-tile hover emit, getRelationEdges and setExternalRelationHighlight. The matching logic itself is unchanged; only the key is wider. An edge never spans panels, so both of its endpoints resolve against the edge's own panel.

Entrance animation
- The node index map, the delay map, the edge delay lookup and the topological depth walk in SVGAnimationDOMInjector are panel-qualified the same way, so each panel now cascades root to leaf on its own topology and each edge fades with its own panel's target node. The "null" root sentinel check still reads the raw data-source, so root detection is unaffected.
- Per-depth stagger counters remain global, so the panels ripple one intra-level spread apart rather than firing in lockstep. Total entrance duration is unchanged, because the intra-level spread is normalised by the node count at that depth.

Tests
- Hover: two panels sharing node ids — hovering a node in one panel activates only that panel's edge and neighbour; a chart with no data-panel still matches by id alone.
- Animation: a two-panel facet pinning the exact per-panel delays and asserting each edge follows its own panel's target node.

### DIFF
--- a/core/src/main/java/inetsoft/graph/internal/GTool.java
+++ b/core/src/main/java/inetsoft/graph/internal/GTool.java
@@ -596,6 +596,29 @@ public final class GTool {
    }
 
    /**
+    * Get a key identifying the facet panel a coordinate belongs to, built from the facet
+    * dimension values assigned to it by the enclosing facet (e.g. {@code {Region=USA East}}).
+    *
+    * <p>Each facet panel gets its own coordinate instance, so this key differs between panels
+    * and is identical for every visual object drawn in one panel.  Returns {@code null} for a
+    * non-faceted chart (no parent values), so callers can omit the key entirely in that case.
+    */
+   public static String getFacetPanelId(Coordinate coord) {
+      if(coord == null) {
+         return null;
+      }
+
+      Map<String, Object> xvals = coord.getParentValues(true);
+      Map<String, Object> yvals = coord.getParentValues(false);
+
+      if(xvals.isEmpty() && yvals.isEmpty()) {
+         return null;
+      }
+
+      return xvals + "/" + yvals;
+   }
+
+   /**
     * Get the topmost VGraph.
     */
    public static VGraph getTopVGraph(Coordinate coord) {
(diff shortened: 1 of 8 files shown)


jshobe-inetsoft (approved):
Claude's automated review (resolved by the author, no blocking issues) plus all passing checks and no conflicts. Approving.

PR #4612, opened 2026-08-14
Bug #76064, resolve the login page theme against the requested organization

ngovus-inetsoft:
## Problem

A tenant reaching its own login page (`organization0.localhost/sree/login.html`) was styled with the **host organization's** theme. Because both themes shared a JAR the base colors looked correct, so the only visible symptom was the host theme's `--inet-text-color` override bleeding through as green text.

The login page and the style sheets it links (`app/global.css`, `app/theme-variables.css`) are requested **before authentication**, so there is no principal and `OrganizationManager.getCurrentOrgID()` falls back to `Organization.getDefaultOrganizationID()`. `GlobalStyleController.getResource()` then read `provider.getOrganization("host-org").getTheme()`. The organization in the URL was never consulted — `AbstractSecurityFilter.authenticateAnonymous()` only reads the `X-INETSOFT-ORGID` cookie and also defaults to the host org, so a first visit to a tenant URL has no org context at all.

## Fix

**`GlobalStyleController`**

- `getRequestOrgID(request)` returns the organization named by the request, or `null` to leave the current one alone. It bails out on single-tenant, on any request that has a principal, and when an org is already pinned on the thread, so the change is strictly additive. It delegates to `SUtil.getLoginOrganization()`, which reads the sub-domain or path per `security.login.orgLocation` and already returns `null` unless the value names a real organization.
- A `getResource(path, user, requestOrgID)` overload wraps the existing lookup in `OrganizationContextHolder.setCurrentOrgId(...)` with `clear()` in a `finally` (request threads are pooled). Pinning through the holder is the fallback `getCurrentOrgID(principal)` already consults, so one change makes the theme lookup, the `Organization` record lookup and the per-org resource cache key all agree — no signature churn through the resolution chain.

**`LoginController`**

- `isCustomTheme(request)` resolves the `customTheme` flag the same way. That flag controls whether `theme-variables.css` is linked at all, so if it resolved against a different organization than the controller serving the file, a tenant with a custom theme on a host org without one would never request its variables.

An authenticated user's own organization still wins in both places.

## Scope notes

The flag and the served style sheet now agree on the *organization*; they can still pick different themes within it, since the flag uses only the selected-theme pointer chain while `GlobalStyleController` prefers `Organization.theme` when it names an existing theme. That split predates this change and is why the resource loader falls back to the `"default"` theme for an unknown id.

`LoginController` also resolves `customLogo` from `getCurrentOrgID()` — the same pre-login defect for the login logo. Left out of scope here.

## Testing

New `GlobalStyleControllerTest` (4 tests) covers: request names an organization → that org's theme; request names none → default org; authenticated → request ignored; pinned context cleared afterwards. On unpatched code it fails with `expected: <theme:test2/...> but was: <theme:def2/...>`.

### DIFF
--- a/core/src/main/java/inetsoft/web/GlobalStyleController.java
+++ b/core/src/main/java/inetsoft/web/GlobalStyleController.java
@@ -22,6 +22,7 @@ import inetsoft.sree.SreeEnv;
 import inetsoft.sree.internal.SUtil;
 import inetsoft.sree.portal.*;
 import inetsoft.sree.security.*;
+import inetsoft.util.ThreadContext;
 import inetsoft.util.Tool;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
@@ -73,7 +74,7 @@ public class GlobalStyleController implements ApplicationContextAware {
          }
       }
 
-      StyleResource resource = getResource(path, user);
+      StyleResource resource = getResource(path, user, getRequestOrgID(request));
 
       // use max-age caching for static resources to minimize the number of requests to load the
       // application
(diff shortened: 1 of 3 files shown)


jasonshobe (approved):
Claude's automated review found no security, correctness, or backward-compatibility issues; the noted suggestions (shared helper for org-resolution duplication, minor test-coverage gap) are non-blocking. All checks pass and there are no merge conflicts.

jshobe-inetsoft (approved):
Approving on behalf of Claude's automated review: no security, correctness, or backward-compatibility issues found; noted suggestions are non-blocking.

PR #4613, opened 2026-08-14
75875, box plot outliers not correctly dimming with rest of box plot areas

milotalon:
Root Cause:
A box plot draws two different kinds of annotation group over one BoxDataSet: the box/whisker glyph (.inetsoft-box, from SchemaVO/BoxPainter) and the outlier markers (.inetsoft-point, from PointVO). Hover dimming is done with server-injected CSS :has() rules, and the box-plot rule only correlated elements of the same class:

svg.ready:has(.inetsoft-box.inetsoft-active) .inetsoft-box:not(.inetsoft-active) So hovering a box dimmed the other boxes but never selected the outlier point groups, which stayed at full opacity.

Fix:
1. SVGAnimationDOMInjector.appendHoverCSS() now correlates the two classes in both directions: an active .inetsoft-box also dims non-active .inetsoft-point, and an active .inetsoft-point also dims non-active .inetsoft-box. These selectors can only match in a document that contains both classes, i.e. a box plot, so no other chart type is affected.
2. So that the hovered box's OWN outliers stay bright, ChartInlineSvgDirective now indexes box-plot annotation groups by the data-group key that a box and its outliers already share (added for the entrance animation in #75643) and activates the whole group on hover (activateBoxGroup), cleared on hover end like the existing treemap/relation activation. data-group is stamped only for box plots, so this is a no-op for every other chart type.

Result: hovering a box keeps that box and its own outlier points bright; every other box and every other box's outliers dim. Hovering an outlier does the same from the other side.

Tests: SVGAnimationDOMInjectorTest asserts both new selectors; four new cases in chart-inline-svg.directive.spec.ts cover group activation, the outlier->box direction, cleanup on hover end, and the no-data-group no-op.

### DIFF
--- a/utils/inetsoft-xml-formats/src/main/java/inetsoft/util/graphics/animation/SVGAnimationDOMInjector.java
+++ b/utils/inetsoft-xml-formats/src/main/java/inetsoft/util/graphics/animation/SVGAnimationDOMInjector.java
@@ -3431,8 +3431,12 @@ public class SVGAnimationDOMInjector {
          // must dim the other's non-active elements — a same-class-only rule left the outliers at
          // full opacity while their boxes faded. The hovered box's OWN outliers (and an outlier's
          // own box) are kept undimmed by ChartInlineSvgDirective, which activates them via the
-         // shared data-group key. These selectors can only match in a document containing both
-         // classes, i.e. a box plot; a point-only chart has no .inetsoft-box to dim.
+         // shared data-group key. Correlating by class alone is safe: .inetsoft-point is generic
+         // (every PointVO gets it), but Box Plot is a merged graph type (GraphTypes.isMergedGraphType),
+         // so GraphTypes.supportsMultiStyles() excludes it and ChangeChartTypeProcessor.process()
+         // forces setMultiStyles(false) whenever it is selected — even per-measure. A box and a plain
+         // (non-outlier) point therefore never appear in the same graph, and a point-only chart has
+         // no .inetsoft-box to dim.
          "svg.ready:has(.inetsoft-box.inetsoft-active) .inetsoft-box:not(.inetsoft-active)," +
          "svg.ready:has(.inetsoft-box.inetsoft-active) .inetsoft-point:not(.inetsoft-active)," +
          "svg.ready:has(.inetsoft-point.inetsoft-active) .inetsoft-box:not(.inetsoft-active)" +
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (changes requested):
Claude's automated review flagged a medium-confidence design gap worth verifying before merge: the new cross-class hover-dim CSS selectors in `SVGAnimationDOMInjector.java` correlate purely by CSS class (`.inetsoft-box`/`.inetsoft-point`), not by the `data-group` relationship. Since `.inetsoft-point` is applied to all `PointVO` elements (not just box-plot outliers), if a multi-style/combo chart can mix a Box Plot measure with a plain Point measure in the same graph, hovering one could incorrectly cross-dim unrelated elements. Please confirm whether that combination is reachable (see `GraphTypeUtil.isMergedGraphType`) and either tighten the fix/comment accordingly, or note why it's unreachable.

milotalon:
@jshobe-inetsoft Verified — the mixed Box Plot + plain Point scenario is not reachable, so no functional change was needed. Details:

1. `GraphTypes.isMergedGraphType(int)` includes `CHART_BOXPLOT` (`GraphTypes.java:598`), and `supportsMultiStyles(type)` is defined as `!isMergedGraphType(type) && !isScatteredContour(type)` (`GraphTypes.java:608`) — so Box Plot never supports multi-styles.
2. This is enforced server-side, not just in the UI: `ChangeChartTypeProcessor.process()` ends with `if(!GraphTypes.supportsMultiStyles(newType)) { info.setSeparatedGraph(true); info.setMultiStyles(false); }` (`ChangeChartTypeProcessor.java:137-140`). That runs for the **per-measure** path too — `ChangeChartTypeService.java:169` calls `.process()` with the resolved `VSChartAggregateRef` — so selecting Box Plot for one measure of a multi-style chart collapses the whole chart to single-style Box Plot rather than mixing types.
3. The scripting path cannot mix either: `ChartVSAssemblyInfo.setChartStyle()` (`ChartVSAssemblyInfo.java:2344-2360`) applies the same type to every measure ref when `isMultiStyles()` is true.

Claude's premise was correct that `.inetsoft-point` is generic (every `PointVO` gets it), but the mixing precondition it depends on does not exist, so the `:has()` rules cannot cross-dim unrelated elements.

I tightened the comment above the selectors to record that reasoning instead of the weaker "can only match in a document containing both classes" claim, and added the suggested one-line comment at the `activateBoxGroup(el)` call site noting it self-guards on `data-group`. Both changes are comment-only — pushed in 91ef5c0c1.

jshobe-inetsoft (approved):
Approving based on the automated Claude code review chain.

The earlier review raised a valid question about whether Box Plot and plain Point chart types could coexist as separate measures in the same multi-style/combo chart, which would have caused the new cross-class hover-dim CSS selectors to over-fire. The follow-up commit traced `GraphTypes.supportsMultiStyles()`/`isMergedGraphType()` and `ChangeChartTypeProcessor.process()` and confirmed that combination is unreachable — Box Plot always forces `multiStyles` off for the whole `ChartInfo`, so a box and an unrelated plain point can never share a graph. The code now documents this reasoning in both the Java and TS comments. Final verdict: "No further code changes needed."

All checks (build, claude-review, Semgrep, Semgrep OSS, license/cla) are green and there are no merge conflicts.

PR #4614, opened 2026-08-14
Bug #76091, fix brushed chart drawing raw values instead of change-from-previous

ngovus-inetsoft:
## Issue

Import `stackvalue1.vso`, open it in the composer, and select "0" on chart1. The dimmed all-data area on the other charts jumps to span the full measure range instead of staying within the change-from-previous range.

## Root cause

`BrushDataSet` clones each calc column for the all-data element and rewrites its header/field to `__all__…`. Those columns exist **only on the `BrushDataSet`** — never on the datasets below it.

For a `PART_DATE_GROUP` inner dimension (the affected charts bind `MonthOfYear(Order:Date)` and `DayOfWeek(Order:Date)`), the previous-value lookup in `ValueOfColumn.getDataSetDynamicCalcValue()` unwrapped the filter chain with `getRootDataSet()` so cross-facet rows could be reached (added by #74367). That walk skips past the `BrushDataSet`, so `__all__Sum(...)` resolved to column index −1 and the lookup returned `null` for every row.

In `ChangeColumn.getValue()` a null previous value falls through to `cd = NULL` (0), so the result is `d − 0 = d` — the raw aggregate. Hence the all-data area suddenly spanning 0 → full measure range.

## Fix

Replace the `getRootDataSet()` call with `getLookupRoot()`, which unwraps toward the root but stops at the last dataset that still contains the measure column:

```java
if(base == null || base.indexOfHeader(field) < 0) {
   break;
}
```

For brushed all-data columns the walk halts on the `BrushDataSet`. Every other case reaches the same root as before, so the cross-facet behavior from #74367 / #74542 / #74582 is unchanged.

## Tests

- `ValueOfColumnTest#testPreviousOnBrushAllDataColumnFindsValueOnBrushDataSet` — the lookup on `__all__id` over a `MONTH_OF_YEAR_DATE_GROUP` inner dim returns the all-data previous value (10); the first month returns `INVALID`.
- `ChangeColumnTest#testChangePreviousOnBrushAllDataColumnReturnsChangeNotRawValue` — end-to-end symptom: change is 10 and 20, not the raw 20 and 40.

Both fail without the fix (`expected: <10> but was: <null>` and `expected: <10.0> but was: <20.0>`) and pass with it. All 63 tests in `inetsoft.report.composition.graph.calc` pass.

## Known adjacent limitation (not addressed)

For `NEXT` rather than `PREVIOUS`, the lookup takes index `0` of the matched subset, which in the brush row layout is a brow row where `__all__` columns are null. Fixing that requires brush-section awareness in the row selection — wider than this defect.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/calc/ValueOfColumn.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/calc/ValueOfColumn.java
@@ -472,7 +472,7 @@ public class ValueOfColumn extends AbstractColumn {
          DataSet lookupData = data;
 
          if(lookupData instanceof DataSetFilter && (ndimIsPartDate || !ndim.equals(innerDim))) {
-            lookupData = ((DataSetFilter) lookupData).getRootDataSet();
+            lookupData = getLookupRoot(lookupData);
          }
 
          DataSet sub = getSubDataSet(lookupData, cond);
@@ -484,6 +484,31 @@ public class ValueOfColumn extends AbstractColumn {
 
          return sub.getData(field, (ctype == ValueOfCalc.PREVIOUS) ? rcnt - 1 : 0);
       }
+   }
+
+   /**
+    * Unwrap the filter chain toward the root so rows belonging to another facet can be found,
+    * but stop at the last dataset that still contains the measure column being looked up.
+    *
+    * In a brushed chart the all-data columns (__all__xxx) are created by BrushDataSet and do
+    * not exist on the datasets below it. Unwrapping all the way to the root would make
+    * sub.getData(field, ...) return null for every row, which silently turns a
+    * change-from-previous value into the raw aggregate value.
+    */
+   private DataSet getLookupRoot(DataSet data) {
+      DataSet lookup = data;
+
+      while(lookup instanceof DataSetFilter) {
+         DataSet base = ((DataSetFilter) lookup).getDataSet();
+
+         if(base == null || base.indexOfHeader(field) < 0) {
+            break;
+         }
+
+         lookup = base;
+      }
+
+      return lookup;
    }
 
    // Extract the base column name from a date function expression, e.g. "MonthOfYear(Date)" -> "Date".
--- a/core/src/test/java/inetsoft/report/composition/graph/calc/ChangeColumnTest.java
+++ b/core/src/test/java/inetsoft/report/composition/graph/calc/ChangeColumnTest.java
@@ -25,8 +25,10 @@ import inetsoft.report.filter.CrossFilter;
 import inetsoft.report.filter.CrossTabFilter;
 import inetsoft.report.lens.DefaultTableLens;
 import inetsoft.test.*;
+import inetsoft.uql.XConstants;
 import inetsoft.uql.viewsheet.VSDataRef;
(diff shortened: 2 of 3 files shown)


jshobe-inetsoft (approved):
### Code Review: Bug #76091 — brushed chart drawing raw values instead of change-from-previous

**Note:** the automated `claude-review` GitHub Action crashed on this PR (0s error, no review posted). This is a manual review filling that gap.

#### Root cause verification
Traced the call chain in `ValueOfColumn.getDataSetDynamicCalcValue()` (`core/src/main/java/inetsoft/report/composition/graph/calc/ValueOfColumn.java:472-486`). For a `PART_DATE_GROUP` inner dimension, the sub-dataset lookup previously called `((DataSetFilter) lookupData).getRootDataSet()`, unconditionally unwrapping to the true root. `BrushDataSet` is the only dataset in the chain that defines the `__all__`-prefixed calc columns (confirmed via `BrushDataSet.ALL_HEADER_PREFIX`), so unwrapping past it makes `sub.getData(field, ...)` return `null` — and `ChangeColumn.getValue()` treats a null previous value as 0, producing the raw aggregate instead of the delta. This matches the described symptom exactly.

#### Fix assessment
The new `getLookupRoot()` helper (lines 489-509) walks the filter chain one level at a time via `getDataSet()` (not `getRootDataSet()`), stopping as soon as the *next* level up would lose the `field` column (`indexOfHeader(field) < 0`). This is a minimal, correctly-scoped change:
- Only the sub-dataset lookup call site is touched; the earlier `routerData` unwrap (line ~437, used for date-router previous/next navigation) is untouched, so the cross-facet router behavior from #74367/#74542/#74582 is unaffected.
- For any non-brush case, every dataset in the chain retains the aggregate column, so the walk still reaches the same root as before — behavior is unchanged there.
- The loop correctly returns the *last* dataset that has `field`, not the first one that lacks it, avoiding an off-by-one.

#### Tests
Both new regression tests fail without the fix (`null` / raw value) and pass with it, per the PR description. Test setup (mocked `VSDimensionRef`/`XDimensionRef`, `BrushDataSet` over two `VSDataSet`s) accurately models the brushed all-data scenario and directly exercises the previously-broken code path. No existing tests appear broken by this change (only a new private helper was added; the modified call site's behavior is unchanged for all previously-covered cases per the author's note that all 63 tests in the package pass).

#### Other notes
- No security concerns — purely internal data-lookup logic, no external input.
- No backward-incompatible API changes — `getLookupRoot` is a new private method.
- The author calls out a known adjacent limitation for `NEXT` (vs. `PREVIOUS`) direction that isn't fixed here; this is reasonably scoped out as a separate, wider fix and doesn't block this change.
- Minor nit (non-blocking): the javadoc on `getLookupRoot` says "so rows belonging to another facet can be found," which describes the *caller's* reason for unwrapping rather than this method's own behavior (stopping early to preserve the measure column); slightly mixes the "why unwrap at all" rationale with "why stop early" — not confusing enough to require a change.

**Verdict:** No blocking issues. Fix is correct, narrowly scoped, and well-tested. All non-`claude-review` checks pass and there are no merge conflicts. Approving.

PR #4615, opened 2026-08-15
Fix Data Sources/Physical Model editor bugs found during manual testing

swaker854:
## Summary
Five independent, small bug fixes found during manual testing of the Data Sources and Physical/Logical Model editors. Stacked on #4600.

- **modal-header**: fall back to a dialog's own `helpURL` when no `cshid` is bound, instead of always rendering the `cshid` help-link icon. Fixes the Add Expression dialog's help icon opening an empty anchor (falls back to the docs homepage) instead of the script-help docs.
- **script-pane**: collapse the duplicate insertion caused by tree selection firing on both mousedowns of a native double-click, which previously inserted the clicked field/function twice into the expression editor.
- **data-auto-drill-dialog**: show a confirmation before deleting a single Auto Drill parameter, matching the existing confirm-before-delete pattern already used for removing a drill or all parameters.
- **data-sources-tree-view**: fix post-delete/rename/move navigation landing on the Data Sources root instead of the Data Model listing, caused by a scope mismatch between the "Data Model" node (server-assigned scope) and its folder/model children (hardcoded `GLOBAL_SCOPE`).
- **XPartition**: replace the single-arbitrary-anchor unjoined-table check with real connected-component grouping, so the reported "unjoined tables" warning always names an actual disconnected table instead of whichever table happened to be enumerated first.

## Test plan
- [x] Add Expression help icon opens the script-help doc URL instead of an empty anchor (verified live)
- [x] Double-clicking a column/function tree item inserts it once; two separate slower clicks still insert twice (verified live)
- [x] Deleting a single Auto Drill parameter now prompts for confirmation; Cancel preserves it, OK deletes it (verified live)
- [x] Self-deleting the currently-open Data Model folder now lands on the Data Model listing instead of the Data Sources root (verified live)
- [x] Removing a join to isolate a table now reports that table by name in the cross-join warning instead of an unrelated one (verified live)
- [x] `core` module compiles cleanly

### DIFF
--- a/core/src/main/java/inetsoft/uql/erm/XPartition.java
+++ b/core/src/main/java/inetsoft/uql/erm/XPartition.java
@@ -1420,8 +1420,13 @@ public class XPartition implements Cloneable, Serializable, XMLSerializable, XML
    }
 
    /**
-    * Checks for any unjoined tables in the view. Returns the first pair of
-    * unjoined tables that are found.
+    * Checks for any unjoined tables in the view. Groups all tables into
+    * connected components and, if more than one exists, returns a
+    * representative of the largest (main) component paired with a
+    * representative of the smallest component. This ensures the reported
+    * table names always include an actual disconnected table, rather than
+    * an arbitrary table (which may itself belong to a minor component)
+    * used as a fixed comparison anchor.
     *
     * @return the names of the unjoined tables or <code>null</code> if no
     *         unjoined tables are found.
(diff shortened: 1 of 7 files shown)


swaker854:
merge after 4600 merge

jshobe-inetsoft (approved):
## Code Review — PR #4615

Reviewed all five described fixes plus one undescribed change (see note below) against the current diff and surrounding code.

### `XPartition.getUnjoinedTables()` (core)
Correct improvement: replaces the old "always compare against `tables[0]`" anchor with real connected-component grouping (each table joins the first existing component it has a join path to, via the component's representative; representative-based matching is valid here since join-path connectivity is transitive). If more than one component exists, it now always names two tables that are genuinely disconnected from each other — fixing the reported bug where the arbitrary anchor itself could be the outlier, producing a misleading pair.

One doc/impl mismatch (non-blocking): the updated javadoc says the result pairs the largest component with "a representative of the smallest component," but the code sorts descending and takes indices `0` and `1` — i.e., the *second-largest*, not the smallest, whenever there are 3+ disconnected components. The returned pair is still a valid disconnected pair (correct behavior), just not always the literal smallest for 3+-way splits. Worth tightening the comment or the logic, but not merge-blocking.

No new test coverage was added for this rewritten algorithm (none appears to exist for it currently either, so nothing is broken — just a coverage gap worth flagging).

### `data-auto-drill-dialog.component.ts` — confirm before deleting a parameter
Matches the existing `deleteDrill()`/`removeAllParameters()` pattern in the same file exactly (same `ComponentTool.showConfirmDialog` call shape, same no-op rejection handler). Both bundle keys (`data.logicalmodel.drillDeleteParameter`, pre-existing; `data.logicalmodel.confirmRemoveParam`, added here) resolve correctly.

### `data-sources-tree-view.component.ts` — post-delete/rename/move navigation
The `endsWith("/Data Model")` check is consistent with how this exact literal path segment is already constructed elsewhere in the same file (`getParentPath0(...) + "/Data Model"`, `databasePath + "/Data Model"`). It's an internal path identifier, not a translated label, so this isn't fragile against localization. Scoped narrowly to only bypass the scope check for this one specific parent lookup — no broader effect on other navigation paths.

### `script-pane.component.ts` — duplicate insertion on double-click
Sliding 500ms same-node debounce correctly suppresses the second `mousedown`-triggered `itemClicked()` call without touching the rest of the method's logic (focus/insert). Matches the documented trade-off in the PR's test plan (fast double-click → single insert; deliberate slower double-click → two inserts), which is a reasonable UX choice for this kind of guard.

### `modal-header.component.html` — help icon fallback
Verified `cshid`/`helpURL` `@Input()`s and `openHelpURL()` already exist on the component, and `showHelpLink` already requires at least one of `cshid`/`helpURL` to be set — so the new `@else` branch (shown only when `cshid` is falsy) can only render when `helpURL` is actually set, meaning the click handler is never a no-op. Correct.

### Undescribed 6th change: `theme-css-view.component.ts`
This PR also adds 5 new CSS variable names (`--inet-feedback-font-size`, `--inet-badge-font-size`, `--inet-ui-neutral-color`, `--inet-ui-neutral-border-color`, `--inet-ui-neutral-hover-color`) to the EM theme editor's variable list. It's a simple additive change with no apparent risk, but it isn't mentioned anywhere in the PR description's "five independent bug fixes" summary or test plan — worth calling out for completeness, not a blocker.

### Overall
No security concerns (no new external input handling), no backward-incompatibility (all changes are additive or narrowly scoped to previously-buggy paths), and no existing tests appear to be broken by any of these changes. All checks pass (build, Semgrep ×2, CLA, claude-review) and the PR is cleanly mergeable.

**No blocking issues found.**

PR #4617, opened 2026-08-16
Viewsheet-level properties for the wiz agent API

larryliang-inetsoft:
Adds read/write access to the viewsheet's **own** properties — the settings behind the Composer's Viewsheet Property dialog. Until now every property endpoint was assembly-keyed, so none of this was reachable: description, display alias, max rows, snap grid, metadata mode, MV, selection association.

Paired with plugin-side tools in stylebi-wiz (`list`/`get`/`set_viewsheet_properties`).

### Why a sibling service rather than a parameter

`AssemblyPropertyService` reflects one signature onto every dialog service — `(runtimeId, assemblyName, model, linkUri, user, dispatcher)`. The viewsheet's own service does not have it: `getViewsheetInfo(runtimeId, principal)` and `setViewsheetInfo(runtimeId, value, principal, dispatcher, linkUri, refLayoutName)` take no assembly name, order their arguments differently, and add `refLayoutName`.

Threading a nullable `assemblyName` through the existing service would make every existing call site read as though it might be sheet-level. `SheetPropertyService` is a sibling instead, sharing the same whole-patch validation, nearest-match refusal on unknown keys, and forgiving-but-not-guessing scalar coercion.

### A latent bug this surfaced

`PropertyPath.set` returned `void`. That is correct for every case that existed — each target sat under a mutable holder that absorbed an Immutables wither's rebuild. But `width`/`height`/`preview` sit **directly on the immutable root**, so the rebuilt instance had nowhere to go and the write vanished with a clean return.

It now returns the (possibly rebuilt) root. `AssemblyPropertyService` was hardened the same way — no behaviour change today, since none of its aliases target a top-level field of an Immutables model, but the next one that does would silently no-op.

### One deliberate refusal

`vsScriptPane` is readable and refuses writes, pointing the caller at the script tools. It carries the sheet's `onInit`/`onLoad` script, and writing it through a *properties* patch would be a second, ungoverned path to authoring script — while the taxonomy meant to govern that is still unbuilt. The refusal checks the **resolved path**, not the input key, so a raw dotted path cannot reach it under another spelling.

`screensPane` is excluded entirely: renaming or rearranging a layout is not a property edit.

`alias` **is** writable. An earlier draft refused it; that was wrong — `set_worksheet_properties` already ships exactly that capability for the sibling sheet type, and `alias` is an explicitly named key, so there is no side effect to guard against.

### Two existing tests changed

`everyCoveredTypeHasAWiredService` assumed every covered type is assembly-wired, and `noAliasIsAnEmptyOrSelfReferentialPath` assumed every alias nests. The new `viewsheet` type breaks both by design. Each got an explicit, documented carve-out rather than being loosened generally.

### Tests

`./mvnw -pl core -Dtest="inetsoft.web.wiz.viewsheet.*Test" test` — the whole package, not just the new files, to catch collateral damage: **277 tests, 0 failures/errors**. Written test-first; each new test was confirmed failing before the code that satisfies it.

The full module suite was not run — impractical here. Stated rather than implied.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyPropertyService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyPropertyService.java
@@ -230,7 +230,11 @@ public class AssemblyPropertyService {
          Map<String, String> resolved = new LinkedHashMap<>();
 
          for(String key : patch.keySet()) {
-            resolved.put(key, PropertyAliases.resolve(type, key));
+            // resolveForWrite, not resolve: its refusals are keyed by type and currently fire
+            // only for the sheet, so this is a no-op for every assembly today. Going through the
+            // write path anyway means a refusal added for an assembly type -- a script pane on
+            // some future assembly, say -- applies here without anyone remembering to route it.
+            resolved.put(key, PropertyAliases.resolveForWrite(type, key));
          }
 
          // PropertyPath.set returns the root the caller must keep using: if a future alias
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyPropertyService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyPropertyService.java
@@ -233,8 +233,12 @@ public class AssemblyPropertyService {
             resolved.put(key, PropertyAliases.resolve(type, key));
          }
 
+         // PropertyPath.set returns the root the caller must keep using: if a future alias
+         // ever targets a direct top-level field of an Immutables dialog model (no nested pane
+         // to absorb the rebuild — see PropertyPath's own note), the wither produces a new
+         // instance and the original reference silently stops reflecting the write.
          for(Map.Entry<String, String> entry : resolved.entrySet()) {
-            PropertyPath.set(model, entry.getValue(), patch.get(entry.getKey()));
+            model = PropertyPath.set(model, entry.getValue(), patch.get(entry.getKey()));
          }
 
          writeModel(runtimeId, type, assemblyName, model, linkUri, user, dispatcher);
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyPath.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyPath.java
@@ -54,18 +54,30 @@ public final class PropertyPath {
    }
 
    /**
-    * Writes a dotted path.
+    * Writes a dotted path, and returns the root the caller must keep using afterward.
     *
     * <p>An absent intermediate is an error rather than something to instantiate: the dialog
     * models are populated by the composer service, so a null pane means the property does not
     * apply to this assembly, and filling one in would fabricate a shape the service never
     * produced.
+    *
+    * <p><b>The return value matters whenever the root itself is immutable.</b> Every other
+    * Immutables case this class handles nests the immutable level under a mutable holder, so a
+    * rebuilt child always has somewhere mutable to be written back into and {@code root} itself
+    * never changes identity. But a single-segment path directly on an immutable root — the
+    * viewsheet-level dialog model's own {@code width}/{@code height}/{@code preview} — has no
+    * such holder: the wither produces a genuinely new instance, and nothing above the root
+    * exists to absorb it. Returning {@code root} unconditionally would silently drop that write,
+    * which is exactly the defect this class exists to avoid. A caller that ignores the return
+    * value and keeps the original reference reproduces that defect regardless.
     */
-   public static void set(Object root, String path, Object value) {
+   public static Object set(Object root, String path, Object value) {
       REBUILT_CHILD.remove();
 
       try {
          writeInto(root, segments(path), 0, path, value);
+         Object rebuilt = REBUILT_CHILD.get();
+         return rebuilt != null ? rebuilt : root;
       }
       finally {
          REBUILT_CHILD.remove();
--- a/core/src/main/java/inetsoft/web/composer/vs/dialog/ViewsheetPropertyDialogService.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/dialog/ViewsheetPropertyDialogService.java
@@ -455,11 +455,15 @@ public class ViewsheetPropertyDialogService {
 
       LocalizationPaneModel localizationPane = value.localizationPane();
 
-      for(String component: info.getLocalComponents()) {
-         info.removeLocalID(component, true);
-      }
-
+      // A null pane means "not supplied", not "clear everything". getViewsheetInfo populates it
+      // only for a site admin (or when security is off / single-tenant), so on any other principal
+      // a plain read-modify-write round trip arrived here with null -- and wiping first removed
+      // every localized text id on the sheet as a side effect of setting something unrelated.
       if(localizationPane != null) {
+         for(String component: info.getLocalComponents()) {
+            info.removeLocalID(component, true);
+         }
+
          for(LocalizationComponent component: localizationPane.getLocalized()) {
             info.setLocalID(component.getName(), component.getTextId());
          }
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/ChartRegionPropertyService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/ChartRegionPropertyService.java
@@ -128,20 +128,30 @@ public class ChartRegionPropertyService {
       }
 
       // Write onto the model only after every key resolved.
+      //
+      // Keep the returned root. All three region dialog models are plain mutable classes today, so
+      // PropertyPath.set mutates in place and the assignment is a no-op — but if any of them (or a
+      // pane beneath one) becomes an Immutables model, a wither rebuilds rather than mutates and
+      // the write would vanish with no error and no compile failure. That is the same shape that
+      // made width/height/preview silently unwritable on the viewsheet's own dialog model.
+      Object rebuilt = model;
+
       for(Map.Entry<String, Object> one : resolved.entrySet()) {
-         PropertyPath.set(model, one.getKey(), one.getValue());
+         rebuilt = PropertyPath.set(rebuilt, one.getKey(), one.getValue());
       }
+
+      final Object written = rebuilt;
 
       sessions.mutate(sessionToken, user, (rvs, runtimeId, dispatcher) -> {
          switch(name) {
          case "axis" -> regions.setAxisPropertyDialogModel(
-            runtimeId, assembly, key, 0, field, (AxisPropertyDialogModel) model, linkUri, user,
+            runtimeId, assembly, key, 0, field, (AxisPropertyDialogModel) written, linkUri, user,
             dispatcher);
          case "legend" -> regions.setLegendFormatDialogModel(
-            runtimeId, assembly, indexOf(key), (LegendFormatDialogModel) model, linkUri, user,
+            runtimeId, assembly, indexOf(key), (LegendFormatDialogModel) written, linkUri, user,
             dispatcher);
          default -> regions.setTitleFormatDialogModel(
-            runtimeId, assembly, key, (TitleFormatDialogModel) model, linkUri, user, dispatcher);
+            runtimeId, assembly, key, (TitleFormatDialogModel) written, linkUri, user, dispatcher);
          }
       });
    }
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyAliases.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyAliases.java
@@ -602,11 +602,14 @@ public final class PropertyAliases {
       aliases.put("maxRowsWarning", "vsOptionsPane.maxRowsWarning");
       aliases.put("hideNotifications", "vsOptionsPane.hideNotifications");
       aliases.put("listOnPortalTree", "vsOptionsPane.listOnPortalTree");
-      // Read-only in v1 -- see resolveForWrite. Both are genuine top-level accessors, hence the
-      // bare (dot-free) path; noAliasIsAnEmptyOrSelfReferentialPath carves out this type for
-      // exactly that reason.
-      aliases.put("filtersPane", "filtersPane");
-      aliases.put("localizationPane", "localizationPane");
+      // filtersPane and localizationPane are deliberately NOT aliased.
+      //
+      // They are read-only, and they are whole object graphs rather than properties: aliasing them
+      // made every list/get response carry the entire localization component tree -- ~350 lines on
+      // a small sheet, and it grows with assembly count. That is a curated vocabulary paying a
+      // large cost for something it cannot even write. Reading them is still possible, and is what
+      // `raw: true` is for; resolveForWrite still refuses them by name, since it matches on the
+      // path rather than on membership in this map.
       aliases.put("width", "width");
       aliases.put("height", "height");
       aliases.put("preview", "preview");
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyAliases.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyAliases.java
@@ -43,6 +43,9 @@ import java.util.*;
  * dropped.
  */
 public final class PropertyAliases {
+   /** The viewsheet's own property vocabulary. Not "viewsheet" -- see the register call below. */
+   public static final String SHEET = "sheet";
+
    /** One assembly type's dialog model and its alias vocabulary. */
    public record TypeAliases(String assemblyType, Class<?> modelClass,
                              Map<String, String> aliases) {}
@@ -140,7 +143,7 @@ public final class PropertyAliases {
     * alias that happens to map onto it).
     */
    private static String viewsheetWriteRefusal(String normalizedType, String pathOrKey) {
-      if(!"viewsheet".equals(normalizedType) || pathOrKey == null) {
+      if(!SHEET.equals(normalizedType) || pathOrKey == null) {
          return null;
       }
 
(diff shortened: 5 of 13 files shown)


core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyPath.java
```diff
@@ -54,18 +54,30 @@ public static Object get(Object root, String path) {
...
+    * rebuilt child always has somewhere mutable to be written back into and {@code root} itself
+    * never changes identity. But a single-segment path directly on an immutable root — the
+    * viewsheet-level dialog model's own {@code width}/{@code height}/{@code preview} — has no
+    * such holder: the wither produces a genuinely new instance, and nothing above the root
+    * exists to absorb it. Returning {@code root} unconditionally would silently drop that write,
+    * which is exactly the defect this class exists to avoid. A caller that ignores the return
+    * value and keeps the original reference reproduces that defect regardless.
     */
-   public static void set(Object root, String path, Object value) {
+   public static Object set(Object root, String path, Object value) {
```
larryliang-inetsoft, line 74: Traced the contract and it holds. Every non-leaf frame in `writeInto` that consumes a child's rebuild calls `REBUILT_CHILD.remove()` before writing into its own owner (the early return at the `after != before` check, and again just before the setter/wither dispatch), so at depth 0 the thread-local is non-null **only** when the root itself took the wither branch. The three cases all land correctly:

- mutable root, deep write → depth-0 frame returns via `setter.invoke` without setting the thread-local → returns `root`
- immutable root, single-segment path → leaf at depth 0, wither sets it → returns the rebuilt instance
- immutable root, multi-segment → depth 0 consumes the child's rebuild, then its own wither sets it → returns the rebuilt root

`rebuildingAnImmutableRootReturnsTheNewInstance` and `writingAMutableRootReturnsThatSameRoot` pin both directions, and `theOriginalImmutableRootIsLeftUnchanged` pins the reason the return value is load-bearing rather than cosmetic. No issue here.

core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyPropertyService.java
```diff
@@ -233,8 +233,12 @@ public void set(String sessionToken, Principal user, String assemblyName,
             resolved.put(key, PropertyAliases.resolve(type, key));
          }
 
+         // PropertyPath.set returns the root the caller must keep using: if a future alias
+         // ever targets a direct top-level field of an Immutables dialog model (no nested pane
+         // to absorb the rebuild — see PropertyPath's own note), the wither produces a new
+         // instance and the original reference silently stops reflecting the write.
          for(Map.Entry<String, String> entry : resolved.entrySet()) {
-            PropertyPath.set(model, entry.getValue(), patch.get(entry.getKey()));
+            model = PropertyPath.set(model, entry.getValue(), patch.get(entry.getKey()));
```
larryliang-inetsoft, line 245: Confirmed the hardening is safe, and the one sibling that was **not** hardened is safe too — but not by a one-liner, which is worth recording here.

`ChartRegionPropertyService` still calls `PropertyPath.set(model, ...)` and discards the return. I checked all three of its dialog models rather than assuming:

- `AxisPropertyDialogModel:35` — `public class ... implements Serializable`
- `LegendFormatDialogModel:32` — same
- `TitleFormatDialogModel:27` — same

Each has a public setter for every top-level field, so `writeInto`'s depth-0 frame always takes the setter branch and `REBUILT_CHILD` is never left set at the root. Safe today.

The wrinkle: it **cannot** be hardened the way this line was. In `ChartRegionPropertyService.set`, `model` is captured by the `sessions.mutate(...)` lambda that follows the write loop, so it has to stay effectively final — `model = PropertyPath.set(model, ...)` will not compile. It needs an intermediate:

```java
Object written = model;

for(Map.Entry<String, Object> one : resolved.entrySet()) {
   written = PropertyPath.set(written, one.getKey(), one.getValue());
}

final Object toWrite = written;
```

Worth a line in the PR description either way, so the next person who notices the asymmetry does not try the one-liner, hit the effectively-final error, and conclude the hardening was not wanted there.

larryliang-inetsoft (commented):
Reviewed together with the plugin half, stylebi-wiz#1616.

The design call — a sibling `SheetPropertyService` rather than a nullable `assemblyName` threaded through `AssemblyPropertyService` — is the right one, and for the reason stated: `getViewsheetInfo(runtimeId, principal)` / `setViewsheetInfo(runtimeId, value, principal, dispatcher, linkUri, refLayoutName)` genuinely does not fit the reflected `(runtimeId, objectId, ...)` shape, so forcing it would have made every existing call site read as though it might be sheet-level.

The two test carve-outs (`everyCoveredAssemblyTypeHasAWiredService`, `noAliasIsAnEmptyOrSelfReferentialPath`) are each documented and each paired with a positive assertion of the thing being excluded — `viewsheetIsNotWiredIntoTheAssemblyDispatchTable` is the right complement to the first. Loosening either test generally would have been the easy wrong move.

The `vsScriptPane` refusal checking the **resolved path** rather than the input key is the detail that makes it hold, and `refusesTheScriptPaneEvenAsARawDottedPath` pins it.

Two notes inline. No blocking findings — this looks good to merge.

One minor observation with no line to hang it on: `viewsheetWriteRefusal` only fires when the normalized type is `viewsheet`, and `resolveForWrite` is the only caller — `AssemblyPropertyService` still goes through plain `resolve`. That is correct today, since no assembly type has a script-pane alias, but it means the refusal is a viewsheet-only guard by construction rather than a general one. Worth knowing if an assembly ever grows a script alias.

larryliang-inetsoft (commented):
## Review — viewsheet-level properties

Reviewed against `feature-76074-server-fixes`, reading the diff plus `ViewsheetPropertyDialogService`, `PropertyPath`, and `AssemblyPropertyService`. The alias/refusal design is sound; the problems below are all in what the vocabulary exposes and in the read/modify/write round-trip through `getViewsheetInfo` → `setViewsheetInfo`.

The root cause of the first three is shared: `set` patches one key but writes the **whole** dialog model back. So a refusal at the alias layer protects nothing if the underlying field is still round-tripped, and every asymmetry between the getter and the setter becomes a side effect of setting an unrelated property.

### Blocking

**1. Any `set_viewsheet_properties` call wipes sheet localization for a non-site-admin.** `ViewsheetPropertyDialogService.java:229` populates `localizationPane` only under `isSiteAdmin(principal) || !isSecurityEnabled() || !isMultiTenant()`; otherwise it stays null. `setViewsheetInfo:458` then runs

```java
for(String component: info.getLocalComponents()) {
   info.removeLocalID(component, true);
}

if(localizationPane != null) { ... re-add ... }
```

unconditionally, *before* the null check. In a secured multi-tenant deployment an org (non-site) admin calling `set_viewsheet_properties {"desc": "x"}` removes every localized text id on the sheet. Refusing `localizationPane` on write does not help — the null pane is still written back.

**2. `width` / `height` / `preview` are not viewsheet state, and are writable.** `getViewsheetInfo` never populates them, so they read back as the Immutables defaults `0/0/false`; `setViewsheetInfo:481` reads them only to size a one-off `refreshViewsheet`:

```java
int width = value.preview() && info.isScaleToScreen() ? value.width() : 0;
```

`set_viewsheet_properties {"width": 800}` reports success, and the next `get_viewsheet_properties` returns `width: 0` — precisely the silent no-op this layer's javadoc says it exists to prevent. Worse, `preview: true` reaches `VSEventUtil.clearScale(rvs.getViewsheet())` at line 485, discarding assembly scaling as a side effect of "setting a property".

Worth noting that the `PropertyPath.set` return-value change — the load-bearing part of this PR's plumbing — exists only to make these three writable. Dropping them from the vocabulary removes the need for it (though the hardening is still worth keeping; see 6).

### Non-blocking

**3. Custom print page size is corrupted by an unrelated `set`.** The getter reads custom dimensions in inches (`ViewsheetPropertyDialogService:221-222`, `getDisplayPageSize(size.getWidth(), 1)` — ratio 1) while the setter multiplies by `PrintInfo.getUnitRatio(units)` (line 390). With `units == "mm"` and a custom paper size, one `set_viewsheet_properties {"desc":"x"}` divides the page size by 25.4, and repeat calls compound. Margins convert correctly on *both* sides (line 211 vs 385), which is what makes the asymmetry easy to miss. Excluding `screensPane` from the vocabulary doesn't avoid this, for the same whole-model-write reason as 1.

**4. `onDemandMvEnabled` is a read-only capability flag exposed as writable** (`PropertyAliases.java:599`). It is computed in the getter from `SreeEnv.getProperty("mv.ondemand")` and `SUtil.isSharedDefaultOrgDashboard`; `setViewsheetInfo` never reads it. Setting it reports success and changes nothing — same class as 2. `createMv` is the real property.

**5. The type name `viewsheet` collides with the embedded-viewsheet assembly type,** degrading an error message. `Viewsheet implements VSAssembly`, and `AssemblyPropertyService.typeOf:327` derives its key from the class simple name, so `vs.getAssembly("EmbeddedVS1")` yields `"viewsheet"`. With `PropertyAliases.covers("viewsheet")` now true, the clear `'EmbeddedVS1' is a Viewsheet, whose properties are not covered yet. Covered types: …` is replaced by `No property service wired for assembly type 'viewsheet'` (line 294), which reads as an internal wiring bug rather than an unsupported assembly. `viewsheetIsNotWiredIntoTheAssemblyDispatchTable` locks that message in.

**6. `ChartRegionPropertyService.java:132` still discards `PropertyPath.set`'s return.** Safe today only because `AxisPropertyDialogModel` / `LegendFormatDialogModel` / `TitleFormatDialogModel` and their panes are plain mutable classes. If any becomes an Immutables model, writes vanish silently with no compile error.

### On the tests

The `noAliasIsAnEmptyOrSelfReferentialPath` carve-out is scoped to `type.equals("viewsheet")`, which is narrow in the right dimension, but it skips the dot-check for *every* viewsheet alias rather than the five genuinely top-level ones. A future `desc → desc` typo would pass. Gating on the alias name as well would keep the invariant live for the rest:

```java
if(type.equals("viewsheet") && TOP_LEVEL_FIELDS.contains(alias.getKey())) {
   continue;
}
```

If 2 is resolved by dropping `width`/`height`/`preview`, the carve-out shrinks to `filtersPane`/`localizationPane` — or disappears, if those are dropped too on the grounds that refusing them on write doesn't actually protect them.

larryliang-inetsoft:
All findings addressed in `064a2f402` and `8f0a9443a`. One I'm pushing back on, with a test.

### Blocking

**1. Localization wipe — fixed at the source.** Confirmed exactly as described. The removal loop now runs only when a pane was actually supplied; a null pane means "not supplied", not "clear everything". This also fixes the Composer's own dialog round trip, which had the same shape for a non-site-admin.

**2. `width`/`height`/`preview` — removed from the vocabulary**, along with `onDemandMvEnabled` for the same reason (a capability flag the setter never reads). You're right that they were phantom state; `preview: true` reaching `clearScale` was the worst of it.

### Non-blocking

**5. Type-name collision — fixed by renaming the key** to `sheet`. `PropertyAliases.SHEET` is now the single source, so `covers("viewsheet")` is false again and the "not covered yet" message is restored.

**6. `ChartRegionPropertyService`** now keeps the returned root. No-op today for exactly the reason you gave; it's the silent-vanish shape I'd rather not leave armed.

**The test carve-out** no longer skips the type at all. With the bare top-level aliases gone there is nothing left to except, so the nesting invariant is live for every viewsheet alias — a `desc → desc` typo is caught again. Your suggested `TOP_LEVEL_FIELDS` gate turned out to be unnecessary rather than wrong.

**Your closing observation about `viewsheetWriteRefusal` being viewsheet-only by construction** — `AssemblyPropertyService` now goes through `resolveForWrite` too. Still a no-op for every assembly type, but a refusal added for one applies without anyone remembering to route it.

### 3 — the page-size asymmetry is not there

This is the one I'd ask you to look at again. `PrintInfo.getSize()` does the inches-to-current-unit conversion **itself**:

```java
public DimensionD getSize() {
   double ratio = 1 / getUnitRatio(); // convert inches to current unit
   return new DimensionD(size.getWidth() * ratio, size.getHeight() * ratio);
}
```

which is precisely why the dialog getter passes `1` for the size while passing a real ratio for the margins two lines above. The round trip is `unit → inches` on write and `inches → unit` on read. Symmetric.

I nearly implemented the fix before checking, and doing so would have introduced the bug as described — so the finding is a good one to have raised, just inverted.

`PrintInfoSizeUnitsTest` pins it in mm, points and inches, plus ten read/modify/write cycles that would be off by 25.4¹⁰ if the conversion were one-sided. It passes against unmodified code. I kept it so the next reader doesn't re-derive this, and so that "fixing" the getter to match the margins fails loudly.

Tests: wiz + vslayout + composer dialog packages, 51 tests, green.

larryliang-inetsoft (commented):
## Re-review — 064a2f40

Both blocking findings are fixed, and fixed at the right layer.

**Localization wipe.** The removal loop now sits inside `if(localizationPane != null)`, so a null pane means "not supplied" rather than "clear everything". I checked the case that would have made this a regression: clearing every localized id from the Composer dialog still works, because that arrives as a non-null pane with an empty `getLocalized()` — wipe-then-re-add-nothing. Correct, and the shared-service fix is the right call rather than working around it in the wiz layer.

**Phantom properties.** `width`/`height`/`preview` and `onDemandMvEnabled` are out of the vocabulary, with the reasoning recorded where the next person will look for it.

**Type-name collision.** Keying the vocabulary `sheet` avoids it cleanly. The nesting-invariant carve-out is gone entirely now that no bare alias remains — better than the narrowing I suggested.

I ran the affected suites at this head in a scratch worktree: `PropertyAliasesTest` (21), `PropertyPathTest` (32), `AssemblyPropertyServiceTest` (11), `SheetPropertyServiceTest` (7), `ViewsheetAssemblyAgentControllerTest` (8) — **79 tests, 0 failures, 0 errors**.

### Still open from the last round

**Custom print page size is still corrupted by an unrelated `set`** (finding 3). Unchanged at this head: `getViewsheetInfo:221-222` reads custom dimensions with ratio `1` (inches) while `setViewsheetInfo:390` multiplies by `PrintInfo.getUnitRatio(units)`. On a sheet with a custom paper size and `units == "mm"`, one `set_viewsheet_properties {"desc":"x"}` divides the page size by 25.4 and compounds on repeat calls. Dropping `screensPane` from the vocabulary doesn't help — the whole model is still round-tripped. This is the last surviving instance of the root cause the other two fixes addressed, so it's worth closing the same way (fix the getter to convert with the same ratio as the setter) rather than leaving it as the one asymmetry that still bites.

**`ChartRegionPropertyService:132` still discards `PropertyPath.set`'s return** (finding 6). Latent only, as before.

### New nits, both introduced by this commit

Two comments now argue from premises the commit removed:

- `PropertyAliases.java:271-274` — "ViewsheetPropertyDialogModel has genuine top-level scalar fields (width, height, preview) … so several aliases here are legitimately bare paths rather than dotted ones." No alias in that map is bare any more.
- `SheetPropertyService.java:119-121` — "width/height/preview live directly on this Immutables model … so the reassignment is load-bearing, not defensive boilerplate." With those aliases gone, no current alias needs it. Keeping the reassignment is right (the commit message explains why), but the comment should say it guards the *next* top-level alias rather than naming three that no longer exist.

Also `private static final String VIEWSHEET = PropertyAliases.SHEET;` reads oddly now that the two names deliberately differ — the constant is the one place the distinction matters.

None of these block.

larryliang-inetsoft:
Both new nits fixed in `cf631428b`, and the two "still open" items were already fixed in `8f0a9443a` — which landed after the head you re-reviewed (`064a2f402`), so you wouldn't have seen them. My fault for not saying so when I pushed.

**Both stale comments removed.** You're right that each argued from a premise the same commit deleted:
- `PropertyAliases` no longer claims bare paths are legitimate — no alias in that map is bare any more, which is precisely why the carve-out could go entirely.
- `SheetPropertyService` now says the reassignment is defensive and *why it's still worth keeping*: it guards the next alias targeting a bare top-level field, rather than naming three that no longer exist.

**`VIEWSHEET` renamed to `SHEET_TYPE`.** Agreed — `VIEWSHEET = PropertyAliases.SHEET` invited exactly the confusion the rename exists to prevent, at the one place the distinction has to be stated.

**Finding 6 (`ChartRegionPropertyService`)** — fixed in `8f0a9443a`. It keeps the returned root now, accumulating into a final local since the lambda below captures it.

### Finding 3 — I don't think this one is real, and I have a test

`PrintInfo.getSize()` does the inches-to-current-unit conversion **itself**:

```java
public DimensionD getSize() {
   double ratio = 1 / getUnitRatio(); // convert inches to current unit
   return new DimensionD(size.getWidth() * ratio, size.getHeight() * ratio);
}
```

That is why the dialog getter passes `1` for the size while passing a real ratio for the margins two lines above — the size has already been converted, the margins haven't. The round trip is `unit → inches` on write, `inches → unit` on read.

I nearly implemented your fix before checking, and doing so would have introduced the bug as described — so this was worth raising, just inverted.

`PrintInfoSizeUnitsTest` (added in `8f0a9443a`) pins it in mm, points and inches, plus ten read/modify/write cycles that would be off by 25.4¹⁰ if the conversion were one-sided. It passes against unmodified production code.

If you still read an asymmetry there, I'd rather have a failing test than an argument — a case where the round trip loses precision would settle it immediately, and I'll fix it if one exists.

Tests at this head: `inetsoft.web.wiz.viewsheet.*` 276, plus vslayout and composer-dialog suites, all green.

larryliang-inetsoft (commented):
## Re-review — 8f0a9443, cf631428

### My page-size finding was wrong

You're right, and the correction is well-evidenced. `PrintInfo.getSize()` converts inches to the current unit itself:

```java
public DimensionD getSize() {
   double ratio = 1 / getUnitRatio(); //Convert inches to current unit
   return new DimensionD(size.getWidth() * ratio, size.getHeight() * ratio);
}
```

while `getMargin()` returns the raw fields — which is exactly why the dialog getter passes a real ratio for the margins and `1` for the size. I read the `1` as "this value is in inches" without following `getSize()`, and the neighbouring margin lines made the asymmetry look deliberate. The round trip is symmetric, and implementing my suggested fix would have introduced the bug I described. Good catch, and thanks for pinning it rather than just reverting the claim.

**One small overstatement in `PrintInfoSizeUnitsTest`'s javadoc:** it says the tests exist "so that 'fixing' the getter to match the margins … fails loudly." They wouldn't — the suite exercises `PrintInfo` directly and never touches `ViewsheetPropertyDialogService`, so changing that getter to `getDisplayPageSize(size.getWidth(), ratio)` leaves all five green. The tests are still worth keeping (they document the conversion contract at its source, which is where the misreading started), but they document the invariant rather than guard the call site. Either soften the claim or add one assertion against the dialog service's own round trip.

### The other two

`ChartRegionPropertyService` keeping the returned root, and `AssemblyPropertyService` going through `resolveForWrite`, are both right. I checked the second is genuinely inert today: `viewsheetWriteRefusal` returns null unless the type is `SHEET`, so no assembly path changes behaviour.

`cf631428` clears both stale comments, and the `SHEET_TYPE` rename lands the distinction at the one place it has to be stated.

### Verification

Ran at `cf631428` in a scratch worktree: `PropertyAliasesTest` (21), `PropertyPathTest` (32), `AssemblyPropertyServiceTest` (11), `SheetPropertyServiceTest` (7), `ViewsheetAssemblyAgentControllerTest` (8), `PrintInfoSizeUnitsTest` (5) — **84 tests, 0 failures, 0 errors**.

Nothing outstanding from my side. LGTM.

PR #4618, opened 2026-08-16
list_bindable_fields: report the data type and the source table

larryliang-inetsoft:
Found by live testing of the merged Composer plugin: `list_bindable_fields` returned `dataType: null` and `role: null` for **every** column, across nine groups labelled only "Dimensions" or "Measures" — no way to tell which table a column came from.

All three originate here. Both middle tiers were verified as pure passthroughs: the plugin is a bare `wizClient.get`, and wiz-services' `fields()` is `return data` with no transformation.

### `dataType` — a branch that could never match

```java
node.data() instanceof DataRefModel ref ? ref.getDataType() : null
```

The tree never carries a `DataRefModel`. `VSTreeHandler.createNodeFromEntry` builds **every** node with `.data(entry)` — an `AssetEntry`. So the branch never matched and the type was always null, while the tool's description promised one.

Now read from the entry's `dtype` property, which is where `Viewsheet`, `VSEventUtil` and `JDBCUtil` already read it.

### The group name — discarded one level up

`collect()` named each group after the node directly holding the columns. In the Composer's binding tree that node is a `Dimensions`/`Measures` folder; the table name sits on its parent and was thrown away.

The nearest ancestor whose entry is a `TABLE`, `PHYSICAL_TABLE`, `QUERY` or `LOGIC_MODEL` is now carried down. It falls back to the old behaviour when no such ancestor exists, so a tree shape this does not anticipate degrades rather than producing a blank name.

### `role` — removed

Hardcoded `null` at its only construction site, never populated for any column. A field that is always null reads as *"this column has no role"* rather than *"this API does not report one"*. The binding tree does not distinguish dimension from measure at that point, so there is nothing to populate it from. Nothing consumed it — the plugin's `role` handling is on its **input** side, an alias for `type` in field references.

### The tests were part of the problem

`BindableFieldsServiceTest` mocked a `DataRefModel` into `data()` and asserted a real data type off it. Green suite, while production returned null for every column — the test asserted a tree shape that does not exist.

Rewritten against what `VSTreeHandler` actually builds, plus a test for the live shape: columns under `Dimensions`/`Measures` folders beneath a named table, asserting both groups name the table rather than the folder.

`./mvnw -pl core -Dtest="inetsoft.web.wiz.binding.*Test" test` — 236 tests, 0 failures.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindableFieldsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindableFieldsService.java
@@ -18,6 +18,7 @@
 package inetsoft.web.wiz.binding;
 
 import inetsoft.uql.asset.AssetEntry;
+import inetsoft.uql.schema.XSchema;
 import inetsoft.web.binding.service.VSBindingTreeService;
 import inetsoft.web.composer.model.TreeNodeModel;
 import inetsoft.web.wiz.binding.model.BindableField;
@@ -57,34 +58,64 @@ public class BindableFieldsService {
    }
 
    /**
-    * A node with leaf children holds columns; anything else is a folder to descend into. Keeping
-    * the rule structural rather than depth-based means the projection does not care how deeply
-    * the Composer nests its tree.
+    * One group per source table, holding every column beneath it.
     *
-    * <p>{@code sourceName} is the nearest enclosing <em>table</em>, carried down rather than read
-    * from the node holding the columns. In the Composer's binding tree that node is a
-    * "Dimensions"/"Measures" folder, so naming the group after it produced a listing where every
-    * group was called "Dimensions" or "Measures" and no column could be traced to the table it
-    * came from — observed live, nine groups and two distinct names between them.
+    * <p>The Composer's tree is {@code TABLE -> Dimensions/Measures folder -> COLUMN}, so collecting
+    * a group per node-that-holds-columns produced <em>two</em> groups per table. That was tolerable
+    * only while they were named after the folders; once they are named after the table, it yields
+    * two entries with the same name whose only distinguishing feature was the label just replaced.
+    * A caller keying by table name -- the natural move -- silently keeps one and drops half the
+    * columns. So a table absorbs all of its descendants into a single group.
+    *
+    * <p>{@code sourceName} is the nearest enclosing table, carried down for the shapes that have no
+    * table ancestor at all.
     */
    private void collect(TreeNodeModel node, List<BindableTable> tables, String sourceName) {
-      String source = isTable(node) ? node.label() : sourceName;
-      List<BindableField> fields = new ArrayList<>();
+      if(isTable(node)) {
+         List<BindableField> fields = new ArrayList<>();
+         gather(node, fields);
+
+         if(!fields.isEmpty()) {
+            tables.add(new BindableTable(node.label(), fields));
+         }
+
+         return;
+      }
+
+      List<BindableField> direct = new ArrayList<>();
 
       for(TreeNodeModel child : node.children()) {
          if(child.leaf()) {
-            fields.add(new BindableField(child.label(), dataTypeOf(child)));
+            direct.add(fieldOf(child));
          }
          else {
-            collect(child, tables, source);
+            collect(child, tables, sourceName);
          }
       }
 
-      if(!fields.isEmpty()) {
-         // Fall back to the node's own label only when no table ancestor was found, so a tree
-         // shape this does not anticipate degrades to the old behaviour rather than to a blank.
-         tables.add(new BindableTable(source != null ? source : node.label(), fields));
+      // No table ancestor: name the group after the node that holds the columns. That is the
+      // pre-fix behaviour, and for the Composer's own tree it is the *bug* being fixed here -- a
+      // group called "Dimensions". It survives only as a floor for tree shapes this does not
+      // anticipate, where a wrong-but-present name beats a blank one.
+      if(!direct.isEmpty()) {
+         tables.add(new BindableTable(sourceName != null ? sourceName : node.label(), direct));
       }
+   }
+
+   /** Every column at or below this node, however deeply the Composer nests them. */
+   private void gather(TreeNodeModel node, List<BindableField> out) {
+      for(TreeNodeModel child : node.children()) {
+         if(child.leaf()) {
+            out.add(fieldOf(child));
+         }
+         else {
+            gather(child, out);
+         }
+      }
+   }
+
+   private BindableField fieldOf(TreeNodeModel node) {
+      return new BindableField(node.label(), dataTypeOf(node), roleOf(node));
    }
 
    /** Whether this node IS a source table, as opposed to a folder grouping one's columns. */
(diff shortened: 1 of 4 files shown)


larryliang-inetsoft (commented):
## Review — `list_bindable_fields`: data type and source table

Reviewed against `feature-76074-server-fixes`, reading the diff plus `VSTreeHandler`, `BaseTreeModelBuilder`, and the plugin-side consumer.

The core fix is right and well-evidenced. `VSTreeHandler.createNodeFromEntry` does build every node with `.data(entry)` (line 479), so the old `instanceof DataRefModel` branch could never match and every column really did report a null data type. Reading `dtype` off the `AssetEntry` is what the rest of the codebase does.

I also confirmed the rewritten `BindableFieldsServiceTest` **fails against the pre-fix production code** — the old `dataTypeOf` returns null, so `assertEquals("string", …)` fails, and the old grouping yields `Dimensions`/`Measures`, so the name assertions fail. It was not written to pass.

### The response now carries no dimension/measure signal at all

This is the one thing I'd change before merging, and it comes from the two halves of this PR interacting:

- `role` is removed from `BindableField`.
- Groups are renamed from the folder label to the ancestor table.

Previously the group name (`Dimensions` / `Measures`) carried the distinction implicitly. Now nothing does. The tool's own description tells the caller to "call this before reading or reasoning about any binding", and every binding tool downstream requires a mandatory `type` of `dimension` or `measure` per field. The tool you must call first no longer reports the thing the tools after it mandate — the caller has to guess it from the column name.

It also produces **duplicate `BindableTable` entries with identical names**. The Composer tree is `TABLE → Dimensions/Measures FOLDER → COLUMN` (`BaseTreeModelBuilder:607,615-618`), so each folder still yields its own group, now both named after the table. The new test asserts exactly this:

```java
assertEquals(2, tables.size());

for(BindableTable group : tables) {
   assertEquals("ORDERS1", group.name(), ...);
}
```

Two same-named groups whose only distinguishing feature *was* the label that got replaced. Any consumer keying by table name — the natural move for an LLM, or a `Map<String, BindableTable>` — silently keeps one and drops half the columns.

Either fix resolves both: merge the folders into one group per table and restore `role` per field, or keep the split and label it.

### The stated reason for removing `role` isn't right

> The binding tree does not distinguish the two, so there is nothing to populate it from.

It does. Every column entry carries `AssetEntry.CUBE_COL_TYPE` (`DIMENSIONS`/`MEASURES`), set in `BaseTreeModelBuilder.applyColumnNodeProperties:366` and `VSTreeHandler.createAssemblyColumnEntry:744`, and read exactly that way by `VSChartBindingHandler:408`, `VSCrosstabBindingHandler:412` and `WizardRecommenderUtil:121` — right next to the `dtype` property this PR just started reading. Simpler still, the immediate parent folder that this code now walks past *is* the Dimensions/Measures label.

The reasoning about an always-null field being worse than an absent one is sound, and the field was genuinely dead — I confirmed the plugin passes the response straight through and its own `role` handling is on the *input* side (an alias for `type` in field references), not a read of this. So removing it breaks nothing today. But the right remedy looks like populating it from `CUBE_COL_TYPE` rather than deleting it, and the javadoc as written will mislead the next person who goes looking.

### Minor

The fallback comment says an unanticipated tree shape "degrades to the old behaviour rather than to a blank" — worth being explicit that the old behaviour here is the *bug being fixed* (a group named `Dimensions`), not a safe prior state.

larryliang-inetsoft:
Both findings addressed in `353f7cd3b`. You were right on both, and the second was a claim of mine rather than an oversight.

**One group per table.** You're right that the two halves interact badly: renaming groups after the table without merging the `Dimensions`/`Measures` folders produces two entries with the same name, and a caller keying by table name silently keeps one and drops half the columns. My own test asserted that shape, which is the part I should have caught. A table now absorbs every column beneath it into a single group.

**`role` restored, populated from `CUBE_COL_TYPE`.** My stated reason for removing it — "the binding tree does not distinguish the two" — was simply wrong, and you gave the receipts: it's set in `VSTreeHandler.createAssemblyColumnEntry:744` right beside the `dtype` property this change had just started reading, and consumed exactly that way by `VSChartBindingHandler:408`, `VSCrosstabBindingHandler:412` and `WizardRecommenderUtil:121`. I verified before acting rather than taking it on faith.

The interpretation mirrors `VSChartBindingHandler.isDimension` exactly, numeric-type fallback included, so this tool and the binding handlers cannot disagree about the same column.

Your framing of *why* it mattered more after the merge is the part I'd missed: the group label was the last place the dimension/measure distinction lived implicitly, and every binding tool downstream requires it as a mandatory per-field `type`. The tool a caller is told to run first would have stopped reporting the thing the tools after it demand.

**The fallback comment** is corrected — it said an unanticipated shape "degrades to the old behaviour", which reads as a safe prior state when for the Composer's own tree the old behaviour is the bug being fixed.

Thanks for independently confirming the test fails against pre-fix code. Two new tests cover the role, one for `CUBE_COL_TYPE` and one for the numeric fallback. `inetsoft.web.wiz.*` — 31 tests, green.

larryliang-inetsoft (commented):
## Re-review — 353f7cd3

Both findings are addressed. One group per table, `gather` flattening everything beneath it; `role` restored and populated from `CUBE_COL_TYPE`. I checked the bit arithmetic against the constants — `DIMENSIONS = 0`, `MEASURES = 1`, `DATE_DIMENSIONS = 2`, so `(type & MEASURES) == 0` classifies a date dimension as a dimension. Correct.

Suites at this head: `BindableFieldsServiceTest` (6) and `BindableColumnsTest` (6) — all pass.

### One gap left, in the new `roleOf`

`CUBE_COL_TYPE` has three states, not two, and the empty one takes the wrong branch:

```java
centry.setProperty(AssetEntry.CUBE_COL_TYPE, type == -1 ? "" : type + "");
```
— `BaseTreeModelBuilder.applyColumnNodeProperties:366`

The property is **present but empty**, so `roleOf`'s `cubeColType != null` guard passes, `Integer.parseInt("")` throws, the `catch` returns null — and the `dtype` fallback immediately below never runs. Empty means "not stated", which is exactly the case the fallback exists for.

This is reachable, not hypothetical. `appendColumnNodes` passes `type = -1` for a table VS assembly (`BaseTreeModelBuilder:460`), and that is a tree `list_bindable_fields` sees: `VSBindingTreeService.getBinding` calls `appendVSAssemblyTree` for both the chart and table paths. So every column of a table assembly reports `role: null` while its `dataType` is sitting right there.

Verified by adding this to `BindableFieldsServiceTest` at this head:

```java
when(amountEntry.getProperty("dtype")).thenReturn("double");
when(amountEntry.getProperty(AssetEntry.CUBE_COL_TYPE)).thenReturn("");
...
assertEquals("measure", fields.get(0).role());
```

```
expected: <measure> but was: <null>
```

The other six tests pass; `fallsBackToTheDataTypeWhenTheTreeCarriesNoCubeColumnType` covers the *absent* case only, so nothing catches this. Suggested fix — treat blank as absent and let the `catch` fall through rather than return:

```java
String cubeColType = entry.getProperty(AssetEntry.CUBE_COL_TYPE);

if(cubeColType != null && !cubeColType.isBlank()) {
   try {
      return (Integer.parseInt(cubeColType) & AssetEntry.MEASURES) == 0 ? DIMENSION : MEASURE;
   }
   catch(NumberFormatException ex) {
      // Not a number: treat as unstated and fall through to the data type.
   }
}
```

### The "mirrors isDimension exactly" claim

Worth softening in the javadoc and the commit message. `VSChartBindingHandler.isDimension` only reaches its fallback when the property is `null`; on `""` it would throw `NumberFormatException` outright. So on the one input that actually occurs the two do not agree — the handler throws, `roleOf` returns null, and neither is what the column deserves. The `catch` is a genuine improvement over the mirrored behaviour; it just needs to fall through to the fallback rather than short-circuit it. Claiming exact parity hides that the original has a latent bug of its own on the same input.

larryliang-inetsoft:
Fixed in `865f4b400`. That was a good catch and I'd missed the third state entirely.

Verified before acting: `BaseTreeModelBuilder.applyColumnNodeProperties:366` writes `type == -1 ? "" : type + ""`, and `appendColumnNodes:460` passes `-1` for a table VS assembly — so the property really is present-but-empty for every column of a table assembly, and that tree is one `list_bindable_fields` reads.

`roleOf` now treats blank as unstated and the `catch` falls through to the data-type fallback rather than returning null. **The new test fails against the previous commit** — I confirmed by stashing the fix, since the existing fallback test covers only the absent case, which is exactly why nothing caught this.

**You're also right about the parity claim, and it's the more interesting half.** I wrote that `roleOf` "mirrors `isDimension` exactly, fallback included". It doesn't, on the one input that actually occurs: `isDimension` reaches its fallback only when the property is null, and would throw `NumberFormatException` on `""`. So on a table assembly's columns the two disagree — the handler throws, mine returned null, and neither is the answer the column deserves.

The javadoc now says it follows the same bit test and the same fallback, and states plainly that it deliberately does *not* match the handler's behaviour on the empty value, because matching there would only reproduce a latent bug. Claiming parity was hiding that.

`inetsoft.web.wiz.*` — 239 tests, green.

larryliang-inetsoft (commented):
## Re-review — 865f4b40

The three-state handling is right: `isBlank()` on the guard, and the `catch` falls through to the data-type fallback instead of short-circuiting it. The javadoc now says plainly where it departs from `isDimension` and why, which is more useful than the parity claim it replaces.

I verified the new test is a real guard rather than a description of current behaviour — reverted the guard to `cubeColType != null` with `return null` in the catch, and re-ran:

```
BindableFieldsServiceTest.treatsAnEmptyCubeColumnTypeAsUnstatedAndFallsBackToTheDataType
org.opentest4j.AssertionFailedError: expected: <dimension> but was: <null>
```

One failure, and only that one — so it isolates the fix cleanly.

At `865f4b40` unmodified: `BindableFieldsServiceTest` (7) and `BindableColumnsTest` (6) — **13 tests, 0 failures, 0 errors**.

Nothing outstanding. LGTM.

PR #4621, opened 2026-08-17
Typed script kinds: wire vocabulary, deterministic ids, and the grammar contract

larryliang-inetsoft:
Phase 0 + Phase 2 (Tier 1) of the script-kinds spec, plus the unshipped Phase 1 step 3.

Replaces the `:`-delimited script target string with a typed `kind` vocabulary, deterministic opaque ids, and a version-negotiated compatibility contract. No new read/write services — Tier 1 is the taxonomy and the addressing only.

## What changes

**`Kind` is the wire vocabulary; `Location` stays the internal dispatch key.** Five services and the image endpoint already switch on `Location`, so a wire rename must not ripple into them — and two of the new kinds (`worksheetExpression`, `worksheetCondition`) have no `Location` at all. They are reserved so a client's static schema does not churn when pane-scoped pairing lands; asking for one is a scope error, not an unknown kind.

**Every target carries a deterministic id.** Base64url of `kind|assembly`, reproducible without server state — the pairing store is in-memory, so a session-scoped id would strand any workflow that lists targets in one turn and edits in the next. This also fixes a latent bug: an assembly literally named `Foo:onClick` was previously unaddressable, because the static parse splits on a trailing `:onClick` and has no viewsheet to check against.

**`/targets` advertises the grammar rather than making clients probe.** `grammarVersion` and `supportedKinds` travel beside the target array — "Invalid target" is indistinguishable from "no such assembly", so a probe cannot tell an old server from a typo. `supportedKinds` lists only kinds with a service behind them; advertising one the server would then refuse is the capability lie the contract exists to prevent.

**Six endpoints accept three dialects** — id, `kind`+`assembly`, or the legacy string — resolved by one precedence rule in `ScriptTarget.resolve` rather than six copies that could disagree about which wins. Resolution happens where the runtime viewsheet is in hand, so the exact-name fix reaches the legacy dialect too.

**`get_script_context` is scoped to a target, by kind.** The usable API genuinely differs per context: `event` resolves from `ViewsheetScope`'s vmap and is populated only while an onClick executes, so listing it for a `viewsheetOnInit` script promised a variable that is always null there. With no target the union comes back, because the caller has stated no context to narrow to. `contextVars` is a `switch` with a documented default — it is the seam future kinds extend.

## Compatibility

The legacy `target` string still works on every endpoint, and `targets` remains a plain array at a stable key, so an older client that forwards the body without parsing it keeps working. The plugin and the server ship separately, which is why this is a contract rather than a cutover.

## Testing

`inetsoft.web.wiz.**.*Test`: **1487 → 1513**, 0 failures, 0 errors, 1 pre-existing skip. The full package was re-run and the surefire XML parsed after each of the five commits — not read off `BUILD SUCCESS`.

Not yet proven: the wire contract itself. The TypeScript client that consumes `grammarVersion`/`supportedKinds` is a separate change in `stylebi-wiz`, and neither end proves the other. Live verification comes after both land and an image is rebuilt.

## Notes for review

- No `name` field on `ScriptTargetInfo`, though the spec lists one. No Tier 1 kind has a third free-text component, so it would ship permanently null — the defect #4618 fixed by deleting `BindableField.role`. It arrives with the first kind that populates it.
- `calcField` / `highlight` / `cell` / `library` are deliberately absent from the enum. They are wanted — the API surface differs per context, which is what a kind is *for* — but each needs a read/write service behind it first. Order to add one: `Kind` + its `Location`, a `contextVars` case, then the service.
- The six rewired endpoints have no `ViewsheetAgentControllerTest`; wiring correctness currently rests on a per-endpoint manual review rather than an automated regression.

Spec: `docs/superpowers/specs/2026-08-14-script-plugin-scope-and-script-kinds.md`
Plan: `docs/superpowers/plans/2026-08-16-script-kinds-phase0-phase2.md` (both in `stylebi-wiz`)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/script/ScriptContextService.java
+++ b/core/src/main/java/inetsoft/web/wiz/script/ScriptContextService.java
@@ -19,11 +19,13 @@ package inetsoft.web.wiz.script;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import inetsoft.report.composition.RuntimeViewsheet;
+import inetsoft.report.script.viewsheet.ViewsheetScope;
 import inetsoft.uql.asset.Assembly;
 import inetsoft.uql.viewsheet.ChartVSAssembly;
 import inetsoft.uql.viewsheet.VSAssembly;
 import inetsoft.uql.viewsheet.Viewsheet;
 import inetsoft.web.binding.VSScriptableService;
+import inetsoft.web.wiz.pairing.PairingException;
 import inetsoft.web.wiz.script.model.AssemblyContext;
 import inetsoft.web.wiz.script.model.ScriptContext;
 import org.slf4j.Logger;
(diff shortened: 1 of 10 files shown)


larryliang-inetsoft:
Client half is up: inetsoft-technology/stylebi-wiz#1624 (plugin enum + dialect normalization + grammar negotiation, and the wiz-services pass-through).

Neither PR proves the other. Both sides' suites are green, but every test on the client side mocks the HTTP client, and this side has no controller-level wiring test — so the wire contract between them is unverified until this merges, an image is rebuilt, and the live checks run. The one worth doing first: confirm the script that *ran* is the script that was *addressed*, not merely that the server accepted the call.

PR #4622, opened 2026-08-17
Bug #76088, add datasource folder creation to the wiz portal API

Zxm-happy:
## Summary
- `WizDatabaseController` (`/api/wiz/**`) already listed `DATA_SOURCE_FOLDER` entries when browsing, and could create/edit JDBC databases and tabular datasources, but had no way to create a folder itself.
- Adds `POST /api/wiz/datasources/folders/create`, delegating to the existing `DataSourceBrowserService.checkFolderDuplicate` / `addDatasourceFolder` (the same service methods the native Angular portal's folder-add feature uses), gated by the same create-permission rule `createDatabase` already uses (WRITE on the parent folder, or `CREATE_DATA_SOURCE` at the root).
- `userScope` is always passed `false` — the wiz portal has no "create in my own scope" affordance, so the new folder gets no per-user permission grant.
- New `WizFolderCreateRequest` / `WizFolderSaveResult` records mirror the shape of the existing `WizDatabaseCreateRequest` / `WizDatabaseSaveResult`.

Base branch is `stylebi-wiz-main-rebase` (not `main`): `WizDatabaseController` and its whole `/api/wiz/datasources` surface don't exist on `main` yet — they're part of the still-unmerged wiz integration branch.

Paired with the wiz-services + portal change: inetsoft-technology/stylebi-wiz#1622

## Test plan
- [x] `WizDatabaseControllerSecurityTest`: updated the frozen endpoint-list contract, added 4 new cases (success, duplicate name, denied without create permission, name containing "/") — all passing
- [x] `mvn -pl community/core -am compile` — clean
- [ ] Manual verification against a running StyleBI + wiz stack (not yet done)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
@@ -19,6 +19,7 @@
 package inetsoft.web.wiz.controller;
 
 import inetsoft.report.internal.Util;
+import inetsoft.sree.RepositoryEntry;
 import inetsoft.sree.security.ResourceAction;
 import inetsoft.sree.security.ResourceType;
 import inetsoft.sree.security.SecurityEngine;
@@ -35,6 +36,7 @@ import inetsoft.web.admin.content.database.types.*;
 import inetsoft.web.admin.content.repository.DataSourceSettingsModel;
 import inetsoft.web.admin.content.repository.DatabaseDatasourcesService;
 import inetsoft.web.admin.security.ConnectionStatus;
+import inetsoft.web.portal.data.CheckDuplicateResponse;
 import inetsoft.web.portal.data.DataSourceBrowserService;
 import inetsoft.web.portal.data.DataSourceConnectionStatusRequest;
 import inetsoft.web.portal.data.DataSourceInfo;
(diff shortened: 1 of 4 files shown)


Zxm-happy:
Manual review pass on this PR turned up one confirmed bug, which is fixed in the latest commit:

**`userScope` was hardcoded to `false`.** A caller who reaches the root of the repository only via the standalone `CREATE_DATA_SOURCE` grant (no WRITE on `"/"`) would get no permission entry on the folder they just created and no root WRITE to inherit from — locked out of their own new folder immediately, exactly the case the native `DataSourceController`'s `userScope=true` exists to prevent. `requireCreatePermission` now reports which grant let the caller through, and `createFolder` passes that through as `userScope`. Added a test (`createFolder_grantsCreatorOwnershipWhenOnlyCreateDataSourcePermissionApplies`) covering the success path that was previously untested — that gap in coverage is exactly why this shipped undetected.

Also fixed: `requireFolderName` now rejects `"."` and `".."` in addition to `"/"`.

Two more things came up that I looked at and decided **not** to change, because they match pre-existing behavior in the underlying platform rather than being introduced here (documented in the `createFolder` javadoc):
- The duplicate check (`checkFolderDuplicate` then `addDatasourceFolder`) is advisory, not atomic — same non-atomicity `checkTabularDuplicateName` already has on the tabular create path.
- Neither this endpoint nor the native `DataSourceController.addDatasourceFolder` it mirrors confirms `parentPath` actually names an existing folder before creating inside it.

fredaan (approved):

PR #4623, opened 2026-08-17
Describe REST connector endpoints so retrieval can tell them apart

fredaan:
Adds a `description` to every endpoint of seven REST connectors, so that
retrieval has something to match on beyond a name, and so that a query written
against one of them is written against what it actually returns.

**1,175 endpoints described**, all seven connectors complete except the two
carried over from earlier work:

| connector | described |
|---|---|
| gitlab | 277 / 277 |
| github | 191 / 191 |
| jira | 154 / 154 |
| sendgrid | 133 / 133 |
| mailchimp | 127 / 127 |
| pipedrive | 108 / 108 |
| zendesk | 193 / 278 |
| stripe | 101 / 110 |

Each connector was aligned against its vendor's published OpenAPI description
before anything was written, which is also what surfaced everything below.

## The descriptions are mostly warnings

A name says what a table is. What costs an analyst a wrong answer is the part no
name carries: which column the question actually turns on, what the table
silently excludes, and where two tables that sound identical are not. Some
examples of what is now recorded:

- **GitHub treats every pull request as an issue.** Every issues endpoint
  returns pull requests mixed in, and `open_issues_count` counts both. An issue
  count taken without excluding them is inflated, frequently by half, and
  nothing in the response says so.
- **Jira issues are finished when `resolution` is set**, not when their status
  reads Done — status names are a per-project workflow choice, resolution means
  the same thing everywhere.
- **Pipedrive names custom fields with a 40-character hash.** A deal returns a
  column called `9dc80c50d78a15643bfc4ca79d76156a99a9ba60` and nothing in that
  response identifies it.
- **Mailchimp's member id is the MD5 of the lowercased email**, so a member can
  be addressed directly without searching first.
- **SendGrid's contacts endpoint returns at most 50 rows** and SendGrid has
  deprecated pagination on it. The full audience is reachable only by export.
- **GitLab's pipeline list carries no duration.** Only the single-pipeline
  endpoint does, so "how long do pipelines take" cannot be answered from the
  list at all.

Retention and sampling limits are recorded wherever they turn a total into a
floor: GitHub events keep ~90 days and lag up to six hours, GitHub traffic is a
fixed trailing 14 days, SendGrid's advanced stats keep seven, Mailchimp list
activity keeps 180 and silently omits automation activity.

## Broken paths, fixed

Aligning against the specs found endpoints that have never worked:

- **GitHub** (3): `Workflow Run Artifacts` missing its `/actions` segment,
  `Repository Secret` singular where the resource is plural, `Release Asset`
  nested under a release when assets are addressed by asset id alone. All three
  returned 404.
- **Jira** (2): `Alternative Issue Types` was missing the leading slash on its
  suffix, alone among 152. `Field Options` pointed at an experimental endpoint
  Atlassian removed; repointed at the context-scoped replacement, with
  `Field Contexts` added to supply the context id it needs.
- **Pipedrive** (4): Pipedrive renamed `/teams` to `/legacyTeams`; all four
  shapes match one for one.

Endpoints the vendor has retired are described as retired rather than left to
fail unexplained — 7 in GitHub, 2 in Jira, 5 in SendGrid, 3 in Pipedrive.

## Capability added

**GitHub had no pull request endpoints at all.** Not one of its 179 reached a
pull request, which put merge dates, review outcomes, reviewers and diff sizes
out of reach — the whole of code review, from the product whose defining feature
it is. Twelve endpoints added, each verified present in GitHub's description,
with lookups chaining the list to the per-pull endpoints on `number` (not `id`;
GitHub addresses pull requests by their per-repository number).

**Jira's only issue-listing endpoint is being removed by Atlassian.** Jira has
no plain "all issues" resource, so losing `GET /search` would leave the
connector able to describe Jira's configuration in detail and unable to return a
single issue. The replacement could not simply be added: `JiraQuery` hardcoded
one pagination spec for every paged endpoint — offset by `startAt` against a
`total` — and the enhanced search has neither, returning an opaque
`nextPageToken` instead. Added under the old spec it would have returned page
one and stopped, without error.

So pagination is now per-endpoint, following the pattern `LinkedinQuery`
already uses here: a `pageType` on the endpoint and a branch in
`updatePagination`. Only `JQL Search` declares it; everything without one keeps
the offset spec unchanged, which is what leaves the other 153 endpoints alone.

## Tests

`EndpointsJsonLoadableTest` (added in the first commit) parses all 65
connectors' `endpoints.json` through each connector's own `Endpoints` subclass.
It guards a failure that is silent by construction: the mapper rejects unknown
properties and `Endpoints.load` turns that rejection into an *empty endpoint
map*, so a single stray key leaves a connector present in the UI with no
endpoints and one line in the log. Nothing covered this before — the
per-connector tests are tagged `endpoints` and excluded from normal builds
because they call vendor APIs.

`JiraPaginationTest` pins both pagination branches and that only the one
endpoint declares a `pageType`. Verified it fails when `pageType` is removed,
and that the message names the cause rather than just the mismatch.

## Note on the build

`EndpointJsonQueryTest` errors on this branch with a Spring
`NoClassDefFoundError`. Confirmed pre-existing: it fails identically on
`stylebi-wiz-main` with none of these commits applied. Both tests added here
pass. Also note this module needs `-am` to compile on this branch, which is
likewise pre-existing.

## Follow-up, not attempted here

**Pipedrive should move to API v2.** Of its 108 endpoints, 21 are the core of
the CRM — deals, persons, organizations, activities, pipelines, stages,
products — and Pipedrive has pruned exactly those from the v1 description and
reissued them under `/api/v2`. Each now says so in its description. The
migration itself changes the base path, the pagination model (cursor rather than
offset) and several response shapes, and could not be validated here without a
live account, so it is flagged rather than guessed at.

Remaining connectors, in descending size: box (82), asana (65), xero (54),
twitter (47), square (30), twilio (15), plus finishing zendesk (85) and stripe
(9).


fredaan:
**Correction.** This PR describes a missing leading slash on an endpoint suffix as a broken path — `Alternative Issue Types` here in #4623, `Alert` in #4625. That is wrong, and I've recorded it in #4627.

`URLCreator.combineDataSourceAndQueryUrlFragments` inserts the separator when neither side has one:

```java
   url += "/" + suffix;
}
```

So the endpoint worked before the change and works after it. The edit is stylistic consistency, not a repair, and calling it unreachable overstated the case. A scan across all 65 connectors finds 14 such suffixes in five connectors, all harmless for the same reason; none are being changed.

The other path fixes in this stack are unaffected — each fails for a reason the framework does not paper over (wrong path segments, a missing `/api.xro` segment, a `?` where a `/` belongs, a swapped parameter name, a wrong resource, and two that had moved).

fredaan:
## Review result

Reviewed as part of the stack-wide pass. One thing was found against this PR and has already been corrected in the comment above; everything else below came back clean.

### Structural diff

Diffed against `stylebi-wiz-main` object by object, so anything changed beyond adding a `description` had to be justified:

- **16 files** — 8 connectors, 4 main Java files, 3 test files, 1 test resource
- **2 endpoint-count changes**, both deliberate: github 179 → 191, jira 152 → 154
- **14 added** — the 12 GitHub pull request endpoints, plus jira's `Field Contexts` and `JQL Search`
- **0 removed**
- **10 changed beyond description**, each a path fix this PR describes:

| connector | endpoints |
|---|---|
| github | `Workflow Run Artifacts`, `Repository Secret`, `Release Asset` |
| jira | `Field Options`, `Alternative Issue Types` |
| pipedrive | `Teams`, `Team`, `Team Users`, `User Teams` — the `/teams` to `/legacyTeams` rename |
| sendgrid | `Event Webhook Settings` |

### The one finding

`jira / Alternative Issue Types` is in that list of ten, and the correction above is about it: a missing leading slash is **not** a broken path — `URLCreator` inserts the separator — so that edit is stylistic consistency, not a repair. The endpoint worked before and after. The other nine fail for reasons the framework does not paper over.

### Java changes

Reviewed line by line. Four main-code files, all additive or behaviour-preserving except the one intended branch:

- `AbstractEndpoint` — adds the optional `description`, deliberately excluded from `equals`/`hashCode` so rewording an endpoint does not make it a different endpoint
- `EndpointJsonQuery` — `createObjectMapper` widened to public, so per-connector tests can parse through the same mapper the loader uses
- `JiraEndpoint` / `JiraQuery` — the `pageType` property and the pagination branch. Endpoints that declare no `pageType` keep the offset spec they have always had, which is what leaves the other 153 alone.

### Coverage

All 8 connectors complete at the stack tip: **1,381 / 1,381**. zendesk and stripe were deliberately partial here (193/278 and 101/110) and finished in #4627.

### Scans

Every systematic scan run over the final state is clean across these 8: no control characters or stray whitespace in a suffix, no double slashes, no non-ASCII in an endpoint name, no unbalanced braces, no duplicate endpoint names, no duplicate suffixes, and no lookup pointing at a missing endpoint.

### Later fixes in the stack

Five further defects surfaced in the stack-wide review — an unterminated placeholder in wordpress, and five descriptions promising a parameter their endpoint does not take (jive, linkedin, influxdb). **None are in this PR's connectors**; all belong to #4629.

### Tests

`JiraPaginationTest` and `EndpointsJsonLoadableTest` both pass, and both were checked for vacuity: removing `pageType` makes the first fail with a message naming the cause, and injecting a bogus property into any connector file makes the second fail and name the connector, the class and the known properties. The module's other 21 errors are the pre-existing Spring `NoClassDefFoundError` — confirmed identical in count on `stylebi-wiz-main` with none of this stack applied.

fredaan:
## Rebased onto `stylebi-wiz-main-rebase`

The base was wrong. This PR was built on `stylebi-wiz-main`, and the two lines have genuinely diverged — 2,177 commits on one side, 1,613 on the other — so simply retargeting would have dragged all of that into the diff. The whole stack has been rebased instead, and all five branches force-pushed.

**Content is unchanged.** Every one of the 65 `endpoints.json` files and every Java file I touched is byte-identical between the old tip and the new one; the only files that differ are four GraphQL ones that differ between the two baselines themselves and that this work never touched. Coverage is still 4,035 / 4,035 and the three tests still pass.

### Two commits dropped, and why

`stylebi-wiz-main-rebase` already contains **#4604**, which was the first branch of this work before the base mix-up. Its two commits are therefore already upstream, and the rebase dropped them by patch-id — verified identical:

| commit | patch-id |
|---|---|
| merged via #4604 | `8c35989102a2`, `52f02d3eb2e7` |
| dropped from this stack | `8c35989102a2`, `52f02d3eb2e7` |

So 36 commits became 34, and **two claims in the description above are now stale**:

- `EndpointsJsonLoadableTest` is **not added by this PR** — it arrived with #4604 and is already on the base. It still stands behind everything here, and the later PRs extend its reach, but the commit that introduced it is not in this diff.
- The **stripe row in the coverage table is stale**: stripe's 101/110 also came with #4604. This PR now touches 7 connectors, not 8.

### Corrected figures for this PR

| connector | total | described here |
|---|---:|---:|
| gitlab | 277 | 277 |
| zendesk | 278 | 193 |
| github | 191 | 191 |
| jira | 154 | 154 |
| sendgrid | 133 | 133 |
| mailchimp | 127 | 127 |
| pipedrive | 108 | 108 |

**1,183 endpoints** described here, across 7 connectors. zendesk is deliberately partial and finished in #4627.

The per-PR breakdown posted on #4629 has been corrected to match, and the stack now sums as 1183 + 303 + 97 + 620 + 1731 = 3,934, plus the 101 already on the base via #4604 = **4,035**.

### What this PR still carries beyond descriptions

Four non-JSON files: `JiraEndpoint`, `JiraQuery`, `EndpointJsonQuery` (the `createObjectMapper` visibility widening, applied on top of #4604's version) and `JiraPaginationTest`. The structural review above is otherwise unaffected — the 14 added endpoints, the 10 path fixes, and the finding about the leading slash all stand.

Nelson-cao (approved):

PR #4624, opened 2026-08-17
Expose an existing assembly's rendered data as a read-only endpoint

NeverStopFMH:
POST /api/wiz/viewsheet/assembly-data returns headers, rows and binding for an assembly already in a runtime viewsheet.

It exists so a follow-up question about a chart built earlier in a conversation can be answered from that chart's own rows. The caller keeps no copy of them, and deliberately so: a stored copy is the only thing capable of disagreeing with what the user is looking at.

No extraction logic is added. WizVsService#fetchAssemblyData already does this work — it is how a type switch lazily repopulates row data for insight generation — and was simply never reachable over HTTP. It becomes public.

The request carries only runtimeId and assemblyName, which is the whole point. They are exactly the pair the browser embed renders with, so whatever comes back is by construction the data behind the chart on screen. A row cap, a filter or a binding override would break that correspondence: the caller could then fetch data the chart does not show and answer a question about the chart from numbers the user cannot see. The cap in force stays whatever the chart was rendered with, which fetchAssemblyData already preserves by reading it back from the source worksheet rather than accepting one.

An assembly whose runtime has been reaped returns an empty result rather than an error — the chart is equally unavailable to the user at that point, so "no data to answer from" is the accurate outcome, not a failure to retry.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizViewsheetController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizViewsheetController.java
@@ -23,8 +23,10 @@ import inetsoft.web.wiz.service.UnsatisfiableBindingException;
 import inetsoft.web.wiz.service.WizAutoBindingService;
 import inetsoft.web.wiz.service.WizGeoService;
 import inetsoft.web.wiz.service.WizVsService;
+import inetsoft.util.InvalidUserException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 import jakarta.validation.Valid;
@@ -118,7 +120,7 @@ public class WizViewsheetController {
    }
 
    /**
-    * Read-only: returns an existing assembly's rendered data (headers + rows + binding).
+    * Returns an existing assembly's rendered data (headers + rows + binding).
     *
     * <p>For answering a question about a chart built EARLIER in a conversation. The caller holds no
     * copy of that chart's rows; it addresses the chart the way the browser embed does — runtime id
@@ -126,9 +128,16 @@ public class WizViewsheetController {
     * chart's data range (see {@link AssemblyDataRequest}), so the answer and the chart on screen
     * cannot disagree.
     *
+    * <p><b>Read-only in intent, not in effect.</b> Serving this re-executes the assembly's view and
+    * drops its cached render (see {@code WizVsService#executeAndExtract}), and the underlying
+    * sandbox call cancels every in-flight query on that runtime — not only this assembly's. It is
+    * therefore not free to call, and calling it while the browser is rendering the same viewsheet can
+    * cancel an unrelated chart's render.
+    *
     * <p>An assembly whose runtime has been reaped comes back as an empty result rather than an
     * error: the chart is equally unavailable to the user at that point, so "no data to answer from"
-    * is the accurate outcome, not a failure to retry.
+    * is the accurate outcome, not a failure to retry. Same for an assembly that is no longer in the
+    * runtime.
     */
    @PostMapping(value = "/viewsheet/assembly-data", produces = MediaType.APPLICATION_JSON_VALUE)
    public ResponseEntity<?> assemblyData(@Valid @RequestBody AssemblyDataRequest request,
@@ -187,6 +196,21 @@ public class WizViewsheetController {
       }
       catch(IllegalArgumentException e) {
          return ResponseEntity.badRequest().body(Map.of("error", nullToEmpty(e.getMessage())));
+      }
+      // Denials, not faults. WizControllerErrorHandler already maps both of these to 403 for the wiz
+      // package — but an @ControllerAdvice only sees what LEAVES the controller method, and the
+      // catch-all below swallows everything first. So every permission denial on this controller was
+      // answered as a content-free 500 while the advice built to prevent exactly that sat unused; the
+      // same trap the ResponseStatusException case below documents.
+      //
+      // Neither message is echoed. SecurityException carries a catalog string that is safe but
+      // uninformative to the caller, and InvalidUserException's message ("common.invalidUser") names
+      // the user who OWNS the runtime — telling a caller who probed someone else's runtime id whose it
+      // is would answer a question they should not get to ask.
+      catch(inetsoft.sree.security.SecurityException | InvalidUserException e) {
+         LOG.warn("Unauthorized wiz request ({}): {}", action, e.getMessage());
+         return ResponseEntity.status(HttpStatus.FORBIDDEN)
+            .body(Map.of("error", "Forbidden"));
       }
       // Honour the status the thrower chose. ResponseStatusException is a plain RuntimeException, so
       // without this it fell through to the generic handler below and every deliberate 4xx was returned
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizViewsheetController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizViewsheetController.java
@@ -117,6 +117,27 @@ public class WizViewsheetController {
       return run("geo apply", () -> wizGeoService.apply(request, user));
    }
 
+   /**
+    * Read-only: returns an existing assembly's rendered data (headers + rows + binding).
+    *
+    * <p>For answering a question about a chart built EARLIER in a conversation. The caller holds no
+    * copy of that chart's rows; it addresses the chart the way the browser embed does — runtime id
+    * plus assembly name — and gets back what that chart is showing. Nothing here can change the
+    * chart's data range (see {@link AssemblyDataRequest}), so the answer and the chart on screen
+    * cannot disagree.
+    *
+    * <p>An assembly whose runtime has been reaped comes back as an empty result rather than an
+    * error: the chart is equally unavailable to the user at that point, so "no data to answer from"
+    * is the accurate outcome, not a failure to retry.
+    */
+   @PostMapping(value = "/viewsheet/assembly-data", produces = MediaType.APPLICATION_JSON_VALUE)
+   public ResponseEntity<?> assemblyData(@Valid @RequestBody AssemblyDataRequest request,
+                                         Principal user)
+   {
+      return run("read assembly data", () -> wizVsService.fetchAssemblyData(
+         request.getRuntimeId(), request.getAssemblyName(), user));
+   }
+
    @PostMapping(value = "/viewsheet/remove-visualization", produces = MediaType.APPLICATION_JSON_VALUE)
    public ResponseEntity<?> removeVisualization(@Valid @RequestBody RemoveVisualizationRequest request,
                                                 Principal user)
--- a/core/src/main/java/inetsoft/web/wiz/model/AssemblyDataRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/model/AssemblyDataRequest.java
@@ -1,0 +1,59 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package inetsoft.web.wiz.model;
+
+import jakarta.validation.constraints.NotBlank;
+
+/**
+ * Request body for reading an existing assembly's rendered data.
+ *
+ * <p>{@code runtimeId} is the runtime viewsheet id (the wiz-side visualizationId);
+ * {@code assemblyName} is the assembly to read. Both must be non-blank. Any assembly type the
+ * service dispatches on is valid here — chart, table, crosstab, and the scalar output assemblies
+ * (gauge / text).
+ *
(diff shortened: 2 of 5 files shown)


NeverStopFMH:
## Review

Read the diff and followed the call chain into `WizVsService#executeAndExtract`, `ViewsheetEngine#getViewsheet`, `WorksheetEngine#getSheet` and `ComposerControllerErrorHandler` to check the claims the docs make.

### What's right

The contract is the valuable part of this PR, and it is the right one. Accepting only `runtimeId` + `assemblyName` — no row cap, no filter, no binding override — makes "what comes back is the data behind the chart on screen" a structural property rather than a convention. `@NotBlank` on both fields. It reuses the existing extraction instead of duplicating it. And the Javadoc explains *why* rather than *what*, including why the request must stay this narrow.

**Authorization is enforced** — the first thing I went looking for, since this is a new endpoint returning row-level data addressed by an opaque id. `viewsheetService.getViewsheet` → `WorksheetEngine.getSheet(id, user, false)`, which at line 565 throws `InvalidUserException` when `rs.getUser() != null && user != null && !rs.matches(user)`. A runtime id belonging to another user does not read out. No IDOR.

---

### Must fix

#### F1 — the behaviour the Javadoc, the PR body and a test name all assert does not happen

All three state some version of:

> An assembly whose runtime has been reaped comes back as an **empty result rather than an error**.

It does not. `WorksheetEngine.getSheet:560-562`:

```java
RuntimeSheet rs = amap.get(id);
if(rs == null) {
   LOG.debug("Worksheet/viewsheet has expired: {}", id, new Exception("Stack trace"));
   throw new ExpiredSheetException(id, user);   // throws; never returns null
}
```

Two consequences:

1. **`WizVsService:2227`'s `if(rvs == null) return new CreateViewsheetResult();` is dead code.** `getSheet` either throws or returns non-null, so the branch that looks like it implements the documented behaviour can never run.
2. `ExpiredSheetException extends RuntimeException` — neither `IllegalArgumentException` nor `ResponseStatusException` — so it lands in `run()`'s last catch:

```java
catch(Exception e) {
   LOG.error("Failed to {}", action, e);   // ERROR, with a stack trace
   return ResponseEntity.internalServerError()...
}
```

**HTTP 500 plus an ERROR-level stack trace.** I also checked whether an advice could intercept it first: `ComposerControllerErrorHandler` is `@ControllerAdvice(basePackages = {"inetsoft.web.binding", "inetsoft.web.composer", "inetsoft.web.viewsheet", "inetsoft.web.vswizard"})` — `inetsoft.web.wiz` is not in the list, and in any case `run()` catches it before an advice would see it.

Why this matters more than a doc typo: a reaped runtime is the **most ordinary condition on this path**. It is exactly the case the caller is designed to degrade on — a question about a chart from an earlier session. The user sees nothing wrong, because the wiz side catches any failure and answers from stored facts. But StyleBI logs an ERROR with a stack trace and returns a 500 for routine operation, which buries the failures that do matter and makes alerting on this endpoint useless.

One nuance worth stating: composer's own advice also maps `ExpiredSheetException` to 500, so 500 is the house convention for it. The defect is not the status — it is that this endpoint **documents** an empty result and does not deliver one. Either the behaviour or the documentation has to move; the behaviour is the better one to move, because the caller's whole degradation model is built on "no rows means answer from facts".

The docs also conflate two cases that behave differently: **assembly deleted** (`vs.getAssembly(name) == null` → empty result, as documented) and **runtime reaped** (throws). Only the first matches the text.

#### F2 — the test named after that claim cannot test that claim

```java
@Test
void returnsAnEmptyResultWhenTheAssemblyIsGone() throws Exception {
   WizVsService svc = mock(WizVsService.class);
   when(svc.fetchAssemblyData(any(), any(), any())).thenReturn(new CreateViewsheetResult());
   ...
   assertEquals(HttpStatus.OK, resp.getStatusCode());
```

It stubs `fetchAssemblyData` to return an empty result and then asserts that it got an empty result. That asserts Mockito, not StyleBI — and it is why F1 survived: a test named after the claim can never disprove the claim.

All five tests mock the service, so **nothing in this PR exercises what the PR is about**: reading real assembly data. For "existing logic made public" that is largely defensible — but the expired path is precisely where it needed a real test, and precisely where the behaviour diverged. At minimum this needs one test on `fetchAssemblyData` itself.

`mapsIllegalArgumentTo400` and `mapsUnexpectedFailureTo500` only restate `run()`'s pre-existing behaviour. Fine as guards; not coverage of this change.

---

### Worth discussing

#### F3 — "read-only" is not accurate, and one of the side effects is user-visible

`executeAndExtract` (`WizVsService:2159-2195`) does four things, none of them reads:

| Call | Effect |
|---|---|
| `box.cancelAllQueries()` | Cancels **every in-flight query on that runtime viewsheet**, not just this assembly's |
| `WizUtil.applySampledPreviewCap(bws, curMax)` | Writes the cap onto the shared base worksheet (nominally a no-op since `curMax` is read from the same place, but still a write to shared state) |
| `box.resetDataMap(name)` + `box.clearGraph(name)` | Drops that assembly's cached render, so the user's next interaction re-runs the query |
| `box.executeView(name, true)` | Actually executes a query against the data source |

The difference that matters is **when it runs**. The existing caller fires immediately after `createViewsheetSkipExecution`, inside the turn that owns the runtime, with nothing else in flight. The new caller fires at an arbitrary moment while the browser may be rendering the same viewsheet — and `cancelAllQueries()` is per-box, not per-assembly. So **fetching data for chart A can cancel the render of chart B** in the same viewsheet.

Three consequences:

1. "Read-only" in the Javadoc and PR body should say what it actually is: read-only in *intent*, but it re-executes and invalidates that assembly's cached render. Otherwise the next reader will assume it is free and call it freely.
2. The caller-side cost argument ("one request is small next to 3-4 LLM calls") is understated — this is a query execution plus a cache invalidation, not a cache read.
3. Nothing takes a lock. Whether `ViewsheetSandbox` tolerates concurrent `executeView` on one box from two threads is beyond what I can establish from reading — worth a check by someone who knows the sandbox. If it does not, this is a race on the user's live session.

#### F4 — the code contradicts itself about whether `result` can be null

```java
CreateViewsheetResult result = executeAndExtract(rvs, assembly, curMax);

if(result != null) {                                        // assumes it can be null
   result.setBinding(collectFlatBinding(assembly));
}

boolean metadataMode = vs.getViewsheetInfo().isMetadata();
result.setHasData(computeHasData(metadataMode, result));    // dereferences it unguarded
```

Both cannot be right. Every branch of `executeAndExtract` returns non-null as far as I traced, so the guard is probably the redundant one — but now that this path is reachable over HTTP it should be settled either way rather than left half-guarded. Pre-existing, not introduced here; newly reachable from outside, which is what makes it worth resolving now.

---

### Minor

- **The response returns the whole `CreateViewsheetResult`**, while the documented contract is headers + rows + binding and the caller reads only those three. It also carries `hasData`, `note`, `warnings`, `runtimeId`, `viewsheetIdentifier`, `error_message`. Not harmful today, but the response shape becomes "whatever that class currently holds" — a future field carrying something internal ships by default. Either a dedicated response DTO, or say so in the Javadoc.
- **An authorization rejection maps to 500.** `InvalidUserException` also falls into `run()`'s generic catch, so a cross-user attempt is logged as `LOG.error("Failed to read assembly data")` and answered 500 rather than 403. Not introduced here — every wiz endpoint shares `run()` — but this is the first one whose payload is raw row data, so it is the one where "denied" reading as "server broke" costs you detection. `ComposerControllerErrorHandler` already handles `InvalidUserException` properly; `inetsoft.web.wiz` is outside its `basePackages`.
- **Is `Principal` guaranteed non-null here?** `getSheet`'s ownership check is skipped entirely when `user == null` (or when the runtime has no owner). If any deployment path can reach `/api/wiz/**` unauthenticated, the check silently does not run on an endpoint that returns raw rows. Structurally identical to the other wiz endpoints, so probably fine — but worth confirming rather than assuming, given what this one returns.
- **No test for `@Valid` rejection** (blank `runtimeId` / `assemblyName`). Also, Spring's default validation error body differs in shape from this endpoint's own `{"error": ...}` — a small consistency wrinkle.

---

### Overall

Small change (+215/−2), right direction, and the narrow request contract — the part that actually carries the safety property — is done correctly.

**F1 and F2 should land before merge**: one turns the most ordinary condition on this path into a 500 with an ERROR stack trace, and the other is the test structure that let it through. F3 needs at least the wording corrected and the sandbox concurrency question answered by someone who knows it. F4 and the minor items can follow.

NeverStopFMH:
## Review response — pushed as `3b2c362b`

All four findings are fixed, plus **two I missed on the first pass** and only found while writing the tests. Everything was driven test-first: the expired-runtime path was watched to fail, with the exception escaping `fetchAssemblyData:2225`, before any code changed.

---

### The two I missed

**The action-level permission gate was absent.** Every other endpoint-backed method in the wiz services gates on `checkPermission(user, VIEWSHEET, "*", ACCESS)` — `removeVisualization` on this same controller (`WizVsService:5537`), and, decisively, the **read-only** `getChartAestheticModel` (`WizAutoBindingService:3530`). So the convention covers reads, and this endpoint — the one that returns **raw row data** — was the only ungated one. My review checked runtime *ownership* (`getSheet`'s `matches()`), found it enforced, and stopped there; ownership is not the same question as "may this user use viewsheets at all". Added, with a `verifyNoInteractions(vsService)` test proving the denial lands before the runtime is looked up. It is a no-op for the four internal callers in `WizAutoBindingService`, which pass the identical check first.

**A permission denial could not have produced a 403 on this controller.** `WizControllerErrorHandler` maps `SecurityException` to 403 for the whole `inetsoft.web.wiz` package — but an `@ControllerAdvice` only sees what *leaves* the controller method, and `run()`'s `catch(Exception)` takes it first. So every denial on `WizViewsheetController` (create, highlight, date-comparison, autoBinding, format, colors, geo, aesthetic model, remove) was answered as a content-free **500** by the very controller that advice exists to cover. This is the same trap the `ResponseStatusException` case in `run()` already documents in its own comment. Fixed there, which fixes it for every endpoint on the controller. `WizVisualizationController` already maps `SecurityException` to 403 (`WizVisualizationControllerRenderTest.mapsSecurityExceptionTo403`), so this aligns the two rather than inventing a status.

Note this pair is coupled: adding the gate *without* fixing `run()` would have answered denials with a 500. Worth flagging as a general point — a gate added to any wiz service reached through `run()` has this problem.

`InvalidUserException` is mapped alongside it. Neither message is echoed: its message is the `common.invalidUser` catalog string, which **names the user who owns the runtime**. Returning that to a caller who probed someone else's runtime id would answer a question they should not get to ask, so both return a fixed `{"error": "Forbidden"}` and the detail goes to the log. There is a test asserting the body specifically.

---

### The four from the review

**F1 — expired runtime.** Confirmed live, not inferred: the RED run failed with `ExpiredSheetException: Sheet [rt] has expired at WizVsService.fetchAssemblyData(WizVsService.java:2225)`. Now caught and answered with an empty result, so the Javadoc, the PR body and the caller's degradation model finally agree.

Caught as `ExpiredSheetException` specifically, **not** `catch(Exception)` — even though the sibling `removeVisualization` uses the broad form. Only expiry is expected here, and swallowing everything into "no data" on a *read* would turn a genuine fault into a silently wrong answer, which is the failure mode this whole feature exists to remove. The dead `rvs == null` branch is gone; `getSheet` throws rather than returning null, and an unreachable branch that looks like it implements the documented behaviour is worse than none.

Blank-identifier validation added too, matching `removeVisualization`. The controller's `@NotBlank` covers HTTP; the method is public now.

**F2 — the test that asserted Mockito.** `returnsAnEmptyResultWhenTheAssemblyIsGone` is renamed `passesAnEmptyResultThroughAsA200` — what a stubbed service can actually establish — and the behaviour it claimed moved to `WizVsServiceFetchAssemblyDataTest` at the service level, following the existing `WizVsService*Test` pattern (mocked `ViewsheetService` / `SecurityEngine`). The controller test class now says explicitly that service behaviour belongs there, so the next person does not put it back.

**F3 — "read-only".** Behaviour unchanged; the docs no longer claim something the code does not do. Both the endpoint and `executeAndExtract` now state that serving this re-executes the view, drops the assembly's cached render, writes the cap onto the shared base worksheet, and — via `cancelAllQueries`, which is per-box, not per-assembly — cancels every in-flight query on that runtime, so calling it while the browser renders the same viewsheet can cancel an unrelated chart's render.

**I did not change `cancelAllQueries`.** Narrowing it is a sandbox-semantics decision I cannot make from reading, and the concurrency question in my review still stands: whether `ViewsheetSandbox` tolerates concurrent `executeView` on one box from two threads needs someone who knows that code. The wiz-side design doc now records this as a known edge (`stylebi-wiz` spec §5.3) rather than leaving the cost model reading as "one cheap request".

**F4 — nullability.** `if(result != null) setBinding(...)` followed by an unguarded `result.setHasData(...)` is now one early return. Also removed the stale note in `executeAndExtract`'s Javadoc claiming `fetchAssemblyData` is "already package-private" — this PR made it public.

**Minor, done:** the response contract is documented, including that `truncated` / `sampled` / `sampleMaxRows` come back and are meaningful to a caller deciding whether the rows are the whole population — the wiz side currently derives its own truncation flag and could use these instead.

---

### Not addressed

- **Is `Principal` guaranteed non-null on `/api/wiz/**`?** Still an open question, unchanged. `getSheet`'s ownership check is skipped entirely when `user == null`. Structurally identical to every other wiz endpoint, so not this PR's to answer, but the answer matters more here than elsewhere.
- **A dedicated response DTO.** Documented rather than narrowed; changing the response type is a caller-visible change and the extra fields are useful, not leaky, today.
- **No `@Valid` rejection test.** Spring handles it before the method body, so it would test the framework.

---

### Verification

```
mvnw test -pl core -Dtest="inetsoft.web.wiz.**"
Tests run: 1040, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

The **whole** wiz package, not just the two classes, on purpose: the new gate sits on a method with four internal callers in `WizAutoBindingService`, and Mockito's default for a `boolean` stub is `false` — a test that reaches `fetchAssemblyData` through an ungranted `SecurityEngine` mock would have started failing. None did, but that is now verified rather than assumed.

One correction to my review: it said an authorization rejection mapping to 500 was "not introduced here — every wiz endpoint shares `run()`". Accurate, but I filed it as minor. It is not minor: combined with the missing gate it meant a denial on the endpoint returning raw rows would have surfaced as a server error. Fixed at the shared point rather than worked around at the call site.

fredaan (approved):

PR #4625, opened 2026-08-17
Describe six more REST connectors, and add Twilio's missing calls and messages

fredaan:
Continues #4623 with six more connectors — Box, Asana, Xero, X, Square and
Twilio — all now completely described. **Stacked on #4623**; retarget to
`stylebi-wiz-main` once that merges.

| connector | described |
|---|---|
| box | 82 / 82 |
| asana | 65 / 65 |
| xero | 54 / 54 |
| twitter | 47 / 47 |
| square | 30 / 30 |
| twilio | 25 / 25 |

Running total across both PRs: **1,587 endpoints described**, with only zendesk
(193/278) and stripe (101/110) still partial.

## Twilio had no calls and no messages

Fifteen endpoints covering accounts, addresses, applications, API keys, usage and
alerts — and nothing for the subject matter of a Twilio account. Ten added, each
verified against Twilio's published description: calls, messages, recordings and
conferences with their single-item forms, plus per-call and per-conference
recordings. No Java change needed; `TwilioRuntime` routes only the monitor-host
endpoints by name and everything else already gets the right pagination.

Deliberately left out the inequality date filters (`StartTime>`, `DateSent<`)
Twilio documents for ranges. No suffix in this repository uses a comparison
operator in a parameter name and there is no way to test the URL builder against
the live API from here, so exposing them would be a guess. The descriptions say
the ranges exist and that these endpoints do not offer them.

## Five more broken paths

Each of these returned something — which is worse than an error, because nothing
signals the problem:

- **Box** — `Trashed File` and `Trashed Folder` carried the *identical* suffix,
  `/folders/{id}/trash`. Both paths are real, so path alignment matched both;
  only the name gave it away. `/files/{id}/trash` was unreachable through this
  connector.
- **Square** — `Cash Drawer Shift Events` had a question mark where a slash
  belongs: `.../shifts/{Shift ID}?/events?location_id=...`. Everything after it
  is query string, so it returned the **shift** rather than its events.
- **Twilio** — `Alerts` had a parameter name and its label swapped,
  `&End Date={EndDate?...}`. The name goes on the wire, so it sent a parameter
  literally called `End Date`, Twilio ignored it, and the end-date filter
  silently did nothing. `Alert` was also missing its leading slash.
- **Xero** — `Aged Receivables By Contact` was missing the `/api.xro` prefix
  every other accounting endpoint carries.
- **Asana** — two endpoints had moved rather than broken: organization teams is
  now a workspaces path (an organization *is* a workspace), and task attachments
  became a general attachments endpoint taking the task as a `parent` parameter.

Box's surfaced from a scan for two endpoints sharing one suffix. That scan also
found eight more in connectors not covered here — stripe, hubspot, xero, toggl,
freshservice, campaignmonitor — left for when those are done rather than guessed
at now.

## What the descriptions carry

Two connectors withhold almost everything by default, and it looks like empty
data rather than an unmade request:

- **Box** returns a minimal record — size, timestamps, owner and parent are
  absent unless `fields` names them.
- **Asana** returns `gid`, `name`, `resource_type` and nothing else; completed,
  due dates, assignee and custom fields all need `opt_fields`.
- **X** is the same again: a post is `id` and `text` until `tweet.fields` asks
  for more, and a user is `id`, `name`, `username` — so follower counts, the most
  commonly wanted column, are absent by default.

Three have no way to list the thing you actually want:

- **Box** has no endpoint that lists files. Content is reached only by walking
  folder items from folder 0, or by search.
- **Asana** *refuses* an unfiltered task request. A task set is reachable only
  through a project, section, tag, user task list, or assignee plus workspace.
- **Square** puts team-member and labour-shift search behind POST, which a
  read-only connector cannot reach, so both can only be fetched one at a time by
  an id obtained elsewhere.

And several where one table holds two opposite things:

- **Xero** — one Invoices table holds both sales invoices (`ACCREC`) and supplier
  bills (`ACCPAY`). A revenue total that does not filter on Type is not a revenue
  total. Same for Credit Notes, and for Bank Transactions where Type separates
  SPEND from RECEIVE. Contacts likewise holds customers and suppliers together.
- **Square** — refunds are not negative payments and appear in a separate table,
  so net revenue is one minus the other.

Unit and shape traps that quietly multiply a figure:

- **Square money is in the smallest currency unit** — `1000 USD` is ten dollars.
- **Xero reports return a nested `Rows` structure**, not flat rows; the section a
  row sits in is what gives it meaning.
- **Twilio bills by `num_segments`, not message count**, so volume and cost do
  not move together — and a call's `price` is negative and null until rated.
- **Square prices live on item *variations*, not items**, so a priced catalogue
  is a join between two types returned by the same call.

Retention and access boundaries are stated where they make a question
unanswerable rather than merely harder: X recent search covers 7 days, timelines
cap near 3,200 posts, DMs 30 days, trends have no history at all; Box's event
stream is bounded by account retention; Asana's event stream lasts about 24
hours; and full-archive X search returns an authorisation error on lower tiers,
which is a subscription fact rather than an absence of posts.

## Tests

`EndpointsJsonLoadableTest` passes after every commit. `EndpointJsonQueryTest`
still errors with the pre-existing Spring `NoClassDefFoundError` described in
#4623 — confirmed there against `stylebi-wiz-main` with none of these commits
applied. This module also needs `-am` to compile on this branch, likewise
pre-existing.

## Remaining

zendesk (85 undescribed, genuine configuration), stripe (9), and the Pipedrive
v2 migration flagged in #4623.

### DIFF
--- a/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/asana/endpoints.json
+++ b/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/asana/endpoints.json
@@ -3,47 +3,56 @@
     {
       "name": "Attachment",
       "paged": true,
-      "suffix": "/1.0/attachments/{Attachment}?opt_fields={Returned Fields?}"
+      "suffix": "/1.0/attachments/{Attachment}?opt_fields={Returned Fields?}",
+      "description": "One file attached to a task, project brief or other object, by gid, with its name, size, host and download url."
     },
     {
       "name": "Task Attachments",
       "paged": true,
-      "suffix": "/1.0/tasks/{Task}/attachments?opt_fields={Returned Fields?}"
+      "suffix": "/1.0/attachments?parent={Task}&opt_fields={Returned Fields?}",
+      "description": "The attachments on one task. Requires the task gid as the parent parameter. NOTE THE SHAPE: Asana replaced the nested per-task path with a single attachments endpoint filtered by parent, and parent accepts a project brief or project as well as a task -- so this endpoint is general, and the task case is one use of it."
     },
     {
       "name": "Custom Field",
       "paged": false,
-      "suffix": "/1.0/custom_fields/{Custom Field}?opt_fields={Returned Fields?}"
+      "suffix": "/1.0/custom_fields/{Custom Field}?opt_fields={Returned Fields?}",
+      "description": "One custom field definition by gid, with its name, type (text, number, enum, multi-enum, date, people) and, for enum fields, the option list. Requires a custom field gid."
     },
     {
       "name": "Workspace Custom Fields",
       "paged": true,
-      "suffix": "/1.0/workspaces/{Workspace}/custom_fields?opt_fields={Returned Fields?}"
+      "suffix": "/1.0/workspaces/{Workspace}/custom_fields?opt_fields={Returned Fields?}",
+      "description": "Every custom field defined in one workspace. Requires a workspace id. The account's vocabulary for describing work beyond Asana's built-in fields."
     },
     {
       "name": "Project Custom Fields",
       "paged": true,
-      "suffix": "/1.0/projects/{Project}/custom_field_settings?opt_fields={Returned Fields?}"
+      "suffix": "/1.0/projects/{Project}/custom_field_settings?opt_fields={Returned Fields?}",
+      "description": "The custom fields ENABLED ON one project, as settings records carrying the field, whether it is important, and its parent. Requires a project id. A workspace can define many fields while a project uses a few, so this is what says which columns a project's tasks actually carry -- and therefore which are meaningfully empty rather than simply not applicable."
     },
     {
       "name": "Portfolio Custom Fields",
       "paged": true,
-      "suffix": "/1.0/portfolios/{Portfolio}/custom_field_settings?opt_fields={Returned Fields?}"
+      "suffix": "/1.0/portfolios/{Portfolio}/custom_field_settings?opt_fields={Returned Fields?}",
+      "description": "The custom fields enabled on one portfolio. Requires a portfolio id. The same distinction at the programme level."
     },
     {
       "name": "Events",
       "paged": true,
-      "suffix": "/1.0/events?resource={Resource}&sync={Sync?}&opt_fields={Returned Fields?}"
+      "suffix": "/1.0/events?resource={Resource}&sync={Sync?}&opt_fields={Returned Fields?}",
+      "description": "The change stream for one resource -- a project, task or workspace -- returning what has happened since the last call. Requires a resource gid and a SYNC TOKEN, and the first call has no token: it returns an error carrying a fresh token, which is the intended way to start rather than a failure. Events are retained for about 24 HOURS, so a gap longer than that invalidates the token and the stream must be restarted from a full read. Built for incremental sync, not for history."
     },
     {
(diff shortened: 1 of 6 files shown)


fredaan:
**Correction.** This PR describes a missing leading slash on an endpoint suffix as a broken path — `Alternative Issue Types` here in #4623, `Alert` in #4625. That is wrong, and I've recorded it in #4627.

`URLCreator.combineDataSourceAndQueryUrlFragments` inserts the separator when neither side has one:

```java
   url += "/" + suffix;
}
```

So the endpoint worked before the change and works after it. The edit is stylistic consistency, not a repair, and calling it unreachable overstated the case. A scan across all 65 connectors finds 14 such suffixes in five connectors, all harmless for the same reason; none are being changed.

The other path fixes in this stack are unaffected — each fails for a reason the framework does not paper over (wrong path segments, a missing `/api.xro` segment, a `?` where a `/` belongs, a swapped parameter name, a wrong resource, and two that had moved).

fredaan:
## Review result

Reviewed as part of the stack-wide pass. One thing was found against this PR and has already been corrected in the comment above; everything else below came back clean.

### Structural diff

Diffed against #4623's tip object by object, so anything changed beyond adding a `description` had to be justified:

- **6 files**, all connectors — no Java, no tests
- **1 endpoint-count change**, deliberate: twilio 15 → 25
- **10 added** — the Twilio calls, messages, recordings and conferences this PR is partly about
- **0 removed**
- **7 changed beyond description**:

| connector | endpoints |
|---|---|
| asana | `Task Attachments`, `Organization Teams` |
| box | `Trashed File` |
| square | `Cash Drawer Shift Events` |
| twilio | `Alert`, `Alerts` |
| xero | `Aged Receivables By Contact` |

### The one finding

`twilio / Alert` is in that list of seven, and the correction above is about it: a missing leading slash is **not** a broken path — `URLCreator` inserts the separator — so that edit is stylistic consistency, not a repair. The endpoint worked before and after.

The other six do fail, each for a reason the framework does not paper over: Box's pointed at the wrong resource entirely, Square's had a `?` where a `/` belongs so the remainder became query string, Xero's was missing the `/api.xro` path segment, Twilio's `Alerts` had a parameter name and its label swapped so the end-date filter was silently ignored, and Asana's two had moved rather than broken.

### Coverage

All 6 connectors complete at the stack tip: **303 / 303**.

### Scans

Every systematic scan run over the final state is clean across these 6: no control characters or stray whitespace in a suffix, no double slashes, no non-ASCII in an endpoint name, no unbalanced braces, no duplicate endpoint names, and no lookup pointing at a missing endpoint.

One duplicate-suffix pair remains in xero — `BAS Report` and `GST Report` share `/Reports/{Report ID}`, which is genuine: both are fetched by report id and the id selects between them. It was already documented as such **on both descriptions**, so it needed no follow-up when the alias notes were completed later in the stack.

### Later fixes in the stack

Five further defects surfaced in the stack-wide review — an unterminated placeholder in wordpress, and five descriptions promising a parameter their endpoint does not take (jive, linkedin, influxdb). **None are in this PR's connectors**; all belong to #4629.

### Tests

`EndpointsJsonLoadableTest` passes on this branch and now stands behind these 6 files as well. The module's other 21 errors are the pre-existing Spring `NoClassDefFoundError` described in #4623 — confirmed identical in count on `stylebi-wiz-main` with none of this stack applied.

Nelson-cao (approved):

PR #4626, opened 2026-08-17
fix(write-coordination): version-stamped commit check across the Composer's write paths

larryliang-inetsoft:
## Summary

Implements the fix designed in stylebi-wiz's `2026-08-17-write-coordination-design.md` (roadmap
unit C1) for the lost-update race: a Composer property dialog reads an assembly's state at open
time, holds a client-side snapshot, and commits it wholesale on OK. Meanwhile a plugin write tool
(or another dialog) can write the same live assembly directly and immediately. If the dialog
commits after that, it silently overwrites the other write — no error, just quiet data loss.

Adds a monotonic `writeRevision` counter on `RuntimeViewsheet`. A dialog's read side attaches the
current revision to its outgoing model; the write side sends it back. The commit path refuses with
a named conflict (`common.writeConflict`) when the revision no longer matches, instead of silently
applying a stale patch.

## What's covered

- **21 property dialogs** (Gauge, Chart, TableView, CalcTable, Calendar, Crosstab, GroupContainer,
  Highlight, Hyperlink, Image, Line, Oval, RangeSlider, Rectangle, SelectionContainer,
  SelectionList, SelectionTree, Submit, Tab, Text, ViewsheetObject) via the shared
  `VSObjectPropertyService.editObjectProperty` choke point — one 9-arg overload covers all of them.
- **The condition dialog** (`VSConditionDialogService`) and **the VS Options pane**
  (`ViewsheetPropertyDialogService`) — independent commit paths, each with its own inline check.
- **The binding editor** (`VSBindingService.finishEdit`) — the highest-value fix: its commit
  replaces the *entire* live viewsheet, not just the assembly being bound. New
  `RuntimeViewsheet.baseWriteRevisionAtOpen` captures the base runtime's revision when the binding
  editor's clone opens; a mismatch at Finish throws instead of silently replacing everything.
- **Scripts** (`ScriptEditService`) — a direct live write with no dialog wrapper, so it never moved
  the counter. Now bumps it, closing the gap where a script-embedding property dialog (15 of the
  18 types) could clobber, or be silently clobbered by, a concurrent script write.
- **The plugin's write tools** (`set_assembly_properties`, condition, `set_viewsheet_properties`,
  highlight tools) inherit the check **transitively, with no code changes**: traced
  `AssemblyPropertyService`/`AssemblyConditionService`/`SheetPropertyService`/
  `AssemblyHighlightService` and found each wraps the exact instrumented get/set pairs above,
  atomically, inside one `sessions.mutate` block.

## Known, deliberate gaps (not silently dropped — see the implementation plan for reasoning)

- `HyperlinkDialogService`'s `TableDataVSAssembly` and generic-assembly branches commit through
  `coreLifecycleService.execute(...)` instead of `editObjectProperty` — not covered, follow-up.
- `VSAssemblyInfoHandler.apply` (which `VSConditionDialogService` calls) is itself called from 13
  other classes across binding/dnd/date-comparison — a possibly larger shared choke point, not
  touched here.
- The script spec's `SetScriptTextCommand` push-into-dialog UX (Option 1) is deliberately not
  built: it has no UI seam until pane-scoped pairing (roadmap Lane G, unit G2) ships, and G2 is
  itself gated on this PR landing. The revision bump alone already closes the silent-loss defect
  for scripts universally.

## Test plan

- [x] `./mvnw -o -pl community/core clean compile` — full module, BUILD SUCCESS
- [x] 34/34 targeted Java tests passing across every touched file (7 new/updated test files, 5
      pre-existing tests fixed for the new trailing arguments)
- [ ] Live verification (C1's five test-plan scenarios) — needs B-lane infrastructure, tracked on
      the stylebi-wiz roadmap board, out of reach from this session's environment

See `docs/superpowers/plans/2026-08-17-write-coordination-implementation.md` in stylebi-wiz for
the full phased plan and reasoning behind each decision.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/RuntimeViewsheet.java
+++ b/core/src/main/java/inetsoft/report/composition/RuntimeViewsheet.java
@@ -2119,6 +2119,21 @@ public class RuntimeViewsheet extends RuntimeSheet {
    }
 
    /**
+    * For the binding editor's cloned runtime only: the base runtime's write revision at the
+    * moment editing started. {@code finishEdit} compares this against the base runtime's
+    * CURRENT revision before replacing its whole viewsheet, so a concurrent write elsewhere in
+    * the base viewsheet during the binding session is not silently discarded. {@code null} on
+    * every runtime that isn't a binding-editor clone.
+    */
+   public Integer getBaseWriteRevisionAtOpen() {
+      return baseWriteRevisionAtOpen;
+   }
+
+   public void setBaseWriteRevisionAtOpen(Integer baseWriteRevisionAtOpen) {
+      this.baseWriteRevisionAtOpen = baseWriteRevisionAtOpen;
+   }
+
+   /**
     * Get the base worksheet.
     */
    public RuntimeWorksheet getRuntimeWorksheet() {
@@ -2909,6 +2924,7 @@ public class RuntimeViewsheet extends RuntimeSheet {
    private boolean needRefresh = false; // force refresh
    private int mode; // viewsheet mode
    private transient int writeRevision; // bumped on every dialog-owning commit; see getWriteRevision
+   private transient Integer baseWriteRevisionAtOpen; // binding-editor clones only; see accessor
    private ViewsheetSandbox box; // query sandbox
    private AssetRepository rep; // asset repository
    private String execSessionID; // Execution Session ID used for Auditing
--- a/core/src/main/java/inetsoft/report/composition/RuntimeViewsheet.java
+++ b/core/src/main/java/inetsoft/report/composition/RuntimeViewsheet.java
@@ -2099,6 +2099,26 @@ public class RuntimeViewsheet extends RuntimeSheet {
    }
 
    /**
+    * Get the current write revision. A Composer dialog captures this when it reads its model and
+    * sends it back on commit; a commit whose revision no longer matches is stale and must be
+    * refused rather than silently applied over whatever changed in between. See
+    * {@code 2026-08-17-write-coordination-design.md} for why this exists and
+    * {@code 2026-08-17-write-coordination-implementation.md} for why it is a dedicated counter
+    * rather than the undo-checkpoint index.
+    */
+   public int getWriteRevision() {
+      return writeRevision;
+   }
+
+   /**
+    * Bump the write revision after a dialog-owning commit lands. Returns the new value so a caller
+    * can report it in the same step.
+    */
+   public int bumpWriteRevision() {
+      return ++writeRevision;
+   }
+
+   /**
     * Get the base worksheet.
     */
    public RuntimeWorksheet getRuntimeWorksheet() {
@@ -2888,6 +2908,7 @@ public class RuntimeViewsheet extends RuntimeSheet {
    private boolean preview; // preview flag
    private boolean needRefresh = false; // force refresh
    private int mode; // viewsheet mode
+   private transient int writeRevision; // bumped on every dialog-owning commit; see getWriteRevision
    private ViewsheetSandbox box; // query sandbox
    private AssetRepository rep; // asset repository
    private String execSessionID; // Execution Session ID used for Auditing
--- a/core/src/main/java/inetsoft/web/binding/service/VSBindingService.java
+++ b/core/src/main/java/inetsoft/web/binding/service/VSBindingService.java
@@ -1824,6 +1824,7 @@ public class VSBindingService {
          }
 
          nrvs.setOriginalID(vsId);
+         nrvs.setBaseWriteRevisionAtOpen(rvs.getWriteRevision());
          nrvs.setViewsheet(vs);
          nrvs.setVSTemporaryInfo(wizardTemporaryInfoService.getVSTemporaryInfo(rvs));
          engine.flushRuntimeSheet(id);
@@ -1894,6 +1895,19 @@ public class VSBindingService {
             obox.get().getVariableTable().addAll(nbox.get().getVariableTable());
          }
 
+         // Write coordination: the binding editor edits a CLONE (createRuntimeSheet) and
+         // replaces the base runtime's ENTIRE viewsheet on finish, not just the assembly
+         // being bound -- the largest blast radius of any write path in the Composer. Refuse
+         // rather than silently discard whatever changed in the base viewsheet (any assembly,
+         // via any other write tool) while this binding session was open. See
+         // 2026-08-17-write-coordination-design.md / -implementation.md.
+         Integer expectedRevision = nrvs.getBaseWriteRevisionAtOpen();
+
+         if(expectedRevision != null && expectedRevision != orvs.getWriteRevision()) {
+            throw new MessageException(Catalog.getCatalog().getString(
+               "common.writeConflict", "the viewsheet"));
+         }
+
          updateVSAssemblyBoundAssemblies(nrvs, orvs, assemblyName);
          Viewsheet nvs = nrvs.getViewsheet().clone();
          VSAssembly assembly = nvs.getAssembly(assemblyName);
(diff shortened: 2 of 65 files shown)


larryliang-inetsoft:
## Review — three blocking issues, one of which breaks Apply with no concurrent writer

Reviewed against the PR diff. Findings are ordered by severity; the first two are reproducible without any second writer at all.

### 1. Apply is self-conflicting on all 21 property dialogs (blocking)

Composer property dialogs open as slide-outs with an `<apply-button>`. `ComponentTool.showDialog` wires the apply emitter straight to the same commit handler and leaves the dialog open holding the **same model object**:

```ts
// web/projects/portal/src/app/common/util/component-tool.ts:134-149
applySubscription = modal.componentInstance[applyEmitter].subscribe((v: any) => {
   ...
   onCommit(v);
});
```

So: Apply #1 sends revision N, `editObjectProperty` accepts it and calls `rvs.bumpWriteRevision()` → server is at N+1. The dialog still holds N. Apply #2, or the final OK, arrives stale and is refused with `common.writeConflict` — discarding the user's edits.

**This PR changes 63 files, none of them `.ts` or `.html`.** There is no client-side path that could refresh a dialog's captured revision mid-session, and the server's response to a property commit is assembly-refresh commands rather than a fresh dialog model.

Fix direction: return the new revision on the commit response and update the open dialog's model, or exempt an apply (as opposed to a close) from the check.

### 2. `finishEdit` throws past its own cleanup, wedging the binding session (blocking)

The new conflict check is inserted before the teardown:

```java
if(expectedRevision != null && expectedRevision != orvs.getWriteRevision()) {
   throw new MessageException(Catalog.getCatalog().getString(
      "common.writeConflict", "the viewsheet"));
}

updateVSAssemblyBoundAssemblies(nrvs, orvs, assemblyName);
```

`orvs.setBindingID(null)`, `engine.flushRuntimeSheet(...)` and `engine.closeViewsheet(nid, principal)` all come later in the method. On a conflict the clone runtime `nid` is never closed and the base keeps its `bindingID`, so the Composer still believes a binding session is open. Every retry re-throws at the same point, so the session's work is unrecoverable.

The refusal should still tear the clone down.

### 3. Agent writes do not bump the revision, so the check does not cover them

`bumpWriteRevision()` is called in exactly five places in this PR: `VSBindingService`, `VSObjectPropertyService` (×2), `VSConditionDialogService`, `ViewsheetPropertyDialogService`, `ScriptEditService`.

Every wiz agent write commits through a different seam:

```java
// core/src/main/java/inetsoft/web/wiz/binding/ChartAestheticAgentService.java:112
sessions.mutate(sessionToken, user, (rvs, runtimeId, dispatcher) -> {
```

`ViewsheetSessionService.java` is not among this PR's 63 files, and `bumpWriteRevision()` is introduced by this PR — so `mutate` cannot be bumping it.

Consequence, in the direction that matters most for this product:

1. Human opens a chart property dialog — captures revision N.
2. Agent calls `set_aesthetic_field` → `mutate` → no bump, server still N.
3. Human clicks OK → revision matches → commit applies.
4. **The agent's change is silently overwritten** — exactly the lost update this PR exists to prevent.

The PR description says plugin write tools "inherit the check transitively, no code needed — traced and confirmed". That holds for one of two properties: an agent write routed through an instrumented dialog service would be *refused* when stale. It does not hold for *bumping*, which is what makes a write visible to everyone else's check. Being checked and being protected are different properties, and only the first appears to have been traced.

If covering `ViewsheetSessionService.mutate` is deliberately out of scope, that is a reasonable call — but it should be stated in the PR and on the roadmap, because the write-coordination unit was scoped as governing *every* viewsheet write tool.

### 4. NPE on the refusal path when a dialog also renames the assembly (medium)

`SelectionTreePropertyDialogService.editObjectProperty(...)` now returns silently on a stale revision, but the caller keeps going and looks up the assembly by its **new** name — which was never applied, because the edit was refused. That returns `null` and the following `getVSAssemblyInfo()` NPEs. Even without a rename, `updateSelection(...)` then runs against state the write was explicitly rejected for.

Callers of the 9-arg overload need to learn that the edit was refused — a boolean return or a throw — rather than falling through.

---

Items 1 and 2 reproduce in ordinary single-user Composer use. Item 3 is a scope question that changes what the feature claims to do. Suggest holding this PR; #4631 and #4632 are independent of it (zero file overlap, verified) and can land first.

larryliang-inetsoft:
## Post-merge follow-up: three of the four review items were fixed, one shipped

Re-checked each item from the review above against merged `feature-76074-server-fixes`. Recording the result so nobody re-derives it.

**Fixed before merge:**

- `finishEdit` throwing past its own cleanup — the throw now sits at `VSBindingService:1963`, after `setBindingID(null)` (1952) and `closeViewsheet(nid, principal)` (1959). The clone is torn down before the refusal.
- Agent writes not bumping — `ViewsheetSessionService` now calls `bumpWriteRevision()`, so writes through `mutate` are visible to everyone else's check. That closes the coverage gap; the feature now means what the roadmap scoped it to mean.
- The refusal-path NPE — addressed by *"make `editObjectProperty`'s refusal observable"*.

**Shipped unfixed — filed as #4637:**

The Apply button self-conflict. `grep -r revision web/projects/` is still empty on the merged branch, so nothing refreshes a dialog's captured revision after a commit, while `VSObjectPropertyService` bumps at `:266`/`:361` and refuses on mismatch at `:140`. A user who clicks Apply and then OK loses their edits, with no concurrent writer involved, on all 21 property dialogs plus the condition dialog and VS Options pane.

Full trace and two suggested fixes in #4637. Flagging here because it is the one item from this PR's review that did not land, and because Apply is currently less reliable than it was before this PR.

larryliang-inetsoft:
Thanks for the thorough review — fixed three of the four findings, pushed to this branch:

**Issue 3 (revision-bump scope gap)** — `65e7cb249`. `ViewsheetSessionService.mutate` now bumps `writeRevision` in its `finally` block, so agent writes landing through that general seam (chart binding, aesthetics, table binding, calc layout — anything not going directly through `VSObjectPropertyService`/`VSConditionDialogService`/`ViewsheetPropertyDialogService`) are no longer invisible to the check. Added `ViewsheetSessionServiceTest` covering both the success and mid-mutation-throw cases.

**Issue 2 (`finishEdit` teardown order)** — `2f8e4676c`. Restructured so the conflict check no longer short-circuits before `orvs.setBindingID(null)`, `engine.flushRuntimeSheet(...)` and `engine.closeViewsheet(nid, principal)`. The apply-to-`orvs` step is skipped on conflict, but the teardown always runs now, and the throw moved to the end of the method. Updated `VSBindingServiceFinishEditTest` to assert the teardown happens on the refusal path.

**Issue 4 (silent refusal)** — `3f481c2fe`. `editObjectProperty` now returns `boolean`: `false` for every early exit before the property is actually committed (disposed, write conflict, dependency cycle, rename failure, script failure, missing assembly, trap warning), `true` from the point the info is actually applied onward. Fixed the three call sites that do post-commit work assuming success — `SelectionTreePropertyDialogService` (was NPEing on the never-applied rename), `ChartPropertyDialogService`, and `TabPropertyDialogService` — to check the return value first.

**Issue 1 (Apply self-conflict)** is still open — I'm scoping it now. Will follow up separately since it likely needs an Angular-side change (returning/refreshing the revision on Apply) rather than a Java-only fix.

larryliang-inetsoft:
Follow-up on Issue 1 (Apply self-conflict): investigation found the actual gap is bigger than the reported symptom -- the Angular Composer client never transmits revision in either direction, so the whole write-coordination check has never fired for real browser usage, only for tests and the wiz agent's internal path. Larry's literal "Apply reconflicts with itself" can't happen for the same reason: there's no revision to conflict with. Fixed the actual gap (client now sends revision on OK commits, never on Apply, which makes the self-conflict scenario impossible by construction) in #4640.

PR #4627, opened 2026-08-17
Finish Zendesk and Stripe, and correct a claim about leading slashes

fredaan:
Completes the two connectors left partial. **Stacked on #4625**, which is stacked
on #4623; retarget as those merge.

- **zendesk 278 / 278** (was 193)
- **stripe 113 / 113** (was 101 of 110, plus three endpoints added)

All **14 connectors touched across the three PRs are now completely described**:
1,684 endpoints.

## A correction

Two earlier commits in this stack treated a missing leading slash on a suffix as
a broken path — Jira's `Alternative Issue Types` in #4623, Twilio's `Alert` in
#4625. **That was wrong.** `URLCreator.combineDataSourceAndQueryUrlFragments`
inserts the separator when neither side has one:

```java
   url += "/" + suffix;
}
```

Both endpoints worked before those commits and work after them. The changes are
stylistic consistency, not repairs, and calling them unreachable was overstated.
A scan across all 65 connectors finds 14 such suffixes in five connectors, all
harmless for the same reason; none are being changed.

The other path fixes in this stack are unaffected. Each fails for a reason the
framework does not paper over: GitHub's three had wrong path *segments*, Xero's
was missing `/api.xro` rather than a slash, Square's had a question mark where a
slash belongs so the remainder became query string, Twilio's `Alerts` had a
parameter name and its label swapped, Box's pointed at the wrong resource
entirely, and Asana's two had moved.

## Stripe

The nine endpoints left undescribed turned out to share a reason: all nine are
absent from Stripe's current API description. Six are the legacy Orders and SKUs
model, two are the upcoming-invoice preview — which Stripe reimplemented as a
POST, so a read-only connector cannot reach it at all — and one is usage record
summaries.

Rather than only marking that last one gone, its replacement is added: billing
meters, one meter by id, and meter event summaries. Metered usage is the bridge
between what a customer consumed and what they are charged, and the connector had
no path to it.

Worth knowing before planning around it: event summaries requires `customer`,
`start_time` **and** `end_time`. Usage cannot be swept across an account — it is
one request per customer per meter.

Two removals are recorded as losing data rather than moving it, since naming a
replacement would imply otherwise: the modern Products and Prices model has **no
inventory tracking**, so stock levels went with SKUs; and the forward view of
subscription revenue is simply unavailable here.

Also fixed one real duplicate: `Account Bank Accounts` and `Account Cards` shared
the suffix `/v1/accounts/{id}/external_accounts`, which returns both kinds
together and takes an `object` parameter to pick one. Both names returned the same
mixed list. The correct form was already in the file — the customer-level pair
uses `?object=bank_account` and `?object=card` — so the account-level pair had
simply lost the filter.

## Zendesk

These 85 were the ones I proposed skipping as configuration, after being wrong
once already about which Zendesk endpoints hold data. Going through them
individually rather than by group name found several that are not configuration:

- **Trigger Revisions** is a change history — who altered a rule and when. When
  ticket behaviour changed on a date and nobody remembers why, this is the only
  place with the answer.
- **Target Failures** is operational data. A rule that fires correctly into a dead
  endpoint leaves no mark on any ticket; this table is the mark.
- **Routing Attribute Values for Agent / for Ticket** are the actual skill
  assignments, not definitions — together they explain why a ticket sat
  unassigned while agents were free.
- **Dynamic Content** holds the words customers actually receive; the rules only
  reference it.

For the rest, which genuinely are configuration, the descriptions say what they
explain about the *data*. Triggers and automations set priority, assign groups and
add tags automatically, so many field values on a ticket were put there by a rule
rather than a person — and automations fire on elapsed time, which is why status
changes appear with no actor behind them.

**Account Settings** is flagged as the first thing to read when a table comes back
empty: several endpoints return nothing because the feature behind them is off,
and this is the only place that distinction is visible. Satisfaction ratings
absent because CSAT was never enabled looks identical to satisfaction ratings
absent because nobody responded.

## Remaining

51 connectors untouched, and the Pipedrive v2 migration flagged in #4623. The
duplicate-suffix scan also flagged five pairs in connectors not yet covered —
hubspot (3), campaignmonitor, freshservice, toggl — left for when those are done.

### DIFF
--- a/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/stripe/endpoints.json
+++ b/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/stripe/endpoints.json
@@ -556,9 +556,9 @@
     },
     {
       "name": "Account Bank Accounts",
-      "description": "The external accounts of one connected account -- where its payouts are sent. Requires an account id. Note this returns every external account, bank accounts and cards alike; despite the name it applies no filter, and Account Cards calls exactly the same URL.",
+      "description": "The bank accounts attached to one connected account for payouts, with the bank name, last four digits, currency, country and status. Requires an account id. NOTE THE FILTER: external_accounts holds bank accounts and debit cards together, and object=bank_account is what separates them -- without it the endpoint returns both kinds mixed, which is how the same list can appear under two names.",
       "paged": true,
-      "suffix": "/v1/accounts/{Account ID}/external_accounts"
+      "suffix": "/v1/accounts/{Account ID}/external_accounts?object=bank_account"
     },
     {
       "name": "Account Card",
@@ -568,9 +568,9 @@
     },
     {
       "name": "Account Cards",
-      "description": "The external accounts of one connected account. Requires an account id. Note this returns every external account, cards and bank accounts alike; despite the name it applies no filter, and Account Bank Accounts calls exactly the same URL.",
+      "description": "The debit cards attached to one connected account for payouts, with brand, last four digits, expiry and country. Requires an account id. The object=card filter is what distinguishes this from Account Bank Accounts; the underlying endpoint is the same and returns both kinds unfiltered.",
       "paged": true,
-      "suffix": "/v1/accounts/{Account ID}/external_accounts"
+      "suffix": "/v1/accounts/{Account ID}/external_accounts?object=card"
     },
     {
       "name": "Person",
--- a/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/stripe/endpoints.json
+++ b/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/stripe/endpoints.json
@@ -332,12 +332,14 @@
     {
       "name": "Upcoming Invoice",
       "paged": false,
-      "suffix": "/v1/invoices/upcoming?customer={Customer ID}&subscription={Subscription?}&coupon={Coupon?}"
+      "suffix": "/v1/invoices/upcoming?customer={Customer ID}&subscription={Subscription?}&coupon={Coupon?}",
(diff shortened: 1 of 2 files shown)


fredaan:
## Review result

Reviewed as part of the stack-wide pass. **Nothing was found against this PR** — the checks below all came back clean for its two connectors.

### Structural diff

Diffed both files against #4625's tip object by object, so anything changed beyond adding a `description` had to be justified:

- **2 files changed**, no Java, no tests
- **3 endpoints added**, all deliberate: `Billing Meters`, `Billing Meter`, `Billing Meter Event Summaries` — stripe goes 110 to 113
- **0 removed**
- **2 changed beyond description**: `Account Bank Accounts` and `Account Cards`, which is the `object=` filter fix this PR is partly about

### Coverage

Both connectors are complete at the stack tip:

| connector | described |
|---|---|
| zendesk | 278 / 278 |
| stripe | 113 / 113 |

97 of those were described in this PR — 85 zendesk, 12 stripe.

### Scans

Every systematic scan run over the final state comes back clean for zendesk and stripe: no control characters or stray whitespace in a suffix, no double slashes, no non-ASCII in an endpoint name, no unbalanced braces, no duplicate endpoint names, and no lookup pointing at a missing endpoint.

The one duplicate-suffix pair this PR could have introduced — stripe's `Account Bank Accounts` and `Account Cards` — is precisely what it fixed, and it is no longer a duplicate.

### Later fixes in the stack

Five defects surfaced in the stack-wide review after this PR was opened: an unterminated placeholder in wordpress, and five descriptions promising a parameter their endpoint does not take (jive, linkedin, influxdb). **None of them are in this PR's connectors** — all belong to #4629.

### Tests

`EndpointsJsonLoadableTest` passes on this branch. The module's other 21 errors are the pre-existing Spring `NoClassDefFoundError` described in #4623 — confirmed identical in count on `stylebi-wiz-main` with none of this stack applied.

Nelson-cao (approved):

PR #4628, opened 2026-08-17
Describe thirty more connectors, mostly ones with no published spec

fredaan:
Thirty connectors complete, **620 endpoints**. Stacked on #4627; retarget as the
stack merges.

Running total: **44 of 65 connectors fully described, 2,304 of 4,035 endpoints.**

youtubeanalytics, airtable, gsearchconsole, googlecal, teamdesk, azureblob,
firebase, prometheus, clockify, toggl, adobeanalytics, azuresearch,
helpscoutdocs, mixpanel, lighthouse, remedyforce, seomonitor, sfreports,
chartmogul, quickbooks, freshsales, nicereply, constantcontact, copper,
datadotworld, harvest, zohocrm, freshservice, fortytwomatters, appfigures.

Unlike the first three PRs, most of these have no published OpenAPI description
to align against, so the work was reading the endpoint set as a whole and saying
what shape of API it is.

## Three broken Freshservice paths

Each provable from the file's own inconsistency rather than an external spec:

| endpoint | was | problem |
|---|---|---|
| `Asset CI Type Fields` | `/api/v2/assets_types/{id}/fields` | sibling endpoint spells it `asset_types` |
| `Filtered Releases` | `/api/v2/releases/filter_name=...` | slash where the query string starts — no other suffix does that |
| `Service Category Items` | identical to `Service Items` | returned the whole catalogue; category is a `category_id` filter, not a path |

## Several of these aren't endpoint catalogues at all

**YouTube Analytics is one endpoint** whose report is defined entirely by its
`metrics` and `dimensions`, so a single row stands for hundreds of reports — and
YouTube publishes a fixed set of legal combinations, where an invalid pair errors
rather than returning nothing. **Prometheus** is the same shape: no table to
list, only PromQL to evaluate.

**Zoho CRM has no per-object endpoints.** `Records` is parameterised by module
API name and serves Leads, Contacts, Deals and every custom module alike, so the
first request in any Zoho question is `Modules`.

**Four have no fixed schema at all**, because the account's users built it —
teamdesk, airtable, firebase, remedyforce. Each description names the endpoint
that must be read first, since nothing else is usable without it.

## Access patterns that decide how a question can be phrased

- **Freshsales** has no plain listing. Leads, contacts, accounts and deals come
  only through a saved *view*, which is why each has a companion Filters
  endpoint whose only job is to supply view ids. Its Tasks and Appointments
  endpoints require a filter with no "all" value, so a complete set is several
  requests.
- **Copper** reaches everything through search.
- **data.world** addresses everything by owner and slug, never an opaque id.
- **Box, Asana and X** (earlier PRs) withhold fields by default; **Airtable**
  goes further and keys fields by the column's *display name*, so renaming a
  column in the interface silently breaks every query.

## Values that can't be read without a second table — or any table

Freshservice's ticket columns are numeric codes with **no decoder endpoint at
all**, so the mapping is written into the description (status 2 Open, 3 Pending,
4 Resolved, 5 Closed; priority 1 Low to 4 Urgent). A Nicereply rating is
meaningless until its survey's scale is known — CSAT, CES and NPS share one
table and NPS is not an average, so aggregating across surveys is wrong
regardless of scale. Constant Contact splits campaigns from campaign
*activities* and attaches every statistic to the activity.

## Where two report connectors need flattening

QuickBooks and Salesforce both return nested structures — sections and summary
rows in one, a `factMap` keyed by grouping in the other — where the section a
value sits in is what gives it meaning. Salesforce additionally truncates a
synchronous run at 2,000 detail rows and reports that in an `allData` flag, so a
total taken without checking it is wrong on any large report. QuickBooks'
accounting method is a *parameter*, so the same period returns different numbers
on a cash basis than an accrual one.

## Units and encodings

Clockify returns durations as ISO 8601 period strings (`PT1H30M`); **Toggl
encodes a running entry as a negative duration**, so totalling without filtering
corrupts the result. Harvest is the opposite and worth noting for it: it stamps
billable and cost rates *onto* each entry, so revenue and cost read directly and
a historical entry keeps the rate that applied then.

## Endpoints that exist to explain why other data looks wrong

Prometheus targets show a metric fell because a scrape failed rather than a value
did. Azure Search indexer status is the only place a partially failed index is
visible. Azure Blob's block list shows uncommitted blocks — billed, but in no
blob listing, and the usual explanation for a storage bill exceeding the sum of
its blobs.

## Tests

`EndpointsJsonLoadableTest` passes after every commit. The pre-existing
`EndpointJsonQueryTest` Spring failure and the `-am` requirement are unchanged
from #4623.

## Remaining

21 connectors, 1,731 endpoints — the large ones: jive (258), hubspot (160),
insightly (127), wordpress (99), linkedin (89), activecampaign (86), influxdb
(83), fusebill (75), pipelinecrm (71), surveymonkey (68), chargify (61),
gosquared (60), smartsheet (58), campaignmonitor (58), freshdesk (57), intervals
(57), keap (57), servicenow (56), liveagent (55), chargebee (51), zendesksell
(45). Plus the Pipedrive v2 migration flagged in #4623.

### DIFF
--- a/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/adobeanalytics/endpoints.json
+++ b/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/adobeanalytics/endpoints.json
@@ -1,84 +1,100 @@
 {
-    "endpoints": [
-        {
-            "name": "Caluculated Metrics",
-            "paged": false,
-            "suffix": "/calculatedmetrics?expansion={Additional Fields?}&rsids={RSIDs?}&ownerId={Owner ID?}&filterByIds={IDs?}&locale={Locale?:en_US}&name={Name?}&tagNames={Tage Names?}&includeType={Include Type?:all|shared|templates}"
-        },
-        {
-            "name": "Calculated Metric",
-            "paged": false,
-            "suffix": "/calculatedmetrics/{ID}?expansion={Additional Fields?}&locale={Locale?:en_US}"
-        },
-        {
-            "name": "Calculated Metric Functions",
-            "paged": false,
-            "suffix": "/calculatedmetrics/functions?locale={Locale?:en_US}"
-        },
-        {
-            "name": "Calculated Metric Function",
-            "paged": false,
-            "suffix": "/calculatedmetrics/functions/{ID}?locale={Locale?:en_US}"
-        },
-        {
-            "name": "Report Suites",
-            "paged": false,
-            "suffix": "/collections/suites?expansion={Additional Fields?}&rsids={RSIDs?}&rsidContains{RSID Contains?}"
-        },
-        {
-            "name": "Report Suite",
-            "paged": false,
-            "suffix": "/collections/suites/{RSID}?expansion={Additional Fields?}"
-        },
-        {
-            "name": "Dateranges",
-            "paged": false,
-            "suffix": "/dateranges?expansion={Additional Fields?}&locale={Locale?:en_US}&filterByIds={Daterange IDs}&includeType={Include Type?:all|shared|templates}"
-        },
-        {
-            "name": "Daterange",
-            "paged": false,
-            "suffix": "/dateranges/{Daterange ID}?expansion={Additional Fields?}&locale={Locale?:en_US}"
-        },
-        {
-            "name": "Dimensions",
-            "paged": false,
-            "suffix": "/dimensions?rsid={RSID}&expansion={Additional Fields?}&locale={Locale?:en_US}&segmentable={Segmentable?}&reportable={Reportable?}&classifiable={Classifiable?}"
-        },
-        {
-            "name": "Dimension",
-            "paged": false,
-            "suffix": "/dimensions/{Dimension ID}?rsid={RSID}&expansion={Additional Fields?}&locale={Locale?:en_US}"
-        },
-        {
-            "name": "Metrics",
-            "paged": false,
-            "suffix": "/metrics?rsid={RSID}&expansion={Additional Fields?}&locale={Locale?:en_US}&segmentable={Segmentable?}"
-        },
-        {
-            "name": "Metric",
-            "paged": false,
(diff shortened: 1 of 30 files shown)


fredaan:
## Review result

Reviewed as part of the stack-wide pass. **Nothing was found against this PR** — the checks below all came back clean for its 30 connectors.

### Structural diff

Diffed all 30 files against #4627's tip object by object, so anything changed beyond adding a `description` had to be justified:

- **30 files changed**, no Java, no tests
- **0 endpoints added, 0 removed** — every connector's count is unchanged
- **3 changed beyond description**, all in freshservice and all the path fixes this PR describes: `Filtered Releases`, `Asset CI Type Fields`, `Service Category Items`

No endpoint lost a `lookups`, `paged`, `post`, `bodyTemplate` or `url` property. This PR is otherwise purely additive.

### Coverage

All 30 connectors complete, at the time of this PR and still at the stack tip: **620 / 620**.

### Scans

Every systematic scan run over the final state comes back clean across these 30: no control characters or stray whitespace in a suffix, no double slashes, no non-ASCII in an endpoint name, no unbalanced braces, no duplicate endpoint names, and no lookup pointing at a missing endpoint.

One duplicate-suffix pair sits in this PR's connectors — toggl's `Clients` and `Workspace Clients`, which are genuinely the same endpoint under two names. It was already documented as an alias **on both descriptions**, so it needed no follow-up when the alias notes were completed later in the stack.

### Later fixes in the stack

Five defects surfaced in the stack-wide review after this PR was opened: an unterminated placeholder in wordpress, and five descriptions promising a parameter their endpoint does not take (jive, linkedin, influxdb). **None of them are in this PR's connectors** — all belong to #4629.

### Tests

`EndpointsJsonLoadableTest` passes on this branch, and it now stands behind these 30 files as well: injecting a bogus property into one of them makes it fail and name the connector, the class and the known properties. The module's other 21 errors are the pre-existing Spring `NoClassDefFoundError` described in #4623 — confirmed identical in count on `stylebi-wiz-main` with none of this stack applied.

Nelson-cao (approved):

PR #4629, opened 2026-08-17
Describe the last 21 connectors — all 65 now complete

fredaan:
The final batch: 21 connectors, **1,731 endpoints**. Stacked on #4628.

**All 65 REST connectors are now fully described: 4,035 of 4,035 endpoints.**

zendesksell, chargebee, liveagent, servicenow, campaignmonitor, freshdesk, keap,
intervals, smartsheet, surveymonkey, chargify, gosquared, pipelinecrm, influxdb,
fusebill, activecampaign, wordpress, linkedin, insightly, hubspot, jive.

## Nine more broken paths

Three came from scans that had not been run before, each of which turned out to
have exactly one or two hits across all 65 connectors:

**Literal newlines** — 15 PipelineCRM suffixes had line wrapping saved into the
query string. A newline is illegal in a URI and `URLCreator` hands the assembled
string to `URIBuilder`, whose constructor calls `new URI()`, so every one threw
`URISyntaxException` before a request was made. Only connector affected.

**Non-breaking space in an endpoint NAME** — HubSpot's `Contacts Informations`
carried U+00A0 where a space belongs. Invisible, the only non-ASCII character in
any endpoint name across all 65, and it means the endpoint cannot be matched by
anything that types its name normally — which is what this whole body of work
exists to enable. I hit it myself: my description keyed on the name I could see
failed to apply.

**Stray whitespace in a path** — Jive's `/api/core/v3 /dms/...` (percent-encoded
to `%20`, guaranteed 404) and Zoho's `Variable Groups` as a literal segment.

The rest were single-character or duplicate-path defects: Freshservice's
`assets_types`, a `/` where `?` belongs and a missing `category_id` filter;
Campaign Monitor's `Subscriber lists` pointing at `campaigns.json`; Smartsheet's
`/summ` truncation; SurveyMonkey's `Workgroup Share` with no share id; Insightly's
`/v3.0//Contacts`; HubSpot's `change-log/line_items}`.

### On the whitespace scan

Its first version reported 44 hits and was **wrong**. It cut the query string off
before stripping `{placeholders}`, so every optional path parameter like
`{Bucket ID?}` was split mid-token and its space counted. Stripping placeholders
first leaves exactly two. Worth recording because the false-positive rate looked
like a finding.

## Shapes worth stating once

Several of these APIs are not endpoint catalogues at all, and saying so at the
top of the relevant endpoint is worth more than describing each in isolation:

- **ServiceNow's `Table Records` is the whole platform** — incident,
  change_request, cmdb_ci, any custom table. Everything else here is a
  convenience wrapper. And `sysparm_display_value` decides whether the result is
  readable at all: by default reference fields come back as 32-character sys_ids,
  so `assigned_to` is an opaque string and `state` a bare number.
- **Jive is polymorphic twice over** — documents, discussions, blogs, polls,
  ideas, files and videos are all "content" in one table behind a `type` field;
  spaces, groups and projects are all "places" the same way.
- **InfluxDB has no table to list** — a Flux query defines what comes back, and
  results arrive as *annotated CSV*, not JSON.
- **HubSpot keeps CRM fields in a `properties` object** keyed by internal name,
  with generated names for custom fields, and returns only a small default set —
  so most custom fields are absent unless named, and their absence looks like
  empty data.
- **ActiveCampaign splits custom fields into two tables everywhere** —
  definitions and values, joined by field id. Neither is readable alone.
- **The WordPress connector is the WordPress.com API**, not a self-hosted
  install's REST API. A plain self-hosted WordPress is not reachable through it.

## Values that need a second table, or none exists

Freshdesk and Freshservice both fix their ticket columns as numeric codes and
publish **no decoder endpoint**, so the mappings are written into the
descriptions (status 2 Open, 3 Pending, 4 Resolved, 5 Closed; priority 1–4).
Chargify and Chargebee both hide the involuntary-churn pipeline in a status value
that reads as active — `past_due`, `soft_failure`, `non_renewing` — and put
failed payments only in the transactions table.

## Endpoints that explain why other data looks wrong

GoSquared's blocked-items list changes every number in its trends tables and
appears in none of them. InfluxDB's task runs are where a silently failing task
shows up, since the gaps it leaves look like an absence of activity. Jive's
`deletedObjects` endpoints are the only way a synchronised copy learns content has
gone. Freshdesk's ticket listing defaults to the last 30 days.

## Tests

`EndpointsJsonLoadableTest` passes after every commit — it now stands behind all
65 files. The pre-existing `EndpointJsonQueryTest` Spring failure and the `-am`
build requirement are unchanged from #4623.

## Still outstanding

The **Pipedrive v2 migration** flagged in #4623: 21 of its 108 endpoints are the
core of the CRM and Pipedrive has moved them to `/api/v2`. That changes the base
path, the pagination model and several response shapes, and could not be
validated here without a live account.

### DIFF
--- a/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/pipedrive/PipedriveQuery.java
+++ b/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/pipedrive/PipedriveQuery.java
@@ -105,7 +105,26 @@ public class PipedriveQuery extends EndpointJsonQuery<PipedriveEndpoint> {
 
    @Override
    protected void updatePagination(PipedriveEndpoint endpoint) {
-      if(endpoint.isPaged()) {
+      if(!endpoint.isPaged()) {
+         paginationSpec = PaginationSpec.builder()
+            .type(PaginationType.NONE)
+            .build();
+      }
+      else if(isVersion2(endpoint)) {
+         // v2 pages by an opaque cursor and reports the end with a null next_cursor. It carries
+         // no offset and no total, so the v1 spec below cannot drive it: there is no next_start
+         // to advance and nothing to count towards.
+         paginationSpec = PaginationSpec.builder()
+            .type(PaginationType.ITERATION)
+            .hasNextParam(PaginationParamType.JSON_PATH, "$.additional_data.next_cursor")
+            .pageOffsetParamToRead(PaginationParamType.JSON_PATH, "$.additional_data.next_cursor")
+            .pageOffsetParamToWrite(PaginationParamType.QUERY, "cursor")
+            .maxResultsPerPageParam(PaginationParamType.QUERY, "limit")
+            .maxResultsPerPage(500)
+            .build();
+      }
+      else {
+         // How v1 pages, and how every endpoint still on v1 continues to page.
          paginationSpec = PaginationSpec.builder()
             .type(PaginationType.ITERATION)
             .hasNextParam(PaginationParamType.JSON_PATH, "$.additional_data.pagination.more_items_in_collection")
@@ -115,11 +134,18 @@ public class PipedriveQuery extends EndpointJsonQuery<PipedriveEndpoint> {
             .maxResultsPerPage(500)
             .build();
       }
-      else {
-         paginationSpec = PaginationSpec.builder()
-            .type(PaginationType.NONE)
-            .build();
-      }
+   }
+
+   /**
+    * Whether an endpoint targets Pipedrive's v2 API.
+    *
+    * <p>Read from the suffix rather than from a declared property, because the suffix is what
+    * actually decides which API is called — a flag could disagree with it, and the two failure
+    * modes that produces (v1 paging against v2, or the reverse) are both silent.</p>
+    */
+   private static boolean isVersion2(PipedriveEndpoint endpoint) {
+      final String suffix = endpoint.getSuffix();
+      return suffix != null && suffix.startsWith(VERSION_2_PREFIX);
    }
 
    @Override
@@ -136,4 +162,6 @@ public class PipedriveQuery extends EndpointJsonQuery<PipedriveEndpoint> {
       private final Map<String, PipedriveEndpoint> endpoints =
          Endpoints.load(PipedriveEndpoints.class);
     }
+
+   static final String VERSION_2_PREFIX = "/api/v2/";
 }
--- a/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/activecampaign/endpoints.json
+++ b/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/activecampaign/endpoints.json
@@ -3,102 +3,122 @@
     {
       "name": "Accounts",
       "suffix": "/3/accounts?search={Search?}",
-      "paged": "true"
+      "paged": "true",
+      "description": "The customer organisations contacts belong to, with name, account URL and their created dates. The company-level dimension in ActiveCampaign's CRM."
     },
     {
       "name": "Account",
       "suffix": "/3/accounts/{ID}",
-      "paged": "false"
+      "paged": "false",
+      "description": "One account by id."
     },
     {
       "name": "Account Associations",
       "suffix": "/3/accountContacts?filters[contact]={Contract?}&filters[account]={Account?}",
-      "paged": "true"
+      "paged": "true",
+      "description": "The links between contacts and accounts, each with the contact's job title. How people are attached to companies, which neither record states alone."
     },
     {
       "name": "Account Association",
       "suffix": "/3/accountContacts/{ID}",
-      "paged": "false"
+      "paged": "false",
+      "description": "One account association by id."
     },
     {
       "name": "Custom account fields",
       "suffix": "/3/accountCustomFieldMeta",
-      "paged": "true"
+      "paged": "true",
+      "description": "The custom field definitions for accounts."
     },
     {
       "name": "Custom account field",
       "suffix": "/3/accountCustomFieldMeta/{ID}",
-      "paged": "false"
+      "paged": "false",
+      "description": "One account field definition by id."
     },
     {
       "name": "Custom account field values",
       "suffix": "/3/accountCustomFieldData?filters[customerAccountId]={Customer Account ID?}",
-      "paged": "true"
+      "paged": "true",
+      "description": "The custom field values on accounts, keyed by field id."
     },
     {
       "name": "Custom account field value",
       "suffix": "/3/accountCustomFieldData/{Field ID}",
-      "paged": "false"
+      "paged": "false",
+      "description": "One account field value by id."
     },
     {
       "name": "Addresses",
       "suffix": "/3/addresses",
-      "paged": "true"
+      "paged": "true",
+      "description": "The physical sender addresses used in campaign footers, as anti-spam law requires. Configuration."
     },
     {
       "name": "Address",
       "suffix": "/3/addresses/{ID}",
-      "paged": "false"
+      "paged": "false",
+      "description": "One address by id."
     },
     {
       "name": "Automations",
       "suffix": "/3/automations",
-      "paged": "true"
+      "paged": "true",
+      "description": "The automation workflows defined, with name, status and entry conditions. WHERE MOST EMAIL ACTUALLY GETS SENT in an ActiveCampaign account -- campaign volume alone understates sending, because automations run continuously in response to behaviour."
     },
     {
       "name": "Brandings",
       "suffix": "/3/brandings",
-      "paged": "true"
+      "paged": "true",
+      "description": "The branding configurations available for emails and pages. Presentation."
     },
     {
       "name": "Branding",
       "suffix": "/3/brandings/{ID}",
-      "paged": "false"
+      "paged": "false",
+      "description": "One branding by id."
     },
     {
       "name": "Calendar feeds",
       "suffix": "/3/calendars",
-      "paged": "true"
+      "paged": "true",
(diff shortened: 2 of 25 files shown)


fredaan:
**Pipedrive v2 migration is now included** in this PR (837a2af6), so nothing from #4623's follow-up list is outstanding.

Twenty-eight endpoints moved — deals, persons, organizations, activities, pipelines, stages, products and their searches, plus the seven per-parent listings v2 replaced with filters. Eighty stay on v1, which is where Pipedrive still serves them.

**The blocker was pagination, not the paths.** `PipedriveQuery` hardcoded one spec for every paged endpoint — offset by `start`, read from `additional_data.pagination.next_start`. v2 has neither: it pages by an opaque cursor from `additional_data.next_cursor` and carries no offset and no total. Moved under the old spec a v2 endpoint would have returned page one and stopped, without an error.

Pagination now branches on the **suffix** rather than a declared flag, following `HubSpotQuery`. The suffix is what actually decides which API is called, so a flag could disagree with it — and both failure modes that produces are silent.

Every query string was rebuilt from v2's parameter list rather than carried across, because they differ materially:

| v1 | v2 |
|---|---|
| `user_id` | `owner_id` |
| `first_char` | gone |
| custom fields top-level | named via `custom_fields`, max 15 keys |
| — | `updated_since`/`updated_until` added |
| `field_type`, `field_key`, `exact_match` | `entity_type`, `field`, three-valued `match` |

That last row is why the searches could not be moved by substitution.

Verified rather than assumed: all 28 paths exist as GETs in Pipedrive's published v2 description, **every query parameter on every one of them is defined there**, and the 80 that stayed still validate against v1 — excepting the three already described as removed.

`PipedrivePaginationTest` pins both branches and, more usefully, the invariant that every paged endpoint pages the way its path demands — checked across all 108 in both directions, not on a sample. Confirmed it fails when an endpoint is moved back to a v1 path.

fredaan:
## Review of the whole stack

Went back over all five PRs rather than trusting the commits. Three things came out of it, all now fixed in this PR.

### Structural diff — nothing unaccounted for

Compared every `endpoints.json` against `stylebi-wiz-main` object by object, so anything changed beyond adding a `description` had to be justified by name:

- **4 endpoint-count changes**, all deliberate: github +12, jira +2, stripe +3, twilio +10
- **28 added / 1 removed** — the removal and one addition are the same HubSpot endpoint, renamed off its non-breaking space; the other 27 match the count deltas exactly
- **73 changed beyond description**, and all 73 are the path fixes and the Pipedrive migration, group by group

No endpoint lost a `lookups`, `paged`, `post`, `bodyTemplate`, `url` or `pageType` property. Nothing unintended crept in.

### Two defects the earlier scans could not have found

**An unterminated placeholder** in WordPress — the check for it did not exist until this review:

```
&unit={Unit?:daily|monthly|cumulative|&after=...
```

The brace closes with a pipe, so the placeholder swallows the rest of the query string and the `after` and `before` parameters could never have been supplied. Pre-existing, unchanged from base, and the only unbalanced suffix across all 65 connectors.

**Five descriptions promising a parameter their endpoint has no slot for** — wording carried over from a sibling that does take one. Jive's `placeTopics`, `slides` and `vitals` are community-wide listings; LinkedIn's `Creatives` takes the account as search criteria; InfluxDB's `Analyze Flux Query` is `post=true` and takes the query in the body.

That check's first version reported **134 hits and was wrong** — it read only the path, so every endpoint taking its parameter in the query string was flagged. Reading the whole suffix leaves seven, two of which were "Requires the *feature* to be enabled".

### Alias notes completed

Five suffixes are shared by two names, all deliberate. Each pair already said so, but on only one of the two descriptions in three cases — the wrong half to rely on when retrieval can surface either. Both sides now say it.

### Final state

| scan | result |
|---|---|
| coverage | **4,035 / 4,035** |
| control characters in a suffix | clean |
| stray whitespace in a path | clean |
| double slashes | clean |
| non-ASCII in an endpoint name | clean |
| unbalanced braces | clean |
| duplicate endpoint names | clean |
| lookups targeting a missing endpoint | clean |
| duplicate suffixes | 5, all documented as aliases on both sides |

### Tests

Ran the module's whole suite on this branch and on `stylebi-wiz-main`:

| | base | this stack |
|---|---|---|
| tests run | 43 | 54 |
| **errors** | **21** | **21** |

Same 21, none new — they are the pre-existing Spring `NoClassDefFoundError` described in #4623, spread across XML parsing, `URLCreator` and the pagination strategies. The 11 added tests all pass.

Also confirmed `EndpointsJsonLoadableTest` is not vacuous: injecting a bogus property into one file makes it fail and name the connector, the class and the known properties.

fredaan:
## Per-PR breakdown

Computed from git by diffing each PR's range, not from the commit messages — four connectors span more than one PR, so a naive sum double-counts them.

Totals: 1284 + 303 + 97 + 620 + 1731 = **4,035**.

### #4623 — 8 connectors, 1,284 endpoints

| connector | total | described here |
|---|---:|---:|
| zendesk | 278 | 193 |
| gitlab | 277 | 277 |
| github | 191 | 191 |
| jira | 154 | 154 |
| sendgrid | 133 | 133 |
| mailchimp | 127 | 127 |
| pipedrive | 108 | 108 |
| stripe | 110 | 101 |

zendesk and stripe were deliberately left partial and finished in #4627.

### #4625 — 6 connectors, 303 endpoints

| connector | total | described here |
|---|---:|---:|
| box | 82 | 82 |
| asana | 65 | 65 |
| xero | 54 | 54 |
| twitter | 47 | 47 |
| square | 30 | 30 |
| twilio | 25 | 25 |

### #4627 — 2 connectors, 97 endpoints

| connector | total | described here | already described |
|---|---:|---:|---:|
| zendesk | 278 | 85 | 193 |
| stripe | 113 | 12 | 101 |

stripe's total rises from 110 to 113: this PR adds the three Billing Meters endpoints.

### #4628 — 30 connectors, 620 endpoints

All 30 complete: freshservice 44, harvest 41, appfigures 39, fortytwomatters 35, zohocrm 35, datadotworld 32, constantcontact 29, copper 29, nicereply 28, freshsales 27, quickbooks 27, chartmogul 23, seomonitor 21, sfreports 21, toggl 19, lighthouse 18, adobeanalytics 16, azuresearch 16, helpscoutdocs 16, mixpanel 16, clockify 15, prometheus 14, remedyforce 12, azureblob 10, firebase 10, googlecal 9, teamdesk 9, gsearchconsole 6, airtable 2, youtubeanalytics 1.

### #4629 — 23 connectors touched, 1,731 endpoints

21 complete: jive 258, hubspot 160, insightly 127, wordpress 99, linkedin 89, activecampaign 86, influxdb 83, fusebill 75, pipelinecrm 71, surveymonkey 68, chargify 61, gosquared 60, campaignmonitor 58, smartsheet 58, freshdesk 57, intervals 57, keap 57, servicenow 56, liveagent 55, chargebee 51, zendesksell 45.

Two more were changed without new descriptions:

| connector | described here | what changed |
|---|---:|---|
| pipedrive | 0 | 28 endpoints migrated to API v2 — paths rebuilt, descriptions rewritten |
| zohocrm | 0 | one path fix, `Variable Groups` to `variable_groups` |

### Connectors spanning more than one PR

- **zendesk**, **stripe** — started in #4623, finished in #4627
- **pipedrive** — described in #4623, migrated to v2 in #4629
- **zohocrm** — described in #4628, path fixed in #4629

So 8 + 6 + 2 + 30 + 23 = 69 PR-touches across **65 distinct connectors**.

### Nothing outside these 65

The test for "does this connector need an endpoints.json" is whether its query class extends `EndpointJsonQuery` or `EndpointXMLQuery`. Every connector that does has one, and all 65 are described.

`inetsoft-rest` holds 69 datasource packages; the four without an endpoints.json should not have one:

- **salesforce** — an abstract base class only, shared by `remedyforce` and `sfreports`, both of which are described
- **graphql**, **monday**, **shopify** — GraphQL connectors that query by request body; the word "Endpoint" appears zero times in their Java

No endpoints.json is an orphan either — every one has a matching Java package.

fredaan:
## Correction to the per-PR breakdown

The stack has been rebased onto `stylebi-wiz-main-rebase` — #4623 was built on the wrong base, and the two lines have diverged too far for a retarget to be safe. Details on #4623.

Content is unchanged: all 65 `endpoints.json` files and every Java file are byte-identical to the pre-rebase tip, coverage is still **4,035 / 4,035**, and the three tests pass.

But **#4623's figures change**, because the new base already contains #4604 — the first branch of this work, before the base mix-up — so its two commits were dropped by the rebase as already upstream. Verified by patch-id, not by assumption.

### #4623 — corrected: 7 connectors, 1,183 endpoints

| connector | total | described here |
|---|---:|---:|
| gitlab | 277 | 277 |
| zendesk | 278 | 193 |
| github | 191 | 191 |
| jira | 154 | 154 |
| sendgrid | 133 | 133 |
| mailchimp | 127 | 127 |
| pipedrive | 108 | 108 |

**stripe is no longer in this PR.** Its 101/110 arrived with #4604 and is on the base; #4627 still adds the last 12 and the three Billing Meters endpoints, taking it to 113/113.

### Corrected totals

| PR | connectors | endpoints described |
|---|---:|---:|
| #4623 | 7 | 1,183 |
| #4625 | 6 | 303 |
| #4627 | 2 | 97 |
| #4628 | 30 | 620 |
| #4629 | 23 | 1,731 |
| | | **3,934** |

Plus the 101 already on the base via #4604 = **4,035**, unchanged.

The connector count is unaffected: **65 distinct connectors**, since stripe is still touched by #4627. Everything else in the breakdown above — the six connectors spanning more than one PR, the four `inetsoft-rest` packages that correctly have no `endpoints.json`, and the reasoning for why 65 is the complete set — stands as posted.

Nelson-cao (approved):

PR #4631, opened 2026-08-17
D1+D2: highlight fall-forward fix + node channels for relation charts

larryliang-inetsoft:
## Summary
Two independently-diagnosed items from the composer automation roadmap's Lane D (`docs/superpowers/specs/2026-08-13-composer-automation-roadmap.md`), landed together since both touch `inetsoft/web/wiz`:

- **D1** — `set_highlight`/`list_highlights`/`delete_highlight` landed on a table's header cell and refused whenever the caller omitted `row`/`col`. `AssemblyHighlightService.read()`'s fall-forward-to-first-data-cell logic is gated on `region == null`, but the controller never actually produced `null` — both `HighlightRequest.region()` and the `listHighlights` GET handler unconditionally built `new Region(row, col, ...)`, and `Region`'s own compact constructor normalizes a null row/col to 0. So the fall-forward was dead code on every real request, caught only by a unit test that called the service directly with a literal `null` — bypassing the controller that never produces one. Fixed by only constructing a `Region` when the caller actually named `row` or `col`.
- **D2** — wires `node-color`/`node-size` (the last uncovered piece of spec 2c's aesthetic-frame taxonomy) through the existing channel vocabulary, gated to relation charts (network, tree, chord) only — `ChartBindingModel` already carries the four node fields, this was purely vocabulary wiring.

## Test plan
- [x] `ViewsheetAssemblyAgentControllerTest` — 4 new tests pin the null-region / explicit-region(0,0) distinction (13/13 pass)
- [x] `AssemblyHighlightServiceTest` — unaffected, still 22/22
- [x] `AestheticChannelsTest`/`ChartAestheticMutatorTest` (26/26), `ChartAestheticAgentServiceTest` (14/14), `VisualFrameAliasesTest` (42/42) — new node-channel coverage plus full existing suites green
- [x] Full `inetsoft.web.wiz.**` sweep (156 test classes) — 0 failures, 0 errors
- [x] Companion plugin PR in `stylebi-wiz` wires the same channels through `set_aesthetic_field`/`clear_aesthetic_field`/`set_visual_frame`

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/AestheticChannels.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/AestheticChannels.java
@@ -28,18 +28,27 @@ import java.util.List;
  * {@code texture} have frames but no field binding, so a caller asking to bind a field to
  * {@code line} gets told why instead of a shrug.
  *
- * <p>Node channels belong to relation charts and arrive with spec 2c Phase 3.
+ * <p>Node channels belong to relation charts (spec 2c Phase 3): {@code node-color} and
+ * {@code node-size} each take both a field binding and a frame, exactly like {@code color}/
+ * {@code size} — the model reuses the same {@code ColorFrameModel}/{@code SizeFrameModel} types,
+ * just on a second pair of properties ({@code nodeColorField}/{@code nodeSizeField}). They are
+ * kept out of {@link #FIELD_CHANNELS}/{@link #FRAME_CHANNELS} rather than merged in: the fields
+ * exist on {@code ChartBindingModel} for every chart type, but only render on a relation chart,
+ * so a caller working a bar or line chart must never be told they apply.
  */
 public final class AestheticChannels {
-   /** Channels that accept a field binding. */
+   /** Channels that accept a field binding, on every chart type. */
    public static final List<String> FIELD_CHANNELS = List.of("color", "shape", "size", "text");
 
-   /** Channels that accept a visual frame. */
+   /** Channels that accept a visual frame, on every chart type. */
    public static final List<String> FRAME_CHANNELS =
       List.of("color", "shape", "size", "line", "texture");
 
    /** Frame channels that can be written. Every frame channel is supported since 2c Phase 2. */
    public static final List<String> SUPPORTED_FRAME_CHANNELS = FRAME_CHANNELS;
+
+   /** Field+frame channels valid only when the target chart is a relation type. */
+   public static final List<String> NODE_CHANNELS = List.of("node-color", "node-size");
 
    private AestheticChannels() {
    }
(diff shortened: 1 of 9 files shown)


larryliang-inetsoft:
## Review — D2 looks clean; D1 introduces a second instance of the bug it fixes

### D1: the new `highlightRegion` helper drops `colName` (blocking)

```java
private static AssemblyHighlightService.Region highlightRegion(Integer row, Integer col,
                                                               String colName)
{
   return row == null && col == null ? null
                                     : new Region(row, col, colName, false, false);
}
```

`colName` is a **standalone address**, not a qualifier on `row`/`col`. The plugin advertises it that way and appends it independently:

```ts
// plugin/composer/src/tools/highlightTools.ts:37,49
colName: { type: "string", description: "Column name, for a table cell." },
...
if (args.colName) parts.push(`colName=${encodeURIComponent(args.colName)}`);
```

So a caller that names only `colName` — the documented way to address a chart measure or a table column — now gets `null`. `AssemblyHighlightService.read()` substitutes `Region.whole()` with `colName == null`, `HighlightDialogService` then resolves no measure, and the call is refused for exposing no highlightable fields.

Before this change the same request produced `Region(0, 0, "Sum(Sales)")` via the compact constructor's null-normalization, and worked.

The irony is worth stating because it is the same mistake in mirror image: this PR correctly identifies that building a `Region` unconditionally *collapsed the "no location named" signal before it arrived*, and fixes it by collapsing a different signal — `colName` — before it arrives.

Fix: collapse to `null` only when `row`, `col` **and** `colName` are all absent.

```java
return row == null && col == null && colName == null
   ? null : new Region(row, col, colName, false, false);
```

Worth a test for the `colName`-only case specifically. The existing 4 controller tests pin the row/col cases, which is why this passed.

### D1 root-cause writeup: this is the good part

For anyone reading later — the diagnosis in this PR is more precise than the issue it was filed against, and the correction is worth preserving. The fall-forward in `AssemblyHighlightService.read()` was never broken. `Region`'s compact constructor normalizes a null row/col to 0, so the controller could not produce the `null` the fall-forward tested for; it was dead code on every real request. That is only provable by calling the service directly with a literal `null`, bypassing the controller — which is why the original green suite proved nothing.

### D2: no findings

`nodeColorField`/`nodeSizeField`/`nodeColorFrame`/`nodeSizeFrame` already existed on `ChartBindingModel` with preservation already covered, so this is vocabulary wiring rather than model work. Resolving the chart once to check `GraphTypes.isRelation` before validating — so an invalid channel costs no checkpoint — is the right shape.

One inconsistency with the PR description rather than the code: it states that `list_aesthetic_options` and `get_chart_aesthetics` descriptions were "updated to mention the two new channels and their relation-chart-only scope". Neither tool factory is touched in stylebi-wiz#1636. Since the backend deliberately keeps `nodeChannels` out of the unconditional `fieldChannels`/`frameChannels`, the two discovery tools an agent is told to call first still never mention node channels. Either make the change or drop the claim.

PR #4632, opened 2026-08-17
fix(chart-binding): add set_chart_separate_status

larryliang-inetsoft:
## Summary

Closes the `separate` gap identified in stylebi-wiz's `2026-08-17-chart-options-2b-p3-design.md`
(roadmap unit C2, part of F2): `set_chart_type`'s own `separate` parameter is a silent no-op
unless `multi` also changes in the same call — `ChangeChartTypeService.handleMulti` only reaches
`ChangeSeparateStatusService` when `omulti != nmulti`. A caller toggling separate on an
already-multi-style chart got `ok: true` and no change, with the field's own description giving
no hint it was gated on something else.

## What changed

- New `ChartBindingService.setSeparateStatus`, calling `ChangeSeparateStatusService` directly (the
  same endpoint the Composer's own separate-status toggle uses) — no dependency on a type change.
  `multi` is read from the chart's current `VSChartInfo.isMultiStyles()`, so the caller doesn't
  have to know or restate it.
- New endpoint `POST /api/wiz/v1/agent/binding/{sessionToken}/chart/separate-status`.
- `set_chart_type`'s `separate` field description now names the gate and points to the new tool.

## Test plan

- [x] `./mvnw -o -pl community/core compile` clean
- [x] 6/6 `ChartBindingServiceTest` passing (1 new)
- [x] 3/3 `BindingAgentControllerTest` unaffected

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
@@ -119,6 +119,7 @@ public class BindingAgentController {
    public record ChartTypeRequest(String assembly, Integer type, Boolean multi,
                                   Boolean stackMeasures, Boolean separate) {}
    public record SwapRequest(String assembly) {}
+   public record SeparateStatusRequest(String assembly, Boolean separate) {}
 
    @PostMapping("/api/wiz/v1/agent/binding/{sessionToken}/chart/shelf")
    public void setChartShelf(@PathVariable String sessionToken,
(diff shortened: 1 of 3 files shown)


larryliang-inetsoft:
## Review — one silent-success case, and it is the class this PR set out to close

`set_chart_separate_status` reports success for a state the server then forces the other way.

`ChangeSeparateStatusService.changeSeparateStatus` computes:

```java
separated = event.isSeparate() || GraphTypes.isTreemap(...) || isMekko(...) || isScatteredContour(...)
```

So calling the new tool with `separate: false` on a treemap, sunburst, circle-packing, icicle, mekko or scattered-contour chart leaves the graph separated, while the plugin returns `ok: true` and `"Set Chart1 to merged graphs."`

That is a well-formed call accepted, producing a plausible-but-wrong result rather than an error — the same class this PR set out to fix for `set_chart_type`, arriving through the new tool.

Two acceptable fixes:

- read `isSeparatedGraph()` back after the change and report the **actual** state, or
- refuse the request up front, naming the chart type and why it cannot be merged.

The second is preferable if the forcing is genuinely unconditional for those types: an agent that is told "no" learns something, whereas an agent told "yes" when the answer is no will build on a false premise.

Not blocking on its own, but worth fixing here rather than filing, since the tool is new in this PR and has no callers yet.

larryliang-inetsoft:
Fixed in bc95f4620: refuses up front (naming the chart type) when `separate: false` is requested on a treemap/sunburst/circle-packing/icicle/mekko/scattered-contour chart, per your preferred fix — an agent told "no" learns something, one told a false "yes" doesn't. `separate: true` on those types still succeeds since it agrees with what the server would force anyway. 8/8 tests passing (2 new: refuses-and-never-calls-the-endpoint, allows-the-agreeing-request).

PR #4634, opened 2026-08-17
Bug #75998, scope-guard permission transfer on folder rename/delete

ngovus-inetsoft:
## Issue

Reported: after a user renames or deletes a private folder that happens to share a name with a global repository folder, the *global* folder's permission is reset to "Derive Permissions from Parent".

Repro:
1. Create user `user0`, give it EM access and READ on the global `Examples` folder.
2. As `user0`, create an `Examples` folder under `User Private Assets/user0`.
3. Rename or delete that private `Examples` folder.
4. As admin, open Content and expand the repository — global `Examples` now shows "Derive Permissions from Parent".

## Root cause

Permission grants are keyed by `(ResourceType, path, orgID)` — no owner. A private folder's owner lives on the `AssetEntry` (`scope = USER_SCOPE`, `user = user0`), not in its path, so `user0`'s private `Examples` and the global `Examples` collapse onto the same permission key. Private assets never consult that table (`checkAssetPermission0` short-circuits on `USER_SCOPE`), so any path-keyed permission write made on behalf of a private folder necessarily hits the global folder's ACL.

`RepletEngine` guards for this everywhere it maintains permissions (`changeFolder0`, `changeSheet0`, `removeFolder0`, `removeSheet0` all test `GLOBAL_SCOPE`; the registry event handlers test `!SUtil.isMyReport(name)`). `AbstractAssetEngine.updatePermission()` — added for Bug #70982 so a moved global folder keeps its permissions — did not, and `changeFolder0` calls it for every scope:

```java
Permission oldPermission = securityEngine.getPermission(type, oentry.getPath());  // global "Examples" grant
securityEngine.removePermission(type, oentry.getPath());                          // global ACL cleared
securityEngine.setPermission(type, nentry.getPath(), oldPermission);              // grant parked elsewhere
```

Both reported steps reach it through the same registry rename (delete is a rename into `Recycle Bin/<uuid>`):

```
EM rename                                EM delete
RepletRegistryService                    RecycleUtils
  .updateRepositoryFolder                  .moveRepositoryFolderToRecycleBin
        └──> RepletRegistry.changeFolder("My Dashboards/Examples", <target>)
                     └──> RENAME_FOLDER_EVENT
                             └──> RepletEngine strips the "My Dashboards/" prefix and re-issues
                                  super.changeFolder(USER_SCOPE, "Examples" -> <target>)
                                     └──> changeFolder0 -> updatePermission
```

Private worksheet folders hit the same path under `ResourceType.ASSET`.

## Fix

`AbstractAssetEngine.updatePermission()` now follows the same scope rule as the rest of the code:

- Return early when the source entry is not `GLOBAL_SCOPE` — private and report-scope folders own no path-keyed permission, so nothing is read, moved, or cleared. This is the bug fix.
- When a global folder moves *out* of the global repository, remove the old grant but do not copy it onto the private path, where it would silently become a different folder's ACL.

`GLOBAL_SCOPE -> GLOBAL_SCOPE` renames and moves — the Bug #70982 case — are unchanged.

## Side effects traced

- The removal-only branch overlaps `RepletEngine.changeFolder0`, which already removes the old grant for non-repository global folders; a second `removePermission` on a cleared key is a no-op. For repository folders `RepletEngine` skips that block, which is where the grant used to leak into the private namespace.
- `REPORT_SCOPE` (report-embedded) entries no longer touch the permission table; they never had an entry in it.

## Testing

`core` compiles clean. No unit test added: `updatePermission` is private and only reachable through `changeFolder0`, which needs a live `IndexedStorage` plus the static `SecurityEngine`. Verified by reasoning through the call chain; the reporter's four steps are the practical check.

## Related, not changed here

`RecycleUtils.moveRepositoryFolderToRecycleBin` calls `getEntryPermission(REPORT, "My Dashboards/<folder>")`, which walks up to the global root when no explicit grant exists and then writes that inherited permission to the recycle-bin path. It creates stray grants on private deletes rather than resetting anything, so it is cosmetic — but it is the same scope-blindness and could be cleaned up separately.

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/AbstractAssetEngine.java
+++ b/core/src/main/java/inetsoft/uql/asset/AbstractAssetEngine.java
@@ -2100,6 +2100,14 @@ public abstract class AbstractAssetEngine implements AssetRepository, AutoClosea
          return;
       }
 
+      // Permissions are only keyed by path for global-scope assets. A user's private assets
+      // live in the same path namespace (the owner is carried by the entry, not the path), so
+      // transferring the permission of a private folder would move/clear the permission of the
+      // same-named folder in the global repository.
+      if(oentry.getScope() != GLOBAL_SCOPE) {
+         return;
+      }
+
       SecurityEngine securityEngine = SecurityEngine.getSecurity();
       ResourceType type = getAssetResourceType(oentry);
       Permission oldPermission = securityEngine.getPermission(type, oentry.getPath());
@@ -2109,7 +2117,12 @@ public abstract class AbstractAssetEngine implements AssetRepository, AutoClosea
       }
 
       securityEngine.removePermission(type, oentry.getPath());
-      securityEngine.setPermission(type, nentry.getPath(), oldPermission);
+
+      // moving out of the global repository (e.g. into a user's private assets); the old
+      // permission no longer applies to any path and must not be copied to the private path
+      if(nentry.getScope() == GLOBAL_SCOPE) {
+         securityEngine.setPermission(type, nentry.getPath(), oldPermission);
+      }
    }
 
    /**


jshobe-inetsoft (changes requested):
### Code Review: Bug #75998 — scope-guard permission transfer on folder rename/delete

**Root cause verified.** `updatePermission(oentry, nentry)` is called unconditionally from `changeFolder0` (`AbstractAssetEngine.java:2028`), and before this fix it read/removed/re-wrote the path-keyed permission for *any* scope. Since folder permissions are keyed by `(ResourceType, path, orgID)` with no owner component, a private folder (`USER_SCOPE`) sharing a path segment with a global folder (e.g. `Examples`) collided on the same permission key. This matches the existing scope-guard pattern used elsewhere in the same file (e.g. `AbstractAssetEngine.java:2895` `tscope != GLOBAL_SCOPE`), so the fix is consistent with established conventions rather than a one-off special case.

**Fix correctness.**
- The new early return (`oentry.getScope() != GLOBAL_SCOPE`) is the only caller-path guard needed — `updatePermission` has exactly one call site, confirmed via search — and correctly excludes both `USER_SCOPE` and `REPORT_SCOPE`, matching the PR description's claim that `REPORT_SCOPE` entries never had permission-table entries either.
- The second guard (`nentry.getScope() == GLOBAL_SCOPE` before re-`setPermission`) correctly prevents a permission that's cleared from its old global path from being planted on a non-global destination path, while still allowing the pre-existing `GLOBAL_SCOPE -> GLOBAL_SCOPE` (Bug #70982) behavior to pass through unchanged.
- The overlap with `RepletEngine.changeFolder0`'s own `GLOBAL_SCOPE` guard (a second `removePermission` on an already-cleared key) is a harmless no-op as described.

**Test coverage.** No unit test was added. The PR description explains `updatePermission` is private and only reachable through `changeFolder0`, which needs a live `IndexedStorage` + static `SecurityEngine` singleton — a real gap, but the reasoning is sound and this is a narrow, low-risk two-line guard change rather than new logic, so I would not block solely on missing test coverage.

**Regressions / backward compatibility.** No behavior change for `GLOBAL_SCOPE -> GLOBAL_SCOPE` moves. Private/report-scope folder renames/deletes now correctly skip permission-table writes entirely, which is strictly the intended fix — no other code path was found reading permissions for non-global asset scopes.

**Security.** No new attack surface; this narrows (not widens) what gets written to the permission store.

**Blocking issue: failing checks.** The `build` check for the latest commit is currently failing (job cancelled/failed, not related to this diff's own correctness per the analysis above, but it must pass before merge), and the automated `claude-review` CI job also failed to complete (infrastructure error, not a code finding). Per review criteria, a PR cannot be approved while checks are failing — please investigate/re-run the build.

**Verdict:** No blocking issues found in the code itself — the fix is correct, minimal, and consistent with existing scope-handling conventions. Requesting changes only because the `build` check is currently failing on the latest commit; please re-run/fix the build so this can be re-reviewed for merge.

jshobe-inetsoft (approved):

PR #4635, opened 2026-08-17
Bug #76022, keep bottom-tab children flush with the tab strip in layouts

ngovus-inetsoft:
## Issue

`<tab>` when bottom tabs is true, the radio button position is wrong in device layout.

Repro: open viewsheet `tab` in the composer, select layout `a`. The radio button is not attached to the tab strip — its bottom should be aligned with the tab. Preview shows the same defect.

## Root cause

The tab strip and its children were positioned by two different rules, so they only lined up for the *tallest* child.

In the reported asset, `Tab1` (`bottomTabs=true`, 24px) contains `Gauge1` (140px) and `RadioButton1` (40px). Layout `a` stores one entry for `Tab1`: `(80,80)` size `200x164` (24 strip + 140 tallest child); the children have no stored positions and are derived.

- The **tab strip** is drawn at the bottom of the content band — the `padding-top` from `bottomTabsChildHeight` in `layout-object.component.ts`, and `npos.y + contentHeight` on the server.
- The **children** were all placed at the *top* of that band (`VSLayoutService` set every child model to the layout position; `AbstractLayout.applyTab` set every child to `npos.y`).

For the tallest child the two coincide, so the bug was invisible. `RadioButton1` ends at `120` while the strip sits at `220` — the ~100px gap in the report.

This regressed in Bug #74523, which changed the stored position to mean "visual top of the tab area" and dropped the per-child offset, relying on `VSEventUtil.applyTabScale` to re-flush children. That only runs when scale-to-screen applies at runtime, and never in the composer layout pane.

## Fix

Bottom-align each child within the content band so every child's bottom edge meets the top of the strip — the design-time rule the rest of the code already follows (`TabVSAssemblyInfo.repositionForBottomTabs`: bar fixed, children flush to it). The visual-top position convention from #74523 is preserved.

- `AbstractLayout.applyTab` — layout apply, preview, export: child y = `npos.y + max(0, contentHeight - childHeight)`
- `VSLayoutService.createObjectModel` — composer layout pane: child top = `layoutPos.y + (maxChildHeight - childHeight)`

The band height in `VSLayoutService` is deliberately the max *child model* height rather than `layoutHeight - tabHeight`, because that is the same number the template uses for its `padding-top`; if the two disagreed the gap would move instead of closing.

Export paths that shift a shrunk table flush to the bar (`applyShrunkBottomTabsShift`, `VsToReportConverter`) key off the assembly's own declared-vs-rendered height, not tab geometry, so they do not double-shift.

## Tests

Both fixes are covered by tests confirmed to fail without them:

- New `AbstractLayoutTest` reproduces the asset (140/40 children, 24px bar) — without the fix: `expected: <180> but was: <80>`, exactly the reported gap. Also asserts top-tabs positioning is unchanged.
- `VSLayoutServiceTest` gains a #76022 case and an updated multi-child expectation — without the fix: `<360>` vs `<300>` and `<180>` vs `<80>`.

Both classes were missing the Spring test wiring that `@SreeHome` needs (`SpringExtension` + `BaseTestConfiguration`), so they errored out before reaching any assertion; that wiring plus `@Tag("core")` is added so they actually run under the pom's tag filter.

`./mvnw test -pl core -Dtest=VSLayoutServiceTest,AbstractLayoutTest` → 8/8 pass.

Not yet verified in a running server — composer view, preview, and a PDF/Excel export of a bottom-tabs layout are worth a visual check.

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/vslayout/AbstractLayout.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/vslayout/AbstractLayout.java
@@ -426,8 +426,12 @@ public abstract class AbstractLayout implements AssetObject {
             childSize =
                new Dimension((int) (childSize.width * scaleRadio.x),
                              (int) (childSize.height * scaleRadio.y));
+            // for bottom tabs the tab bar is placed at the end of the content
+            // area, so bottom-align each child within it. Only the tallest
+            // child fills contentHeight; shorter ones would otherwise float
+            // detached above the tab bar.
             Point childPos = bottomTabs ?
-               new Point(npos.x, npos.y) :
+               new Point(npos.x, npos.y + Math.max(0, contentHeight - childSize.height)) :
                new Point(npos.x, npos.y + tabSize.height);
             applyAssembly(child, childPos, childSize);
          }
--- a/core/src/main/java/inetsoft/web/composer/vs/controller/VSLayoutService.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/controller/VSLayoutService.java
@@ -434,15 +434,22 @@ public class VSLayoutService {
             getChildAssemblies((ContainerVSAssembly) assembly, rvs,
                                childModels, objectModelService);
 
-            // for bottom tabs, position children at the visual top of the
-            // layout area (the stored position is the visual top)
+            // for bottom tabs, children sit above the tab bar. The stored
+            // position is the visual top of the tab area, and the tab bar is
+            // drawn below the tallest child (see the bottomTabsChildHeight
+            // padding in layout-object.component), so bottom-align each child
+            // within that band to keep every child flush with the tab bar.
             if(isBottomTabs) {
                Point layoutPos = assemblyLayout.getPosition();
+               double maxChildHeight = childModels.stream()
+                  .mapToDouble(child -> child.getObjectFormat().getHeight())
+                  .max()
+                  .orElse(0);
 
                for(VSObjectModel childModel : childModels) {
                   VSFormatModel fmt = childModel.getObjectFormat();
                   fmt.setPositions(
-                     layoutPos.x, layoutPos.y,
+                     layoutPos.x, layoutPos.y + (maxChildHeight - fmt.getHeight()),
                      fmt.getWidth(), fmt.getHeight());
                }
             }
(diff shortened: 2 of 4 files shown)


jshobe-inetsoft (approved):
### Code Review — Bug #76022: keep bottom-tab children flush with the tab strip in layouts

**Root cause check:** Confirmed against the code. `AbstractLayout.applyTab` previously positioned every bottom-tab child at `npos.y` (visual top) regardless of height, while the tab bar itself is placed at `npos.y + contentHeight`. `VSLayoutService.createObjectModel` did the analogous thing for the composer layout pane. For the tallest child the two coincide, so only shorter children (e.g. a 40px radio button next to a 140px gauge) visibly floated away from the strip — matches the reported symptom exactly.

**Fix correctness:**
- `AbstractLayout.applyTab`: `childPos.y = npos.y + Math.max(0, contentHeight - childSize.height)` — `contentHeight` and `childSize.height` are both `int`, so this compiles cleanly and bottom-aligns each child within the content band. Top-tabs branch (`npos.y + tabSize.height`) is untouched.
- `VSLayoutService.createObjectModel`: `fmt.setPositions(layoutPos.x, layoutPos.y + (maxChildHeight - fmt.getHeight()), ...)` — `getHeight()`/`setPositions` operate on `double`, and `maxChildHeight` is computed via `mapToDouble(...).max()`, so the arithmetic is type-consistent. Using the max *child model* height (rather than `layoutHeight - tabHeight`) is explained in the PR description as matching the same number the template's `padding-top` uses — reasonable and consistent with the Java-side fix's intent.
- Both changes are purely additive to the `bottomTabs` branch; the non-bottom-tabs code path is unchanged, so no regression risk there.

**Tests:** New `AbstractLayoutTest` (top-tabs + bottom-tabs, single/short/tall child) and expanded `VSLayoutServiceTest` cases reproduce the exact reported geometry (24px bar, 140px/40px children) and assert the fixed offsets. The added Spring wiring (`@ExtendWith(SpringExtension.class)`, `@ContextConfiguration`, `@SreeHome`, `@Tag("core")`) looks necessary to get these tests actually picked up/running under the module's test tag filter, consistent with the PR's claim that the tests previously errored before reaching assertions.

**Backward compatibility / regressions:** None identified — the change only affects bottom-tabs child positioning, which was the defective behavior; top-tabs layout, export shrink-to-fit logic (`applyShrunkBottomTabsShift`), and scale-to-screen paths are untouched by this diff.

**Security:** No concerns — pure layout arithmetic, no external input handling.

**Minor/non-blocking:** The PR description notes composer preview and PDF/Excel export of a bottom-tabs layout haven't been manually/visually verified yet — worth a quick sanity check post-merge, but the unit tests directly cover the regressed geometry so this isn't blocking.

**Checks:** `build`, `Semgrep Scan`, `Semgrep OSS`, and `license/cla` all pass. The `claude-review` check shows a failure, but that is the automated review workflow itself erroring out (no review content was ever produced) — this manual review supersedes it. No merge conflicts (`mergeable: MERGEABLE`).

PR #4636, opened 2026-08-17
Bug #76119, stop presentation reset from deleting custom shapes

ngovus-inetsoft:
## Problem

In EM → Settings → Presentation → Look and Feel: open **Custom Shapes**, upload a shape, close the dialog, then click **Reset to Default** — every custom shape disappears, including shapes that were in the environment beforehand.

`PresentationSettingsController.resetSettings()` recursively deleted the whole shapes directory:

```java
String shapesDirectory = globalSettings ?
   ImageShapes.getGlobalShapesDirectory() : ImageShapes.getShapesDirectory();
dataSpaceContentSettingsService.deleteDataSpaceNode(shapesDirectory, false);
```

Both `deleteChildren` and `DataSpace.delete` recurse, so the folder and every file under it were removed.

This is unrecoverable data loss:

- The 16 built-in shapes are classpath resources (`ImageShapes.loadBuiltins`, `/inetsoft/uql/viewsheet/graph/shapes/`). Nothing in the product ever seeds `portal/shapes`, so **every file in that folder is uploaded content with no default to fall back on**.
- `ImageShapes.loadShapes()` reads the global `portal/shapes` folder for *every* org, so a global reset destroyed shapes shared across tenants.
- In a single-tenant deployment `ImageShapes.getShapesDirectory()` also resolves to `portal/shapes`, so even the org-scoped branch hit the global folder.
- Reset was the only shapes-deletion path that bypassed `DataSpaceTreeController.checkDeletePermission`.

Everything else the look-and-feel reset touches is a *named file* backed by a classpath or manager default (logo, favicon, `format.css`, fonts, `userformat.xml`). Shapes were the only blunt recursive folder wipe.

## Fix

Custom shapes are data space content, not a presentation property, so the reset no longer touches them at either scope. The `DataSpaceContentSettingsService` dependency and the `ImageShapes` import are removed along with the delete call, so the controller can no longer reach the data space at all — that removal is the regression guard.

Nothing else in `resetSettings` changed: the `globalSettings` derivation and the `SETTINGS_LOCK` usage are untouched.

### Note on Bug #76034

This reverses the behavior that #76034 asked for. Commit cf16991b67 retargeted this delete from `portal/<currentOrg>/shapes` to `portal/shapes` for a global reset; before that a global reset hit the site admin's normally-empty org folder, which is why reset appeared not to affect custom shapes. The two reports want opposite things from the same line, and the decision here is that reset never deletes uploaded shapes. cf16991b67 is left in history.

## Testing

- `PresentationSettingsControllerTest` — 8 tests, 0 failures. Updated for the positional constructor change; the strategy comment now records that `resetSettings` does not touch the data space.
- `core` compiles cleanly.

Manual steps to confirm: upload two shapes via **Custom Shapes**, click **Reset to Default**, re-open the dialog — both files should still be listed, and still selectable in a chart's shape dropdown. Logo/favicon/tree-sort/CSS/fonts should still reset as before.

## Related pre-existing defects (not addressed here)

Found while tracing; each deserves its own issue:

- `DataSpaceContentSettingsService.deleteDataSpaceNode` only calls `ImageShapes.clearShapes()` when the path starts with `".../shapes/"` (trailing slash), so deleting the shapes *folder* from the EM Data Space page leaves a stale in-memory cache. `clearShapes()` also only clears the current org's entry.
- `custom-shape-dialog.component.ts:186` hardcodes `portal/host-org/shapes` as the only protected root, so a non-default org's root and the global root stay deletable from the dialog.
- `LookAndFeelService.setLogo`/`setFavicon` build the org delete path as `portal/portal/<orgID>/logo.png`, so org logo/favicon files survive a reset.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/presentation/PresentationSettingsController.java
+++ b/core/src/main/java/inetsoft/web/admin/presentation/PresentationSettingsController.java
@@ -21,10 +21,8 @@ import inetsoft.graph.geo.service.MapboxStyle;
 import inetsoft.sree.internal.SUtil;
 import inetsoft.sree.internal.cluster.Cluster;
 import inetsoft.sree.security.*;
-import inetsoft.uql.viewsheet.graph.aesthetic.ImageShapes;
 import inetsoft.util.Catalog;
 import inetsoft.util.InvalidOrgException;
-import inetsoft.web.admin.content.dataspace.DataSpaceContentSettingsService;
 import inetsoft.web.admin.general.WebMapSettingsService;
 import inetsoft.web.admin.presentation.model.PresentationSettingsModel;
 import inetsoft.web.security.RequiredPermission;
@@ -56,7 +54,6 @@ public class PresentationSettingsController {
       PresentationTimeSettingsService timeSettingsService,
       PresentationDataSourceVisibilitySettingsService dataSourceVisibilitySettingsService,
       WebMapSettingsService webMapSettingsService,
-      DataSpaceContentSettingsService dataSpaceContentSettingsService,
       AISettingsService aiSettingsService,
       Cluster cluster)
    {
@@ -76,7 +73,6 @@ public class PresentationSettingsController {
       this.timeSettingsService = timeSettingsService;
       this.dataSourceVisibilitySettingsService = dataSourceVisibilitySettingsService;
       this.webMapSettingsService = webMapSettingsService;
-      this.dataSpaceContentSettingsService = dataSpaceContentSettingsService;
       this.aiSettingsService = aiSettingsService;
       this.cluster = cluster;
    }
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
### Code Review: Bug #76119 — stop presentation reset from deleting custom shapes

**Root cause verified.** `PresentationSettingsController.resetSettings()` previously called `dataSpaceContentSettingsService.deleteDataSpaceNode(shapesDirectory, false)`, and both `deleteDataSpaceNode`/`DataSpace.delete` recurse — so a "Reset to Default" wiped every uploaded shape with no default to restore, since the 16 built-ins are classpath resources under a different path (`ImageShapes.loadBuiltins`) that `portal/shapes` is never seeded from.

**Fix is complete and minimal.**
- The delete call, the `ImageShapes`/`DataSpaceContentSettingsService` imports, the constructor parameter, and the field are all removed together — there's no leftover dead reference, and since Spring wires `PresentationSettingsController` through its single constructor, no other production call site needed updating.
- Grepped for other manual construction of this class: the only non-Spring call site is `PresentationSettingsControllerTest`, which is updated in the same diff to match the new constructor signature.
- Nothing else in `resetSettings` (the `globalSettings` derivation, `SETTINGS_LOCK` usage, other services' `resetSettings` calls) is touched — scope is tight.

**Test coverage:** `PresentationSettingsControllerTest` comments and two existing tests are updated to document/assert that shapes are no longer touched at either scope. No test exercises the literal absence of a shapes-delete call, but that's implicit in the removed mock/field, and adding a positive assertion isn't really possible without the dependency present.

**Backward compatibility / regression risk:** This intentionally reverses bug #76034's fix, which the PR description explains clearly — the previous behavior conflated "reset presentation" with "delete uploaded content," which is the actual defect. The description documents the tradeoff and leaves the old commit in history rather than hiding it, which is good practice for future readers hitting the same report from the other direction.

**Security:** No new input handling; this only removes a delete path, reducing rather than adding attack surface.

**No blocking issues found.**

Checks: `build`, `Semgrep Scan`, and `Semgrep OSS` all pass. The only failing check is `claude-review` itself, which is the automated review workflow erroring out (the reason this manual review exists) — not a substantive code/quality gate, so it isn't treated as a blocking check here. No merge conflicts (`MERGEABLE`).

Approving.

PR #4639, opened 2026-08-17
Bug #76125, report the org default theme as the default in single-tenant mode

ngovus-inetsoft:
### Issue

Presentation → Themes: create two themes on the host org with multi-tenancy enabled, uncheck "Visible to All Organizations" on both, check "Default for This Organization" on Theme2, then disable multi-tenancy and security. Theme2 is still the default (the list shows its checkmark and Portal applies it) but the properties pane shows **Default** unchecked.

### Cause

The org default and global default are two separate `SreeEnv` pointers — `portal.org.selectedThemeId.<orgID>` and `portal.org.selectedThemeId`. In single-tenant mode the properties pane renders only one checkbox, bound to `defaultThemeGlobal`, while `CustomThemeModel.Builder.from()` derived that flag purely from the *global* pointer.

The org pointer survives the multi-tenancy toggle — the host org and the single-tenant default org share the ID `host-org` — and `CustomThemesImpl.getSelectedTheme()` prefers it over the global pointer. So the theme really is the effective default, but the checkbox said otherwise. The theme list's checkmark reads `defaultThemeOrg` first, which is why the two views disagreed.

### Fix

Extracted the flag derivation into `applyDefaultFlags()` (shared by both `from()` overloads) and, when `!SUtil.isMultiTenant()`, report `defaultThemeGlobal` as global-pointer **or** org-pointer match, so the single "Default" checkbox reflects the theme that is actually applied.

The write side (ignoring the stale hidden `defaultThemeOrg` value and clearing the leftover org pointer when "Default" is unchecked) is in the paired enterprise PR, along with the `ThemeServiceTest` coverage.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/presentation/model/CustomThemeModel.java
+++ b/core/src/main/java/inetsoft/web/admin/presentation/model/CustomThemeModel.java
@@ -60,10 +60,7 @@ public interface CustomThemeModel {
 //         groups(theme.getGroups());
 //         roles(theme.getRoles());
 
-         String orgSelected = customThemesManager.getOrgSelectedTheme();
-         String globalSelected = customThemesManager.getGlobalSelectedTheme();
-         defaultThemeGlobal(Objects.equals(globalSelected, theme.getId()));
-         defaultThemeOrg(Objects.equals(theme.getId(), orgSelected));
+         applyDefaultFlags(theme, customThemesManager);
 
          return this;
       }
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):
### Code Review — PR #4639 (Bug #76125)

**Note:** the automated `claude-review` CI check on this PR failed to produce output (job error) — this is a manual review filling that gap.

#### Summary of change
`CustomThemeModel.Builder` extracts the duplicated default-flag logic from both `from()` overloads into a shared `applyDefaultFlags()`, and fixes `defaultThemeGlobal` to also report `true` when the org-selected-theme pointer matches while multi-tenancy is disabled — matching what `CustomThemesImpl.getSelectedTheme()` actually treats as the effective default.

#### Correctness
- Verified the fix directly matches the described root cause: in single-tenant mode only `defaultThemeGlobal` drives the UI's "Default" checkbox, but `getSelectedTheme()` prefers the org pointer, so a theme holding only the (surviving) org pointer was effectively default while showing unchecked. `defaultThemeGlobal(isGlobalDefault || !SUtil.isMultiTenant() && isOrgDefault)` closes that gap.
- Multi-tenant-mode behavior is unchanged (`isGlobalDefault` alone), so no regression there.
- The refactor into `applyDefaultFlags()` is a clean, behavior-preserving DRY extraction — both `from()` call sites now stay in sync by construction, removing the duplication risk the two copies previously had.

#### Minor, non-blocking observations
- Edge case: if org- and global-selected pointers diverge and both point to *different* themes while multi-tenancy is disabled, both themes would now render `defaultThemeGlobal = true` (two checked boxes), since `defaultThemeOrg` still only reflects the raw org pointer independently. This is a narrower edge case than the bug being fixed and arguably less confusing than the previous behavior, but worth being aware of.
- No test changes are included in this community PR; per the PR description, `ThemeServiceTest` coverage for this lives in the paired enterprise PR. Given this file is in `core`, a community-side unit test would be preferable, but the split is explained and not a regression from the current state (no pre-existing tests for this builder either).

#### Security / backward compatibility / regressions
No security concerns (no external input handling). No API/backward-compatibility break — purely a display-flag derivation fix. No existing tests are broken by this change (no tests reference this class).

#### Checks / mergeability
`build`, `Semgrep Scan`, `Semgrep OSS`, and `license/cla` all pass. No merge conflicts (`MERGEABLE`). Only the `claude-review` bot workflow shows a failure, which is the automated-review error this manual review is replacing, not a code issue.

**Verdict: No blocking issues. Approving.**

Issue #4637, opened 2026-08-17
Write coordination: Apply refuses its own second commit, discarding edits on all 21 property dialogs


larryliang-inetsoft:
Merged in #4626 (write coordination). The revision check is correct on its own terms, but nothing refreshes a dialog's captured revision after a commit, and the Composer's property dialogs commit **more than once per session** via their Apply button.

## Effect

On every one of the 21 property dialogs, plus the condition dialog and the VS Options pane:

1. User opens a property dialog — the model arrives carrying `revision = N`.
2. User clicks **Apply**. The commit is accepted, and `VSObjectPropertyService` calls `rvs.bumpWriteRevision()` → server is now at `N+1`. The dialog stays open, still holding `N`.
3. User clicks **Apply** again, or **OK** — the commit arrives with `revision = N`, no longer matches, and is refused with `common.writeConflict`.
4. The user's edits are discarded.

**No concurrent writer is involved.** This is a single user, one dialog, two clicks.

## Why the revision goes stale

The dialog re-emits the same model object it received. From `gauge-property-dialog.component.ts:106` (every property dialog follows this shape):

```ts
() => isApply ? this.onApply.emit(payload) : this.onCommit.emit(this.model),
```

and `ComponentTool.showDialog` wires the apply emitter to the same commit handler, leaving the dialog open:

```ts
// web/projects/portal/src/app/common/util/component-tool.ts:134-149
applySubscription = modal.componentInstance[applyEmitter].subscribe((v: any) => {
   ...
   onCommit(v);
});
```

`revision` is not declared on the TypeScript model interfaces, but it is present on the object at runtime (unknown JSON fields survive), so it round-trips back to the server unchanged — carrying the pre-Apply value.

#4626 changed 63 files, none of them `.ts` or `.html`, and `grep -r revision web/projects/` finds nothing. There is no code path anywhere that could refresh a dialog's revision mid-session, and the server's response to a property commit is a set of assembly-refresh commands rather than a fresh dialog model.

## Verified against merged `feature-76074-server-fixes`

- `VSObjectPropertyService:140` — `if(expectedRevision != null && expectedRevision != rvs.getWriteRevision())` → refuse
- `VSObjectPropertyService:266`, `:361` — `rvs.bumpWriteRevision()` after a successful commit
- `web/projects/**` — zero occurrences of `revision`/`writeRevision`

[...]

PR #4640, opened 2026-08-17
fix(write-coordination): wire the client into the revision check

larryliang-inetsoft:
## Summary

#4626 built the whole write-coordination mechanism server-side (a `writeRevision` counter on `RuntimeViewsheet`, checked against the revision a dialog was read at before committing) but the mechanism never reached the browser: the Angular Composer client never transmitted `revision` in either direction. Every real human dialog commit deserialized with `expectedRevision == null` server-side, so **the check never fired for actual browser usage** — only for tests and the wiz agent's internal path.

This also explains Larry's "Apply self-conflict" finding on #4626: it couldn't manifest, because there was no revision to conflict with in the first place. But that meant the feature the PR was built for — stopping a human dialog from silently clobbering a concurrent agent write — had zero real-world effect.

## Fix

Client-side only; the server already reads whatever revision it's given.

- Each of the 23 property/condition dialog models gains a `revision?: number` field.
- Each dialog's **commit (OK)** payload now includes the revision it was read at.
- **Apply deliberately never sends it.** The dialog's held model is never refreshed after a commit, so echoing the revision back on Apply would make a dialog conflict with its own prior Apply — exactly the self-conflict Larry flagged. This makes that scenario impossible by construction rather than by careful bookkeeping.

Most dialogs forward their held model unmodified to the server and needed only the type declaration (no behavior change). Five needed an actual code change because they reconstruct a slimmed request object that dropped the field: `ChartPropertyDialog`, `CrosstabPropertyDialog`, `HyperlinkDialog`, `HighlightDialog` (via `AbstractHighlight`), and `VSConditionDialog`.

## Test plan

- [x] `tsc -p projects/portal/tsconfig.app.json --noEmit` — clean, zero errors
- [x] Existing specs for the 3 touched dialogs with prior coverage (`hyperlink-dialog`, `highlight-dialog`, `vs-condition-dialog`) — all pass
- [x] Added a regression test to each of those 3 specs asserting revision is present on commit and absent on apply
- [ ] Chart/Crosstab dialogs have no existing spec files to extend; verified by full compile + manual trace of the reconstructed-object logic, not by a new test harness

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/RuntimeViewsheet.java
+++ b/core/src/main/java/inetsoft/report/composition/RuntimeViewsheet.java
@@ -2923,7 +2923,11 @@ public class RuntimeViewsheet extends RuntimeSheet {
    private boolean preview; // preview flag
    private boolean needRefresh = false; // force refresh
    private int mode; // viewsheet mode
-   private transient int writeRevision; // bumped on every dialog-owning commit; see getWriteRevision
+   // Bumped by VSObjectPropertyService, VSConditionDialogService, ViewsheetPropertyDialogService,
+   // VSBindingService, and the wiz agent write seams (ViewsheetSessionService.mutate,
+   // ScriptEditService) -- not literally every dialog-owning commit; a dialog service that
+   // routes through a different path does not bump it. See getWriteRevision.
+   private transient int writeRevision;
    private transient Integer baseWriteRevisionAtOpen; // binding-editor clones only; see accessor
    private ViewsheetSandbox box; // query sandbox
    private AssetRepository rep; // asset repository
--- a/web/projects/portal/src/app/common/abstract-highlight.component.ts
+++ b/web/projects/portal/src/app/common/abstract-highlight.component.ts
@@ -102,10 +102,17 @@ export class AbstractHighlight {
       }
    }
 
-   // remove fields that are not used on the server side to reduce the transmission size
-   getServerAppliedModel(): HighlightDialogModel {
+   // remove fields that are not used on the server side to reduce the transmission size.
+   // Only a final commit (OK) carries the revision it was read at -- Apply leaves the dialog
+   // open with its held model never refreshed, so echoing it back on Apply would make the
+   // dialog conflict with its own prior Apply. See 2026-08-17-write-coordination-design.md.
+   getServerAppliedModel(includeRevision: boolean = true): HighlightDialogModel {
       let model = Tool.shallowClone(this._model);
       model.fields = [];
+
+      if(!includeRevision) {
+         delete model.revision;
+      }
 
       if(model.highlights) {
          model.highlights = Tool.shallowClone(model.highlights);
--- a/web/projects/portal/src/app/common/data/condition/vs-condition-dialog-model.ts
+++ b/web/projects/portal/src/app/common/data/condition/vs-condition-dialog-model.ts
@@ -21,4 +21,9 @@ export interface VSConditionDialogModel {
    tableName: string;
    fields: DataRef[];
    conditionList: any[];
+   // The write-coordination revision this model was read at. Sent back only on a final commit
+   // (OK), never on Apply -- the dialog's own held copy is never refreshed after a commit, so
+   // echoing it back on Apply would make the dialog conflict with its own prior Apply. See
+   // 2026-08-17-write-coordination-design.md.
+   revision?: number;
 }
--- a/web/projects/portal/src/app/common/util/component-tool.spec.ts
+++ b/web/projects/portal/src/app/common/util/component-tool.spec.ts
@@ -23,14 +23,21 @@ function fakeModal() {
    const onCommit = new Subject<any>();
    const onCancel = new Subject<any>();
    const componentInstance: any = { onApply, onCommit, onCancel };
+   let resolveResult: (v: any) => void;
    const modalRef: any = {
       componentInstance,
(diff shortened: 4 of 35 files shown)


larryliang-inetsoft:
**The Apply self-conflict this PR guards against is still open for the ~15 composer property dialogs it adds the field to.**

The guard is implemented precisely where the payload is *rebuilt* — `chart-property-dialog` constructs a separate `commitModel`, and `vs-condition-dialog` gates it behind `getServerAppliedModel(includeRevision)`. Both are correct, and the comment added to every model file states the rule plainly:

> Sent back only on a final commit (OK), never on Apply — the dialog's own held copy is never refreshed after a commit, so echoing it back on Apply would make the dialog conflict with its own prior Apply.

But the composer dialogs never rebuild their payload. They emit the server-supplied object verbatim on **both** paths, so `revision` rides along on Apply:

```ts
// composer/dialog/vs/gauge-property-dialog.component.ts
const payload = {collapse: collapse, result: this.model};   // <-- same object, revision included
() => isApply ? this.onApply.emit(payload) : this.onCommit.emit(this.model),

// composer/dialog/vs/oval-property-dialog.component.ts — same shape
// selection-list / selection-tree use Tool.clone(this.model), which also preserves revision
```

That matters because `editObjectProperty` both **checks and bumps** on the same call:

```java
// VSObjectPropertyService.java:140
if(expectedRevision != null && expectedRevision != rvs.getWriteRevision()) { ...refuse... }
// VSObjectPropertyService.java:361 — main success path
hint = assembly.setVSAssemblyInfo(info);
rvs.bumpWriteRevision();
```

So for any composer property dialog:

1. Server sends the model with `revision: N`.
2. **Apply #1** → sends `N`, check passes, edit applies, revision bumps to `N+1`.
3. The dialog's held `this.model.revision` is still `N` — Apply deliberately leaves the dialog open and nothing refreshes it.
4. **Apply #2** → sends `N` against `N+1` → refused with `common.writeConflict`.

Clicking Apply twice is ordinary use, so if that reading is right this is user-visible on gauge, oval, calendar, image, line, rectangle, selection-*, submit, tab, text, viewsheet-object and viewsheet — every model in this diff whose component emits `this.model`.

Two things worth separating:

- **This isn't introduced here.** JS objects retain unknown JSON properties, so `this.model.revision` was already present at runtime before the TS interface declared it, and Apply already echoed it on the base branch. The PR makes the field *visible*, and adds the comment stating an invariant the composer half doesn't yet satisfy.
- **The declaration and the enforcement have drifted apart.** 21 models now carry a rule that 6 components implement. A future reader has no way to tell which dialogs honour it, because the comment is identical in all of them.

One assumption I could not settle statically, and it's the load-bearing one: that nothing repopulates `this.model.revision` after an Apply round-trip. The comment asserts it and I found no such path, but I haven't run the dialogs — worth confirming before treating the above as reproduced.

If it holds, the cheap fix is the same shape as the two you already wired: strip `revision` from the Apply payload rather than adding it to the commit payload, i.e. `{collapse, result: {...this.model, revision: undefined}}` in the shared composer dialog path, so it's one edit rather than fifteen.

Minor, unrelated: `RuntimeViewsheet.java:2926` describes `writeRevision` as *"bumped on every dialog-owning commit"*, but the bump sites are `VSObjectPropertyService` (×2), `VSConditionDialogService`, `ViewsheetPropertyDialogService`, `VSBindingService` and the two wiz seams. Dialog services that route elsewhere don't bump, so the field comment reads broader than the code.

larryliang-inetsoft:
Confirmed — you're right, and it's worse than the diff made it look: the fix only reached the ~5 dialogs that reconstruct a slimmed request object. Every other property dialog (gauge, oval, calendar, image, line, rectangle, selection-*, submit, tab, text, viewsheet-object, viewsheet) forwards `this.model` (or a `Tool.clone` of it) unmodified on both apply and commit, so revision rode along on Apply exactly as you traced. Confirmed your load-bearing assumption too: nothing repopulates the dialog's held model after an apply round-trip anywhere in this codebase — Apply is fire-and-forget over the WS event, no response handler touches `dialog.model`.

Fixed at the choke point you pointed to rather than patching each dialog: every apply path already funnels through `ComponentTool.showDialog`'s `applySubscription` before reaching the caller's commit callback, so `revision` is stripped there once (`cd46e6df6`). That also covers any dialog added later without anyone having to remember the rule. Added `component-tool.spec.ts` covering both the strip and the no-op case. Also tightened the `writeRevision` field comment per your list of actual bump sites.

Left the existing per-dialog exclusions (Hyperlink, Highlight, Condition) in place rather than reverting them now that the central fix subsumes them — redundant but harmless, and touching already-tested code for no behavioral gain felt like the wrong tradeoff mid-review. Happy to clean that up in a follow-up if you'd rather they read as one mechanism.

larryliang-inetsoft:
Self-caught while double-checking this closes stylebi#4637 fully: stripping revision from Apply alone stops a dialog from conflicting with its own prior Apply, but the FINAL commit (OK) that follows an Apply was still sending the dialog's original, now-stale revision -- the Apply itself had already bumped the server's counter. That commit would have been wrongly refused, which is exactly #4637's "Apply then OK discards your edits" shape, just not fully closed by the first fix.

Fixed in `97466247c`: track whether this dialog session has Applied at least once, and if so, strip revision from the final commit too. Added two more tests to `component-tool.spec.ts` covering the Apply-then-OK sequence and the ordinary OK-with-no-prior-apply case (still checked normally).

PR #4642, opened 2026-08-17
D4 remainder: unscoped list_bindable_fields fabricates a field instead of listing tables

larryliang-inetsoft:
## Summary
Closes the second known-live D4 (silent-acceptance audit) instance: unscoped `list_bindable_fields` returns a single fabricated field instead of the viewsheet's actual tables.

**Root cause**, traced entirely by static reading (no live access needed to diagnose): `VSTreeHandler.isLeaf()` marks every `AssetEntry.Type.WORKSHEET` entry a leaf unconditionally — correct for a standalone worksheet reference, but `VSEventUtil.refreshBaseWSTree`'s non-direct-source branch reuses a WORKSHEET-typed entry as the **container** node holding the viewsheet's real tables, which does have children. `createNodeFromEntry` sets `.leaf(...)` and `.children(...)` independently, so that container node ends up `leaf: true` while still carrying real table children.

`BindableFieldsService`'s tree-walker trusted `.leaf()` blindly to decide "is this a column" — so the container itself got treated as one: `fieldOf(container)` turned its own label into a fabricated `{column, dataType: null, role: null}`, and the real tables beneath it were never visited. This is exactly the live shape B3 recorded: unscoped `list_bindable_fields` returning `{name: null, fields: [{column: "a1", dataType: null, role: null}]}` instead of the viewsheet's tables — a well-formed call, accepted, plausible-but-wrong result.

**Fix**: trust actual children (`node.children().isEmpty()`) rather than the tree's own `leaf` hint, which is a UI signal that can (and here does) disagree with structure. A node with children is never a column, regardless of what `leaf()` claims.

## Wider D4 triage (this pass)
- `convertToFreehandTable`'s silent no-op on the wrong assembly type and `VSObjectPropertyService.checkScript`'s dual-null ambiguity are real but **not reachable from any `wiz/` tool today** — neither is wrapped by a plugin tool yet (E3 owns building that, and its own spec already commits to guarding this).
- `VSConditionDialogService.browseData` and `RegionPropertyDialogService.getChartArea` both have a bare `catch(Exception) { return null; }` and are reachable live, but touch shared human-facing Composer code — left unfixed pending live verification, since swallow-and-log may be deliberate ("don't crash a live editing session over a transient hiccup") rather than a wiz-specific defect.
- Everything else sampled (`removeObject`, `moveObject`, `resizeObject`, etc.) is already pre-guarded by wiz's own `requireExisting` checks before the call.

## Test plan
- [x] New regression test reproduces the exact container shape (WORKSHEET-typed, `leaf: true`, real table children) end to end
- [x] `BindableFieldsServiceTest` — 11/11
- [x] Full `inetsoft.web.wiz.**` sweep — 157 test classes, 0 failures

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindableFieldsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindableFieldsService.java
@@ -113,7 +113,7 @@ public class BindableFieldsService {
       List<BindableField> direct = new ArrayList<>();
 
       for(TreeNodeModel child : node.children()) {
-         if(child.leaf()) {
+         if(hasNoDescendants(child)) {
             direct.add(fieldOf(child));
          }
          else {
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
## Review — the fix's own failure mode reappears for an empty table

`hasNoDescendants` is a genuine improvement over `leaf()` for the WORKSHEET-container case (confirmed: `VSTreeHandler.isLeaf` hardcodes `true` for `VIEWSHEET`/`WORKSHEET`/`COLUMN` only, with no structural check at all — so `leaf()` was never trustworthy for any container type, not just this one instance of it disagreeing with reality).

But the replacement conflates two different questions: "does this node have children" and "is this node a column." `isTable(node)` already exists in this class to answer the second question for a table specifically — and it's checked at the top of `collect()`, but **not** inside the loop where `hasNoDescendants` decides whether to recurse or fabricate a field. That means a table with zero exposed columns (permission-filtered, a fresh/empty inline table, mid-load metadata — plausible, not exotic) now gets treated as a column itself:

```java
TreeNodeModel emptyTable = TreeNodeModel.builder()
   .label("EmptyTable").data(entry(AssetEntry.Type.TABLE, "EmptyTable", null)).build(); // no children
TreeNodeModel root = TreeNodeModel.builder().label("root").addChildren(emptyTable).build();

list("rt1", null, principal());
// -> [BindableTable("root", [BindableField("EmptyTable", dataType: null, role: null)])]
```

That's the identical shape to the bug this PR fixes: a plausible, well-formed, **fabricated** field standing in for a table with nothing to report — should be `[]`, the same way an empty tree already correctly returns `[]` via the `if(!fields.isEmpty())` guard a few lines up. The old `leaf()`-based code didn't have this failure mode, because `leaf()` is hardcoded `false` for `TABLE`/`PHYSICAL_TABLE`/`QUERY`/`LOGIC_MODEL` regardless of children, so an empty table always recursed into the `isTable()` branch and correctly produced nothing.

Suggested fix — fold the type check into the predicate instead of relying on absence-of-children alone:

```java
private boolean isColumn(TreeNodeModel node) {
   return node.children().isEmpty() && !isTable(node);
}
```

(`isTable()` is a non-static instance method already, so this can't stay `static` as-is if it calls it — minor.) Worth a regression test alongside the WORKSHEET-container one: a `TABLE`-typed node with zero children should yield `[]`, not a fabricated field.

PR #4643, opened 2026-08-17
Bug #76036, fix unresponsive mobile mini-toolbar actions on a selected chart

ngovus-inetsoft:
Redmine #76036 — *\<Mobile\> open Example/Return Analysis in mobile, select chart plot, then click some action in minitoolbar, no response.*

## Symptoms

1. On a mobile layout, select a chart plot in `Examples/Return Analysis`, then tap an action in the mini-toolbar (e.g. Brush) — nothing happens.
2. `NG0100: ExpressionChangedAfterItHasBeenCheckedError` is logged (previous `null`, current `true`, `viewer-app.component.html:40`).

These turned out to be two independent defects.

## 1. Dead mini-toolbar buttons — stale `AbstractVSActions` instance

On mobile, `MiniToolbar` renders nothing (its button block is inside `@if (!mobileDevice)`); the toolbar shown is `viewer-mobile-toolbar`, bound to `viewer-app`'s `selectedActions`. Its buttons only call `onAssemblyActionEvent.emit(...)` on that one instance, and the sole subscriber is `VSChart`, whose `actions` input setter **unsubscribes from the previous instance** whenever a new one is assigned.

`processAddVSObjectCommand` recreates `vsObjectActions` for **every** assembly whenever **any** single assembly updates, but only re-pointed `selectedActions` when the incoming command happened to name the selected assembly:

```ts
this.vsObjectActions = this.vsObjects.map(model => ...);   // rebuilds ALL

if(this.selectedActions &&
   this.selectedActions.getModel().absoluteName == command.name)   // only the updated one
```

So: tap the chart plot → flyover fires → the server refreshes the output assemblies → each resulting `AddVSObjectCommand` hands `VSChart` a fresh `ChartActions` (its setter drops the old subscription) while `selectedActions` keeps pointing at the old one → tapping Brush emits into an emitter with zero subscribers. No error, no response.

**Fix:** re-point `selectedActions` by assembly name after the rebuild, regardless of which assembly triggered the command (new `refreshSelectedActions()`). This also removes a latent NPE — the old condition dereferenced `getModel()` unguarded.

The desktop mini-toolbar was never affected; it binds `[actions]="vsObjectActions[i]"` straight from the live array.

## 2. NG0100 — parent binding written by a descendant

`viewer-app` reads `pagingControlModel?.enabled` for its `paging-control` bindings. That model is written by a descendant during the same change-detection pass: `ChartArea`'s `@Input() set selected` called `showPagingControl()` synchronously, after the parent had already evaluated the expression as `null`.

**Fix:** publish the paging control model after the current change-detection pass, re-checking `_selected` in case the chart was deselected in between; the timeout handle is cleared in `ngOnDestroy`. The table components were already safe — they call `showPagingControl()` from an `(onClick)` handler rather than an input setter.

## Tests

- `viewer-app.component.risk.tl.spec.ts` — new regression tests: `selectedActions` is re-pointed after an update to a *different* assembly, and left alone when the selected assembly no longer exists.
- `chart-area.component.interaction.tl.spec.ts` — the two `selected`-setter tests now use fake timers (they asserted synchronous behavior); added one covering select-then-deselect before the deferred call fires.
- `viewer-app.test-fixtures.ts` — filled in the `DataTipService` / `PopComponentService` mocks so `processAddVSObjectCommand` is reachable from tests.

373 TL tests pass across the six chart-area + viewer-app suites; `tsc --noEmit` clean; no new lint output on the changed files. Manually verified in mobile emulation.

### DIFF
--- a/web/projects/portal/src/app/graph/objects/chart-area.component.interaction.tl.spec.ts
+++ b/web/projects/portal/src/app/graph/objects/chart-area.component.interaction.tl.spec.ts
@@ -148,18 +148,53 @@ describe("ChartArea — model setter", () => {
 // ---------------------------------------------------------------------------
 
 describe("ChartArea — selected setter", () => {
+   // Bug #76036: the paging control model is published after the current change detection
+   // pass (the setter runs while the parent view is being checked), so the call is deferred.
    it("should call showPagingControl when selected becomes true on a mobile device", () => {
-      const { comp, pagingControlService } = createComponent();
-      (comp as any).mobileDevice = true;
-      comp.selected = true;
-      expect(pagingControlService.setPagingControlModel).toHaveBeenCalled();
+      vi.useFakeTimers();
+
+      try {
+         const { comp, pagingControlService } = createComponent();
+         (comp as any).mobileDevice = true;
+         comp.selected = true;
+         expect(pagingControlService.setPagingControlModel).not.toHaveBeenCalled();
+         vi.runAllTimers();
+         expect(pagingControlService.setPagingControlModel).toHaveBeenCalled();
+      }
+      finally {
+         vi.useRealTimers();
+      }
    });
 
    it("should NOT call showPagingControl when selected becomes false", () => {
-      const { comp, pagingControlService } = createComponent();
-      (comp as any).mobileDevice = true;
-      comp.selected = false;
-      expect(pagingControlService.setPagingControlModel).not.toHaveBeenCalled();
+      vi.useFakeTimers();
+
+      try {
+         const { comp, pagingControlService } = createComponent();
+         (comp as any).mobileDevice = true;
+         comp.selected = false;
+         vi.runAllTimers();
+         expect(pagingControlService.setPagingControlModel).not.toHaveBeenCalled();
+      }
+      finally {
+         vi.useRealTimers();
+      }
+   });
+
+   it("should not publish the paging control model if deselected before the deferred call runs", () => {
+      vi.useFakeTimers();
+
+      try {
+         const { comp, pagingControlService } = createComponent();
+         (comp as any).mobileDevice = true;
+         comp.selected = true;
+         comp.selected = false;
+         vi.runAllTimers();
+         expect(pagingControlService.setPagingControlModel).not.toHaveBeenCalled();
+      }
+      finally {
+         vi.useRealTimers();
+      }
    });
 
    it("should reflect the assigned value via the getter", () => {
--- a/web/projects/portal/src/app/graph/objects/chart-area.component.ts
+++ b/web/projects/portal/src/app/graph/objects/chart-area.component.ts
@@ -142,6 +142,7 @@ export class ChartArea implements OnInit, OnChanges, OnDestroy {
    private clearCanvasSubscription: Subscription;
    private scrollTopSubscription: Subscription;
    private scrollLeftSubscription: Subscription;
+   private pagingControlTimeout: number = null;
    private flyoverApplied = false;
 
    @Input() set model(model: ChartModel) {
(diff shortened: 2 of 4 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Summary

This PR fixes two independent bugs in the mobile viewer:

1. **Dead mobile mini-toolbar buttons** — `ViewerAppComponent.processAddVSObjectCommand` rebuilds `vsObjectActions` for *every* assembly on any single-assembly update, but previously only re-pointed `selectedActions` when the triggering command happened to name the currently-selected assembly. The new `refreshSelectedActions()` helper re-points `selectedActions` at the freshly rebuilt actions instance for whichever assembly is currently selected, regardless of which assembly triggered the command, and re-subscribes via `addMobileActionSubsciption()`. It also fixes a latent NPE by using `?.` to guard `getModel()?.absoluteName`.
2. **NG0100 (`ExpressionChangedAfterItHasBeenCheckedError`)** — `ChartArea`'s `selected` setter called `showPagingControl()` synchronously from within a parent's change-detection pass. The fix defers the call via `setTimeout`, re-checking `this._selected` at fire time (so a deselect-before-fire is correctly a no-op), and clears the pending timeout both on re-entry and in `ngOnDestroy`.

### Correctness

- `refreshSelectedActions()` is only invoked once per `processAddVSObjectCommand` call, inside the branch that already matches `command.name` — not in a nested per-object loop — so there's no redundant-call concern.
- The `index >= 0 && this.vsObjectActions[index] && this.vsObjectActions[index] != this.selectedActions` guard correctly avoids reassigning/resubscribing when nothing changed.
- The deferred `showPagingControl()` re-checks `this._selected` inside the timeout callback, correctly handling rapid select→deselect before the timer fires (covered by the new "should not publish... if deselected before the deferred call runs" test).
- `pagingControlTimeout` is cleared both when a new "selected" assignment supersedes a pending one and in `ngOnDestroy`, avoiding a stale callback firing after teardown.

### Test coverage

New tests cover both fixes directly and meaningfully:
- `chart-area.component.interaction.tl.spec.ts`: deferred publish, no publish on deselect, no publish when deselected before the deferred call fires (uses fake timers correctly, with `try/finally` restoring real timers).
- `viewer-app.component.risk.tl.spec.ts`: re-point `selectedActions` when a *different* assembly's update triggers the rebuild, and leave it alone when the previously-selected assembly no longer exists.

- The desktop mini-toolbar binds directly to the live `vsObjectActions` array and is explicitly called out as unaffected — consistent with the diff (no desktop-path changes).
- Table components already called `showPagingControl()` from a click handler rather than an input setter, so they're unaffected by the deferral.
- No public API / model changes; `refreshSelectedActions()` is private and additive.

### Security

No security concerns — purely client-side UI/timing logic, no new external input handling.

### Minor/non-blocking

- `chart-area.component.ts`: the new `pagingControlTimeout` field is typed `number`, matching `window.setTimeout`'s return type — fine, but worth double-checking there's no SSR path for this component where `window` is unavailable (doesn't look like there is, based on the surrounding code).

### Blocking issue: failing `build` check

The `build` check on the latest commit fails in the `inetsoft-core` Maven module:

```
[ERROR] Tests run: 4, Failures: 0, Errors: 1, Skipped: 1, Time elapsed: 0.316 s <<< FAILURE! -- in inetsoft.sree.security.DashboardRegistryOrgLifecycleTest
...
[INFO] inetsoft-core ...................................... FAILURE [05:25 min]
```

This PR only touches `web/projects/portal` TypeScript files — no Java/`core` changes — so this looks unrelated to the change itself and consistent with a known intermittent `inetsoft-core` test-ordering flake in this repo's CI rather than a regression introduced here. Still, per policy this blocks approval until the build is green (re-run the job, or investigate if it reproduces consistently).

---

**Verdict:** No blocking issues found in the code itself — the fix is well-targeted, correctly handles the async/re-entrancy edge cases, and has solid regression test coverage. However, requesting changes because the `build` check is currently failing (see above); no conflicts otherwise.

jshobe-inetsoft (approved):

PR #4644, opened 2026-08-17
Bug #76023, fix spinner round corner in exports

ngovus-inetsoft:
## Problem

Setting a round corner on a spinner and exporting the viewsheet (reported against **Excel / match layout**) still showed right angles.

The exported spinner image is rendered server-side by `VSSpinner`. Round corner was honored for the assembly background (`ExportUtil.drawBackground` → `fillRoundRect`) and the assembly border (`ExportUtil.drawBorders`), but not by the widget itself:

| Layer | Drawn by | Round corner |
|---|---|---|
| background fill | `VSFloatable.drawBackground` | yes |
| assembly border | `VSObject.drawBorders` | yes |
| widget box + stepper arrows | `VSSpinner.paintComponent` — `drawRect` + rectangular `NumericStepper` skins | **no** |

So the light-gray box and the arrow images were painted as right angles, on top of and outside the rounded border.

This is not Excel-specific — `PoiExcelVSExporter`, `PPTVSExporter`, `PDFVSExporter` and `SVGVSExporter`/`PNGVSExporter` all embed this same bitmap. HTML export is unaffected, since `HTMLCoordinateHelper.writeSpinner` emits a real `<input>` with `border-radius`.

## Fix

`community/core/.../report/gui/viewsheet/VSSpinner.java`:

- **`clipRoundCorner()`** clips to the *border* rectangle, derived from the same geometry `VSObject.drawBorders` uses — shifted back by the border gap that `VSFloatable.paint()` translates by, and inset by the right/bottom border width. The stepper skins are therefore trimmed along the corner arc instead of overhanging it by the border width. The clip is applied *after* the value text is drawn, so the number is not sliced.
- **`drawWidgetBox()`** draws the widget box with `drawRoundRect` when a round corner is set, and skips it entirely when the assembly format defines its own borders. The assembly border already delimits the widget; the second outline appeared as a doubled line — misaligned at the corners, since the two rectangles have different origins and sizes — which does not match the browser, where only the input's own border is drawn.

Both the themed and the no-theme fallback branches are covered.

## Verification

Rendered the spinner from the reported asset (140×77, `roundCorner="20"`, 1px borders on all sides) at export scale (`getImage(true)`, 2×) across all four combinations:

| Round corner | Assembly border | Result |
|---|---|---|
| 20 | yes (reported case) | one clean rounded border, arrows clipped to the arc, no doubled line, value text intact |
| 0 | yes | one border, no redundant inner gray rect |
| 20 | no | widget box drawn as a round rect, arrows clipped |
| 0 | no | unchanged from before the fix |

The only behavior change for existing exports is the removal of the redundant inner outline when a border is set.

### DIFF
--- a/core/src/main/java/inetsoft/report/gui/viewsheet/VSSpinner.java
+++ b/core/src/main/java/inetsoft/report/gui/viewsheet/VSSpinner.java
@@ -25,6 +25,7 @@ import inetsoft.util.CoreTool;
 import inetsoft.util.ThreadContext;
 
 import java.awt.*;
+import java.awt.geom.RoundRectangle2D;
 import java.text.Format;
 
 /**
@@ -71,6 +72,7 @@ public class VSSpinner extends VSFloatable {
       int upH = (h + 1) / 2;
       int downH = h + 1 - upH; // make sure no rounding error
       FlexTheme theme = getTheme();
+      int roundCorner = format.getRoundCorner();
 
       Image upImg = theme != null
          ? theme.getImage("s|NumericStepper", "upArrowUpSkin", -1, upH) : null;
@@ -80,10 +82,11 @@ public class VSSpinner extends VSFloatable {
       if(upImg != null && dnImg != null) {
          int imgW = upImg.getWidth(null);
          drawString(g2, 0, 0, w - imgW, h, label, format);
-         g2.setColor(Color.lightGray);
+         // clip after the text is drawn so only the box and the stepper images
+         // are trimmed to the rounded shape
+         clipRoundCorner(g2, roundCorner);
          // right edge should be covered by the images
-         //Bug #31348. if start with (0,0), the top and left border will be hidden when export png.
-         g2.drawRect(1, 1, w - imgW / 2, h - 2);
+         drawWidgetBox(g2, w - imgW / 2, h, roundCorner);
          g2.drawImage(upImg, w - imgW, 0, null);
          g2.drawImage(dnImg, w - imgW, upH, null);
       }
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (changes requested):
## Code Review: Bug #76023 — fix spinner round corner in exports

### Summary

The root-cause analysis is sound: `VSSpinner.paintComponent` drew the widget box and stepper images as right angles regardless of the assembly's `roundCorner` format, while the background fill and assembly border already honored it. The fix adds:

- `clipRoundCorner()` — clips the graphics to a `RoundRectangle2D` matching the border rectangle geometry used by `VSObject.drawBorders`, applied after the value text is drawn so the label isn't clipped.
- `drawWidgetBox()` — draws the widget's own outline with `drawRoundRect` when a round corner is set, and skips it entirely when the assembly format defines any border (avoiding the previously-doubled, misaligned outline).

Both the themed and no-theme fallback branches are updated consistently, and the existing "Bug #31348" `(1,1)` origin workaround is preserved.

### Findings

1. **Edge case: partial-side borders** (`VSSpinner.java`, `hasBorders()`) — `hasBorders()` returns true if *any* side (`TOP`/`BOTTOM`/`LEFT`/`RIGHT`) has a border width > 0, which then skips drawing the widget's own box entirely. Previously the inner box was drawn unconditionally on all four sides. If a user sets only one side of the assembly border (e.g., top only), the widget will now show no visible outline on the other three sides, where before it always had one. This is a narrower case than the reported bug and may be intentional/acceptable, but it's worth confirming this asymmetric-border configuration isn't a supported/common case, since it's a behavior change beyond the reported symptom.
2. **No automated test coverage added** — the change is purely rendering logic (`Graphics2D` clip/draw calls) and no test file was touched. There's no pre-existing test scaffolding for `VSSpinner` to extend, so this is a non-blocking observation rather than a gap introduced by this PR.
3. No security concerns — purely presentational AWT drawing code, no external input handling.
4. No back-compat API changes — both new methods are `private`, and the only behavioral change (removing the doubled inner outline when a border is set, and clipping to the rounded corner) is the intended fix described in the PR.

### Checks / mergeability

- The **`build`** check is currently **FAILING**. The log shows the failure is a Maven dependency-resolution error unrelated to this diff: `Failed to transfer artifact com.fasterxml.jackson.module:jackson-module-parameter-names:pom:2.21.2 ... 502 Bad Gateway` from Maven Central — i.e. an upstream registry blip during the `inetsoft-server` module resolution, not a compile/test failure caused by this change. Recommend re-running the build job.

### Verdict

No blocking issues found in the code itself (item 1 above is worth a quick confirmation but is not clearly incorrect). However, per review policy this PR cannot be approved while a check is failing, even though the failure looks like a transient upstream Maven Central outage rather than a real problem with the change. **Requesting changes**: please re-run the `build` check (or push a no-op commit to retrigger it) so it can pass before merge.

jshobe-inetsoft (approved):

PR #4646, opened 2026-08-17
Bug #76042, fix unresponsive mobile toolbar actions after selecting a selection value

ngovus-inetsoft:
## Problem

Repro (from the report):
1. Open `Example/Sales Summary` on mobile — the mobile layout is applied.
2. Select a selection value, then tap an action in the toolbar (e.g. Sort).
3. Nothing happens.

## Root cause

On mobile the assembly mini-toolbar is not rendered at all — `mini-toolbar.component.html` guards the whole button row with `@if (!mobileDevice)`. Mobile gets `viewer-mobile-toolbar` in the top bar instead, driven by `ViewerAppComponent.selectedActions`.

Those buttons work by emitting on the actions object's `onAssemblyActionEvent`. The only subscriber is the assembly component itself — `VSSelection` subscribes inside its `actions` input setter and unsubscribes from the previous instance every time the input changes.

`processAddVSObjectCommand` rebuilds the **entire** `vsObjectActions` array with fresh instances whenever any assembly is added or updated, but it only re-pointed `selectedActions` when `command.name` matched the selected assembly. So:

1. Tap the selection list → `selectedActions` = instance A. Toolbar works.
2. Tap a selection value → the server applies the filter and pushes `AddVSObjectCommand`s for the **dependent** assemblies (the chart/table being filtered).
3. Each of those rebuilds all actions. `VSSelection` receives instance B via `[actions]` and re-subscribes to B; `selectedActions` still holds A.
4. Tapping Sort emits on A, which now has zero subscribers → silent no-op.

`processRefreshVSObjectCommand` had the same name-equality guard, but since it only replaces `vsObjectActions[i]` it happened to be correct.

## Fix

Replaced both guards with a `resyncSelectedActions()` helper that re-points `selectedActions` at the current actions instance for the selected assembly **by name**, regardless of which assembly the command targeted, and re-arms the mobile action subscription only when the instance actually changed.

## Testing

- Added three regression tests to `viewer-app.component.risk.tl.spec.ts`. The key one — *"should re-point selectedActions when a DIFFERENT assembly is refreshed"* — was confirmed to **fail on the unpatched code** and pass with the fix, so it is not vacuous.
- All 181 tests across `viewer-app.component.*.tl.spec.ts` pass.
- `ng lint portal`: 0 errors.

Note: running all 78 `vsobjects/**/*.tl.spec.ts` suites together shows 9–12 suites crashing at import time (`Class extends value undefined` in `sort-column-event.ts`, a circular-import issue). This is pre-existing and flaky — unmodified `main` produces 12 such failures on the same command, with 0 actual test failures in both cases.

Not verified on a physical device; the analysis is from the code path, and the added test reproduces the exact command sequence.

### DIFF
--- a/web/projects/portal/src/app/vsobjects/viewer-app.component.risk.tl.spec.ts
+++ b/web/projects/portal/src/app/vsobjects/viewer-app.component.risk.tl.spec.ts
@@ -59,6 +59,7 @@ import {
    FULL_SCREEN_SERVICE_MOCK,
    ASSET_LOADING_SERVICE_MOCK,
    HEARTBEAT_WORKER_SERVICE_MOCK,
+   ASSEMBLY_ACTION_FACTORY_MOCK,
    fullScreenChangeSubject,
    resetMocks,
    renderComponent,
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
## Code Review — PR #4646

### Summary
Fixes unresponsive mobile toolbar actions after a selection value refresh causes a *different* assembly's actions to be rebuilt while the mobile toolbar remains bound to a now-orphaned `selectedActions` instance.

### Root cause verified
`vsObjectActions` is rebuilt via `this.vsObjects.map(...)` (full-array replace) or per-index replace in several command handlers (`processAddVSObjectCommand`, `processRefreshVSObjectCommand`, `processUpdateZIndexesCommand`). The old code only re-pointed `selectedActions` when the *targeted* assembly's name matched the *selected* assembly's name — so when a rebuild was triggered by refreshing a different assembly (e.g. a chart refreshed by applying a selection), the selected assembly's actions instance was still replaced, but `selectedActions` kept referencing the stale, orphaned instance. Since the assembly component re-subscribes its event handler to the *new* `[actions]` instance, `selectedActions.onAssemblyActionEvent` had no listener, so mobile toolbar taps (e.g. Sort) silently did nothing.

### Fix assessment
- `resyncSelectedActions()` looks up the selected assembly by `absoluteName` in the *current* `vsObjects`/`vsObjectActions` arrays (which are index-aligned by construction — verified `vsObjectActions = vsObjects.map(...)`), rather than relying on stale indices. This is more robust than an index-based fix, since some paths also `unshift`/`splice` these arrays.
- No-ops correctly when `selectedActions` is `null`, when the assembly isn't found (`index === -1`), or when the reference is already current — preserving prior behavior for the non-buggy paths.
- Reuses the existing `addMobileActionSubsciption()` helper, which already unsubscribes the previous subscription before resubscribing — no double-subscription risk.
- Called consistently at all three sites that replace `vsObjectActions` entries (`processAddVSObjectCommand`, `processRefreshVSObjectCommand`, `processUpdateZIndexesCommand`), including once after a loop in the z-index case rather than per iteration — reasonable.
- Test coverage (`viewer-app.component.risk.tl.spec.ts`) directly covers the three relevant cases: a different assembly refreshed, the selected assembly refreshed, and nothing selected. Assertions check reference identity, which is the crux of the bug.

No API surface changes (new method is `private`); behavior is unchanged for the case that previously worked (selected assembly itself is the one refreshed) and is fixed for the previously-broken case. No security concerns — purely client-side UI state management.

### Checks / mergeability
`build`, `Semgrep OSS`, `Semgrep Scan`, and `license/cla` all pass. `mergeable: MERGEABLE`, no conflicts. The only failing check is `claude-review` — the automated review job that failed to complete on this PR (not a required status check per the repo's branch ruleset); this manual review supersedes it.

**Verdict: Approve — no blocking issues found.**

PR #4647, opened 2026-08-17
Bug #76065, keep the color legend of elements that don't share it

ngovus-inetsoft:
## Problem

On a boxplot with multiple measures and a per-measure texture bound, the box body's color legend disappears — it is merged into the outlier texture/shape legend, so only one combined legend renders instead of separate "Color" and "Shape"/texture legends.

## Root cause

A boxplot builds a `SchemaElement` (box) and a `PointElement` (outliers) **per measure**, and each element gets its own equal-valued `MultiMeasureColorFrame`. The outlier elements carry the texture frame, so they share their colors into the texture legend and their color frames are correctly suppressed by this block in `EGraph.getVisualFrames()`:

```java
// don't show legend if color is already painted by other legends.
for(VisualFrame sharedColor : sharedColors) {
   if(sharedColor != null && otherIds.contains(sharedColor.getShareId())) {
      set.remove(sharedColor);   // matched by equals(), not identity
   }
}
```

`MultiMeasureColorFrame.equals()` is value-based, so once 127d4c7f3 (Bug #72971) made `set` graph-wide, this `remove()` also deleted the *box* elements' distinct color frames — which are not painted by that legend and are entitled to their own.

The two bugs pull in opposite directions, and the real discriminator is identity vs. equality, not scope:

- **#72971** — several elements share the *same* color frame instance, so suppression must be graph-wide.
- **#76065** — each element owns a *distinct but equal* clone, so suppressing one must not remove the others.

## Fix

Candidates stay graph-wide (preserving #72971) but are tracked in an `IdentityHashMap`-backed set, and the suppression moved out of the element loop to run once with `removeIf(frame -> frame == sharedColor)`. Only the frame actually painted by another legend is removed, and the outcome no longer depends on element order.

## Testing

New `EGraphVisualFramesTest` pins **both** directions, since each has historically regressed the other. Every test was validated to fail on the code that broke it:

| test | `main` | pre-#72971 | this PR |
|---|---|---|---|
| `boxplotKeepsColorLegendWithTextureLegend` | ✗ | ✓ | ✓ |
| `equalColorFrameOfOtherElementKeepsItsLegend` | ✗ | ✗ | ✓ |
| `sharedColorFrameDoesNotGetItsOwnLegend` (#72971) | ✓ | ✗ | ✓ |
| `boxplotWithoutTextureHasSingleColorLegend` | ✓ | ✓ | ✓ |

I also compared legend output across 11 chart configurations (boxplot ±texture, shared/distinct color frames with the shape frame on either element, stacked bar, bar, line, point) against `main`. The result is **identical to `main` in every case except the two #76065 cases**, including all #72971 shared-instance cases and the stacked-measures path.

77 tests across the graph, chart, legend and binding suites pass.

Note on the `isIdenticalStackedMeasuresFrame` tail dedup: stacked charts share a single frame *instance* across elements, so identity suppression still removes it and no color frame survives into that path. The frames that newly survive are per-element clones, which are not `StackedMeasuresFrame`.

### DIFF
--- a/core/src/main/java/inetsoft/graph/EGraph.java
+++ b/core/src/main/java/inetsoft/graph/EGraph.java
@@ -367,7 +367,11 @@ public class EGraph implements Cloneable, Serializable {
       Map<Object, VisualFrame> vmap = new HashMap<>();
       Set<String> otherIds = new HashSet<>(); // non-color-frame
       List<VisualFrame> sharedColors = new ArrayList<>();
-      Set<VisualFrame> set = new HashSet<>();
+      // candidate legend frames of all elements, in element order. tracked by identity since
+      // different elements may carry distinct but equal frames, each of which may still need
+      // its own legend. equal frames are merged by vmap below. (76065)
+      List<VisualFrame> candidates = new ArrayList<>();
+      Set<VisualFrame> added = Collections.newSetFromMap(new IdentityHashMap<>());
       Map<Integer, Integer> frameElemIdx = new HashMap<>();
 
       for(int i = 0; i < getElementCount(); i++) {
@@ -399,18 +403,24 @@ public class EGraph implements Cloneable, Serializable {
                }
             }
 
-            set.add(frames[j]);
-         }
-
-         // don't show legend if color is already painted by other legends.
-         for(VisualFrame sharedColor : sharedColors) {
-            if(sharedColor != null && otherIds.contains(sharedColor.getShareId())) {
-               set.remove(sharedColor);
+            if(added.add(frames[j])) {
+               candidates.add(frames[j]);
             }
          }
       }
 
-      for(VisualFrame frame : set) {
+      // don't show legend if color is already painted by other legends. the frame is matched by
+      // identity so that an equal color frame belonging to another element, which is not painted
+      // by that legend, keeps its own legend. (76065)
+      for(VisualFrame sharedColor : sharedColors) {
+         if(sharedColor != null && otherIds.contains(sharedColor.getShareId())) {
+            candidates.removeIf(frame -> frame == sharedColor);
+         }
+      }
+
+      // all shared color frames have been collected at this point, so the legends can be
+      // finalized. (72971)
+      for(VisualFrame frame : candidates) {
          List<VisualFrame> key = new ArrayList<>();
          key.add(frame);
          VisualFrame sharedColor = null;
--- a/core/src/test/java/inetsoft/graph/EGraphVisualFramesTest.java
+++ b/core/src/test/java/inetsoft/graph/EGraphVisualFramesTest.java
@@ -1,0 +1,219 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (approved):
### Code Review: Bug #76065 — keep the color legend of elements that don't share it

*(Posting this manually — the automated `claude-review` GitHub Action errored out on this PR without producing a review.)*

**Files reviewed:** `core/src/main/java/inetsoft/graph/EGraph.java`, `core/src/test/java/inetsoft/graph/EGraphVisualFramesTest.java` (new)

#### Root cause verification

Confirmed against `main`: `getVisualFrames()` collected legend candidates into a `HashSet<VisualFrame>` (`set`), and the "don't show legend if painted by another legend" suppression called `set.remove(sharedColor)`. Since `MultiMeasureColorFrame.equals()`/`hashCode()` are value-based, `HashSet.remove()` matches by *equality*, not identity — so suppressing one element's shared color frame also silently removed every other element's distinct-but-equal color frame from the candidate set, even though those frames aren't painted by any other legend and are entitled to their own. This exactly matches the reported boxplot-with-per-measure-texture symptom (box body's color legend disappearing).

#### Fix correctness

- Candidates are now collected into an order-preserving `List<VisualFrame>` (`candidates`), de-duplicated only by *identity* via an `IdentityHashMap`-backed `added` set — so two distinct instances that happen to be equal are both kept as separate candidates.
- The suppression loop now does `candidates.removeIf(frame -> frame == sharedColor)` — identity-based removal — so only the actual instance that is painted by another legend is dropped; equal-but-distinct sibling frames survive.
- Equal candidates that were *not* individually suppressed are still correctly merged into a single legend afterward via the existing `vmap` keyed on `List<VisualFrame>` (which uses `equals()`), so the #72971 "shared same instance" case still collapses to one legend, and general value-equal-frame merging (unrelated to sharing) is preserved.
- Moving the suppression loop to run once after the full element loop (instead of once per outer-loop iteration) is a behavior-preserving simplification: `sharedColors`/`otherIds` are only ever consumed after being fully populated in the final iteration anyway, so this is equivalent to the old code's net effect, just without redundant re-scans on every iteration.

I traced this against the four pinned scenarios in the new test and the described 11-configuration manual comparison in the PR description, and the logic holds up:
- Distinct equal color frames, only one shared into another legend → only the shared one is suppressed (76065, fixed).
- Same color frame instance shared by two elements, one paints it into a shape/texture legend → suppressed once, no duplicate legend (72971, preserved).
- Boxplot without texture → single merged color legend (unchanged path).

#### Other checks

- No security implications — pure in-memory legend/frame bookkeeping, no external input.
- No backward-incompatible API changes — `getVisualFrames()` signature and return semantics (dedup by equals) are unchanged for all cases that don't hit this specific corner case.
- Test coverage is solid: the new `EGraphVisualFramesTest` pins both directions (76065 and 72971) plus the no-texture baseline, and per the PR description each assertion was validated to fail against the code that introduced/reintroduced the corresponding bug.
- No existing tests are broken by this change (only additive test file + internal method-body rewrite of unchanged public signature).

**Verdict:** No blocking issues. `build`, `Semgrep Scan`, `Semgrep OSS`, and `license/cla` all pass, and there are no merge conflicts. The only failing check is the automated `claude-review` Action itself, which errored out without reviewing the code (this comment supersedes it). Approving.

PR #4648, opened 2026-08-17
Bug #76066, fix circular network node fixed color mapping dialog not opening

jkidacki-inetsoft:
## Summary

For a circular network (relation) chart, binding a plain dimension column (e.g. "state") as the node color field and clicking "assign fixed mapping" in the node pane's color editor silently failed to open the dialog — no error, no console output.

## Root Cause

`ColorMappingDialogService.getColorMappingDialogModel()` looks up the aesthetic ref via `GraphUtil.getChartAesRef(cinfo, dimensionName, false)`, which searches `cinfo.getAestheticRefs(false)`. For relation charts, `AbstractChartInfo.getAestheticRefs(ChartBindable)` intentionally excludes the node color field when it is `VSDimensionRef`-backed (a deliberate fix for bug #75253, to keep it out of the query's GROUP BY). Because "state" is a dimension (not an aggregate), the lookup returned `null`, and the service returned `null` after logging a warning. The Angular frontend (`categorical-color-pane.component.ts`) filters out `null` responses before opening the dialog (`filter(data => data.body != null)`), so the failure was silent.

## Fix

In `ColorMappingDialogService.getColorMappingDialogModel`, when the standard `getChartAesRef` lookup misses and `cinfo instanceof RelationChartInfo`, fall back to `RelationChartInfo.getNodeColorField()` directly (matched by name via `GraphUtil.equalsName`). This only changes what the color-mapping dialog can find — `getAestheticRefs()` itself is untouched, so the GROUP BY exclusion from bug #75253 remains intact.

## Test Plan

- [x] Imported `circular_network.zip`, opened the viewsheet in Composer, edited the node pane's "state" color field, clicked "assign fixed mapping" — dialog now opens correctly.
- [x] `mvn compile -pl core -am` succeeds.
- [x] Code review found no issues.
- [x] User manually verified the fix resolves the reported behavior.

Fixes Redmine #76066.

### DIFF
--- a/core/src/main/java/inetsoft/web/binding/controller/ColorMappingDialogService.java
+++ b/core/src/main/java/inetsoft/web/binding/controller/ColorMappingDialogService.java
@@ -25,6 +25,7 @@ import inetsoft.report.composition.WorksheetEngine;
 import inetsoft.report.composition.graph.GraphUtil;
 import inetsoft.uql.viewsheet.*;
 import inetsoft.uql.viewsheet.graph.AestheticRef;
+import inetsoft.uql.viewsheet.graph.RelationChartInfo;
 import inetsoft.uql.viewsheet.graph.VSChartInfo;
 import inetsoft.uql.viewsheet.internal.DateComparisonUtil;
 import inetsoft.util.Tool;
@@ -76,6 +77,19 @@ public class ColorMappingDialogService {
       }
 
       AestheticRef cfield = (AestheticRef) GraphUtil.getChartAesRef(cinfo, dimensionName, false);
+
+      // Relation (e.g. circular network) chart node color fields bound to a dimension are
+      // excluded from getAestheticRefs() (bug #75253, to keep them out of the query GROUP BY),
+      // so getChartAesRef() above never finds them. Fall back to the node color field directly
+      // so the "assign fixed mapping" dialog can still open for it. (Bug #76066)
+      if(cfield == null && cinfo instanceof RelationChartInfo) {
+         AestheticRef nodeColorField = ((RelationChartInfo) cinfo).getNodeColorField();
+
+         if(nodeColorField != null && GraphUtil.equalsName(nodeColorField, dimensionName)) {
+            cfield = nodeColorField;
+         }
+      }
+
       List list = null;
 
       if(cfield == null) {


jshobe-inetsoft (approved):
## Code Review — PR #4648 (Bug #76066)

**Note:** The automated `claude-review` CI job errored out on this PR without producing a review, so this is a manual comprehensive review covering the same checklist.

### Change

`ColorMappingDialogService.getColorMappingDialogModel()` now falls back to `RelationChartInfo.getNodeColorField()` when the standard `GraphUtil.getChartAesRef(cinfo, dimensionName, false)` lookup returns `null` and the chart is a `RelationChartInfo` (circular network). This restores the "assign fixed mapping" dialog for a dimension-backed node color field, which `getAestheticRefs()` intentionally excludes per bug #75253 (to keep it out of the query GROUP BY).

- The fallback mirrors the exact pattern already used in `GraphUtil.getChartAesRef` (`ref != null && GraphUtil.equalsName(ref, fld)`), and `equalsName(DataRef, String)` accepts an `AestheticRef` the same way it's called elsewhere in `GraphUtil`, so the type usage is consistent with the existing codebase.
- The fallback only assigns `cfield` when the name actually matches, so it can't accidentally return an unrelated field.
- Downstream logic (`cfield.getDataRef() instanceof VSDimensionRef` → `browseDimensionData`) already handles a dimension-backed `cfield`; this code path was previously unreachable for relation charts and is now correctly exercised.
- `getAestheticRefs()` itself is untouched, so the bug #75253 GROUP BY exclusion is preserved — this is purely an additive fallback for dialog lookup, not a change to query behavior.

- Non-relation charts are unaffected (`cinfo instanceof RelationChartInfo` guard).
- Relation charts whose node color field is aggregate-backed are unaffected (the original `getChartAesRef` lookup already finds those; the fallback only triggers on a `null` result).

### Style / readability
- Small, well-scoped change with a clear comment explaining *why* the fallback is needed and cross-referencing both bug numbers.

### Security
- No new external input handling; purely internal lookup logic.

### Test coverage
- No automated test was added; verification is manual (per the PR's test plan: import `circular_network.zip`, reproduce, confirm dialog opens). Given the fix is a narrowly-scoped null-check fallback with an obvious repro, this is a minor gap but not blocking.

### Verdict
No blocking issues found. Build, Semgrep, and license/cla checks all pass; the only failing check is the `claude-review` bot itself erroring out (unrelated to code correctness — this review supersedes it). No merge conflicts. **Approving.**

PR #4649, opened 2026-08-17
Bug #76081, apply ranking set on a tree chart's Node Color/Node Size field

ngovus-inetsoft:
## Problem

In a tree (relation) chart, a Top N / Bottom N ranking set on the **Node Color** (or **Node Size**) field has no effect — all nodes are rendered instead of the top N.

Reproduced with the attached `tree.vso`: Source = `Customer:Region`, Target = `Customer:State`, Node Color = `Customer:State` with `Top 3 by Sum(Order:Paid)` and `rankPerGroup=true`. All 14 states render; only 3 per region should.

## Root cause

`AbstractChartInfo.getAestheticRefs()` deliberately excludes `VSDimensionRef`-backed node color/size fields (bug #75253) to keep the node dimension out of the `GROUP BY`; `ChartVSAQuery` re-adds the column later as a `MAX` aggregate instead.

The side effect is that those refs also never appear in `cinfo.getRTFields()`, so the ranking-merge loop in `ChartVSAQuery.createTableAssembly()` never sees them and the `RankingCondition` is silently dropped. The earlier tree chart ranking fix (#4186) only covered the Source/Target wells, so ranking set from the Node Color/Node Size well remained a no-op.

## Fix

Merge the node color/size `RankingCondition` inside the existing `RelationChartInfo` ranking block.

Scope guards, to avoid regressions:

- **Only when the node dimension is the same field as Source or Target.** Those are the only real groups in a relation chart query, so a ranking on any other node dimension can't be evaluated and is skipped rather than injected into the query as a condition on an ungrouped column.
- The existing `rankingFields` dedup prevents a ranking already merged from Source/Target from being applied twice on the same column.
- `mgrcond` is cleared when the ranking lands on Source while Target is bound, matching the existing non-innermost-dimension handling so per-group (`rankPerGroup`) ranking is still post-processed.
- Uses `getRTDataRef()` with a `getDataRef()` fallback, consistent with how `getRTFields()` resolves aesthetic refs.

No behavior change for charts without a `VSDimensionRef` node color/size field, or where that field carries no ranking.

## Testing

- `./mvnw -o compile -pl core` passes.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/ChartVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/ChartVSAQuery.java
@@ -1548,6 +1548,57 @@ public class ChartVSAQuery extends CubeVSAQuery implements BindableVSAQuery {
                mergeDimRanking((VSDimensionRef) targetField, cols, ainfo, rconds,
                                 rankingFields, usedGroups);
             }
+
+            // VSDimensionRef-backed node color/size fields are excluded from
+            // getAestheticRefs() (bug #75253) and so never appear in cinfo.getRTFields()
+            // either, which means a RankingCondition set on the Node Color/Node Size well
+            // would never be merged into the query. Merge it here. Only source and target
+            // are actual groups in a relation chart query (the node dimension is added as a
+            // MAX aggregate), so a ranking is only meaningful when the node dimension is the
+            // same field as source or target - ranking on an ungrouped column can't be
+            // evaluated. rankingFields dedup keeps a ranking already merged from
+            // source/target from being applied twice.
+            String sourceName = sourceField instanceof VSDimensionRef ?
+               ((VSDimensionRef) sourceField).getFullName() : null;
+            String targetName = targetField instanceof VSDimensionRef ?
+               ((VSDimensionRef) targetField).getFullName() : null;
+            List<AestheticRef> nodeRankRefs = new ArrayList<>();
+
+            if(cinfo instanceof RelationVSChartInfo) {
+               nodeRankRefs.add(((RelationVSChartInfo) cinfo).getNodeColorField());
+               nodeRankRefs.add(((RelationVSChartInfo) cinfo).getNodeSizeField());
+            }
+
+            for(AestheticRef nodeARef : nodeRankRefs) {
+               if(nodeARef == null) {
+                  continue;
+               }
+
+               DataRef nodeRef = nodeARef.getRTDataRef() != null ?
+                  nodeARef.getRTDataRef() : nodeARef.getDataRef();
+
+               if(!(nodeRef instanceof VSDimensionRef)) {
+                  continue;
+               }
+
+               VSDimensionRef nodeDim = (VSDimensionRef) nodeRef;
+               String nodeName = nodeDim.getFullName();
+               boolean isSource = nodeName != null && nodeName.equals(sourceName);
+               boolean isTarget = nodeName != null && nodeName.equals(targetName);
+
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):
### Code Review: Bug #76081 — apply ranking set on a tree chart's Node Color/Node Size field

**Scope:** `core/src/main/java/inetsoft/report/composition/execution/ChartVSAQuery.java` (+51/-0, single file).

**Root cause verification:** Confirmed against the code. `AbstractChartInfo.getAestheticRefs()` excludes `VSDimensionRef`-backed node color/size fields (per the referenced bug #75253 comment at line ~1442), so they never surface via `cinfo.getRTFields()`, and the existing ranking-merge block (added for #4186, source/target only) never sees them — matching the reported symptom that a ranking set on Node Color/Node Size is silently dropped for tree/relation charts.

**Fix correctness:**
- The new block only merges the node dimension's ranking when its `getFullName()` matches the current Source or Target field — the only two real `GROUP BY` groups in a relation chart query (node color/size dimensions are otherwise folded into a `MAX` aggregate a few lines above, not grouped). This correctly avoids injecting a `RankingCondition` on an ungrouped column, which `mergeDimRanking`'s group-lookup would otherwise silently no-op or misresolve.
- Reuses the pre-existing `rankingFields` dedup set shared across the x/y loop, source, and target merges, so a ranking already merged from Source/Target is not double-applied when Node Color/Size is bound to that same field.
- Correctly propagates `mgrcond = false` when the matched node ranking lands on the outer/Source grouping while Target is also bound, mirroring the existing non-innermost-dimension handling directly above.
- Uses `getRTDataRef()` with `getDataRef()` fallback consistently with how aesthetic refs are resolved elsewhere in this method.
- Guarded by `cinfo instanceof RelationVSChartInfo`, so no behavior change for non-relation charts, and by `nodeARef == null` / non-`VSDimensionRef` checks, so no behavior change for charts without a node color/size ranking.

**Minor, non-blocking observations:**
1. The new `isSource`/`isTarget` match uses plain `nodeName.equals(sourceName/targetName)`, whereas `mergeDimRanking`'s own group-lookup fallback (a few lines below, in the same file) additionally tolerates a table-qualified name via `gname.endsWith("." + vname)`. If a node color/size ref's `getFullName()` and the source/target ref's `getFullName()` ever differ only by table qualification, the new code would silently skip the ranking instead of merging it (whereas the same field bound directly as source/target would still resolve via that fallback). Likely rare in practice since both refs are typically derived from the same worksheet column, but worth keeping in mind if a similar report resurfaces with a qualified column name.
2. No new test coverage was added for this change (PR testing note is only `./mvnw -o compile -pl core`). The existing `ChartVSAQueryTest.java` is minimal and doesn't currently exercise ranking or relation-chart logic at all, so this isn't a regression in test coverage introduced by this PR, but a targeted unit test around `mergeDimRanking`/node ranking merge (if feasible given the private method and heavy setup) would help prevent a future regression here, similar to how #4186 apparently also shipped without one.

**Security / backward compatibility / regressions:** No security concerns (pure server-side query-condition construction, no new external input handling). No backward-incompatible changes — the new block is strictly additive and only activates when a node color/size dimension shares a name with Source/Target and carries a ranking condition; all other charts are unaffected. No existing tests are touched or broken.

**Checks:** `build`, `Semgrep Scan`, `Semgrep OSS`, and `license/cla` all pass. The `claude-review` bot check failed to complete (that's why this manual review exists) — not treated as a blocking check here. No merge conflicts (`MERGEABLE`).

**Verdict:** No blocking issues found. Approving.

PR #4650, opened 2026-08-17
Fix the selection converts' container guard: AND where it needed OR

larryliang-inetsoft:
Both selection conversions guard with `&&` where the body requires **both** conditions, so the guard lets through exactly the cases it exists to stop.

```java
// ComposerVSSelectionListService.convertToRangeSlider
if(!(assembly instanceof SelectionListVSAssembly) &&
   !(container instanceof CurrentSelectionVSAssembly)) { return null; }

final CurrentSelectionVSAssembly containerAssembly = (CurrentSelectionVSAssembly) container;
final String[] assemblies = containerAssembly.getAssemblies();   // unguarded
```

It fails in **both** directions, which is more than the obvious one:

| Case | Old behaviour |
|---|---|
| Right type, **no container** — a standalone selection list, the common case | `getContainer()` returns `null`, the cast of null succeeds, `getAssemblies()` → **NullPointerException** |
| Right type, container is a **Tab / GroupContainer** | **ClassCastException** on the cast |
| **Wrong** type, right container — e.g. a chart inside a selection container | container half satisfied ⇒ passes ⇒ **ClassCastException** casting to `SelectionListVSAssembly` |

`ComposerRangeSliderService.convertCSComponent` has the same defect with three tests chained instead of two.

## Why it survived

Unreachable from the Composer: both menu items are hidden unless the assembly is in a selection container — `visible: () => this.composer && !this.model.adhocFilter && this.inSelectionContainer` (`selection-list-actions.ts`) and `visible: () => this.composer && this.inSelectionContainer` (`range-slider-actions.ts`). **That precondition lives only in Angular.** Any caller that isn't the menu reaches the bug immediately.

Found while wrapping these endpoints for the composer plugin (`stylebi#4645`). The plugin tier now refuses the case with a named error; this guard is the backstop underneath it, and the right place for the invariant.

## The fix

Require both. The type test stays an either-or for `convertCSComponent`, which legitimately accepts a selection list **or** a time slider — only the container is mandatory:

```java
if((!(assembly instanceof SelectionListVSAssembly) &&
    !(assembly instanceof TimeSliderVSAssembly)) ||
   !(container instanceof CurrentSelectionVSAssembly)) { return null; }
```

Behaviour on the paths the Composer actually exercises is unchanged: a selection list or time slider inside a `CurrentSelectionVSAssembly` still converts. `return null` is kept rather than raised to an exception, since that matches the surrounding style and the UI never sends these cases.

## Tests

5 new tests — there were none for either service — and **they discriminate**: restoring either `&&` fails all five with the original exceptions, three `NullPointerException`s naming `containerAssembly` and two `ClassCastException`s. That includes the wrong-type-with-container case, which the old guard also got wrong.

**Composer + wiz sweep: 1695 tests, 183 classes, 0 failures.**

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/vs/objects/controller/ComposerRangeSliderService.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/objects/controller/ComposerRangeSliderService.java
@@ -71,14 +71,15 @@ public class ComposerRangeSliderService {
 
       final VSAssembly container = assembly.getContainer();
 
-      // The type test is an either-or; the container is mandatory, because the body below
-      // dereferences it unconditionally. Chaining all three with && let an assembly of the right type
-      // through with no container, whereupon the cast of getContainer()'s null succeeded and
-      // getAssemblies() threw NullPointerException -- or ClassCastException inside a Tab or
-      // GroupContainer. Unreachable from the Composer, which hides the menu item unless
-      // inSelectionContainer, but reachable by any other caller.
-      if((!(assembly instanceof SelectionListVSAssembly) &&
-          !(assembly instanceof TimeSliderVSAssembly)) ||
+      // The container is mandatory, because the body below dereferences it unconditionally, and
+      // createSelectionListVSAssembly's only caller (below) unconditionally casts to
+      // TimeSliderVSAssembly -- this converts a time slider TO a selection list, not the other
+      // way around, so a SelectionListVSAssembly must be refused here rather than accepted
+      // alongside it. Accepting either type (as this once did, mirroring
+      // ComposerVSSelectionListService's own guard for its own single type) let a
+      // SelectionListVSAssembly with a valid container reach the cast below and throw
+      // ClassCastException.
+      if(!(assembly instanceof TimeSliderVSAssembly) ||
          !(container instanceof CurrentSelectionVSAssembly))
       {
          return null;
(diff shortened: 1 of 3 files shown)


larryliang-inetsoft:
## Review — the AND→OR fix is correct, but the guard it produces still lets an incompatible type through

Checked the boolean-logic change by truth table (De Morgan's on both call sites) and it holds: `ComposerVSSelectionListService`'s `!A || !C` and `ComposerRangeSliderService.convertCSComponent`'s `(!A && !B) || !C` both correctly reduce to "proceed only when (right type) ∧ (right container)" for every combination. Good fix, correctly generalized past the one reported case.

But `convertCSComponent`'s type check still accepts *either* `SelectionListVSAssembly` or `TimeSliderVSAssembly` (line 80-81), and its only downstream path unconditionally casts to the latter:

```java
// ComposerRangeSliderService.java:111-112
private VSAssembly createSelectionListVSAssembly(Viewsheet vs, VSAssembly cobj) {
   TimeSliderVSAssembly slider = (TimeSliderVSAssembly) cobj;   // <-- unconditional
```

So `convertCSComponent(assembly = SelectionListVSAssembly, container = a valid CurrentSelectionVSAssembly)` now passes the (correctly-fixed) guard and throws `ClassCastException` on this cast. Not introduced by this PR — the three-term `&&` chain already accepted both types before this fix — but the new tests don't reach it either: `aStandaloneSelectionListIsDeclinedByTheRangeSliderConvertToo` uses `container = null`, so it's declined by the guard before ever reaching the cast, not because of the type check.

Given this is exactly the call path `#4645`'s plugin wrapper reaches (per this PR's own description of how the bug was found), worth either refusing the `SelectionListVSAssembly` case explicitly here (mirroring what `ComposerVSSelectionListService`'s guard does for its own type), or confirming there's a reason a selection list can never legitimately reach this method with a valid container that I'm missing.

PR #4651, opened 2026-08-17
Bug #76113, enforce the monitoring/summary/debug EM component permission

ngovus-inetsoft:
## Problem

With security enabled and multi-tenancy off, a user granted EM access but explicitly **denied** `Monitoring/Summary/Debug` on the Actions tab could still open `em/monitoring/summary/debug` and download both dumps.

The Debug page is registered as a permissible EM component by its `@Secured` decorator — `em-metadata.js` scrapes that into `view-components.json`, which is what renders the node on the Actions tab. But nothing enforced it:

1. **The route had no guard.** `summary.routes.ts` declared `path: "debug"` with no `canActivate` and no `data.permission*`, unlike every other secured EM route. The only check that ran was the parent `summary` route's guard, so `monitoring/summary` access alone was sufficient.
2. **The endpoints checked the wrong resource.** Both `@Secured` blocks in `DebugMonitoringController` declared `resource = "monitoring/summary"`, so `/em/monitoring/server/debug/key-value-storage-dump` and `/blob-path-dump` stayed directly fetchable even with the route guarded.

## Fix

- Add `canActivateDebugPage`, which reads `monitoring/summary/debug` from `/api/em/authz` and returns a `UrlTree` to `/monitoring/summary` when denied.
- Correct the resource on both endpoints to `monitoring/summary/debug`.
- Mark the node `hiddenForMultiTenancy: true`, matching `/monitoring/summary` — the blob dump enumerates the blobs of every organization, so it should not be reachable by an org admin.
- Add `summary.routes.spec.ts` covering the allow and deny/redirect paths.

### Why not the generic `authorizationGuard`?

Its deny path picks the first permitted sibling and calls `router.navigate` on it. The other 10 children of `monitoring/summary` (`heapMemory`, `cpuUsage`, `swapping`, …) are widget-level permission nodes with **no route**, so it would have navigated to a nonexistent path. Returning a `UrlTree` to the parent page is the correct deny behavior, and follows the existing custom-guard precedent in `log.routes.ts`.

## Side effects considered

`ComponentAuthorizationController.anyChildAvailable` hides a parent node when all of its children are denied, so adding an enforced child under `monitoring/summary` could in principle have hidden **Summary itself** from the monitoring sidenav. I regenerated `view-components.json` and confirmed `summary` still has its 10 widget children alongside `debug`, so denying only `debug` leaves Summary visible. `SecuredAspect.isComponentAccessible()` resolves the new resource string against that same tree; the `debug` node is present, so the change does not hit the deny-by-default path.

## Testing

- `npx ng test em --include=…/summary.routes.spec.ts` — 2/2 passing
- `npx tsc --noEmit -p projects/em/tsconfig.app.json` — clean
- `npx eslint` on all changed TS files — clean
- `npx gulp em:metadata` — node emits as `{"name":"debug","label":"Debug","children":{},"hiddenForMultiTenancy":true}`

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/server/DebugMonitoringController.java
+++ b/core/src/main/java/inetsoft/web/admin/server/DebugMonitoringController.java
@@ -53,7 +53,7 @@ public class DebugMonitoringController {
    @Secured(
       @RequiredPermission(
          resourceType = ResourceType.EM_COMPONENT,
-         resource = "monitoring/summary",
+         resource = "monitoring/summary/debug",
          actions = ResourceAction.ACCESS
       )
    )
@@ -99,7 +99,7 @@ public class DebugMonitoringController {
    @Secured(
       @RequiredPermission(
          resourceType = ResourceType.EM_COMPONENT,
-         resource = "monitoring/summary",
+         resource = "monitoring/summary/debug",
          actions = ResourceAction.ACCESS
       )
    )
--- a/web/projects/em/src/app/monitoring/summary/debug-monitoring-page/debug-monitoring-page.component.ts
+++ b/web/projects/em/src/app/monitoring/summary/debug-monitoring-page/debug-monitoring-page.component.ts
@@ -26,7 +26,8 @@ import { MatButton } from "@angular/material/button";
 
 @Secured({
    route: "/monitoring/summary/debug",
-   label: "Debug"
+   label: "Debug",
+   hiddenForMultiTenancy: true
 })
 @Component({
     selector: "em-debug-monitoring-page",
--- a/web/projects/em/src/app/monitoring/summary/summary.routes.spec.ts
+++ b/web/projects/em/src/app/monitoring/summary/summary.routes.spec.ts
@@ -1,0 +1,78 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+import { TestBed } from "@angular/core/testing";
+import {
+   ActivatedRouteSnapshot,
+   provideRouter,
+   Router,
+   RouterStateSnapshot,
+   UrlTree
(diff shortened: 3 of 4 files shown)


jshobe-inetsoft (approved):
### Code Review: Bug #76113 — enforce the `monitoring/summary/debug` EM component permission

**Note:** The automated `claude-review` GitHub Action check failed to complete on this PR (errored with no output) — this review was performed manually to fill that gap.

#### Correctness
Verified the root cause and fix directly against the current `main` branch source:

- `DebugMonitoringController.java` (both `getKVDump` and `getBlobPathsDump`) previously declared `resource = "monitoring/summary"` on their `@Secured` annotations — identical to the parent page's permission — so a user with only `monitoring/summary` access (and an explicit deny on the `debug` child) could still hit both dump endpoints directly. Changing the resource to `monitoring/summary/debug` on both endpoints correctly closes that gap and matches the resource string the EM Actions tab renders from `@Secured({route: "/monitoring/summary/debug", ...})` on the component.
- `summary.routes.ts` previously had no `canActivate` on the `debug` child route, so route-level access relied only on the parent `summary` guard. The new `canActivateDebugPage` guard queries `AuthorizationService.getPermissions("monitoring/summary")` and checks the `debug` child permission, redirecting to `/monitoring/summary` (via `UrlTree`) when denied.
- The custom-guard approach (rather than the generic `authorizationGuard`) is justified and matches existing precedent — `log.routes.ts` uses the same `CanActivateFn` + `map` pattern for a similar case where the generic guard's "redirect to first permitted sibling" fallback would not apply cleanly (here, the siblings are routeless widget nodes).
- `AuthorizationService.getPermissions` docstring confirms it returns permissions for *child* components of the given path, which matches how the guard and test consume the `debug` key from `getPermissions("monitoring/summary")`.
- `hiddenForMultiTenancy: true` on the component decorator matches the stated intent (hide from org admins, consistent with the parent `/monitoring/summary` node) and is additive UI metadata — doesn't change any existing behavior for single-tenant/site-admin cases.

#### Test coverage
`summary.routes.spec.ts` covers both branches of `canActivateDebugPage` (allow when permitted, redirect when denied) with a reasonably mocked `AuthorizationService`. No existing tests reference the changed backend resource strings, so nothing here appears to break existing coverage.

#### Backward compatibility / regressions
- The `@Secured` resource change is scoped to exactly the two endpoints described — verified no other endpoints in `DebugMonitoringController.java` reference `monitoring/summary` as a resource.
- Existing users who already have `monitoring/summary/debug` granted (the EM permission tree node has existed since the component was registered) are unaffected; this only tightens enforcement for cases where `debug` was explicitly denied while the parent was allowed.
- No API signature or response-shape changes.

#### Security
This is a net security improvement — it closes a permission-bypass path for a sensitive endpoint (dumps KV storage and blob paths across all organizations).

#### Minor, non-blocking notes
- `AuthorizationService.getPermissions` caches by path with `useCache=true` by default; if an admin's permissions change without a full reload, the guard could act on a stale cached value. This is pre-existing behavior shared by other guards/pages using the same service, not something introduced by this PR.

No blocking issues found. Checks (build, Semgrep, license/cla) are passing and there are no merge conflicts.

PR #3933, opened 2026-06-08
Bug #75252, prevent EM logout when Google OAuth2 SMTP sign-in fails on unlicensed instances

larryliang-inetsoft:
## Summary

In Enterprise Manager → Settings → General → Email, selecting **Google OAuth2 SMTP** and clicking **"Sign in with Google"** refreshed the page and bounced the admin back to the EM Summary page on opensource/unlicensed StyleBI instances.

## Root Cause

The Google OAuth2 SMTP (and SASL XOAuth2) sign-in flow acquires tokens through InetSoft's hosted OAuth proxy at `https://data.inetsoft.com/login`, which authenticates using the `license.key`. On unlicensed/opensource instances the proxy responds **HTTP 403**. The globally-registered `InvalidSessionInterceptor` taps every `HttpClient` response — including this cross-origin one — and on a 401/403 calls `LogoutService.sessionExpired()`, which does a full `document.location.replace("../sessionexpired?...")`. Since the real server session is still valid, EM bounces back to the Summary page.

Note: this flow is the same in Style Intelligence 13.6/13.7, where it works only because those installs have a valid commercial license that the proxy accepts.

## Fix

- **`invalid-session-interceptor.ts`** — only trigger `sessionExpired()` for **same-origin** responses. A 401/403 from an absolute URL pointing at an external host (e.g. `data.inetsoft.com`) is unrelated to the StyleBI session and no longer logs the EM admin out. (Protocol-relative URLs are normalized before the origin comparison.)
- **`email-settings-view.component.ts`** — `authorize0()` now surfaces a snackbar on authorization failure instead of failing silently.
- **`srinter.properties`** — added `em.mail.smtp.authorizeFailed` message.

This is a symptom-only fix (stop the spurious logout + show a clear error). Completing the OAuth flow still requires a valid commercial license, because the hosted proxy rejects unlicensed requests by design.

## Test Plan

1. Run an unlicensed/opensource (community) server.
2. EM → Settings → General → Email → SMTP Authentication Type → **Google OAuth2 SMTP**.
3. Enter any Gmail user / Client ID / Client Secret and click **"Sign in with Google"**.
4. **Expected:** you stay on the Email settings page and see *"Authorization failed. Please verify your OAuth settings and try again."* — you are **not** logged out / bounced to EM Summary.
5. Regression: same-origin 401/403 responses still trigger the normal session-expiry handling.

Fixes Redmine #75252.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4652, opened 2026-08-17
Bug #76095, prevent EM logout when a page requests a resource the user lacks permission for

ngovus-inetsoft:
## Issue

Enable security (multi-tenancy off), then:

1. Create `user0` and `user1`
2. Grant `user0` permission on **Portal Tabs / Schedule**
3. Grant `user0` permission on **Enterprise Manager**
4. Grant `user1` permission on **Enterprise Manager / Settings / Schedule / Tasks**
5. Log into the EM as `user0` and click the **Settings** tab

Expected: the Settings page opens. Actual: the EM logs the user out.

It looks intermittent, which is why the report suggests clearing the cache and re-entering the EM URL — see below.

## Root cause

Two defects stacked.

**1. Settings/General eagerly requests a task-gated endpoint.** Clicking Settings always lands on `/settings/general` (`settings.routes.ts` — `{ path: "**", redirectTo: "general" }`). That page renders `EmailSettingsViewComponent`, which injects the root-provided `ScheduleUsersService`; its *constructor* issues `GET /api/em/schedule/users-model`. That endpoint was secured on `EM_COMPONENT "settings/schedule/tasks"` alone, so `user0` gets a 403. The service's error callback is empty, so nothing surfaces in the UI.

**2. `InvalidSessionInterceptor` treated 403 as an expired session.** It fired `LogoutService.sessionExpired()` on any same-origin **401 or 403**, hard-redirecting to `../sessionexpired`. That is the logout.

Because the service is a singleton constructed once, the request never repeats after the first bootstrap — hence the apparent flakiness and the "clear the cache" workaround.

This surfaced as a regression from #4355 (Bug #75734), which correctly changed `@Secured` denials from 500 to 403. Before it, the interceptor ignored the 500. That commit is right; it just exposed the latent interceptor bug.

## Fix

**`invalid-session-interceptor.ts` — only 401 triggers `sessionExpired()`.**

Only 401 indicates a lost session: when a session expires or is invalidated, the filter chain answers XHR requests with 401 (`AbstractSecurityFilter.shouldSendAuthenticationRedirect`), and this interceptor stamps `X-Requested-With: XMLHttpRequest` on every request, so that path always applies. A 403 means the session is valid but the user is not permitted to access that resource.

This is the important half — the same failure mode was latent across a large set of schedule/task-secured EM endpoints (`EMScheduleTaskController`, `EMScheduleTaskActionController`, `EMScheduleBatchActionController`, `ExportTaskController`, `ImportTaskController`, `EMScheduleTaskFolderController`, …). It also un-breaks call sites that already handle 403 themselves but were losing the race against the logout, e.g. the "Unauthorized" dialog in `schedule-task-list.component.ts`.

**`EMScheduleController.getUsersModel` — widen the permission to match its callers.**

The user/group list is not specific to schedule tasks: it also backs the email recipient pickers on the General settings page and the Security settings pages, none of which require any schedule permission. Now requires `settings/general` **OR** `settings/schedule/tasks` **OR** `settings/security/users` — the same set already used by `/api/em/settings/schedule/check-mail`.

`/api/em/schedule/task-names` is deliberately left alone: `ScheduleTaskNamesService` is only injected by the task-editor pages, which are already behind the tasks permission.

**`navbar.component.ts`** — corrected a comment that claimed the interceptor redirects on 403.

## Testing

New `invalid-session-interceptor.spec.ts`, 7 tests, passing via `ng test em`:

- 401 logs out (server's signal for a lost session)
- **403 does not log out** (regression-sensitive)
- 500 / 404 do not log out
- cross-origin 401 does not log out (preserves #3933 / Bug #75252)
- absolute same-origin 401 does log out
- GET requests carry `X-Requested-With: XMLHttpRequest`
- success responses do not log out

`./mvnw -o compile -pl core` passes clean.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/schedule/EMScheduleController.java
+++ b/core/src/main/java/inetsoft/web/admin/schedule/EMScheduleController.java
@@ -221,12 +221,32 @@ public class EMScheduleController {
       this.usersChangeService.addSubscriber(stompHeaderAccessor);
    }
 
+   /**
+    * The user/group list returned here is not specific to schedule tasks: it also backs the
+    * email recipient pickers on the General settings page and the Security settings pages, all
+    * of which are reachable without any schedule permission. Gating it on
+    * "settings/schedule/tasks" alone made simply opening Settings fail for such users. Mirrors
+    * the permission set of {@code /api/em/settings/schedule/check-mail}.
+    */
    @Secured(
-      @RequiredPermission(
-         resourceType = ResourceType.EM_COMPONENT,
-         resource = "settings/schedule/tasks",
-         actions = ResourceAction.ACCESS
-      )
+      value = {
+         @RequiredPermission(
+            resourceType = ResourceType.EM_COMPONENT,
+            resource = "settings/general",
+            actions = ResourceAction.ACCESS
+         ),
+         @RequiredPermission(
+            resourceType = ResourceType.EM_COMPONENT,
+            resource = "settings/schedule/tasks",
+            actions = ResourceAction.ACCESS
+         ),
+         @RequiredPermission(
+            resourceType = ResourceType.EM_COMPONENT,
+            resource = "settings/security/users",
+            actions = ResourceAction.ACCESS
+         )
+      },
+      operator = "OR"
    )
    @GetMapping("/api/em/schedule/users-model")
    public UsersModel getUsersModel(@PermissionUser Principal principal) throws Exception
--- a/web/projects/em/src/app/invalid-session-interceptor.spec.ts
+++ b/web/projects/em/src/app/invalid-session-interceptor.spec.ts
@@ -1,0 +1,97 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import { HTTP_INTERCEPTORS, HttpClient } from "@angular/common/http";
+import { HttpClientTestingModule, HttpTestingController } from "@angular/common/http/testing";
+import { TestBed } from "@angular/core/testing";
+import { LogoutService } from "../../../shared/util/logout.service";
+import { InvalidSessionInterceptor } from "./invalid-session-interceptor";
+
+describe("InvalidSessionInterceptor", () => {
+   let http: HttpClient;
+   let httpMock: HttpTestingController;
+   let logoutService: { sessionExpired: any };
+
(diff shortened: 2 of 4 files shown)


jshobe-inetsoft (approved):
### Code Review: Bug #76095 — prevent EM logout when a page requests a resource the user lacks permission for

*(Note: the automated `claude-review` CI check on this PR failed to complete — it appears to be a tooling/CI issue, not a finding about this code. This is a manual review to replace it.)*

#### Root-cause verification

I traced the claims in the PR description against the actual code and they check out:

- `AbstractSecurityFilter.shouldSendAuthenticationRedirect` (`core/src/main/java/inetsoft/web/security/AbstractSecurityFilter.java:679-694`) sends a 401 for any XHR/websocket request (or non-navigation request) when the session is unauthenticated/expired — confirming 401 is the server's actual "session is gone" signal.
- `AdminExceptionHandler.accessDenied` (`core/src/main/java/inetsoft/web/admin/AdminExceptionHandler.java`) maps both the checked `SecurityException` and `SecuredAspect`'s unchecked `java.lang.SecurityException` (thrown on any `@Secured` permission denial) to a 403 — confirming 403 in this codebase means "authenticated but not permitted," never session expiry.
- `EMScheduleController.getUsersModel`'s old single-permission `@Secured` (`settings/schedule/tasks`) is exactly what a user with only `settings/general` or `settings/security/users` access would fail, and `EmailSettingsViewComponent` (reached via the `**` → `general` default route in `settings.routes.ts`) does inject `ScheduleUsersService`, which fires this request from its constructor — confirming the trigger path.
- The widened permission set (`settings/general` OR `settings/schedule/tasks` OR `settings/security/users`, `operator = "OR"`) is an exact match for the existing pattern on `/api/em/settings/schedule/check-mail` (`SchedulerConfigurationController.java:96-104`), so this isn't a new/ad-hoc permission scheme — it's reusing an established one for the same underlying user/group list.
- `schedule-task-list.component.ts:459-460` already has its own 403 → "Unauthorized" dialog handler, confirming the PR's claim that this fix un-breaks an existing 403 handler that was losing the race against the interceptor's logout.
- `navbar.component.ts`'s comment update accurately reflects the new behavior (401 only triggers the interceptor's redirect; 403 is a separate, permission-related case).

#### Code quality

- `invalid-session-interceptor.ts`: the fix is a single, well-targeted condition change (`401 || 403` → `401`), with the doc comment clearly explaining *why* (not just what). `isSameOrigin`/`intercept()` (X-Requested-With stamping, same-origin check) are unchanged by this PR and out of scope here.
- `EMScheduleController.java`: new Javadoc on `getUsersModel` clearly explains why the permission is now an OR of three components, matching the surrounding code's style.
- New `invalid-session-interceptor.spec.ts` (7 tests) directly covers the regression case (403 no longer logs out), the normal case (401 does), cross-origin isolation, and the XHR header stamping — good, targeted coverage using `HttpClientTestingModule`/`HttpTestingController` consistent with this codebase's testing conventions.

#### Regressions / backward compatibility

- No other call site in the diff or elsewhere relies on the interceptor's previous "403 also logs out" behavior as intended behavior — the PR description calls out that this was itself a latent bug across many schedule/task-secured EM endpoints, and the one call site that does explicitly handle 403 (`schedule-task-list.component.ts`) is fixed rather than broken by this change.
- Widening `getUsersModel`'s permission is strictly additive (OR of a superset) — no existing caller that could reach it under the old, narrower permission loses access.
- No security concern: the widened permission still requires *some* EM settings access; it does not open the endpoint to unauthenticated or fully unprivileged users.

#### Verdict

No blocking issues found. All substantive checks (build, Semgrep OSS, Semgrep Scan, license/cla) pass, and there are no merge conflicts (`mergeable: MERGEABLE`). Approving.

PR #4653, opened 2026-08-18
Serve connector endpoint catalogues to the wiz portal

Nelson-cao:
Adds an API that returns a connector's endpoint catalogue, so the wiz portal has something to index and retrieve on. This is the StyleBI half of Tier 1 tabular annotation.

**The consumer is inetsoft-technology/stylebi-wiz#1643**, which indexes what this returns and retrieves against it. This PR merges first — nothing here depends on that one.

`POST /api/wiz/datasources/endpoint-catalog` takes a list of data source types and answers with one catalogue each:

```jsonc
{ "types": ["Rest.Stripe", "Rest.Github"] }
```
```jsonc
{
  "catalogs": { "Rest.Stripe": { "endpoints": [ { "name": "Balance Transactions",
      "description": "Every movement of money through the Stripe balance …",
      "suffix": "/v1/balance_transactions?…", "paged": true, "lookups": [ … ] } ] } },
  "notCatalogued": ["Rest.Foo"],
  "unavailable": ["Rest.Bar"]
}
```

## Keyed by type, not by data source

`endpoints.json` is one file per connector, so three Stripe connections in an organization share one catalogue. Asking per connection would return the same 110 entries three times, and the wiz side would index them three times over.

## The four buckets are not interchangeable

`notCatalogued` means the connector ships no `endpoints.json`. That is the normal state of a REST data source whose endpoints can only come from user-supplied documentation, and the portal has to ask the user for that documentation.

`unavailable` means the type IS registered but its query class could not be loaded — a connector plugin that is not installed — or the resource exists and does not parse. That is an environment problem and goes away when the plugin is installed.

`unknownType` means the type string has no entry in the registry at all: a client typo, or a type from a newer build than this server. That is **permanent** — no plugin install fixes it, and a caller must not retry it the way it retries `unavailable`.

Splitting the last two is not guesswork; the signals were already distinct and were being discarded. `Config.getQueryClass` is a plain map lookup that answers null for an unregistered type, whereas a registered type gets past it and fails later in `getClass`. Before the split, a typo was handed to the portal as a transient environment fault: it would keep asking for a type that will never exist, and an operator would be pointed at plugin installation for a client-side bug.

Collapsing any of these would send the portal to act on a fault that is not the one it has. This is the same distinction `WizDatasourceEntry` already draws between `UNKNOWN` and `UNSUPPORTED`: a classification failure must never be readable as a classification result.

## Reading the resource, not calling the loader

`EndpointCatalogReader` parses `endpoints.json` off `queryClass.getResourceAsStream(...)` with core's own mapper rather than going through `EndpointJsonQuery.Endpoints.load`. Three reasons: `load` binds each connector's own `Endpoints` subclass, which core cannot reference; it reads `SreeEnv`; and it converts a parse failure into an empty map — the right trade for a query path, the wrong one for a catalogue read, where "empty" and "broken" must stay distinguishable. `read` returns `null` for an absent resource and lets an `IOException` propagate for a broken one.

Review found two documents that slipped between those two states. A resource whose whole content is the JSON literal `null` deserializes to a null object, and the log line then dereferenced it — so the method escaped with an NPE instead of the `IOException` its own javadoc declares, and the caller's only evidence was a log message reading `null` with no stack. And `{}` produced a catalogue with `endpoints == null`, which is neither the documented "empty" nor the documented "broken": it was returned as a success and put straight into `catalogs`, so the response carried `"endpoints": null` and any consumer iterating it dereferenced a null. The first now throws; the second is normalized to an empty list, because `{}` is structurally valid JSON that simply declares no endpoint.

## The catalogue model ignores unknown properties, which is the reverse of the loader's requirement

Sixteen connectors declare endpoint properties of their own — `pageType`, `post`, `bodyTemplate`, `pageLimit`, `pagePath`, `pageRequiredParameter`, `paginationPath`, `freePageLimit`, `url`. The catalogue binds a five-field subset, so a strict mapper here would take those connectors' catalogues to zero while `EndpointsJsonLoadableTest`'s original case stayed green.

`EndpointsJsonLoadableTest` gains a second case that parses all 65 files through the catalogue model. Removing the `@JsonIgnoreProperties` annotation turns it red naming the affected connectors.

That case originally asserted only that every entry has a `name`, which review pointed out is the one field Tier 1 retrieval does not use: rename a record component or misspell a JSON name and Jackson binds `description`, `suffix` and `lookups` to null for all 4035 entries while the test stays green — the exact silent degradation the original case in this file exists to prevent. It now also asserts a non-null `suffix` and `description` per file, and non-empty `lookups` on a connector that has them, in both the array and the single-string spelling. Renaming the `suffix` component turns it red.

## Both lookup spellings reach the consumer as they are

A lookup names its targets as either `endpoint` (a single string, GitHub) or `endpoints` (an array, Stripe and Zendesk). `WizEndpointLookup` carries both fields and passes both through unchanged; the collapsing happens in the consumer.

An earlier revision of this branch had a `targets()` accessor here that claimed to reconcile them. Review established it was dead: `targets` is not a record component, so Jackson never put it on the wire, and nothing called it. It has been removed and the javadoc corrected — the consumer already collapses both spellings and has tests for each, so exposing a derived `targets` array would only add a third representation of the same fact next to the two raw fields.

## No permission gate

A catalogue is compiled into the connector jar and is identical for every customer, so there is no instance, no path and no stored data to authorize against.

Review questioned this, and an earlier revision answered it with a paragraph about what the wiz authentication filter guarantees. That paragraph was wrong on the mechanism and, more to the point, not this controller's concern — authentication for `/api/wiz/**` is settled by the filter chain, not here. It has been removed rather than corrected.

`WizDatabaseControllerSecurityTest` pins the set of mapped endpoints, so the new path is added to that frozen list; no authorization assertion changed.

## Testing

```
EndpointsJsonLoadableTest            2 tests  — all 65 files parse through both models, and bind
                                                description, suffix and lookups rather than name alone
WizDatabaseControllerSecurityTest   19 tests  — the frozen mapping contract plus the type-classification
                                                cases: null/blank filtered, unregistered vs unloadable
./mvnw -pl core -am -DskipTests install        — BUILD SUCCESS
```

Every behaviour this branch relies on was mutation-checked rather than assumed. Removing `@JsonIgnoreProperties` reds the catalogue-parse case; dropping the null/blank filter, or collapsing `unknownType` back into `unavailable`, reds the corresponding controller cases; renaming the `suffix` record component reds the new field assertions.

`stylebi-wiz-main-rebase` has been merged in. Its only conflict was the import block of `WizDatabaseControllerSecurityTest`, where the base added a folder-create endpoint while this branch added the catalogue one. The parts of that test that carry risk merged cleanly and were verified rather than assumed: the frozen mapped-path set holds both sides' new entries, and the fixture absorbed both sides' new constructor arguments.

### DIFF
--- a/connectors/inetsoft-rest/src/test/java/inetsoft/uql/rest/json/EndpointsJsonLoadableTest.java
+++ b/connectors/inetsoft-rest/src/test/java/inetsoft/uql/rest/json/EndpointsJsonLoadableTest.java
@@ -18,6 +18,8 @@
 package inetsoft.uql.rest.json;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import inetsoft.web.wiz.model.WizEndpointCatalog;
+import inetsoft.web.wiz.model.WizEndpointCatalogEntry;
 import org.junit.jupiter.api.Test;
 
 import java.io.*;
@@ -93,6 +95,52 @@ class EndpointsJsonLoadableTest {
       assertTrue(failures.isEmpty(),
                  "endpoints.json failed to parse for " + failures.size() + " connector(s):\n  "
                     + String.join("\n  ", failures));
+   }
+
+   /**
+    * The same 65 files must ALSO parse through the wiz catalog DTO, which has the opposite
+    * requirement to the loader above: it binds a subset of the properties and must therefore
+    * IGNORE the rest. Fifteen connectors declare extras of their own (pageType, post, bodyTemplate,
+    * pageLimit, pagePath, pageRequiredParameter, paginationPath, freePageLimit, url); a strict
+    * mapper here would take those connectors' catalogs to zero while the loader stayed green.
+    */
+   @Test
+   void everyConnectorEndpointsFileParsesAsCatalog() throws Exception {
+      List<Path> roots = findDatasourceRoots();
+      List<Path> files = findEndpointFiles(roots);
+
+      assertFalse(files.isEmpty(),
+                  "found no endpoints.json under " + DATASOURCE_PACKAGE + " (searched " + roots + ")");
+
+      ObjectMapper mapper = new ObjectMapper();
+      List<String> failures = new ArrayList<>();
+
+      for(Path file : files) {
+         String connector = file.getParent().getFileName().toString();
+
+         try(InputStream input = Files.newInputStream(file)) {
+            WizEndpointCatalog catalog = mapper.readValue(input, WizEndpointCatalog.class);
+
+            if(catalog.endpoints() == null || catalog.endpoints().isEmpty()) {
+               failures.add(connector + ": parsed but declares no endpoints");
+               continue;
+            }
+
+            for(WizEndpointCatalogEntry entry : catalog.endpoints()) {
+               if(entry.name() == null || entry.name().isBlank()) {
+                  failures.add(connector + ": an entry has no name");
+                  break;
+               }
+            }
+         }
+         catch(Exception e) {
+            failures.add(connector + ": " + e.getMessage());
+         }
+      }
+
+      assertTrue(failures.isEmpty(),
+                 "endpoints.json failed to parse as a catalog for " + failures.size()
+                    + " connector(s):\n  " + String.join("\n  ", failures));
    }
 
    /**
--- a/connectors/inetsoft-rest/src/test/java/inetsoft/uql/rest/json/EndpointsJsonLoadableTest.java
+++ b/connectors/inetsoft-rest/src/test/java/inetsoft/uql/rest/json/EndpointsJsonLoadableTest.java
@@ -20,6 +20,7 @@ package inetsoft.uql.rest.json;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import inetsoft.web.wiz.model.WizEndpointCatalog;
 import inetsoft.web.wiz.model.WizEndpointCatalogEntry;
+import inetsoft.web.wiz.model.WizEndpointLookup;
 import org.junit.jupiter.api.Test;
 
 import java.io.*;
@@ -115,6 +116,14 @@ class EndpointsJsonLoadableTest {
       ObjectMapper mapper = new ObjectMapper();
       List<String> failures = new ArrayList<>();
 
+      // name is a poor guard against a renamed or misspelled JSON key: it is the one field every
+      // one of the 4035 entries across all 65 files always carries, and the portal's Tier 1
+      // retrieval does not depend on it. A binding failure on suffix, description, or lookups --
+      // the fields retrieval actually uses -- would null them out for every connector while this
+      // check alone stayed green. The per-field checks below close that gap.
+      boolean sawStripeLookupWithEndpointsArray = false;
+      boolean sawGithubLookupWithSingularEndpoint = false;
+
       for(Path file : files) {
          String connector = file.getParent().getFileName().toString();
 
@@ -126,16 +135,71 @@ class EndpointsJsonLoadableTest {
                continue;
             }
 
+            boolean hasSuffix = false;
+            boolean hasDescription = false;
+
             for(WizEndpointCatalogEntry entry : catalog.endpoints()) {
                if(entry.name() == null || entry.name().isBlank()) {
                   failures.add(connector + ": an entry has no name");
                   break;
                }
             }
+
+            for(WizEndpointCatalogEntry entry : catalog.endpoints()) {
+               if(entry.suffix() != null && !entry.suffix().isBlank()) {
+                  hasSuffix = true;
+               }
+
+               if(entry.description() != null && !entry.description().isBlank()) {
+                  hasDescription = true;
+               }
+
+               if(entry.lookups() != null) {
+                  for(WizEndpointLookup lookup : entry.lookups()) {
+                     if("stripe".equals(connector) && lookup.endpoints() != null
+                        && !lookup.endpoints().isEmpty())
+                     {
+                        sawStripeLookupWithEndpointsArray = true;
+                     }
+
+                     if("github".equals(connector) && lookup.endpoint() != null
+                        && !lookup.endpoint().isBlank())
+                     {
+                        sawGithubLookupWithSingularEndpoint = true;
+                     }
+                  }
+               }
+            }
+
+            // Every one of the 65 files has a non-blank suffix on at least one entry; a renamed or
+            // misspelled "suffix" JSON key would null it out everywhere while the name-only check
+            // above stayed green.
+            if(!hasSuffix) {
+               failures.add(connector + ": no entry has a non-blank suffix");
+            }
+
+            // The description backfill has in fact reached every connector: every entry in every
+            // one of the 65 files already carries a non-blank description, so this is asserted for
+            // all of them rather than only stripe and github.
+            if(!hasDescription) {
+               failures.add(connector + ": no entry has a non-blank description");
+            }
          }
          catch(Exception e) {
             failures.add(connector + ": " + e.getMessage());
          }
+      }
+
+      // Both lookup spellings are exercised in the wild -- Stripe/Zendesk write an "endpoints"
+      // array, GitHub writes a singular "endpoint" -- and WizEndpointLookup carries both fields as
+      // written (see its javadoc). A binding regression on either field would only show up here,
+      // never in the entry-level checks above.
+      if(!sawStripeLookupWithEndpointsArray) {
+         failures.add("stripe: no lookup bound a non-empty \"endpoints\" array");
+      }
+
+      if(!sawGithubLookupWithSingularEndpoint) {
+         failures.add("github: no lookup bound a non-blank singular \"endpoint\"");
       }
 
       assertTrue(failures.isEmpty(),
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
@@ -252,21 +252,8 @@ public class WizDatabaseController {
     * <p>POST rather than GET because the type list is unbounded, matching {@code
     * /datasources/search} and {@code /datasources/statuses} in this same controller.</p>
     *
-    * <p>No permission gate, but that reasoning only covers half the response. {@code catalogs} is a
-    * property of the CONNECTOR — identical for every customer, compiled into the connector jar,
-    * carrying nothing about any one of them. There is no instance, no path and no stored data to
-    * authorize against, so no permission applies. {@code notCatalogued}, {@code unavailable} and
-    * {@code unknownType} are not that: they report which connector plugins THIS deployment has
-    * installed and how its type registry is configured, which is instance-specific information a
-    * customer-neutral catalogue is not. Do not extend the "no instance to authorize against"
-    * reasoning to those three fields — it does not hold for them.</p>
-    *
-    * <p>The actual boundary protecting this endpoint today is authentication, not permission: the
-    * whole {@code /api/wiz/**} prefix sits behind {@code WizServiceAuthenticationFilter}'s bearer
-    * token check, so the exposure is "any authenticated wiz caller can enumerate this server's
-    * installed connector plugins by POSTing every known {@code Rest.*} type", not "anyone on the
-    * network can". Whether that residual exposure warrants its own permission is a separate policy
-    * decision, not made here — this handler does not even take a {@code Principal} yet.</p>
+    * <p>No permission gate: a catalogue is compiled into the connector jar and is identical for
+    * every customer, so there is no instance, no path and no stored data to authorize against.</p>
     *
     * <p>{@code unavailable} and {@code unknownType} answer two different questions and must not be
     * merged. Both start from {@code Config.getQueryClass(type)}, a plain map lookup that either
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
@@ -47,6 +47,7 @@ import inetsoft.web.security.RequiredPermission;
 import inetsoft.web.security.Secured;
 import inetsoft.web.wiz.model.*;
 import inetsoft.web.wiz.request.*;
+import inetsoft.web.wiz.service.EndpointCatalogReader;
 import inetsoft.web.wiz.service.UnsupportedDatasourceException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -103,7 +104,8 @@ public class WizDatabaseController {
                                 DatabaseTypeService databaseTypeService,
                                 SecurityEngine securityEngine,
                                 Config uqlConfig,
-                                XRepository xrepository)
+                                XRepository xrepository,
+                                EndpointCatalogReader endpointCatalogReader)
    {
       this.dataSourceBrowserService = dataSourceBrowserService;
       this.dataSourceStatusService = dataSourceStatusService;
@@ -112,6 +114,7 @@ public class WizDatabaseController {
       this.securityEngine = securityEngine;
       this.uqlConfig = uqlConfig;
       this.xrepository = xrepository;
+      this.endpointCatalogReader = endpointCatalogReader;
    }
 
    /**
@@ -239,6 +242,67 @@ public class WizDatabaseController {
       }
 
       return result;
+   }
+
+   /**
+    * The endpoint catalogues for a set of data source types.
+    *
+    * <p>POST rather than GET because the type list is unbounded, matching {@code
+    * /datasources/search} and {@code /datasources/statuses} in this same controller.</p>
+    *
+    * <p>No permission gate: a catalogue is a property of the CONNECTOR, identical for every
+    * customer and carrying nothing about any one of them. There is no instance, no path and no
+    * stored data to authorize against. The listing endpoints, which do expose what an organization
+    * has configured, remain gated as before.</p>
+    *
+    * @param request the types to look up. An absent or empty list yields an empty answer rather
+    *                than an error — asking about nothing is not a fault.
+    */
+   @PostMapping(value = "/datasources/endpoint-catalog", produces = MediaType.APPLICATION_JSON_VALUE)
+   public WizEndpointCatalogResponse getEndpointCatalog(@RequestBody WizEndpointCatalogRequest request) {
+      Map<String, WizEndpointCatalog> catalogs = new LinkedHashMap<>();
+      List<String> notCatalogued = new ArrayList<>();
+      List<String> unavailable = new ArrayList<>();
+      List<String> types = request.getTypes() == null ? List.of() : request.getTypes();
+
+      for(String type : new LinkedHashSet<>(types)) {
+         Class<?> queryClass;
+
+         try {
+            String className = uqlConfig.getQueryClass(type);
+            queryClass = className == null ? null : uqlConfig.getClass(type, className);
+         }
+         catch(Throwable ex) {
+            // A connector whose plugin is absent cannot be read. That is an environment problem,
+            // not a verdict about the connector, so it must not land in notCatalogued.
+            LOG.debug("Failed to load the query class for type {}: {}", type, ex.getMessage());
+            queryClass = null;
+         }
+
+         if(queryClass == null) {
+            unavailable.add(type);
+            continue;
+         }
+
+         try {
+            WizEndpointCatalog catalog = endpointCatalogReader.read(queryClass);
+
+            if(catalog == null) {
+               notCatalogued.add(type);
+            }
+            else {
+               catalogs.put(type, catalog);
+            }
+         }
+         catch(Exception ex) {
+            // The resource is there but unreadable. Reporting it as "has no catalogue" would send
+            // the portal to ask the user for documentation it does not need.
+            LOG.warn("Failed to read the endpoint catalogue for type {}: {}", type, ex.getMessage());
+            unavailable.add(type);
+         }
+      }
+
+      return new WizEndpointCatalogResponse(catalogs, notCatalogued, unavailable);
    }
 
    /**
@@ -1073,6 +1137,7 @@ public class WizDatabaseController {
    private final SecurityEngine securityEngine;
    private final Config uqlConfig;
    private final XRepository xrepository;
+   private final EndpointCatalogReader endpointCatalogReader;
 
    /** Annotation classes. See {@link WizDatasourceEntry} for what each one means to the portal. */
    private static final String JDBC = "JDBC";
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
@@ -252,10 +252,33 @@ public class WizDatabaseController {
     * <p>POST rather than GET because the type list is unbounded, matching {@code
     * /datasources/search} and {@code /datasources/statuses} in this same controller.</p>
     *
-    * <p>No permission gate: a catalogue is a property of the CONNECTOR, identical for every
-    * customer and carrying nothing about any one of them. There is no instance, no path and no
-    * stored data to authorize against. The listing endpoints, which do expose what an organization
-    * has configured, remain gated as before.</p>
+    * <p>No permission gate, but that reasoning only covers half the response. {@code catalogs} is a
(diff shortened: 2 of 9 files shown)


core/src/main/java/inetsoft/web/wiz/model/WizEndpointLookup.java
```diff
@@ -0,0 +1,51 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record WizEndpointLookup(
+   String endpoint,
+   List<String> endpoints,
+   String jsonPath,
+   String key,
+   String parameterName)
+{
+   /** Both spellings collapsed into one list. Never null. */
+   public List<String> targets() {
```
Nelson-cao: **[medium] `targets()` is unreachable - the normalization it promises never leaves the server.**

`targets` is not a record component, so Jackson does not treat it as a property. Verified by serializing this exact record with jackson-databind 2.16.1:

```
{"endpoint":"Issue Event","endpoints":null,"jsonPath":"$.x","key":"id","parameterName":"p"}
{"endpoint":null,"endpoints":["Disputes","Refunds"],"jsonPath":"$.x","key":"id","parameterName":"p"}
```

No `targets` field in either. And `grep -rn "targets()"` over the repo returns only this definition and the javadoc that references it - there is no caller.

Concrete consequence: the class javadoc and the PR description both say core reconciles the two spellings because the connector's deserializer modifier cannot be reused here. It does not. The portal receives GitHub's 325 single-`endpoint` lookups and Stripe/Zendesk's 249 `endpoints`-array lookups in two different shapes and has to write the normalization anyway - with no test on either side covering it.

Either expose it (`@JsonProperty` on a derived accessor, and consider dropping the raw pair from the payload) or delete the method and state plainly that the consumer normalizes.

core/src/main/java/inetsoft/web/wiz/service/EndpointCatalogReader.java
```diff
@@ -0,0 +1,68 @@
...
+    * @throws java.io.IOException when the resource exists but does not parse. A broken catalogue is
+    *                             reported, never silently rendered as an empty one.
+    */
+   public WizEndpointCatalog read(Class<?> queryClass) throws java.io.IOException {
+      try(InputStream input = queryClass.getResourceAsStream(RESOURCE)) {
+         if(input == null) {
+            return null;
+         }
+
+         WizEndpointCatalog catalog = MAPPER.readValue(input, WizEndpointCatalog.class);
```
Nelson-cao, line 60: **[low] The log line NPEs on a resource whose content is the JSON literal `null`, breaking this method's own contract.**

`new ObjectMapper().readValue("null", WizEndpointCatalog.class)` returns `null` (verified). Line 60 then calls `catalog.endpoints()` in the log arguments and throws NPE, so `read` escapes with an NPE rather than the `IOException` the javadoc above declares for a resource that "exists but does not parse".

The controller's `catch(Exception)` still routes it to `unavailable`, so nothing 500s - but the caller-visible evidence is `Failed to read the endpoint catalogue for type X: null` with no stack trace, which is the hardest possible form of "broken catalogue" to diagnose, in a class whose whole justification is keeping "empty" and "broken" distinguishable.

Null-check `catalog` before the log statement and throw for a null document.

core/src/main/java/inetsoft/web/wiz/service/EndpointCatalogReader.java
```diff
@@ -0,0 +1,68 @@
...
+         LOG.debug("Read {} endpoint(s) for {}",
+                   catalog.endpoints() == null ? 0 : catalog.endpoints().size(), queryClass);
+         return catalog;
```
Nelson-cao, line 82: **[low] A document with no `endpoints` array is returned as a successful catalogue with `endpoints == null`.**

`readValue("{}", WizEndpointCatalog.class)` yields `WizEndpointCatalog[endpoints=null]` (verified) - non-null, so it is returned here, and `getEndpointCatalog` puts it straight into `catalogs`. The response then carries `"catalogs": {"Rest.X": {"endpoints": null}}`.

That is a fourth state none of the three documented buckets covers, and the consumer that iterates `catalogs[type].endpoints` to index it dereferences a null. It also undercuts the class rationale: `{}` is neither the documented "empty" nor the documented "broken".

Normalize a missing array to `List.of()` here, or treat it as broken and let it land in `unavailable`.

core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
```diff
@@ -243,6 +246,78 @@ public List<WizDatasourceStatus> getDatasourceStatuses(
...
+         }
+         catch(Throwable ex) {
+            // A connector whose plugin is absent cannot be read. That is an environment problem,
+            // not a verdict about the connector, so it must not land in notCatalogued.
+            LOG.debug("Failed to load the query class for type {}: {}", type, ex.getMessage());
+            queryClass = null;
+         }
+
+         if(queryClass == null) {
+            unavailable.add(type);
```
Nelson-cao, line 328: **[low] An unrecognized type string is reported as `unavailable`, i.e. as a missing plugin.**

`Config.getQueryClass` is a plain `dxmap.get(type)` on a `HashMap`, so a type that was never registered - a client typo such as `"Rest.Strip"`, or a type from a newer build - returns null and reaches this line indistinguishably from a connector whose plugin failed to load.

`WizEndpointCatalogResponse`'s javadoc tells the portal that `unavailable` is "an environment problem and goes away when the plugin is installed". So a typo is handed to the portal as a transient, retryable environment fault: it will keep asking for a type that will never exist, and the operator is pointed at plugin installation for a client-side bug.

This is the same distinction the PR argues so carefully for between `notCatalogued` and `unavailable` - "a classification failure must never be readable as a classification result" - applied one level up. Either add an `unknownType` bucket, or document explicitly that `unavailable` also means "no such type".

core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
```diff
@@ -243,6 +246,78 @@ public List<WizDatasourceStatus> getDatasourceStatuses(
       return result;
    }
 
+   /**
+    * The endpoint catalogues for a set of data source types.
+    *
+    * <p>POST rather than GET because the type list is unbounded, matching {@code
+    * /datasources/search} and {@code /datasources/statuses} in this same controller.</p>
+    *
+    * <p>No permission gate: a catalogue is a property of the CONNECTOR, identical for every
```
Nelson-cao: **[low] The "no permission gate" rationale does not hold for two of the three response fields.**

The argument is that a catalogue "carr[ies] nothing about any one [customer]". True of `catalogs` - the payload is compiled-in connector metadata. Not true of `notCatalogued` and `unavailable`: those report which connector plugins are installed in *this* deployment.

Concretely, any caller that can reach `/api/wiz/**` can POST a list of every known `Rest.*` type and read back the server's installed-plugin inventory from the `unavailable` array. This is also the only handler in this controller that takes no `Principal` at all, so there is nothing to gate on even conditionally.

The fix does not need a DATA_SOURCE gate on the catalogue itself - but the per-deployment half of the answer should either be gated or the rationale in this javadoc corrected so the next reader does not extend "no gate needed" to a genuinely instance-specific field.

Nelson-cao (commented):
Careful single-pass review of the diff. Six findings, one medium and five low. Everything below was checked against the actual resource files and, where behaviour was in question, against jackson-databind directly rather than from memory.

Verified as correct while reviewing: all 65 `endpoints.json` files parse through the catalog model (ran it), `paged` coerces from the string form used by 1412 entries, every lookup `endpoint`/`endpoints` value in the tree is a string, the resource is resolved relative to the query class exactly as `Endpoints.load` and `classifyQueryClass` already do, `Config.dxmap` is a plain `HashMap` so the `getQueryClass(null)` comment is accurate, `inetsoft-core` is `provided` scope in `inetsoft-rest` so the new test import compiles, and `EndpointCatalogReader` is inside the `inetsoft.web` component scan.

connectors/inetsoft-rest/src/test/java/inetsoft/uql/rest/json/EndpointsJsonLoadableTest.java
```diff
@@ -95,6 +97,52 @@ void everyConnectorEndpointsFileParses() throws Exception {
...
+
+         try(InputStream input = Files.newInputStream(file)) {
+            WizEndpointCatalog catalog = mapper.readValue(input, WizEndpointCatalog.class);
+
+            if(catalog.endpoints() == null || catalog.endpoints().isEmpty()) {
+               failures.add(connector + ": parsed but declares no endpoints");
+               continue;
+            }
+
+            for(WizEndpointCatalogEntry entry : catalog.endpoints()) {
```
Nelson-cao, line 141: **[low] The new case asserts only `name`, which is the one field Tier 1 retrieval does not depend on.**

`description`, `suffix` and `lookups` are what the portal indexes and retrieves against, and none of them is checked. Rename a record component in `WizEndpointCatalogEntry`, or misspell a JSON name, and Jackson binds it to null for all 4035 entries across all 65 files - while this test stays green, because every entry still has a `name`. The portal would then index 65 catalogues of names with no descriptions and no suffixes, which is exactly the silent-degradation failure the original case in this file was written to prevent.

Add assertions that each file yields at least one non-null `description` and `suffix`, and that a lookup-bearing connector (stripe or github) yields non-empty `lookups` - the second of those would also have caught the `targets()` issue flagged on `WizEndpointLookup`.

core/src/main/java/inetsoft/web/wiz/model/WizEndpointLookup.java
Nelson-cao: Confirmed dead. `grep -rn "targets()" --include="*.java"` over the whole repo returns two hits: the method itself and the javadoc line that references it. No caller.

One correction to the diagnosis, because it changes the remedy. The normalization is not missing and is not untested — it just lives on the consumer, which is where the only caller is:

- `wiz-services/src/services/tabular/endpointCatalogClient.ts:66-68` collapses both spellings: `raw.endpoints?.length ? raw.endpoints : (raw.endpoint ? [raw.endpoint] : [])`.
- Covered by `endpointCatalogClient.test.ts:79` ("accepts the singular lookup spelling", GitHub's `{"endpoint": "Issue Event"}`) and again at `:142`, whose fixture carries `endpoint: null, endpoints: ["Refunds"]` to model what a Java record with no `@JsonInclude` actually puts on the wire.

So the defect is narrower than "the portal has to write the normalization anyway with no test": it is dead code plus two pieces of prose that claim credit for work the consumer does.

Fixing it by deleting `targets()` and correcting the class javadoc and the PR description, not by exposing it. Exposing would put a third representation of the same fact on the wire next to the two raw fields, and the sole consumer already collapses them — a `@JsonProperty` derived accessor would give it a `targets` array to prefer plus an `endpoint`/`endpoints` pair to keep ignoring.

core/src/main/java/inetsoft/web/wiz/service/EndpointCatalogReader.java
Nelson-cao: Correct, and it breaks the contract the javadoc right above it declares. Adding a null check that throws `IOException` for a null document, before the log statement, so "exists but does not parse" stays the documented outcome instead of an NPE that the controller's `catch(Exception)` then flattens into an unexplained `unavailable`.

core/src/main/java/inetsoft/web/wiz/service/EndpointCatalogReader.java
Nelson-cao: Correct. Normalizing a missing array to `List.of()` in the reader, so `{}` becomes the documented "empty" rather than a fourth state.

For the record on blast radius: the current consumer is null-safe here — `endpointCatalogClient.ts` reads `(catalog?.endpoints ?? []).map(...)` — so nothing breaks today. That does not make `"endpoints": null` acceptable on the wire; it means the defect is currently absorbed by one specific caller rather than surfaced.

core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
Nelson-cao: The two cases are distinguishable in code, which makes this sharper than the comment suggests — the information is being discarded, not guessed at:

- `Config.getQueryClass(type)` is `dxmap.get(type)` and returns null when the type was never registered → unknown type.
- A registered type whose plugin failed to load gets past that and throws from `uqlConfig.getClass(type, className)` → genuine environment problem.

The current code collapses both into `unavailable` by catching one and null-checking the other into the same bucket. An `unknownType` bucket would therefore be precise, not heuristic.

One consequence before I add it, and the reason I want your call: the wiz client reads the three fields it knows by name. If this repo starts returning a fourth bucket and the consumer is not updated in the same wave, a typo'd type stops appearing in the wiz-side ingest report altogether — today it at least shows up under `unavailable`. Silently dropping it is worse than mislabeling it.

So: add `unknownType` here **and** a passthrough in inetsoft-technology/stylebi-wiz#1643 (small — one field on the response type, one on `IngestReport`, one line in the route), or document in `WizEndpointCatalogResponse` that `unavailable` also covers "no such type" and leave the split for later? I will not expand into the other PR without you saying so.

core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
Nelson-cao: The facts hold. `catalogs` is compiled-in connector metadata; `notCatalogued` and `unavailable` are per-deployment, and enumerating every known `Rest.*` type does read back an installed-plugin inventory. Confirmed separately that this is the only handler in the controller with no `Principal` parameter at all, so there is currently nothing to gate on.

For calibration rather than as a defence: `/api/wiz/**` is not anonymous — `WizServiceAuthenticationFilter` validates a wiz bearer token on that prefix, which is the same gate every other endpoint in this controller sits behind. So the exposure is to any authenticated wiz caller, not to the internet.

Two ways to close it and I would rather you pick, because one is a product decision:

1. Correct the javadoc rationale only — state that `catalogs` is customer-independent while the two failure buckets report this deployment's installed connectors, so the next reader does not extend "no gate needed" to an instance-specific field.
2. Also gate it, which means taking a `Principal` and asserting something — and it is not obvious what: there is no data source path to check, so it would have to be a coarse permission, which is a new policy decision rather than a repair.

Doing 1 regardless. Tell me if you want 2 on top.

connectors/inetsoft-rest/src/test/java/inetsoft/uql/rest/json/EndpointsJsonLoadableTest.java
Nelson-cao: Correct, and it is the same failure shape the original case in this file exists to prevent — a green test over a catalogue that bound to nothing but names. Adding: every file yields at least one non-null `description` and `suffix`, and a lookup-bearing connector (stripe) yields non-empty `lookups`.

One correction on the last clause: the lookups assertion would not have caught the `targets()` issue. This test only deserializes — `mapper.readValue(file, endpointsClass)` — so it exercises JSON→record binding. `targets()` failing to appear is a *serialization* property of the record, on the response side, which nothing in this file touches. A non-empty-lookups assertion would catch a renamed or misspelled `lookups`/`jsonPath`/`key` component; it would stay green whatever `targets()` does.

core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
Nelson-cao: Decided: adding the bucket, in both PRs together.

This repo gains `unknownType` on `WizEndpointCatalogResponse`, split on the signal that already exists — `getQueryClass` returning null means the type was never registered, whereas getting past that and having `getClass` throw means a registered type whose plugin did not load. inetsoft-technology/stylebi-wiz#1643 gains the passthrough in the same wave, so a typo cannot vanish from the wiz-side ingest report.

(Re-posted: the first version of this reply lost every backtick-quoted identifier to a shell-quoting mistake on my side.)

core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
Nelson-cao: Decided: correcting the rationale, not adding a gate.

The javadoc will state that `catalogs` is compiled-in connector metadata and customer-independent, while `notCatalogued` and `unavailable` — and now `unknownType` — report what this deployment has installed. So the "nothing to authorize against" argument covers the catalogue payload only, and the next reader does not extend it to the per-deployment fields.

Not gating, because what to check is a policy decision rather than a repair: there is no data source path here, so it would have to be a coarse permission this controller does not otherwise use, and the endpoint already sits behind the wiz bearer token like every other handler in the class. Worth raising as its own change if you want that gate.

(Re-posted: the first version of this reply lost every backtick-quoted identifier to a shell-quoting mistake on my side.)

core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
```diff
@@ -243,6 +246,123 @@ public List<WizDatasourceStatus> getDatasourceStatuses(
...
+    * {@code unknownType} are not that: they report which connector plugins THIS deployment has
+    * installed and how its type registry is configured, which is instance-specific information a
+    * customer-neutral catalogue is not. Do not extend the "no instance to authorize against"
+    * reasoning to those three fields — it does not hold for them.</p>
+    *
+    * <p>The actual boundary protecting this endpoint today is authentication, not permission: the
+    * whole {@code /api/wiz/**} prefix sits behind {@code WizServiceAuthenticationFilter}'s bearer
+    * token check, so the exposure is "any authenticated wiz caller can enumerate this server's
+    * installed connector plugins by POSTing every known {@code Rest.*} type", not "anyone on the
+    * network can". Whether that residual exposure warrants its own permission is a separate policy
```
Nelson-cao: This paragraph states a security property the filter does not provide, and it is the premise the "no permission gate" decision now rests on.

`WizServiceAuthenticationFilter` is an authenticator, not a gate. Two paths through `doFilter` let an unauthenticated request straight through to this handler:

- No `Authorization` header: `if(authHeader == null || !authHeader.startsWith(BEARER_PREFIX)) { chain.doFilter(request, response); return; }` — the comment there says "No token provided, let other filters handle authentication". It does not 401.
- `wiz.auth.enabled=false` (`SreeEnv`, defaults true but is settable): the filter returns via `chain.doFilter` before looking at the header at all.

So a bearer token is verified when one is presented, but nothing in this prefix requires one. `AbstractSecurityFilter.isWizRequest` matches `/api/wiz/**` only to keep wiz sessions from being invalidated (`InvalidateSessionFilter`); it does not impose a token check either. Whether an anonymous caller is rejected depends on the deployment's general security configuration, which this javadoc does not establish.

The residual exposure is therefore not bounded by "any authenticated wiz caller" the way the text claims. Suggest narrowing the wording to what the code supports — a bearer token is validated when present, and the anonymous case falls through to the standard chain — or taking a `Principal` and gating the three deployment-specific fields.

Nelson-cao (commented):
Re-reviewed the three fix commits (1323254c8..89a741fd4). Five of the six findings are resolved and verified:

- `targets()` removed; the singular/array spellings stay as written and `endpointCatalogClient.ts` does the normalization (confirmed it reads both `endpoint` and `endpoints`).
- A JSON `null` document now throws `IOException` instead of NPE-ing in the `LOG.debug` argument.
- `{}` / `{"endpoints": null}` is normalized to an empty list in the reader, so the fourth undocumented state is gone.
- `unknownType` separates a client typo / version mismatch from a plugin that failed to load, with tests covering both directions; `endpointCatalogClient.ts` already knows the field.
- `EndpointsJsonLoadableTest` now asserts `suffix`, `description` and both lookup spellings. I ran the equivalent checks over all 65 `endpoints.json` (4035 entries): every connector has a non-blank suffix and non-blank descriptions, and the stripe/github lookup flags both hold, so the new assertions pass.

One new item on the permission-gate paragraph, inline.

Also correcting my own earlier remark: this is not the only handler in the controller without a `Principal` — `getDatabaseMeta` has none either.

core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
Nelson-cao: Verified against the filter, and the claim in that paragraph does not hold. `WizServiceAuthenticationFilter.doFilter`:

- lines 93-96 — `if(!isWizAuthEnabled()) { chain.doFilter(request, response); return; }`, and `isWizAuthEnabled` is `"true".equalsIgnoreCase(SreeEnv.getProperty(WIZ_AUTH_ENABLED_PROPERTY, "true"))` (line 159). Default true, but settable, and when it is off the filter never looks at the header.
- lines 101-105 — `if(authHeader == null || !authHeader.startsWith(BEARER_PREFIX)) { chain.doFilter(request, response); return; }`, with the filter's own comment reading "No token provided, let other filters handle authentication". No 401.

A token is validated when one is presented; nothing on this prefix requires one. So "the exposure is to any authenticated wiz caller rather than to anyone" is not something this filter establishes, and neither is the same sentence in the PR description.

This matters beyond the wording. That sentence was the premise I put in front of my human partner when the gate-or-document question was decided — the decision to correct the rationale and not gate was taken on the understanding that the endpoint was already authenticated-only. That premise was mine and it was wrong, so I am reopening the decision with them rather than quietly narrowing the text to fit a choice made on bad information.

Two things will change regardless of how that lands: the javadoc will describe only what the code does (a bearer token is validated when present; the anonymous case falls through to the standard chain, and whether it is rejected depends on the deployment's general security configuration), and the PR description loses the same claim. Whether the three deployment-specific fields also get a `Principal` and a gate is what I am putting back to them.

Nelson-cao: Resolved, and the resolution corrects me as well as the paragraph.

From the maintainer: the wiz filter calling `chain.doFilter` on a tokenless request is not the request reaching this handler unauthenticated — it hands off to the rest of the chain, and authentication is enforced there, on user information available from the session. So neither my sentence ("bounded to any authenticated wiz caller") nor the reading it invited ("anonymous callers reach the handler") describes the system.

The decision is to remove the discussion rather than fix it. Authentication on `/api/wiz/**` is settled by the filter chain and is not this controller's concern, so a javadoc paragraph reasoning about it will drift out of date and mislead whoever reads it next — which is what happened here within a single review round.

`65b05b0de` cuts those two paragraphs down to the one sentence that is actually about this endpoint: a catalogue is compiled into the connector jar and identical for every customer, so there is nothing instance-specific to authorize against. No gate, no `Principal`. The same claim is out of the PR description.

19 tests still green; the change is javadoc only.

core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
```diff
@@ -243,6 +246,110 @@ public List<WizDatasourceStatus> getDatasourceStatuses(
...
    }
 
+   /**
+    * The endpoint catalogues for a set of data source types.
+    *
+    * <p>POST rather than GET because the type list is unbounded, matching {@code
+    * /datasources/search} and {@code /datasources/statuses} in this same controller.</p>
+    *
+    * <p>No permission gate: a catalogue is compiled into the connector jar and is identical for
+    * every customer, so there is no instance, no path and no stored data to authorize against.</p>
```
Nelson-cao, line 256: Removing the authentication paragraph also took out the half of the earlier text that was about this endpoint rather than about the filter chain: `notCatalogued`, `unavailable` and `unknownType` are not compiled-in connector metadata — they report which connector plugins this deployment has installed and how its type registry is configured. The retained sentence justifies `catalogs` only, but reads as covering the whole response.

That caveat is independent of who enforces authentication, so it does not carry the drift risk the auth paragraph did. Suggest keeping one clause of it, e.g. "…nothing instance-specific to authorize against in `catalogs`; the three type lists do reflect this deployment's installed plugins." Non-blocking.

Nelson-cao (commented):
Re-reviewed `65b05b0de`. Javadoc-only, and it resolves the authentication point: the incorrect claim is gone, and I accept the maintainer's correction that a tokenless request handed to the rest of the chain is authenticated there — my reading that it reaches the handler unauthenticated was wrong. Agreed that this controller's javadoc is the wrong place to reason about the filter chain.

One leftover from the cut, inline. Nothing else outstanding on my side — the other five findings stayed fixed.

fredaan (approved):

PR #4654, opened 2026-08-18
Pane-scoped pairing: scope an agent session to one script location

larryliang-inetsoft:
Makes an agent pairing session scopeable to a single **script location** — one chart's script, one
calculated field's formula, one worksheet column's expression — instead of a whole sheet.

Two reasons it exists. Expression-level locations **cannot be named at sheet level** at all, so
they were unreachable by any agent. And a whole-sheet write handle is far more authority than
editing one formula requires: a session minted to fix a typo in a chart's `onClick` could rewrite
every script on the sheet.

Implements Phase 3 of `2026-08-14-script-plugin-scope-and-script-kinds.md`. Client half:
inetsoft-technology/stylebi-wiz#1644.

## Three enforcement layers, and only the third is a policy

1. **Addressability** — `Kind.requiresPaneSession()`. An expression-level kind asked for from a
   whole-sheet session fails with the *reason*, never the policy name: *"a calculated field has no
   name at sheet level — open its formula editor and click Connect Agent."* A user who cannot act
   on an error has been told nothing.
2. **Grant scope** — `PaneScopeService.matchesGrant`, **unconditional**. A pane session refusing a
   target outside its grant is the scope of the grant, not a configurable posture, and it does not
   weaken when the strict flag is off. If it did, a session minted for one chart's script could
   edit another's.
3. **Strict posture** — `wiz.agent.script.require-script-pane`, default `false`. The only policy
   here; it *additionally* refuses viewsheet-level kinds from a whole-sheet session.

## Validation happens at mint, not at join

An `editorContext` naming an assembly the runtime lacks fails while the user is still looking at
the editor that produced it. At join the same failure surfaces to an agent minutes later with no
path back to the user.

That change had a cost worth stating: mint began resolving a caller-supplied `runtimeId` and
inspecting it, and the ownership check lived at join. So a user could mint against a **foreign**
runtime and learn from the outcome whether an assembly existed on someone else's open sheet. Fixed
by checking ownership first — and the refusal is deliberately the *same kind and message* a missing
runtime produces, never varying with what was asked for. A check that refused foreign runtimes with
a distinct message would have closed the ownership hole and left the oracle standing.

## Two findings worth reading even if you skim the diff

**A guard planted three tasks early caught the exact defect it was built for.** `requiresPaneSession()`
is set membership, not `location() == null`. Landing the reserved worksheet kinds gives them real
`Location` values — under a null-check both would have flipped silently from pane-scoped to
whole-sheet-addressable, compiling cleanly, every test green. The planted test failed with
`worksheetExpression ==> expected: <true> but was: <false>` exactly as intended.

**A stale comment did the same damage where no guard existed.** `matchesGrant`'s `CALC_FIELD`-only
branch carried a comment asserting the worksheet kinds "carry no assembly/name" — true when
written, false once they became servable. A grant on one column would have matched a target naming
a **different column**. A comment is an assumption with no enforcement; the mechanical half was
caught by a test, this half only by re-reading the branch being edited.

## Scope discipline

Worksheet writes route onto `WorksheetAgentController`'s existing `edit_expression` /
`edit_condition` ops — a front door, not a second writer, because that op is what keeps undo/redo
and the refresh broadcast coherent. Only expression **text** is writable; `type`, `sql`, and column
existence are refused with a redirect to worksheet-chat, before any request is built.

A pane session also dies with its editor (`ngOnDestroy` → detach, plus socket loss). Without that,
"only while a script pane is active" is enforced at mint time only, and a user who pairs then
cancels leaves a live write handle behind. Whole-sheet sessions keep today's TTL untouched — that
regression is pinned by its own test, because leaking pane lifetime into toolbar sessions would log
out every existing user on a socket blip.

## Testing

**1722 tests, 0 failures**, full `inetsoft.web.wiz.**` on the merged tree, from a clean build.

The merge with `feature-76074-server-fixes` is worth noting: git reported **no conflicts** and the
result **did not compile**. New base-side tests construct `JoinSession` with nine arguments and this
branch made it ten; E2/E3 inserted four services into a constructor whose test helper predates them.
Different lines of the same files — not a textual conflict, and only compiling finds it.

Every test in this branch was verified by reverting the **behaviour** it covers and confirming it
fails and names itself, not by observing a green run. That standard found several tests that were
green for the wrong reason, including one that passed under its own mutation because it asserted a
call succeeded rather than which target reached the wire.

**Known limit, stated rather than glossed:** the Angular figures come from scoped `--include` runs.
`portal:test-tl` cannot be run as a whole (a pre-existing circular import unrelated to this work)
and the top-level `npm test` omits that target entirely.

## Post-review fix wave (`18d5f27ee`)

The whole-branch review found four defects **at the seams between tasks** — every per-task review
had passed. Worth listing, because they are the interesting ones:

1. **Critical: a pane-scoped code was still a whole-sheet write handle.** The check was reached only
   from the script surfaces; the worksheet `edit` dispatcher, the assembly controller and ~25
   binding endpoints resolved the same token and never read `editorContext`. Worst path:
   `open_base_worksheet` minted a *new* whole-sheet session with `editorContext = null`, laundering
   a narrow grant into an unscoped handle that socket-loss would never reap. The code there carried
   a comment explaining why the new session was null — correct about the value, blind to the
   consequence.
2. **`detach` keyed on the live context, not the minted one.** Toggle the onInit/onLoad radio, then
   Cancel, and detach sent a context matching nothing; the session lived out its TTL with the editor
   gone. Its revert-check is the tell: the two new tests fail and *the three pre-existing detach
   tests stay green* — they vary the context, but never after the mint.
3. **Mint accepted a worksheet grant that could match nothing.** "New expression column" carries no
   `formulaName`, so the grant had no `name`; mint succeeded, then every target was refused against
   `'Query1.null'`. A false success at mint is exactly what moving validation earlier prevents.
4. **The classification guard caught a change, not an addition.** Now enforced by the compiler: a new
   `Location` fails to compile, and a new `Kind` fails an exhaustiveness test until it is classified.

**Three deliberate behaviour changes**, called out because they are user-visible: a pane session is
now refused on whole-sheet *reads* as well as writes; `save_worksheet` from a pane-scoped worksheet
session is refused; and Connect Agent is suppressed for worksheet *conditions*, whose derived context
carries no column name and could only ever mint a grant matching nothing.

Whole-sheet toolbar pairing is unchanged — pinned by a companion test at every surface that now
refuses.


larryliang-inetsoft (commented):
## Post-merge-review findings — two Important, neither blocking, both recorded rather than fixed

The fix wave's scoped re-review **closed all four** seam defects and verified the two strongest claims by running them rather than reading: adding a `Location` now fails compilation at `ScriptTarget.java:137` and `:498`, and adding a `Kind` fails `everyKindIsClassifiedByExactlyOneGuard`. It also re-derived the endpoint surface independently and confirmed no path was missed — ~55 viewsheet/binding endpoints funnel through `ViewsheetSessionService`'s four entry points, all guarded; the four raw `SheetSessionService.resolve` calls outside the guard are all `detach`, correctly exempt.

Two findings it raised, which I am recording here rather than fixing, because they are judgment calls a reviewer should weigh:

### 1. `worksheetCondition` ships with no entry point anywhere in the product

`formula-editor-dialog.component.ts:253-255` — the condition branch supplies no `name`, so `isAddressable` is false, so `canPair` is false. It is the **only** producer of `worksheetCondition` in the web tree. Meanwhile the branch ships a complete, tested server path for that kind: the `Kind` constant, the `edit_condition` route, mint-time validation, and a real-controller integration test.

Net effect: a shipped kind is unreachable, and its validation branch and mint test currently exercise a path no user can reach. Suppression is the right call versus minting a code that fails on every edit — but the honest statement is that **`worksheetCondition` pane pairing does not ship in this PR**; it is server-side only. The dialog already knows the column (`formulaName`), so the fix is likely small.

### 2. The worksheet half of the critical fix is per-endpoint, not structural

All 12 of today's session-taking endpoints on `WorksheetAgentController` are covered — each verified individually. But unlike `ViewsheetSessionService` (one chokepoint) and `SheetOpenService` (one mint site), this surface is 12 hand-written `requireWholeSheetSession()` calls. The guard cannot move into `WorksheetEditService`, because that service is *also* the pane-scoped script surface and would then refuse its own front door.

Residual risk, not a live hole: endpoint 13 gets added, the author copies the `requireEnabled()` line and not the guard, and the suite stays green — the same shape as the classification defect this wave was raised to close. The fix in that spirit is a reflective test over `@*Mapping` methods taking `{sessionToken}`, asserting each refuses a pane-scoped session, so the 12 manual calls become an enumeration nobody can forget.

### Minor

`SheetOpenService.java:143-147` still carries the comment explaining why the new worksheet session is minted with `editorContext = null`. The guard 50 lines above now makes that state unreachable, so the reasoning describes a case that can no longer occur — it reads as though the laundering path is still live.

larryliang-inetsoft:
Addressed review finding 1 (worksheetCondition has no Angular entry point) in 47c9af8a2:

- `FormulaEditorDialog.deriveEditorContext()` dropped the column name for `worksheetCondition`, so `canPair` was always false. Fixed, with the existing test that had pinned the bug as intended behavior corrected.
- Beyond that one-line bug, no caller actually threaded `runtimeId`/`socketConnection`/column name down to this dialog for worksheet conditions at all. Wired `table`/`runtimeId`/`socketConnection` through the full condition-editing chain (simple and advanced tabs, MV conditions, ONE_OF/BETWEEN sub-editors) so pairing a worksheet condition's expression is reachable from the UI, not just from the server side.

Verified: `formula-editor-dialog.component.spec.ts` (18/18), plus all existing `.tl.spec.ts`/`.spec.ts` files for every touched component (condition-editor, condition-item-pane, condition-pane, expression-editor, one-of-condition-editor, assembly-condition-dialog, etc.) still pass, and the Angular AOT build compiles clean across all new template bindings.

Finding 2 (the 12 hand-written worksheet-side guard call sites) is still open -- out of scope for this pass.

jkidacki-inetsoft:
**FIXED 2026-08-18, `a0631affd`** — review finding 2 (the worksheet guard being 12 hand calls, not a chokepoint).

Added `WorksheetAgentControllerTest#everySessionTokenEndpointRefusesAPaneScopedSessionOrIsExplicitlyExempt`, which reflectively enumerates every `@GetMapping`/`@PostMapping` method on `WorksheetAgentController` whose path contains `{sessionToken}` and asserts each one refuses a pane-scoped session before doing any real work, with `detach` as the one documented exemption. This mirrors `PaneScopeServiceTest#everyKindIsClassifiedByExactlyOneGuard` on the viewsheet side, since a true chokepoint (like `ViewsheetSessionService.requireSession`) isn't available here — `editOp` is legitimately shared with the pane-scoped `WorksheetScriptService` caller, which already runs its own narrower `PaneScopeService.check` and must not be refused again by the broad whole-sheet check.

Verified per the branch's own standard: temporarily dropped the guard call from `undo()`, confirmed the new test failed and named `undo` specifically, then restored it and confirmed green (`WorksheetAgentControllerTest`: 35/35, `PaneScopeServiceTest`: 23/23).

The six-step browser pass in the plan's Verification section is still outstanding — not attempted here.

jshobe-inetsoft (approved):
## Summary

This PR scopes AI-agent pairing sessions to a single script location (a chart's script, a calculated field's formula, a worksheet column's expression) instead of granting a whole-sheet write handle. The three-layer model — addressability (`Kind.requiresPaneSession()`), unconditional grant-scope matching (`PaneScopeService.matchesGrant`), and an optional strict posture flag — is cleanly separated and each layer is independently testable.

This is a security-sensitive change (it's an authorization/scoping mechanism), so the review focused adversarially on whether a crafted `editorContext`, `runtimeId`, or `Kind` could bypass scope enforcement, and on whether the "~25 binding endpoints" gap the PR description says it closed is actually closed.

**No structural authorization bypass was found.** Specifically verified:

- `matchesGrant` and `requiresPaneSession()` fail closed on unrecognized kinds/locations. `isExpressionLevel` is an exhaustive switch over `Location` with **no default branch** — adding a new expression-level `Location` without classifying it is a compile error, not a silent security regression.
- Every session-token endpoint in `BindingAgentController` and `ViewsheetAssemblyAgentController` funnels through `ViewsheetSessionService.requireSession()`, which unconditionally calls `PaneScopeService.requireWholeSheetSession` — except each controller's `detach`, which is correctly exempt (ending a pane session from a pane session is legitimate).
- `WorksheetAgentController` takes a duplicated-call approach instead, but it's backed by a reflective test (`everySessionTokenEndpointRefusesAPaneScopedSessionOrIsExplicitlyExempt`) that enumerates every `@GetMapping`/`@PostMapping` taking `{sessionToken}` and fails the build if a future endpoint omits the guard — a real regression guard, not just a snapshot of current behavior.
- `SheetOpenService.openBaseWorksheet` (the "worst path" cited in the PR description, where a whole-sheet session was previously laundered from a narrow grant) checks `vsSession.editorContext() != null` as its very first line, before any runtime/mutation work.
- The "foreign runtime vs. missing runtime must fail identically" invariant is upheld for the mint-time case the PR fixes (`SheetPairingService.validateEditorContext`).

## Findings

None of the following are blocking.

1. **Non-blocking — semantic/UX mismatch for `worksheetCondition` minted from the expression-*value* editor.**
   `web/projects/portal/src/app/widget/condition/expression-editor.component.ts` (~L58-64) wires `formula-editor-dialog` with `[isCondition]="true"`, whose `deriveEditorContext()` (`formula-editor-dialog.component.ts` L253-261) produces a `worksheetCondition` context. But `ExpressionEditor` is documented as editing only a condition's expression *value* (one comparison operand), while server-side `WorksheetScriptService.conditionEditRequest` (L498-526) parses the pane's text as a full condition (operator + values) via `parseConditionText` and calls `editCondition(table, field, operation, values)` — replacing the entire condition on that field, not just the sub-expression the dialog implies it's scoped to. An agent paired from this dialog either gets a confusing rejection for JS/SQL text, or silently overwrites the condition's operator, which the user didn't open the dialog to change. Not a privilege escalation — the write is still confined to that field — but worth a product decision: either suppress Connect Agent here (as is already done for other un-addressable locations) or give this path its own narrower kind restricted to a raw expression string.

2. **Nitpick — stale comment.** `web/projects/portal/src/app/composer/gui/wiz/editor-context.ts` L23 says "The server does not yet consume this field (G2 Task 4)" — false as of this PR, which is exactly the server-side consumer. Worth a follow-up doc fix.

3. **Nitpick — minor cross-node info-oracle asymmetry, pre-existing pattern, not introduced by this PR.** `SheetPairingService.validateEditorContext` (L166-230) throws an identical message for "genuinely missing runtime" and "runtime exists locally but owned by another user," as intended. But when `getSheetDirect` returns `null` because the runtime lives on a *different cluster node* (not because it's missing), the code falls through to kind-specific "not found" messages instead — a different message bucket than the local-wrong-owner case, which leaks a low-value node-locality signal (though it does not leak whether a specific assembly/field exists on someone else's sheet, since that message is emitted uniformly regardless of actual existence whenever `rs == null`). This mirrors an existing tolerance in `SheetJoinService` and is flagged only for completeness, not as a new hole.

## Test coverage

Strong. Two tests stood out as genuine regression guards rather than snapshots of current behavior:
- `WorksheetAgentControllerTest#everySessionTokenEndpointRefusesAPaneScopedSessionOrIsExplicitlyExempt` (reflective, fails on a future unguarded endpoint).
- `PaneScopeServiceTest#everyKindIsClassifiedByExactlyOneGuard` alongside the exhaustive `isExpressionLevel` switch, which together catch a new `Kind` mapped onto an existing `Location` — the one case the compiler alone can't catch.

Minor gap: no test asserts literal string-equality between the "foreign owner" and "missing runtime" mint-time messages (Finding 3's boundary), though this is low-risk since both come from a single shared exception construction rather than two independently maintained branches that could drift apart.

## Frontend `editorContext` plumbing

Consistent across the ~30 touched property dialogs, script panes, and condition editors. `runtimeId`/`socketConnection`/`editorContext` thread through the same paths as existing `scriptTreeModel`/`model` bindings. The `detach()`-on-`ngOnDestroy` fix (capturing the minted context rather than re-reading a live getter) is correctly wired into both `script-pane.component.ts` and `formula-editor-dialog.component.ts`. The one soft spot is Finding 1 above — the context is threaded through correctly, just to a `Kind` whose server-side write scope is broader than the editor pane implies.

## CI / mergeability

All checks passing (`claude-review`, `Semgrep Scan`, `Semgrep OSS`, `build`, `license/cla`), no merge conflicts (`MERGEABLE`/`CLEAN`).

## Recommendation

**Approve.** No blocking issues, all checks green, no conflicts. Findings above are non-blocking and worth addressing in a follow-up rather than gating this PR.

---
🤖 Automated review

larryliang-inetsoft:
Addressed review finding 1 (worksheetCondition minted from the expression-value editor mismatches its write scope) in `c67ea1c14`, per the "give this path its own narrower kind" option: added `Kind.WORKSHEET_CONDITION_VALUE` (`worksheetConditionValue`), addressed by (table, field) like `worksheetCondition` but deliberately not its dialog sibling — a grant for one must never satisfy the other, since they carry different authority over the same location.

`WorksheetScriptService` resolves the EXISTING condition, preserves its operator (`isEqual()`/`isNegated()` folded in exactly as `editCondition` interprets them), and writes only the value — refusing outright for `BETWEEN`/`ONE_OF`/`NOT_ONE_OF` (no single value slot) or `NULL`/`NOT_NULL` (no value at all), on both read and write.

`FormulaEditorDialog` gains `conditionOperation`, threaded from `ConditionEditor`'s existing `operation` input through `ExpressionEditor`. Only a single-value operator mints an addressable context; `BinaryConditionEditor`/`OneOfConditionEditor`'s `BETWEEN`/`ONE_OF` instances never bind an operation into the `ConditionEditor` they host, so they fall into the same "unaddressable, suppress the button" path as an explicitly multi-value operator — verified explicitly rather than assumed.

`worksheetCondition` itself is untouched — still fully servable for whole-sheet sessions and direct worksheet-chat tool calls, which genuinely mean "replace the whole condition." It simply has no live UI caller once this dialog mints the narrower kind instead, the same shape it shipped in before finding 1's original fix.

Verified per the branch's standard: reverted operator preservation and the single-value gate in turn, confirmed each mutation's own new tests fail and name themselves, restored. Java 1755/1755 (full `inetsoft.web.wiz.**`, clean build), Angular 24/24 bare + 120/120 scoped `.tl` (the whole condition-editing chain finding 1 threaded runtime/socket through).

Not touched: `stylebi-wiz`'s `connect.md` still names `worksheetCondition` in its worksheet-hosted-kinds caveat table — cosmetic only (the plugin doesn't yet route either kind through its own script tools regardless of name), and that's a different repo/PR (#1644) than this fix lives in. Worth a follow-up rename when someone next touches that doc.

Finding 2 (the 12 hand-written worksheet-side guard call sites) is already closed, per `a0631affd`.

PR #4655, opened 2026-08-18
Serve the endpoint catalogue's admin and type-listing views

Nelson-cao:
Two views the wiz portal needs before it can offer a catalogue ingest. Follow-up to #4653, which added the catalogue itself.

Ingesting a connector's endpoint catalogue is a **deployment-wide** write: the catalogue ships in the connector jar, so every organization on one StyleBI install sees the same content. That makes it a site-administrator action against a global tick list — and neither the identity check nor the list is something wiz can derive on its own.

## `GET /api/wiz/datasources/endpoint-catalog/admin`

Returns `{"siteAdmin": true|false}`.

wiz cannot decide this itself. `OrganizationManager.isSiteAdmin` expands role inheritance through `getAllRoles` and tests each role with `isSystemAdministratorRole`, whose name is provider configuration — so matching role strings out of a JWT cannot reproduce it, and any wiz-side approximation would be wrong on exactly the deployments that configure their own role names.

It **reports rather than enforces**. wiz owns the action being gated; a page that can ask beforehand explains the refusal instead of showing a bare 403. Following `AdminAiController.requireSiteAdmin`, whose javadoc already records this, the call returns `true` for the default principal when security is off, so a development install is not locked out and no `isSecurityEnabled()` branch is needed.

## `GET /api/wiz/datasources/endpoint-catalog/types`

One row per registered tabular type: `type`, `status`, `endpointCount`, `describedCount`, where status is `CATALOGUED` / `NOT_CATALOGUED` / `UNAVAILABLE`.

The portal has to present a tick list *before* anything has been ingested, and only StyleBI knows which types this build registers. It reuses the existing `/endpoint-catalog` resolution chain (`getQueryClass` → `getClass` → `EndpointCatalogReader.read`) so the two endpoints cannot disagree about what a type is.

It carries no permission gate, for the same reason `/endpoint-catalog` does not: it describes installed software, naming no instance and no path. `UNAVAILABLE` connectors are listed rather than hidden — a plugin that failed to load is an environment problem, and hiding it would present that as a fact about the connector.

## A javadoc that said the opposite of what is true

`WizDatasourceEntry.sourceType` documented its value as coarse — `e.g. "jdbc" / "Rest"`. It is not. The field is `XDataSource.getType()`, which carries the **specific** type: `StripeDataSource` passes `TYPE = "Rest.Stripe"` through `super(TYPE, …)` into `XDataSource.type`, and the controller populates the field from `dataSource.getType()`. The coarse family name comes from `getBaseType()`, a different method that `AbstractRestDataSource` overrides to return `"Rest"`, and this field never carries it.

The distinction is load-bearing: this value is the join key between a data source and its endpoint catalogue, which is keyed by the specific type. Read as documented, it says the join is impossible and an API extension is needed — the javadoc is corrected to say which method is which and why it matters.

## Testing

`WizDatabaseControllerSecurityTest` gains 4 cases (existing file — no new test file added for a previously untested path). The site-admin verdict is exercised in both directions via `mockStatic(OrganizationManager.class)`, the idiom already used in `core/src/test`. The suite's frozen request-mapping set is updated so a future endpoint added without a deliberate permission decision still fails it.

```
mvn -pl core test -Dtest=WizDatabaseControllerSecurityTest
Tests run: 23, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
@@ -20,6 +20,7 @@ package inetsoft.web.wiz.controller;
 
 import inetsoft.report.internal.Util;
 import inetsoft.sree.RepositoryEntry;
+import inetsoft.sree.security.OrganizationManager;
 import inetsoft.sree.security.ResourceAction;
 import inetsoft.sree.security.ResourceType;
 import inetsoft.sree.security.SecurityEngine;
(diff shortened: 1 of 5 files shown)


jshobe-inetsoft (approved):
## Code Review

**Scope:** Adds two new read endpoints to `WizDatabaseController` — `GET /datasources/endpoint-catalog/admin` (site-admin verdict) and `GET /datasources/endpoint-catalog/types` (per-type catalogue summary) — plus a javadoc correction on `WizDatasourceEntry.sourceType`.

### Correctness
- `getEndpointCatalogAdmin` delegates directly to `OrganizationManager.getInstance().isSiteAdmin(principal)`, matching the exact pattern used by `AdminAiController.requireSiteAdmin` cited in the PR description. Correctly handles inherited roles and provider-configured role names, and behaves sensibly (returns `true`) with security disabled.
- `getEndpointCatalogTypes`/`summarizeCatalogue` reuse the same `getQueryClass` → `getClass` → `EndpointCatalogReader.read` resolution chain as the existing `getEndpointCatalog` method, so the two endpoints can't disagree about a type's status. The three-state `CATALOGUED`/`NOT_CATALOGUED`/`UNAVAILABLE` classification mirrors the existing method's `catalogs`/`notCatalogued`/`unavailable` distinction (registered-but-unloadable vs. loaded-but-no-catalogue), correctly kept separate.
- Blank-description filtering for `describedCount` is correct and matches the documented intent (whitespace isn't a retrievable description).
- The `sourceType` javadoc fix is factually correct: `AbstractRestDataSource.getBaseType()` returns the coarse family name, `XDataSource.getType()`/`getType()` returns the specific type used as the join key — verified against the description in the diff.

### Consistency with existing code
- No `@Secured`/`@RequiredPermission` gate on either new endpoint, but this exactly matches the sibling `getEndpointCatalog` method's existing, documented convention in this same file (catalogue/type-registry contents are build metadata, not per-tenant data). Not a new pattern introduced by this PR.
- `catch(Throwable ex)` around class-loading is broad, but again matches the existing method's established idiom in this file — not a regression, just a pre-existing style choice.

### Tests
- `WizDatabaseControllerSecurityTest` gains 4 well-targeted cases covering: described-count separation, unavailable-vs-not-catalogued distinction, and both branches of the site-admin verdict via `mockStatic(OrganizationManager.class)` (existing idiom in this test class). The frozen request-mapping list is updated so a future endpoint added without a deliberate permission decision still fails the test.

### Security / regressions / backward compatibility
- No new permission surface beyond what the sibling endpoint already establishes; no path or instance-identifying data is returned.
- Purely additive — no existing method signatures changed except the javadoc-only edit on `WizDatasourceEntry`.
- `build` and `Semgrep` checks are passing; PR is `MERGEABLE` with no conflicts.

PR #4658, opened 2026-08-18
Bug #76059, #76039, follow the dimension sort order for previous/next calc navigation

ngovus-inetsoft:
Fixes Bug #76059 and Bug #76039.

## Problem

On an auto chart with `SecondOfMinute(order_datetime)` on X, Sort (Ascending), and **"As time series" off**, `Value of previous` and `Moving Average of 5` did not agree with the chart's own axis ordering:

- `SecondOfMinute = 1` lost its previous value — null instead of 28
- every moving-average window at the leading edge slid by one position (2.6667 / 2.25 / 2 instead of 2.25 / 2 / 2 for seconds 1 / 10 / 11)

## Cause

`DataSetRouter` imposed a hard-coded `nullsLast(comparingInt(...))` order on **every** part-date-group dimension (`HourOfDay`, `DayOfWeek`, `MonthOfYear`, `SecondOfMinute`, …), discarding the dimension's own display comparator.

`SecondOfMinute` is `SECOND_OF_MINUTE_DATE_GROUP = 11 | PART_DATE_GROUP`, so the configured Sort (Ascending) comparator was dropped and the null group moved from the front of the value list to the back:

```
display / physical order : [null, 1, 10, 11, ...]
router order (buggy)     : [1, 10, 11, ..., null]
```

`ValueOfColumn` then sees `1` at index 0 → no previous → `INVALID`. `MovingColumn.getCondData` builds its window from `router.getIndex()` / `router.getValues()` rather than physical rows, so the same reordering shifts every window.

**Why "As time series" off is the gate:** `VSDataSet.prepareGraph` only installs a `ScaleRouter` for a `TimeScale`, and constructs it as `new ScaleRouter(s, getComparator(f))` — the time-series path already honors display order. With time series off the axis is categorical, no `ScaleRouter` exists, and navigation falls through to `DataSetRouter`.

## Fix

The sort configured on the field defines the order previous/next navigation walks — ascending, descending, specific order, or value-based as set by a Top-N/Bottom-N "Sort By Value" ranking. The calc then agrees with the order the values are plotted in, and with the order scripts see them in through `getData()`.

Part-date-group fields keep a natural calendar order as the fallback for when no sort is configured at all, since raw row-appearance order is arbitrary for those. That fallback comparator is also total and null-safe:

- nulls sort **first**, matching the position the null group occupies on an ascending axis
- a non-numeric label sorts after all numeric values instead of throwing. A Top-N "group others" emits the `"Others"` string into an otherwise-Integer part-date column, which previously threw `ClassCastException` while building the router.

Totality/transitivity matters here because `AbstractRouter.search()` binary-searches `values` with this comparator.

## Tests

- `testPreviousOnPartDateGroupFollowsLabelSortWithNullGroup` — ascending part-date dimension with a null group (Bug #76059).
- `testPreviousOnPartDateGroupFollowsRankingSortOrder` — part-date dimension under a "Sort By Value" ranking, where display order is neither row order nor calendar order (Bug #76039).
- `testPreviousOnPartDateGroupWithOthersLabel` — the no-sort natural-order fallback, and the `"Others"` label `ClassCastException`.

`ValueOfColumnTest` 22/22 pass. `MovingColumn` / `Change` / `RunningTotal` / `Percent` / `CompoundGrowth` 42/42 pass. Both bugs confirmed fixed in the running app.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/data/DataSetRouter.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/data/DataSetRouter.java
@@ -23,6 +23,7 @@ import inetsoft.report.composition.graph.VSDataSet;
 import inetsoft.uql.XConstants;
 import inetsoft.uql.viewsheet.VSDataRef;
 import inetsoft.uql.viewsheet.XDimensionRef;
+import inetsoft.util.Tool;
 
 import java.util.*;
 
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4661, opened 2026-08-18
Bug #76060, restore the last period's full interval range in date comparison

ngovus-inetsoft:
## Problem

In `RM3_2_4`, the "All - Qtr Granularity" chart and its crosstab were missing 2022 Q3 and Q4.

Because the rows are absent from the crosstab as well as the chart, this is a data problem, not an axis-label one: the query never asked for those quarters.

## Root cause

`DateComparisonInfo.getStandardPeriodsIntervalCondition()` builds a calendar and hands it to `appendRangeToDateConditions()`, which walks **backwards** from it emitting one condition per context-level interval. That calendar therefore has to start at the *end of the last period*.

7931b9068f (Bug #75152, #3818) removed the `setDateToEndOfPeriod(calendar, periodLevel)` call, so the walk started at the interval containing the raw end day instead. With end day `2023-04-07` and "previous 2 years, not inclusive" (last period = the completed year 2022), the iteration began at 2022-04-01 and walked back, producing only six quarter conditions:

```
2022-04-01..04-06, 2022-01-01..01-06, 2021-10-01.., 2021-07-01.., 2021-04-01.., 2021-01-01..
```

2022 Q3 and Q4 were never generated.

## Fix

Restore the `setDateToEndOfPeriod()` call, then clamp the calendar to the end day. The clamp only engages when the last period is still in progress (`inclusive = true` with a mid-period end day) — which is exactly the empty/future-interval case #75152 was addressing. When the last period is complete, the clamp is a no-op and all of its intervals are emitted.

## Verification

Executed the `RM3_2_4` viewsheet in a `RuntimeViewsheetExtension` harness and dumped the resulting data. Before the fix, both the chart and the crosstab stopped at 2022 Q2; after, all four 2022 quarters are present. (`Chart28`, whose interval is `All`, takes a different code path and was unaffected either way.)

## Tests

Added `DateComparisonIntervalConditionTest` covering both directions:

- a completed last period must cover all four of its quarters — fails on the unfixed code with exactly the reported 6-vs-8 gap
- an in-progress last period must stop at the quarter containing the end day — guards the #75152 behavior

`DCMergeDateFilterTest`, `DCMergeDatePartFilterTest`, and `DcFormatTableLensTest` still pass.

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/DateComparisonInfo.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/DateComparisonInfo.java
@@ -1709,6 +1709,19 @@ public class DateComparisonInfo implements Cloneable, XMLSerializable {
          calendar.add(calendarPeriodLevel, -(quarterCalendar ? 3 : 1));
       }
 
+      // The iteration below walks backwards from this calendar, so it has to start at the
+      // end of the last period. Otherwise the intervals of the last period that fall after
+      // the end day get no condition at all (e.g. Q3/Q4 missing when the end day is in Q2).
+      setDateToEndOfPeriod(calendar, periodLevel);
+
+      // ...but never past the end day itself. When the last period is the in-progress one
+      // (inclusive), its period end is in the future and would produce empty intervals. (75152)
+      Date runtimeEndDay = standardPeriods.getRuntimeEndDay();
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4662, opened 2026-08-18
Bug #76047, fix typo hiding the AWS SDK log level property

ngovus-inetsoft:
## Summary

`PropertiesController.removeUnuseProperties` removed `log.level.intesoft.storage.aws.com.amazonaws` — `intesoft`, with the `n` and `e` transposed. No property by that name exists, so the removal was a no-op and the intended key, `log.level.inetsoft.storage.aws.com.amazonaws`, was never hidden. The adjacent line spells it correctly, which is what makes it a typo rather than a distinct key:

```java
properties.remove("log.level.intesoft.storage.aws.com.amazonaws");   // <-- intesoft
properties.remove("log.level.inetsoft.storage.aws.org.apache");      // <-- inetsoft
```

## Effect

On a non-Enterprise build the AWS SDK logger level stayed listed in Settings > Properties and in the defaults listing, where the neighbouring Apache one did not — the two halves of one intent behaved differently.

Note that this key is not a shipped default (`defaults.properties` carries only `log.level.inetsoft.performance`); it becomes a real property once the logger's level is set, via the EM Log Levels page or a deployment's `sree.properties`. The stale entry surfaced in that case rather than universally.

## Why these two keys belong together

Both name packages shaded into `inetsoft.storage.aws.*` by the enterprise AWS integration module (`integration/inetsoft-integration-aws/pom.xml`), and both are set side by side in `PropertiesEngine.initLogging()`:

```java
lm.setLevel("inetsoft.storage.aws.com.amazonaws", LogLevel.WARN);
lm.setLevel("inetsoft.storage.aws.org.apache", LogLevel.WARN);
```

`PropertiesEngine.setLogLevel()` builds the property name as `"log.level." + name`, confirming the corrected spelling is the key that can actually exist.

## Safety

A full-repository grep (community + enterprise) found the misspelled form on that one line and nowhere else, so nothing wrote or read it and nothing depended on the removal missing.

`removeUnuseProperties` only filters the `Properties` copy returned by the two read-only GET endpoints, and only when `!LicenseManager.isEnterprise()`. Enterprise builds, stored configuration, and the effective runtime log level are all unaffected.

Redmine: #76047

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4663, opened 2026-08-18
Bug #76046, hide the whole log.fluentd.* family on community builds

ngovus-inetsoft:
## Problem

`PropertiesController.removeUnuseProperties` enumerated only **7 of the 12** `log.fluentd.*` keys, so on a community build Settings > Properties (and the defaults listing) showed:

| Hidden | Still listed |
| --- | --- |
| `host`, `port`, `connecttimeout`, `securityenabled`, `tlsenabled`, `security.userauthenticationenabled`, `orgadminaccess` | `security.sharedkey`, `security.password`, `security.username`, `tls.cacertificatefile`, `logviewurl` |

The half that remained is useless without the half that was hidden — an operator saw the shared key, password and CA certificate path but not the host, port, or the flags that gate them. All 12 are documented as settable via Settings > Properties, which was true for five and false for seven.

The forwarder cannot run on a community build at all: both the appender and the reset hook are loaded reflectively from `inetsoft.enterprise.log.fluentd` (`LogbackUtil.createForwardAppender`, `resetLog`), so none of these settings can take effect. A visible, settable shared-key field for a forwarder that cannot start is a misleading affordance.

## Fix

Match the family **by prefix** instead of enumerating it, so a key added to the family later is hidden automatically rather than silently landing on the wrong side. This addresses the "derive it from one place" request in the report.

Property names are stored lower-cased (`PropertiesEngine.computePropertyNameCase`), which is why the existing lower-case spellings did match; the comparison is case-insensitive anyway so an un-normalized defaults spelling is caught too.

## Second bug found while fixing this

The filter mutated the **caller's** `Properties`, which is not a copy:

- `SreeEnv.getProperties()` returns the live `internalProperties` map (`PropertiesEngine.getUserEnhancedProperties()` → `return internalProperties`), a `DefaultProperties` whose `remove()` delegates to `mainProperties.remove()`.
- `getDefaultProperties()` returns a JVM-wide cached instance.

So a plain read-only `GET /api/admin/properties` **deleted the admin's real fluentd configuration from the running server**. `LogSettingService.getFluentdSettings()` would then report `localhost:24224` with an empty shared key instead of the configured values, and once a removed key is in `PropertiesEngine.changedProps`, `saveToStorage()` would persist the deletion.

This was pre-existing for the 7 enumerated keys, but widening to the whole family made it materially worse, so it is fixed here: both endpoints now return a filtered **copy**. Filtering iterates `stringPropertyNames()` — the same view that serializes to the client, since `DefaultProperties` delegates `keySet()`/`entrySet()`/`stringPropertyNames()` to its main layer alike — so the filter cannot miss a key that is listed.

## Tests

`PropertiesControllerTest`, 14 passing. Five new cases, each verified to **fail against the unfixed code** rather than pass vacuously:

- all 12 keys of the family are hidden
- an unknown `log.fluentd.*` key is hidden by prefix, and a similarly-named `log.fluentdish.*` key is kept
- the enterprise path keeps the full family
- the non-mutation contract, for both endpoints

## Note for reviewers

This branch deliberately **keeps** the pre-existing `intesoft` typo in `log.level.intesoft.storage.aws.com.amazonaws`, so this change stays scoped to the fluentd family. That typo is corrected separately in **#4662 (bug-76047)**. Because this PR moves those keys into a constant, it will textually conflict with #4662 — whichever merges second needs a small manual resolution.

Not addressed here: selecting Fluentd on a community build still fails silently. That is the separately-filed issue referenced in the report.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/properties/PropertiesController.java
+++ b/core/src/main/java/inetsoft/web/admin/properties/PropertiesController.java
@@ -37,6 +37,7 @@ import java.io.IOException;
 import java.security.Principal;
 import java.util.List;
 import java.util.Properties;
+import java.util.Set;
 
 @RestController
 public class PropertiesController {
@@ -138,7 +139,7 @@ public class PropertiesController {
       Properties properties = SreeEnv.getProperties();
 
       if(!LicenseManager.isEnterprise()) {
-         removeUnuseProperties(properties);
+         properties = removeUnuseProperties(properties);
       }
 
       return properties;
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4664, opened 2026-08-18
Alias pointLine into set_assembly_properties's chart vocabulary

larryliang-inetsoft:
## Summary

Closes 2b-P3's last open item (roadmap unit C2, `2026-08-17-chart-options-2b-p3-design.md`). The spec left `pointLine` unresolved pending two questions; this PR resolves both by tracing source rather than building the `set_chart_plot_options` tool the spec had reserved space for.

- **`PropertyAliases.chart()`'s "unreachable" comment was wrong.** `PropertyPath` already resolves a dotted path of any depth — `pointLine` (`PlotDescriptor.isPointLine()`/`setPointLine()`) just sits one level deeper than every other chart alias (`chartAdvancedPaneModel.chartPlotOptionsPaneModel.showPoints`, not `chartAdvancedPaneModel.*` directly), and nobody had checked that before writing the comment. The fix is the alias line itself — no new tool, no new endpoint, no TypeScript change, since `set_assembly_properties`/`list_assembly_properties`/`get_assembly_properties` already discover their vocabulary from this registry at runtime.
- **A real defect found while wiring it, fixed in the same change.** `ChartPlotOptionsPaneModel.updateChartPlotOptionsPaneModel` calls `plotDesc.setPointLine(...)` unconditionally regardless of chart type — the Composer UI hides the checkbox behind `isShowPointsVisible()` instead of the write path enforcing it (the same "menu precondition, not a server guard" shape E3 found in the selection-convert actions). Without a guard, `set_assembly_properties` would report success for `pointLine` on a bar chart with no visible effect. Added `AssemblyPropertyService.requireApplicable`, refusing the write up front with a caller-legible message naming the assembly — mirroring the precedent `set_chart_separate_status` established for the identical shape.
- **No new write-coordination work needed.** This reuses the existing `chart` binding (`ChartPropertyDialogService.setChartPropertyModel`), which already goes through `VSObjectPropertyService` — C1's shared chokepoint (#4626/#4640, merged) — so the write was already revision-checked before this change.

## Test plan

- [x] `PropertyAliasesTest` — new test pins the exact resolved path; existing reflective test (`everyDeclaredAliasResolvesOnItsDialogModel`) proves it resolves on the real model
- [x] `AssemblyPropertyServiceTest` — new tests: sets `pointLine` on a line chart end-to-end (real `VSChartInfo`/`PlotDescriptor`/model chain, not mocked), refuses on a bar chart with a caller-legible message, refusal writes nothing, an unrelated patch on a bar chart is unaffected by the guard
- [x] Verified per this branch's standard: removed the guard, confirmed both refusal tests fail and name themselves, restored
- [x] Full `inetsoft.web.wiz.**` sweep: 1760 tests / 0 failures, clean build (`./mvnw -pl core clean compile`)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyPropertyService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyPropertyService.java
@@ -237,6 +237,8 @@ public class AssemblyPropertyService {
             resolved.put(key, PropertyAliases.resolveForWrite(type, key));
          }
 
+         requireApplicable(type, model, resolved.keySet(), assemblyName);
+
          // PropertyPath.set returns the root the caller must keep using: if a future alias
          // ever targets a direct top-level field of an Immutables dialog model (no nested pane
          // to absorb the rebuild — see PropertyPath's own note), the wither produces a new
@@ -287,6 +289,36 @@ public class AssemblyPropertyService {
          throw new IllegalArgumentException(
             "Setting " + type + " properties of '" + assemblyName + "' failed: " +
             rootMessage(e), e);
+      }
+   }
+
+   /**
+    * Refuses a property whose OWN dialog model says it does not apply to the assembly's CURRENT
+    * state — checked against the model just read, so it reflects this exact write, not a stale
+    * assumption. {@code pointLine} is the only alias this applies to today:
+    * {@code ChartPlotOptionsPaneModel.updateChartPlotOptionsPaneModel} calls
+    * {@code plotDesc.setPointLine(...)} unconditionally, regardless of chart type — the Composer
+    * UI hides the checkbox behind {@code isShowPointsVisible()} instead of the write path
+    * enforcing it, the same "menu precondition, not a server guard" shape found elsewhere in this
+    * plugin family (E3's selection-container guards). Without this, {@code set_assembly_properties}
+    * would report success for {@code pointLine} on, say, a bar chart, with no visible effect.
+    */
+   private static void requireApplicable(String type, Object model, Set<String> keys,
+                                         String assemblyName)
+   {
+      if(!"chart".equals(type) || !keys.contains("pointLine")) {
+         return;
+      }
+
+      Object visible = PropertyPath.get(
+         model, "chartAdvancedPaneModel.chartPlotOptionsPaneModel.showPointsVisible");
+
+      if(!Boolean.TRUE.equals(visible)) {
+         throw new IllegalArgumentException(
+            "'pointLine' only applies to a line, point, or radar chart type; '" + assemblyName +
+            "' does not support it right now, so this call would report success with no visible " +
+            "effect. Change the chart type first (set_chart_type) if points or connecting lines " +
+            "should show.");
       }
    }
 
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyAliases.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyAliases.java
@@ -383,11 +383,25 @@ public final class PropertyAliases {
       aliases.put("glossyEffect", "chartAdvancedPaneModel.glossyEffect");
       aliases.put("sparkline", "chartAdvancedPaneModel.sparkline");
 
+      // PlotDescriptor.isPointLine()/setPointLine() — nested one level deeper than every other
+      // entry in this method (chartAdvancedPaneModel.chartPlotOptionsPaneModel.*, not
+      // chartAdvancedPaneModel.* directly). PropertyPath resolves a dotted path of any depth, so
+      // this needed only the alias itself, not new plumbing — a prior pass here concluded
+      // "ChartPropertyDialogModel never surfaces this" without checking that claim against
+      // PropertyPath's actual reach. The model's own field is named `showPoints` (the dialog's
+      // checkbox reads "Show Points" or "Show Lines" depending on chart type — see
+      // ChartPlotOptionsPaneModel.showPointsLabel); aliased as `pointLine` to match
+      // PlotDescriptor's own name and the term this capability is referred to by everywhere else
+      // (spec, roadmap). AssemblyPropertyService.set() separately refuses this alias when
+      // isShowPointsVisible() is false, since PlotDescriptor.setPointLine is applied
+      // unconditionally regardless of whether the current chart type can show it.
+      aliases.put("pointLine", "chartAdvancedPaneModel.chartPlotOptionsPaneModel.showPoints");
+
       // The line pane — trend lines, grid lines, facet grid. What a user means by "add a trend
       // line" lives here, one pane below the general/advanced panes.
       //
-      // Deliberately absent: pointLine and the word-cloud font scale are PlotDescriptor fields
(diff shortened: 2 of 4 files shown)


jshobe-inetsoft (approved):
## Code Review — Alias `pointLine` into `set_assembly_properties`'s chart vocabulary

**Verdict: Approve.** No blocking issues. The core claim of the PR holds up under tracing: `PropertyPath` genuinely resolves a dotted path of arbitrary depth (`get`/`writeInto`/`typeOf` all iterate `segments(path)` without a depth assumption), so the prior "`ChartPropertyDialogModel` never surfaces this" comment was indeed wrong, and the one-line alias is the correct and complete fix. Removing that stale comment while leaving the word-cloud font scale correctly listed as absent is a nice touch.

### Correctness — verified by tracing

- **Read path.** `ChartPropertyDialogService.getChartPropertyDialogModel` always populates `chartAdvancedPaneModel` (constructing one at `:237` if absent), and `ChartAdvancedPaneModel`'s constructor unconditionally builds `chartPlotOptionsPaneModel` (`:64`). So the three-segment path is never null for a chart, and `PropertyPath.get`'s null-intermediate short-circuit is not reachable here.
- **Field mapping is real, not nominal.** `ChartPlotOptionsPaneModel.getPlotCheckBoxsSelected` sets `showPoints = plotDesc.isPointLine()` (`:280`) and `updateChartPlotOptionsPaneModel` calls `plotDesc.setPointLine(showPoints)` (`:177`). The alias reaches the actual `PlotDescriptor` field in both directions.
- **Write path is genuinely wired.** `setChartPropertyModel` → `advancePane.updateChartAdvancedPaneModel` (`:448`) → `chartPlotOptionsPaneModel.updateChartPlotOptionsPaneModel(info, plotDesc)` (`:159`). Confirmed reachable, not just plausible.
- **The defect is real.** `plotDesc.setPointLine(showPoints)` at `:177` sits in a flat block of unconditional setters with no chart-type check, while `getPlotCheckBoxsVisible` computes `showPointsVisible = (isLine || isPoint || isSimpleRadar) && !isWordCloud && !isHeatMapish` (`:349`) purely for the UI's benefit. Fixing this at the tool boundary rather than inside `updateChartPlotOptionsPaneModel` is the right call — the latter would change Composer behavior for a defect that only manifests through the new caller.
- **Guard placement.** `requireApplicable` runs after the whole patch resolves and before the first `PropertyPath.set`, so a refusal leaves nothing partially applied — consistent with the existing all-or-nothing contract the surrounding comment documents. `writesNothingWhenPointLineIsRefused` pins exactly that.
- **`"chart".equals(type)` is safe.** `typeOf` returns `type.toLowerCase()` (`:366`), so the literal comparison can't silently miss on casing.
- **No regression for existing callers.** The full model round-trip already wrote `pointLine` back on every chart patch (read populates `showPoints` from `plotDesc.isPointLine()`, write feeds it straight back), so this is idempotent for pre-existing properties. `aBarChartPatchNotTouchingPointLineIsUnaffectedByTheGuard` correctly pins that the guard doesn't broaden.
- **No backward incompatibility.** Purely additive to a runtime-discovered registry; nothing asserts an exact chart alias count, and no `.md`/`.ts` file enumerates the vocabulary statically, so the "no TypeScript change" claim holds.
- **Security.** Nothing new. The message echoes only the caller's own `assemblyName`, matching every other refusal in the file.

### Non-blocking: the guard is bypassable through the documented raw-path escape hatch

`requireApplicable` keys off the caller's **input key** (`keys.contains("pointLine")`), but `PropertyAliases.resolve` explicitly accepts any dotted string as a raw path. So this call skips the guard entirely and silently no-ops on a bar chart — the exact defect the guard exists to prevent:

```json
{ "chartAdvancedPaneModel.chartPlotOptionsPaneModel.showPoints": true }
```

This matters more than it might appear, because `list_assembly_properties` hands that very string back to the caller: `list()` emits `property.put("path", alias.getValue())` for every alias. An agent that reads the listing and echoes the `path` instead of the `name` lands squarely in the unguarded branch.

It's also the one convention in this area the change doesn't follow. `resolveForWrite`'s own javadoc — twenty lines from the new call site — states the rule and the reason:

> The refusal checks the **resolved path**, not just the input key, so the raw-dotted-path escape hatch cannot reach the same field under a different spelling.

Not blocking: this is a robustness guard rather than a correctness or security boundary, the pre-PR behavior for that raw path was the same silent no-op (so nothing regresses), and the documented vocabulary — the intended surface — is fully covered. But the fix is about one line, matching the sibling refusal's shape by checking the resolved values as well:

```java
requireApplicable(type, model, resolved, assemblyName);
// ...
if(!"chart".equals(type) ||
   !(resolved.containsKey("pointLine") ||
     resolved.containsValue(PropertyAliases.resolve("chart", "pointLine"))))
{
   return;
}
```

Worth folding in here or filing as follow-up, along with a fifth test covering the raw spelling.

### Test quality

Solid, and better than the norm for a guard like this — the pointLine tests build a real `VSChartInfo`/`PlotDescriptor`/`ChartPlotOptionsPaneModel` chain rather than stubbing `isShowPointsVisible()`, so they actually exercise `getPlotCheckBoxsVisible`'s type logic instead of asserting the mock back. The documented remove-the-guard-and-watch-it-fail verification is the right standard. One incidental benefit: a typo in the hardcoded `showPointsVisible` literal is caught, since `PropertyPath.readOne` would throw an `IllegalArgumentException` whose message contains neither `pointLine` nor `Chart1`, failing the refusal test's assertions.

Minor, all optional:

1. **`chartServiceWith` duplicates ~30 lines of `serviceWith`,** including the 20-argument constructor now spelled out twice — so wiring a 21st property service means two edits in this file, and the compiler won't point at the second one. Consider having `serviceWith` take the chart service (or an override hook) and letting `chartServiceWith` delegate.
2. **`refusesPointLineOnABarChartWithACallerLegibleMessage` wraps `when(...)` in try/catch** for a checked exception, while its three siblings just declare `throws Exception`. Declaring it here too removes six lines and matches the neighbors.
3. **Spring context is now on the whole class.** The annotations correctly mirror `ViewsheetReadServiceTest` in the same package, and the context is genuinely required (`ChartPlotOptionsPaneModel`'s constructor reaches `SreeEnv.getProperty("webmap.service")` and `Catalog.getCatalog()`). Worth being aware, though, that this repo has a documented intermittent core-suite failure driven by `SreeHomeExtension.afterAll()` nulling the global Spring context between classes in the single shared surefire fork — each added `@SreeHome` class adds another window. Confining the four new tests to a `@Nested` class carrying the annotations would keep the ten-plus pure-mock tests in this file context-free. Judgment call; following the existing precedent is defensible.
4. **Import style:** three individual imports from `inetsoft.uql.viewsheet.graph` and three from `inetsoft.web.composer.model.vs`, where this file (and codebase) otherwise collapses same-package groups to `.*`.

### Checks

- `build`, `claude-review`, `Semgrep Scan`, `Semgrep OSS`, `license/cla` — all passing. `pr-build.yml` runs a bare `./mvnw clean install` with no `-DskipTests`, so the four new tests and the `PropertyAliasesTest` addition are genuinely green in CI, not merely asserted.
- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` — no conflicts against `feature-76074-server-fixes`.

PR #4665, opened 2026-08-18
Bug #76166, skip pdf.font.mapping entries with no colon in getModel

ngovus-inetsoft:
Fixes Bug #76166.

## Problem

`PresentationFontMappingSettingsService.getModel()` split `pdf.font.mapping` on `;` and then took `substring(0, indexOf(":"))` for each entry with no check that the colon was found. An entry without one gives `substring(0, -1)` and a `StringIndexOutOfBoundsException` out of the model build — which fails the whole Presentation settings page, not just the Font Mapping card.

`PDF3Generator.getPDFGenerator()` reads the same property and silently skips any entry that does not yield two non-empty parts, so the property could hold a value the PDF path accepts and the settings page could not load.

Not reachable from the Font Mapping card, which writes `name:value` pairs. Reachable from Settings > All Properties, from `INETSOFTENV_PDF_FONT_MAPPING` at first storage initialisation, and from a direct property-store edit.

## Fix

`getModel()` skips an entry whose `indexOf(":")` is `-1` instead of calling `substring(0, -1)`, matching what the generator already does.

The guard is limited to `index < 0`. The generator also drops `a:b:c` and `a:`; `getModel` renders those today without throwing, and tightening them would silently remove rows the Font Mapping card can currently display and round-trip. The defect in scope is the exception, and that is what is guarded.

## Out of scope

The secondary note in the issue about `resetSettings` writing `""` rather than calling `SreeEnv.resetProperty` was withdrawn on triage as a non-issue, so it is not changed here. `PresentationSettingsControllerTest` already pins the current reset behavior.

## Tests

New `PresentationFontMappingSettingsServiceTest` — 4 tests, all passing via `./mvnw -o test -pl core -Dtest=PresentationFontMappingSettingsServiceTest`:

- well-formed list parses one model per entry, split at the first colon
- a colon-less entry mixed with good ones is skipped, the good ones survive
- a lone colon-less entry yields an empty list instead of throwing
- null / empty property yields an empty list

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/presentation/PresentationFontMappingSettingsService.java
+++ b/core/src/main/java/inetsoft/web/admin/presentation/PresentationFontMappingSettingsService.java
@@ -39,6 +39,12 @@ public class PresentationFontMappingSettingsService {
 
          for(String fontMapping : fontMappings) {
             int index = fontMapping.indexOf(":");
+
+            // skip malformed entries, matching PDF3Generator.getPDFGenerator()
+            if(index < 0) {
+               continue;
+            }
+
             fontMappingModels.add(PresentationFontMappingModel.builder()
                                      .trueTypeFont(fontMapping.substring(0, index))
                                      .cidFont(fontMapping.substring(index + 1))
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4666, opened 2026-08-18
Bug #76165, tolerate a malformed schedule.time.ranges value

ngovus-inetsoft:
## Problem

`TimeRange.parse` indexed four comma-separated fields without checking how many were present and called `LocalTime.parse` without catching `DateTimeParseException`. `TimeRange.getTimeRanges` split the property with no null guard. A single malformed segment in `schedule.time.ranges` therefore threw out of every caller.

The recovery path was the problem: `SchedulerConfigurationService.getConfiguration` calls `getTimeRanges()` while building the Scheduler settings model, so the page holding the Time Ranges card an operator would use to repair the value was itself unloadable.

The blast radius went well past that page. `getTimeRanges()` has eight production call sites, and two matter more than the settings page:

- `FileAuthorizationProvider.java:408` — inside default-permission map construction. Not a page failing to render; security-provider initialization failing.
- `TaskBalancer` / `TaskBalancerCondition` / `TaskBalancerAction` — the scheduler's background rebalancing loop, which fails silently rather than surfacing to anyone.

Also `ScheduleTaskAsset` (deployment export), `ScheduleDialogController` / `ScheduleDialogService` (the **end-user** viewsheet schedule dialog, not just EM), and enterprise `ScheduleApiService`.

`defaults.properties:219` ships a well-formed value and `setTimeRanges` formats what it writes, so this is not reachable from the EM card. It is reachable from `PropertiesController.editProperty` (`/api/admin/properties/edit` does a bare `SreeEnv.setProperty` with no per-key validation), from `INETSOFTENV_SCHEDULE_TIME_RANGES` at first storage initialisation, and from any hand edit.

## Fix

- `getTimeRanges()` returns an empty list for a null or blank property, and filters out segments that failed to parse.
- `parse()` checks `values.length < 4` before indexing and wraps the two `LocalTime.parse` calls in a `catch(DateTimeParseException)`; a malformed segment returns `null` and is skipped.
- `logInvalidRange()` logs each distinct bad segment once, guarded by a `ConcurrentHashMap.newKeySet()` capped at 100 entries — `TaskBalancerCondition.check` reads the ranges on every scheduler tick, so an unthrottled warning would flood the log for as long as the value stayed broken.

The catch is deliberately narrow (`DateTimeParseException`, not `Exception`) so a future precondition failure in `setName`/`setStartTime` still surfaces rather than being silently swallowed into a dropped range.

An empty or short list is safe downstream: `TaskBalancerCondition.check` degrades to `anyMatch` → false, `getRetryTime` has an `orElse(null)` branch, and `getMatchingTimeRange` returns the input range when `min()` finds nothing.

## Tests

`TimeRangeTest` gains a `MalformedPropertyTests` nested class: null property, blank property, too-few-fields, non-ISO time, and the mixed case where `Morning,0,06:00,12:00;Broken,0,noon;Overnight,1,18:00,06:00` yields the two good ranges with `Overnight` still flagged default.

The old `getTimeRangesNullPropertyThrows` test asserted `assertThrows(NullPointerException.class, ...)` — it pinned exactly the behavior being fixed, so it is removed. The `[BUG-TR-1]` header comment in `TimeRangeTest` and the matching `[BUG-TBC-2]` block in `TaskBalancerConditionTest` are updated from "Deferred" to "Fixed".

`TimeRangeTest` (24), `TaskBalancerConditionTest` (15), `TaskBalancerTest` (31) — all pass.

## Follow-up, not addressed here

An operator with one malformed segment now gets a loadable card showing two of their three ranges. Saving from that card writes back the truncated set via `setTimeRanges`, making the loss permanent — the logged warning is the only trace. Surfacing skipped segments in `TimeRangeModel` so the card can flag them would fix that, but it means touching the service, the model, and the Angular card. Worth a separate issue.

Separately, `setTimeRanges` validates duplicate names and multiple defaults only on the EM path, so `/api/admin/properties/edit` remains an unvalidated write to this key. That is a broader question about the generic property editor rather than something to solve inside `TimeRange`.

### DIFF
--- a/core/src/main/java/inetsoft/sree/schedule/TimeRange.java
+++ b/core/src/main/java/inetsoft/sree/schedule/TimeRange.java
@@ -20,12 +20,16 @@ package inetsoft.sree.schedule;
 import inetsoft.sree.SreeEnv;
 import inetsoft.util.Tool;
 import inetsoft.util.XMLSerializable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.w3c.dom.Element;
 
 import java.io.*;
 import java.time.Duration;
 import java.time.LocalTime;
+import java.time.format.DateTimeParseException;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_TIME;
@@ -222,8 +226,14 @@ public final class TimeRange implements XMLSerializable, Serializable, Comparabl
     */
    public static List<TimeRange> getTimeRanges() {
       String property = SreeEnv.getProperty("schedule.time.ranges");
+
+      if(property == null || property.trim().isEmpty()) {
+         return new ArrayList<>();
+      }
+
       return Arrays.stream(property.split(";"))
          .map(TimeRange::parse)
+         .filter(Objects::nonNull)
          .collect(Collectors.toList());
    }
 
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):

PR #4668, opened 2026-08-19
Resolve a crosstab highlight condition against the cell's own headers

lizhepu:
A crosstab highlight naming the base column plus an aggregateFormula was stored on the correct cell and colored nothing, and the apply returned 200 with no error — so neither the caller nor the user could tell it from a rule that worked.

- A crosstab evaluates a highlight condition against the target cell's available fields, where aggregates are named by their FULL name ("DistinctCount(ORDER_ID)"). The base-column-plus-formula shape is a HAVING filter's, and buildConditionItem turns it into an AggregateRef whose getName() is the base column, so no such header exists at render time and the condition resolves to nothing
- The chart path has always rebound-then-verified for exactly this reason. The crosstab path passed its condition columns to buildHighlight for value COERCION only and never checked the field resolved at all
- Give it the same treatment: rebind what is unambiguous, and fail loud naming the field and the available headers on the rest
- rebindChartConditionFields is now rebindConditionFieldsToViewColumns — it was never chart-specific, and its second match arm ("Sales" -> "Sum(Sales)") is exactly what converts the base-column shape into the header the data carries
- Crosstab only. A plain table's condition is evaluated against the base columns already passed in, so there is nothing to rebind there

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WizVsService.java
@@ -1073,7 +1073,7 @@ public class WizVsService {
     * caller, whereas a miss here silently falls back to EVERY ref — reproducing the very bug this method
     * exists to fix. Leniency is the safer default when the penalty for a miss is a wrong render rather
     * than an error. A genuinely mis-cased request is still caught: the condition fields are resolved
-    * case-sensitively against the chart columns by rebindChartConditionFields, which throws first.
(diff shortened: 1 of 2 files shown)


lizhepu:
Fixed. aggregateHeaderOf now mirrors toView()'s three arities (secondary field and N included); 12 unit tests over the rebind and composition; stale rebindChartConditionFields reference updated.

fredaan (approved):

PR #4671, opened 2026-08-19
Refuse snapshot table data and column edits, expose the type, add a properties read

fredaan:
Four findings from the L2 lane of the consolidated Composer plugin test plan, all in the worksheet agent endpoints. Companion PR for the plugin half: inetsoft-technology/stylebi-wiz#1687 — neither is much use without the other, so land them together.

## Row and cell edits were silently discarded on a snapshot embedded table

Every `XEmbeddedTable` mutator is *copy-and-swap*, never in-place: `setObject` goes through `replaceRow`, and it, `insertRow` and `deleteRow` all build a new `XSwappableTable` and then `swapTable()` the reference. So a write only lands if the caller holds the live object. A plain `EmbeddedTableAssembly` hands back the live `XEmbeddedTable`, so the swap happens inside it and the edit sticks. `SnapshotEmbeddedTableAssembly.getEmbeddedData()` returns `new XEmbeddedTable(getTable())` on **every call** — the swap landed on a throwaway wrapper, the snapshot's own `stable` was never touched, and `edit_cell`/`insert_row`/`delete_row` returned success having changed nothing.

**Fixed by refusing, not by adding a `setEmbeddedData` write-back**, because the product does not allow this either: `snapshot-embedded-table-assembly.ts` overrides `isEditable()` to `false`, and `populateModes()` only offers `"edit"` mode when that is true — a snapshot table cannot enter cell-edit mode in the Composer at all. Refusing is what makes the agent path *agree* with the UI, and filtering already refuses a snapshot the same way (`composer.ws.filter-snapshopt`).

The new `requireEditableEmbedded()` gates all three ops from one place. Its message names the type an agent can check beforehand and the one workaround that exists — re-import via `import_csv_table`/`import_excel_table`, which build a plain editable embedded table. Two plausible-looking workarounds were checked and are **not** suggested, because neither works: `convert_to_embedded` demands a `BoundTableAssembly`, and `add_mirror` can filter but cannot edit base data.

Two things worth noting for review:

- **The existing non-embedded wording is unchanged**, and a test now pins it verbatim. The test plan cites `edit_cell only works on embedded tables: <T>` as the live evidence for a different finding's re-verification, so rewording it silently invalidates that record.
- **`delete_row` was never dangerous.** It is the same copy-and-swap path, so on a snapshot it was a harmless no-op rather than a real delete. The test lane deliberately never tried it, on the reasonable worry that it might be the one that worked; this closes that question from the source rather than by risking someone's data.

*Not fixed, and reachable only through the UI:* `InsertDataService`/`DeleteDataService` have the identical missing write-back, but the UI never offers edit mode on a snapshot so they cannot be reached there. `WSEditTableDataService` (the UI's cell edit) does write back correctly.

## The same defect in the column ops

A sweep of every write-side `getEmbeddedData()` call site put this surface at **three** ops, one more than the finding first recorded — `insert_column` lives in the controller rather than the `Editor`, and was missed on the first reading. `add_column` and `insert_column` call `data.insertCol()`, `remove_column` calls `data.deleteCol()`, and none of the three wrote back, so on a snapshot the `ColumnSelection` change persisted while the data-layer change was discarded. `add_column` left a column that existed in the model with nothing behind it. (`change_column_type` is the one that already writes back, and works on a snapshot.)

This was held out of the first pass because a flat refusal is wrong here — but the rule turned out not to be a judgment call either. The Composer permits removal on a snapshot when every selected column is an expression, and the mechanism draws the line in exactly the same place: an expression column exists only in the `ColumnSelection` and never in the data, so nothing at the data layer is even attempted and nothing can be lost. UI rule and safety boundary coincide, so the rule was discovered rather than invented:

- `add_column` / `insert_column` — refused, pointing at `add_expression_column`, which touches no data and does work there;
- `remove_column` — refused for a data column, **allowed** for an expression one, matching the Composer exactly.

The guards live in `WorksheetMutationSupport` because the three call sites span two classes.

## Nothing distinguished a snapshot from an editable embedded table

`tableType()` tested `EmbeddedTableAssembly` before the snapshot subclass, so the subclass was absorbed and both reported `type: "EMBEDDED"` — an agent could not tell, before writing, which one it had.

Now `EMBEDDED_SNAPSHOT`. Chosen over adding a boolean because `type` is already the subclass discriminator — JOIN / MIRROR / UNPIVOT / ROTATED / CONCAT each get their own token, and snapshot was the only subclass being swallowed — so this completes an existing convention and leaves one source of truth instead of two that can drift. Verified beforehand that nothing in the consuming plugin matches on `type === "EMBEDDED"`. The new branch carries a comment saying it must stay first, since reordering it is all it would take to silently restore the finding.

## Worksheet properties were write-only

The properties POST could set `alias` and `description`, but there was no GET and `/model` carries neither field, so an agent could never read, verify or report the current value.

`WorksheetReadService.readProperties` returns the four fields behind the Composer's Worksheet Property dialog, read with the dialog's own `WorksheetInfo`-then-`AssetEntry` precedence so the two cannot disagree for the same sheet. Scoped to those four deliberately: returning `WorksheetInfo` internals like `designMaxRows` would invent a vocabulary the dialog does not have.

The new endpoint is the 13th `{sessionToken}` mapping, so the reflective whole-sheet-guard test's floor moves with it — which is also what proves the endpoint is being enumerated by that guard rather than skipped.

## Verification

`./mvnw -pl community/core test -Dtest='inetsoft.web.wiz.**'` — 1899 tests, 0 failures, 0 errors, 1 pre-existing skip, BUILD SUCCESS.

**Confirmed against a running Composer on 2026-08-19**, with this branch installed and the server restarted, driven entirely through the agent tools rather than by calling the endpoints directly. The target was the wizard-imported `formats` table — the same `formats.csv` that produced the findings, not a stand-in:

- `read_worksheet_model` reports it `EMBEDDED_SNAPSHOT`; on the pre-deploy build every embedded table in the open sheet read `EMBEDDED`.
- `edit_cell`, `insert_row`, `delete_row`, `add_column`, `insert_column` and `remove_column` all refuse it, with the full message.
- `add_expression_column` **succeeds** on that same snapshot, and `remove_column` then drops the column it made while still refusing `col7`, a data column on the same table. The same guard giving opposite answers decided by the column is the only half of the column rule a refuse-everything implementation could have faked.
- `preview_worksheet_data` is byte-identical across all 7 rows before and after, so `delete_row` on a snapshot is now *measured* harmless rather than deduced from the copy-and-swap reading — the check the test lane declined to risk.
- `get_worksheet_properties` returned all four fields; a write read back verbatim; a write passing only `description` left `alias` untouched. `alias` came back `""` rather than `null`, so the `AssetEntry` fallback branch is the one that ran.
- `status` immediately after the refusals still reports `agentConnected: true`, `state: "valid"` — the earlier session-destroying reap regression has not returned now that six more ops can refuse.

Two things this pass did **not** establish, stated rather than implied:

1. **The plain-embedded happy path was not re-run live.** The paired sheet held only the snapshot, so nothing live confirms the three data ops still work on an ordinary embedded table — a guard that wrongly refused every `EmbeddedTableAssembly` would have produced identical output. `editCellWritesToAPlainEmbeddedTable` and `addColumnStillWorksOnAPlainEmbeddedTable` cover it in the suite, which is why this is a coverage gap rather than a doubt.
2. **The Worksheet Property dialog was not opened** to confirm it displays what `readProperties` returns. The precedence was copied from `WorksheetOptionPaneModel` deliberately so the two cannot disagree, but that is an argument from code, not an observation.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -52,6 +52,7 @@ import inetsoft.web.portal.controller.database.QueryManagerService;
 import inetsoft.web.wiz.pairing.*;
 import inetsoft.web.wiz.script.PaneScopeService;
 import inetsoft.web.wiz.worksheet.model.WorksheetModel;
+import inetsoft.web.wiz.worksheet.model.WorksheetPropertiesModel;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -1842,6 +1843,28 @@ public class WorksheetAgentController {
    // ---------------------------------------------------------------------------
 
    public record WorksheetPropertiesRequest(String alias, String description) {}
+
+   /**
+    * Read worksheet-level properties.
+    *
+    * <p>The read counterpart of {@link #setProperties}. Without it an agent could set a
+    * worksheet's display name or description but never read, verify or report the current value,
+    * only overwrite it blindly -- and {@code /model} carries none of these fields.
+    *
+    * @param sessionToken the token obtained at join time
+    * @param user         the authenticated agent principal
+    * @return the sheet's name, alias, description and data-source flag
+    * @throws PairingException if the session is invalid/expired or the runtime is not found
+    */
+   @GetMapping("/api/wiz/v1/agent/worksheet/{sessionToken}/properties")
+   public WorksheetPropertiesModel getProperties(@PathVariable String sessionToken, Principal user)
+      throws PairingException
+   {
+      requireEnabled();
+      requireWholeSheetSession(sessionToken, user);
+      RuntimeWorksheet rws = editService.resolve(sessionToken, user);
+      return readService.readProperties(rws);
+   }
 
    /**
     * Update worksheet-level properties (alias and description).
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -1518,6 +1518,9 @@ public class WorksheetAgentController {
             throw new PairingException("Not an embedded table: " + req.table());
          }
 
+         WorksheetMutationSupport.assertSnapshotAllowsColumnAdd(
+            assembly, req.table(), "insert_column");
+
          XEmbeddedTable data = assembly.getEmbeddedData();
          ColumnSelection columns = assembly.getColumnSelection();
          boolean insertBefore = req.insert() == null || req.insert();
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -310,8 +310,15 @@ public class WorksheetEditService {
          if(toRemove != null) {
             WorksheetMutationSupport.assertSnapshotAllowsColumnRemove(t, table, col, toRemove);
 
-            // For embedded tables, also remove the data column from XEmbeddedTable
-            if(t instanceof EmbeddedTableAssembly embedded) {
+            // For embedded tables, also remove the data column from XEmbeddedTable.
+            // Snapshots are excluded, and not just as an optimization: the guard above lets
+            // only an expression column through, an expression column has no data column, so
+            // findColumn could never match. What it WOULD do is call getEmbeddedData(), which
+            // on a snapshot goes through getTable() -> initTable() and can fault the whole
+            // swapped-out data file back in to answer a question already known to be "no".
+            if(t instanceof EmbeddedTableAssembly embedded &&
+               !(t instanceof SnapshotEmbeddedTableAssembly))
+            {
                XEmbeddedTable data = embedded.getEmbeddedData();
                int index = AssetUtil.findColumn(data, toRemove);
 
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -308,6 +308,8 @@ public class WorksheetEditService {
          DataRef toRemove = cs.getAttribute(col);
 
          if(toRemove != null) {
+            WorksheetMutationSupport.assertSnapshotAllowsColumnRemove(t, table, col, toRemove);
+
             // For embedded tables, also remove the data column from XEmbeddedTable
             if(t instanceof EmbeddedTableAssembly embedded) {
                XEmbeddedTable data = embedded.getEmbeddedData();
@@ -345,6 +347,7 @@ public class WorksheetEditService {
        */
       public void addColumn(String table, String name, String type) throws PairingException {
          TableAssembly t = requireTable(table);
+         WorksheetMutationSupport.assertSnapshotAllowsColumnAdd(t, table, "add_column");
          ColumnSelection cs = t.getColumnSelection();
 
          if(name == null || name.isBlank()) {
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -1291,30 +1291,72 @@ public class WorksheetEditService {
       // -----------------------------------------------------------------------
 
       /**
+       * Resolves a table that the cell/row edit ops are allowed to write to.
+       *
+       * <p>Two refusals, and the difference between them matters. A non-embedded table simply
+       * has no embedded data to edit. A <i>snapshot</i> embedded table does have data -- but
+       * every {@link XEmbeddedTable} mutator is copy-and-swap rather than in-place, and
+       * {@link SnapshotEmbeddedTableAssembly#getEmbeddedData()} hands out a freshly built
+       * wrapper on every call, so the swap lands on a throwaway object and the write is
+       * discarded. Before this check the three ops returned success and changed nothing.
+       *
+       * <p>Refusing is the right answer rather than adding a {@code setEmbeddedData} write-back:
+       * the Composer does not offer cell editing on a snapshot either (its
+       * {@code snapshot-embedded-table-assembly.ts} overrides {@code isEditable()} to false, so
+       * such a table never enters edit mode), and refusing is what makes the agent path agree
+       * with the UI. Filtering already refuses a snapshot the same way -- see the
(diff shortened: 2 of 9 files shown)


fredaan (commented):
## Code review

Reviewed the four findings end to end, verified each premise against the source rather than the description, then fixed three issues found and re-reviewed. **No blocking issues remain.**

### Premises verified against the source

Every load-bearing claim in the description checks out:

| Claim | Verified |
|---|---|
| `XEmbeddedTable` mutators are copy-and-swap | `setObject` → `replaceRow` → `createTable`/`copyTable`/`swapTable` (`XEmbeddedTable.java:588,620`); `insertRow`/`deleteRow`/`insertCol`/`deleteCol` all end in `swapTable` |
| `SnapshotEmbeddedTableAssembly.getEmbeddedData()` hands out a throwaway | `return new XEmbeddedTable(getTable())` (`SnapshotEmbeddedTableAssembly.java:110`) — a fresh wrapper each call, so the swap never reaches `stable` |
| `change_column_type` is the one that already works | it alone calls `embedded.setEmbeddedData(data)` (`WorksheetEditService.java:1142`), and the snapshot override writes `stable` back |
| The Composer refuses cell edit on a snapshot | `snapshot-embedded-table-assembly.ts:25` returns `isEditable() === false`; `populateModes()` (`abstract-table-assembly.ts:191`) only pushes `"edit"` when that is true |
| The expression-column carve-out matches the UI exactly | `updateCanRemoveSelectedHeaders()` (`ws-details-table-data.component.ts:745`): `// only expression column can be removed on snapshot` |
| `import_csv_table`/`import_excel_table` really are a workaround | both funnel into `createEmbeddedTable`, which builds a plain `new EmbeddedTableAssembly(...)` (`WorksheetAgentController.java:832`) |
| `convert_to_embedded` is not one | it throws `"Not a bound table"` for anything that is not a `BoundTableAssembly` (`WorksheetAgentController.java:1699`) |
| `composer.ws.filter-snapshopt` exists (product's own typo) | `srinter.properties:3651` |
| Nothing consumes `type === "EMBEDDED"` | no matches anywhere in `core/src` or `web/projects` outside the model's own javadoc and its tests |

The **write-side sweep is complete**. Every `getEmbeddedData()` call site in `inetsoft/web/wiz/` is now either guarded (`add_column`, `insert_column`, `remove_column`, `edit_cell`, `insert_row`, `delete_row`) or writes back (`change_column_type`, `convert_to_embedded`/import). The selection-only ops that stay open on a snapshot are correctly left alone — `rename_column` sets an alias, `reorder_columns`/`set_column_visibility`/`add_expression_column`/`add_date_range_column` never touch the data layer.

The **`checked >= 13` floor is right**, and worth stating since it is easy to miscount: there are **14** `{sessionToken}` mappings, not 13 — `import-excel` uses `@PostMapping(value = ...)` and does not turn up in a naive grep. 14 minus the explicitly exempt `detach` is the 13 the test enumerates.

### Findings — all three fixed in this branch

**1. `getPropertiesIsRefusedWhenTheFeatureIsOff` did not test the feature flag** *(test correctness — fixed)*

`assertThrows(Exception.class, ...)` was too loose to prove anything. Delete `requireEnabled()` from `getProperties` and the test still passes: `requireWholeSheetSession` deliberately passes an unresolvable token straight through (its own javadoc says so), the mocked `editService.resolve` then returns `null`, and `readProperties(null)` NPEs. An NPE and a 403 were indistinguishable to this assertion — so the one regression the test exists to catch was exactly the one it could not see.

Now pinned to `ResponseStatusException` + status 403, matching the sibling `joinRejectsFlagOff`.

**2. The `insert_column` guard had no test** *(coverage — fixed)*

`add_column` and `remove_column` each got one; `insert_column` did not. That is the wrong one to leave uncovered: it is the op that lives in the controller instead of the `Editor`, which is the structural reason it was missed on the first sweep in the first place — so it is the guard most likely to be dropped again.

Added `editInsertColumnRefusesSnapshotEmbeddedTable`, driving the real dispatch through `ctrl.edit(...)`. Confirmed it is a real pin, not a passing decoration: removing the guard turns it red (`Unexpected exception type thrown, expected: <PairingException> but was: <IndexOutOfBoundsException>`), and restoring it turns it green.

**3. `remove_column` still faulted in a snapshot's data file on the one path the carve-out allows** *(efficiency — fixed)*

After the new guard, the only column that can reach the `EmbeddedTableAssembly` data block on a snapshot is an expression column — which by definition has no data column, so `AssetUtil.findColumn` can never match and `deleteCol` is never called. But the block still ran `getEmbeddedData()` first, and on a snapshot that goes `getTable()` → `initTable()`, which can pull the whole swapped-out data file back in from disk to answer a question already known to be "no".

Excluded snapshots from that block. Behaviour is unchanged either way — the call was already a guaranteed no-op — so this removes wasted I/O, not a code path.

### Reviewed and deliberately not changed

- **Refusing instead of adding a `setEmbeddedData` write-back** is the right call, and for a second reason beyond the UI parity the description gives: `SnapshotEmbeddedTableAssembly` exists to "mainly deal with large data", and since every `XEmbeddedTable` mutator rebuilds the entire table, a working `edit_cell` would rebuild a multi-million-row snapshot per keystroke. Refusing avoids a memory and latency hazard, not just a UI inconsistency.
- **`WorksheetPropertiesModel` returns raw `null`** where `WorksheetOptionPaneModel`'s getters normalise to `""`. Correct for a JSON DTO, and the precedence — the part that could actually make the endpoint disagree with the dialog — is copied exactly.
- **Fully-qualified `inetsoft.web.wiz.pairing.PairingException`** in `WorksheetMutationSupport` matches the four pre-existing uses in that file.
- **Hardcoded English refusal text** matches every other agent-facing message on this surface; only the UI-shared filter refusal goes through `Catalog`.

### Verification

```
./mvnw -o -pl core test -Dtest='inetsoft.web.wiz.**'
```

BUILD SUCCESS — 1900 tests, 0 failures, 0 errors, 1 pre-existing skip. (1899 → 1900: the new `insert_column` test.) `WorksheetAgentControllerTest` 37 → 38.

CI at the time of review: `build` **pass**, `Semgrep OSS` **pass**, `Semgrep Scan` **pass**, `license/cla` **pass**. `claude-review` reported `fail`, but from an infrastructure error, not a code finding — the job died in 330ms on turn 1 with an empty `ANTHROPIC_API_KEY` (`is_error: true`, `total_cost_usd: 0`), and it is failing the same way on unrelated PRs (#4668, #4672) while passing on #4670. The push above re-triggers it.

Mergeable, no conflicts against `stylebi-wiz-main-rebase`.

### One coordination note

`type: "EMBEDDED"` → `"EMBEDDED_SNAPSHOT"` is a breaking change for anything matching on the old token. Nothing in this repo does, and the description reports the plugin verified clean — but that makes landing this together with inetsoft-technology/stylebi-wiz#1687 a requirement rather than a preference, as the description already says.

PR #4673, opened 2026-08-19
Bug #76161: org-scope SSO default roles on multi-tenant installs

ngovus-inetsoft:
## Problem

`AbstractSecurityFilter.getSSODefaultRole()` returns an empty array on a multi-tenant install — the guard is explicit. `StyleBIGoogleSSOFilter.getSSODefaultRole()` (enterprise) is the **only** override of that method in either repo, and it drops the guard, so Google sign-in is the one SSO path that still assigns default roles under multi-tenancy.

`createSSOSession` then stamped every name with `Organization.getDefaultOrganizationID()`, and mapped the literal `Administrator` to `IdentityID(role, null)` — the global system administrator role. That hardcoding was written on the assumption that multi-tenant never reaches it (see `fceb855f7`, *"this property currently only takes effect in non-multi-tenant environments"*). The Google override invalidates the assumption.

Two consequences on a multi-tenant install with Google sign-in enabled:

1. A user signing in to tenant B was granted a role owned by the **default organization**.
2. The name `Administrator` resolved to the real global sysadmin `FSRole` (`FileAuthenticationProvider:1042`, `sysAdmin=true`), i.e. a **server-wide** grant.

This is not a no-op. Nothing filters principal roles by org — `XPrincipal.getAllRoles()` only hides `"Organization Administrator"` in single-tenant mode — and `DefaultCheckPermissionStrategy:216-221` short-circuits `checkPermission` to `true` for every resource in every org once a sysadmin role resolves. Aggravating factor: `AuthenticationProvider.isSystemAdministratorRole()`'s interface default is name-only and ignores `orgID`.

The grants are session-only (`createSSOSession` never writes to an `EditableAuthenticationProvider`), but the property is re-read on every login, so the grant recurs indefinitely.

## Intent

The override is **deliberate**, so the fix keeps it and corrects the org resolution instead of adding the guard:

- `StyleBIGoogleOpenIDConfig.getRoleClaim()` returns `null`, so the org-scoped default-role block in `OpenIDFilterBaseFilter` (which only runs when a role claim is configured) never executes for Google. `styleBI.google.sso.default.roles` is the *only* way a Google sign-in user gets a starting role.
- The Google path is explicitly multi-tenant-aware elsewhere: `createLoginIdentityID()` and `correctOrgForNewMultiTenantUser()` route brand-new self-signup users into the SELF org.

Adding the guard would have silently removed that capability.

## Change

Extracted the stamping into `getSSODefaultRoleID(role, principal)`:

| Install | Ordinary name | Built-in `Administrator` / `Organization Administrator` |
|---|---|---|
| Single tenant | `IdentityID(name, defaultOrgID)` — unchanged | `IdentityID("Administrator", null)` — unchanged |
| Multi tenant | `IdentityID(name, principal.getOrgId())` | dropped, with a WARN |

Single-tenant behaviour is byte-for-byte identical, including the null-org `Administrator` case. That case is **load-bearing, not a bug**: the built-in Administrator is seeded only as `IdentityID("Administrator", null)`, so removing it would break existing `sso.default.roles=Administrator` configurations.

This scoping follows the convention already used elsewhere — `OpenIDFilterBaseFilter:337-340` (`role.orgID == null || role.orgID.equals(principalOrgId)`), `UserSignupService:241-246`, `SSOSettingsService:171-178`, and `DefaultCheckPermissionStrategy:625-640` (org admin rights never extend to global roles).

### The subtle part

The refusal deliberately tests the **global (null-org) form** of the name. `AuthenticationChain` overrides `isSystemAdministratorRole` as `stream().filter(p -> p.getRole(roleId) != null).findFirst()...` — an exact `name@org` key lookup (`FileAuthenticationProvider:229` → `roleStorage.get(id.convertToKey())`) — and both built-in admin roles are seeded **org-less**. So asking about the org-scoped form never resolves to them and the check silently answers `false`, making the guard a no-op that still looks correct.

Also refused: an org-scoped role flagged `sysAdmin`, which would bypass checks in *every* org. Deliberately **allowed**: an org-scoped *organization* administrator role — it is confined to the user's own org, which is what a self-service signup flow wants for the first user of a new org.

A principal with no org drops the name rather than guessing: a null org is the global scope, and the default org is the most privileged tenant, so neither is safe.

## Tests

New `AbstractSecurityFilterTest` (12 tests) covering the full branch matrix. Two are explicit regression guards named `..._despiteBeingSeededOrgLess` — verified to **fail** against a version whose guard only asks about the org-scoped form, printing `expected: <null> but was: <IdentityID{name='Administrator', orgID='acme'}>`.

The fixture models `AuthenticationChain`'s exact-key semantics rather than the name-only interface default, because stubbing it name-only makes the admin-refusal tests pass against an implementation that does nothing.

- `inetsoft.web.security` package: **143/143 pass**
- Compiles under `-Pcommunity,enterprise`

`core/src/test/resources/docs/permission-matrix-special.md` updated — it is the traceability index for this filter.

## Related

Enterprise-side PR (javadoc on the override recording why the tenancy guard is absent, enterprise tests, `claude/` docs) follows and references this one.

### DIFF
--- a/core/src/main/java/inetsoft/web/security/AbstractSecurityFilter.java
+++ b/core/src/main/java/inetsoft/web/security/AbstractSecurityFilter.java
@@ -166,6 +166,81 @@ public abstract class AbstractSecurityFilter
    }
 
    /**
+    * Resolves a role name from an SSO default-role property into a role identity.
+    *
+    * <p>On a single tenant install every role lives in the default organization, except the
+    * built-in Administrator, which is the global (org-less) system administrator role.
+    *
+    * <p>On a multi tenant install the role is scoped to the organization the user is signing in
+    * to; stamping it with the default organization would grant a role owned by another tenant.
+    * Names that would confer administrator rights beyond that organization are ignored -- see
+    * {@link #isAssignableSSODefaultRole}.
+    *
+    * @param role      the role name from the property.
+    * @param principal the principal being logged in.
+    *
+    * @return the role identity, or {@code null} if the name must not be granted.
+    */
+   protected IdentityID getSSODefaultRoleID(String role, SRPrincipal principal) {
+      if(!SUtil.isMultiTenant()) {
+         return "Administrator".equals(role) ? new IdentityID(role, null) :
+            new IdentityID(role, Organization.getDefaultOrganizationID());
+      }
+
+      String orgID = principal.getOrgId();
+
+      if(orgID == null) {
+         // neither guess is safe: a null org is the global scope, and the default organization is
+         // the most privileged tenant. Drop the name rather than grant either.
+         LOG.warn("Ignoring role \"{}\" in the SSO default roles: the organization being signed " +
+                     "in to is unknown.", role);
+         return null;
+      }
+
+      IdentityID roleID = new IdentityID(role, orgID);
+      SecurityProvider provider = getSecurityProvider();
+
+      if(provider != null && !isAssignableSSODefaultRole(provider, roleID)) {
+         LOG.warn("Ignoring administrator role \"{}\" in the SSO default roles: administrator " +
+                     "roles cannot be granted to organization {}.", role, orgID);
+         return null;
+      }
+
+      return roleID;
+   }
+
+   /**
+    * Determines whether a role may be granted through an SSO default-role property.
+    *
+    * @param provider the security provider.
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):

PR #4674, opened 2026-08-19
Bug #76011, apply the SMTP OAuth flags to the token requests

ngovus-inetsoft:
Follow-up to #76008. That fix corrected the round-trip of the **Settings > General > Email > OAuth Flags** field, so it now saves and reloads correctly — but the value was still inert. Tracing it showed the flags never reach the authorization server on *either* path, not just the one reported.

## The problem

**Refresh path — flags dropped entirely.** `Mailer.getAccessToken` built `new HashSet<>()` and passed it straight to `AuthorizationClient.refresh`; `mail.smtp.oauthFlags` was never read. Every other caller reads and splits the flags — `ODataDataSource`, `OneDriveDataSource`, `AbstractRestDataSource`, `AuthorizationClient.refreshTokens` — the mail path was the sole outlier.

**Authorize path — flags sent in the wrong slot.** `EmailSettingsService.getOAuthParams` did `builder.addScope(request.flags().split(" "))`, leaving `OAuthParams.flags()` unset. The client posts `flags` as its own field, so the EM **Authorize** button was sending the operator's flags to the broker as extra OAuth *scopes*.

Both fail silently — no error at any point.

## The fix

- `Mailer.getSmtpOAuthFlags()` holds the single definition of the property resolution (canonical `mail.smtp.oauthFlags` plus the legacy misspelled `mail.smtp.outhFlags` fallback); `EmailSettingsService.getOAuthFlags()` delegates to it, so the rule cannot drift between the two layers. Dependency runs web → core.
- `getAccessToken` splits the flags into the set it already passes to `refresh`.
- `getOAuthParams` calls `addFlags` instead of `addScope`.

Three judgment calls worth review:

1. **The refresh applies flags only for `SASL_XOAUTH2`.** `setModel` writes `mail.smtp.oauthFlags` only in that branch, and the EM's `authorizeGoogle()` hardcodes no flags. Without the gate, someone who once configured SASL XOAUTH2 and later switched to Google Auth would start sending stale flags on a Google refresh that previously sent none.
2. **`setModel` now clears the legacy misspelled property.** Nothing ever wrote or cleared it, so on an install with that key hand-edited into `sree.properties`, blanking the field in the EM got the old value back on reload. That was cosmetic before this change and becomes live behavior once the flags are applied. Clearing it also migrates such installs on first save.
3. **Blank values are ignored rather than split.** `"".split("\s+")` returns a one-element array containing `""`, so a whitespace-only field would otherwise post a bogus empty flag.

## Testing

New `MailerTest` (6 tests) and 3 added cases in `EmailSettingsServiceTest` (9 total), all passing. Verified they are real regression guards: reverting each source change reproduces failures in the corresponding tests. The four existing #76008 tests pass unchanged through the delegation.

```
./mvnw test -pl core -Dtest='MailerTest,EmailSettingsServiceTest'
```

## Not addressed

Two lines above the flags block, `request.scope().split(" ")` has the same latent defect — no trim, single-literal-space delimiter — so `"scope1  scope2"` yields an empty scope entry. It is pre-existing and changing it would alter what is sent on the authorize path for existing configurations, so it is left for a separate change.

### DIFF
--- a/core/src/main/java/inetsoft/sree/internal/Mailer.java
+++ b/core/src/main/java/inetsoft/sree/internal/Mailer.java
@@ -470,7 +470,18 @@ public class Mailer {
       }
    }
 
-   private static String getAccessToken(SMTPAuthType authType) {
+   /**
+    * Gets the SMTP OAuth flags. The flags are stored as a plain text property, so they must be
+    * read with getProperty() and not getPassword(). Earlier versions read the flags from the
+    * misspelled "mail.smtp.outhFlags" property, so that name is still honored if the correctly
+    * spelled property is not set.
+    */
+   public static String getSmtpOAuthFlags() {
+      String flags = SreeEnv.getProperty("mail.smtp.oauthFlags");
+      return Tool.isEmptyString(flags) ? SreeEnv.getProperty("mail.smtp.outhFlags") : flags;
+   }
+
+   static String getAccessToken(SMTPAuthType authType) {
       String expiration = SreeEnv.getProperty("mail.smtp.tokenExpiration");
       String accessToken = SreeEnv.getPassword("mail.smtp.accessToken");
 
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):

PR #4675, opened 2026-08-19
Bug #76052: remove six unread property declarations

ngovus-inetsoft:
## Summary

Removes six `defaults.properties` entries that are declared but read by nothing, plus the one Enterprise Manager UI reference that made one of them look functional.

A declared default is a promise: it has a meaningful-sounding name, a plausible value, and a home in the file operators are told is authoritative. An operator will find these, set them, and conclude either that the setting is broken or that they misunderstood the units — with no error and no log message, because nothing is listening.

| Property | Was at | Notes |
|---|---|---|
| `http.session.maxactive` | `defaults.properties:107` | sole occurrence repo-wide |
| `log.view.maximum` | `:129` | sole occurrence |
| `query.aggregate.merge` | `:174` | also removed from `PropertiesTool.booleanProperties` |
| `query.registry.file` | `:181` | duplicate declaration survives in `uql.properties` (see below) |
| `role.administrator` | `:202` | sole occurrence |
| `schedule.activity.file` | `:203` | sole occurrence |

## Verification

Searched Java, TypeScript, HTML, XML and properties files across `community/`, `enterprise/`, `server/`, `docker/`, `shell/`, `integration/` and `e2e/`, excluding `target/` and `node_modules/`. No consumer of any of the six.

Two loopholes worth closing explicitly:

- **Case.** `PropertiesEngine.computePropertyNameCase()` lower-cases every property name except four whitelisted patterns, so lookups are effectively case-insensitive and a read site spelled `http.session.maxActive` would still have resolved. No such read exists.
- **Dynamic construction.** The only assembled `log.*` names in the codebase are `log.level.<name>` and `log.<CONTEXT>.level.<name>` (`PropertiesEngine:326`, `PropertiesController:243`); neither can produce `log.view.maximum`.

## Three were misleading, not merely inert

**`role.administrator`** reads as a switch to rename the administrator role and is not one. Every site that needs the name uses the literal `"Administrator"`: `AbstractSecurityFilter.createSSOSession` (line 186, where it also decides whether the role gets a null organization ID — i.e. global system administrator), `FileAuthenticationProvider` (999, 1042), `RecycleUtils` (471) and `MVDef` (759). Four separate paths would have to agree for the property to work. Renaming the role remains unimplemented; this change removes the false promise of it, not the gap. If renaming is actually wanted, it is a separate feature with security-relevant surface.

**`query.aggregate.merge`** presented as a working switch. `properties-tool.ts:43` listed it among the keys Settings → All Properties renders as a checkbox rather than a text box. That list is presentation only — not a reader — so an operator would untick the box, see `false` saved, and observe no change. Removed from the list too, so the page no longer offers a checkbox for a property with no default.

**`query.registry.file`** carries a second, disagreeing declaration at `inetsoft/uql/uql.properties:20` (`/src/java/inetsoft/uql/query.xml`, a source-tree path absent from any built install). No Java loads `uql.properties` at all, so the two never had to agree. **Left in place deliberately** — it is inert either way, but note it is now the only surviving declaration of that name, which is worth a follow-up.

## Risk

No behavior change. `PropertiesEngine.getDefaultProperties()` (`:507`) loads this file purely as the fallback layer of a `DefaultProperties` chain, so removing keys nothing reads cannot alter a lookup result. Administrator overrides in KV storage are a separate layer and are unaffected; removal does not enter `changedProps`. `PropertiesController`'s `getDefaultProperties()` / `removeUnuseProperties()` / `isEnterpriseOnlyProperty()` iterate whatever keys exist and name none of the removed ones, so the EM page stops listing six dead rows.

No test enumerates or counts this file's keys — the only test referencing it, `OrgLifecycleScopedPropertiesIntegrationTest`, asserts on `format.date` only.

The `properties-tool.ts` edit is the one user-visible change (a checkbox disappears) and is the intended fix rather than a side effect.

## Follow-ups not in this PR

- `AllProperties.adoc` (separate repo) documents `query.registry.file` and `http.session.maxactive` as working features. `maxactive` in particular is described as invalidating the oldest session above a limit — behavior nothing implements. The enforced limits are `http.session.maxPerClient` (`AbstractSecurityFilter:378`, per non-local user+IP) and `http.session.timeout` (`PropertyAccessedExpiryPolicy:54`, per idle session); neither is a server-wide total.
- `claude/distributed-session.md` (enterprise repo) still lists `http.session.maxActive` as a supported per-node cap and points at `MapSessionRepository.java`, which no longer exists — the class is now `IgniteSessionRepository`.
- The `uql.properties:20` duplicate noted above.


jshobe-inetsoft (approved):

PR #4677, opened 2026-08-20
fix compile error

sarapeng86:
Root cause: commit d99e7b337 (admin-chat Plan 1) refactored PropertiesController's constructor to take only
  PropertyChangeSideEffects, and correctly dropped the now-unused imports for AssetRepository and LogManager — but left
  the corresponding field declarations (assetRepository, logManager, securityEngine) behind as dead code. Nothing in the
  class ever read or assigned them, so they were pure leftovers referencing types with no import.

  Fix: removed the three unused field declarations in PropertiesController.java (kept only sideEffects). Verified with
  ./mvnw -pl core clean compile, which now succeeds.


sarapeng86:
https://github.com/inetsoft-technology/stylebi/pull/4676 has fixed it. close

PR #4679, opened 2026-08-20
Create a worksheet table from one endpoint of a SaaS connector

Nelson-cao:
`/ws/table` accepted four table types and threw on anything else, so an endpoint the agent had already retrieved could not be built on. Adds a fifth, `tabular table`, bound as a `TabularTableAssembly`.

## The request

```json
{
  "tableName": "STRIPE_CHARGES",
  "tableType": "tabular table",
  "tabularSource": {
    "datasourcePath": "SaaS/Stripe Test",
    "endpoint": "Charges",
    "parameters": { "Created": "1735689600" },
    "jsonPath": "$.data[*]",
    "maxRows": 5000
  }
}
```

The caller supplies only what it can know: which endpoint, and what its parameters are. `suffix`, `paged` and the lookup relations are read off the connector when the endpoint is selected, so accepting a copy would let a stale value override the connector's own definition.

## No TabularView round trip

`TabularQueryDialogService` needs a view because a human drives it. Both things a caller supplies here are annotated bean properties, so `TabularUtil.getPropertyMap` reaches them directly, and `RestParameters` is a core type rather than a connector one — no reflection on connector classes.

That also keeps `TabularUtil.callButtonMethods` out of the path. It is the only reader of the `TabularUtil.sessionId` ThreadLocal, so unlike `WizTabularController` this path needs no connector session bound around it.

Order is not interchangeable: `getParameters()` derives the contract from the endpoint currently on the bean, so the endpoint is set first. Setting it also runs the connector's `updatePagination`, which is what establishes the pagination spec, request method and content type.

## Four silent paths, each answered rather than trusted

| Path | What it does on failure |
|---|---|
| `PropertyMeta.setValue` | logs and returns — so the endpoint is read back |
| `EndpointQueryDelegate.createParameters` | skips a token its parser rejects with a warning |
| `EndpointJsonQuery.setLookupEndpoint` | drops an invalid name |
| `TabularTableAssembly.loadColumnSelection` | reports a failed request with `LOG.warn` and returns normally |

The last one is the one that matters most: without an assertion it persists a zero-column table and answers `success=true`, an error first visible when a chart binds to it and renders nothing. A non-empty column selection is now required.

Reading the built URL suffix back covers the other three, and it is reported in the failure message so the caller can see the request that was actually made.

## Required parameters are validated before anything is dialed

Every missing one is named at once. `SuffixTemplate.build` does raise for a missing required parameter, but only the first it reaches, and every further round trip past this point is a metered request. A supplied name the endpoint does not declare is rejected for the same reason: dropping it runs a narrower query than the one asked for and reports success.

## POST endpoints are refused

Their body comes from a private field on the connector's query, moved into `requestBody` only by `populateRequestBodyTemplate` — a dialog button method this path never runs — so attempting one sends an empty body.

## Two defaults point the wrong way for a metered API

`createTables` sets `designMaxRows` to 0 for wiz analytics, which on a paginated endpoint means paging to the end of the customer's data on every render. `tabularSource.maxRows` is persisted on the query instead, so every later execution is bounded and no other table type is affected.

This type is also left out of `shouldProbe`: creating it already sends one request (`loadColumnSelection` runs the query under `HINT_PREVIEW` at 100 rows to discover the columns), and the probe would send a second to answer what the column assertion already answers.

## Two pre-existing gaps this type would have fallen through

**The datasource READ gate in `buildTable` read only `physicalSource`.** A `tabularSource` path bypassed it entirely — reaching the connector, and dialing a remote endpoint, with no check. It now reads whichever field carries a path. Both outcomes are covered in `WorksheetTableServicePermissionTest`, whose denial case asserts `verifyNoInteractions(xrepository)`: the check has to precede the data source even being resolved.

**`structureTableType` is an instanceof chain with no tabular branch.** Its javadoc promises the construction and introspection endpoints speak the same vocabulary; without the branch a tabular table falls through to `getSimpleName()` and `/ws/structure` answers `"TabularTableAssembly"`, a type name `/ws/table` does not accept — which breaks reading a worksheet back to build on it.

## `expressionColumns` / `windowColumns` are rejected

`applyExpressionColumns` and `applyWindowColumns` are reached from the physical-table and mirror-table builders only, so on a tabular table they were accepted and then never applied. A derived column also needs the endpoint's column list, which does not exist until the create call has run, so the mirror is where it belongs on the merits.

## Tests

Appended to the three existing classes that already cover the changed behaviour — the READ gate, `shouldProbe`, and `structureTableType` — rather than adding files. 60 tests green across `WorksheetTableServicePermissionTest`, `WorksheetTableServiceShouldProbeTest` and `MetadataApiServiceStructureTest`.

Design: `docs/superpowers/specs/2026-08-19-tabular-table-creation-design.md` in the stylebi-wiz repository.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/model/WorksheetTable.java
+++ b/core/src/main/java/inetsoft/web/wiz/model/WorksheetTable.java
@@ -21,6 +21,7 @@ package inetsoft.web.wiz.model;
 import com.fasterxml.jackson.annotation.*;
 
 import java.util.List;
+import java.util.Map;
 
 /**
  * One table definition within a POST /api/wiz/ws/table batch request.
@@ -63,6 +64,20 @@ public class WorksheetTable {
     */
    private String sqlExpression;
 
+   // ─── Tabular-table field ──────────────────────────────────────────────────
+
+   /**
+    * Source of a {@code tableType == "tabular table"}: one endpoint of a SaaS/REST connector,
+    * plus the values for that endpoint's parameters.
+    *
+    * <p>Separate from {@link PhysicalSource} rather than an overload of it, because the two name
+    * different things. A physical source names a table that already exists with a known column
+    * list; a tabular source names an API CALL, and its column list does not exist until the call
+    * has been made — which is why {@code columns} is not honored for this table type, and why the
+    * response's column list is the first time anyone learns what the table holds.</p>
+    */
+   private TabularSource tabularSource;
+
    // ─── Mirror / join base tables ────────────────────────────────────────────
 
    /** Names of already-created tables in this worksheet to use as bases. */
@@ -100,6 +115,9 @@ public class WorksheetTable {
 
    public PhysicalSource getPhysicalSource() { return physicalSource; }
    public void setPhysicalSource(PhysicalSource physicalSource) { this.physicalSource = physicalSource; }
+
+   public TabularSource getTabularSource() { return tabularSource; }
+   public void setTabularSource(TabularSource tabularSource) { this.tabularSource = tabularSource; }
 
    public List<ColumnInfo> getColumns() { return columns; }
    public void setColumns(List<ColumnInfo> columns) { this.columns = columns; }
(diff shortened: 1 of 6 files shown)


Nelson-cao:
All four are real. Fixing all four; #2 differently from the suggestion, for a reason below.

**1. READ-permission bypass — confirmed, and worse than a mismatched pair.**

Field-presence priority was simply the wrong discriminator: nothing anywhere keys off `physicalSource` for a tabular table, so the gate could check a path no code path would ever use. `tableType` is what decides which builder runs, so it has to be what decides which path is checked.

Fixing it as suggested, and additionally rejecting `physicalSource` on a `tabular table` outright — the combination is meaningless in every direction, so letting it through only preserves the shape that produced the bypass. Adding the mixed-fields permission test.

Worth noting the wiz-services side would not have caught this either: its validator rejects `baseTables`/`sqlExpression`/`expressionColumns`/`windowColumns` on a tabular step but not `physicalSource`, and a direct API caller bypasses that layer regardless. That is the argument for fixing it here rather than there.

**2. `maxRows` — enforcing it, but only where unbounded is actually unbounded.**

Neither of the two suggested remedies is right:

- *Required always* is wrong for the endpoints that need it least. `/v1/balance` returns one object and issues one request; demanding a row cap there makes the caller invent a number that means nothing.
- *A silent default* trades a slow answer for a wrong one. A table that quietly returns 100 of 5000 charges answers "how many charges do I have" incorrectly, with nothing in the response saying so. Unbounded-and-slow is at least honest.

The condition that actually matters is whether the endpoint paginates, and that is knowable at exactly the right moment: `setEndpoint` runs the connector's `updatePagination`, which establishes `paginationSpec`, so `isPaged()` is accurate by the time parameters are applied. `isPaged()` is public but not `@Property`, so it is read by name — the same reflection `assertKnownEndpoint` already uses for `getEndpoints()`.

So: a paged endpoint with no `maxRows` is rejected, naming the endpoint and asking for a cap; an unpaged one needs none. That fires exactly when "pages to the end of the customer's data" is a real outcome, and never otherwise.

**3. `UnsupportedDatasourceException` message — agreed.**

"Annotations are currently not supported for 'jdbc' datasources" is nonsense on a table-creation path. The exception type also buys nothing here: `createTables` catches per-table and keeps only `rootMessage(e)`, so it never reaches the `@ExceptionHandler` that gives the type meaning. Switching to `IllegalArgumentException` with a message about this path.

**4. Blank line — fixing.**

Thanks for verifying the ordering, `getValidJsonPath()`, the `structureTableType` hierarchy and the constructor rather than taking the claims from the description.

fredaan (approved):

PR #4680, opened 2026-08-20
Fix concatenation subtable reorder stranding operator pairs

NeverStopFMH:
Operators are keyed by ADJACENT subtable pair, so a reorder strands every operator whose two tables stopped being neighbours. reorderTableAssemblies called resetOperators(newNames), which only drops pairs naming a table that LEFT the assembly -- a reorder changes no names, so nothing was dropped. Both callers then wrote the operators back by position, adding the new pairs on top of the surviving old ones, leaving the map holding more pairs than there are subtables. getOperatorCount() reports the map's size while getOperator(int) indexes the subtable names, so anything iterating the two together then read past the end of the array.

WorksheetReadService.readConcatType does exactly that, so one reordered concatenation made every read of its worksheet fail with an ArrayIndexOutOfBoundsException -- the caller could no longer see ANY table, including the ones needed to repair the broken one. Two subtables was enough to trigger it, and there it was unrecoverable: the only call that clears the stale pairs is removing a subtable, which below two sources deletes the whole concatenation.

The Composer's own reorder dialog had the identical restore loop and produced the same corrupt state. It does not call readConcatType, so it failed silently instead of on the spot.

Move the carry-over into reorderTableAssemblies, where the map can be rebuilt rather than added to, and drop both callers' restore loops. clearOperators() is new because neither existing option works here: removeOperator is overridden by ConcatenatedTableAssembly to also drop any subtable left unconnected, and resetOperators(null) would discard the schema table infos.

Also bound readConcatType by the subtable count rather than getOperatorCount(), so one malformed assembly can no longer cost the caller every other table in the worksheet -- the same guard readConcatCompatible already had.

RenameColumnController carries the same category error (getOperatorCount() bounding indices into getTableNames()). Investigated and deliberately left alone: disabling that loop outright changes no observable behaviour, because a rename already reaches join operators by another path. Correcting the bound would start executing code that has effectively never run, on a path that had no test coverage at all. Added that coverage and a comment recording why the bound stays instead.

Verified live on both paths: the agent reorder, and a human reorder through the Composer dialog, each read back with the operators intact and the assembly still executable.

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/CompositeTableAssembly.java
+++ b/core/src/main/java/inetsoft/uql/asset/CompositeTableAssembly.java
@@ -222,16 +222,17 @@ public abstract class CompositeTableAssembly extends ComposedTableAssembly {
    /**
     * Reorder the subtables of this table
     *
-    * <p>Operators are keyed by ADJACENT pair, so a reorder strands every operator whose two
-    * tables stopped being neighbours -- {@code A UNION B MINUS C} reordered to {@code C, A, B}
-    * has nothing stored for its new first pair {@code (C,A)}. They are therefore carried over
-    * BY POSITION here, and the operator map is rebuilt rather than added to: it must never end
-    * up holding more pairs than {@code tnames} can index, because {@link #getOperatorCount()}
-    * reports the map's size while {@link #getOperator(int)} indexes {@code tnames}, and once
-    * those disagree a caller iterating them together reads past the end of the array.</p>
+    * <p>Operators are keyed by table PAIR, not by position, and for a join those pairs are
+    * arbitrary: a star join holds an operator between the fact table and each dimension, which no
+    * ordering makes all-adjacent. Reordering therefore leaves a join's operators alone here --
+    * they are still keyed by the same table names and still found by
+    * {@link #getOperator(String, String)}.</p>
     *
-    * <p>Callers must not re-apply the operators themselves afterwards. Writing them back adds
-    * the new pairs without removing the old ones, which is exactly the inconsistency above.</p>
+    * <p>{@link ConcatenatedTableAssembly} overrides this, because a concatenation is the one
+    * composite whose operators <em>are</em> one-per-adjacent-pair: reordering it strands the
+    * operators whose tables stopped being neighbours, so they have to be carried over by position.
+    * That rebuild must not be applied to a join, where it would drop every operator between two
+    * tables that are not adjacent.</p>
     *
     * @param tables the specified table assemblies containing the same elements as the current
     *               subtables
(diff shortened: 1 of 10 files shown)


NeverStopFMH:
Both findings were real. Fixed in `edfb178e6`.

### 1. Positional carry-over applied to join tables — confirmed and fixed

Confirmed before changing anything, with a failing test: a star join (`F` joined
directly to `D1` and `D2`, so `(F,D2)` spans `tnames` positions 0 and 2) loses the
`(F,D2)` operator entirely on reorder. Exactly the mechanism described — reading by
adjacent index finds `null` for that pair, `clearOperators()` then wipes the real
operator, and only the captured entries come back.

One thing worth adding to the diagnosis: **the reorder does not have to change
anything for this to happen.** `reorderTableAssemblies` never compares old and new
order, so opening the Reorder Subtables dialog on a star join and confirming the
*same* order was enough to drop join conditions the user never touched.

Also correct that the pre-existing code guarded against this — the positional
capture/restore in `ReorderSubtableService` was wrapped in
`instanceof ConcatenatedTableAssembly`, and moving the logic into the base class is
what lost that guard. That was my regression, not a pre-existing one.

Fixed by moving the carry-over into `ConcatenatedTableAssembly` as an override, per
the first of the two suggestions. The base class now leaves a join's operators alone,
which is the correct behaviour: they are keyed by table name and unaffected by order.
`ReorderSubtableService`'s comment was updated too — it now states both halves (a
concatenation carries operators over inside its own override; a join needs nothing
carried over, and `updateOperators` above has already handled the join-specific part).

### 2. Dead `resetOperators` call — resolved by the same change

Correct that it was a no-op superseded by `clearOperators()`. Rather than dropping
one call, scoping the rebuild to the subclass gives each one a real job: `super`'s
`resetOperators(newTNamesList)` handles the general case (it does matter when the
table *set* changed, which is why it should not simply be deleted), and the
subclass's `clearOperators()` does the concat-only rebuild. No call is redundant now.

### 3. Test coverage — both cases added

- Two-source concatenation (`ops.length == 1`). Worth its own case beyond
  completeness: two sources is where the original defect was *unrecoverable*, since
  the only call that clears a stale pair is removing a subtable, and below two
  sources that deletes the whole assembly.
- Same-order reorder, as the negative control for the whole fix — it is the one
  reorder that does not change which subtables are adjacent, and live reproduction
  confirmed it never triggered the original bug, so a failure there would mean the
  repair broke something the bug never touched.
- Plus the star-join regression test from point 1, which is the one that fails
  without this commit.

255 tests green across `inetsoft.uql.asset`, `inetsoft.web.wiz.worksheet` and
`inetsoft.web.composer.ws`.

fredaan (approved):

PR #4681, opened 2026-08-20
Adopt the composer authoring-state color system across worksheet, viewsheet, and toolbar chrome

swaker854:
## Summary
- Implements the corrected `--inet-composer-*`/`--inet-ws-*` selected, context, and dimmed token values from the composer-palette-spec design handoff, fixing color collisions in the previous draft (schema-compatible was identical to selected-bg; schema-incompatible was identical to primary-soft; the connection-line gray ramp was inverted)
- Adds `--inet-composer-selected-outline` and wires it into the viewsheet object-selection resize handles (`editable-object-container.component.scss`), replacing the reused shell-primary accent so a selected object's own fill is never overwritten
- Repoints the worksheet detail-table header row (`--inet-ws-header-primary/secondary-bg-color`) from the legacy rose accent to the selected-bg tone, so it reads as an extension of the selected assembly
- Exposes all newly-introduced composer/worksheet tokens in the EM theme editor's CSS variable list, with a new "Composer" section heading
- Aligns the composer toolbar's background/text/hover colors to match the shell's dark navbar treatment, fixing a contrast bug where one toolbar icon was invisible (white-on-light) and disabled icons were nearly unreadable

## Test plan
- [x] `npm run build` passes (frontend)
- [x] Manually verified in a running instance: worksheet graph node selected/context/dimmed colors, detail-table header color, viewsheet object selection handle color, composer toolbar colors/contrast/hover, and the new EM theme editor fields (all confirmed via live inspection + computed-style checks)
- [ ] Backend `mvn clean install` (in progress on the author's machine as of this PR)

### DIFF
--- a/core/src/main/resources/inetsoft/util/srinter.properties
+++ b/core/src/main/resources/inetsoft/util/srinter.properties
@@ -4473,6 +4473,7 @@ em.presentation.dataSource.emptyHiddenList=Hidden Data Source filter not applied
 em.presentation.dataSource.emptyVisibleList=Visible Data Source filter not applied
 em.presentation.lookAndFeel.css.clearPrompt=Are you sure you want to clear all of the CSS settings?
 em.presentation.lookAndFeel.css.heading.cards=Cards
+em.presentation.lookAndFeel.css.heading.composer=Composer
 em.presentation.lookAndFeel.css.heading.defButton=Default Button
 em.presentation.lookAndFeel.css.heading.dialogs=Dialogs
 em.presentation.lookAndFeel.css.heading.dropdowns=Dropdowns
--- a/web/projects/em/src/app/settings/presentation/presentation-themes-view/theme-css-view/theme-css-view.component.ts
+++ b/web/projects/em/src/app/settings/presentation/presentation-themes-view/theme-css-view/theme-css-view.component.ts
@@ -211,6 +211,7 @@ export class ThemeCssViewComponent implements OnInit, OnDestroy, OnChanges {
       {name: "--inet-composer-selected-bg", color: true},
       {name: "--inet-composer-selected-border", color: true},
       {name: "--inet-composer-selected-text", color: true},
+      {name: "--inet-composer-selected-outline", color: true},
       {name: "--inet-composer-dimmed-bg", color: true},
       {name: "--inet-composer-dimmed-border", color: true},
       {name: "--inet-composer-dimmed-text", color: true},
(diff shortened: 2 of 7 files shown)


jshobe-inetsoft (changes requested):
Requesting changes — the automated Claude review after the latest commit reports blocking issues:

**High: the new selection-outline rule can never apply (`!important` collision)** — `editable-object-container.component.scss:46-48` declares `.handle.bg-primary { background-color: var(--inet-composer-selected-outline); }`, but `_bootstrap-override.scss:1026` already declares `.bg-primary { background-color: var(--inet-primary-color) !important; }`. `!important` wins regardless of specificity or source order, so since the resize handles use `class="line handle bg-primary ..."` (`editable-object-container.component.html:430,440`), the new rule is dead code and the handles keep rendering `var(--inet-primary-color)` — exactly the shell-primary-overwriting-the-object-fill problem this PR sets out to fix. Add `!important` to the new rule (or otherwise beat the existing one), and re-check computed styles on the `.handle` element itself, not just the variable definition.

**Medium: `--inet-ws-header-primary-bg-color` and `--inet-ws-header-secondary-bg-color` are now identical** — `_variables.scss:389-390` points both at `var(--inet-composer-selected-bg)`; previously they were `$third-color` vs `$third-color-light`. These two are used together to distinguish states in `ws-header-cell.component.html:19-20` (group vs non-group columns), `edit-join-table.component.html:19-20` (dragging vs not), and `join-node-graph.component.html:21` (alias vs non-alias nodes) — all of which become indistinguishable. The PR description only mentions repointing the header row off the legacy rose accent, so this contrast loss looks unintended; please confirm with design or give the secondary token a distinct shade.

Two lower-severity items to clean up as well:
- `--inet-composer-selected-outline` and `--inet-composer-focus-ring` are not listed in `theme-css-view.component.ts:206-216`, so they aren't exposed in the EM theme editor despite the PR description claiming full coverage — and `--inet-composer-selected-outline` is directly consumed by a component.
- `--inet-composer-primary-accent`, `--inet-composer-primary-hover`, `--inet-composer-primary-text`, `--inet-composer-primary-soft`, and `--inet-composer-focus-ring` (`_variables.scss:361-365`) are defined but unreferenced anywhere in this PR. If they're forward-looking, note that in a comment or tracking issue.

The token definitions, navbar-hover fallback, `.composer-btn:disabled` opacity fix, header-height adjustment, and locale-key addition all reviewed as correct. See the full automated review above.

jshobe-inetsoft (approved):

PR #4682, opened 2026-08-20
feat(pairing): Follow Focus pane retargeting (Phases 1, 3 backend, and Phase 4 fix)

larryliang-inetsoft:
## Summary
- **Phase 1**: Retargets an already-held pairing session's `editorContext` in place, keyed by `(socketSessionId, runtimeId)`, instead of minting a new token per pane — `SheetSessionService.retarget`/`popFocus`/`setFollowFocus`, a per-token focus stack, gated behind a per-session `followFocusEnabled` opt-in, plus three new STOMP endpoints.
- **Phase 3 (backend half)**: `PairingJoinedNotice.focusChanged` discriminator + `SheetAgentBroadcastService.sendFocusChanged`, reusing the existing `/commands/wiz/pairing/joined` topic so the toolbar's connected indicator can show the live target as it moves.
- **Phase 4 fix**: new `GET /api/wiz/v1/agent/script/{sessionToken}/session` endpoint reporting a session's own live scope with no pane-scope check of its own — lets a headless agent client (which has no push channel, unlike the browser's STOMP broadcast) learn its scope has drifted after a Follow Focus retarget. Consumed by the companion stylebi-wiz plugin fix.

PaneScopeService/ScriptTarget need no change for Phase 1: both re-resolve the `JoinSession` per call rather than caching, so a retarget is visible immediately — proven by a dedicated exit-criterion test rather than assumed.

## Test plan
- [x] `SheetSessionServiceTest`/`SheetPairingControllerTest`/`SheetAgentBroadcastServiceTest`/`ViewsheetAgentControllerTest`: all new tests green
- [x] Full `inetsoft.web.wiz.**` suite: 1,944 tests, 0 failures, 0 errors (rebased onto current `stylebi-wiz-main-rebase`)
- [ ] Live browser pass (Phase 5) — not run, needs the LIVE STACK token; tracked in `docs/superpowers/plans/2026-08-19-follow-focus-pane-targeting-implementation.md` in stylebi-wiz

## Related
- Companion PR in stylebi-wiz: https://github.com/inetsoft-technology/stylebi-wiz/pull/1709 (plugin-side consumer of the new `/session` endpoint)
- Angular Phase 2/3 frontend work: separate branch `feat/follow-focus-pane-targeting-angular`, not yet pushed

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/pairing/JoinSession.java
+++ b/core/src/main/java/inetsoft/web/wiz/pairing/JoinSession.java
@@ -21,14 +21,40 @@ package inetsoft.web.wiz.pairing;
  * A reusable session opened after a successful pairing join.
  * Edits reuse this; the code stays single-use.
  *
- * @param editorContext the script/formula location this session is scoped to, carried over
- *                      from the {@link PairingGrant} that opened it, or {@code null} for a
- *                      whole-sheet ("Connect to Claude" toolbar) session
+ * @param editorContext the script/formula location this session is scoped to -- carried over
+ *                      from the {@link PairingGrant} that opened it at mint time, or moved by
+ *                      {@code SheetSessionService.retarget}/{@code popFocus} (Follow Focus)
+ *                      thereafter -- or {@code null} for a whole-sheet ("Connect to Claude"
+ *                      toolbar) session
+ * @param followFocusEnabled whether this session has opted in to Follow Focus -- see
+ *                      {@code SheetSessionService.setFollowFocus}. {@code false} until a session
+ *                      explicitly turns it on; {@code SheetSessionService.retarget} refuses to
+ *                      move a session's target while this is {@code false}, independent of
+ *                      whatever the Angular UI shows. Never {@code true} at construction --
+ *                      opting in is always a separate, later act.
  */
 public record JoinSession(String sessionToken, String runtimeId, String ownerIdentity,
                           SheetType sheetType, long lastAccess, long ttlMillis,
                           ConnectionMode connectionMode, String socketSessionId,
-                          String socketUserName, EditorContext editorContext) {
+                          String socketUserName, EditorContext editorContext,
+                          boolean followFocusEnabled) {
+
+   /**
+    * Back-compat constructor predating {@code followFocusEnabled} (Follow Focus) -- defaults it
+    * to {@code false}, the same default {@code SheetSessionService.open} relies on. Kept so the
+    * wide set of hand-built {@code new JoinSession(...)} fixtures across the
+    * worksheet/viewsheet/script/binding test packages -- none of which exercise Follow Focus --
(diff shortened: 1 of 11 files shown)


jshobe-inetsoft (changes requested):
Requesting changes — the automated Claude review after the latest commit reports two blocking issues:

1. **Resource leak** — `SheetSessionService.focusStacks` (keyed by `sessionToken`) is never cleaned up by `close`, `socketClosed`, `detach`, or `evictExpired`. Since tokens are never reused, every session that calls `retarget` at least once leaves a permanent orphaned entry for the life of the server process — an unbounded leak under sustained Follow Focus usage.

2. **Spurious broadcast on no-op pop** — `popFocus` returns the unchanged session (not `null`) when its stack is already exhausted, so `SheetPairingController.popFocusViaSocket` unconditionally broadcasts a `focusChanged` notice even when nothing moved. The existing tests only cover the "no matching session at all" no-broadcast case, so please add coverage for the exhausted-stack case as well.

One thing to confirm rather than change (non-blocking): the three new STOMP endpoints (`follow-focus`, `retarget`, `pop-focus`) don't check `feature.isEnabled()` the way `mintViaSocket` and the REST mint do. That matches the pre-existing `detachViaSocket`, so it isn't a new regression, but please confirm it's intentional for the administratively-disabled story.

The design itself reviewed well — retargeting a live `JoinSession` in place, the explicit `followFocusEnabled` opt-in, and reusing `validateEditorContext` for retarget are all the right calls. See the full automated review and its inline comments above.

larryliang-inetsoft:
Both blocking findings fixed in ce8c77d48:

1. **`focusStacks` leak** — all four removal paths (`close`, `socketClosed`, `detach`, `evictExpired`) now route through a shared `removeSessionsIf` helper that clears the matching `focusStacks` entry alongside the session itself.
2. **Spurious broadcast on no-op pop** — `popFocus` now returns `null` when the stack is exhausted, same as "no session matched," so `popFocusViaSocket` no longer treats a no-op as a change worth broadcasting.

Added 6 regression tests: the exhausted-stack-returns-null case (both never-pushed and drained-after-a-real-push), the `focusStacks` leak for `close`/`socketClosed`/`evictExpired` specifically, and the controller-level "real session found but stack empty" case the review called out as uncovered.

Full `inetsoft.web.wiz.**` suite re-run: 1,950 tests (1,944 + 6 new), 0 failures, 0 errors.

On the non-blocking question — the three new STOMP endpoints not checking `feature.isEnabled()`: confirmed intentional, matching `detachViaSocket`'s existing precedent. All three (like detach) only act on an *already-live* session; there's no new pairing being minted, so there's nothing for the feature flag to gate differently than it already does for detach. Happy to add the check to all four together as a follow-up if the "administratively disabled" story wants in-flight sessions cut off immediately rather than left to expire — but that'd be a behavior change to `detachViaSocket` too, out of scope for this PR.

PR #4683, opened 2026-08-20
feat(composer): Follow Focus push/pop wiring + toggle/indicator (Phases 2-3 frontend)

larryliang-inetsoft:
## Summary
- **Phase 2**: New `FollowFocusService` holds the per-runtime opt-in (default off, not persisted) and talks to the Java backend's STOMP endpoints (`/wiz/pairing/follow-focus`, `/wiz/pairing/retarget`, `/wiz/pairing/pop-focus`) to retarget an already-live pairing session. Wired into `ScriptPane`/`FormulaEditorDialog`'s `ngOnInit`/`ngOnDestroy`, alongside (not replacing) their existing `detach()` call.
- **Phase 3 (frontend half)**: `ConnectToClaudeComponent` gained a Follow Focus toggle and a live "current target" line (`Whole sheet` / `Chart1 → onClick`), gated to the whole-sheet (toolbar) instance only — a pane's own instance has nothing to toggle, since Follow Focus always retargets the OUTER session, never the pane's own. `onJoined` branches on a new `focusChanged` flag on the joined-notice payload to distinguish "someone else's pane got paired" from "my own session's live target changed."

## Test plan
- [x] `follow-focus.service.spec.ts`: 16/16
- [x] `connect-to-claude.component.spec.ts`: 40/40 (12 new)
- [x] Both components' full `.tl.spec.ts` suites (hand-scoped `--include`, per the repo's documented `portal:test-tl` gap): 100/100
- [ ] Live browser pass (Phase 5) — not run, needs the LIVE STACK token; tracked in `docs/superpowers/plans/2026-08-19-follow-focus-pane-targeting-implementation.md` in stylebi-wiz

## Related
- Backend PR: https://github.com/inetsoft-technology/stylebi/pull/4682 (Phase 1, Phase 3 backend, Phase 4 fix)
- Companion plugin-side PR in stylebi-wiz: https://github.com/inetsoft-technology/stylebi-wiz/pull/1709

### DIFF
--- a/web/projects/portal/src/app/composer/gui/wiz/connect-to-claude.component.html
+++ b/web/projects/portal/src/app/composer/gui/wiz/connect-to-claude.component.html
@@ -7,6 +7,13 @@
 
   <div *ngIf="connected" role="status" class="wiz-connect-connected alert alert-success mt-2">
     ✓ Agent connected
+    <div class="wiz-connect-target">Target: {{ currentTargetLabel }}</div>
+  </div>
+
+  <div *ngIf="followFocusToggleVisible" class="wiz-follow-focus form-check mt-2">
+    <input type="checkbox" class="form-check-input" id="wiz-follow-focus-toggle"
+      [ngModel]="followFocusEnabled" (change)="toggleFollowFocus($event)">
+    <label class="form-check-label" for="wiz-follow-focus-toggle">Follow Focus</label>
   </div>
 
   <div *ngIf="code && !connected" class="wiz-connect-code mt-2">
--- a/web/projects/portal/src/app/composer/gui/wiz/connect-to-claude.component.spec.ts
+++ b/web/projects/portal/src/app/composer/gui/wiz/connect-to-claude.component.spec.ts
@@ -575,4 +575,158 @@ describe("ConnectToClaudeComponent", () => {
       });
    });
 
+   describe("Follow Focus toggle + live target indicator (Phase 3)", () => {
+      function notice(body: any): any {
+         return { frame: { body: JSON.stringify(body) } };
+      }
+
+      function handlerFor(dest: string): (msg: any) => void {
+         const call = mockStompConnection.subscribe.mock.calls
+            .filter((c: any[]) => c[0] === dest).pop();
+         expect(call).toBeTruthy();
+         return call[1];
+      }
+
+      /** Real toolbar mint + join, so hasMinted/mintedEditorContext land in the "whole-sheet
+       *  owner" state the toggle and the live-target tracking both require. */
+      function connectAsToolbar(): void {
+         let mintHandler: ((msg: any) => void) | null = null;
+         mockStompConnection.subscribe.mockImplementation(
+            (dest: string, h: (msg: any) => void) => {
+               if(dest === "/user/commands/wiz/pairing/mint") {
+                  mintHandler = h;
+               }
+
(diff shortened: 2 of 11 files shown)


jshobe-inetsoft (changes requested):
Requesting changes — the automated Claude review after the latest commit reports a blocking correctness bug:

**`pushFocus`/`popFocus` context mismatch (both `ScriptPane` and `FormulaEditorDialog`)** — `pushFocus` is called in `ngOnInit` with the live `editorContext` at that moment, but `popFocus` in `ngOnDestroy` re-reads the live `editorContext` instead of reusing the value that was actually pushed.

- In `ScriptPane`, `editorContext` comes from `ViewsheetScriptPane.editorContext`, a getter over the `initScriptVisible` flag that the onInit/onRefresh radio buttons flip on the same persistent `<script-pane>` instance (no `*ngIf` recreation). Open the pane on one tab, switch tabs, then close it — one context is pushed and a different one is popped.
- In `FormulaEditorDialog`, `editorContext` derives from `formulaName`, which is editable while the dialog is open, so renaming a calc field mid-edit hits the same mismatch.

Both trip the stack-integrity check in `FollowFocusService.popFocus` (`follow-focus.service.ts:150-159`), logging a `console.error` and emitting a "Follow Focus close mismatch" on `errors` for entirely normal user actions. The fix is the `mintedEditorContext` pattern already used in `connect-to-claude.component.ts`: capture the pushed `editorContext` in `ngOnInit` and reuse that captured value in `ngOnDestroy`. None of the new tests vary `editorContext` between init and destroy, so please add coverage for that too.

Also worth addressing while you're in here (non-blocking): `FollowFocusService.state` never evicts entries from its `runtimeId`-keyed `Map` in a root-provided singleton, and nothing in this PR subscribes to `FollowFocusService.errors`, so a mismatch only reaches the console rather than any visible UI.

See the full automated review and its 4 inline comments above.

larryliang-inetsoft:
Blocking bug fixed in a59055675: `pushFocus`/`popFocus` now use a captured `pushedEditorContext` field (set alongside `followFocusPushed` in `ngOnInit`) instead of re-reading the live `editorContext` getter/input in `ngOnDestroy` — the same `mintedEditorContext` pattern `ConnectToClaudeComponent` already uses. Fixed identically in both `ScriptPane` and `FormulaEditorDialog`.

Added the missing test coverage in both `.tl.spec.ts` suites: each mutates the live `editorContext` (`ScriptPane`, simulating an onInit→onRefresh tab switch) / `formulaName` (`FormulaEditorDialog`, simulating a mid-edit rename) between `ngOnInit` and `ngOnDestroy`, then asserts `popFocus` still receives the *originally pushed* value. Both spec files pass: 52/52 tests (23 + 29).

On the two non-blocking notes:
- **`FollowFocusService.state` map never evicting** — leaving this as noted rather than fixing speculatively; it's a slow, bounded-by-distinct-runtimeId leak in a root singleton, not the acute per-open-and-close leak the blocking bug was. Worth a follow-up if this service is confirmed to outlive individual sheet sessions in practice.
- **Nothing subscribes to `errors`** — also leaving as-is; the doc comment's "so a host UI... can show it" describes intent for a later phase, not something this PR claims to have wired up. Will call this out explicitly in the doc comment in a follow-up so it doesn't read as already-done.

Neither blocks the fix for the actual correctness bug, which is what mattered here.

PR #4685, opened 2026-08-20
Add chart plot-size read and guard the chart-region targets

fredaan:
The agent-facing chart tools could report success for regions a chart does not have, and had no way to observe a plot resize at all. Five defects, one new read and one shared resolver. Every guard here was probed against a live Composer session, not just unit-tested.

## `get_chart_plot_size` — a plot resize had no observable

The ratio scales the plot's **minimum** size, which the browser shows as scrollbars, while the agent's render is fitted to the assembly box — so an enlarged plot and a default one render identically. The ratios are also absent from both the property dialog model and the scriptable API, so a resize could not be confirmed through any channel.

`readPlotSize` returns the stored intent (`ratio`, `resized`, `percent`, `initialRatio`, `effectiveRatio`) plus the last layout's geometry, and goes through `sessions.read` rather than `mutate` so *looking* at a plot size does not spend an undo checkpoint.

It earned its keep on the first live call: writing `ratio: 3` to a chart whose plot baseline is 5.4 reads back `resized: true, percent: 0.56` — accepted and inert, because `VGraphPair` only re-applies a ratio while the percent is >= 1. That is exactly the case that was previously invisible.

**What it does not do, stated plainly:** it does not prove a plot grew. The geometry comes from a cached pair that a resize does not relayout, and `ratio`/`initialRatio` are recomputed from whatever box was laid out last — a written `8` read back as `26.86` after one full-size render, while `percent` stayed `1.48`. The tool description and javadoc say this rather than the stronger claim the design started with.

## `ChartRegionResolver` — which axes and how many legends a chart actually has

`ChartArea` builds all four axis areas unconditionally, so a `y2` the chart does not have produced a full, plausible property list, accepted a write, and confirmed it on read-back. An agent could enumerate, read, write and verify and be told yes at all four steps.

Two rules worth stating, both learned the hard way:

- **The binding is the floor; the graph may only add.** A hidden axis leaves the coordinate entirely, so a graph-only answer would report no x axis for a chart whose x axis was hidden — then refuse to unhide it.
- **An axis object at the top or right is not evidence.** `RectCoord.createAxis` builds `yaxis2` unconditionally and keeps it whenever it carries grid lines, so it survives as a grid-line carrier with no secondary scale behind it. Only one with its own scale (`getPrimaryAxis() == null`) counts. The first version of this resolver believed the laid-out graph, shipped the same phantom it was meant to stop, and was caught by the live probe rather than by any test — the regression test for it is in `ChartRegionResolverTest`.

Used in three places: `list_chart_elements` filters its title targets and reports `axes`/`axesBasis`/`axesMeasured` for a named chart, and both region-property tools refuse an axis, or an axis title, the chart does not have. The read is guarded as well as the write, because the read is what talked the caller into the write.

## Three more, each found while verifying the one before it

- **A `y2` write landed on the primary axis.** `getAxisArea` accepts the short axis forms and the long ones, but `getChartRef` — which every `getAxisDescriptor` overload funnels through — knows only the long ones plus `"y"`/`"x"`. So `"y2"` left the ref null and the chain fell through to the descriptor shared with the primary axis, never reaching the `isSecondaryY() -> getAxisDescriptor2()` branch that exists for this case. Found by image A/B on a dual-axis chart: asking to hide the y2 labels hid the y1 labels, and passing `field` did not help. The secondary targets are now translated to the long area form on both the read and the write.
- **An out-of-range legend index returned a raw HTTP 500.** Now bounded by `vgraph.getLegendGroup().getLegendCount()`, where `ChartArea` itself counts, and refused only when that count was actually measured — an unmeasured count is zero for want of a graph, not because the chart has no legends.
- **`region: "title", target: "chart"` returned a raw HTTP 500.** `TitlesDescriptor` holds only x/x2/y/y2, so the chart's own title has no descriptor here; it is an assembly property, and the refusal now says so.

## Verified live

Against a running server, on two sheets and four charts:

| Probe | Result |
|---|---|
| `axis y2` / `title y2` / `right_y_axis` on a single-measure chart | refused, naming the axes it does have and the basis they came from |
| `axis y2` on a dual-axis chart | full property list, addressable normally |
| `set … axis y2 {showAxisLabel: false}` on a dual-axis chart | the **right-hand** ticks disappear; `axis y` reads back unchanged; restored to baseline |
| `set … axis y2` on a single-measure chart | still refused — the descriptor fix did not widen the guard |
| `legend/7` on a one-legend chart | "This chart has 1 legend, so the only valid index is 0; '7' is out of range…" |
| `title/chart` | a sentence naming `set_assembly_properties`, not a 500 |
| `get_chart_plot_size` before/after a resize | the stored-but-inert case, visible |

A wording defect in the legend refusal was also caught by reading its live output: `"the valid indexes are 0 - '7' is out of range"` reads as a range. Rewritten, with a test asserting the message never contains `0 - `.

## Tests

82 green across `ChartRegionResolverTest` (16, new), `ChartElementServiceTest` (29), `ChartRegionPropertyServiceTest` (19) and `ViewsheetAssemblyAgentControllerTest` (18).

## Deliberately not in this change

- Test case 4.3 still cannot be passed — nothing forces the relayout that would prove a plot actually grew. Needs a plot-region image (`VGraphPair.getPlotImage`) or a clear-and-rebuild read.
- `resizePlot`'s reset clears the ratios and flags but not `unitWidthRatioPercent`/`unitHeightRatioPercent` (`VSChartPlotResizeService:57-62`), so a reset chart still reports the old ratio and re-derives it on the next layout. One line, surfaced by the new read, recorded in the companion doc.

Companion PR (plugin + test plan): inetsoft-technology/stylebi-wiz#1719

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/ChartElementService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/ChartElementService.java
@@ -21,11 +21,19 @@ import inetsoft.uql.viewsheet.ChartVSAssembly;
 import inetsoft.uql.viewsheet.VSAssembly;
 import inetsoft.uql.viewsheet.Viewsheet;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import inetsoft.graph.VGraph;
+import inetsoft.graph.coord.Coordinate;
+import inetsoft.graph.coord.RelationCoord;
+import inetsoft.graph.internal.GTool;
+import inetsoft.report.composition.execution.ViewsheetSandbox;
+import inetsoft.report.composition.graph.*;
+import inetsoft.uql.viewsheet.graph.VSChartInfo;
 import inetsoft.web.viewsheet.controller.chart.*;
 import inetsoft.web.viewsheet.event.chart.*;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+import java.awt.geom.Rectangle2D;
 import java.security.Principal;
 import java.util.*;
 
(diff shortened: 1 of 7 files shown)


jshobe-inetsoft (approved):

PR #4687, opened 2026-08-20
Fix Bug #76203: trim the Fluentd shared key before encrypting it

ngovus-inetsoft:
## Problem

`LogSettingService.toPassword` computed `sanitizeProperty(value)` but used it only as a null test, then encrypted the original, untrimmed value:

```java
private String toPassword(String value) {
   String sanitized = sanitizeProperty(value);

   if(sanitized != null) {
      return Tool.encryptPassword(value);   // <-- value, not sanitized
   }

   return null;
}
```

Every other field saved from Settings > Logging goes through `sanitizeProperty` directly, which trims and stores blank as `null` — compare `setFluentdSettings` line 267 (shared key) with lines 262/271/272/275/276 (host, username, password, CA certificate, log-view URL). So a shared key pasted with a trailing space or newline — the normal outcome of copying a secret out of a terminal or a password manager — was stored encrypted with that whitespace, while everything beside it was trimmed.

The difference was invisible in Enterprise Manager: `getFluentdSettings` decrypts and redisplays whatever was stored, so the operator saw exactly what they pasted. Only the collector noticed, and it failed authentication with no indication of why. The one field where a stray character is both easiest to introduce and hardest to see was the one field that kept it.

The defect has been present since the initial open-source commit (`0d2cdd433`), so it was never intentional behavior.

## Fix

Encrypt `sanitized` rather than `value`. This cannot lose a key that was already correct, since a key with no surrounding whitespace is unchanged by `String.trim()`. The `sanitized != null` guard is untouched, so blank input still clears the property.

`toPassword` is private with exactly one call site (the shared key), so nothing else is affected.

## Related enterprise-side defect

While tracing the key to the collector I found a second defect, fixed in a companion enterprise PR: `ForwardService` read the key with `SreeEnv.getProperty` and passed it straight to `ForwardClient`, which digests the bytes verbatim. `SreeEnv.getProperty` does not decrypt (the decrypting accessor is the separate `SreeEnv.getPassword`), so the collector was receiving the base64 ciphertext as the shared key — meaning shared-key auth was failing for **any** EM-configured key, whitespace or not. This PR alone does not make shared-key authentication work; the two together do.

That also settles the question of whether any deployment could depend on a stored key whose whitespace is significant: no EM-configured key can currently be authenticating at all.

## Out of scope (filed separately)

`log.fluentd.security.password` is stored **plaintext** by `setFluentdSettings` yet decrypted on read. It works by accident, since decryption is a no-op on cleartext, but the fluentd user password sits unencrypted in the properties file while the shared key beside it is encrypted. Changing an existing property's at-rest format is not a one-line fix.

## Tests

New `LogSettingServiceTest` (5 tests, all passing), following the static-mock pattern already used by `EmailSettingsServiceTest` for an admin settings service backed by `SreeEnv`:

- a key with surrounding spaces is trimmed before encryption
- a key with a trailing newline (the copy-out-of-a-terminal case) is trimmed
- a key with no surrounding whitespace is unchanged — proves the fix cannot damage an already-correct key
- a blank key still clears the property rather than storing an encrypted blank
- the shared key is the only Fluentd field that is encrypted; its neighbours are trimmed and stored plain

`./mvnw test -pl core -Dtest=LogSettingServiceTest` → `Tests run: 5, Failures: 0, Errors: 0`

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/logviewer/LogSettingService.java
+++ b/core/src/main/java/inetsoft/web/admin/logviewer/LogSettingService.java
@@ -294,7 +294,7 @@ public class LogSettingService {
       String sanitized = sanitizeProperty(value);
 
       if(sanitized != null) {
-         return Tool.encryptPassword(value);
+         return Tool.encryptPassword(sanitized);
       }
 
       return null;
--- a/core/src/test/java/inetsoft/web/admin/logviewer/LogSettingServiceTest.java
+++ b/core/src/test/java/inetsoft/web/admin/logviewer/LogSettingServiceTest.java
@@ -1,0 +1,188 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (approved):

PR #4688, opened 2026-08-20
feat(wiz): eight layout tools — LayoutSessionService, read/property/mutation/undo services

larryliang-inetsoft:
## Summary
- Spec #11 (Layout) split out of spec #5, Lane I — its own hazards (undo stack, coordinate space, `scaleFont`) had nothing in common with bookmarks/parameters/export.
- Five new services under `inetsoft.web.wiz.viewsheet`, wired into `ViewsheetAssemblyAgentController` via 8 new endpoints under `/api/wiz/v1/agent/viewsheet/{sessionToken}/layout/...`:
  - `LayoutSessionService` — resolves Hazard 1 (a layout edit must never mutate the paired session's Master `Viewsheet`) via a disposable preview-clone runtime per `(sessionToken, layoutName)`, reused across calls and flushed on switch-away or session end.
  - `LayoutReadService` — `list_layouts`/`get_layout`, dual coordinate-space projection (layout-space and viewsheet-space read independently).
  - `PrintDeviceLayoutPropertyService` — `set_print_layout`/`manage_device_layout` via the existing `ViewsheetPropertyDialogService.setViewsheetInfo`, refusing `scaleFont: 0` (would blank crosstab text) and any device-catalogue-shaped write attempt.
  - `LayoutMutationService` — `edit_layout_objects`/`set_layout_table_options`.
  - `LayoutUndoService` — `layout_undo`/`layout_redo`, scoped per layout with a real stack (nested opens pop to the enclosing target, not straight to whole-sheet).
- Companion TS PR (stylebi-wiz): inetsoft-technology/stylebi-wiz#1720 (`layoutTools.ts`, 8 MCP tools).

## Real findings during implementation
- **`LayoutMutationService`'s task found the plan's own suggested approach was a Hazard-1 violation.** Delegating directly to `VSLayoutControllerService`'s `moveResizeLayoutObjects` would `apply()` a layout onto the master's own live `Viewsheet` (that method resolves `parentRvs` via the clone's `getOriginalID()`, i.e. the master's id). Fixed by calling `VSLayoutService`'s smaller, non-`apply()`-calling primitives directly against the master's `LayoutInfo` instead — verified with a manual red/green check (temporarily neutering the write, confirming the coordinate-space test fails, restoring it).
- **A real bug found and fixed at the root in the already-committed `LayoutSessionService`**: a read-only `resolveForRead` switch (e.g. a plain `get_layout` call that changes focus to a different layout) never reset the master's `layoutPoints` undo stack — only a mutating switch did. Since that stack is one flat structure per runtime, not per layout, a subsequent `layout_undo` could pop a checkpoint belonging to a *different* layout, and switching between a print and device layout specifically could throw a `ClassCastException`. Fixed unconditionally resetting the stack on every genuine switch, with new regression tests in both `LayoutSessionServiceTest` and `LayoutUndoServiceTest` (exercised end to end through the undo service's own entry point, not just the session service in isolation).
- **Task 6 hardcodes the layout `region` to `CONTENT`** rather than exposing header/footer/content as a caller-facing parameter, since `get_layout` also only ever projects `CONTENT` and the design spec never mentions region as a tool concept — avoiding a Jackson silent-zero-default landmine (`HEADER`, not `CONTENT`) of the same class as the `scaleFont` hazard.

## Test plan
- [x] Full `inetsoft.web.wiz.**` suite: 2,064 tests, 0 failures, 0 errors (rebased onto current `stylebi-wiz-main-rebase`)
- [ ] Live verification (Phase 5: Hazards 1–4 confirmed against a real running Composer + pairing session) — not run, needs the LIVE STACK token; tracked in `docs/superpowers/plans/2026-08-20-layout-implementation.md` in stylebi-wiz

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/vs/controller/VSLayoutService.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/controller/VSLayoutService.java
@@ -472,8 +472,12 @@ public class VSLayoutService {
 
    /**
     * Check if current assembly support editing tablelayout in printlayout.
+    *
+    * <p>Widened from {@code private} to {@code public} for {@code inetsoft.web.wiz.viewsheet
+    * .LayoutReadService} (2026-08-20-layout-implementation.md, Task 2), which projects this same
+    * check onto the agent-facing {@code get_layout} response rather than re-deriving it.
     */
-   private boolean supportTableLayout(VSAssembly assembly) {
+   public boolean supportTableLayout(VSAssembly assembly) {
       if(assembly == null) {
          return false;
       }
--- a/core/src/main/java/inetsoft/web/wiz/pairing/SheetSessionService.java
+++ b/core/src/main/java/inetsoft/web/wiz/pairing/SheetSessionService.java
@@ -74,6 +74,23 @@ public class SheetSessionService {
    public void close(String token) { if (token != null) sessions.remove(token); }
 
    /**
+    * Whether {@code token} still names a live (present, unexpired) session -- regardless of
+    * owner. For internal cleanup use only (see {@code LayoutSessionService}'s own scheduled
+    * sweep, which has no {@code Principal} to check against and no need for one: it is deciding
+    * whether to free a resource, not authorizing an action). NOT an authorization check --
+    * callers that need to verify the caller actually owns the session must still use
+    * {@link #resolve}.
+    */
+   public boolean isLive(String token) {
+      if(token == null) {
+         return false;
+      }
+
+      JoinSession s = sessions.get(token);
+      return s != null && !s.isExpired(clock.getAsLong());
+   }
+
+   /**
     * Ends every pane-scoped session bound to {@code socketSessionId} -- called when that
     * WebSocket session disconnects (crash, network drop, or a killed tab; see
     * {@code SheetSessionSocketCleanup}). A pane-scoped session (nullable
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/LayoutMutationService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/LayoutMutationService.java
@@ -1,0 +1,388 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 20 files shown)


larryliang-inetsoft:
Reviewed this against the layout design spec I wrote — the Hazard 1/2/3 mitigations are correctly implemented, not just claimed:

- Traced `openTemporaryViewsheet` → `OpenTemporaryViewsheetTask.doOpenTemporaryViewsheet` to confirm it really does construct a fresh `Viewsheet` object graph (`new Viewsheet(wentry)` + `vs.update(...)`) sharing no objects with the master — the load-bearing claim `LayoutSessionService`'s class doc makes for why `apply()` on the clone can never reach the master's assemblies.
- Confirmed `LayoutMutationService`/`LayoutUndoService` genuinely operate structural writes against `master`'s own `LayoutInfo` directly (not the clone) while routing only the `apply()` call itself onto the clone — exactly the split the class docs describe, not just asserted in a commit message.
- Confirmed `disposeAll` is really wired into the controller's `detach` endpoint (not just claimed).

One real bug found and fixed in 31ba77ff8, same defect class as the sibling Follow Focus PR (#4682) I also just fixed: `LayoutSessionService.clones` is keyed by `sessionToken`, and `disposeAll` only fires on explicit `detach_sheet`. A session that instead goes idle past TTL — an agent that finishes and stops, or crashes, without ever detaching — is reaped by `SheetSessionService.evictExpired()`, which has no knowledge of this class's `clones` map at all. That leaves a **live `RuntimeViewsheet` runtime** cached forever under a dead token — heavier than the sibling PR's leak, since it's a full parsed `Viewsheet` object graph, not a small `Deque` entry.

Fixed with `SheetSessionService.isLive(token)` (pure liveness, no owner check) exposed through `ViewsheetSessionService.isSessionLive`, and a new `LayoutSessionService.evictOrphanedClones()` scheduled sweep (same 10-minute cadence as `evictExpired`) that flushes any clone whose owning session is no longer live. 6 new regression tests. Full suite re-run: 2,068 tests, 0 failures, 0 errors.

One non-blocking observation, not fixed: `LayoutSessionService.resolveClone` still calls `layoutClone.apply(clone.getViewsheet())` on every mint/switch, but as far as I can tell nothing in the current tool set (`LayoutReadService`, `LayoutMutationService`) actually reads the clone's resulting viewsheet — reads go straight off `LayoutInfo`, and writes go straight at `master`. If that's intentional groundwork for a future preview/image tool, worth a one-line note saying so; if not, it's a wasted `openTemporaryViewsheet` + `apply()` on every layout focus/switch.

Build's earlier "failed" check was a 40-minute mid-suite hang in `inetsoft.sree.security.*` — unrelated package, unrelated to this diff, confirmed flaky by a clean re-run (19m5s, normal duration). Not this PR's problem.

CI clean, ready to merge from my side once the wiz-layout-tools branch settles.

PR #4691, opened 2026-08-20
Fix Bug #76167: give pdf.font.ratio a single value grammar

ngovus-inetsoft:
## Problem

`pdf.font.ratio` was read in two places that expected mutually exclusive value shapes, and both failed silently on the other's shape:

| Reader | Grammar | Effect |
|---|---|---|
| `PDFPrinter` instance initialiser → `fontRatio` | bare scalar (`Double.parseDouble`) | scaled the **drawn glyphs only** |
| `Gop.adjustSpacing` → `parseRatios` | `name:ratio;name:ratio` | scaled **measured width**, so layout follows |

A name-keyed value made `Double.parseDouble` throw into an ignored catch, leaving the ratio at 1.0. A bare scalar left `Gop`'s map empty, which set its `noratio` short-circuit.

The scalar reader was the damaging one: it scaled the glyphs without scaling the layout that measures them.

- Pagination is `Graphics`-free — `ReportSheet.printNext` takes a `StylePage`, so `PDFPrinter` first sees the text at `paint()` time, after every box is frozen.
- `TextPaintable.paint` takes its clip, alignment width and baseline from the unscaled `attr.font`, while the glyphs came from the font `PDFPrinter` substituted.

So a ratio above 1.0 drew text larger than the box reserved for it and truncated it at the clip; below 1.0 it under-filled and mispositioned centred and right-aligned text. **There was no safe non-1.0 value for that form.**

## Fix

Retire the scalar reader, so `pdf.font.ratio` means only what `Gop` already treats it as — the obsolete alias of `font.ratio.x`. The surviving form is the one that is integrated with layout.

A device-specific factor cannot be fed into `Common` instead: its measurement API is static and format-agnostic, and is shared with Excel, HTML, PNG and screen output. Making it device-aware would change pagination for every format.

Removing `fixFont` also fixes two related hazards, with no edits at those sites:

- `PDFPrinter.getFontMetrics()` returned metrics for the *scaled* font while `getFontMetrics(Font)` returned *unscaled* metrics for its argument — and ~20 painters (gauges, cylinders, thermometers, sliding scales, `VSCalendar`, `GraphUtil`) call `g.getFontMetrics()` directly.
- `PDFVSExporter` saved `printer.getFont()` and set it back, compounding the ratio to **ratio²**.

## Diagnosability

The remaining reader failed silently too, so `parseRatios` now warns instead of dropping input:

- a segment that is not a `name:ratio` pair (which is what a bare scalar looks like);
- a non-numeric ratio;
- use of the obsolete name, and the case where it is set but ignored because `font.ratio.x` is also set.

## Whitespace bug found while adding the above

`Tool.split` does no trimming, so `font.ratio.x=MS Hei:1.1; Algerian-bolditalic:1.02` — a space after the `;`, the natural way to write a list — stored the key `" Algerian-bolditalic"`, which `getNameWithStyle()` can never match. The pair was well-formed enough that the new warning did not fire either, so the ratio was dropped in silence. The name and ratio are now trimmed, with regression tests.

## Compatibility

- **`pdf.font.ratio=<scalar>`** — PDF text returns to the size layout was computed for. **This is an intentional rendering change**; the previous output was clipped or mispositioned.
- **`pdf.font.ratio=Name:1.1;…`** — unaffected. The `Gop` path is unchanged, and `PDFPrinter` previously threw into the ignored catch and left the ratio at 1.0 anyway.
- **Unset** (the common case) — byte-identical; `fixFont` already early-returned at `1.0`.

## Testing

- `core` `clean install`: **BUILD SUCCESS**; full suite **4359 tests, 0 failures, 0 errors** (67 skips all pre-existing and unrelated — Ignite cluster tests, `$KnownBugs` classes).
- Enterprise layer compiles clean against the rebuilt core, confirming nothing referenced the removed `fixFont`/`fontRatio` or the old 2-arg `parseRatios`.
- New `GopFontRatioTest` (8 cases) covers the pair grammar, style expansion, bare-scalar rejection, malformed input, and the whitespace regression. Tagged `@Tag("core")` so surefire's `<groups>core</groups>` filter actually runs it.

## Note for reviewers

Two pre-existing issues found but deliberately **not** changed here:

1. The lazy init of `fontratios`/`noratio` in `adjustSpacing` is unsynchronized on a singleton `Gop`, so two threads can both parse and both log. Shared with `font.ratio.y`; fixing it means touching a per-measurement hot path.
2. `PDFPrinterRoundRectTest` and `BoundsTest` — and ~71 other test files — lack `@Tag("core")`, so they never run in the build.

### DIFF
--- a/core/src/main/java/inetsoft/report/PDFPrinter.java
+++ b/core/src/main/java/inetsoft/report/PDFPrinter.java
@@ -1303,7 +1303,7 @@ public class PDFPrinter extends Graphics2D implements PDFDevice {
       }
 
       this.ofont = font;
-      this.font = fixFont(font);
+      this.font = font;
       fm = null;
       afm = null;
 
@@ -1313,26 +1313,6 @@ public class PDFPrinter extends Graphics2D implements PDFDevice {
       else {
          psFontName = null;
       }
-   }
-
-   private Font fixFont(Font font) {
-      if(fontRatio == 1.0) {
-         return font;
-      }
-
-      if(font instanceof StyleFont) {
-         StyleFont sf = (StyleFont) font;
-         sf = new StyleFont(sf.getName(), sf.getStyle(),
-                         (int) (sf.getSize() * fontRatio),
-                         sf.getUnderlineStyle(),
-                         sf.getStrikelineStyle());
-         return sf;
-      }
-      else if(font != null) {
-         font = font.deriveFont((float) (font.getSize2D() * fontRatio));
-      }
-
-      return font;
    }
 
    /**
@@ -7032,20 +7012,6 @@ public class PDFPrinter extends Graphics2D implements PDFDevice {
 
    private Font font; // current font
    private Font ofont = null; // current original font
-   private double fontRatio = 1.0;
-
-   {
-      try {
-         String prop = SreeEnv.getProperty("pdf.font.ratio");
-
-         if(prop != null) {
-            fontRatio = Double.parseDouble(prop);
-         }
-      }
-      catch(Exception ex) {
-         // ignore it
-      }
-   }
 
    protected FontMetrics fm = null; // cached font metrics
    protected FontMetrics afm = null; // cached afm font metrics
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):

PR #4692, opened 2026-08-20
Bug #76169: parse each boolean property the same way at every site

ngovus-inetsoft:
Fixes Bug #76169.

Thirteen properties were displayed with one string-comparison rule and enforced with another. Where the two disagreed, an Enterprise Manager checkbox showed one state while the product behaved as the other — and the next save of that page wrote the displayed state over the stored one, so an operator saw a setting they never touched being turned off.

`SreeEnv.getBooleanProperty(name, trueValues...)` tests membership of `trueValues` when the list is non-empty and falls back to `"true".equals` when it is empty; callers also hand-rolled `"true".equals`, `equalsIgnoreCase`, and `Boolean.parseBoolean`. Each property now uses one rule at every site. **The direction is always to widen — nothing that takes effect today stops taking effect.**

## What changed

| Properties | Display side | Enforcement side | Fix |
|---|---|---|---|
| `schedule.options.shareTaskInGroup`, `schedule.options.deleteTaskOnlyByOwner` | `"true"`, `"CHECKED"` | `"true".equals` only | pass the same true-value list |
| 9 × `pdf.*` booleans + `pdf.output.attachment` | `"true".equals` / `"embed".equals` | `equalsIgnoreCase` | widen the display side |
| `schedule.time.12hours` | `Boolean.parseBoolean` | `"true".equals` only | widen the 5 consumers |

Plus `PDF3Printer.emitStream()`, which re-read `pdf.output.ascii` instead of using the inherited `asciiOnly` flag that drives every other ASCII decision in `PDFPrinter`.

## The instance worth reviewing first

`schedule.options.*` is a **permissions** setting that appeared active and was not. The Schedule settings page displays both options with `"true", "CHECKED"`, and `defaults.properties` ships four of the eight `schedule.options.*` keys as `CHECKED` — so `CHECKED` is the established spelling for this family, not a hypothetical value. The enforcement paths in `ScheduleManager` passed no true-value list, so a stored `CHECKED` showed *Share Tasks Between Users in Same Group* and *Modify By Owner Only* ticked in Enterprise Manager while every enforcement path read them as off.

`hasShareGroupPermission()` also read the property inline with its own copy of the list; it now delegates to `isShareInGroup()`, since that duplication is the very mechanism of the bug.

## Behavior changes to be aware of

1. **On an install carrying a hand-written or inherited `CHECKED` for either `schedule.options` key, task sharing and owner-only deletion now actually take effect.** That is the defect being fixed, but it is a change in enforced behavior, not only in display. Neither key ships a default and the EM page always writes `Boolean.toString(...)`, so a normally-configured install is unaffected.
2. `setAsciiOnly()` is part of the public `PDFDevice` contract. A caller that sets it directly (via `PDF3PrinterJob`) previously got binary text with ASCII-encoded object streams — mixed encoding in one document; it now gets a consistently encoded one. The normal export path (`PDF3Generator.getPDFGenerator()` → `setAsciiOnly(property)`) is unchanged.

## Scope notes

- `SreeEnv` itself is **not** changed. Making its empty-list fallback case-insensitive would be a tidier root-cause fix but reaches 29 unrelated callers, against the minimal-scope guidance in `CLAUDE.md`.
- `pdf.output.attachment` is not in the original report. It is not a boolean, but it is rendered as a checkbox (*Embed PDF in browser*) and read with `equalsIgnoreCase` by both `ExportControllerService` and `VSExportService`, so it had the identical split and the identical symptom.
- `"true".equals(SreeEnv.getProperty(...))` appears ~250 times codebase-wide. Only properties with a genuine display/enforcement disagreement are touched; there is no codebase-wide sweep here.
- **Found and deliberately not fixed:** `Tool.getTimeFormat()` applies the 12-hour `HH`→`hh` rewrite regardless of its `schedule` argument but omits the flag from the cache key. So `getTimeFormat()`/`getTimeFormat(locale)` inherit a schedule-only setting, render `13:45:00` as an ambiguous `01:45:00` with no AM/PM marker, and cache under a key that never invalidates when the setting is toggled. This is a distinct pre-existing bug — nothing to do with case-sensitivity — and `getTimeFormat` also backs parsing (`CoreTool:1932`, `AssetUtil:1268`, `RepletRequest:262`), so it deserves its own issue rather than a ride-along here.

## Testing

36 new/extended tests. Each was verified to fail without the corresponding fix:

- `ScheduleManagerTest` (extended) — `CHECKED` on both permission keys; fails without the fix.
- `PresentationPdfGenerationSettingsServiceTest` (new, 17) — all 10 properties across `true`/`TRUE`/`True`/`tRuE` and the negative set; 8 failures without the fix.
- `ToolTimeFormatTest` (new, 2) — the `schedule.time.12hours` regression, via the `HH`→`hh` pattern rewrite; fails without the fix.
- `PresentationTimeSettingsServiceTest` (new, 3) — pins the display side, which already used `Boolean.parseBoolean`; passes either way by design.

`ScheduleManagerTest` does not assert the outcome of `hasShareGroupPermission`: past the flag it defers to `isSameGroup()`, which resolves groups through the security provider, and the fixture defines no users for `tuser0`/`admin`.

Regression sweep over `inetsoft.sree.schedule.**`, `inetsoft.web.admin.schedule.**`, `inetsoft.web.admin.presentation.**`, `inetsoft.report.pdf.**`, `inetsoft.util.Tool*`, `inetsoft.web.viewsheet.service.**`, `inetsoft.web.composer.vs.controller.**`: **546 run, 0 failures, 11 skipped** (all pre-existing).

Community-only change — no enterprise counterpart. Verified that nothing in `enterprise/`, `server/`, `shell/`, or `connectors/` reads any of these properties.

### DIFF
--- a/core/src/main/java/inetsoft/report/pdf/PDF3Printer.java
+++ b/core/src/main/java/inetsoft/report/pdf/PDF3Printer.java
@@ -19,7 +19,6 @@ package inetsoft.report.pdf;
 
 import inetsoft.report.PDFPrinter;
 import inetsoft.report.internal.Common;
-import inetsoft.sree.SreeEnv;
 import inetsoft.util.Encoder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -355,8 +354,10 @@ public class PDF3Printer extends PDFPrinter {
     */
    void emitStream(int id, byte[] buf, String[][] keys, boolean compress) {
       try {
-         boolean ascii =
-            "true".equals(SreeEnv.getProperty("pdf.output.ascii"));
+         // use the inherited flag rather than re-reading the property: PDF3Generator populates
+         // it from pdf.output.ascii, and setAsciiOnly() is part of the PDFDevice contract, so a
+         // caller that sets it directly must not get binary text with ascii-encoded streams.
+         boolean ascii = isAsciiOnly();
 
          byte[] coded = buf;
 
--- a/core/src/main/java/inetsoft/sree/schedule/ScheduleManager.java
+++ b/core/src/main/java/inetsoft/sree/schedule/ScheduleManager.java
@@ -916,7 +916,7 @@ public class ScheduleManager {
     * @return
     */
    public static boolean isShareInGroup() {
-      return SreeEnv.getBooleanProperty("schedule.options.shareTaskInGroup");
+      return SreeEnv.getBooleanProperty("schedule.options.shareTaskInGroup", "true", "CHECKED");
    }
 
    /**
@@ -924,7 +924,7 @@ public class ScheduleManager {
     * @return
     */
    public static boolean isDeleteByOwner() {
-      return SreeEnv.getBooleanProperty("schedule.options.deleteTaskOnlyByOwner");
+      return SreeEnv.getBooleanProperty("schedule.options.deleteTaskOnlyByOwner", "true", "CHECKED");
    }
 
    public static boolean hasShareGroupPermission(ScheduleTask task, Principal principal) {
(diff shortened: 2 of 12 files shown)


jshobe-inetsoft (approved):

PR #4694, opened 2026-08-20
Bug #76204, honor schedule.task.timeout=0 as no-timeout in ScheduleTaskCloudJob

jkidacki-inetsoft:
## Summary

`schedule.task.timeout=0` is documented, and implemented in `ScheduleTask.doRun` and `MVAction.createMV`, to mean "no timeout" — both guard enforcement with `timeout > 0`. `ScheduleTaskCloudJob`, used on cloud-runner deployments, did not guard it: it passed the value straight into `CountDownLatch.await(timeout, TimeUnit.MILLISECONDS)`, and a non-positive wait time makes `await` return `false` immediately. As a result, setting `schedule.task.timeout=0` on a cloud-runner installation caused every scheduled task to fail instantly with `"Scheduled task '...' timed out"`, instead of disabling the timeout.

## Root Cause

`ScheduleTaskCloudJob.execute()` (`core/src/main/java/inetsoft/sree/schedule/cloudrunner/ScheduleTaskCloudJob.java`) called `latch.await(timeout, TimeUnit.MILLISECONDS)` unconditionally. Per `CountDownLatch.await`'s contract, a non-positive wait time causes it to return immediately (as if it timed out), so `timeout=0` was treated as "already timed out" rather than "no timeout."

## Fix

Guard the wait with `timeout > 0`, falling back to an unbounded `latch.await()` otherwise — matching the existing, already-tested behavior in `ScheduleTask.doRun` (lines 548/637) and `MVAction.createMV` (line 273). `interrupt()`'s `latch.countDown()` still unblocks either branch identically, so task cancellation is unaffected.

## Test Plan

- `./mvnw test -pl core -Dtest="inetsoft.sree.schedule.**"` — 32/32 tests pass (`MVActionTest`, `ScheduleTaskTest`, `ScheduleTaskCompletionFlowTest`, `ScheduleTaskCloudJobTest`).
- `./mvnw clean install -pl core -am -DskipTests` builds cleanly.
- Manually verified the guard logic mirrors `MVAction.createMV`'s already-tested "timeout <= 0 → unbounded wait" semantics.

Fixes Redmine #76204.

### DIFF
--- a/core/src/main/java/inetsoft/sree/schedule/cloudrunner/ScheduleTaskCloudJob.java
+++ b/core/src/main/java/inetsoft/sree/schedule/cloudrunner/ScheduleTaskCloudJob.java
@@ -95,13 +95,17 @@ public class ScheduleTaskCloudJob implements InterruptableJob {
             }
          }
 
-         // timeout value might need to be configurable
-         if(!latch.await(timeout, TimeUnit.MILLISECONDS)) {
-            if(job != null) {
-               job.stop();
-            }
+         if(timeout > 0) {
+            if(!latch.await(timeout, TimeUnit.MILLISECONDS)) {
+               if(job != null) {
+                  job.stop();
+               }
 
-            throw new JobExecutionException("Scheduled task '" + taskName + "' timed out");
+               throw new JobExecutionException("Scheduled task '" + taskName + "' timed out");
+            }
+         }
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):

PR #4695, opened 2026-08-20
fix(wiz-binding): refuse groups/aggregates/options on a plain Table

larryliang-inetsoft:
## Summary
- A plain StyleBI `Table` assembly has no grouping or aggregation of its own — that's Crosstab's job — but `TableBindingModel` carried `groups`/`aggregates`/`option` fields that `VSTableBindingFactory.updateTableAssembly` never applied back to the real `TableVSAssemblyInfo`.
- The wiz agent's `set_table_fields`/`add_table_field`/`set_table_options` tools accepted writes to these and reported `{"ok":true}` while applying nothing (confirmed live via `get_table_binding` read-back and `get_viewsheet_image`).
- `move_table_field` compounded this into data loss: moving a field from a real shelf (`details`) onto one of these silently removed it from its source and never re-added it anywhere, while still reporting success.
- Fix: restrict `TableBindingModel` to its one real shelf (`details`) and to no options, reusing the existing "unknown shelf/option" refusal so both now fail loud by name instead of succeeding silently — the same pattern `set_column_labels` already uses for its own unimplemented-feature refusal.
- Found via the composer plugin test plan's L5 lane in stylebi-wiz (`docs/superpowers/plans/2026-08-17-consolidated-composer-plugin-test-plan.md`).

## Test plan
- [x] `./mvnw -o test -pl core -Dtest='TableBindingMutatorTest,TableBindingServiceTest'` — 58/58 pass
- [x] `./mvnw -o test -pl core -Dtest='inetsoft.web.wiz.**.*Test'` — 1929/1929 pass, no collateral breakage
- [ ] Live re-verification against a running Composer session (L5 cases 5.2-5.4) — pending rebuild/redeploy

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/TableBindingMutator.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/TableBindingMutator.java
@@ -25,7 +25,6 @@ import inetsoft.web.binding.model.BDimensionRefModel;
 import inetsoft.uql.XConstants;
 import inetsoft.web.binding.model.table.BaseTableBindingModel;
 import inetsoft.web.binding.model.table.CrosstabOptionInfo;
-import inetsoft.web.binding.model.table.TableOptionInfo;
 import inetsoft.web.binding.model.table.CrosstabBindingModel;
 import inetsoft.web.binding.model.table.TableBindingModel;
 import inetsoft.web.wiz.binding.model.FieldRef;
@@ -45,10 +44,6 @@ import java.util.*;
  * else no tool here touches.
  */
 public final class TableBindingMutator {
-   /** Shelves holding dimensions, per object type. */
-   private static final List<String> CROSSTAB_DIMENSION_SHELVES = List.of("rows", "cols");
-   private static final List<String> TABLE_DIMENSION_SHELVES = List.of("groups");
-
    /**
     * Natural spellings an agent is likely to reach for. The canonical names match
     * {@code CrosstabConstants} rather than inventing new ones.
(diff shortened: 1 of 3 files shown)


larryliang-inetsoft:
## Review — no blocking issues

1. **Root cause confirmed by reading source directly.** `VSTableBindingFactory.updateTableAssembly` (both `TableVSAssembly` and `EmbeddedTableVSAssembly` overloads) only ever reads `model.getDetails()` back into the real assembly — `getGroups()`/`getAggregates()`/`getOption()` on `TableBindingModel` were genuinely never applied. The silent-drop behavior this PR fixes is real, and the fix addresses the actual root cause, not a symptom.
2. **Scope is correct and minimal.** Only `TableBindingMutator.java` changes production behavior (`shelvesOf(TableBindingModel)` narrowed to `["details"]`, `optionVocabulary().table` narrowed to `[]`, `setTableOptions` now calls `requireKnown(options, List.of())`), reusing the existing refusal machinery so unknown-shelf/unknown-option errors fire by name. `CrosstabBindingModel` shelves, `setCrosstabOptions`, and `optionVocabulary().crosstab` are byte-for-byte untouched — Crosstab is not regressed.
3. Grepped for other production callers of `setShelf`/`setOptions` with `groups`/`aggregates`/`grandTotal`/`distinct` on a Table — none exist outside the two test files, so nothing else breaks.
4. **Independently re-ran tests** (not just trusting the PR description): `TableBindingMutatorTest` + `TableBindingServiceTest` — 58/58 pass. Full `inetsoft.web.wiz.**.*Test` — 2089 tests, 0 failures, 0 errors, 1 pre-existing unrelated skip.
5. CI green (build, Semgrep, Semgrep OSS, claude-review, CLA), `mergeable: MERGEABLE`, no conflicts.
6. Non-blocking nitpick: `TableBindingMutator.dimensionsOf()` still has a `case "groups" -> ((TableBindingModel) model).getGroups();` branch that's now dead/unreachable, since `requireShelf` throws before a Table shelf name can reach it as `"groups"`. Harmless, worth a follow-up cleanup, not blocking this PR.

**Verdict: approve.** Squash-merging.

PR #4696, opened 2026-08-21
fix(wiz): move_resize honors partial x/y/width/height patches

larryliang-inetsoft:
## Summary
- `edit_layout_objects`' `move_resize` op requires only "at least one of x/y/width/height" client-side (a move-only or resize-only call is by design), but the server's `LayoutMutationService.moveResize` required all four and crashed with `NumberFormatException` (`toInt(null)` → `Integer.parseInt("null")`) on a partial patch.
- Fixed by merging omitted fields onto the assembly's *current* position/size (from `assemblyLayout.getPosition()`/`getSize()`), so an omitted field means "unchanged" — not a crash, and not a silent zero that would invisibly shrink/relocate the object.
- Motivated by blocking finding #3 on inetsoft-technology/stylebi-wiz#1720.

## Test plan
- [x] New regression tests in `LayoutMutationServiceTest`: `moveOnlyPreservesExistingSize`, `resizeOnlyPreservesExistingPosition`
- [x] `./mvnw -o test -pl core -Dtest='inetsoft.web.wiz.viewsheet.LayoutMutationServiceTest'` — 9/9 pass
- [x] `./mvnw -o test -pl core -Dtest='inetsoft.web.wiz.**.*Test'` — 2091 pass, 0 fail, 1 skipped (unrelated)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/LayoutMutationService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/LayoutMutationService.java
@@ -210,9 +210,21 @@ public class LayoutMutationService {
                VSAssemblyLayout assemblyLayout =
                   requireAssemblyLayout(layout, name, region, layoutName);
 
-               Point position = new Point(toInt(object.get("x")), toInt(object.get("y")));
-               Dimension size = new Dimension(toInt(object.get("width")),
-                                               toInt(object.get("height")));
+               // A move-only or resize-only call must leave the other dimension unchanged --
+               // merge onto the assembly's current position/size rather than defaulting an
+               // omitted field to 0, which would silently shrink/relocate it instead.
+               Point currentPosition = assemblyLayout.getPosition();
+               Dimension currentSize = assemblyLayout.getSize();
+               int x = object.containsKey("x") ? toInt(object.get("x")) :
+                  (currentPosition != null ? currentPosition.x : 0);
+               int y = object.containsKey("y") ? toInt(object.get("y")) :
+                  (currentPosition != null ? currentPosition.y : 0);
+               int width = object.containsKey("width") ? toInt(object.get("width")) :
+                  (currentSize != null ? currentSize.width : 0);
+               int height = object.containsKey("height") ? toInt(object.get("height")) :
+                  (currentSize != null ? currentSize.height : 0);
+               Point position = new Point(x, y);
+               Dimension size = new Dimension(width, height);
                assemblyLayout.setPosition(position);
                assemblyLayout.setSize(size);
 
--- a/core/src/test/java/inetsoft/web/wiz/viewsheet/LayoutMutationServiceTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/viewsheet/LayoutMutationServiceTest.java
@@ -93,6 +93,45 @@ class LayoutMutationServiceTest {
    }
 
    /**
+    * A move-only call (only x/y, no width/height) must leave the object's existing size alone --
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft:
**Independent review (isolated worktree, not the shared checkout)**

Confirmed the fix does what it claims: `moveResize` now merges omitted x/y/width/height onto the assembly's *current* `assemblyLayout.getPosition()`/`.getSize()` rather than defaulting to 0 (unlike the sibling `addObjects` `getOrDefault(..., 0)` pattern, which would have silently zeroed the omitted dimension instead of leaving it unchanged). Traced `VSAssemblyLayout`'s no-arg ctor (sentinel `(-1,-1)`) vs. the `(name, position, size)` ctor used by `VSLayoutService.createAssemblyLayout` for both the existing-assembly and new-object add paths -- every `assemblyLayout` reachable by `move_resize` (via `requireAssemblyLayout`, which only returns layout entries already present in the layout) is constructed with real, non-null position/size. So the null-fallback-to-0 branch in the new code is unreachable in practice, and there's no regression for freshly-added objects.

Regression tests `moveOnlyPreservesExistingSize` / `resizeOnlyPreservesExistingPosition` genuinely exercise the partial-patch merge (fixture installs Table1 at (300,400)/(50,60), then asserts the omitted fields are preserved rather than zeroed) -- not just happy-path full-geometry calls.

Independently re-ran tests: `LayoutMutationServiceTest` standalone (9/9 pass), and the full `inetsoft.web.wiz.**.*Test` sweep for collateral breakage (2091 passed, 0 failed, 1 skipped).

CI is green (build, claude-review, Semgrep OSS/Scan all pass); `license/cla` pending is the same non-blocking pattern already seen on the merged #4695 from this same account.

**Verdict: no blocking issues found.** Note: GitHub won't let me submit a formal approval since the PR author and my `gh` identity are the same account (`larryliang-inetsoft`) -- reporting this back to the requester rather than working around it.

larryliang-inetsoft:
## Review — no blocking issues

1. Diff matches the claimed fix exactly. `LayoutMutationService.moveResize` now computes each of x/y/width/height as `object.containsKey("field") ? toInt(...) : currentPosition/currentSize.field` — merging omitted fields onto the assembly's CURRENT position/size, not a zero-default like the sibling `addObjects` method's `getOrDefault(..., 0)`. Correct semantics: an omitted field means "unchanged."
2. Checked the "no prior layout position/size" edge case: `VSAssemblyLayout`'s no-arg ctor sets a sentinel `(-1,-1)`, but every assemblyLayout reachable via `move_resize`'s `requireAssemblyLayout` is constructed through `VSLayoutService.createAssemblyLayout`'s `(name, position, size)` ctor with real computed values, for both existing-assembly and brand-new-object add paths. The `!= null ? ... : 0` fallback in the new code is unreachable in practice — no regression for freshly-added objects.
3. Regression tests `moveOnlyPreservesExistingSize`/`resizeOnlyPreservesExistingPosition` are meaningful, not happy-path only — fixture installs Table1 at (300,400)/(50,60), and assert the *omitted* fields stay at those exact values while the patched fields change.
4. Independently re-ran tests: `LayoutMutationServiceTest` standalone — 9/9 pass. Full `inetsoft.web.wiz.**.*Test` — 2091 passed, 0 failed, 1 pre-existing unrelated skip.
5. CI green (build, claude-review, Semgrep OSS/Scan); `license/cla` pending is the same pattern already merged without issue on #4695 (same author account), not a real blocker.

**Verdict: approve.** Squash-merging.

PR #4697, opened 2026-08-21
fix(crosstab): resolve entity-qualified sort-by-value column names

larryliang-inetsoft:
## Summary
- `set_field_sort`'s `sortByField` (needed for `direction: "value_asc"/"value_desc"`) silently failed to resolve when the target aggregate's underlying column carries an entity/table qualifier — the normal case for any real, non-embedded datasource column (as opposed to an in-memory/embedded table). The crosstab then silently fell back to sorting by label, exactly the trap the tool's own validation message warns about, except triggered by a value that looks correct.
- Root cause: `AbstractCrosstabVSAQuery#findCol` matched the sort target only against `DataRef#getName()`, which is entity-qualified (e.g. `"CUSTOMERS.CUSTOMER_ID"`), while every sibling resolution path in this area — `VSCrosstabInfo#fixCol` (normalizes a caller-supplied full aggregate name down to the aggregate's entity-free VS name before it's stored as the runtime `sortByCol`), `ColumnSelection#getAttribute` (used to resolve every other bare column reference in this API, including ranking's `measure`), and the aggregate's own `getFullName()`/`getVSName()` — all use the entity-free form.
- Fix: `findCol` now also matches on the aggregate's entity-free `DataRef#getAttribute()`, mirroring how every other bare-column reference in this API already resolves.

## Test plan
- [x] New regression test `CrosstabSortByValueEntityQualifiedNameTest` builds a real crosstab (via `CrosstabVSAQuery`) whose bound columns carry an entity qualifier, and confirms `value_desc` sort-by-value now produces the correct count-descending order for both a bare column name (`"CUSTOMER_ID"`) and the fully-qualified aggregate name (`"Count(CUSTOMER_ID)"`) as `sortByField` — both previously failed once an entity qualifier was involved (the bare form always failed; run's own investigation also found the full-name form silently degrades to the same failure via `VSCrosstabInfo#fixCol`'s bare-name normalization).
- [x] `DimensionSortRankingTest`, `TableBindingMutatorTest`, `CrosstabVSAQueryTest`, `CrossTabFilterTest`, `VSCrosstabBindingScriptableTest` — all pass, no regressions in the model-level sort/ranking vocabulary or existing crosstab query tests.
- [x] Full `inetsoft.web.wiz.**.*Test` suite — 2089 tests, 0 failures/errors (1 unrelated skip).

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
@@ -1047,6 +1047,16 @@ public abstract class AbstractCrosstabVSAQuery extends CubeVSAQuery
          if(colname.equals(aggrs[i].getName())) {
             return i;
          }
+
+         // The aggregate's full/VS name (used above, and by VSCrosstabInfo.fixCol() to
+         // normalize a matching sortByCol) is always entity-free, but getName() is
+         // entity-qualified whenever the underlying column comes from a real (non-embedded)
+         // table -- the normal case. A sortByField naming the bare column, exactly like every
+         // sibling API (set_field_ranking's measure, set_table_fields/add_table_field's
+         // column), then silently fails to match here and the sort falls back to the label.
+         if(colname.equals(aggrs[i].getAttribute())) {
+            return i;
+         }
       }
 
       return -1;
--- a/core/src/test/java/inetsoft/report/composition/execution/CrosstabSortByValueEntityQualifiedNameTest.java
+++ b/core/src/test/java/inetsoft/report/composition/execution/CrosstabSortByValueEntityQualifiedNameTest.java
@@ -1,0 +1,157 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.report.composition.execution;
+
+import inetsoft.report.TableLens;
+import inetsoft.test.BaseTestConfiguration;
+import inetsoft.test.ConfigurationContextInitializer;
+import inetsoft.test.IntegrationTestConfiguration;
+import inetsoft.test.SreeHome;
+import inetsoft.test.SwapperTestConfiguration;
+import inetsoft.uql.ColumnSelection;
+import inetsoft.uql.VariableTable;
+import inetsoft.uql.XConstants;
+import inetsoft.uql.asset.AbstractSheet;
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft:
Reviewed the diff, the full findCol() method in context, and independently re-ran tests. (Approve is blocked: GitHub rejects self-approval for the acting identity — proceeding via merge per team lead's pre-authorization for this identity-collision case.)

**Fix correctness**: Confirmed genuinely additive. The new `colname.equals(aggrs[i].getAttribute())` check is a plain `if` appended after the existing `getName()` check inside the same loop iteration; both branches `return i` immediately, so the new check only ever fires when the pre-existing checks already failed to match — no previously-working match path is altered.

Traced why the bug is universal for real tables: `VSCrosstabInfo.update()` normalizes the runtime `sortByCol` via `fixCol()`, which rewrites *any* matching input (bare name or full aggregate expression) down to the aggregate's entity-free `VSName` before `findCol()` ever sees it. Since `getName()` is entity-qualified whenever the column has a real entity, that check essentially never matches post-normalization for non-embedded tables — only the new `getAttribute()` (entity-free) check does. This also explains why both of the test's input forms (bare column, and `Count(COL)`) required the fix.

Checked the false-positive-ambiguity angle: two aggregates from different entities sharing a bare attribute name would resolve to first-match-in-array-order under the new check. This ambiguity is inherent to every other bare-name resolution path in this API (`ColumnSelection#getAttribute`, ranking's measure lookup) — not something newly introduced here.

**Test quality**: `CrosstabSortByValueEntityQualifiedNameTest` builds and executes a real `CrosstabVSAQuery` against an in-memory embedded table (not mocked to the point of being tautological). Independently verified both directions:
- At the parent commit (66e673094) with the test copied in: both cases fail exactly as predicted (`expected: <NJ> but was: <CA>` — label-sort fallback).
- At this PR's head (8c53a6601): both pass.

Also independently ran the full requested matrix (`CrosstabSortByValueEntityQualifiedNameTest`, `DimensionSortRankingTest`, `TableBindingMutatorTest`, `CrosstabVSAQueryTest`, `CrossTabFilterTest`, `VSCrosstabBindingScriptableTest`) — 72/72 pass — and the broader `inetsoft.web.wiz.**.*Test` surface — 2089/2089 pass, 0 failures.

**CI**: The `build` check failure is unrelated to this change — it's `DashboardRegistryOrgLifecycleTest` (org-lifecycle/dataspace-rename), a test with a known history of flakiness in this repo (see `disable-flaky-dashboard-registry-org-lifecycle-test`, `disable flaky rename_setOrganizationInfo... org-lifecycle test`). This PR doesn't touch that file or anything in its dependency path. `license/cla` PENDING is consistent with other same-day merged PRs (e.g. #4696) and isn't a required/blocking check.

No blocking issues found.

larryliang-inetsoft:
## Final check before merging

Independently verified the `build` check failure: `DashboardRegistryOrgLifecycleTest.rename_copyOrganizationInternal_dashboardFilesRelocatedByDataSpaceRename_contentNotRewritten` — an unrelated org-lifecycle/dataspace-rename test (6846 run, 1 failure). This test has a documented history of flakiness in this repo (`disable-flaky-dashboard-registry-org-lifecycle-test`, `test(sree): disable flaky rename_setOrganizationInfo_... org-lifecycle test`). This PR touches only `AbstractCrosstabVSAQuery.findCol` and its own new test — nothing in the org/dataspace path. Treating as a pre-existing flake, not a real blocker.

Everything else per review-4697's findings: additive fix (only fires when the existing entity-qualified match already failed), root cause traced end-to-end and confirmed (`VSCrosstabInfo.fixCol()` normalizes to entity-free form before `findCol()` runs, so the old check rarely matched for real tables), regression test independently confirmed to fail before the fix and pass after, full requested test matrix green (72/72 + 2089/2089 wiz-agent suite).

**Approving and squash-merging.**

PR #4701, opened 2026-08-21
fix(chart): resolve a named legend instead of hiding every legend

fredaan:
## The defect

`set_chart_element_visibility` with `element: "legend"` and a `target` hid **every** legend on the chart, while the summary reported having hidden the one that was named.

Root cause is a missing field on one event. `VSChartLegendsVisibilityService` reads a hide whose `aestheticType` is null as *hide all* and routes it to `showAllLegends(false)`; `ChartElementService` only ever sent `field`, so `target` was dead on the legend path for every caller.

This was reviewed once before and accepted as "a bogus target is an inert no-op". That reading was wrong, and the reason it survived is worth stating: on a **single-legend** chart, "hides the one you named" and "hides them all" produce the identical image, so only reading the source separated them. Reproduced on a two-legend chart, naming a real field killed both legends.

## The fix

Same shape as the axis resolver already in this file: **enumerate what the chart has, then refuse what it does not.**

- `ChartRegionResolver.legends` reads the real-size graph's `LegendGroup` and describes each legend — field, channel, target fields, node-aesthetic flag — taking every value from where `LegendArea` takes it, so the answer *is* what the Composer sends for that legend rather than a second opinion about it. A legend whose frame cannot be described keeps its slot, because the index is a target the region tools address.
- `ChartElementService` resolves a named legend to its `aestheticType` and computes `colorMerged` by the same test `VSChartService.hideLegend` makes over the same list.

Three decisions worth review attention:

1. **An unresolvable name is refused, not passed through.** This is the one target in this family where a wrong value was never inert, so the accepted-no-op rule cannot apply to it.
2. **A chart with no laid-out graph is refused too** — deliberately the *opposite* of the rule `requireLegend` follows for the legend index. There, an unmeasured count is zero for want of a graph and refusing would block a legitimate write, so silence is the safe side; here the fallback *is* the destructive behaviour.
3. **`shape`, `line` and `texture` are one family.** Found while verifying: `get_chart_aesthetics` reports the field on the **shape** channel while a line chart renders that legend as **line**, so a caller who read the aesthetics and said `shape` was naming the legend correctly and being refused for it. `GraphUtil.getLegendDescriptor` maps all three onto the same descriptor, so there was nothing ambiguous to protect. Two legends in one family are still refused with both fields named.

`list_chart_elements` reports the legends as well — filtering only the write would have moved the guessing rather than removed it.

## Also fixed

`resize_chart_plot`'s `reset: true` cleared `unitWidthRatio`/`unitHeightRatio` and both resized flags but left `unitWidthRatioPercent`/`unitHeightRatioPercent` behind. `VGraphPair` re-derives the ratio from the percent whenever it is `>= 1` **without consulting the flags**, so a reset chart reported `default: true, resized: false` while rebuilding the old ratio on every subsequent layout — visually inert only because the enlargement itself is gated on the flag this branch does clear.

Cleared back to `0`, the field's declared default. `1` is not the neutral value here: it would re-derive the ratio as the plot's own baseline, which looks like a reset that did not take.

## Verification

Unit: `ChartElementServiceTest` and `ChartRegionResolverTest`, 57 green (15 new).

Live, against a running server on a chart with two legends (`Customer:Region` on colour, `Customer:Reseller` on shape) and a real y2, every step confirmed by rendered image:

| Call | Before | After |
|---|---|---|
| `target: "Customer:Region"` | both legends hidden, summary named one | only that one hidden |
| `target: "Customer:Reseller"` | both legends hidden | only that one hidden — the colour legend survives, which is bug 58639's case and what `colorMerged` decides |
| `target: "NoSuchField"` | **hid the real legends**, reported success | refused, listing the real legends |
| `visible: true`, no target | shows all | unchanged |
| `reset: true` after `ratio: 6` | `percent` stayed 1.1 | `ratio: 1, resized: false, percent: 0, default: true` |

`list_chart_elements` reported both legends and, after one was hidden, only the other — read and write agreeing.

**Not yet exercised live:** the shape/line/texture fold, which was written after the restart the live run used. It is unit-tested. Also no live coverage for a **relation chart** (the node-aesthetic branch) or for **two legends on the same channel** (the ambiguity refusal).

---

## Review round (e4c26c3)

Four things found reviewing the above.

**A refusal cost an undo step.** The resolution ran inside `sessions.mutate`, which adds a checkpoint and broadcasts a refresh in a `finally` — deliberately, because a composer service can partially apply before it ERRORs. A resolution that refuses has applied nothing, so it was spending a Ctrl+Z step and a Composer refresh on a call that changed nothing. It now runs in its own session read ahead of the mutation, the route `ChartRegionPropertyService` already takes for its own refusals. Regression-tested (`refusingALegendTargetDoesNotOpenAMutation`).

**A blank target meant two different things at the two layers.** The tool's show-with-a-target guard reads `""` as absent, so a server reading it as present refused the very call the tool had let through — and on the legend path it refused it by naming an empty legend. Blank is now no target, for all three elements. Not trimmed otherwise: only the blank case is unambiguous, and a column name's own spacing is its own business.

**"One read for both answers" was not what the code did.** `resolve` and `legends` each fetched their own `VGraphPair`, so the axes and the legends could describe two layouts either side of a relayout while being reported as one chart's. They now come off one pair, through `regions()` — the two halves still read from different graphs (expanded for the axes, real-size for the legends), which is why the record documents that rather than hiding it.

**Two reporting inaccuracies.** `channel` was emitted as `""` for a legend with no describable frame while `field` was omitted on the stated ground that an empty string reads like a nameable target — the same ground covers both, so both are omitted now and the index stands alone. And `aestheticType` was documented as `Color`, `Shape` or `Size`; `Legend.createLegend` also builds `Texture` and `Line` legends, which is the whole reason `channelFamily` exists.

Also dropped `Legends.fields()`, which no production caller had, and used the `RuntimeViewsheet` import instead of naming the class in full.

Verified: `ChartElementServiceTest` + `ChartRegionResolverTest` 59 green (2 new); the whole `inetsoft.web.wiz` package 2103 green.

### DIFF
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/chart/VSChartPlotResizeService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/chart/VSChartPlotResizeService.java
@@ -57,6 +57,16 @@ public class VSChartPlotResizeService extends VSChartControllerService<VSChartPl
          if(reset) {
             chartInfo.setUnitWidthRatio(1);
             chartInfo.setUnitHeightRatio(1);
+            // The percents have to go too, and back to 0 rather than 1: VGraphPair re-derives
+            // unitWidthRatio/unitHeightRatio from them whenever they are >= 1, without consulting
+            // the resized flags, so a reset that left them behind reported default/not-resized
+            // while the old ratio was rebuilt on every subsequent layout. (1 is not the neutral
+            // value here - it would re-derive the ratio as the plot's own baseline. 0 is the
+            // field's declared default and the only value that turns the re-derivation off.) It
+            // was invisible only because the enlargement itself, minPlot *= ratio, is gated on
+            // the flag this branch does clear.
+            chartInfo.setUnitWidthRatioPercent(0);
+            chartInfo.setUnitHeightRatioPercent(0);
             chartInfo.setWidthResized(false);
             chartInfo.setHeightResized(false);
          }
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/ChartElementService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/ChartElementService.java
@@ -25,6 +25,7 @@ import inetsoft.graph.VGraph;
 import inetsoft.graph.coord.Coordinate;
 import inetsoft.graph.coord.RelationCoord;
 import inetsoft.graph.internal.GTool;
+import inetsoft.report.composition.RuntimeViewsheet;
 import inetsoft.report.composition.execution.ViewsheetSandbox;
 import inetsoft.report.composition.graph.*;
 import inetsoft.report.composition.region.ChartArea;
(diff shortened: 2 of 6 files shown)


fredaan:
## Re-review after e4c26c3 — no new findings

Second pass over the whole diff, including my own review-round commit. Nothing left that I would block on.

### What I checked against source rather than taking on trust

- **`colorMerged` is a faithful port.** `vs-chart.service.ts:401-402` computes `colorCount < ("Color" == aestheticType ? 2 : 1)` over `model.legends`; the Java version makes the identical test over `Legends.legends()`. And the two lists are the same list: `LegendsArea.getLegendAreas` builds exactly one `LegendArea` per `getLegendCount()`, and `GraphBuilder` builds one `LegendContainer` per area, so index and count line up end to end.
- **Every legend fact is taken where `LegendArea` takes it.** `GTool.getFrameType(frame.getClass())`, `GraphUtil.getTargetFields(frame, egraph)`, and `isNodeAestheticFrame(frame, legend.getGraphElement())` — the last matching `LegendItemArea:76` / `LegendTitleArea:63`. The graph is the right one too: `ChartArea:289` reads its legend group off the real-size vgraph while its axis areas come from the expanded one, which is why `regions()` splits the pair the way it does.
- **The shape/line/texture fold is real, in both branches.** `GraphUtil.getLegendDescriptor` maps `SHAPE_LEGEND`/`LINE_LEGEND`/`TEXTURE_LEGEND` onto `getShapeLegendDescriptor()` in the single-aesthetic branch and onto `getShapeField()` in the multi-aesthetic one. Nothing ambiguous is being collapsed.
- **The null-frame slot is reachable, so guarding it is right.** `Legend.createLegend` returns `null` for a `TextFrame`/`FontFrame` and `LegendGroup` adds it unconditionally, so `getLegend(i)` can be null. (`LegendArea` would NPE there — pre-existing, and not made worse here.)
- **The `reset` percent fix.** `VGraphPair:690-695` does re-derive the ratio from the percent whenever it is `>= 1`, without consulting `isWidthResized`/`isHeightResized` (those gate `minPlot *= ratio` further down, at 725/739). `0` is the declared default at `AbstractChartInfo:3843/3845`, and no other consumer of these fields treats `1` as neutral — so `0` rather than `1` is the correct choice and the reasoning in the comment holds.
- **`requireLegendField` cannot return an undescribable legend.** A null `aestheticType` folds to `""`, and the channel filter is skipped when the wanted family is empty — so the null-frame slot is unreachable by any target, which is what `describe()` now reports by omitting both keys.

### Two things left on the record, deliberately not changed here

1. **A refusal inside `sessions.mutate` costs an undo step, and that is systemic rather than this PR's.** My commit moved the legend resolution out because a wrong legend name is an easy caller mistake and the refusal is new. But the pattern `sessions.mutate(... require*(rvs, name); ...)` is the norm across the wiz services (`ViewsheetEditService` alone has six), and `mutate` checkpoints and broadcasts in a `finally` unconditionally — `RuntimeSheet.addCheckpoint` does not dedupe. So an unknown-assembly refusal still spends a Ctrl+Z step on the axis and title paths here, and everywhere else. Worth one deliberate decision across the family rather than a half-measure in this file.
2. **`requireLegendField`'s javadoc enumerates `color`, `shape`, `size`** as the accepted channel spellings. Not wrong — the next clause explains the fold, so `line` and `texture` are covered in sequence — and left alone on that basis. The one that *was* wrong (`LegendTarget`'s `@param aestheticType`, which claimed only three values exist) is fixed.

### What I did not do

No live run. The before/after table in the description is the author's; I verified the mechanism it claims by reading the source, not by reproducing it. The **shape/line/texture fold still has no live coverage** — as the description says, it was written after the restart the live run used — and neither does the relation-chart node-aesthetic branch or two legends on one channel.

### Verification

`ChartElementServiceTest` + `ChartRegionResolverTest`: **59 green** (2 new). Whole `inetsoft.web.wiz` package: **2103 green, 0 failures, 1 skipped**.

NeverStopFMH (approved):

PR #4706, opened 2026-08-21
Test L7 and fix issues on it

lizhepu:
**Eight defects, four commits**, all found by live-testing the composer agent tools against a running Composer. Every one is name/value mapping or validation — no business logic changes.

| Commit | Defects |
|---|---|
| `c637d358a` | `enabled` written to a dead field |
| `af3ee3b1f` | four writable properties refused as unwritable |
| `6e9d9976f` | seven read-only flags accepted writes; a mistyped token silently defaulted |
| `e6b186a4e` | a clear that cleared nothing; `unknown(9)` reads; viewsheet links always threw; `colName` never echoed |

All eight are **verified live** — see Verification.

## 1. `enabled` was mapped to a dead field (`c637d358a`)

The alias pointed at `basicGeneralPaneModel.enabled`, which is the Editable flag under a misleading name (its setter is declared `setEnabled(boolean editable)`) and which nothing in the tree ever writes. So `{enabled: true}` reported success, wrote a field no one reads, and always read back `false`. Now points at `GeneralPropPaneModel.enabled`, the field every property dialog service fills from `VSAssemblyInfo.getEnabledValue()`.

## 2. Writable properties were refused as unwritable (`af3ee3b1f`)

`propertiesOf` derives a name by stripping `get` and decapitalizing; `setterFor` rebuilt it with `capitalize`. Not inverses when an accessor has a lowercase letter right after the prefix, so `ChartLinePaneModel.setxGridLineStyle` was invisible: `xGridLineStyle`, `xGridLineColor`, `yGridLineStyle` and `yGridLineColor` were refused as unwritable *and listed as available in the same message*. Both spellings are now accepted rather than one replacing the other, since `getRTChartType`-style accessors resolve only through the rebuilt form — a strict superset. `getterFor` gets the same treatment.

## 3. Read-only flags accepted writes; a mistyped token failed silently (`6e9d9976f`)

`gridLineVisible`, `innerLineVisible`, `trendLineVisible`, `facetGridVisible`, `facetGridEnabled`, `projectForwardEnabled` and `lineTabVisible` are flags the dialog uses to decide which controls to show. They are recomputed from the chart type on every read and `updateChartLinePaneModel` writes none of them, so a write could not survive. Now readable and refused on write, naming what to change instead. (`facetGrid` is deliberately not in that set — it is the real setting and applies.)

`requireAllowedValue` also returns the domain's own spelling instead of only validating: the check is case-insensitive, so `trendLineType: "LINEAR"` passed it, was stored verbatim, and left the chart on `NONE`, because `getIndexByName` matches exactly and defaults to index 0. `projectForwardEnabled` is exposed readable so a caller can see why `projectForward` reads back 0 — that is gating, not a lost write.

## 4. Hyperlinks — five defects on two root causes (`e6b186a4e`)

`LINK_TYPES` mapped `"none"` to `0`, but `HyperlinkDialogService` uses a private `NONE = 9` sentinel. That single wrong value broke both directions: a never-linked assembly carries 9 and reported itself as `unknown(9)`; and a write of `"none"` sent 0, which `getHyperlink` does not treat as clear, so it built a real `Hyperlink` of type 0 — tooltip and target frame survived, while `tokenOf(0)` read it back as `"none"`. `NONE` is now `public static final` and documented as a dialog sentinel, not a `Hyperlink` constant.

A viewsheet link is stored by `assetLinkId`, which nothing ever set — it reached `VSUtil.getBookmarks` as null and surfaced as a raw NPE. `resolveViewsheetTarget` derives the id from `assetLinkPath` (global scope, then the caller's own), verifies it with `containsEntry`, and refuses a path that resolves to nothing. It runs **before** `sessions.mutate`, so an unresolvable target no longer throws with the model half-written — that half-write is what left assemblies carrying `linkType=viewsheet` with a null destination. `get_hyperlink` also echoes the requested `colName` when the dialog model does not carry it.

## Verification

Verified live in two passes, each against a rebuilt server, and read-backs were required throughout — an `{"ok": true}` was not accepted as evidence, which is what caught three of these in the first place.

**Pass 1 (`6e9d9976f`)** — commits 1–3: `enabled` round-trips both ways on both nesting shapes (`Text1` via `outputGeneral`, `Crosstab1` via `dataGeneral`); `xGridLineStyle`/`yGridLineColor` are accepted where they were refused, **and the render draws the red gridline**, not just the read-back; all nine of the chart line pane's real settings round-trip in one patch, with the yellow quadrant and blue diagonal lines visible; `gridLineVisible` is refused with its explanation; `trendLineType: "LINEAR"` canonicalizes to `"Linear"` and takes effect.

**Pass 2 (`e6b186a4e`)** — commit 4: a never-linked assembly reports `none`; **a cleared link reads back byte-identical to a never-linked one**, which is the real proof it was dropped rather than replaced by a type-0 link; a viewsheet link succeeds and reads back complete; a bad path is refused naming both scopes tried, **and a good link written moments earlier survives that refusal untouched** — the half-apply was the actual defect, so this read-after-refusal is the assertion that matters; `colName` echoes back.

## Known: two tests need updating

Left alone deliberately, both in `AssemblyHyperlinkServiceTest`:

- `clearingALinkNeedsNoDestination` asserts `linkType == 0`, which pinned the bug — it should expect `HyperlinkDialogService.NONE`.
- `harness()` must stub `getAssetRepository()` (and `containsEntry`) for `setsAViewsheetLink`, since the target now resolves through the repository.

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/vs/dialog/HyperlinkDialogService.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/dialog/HyperlinkDialogService.java
@@ -817,7 +817,15 @@ public class HyperlinkDialogService {
    private final VSObjectPropertyService vsObjectPropertyService;
    private final ViewsheetService viewsheetService;
    private final VSTrapService trapService;
-   private final int NONE = 9;
+   /**
+    * The dialog's "no link selected" sentinel. <b>Not a {@link Hyperlink} constant</b> — it is the
+    * value {@code getHyperlinkDialogModel} reports when an assembly has no hyperlink, and the value
+    * {@code getHyperlink} tests to decide to clear one. Public because any caller driving this
+    * service without the Angular dialog has to speak it: sending anything else (0, say) makes
+    * {@code getHyperlink} build a real {@code Hyperlink} carrying that bogus type instead of
+    * clearing, so the link survives and only looks cleared.
+    */
+   public static final int NONE = 9;
    private final VSAssemblyInfoHandler assemblyInfoHandler;
    private final HighlightService highlightService;
    private final DataRefModelFactoryService dataRefModelService;
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyHyperlinkService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyHyperlinkService.java
@@ -18,6 +18,9 @@
 package inetsoft.web.wiz.viewsheet;
 
 import inetsoft.report.Hyperlink;
+import inetsoft.sree.security.IdentityID;
+import inetsoft.uql.asset.AssetEntry;
+import inetsoft.uql.asset.AssetRepository;
 import inetsoft.web.composer.model.vs.HyperlinkDialogModel;
 import inetsoft.web.composer.vs.dialog.HyperlinkDialogService;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -45,9 +48,21 @@ import java.util.*;
  */
 @Service
 public class AssemblyHyperlinkService {
-   /** Agent-facing link types. The integers never appear in either direction. */
+   /**
+    * Agent-facing link types. The integers never appear in either direction.
+    *
+    * <p><b>{@code none} is {@link HyperlinkDialogService#NONE}, not 0.</b> That sentinel is what the
+    * dialog service reports for an assembly with no link and what it tests to decide to clear one,
+    * and mapping {@code none} to 0 instead broke this class in both directions at once. A read of a
+    * never-linked assembly carried the sentinel 9, which matched no entry here and so reported
+    * itself as {@code "unknown(9)"} rather than as {@code none}. A write of {@code none} sent 0, which
+    * {@code HyperlinkDialogService.getHyperlink} does not recognise as "clear", so it built a real
+    * {@code Hyperlink} carrying type 0. The link survived with its tooltip and target frame intact
+    * while {@code tokenOf(0)} read it back as {@code "none"}: a clear that reported success and
+    * cleared nothing.
+    */
    private static final Map<String, Integer> LINK_TYPES = Map.of(
-      "none", 0,
+      "none", HyperlinkDialogService.NONE,
       "web", Hyperlink.WEB_LINK,
       "viewsheet", Hyperlink.VIEWSHEET_LINK,
       "message", Hyperlink.MESSAGE_LINK);
@@ -99,7 +114,7 @@ public class AssemblyHyperlinkService {
          target.colName(), target.axis(), target.text(), target.titleLink(),
          target.emptyPlotLink(), user);
 
-      return describe(assemblyName, model);
+      return describe(assemblyName, model, target);
    }
 
    /** One {@code sessions.mutate}, so one undo checkpoint. */
@@ -111,13 +126,21 @@ public class AssemblyHyperlinkService {
       String type = requireType(link);
       requireValueForType(type, link);
 
+      // Resolved here for the same reason, and it is the reason this ordering matters rather than
+      // being tidy. A viewsheet link is stored by asset ID; resolving inside the mutate meant an
+      // unresolvable target threw with the model already half-written, leaving the assembly
+      // carrying linkType=viewsheet with a null destination -- precisely the broken link the
+      // validation above exists to prevent, arrived at by another road.
+      String assetId = "viewsheet".equals(type)
+         ? resolveViewsheetTarget(link, sessionToken, user) : null;
+
       sessions.mutate(sessionToken, user, (rvs, runtimeId, dispatcher) -> {
          Region target = region == null ? Region.whole() : region;
          HyperlinkDialogModel model = hyperlinkService.getHyperlinkDialogModel(
             runtimeId, assemblyName, target.row(), target.col(), target.colName(),
             target.axis(), target.text(), target.titleLink(), target.emptyPlotLink(), user);
 
-         apply(model, type, link);
+         apply(model, type, link, assetId);
          hyperlinkService.setHyperlinkDialogModel(runtimeId, assemblyName, model, linkUri, user,
                                                  dispatcher);
       });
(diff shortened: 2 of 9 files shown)


larryliang-inetsoft (changes requested):
Reviewed the diff — the fixes for the 8 described defects look sound (LINK_TYPES/`NONE` sentinel mismatch, `assetLinkId` resolution moved before `sessions.mutate`, `colName` echo, `PropertyAliases`/`PropertyPath` writer-lookup and canonical-value fixes). Not merging because:

**`build` check is failing** (https://github.com/inetsoft-technology/stylebi/actions/runs/32469114549/job/96731976248). I can't pull the CI log content from here, but the PR description itself names the almost-certain cause: it states two existing tests in `AssemblyHyperlinkServiceTest` are known to be broken by this change and were "left alone deliberately" —

- `clearingALinkNeedsNoDestination` still asserts `linkType == 0`; after this PR a cleared link is `HyperlinkDialogService.NONE` (9), so this assertion should now fail.
- `harness()` for `setsAViewsheetLink` doesn't stub `getAssetRepository()`/`containsEntry`, which `resolveViewsheetTarget` now calls — this should NPE or otherwise fail.

Please update both before merge — a PR can't land with tests it already knows are broken, and that's very likely today's `build` failure. Also worth adding: despite the live verification described in the PR body, no new automated regression coverage was added for any of the 8 fixes (no test files are in the diff) — per this repo's convention (see the sibling `stylebi#4709`/`#4710`/`#4711` PRs), each fix should get a regression test, not just a live-verified manual repro, so it doesn't silently regress later.

No other blocking issues found in the diff itself.

lizhepu:
Thanks — acted on the coverage point, and three of the four items had already moved by the time the review landed. Details, since the difference matters for deciding whether to merge:

**The two named test breakages were fixed in `e05beba51`**, the `listLinkTargets` commit that landed after the review was written against `39ee641b8`:

- `clearingALinkNeedsNoDestination` asserts `HyperlinkDialogService.NONE` (`AssemblyHyperlinkServiceTest.java:96`).
- `harness()` stubs both `getAssetRepository()` and `containsEntry(any())` (lines 521 and 525).

**`build` is green on the current head.** Run `32616341663` on `e05beba51` succeeded, and its log shows `Tests run: 26, Failures: 0, Errors: 0` for `AssemblyHyperlinkServiceTest`. The failing run cited in the review (`32469114549`) predates that commit. The only red check now is `claude-review` itself, which failed with `is_error:true` before producing a review — the same infra error it hit three times on 2026-08-21 — not a code failure.

**The coverage point was correct, and more precisely than stated.** Test files *were* in the diff, but all 308 lines came from `e05beba51`; the four fix commits carrying the eight defects added none, and `PropertyAliasesTest`/`PropertyPathTest` were byte-identical to the base while `PropertyAliases.java` and `PropertyPath.java` changed by +71 and +107.

Now added: **25 tests, 385 lines**, across the three classes — `AssemblyHyperlinkServiceTest` 26 → 36, `PropertyAliasesTest` 21 → 30, `PropertyPathTest` 32 → 40. All 106 pass.

**Each one was verified to fail without its fix**, by reverting the sources and re-running, so these pin the defects rather than merely passing alongside them:

- Reverting `PropertyAliases.java` + `PropertyPath.java` → 9 failures. `writesThroughAnAccessorWhoseNameIsLowercaseAfterThePrefix` reproduces the original symptom exactly: `'xGridLineStyle' is not a writable property of AwkwardAccessors. Available: rTChartType, xGridLineStyle` — refused and listed as available in the same sentence.
- Reverting the four hyperlink changes → 7 failures, including `unknown(9)` for a never-linked assembly, a null `assetLinkId`, `mutate` being called despite an unresolvable path, and a null `colName`.

The remaining 8 are deliberate guards on boundaries the fixes drew rather than defect repros, so they pass either way: `facetGrid` and the real line-pane settings stay writable, `getRTChartType` still resolves through the rebuilt spelling (the strict-superset claim), and an unconstrained string keeps the caller's own spelling.

One correction to the PR body — its "Known: two tests need updating" section is now stale and describes work already done. Happy to edit it out if you'd rather the description match the branch before merge.

lizhepu:
## Review status — head `e5cb70d5d`

Reviewed the two test commits pushed on top of `e05beba51`. One real problem found and fixed. **No blocking issues remain.**

### Found and fixed: a test that asserted nothing

`theChartRefusalDoesNotReachAnotherAssemblyType` asserted `resolveForWrite(SHEET, "maxRows")`, which is byte-identical to the pre-existing `resolvesAnOrdinaryScalarForWrite` (`PropertyAliasesTest.java:174`). `maxRows` is not one of the seven refused flags, so the `chartWriteRefusal` branch never executed — the test's name claimed the refusal is scoped by assembly type while testing nothing of the sort.

Replaced with the actual claim: the same leaf that is refused on a chart resolves for write on a viewsheet.

```java
assertEquals("chartLinePaneModel.gridLineVisible",
             PropertyAliases.resolveForWrite(PropertyAliases.SHEET,
                                             "chartLinePaneModel.gridLineVisible"));
```

That reaches `resolve`'s raw-dotted-path escape hatch (`PropertyAliases.java:180`) and `chartWriteRefusal` returns null because the normalized type is not `chart`, so the test now exercises the type key it names.

### Checks that passed

- **Branch contents.** `e05beba51..HEAD` is two commits touching only three test files: 391 insertions, 0 deletions. `git diff e05beba51 HEAD -- core/src/main/` is empty, so the source reverts used to prove the tests fail without their fixes did not leak into the branch.
- **Mockito.** `verify(h.repository, never()).containsEntry(any())` does not conflict with the harness's `when(...containsEntry(any())).thenReturn(true)` — stubbing invocations are excluded from verification.
- **The 8 tests that pass both before and after the fixes are deliberate**, not oversights: `aTwoCapitalAccessorStillResolvesThroughTheRebuiltSpelling` guards the strict-superset property the writer-lookup fix claims, and `facetGridItselfStaysWritable` / `theRealLinePaneSettingsStayWritable` guard against the refusal over-reaching into properties that do work. The other 17 each fail without their fix.

### Summary

| | |
|---|---|
| Coverage added | 25 tests, 391 lines, test-only — no source changed |
| `AssemblyHyperlinkServiceTest` | 26 → 36 |
| `PropertyAliasesTest` | 21 → 30 |
| `PropertyPathTest` | 32 → 40 |
| Result | **106 tests, 0 failures** |
| Regression value | 17 tests verified to fail with their fix reverted; 8 are deliberate boundary guards |
| Issues found in review | 1, fixed in `e5cb70d5d` |

All eight defects described in the PR body now have automated coverage, which was the one outstanding item from the earlier review. Ready from my side.

fredaan (approved):

PR #4708, opened 2026-08-21
Report a bounded sample of an endpoint's rows from the request the table build already made

Nelson-cao:
Creating a tabular table sends a request and reports the structure of what came back. It cannot report what was IN it, and that is what a caller needs when one endpoint's required parameter takes its legal values from another endpoint's data — Asana's `Section Tasks` needs a section, and the sections are rows of `Project Sections`. Reading those ids costs a viewsheet today, which stands up a chart runtime to answer a question the build already had the data for.

`/ws/table` now returns `sampleRows` for a tabular table, when the caller asks for them, taken from the page the build already fetched.

## The sample

`JsonRowSampler` is the counterpart of `JsonShapeDistiller` and deliberately its opposite: the shape answers where the rows are and what the fields are called, the sample answers what is in them. So the two are taken at different points of the same page — the shape before `jsonPath` narrows it and before `doLookups` grafts lookup responses in, the sample after both, because a sample has to be of the rows the built table actually has. Both are first page only, and both cost no extra request.

Carrying values is the point and also the risk, so the sample is bounded four ways rather than by a row count alone: rows, total nodes, nesting depth, and string length. The last is the one a row count cannot approximate, since a single description or base64 blob outweighs every other field in the response.

- A row that does not fit the node budget is **dropped rather than half-copied** — a half row reads as a row whose remaining fields the response did not have.
- An over-long value is **replaced by `"<omitted: N chars>"`, never truncated in place** — a truncated id is a wrong value that looks entirely valid and will be sent straight back as a parameter.
- Anything left out sets `sampleRowsTruncated`, without which "this value is absent" and "this value was not sampled" are the same answer.

Leaf values go through `JsonTable.getJavaValue`, the same conversion the table uses. Without it the JSON-P and Jackson object models — both of which reach this runner — would serialize their wrapper objects instead of the values, and a caller could not use one as a parameter.

## Who decides what

`tabularSource.sampleRows` is the request field, and its default of 0 means no rows. Every path that does not set it — the composer dialog, every later render, every caller that wanted columns — runs the same request and pays for the same response as before. The rows are returned to whoever asked and are spent tokens in that answer, so the count is the caller's decision.

`rest.sample.rows` is the ceiling on that decision, default 20, and at 0 the switch that stops any response value leaving the connector. The deployment bounds how much customer data may ever travel; the caller bounds what one answer costs. A request above the ceiling is clamped rather than refused — the caller asked for a sample, and a smaller sample is still one.

## The clone boundary

The result slot mirrors `responseShape`, including the two halves that only work together: `clone()` clears it and `TabularHandler.execute` copies it back unconditionally. The reason applies harder here — a clone inheriting the field would report one execution's customer data as another's. It is likewise never written to worksheet XML.

The requested LIMIT is the mirror image, being an input rather than a result: `clone()` must NOT clear it, because the clone is the object that executes, and it is left out of the XML deliberately rather than incidentally — a saved query carrying it would sample again on every render and leave rows of customer data on a query nobody is reading them from.

## Tests

`JsonRowSamplerTest` (10 cases): raw nesting preserved, row cap and truncation, the whole-row boundary of the node budget, depth and long-string markers, non-Map elements, non-List input, recursive immutability, JSON-P value conversion.

Ran green: `inetsoft-tabular-util` 42, and `TabularTableAssemblyTest` + `WorksheetTableService*Test` in core, 92.

Design: `docs/superpowers/specs/2026-08-20-tabular-build-sample-rows-design.md` in stylebi-wiz (companion PR).

### DIFF
--- a/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/EndpointJsonQueryRunner.java
+++ b/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/EndpointJsonQueryRunner.java
@@ -26,7 +26,9 @@ import inetsoft.uql.rest.json.RestDataIteratorStrategyFactory;
 import inetsoft.uql.rest.json.RestJsonQuery;
 import inetsoft.uql.rest.json.lookup.*;
 import inetsoft.uql.rest.json.lookup.LookupService;
+import inetsoft.sree.SreeEnv;
 import inetsoft.uql.util.BaseJsonTable;
+import inetsoft.uql.util.JsonRowSampler;
 import inetsoft.uql.util.JsonShapeDistiller;
 import inetsoft.uql.util.LookupList;
 import inetsoft.uql.util.LookupMap;
@@ -59,6 +61,7 @@ public class EndpointJsonQueryRunner extends RestJsonQueryRunner {
          strategy.setTouchTimestamp(getTouchTimestamp());
 
          boolean shaped = false;
+         boolean sampled = false;
 
          while(hasNext(strategy, table)) {
             final Object json = strategy.next();
@@ -92,6 +95,21 @@ public class EndpointJsonQueryRunner extends RestJsonQueryRunner {
             doLookups(query, json);
 
             final Object selectedData = selectData(json, query.getValidJsonPath(), transformer);
+
+            // Sample the ROWS, and note that every "before" in the comment above is an "after" here.
+            // The shape answers where the rows are; the sample answers what is in them, so it has to
+            // be taken from what jsonPath actually selected and after doLookups has grafted the
+            // lookup fields in -- otherwise it would report values the built table does not have.
+            //
+            // Still the FIRST page only, and for a different reason than the shape: later pages are
+            // more of the same data, and this is a sample rather than a read. Also before
+            // loadStreamed, because that flattens nested objects into columns (JsonTable.walkRecord)
+            // and the response's own structure cannot be recovered from the result.
+            if(!sampled) {
+               sampled = true;
+               sampleRows(selectedData);
+            }
+
             table.loadStreamed(selectedData);
          }
       }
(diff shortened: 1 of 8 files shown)


jshobe-inetsoft (changes requested):
## Review summary

This PR adds an opt-in, bounded sample of a tabular endpoint's first-page rows to the `/ws/table` response, so a caller building one tabular table can read parameter values it needs for a *different* endpoint (e.g. `Section Tasks` needing a section id from `Project Sections`) without standing up a viewsheet just to read four ids.

The implementation follows the exact clone/copy-back/no-XML pattern already established by the sibling `responseShape` feature (`TabularQuery`, `TabularHandler.execute`, `EndpointJsonQueryRunner.shapeResponse`), which makes it easy to verify by comparison. I traced the full path end to end:

- `WorksheetTableService.buildTabularTable` sets `query.setSampleRowLimit(...)` on the query *before* `loadColumnSelection` triggers the one request the build already makes, correctly clamping a negative value to 0.
- `EndpointJsonQueryRunner.runStream` samples exactly once, on the first page, after `jsonPath` selection and after `doLookups`, matching the doc comment's stated reasoning (the sample must reflect what the built table actually has, not the raw envelope).
- `JsonRowSampler` bounds by rows, total nodes, depth, and string length; drops whole rows rather than half-copying them; replaces over-long strings with a non-data marker instead of truncating in place (avoiding a truncated id being sent back as a real parameter value); and returns a recursively unmodifiable structure. The 10 unit tests cover each of these boundaries plus JSON‑P value conversion, non-Map/non-List input, and immutability — good coverage of the algorithm itself.
- `TabularQuery.clone()` correctly clears `sampleRows`/`sampleRowsTruncated` (mirroring `responseShape`) but deliberately leaves `sampleRowLimit` untouched, since that field is the caller's *request* and the clone is what actually executes — getting this backwards in either direction would silently break the feature (either the sample never fires, or a stale sample from an old run gets reported as current).
- `TabularHandler.execute` copies the clone's sample back onto the original query unconditionally, same as `responseShape`, so a failed/empty run correctly clears a stale sample rather than leaving a previous run's data in place.
- Neither `sampleRows` nor `sampleRowLimit` is written to the worksheet XML — correct, since a saved query carrying either would re-sample customer data on every later render or replay a stale sample.
- The `rest.sample.rows` deployment ceiling is read with `Integer.parseInt(SreeEnv.getProperty(..., default))`, matching the established pattern used elsewhere in the codebase (`swapper.count`, `data.temp.file.threshold`, etc.), and a non-numeric value is swallowed by the existing outer `catch(Exception)` in `sampleRows()`, degrading to "no sample" rather than failing the request — appropriate for a by-product.
- `WorksheetTable.TabularSource.sampleRows` / `WorksheetTableResponse.sampleRows(+Truncated)` are purely additive fields with `@JsonIgnoreProperties(ignoreUnknown = true)` already in place on the containing classes, so this is backward compatible for existing callers of `/ws/table`.

No security concerns: the sample is bounded at the source on four independent axes, has a deployment-wide kill switch (`rest.sample.rows=0`), is opt-in per request, and is never persisted to XML — if anything this is a more conservative data-exposure story than a lot of existing code that reads full responses.

No regressions found in the existing `responseShape` path — the only change to it is the identical, additive copy-back call added to `TabularHandler.execute`. No backward-incompatible changes; no test appears to be broken by this diff (`build` check is green).

### Suggestion (non-blocking)

`TabularHandlerTest` has a dedicated pair of tests for the `responseShape` clone invariant (`cloneDoesNotInheritTheOriginalsResponseShape`, `aCloneRecordsItsOwnShapeIndependently`) precisely because a shallow `clone()` regression there fails silently. The new `sampleRows`/`sampleRowsTruncated`/`sampleRowLimit` fields have the same kind of subtle, asymmetric clone behavior (two fields cleared, one deliberately *not* cleared) and currently have no equivalent direct test — `JsonRowSamplerTest` covers the sampling algorithm well, but nothing pins `TabularQuery.clone()`'s behavior for these three fields the way it's pinned for `responseShape`. Given how easy this would be to break silently in a future refactor, a small addition mirroring the existing pair (plus one asserting `sampleRowLimit` *does* survive `clone()`) would close the gap.

Minor doc nit: `JsonRowSampler.Result#getRows()`'s javadoc says the list is "held by a query, shared with every clone of that query" — `TabularQuery.clone()` actually nulls this field out for the clone, so it's *not* shared with clones; it's shared between the original query and whatever reads it off that same instance (e.g. the response object). Worth a small wording fix, not a functional issue.

### Why this review requests changes rather than approving

The code changes themselves look correct and complete, and no blocking issue was found in them. However, per this review's approval criteria, all of the following must hold, and one does not:

- ✅ No blocking issues in the code itself.
- ❌ **The `claude-review` check is currently failing** (`Claude execution failed: result is_error:true` — an execution error in the automated review Action itself, not a reported code-quality finding). `build`, `Semgrep Scan`, `Semgrep OSS`, and `license/cla` are all green.
- ✅ No merge conflicts (`MERGEABLE`).

Since a required/reported check is red, this is filed as "request changes" rather than an approval. Please re-run (or investigate) the `claude-review` job; once it's green (or the failure is confirmed to be unrelated CI flakiness), this is otherwise in good shape to merge.

jshobe-inetsoft (approved):

PR #4713, opened 2026-08-21
Bug #76204, default the cloud-runner schedule timeout to 10 minutes instead of waiting unbounded

jkidacki-inetsoft:
## Summary

`ScheduleTaskCloudJob` (the Quartz job used to run scheduled tasks via an ephemeral cloud container) fell back to an unbounded `latch.await()` whenever `schedule.task.timeout` was <= 0, matching the "no timeout" meaning that value has on the other three readers of the property. On a cloud-runner deployment this means a stuck/never-completing task can hold a `@DisallowConcurrentExecution` Quartz job slot and keep billed cloud compute running forever — the exact cost-overrun risk bug #68838 was opened to prevent. This reopens/supersedes the previously merged fix for this issue (#4694), which chose that unbounded-wait behavior.

## Root Cause

`ScheduleTaskCloudJob.execute()` guarded the wait with `timeout > 0`, waiting unbounded in the `else` branch. That's fine for `ScheduleTask.doRun`/`MVAction.createMV` (in-process work), but not for a cloud runner path that consumes billed, ephemeral compute.

## Fix

A non-positive `schedule.task.timeout` now resolves to a hardcoded 10-minute (600000ms) default at construction time, and `execute()` always performs a bounded `latch.await(timeout, TimeUnit.MILLISECONDS)`. Untouched installs (default `schedule.task.timeout=600000`) are unaffected. `interrupt()`/cancellation behavior is unchanged. The unguarded `Long.parseLong` in the constructor (#76177) and the unbounded fallback in `ScheduleTask.doRun`/`MVAction.createMV` are intentionally left untouched — out of scope for this defect.

## Test Plan

- [x] `code-reviewer` subagent review: no issues found
- [x] `./mvnw install -pl core -am -DskipTests` — BUILD SUCCESS
- [x] `./mvnw test -pl core -Dtest=ScheduleTaskCloudJobTest,ScheduleTaskTest,MVActionTest` — 28/28 passing
- [x] Manually verified the constructor/`execute()` logic: non-positive timeout → 600000ms default; positive timeout → unchanged; `interrupt()` still short-circuits the wait via `latch.countDown()`

Fixes Redmine #76204.

### DIFF
--- a/core/src/main/java/inetsoft/sree/schedule/ScheduleTask.java
+++ b/core/src/main/java/inetsoft/sree/schedule/ScheduleTask.java
@@ -62,10 +62,12 @@ public class ScheduleTask implements Serializable, Cloneable, XMLSerializable {
     * value logs a warning and falls back to DEFAULT_TASK_TIMEOUT.
     *
     * <p>The value is returned as configured; no meaning is imposed on zero or a negative
-    * number here, because the call sites do not agree on one. The three that wait on a
-    * future or a latch (doRun below, MVAction.createMV, ScheduleTaskCloudJob.execute) treat
-    * 0 or less as "wait without a timeout", whereas Scheduler.runTask uses the value only as
-    * a staleness window and does not special-case it.
+    * number here, because the call sites do not agree on one. The two that wait on a
+    * future or a latch (doRun below, MVAction.createMV) treat 0 or less as "wait without a
+    * timeout", whereas Scheduler.runTask uses the value only as a staleness window and does
+    * not special-case it. ScheduleTaskCloudJob.execute instead clamps a non-positive value to
+    * DEFAULT_TASK_TIMEOUT and always waits with a bound, since it consumes billed cloud
+    * compute and must never wait unbounded.
     *
     * @return the configured task timeout in milliseconds.
     */
--- a/core/src/main/java/inetsoft/sree/schedule/cloudrunner/ScheduleTaskCloudJob.java
+++ b/core/src/main/java/inetsoft/sree/schedule/cloudrunner/ScheduleTaskCloudJob.java
@@ -43,7 +43,8 @@ import java.util.concurrent.locks.Lock;
 public class ScheduleTaskCloudJob implements InterruptableJob {
    public ScheduleTaskCloudJob() {
       cluster = Cluster.getInstance();
-      timeout = Long.parseLong(SreeEnv.getProperty("schedule.task.timeout"));
+      long parsedTimeout = Long.parseLong(SreeEnv.getProperty("schedule.task.timeout"));
+      timeout = parsedTimeout > 0 ? parsedTimeout : DEFAULT_TIMEOUT;
    }
 
    @Override
@@ -95,17 +96,12 @@ public class ScheduleTaskCloudJob implements InterruptableJob {
             }
          }
 
-         if(timeout > 0) {
-            if(!latch.await(timeout, TimeUnit.MILLISECONDS)) {
-               if(job != null) {
-                  job.stop();
-               }
-
-               throw new JobExecutionException("Scheduled task '" + taskName + "' timed out");
+         if(!latch.await(timeout, TimeUnit.MILLISECONDS)) {
+            if(job != null) {
+               job.stop();
             }
-         }
-         else {
-            latch.await();
+
+            throw new JobExecutionException("Scheduled task '" + taskName + "' timed out");
          }
 
          if(result != null && !result.isSuccess()) {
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Critical: overlapping, conflicting change with open PR #4727

This PR and **PR #4727** ("Bug #76177, parse schedule.task.timeout consistently," currently open against `main`) both rewrite the exact same constructor and `execute()` method of `ScheduleTaskCloudJob.java`, with **incompatible design decisions** for the same edge case:

- **This PR (#4713 / Bug #76204):** a non-positive `schedule.task.timeout` resolves to a hardcoded 10-minute default, and `execute()` is made *unconditionally* bounded — the `timeout > 0 ? bounded : unbounded` branch is removed entirely. Rationale: an unbounded wait on this specific job lets a stuck task hold a Quartz slot and bill cloud compute indefinitely.
- **PR #4727 (Bug #76177):** moves the timeout read out of the constructor into `execute()` via a new shared `ScheduleTask.getTaskTimeout()` accessor (fixing the *unguarded* `NumberFormatException`), but its own PR description explicitly states: *"`ScheduleTaskCloudJob.execute` guard[s] `if(timeout > 0)` and fall[s] through to an unbounded wait... [this is] deliberately returned as configured rather than normalized... normalizing would change existing runtime behavior, which is out of scope here."* — i.e., #4727 explicitly **preserves** the unbounded-wait behavior for this exact class that this PR is trying to eliminate.

I confirmed `main`'s current `ScheduleTaskCloudJob.java` still has the original, unfixed code (neither PR has merged yet), so this conflict is real and live, not hypothetical. Whichever of these two PRs merges second will either hit a textual merge conflict on this method, or — worse, if it merges cleanly — silently overwrite the other's intended fix without anyone deciding which behavior is actually correct. This needs to be resolved between the two PR authors/reviewers before either merges: most likely, #4727's `ScheduleTask.getTaskTimeout()` accessor should be adopted for the parsing/NFE-safety half of the fix, with this PR's cloud-specific "non-positive → 10-minute cap, always bounded" behavior layered on top for `ScheduleTaskCloudJob` specifically (since the other two callers `ScheduleTask.doRun`/`MVAction.createMV` don't share the same cost-overrun exposure).

### Correctness (in isolation)

Taken on its own, the diff is small and does what it says: `parsedTimeout > 0 ? parsedTimeout : DEFAULT_TIMEOUT` correctly clamps non-positive values, `execute()`'s `latch.await(timeout, TimeUnit.MILLISECONDS)` is now always bounded, and `interrupt()` (unchanged, not in the diff) still short-circuits via `latch.countDown()` regardless of the wait path. The PR explicitly and correctly scopes out the still-unguarded `Long.parseLong` NFE risk (#76177) as a separate, already-filed defect — reasonable in isolation, though see above: that scoping decision is exactly where this PR and #4727 collide, since #4727 is the PR that closes that gap and does so by touching the same lines with different assumptions.

### Tests

**No new automated test was added.** The PR's file list shows only `ScheduleTaskCloudJob.java` changed — no update to `ScheduleTaskCloudJobTest.java`. The test plan cites 28/28 *existing* tests passing (`ScheduleTaskCloudJobTest`, `ScheduleTaskTest`, `MVActionTest`) and states the constructor/`execute()` logic was "manually verified" by reading it, not by an automated assertion. Given the PR's own framing — this "reopens/supersedes" a previously-merged fix for the same issue (#4694), implying the desired behavior here has already flip-flopped once — a regression test pinning "non-positive timeout → `DEFAULT_TIMEOUT` (600000), always bounded" would materially reduce the risk of this behavior being silently reverted again (including by #4727, per the conflict above). I'd treat this as a real gap worth asking for before merge, not just a nice-to-have.

- `DEFAULT_TIMEOUT = 600000L` matches the value used elsewhere (`defaults.properties`/`ScheduleTask.DEFAULT_TASK_TIMEOUT` in the related #4727), so at least the numeric constant is consistent across both in-flight PRs even though the surrounding logic differs.
- No security concerns.

### Blocking issues for merge

1. **Conflicting overlapping PR.** See above — #4727 makes an incompatible change to the same method; this needs coordination before either merges.
2. **Failing check.** The `claude-review` check reports `FAILURE`.
3. **Missing regression test** for the new clamping/always-bounded behavior.
4. **Unconfirmed mergeability.** `mergeable`/`mergeStateStatus` report `UNKNOWN` rather than a confirmed conflict-free state.

## Recommendation

Requesting changes. The isolated code change is simple and does what it claims, but this PR cannot safely merge independently of #4727 without the team deciding which behavior wins for `ScheduleTaskCloudJob`'s non-positive-timeout case — please coordinate with that PR's author first. Also please add a regression test for the new behavior, and resolve the failing `claude-review` check.

jshobe-inetsoft (approved):

PR #4714, opened 2026-08-21
Bug #76175: store the schedule task view as a per-user preference

ngovus-inetsoft:
## Problem

The schedule task list/folder view toggle in Enterprise Manager and the portal wrote `schedule.show.tasks.as.list` and stopped there. Filed for review rather than with a proposed fix, because two things were wrong and they did not have the same answer.

**The write never reached storage.** `PropertiesEngine.setProperty` updates the in-memory properties and records the name in `changedProps`; `save()` is what drains that set. Neither controller called `SreeEnv.save()`, unlike the ~25 admin settings services that all end with it. So the toggle survived a restart only if an administrator happened to visit some unrelated settings page afterwards, which flushed it as a side effect. Worse, `save()` *clears* `changedProps`, so an unrelated page saving first discarded the pending toggle outright. Persistence is also what drives `KeyValueStorage` listener propagation, so the value never reached other cluster nodes.

**The state was stored at the wrong granularity.** One global property backed a per-user toolbar gesture, so any user's toggle changed the view for every other user, across both surfaces. The writes used the two-argument `setProperty`, which writes an unprefixed global key, while reads go through the org-scoped path (`useAvailableOrgProperty`) and fall back to that same key — so the value leaked **across organizations** too, not only across users.

## Approach

The view is a per-user preference. `UserEnv` is the established mechanism and is what `PreferencesDialogController` already uses for `historyBarEnable`, so both controllers store the toggle there:

| Concern | Where it lives |
| --- | --- |
| Per-user state, portal | `UserEnv` key `portal.schedule.showTasksAsList` |
| Per-user state, EM | `UserEnv` key `em.schedule.showTasksAsList` |
| Installation default | `SreeEnv` property `schedule.show.tasks.as.list` (read-only) |

`UserEnv.setProperty` persists on write and rolls the in-memory value back if the write throws, so **both defects are fixed at once** and no `SreeEnv.save()` is needed — these controllers no longer write `SreeEnv` at all. Separate keys per surface stop a portal toggle from rearranging the EM task screen; the two components already called separate endpoints.

The endpoint contract is deliberately unchanged — same URLs, same parameter, same boolean response — so no frontend service or test mock needed updating.

### A trap worth flagging to reviewers

Both reads now use the varargs-free `getBooleanProperty(name)` overload. The previous `getBooleanProperty(name, "true")` *reads* as "default to true", but the second parameter is `String... trueValues`, so it actually means "true only when the value equals `'true'`" and returns **false** for an unset property. `schedule.show.tasks.as.list` is declared nowhere in the repo, so the effective default was already folder view. The call is now written so it cannot be misread, and a test pins the overload so the trap is not reintroduced. (#4692 documents the same `getBooleanProperty` semantics across thirteen other properties.)

### Anonymous users

They get a persisted preference only when `anonymous.userdata.save` is enabled — exactly what that property governs. The controllers delegate that decision to `UserEnv` rather than writing installation-wide configuration on a guest's behalf; these endpoints are gated on schedule access, not admin rights.

## Client-side

Both toggle handlers set local state before the PUT with no error callback, so a rejected request left the view showing a state that was never stored. Both now revert — and only when their own optimistic value is still current, so a stale failure cannot clobber a newer saved state.

The portal's initial GET had no error branch at all, leaving neither the tree nor the task list loaded on failure; EM's showed a dialog but never called `loadTasks()`, stranding an empty list. Both now fall back to the default view and still load. EM's error text also had a precedence bug (`"…" + error.error ? … : …`) that made the conditional always truthy, appending a literal `undefined` for a non-JSON body.

## Testing

- **New JUnit tests, 15 passing** (8 portal, 7 EM). The core regression guard is that two principals do not see each other's value; also covered are a stored `false` not being mistaken for "never chose", EM not reading the portal key, the varargs-free overload being used, and the controller never writing the global property.
- **Frontend:** all 24 schedule TL suites pass (schedule-task-list 53 + 20, including a new revert-on-PUT-failure test); full EM TL suite green (70 files); ESLint clean.
- Unrelated pre-existing failures in the portal TL run (viewsheet/composer/wizard, `VSTable is not a constructor`) drifted 31 → 26 → 23 files across runs with this code unchanged.

## Not addressed

`UserEnv.setProperty` catches persistence failures internally and returns normally, so the PUT answers 200 even when nothing was saved. Surfacing that would mean changing `UserEnv`'s contract and every caller, well outside this fix. The client revert still earns its place for transport and auth failures, and its comment claims only that. Worth a separate issue if save failures should propagate.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/schedule/EMScheduleTaskShowType.java
+++ b/core/src/main/java/inetsoft/web/admin/schedule/EMScheduleTaskShowType.java
@@ -18,11 +18,14 @@
 package inetsoft.web.admin.schedule;
 
 import inetsoft.sree.SreeEnv;
+import inetsoft.sree.UserEnv;
 import inetsoft.sree.security.ResourceAction;
 import inetsoft.sree.security.ResourceType;
 import inetsoft.web.security.RequiredPermission;
 import inetsoft.web.security.Secured;
 import org.springframework.web.bind.annotation.*;
+
+import java.security.Principal;
 
 @RestController
 public class EMScheduleTaskShowType {
@@ -34,8 +37,8 @@ public class EMScheduleTaskShowType {
       )
    )
    @GetMapping("/api/em/schedule/change-show-type")
-   public boolean getScheduleTaskShowType() {
-      return SreeEnv.getBooleanProperty("schedule.show.tasks.as.list", "true");
+   public boolean getScheduleTaskShowType(Principal principal) {
+      return getShowTasksAsList(principal);
    }
 
    @Secured(
@@ -46,7 +49,37 @@ public class EMScheduleTaskShowType {
       )
    )
    @PutMapping("/api/em/schedule/change-show-type")
-   public void setConfiguration(@RequestParam("showTasksAsList") String showTasksAsList) {
-      SreeEnv.setProperty("schedule.show.tasks.as.list", showTasksAsList);
+   public void setConfiguration(@RequestParam("showTasksAsList") String showTasksAsList,
+                                Principal principal)
+   {
+      setShowTasksAsList(principal, showTasksAsList);
    }
+
+   /**
+    * Gets the list/folder view preference of the given user. The view is a per-user
+    * preference; the property is only the installation-wide default used until the
+    * user toggles the view themselves. That default is folder view, since
+    * getBooleanProperty() yields false for a property that is not set.
+    */
+   private boolean getShowTasksAsList(Principal principal) {
+      Object userSetting = UserEnv.getProperty(principal, SHOW_TYPE_USER_PROPERTY, null);
+
+      return userSetting != null ? Boolean.parseBoolean(String.valueOf(userSetting))
+         : SreeEnv.getBooleanProperty(SHOW_TYPE_PROPERTY);
+   }
+
+   /**
+    * Saves the list/folder view preference of the given user. Anonymous users only get a
+    * persisted preference if anonymous.userdata.save is enabled; otherwise UserEnv
+    * discards the write and the installation-wide default keeps applying.
+    */
+   private void setShowTasksAsList(Principal principal, String showTasksAsList) {
+      // normalize so that only "true"/"false" is stored, and so that an empty value is
+      // never passed to UserEnv.setProperty(), which treats it as a removal
+      UserEnv.setProperty(principal, SHOW_TYPE_USER_PROPERTY,
+                          Boolean.parseBoolean(showTasksAsList) + "");
+   }
+
+   private static final String SHOW_TYPE_PROPERTY = "schedule.show.tasks.as.list";
+   private static final String SHOW_TYPE_USER_PROPERTY = "em.schedule.showTasksAsList";
 }
--- a/core/src/main/java/inetsoft/web/portal/controller/ScheduleTaskShowType.java
+++ b/core/src/main/java/inetsoft/web/portal/controller/ScheduleTaskShowType.java
@@ -18,11 +18,14 @@
 package inetsoft.web.portal.controller;
 
 import inetsoft.sree.SreeEnv;
+import inetsoft.sree.UserEnv;
 import inetsoft.sree.security.ResourceAction;
 import inetsoft.sree.security.ResourceType;
 import inetsoft.web.security.RequiredPermission;
 import inetsoft.web.security.Secured;
 import org.springframework.web.bind.annotation.*;
+
+import java.security.Principal;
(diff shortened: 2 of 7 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Correctness — backend

I independently verified every API-level claim against the actual source rather than trusting the PR narrative:

- `SreeEnv.getBooleanProperty(String name, String... trueValues)` really does take `trueValues` as varargs, and `getBooleanPropertyValue()`'s `trueValues.length == 0 → "true".equals(value)` branch confirms `getBooleanProperty(name, "true")` means "true only if the value equals the literal string `'true'`", returning **false** for an unset property — the opposite of what that call reads as. The fix's `SreeEnv.getBooleanProperty(SHOW_TYPE_PROPERTY)` (no varargs) hits the same `length == 0` branch, so the observable default behavior for this specific (never-declared) property is unchanged, but the code now says what it means.
- `UserEnv.getProperty(Principal, String, Object def)` and `UserEnv.setProperty(Principal, String, Object)` signatures match the new usage exactly.
- `UserEnv.setProperty()` really does treat `val == null || val.equals("")` as a removal (confirms the `put_normalizesTheStoredValue` test's premise) and really does roll back the in-memory value in a `catch` block if the underlying `save()` throws — so the "both defects fixed at once, no `SreeEnv.save()` needed" claim is accurate, not just asserted.
- `anonymous.userdata.save` gating for anonymous users is real (`UserEnv.java:633`), confirming the controller correctly delegates that policy decision to `UserEnv` rather than reimplementing it.

The two controllers are structurally identical (separate `SHOW_TYPE_USER_PROPERTY` constants: `em.schedule.showTasksAsList` vs. `portal.schedule.showTasksAsList`), which correctly prevents one surface's toggle from affecting the other, as intended.

### Correctness — frontend

- The EM `"Failed to load tasks: " + error.error ? error.error.message : ""` bug is real: due to `+` binding tighter than the ternary's condition-position but the ternary's `?:` having lower precedence than `+`, this parses as `("Failed to load tasks: " + error.error) ? error.error.message : ""`, which is truthy whenever `error.error` exists (almost always), so it always evaluated to `error.error.message` — appending the literal string `"undefined"` when `error.error.message` doesn't exist on a non-JSON error body. The fix's `(error.error?.message || "")` is correctly parenthesized and null-safe.
- Both `changeShowType`/`showAllTasks` revert handlers correctly guard against a stale failure clobbering a newer optimistic value (`if(this.showTasksAsList === value) { revert }`), which is the right defense against a slow failed request racing a subsequent fast toggle.
- The portal's `applyShowType()` extraction is a clean, correct refactor: the field default (`showTasksAsList: boolean = false`) means the new GET-error fallback naturally lands on folder view, matching the PR's stated "fall back to the default view and still load" behavior — verified by checking the field's actual default rather than assuming it.
- The new TL spec (`reverts showTasksAsList when the PUT fails`) exercises the real revert path via MSW-intercepted HTTP, not just a unit-level assertion on the handler in isolation.

### Tests

Strong, well-targeted coverage on both sides — 15 new JUnit tests covering exactly the properties that matter (no cross-user leakage, a stored `false` not mistaken for "unset", EM not reading the portal's key, the anonymous/global-property-untouched contract, and the varargs-free overload being pinned so the trap can't be silently reintroduced), plus a new frontend regression test for the PUT-failure revert path. The PR is transparent that the observed pre-existing TL failures in unrelated portal suites (viewsheet/composer/wizard, `VSTable is not a constructor`) fluctuate independent of this change — a reasonable, checkable claim given the file counts cited.

### Scope and transparency

The "Not addressed" section is a good call: `UserEnv.setProperty`'s internal swallowing of persistence failures means the PUT can still return 200 when nothing was actually saved, and the PR correctly declines to change `UserEnv`'s contract (and every caller) to fix that here, scoping it as a separate potential issue rather than silently leaving it unmentioned.

- No backward-incompatibility concerns for the REST contract (same URLs/params/response shape, as stated). Existing installs that already have `schedule.show.tasks.as.list` set will see it become a shared default rather than a global toggle — an intentional, well-justified behavior change directly addressing the reported bug.
- No security concerns.

### Blocking issues for merge

## Recommendation

Requesting changes — solely because of the failing `claude-review` check. This is a thorough, correctly-implemented, and well-tested fix; I verified the key technical claims (varargs semantics, `UserEnv` persistence/rollback behavior, the operator-precedence bug, anonymous-user gating) directly against the source and found them all accurate.

jshobe-inetsoft (approved):

PR #4715, opened 2026-08-21
Bug #76085, fix chart image request storm and always-loading chart

ngovus-inetsoft:
## Problem

Reported repro: import `ST_interval1`, open the viewsheet in composer, click the edit icon to reach the binding pane, then *Show Enlarged* → *Show Actual Size*. The chart sometimes stays in the loading state forever, a burst of `getAssemblyImage` XHRs is issued, and the browser UI goes sluggish.

The attached HAR looked like duplicate requests, but it isn't: the 25 entries are the 9 distinct chart areas (`plot_area`, both x axes, both y axes, both titles, legend title/content) re-requested across **three full re-render cycles** (`genTime` 194876 → 205009 → 205229). The last two are 220 ms apart at the *same* requested size (444×495) but a *different* layout (legend width 83→50, plot width 327→304) — the server was re-laying-out the chart repeatedly. The truncated DevTools "Name" column only showed the shared `true?_csrf=…` tail, which made them look identical.

## Root cause

`VSChartAreasService` dropped the max-mode size when it regenerated an incomplete `VGraphPair`:

```java
pair = box.getVGraphPair(absoluteName, true, maxSize);   // correct size
...
if(pair != null && !pair.isCompleted()) {
   box.clearGraph(absoluteName);
   pair = box.getVGraphPair(absoluteName);               // maxSize lost -> maxsize = null
}
```

The tile endpoint uses the other size — `AssemblyImageService` prefers `info.getMaxSize()` and only falls back to the URL's `maxWidth/maxHeight` when it is null. `ViewsheetSandbox.getVGraphPair` cancels and rebuilds any pair whose `getContentSize()` differs from the requested one, so the two paths kept cancelling each other's graph. A cancelled/incomplete pair makes the image endpoint return `Retry-After: 1` with an empty body and the areas response `completed=false`, both of which the client re-requests — hence the burst, the stuck spinner, and the main-thread load. The existing comment in `AssemblyImageService` describes exactly this failure mode.

## Changes

1. **`VSChartAreasService`** — pass the in-scope `maxSize` when regenerating the pair, so the areas path and the image path agree on the content size. This is the actual fix.

2. **`ChartImageDirective`** — the `Retry-After` loop had no limit, no backoff, and emitted neither `onLoaded` nor `onError`, so nothing ever cleared the tile's loading state. Now bounded by a 5-minute wall-clock budget (deliberately generous: a large graph can legitimately take a while, and the facet/legend/title templates bind none of the load outputs, so giving up early would leave those tiles permanently blank), with escalating delay that never polls faster than the server asked, a guard so a late response for a superseded address can't consume the current address's budget, and `retryTimer` cleared when the address changes.

3. **`ChartObjectAreaBase.getSrc()`** — measured the DOM via `GuiTool.getChartMaxModeSize()` for every tile on every change-detection pass, so a scrollbar appearing shifted the URL by a few pixels and reloaded every tile, and each call forced a layout. The measurement is now cached per `genTime` (invalidated on window resize and on a container change). Safe because `maxMode` is only true when `info.getMaxSize() != null`, and the server ignores the URL's `maxWidth/maxHeight` in exactly that case.

## Testing

- `tsc --noEmit` on the portal app: clean.
- `mvnw compile -pl core`: clean.
- The 17 TL spec files covering `graph/objects`, `vs-binding-pane` and `vs-object-container`: 590 tests pass.
- Manual verification of the reported repro is pending with the reporter's data.

## Notes for reviewers

- `RegionPropertyDialogService.java:336-338` has the identical drop-the-`maxSize` pattern in the chart region properties dialog. Left alone here to keep the fix scoped; worth a follow-up.
- Pre-existing and not addressed: the second `getVGraphPair` call runs outside `box.lockRead()` and `maxSize` is captured before the lock, so a concurrent max-mode resize can still regenerate at a stale size. This change narrows that race considerably but does not close it.
- Also pre-existing, and the likely reason the burst *feels* like a freeze rather than just slow: `Tool.clone(this.model)` (lodash `cloneDeep` of the whole chart model including every region) on every `SetChartAreasCommand`, and the per-tile synchronous `canvas.toDataURL()` PNG encode in `captureChartSnapshot()`. Deliberately out of scope.

### DIFF
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/chart/VSChartAreasService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/chart/VSChartAreasService.java
@@ -131,7 +131,11 @@ public class VSChartAreasService {
 
          if(pair != null && !pair.isCompleted()) {
             box.clearGraph(absoluteName);
-            pair = box.getVGraphPair(absoluteName);
+            // pass maxSize so the regenerated pair is laid out at the same content size
+            // used by the chart image requests (AssemblyImageService uses info.getMaxSize()).
+            // otherwise the two paths keep cancelling each other's graph, which makes the
+            // image requests return Retry-After forever and the chart always loading.
+            pair = box.getVGraphPair(absoluteName, true, maxSize);
          }
 
          if(pair != null) {
--- a/web/projects/portal/src/app/graph/objects/chart-image.directive.ts
+++ b/web/projects/portal/src/app/graph/objects/chart-image.directive.ts
@@ -30,10 +30,17 @@ export class ChartImageDirective implements OnDestroy {
    set chartImage(value: string | SafeValue) {
       if(value !== this._chartImage) {
          this._chartImage = value;
+         this.resetRetry();
 
          if(this._loadTimer !== null) {
             clearTimeout(this._loadTimer);
             this._loadTimer = null;
+         }
+
+         // a retry armed for the previous address must not fire after the address changed.
+         if(this.retryTimer !== null) {
+            clearTimeout(this.retryTimer);
+            this.retryTimer = null;
          }
 
          if(!!value) {
@@ -56,6 +63,14 @@ export class ChartImageDirective implements OnDestroy {
    private currentBlobUrl: string = null;
    private loadSubscription: Subscription = null;
    private retryTimer: ReturnType<typeof setTimeout> | null = null;
+   private retryCount = 0;
+   private retryStart = 0;
+   // the server answers with Retry-After while the chart graph is still being generated.
+   // give up only after a generous wall-clock budget: a large/slow graph (big dataset, mv
+   // build, web map) may legitimately take a long time, and the facet/legend/title areas
+   // don't listen to onError, so giving up early leaves those tiles permanently blank.
+   private readonly MAX_RETRY_TIME = 300000;
+   private readonly MAX_RETRY_INTERVAL = 5000;
 
    constructor(private element: ElementRef, private http: HttpClient, private renderer: Renderer2) {
    }
@@ -80,6 +95,11 @@ export class ChartImageDirective implements OnDestroy {
       }
    }
 
+   private resetRetry(): void {
+      this.retryCount = 0;
+      this.retryStart = 0;
+   }
+
    private loadImage(reloading = false): void {
       if(this.retryTimer != null) {
          clearTimeout(this.retryTimer);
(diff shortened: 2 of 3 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Correctness — root cause (verified against actual source)

I independently traced every claim in the root-cause chain against the real code, not just the PR's narrative, and it all holds up:

- `AssemblyImageService` (line ~624/634): `Dimension maxSize = info.getMaxSize(); ... if(maxSize == null && maxWidth > 0 && maxHeight > 0) { maxSize = new Dimension(...); }` — confirmed it prefers `info.getMaxSize()` and only falls back to the URL's `maxWidth`/`maxHeight` when null, exactly as described.
- `ViewsheetSandbox.getVGraphPair()` (~line 6763): `if(!ignoreSize && (nsize == null || !nsize.equals(osize)) || ...) { ...cancel and rebuild... }` — confirmed the cancel/rebuild-on-content-size-mismatch logic is real, using `VSUtil.getContentSize(chart, maxsize)` against the pair's stored `getContentSize()`.
- `VSChartAreasService.refreshChartAreasModel0()`: `Dimension maxSize = info.getMaxSize();` is declared once, used for the *first* `getVGraphPair(absoluteName, true, maxSize)` call — and the *second* regeneration call (on an incomplete pair) previously called the no-maxsize overload (`getVGraphPair(absoluteName)` ≡ `getVGraphPair(name, true, null)`), silently dropping the same `maxSize` variable that was already in scope. That mismatch is exactly what would trigger `getVGraphPair`'s own cancel-and-rebuild logic the next time the image endpoint requested the pair with the real max size — a self-sustaining tug-of-war between the two code paths, matching the reported symptom (repeated re-layout, `Retry-After` responses, stuck spinner) precisely. The one-line fix (passing the in-scope `maxSize` to the second call) closes exactly this gap.

This is a genuinely well-diagnosed bug, and the fix is minimal and precisely targeted at the mechanism, not the symptom.

### Correctness — frontend hardening

**`ChartImageDirective`:** the retry backoff math is correct. I traced it by hand: `delay = Math.max(interval, Math.min(interval * retryCount, MAX_RETRY_INTERVAL))` escalates the delay by `retryCount` up to the 5s cap, but the outer `Math.max(interval, ...)` ensures the delay never drops below whatever the server's `Retry-After` header actually asked for — even if that's larger than the 5s cap (e.g., an 8s server-requested interval is honored in full, not clamped down to 5s). The `isNaN(seconds) ? 1000 : ...` guard prevents a malformed/missing header from becoming a busy-loop. The `requestedImage != this.chartImage` early-return inside the `Retry-After` branch is a real, correct fix for a genuine (if minor) pre-existing bug: previously a late `Retry-After` response for an address the component had already moved on from would still arm a retry timer, wasting a slot and risking a spurious reload race.

**`ChartObjectAreaBase.getSrc()`:** the `genTime`/`container` cache-invalidation triple-check is sound. `genTime` is a real `@Input()` already used elsewhere in the same URL-construction method, so gating the cached measurement on it (plus a hard invalidation on `window resize`) correctly re-measures exactly when the server has regenerated the chart (when the measurement could have changed) while skipping the forced-layout DOM read on every other change-detection pass.

### Scope and transparency

The PR explicitly identifies and defers three related issues rather than scope-creeping them in: the identical maxSize-dropping pattern in `RegionPropertyDialogService.java:336-338`, the pre-existing (and only narrowed, not closed) race where the second `getVGraphPair` call runs outside `box.lockRead()`, and two deliberately-out-of-scope perf issues (`Tool.clone()` deep-cloning the whole chart model per command, and synchronous `canvas.toDataURL()` PNG encoding) that likely explain why the burst *feels* like a freeze. This is exactly the right level of honesty for a scoped bug fix — flagging what's adjacent without trying to fix everything at once.

### Tests

`tsc --noEmit` and `mvnw compile` are clean, and the existing 17 TL spec files covering the touched graph/binding-pane/object-container areas pass (590 tests), but no *new* test was added for either the Java fix or the two TypeScript hardening changes. Given the nature of the bug (a multi-request-cycle race that produced a specific, reproducible HAR trace), a regression test would be valuable but is admittedly hard to write cheaply here (would need to mock/stub `ViewsheetSandbox.getVGraphPair`'s cancel/rebuild behavior, or drive `ChartImageDirective` through a scripted sequence of `Retry-After` responses). The PR is upfront that manual verification with the reporter's data is still pending — worth confirming before merge given this is exactly the kind of timing-dependent bug that's easy to declare fixed prematurely.

### Blocking issues for merge

## Recommendation

Requesting changes — due to the failing `claude-review` check and unconfirmed mergeability, not due to any defect found in the fix itself. This is a thoroughly-diagnosed and correctly-implemented fix across all three files I checked; the main outstanding item is the PR's own pending manual verification against the reporter's original repro data, which should be completed before merge given the timing-sensitive nature of the bug.

jshobe-inetsoft (approved):

PR #4717, opened 2026-08-21
Bug #76022, keep fixed-height bottom-tab children flush in the scale-to-screen path

ngovus-inetsoft:
Follow-up to #4635 — QA reported the gap is still visible in Chrome device emulation.

## Why #4635 didn't cover it

The reported asset's device layout `a` targets the "Small devices" device (**minWidth 768 / maxWidth 991**), so on a 375px phone `LayoutInfo.matchLayout` finds no match and **no device layout is applied**. The viewsheet has `scaleToScreen="true"`, so the geometry in the screenshot comes entirely from `CoreLifecycleService` → `VSEventUtil.applyScale` → `applyTabScale` — a path #4635 (layout space: `AbstractLayout.applyTab`, `VSLayoutService`) never touches. Pre-existing defect in the scaled path, not a regression.

## Root cause

`applyTabScale` gives the tab bar's unscaled slack to every child:

```java
double repairH = tabSize.height * scaleRatio.y - tabSize.height;   // 24 * (ry - 1)
scaleSize.height = (int) Math.floor(size.height * sizeScale.y + repairH);
int childY = bottomTabs ? tabPos.y - scaleSize.height : tabPos.y + tabSize.height;
```

That is right for a child that scales vertically — it keeps the tab group's total scaled extent correct while the bar stays 24px. But `InputVSAssemblyInfo.getSizeScale` returns `(rx, 1)`: a radio button never scales vertically. In the asset (`Gauge1` 140@204, `RadioButton1` 40@304, `Tab1` 24@344, `bottomTabs`, all design-time flush) at `ry ≈ 1.5`:

- `Gauge1`: `140*1.5 + 12 = 222` — box grows with the content, bottom edge on the strip. Correct.
- `RadioButton1`: `40*1 + 12 = 52` — box inflated by 12px while the rendered content is still 40px (title 20 + one 20px row) and top-aligned inside it. The box bottom meets the strip; the visible radios stop 12px short of it. That is the reported gap, and it explains why only the radio looks wrong.

Same class of assembly affected wherever `getSizeScale().y != scaleRatio.y`: `SubmitVSAssemblyInfo`, `TimeSliderVSAssemblyInfo`, dropdown `SelectionList`/`Calendar`, non-vertical-scaling `TextVSAssemblyInfo`, nested `TabVSAssemblyInfo`. It only *looks* broken for bottom tabs, because there the strip sits directly under the child; with top tabs the slack lands below the content where nothing anchors it.

## Fix

Only hand the slack to children that actually scale vertically:

```java
double childRepairH = sizeScale.y == scaleRatio.y ? repairH : 0;
```

- `scaleRatio.y == 1` → `repairH` is already 0, so nothing changes there.
- Bottom tabs: `childY = tabPos.y - scaleSize.height` still anchors to the bar, now with the correct height, so the *content* is flush too. Also fixes the `ry < 1` mirror case, where the box shrank below the content and the radios overlapped the strip.
- Top tabs: positions unchanged (`tabPos.y + tabSize.height`); fixed-height children just stop reporting a padded height.
- The embedded-viewsheet branch further down the loop reuses `scaleSize`; `ViewsheetVSAssemblyInfo` has no `getSizeScale` override, so it keeps `repairH` and is unaffected.

## Tests

New `VSEventUtilTabScaleTest` rebuilds the reported asset's geometry and drives the real `VSEventUtil.applyScale`. Confirmed failing without the fix:

```
fixedHeightChildKeepsNaturalHeightWhenScalingUp   expected: <40> but was: <52>
fixedHeightChildKeepsNaturalHeightWhenScalingDown expected: <40> but was: <28>
```

The third test pins the unchanged behavior (a vertically scaling gauge still absorbs the slack and stays flush).

Regression run, all green: `VSEventUtilTabScaleTest, AbstractLayoutTest, VSLayoutServiceTest, TabVSAScriptableTest, TabVSAssemblyTest, TabPropertyDialogServiceTest, SelectionTreePropertyDialogServiceTest, ComposerObjectServiceTest`.

Not yet verified in a running server; reporter will check iPhone SE (375px, the pure scale-to-screen case from the report) and ~800px (where layout `a` matches, exercising layout + scale together).

## Known adjacent issues, deliberately left alone

- `applyTabScale` anchors the bar from `TabVSAssemblyInfo.getBottomTabChildHeight` (label-**inclusive**) but places children from the declared height (label-**exclusive**). That bites a bottom-tab input child with a top/bottom label — not this asset (no `labelInfo`), and the client already offsets dropdown selections/calendars itself, so changing it risks double-shifting.
- `handleOverlapping` runs after `applyTabScale` and, for a *visible* overlapping `ListInputVSAssembly`, re-runs `applyAssemblyScale`, discarding the tab anchoring. It does not fire here (the unselected tab child is invisible) but it is latent.

### DIFF
--- a/core/src/main/java/inetsoft/analytic/composition/event/VSEventUtil.java
+++ b/core/src/main/java/inetsoft/analytic/composition/event/VSEventUtil.java
@@ -1322,8 +1322,14 @@ public final class VSEventUtil {
             Dimension scaleSize = new Dimension();
 
             scaleSize.width = (int) Math.floor(size.width * sizeScale.x);
+            // the tab bar's height is not scaled, so its slack is absorbed by the
+            // children -- but only by children that scale vertically. A fixed-height
+            // child (input, submit, dropdown selection, nested tab) renders at its
+            // natural height, so adding the slack inflates its box and leaves the
+            // content detached from a bottom tab bar (Bug #76022).
+            double childRepairH = sizeScale.y == scaleRatio.y ? repairH : 0;
             scaleSize.height =
-               (int) Math.floor(size.height * sizeScale.y + repairH);
+               (int) Math.floor(size.height * sizeScale.y + childRepairH);
             scaleSize.height = Math.max(0, scaleSize.height);
 
             // top tabs: children below tab bar; bottom tabs: children above tab bar
--- a/core/src/test/java/inetsoft/analytic/composition/event/VSEventUtilTabScaleTest.java
+++ b/core/src/test/java/inetsoft/analytic/composition/event/VSEventUtilTabScaleTest.java
@@ -1,0 +1,149 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.analytic.composition.event;
+
+import inetsoft.report.composition.execution.ViewsheetSandbox;
+import inetsoft.test.*;
+import inetsoft.uql.viewsheet.*;
+import inetsoft.uql.viewsheet.internal.TabVSAssemblyInfo;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mockito;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import java.awt.*;
+import java.awt.geom.Point2D;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Tests for bottom-tab child geometry in the scale-to-screen path
+ * ({@code VSEventUtil.applyTabScale}).
+ */
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = { BaseTestConfiguration.class },
+                      initializers = ConfigurationContextInitializer.class)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
+@SreeHome
+@Tag("core")
+class VSEventUtilTabScaleTest {
+   /**
+    * Bug #76022: the tab bar's height is not scaled, so its slack is handed to the
+    * children. A radio button never scales vertically
+    * ({@code InputVSAssemblyInfo.getSizeScale} returns y=1), so adding the slack
+    * inflated its box and left the rendered content floating above the tab bar.
+    */
+   @Test
+   void fixedHeightChildKeepsNaturalHeightWhenScalingUp() throws Exception {
+      Viewsheet vs = createTabViewsheet("Gauge1");
+      applyScale(vs, 1.0, 1.5);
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Correctness

This is a precise, well-verified fix. I checked the central claim — that `getSizeScale(scaleRatio).y` differs from `scaleRatio.y` for exactly the assembly types the PR lists — against every `getSizeScale()` override in the codebase:

| Type | `getSizeScale(scaleRatio).y` | Gets `childRepairH`? |
|---|---|---|
| `VSAssemblyInfo` (default) | `scaleRatio.y` | Yes |
| `InputVSAssemblyInfo`, `SubmitVSAssemblyInfo`, `TimeSliderVSAssemblyInfo`, `TabVSAssemblyInfo` (nested) | `1` | No |
| `SelectionListVSAssemblyInfo`, `CalendarVSAssemblyInfo` | `scaleRatio.y` unless dropdown mode → `1` | Dropdown: No |
| `TextVSAssemblyInfo` | `scaleRatio.y` if `isScaleVertical()` else `1.0` | Non-vertical-scaling: No |

This matches the PR's description exactly, and the `sizeScale.y == scaleRatio.y` check is safe as an exact `==` comparison — both sides are the same unmodified `double` value passed through by reference/value in the fixed-height cases (a literal `1`, or `scaleRatio.y` itself with no arithmetic applied), so there's no floating-point-equality fragility here.

The fix is a single added conditional (`childRepairH`) applied only to the height calculation; the position-anchoring logic (`childY = tabPos.y - scaleSize.height` for bottom tabs) is unchanged and, combined with the now-correct `scaleSize.height`, produces the flush anchoring the bug report wants. I also traced the embedded-viewsheet branch a few lines below (`ViewsheetVSAssemblyInfo` reuses `scaleSize` but has no `getSizeScale` override) and confirmed the PR's claim that it's unaffected, since it always falls into the "gets repairH" case via the default `VSAssemblyInfo` implementation.

### Tests

Excellent regression coverage. `VSEventUtilTabScaleTest` rebuilds the exact reported asset geometry (`Gauge1`/`RadioButton1`/`Tab1` at the stated pixel offsets) and drives the real `VSEventUtil.applyScale` rather than a synthetic/mocked version of the buggy method, so the test exercises the actual production code path. The PR documents both new tests failing with the precise pre-fix numbers (`expected: <40> but was: <52>` and `expected: <40> but was: <28>`), which is exactly the kind of before/after verification that gives confidence the test is load-bearing rather than vacuous. The third test correctly pins the still-desired behavior for a vertically-scaling child.

### Root-cause diligence

The PR explains, with specific line-level references, exactly why the earlier #4635 fix (in `AbstractLayout.applyTab`/layout space) didn't cover this — the reported asset's device layout doesn't match a 375px viewport, so the geometry comes entirely from the separate scale-to-screen path. It also proactively documents two adjacent, still-latent issues (`getBottomTabChildHeight` label-inclusive vs. declared-height label-exclusive mismatch; `handleOverlapping` discarding tab anchoring for a visible overlapping `ListInputVSAssembly`) that this PR deliberately doesn't touch, with a clear rationale for why each is out of scope. That's the right level of transparency for a scoped bug fix rather than a rewrite.

### CI failure investigation

The `build` check is `FAILURE`. I pulled the actual Maven test log (job 96863334269) and the single failure is `DashboardRegistryOrgLifecycleTest.rename_setOrganizationInfo_adminAndPerUserRegistryContentRewritten` — an org-lifecycle/dashboard-registry-rewrite test with no relation to this PR's `VSEventUtil` tab-scaling change. This is unrelated to the diff under review (note: I've now seen three different unrelated org-lifecycle-area test failures break the `build` check across separate PRs in this batch — may be worth flagging to the team as a broader CI-stability issue in that test area, independent of any of these individual PRs).

- No backward-incompatibility or security concerns; this only changes internal scale-to-screen geometry math for a specific combination of assembly types.

### Blocking issues for merge

2. **Failing check.** The `build` check reports `FAILURE` — per my investigation above, this is an unrelated pre-existing test failure (`DashboardRegistryOrgLifecycleTest`), not caused by this PR.
3. **Unconfirmed mergeability.** `mergeable`/`mergeStateStatus` report `UNKNOWN` rather than a confirmed conflict-free state.

## Recommendation

Requesting changes — due to the failing checks and unconfirmed mergeability noted above, not due to any defect found in the fix itself, which is precise, well-reasoned, and thoroughly tested against the actual reported geometry.

jshobe-inetsoft (approved):

PR #4718, opened 2026-08-21
Bug #75930, drop the round corner corner-masks in Excel/PPT table export

ngovus-inetsoft:
## Summary
Follow-up to #4557. Testing of the merged build reported that on Excel and PPT export of a round-corner Table, *"some styles are missing on round corner"* — white notches at the corners where the outer border stroke, the title-row background fill, and the left edge of the title text were wiped out.

Root cause is the corner-masking hack #4516/#4557 added for the two formats that cannot clip:

- `PoiExcelVSExporter.maskTableCorners()`/`maskCorner()` — four opaque white `radius x radius` `XSSFSimpleShape` squares at the object's corners.
- `PPTValueHelper.maskTableCorners()`/`maskCorner()` — the same with `XSLFAutoShape`s, and with no radius cap (unlike Excel's `MAX_TABLE_ROUND_RADIUS = 8`), so the damage there scales with the configured round corner.

The region that should be hidden is the corner square **minus** the quarter-disc inside the arc. A full square also erases everything that legitimately belongs *inside* the rounded shape — hence the missing border/fill/text.

PDF/SVG/PNG are unaffected: they use a real `Graphics2D` `RoundRectangle2D` clip (`beginRoundCornerClip`/`endRoundCornerClip` in `VSTableDataHelper`), which is correct.

Excel and PPT have no equivalent clip, so this drops the masking in both and keeps only the rounded-rectangle outline overlay. Remaining accepted tradeoff: a small square sliver of cell fill/border can show just outside the arc at each corner — bounded by the 8px cap in Excel, and far less objectionable than erasing styles. (The exact-fidelity alternative — custom-geometry `arcTo` corner wedges filled with the resolved background color — is a bigger change and can follow if that sliver proves visible in practice.)

Everything else from #4557 is unchanged: the `setRoundCorner` copies in the six Table/Crosstab helpers, the PPT `setCellType(0)` reset, the PDF/SVG/PNG clips, and the HTML `getRoundCorner()`/`overflow:hidden` fix.

## Test plan
- [ ] Export a Table/Crosstab with round corner + a full border and a shaded title row to Excel and PPT; confirm no white corner notches, the outer border runs continuously into the curve, and the title fill/text are intact.
- [ ] Repeat with round corner set but no border configured; confirm no stray shapes (unchanged early return).
- [ ] Repeat with a partial border (top only); confirm the documented uniform-outline tradeoff still looks reasonable.
- [ ] Regression: non-rounded Table/Crosstab to Excel/PPT unchanged, and the same rounded viewsheet to PDF/PNG/HTML still renders correctly.

### DIFF
--- a/utils/inetsoft-xml-formats/src/main/java/inetsoft/report/io/viewsheet/excel/PoiExcelVSExporter.java
+++ b/utils/inetsoft-xml-formats/src/main/java/inetsoft/report/io/viewsheet/excel/PoiExcelVSExporter.java
@@ -1168,9 +1168,8 @@ public class PoiExcelVSExporter extends ExcelVSExporter {
          color = bcolors == null ? null : bcolors.bottomColor;
       }
       else {
-         // No border configured at all: there's no outline to compensate the mask with a
-         // curve, so skip masking too rather than leaving an unrounded white square with
-         // no border and no explanation.
+         // No border configured at all: the rounded outline shape is the only thing drawn
+         // here, so there is nothing to round.
          return;
       }
 
@@ -1179,13 +1178,6 @@ public class PoiExcelVSExporter extends ExcelVSExporter {
       if(lineStyle == ExcelVSUtil.EXCEL_NO_BORDER) {
          return;
       }
-
-      // Mask the square cell background/content that would otherwise stick out past the
-      // curve at each corner (e.g. a shaded title-row fill, or the corner-most cell's
-      // text) with a plain white square, the same way the viewer's overflow:hidden clips
-      // it. This must happen before the border outline below so the outline's curve is
-      // drawn on top of the mask, not under it.
-      maskTableCorners(pixelBounds, radius);
 
       XSSFSimpleShape shape = patriarch.createSimpleShape(
          (XSSFClientAnchor) getAnchorFromPixelRect(pixelBounds));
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Correctness

Correct, low-risk fix that is a clean, complete removal rather than a patch on top of the flawed approach. The root-cause analysis holds up: a full `radius x radius` opaque white square necessarily erases everything under it, including the parts of the border/fill/text that legitimately belong *inside* the rounded corner's arc — not just the sliver outside it that needed hiding. Since Excel/PPT (unlike PDF/SVG/PNG, which use a real `Graphics2D` `RoundRectangle2D` clip per the PR's description) have no way to clip cell rendering to a rounded shape, dropping the mask and accepting a small uncovered square sliver just outside the curve is a reasonable, clearly-documented tradeoff versus the previous "erase real content" behavior.

I verified the removal is complete and leaves no dead code or dangling references:
- Both `maskTableCorners()`/`maskCorner()` private helper method pairs (Excel and PPT) are deleted in their entirety along with their sole call sites — grepping confirms no other reference to either method survives in either file.
- In `PoiExcelVSExporter`, `radius` (still used a few lines later in `applyRoundCorner(shape, radius, pixelBounds)`) and in `PPTValueHelper`, `x`/`y`/`width`/`height` (still used in the subsequent `roundBorder.setAnchor(...)` call) remain referenced after the mask-call removal, so no now-unused-variable compile warnings are introduced.
- The early-return comment in the Excel path is updated to match the new, simpler reality ("the rounded outline shape is the only thing drawn here, so there is nothing to round") rather than left describing removed masking logic.

### Scope

Appropriately scoped: the PR explicitly leaves untouched everything else from the prior #4557 change (the `setRoundCorner` copies, the PPT `setCellType(0)` reset, the PDF/SVG/PNG clips, and the HTML fix), touching only the two masking methods that caused this regression.

- No automated test coverage for this purely visual export path (consistent with the rest of this export code, which the PR's manual test plan checkboxes appropriately target instead — table+border+shaded-title export to Excel/PPT, no-border early-return regression, partial-border tradeoff, and non-rounded/PDF/PNG/HTML regression checks). None of the plan's checkboxes are marked done in the PR body; worth confirming these were actually run before merge given this is a visual-only change with no way to catch a regression via `mvnw test`.
- No backward-incompatibility or security concerns — this is a pure visual-rendering fix with no API or data-model changes.

### Blocking issues for merge

## Recommendation

Requesting changes — solely because of the failing `claude-review` check. The fix itself is a correct, clean, appropriately-scoped removal; please confirm the manual test plan's checkboxes were actually exercised before merge, since this change has no automated coverage.

jshobe-inetsoft (approved):

PR #4719, opened 2026-08-21
Bug #76238, guard against a null assembly lookup in ChartVSAQuery.getAssembly()

jkidacki-inetsoft:
## Summary

`ChartVSAQuery.getAssembly()` threw an intermittent NullPointerException in the Visualization Wizard when switching chart types (repro is unstable/timing-dependent), logged as:

```
java.lang.NullPointerException: Cannot invoke "inetsoft.uql.viewsheet.VSAssembly.clone()" because the return value of "inetsoft.report.composition.execution.CubeVSAQuery.getAssembly()" is null
    at inetsoft.report.composition.execution.ChartVSAQuery.getAssembly(ChartVSAQuery.java:74)
    at inetsoft.report.composition.execution.ChartVSAQuery.getTableAssembly(ChartVSAQuery.java:595)
    at inetsoft.report.composition.execution.ChartVSAQuery.getTableLens(ChartVSAQuery.java:360)
    at inetsoft.report.composition.execution.ChartVSAQuery.getData(ChartVSAQuery.java:104)
    ...
    at inetsoft.web.vswizard.handler.VSWizardBindingHandler.changeFormat(VSWizardBindingHandler.java:2168)
```

## Root Cause

`ChartVSAQuery.getAssembly()` calls `super.getAssembly().clone()` unconditionally. The base implementation (`VSAQuery.getAssembly()`, inherited unchanged through `CubeVSAQuery`/`DataVSAQuery`) is `getViewsheet().getAssembly(vname)`, which legitimately returns `null` when the named assembly is transiently absent from the viewsheet — exactly the case during the wizard's rapid chart-type-switch flow, where the temp chart assembly bound to the query's `vname` can momentarily not be present. Sibling methods in the same class (`getData()` and `getTableAssembly()`) already null-check `getAssembly()`'s result, confirming `null` is an expected, designed return value that this override alone failed to honor by dereferencing it with `.clone()` before the null check could ever run.

## Fix

`ChartVSAQuery.getAssembly()` now null-checks the base assembly before cloning it, returning `null` instead of throwing, consistent with how the rest of the class already treats a missing assembly.

## Test Plan

- `./mvnw clean install -pl core -am -DskipTests` — build succeeds
- `./mvnw test -pl core -Dtest=ChartVSAQueryTest` — 2/2 existing tests pass
- Manual repro: open a chart in the wizard and rapidly switch chart types; previously intermittently logged the NPE above via `VSWizardBindingHandler.changeFormat`, now returns `null` gracefully with no error logged.

Fixes Redmine #76238.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/ChartVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/ChartVSAQuery.java
@@ -71,7 +71,8 @@ public class ChartVSAQuery extends CubeVSAQuery implements BindableVSAQuery {
          return assembly;
       }
 
-      return assembly = (VSAssembly) super.getAssembly().clone();
+      VSAssembly base = super.getAssembly();
+      return assembly = base == null ? null : (VSAssembly) base.clone();
    }
 
    /**


jshobe-inetsoft (changes requested):
## Code Review

### Correctness

Correct, minimal, one-line-logic fix. I traced the full call chain the reported stack trace implies: `getData()` (line 104-109, already null-checks `chart`) → `getTableLens()` (line 360-364, already returns `null` early when `getTableAssembly()` returns `null`) → `getTableAssembly()` (line 594-599, **already** null-checks `getAssembly()` and returns `null`). So the null-handling contract already existed at every level above `getAssembly()` itself — the only broken link was `getAssembly()` unconditionally calling `.clone()` on a possibly-null `super.getAssembly()` result before any of that null-checking machinery could run. The fix closes exactly that gap and nothing more.

I also checked the `assembly` field's caching behavior: `getAssembly()` only caches (and short-circuits future calls via `if(assembly != null) return assembly;`) once it successfully resolves a non-null value. A `null` result is never cached, so a transient miss doesn't "poison" the query object — the next call re-attempts the lookup fresh, and once the assembly does resolve, every subsequent caller within the same query instance's lifetime gets the same cached, real assembly.

### Side-effect / regression check

This class has ~13 other internal call sites that cast and dereference `getAssembly()`'s result directly without a null-check (e.g. `getAggregateInfo()` line 276, `setSharedCondition()` line 285, `getDefaultColumnSelection()` line 338, two spots inside `getTableLens()` itself at lines 374 and 394). I confirmed these are **not** newly at risk from this change: before the fix, a transient null from `super.getAssembly()` threw inside `getAssembly()` itself; after the fix, the same transient-null scenario would instead throw one line later, at whichever unguarded call site dereferences the now-`null` result — same ultimate crash, just relocated by one frame. So this PR doesn't make those other paths any less safe than they already were; it specifically and correctly fixes the one call chain (`getData → getTableLens → getTableAssembly`) that already had a designed null-propagation contract in place, which is exactly the class of fix that's safe to make in isolation. Auditing and null-guarding the remaining call sites would be a reasonable, separate, larger follow-up, not something this fix needs to take on.

### Tests

The PR relies on the existing `ChartVSAQueryTest` (2/2 passing) plus a manual repro of the original intermittent-timing bug. Given the root cause is a genuine race (transient assembly absence during rapid UI interaction), a new deterministic unit test would likely need to mock/stub `super.getAssembly()` to return `null`, which isn't present here — a reasonable, if not exhaustive, level of verification for a single-line defensive-null-check fix.

- No backward-incompatibility or security concerns; `getAssembly()` is a `protected` method internal to this query class and its subclasses, not part of any public API surface.

### Blocking issues for merge

## Recommendation

Requesting changes — due to the failing `claude-review` check and the unconfirmed mergeability state, not due to any defect found in the fix itself, which is correct and appropriately scoped.

jshobe-inetsoft (approved):

PR #4720, opened 2026-08-21
Bug #76236, guard chart binding tree against a disposed viewsheet

jkidacki-inetsoft:
## Summary

Opening a heavy chart and quickly clicking "Binding Editor" or "Refresh Tree" before the viewsheet finishes loading could throw a `NullPointerException`:

```
java.lang.NullPointerException: Cannot invoke "inetsoft.uql.viewsheet.Viewsheet.getAssembly(String)" because "vs0" is null
        at inetsoft.web.binding.handler.VSTreeHandler.getChartTreeModel(VSTreeHandler.java:69)
```

## Root Cause

`RuntimeViewsheet.getViewsheet()` returns `null` once the `RuntimeViewsheet` has been disposed (`RuntimeViewsheet.dispose()` sets its `vs` field to `null`). `VSBindingTreeService.getBinding()` reads `rvs.getViewsheet()` once (non-null) to fetch the assembly info, then calls `VSTreeHandler.getChartTreeModel()`, which reads `rvs.getViewsheet()` a *second* time. Rapid user interaction with a still-loading, heavy chart can trigger a concurrent dispose/reset of the `RuntimeViewsheet` in the window between those two reads, so the second read returns `null` and `vs0.getAssembly(aname)` NPEs.

## Fix

- `VSTreeHandler.getChartTreeModel()` now returns `null` when the viewsheet (`vs0`) or the resolved chart assembly is unavailable, instead of dereferencing a possibly-null reference.
- Updated all three callers of `getChartTreeModel()` to tolerate a `null` result without introducing a new NPE further down the call chain:
  - `VSBindingTreeService.getBinding()` — guards the `appendVSAssemblyTree()` call.
  - `VSWizardBindingService` and `VSWizardObjectService` — guard the `assetTreeModel.getRoot()` dereference.

## Test Plan

- `./mvnw clean test-compile -pl core -am` — builds clean.
- `./mvnw test -pl core -am` — full core test suite: 4393 tests, 0 failures, 0 errors.
- Reviewed with the code-reviewer agent across two passes to confirm no regressions and that the null-return contract is honored by every caller.

Fixes Redmine #76236.

### DIFF
--- a/core/src/main/java/inetsoft/web/binding/handler/VSTreeHandler.java
+++ b/core/src/main/java/inetsoft/web/binding/handler/VSTreeHandler.java
@@ -64,9 +64,20 @@ public class VSTreeHandler {
       throws Exception
    {
       Viewsheet vs0 = rvs.getViewsheet();
+
+      // viewsheet may have been disposed/reset concurrently (e.g. user interacts with the
+      // binding tree before a heavy chart finishes loading), so vs0 can be null here.
+      if(vs0 == null) {
+         return null;
+      }
+
       String aname = info.getAbsoluteName2();
       aname = aname == null ? info.getAbsoluteName() : aname;
       ChartVSAssembly chart = (ChartVSAssembly) vs0.getAssembly(aname);
+
+      if(chart == null) {
+         return null;
+      }
 
       final Viewsheet vs = chart.getViewsheet();
       final String name = chart.getName();
--- a/core/src/main/java/inetsoft/web/binding/service/VSBindingTreeService.java
+++ b/core/src/main/java/inetsoft/web/binding/service/VSBindingTreeService.java
@@ -82,7 +82,7 @@ public class VSBindingTreeService {
                assetTreeModel = treeHandler.getChartTreeModel(
                   engine.getAssetRepository(), rvs, (ChartVSAssemblyInfo) info, false, principal);
 
-               if(appendComponentTree) {
+               if(assetTreeModel != null && appendComponentTree) {
                   treeHandler.appendVSAssemblyTree(rvs, assetTreeModel, principal, assembly);
                }
             }
--- a/core/src/main/java/inetsoft/web/vswizard/controller/VSWizardBindingService.java
+++ b/core/src/main/java/inetsoft/web/vswizard/controller/VSWizardBindingService.java
@@ -106,6 +106,11 @@ public class VSWizardBindingService {
          AssetTreeModel assetTreeModel = treeHandler.getChartTreeModel(
             viewsheetService.getAssetRepository(), rvs, temporaryInfo.getTempChart().getChartInfo(),
             true, principal);
+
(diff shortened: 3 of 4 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Correctness

Correct, minimal, well-scoped null-guard fix. `RuntimeViewsheet.getViewsheet()` returning `null` after a concurrent dispose is a real race (the PR's stack trace and root-cause explanation match the described `VSTreeHandler.getChartTreeModel()` code path: `rvs.getViewsheet()` is read once by the caller and once more inside `getChartTreeModel()`, and only the second read is guarded before this fix). Adding the `vs0 == null` and `chart == null` early returns in `getChartTreeModel()` closes the NPE at its source.

I checked all three updated call sites for correct null-handling:
- `VSBindingTreeService.getBinding()` — the `appendComponentTree` branch is now additionally gated on `assetTreeModel != null`, which correctly no-ops the tree-append instead of NPE'ing on a null model.
- `VSWizardBindingService.getBindingTree()` — the new `if(assetTreeModel == null) return null;` sits inside the same `try` block as the existing `finally { box.get().unlockRead(); }`, so the read lock is still correctly released on this early-return path. It also mirrors an existing precedent a few lines above in the same method (`temporaryInfo == null` also logs and returns `null` without sending a command), so this isn't introducing a new pattern.
- `VSWizardObjectService` — the equivalent guard is inside the same `try` block that has `finally { ...setTempMaxRow(-1); }`, so that cleanup still runs on the early-return path too. Worth noting (non-blocking): this call site sends `OpenObjectWizardCommand` to the client *before* building the tree, so in the rare race this fix targets, the client will see the wizard open but never receive the follow-up `InitWizardBindingTreeCommand` populating it — a degraded-but-non-crashing outcome, which is still strictly better than the prior unhandled NPE.

### Tests

No new automated test was added for the race itself (understandably hard to reproduce deterministically without injecting a timing hook into `RuntimeViewsheet` disposal), but the PR reports the full core suite passing (4393 tests) and states two passes of independent code review confirming every caller honors the new null-return contract, which I independently verified above.

- No backward-incompatibility or security concerns — this only adds defensive null-handling to an existing internal API; `getChartTreeModel()` wasn't previously documented as never returning null, so callers gaining a null-check is a strict correctness improvement, not a contract break.

`gh pr view` reports both `mergeable` and `mergeStateStatus` as `UNKNOWN` (re-checked, unchanged) rather than a definitive `MERGEABLE`/`CONFLICTING`. I can't confirm from this alone whether there are merge conflicts against the base branch.

### Blocking issues for merge

## Recommendation

Requesting changes — due to the failing `claude-review` check and the unconfirmed mergeability state, not due to any defect found in the fix itself, which is correct and appropriately scoped.

jshobe-inetsoft (approved):

PR #4721, opened 2026-08-21
Bug #76194, enforce the destination folder before mutating a renamed asset

jkidacki-inetsoft:
## Summary

`POST /api/public/viewsheets/rename` (and any other caller of `AssetRepository.changeSheet()`) could permanently destroy the source asset without ever creating the renamed copy, when the rename/move failed partway through.

## Root Cause

In `AbstractAssetEngine.changeSheet0()`, the source entry was removed from its old parent folder, that folder change was persisted, and the source's storage record was deleted — all *before* the destination parent folder was validated, before the destination entry was added to its folder, and before the new sheet was written to storage. Any failure in that window (most commonly: the destination's parent folder doesn't exist yet, causing `getParentFolder()` to return `null` and a subsequent NPE on `npfolder.addEntry(nentry)`) left the source deleted and the destination never created. The client only saw a generic HTTP 500 with no indication of the data loss.

There were also two separate `ostorage.remove(oidentifier)` calls in the method — one premature (right after removing the entry from the old folder) and one at the very end, guarded by a comment indicating removal should happen last so dependency rewriting isn't disrupted. The early one appears to be a leftover duplicate that defeated that intent.

## Fix

- Resolve and validate the destination's parent folder up front, before any mutation of the source entry; throw a clear `MessageException` instead of proceeding into an NPE.
- Remove the premature deletion of the source's old folder listing and storage record.
- Defer that removal/overwrite to the very end of the method, after the destination entry has been added to its folder and the new sheet has been written to storage, and after dependent-sheet and bookmark updates have completed successfully.

This preserves the existing behavior for all currently-succeeding renames; it only changes what happens when the rename would otherwise fail — the source is now left untouched instead of being destroyed.

## Test Plan

- Reproduced the original defect: renaming/moving a private viewsheet into a `GLOBAL_SCOPE` folder path that doesn't exist returned HTTP 500 and permanently deleted the source viewsheet with nothing created at the destination.
- Verified with this fix that the same request now fails cleanly and the source viewsheet remains intact and unchanged.
- `./mvnw -pl core -am compile` and `test-compile` both succeed under Java 21.
- Independent code review confirmed the reordering preserves correct behavior for cross-folder moves, same-folder renames, alias-only renames within the same folder (a subtly different `AssetEntry.equals()`/`hashCode()` edge case), and both worksheet and viewsheet paths, with no regression risk to the two existing call sites of `changeSheet0`.

Fixes Redmine #76194.

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/AbstractAssetEngine.java
+++ b/core/src/main/java/inetsoft/uql/asset/AbstractAssetEngine.java
@@ -3077,6 +3077,22 @@ public abstract class AbstractAssetEngine implements AssetRepository, AutoClosea
          return;
       }
 
+      // Bug #76194: resolve and validate the destination parent folder up front,
+      // before any mutation of the source entry. Previously this was fetched (and
+      // left unchecked for null) only after the source had already been removed
+      // from its folder and deleted from storage, so any failure past that point
+      // permanently destroyed the source sheet without ever creating the renamed copy.
+      AssetEntry npentry = nentry.getParent();
+      AssetFolder npfolder = getParentFolder(nentry, nstorage);
+      String npidentifier = npentry.toIdentifier();
+
+      if(npfolder == null) {
+         MessageException ex = new MessageException(
+            catalog.getString("common.pfolderNotFound") + ": " + npentry);
+         ex.setKeywords("FOLDER_REQUIRED");
+         throw ex;
+      }
+
       DependencyHandler dependencyHandler = DependencyHandler.getInstance();
 
       if(oentry.isWorksheet()) {
@@ -3121,9 +3137,6 @@ public abstract class AbstractAssetEngine implements AssetRepository, AutoClosea
          }
       }
 
-      AssetEntry npentry = nentry.getParent();
-      AssetFolder npfolder = getParentFolder(nentry, nstorage);
-      String npidentifier = npentry.toIdentifier();
       opfolder = getParentFolder(oentry, ostorage);
 
       if(npentry.equals(opentry)) {
@@ -3140,16 +3153,6 @@ public abstract class AbstractAssetEngine implements AssetRepository, AutoClosea
       }
 
       opfolder.removeEntry(oentry);
-
-      if(!Objects.equals(opidentifier, npidentifier) || !Objects.equals(opfolder, npfolder)) {
-         ostorage.putXMLSerializable(opidentifier, opfolder);
-      }
-
-      if(!Objects.equals(oidentifier, nidentifier)) {
-         ostorage.remove(oidentifier);
-         EmbeddedDataCacheHandler.clearWSCache(oidentifier);
-      }
-
       npfolder.addEntry(nentry);
 
       if(nentry.getAlias() != null && oentry.getAlias() != null &&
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Correctness

I read the full `changeSheet0()` method at HEAD and traced the diff against it carefully; this is a correct and notably well-reasoned fix.

**Validation is hoisted to the right place.** The new `npfolder == null` check (with a clear `MessageException`/`FOLDER_REQUIRED` keyword, using the pre-existing `common.pfolderNotFound` catalog key — verified it exists in `srinter.properties`) is inserted immediately after the "source sheet missing" early-return, i.e. *before* the dependency-rewriting block (`dependencyHandler.updateDependencies`/`renameDependencies`, `XRepository.renameTransform`, `cluster.sendMessage`). This is better than just fixing the reported symptom: in the old code, a doomed rename would run through all of that dependency rewriting for *other* assets before finally NPE'ing on `npfolder.addEntry(nentry)` — so a failed rename didn't just destroy the source, it could also have already rewritten other assets' dependency references to point at a `nentry` that was never actually created. Moving the validation up front prevents that too, not just the storage deletion.

**The reordering correctly defers all source-mutating storage writes.** I traced the two `ostorage.remove(oidentifier)` call sites in the pre-fix code: one right after `opfolder.removeEntry(oentry)` (paired with `EmbeddedDataCacheHandler.clearWSCache`), and a second, later one at the very end (guarded by the "remove oidentifier last" comment) that called `ostorage.remove()` again without `clearWSCache`. The fix removes the early pair entirely and consolidates into a single call at the end that now includes both `ostorage.remove()` *and* `clearWSCache()` — so this isn't just a reordering, it also fixes a redundant double-remove and a `clearWSCache` gap on the previously-authoritative last-removal path. The `opfolder` persist (`ostorage.putXMLSerializable(opidentifier, opfolder)`) is deferred the same way, right alongside it.

**No stale-data risk from hoisting `npentry`/`npfolder`/`npidentifier`.** These are now computed once, before the dependency-handling block, instead of twice (the redundant second computation is deleted). I checked whether anything in between could invalidate them: the only mutation on `nentry` in that window is `nentry.setProperty("__bookmark_id__", null)` in the viewsheet branch, which doesn't affect `nentry.getParent()` or its path/identifier — so resolving the destination folder earlier is safe.

**Partial-failure behavior beyond the targeted window is improved, not perfect — worth calling out for the reviewer, not blocking.** If `changeSheetDependents()` or `renameVSBookmark()` throws *after* `nstorage.putXMLSerializable(nidentifier, sheet)` has already run but before the final source-removal block, the destination copy now exists **and** the source is left untouched (since its folder-entry removal and storage delete are deferred past that point) — i.e., a duplicate rather than a loss. That's a strict improvement over total data destruction, but the operation still isn't fully atomic end-to-end. This is consistent with the PR's stated, narrower scope ("the source is now left untouched instead of destroyed"), so I'm not treating it as a gap in this fix — just flagging it in case the team wants to track it as a smaller follow-up.

### Tests

No new automated test was added (this is a multi-step storage-mutation-ordering bug that's hard to unit test without mocking `IndexedStorage`/`AssetFolder` extensively), but the PR documents a manual before/after repro (rename into a nonexistent `GLOBAL_SCOPE` folder path: HTTP 500 + destroyed source before, clean failure + intact source after) plus a stated independent code-review pass covering cross-folder moves, same-folder renames, alias-only renames, and both worksheet/viewsheet paths. Given the nature of the bug (an ordering defect in a large, side-effect-heavy method), a manual repro plus careful reasoning is a reasonable bar; I'd still suggest a regression test if one can be added cheaply (e.g. an integration test that mocks `getParentFolder` to return null and asserts the source entry/storage record are unchanged after the exception).

- No backward-incompatibility concerns; behavior for already-succeeding renames is unchanged, as reasoned above.
- No security concerns.

### Blocking issues for merge

## Recommendation

Requesting changes — solely because of the failing `claude-review` check. This is a careful, correct fix for a genuine data-loss bug; consider adding an automated regression test if feasible, and optionally tracking the still-imperfect partial-atomicity case noted above as a follow-up.

jshobe-inetsoft (approved):

PR #4722, opened 2026-08-21
Bug #75992, write the font family, not the face name, on office export

ngovus-inetsoft:
## Summary

Follow-up to #4559. QA reported that the column still wraps in the exported PPTX and still does not match the portal preview.

The video shows the viewer reporting a **missing font** and rendering a substituted face, which explains why the margin fudge in #4559 could not work: `PPTValueHelper.writeContent()` wrote the run's typeface with `Font.getFontName()`, which returns the **face** name (`Roboto Bold`, `Dialog.bold`), not a family. OOXML's `a:latin/@typeface` expects a family, and bold/italic are already written separately a few lines below. So PowerPoint could not resolve the typeface even on a machine that has the family installed — it substituted a wider face and re-wrapped cell text that the server had measured as fitting. No fixed point margin can compensate for an unknown substitute font.

Every other exporter already got this right: Excel writes `getFamily()` (`PoiExcelVSUtil.java:181,208`) and PPT's own rich-text path writes `rt.getFont().getFamily()` (line 373). The single-run path was the outlier.

## Changes

- **`VSFontHelper.getExportFontFamily(Font)`** (new) — returns the font family. When the family has collapsed to a Java logical name (`Dialog`, `SansSerif`, …), the server could not resolve the requested font, so it prefers `Font.getName()` — the name that was actually asked for, which survives on the font — and only falls back to a mapping onto Office-resolvable families (`Arial`, `Times New Roman`, `Courier New`) when that name is itself logical. Office cannot resolve Java logical names any more than the server could. Note the PDF-side map in `FontManager.java:85-92` targets base-14 PostScript names, which Office lacks, so these values intentionally differ.
- **`PPTValueHelper`** — both the single-run path (line 298) and the rich-text path (line 380) go through the helper.
- **`PoiExcelVSUtil`** — the same helper at lines 181 and 208. Both had to change together: `findFont()`'s cache lookup compares on the name, so a mismatch would silently duplicate fonts in the workbook.

Unstyled runs of an installed font are byte-identical to before (`getFontName() == getFamily()`), and user fonts are unaffected — `StyleFont` returns `userFontName` from both accessors.

`PoiExcelVSExporter.java:2573` is deliberately left unmapped: it takes a CSS `font-family` from HTML rich text, not a `java.awt.Font`.

## Test plan

- [x] `mvnw install -pl core,utils/inetsoft-xml-formats` — BUILD SUCCESS
- [x] New `VSFontHelperTest` (5 tests) covers the logical-family mapping, case insensitivity, the face-name regression, and that an unresolvable font keeps its requested name
- [x] Existing `StyleFontDefaultFamilyTest` + `GopFontRatioTest` still pass (15 tests total, 0 skipped)
- [ ] Export the reported viewsheet to PowerPoint with **match layout**; the column no longer wraps and matches the portal preview
- [ ] Unzip the `.pptx` and check `ppt/slides/slide1.xml`: table `<a:rPr>` carries `<a:latin typeface="Roboto"/>`, never `typeface="Roboto Bold"`, with `b="1"` carrying the weight
- [ ] Regression: wrap-text-enabled columns still wrap; bold/italic/underline still render; crosstab and selection-list titles unchanged
- [ ] Export the same viewsheet to Excel and confirm font names are unchanged for resolved fonts

## Known residual

If the machine opening the deck has no Roboto at all, this does not fully close the gap — the PPTX embeds no fonts, and POI's XSLF exposes no API for a `p:embeddedFontLst` part. Today's remedies are setting `default.font.family` (`defaults.properties:56`) to an Office-safe family, or installing the font on client machines. Worth a separate enhancement if drift remains.

Separately worth its own ticket: `VSFontHelper.getFontSize()` scales font size by 0.85 while `PPTVSUtil.PIXEL_TO_POINT` scales geometry by 0.75, so PPT text always occupies ~13% more of its box than the preview. That is the pressure making these near-fit cases fragile in the first place.

### DIFF
--- a/core/src/main/java/inetsoft/report/io/viewsheet/VSFontHelper.java
+++ b/core/src/main/java/inetsoft/report/io/viewsheet/VSFontHelper.java
@@ -21,6 +21,7 @@ import inetsoft.uql.viewsheet.internal.VSUtil;
 
 import java.awt.*;
 import java.util.Hashtable;
+import java.util.Locale;
 
 /**
  * Viewsheet font exporting helper.
@@ -35,6 +36,44 @@ public class VSFontHelper {
     */
    public static Font getDefaultFont() {
       return VSUtil.getDefaultFont();
+   }
+
+   /**
+    * Get the font family name to write out when exporting. The family (not the face)
+    * name is what office formats expect, since bold/italic are recorded separately -- a
+    * face name such as "Roboto Bold" cannot be resolved by the viewer, which then
+    * substitutes a font whose metrics differ from the ones used to lay the text out.
+    * See bug #75992.
+    * @param font the specified font.
+    * @return the font family to use when exporting.
+    */
+   public static String getExportFontFamily(Font font) {
+      if(font == null) {
+         return null;
+      }
+
+      String family = font.getFamily();
+
+      if(family == null || !exportFamilies.containsKey(key(family))) {
+         return family;
+      }
+
+      // the family collapsed to a java logical family, which means the server could not
+      // resolve the requested font. Write out the name that was actually asked for, so a
+      // client that does have it renders the same as the browser; only fall back to the
+      // mapping when the requested name is itself a logical name, since office cannot
+      // resolve those either.
+      String name = font.getName();
+
+      if(name != null && !exportFamilies.containsKey(key(name))) {
+         return name;
+      }
+
+      return exportFamilies.get(key(family));
+   }
+
+   private static String key(String family) {
+      return family.toLowerCase(Locale.ROOT);
    }
 
    /**
@@ -72,5 +111,18 @@ public class VSFontHelper {
    static {
       fontrates.put("dialog", DEFAULT_FONT_RATE);
       fontrates.put("comic sans ms", DEFAULT_FONT_RATE);
+   }
+
+   // java logical font family -> family that office applications can resolve. The
+   // pdf equivalent (FontManager) maps to the base-14 postscript names, which office
+   // does not have, so the values here are intentionally different.
+   private static final Hashtable<String, String> exportFamilies = new Hashtable<>();
+   static {
+      exportFamilies.put("dialog", "Arial");
+      exportFamilies.put("sansserif", "Arial");
+      exportFamilies.put("sans-serif", "Arial");
+      exportFamilies.put("serif", "Times New Roman");
+      exportFamilies.put("monospaced", "Courier New");
+      exportFamilies.put("dialoginput", "Courier New");
    }
 }
--- a/core/src/test/java/inetsoft/report/io/viewsheet/VSFontHelperTest.java
+++ b/core/src/test/java/inetsoft/report/io/viewsheet/VSFontHelperTest.java
@@ -1,0 +1,95 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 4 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Correctness

The root-cause analysis and fix are correct. I verified the core claim directly: for a logical Java font, `Font.getFontName()` really does append the style as a literal suffix (`new Font("Dialog", Font.BOLD, 10).getFontName()` → `"Dialog.bold"`, exactly as the new `styledFontYieldsFamilyNotFaceName` test asserts), which is indeed unresolvable as an OOXML `a:latin/@typeface` — explaining why the #4559 margin-only fix couldn't work.

`VSFontHelper.getExportFontFamily()`'s three-branch logic is sound:
1. Normal case — `family` isn't a Java logical name → return it as-is (unstyled runs of an installed font are byte-identical to before, since `getFontName() == getFamily()` for physical fonts).
2. `family` collapsed to a logical name (`Dialog`, `SansSerif`, …) but the originally-requested `Font.getName()` isn't itself logical → return the requested name, so a client that actually has that font renders correctly rather than being pinned to a generic substitute.
3. The requested name is itself a logical name → fall back to the Office-resolvable mapping (`Arial`/`Times New Roman`/`Courier New`).

This is a sensible, minimal-risk resolution order, and the new `VSFontHelperTest` covers all three branches plus case-insensitivity.

I also checked the `PoiExcelVSUtil` change for the cache-consistency concern the PR calls out: `findFont()`'s cache lookup keys on the font name string, and both the `findFont()` call (line 181) and the subsequent `xfont.setFontName()` call (line 208) were updated together to use `VSFontHelper.getExportFontFamily()`, so the two stay consistent and won't silently duplicate fonts in the workbook as the PR describes.

### CI failure investigation

The `build` check is `FAILURE`, but I pulled the actual Maven test log (job 96899307932) and confirmed the single test failure is `OrgLifecycleRepletRegistryIntegrationTest.delete_updateRepletRegistry_removalIsInMemoryOnly_notPersistedToDisk` — a pre-existing, unrelated test in the org-lifecycle/replet-registry persistence area that has nothing to do with this PR's font-export changes (`VSFontHelper`, `PPTValueHelper`, `PoiExcelVSUtil`). The test's own failure message documents it as a known gap ("documents a real gap: ... only appears correct in production because ... already removed/relocated the physical file a step earlier"), consistent with a flaky/order-dependent pre-existing issue rather than something this diff introduced.

- Good discipline in the PR body: it explicitly separates fixed scope from two known residual issues (missing font embedding in PPTX, and the pre-existing 0.85 vs 0.75 scale mismatch between `VSFontHelper.getFontSize()` and `PPTVSUtil.PIXEL_TO_POINT`) rather than trying to solve everything in one change.
- `PoiExcelVSExporter.java:2573` is correctly left untouched since it takes a CSS `font-family` string from HTML rich text, not a `java.awt.Font` — verified this is a genuinely different input type, not an inconsistency.

### Blocking issues for merge

2. **Failing check.** The `build` check reports `FAILURE` — however, per my investigation above, this failure is an unrelated pre-existing test (`OrgLifecycleRepletRegistryIntegrationTest`) and not caused by this PR's changes. Still, per policy any FAILURE-conclusion check blocks approval regardless of cause; re-running the build (or fixing/quarantining the unrelated flaky test) should clear this.

## Recommendation

Requesting changes — due to the two failing checks noted above, not due to any defect found in the fix itself, which is correct, well-tested, and appropriately scoped.

jshobe-inetsoft (approved):

PR #4723, opened 2026-08-21
Bug #76171, gate Google Chat sharing on the webhook URL property value

ngovus-inetsoft:
## Problem

`ShareController.getConfig` decides which entries of a dashboard's share menu are live. The Google Chat branch tested a string *literal* — the property's name — instead of reading the property:

```java
boolean googleChatEnabled = "true".equals(SreeEnv.getProperty("share.googlechat.enabled")) &&
   StringUtils.hasText("share.googlechat.url") && checkPermission("googlechat", user);
```

`StringUtils.hasText` of a non-empty literal is always `true`, so `share.googlechat.url` had no bearing on the result. The Slack branch five lines below is the same logic written correctly, which is what makes this a copy-paste slip rather than a design difference — `ShareSettingsService` reads and writes both channels' `enabled`/`url` pairs identically, and the Enterprise Manager form applies matching `urlRequired`/`invalidPrefix` validators to both.

## What an operator saw

With `share.googlechat.enabled` true and no webhook URL, the Google Hangouts Chat entry was live in the share menu. The failure arrived only on click: `sendGoogleChatMessage` re-reads `share.googlechat.url`, finds it blank, and throws `IllegalStateException("The Google Hangouts webhook URL is not set")`. Slack in the same state greys the entry and no user reaches the failure.

The Enterprise Manager Presentation form will not save the checkbox on with the URL field empty, so this state is not reachable from **Settings > Presentation**. It is reachable from **Settings > All Properties**, from `INETSOFTENV_SHARE_GOOGLECHAT_ENABLED` at first storage initialisation, and — most plausibly in practice — from an organization that enables Google Chat while inheriting the empty global URL from `defaults.properties`.

## Fix

Read the property, formatted to mirror the Slack branch:

```java
boolean googleChatEnabled = "true".equals(SreeEnv.getProperty("share.googlechat.enabled")) &&
   StringUtils.hasText(SreeEnv.getProperty("share.googlechat.url")) &&
   checkPermission("googlechat", user);
```

## Side effects

`getConfig` gates three callers: the `GET /api/share/config` endpoint, and both `sendGoogleChatMessage` and `sendSlackMessage` as their own precondition. The only behavior that changes is `googleChatEnabled` going from `true` to `false` in exactly the enabled-but-no-URL state. In that state the share attempt already failed — the exception moves from "The Google Hangouts webhook URL is not set" to "Sharing with Google Hangouts is not enabled", the same message Slack gives, and users no longer reach either because the entry is now inert. No install with a URL configured is affected.

## Note for the docs

`AllProperties.adoc` states that `share.googlechat.enabled` "Requires `share.googlechat.url` to also be set, otherwise the option remains hidden." This change makes the first half true. The "hidden" wording is still wrong, and wrong for the Slack entry too — `viewer-app.component.ts` maps both flags to `isShareHangoutsDisabled()`/`isShareSlackDisabled()`, so a disabled entry stays in the menu, greyed and inert. That correction belongs in the properties documentation, not here.

### DIFF
--- a/core/src/main/java/inetsoft/web/share/ShareController.java
+++ b/core/src/main/java/inetsoft/web/share/ShareController.java
@@ -70,7 +70,8 @@ public class ShareController {
          boolean facebookEnabled = "true".equals(SreeEnv.getProperty("share.facebook.enabled")) &&
             checkPermission("facebook", user);
          boolean googleChatEnabled = "true".equals(SreeEnv.getProperty("share.googlechat.enabled")) &&
-            StringUtils.hasText("share.googlechat.url") && checkPermission("googlechat", user);
+            StringUtils.hasText(SreeEnv.getProperty("share.googlechat.url")) &&
+            checkPermission("googlechat", user);
          boolean linkedinEnabled = "true".equals(SreeEnv.getProperty("share.linkedin.enabled")) &&
             checkPermission("linkedin", user);
          boolean slackEnabled = "true".equals(SreeEnv.getProperty("share.slack.enabled")) &&


jshobe-inetsoft (changes requested):
## Code Review

### Correctness

Correct, minimal, one-line fix for an obvious copy-paste slip: `StringUtils.hasText("share.googlechat.url")` was checking the truthiness of a string *literal* (always true for a non-empty string) instead of reading the property value via `SreeEnv.getProperty("share.googlechat.url")`. The fix now mirrors the Slack branch five lines below it exactly, which is the correct reference implementation per the PR's own analysis (I confirmed `ShareSettingsService` does read/write both channels' `enabled`/`url` pairs symmetrically).

The described behavior change is precisely scoped: `googleChatEnabled` flips from `true` to `false` only in the pre-existing broken state (enabled with no URL), where the share attempt already failed downstream in `sendGoogleChatMessage` — this just moves the failure earlier (menu entry now correctly greyed/inert) and changes the error message users would have hit had they clicked through, consistent with Slack's existing behavior in the same state. No install with a URL actually configured is affected.

- No existing `ShareControllerTest` exists to extend, and none was added here. Given this is exactly the kind of copy-paste-typo bug a simple unit test (`googleChatEnabled` false when URL is blank, true when set) would catch permanently, it'd be worth adding — but the fix is low-risk and obviously correct as a one-line change, so I wouldn't block merge solely on missing test coverage for a fix this small.

### Blocking issues for merge

## Recommendation

Requesting changes — solely because of the failing `claude-review` check. The fix itself is correct and low-risk; consider adding a small regression test for this class of typo if convenient, but it's not blocking.

jshobe-inetsoft (approved):

PR #4725, opened 2026-08-21
Bug #76172, remove the undocumented portal.customLogo.enabled gate

ngovus-inetsoft:
## Summary

`portal.customLogo.enabled` gated the **Default Logo** and **Default Favicon** controls on the EM Look and Feel page. It ships `false` and **nothing in the codebase writes it**, so on a stock install the controls were absent and the only way to reveal them was to type the property name into Settings > Properties — a page that is itself `hiddenForMultiTenancy: true`.

The issue was filed for review rather than as a defect, asking whether this was an oversight or a deliberately gated legacy feature. The evidence says oversight:

1. **The gate has no effect on rendering.** Every display-side consumer of `PortalThemesManager.hasCustomLogo()` / `isFaviconStyle()` ignores the property — `PortalThemeController:64`, `PortalController:131`, `EmNavBarController:94`, `AdminPageController:76`, `LoginController:86`, `SignupController:62,107`. A custom logo kept rendering in the portal and EM navbar with the property `false`, so the gate was never protecting a capability.
2. **A sibling control shows the intended shape.** `vsEnabled` gates the Viewsheet CSS checkbox with the identical template idiom and is hardcoded `.vsEnabled(true)`.

## Two defects the gate was masking

- **Silent loss of a configured logo.** `getModel()` forced `defaultLogo`/`defaultFavicon` to `true` while the gate was off, the Angular form initializes both to `true`, and `setModel()` never read the gate. Saving Look and Feel for any unrelated reason (tree sort order, fonts) therefore ran `setLogoStyle(false)` + `setLogo(null, …)` and cleared a configured logo. Now that `getModel()` reports real state, the round-trip is lossless.
- **Org scope discarded the gate anyway.** `getModel()` recomputes both flags from `getLogoEntries()`/`getFaviconEntries()`, overwriting the gated values. An org with no logo could never upload one, because the file chooser is gated on the checkbox value.

## Orphaned blob fix

Un-gating the checkbox makes the org-scope clear branch reachable for the first time (previously the checkbox was hidden while `defaultLogo` was `false`, so it could not be ticked), which exposed a path bug. `setLogo`/`setFavicon` store the map entry as a **full** path (`portal/<org>/logo.png`), but the clear branch passed it back through `space.delete(directory, file)` and `DataSpace.getPath()` concatenates → `portal/portal/<org>/logo.png`, orphaning the blob. Passing `null` as the directory uses the stored path as-is, matching how `setViewsheet` already behaves. The persisted map format is unchanged, so no migration is needed.

## Compatibility

- Installs that already set the property to `true`: behavior is unchanged — the gate was a no-op for them.
- Stock installs: the two checkboxes now appear, defaulted from real manager state. The file choosers were always conditioned on the checkbox value, not on the gate, so there is no visual regression.
- A stale key left in an existing `sree.properties` is inert; nothing reads it.

## Testing

- `mvnw compile -pl core` — BUILD SUCCESS; `ImmutableLookAndFeelSettingsModel` regenerates without the removed field.
- `LookAndFeelSettingsControllerTest` + `PresentationSettingsControllerTest` — 11/11 pass.
- `npm run test:em` — 365/365 pass; ESLint clean on the touched files.
- Manual verification against a running server has **not** been done. Worth exercising at review: (a) configure a custom logo, change only *Sort Repository Tree*, Apply, and confirm the logo survives; (b) in org scope, upload a logo, tick Default Logo, Apply, and confirm the blob is removed from the data space.

## Follow-ups (not in this change)

- `vsEnabled` is hardcoded `true` at `LookAndFeelService.java:136` and its template `@if` wrappers are equally dead.
- `hasCustomLogo()` guards with `!Tool.isEmptyString(logo)` but `isFaviconStyle()` has no equivalent guard, so `faviconStyle=true` with an empty path renders an unsatisfiable required-file validator.
- Logo/favicon extension validation is client-side only; the server accepts any extension, and `PortalThemeController.getIcon` then rejects anything outside gif/jpg/png/bmp/ico.


jshobe-inetsoft (changes requested):
## Code Review

### Correctness

This is a thorough, well-verified fix. I independently confirmed the two claims that matter most:

1. **The gate never actually protected the rendering path.** The PR body lists six display-side consumers of `hasCustomLogo()`/`isFaviconStyle()`; the removal of `customLogoEnabled` from `LookAndFeelService.getModel()` is a clean no-op for rendering and only changes whether the two checkboxes are visible/settable in the EM UI.
2. **The orphaned-blob fix is correct.** I traced `addLogoEntry(orgID, directory + "/" + logoName)` (line 540) to confirm the map stores a path already prefixed with `directory`, then checked `DataSpace.getPath(dir, file)` — it concatenates `dir + "/" + file` whenever `dir` is non-empty. So the old `space.delete(directory, file)` call really did produce `portal/portal/<org>/logo.png` for the org-scope clear branch, and the fix (`space.delete(null, file)`, treating `file` as the already-complete path) is exactly right. Same reasoning applies identically to the favicon path.

### Consistency check

I grepped for every reference to `customLogoEnabled`/`portal.customLogo.enabled` across `core/` and `web/` (excluding generated Immutables sources, which regenerate automatically) and confirmed the PR removes it consistently end-to-end: `LookAndFeelService.java`, the `LookAndFeelSettingsModel` interface, the Angular model/template, `properties-tool.ts`'s `boolean_options` list, `defaults.properties`, and the internal test-docs matrix note. No test file (Java or TypeScript) references the removed field, which is consistent with the PR's claimed 11/11 and 365/365 passing test runs — there's nothing left dangling that would have silently broken compilation.

- Good, honest scoping: the PR explicitly calls out three follow-up issues it deliberately didn't fix (hardcoded `vsEnabled`, missing empty-path guard on `isFaviconStyle()`, client-only extension validation) rather than scope-creeping them in.
- The PR is transparent that manual server verification of the two suggested scenarios (logo survives an unrelated save; org-scope clear removes the blob) was not done — worth doing before merge given this touches a real data-loss bug (silent logo clearing) and a storage cleanup path, but the static analysis here strongly supports both fixes being correct.
- No backward-incompatibility or security concerns. A stale `portal.customLogo.enabled=true` left in an existing `sree.properties` is correctly inert (nothing reads it after this change).

### Blocking issues for merge

## Recommendation

Requesting changes — solely because of the failing `claude-review` check. The fix itself is correct, well-reasoned, and verified consistent across the codebase; recommend doing the manual server verification the PR author already flagged before merging, given the data-loss/orphaned-blob nature of the underlying bugs.

jshobe-inetsoft (approved):

PR #4726, opened 2026-08-21
Bug #76176, pin portal.themes.file for the manager's lifetime

ngovus-inetsoft:
## Problem

`PortalThemesManager` resolved `portal.themes.file` in two places with different lifetimes:

- `loadThemes()` — `@PostConstruct`, so once per bean
- `saveUnderLock()` — on every write

Nothing re-ran the load when the property changed. After an admin changed it on a running server, the manager kept serving the configuration loaded from the old file while every save went to the new one. The two diverged silently, and it was the old file that stayed in effect. A restart then switched to whatever the new file had accumulated, possibly a partial configuration written across several saves.

Two things that are easy to miss:

- **The property is reachable at runtime.** There is no dedicated UI field and the key is absent from `defaults.properties`, but the generic EM property editor (Settings → Properties, `PropertiesController`) lists and edits whatever `SreeEnv.getProperties()` returns, unfiltered.
- **The mismatch also stranded the change listener.** `DataChangeListenerManager` keys registrations on `(space, dir, file, listener)`, so `saveUnderLock()`'s `removeChangeListener` named a file that was never registered and matched nothing — leaving the old file's listener in place and adding a second one on the new file. The old file then kept driving `reset(); loadThemes();`.

The bundled fallback made it quieter still: when the named file is missing, `loadThemes()` falls back to the classpath copy and writes it out under the configured name, so pointing the property at an unwritten name produces a fresh default configuration rather than an error.

## Fix

Resolve the property once in `getThemesFile()` and use it from both the load and the save path, so those two and the change-listener registration can never name different files.

A change to the property now takes effect on the next restart — which is how the sibling `CustomThemesManager` already treats `custom.themes.file` (resolved in its constructor).

Live reload was considered and left out of scope. It would need a hook in `PropertiesController` for the local node *plus* a `PropertiesEngine.addPropertyChangeListener` for peers, because `PropertiesEngine.save()` deregisters its own KV-storage listener during the write, so the node making the change never fires its own property listener.

## Tests

New `PortalThemesManagerTest` (3 tests, `mockStatic(SreeEnv)`):

- `save_afterPropertyChanged_writesToTheFileThatWasLoaded`
- `save_afterPropertyChanged_unregistersTheListenerItRegistered`
- `propertyUnset_usesTheDefaultFileName`

Both regression tests were verified to fail against the unfixed code — the save test wrote `portalthemes-b.xml` instead of `-a.xml`, and the listener test caught the stranding.

`mvnw test -pl core -Dtest='*Theme*,*LookAndFeel*,*Presentation*,PermissionMatrixActionsS8Test,*Portal*'` → 151 tests, BUILD SUCCESS.

## Notes for the reviewer

Behavior is unchanged unless `portal.themes.file` is edited on a running server; with the property unset both paths resolve `portalthemes.xml` exactly as before. Every writer reaches the file only through `manager.save()` (`LookAndFeelService`, `PresentationLoginBannerSettingsService`, `PortalIntegrationViewSettingsService`, `WelcomePageService`), and none resolve the path themselves.

Three pre-existing weaknesses in this class were found while investigating and deliberately **not** touched, to keep the fix minimal:

- `clear()` is dead code (no callers) that reloads without `reset()` and without the distributed lock.
- `reset()` does not clear `cssEntries` / `faviconEntries` / `logoEntries` / `welcomePageEntries` / `userFontFaces`, so a remote change that *removes* an element leaves stale per-identity entries.
- The no-arg constructor passes `cluster = null`, so `save()` and the change listener NPE in any non-Spring instantiation that reaches them.

A companion enterprise PR updates `claude/theme.md` with the load/save/listener invariant.

### DIFF
--- a/core/src/main/java/inetsoft/sree/portal/PortalThemesManager.java
+++ b/core/src/main/java/inetsoft/sree/portal/PortalThemesManager.java
@@ -564,7 +564,7 @@ public class PortalThemesManager implements XMLSerializable, AutoCloseable {
     * Must only be called by save(), which acquires the lock first.
     */
    private void saveUnderLock() {
-      String name = SreeEnv.getPath("portal.themes.file", "portalthemes.xml");
+      String name = getThemesFile();
       DataSpace space = dataSpace;
 
       try {
@@ -943,12 +943,32 @@ public class PortalThemesManager implements XMLSerializable, AutoCloseable {
    }
 
    /**
+    * Get the data space file that holds the portal themes. The name is resolved from the
+    * portal.themes.file property on first use and pinned for the lifetime of this manager.
+    * The load path, the save path and the change listener registration must all name the
+    * same file; if the property were re-read on each save, changing it on a running server
+    * would leave the loaded configuration coming from the old file while every subsequent
+    * save went to the new one, and would strand the change listener on the old file. A
+    * change to the property therefore takes effect on the next restart.
+    */
+   private String getThemesFile() {
+      String name = themesFile;
+
+      if(name == null) {
+         // a race here is benign, getPath() is deterministic
+         name = themesFile = SreeEnv.getPath("portal.themes.file", "portalthemes.xml");
+      }
+
+      return name;
+   }
+
+   /**
     * Build up the portal manager by parse a .xml file.
     */
    @PostConstruct
    public void loadThemes() {
       DataSpace space = dataSpace;
-      String name = SreeEnv.getPath("portal.themes.file", "portalthemes.xml");
+      String name = getThemesFile();
       boolean saveFile = false;
 
       try {
@@ -1068,6 +1088,7 @@ public class PortalThemesManager implements XMLSerializable, AutoCloseable {
    private PortalWelcomePage welcomePage;
    private List<PortalTab> portalTabs = new ArrayList<>();
    private String copyright;
+   private volatile String themesFile;
    private final DataChangeListenerManager dmgr = new DataChangeListenerManager();
 
    private static final Logger LOG = LoggerFactory.getLogger(PortalThemesManager.class);
--- a/core/src/test/java/inetsoft/sree/portal/PortalThemesManagerTest.java
+++ b/core/src/test/java/inetsoft/sree/portal/PortalThemesManagerTest.java
@@ -1,0 +1,149 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Correctness

The core fix is right: `getThemesFile()` resolves `portal.themes.file` once and lazily caches it in a `volatile` field, and both `loadThemes()` and `saveUnderLock()` now go through it, so the load path, save path, and `DataChangeListenerManager` registration can never name different files. The "a race here is benign, getPath() is deterministic" comment is accurate — concurrent first-callers compute the same value and a `volatile String` assignment can't tear, so no synchronization is needed beyond that.

I traced one subtlety worth flagging (not fixed, just documented, in the spirit of the PR's own "notes for reviewer" section): `portal.themes.file` is *not* in `PropertiesEngine.EXCLUDED_ORG_PROPERTIES`, so `SreeEnv.getPath()` → `getProperty(name, def)` → `fixPropertyName()` → `useAvailableOrgProperty()` will resolve an org-scoped override (`inetsoft.org.<id>.portal.themes.file`) if one exists *and* a principal is bound to the calling thread when `getThemesFile()` first runs. Since `PortalThemesManager` is a plain `@Service` singleton (one instance per node, not per-org) and `loadThemes()` is `@PostConstruct` — which fires during context startup with no request-bound principal — the cached value in practice will almost always end up being the base (non-org-scoped) file name, which matches this class's actual design (one shared XML file holding per-identity `cssEntries`/`logoEntries`/`faviconEntries` maps, not one file per org). So this isn't a regression: the pre-fix code had the equivalent-or-worse problem (whichever org's thread saved *last* silently became the effective file for everyone), and an org-scoped override of this particular property was already a semantically murky configuration before this PR touched it. Worth a one-line comment on `themesFile` noting that the cached value intentionally ignores any org-scoped override of the property, so nobody "fixes" it later by innocently re-reading the property per-call and reintroducing the exact bug this PR closes.

### Tests

The three new `PortalThemesManagerTest` cases target the described defect directly and are well-isolated (`mockStatic(SreeEnv)`, no Spring context needed). Verifying that both regression tests fail against the unfixed code, per the PR body, is exactly the right level of rigor. `propertyUnset_usesTheDefaultFileName` guards the common case explicitly, which is good since it's the path every install without a customized property takes.

- The PR is transparent and well-scoped: it explicitly calls out three pre-existing weaknesses in this class (`clear()` dead code, `reset()` not clearing entry maps, the no-arg constructor's `cluster = null`) and deliberately leaves them alone rather than scope-creeping the fix. Good judgment.
- No backward-incompatibility or security concerns; behavior is unchanged when `portal.themes.file` is unset (the common case), and the only behavior change (a mid-run property edit now requires a restart to take effect) is explicitly documented and matches the sibling `CustomThemesManager`'s existing contract.

### Blocking issues for merge

## Recommendation

Requesting changes — solely because of the failing `claude-review` check. The fix itself is correct, minimal, and well-tested; I'd suggest considering the org-scoped-property caveat above for a follow-up comment, but it is not a regression and not blocking.

jshobe-inetsoft (approved):

PR #4727, opened 2026-08-21
Bug #76177, parse schedule.task.timeout consistently

ngovus-inetsoft:
## Problem

Every reader of `schedule.task.timeout` parsed it with a bare `Long.parseLong(SreeEnv.getProperty(...))`. Two caught `NumberFormatException` and fell back to ten minutes; the rest did not, so a single mistyped value degraded quietly on some paths and threw on others.

There are **nine** readers in total — four here, five in the enterprise `integration/` cloud job factories. Split was 2 guarded / 7 unguarded:

| Reader | Before | Consequence of a bad value |
|---|---|---|
| `ScheduleTask.doRun` | guarded, silent | silent fallback |
| `MVAction.createMV` | guarded, silent | silent fallback |
| `Scheduler.runTask` | **unguarded** | unchecked NFE — Run Now fails for any task whose status is `STARTED` |
| `ScheduleTaskCloudJob` ctor | **unguarded** | thrown from the **constructor**, so Quartz cannot instantiate the job; the trigger misfires and `execute()` never runs |

The constructor case is the least recoverable: a job that cannot be constructed does not run, and the failure is not attributable to a property value from where it surfaces.

Not reachable while `defaults.properties` supplies a number, so this is latent. It becomes reachable through EM **Settings → All Properties**, `INETSOFTENV_SCHEDULE_TASK_TIMEOUT` at first storage initialisation, or a direct property-store edit — none of which validate. `PropertiesController.editProperty` trims the value and persists it verbatim with no type check.

## Change

One accessor, one behaviour. `ScheduleTask.getTaskTimeout()` logs a warning naming the property and the offending value, then falls back to `DEFAULT_TASK_TIMEOUT` (600000, matching `defaults.properties:275`). All four readers here route through it; exactly one `SreeEnv.getProperty("schedule.task.timeout")` call now exists in the tree.

`ScheduleTaskCloudJob` reads the timeout in `execute()` rather than its constructor, so an unusable value can no longer prevent construction. The field drops `final`: safe because it is written as the first statement of `execute()` and read only later on that same thread — `interrupt()` and the `MessageListener` never touch it, and Quartz instantiates the job per firing. Side benefit: the value is re-read per execution, so a config change takes effect without a restart.

### On zero and negative

Deliberately returned as configured rather than normalised, because the call sites do not agree on a meaning. `ScheduleTask.doRun`, `MVAction.createMV`, and `ScheduleTaskCloudJob.execute` all guard `if(timeout > 0)` and fall through to an unbounded wait. `Scheduler.runTask` does not: with `timeout <= 0`, `(currTime - startTime) < timeout` is always false, so every `STARTED` task is treated as stale and re-triggerable — the inverse of "no timeout". The javadoc records this rather than implying a uniform contract. Normalising it would change existing runtime behaviour, which is out of scope here.

## Tests

`ScheduleTaskTest` gains coverage for the configured value, the fallback (`null`, `abc`, `""`, `" "`, `600000ms`, `600,000`, `1e5`), the warning itself (WARN level, message contains both property name and offending value), and drift between `DEFAULT_TASK_TIMEOUT` and `defaults.properties`. `ScheduleTaskCloudJobTest` gains a guard that construction survives an unusable property.

Each new test was verified to be load-bearing by reverting the behaviour it covers and confirming it fails:

| Revert | Failure |
|---|---|
| throwing parse back in the constructor | `NumberFormatException: Cannot parse null string` |
| `LOG.warn` deleted | `the fallback must report itself exactly once ==> expected: <1> but was: <0>` |
| `defaults.properties` set to `300000` | `expected: <300000> but was: <600000>` |

366 tests pass across `inetsoft.sree.schedule.**` (up from 364). The pre-existing `MVActionTest` cases that stub this property pass unchanged through the new indirection.

## Related

The five `integration/` cloud job factories share this defect and are fixed in the enterprise PR, which depends on the accessor added here. Merge this first.

## Out of scope

`schedule.concurrency` has the identical unguarded pattern at `Scheduler.java:544` (during `startScheduler`, so a bad value stops the scheduler from starting) and `TaskBalancer.java:212`; `security.cache.interval` at `AbstractAuthenticationProvider.java:238`. Type validation in `PropertiesController.editProperty` would fix the whole class at configuration time and is the more valuable follow-up. Filing separately.

### DIFF
--- a/core/src/main/java/inetsoft/sree/schedule/MVAction.java
+++ b/core/src/main/java/inetsoft/sree/schedule/MVAction.java
@@ -20,7 +20,6 @@ package inetsoft.sree.schedule;
 import inetsoft.mv.*;
 import inetsoft.mv.fs.FSService;
 import inetsoft.mv.fs.internal.ClusterUtil;
-import inetsoft.sree.SreeEnv;
 import inetsoft.sree.internal.Mailer;
 import inetsoft.sree.internal.cluster.Cluster;
 import inetsoft.sree.internal.cluster.SimpleMessage;
@@ -256,14 +255,7 @@ public class MVAction implements AssetSupport, Cloneable, XMLSerializable, Cance
          boolean exists = mv.hasData();
 
          if(createInScheduler) {
-            long timeout;
-
-            try {
-               timeout = Long.parseLong(SreeEnv.getProperty("schedule.task.timeout"));
-            }
-            catch(NumberFormatException e) {
-               timeout = 600000L;
-            }
+            long timeout = ScheduleTask.getTaskTimeout();
 
             try {
                MVCallable creator = new MVCallable(mv, principal);
--- a/core/src/main/java/inetsoft/sree/schedule/ScheduleTask.java
+++ b/core/src/main/java/inetsoft/sree/schedule/ScheduleTask.java
@@ -51,6 +51,37 @@ import java.util.stream.Stream;
  * @author InetSoft Technology Corp
  */
 public class ScheduleTask implements Serializable, Cloneable, XMLSerializable {
+   /**
+    * The default schedule task timeout, in milliseconds. Matches the value shipped in
+    * defaults.properties; used when schedule.task.timeout is missing or not a number.
+    */
+   public static final long DEFAULT_TASK_TIMEOUT = 600000L;
+
+   /**
+    * Gets the configured schedule.task.timeout, in milliseconds. A missing or non-numeric
+    * value logs a warning and falls back to DEFAULT_TASK_TIMEOUT.
+    *
+    * <p>The value is returned as configured; no meaning is imposed on zero or a negative
+    * number here, because the call sites do not agree on one. The three that wait on a
+    * future or a latch (doRun below, MVAction.createMV, ScheduleTaskCloudJob.execute) treat
+    * 0 or less as "wait without a timeout", whereas Scheduler.runTask uses the value only as
+    * a staleness window and does not special-case it.
+    *
+    * @return the configured task timeout in milliseconds.
+    */
+   public static long getTaskTimeout() {
+      String value = SreeEnv.getProperty("schedule.task.timeout");
+
+      try {
+         return Long.parseLong(value);
+      }
+      catch(NumberFormatException e) {
+         LOG.warn("Invalid schedule.task.timeout value \"{}\", using the default of {} ms",
+                  value, DEFAULT_TASK_TIMEOUT);
+         return DEFAULT_TASK_TIMEOUT;
+      }
+   }
+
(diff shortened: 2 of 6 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Correctness

This is a solid, well-scoped fix. It consolidates four separate `Long.parseLong(SreeEnv.getProperty("schedule.task.timeout"))` call sites (two silently guarded, two not) into a single `ScheduleTask.getTaskTimeout()` accessor that logs a `WARN` naming both the property and the offending value before falling back to `DEFAULT_TASK_TIMEOUT` (600000, matching `defaults.properties`).

I specifically checked the `ScheduleTaskCloudJob` change, since moving `timeout` out of the constructor and into `execute()` also drops the field's `final` modifier — a real thread-safety-relevant change. Verified by grep: `timeout` is written exactly once per `execute()` call (as its first statement), and the only other reads are later in that same `execute()` method (`latch.await(timeout, ...)`). `interrupt()` never touches `timeout`, and Quartz instantiates a fresh job object per firing for a stateless job like this one, so there's no cross-thread or cross-firing visibility hazard. The description's own reasoning for dropping `final` checks out.

The "zero/negative is returned as-configured, not normalized" decision is correctly justified and documented — `Scheduler.runTask`'s `(currTime - startTime) < timeout` genuinely has inverted semantics from the other three wait-based call sites for `timeout <= 0`, and normalizing here would silently change existing runtime behavior outside this bug's scope. Good instinct to document rather than "fix" that inconsistency as a drive-by.

### Tests

Thorough and load-bearing — the PR body documents reverting each behavior and confirming the corresponding test fails, which is good practice. The `getTaskTimeoutDefaultTracksShippedProperty` test (reading `defaults.properties` rather than hardcoding `600000` twice) is a nice touch that prevents the constant and the shipped default from silently drifting apart. `ScheduleTaskCloudJobTest.constructorSurvivesAnUnusableTimeoutProperty` directly targets the most severe of the four original defects (an NFE thrown from a Quartz job constructor, which prevented instantiation entirely).

- `MVAction.java`'s removed `import inetsoft.sree.SreeEnv;` is correctly dropped since the direct call is replaced by `ScheduleTask.getTaskTimeout()` — verified no other use of `SreeEnv` remains in that file's diff context.
- No backward-incompatibility or security concerns; this only changes fallback/logging behavior for an already-latent misconfiguration path.

### Blocking issues for merge

## Recommendation

Requesting changes — solely because of the failing `claude-review` check; I found no issues with the code itself, which is correct, well-tested, and appropriately scoped. Please resolve/re-run that check before merge.

jshobe-inetsoft (approved):

PR #4728, opened 2026-08-22
Add catch-all exception handler to WizControllerErrorHandler

larryliang-inetsoft:
## Summary
- `WizControllerErrorHandler` (`@ControllerAdvice(basePackages = "inetsoft.web.wiz")`) only maps 7 specific exception types. Any other exception thrown by a wiz controller (viewsheet, script, worksheet, pairing, binding) fell through to Tomcat's own default HTML error page, which ignores the client's `Accept: application/json` header — leaking a raw, multi-KB HTML stack-trace dump to the calling MCP agent instead of a clean JSON error.
- Live-confirmed trigger: `get_binding`, `list_bindable_fields` (scoped and unscoped), and `set_table_source` all raw-500 on a crosstab assembly with no source bound yet. Also explains L4 Finding 5 (raw Tomcat 500 on an out-of-range legend index) and L6's raw `HttpRequestMethodNotSupportedException` note.
- Adds a last-resort `@ExceptionHandler(Exception.class)` that logs the exception and returns a plain JSON 500 (`{"error": "..."}`). Ordered last relative to the existing handlers — Spring dispatches to the most specific matching handler, so this only catches what nothing else already does.

## Test plan
- [x] `./mvnw test -pl core -Dtest=WizControllerErrorHandlerTest` — 9/9 pass, including the new `anUnrecognizedExceptionMapsToInternalServerErrorWithJsonBody` case (bare `NullPointerException` -> HTTP 500 JSON body containing "unexpected server error", not HTML)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/WizControllerErrorHandler.java
+++ b/core/src/main/java/inetsoft/web/wiz/WizControllerErrorHandler.java
@@ -227,5 +227,25 @@ public class WizControllerErrorHandler {
       return new ResponseEntity<>(payload, null, HttpStatus.CONFLICT);
    }
 
+   /**
+    * Maps any exception none of the handlers above recognize to a plain JSON 500.
+    *
+    * <p>Without this, an unanticipated exception (an NPE deep in a binding/viewsheet service, for
+    * example) fell through to Tomcat's own default HTML error page, which ignores the client's
+    * {@code Accept: application/json} header — leaking a raw, multi-KB HTML/stack-trace dump to
+    * the calling agent instead of a clean JSON error. Ordered last: Spring dispatches to the most
+    * specific matching handler, so this only catches what nothing else already does.
+    */
+   @ExceptionHandler(Exception.class)
+   public ResponseEntity<Map<String, String>> handleUnexpected(Exception e) {
+      LOG.error("Unexpected error in wiz agent request", e);
+
+      Map<String, String> payload = new HashMap<>();
+      payload.put("error",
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
**Review verdict: approve (code is sound).** Note: could not submit a formal GitHub review — GitHub blocks self-approval since this PR is authored by the same account I'm running as.

Reviewed the diff and the full `WizControllerErrorHandler.java` for context.

- The new `@ExceptionHandler(Exception.class)` is correctly the last handler and, per Spring's dispatch semantics, only ever catches what the 7 more-specific handlers (including `SecurityException` -> 403) don't already claim — declaration order doesn't matter for that resolution, so there's no risk of it masking the auth/authz handler above it. Checked for over-reach specifically and don't see any exception type here that needs to propagate unhandled to another layer.
- Response body is a generic, non-leaking message; stack trace only goes to the server log via `LOG.error`. No new info disclosure.
- New test (`anUnrecognizedExceptionMapsToInternalServerErrorWithJsonBody`) matches the claimed fix.

**CI note (not a code issue, flagging for the merge decision):**
- `build` failed on a single unrelated test: `WorksheetAgentControllerTest.importCsvDetectTypeFalseMakesEveryColumnString` (CSV type-detection assertion). Nothing in this diff touches CSV import or that test file — looks like a pre-existing/flaky failure, not caused by this change. Worth a re-run to confirm before merging.
- `claude-review` failed due to an empty `ANTHROPIC_API_KEY` in that workflow run (infra/tooling issue, 0 turns executed) — not a review finding.

PR #4729, opened 2026-08-22
fix(wiz): verify the runtime, not just the pairing record, in session()

larryliang-inetsoft:
## Summary
- `ViewsheetAgentController.session()` (`GET /api/wiz/v1/agent/script/{token}/session`) resolved only the in-memory `JoinSession` pairing record via `resolveSession()`, never dereferencing the `RuntimeViewsheet` its `runtimeId` names. A pairing record can stay valid (present, unexpired, owned) long after the runtime it points at has been evicted independently, so this status probe reported the session `valid` while every other endpoint on this controller (`targets`, `readScript`, etc., via `ScriptEditService.resolve`) already reported it expired.
- Live-confirmed impact: right after an MCP server reconnect, `status` reported a session as `valid:true` while the very next call against the same session immediately failed with "runtime not found or expired".
- Fix reuses the existing `editService.resolve(sessionToken, user)` call — the same pattern `targets()` already follows two methods below — so `session()` now also verifies the runtime through `SheetRuntimeAccess.getSheetForPairing`, surfacing a 404 `SESSION_EXPIRED` once the runtime is gone instead of a stale 200. No new bean injection needed since `ScriptEditService` was already a constructor dependency of this controller.
- Scoped to `session()` only — `resolveSession()`'s pairing-only contract is left untouched, since other callers (`requirePaneScope`, `targets`) rely on it and separately dereference the runtime elsewhere already.
- No `stylebi-wiz` (plugin) change needed — the plugin's existing 404→"expired" classification in `classifyProbeError()` already handles this correctly once the backend returns 404.

## Test plan
- [x] Added `sessionRefusesAPairingRecordWhoseRuntimeIsGone` to `ViewsheetAgentControllerTest`, covering a live pairing record whose `runtimeId` no longer resolves to a real runtime.
- [x] `./mvnw -pl core -Dtest=ViewsheetAgentControllerTest test -o` — 12/12 pass (11 pre-existing + 1 new).

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/script/ViewsheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/script/ViewsheetAgentController.java
@@ -133,12 +133,19 @@ public class ViewsheetAgentController {
     * RIGHT NOW" and get a live answer, independent of whatever it minted or cached at connect
     * time. Requires no pane-scope check of its own: a session is always entitled to know its own
     * scope, whole-sheet or pane-scoped alike.
+    *
+    * <p>Also verifies the runtime the pairing names still exists (mirrors {@link #targets}) --
+    * {@link #resolveSession} alone only checks the in-memory pairing record, not whether the
+    * {@link RuntimeViewsheet} it points at has since been evicted independently. Without this, a
+    * pairing that outlives its runtime read back as {@code valid} here while every other endpoint
+    * on this controller correctly reported the session as expired.
     */
    @GetMapping("/api/wiz/v1/agent/script/{sessionToken}/session")
    public SessionInfo session(@PathVariable String sessionToken, Principal user)
       throws PairingException
    {
       requireEnabled();
+      editService.resolve(sessionToken, user);
       JoinSession session = resolveSession(sessionToken, user);
       return new SessionInfo(session.runtimeId(), session.sheetType().name().toLowerCase(),
                              session.editorContext(), session.followFocusEnabled());
--- a/core/src/test/java/inetsoft/web/wiz/script/ViewsheetAgentControllerTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/script/ViewsheetAgentControllerTest.java
@@ -292,6 +292,42 @@ class ViewsheetAgentControllerTest {
       assertEquals(PANE_RUNTIME_ID, after.runtimeId());
    }
 
+   /**
+    * A pairing record can stay valid (present, unexpired, owned) in {@code SheetSessionService}
+    * long after the {@link RuntimeViewsheet} its {@code runtimeId} names has been evicted
+    * independently -- {@link #resolveSession} only checks the pairing record. {@code session()}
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft:
**Review verdict: approve (self-review-blocked, cannot use `gh pr review --approve` on own PR — posting as comment instead)**

Reviewed the diff, the touched service (`ScriptEditService.resolve`), and the plugin's consumer of this endpoint (`sessionProbe.ts`).

**Fix verified correct:**
- `ScriptEditService.resolve()` really does dereference the runtime: `runtimeAccess.getSheetForPairing(SheetType.VIEWSHEET, session.runtimeId(), agent)` throws `PairingException(SESSION_EXPIRED, ...)` when the runtime is gone (core/src/main/java/inetsoft/web/wiz/script/ScriptEditService.java:65-71).
- The added `editService.resolve(sessionToken, user)` call in `session()` is placed before the existing pairing-only `resolveSession()` check and is byte-for-byte the same ordering `targets()` already uses two methods below — consistent with existing convention, not a new pattern.
- `SESSION_EXPIRED` maps to HTTP 404 in this controller's `@ExceptionHandler` (line 481), which matches `classifyProbeError()`'s `HTTP_404 -> "expired"` on the plugin side (plugin/shared/core/src/state/sessionProbe.ts) — no plumbing mismatch.

**No regression for legitimate callers:** checked `sessionProbe.ts` — the plugin's `status` tool calls this exact `/session` endpoint specifically as a liveness probe and already expects a 404 to mean "expired." No caller anywhere wants `session()` to report `valid:true` while the runtime is briefly gone; this fix makes the endpoint match the intent already documented in `probeSession`'s own docstring.

**Regression test is meaningful:** `sessionRefusesAPairingRecordWhoseRuntimeIsGone` wires a real `ScriptEditService` against a mocked `SheetRuntimeAccess` that throws `PairingException(SESSION_EXPIRED, ...)` from `getSheetForPairing`, with `SheetSessionService` still returning a valid pairing record — exactly the "pairing alive, runtime gone" seam the bug was in, and it would fail against the pre-fix code. The pre-existing `sessionReportsTheLiveEditorContextOnEveryCall` test still passes unaffected since it uses a bare `mock(ScriptEditService.class)` whose unstubbed `resolve()` returns null, which `session()` discards.

Scope is minimal: one call added, one doc comment, one new test. No unrelated refactors.

**CI:** `build` and `claude-review` show failing, but neither is caused by this PR:
- `build`'s only failure is `WorksheetAgentControllerTest.importCsvDetectTypeFalseMakesEveryColumnString` (a file this PR never touches). Confirmed the identical failure occurs in PR #4731 (an unrelated, already-merged named-group PR) on this same base branch — a pre-existing/flaky test.
- `claude-review` fails with "Claude encountered an error after 0s" — same infra error reproduces on #4731 too, not a real review signal.
- Semgrep (both) and CLA pass. PR is mergeable, no conflicts.

No blocking issues found.

PR #4734, opened 2026-08-22
Fix locale-dependent CSV numeric type detection flake in AssetUtil

larryliang-inetsoft:
## Root cause

`AssetUtil.NFMT`/`PFMT`/`CFMT`/`C2FMT` (used by `AssetUtil.getType()` for CSV/Excel import-time numeric type detection) were declared `private static final` and built once, at whatever `Locale.getDefault()` was in effect the moment `AssetUtil` was first class-loaded in a JVM. If any earlier-running test in the same Surefire fork left a non-US default locale in place (e.g. via `Locale.setDefault(...)` without restoring it), these formatters got permanently poisoned for the rest of that fork's life. Depending on the locale's grouping separator, this can either produce a silently-wrong parsed value or (for e.g. French, whose grouping separator isn't `.`) cause the string to fail to fully parse, falling back to `STRING` type.

This caused `WorksheetAgentControllerTest.importCsvDetectTypeFalseMakesEveryColumnString` to fail intermittently and order-dependently in CI's Linux Surefire run (`filesystem` run order differs from Windows), while always passing locally on Windows and in isolation — blocking merges on several unrelated PRs (#4732 and others). Confirmed this is not a real regression: both the test and the underlying CSV-import behavior (`0fc080828`) are correct as committed.

## Fix

Pin `NFMT`/`PFMT`/`CFMT`/`C2FMT` explicitly to `Locale.US` (`core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java:3496-3507`) instead of relying on the JVM default at class-load time. These formatters exist to recognize StyleBI's own canonical numeric/currency/percent text representations during import-time type sniffing — not to render numbers back to a user in their locale — so they should never have varied with the runtime default locale.

Left `AssetUtil.getNumberFormat()` (`numFmt`, line ~1355) alone — it's a distinct field used by `format()` for user-facing value rendering/export and is configurable via the `format.number` sree property, which is a legitimately locale-aware, different concern from import-time type sniffing.

## Regression test

Added `core/src/test/java/inetsoft/uql/asset/internal/AssetUtilTest.java`, which sets `Locale.setDefault(Locale.FRANCE)` (a locale whose grouping separator isn't `.`, so it reproduces the actual parse failure rather than a silently-wrong-value pass), calls `AssetUtil.getType()` on `"299.99"` and `"$499.99"`, and asserts both are still detected as numeric. Restores the original default locale in `@AfterEach`. Verified this test fails without the fix and passes with it.

## Testing

- `./mvnw -pl core -o test -Dtest=AssetUtilTest` — 1/1 pass
- `./mvnw -pl core -o test -Dtest=WorksheetAgentControllerTest` — 50/50 pass (including the previously flaky test)
- Verified the new test fails without the fix (locale poisoning reproduced) and passes with it

## Note on the leaking test

Could not pin the exact test/call site that leaks a non-US default locale on CI's Linux runner (grepped `core/src/test` for `Locale.setDefault`; the two call sites found both already restore correctly in `@AfterAll`). This fix is independent of finding it — it closes the underlying fragility in `AssetUtil` regardless of which test (if any) is the culprit.

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java
+++ b/core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java
@@ -3495,10 +3495,17 @@ public class AssetUtil {
 
    public static boolean debug = false;
    private static DecimalFormat numFmt = null;
-   private static final DecimalFormat NFMT = new DecimalFormat("#,###.#");
-   private static final NumberFormat PFMT = NumberFormat.getPercentInstance();
-   private static final NumberFormat CFMT = new DecimalFormat("$#,##0.##;-$#,##0.##");
-   private static final NumberFormat C2FMT = new DecimalFormat("$#,##0.##;($#,##0.##)");
+   // these recognize StyleBI's own canonical numeric/currency/percent text
+   // representations during import-time type sniffing (not user-facing
+   // rendering), so they must stay pinned to Locale.US regardless of the
+   // JVM default locale in effect when AssetUtil is first class-loaded.
+   private static final DecimalFormat NFMT =
+      new DecimalFormat("#,###.#", DecimalFormatSymbols.getInstance(Locale.US));
+   private static final NumberFormat PFMT = NumberFormat.getPercentInstance(Locale.US);
+   private static final NumberFormat CFMT =
+      new DecimalFormat("$#,##0.##;-$#,##0.##", DecimalFormatSymbols.getInstance(Locale.US));
+   private static final NumberFormat C2FMT =
+      new DecimalFormat("$#,##0.##;($#,##0.##)", DecimalFormatSymbols.getInstance(Locale.US));
 
    private static final String REPOSITORY_KEY = AssetUtil.class.getName() + ".repository";
    private static final String DESIGN_REPOSITORY_KEY =
--- a/core/src/test/java/inetsoft/uql/asset/internal/AssetUtilTest.java
+++ b/core/src/test/java/inetsoft/uql/asset/internal/AssetUtilTest.java
@@ -23,7 +23,12 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
+import java.lang.reflect.Field;
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
 import java.text.Format;
 import java.util.HashMap;
 import java.util.Locale;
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft:
Addressed the reviewer's finding: the original regression test's end-to-end `AssetUtil.getType()` parse assertion relied on this test class being the one that first class-loads `AssetUtil` in the shared full-suite JVM, which Surefire run order doesn't guarantee. Replaced the sole reliance on that with a reflection-based `@ParameterizedTest` (`formattersAreLocaleInvariant`) that inspects `NFMT`/`PFMT`/`CFMT`/`C2FMT`'s actual embedded `DecimalFormatSymbols` directly -- order-independent by construction, since it checks whatever state the fields already hold rather than trying to control when they're built. Kept the original end-to-end test as a secondary smoke test.

Verified both red (fails on pre-fix source) and green (passes on fixed source) under a genuinely poisoned starting locale, by temporarily forcing `core/pom.xml`'s argLine to `-Duser.language=fr -Duser.country=FR` (simulating the real leaked-locale-at-class-load scenario) rather than relying on incidental single-test-class run order -- this is what caught the original test's blind spot in the first place. Reverted the pom.xml change before pushing; it was verification-only.

Re-ran:
- `./mvnw -pl core -o test -Dtest=AssetUtilTest` -- 5/5 pass
- `./mvnw -pl core -o test -Dtest="AssetUtilTest,WorksheetAgentControllerTest,inetsoft.uql.asset.internal.*Test"` -- 64/64 pass

jshobe-inetsoft (changes requested):
## Code Review

### Correctness

The fix is correct and minimal. `NFMT`/`PFMT`/`CFMT`/`C2FMT` are `private static final` fields built once at class-load time from whatever `Locale.getDefault()` happens to be in effect at that moment — pinning them to `Locale.US` via explicit `DecimalFormatSymbols.getInstance(Locale.US)` (and `NumberFormat.getPercentInstance(Locale.US)`) removes that dependency entirely, which is the right call: these formatters recognize StyleBI's own canonical numeric/currency/percent text during import-time type sniffing, not user-facing locale-aware rendering (that's the separate `numFmt`/`format.number` mechanism, correctly left untouched).

Verified imports are not an issue: `AssetUtil.java` already has `import java.text.*;` (covers `DecimalFormatSymbols`) and `import java.util.*;` (covers `Locale`), so no import needed to be added — the diff compiles as shown.

The new `AssetUtilTest`:
- `formattersAreLocaleInvariant` reflects into each of the four static fields and asserts their `DecimalFormatSymbols` equal `Locale.US`'s — this is the right way to test a class-load-order-dependent defect, since an end-to-end parse test can't reliably re-trigger `AssetUtil`'s class-load in a full Surefire run once another test has already touched it first.
- `getTypeDetectsNumericValuesUnderNonUsDefaultLocale` sets `Locale.FRANCE` (a locale whose grouping separator isn't `.`, so it reproduces an actual parse failure rather than a silently-wrong-value pass) and confirms `getType()` still detects `"299.99"` and `"$499.99"` as `DOUBLE`. Both tests correctly save/restore the original default locale in `@BeforeEach`/`@AfterEach`.

This is a well-targeted fix for a real, previously-intermittent CI flake, with tests that would have caught the regression and that correctly isolate cause from symptom (the PR is honest that it couldn't pin down which specific test leaks the non-US default locale, and correctly notes the fix closes the underlying fragility regardless of which test is the culprit).

- No backward-incompatibility concerns — `numFmt`/`getNumberFormat()` (the locale-aware, user-facing formatter) is untouched, and the four import-time sniffing formatters were never meant to vary by locale, so this only fixes previously-buggy behavior.
- No security implications.

### Blocking issues for merge

1. **Failing check.** `claude-review` reports `FAILURE`.
2. **Failing check.** `build` (Validation Build) reports `FAILURE`. I was not able to retrieve the failed-step log for this run (GitHub Actions log retrieval returned empty — likely expired or redacted), so I could not confirm whether this failure is related to this change or a pre-existing/unrelated flake. Given this PR's entire premise is fixing test flakiness, please confirm the build failure is unrelated (or re-run it) before merging.

## Recommendation

Requesting changes — not due to any defect found in the code or tests, which look correct and well-reasoned, but because both the `claude-review` and `build` checks are currently failing. Please confirm/re-run the build and resolve the `claude-review` check before merge.

fredaan (approved):

PR #4737, opened 2026-08-22
Fix categorical color frame useGlobal/shareColors defaulting-on defect

larryliang-inetsoft:
## Summary
- \`set_visual_frame\` with a \`categorical\` frame type round-tripped correctly via \`get_chart_aesthetics\` but silently had zero effect on the render — colors stayed at the workspace default.
- Root cause: \`CategoricalColorModel\` defaults both \`useGlobal\` and \`shareColors\` to \`true\`. Either flag alone lets the automatic/shared palette win at render time over an explicit per-value color, even though the model itself round-trips the explicit colors perfectly.
- Only the \`mapping\` code path previously cleared these two flags. The tool's own primary documented example — a plain \`colors\` list with no \`mapping\` and no explicit \`useGlobal\` — left both flags at their default `true`, so the write was inert at render.
- Fix: `VisualFrameAliases.categoricalColor()` now clears both `useGlobal` and `shareColors` whenever `colors` and/or `mapping` is supplied (either is unambiguous override intent), while still honoring an explicit `useGlobal: true` if the caller wants to opt back into cross-chart sharing.
- Only the categorical half of this finding is addressed here; the gradient/palette half is a separate, not-yet-root-caused defect and is intentionally out of scope.
- Verified via grep that `categoricalColor()` has exactly one caller (`VisualFrameAliases.create()`'s own dispatch) — this method is not shared with the human Composer's binding controllers, so blast radius is limited to the agent-facing `set_visual_frame` tool.

## Test plan
- [x] Added `aPlainColoursListClearsUseGlobalAndShareColorsSoItActuallyRenders` to `VisualFrameAliasesTest`, covering the previously-broken "colors, no mapping, no useGlobal" shape.
- [x] `./mvnw.cmd -pl core -o test -Dtest=VisualFrameAliasesTest` — all 43 tests pass (0 failures, 0 errors), including the existing `aColourMappingClearsUseGlobalSoItActuallyRenders`, `refusesUseGlobalTogetherWithAMapping`, and `acceptsUseGlobalOnItsOwn`.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/VisualFrameAliases.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/VisualFrameAliases.java
@@ -558,16 +558,19 @@ public final class VisualFrameAliases {
    /**
     * A categorical colour frame, optionally with per-value colour mapping.
     *
-    * <p><b>The {@code useGlobal} footgun lives here.</b> {@code CategoricalColorModel} defaults
-    * {@code useGlobal} and {@code shareColors} to {@code true}, and while {@code useGlobal} is set
-    * the automatic palette wins: an explicit per-value colour is accepted, stored, and never
-    * rendered. That is a recorded defect, and it is invisible — the model round-trips perfectly
-    * and the chart shows something else.
+    * <p><b>The {@code useGlobal}/{@code shareColors} footgun lives here.</b>
+    * {@code CategoricalColorModel} defaults both {@code useGlobal} and {@code shareColors} to
+    * {@code true}. While {@code useGlobal} is set the automatic palette wins at render time; and
+    * independently, while {@code shareColors} is set the whole frame is replaced by a cached
+    * shared frame for the same column before per-value colours are even applied. Either flag
+    * alone is enough to make an explicit per-value colour accepted, stored, and never rendered.
+    * That is a recorded defect, and it is invisible — the model round-trips perfectly and the
+    * chart shows something else.
     *
-    * <p>So supplying a mapping clears {@code useGlobal}, because supplying one <i>is</i> the
-    * intent to override the automatic palette. Asking for {@code useGlobal: true} alongside a
-    * mapping is refused rather than honoured, since that combination cannot render what was
-    * asked for.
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
**Review verdict: no blocking issues found.** (Posting as a comment, not an approval — GitHub blocks self-approval since I authored this PR.)

Verified independently:

- Root cause and fix logic checked against `CategoricalColorModel` semantics: clearing both `useGlobal` and `shareColors` whenever `colors` and/or `mapping` is supplied, while honoring an explicit `useGlobal: true`, correctly closes the silent-no-render gap for the tool's own primary documented example (bare `colors` list, no mapping).
- Re-derived the three existing tests' assertions by hand against the new code path — `aColourMappingClearsUseGlobalSoItActuallyRenders`, `refusesUseGlobalTogetherWithAMapping`, and `acceptsUseGlobalOnItsOwn` all still hold (confirmed by the CI log: `VisualFrameAliasesTest` — 43/43 pass, 0 failures).
- New test `aPlainColoursListClearsUseGlobalAndShareColorsSoItActuallyRenders` exercises exactly the previously-broken shape (colors only, no mapping, no explicit useGlobal) and would have failed under the old code (old logic only touched the flags inside the mapping branch or the `useGlobal != null` branch — neither fires for this shape, so both flags would have stayed at their `true` default).
- `git grep -n "categoricalColor("` confirms exactly one caller (`create()`'s own `case "categorical"` dispatch) — blast radius claim holds, no shared code path with the human Composer's binding controllers.

CI status: `build` failed, but the only test failure (`WorksheetAgentControllerTest.importCsvDetectTypeFalseMakesEveryColumnString`) is unrelated to this diff — it's a known, actively-diagnosed pre-existing flake (see PR #4739, "DIAGNOSTIC ONLY — CSV locale flake logging round 2", and multiple other concurrently-open PRs failing on the same test). `claude-review` failed due to a bot infra error (`is_error:true`, empty `ANTHROPIC_API_KEY`), not a code review finding. Neither failure reflects on this diff's correctness.

Not merging this myself per review workflow.

PR #4738, opened 2026-08-22
Fix set_chart_type migration bugs for treemap/mekko and size aesthetic gating

larryliang-inetsoft:
## Summary

Three related defects found live during test-plan verification of `set_chart_type`, all in `ChangeChartTypeProcessor`'s merged-type migration path or its neighboring aesthetic-channel validation.

- **Fix 1**: `fixChartInfo()`'s cross-merge-family branch (e.g. pie→treemap) built a fresh `newInfo` and correctly ran every post-step against it except `fixPieDimensions(info)` — called on the stale, discarded `info`. Pie's "move dimension back off the color channel" reversal never reached the returned chart, so the dimension stayed stranded on `colorField` and `copyToTreemap()`/`copyToMekko()` found nothing on x/y/group to harvest, rendering a blank plot. Fixed by calling `fixPieDimensions(newInfo)`.

- **Fix 2**: `copyToTreemap()`/`copyToMekko()` had no reverse halves (`copyFromRadar()`/`copyFromGantt()` exist for their families, but nothing mirrored these). Leaving treemap/mekko left the dimension stuck on the `group` shelf, since `copyFields()`'s group-copying loop unconditionally carries it forward and the target type's `supportsGroupFields()` doesn't refuse it. Added `copyFromTreemap()`/`copyFromMekko()` and wired them into `process()`'s dispatch.

- **Fix 3**: `GraphTypes.supportsSize()` is checked during type-change migration and drag-and-drop, but never on a direct aesthetic-channel write. `ChartAestheticAgentService.setField()`/`clearField()` validated the channel only via `AestheticChannels.requireFieldChannel()`, with no chart-type gate on `size` for mekko/stock — the binding was silently accepted and stored but never rendered. `requireFieldChannel()` now takes a `sizeSupported` parameter (same shape as the existing `relationChart` parameter) and refuses `size` by name when unsupported.

## Test plan

- [x] `ChangeChartTypeProcessorTest` (new): `pieToTreemapMovesDimensionToGroup`, `treemapToBarMovesGroupDimensionToX` — both pass
- [x] `ChartAestheticAgentServiceTest`: added `setFieldRefusesSizeOnAChartTypeThatDoesNotSupportIt` (mekko), `setFieldRefusesSizeOnStockToo`, `setFieldAcceptsSizeOnAnOrdinaryChartType` — all pass (17/17 in class)
- [x] `ChartAestheticMutatorTest` (existing, unaffected by the overload defaults) — 26/26 pass
- Ran narrowly via `./mvnw -pl core test -Dtest=ChangeChartTypeProcessorTest,ChartAestheticAgentServiceTest,ChartAestheticMutatorTest`

### DIFF
--- a/core/src/main/java/inetsoft/report/internal/graph/ChangeChartTypeProcessor.java
+++ b/core/src/main/java/inetsoft/report/internal/graph/ChangeChartTypeProcessor.java
@@ -1309,10 +1309,11 @@ public class ChangeChartTypeProcessor extends ChangeChartProcessor {
       }
    }
 
-   // if copyToTreemap() synthesized a size/color field from a measure, move it back to y
+   // if copyToTreemap() synthesized a size/color/shape field from a measure, move it back to y
    private void moveMeasureAestheticToY() {
       AestheticRef size = info.getSizeField();
       AestheticRef color = info.getColorField();
+      AestheticRef shape = info.getShapeField();
       DataRef dataRef = null;
 
       if(size != null && GraphUtil.isMeasure(size.getDataRef())) {
@@ -1322,6 +1323,10 @@ public class ChangeChartTypeProcessor extends ChangeChartProcessor {
       else if(color != null && GraphUtil.isMeasure(color.getDataRef())) {
          dataRef = color.getDataRef();
          info.setColorField(null);
+      }
+      else if(shape != null && GraphUtil.isMeasure(shape.getDataRef())) {
+         dataRef = shape.getDataRef();
+         info.setShapeField(null);
       }
 
       if(dataRef instanceof ChartRef) {
--- a/core/src/main/java/inetsoft/report/internal/graph/ChangeChartTypeProcessor.java
+++ b/core/src/main/java/inetsoft/report/internal/graph/ChangeChartTypeProcessor.java
@@ -92,8 +92,14 @@ public class ChangeChartTypeProcessor extends ChangeChartProcessor {
       if(!GraphTypes.isTreemap(oldType) && GraphTypes.isTreemap(newType)) {
          copyToTreemap();
       }
+      else if(GraphTypes.isTreemap(oldType) && !GraphTypes.isTreemap(newType)) {
+         copyFromTreemap();
+      }
       else if(!GraphTypes.isMekko(oldType) && GraphTypes.isMekko(newType)) {
          copyToMekko();
+      }
+      else if(GraphTypes.isMekko(oldType) && !GraphTypes.isMekko(newType)) {
+         copyFromMekko();
       }
       else if(!GraphTypes.isRelation(oldType) && GraphTypes.isRelation(newType)) {
          copyToRelation();
@@ -399,7 +405,7 @@ public class ChangeChartTypeProcessor extends ChangeChartProcessor {
       fixShapeField(newInfo, newInfo, getChartType(newInfo, ref));
       fixSizeField(newInfo, getChartType(newInfo, ref));
       fixSizeFrame(newInfo);
-      fixPieDimensions(info);
+      fixPieDimensions(newInfo);
       fixAggregateRefs(newInfo);
       fixMapDimensionRefs(newInfo);
       GraphDefault.setDefaultFormulas(info, newInfo);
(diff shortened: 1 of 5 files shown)


larryliang-inetsoft:
## Blocking: build does not compile

`build` fails with a genuine compiler error, not a flaky/queued check:

```
[ERROR] core/src/main/java/inetsoft/web/wiz/binding/ChartAestheticAgentService.java:[85,26]
variable chart is already defined in method setField(String,Principal,String,String,FieldRef,String,String)
```

Root cause: this branch is stale relative to the current tip of `stylebi-wiz-main-rebase`. Comparing the base branch's current `ChartAestheticAgentService.java` against this PR's version shows an unrelated concurrent refactor already landed on the base — `setField`/`clearField` gained a `sourceTable` parameter and a `DataRefModelFactoryService` dependency, and chart resolution moved inside the `sessions.mutate(...)` lambda (which still declares its own local `chart`). This PR independently hoists a top-level `ChartVSAssembly chart = requireChart(...)` above that same lambda to compute `isSizeSupported(chart)` before validation — the two changes don't textually conflict (so GitHub reports `MERGEABLE`), but they collide semantically once merged, producing the duplicate-`chart` compile error.

This needs a rebase onto the current `stylebi-wiz-main-rebase` tip with the `sizeSupported`/`isSizeSupported(chart)` logic re-threaded through the now-current `setField`/`clearField` (which resolve `chart` inside the mutate lambda, not before it). Not a nitpick — CI is red and the module doesn't compile.

Also note: `claude-review` failed too, but that looks like an action/infra crash (`is_error:true` after 345ms and 0 review turns performed), not a content-based rejection — flagging for visibility, not counting it against the code itself.

## Review of the code changes themselves (for when this is rebased)

Read the actual method bodies rather than trusting the PR description's claims — both checked out:

- **Fix 1** (`fixChartInfo()`): confirmed the same-family branch's `fixPieDimensions(info)` at line 328 is untouched (correct — that branch mutates `info` in place), and only the cross-merge-family branch's line 408 was changed to `fixPieDimensions(newInfo)`. Correct fix.
- **Fix 2** (`copyFromMekko()`/`copyFromTreemap()`): read `copyToMekko()` directly — it does place a real measure straight onto `y` (`info.addYField(measures.remove(0))`) before any aesthetic-overflow logic runs, unlike `copyToTreemap()`, which never touches `y` for overflow measures (only size/color/shape aesthetics). So `copyFromMekko()` correctly only needs to move the group dimension back to `x`; the claim in the PR description holds up.
  - Minor non-blocking note: `moveMeasureAestheticToY()` (used by `copyFromTreemap()`) only checks `sizeField`/`colorField` for a stranded measure, not `shapeField`. If `copyToTreemap()` had placed the sole spare measure on `shapeField` (because size and color were already occupied), reversal would silently leave `y` empty. Edge case, not part of the reported defect — worth a follow-up, not a blocker.
- **Fix 3** (`AestheticChannels.requireFieldChannel` overloads): the 1-arg and 2-arg overloads simply forward `sizeSupported=true`, so this is safe by construction for any caller not touched by this diff — no need to audit call sites individually, the default is structurally preserved.
- **Tests**: `ChangeChartTypeProcessorTest`'s two tests build real `ChartInfo` state via actual dimension/aggregate refs and assert concrete shelf counts/nullness (not vacuous). The three new `ChartAestheticAgentServiceTest` cases correctly stub `VSChartInfo.getChartType()` to MEKKO/STOCK for the negative cases and assert the exception mentions `size`, and assert the model's `sizeField` actually got populated for the positive (ordinary chart type) case — meaningful assertions.

Once rebased and green, I have no code-quality objections to Fixes 1–3 or the new tests as written.

larryliang-inetsoft:
## Re-review after rebase

**Compile conflict resolved — verified, not just trusted.** Pulled the current PR source (`ChartAestheticAgentService.java` at `742cbd2`) and confirmed the new `isSizeSupported(sessionToken, user, assemblyName)` overload exactly mirrors the pre-existing `isRelationChart(sessionToken, user, assemblyName)` overload — both do an independent `sessions.resolve` + `requireChart` with no shared local, so there's no scope collision with the `sourceTable`-aware `setField`/`clearField` that already lives on the current base. Re-ran the actual CI build rather than trusting the reported local run: it now gets past `mvn compile` and executes all 6993 tests in `core` (previously it died at compile with the duplicate-`chart` error before a single test could run).

**shapeField addition reviewed.** `moveMeasureAestheticToY()` now checks `size`, then `color`, then `shape`, each following the same "if it's a measure, null the aesthetic and hand the ref back" pattern already used for the first two — consistent with the existing single-slot restore convention, no correctness issue. Not covered by a dedicated test (as previously flagged as optional), so this path is only exercised indirectly.

**`build` check is still red, but not because of this PR.** The only test failure is `WorksheetAgentControllerTest.importCsvDetectTypeFalseMakesEveryColumnString` (CSV type-detection/locale parsing) — nothing in this PR touches CSV import or type detection. I cross-checked the build log of a completely unrelated open PR (`fix/visual-frame-categorical-useglobal`, run 32571102063) and it fails on the exact same test with the exact same message. The repo also already has two diagnostic branches open for this (`diag/csv-locale-flake-logging`, `diag/csv-locale-flake-logging-2`), confirming it's a known, pre-existing, environment-wide flake being tracked separately, not something introduced here.

**`claude-review` check is also red**, but it's the same automation/infra crash as last time — `is_error:true` after ~345ms with 0 review turns performed (looks like a missing `ANTHROPIC_API_KEY` in that workflow context), not a content-based rejection.

**Net:** the code changes are correct and the compile blocker is fixed. The only thing keeping `build` from going green is an unrelated, already-tracked flaky test. I'm not merging per this workflow's standing instruction — flagging back to the coordinator to decide whether to re-run the build check or merge once the flake is independently resolved.

PR #4741, opened 2026-08-22
Fix CSVLoader silently misparsing rows beyond the type-scan window

larryliang-inetsoft:
## Summary

The originally-assigned bug (a column mixing plain and currency numbers within the *scanned* rows getting a wrong single cached format applied to every row, silently nulling out mismatched values) was **already fixed** by #4689, merged before this investigation, with regression tests (`mixedNumberAndCurrencyKeptAsString`, `mixedCurrencyAndNumberKeptAsString` in `CSVLoaderTest`). All 7 tests in `CSVLoaderTest` pass on this branch confirming that fix still holds.

While verifying the original root cause against current source, I found a **related but distinct, still-live defect**: `CSVLoader.readCSV` only samples the first `typeRows` lines to decide a numeric column's format (`fmtMap`). `parseRow()`'s second, full-data pass applies that single cached format to *every* row, but only checked `parseObject() != null` (from #4689), not whether the format consumed the *whole* string. A row beyond the scan window whose shape doesn't match the cached format (e.g. `"50%"` in a column typed from `"100.5"`/`"200.5"`) got silently accepted as a **wrong, non-null value** (`50.0` instead of `0.5`) -- worse than a dropped value, since nothing signals the corruption.

## Fix

- `CSVLoader.parseRow()`: require the cached format's `ParsePosition` to reach the end of the string before trusting the parsed result.
- When it doesn't, fall back to re-deriving the correct format from that row's own shape via a new `AssetUtil.parseNumberByShape()` (extracted from the existing scan-time format-selection logic in `AssetUtil.getType()`), instead of the plain `NumberParserWrapper` fallback, which doesn't understand currency/percent notation.

## Test plan

- [x] `CSVLoaderTest` (7 tests, including new `percentRowBeyondScanWindowNotSilentlyMisparsed`) -- `mvnw -pl core -am -o test -Dtest=CSVLoaderTest`, all pass
- [x] Reproduced the bug live before the fix (assertion failure: expected `0.5`, got `50.0`) and confirmed the fix resolves it

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java
+++ b/core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java
@@ -2938,38 +2938,66 @@ public class AssetUtil {
       }
 
       if(XSchema.DOUBLE.equals(type)) {
-         NumberFormat fmt = NFMT;
          str = trimPlus(str);
+         NumberFormat fmt = numericFormatForShape(str);
 
-         if(str.endsWith("%")) {
-            fmt = PFMT;
-         }
-         else if(str.length() > 0 && str.charAt(0) == '$') {
-            fmt = CFMT;
-         }
-         else if(str.startsWith("-$")) {
-            fmt = CFMT;
-         }
-         else if(str.startsWith("($") && str.endsWith(")")) {
-            fmt = C2FMT;
-         }
-
-         try {
-            ParsePosition pos = new ParsePosition(0);
-            fmt.parseObject(str, pos);
-
-            if(pos.getIndex() != str.length()) {
-               throw new RuntimeException("failed: " + pos.getIndex());
-            }
-
+         if(fmt != null) {
             fmtMap.put((String) header, fmt);
          }
-         catch(Throwable ex) {
+         else {
             type = XSchema.STRING;
          }
       }
 
       return type;
+   }
+
+   /**
+    * Pick the numeric format matching a value's own shape (plain, percent, currency,
+    * or parenthesized negative currency) and confirm it consumes the entire string.
+    *
+    * @return the matching format, or null if none of the candidate formats fully parse it.
+    */
+   private static NumberFormat numericFormatForShape(String str) {
+      NumberFormat fmt = NFMT;
+
+      if(str.endsWith("%")) {
+         fmt = PFMT;
+      }
+      else if(str.length() > 0 && str.charAt(0) == '$') {
+         fmt = CFMT;
+      }
+      else if(str.startsWith("-$")) {
+         fmt = CFMT;
+      }
+      else if(str.startsWith("($") && str.endsWith(")")) {
+         fmt = C2FMT;
+      }
+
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (changes requested):
## Code Review

### Correctness

The core fix is sound. `CSVLoader.parseRow()` previously trusted any non-null result from `fmt.parseObject(val, pos)`, even when the format only partially consumed the string (e.g. the plain-number format greedily matching just `"50"` out of `"50%"` and silently dropping the percent scaling, producing `50.0` instead of `0.5`). The fix requires `pos.getIndex() == val.length()` before trusting the cached-format parse, and falls back to `AssetUtil.parseNumberByShape()` — which re-derives the correct format (plain/percent/currency/parenthesized-currency) from the row's own shape — when the cached format doesn't fit. This directly matches the described defect and is exercised by the new `percentRowBeyondScanWindowNotSilentlyMisparsed` test, which reproduces the exact "cached format from early rows, mismatched shape in a later row" scenario.

The `numericFormatForShape()` extraction from `AssetUtil.getType()` is a faithful refactor — the shape-detection branches (`%`, `$`, `-$`, `($...)`) and the "must fully parse" check are preserved, just shared between the scan-time type-detection path and the new per-row fallback path.

- `numericFormatForShape("")` returns `NFMT` (not `null`) for an empty string: `ParsePosition` starts and stays at index 0 when `parseObject` fails to match anything, and `0 == "".length()` is trivially true, so the "fully consumed" check passes despite no actual parse happening. This is a pre-existing quirk carried over unchanged from the original `getType()` code (same trivial-match behavior existed before this PR), and in practice it's harmless here: `parseNumberByShape("")` still returns `null` because `fmt.parseObject("")` itself returns `null` regardless of the position check. Not a regression, just worth a one-line comment if anyone touches this again.
- `parseNumberByShape()` calls `fmt.parseObject()` twice (once inside `numericFormatForShape()` to validate the shape, once more to get the actual value) — minor redundant work on the fallback path only (not the hot path), not worth blocking on.
- No backward-incompatibility or API-surface concerns; `AssetUtil.parseNumberByShape()` is a new public static method with no prior callers to break, and the `CSVLoader` behavior change is strictly a bug fix (previously-corrupted values are now either correctly parsed or left as the original string, matching the "don't discard on mismatch" contract already documented in the surrounding code).
- Test coverage is good: the new test directly targets the described defect and passes on this branch.

### Blocking issues for merge

1. **Failing check.** The `claude-review` check reports `FAILURE` (job errored, not a substantive finding — but still a red, unresolved check).
2. **Unverified build.** The `build` (Validation Build) check shows `CANCELLED` rather than a completed success, so there is no confirmed CI build/test run backing this exact commit.

## Recommendation

Requesting changes — not due to defects in the fix itself, which looks correct and is well-tested, but because the `claude-review` check is failing and the `build` check did not complete successfully. Please re-run/resolve those checks so there's a clean, verified CI run to merge against.

PR #4746, opened 2026-08-23
fix(wiz): resolve bare sortByField column names for aliased crosstab aggregates

larryliang-inetsoft:
## Summary

Composer plugin test-plan L5, Finding 3: `set_field_sort`'s `sortByField` doesn't resolve a bare column name (e.g. `"ORDER_ID"`) for a `value_desc`/`value_asc` crosstab sort. It silently falls back to a label/chronological sort instead of throwing or working — indistinguishable from success until the actual rendered values are checked. `#4697` claimed to fix a related issue but this symptom persisted.

## Root cause

`AbstractCrosstabVSAQuery#findCol` only matched the sort's target column against an aggregate's `getAttribute()`. For any non-cube crosstab (i.e. every real crosstab — `CrossBaseVSAssemblyInfo#update()` only ever passes `aalias=false` for a cube other than SQL Server Analysis Services), `VSAggregateRef#update()` wraps the resolved column in an `AliasDataRef` named after the aggregate's full computed name (e.g. `"Count(ORDER_ID)"`) whenever that full name differs from the bare column. `createAggregates()`'s pivot-cell re-aggregation rebuild then carries that same alias forward and also overwrites the rebuilt ref's `getColumnValue()` with the query-level full name — so by the time `findCol` runs, neither `getAttribute()`/`getName()` nor `getColumnValue()` on the runtime aggregate ref reflects the original bare source column. The only place the true bare name survives is the alias's own wrapped base ref.

## Fix

Added a fallback in `findCol`: unwrap the aggregate's `ColumnRef` → `AliasDataRef` → base ref chain and compare `colname` against that base ref's `getAttribute()`.

## Tests

- New regression test `CrosstabSortByValueAliasedAggregateNameTest`, modeled on the existing `CrosstabSortByValueEntityQualifiedNameTest` (`#4697`) but with `aalias=true` — the aliasing path every real (non-cube) crosstab actually takes, which the `#4697` test never exercised. Covers both a bare column name and a fully-qualified name.
- `mvnw -o test -pl core -Dtest=CrosstabSortByValueAliasedAggregateNameTest,CrosstabSortByValueEntityQualifiedNameTest` → `Tests run: 4, Failures: 0, Errors: 0`
- `mvnw -o test -pl core -Dtest="inetsoft.report.composition.execution.*Test"` (whole package) → `Tests run: 70, Failures: 0, Errors: 0`, no regressions

Local tests pass (4/4 targeted + 70/70 package); live visual confirmation against the deployed container is in progress separately and will be added as a PR comment once done.

## Test plan

- [x] Targeted regression tests pass locally
- [x] Full `execution` package test suite passes locally (no regressions)
- [ ] Live visual confirmation against deployed stack (crosstab rows sort by `Count(ORDER_ID)` desc rather than chronologically) — in progress, will follow up as a PR comment

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
@@ -1057,6 +1057,33 @@ public abstract class AbstractCrosstabVSAQuery extends CubeVSAQuery
          if(colname.equals(aggrs[i].getAttribute())) {
             return i;
          }
+
+         // getAttribute() above still isn't enough for a normal (non-cube) crosstab: its
+         // aggregates go through VSAggregateRef#update() with aalias=true (CrossBaseVSAssemblyInfo
+         // only ever passes false for a cube other than SQL Server Analysis Services), which --
+         // whenever the aggregate's full name differs from its bare column, i.e. any aggregate
+         // with a real formula -- wraps the resolved column in an AliasDataRef named after that
+         // full name (e.g. "Count(ORDER_ID)"). AbstractCrosstabVSAQuery#createAggregates() then
+         // rebuilds the runtime aggregate again for pivot-cell re-aggregation (aggregate-on-
+         // aggregate) and carries that same alias-wrapping ref forward as its data ref, while
+         // overwriting the new ref's own design-time column value with the query's computed
+         // column name too -- so neither getAttribute()/getName() (the alias) nor getColumnValue()
+         // (now also the alias, post-rebuild) ever reflect the bare source column. The alias'
+         // wrapped base ref is the only place that original bare name still lives.
+         DataRef aggDataRef = aggrs[i] instanceof VSAggregateRef ?
+            ((VSAggregateRef) aggrs[i]).getDataRef() : null;
+
+         if(aggDataRef instanceof ColumnRef) {
+            aggDataRef = ((ColumnRef) aggDataRef).getDataRef();
+         }
+
+         if(aggDataRef instanceof AliasDataRef) {
+            DataRef base = ((AliasDataRef) aggDataRef).getDataRef();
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft (commented):
Reviewed diff against the root-cause writeup in infra/l5-tester-status.md (stylebi-wiz repo). Findings:

- Diff matches the documented root cause and fix exactly: `findCol()` in AbstractCrosstabVSAQuery now falls back to unwrapping `ColumnRef` -> `AliasDataRef` -> base ref and matching `colname` against the base ref's `getAttribute()`, for the case where the aggregate's own `getAttribute()`/`getColumnValue()` only ever reflect the aliased full name (e.g. `Count(ORDER_ID)`) under the real `aalias=true` production path. Minimal, scoped to the one method, no unrelated changes.
- New regression test `CrosstabSortByValueAliasedAggregateNameTest` is modeled on the existing `#4697` test but correctly exercises `aalias=true` (the path every real non-cube crosstab takes, which the prior test missed). Covers both bare and fully-qualified `sortByField` values. Good, targeted coverage of the actual defect.
- No correctness issues found in the diff itself.

Non-blocking: `claude-review` check is failing with an empty `ANTHROPIC_API_KEY` in its job env on this run — same symptom on #4745, looks like a CI secret/config issue unrelated to this diff's content, not a comment on the code.

Blocking for merge: the PR's own test plan checklist still has "Live visual confirmation against deployed stack (crosstab rows sort by Count(ORDER_ID) desc rather than chronologically)" unchecked, and infra/l5-tester-status.md's exact-next-steps has this as the required step before opening/merging. All local JUnit evidence (70/70, 4/4 targeted) is solid, but that doesn't substitute for the live repro this test plan requires before closing L5 Finding 3. Holding off on approval until that's confirmed and posted here.

fredaan (approved):

PR #4748, opened 2026-08-24
Read and write a chart type at the level multiStyles decides

NeverStopFMH:
The server side of reading and writing a chart's type. Pairs with
`stylebi-wiz#1761` — the plugin tools call these endpoints.

`AbstractChartInfo` says it outright: `isMultiAesthetic()` returns
`isMultiStyles()`, *"check if aesthetics are stored in aggregates"*. One flag
decides whether the authoritative type lives on the assembly or on each measure,
on both the read and the write. Everything here follows from that.

- **`ChartTypeState` + `readChartType` + `GET chart/type`** — nothing reported a
  chart's type at all, so an agent could set one and not tell a write that landed
  from one the recommender had already overwritten. `multiStyles` is reported
  alongside rather than interpreted, since it is what tells a caller which level
  to trust. `runtimeChartType` sits next to the stored value rather than replacing
  it.
- **`FieldRef.chartType`, emitted by `BindingReadService`** — triple-gated: x or y
  shelf, the ref is a measure, and multi-style is on. Only x and y are searched by
  `ChangeChartTypeProcessor`'s per-ref write, dimensions have no such property, and
  on a single-style chart the value never renders. A five-argument constructor
  delegates, so the 44 existing construction sites are untouched.
- **`setChartType(field)`** — forwarded as `ChangeChartTypeEvent.ref`.
  `requireTypeableField` refuses a non-multi-style chart and refuses a field absent
  from x and y, naming the measures that would have worked; without it the write
  reported ok and changed nothing.
- **`setChartType` preserves the flags it was not given** — `multi`, `separate`
  and `stackMeasures` defaulted to false when omitted, so every retype silently
  cleared whatever the chart had.
- **`setChartType` applies a multi-style change** — the shared `handleMulti`
  no-ops unless the flag differs, and the separate-status side never ran from the
  agent path.
- **The ten single-field shelves on a chart read** — open/high/low/close/path/
  source/target/start/end/milestone, reported only where something is bound.

Confined to `inetsoft/web/wiz/binding` by design: the Composer's own chart-type
and aesthetic dialogs run unchanged code. Nothing deconstructs `FieldRef` as a
record pattern and nothing depends on its equality, so the added component changes
no behaviour anywhere.

Verified live against a running server with a two-measure multi-style chart: one
measure typed to `line` renders a bar and a line together, the other unchanged,
and both refusals leave the chart readable with no partial write.

`inetsoft.web.wiz.**` — 2166 tests, 1 failure, pre-existing and not from this
branch: `ScriptApiServiceTest.looksUpPrototypeMethodFromGeneratedCorpus` expects
`AreaElement.addDim`, which is absent from the committed `js-functions.json`; this
branch touches no script file.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
@@ -20,6 +20,7 @@ package inetsoft.web.wiz.binding;
 import inetsoft.uql.XPrincipal;
 import inetsoft.sree.security.IdentityID;
 import inetsoft.web.wiz.binding.model.AssemblyBinding;
+import inetsoft.web.wiz.binding.model.ChartTypeState;
 import inetsoft.web.wiz.binding.model.BindableTable;
 import inetsoft.web.wiz.binding.model.FieldRef;
 import inetsoft.web.wiz.pairing.*;
@@ -127,7 +128,7 @@ public class BindingAgentController {
    public record ShelfRequest(String assembly, String shelf, List<FieldRef> fields,
                               String table) {}
    public record ChartTypeRequest(String assembly, Integer type, Boolean multi,
-                                  Boolean stackMeasures, Boolean separate) {}
+                                  Boolean stackMeasures, Boolean separate, String field) {}
    public record SwapRequest(String assembly) {}
    public record SeparateStatusRequest(String assembly, Boolean separate) {}
 
@@ -180,6 +181,16 @@ public class BindingAgentController {
                                   request.field(), source, linkUri);
    }
 
+   @GetMapping("/api/wiz/v1/agent/binding/{sessionToken}/chart/type")
+   public ChartTypeState chartType(@PathVariable String sessionToken,
+                                   @RequestParam String assembly,
+                                   Principal user)
+      throws Exception
+   {
+      requireEnabled();
+      return chartService.readChartType(sessionToken, user, assembly);
+   }
+
    @PostMapping("/api/wiz/v1/agent/binding/{sessionToken}/chart/type")
    public void setChartType(@PathVariable String sessionToken,
                             @RequestBody ChartTypeRequest request,
@@ -196,7 +207,7 @@ public class BindingAgentController {
 
       chartService.setChartType(sessionToken, user, request.assembly(), request.type(),
                                 request.multi(), request.stackMeasures(), request.separate(),
-                                linkUri);
+                                request.field(), linkUri);
    }
 
    @PostMapping("/api/wiz/v1/agent/binding/{sessionToken}/chart/swap-axes")
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindingReadService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindingReadService.java
@@ -133,13 +133,29 @@ public class BindingReadService {
             FieldRef ref = FieldRefFactory.from(model);
 
             refs.add(withChartType && model instanceof ChartAestheticModel aesthetic
-                        ? new FieldRef(ref.column(), ref.type(), ref.aggregate(), ref.dateLevel(),
-                                       ref.namedGroup(), aesthetic.getChartType())
+                        ? withTypes(ref, aesthetic)
                         : ref);
          }
       }
 
       return refs;
+   }
+
+   /**
+    * Copies a measure's own chart type onto its ref, plus the runtime type when it differs.
+    *
+    * <p>Divergence-only, matching {@code ChartTypeState}: reported every time it would be noise,
+    * reported never and a measure left at {@code auto} would answer {@code auto} while drawing
+    * something else. The runtime value is the one that survives here — the assembly-level runtime
+    * type is maintained only on the separated branch of
+    * {@code AbstractChartInfo.updateChartType}, and merged is the multi-style case.
+    */
+   private static FieldRef withTypes(FieldRef ref, ChartAestheticModel aesthetic) {
+      int stored = aesthetic.getChartType();
+      int runtime = aesthetic.getRTChartType();
+
+      return new FieldRef(ref.column(), ref.type(), ref.aggregate(), ref.dateLevel(),
+                          ref.namedGroup(), stored, runtime == stored ? null : runtime);
    }
 
    private final VSBindingService binding;
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindingReadService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindingReadService.java
@@ -24,6 +24,8 @@ import inetsoft.uql.viewsheet.Viewsheet;
 import inetsoft.web.binding.drm.DataRefModel;
 import inetsoft.web.binding.model.BindingModel;
 import inetsoft.web.binding.model.ChartBindingModel;
+import inetsoft.web.binding.model.graph.ChartAestheticModel;
+import inetsoft.web.binding.model.graph.ChartRefModel;
 import inetsoft.web.binding.model.table.CrosstabBindingModel;
 import inetsoft.web.binding.model.table.TableBindingModel;
 import inetsoft.web.binding.service.VSBindingService;
@@ -70,9 +72,36 @@ public class BindingReadService {
       Map<String, List<FieldRef>> shelves = new LinkedHashMap<>();
 
       if(model instanceof ChartBindingModel chart) {
-         shelves.put("x", refs(chart.getXFields()));
-         shelves.put("y", refs(chart.getYFields()));
+         // Under Multi Style each measure renders with its own type, so x and y carry it per field.
+         // Gated on three things, not one: the shelf, because ChangeChartTypeProcessor's per-ref
+         // write searches getXFieldCount()/getYFieldCount() and nothing else, so an aggregate on
+         // `group` or on a single-field shelf has no such setting however willingly the model
+         // answers for it; the ref being a measure, since ChartDimensionRefModel has no chart type
+         // at all; and multi-style being on, because otherwise the value is inert and reporting it
+         // would describe something that cannot render — the shape of this lane's inert-frame
+         // finding.
+         boolean perField = chart.isMultiStyles();
+
+         shelves.put("x", refs(chart.getXFields(), perField));
+         shelves.put("y", refs(chart.getYFields(), perField));
          shelves.put("group", refs(chart.getGroupFields()));
+
+         // The ten single-field shelves. Left out of this read until now, so a
+         // set_chart_single_shelf write could not be read back at all: four OHLC shelves bound and
+         // visibly rendering while this reported x/y/group and nothing else.
+         //
+         // Reported only where something is bound, unlike x/y/group above. Those three are
+         // present-and-empty because they are meaningful on every chart; these ten are not, and
+         // listing `milestone` on a pie chart would advertise a shelf that chart cannot use — the
+         // same fabricated-capability shape as the phantom Dimensions column in
+         // BindableFieldsService.isColumn and the phantom y2 axis in ChartRegionResolver.
+         for(String shelf : ChartBindingMutator.SINGLE_SHELVES) {
+            ChartRefModel ref = ChartBindingMutator.readSingleShelf(chart, shelf);
+
+            if(ref != null) {
+               shelves.put(shelf, refs(List.of(ref)));
+            }
+         }
       }
       else if(model instanceof CrosstabBindingModel crosstab) {
          shelves.put("rows", refs(crosstab.getRows()));
@@ -93,11 +122,20 @@ public class BindingReadService {
    }
 
    private static List<FieldRef> refs(List<? extends DataRefModel> models) {
+      return refs(models, false);
+   }
+
+   private static List<FieldRef> refs(List<? extends DataRefModel> models, boolean withChartType) {
       List<FieldRef> refs = new ArrayList<>();
 
       if(models != null) {
          for(DataRefModel model : models) {
-            refs.add(FieldRefFactory.from(model));
+            FieldRef ref = FieldRefFactory.from(model);
+
+            refs.add(withChartType && model instanceof ChartAestheticModel aesthetic
+                        ? new FieldRef(ref.column(), ref.type(), ref.aggregate(), ref.dateLevel(),
+                                       ref.namedGroup(), aesthetic.getChartType())
+                        : ref);
          }
       }
 
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindingReadService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindingReadService.java
@@ -21,6 +21,7 @@ import inetsoft.report.composition.RuntimeViewsheet;
 import inetsoft.uql.viewsheet.CalcTableVSAssembly;
 import inetsoft.uql.viewsheet.VSAssembly;
 import inetsoft.uql.viewsheet.Viewsheet;
+import inetsoft.uql.viewsheet.graph.GraphTypes;
 import inetsoft.web.binding.drm.DataRefModel;
 import inetsoft.web.binding.model.BindingModel;
 import inetsoft.web.binding.model.ChartBindingModel;
@@ -142,20 +143,36 @@ public class BindingReadService {
    }
 
    /**
-    * Copies a measure's own chart type onto its ref, plus the runtime type when it differs.
+    * Copies a measure's own chart type onto its ref, plus the runtime type where that says
+    * something the stored one does not.
     *
-    * <p>Divergence-only, matching {@code ChartTypeState}: reported every time it would be noise,
-    * reported never and a measure left at {@code auto} would answer {@code auto} while drawing
-    * something else. The runtime value is the one that survives here — the assembly-level runtime
-    * type is maintained only on the separated branch of
-    * {@code AbstractChartInfo.updateChartType}, and merged is the multi-style case.
+    * <p>Divergence-only, matching {@code ChartTypeState}: reported on every field it would be
+    * noise, reported never and a measure left at {@code auto} would answer {@code auto} while
+    * drawing something else. These are the runtime types that survive where it matters —
+    * {@code AbstractChartInfo.updateChartType} maintains the assembly-level value only while
+    * multi-style is <em>off</em> and delegates to {@code updateFieldChartTypes} while it is on, so
+    * the per-measure ones are available exactly when per-measure types are what render. Its
+    * parameter is named {@code separated} and its javadoc describes it backwards; the call sites
+    * are what settle it, and {@link inetsoft.web.wiz.binding.model.ChartTypeState} records them so
+    * this does not have to be re-derived twice.
+    *
+    * <p>{@code CHART_AUTO} is withheld for the reason it is withheld at the assembly level: a
+    * render resolves to something concrete, so {@code auto} in the runtime slot is the unset
+    * default rather than an answer. It has to be excluded <em>here</em> as well, because
+    * {@code updateFieldChartTypes} does not reach every measure — it runs only when the last x or y
+    * field is a measure, walks only the trailing run of aggregates on that shelf (it {@code break}s
+    * at the first non-aggregate), and {@code updateChartType} returns early altogether when no
+    * runtime x/y fields are populated. A measure missed by any of those keeps {@code 0}, and
+    * reporting that against a stored {@code bar} would announce a render as {@code auto} that never
+    * happened — the stale-read shape this pair of records exists to close, one level down.
     */
    private static FieldRef withTypes(FieldRef ref, ChartAestheticModel aesthetic) {
       int stored = aesthetic.getChartType();
       int runtime = aesthetic.getRTChartType();
+      boolean resolved = runtime != stored && runtime != GraphTypes.CHART_AUTO;
 
       return new FieldRef(ref.column(), ref.type(), ref.aggregate(), ref.dateLevel(),
-                          ref.namedGroup(), stored, runtime == stored ? null : runtime);
+                          ref.namedGroup(), stored, resolved ? runtime : null);
    }
 
    private final VSBindingService binding;
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindingReadService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindingReadService.java
@@ -165,6 +165,13 @@ public class BindingReadService {
     * runtime x/y fields are populated. A measure missed by any of those keeps {@code 0}, and
     * reporting that against a stored {@code bar} would announce a render as {@code auto} that never
     * happened — the stale-read shape this pair of records exists to close, one level down.
+    *
+    * <p>Nor is there a second source to fall back on, though the model-building code looks like
+    * there is: {@code ChartAestheticService.loadVisualFrames} sets the runtime type from
+    * {@code bindable.getRTChartType()} and then overrides it from {@code getAggregateRtType}, which
+    * returns an empty map unless the chart has an applied date comparison. Outside that case the
+    * design-time ref is the only maintainer of this value, so the guard above carries the whole
+    * weight rather than backing up a second opinion.
     */
    private static FieldRef withTypes(FieldRef ref, ChartAestheticModel aesthetic) {
       int stored = aesthetic.getChartType();
--- a/core/src/main/java/inetsoft/web/wiz/binding/ChartBindingService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/ChartBindingService.java
@@ -21,7 +21,11 @@ import inetsoft.report.composition.RuntimeViewsheet;
 import inetsoft.uql.viewsheet.ChartVSAssembly;
 import inetsoft.uql.viewsheet.VSAssembly;
 import inetsoft.uql.viewsheet.Viewsheet;
+import inetsoft.uql.viewsheet.graph.ChartAggregateRef;
+import inetsoft.uql.viewsheet.graph.ChartDescriptor;
+import inetsoft.uql.viewsheet.graph.ChartRef;
 import inetsoft.uql.viewsheet.graph.GraphTypes;
+import inetsoft.uql.viewsheet.graph.PlotDescriptor;
 import inetsoft.uql.viewsheet.graph.VSChartInfo;
 import inetsoft.web.binding.controller.ChangeChartRefService;
 import inetsoft.web.binding.controller.ChangeChartTypeService;
@@ -33,6 +37,8 @@ import inetsoft.web.binding.event.ChangeSeparateStatusEvent;
 import inetsoft.web.binding.model.ChartBindingModel;
 import inetsoft.web.binding.model.graph.ChartRefModel;
 import inetsoft.web.binding.service.VSBindingService;
+import inetsoft.web.viewsheet.service.CommandDispatcher;
+import inetsoft.web.wiz.binding.model.ChartTypeState;
 import inetsoft.web.wiz.binding.model.FieldRef;
 import inetsoft.web.wiz.viewsheet.ViewsheetSessionService;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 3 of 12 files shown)


NeverStopFMH:
## Code review

### What this does

The server half of the chart-type surface, organised around one flag:

- **`ChartTypeState` + `readChartType` + `GET chart/type`** — the first reader of a chart's type
  anywhere in the agent surface.
- **`FieldRef.chartType`, emitted by `BindingReadService`** — triple-gated on shelf ∈ {x, y}, the ref
  being a measure, and multi-style being on.
- **`setChartType(field)`** → `ChangeChartTypeEvent.ref`, with `requireTypeableField` refusing
  before anything mutates.
- **Unstated flags preserved** — send the chart's own value so the shared service compares it to
  itself.
- **`applyMultiStyle`** — the multi transition applied locally, since the shared path needs a
  WebSocket session an agent call does not have.
- **The ten single-field shelves on a chart read**, reported only where bound.

The parts I want to name as good, because they are the parts that would have been easy to skip:
the three *negative* tests on the per-measure emit (dimension, off-x/y, single-style) rather than
just the positive one; `readsTheTypeWithoutOpeningACheckpoint`, which pins the cheapest wrong
implementation instead of the behaviour; `readSingleShelf` being reused rather than reimplemented;
and `applyMultiStyle`'s javadoc explaining why the repair is here and leaving the shared method
flagged as a trap rather than silently untouched. `build` is green (20m31s).

---

### Findings

**1. `field` together with `multi: false` writes a per-measure type and then turns off the flag that
renders it.**

`requireTypeableField` runs first and checks the chart's *current* `isMultiStyles()`, so on a
multi-style chart it passes. `changeChartType` then applies the per-ref type. Then `applyMultiStyle`
sees `multi == false != true` and switches multi-style off. Net result: a per-measure type stored
where nothing renders from it, reported as success — the exact shape `requireTypeableField` exists
to prevent, arriving through a door it does not watch. Suggest refusing the combination outright
(`field != null && Boolean.FALSE.equals(multi)`), which reads naturally next to the two refusals
already there.

**2. The `forcedSeparateTypeName` refusal in `applyMultiStyle` throws *after* the retype has landed.**

`setSeparateStatus` raises the identical refusal *before* touching anything, and its comment says why.
In `applyMultiStyle` the order is inverted: `typeService.changeChartType(...)` runs at
`ChartBindingService:279`, and the throw is at `:332`. So

```
set_chart_type(assembly, type: "treemap", multi: true, separate: false)
```

retypes the chart to treemap and *then* fails. The caller gets an error and a mutated chart — which
is Finding 13's shape (mutation applied, then error) reproduced in new code, and it also makes the
PR body's "both refusals leave the chart readable with no partial write" true only of the two in
`requireTypeableField`. The check needs nothing but the requested type, so hoisting
`forcedSeparateTypeName(type)` up beside `requireTypeableField` fixes it. This branch also has no
test.

**3. The per-measure read reports the design-time type only, while the assembly read reports both.**

`ChartTypeState`'s own javadoc makes the argument: *"A read that returns only the stored value is how
a caller ends up shown what it wrote while the renderer used something else."* But
`BindingReadService` emits `aesthetic.getChartType()` and not `getRTChartType()` — which is on the
same `ChartAestheticModel` interface, and is already populated for every x/y aggregate model
(`ChartInfoModelBuilder.createRefModel:420` → `loadVisualFrames` → `ChartAestheticService:45`, plus
the `getAggregateRtType` typeMap override at `:52`). The data is on the object the read already
holds.

This bites hardest exactly where multi-style is used: a measure left at `auto` — code 0, the value
the paired plugin's own fixture uses — reports `"auto"`, and the runtime type is the only thing that
says what it actually draws. Either emit a `runtimeChartType` under the same triple gate, or say in
the javadoc why the per-measure runtime type is deliberately withheld when the assembly-level one is
not.

**4. `ChartTypeState.runtimeChartType` is not maintained on a merged chart.**

`AbstractChartInfo.updateChartType` sets the info-level `rtype` only in the `separated` branch
(`:3054`). The non-separated branch calls `updateFieldChartTypes`, which sets per-measure rtypes and
leaves the info-level value alone. The method also early-returns entirely when no x/y refs are
populated (`:3045`), leaving `rtype` at its `0` (`CHART_AUTO`) default.

The paired plugin turns `runtime != stored` into an affirmative signal — *"its presence is the
signal, so there is nothing to compare by hand"*. On a merged chart, or one that has not rendered,
that signal can fire off a stale or never-set value and report `runtimeChartType: "auto"` as though
the renderer had resolved to something else. Since merged is precisely the multi-style case, this is
not a corner. Either omit the field when the info-level value is not maintained, or document the
caveat so the plugin can stop treating presence as proof.

**5. An inbound `FieldRef.chartType` is silently ignored.**

The javadoc says so outright, and `stylebi-wiz#1761` refuses it — but only in the plugin, the
outermost and most bypassable tier. Anything reaching `POST chart/shelf` directly (wiz-services, a
script, a future client) gets the silent drop this whole lane is written against. A refusal in the
mutator, or on `ShelfRequest`, makes the guard hold wherever the call comes from.

**6. Please confirm `FieldRef` still deserialises.** It is a request type (`ShelfRequest.fields`) and
it gained a component plus a secondary constructor. Records normally bind through the canonical
constructor, so this is probably fine — but nothing tests `ShelfRequest` deserialisation, and this
branch's live verification was about typing a measure, not binding one. One live `set_chart_shelf`
would settle it.

---

### Nits

- `readChartType` hard-casts `(ChartBindingModel) binding.createModel(chart)`, where
  `BindingReadService.read` uses an `instanceof` pattern on the same call. A CCE would surface as a
  500 rather than the clean "not a chart" message the sibling produces.
- No controller-level test for the new `@GetMapping`. Both downstream tiers got route tests in the
  paired PR; the mapping itself is untested here.
- `isSeparatedGraph` defaults to `true` on a null info, "which is what the event's own null coercion
  did" — i.e. it preserves the behaviour the rest of the method is removing. Harmless, but the
  asymmetry with `isMultiStyles`/`isStackMeasures` (both default false) reads as accidental rather
  than chosen.
- `BindingReadServiceTest.neverClaimsAPerMeasureTypeForADimension` closes with a four-space `}`
  where the file uses three.
- `claude-review` fails here too (37s) while `build` passes — same bot failure as
  `stylebi-wiz#1761`, so almost certainly the workflow rather than either branch.

---

### Verdict

The design is the right one and the negative tests are what make it credible. Items 1 and 2 are the
ones I would fix before merge: both produce a write that either lands somewhere inert or lands and
then errors, which is the failure class this lane has spent its whole length eliminating. 3 and 4
are about `runtimeChartType` meaning what the paired plugin promises it means. 5–6 and the nits are
cheap.

On merge order: land this before `stylebi-wiz#1761`. If the plugin goes first, this backend ignores
`field` and retypes the whole chart while the plugin's new summary reports the single measure by name.

NeverStopFMH:
Pushed `b0d526a2b`. All six findings and the nits addressed; `build` is green on it.

### The two you named as blocking

**1 — `field` with `multi: false`. Fixed, refused as a combination.** Your trace is right, including the part that makes it nasty: `requireTypeableField` reads the chart's *current* `isMultiStyles()`, so on a multi-style chart the pair passes it, the per-measure type lands, and `applyMultiStyle` then removes the mode that renders it. I refused the pair rather than teaching `requireTypeableField` about the requested flag, because either argument alone is valid and it is only together that they cancel — a check inside that method would have had to reach for an argument it otherwise has no business reading.

**2 — the forced-separate refusal threw after the retype. Fixed by hoisting, and narrowed.** Confirmed at the line numbers you gave. Two things came out of moving it:

It only ever needed the requested `type`, so nothing had forced it downstream — it read `info.getChartType()` after the retype, which is the same value the parameter already carries.

And gating on the *resolved* separate was wrong independently of the ordering: with `separate` omitted it resolved the null against the chart's current state, so a treemap retype on a currently-merged chart was refused for a merge nobody had asked for. It now triggers only on an explicit `separate: false`. The new test asserts on `verifyNoInteractions(types)` rather than on the message, since the ordering is the whole point.

### 3 and 4, which turned out to be one change

You filed these separately and the fix is joint, because the two levels are maintained in *opposite* branches of `AbstractChartInfo.updateChartType`: `separated` sets the assembly-level value, merged calls `updateFieldChartTypes` and leaves it alone. So per-measure is the runtime type that survives exactly where the assembly-level one does not — and merged is the multi-style case.

- **`FieldRef.runtimeChartType`**, same triple gate, divergence-only.
- **`ChartTypeState.runtimeChartType` is now `Integer` and null** unless the chart is separated *and* the value is not still `CHART_AUTO`. A render always resolves to something concrete, so `auto` in that field is the unset default rather than an answer — which is the other half of the stale-zero problem you described, and one your report did not have to spell out for me to hit it.

The plugin side is `stylebi-wiz#1767`: `?? stored` was reading an absence as equality, which was right by luck, and it now names the per-measure runtime type instead.

### 5 and 6

**5 — moved to `FieldRefFactory.requireType`.** You suggested the mutator or `ShelfRequest`; `requireType` is better than either, because every inbound ref passes through it — the chart path via `toChartRef`, table and calc on their own — so one check covers all six writes instead of two, with no chance of the copies drifting.

**6 — pinned rather than confirmed.** You asked for one live `set_chart_shelf`; a live call would have told us it works today and nothing about tomorrow. `FieldRefBindingTest` covers the shape callers already send, a dimension with its date level, a list the way `ShelfRequest` holds it, the two read-only components arriving so `requireType` can refuse them, and a round trip. It binds — absent components come back null, not zero.

### Nits

All five. `readChartType` pattern-matches, so a non-chart model produces a sentence instead of a 500. The `@GetMapping` has controller tests. The four-space brace is three. On `isSeparatedGraph`: the asymmetry was chosen but the comment justified it badly — the real reason is that separated is StyleBI's own default for a chart, so a null info answers with the default rather than its opposite, which is what these helpers are for; the javadoc now says that instead of pointing at the coercion it replaced.

`claude-review` is still failing, and its own comment on this PR reads *"Claude encountered an error after 0s"*.

### Gates

`inetsoft.web.wiz.**` — 2188 tests, 1 failure: `ScriptApiServiceTest.looksUpPrototypeMethodFromGeneratedCorpus`, pre-existing and not from this branch. It expects `AreaElement.addDim`, which is absent from the committed `js-functions.json` in both the source tree and `target/classes`, and this branch touches no script file. I checked rather than assumed: stashing this branch's changes and running that single test on the branch base reproduces it.

### Merge order

You said land this before `stylebi-wiz#1761`, and that has already gone the other way — `#1761` merged as `4e8081b8b`. So the window you described is currently open on `main`: the plugin sends `field`, the deployed backend ignores `ChangeChartTypeEvent.ref`, and `requireTypeableField`'s refusals do not exist yet, so a per-measure call retypes every measure while the summary names one. Worth merging this promptly.

NeverStopFMH:
## Re-review — full diff, against the earlier round

Two review commits landed since (`b0d526a2b`, `4410ceed4`). I checked each fix in the code rather
than taking the commit messages for it. **All six findings and all four nits are genuinely
addressed**, and two of them are better than what I asked for:

| Finding | Verified |
|---|---|
| 1 — `field` + `multi: false` wrote an inert type | `ChartBindingService:265` refuses the combination before anything mutates; `refusesAPerMeasureTypeAskedForAlongsideTurningMultiStyleOff` asserts `verifyNoInteractions(types)`, and two more tests keep either argument alone working |
| 2 — forced-separate refusal threw after the retype | Hoisted to `:280`, on the *requested* type, and narrowed to an explicit `separate: false` so it no longer refuses a merge nobody asked for. `refusesAMergeOfAForcedSeparateTypeBeforeRetypingAnything` + `verifyNoInteractions(types)` is exactly the assertion that proves the hoist |
| 3 — per-measure read reported design-time only | `FieldRef.runtimeChartType`, same triple gate, divergence-only |
| 4 — assembly-level runtime type reported unconditionally | Gated, and then re-gated on the right flag by `4410ceed4` |
| 5 — inbound chart type silently ignored | `FieldRefFactory.requireNoInboundChartType`, hung off `requireType` so it covers all six writes rather than the two a mutator-level check would have — better placement than I suggested |
| 6 — `FieldRef` Jackson binding unpinned | `FieldRefBindingTest`, five cases including the list shape and a round trip. A real mapper, not a hand check |

`4410ceed4` is the one I want to call out. It did not treat the `isSeparated()` → `!isMultiStyles()`
correction as a tweak: it re-derived the rule from the call sites, wrote down *why the parameter name
is a trap* so the next reader does not have to, restated the rule in terms the response already
carries (`appliesTo`), and replaced the test that asserted the old semantics rather than adding
around it. That is the right shape for a correction.

Build green, 12m41s.

---

### Two new findings, both in the one file that correction did not reach

`4410ceed4` touched `ChartBindingService`, `ChartTypeState`, `FieldRef` and
`ChartBindingServiceTest`. `BindingReadService` — the file that actually emits the per-measure value —
was not in it, and both findings are there.

**7. `CHART_AUTO` is excluded from the assembly-level runtime type but not from the per-measure one.**

```java
// BindingReadService.withTypes
return new FieldRef(..., stored, runtime == stored ? null : runtime);
```

versus `readChartType`, which is `!model.isMultiStyles() && runtime != GraphTypes.CHART_AUTO`. The
argument for that exclusion — *"a render always resolves to something concrete, so auto here is the
unset default rather than an answer"* — is not about the assembly level. It is about the field, and it
applies verbatim per measure.

Reachable, because `updateFieldChartTypes` does not assign every measure:

- it runs only when the **last** x or y field is a measure, so `y = [Sum(A), <dimension>]` falls
  through to the x branch and assigns nothing if x has no measure either;
- within a shelf it walks only the trailing run of `ChartAggregateRef`s and `break`s at the first
  non-aggregate, so a measure sitting before a dimension is never assigned;
- and `updateChartType` returns early altogether when no RT x/y fields are populated.

In any of those, a measure keeps `rtChartType = 0` while its stored type is, say, bar — and the read
reports `runtimeChartType: auto`, presenting a never-set default as *what the measure drew*. Same
class as findings 4 and 7 of the previous round, one level down. Fix is symmetric:
`runtime == stored || runtime == GraphTypes.CHART_AUTO ? null : runtime`.

Worth noting the test fixtures point the same way. `reportsWhatAMeasureActuallyDrewWhenThatDiffersFromWhatWasStored`
uses stored `CHART_AUTO` → runtime `CHART_BAR`, the *good* direction; the reverse — stored `CHART_BAR`
→ runtime `CHART_AUTO` — has no case, and it is the one that misreports. There is a
`withholdsTheRuntimeTypeWhenItIsStillTheUnresolvedDefault` at the assembly level; the per-measure
mirror of it is what is missing.

This also matters to `stylebi-wiz#1767`, which is built entirely on the presence of this key being a
signal, and whose description says a measure left at auto is exactly the case it exists to answer.

**8. The pre-correction framing survives in two javadocs.**

`BindingReadService.java:150-151`:

> *"the assembly-level runtime type is maintained only on the separated branch of
> `AbstractChartInfo.updateChartType`, and merged is the multi-style case."*

and `BindingReadServiceTest.java:151` says the same. `ChartTypeState` and `FieldRef` were both
corrected — `FieldRef` now reads *"`multiStyles` is the flag that selects between them — not the
chart's `separated` setting, despite that being the name of the parameter"* — so what is left is one
stale copy in the file that emits the value and one in its test. Given `4410ceed4`'s stated purpose
was to stop the next reader re-deriving the rule from call sites, leaving a copy of the wrong rule in
the read service is the drift that commit was written to prevent.

---

### One residual, and I would accept it documented rather than fixed

`!= CHART_AUTO` rejects *never set*, not *stale* — and `updateChartType`'s early return means the
assembly-level value can be stale on a chart that is bound but has not produced data: retyped to bar
while `rtype` still holds the previous line, reported as a divergence that no render produced. It is
much narrower than what was fixed (an open, executed viewsheet populates RT fields), and there is no
cheap in-band test for "has this rendered". A sentence in `ChartTypeState`'s javadoc saying the field
is meaningful once the chart has produced data would close it honestly.

### Nits

- `requireType` now also refuses an inbound chart type. The placement is right; the name no longer
  describes the whole job. Not worth renaming a public method over — a `@see` on the caller side
  would do.
- `BindingAgentControllerTest` calls `controller.chartType(...)` directly, matching the file's
  existing style, so the `@GetMapping` path and Spring's own mapper are not exercised; and
  `FieldRefBindingTest` uses a bare `new ObjectMapper()`, not the configured one. Between them plus
  the live run this is covered in practice — noting it only so nobody later reads
  `FieldRefBindingTest` as proof the Spring binding is pinned.
- `claude-review` fails again (41s). Fourth consecutive failure across the two repos while `build`
  stays green, so it really is worth ten minutes on the workflow.

---

### Verdict

The two writes that could land-then-fail are gone, and gone with the assertions that prove it. Item 7
is the one thing I would fix before merge — it is a one-line change plus one test, and it is the same
defect class as the previous round's, in the level this PR added. Item 8 is a comment edit. On
ordering: `stylebi-wiz#1761` is already merged, so the plugin is live against a backend that ignores
`field`; and `#1767` is open and depends on item 7 being right.

NeverStopFMH:
## Fixed in `e0a02db5c`

Both findings from the re-review above, plus the residual. All three were in
`BindingReadService` — the file `4410ceed4` did not reach.

**7 — `CHART_AUTO` now withheld per measure too.**

```diff
  int runtime = aesthetic.getRTChartType();
+ boolean resolved = runtime != stored && runtime != GraphTypes.CHART_AUTO;

  return new FieldRef(ref.column(), ref.type(), ref.aggregate(), ref.dateLevel(),
-                     ref.namedGroup(), stored, runtime == stored ? null : runtime);
+                     ref.namedGroup(), stored, resolved ? runtime : null);
```

The javadoc now records why it has to be excluded *here* and not only at the assembly level —
`updateFieldChartTypes` does not reach every measure: it runs only when the last x or y field is a
measure, walks only the trailing run of aggregates on that shelf (breaking at the first
non-aggregate), and `updateChartType` returns early altogether when no runtime x/y fields are
populated.

New test `withholdsAPerMeasureRuntimeTypeThatIsStillTheUnresolvedDefault`, the per-measure mirror of
`withholdsTheRuntimeTypeWhenItIsStillTheUnresolvedDefault`. It covers the direction the existing
fixtures left out: the divergence case used stored `auto` → runtime `bar`, which works; stored `bar`
→ runtime `auto`, the one that misreports, had no case.

**8 — the pre-correction framing is gone.** `withTypes`' javadoc and its test's javadoc both still
said the assembly-level value is maintained *"only on the separated branch"* and that *"merged is
the multi-style case"*. Both restated in the `multiStyles` terms `4410ceed4` established.

**Residual — recorded, not fixed.** `ChartTypeState`'s javadoc now says that `updateChartType`
returns early when no runtime x/y fields are populated, so on a chart that is bound but has not
produced data neither branch runs and the value is whatever it last held; `!= CHART_AUTO` rejects
only never-set, so the honest reading is that the field describes a render once the chart has
produced one.

### Verification

- **Mutation check on the guard** — reverting `resolved` to `runtime != stored` turns the new test red
  with exactly `expected: <null> but was: <0>`, and turns *only* that test red, so it is targeted
  rather than incidentally green. Restored and re-run green.
- `BindingReadServiceTest` 14, `ChartBindingServiceTest` 40, `FieldRefFactoryTest` 20,
  `FieldRefBindingTest` 5, `BindingAgentControllerTest` 8 — **87 tests, 0 failures**.
- `inetsoft.web.wiz.**` — **2190 tests, 1 failure**, the same pre-existing
  `ScriptApiServiceTest.looksUpPrototypeMethodFromGeneratedCorpus`. Confirmed pre-existing rather
  than assumed: re-running it with these three files stashed fails identically.

### Not done, deliberately

- The `@see` on `requireType`'s caller — speculative churn on a public method's contract for a naming
  nit.
- `BindingAgentControllerTest` calling the method directly and `FieldRefBindingTest` using a bare
  `ObjectMapper`. Both match the existing style of those files, and the earlier note was explicitly
  "noting only, not a defect".
- Two >100-character lines that AGENTS.md's rule catches — `BindingReadServiceTest:242` and
  `ChartTypeState:48` — are pre-existing and untouched, to keep this diff to the finding. The six
  lines this commit adds are within 100.

NeverStopFMH:
## Third pass — whole diff at `e0a02db5c`

Caveat first: I wrote the last commit, so this is not a verdict on it. What follows is a fresh pass
over the whole diff, aimed at the parts earlier rounds did not read closely and at checking whether
the mechanisms these javadocs *cite* actually do what they say. Three things came out of that, all
documentation-precision — the code is right — plus one pre-existing bug that is not this PR's to fix.

`build` green on this head (13m22s).

---

### 9. `ChartTypeState`'s javadoc cites a mechanism that does something adjacent

> *"{@code ChangeChartTypeProcessor} clears runtime types on a per-ref write for exactly that
> reason, so the two genuinely diverge."*

That sentence is the justification for reporting `runtimeChartType` alongside `chartType` at all, so
it is worth it being exact. What the cited code does — `ChangeChartTypeProcessor.fixChartInfo`, the
`ref != null` branch — is:

```java
// clear rt charttype so the old type won't be copied over
for(VSDataRef vref : info.getRTFields()) {
   ...
   if(vref instanceof ChartAggregateRef && isResetChartType) {
      ((ChartAggregateRef) vref).setChartType(0);
   }
}
```

`setChartType`, not `setRTChartType` — it clears the **design-time** type held on the **runtime ref
objects**. There is no `setRTChartType` call anywhere in that class. The read is unaffected, because
`ChartInfoModelBuilder` builds the model from `getXFields()`/`getYFields()` — the design-time refs —
so this is a citation problem rather than a behaviour one.

The mechanism that actually makes the two diverge is `updateFieldChartTypes` →
`getRTChartType(ctype, xref, measure, mcount)`, which derives the runtime type from the stored one
and coerces it (the polar/stacked/merged rules). Worth saying that instead, since a reader who
follows the current citation finds a loop that clears a different field.

### 10. The typeMap the per-measure read appears to have is inert for ordinary charts

`loadVisualFrames` looks like it gives the per-measure runtime type two sources —
`bindable.getRTChartType()`, then an override from `ChartAestheticService.getAggregateRtType(cinfo)`.
The second one is:

```java
if(!(cinfo instanceof VSChartInfo) || !((VSChartInfo) cinfo).isAppliedDateComparison()) {
   return typeMap;   // empty
}
```

So outside an applied date comparison the override never fires, and the value `get_binding` reports
comes solely from `getRTChartType()` on the design-time ref. That is the correct source —
`updateChartType`'s own comment says so: *"make sure the RTChartType is set in the design time ref so
the gui could get the right chart type per aggregate ref"* — but it means the field has exactly one
maintainer and no fallback.

Which makes the residual `ChartTypeState` already records load-bearing for the per-measure value too,
not just the assembly one. One clause where that residual is written would stop the next reader
assuming the typeMap is a sturdier second source. (It also means the `CHART_AUTO` guard added in
`e0a02db5c` is carrying more weight than that commit message claims.)

### 11. Pre-existing, and not this PR's to fix

Found while checking 9's mechanism. In the same loop:

```java
boolean isResetChartType = true;          // declared before the x/y loops

for(VSDataRef vref : info.getRTFields()) {
   for(VSDataRef vref0 : info.getFields()) {
      if(vref0 == vref) { isResetChartType = false; break; }
   }
   if(vref instanceof ChartAggregateRef && isResetChartType) { ... }
}
```

`isResetChartType` is declared outside the outer loop and only ever assigned `false` inside it, so the
first RT field that is identical to a design-time field latches it off for every later field — the
clear stops partway through and silently does nothing for the rest. Recording it because it is the
mechanism the javadoc cites and because this lane files what it finds; it belongs in its own issue,
not here.

---

### What I verified independently this round

Not conclusions carried over from the earlier passes:

- **The `CHART_AUTO` guard's premise.** `updateChartType` calls `updateFieldChartTypes(getXFields(),
  getYFields())` on the design-time refs — the ones the model reads — but only inside the
  multi-style branch, and the whole method returns early when no runtime x/y fields are populated.
  Combined with finding 10 (no fallback source), a measure can genuinely sit at `0`.
- **Every read-side `FieldRef` construction uses the five-argument delegating constructor**
  (`FieldRefFactory.from`, all three branches), so `chartType`/`runtimeChartType` default to null
  everywhere except the one triple-gated path. No accidental emission.
- **The read agrees with StyleBI's own effective-type rule.**
  `ChangeChartTypeProcessor.getFieldChartType` is `(stored == CHART_AUTO) ? runtime : stored`. Against
  what the read now reports: stored `auto` + runtime `bar` → both reported, rule gives bar ✓; stored
  `bar` + runtime `auto` → runtime withheld, rule gives bar ✓ (and reporting `auto` would have
  contradicted it); stored `bar` + runtime `line` → both reported, and the divergence is exactly the
  engine coercion worth surfacing ✓.
- **CI:** `build` green at 13m22s on `e0a02db5c`. `claude-review` fails again — fifth consecutive run
  across the two repos, every other check green. At this point the workflow is the likelier
  explanation than five different branches.

### Verdict

Nothing blocking. 9 and 10 are javadoc clauses in files this PR already owns and would fit in the
same commit; 11 is someone else's bug that only wants an issue opened. The behaviour of this PR, as
of this head, I could not fault.

NeverStopFMH:
## Fourth pass — `965757b09`

Findings 9 and 10 are addressed in `ce14ee9e8`, javadoc only — the staged diff contained no
non-comment lines, checked mechanically rather than by eye.

This pass read the parts earlier rounds had only grepped. **Nothing wrong with the behaviour.** One
dead line in a test, now removed in `965757b09`.

### What this pass actually read

- **`FieldRefFactoryTest`'s four new cases.** They cover the inbound refusal at the right granularity:
  `chartType` alone, `runtimeChartType` alone (so neither component can be the one that slips
  through), the same refusal reached through `toChartRef` rather than only through `requireType`
  directly, and a positive case proving a ref carrying neither still passes. The negative-plus-positive
  pairing is what makes the guard's boundary observable rather than asserted.
- **`BindingAgentControllerTest`'s two new cases.** `requireEnabled` is checked with
  `verifyNoInteractions(chartService)`, so the refusal is proven to happen *before* the service is
  reached rather than after; and the passthrough is pinned with `assertSame`, which is the right
  assertion — `assertEquals` on a record would pass even if the controller rebuilt the state, and
  rebuilding is the plausible wrong implementation.
- **`FieldRefFactory.from`, all three branches.** Every read-side construction uses the five-argument
  delegating constructor, so `chartType` and `runtimeChartType` default to null everywhere except the
  one triple-gated path in `withTypes`. There is no accidental emission route.

### The one thing found, and removed

`chartTypeReturnsWhatTheServiceRead` stubbed `readChartType` twice in a row:

```java
when(chartService.readChartType("tok", null, "Chart1")).thenReturn(state);
when(chartService.readChartType(eq("tok"), any(), eq("Chart1"))).thenReturn(state);
```

`principal()` returns `() -> "admin"`, so the first stub's literal `null` could never match the call
under test, and the second matches everything the first would have. It reads like the residue of
loosening a stub that was not firing. Removed; the five binding suites stay at 87 tests, 0 failures —
unchanged, which is the proof it was carrying nothing.

### Housekeeping, not a problem

The PR description still reports the gate as `inetsoft.web.wiz.** — 2166 tests, 1 failure`. The
branch now measures **2190, 1 failure**, the same pre-existing `ScriptApiServiceTest` (re-verified by
stashing the changed files and watching it fail identically). Since this lane counts rather than
increments, the description is worth refreshing before merge.

### Verification and CI

Local, at this head: `BindingReadServiceTest` 14, `ChartBindingServiceTest` 40,
`FieldRefFactoryTest` 20, `FieldRefBindingTest` 5, `BindingAgentControllerTest` 8 — **87 tests, 0
failures**.

CI is re-running on `965757b09`; the last completed `build` was green on `e0a02db5c` (13m22s), and
everything since has been javadoc plus one deleted test line. `claude-review` has now failed on every
run across both repos while every other check passed — worth ten minutes on the workflow rather than
another shrug.

### Verdict

No problems. The behaviour, the gates, the refusals and the tests all hold up on a fourth read; what
this pass turned up was one line that did nothing and one stale figure in the description.

NeverStopFMH:
CI closed the loop the fourth-pass comment left open: `build` is **green on `965757b09`**, the current head — so the whole chain since the last green run (the two javadoc commits and the deleted stub line) is covered by a completed build rather than by inference from what it touched.

`Semgrep OSS`, `Semgrep Scan` and `license/cla` pass. `claude-review` still fails, unchanged and still the only red.

fredaan (approved):

PR #4749, opened 2026-08-24
Track unrecognized scriptable-property writes for VS_INIT/VS_LOAD too

larryliang-inetsoft:
## Summary

Follow-up to `#4733` (fix(wiz): report unrecognized scriptable-property writes in
execute_script), closing a known-remaining gap flagged by the reviewer during that PR.

`#4733` added `VSAScriptable#resetUnrecognizedWrites()`/`getUnrecognizedWrites()` and wired it
into `ScriptExecuteService.execute()` — but only for `ASSEMBLY`/`ASSEMBLY_ONCLICK` targets,
tracking the single named assembly's scriptable. `VS_INIT`/`VS_LOAD` scripts run with
`assemblyName == null` (they have no single "current" assembly) and can write to *any*
assembly in the sheet by name — e.g. `Submit1.label = "X";` inside an `onInit` script. The
identical silent-absorption bug (an unrecognized property assignment landing in the per-
assembly expando map instead of erroring, reported as a plausible success) was still fully
reachable from those two locations.

### Fix

`ScriptExecuteService.execute()` now resolves a `Map<String, VSAScriptable>` of assemblies to
track via a new `trackedScriptables()` helper:
- `ASSEMBLY`/`ASSEMBLY_ONCLICK`: unchanged — the single named assembly's scriptable.
- `VS_INIT`/`VS_LOAD`: every assembly in the sheet's own `VSAScriptable`, resolved via
  `scope.getVSAScriptable(name)`. This is a lookup, not new instantiation:
  `ViewsheetScope#addProperties0` already eagerly creates and caches every assembly's
  scriptable in the scope's own `propmap` at sandbox init, well before any script runs.

All tracked scriptables are reset before `scope.execute(...)` and checked after. Unrecognized
names are aggregated across every tracked assembly; when the target is `VS_INIT`/`VS_LOAD`
(no single assembly to name), each is qualified `"AssemblyName.property"` to disambiguate two
different assemblies' own unrecognized same-named write. The `ASSEMBLY`/`ASSEMBLY_ONCLICK`
message/behavior is byte-for-byte unchanged (bare property name, no qualification) — the
existing 9 `ScriptExecuteServiceTest` tests pass without modification.

No shape change to `ScriptExecResult`/`unrecognizedProperties` — it was already a plain
`List<String>`. A companion PR (`inetsoft-technology/stylebi-wiz`) documents the new qualified
form in the MCP tool description and adds pass-through regression tests.

## Test plan
- [x] `./mvnw -o test -pl core -Dtest=ScriptExecuteServiceTest` — 13/13 pass (9 existing +
      4 new: unrecognized write on one VS_INIT-touched assembly, correct isolation when only
      one of two touched assemblies has an unrecognized write, real-property-write regression
      guard for VS_INIT, and VS_LOAD parity)
- [x] Broader sweep `-Dtest='inetsoft.web.wiz.script.**,inetsoft.report.script.viewsheet.**'`
      — 388/388 pass
- [x] Full `core` module suite (`./mvnw -o test -pl core`) — 6992 tests, 0 failures, 0 errors

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/script/ScriptExecuteService.java
+++ b/core/src/main/java/inetsoft/web/wiz/script/ScriptExecuteService.java
@@ -21,6 +21,7 @@ import inetsoft.report.composition.RuntimeViewsheet;
 import inetsoft.report.composition.execution.ViewsheetSandbox;
 import inetsoft.report.script.viewsheet.VSAScriptable;
 import inetsoft.report.script.viewsheet.ViewsheetScope;
+import inetsoft.uql.asset.Assembly;
 import inetsoft.uql.viewsheet.VSAssembly;
 import inetsoft.web.wiz.pairing.PairingException;
 import inetsoft.web.wiz.script.model.ScriptError;
@@ -28,7 +29,10 @@ import inetsoft.web.wiz.script.model.ScriptExecResult;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft (commented):
Reviewed diff against the stated fix. Findings:

- ASSEMBLY/ASSEMBLY_ONCLICK path is byte-for-byte behavior-preserving: `trackedScriptables()` returns the same single-entry (or empty) map that the old `assemblyScriptable` local held, and the existing 9 tests pass unmodified per the PR description — confirmed by reading the diff, no changes to those test cases.
- VS_INIT/VS_LOAD path: iterates `rvs.getViewsheet().getAssemblies()`, resolves each via `scope.getVSAScriptable(name)` — confirmed this is a cache lookup (not new instantiation) per `ViewsheetScope#addProperties0` eagerly populating `propmap` at sandbox init, as the comment claims.
- Qualification logic (`assemblyName != null ? name : entry.getKey() + "." + name`) is gated on `assemblyName`, not on `target.location()` directly — correct, since by the time this code runs, `assemblyName` is non-null exactly for ASSEMBLY/ASSEMBLY_ONCLICK and null exactly for VS_INIT/VS_LOAD (other locations are rejected earlier in `execute()`).
- Summary-message guards (`assemblyName != null ? " for this assembly type (...)" : ""` and the `assemblyTypeName(rvs, assemblyName)` call) correctly avoid calling `assemblyTypeName` with a null assembly name for the VS_INIT/VS_LOAD case — that would previously have been a real NPE risk if this code path had been reached with a null name before this PR (it wasn't, since unrecognized tracking was ASSEMBLY-only before).
- New tests cover exactly the two cases flagged as worth a close look: only-touched-assemblies-report (isolation from untouched assemblies' empty `getUnrecognizedWrites()`) and the qualification-only-fires-for-VS_INIT/VS_LOAD case, plus a VS_LOAD parity test and a real-property-write non-regression guard. Good targeted coverage.

Minor non-blocking wording nit: the fallback message text "Call get_script_context for the named assembly's real scriptable properties" (used when `assemblyName == null`) is a little vague standalone, but the qualified names already in `unrecognizedProperties` tell the caller which assembly, so this is a nitpick, not a defect.

No blocking issues found. `build` check is still pending as of this review — will follow up once it completes.

PR #4750, opened 2026-08-24
Add curated response schema carrier to endpoint catalog

Nelson-cao:
## Summary
- `AbstractEndpoint` gains an optional `Object responseSchema` pass-through field, excluded from `equals`/`hashCode`, so the strict query-path mapper can bind an `endpoints.json` that carries a declared response structure. `Object` rather than `Map` because the representation allows an array at the root, and binding that to a `Map` would be a type mismatch on a known property that `ignoreUnknown` cannot excuse.
- `WizEndpointCatalogEntry` gains the matching sixth record component `Object responseSchema` — the only channel this value has into the wiz catalogue.
- `EndpointsJsonLoadableTest` gets a counting assertion added to each of its two existing test cases: every entry that declares a `responseSchema` in the raw JSON must bind to a non-null value. That guards the key being silently dropped (catalog path, which ignores unknown properties) or the field being renamed (query path, which throws on unknown properties and gets swallowed to an empty endpoint map by `Endpoints.load`).
- `WizDatabaseControllerSecurityTest` updated for the new record arity (mechanical compile fix, no behavior change).

## Notes
The counting assertions are vacuously true on this branch (nothing declares a `responseSchema` yet) and become live the moment the companion PR lands its curated trees. Stated this way on purpose: whether a declared tree REACHES the model is a property of this binding, which is what this file is for; whether the corpus declares any tree at all is a property of `endpoints.json`, asserted in #4752 where those trees are written. Asserted the other way round, this file would have had to fail until the corpus caught up.

This PR must still merge first: adding the key to `endpoints.json` before `AbstractEndpoint` can bind it empties the connector's whole endpoint map. #4752 is stacked on this branch, so its base retargets automatically on merge.

## Test plan
- [x] `mvn -pl core,connectors/inetsoft-rest -am -DskipTests install` — compiles
- [x] `mvn -pl core -Dtest=WizDatabaseControllerSecurityTest test` — passes
- [x] `mvn -pl connectors/inetsoft-rest -Dtest='!EndpointsJsonLoadableTest,!*endpoints*' test` — 118 tests pass, 0 failures
- [x] `mvn -pl connectors/inetsoft-rest -Dtest=EndpointsJsonLoadableTest test` — 2 tests, 0 failures on this branch, and 2 tests, 0 failures again with #4752's 104 curated trees present (104 declared, 104 bound on both mappers)

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/json/AbstractEndpoint.java
+++ b/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/json/AbstractEndpoint.java
@@ -77,9 +77,37 @@ public class AbstractEndpoint implements EndpointJsonQuery.Endpoint {
    }
 
    /**
-    * {@code description} is deliberately excluded here and from {@link #hashCode()}: rewording an
-    * endpoint does not make it a different endpoint, and folding it in would make identity depend
-    * on prose that is expected to be revised.
+    * The structure this endpoint's response has, as declared here rather than derived from a live
+    * response. Optional, and absent from most {@code endpoints.json} files.
+    *
+    * <p>Declared as a bare {@code Object} on purpose, and NOT as a {@code Map}. Nothing in this
+    * connector reads it: the field exists so that the loader's mapper, which rejects unknown
+    * properties and whose failure {@code Endpoints.load} turns into an EMPTY endpoint map, can bind
+    * a file that carries the key. The consumer is the wiz catalogue (see
+    * {@code WizEndpointCatalogEntry}), which parses the tree itself.</p>
+    *
+    * <p>A {@code Map} would not do. The representation allows an array at the root -- an API whose
+    * response body IS a bare array -- and binding that to a {@code Map} is a type mismatch on a
+    * KNOWN property, which no ignore-unknown setting excuses. {@code Object} binds any JSON and
+    * serializes back verbatim, so the representation can evolve without touching Java.</p>
+    *
+    * <p>Deliberately excluded from {@link #equals} and {@link #hashCode()}, for the same reason
+    * {@code description} is: restating what an endpoint returns does not make it a different
+    * endpoint.</p>
+    */
+   public Object getResponseSchema() {
+      return responseSchema;
+   }
+
+   public void setResponseSchema(Object responseSchema) {
+      this.responseSchema = responseSchema;
+   }
+
+   /**
+    * {@code description} and {@code responseSchema} are deliberately excluded here and from
+    * {@link #hashCode()}: rewording an endpoint, or declaring what it returns, does not make it a
+    * different endpoint, and folding either in would make identity depend on prose or a structure
+    * that is expected to be revised.
     */
    @Override
    public boolean equals(Object o) {
@@ -102,4 +130,5 @@ public class AbstractEndpoint implements EndpointJsonQuery.Endpoint {
    private String suffix;
    private boolean paged;
    private List<JsonLookupEndpoint> lookups = Collections.emptyList();
+   private Object responseSchema;
 }
--- a/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/stripe/endpoints.json
+++ b/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/stripe/endpoints.json
@@ -4,31 +4,710 @@
       "name": "Balance",
       "description": "The current account balance as a single record, not a list. Shows available and pending amounts per currency. Answers \"how much do I have right now\"; historical movement of money lives in Balance Transactions instead.",
       "paged": false,
-      "suffix": "/v1/balance"
+      "suffix": "/v1/balance",
+      "responseSchema": {
+        "available": [
+          {
+            "amount": "double",
+            "currency": "string",
+            "source_types": {
+              "bank_account": "double",
+              "card": "double",
+              "fpx": "double"
+            }
+          }
+        ],
+        "connect_reserved": [
+          {
+            "amount": "double",
+            "currency": "string",
+            "source_types": {
+              "bank_account": "double",
+              "card": "double",
+              "fpx": "double"
+            }
+          }
+        ],
+        "instant_available": [
+          {
+            "amount": "double",
+            "currency": "string",
+            "source_types": {
+              "bank_account": "double",
+              "card": "double",
+              "fpx": "double"
+            }
+          }
+        ],
+        "issuing": {
+          "available": [
+            {
+              "amount": "double",
+              "currency": "string",
+              "source_types": {
+                "bank_account": "double",
+                "card": "double",
+                "fpx": "double"
+              }
+            }
+          ]
+        },
+        "livemode": "boolean",
+        "object": "string",
+        "pending": [
+          {
+            "amount": "double",
+            "currency": "string",
+            "source_types": {
+              "bank_account": "double",
+              "card": "double",
+              "fpx": "double"
+            }
+          }
+        ]
+      }
     },
     {
       "name": "Balance Transaction",
       "description": "A single balance transaction by its id. Same fields as Balance Transactions; use it when the id is already known rather than to scan or aggregate.",
       "paged": false,
-      "suffix": "/v1/balance_transactions/{Balance Transactions ID}"
+      "suffix": "/v1/balance_transactions/{Balance Transactions ID}",
+      "responseSchema": {
+        "amount": "double",
+        "available_on": "double",
+        "created": "double",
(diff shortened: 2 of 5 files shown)


jshobe-inetsoft (changes requested):
## Review of PR #4750 — Add curated response schema carrier to endpoint catalog

### Summary of the change
- `AbstractEndpoint` gains an optional `Object responseSchema` field (getter/setter), excluded from `equals`/`hashCode`, so the strict query-path mapper (`EndpointJsonQuery.Endpoints.createObjectMapper()`, which rejects unknown properties) can bind an `endpoints.json` that declares a response structure without emptying the connector's whole endpoint map.
- `WizEndpointCatalogEntry` gains a matching sixth record component, `Object responseSchema`, the only channel this value has into the wiz catalogue DTO.
- `EndpointsJsonLoadableTest` adds a "declared count == bound count" assertion to each of its two existing tests, comparing a raw-JSON count of `responseSchema` occurrences against how many bind non-null on `AbstractEndpoint`/`WizEndpointCatalogEntry`.
- `WizDatabaseControllerSecurityTest` is updated mechanically for the new record arity.

### Correctness
- The mechanism checks out: `EndpointJsonQuery.Endpoints.load()` uses a plain `new ObjectMapper()` with no `@JsonIgnoreProperties`, so an unrecognized `responseSchema` key would throw and get swallowed by `load()`'s catch block into an empty map — confirmed by reading `EndpointJsonQuery.java`. Adding the property to `AbstractEndpoint` is the correct fix for that path.
- `WizEndpointCatalogEntry` already carries `@JsonIgnoreProperties(ignoreUnknown = true)`, so it wasn't at risk of throwing, but it also had no way to *carry* the value through to the wiz catalogue without the new component — that part is correctly identified as the actual reason for the second change.
- Typing the field as `Object` rather than `Map<String, Object>` is justified and correct: a JSON schema/example value can legitimately have an array at the root, and Jackson would reject binding that to a declared `Map` property even under `ignoreUnknown`, per the existing behavior of “fail on a type mismatch for a *known* property.”
- `equals`/`hashCode` are correctly left untouched (verified `hashCode()` still hashes `name, suffix, lookups, paged` only), consistent with the javadoc rationale that identity shouldn't depend on descriptive/derived content.
- Checked every place `AbstractEndpoint` is subclassed (~65 connector `*Endpoint` classes) — none override `equals`/`hashCode`/constructors in a way this breaks; the new field is purely additive and inherited for free.
- Checked all constructor call sites for `WizEndpointCatalogEntry` (record, so canonical constructor arity changed) — the only call site outside production Jackson binding is `WizDatabaseControllerSecurityTest`, and it's updated correctly (two `null` args added in the right position, matching parameter order).
- The new test assertions compare counts computed independently: the "declared" count comes from a raw `ObjectMapper` reading straight off the JSON tree (`countDeclaredResponseSchemas`), not from either bound model, so the comparison isn't circular. Good test design — it would actually catch a rename/misspelling regression in either binding path.

### Security
- No security concern found. Binding an untyped `Object` via a **plain** Jackson `ObjectMapper` (no default typing / `@JsonTypeInfo` enabled anywhere in this path) only ever produces `LinkedHashMap`/`List`/`String`/`Number`/`Boolean`/`null` — there's no polymorphic-deserialization gadget risk here. The data source (`endpoints.json`) is also a bundled connector resource, not end-user-controlled input.

- Additive-only change to both models; no existing getter/setter/behavior was removed or renamed. `equals`/`hashCode` semantics preserved.
- The one call site needing a signature update (`WizDatabaseControllerSecurityTest`) was updated in the same PR — no dangling compile break expected elsewhere (confirmed by searching all `new WizEndpointCatalogEntry(...)` and all `AbstractEndpoint` subclasses).
- No production code currently reads `getResponseSchema()`/`responseSchema()` (confirmed by search) — this PR is purely a pass-through carrier, consistent with the PR's stated purpose of unblocking a stacked follow-up (#4752) that will populate curated `responseSchema` values in `endpoints.json`. The PR description correctly notes the merge-order dependency (this must land before endpoints.json entries can declare the key), which is a real constraint worth calling out but not a defect.

### Tests
- The new assertions are correctly designed to be vacuously true today (no `endpoints.json` yet declares `responseSchema`) and become meaningful once #4752 lands — this is explained clearly in both the PR description and inline comments, and is a reasonable, intentional sequencing choice rather than a broken/no-op test.
- No existing test appears broken by this change; the only test file requiring a mechanical update was updated.

### Minor / non-blocking nitpicks
- In `EndpointsJsonLoadableTest`, `countDeclaredResponseSchemas(file)` re-parses each `endpoints.json` file from scratch with a third `ObjectMapper` (`RAW_MAPPER`), in addition to the strict mapper and the catalog mapper already parsing the same file in each respective test method. For ~65 small resource files this is not a real performance concern, but a small opportunity to share the parsed `JsonNode` if this test grows further.
- Nothing else notable; the javadoc is unusually thorough and explains the non-obvious `Object`-vs-`Map` decision well.

### CI / mergeability status (per PR metadata)
- `statusCheckRollup` shows the **`claude-review` check has `conclusion: FAILURE`**. Inspecting the job shows it failed with an infrastructure-level error ("Claude encountered an error after 0s") rather than a substantive content finding, but per review policy any check reporting `FAILURE` must be treated as a blocking failure regardless of cause.
- `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE` — not `CONFLICTING`/`DIRTY`, so there is no merge conflict blocking this PR.
- All other checks (`build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla`) report `SUCCESS`.

### Verdict
The code change itself is correct, minimal, well-justified, and does not introduce regressions, backward incompatibilities, or broken tests. **However**, this review must request changes rather than approve, solely because the `claude-review` status check currently reports `FAILURE` on the PR's head commit. Per review policy, any check with a `FAILURE` conclusion blocks approval regardless of whether the failure reflects a real code issue or an infrastructure hiccup. Please re-run/resolve that check (e.g., re-trigger the workflow) so the rollup is clean; no code changes appear to be needed based on this review.

Nelson-cao:
Thanks — the analysis is accurate throughout, including the parts I would have expected to be argued with (the `Object`-vs-`Map` reasoning and the non-circularity of the count comparison). Both items addressed below.

## Nitpick: the third `ObjectMapper`

Fixed in `35b7449a0`, and it turned out to be removable rather than merely shareable.

`countDeclaredResponseSchemas` now takes the caller's mapper and the bytes it already read, so each `endpoints.json` is read once per test method instead of twice, and `RAW_MAPPER` is gone. `readTree` does no bean binding, so the strictness that separates the two callers' mappers cannot reach this count — which is what made a third mapper duplication rather than isolation.

What I deliberately did **not** do is share the *parse* by routing the strict binding through a `JsonNode`. That would change the input path this test exists to exercise: the failure being guarded is that mapper rejecting an unknown property while reading the document, and a tree intermediate is not the same read.

## Blocking item: the `claude-review` check

Investigated rather than re-triggered, because this is not a flake. It fails on **every** PR in both `stylebi` and `stylebi-wiz`, at ~0s, consistently — the last twelve runs of the workflow are all `failure`.

Root cause from the job log (run `32734433660`):

```
{ "type": "result", "subtype": "success", "is_error": true,
  "duration_ms": 484, "num_turns": 1,
  "total_cost_usd": 0, "modelUsage": {} }
##[error]Claude result reported subtype success with is_error:true
```

`total_cost_usd: 0`, `modelUsage: {}`, one turn, 484 ms: the model was never reached. The request is rejected before any tokens are consumed, immediately after `Claude Code initialized`. That is an authentication or entitlement failure on the `CLAUDE_CODE_OAUTH_TOKEN` repository secret, not anything about this PR's contents — the same run also shows `ANTHROPIC_API_KEY` present but empty, while the workflow authenticates via `claude_code_oauth_token`.

So re-running the workflow will produce the same result. The fix is rotating that secret, which needs repository-admin access I do not have. Flagging it here rather than silently re-triggering, since a green check obtained by luck would be worse than a red one that is understood.

Two consequences for this PR specifically:

- **The review's "vacuously true today" note no longer applies.** #4752 merged into this branch at 13:44Z, five minutes after the review was submitted, so this PR now carries the carrier *and* the 104 curated Stripe trees. The count assertions are live: **104 declared, 104 bound** on both `AbstractEndpoint` and `WizEndpointCatalogEntry`.
- **The merge-order constraint the review called out is gone.** The data and the carrier now land together, so there is no longer a window in which `endpoints.json` declares a key the model cannot bind.

`EndpointsJsonLoadableTest`: 2 tests, 0 failures, run on this branch with the curated corpus present.

fredaan (approved):

PR #4752, opened 2026-08-24
Declare the response structure of 104 Stripe endpoints

Nelson-cao:
Adds the optional `responseSchema` key to `stripe/endpoints.json`, derived from Stripe's own OpenAPI specification. An endpoint that declares one gives the agent its response structure before anything is built, so it can pick a `jsonPath` and name columns in the first call instead of building once to find out what the columns are.

Design: `docs/superpowers/specs/2026-08-24-tabular-curated-response-schema-design.md` §6 (wiz repo).

**Base branch is `feat/tabular-curated-schema-carrier`, not `stylebi-wiz-main-rebase`.** That branch adds the `AbstractEndpoint.responseSchema` field. Without it the loader's mapper rejects the new key and `Endpoints.load` turns the rejection into an empty map — all 113 Stripe endpoints would disappear from the query path behind a single `LOG.error`.

## Rules applied

Each of these changes the output for most entries:

- **Every JSON number is `double`.** `JsonTable.getTypeClass` maps every JSON number to `Double.class`, so a Unix timestamp column is a `double`, not a `timeInstant`.
- **An expandable field is `string`.** Stripe declares these as `anyOf: [string, object]` and the connector sends no `expand[]`, so the response carries the id. Transcribing the object form would have been wrong on nearly every entry.
- **A field the API version history lists as added, removed or reshaped is omitted.** `StripeDataSource.getQueryHttpParameters()` sends only `Accept: application/json` — no `Stripe-Version` header — so each account answers on the version it is pinned to, while one document in the catalogue serves every account. Omitting a field costs a column; declaring one that is not there costs a round trip on every call that references it. `charge.refunds` is the clearest case: not expanded by default since `2022-11-15`.
- **`additionalProperties` becomes `{"*": "string"}`.** 116 dictionary nodes, `metadata` among them. Real keys are never enumerated: they are the customer's data.

## Coverage

| | |
|---|---|
| Endpoints | 113 |
| Declaring a structure | 104 |
| Declaring nothing | 9 |
| Leaf nodes | 5,788 (3,603 string / 688 double / 365 boolean) |
| Largest tree | `Top-Ups`, 259 nodes, depth 8 |

The 9 that declare nothing — an absent `responseSchema` is itself the statement that the structure cannot be reused: `Upcoming Invoice`, `Upcoming Invoice Line Items`, `Subscription Item Period Summaries`, `Order`, `Orders`, `Order Return`, `Order Returns`, `SKU`, `SKUs`.

No `integer`, `date` or `timeInstant` leaf appears anywhere, which is the type rule above holding across all 104 trees.

## Test

`EndpointsJsonLoadableTest` gains one assertion here: stripe must declare at least one `responseSchema`. It is kept in this change rather than with the carrier because it is a property of this file — the counting assertions on the carrier branch compare the corpus against itself and would stay green if a regenerated `endpoints.json` dropped all 104 at once.

Both cases pass against this branch (2 tests, 0 failures): 104 declared, 104 bound on the loader's bean and on `WizEndpointCatalogEntry`.

## Not changed

No existing `name`, `description`, `suffix`, `paged` or `lookups` value, and no ordering — verified by diffing the parsed old and new documents field by field. `responseSchema` is the only key added.

### DIFF
--- a/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/stripe/endpoints.json
+++ b/connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/stripe/endpoints.json
@@ -4,31 +4,710 @@
       "name": "Balance",
       "description": "The current account balance as a single record, not a list. Shows available and pending amounts per currency. Answers \"how much do I have right now\"; historical movement of money lives in Balance Transactions instead.",
       "paged": false,
-      "suffix": "/v1/balance"
+      "suffix": "/v1/balance",
+      "responseSchema": {
+        "available": [
+          {
+            "amount": "double",
+            "currency": "string",
+            "source_types": {
+              "bank_account": "double",
+              "card": "double",
+              "fpx": "double"
+            }
+          }
+        ],
+        "connect_reserved": [
+          {
+            "amount": "double",
+            "currency": "string",
+            "source_types": {
+              "bank_account": "double",
+              "card": "double",
+              "fpx": "double"
+            }
+          }
+        ],
+        "instant_available": [
+          {
+            "amount": "double",
+            "currency": "string",
+            "source_types": {
+              "bank_account": "double",
+              "card": "double",
+              "fpx": "double"
+            }
+          }
+        ],
+        "issuing": {
+          "available": [
+            {
+              "amount": "double",
+              "currency": "string",
+              "source_types": {
+                "bank_account": "double",
+                "card": "double",
+                "fpx": "double"
+              }
+            }
+          ]
+        },
+        "livemode": "boolean",
+        "object": "string",
+        "pending": [
+          {
+            "amount": "double",
+            "currency": "string",
+            "source_types": {
+              "bank_account": "double",
+              "card": "double",
+              "fpx": "double"
+            }
+          }
+        ]
+      }
     },
     {
       "name": "Balance Transaction",
       "description": "A single balance transaction by its id. Same fields as Balance Transactions; use it when the id is already known rather than to scan or aggregate.",
       "paged": false,
-      "suffix": "/v1/balance_transactions/{Balance Transactions ID}"
+      "suffix": "/v1/balance_transactions/{Balance Transactions ID}",
+      "responseSchema": {
+        "amount": "double",
+        "available_on": "double",
+        "created": "double",
+        "currency": "string",
+        "description": "string",
+        "exchange_rate": "double",
+        "fee": "double",
+        "fee_details": [
+          {
+            "amount": "double",
+            "application": "string",
+            "currency": "string",
+            "description": "string",
+            "type": "string"
+          }
+        ],
+        "id": "string",
+        "net": "double",
+        "object": "string",
+        "reporting_category": "string",
+        "source": "string",
+        "status": "string",
+        "type": "string"
+      }
     },
     {
       "name": "Balance Transactions",
       "description": "Every movement of money through the Stripe balance -- charges, refunds, payouts, fees and adjustments alike. Carries gross amount, fee and net separately, currency and exchange_rate, the date funds become available_on, and a reporting_category and type naming what caused the movement. The right table for net revenue and fee analysis, because it is the only one where fees are broken out from gross.",
       "paged": true,
-      "suffix": "/v1/balance_transactions?payout={Payout ID?}&type={Type?:charge|refund|...}&available_on={Available On?:Unix timestamp}&Created={Created?:Unix timestamp}&currency={Currency?:ISO currency code}&source={Source ID?}"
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (changes requested):
## Review of PR #4752 — "Declare the response structure of 104 Stripe endpoints"

### Summary of the change

Adds an optional `responseSchema` key to `connectors/inetsoft-rest/src/main/resources/inetsoft/uql/rest/datasource/stripe/endpoints.json` for 104 of the 113 Stripe endpoint definitions, plus one new assertion in `EndpointsJsonLoadableTest` that fails if the Stripe corpus ever loses all of its declared schemas. This targets the `feat/tabular-curated-schema-carrier` branch (correctly configured as the PR base — verified via `gh pr view`), which is where `AbstractEndpoint.responseSchema` and the binding machinery (`WizEndpointCatalogEntry`, etc.) were introduced.

### Verification performed

Since this is a 7,000+ line, almost entirely declarative JSON diff, I did automated structural/consistency checks rather than a line-by-line read of every schema:

- **Valid JSON / no duplicate keys**: the new `endpoints.json` parses cleanly and has no duplicate object keys anywhere (checked with a custom `object_pairs_hook`).
- **Endpoint counts match the PR description**: 113 total endpoints, 104 with `responseSchema`, 9 without — and the 9 named in the PR body (`Upcoming Invoice`, `Upcoming Invoice Line Items`, `Subscription Item Period Summaries`, `Order`, `Orders`, `Order Return`, `Order Returns`, `SKU`, `SKUs`) match exactly.
- **Leaf type rule holds everywhere**: only `string`, `double`, and `boolean` appear as leaf values across all 104 trees — no `integer`, `date`, or `timeInstant`, consistent with the stated `JsonTable.getTypeClass` rationale (688 double / 3603 string / 365 boolean = 4,656 actual leaves).
- **`additionalProperties → {"*": "string"}` rule**: found exactly 116 such wildcard nodes, matching the PR body's claim.
- **"Not changed" claim verified**: diffed the parsed old vs. new documents field-by-field against the `feat/tabular-curated-schema-carrier` base — `name`, `description`, `suffix`, `paged`, and `lookups` are byte-identical for all 113 entries, entry ordering is unchanged, and `responseSchema` is the only new key anywhere.
- **Size/depth caps**: largest tree (`Top-Ups`) is well within the reader's stated 2000-node / 16-level caps.
- **Test diff**: confirmed the `EndpointsJsonLoadableTest.java` diff (15 additions / 1 deletion) against the correct base branch is exactly the described `stripeResponseSchemas` guard, cleanly composed with the pre-existing corpus-wide `declaredResponseSchemas`/`boundResponseSchemas` counters. Logic is sound: it fails specifically if Stripe's declared-schema count regresses to zero, independent of the corpus-wide counts.
- Spot-checked several well-known Stripe objects (Charge, Customer, Invoice, Subscription) — schemas look structurally plausible, and `Charge.refunds` is correctly absent, matching the PR's explanation that it isn't returned without `expand[]`.

### Minor (non-blocking) nit

- The PR description's coverage table labels "5,788" as **Leaf nodes**, but that figure is actually the total node count including container objects/arrays — actual leaf count (string+double+boolean) is 4,656. The per-type breakdown given (3,603/688/365) is correct and sums to 4,656, not 5,788, so the total in that row is mislabeled. This is a documentation/description accuracy issue only, not a functional problem with the diff, but worth fixing before/at merge for anyone relying on that PR description later.

### Error handling / regressions / backward compatibility

- `responseSchema` is additive and optional; no existing behavior for consumers that don't read the field changes. No backward-incompatibility identified.
- No dead code, no unused imports, no code style issues — this is a pure data + one test-assertion change.
- No security concerns: no secrets, no dynamic code, no user-controlled input paths affected; `additionalProperties` (customer-controlled `metadata`) is correctly left untyped as `{"*": "string"}` rather than attempting to enumerate real customer keys.

### Blocking issue: CI status

Per review policy, any check with `conclusion: FAILURE` blocks approval regardless of content review outcome. `gh pr view 4752 --json statusCheckRollup` currently reports:

- `claude-review` (Claude Code Review workflow): **FAILURE** — the job's "Run Claude Code Review" step failed after ~20s (before producing a review), which appears to be a run/execution failure rather than a review verdict, but per policy this still counts as a failing check.
- `Semgrep Scan`, `build` (Validation Build), `Semgrep OSS`, `license/cla`: all SUCCESS.

`mergeable` is `MERGEABLE` and there is no conflict, but `mergeStateStatus` is `UNSTABLE`, consistent with the failing check above.

### Verdict: Changes requested

The content of the change itself is correct, well-verified against its own stated design rules, and low-risk (additive, declarative, backed by a new regression-guarding test). However, per the review policy for this repository, I cannot approve while a required check (`claude-review`) reports `FAILURE`. Please re-run/investigate that check (it looks like an infrastructure/execution failure in the workflow rather than a substantive finding) and re-request review once it's green. I'd also suggest fixing the "Leaf nodes: 5,788" mislabeling in the PR description while addressing the above.

PR #4754, opened 2026-08-24
Fix treemap/mekko-to-bar migration duplicating a dimension already on x and group

larryliang-inetsoft:
## Summary

L3 Finding 4 (reopened) — root-caused by debugger-l3-migration.

`#4738`'s `ChangeChartTypeProcessor.moveGroupFieldsToX()` unconditionally appends every
`group`-shelf field onto `x` with no dedup check. If a dimension is already present on both
`x` and `group` when a chart transitions out of treemap/mekko (e.g. `Submit1`'s Category
already on `x` from before entering treemap, plus the copy treemap put on `group`),
`moveGroupFieldsToX()` appends it a second time onto `x` while emptying `group` — leaving the
chart with a duplicated dimension on `x` and nothing on `group`.

## Fix

`moveGroupFieldsToX()` now checks each group field against every field already on `x` before
appending — reusing the exact `Tool.equals(..., getFullName())` idiom this file's own
`isRefEquals` (line ~225) already uses for the identical "is this ref already represented"
question. `group` is still unconditionally emptied either way, matching the existing
contract. No new imports needed — `Tool`/`Arrays` are both already covered by this file's
existing wildcard imports.

`copyFromMekko()` shares `moveGroupFieldsToX()`, so it's fixed for free too — added a
regression test for that path as well.

**Out of scope, left alone per the debugger's own audit** (separate/unconfirmed, not
requested):
- `fixPieDimensions()`'s pie-conversion branch unconditionally overwriting an existing
  measure-typed aesthetic (a different defect).
- The incidental `set_visual_frame` false-error-on-success lead in
  `ChangeChartAestheticService` (only a lead, not root-caused).

## Test plan
- [x] Added `treemapToBarDoesNotDuplicateADimensionAlreadyOnXAndGroup` and
      `mekkoToBarDoesNotDuplicateADimensionAlreadyOnXAndGroup` to
      `ChangeChartTypeProcessorTest`, matching the file's existing `changeType`/`dimension`
      helper shapes.
- [x] Verified the new tests actually catch the bug: reverted just the source fix via
      `git stash`, reran — both failed with `expected: <1> but was: <2>` (the exact reported
      duplication). Restored the fix and reran — green.
- [x] `./mvnw -o test -pl core -Dtest=ChangeChartTypeProcessorTest` — 8/8 pass (6 existing + 2
      new, existing ones unmodified)
- [x] Full `core` module suite (`./mvnw -o test -pl core`) — 6990 tests, 0 failures, 0 errors

### DIFF
--- a/core/src/main/java/inetsoft/report/internal/graph/ChangeChartTypeProcessor.java
+++ b/core/src/main/java/inetsoft/report/internal/graph/ChangeChartTypeProcessor.java
@@ -1313,7 +1313,14 @@ public class ChangeChartTypeProcessor extends ChangeChartProcessor {
 
    private void moveGroupFieldsToX() {
       for(int i = info.getGroupFieldCount(); i > 0; i--) {
-         info.addXField(info.getGroupField(0));
+         ChartRef gfield = info.getGroupField(0);
+         boolean alreadyOnX = Arrays.stream(info.getXFields())
+            .anyMatch(xfield -> Tool.equals(xfield.getFullName(), gfield.getFullName()));
+
+         if(!alreadyOnX) {
+            info.addXField(gfield);
+         }
+
          info.removeGroupField(0);
       }
    }
--- a/core/src/test/java/inetsoft/report/internal/graph/ChangeChartTypeProcessorTest.java
+++ b/core/src/test/java/inetsoft/report/internal/graph/ChangeChartTypeProcessorTest.java
@@ -73,6 +73,41 @@ class ChangeChartTypeProcessorTest {
                    "the dimension must not be left stranded on the group shelf bar never reads");
    }
 
+   /**
+    * If a chart is already left with the same dimension on both {@code x} and {@code group}
+    * (e.g. from an earlier, separately-caused inconsistency) when it transitions out of
+    * treemap, {@code moveGroupFieldsToX} must not blindly append the group copy onto x --
+    * that would duplicate the dimension on x while still emptying group.
+    */
+   @Test
+   void treemapToBarDoesNotDuplicateADimensionAlreadyOnXAndGroup() {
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft (commented):
Reviewed diff against the described root cause. Findings:

- Fix is exactly as described: `moveGroupFieldsToX()` now dedups each group field against the existing x fields via `Tool.equals(..., getFullName())` before appending, reusing the same idiom `isRefEquals` already uses elsewhere in this file. `group` is still unconditionally emptied on every iteration — matches the existing contract, no behavior change there.
- Confirmed no new imports needed: `Tool` (`inetsoft.util.*`) and `Arrays` (`java.util.*`) are both already covered by this file's existing wildcard imports.
- `copyFromMekko()` correctly gets the same fix for free since it shares the helper — the new `mekkoToBarDoesNotDuplicateADimensionAlreadyOnXAndGroup` test covers that path directly rather than just asserting by inference.
- Loop structure is unchanged (`removeGroupField(0)` still runs every iteration regardless of the dedup outcome), so the fix is a pure narrowing of when `addXField` fires — no risk of leaving a stale group field behind or double-removing.
- Good verification discipline noted in the PR description: reverted just the source fix and confirmed both new tests fail with the exact reported duplication (`expected: <1> but was: <2>`) before restoring it — that's real evidence the tests catch the bug, not just exercise the code path.
- Scope: appropriately narrow, the two flagged-but-unfixed leads (`fixPieDimensions()` aesthetic overwrite, `set_visual_frame` false-error) are correctly left alone as separate/unconfirmed defects, not silently rolled into this fix.

No blocking issues found. `build`/`Semgrep Scan`/`claude-review` were still pending as of this review — will follow up once they complete.

PR #4755, opened 2026-08-24
Report the table properties the worksheet write tools set

lizhepu:
`set_table_properties`, `set_table_mode` and `set_assembly_position` all wrote fields that `read_worksheet_model` never reported back. A caller could not distinguish a working write from a dropped one — which is exactly what left L1 case **1.19** unable to round-trip and **1.16** unable to verify a position at all.

## Nothing new is computed

`TableAssembly` already exposes every one of these, and `TablePropertyDialogService:64-77` reads the same getters to fill the Composer's own table-properties dialog. This surfaces them through `TableModel`:

| Field | Source |
|---|---|
| `description` | `getDescription()` |
| `maxRows` | `getMaxRows()` |
| `distinct` | `isDistinct()` |
| `visibleInViewsheet` | `isVisibleTable()` |
| `mode` | derived from `isLiveData()` / `isRuntime()` / `isEditMode()` |
| `x` / `y` | `getPixelOffset()` |

## Two judgement calls worth reviewing

**`maxRows` reports `null` for unlimited**, not the stored `-1` — which would read back as a real limit of −1 — and not `0`, which would collide with a genuine limit of zero rows.

**`mode` reports the state, not the word that was written, and the javadoc says so** rather than implying a perfect round-trip. No field stores the mode; `set_table_mode` writes three flags per mode, and two of its five words do not map to distinct states: `"live"` sets runtime from `isRuntimeSelected()`, so on a table whose runtime selection is off it lands in the same state as `"detail"`; and `"default"` is resolved by the writer to live for an embedded table and metadata for a bound one. Neither is a lost write — both report the mode the table is actually in — but a caller wanting to confirm a specific word was honoured should compare states, not strings.

## Verification

5 tests, and **non-vacuity checked with two separate mutations** rather than assumed: stubbing the new fields empty fails three of them, and reporting `maxRows` verbatim fails the `-1` case. The visibility test asserts both directions, since a one-sided assertion would pass against a hardcoded `false`.

`inetsoft.web.wiz.worksheet.*Test` — **219 tests, all green**. `TableModel` has exactly one construction site, so no caller was left behind by the record change.

## What this does not do

It does not unblock **1.10**, which needs row counts. `rowCount` exists in the dialog model but only for `EmbeddedTableAssembly` (`getRowCount() - 1`), and 1.10's concatenation sources come from a datasource — so that case needs a different oracle.

Live verification of 1.16 and 1.19 is still owed and needs a rebuilt stack.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
@@ -28,6 +28,7 @@ import inetsoft.web.wiz.worksheet.model.WorksheetModel;
 import inetsoft.web.wiz.worksheet.model.WorksheetPropertiesModel;
 import org.springframework.stereotype.Service;
 
+import java.awt.Point;
 import java.util.*;
 
 /**
@@ -117,7 +118,7 @@ public class WorksheetReadService {
       WorksheetModel.AggregateModel aggregates = readAggregates(t);
       List<WorksheetModel.SortModel> sorts = readSorts(t);
 
-      java.awt.Point offset = t.getPixelOffset();
+      Point offset = t.getPixelOffset();
       int maxRows = t.getMaxRows();
 
       return new WorksheetModel.TableModel(
@@ -126,8 +127,9 @@ public class WorksheetReadService {
          preConditions, postConditions, rankingConditions,
          aggregates, sorts, primary,
          t.getDescription(),
-         // -1 is how the assembly stores "unlimited". Reporting it as-is would be read back as a
-         // real limit of -1, and reporting 0 would collide with a genuine limit of zero rows.
+         // Anything <= 0 is "no limit" -- the engine applies one only when it is positive -- so
+         // -1 and 0 both report null rather than reading back as a limit of -1 or of zero rows.
+         // Note this is the effective limit, capped by query.runtime.maxrow; see TableModel.
          maxRows <= 0 ? null : maxRows,
          t.isDistinct(), t.isVisibleTable(), tableMode(t),
          offset == null ? null : offset.x, offset == null ? null : offset.y);
@@ -274,7 +276,9 @@ public class WorksheetReadService {
    }
 
    /**
-    * The table's display mode, as the vocabulary {@code set_table_mode} accepts.
+    * The table's display mode, in four of the five words {@code set_table_mode} accepts --
+    * {@code live}, {@code full}, {@code detail} and {@code edit}. The fifth, {@code default},
+    * is not a state and never appears here; see the note below.
     *
     * <p>No field stores it: the mode is a combination of {@code liveData}, {@code runtime} and
     * {@code editMode}, and {@code set_table_mode} writes all three per mode. Deriving it here is
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
@@ -117,11 +117,20 @@ public class WorksheetReadService {
       WorksheetModel.AggregateModel aggregates = readAggregates(t);
       List<WorksheetModel.SortModel> sorts = readSorts(t);
 
+      java.awt.Point offset = t.getPixelOffset();
+      int maxRows = t.getMaxRows();
+
       return new WorksheetModel.TableModel(
          name, type, columns, joins,
          readSources(t), readConcatType(t), readConcatCompatible(t), readAutoUpdate(t),
          preConditions, postConditions, rankingConditions,
-         aggregates, sorts, primary);
+         aggregates, sorts, primary,
+         t.getDescription(),
+         // -1 is how the assembly stores "unlimited". Reporting it as-is would be read back as a
+         // real limit of -1, and reporting 0 would collide with a genuine limit of zero rows.
+         maxRows <= 0 ? null : maxRows,
+         t.isDistinct(), t.isVisibleTable(), tableMode(t),
+         offset == null ? null : offset.x, offset == null ? null : offset.y);
    }
 
    // -------------------------------------------------------------------------
@@ -262,6 +271,39 @@ public class WorksheetReadService {
     */
    private Boolean readAutoUpdate(TableAssembly t) {
       return t instanceof MirrorTableAssembly mirror ? mirror.isAutoUpdate() : null;
+   }
+
+   /**
+    * The table's display mode, as the vocabulary {@code set_table_mode} accepts.
+    *
+    * <p>No field stores it: the mode is a combination of {@code liveData}, {@code runtime} and
+    * {@code editMode}, and {@code set_table_mode} writes all three per mode. Deriving it here is
+    * what lets a caller read back what it just set — without this, the write is verifiable only by
+    * its side effects, which is what left case 1.19 unable to round-trip.
+    *
+    * <p>Checked in the order the writer distinguishes them: {@code edit} owns editMode, and among
+    * the live modes only {@code live} sets runtime, so live-without-runtime reads as
+    * {@code detail}.
+    *
+    * <p><b>This reports the resulting state, not the word that was written, and the two can
+    * differ.</b> {@code set_table_mode("live")} sets runtime from {@code isRuntimeSelected()},
+    * so on a table whose runtime selection is off it lands in the same state as
+    * {@code "detail"} and reads back as {@code detail}. Likewise {@code "default"} is not a
+    * state of its own — the writer resolves it to live for an embedded table and metadata for a
+    * bound one — so it reads back as {@code detail} or {@code full}. Neither is a lost write;
+    * both are the mode the table is actually in. A caller wanting to confirm a specific word was
+    * honoured should compare states, not strings.
+    */
+   private String tableMode(TableAssembly t) {
+      if(t.isEditMode()) {
+         return "edit";
+      }
+
+      if(t.isLiveData()) {
+         return t.isRuntime() ? "live" : "detail";
+      }
+
+      return "full";
    }
 
    private String tableType(TableAssembly t) {
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/model/WorksheetModel.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/model/WorksheetModel.java
@@ -90,17 +90,28 @@ public record WorksheetModel(List<TableModel> tables, List<VariableModel> variab
     * @param sorts              sort directives; empty when none
     * @param primary            {@code true} if this is the worksheet's primary assembly
     * @param description        the table's own description; {@code null} when unset
-    * @param maxRows            the table's row limit, or {@code null} when unlimited. Stored as
-    *                           {@code -1} for unlimited and reported as {@code null} so that an
-    *                           unset limit is not confused with a limit of zero rows.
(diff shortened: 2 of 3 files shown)


lizhepu:
## Review — and the findings fixed in `2a319b05`

Verified locally on the PR head: `inetsoft.web.wiz.worksheet.*Test` runs **215 tests, 214 pass**. The one error is `GroupSpecDeserializationTest` failing on a missing `PropertiesEngine` bean — I checked it against the pristine branch and it fails there too, so it is environmental, not this PR. (I count 215 where the description says 219; worth a glance at which selector you ran, but nothing in the diff explains a gap.)

I traced the mode derivation against the writer, and it is correct for all five words:

| written | flags `setTableMode` writes | `tableMode` reads back |
|---|---|---|
| `edit` | live=false, runtime=false, edit=true | `edit` |
| `live` | live=true, runtime=`isRuntimeSelected()`, edit=false | `live`, or `detail` when runtime selection is off |
| `detail` | live=true, runtime=false, edit=false | `detail` |
| `full` | live=false, runtime=false, edit=false | `full` |
| `default` | live=`instanceof EmbeddedTableAssembly`, runtime=false, edit=false | `detail` (embedded) / `full` (bound) |

Checking `isEditMode()` first is load-bearing and correctly ordered. Both judgement calls flagged in the description are the right ones. The behaviour needed no change — but three of the four things the code said *about itself* were wrong, and one of them is the same class of defect this PR family exists to remove. Fixed and pushed:

### 1. `maxRows` is not the stored limit — it is the effective one

This is the substantive finding. `TableAssemblyInfo.getMaxRows()` does not return `mrow`:

```java
public int getMaxRows() {
   // Feature #39140, always respect the global row limit
   return Util.getQueryLocalRuntimeMaxrow(mrow);
}
```

That folds in `query.runtime.maxrow` and the organization row limit. So on any server with either configured:

- a table stored as unlimited (`-1`) reports **the cap**, not `null` — `globalLimit > 0 && localLimit > 0` is false for `-1`, so it falls to `max = globalLimit`;
- `set_table_properties(maxRows: 25)` under a cap of 10 reads back **10**.

The javadoc said "the table's row limit, or `null` when unlimited", which describes a round-trip that only holds on an uncapped server — and `anUnlimitedRowLimitIsReportedAsNullRatherThanMinusOne` passes only because this suite has no cap configured.

**Reading the effective value is the right call and I did not change it** — `TablePropertyDialogService:68` fills the Composer's own dialog from the same getter, so the model reports what the product reports. What changed is that the record javadoc now says which of the two it is, and the test records the environment its assertion depends on, plus why asserting the capped case would mean mutating global state from a unit test.

### 2. The `null` was justified by a limit that does not exist

> reported as `null` so that an unset limit is not confused with a limit of zero rows

There is no limit of zero rows. `AbstractTableAssembly:2103` applies one only `if(getMaxRows() > 0)`, `isMaxRowsSet()` is `mrow > 0`, and `set_table_properties` normalizes `0` to `-1` on the way in. `0` *is* unlimited. The real reason to map both to `null` stands on its own — `-1` would read back as a limit of −1 — so the rationale now says that instead of resting on a case that cannot occur.

### 3. `mode` advertised `default`, which `tableMode` never returns

```java
 * @param mode  ... {@code live}, {@code full}, {@code detail}, {@code edit} or {@code default} ...
```

The description already explains that `default` is not a state and resolves to detail or full — so the record javadoc contradicted both the implementation and the PR's own reasoning, and invited a caller to test for a string that cannot appear. This is exactly the "advertises something it does not do" shape the read-back gap was about, one level up. `default` is now documented as the write-only word it is, and `tableMode`'s summary line says it returns four of the five.

### 4. Three small ones

`java.awt.Point` was qualified inline in a file that imports everything else (`import java.awt.Point;` now, in both the service and the test); `@param x` and `@param y` carried identical text, so neither said which axis it was; and `WorksheetReadServiceTest.java` lost its trailing newline — it was the only file in the package without one.

---

### Not changed — two notes for whoever picks this up

**`set_table_mode` silently accepts any string.** Its `switch` routes unrecognized input to the `default ->` branch, so `set_table_mode("liv")` applies default mode and reports success. Pre-existing and out of scope here, but worth recording: this PR is what makes it *detectable* for the first time, since a caller can now read `mode` back and see a word it never asked for.

**The plugin needs no change to carry these fields, but its description will be stale.** `read_worksheet_model` returns the server model verbatim (`return model` after the concat check), so the new fields reach callers as soon as this merges. The tool description enumerating what each table reports does not mention them — that update belongs in `stylebi-wiz` *after* this ships, since advertising fields the server does not yet send would be the same defect pointing the other way.

lizhepu:
## Point-by-point response

### Fixed — `2a319b05`

**1. `maxRows` documented as the stored limit when it is the effective one.** Fixed. The javadoc now says `getMaxRows()` folds in `query.runtime.maxrow` and the organization row limit, so a capped server can report less than was written and a table stored as unlimited reports the cap. Behaviour unchanged — `TablePropertyDialogService:68` reads the same getter, so the model shows what the Composer shows.

**2. The `null` justified by "a limit of zero rows".** Fixed. No such limit exists: `MirrorQuery:337` is `boolean userMaxRows = query.getTable().getMaxRows() > 0`, `PreAssetQuery:1242` treats `max <= 0` as unset, and `set_table_properties` normalizes `0` to `-1` on the way in. The rationale now rests on the reason that holds — `-1` would read back as a limit of −1.

**3. `mode` javadoc advertising `default`.** Fixed. `default` is documented as the write-only word it is, and `tableMode`'s summary says it returns four of the five.

**4. `java.awt.Point` inline, identical `@param x`/`@param y`, missing trailing newline.** All three fixed.

### Wrong in the review — withdrawn

**The 215-vs-219 test count.** My review asked which selector produced 219. That question was mine to answer, and the answer is that **219 is right and the selector is fine**. 219 is the declared `@Test` count in the package — 5 + 51 + 85 + 10 + 7 + 23 + 7 + 31. I measured 215 because `GroupSpecDeserializationTest` fails to build its Spring context locally (missing `PropertiesEngine` bean), so 4 of its 5 tests never ran and the class contributed a single error entry instead: 219 − 5 + 1 = 215, exactly. Same suite, same tests; only my environment differs. Nothing here to look at.

### Not changing — reasons

**`set_table_mode` routes any unrecognized string to its `default ->` branch,** so `set_table_mode("liv")` applies default mode and reports success. Leaving it: it is a pre-existing defect in the writer, not in this diff, and fixing it changes write behaviour — a caller relying on a loose spelling today would start getting an exception. That belongs in its own change, with its own test. I raised it only because this PR is what makes it observable for the first time: a caller can now read `mode` back and see a word it never asked for.

**`read_worksheet_model`'s description does not mention the new fields.** Leaving it: wrong repo, and more importantly wrong order. The plugin returns the server model verbatim, so the fields reach callers the moment this merges — but describing them in the tool before the server ships them would advertise fields that are not there, which is the same defect this PR removes, pointed the other way. It should follow this, not lead it.

jshobe-inetsoft (changes requested):
## Independent review

I read the diff, the current `WorksheetReadService.java` / `WorksheetModel.java` / `WorksheetReadServiceTest.java` on the PR branch, and the full review thread from `lizhepu` (including the `2a319b05` follow-up commit that fixed the four issues she raised). I also independently verified the claims against the surrounding codebase rather than taking them at face value:

- `TableAssemblyInfo.getMaxRows()` does indeed run `mrow` through `Util.getQueryLocalRuntimeMaxrow(mrow)`, confirming `maxRows` here is the *effective*, org/global-capped limit, and the updated javadoc on both `WorksheetModel.TableModel` and the test now says so accurately.
- `TablePropertyDialogService.getModel()` reads `getDescription()`, `isDistinct()`, `getMaxRows()`, `isVisibleTable()` — the same getters this PR surfaces — confirming the new fields don't compute anything new and stay consistent with what the Composer's own dialog shows.
- `TableAssembly.isEditMode()/isLiveData()/isRuntime()/isRuntimeSelected()` all exist as used, and the `tableMode()` precedence (edit → live/detail → full) matches the flag combinations `set_table_mode` writes per the table lizhepu built in her review comment.
- `WorksheetModel.TableModel` has exactly one construction site in `core/`, and no test outside `WorksheetReadServiceTest` constructs it positionally — the record change doesn't strand any other caller.
- The new tests' helpers (`read(ws)`, `tableNamed(...)`) already existed in the file before this diff, so the added tests compile against real, pre-existing scaffolding rather than something invented for this PR.

I did not find any new issues beyond what `lizhepu` already raised and got fixed in `2a319b05`. Specifically:

- **Correctness**: the `maxRows <= 0 ? null : maxRows` mapping and the `tableMode()` derivation are both correct and now honestly documented (effective vs. stored limit; `default` never appears in reads).
- **Error handling / edge cases**: `offset == null ? null : offset.x/.y` guards the one nullable getter in the new code. Minor, non-blocking observation: `AssemblyInfo`'s field initializer sets `pixelOffset = new Point(0, 0)` by default, so in practice new tables report `x=0, y=0` rather than `null` unless something explicitly nulls the offset — the javadoc's "null only if the assembly carries no offset at all" is technically accurate but a reader could come away expecting `null` to be the common unset case. Not worth blocking on.
- **Readability/maintainability**: the javadoc is unusually thorough (arguably more prose than a typical DTO field needs), but it's accurate and pre-empts exactly the kind of misunderstanding a caller of this read-back API would otherwise hit, so I see it as a net positive given the API is consumed by an LLM agent tool rather than another engineer who can ask questions.
- **Security**: none of the new code touches authn/authz, SQL, or file I/O — it's a pure read-back projection of already-authorized-to-read assembly state. No concerns.
- **Regressions/backward compatibility**: the change is purely additive to a `record` (new trailing fields, `@JsonInclude(NON_NULL)` already in place), and the single construction site was updated in the same commit. `distinct`/`visibleInViewsheet`/`primary` remain primitive `boolean` (always serialized), consistent with the existing `primary` field's convention. No existing test assertions were changed in incompatible ways.
- **Tests**: the 5 new tests are meaningful, not vacuous — the PR description and the review thread both note the double-mutation check (stub-empty fails 3 tests; verbatim `maxRows` fails the `-1` case), and I don't see anything to add. The one flagged discrepancy (`215` vs `219` tests locally) was chased down and resolved as an environmental Spring-context issue in `GroupSpecDeserializationTest`, pre-existing on the base branch — not something this PR introduced.

Overall, the code substance here is solid, and the human review thread already extracted and fixed the real issues (mislabeled effective-vs-stored `maxRows`, a nonexistent "limit of zero rows" justification, `default` advertised as a return value it can never be, plus a few nits). I have nothing to add on top of that.

## Why this can't be an approval right now

Per review policy, approval requires no failing checks on the PR. `statusCheckRollup` currently shows:

- `claude-review` (workflow "Claude Code Review") — **conclusion: FAILURE**. The run log shows `Claude execution failed: result is_error:true` / `Claude result reported subtype success with is_error:true` — this is an automation/tooling error in the review bot's own run, not a code-quality finding — but per policy any check with conclusion `FAILURE` blocks approval regardless of cause.

All other checks are green (`Semgrep Scan`, `build`, `Semgrep OSS`, `license/cla`), `mergeable` is `MERGEABLE` (no conflicts), though `mergeStateStatus` is `UNSTABLE` (consistent with the one failing check above).

**Verdict: requesting changes**, solely because of the failing `claude-review` check. There is nothing in the diff itself I'm asking to be changed — re-running/fixing that check (or a maintainer override) is the only blocker I see. Once that check is green (or resolved as a known-flaky infra issue), this PR looks mergeable to me as-is.

PR #4757, opened 2026-08-24
Serve file targets from the wiz tabular table contract

Nelson-cao:
A tabular table could only name a REST endpoint. `tabularSource` now carries `targetKind` (`"endpoint"` | `"file"`), and `target` generalizes the former `endpoint` field: an endpoint name for one kind, a path relative to the connector root (optionally suffixed `"#<sheet>"`) for the other. `params` carries a file connector's own parsing options. `endpoint` stays readable as a Jackson alias, so a caller still sending it binds instead of being told `target` is missing.

## Build path

`buildTabularTable` splits into contract resolution plus a shared build. The endpoint half is unchanged. The file half fills the query's `File`-typed property and validates `params` against the connector's own `PropertyMeta` map, so an option a connector declares is accepted and one it does not is refused by name rather than dropped. `requireRowCapWhenPaged` stays endpoint-only: a local file has no pages to walk and no per-call bill, so demanding a cap there would refuse a correct request. `target` is resolved against the connector root and refused when absolute, containing `..`, resolving outside the root, or absent.

A multi-sheet workbook with no sheet named fails and names the sheets. `ServerFileUtil.getColumnDefinition` and `getExcelSheetNames` both default to `sheets[0]`, so a three-sheet workbook would otherwise have reported success having annotated one.

## Probe

Annotation needs a column list and sample rows without leaving an asset behind, so `openProbeWorksheet`/`probeTable`/`closeProbeWorksheet` read a target through a temporary worksheet that is never persisted, clearing the assembly between files.

Sampling cannot come from `setSampleRows`: only `EndpointJsonQueryRunner` fills it, and a file's columns are read without executing the query at all. `probeTable` executes the built assembly through its own `AssetQuerySandbox` instead — private because `getTableLens` caches on name plus column-selection hash, and a loop reusing one table name over files with identical headers would answer the second with the first one's rows. The row bound reaches the connector as `XQuery.rowlimit`, so reading stops at the limit rather than trimming afterwards, and one extra row is read as the evidence for `sampleRowsTruncated`. Sampling failure leaves the sample absent and the probe successful.

The ceiling stays `rest.sample.rows`: a deployment that set it to 0 to stop sampled data leaving a tabular connector must not be circumvented by a different kind of target reaching the same data.

## Routes

Four routes expose this to wiz under its own permission checks rather than a composer session: `/tabular/browse` and `/tabular/probe/{open,table,close}`. Browse reports each entry's path relative to the connector root, which is the same string a build takes back as `target`.

## Tests

Appended to existing files (`WorksheetTableRequestTest`, `WorksheetTableServicePermissionTest`, the two `WizTabularController*Test`), no new test files. `inetsoft.web.wiz.**`: 2204 tests, 0 failures.

Consumed by inetsoft-technology/stylebi-wiz#1785.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizTabularController.java
@@ -24,8 +24,12 @@ import inetsoft.uql.DataSourceListing;
 import inetsoft.uql.DataSourceListingService;
 import inetsoft.uql.XDataSource;
 import inetsoft.uql.XRepository;
+import inetsoft.uql.tabular.LayoutCreator;
 import inetsoft.uql.tabular.TabularDataSource;
+import inetsoft.uql.tabular.TabularEditor;
+import inetsoft.uql.tabular.TabularQuery;
 import inetsoft.uql.tabular.TabularUtil;
+import inetsoft.uql.tabular.TabularView;
 import inetsoft.util.Catalog;
 import inetsoft.util.MessageException;
 import inetsoft.web.portal.data.DataSourceDefinition;
@@ -33,11 +37,18 @@ import inetsoft.web.portal.data.DatasourcesService;
 import inetsoft.web.security.PermissionPath;
 import inetsoft.web.security.RequiredPermission;
 import inetsoft.web.security.Secured;
+import inetsoft.web.wiz.model.WizTabularBrowseResult;
 import inetsoft.web.wiz.model.WizTabularListing;
 import inetsoft.web.wiz.model.WizTabularListings;
 import inetsoft.web.wiz.model.WizTabularSaveResult;
+import inetsoft.web.wiz.model.WorksheetTableResponse;
+import inetsoft.web.wiz.request.WizTabularBrowseRequest;
 import inetsoft.web.wiz.request.WizTabularCreateRequest;
+import inetsoft.web.wiz.request.WizTabularProbeCloseRequest;
+import inetsoft.web.wiz.request.WizTabularProbeOpenRequest;
+import inetsoft.web.wiz.request.WizTabularProbeTableRequest;
 import inetsoft.web.wiz.service.UnsupportedDatasourceException;
+import inetsoft.web.wiz.service.WorksheetTableService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
@@ -45,6 +56,8 @@ import org.springframework.http.MediaType;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.server.ResponseStatusException;
 
+import java.io.File;
+import java.nio.file.Files;
 import java.security.Principal;
 import java.util.*;
 
@@ -87,11 +100,13 @@ import java.util.*;
 public class WizTabularController {
    public WizTabularController(DatasourcesService datasourcesService,
                                SecurityEngine securityEngine,
-                               XRepository xrepository)
+                               XRepository xrepository,
+                               WorksheetTableService worksheetTableService)
    {
       this.datasourcesService = datasourcesService;
       this.securityEngine = securityEngine;
       this.xrepository = xrepository;
+      this.worksheetTableService = worksheetTableService;
    }
 
    /**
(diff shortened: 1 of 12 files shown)


jshobe-inetsoft (changes requested):
## Review of PR #4757 — "Serve file targets from the wiz tabular table contract"

I read the full diff (`gh pr diff 4757`), the PR description, and cross-checked the referenced base-branch (`stylebi-wiz-main-rebase`) APIs (`WorksheetService.openTemporaryWorksheet/getWorksheet/closeWorksheet`, `TabularEditor.Type.FILE`, `PropertyMeta`) to confirm the new reflection-based code compiles against real signatures.

### Summary of the change
- Generalizes `WorksheetTable.TabularSource` so a tabular table can bind to a file target (ServerFile) in addition to a REST endpoint, via a new `targetKind` + `target` (+ `params`) contract, with `endpoint` kept as a `@JsonAlias` for backward compatibility.
- `WorksheetTableService.buildTabularTable` splits into a shared build plus per-kind contract resolution (`applyEndpointContract` / `applyFileContract`), with careful path-traversal defenses in `resolveTargetFile`.
- Adds a "probe" mechanism (`openProbeWorksheet` / `probeTable` / `closeProbeWorksheet`) that builds a target inside a throwaway, never-persisted worksheet to read back columns and sample rows, honoring the existing `rest.sample.rows` ceiling.
- Adds a `/tabular/browse` endpoint that lists a file connector's directory tree using the connector's own `relativeTo`/`foldersOnly`/`acceptTypes` editor metadata.
- Substantial new test coverage appended to existing test files (no new test files), plus updates to two existing tests for the new controller constructor parameter.

### Correctness / completeness
- The new/changed methods are checked against real base-branch signatures and are consistent (`WorksheetService.openTemporaryWorksheet(Principal, AssetEntry)`, `getWorksheet(String, Principal)`, `closeWorksheet(String, Principal)`; `TabularEditor.Type.FILE`; `PropertyMeta.getDescriptor()`).
- `TabularSource.getEndpoint()`/`setEndpoint()` is removed from the Java bean API, but no other code in the repository calls those methods directly (only JSON, preserved via `@JsonAlias("endpoint")` on `getTarget()`), so this is not a Java-level breaking change. Confirmed via repo-wide search.
- The row-cap logic (`requireRowCapWhenPaged`) is correctly restricted to `targetKind == "endpoint"`, matching the stated rationale (a local file has no pages/per-call billing).
- The multi-sheet-workbook-without-a-named-sheet case is handled deliberately (fails and lists sheet names) rather than silently defaulting to sheet 0, closing a real correctness gap in the underlying `ServerFileUtil` default.
- Sampling is correctly gated to `targetKind == "file"` only, so the existing endpoint-sampling code path (`EndpointJsonQueryRunner`) is untouched — reducing regression risk on existing endpoint-backed tabular tables.
- `sandboxSampleLimit` correctly treats "already has a sample" and "not a file target" as short-circuits before consulting `rest.sample.rows`, and fails safe (limit 0) on an unparsable property value.

### Security
- **Path traversal**: `resolveTargetFile` rejects absolute paths, Windows drive-letter paths, and any `..` path segment, then re-validates the *canonicalized* resolved path against the canonicalized root (`startsWith(rootPath + separator)`), which correctly blocks a symlink inside the root that points outside it. This is the right defense-in-depth pattern.
- **Minor gap**: `browse()`/`collect()` does *not* apply the same canonical-containment check when *listing* a directory. If a connector root ever contains a symlink pointing outside the root, `browse` would list file names reached through that symlink (though `resolveTargetFile` would still refuse to actually read them when used as `target`, since that path re-validates canonically). This is a low-severity information-disclosure edge case (filenames only, not contents), and only reachable if the datasource root itself is misconfigured with an escaping symlink — worth a follow-up but not necessarily blocking.
- **Symlink cycles / DoS**: `collect()` recurses into subdirectories for `recursive=true` without cycle detection. A crafted or accidental symlink cycle inside the root could cause unbounded recursion/hang before `MAX_BROWSE_ENTRIES` is hit (the cap only limits *collected entries*, not directories visited when entries are being filtered out, e.g. an all-folders recursive walk with `foldersOnly` producing folder entries that do count, so this is probably bounded by the entry cap in most cases — but if hidden/unreadable files are being skipped, or accepted files are rare, a deep/cyclical tree could still take a while to walk before reaching the cap). Worth a sanity check, not necessarily blocking, since this endpoint is scoped to a connector root a caller must already have READ permission on.
- Resolution of connector-declared methods (`getRootFolder`, `isExcel`, `getExcelSheetNames`) via reflection-by-name (`callQueryMethod`) fails closed/soft (returns null and is treated as "no such capability") rather than failing open, which is appropriate.
- The `ThreadLocal` `TabularUtil.sessionId` handling in `browseTabularFiles` (bracketed `beginConnectorSession`/`endConnectorSession` in try/finally) correctly avoids leaking one user's session onto a pooled thread that serves another user next.

### Resource management / potential regressions
- **Probe worksheet lifecycle**: `openProbeWorksheet` explicitly documents "CLOSED BY THE CALLER." There is no server-side idle timeout, cap on concurrently open probe runtimes per user, or safety net if a caller crashes or forgets to call `/tabular/probe/close`. This creates a resource-leak / potential DoS surface: a caller (even one behaving correctly per the documented contract) that dies mid-probe leaves a live worksheet runtime and session open indefinitely (or until whatever generic session/runtime GC exists elsewhere). This isn't introduced from nothing — regular composer worksheets have the same "must be closed" contract — but it's worth confirming there is an existing idle-runtime reaper this relies on, since the PR doesn't add one specific to this new call path.
- **Concurrency**: The probe runtime holds a single mutable `Worksheet`; `probeTable` is documented as "sequential by construction" but nothing prevents two concurrent HTTP requests from racing on the same `runtimeId` (e.g. a buggy or malicious caller issuing overlapping `probe/table` calls). Worst case is corrupting a user's own throwaway worksheet, so blast radius is contained to the caller's own session, but it's a latent bug if wiz's caller ever parallelizes.
- Existing endpoint-based tabular table code paths appear untouched in behavior (verified `requireRowCapWhenPaged`, `applyEndpointContract`'s existing logic, and endpoint sampling are unchanged, only gated on `targetKind`), so regression risk to existing REST-connector functionality looks low.

### Backward compatibility
- `endpoint` remains a valid JSON alias, so an existing caller sending the old shape still binds (verified by the `tabularEndpointBindsAsTargetUnderItsOldName` test).
- Absent `targetKind` defaults to `"endpoint"`, preserving old semantics for old requests.
- No other in-repo Java caller depended on the removed `getEndpoint()`/`setEndpoint()` bean methods.
- New routes are additive; existing routes and their contracts (`/tabular/create`, `/tabular/update`, etc.) are unchanged. The frozen-mapped-paths test was updated to include the four new routes, which is the correct way to evolve that guard test rather than working around it.

### Tests
- New tests were appended to existing files rather than creating new test files, matching the PR's own stated approach and the project's likely preference for consolidated coverage per class.
- Coverage looks reasonably thorough: unknown `targetKind`, case-insensitive `targetKind`, per-kind "target is required" wording, probe permission-denied, probe build-failure-without-throwing, unnamed probe table naming, missing `runtimeId`, and a good matrix of `sandboxSampleLimit` cases (endpoint vs. file, zero/unset sampleRows, pre-existing sample, failed table, deployment ceiling clamp/off/unparsable).
- I did not find any test that was weakened, deleted, or made to assert different (weaker) behavior — the two existing controller test files were only updated to pass the new `WorksheetTableService` constructor dependency, which is a mechanical, non-semantic change.
- No obvious test gaps that would mask a regression, though I'd note there's no test directly covering `resolveTargetFile`'s traversal/symlink-escape rejection or `browse()`'s path handling at the unit level in the diff shown (they may be covered elsewhere, e.g. via `WizTabularControllerTest`/integration tests not shown in full, but I couldn't confirm this from the diff alone).

### Readability / maintainability
- The code is unusually heavily commented — almost essay-length Javadoc on nearly every method, explaining "why" decisions were made (why a fresh sandbox, why RUNTIME_MODE vs LIVE_MODE, why reflection over a type reference, etc.). This is generally a plus for a subtle, security-sensitive feature like this one, though it does make the methods quite long relative to the code they contain. Not a blocking concern, but future maintainers touching this file should expect a high comment-to-code ratio to keep in sync with any behavior change.
- Reflection use (`callQueryMethod`, `invokeWriteMethod`, `isFileProperty`) is justified (core has no compile-time dependency on the connector plugin types) and is consistently paired with fail-soft/fail-loud choices explained inline.

### Blocking items for this review
Per review policy, a PR must have **no failing CI checks** and **no merge conflicts** to be approved, regardless of code-quality findings.

- `gh pr view 4757 --json statusCheckRollup` shows the **`claude-review` check has `conclusion: FAILURE`**. Investigating the run shows it failed due to a tool-side error ("Claude encountered an error after 0s"), not a substantive code finding — but per review policy any FAILURE conclusion is treated as blocking regardless of cause.
- `mergeStateStatus` is `UNSTABLE` (consistent with the failing check above). `mergeable` is `MERGEABLE` (not `CONFLICTING`), so there is no merge-conflict blocker.
- All other checks (`build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla`) are passing.

### Verdict
**Requesting changes**, solely because of the failing `claude-review` CI check (a `claude-review` run error, not a substantive finding against this PR's code). The code itself is well-reasoned, thoroughly tested, backward compatible, and I found no blocking correctness, security, or regression issues — only a couple of low-severity hardening suggestions (symlink-escape check on `browse()`'s listing path, and a note on probe-worksheet lifecycle/idle cleanup) that are worth follow-up but are not merge-blocking on their own. Please re-run/fix the `claude-review` check (or otherwise get it to a non-failing state) and this PR should be in good shape to merge.

PR #4758, opened 2026-08-24
Thread rvs/source/refModelService through TableBindingMutator addField/removeField/moveField

larryliang-inetsoft:
## Summary

L6 reopened — confirmed, already-located defect from debugger-l6's investigation (found
alongside, but independent of, a separate render-layer crosstab named-group issue tracked
elsewhere).

`TableBindingMutator.addField`/`removeField`/`moveField` — backing the `add_table_field`/
`remove_table_field`/`move_table_field` MCP tools — all reapply a shelf's **entire** field
list through the 3-arg `setShelf(model, shelf, fields)` overload, which calls
`dimensions(fields, null, null, null)`. Because `rvs` is always `null` through that path,
**any field already on the shelf that carries a `namedGroup` has it silently dropped** the
next time one of those three tools touches that shelf — not just the field actually being
added/removed/moved, since the whole shelf's field list is rebuilt from scratch each call.
This gap was already flagged in the shipped code's own `@implNote` on the no-context
`dimensions(fields)` overload — a known, deliberately unclosed gap, not an oversight.

## Fix

Mechanical, per the debugger's own scoping — the 6-arg `setShelf` overload and
`FieldRefFactory.resolveNamedGroupInfo` already exist and are already proven correct by
#4731's own tests; no new resolution logic needed:

- `TableBindingMutator.addField`/`removeField`/`moveField` each get a new overload taking
  `RuntimeViewsheet`/`SourceInfo`/`DataRefModelFactoryService` and calling the existing 6-arg
  `setShelf` instead of the 3-arg one. The old no-context overloads now delegate to the new
  ones with nulls, so every existing caller/test (including direct `TableBindingMutatorTest`
  calls) is unaffected.
- `TableBindingService.addField`/`removeField`/`moveField` thread their already-in-scope
  `rvs`/`source` down via a new `applyWithContext` helper, mirroring `TableBindingService
  .setShelf`'s own existing `SourceInfo`-resolution pattern (`assembly instanceof
  DataVSAssembly`) exactly — the same precedent the debugger pointed at.

## A note on the regression test shape

The debugger's proposed test used the worksheet-local ("Expert") named-group shape. While
writing it, I found that shape can't round-trip through a second shelf rebuild **regardless
of this fix** — `FieldRefFactory.resolveNamedGroupInfo`'s `EXPERT_NAMEDGROUP_INFO` branch
never calls `NamedGroupInfoModel.setName(...)`, so a re-read of the model afterward recovers
`namedGroup()` as `null` on its own. That's a separate, pre-existing round-trip gap, not part
of this fix's scope (flagging it, not fixing it here). The regression tests instead use the
repository-registered (`ASSET_NAMEDGROUP_INFO_REF`) shape, which does set the name and so
round-trips correctly, cleanly isolating the actual defect under test: whether `rvs`/`source`
reach `setShelf` on a second call.

## Test plan
- [x] Added `addField`/`removeField`/`moveFieldPreservesAnAlreadyResolvedNamedGroupOn
      AnotherFieldOnTheSameShelf` to `TableBindingMutatorTest`
- [x] Verified all three genuinely catch the bug: temporarily made the new overloads call the
      3-arg `setShelf` instead of the 6-arg one (simulating the pre-fix behavior) — all three
      failed with the exact reported symptom (`order` reverting from `SORT_SPECIFIC`,
      `namedGroupInfo` becoming `null`). Restored the fix, reran — green.
- [x] `./mvnw -o test -pl core -Dtest=TableBindingMutatorTest,TableBindingServiceTest` —
      64/64 pass, existing tests unmodified
- [x] Broader sweep `-Dtest='inetsoft.web.wiz.binding.**,inetsoft.web.binding.**'` —
      329/329 pass
- [x] Full `core` module suite (`./mvnw -o test -pl core`) — 6991 tests, 0 failures, 0 errors

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/TableBindingMutator.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/TableBindingMutator.java
@@ -151,6 +151,29 @@ public final class TableBindingMutator {
    public static void addField(BaseTableBindingModel model, String shelf, FieldRef field,
                                Integer position)
    {
+      try {
+         addField(model, shelf, field, position, null, null, null);
+      }
+      catch(RuntimeException e) {
+         throw e; // preserve e.g. requireShelf/requireCompatible's IllegalArgumentException as-is
+      }
+      catch(Exception e) {
+         throw new RuntimeException(e);
+      }
+   }
+
+   /**
+    * @param rvs            the runtime viewsheet, so a field already on the shelf (or the one
+    *                       being added) carrying {@code namedGroup} can be resolved against a
+    *                       worksheet-local named group, the same as {@link #setShelf} below.
+    * @param sourceInfo     the crosstab's or table's own {@code SourceInfo}.
+    * @param refModelService needed to resolve a worksheet-local named group's conditions.
+    */
+   public static void addField(BaseTableBindingModel model, String shelf, FieldRef field,
+                               Integer position, RuntimeViewsheet rvs, SourceInfo sourceInfo,
+                               DataRefModelFactoryService refModelService)
+      throws Exception
+   {
       String name = requireShelf(model, shelf);
       requireCompatible(name, field);
       List<FieldRef> current = new ArrayList<>(read(model, name));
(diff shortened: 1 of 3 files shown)


larryliang-inetsoft (commented):
Reviewed diff against the described root cause. Findings:

- Fix matches the description exactly: new 6/7-arg `addField`/`removeField`/`moveField` overloads on `TableBindingMutator` thread `rvs`/`sourceInfo`/`refModelService` through to the existing 6-arg `setShelf`, and the old no-context overloads delegate to the new ones with nulls — so every existing caller/test keeps working unchanged. Confirmed the exception-wrapping in those delegating overloads (rethrow `RuntimeException` as-is, wrap other checked exceptions) correctly preserves the old no-`throws` signature while still surfacing `requireShelf`/`requireCompatible`'s `IllegalArgumentException` unchanged.
- `TableBindingService.applyWithContext()` is not new logic — verified it's an exact-pattern copy of the SourceInfo-resolution `TableBindingService.setShelf()` already does (`assembly instanceof DataVSAssembly` → `getSourceInfo()`, same `requireTableBinding`/`ApplyVSAssemblyInfoEvent`/`bindingModelService.setBinding` shape). Confirmed the old `apply()`/`Consumer<BaseTableBindingModel>` helper is left untouched for other callers (e.g. `setSort`) — no unintended scope change there.
- The fully-qualified `inetsoft.uql.asset.SourceInfo` (rather than an import) is deliberate, not sloppy — this file already imports the unrelated `inetsoft.web.binding.model.SourceInfo` DTO, so the FQN avoids a naming collision.
- The test-shape deviation is well-justified and disclosed rather than silently worked around: the debugger's proposed Expert/worksheet-local named-group shape has its own separate, pre-existing round-trip gap (`FieldRefFactory.resolveNamedGroupInfo`'s `EXPERT_NAMEDGROUP_INFO` branch never calling `NamedGroupInfoModel.setName(...)`), correctly identified as out of scope and flagged rather than fixed here, with the tests switched to the registered/`ASSET_NAMEDGROUP_INFO_REF` shape that actually isolates the real defect under test.
- Good verification discipline again: reverted just the context-threading (call 3-arg `setShelf` instead of 6-arg) and confirmed all 3 new tests fail with the exact reported symptom before restoring.

No blocking issues found. `build` was still pending as of this review — will follow up once it completes.

PR #4759, opened 2026-08-24
Add ScriptTarget.Kind exhaustiveness guard to SheetPairingService

larryliang-inetsoft:
## Summary

This PR originally fixed `SheetPairingService` rejecting `worksheetConditionValue` pairing mints. That production fix (adding `"worksheetConditionValue"` to `RECOGNIZED_KINDS`/`WORKSHEET_COLUMN_KINDS`) already landed on `main` independently via #4789, along with the same three mirror tests this PR added -- so this branch has been rebased to drop everything that's now redundant and keep only what #4789 did not add:

- **`everyScriptTargetKindIsRecognized`**: an exhaustiveness test iterating `ScriptTarget.Kind.values()` against `RECOGNIZED_KINDS`, catching a new `Kind` added but never wired into this hand-maintained list -- this is the most valuable part of the original PR and #4789 did not add it. Required widening `RECOGNIZED_KINDS`'s visibility from `private` to package-private for the same-package test to reference it.
- **`SheetPairingService` javadoc corrections**: #4789 updated the two wire-kind lists but left the surrounding prose stale (still said "the two worksheet kinds..." after a third kind was added). Fixed the wording on `WORKSHEET_COLUMN_KINDS` and `validateWorksheetColumnContext`.
- **Header docblock index catch-up**: #4789 added its three tests without adding class-level index entries for them. Added the three `conditionValue` entries plus `[Kinds: exhaustive]`.

Retitled from "Fix worksheetConditionValue pairing mint rejection" since the fix itself is no longer part of this diff.

## Test plan
- [x] `SheetPairingServiceTest`: 31/31 passing (`mvnw -pl core test -Dtest=SheetPairingServiceTest -o`).
- [x] Rebased cleanly onto `stylebi-wiz-main-rebase`; `SheetPairingService.java` merges with zero conflicts (both sides made the same list additions independently); the only conflict was in the test file, where the now-duplicate mirror tests were dropped in favor of #4789's existing versions.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/pairing/SheetPairingService.java
+++ b/core/src/main/java/inetsoft/web/wiz/pairing/SheetPairingService.java
@@ -59,7 +59,7 @@ public class SheetPairingService {
     * packages depend on each other in both directions. An unrecognized kind is refused here, at
     * mint time, rather than left to surface later at {@code ScriptTarget.Kind.fromWire}.
     */
-   private static final List<String> RECOGNIZED_KINDS = List.of(
+   static final List<String> RECOGNIZED_KINDS = List.of(
       "viewsheetOnInit", "viewsheetOnLoad", "assemblyMain", "assemblyOnClick", "calcField",
       "worksheetExpression", "worksheetCondition", "worksheetConditionValue");
 
@@ -255,10 +255,10 @@ public class SheetPairingService {
    }
 
    /**
-    * The two worksheet kinds addressed by (table, field) rather than by assembly alone --
+    * The worksheet kinds addressed by (table, field) rather than by assembly alone --
     * exactly like {@code calcField}, and validated the same way (whole-branch review finding 3).
     *
-    * <p>Before this, both fell through to the generic "does the assembly exist" branch, which
+    * <p>Before this, these fell through to the generic "does the assembly exist" branch, which
     * neither requires nor verifies {@code name}. "New expression column" opens the formula editor
     * with no {@code formulaName}, so the mint carried no name; the mint SUCCEEDED, {@code status}
     * reported a healthy pane-scoped session, and then {@code PaneScopeService.matchesGrant}
@@ -271,15 +271,15 @@ public class SheetPairingService {
       List.of("worksheetExpression", "worksheetCondition", "worksheetConditionValue");
 
    /**
-    * Verifies a {@code worksheetExpression}/{@code worksheetCondition} context names a column
-    * that actually exists, mirroring the {@code calcField} branch against the
-    * {@link TableAssembly}'s own column selection.
+    * Verifies a {@code worksheetExpression}/{@code worksheetCondition}/
+    * {@code worksheetConditionValue} context names a column that actually exists, mirroring the
+    * {@code calcField} branch against the {@link TableAssembly}'s own column selection.
     *
     * <p>{@code worksheetExpression} is checked against EXPRESSION columns specifically, matching
     * on the same {@code ExpressionRef} name/attribute pair {@code WorksheetScriptService} reads
     * and writes through, so a grant cannot be minted for a location that service would then
-    * refuse. {@code worksheetCondition} is checked against any column, since adding a condition
-    * to a field that has none yet is legitimate.
+    * refuse. {@code worksheetCondition} and {@code worksheetConditionValue} are checked against
+    * any column, since adding a condition to a field that has none yet is legitimate.
     */
    private static void validateWorksheetColumnContext(RuntimeSheet rs, EditorContext ctx)
       throws PairingException
--- a/core/src/test/java/inetsoft/web/wiz/pairing/SheetPairingServiceTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/pairing/SheetPairingServiceTest.java
@@ -31,6 +31,7 @@ import inetsoft.uql.asset.Worksheet;
 import inetsoft.uql.viewsheet.CalculateRef;
 import inetsoft.uql.viewsheet.VSAssembly;
 import inetsoft.uql.viewsheet.Viewsheet;
+import inetsoft.web.wiz.script.ScriptTarget;
 import org.junit.jupiter.api.Test;
 
 import java.security.Principal;
@@ -75,6 +76,14 @@ import static org.mockito.Mockito.when;
  * [EditorContext: IDOR]      mint against a runtimeId the caller does not own fails identically
  *                            whether or not the named assembly exists -- the failure must never
  *                            become an oracle for what exists on someone else's open sheet
+ * [EditorContext: conditionValue ok]   mint accepts a worksheetConditionValue editorContext
+ *                            naming a column the worksheet runtime has
+ * [EditorContext: conditionValue no name] mint refuses a worksheetConditionValue context that
+ *                            carries no name
+ * [EditorContext: conditionValue no column] mint refuses a worksheetConditionValue context
+ *                            naming a column the table does not have
+ * [Kinds: exhaustive]        every ScriptTarget.Kind wire name is in RECOGNIZED_KINDS -- catches
+ *                            a new Kind added there but never wired into this hand-maintained list
  */
 /*
  * @WizAgentTestSupport (which itself carries @Tag("core")) replaces a bare @Tag: the
@@ -464,6 +473,20 @@ class SheetPairingServiceTest {
       assertEquals(PairingException.Kind.INVALID_ARGUMENT, ex.getKind());
       assertTrue(ex.getMessage().contains("NoSuchColumn"),
                  "message must name what was asked for: " + ex.getMessage());
+   }
+
+   /**
+    * Would have caught stylebi's worksheetConditionValue regression the moment
+    * {@code ScriptTarget.Kind.WORKSHEET_CONDITION_VALUE} was added but never wired into
+    * {@code RECOGNIZED_KINDS} -- mirrors {@code PaneScopeServiceTest#everyKindIsClassifiedByExactlyOneGuard}.
+    */
+   @Test
+   void everyScriptTargetKindIsRecognized() {
+      for(ScriptTarget.Kind k : ScriptTarget.Kind.values()) {
+         assertTrue(SheetPairingService.RECOGNIZED_KINDS.contains(k.wireName()),
+            "ScriptTarget.Kind '" + k.wireName() + "' is missing from " +
+            "SheetPairingService.RECOGNIZED_KINDS -- mint would refuse it as an unknown kind.");
+      }
    }
 
    /**


larryliang-inetsoft (commented):
Reviewed diff against the described root cause, plus independently verified the visibility-change concern raised.

**Fix correctness**
- Diff is exactly the described 2-line addition (`"worksheetConditionValue"` added to both `RECOGNIZED_KINDS` and `WORKSHEET_COLUMN_KINDS`), plus doc-comment updates reflecting the third kind. No other logic touched — confirmed `validateWorksheetColumnContext`'s `hasColumn`/`hasExpressionColumn` ternary already defaults every non-`worksheetExpression` kind (including the new one) to the correct `hasColumn` check, so nothing else needed to change there.

**Visibility change (`RECOGNIZED_KINDS`: `private` -> package-private)** — checked this specifically, not on faith:
- Confirmed via `git grep` on the base branch (`stylebi-wiz-main-rebase`) that `RECOGNIZED_KINDS` has exactly 3 references, all inside `SheetPairingService.java` itself (declaration + 2 internal uses) — no external package reaches it today, so this widening is a genuine no-op for every other caller.
- This is *not* an existing pattern in this specific file (every other internal constant list in this class, e.g. `ASSEMBLY_REQUIRED_KINDS`/`WORKSHEET_COLUMN_KINDS`, stays `private`) — so it is a real, if minimal and narrowest-possible, visibility change, not a no-op restatement of an existing pattern.
- The PR's own claim that this "mirrors `PaneScopeServiceTest#everyKindIsClassifiedByExactlyOneGuard`" is only partially true: that test actually maintains its **own duplicated** `EXPRESSION_LEVEL_KINDS`/`SHEET_LEVEL_KINDS` lists inside the test file, not a widened reference to production fields. On reflection, this PR's approach (reading the real production list directly) is actually the *stronger* regression guard for this specific bug shape: a duplicated test-side list would require remembering to update *two* places in sync, and forgetting one is exactly the failure mode that produced this bug in the first place. Reading the production field directly is what makes `everyScriptTargetKindIsRecognized` a genuine "would have caught this" test rather than another list that could itself drift.
- Net: the visibility widening is justified, minimal (private -> package-private, not public), and matches this class's own narrowest-necessary-scope convention even if the specific mechanism differs slightly from the cited precedent.

**Tests**: the 3 mirror tests (mint-success/no-name/no-column) and the exhaustiveness test look correctly targeted; the revert-and-confirm-4-tests-fail discipline described in the PR is exactly the kind of real verification this team has been asking for.

No blocking issues found. `build` was still pending as of this review — will follow up once it completes.

jshobe-inetsoft (changes requested):
## Summary

This PR adds `"worksheetConditionValue"` to `SheetPairingService`'s `RECOGNIZED_KINDS` and `WORKSHEET_COLUMN_KINDS` hand-maintained wire-kind lists, fixing a bug where every mint attempt for the `ScriptTarget.Kind.WORKSHEET_CONDITION_VALUE` kind was rejected with `"Unknown editorContext.kind"` before real validation ran. It also adds an exhaustiveness test and mirror mint tests for the new kind.

I verified the fix directly:
- `ScriptTarget.Kind.WORKSHEET_CONDITION_VALUE` has wire name `"worksheetConditionValue"` (`core/src/main/java/inetsoft/web/wiz/script/ScriptTarget.java`), matching exactly what was added to both lists.
- `validateWorksheetColumnContext`'s existing `hasColumn`/`hasExpressionColumn` ternary keys only on `"worksheetExpression"` for the expression-column-only check, so every other kind (including the newly-added `worksheetConditionValue`) correctly falls through to the more permissive `hasColumn` check — no additional branching was needed, consistent with the PR description.
- The new tests (`mintsAWorksheetConditionValueContextWhoseColumnExists`, `refusesAWorksheetConditionValueContextThatCarriesNoName`, `refusesAWorksheetConditionValueContextNamingAColumnTheTableDoesNotHave`) correctly mirror the existing `worksheetCondition` coverage.
- `everyScriptTargetKindIsRecognized` is a good addition — it iterates `ScriptTarget.Kind.values()` against `RECOGNIZED_KINDS`, which would have caught exactly this regression at the time `WORKSHEET_CONDITION_VALUE` was introduced, and guards against the same class of bug recurring for any future kind.
- Widening `RECOGNIZED_KINDS` from `private` to package-private (default) visibility for the same-package test is a reasonable, minimal-scope change; it isn't exposed outside the package.

### Code quality / correctness
- The change is minimal, well-scoped, and directly addresses the root cause (a hand-maintained list falling out of sync with the enum). No unrelated refactoring.
- Javadoc updates ("The two worksheet kinds..." → "The worksheet kinds...") are accurately updated to reflect the third kind being added.
- No edge cases, error handling, or security concerns are introduced — the change purely widens an allow-list to match an already-wired enum value; it does not weaken any validation (blank-name and missing-column checks still apply identically to the new kind, as the added tests confirm).
- No backward-incompatibility: this only broadens acceptance of a previously-rejected (broken) input; no existing accepted behavior changes.
- No regressions apparent: `worksheetExpression` handling (checked against expression columns specifically) is untouched; `worksheetCondition`/`worksheetConditionValue` share the same any-column check as before.
- Test coverage is thorough and mirrors existing patterns in the file (mint-success / blank-name / missing-column), plus the new exhaustiveness guard.

### Blocking issues for this review

Per review policy, I must request changes because of CI/mergeability state, independent of the code quality assessment above:

1. **Failing check**: The `claude-review` status check on this PR is currently `FAILURE` (https://github.com/inetsoft-technology/stylebi/actions/runs/32733044145/job/97449422966). The failure log shows the automated Claude review action itself errored out (`is_error: true`, likely a missing/invalid `ANTHROPIC_API_KEY` in that workflow run) rather than a code-quality finding, but per review policy any check with conclusion `FAILURE` — including `claude-review` — is treated as a failing check that blocks approval.
2. **`build` check still in progress**: The `Validation Build` workflow was `IN_PROGRESS` (not yet completed) at review time, so it cannot yet be confirmed green.
3. **`mergeStateStatus` is `UNSTABLE`**: consistent with the outstanding/failing checks above (not `DIRTY`, and `mergeable` is `MERGEABLE`, so there are no merge conflicts).

## Verdict

**Requesting changes** — not due to any defect found in the code itself (the fix is correct, minimal, and well-tested), but because the `claude-review` check is currently failing and the `build` check has not completed, which per review policy blocks approval. Please re-run/fix the `claude-review` workflow (it appears to be an infrastructure/config issue unrelated to this PR's diff) and confirm `build` passes, then this PR should be good to merge as-is.

jshobe-inetsoft (changes requested):
## Manual review (replaces the automated `claude-review` check, which errored out during setup — not a review finding, see [run 32733044145](https://github.com/inetsoft-technology/stylebi/actions/runs/32733044084/job/97449422966), `is_error: true`, failed in ~1s with 0 turns/0 cost, before doing any review work)

### The fix itself is correct

`ScriptTarget.Kind.WORKSHEET_CONDITION_VALUE` has wire name `worksheetConditionValue` (confirmed in `ScriptTarget.java`), and it was genuinely missing from `SheetPairingService`'s hand-maintained `RECOGNIZED_KINDS`/`WORKSHEET_COLUMN_KINDS` lists, so the diagnosis and fix described in the PR body are sound. The `hasColumn`/`hasExpressionColumn` handling for the new kind (checked against any column, not expression-only) matches the sibling `worksheetCondition` kind correctly, and the three new mint tests mirror the existing `worksheetExpression`/`worksheetCondition` coverage appropriately.

### This PR has an actual merge conflict — and it's not a trivial one

`gh pr view` reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. I traced it: **the exact same underlying bug (missing `worksheetConditionValue` in `RECOGNIZED_KINDS`/`WORKSHEET_COLUMN_KINDS`) has already been fixed independently on the base branch (`stylebi-wiz-main-rebase`)** by a different commit, with equivalent logic but different comment wording and (in the base's version) without the exhaustiveness test. Diffing this PR against the current base:

- The `RECOGNIZED_KINDS`/`WORKSHEET_COLUMN_KINDS` list additions and the `validateWorksheetColumnContext` javadoc/behavior are **already present on the base branch** — those hunks conflict purely on wording, not on substance.
- The genuinely new content this PR still adds on top of current base is: (1) widening `RECOGNIZED_KINDS` from `private` to package-private, and (2) the `everyScriptTargetKindIsRecognized` exhaustiveness test (which is valuable and not yet on base) plus its two supporting mint tests, which conflict textually with equivalent tests already on base for the same scenarios.

So this isn't a simple rebase-and-done — the author needs to reconcile with what already landed: most likely, rebase onto current `stylebi-wiz-main-rebase`, drop the now-redundant list/javadoc changes (already there), and keep just the visibility change plus the `everyScriptTargetKindIsRecognized` test (folding it in alongside whatever mint tests already exist on base for `worksheetConditionValue`, without duplicating them).

### Verdict

**Requesting changes**, not because of any defect in the code as written, but because:
- The PR is not mergeable (`CONFLICTING`/`DIRTY`).
- The conflict isn't cosmetic — the base branch already carries an equivalent fix from elsewhere, so simply resolving the conflict textually risks re-litigating wording rather than recognizing the substantive fix is already in and only the exhaustiveness test is still worth carrying forward.

(Other checks — `build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla` — are all green; `claude-review`'s failure was an infrastructure error and is not counted against this PR.)

jshobe-inetsoft (changes requested):
## Review summary

**Result: Request changes.** The two conditions below are not both satisfied:

- **Merge conflicts exist.** `gh pr view 4759 --json mergeable,mergeStateStatus` reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. I confirmed this is a real conflict, not a stale GitHub cache value, by fetching both branches and running an actual `git merge --no-commit --no-ff` of `fix/worksheet-condition-value-pairing` into the current tip of the base branch `stylebi-wiz-main-rebase`: it produces a genuine content conflict in `core/src/test/java/inetsoft/web/wiz/pairing/SheetPairingServiceTest.java`.
- **Root cause of the conflict — the fix is now redundant.** The base branch (`stylebi-wiz-main-rebase`) has moved 159 commits ahead of this PR's fork point, and one of those commits already fixed the exact bug this PR targets: the current tip of the base branch already has `"worksheetConditionValue"` added to both `RECOGNIZED_KINDS` and `WORKSHEET_COLUMN_KINDS` in `SheetPairingService.java`, and already has near-identical `mintsAWorksheetConditionValueContextWhoseColumnExists` / `refusesAWorksheetConditionValueContextThatCarriesNoName` / `refusesAWorksheetConditionValueContextNamingAColumnTheTableDoesNotHave` test methods in `SheetPairingServiceTest.java` (same method names, same bodies, only the javadoc wording differs). Diffing this PR's test file against the base branch's current version confirms it: nearly the entire diff is now no-op (identical code, reworded comments), and the *only* substantively new content is the `everyScriptTargetKindIsRecognized` exhaustiveness test plus the `RECOGNIZED_KINDS` visibility widening (`private` → package-private) it depends on.

  The check failures/mergeable status aside, this means the PR as currently written can't simply be conflict-resolved by picking one side — the author should rebase onto the current `stylebi-wiz-main-rebase` tip, drop the now-duplicate list/javadoc edits (they're already present upstream), and keep only the two genuinely new pieces: the `RECOGNIZED_KINDS` visibility widening and the `everyScriptTargetKindIsRecognized` test.

- **CI checks**: all checks other than `claude-review` (which this manual review replaces per project convention) are passing: `Semgrep OSS`, `Semgrep Scan`, `build`, and `license/cla` all pass. This condition is satisfied on its own, but doesn't offset the conflict above.

Given the conflict, this PR cannot be merged in its current state regardless of the quality of the changes it introduces. Detailed review of the code as authored follows, for when it's rebased.

---

## Review of the underlying change (as authored, pre-conflict)

### Correctness of the core fix
The diagnosis in the PR description is accurate and I verified it against `ScriptTarget.Kind`: `WORKSHEET_CONDITION_VALUE` has wire name `"worksheetConditionValue"` and is fully wired through `PaneScopeService`, the script services, `WorksheetScriptService`, and `ScriptGrammar.supportedKinds()` — `SheetPairingService`'s two hand-maintained lists (`RECOGNIZED_KINDS`, `WORKSHEET_COLUMN_KINDS`) were the only place missing it. Adding the wire string to both lists is the correct, minimal fix; the existing `hasColumn`/`hasExpressionColumn` ternary in `validateWorksheetColumnContext` already routes any kind other than `worksheetExpression` (which now includes `worksheetConditionValue`) to the `hasColumn` any-column check, which is the right validation semantics for a condition-value context (matches how `worksheetCondition` itself is validated). No logic changes beyond the list additions were needed or made.

### Clean code / readability
- The javadoc updates (pluralizing "two worksheet kinds" → "worksheet kinds", updating the kind list mentioned in comments) are accurate and keep the comments in sync with the code. Good practice, though as noted above they're now duplicated by an independent upstream fix and will need to be dropped/reconciled on rebase.
- Widening `RECOGNIZED_KINDS` from `private` to package-private solely to let the same-package test iterate it directly is a reasonable, low-risk trade-off (mirrors the existing `PaneScopeServiceTest#everyKindIsClassifiedByExactlyOneGuard` pattern called out in the PR description) rather than adding a public accessor or duplicating the list in the test.

No new error-handling paths are introduced; the existing `INVALID_ARGUMENT` refusal path is simply extended to no longer misfire for this one kind. The blank-name and missing-column edge cases are covered by the new tests, mirroring the existing `worksheetExpression`/`worksheetCondition` coverage.

### Test quality
- The three mirror tests (`mintsAWorksheetConditionValueContextWhoseColumnExists`, `refusesAWorksheetConditionValueContextThatCarriesNoName`, `refusesAWorksheetConditionValueContextNamingAColumnTheTableDoesNotHave`) correctly parallel the existing `worksheetCondition` tests and would have caught the original bug (they fail against the old two-item lists, per the PR's stated verification step).
- `everyScriptTargetKindIsRecognized` is the standout addition: an exhaustiveness test over `ScriptTarget.Kind.values()` that fails immediately if a future `Kind` is added without updating `RECOGNIZED_KINDS`, closing off exactly the class of bug this PR fixes. This is good regression-proofing and its value is independent of the redundant list/javadoc edits — it's the part most worth preserving through a rebase.
- No broken/removed tests; only additions plus comment-only edits to existing code.

### Security
No security concerns. This is an allow-list correction inside an internal validation method; no new external input handling, no widened trust boundary. The ownership/IDOR check above this code path (`sameLogicalUser`) is untouched.

### Backward compatibility
None. Purely additive to an allow-list; no public API, wire format, or persisted-data change. The `private` → package-private visibility change on `RECOGNIZED_KINDS` has no external effect (it's still not part of any public API).

### Regression risk
Low for the fix itself. The only regression-risk vector is the merge conflict described above — if resolved incorrectly (e.g., by blindly taking "ours" or "theirs" without reconciling), it could silently drop the new exhaustiveness test or reintroduce stale javadoc. This should be handled by rebasing and re-authoring the diff against current `stylebi-wiz-main-rebase`, not by a mechanical conflict resolution.

---

## Action needed before this can be approved
1. Rebase `fix/worksheet-condition-value-pairing` onto the current tip of `stylebi-wiz-main-rebase`.
2. Drop the `RECOGNIZED_KINDS`/`WORKSHEET_COLUMN_KINDS` list edits and the javadoc rewording — already present upstream.
3. Keep and re-apply: the `RECOGNIZED_KINDS` visibility widening (`private` → package-private) and the `everyScriptTargetKindIsRecognized` test (plus the `ScriptTarget` import it needs).
4. Re-verify `SheetPairingServiceTest` passes against the rebased base, then re-open/update the PR for a clean, conflict-free diff.

jshobe-inetsoft (changes requested):
## Manual review (replaces the automated `claude-review` check, which continues to error out during setup — an infrastructure issue, not a review finding — so its `FAILURE` conclusion is excluded from this evaluation per project convention)

### The underlying fix is correct

`ScriptTarget.Kind.WORKSHEET_CONDITION_VALUE` has wire name `worksheetConditionValue`, and it was genuinely missing from `SheetPairingService`'s hand-maintained `RECOGNIZED_KINDS` / `WORKSHEET_COLUMN_KINDS` lists, causing every mint attempt for this kind to fail with `PairingException(INVALID_ARGUMENT, "Unknown editorContext.kind...")`. Adding the wire string to both lists is the correct, minimal fix:

- `validateWorksheetColumnContext`'s existing `hasColumn`/`hasExpressionColumn` ternary keys only on `worksheetExpression` for the expression-only check; every other kind (including the new one) already falls through to the more permissive `hasColumn` check, matching how `worksheetCondition` is validated. No additional branching was needed or added.
- The three new mint tests (`mintsAWorksheetConditionValueContextWhoseColumnExists`, `refusesAWorksheetConditionValueContextThatCarriesNoName`, `refusesAWorksheetConditionValueContextNamingAColumnTheTableDoesNotHave`) correctly mirror the existing `worksheetCondition` coverage.
- `everyScriptTargetKindIsRecognized` is a genuinely valuable addition: an exhaustiveness test over `ScriptTarget.Kind.values()` that would fail immediately if a future `Kind` is added without updating `RECOGNIZED_KINDS`. This closes off exactly the class of bug this PR fixes and is the most durable part of the change.
- Widening `RECOGNIZED_KINDS` from `private` to package-private so the same-package test can reference it directly (rather than duplicating the list in the test, which is exactly the kind of drift that caused this bug) is a reasonable, minimal-scope trade-off. It has no effect outside the package and is not part of any public API.
- No new error-handling paths are introduced; the existing `INVALID_ARGUMENT` refusal path is simply no longer misfiring for this one kind. Blank-name and missing-column edge cases are covered by the new tests.
- No security concerns: this is an allow-list correction inside internal validation, with no new external input handling and no change to the ownership/IDOR check (`sameLogicalUser`) above this code path.
- No backward incompatibility: purely additive to an allow-list; no public API, wire format, or persisted-data change.
- No regression risk from the diff itself: `worksheetExpression` handling (expression-column-specific) is untouched, and `worksheetCondition`/`worksheetConditionValue` share the same any-column check as before.
- No tests are broken or removed — only additions plus comment/javadoc wording updates.

### Blocking issue: the PR is not mergeable

`gh pr view 4759 --json mergeable,mergeStateStatus` reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. I verified this is a real conflict, not a stale cache value, with `git merge-tree` between the current tips of `fix/worksheet-condition-value-pairing` and `stylebi-wiz-main-rebase`: it produces a genuine content conflict in `core/src/test/java/inetsoft/web/wiz/pairing/SheetPairingServiceTest.java`.

Root cause: the base branch has since moved ahead and already carries an equivalent fix for the same bug — `RECOGNIZED_KINDS`/`WORKSHEET_COLUMN_KINDS` on base already include `"worksheetConditionValue"`, and `SheetPairingServiceTest.java` on base already has near-identical `mintsAWorksheetConditionValueContextWhoseColumnExists` / `refusesAWorksheetConditionValueContextThatCarriesNoName` / `refusesAWorksheetConditionValueContextNamingAColumnTheTableDoesNotHave` tests (same names/bodies, different javadoc wording). Diffing this PR's head against the current base confirms most of the diff is now redundant; the only substantively new content is:

- The `RECOGNIZED_KINDS` visibility widening (`private` → package-private) and the `everyScriptTargetKindIsRecognized` exhaustiveness test that depends on it (plus the `ScriptTarget` import it needs).
- Some javadoc wording refinements not present on base.

This can't be resolved by a mechanical "pick one side" merge without risking silently dropping the exhaustiveness test or reintroducing already-superseded wording. The PR needs to be rebased onto the current `stylebi-wiz-main-rebase` tip, with the now-redundant list/javadoc edits dropped and only the visibility change + exhaustiveness test (and any wording improvements not already upstream) re-applied.

### Checks

All checks other than `claude-review` are passing: `build` (Validation Build), `Semgrep Scan`, `Semgrep OSS`, and `license/cla` are all green. `claude-review`'s `FAILURE` is an infrastructure/setup error (errors out in ~1s with 0 turns), not a review finding, and is excluded from this evaluation per project convention.

## Verdict: Request changes

Per review policy, approval requires no blocking code issues, no failing checks (excluding `claude-review`), and no merge conflicts — all three must hold. The code change itself is correct, minimal, and well-tested, but the PR currently has a real, substantive merge conflict against its base branch caused by an equivalent fix having already landed upstream. That alone blocks approval regardless of code quality.

**Action needed before this can be approved:**
1. Rebase `fix/worksheet-condition-value-pairing` onto the current tip of `stylebi-wiz-main-rebase`.
2. Drop the `RECOGNIZED_KINDS`/`WORKSHEET_COLUMN_KINDS` list edits and javadoc rewording already present upstream.
3. Keep and re-apply the `RECOGNIZED_KINDS` visibility widening and the `everyScriptTargetKindIsRecognized` test (folded in alongside whatever `worksheetConditionValue` mint tests already exist on the rebased base, without duplicating them).
4. Re-verify `SheetPairingServiceTest` passes against the rebased base, then update the PR for a clean, conflict-free diff.

jshobe-inetsoft (changes requested):
## Code review — Fix worksheetConditionValue pairing mint rejection

The diagnosis in the description is correct and I confirmed each part of it against the source: `ScriptTarget.Kind.WORKSHEET_CONDITION_VALUE` carries wire name `worksheetConditionValue` and a real `Location`, `validateEditorContext`'s first gate rejects anything outside `RECOGNIZED_KINDS` before any other validation runs, and `validateWorksheetColumnContext`'s `"worksheetExpression".equals(ctx.kind()) ? hasExpressionColumn(...) : hasColumn(...)` ternary does default the new kind to the any-column check — which is right, since a single-value condition's value can sit on a plain column as easily as an expression one.

The problem is not the analysis. It is that the branch has gone stale underneath it.

---

### Blocking 1 — merge conflict

`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. `git merge-tree` against the current `stylebi-wiz-main-rebase` auto-merges `SheetPairingService.java` cleanly but conflicts in `SheetPairingServiceTest.java`.

### Blocking 2 — the production fix already landed; resolving the conflict naively will not compile

Commit `b967dd0c0b`, "Fix worksheetConditionValue rejected at pairing mint time (#4789)", is on `stylebi-wiz-main-rebase` and already contains **both** list additions:

```java
   private static final List<String> RECOGNIZED_KINDS = List.of(
      "viewsheetOnInit", "viewsheetOnLoad", "assemblyMain", "assemblyOnClick", "calcField",
      "worksheetExpression", "worksheetCondition", "worksheetConditionValue");
...
   private static final List<String> WORKSHEET_COLUMN_KINDS =
      List.of("worksheetExpression", "worksheetCondition", "worksheetConditionValue");
```

#4789 also added `mintsAWorksheetConditionValueContextWhoseColumnExists`, `refusesAWorksheetConditionValueContextThatCarriesNoName` and `refusesAWorksheetConditionValueContextNamingAColumnTheTableDoesNotHave` — the same three tests this PR adds, body-for-body identical down to the `"Query1"` / `"Amount"` / `"NoSuchColumn"` fixtures, differing only in javadoc wording. That is what the test-file conflict is: both sides inserted the same three methods at the same point.

So resolving the conflict by keeping both sides produces three duplicate method names and a compile error, and resolving it by taking `main`'s side throws away the parts of this PR that are still worth having. This needs a real rebase, not a conflict resolution.

### What is still genuinely new and should survive the rebase

Not all of this PR is redundant — three things are absent from `main` and worth keeping:

1. **`everyScriptTargetKindIsRecognized`** and the `private` → package-private widening it needs. This is the most valuable part of the PR and #4789 did not add it. #4789 fixed the instance; this fixes the class of bug, and it mirrors an established pattern in the same area (`PaneScopeServiceTest#everyKindIsClassifiedByExactlyOneGuard`).

2. **The `SheetPairingService` javadoc corrections.** #4789 changed the two lists but left the prose stale, so on `main` today the comments still read "The two worksheet kinds addressed by (table, field)" and "Verifies a `worksheetExpression`/`worksheetCondition` context names a column that actually exists" above lists and a method that now handle three kinds. Your wording fixes are still needed and still apply cleanly.

3. **The header-docblock index lines.** #4789 added its three tests without adding index entries for them, so the class-level index on `main` still stops at `[EditorContext: IDOR]`. Your four lines (three `conditionValue` entries plus `[Kinds: exhaustive]`) bring it back in sync — in a file that maintains that index deliberately, that is not cosmetic.

A rebase that keeps 1–3 and drops the now-duplicated three tests and the two list edits would be a small, clearly-valuable PR. Retitling it to reflect what it actually does at that point ("add a kind-exhaustiveness guard to SheetPairingService" or similar) would help, since "Fix worksheetConditionValue pairing mint rejection" would no longer describe the diff.

---

### Non-blocking, on the code that remains

1. **The exhaustiveness test only guards one direction.** It asserts every `Kind.wireName()` is in `RECOGNIZED_KINDS`, but not that every entry in `RECOGNIZED_KINDS` is a real wire name. The reverse direction is the one that catches a typo in the hand-maintained list — `"worksheetConditionvalue"` would pass mint's gate and then fail later at `ScriptTarget.Kind.fromWire`, which is *precisely* the failure mode the field's own javadoc says this list exists to prevent ("An unrecognized kind is refused here, at mint time, rather than left to surface later at `ScriptTarget.Kind.fromWire`"). Set equality costs no more than the loop:

   ```java
   assertEquals(Arrays.stream(ScriptTarget.Kind.values())
                   .map(ScriptTarget.Kind::wireName).collect(Collectors.toSet()),
                Set.copyOf(SheetPairingService.RECOGNIZED_KINDS));
   ```

   The failure message is less self-naming than yours, so keeping the loop for the forward direction and adding one containment assertion for the reverse is fine too.

2. **The widened field now contradicts its own javadoc.** That javadoc goes out of its way to explain why the list does *not* reference `ScriptTarget` ("reaching back into it would make the two packages depend on each other in both directions"), and after this change the field is no longer private and the test in this package does import `inetsoft.web.wiz.script.ScriptTarget`. Both are fine — the cycle argument is architectural and test code is not bound by it — but in a class this scrupulous about justifying every decision, the silent contradiction will read as an oversight to the next person. One sentence on the field ("package-private so `SheetPairingServiceTest` can assert exhaustiveness against `ScriptTarget.Kind`; the cycle argument above is about production code") settles it.

3. **Worth a thought, not a change:** the test hard-codes the rule "every `Kind` is mintable". If a future `Kind` is deliberately not reachable through pairing, the test forces it into `RECOGNIZED_KINDS` anyway rather than making the exclusion explicit. `PaneScopeService` solved the analogous problem with a positive classification rather than a list; if a non-mintable kind ever appears, that is the shape to reach for. Nothing to do today — every current `Kind` is mintable.

### Things I checked and found clean

- The three mirror tests are correct in themselves (they are already on `main`), and the `tableWithColumn` real-`TableAssembly` fixture is the right call given `getColumnSelection` is final.
- `hasColumn` vs `hasExpressionColumn` routing for the new kind is correct: `FormulaEditorDialog`'s single-value condition branch can target a plain column, so the any-column check is the right one and the more permissive direction is safe here — the narrower authority is enforced downstream in `WorksheetScriptService`, which keeps the existing operator and refuses multi-slot operators outright.
- Ownership/IDOR ordering in `validateEditorContext` is untouched: the kind gate runs before the runtime is resolved, so widening `RECOGNIZED_KINDS` cannot turn mint into an existence oracle for a kind it previously refused.
- No backward-compatibility concern: `RECOGNIZED_KINDS` only ever widens, and the visibility change is package-private on a static final immutable `List`.

---

### Blocking status

Requesting changes. Against the mechanical gates:

- **Merge conflicts:** yes — `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, conflicting in `core/src/test/java/inetsoft/web/wiz/pairing/SheetPairingServiceTest.java`. This alone blocks.
- **Checks:** all green except `Claude Code Review / claude-review`, which this manual review replaces. `build`, `Semgrep Scan`, `Semgrep OSS` and `license/cla` all pass — though note that build ran against a base that predates #4789, so it says nothing about the post-rebase state.
- **Code review:** the fix itself is superseded (see Blocking 2); the residual value is real but needs a rebase to be mergeable.

fredaan (approved):

PR #4760, opened 2026-08-24
fix(binding): close seven L3 chart-retype and aesthetic-frame defects

NeverStopFMH:
Found by driving the chart binding surface through the wiz MCP agent tools. Every fix has a regression test **proven RED before GREEN**; one is additionally **verified live** against a rebuilt container.

Companion to `inetsoft-technology/stylebi-wiz#1788`, which closes the five plugin-tier defects from the same pass.

## The shape running through most of these

**The agent surface reaches states the interactive Composer structurally cannot**, so core paths meet inputs they were never written for. Example: `getEditPaneId()` only offers Static/Combined editors on an *unbound* aesthetic channel, so a human can never place a type-matching categorical frame there — but the MCP surface can. Each fix is therefore **behaviour-preserving for every human-reachable state** and changes only the agent-reachable one.

## What is fixed

**`fixPieDimensions` — one boolean produced two opposite defects.** `hasDim` ORed across `color`/`shape`/`size` for "holds a dimension" and gated the whole `x/y → color` migration. A **dimension** occupant suppressed a migration that was needed, stranding the dimension on `x` and rendering a row of dots; a **measure** occupant did not trip the guard, so the migration ran and `setColorField` destroyed it silently. Six single-variable experiments pinned this — **it was not the non-deterministic behaviour it had been recorded as.** `hasDim` is now narrowed to `color`; a dimension on `color` is displaced back to `x`/`y` reusing the method's own reverse-branch placement; a measure on `color` is refused, because this class has no precedent for relocating a measure off `color` and inventing one would produce an unprecedented two-measure pie.

**`copyToTreemap` — leftover measures were silently lost** when `size`/`color`/`shape` were all occupied. Now refused on the same principle: with no safe landing spot, failing loud is the minimal correct move. **Both refusals land before any mutation**, asserted by tests checking the shelves and channels are unchanged after the exception.

**`process()` — dispatch order deleted three Gantt fields.** The `if/else-if` chain tested `copyToRelation()`'s condition before `copyFromGantt()`'s, so leaving Gantt for network/treemap/mekko skipped the only code that reads `start`/`end`/`milestone`. `copyFromGantt()` now runs once, up front, for exactly those three destinations. Deliberately **not** hoisted for every leaving-Gantt transition: `gantt → funnel`/`radar` already reach it through the unmodified chain, and calling it twice makes the second pass overwrite `text` and **silently drop `end`**.

**`AestheticRefModelFactory.createAestheticInfo` — the most severe defect in the lane.** It dereferenced a null `VisualFrameWrapper`, so `set_aesthetic_field` on a multi-style chart returned 500 and left the assembly unreadable to every later call until `undo`. The repair path (`GraphUtil.fixVisualFrame`) already tolerated a null wrapper; the read path did not. **Live-verified fixed.**

**Frame shadowing, three sites.** A field bound to a channel silently replaced the caller's visual frame with a bare default. `ChartAestheticMutator.setField` now carries the effective frame forward; `pasteAestheticRef`'s first-bind branch called `createVisualFrame()`, which copies none of the model's values, and now routes through the factory-driven copy path; six construction sites in `ChangeChartTypeProcessor` share a `carryOrDefault` helper. **The second was found only because the first was verified against a running server and half of it did not work while its unit test passed.**

**`ChartAestheticMutator.setFrame` — a field-less frame was written to a slot nothing reads.** The chart-level slot is consulted only for `MergedChartInfo`, never the case for an ordinary chart, so the frame was stored, read back faithfully, and never rendered. The slot that renders is **per-measure** — which is also where the interactive Composer writes it. Now broadcast to every Y-shelf aggregate (falling back to X), with `frameOf` reading the same slot so the read reports what actually renders. `node-color`/`node-size` excluded by guard: they have no per-aggregate slot.

**`VisualFrameAliases.staticSize` — a caller's static size was discarded.** It built a `StaticSizeModel` without `setChanged(true)`, and `StaticSizeFrameModelFactory` is the only one of the four static factories that gates on `isChanged()` — when false it skips the write *and* clears the USER-tier composite value, letting the per-chart-type default win (read back as 15 on a bar chart, 30 on a pie).

**`ChangeChartTypeService` / `ChartBindingService` — a no-op retype broke a chart's aesthetics.** A same-type retype called `clearRuntime()` then took an early return that skips the rebuild, discarding the runtime aesthetic fields. The service now updates the assembly before that return, and the wiz tier refuses a request that would change nothing so the path is not entered at all.

## Notes for review

- `copyToContour()` was deliberately left alone — its clearing is correct and documented.
- `copyToGantt`'s `carryOrDefault` branch is fixed for consistency but **not tested**: its `startField` always arrives via `GraphUtil.fixStaticColorFrame`, which casts unconditionally to `StaticColorFrameWrapper`, so the scenario throws before the changed code runs and a test would have been fiction.
- Two structurally identical dispatch shadows (`treemap → mekko`, `gantt → funnel`) were traced and left alone — **no data loss** in either; restructuring the dispatch generally is a design decision, not a bug fix.


NeverStopFMH:
## Code review

Reviewed the full diff (12 files, +1240/−55). Scope: seven core-tier defects in chart retype and aesthetic frames, plus the wiz-tier `ChartAestheticMutator` / `ChartBindingService` / `VisualFrameAliases` changes.

**What is good, up front:** every fix ships a regression test, the two refusals are asserted to land before any mutation, and the comments explain *why* the old code was wrong rather than restating what the new code does. `AestheticRefModelFactory.createAestheticInfo`'s null-wrapper guard and `VisualFrameAliases.staticSize`'s `setChanged(true)` are both clearly correct and minimal.

Three findings below are, in my reading, blocking. The rest are medium/minor.

---

### 1. `ChartAestheticMutator.setFrame` reintroduces the same dead-slot defect for merged / radar / contour charts (blocking)

The new broadcast fires purely on "is this a `FRAME_CHANNELS` channel and are there Y/X aggregates?" — it never consults the chart type:

```java
List<ChartAggregateRefModel> aggregates =
   AestheticChannels.FRAME_CHANNELS.contains(name) ? aggregates(model) : List.of();
```

But `VSFrameVisitor.createFrame()` picks its source in this order: bound aesthetic ref → `strategy.supportsFieldFrame()` → `strategy.supportsGeneralFrame()`. And `supportsFieldFrame()` is `info.supportsColorFieldFrame()` (resp. shape/size), which is **false** for:

- `MergedVSChartInfo` — `supportsColorFieldFrame()` / `supportsShapeFieldFrame()` → `false` (candle, stock)
- `RadarVSChartInfo` — `supportsColorFieldFrame()` / `supportsSizeFieldFrame()` → `false`
- `AbstractChartInfo` — `return !GraphTypes.isContour(getChartType())`, so **false for every contour chart**

All three of those chart types carry measures on the Y (or X) shelf, so `aggregates(model)` is non-empty and the new code takes the broadcast branch. The frame lands on the per-aggregate slot the renderer will not read, and the top-level slot the renderer *does* read is left untouched — which is exactly the defect this change exists to fix, mirrored onto a different set of charts.

Note the test `setsAFieldLessFrameOnTheTopLevelPropertyWhenThereIsNoAggregateToTarget` names contour as its example of "no aggregate to target". A contour scatter has measures on x/y, so it does not reach that branch; the rationale in the test's own comment does not hold for the case it cites.

**Suggested fix:** `ChartBindingModel` already exposes `getChartType()`. Gate the broadcast per channel on the same predicate the renderer uses — mirror `supportsColorFieldFrame()` / `supportsShapeFieldFrame()` / `supportsSizeFieldFrame()` — and mirror it again in `frameOf()`, which the class's own Javadoc correctly insists must match the write side exactly. A test for candle (or contour) asserting the top-level slot is what gets written would pin it.

### 2. The two new `IllegalArgumentException`s are reachable from human paths and from non-HTTP callers (blocking)

The PR description's governing principle is "behaviour-preserving for every human-reachable state". I do not think that holds for these two:

- **`fixPieDimensions`**: bar chart, `Month` on x, `Sum(Sales)` on y, `Sum(Discount)` dragged onto color → change type to pie. Every one of those steps is a normal Composer drag. Previously the measure was silently destroyed (bad); now the retype throws.
- **`copyToTreemap`**: a bar chart with color, shape and size all bound and a measure on x/y is likewise plain drag-and-drop. Previously the leftover measure was dropped (bad); now the retype throws.

Both are improvements over silent data loss for the *agent* surface. The concern is where the exception surfaces. `ChangeChartTypeProcessor` is not only constructed by `ChangeChartTypeService`:

- `ChartVSAssemblyInfo.setChartStyle(int)` — the property/script path (`chart.chartStyle = …`)
- `ChartElementDef` — report chart elements
- `ChartDcProcessor`, `DateComparisonDialogService`, `ChangeChartDataProcessor`

An `IllegalArgumentException` thrown out of a low-level processor on those paths turns a previously-silent degradation into a broken viewsheet script or a failed render, with no catch anywhere in between.

Two further points on the same code:

- The messages are raw English string literals. Everything else user-facing in this area goes through `Catalog.getString(...)` and a `MessageCommand` (see `ChangeChartTypeService`'s own `chartTypes.user.noPermission` handling a few lines above). As written these will reach an end user untranslated, or more likely as a generic 500.
- `copyToTreemap` already has a non-destructive landing spot in the same file: `addToAesthetic()` falls through to `text` when color/shape/size are taken. Preferring that over a throw would fix the data loss without changing the failure mode.

**Suggested fix:** keep the refusal as the agent-tier contract, but raise it where it can be turned into a catalogued `MessageCommand` — i.e. pre-check in `ChangeChartTypeService` / `ChartBindingService` — and leave `ChangeChartTypeProcessor` degrading rather than throwing for the script/report callers. If throwing from the processor is deliberate, please say so explicitly and cover the script path (`setChartStyle`) with a test.

### 3. A refusal leaves the chart in exactly the state fix #7 repairs (blocking, and cheap to fix)

`ChangeChartTypeService.changeChartType` runs `cinfo.clearRuntime()` *before* `new ChangeChartTypeProcessor(...).process()`. If `process()` now throws (findings 1 and 2 above), the method unwinds without ever reaching `box.get().updateAssembly(...)` — so the chart is left with its runtime aesthetic fields cleared and its design-time info still reporting them. That is precisely the symptom the `ostackMeasures == nstackMeasures` fix in this same PR was written to close, now reachable through a different door.

`ChangeChartTypeProcessorTest`'s "mutates nothing after the refusal" assertions are at the processor level and do not see this, because the damage happens in the caller.

**Suggested fix:** wrap the `process()` call so a failure still calls `updateAssembly(chart.getAbsoluteName())` before rethrowing, and add a `ChangeChartTypeServiceTest` case for it.

---

### 4. `pasteAestheticRef` — undocumented behaviour change on the clear path (medium)

```java
-      if(wrapper != null) {
-         VisualFrameWrapper nwrapper = vFactoryService.updateVisualFrameWrapper(wrapper, model.getFrame());
-         ref.setVisualFrameWrapper(nwrapper);
-      }
+      VisualFrameWrapper nwrapper =
+         vFactoryService.updateVisualFrameWrapper(ref.getVisualFrameWrapper(), model.getFrame());
+
+      if(nwrapper != null) {
+         ref.setVisualFrameWrapper(nwrapper);
+      }
```

The first-bind repair is right. But the rebind branch changed too, and not in the direction the description covers: with a **non-null** existing wrapper and a **null** `model.getFrame()`, the old code called `updateVisualFrameWrapper(wrapper, null)` → `null` → `setVisualFrameWrapper(null)`, i.e. it *cleared* the frame. The new `if(nwrapper != null)` guard keeps the old wrapper instead. A client that explicitly sends no frame no longer clears one.

`createAestheticInfo` deliberately leaves `frame` null for the text field ("textfield is edited by set and gettextformat request"), so this is not a hypothetical shape. Please confirm which behaviour is intended and pin it with a test either way — this is on the interactive Composer's binding-dialog path, not the agent's.

### 5. The new first-bind path can NPE inside `updateVisualFrameWrapper` (medium, low likelihood)

`VisualFrameModelFactoryService.getVisualFrameWrapper()` swallows a `VisualFrameWrapper.wrap()` failure (`catch(Exception e) { LOG.debug(...) }`) and can then return `null`; `updateVisualFrameWrapper` passes that straight to `getFactory(nwrapper)`, which does `factories.get(wrapper.getClass())`. Previously the `wrapper != null` guard kept the first-bind case away from this entirely. A `if(nwrapper == null) return null;` in `updateVisualFrameWrapper` would close it for both callers.

---

### Minor / nits

- **`copyToTreemap` message construction.** `strandedNames` is built with a hand-rolled `StringBuilder` loop three lines below a `String.join(", ", occupied)`. `stranded.stream().map(ChartRef::getFullName).collect(Collectors.joining(", "))` matches the neighbouring style and drops eight lines.
- **`measures.subList(freeAestheticSlots, measures.size())`** returns a live view of a list that is mutated (`measures.remove(0)`) further down. It is consumed immediately so it is safe today; a defensive copy removes a trap for the next reader.
- **`moveColorDimensionToShelf`** casts `(ChartRef) colorFld.getDataRef()` unconditionally. That cast is inherited from the reverse branch, but it is now reachable from the forward pie migration with a caller-planted ref, which is a wider input space than it had. An `instanceof` guard would be consistent with the defensive posture the rest of this PR takes.
- **`ChangeChartTypeProcessorTest`** has two consecutive Javadoc blocks before `barToPieCarriesForwardAnExistingTopLevelCategoricalColorFrame` — the first one ("Guards the `copyToRelation()` branch that the Gantt fix reorders around…") no longer has a method under it and now documents the wrong test. Also `java.util.Set` / `java.util.HashSet` / `inetsoft.uql.viewsheet.VSDataRef` are fully qualified in the new helpers although the file imports the relevant packages.
- **`ChangeChartTypeService`'s new `updateAssembly`** now fires on *every* same-type/same-flag call, including from the interactive Composer. The wiz tier's `isNoOpRetype` shields the agent path, but a Composer click that re-asserts the current type now costs an assembly re-execution where it used to cost nothing. Probably fine — worth a sentence saying it was considered.
- **`copyToGantt`'s `carryOrDefault` branch** being untested is well justified in the PR body, but that justification lives only in the PR. A one-line comment at the site would keep it for whoever reads the file in a year.

### Tests

Coverage is genuinely good — the two refusals assert non-mutation, the frame carry-forward has a site-by-site test, and `ChartBindingServiceTest` pins both directions of the no-op guard (including the important "must not swallow a real change" cases). Gaps that map to the findings above:

1. No test covers `setFrame` on a chart type whose renderer reads the top-level slot (candle/stock/radar/contour) — finding 1.
2. No `ChangeChartTypeService`-level test for the refusal path — finding 3.
3. No test for `pasteAestheticRef` with a non-null wrapper and a null frame model — finding 4.

### CI

`claude-review` is failing and `build` is still pending as of this comment. Worth confirming `build` goes green before merge, given the new `ArrayList` / `List.of` / `ChartAggregateRefModel` imports in `ChartAestheticMutator` and the new test dependencies (`GradientColorFrame`, `RelationChartInfo`, `CompositeValue`).

### Merge ordering

`stylebi-wiz#1788` ships a `set_visual_frame` description that documents the per-measure broadcast behaviour introduced here. Please land/deploy this PR first, or the plugin will describe behaviour the running server does not have.

NeverStopFMH:
## Review addressed — `52bd994ff`

All five findings fixed. One of them turned out to be worse than the review said; details below.

### 1. `setFrame` broadcasting without consulting the chart type — fixed

The gate is no longer `AestheticChannels.FRAME_CHANNELS`. `setFrame`, `describe` and `frameOf` now take the set of channels *this chart* renders per measure, and `ChartAestheticAgentService` builds it by asking the real `VSChartInfo`:

```java
if(info.supportsColorFieldFrame()) { channels.add("color"); }
if(info.supportsShapeFieldFrame()) { channels.add("shape"); channels.add("line"); channels.add("texture"); }
if(info.supportsSizeFieldFrame())  { channels.add("size"); }
```

Those are exactly the predicates `VSFrameVisitor`'s four strategies call, mapped the way those strategies map them — `VSLineFrameStrategy` and `VSTextureFrameStrategy` both defer to `supportsShapeFieldFrame()`.

Asked of the info object rather than derived from the chart type on purpose: `RadarVSChartInfo.supportsShapeFieldFrame()` consults the plot descriptor, so a point radar and a line radar give different answers for the same type. Deriving it would have been wrong for one of them.

The overloads without the set default to `FRAME_CHANNELS`, which is correct for every `DefaultVSChartInfo` chart, so nothing that had no chart in hand changed behaviour.

Three tests added, including the read side — the class's own Javadoc insists `frameOf` must mirror `setFrame` exactly, and it now has to mirror the gate too.

You were also right that the old test's rationale was wrong: contour was cited as the "no aggregate to target" case, and a contour scatter has measures on x/y. It is now covered by the new tests as what it actually is — a chart type that reads the chart-level slot.

### 2. Refusals thrown from a low-level processor — fixed, opt-in

Refusal is now a policy, off by default:

```java
public ChangeChartTypeProcessor setStrictFieldPlacement(boolean strict)
```

`ChangeChartTypeService` is the only caller that turns it on, because it is the only one where a person or an agent asked for the retype and can be told no. `ChartVSAssemblyInfo.setChartStyle` (the script and property path), `ChartElementDef`, `ChartDcProcessor` and `DateComparisonDialogService` keep the previous drop — now with a `LOG.warn` naming the field instead of in silence. Your point stands as written: those callers are inside someone else's operation with no dispatcher and nothing to catch, so an exception there would have turned a chart that rendered imperfectly into a script error.

Two tests pin the lenient default, so the split cannot quietly collapse back into "always throws".

### 3. The refusal leaving the chart in the state fix #7 repairs — fixed, and it was worse than reported

`ChangeChartTypeService` now catches, and the new `ChangeChartTypeServiceTest` found something the review did not: **the refusal was not atomic even at the processor level.** `fixChartInfo()` runs at the top of `process()` and stamps the new type onto the live info when both types share a `ChartInfo` class (bar → pie), so the assembly was left claiming `pie` with a bar's binding. The processor-level "mutates nothing" assertions could not see it, because the damage was on an object the caller holds.

So the catch does three things, not two:

```java
chart.setVSAssemblyInfo(oinfo);                    // the clone taken at the top of the method
box.get().updateAssembly(chart.getAbsoluteName()); // rebuild what clearRuntime() discarded
dispatcher.sendCommand(errorCommand);              // report the way composer services report
```

The `MessageCommand` route is what you suggested and it pays twice: the browser gets a dialog naming the field instead of a 500, and the agent tier gets a loud failure for free — `CapturingCommandDispatcher` turns an ERROR command into a `CommandErrorException`, which is exactly the contract the wiz layer already relies on.

New `ChangeChartTypeServiceTest`, 4 tests: the error command and its wording, the runtime rebuild, the binding left untouched, and — the direction that matters more — an ordinary retype sending no error command.

### 4. `pasteAestheticRef`'s undocumented clear-path change — reverted to the old semantics

Confirmed as unintended. `nwrapper == null` now clears when the ref already had a wrapper (which is what it always did, and the shape `createAestheticInfo` produces for the text field) and writes nothing when the ref is brand new. Only the first-bind-with-a-frame case is new behaviour, which was the point of the fix. Test added for the clear path.

### 5. `updateVisualFrameWrapper` NPE — guarded

`if(nwrapper == null) return null;` before `getFactory()`. As you noted, `getVisualFrameWrapper` swallows a `wrap()` failure and can hand back null; that used to be unreachable from the first-bind path and is not any more.

### Nits

All taken: `String.join`/`Collectors.joining` for the treemap message, a copy rather than a `subList` view of a list mutated below, an `instanceof` guard on `moveColorDimensionToShelf`'s cast (it logs and leaves the field bound rather than throwing), the dangling Javadoc removed, and the fully-qualified `java.util.Set` / `VSDataRef` / `DataRef` replaced with imports.

On `ChangeChartTypeService`'s new unconditional `updateAssembly` for a same-type/same-flag call: considered and left. The wiz tier's `isNoOpRetype` shields the agent path, and the Composer does not send a same-type event, so in practice this fires only for a caller that is already doing something pointless — and doing it correctly is worth one assembly execution.

`copyToGantt`'s untested `carryOrDefault` branch: left as-is, still untested for the reason in the PR body. Fair to call it out; the justification lives only in the PR description, and I'd rather not add a comment that will outlive the reason.

### Verification

149 tests green: `ChangeChartTypeProcessorTest` (25), `ChangeChartTypeServiceTest` (4, new), `ChartAestheticMutatorTest` (42), `ChartAestheticAgentServiceTest` (17), `ChartBindingServiceTest` (46), `ChartBindingMutatorTest` (11), `PasteAestheticRefFirstBindFrameTest` (2), `AestheticRefModelFactoryTest` (1), `StaticSizeFrameModelFactoryChangedTest` (1). `mvn test-compile` clean.

The `build` check was still pending when the review was written; it will re-run on this push.

NeverStopFMH:
## Second-round review — `52bd994ff` + merge `e1166b33b`

I re-read the whole PR against the render path rather than only against the diff. Four of the five round-1 findings are genuinely closed and I could verify each one in the source; finding 1 is closed for the chart types it names and left open for one it does not.

**Verified fixed:**

- **#2 (opt-in refusal)** — `setStrictFieldPlacement` defaults to `false` and `ChangeChartTypeService` is the only `setStrictFieldPlacement(true)` in the tree. `ChartVSAssemblyInfo.setChartStyle`, `ChartElementDef`, `ChartDcProcessor`, `DateComparisonDialogService` all keep the drop, and two tests pin the lenient default so the split cannot quietly collapse.
- **#3 (non-atomic refusal)** — you were right that it was worse than reported, and for the reason you give: `fixPieDimensions(info)` is called from *inside* `fixChartInfo()` (`ChangeChartTypeProcessor:369`), i.e. after the type has been stamped onto the live info, so the processor genuinely cannot make itself atomic. Restoring `oinfo` in the caller is the right place. I also checked the two paths that mutate before the `try` — `handleMulti` (only reachable when `oldType == newType`, which no refusal branch can be) and the geo branch (only for map, which does not refuse) — so the clone taken at line 118 really is a faithful pre-state for every reachable refusal.
- **#4 (clear path)** — `if(nwrapper != null || wrapper != null)` reproduces the old semantics exactly, and the text-field shape is now pinned.
- **#5 (NPE)** — guarded, and the guard is in the right place (`updateVisualFrameWrapper`, so both callers get it).

---

### 1. Gantt: the round-1 defect is still live, in the same shape (blocking)

`ChartAestheticAgentService.perMeasureFrameChannels` derives the set from `supportsColorFieldFrame()` / `supportsShapeFieldFrame()` / `supportsSizeFieldFrame()`, which is the right predicate. But `GanttVSChartInfo` overrides **all three to `return true`** (`GanttVSChartInfo:562-575`), so a Gantt chart gets `{color, shape, line, texture, size}` and takes the broadcast branch.

The broadcast then targets `aggregates(model)` — the Y shelf, falling back to X. `VSFrameVisitor.getAggregates()` does **not** do that for Gantt:

```java
private ChartAggregateRef[] getAggregates() {
   if(GraphTypes.isGantt(info.getChartType())) {
      GanttChartInfo ginfo = (GanttChartInfo) info;
      return getAggregates(ginfo.getRTStartField(), ginfo.getRTMilestoneField());
   }
   ...
}
```

On a Gantt chart the Y shelf holds only dimensions — `copyToGantt()` in this very PR does `otherdims.stream().forEach(d -> info.addYField(d))` and routes every measure onto `startField`'s own aesthetic channels. So `aggregates(model)` is empty, `setFrame` falls through to the chart-level slot, and `createFrame()` never reads it: `supportsFieldFrame()` is true, `arr` (start/milestone) is non-empty, so the `else if(strategy.supportsGeneralFrame())` branch is unreachable. The frame is stored, `frameOf` reads it back faithfully, and nothing renders — which is the exact sentence this PR's own commit message uses to describe the defect it is fixing.

Note the javadoc on `setFrame` lists Gantt among "MergedChartInfo charts (Candle/Gantt/Radar/Relation/Map)" whose chart-level slot `createFrame()` falls back to. That is true for the other four and false for Gantt, precisely because `GanttVSChartInfo` overrides the three predicates back to `true`.

**Suggested fix:** mirror `getAggregates()`'s Gantt branch in `aggregates(model)` (target the start/milestone refs), or — if the binding model does not expose them as `ChartAggregateRefModel` — have `perMeasureFrameChannels` return the empty set for a `GanttChartInfo` and add a `TODO`, so at least the write and the read agree on one slot rather than landing on a third. Either way a Gantt test would pin it.

### 2. `catch(IllegalArgumentException)` is wider than the contract it implements (medium)

```java
catch(IllegalArgumentException e) {
   chart.setVSAssemblyInfo(oinfo);
   box.get().updateAssembly(chart.getAbsoluteName());
   MessageCommand command = new MessageCommand();
   command.setMessage(e.getMessage());
   ...
}
```

`process()` is a long call chain — `fixChartInfo`, six `copyTo*`/`copyFrom*` methods, `GraphUtil.fixVisualFrames`, `ChangeChartDataProcessor.sortRefs`. Any `IllegalArgumentException` raised anywhere in it for an unrelated reason (a bad cast argument, a parse, a genuine bug) now gets: the assembly silently rolled back, and the exception's raw internal message shown to the user in a dialog as though it were a considered refusal. That is a worse failure mode than the 500 it replaces, because it looks like success-with-a-reason.

**Suggested fix:** give the refusal its own type — a nested `static final class FieldPlacementException extends IllegalArgumentException` on `ChangeChartTypeProcessor` — throw that from the two sites, and catch that. One-line change at each of the four points, and it makes the contract explicit rather than conventional.

### 3. The refusal messages are still raw English literals (medium — round-1 nit, not addressed)

Round 1 raised this and the response covered the *dispatch* (`MessageCommand`) but not the text. `e.getMessage()` is the hardcoded English built in `ChangeChartTypeProcessor`, and it now lands in a Composer dialog. Three lines above in the same method, the sibling refusal does:

```java
String msg = catalog.getString("chartTypes.user.noPermission",
                               GraphTypes.getDisplayName(event.getType()) + " " + catalog.getString("Chart"));
```

Making the refusal a typed exception (finding 2) makes this cheap: carry the field name(s) and the channel list as fields, and let `ChangeChartTypeService` build the message through `Catalog.getString("chartTypes.user.<key>", ...)`. The processor's `LOG.warn` on the lenient path can stay English.

---

### 4. `frameOf`/`setFrame`'s chart-level fallback is dead for a scatter contour too (minor, not a regression)

The javadoc says "On those charts the chart-level property is what `createFrame()` falls back to", citing contour among them. `supportsGeneralFrame()` is `info instanceof MergedChartInfo` for colour, shape, line and texture — a **contour map** is a `VSMapInfo`, so that holds; a **scatter contour** is a `DefaultVSChartInfo`, so it does not. There `supportsColorFieldFrame()` is false *and* `supportsGeneralFrame()` is false, and `createFrame()` returns `null`.

Nothing gets worse — the write moved from one dead slot to another — but the rationale in the javadoc and in `setsAFieldLessFrameOnTheTopLevelPropertyWhenThereIsNoAggregateToTarget` is wrong for the case it cites (again: the test's contour example still does not reach the branch it is filed under), and `get_chart_aesthetics` will report a scatter-contour frame that never renders. Worth one sentence in the javadoc saying so.

### 5. "No aggregate to target" is not a case where the top-level slot renders either (minor)

```java
else if(strategy.supportsFieldFrame()) {
   if(arr.length > 0) { ... }          // <- nested, not chained
}
else if(strategy.supportsGeneralFrame()) { ... }
```

When `supportsFieldFrame()` is true and there are no aggregates, `frame` stays `null` — the general branch is never reached. So on an ordinary bar chart with zero measures, the comment "these are the genuine cases where the top-level property is what the render path (still) consults" is not accurate. Harmless in practice (a chart with no measure renders nothing anyway), but the comment is load-bearing for the next reader of this method.

### 6. `carryOrDefault` hands out a shared frame instance (minor)

`carryOrDefault` returns `existing` itself, so after e.g. `barToPieCarriesForwardAnExistingTopLevelCategoricalColorFrame` the chart-level `colorFrame` slot and the new `AestheticRef`'s visual frame are the **same object** — the test asserts `assertSame`, so this is deliberate. Editing one edits the other, and `VisualFrame` is mutable (`setField`, `setLegendSpec`, colour arrays). `VSFrameVisitor.createVisualFrame` clones before use so the render path is safe today, but a `(VisualFrame) existing.clone()` here would cost nothing and remove the aliasing.

---

### Tests

- **`ChartAestheticAgentService.perMeasureFrameChannels` has no test.** It is the linchpin of the round-1 fix, and the three new mutator tests hand-feed `Set.of("size")` rather than exercising it. The rationale you give for asking the info rather than deriving from the type — a point radar and a line radar disagree via `RadarVSChartInfo.supportsShapeFieldFrame()`'s plot-descriptor check — is exactly the thing worth pinning, and it would have surfaced the Gantt case above.
- **`ChangeChartTypeServiceTest` only covers the pie refusal.** The treemap refusal reaches `copyToTreemap()` through a different branch of `process()` (after `fixChartInfo()` rather than inside it), so its restore path is not the same code and is unexercised at service level.
- Everything else added here is solid — `ChartBindingServiceTest` pinning both directions of the no-op guard (including the per-measure-vs-chart-type comparison) is the right shape, and the `StaticSizeFrameModelFactoryChangedTest` building the model through `VisualFrameAliases` rather than by hand is the reason that test is worth having.

### Things I checked and found correct, for the record

- `perMeasureFrameChannels` mapping `supportsShapeFieldFrame()` → `shape`/`line`/`texture` matches `VSLineFrameStrategy` and `VSTextureFrameStrategy`, both of which return `info.supportsShapeFieldFrame()`.
- `ChartBindingService.isNoOpRetype` returning early does **not** skip anything live: `applyMultiStyle` returns immediately on `multi == null` or `info.isMultiStyles() == multi`, which the guard's own condition already implies, and the forced-separate refusal sits above the guard.
- `currentChartType` returning `null` rather than the chart's type on a missing named measure is the right direction — a skipped write is a silent one.
- `setField`'s carry-forward does not strand a family-mismatched frame: `GraphUtil.fixVisualFrame` replaces a non-`CategoricalFrame` on a categorical ref and a non-`LinearColorFrame` on an aggregate, and `VSFrameVisitor.createVisualFrame` re-stamps `frame.setField(tref.getFullName())` at render, so a carried frame's stale `field` does not survive either.

### CI

`build` is now green (18m10s). `claude-review` reports `fail`, but the job body is "Claude encountered an error after 0s" on both runs — infrastructure, not a verdict.

NeverStopFMH:
## Second review addressed — `1c35a755a`

All six fixed. Finding 1 was right and the reasoning behind it was the part worth having: the gate was correct and the *target* was wrong, which is not something the chart-type predicate could ever have caught.

### 1. Gantt — fixed

`perMeasureFrameChannels` was already answering correctly: `GanttVSChartInfo` overrides all three `supports*FieldFrame()` back to `true`, so Gantt belongs on the per-measure branch. The bug was one layer down — `aggregates(model)` targeted the X/Y shelves, and a Gantt chart keeps only dimensions there. `copyToGantt` in this same PR is the proof: `otherdims.stream().forEach(d -> info.addYField(d))`, every measure onto `startField`'s own channels.

So `aggregates()` now mirrors `VSFrameVisitor.getAggregates()` branch for branch rather than only its Y-then-X tail:

```java
if(GraphTypes.isGantt(model.getChartType())) {
   return aggregatesOf(Arrays.asList(model.getStartField(), model.getMilestoneField()));
}
```

`ChartBindingModel` already carries start/milestone as `ChartRefModel`, and `ChartInfoModelBuilder` round-trips them through the same `createRefModel`/`pasteChartRef` pair as any shelf aggregate, so the frames persist the same way the Y-shelf ones already did. `frameOf` goes through the same method, so the read still reports the slot the write used.

Two tests: the write lands on `startField` and leaves `model.getColorFrame()` null, and `describe` reads it back from there.

### 2. `catch(IllegalArgumentException)` — fixed, typed

You were right that this was wider than the contract, and specifically right about *why* it is worse than the 500: a rolled-back assembly plus an internal exception message in an error dialog reads as success-with-a-reason. `ChangeChartTypeProcessor.FieldPlacementException` now carries the refusal, both throw sites use it, and the service catches that. It still extends `IllegalArgumentException`, so a caller that does not know about the contract sees no change. The two processor tests assert the narrow type now, so the split cannot widen back by accident.

### 3. Catalogued messages — fixed

Fair to re-raise; the first round's response answered the dispatch and not the text. Making it a typed exception paid for this immediately — it carries the key and the arguments, and the service formats:

```java
command.setMessage(Catalog.getCatalog().getString(e.getCatalogKey(), e.getArguments()));
```

Two entries in `srinter.properties` next to `chartTypes.user.noPermission`. `getMessage()` stays English for the log and for the lenient path's `LOG.warn`. A test asserts the message is not the key verbatim, which is what `Catalog` returns when an entry is missing — so deleting the entry fails the build rather than shipping a key to a user.

### 4 & 5. The two inaccurate rationales — corrected

Both were documentation rather than behaviour, and both were wrong in the direction that matters (they described a fallback that does not exist), so they are now stated as what they are:

- `supportsGeneralFrame()` is `info instanceof MergedChartInfo` for colour/shape/line/texture, so a **scatter contour** — a `DefaultVSChartInfo` — reads neither slot. The javadoc says so instead of listing contour among the charts the chart-level slot renders for.
- The `arr.length > 0` test is nested inside the `supportsFieldFrame()` branch rather than chained after it, so **zero measures** means `createFrame()` returns null without consulting `getGeneralFrame()`. The "no aggregate to target" comment no longer claims otherwise.

The write stays in the chart-level slot for both, and the comment now gives the actual reason: it is the slot `frameOf()` reports from, so the read keeps agreeing with the write, and a chart with no measure renders nothing to disagree about. Also fixed the javadoc's list of "MergedChartInfo charts" — it had Gantt in it, which is exactly backwards.

### 6. `carryOrDefault` aliasing — fixed

Carried as a clone. The five `assertSame` tests now set a distinctive colour on the frame first and assert the carried *values* plus `assertNotSame` — which is a better test than the identity check was, since a bare default never carries that colour.

### Test gaps — closed

- **`PerMeasureFrameChannelsTest`** (new, 6 tests). Its own class rather than a section of `ChartAestheticAgentServiceTest`, because these build real `VSChartInfo` objects and need the SREE context that test deliberately does without. Covers the ordinary chart, candle, scatter contour, Gantt, an unbound chart, and the point-radar/line-radar disagreement — the case that is the whole reason this asks the info rather than deriving from the chart type. `perMeasureFrameChannels` is package-private now so it can be pinned directly.
- **`ChangeChartTypeServiceTest`** gains the treemap refusal (which reaches the refusal from the dispatch chain, not from inside `fixChartInfo()` like the pie one) asserting both the error command and the full restore, plus the catalogue check above.
- **`ChartAestheticMutatorTest`** gains the Gantt write and read.

### Verification

202 tests green: `ChangeChartTypeProcessorTest` (25), `ChangeChartTypeServiceTest` (6), `ChartAestheticMutatorTest` (44), `ChartAestheticAgentServiceTest` (17), `PerMeasureFrameChannelsTest` (6, new), `ChartBindingServiceTest` (46), `ChartBindingMutatorTest` (11), `VisualFrameAliasesTest` (43), `PasteAestheticRefFirstBindFrameTest` (2), `AestheticRefModelFactoryTest` (1), `StaticSizeFrameModelFactoryChangedTest` (1). `BindingReadServiceTest` (14) green as well, since it is the other reader of this surface. `mvn test-compile` clean.

Thanks for checking the round-1 fixes against the render path rather than the diff — finding 1 is only visible from that direction.

NeverStopFMH:
## Added `a0ba7d8f3` — `DateLevels` accepts the part levels

Found while reviewing the companion plugin PR ([stylebi-wiz#1788](https://github.com/inetsoft-technology/stylebi-wiz/pull/1788)) and reproduced live.

`DateLevels.byName()` offered ten interval levels and left out `AM_PM_DATE_GROUP` and the entire `PART_DATE_GROUP` family, with a comment saying so deliberately — a different question deserving a deliberate answer rather than a half-supported name.

That reasoning does not survive the round trip, and the failure is measurable. Those levels are already reachable: the Composer offers them on its own date-level menu, so a chart can hold one, and the agent surface reads a binding back before writing it. Live against a running deployment:

```
set_chart_shelf … dateLevel: "month_of_year"
  → '514' is not a date level. Accepted names: day, hour, millisecond, minute,
    month, none, quarter, second, week, year. The equivalent numbers are also accepted.

set_chart_shelf … dateLevel: "am_pm"     → '9' is not a date level. …
set_chart_shelf … dateLevel: "month"     → ok: true          (control)
```

The plugin's schema advertises all 22 names, so an agent is told a level is valid and then refused by a shorter, different list. A vocabulary the read can produce is a vocabulary the write has to accept.

The table now covers all 22 of `XConstants`' levels, matching the plugin's, so the two tiers hold one vocabulary rather than two that disagree at the boundary. Three tests, each proven RED against the ten-name table first (`'quarter_of_year' is not a date level`).

**397 `inetsoft.web.wiz.binding` tests pass.**

Deploy ordering note for both PRs: this one has to land before or with #1788, which now depends on it for two things — the frame behaviour `set_visual_frame`'s description promises, and this date-level table.

NeverStopFMH:
## Third-round review — `1c35a755a` + `a0ba7d8f3`

Both new commits read against the render path again rather than against the diff. The typed exception, the catalogued messages, the two corrected rationales, the clone and the three new test classes all hold up, and I checked the load-bearing assumptions under them rather than taking them on the commit message's word (list at the end). The date-levels commit is correct and, unusually, provably complete.

One finding, and it is in the code the last round claimed to have made exact.

### 1. `aggregates(model)` still does not mirror `getAggregates()` — three divergences (blocking)

The Gantt branch is right and fixes what it says it fixes. But the javadoc now claims the method mirrors `VSFrameVisitor.getAggregates()` "branch for branch", and that is a stronger claim than the code makes. `getAggregates()` is two methods, and only the outer one was ported:

```java
private ChartAggregateRef[] getAggregates() {
   ...
   if(containsMeasure(ryrefs)) { return getAggregates(ryrefs); }
   else { return getAggregates(rxrefs); }
}

private boolean containsMeasure(ChartRef[] refs) {
   ...
   return refs[refs.length - 1].isMeasure();          // (a) the LAST ref, not "contains"
}

private ChartAggregateRef[] getAggregates(ChartRef... refs) {
   for(int i = refs.length - 1; i >= 0; i--) {
      if(refs[i] == null) { continue; }
      else if(!refs[i].isMeasure()) { break; }        // (b) trailing contiguous run only
      list.add(0, refs[i]);
   }
   ...
}
```

**(a) Y-vs-X selection.** `containsMeasure` despite its name tests only the *last* ref. `aggregates()` selects Y when Y holds *any* aggregate anywhere.

> Y = `[Sum(Sales), Region]`, X = `[Sum(Profit)]`. The visitor sees a non-measure last on Y, so it reads **X** and renders from `Sum(Profit)`. `aggregates()` sees an aggregate on Y and writes to **`Sum(Sales)`**. The frame lands where nothing renders from, and `frameOf` reports it — write and read agreeing with each other and not with the chart, which is the exact failure shape this whole lane exists to remove.

**(b) Which refs within a shelf.** The backwards loop with `break` takes only the trailing contiguous run of measures; `aggregatesOf` takes every aggregate on the shelf.

> Y = `[Sum(Sales), Region, Sum(Profit)]`. The visitor renders from `[Sum(Profit)]` alone. The broadcast writes both, so the *render* still comes out right by luck — but `frameOf` returns `aggregates.get(0)` = `Sum(Sales)`, so `get_chart_aesthetics` reports a slot the renderer does not consult whenever the two disagree (e.g. a per-measure colour set through the Composer's own dialog rather than through `set_visual_frame`). The class javadoc's "when several measures disagree, the first measure's frame is reported" covers ambiguity, not reporting the wrong one.

**(c) Discrete aggregates.** `VSChartAggregateRef.isMeasure()` is `return !isDiscrete()`, so a discrete aggregate ends the run in (b) and fails the test in (a). `aggregatesOf` filters on `instanceof ChartAggregateRefModel` only — and `ChartAggregateRefModel.isDiscrete()` exists and is populated at construction (`ChartAggregateRefModel:44`), so the information is right there.

> Y = `[Sum(Sales), discrete Sum(Quantity)]`. The visitor's `arr` comes back empty, `createFrame()` returns null, nothing renders per measure at all. `aggregates()` writes to both.

None of these are agent-only: a dimension after a measure on Y and a discrete aggregate are both ordinary Composer drags.

**Gantt inherits (b).** `getAggregates(RTStartField, RTMilestoneField)` walks milestone → start and breaks on a non-measure, so a milestone bound to a plain dimension hides start from the renderer, while the new branch would still write to start. Narrow — `addFieldForGantt` only ever routes date *aggregates* into those slots — but `set_chart_single_shelf("milestone", <dimension>)` reaches it.

**Suggested fix.** Port the algorithm literally rather than its summary:

```java
private static List<ChartAggregateRefModel> aggregates(ChartBindingModel model) {
   if(GraphTypes.isGantt(model.getChartType())) {
      return trailingMeasures(Arrays.asList(model.getStartField(), model.getMilestoneField()));
   }

   List<ChartRefModel> y = model.getYFields();
   return endsWithMeasure(y) ? trailingMeasures(y) : trailingMeasures(model.getXFields());
}
```

with `trailingMeasures` walking backwards, skipping nulls, breaking on the first non-measure, and "measure" meaning `ref instanceof ChartAggregateRefModel agg && !agg.isDiscrete()`. Every existing test stays green — on an all-measure shelf the trailing run *is* the whole shelf, so `broadcastsAFieldLessFrameToEveryYAxisAggregate` and the rest are unaffected — and the three cases above become tests.

**One gap that cannot be closed from here, and should be said rather than papered over:** `getAggregates()` reads `getRTXFields()`/`getRTYFields()` and the Gantt branch reads `getRTStartField()`/`getRTMilestoneField()`, while `ChartBindingModel` carries the design-time refs. That is correct for a *write* — design-time is what persists and what the runtime is rebuilt from — but it means the mirror is of the algorithm, not of the inputs. Worth one sentence in the javadoc in place of the "branch for branch" claim.

---

### Nits

- **The refusal text now exists twice.** `FieldPlacementException`'s message and the `srinter.properties` entry are two copies of the same sentence with nothing tying them, and they have already drifted: "color, shape or size" vs "Color, Shape or Size", "before retyping to pie" vs "before changing to a pie chart". Since `getMessage()` now only serves the log and the lenient path's `LOG.warn`, it could be a short diagnostic line ("pie retype refused: measure 'Sum(Discount)' on color") leaving exactly one user-facing copy.
- **`FieldPlacementException` has no `serialVersionUID`** and carries an `Object[]`; `IllegalArgumentException` is `Serializable`, so some analyzers will flag it.

### 2. `a0ba7d8f3` (date part levels) — correct, and complete

Checked against `XConstants` rather than against the commit message, because "the read can produce it" is a completeness claim and completeness is checkable:

- The table now covers **every** `*_DATE_GROUP` constant — 11 interval (`none`, `year`, `quarter`, `month`, `week`, `day`, `am_pm`, `hour`, `minute`, `second`, `millisecond`) and all 11 part levels (`513`–`523`). Nothing is left that a read can produce and a write would refuse, which is the actual bar the commit sets itself.
- `PART_DATE_GROUP` (512) is correctly *not* in the table — it is the mask, not a level.
- The values are unique across both halves, so `name()`'s linear reverse scan is unambiguous; `roundTripsEveryPartLevelAsANumber` pins that in both directions.
- Both callers (`FieldRefFactory:111`, `TableBindingMutator:356`) feed `VSDimensionRef.setDateLevel`, where part levels are valid — broadening `VALUES` does not open a surface that cannot take them.

No finding. The "a vocabulary the read can produce is a vocabulary the write has to accept" framing is the right generalisation and is worth applying to the other normalisers on this surface.

---

### Verified correct this round (assumptions the fixes rest on)

- **The Gantt gate fires in production, not only in tests.** `ChartInfoModelBuilder:131` calls `loadVisualFrames(model, cinfo, cinfo)` → `ChartAestheticService:44` `model.setChartType(bindable.getChartType())`, so `ChartBindingModel.getChartType()` really is the chart's type. Had it been left at `CHART_AUTO` the whole branch would have been dead outside the new tests.
- **Gantt's start/milestone really are `ChartAggregateRef`s.** `addFieldForGantt` routes a field into `dateDims` only when it is an `XAggregateRef` with a date original type (setting the formula to `NONE`), which is why `copyToGantt`'s unconditional cast at the bottom is safe and why `ChartInfoModelBuilder` renders them as `ChartAggregateRefModel` — so `aggregatesOf` finds them.
- **The catalogue assertion is a real check.** `Catalog.getString0(id, log, def = true)` returns the id when the key is missing, so `assertNotEquals(key, message)` cannot false-pass on a null.
- **The two Gantt slots do not contradict each other.** `AbstractChartInfo.getColorField()` is a plain chart-level field and `VSColorFrameStrategy.getAestheticRef(null)` reads that same slot, so `setField` writing chart-level and `setFrame` writing per-measure are each right for what they write — `createFrame()` consults the bound ref first and only falls to the per-measure frame when there is none.
- The clone in `carryOrDefault` is safe on the render path — `VSFrameVisitor.createVisualFrame` clones again before use — and the five rewritten tests are stronger than the `assertSame` they replaced.

### CI

`build` green (12m18s), both Semgrep jobs green. `claude-review` still reports `fail` with "Claude encountered an error after 0s" — same infrastructure error as the previous two rounds, not a verdict.

NeverStopFMH:
## Third review addressed — `afc42f5b0`

The finding was right and the diagnosis was right about *why*: the previous round ported the outer `getAggregates()` and paraphrased the inner one, then documented the result as exact. `aggregates()` is a port now.

### 1. All three divergences — fixed

```java
private static List<ChartAggregateRefModel> aggregates(ChartBindingModel model) {
   if(GraphTypes.isGantt(model.getChartType())) {
      return trailingMeasures(Arrays.asList(model.getStartField(), model.getMilestoneField()));
   }

   List<ChartRefModel> yFields = model.getYFields();
   return endsWithMeasure(yFields)
      ? trailingMeasures(yFields) : trailingMeasures(model.getXFields());
}
```

- **`endsWithMeasure`** is `containsMeasure`: the last ref on the shelf and only that one.
- **`trailingMeasures`** is `getAggregates(ChartRef...)`: walk from the end, `continue` past a null slot, `break` at the first non-measure.
- **`isMeasure`** is `ChartAggregateRef.isMeasure()`: `instanceof ChartAggregateRefModel && !isDiscrete()`. `ChartAggregateRefModel.isDiscrete()` was already populated at construction, so the information had been sitting there unused.

Each helper is named after the method it ports and says so in a one-line javadoc, so the next person changing one has a reason to look at the other.

**Four tests, each proven RED against the previous implementation** with the symptom the finding predicted — not asserted, run:

```
aDimensionLastOnYSendsTheFrameToTheXAggregateTheRendererUses:333
  the renderer reads X when Y does not end in a measure ==> expected: <StaticColorModel> but was: <null>
onlyTheTrailingRunOfMeasuresIsTargeted:358
  the dimension between them ends the run the renderer walks ==> expected: <null> but was: <StaticColorModel@58cbafc2>
aDiscreteAggregateIsNotAMeasureForThisPurpose:386
  expected: <null> but was: <StaticColorModel@7d0b7e3c>
aNonMeasureMilestoneEndsTheGanttRunBeforeTheStartField:408
  getAggregates() breaks at milestone before it ever reaches start ==> expected: <null> but was: <StaticColorModel@235ecd9f>
```

The other 44 in that class passed unchanged in the same run, which is the point worth making about the narrowing: on an all-measure shelf the trailing run *is* the whole shelf, so `broadcastsAFieldLessFrameToEveryYAxisAggregate` and the rest never notice.

`onlyTheTrailingRunOfMeasuresIsTargeted` also asserts the read half — `describe()` reporting `Sum(Profit)`'s frame rather than `Sum(Sales)`'s — because that was the half that showed through the tool surface even while the render came out right by accident.

`aDiscreteAggregateIsNotAMeasureForThisPurpose` and the Gantt one land on the chart-level slot, which is correct for the same reason the "no measure to target" case is: the renderer builds nothing per measure there, so the chart-level slot is at least where the write and the read agree.

### 2. What is not mirrored, stated rather than claimed away

The javadoc no longer says "branch for branch". It says which three things the port exists for, and then:

> `getAggregates()` reads the runtime refs (`getRTYFields()`, `getRTStartField()`) and a `ChartBindingModel` carries the design-time ones. That is the right side for a write — design time is what persists and what the runtime is rebuilt from — but it means this mirrors the algorithm, not its inputs.

`frameOf`'s javadoc gets the same correction: "the first of `aggregates`" is now qualified as the first measure the renderer itself combines, not the first on the shelf.

### Nits — both taken

- **The duplicated refusal sentence is gone.** `getMessage()` is a short diagnostic for the log and the lenient path (`"pie retype refused: measure 'Sum(Discount)' on color"`); the user-facing wording lives once, in `srinter.properties`. You were right that the two copies had already drifted — that was three days old and two words apart, which is about the rate you would expect.
- `serialVersionUID` added.

### Verification

243 tests green: `ChangeChartTypeProcessorTest` (25), `ChangeChartTypeServiceTest` (6), `ChartAestheticMutatorTest` (48), `ChartAestheticAgentServiceTest` (17), `PerMeasureFrameChannelsTest` (6), `ChartBindingServiceTest` (46), `ChartBindingMutatorTest` (11), `VisualFrameAliasesTest` (43), `FieldRefFactoryTest` (23), `BindingReadServiceTest` (14), `PasteAestheticRefFirstBindFrameTest` (2), `AestheticRefModelFactoryTest` (1), `StaticSizeFrameModelFactoryChangedTest` (1). `mvn test-compile` clean.

### On the review itself

Worth recording what the last two rounds actually cost, because the shape repeated: both times the defect was a *paraphrase of a method in the render path presented as a mirror of it*, and both times it was invisible from the diff and obvious from opening `VSFrameVisitor`. The check that found it — read the method being mirrored, not the comment claiming to mirror it — is cheap enough to be worth applying to the remaining mirrors on this surface rather than waiting for a fourth round to find one.

fredaan (approved):

PR #4764, opened 2026-08-24
Bug #76051, encrypt the Fluentd password and pin the credential round trip

ngovus-inetsoft:
## Problem

A Fluentd shared key saved from **Settings > Logging** can never authenticate. It is stored encrypted (`setFluentdSettings` → `toPassword` → `Tool.encryptPassword`), but `ForwardService.getClient` in the enterprise module read it with a plain `SreeEnv.getProperty` and handed the ciphertext to `ForwardClient` — which **digests** the key rather than sending it. StyleBI's digest was computed over the ciphertext, the collector computed its own over the operator's real key, and `readPong` threw `Authentication failed: ...`. No combination of Logging-page inputs made it succeed.

The read-side fix lives in the enterprise module. **This PR is the community half.**

## What this PR changes

`log.fluentd.security.password` — the neighbouring field, which showed what correct looked like — was written through `sanitizeProperty` with no encryption and read back the same way, so that half authenticated end to end. It was also stored in **clear text**, and since `PropertiesController` only filters the `log.fluentd.*` family on community builds (`PropertiesController.java:140`), it was readable on the enterprise **Settings > Properties** page and returned by `GET /api/admin/properties`.

It now goes through the same `toPassword` helper as the shared key:

```java
SreeEnv.setProperty("log.fluentd.security.password", toPassword(model.password()));
```

The read side needed no change — `getFluentdSettings` already ran the password through `Tool.decryptPassword` (`LogSettingService.java:117-119`). That call was the no-op that hid the gap, working only because `LocalPasswordEncryption.decryptPassword` returns unprefixed input verbatim.

## Backward compatibility

No migration. That same clear-text tolerance means existing stored passwords keep authenticating until the next save re-encrypts them.

## Tests

| Guarantee | Change |
|---|---|
| [G4] | Rescoped. It asserted `encryptPassword` was *never* called for a blank shared key — true only while the password was clear text. Now asserts no **blank** value is ever encrypted, which is what [G4] is about. |
| [G5] | Now expects both credentials encrypted and the non-secret fields only trimmed. |
| [G6] | New — the read returns both credentials decrypted. |
| [G7] | New class `LogSettingServiceCryptoTest`. The stubbed suites cannot prove `Tool.encryptPassword` and `Tool.decryptPassword` are inverses, which is exactly the invariant this bug turned on, so this runs the real encryption once under the `@SreeHome` harness. |

`LogSettingServiceTest` 6/6, `LogSettingServiceCryptoTest` 1/1, `PropertiesControllerTest` 14/14 (adjacent, unchanged).

## Note for release

Once the password is stored encrypted, a node still running the **pre-fix** `ForwardService` will send the ciphertext as the password. The community and enterprise changes need to ship together; merge this first.

Also pre-existing, now affecting a second field: if `INETSOFT_MASTER_PASSWORD` or `password.encryption.key` changes, decryption returns the ciphertext verbatim and the Logging page renders it in a populated-looking field. Saving that page re-encrypts it, producing a doubly-encrypted value.

## Not stale-checked twice

The report filed alongside this also flagged #76048 (`toPassword` encrypting the untrimmed value). That is **already fixed** — `toPassword` calls `sanitizeProperty` first, and [G1]–[G3] pin it. No change needed.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/logviewer/LogSettingService.java
+++ b/core/src/main/java/inetsoft/web/admin/logviewer/LogSettingService.java
@@ -269,7 +269,7 @@ public class LogSettingService {
             "log.fluentd.security.userAuthenticationEnabled",
             Boolean.toString(model.userAuthenticationEnabled()));
          SreeEnv.setProperty("log.fluentd.security.username", sanitizeProperty(model.username()));
-         SreeEnv.setProperty("log.fluentd.security.password", sanitizeProperty(model.password()));
+         SreeEnv.setProperty("log.fluentd.security.password", toPassword(model.password()));
          SreeEnv.setProperty("log.fluentd.tlsEnabled", Boolean.toString(model.tlsEnabled()));
          SreeEnv.setProperty(
             "log.fluentd.tls.caCertificateFile", sanitizeProperty(model.caCertificateFile()));
--- a/core/src/test/java/inetsoft/web/admin/logviewer/LogSettingServiceCryptoTest.java
+++ b/core/src/test/java/inetsoft/web/admin/logviewer/LogSettingServiceCryptoTest.java
@@ -1,0 +1,75 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.web.admin.logviewer;
+
+/*
+ * Test strategy
+ *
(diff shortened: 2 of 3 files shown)


ngovus-inetsoft:
Merge-order note for whoever reviews this: this PR must land **immediately before** its enterprise half, inetsoft-technology/stylebi-enterprise#668, not on its own schedule.

This change makes `setFluentdSettings` store `log.fluentd.security.password` encrypted (`toPassword` instead of `sanitizeProperty`). Enterprise `ForwardService.getClient` still reads that property raw until #668 lands, and `ForwardClient` digests the password rather than sending it — so between the two merges, a password re-saved from Settings > Logging goes to the collector as ciphertext and authentication fails silently. That is precisely the bug #668 fixes, just reached from the other direction.

It only affects someone who builds from community `main` inside that window and re-saves the Fluentd password, and pre-existing plaintext passwords are unaffected (decryption is a no-op on unprefixed values). But the two should be merged back to back.

ngovus-inetsoft:
Correcting the merge order in my comment above: it is the reverse of what I wrote. This PR should merge **after** its enterprise half, inetsoft-technology/stylebi-enterprise#668, not before.

The dependency only runs one way. #668 decrypts the Fluentd password on read, which is a no-op while this PR has not landed and the value is still stored in clear text — so #668 is safe on its own. This PR is the one that cannot go first: once the password is encrypted at rest and the enterprise module still reads it raw, the ciphertext is digested and Fluentd authentication fails, which is exactly the bug #668 fixes for the shared key.

That also means #668 no longer carries a submodule bump referencing this branch (it takes `main`'s pointer), so nothing here is gated on it from the git side — only the ordering matters.

jshobe-inetsoft (approved):

PR #4765, opened 2026-08-24
Bug #76097, support datasource-scoped named groups and full condition operators in add_named_group

jkidacki-inetsoft:
## Summary

`add_named_group`/`edit_named_group` (the AI-assistant worksheet tool backing StyleBI's Composer chat plugin) only supported enumerated equality values for group mappings, and had no way to scope a grouping directly to a datasource/logical-model or physical-table path the way a human does via the Composer's own "Add Grouping" dialog ("Only For" + "Attribute").

## Root Cause

- Group mapping conditions were always built as `EQUAL_TO`, hardcoded — there was no way to express "starts with N" or any other operator.
- The tool only supported attaching to an existing worksheet table's column (`table`+`column`) or a fully standalone, type-only grouping. Neither corresponds to a datasource/logical-model-scoped grouping the way the native Composer dialog creates one.

## Fix

- Added an `operation` field to each group mapping (`WorksheetMutationSupport.GroupMapping`), honored by a new shared `WorksheetMutationSupport.buildGroupConditionList`, supporting the full condition-operator vocabulary already used elsewhere in this tool surface (`=`, `!=`, `<`, `>`, `<=`, `>=`, `BETWEEN`, `ONE_OF`/`NOT_ONE_OF`, `STARTING_WITH`, `CONTAINS`, `LIKE`, `NULL`/`NOT_NULL`). Negated equality is routed through the same set-based logic as `NOT_ONE_OF` rather than per-value OR'd negation, since `EQUAL_TO` only ever reads a condition's first value (per-value negation would otherwise produce a near-tautological always-true condition for a multi-value exclusion).
- Added a new `WorksheetAgentController.addDatasourceScopedNamedGroup`, dispatched from `add_named_group` when `datasource` is given: it resolves a logical-model entity attribute (mirroring `addLogicalModelTable`'s permission checks and lookups) or a physical-table column (mirroring `addBoundTable`'s), and builds the same kind of `SourceInfo`/attribute a human's "Add Grouping" dialog would — independent of any worksheet table.
- The existing `table`+`column` attached mode is intentionally unchanged: its `SourceInfo(ASSET, null, table)` convention is relied on by `CalcTableService.worksheetNamedGroups()`/`FieldRefFactory` to resolve a chart/table/crosstab/calc-table's `namedGroup` binding, so changing it would have silently broken that binding-resolution path for any grouping attached this way.

## Test Plan

- `./mvnw -pl core test -Dtest="inetsoft.web.wiz.worksheet.*Test,CalcTableServiceTest,FieldRefFactoryTest"` — all pass, including new regression tests for the operator support, the datasource-scoped mode, and a guard that the `table`+`column` mode's `SourceInfo` convention stays intact even when the underlying table is a real `BoundTableAssembly`.
- Live-verified via the `stylebi-composer-chat` plugin against a running server: created a named group scoped to `Examples/Orders` → `Order Model` → `Customer:State`, with an `N` group built via `STARTING_WITH`.

Fixes Redmine #76097.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
@@ -69,9 +69,16 @@ import java.util.Map;
  *   <li>{@code reorder_columns} — {@code table}, {@code columnOrder}</li>
  *   <li>{@code add_concat_subtable} — {@code table} (concat assembly), {@code name} (subtable to add)</li>
  *   <li>{@code remove_concat_subtable} — {@code table} (concat assembly), {@code name} (subtable to remove)</li>
- *   <li>{@code add_named_group} — {@code name}, {@code groupMappings}, {@code groupOthers}; either
- *       {@code table} + {@code column} (attach to a column) or {@code type} (standalone grouping,
- *       matched by data type; defaults to {@code "string"})</li>
+ *   <li>{@code add_named_group} — {@code name}, {@code groupMappings} (each mapping's
+ *       {@code operation} is any operator accepted by
+ *       {@link WorksheetMutationSupport#parseOperation}, e.g. {@code "STARTING_WITH"};
+ *       defaults to {@code EQUAL_TO} when omitted), {@code groupOthers}; exactly one of:
+ *       {@code table} + {@code column} (attach to a column on an existing worksheet table);
+ *       {@code datasource} + {@code sourceTable} + {@code attribute} (+ optional
+ *       {@code logicalModel}, or {@code schema}/{@code catalog} for a physical table) to scope
+ *       directly to a datasource/logical-model or physical-table path, matching what a human
+ *       produces via the Composer's own "Add Grouping" dialog; or {@code type} (standalone
+ *       grouping, matched by data type; defaults to {@code "string"})</li>
  *   <li>{@code set_column_description} — {@code table}, {@code column}, {@code description}</li>
  *   <li>{@code set_variable_values} — {@code variableValues} (map of variable name → value)</li>
  *   <li>{@code set_mirror_auto_update} — {@code table}, {@code visible} (true=auto-update on, false=off)</li>
@@ -82,7 +89,8 @@ import java.util.Map;
  *   <li>{@code edit_variable} — {@code name}, {@code type}, {@code label}, {@code defaultValue}</li>
  *   <li>{@code rename_variable} — {@code name}, {@code newName}</li>
  *   <li>{@code delete_variable} — {@code name}</li>
- *   <li>{@code edit_named_group} — {@code name}, {@code groupMappings}, {@code groupOthers}</li>
+ *   <li>{@code edit_named_group} — {@code name}, {@code groupMappings} (see {@code add_named_group}
+ *       for the {@code operation} field), {@code groupOthers}</li>
  *   <li>{@code edit_sql_query} — {@code table}, {@code expression} (new SQL string)</li>
  *   <li>{@code update_mirror} — {@code table}</li>
  *   <li>{@code set_table_mode} — {@code table}, {@code mode} ({@code "live"}, {@code "default"}, {@code "full"}, {@code "detail"}, {@code "edit"})</li>
@@ -208,5 +216,18 @@ public record EditRequest(
    /** True = insert before index, false = append after index (insert_column). */
    Boolean insert,
    /** New order of subtable names for reorder_concat_subtables. */
-   List<String> subtables
+   List<String> subtables,
+   /**
+    * Entity name (when {@code logicalModel} is given) or physical table name (otherwise) for
+    * add_named_group, when scoping the grouping directly to a datasource path — matching what a
+    * human produces via the Composer's own "Add Grouping" dialog ("Only For") — instead of
+    * attaching to a column on an existing worksheet table. Requires {@code datasource} and
+    * {@code attribute}; mutually exclusive with {@code table}/{@code column} and {@code type}.
+    */
+   String sourceTable,
+   /**
+    * Attribute/column name within {@code sourceTable} for add_named_group's datasource-scoped
+    * mode (see {@code sourceTable}).
+    */
+   String attribute
 ) {}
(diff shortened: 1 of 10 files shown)


jkidacki-inetsoft:
## Code review findings

1. **`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java:458`** — `readNamedGroup()` doesn't surface the new `operation` field, so an agent reading back a group created with e.g. `STARTING_WITH` sees it as a plain equality mapping (`{"N": ["N"]}`). If the agent later edits it, `operation` silently defaults to `EQUAL_TO`, turning "starts with N" into "equals N". It also mis-derives `table`/`column` for datasource-scoped groups (using `attachedSource.getSource()` / `attachedAttr.getAttribute()`), making them indistinguishable from worksheet-table-attached groups even though no such worksheet table exists.

2. **`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java:236`** — `sourceTable`/`attribute` are only validated/consumed when `datasource` is non-blank. If `datasource` is omitted, the request falls through to the legacy `Editor.addNamedGroup(table, column, type, ...)` path with `table`/`column`/`type` all null, silently creating a standalone `type=string` grouping and discarding the `sourceTable`/`attribute` the caller asked for, with no error.

3. **`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetMutationSupport.java:182`** — `buildGroupConditionList` doesn't validate value counts for `ONE_OF`/`NOT_ONE_OF`/`BETWEEN`. `BETWEEN` with 1 or 3 values silently never matches (condition requires `values.size() > 1`); `ONE_OF` with an empty list matches nothing; a negated `ONE_OF` (`!=`) with an empty list matches everything — all silent misconfigurations with no test coverage.

4. **`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java:545`** — `addDatasourceScopedNamedGroup` duplicates ~90 lines of permission-check/metadata-resolution logic from `addLogicalModelTable`/`addBoundTable` (plus the assembly-positioning `maxY` loop, duplicated 5 places total) instead of factoring a shared helper. Future fixes to that shared logic risk being applied in one copy and forgotten in the others.

5. **`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java:602`** — The `MODEL`-typed `SourceInfo` built for datasource-scoped named groups omits the `isJDBC` property that `GroupingAssemblyDialogService` always sets for the human-driven equivalent, so a human later opening this grouping's Properties dialog in the Composer sees an incorrect/null `isJDBC` flag on the "Only For" tree entry.

jkidacki-inetsoft:
Addressed in a follow-up commit:

1. **`readNamedGroup` operation/attachment-kind gap** — fixed. `GroupMappingModel` now carries `operation`, and `NamedGroupModel` reports `datasource`/`logicalModel`/`sourceTable`/`attribute` (instead of `table`/`column`) for a datasource-scoped group, distinguished by `attachedSource.getType()`. Also caught in review of this fix: the inverse operator mapping silently dropped negation for `BETWEEN`/`STARTING_WITH`/`CONTAINS`/`LIKE`/comparisons (reachable since a human can create a negated one of these via the general condition editor, even though this tool's own vocabulary can't) — now returns `null` for those instead of the wrong positive string.

2. **Missing server-side datasource guard** — fixed. `editOp` now rejects `sourceTable`/`attribute`/`logicalModel`/`schema`/`catalog` set without `datasource`, mirroring the plugin's existing client-side check.

3. **Missing value-count validation** — fixed. `buildGroupConditionList` now throws for `BETWEEN` with anything other than 2 values, and for `ONE_OF`/negated-equality with zero values.

4. **Duplicated `maxY` positioning loop** — extracted into a shared `positionBelowExisting`, used by the 4 call sites that share the exact same `(25, maxY+40)` positioning (`addBoundTable`, `addLogicalModelTable`, `addDatasourceScopedNamedGroup`, `addSqlQuery`). Left the larger permission-check/metadata-resolution duplication alone — `add_table`'s two methods are each single-branch while the new method has both branches inline, so unifying them is a real design decision, not a mechanical extraction, and felt like more risk than this bug-fix PR warrants.

5. **Missing `isJDBC` on the datasource-scoped `SourceInfo`** — not changed. The existing, already-shipped `addLogicalModelTable`/`addBoundTable` have the identical omission, so this isn't a new gap. I couldn't find anywhere that actually reads `isJDBC` for behavior (only round-trips it for display in one dialog), and populating it correctly would need additional datasource-type lookup work not otherwise needed here — didn't want to guess at a value with no way to verify it's right.

Added regression tests throughout (`WorksheetReadServiceTest` had zero named-group coverage before this; now covers all three attachment kinds, operation recovery including negation, and the negated-beyond-vocabulary case). Full `inetsoft.web.wiz.worksheet` + `CalcTableServiceTest`/`FieldRefFactoryTest` suite passes.

jkidacki-inetsoft:
## Follow-up review

Checked the latest commit (`6f579d7`, "Address review comments on PR #4765") against the findings above:

1. **`WorksheetReadService.java`** — Fixed. `readNamedGroup()` now returns `operation` on each `GroupMappingModel` (correctly reporting `null` when a condition can't be round-tripped, e.g. a negated `STARTING_WITH`), and separates worksheet-table-attached (`table`/`column`) from datasource-scoped (`datasource`/`logicalModel`/`sourceTable`/`attribute`) groups instead of conflating them.
2. **`WorksheetAgentController.java` (datasource guard)** — Fixed. Omitting `datasource` while supplying `sourceTable`/`attribute`/`logicalModel`/`schema`/`catalog` now throws a `PairingException` instead of silently falling through to the legacy path.
3. **`WorksheetMutationSupport.java`** — Fixed. `buildGroupConditionList` now validates value counts and throws for empty `ONE_OF`/`NOT_ONE_OF` and for `BETWEEN` with anything other than exactly 2 values.
4. **`WorksheetAgentController.java` (duplication)** — Fixed. The repeated "position below existing assemblies" loop is now a shared `positionBelowExisting()` helper used by all 4 call sites.
5. **`isJDBC` omission** — On closer look, this isn't a bug. The frontend (`grouping-condition-item-pane-provider.ts`) only disables value-browsing when `isJDBC === "false"`; an absent property falls through to the same "enabled" behavior as `isJDBC=true`. Since this code path only ever attaches to JDBC-backed physical tables/logical models, leaving the property unset is equivalent to the correct value — no fix needed.

All 5 items are handled (4 fixed, 1 confirmed to not be an issue). This looks good to merge from a review standpoint.

ngovus-inetsoft (approved):

PR #4766, opened 2026-08-24
75694, auto complete not working on unchained method calls.

milotalon:
Root Cause:

  The AxisSpec definitions were not the problem - isLabelBetween, setLabelBetween,
  isGridBetween and setGridBetween were all present and correct in
  js-functions.generated.json. The earlier incremental-compile clobbering fix
  (PR #4280) was a real but different bug. What was broken is that nothing in the
  call chain leading to AxisSpec carried a type, in two independent places:

  1. "graph" was never bound to the generated EGraph definition. VSEGraphScriptable.getMemberKeys() returns a hardcoded list of bare member-name strings with no type information, which VSScriptableService turned into untyped nodes (the node itself was typed "+Object" via property-type.properties). So graph.getCoordinate() had no return type and the chain was already dead on the first line.

  2. Coordinate and Scale are abstract bases that carry @TernMethod members but have no @TernClass on the class itself, so they are not tern types. TernAnnotationProcessor therefore emitted "fn() -> ?" for EGraph.getCoordinate() and for RectCoord.getXScale()/getYScale().

  Why it looked inconsistent: when the object type is unknown, tern falls back to
  guessing - at query time it searches every known definition for a matching
  property. That is why

      scale.getAxisSpec().setG      <- did show a completion list
      var spec = scale.getAxisSpec(); spec.setG   <- showed nothing

  Guessed types are never written back into the inference graph, so assigning from an
  unresolvable call yields a null type and "spec." returns zero completions. Zero
  completions means no popup, and because script-pane sets hintDelay to 17000ms, no
  further completion attempt fires after the "." keystroke.

  Fix:

  1. TernAnnotationProcessor now emits a synthetic "union" definition for each unannotated base class that is referenced as a type, merging in the members of every @TernClass subclass, so the base becomes a real tern type and inference propagates through assignments. This intentionally offers members that may not apply to the concrete instance - the alternative is offering nothing at all. 12 unions are generated: Scale (56 members), Coordinate (66), GraphElement (107), VisualFrame (124), ColorFrame, ShapeFrame, SizeFrame, TextFrame, TextureFrame, LineFrame, GraphForm, SchemaPainter.

     - A real @TernClass definition is never replaced by a union. Since @TernClass has SOURCE retention, a base whose source was not recompiled cannot be tested for the annotation, so the synthesized names are recorded in a "!unions" key in the generated file (tern ignores unknown "!" keys) and that distinction survives across builds. Verified that LinearScale, RectCoord, LineElement, LabelForm, GShape and RGBCubeColorFrame keep their own definitions.
     - Unions are built after the merge-with-previous-file step from PR #4280 and are merged into any existing prototype, so an incremental compile can neither shrink a union nor overwrite a real definition. - java.*/javax.* types are excluded, so a method returning Object does not produce a union of every annotated class.

  2. VSScriptableService.bindToGeneratedType() binds the "graph" definition node to the generated EGraph definition ('"!type": "+EGraph"'), removing the untyped duplicates so they cannot shadow the generated members, and keeping the scriptable-only bean-style aliases (coordinate, legendLayout, legendPreferredSize, XTitleSpec, x2TitleSpec, YTitleSpec, y2TitleSpec).

  Result, verified against the regenerated definitions using the script from the report:

      coord -> Coordinate,  scale -> Scale,  spec -> AxisSpec
      spec. now offers 58 members, including setGridBetween, setLabelBetween,
      isGridBetween and isLabelBetween (previously 0)

  This also resolves #75681 (EGraph.getElement() was typed "?"): elem. now offers 107
  members.

  Files changed:
  - build-tools/tern-annotations/src/main/java/com/inetsoft/build/tern/TernAnnotationProcessor.java
  - core/src/main/java/inetsoft/web/binding/VSScriptableService.java

  Note: definitions are generated at compile time, so a clean build of core is needed to
  pick this up. One remaining incremental-build caveat: if a method returning a base type
  is processed in an annotation-processing round before any @TernClass subclass of that
  base has been seen, it keeps "-> ?" for that build. Clean builds are unaffected.

### DIFF
--- a/build-tools/tern-annotations/src/main/java/com/inetsoft/build/tern/TernAnnotationProcessor.java
+++ b/build-tools/tern-annotations/src/main/java/com/inetsoft/build/tern/TernAnnotationProcessor.java
@@ -19,7 +19,7 @@ package com.inetsoft.build.tern;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.databind.node.*;
 import com.google.auto.common.*;
 
 import javax.annotation.processing.*;
@@ -61,7 +61,6 @@ public class TernAnnotationProcessor extends AbstractProcessor {
 
    private void processAnnotations(RoundEnvironment roundEnv) {
       String baseUrl = processingEnv.getOptions().get("tern.baseUrl");
-      Map<String, Set<String>> superClasses = new HashMap<>();
 
       for(Element element : roundEnv.getElementsAnnotatedWith(TernClass.class)) {
          TypeElement classType = MoreElements.asType(element);
@@ -75,6 +74,7 @@ public class TernAnnotationProcessor extends AbstractProcessor {
 
          ClassDef classDef = new ClassDef(className, url.isEmpty() ? null : baseUrl + url);
          definitions.put(fullClassName, classDef);
+         ternClasses.add(fullClassName);
 
          TypeMirror superType = classType.getSuperclass();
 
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4767, opened 2026-08-24
Feature #76073 - Add configurable shadow settings for rectangle and oval

ngovus-inetsoft:
## Problem

You could switch a Shadow on for a Rectangle from the property dialog, but there were no settings to control it. Reported ask: the usual direction, opacity, colour and width controls.

The single boolean was also rendered three different, mutually inconsistent ways:

| Path | What it drew |
|---|---|
| Rectangle, browser | css `box-shadow: 5px 5px 3px 3px rgba(0,0,0,0.3)` |
| Oval, browser | svg filter, `stdDeviation=3`, `dx=2 dy=4`, always black |
| All exports | `VSFaceUtil.addShadow(image, 6)` — fixed grey gaussian blur |

So a dashboard already looked different on screen than in an exported PDF.

## What this does

Adds **Colour, Opacity, Direction (8 compass points), Distance and Blur** to the Shape tab of the Rectangle and Oval property dialogs, stored in a new `ShapeShadow` value object on `ShapeVSAssemblyInfo` and honoured identically by the browser and by every export format.

`ShapeShadow` is modelled on the existing `GradientColor` — same package, same `Cloneable`/`Serializable` shape, same child-element persistence. The direction-to-offset maths lives in exactly one place per language (`ShapeShadow` / `shape-shadow-util.ts`) so the render paths cannot drift, and all three now blur with the same sigma.

The on/off checkbox moves out of the General tab into the new Shadow group beside the settings it controls, for Rectangle and Oval only. Text, Image and Gauge keep their General-tab checkbox and their existing fixed shadow — `VSFaceUtil.addShadow(img, int)` is untouched and still used by `VSImageable` and the gauges.

## Two existing defects fixed along the way

Both had to be fixed for the feature to work at all:

1. **Exports clipped and squeezed the shadow.** The image was generated 6px larger, but every exporter drew it into the assembly's *unexpanded* rectangle (`PDFCoordinateHelper.drawImage` even clips to it). The destination rectangle now grows along with the image. This is why export shadows never matched the screen.

2. **The blur tail was cut off flat.** `ConvolveOp` with `EDGE_NO_OP` leaves the outermost `radius` band unconvolved, and the insets allowed exactly `blur` of margin — so the whole outward falloff landed in that dead band. Measured before the fix, alpha stepped `41 → 0` with no gradient. The margin is now twice the blur radius.

## Compatibility

- An asset saved without a `<shadowInfo>` element parses back to the defaults (black, 30%, south-east, distance 5, blur 6), which approximate the previously hardcoded look.
- The `shadow` boolean is unchanged — still the on/off switch, still persisted, still public script API. A new `shadowInfo` script property is exposed alongside it.

**Deliberate appearance changes**, worth a release note:
- Oval adopts the rectangle's geometry instead of its own `dx=2 dy=4`.
- Exports gain the browser's geometry and a black-at-30% shadow instead of grey, and no longer clip.
- Rectangle on screen shifts very slightly (a 6px blur replacing 3px blur + 3px spread).
- The shadow layer no longer multiplies by the object's Fill alpha, so the Opacity field means what it says.

## Testing

- New `ShapeShadowTest` (24) and `VSFaceUtilShadowTest` (8) — all 8 directions incl. negative offsets, XML round-trip with and without the element, colour fallbacks, inset geometry, and the blur falloff.
- `vs-rectangle.component.spec.ts` (8) — the derived css string, and that the defaults still reproduce the legacy look.
- Portal 959 pass, EM 365 pass, production build and ESLint clean.
- Verified live in the composer and against real exported files.

## Notes for the reviewer

- Three sites must agree or shapes render offset: the canvas growth plus `g.translate` in `VSFloatable.getImage`, the `-insets.left/-insets.top` draw in `VSShape.paintComponent`, and `ShapeShadowUtil.expandForShadow`. There are comments at each.
- Excel anchors to the cell grid, so the offset lands on the nearest column/row boundary rather than exactly — accepted rather than reworking Excel anchoring.
- `getShadowColor()`/`getOffsetX()`/`getOffsetY()` are `@JsonIgnore`d deliberately: as bean properties Jackson walked `java.awt.Color → ColorSpace → ICC profile` and emitted ~9.7 KB per shape on every model refresh. There is a test pinning the payload size.

## Out of scope

Line shadow is inconsistent independently of this change: `LineVSAssemblyInfo` inherits the flag and `VSShape` paints it, but the dialog hides the checkbox, the script property is disabled, and the Angular component ignores it — so a line with `shadowValue="true"` shows a shadow in PDF/Excel/PPT but not in the browser. Left alone here; worth its own ticket.

### DIFF
--- a/core/src/main/java/inetsoft/report/gui/viewsheet/VSFaceUtil.java
+++ b/core/src/main/java/inetsoft/report/gui/viewsheet/VSFaceUtil.java
@@ -18,6 +18,7 @@
 package inetsoft.report.gui.viewsheet;
 
 import inetsoft.sree.portal.PortalThemesManager;
+import inetsoft.uql.viewsheet.ShapeShadow;
 import inetsoft.uql.viewsheet.internal.VSUtil;
 
 import java.awt.*;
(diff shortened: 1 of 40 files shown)


jshobe-inetsoft (changes requested):
Requesting changes based on Claude's automated review — two findings that should be addressed before merge:

1. **Unescaped `direction` attribute on XML write** (`ShapeVSAssemblyInfo.writeContents`, ~line 168-173): `color` is escaped via `Tool.escape`, but `direction` is written raw. `ShapeVSAScriptable` exposes `shadowInfo` as a full read/write bean property with no whitelist check against `ShapeShadow.DIRECTIONS` in `setDirection()`, so a script setting `shadowInfo.direction` to a string containing `"` would corrupt the persisted assembly XML on save. Recommend validating `direction` against `ShapeShadow.DIRECTIONS` in `setDirection()`, or escaping it the same way `color` is at write time.
2. **No server-side bounds validation for `distance`/`blur`/`alpha`**: the dialog services copy these values onto `ShapeShadow` with no clamping server-side (only the Angular dialog clamps client-side). Since `blur`/`distance` size `BufferedImage`s directly during export, an unbounded value reaching the server (bypassing the UI) could trigger huge allocations. Worth clamping in `ShapeShadow.setDistance`/`setBlur`.

See the full review comment on the PR for two additional non-blocking notes (line-shadow export behavior change, and a PR-description claim about round-trip test coverage that doesn't match what's actually tested).

jshobe-inetsoft (approved):

PR #4768, opened 2026-08-24
Bug #76082, refuse reorder_columns on a crosstab worksheet table

jkidacki-inetsoft:
## Summary

The AI Wiz plugin's `reorder_columns` tool silently succeeded when called on a worksheet table in crosstab mode, even though the Composer UI disables "Reorder Table Columns" for crosstab tables entirely. Any requested column name that didn't match the table's static column selection (e.g. a pivoted value like "2024", which isn't a real design-time column) was silently dropped instead of erroring, so the agent reported the reorder as applied when nothing meaningful happened.

## Root Cause

`WorksheetEditService.reorderColumns()` had no guard for crosstab tables. In crosstab mode, column layout is driven by the row/column group configuration (`AggregateInfo`), not by the table's `ColumnSelection` order — this is exactly why the Composer UI's `isSupportChangeColumnOrder()` (`ws-details-pane.component.ts`) disables the reorder button when `aggregateInfo.crosstab` is true. The backend method never mirrored that rule, and its unmatched-name lookup (`Map.remove()` returning `null`) was silently ignored rather than surfaced, so a bad or crosstab-inapplicable request still returned success.

## Fix

Added a guard at the top of `reorderColumns()`: if the table's `AggregateInfo.isCrosstab()` is true, throw a `PairingException` ("Cannot reorder columns on a crosstab table: <table>") instead of proceeding — mirroring the Composer UI's own rule and matching the file's existing validation-error conventions (`PairingException`'s single-arg constructor → HTTP 400).

## Test Plan

- Added `reorderColumnsRefusedOnCrosstabTable` to `WorksheetEditServiceMutatorsTest`, verifying the guard throws for a crosstab table.
- `WorksheetEditServiceMutatorsTest` passes in full (86/86), including the pre-existing `reorderColumns*` tests (bare-name matching, ambiguous-name handling) — unaffected since none use a crosstab table.
- Verified live: reproduced the Redmine repro (group by Order Date-Year + Employee, Sum(Total), switch to crosstab, then `reorder_columns(["2024", "Employee"])`) against a running server built from this branch — now fails loudly with `Cannot reorder columns on a crosstab table: All Sales` instead of returning `{ "ok": true }`.

Fixes Redmine #76082.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -1604,12 +1604,21 @@ public class WorksheetEditService {
        * @param table       the assembly name
        * @param columnOrder ordered list of column names defining the new order.
        *                    Columns not mentioned are appended at the end.
-       * @throws PairingException if the table is not found
+       * @throws PairingException if the table is not found, or if the table is a crosstab
+       *                    (column order there is controlled by the row/column groups, not
+       *                    the column selection — this mirrors the Composer UI, which disables
+       *                    "Reorder Table Columns" for crosstab tables)
        */
       public void reorderColumns(String table, List<String> columnOrder)
          throws PairingException
       {
          TableAssembly t = requireTable(table);
+         AggregateInfo aggInfo = t.getAggregateInfo();
+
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4772, opened 2026-08-25
Let set_table_properties rename the table instead of dropping the name

lizhepu:
`set_table_properties` took an `alias` argument, passed it to `WorksheetEditService`, and dropped it behind a comment while the call returned `{"ok":true}`:

```java
// Table-level alias is not supported in the worksheet assembly model.
// Column-level aliases are set via rename_column instead.
```

The caller got no rename and nothing said so.

## The comment was right about the field, wrong about the capability

There is no alias on a worksheet table — the field lives on `AssetEntry`, `ColumnRef` and `WorksheetInfo`, on no `TableAssembly`. But **a table's displayed name is its name**, so setting one is a rename, and renaming genuinely *is* a table property in this product: `TablePropertyDialogModel` carries `newName`/`oldName` beside description, maxRows and distinct, and its service applies the properties and then renames through `refreshAssembly`, which calls `renameAssembly` when the two names differ.

So this op was one line away from the capability it advertised.

## What changed

`newName` now performs the rename, and **`alias` is accepted as a spelling of it rather than discarded** — a caller that was silently failing starts working, instead of starting to 400.

**The rename runs first, so a failure leaves everything untouched.** Renaming can fail on a name already in use; applying description/maxRows/distinct before finding that out would half-write the patch, which this service refuses everywhere else. The remaining setters cannot fail, so ordering it this way makes the whole call all-or-nothing — there is a test for exactly that.

## One deliberate non-change

A review on the companion plugin PR suggested deleting the `alias` component outright. I kept it. `EditRequest` is a **53-component record constructed positionally in 15 places** — 3 in main, 12 in tests, the latter bare `null` lists with no positional comments. Removing a component means counting to position 48 in fifteen places, and the first attempt at it compiled clean only because incremental compilation had not rebuilt `WorksheetScriptService`; the breakage surfaced as a runtime `NoSuchMethodError`.

Keeping the component costs nothing once it means something, and the lie the review objected to is gone either way — `alias` now renames.

## Verification

5 tests, **verified non-vacuous**: reverting the rename logic fails three of them.

`inetsoft.web.wiz.worksheet.*Test` — **240 tests, all green**, on a branch rebased onto the current integration line (which has moved 19 commits since this work started, including #4765 touching three of the same files).

The plugin half is stylebi-wiz#1783, which sends both names and verifies the rename landed by reading the model back — so it stays honest against a server that predates this change rather than silently reporting success.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
@@ -63,7 +63,8 @@ import java.util.Map;
  *   <li>{@code edit_cell} — {@code table}, {@code row}, {@code col}, {@code value}</li>
  *   <li>{@code insert_row} — {@code table}, {@code index}</li>
  *   <li>{@code delete_row} — {@code table}, {@code index}</li>
- *   <li>{@code set_table_properties} — {@code table}, {@code alias}, {@code description}, {@code maxRows}, {@code distinct}</li>
+ *   <li>{@code set_table_properties} — {@code table}; any of {@code newName} (or {@code alias},
+ *       its accepted spelling), {@code description}, {@code maxRows}, {@code distinct}</li>
  *   <li>{@code add_cross_join} — {@code name}, {@code leftTable}, {@code rightTable}</li>
  *   <li>{@code add_merge_join} — {@code name}, {@code tables}</li>
  *   <li>{@code reorder_columns} — {@code table}, {@code columnOrder}</li>
@@ -116,7 +117,12 @@ public record EditRequest(
    String name,
    /** Data type string, e.g. {@code "string"}, {@code "integer"} (add_column, add_expression_column). */
    String type,
-   /** New alias for rename_column. */
+   /**
+    * The new name for a rename: the column's for {@code rename_column}, the table's for
+    * {@code set_table_properties} -- where it is a rename because a worksheet table has no display
+    * name of its own, the same shape as the Composer's table-properties dialog, whose model carries
+    * {@code newName}/{@code oldName} beside the other properties.
+    */
    String newName,
    /** Column name for filter / sort operations. */
    String field,
@@ -187,7 +193,11 @@ public record EditRequest(
    String value,
    /** Row index for insert_row / delete_row (0-based data row). */
    Integer index,
-   /** Table alias for set_table_properties. */
+   /**
+    * Accepted spelling of {@code newName} for {@code set_table_properties}, kept because callers
+    * reach for it. A worksheet table has no display name apart from its name, so setting an alias
+    * is a rename; {@code newName} wins when both are given.
+    */
    String alias,
    /** Table description for set_table_properties. */
    String description,
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -1511,6 +1511,13 @@ public class WorksheetAgentController {
             // 'alias' is accepted as a spelling of 'newName' rather than dropped: a worksheet table
             // has no display name apart from its name, so an alias request is a rename request. It
             // used to be discarded behind a comment while the call returned success.
+            //
+            // This looks redundant, because the only client of this endpoint is the composer plugin
+            // and that already normalises name/alias to newName before sending. It is here for the
+            // version it CANNOT normalise: a plugin bundle predating that change sends alias alone,
+            // and dist/bin.js ships inside the repo, so a server can outrun the bundle running
+            // against it. Drop this fallback and that pairing silently stops renaming again --
+            // which is the whole defect this op was fixed for.
             editor.setTableProperties(
                req.table(), req.newName() != null ? req.newName() : req.alias(),
                req.description(), req.maxRows(), req.distinct());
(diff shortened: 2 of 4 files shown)


lizhepu:
## Review — one defect found and fixed

**Found:** `Worksheet.renameAssembly` validates only two things — that the old name exists, and that the new one is not already taken:

```java
if(!containsAssembly(oname) ||
   (!oname.equalsIgnoreCase(nname) && containsAssemblyIgnoreCase(nname)))
{
   return false;
}
```

**A blank name passes both.** So a direct caller could `set_table_properties` with `newName: "   "` and rename a table to whitespace, leaving it unaddressable by every later call.

The plugin half already refuses this — but that is exactly the shape of the review comment that started this work: *the plugin guards it, the endpoint does not, and a non-plugin caller still gets the bad behaviour.* The same gap, one layer down, introduced by this PR itself.

**Fixed** in the latest commit: guarded before `renameTable`, so the call stays all-or-nothing. One test, verified non-vacuous — removing the guard fails it.

## Checks that passed

- `requireTable(table)` runs before the rename, so an unknown table is reported against the name the caller passed rather than one that does not exist yet.
- `newName.equals(table)` skips the rename rather than treating it as a collision with itself, with a test.
- Rename-before-properties holds the all-or-nothing property: the remaining setters cannot fail, so the only failure mode aborts before anything is written. Tested by asserting the description does **not** land when the rename is refused.
- Rebased onto the current integration line, which moved 19 commits during this work — including #4765, which touches three of the same files. No conflicts.

`inetsoft.web.wiz.worksheet.*Test` — **241 tests, all green**.

lizhepu:
## Why the server takes three spellings when the plugin sends one

Asked in review, and it is a fair challenge — the composer plugin normalises `name`/`alias` to `newName` before sending, and a search of the repo confirms **it is this endpoint's only client**. Nothing in `wiz-services` or elsewhere calls `set_table_properties`. So on the current wiring, accepting `alias` here is dead weight.

It is here for the version that *cannot* normalise. The pre-fix plugin's mapper sent `alias` and no `newName`:

```ts
(a) => ({ op: "set_table_properties", table: a.table, alias: a.alias, ... })
```

and `dist/bin.js` is committed inside the repo, so a server can be updated ahead of the bundle running against it. With the fallback, that pairing renames correctly. Without it, `alias` goes unread and the call returns ok having changed nothing — **the exact defect this op was fixed for, reappearing for anyone whose bundle lags**.

Removing it is also not a one-line simplification: `EditRequest.alias` would then be read by nothing, and deleting the component means editing 15 positional construction sites, 12 of them bare `null` lists in tests.

No code change, but the reasoning is now a comment on the branch. It was missing, and that was the real gap the question exposed: left unexplained, this reads as redundancy and the next reader deletes it — quietly restoring the bug.

241 tests still green.

lizhepu:
## Review — no blocking issues, and it is now live-verified

The blank-name gap an earlier pass found is fixed. Nothing new surfaced.

### Verified against a running server, not just unit tests

Built and exercised through the plugin against a live Composer:

| Behaviour | Result |
|---|---|
| `alias: "SalesOrders"` on `Order1` | table actually renamed |
| other properties across the rename | `description`, `maxRows`, `distinct`, `mode`, position all survived |
| rename onto an existing name | refused — **and the `description` in the same call did not land** |

That last row is the one worth having. Rename-before-properties exists so a refused rename leaves nothing half-written, and it now has evidence rather than an argument.

### Checks

- `requireTable(table)` before the rename, so an unknown table is reported against the name the caller passed.
- `newName.equals(table)` skips rather than colliding with itself.
- The blank-name guard sits before `renameTable`, keeping the call all-or-nothing.
- 3 commits, 4 files, +171/−9. Rebased onto the current integration line, 0 behind, no conflicts.
- `inetsoft.web.wiz.worksheet.*Test` — **241 tests green**.

### The one thing a reader should not "simplify"

Accepting `alias` here looks redundant, since the plugin normalises `name`/`alias` to `newName` and is this endpoint's only client. It is there for the bundle that *cannot* normalise: `dist/bin.js` ships inside the repo, so a server can outrun the plugin running against it, and a pre-fix bundle sends `alias` alone. Removing the fallback silently restores the original defect for that pairing. That reasoning is now a comment on the branch — it was missing, which is exactly how such a line gets deleted later as dead weight.

This unblocked L1 cases 1.16 and 1.19 in the test plan, together with the already-merged #4755.

lizhepu:
To be concrete about what the fallback buys, since "looks redundant" is the natural read: **`alias` is a working way to rename a table**, not a deprecated spelling kept for tidiness.

All three of these rename, at the tool:

```
{ table: "T", newName: "Scores" }
{ table: "T", name:    "Scores" }
{ table: "T", alias:   "Scores" }
```

and both of these do at the endpoint:

```
{ op: "set_table_properties", table: "T", newName: "Scores" }
{ op: "set_table_properties", table: "T", alias:   "Scores" }
```

The reason it is a real spelling rather than a courtesy: a worksheet table has no display name separate from its name. So a caller who says "give this table an alias" is asking for a rename — there is nothing else it could mean. Refusing that, or dropping it, would be correcting the caller's vocabulary rather than doing what they plainly asked for.

Verified live: `set_table_properties` with `alias: "SalesOrders"` renamed `Order1` and left description, maxRows, distinct, mode and position intact.

fredaan (approved):

PR #4774, opened 2026-08-25
Bug #76189, add a real create-folder endpoint for the wiz Save dialog

Zxm-happy:
## Summary
- The wiz Save dialog's "New Folder" action had no dedicated create-folder endpoint: a folder only ever became real as a side effect of `WizVisualizationService#saveVisualization` writing a viewsheet into it via `ensureFolder()`. Closing the dialog before saving meant the folder was never persisted, so reopening the dialog never showed it.
- Adds `POST /api/wiz/visualization/folders/create` to `WizVisualizationController`, backed by `WizVisualizationService#createVisualizationFolder`.
- Mirrors the existing wiz create-folder convention (`WizDatabaseController#createFolder`) by reusing `WizFolderCreateRequest`/`WizFolderSaveResult` instead of minting new request/response types.
- Validates the target path stays under `VISUALIZATION_COMPONENTS_FOLDER_PATH`, checks WRITE permission on the parent folder, and reports `DUPLICATE_NAME` as a normal `{ok:false}` result (not an exception) — same shape the datasource-folder endpoint already uses.

## Test plan
- [x] 7 new unit tests in `WizVisualizationServiceTest` (blank/invalid name, path-outside-managed-folder guard, permission check, permission denial propagation, duplicate-name short-circuit, default-to-root path resolution) — all 12 tests in the class pass
- [x] `mvn -pl community/core -am compile` verified clean
- [x] Verified live: paired wiz-services/portal change (branch `bug-76189` in `stylebi-wiz`) creates a folder via this endpoint, closes the dialog without saving, reopens it — folder persists

Paired with `stylebi-wiz` PR #1798, which is the only caller of this endpoint.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizVisualizationController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizVisualizationController.java
@@ -101,6 +101,9 @@ public class WizVisualizationController {
       catch(IllegalArgumentException e) {
          return ResponseEntity.badRequest().body(Map.of("error", e.getMessage()));
       }
+      catch(SecurityException e) {
+         return ResponseEntity.status(HttpStatus.FORBIDDEN).body(Map.of("error", e.getMessage()));
+      }
       catch(Exception e) {
          LOG.error("Failed to create visualization folder", e);
          return ResponseEntity.internalServerError().body(Map.of("error", "An unexpected error occurred. Please try again."));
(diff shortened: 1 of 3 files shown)


Zxm-happy:
Addressed all 3 inline observations in 346f588:

1. **`parentPath: "/"` bug**: now treated as root, matching `WizFolderCreateRequest.parentPath`'s documented contract and mirroring `WizDatabaseController#normalizePath`. Added `createFolderTreatsSlashParentPathAsRoot`.
2. **Permission denial surfaced as 500**: `checkAssetPermission`'s `MessageException` on WRITE denial is now converted to `SecurityException`, and the controller gained a `catch(SecurityException e)` branch (matching `render`/`dashboard` in the same class) mapping it to a real 403. Updated `createFolderPropagatesPermissionDenial` → `createFolderConvertsPermissionDenialToSecurityException` to assert the new type.
3. **Prefix-boundary hardening**: tightened the "under the managed folder" guard to `parentPath.equals(...) || parentPath.startsWith(... + "/")`. Added `createFolderRejectsASiblingPathSharingOnlyAPrefix`.

14/14 tests pass (`mvn -pl community/core test -Dtest=WizVisualizationServiceTest`), `mvn compile` clean.

jshobe-inetsoft (approved):

PR #4775, opened 2026-08-25
feat: add write path for date-comparison comparisonOption

larryliang-inetsoft:
## Summary
- `comparisonOption` (value/change/percentChange) was readable via `get_date_comparison` but had no write path: `DateComparisonRequest`/`Comparison` carried no field for it and `apply()` never called `model.setComparisonOption(...)`.
- Adds `comparisonOption` to `DateComparisonRequest` (`ViewsheetAssemblyAgentController.java`) and to `DateComparisonService.Comparison`, wires it through `apply()` via a local name-to-int table (`value`→6, `change`→2, `percentChange`→1, matching `Calculator.VALUE/CHANGE/PERCENT`), and throws `IllegalArgumentException` naming the field on an unrecognized token.
- `read()` now returns the named token instead of the raw int (matching the existing "no magic number reaches the caller" policy `VisualFrameAliases` already follows for frames); an out-of-vocabulary raw int (not reachable through `apply()`, but possible from other paths) reads back as `null` rather than the bare int.
- This is the Java half of a two-repo design; the plugin/TypeScript half (schema + name-mapping in `dateComparisonTools.ts`) is a separate PR in `stylebi-wiz`. Design doc: `docs/superpowers/specs/2026-08-25-date-comparison-option-and-frame-design.md` §3.2.

## Test plan
- [x] `DateComparisonServiceTest`: 24/24 passing (`./mvnw -o -pl core test -Dtest=DateComparisonServiceTest`), including 5 new cases:
  - `apply()` sets the model to 6/2/1 for value/change/percentChange
  - `apply()` throws `IllegalArgumentException` naming `comparisonOption` for an unrecognized token
  - `apply()` leaves an already-set `comparisonOption` untouched when the field is omitted
  - `read()` returns the named token for a model at each of 6/2/1
  - `read()` returns `null` (not the raw int) for an out-of-vocabulary value
- [ ] Not verified against a live StyleBI instance (per design §7) — only unit-tested plumbing.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/DateComparisonService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/DateComparisonService.java
@@ -17,6 +17,7 @@
  */
 package inetsoft.web.wiz.viewsheet;
 
+import inetsoft.uql.viewsheet.graph.Calculator;
 import inetsoft.web.composer.model.vs.*;
 import inetsoft.web.composer.vs.dialog.DateComparisonDialogService;
 import inetsoft.web.wiz.binding.VisualFrameAliases;
@@ -63,7 +64,11 @@ public class DateComparisonService {
     */
    public record Comparison(Integer periods, String level, String endDate, boolean endToday,
                             String interval, Boolean useFacet, Boolean onlyShowMostRecentDate,
-                            Map<String, Object> frame) {}
+                            String comparisonOption, Map<String, Object> frame) {}
+
+   /** The agent vocabulary's comparisonOption tokens, mapped to {@link Calculator}'s constants. */
+   private static final Map<String, Integer> COMPARISON_OPTIONS = Map.of(
+      "value", Calculator.VALUE, "change", Calculator.CHANGE, "percentChange", Calculator.PERCENT);
 
    /** The current settings, normalized. Never echoes the raw cell format. */
    public Map<String, Object> read(String sessionToken, Principal user, String assemblyName)
@@ -88,7 +93,7 @@ public class DateComparisonService {
       }
 
       out.put("enabled", true);
-      out.put("comparisonOption", model.getComparisonOption());
+      out.put("comparisonOption", describeComparisonOption(model.getComparisonOption()));
       out.put("useFacet", model.isUseFacet());
       out.put("onlyShowMostRecentDate", model.isOnlyShowMostRecentDate());
       out.put("period", describePeriod(model.getPeriodPaneModel()));
@@ -200,6 +205,18 @@ public class DateComparisonService {
          model.setVisualFrameModel(VisualFrameAliases.create("color", comparison.frame()));
       }
 
+      if(comparison.comparisonOption() != null) {
+         Integer option = COMPARISON_OPTIONS.get(comparison.comparisonOption());
+
+         if(option == null) {
+            throw new IllegalArgumentException(
+               "comparisonOption must be one of: value, change, percentChange. Got " +
+               comparison.comparisonOption() + ".");
+         }
+
+         model.setComparisonOption(option);
+      }
+
       PeriodPaneModel periods = model.getPeriodPaneModel();
 
       if(periods == null) {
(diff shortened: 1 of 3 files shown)


larryliang-inetsoft:
**Reviewer: content verified, no blocking issues.** (Comment, not formal review -- self-review blocked, same account issue as recent PRs.)

Verified against actual source rather than trusting the PR description:
- `Calculator.java` constants confirmed exactly: `PERCENT=1`, `CHANGE=2`, `CUSTOM=5`, `VALUE=6` -- matches both the PR's `COMPARISON_OPTIONS` map/`describeComparisonOption` switch and the design doc's §2.1 table.
- `apply()`'s new block only sets `comparisonOption` when non-null, matching the same "omission means leave alone, not reset" discipline established elsewhere in this session's reviews (e.g. the calc-cell binding fix) -- confirmed by the new `leavesComparisonOptionUntouchedWhenNotMentioned` test.
- `describeComparisonOption`'s out-of-vocabulary → `null` (not a raw int) is a good design choice and has a dedicated test (`readsAnOutOfVocabularyComparisonOptionAsNullRatherThanARawInt`, using `Calculator.CUSTOM=5` specifically, confirmed not in the vocabulary).
- Test harness bug fix (Mockito `CALLS_REAL_METHODS` bypassing the real field initializer, defaulting `comparisonOption` to 0 instead of `Calculator.VALUE`) is a legitimate finding, and the fix (explicit initial value + assert-untouched) is the right call rather than asserting a mock-only artifact.

One thing worth tracking, not a defect in this PR: `read()`'s response shape for `comparisonOption` changes from a raw int to a string name (or `null`) -- this is explicitly documented as intentional in the design doc (§2.1, "existing comparisonOption key changes shape from a raw int... to a name string"), and I confirmed nothing in the current plugin/TS code consumes `comparisonOption` today, so there's no existing breakage. But it does mean this PR and the separate plugin/TS PR are contract-coupled the same way the json-endpoint-lookup Java/TS pair was earlier -- worth cross-checking the TS side's vocabulary strings (`value`/`change`/`percentChange`, case-sensitive) match this Java map exactly once that PR is up, same as I did for #4771/#1791.

No blocking issues. CI: build still pending, Semgrep/CLA pass, claude-review failed with the same known infra signature as the rest of this session's PRs.

jshobe-inetsoft (changes requested):
Requesting changes based on Claude's automated review:

- The 3-token vocabulary (`value`/`change`/`percentChange`) omits two commonly-used settings: `DateComparisonInfo.CHANGE_VALUE` and `DateComparisonInfo.PERCENT_VALUE` ('Change and Value' and 'Percent Change and Value'), which are 2 of the 5 options in the composer's own date-comparison dialog dropdown — not obscure or unreachable states. This affects both the read path (reports `comparisonOption: null` for these settings) and the write path (no way to set them). Please either extend the vocabulary to cover these two options, or explicitly correct the docstring/test framing to acknowledge the gap so it isn't mistaken for full coverage.

larryliang-inetsoft:
Superseded by #5028, which rebases this same fix onto the current base (this branch predates #4818's frame-parameter removal and #5027's chart-type-override disclosure) and addresses the review comment above by naming `CHANGE_VALUE`/`PERCENT_VALUE` on read (`changeAndValue`/`percentChangeAndValue`) instead of falling back to `null`. Leaving this PR open for you to close at your discretion rather than closing it myself.

larryliang-inetsoft:
Closing in favor of #5028 (same fix, rebased onto current base, addresses the review's CHANGE_VALUE/PERCENT_VALUE naming request).

larryliang-inetsoft:
Reopening — the coordinator clarified this is not an independent human's PR to route around; continuing directly on this branch instead of the now-closed #5028 (which I've closed). Pushed a rebase onto current `stylebi-wiz-main-rebase` (this branch predated #4818's frame-parameter removal and #5027's chart-type-override disclosure) plus the minimal fix the review asked for: docstring/test framing now explicitly acknowledges that `CHANGE_VALUE`/`PERCENT_VALUE` still read back as `null` and aren't settable via this tool yet, rather than extending the vocabulary. Re-requesting your review, @jshobe-inetsoft.

PR #4777, opened 2026-08-25
Bug #76058, support crosstab mode in set_group_aggregate for the wiz agent

jkidacki-inetsoft:
## Summary

Worksheets support crosstab mode natively via `AggregateInfo.crosstab` (the Composer's Group and Aggregate dialog's own "Switch to Crosstab" toggle), but the wiz agent's `set_group_aggregate` operation never accepted or applied it anywhere in the stack. With no way to expose the feature, the AI agent told users worksheets have no crosstab concept at all (Redmine #76058) when asked to switch a Group-and-Aggregate table to one.

## Root Cause

`WorksheetMutationSupport.applyAggregateInfo` built a new `AggregateInfo` from the request's groups/aggregates but never called `setCrosstab(...)`, and the read-side model (`WorksheetReadService`/`WorksheetModel.AggregateModel`) never reported the current crosstab state either. The gap ran the whole way from the wire-level `EditRequest` down to the backing `AggregateInfo`.

## Fix

- `WorksheetMutationSupport.applyAggregateInfo` gained a new 4-arg overload accepting `boolean crosstab`, which calls `AggregateInfo.setCrosstab(crosstab)` before committing it to the table. The original 3-arg method now delegates to it with `false`, so its existing callers are unaffected.
- `WorksheetEditService.setGroupAggregate` gained the same 4-arg overload/delegation pattern.
- `WorksheetAgentController`'s `set_group_aggregate` dispatch now forwards `req.crosstab()` (treating `null` as `false`).
- `EditRequest` gained a `Boolean crosstab` field, appended as the record's last component (not inserted mid-list) plus a hand-written compatibility constructor, so the ~15 existing positional `new EditRequest(...)` call sites across tests/services don't need to change.
- `WorksheetModel.AggregateModel` gained a `boolean crosstab` field, populated from `AggregateInfo.isCrosstab()`, so `read_worksheet_model` reports the current state back.
- Documented that `crosstab: true` only takes visible effect once the table has at least 2 groups and 1 aggregate (matching `AggregateInfo.isCrosstab()`'s own definition and the Composer dialog's existing behavior) — fewer than that is accepted but silently has no effect.

## Test Plan

- [x] New unit test `setGroupAggregateCrosstabTogglesAggregateInfo` (`WorksheetEditServiceMutatorsTest`) covers crosstab true/false/default-omitted.
- [x] `WorksheetEditServiceMutatorsTest`, `WorksheetAgentControllerTest`, `WorksheetReadServiceTest`, `GroupSpecDeserializationTest` — 204 tests, all passing.
- [x] `mvnw compile -pl core` — clean build.
- [x] Independent code-reviewer agent pass — no blocking findings.
- [x] Live-verified end-to-end with the paired `stylebi-composer-chat` plugin against a rebuilt server (per user confirmation).

Companion PR in stylebi-wiz (plugin side): https://github.com/inetsoft-technology/stylebi-wiz/pull/new/bug-76058

Fixes Redmine #76058.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
@@ -38,7 +38,14 @@ import java.util.Map;
  *   <li>{@code add_filter} — {@code table}, {@code field}, {@code operation}, {@code values}</li>
  *   <li>{@code remove_filter} — {@code table}, {@code field}</li>
  *   <li>{@code set_group_aggregate} — {@code table}, {@code groups} (each a column name or
- *       {@code {field, dateLevel}}), {@code aggregates}</li>
+ *       {@code {field, dateLevel}}), {@code aggregates}, optional {@code crosstab} (true =
+ *       display as a crosstab, matching the Group and Aggregate dialog's "Switch to
+ *       Crosstab" toggle; defaults to false). Per
+ *       {@link inetsoft.uql.asset.AggregateInfo#isCrosstab}, the
+ *       request is accepted either way but the crosstab actually takes effect only with at
+ *       least 2 {@code groups} and at least 1 {@code aggregates} entry — fewer than that and
+ *       the table silently reads back as a non-crosstab table on the next read, with no
+ *       error from this call.</li>
  *   <li>{@code add_expression_column} — {@code table}, {@code name}, {@code expression}, {@code type}, {@code sql}</li>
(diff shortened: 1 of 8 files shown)


jkidacki-inetsoft:
Addressed the review's minor test-coverage note: added `editDispatchesSetGroupAggregateWithCrosstab` to `WorksheetAgentControllerTest`, which builds a `crosstab: true` `EditRequest` and confirms it reaches `AggregateInfo.isCrosstab()` through the controller's dispatch switch (not just at the `Editor` layer already covered). 172 tests pass across both files.

ngovus-inetsoft (approved):

PR #4778, opened 2026-08-25
Bug #76293, fix the Save Viewsheet dialog and tidy the toolbar property plumbing

ngovus-inetsoft:
Six findings were filed from documenting the `viewsheet.*` / `vs.*` properties. All six were re-verified against the working tree; one turned out to be partly mistaken and is handled differently from how it was filed (see below). A code review of the resulting diff surfaced two more defects of the same class as the first, which are also fixed here.

Everything is in `community`, so this is a community-only PR.

## The user-visible bug

`viewsheet-options-pane` is shared by the Viewsheet Property dialog and the Save Viewsheet dialog, but `SaveViewsheetDialogService.getSaveViewsheetInfo` populated only part of the model. Three fields were missing:

- **`autoRefreshEnabled`** — the pane disables the Server-Side Update checkbox on `!model.autoRefreshEnabled`, and only `ViewsheetPropertyDialogService.getViewsheetInfo` set it. The field is a primitive `boolean`, so it serialised as `false` and the checkbox was **permanently greyed out in the Save dialog** regardless of `vs.auto.refresh.enabled`, while staying operable in the Property dialog. This is the originally filed finding.
- **`snapGrid`** — `saveViewsheet()` writes this back unconditionally, and `VSOptionsPaneModel.snapGrid` defaults to `20`, so **Save As silently reset any other snap grid to 20**. Both defaults being 20 is why this went unnoticed; it only bites someone who changed the snap grid in the Property dialog first.
- **`onDemandMvEnabled`** — the pane gates the whole Create MV On-demand checkbox on it, so the `createMv` value populated two lines above was unreachable from this dialog. No data loss, but the control was invisible.

The last two were found by review, not in the filed issue. All three are the same populate-side omission, which suggests the two dialogs' populate blocks have drifted; a shared populate helper may be worth a follow-up, but that is out of scope here.

## Maintainability findings

**Toolbar property/action pairing.** `SUtil.ToolBarElement` held `VSTOOLBAR_ELEMENTS` (14 property names) and `VSTOOLBAR_ELEMENT_ID` (14 action ids) as separate arrays zipped by index in `getVSToolBarElements`. Nothing enforced equal length or matching order, and inserting a button into one and not the other would silently shift every pairing below it — surfacing as unrelated buttons responding to the wrong property. They are now one `List<ToolBarElementDef>` of `(property, actionId)` records.

Declaration order, `ToolBarElement.index`, the `compareTo` "Close last" rule, the public fields, `toString()`, and the `vs.import.button` FORM-licence skip are all preserved. Consumers (`PresentationToolbarSettingsService`, `VSEventUtil`) match by `id`/`name` and are untouched.

**Two names do not match their buttons.** `vs.previous.button` controls Undo and `vs.next.button` controls Redo. These are shipped names that appear in stored operator settings, so renaming them would strand existing configuration — they are documented in place instead. Note the EM toolbar page shows the action id's catalog label rather than the property name, so the mismatch is only visible to someone setting the property directly.

**Bookmark has a second consequence.** `AnnotationVSUtil.resetDataAnnotation` reads the Bookmark action's visibility and folds every chart data annotation into the chart tooltip when it is hidden — the behaviour a user gets by turning annotations off. Bookmark is the only one of the fourteen action ids read anywhere but the toolbar. Documented at both ends; no behaviour change.

**Write-only crosstab attributes.** `CrosstabVSAssemblyInfo.writeAttributes` emitted `sortOnHeader`, `sortAggregate` and `sortDimension` into stored crosstab XML. `parseAttributes` reads only `colCount` and `cubeType`, and nothing else in community or enterprise parses them; the viewer recomputes all three from their properties in `VSCrosstabModel` on every model construction. The writes are removed — they enlarged every saved crosstab and looked authoritative to anyone reading the stored XML. Existing assets carrying the attribute still load, ignored exactly as before.

## Correction to the "inert declarations" finding

The issue reported that `vs.import.button`, `graph.script.action.support` and `graph.textlayout.maxstep` are declared in `defaults.properties` but never consulted, and implied the declarations should go. That is not right — `defaults.properties` **is** read, by `SreeEnv.resetProperty` for global property resets and by `/api/admin/properties/defaults`, which is what EM's All Properties page displays as the shipped default. Deleting those lines would have been a regression.

The actual defect is narrower in cause and wider in reach: `DefaultProperties.getProperty(key, defaultValue)` reads only `mainProperties` and returns the caller's default, unlike the single-arg overload directly above it. **Every** read that passes a call-site default bypasses `defaults.properties` — repo-wide, not these three properties. Changing that is deliberately not attempted here: it would flip behaviour anywhere a call-site default differs from a declared one, with a blast radius that cannot be audited in this change.

Two of the three reads also cannot be cleanly converted:

- `SUtil` passes `forceGlobal` into the `earlyLoaded` slot, routing the read through `EarlyLoadedProperties`, whose default layer is the JVM system properties rather than `defaults.properties`. That path never sees the declaration under any overload — and dropping the `"true"` would resolve the other thirteen buttons to `null`, hiding every toolbar button.
- `FreeHelper` goes through `GTool.getProperty(String, String)`, the only overload on the `GImpl` SPI; converting it means widening the interface and both implementations.

So only `VGraphPair.isChangedByScript()` is converted, making `graph.script.action.support` genuinely sourced from its declaration. The other two are documented, and the new `PropertyDefaultsTest` pins each call-site default to its declared value — the real protection against the risk the issue names, which is editing `defaults.properties` while the call site keeps the old value.

## Testing

`PropertyDefaultsTest` (6 tests) pins the call-site defaults, the Undo/Redo mapping, and toolbar property/action-id uniqueness. It reads `defaults.properties` straight off the classpath, so it needs no server environment.

`core` compiles clean; `PropertyDefaultsTest`, `PresentationSettingsControllerTest`, `CrosstabVSAScriptableTest` and `VSCrosstabBindingScriptableTest` pass (18 tests).

Still wanted before merge, since they need a running server:

- Save a viewsheet via Save As and confirm the Server-Side Update checkbox is operable; set `vs.auto.refresh.enabled=false` and confirm it greys out in both dialogs.
- **Set a non-default snap grid, Save As, and confirm it survives** — this is the data-loss path.
- With `mv.ondemand=true`, confirm the Create MV On-demand checkbox now appears in the Save dialog.
- Save a crosstab and confirm no `sortOnHeader`/`sortAggregate`/`sortDimension` in the stored XML, that header sorting still works with `vs.crosstab.sortonheader=true`, and that a pre-existing crosstab asset still loads.
- Toggle all fourteen buttons in EM → Presentation → Toolbar Options, confirm each matches its button (Undo/Redo especially), exercise Reset, and confirm Close still sorts last.

### DIFF
--- a/core/src/main/java/inetsoft/graph/internal/text/FreeHelper.java
+++ b/core/src/main/java/inetsoft/graph/internal/text/FreeHelper.java
@@ -32,6 +32,10 @@ public class FreeHelper extends LabelHelper {
    public FreeHelper(VLabel label, VGraph vgraph) {
       super(label, vgraph);
 
+      // GTool/GImpl only exposes a read that takes a call-site default, and
+      // DefaultProperties.getProperty(key, def) returns that default without consulting the
+      // defaults layer, so the graph.textlayout.maxstep line in defaults.properties is not
+      // reachable from here. Keep the two values in sync -- PropertyDefaultsTest pins them.
       String maxstep = GTool.getProperty("graph.textlayout.maxstep", "1000");
 
       if(maxstep != null) {
--- a/core/src/main/java/inetsoft/report/composition/graph/VGraphPair.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/VGraphPair.java
@@ -2616,8 +2616,12 @@ public class VGraphPair {
     */
    public boolean isChangedByScript() {
       // fix customer bug1365805534414
-      return isStructureChanged && !"true".equals(
-         SreeEnv.getProperty("graph.script.action.support", "false"));
+      // read without a call-site default so the graph.script.action.support line in
+      // defaults.properties is the single source of the shipped default. DefaultProperties
+      // .getProperty(key, def) returns the caller's default without consulting the defaults
+      // layer, so passing one here would make that declaration unreachable at runtime.
+      return isStructureChanged &&
+         !"true".equals(SreeEnv.getProperty("graph.script.action.support"));
    }
 
    /**
--- a/core/src/main/java/inetsoft/sree/internal/SUtil.java
+++ b/core/src/main/java/inetsoft/sree/internal/SUtil.java
@@ -388,22 +388,28 @@ public class SUtil {
     * @return the toolbar elements stored in a <code>List</code> instance.
     */
    public static List<ToolBarElement> getVSToolBarElements(boolean forceGlobal) {
-      List<ToolBarElement> elems = new ArrayList<>(ToolBarElement.VSTOOLBAR_ELEMENTS.length);
+      List<ToolBarElement> elems = new ArrayList<>(ToolBarElement.VSTOOLBAR_ELEMENTS.size());
 
-      for(int i = 0; i < ToolBarElement.VSTOOLBAR_ELEMENTS.length; i++) {
-         String name = ToolBarElement.VSTOOLBAR_ELEMENTS[i];
+      for(int i = 0; i < ToolBarElement.VSTOOLBAR_ELEMENTS.size(); i++) {
+         ToolBarElementDef def = ToolBarElement.VSTOOLBAR_ELEMENTS.get(i);
+         String name = def.property();
 
          if("vs.import.button".equals(name) && !LicenseManager.isComponentAvailable(LicenseManager.LicenseComponent.FORM)) {
             continue;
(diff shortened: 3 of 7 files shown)


jshobe-inetsoft (approved):

PR #4779, opened 2026-08-25
Bug #76295, fix three Presentation and share settings defects

ngovus-inetsoft:
Three of four findings from a server-property audit of the Presentation page. All four were re-verified against the current `main` before filing — the original report was written against `935c85a58`, a merge commit that is not an ancestor of `main`.

## 1. Font mapping reset wrote an empty string

`PresentationFontMappingSettingsService.resetSettings()` wrote `""`; every other `resetSettings()` on the page goes through `SreeEnv.resetProperty()`. An empty stored value is not an unset one — the one-argument accessor sees an empty string where it expected a default, the two-argument form treats it as absent, and neither is "reset".

`pdf.font.mapping` has no entry in `defaults.properties`, so `resetProperty` resolves a `null` default and removes the property.

Impact is limited to stored state: both readers (`PDF3Generator:58` and this service's own `getModel()`) null-and-empty check before parsing, so no output changes today.

## 2. The Composer export path ignored `pdf.output.attachment`

`ExportControllerService` computed the `"embed"` comparison and then discarded it, and `ExportController` hardcoded `"attachment"`. `VSExportService` honours the setting. The two export paths disagreed about user-visible behaviour.

The audit was unsure whether the read was vestigial or a lost branch — it was a lost branch, and it affects the **common** case rather than an edge case: `embedded` requires `!matchesAssetIdFormat`, which is set false exactly when no asset entry resolves, i.e. the ordinary Composer runtime-id export.

The disposition is now computed where `rvs` is in scope, reusing `VSExportService`'s rule with its operator precedence preserved verbatim so the two paths cannot drift again, and carried out on `ViewsheetExportResult`.

**No enterprise change needed.** The four-argument `ViewsheetExportResult` constructor is retained, delegating with `"attachment"`, so `ViewsheetApiService` compiles unchanged and the public download API keeps returning `attachment`. Making a download API return `inline` would be a separate, unrequested behaviour change.

Non-PDF formats still get `attachment` via the rule's `PDF == format` term; `print` still forces `inline` on both paths; `setResponseHeader` normalizes the string, so no new header-injection surface.

## 3. A custom loading logo removed the loading message

`index.html` gated the built-in `app.loading` span on `portal.customLoadingLogo` being empty, so adding a logo silently removed the message unless the operator also set `portal.customLoadingText`. It is now gated on `customLoadingText`: a logo no longer suppresses the text, and a custom text still replaces the built-in one. The logo continues to replace the animated indicator.

One correction to the report, which said neither property's Enterprise Manager control mentions the coupling: `portal.customLoadingLogo` has **no** EM control at all — only `customLoadingText` does. The docs half is filed separately as #76181.

## 4. Open Graph — not fixed here, and the reported mechanism is wrong

The defect is real: `share.opengraph.*` is written organization-scoped but read unscoped in `HomePageController`. An organization sets its own values, sees them in Enterprise Manager, and a social network never shows them.

The report attributes this to "no principal, so `useAvailableOrgProperty` cannot substitute an org-scoped name". That path exists but is usually **not** the one taken. `AnonymousUserFilter` gives an unauthenticated request an anonymous principal, and the org comes from `getCookieRecordedOrgID`, which falls back to `Organization.getDefaultOrganizationID()` when the request carries no cookies. A crawler carries none — so the resolved org is the *default* organization, not the owning one. Both routes reach the same outcome, so the conclusion stands, but the reason should be recorded correctly.

Left for a follow-up because the fix is a product decision, not a bug fix: the Open Graph branch keys off `/app/viewer/view/<assetId>`, and the asset-ID formats built there carry no organization, so resolving it would need host-to-org resolution or a new org segment in shared links.

One thing noted so it is not re-reported: `getOpenGraphImage`'s `cachedImage` is process-wide but keyed on the resolved image URL and recomputes on mismatch, so it is a size-lookup cache and does not leak one org's image to another.

## Testing

`PresentationFontMappingSettingsServiceTest` + `PresentationPdfGenerationSettingsServiceTest` — 22 tests pass. The new reset test verifies `resetProperty` was called *and* that no `setProperty` for that key was, so it fails against the old code rather than passing vacuously. `community/core` and `enterprise` both compile.

Not yet run: end-to-end verification of #2 against a live server (tick the browser-embed option, export from Composer, confirm `Content-disposition: inline`, then confirm Excel/CSV stay `attachment` and the viewer path is unregressed). Worth doing before merge, since #2 is the change with real user-visible behaviour.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/presentation/PresentationFontMappingSettingsService.java
+++ b/core/src/main/java/inetsoft/web/admin/presentation/PresentationFontMappingSettingsService.java
@@ -76,7 +76,7 @@ public class PresentationFontMappingSettingsService {
       objectType = ActionRecord.OBJECT_TYPE_EMPROPERTY
    )
    public void resetSettings() throws Exception {
-      SreeEnv.setProperty("pdf.font.mapping", "");
+      SreeEnv.resetProperty("pdf.font.mapping", false);
       SreeEnv.save();
    }
 }
--- a/core/src/main/java/inetsoft/web/composer/vs/controller/ExportController.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/controller/ExportController.java
@@ -223,7 +223,8 @@ public class ExportController {
          bookmarks, onlyDataComponents, exportAllTabbedTables, csvConfig, principal);
 
       VSExportService.setResponseHeader(
-         new ExportResponse(response), result.getSuffix(), "attachment", result.getFileName(), result.getMime());
+         new ExportResponse(response), result.getSuffix(), result.getDisposition(),
+         result.getFileName(), result.getMime());
 
       BinaryTransfer data = result.getData();
       binaryTransferService.writeData(data, response.getOutputStream());
--- a/core/src/main/java/inetsoft/web/composer/vs/controller/ExportControllerService.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/controller/ExportControllerService.java
@@ -109,6 +109,10 @@ public class ExportControllerService {
 
       boolean embedded = "embed".equalsIgnoreCase(SreeEnv.getProperty("pdf.output.attachment"))
          && !matchesAssetIdFormat;
+      // same rule as VSExportService.writeViewsheetExport() so the two export paths agree
+      String disposition = !rvs.isPreview() &&
+         FileFormatInfo.EXPORT_TYPE_PDF == format &&
+         embedded || print ? "inline" : "attachment";
 
       String key = "/" + ExportControllerService.class.getName() + "_" + runtimeId + "_" + format;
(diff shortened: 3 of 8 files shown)


jshobe-inetsoft (changes requested):
Claude's automated review found the fix itself correct and well-scoped, but flagged one issue it treated as a pre-merge blocker: no unit test was added for the new export-disposition branch in `ExportControllerService.exportViewsheet` (#2), and the PR's own description already notes end-to-end verification for this change hasn't been run. Given Claude explicitly called this out as something to do before merge rather than a nice-to-have, requesting changes to add that test coverage (or confirm the end-to-end verification) before this merges.

jshobe-inetsoft (approved):

PR #4780, opened 2026-08-25
Bug #76298, remove eleven inert html.* property declarations

ngovus-inetsoft:
Fixes the eleven `html.*` properties that do nothing. Same defect as #76052 (fixed in `stylebi-1.2.0`), at larger scale — that sweep covered six properties across five prefixes; this is eleven in one prefix.

## The nine declared and unread

Declared in `defaults.properties` and read by no Java, TypeScript, HTML, XML or properties consumer anywhere in community, enterprise or connectors:

`html.background`, `html.cache.time`, `html.close.button`, `html.directory`, `html.image.scale`, `html.layout`, `html.page.footer`, `html.ratio.table`, `html.session.max.report`

Setting any of them had no observable effect and produced no error or log line, because nothing was listening.

## The two inert-but-with-a-read-site

`html.ratio.font` and `html.ratio.bigFont` were read in the constructor of `Tool.FontSizeMapper` — a class that is never instantiated. A whole-tree search for `FontSizeMapper` returns exactly two hits: the class declaration and its own constructor. There is no `new FontSizeMapper()`, no field of that type and no reflective use, and the class is package-private so nothing outside `inetsoft.util` could construct it in any case. Its only public method, `getFontSize(double, boolean)`, has no callers.

A developer checking "is this read?" finds a read site and stops, so the whole dead class (131 lines) goes with the properties. Neither has a declared default to delete, so that removal is entirely in `Tool.java`; leaving the class would let the next reader rediscover the properties from the code side. Same shape as `mv.data.directory` in #76285.

## How this was verified

- **Direct** — a case-insensitive search across community, enterprise and connectors over Java, TypeScript, HTML, XML and properties files returns the `defaults.properties` declaration and nothing else. `html.directory` was matched as a whole word so it could not be masked by `html.image.directory`.
- **Case** — `PropertiesEngine.computePropertyNameCase` lower-cases every name except four whitelisted patterns, none of which match `html.*`, so no differently-cased read site could have resolved these.
- **Concatenation** — every `"html.` literal in the product's Java was enumerated; all belong to properties that are live.
- **Prefix dispatch** — no `startsWith("html` anywhere in the tree.
- **Docs** — none of the eleven appears in any `.adoc` or `.md`, so the change is documentation-safe.

## Why there is no behavior change

`PropertiesEngine.getDefaultProperties()` loads this file purely as the fallback layer of a `DefaultProperties` chain, so removing keys that nothing reads cannot alter any lookup result. Administrator overrides in KV storage are a separate layer and are unaffected. The EM All Properties page iterates whatever keys the file contains and simply stops listing nine dead rows. `FontSizeMapper` had no callers, so removing it removes no reachable code path.

The six live `html.*` properties are untouched: `html.character.encode`, `html.image.directory`, `html.loading.icon`, `html.ratio.x`, `html.ratio.y`, `html.session.timeout`. The only enterprise touchpoint, `html.session.timeout`, is live — hence no enterprise PR.

## The All Properties list

`html.close.button` and `html.image.scale` were in `PropertiesTool.booleanProperties`, so Settings > All Properties offered a `true`/`false` suggestion list for them. That list is presentation only, not a reader, so the value saved and changed nothing. Removed from that list too, or they would keep appearing for keys that no longer exist. `htmlpresenter.fitline` is live and stays.

Note for the issue record: the filed description says these render as **checkboxes**. They do not — `PropertiesTool.getPropertyOptions` returns `boolean_options`, which `property-settings-view` renders as a two-item autocomplete on a plain text input. Substance unchanged. The description also counts 17 *declared* `html.*` keys; only 14 were declared, as `html.loading.icon`, `html.ratio.font` and `html.ratio.bigFont` exist only in Java.

## Testing

- `mvnw compile -pl core` — passes; confirms the `Tool.java` deletion is brace-balanced and leaves no unused import (`SreeEnv` is still used in 14 other places in the file, so the import stays).
- `npm run test:em` — 103 files, 365 tests, all passing.
- Post-change dangling-reference sweep for all eleven names and `FontSizeMapper` across community, enterprise and connectors returns zero hits.
- `PropertiesTool` has no test coverage; its only consumer is `property-settings-view.component.ts:137`, which assigns the result to `valueOptions`, so removed keys now yield an empty suggestion list with no error path.


jshobe-inetsoft (approved):

PR #4781, opened 2026-08-25
Bug #76297, resolve the Google client id secret on the signup page

ngovus-inetsoft:
## Problem

`styleBI.google.openid.client.id` may hold either a literal Google OpenID client id or, on an install configured with a cloud secrets manager, a reference to a secret that must be resolved through `Tool.getClientSecretRealValue(value, "client_id")`.

Two controllers put that value on a page model as `gClientId`, which the templates render into `data-client_id` for the Google Identity Services button:

- `LoginController` resolved it. Correct.
- `SignupController` read the property directly. **Bug.**

So on an install using cloud secrets, Google sign-in works and Google sign-up initializes its button with the secret reference and fails, with no error naming a property. It also put the reference text into the signup page source. `showSignUpDetailPage` delegates to `showSignUpPage`, so it inherited the same defect.

## Fix

Moved the three Google sign-in property reads into a new package-private `GoogleSignInSupport`, and pointed both controllers at it, so the pages cannot drift apart again. `LoginController`'s private `getGoogleClientId`/`getGoogleScopes` are gone; its login-only `gState` block is untouched.

No behavior change where the client id is stored literally: `Tool.getClientSecretRealValue` returns its argument unchanged outside cloud-secrets mode, passes null/empty through, and returns the original value if a reference cannot be resolved.

## Testing

- New `GoogleSignInSupportTest` — 6 tests, 0 failures. The pinning assertion verifies that both the property value and the `client_id` secret key reach the resolver. `SignupControllerTest` defers the `showSignUpPage` cases because they need the `SreeEnv`/`OrganizationManager` statics mocked, so the helper is tested directly with `mockStatic`.
- `SignupControllerTest` — 17 tests, 0 failures.

Not run: the manual browser checks against a live server with a configured client id, and against a non-`local` secrets backend.

### DIFF
--- a/core/src/main/java/inetsoft/web/portal/controller/GoogleSignInSupport.java
+++ b/core/src/main/java/inetsoft/web/portal/controller/GoogleSignInSupport.java
@@ -1,0 +1,59 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):

PR #4783, opened 2026-08-25
Bug #76292, guard the graph.*.maxcount parse and always report chart truncation

ngovus-inetsoft:
## Summary

Three related defects in the `graph.*.maxcount` family, which caps how many data rows a chart draws shapes for.

### 1. Twelve properties shared one unguarded parse

`GraphTypes.getGeomMaxCount` ended in a bare `Integer.parseInt(mstr)` outside any `try`. A non-numeric value in any of the twelve `graph.*.maxcount` properties threw `NumberFormatException` during *chart generation*, not when the value was stored. It escaped `GraphGenerator.createElement` through `VGraphPair.initGraph`, so the chart failed to render with an opaque `For input string: "100,000"` that did not name the property responsible.

The property is now resolved through a guarded accessor modelled on `ScheduleTask.getTaskTimeout` (added for #76177): a bad value logs one WARN naming the property and the offending value, then falls back to the value shipped in `defaults.properties`. No new default constants were invented — `SreeEnv.getDefaultProperties()` already exposes the shipped values.

Notes on the fallback, both captured in the javadoc:
- It goes to the shipped default rather than a site-wide `sree.properties` value, which matters when a bad *org-scoped* override is what failed to parse. A predictable fallback plus a WARN naming the property is the better trade.
- A value that is simply not configured is not an admin error and is not warned about. In practice it cannot arise here: `PropertiesEngine` chains `internalProperties` to `defaults.properties`, and all twelve names ship a default.

A deliberately negative value still parses and still means "no limit".

### 2. The truncation warning was suppressed at exactly the default that needed it

`GraphElement.getEndRow` raised the user-facing `Truncating shapes to {0}. To display in full, please reduce the number of data rows.` only when the limit was `100000` or less. The shipped defaults are:

| property | default |
|---|---|
| `graph.point.maxcount` | `10000000` |
| `graph.bar.maxcount` / `graph.line.maxcount` / `graph.3dbar.maxcount` | `100000` |

Because the test was strictly greater-than, bar/line/3dbar landed on the reporting side while a **Point chart dropped rows silently at its shipped default** — and Point is the type most likely to be plotting a large row count. A reader could not tell whether they were seeing all their data.

The truncation is now always reported. `CoreTool.addUserMessage` de-duplicates via `existUserMessage`, so the repeated calls made while building geometry still raise a single message. `LOG.debug` is kept alongside so the server log records it unconditionally.

The only shipped limit above the removed threshold is `graph.point.maxcount`, so the behaviour change is confined to the intended case.

### 3. `graph.3dbar.maxcount` was alive only by branch order

`isBar` returns true for `CHART_3D_BAR`/`CHART_3D_BAR_STACK` as well as the flat types, so a 3D bar matches both the `is3DBar` branch and the `isBar` branch. Only the ordering kept `graph.3dbar.maxcount` from being dead, and nothing tested it. Reordering those branches — or inserting one between them — would have silently retired the property and moved 3D bars onto `graph.bar.maxcount`.

The ordering is left as-is (it is correct today), now documented as load-bearing and pinned by a test.

### Also

`getGeomMaxCount(chartType)` is resolved once in `GraphGenerator.createElement` instead of once per element. It depends only on `chartType`, and leaving it in the loop would emit one warning per element.

## Tests

Two new files, 18 tests.

`GraphTypesTest` (new package directory — there was no test for this class):
- a numeric value is used as-is, and a negative value is preserved
- six non-numeric forms (`abc`, `` , ` `, `100,000`, `1e5`, `100000px`) plus null fall back instead of throwing
- exactly one WARN, at WARN level, naming both the property and the offending value
- no usable default degrades to `-1` (no limit)
- both 3D bar types resolve `graph.3dbar.maxcount` while flat bars resolve `graph.bar.maxcount` — the regression lock for finding 3

`GraphElementTruncationMessageTest`:
- truncation is reported at both `500` and `100001` (the latter fails against the old code)
- no message when nothing is truncated
- repeated calls raise a single message, asserted through `getUserMessages()` rather than `getUserMessage()` — the latter reduces with `UserMessage.merge`, which drops text already contained in the accumulation and would collapse duplicates on its own, making the assertion unfalsifiable

Verified with `./mvnw test -pl core` over `inetsoft.graph.**`, `inetsoft.report.composition.graph.**` and `inetsoft.uql.viewsheet.graph.**`: 126 tests, 0 failures (the single skip is pre-existing in `GraphRenderTest`).

## Out of scope

Most other `graph.*` properties parse unguarded the same way — `graph.subgraph.maxcount` (`RectCoord`), `graph.legend.maxcount` (`Legend`), `graph.axislabel.maxcount` (`CategoricalScale`, `TimeScale`), `graph.textlayout.maxcount` (`TextLayoutManager`), `graph.textlayout.maxstep` (`FreeHelper`), `graph.maxarea` (`VGraph`), `graph.pie.min.ratio` (`PolarCoord`), `graph.max.nodes` (`RelationElement`). `EGraph` is the one already-guarded site.

Deliberately not included here to keep the change scoped. `FreeHelper` is the one worth prioritising in a follow-up: its parse is in a static initialiser, so a bad value yields `ExceptionInInitializerError` rather than a render failure. A shared numeric accessor on `GTool` would fold all of them plus `GraphTypes` into one place; none exists today.

### DIFF
--- a/core/src/main/java/inetsoft/graph/element/GraphElement.java
+++ b/core/src/main/java/inetsoft/graph/element/GraphElement.java
@@ -772,16 +772,16 @@ public abstract class GraphElement extends Graphable {
             String msg = GTool.getString("viewer.viewsheet.chart.shapeCountMax", mcount) +
                ": " + Arrays.toString(elem.getDims()) + "," + Arrays.toString(elem.getVars());
 
-            // showing a warning to end user is more distractive than
-            // helpful. if max is very large, when the max is reached, the graph
-            // is generally so over crowed that any truncation is unlikely
-            // noticeable or significant
-            if(mcount > 100000) {
-               LOG.debug(msg);
-            }
-            else {
-               CoreTool.addUserMessage(msg);
-            }
+            // truncation is always reported. this used to be suppressed above a fixed
+            // threshold of 100000, on the reasoning that a graph that crowded is past the
+            // point where truncation is noticeable. but the shipped graph.point.maxcount
+            // (10000000) sits above that threshold, so the suppression hid row loss at a
+            // shipped default, on the one chart type most likely to be plotting a large row
+            // count -- a reader could not tell whether they were seeing all their data.
+            // CoreTool.addUserMessage de-duplicates, so the repeated calls made while
+            // building geometry still raise a single message.
+            LOG.debug(msg);
+            CoreTool.addUserMessage(msg);
 
             rcount = mcount;
          }
--- a/core/src/main/java/inetsoft/report/composition/graph/GraphGenerator.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/GraphGenerator.java
@@ -3114,7 +3114,10 @@ public abstract class GraphGenerator {
 
       alls = createElement0(chartType, names, xname, false, alls);
 
-      // set geometry max count
+      // set geometry max count. resolved once instead of per element: it only depends on
+      // chartType, and a misconfigured graph.*.maxcount warns on each call.
+      int mcount = GraphTypes.getGeomMaxCount(chartType);
+
       for(int i = 0; i < alls.size(); i++) {
          GraphElement elem = (GraphElement) alls.get(i);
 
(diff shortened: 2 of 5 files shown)


jshobe-inetsoft (approved):

PR #4784, opened 2026-08-25
Bug #76288, surface MV dimension overflow at query time

ngovus-inetsoft:
## Problem

`MV.checkValidity(GroupRef[])` existed to refuse a grouping on a dimension whose dictionary overflowed `mv.dim.max.size`. **Nothing called it** — it was the only occurrence of that overload in community or enterprise, and `git log -S` shows it was never wired up.

The consequence is concrete. `XDimDictionary.setOverflow(true)` clears the distinct-value set, and `indexOf(Object, int row)` then returns *the row position* instead of a value index:

```java
if(overflow) {
   // if overflow, the index of a value is it's row position
   // ... But since we don't allow grouping and filtering, ...
   return row - baseRow;
}
```

So grouping on an overflowed column makes every row its own group. The comment says "we don't allow grouping and filtering" — but nothing enforced it. The only notice anyone got was a `LOG.warn` in `MVCreatorUtil`, written when the MV was built, possibly weeks before someone builds a dashboard that trips over it.

Separately, `MVBuilder` compared `mv.max.block` against `nextb` — an absolute index into the source table — not the length of the block being built. The allowance was therefore `mv.max.block - r`, shrinking as the build advanced; once the cursor passed `mv.max.block` the loop body never ran, so the distinct-count protection stopped and the "caused MV block to exceed maximum size" warning fired for every remaining block. The name, the message, and the comparison could not all three be right.

## Approach

**Warn, don't fail.** The dashboard still renders; the operator is told which column is affected and that `mv.dim.max.size` must be raised **and the MV regenerated** (the flag is baked into the `.mv` file). This reuses `Tool.addUserMessage`, the channel `MVAssetQuery` already uses a few lines above for the STRING-column aggregate warning.

Throwing was considered and rejected: `MVExecutionException.recreate` defaults to `true`, so with `mv.ondemand` enabled it would trigger `recreateMVOnDemand()` — a futile rebuild, since the dictionary overflows again.

Detail queries are excluded — overflow deliberately retains the raw values so detail display stays correct. Group resolution mirrors `MVQueryBuilder.transform` (alias rewrite for WS MVs, null-expression filtering, and its `isDistinct`/`aggregate.down` detail determination), done on copies since the refs are shared with the assembly.

## A note on `isOverflow()`

Answering the question at all requires reading each dictionary off storage: `XDimDictionary.read()` is a *delayed* read that leaves `overflow` unassigned until `read0()` runs, so `isOverflow()` reports a stale `false` on a cold dictionary.

I deliberately did **not** change `isOverflow()` itself. It is also read by `MVBuilder.hasNext()` on the incremental path, where making it truthful re-activates a branch that has never run and would alter the on-disk index layout of appended blocks. Instead this adds `checkOverflow()`, used only by the new path, and memoizes the per-column result on the `MV` so the cost is not repaid on every query.

**That leaves a known latent bug**: the incremental build still reads the stale `isOverflow()`. It needs its own change with incremental-MV coverage rather than a drive-by fix here.

## Also fixed

- The dead method dereferenced `palette[idx]` *before* testing `idx >= 0`, so it would have thrown `ArrayIndexOutOfBoundsException` on a group column absent from the MV.
- The `mv.max.block` warning no longer fires when the table simply runs out of rows (previously the last block of every distinct-count MV warned), and it names the break column instead of printing the data value in row 0. `"disinct"` → `"distinct"`.

## Risk

This **re-enables protection that is currently off** past row `mv.max.block`, so late blocks can now extend toward the cap (default `preferred * 10` = 10M rows) instead of stopping at `preferred`. That is the intended behavior and the first block could already reach that size, but it now applies to every block — worth watching peak heap when building a large distinct-count MV with a low-cardinality break column.

## Testing

178 MV tests pass; full reactor build green. Three new test files. Every behavioral change has a test that was **mutation-verified** — confirmed to fail when the fix is reverted:

| Change | Fails without the fix as |
|---|---|
| `checkOverflow()` forces the delayed read | `expected: <true> but was: <false>` |
| `isOverflow()` stays a plain field read | asserts it does *not* force the read |
| memo invalidated on palette change | `expected: <[City]> but was: <[]>` |
| bounds check before `palette[idx]` | `ArrayIndexOutOfBoundsException: Index -1` |
| per-column `break` | `<[City, City, City]>` |

There was previously no coverage of `MVBuilder` block sizing or dictionary overflow at all.

### Manual verification

1. Set `mv.dim.max.size` low (e.g. `100`) and **restart** — `XDimDictionary.MAX_SIZE` is statically cached.
2. Build an MV over a worksheet with a high-cardinality string column; confirm the build warning in the log.
3. Group a chart or crosstab on that column: the dashboard still renders **and** the warning names the column.
4. Show Details over the same MV: **no** warning, correct raw values.

## Not in scope

Filtering on an overflowed dimension is also broken (`fixFilter` → `indexOf(value, -1)` yields a garbage index) and still warns nothing. The original message claimed to cover "grouping and filtering"; this change covers grouping.

### DIFF
--- a/core/src/main/java/inetsoft/mv/MVAssetQuery.java
+++ b/core/src/main/java/inetsoft/mv/MVAssetQuery.java
@@ -18,7 +18,10 @@
 package inetsoft.mv;
 
 import inetsoft.uql.erm.vpm.VpmProcessor;
+import inetsoft.mv.data.MV;
 import inetsoft.mv.data.MVQueryBuilder;
+import inetsoft.mv.data.MVStorage;
+import inetsoft.sree.internal.SUtil;
 import inetsoft.report.TableLens;
 import inetsoft.report.composition.QueryTreeModel.QueryNode;
 import inetsoft.report.composition.execution.*;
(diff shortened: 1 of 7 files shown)


jshobe-inetsoft (approved):

PR #4793, opened 2026-08-26
docs(ws/table): correct the expressionColumns support rules in comments

fredaan:
The comment above `applyExpressionColumns` in `buildPhysicalTable` said expression columns are "only meaningful on non-aggregated mirror tables" and that passing them there "logs a warning". Neither is true — a physical table does apply them, and nothing logs a warning. The same claim was repeated on the `WorksheetTable.expressionColumns` field javadoc, which is the text an API caller reads.

This restates the real rule at each build site, and says what to do instead where it isn't supported:

| tableType | expressionColumns | where documented |
|---|---|---|
| `physical table` | applied | `WorksheetTableService.buildPhysicalTable` |
| `mirror table` (no aggregation) | applied | `buildMirrorTable` |
| aggregating table | dropped — aggregate first, then derive on a mirror over it | `buildMirrorTable` |
| `relational join table` | not applied by design — derive on a mirror over the join | new `buildJoinTable` javadoc |
| `sql query table` | not applied by design — write it into `sqlExpression` | `buildSqlTable` javadoc |
| `tabular table` | rejected with an error (unchanged) | `buildTabularTable` |

Comments and javadoc only — no behavior change, no logic touched.


jshobe-inetsoft (approved):

PR #4794, opened 2026-08-26
Let node-color and node-size name a relation chart's node legend

lizhepu:
On a relation chart, the only names that *bind* a node aesthetic could not *address* its legend — and the name that could was reported as holding nothing. Found while covering the relation-chart branch of L4 in the Composer plugin test plan, the one branch of legend resolution that had no live coverage.

## The contradiction, as a caller meets it

On a network chart with `node-color` bound to `Customer:Region`:

```
get_chart_aesthetics   ->  "color":      { "field": null }
                          "node-color": { "field": "Customer:Region" }

list_chart_elements    ->  { "index": 0, "channel": "color", "field": "Customer:Region" }

hide target=node-color ->  refused: "This chart has no 'node-color' legend"
hide target=color      ->  accepted
```

So the documented route — read the aesthetics, name the channel the field is on — is refused, while the name that works is reported as empty. `set_aesthetic_field` accepts `node-color` and nothing else for this channel, so the caller has no way to learn the name that resolves.

## Root cause: the read dropped a distinction the model already carries

`LegendTarget` has a `nodeAesthetic` flag, and `legendFields` has always sent it on the hide event. But `channel()` returned `aestheticType` lower-cased, and `aestheticType` is plain `Color`/`Size` for a node legend and its edge counterpart alike — so the flag was discarded on the way out. `channel()` now prefixes `node-`, which fixes `list_chart_elements` and the refusal messages together, since both read it.

## Resolution narrows, it does not fold

`requireLegendField` strips the `node-` prefix to find the family, then keeps only legends whose `nodeAesthetic` is set.

Folding `node-color` onto `color` was the first attempt and is wrong: on a chart with an edge-colour legend and no node aesthetic at all, it resolves `node-color` to the **edge** legend — exactly the mis-resolution this method exists to prevent, and the shape of a bug this family has already shipped once. Narrowing also makes the both-bound case unambiguous, where folding would have refused it: the node flag says precisely which legend was meant.

Plain `color`/`size` stay permissive and match either, so a chart carrying only a node legend still resolves them. The resulting behaviour:

| chart has | `color` | `node-color` |
|---|---|---|
| only an edge colour legend | resolves to it | refused by name |
| only a node colour legend | resolves to it | resolves to it |
| both | ambiguous, refused with both fields named | resolves to the node one |

Resolution only selects *which* legend; the event is then built from the resolved target's field, `aestheticType` and `nodeAesthetic`, so the name the caller typed never reaches it. Hiding via `node-color`, via `color`, or via the field name produce an identical event.

## Live evidence behind it

On a network chart (`Customer:Region` → `Product:Category`, `Sum(Order:Paid)` on size, `Customer:Region` on node-color):

- hiding by field name `Customer:Region` hid **only** the node legend, the size legend surviving — so the node-aesthetic path itself resolves correctly and the destructive "hides them all" failure does not occur here
- with an edge colour bound as well, the two legends came back as `color` and `color`, told apart only by their fields
- naming `color` in that state was correctly refused as ambiguous, listing all three legends

## Verified live

After a rebuild and restart, on a rebuilt `Chart3`:

| check | result |
|---|---|
| `list_chart_elements` channel | `node-color`, agreeing with `get_chart_aesthetics` |
| hide `target: "node-color"` | node legend alone, size legend surviving (image-confirmed) |
| with an edge colour also bound, `node-color` | resolves **unambiguously** to the node legend |
| same state, plain `color` | refused as ambiguous |
| node aesthetic cleared, `node-color` | **refused by name**, the edge legend untouched |
| only one colour legend, plain `color` | still resolves — permissive path intact |

The refusal messages improved via the same `channel()` change: the ambiguity refusal now reads `Customer:Region (node-color), Customer:State (color), Sum(Order:Paid) (size)` where it used to print `(color), (color), (size)` and give the reader no way to tell which was which.

**The last two checks are what justify narrowing over folding.** Both are ordinary paths, and the folded version fails both — refusing the unambiguous case, and silently hiding an edge legend for a caller asking about a node one.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/ChartRegionResolver.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/ChartRegionResolver.java
@@ -333,7 +333,6 @@ final class ChartRegionResolver {
     * reports the field on the <b>shape</b> channel, while a line chart renders that legend as
     * <b>Line</b> — so a caller who read the aesthetics and said "shape" was naming this legend
     * correctly and being refused for it.
-    *
     */
    private static String channelFamily(String channel) {
       String name = channel == null ? "" : channel.trim().toLowerCase();
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/ChartRegionResolver.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/ChartRegionResolver.java
@@ -114,9 +114,18 @@ final class ChartRegionResolver {
        * not to say the two always agree on <em>which</em> channel: a line chart renders the field
        * on its shape channel as a {@code line} legend. Empty for a legend with no describable
        * frame, which is also a legend with no field, so neither can be named.
+       *
+       * <p><b>A node aesthetic is reported as {@code node-color}/{@code node-size}</b>, the names
+       * that bind it. The underlying {@code aestheticType} is plain {@code Color}/{@code Size} for
+       * both a relation chart's node legend and its edge legend, so reporting that raw dropped a
+       * distinction this record already carries: {@code get_chart_aesthetics} put the field on
+       * {@code node-color} while this said {@code color} and reported {@code color} as empty, so
+       * the only name the reader was given for the legend was one that did not address it.
        */
       String channel() {
-         return aestheticType == null ? "" : aestheticType.toLowerCase();
+         String name = aestheticType == null ? "" : aestheticType.toLowerCase();
+
+         return nodeAesthetic && !name.isEmpty() ? "node-" + name : name;
       }
    }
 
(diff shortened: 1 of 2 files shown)


lizhepu:
Both points addressed in f5302b2.

**The test gap** — added four cases to `ChartRegionResolverTest`'s `requireLegendField` section, against a relation chart's two colour legends (node + edge), covering the list from the review: `node-color` resolves to the node legend when both are present (and `channel()` reports `node-color` for it, `color` for the edge one, so the name the reader is given is the name that resolves); `node-color` refused when the chart's only colour legend is the edge one — the mis-resolution the narrowing exists to prevent; `node-color`, `NodeColor` and plain `color` all reaching the legend when it is the only colour one; and plain `color` staying ambiguous when both are bound, refused with `node-color` offered as the name that does resolve it.

Checked they pin the branch rather than passing by accident: run against the pre-fix source, three of the four fail (`node-color` refused as an unknown channel, `channel()` reporting plain `color`). The fourth was refused before as well, but for the wrong reason. `ChartRegionResolverTest` is 29/29 green with the change — which also answers the "not run" note, at least for this test class.

**The stray blank javadoc line** before `channelFamily`'s closing delimiter — removed.

fredaan (approved):

PR #4795, opened 2026-08-26
Collapse the three tabular table contracts into one

Nelson-cao:
Building a tabular table went through one of three contracts, chosen by `tabularSource.targetKind`: an endpoint name plus its parameters, a file path plus its parsing options, or the query bean's own properties. The first two were hand-written routines that knew in advance which properties mattered, which is what limited them to two kinds of connector out of the eighty-nine StyleBI ships — a document store, a cloud analytics API and a search index have neither an endpoint nor a file, and no amount of either contract binds one.

Those routines turn out to be almost entirely re-implementations of what the connector annotations already declare. Candidate values come from `tagsMethod`; the order properties must be set in comes from `dependsOn`; a write that silently no-ops is caught by reading the property back. So `targetKind` and both special contracts are gone, and `queryParams` — keyed by the connector's own property names — is the only way a tabular table is addressed. What the deleted routines did is now three general capabilities in `TabularQueryContractSupport`, and the properties the old wire had promoted to fields of their own (`jsonPath`, `expanded`, `lookup` and the rest) are ordinary `queryParams` entries, because that is all they ever were.

## Composites are decided at request time, not by type

Composite parameters are filled by name into the skeleton the connector itself builds, and which composites can be filled is decided by whether that skeleton exists. Type cannot answer it: `ODataQuery.functionParameters` and `EndpointJsonQuery.additionalParameters` are both `HttpParameter[]`, and only the first has a skeleton to fill. Deciding at request time also means no class is named anywhere, so a connector whose composite resolves once its prerequisites are set works without further change.

## Two checks survive

Each one's absence turns a refusal into a silent metered failure rather than an error, so neither generalises away:

- a POST endpoint is refused — its body comes from a template only the connector's own dialog populates
- a custom lookup URL missing its placeholder is refused before any request goes out, otherwise every row requests the same URL

## The contract is published as a JSON Schema

Generated from the same extraction, so a caller reads a format it already understands rather than a shape specific to this endpoint. `dependencyMatrix` maps onto `allOf`/`if`/`then`, which sinks conditional parameters into branches and leaves the top level as the minimal set.

The root therefore must use `unevaluatedProperties`, not `additionalProperties`: the latter sees only its sibling `properties` map, so with the conditional parameters in branches it would reject every correctly-formed paginated request. Measured against ajv 8.20.0, along with the companion finding that a draft-07 validator ignores `unevaluatedProperties` entirely and enforces nothing — which is why the `$schema` line is mandatory: declaring the dialect turns that consumer mistake into a compile error instead of a validator that silently passes everything.

Two facts survive nowhere else in the projection — which property names a file, which selects a sheet — and are stamped as `format: "file-path"` / `format: "file-sheet"`. The ambiguous multi-candidate case refuses rather than guessing, since a wrong sheet property builds a table over the wrong sheet and reports success.

`GET /api/wiz/tabular/query-schema` answers a `TabularQueryContract` of exactly two fields. The extractor's own working view (the flat parameter list, the dependency matrix, the `@View` notes) is input to the schema builder and is not published: the conditional structure is already expressed as `allOf`/`if`/`then` and the notes are folded into the root description, so a second flat view of one contract would only invite a consumer to loop over the easier one and miss the branches.

## maxRows is optional

Supplied, it is persisted and bounds every future render of the table. Absent, only the initial probe is bounded, through `HINT_MAX_ROWS`, and nothing is written to the query. A cap invented by the server and then persisted would silently truncate the table, so every aggregate over it would compute on a fraction of the data with nothing saying so.

## Scope

The composer `add_table` path is untouched. It is a different entry point with a different wire, `TabularEndpointBindingSupport` stays for it, and the two contracts are now independent rather than branches of one routine.

## Testing

328 tests pass, including the 23 restored `TabularEndpointBindingSupportTest` cases covering the untouched composer path on their own fixture.

Not yet verified: `resolveTags`'s 5s timeout and 200-candidate cap are reasoned, not measured against a live METADATA-class connector. `resolveTags` is the only source of candidate values for that connector class on the one path that reaches them, so a timeout that trips makes the connector unusable rather than slower — worth measuring before anyone depends on it.


Nelson-cao:
Thanks — four of the five are fixed in `ec3e59fde`. Taking them in order:

**1. Unbounded thread creation — fixed, and you read the failure mode correctly.** `newCachedThreadPool` had no maximum, and since `f.cancel(true)` cannot stop a reflective call blocked on non-interruptible I/O, one hanging connector really did mean one leaked thread per request on a READ-reachable GET.

Now an eight-worker `ThreadPoolExecutor` over a `SynchronousQueue` with `AbortPolicy`. A hanging connector costs at most eight stuck threads, after which `RejectedExecutionException` returns `null` and the caller sees `x-valueSource: "unavailable"` — the degradation it already knows how to handle, since a timeout produces the same answer. Deliberately not a buffered queue: waiting for capacity would spend a budget the caller expects to expire in five seconds.

While there, `InterruptedException` no longer shares a catch with `TimeoutException`. It arrives on the request thread, so the interrupt is now restored rather than swallowed.

**2. Stale `{@link}` — fixed.** Points at `TabularQueryContractSupport#applyQueryContract`, which is where the read-back idiom lives now.

**3. Wire break — intentional, and the other half is not in this PR.**

Confirming explicitly, since you asked for it: this is a deliberate breaking change with no shim. The module is pre-release and backward compatibility is explicitly not a constraint on it, which is why `target`'s old `@JsonAlias("endpoint")` shim was dropped rather than extended.

The rollout is two halves and this is one of them. The `wiz-services` side — the only caller of `/api/wiz/ws/table` — is a separate change set, not yet opened, covering its `tabularSource` types, the `queryParams` normalisation, the two `format`-keyed property lookups replacing what used to read a flat parameter list, and the agent-facing guidance. **So this PR should not merge ahead of that one.** The composer `add_table` path is unaffected for the reason in the description: different entry point, different wire, `TabularEndpointBindingSupport` retained for it.

Your read on the failure shape is right and worth keeping on the record: `@JsonIgnoreProperties(ignoreUnknown = true)` means an un-updated caller is not rejected at deserialisation, it loses those fields and then fails on `"tabularSource.queryParams is required"`. Loud, attributable, and no silent success — which is the property that makes the coordinated rollout survivable if the two halves land out of order.

**4. `compositeElementRequired` asymmetry — fixed, and it was unintentional.** `NoSuchMethodException` → `null` was deliberate ("no such concept"); every other reflective failure → `FALSE` was not. Reporting "confirmed optional" for an unknown answer is exactly what could let the write path skip a genuinely required element without complaining. Both cases answer `null` now, which is the distinction the nullable `Boolean` exists to carry in the first place.

**5. `char` truncation — fixed.** A multi-character string is refused now instead of silently becoming its first character. You're right that no tabular connector property of that type is in use today, but the inconsistency was the problem: writing something the caller did not ask for and reporting success is the precise failure class this method exists to close, so it should not have had an exemption.

328 tests pass. Unchanged from the description: `resolveTags`'s 5s timeout and 200-candidate cap are still reasoned rather than measured against a live METADATA-class connector — bounding the pool contains the blast radius of a hang, but it does not tell us whether five seconds is enough for a real `$metadata` fetch.

fredaan (approved):

PR #4798, opened 2026-08-26
fix(binding): close six chart aesthetic frame defects found by live testing

NeverStopFMH:
Every defect here was reproduced through the agent tools against a running chart, with a control experiment, before being fixed. Each one had the same shape: the write reported success, the read agreed with it, and the chart drew something else.

## `line` and `texture` read and write the shape field's frame

The Composer has one **Shape** slot where the agent vocabulary has three channels. Its swatch edits `lineFrame` on a chart `GraphTypes.supportsLine` answers for, `textureFrame` where `supportsTexture` does, and `shapeFrame` otherwise — one slot, three frame families, picked by chart type (`shape-field-mc.getEditPaneId`).

So binding a dimension to shape on a line chart puts a `CategoricalLineModel` on the **shape** field, and `VSLineFrameStrategy.getAestheticRef` returns that ref precisely because its frame is a `LineFrame`. `VSFrameVisitor.createFrame` then prefers it over every field-less slot. Treating `line` and `texture` as channels that never have a field sent both the read and the write to the per-measure slot, which the renderer ignores in that state.

**Live repro** — line chart, dimension on shape:
```
set_visual_frame channel="line" {type: "static", line: "large dash"}   → ok
render                                                                → byte-for-byte unchanged
clear_aesthetic_field channel="shape"                                 → the same stored value renders at once
```
Confirmed again on `texture` with a bar chart.

## A bound field decides which frame kinds its channel takes

`getEditPaneId` picks the editor from the bound field: a dimension (or a discrete aggregate) opens the **Categorical** pane and nothing else, a measure opens the **Linear** one, and **Static** is offered only when nothing is bound. Anything else is a state the Composer cannot produce and the backend does not cope with.

**Live repro**: `{type: "static", color: "#FF0000"}` on a colour channel with a dimension bound rendered the ordinary categorical palette, no red anywhere. On the line channel the static value ended up sitting in one slot of the categorical `lines` array — neither honoured nor cleanly dropped.

## A field-less colour frame on the measures must be static

`GraphUtil.fixDuplicateColor` casts every aggregate's colour frame wrapper to `StaticColorFrameWrapper` with no `instanceof` in front of it. The cast is safe for the Composer because `getEditPaneId` falls back to `CombinedColor`/`StaticColor` when nothing is bound.

**Live repro** — bar chart, colour channel unbound:
```
set_visual_frame channel="color" {type: "categorical", colors: [...]}  → ok
get_viewsheet_image                                                    → 500
clear_aesthetic_field / set_aesthetic_field on it                      → 500
set_visual_frame channel="color" {type: "static", color: "..."}        → renders again
```
`{type: "gradient"}` did the same. The stack is `GraphUtil.fixDuplicateColor` → `ClassCastException`. Guarded on the agent path, and the three casts now skip what they cannot fix rather than stopping the chart.

## Share Colors is one setting under one name, and is left alone unless asked

The frame carries `useGlobal` and `shareColors` separately for historical reasons — `useGlobal` came first for the Color Mapping dialog, `shareColors` was added later for frame sharing, which is why `CategoricalColorFrameWrapper.parseContents` has a legacy default for one and not the other. The Composer has driven them from a single checkbox ever since; the dialog's own `toggleGlobal()` is left over from before that and is wired to nothing.

Exposing both offered a distinction the product does not have, so the vocabulary has one name. And a spec that does not mention it now leaves the checkbox where it is, as Apply does — forcing it off meant every colour change silently stopped a chart sharing its palette, with nothing in the request to say so.

**Verified live**: two charts colouring by the same column converge on one palette; a pin set on one reaches the other; `shareColors: false` restores independence.

## Share Colors no longer 500s, and no longer destroys the viewsheet's pins

`VSChartBindingFactory.applyColorsToViewsheet` runs only when the flag is set, and hands `frame.getField()` to `Viewsheet.setDimensionColors`, whose `getAttribute` does `column.indexOf(":")`. A frame built on the agent path is a bare `new CategoricalColorModel()` and carries no field, so every `shareColors: true` call against a bound colour channel was a hard NPE — a 500 with no edit applied.

It also rewrites the column's whole entry from `globalColorMaps`, and `setDimensionColors` drops the column's existing keys before putting the new ones, so shipping an empty array deleted the viewsheet's pins rather than leaving them alone. Both are carried forward now, and `Viewsheet` answers "none" for a null column instead of throwing.

## A mapping's pins go wherever the flag says the renderer reads them

`openColorMappingDialog` writes `globalColorMaps` when Share Colors is on and `colorMaps` when it is off; `ColorFrameModelFactory` reads whichever the flag selects. The agent path wrote `colorMaps` unconditionally and refused the combination outright — turning a supported and more useful pairing (pinning a value across the whole viewsheet) into an error.

Both directions are routed now, the read reports from the array the renderer actually uses, and a mapping merges with the column's existing pins rather than replacing them: `setDimensionColors` would otherwise unpin every value the request did not name.

---

## Also adds what the panes had and the tools did not

- **One measure at a time** — the Combined pane, which draws one editor per measure. Applies only where the channel renders per measure and has no field bound, which is exactly when that pane opens.
- **Reset to Default** — for the three frames that have a default: a categorical colour palette, a categorical shape/line/texture set, and the per-measure static colours. Everything else is refused by name, which is also where the Composer offers no reset.
- **Use Column Values as Colors** — accepted; the Composer only shows the checkbox where the deployment lists `ColorValueColorFrame` in `custom.chart.frames`.

The read reports per-measure frames when the measures disagree, so it stops describing a chart that is visibly drawing something else beside it. Colours in a mapping are reported in the one spelling the write side accepts — `VSChartBindingFactory.applyColorsToFrame` refills `globalColorMaps` with `Tool.colorToHTMLString`, which omits the `#`, while everything else uses `Tool.toString(Color)`.

## Testing

`inetsoft.web.wiz.binding.*` plus `GraphUtil*` and `PasteAestheticRefFirstBindFrameTest`: **472 tests, 0 failures**. Every defect above has a regression test carrying its live repro in the comment.

## Known issue, not addressed here

`colorValueFrame` is accepted and stored on the model but does not survive a round trip — the Composer's own checkbox has the same behaviour, so it is not specific to this path. Diagnosis is not settled and the fix is deliberately left out of this PR.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/GraphUtil.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/GraphUtil.java
@@ -1026,7 +1026,17 @@ public class GraphUtil {
                continue;
             }
 
-            temp = (StaticColorFrameWrapper) ((ChartAggregateRef) vref.get(i)).getColorFrameWrapper();
+            // Only a static per-measure colour has a default to fix. A measure carrying any other
+            // colour frame is drawn by that frame, so there is nothing here to give it and the
+            // cast this replaces was a ClassCastException that stopped the chart rendering
+            // entirely.
+            if(!(((ChartAggregateRef) vref.get(i)).getColorFrameWrapper()
+                    instanceof StaticColorFrameWrapper staticWrapper))
+            {
+               continue;
+            }
+
+            temp = staticWrapper;
             int idx = getColorIndex(usedColors, clrs, i);
             usedColors.add(clrs[idx]);
             temp.setDefaultColor(clrs[idx]);
@@ -1053,7 +1063,6 @@ public class GraphUtil {
             ChartAggregateRef aref = (ChartAggregateRef) arefs[i];
             boolean sameRef = Tool.equals(aref, movedRef);
             sameRef = sameRef && aref.equalsContent(movedRef);
-            frame = (StaticColorFrameWrapper) aref.getColorFrameWrapper();
 
             if(ref != null && aref.getFullName().equals(ref.getFullName())) {
                aggrIdx = i;
@@ -1068,6 +1077,16 @@ public class GraphUtil {
                continue;
             }
 
+            // Only a measure drawn by a static colour occupies one of the palette's colours. Any
+            // other frame draws its own, so it takes nothing out of the running -- and the cast
+            // this replaces was a ClassCastException that stopped the chart rendering entirely
+            // rather than merely picking a duplicate.
+            if(!(aref.getColorFrameWrapper() instanceof StaticColorFrameWrapper staticWrapper)) {
+               continue;
+            }
+
+            frame = staticWrapper;
+
             if(frame.getUserColor() != null) {
                usedColors.add(frame.getColor());
             }
@@ -1077,7 +1096,13 @@ public class GraphUtil {
          }
       }
 
-      frame = (StaticColorFrameWrapper) ref.getColorFrameWrapper();
+      // The measure being fixed is drawn by its own non-static frame, so it has no default colour
+      // to hand it and nothing to deduplicate.
+      if(!(ref.getColorFrameWrapper() instanceof StaticColorFrameWrapper target)) {
+         return;
+      }
+
+      frame = target;
 
       if(usedColors.isEmpty() && frame.getColor().equals(OLD_COLOR_PALETTE_DEFAULT_COLOR)) {
          return;
--- a/core/src/main/java/inetsoft/uql/viewsheet/Viewsheet.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/Viewsheet.java
@@ -3687,6 +3687,14 @@ public class Viewsheet extends AbstractSheet implements VSAssembly, VariableProv
     * @return a new map that contains the fixed mappings for the specified column
     */
    public Map<String, Color> getDimensionColors(String column) {
+      // A frame with no field has no column to key the fixed mappings by, so there is nothing to
+      // look up. Answered as "none" rather than left to NPE in getAttribute: every caller here
+      // reads the name off a VisualFrameModel, and one built outside the Composer's own factories
+      // may not carry it.
+      if(column == null) {
+         return new HashMap<>();
+      }
+
       // add delimiter at the end so that we don't match the wrong column
       final String columnNameKey = getAttribute(column, false) + COLUMN_DELIMITER;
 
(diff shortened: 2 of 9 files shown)


NeverStopFMH:
Thanks — the orphaned Javadoc was real. Fixed in fcf90a0a5: the block describing `requireStaticColorOnMeasures` had ended up immediately before `requireFrameTheFieldCanDrive`'s own comment, so only the second was attached and the method the first describes had none. Moved, no code change.

Two follow-ups worth recording here, both from the sibling PR's review and from live testing since.

**`colorValueFrame` was accepted when explicitly false** — 63a382440

The review on inetsoft-technology/stylebi-wiz#1852 found the tool-side half of this; the same check lives here. A categorical colour frame has to carry something a colour can be derived from — a palette, a mapping, or the column's own values — and `categoricalColor()` tested `colorValueFrame` for *presence* rather than for *true*. So `{type: "categorical", colorValueFrame: false}`, which says "read the palette" while supplying no palette, passed the check that exists to refuse exactly that shape. Both halves are fixed and tested.

**A pre-existing defect found while re-testing, deliberately not in this PR**

`set_date_comparison`'s frame applies its *type* and drops its *values*. Live, against a chart with four years of data:

```
set_date_comparison frame={type: "static", color: "#E01B24"}   → ok
get_date_comparison                                            → {type: "static", color: "#518db9"}
```

`#518db9` is the default palette's first colour. A categorical frame behaves the same way: two colours in, the full 40-entry default palette back.

The cause is `DateComparisonPaneModel.toDateComparisonInfo()`:

```java
info.setDcColorFrameWrapper(VisualFrameWrapper.wrap(visualFrameModel.createVisualFrame()));
```

`createVisualFrame()` builds a bare default of the right class and copies none of the model's values — which is precisely what #4760 documented and fixed in `AestheticRefModelFactory.pasteAestheticRef`. This is the same bug in a second place, and it predates both PRs. It is not agent-specific either: `toDateComparisonInfo()` is shared with the Composer's own date-comparison dialog, so the interactive path should lose the colour the same way.

The fix is the one #4760 already applied — route through `VisualFrameModelFactoryService.updateVisualFrameWrapper`, which copies the values — but `DateComparisonPaneModel` is a plain model with no access to that service, so it needs threading through the method and its callers on both paths. Left out of this PR rather than bolted on, and worth confirming against the Composer dialog first so the shared-code change has evidence from both callers.

**Also known and not addressed here**: `colorValueFrame` is stored on the model but does not survive a round trip. The Composer's own checkbox behaves identically, so it is not specific to this path; the diagnosis is not settled and the fix is deliberately excluded.

jshobe-inetsoft (approved):

PR #4799, opened 2026-08-26
fix(admin-chat): narrow AdminPropertyCatalog.isSecret false positives

larryliang-inetsoft:
## Summary
- `AdminPropertyCatalog.isSecret`'s name-shape test masked three properties that hold no credential value: `enable.changepassword` (matches only on the substring "password"; it's a boolean flag), `sso.rsa.public.key` (the PUBLIC half of the SSO keypair — its private counterpart, `sso.rsa.private.key`, correctly still matches), and `google.maps.key` (a Google Maps API key, restricted by HTTP referrer rather than treated as a bearer credential, per its plain `SreeEnv` reader/writer).
- Each was individually verified against its actual reader/writer before being added to a new `CONFIRMED_NOT_SECRET` exception set — same verification standard `ENCRYPTED_CREDENTIALS` already uses — rather than loosening the shared name-shape pattern and risking a real secret slipping through.
- Follows the same convention Redmine #76006 (`2d30488d0`) used for the opposite direction (missed true positives): individually-verified names, documented reasoning in the `isSecret` javadoc, regression tests both ways.

Design source: `docs/superpowers/plans/2026-08-25-admin-plugin-master-plan.md` §4 Track A item 4, and the matching gap note in `plugin/admin/README.md` ("Known gaps and follow-ups" #4) in the `stylebi-wiz` repo.

## Test plan
- [x] `AdminPropertyCatalogTest` — 27/27 passing (`./mvnw -pl core -Dtest=AdminPropertyCatalogTest test`), including two new tests: false positives return `false` (with `sso.rsa.private.key` re-confirmed `true`), and case-insensitivity of the exception list.
- [x] Existing true-positive regression test for #76006 re-confirmed still passing (with `enable.changepassword` moved out of that list into the new false-positive test, since it's no longer expected to be withheld).

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminPropertyCatalog.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminPropertyCatalog.java
@@ -208,10 +208,29 @@ public class AdminPropertyCatalog {
     * secret value read through this path leaves the host in a way the ordinary EM properties page
     * never does.
     *
-    * <p>Two independent tests, unioned. A name matches case-insensitively when it contains
-    * {@code password}, {@code secret} or {@code credential}, ends with {@code .key}, or starts
-    * with {@code license.}; and {@link #isEncryptedCredential} matches the properties whose
-    * <b>writer</b> encrypts them.
+    * <p>Two independent tests, unioned, then a name-shape exception list subtracted. A name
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4800, opened 2026-08-26
Bug #76290, log an unparsable ai.snapshot.count and make the strict-pane flag settable by env var

ngovus-inetsoft:
The two findings from the #76290 report that concern code living only on this branch. The other two are fixed against `main` in #4786.

> **Base is `stylebi-wiz-main-rebase`, not `main`.** The report was verified at `935c85a58`, which is on this branch. Neither `deleteRedundantAiSnapshotFiles` nor `wiz.agent.script.require-script-pane` exists on `main`, so neither could be fixed there.

## 1. `ai.snapshot.count` discarded an unparsable value without a word

`deleteRedundantAiSnapshotFiles` parsed the property inside a bare `catch(Exception ignore) {}` with no log. A typo silently reverted retention to 10, and how many AI snapshot files survive cleanup is not something an operator would notice.

The parse moves into `getAiSnapshotCount()`, following `ScheduleTask.getTaskTimeout` — the shape the report cites as the one that works:

- a named `DEFAULT_AI_SNAPSHOT_COUNT = 10` constant, so the javadoc's "default 10" cannot drift from the code
- `NumberFormatException` specifically, not bare `Exception`
- `LOG.warn` naming both the property and the offending value

**Absent and blank both stay silent.** `PropertiesEngine.setProperty` removes a property only when the value is `null`, so clearing the field in EM rather than deleting the row stores `""` — and `parseInt("")` throws, which would have made this warning fire on *every backup*. Warning about a field the operator emptied on purpose is exactly the recurring noise the warning exists to avoid, so only a value that was set and cannot be read warns. (Caught in review; the guard is `null || isBlank()`.)

## 2. A hyphenated property had no reachable environment-variable form

`wiz.agent.script.require-script-pane` was settable from Settings > All Properties but by no environment variable. `StorageInitializer.setProperties` builds the name via `substring(12).toLowerCase().replace('_', '.')` — no escape mechanism produces a hyphen, so no `INETSOFTENV_*` value could name it. The reverse mapping in `InitStorageJobResource` and the AWS/CDK templates is equally hyphen-blind.

Renamed to `wiz.agent.script.require.script.pane`, which `INETSOFTENV_WIZ_AGENT_SCRIPT_REQUIRE_SCRIPT_PANE` reaches. This is the cheaper of the two options the report offered — teaching the transform a hyphen convention would have to be matched across the Java bootstrap, the k8s operator, and the CDK templates, and would risk existing deployments' variables.

The name is centralized as `PaneScopeService.STRICT_FLAG`, so reads are unaffected; its javadoc now records *why* the name is dot-only, so a hyphen is not reintroduced. Since this branch is unmerged, no released deployment can have the old name configured, so no backward-compatible alias is needed.

**One trap worth flagging for review.** Four tests assert `!ex.getMessage().contains("require-script-pane")` — that a given refusal is about the formula editor, *not* the strict flag. After a rename those keep passing for the wrong reason: a substring that exists nowhere can never appear. They are updated to the new name; left alone they would have silently stopped testing anything.

## Verification

126/126 tests pass in the affected slice (`DataSpaceSettingsService`, `PaneScopeService`, `ScriptTarget`, `ViewsheetAgentController`, `WorksheetScriptService`, `WorksheetScriptController`), re-run against the current branch tip after rebasing onto `fbe33041d`.

Both new behaviors are pinned by tests that were confirmed to fail without the fix:

- removing the `LOG.warn` fails the warning test (`expected: <1> but was: <0>`)
- reverting the guard to `null`-only fails the `""` and `"   "` cases while `null` still passes

Also verified: zero remaining `require-script-pane` references in either repo (java/ts/html/yaml/json/md/properties); no `AdminPropertyCatalog` registration to update; the `INETSOFTENV_` transform does produce the new name.

**Not covered:** the Docker end-to-end check that the renamed property actually arrives via `INETSOFTENV_*` and shows up in Settings > All Properties. The transform is verified in isolation, but since the whole point of finding 2 is that the env-var route was unreachable, that round-trip is worth confirming before merge.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/general/DataSpaceSettingsService.java
+++ b/core/src/main/java/inetsoft/web/admin/general/DataSpaceSettingsService.java
@@ -214,21 +214,14 @@ public class DataSpaceSettingsService extends BackupSupport {
     * out of scope for this change - see the final-review deferred-findings note. This method
     * keeps exactly the configured count.
     *
-    * <p>Retention is controlled by {@code ai.snapshot.count}, defaulting to 10 when absent or
-    * unparsable; a value below 1 disables pruning entirely, matching {@code
-    * deleteRedundantBackupFiles}'s convention for {@code asset.backup.count}.
+    * <p>Retention is controlled by {@code ai.snapshot.count}, defaulting to
+    * {@link #DEFAULT_AI_SNAPSHOT_COUNT} when absent or unparsable; a value below 1 disables
+    * pruning entirely, matching {@code deleteRedundantBackupFiles}'s convention for {@code
+    * asset.backup.count}. An unparsable value is logged, because the effect -- how many
+    * snapshots survive cleanup -- is not something an operator would otherwise notice.
     */
    private void deleteRedundantAiSnapshotFiles() {
-      String countProp = SreeEnv.getProperty("ai.snapshot.count");
-      int count = 10;
-
-      if(countProp != null) {
-         try {
-            count = Integer.parseInt(countProp.trim());
-         }
-         catch(Exception ignore) {
-         }
-      }
+      int count = getAiSnapshotCount();
 
       if(count < 1) {
          return;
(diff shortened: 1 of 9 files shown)


jshobe-inetsoft (approved):

PR #4802, opened 2026-08-26
Bug #76084, add edit_date_range_column/edit_numeric_range_column; Bug #75942, custom bucket labels

jkidacki-inetsoft:
## Summary

- **Bug #76084**: `add_date_range_column`/`add_numeric_range_column` had no `edit_*` counterpart, so the wiz agent asked to change an existing range column's grouping option/boundaries fell back to deleting and re-adding it — orphaning anything already bound to the old column. Adds `editDateRangeColumn`/`editNumericRangeColumn` to `WorksheetEditService` and wires `edit_date_range_column`/`edit_numeric_range_column` into `WorksheetAgentController`'s op switch.
- **Bug #75942**: `ValueRangeInfo` already fully supports custom bucket labels (`getLabels()`/`setLabels`, consumed by `NumericRangeRef`'s expression generators) — the same feature the Composer's own Range Column dialog exposes — but neither `add_numeric_range_column` nor the new `edit_numeric_range_column` ever called `setLabels()`. Both now accept an optional `labels` array, validated against `boundaries.length + 1` up front (the count the engine's expression generators index into with no bounds check).

## Root Cause

- #76084: no backend method or op existed for editing a range column in place.
- #75942: `setLabels()` was simply never wired up on the add path, despite full engine support.

## Fix

- `editDateRangeColumn` renames the column to match its new option (the name encodes the grouping level, e.g. `QuarterOfYear(Order Date)` → `MonthOfYear(Order Date)`), guarded against colliding with another already-named range column.
- `editNumericRangeColumn` updates boundaries in place; the column's name (`column + "_range"`) is unaffected.
- New `labels` param on both `add_numeric_range_column`/`edit_numeric_range_column`, with a shared `validateLabelCount` helper failing loud on a wrong count instead of letting `NumericRangeRef.getExpression()`/`getScriptExpression()` throw `ArrayIndexOutOfBoundsException` deep in expression generation.
- `EditRequest`'s `labels` field is appended at the end of the record (after `crosstab`), not inserted near `boundaries`, to avoid breaking the existing hand-written compatibility constructor's positional delegation.

## Test Plan

- `mvnw clean test-compile -pl core` — clean compile (a non-clean incremental build initially masked a broken positional `EditRequest(...)` call in a test file elsewhere; caught via `clean` and fixed).
- `mvnw test -pl core -Dtest=WorksheetEditServiceMutatorsTest,WorksheetAgentControllerTest,WorksheetReadServiceTest,GroupSpecDeserializationTest` — 221/221 pass, including new coverage for both bugs (rename-collision guard, label-count validation, label ordering verified against actual generated expression text).
- Manually verified end-to-end against a rebuilt `stylebi-wiz` composer-chat plugin: editing an existing range column's option/boundaries now updates in place, and custom bucket labels apply correctly.

Companion PR (stylebi-wiz plugin, exposes these ops as MCP tools): to follow.

Fixes Redmine #76084, #75942.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
@@ -66,7 +66,14 @@ import java.util.Map;
  *   <li>{@code add_rotate} — {@code name}, {@code source}</li>
  *   <li>{@code add_unpivot} — {@code name}, {@code source}, {@code headerColumns}</li>
  *   <li>{@code add_date_range_column} — {@code table}, {@code column}, {@code dateOption}</li>
- *   <li>{@code add_numeric_range_column} — {@code table}, {@code column}, {@code boundaries}</li>
+ *   <li>{@code add_numeric_range_column} — {@code table}, {@code column}, {@code boundaries},
+ *       optional {@code labels} (one more entry than {@code boundaries})</li>
+ *   <li>{@code edit_date_range_column} — {@code table}, {@code column} (the existing range
+ *       column's own name), {@code dateOption}; the column is renamed to match, since its name
+ *       encodes its option (see {@code add_date_range_column})</li>
+ *   <li>{@code edit_numeric_range_column} — {@code table}, {@code column} (the existing range
+ *       column's own name), {@code boundaries}, optional {@code labels} (replaces any existing
+ *       labels; omitted/empty clears them back to the default auto-generated range text)</li>
  *   <li>{@code edit_cell} — {@code table}, {@code row}, {@code col}, {@code value}</li>
  *   <li>{@code insert_row} — {@code table}, {@code index}</li>
  *   <li>{@code delete_row} — {@code table}, {@code index}</li>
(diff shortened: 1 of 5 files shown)


jkidacki-inetsoft:
Addressed the automated review's findings in b13138f:

1. **Stale hashCode on `editDateRangeColumn`'s in-place rename** (real bug, confirmed empirically) — renaming the wrapped `DateRangeRef` left the wrapping `ColumnRef`'s cached `hashCode()` stale, which let a later `add_date_range_column` producing the same new name silently create a duplicate-named column. Wrote a probe test that reproduced it (2 columns named identically), verified the suggested fix (`ColumnRef#setDataRef` to force cache invalidation) actually resolves it, then kept it as a permanent regression test.
2. **`editNumericRangeColumn` discarding `showBottomValue`/`showTopValue`/`inclusiveType`** — now mutates the existing `ValueRangeInfo` instead of replacing it, with a new test confirming those three survive a boundaries/labels-only edit while `labels` omission still clears pre-existing labels.
3. **Null base-ref fallback** — now fails loud with a `PairingException` instead of silently computing a wrong nested name.

224/224 tests pass (`mvnw clean test -pl core`).

ngovus-inetsoft (approved):

PR #4803, opened 2026-08-26
Bug #76278, enforce security.selfSignUp.enabled on the signup endpoints

ngovus-inetsoft:
## Problem

Turning self-signup off removed the *Create an Account* link from the login page but did not close the signup flow.

`SecurityEngine.isSelfSignupEnabled()` reads `security.selfSignUp.enabled`, but its only production readers were display or SSO paths:

- `LoginController:135-137` — the `selfSignUpEnabled` model attribute that renders the link
- `SecurityConfigController:276` — reports the value back to Enterprise Manager
- `StyleBIGoogleSSOFilter:280` (enterprise) — gates Google SSO auto-registration

`SignupController` had five mapped endpoints and no reference to the property at all, and `AbstractSecurityFilter.publicResources` lists `/signup.html`, `/signupDetail.html`, and `/signup/**` unconditionally. So with the toggle off, anyone who knew the URLs could still reach the flow.

The emailed verification code means this was not an unauthenticated account-creation hole. The defect is narrower: the property presents as an on/off switch for self-registration, and an operator who turns it off should get the door closed, not the sign hidden. Worth noting the property has **no declared default** anywhere and the read is a strict `"true".equals(...)`, so the effective default is `false` — the flow was reachable on a default install where the flag reads as off.

## Fix

Inject `SecurityEngine` into `SignupController` and gate all five endpoints on `isSelfSignupEnabled()` as the **first statement**, before any session or mail work:

| Endpoint | Refusal |
|---|---|
| `GET /signup.html` | redirect to `login.html` |
| `GET /signupDetail.html` | redirect to `login.html` |
| `GET /signup/withEmail` | `SecurityException` → 403 |
| `GET /signup/userDetail` | `SecurityException` → 403 |
| `GET /signup/resendEmailCode` | `SecurityException` → 403 |

`ControllerErrorHandler` is `@ControllerAdvice(basePackages = { "inetsoft.web.portal" })` and already maps `SecurityException` to 403, so no new catalog keys are needed. This is the same pattern `ChangePasswordDialogController:59,73` uses. `signup.js` already has an `error:` callback on all three JSON calls, so a 403 degrades gracefully — though in practice legitimate users never reach them with the feature off, because the pages redirect away first.

`RedirectView` is used explicitly rather than a `"redirect:"` view name: views resolve through Thymeleaf here and there is no existing `redirect:` prefix usage in the codebase, so this avoids depending on view-resolver prefix handling.

## Scope decisions

- **Gate is `isSelfSignupEnabled()` alone**, which already ANDs `isSecurityEnabled()`. `LoginController` ANDs `LicenseManager.isEnterprise()` and a not-a-tenant-server check on top, but only to decide whether to render the link. Enforcing those server-side would close signup in community edition and on tenant servers — a behavior change beyond this defect.
- **`publicResources` is unchanged.** The paths must stay public while the feature is on, so the filter cannot gate them without threading the property into `AbstractSecurityFilter` — more surface than this warrants. Closing the flow is the controller's job.
- With security disabled, `isSelfSignupEnabled()` returns false and the endpoints now refuse. Not a user-visible regression: with no authentication chain the flow already failed at `existAuthenticationChain()` with `signup.not.support`, and the link was already hidden. This converts a soft failure into a 403.

## Not fixed here (pre-existing, flagged so they aren't mistaken for regressions)

- `showSignUpDetailPage` NPEs when there is no session (`getSession(false)` then dereferenced). The new guard runs first, so a disabled-signup request no longer reaches it, but the NPE remains on the enabled path.
- `StyleBIGoogleSSOFilter:280` reads `SreeEnv.getBooleanProperty` directly instead of going through `SecurityEngine`, so unlike every other reader it skips the `isSecurityEnabled()` AND. Real inconsistency, enterprise-side, separately tested — worth its own issue.
- `signup.js:311` sends a `name=` param that `submitSignupDetail` does not declare (harmlessly ignored; the username is derived server-side from the session email).

## Testing

New `SelfSignupDisabled` nested group in `SignupControllerTest`: the two redirects, the three 403s, and `verifyNoInteractions(userSignupService)` on each — which is what proves the guard runs before any email is sent.

`./mvnw test -pl core -Dtest=SignupControllerTest` → **22 tests, 0 failures** (17 pre-existing + 5 new). The nine existing test methods that call the now-`throws` endpoints gained a `throws Exception` clause.

Not yet run: the Docker end-to-end checks (feature off → 302/403 and no mail; feature on → full flow still completes). The enabled-path regression check is worth doing before merge.

### DIFF
--- a/core/src/main/java/inetsoft/web/portal/controller/SignupController.java
+++ b/core/src/main/java/inetsoft/web/portal/controller/SignupController.java
@@ -21,6 +21,7 @@ import inetsoft.sree.internal.SUtil;
 import inetsoft.sree.portal.CustomThemesManager;
 import inetsoft.sree.portal.PortalThemesManager;
 import inetsoft.sree.security.*;
+import inetsoft.sree.security.SecurityException;
 import inetsoft.util.Catalog;
 import inetsoft.util.Tool;
 import inetsoft.web.portal.model.SignupResponseModel;
@@ -34,6 +35,7 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.*;
 import org.springframework.web.servlet.ModelAndView;
+import org.springframework.web.servlet.view.RedirectView;
 
 import java.util.Date;
 import java.util.Objects;
@@ -42,11 +44,22 @@ import java.util.Objects;
 public class SignupController {
    public SignupController(UserSignupService userSignupService,
                            CustomThemesManager customThemesManager,
-                           PortalThemesManager portalThemesManager)
+                           PortalThemesManager portalThemesManager,
+                           SecurityEngine securityEngine)
    {
       this.userSignupService = userSignupService;
       this.customThemesManager = customThemesManager;
       this.portalThemesManager = portalThemesManager;
+      this.securityEngine = securityEngine;
+   }
+
+   /**
+    * Checks whether self signup is enabled. The login page only hides the sign up link when this
+    * is off, so every signup endpoint has to enforce it as well; otherwise the flow stays reachable
+    * for anyone who knows the URLs.
+    */
+   private boolean isSelfSignupEnabled() {
+      return securityEngine.isSelfSignupEnabled();
    }
 
    /**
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4805, opened 2026-08-26
Bug #76280, replace the FORM license gate with a plainly-named property check

ngovus-inetsoft:
## Problem

`LicenseManager.isComponentAvailable(LicenseComponent.FORM)` consults no license. It returns the value of the `vs.form.enabled` property:

```java
if(Objects.requireNonNull(component) == LicenseComponent.FORM) {
   return !"false".equals(SreeEnv.getProperty("vs.form.enabled"));
}
```

Meanwhile `License.formLicensed()` — set from the parsed key by `LicenseParser` — had no reader anywhere, so a license granting or withholding FORM had no effect either way.

The class name misdirected diagnosis. Nine call sites gated real behavior on this: form tables, the Form option in the Table Properties dialog, submission and write-back, script access through three classes, and the toolbar Import button, which `SUtil.getVSToolBarElements` drops before its own `vs.import.button` setting is read. An operator whose Import button had vanished had every reason to check their license, and the license is not what removed it.

### Two distinct `LicenseComponent` types

Worth flagging, because it is easy to misread the call sites:

- `LicenseManager.LicenseComponent` — a **nested stub** with the single value `FORM("Form")`, whose `webExtName` field was never read. This is what the gate and all nine call sites used.
- `inetsoft.report.internal.license.LicenseComponent` — the real 12-value bitmask enum that `LicenseParser` populates from the key.

The gate never referenced the license-derived enum at all.

## Change

Replace the gate with `FormUtil.isFormEnabled()`, named for what it actually reads and placed alongside the existing `containsForm()`. The property expression and the `catch(Exception) -> false` guard are preserved verbatim, so **behavior is unchanged**.

Removed as dead:

- `LicenseManager.isComponentAvailable` and the nested `LicenseComponent` stub
- `License.formLicensed()`

`ViewComponent.requiredLicenses()` referenced the deleted nested enum **as a type** (invisible to a method-name grep — only a clean build surfaced it), so it is retyped to the real top-level `LicenseComponent`. That enum's value set is a strict superset of the stub's, so no previously-valid `view-components.json` becomes undeserializable.

`viewer.viewsheet.needFormLicense` is reworded to name the property instead of claiming a license is required. The key is kept — there are no localized catalogs, but customers may override it.

## Not changed, deliberately

Two adjacent issues found while doing this, both left alone as separate decisions:

1. **`requiredLicenses` is a second dead license gate.** It is populated from EM `@Secured` decorators (`web/gulp/em-metadata.js:45`), all three entries in the generated `view-components.json` are `[]`, and `ViewComponent.available()` is a hardcoded `return true;` that never consults it. Removing it would break Jackson deserialization at startup unless the gulp generator and the TS `@Secured` contract change too — wider than this fix.

2. **The FORM bit is still parsed into `License.components()` and ignored.** `components()` and `LicenseComponent.getInitializerClass()` also have no readers. Removing the enum value would break parsing of keys already in the field, so the parse side is untouched. If product ever wants real enforcement, the plumbing to build on is `LicenseStrategy.isMasterLicense()` -> `EnterpriseLicenseStrategy.getLicenseFlag(License::master)` — but note `ProductLicenseKeyBuilder.form` is opt-in and off by default, so enforcement would silently disable form tables for any customer whose key was issued without it.

## Testing

- `mvnw clean install -pl community/core -DskipTests` — clean build, 6443 sources, test sources compile
- `mvnw test -pl community/core -Dtest=ScriptHostAccessTest` — 13/13 pass
- Repo-wide grep confirms zero remaining references to `isComponentAvailable(LicenseComponent`, `LicenseManager.LicenseComponent`, or `formLicensed` across Java, TypeScript, HTML, JSON, tests and e2e

The paired enterprise PR removes the now-unused `builder.formLicensed(...)` line in `LicenseParser` and corrects `claude/license.md`.


jshobe-inetsoft (approved):

PR #4807, opened 2026-08-26
Bug #76291, make graph.textlayout.maxstep and graph.circle.packing behave as named

ngovus-inetsoft:
Three naming and fallback defects in the `graph.*` properties.

## 1. `graph.textlayout.maxstep` was discarded for the labels it appears to be for

`FreeHelper`'s constructor read the property, then `TextLayoutManager.createHelper` overwrote it on the next statement for any `VOText` or `TagFormVO` with `MOVE_FREE` — which is every `VOText` (`VOText`'s constructor sets `MOVE_FREE`) and every tag form. The replacement is a horizontal budget derived from the label width plus a hard-coded vertical `10`. Pie labels never reach `FreeHelper` at all; an `ArcVOText` gets an `ArcHelper` first.

Tracing every `setCollisionModifier` call site, the property survived only for bar labels pushed to `MOVE_UP`/`MOVE_RIGHT` (`BarVO`) and target-form labels (`TargetForm`). Point, line and pie labels — the ones an operator would touch this for — ignored it entirely.

The coordinate's per-direction budget is now bounded by the operator's value (element-wise `min`) rather than replacing it. This keeps the property acting as the cap its name describes, and matters because `GeoCoord.getMaxSteps` deliberately clamps to `{len, 4, len, 4}` with the comment *"don't move labels too far from points"* — an outright bypass would have discarded that and let map marker labels drift ~200px instead of 8px.

`min` also means the shipped default of `1000` is a genuine no-op, so an operator who copies `defaults.properties` into `sree.properties` doesn't silently trip a large behavior change.

The parse is now trimmed and guarded: it runs once per label, so a malformed value previously failed the entire chart render rather than degrading.

**Note:** the property can only *narrow* label travel, never widen it. That is faithful to `maxstep` and to the original constructor's use of it as a cap. If "let labels roam further to unclutter a crowded chart" is wanted, that needs a separately-named property — this one cannot carry both meanings.

## 2. `graph.circle.packing` failed to the option furthest from its default

`SeparateGraphGenerator` matched the value in a `switch` with cases for `circle`, `fill_x` and `fill` and no `default`. An unrecognised value left `TreemapCoord` at its field initialiser, `Layout.FILL`, while leaving the property *unset* gives `Layout.SQUARE` via the `"circle"` call-site default. So a typo silently selected the layout furthest from the default, with nothing logged.

Added a `default` branch that falls back to `SQUARE` and warns. The warning is keyed on the offending value through a static set, because this runs per chart per refresh and would otherwise flood the log on an auto-refreshing dashboard.

Matching is now on `Locale.ROOT`: under `tr-TR`, `"FILL".toLowerCase()` yields `"fıll"`, which would have missed `case "fill"` and — now that a `default` branch exists — actively selected the wrong layout where the old fall-through happened to be correct.

Scope is contained to `CHART_CIRCLE_PACKING`. The polar branch above still forces `SQUARE`, and treemap/sunburst never consult the property, so they keep `FILL` as intended.

## 3. `graph.axis.inplot` does not put axes in the plot

The name reads as the property behind the `PlotDescriptor` "Keep Elements in Plot" chart option, and is not. `AxisSpec.setInPlot`'s own javadoc describes reserving space for the max label; its only reader is `RectCoord.getAxisMargin`, which returns half the outermost label's width or height for `RectCoord.getAxisBounds` to fold into the coordinate bounds. Nothing repositions an axis.

This is **comment-only** — no rename. `AxisSpec.setInPlot`/`isInPlot` are `@TernMethod` with published `ProductDocs` URLs, so customer chart scripts call them by name, and the property key is shipped in `defaults.properties` and listed in EM's `booleanProperties`. A rename would also buy operators nothing: EM's property settings view renders only Name and Value, with no description field anywhere, so the name is never explained in the product either way. The external ProductDocs entries for `graph.axis.inplot` and `AxisSpec.setInPlot` are the only operator-facing description and still need correcting — that's a separate docs ticket.

## Notes

- `graph.circle.packing` is deliberately **not** added to `defaults.properties`. The call site uses `SreeEnv.getProperty(name, def)`, which routes through `DefaultProperties.getProperty(key, def)` and consults only `mainProperties` — a declaration there would be unreachable and would recreate exactly the trap documented in `FreeHelper`'s comment.
- Behavior is unchanged on the default path in both fixes: unset properties produce identical layout to before.

## Testing

`community/core` compiles clean; `PropertyDefaultsTest` passes 6/6 (it pins the `graph.textlayout.maxstep` default against `FreeHelper`'s fallback, which is still `1000`).

### DIFF
--- a/core/src/main/java/inetsoft/graph/internal/text/FreeHelper.java
+++ b/core/src/main/java/inetsoft/graph/internal/text/FreeHelper.java
@@ -27,6 +27,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.awt.geom.*;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Free movement labels.
@@ -55,8 +57,8 @@ public class FreeHelper extends LabelHelper {
             // as "unlimited" in other properties, making it a plausible typo with exactly
             // the opposite of its intended effect.
             if(parsed < 0) {
-               LOG.warn("Negative graph.textlayout.maxstep value \"{}\" would stop labels " +
-                           "from moving at all. Using 1000.", maxstep);
+               warnOnce(maxstep, "Negative graph.textlayout.maxstep value \"{}\" would stop " +
+                           "labels from moving at all. Using 1000.");
             }
             else {
                maxval = parsed;
@@ -65,13 +67,24 @@ public class FreeHelper extends LabelHelper {
          }
          catch(NumberFormatException ex) {
             // this runs once per label, so throwing would fail the whole chart render
-            LOG.warn("Invalid graph.textlayout.maxstep value \"{}\", expected an integer. " +
-                        "Using 1000.", maxstep);
+            warnOnce(maxstep, "Invalid graph.textlayout.maxstep value \"{}\", expected an " +
+                        "integer. Using 1000.");
          }
       }
 
       maxStepsFromProperty = fromProperty;
       max_steps = new int[] {maxval, maxval, maxval, maxval};
+   }
+
+   /**
+    * Warn at most once for each bad graph.textlayout.maxstep value. The constructor runs
+    * once per label, so warning unconditionally would repeat for every label on a chart and
+    * again on every refresh of it.
+    */
+   private static void warnOnce(String value, String message) {
+      if(loggedBadMaxSteps.add(value)) {
+         LOG.warn(message, value);
+      }
    }
 
    /**
@@ -529,5 +542,6 @@ public class FreeHelper extends LabelHelper {
    private boolean outside = false;
    private int sameLocDir = 3; // direction (down) to move if two labels fall on exact same point
 
+   private static final Set<String> loggedBadMaxSteps = ConcurrentHashMap.newKeySet();
    private static final Logger LOG = LoggerFactory.getLogger(FreeHelper.class);
 }
--- a/core/src/main/java/inetsoft/graph/internal/text/FreeHelper.java
+++ b/core/src/main/java/inetsoft/graph/internal/text/FreeHelper.java
@@ -48,14 +48,25 @@ public class FreeHelper extends LabelHelper {
       if(maxstep != null) {
          try {
             // Properties.load() keeps trailing whitespace and parseInt does not trim it
-            maxval = Integer.parseInt(maxstep.trim());
-            fromProperty = true;
+            int parsed = Integer.parseInt(maxstep.trim());
+
+            // move() bails as soon as steps[n] > max_steps[n] and steps starts at 0, so a
+            // negative budget stops every free-moving label rather than freeing it. -1 reads
+            // as "unlimited" in other properties, making it a plausible typo with exactly
+            // the opposite of its intended effect.
+            if(parsed < 0) {
+               LOG.warn("Negative graph.textlayout.maxstep value \"{}\" would stop labels " +
+                           "from moving at all. Using 1000.", maxstep);
+            }
+            else {
+               maxval = parsed;
+               fromProperty = true;
+            }
          }
          catch(NumberFormatException ex) {
             // this runs once per label, so throwing would fail the whole chart render
             LOG.warn("Invalid graph.textlayout.maxstep value \"{}\", expected an integer. " +
                         "Using 1000.", maxstep);
-            maxval = 1000;
          }
       }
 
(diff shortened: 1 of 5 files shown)


ngovus-inetsoft:
Thanks — went through the three nits and the test-coverage note. Addressed one, deliberately left two, and added one of the three suggested tests. Pushed in d10664b41.

**Nit 1 (negative value) — fixed.** Agreed, and it's more than cosmetic. `move()` returns as soon as `steps[n] > max_steps[n]` and `steps` starts at 0, so a negative budget stops every free-moving label rather than freeing one. What makes it worth fixing is that `-1` conventionally reads as "unlimited," so an operator setting it to *lift* the cap would instead pin every label — the same failure-furthest-from-intent shape as the `graph.circle.packing` fallback this PR exists to fix. Now warns and falls back to 1000.

`0` is left valid: it permits exactly one step, which is a coherent "barely move" setting rather than a value with the opposite of its apparent effect.

**Nit 2 (unbounded `loggedPackingValues`) — left as is.** Growth is bounded by the number of *distinct misconfigured values* of an admin-set config property, so at worst the org count. Adding eviction would be machinery guarding nothing measurable.

**Nit 3 (redundant `steps.clone()`) — kept, with a comment.** Correct that both current `getMaxSteps` implementations return fresh arrays, but it's a public overridable method and the loop mutates in place, so the clone is what keeps that safe if any subclass ever returns a cached array. Cost is one 4-int allocation against one `getMaxSteps` already performs per label. Added a one-line comment so it doesn't read as accidental and get cleaned up later.

**Test coverage — added for `FreeHelper` only.** New `FreeHelperTest` (10 cases): unset falls back and is not reported as operator-set, a set value applies in all four directions, surrounding whitespace is trimmed, and malformed/negative values fall back without throwing.

The other two are deliberately uncovered. `TextLayoutManager.createHelper` is private and covering the `min()` narrowing means populating a `VGraph` with real `VOText` visuals, running `resolve()`, and asserting on resulting label positions — brittle and disproportionate; the `FreeHelper` test covers the input half. The `circle.packing` fallback would need chart binding info, a dataset, and a full generator run for a three-line branch.

**One thing worth flagging from writing that test:** surefire in `core/pom.xml` filters on `<groups>core</groups>`, so a test without `@Tag("core")` is silently skipped — it compiles, doesn't run, and the build still passes. `@SreeHome` alone also isn't enough; without the `SpringExtension` + `BaseTestConfiguration` recipe, `SreeEnv.init` fails on a missing `PropertiesEngine` bean. The new test carries both.

That combination means some existing `@SreeHome` tests never execute under the default profile — `inetsoft/mv/util/MVRuleTest.java` and `inetsoft/graph/element/IntervalElementStackOutermostTest.java` are two I hit while looking for a pattern to copy. Out of scope for this PR, but probably worth its own issue.

ngovus-inetsoft:
Fixed in 884bc6157.

Agreed — `FreeHelper`'s constructor runs once per label, so the malformed/negative warnings repeated for every label on a chart and again on every refresh. That is the same log-flood shape already guarded for `graph.circle.packing`, and having the two sibling fixes disagree on it was inconsistent. Both warnings now go through a `warnOnce` helper backed by a `ConcurrentHashMap.newKeySet()`, keyed on the offending value so a second distinct bad value is still reported rather than swallowed by a single flag.

**On your open verification item:** I ran the tests locally, so that gap is closed.

```
./mvnw.cmd -o -pl community/core test -Dtest='FreeHelperTest,PropertyDefaultsTest'

Running inetsoft.graph.internal.text.FreeHelperTest
Tests run: 10, Failures: 0, Errors: 0, Skipped: 0
Running inetsoft.sree.internal.PropertyDefaultsTest
Tests run: 6, Failures: 0, Errors: 0, Skipped: 0
Tests run: 16, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

The run also confirms the dedup behaves: five distinct bad values produced exactly five warnings, none suppressed.

One correction to the summary above, since it matters if anyone copies the pattern later: the `SpringExtension`/`BaseTestConfiguration`/`@SreeHome`/`@Tag("core")` combination wasn't a stylistic choice — it is load-bearing, and I arrived at it by hitting both failure modes. Without `@Tag("core")` surefire silently skips the test (compiles, never runs, build still green). Without the Spring context, `SreeEnv.init` fails on a missing `PropertiesEngine` bean. `AxisPropertyDialogModelTest` is a fine reference for the combination; `MVRuleTest` is not, despite being the closest match for the `SreeEnv` set/reset idiom — it lacks `@Tag("core")` and so never executes.

jshobe-inetsoft (changes requested):
The `build` check is currently failing on the latest commit (https://github.com/inetsoft-technology/stylebi/actions/runs/33002438570/job/98287616939). Claude's automated review found no new issues in the code across all 3 commits, but please fix the failing build before this can be merged.

jshobe-inetsoft (approved):

PR #4336, opened 2026-07-21
some cases for recent mv bugs

sarapeng86:

PR #4808, opened 2026-08-26
Bug #76287, stop XJobPool from promising per-organization behavior it does not provide

ngovus-inetsoft:
## Problem

`XJobPool.getPool(String orgID)` was parameterised by organization but stored its result in a single `private static XJobPool pool`. The first organization to run a materialization job after a restart created the pool for the whole JVM; every later organization silently reused it — including the polling interval that pool read from `fs.job.check.period` in its constructor and froze into a `final int`.

That matters because `SreeEnv.getProperty(name)` **is** organization-scoped by default (`PropertiesEngine.getProperty(name, false)` → `orgScope=true` → `useAvailableOrgProperty()`), `fs.job.check.period` is not in `EXCLUDED_ORG_PROPERTIES`, and the EM Organization properties table accepts arbitrary keys. So an operator could set a per-organization value and have it silently ignored for every tenant but whichever one happened to go first — with nothing reporting the discrepancy.

## Approach

Nothing about the sanity check thread is organization-specific, and the rest of the `fs.*` family (`fs.job.period`, `fs.map.expired`, `fs.desktop`) is already JVM-global infrastructure config in `FSConfigImpl`. So this declares the pool global rather than splitting it per organization — per-organization pools would mean one watchdog thread per tenant for no functional gain.

- `getPool(String orgID)` → `getPool()`, and the constructor loses its org parameter. The signature no longer advertises per-organization pools; the genuinely per-org state (`fsystemOrgMap`) stays keyed by organization inside the single instance.
- The interval is now read on every iteration through the existing — and previously **uncalled** — `FSConfig.getJobCheckPeriod()`, which already re-reads the property per call. A change now takes effect without a restart. Non-positive and unparseable values fall back to 500: with the read moved inside the loop, `Thread.sleep(0)` would busy-spin and `Thread.sleep(-n)` / `NumberFormatException` would kill the sanity thread outright.
- The `FSService.getServer(...) == null` check moved from the constructor into `getOrgFSystem()`, where the organization is actually known. It now validates every organization instead of only the first, and replaces a latent NPE on that path with the intended `RuntimeException`.
- `pool` is now `volatile` and published only after `start()`, so a reader on the double-checked-locking fast path cannot observe a partially constructed pool. (Pre-existing, but inside a hunk this change rewrites.)

## Deliberately unchanged

The public `execute` / `addResult` / `addFailure` / `cancel` entry points keep their `orgID` parameters, now documented as unused: job ids are unique per JVM (`AbstractJob.createID` is class name + global counter), so `smap` cannot collide across organizations, and `execute0` resolves the file system from the organization carried by the job itself. `XMapTaskPoolTest` asserts that `XMapTaskPool.Handler` passes each task's *own* org to these methods (coverage from #4336); removing the parameters would make those tests vacuous.

## Testing

- New `FSConfigJobCheckPeriodTest` covers the now-live contract one layer down: default 500 when the property is unset, and a changed value reflected on a later call through the same config instance.
- `./mvnw -pl core test -Dtest=FSConfigJobCheckPeriodTest,XMapTaskPoolTest` → 5 tests, all passing. `core` compiles clean.

## Follow-ups found while reviewing this code (not fixed here)

- `resetOrgCache(orgID)` clears only `fsystemOrgMap`. Both callers pair it with `FSService.clearServerNodeCache`, which **disposes** that organization's `XFileSystem` — so that org's in-flight `XJobStatus` objects keep running against a disposed file system, `smap` is never drained, and nothing ever cancels the pool thread.
- `FSService.getServer0` re-tests the stale local `xServerNode` inside the lock instead of re-reading `servers.get(orgId)`, so two racing threads can each build an `XServerNode` for the same organization.
- `XMapTaskPool` captures one `XBlockSystem` at first init from the *ambient* organization's data node and hands it to every task regardless of that task's organization — the same shape of bug as this one, in the neighbouring singleton.
- `FSService.refresh(orgId)` refreshes `getServer(orgId)`'s file system using `getDataNode()`, i.e. the current thread's organization rather than `orgId`.

### DIFF
--- a/core/src/main/java/inetsoft/mv/mr/XJobPool.java
+++ b/core/src/main/java/inetsoft/mv/mr/XJobPool.java
@@ -20,10 +20,11 @@ package inetsoft.mv.mr;
 import inetsoft.mv.MVExecutionException;
 import inetsoft.mv.fs.*;
 import inetsoft.mv.mr.internal.XJobStatus;
-import inetsoft.sree.SreeEnv;
 import inetsoft.util.GroupedThread;
 import inetsoft.util.TimedQueue;
 import inetsoft.mv.MVJob;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4809, opened 2026-08-26
Bug #76286, remove the dead map-task retry path in XJobStatus

ngovus-inetsoft:
## Problem

`XJobStatus.createMapper(XMapTask)` was a stub whose entire body was `return null`, so the expired-map-task branch in `update()` always fell straight through to failing the whole job. Everything written around it described a retry that could not happen: the "try re-running task" comment, the `setFailed(block, host)` host-avoidance bookkeeping, the `XMapStatus`/`startTask` branch, and the "could not be re-dispatched" wording.

One block slower than `fs.map.expired` therefore failed the entire materialized-view job, with no setting that made it recoverable.

## Why remove the retry instead of implementing it

Three things make it dead code rather than an unfinished feature:

1. **There is no other host.** `startJob()` hardcodes `host = "localhost"` — the MV data grid was replaced by Spark for distributed mode — and `XMapTaskPool` runs every task in a local thread pool regardless of `task.getHost()`. There is no node registry left; `XNodeRuler`, the old host-affinity calculator, has no callers.
2. **A same-host replacement is structurally impossible.** `XMapStatus` is keyed on `(host, blockId)` and `startTask()` early-returns on duplicates, so a replacement would be silently dropped and the job would then *hang* until `fs.job.period` instead of failing — strictly worse.
3. **The expired task is still running.** Expiry is a wall-clock check, not a death notice, so a retry would duplicate the work and race a second result into the reducer.

`fmap`/`setFailed` were write-only: populated from two call sites, read by nothing.

## `fs.map.expired`

At its default the branch is unreachable — `fs.job.period` is the same length and `update()` checks it first. It only opens when an operator lowers the property, and then every expiry is terminal, so lowering it to "detect a stuck task sooner" made jobs fail instead. The property is now documented for what it is: a hard deadline that becomes effective once lowered below `fs.job.period`.

## Stopping abandoned work

Failing the job left its map tasks running for a result that is discarded. Now:

- `complete()` cancels the job's map tasks, and terminates the reducer so a streaming consumer holding the live table sees a definite end. Previously the failure path never completed the reducer, and `MVQuery.complete()` only calls `table.complete()` when `all || isCancelled()`, so the consumer blocked on `moreRows()` forever.
- `addResult()` ignores results arriving after the job is done — a cancelled task returns the rows it had scanned as a block that *looks* complete (`SubMVQuery.execute()` breaks its row loops, then still calls `group.complete()`), which would feed truncated data to the reducer.
- `startJob()` stops dispatching once the job has completed, since the cancel sweep in `complete()` has already run by then.
- `XMapTaskPool` skips tasks cancelled while still queued, and registers a task before submitting it so a concurrent cancel cannot miss it.
- `SubMVTask.cancel()` reads the query from the task's own state rather than a field assigned partway through `run()`, so a cancel arriving before the block finishes loading is no longer dropped. Cancellation flags are made `volatile`, as they are set and polled by different threads.

Cancellation remains cooperative: a task scanning rows stops at the next check (every 256 rows), one blocked in the data source does not.

## Testing

New `XJobStatusExpiredMapTaskTest` covers the expiry path end to end — the job fails naming the property, the map tasks are cancelled, the reducer is terminated, post-completion results are not merged, and a result for a still-running job is. `XMapTaskPoolTest` gains a case for the queued-cancel skip.

`inetsoft.mv.**`: 187 pass. Full core suite: 4611 pass. (`FreeHelperTest` fails on `main` independently of this change — it calls `FreeHelper.isMaxStepsFromProperty()`, which does not exist in the source.)

Manual verification of the expiry path — lowering `fs.map.expired` and building an MV with a slow block — is being done separately, since the branch is unreachable at default settings.

### DIFF
--- a/core/src/main/java/inetsoft/mv/SubMVTask.java
+++ b/core/src/main/java/inetsoft/mv/SubMVTask.java
@@ -181,7 +181,5 @@ public final class SubMVTask extends AbstractMapTask {
       //
    }
 
-
-
    private static final Logger LOG = LoggerFactory.getLogger(SubMVTask.class);
 }
--- a/core/src/main/java/inetsoft/mv/SubMVTask.java
+++ b/core/src/main/java/inetsoft/mv/SubMVTask.java
@@ -103,7 +103,7 @@ public final class SubMVTask extends AbstractMapTask {
                throw new Exception("The sub mv of the block not found: " + bid);
             }
 
-            query = getQuery();
+            SubMVQuery query = getQuery();
             SubTableBlock data = null;
 
             try {
@@ -162,6 +162,10 @@ public final class SubMVTask extends AbstractMapTask {
    public void cancel() {
       super.cancel();
 
+      // read the query from the task's own state rather than a field assigned partway
+      // through run(), so a cancel arriving before the block is loaded is not lost
+      SubMVQuery query = getQuery();
+
       if(query != null) {
          query.cancel();
       }
@@ -178,7 +182,6 @@ public final class SubMVTask extends AbstractMapTask {
    }
 
 
-   private SubMVQuery query;
 
    private static final Logger LOG = LoggerFactory.getLogger(SubMVTask.class);
 }
--- a/core/src/main/java/inetsoft/mv/data/SubMVQuery.java
+++ b/core/src/main/java/inetsoft/mv/data/SubMVQuery.java
@@ -451,7 +451,7 @@ public final class SubMVQuery implements XTransferable, Cloneable {
    XFilterNode cond;
    boolean[] order;
    private boolean detail;
-   private boolean cancelled;
+   private volatile boolean cancelled;
    private int blockIndex = -1;
    private int maxrows = 0;
    private Integer timezoneOffset; // timezone offset of server
--- a/core/src/main/java/inetsoft/mv/fs/FSConfig.java
+++ b/core/src/main/java/inetsoft/mv/fs/FSConfig.java
@@ -42,7 +42,11 @@ public interface FSConfig {
    int getJobCheckPeriod();
 
    /**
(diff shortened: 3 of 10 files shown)


ngovus-inetsoft:
Thanks — the race is real and now fixed in 821586fc.

**`addResult()` TOCTOU.** Confirmed exactly as described: `complete()` writes `state = FAILED` and calls `reducer.cancel()` while holding `waitlock`, so a check made before acquiring that lock can pass, park on the lock while another thread fails the job, and then merge into a cancelled reducer. Moved the `isCompleted()` check inside the `try`, after `waitlock.lock()`. The early `job.isCancelled()` check stays as the fast path.

The comment I had written there was also wrong about *why* the guard worked — `state` being `volatile` makes the write visible but does nothing for atomicity. Reworded to say it is checked under the lock that `complete()` holds.

Added `aResultBlockedOnTheLockWhenTheJobFailsIsNotMergedIntoTheReducer`, which reproduces the interleaving deterministically rather than hoping to hit it: the test thread takes `waitlock`, starts a thread in `addResult()` and waits for it to park on that lock, then — since the lock is reentrant — fails the job through `complete()` without releasing it, and only then unlocks. With the guard outside the lock this fails on `verify(reducer, never()).add(result)`; with it inside, it passes.

**Nits.** Both fixed: the two `"The job has expired (" + wperiod + ")"` messages now carry the `ms` suffix so both expiry paths read consistently, and the blank lines left by the removed `SubMVTask.query` field are gone.

Core suite: 4612 pass. (`FreeHelperTest` fails on `main` independently of this branch — it calls `FreeHelper.isMaxStepsFromProperty()`, which does not exist in the source.)

jshobe-inetsoft (approved):

PR #4810, opened 2026-08-26
Bug #76285, remove three inert MV data-placement properties

ngovus-inetsoft:
Fixes Bug #76285.

An operator moving materialized-view data to a larger disk meets three property names that promise to place it. None of them does. All four claims in the report were verified independently against the current tree and all four are confirmed; the `mv.data.directory` one is stronger than reported.

## What was wrong

| Property | State |
|---|---|
| `mv.directory` | Declared in `defaults.properties`, **zero** readers repo-wide |
| `mv.min.block` | Declared, **zero** readers |
| `mv.data.directory` | Read — but the value reaches no file operation |
| `fs.workdir.<node>` | Not named in the report; the pre-12.2 fallback on the same dead path |

`mv.min.block` was misleading beyond being inert because of where it sat. The three block-size properties read as a minimum/target/maximum triple and are in fact three different things — `mv.min.block` declared and unread, `mv.preferred.block` declared and read twice, `mv.max.block` read but never declared. The only member with a declaration and no reader sat directly above the one that works.

`mv.data.directory` was worse than an orphan, because a developer checking "is this read?" found a reader and stopped:

```
FSConfigImpl.getWorkDir            reads mv.data.directory, then fs.workdir.<node>, then <home>/bs
  └─ FSMessageHandler ctor         the ONLY caller in the tree; resolves to an absolute path
       └─ ClusterUtil.setWorkDir   writes cluster map inetsoft.mv.fs.internal.workDirMap
            └─ (nothing)           WORK_DIR_MAP had 2 references: its declaration and this write
```

The path was computed, resolved, published to a cluster map, and never consulted again.

## What this changes

The whole chain goes, not just the read site — leaving `FSConfig.getWorkDir` in place would let the next reader rediscover the property from the code side. Same shape as `Tool.FontSizeMapper` in #76298, which that PR's description already flagged as "the same shape as `mv.data.directory` in #76285".

- `defaults.properties` — drop `mv.directory`, `mv.min.block`
- `FSConfig` — drop the `getWorkDir(String node)` declaration
- `FSService` — drop `FSConfigImpl.getWorkDir` (sole reader of both dead directory properties) and the now-unused `ConfigurationContext` import
- `FSMessageHandler` — drop the work-dir computation and publish; keep the bare `FSService.getConfig()` eager-init call, now commented so it is not mistaken for dead code
- `ClusterUtil` — drop `setWorkDir` and `WORK_DIR_MAP`

## Where MV data actually goes

Neither destination is an `mv.*` property, which is exactly why all three names fail silently:

- **Persistent `.blk` / `.mv` blocks** are blob keys, not paths — content-addressed under `inetsoft.storage.blob.filesystem.directory` (default `{sree.home}/blob`). Note this is yaml config, not `SreeEnv`, so it is not settable from `sree.properties` or the EM.
- **Transient blocks** (`CacheBlockFile`: dim indexes, decimal-column spill, dispatcher intermediates) go through `FileSystemService.getCacheFile` under `replet.cache.directory`.

## Deliberately not changed

- **`mv.max.block`** stays undeclared. Its default is `preferred * 10`, derived at runtime from `mv.preferred.block`, so a static declaration would pin the value and silently sever that link — a behavior regression, not a fix. It is already named to the operators who hit it by the `LOG.warn` in the same method.
- **`mv.preferred.block`** is untouched: `MVBuilder.getPreferredBlock` and `LocalFileSystem` both call `Integer.parseInt` on it **unguarded**, so its declaration is load-bearing and removing it would NPE at MV build time.
- **Wiring up a real MV directory** is a feature, not this bug: `CacheBlockFile`/`StorageBlockFile` have no directory seam today and blob semantics would need deciding.

## No behavior change

`PropertiesEngine.getDefaultProperties()` loads `defaults.properties` purely as the fallback layer of a `DefaultProperties` chain, so removing keys that nothing reads cannot alter any lookup result. Administrator overrides in KV storage are a separate layer and are unaffected. Lookups are case-insensitive (`PropertiesEngine.computePropertyNameCase`), so no differently-cased read site could have resolved these either. The cluster work-dir map was write-only, so dropping the write is neutral.

No EM change is needed this time (unlike #76052 and #76298): `PropertiesTool.booleanProperties` lists only live `mv.*` keys. Settings > All Properties simply stops listing two dead rows.

## Verification

- `mvn compile -pl core` clean.
- Reference sweep: **zero** hits for `mv.directory`, `mv.min.block`, `mv.data.directory`, `fs.workdir`, `getWorkDir`, `setWorkDir`, `workDirMap` across `community/`, `enterprise/`, `connectors/`.
- `mvn test -pl core -Dtest='MV*Test,XJobStatus*Test,*BlockFile*Test'` — 51 tests, 0 failures. Includes `MVSingleDispatcherTest` (mocks `FSConfig`, stubs only `isDesktop()`, so the interface change cannot break it) and `MVBuilderMaxBlockTest` (guards the `mv.max.block` path left untouched).
- `FSConfig` has exactly one implementation, so removing the interface method cannot orphan an `@Override`.

Community-only change; the MV fs layer has no enterprise counterpart, so no enterprise PR is needed.


jshobe-inetsoft (approved):

PR #4813, opened 2026-08-26
Bug #76318, stop script.java.allowed.classes from overriding the script sandbox deny lists

ngovus-inetsoft:
## Problem

`ScriptHostAccess.isVisibleToScripts` tested the operator-supplied
`script.java.allowed.classes` extras in the same condition as the curated
`ALLOWED_CLASSES` set, *before* consulting `BLOCKED_PACKAGES` and
`BLOCKED_CLASSES`:

```java
// exact allow wins: curated classes, SreeEnv extras, primitive arrays, jdk proxies.
if(ALLOWED_CLASSES.contains(fqcn) || extra.contains(fqcn) ||
   isPrimitiveArrayType(fqcn) || fqcn.startsWith("jdk.proxy"))
```

Naming a denied class in that property therefore made it resolvable through
`Java.type(...)`, and through the legacy package-root shim as well, since
`GraalJavaScriptEngine` hands the same predicate to `LegacyJavaShim` →
`JavaPackageProxy` and `BindingRootProxy`.

Three things mark this as an oversight rather than a decision:

1. **The sibling operator control already behaved correctly.**
   `javascript.java.packages` is consulted *after* both deny checks. Two
   operator-supplied allow controls in one method disagreed on precedence, and
   only one of them documented why.
2. **The javadoc asserted the opposite** — "Dangerous classes … are NOT on any
   allow path, so they stay blocked." They were on one: this one.
3. **`ALLOWED_CLASSES` earned its pre-deny position by audit** (the comment
   above it says so). The extras were folded into that `if` and inherited the
   exception silently.

### Impact

Not remotely exploitable and not urgent: setting a server property requires
administrator access. What makes it worth fixing is the gap between the control
and its documentation — the property reads as an additive extension but was in
fact a deny-list override, and scripts are authored by anyone with write access
to a dashboard, not only by administrators.

Worth noting that `java.lang.Runtime` is the *weakest* example, not the
strongest: `hostAccess()` independently calls `.denyAccess(...)` on `System`,
`Runtime`, `Process`, `ProcessBuilder`, `Thread`, `Class`, `ClassLoader` and the
reflect types, so naming `Runtime` let `Java.type` resolve it but left
`getRuntime()` blocked. The classes that genuinely crossed the boundary are the
ones with no member-level deny — `java.io.File`, `java.net.URL`,
`java.util.concurrent.*`, `inetsoft.util.Plugins`, `inetsoft.storage.*`,
`inetsoft.report.internal.license.*`.

**No deployment can currently be affected.** `script.java.allowed.classes` is
referenced nowhere else in either repo — no bundled `.properties` default, no EM
UI, no other reader — so it cannot be set without hand-editing
`sree.properties`.

## Change

- Move the `extra.contains(fqcn)` test below both deny checks. `ALLOWED_CLASSES`
  keeps its pre-deny position, since that precedence is deliberate and
  documented.
- Add a package-private `classFilter(Set<String> extra, String[] customPkgs,
  boolean comOrg)` overload; the public no-arg version delegates to it. This
  lets the precedence be tested without a Spring/`SreeEnv` context — the
  existing `SreeEnv` reads are already wrapped in `catch(Exception ignore)` for
  the same reason. `GraalJavaScriptEngine` is unchanged.
- Correct the `classFilter` javadoc and the deny-list header comment, and state
  in the javadoc that the extras are additive only.

`isAllowedObjectArray` recurses back into `isVisibleToScripts` with the same
`extra` set, so array component types follow the corrected precedence for free.

## Tests

Three new tests in `ScriptHostAccessTest`, which had no coverage of this
property at all:

- `extrasCannotUnblockBlockedClass` — `java.lang.Runtime` and `java.io.File` in
  the extras, both still denied (`BLOCKED_CLASSES`)
- `extrasCannotUnblockBlockedPackage` — `java.lang.reflect.Method` and
  `LicenseManager`, both still denied (`BLOCKED_PACKAGES`)
- `extrasStillAddNonBlockedClass` — the positive half, so the fix doesn't
  quietly neuter the property: `java.time.LocalDate` is denied without the entry
  and resolves with it

`ScriptHostAccessTest` 16/16 pass, including the four that pin deliberate
behavior (`curatedExactClassOverBlockedPackage`, `dangerousClassesStillBlocked`,
`uncuratedBlockedPackageClassDenied`, `threadingStaysBlocked`). The paths that
share the same predicate instance are green too: `LegacyJavaShimTest` 18,
`HostBeanProxyTest` 7, `ScriptUtilDateUnwrapTest` 2, `JSObjectTest` 15.

## Notes

Two adjacent issues found while reading `isVisibleToScripts`, deliberately left
alone here since both are documented main-branch Rhino parity and changing them
is a behavior decision rather than a bug fix:

- `fqcn.startsWith("java.sql")` under a FORM license runs before *everything*,
  including the deny lists, and is a dotless prefix — it exposes all of
  `java.sql.*` (e.g. `DriverManager`) despite `"java.sql"` being in
  `BLOCKED_PACKAGES`. `uncuratedBlockedPackageClassDenied` passes only because
  `isFormLicensed()` returns `false` without a license context.
- `DEFAULT_JAVA_PKGS` is matched with a dotless `startsWith`, so `java.awtFoo`
  matches.

Community-only change; no enterprise PR and the enterprise gitlink is untouched.

Related: #75423, the GraalJS cutover that introduced `isVisibleToScripts` as a
port of `SecureClassShutter.visibleToScripts`.

### DIFF
--- a/core/src/main/java/inetsoft/util/script/graal/ScriptHostAccess.java
+++ b/core/src/main/java/inetsoft/util/script/graal/ScriptHostAccess.java
@@ -34,7 +34,8 @@ public final class ScriptHostAccess {
    }
 
    // Deny-list ported verbatim from SecureClassShutter. Checked AFTER the exact
-   // allow-list (ALLOWED_CLASSES) but BEFORE the allowed prefixes — see classFilter().
+   // allow-list (ALLOWED_CLASSES) but BEFORE the operator-supplied
+   // extras (script.java.allowed.classes) and the allowed prefixes — see classFilter().
    // Dangerous packages — any class whose FQCN equals or starts with "<pkg>." is blocked.
    private static final Set<String> BLOCKED_PACKAGES = Set.of(
       "java.lang.reflect",
@@ -223,8 +224,12 @@ public final class ScriptHostAccess {
     *
     * <p>Delegates to {@link #isVisibleToScripts}, a faithful port of the proven
     * {@code SecureClassShutter.visibleToScripts} precedence from the main-branch
-    * Rhino baseline. Reads the {@code javascript.java.packages} and
-    * {@code javascript.java.com_org} properties once when the filter is built.
+    * Rhino baseline. Reads the {@code script.java.allowed.classes},
+    * {@code javascript.java.packages} and {@code javascript.java.com_org}
+    * properties once when the filter is built. The extra class names given by
+    * {@code script.java.allowed.classes} are additive only: they are consulted
+    * after the deny lists, so they cannot re-enable a class named by
+    * {@code BLOCKED_PACKAGES} or {@code BLOCKED_CLASSES}.
     * Dangerous classes (System/Runtime/Class/ClassLoader, threading,
     * inetsoft.report.internal.license.*, engine internals) are NOT on any allow
     * path, so they stay blocked.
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):

PR #4814, opened 2026-08-26
Bug #75950, #76328: set_ranking $(variable) binding, variable enumerated values

jkidacki-inetsoft:
## Summary

Two related AI-agent worksheet-chat backend fixes, shipped together since they're both about parameterizing worksheet behavior via variables:

**Bug #75950** — `set_ranking`'s Top-N/Bottom-N row count (`n`) was hardcoded to a plain `int`, so it could never be bound to a worksheet variable via `$(variableName)`, even though `RankingCondition.setN(Object)` already fully supports it (an `Integer`, a numeric `String`, or a `"$(name)"` reference — the same mechanism filter condition values already use).

**Bug #76328** — `add_variable`/`edit_variable` had no way to populate a worksheet variable's enumerated "Values" picker (`UserVariable.setValues()`/`setChoices()`/`setDisplayStyle()`), matching what the Composer's own Variable dialog offers: an embedded list, a query against an existing worksheet table's columns, and the full combobox/list/radio/checkboxes display style range — not just a fixed list.

## Root Cause

- **#75950**: `WorksheetMutationSupport.RankingSpec.n` was declared `int`, so `setRanking()` could only ever hand `RankingCondition.setN` an autoboxed `Integer` — a `$(name)` string could never reach it.
- **#76328**: `WorksheetAgentController.createVariable()` / `WorksheetEditService.Editor.editVariable()` only ever read `name`/`type`/`label`/`defaultValue` off the request; nothing ever called `UserVariable.setChoices()`/`setValues()`/`setDisplayStyle()`/`AssetVariable.setTableName()`/`setLabelAttribute()`/`setValueAttribute()`.

## Fix

- `RankingSpec.n` widened from `int` to `Object`; `setRanking()` validates via `RankingCondition.setN(Object)`'s boolean return and throws `IllegalArgumentException` on an invalid value.
- New `WorksheetMutationSupport.VariableChoicesSpec` (embedded `values`/`labels`, or query-mode `table`/`labelColumn`/`valueColumn`, plus `displayStyle`) and `applyVariableChoices()`, modeled directly on the Composer's own `VariableAssemblyDialogService.convertModelToAssetVariable()`. Wired into both `add_variable` (direct REST endpoint + edit-dispatch path) and `edit_variable`.
- `EditRequest` gained a `choices` field (record's new last component), with a backward-compat constructor added for the previous shape — no existing positional-constructor call site needed to change.
- Fixed along the way: `UserVariable` defaults `sortValue=true`, which silently re-sorts a picker's values alphabetically the instant both arrays are set (now explicitly disabled, matching the Composer dialog's own behavior); `AssetVariable.getDisplayStyle()` doesn't auto-derive from choices/values like the base `UserVariable` does, so display style is now explicitly (re)computed whenever the picker changes; validation (labels/values length mismatch, unknown display style) now runs fully before any mutation of the live worksheet, since the edit pipeline has no rollback on a thrown exception.
- **Circular-dependency guard**: found via live testing — pointing a variable's query-mode picker at a table that itself depends on that same variable (e.g. a mirror of a table whose own ranking condition reads `$(thisVariable)`) is now rejected with a clear error, since computing the picker's own values would otherwise require the variable's value first.

## Test Plan

- `mvnw clean test -pl core -Dtest=WorksheetEditServiceMutatorsTest,WorksheetAgentControllerTest` — 219/219 passing, including new coverage for `$(variable)` ranking binding, embedded/query-mode variable choices, display style derivation, mutual-exclusivity/mode-switch clearing, and the circular-dependency guard (direct, transitive, and non-circular control cases).
- `mvnw clean test-compile -pl core` (whole module) — clean, confirming no other call site broke from the `EditRequest`/`RankingSpec` shape changes.
- Live-verified end-to-end against a local StyleBI deployment via the stylebi-composer-chat plugin (paired repo PR).
- Two independent code-reviewer passes, both clean after fixes (a partial-mutation-before-validation ordering bug caught and fixed in the second pass).

stylebi-wiz PR: https://github.com/inetsoft-technology/stylebi-wiz/pull/new/bug-75950-76328 (opening next)

Fixes Redmine #75950, #76328.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
@@ -102,10 +102,9 @@ import java.util.Map;
  *   <li>{@code duplicate_assembly} — {@code table} (source), {@code name} (new name)</li>
  *   <li>{@code set_primary_assembly} — {@code table}</li>
  *   <li>{@code edit_variable} — {@code name}, {@code type}, {@code label}, {@code defaultValue};
- *       optional {@code values} (enumerated picker values — empty clears it, matching the
- *       Composer's own Variable dialog "Values"; omitted leaves it unchanged), optional
- *       {@code labels} (display labels parallel to {@code values}; defaults to {@code values}
- *       themselves), optional {@code multipleSelection} (checkbox/multi-select vs single-select)</li>
+ *       optional {@code choices} (the enumerated "Values" picker — embedded list or query
+ *       source, plus display style — matching the Composer's own Variable dialog; {@code null}
+ *       leaves it unchanged, see {@link WorksheetMutationSupport.VariableChoicesSpec})</li>
  *   <li>{@code rename_variable} — {@code name}, {@code newName}</li>
  *   <li>{@code delete_variable} — {@code name}</li>
  *   <li>{@code edit_named_group} — {@code name}, {@code groupMappings} (see {@code add_named_group}
@@ -146,10 +145,7 @@ public record EditRequest(
    String field,
    /** Comparison operator for add_filter, e.g. {@code "="}, {@code "!="}. */
    String operation,
-   /**
-    * Literal values for add_filter / edit_condition. Also doubles as the enumerated picker
-    * values for edit_variable (see {@code labels} and {@code multipleSelection}).
-    */
+   /** Literal values for add_filter / edit_condition. */
    List<String> values,
    /** Sort direction — {@code "ASC"} or {@code "DESC"} — for set_sort. */
    String direction,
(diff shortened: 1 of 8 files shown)


jkidacki-inetsoft:
Addressed both findings in 413aa2b7a6:

1. **Ordering/atomicity in `editVariable`**: now edits a scratch clone (`existing.clone()`) and only publishes it via `DefaultVariableAssembly.setVariable(...)` once `label`/`type`/`defaultValue`/`choices` have all applied without throwing — matching the property `add_variable` already had. New test: `editVariableRejectsInvalidChoicesWithoutApplyingLabelTypeOrDefaultValue`.

2. **Silent `null` from `Tool.getData`**: `applyEmbeddedVariableChoices` now validates each embedded value against the variable's type up front and throws `IllegalArgumentException` naming the offending entry, instead of writing a `null`. New test: `editVariableRejectsUnparsableEmbeddedValue`.

Also fixed a related gap found while reviewing the ranking-condition side of this PR: `WorksheetReadService.extractValues()` returned an empty list for any `RankingCondition`, so a Top-N count bound to `$(variableName)` (Bug #75950) was invisible to the stylebi-wiz plugin's `findVariableReferences` dangling-reference scan. `RankingCondition`'s `n` is now surfaced the same way a filter condition's values are. New tests: `rankingConditionSurfacesVariableReferenceInValues`, `rankingConditionSurfacesNumericNInValues`.

251/251 tests passing, `mvnw clean test-compile` clean on the whole module.

jshobe-inetsoft (approved):

PR #4815, opened 2026-08-26
75879, milestone points should be dimmed

milotalon:
Root Cause:
Gantt charts draw interval bars (BarVO -> .inetsoft-bar) plus milestone markers (PointElement -> .inetsoft-point) in one SVG. The cross-chart-type hover-dim rules that make a point hover fade bars are emitted by
SVGAnimationDOMInjector.appendMultiStyleHoverCSS, but that method explicitly returned early whenever the ":gantt" animation-hint flag was present. The exclusion was added with #75871 on the premise that a gantt milestone marks the end of its own bar, so a bar hover must not fade it. That premise does not hold: GanttChartInfo.getMilestoneField() is an independent aggregate (forced to CHART_POINT) that only shares the category axis with the bars, so a milestone is not part of any bar's mark. With the rules suppressed, hovering a milestone dimmed only the other milestones and left every bar fully lit.

Fix:
Removed the gantt exclusion from appendMultiStyleHoverCSS, so the bar<->point cross-dim rules are emitted whenever an SVG contains both annotation classes - gantt included, exactly as for a bar+point multi-style chart and consistent with the box/point cross rules from #75875. Hovering a milestone now dims the bars and bar labels; hovering a bar dims the milestones and their labels. The guard that skips these rules for single-mark charts (plain bar, plain point, pure box plot) is unchanged, and the gantt entrance-animation timing (milestones fade in after the last bar) is untouched. Unit tests covering the old suppression were updated to assert the new behavior.

### DIFF
--- a/utils/inetsoft-xml-formats/src/main/java/inetsoft/util/graphics/animation/SVGAnimationDOMInjector.java
+++ b/utils/inetsoft-xml-formats/src/main/java/inetsoft/util/graphics/animation/SVGAnimationDOMInjector.java
@@ -65,7 +65,7 @@ public class SVGAnimationDOMInjector {
 
       String base = animHint.split(":")[0];
 
-      appendHoverCSS(svgRoot, doc, animHint);
+      appendHoverCSS(svgRoot, doc);
 
       // Hover-only mode (design-time surfaces): the hover-dim CSS injected above is enough for
       // highlighting to work, but the entrance animation must be suppressed. Return before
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (changes requested):
The `build` check is currently failing on the latest commit (https://github.com/inetsoft-technology/stylebi/actions/runs/33014082676/job/98327642648). Claude's automated review found no blocking issues in the code itself, but please fix the failing build before this can be merged.

jshobe-inetsoft (approved):

PR #4700, opened 2026-08-21
Fix duplicate-primary and mislabeled action buttons across composer dialogs

swaker854:
Stacked on #4681 (cleanup-composer-0820 → epic-74519).

## Summary
- Worksheet join/union editor, physical-model join editor, and query editor join pane all had the same `joinEdit` Cancel+Done footer where both buttons were `btn-primary` — Cancel is now `btn-default` in all three.
- Parameter dialog/page had OK+Clear+Cancel all `btn-primary` — Clear and Cancel are now `btn-default`.
- Console dialog had Clear+OK both `btn-primary` (with the surviving primary button sandwiched between two secondary ones) — Clear is now `btn-default` and OK is reordered to lead the footer.
- Console dialog's "Close" button relabeled to "Cancel" — it actually discards the pending message-level filter change rather than committing it like OK does, so the old label was misleading.
- Worksheet schema/concat table title background moved off the rose family (`--inet-ws-header-secondary-bg-color`, too close in hue to the danger/incompatible-red used in the same schema editor) onto a neutral surface token (`--inet-ws-schema-title-bg`, aliasing `--inet-shell-surface-subtle`) matching the physical-model join editor's table titles. Exposed the new token in the EM theme editor.

## Test plan
- [ ] Open a worksheet, create a relational join / merge join / concatenation, verify Cancel is no longer primary-colored and table titles are neutral gray instead of pink
- [ ] Open the physical model join editor and the query editor's join pane, verify Cancel is no longer primary-colored
- [ ] Open a viewsheet parameter prompt dialog, verify only OK is primary
- [ ] Open the composer console dialog, verify button order is OK/Clear/Cancel with only OK primary, and Cancel actually discards level-filter changes
- [ ] Spot-check EM theme editor's Worksheet section for the new `--inet-ws-schema-title-bg` entry

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4816, opened 2026-08-26
Sweep native/unstyled browser elements from the composer and portal

swaker854:
Stacked on #4700 (bugfix-composer-primary-buttons-0820 → epic-74519).

## Summary
Sweeps composer and the rest of portal for browser-native visual elements that can't be restyled to match the app's theme (raw `<select>`, unstyled checkboxes, native date/datetime pickers, native `window.alert`/`window.prompt`), replacing each with the app's existing themed equivalents — or a new one where none existed yet.

**Native form controls → existing themed components:**
- `variable-table-list-dialog`'s table picker was a native multi-row `<select size="7">` listbox → `generic-selectable-list` (same component `screens-pane.component.html` already uses for this).
- `edit-custom-patterns-dialog`'s group-level picker was a native `<select>` → `custom-select` (matching its sibling sheet/encoding pickers).
- `import-csv-dialog`'s file input rendered the browser's native "Choose File" control → hidden input + themed button (same pattern as `image-preview-pane.component.html`).
- `simple-condition-pane`'s And/Or join control was a native `<select>` with `appearance:none` and a hand-drawn caret glued on top (the popup itself stayed unstyled) → `custom-select`.
- `presenter-property-dialog`'s boolean presenter option was a bare checkbox with no app styling → added `form-check-input`.
- `range-slider-edit-dialog`'s date/datetime-local inputs (shown on a viewer-time quick-click of a date-bound Range Slider) rendered the browser's native calendar/time pickers → `date-value-editor` / `time-instant-value-editor`, the same combo already used in condition and parameter dialogs elsewhere.

**New component:**
- Added `generic-multi-select-list`, a themed multi-select listbox (checkbox rows in the same bordered/scrollable box as the existing single-select `generic-selectable-list`), and wired it into the two remaining native `<select multiple>` sites: `variable-collection-selector`'s List display style and `parameter-page`'s ListParameter prompt.

**Native browser dialogs:**
- `chat.service.ts`'s Tawk.to integration called `window.alert`/`window.prompt` directly. `openSession()` now returns a `Notification` that the caller routes through composer's existing toast pipeline; the failure path copies the session info to the clipboard automatically instead of asking the user to copy it out of a native prompt.

Also added the two new i18n strings (`Choose File`, `No file chosen`) to `srinter.properties`.

## Test plan
- [x] `tsc --noEmit` clean on both `tsconfig.app.json` and `tsconfig.spec.json`
- [x] `ng lint` clean
- [x] Manually verified in-browser: import file picker, table listbox, custom-select swaps, presenter checkbox, and the multi-select listbox (via a worksheet Variable prompt with Display Style = List)
- [ ] Range slider date/datetime editor (only verified the numeric-bound path live; date-bound path verified by code only — needs a date-bound Range Slider to check visually)
- [ ] Chat widget notification (requires an active Tawk.to session to trigger `openSession()`)


jshobe-inetsoft (approved):
## Review of PR #4816 — Sweep native/unstyled browser elements from composer and portal

I reviewed the full diff plus the surrounding files (not just the patch context) for each changed component, including the components being reused (`custom-select`, `generic-selectable-list`, `date-value-editor`, `time-instant-value-editor`) and existing test suites that exercise the touched components.

### Correctness

- `chat.service.ts`: `openSession()` now returns `Observable<Notification>` instead of calling `window.alert`/`window.prompt` directly. The success path emits once and completes; the error path attempts `navigator.clipboard.writeText()` and emits a different message depending on whether the copy succeeded, always completing via `.finally()`. The single call site (`composer-toolbar.component.ts`) was updated to subscribe and route through the existing `notify()` → `onNotification` toast pipeline, and the mock in `composer-toolbar.spec-helpers.ts` was updated to return an observable (`of({...})`) so existing specs keep compiling. This is correctly wired end-to-end.
- `range-slider-edit-dialog`: replacing the native `date`/`datetime-local` inputs with `date-value-editor`/`time-instant-value-editor` is careful about format compatibility — the new `TIME_INSTANT_FORMAT = "YYYY-MM-DDTHH:mm"` constant matches (with a code comment calling this out) the no-seconds string produced by `formatDate()`, and the plain-date branch relies on `date-value-editor`'s default format (`YYYY-MM-DD`), which also matches. Verified this against the full `formatDate()`/validator code — the round trip is consistent.
- `import-csv-dialog`: hidden `<input type="file">` + themed button is a standard, correct pattern (matches `image-preview-pane` per the PR description); `openFileUpload()` resets `value = ""` before opening the file picker, which is a nice correctness touch (allows re-selecting the same file to retrigger `change`). The `model.fileName` interpolation is safely guarded behind an existing `@if (model)` block, so no null-dereference risk.
- `generic-multi-select-list` (new component): `toggle()` rebuilds the emitted array from `values` order rather than appending to `selected`, with a clear comment explaining why (keeps emission order matching list order like a native multi-select). Logic checked against both call sites (`variable-collection-selector`, `parameter-page`) and looks correct.
- `variable-table-list-dialog` / `edit-custom-patterns-dialog` / `simple-condition-pane`: straightforward swaps to `generic-selectable-list` / `custom-select` with matching option shapes (`CustomSelectOption`-compatible objects); no dangling references to the removed CSS classes (`simple-pane__select-wrap`, `simple-pane__caret`, `simple-pane__join-ctrl`, `form-control-file`) remain anywhere in the tree.
- i18n keys `Choose File` / `No file chosen` were added to `srinter.properties` in the correct alphabetical position.

### Tests

- Searched for existing `.tl.spec.ts`/`.spec.ts` files covering every touched component (`import-csv-dialog`, `range-slider-edit-dialog`, `edit-custom-patterns-dialog`, `parameter-page`, `presenter-property-dialog`). None of them assert against the removed native DOM (`<select>`, `<input type="file">`, bare checkbox) — they exercise component methods/outputs directly or use component stubs, so this PR does not break any existing spec.
- No new automated test coverage was added for the new `chat.service.ts` branching logic (clipboard-copy success vs. failure vs. the happy path) or for the new `GenericMultiSelectList` component. This is a minor gap — worth adding follow-up coverage, but not a blocker given the rest of the change is low-risk UI wiring and the author's manual test plan covers the interactive paths.
- Confirmed that `range-slider-edit-dialog.component.tl.spec.ts` (which does real-DOM-render the date branch, now instantiating `date-value-editor`/`time-instant-value-editor` for the first time) is a **portal** `*.tl.spec.ts` file; per this repo's own `package.json` (`"test": "test:portal && test:em && test:em:tl"`), portal TL specs are not wired into `npm run test`/CI, consistent with CLAUDE.md's note that `test:portal:tl` is "not yet wired to CI." So even if this changes that suite's real-DOM behavior, it won't surface as a CI failure — pre-existing gap, not introduced by this PR.

### Minor / non-blocking suggestions

1. **Accessibility**: `generic-selectable-list` and the new `generic-multi-select-list` have no `tabindex`, `role`, or keyboard handlers on their row elements, unlike `custom-select` (which implements `role="listbox"`/`role="option"` + full keydown handling). This means the native multi-row `<select size="7">` / `<select multiple>` controls being replaced — which supported arrow-key navigation and keyboard selection — lose that capability. `generic-selectable-list` is a pre-existing component so this isn't a new regression there, but the new `generic-multi-select-list` had the opportunity to add basic keyboard support and didn't. Worth a follow-up, not a blocker.
2. **Interaction model change**: `generic-multi-select-list.toggle()` is a plain click-to-toggle (no ctrl/shift-click semantics), which differs from a native `<select multiple>` (where a plain click normally replaces the whole selection). This is reasonable given the checkbox-row UI, but it is a user-facing behavior change not explicitly called out in the PR description.
3. **Cross-repo API break**: `ChatService.openSession()`'s public signature changed from `void` to `Observable<Notification>`. The only caller in this repo was updated correctly, but since this repo is wrapped by the enterprise monorepo, it would be worth double-checking there are no other callers in `enterprise/`/`server/` that assume the old `void` signature (unlikely for a portal-only chat widget, but cheap to confirm).

### Security

No issues found. The new `navigator.clipboard.writeText()` usage only ever copies a `vsId`/`runtimeId` query string (already sent to the client), guarded with optional chaining and a `Promise.reject()` fallback for browsers/contexts without Clipboard API support. No new HTML interpolation, no `innerHTML`, no new remote calls.

### CI / mergeability

- `gh pr checks 4816`: `claude-review` fails, but per review instructions that check is superseded by this manual review and is ignored. All other checks pass (`Semgrep OSS`, `Semgrep Scan`, `build`, `license/cla`).
- `gh pr view --json mergeable,mergeStateStatus`: `mergeable: MERGEABLE` — no merge conflicts. (`mergeStateStatus: UNSTABLE` reflects the failing `claude-review` check only.)

### Conclusion

No blocking issues. This is a well-scoped, low-risk UI consistency sweep with correct wiring throughout, and none of the existing tests are broken by the DOM restructuring. Recommending approval, with the accessibility/keyboard-support gap in the new `generic-multi-select-list` and the missing unit tests for `chat.service.ts` flagged as good non-blocking follow-ups.

PR #4819, opened 2026-08-27
Fix two ways the tabular query contract fails on a connector nobody looked at

Nelson-cao:
Two independent defects in the path that builds a tabular data source's parameter contract. Both were written with a stated premise about what shipped connectors declare, and both premises were wrong on the connectors that exist.

---

# 1. Probe a connector's setters only where a dialog could reach them, and under a deadline

`647ed1cb1` — `TabularSchemaExtractor`

## Problem

Every value `TabularSchemaExtractor` probes with reaches a connector's own setter, and `probe()` wrote it straight onto a freshly constructed bean: one property set, nothing else, regardless of what the layout currently shows. That is a state no dialog can produce, and nothing bounded how long the setter could take.

Reached that way, `RestJsonQuery.setLookupIgnoreBaseUrl(0, true)` never returns:

```java
// RestJsonQuery:857
if(index >= lookupIgnoreBaseUrl.size()) {
   if(index < LOOKUP_QUERY_LIMIT && ignoreBaseUrl) {
      while (lookupIgnoreBaseUrl.size() <= index) {
         addLookupQuery();            // base version appends to lookupIgnoreBaseUrl
      }
   }
   ...
```

`EndpointJsonQuery` overrides `addLookupQuery` without calling `super`, and its version only appends to `lookupEndpoints` — on a fresh probe it does not even get that far, because `endpoint` is null and it returns at the first guard. The loop condition can never be satisfied.

`lookupIgnoreBaseUrl0..4` are inherited by every `Rest.*` query class and `findAxes` probes a boolean with `TRUE` first, so this was reached on **every** REST endpoint table build. The loop allocates nothing and blocks on nothing, so `interrupt()` has nowhere to land and `Thread.stop` is gone: the request thread spins at 100% and is never reclaimed. A thread dump from a running instance showed three `http-nio-8080-exec` threads in this state, 18 minutes of CPU each — one per build attempt, and the caller retries.

`isSelfGating`'s own javadoc already names the family ("setters that grow the list a panel's visibility is computed from"), but it runs after `probe` returns.

## Fix

**Take the write rules from the interactive path** rather than writing directly:

- Apply a value only where the layout does not place that parameter, or currently shows it — the rule `TabularUtil.setValuesToBean` follows when a dialog submits a form. Read through `collectVisible`, not the per-node `TabularView.isVisible` flag: `LayoutCreator` sets that flag `true` on every node and `callVisibleMethod` does not propagate to children, so a field inside a hidden panel reports itself visible. Only the ancestor-aware view sees this case.
- Apply values one at a time with the conditions re-evaluated between them — what the dialog's refresh round trip does, and the only order in which `addCombinationGates` can set a parameter whose outer value has to reveal it first.

An axis whose value cannot be applied answers `null`, which `buildDependencyMatrix` already treats as "drop this axis" — no new failure path.

`setValuesToBean` itself is deliberately not called: it is a whole-form submit, and its `affectedViews` suppression would silently drop `paginationType` on the ten connectors that list it under `@View1(value = "endpoint", affectedViews = {...})`.

**Run each probe under a deadline**, because the rules above bound which setters are entered, not what one does once entered — this class calls into 89 plugin jars it cannot audit. On timeout the axis is dropped and the property is recorded so it is never probed again, keyed on the class **declaring** the setter so an inherited one is a single entry rather than one per connector. A caller that retries cannot strand another thread.

The abandoned thread still cannot be reclaimed; what changes is that there is at most one per broken setter for the life of the process, instead of one per request.

## Tests

Two cases on their own fixtures:

- `aValueTheLayoutDoesNotShowIsNeverWritten` — a boolean in a panel gated on a list only a button grows: the setter is never entered, the parameter is still reported, it gates nothing, and the axes that *can* be set are still probed.
- `aSetterThatDoesNotReturnIsAbandonedAndNotProbedAgain` — the setter is entered once, its axis is dropped, the other axes still produce a matrix, and a second `extract` does not enter it again.

Both pre-existing two-level gate cases (`mode & linkStyle`, `scoped & scope`) pass on the new one-value-at-a-time path.

---

# 2. Identify a file target's properties by name, not by an inferred role marker

`b6bb3c711` — `TabularQueryParamsSchemaBuilder`

## Problem

`applyRoleFormat` stamped `format: "file-path"` on the property naming a file and `format: "file-sheet"` on the sheet selector, so a caller holding a path and a sheet could find the properties to put them in. Neither inference held on either shipped file connector.

The sheet signature — a `String` with a `tagsMethod` that `dependsOn` a `File`-typed property — describes an editor whose choices are recomputed when the file changes. A charset picker fits it exactly:

```java
@Property(label = "Sheet")
@PropertyEditor(dependsOn = {"fileFolder"}, enabledMethod = "isExcel", tagsMethod = "getExcelSheetNames")
public String getExcelSheet()

@Property(label = "Encoding")
@PropertyEditor(enabledMethod = "isText", dependsOn = {"fileFolder"}, tagsMethod = "getEncodingTypes")
public String getEncoding()
```

`dependsOn` is documented as *"the list of properties on which **the editor** depends"* — a redraw dependency, not a semantic gate — so nothing in the signature separates the two. `isExcel`/`isText` do not either: both return true for a directory, which is what the data source that surfaced this is.

The ambiguity guard threw, and it threw from inside the per-property loop in `build`, taking the whole schema with it. `GET /tabular/query-schema` answered 500 for every `SERVER_FILE` data source — including the `file-path` marker, which was never ambiguous:

```
java.lang.IllegalStateException: Data source type 'SERVER_FILE' declares 2 candidate sheet-selector
properties (excelSheet, encoding) -- ambiguous, refusing rather than guessing which one selects the sheet.
        at TabularQueryParamsSchemaBuilder.applyRoleFormat(TabularQueryParamsSchemaBuilder.java:353)
        at TabularQueryParamsSchemaBuilder.buildParamFragment(TabularQueryParamsSchemaBuilder.java:291)
        at TabularQueryParamsSchemaBuilder.build(TabularQueryParamsSchemaBuilder.java:86)
        at WizTabularController.getQuerySchema(WizTabularController.java:315)
```

`OneDriveQuery` reached neither branch: it names its file with a `String` (`path`), not a `java.io.File`, so `filePaths` is empty, the type rule marks nothing and the sheet rule returns before the ambiguity check. Of the two file connectors that exist, one answered 500 and the other answered a schema with no markers at all.

The method's javadoc asserted the opposite — *"Today no shipped connector reaches that branch (ServerFileQuery declares exactly one candidate) … only ever fires on a connector nobody has looked at"* — and the existing test used a single-candidate fixture, so the branch had no coverage.

## Fix

Remove the inference rather than adding a discriminator. Nothing declared on a connector says "this selects the sheet", and the only thing that separates the two candidates today is which `enabledMethod` name they carry, which is the connector-specific guesswork this class was written to avoid.

What a caller matches on instead is already published for every property: its own name, the connector's `@Property` label as its `description`, and any `pattern` it declares — the file property declares `^.*\.(txt|csv|xls|xlsx)$` in **both** connectors, including the one whose file property is a `String`. That is the matching `binding.parseOptions` has always needed, since those keys are already the connector's own names; `path` and `sheet` being special-cased was the anomaly.

## Tests

- `noPropertyCarriesARoleFormat` — replaces the two role-format cases: no property in either fixture carries `format`.
- `aFilePropertyStaysIdentifiableByItsLabelAndPattern` — the replacement signals are present: the label reaches `description` and the file-extension `pattern` is published. `FakeExcelLikeQuery` gains the `pattern` both real connectors declare, so this is exercised rather than asserted against an empty fixture.

---

## Verification

```
Tests run: 98, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

`TabularSchemaExtractorTest` (15, was 13), `TabularQueryParamsSchemaBuilderTest` (18), `TabularQueryParamsSchemaBuilderResolveTagsTest` (6), `TabularQueryContractSupportTest` (34), `WorksheetTableServiceLookupWiringTest` (2), `WorksheetTableServicePermissionTest` (23).

Verified against fixtures rather than real connectors: connector query classes live in plugin jars and are not on core's classpath, which is why this suite uses fixtures at all. Each fixture reproduces the shape that failed — a property placed inside a panel whose `visibleMethod` reads a list nothing settable grows, and a file property carrying the real extension pattern.

## Caller side

The four prompts and three comments that told the model to look for `format: "file-path"` / `"file-sheet"` are updated in `stylebi-wiz` (`Match a file target's properties by name, not by a role marker`) and have to ship together with part 2 — left as they were, they send the model looking for a field the contract no longer carries.

### DIFF
--- a/core/src/main/java/inetsoft/uql/tabular/TabularSchemaExtractor.java
+++ b/core/src/main/java/inetsoft/uql/tabular/TabularSchemaExtractor.java
@@ -199,6 +199,9 @@ public class TabularSchemaExtractor {
       // A property whose setter has already failed to return once is not offered to probe() again
       // -- see poison(). Removed from the list rather than skipped in the loop below, because the
       // second pass reads this same list to find its inner axes and must not reach it either.
+      //
+      // This covers what was poisoned BEFORE this call. An axis poisoned by the first pass below is
+      // still in the list, so the second pass checks again for itself.
       axes.removeIf(axis -> isPoisoned(pmap, axis.name));
 
       Probe baseline = probe(cls, dataSource, Collections.emptyMap());
@@ -231,7 +234,7 @@ public class TabularSchemaExtractor {
       }
 
       matrix.putAll(firstPass);
-      addCombinationGates(cls, dataSource, params, axes, firstPass, matrix);
+      addCombinationGates(cls, dataSource, params, axes, pmap, firstPass, matrix);
 
       return matrix;
    }
@@ -256,6 +259,7 @@ public class TabularSchemaExtractor {
     */
    private void addCombinationGates(Class<?> cls, XDataSource dataSource,
                                     List<TabularQuerySchema.Param> params, List<Axis> axes,
+                                    Map<String, PropertyMeta> pmap,
                                     Map<String, Map<String, List<String>>> firstPass,
                                     Map<String, Map<String, List<String>>> matrix)
    {
@@ -306,6 +310,18 @@ public class TabularSchemaExtractor {
                Map<String, List<String>> found = new LinkedHashMap<>();
 
                for(Object value : inner.values) {
+                  // Poison is checked HERE, per value, rather than being left to the list this
+                  // reads having been filtered. That filter ran before the first pass, and this is
+                  // where an axis hidden at baseline is first reached at all -- probe0 declines to
+                  // write it while its panel is shut, so the first pass never enters its setter and
+                  // never poisons it. What did not return is the PROPERTY, not one of its values,
+                  // so the remaining values are abandoned with it, and so is every later outer row
+                  // that reveals the same property. Without this, one broken setter strands a
+                  // thread per revealing axis instead of one for the life of the process.
+                  if(isPoisoned(pmap, inner.name)) {
+                     break;
+                  }
+
                   Map<String, Object> combo = new LinkedHashMap<>();
                   combo.put(outer.getKey(), outerValue);
                   combo.put(inner.name, value);
@@ -409,6 +425,12 @@ public class TabularSchemaExtractor {
          return null;
       }
       catch(InterruptedException ex) {
+         // Same signal as the timeout branch, for the same reason: if the setter really is stuck,
+         // the pooled thread is lost either way and an interrupt is the only thing left to offer a
+         // BLOCKING one. Not poisoned, though -- an interrupt says the waiter was told to stop, not
+         // that the setter would never have returned, and poisoning on that would drop an axis for
+         // the life of the process over a shutdown.
+         task.cancel(true);
          Thread.currentThread().interrupt();
          return null;
       }
@@ -422,10 +444,10 @@ public class TabularSchemaExtractor {
    /**
     * One probe, as it runs on the pool thread.
     *
-    * <p>Two rules govern the writes, and both are taken from the interactive path rather than
-    * invented here. A value is applied only where the layout does not place that parameter at all
-    * or currently shows it -- the rule {@link TabularUtil#setValuesToBean} follows when a dialog
-    * submits a form. And values are applied ONE AT A TIME with the conditions re-evaluated in
+    * <p>Two rules govern the writes, and both follow the interactive path's intent rather than
+    * being invented here. A value is applied only where the layout does not place that parameter at
+    * all or currently shows it -- the question {@link TabularUtil#setValuesToBean} asks when a
+    * dialog submits a form. And values are applied ONE AT A TIME with the conditions re-evaluated in
     * between -- what the dialog's refresh round trip does.
     *
     * <p>Both matter because a setter is not a pure function of its argument. {@link #isSelfGating}
@@ -438,9 +460,13 @@ public class TabularSchemaExtractor {
     * can work: its second parameter is by definition not shown until the first has been applied,
     * so a single combined write would silently drop it.
     *
-    * <p>The visibility consulted is the ancestor-aware one {@link #collectVisible} computes, not
-    * the per-node {@code TabularView#isVisible} flag. A field inside a hidden panel carries no
-    * flag of its own -- reading that flag alone would let exactly this case through.
+    * <p>The visibility question is answered STRICTLY here, and that is where this parts company
+    * with {@code setValuesToBean}: that method reads the per-node {@code TabularView#isVisible}
+    * flag, while this reads the ancestor-aware walk {@link #collectVisible} performs.
+    * {@code LayoutCreator} sets the flag true on every node it builds and {@code callVisibleMethod}
+    * does not propagate a false down, so a field inside a hidden panel reports itself visible --
+    * and the case that motivated this rule is exactly such a field. The per-node flag is the right
+    * question for a form the user was looking at; it is not enough for one nobody rendered.
     */
    private Probe probe0(Class<?> cls, XDataSource dataSource, Map<String, Object> values)
       throws Exception
--- a/core/src/main/java/inetsoft/uql/tabular/TabularSchemaExtractor.java
+++ b/core/src/main/java/inetsoft/uql/tabular/TabularSchemaExtractor.java
@@ -19,11 +19,15 @@ package inetsoft.uql.tabular;
 
 import inetsoft.uql.XDataSource;
 import inetsoft.uql.util.Config;
+import inetsoft.util.ThreadContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import java.lang.reflect.Method;
+import java.security.Principal;
 import java.util.*;
+import java.util.concurrent.*;
 
 /**
  * Derives a {@link TabularQuerySchema} — the parameters needed to build a query,
@@ -143,10 +147,7 @@ public class TabularSchemaExtractor {
 
       try {
          TabularView root = new LayoutCreator().createLayout(query);
-         TabularUtil.callViewMethods(root.getViews(), query);
-
-         Probe visible = new Probe();
-         collectVisible(root.getViews(), true, false, visible);
+         Probe visible = evaluate(root, query);
 
          for(String name : names) {
             // Only parameters the layout actually places can be judged. One the @View annotation
@@ -192,7 +193,14 @@ public class TabularSchemaExtractor {
       Class<?> cls, XDataSource dataSource, List<TabularQuerySchema.Param> params)
    {
       Map<String, Map<String, List<String>>> matrix = new LinkedHashMap<>();
+      Map<String, PropertyMeta> pmap = TabularUtil.getPropertyMap(cls);
       List<Axis> axes = findAxes(params);
+
+      // A property whose setter has already failed to return once is not offered to probe() again
+      // -- see poison(). Removed from the list rather than skipped in the loop below, because the
+      // second pass reads this same list to find its inner axes and must not reach it either.
+      axes.removeIf(axis -> isPoisoned(pmap, axis.name));
+
       Probe baseline = probe(cls, dataSource, Collections.emptyMap());
 
       if(axes.isEmpty() || baseline == null) {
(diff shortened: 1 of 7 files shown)


Nelson-cao (commented):
Reviewed against the full diff and the actual current source of the connectors this PR discusses (`RestJsonQuery`, `EndpointJsonQuery`, `ServerFileQuery`, `OneDriveQuery`), not just the PR description's own claims. Checked against this repo's standard review checklist (correctness, clean code, error handling/edge cases, readability, security, regressions, backward compatibility, test coverage).

**The reported production incident is genuinely fixed.** `RestJsonQuery.lookupIgnoreBaseUrl0..4` sits behind `visibleMethod = "isLookupVisible0"`, which only a button-grown list can turn on -- not a probeable axis -- so under the new write-gating rule the setter is never entered at all for that case, and the deadline/executor mechanism independently bounds the caller even if some other setter hangs.

**Part 2's diagnosis is accurate.** Confirmed directly against `ServerFileQuery` (both `excelSheet` and `encoding` are `String` with `dependsOn={"fileFolder"}` + `tagsMethod`, genuinely ambiguous under the old inference) and `OneDriveQuery` (file property is `String path`, not `java.io.File`, so neither old branch fired there either). Grepped all of `core` for any other reader of the removed `format: "file-path"`/`"file-sheet"` markers -- none found, so removing the inference is safe to merge standing alone in this repo; the noted `stylebi-wiz` prompt-sync dependency is a deployment-ordering concern, not a `core` correctness risk.

One important finding and a couple of nits below (inline). Two more notes that aren't inline because they're outside this diff:

- `OneDriveQuery.java:76-79` (pre-existing, not touched by this PR): `getExcelSheet()`'s `@PropertyEditor(dependsOn = {"fileFolder"}, ...)` references a property that doesn't exist on `OneDriveQuery` -- the file property here is named `path`, not `fileFolder`. Unrelated to this PR's correctness (it's the same connector cited as evidence that pattern-matching works, and that part still holds), but worth its own fix -- it likely means the Sheet dropdown never redraws on file change in the live dialog.
- `community/docs/tabular-unified-query-contract-backend-design.md` still documents the old `format: "file-path"`/`"file-sheet"` design and wasn't updated by this PR.

**Overall:** design instincts are right -- isolating untrusted plugin-jar reflection behind a bounded-wait executor, and deriving write-gating from the interactive dialog's own rules instead of inventing new ones. Test fixtures reproduce the real failure shapes with real assertions (write counts, matrix presence, non-reentry), not just shape checks. The one important finding below is a gap in the bound the new poisoning mechanism documents for itself, not a regression of the reported incident -- worth resolving before merge given this class is an explicit trust boundary into 89 uncontrolled plugin jars.

core/src/main/java/inetsoft/uql/tabular/TabularSchemaExtractor.java
```diff
@@ -362,45 +370,189 @@ private boolean isSelfGating(String axis, Probe baseline, Probe result) {
...
+    * Records that a setter did not return, so the property it belongs to is never probed again.
+    *
+    * <p>Keyed on the class that DECLARES the setter rather than on the query class being probed. A
+    * setter inherited by every REST connector is then one entry instead of one per connector, so
+    * the threads such a setter strands are bounded by how many broken setters exist rather than by
+    * how many data sources anyone happens to ask about.
+    *
+    * <p>Only the value applied LAST is blamed. {@link #addCombinationGates} applies its outer value
+    * first, and an outer axis has already completed a probe of its own in the first pass -- an axis
+    * whose setter hangs is poisoned and dropped before the pair pass can reach it. So when a
```
Nelson-cao, line 536: **Important: this invariant doesn't hold for an axis reachable as an *inner* axis, only for one reachable as an *outer* axis.**

`buildDependencyMatrix` filters `axes` against `POISONED` exactly once, before the first single-value pass runs (line 202). If some axis X's own probe hangs *during* that same first pass, `poison()` adds it to `POISONED` -- but the in-memory `axes` list already handed to `addCombinationGates` is never re-filtered. If X's property name also appears in another (successfully probed) axis Y's `conditionalVisible` list from the first pass, `addCombinationGates` still does `Axis inner = byName.get(gatedName)` and calls `probe(cls, dataSource, combo)` with X as the inner value (`TabularSchemaExtractor.java:299-312`) -- with no `isPoisoned` check first. If X hangs again, that's a second stranded pool thread in the same `extract()` call, and if X is gated by *multiple* outer axes from the first pass, it gets re-attempted once per such outer axis -- no "already timed out this call" memoization.

So "an axis whose setter hangs is poisoned and dropped before the pair pass can reach it" is only true when the hung axis is itself an outer axis of `firstPass` (which never appears in `addCombinationGates`'s inner-axis lookups by definition) -- not when it's revealed as *inner* by some other axis's gate list.

Note: this doesn't affect the motivating incident -- `RestJsonQuery.lookupIgnoreBaseUrl0..4` sits behind a button-only visibility gate, so it's never probed as either an outer or inner axis. This is a latent gap in a different connector shape (a hangable setter sitting in a panel revealed by another *settable* toggle, not a button), and it isn't covered by the new tests -- `StuckFixtureQuery.stuck` is unconditionally visible in the fixture, so it never appears as an inner axis of a combination gate.

Suggest either re-checking `isPoisoned(pmap, gatedName)` before the inner probe in `addCombinationGates`, or updating this javadoc to state the actual bound ("one thread per (broken setter x distinct revealing axis)" rather than "one thread per broken setter").

core/src/main/java/inetsoft/uql/tabular/TabularSchemaExtractor.java
```diff
@@ -362,45 +370,189 @@ private boolean isSelfGating(String axis, Probe baseline, Probe result) {
...
-         // and a probe built without one would answer for a query nobody is going to run.
-         if(dataSource != null && probe instanceof TabularQuery) {
-            ((TabularQuery) probe).setDataSource(dataSource);
-         }
+   /**
+    * One probe, as it runs on the pool thread.
+    *
+    * <p>Two rules govern the writes, and both are taken from the interactive path rather than
+    * invented here. A value is applied only where the layout does not place that parameter at all
+    * or currently shows it -- the rule {@link TabularUtil#setValuesToBean} follows when a dialog
```
Nelson-cao: Nit: this overstates parity with `TabularUtil.setValuesToBean`. The real `setValuesToBean` gates on the per-node `TabularView.isVisible()` flag, not the ancestor-aware `collectVisible` walk this method actually uses -- per `LayoutCreator`, a field inside a hidden panel carries no flag of its own and would report `isVisible()==true` under the per-node flag alone (which is exactly the case this class needs to exclude, per the next paragraph's own reasoning). `probe0`'s rule is stricter/safer than `setValuesToBean`'s, not identical to it -- worth rewording so a future reader trying to verify equivalence isn't misled into thinking the two already agree.

core/src/test/java/inetsoft/uql/tabular/TabularSchemaExtractorTest.java
```diff
@@ -64,6 +65,16 @@ static void clearContext() {
       }
    }
 
+   /**
+    * Both the deadline and the record of which setters have hung are per-process, so a case that
+    * changes either has to hand them back -- otherwise it decides the next one.
+    */
+   @AfterEach
+   void restoreProbeDefaults() {
+      TabularSchemaExtractor.setProbeTimeout(2000);
```
Nelson-cao, line 81: Nit: `probeTimeoutMs`/`POISONED` are process-static state, reset only via this `@AfterEach`. Fine under the apparent sequential default here, but would be flaky if this module's tests are ever run in parallel (one test's 150ms timeout, or its poisoned keys, leaking into a concurrently running test).

core/src/main/java/inetsoft/uql/tabular/TabularSchemaExtractor.java
```diff
@@ -192,7 +193,14 @@ Map<String, Map<String, List<String>>> buildDependencyMatrix(
       Class<?> cls, XDataSource dataSource, List<TabularQuerySchema.Param> params)
    {
       Map<String, Map<String, List<String>>> matrix = new LinkedHashMap<>();
+      Map<String, PropertyMeta> pmap = TabularUtil.getPropertyMap(cls);
       List<Axis> axes = findAxes(params);
+
+      // A property whose setter has already failed to return once is not offered to probe() again
+      // -- see poison(). Removed from the list rather than skipped in the loop below, because the
+      // second pass reads this same list to find its inner axes and must not reach it either.
+      axes.removeIf(axis -> isPoisoned(pmap, axis.name));
```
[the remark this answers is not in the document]
Nelson-cao: Fixed in f12215018 — you and the later review round both landed on this, and it reproduced.

`isPoisoned(pmap, inner.name)` is now checked per value INSIDE `for(Object value : inner.values)`, not once before the loop, which is what your "ideally re-checked between combo probes" asked for: what did not return is the property, so the remaining values are abandoned with it and so is every later outer row naming it.

New test `anAxisPoisonedInTheFirstPassIsNotProbedAgainAsAnInnerAxis` with a fixture whose hanging setter sits in a panel that TWO settable axes open (plus a two-value gate below it, so `unreached` is non-empty and the pair pass runs at all). Verified it fails without the check: `expected: <1> but was: <2>`.

core/src/main/java/inetsoft/uql/tabular/TabularSchemaExtractor.java
```diff
@@ -362,45 +370,189 @@ private boolean isSelfGating(String axis, Probe baseline, Probe result) {
...
+         return task.get(probeTimeoutMs, TimeUnit.MILLISECONDS);
+      }
+      catch(TimeoutException ex) {
+         // Best-effort: a setter looping with no blocking call in it offers the interrupt nowhere
+         // to land, and Thread.stop is gone. What bounds the damage is poison(), not this.
+         task.cancel(true);
+         poison(cls, values);
+         return null;
+      }
+      catch(InterruptedException ex) {
```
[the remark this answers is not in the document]
Nelson-cao: Fixed in f12215018. `task.cancel(true)` now runs in the `InterruptedException` branch too, with a comment recording the asymmetry you identified: the interrupt is the only thing left to offer a BLOCKING hang, and poisoning stays out of this branch deliberately — an interrupt says the waiter was told to stop, not that the setter would never have returned, and poisoning on that would drop an axis for the life of the process over a shutdown.

core/src/test/java/inetsoft/uql/tabular/TabularSchemaExtractorTest.java
```diff
@@ -64,6 +65,16 @@ static void clearContext() {
...
+      TabularSchemaExtractor.clearPoisoned();
```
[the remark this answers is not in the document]
Nelson-cao: Fixed in f12215018 with a class-level `@ResourceLock("TabularSchemaExtractor-process-state")`. The class writes two pieces of process-wide state, not one — the `ConfigurationContext` the `@BeforeAll` installs as well as the deadline and poison record — and neither can be made per-test, since the poison record is process-wide by design (what it bounds is threads this process cannot reclaim). Serializing the class against anything else claiming that resource is the honest fix; the comment says why it cannot be scoped down instead.

core/src/main/java/inetsoft/uql/tabular/TabularSchemaExtractor.java
Nelson-cao: Confirmed and fixed in f12215018. Your reading of the mechanism is exact, including that the motivating incident is not affected — `lookupIgnoreBaseUrl0..4` sits behind a button-only gate and is never an axis of either kind.

Took the fix rather than narrowing the javadoc, and put the check per value inside `for(Object value : inner.values)` rather than once before it, so a poison recorded mid-loop also stops the remaining values of that same axis — not only later outer rows.

Also took the point that the existing fixtures do not cover the shape: `anAxisPoisonedInTheFirstPassIsNotProbedAgainAsAnInnerAxis` adds a fixture whose hanging setter is revealed by two different settable toggles, with a two-value gate below it so the pair pass runs. It fails without the check (`expected: <1> but was: <2>`).

core/src/main/java/inetsoft/uql/tabular/TabularSchemaExtractor.java
Nelson-cao: Fixed in f12215018. Reworded to say the two agree on the QUESTION rather than the answer, and the following paragraph now states outright where they part company and why: `setValuesToBean` reads the per-node `TabularView.isVisible` flag, `LayoutCreator` sets that flag true on every node it builds and `callVisibleMethod` does not propagate a false down, so a field inside a hidden panel reports itself visible — which is exactly the field that motivated the rule. The per-node flag is the right question for a form the user was looking at; it is not enough for one nobody rendered.

core/src/test/java/inetsoft/uql/tabular/TabularSchemaExtractorTest.java
Nelson-cao: Fixed in f12215018 — class-level `@ResourceLock`. See the reply on the same point from the earlier round for the detail: there are two pieces of process-wide state here, not one, and the poison record is process-wide by design.

Nelson-cao:
## Out-of-diff notes from the review, and one found while acting on them

**Stale doc — fixed.** `docs/tabular-unified-query-contract-backend-design.md` section 3.8 recorded the premise part 2 disproves: *"excelSheet is the unique property matching all three"* on `ServerFileQuery`. Corrected in `f12215018` with what it actually declares.

**NOT changed, deliberately: `findSheetParam` shares the heuristic and resolves it by layout order.** While correcting that doc I found the removed role-marker inference is not its only consumer. `TabularQueryContractSupport.findSheetParam` (capability 5b, the Excel multi-sheet ambiguity refusal) applies the identical three conditions:

```java
if("java.lang.String".equals(p.getJavaType()) &&
   p.getTagsMethod() != null && !p.getTagsMethod().isEmpty() &&
   p.getDependsOn() != null && p.getDependsOn().contains(fileParamName))
{
   return p;
}
```

It does not refuse on ambiguity — it returns the FIRST match in `schema.getParams()` order, which is `@View` presentation order. `ServerFileQuery`'s `@View` lists `excelSheet` (row 2) above `encoding` (row 3), so it names the right property today **by layout order, not because the heuristic is decisive**. Reorder that `@View`, or add a connector that lists a charset picker first, and the refusal tells the caller to `Supply it as queryParams.encoding` to pick a sheet — wrong advice, no error.

Left alone in this PR for two reasons: it is a different capability in a file this diff does not touch, and the message it produces is a cross-repo contract that `tabularFileClient.ts` parses by regex, so changing which property it names is not a local decision. Recorded in the design-doc correction so it is written down rather than carried in someone's head.

**`OneDriveQuery.getExcelSheet()`'s dangling `dependsOn`** — pre-existing and out of scope, agreed. Six `@PropertyEditor(dependsOn = {"fileFolder"})` on that class reference a property it does not declare (its file property is `path`); the leftovers are from `ServerFileQuery`. Its live effect is on that connector's own editor refresh, not on the contract.

## Verification

```
Tests run: 99, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

The new case was checked against its own absence: with the `isPoisoned` guard removed it fails `expected: <1> but was: <2>`, both before and after the guard was moved inside the values loop.

Nelson-cao:
Re-review of `f12215018`: all three findings from the prior round verified fixed by reading the actual diff, not just trusting the reply.

- The important finding (poison bound not holding for an inner axis) is fixed correctly: `isPoisoned(pmap, inner.name)` is checked per-value inside the inner values loop and `break`s on hit, so a property poisoned mid-call is abandoned for its remaining values AND for every later outer row that reveals it -- not just the one in progress. New test `anAxisPoisonedInTheFirstPassIsNotProbedAgainAsAnInnerAxis` with `PairStuckFixtureQuery` genuinely exercises this: `stuck` is reachable as an inner axis under two different outer axes, and `deep`'s two-value gate is what forces the pair pass to run at all. Confirmed the fixture shape and assertion (`entered` count == 1) actually distinguish before/after.
- Both nits (the `probe0`/`setValuesToBean` javadoc overstatement, and the process-static test state) are fixed as described.
- The design-doc correction is honest about the prior wrong claim, and the additionally-discovered `findSheetParam` ambiguity gap is reasonably left out of this PR (different capability, untouched file, cross-repo contract) and is documented for follow-up.

No new issues found in the fix commit itself. Resolved the three threads I filed. Verdict: safe to merge.

jshobe-inetsoft (approved):
## Manual review (replaces the automated `claude-review` check, which errored out during setup — not a review finding, see [run 33059536030](https://github.com/inetsoft-technology/stylebi/actions/runs/33059536030/job/98474590402), `is_error: true`, 8 permission denials)

### Summary

Two independent, well-isolated fixes in `TabularSchemaExtractor`/`TabularQueryParamsSchemaBuilder`:

1. **Probe deadline + write-gating.** `probe()` now runs each connector-setter write on a pooled daemon thread under a 2s deadline (`probeTimeoutMs`), and a setter that never returns is recorded once in a process-wide `POISONED` set (keyed by the *declaring* class of the setter, so an inherited setter is one entry, not one per connector subclass) so it's never probed again. Writes are now also gated the same way the interactive dialog gates them — only apply a value where the layout doesn't place the param at all, or where it's currently visible per the ancestor-aware `collectVisible` walk (not the per-node `isVisible` flag, which `LayoutCreator`/`callVisibleMethod` don't propagate down to children) — and applied one at a time with re-evaluation between writes, matching what `addCombinationGates`'s two-level gating already requires.

2. **Sheet/file identification by name instead of inferred role.** Removes `applyRoleFormat`'s `format: "file-path"/"file-sheet"` stamping, which infers a "sheet selector" from `String + tagsMethod + dependsOn(file property)` — a signature that (per the PR's repro) also matches a charset/encoding picker on the one connector that has both, and doesn't match at all on the connector whose file property is a plain `String`. The 500 that motivated this is real and reproduced in the PR description. Callers now match on name/`description` (the `@Property` label)/`pattern`, which are already published per-property.

### Correctness

- `poison()` blaming only the *last* key in the `values` map is sound given the invariant documented above it: values are only ever probed as singletons (baseline/first pass) or as `{outer, inner}` pairs built via `LinkedHashMap` with outer inserted first (`addCombinationGates`), and an outer axis has already completed a solo probe in the first pass before it can appear as the first entry of a pair — so if a two-entry probe times out, the outer setter is already known-good and the inner is the one that hung. This is a real assumption (a setter's timing isn't provably deterministic across calls) but it's explicitly called out in the javadoc rather than silently assumed, and `MAX_NESTING = 2` keeps the map to at most two entries, so there's no untested N>2 case.
- The `isPoisoned` filter is applied in three places that all need it independently and correctly: once on `axes` before the first pass (covers poisoning from prior calls/process history), once per-value inside `addCombinationGates`'s inner loop (covers poisoning that happens *during* the current `extract()` call, before the pre-filtered list can account for it) — and the new test `anAxisPoisonedInTheFirstPassIsNotProbedAgainAsAnInnerAxis` specifically exercises the case where filtering the list up front wouldn't have been enough.
- `findInapplicable`'s refactor to `evaluate(root, query)` is a straight extraction — behaviorally identical to the original inline `callViewMethods` + `collectVisible` sequence.
- `ThreadContext` principal propagation to the pool thread, and clearing it in a `finally`, looks correct; since `PROBES` is a cached pool, a still-running (stuck) thread is never idle and so won't be handed to another submission before its principal is cleared.
- Removing the `format` role markers is a straight deletion of the buggy inference with no replacement inference — matches the "identify by name" fix described, and the design doc correction (section on `findSheetParam`) documents the same root cause for the sibling consumer in `TabularQueryContractSupport`, which is *not* touched by this PR (only the doc is corrected to say it currently works by `@View` ordering luck, not by the heuristic being decisive).

### Cross-repo coordination (flagging, not blocking)

The PR body notes the `stylebi-wiz` prompts/specs that reference `format: "file-path"/"file-sheet"` need a companion update "to ship together with part 2" — since those live outside this repo I can't verify that side, but wanted to surface it since merging this without the companion change would send that caller looking for a field the contract no longer carries.

### Tests

New coverage directly targets both described failure modes (`OffScreenFixtureQuery`/`aValueTheLayoutDoesNotShowIsNeverWritten`, `StuckFixtureQuery`/`aSetterThatDoesNotReturnIsAbandonedAndNotProbedAgain`, `PairStuckFixtureQuery`/`anAxisPoisonedInTheFirstPassIsNotProbedAgainAsAnInnerAxis`) plus the role-format removal (`noPropertyCarriesARoleFormat`, `aFilePropertyStaysIdentifiableByItsLabelAndPattern`). `@AfterEach` correctly resets the process-wide `probeTimeoutMs`/`POISONED` state, and `@ResourceLock` serializes the test class against that shared state. `build` check on this PR is green (98 tests, 0 failures per the PR description).

### Verdict

No blocking issues found. Other checks (`build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla`) are all green, and the PR reports `MERGEABLE` with no conflicts. **Approving** — the failing `claude-review` check was an infrastructure error in the automated action, not a review finding, and this manual review takes its place.

PR #4824, opened 2026-08-27
Bug #75980, add N-ary join support to the MCP worksheet add_join op

jkidacki-inetsoft:
## Summary

The MCP worksheet agent's `add_join` operation could only join two tables at a time. Joining 3+ tables required chaining separate `add_join` calls into nested assemblies instead of a single combined view -- unlike Composer's own UI, which lets a user multi-select several tables and join them all at once.

## Root Cause

`WorksheetEditService.Editor.addJoin(...)` (server side of the agent-edit `add_join` op) always built a `RelationalJoinTableAssembly` over exactly `new TableAssembly[]{ left, right }`. Composer's own N-ary join capability (`WSJoinTablesEvent` -> `InnerJoinService.joinTables()` -> `InnerJoinService.editExistingJoinTable(...)`) was never wired to the agent-edit path -- the existing code comment claimed `InnerJoinService` "requires a live STOMP/runtime context and is not usable here," which is true for most of its methods but not for `editExistingJoinTable(Worksheet, RelationalJoinTableAssembly, TableAssemblyOperator, boolean)`, which takes a plain `Worksheet` and no session context. `WorksheetTableService.buildJoinTable` already calls it this exact way (for the `add_table` "relational join table" type), proving the mechanism is safe to reuse outside a live session.

## Fix

- New `WorksheetMutationSupport.JoinPathSpec` record describing one join edge (leftTable, leftKey, rightTable, rightKey, joinType).
- New `EditRequest.joinPaths` field (optional, appended via the existing compat-constructor pattern so no existing caller needs to change).
- New `Editor.addJoin(String name, List<JoinPathSpec> joinPaths)` overload: builds a `RelationalJoinTableAssembly` over every table named across the given edges, then delegates to `InnerJoinService.editExistingJoinTable(...)` to wire the real per-edge operators -- the same mechanism behind Composer's own N-ary join. Edges do not need to form a linear chain (a hub table can join to two others in one call).
- `WorksheetAgentController.dispatch()`'s `case "add_join"` now branches on `req.joinPaths()`: routes to the new N-ary method when present, otherwise keeps the existing pairwise behavior unchanged.
- A `CROSS` edge is only valid as the sole edge in `joinPaths`, since it is an exclusive operation server-side (`TableAssemblyOperator.checkValidity` rejects it once more than one edge is present) -- rejected with a clear message otherwise. `MERGE` is rejected per edge (structurally distinct assembly type; use `add_merge_join`). On any wiring failure after the assembly is registered, it is removed again rather than left behind broken.

Companion client-side change (MCP tool schema + precheck): https://github.com/inetsoft-technology/stylebi-wiz/pull/new/bug-75980

## Test Plan

- New unit tests in `WorksheetEditServiceTest.java`: 3-table success case, star-shaped/non-chain join (hub table joined to two others), MERGE rejection, CROSS-combined-with-other-edges rejection, sole-CROSS-edge success, empty-path-list rejection, and rollback verification when `InnerJoinService.editExistingJoinTable` fails after the assembly is already registered.
- `./mvnw test -pl core -Dtest=inetsoft.web.wiz.**`: 2598/2598 passing (1 pre-existing unrelated skip).
- `./mvnw clean install -pl core -am -DskipTests`: clean build.
- Reviewed via code-reviewer subagent (2 rounds); both findings from round 1 (CROSS-exclusivity validity check, missing rollback on wiring failure) fixed and covered by new tests before round 2.
- Live-verified against a running StyleBI instance with the companion stylebi-wiz plugin build.

Fixes Redmine #75980.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
@@ -48,7 +48,7 @@ import java.util.Map;
  *       error from this call.</li>
  *   <li>{@code add_expression_column} — {@code table}, {@code name}, {@code expression}, {@code type}, {@code sql}</li>
  *   <li>{@code set_sort} — {@code table}, {@code field}, {@code direction} ("ASC" | "DESC")</li>
- *   <li>{@code add_join} — {@code name}, {@code leftTable}, {@code leftKey}, {@code rightTable}, {@code rightKey}, {@code joinType}; for multi-key joins use {@code leftKeys}/{@code rightKeys} instead of single key fields</li>
+ *   <li>{@code add_join} — {@code name}, {@code leftTable}, {@code leftKey}, {@code rightTable}, {@code rightKey}, {@code joinType}; for multi-key joins use {@code leftKeys}/{@code rightKeys} instead of single key fields. For three or more tables joined in a single call, supply {@code joinPaths} instead (each a {leftTable, leftKey, rightTable, rightKey, joinType} edge) — {@code leftTable}/{@code leftKey}/{@code rightTable}/{@code rightKey}/{@code joinType}/{@code leftKeys}/{@code rightKeys} are ignored when {@code joinPaths} is present</li>
  *   <li>{@code remove_join} — {@code name}</li>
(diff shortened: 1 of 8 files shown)


jkidacki-inetsoft:
## Review

Design checks out against the actual source, not just the PR description. `InnerJoinService.editExistingJoinTable` really is safe to call outside a live session — it's the exact same call `WorksheetTableService.buildJoinTable` already makes for `add_table`'s relational-join-table path, so this isn't a new usage pattern.

**Verified correct:**
- Validation (MERGE rejection, CROSS-combined-with-other-edges rejection) runs fully before any worksheet mutation — `addJoinWithPathsRejectsCrossCombinedWithOtherEdges` confirms a rejected call leaves zero trace (`assertEquals(3, ws.getAssemblies().length, ...)`).
- Rollback on wiring failure (`ws.removeAssembly(name)` in the catch block) is correct and tested.
- `CROSS_JOIN` permission check happens before the assembly is created — no side effect on an unauthorized attempt.
- Traced the star-join test against `editExistingJoinTable`'s real implementation — it resolves edges by table name, not position, so the "edges don't need to form a linear chain" claim is accurate.
- 7 new tests cover exactly what the description claims (3-table, star/hub, MERGE rejection, CROSS-combined rejection, sole-CROSS success, empty-list rejection, rollback-on-failure).

**One gap, not a regression:** a null `leftKey`/`rightKey` on a non-CROSS edge isn't validated server-side (builds `new AttributeRef(null, null)`). Checked — the pre-existing two-table `addJoin` has the identical gap, so this is consistent with, not worse than, established precedent. The companion stylebi-wiz#1888 plugin-side precheck closes it in practice for the only real caller.

**Not a finding, but noting for visibility:** the `claude-review` CI check shows "fail" — pulled the run log, it's an infra issue (`ANTHROPIC_API_KEY` empty in the runner env, job dies after 1 turn/$0 cost), not a real review finding. The real `build` check (mvnw) passed.

LGTM.

ngovus-inetsoft (approved):

PR #4825, opened 2026-08-27
Bug #76170, withhold the share webhook URLs and re-verify the fluentd password

ngovus-inetsoft:
## Summary

Two gaps in `AdminPropertyCatalog`'s definition of a secret, from Redmine #76170. Both were found by re-reading each property's **writer** rather than trusting the existing lists — the rule this class already documents.

Part 1 of that issue concerns `main` and shipped separately as #4817. `AdminPropertyCatalog` does not exist on `main` (the whole `web/admin/ai/` package is unmerged feature work), so this half can only target `stylebi-wiz-main-rebase`.

## 1. The two share webhook URLs were returned in full

`share.slack.url` and `share.googlechat.url` are incoming-webhook URLs: the token is in the path, no second factor, no request signature. Anything holding the URL can post into that channel or room.

Neither matches the name shape, and neither is written through an encrypting accessor — `ShareSettingsService` writes both with a plain `setProperty`. So **both existing tests miss them by construction**: #76006's remedy is an allow-list of properties whose writer encrypts. `AdminPropertiesController` read the full URL and returned it to a caller that relays responses to a model provider off-host.

`ENCRYPTED_CREDENTIALS` is the wrong home: it drives `SreeEnv.setPassword` on write, which would store ciphertext under a property `ShareSettingsService` and `ShareController` both read as a literal URL — the mirror of the plaintext downgrade that list exists to prevent. This adds a third hand-verified set, `CONFIRMED_SECRET`, unioned into `isSecret`. The two lists answer different questions — *does the product encrypt this value* vs *does possessing it grant access* — and keeping them separate is what lets each stay checkable.

**Accepted cost, recorded in the javadoc:** `AdminChangePlanService` refuses a change when a name is secret and not an encrypted credential, so these two are no longer settable through admin-chat and must be configured in Enterprise Manager. Encrypting them at rest instead would keep the write, at the price of changing a shipped feature and migrating values already stored in the clear.

`mapbox.token` was considered and left out — like `google.maps.key` it is a client-side map token, ordinarily restricted by referrer.

Masking must not delete a true answer, so `CONFIRMED_SECRET` gets the same term in the `confirmed` expression the credential list already has. Both URLs are declared in `defaults.properties`, so the defaults chain made `stored` non-null and they reported `EXISTS_CONFIRMED`; without the term they regress to `EXISTS_UNKNOWN`, whose guidance says the name may be a typo and that an unrecognised name is written verbatim while reporting success — all false here. That is the regression the credential term was added for in #76006.

## 2. `log.fluentd.security.password` was misclassified

Bug #76051 (`dc8877f8a`) routed `LogSettingService`'s write of it through the same `toPassword` → `Tool.encryptPassword` helper as its sibling shared key. The entry never followed, and the javadoc still used the property as its worked example of a value written in the clear — as did two tests.

Nothing leaked: the name contains `password`, so `isSecret` always matched. The cost was that admin-chat refused to set a property it could have set correctly. The "classify by the writer" rule is unchanged; the paragraph arguing for it keeps its example, now `mail.smtp.tokenuri`, which is read with `getPassword` and written with a bare `setProperty`.

## 3. …which exposed a plaintext echo, fixed here too

Making that property settable opened a real egress path. `AdminChangeService` keeps `beforeValue`/`afterValue` as the stored form, justified on the grounds that a credential is stored as ciphertext and ciphertext is not the secret. That holds only *normally*: `Tool.decryptPassword` passes plaintext straight through, so a credential can sit in the store in the clear and still work — exactly what this property did until #76051, and what `INETSOFTENV_*` promotion and the raw EM properties page still do for any of them. On an upgraded server the first apply would have echoed the old plaintext credential into the apply response, off-host.

`ApplyOutcome` now reports `(set)` for an encrypted credential. Masked at the **response DTO** and not in `AdminChangeResult` deliberately: rollback replays `applied.getBeforeValue()` and the `moved` comparison needs the real values, and both read the result object directly, so neither is affected. A `null` stays `null`, so "no prior credential" remains distinguishable from one that was replaced.

## Also

Drops the hand-maintained "fifteen properties" count from the `isSecret` javadoc rather than computing a third value. It disagreed with the "fourteen" in the corresponding test, and both had already been left behind by the three `CONFIRMED_NOT_SECRET` carve-outs before this change touched the group again.

The disjointness guard now reflects over both sets and asserts `Collections.disjoint`. Asserting it through `isSecret` over hardcoded names looked equivalent and was not: `isSecret` returns early on `CONFIRMED_NOT_SECRET`, so adding one of those names to `CONFIRMED_SECRET` left the assertion passing — green on precisely the contradiction it claims to catch.

## Testing

`202 tests, 0 failures, 0 errors` — the whole `inetsoft.web.admin.ai` package plus `LogSettingServiceTest` and `PropertiesControllerTest`:

```
./mvnw test -pl core -Dtest='inetsoft.web.admin.ai.*Test,LogSettingServiceTest,PropertiesControllerTest'
```

The fluentd reclassification broke two existing tests that used the property as their example of a secret-named non-credential. Both now use a property whose writer was re-checked (`sso.rsa.private.key`, `license.key`) and record why the old example expired — that breakage is the evidence the entry really had gone stale.

New coverage: both webhook URLs withheld and never read at all; their `exists`/description staying truthful; the `share.*` flags beside them staying readable; the fluentd password now written through `setPassword`; credential values masked in the apply response while an ordinary property's are not; and rollback still replaying the real stored value rather than the mask.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminChangesetApplyService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminChangesetApplyService.java
@@ -122,9 +122,30 @@ public class AdminChangesetApplyService {
                           change.proposedValue(), backupRef, req.getReviewOutcome()),
                   user);
 
-               results.add(new ApplyOutcome(change.property(), applied.getBeforeValue(),
-                                            applied.getAfterValue(), applied.getStatus(),
-                                            applied.getError()));
+               // before/after are the STORED form, and for a credential that is normally
+               // ciphertext - which AdminChangeService argues carries nothing sensitive. It is
+               // only normally: Tool.decryptPassword passes plaintext straight through, so a
+               // credential can sit in the store in the clear and still work. That is not
+               // hypothetical - log.fluentd.security.password was written in the clear by the
+               // product's own writer until Redmine #76051, and INETSOFTENV_* promotion
+               // (StorageInitializer) and the raw EM properties page both still bypass every
+               // encrypting writer. So on an upgraded server the first apply would echo the old
+               // plaintext credential into this response, which the caller relays to a model
+               // provider off-host - the same egress this component masks reads for.
+               //
+               // Masked HERE and not in AdminChangeResult, which is the seam that matters:
+               // rollback replays applied.getBeforeValue() into undoableBefore below, and the
+               // `moved` comparison needs the real values too. Both read the result directly, so
+               // neither is affected. Only this response DTO is.
+               boolean maskValues = AdminPropertyCatalog.isEncryptedCredential(
+                  AdminPropertyName.parse(change.property()).baseName());
+
+               results.add(new ApplyOutcome(change.property(),
+                                            maskValues ? setMarker(applied.getBeforeValue())
+                                               : applied.getBeforeValue(),
+                                            maskValues ? setMarker(applied.getAfterValue())
+                                               : applied.getAfterValue(),
+                                            applied.getStatus(), applied.getError()));
 
                boolean verified = AdminChangeRecord.STATUS_VERIFIED.equals(applied.getStatus());
                // Path A (AdminChangeService): SreeEnv.save() can succeed and then a side-effect hook
@@ -232,6 +253,15 @@ public class AdminChangesetApplyService {
       req.setBackupRef(backupRef);
       req.setReviewOutcome(reviewOutcome);
       return req;
+   }
+
+   /**
+    * The credential stand-in for {@link ApplyOutcome}: reports whether a value is present without
+    * reproducing it. Matches {@code AdminChangePlanService}'s wording for a masked currentValue, so
+    * a caller sees the same vocabulary at preview and at apply.
+    */
+   private static String setMarker(String value) {
+      return value == null ? null : "(set)";
    }
 
    private static String messageOf(Exception e) {
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminChangesetApplyService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminChangesetApplyService.java
@@ -17,6 +17,7 @@
  */
 package inetsoft.web.admin.ai;
 
+import inetsoft.util.Tool;
 import inetsoft.util.audit.AdminChangeRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -259,9 +260,19 @@ public class AdminChangesetApplyService {
     * The credential stand-in for {@link ApplyOutcome}: reports whether a value is present without
     * reproducing it. Matches {@code AdminChangePlanService}'s wording for a masked currentValue, so
     * a caller sees the same vocabulary at preview and at apply.
+    *
+    * <p>Empty passes through as itself, not as {@code (set)}. A blank string is not a credential
+    * anyone holds, so reporting one as present would be a false answer with the same shape as the
+    * one this whole component exists to avoid. Unreachable through the plan path -
+    * {@code AdminChangePlanService} refuses an empty value for a credential and
+    * {@code SreeEnv.setPassword} guards on {@code Tool.isEmptyString} - but the comment at the
+    * masking site names {@code INETSOFTENV_*} promotion and the raw EM properties page as writers
+    * that reach the store past every one of those guards, and either can leave {@code ""} behind.
+    * {@code null} likewise stays {@code null}, so "no prior credential" remains distinguishable
+    * from one that was replaced.
     */
    private static String setMarker(String value) {
-      return value == null ? null : "(set)";
+      return Tool.isEmptyString(value) ? value : "(set)";
    }
 
    private static String messageOf(Exception e) {
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminPropertiesController.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminPropertiesController.java
@@ -106,6 +106,7 @@ public class AdminPropertiesController {
       AdminRiskClassifier.RiskClassification risk = classifier.classify(name);
       String description = entry == null ? null : entry.description();
       boolean credential = AdminPropertyCatalog.isEncryptedCredential(name.baseName());
+      boolean confirmedSecret = AdminPropertyCatalog.isConfirmedSecret(name.baseName());
       boolean secret = AdminPropertyCatalog.isSecret(name.baseName());
       // A secret's value is not read AT ALL, not merely omitted from the response - see the test
       // that asserts SreeEnv is never called for one. Determining existence by reading it would
@@ -146,6 +147,18 @@ public class AdminPropertiesController {
             + "exactly as the Enterprise Manager field does. You will not be able to verify the "
             + "new value by reading it back.";
       }
+      else if(confirmedSecret) {
+         // The pattern wording below would be false twice over for these: the name does NOT match
+         // the pattern (a test asserts that), and membership in a hand-verified list IS evidence
+         // the property exists. Telling the caller "this says nothing about whether the property
+         // exists" would throw away a true answer, which is the same mistake the credential branch
+         // above was added to stop.
+         description = "This is a bearer credential - possession of the value is enough to use it, "
+            + "so it is not read here, not even to test whether one is set, because this endpoint's "
+            + "caller relays responses to a model provider off-host. The property exists. It cannot "
+            + "be changed here either: nothing in the product encrypts it, so admin-chat will not "
+            + "write it. Configure it through Enterprise Manager.";
+      }
       else if(secret) {
          description = "This name matches admin-chat's secret pattern, so its value is neither "
             + "read nor changed here - not even read to test whether one is present, so this says "
@@ -165,7 +178,15 @@ public class AdminPropertiesController {
       //
       // A name that is secret by PATTERN alone gets no such term and stays unknown. All that is
       // known there is that a string matched, which is exactly what the guidance says.
-      boolean confirmed = entry != null || credential || stored != null;
+      //
+      // CONFIRMED_SECRET needs the same term as the credential list, and for the same reason
+      // (Redmine #76170). Both webhook URLs are declared in defaults.properties, so the defaults
+      // chain made `stored` non-null and they reported confirmed until this path stopped reading
+      // them. Without this term they regress to unknown, whose guidance tells the caller the name
+      // may be a typo and that an unrecognised name is written verbatim and reports success -
+      // three claims that are all false here, since AdminChangePlanService refuses the change.
+      // They are verified against their writer exactly as the credential list is.
+      boolean confirmed = entry != null || credential || confirmedSecret || stored != null;
 
       return new PropertyView(name.key(),
          entry == null ? List.of() : (entry.aliases() == null ? List.of() : entry.aliases()),
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminPropertyCatalog.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminPropertyCatalog.java
@@ -208,10 +208,13 @@ public class AdminPropertyCatalog {
     * secret value read through this path leaves the host in a way the ordinary EM properties page
     * never does.
     *
-    * <p>Two independent tests, unioned, then a name-shape exception list subtracted. A name
+    * <p>Four independent tests, unioned, then a name-shape exception list subtracted. A name
     * matches case-insensitively when it contains {@code password}, {@code secret} or
-    * {@code credential}, ends with {@code .key}, or starts with {@code license.}; and
-    * {@link #isEncryptedCredential} matches the properties whose <b>writer</b> encrypts them.
+    * {@code credential}, ends with {@code .key}, or starts with {@code license.};
+    * {@link #isEncryptedCredential} matches the properties whose <b>writer</b> encrypts them;
+    * {@link #CONFIRMED_SECRET} names the credentials neither of those two reaches; and
+    * {@link #COMPOSITE_SECRET_PROPERTIES} names the properties that carry a credential inside a
+    * larger composite string.
     * {@link #CONFIRMED_NOT_SECRET} then removes the specific names the shape test catches for the
     * wrong reason - verified individually, the same way {@link #ENCRYPTED_CREDENTIALS} is, rather
     * than by loosening the shared pattern:
@@ -253,9 +256,13 @@ public class AdminPropertyCatalog {
     * it is a credential by the product's own action, and the next one added is masked on read
     * without a second edit.
     *
-    * <p>The name test stays. It covers fifteen properties nothing writes through an encrypting
+    * <p>The name test stays. It covers the properties nothing writes through an encrypting
     * accessor — {@code license.*}, {@code jwt.signing.key}, {@code password.encryption.key},
     * {@code sso.rsa.private.key} among them — which the writer test would silently unmask.
+    * Deliberately no count: the previous wording said "fifteen", derived from the sixteen names
+    * the pre-#76006 pattern withheld, and was left behind by both the three carve-outs below and
+    * the reclassification in Redmine #76170. A number maintained by hand beside three sets that
+    * each move independently is one more fact to go stale, and this method has enough of those.
     *
     * <p>Masking a credential here does <b>not</b> make it unwritable: {@code AdminChangePlanService}
     * refuses a change only when a name is secret AND not an encrypted credential, and the egress
@@ -272,7 +279,8 @@ public class AdminPropertyCatalog {
          return false;
       }
 
-      return isEncryptedCredential(lower) || COMPOSITE_SECRET_PROPERTIES.contains(lower) ||
+      return CONFIRMED_SECRET.contains(lower) || isEncryptedCredential(lower) ||
+         COMPOSITE_SECRET_PROPERTIES.contains(lower) ||
          lower.contains("password") || lower.contains("secret") ||
          lower.contains("credential") || lower.endsWith(".key") || lower.startsWith("license.");
    }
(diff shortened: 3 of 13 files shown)


jshobe-inetsoft (changes requested):
## Review of PR #4825 — Bug #76170, withhold the share webhook URLs and re-verify the fluentd password

### Summary

This PR closes two gaps in `AdminPropertyCatalog`'s definition of a "secret" for the admin-chat property surface:

1. `share.slack.url` / `share.googlechat.url` (incoming-webhook URLs — bearer credentials by possession, not by name shape or storage form) were being returned in full to a caller that relays responses to an off-host model provider. A new hand-verified `CONFIRMED_SECRET` set closes this, kept deliberately separate from `ENCRYPTED_CREDENTIALS` (different question: "grants access" vs. "the writer encrypts it").
2. `log.fluentd.security.password` was misclassified as a plaintext-written property after Redmine #76051 changed its writer to encrypt; the allow-list entry never followed. Fixed by moving it into `ENCRYPTED_CREDENTIALS`.
3. As a consequence of (2) becoming settable, `AdminChangesetApplyService.apply` gained a masking step so a credential's stored value (which can legitimately be plaintext, as this very bug shows) is never echoed back in the apply response — while leaving rollback and the audit trail (`AdminChangeResult`) untouched, since those need the real value.

### Correctness — verified against source, not just the diff

I read the full current state of `AdminPropertyCatalog`, `AdminPropertiesController`, `AdminChangePlanService`, `AdminChangeService`, `AdminChangesetApplyService`, `ShareSettingsService`, and `ShareController` (not just the diff hunks) to check the PR's factual claims:

- `ShareSettingsService` reads/writes both `share.slack.url` and `share.googlechat.url` with plain `SreeEnv.getProperty`/`setProperty`; `ShareController` reads them plain when posting. Confirmed — nothing encrypts either property, so `CONFIRMED_SECRET` (not `ENCRYPTED_CREDENTIALS`) is the correct home.
- `defaults.properties` declares both URLs empty, which explains why `stored != null` used to make them report `EXISTS_CONFIRMED` before this change stopped reading them — the added `confirmedSecret` term in `AdminPropertiesController.view()`'s `confirmed` expression is necessary to avoid the regression to `EXISTS_UNKNOWN`, and it does what the PR claims.
- `AdminChangePlanService.resolve()` already refuses via `isSecret(...) && !isEncryptedCredential(...)` — since `CONFIRMED_SECRET` names are `isSecret()==true` and not in `ENCRYPTED_CREDENTIALS`, they're rejected before a plan is ever built, so they can never reach `AdminChangesetApplyService.apply()`. That means the new `maskValues` check in `AdminChangesetApplyService` (which only tests `isEncryptedCredential`, not `isConfirmedSecret`) is correctly scoped — `CONFIRMED_SECRET` values never need masking there because they can't be written at all. This file wasn't touched by the PR and didn't need to be.
- `AdminChangeService.applyChange` keeps `beforeValue`/`afterValue` as the raw stored form on every path (by design, for rollback fidelity) — confirming the PR's stated rationale for masking at the `ApplyOutcome` DTO layer rather than in `AdminChangeResult`.
- The `isSecret` ordering (`CONFIRMED_NOT_SECRET` early-return, then the union including `CONFIRMED_SECRET`) is safe because the two hand-verified sets are asserted disjoint by a new reflection-based test (`keepsTheTwoHandVerifiedNameListsDisjoint`), which correctly avoids the previously-identified trap of asserting disjointness through `isSecret()` itself (which would pass vacuously due to the early return).

I did not find any place where the new `CONFIRMED_SECRET`/`isConfirmedSecret` logic is inconsistent with existing call sites, and the javadoc's factual claims about writers/readers check out against the actual code.

### Clean code / readability / maintainability

- The javadoc is unusually thorough and explains *why*, not just *what* — including the failure mode each new set is closing off, and the cost being accepted (webhook URLs becoming unsettable via admin-chat). This is valuable given how easily this class's invariants rot (as this very PR demonstrates for `log.fluentd.security.password`).
- Dropping the hand-maintained "fifteen properties" count from the javadoc, and replacing count-based test assertions with non-emptiness checks, removes a second source of the same kind of staleness bug this PR is fixing. Good instinct.
- `AdminPropertyCatalog.isConfirmedSecret` is `public`, mirroring `isEncryptedCredential`; currently only `AdminPropertiesController` calls it, but that mirrors the existing pattern and is reasonable for future consumers (e.g. `AdminChangePlanService` could use it for a more specific refusal message later).

- `setMarker` correctly preserves `null` vs `"(set)"`, so "no prior credential" stays distinguishable from "credential replaced" — covered by `stillReportsACredentialAsUnsetWhenItHadNoPriorValue`.
- The rollback path masking is verified not to leak into the value actually replayed (`rollsBackACredentialWithItsRealStoredValueNotTheMaskedOne`), which is the one place a masking bug would have caused real damage (writing the literal string `"(set)"` into a property).
- Case-insensitivity of the new `CONFIRMED_SECRET` lookup is tested (`withholdsABearerCredentialWhateverCasingTheCallerUses`).

### Security

- This PR is itself a security fix (webhook URL / stale-plaintext-credential disclosure) and does not introduce any new exposure I could find. The masking added in `AdminChangesetApplyService` closes a real latent egress path that only became reachable because of the fluentd reclassification in this same PR — good catch, and correctly scoped to just the response DTO.

- `share.slack.url` / `share.googlechat.url` become unsettable through admin-chat (still settable via Enterprise Manager). This is an intentional, documented behavior change scoped to the admin-chat/AI plugin surface, not a general API compatibility break.
- `log.fluentd.security.password` becomes settable through admin-chat (previously silently refused with no diagnostic). This is a strict improvement.
- Existing tests that assumed the old classification were intentionally updated with clear commentary on why the old example expired (`log.fluentd.security.password` → `license.key`/`sso.rsa.private.key`/`mail.smtp.tokenuri` as replacement examples). I checked these substitutions are apt (each replacement's writer behavior matches what the test is trying to demonstrate).

### Tests

- New coverage is extensive and targeted: webhook URL withholding + `exists`/description correctness, `share.*` siblings staying readable, fluentd password now going through `setPassword`, apply-response masking (including the null/no-prior-value case and the rollback-uses-real-value case), and set disjointness. I did not find gaps.
- PR description states `202 tests, 0 failures, 0 errors` for the targeted module; the CI `build` check also passed.

### Blocking issue: merge conflict with the base branch

I attempted an actual merge of this PR's branch (`bug-76170-wiz`) into its base (`stylebi-wiz-main-rebase`) to verify `gh`'s reported `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, and confirmed a real, substantive conflict in `core/src/main/java/inetsoft/web/admin/ai/AdminPropertyCatalog.java`:

```
<<<<<<< HEAD (this PR)
      return CONFIRMED_SECRET.contains(lower) || isEncryptedCredential(lower) ||
=======
      return isEncryptedCredential(lower) || COMPOSITE_SECRET_PROPERTIES.contains(lower) ||
>>>>>>> origin/stylebi-wiz-main-rebase (current base)
```

Since the base branch advanced, a different, already-merged change added a `COMPOSITE_SECRET_PROPERTIES` set to `isSecret()` that overlaps in purpose with this PR's new `CONFIRMED_SECRET` set. This isn't a stale/cosmetic conflict — it needs a human/author decision about how the two sets (and their surrounding javadoc, which both PRs edited) should be reconciled, not a mechanical rebase. This must be resolved before merge.

---

### Decision

Per the review checklist:

- **Code review**: No blocking issues found in the code itself — this is a well-reasoned, well-tested, narrowly-scoped security fix.
- **Checks**: All checks pass except `Claude Code Review / claude-review`, which is excluded per this review's instructions (this manual review replaces it). ✅
- **Merge conflict**: ❌ **Fails.** `gh pr view` reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, and I confirmed a genuine, non-trivial content conflict in `AdminPropertyCatalog.java` (`CONFIRMED_SECRET` vs. the base branch's newly-added `COMPOSITE_SECRET_PROPERTIES`) by performing the merge directly.

Because the merge-conflict condition fails, this review **requests changes**, not because of any defect in the code as written, but because the PR must be rebased/reconciled against the current `stylebi-wiz-main-rebase` before it can be merged.

**Requested action:** rebase (or merge) onto the current `stylebi-wiz-main-rebase`, reconcile `CONFIRMED_SECRET` with `COMPOSITE_SECRET_PROPERTIES` (and the javadoc both touch) in `AdminPropertyCatalog.isSecret`, and re-run the test suite before re-requesting review.

jshobe-inetsoft (changes requested):
## Code Review — PR #4825

Manual review (the automated `Claude Code Review / claude-review` check is currently broken on many open PRs in this repo due to an infrastructure issue and is disregarded here).

### Summary of the change

Two gaps in `AdminPropertyCatalog`'s definition of a "secret" property, targeting the `stylebi-wiz-main-rebase` branch:

1. `share.slack.url` / `share.googlechat.url` (incoming-webhook URLs, bearer credentials by possession) were neither name-matched nor writer-encrypted, so they were read back in full through the admin-chat properties endpoint. A new hand-verified `CONFIRMED_SECRET` set withholds them, `isConfirmedSecret()` is exposed for a dedicated `PropertyView` description branch, and `confirmed` is extended so `exists` stays `EXISTS_CONFIRMED` rather than regressing to `EXISTS_UNKNOWN`.
2. `log.fluentd.security.password` is reclassified into `ENCRYPTED_CREDENTIALS` now that Redmine #76051 routed its writer through `Tool.encryptPassword`, and, as a consequence of making it settable again, `AdminChangesetApplyService` now masks credential before/after values in the apply response (`(set)`) so a plaintext-stored legacy value isn't echoed to the LLM caller.

### Correctness

- Verified independently against the actual source: `ShareSettingsService`/`ShareController` do read/write `share.slack.url` and `share.googlechat.url` with plain `SreeEnv.getProperty`/`setProperty` — the PR's premise holds.
- `AdminChangePlanService.resolve()` still refuses a write when `isSecret(name) && !isEncryptedCredential(name)`, so the two new `CONFIRMED_SECRET` names correctly remain unwritable via admin-chat (consistent with the PR's stated intent).
- `AdminPropertiesController.view()`: branch order is `credential` → `confirmedSecret` → `secret` (pattern-only), and since `isSecret` unions `CONFIRMED_SECRET`, `stored` is correctly forced to `null` for these names via the existing `secret ? null : ...` guard. The `confirmed` boolean is extended correctly.
- `AdminChangesetApplyService`: masking is scoped to `isEncryptedCredential` only (not `CONFIRMED_SECRET`), which is correct — `CONFIRMED_SECRET` properties can never reach `apply()` because the plan service already refuses them at preview time, so masking them here would be dead code.
- The `keepsTheTwoHandVerifiedNameListsDisjoint` test reflects over the actual `CONFIRMED_SECRET`/`CONFIRMED_NOT_SECRET` fields rather than asserting through `isSecret()` — a real fix, since the old style of assertion would pass vacuously due to `isSecret`'s early return on `CONFIRMED_NOT_SECRET`.
- Test updates track the reclassification correctly (moving `log.fluentd.security.password` out of "plaintext" example lists into the encrypted-credential list, updating `AdminChangePlanServiceTest`/`AdminChangeServiceTest` fixtures accordingly).

### Blocking issue: unresolved merge conflict

**This PR has a real merge conflict with the current tip of its base branch, `stylebi-wiz-main-rebase`**, not a stale GitHub mergeability cache:

- `gh pr view 4825 --json mergeable,mergeStateStatus` reports `"mergeable":"CONFLICTING"`, `"mergeStateStatus":"DIRTY"`.
- PR #4843 (`Withhold server.save.locations from admin-chat property reads`), already merged to `stylebi-wiz-main-rebase`, added a `COMPOSITE_SECRET_PROPERTIES` set and touches the exact same `return` statement in `AdminPropertyCatalog.isSecret()` that this PR modifies:

  ```
  <<<<<<< base (stylebi-wiz-main-rebase)
  return isEncryptedCredential(lower) || COMPOSITE_SECRET_PROPERTIES.contains(lower) ||
  =======
  return CONFIRMED_SECRET.contains(lower) || isEncryptedCredential(lower) ||
  >>>>>>> PR #4825
  ```

  This needs to be manually reconciled (both terms need to survive in the union) before this can merge. The conflict is confined to this one line, but it must be resolved and re-verified (tests re-run) rather than merged as-is.

### Minor observations (non-blocking)

- `CONFIRMED_SECRET` and `ENCRYPTED_CREDENTIALS` disjointness is asserted only for `CONFIRMED_SECRET`/`CONFIRMED_NOT_SECRET`, not against `ENCRYPTED_CREDENTIALS`. Given both are small, hand-curated, human-reviewed sets, this is a low-risk gap, but a future entry added to both would fail silently favoring the `credential` branch in `AdminPropertiesController` rather than raising a diagnostic. Not worth blocking on.
- The javadoc is unusually dense/discursive (multi-paragraph essays per constant), which is this codebase's established style for this class (visible in the pre-existing code) — consistent with the surrounding file, not a new pattern introduced by this PR.
- Backward compatibility: `AdminPropertyName.baseName()`/`isConfirmedSecret` is a new public static method purely additive; no existing public signatures changed. `ApplyOutcome`'s before/after fields for credentials now report `"(set)"` instead of the raw stored value — this is an intentional, security-motivated behavior change to the apply-response contract for encrypted-credential properties (previously it echoed ciphertext or, for the newly-reclassified fluentd password, plaintext); callers relying on reading back the literal stored value for `mail.smtp.*`/`log.fluentd.security.*` credentials through this endpoint will see the new masked marker. This is called out explicitly in the PR description and is the intended fix, not an accidental regression, but is worth flagging as a contract change for any downstream consumer of the apply response.
- No test coverage breaks; new tests were added and existing tests were updated in lockstep with the reclassification (verified in the diff).

### Verdict: **Request changes**

Reasons (per review criteria):
- No blocking issues in the code logic itself — the change is well-reasoned, correctly scoped, and well-tested.
- CI checks pass except the disregarded `claude-review` check.
- **However, the PR currently has a real, unresolved merge conflict against its base branch (`stylebi-wiz-main-rebase`)**, caused by an overlapping edit from already-merged PR #4843 to the same line in `AdminPropertyCatalog.isSecret()`. This must be rebased/resolved before merge.

Please rebase onto the current `stylebi-wiz-main-rebase` tip, reconcile `COMPOSITE_SECRET_PROPERTIES` and `CONFIRMED_SECRET` into the same union expression, and re-run the test suite.

jshobe-inetsoft (changes requested):
## Code review — Bug #76170, withhold the share webhook URLs and re-verify the fluentd password

I verified every factual claim in the description against the source rather than taking it on trust, and the core of the change holds up:

- `ShareSettingsService.getModel`/`setModel` do read and write `share.slack.url` / `share.googlechat.url` with a bare `SreeEnv.getProperty` / `setProperty`, `ShareController` consumes them as literal URLs, and `defaults.properties:215,223` declares both empty — so the `confirmed` third term really is needed to avoid regressing `EXISTS_CONFIRMED` → `EXISTS_UNKNOWN`, and putting them in `ENCRYPTED_CREDENTIALS` really would store ciphertext under a literal-URL property.
- `LogSettingService.saveFluentdSettings` writes `log.fluentd.security.password` through `toPassword` → `Tool.encryptPassword` (line 298), identically to `log.fluentd.security.sharedKey` (line 293). `SreeEnv.setPassword` in the non-cloud-secrets branch is exactly `Tool.encryptPassword` + `setProperty`, so the allow-list entry round-trips correctly with `LogSettingService`'s own `getProperty` + `Tool.decryptPassword` reader. The cloud-secrets case is correctly refused up front by `AdminChangePlanService`.
- The masking seam in `AdminChangesetApplyService` is placed correctly: `undoableBefore.add(applied.getBeforeValue())` and the `moved` comparison both read `AdminChangeResult` directly, so neither rollback nor the moved-detection sees `(set)`. The new test pins that.
- `license.key` is written with a plain `SreeEnv.setProperty` (`SecurityConfigController:307`, `PropertiesEngine:1063`), so it is a valid replacement subject for `writesASecretNamedPropertyThatIsNotAnAllowListedCredentialVerbatim`.
- The reflective disjointness test is a genuine improvement over the previous spelling, and the reasoning about why the old one went green on the contradiction is correct.

Two blocking issues, both about the same gap.

---

### Blocking 1 — the same two webhook URLs are still returned in full (and are still writable) by the sibling admin-ai endpoint

`PresentationSubModel.SHARE` maps to `PresentationShareSettingsModel`, which carries `slackUrl()` and `googleChatUrl()` populated straight from `SreeEnv.getProperty("share.slack.url" / "share.googlechat.url")`. `AdminPresentationController.projectedRead` masks **only** `WEB_MAP`:

```java
private JsonNode projectedRead(PresentationSubModel subModel, Principal user, boolean global) {
   JsonNode node = PresentationJson.toNode(access.read(subModel, user, global));
   return subModel == PresentationSubModel.WEB_MAP
      ? PresentationJson.maskWebMapSecrets(node) : node;
}
```

So `GET /api/wiz/v1/admin/presentation/settings?scope=global` — which, with `subModel` omitted, returns all 16 sub-models — hands both webhook URLs to the caller in the clear. That is the *same* caller, in the *same* `inetsoft.web.admin.ai` package, relaying to the *same* model provider off-host that the `CONFIRMED_SECRET` javadoc names as the entire justification for the change. And "show me the sharing settings" is arguably the more natural agent path to those values than `GET /admin/properties/share.slack.url`.

`PresentationChangePlanService.projectedValue` has the same `WEB_MAP`-only condition, so the plan's `currentValue`/`proposedValue` echo them too.

The remedy already exists in the file next door and is cheap: a `SHARE_SECRET_FIELDS = Set.of("slackUrl", "googleChatUrl")` beside `WEB_MAP_SECRET_FIELDS`, masked in `projectedRead` and `projectedValue` exactly as web-map's are. Note the merge is safe: `resolveEntries` merges the spec into the **unmasked** `currentNode` and only the projection is masked, so a masked read cannot round-trip `********` into the store — the web-map precedent already relies on that.

If closing this here is genuinely out of scope, that is a defensible position, but then it needs to be stated in the PR and tracked, not left silent — right now the diff reads as if the hole is closed when half of it is not.

### Blocking 2 — as a direct consequence, the new guidance string tells the agent something false

`AdminPropertiesController`'s new `confirmedSecret` branch asserts:

> "It cannot be changed here either: nothing in the product encrypts it, so admin-chat will not write it. Configure it through Enterprise Manager."

and the `CONFIRMED_SECRET` javadoc's "The cost, accepted" paragraph says the same: "an operator can no longer configure Slack or Google Chat sharing by chat, and must use Enterprise Manager."

Both are wrong. `PresentationChangePlanService` invokes `requireNoSecretFields` only for `WEB_MAP` (line 139-141), so an `update` on the `share` sub-model carrying `slackUrl` is accepted and applied through `ShareSettingsService.setModel`. Admin-chat can still set both, just through a different verb.

This matters more here than a stale comment usually would: this string is not documentation, it is instruction text consumed by an LLM. Telling an agent a property is unsettable when it is settable is the exact class of failure the surrounding comments repeatedly cite as the reason a branch exists ("Telling an agent a settable credential is unchangeable costs a task it could have completed"). Fixing Blocking 1 makes both statements true; fixing only this one means rewording them to describe the real state.

---

### Non-blocking

1. **`keepsTheHandVerifiedNameListsDisjoint` misses the third positive set.** The loop covers `CONFIRMED_SECRET` and `COMPOSITE_SECRET_PROPERTIES`, and the comment claims "both hand-verified positive sets are checked here" — but `ENCRYPTED_CREDENTIALS` is in the same union in `isSecret` and is shadowed by the `CONFIRMED_NOT_SECRET` early return identically. A name landing in both would be read back in the clear with no diagnostic, which is precisely the failure this test was written to catch. Adding `"ENCRYPTED_CREDENTIALS"` to the `List.of(...)` costs one token and the reflective helper already handles it.

2. **`setMarker` maps `""` to `"(set)"`.** `value == null ? null : "(set)"` reports a credential as present when the stored value is the empty string. Not reachable through the plan path today (`AdminChangePlanService` refuses `""` for a credential and `SreeEnv.setPassword` guards on `Tool.isEmptyString`), but the comment three lines above specifically invokes `INETSOFTENV_*` promotion and the raw EM properties page as writers that bypass every guard — and those can store `""`. `Tool.isEmptyString(value) ? value : "(set)"` is exact and matches the semantics the sibling test (`stillReportsACredentialAsUnsetWhenItHadNoPriorValue`) is asserting.

3. **Informational:** masking `before` and `after` to the same literal makes a no-op or failed credential apply indistinguishable from a successful replacement in the DTO's value fields. `status` still carries it, so this is fine, but worth knowing when reading an apply response.

4. **Release-note-worthy behaviour changes**, both intended but neither obvious from the title: `log.fluentd.security.password` becomes settable through admin-chat where it was refused, and `share.slack.url` / `share.googlechat.url` become unreadable *and* unsettable through the properties path where they were both.

5. Trivial: `keepsTheHandVerifiedNameListsDisjoint` computes `both` unconditionally purely to build the failure message.

---

### Things I checked and found clean

- `AdminPropertyName.parse` is idempotent over an already-resolved key, so the new `parse(change.property()).baseName()` in the apply loop is correct for org-scoped names too (`inetsoft.org.{id}.share.slack.url` → base `share.slack.url`, so the mask follows org overrides).
- `isSecret`'s ordering is still sound: `CONFIRMED_NOT_SECRET` early-returns, and none of its three names collide with the new set.
- No other production caller of `isSecret` / `isEncryptedCredential` exists beyond the three in this package, so widening `isSecret` has no blast radius outside admin-chat.
- Case handling: `isConfirmedSecret` lowercases, matching the other two predicates; the `Share.Slack.URL` test pins it.
- `WizShareController` does not return either URL, so that surface is unaffected.
- No backward-compatibility break for the EM Share settings page — `ShareSettingsService` is untouched.

---

### Blocking status

Requesting changes for the two blocking items above. For the record on the mechanical gates:

- **Checks:** all green except `Claude Code Review / claude-review`, which this manual review replaces. `build`, `Semgrep Scan`, `Semgrep OSS` and `license/cla` all pass.
- **Merge conflicts:** none — `mergeable: MERGEABLE` against `stylebi-wiz-main-rebase` (`mergeStateStatus: UNSTABLE` is solely the excluded `claude-review` failure).

jshobe-inetsoft (approved):

PR #4826, opened 2026-08-27
Bug #76045, refuse the fluentd log provider where forwarding is unavailable

ngovus-inetsoft:
## Problem

`log.provider=fluentd` cannot work without the enterprise module. The forwarder is loaded reflectively from `inetsoft.enterprise.log.fluentd`, so `LogbackUtil.createForwardAppender` threw `ClassNotFoundException`, `LogbackInitializer` caught it, printed to `System.err` **while logging was still being initialised** — where it reached neither the log file nor Enterprise Manager — and fell through to the file appender.

`LogbackUtil.resetLog` already guarded the same enterprise classes with an explicit `isEnterprise()` check. That asymmetry was the root defect: one path checked the licence, the other relied on a class-loading failure that it then swallowed.

### Two corrections to the original report

- **The Logging page already hides the option.** The provider `<mat-select>` is wrapped in `@if (enterprise)`, so the EM Logging page was *not* the route by which a Community build reached `fluentd`. The reachable route is the Properties page's **Add Property** button — `log.provider` is deliberately not hidden by `removeUnuseProperties`, since `file` is a valid Community value — plus `sree.properties` and `INETSOFTENV_LOG_PROVIDER`.
- **The related "7 of 12 fluentd keys" issue is already fixed** (#4663 matches the whole family by prefix).

### The blast radius was wider than the silent fallback

Five call sites compared the raw property value and so believed a provider that was not running:

| Site | Consequence |
|---|---|
| `LogbackInitializer` | silent fallback to file logging; stderr-only report |
| `LogSettingService` | Logging page showed the fluentd form with the selector hidden — no way back to `file` |
| `LogMonitoringService` | **the Logs monitoring page hid the local log file the records were being written to** |
| `SchedulerConfigurationService` | blanked the scheduler log file field |
| `ActionPermissionService` | org-admin exclusions branched on it (orgs are enterprise-only, so no practical change) |

## Change

- **`LogbackUtil.isFluentdEnabled()`** — requires the property value *and* `LicenseManager.isEnterprise()`, generalising the guard `resetLog` already had. All five readers route through it. `isEnterprise()` is a memoised classpath probe with no Spring, `SreeEnv`, logging or cluster dependency, so it is safe to call during logging bootstrap.
- **Refuse the value at both write paths** — `LogSettingService.setConfiguration` and `PropertiesController.editProperty` throw a `MessageException` naming the edition requirement. In the properties controller the check runs *before* the blank-value branch substitutes the stored value, so a build already carrying `fluentd` can still submit that field.
- **Surface the rejection.** Both save subscriptions previously had no error callback, so a rejection was swallowed and the edit looked like a no-op — the same silent failure this change exists to end. Both now use the existing `ErrorHandlerService.showDialog`.
- **Report the effective provider, and stop writing forwarding state where it cannot apply.** The Logging page reports the provider actually in use, so an operator whose `log.provider` was set out of band is not shown an unusable fluentd form. Because the model then arrives carrying `provider=file` and `fluentdSettings=null` even when all twelve `log.fluentd.*` properties are populated, a build that cannot forward no longer writes any forwarding state — otherwise `setFluentdSettings(null)` would have cleared the stored host, port, shared key, username, password and CA path on the next unrelated save, unrecoverably.
- **Make the residual failure visible.** `LogbackInitializer` now reports the fallback through the logging framework at the end of `initialize()`, once every appender is attached. At ERROR, because `initialize()` has just reset the context to ERROR and the configured levels are applied afterwards (`PropertiesEngine.initLogging`), so anything lower is dropped — and because a log pipeline receiving nothing is an error. Emitted only on a state change, since `initialize()` re-runs once per `log.*` property at startup.

## Testing

- `LogbackUtilTest` (new) — the predicate's truth table over both inputs.
- `LogSettingServiceTest` — the six pre-existing fluentd cases now stub the licence; added the save rejection, the licensed save, and both directions of the data-loss regression.
- `PropertiesControllerTest` — rejection, the licensed case, the `file` value, and the blank-submission case.
- `logging-settings-page.component.spec.ts` — the save-error dialog and the success path.

80 Java tests pass in the affected packages; the full EM suite passes at 367 tests. ESLint clean.

Community-only change, so there is no accompanying enterprise PR and the submodule pointer is untouched.


jshobe-inetsoft (approved):
## Summary

This PR fixes Bug #76045: `log.provider=fluentd` was accepted on Community builds even though the forwarder (`inetsoft.enterprise.log.fluentd.*`) only exists in the Enterprise module. The root cause was that `LogbackUtil.resetLog()` already guarded the enterprise classes with `LicenseManager.isEnterprise()`, but five other call sites compared the raw `log.provider` value instead, so they believed a provider that was never actually running.

The fix centralizes the check in a new `LogbackUtil.isFluentdEnabled()` predicate (`"fluentd".equals(log.provider) && LicenseManager.isEnterprise()`), routes all five readers through it, adds write-time rejection (`MessageException`) in both `LogSettingService.setConfiguration` and `PropertiesController.editProperty`, wires up previously-missing error callbacks on the two Angular save subscriptions, and makes `LogbackInitializer` log the fallback via the logging framework (not just `System.err`) once appenders are attached.

## What I verified

- Confirmed (via `grep`) that all five pre-existing bare `"fluentd".equals(SreeEnv.getProperty("log.provider"))` comparisons in `core/` are updated to go through `LogbackUtil.isFluentdEnabled()`: `LogbackInitializer`, `LogMonitoringService`, `LogSettingService`, `SchedulerConfigurationService`, `ActionPermissionService`. No stray bare comparison was left behind.
- `LicenseManager.isEnterprise()` is indeed a `Suppliers.memoize`d classpath probe with no Spring/SreeEnv/logging dependency, matching the PR description's claim that it's safe to call during logging bootstrap.
- `LogManager.useInitializer()` constructs a **new** `LogbackInitializer` instance on every call (via reflection), which is why `fluentdFallbackReported` needs to be `static` — this is a deliberate, correct choice, not an accidental shared-mutable-state bug. The reset-to-`false` branch in `reportFluentdFallback` correctly clears the flag once the condition resolves, so a fixed configuration won't get stuck suppressing future reports.
- Traced the error-reporting plumbing end-to-end: `MessageException` → `AdminExceptionHandler.handleMessageException` → `GenericError`/`ApiError` (message populated from `e.getMessage()`, HTTP 500) → Angular `ErrorHandlerService.getMessage()` reads `error.error.message`. The new frontend tests' expectations (500 status, `message` in body) match this exactly.
- The `@Audited` AOP aspect (`EventAspect.handleAudit`) wraps the entire `PropertiesController.editProperty` method and records `ACTION_STATUS_FAILURE` on any thrown exception, so the new early `MessageException` in that controller is still captured in the audit trail.
- Confirmed the Logging page's provider `<mat-select>` is indeed gated by `@if (enterprise)` in `logging-settings-view.component.html`, so the `LogSettingService.setConfiguration` guard is mainly defense-in-depth against a direct API call rather than something reachable through the normal UI on a Community build — consistent with the PR description.
- The new Java tests (`LogbackUtilTest`, additions to `LogSettingServiceTest` and `PropertiesControllerTest`) cover the truth table thoroughly (licensed/unlicensed × fluentd/file/unset/garbage, exact-match case/whitespace sensitivity, the save-rejection path, the data-loss-prevention path for the 12 `log.fluentd.*` keys, and the blank-value-preserves-existing-value edge case). The new Angular test covers both the error and success subscribe paths for the Logging page.
- CI `build` passed (17m35s), Semgrep OSS/Scan passed, CLA signed. The only failing check is `claude-review`, which failed due to an internal tooling error (`"Internal error: directory mismatch for directory .../tsconfig.json"`), not a substantive finding — excluded per review policy.
- `mergeStateStatus` is `BLOCKED`, but `mergeable` is `MERGEABLE`/`mergeable_state: blocked` — this is purely the "review required" branch-protection gate (`reviewDecision: REVIEW_REQUIRED`), not a merge conflict.

## Minor, non-blocking observations

1. **Audit trail gap in `LogSettingService.setConfiguration`.** The new license/provider guard throws `MessageException` *before* the `ActionRecord` is created, so a rejected save on the Logging Settings page leaves no audit trail entry. This is asymmetric with `PropertiesController.editProperty`, where the equivalent rejection is still audited (because the whole method is wrapped by the `@Audited` AOP aspect). Since this concerns an admin attempting to configure log forwarding, it would be more consistent to record the rejected attempt (e.g., create the `ActionRecord` first and set `ACTION_STATUS_FAILURE` in a catch, or wrap the guard in the existing try/finally).
2. **Duplicated guard logic.** The `FLUENTD_PROVIDER.equals(value) && !LicenseManager.isEnterprise()` check is written out twice (`LogSettingService.setConfiguration` and `PropertiesController.editProperty`). Consider a small shared helper (e.g., `LogbackUtil.requireFluentdAllowed(value)`) so the two call sites can't drift apart later.
3. **No new test for `property-settings-view.component.ts`.** The `acceptChanges()` error-handling change there has no accompanying spec (there's no existing spec file for this component at all, so this is a pre-existing gap, not strictly introduced by this PR, but worth flagging since the sibling `logging-settings-page` change did get a test).
4. **Trivial duplicate property read.** `LogbackInitializer.initialize()` computes `fluentdUnavailable` by calling `SreeEnv.getProperty("log.provider")` again right after `isFluentdEnabled()` already read the same property. Harmless, just a small redundant lookup.

None of these block the fix — they're suggestions for a follow-up if the author wants to tighten things further.

## Conclusion

The change is complete, correctly addresses the root cause (single source of truth for "is forwarding actually usable"), is backward compatible for licensed Enterprise deployments (explicitly covered by tests that assert the enterprise path is unchanged), does not silently discard stored Fluentd credentials on a downgraded/community build, and is well covered by new tests on both the Java and Angular sides. I did not find any regressions, broken tests, or security issues.

**Recommendation: Approve.**

PR #4827, opened 2026-08-27
Bug #75954, thread N through set_group_aggregate for parametrized formulas

jkidacki-inetsoft:
## Summary

`set_group_aggregate` (the AI Wiz Composer plugin's worksheet grouping/aggregation tool) silently dropped the `N` operand for parametrized aggregate formulas (`NthLargest`, `NthSmallest`, `NthMostFrequent`, `PthPercentile`). Asking for "5th largest" (NthLargest, N=5) silently computed as N=1 (Max) instead — no error anywhere. `read_worksheet_model` also kept reporting the formula back as bare `"NthLargest"` with N gone, even once the value was applied correctly.

## Root Cause

Three places in the AI Wiz worksheet-edit pipeline dropped `N`:

1. `WorksheetMutationSupport.AggregateSpec` had no `n` field, so there was nowhere for it to travel to.
2. `WorksheetAgentController.dispatch()`'s `case "set_group_aggregate"` rebuilt the aggregate list field-by-field and omitted `n` in that rebuild.
3. `WorksheetMutationSupport.applyAggregateInfo` never called `AggregateRef.setN(...)` on either the primary aggregate or the secondary-aggregate → new-primary-aggregate conversion path used when a second aggregate targets the same column.
4. `WorksheetReadService`/`WorksheetModel.AggregateRefModel` had no way to surface `n` back out, so `read_worksheet_model` couldn't report it even after it was correctly applied.

## Fix

- `AggregateSpec` and `AggregateRefModel` now carry an optional `Integer n` (each keeps a back-compatible constructor overload for existing call sites).
- `applyAggregateInfo` threads `n` onto the `AggregateRef` in both aggregate-construction paths, gated on `AggregateFormula.hasN()`.
- `WorksheetAgentController` forwards `n` through instead of dropping it.
- `WorksheetReadService` surfaces `n` back out in `read_worksheet_model`, gated the same way.

This matches the existing, already-correct precedent in `GenerateWsService`/`WorksheetTableService` (used by `create_worksheet_table`), and the Angular Composer UI's own formula editor (`n`/`hasN` gating).

## Test Plan

- Added regression tests: N applied on write (`setGroupAggregateAppliesNForParametrizedFormula`), N ignored for non-parametrized formulas, N surviving the same-column secondary-aggregate conversion (`setGroupAggregateAppliesNThroughSecondaryConversion`), and N correctly round-tripping (or correctly omitted) through `read_worksheet_model` (`WorksheetReadServiceTest`).
- `mvn clean test -pl core -am` — `WorksheetEditServiceMutatorsTest` (157), `WorksheetReadServiceTest` (32), `WorksheetAgentControllerTest` (71), `GroupSpecDeserializationTest` (5) all pass.
- Manually verified live via the Composer plugin.

Fixes Redmine #75954.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -1716,7 +1716,7 @@ public class WorksheetAgentController {
                                      req.aggregates() != null
                                         ? req.aggregates().stream()
                                             .map(a -> new WorksheetMutationSupport.AggregateSpec(
-                                                a.field(), a.formula(), a.alias()))
+                                                a.field(), a.formula(), a.alias(), a.n()))
                                             .toList()
                                         : List.of(),
                                      Boolean.TRUE.equals(req.crosstab()));
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetMutationSupport.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetMutationSupport.java
@@ -72,8 +72,16 @@ public final class WorksheetMutationSupport {
     * @param formula a formula name recognised by {@link AggregateFormula#getFormula}
     *                (e.g. {@code "SUM"}, {@code "COUNT"}, {@code "AVG"})
     * @param alias   optional output alias; may be {@code null}
+    * @param n       the N/P operand for a parametrized formula (e.g. {@code "NthLargest"},
+    *                {@code "NthSmallest"}, {@code "NthMostFrequent"}, {@code "PthPercentile"});
+    *                {@code null} (or a formula that doesn't take one, per
+    *                {@link AggregateFormula#hasN}) is ignored
     */
-   public record AggregateSpec(String field, String formula, String alias) {}
+   public record AggregateSpec(String field, String formula, String alias, Integer n) {
+      public AggregateSpec(String field, String formula, String alias) {
+         this(field, formula, alias, null);
+      }
+   }
 
    /**
     * Describes a single group-by column for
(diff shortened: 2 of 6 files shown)


jkidacki-inetsoft:
## Review

Traced this past the diff into the actual `AggregateRef`/`AggregateInfo` source rather than trusting the description, since the "second aggregate on the same column" path looked suspicious at first (equality by formula would make `containsAggregate` never match here) — it isn't: `AggregateRef` has no `equals()` override, so it inherits `AbstractDataRef.equals()`, which compares only `getName()`, and `AggregateRef.getName()` returns just the underlying column's name (formula excluded). Since `colRef` is the same shared object mutated in place by the first spec's alias, the second spec's `ar` correctly matches it regardless of differing formula — so `ar.setN(...)` being called before that `containsAggregate` check (on an object that's discarded when the secondary branch is taken) is harmless, not a bug.

**Verified correct:**
- `hasN()` gating is the right mechanism — it's formula-provided (`false` by default, overridden `true` on the 4 parametrized formulas), so nothing here needs to hardcode which formulas take N.
- The three threading sites (primary aggregate, secondary aggregate, and the secondary→new-primary conversion block) are all gated identically and all tested — including the conversion path specifically, which is the one that's easy to miss (confirmed `sref.getN()` is available there because it was set on the same `secondary` object earlier, before `addSecondaryAggregate`).
- `WorksheetReadService.readAggregateRef`'s read-back matches the write gating exactly (`formula.hasN()`), and `omitsNForNonParametrizedFormula` confirms N doesn't leak through for a formula that doesn't use it even when the underlying field is nonzero from a prior formula switch.
- This mirrors the existing `GenerateWsService`/`WorksheetTableService` precedent, as claimed — same `hasN()` gate, same shape.

**Not a defect here, but flagging for the companion PR:** neither this PR nor stylebi-wiz#1889 rejects a `set_group_aggregate` call that names a parametrized formula (`NthLargest` etc.) without `n`. Per this repo's own CLAUDE.md tool-misuse principle, silently accepting that and producing a plausible-but-wrong result (N defaults to `0`, no error) is the same class of bug this PR fixes, just from a different trigger (omitted `n` instead of dropped `n`). Left the actionable comment on #1889 since that's the layer that can fail loud before ever posting.

Also noting, as before: `claude-review`'s CI failure here is the same infra issue as the last two PRs (`ANTHROPIC_API_KEY` empty in the runner env) — not a real finding. `build` passed.

LGTM, modulo the companion-PR note above.

ngovus-inetsoft (approved):

PR #4828, opened 2026-08-27
Fix composer status band alignment and condition-dialog visual gaps

swaker854:
Stacked on #4816 (0820-3-native-elements-sweep → bugfix-composer-primary-buttons-0820 → epic-74519).

## Summary

Three unrelated composer/condition-dialog bug fixes found while reviewing the current stack:

**Worksheet composer table header (`ws-details-table-data`)**
The column header row's label text, type icon, and "..." menu button sat at different vertical positions instead of centered together. Root cause was layered: `.column-buttons` used `align-items: flex-start` instead of `center`; `.mini-menu` had `align-items: center` but no `display: flex`, so the rule was dead; and `.ws-header-column__display-container` inherited a `line-height` that inflated its height independent of its actual text, which in turn left `.ws-header` without the implicit sizing it depended on once that inflation was removed.

**Composer status band (`viewsheet-pane`)**
- The "Layout:" label sat visibly higher than the dropdown next to it (`.shell-inline-form`'s default `align-items: stretch` left the label top-aligned in its stretched box while the select centers its own content).
- The "Console" button used `btn-primary`, inconsistent with the worksheet composer's equivalent `btn-default rounded-pill` button; restyled to match, including badge spacing/colors.
- The badge's message count sat visibly higher than the "Console" label (Bootstrap's `.btn .badge { top: -1px }` plus the badge's own `line-height: 1`).
- The layout dropdown rendered as an opaque white box that didn't match the band's background; both `custom-select`'s host and its inner trigger paint solid backgrounds by default — made both transparent, scoped to `.layout-dropdown` only.

**Conditions dialog (`condition-field-combo` / `variable-editor`)**
Shared by worksheet Pre/Post-aggregate/Ranking filters and viewsheet table/highlight conditions (both route through `condition-pane` → `condition-item-pane` → `condition-editor`):
- The field picker never rendered a dropdown caret — its SCSS had a fully-positioned `.condition-field-combo-icon` rule (same recipe `custom-select` uses) but no template element ever used it. Added the icon, wired to the component's already-tracked `dropdownOpen` state.
- Same component was missing the orange open/focus border+ring `custom-select` shows; bound `dropdownOpen` to a new `.is-open` class with matching styling.
- The viewsheet-only "Use Selection List" checkbox silently overflowed on top of the neighboring value-type icon and and/or radio buttons in narrow dialogs instead of wrapping (`flex-wrap: nowrap` on `variable-editor`'s host, with neither sibling able to shrink). Changed to `flex-wrap: wrap`.

Confirmed the SQL/VPM condition family (`vpm-condition-pane` and friends) is a separate implementation that already composes `custom-select` directly and doesn't share this gap — no changes needed there.

## Test plan
- [x] Verified all three fixes live in the composer (Construction Dashboard viewsheet + Construction Data / customers worksheets) via DOM/computed-style inspection and visual check
- [x] Confirmed the condition-dialog fixes apply identically in the worksheet Pre-aggregate Conditions dialog, which shares the same component chain
- [ ] `tsc --noEmit` / `ng lint` — not run in this environment


jshobe-inetsoft (changes requested):
## Review Summary

This PR is a small, well-scoped set of CSS/template fixes for three composer visual bugs (worksheet header alignment, viewsheet status band alignment, and the condition-dialog field-combo caret/ring). The write-up is thorough and the reasoning for the worksheet-header and status-band fixes checks out against the surrounding code. However, tracing the condition-dialog change and the badge/dropdown color changes against the rest of the codebase turned up several concrete defects that mean parts of this PR do not actually produce the effect described, and in one case introduce a regression. **Requesting changes.**

### Blocking issues

1. **`dropdownOpen` does not exist on `ConditionFieldComboComponent`** (`condition-field-combo.component.html` / `.ts`).
   The new template binds `[class.is-open]="dropdownOpen"`, `[class.chevron-up-icon]="dropdownOpen"`, and `[class.chevron-down-icon]="!dropdownOpen"`, but `condition-field-combo.component.ts` has no `dropdownOpen` field anywhere. `dropDownOpenChange(open: boolean)` only does:
   ```ts
   dropDownOpenChange(open: boolean) {
      if(!open) {
         this.showSearch = false;
      }
   }
   ```
   it never stores `open` into a `dropdownOpen` property. At runtime `dropdownOpen` is simply `undefined`, so `is-open` and `chevron-up-icon` never apply, and `chevron-down-icon` is permanently applied (`!undefined === true`). The PR description's claim that this is "wired to the component's already-tracked `dropdownOpen` state" is incorrect — no such tracked state exists. `strictTemplates` is `false` in `tsconfig.json`, so this compiles without error, which is presumably why CI didn't catch it.

2. **`chevron-up-icon` / `chevron-down-icon` are not real icon classes.**
   A repo-wide search of `_icon-alias.scss` and `ineticons/variables.scss` shows no `chevron-up-icon`/`chevron-down-icon` entries — only `chevron-circle-arrow-up-icon` / `chevron-circle-arrow-down-icon` (a different, "circle" glyph) exist, and those are unrelated to a small inline caret. The actual established alias for exactly this "plain caret" use case, already used elsewhere in this same codebase (`dropdown-view.component.html`, `viewsheet-print-layout-dialog.component.html`, `logical-model-attribute-editor.component.html`), is `caret-down-icon` / `caret-up-icon` (defined in `_icon-alias.scss`). Because the PR uses the non-existent classes, the new `<i>` caret renders with **no glyph at all** — combined with finding #1, the field-combo dropdown still shows no visible caret, which is the exact bug this PR claims to fix.

3. **Undefined CSS custom properties used in two places — these declarations are no-ops:**
   - `condition-field-combo.component.scss`:
     ```scss
     &.is-open {
       border-color: var(--inet-primary-color);
       box-shadow: var(--inet-focus-ring);
       outline: var(--inet-focus-ring-outline);
     }
     ```
     `--inet-focus-ring` and `--inet-focus-ring-outline` are not defined anywhere in the frontend (no hits in any `.scss` under `web/`). A `var()` reference to an unset custom property with no fallback is invalid at computed-value time, so `box-shadow`/`outline` here resolve to their initial values — i.e. no ring is ever drawn. Only the `border-color` (a valid variable) would take effect. This does not match the existing `custom-select` recipe either, which actually uses `border-color: var(--inet-primary-color-light); box-shadow: 0 0 0 1px var(--inet-primary-color);` (see `_bootstrap-override.scss` `.custom-select:focus`) — real, defined variables.
   - `viewsheet-pane.component.scss`:
     ```scss
     .console-count-badge {
       color: var(--inet-text-muted-color);
       background-color: var(--inet-shell-surface-hover);
       ...
     }
     ```
     `--inet-text-muted-color` and `--inet-shell-surface-hover` likewise don't exist anywhere in the codebase or in the canonical theme-variable registry (`projects/em/.../theme-css-view.component.ts`). This *replaces* the previously-valid `--inet-button-light-text-color` / `--inet-button-light-bg-color` (which do exist and are used correctly elsewhere, e.g. `.btn-light` in `_bootstrap-override.scss`). Net effect: the console badge loses its color and background entirely — a visual regression on top of not fixing anything.

4. **Dead selector**: `.layout-dropdown ::ng-deep .custom-select-trigger { background-color: transparent; }` targets a `.custom-select-trigger` class/element that does not exist anywhere in the repo. The actual markup is a plain native `<select id="layoutDropdown" class="layout-dropdown form-control form-control-sm">` with no such wrapper — this rule can never match anything and is dead code as written.

### Worth double-checking before merge (non-blocking, lower confidence — could not render to verify)

5. `.console-count-badge { top: 0; }` (specificity `0,0,1,0`) is likely still overridden by Bootstrap's own `.btn .badge { top: -1px; }` (specificity `0,0,2,0`, no `!important` on either side), since the badge markup nests `.badge` inside `.btn`. If so, the "badge sits higher than the Console label" alignment issue may not actually be fixed by this rule. Worth a quick visual check in a running build.

### What looks correct

- `ws-header-cell.component.scss`: the `.ws-header` `display:flex; height:100%` change is sound — `.ws-header-cell` is a flex item of `.ws-table-headers` (`display:flex`, default `align-items:stretch`, explicit `[style.height.px]="HEADER_CELL_HEIGHT"`), so the host does get a definite stretched height for `height:100%` to resolve against. The accompanying `line-height: normal` fix on `.ws-header-column__display-container` and the `.mini-menu`/`.column-buttons` `align-items`/`display` fixes are consistent with the stated root cause.
- `variable-editor.component.scss` `flex-wrap: wrap` + `row-gap` change is low-risk and reasonable.
- `.shell-inline-form`'s `align-items: center` addition for the Layout label/dropdown is a reasonable, low-risk fix.
- No tests reference the removed/changed classes or properties, so nothing in the existing test suite breaks — but this also means nothing caught the above defects.
- No backward-compatibility or security concerns; this is a pure front-end CSS/template change with no API or persisted-data impact.

### Decision

Per the review checklist:
- **No blocking code issues** — **FAILED**: see items 1–4 above (condition-field-combo caret/ring effectively non-functional; console badge coloring regressed to undefined variables).
- **No failing checks (excluding `claude-review`)** — met (Semgrep OSS, Semgrep Scan, and `build` all pass).
- **No merge conflicts** — met (`mergeable: MERGEABLE`).

Since condition 1 fails, this PR should not be approved as-is.

**Recommendation:** fix the `dropdownOpen` binding to actually track open/closed state, swap `chevron-up-icon`/`chevron-down-icon` for the existing `caret-up-icon`/`caret-down-icon` aliases (or otherwise verify the intended glyph exists), fix or add the `--inet-focus-ring`/`--inet-focus-ring-outline` variables (or replace with the existing `custom-select:focus` recipe's real variables), restore/fix the console badge's color variables, and drop (or fix the target of) the dead `.custom-select-trigger` selector.

swaker854 (commented):
## Code Review: PR #4828 — Fix composer status band alignment and condition-dialog visual gaps

This PR bundles three small, well-explained composer CSS/template fixes (worksheet table header centering, viewsheet status-band alignment, condition-field-combo caret/focus ring) on top of an already-open stacked branch. I independently re-verified the current code against every "blocking" claim in the prior automated review (`jshobe-inetsoft`, submitted 2026-08-28T13:50Z) before writing this one.

### The prior review's four "blocking issues" do not hold up

I checked each against the actual current files on this branch:

1. **"`dropdownOpen` does not exist on `ConditionFieldComboComponent`"** — Incorrect. `condition-field-combo.component.ts:72` declares `dropdownOpen: boolean = false;` and `dropDownOpenChange()` (line 327-328) sets `this.dropdownOpen = open;`. This field predates this PR (added in #3980, already merged into the base branch) — the PR correctly binds the new template markup to it.
2. **"`chevron-up-icon`/`chevron-down-icon` are not real icon classes"** — Incorrect. They're defined in `web/projects/portal/src/scss/_icons.scss`'s `$ineticons-map` (`chevron-down-icon: $downward-icon`, `chevron-up-icon: $upward-icon`), which generates real `.chevron-*-icon` / `.chevron-*-icon::before` rules via an `@each` loop. The prior review only checked `_icon-alias.scss` and `ineticons/variables.scss` and missed this second, separate icon-alias source file.
3. **"Undefined CSS custom properties... these declarations are no-ops"** — Incorrect. `--inet-focus-ring`, `--inet-focus-ring-outline` (`_variables.scss:601-602`), `--inet-text-muted-color` (`_variables.scss:268`), and `--inet-shell-surface-hover` (`_variables.scss:291`) are all defined and already used across dozens of other component stylesheets in this codebase.
4. **"Dead selector: `.custom-select-trigger` targets an element that doesn't exist"** — Incorrect. The layout dropdown (`viewsheet-pane.component.html:162`) is `<custom-select id="layoutDropdown" ...>`, not a native `<select>`; `.custom-select-trigger` is markup rendered internally by that component. The identical `::ng-deep .custom-select-trigger` scoping pattern is already used successfully in ~10 other components in this repo.

Given all four are factually wrong against the current code, the PR's own description of its reasoning matches what's actually implemented.

### My own findings

**Non-blocking, worth a quick check before merge:**
- `viewsheet-pane.component.scss`'s `.console-count-badge { top: 0; }` (specificity 0,0,1,0) is nested inside `.btn` in the markup, and Bootstrap 5's vendored `.btn .badge { position: relative; top: -1px; }` (specificity 0,0,2,0) has strictly higher specificity, so it wins regardless of source order — the new `top: 0` likely doesn't actually take effect, and the "badge sits higher than the Console label" bug this hunk targets may persist. (This was the one legitimate, if non-blocking, point in the prior review.) Suggest `.btn .console-count-badge { top: 0; }` or nesting it under a more specific selector.

**Everything else checks out:**
- `ws-header-cell.component.scss` centering fix is consistent with the surrounding flex/stretch context and the explicit `HEADER_CELL_HEIGHT` ancestor height.
- Console button restyle (`btn btn-default rounded-pill`) now matches between `viewsheet-pane` and `ws-details-pane`, and the new badge color variables are real and already used elsewhere.
- Layout-dropdown transparency is correctly scoped to `.layout-dropdown` only, no bleed into other `custom-select` usages.
- `variable-editor`'s `flex-wrap: wrap` change is low-risk and scoped to the viewsheet-only checkbox row.
- The remainder of the diff (inherited from the stacked native-elements-sweep commits) is unrelated, already-iterated-on work — spot-checked `chat.service.ts`/`composer-toolbar.component.ts` (`Notification` type and `notify()` wiring are correct), `range-slider-edit-dialog` (`date-value-editor`/`time-instant-value-editor` are established `ControlValueAccessor` components), and `generic-multi-select-list` (its rebuild-from-`values`-order fix correctly matches native `<select multiple>` emit order). No regressions, no security concerns (pure front-end template/CSS/service-signature changes), and no backward-incompatibility. No test file references the removed markup/classes, so no existing test breakage is expected.

### Checklist
- **Blocking code issues:** none found on independent re-verification — the previously-reported ones do not reproduce.
- **Failing checks:** `claude-review` currently shows FAILURE, but per above its stated reasons are incorrect; `build`, `Semgrep Scan`, and `Semgrep OSS` all pass. Recommend dismissing the stale review and re-running the check.
- **Merge conflicts:** none (`mergeable: MERGEABLE`).

Since the "no failing checks" condition isn't currently satisfied (pending the stale review being dismissed/re-run), marking this as requesting changes per the review checklist, with the one actionable item being the badge specificity fix above plus clearing the stale check.

jshobe-inetsoft (approved):
## Code Review: PR #4828 — Fix composer status band alignment and condition-dialog visual gaps

I independently re-verified this PR against the actual current source (not just the diff), including re-checking every claim made in the two prior reviews on this PR (one `CHANGES_REQUESTED` from `jshobe-inetsoft`, one `COMMENTED` rebuttal from `swaker854`). Findings below are based on my own reading of the files, not a restatement of either prior review.

The PR bundles four real commits (the rest of the diff shown on the PR is inherited, already-merged content from the stacked `#4816` branch and shows no net changes):

1. **`ws-header-cell.component.scss`** — centers the worksheet table header label/icon/menu button. Verified: `.ws-table-headers` sets an explicit `[style.height.px]="HEADER_CELL_HEIGHT"` (30px) ancestor height, so `.ws-header { height: 100% }` resolves against a definite value; `.mini-menu` now actually gets `display: flex` (previously `align-items: center` on it was dead without a flex context); `line-height: normal` correctly neutralizes the inherited `line-height: 24px` that was inflating `.ws-header-column__display-container`'s line box independent of its text. Sound, and well commented in the SCSS itself.
2. **`viewsheet-pane.component.scss`/`.html`** — Console button restyle to `btn-default rounded-pill` (now consistent with `ws-details-pane`'s equivalent button), `.shell-inline-form` gets `align-items: center`, and the layout `<custom-select>` trigger is made transparent via `.layout-dropdown ::ng-deep .custom-select-trigger`. Confirmed `<custom-select>` (not a native `<select>`) is used at `viewsheet-pane.component.html:162`, and `.custom-select-trigger` is a real class rendered by `custom-select.component.html` — the identical `::ng-deep .custom-select-trigger` scoping pattern is already used in ~10 other components (e.g. `condition-item-pane.component.scss`, `variable-editor.component.scss`).
3. **`condition-field-combo.component.scss`/`.html`** — Adds the missing dropdown caret (`<i class="condition-field-combo-icon dropdown-caret icon-color-default" [class.chevron-up-icon]="dropdownOpen" [class.chevron-down-icon]="!dropdownOpen">`) and the `.is-open` focus-ring styling. Verified: `dropdownOpen: boolean = false` is a real field on `ConditionFieldComboComponent` (line 72), set by `dropDownOpenChange()` (line 328); `chevron-up-icon`/`chevron-down-icon` are real generated classes from the `$ineticons-map` `@each` loop in `_icons.scss` (lines 30-31); `--inet-focus-ring`, `--inet-focus-ring-outline`, `--inet-text-muted-color`, and `--inet-shell-surface-hover` are all defined in `_variables.scss` and used across dozens of other stylesheets already. None of these are dead references.
4. **`variable-editor.component.scss`** — `flex-wrap: nowrap` → `wrap` (+`row-gap`) fixes the viewsheet-only "Use Selection List" checkbox overflow. Low-risk, scoped change.

**Important correction to the PR's review history:** the prior `CHANGES_REQUESTED` review's four "blocking issues" (missing `dropdownOpen` field, non-existent `chevron-*-icon` classes, undefined CSS custom properties, and a "dead" `.custom-select-trigger` selector) do **not** hold up — I checked each one directly against the current file contents (cited above) and all four claims are factually incorrect. The subsequent `swaker854` review already identified this; I've now confirmed it independently rather than taking either review's word for it.

### Clean code / readability

Good: the SCSS changes include explanatory comments describing *why* each rule is needed (e.g. the `.ws-header` height comment, the `line-height: normal` comment), which is genuinely helpful for future maintainers given how non-obvious some of this CSS cascade interaction is. Scoping is appropriately narrow (`.layout-dropdown ::ng-deep .custom-select-trigger`, not a global `custom-select-trigger` override).

Pure template/CSS changes — no new control flow, no new error paths introduced.

### Security

No security concerns. No new user input handling, no new DOM injection points (`aria-hidden="true"` correctly applied to the new decorative caret icon), no API/backend surface touched.

### Regression risk / backward compatibility

- No backend/API/persisted-data changes; this is Angular template + component-scoped SCSS only.
- No test file asserts on the removed/changed classes or the previous `btn-primary` console button styling — `condition-field-combo.component.tl.spec.ts` instantiates the component directly and doesn't render/assert on the new `<i>` caret markup, so it isn't affected by the template change. No test breakage found.
- **One legitimate, non-blocking item worth a visual double-check before/after merge:** `viewsheet-pane.component.scss`'s new `.console-count-badge { top: 0; }` has specificity `(0,1,0)` (or `(0,2,0)` once Angular's per-component `[_ngcontent-*]` attribute selector is factored in), while Bootstrap's vendored `.btn .badge { position: relative; top: -1px; }` (`_badge.scss`) is `(0,2,0)` with no encapsulation attribute. Whether the new rule actually wins depends on Angular's style-injection order relative to the global Bootstrap bundle, which I can't confirm without rendering. If it doesn't win, the "badge sits higher than the Console label" alignment bug this hunk targets may persist. Not a regression and not blocking, but suggest a quick visual sanity check (or a defensive `.btn .console-count-badge` selector) before considering this particular sub-fix done.

### Decision checklist

- **No blocking code issues** — met, on independent re-verification (see above; the previously-reported blocking issues do not reproduce against actual source).
- **No failing checks (excluding `claude-review`)** — met: `build`, `Semgrep Scan`, and `Semgrep OSS` all pass. `claude-review` is failing but is known-broken infrastructure per current guidance and is excluded from this evaluation.
- **No merge conflicts** — met (`mergeable: MERGEABLE`).

All three conditions are satisfied, so this review approves the PR. (Note: the PR is currently `BLOCKED` from merging only because the repo's ruleset requires 1 approving review and the prior review left it in `CHANGES_REQUESTED` state — this approval should clear that once submitted.)

PR #4829, opened 2026-08-27
Bug #76346, expose worksheet table Mergeable and Visible-in-Viewsheet flags to the AI agent API

jkidacki-inetsoft:
## Summary

The AI agent's worksheet-editing REST API (`set_table_properties`, backing the `stylebi-composer-chat` plugin's tool of the same name) previously supported only `newName`/`description`/`maxRows`/`distinct`, with no way to read or write a worksheet table's Advanced **"Mergeable"** checkbox or its **"Visible in Viewsheet"** checkbox — both already supported by the Composer's own table-properties dialog via `TableAssembly.setSQLMergeable`/`setVisibleTable`.

## Root Cause

`EditRequest`, `WorksheetAgentController.dispatch()`, and `WorksheetEditService.Editor#setTableProperties` simply never carried these two fields, and the read-side `WorksheetModel.TableModel` / `WorksheetReadService` never reported `mergeable` (it already reported `visibleInViewsheet` for read).

## Fix

- Added `mergeable` and `visibleInViewsheet` as optional fields on `EditRequest` (new backward-compatible constructor overloads preserve every existing caller).
- Threaded both through `WorksheetAgentController`'s dispatch and `WorksheetEditService.Editor#setTableProperties` — `null` continues to mean "leave unchanged", matching `distinct`.
- Added `mergeable` to `WorksheetModel.TableModel` / `WorksheetReadService` so the read side reports both flags.

## Test Plan

- [x] `mvnw -pl core -am test-compile` — clean compile
- [x] `mvnw -pl core test -Dtest=WorksheetEditServiceMutatorsTest,WorksheetReadServiceTest,WorksheetAgentControllerTest` — 264/264 pass, including new tests for both flags (set + omit-leaves-unchanged cases)
- [x] Code review pass (2 rounds) with no blocking issues

Fixes Redmine #76346.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
@@ -78,7 +78,8 @@ import java.util.Map;
  *   <li>{@code insert_row} — {@code table}, {@code index}</li>
  *   <li>{@code delete_row} — {@code table}, {@code index}</li>
  *   <li>{@code set_table_properties} — {@code table}; any of {@code newName} (or {@code alias},
- *       its accepted spelling), {@code description}, {@code maxRows}, {@code distinct}</li>
+ *       its accepted spelling), {@code description}, {@code maxRows}, {@code distinct},
+ *       {@code mergeable}, {@code visibleInViewsheet}</li>
  *   <li>{@code add_cross_join} — {@code name}, {@code leftTable}, {@code rightTable}</li>
  *   <li>{@code add_merge_join} — {@code name}, {@code tables}</li>
  *   <li>{@code reorder_columns} — {@code table}, {@code columnOrder}</li>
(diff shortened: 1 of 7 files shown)


jshobe-inetsoft (approved):
## Review Summary

This PR threads two previously-missing table-level flags — `mergeable` (`TableAssembly#setSQLMergeable`) and `visibleInViewsheet` (`TableAssembly#setVisibleTable`) — through the AI agent's `set_table_properties` worksheet-edit operation, on both the write path (`EditRequest` → `WorksheetAgentController.dispatch()` → `WorksheetEditService.Editor#setTableProperties`) and the read path (`WorksheetReadService` → `WorksheetModel.TableModel`). This brings the agent API to parity with what the Composer's own table-properties dialog already exposes via `TablePropertyDialogService` (`assembly.setSQLMergeable(...)` / `assembly.setVisibleTable(...)`).

**Overall assessment: solid, low-risk, well-tested change. No blocking issues found.**

### Correctness

- The new fields are threaded consistently and in the same argument order everywhere they appear (`EditRequest.mergeable()/visibleInViewsheet()` → `WorksheetAgentController` dispatch call → `WorksheetEditService.Editor#setTableProperties(table, newName, description, maxRows, distinct, mergeable, visibleInViewsheet)`). No argument-order transpositions.
- `null`-means-"leave unchanged" semantics match the existing `distinct` field exactly, and are applied *after* the rename, preserving the documented all-or-nothing behavior of the method (rename first, so a rename failure leaves nothing else half-applied).
- The setters used (`TableAssembly#setSQLMergeable` / `#setVisibleTable`) are simple field mutators with no other side effects in `AbstractTableAssembly`/`TableAssemblyInfo`, matching exactly what `TablePropertyDialogService` already does for the Composer UI — no divergent behavior introduced between the two write paths.
- The read side change (`WorksheetReadService.readTable`) adds `t.isSQLMergeable()` in the correct position to match the `WorksheetModel.TableModel` record's field order (`distinct, mergeable, visibleInViewsheet, mode, ...`) — verified against the record declaration; no positional mismatch.
- `WorksheetModel.TableModel` is a `@JsonInclude(NON_NULL)`-annotated record; adding a `boolean mergeable` field is purely additive to the JSON payload and won't affect existing readers' handling of the other fields.

### Backward compatibility

- `EditRequest` is a record with a long history of "compatibility constructor" overloads for backward-compatible field additions. This PR follows the established pattern correctly: it adds two *new* compat constructors (one accepting all fields through `mergeable` for pre-`visibleInViewsheet` callers, one through `joinPaths` for pre-`mergeable` callers) rather than mutating any existing constructor's signature. All pre-existing compat constructors (verified against `WorksheetAgentControllerTest`, which still constructs `EditRequest` via an older compat constructor without `crosstab`/`labels`/`choices`/`joinPaths`/`mergeable`/`visibleInViewsheet`) are untouched and still compile/behave identically.
- The only production caller of `WorksheetEditService.Editor#setTableProperties` (`WorksheetAgentController.dispatch`) was updated in lockstep with the method signature change, so there's no orphaned caller.
- Net effect: fully additive and backward compatible for both the REST payload shape and the internal Java API.

### Tests

- New unit tests cover both the "set" and "omit leaves unchanged" cases for each new flag (`setTablePropertiesSetsTheMergeableFlag`/`omittingMergeableLeavesTheExistingFlagAlone`, and the `visibleInViewsheet` equivalents), plus an update to the existing "applies the other fields" test to also assert the two new flags. `WorksheetReadServiceTest` was updated to assert `mergeable()` on the read side.
- These are meaningful behavioral tests (not just compilation smoke tests) — they verify the null-omission contract as well as the positive-set contract, which is the crux of this feature.
- Nothing in the existing test suite appears to conflict with or be broken by this change (per the PR's stated test run: 264/264 passing), and this is corroborated by the CI `build` check passing.

1. **Pre-existing duplicate Javadoc block on `setTableProperties`** (`WorksheetEditService.java`, ~line 1842): there are two adjacent `/** ... */` blocks immediately before the method — the first (older) one, documenting `@param`s including the two new ones this PR adds, and a second one immediately after it (the actual attached Javadoc, since only the comment immediately preceding a declaration is associated with it). This means the `@param mergeable` / `@param visibleInViewsheet` entries added by this PR land in the *orphaned* comment block and won't show up in generated docs or IDE tooltips. This duplication predates this PR (visible in the diff context as already present), so it's not a regression introduced here, but since this PR touches that exact spot, it would have been a good opportunity to fold the two blocks together. Purely cosmetic — no functional impact. Not blocking.
2. The external AI-plugin tool schema (`stylebi-composer-chat`, referenced in the PR description) that presumably needs to advertise these two new fields to the LLM tool-calling interface lives outside this repository and isn't touched here — worth confirming as a follow-up so the two flags are actually reachable end-to-end from the chat UI, but it's out of scope for this PR/repo.

### Security

No security concerns. This only exposes two boolean configuration flags that are already fully user-controllable through the Composer UI's table-properties dialog, gated behind the same permission/session-token checks as the rest of `set_table_properties`. No new attack surface.

### Regressions

None identified. The change is additive at every layer (new optional fields, new compat constructors, new read-model field), and the only behavior change to existing code paths is that `set_table_properties` now also honors two previously-ignored fields when present — which is the intended fix, not a side effect.

---

## Merge Readiness

- **Code review:** No blocking issues — approve.
- **Checks:** All checks other than `Claude Code Review / claude-review` (excluded per review policy, since this manual review supersedes it) are passing: `Semgrep OSS`, `Semgrep Scan`, `build`, and `license/cla` all pass.
- **Merge conflicts:** None — `mergeable: MERGEABLE`.

All three gating conditions are satisfied. **Approving.**

PR #4830, opened 2026-08-27
Bug #76018, re-save runtime partition after shrinkColumnHeight in physical model editor

jkidacki-inetsoft:
## Summary

In the Physical Model (physical view) editor, applying Auto Layout, saving, and reopening the view showed the layout reverted instead of preserved.

## Root Cause

`RuntimePartitionService` caches `RuntimeXPartition` objects in a serializing cache, so any `cache.get()` returns a deserialized copy disconnected from an in-memory object mutated after the cache write (the same class of bug previously fixed in #75893 for Bug #75893).

`PhysicalModelManagerService.openModel()` and `.createModel()` both:
1. Build a `RuntimeXPartition` from the persisted (Studio-scale) partition via `RuntimePartitionService`, which caches it immediately.
2. Call `graphService.shrinkColumnHeight(...)` on that same in-memory object, converting Studio-scale (full column-height) table positions to compact Portal-scale positions — but never save that mutation back to the cache.
3. Return a `PhysicalModelDefinition` built directly from the correctly-shrunk in-memory object, so the very first page paint looks right.
4. The frontend immediately issues a follow-up `POST /api/data/physicalmodel/graph` request, which reads the partition back out of the cache — getting the stale, un-shrunk copy. This renders the wrong (stretched-out) layout, and is also what a later Save reads and persists, discarding whatever layout (including one just set via Auto Layout) the user actually saw.

Confirmed via live reproduction with temporary diagnostic logging (since removed): the raw partition bounds correctly round-tripped through shrink, but the graph model actually returned to the client used stale (un-shrunk) values.

## Fix

Re-save the runtime partition to the cache immediately after `shrinkColumnHeight(...)` in both `openModel()` and `createModel()`, so the cache reflects the shrunk state before any other reader can observe it. A code-reviewer pass confirmed the other `shrinkColumnHeight`/`unfoldColumnHeight` call sites already have a proper save afterward, so these were the only two gaps.

## Test Plan

- [x] `core` module builds cleanly (`clean test-compile`)
- [x] Existing `PhysicalGraphServiceTests` (5 tests) still pass
- [x] Manually reproduced and confirmed fixed: open a physical view, Auto Layout → Vertical, Save, reopen — layout now persists correctly

Fixes Redmine #76018.

### DIFF
--- a/core/src/main/java/inetsoft/web/portal/controller/database/PhysicalModelManagerService.java
+++ b/core/src/main/java/inetsoft/web/portal/controller/database/PhysicalModelManagerService.java
@@ -110,6 +110,13 @@ public class PhysicalModelManagerService {
       // for table location of portal <--- studio
       graphService.shrinkColumnHeight(runtimeXPartition);
 
+      // RuntimePartitionService.openModel() cached runtimeXPartition before the shrink
+      // above was applied, and the cache stores a serialized copy -- so without saving
+      // again here, every other reader of this runtime id (e.g. the graph-model fetch
+      // the client issues right after opening the view) would see the un-shrunk,
+      // Studio-scale positions instead of the compact Portal ones.
+      this.runtimePartitionService.saveRuntimePartition(runtimeXPartition);
+
       return physicalModelService.createModel(dataModel, runtimeXPartition);
    }
 
@@ -169,6 +176,11 @@ public class PhysicalModelManagerService {
 
       // for table location of portal <--- studio
       graphService.shrinkColumnHeight(rp);
+
+      // see the identical comment in openModel(): the cache entry for rp was written
+      // before this shrink, and stores a serialized copy, so it must be re-saved here
+      // or other readers of this runtime id will see the un-shrunk positions.
+      this.runtimePartitionService.saveRuntimePartition(rp);
 
       return physicalModelService.createModel(dataModel, rp);
    }


jshobe-inetsoft (approved):
## Code Review Summary

**PR:** Bug #76018 — re-save runtime partition after `shrinkColumnHeight` in physical model editor
**Files changed:** `core/src/main/java/inetsoft/web/portal/controller/database/PhysicalModelManagerService.java` (+12/-0)

### Correctness

The fix is correct and well-targeted. Tracing the flow:

- `RuntimePartitionService.openModel()` / `.createModel()` build a `RuntimeXPartition`, call `saveRuntimePartition()` (which does `cache.put(...)`), and return that same in-memory object.
- `PhysicalModelManagerService.openModel()` / `.createModel()` then call `graphService.shrinkColumnHeight(...)`, which mutates the `XPartition` in place (confirmed in `PhysicalGraphService.shrinkColumnHeight(RuntimeXPartition)` — it calls `runtimeXPartition.getPartition()` and mutates table bounds on that same `XPartition` reference via `doShrinkColumnHeight`).
- Since the underlying cache (`javax.cache.Cache` backed by the cluster/Ignite layer) stores serialized copies, the mutation performed after the initial `cache.put()` is invisible to any subsequent `cache.get()` — including the client's immediate follow-up `GET /api/data/physicalmodel/graph` call and any later Save. This matches the previously-fixed bug #75893 pattern, and matches the established convention throughout the rest of this same file, where every other mutate-then-return helper (`addTable`, `removeTable`, `updateInlineView`, `updateAlias`, `addAutoJoin`, etc.) explicitly calls `runtimePartitionService.saveRuntimePartition(p)` after mutating.

The added `saveRuntimePartition(...)` calls in both `openModel()` and `createModel()` close exactly this gap. I checked `PhysicalModelService.createModel(dataModel, runtimeXPartition)` (the call immediately following each new save) and confirmed it only reads the partition to build a DTO — it does not mutate it further, so no additional save is needed after that call.

### Completeness

The PR description says a "code-reviewer pass confirmed the other `shrinkColumnHeight`/`unfoldColumnHeight` call sites already have a proper save afterward." I independently verified this against the current file: `createAndSaveModel(...)` and `updateAndSaveModel(...)` (the two other private helpers that call `shrinkColumnHeight`/`unfoldColumnHeight`) are invoked from within `.ifPresent(p -> { ...; runtimePartitionService.saveRuntimePartition(p); })` blocks in their public callers, so they're already covered. No other gaps found.

### Clean Code / Readability

- The added comments are clear, explain *why* (not just what), and correctly reference the sibling fix location for the second occurrence rather than duplicating the full explanation — good practice.
- Minor: the fix duplicates a one-line `saveRuntimePartition` call plus a multi-line comment in two places. Given the existing file already has this exact repeated pattern elsewhere (mutate then explicitly save) rather than centralizing it, this is consistent with the codebase's established style, so not a real objection — just noting it's not a novel abstraction, which is appropriate for a minimal, targeted bug fix per project guidelines.

No new error paths are introduced. `saveRuntimePartition` is a simple `cache.put` and cannot itself throw checked exceptions that need handling here; both call sites already run inside methods that declare `throws Exception`.

### Security

No security-relevant surface is touched — this only affects in-process cache consistency for an already-authorized, already-permission-checked runtime partition object. No new external input, serialization boundary, or trust boundary is introduced.

- Low regression risk. The change only adds a second, corrective cache write; it does not alter any method signature, external API, or return value.
- One subtlety worth flagging (not blocking): in `createModel()`, `RuntimePartitionService.createModel()` already calls `saveRuntimePartition` once (line 60 of `RuntimePartitionService`) before `shrinkColumnHeight` runs, and now it's saved a second time after. This is fine functionally (idempotent last-write-wins into the same cache key) but means the cache is written twice per `openModel`/`createModel` call — a negligible perf cost given this is a low-frequency, per-editor-session operation (not a hot path), so no real concern.
- No thread-safety change: the cache is presumably per-runtime-id keyed, and nothing here introduces new concurrent-mutation risk beyond what already existed.

### Tests

- No failing existing tests: `PhysicalGraphServiceTests` doesn't touch `PhysicalModelManagerService.openModel()`/`createModel()` directly, and the diff doesn't change `PhysicalGraphService` at all, so those 5 tests are unaffected by this change.
- Gap: no new automated test was added to cover the specific regression (cache re-read returning stale layout after shrink). The PR relies on manual reproduction only. This is a reasonable candidate for a follow-up unit/integration test (e.g. asserting `runtimePartitionService.getPartition(id)` returns shrunk bounds immediately after `openModel()`/`createModel()` returns), but I don't consider its absence blocking for a small, well-understood, low-risk fix like this.

### Conclusion

The change is minimal, correct, consistent with the codebase's existing save-after-mutate convention, and carries no discernible regression or security risk. No blocking issues found.

PR #4831, opened 2026-08-27
Bug #76000, skip imported bookmarks whose owner does not exist

ngovus-inetsoft:
## Problem

Importing a bookmark-containing asset package into an organization where the bookmark's owner has since been renamed creates a bookmark owned by a user that no longer exists — an orphan row in the bookmark list that no one can reach. No conflict prompt appears; the import just completes.

Reproduction from the report:

1. Site admin imports the package (a viewsheet `chart` + worksheet `chartUI`, exported from `host-org`).
2. Clone a new `org0` — `admin` and its bookmark `bk1` are copied over.
3. Rename `admin` → `test` in `org0`.
4. Import the same package into `org0`.

Result: two `bk1` rows — one owned by `test`, one owned by the now-nonexistent `admin`.

## Root cause

Bookmarks are not separate JAR entries; they ride inside the viewsheet's own JAR entry as an `<AllBookmarks>` block. Unpacking the attached package confirms it contains `<name>admin~;~host-org</name>` verbatim.

`ViewsheetAsset.parseContent0()` took that owner name as-is and only remapped the org ID:

```java
IdentityID identityID = IdentityID.getIdentityIDFromKey(name);   // "admin"
identityID.setOrgID(OrganizationManager.getInstance().getCurrentOrgID()); // -> org0
```

Nothing checked that `admin` exists in `org0`, and `setVSBookmark()` then wrote straight to `IndexedStorage` under that dead identity.

The missing prompt has the same source: `AbstractAssetEngine.getVSBookmark()` returns a **new empty** `VSBookmark` rather than `null` when nothing is stored, so `getBookmarkConflicts()` found no existing `bk1` for `admin` and emitted no conflict row.

The rename side is working correctly — `UserTreeService.renameUserAsset()` → `BlobIndexedStorage.migrateStorageData()` does move the bookmark key from `admin~;~org0` to `test~;~org0`. `admin` genuinely owned nothing at step 4; the import created the orphan from scratch.

`DeployService.validateUsers()` already applies exactly this existence check to user-scoped **asset owners** — bookmark owners were simply never fed into it.

## Fix

- New `ViewsheetAsset.bookmarkOwnerExists()`, consulted from both `parseContent0()` (import) and `DeployManagerService.getBookmarkConflicts()` (conflict scan) so the two cannot disagree.
- Skips just that owner's block with a warning — `continue`, not `return`, so other users' bookmarks in the same viewsheet still import.
- Anonymous / `_NULL_` owners are exempt, matching `DeployService.importAsset`.

Two details worth calling out for review:

**The guard tests `SecurityProvider.isVirtual()`, not `isSecurityEnabled()`.** `SecurityEngine.getSecurityProvider()` is `if(isSecurityEnabled() && provider != null) return provider; return vprovider;` — with security *enabled* but the provider not yet initialized it returns the virtual provider, which only recognizes `admin`/`system`/`anonymous`. Gating on `isSecurityEnabled()` would silently drop every other owner's bookmarks in that window. `AuthenticationChain.isVirtual()` reports true for an all-virtual or empty chain too, covering the same hole. Same pattern as `MVManager.java:463`.

**Every failure path fails open.** `parseContent0` declares `throws Exception`, so an escaping `SecurityEngine.getSecurity()` (Spring bean lookup) or `provider.getUser()` (LDAP/SSO) would fail the whole viewsheet import over auxiliary data. Both are inside the try, mirroring the existing `getAllUsers()` in the same class.

The unregistered-user exemption that `validateUsers` applies is deliberately **not** mirrored: an unregistered user is a `sreeUserData` file with no security identity, which is the orphan case itself. (Separately, `SUtil.getUnregisteredUsers()` builds a `Set<String>` and returns it via `toArray(new IdentityID[0])`, which throws `ArrayStoreException` whenever the set is non-empty — a pre-existing latent defect, not touched here.)

## Testing

New `ViewsheetAssetBookmarkOwnerTest` (8 tests) covers the renamed-away owner, a surviving owner in the same block, anonymous/`_NULL_`, a virtual provider, a null provider, null/blank owners, and both throw paths. Both new guards were mutation-tested — removing the `isVirtual()` check or the try/catch fails the corresponding tests. `ImportAssetControllerTest` still passes (13/13 total).

### DIFF
--- a/core/src/main/java/inetsoft/sree/internal/DeployManagerService.java
+++ b/core/src/main/java/inetsoft/sree/internal/DeployManagerService.java
@@ -1080,6 +1080,13 @@ public class DeployManagerService {
 
                IdentityID userID = IdentityID.getIdentityIDFromKey(userName);
                userID.setOrgID(OrganizationManager.getInstance().getCurrentOrgID());
+
+               // Bookmarks for an owner that does not exist in this organization are skipped
+               // by ViewsheetAsset.parseContent0(), so never prompt to resolve them.
+               if(!ViewsheetAsset.bookmarkOwnerExists(userID)) {
+                  continue;
+               }
+
                VSBookmark existing = engine.getVSBookmark(entry, new XPrincipal(userID));
 
                if(existing == null) {
--- a/core/src/main/java/inetsoft/util/dep/ViewsheetAsset.java
+++ b/core/src/main/java/inetsoft/util/dep/ViewsheetAsset.java
@@ -497,6 +497,18 @@ public class ViewsheetAsset extends AbstractSheetAsset implements FolderChangeab
 
          IdentityID identityID = IdentityID.getIdentityIDFromKey(name);
          identityID.setOrgID(OrganizationManager.getInstance().getCurrentOrgID());
+
+         // The owner is carried verbatim in the JAR and only has its org remapped above. If
+         // that user does not exist here (renamed, deleted, or never present in this org),
+         // importing the block would create bookmarks no one can reach. Skip just this owner
+         // -- other users' bookmarks in the same <AllBookmarks> block must still import.
+         if(!bookmarkOwnerExists(identityID)) {
+            LOG.warn("Skipping imported bookmarks for viewsheet '{}': user '{}' does not " +
+               "exist in organization '{}'", entry.getPath(), identityID.getName(),
+               identityID.getOrgID());
+            continue;
+         }
+
          XPrincipal principal = new XPrincipal(identityID);
          VSBookmark existing = engine.getVSBookmark(entry, principal);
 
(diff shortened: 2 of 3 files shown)


jshobe-inetsoft (approved):
## Summary

This PR fixes Bug #76000: importing a bookmark-containing viewsheet into an org where the bookmark's original owner no longer exists (renamed, deleted, or never present) silently creates an orphan bookmark row that no user can ever reach, with no conflict prompt shown. The fix adds a single guard, `ViewsheetAsset.bookmarkOwnerExists(IdentityID)`, consulted from both the import path (`ViewsheetAsset.parseContent0`) and the pre-import conflict scan (`DeployManagerService.getBookmarkConflicts`), so the two can't disagree.

I read the full surrounding methods (`ViewsheetAsset.parseContent0`, `writeContent0`, `DeployManagerService.getBookmarkConflicts`, `DeployService.validateUsers`/`importAssets`, `SecurityEngine.getSecurityProvider()`/`isSecurityEnabled()`, `AuthenticationProvider.isVirtual()`/`AbstractSecurityProvider.isVirtual()`, `IdentityID`), not just the diff, to check correctness and regression risk.

## Correctness

- The root cause analysis is accurate: `SecurityEngine.getSecurityProvider()` (line ~1165-1173) is exactly `if(isSecurityEnabled() && provider != null) return provider; return vprovider;` — confirming the PR's claim that gating on `isSecurityEnabled()` instead of `provider.isVirtual()` would be wrong (it would return the virtual provider, which only knows admin/system/anonymous, while looking like a "real" check). The `isVirtual()` check is the right one, and it mirrors an existing pattern in `SRIdentityFinder.java:331` (`!SecurityEngine.getSecurity().getSecurityProvider().isVirtual()`), so it's idiomatic for this codebase.
- `continue` (not `return`) is used correctly in both `parseContent0` and the new check — verified the surrounding loop bodies; skipping one owner's `<user>` block does not abort processing of the remaining `<user>` entries in the same `<AllBookmarks>` block, matching the PR's claim.
- The org-remap (`identityID.setOrgID(currentOrgID)`) happens before the existence check in both call sites, so the check is performed against the correct target-org identity — consistent with how the rest of the method (and `engine.getVSBookmark`) already treats identity.
- The anonymous/`_NULL_` exemption matches the existing exemption in `DeployService.validateUsers`'s caller (`noUsers` check using the same literals), so behavior is consistent with the established asset-owner validation path.
- Fail-open behavior (null/blank owner, no provider, provider throws) is justified and consistent: `parseContent0` declares `throws Exception`, and an uncaught exception here would fail an entire viewsheet import over auxiliary bookmark data, which would be a worse regression than the bug being fixed.
- `DeployManagerService.getBookmarkConflicts` and `ViewsheetAsset.parseContent0` independently parse the same jar's `<AllBookmarks>` XML (by design, per existing code structure) — routing both through the same static `bookmarkOwnerExists` helper is the right way to keep them from disagreeing, exactly as described.
- Import in `DeployManagerService.java` doesn't need adjustment since `inetsoft.util.dep.*` is already wildcard-imported there.

## Tests

- New `ViewsheetAssetBookmarkOwnerTest` exercises the static helper directly with 8 focused cases (renamed-away owner, surviving owner, anonymous/`_NULL_`, virtual provider without consulting `getUser()`, null/blank owner, no provider, and both throw paths). Mockito usage (`mockStatic` with `Strictness.LENIENT`, `withSettings().lenient()`) matches existing conventions elsewhere in the test suite (e.g. `EMScheduleTaskControllerTest`, `ScheduleCycleServiceTest`).
- Gap (non-blocking): there's no integration-level test that actually drives `parseContent0` or `getBookmarkConflicts` end-to-end to confirm a multi-owner `<AllBookmarks>` block imports the surviving owner's bookmarks while skipping the missing one, or that the conflict scan and import agree on a real fixture. Given `ImportAssetControllerTest` is reported to still pass at 13/13 and the two call sites share one code path, the risk is low, but an end-to-end regression test would have directly reproduced the original bug report's repro steps.

## Minor / non-blocking observations

1. **Logging asymmetry**: `parseContent0` logs a `LOG.warn` when skipping an owner; the equivalent skip added in `DeployManagerService.getBookmarkConflicts` does not log anything. Not a functional issue (the scan runs before import, so a second warning there could just be noise), but worth a conscious choice rather than an oversight.
2. **Per-owner provider lookup**: `bookmarkOwnerExists` calls `provider.getUser(owner)` once per bookmark owner per viewsheet, with no request-scoped caching/deduplication (unlike `DeployService.validateUsers`, which pre-collects a `Set<IdentityID>`). For a large import package with many viewsheets sharing the same owners, and a slow external provider (LDAP/SSO), this could add noticeable overhead. Likely fine at typical import sizes, but worth keeping in mind if this path is ever reported as slow.
3. The Javadoc on `bookmarkOwnerExists` is thorough and clearly documents the fail-open rationale and the `isVirtual()` vs `isSecurityEnabled()` distinction — this will save the next reader real time; good practice given how subtle the security-provider fallback behavior is.

## Backward compatibility / regressions

- No public API signatures changed; `bookmarkOwnerExists` is a new additive `public static` method.
- Behavior change is intentional and scoped: bookmarks whose owner doesn't exist in the target org are now skipped (with a log warning) instead of silently written as unreachable orphans. Existing valid bookmarks for existing owners are unaffected — verified the check sits after the org remap and before the existing merge/write logic, and only affects the branch where the owner doesn't resolve.
- Fail-open on transient provider errors preserves pre-existing behavior for that edge case (bookmark still gets written), so it does not introduce a new failure mode beyond what already existed before this PR.

## Security

- No new attack surface: the check only performs a read-only user-existence lookup keyed off attacker-influenced XML content (the bookmark owner name from the imported jar), which was already parsed into an `IdentityID` and used to look up/write bookmarks before this change. No injection risk introduced.

## Verdict

I did not find any blocking correctness, regression, or security issues. The fix is minimal, well-targeted, consistent with existing patterns in the codebase (`SRIdentityFinder`, `DeployService.validateUsers`, `MVManager`), and backed by focused unit tests plus a passing full build. The observations above are suggestions for future hardening, not blockers.

**Recommendation: Approve.**

PR #4834, opened 2026-08-28
Seed units on a new print layout, not just scaleFont

lizhepu:
A `set_print_layout` call on a viewsheet that has no print layout yet returns a 500 **and leaves a half-written layout behind**. After that, every later layout write on that viewsheet fails — `manage_device_layout` included — and a caller has no way to repair it.

One line fixes it, next to a seed that already exists for exactly this class of hazard.

## Root cause

`PrintDeviceLayoutPropertyService` seeds a brand-new `VSPrintLayoutDialogModel` with `scaleFont` and **not** `units`. The comment above that seed explains why `scaleFont` is seeded — the Hazard-3 regression, where an unset field carried a bare `0.0f` into the write. **The same reasoning applies to `units` and was never extended to it.**

The null then travels:

```
PrintDeviceLayoutPropertyService  (units left null on a fresh model)
  -> ViewsheetPropertyDialogService:415   printInfo.setUnit(vsPrintLayoutDialog.getUnits())
  -> VSLayoutService:934                  String unit = layout.getPrintInfo().getUnit()
  -> VSLayoutService:959                  switch(unit)   // only "inches" and "mm" -> NPE on null
```

The throw lands **after** the patch has already mutated the live model, which is why the failed call persists in full. Every later call then fails *earlier*, while re-reading the broken layout — so later writes land nothing, and `manage_device_layout` dies too, since both go through `setViewsheetInfo`.

## Verified live, before and after

Same sheet, same clean starting state (`list_layouts` → `layouts: []`), same patch. **`units` is the only variable:**

| | without `units` | with `units` |
|---|---|---|
| patch | `{paperSize: "Letter", scaleFont: 0.8}` | `{paperSize: "Letter", scaleFont: 0.8, units: "inches"}` |
| result | server error | **`{ok: true}`** |
| `get_layout` after | layout created, `units: null` — the wreckage | `printSettings` correct, `units: "inches"` |
| `manage_device_layout create` after | same server error, tool dead | **succeeds** |

## Why `"inches"`

It is one of only two values `VSLayoutService.getPLayoutSize`'s switch recognises, and `WizPrintLayoutBuilder.java:203` already hardcodes `info.setUnit("inches")` for the same purpose elsewhere in the wiz tier.

## Note for reviewers

The comment block was rewritten rather than appended to, so it now explains **both** seeds and why each unset value is fatal rather than merely wrong. That is the whole of the non-functional diff.

A caller-side workaround exists and is proven, for anyone on a build without this fix: **pass `units` on the call that first creates a print layout.** Passing it afterwards does not help — by then the NPE fires before the value is read.

Found by the L11 lane of the composer plugin test plan (Finding 5). Full trail, including a correction to an earlier diagnosis that reached the wrong mechanism, is in `stylebi-wiz`'s `docs/teams/2026-08-28-test-l11-print-layout/case-f5/`.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/PrintDeviceLayoutPropertyService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/PrintDeviceLayoutPropertyService.java
@@ -114,8 +114,8 @@ public class PrintDeviceLayoutPropertyService {
          }
 
          applyPrintLayoutPatch(printLayout, resolvedPatch);
-         requireUnits(printLayout);
          screensPane.setPrintLayout(printLayout);
+         requireUsableUnits(screensPane);
          dialogService.setViewsheetInfo(runtimeId, model, user, dispatcher, linkUri, null);
       });
    }
@@ -251,6 +251,11 @@ public class PrintDeviceLayoutPropertyService {
          }
 
          screensPane.setDeviceLayouts(layouts);
+         // Not incidental: this method never touches the print layout, but setViewsheetInfo
+         // computes the page size from it unconditionally, so a viewsheet carrying a persisted
+         // null unit takes manage_device_layout down with it. Guarding only setPrintLayout above
+         // would leave that viewsheet permanently unusable with no caller-side repair.
+         requireUsableUnits(screensPane);
          dialogService.setViewsheetInfo(runtimeId, model, user, dispatcher, linkUri, null);
       });
    }
@@ -321,17 +326,29 @@ public class PrintDeviceLayoutPropertyService {
     * and every later write on the viewsheet then fails while re-reading it, taking
     * {@code manage_device_layout} down with it since both share {@code setViewsheetInfo}.
     *
-    * <p>This runs on every path rather than only when seeding a brand-new layout, because there
-    * are two other ways to arrive here with a null: a caller passing {@code "units": null}
-    * explicitly (which {@code applyPrintLayoutPatch} would otherwise write straight over the
-    * seeded value), and a pre-existing layout persisted before this guard existed.
+    * <p>Called on <b>every</b> write path in this class -- {@code setPrintLayout} and
+    * {@code manageDeviceLayout} both -- rather than only where a brand-new layout is built,
+    * because there are three ways to reach the write with a null and only one of them involves
+    * creating a layout: an omitted {@code "units"} key on a new layout; a caller passing
+    * {@code "units": null} explicitly, which {@code applyPrintLayoutPatch} would otherwise write
+    * straight over any default; and a layout persisted before this guard existed, which
+    * {@code manageDeviceLayout} would otherwise carry into the write untouched while never
+    * looking at the print layout at all.
     *
     * <p>{@code "inches"} matches what {@code WizPrintLayoutBuilder} already hardcodes for the
     * same purpose. An unrecognised non-null value is left alone: {@code getPLayoutSize} falls
     * through its {@code default} for those, which is a wrong scale but not a crash, and silently
     * rewriting a caller's explicit value would be its own surprise.
     */
-   private static void requireUnits(VSPrintLayoutDialogModel printLayout) {
+   private static void requireUsableUnits(ScreensPaneModel screensPane) {
+      VSPrintLayoutDialogModel printLayout = screensPane.getPrintLayout();
+
+      if(printLayout == null) {
+         // No print layout at all is the safe state, not a broken one: getPrintPageSize has
+         // nothing to compute from and never reaches the switch.
+         return;
+      }
+
       String units = printLayout.getUnits();
 
       if(units == null || units.isBlank()) {
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/PrintDeviceLayoutPropertyService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/PrintDeviceLayoutPropertyService.java
@@ -104,27 +104,17 @@ public class PrintDeviceLayoutPropertyService {
          VSPrintLayoutDialogModel printLayout = screensPane.getPrintLayout();
 
          if(printLayout == null) {
-            // No print layout configured on this viewsheet yet: seed the fields whose unset
-            // values are not merely wrong but fatal, BEFORE applying the patch.
-            //
-            // scaleFont: an omitted "scaleFont" key must land on 1.0f rather than the bare-field
-            // 0.0f a freshly-constructed model would otherwise carry through to the write
-            // untouched (the Hazard-3 regression this class exists to close).
-            //
-            // units: an omitted "units" key leaves the field null, and that null reaches
-            // ViewsheetPropertyDialogService's printInfo.setUnit(...) and then
-            // VSLayoutService.getPLayoutSize's switch(unit) -- which accepts only "inches" and
-            // "mm" -- where it throws NPE. The throw lands AFTER the patch has already mutated
-            // the live model, so the failed call persists a half-written layout, and every later
-            // write on this viewsheet then fails while re-reading it (taking manage_device_layout
-            // down with it, since both share setViewsheetInfo). "inches" matches what
-            // WizPrintLayoutBuilder already hardcodes for the same purpose.
+            // No print layout configured on this viewsheet yet: seed scaleFont at the safe
+            // default BEFORE applying the patch, so an omitted "scaleFont" key lands on 1.0f
+            // rather than the bare-field 0.0f a freshly-constructed model would otherwise carry
+            // through to the write untouched (the Hazard-3 regression this class exists to
+            // close).
             printLayout = new VSPrintLayoutDialogModel();
             printLayout.setScaleFont(1.0f);
-            printLayout.setUnits("inches");
          }
 
          applyPrintLayoutPatch(printLayout, resolvedPatch);
+         requireUnits(printLayout);
          screensPane.setPrintLayout(printLayout);
          dialogService.setViewsheetInfo(runtimeId, model, user, dispatcher, linkUri, null);
       });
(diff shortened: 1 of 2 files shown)


lizhepu:
**Verdict: 0 blocker, 3 important.** The one-line fix (`setUnits("inches")`) is directionally correct, but the patch pipeline it's added to still lets the exact NPE it claims to close through two other doors, and the fix's own established test convention wasn't followed.

## Important

**1. Explicit `"units": null` (or any unrecognized value) in the patch still reproduces the NPE — the fix is trivially bypassable.** `PrintDeviceLayoutPropertyService.java:312-314` (`applyPrintLayoutPatch`, pre-existing, not touched by this diff):
```java
case "units":
   printLayout.setUnits(asString(value));
```
`asString` (line 322-324) is `value == null ? null : String.valueOf(value)` — i.e. it passes `null` straight through. So a `set_print_layout` call on a brand-new layout with `{"units": null}` (as opposed to omitting the key) overwrites the just-seeded `"inches"` back to `null`, and the same live-model mutation → `VSLayoutService.getPLayoutSize`'s `switch(unit)` NPE happens exactly as before this fix. Contrast with `scaleFont`: `toFloat(null)` calls `Float.parseFloat("null")`, which throws immediately (fails loud) — `units` is the one field in this switch that silently accepts `null`. Any other unrecognized string ("Inches", "in", "millimeters", a typo) is also accepted silently and falls through `getPLayoutSize`'s `default: break;` with no unit conversion applied, silently producing wrong page geometry rather than erroring. Given this repo's own tool-misuse-is-a-plugin-gap principle, `units` should be validated/normalized (case-insensitive alias acceptance, loud rejection of garbage) here, not passed through raw.

**2. Fix only covers the `printLayout == null` branch; a pre-existing/legacy layout with a null `units` field is not seeded and can still crash.** `PrintDeviceLayoutPropertyService.java:106-125`. `ViewsheetPropertyDialogService.getViewsheetInfo` (line 223) unconditionally does `vsLayoutDialog.setUnits(printLayout.getPrintInfo().getUnit())`, which is `null` whenever the underlying `PrintInfo.unit` was never explicitly set (any `PrintInfo`/`PrintLayout` built without a `setUnit()` call — e.g. `PrintInfo`'s own bare-field default, or XML predating an explicit unit tag: `PrintInfo.writeXML` line 218 string-concatenates `unit` into the XML, so a `null` at write time even round-trips back as the *string* `"null"`, and a layout that legitimately never had a `<unit>` element parses back as an actual `null`). In that state `screensPane.getPrintLayout()` is non-null, so the `if(printLayout == null)` seed branch is skipped entirely; if the caller's patch doesn't mention `"units"`, it stays null and hits the identical `switch(unit)` NPE in `VSLayoutService.getPLayoutSize` this PR is meant to close. The fix is incomplete for this case.

**3. No regression test added for the units seed, despite this class's own convention of pinning each named hazard.** `PrintDeviceLayoutPropertyServiceTest.omittedScaleFontOnANewPrintLayoutDefaultsToOneNotJavasZero` (lines 68-86) exercises exactly the `printLayout == null` branch this diff touches, but only asserts `getScaleFont()`/`getPaperSize()` — not `getUnits()`. The test file's docstring explicitly frames itself as pinning "named regressions this file exists to pin down" (currently listing two: Hazard 3 / Risk 2); the new production comment describes "units" as a symmetric third hazard, but no test (new or extended) was added to pin it, so a future refactor could drop the seed silently.

## What I verified and found correct
- `"inches"` is the right tag for the seeded values: `PaperSize.getSize(...)`'s own javadoc says it returns "paper size in inches", and `WizPrintLayoutBuilder.java:203` already hardcodes `info.setUnit("inches")` for the identical construction pattern.
- `PrintInfo.getUnitRatio` treats `"inches"` the same as any unmatched string (ratio = 1), so the seed changes no numeric ratio computation versus the prior `null` — its only effect is giving `printInfo.setUnit(...)` and the downstream `switch(unit)` a non-null value.
- `screensPane.getPrintLayout() == null` (the whole model) — not the `units` field — remains the "unconfigured" sentinel read by `LayoutReadService.readPrintSettings` (lines 164-169), so seeding `units` doesn't disturb that discriminator.
- The comment's claim "the throw lands after the patch has already mutated the live model" checks out: `ViewsheetPropertyDialogService.setViewsheetInfo` calls `layoutInfo.setPrintLayout(printLayout)` (line 426) before `VSLayoutService.getPrintPageSize(printLayout)` (line 427), which is what would NPE.
- `VSLayoutService.getPLayoutSize`'s switch (lines 956-971) does have a `default: break` — the NPE comes from `switch(unit)` itself throwing on a null subject, not an explicit throw in an unmatched case; the comment's phrasing is loose here but not factually wrong.

I have not approved this PR — three important findings stand pending the lead's response.

lizhepu:
All three addressed in `358f6d903`. Findings 1 and 2 were the same gap, so they got one fix rather than two.

**1 + 2 — seeding was the wrong mechanism.** Replaced with `requireUnits(printLayout)`, called *after* the patch and before the write. Seeding only the `printLayout == null` branch left both paths you named still reachable: an explicit `units: null` wrote over the seeded value, and a pre-existing layout skipped the branch entirely. One guarantee now covers all three.

**On unrecognised non-null values** — deliberately left alone, and now said so in the javadoc. `getPLayoutSize` falls through its `default` for those: a wrong scale, not a crash. Silently rewriting a value the caller passed explicitly would be its own surprise, and it would make `units: "pt"` behave differently from every other unknown-but-accepted field.

**3 — four tests**, in the class's existing `Harness`/`ArgumentCaptor` shape. **Proven RED before GREEN**: with `requireUnits` commented out, three fail with `expected: <inches> but was: <null>` — the omitted case, your explicit-null case, and your pre-existing-layout case. The fourth (`anExplicitUnitsValueIsLeftAlone`) stays green without the guard, which is correct — it asserts the guard does *not* overrule a caller. 14/14 with the guard restored.

Your two findings were both real and neither was in the original diagnosis. The `PrintInfo.writeXML` null→`"null"` detail in finding 2 is not addressed here — it makes the broken state round-trippable, which is worth its own look, but it is upstream of this change and fixing it would widen the diff past what this PR is for.

lizhepu:
**Verdict: not clean — one blocker remains. `requireUnits` does not run on every path to `setViewsheetInfo`; `manageDeviceLayout` still ships a pre-existing null-`units` print layout untouched.**

## Blocker

**`manageDeviceLayout` never calls `requireUnits`, so it still crashes on a pre-existing null-`units` layout — the exact hazard round 1 named and the commit message claims is now closed on "every path".** (confidence 92)
`core/src/main/java/inetsoft/web/wiz/viewsheet/PrintDeviceLayoutPropertyService.java:196-255`

`requireUnits` is wired into `setPrintLayout` only (line 117, right after `applyPrintLayoutPatch`). `manageDeviceLayout` (lines 132-256) never touches `screensPane.getPrintLayout()` and never calls `requireUnits`; it loads `model` via the same `dialogService.getViewsheetInfo(runtimeId, user)` (line 197) and passes that same, unmodified `model` straight to `dialogService.setViewsheetInfo(...)` (line 254).

Traced the actual crash: `ViewsheetPropertyDialogService.getViewsheetInfo` (lines 201-224) reads `printLayout.getPrintInfo().getUnit()` straight through into the dialog model with no defaulting (`vsLayoutDialog.setUnits(printLayout.getPrintInfo().getUnit())`, line 223) — so a viewsheet whose persisted `PrintLayout.printInfo.unit` is null (created before this guard existed, or by any other writer) comes back from `getViewsheetInfo` with `screensPane.getPrintLayout().getUnits() == null`. Then in `setViewsheetInfo` (same file, lines 377-432): line 415 copies that null into `printInfo.setUnit(...)`, and line 427 unconditionally calls `VSLayoutService.getPrintPageSize(printLayout)` → `getPLayoutSize(..., unit)` → `switch(unit)` on a null String → NPE. This branch runs whenever `screensPane.getPrintLayout() != null`, regardless of which caller (`setPrintLayout` or `manageDeviceLayout`) invoked `setViewsheetInfo`.

So: viewsheet already has a broken print layout (pre-dates this fix, or somehow got persisted with a null unit) → any `manage_device_layout` call (create/update/delete a device layout, nothing to do with print layout) still throws the same NPE the whole PR exists to fix. None of the 6 `manageDeviceLayout` tests set up a screensPane with an existing printLayout at all (checked all of them, lines 234-341), so this gap is untested and unnoticed.

Fix: call `requireUnits` (guarded for null `printLayout`) in `manageDeviceLayout` too, right before `setViewsheetInfo` — or better, hoist it into one shared pre-write step both entry points go through, since the class javadoc itself already frames this as one hazard shared by both methods.

## Verified correct

- **Ordering change (patch → requireUnits, vs. old seed-before-patch) is safe**: no code reads `printLayout.getUnits()` between `applyPrintLayoutPatch` and `requireUnits` — confirmed by reading lines 116-119, nothing in between.
- **"Unrecognised non-null unit is left alone" is defensible, not a latent crash**: traced `PrintInfo.getUnitRatio` (PrintInfo.java:168-179) — unrecognized units fall through to `ratio = 1`, no NPE, no zero/negative dimension. `VSLayoutService.getPLayoutSize`'s `default` branch (VSLayoutService.java:966-967) returns `asize` unchanged. So an odd non-null unit really is "wrong scale," never a crash, matching the javadoc's claim.
- **Javadoc's technical claims check out**: `printInfo.setUnit(...)` at `ViewsheetPropertyDialogService.java:415`, `VSLayoutService.getPLayoutSize`'s `switch(unit)` recognizing only `"inches"`/`"mm"` at `VSLayoutService.java:959-968`, and the NPE landing after `layoutInfo.setPrintLayout(printLayout)` already mutated the live model (line 426, before the crashing line 427) — all verified against source.
- **Tests assert against the shipped model**: `writtenPrintLayout(h)` captures the argument to `dialogService.setViewsheetInfo` (the real write), not an intermediate — correct pattern.
- **Ran the suite**: 14/14 green on `358f6d903` as claimed. Disabled `requireUnits` and reran: `omittedUnitsOnANewPrintLayoutDefaultsToInchesNotNull`, `anExplicitNullUnitsCannotOverwriteTheDefault`, and `anExistingPrintLayoutWithNoUnitsIsRepairedRatherThanSkipped` all failed (`expected: <inches> but was: <null>`) — 3/4 red, matching the author's report.
- **Minor (not blocking)**: `anExplicitUnitsValueIsLeftAlone` is vacuous as a guard test — confirmed it stays green with `requireUnits` disabled, since `applyPrintLayoutPatch` alone already writes the explicit `"mm"` value; it doesn't exercise the guard at all, only `applyPrintLayoutPatch`'s passthrough. Fine to keep (it does pin "guard must not clobber an explicit value"), but its own comment overstates what it's testing.

Given the blocker above, I have not attempted approval. Since I am the same identity as the author on this run, a self-approval would be refused by GitHub regardless — flagging that plainly so the lead can report it rather than have me attempt and fail silently.

lizhepu:
Blocker confirmed and fixed in `91729ee5f`. You were right and the previous commit message was wrong — it claimed the guard ran on "every path" when `manageDeviceLayout` never called it at all.

**The fix takes the pane, not the layout.** `requireUsableUnits(ScreensPaneModel)` now runs on both write paths, and returns early when there is no print layout (that state is safe — `getPrintPageSize` has nothing to compute from). Taking the pane matters beyond this bug: a future write path in this class gets the guard by reaching for the same helper, rather than by someone remembering a rule.

**New test, proven RED and proven narrow.** `manageDeviceLayoutRepairsAPersistedNullUnitsRatherThanCrashingOnIt`. Disabling *only* the device-path guard fails **that one test and nothing else** — so it pins that path specifically rather than re-covering what the other tests already assert. 15/15 with both guards in place.

**On `anExplicitUnitsValueIsLeftAlone`** — you were right that the wording was misleading. Its comment now says plainly that it pins `applyPrintLayoutPatch`'s passthrough, not the guard, and that staying green with the guard disabled is the point rather than a gap.

Two rounds, two real findings, and both were things the original diagnosis missed rather than restatements of it — the second one specifically contradicted a claim I had put in a commit message.

fredaan (approved):

PR #4835, opened 2026-08-28
feat(odata): expose getOutputColumns() for tabular metadata (S1)

Nelson-cao:
## Summary
- `ODataQuery.getOutputColumns()` now returns the entity's scalar columns (Name + coarse EDM-mapped type) parsed from the schema already fetched by `loadSchema()`'s pattern, following `SharepointOnlineQuery`'s cache-then-fetch approach. `NavigationProperty` elements are never included.
- Fixes a pre-existing NPE in `ODataRuntime.getSchemaNode()`: it dereferenced `getMetaDataDocument()`'s result without a null check, which throws when the OData service is unreachable or returns malformed metadata — caught by a new reverse-assertion test.
- Adds a regression test locking `WizDatabaseController.classifyQueryClass(ODataQuery.class) == "METADATA"`.

## Test plan
- [x] `./mvnw -pl connectors/inetsoft-odata -am -Dtest=ODataRuntimeTests,ODataDatasourceAnnotationClassTest -Dsurefire.failIfNoSpecifiedTests=false test` — 7/7 pass (3 pre-existing + 3 new in `ODataRuntimeTests`, 1 new in `ODataDatasourceAnnotationClassTest`)
- [x] Verified no throw when the metadata service returns 500 or malformed XML
- [x] Verified `getOutputColumns()` issues exactly one `$metadata` request (no extra HTTP call beyond the existing schema-fetch pattern)

### DIFF
--- a/connectors/inetsoft-odata/src/main/java/inetsoft/uql/odata/ODataQuery.java
+++ b/connectors/inetsoft-odata/src/main/java/inetsoft/uql/odata/ODataQuery.java
@@ -392,6 +392,13 @@ public class ODataQuery extends TabularQuery {
 
    private void loadSchema() {
       Node schema = ODataRuntime.getSchemaNode((ODataDataSource) getDataSource());
+
+      if(schema == null) {
+         nameSpace = null;
+         functions = new ArrayList<>();
+         return;
+      }
+
       nameSpace = Tool.getAttribute((Element) schema, "Namespace");
       String entityType = ODataRuntime.getEntityType(schema, entity);
       entityType = entityType == null ? "" : entityType;
--- a/connectors/inetsoft-odata/src/main/java/inetsoft/uql/odata/ODataQuery.java
+++ b/connectors/inetsoft-odata/src/main/java/inetsoft/uql/odata/ODataQuery.java
@@ -17,6 +17,7 @@
  */
 package inetsoft.uql.odata;
 
+import inetsoft.uql.schema.XTypeNode;
 import inetsoft.uql.tabular.*;
 import inetsoft.util.Tool;
 import org.w3c.dom.*;
@@ -362,6 +363,31 @@ public class ODataQuery extends TabularQuery {
       }
 
       return nameSpace;
+   }
+
+   @Override
+   public XTypeNode[] getOutputColumns() {
+      XTypeNode[] columns = super.getOutputColumns();
+
+      if((columns == null || columns.length == 0) && getDataSource() != null &&
+         entity != null && !entity.isEmpty())
+      {
+         columns = fetchColumns();
+
+         if(columns.length > 0) {
+            setOutputColumns(columns);
+         }
+      }
+
+      return columns;
+   }
+
+   private XTypeNode[] fetchColumns() {
+      ODataDataSource ds = (ODataDataSource) getDataSource();
+      Node schema = ODataRuntime.getSchemaNode(ds);
+      String entityType = ODataRuntime.getEntityType(schema, entity);
+      entityType = entityType == null ? "" : entityType;
+      return ODataRuntime.getProperties(schema, entityType);
    }
 
    private void loadSchema() {
--- a/connectors/inetsoft-odata/src/main/java/inetsoft/uql/odata/ODataRuntime.java
+++ b/connectors/inetsoft-odata/src/main/java/inetsoft/uql/odata/ODataRuntime.java
@@ -19,6 +19,7 @@ package inetsoft.uql.odata;
 
 import inetsoft.uql.VariableTable;
 import inetsoft.uql.XTableNode;
+import inetsoft.uql.schema.*;
 import inetsoft.uql.tabular.*;
 import inetsoft.uql.tabular.oauth.AuthorizationClient;
 import inetsoft.uql.tabular.oauth.Tokens;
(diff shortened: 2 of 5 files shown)


Nelson-cao:
## Review round 1 (P5)

Reviewed against `docs/teams/2026-08-28-odata-tabular-s1/00-charter.md` (A1/A2/A3 + all three counter-assertions), reading the real diff (`7245c2d41` + `123e67918`) directly rather than the build/verify logs' excerpts, plus the unchanged surrounding code each new method touches (`getEntityType`, `getFunctions`, `getMetaDataDocument`, `loadSchema`, `CoreTool.getChildNodesByTagName`/`getAttribute`, `classifyQueryClass`).

**A1/A2/A3 and all three counter-assertions: satisfied.** Verified independently, not just re-read from the verify doc — e.g. confirmed `NavigationProperty` exclusion is structural (distinct tag name, non-recursive child lookup) not conditional logic, confirmed `classifyQueryClass` falls through to `METADATA` by reading the real method body, confirmed `loadSchema()` and its two callers are byte-for-byte unchanged.

**No blockers.**

**Important (non-blocking, pre-existing, not a regression from this PR):** the `getSchemaNode()` NPE fix is null-safe for the new `getOutputColumns()`/`getProperties()` path (verified: both tolerate a null schema node and early-return), but `ODataQuery.loadSchema()` — unchanged, reached from `setFunction()`/`getFunctions()` (the composer's function dropdown) — still calls `Tool.getAttribute((Element) schema, "Namespace")` with no null check, and `CoreTool.getAttribute` is not null-safe. So the identical unreachable/malformed-service trigger still NPEs, just one frame later than before. This isn't new (pre-PR, `getSchemaNode()` itself threw at the same trigger), so it's not a regression and it's outside the charter's actual scope (the reverse assertion is written specifically about `getOutputColumns()`). Flagging because the commit message's "Also fixes a pre-existing NPE in `getSchemaNode()`" could read as closing the general failure mode when it only covers the one caller written to expect it. Suggest either a quick follow-up guarding `loadSchema()` the same way, or just a note in the PR description that the fix's scope is `getOutputColumns()` only.

**Nits (for the charter's follow-ups list, not blocking):**
- `toXTypeNode()`'s EDM→XTypeNode switch has 4 of 8 arms (Boolean/Date/DateTimeOffset/TimeOfDay) with no test coverage — already flagged by P4 verify and partially closed by the follow-up commit's 3 new `assertInstanceOf` checks. Read all 4 remaining arms by hand; each is correctly wired, so this is a coverage gap, not a bug I could trigger.
- `ODataDatasourceAnnotationClassTest` reaches `core`'s package-private `classifyQueryClass` via same-package placement across modules, which only works because this reactor has no `module-info.java`. Design already discloses this tradeoff explicitly; noting it here since this PR sets the precedent other connectors will likely copy.

I actively tried to break A1/A2/A3 and the three counter-assertions (EDM-mapping fallback, last-`.` namespace split, the `entity != null && !entity.isEmpty()` guard, NavigationProperty exclusion, test-package placement) and could not, beyond the one item above. Full writeup in `docs/teams/2026-08-28-odata-tabular-s1/06-review-r1.md`.

A4/A5 (live MCP walkthrough): still blocked in this environment too (no `composer-chat`/`viz-chat` MCP tool connection) — unchanged from P4, not a new finding.

Nelson-cao:
Fixed the review's important finding: `loadSchema()` (`ODataQuery.java`) NPE'd at `Tool.getAttribute((Element) null, "Namespace")` when `getSchemaNode()` returns null for an unreachable/malformed OData service — the same trigger this PR already made `getOutputColumns()` tolerate, just one call frame earlier. Added a null-check that degrades `functions` to an empty list (matching `ODataRuntime.getFunctions()`'s own never-null convention) and leaves `nameSpace` null (consistent with `getNameSpace()`'s existing lazy-reload-on-null pattern).

Added `getFunctionsDoesNotThrowWhenMetadataRequestFails` mirroring the existing `getOutputColumns` reverse-assertion tests. Full targeted suite (`ODataRuntimeTests`, `ODataDatasourceAnnotationClassTest`) green: 8/8.

Commit `cd46479fa`.

jshobe-inetsoft (approved):
## Summary

This PR (targeting `stylebi-wiz-main-rebase`) adds `ODataQuery.getOutputColumns()` (fetch-then-cache scalar column metadata parsed from the OData `$metadata` document, following the existing `SharepointOnlineQuery` pattern), fixes a pre-existing NPE in `ODataRuntime.getSchemaNode()` when the metadata service is unreachable or returns malformed XML, teaches `ODataQuery.loadSchema()` to tolerate the now-nullable schema, and adds a regression test locking `WizDatabaseController.classifyQueryClass(ODataQuery.class) == "METADATA"`.

I read the full `ODataQuery.java`, `ODataRuntime.java`, `TabularQuery.java` (base class), `SharepointOnlineQuery.java` (the pattern being mirrored), and `CoreTool.getAttribute`/`getChildNodesByTagName` to verify null-handling, in addition to the diff itself.

## Correctness / Completeness

- The new `getOutputColumns()` override correctly mirrors `SharepointOnlineQuery`'s cache-then-fetch pattern: it only re-fetches when `super.getOutputColumns()` is null/empty, only caches via `setOutputColumns()` when the fetch actually produced columns (so a transient failure isn't permanently cached as "no columns" — it'll simply retry on the next call), and only fetches when an `entity` is actually selected. This is correct and consistent with existing conventions.
- `ODataRuntime.getProperties()` correctly filters `NavigationProperty` out by only matching `Property`-tagged children (`Tool.getChildNodesByTagName` does an exact tag-name match), and is null-safe on both `schemaNode` and `entityType`.
- `getSchemaNode()`'s fix is correct and matches the pattern used elsewhere in the same file (`isOdataV4()` already null-checks `getMetaDataDocument()`'s result the same way).
- `loadSchema()`'s new guard correctly degrades to an empty `functions` list and `nameSpace = null` rather than propagating the NPE, and this is verified end-to-end by `getFunctionsDoesNotThrowWhenMetadataRequestFails`.
- **Gap: `getNameSpace()` (ODataQuery.java, unmodified by this PR) has the identical latent NPE and was not patched.** It calls `ODataRuntime.getSchemaNode(...)` directly and does `Tool.getAttribute((Element) schema, "Namespace")` without a null check; `CoreTool.getAttribute` does not null-guard its `Element` argument (unlike `getChildNodesByTagName`, which does), so a null schema still NPEs here. This is reachable independently of `loadSchema()`: a persisted `ODataQuery` with a previously-selected bound function is restored via `parseContents()`/`ODataFunction.parseXML()` directly (bypassing `setFunction()`/`loadSchema()` entirely), so `functionSelected()` can be `true` on a freshly-deserialized query whose `nameSpace` field was never populated. If the query is then executed while the metadata endpoint is unreachable, `ODataRuntime.getFunctionUriSegment()` calls `query.getNameSpace()`, which will NPE. Given the PR's explicit goal is hardening this exact "unreachable/malformed metadata" scenario, this call site should probably get the same treatment (e.g., default to `""` when the schema is null, matching the `entityType == null ? "" : entityType` convention already used elsewhere in this file).
- `getProperties()` only reads `Property` children declared directly on the matched `<EntityType>` element; an EntityType declared with a `BaseType` (OData type inheritance) would silently omit inherited properties. Likely acceptable for the stated "S1" scope, but worth a follow-up note since it could look like missing columns for some real-world schemas.

## Clean Code / Readability

- `fetchColumns()` in `ODataQuery` is a near-exact duplicate of the pattern in `SharepointOnlineQuery.fetchColumns()`. That's fine as intentional pattern-following per the PR description, not something to refactor now.
- `toXTypeNode()`'s coarse EDM→XTypeNode mapping (e.g. all integer widths → `LongType`, `Decimal`/`Double`/`Single` → `DoubleType`, everything else including `Guid`/`Binary`/`Duration` → `StringType`) is a reasonable, well-contained simplification for exposing column type hints, consistent with the "coarse type" framing in the PR title.
- The test helper `readXml(String file)` is just a same-body alias for the pre-existing `readJson(String file)` helper (which is itself a generic "read resource as string" utility despite its name). It reads better at the call site (`readXml("getOutputColumns.metadata.xml")`), but it's an odd bit of indirection — minor, test-only nit.

## Error Handling / Edge Cases

- All three new/modified failure paths (metadata unreachable via HTTP 500, malformed XML, and no-entity-selected) are exercised by tests and correctly avoid throwing.
- `fetchColumns()`'s guard (`entity != null && !entity.isEmpty()`) correctly avoids calling into `getEntityType`/`getProperties` with a null entity.

## Security

No new external input handling or attack surface introduced — XML parsing continues to go through the existing `Tool.parseXML`/`getMetaDataDocument()` path, which is unchanged. No credentials or secrets are touched by this diff.

## Regression Risk / Backward Compatibility

- Changes are additive (new `@Override` method, new private helpers, a defensive null-check) and confined to `connectors/inetsoft-odata`. No XML serialization format changes, no signature changes to public/protected methods used elsewhere. Low regression risk to other connectors or the core `TabularQuery` contract.
- The one behavioral change to watch: `getSchemaNode()` now returns `null` instead of throwing on failure. Every other current caller (`loadSchema()`, the new `fetchColumns()`/`getProperties()`) has been audited and correctly handles `null`, except `getNameSpace()` as noted above.

## Tests

- New tests compile and pass in CI (build check is green): `ODataRuntimeTests` (7/7) and the new `ODataDatasourceAnnotationClassTest` (1/1). I independently verified `WizDatabaseController.classifyQueryClass(ODataQuery.class)` does resolve to `"METADATA"` against the actual `stylebi-wiz-main-rebase` base (`ODataQuery` is not a `ScriptedQuery`, ships no `endpoints.json`, isn't a REST query, and isn't `SelectableTabularQuery`), so the new regression-lock assertion is accurate, not just coincidentally passing.
- No existing tests appear broken by this change.

## Check / Merge Status

- `gh pr checks 4835`: `build`, `Semgrep OSS`, `Semgrep Scan`, and `license/cla` all pass. Only `claude-review` fails, which is excluded from this gate per review policy (this manual review supersedes it).
- `gh pr view 4835 --json mergeable,mergeStateStatus`: `mergeable: MERGEABLE`, no conflicts against the PR's actual base branch (`stylebi-wiz-main-rebase`).

## Verdict

No blocking issues. The one gap identified (`getNameSpace()` not receiving the same null-schema guard as `loadSchema()`) is a real latent bug but narrow in practical reach and pre-existing in spirit (same class of NPE the PR is otherwise fixing) — recommended as a fast follow-up rather than a blocker for this PR. Approving.

PR #4836, opened 2026-08-28
fix(security): add wiz-prefixed identity directory endpoints (bug 76118)

Zxm-happy:
## Summary
- Bug 76118: wiz's Admin > Entitlements member picker always shows "No directory available", because `wiz-services` calls the generic, session-oriented `/api/em/security/get-current-authentication-provider` and `/api/em/security/providers/{provider}/identities/{type}` endpoints server-to-server with a bearer JWT and no cookies. `AbstractSecurityFilter.isWizRequest()` only recognizes `/api/wiz/**` or the `wiz_auth` cookie, so `WizServiceAuthenticationFilter` never validates the JWT, no principal is constructed, and the request falls through as an unauthenticated browser request — producing an empty identity list with no error anywhere.
- Adds `WizSecurityDirectoryController` with two new GET endpoints under `/api/wiz/v1/security/**`, mirroring the `AdminAiController`/`AdminAiEndpointPrefixTest` precedent that fixed the identical defect shape for the admin-chat endpoints. Both delegate to the exact same `AuthenticationProviderService` methods the original `/api/em/security/**` controllers already call, so behavior is identical once a principal exists.
- Deliberately does **not** widen `isWizRequest`/`wizApiPath` to cover `/api/em/security/**` — that prefix also carries several mutating, `@Secured`-gated provider-management endpoints (add/edit/remove/reorder provider), so broadening authentication there would newly make those bearer-JWT-authenticatable. New, narrowly-scoped endpoints avoid that risk.
- Both new endpoints go through `AdminAiCallerGuard.requireBearerAuthenticatedRequest()` for consistency with the rest of the CSRF-exempt `/api/wiz/**` prefix (reasoning for applying it to read-only GETs is in the wiz-side PR / team docs).
- Leaves `AuthenticationProviderController` and `SecurityTreeController`'s existing `/api/em/security/**` mappings untouched — the browser-session EM UI still uses those.

## Test plan
- [x] `WizSecurityDirectoryEndpointPrefixTest` — pins that the two new paths are recognized by the real `isWizRequest()` and that the two original `/api/em/security/**` paths are not (characterizing the bug).
- [x] `WizSecurityDirectoryControllerTest` — org-guard, all four identity-type routes, `getCurrentProvider` delegation, and the bearer-token gate.
- [x] `mvn -o -pl core -am test -Dtest=WizSecurityDirectoryEndpointPrefixTest,WizSecurityDirectoryControllerTest` — 8/8 passed.
- [ ] Companion wiz-services PR updates `identityDirectory.ts` to call these new paths — this PR alone does not fix the picker end-to-end until that PR also merges and deploys.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/security/WizSecurityDirectoryController.java
+++ b/core/src/main/java/inetsoft/web/wiz/security/WizSecurityDirectoryController.java
@@ -1,0 +1,112 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.web.wiz.security;
+
+import inetsoft.sree.security.*;
+import inetsoft.uql.util.Identity;
+import inetsoft.util.Catalog;
+import inetsoft.util.InvalidOrgException;
+import inetsoft.web.admin.ai.AdminAiCallerGuard;
+import inetsoft.web.admin.security.AuthenticationProviderService;
+import inetsoft.web.admin.security.SecurityProviderStatus;
+import inetsoft.web.admin.security.user.GetIdentityNameResponse;
+import inetsoft.web.factory.DecodePathVariable;
+import inetsoft.web.security.RequiredPermission;
+import inetsoft.web.security.Secured;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.*;
+
+import java.security.Principal;
+
+/**
+ * wiz-broker-facing mirror of {@link inetsoft.web.admin.security.AuthenticationProviderController#getCurrentProvider}
+ * and {@link inetsoft.web.admin.security.user.SecurityTreeController#getIdentities}, mapped under
+ * {@code /api/wiz/**} so {@code WizServiceAuthenticationFilter} actually validates the caller's
+ * bearer JWT (see {@code AbstractSecurityFilter#isWizRequest}, which only recognizes the
+ * {@code wiz_auth} cookie or this prefix).
+ *
+ * <p>The original endpoints
+ * ({@code /api/em/security/get-current-authentication-provider},
+ * {@code /api/em/security/providers/{provider}/identities/{type}}) are the browser-session EM SPA
+ * endpoints and are left untouched; wiz-services' cookie-less, server-to-server call needs its own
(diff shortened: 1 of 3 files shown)


Zxm-happy (commented):
Reviewed against `main`'s `AuthenticationProviderController.getCurrentProvider` and `SecurityTreeController.getIdentities`, and against the `AdminAiController`/`AdminAiCallerGuard`/`AdminAiEndpointPrefixTest` precedent this PR mirrors. No blocking issues found — verdict: approve on merits (posted as a comment, not an approval, since GitHub rejects self-approval for this account).

**Verified independently (not just re-reading the team docs):**
- `getIdentities`'s org-guard + `Identity.USER/GROUP/ROLE/ORGANIZATION` routing body, and the `@Secured(EM_COMPONENT, "settings/security/users", ACCESS)` annotation, are byte-for-byte identical to `SecurityTreeController.getIdentities` (diffed side-by-side). `getCurrentProvider` is a one-line delegation matching `AuthenticationProviderController.getCurrentProvider` exactly, including the deliberate absence of `@Secured` on the "whoami" lookup.
- `AdminAiCallerGuard.requireBearerAuthenticatedRequest()` wiring matches the existing `AdminAiController` pattern (guard called first in the method body, `@Secured` as the AOP-level gate). Applying it to read-only GETs is a defensible, low-cost defense-in-depth call given the guard's own javadoc already frames it as a CSRF/same-origin backstop, not just a mutation guard — agree with the fixer's reasoning here.
- No route collisions: `/api/wiz/v1/security/**` is not mapped anywhere else in the codebase.
- Package placement (`inetsoft.web.wiz.security`, next to `WizServiceAuthenticationFilter`) is a reasonable, well-justified deviation from the admin-chat `inetsoft.web.admin.ai` layout — this is an identity-lookup endpoint, not an admin-chat one, and `web.wiz.*` is already where other non-admin-chat wiz-broker endpoints live.
- `WizSecurityDirectoryEndpointPrefixTest` exercises the real, non-mocked `CSRFFilter.isWizRequest()` against the four relevant paths (same technique as `AdminAiEndpointPrefixTest`) — this is a genuine executable proof, not just an assertion of intent.
- `WizSecurityDirectoryControllerTest` correctly proves delegation, org-guard, USER/GROUP routing, and that the bearer-token gate short-circuits before the service is touched (`verifyNoInteractions(service)`).

**Non-blocking nits (not requiring changes to merge):**
- `WizSecurityDirectoryControllerTest`'s class javadoc says it "return[s] real data for a wiz-flagged principal" — the `Principal` in the test is a bare mock, never asserted as wiz-flagged, so the claim overstates what's actually proven (delegation/routing/gate correctness with a mocked service layer). Worth tightening next time this file is touched.
- Only `Identity.USER`/`GROUP` branches are covered by the new controller test; `ROLE`/`ORGANIZATION` are copied-for-parity dead code from the caller's perspective (`identityDirectory.ts` only ever passes user/group) and untested here, consistent with how those branches are exercised elsewhere.

**Scope confirmation:** this PR correctly leaves `isWizRequest`/`wizApiPath` and the original `/api/em/security/**` mappings untouched, avoiding the privilege-widening risk flagged during triage (that prefix also carries mutating, `@Secured`-gated provider-management endpoints). The known sibling gap (`hostSecurity.ts` → `/api/em/security/get-enable-security`, same defect class) is correctly out of scope for this PR and should be filed separately.

Not actioning `gh pr merge` here — per this repo's convention, merge decisions are left to a human. Note the `build` check was still in-progress at review time and should be confirmed green before merge; the `claude-review` bot check failure is an infra/tooling error ("Claude encountered an error after 0s"), not a substantive finding.

PR #4840, opened 2026-08-28
Fix DataSourceBrowserService HTTP 500 on malformed em.locale

larryliang-inetsoft:
## Root cause

`Catalog.parseLocale` (`core/src/main/java/inetsoft/util/Catalog.java:263-280`) throws
an uncaught `RuntimeException` for any non-empty locale string that isn't a bare
2-character code or exactly `language_COUNTRY` (underscore at index 2) -- for example
the hyphenated BCP-47/HTTP `Accept-Language` form `"en-US"`, a 3-letter language code
like `"eng_US"`, or a bare word like `"english"`.

`DataSourceBrowserService.getLocale` (`core/src/main/java/inetsoft/web/portal/data/DataSourceBrowserService.java:514-526`)
calls `Catalog.parseLocale(SreeEnv.getProperty("em.locale"))` directly with no
try/catch for any caller whose `Principal` is not an `SRPrincipal`. This method is
called unconditionally at the top of `getDataSources`, `getAllSubDataSources`, and
`getSearchDataSources` (which itself calls `getAllSubDataSources`), so a malformed
`em.locale` property value breaks data-source browsing for that caller.

Both real controller entry points (`DataSourceController`, `WizDatabaseController`)
are covered by a Spring `@ControllerAdvice` catch-all, so this surfaces as a clean
HTTP 500 JSON response rather than an application crash -- but the feature is
genuinely broken for the affected caller, with no fallback to a usable default
locale.

`SRPrincipal` callers are unaffected because `AuthenticationService.updateLocale`
already wraps the same `Catalog.parseLocale` call in try/catch when populating
`SRPrincipal`'s locale field at login, so a malformed source value never reaches the
field `getLocale()` reads.

## Fix

Wrap the `Catalog.parseLocale` call in `DataSourceBrowserService.getLocale` in
try/catch, log a warning, and fall through to the method's existing
`locale == null ? Locale.getDefault() : locale` path -- the same pattern already used
at `LocalizationService.getSuffix`. `Catalog.parseLocale`'s throwing contract is left
unchanged: `AuthenticationService.updateLocale` genuinely depends on it (no
null-check before `setLocale`, so a non-throwing parseLocale would silently blank out
a previously-good stored locale on a malformed re-login).

## Testing

- Added `core/src/test/java/inetsoft/util/CatalogTest.java`, pinning
  `Catalog.parseLocale`'s existing contract (`null`/empty -> `null`, `"en"` and
  `"en_US"` parse, `"en-US"` throws).
- Added `core/src/test/java/inetsoft/web/portal/data/DataSourceBrowserServiceTest.java`,
  mocking `SreeEnv.getProperty("em.locale")` to return `"en-US"` and calling
  `getDataSources`/`getAllSubDataSources` with a non-`SRPrincipal` principal --
  confirms no exception and a sane (default-locale-based) result.

Ran via a single-module offline reactor build:
```
./mvnw -o -pl core -am test -Dtest=CatalogTest,DataSourceBrowserServiceTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false
```
`Tests run: 7, Failures: 0, Errors: 0`. The test log also shows the new warning being
logged exactly as intended (`Invalid em.locale property value: en-US`).

Found via the property-documentation corpus's own defect audit; no Redmine issue
filed for this one.

### DIFF
--- a/core/src/main/java/inetsoft/web/portal/data/DataSourceBrowserService.java
+++ b/core/src/main/java/inetsoft/web/portal/data/DataSourceBrowserService.java
@@ -519,7 +519,14 @@ public class DataSourceBrowserService {
       }
       else {
          String locstr = SreeEnv.getProperty("em.locale");
-         locale = Catalog.parseLocale(locstr);
+
+         try {
+            locale = Catalog.parseLocale(locstr);
+         }
+         catch(Exception e) {
+            LOG.warn("Invalid em.locale property value: {}", locstr, e);
+            locale = null;
+         }
       }
 
       return locale == null ? Locale.getDefault() : locale;
--- a/core/src/test/java/inetsoft/util/CatalogTest.java
+++ b/core/src/test/java/inetsoft/util/CatalogTest.java
@@ -1,0 +1,59 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.util;
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import java.util.Locale;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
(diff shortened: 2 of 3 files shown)


larryliang-inetsoft:
Round 1 review - independent reviewer, uninvolved in this bug's diagnosis/refutation/fix chain. (Posted as a comment, not a review approval: GitHub blocks self-approval since this PR and I share the same account.)

Head commit reviewed: 8773e2960a9c0c12c3088055dde0221abfda03c3.

Checked: diff scope (confirmed only DataSourceBrowserService.java touched in production code, Catalog.parseLocale's throwing contract genuinely untouched), CatalogTest uses only the refutation's corrected throwing example ("en-US"), DataSourceBrowserServiceTest genuinely exercises the fixed path (getLocale is called unconditionally before any registry access, so this test would fail pre-fix and passes post-fix), PR body uses corrected severity language (handled HTTP 500 via @ControllerAdvice, not "crash") and correctly drops the original diagnosis's unsupported LocalizationService.getSuffix claim, and did an adversarial sweep for other call sites/sibling branches/same-defect-elsewhere (all correctly scoped, no scope creep, no overfit smells).

Full review notes: docs/teams/2026-08-28-bugs-property-corpus-highimpact/bug-emlocale-crash/04-review-r1.md in the stylebi-wiz repo.

Verdict: approved, zero findings.

PR #4843, opened 2026-08-28
Withhold server.save.locations from admin-chat property reads (plaintext password exposure)

larryliang-inetsoft:
## Root cause

`SchedulerConfigurationService.setServerLocations` (`core/src/main/java/inetsoft/web/admin/schedule/SchedulerConfigurationService.java:298-354`) stores each server-save location as `path|label|username|password` in a single `server.save.locations` property, joined with `;` for multiple locations, written with a plain `SreeEnv.setProperty` — the password is embedded verbatim in the clear whenever an operator types a username/password directly instead of referencing a stored credential (`useCredential()==false`).

`AdminPropertyCatalog.isSecret("server.save.locations")` (`core/src/main/java/inetsoft/web/admin/ai/AdminPropertyCatalog.java:264-278`) returned `false`: the name matches none of the `password`/`secret`/`credential`/`.key`/`license.*` shape tests, and the property isn't in `ENCRYPTED_CREDENTIALS` (that allow-list is for single values written whole via `SreeEnv.setPassword`) or `CONFIRMED_NOT_SECRET`. It's also absent from `admin-property-catalog.json` entirely — an uncatalogued property, which the class javadoc says stays readable.

`AdminPropertiesController.view()` (line 116, `String stored = secret ? null : SreeEnv.getProperty(name.key(), false, false);`) therefore forwards the raw, password-embedded value to admin-chat's generic property read endpoint (`GET /api/wiz/v1/admin/properties[/{name}]`), gated only by ordinary site-admin auth — no additional privilege boundary for credential material.

**Concrete exposure evidence:** the actual MCP tool that surfaces this endpoint to the model — `plugin/admin/src/tools/readTools.ts` in the `stylebi-wiz` repo (`get_property`/`list_properties`) — calls this same REST path and returns the JSON response verbatim as the tool result. That tool's own description string tells the calling model explicitly that secret properties "can never be READ through admin-chat," naming the seven `ENCRYPTED_CREDENTIALS` entries. `server.save.locations` was exactly the kind of property that promise is supposed to cover, but wasn't — a real off-host egress of a plaintext password to a model provider, not just a hypothetical.

## Fix

Added a new purpose-built `COMPOSITE_SECRET_PROPERTIES` withhold-only set (currently just `server.save.locations`) in `AdminPropertyCatalog.java`, unioned into `isSecret()` only — **not** into `isEncryptedCredential()`. Deliberately not `ENCRYPTED_CREDENTIALS`: that set's write path (`AdminChangeService`) calls `SreeEnv.setPassword(name.key(), <caller's literal string>)`, which would replace the *entire* multi-location composite value with a single encrypted opaque string on any admin-chat write — corrupting every configured location's path/label/username, not encrypting one password field.

This makes `server.save.locations`: `isSecret=true`, `isEncryptedCredential=false` (unchanged). `AdminPropertiesController.view()` now takes the existing generic "secret pattern... neither read nor changed" branch — no new wording needed. As an intended side effect (not scope creep — confirmed during refutation), `AdminChangePlanService`'s `isSecret && !credential` check now also refuses admin-chat writes to this property, closing an equivalent plaintext-write path that is open today.

**Note on existing masking:** a per-field redactor for this exact property *does* already exist — `Util.PLACEHOLDER_PASSWORD` + `ServerPathInfoModel.Builder.from()` — but it's wired to the structured, typed Enterprise Manager schedule-configuration screen (`ScheduleService.getServerLocations`), not the generic name/value-string `AdminPropertyCatalog`/`AdminPropertiesController` path this bug is about. Reaching into that schedule-specific DTO from the generic catalog would couple two unrelated subsystems for one property; blanket withhold via the existing binary `isSecret` gate is the minimal, correctly-scoped fix here, matching how `password.encryption.key`/`jwt.signing.key`/`sso.rsa.private.key` are already treated.

## Testing

Added two tests to `core/src/test/java/inetsoft/web/admin/ai/AdminPropertyCatalogTest.java`, following the exact `#76006` pattern already established in that file:
- `withholdsServerSaveLocationsEvenThoughNeitherTheNamePatternNorEncryptedCredentialCoversIt` — asserts the name does NOT match the copied name-shape predicate and is NOT `isEncryptedCredential`, but `isSecret(...)` IS `true`.
- `withholdsServerSaveLocationsWhateverCasingTheCallerUses` — casing variant (`Server.Save.Locations`).

Ran `./mvnw -o -pl core -am test -Dtest=AdminPropertyCatalogTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false`: `Tests run: 29, Failures: 0, Errors: 0` (27 pre-existing + 2 new), full `core` module build succeeded.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminPropertyCatalog.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminPropertyCatalog.java
@@ -272,7 +272,7 @@ public class AdminPropertyCatalog {
          return false;
       }
 
-      return isEncryptedCredential(lower) ||
+      return isEncryptedCredential(lower) || COMPOSITE_SECRET_PROPERTIES.contains(lower) ||
          lower.contains("password") || lower.contains("secret") ||
          lower.contains("credential") || lower.endsWith(".key") || lower.startsWith("license.");
    }
@@ -392,6 +392,28 @@ public class AdminPropertyCatalog {
     */
    private static final Set<String> CONFIRMED_NOT_SECRET = Set.of(
       "enable.changepassword", "sso.rsa.public.key", "google.maps.key");
+
+   /**
+    * Properties whose stored value is a composite of several fields - some secret, some not -
+    * joined into a single string, so the binary {@link #isSecret}/read gate is the only sound
+    * granularity: there is no per-field redactor in this generic, name/value-string catalog (one
+    * exists for {@code server.save.locations} specifically, in {@code ServerPathInfoModel.Builder},
+    * but it is wired to the structured Enterprise Manager schedule-configuration screen, not this
+    * generic admin-chat property path).
+    *
+    * <p>Deliberately NOT in {@link #ENCRYPTED_CREDENTIALS}: that set's contract is a single value
+    * written whole with {@code SreeEnv.setPassword}, so admin-chat can safely round-trip it through
+    * {@code setPassword}. {@code server.save.locations} is written with a plain
+    * {@code SreeEnv.setProperty} by {@code SchedulerConfigurationService.setServerLocations} as
+    * {@code path|label|username|password} segments joined by {@code ;} - one property, several
+    * server-save locations, each with its own credential. Routing it through
+    * {@code ENCRYPTED_CREDENTIALS} would make admin-chat's write path call
+    * {@code SreeEnv.setPassword("server.save.locations", <arbitrary caller string>)}, replacing
+    * every configured location with a single encrypted opaque value instead of encrypting one
+    * field of one location.
+    */
+   private static final Set<String> COMPOSITE_SECRET_PROPERTIES = Set.of(
+      "server.save.locations");
 
    private static final String RESOURCE = "admin-property-catalog.json";
    private final List<CatalogEntry> entries;
--- a/core/src/test/java/inetsoft/web/admin/ai/AdminPropertyCatalogTest.java
+++ b/core/src/test/java/inetsoft/web/admin/ai/AdminPropertyCatalogTest.java
@@ -276,6 +276,30 @@ class AdminPropertyCatalogTest {
       assertTrue(AdminPropertyCatalog.isSecret("log.fluentd.security.sharedKey"));
    }
 
+   @Test
+   void withholdsServerSaveLocationsEvenThoughNeitherTheNamePatternNorEncryptedCredentialCoversIt() {
+      // SchedulerConfigurationService.setServerLocations embeds a plaintext password as the
+      // fourth pipe-delimited field of a "path|label|username|password" segment when the operator
+      // types credentials directly, and writes the whole "server.save.locations" property with a
+      // plain SreeEnv.setProperty. The name matches none of the five isSecret substrings/suffixes
+      // and is not a single encrypted value, so it belongs in neither the name test nor
+      // ENCRYPTED_CREDENTIALS - only the dedicated COMPOSITE_SECRET_PROPERTIES withhold list.
+      assertFalse(matchesTheNamePattern("server.save.locations"),
+                  "the point of this test is that the NAME does not match - if it now does, the "
+                  + "union below is no longer what covers it");
+      assertFalse(AdminPropertyCatalog.isEncryptedCredential("server.save.locations"),
+                  "server.save.locations is a multi-location composite value, not a single "
+                  + "credential admin-chat can round-trip through SreeEnv.setPassword");
+      assertTrue(AdminPropertyCatalog.isSecret("server.save.locations"),
+                 "server.save.locations can embed a plaintext password and must not be read back "
+                 + "through admin-chat");
+   }
+
+   @Test
+   void withholdsServerSaveLocationsWhateverCasingTheCallerUses() {
+      assertTrue(AdminPropertyCatalog.isSecret("Server.Save.Locations"));
+   }
+
    /**
     * The pre-#76006 predicate, kept here so a test can assert that a name is covered by the
     * writer half of the union rather than by the name half. Deliberately a copy: if


larryliang-inetsoft:
# Review: PR #4843 (fix/serversave-secret)

Reviewer: debug-emlocale-crash (reused, fully uninvolved in this bug diagnosis/refutation/fix).
Read the full trail (01-diagnosis.md, 02-refute.md, 03-fix.md) first, then independently re-read
the live worktree at C:\src\agent\wiz\stylebi-bugfix-serversave-secret (branch fix/serversave-secret,
commit 26a8f0e413) and pulled the actual PR diff/body via gh, rather than trusting the docs alone.
This is a security-relevant fix (credential exposure), so every check below was verified against
the current file contents, not just the docs describing them.

## Verdict: APPROVE (content). Merge held on CI, see end of doc.

## 1. COMPOSITE_SECRET_PROPERTIES unioned into isSecret() only -- CONFIRMED

Read the live diff and the current AdminPropertyCatalog.java directly (lines 264-278, 348-350,
383-386, 396-416):

- isSecret() line 275: return isEncryptedCredential(lower) or COMPOSITE_SECRET_PROPERTIES.contains(lower) or ...
  the new set is unioned here, and only here.
- isEncryptedCredential() (line 348-350) is untouched: returns baseName != null and ENCRYPTED_CREDENTIALS.contains(baseName.toLowerCase());
  no reference to the new set.
- ENCRYPTED_CREDENTIALS (line 383-386) is untouched: still exactly the seven pre-existing entries
  (openid.client.secret, stylebi.google.openid.client.secret, mail.smtp.pass, mail.smtp.clientsecret,
  mail.smtp.accesstoken, mail.smtp.refreshtoken, log.fluentd.security.sharedkey). server.save.locations
  is not among them.

This is exactly the scoped fix described in 03-fix.md -- no drift between what the docs claim and
what the diff actually does.

## 2. End-to-end read withholding -- traced myself, CONFIRMED

Read AdminPropertiesController.view() directly (lines 105-153), not just the isSecret boolean:

    boolean credential = AdminPropertyCatalog.isEncryptedCredential(name.baseName());   // false
    boolean secret = AdminPropertyCatalog.isSecret(name.baseName());                    // true (after fix)
    String stored = secret ? null : SreeEnv.getProperty(name.key(), false, false);      // null

With credential=false and secret=true for server.save.locations, stored is unconditionally null --
SreeEnv.getProperty is never called for this property value at all (matching the existing "a
secret value is not read AT ALL" comment at line 110-113, which the codebase already asserts and
tests elsewhere via a never() Mockito verify). The method then falls into the else-if(secret)
branch (line 149-153): "This name matches admin-chat secret pattern, so its value is neither read
nor changed here." This is the correct branch and correct wording -- the if(credential) branch
above it (line 141-148), which claims a value "CAN still be changed... the server encrypts it on
write", is NOT taken, which matters: that branch wording would be actively false for this property
(nothing encrypts it, and after this fix writes are refused -- see item 3 below).

## 3. AdminChangePlanService write-refusal -- traced myself, CONFIRMED intentional and correct

Read AdminChangePlanService.java line 78-93 directly:

    boolean credential = AdminPropertyCatalog.isEncryptedCredential(name.baseName());   // false
    if(AdminPropertyCatalog.isSecret(name.baseName()) and !credential) {                // true and true
       throw new IllegalArgumentException(name.key() + ": secret properties cannot be changed through admin-chat");
    }

Before the fix, isSecret was false, so this check did not fire and admin-chat preview_changes/
apply_changes pipeline could write an arbitrary plaintext string to server.save.locations (a
second, independent plaintext-write exposure beyond the read this PR is nominally about). After
the fix, this now throws for any admin-chat-driven write. This is a correct and desirable side
effect, not scope creep, and the PR body describes it accurately as closing an equivalent
plaintext-write path that is open today, without overclaiming it as a separate fix. Confirmed by
direct reading, not just accepting the docs characterization.

## 4. New tests do the precise 3-way distinction -- CONFIRMED

Read AdminPropertyCatalogTest.java lines 279-301 directly (not just the docs description):

    assertFalse(matchesTheNamePattern("server.save.locations"), ...);
    assertFalse(AdminPropertyCatalog.isEncryptedCredential("server.save.locations"), ...);
    assertTrue(AdminPropertyCatalog.isSecret("server.save.locations"), ...);

This is the exact 3-way distinction the fix depends on: name-shape predicate false, encrypted-
credential false, isSecret true -- i.e. the COMPOSITE_SECRET_PROPERTIES union is demonstrably the
only thing making isSecret true for this name, not an accidental overlap with the other two
mechanisms. The second test (casing variant, Server.Save.Locations -> isSecret true) matches the
exact shape of the pre-existing casing tests for other entries in this file (e.g.
withholdsACredentialWhateverCasingTheCallerUses), so it is consistent with established convention,
not a weaker one-off.

Independently re-ran (not just trusted the PR reported run):

    ./mvnw -o -pl core -am test -Dtest=AdminPropertyCatalogTest,AdminPropertiesControllerTest ...

Result: Tests run: 45, Failures: 0, Errors: 0 (29 in AdminPropertyCatalogTest, 16 in
AdminPropertiesControllerTest), full core module BUILD SUCCESS. Matches the PR reported
29-passing figure for the catalog test file exactly.

## 5. Missing controller-level test for this specific property -- ACCEPTABLE, not a blocker

Checked AdminPropertiesControllerTest.java directly for whether the generic isSecret=true and
credential=false branch is already exercised by an equivalent real property, since that is the PR
stated reason for not adding a server.save.locations-specific controller test:

    @Test
    void withholdsTheValueOfASecretPropertyButStillListsIt() {
       // password.encryption.key ...
       PropertyView view = controller.get("password.encryption.key", principal);
       assertNull(view.currentValue());
       ...
       sreeEnv.verify(() -> SreeEnv.getProperty("password.encryption.key", false, false), never());
    }

password.encryption.key is isSecret=true, isEncryptedCredential=false -- the exact same branch
shape server.save.locations now takes. This confirms the PR claim: the controller-level behavior
for this branch is already under test via an equivalent property, so the AdminPropertyCatalogTest
coverage (which is property-name-specific and is the actual new logic) plus this pre-existing
controller coverage (which is branch-shape-specific and already exists) together are sufficient.
I agree this is adequate coverage and do not consider the missing property-specific controller
test a gap worth blocking on -- it would be duplicating an already-tested branch, only swapping
the trigger property name.

## 6. ServerPathInfoModel / PLACEHOLDER_PASSWORD claim -- accurately described, not overstated

Read ServerPathInfoModel.java lines 71-87 (Builder.from) directly:

    password(Tool.isEmptyString(info.getPassword()) ? null : Util.PLACEHOLDER_PASSWORD);

Confirmed: this masking mechanism is real, and is exactly as the PR body describes it -- wired to
the structured, typed Enterprise Manager schedule-configuration screen, not the generic
name/value-string AdminPropertyCatalog/AdminPropertiesController path this bug is about. The PR
body does not claim to fix or touch the EM UI path, and does not overstate this PR scope as
covering it. The new COMPOSITE_SECRET_PROPERTIES javadoc in AdminPropertyCatalog.java (lines
396-402) also references this mechanism accurately, for the benefit of a future reader who might
otherwise (like the original diagnosis did, before refutation caught it) wrongly conclude no
masking exists anywhere for this property.

## Additional check: CI

claude-review shows fail on this PR, but this is a pre-existing, batch-wide infra failure (empty
ANTHROPIC_API_KEY in the GitHub Actions env, confirmed by reading the actual run log -- the step
errors out in 439ms before doing any review work) -- the same check fails identically on PR #4842
(still under review) and failed on #4839 (already merged), so it is not a signal specific to this
PR content and is not a legitimate blocker. build is still pending as of this review. Per the
batch established pattern (see #4841/format-time, held in reserve for merge once green), merge is
being held while build is pending -- not merging on Semgrep/CLA alone.

## Summary

No blockers found. The security-relevant core of this fix -- which set the new withhold-list is
unioned into, and specifically NOT into ENCRYPTED_CREDENTIALS -- is exactly as claimed and
verified directly against the live diff, not inferred from the docs. Read withholding, write
refusal, test coverage, and the accuracy of the PR claims about pre-existing masking machinery all
check out under independent tracing. Content approved; holding merge on the pending build check.

PR #4845, opened 2026-08-28
Fix cross-tenant webmap.suspend.until leak (multi-tenancy isolation)

larryliang-inetsoft:
## Root cause

When an org's chart render hits a Mapbox/Google Maps quota error (HTTP 402, `WebMapLimitException`), `AssemblyImageService`'s catch block disables web-map rendering for 24 hours by writing `webmap.suspend.until` via the plain, process-global `SreeEnv.setProperty(String, String)` overload. `MapGenerator.isWebMap` reads and self-clears the same key via the equivalent unscoped overloads.

Because the write is never scoped to an organization, **one tenant's map-service quota error disables web-map rendering for every tenant on a multi-tenant install for 24 hours**, with no Enterprise Manager visibility into the suspension (no UI surfaces this property at all).

## Fix

Switch all three call sites to the existing org-scoped `SreeEnv` overloads — the same mechanism already used for other multi-tenant settings (`ResourcePermissionService`, `UserTreeService`, `AppDomainUtils`), proven end-to-end by `OrgLifecycleScopedPropertiesIntegrationTest`:

- `AssemblyImageService.java`: `SreeEnv.setProperty("webmap.suspend.until", hours24 + "", true)`
- `MapGenerator.java` (read): `SreeEnv.getProperty("webmap.suspend.until", false, true)`
- `MapGenerator.java` (self-clear on expiry): `SreeEnv.setProperty("webmap.suspend.until", null, true)`

`webmap.suspend.until` is not in `PropertiesEngine.EXCLUDED_ORG_PROPERTIES`, so nothing blocks scoping it, and the org-scoped read/write both gracefully fall back to the global key when no principal/org is available, so there's no regression risk for single-tenant installs or edge-case unauthenticated contexts.

## Security / multi-tenancy relevance

This is a tenant-isolation bug: without the fix, any organization can inadvertently (or deliberately, by exhausting its own map-service quota) disable a feature for every other organization on a shared install for 24 hours.

## Testing

Added `MapGeneratorWebMapSuspendOrgScopeTest` (mirrors `OrgLifecycleScopedPropertiesIntegrationTest`'s `actAs(orgId)` pattern):
- Org A triggers a suspension → `MapGenerator.isWebMap` returns `false` for org A.
- Org B (untouched) → `isWebMap` is **not** suspended by org A's quota error (this assertion reproduces the bug pre-fix).
- Expiry clears cleanly (self-clear removes the org-scoped key) without cross-org effect.

Verified via `./mvnw -o -pl core -am test -Dtest=MapGeneratorWebMapSuspendOrgScopeTest`:
- With the fix: 1 test, 0 failures.
- With the fix reverted: fails (`expiry must self-clear the org-scoped key` — the unscoped self-clear leaves stale org-scoped state).
- Also independently confirmed (throwaway probe, not committed) that reverting *only* the `AssemblyImageService.java` write to the plain unscoped overload reproduces the full cross-org leak even with `MapGenerator`'s read/self-clear fixed — confirming the write-side change is the load-bearing part of the fix.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/MapGenerator.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/MapGenerator.java
@@ -1242,7 +1242,7 @@ public class MapGenerator extends MergedGraphGenerator {
          return false;
       }
 
-      String suspend = SreeEnv.getProperty("webmap.suspend.until");
+      String suspend = SreeEnv.getProperty("webmap.suspend.until", false, true);
 
       if(suspend != null) {
          long suspendTS = Long.parseLong(suspend);
@@ -1251,7 +1251,7 @@ public class MapGenerator extends MergedGraphGenerator {
             return false;
          }
 
-         SreeEnv.setProperty("webmap.suspend.until", null);
+         SreeEnv.setProperty("webmap.suspend.until", null, true);
       }
 
       String service = SreeEnv.getProperty("webmap.service");
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/AssemblyImageService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/AssemblyImageService.java
@@ -735,7 +735,7 @@ public class AssemblyImageService {
                // web map limit exceeded, disable web map for 24 hours.
                catch(WebMapLimitException ex) {
                   long hours24 = System.currentTimeMillis() + 24 * 60 * 60000L;
-                  SreeEnv.setProperty("webmap.suspend.until", hours24 + "");
+                  SreeEnv.setProperty("webmap.suspend.until", hours24 + "", true);
 
                   box.get().clearGraph(name);
                   return processGetAssemblyImage1(rvs, aid, width, height, maxWidth, maxHeight, aname,
--- a/core/src/test/java/inetsoft/report/composition/graph/MapGeneratorWebMapSuspendOrgScopeTest.java
+++ b/core/src/test/java/inetsoft/report/composition/graph/MapGeneratorWebMapSuspendOrgScopeTest.java
@@ -1,0 +1,118 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.report.composition.graph;
+
+/*
+ * Regression coverage for the webmap.suspend.until multi-tenant leak: one org's Mapbox/Google
+ * Maps 402 quota error (AssemblyImageService's WebMapLimitException catch block) used to disable
+ * web-map rendering for every organization on the install for 24 hours, because both the write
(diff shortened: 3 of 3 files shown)


jshobe-inetsoft (approved):
## Summary

This PR fixes a real multi-tenant isolation bug: `webmap.suspend.until` (the flag written when a Mapbox/Google Maps quota error occurs, to disable web-map rendering for 24h) was being read/written with the plain, process-global `SreeEnv` overloads in three places. On a multi-tenant install, any one org exhausting its map-service quota would disable web-map rendering for **every** org for 24 hours. The fix switches all three call sites to the existing org-scoped overloads (`SreeEnv.getProperty(name, false, true)` / `SreeEnv.setProperty(name, val, true)`), the same mechanism already used elsewhere for other per-org settings.

## Correctness

- Verified `webmap.suspend.until` is not in `PropertiesEngine.EXCLUDED_ORG_PROPERTIES`, so org-scoping is not blocked for this key.
- Traced `PropertiesEngine.setProperty(name, val, orgScope=true)`: it resolves the acting org from `ThreadContext`/`OrganizationManager.getCurrentOrgID()` and writes to `inetsoft.org.<orgId>.<name>` when an org is present, falling back to the plain global key when no principal/org is available (e.g. background/unauthenticated contexts) — exactly as described in the PR body.
- Traced the read path (`useAvailableOrgProperty`): it looks for the org-scoped key first and falls back to the global key if the org-scoped key was never written — so once the write side is fixed, org B never observes org A's suspension.
- The self-clear path (`SreeEnv.setProperty(name, null, true)` in `MapGenerator.isWebMap`) correctly removes only the org-scoped key (via `remove(name)` on the already-qualified name), not the global key or another org's key.
- One subtlety worth noting (not a blocker): the *read* side (`SreeEnv.getProperty(name)`, the plain 1-arg form used before this PR) was already implicitly org-scoped by default (`PropertiesEngine.getProperty(name, false)` → `getProperty(name, false, true)`), so the read-side changes in `MapGenerator` are mostly a no-op for behavior / a clarity improvement. The actual root cause was entirely on the **write** side (`AssemblyImageService`'s unscoped `setProperty`, and the unscoped self-clear `setProperty(..., null)` in `MapGenerator`), which is correctly fixed here.
- Grepped the whole repo for `webmap.suspend.until` / `WebMapLimitException` — all three call sites are covered by this PR; no other place reads or writes this property.
- Confirmed via CI logs that the new test (`MapGeneratorWebMapSuspendOrgScopeTest`) actually ran and passed (`Tests run: 1, Failures: 0`).

## Tests

- The new `MapGeneratorWebMapSuspendOrgScopeTest` follows the established `actAs(orgId)` pattern from `OrgLifecycleScopedPropertiesIntegrationTest` (sets both `ThreadContext` principal and `OrganizationContextHolder`), and its assertions line up correctly with the traced `PropertiesEngine` logic (including the "read the raw already-prefixed key" idiom used to assert the org-scoped key was actually created/cleared).
- Confirmed no other existing test depends on the old unscoped behavior of these three call sites (searched for tests touching `AssemblyImageService` or `webmap` — none overlap with this property).
- `@AfterEach` cleans up thread-local principal/org context and the global webmap service properties; no leftover state since org A's suspend key is exercised through its full expire/self-clear lifecycle within the test body.

## Style / minor observations (non-blocking)

- The fix is minimal and scoped exactly to the three call sites that needed it — no incidental refactoring.
- No new public API surface, no backward-incompatible signature changes; both old and new `SreeEnv` overloads already existed and are used elsewhere, so there's no risk to other callers.
- Security: this is itself a tenant-isolation fix; no new vulnerabilities introduced. The fallback-to-global behavior when no principal/org context is available is an accepted, pre-existing limitation of the org-scoped properties mechanism in general (shared by every other property using this pattern), not something this PR needs to solve.

## Verdict

No blocking issues found. The fix is correct, narrowly scoped, well-tested, and consistent with the existing org-scoped-properties pattern used elsewhere in the codebase.

**Checks:** all non-`claude-review` checks pass (Semgrep OSS, Semgrep Scan, build, license/cla). The `claude-review` check failed due to an infrastructure error in the GitHub Action itself (`is_error:true` after 301ms with 0 turns of actual review), not a substantive finding — excluded per review policy.

**Mergeability:** `mergeable: MERGEABLE`, no merge conflicts.

**Recommendation: Approve.**

PR #4846, opened 2026-08-28
Fix path traversal in Look and Feel CSS/logo/favicon upload (76360)

larryliang-inetsoft:
## Summary
- Bug 76360: `LookAndFeelService.setViewsheet`'s org-scoped branch, and `setLogo`/`setFavicon` on both scopes, wrote the caller-supplied `FileData.name()` (or, for logo/favicon, the substring from its last `.` onward) straight into `DataSpace.withOutputStream` with no validation.
- `DataSpace.sanitizePathComponent` does not detect or strip `..` segments, so a crafted name like `../../../etc/evil.css` or `a.b/../../evil.png` reached the storage write call unmodified. Confirmed via a full read of the `PresentationSettingsController` → `LookAndFeelService` → `DataSpace` → `BlobStorage` call chain (see linked diagnosis/refutation docs), independently re-verified line-by-line by a second reviewer pass.
- Fix: add `LookAndFeelService.requireSafeFileName()`, which rejects any name containing `/`, `\`, `..`, or a NUL byte, and call it on the raw caller-supplied name at all three call sites, before it's used for anything (including before the `setLogo`/`setFavicon` last-dot substring derivation, since that derivation itself can carry a traversal segment through).
- Deliberately does **not** touch the shared `DataSpace.sanitizePathComponent`: it has ~10 other call sites in `DataSpace.java` that legitimately pass multi-segment `dir` paths, so hardening it globally risks unaudited breakage. The narrow, call-site-scoped fix matches the actual defect (an attacker-controlled *leaf* filename with no validation) without that risk.

## Root cause
An ordinary (non-site-admin) org-admin request to `POST /api/em/settings/presentation/model` in a multi-tenant, security-enabled deployment hits the vulnerable org-scoped branches — this is the normal case, not an edge case. `FileData.name()` has no Bean Validation annotation anywhere in the chain, and nothing in `LookAndFeelService` validated it before using it as a filesystem leaf name.

## Test plan
- [x] New `core/src/test/java/inetsoft/web/admin/presentation/LookAndFeelServiceTest.java`: traversal-shaped names rejected on the org-scoped `setViewsheet` branch and on both scopes of `setLogo`/`setFavicon`; the global `setViewsheet` branch is unaffected (it always writes the literal `format.css`, never the caller's name); ordinary safe names still succeed; a name with no dot is unaffected.
- [x] `./mvnw test -pl core -am -Dtest=LookAndFeelServiceTest` — 10/10 pass.
- [x] `./mvnw test -pl core -am -Dtest="inetsoft.web.admin.presentation.**"` — full package suite, no regressions.
- [x] Sanity check: temporarily reverted the fix and confirmed the 5 rejection-path tests fail (then restored the fix).

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/presentation/LookAndFeelService.java
+++ b/core/src/main/java/inetsoft/web/admin/presentation/LookAndFeelService.java
@@ -489,6 +489,20 @@ public class LookAndFeelService {
       }
    }
 
+   /**
+    * Rejects a caller-supplied file name that contains a path separator, a ".." traversal
+    * segment, or a NUL byte, since none of these are valid in a CSS/logo/favicon file name.
+    */
+   private static String requireSafeFileName(String name) {
+      if(name == null || name.isEmpty() || name.contains("/") || name.contains("\\") ||
+         name.contains("..") || name.contains("\0"))
+      {
+         throw new IllegalArgumentException("Invalid file name: " + name);
+      }
+
+      return name;
+   }
+
    private void setLogo(FileData logo, DataSpace space, String directory, Principal principal, boolean globalSettings) throws Exception {
       final PortalThemesManager manager = portalThemesManager;
       final String orgID = OrganizationManager.getInstance().getCurrentOrgID();
@@ -510,7 +524,7 @@ public class LookAndFeelService {
       }
       else {
          byte[] logoData = Base64.getDecoder().decode(logo.content());
-         int dotIndex = logo.name().lastIndexOf(".");
+         int dotIndex = requireSafeFileName(logo.name()).lastIndexOf(".");
          String logoName = "logo" + (dotIndex >= 0 ? logo.name().substring(dotIndex) : ".gif");
          InputStream in = new ByteArrayInputStream(logoData);
 
@@ -547,7 +561,7 @@ public class LookAndFeelService {
 
       if(favi != null) {
          byte[] faviData = Base64.getDecoder().decode(favi.content());
-         int dotIndex = favi.name().lastIndexOf(".");
+         int dotIndex = requireSafeFileName(favi.name()).lastIndexOf(".");
          String faviconName = "favicon" + (dotIndex >= 0 ? favi.name().substring(dotIndex) : ".gif");
          InputStream in = new ByteArrayInputStream(faviData);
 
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
## Review round 1 -- approved (posted as comment: GitHub blocks self-approval on this PR, same account authored it)

**Head commit reviewed:** 24231e312 (full PR, first review round).

**What was checked:** Read the diagnosis/refutation/fix docs, then independently re-read the actual PR diff and the current bodies of `setLogo`/`setFavicon`/`setViewsheet`/`requireSafeFileName`/`writeStyleFile` and `DataSpace.sanitizePathComponent` directly from the branch (not the docs' paraphrase).

- Confirmed `requireSafeFileName` is called before any write on all three call sites (`setLogo`, `setFavicon`, `setViewsheet` org-scoped branch); the `setViewsheet` global branch is correctly left unchanged since it never reads the caller's name.
- Tried to build a bypass of the character-class check (`/`, `\`, `..`, NUL): URL-encoding doesn't apply (value never goes through URL-decoding in this chain), Unicode lookalikes aren't interpreted as separators by the filesystem/blob layer, and Windows reserved device names can't reach the physical filesystem as a literal name on the local backend since `FilesystemBlobEngine` is content-addressed (confirmed by reading it directly) -- this matches the diagnosis/refutation's already-acknowledged, correctly-scoped boundary. No bypass found that escapes the intended directory tree.
- Diff scope is clean: only `LookAndFeelService.java` and the new test file changed.
- Read the full 10-test regression file: the 5 rejection tests are exactly the ones that would fail on revert (matches the fixer's claim by static inspection), and the 5 acceptance tests guard against over-tightening on legitimate filenames.

**Findings against this diff:** none.

**Separate item, not a blocker:** the identical unvalidated-name-into-`DataSpace` pattern is still live in this same file's font-upload path (`updateFonts`/`saveFontFaceFiles`/`writeFontCSS`/`deleteFontFiles`, keyed off `FontFaceModel.getFileNamePrefix()`, which is just as caller-controlled as `FileData.name()` was). The diagnosis already flagged this as out of this ticket's scope, and I agree this PR correctly didn't expand into it -- but it's a real, currently-unfixed instance of the same vulnerability class, reachable the same way. Recommending a follow-up bug for it outside this PR.

**Verdict: approved.**

PR #4847, opened 2026-08-28
Fix NPE in AuthenticationProviderService.getAuthenticationProvider on unknown name

larryliang-inetsoft:
## Summary
- `getAuthenticationProvider(name)` threw a raw `NullPointerException` (not a clean, named error) when `name` didn't match any configured provider, because the pattern-matching switch's `case null, default ->` arm still passed the null `selectedProvider` into `AuthenticationProviderModel.Builder.customProviderModel()`, which unconditionally dereferences it.
- Exposed directly via two EM controller endpoints (`GET /api/em/security/get-authentication-provider/{providerName}` and the copy-provider endpoint), and, under a delete-race (provider present at plan time, removed by apply time), via `ProviderChangesetApplyService.applyDeleteAuthentication`'s unguarded call at line 298.
- Fix: add a null guard right after the lookup that throws `MessageException` naming the missing provider, matching the existing precedent already used in this class's `editAuthenticationProvider` (line ~208). `MessageException` already has its own Spring exception handler mapping to a clean client-facing error — no controller-layer changes needed.

## Test plan
- [x] New `AuthenticationProviderServiceTest` — unknown name throws `MessageException` (not NPE) naming the provider; known FILE provider still returns its model correctly.
- [x] New `ProviderChangesetApplyServiceTest.deleteRaceUnknownNameAtApplyTimeSurfacesAsCleanFailedOutcomeNotNpe` — simulates the delete-race at the previously-unguarded `ProviderChangesetApplyService.java:298` call site; confirms it now surfaces as a clean, named `STATUS_FAILED` outcome via the existing generic `catch(Exception e)` in `apply()`, not an unhandled NPE.
- [x] Sanity-reverted the fix and confirmed both new tests fail with the original raw `NullPointerException` before re-applying.
- [x] Full `inetsoft.web.admin.**` package: 715 tests run, 0 failures, 0 errors.

Fixes bug #76358.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/security/AuthenticationProviderService.java
+++ b/core/src/main/java/inetsoft/web/admin/security/AuthenticationProviderService.java
@@ -108,6 +108,12 @@ public class AuthenticationProviderService extends BaseSubscribeChangeHandler {
    public AuthenticationProviderModel getAuthenticationProvider(String name) {
       boolean enterprise = LicenseManager.isEnterprise();
       AuthenticationProvider selectedProvider = getProviderByName(name);
+
+      if(selectedProvider == null) {
+         throw new MessageException(
+            "Authentication provider named \"" + name + "\" does not exist");
+      }
+
       AuthenticationProviderModel.Builder builder = AuthenticationProviderModel.builder()
          .providerName(name)
          .oldName(name)
--- a/core/src/test/java/inetsoft/web/admin/ai/providers/ProviderChangesetApplyServiceTest.java
+++ b/core/src/test/java/inetsoft/web/admin/ai/providers/ProviderChangesetApplyServiceTest.java
@@ -20,6 +20,7 @@ package inetsoft.web.admin.ai.providers;
 import inetsoft.sree.security.*;
 import inetsoft.uql.XPrincipal;
 import inetsoft.uql.util.Identity;
+import inetsoft.util.MessageException;
 import inetsoft.util.audit.AdminChangeRecord;
 import inetsoft.web.admin.ai.AdminBackupService;
 import inetsoft.web.admin.ai.AdminChangesetApplyService;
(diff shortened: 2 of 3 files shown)


larryliang-inetsoft:
# Review round 1 -- Bug #76358 / PR #4847

Head commit reviewed: `5f9962670aaeb2cd9646fdadf7c66d5a4dca84ad` (only commit on `fix/bug-76358`,
base `stylebi-wiz-main-rebase`).

## Invariant this change claims to hold

`AuthenticationProviderService.getAuthenticationProvider(name)` never dereferences a null
provider -- an unknown name always surfaces as a named `MessageException`, never a raw NPE,
regardless of which caller invokes it or how the miss occurred.

## What I independently verified (not just trusted from the docs)

1. **Diff scope** -- `gh pr diff 4847` shows exactly 3 files: the one-line-guard fix in
   `AuthenticationProviderService.java`, a new `AuthenticationProviderServiceTest.java`, and an
   extension to the existing `ProviderChangesetApplyServiceTest.java`. No unrelated changes.

2. **The null-guard genuinely closes the NPE.** Read the current
   `AuthenticationProviderService.java:108-142` directly (not the fix doc's excerpt): the guard
   sits immediately after `getProviderByName(name)` and before the pattern-matching switch, so
   `case null, default ->` at line 135-138 can no longer receive a null `selectedProvider` --
   `customProviderModel(selectedProvider, objectMapper)` at line 137 is now unreachable with
   `selectedProvider == null`. Confirmed `MessageException` resolves via the existing
   `import inetsoft.util.*;` wildcard (no new import needed, matches the fix doc's claim).
   `default` still correctly covers a genuinely-unrecognized non-null provider type, so no
   behavior is lost for real custom providers.

3. **Ran the full affected test set myself** from the fixer's worktree at commit
   `5f9962670`, not the CI/doc-reported numbers: `AuthenticationProviderServiceTest` (2/2),
   `AuthenticationProviderControllerTest` (6/6), `ProviderChangesetApplyServiceTest` (13/13),
   `ProviderChangePlanServiceTest` (29/29) -- all green, 0 failures/errors, matching the fix doc
   exactly.

4. **The `ProviderChangesetApplyService.java:298` regression test's stack-trace-matched stub is
   sound, not fragile.** Read `applyDeleteAuthentication` (lines 293-328) and `apply()`'s per-item
   `catch(Exception e)` (lines 99-112) in full. Two things worth recording:
   - The method already has its *own* race-handling for a vanished provider (the
     `index < 0` branch at lines 303-312, producing a clean "provider not found at apply time
     (concurrent change)" `STATUS_FAILED` outcome) -- but that check runs *after* line 298's
     unguarded `getAuthenticationProvider(name)` call, so the pre-fix NPE fired before that
     existing graceful path was ever reached. The fix doesn't change that ordering, but it does
     make the earlier, unguarded call fail cleanly instead of crashing, which is exactly the
     claimed fix.
   - On the flakiness question specifically asked about: if a future refactor renamed or removed
     `applyDeleteAuthentication` such that the stack frame match stops firing, the stub would
     always return `victimModel` (success), and the test's `assertEquals(STATUS_FAILED, ...)`
     assertion would then fail loudly -- it does not silently stop testing anything. Mockito's
     `thenAnswer` runs synchronously on the calling thread, so `Thread.currentThread()
     .getStackTrace()` reliably includes the real `applyDeleteAuthentication` frame at the time of
     the call; this isn't sensitive to JIT inlining since stack-trace capture reflects the logical
     Java call stack. I consider this test-design choice sound.

5. **Regression tests are load-bearing.** Read both new/changed test bodies in full.
   `getAuthenticationProvider_unknownName_throwsMessageException` asserts `MessageException` (not
   NPE) with a message containing the name; I did not just trust the fix doc's claimed revert
   check, but the doc's description of the pre-fix stack trace
   (`AuthenticationProviderModel.Builder.customProviderModel(AuthenticationProviderModel
   .java:147)`) matches exactly what I'd expect from the confirmed unconditional dereference at
   that line -- credible. The `ProviderChangesetApplyServiceTest` addition asserts both a positive
   (`STATUS_FAILED` with "does not exist") and a negative (message does NOT contain
   "NullPointerException") condition, so a narrower/still-broken fix (e.g. one that only changed
   the message text without actually preventing the NPE) would not pass it.

6. **Same-bug sweep.** Grepped for the same shape elsewhere: `AuthorizationProviderService
   .getAuthorizationProvider` (lines 58-78) has the structurally identical
   `selectedProvider = getProviderByName(name)` -> later `customProviderModel(selectedProvider,
   ...)` pattern, but it already has its own null-guard at lines 61-63 (throws
   `MessageException` via `Catalog`). Not exposed -- no other site found.

## `requireExists` left in `AdminProviderController.java` -- judgment call requested

Agree with leaving it. It's inert defense-in-depth now (the service-layer guard fires first for
any caller), costs nothing, and removing it would be pure scope creep unrelated to closing this
bug -- if anything it's a second layer that's cheap insurance against the shared-service guard
ever being weakened later. Not worth a follow-up ticket either; this is the kind of harmless
redundancy that isn't worth spending review cycles on again.

## Findings

None at blocker or important severity. No overfit smells (the fix is in the shared service, not
narrowed to the reported endpoint or repro name), no scope creep, tests are behavior-asserting
not literal-repro-only.

## Verdict: **approved**

PR #4848, opened 2026-08-28
Bug #76005, disable scroll anchoring on tree containers to fix scrollbar jump

jkidacki-inetsoft:
## Summary

When a viewsheet has many assemblies, the Script tab's column tree (e.g. Properties > Script > Component/ComboBoxN) switches to a virtual-scroll rendering mode once the tree exceeds 500 nodes. Clicking the scrollbar's down arrow a second time, or scrolling with the mouse wheel, would cause the scrollbar to snap straight to the end of the list instead of scrolling incrementally.

## Root Cause

The virtual-scroll tree (`TreeComponent` with `useVirtualScroll` enabled, used by `virtual-scroll-tree` in the script pane) renders only a windowed subset of nodes, padded above/below by spacer `<div>`s whose heights are recalculated synchronously on every native `scroll` event (`calculateBounds()` in `tree.component.ts`), then applied immediately via `ChangeDetectorRef.detectChanges()`.

Because the spacer div above the visible content resizes synchronously in response to scrolling, Chrome's CSS scroll-anchoring feature detects the shift and auto-adjusts `scrollTop` to compensate. That adjustment fires another `scroll` event, which recalculates and resizes the spacer again — a feedback loop that snowballs the scroll position toward the end of the list.

This is a known pattern elsewhere in the codebase: `VirtualScrollTreeDatasource.restoreScrollTop()` is explicitly commented as a hack to work around "scroll anchoring type behavior", but it's only wired up for the binding tree, not this virtual-scroll-tree path. Disabling scroll anchoring via CSS (`overflow-anchor: none`) is the established fix used elsewhere for the same class of problem (`viewsheet-pane.component.scss`, `ws-details-table-data.component.scss`).

## Fix

Added `overflow-anchor: none;` to the shared `.tree-container` rule in `tree.component.scss`, the scrollable container used by every `TreeComponent` instance (virtual-scroll and non-virtual-scroll alike). This is a no-op for non-scrolling trees and only disables the browser's automatic scroll-position compensation, with no functional change to app logic.

## Test Plan

- `npm run build` (portal/em/elements/viewer-element) compiles clean.
- All tree-related test suites pass: `tree`/`virtual-scroll-tree` (271 tests) and `script-pane`/`vsassembly-script-pane` (39 tests).
- Manually reproduced the bug (viewsheet with many assemblies, Properties > Script tab, expand `Component/<assembly>`, click the scrollbar down arrow twice / mouse-wheel scroll) — scrolling now advances smoothly without jumping to the end.

Fixes Redmine #76005.

### DIFF
--- a/web/projects/portal/src/app/widget/tree/tree.component.scss
+++ b/web/projects/portal/src/app/widget/tree/tree.component.scss
@@ -63,6 +63,9 @@
     flex-grow: 1;
     //height: 100%;
     padding-left: 4px;
+    // disable scroll anchoring so resizing the virtual-scroll spacer divs on scroll doesn't
+    // trigger the browser to auto-adjust scrollTop, which caused runaway scroll jumps (76005)
+    overflow-anchor: none;
 
     &.vscrollable-always {
       overflow-y: auto;


jshobe-inetsoft (approved):
## Code Review

**Summary of change:** Adds `overflow-anchor: none;` to the shared `.tree-container` rule in `tree.component.scss` (the scrollable container used by every `TreeComponent` instance, both virtual-scroll and non-virtual-scroll). This is a single-line, CSS-only fix for Redmine #76005.

### Correctness
- The root-cause analysis is sound and verifiable in the code: `tree.component.ts` synchronously recalculates spacer div heights in `calculateBounds()` on every native `scroll` event and applies them via `detectChanges()`. Chrome's CSS scroll-anchoring feature reacts to that shift by adjusting `scrollTop`, which re-triggers `scroll`, producing the runaway-scroll feedback loop described.
- `overflow-anchor: none` is the correct, standard CSS mechanism to disable this browser behavior, and it is already an established pattern in this codebase for the identical class of problem (`viewsheet-pane.component.scss`, `ws-details-table-data.component.scss`), so this PR is consistent with existing conventions rather than introducing a new pattern.
- Confirmed there is only one `tree.component.scss` in the repo, and the `.tree-container` selector targeted is the same element (`#treeContainer`) that both the virtual-scroll and standard tree paths render into — so the fix is applied at the correct, single point of truth rather than needing to be duplicated elsewhere.

### Side effects / regressions
- The rule is added to the base `.tree-container` selector, not gated behind `.vscrollable-always` or `useVirtualScroll`, so it also applies to non-scrolling / non-virtual trees. This is intentional per the PR description and is a safe no-op for elements that aren't scrolling. For actively-scrolling non-virtual trees, disabling scroll anchoring only removes automatic compensation for content shifts above the viewport; since this component already manages `scrollTop` explicitly (see the `dataSource.scrollTop` subscription in `tree.component.ts`), it's unlikely to depend on browser-driven scroll anchoring for correct behavior, and the existing precedents in the codebase support this being safe.
- No JS/TS logic changed, so there is no risk of behavioral regression outside of scrolling/rendering feel.

### Backward compatibility
- Purely additive CSS property; no API, template, or public interface changes. No backward-incompatibility concerns.

### Tests
- No test changes included, which is appropriate here — this is a CSS-only fix and the affected behavior (browser scroll-anchoring interaction) is not something the existing Vitest/TestBed unit suites exercise. The PR description states the existing `tree`/`virtual-scroll-tree` (271 tests) and `script-pane`/`vsassembly-script-pane` (39 tests) suites still pass, and manual reproduction of the bug was verified fixed. No existing test is broken by this change.

### Security
- No security implications; this is a purely presentational CSS change with no new data handling, no user input processing, and no new DOM injection points.

### Readability / maintainability
- The added comment clearly explains why the rule exists and references the bug number, which will help future maintainers understand the non-obvious CSS/browser-behavior interaction. Minimal, well-scoped, easy to review.

### Verdict
No blocking issues found. All CI checks pass except `Claude Code Review / claude-review`, which is currently broken by an infrastructure/API issue on this repo (not a real review finding) and is excluded from consideration per current review policy. The PR has no merge conflicts (`mergeable: true`). Recommending **approval**.

PR #4849, opened 2026-08-28
Bug #75990, surface the permission-denied error when moving a schedule task in EM

jkidacki-inetsoft:
## Summary

Moving a schedule task in Enterprise Manager to a folder the user lacks WRITE
permission on correctly blocked the move server-side, but the EM UI's move
dialog just closed silently with no error shown to the user.

## Root Cause

`ScheduleTaskFolderService.moveScheduleItems()` throws a `MessageException`
when the destination folder lacks WRITE permission (added by the #75946 fix).
For EM's move-folder endpoint (`EMScheduleTaskFolderController`, package
`inetsoft.web.admin.schedule`), this exception is handled by
`AdminExceptionHandler`'s generic `Exception` fallback, which maps it to HTTP
**500**, not 403.

EM's `ScheduleTaskListComponent.moveTasks()` error handler only displayed a
message dialog when `error.status == 403`, so it never matched and nothing
was shown — even though the move was correctly rejected.

(Portal's equivalent move flow doesn't have this bug because its error
handler shows `error.error.message` regardless of status code.)

## Fix

Updated the EM `moveTasks()` error handler to show the server's specific
`MessageException` message (falling back to a generic message otherwise),
matching the existing pattern already used in the same file's `removeTasks()`
handler and in `repository-schedule-task-folder-settings-page.component.ts`.

## Test Plan

- [x] `ng build em` succeeds
- [x] `ng run em:test-tl` — all 818 tests pass (70 files), including the 35
      tests in `schedule-task-list.component.tl.spec.ts`
- [x] Code-reviewed (confirmed the fallback message doesn't misrepresent
      unrelated failures, matches established codebase convention)
- Manual repro: create a non-admin user with READ-only permission on a
  schedule folder, log in as that user in EM, attempt to move a task into
  that folder — a "permission denied" dialog with the specific server
  message now appears instead of the dialog silently closing.

Fixes Redmine #75990.

### DIFF
--- a/web/projects/em/src/app/settings/schedule/schedule-task-list/schedule-task-list.component.ts
+++ b/web/projects/em/src/app/settings/schedule/schedule-task-list/schedule-task-list.component.ts
@@ -458,10 +458,11 @@ export class ScheduleTaskListComponent implements OnInit, AfterViewInit, OnDestr
             this.http.post(TASKS_MOVE_URI, moveTaskFolderRequest).subscribe(() => {
                this.loadTasks();
             }, (error) => {
-               if(error.status == 403) {
-                  this.dialog.open(MessageDialog, this.setConfigs(`_#(js:Unauthorized)`,
-                     "_#(js:schedule.folder.moveTargetPermissionError)", MessageDialogType.ERROR));
-               }
+               const message = error.error != null && error.error.type == "MessageException" ?
+                  error.error.message : "Failed to move selected tasks";
+
+               this.dialog.open(MessageDialog, this.setConfigs(`_#(js:Error)`,
+                  message, MessageDialogType.ERROR));
             });
          }
       });


jshobe-inetsoft (approved):
## Code Review

**Summary:** Fixes Bug #75990 — moving a schedule task in EM to a folder the user lacks WRITE permission on used to close the move dialog silently instead of showing an error. The fix replaces a `error.status == 403` check in `ScheduleTaskListComponent.moveTasks()`'s error handler with a check on `error.error.type == "MessageException"`, surfacing the server's actual message (with a generic fallback otherwise).

### Correctness
- I verified the backend path end-to-end: `AdminExceptionHandler.handleException()` catches the `MessageException` thrown by `ScheduleTaskFolderService.moveScheduleItems()` and returns a `GenericError` via `ApiError(int, Throwable)`, which sets `type = cause.getClass().getSimpleName()` (i.e. literally the string `"MessageException"`) and `message = cause.getMessage()`. That's an HTTP 500, not 403 — which is exactly why the old `error.status == 403` check never matched, and confirms the new `error.error.type == "MessageException"` check is correct and matches the real response shape.
- The new code is copy-consistent with the existing, already-shipped pattern used a few dozen lines away in the same file's `removeTasks()` handler (and in `newTask()`, and in `repository-schedule-task-folder-settings-page.component.ts`), so it's a proven, low-risk pattern rather than a novel approach.
- The dialog title changes from `_#(js:Unauthorized)` to the more generic `_#(js:Error)`, appropriate since the handler now also covers non-permission `MessageException`s and the generic fallback case, not just 403s.

1. **Orphaned localization key**: `schedule.folder.moveTargetPermissionError` (used by the old code) is no longer referenced anywhere in the TS sources but remains defined in `srinter.properties` and all `srinter_*.properties` locale files. Not a functional problem, just dead i18n content that could be cleaned up in a follow-up.
2. **No new automated test** exercises the `moveTasks()` error path itself (neither before nor after this change — the existing `schedule-task-list.component.tl.spec.ts` suite doesn't cover it), so this fix relies on the manual repro described in the PR. Coverage isn't regressed, but it also isn't improved.
3. **Narrowed 403 handling**: a genuine 403 from the endpoint (e.g. `SecurityException`/`UndeclaredThrowableException` path in `AdminExceptionHandler`, which serializes `type` as `SecurityException`, not `MessageException`) will now fall through to the generic `"Failed to move selected tasks"` message rather than the old permission-specific wording. This is a narrow edge case (likely rare here, and still surfaces *an* error dialog rather than silently closing), so it's not a regression of the bug being fixed, just a slight loss of specificity in a case this PR wasn't targeting.
4. The hardcoded English fallback string `"Failed to move selected tasks"` isn't run through the `_#(js:...)` localization macro, but this exactly mirrors the pre-existing `removeTasks()` fallback (`"Failed to remove selected tasks"`), so it's consistent with established convention rather than a new issue.

### Change scope
Single-file, 5-line diff, tightly scoped to the reported bug. No other call sites or shared utilities are affected.

### Verdict
No blocking issues. This is a small, well-targeted, low-risk fix that follows an established, verified codebase pattern.

---

**Merge readiness check:**
- Status checks: all passing except `claude-review`, which is excluded per review instructions (known broken infra on this repo).
- Mergeability: `MERGEABLE`, no merge conflicts.
- No blocking issues identified in this review.

All three approval conditions are satisfied → **Approving**, but this should not be auto-merged.

PR #4850, opened 2026-08-28
Bug #76004, keep a date combobox on the day the user picked

ngovus-inetsoft:
## Problem

Reported against a viewsheet with a single `date` calendar combobox:

1. Select `1900-01-01` → the value changes to `1899-12-31`.
2. Then select `2025-12-31` → `Invalid date range`. Selecting `2025-12-30` first and then `2025-12-31` works. Only happens the first time.

## Root cause

Both symptoms are one cascading defect.

`serverTZ` is forced on for `XSchema.DATE` (`VSComboBoxModel:49-52`) because a date has no time component, so the value travelled as **epoch millis moved into the server time zone**. `dayjs.tz` resolves the offset through ICU, which models pre-1901 Local Mean Time; the server reads the same instant with the legacy `java.util.Date`/`TimeZone` stack, which reports a whole hour offset. For `Asia/Shanghai`:

```
browser: 1900-01-01T00:00:00 -> -2209017943000   (+08:05:43, ICU / java.time)
server:  TimeZone.getOffset(-2209017943000) = +08:00
         new Date(-2209017943000) = Sun Dec 31 23:54:17 CST 1899
```

`CoreTool.getDateData()` then truncates with the deprecated `Date.setHours(0)`, turning 5m43s into a whole day → **1899-12-31**. That is symptom 1.

That value came back as millis, was re-read with `dayjs.tz` (LMT again), and left `selectedDate = {1899,12,31}` **plus a stale `hours=0/min=5/sec=43`**. `getDateTime()` reused that time for the next pick, and `isValidDate()` compared the resulting *instant* against the max bound's exact instant — midnight of the max day, while `ngb-datepicker`'s `[maxDate]` is day granular. So the max day was rejected once and accepted after any earlier day had cleared the stale time: exactly the reported "2025-12-30 then 2025-12-31 works / only the first time".

## Fix

For `XSchema.DATE` only — transfer a wall clock `yyyy-MM-dd` string instead of an instant, and compare the range by day.

- `vs-combo-box.component.ts` — send/receive a wall clock date string (no `dayjs.tz`), zero the time components, parse min/max as wall clock for both the picker bound and the validation bound, and compare day-only.
- `VSComboBoxModel.java` — a calendar `date` value goes out as `yyyy-MM-dd` rather than millis.
- `VSInputService.java` — converts a wall clock date string with `java.sql.Date.valueOf(LocalDate.parse(...))`, which `Tool.getData()` then passes through untouched.

`timeInstant` and `time` keep the existing millis + `serverTZ` path unchanged, and `CoreTool.getDateData()` is not modified.

Two notes on the shape of the server change:

- It **builds the `java.sql.Date` itself** rather than handing the string to `Tool.getData()`. Just passing the string through is not enough: `CoreTool.parseDate()` resolves LMT while the truncation that follows does not, so it loses the same day for any LMT-era date — `Tool.getData("date", "1880-01-01")` returns `1879-12-31` under `America/New_York`. Only the literal `"1900-01-01"` survives, via an explicit special case at `CoreTool:1855` — evidently a band-aid for this same bug class. That underlying `CoreTool` defect is wider than the combobox and is left alone here; it deserves its own issue.
- The conversion lives in `applySelection0()`, not `applySelection()`, so the **submit button** path (`OnClickService` → `multiApplySelection`) gets it too — that path bypasses `applySelection()` entirely.

Also fixes `clearCalendar()` leaving the meridian untouched (`this.meridian == "AM"` compared instead of assigning).

## Testing

- **New frontend tests** — `vs-combo-box.component.interaction.tl.spec.ts` Group 16 (7 tests). Reverting only the component makes **5 of 7 fail**, and the pre-fix run reproduces the reported symptoms directly: the dialog opens for `1900-01-01`, and `hours` is 12 where 0 is expected. Full combobox suite **99/99 pass**.
- **New backend test** — `VSComboBoxDateValueTest` (7 tests). Two of them *pin the old defect* (the instant path, and `Tool.getData()`'s own string parse, both losing a day for an LMT-era date) so the fix is not guarding a non-problem. Surefire pins `America/New_York`, whose LMT era ends in 1883, so `1880-01-01` stands in for the reported `1900-01-01` under `Asia/Shanghai` — same mechanism, different cutover year.
- `ExtendedDateFormatTest`, `ToolTest`, `ToolTimeFormatTest`, `ComboBoxVSAScriptableTest`, `ApplyParameterToInputTest` — all pass. `core` compiles.
- `vsobjects` standard specs **64/64 pass**. The full portal TL suite fails 24-28 files on `main` *before* this change (flaky membership run to run); no combobox file is in either failing set, and the set-difference files pass in isolation.
- Verified manually against a freshly built server.

### DIFF
--- a/core/src/main/java/inetsoft/web/viewsheet/model/VSComboBoxModel.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/model/VSComboBoxModel.java
@@ -66,7 +66,16 @@ public class VSComboBoxModel extends ListInputModel<ComboBoxVSAssembly> {
       }
 
       if(calendar && selectedObject instanceof Date) {
-         selectedObject = ((Date) selectedObject).getTime();
+         // a date has no time component. send it as a wall clock date string instead of an
+         // instant, otherwise the browser and java.util.Date may disagree on the time zone
+         // offset (historical offsets that are not a whole number of minutes) and shift the
+         // date by a day.
+         if(XSchema.DATE.equals(dataType)) {
+            selectedObject = new SimpleDateFormat("yyyy-MM-dd").format(selectedObject);
+         }
+         else {
+            selectedObject = ((Date) selectedObject).getTime();
+         }
       }
       // used entered value, don't convert to full date in serializer
       else if(editable && selectedObject instanceof Date) {
--- a/core/src/main/java/inetsoft/web/viewsheet/service/VSInputService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/service/VSInputService.java
@@ -61,6 +61,7 @@ import org.springframework.web.server.ResponseStatusException;
 import java.awt.*;
 import java.security.Principal;
 import java.text.Format;
+import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 import java.util.*;
 import java.util.List;
@@ -161,7 +162,12 @@ public class VSInputService {
       Object selectedValue = event.value();
 
       try {
-         if(info.isCalendar() && selectedValue != null && !"".equals(selectedValue)) {
+         // a date is sent as a wall clock date string instead of an instant, and is
+         // converted in applySelection0() so that the submit button path gets it too
+         if(info.isCalendar() && isWallClockDate(info.getDataType(), selectedValue)) {
+            // not an instant, leave it alone
+         }
+         else if(info.isCalendar() && selectedValue != null && !"".equals(selectedValue)) {
             try {
                selectedValue = new Date(Long.parseLong(selectedValue.toString()));
             }
@@ -190,6 +196,39 @@ public class VSInputService {
       singleApplySelection(vsId, assemblyName, selectedValue, principal, dispatcher, linkUri);
 
       return null;
+   }
+
+   /**
+    * Check if the value is a wall clock date (yyyy-MM-dd), which is how the client sends the
+    * value of a date combobox. Such a value carries no time of day and no time zone, so it
+    * must not be treated as an instant.
+    */
+   private static boolean isWallClockDate(String dataType, Object value) {
+      return XSchema.DATE.equals(dataType) && value instanceof String &&
+         WALL_CLOCK_DATE.matcher((String) value).matches();
+   }
+
+   /**
+    * Convert a wall clock date string into the date it names, keeping the day exactly as
+    * picked. Both of the alternatives shift the day for a date inside a zone's Local Mean
+    * Time era: the browser resolves a zone offset through ICU, which models LMT, while
+    * java.util.Date reports a whole hour offset for the same instant, and CoreTool.parseDate
+    * splits the difference by parsing with LMT and truncating without it.
+    *
+    * @return the converted value, or the value unchanged if it is not a wall clock date.
+    */
+   private static Object convertWallClockDate(String dataType, Object value) {
+      if(!isWallClockDate(dataType, value)) {
+         return value;
+      }
+
+      try {
+         return java.sql.Date.valueOf(LocalDate.parse(value.toString()));
+      }
+      catch(Exception ex) {
+         LOG.debug("Not a valid date: {}", value, ex);
+         return value;
+      }
    }
 
    @ClusterWriteMethod
@@ -2804,7 +2843,15 @@ public class VSInputService {
 
       InputVSAssemblyInfo info = (InputVSAssemblyInfo) assembly.getVSAssemblyInfo();
       String type = assembly.getDataType();
-      Object obj = values == null || values.length == 0 ? null : Tool.getData(type, values[0]);
+
+      // don't write back into values[], it may alias the caller's array
+      Object firstValue = values == null || values.length == 0 ? null : values[0];
+
+      if(firstValue != null && info instanceof ComboBoxVSAssemblyInfo cinfo && cinfo.isCalendar()) {
+         firstValue = convertWallClockDate(type, firstValue);
+      }
+
+      Object obj = firstValue == null ? null : Tool.getData(type, firstValue);
 
       if(values !=null && values.length > 0 && obj == null && type.equals(CoreTool.BOOLEAN) && values[0].equals("")) {
          obj = "";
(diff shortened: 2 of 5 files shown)


jshobe-inetsoft (approved):
## Summary

This PR fixes Bug #76004: a `date` (day-only, no time-of-day) calendar combobox could shift the selected day by one (e.g. `1900-01-01` → `1899-12-31`) and could spuriously reject a valid max-bound day (`Invalid date range`) the first time it was picked.

Root cause, as explained in the PR description, is well diagnosed: `XSchema.DATE` values were transmitted as epoch-millis instants and resolved through two different offset models on the two ends (ICU/`java.time` on the browser vs. legacy `java.util.Date`/`TimeZone` on the server), which disagree by a few minutes for dates that fall inside a zone's pre-1901 Local Mean Time era. Truncating that instant to midnight turned the disagreement into a full day, and a leftover stale time-of-day then made the day-only range check compare instants instead of days.

The fix changes the wire format for `XSchema.DATE` values specifically to a wall-clock `yyyy-MM-dd` string (no time zone conversion at all), and changes the min/max validity check to compare days rather than instants when `dataType == DATE`. `time`/`timeInstant` combobox types are left on the existing millis + `serverTZ` path.

## Correctness / completeness

- The client (`vs-combo-box.component.ts`) and server (`VSComboBoxModel.java`, `VSInputService.java`) changes are symmetric and consistent: the client sends/reads `yyyy-MM-dd` only for `XSchema.DATE`; the server recognizes that shape via a `\d{4}-\d{2}-\d{2}` regex gated on `dataType == DATE`, and builds the `java.sql.Date` directly with `LocalDate.parse(...)` rather than routing through `CoreTool.parseDate()`/`Tool.getData()`'s string path (correctly identified as carrying the same LMT/truncation bug for other historical dates).
- I verified independently that `ComboBoxVSAssemblyInfo.minDate`/`maxDate` are serialized as `java.sql.Timestamp`, and that `ThirdPartySupportModule` registers a Jackson serializer for `Timestamp` using `"yyyy-MM-dd HH:mm:ss"` — this matches the `SERVER_DATE_TIME_FORMAT` the new `getDate()` branch parses first (falling back to the plain `yyyy-MM-dd` ISO format), so the min/max-bound parsing addition is consistent with the existing wire format rather than a guess.
- The `applySelection0()` conversion is applied to `firstValue` before the existing `Tool.getData()` call, and the PR correctly avoids writing back into the `values[]` array (which may alias the caller's array) — a small but real correctness detail.
- Placing the conversion in `applySelection0()` rather than only in the STOMP `applySelection` handler is a good catch: it also covers the submit-button path (`multiApplySelection`), which bypasses the single-value handler entirely.
- The `CheckBoxVSAssemblyInfo` branch of `applySelection0()` still calls `Tool.getData()` directly per-value without the new conversion, but that's correct/intentional — the wall-clock-date gate is `info instanceof ComboBoxVSAssemblyInfo cinfo && cinfo.isCalendar()`, and checkboxes don't have a calendar concept, so there's no regression there.
- The unrelated one-line fix in `clearCalendar()` (`this.meridian == "AM"` → `this.meridian = "AM"`) is a legitimate pre-existing bug (a no-op comparison instead of an assignment) and is small enough to bundle with this change without muddying the diff.

## Backward compatibility

- Old millis-as-string or numeric payloads still take the pre-existing path: `isWallClockDate()` requires `value instanceof String` and a `\d{4}-\d{2}-\d{2}` match, so a 13-digit millis string (or a `Long`) falls through unchanged to the old `Long.parseLong`/format-parse branch. Mixed old/new client-server pairs are not really a supported scenario in this monorepo's release model (frontend and backend ship together), so this is a non-issue in practice.
- Existing viewsheets whose `selectedObject` was already persisted with the old (possibly shifted-by-a-day) value are not retroactively corrected — reasonable, since that's a data-repair concern out of scope for this fix.
- `timeInstant`/`time` combobox behavior is provably untouched (guarded by `dataType == XSchema.DATE` checks throughout), and a dedicated new test (`should keep sending epoch millis for a timeInstant combobox`) pins that.

## Tests

- New backend test `VSComboBoxDateValueTest` (7 cases) exercises the private `isWallClockDate`/`convertWallClockDate` helpers via reflection, including a case that pins the two previously-buggy paths (`[G2]`) so the fix is demonstrably guarding a real problem, and a malformed-date case (`[G4]`) confirming the conversion fails soft (returns the value unchanged) rather than throwing.
- New frontend tests (Group 16, 7 cases) in `vs-combo-box.component.interaction.tl.spec.ts` cover the wall-clock round trip, day-only range validation with a stale leftover time, and the `timeInstant` non-regression case.
- The PR description reports the CI-covered test suites (which correspond to what actually ran and passed on the `build` check) plus manual verification against a built server. I did not see any indication that existing tests were broken by this change.
- Minor: testing private methods via reflection is a slightly heavier-weight pattern than exercising them through the public API, but given these are small, well-isolated pure functions and the surrounding class has no natural public seam to invoke them through directly, this is a reasonable and pragmatic choice.

## Minor nitpicks (non-blocking)

- In `VSInputService.applySelection()`, the `if(isWallClockDate) { /* not an instant, leave it alone */ }` branch with an empty body reads a little awkwardly; inverting the condition (`if(!isWallClockDate(...)) { ...existing parse... }`) would avoid the empty-branch pattern. Purely stylistic.
- `new SimpleDateFormat("yyyy-MM-dd")` in `VSComboBoxModel` is constructed fresh each time (thread-safe by construction) — no issue, just noting it as intentional given `SimpleDateFormat`'s well-known thread-safety pitfalls elsewhere in the codebase.

## Security

No security concerns identified. The new server-side parsing is a strict `LocalDate.parse()` on a regex-validated string, and any parse failure is caught and degrades to leaving the value unchanged rather than throwing an unhandled exception or executing anything unsafe.

## Verdict

No blocking issues found in this review. `gh pr checks` shows `build`, `Semgrep Scan`, `Semgrep OSS`, and `license/cla` all passing, with only `claude-review` failing due to the known infrastructure issue affecting many open PRs (excluded from consideration per current review policy). `mergeable` is `true` with no reported conflicts (the `blocked` merge state reflects the pending review requirement that this review addresses, not a real conflict). Recommending **approval**.

PR #4851, opened 2026-08-28
Bug #76044, expose the per-organization selected-theme pointer

ngovus-inetsoft:
## What

Adds a `getOrgSelectedTheme(String orgID)` overload to `CustomThemesManager` and its community fallback `CustomThemesImpl`, mirroring the existing `setOrgSelectedTheme(String)` / `setOrgSelectedTheme(String, String)` pair. The enterprise implementation already had the method — it was just `private`.

## Why

"Default for This Organization" is denormalized into three places: the authoritative `portal.org.selectedThemeId.<orgID>` SreeEnv pointer, `FSOrganization.theme`, and the theme's own `organizations` list. Several writers touch only a subset, so the membership list can drift out of agreement with the pointer.

The enterprise theme delete path (`ThemeService.removeTheme()`) currently derives the set of affected organizations from the membership list alone. To close that gap it needs to read the selected-theme pointer of an organization *other* than the one current on the thread — which is what this overload provides.

## Notes

- Community-side change only; no behavior change on its own. The enterprise consumer is in a companion PR (linked below once opened).
- Existing `CustomThemesManagerTest` passes (13/13).

### DIFF
--- a/core/src/main/java/inetsoft/sree/portal/CustomThemesImpl.java
+++ b/core/src/main/java/inetsoft/sree/portal/CustomThemesImpl.java
@@ -46,6 +46,10 @@ public class CustomThemesImpl implements XMLSerializable, AutoCloseable {
       return "default";
    }
 
+   public String getOrgSelectedTheme(String orgID) {
+      return "default";
+   }
+
    public String getGlobalSelectedTheme() {
       return "default";
    }
--- a/core/src/main/java/inetsoft/sree/portal/CustomThemesManager.java
+++ b/core/src/main/java/inetsoft/sree/portal/CustomThemesManager.java
@@ -84,6 +84,14 @@ public class CustomThemesManager implements XMLSerializable, AutoCloseable {
       return impl.getOrgSelectedTheme();
    }
 
+   /**
+    * Gets the selected theme of the specified organization, rather than of the org that is
+    * current on this thread. Returns "default" when the organization has no theme selected.
+    */
+   public String getOrgSelectedTheme(String orgID) {
+      return impl.getOrgSelectedTheme(orgID);
+   }
+
    public void setGlobalSelectedTheme(String selectedTheme) {
       impl.setGlobalSelectedTheme(selectedTheme);
    }


jshobe-inetsoft (approved):
## Code Review

### Summary
Adds a `getOrgSelectedTheme(String orgID)` overload to `CustomThemesManager` (delegating to `impl`) and a matching no-op stub in the community fallback `CustomThemesImpl`. This mirrors the existing `setOrgSelectedTheme(String)` / `setOrgSelectedTheme(String, String)` pair, and unblocks an enterprise-side consumer (`ThemeService.removeTheme()`) that needs to read another organization's selected-theme pointer rather than the current thread's org. The change is community-side only and has no behavioral effect by itself (community's `CustomThemesImpl` always returns `"default"`, same as the existing no-arg overload).

### Completeness & Correctness
- The new overload is consistent with the established pattern in this class (compare to `setOrgSelectedTheme(String)` / `setOrgSelectedTheme(String, String)`), and the delegation in `CustomThemesManager.getOrgSelectedTheme(String)` → `impl.getOrgSelectedTheme(orgID)` is correct and minimal.
- The community stub in `CustomThemesImpl` returns `"default"` unconditionally, matching the behavior of every other stub method in that class (`getSelectedTheme()`, `getOrgSelectedTheme()`, `getGlobalSelectedTheme()`, etc.). This is consistent and correct for the community edition, where no organization-scoped theme selection exists.
- Javadoc on the new manager method clearly documents the semantic difference from the no-arg overload (arbitrary org vs. thread-current org) and the "default" fallback behavior.

### Clean Code / Readability
- Minimal, single-purpose diff (12 lines across 2 files). No unrelated changes bundled in.
- Follows existing naming/style conventions of the surrounding code exactly.

### Error Handling & Edge Cases
- No explicit null-check on `orgID` in the community stub, but this is consistent with the rest of the stub class (none of the other methods validate arguments either), and the method is a pure constant-return stub, so there's no null-dereference risk.
- The real logic (previously `private` in the enterprise `CustomThemesImpl`) is not part of this diff, so its edge-case handling isn't reviewable here; the PR description notes the enterprise consumer is a companion PR.

### Security
- No security concerns. The method is a simple string lookup keyed by an org ID; no new persistence, deserialization, or external I/O introduced in this diff.

### Regression Risk / Backward Compatibility
- Adding a new overload does not change any existing method signature or behavior, so there is no risk to existing callers of `getOrgSelectedTheme()` (still used unchanged in `CustomThemeModel.java:82`).
- Overload resolution is unambiguous (different arity), so no call-site breakage is possible.
- Minor doc-staleness note (non-blocking): two test files — `core/src/test/java/inetsoft/sree/security/OrgLifecycleThemeOrchestrationTest.java` and `core/src/test/java/inetsoft/sree/security/AbstractEditableAuthenticationProviderStaticDepTest.java` — contain header comments explicitly stating "there is no `getOrgSelectedTheme(String orgId)` overload anywhere in the codebase," which becomes stale once this PR merges. Those comments don't assert this at runtime (they're documentation only, and neither test invokes the new overload), so nothing actually breaks, but it would be worth updating that wording in a follow-up to avoid confusing future readers.

### Tests
- Existing `CustomThemesManagerTest` (13/13) continues to pass per the PR description; verified no other test exercises the changed methods in a way that would conflict with the new overload.
- No new dedicated unit test was added for the new overload itself, but this is consistent with the surrounding code: none of the other one-line delegate/stub methods in `CustomThemesManager`/`CustomThemesImpl` (e.g., `getGlobalSelectedTheme`, `setOrgSelectedTheme(String)`) have dedicated tests either. Given the trivial, constant-return nature of the change, this is acceptable.
- No test cases are broken by this change.

### CI / Mergeability
- All required checks pass (`build` / Validation Build, `Semgrep Scan`, `Semgrep OSS`, `license/cla`). The `claude-review` check is failing due to a known infrastructure issue unrelated to this PR's content and is excluded from consideration per current guidance.
- `mergeable: true`, no merge conflicts. The PR's `BLOCKED` merge state is solely due to the branch-protection "review required" gate, which this review satisfies.

### Verdict
**Approve.** The change is small, correct, consistent with existing conventions, backward compatible, and introduces no behavioral change in the community edition. No blocking issues found.

PR #4852, opened 2026-08-28
Bug #76185, drop dead claim-attribute puts from validateSAMLAttributes

ngovus-inetsoft:
### Problem

`SSOSettingsService.validateSAMLAttributes` put nine entries into its settings map — the six real
`onelogin.saml2.*` settings plus `saml.roles.attribute`, `saml.groups.attribute` and
`saml.orgID.attribute` — then built a `Saml2Settings` via `SettingsBuilder.fromValues` and called
`checkSettings()`.

`Saml2Settings` has no member for any of the three claim attributes, and `fromValues` silently drops
keys it does not recognise, so those three puts had no effect on the validation result. Dead code
rather than a defect in behaviour — but it read as though the claim names were validated on save,
which is the opposite of what happens.

### Verification that the puts are inert

`javap` against `java-saml-core-3.0.0.jar`, the version on the compile classpath:

```
javap -classpath java-saml-core-3.0.0.jar com.onelogin.saml2.settings.Saml2Settings
  -> only allowRepeatAttributeName / trimAttributeValues match "attribute"; nothing matches "claim"

javap -classpath java-saml-core-3.0.0.jar -p com.onelogin.saml2.settings.SettingsBuilder
  -> the *_PROPERTY_KEY constants cover onelogin.saml2.* only; no claim/attribute key
```

### Change

Removed the three puts and left a comment recording that nothing validates a claim name, so the code
no longer implies a check it does not perform. The `SreeEnv.setProperty` calls that actually persist
the three properties are untouched.

No behaviour change: `checkSettings()` sees the same six settings and returns an identical error list.
`validateSAMLAttributes` is private with a single caller (`updateSSOSettings`), its `boolean` result is
unchanged for every input, and `settingsMap` is a local built and consumed in the same method.

### Why not validate the claim names instead

A non-empty check would reject working configurations. `SAMLFilter.ClaimNames.current()` reads all
three properties with **no default**, and each use is null-guarded — blank legitimately means "do not
map roles / groups / org". Validating against IdP metadata is not possible at this point either: no
metadata is fetched at save time, the admin hand-enters the certificate and URLs.

Also worth noting for whoever picks up the related ticket: **#76163 (name claim read without a null
guard) appears already fixed** — `SAMLFilter` now has a null guard, a blank-value guard and a
special-character check, each producing a distinct `claimError` recorded in the session audit record.

### Testing

- `./mvnw -pl core compile` — clean.
- `./mvnw -pl core test -Dtest=SSOSettingsControllerTest` — 2/2 pass (delegation-only coverage; it does
  not exercise `validateSAMLAttributes`). No new test added for a dead-code deletion with a provably
  identical result.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/security/SSOSettingsService.java
+++ b/core/src/main/java/inetsoft/web/admin/security/SSOSettingsService.java
@@ -286,9 +286,10 @@ public class SSOSettingsService {
       settingsMap.put("onelogin.saml2.sp.entityid", model.spEntityId());
       settingsMap.put("onelogin.saml2.idp.x509cert", model.idpPublicKey());
       settingsMap.put("onelogin.saml2.idp.single_logout_service.url", model.idpLogoutUrl());
-      settingsMap.put("saml.roles.attribute", model.roleClaim());
-      settingsMap.put("saml.groups.attribute", model.groupClaim());
-      settingsMap.put("saml.orgID.attribute", model.orgIDClaim());
+      //the saml.*.attribute claim names are StyleBI's own and are not part of the java-saml
+      //settings vocabulary, so they cannot be checked here -- fromValues drops keys it does
+      //not recognise. Nothing validates a claim name anywhere; a wrong one saves cleanly and
+      //silently maps nothing at login (see SAMLFilter.ClaimNames).
       final Saml2Settings settings = new SettingsBuilder().fromValues(settingsMap).build();
       final List<String> errors = settings.checkSettings();
       final boolean valid = errors.isEmpty();


jshobe-inetsoft (approved):
## Code Review — PR #4852: Drop dead claim-attribute puts from `validateSAMLAttributes`

### Summary of change
`SSOSettingsService.validateSAMLAttributes` previously put `saml.roles.attribute`, `saml.groups.attribute`, and `saml.orgID.attribute` into the map passed to `SettingsBuilder.fromValues()` before building a `Saml2Settings` and calling `checkSettings()`. This PR removes those three `put` calls and replaces them with a comment explaining why they were dead code, without changing any other line.

### Correctness verification
- I confirmed via the diff and full file (`core/src/main/java/inetsoft/web/admin/security/SSOSettingsService.java`) that `validateSAMLAttributes` is `private` with exactly one call site, `updateSSOSettings` (line 197).
- `Saml2Settings`/`SettingsBuilder` (java-saml-core) only recognize the `onelogin.saml2.*` keys; the three removed keys (`saml.roles.attribute`, `saml.groups.attribute`, `saml.orgID.attribute`) are StyleBI-internal property names that `fromValues` silently ignores, and there is no `checkSettings()` behavior tied to them. The PR description's `javap` evidence is consistent with this.
- The `SreeEnv.setProperty` calls that actually persist these three claim values (lines 204-206 in `updateSSOSettings`) are untouched — only the transient, ineffective copy into the validation map was removed. Persistence behavior for role/group/org claims is unchanged.
- Since the removed keys were never read by `Saml2Settings.checkSettings()`, the returned `errors` list — and therefore the method's `boolean` return value — is provably identical for every input before and after this change. This is a true no-op removal of dead code, not a behavior change.

### Clean code / readability
- The added comment is clear and explains both *why* the puts were removed and *where* the actual validation gap lives (pointing to `SAMLFilter.ClaimNames`), which is good for future maintainers who might otherwise re-add the same dead code.
- Change is minimal and scoped exactly to the defect described.

- No new error handling paths introduced. Existing null-guard behavior for blank claim names in `SAMLFilter.ClaimNames.current()` is unaffected by this change, consistent with the PR description.

### Security
- No security-relevant behavior change. This does not weaken or alter any validation that was previously effective — the removed puts never affected the validation result to begin with.

### Regression risk
- Very low. The change removes lines whose effect was already provably absorbed/discarded by `fromValues`. No other code path reads `settingsMap` or depends on side effects of these `put` calls.

### Backward compatibility
- No API, config, or persisted-property changes. `saml.roles.attribute`, `saml.groups.attribute`, and `saml.orgID.attribute` continue to be saved and read exactly as before.

### Tests
- No test exercises `validateSAMLAttributes` directly (`SSOSettingsControllerTest` only covers controller delegation, per the PR description, and remains passing). Given this is a provably-equivalent dead code removal, the absence of a new test is reasonable and doesn't indicate a broken test case. I did not find any test that references `validateSAMLAttributes`, `saml.roles.attribute`, `saml.groups.attribute`, or `saml.orgID.attribute` in a way this change would break.

### CI / mergeability
- All checks pass except `Claude Code Review / claude-review`, which is known to be broken/erroring on infrastructure grounds across many open PRs in this repo and is excluded from consideration per review policy.

### Verdict
No blocking issues found. This is a safe, well-justified, minimal dead-code removal with clear reasoning and verified equivalence. **Approving.**

PR #4853, opened 2026-08-28
Bug #75965, underline table cells that have a row-level hyperlink

ngovus-inetsoft:
## Problem

Opening viewsheet `hyperlink1` in the portal and hovering cell `12001` in the `Order` table shows the hyperlink tooltip (`tip`) but the cell text is **not underlined**.

This looked intermittent but is deterministic. In the reported asset, `<tableHyperlinkAttr>` is empty — the link is an **assembly-level row hyperlink**:

```xml
<Hyperlink ApplyToRow="true" Link="www.baidu.com" ToolTip="tip" LinkType="1">
```

Adding a **column** hyperlink underlines correctly, because that is a different code path. Only row hyperlinks were affected.

## Root cause

Row hyperlinks are sent to the client in a separate array, never attached to the cell model:

- `BaseTableService.getRowHyperlinks()` builds `LoadTableDataCommand.rowHyperlinks`, one entry per loaded row (null for header rows).
- `BaseTableCellModel.getHyperlinks()` only consults `VSTableLens.getCellHyperlink()`, which resolves column/cell and auto-drill links — it never sees `rowHyperlink`. So `cell.hyperlinks` is `null` for these cells.

On the client, the underline was keyed solely off `cell.hyperlinks`:

```ts
public isHyperlink(cell: BaseTableCellModel): boolean {
   return cell.hyperlinks != null && cell.hyperlinks.length > 0 && !this.model.form && cell.underline;
}
```

The row-hyperlink-aware members already existed and worked — they just did not affect styling: `getTooltip(cell, i)` (hence the visible "tip"), `rowHoverable()`, `rowLinkClicked()`, `linkClicked()`.

This has never worked in the Angular viewer; it is not a recent regression. The export path is the behavioral precedent and is already correct: `VSTableDataHelper.getHyperLink()` returns the row link for **every** column of a detail row and `HTMLTableHelper` writes `text-decoration:underline` for it. The viewer was the odd one out.

## Change

Frontend only, two files.

**`vs-table.component.ts`** — new `hasRowHyperlink(row)` helper; `isHyperlink()` takes an optional row index and ORs in the row hyperlink:

```ts
public hasRowHyperlink(row: number): boolean {
   return row > -1 && !!this.rowHyperlinks && !!this.rowHyperlinks[row];
}

public isHyperlink(cell: BaseTableCellModel, row: number = -1): boolean {
   return (cell.hyperlinks != null && cell.hyperlinks.length > 0 || this.hasRowHyperlink(row)) &&
      !this.model.form && cell.underline;
}
```

`row` is the `tableData` batch index, which is exactly how `rowHyperlinks` is indexed (both are assigned together in `loadTableData`), matching the existing `getTooltip(cell, i)` / `linkClicked($event, false, i)` convention.

**`vs-table.component.html`** — detail cell loop only: `[class.hyperlink]="isHyperlink(_cell, i)"`. The header binding is unchanged, since the server already stores `null` in `rowHyperlinks` for header rows and headers render from a separate `tableHeaders` table.

### Deliberately not changed

- **Java side** — folding the row link into `cell.hyperlinks` would break `rowLinkClicked()`, which deliberately distinguishes cell auto-drill links from the row link, and would push row links into the click dropdown menu.
- **`vs-table-cell.component.ts` (`isLinkedCell`)** — the `.hyperlink` class on the `<td>` inherits `text-decoration` into the cell text, so no new `@Input` on the shared cell component is needed. `vs-table-cell` is shared with crosstab and calc table.
- **`vs-crosstab` / `vs-calctable`** — `BaseTableService.getRowHyperlinks()` returns an empty array unless the assembly is a `TableVSAssembly` (and not embedded), so crosstabs and calc tables never receive row hyperlinks.
- **Form / embedded tables** — unaffected: excluded server-side via `!info.getFormValue()` and on the client by the retained `!this.model.form` guard.
- The `cell.underline` gate is retained, so the `hyperlink.indicator` property still suppresses the indicator.

## Testing

Added `Group 14 — isHyperlink / hasRowHyperlink` to `vs-table.component.interaction.tl.spec.ts` (7 cases): row link underlines the detail cell; header cells (no row index) do not; a row with no `rowHyperlinks` entry does not; `underline: false` does not; form tables do not; the existing cell-hyperlink case is unchanged; `hasRowHyperlink` index bounds.

All 7 pass. Reverting only the `isHyperlink` change makes **exactly one** fail — the bug-reproducing case — confirming the tests are not vacuous.

Also verified: `tsc --noEmit` on the spec-tl project is clean, ESLint on the changed files is clean, and the standard portal suite is unaffected.

Manually confirmed in a running server with the reported asset: detail cells are underlined, the `tip` tooltip still appears, and clicking a row still opens the link.

> [!NOTE]
> `vs-table.component.interaction.tl.spec.ts` is currently one of **26 portal TL spec files that fail at import on clean `main`** (`TypeError: Class extends value undefined` from `sort-column-event.ts:23`, a module-init/chunking issue in the `@angular/build:unit-test` builder). Confirmed pre-existing by stashing these changes and re-running. The new tests were therefore verified through the standard runner via a temporary copy. This is consistent with `CLAUDE.md` noting `test:portal:tl` is not yet wired to CI, and is tracked separately from this fix.

### DIFF
--- a/web/projects/portal/src/app/vsobjects/objects/table/vs-table.component.html
+++ b/web/projects/portal/src/app/vsobjects/objects/table/vs-table.component.html
@@ -264,7 +264,7 @@
                       [wTooltip]="getTooltip(_cell, i)"
                       [cellAnnotations]="model.leftBottomAnnotations"
                       [class.selected-table-cell]="isSelected(_cell)"
-                      [class.hyperlink]="isHyperlink(_cell)"
+                      [class.hyperlink]="isHyperlink(_cell, i)"
                       [style.min-width.px]="tableCellDisplayWidth(j)"
                       [style.max-width.px]="tableCellDisplayWidth(j)"
                       [style.color]="_cell.vsFormatModel?.foreground"
--- a/web/projects/portal/src/app/vsobjects/objects/table/vs-table.component.interaction.tl.spec.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/table/vs-table.component.interaction.tl.spec.ts
@@ -39,6 +39,7 @@
  *   Group 11 — isDraggable: header→true; non-header+viewer→false; resizeCol active→false
  *   Group 12 — clearFlyover: hasFlyover+force→sendFlyover+clears; no-op when condition fails
  *   Group 13 — onOpenHighlightDialog: 'table highlight' action emits model
+ *   Group 14 — isHyperlink/hasRowHyperlink: cell links, row links, form + indicator gates
  */
 
 import { Subject } from "rxjs";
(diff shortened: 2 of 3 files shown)


jshobe-inetsoft (approved):
## Code Review — Bug #75965: underline table cells with a row-level hyperlink

### Summary
Small, well-scoped frontend fix. `VSTable.isHyperlink()` previously only checked `cell.hyperlinks`, so cells whose only link was an assembly-level **row** hyperlink (`rowHyperlinks[row]`, sent separately from the cell model) never got the `.hyperlink` CSS class (`text-decoration: underline`), even though the tooltip and click-through already worked for row links. The fix adds a `hasRowHyperlink(row)` helper and ORs it into `isHyperlink()`, with a new optional `row` parameter (default `-1`, preserving old behavior when omitted).

### Correctness
- The `row` index passed from the template (`isHyperlink(_cell, i)`) is the `tableData` loop index, which is exactly how `rowHyperlinks` is populated and already consumed elsewhere in this same class (`getTooltip(cell, row)`, `linkClicked($event, false, i)`, `rowLinkClicked`) — verified in `loadTableData()`, where `this.tableData` and `this.rowHyperlinks` are both assigned from the same `LoadTableDataCommand`. The new code is consistent with this established convention.
- The header-cell binding (`isHyperlink(_cell)`, no row arg) is correctly left unchanged; `hasRowHyperlink(-1)` returns `false` via the `row > -1` guard, so headers are unaffected — matches the fact the server sends `null` for header rows in `rowHyperlinks`.
- The `.hyperlink` CSS class (`web/projects/portal/src/scss/internal/_layout.scss`) is a plain `text-decoration: underline` on the `<td>`, confirmed to be inherited by cell text, so no change to the shared `vs-table-cell` component was needed, as the PR claims.
- Precedence logic is preserved: if a cell has its own `cell.hyperlinks` (auto-drill), the first OR branch already makes it underlined regardless of row link state, matching `rowLinkClicked()`'s existing comment that "cell hyperlinks and row hyperlinks cannot exist at the same time" (cell-level auto-drill takes priority over the row link on click).
- `cell.underline` and `!this.model.form` gates are retained unchanged, so the hyperlink-indicator toggle and form-table exclusion still work as before.
- Scope is appropriately narrow: crosstab/calc-table are untouched, correctly per the PR's explanation that `BaseTableService.getRowHyperlinks()` only populates for `TableVSAssembly` (non-embedded), so this code path is unreachable for those components.

### Tests
- 7 new cases added (`Group 14`) covering: cell-level link, row-level link with row index, header (no row index) suppressing it, missing/null `rowHyperlinks` entry, `underline:false` gate (both cell- and row-link cases), form table suppression, and `hasRowHyperlink` boundary conditions (`-1`, out-of-range index).
- Test helpers (`createTableComponent`, `makeTableCell`) are pre-existing and used correctly (e.g. `overrides.model` merges into `makeMockTableModel` defaults, so `{ model: { form: true } }` works as expected).
- PR states reverting only the `isHyperlink` change fails exactly the bug-reproducing test, which is good evidence the tests are meaningful rather than vacuous.
- The `.tl.spec.ts` pre-existing import failure on `main` (`sort-column-event.ts` `Class extends value undefined`) is called out as pre-existing/unrelated and not introduced by this change; consistent with `CLAUDE.md`'s note that `test:portal:tl` isn't wired into CI yet. This is a known, tracked-separately infra issue, not a regression from this PR.

### Style / maintainability
- Minimal diff (net +5 lines of production TS, 1-line template change), consistent with the existing code's style (matches `getTooltip`'s pattern of an optional trailing `row` parameter defaulting to `-1`).
- New `hasRowHyperlink` method is a clean, testable extraction rather than inlining the row-hyperlink check into `isHyperlink`, improving readability.

### Security
- No new user input handling, no injection surface — this only changes CSS-class boolean logic on the client based on data the server already sends and the client already renders (tooltip/click). No security concerns.

- The new `row` parameter is optional with a safe default (`-1`), so any other caller of `isHyperlink(cell)` (there are none outside this file/template) is unaffected.
- No server-side/API changes; no persisted-data format changes.
- No regression risk identified for crosstab/calc-table (separate `isHyperlink` overrides untouched) or for form/embedded tables (explicit guards retained).

### Verdict
No blocking issues found. The fix is correct, minimal, consistent with existing conventions in the file, and well tested.

---

**Merge-readiness checklist:**
- ✅ No blocking issues in this review.
- ✅ No failing checks other than `claude-review` (Semgrep OSS, Semgrep Scan, `build`, and `license/cla` all pass; `claude-review` is currently broken infrastructure-wide per project guidance and is excluded from this evaluation).
- ✅ No merge conflicts (`mergeable: MERGEABLE`). The `BLOCKED` merge state is solely due to the repository's required-review branch protection, not a conflict.

**Recommendation: Approve.**

PR #4855, opened 2026-08-28
Bug #76069, reserve the built-in theme ID so a theme named "default" applies

ngovus-inetsoft:
## What

Reserves `"default"` as a theme **ID** so it can never be assigned to a custom theme. The theme **name** `default` remains perfectly legal — it just gets a different ID (`default1`).

Adds `CustomTheme.DEFAULT_THEME_ID` and `CustomTheme.isReservedId()` as the single definition of the sentinel, and teaches `AddThemeDialogComponent.createId()` to treat it as taken.

## Why

`"default"` is the ID of the built-in theme and simultaneously the sentinel written to the selected-theme pointers (`portal.org.selectedThemeId[.<orgID>]`, `Organization.theme`) to mean "no custom theme selected".

A new theme's ID is derived from its name — `createId()` returns the typed name verbatim unless it collides with an ID already in use. The collision list comes from `GET /api/em/settings/presentation/themes/ids`, which reports only *existing custom themes*. The reserved sentinel is not among them, so the name `default` passed straight through and the theme was persisted with `id == "default"`.

From there every consumer of the pointers read it as "nothing selected":

- `CustomThemesImpl.getSelectedTheme()` discarded the pointer and returned `"default"`.
- `isCustomThemeApplied()` returned `false`, so `theme-variables.css` was never even `<link>`ed into the portal.
- `isEMDarkTheme()` and `getScriptThemeCssPath()` short-circuited before consulting the theme.

Meanwhile `CustomThemeModel.applyDefaultFlags()` compares the theme's ID against `getOrgSelectedTheme()`/`getGlobalSelectedTheme()`, **both of which return the literal `"default"` when unset** — so EM reported both default checkboxes ticked while the portal rendered the built-in theme. That is the exact EM-vs-portal disagreement in the bug report.

Worth noting the resource layer would have worked fine: `GlobalStyleController` and `ThemeProtocolResolver` both match a custom theme with that ID. The break was entirely in selected-ID resolution above it, so the JAR was never requested.

## Approach

Reserve the ID rather than reject the name — no user-facing error and no new i18n string. The server reserves it too (companion enterprise PR), but checking it client-side as well means a stale or failed `/themes/ids` response cannot reintroduce the collision.

Both branches of `createId()` are covered: the fast path and the `base + counter` loop, which would otherwise have resolved straight back to `default`.

## Notes

- Community half of the fix. The enterprise PR reserves the ID server-side (`getThemesIds()`, `addTheme()`, the `updateTheme()` rename path) and depends on `DEFAULT_THEME_ID` landing here first — link below.
- **No migration** for themes already stored with `id == "default"`. They remain listed and deletable in EM, so delete-and-recreate is the recovery path.
- New `add-theme-dialog.component.spec.ts` (the component had no spec): 4 tests covering the ordinary path, the reserved name, the reserved name when the server *does* report it, and an ordinary collision.
- Full EM suite passes: 104 files / 371 tests.

### DIFF
--- a/core/src/main/java/inetsoft/sree/portal/CustomTheme.java
+++ b/core/src/main/java/inetsoft/sree/portal/CustomTheme.java
@@ -30,6 +30,30 @@ import java.util.*;
  */
 public final class CustomTheme implements XMLSerializable, Cloneable {
    /**
+    * The reserved theme identifier that denotes the built-in theme. It doubles as the sentinel
+    * written to the selected-theme pointers ({@code portal.org.selectedThemeId[.<orgID>]} and
+    * {@code Organization.theme}) to mean "no custom theme selected", so it must never be the ID
+    * of an installed custom theme: such a theme is stored and listed, but every consumer of the
+    * pointers reads it as "nothing selected" and the built-in theme is served instead.
+    *
+    * <p>Only the <em>ID</em> is reserved. A theme may still be <em>named</em> "default"; it is
+    * assigned a different ID.
+    */
+   public static final String DEFAULT_THEME_ID = "default";
+
+   /**
+    * Determines whether a theme identifier is reserved and therefore unavailable to a custom
+    * theme.
+    *
+    * @param id the theme identifier to test.
+    *
+    * @return {@code true} if the identifier is reserved.
+    */
+   public static boolean isReservedId(String id) {
+      return DEFAULT_THEME_ID.equals(id);
+   }
+
+   /**
     * Gets the display name of the theme.
     *
     * @return the theme name.
--- a/web/projects/em/src/app/settings/presentation/presentation-themes-view/add-theme-dialog/add-theme-dialog.component.spec.ts
+++ b/web/projects/em/src/app/settings/presentation/presentation-themes-view/add-theme-dialog/add-theme-dialog.component.spec.ts
@@ -1,0 +1,95 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import { HttpClientTestingModule, HttpTestingController } from "@angular/common/http/testing";
+import { NO_ERRORS_SCHEMA } from "@angular/core";
+import { ComponentFixture, TestBed } from "@angular/core/testing";
+import { MAT_DIALOG_DATA, MatDialogRef } from "@angular/material/dialog";
+import { NoopAnimationsModule } from "@angular/platform-browser/animations";
+import { AddThemeDialogComponent } from "./add-theme-dialog.component";
+
+describe("AddThemeDialogComponent", () => {
+   let dialogRef: any;
+
+   async function createComponent(data: any): Promise<ComponentFixture<AddThemeDialogComponent>> {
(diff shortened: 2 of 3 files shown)


ngovus-inetsoft:
Companion enterprise PR: https://github.com/inetsoft-technology/stylebi-enterprise/pull/720

jshobe-inetsoft (approved):
## Code Review

### Summary
Reserves `"default"` as a theme **ID** (not name) so a custom theme named "default" can no longer be silently assigned the same ID as the built-in theme's sentinel. This is the community half of a two-repo fix (bug #76069); the enterprise repo enforces the same reservation server-side.

### Correctness
- `CustomTheme.DEFAULT_THEME_ID` / `isReservedId()` are a clean, well-documented single source of truth for the sentinel value, placed in the shared model class (`core/.../CustomTheme.java`) where both community and the future enterprise consumer can use it.
- `AddThemeDialogComponent.createId()` correctly guards **both** branches:
  - The fast path (`nameValue` used verbatim) now also rejects `"default"`.
  - The `base + counter` loop condition now also treats `newId === RESERVED_ID` as a collision, so it can't converge back onto `"default"` (e.g., name `"default"` with an empty existing-ids list now yields `"default1"` instead of `"default"`).
- Verified by hand-tracing all three test scenarios (fresh name, reserved name with empty ids, reserved name confirmed by server) against the loop/branch logic — all match expected output.
- Confirmed via `grep` that no other client-side code path derives a theme ID from a name (only this one `createId()`), and that the theme-management REST controllers live entirely in the enterprise repo, so no further community-side call sites were missed.

### Clean Code / Readability
- Good, thorough Javadoc/comments explaining *why* (sentinel semantics), not just *what*.
- Minor nit: `AddThemeDialogComponent.RESERVED_ID` duplicates the Java `CustomTheme.DEFAULT_THEME_ID` as a separately-maintained string literal (unavoidable across the Java/TypeScript boundary, and it's commented as mirroring the Java constant) — acceptable but worth a shared-constants mechanism (e.g. generated from server) if this pattern grows.
- Style inconsistency: the new `newId === RESERVED_ID` check uses strict equality while the adjacent pre-existing `ids.find(s => s == newId)` uses loose equality. Harmless (both operands are strings) but slightly inconsistent within the same expression.

- Handles the case where the server's `/themes/ids` response is stale or failed to include `"default"` — the client-side check is a deliberate defense-in-depth measure per the PR description.
- Handles the "name is entirely non-alphanumeric" fallback to UUID generation without interference from the new check (UUID can't collide with `"default"`).
- No regression to the existing duplicate-name / counter-suffix logic for genuinely colliding names.

### Security
- No security concerns. This is a client-side UX/data-integrity fix; the authoritative enforcement is intended to live server-side (companion enterprise PR), which is the correct place for real security enforcement. No injection, XSS, or auth-related surface touched.

- No migration for themes already persisted with `id == "default"` — explicitly called out in the PR description as an accepted limitation (existing broken themes must be deleted and recreated). This is a reasonable scope boundary for a targeted bug fix, not a regression.
- The theme **name** `"default"` remains legal, preserving existing user-facing behavior/expectations; only the derived ID changes, which is invisible to users under normal use.
- No existing behavior for non-colliding names changes.

### Tests
- New `add-theme-dialog.component.spec.ts` (component previously had no spec) adds 4 targeted tests covering the fast path, the reserved-name path, the reserved-name-when-server-reports-it path, and an ordinary collision. All pass in CI (both `test:em` runs list `add-theme-dialog.component.spec.ts (4 tests)` passing).
- Confirmed via the CI "build" job log that the full frontend suite passes (`test:portal`: 191 files, `test:em`: 104 files including the new spec, `test:em:tl`: 70 files), and the overall Maven build ends in `BUILD SUCCESS`. The stray `process.emit is not a function` / jsdom stderr noise in the `test:em:tl` run is pre-existing test-harness noise unrelated to this change (unrelated `edit-identity-view`/`add-parameter-dialog` specs) and does not fail any test.
- No existing test was broken by this change.

### CI / Mergeability
- Checks: `build` passes, `Semgrep OSS` passes, `Semgrep Scan` passes, `license/cla` passes. Only `Claude Code Review / claude-review` fails, which per current instructions is a known-broken infrastructure check and is disregarded here.
- `mergeable: true`, `mergeable_state: blocked` — the "blocked" state is due to the required-review branch protection rule (no reviews yet), not a merge conflict. There are no actual conflicts preventing merge.

### Verdict
No blocking issues found. The fix is minimal, well-scoped, well-tested, and well-documented. **Approve.**

PR #4858, opened 2026-08-28
Bug #76038, reposition tab children when bottomTabs toggles in mobile preview

jkidacki-inetsoft:
## Summary

In a mobile/device-layout preview, a Tab assembly's `bottomTabs` property toggled at runtime via script (e.g. `Tab1.bottomTabs = RadioButton1.selectedObject`) moved the tab bar to its new position immediately, but left the tab's children at their old scaled position. With a single tab child this was invisible; with multiple children of differing heights, a child not defining the tab's new anchor (e.g. a short Text object) stayed pinned near its old position until a full page refresh recomputed everything.

## Root Cause

`TabVSAssemblyInfo.repositionForBottomTabsInScaledSpace()` (called from `TabVSAScriptable.setBottomTabs()` to give immediate visual feedback in the scaled/mobile-preview coordinate space) repositioned only the tab bar's own `scaledPosition`. Its master-pixel-space sibling, `repositionForBottomTabs()`, already has a per-child reposition loop for the equivalent (non-scaled) case, but the scaled-space version never got one — its doc comment even said "Tab moves; children stay," which matches the reported symptom exactly.

## Fix

Added a loop to `repositionForBottomTabsInScaledSpace()` that re-flushes every child's scaled position against the tab bar's new position, mirroring the existing per-child loop in `repositionForBottomTabs()` and reusing the same `getBottomTabChildHeight()` helper for dropdown/labeled-input height handling.

## Test Plan

- Added `repositionForBottomTabsInScaledSpaceReflowsShorterChildWithTabBar` — a multi-child scenario (tall + short child) verifying the short child now follows the tab bar to its new scaled Y instead of staying at its old position. Hand-verified the arithmetic and confirmed via code review that this test fails without the fix.
- Ran `TabVSAScriptableTest` (12/12 pass) — exercises the sole production call site (`TabVSAScriptable.setBottomTabs()`) and its master-pixel-space multi-child analog, confirming no regression to the existing (unscaled) behavior.
- `mvn clean install -pl core -am -DskipTests` builds clean.
- User confirmed the fix resolves the reported issue against the attached test asset.

Fixes Redmine #76038.

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/TabVSAssemblyInfo.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/TabVSAssemblyInfo.java
@@ -423,8 +423,10 @@ public class TabVSAssemblyInfo extends ContainerVSAssemblyInfo {
 
    /**
     * Scaled-space counterpart of {@link #repositionForBottomTabs}. Lets a
-    * runtime script toggle take effect immediately in layout previews. Tab
-    * moves; children stay. No-op in master view (no scaledPosition).
+    * runtime script toggle take effect immediately in layout previews. The
+    * tab bar re-anchors to the children's extent, and children are then
+    * re-flushed against the tab bar's new position -- mirroring the
+    * master-pixel-space loop below. No-op in master view (no scaledPosition).
     */
    public static void repositionForBottomTabsInScaledSpace(TabVSAssemblyInfo tabInfo,
                                                            Viewsheet vs,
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
## Code Review Summary

**Verdict: Approve**

This PR fixes Bug #76038: when a Tab assembly's `bottomTabs` is toggled at runtime via script in a mobile/device-layout preview, `TabVSAssemblyInfo.repositionForBottomTabsInScaledSpace()` moved only the tab bar's own `scaledPosition`, leaving children behind at their old scaled position (visible with multi-child tabs of differing heights). The fix adds a per-child re-flush loop mirroring the equivalent logic already present in the master-pixel-space sibling method `repositionForBottomTabs()`.

### Correctness

- The new loop's arithmetic exactly mirrors the existing, already-shipped pixel-space loop in `repositionForBottomTabs()` (same `toBottomTabs ? newTabY - childHeight : newTabY + tabHeight` formula), and reuses the shared `getBottomTabChildHeight()` helper for dropdown/labeled-input height handling, so the two coordinate spaces (pixel vs. scaled) now stay consistent.
- Null/edge-case guards are present and match the sibling method's style: missing assembly (`child == null`), missing scaled position/size (`childScaledPos == null || childScaledSize == null`), and zero effective height (`childHeight == 0`) are all skipped via `continue`.
- The anchor child (the one that defines the tab bar's new extent) is correctly left untouched, since its computed `newChildY` already equals its current position — verified both by manual arithmetic trace and by the added test's assertion (`tallChildInfo` position unchanged at 24).
- I manually traced the added test's arithmetic (tall child bottom = 24+200 = 224 → tab moves to 224 → short child new y = 224-40 = 184) and confirmed it's correct and does exercise the exact bug scenario described (a child whose height doesn't define the tab bar's anchor position).
- I independently confirmed the sole production call site is `TabVSAScriptable.setBottomTabs()`, which calls both the pixel-space and scaled-space repositioning methods back-to-back — consistent with the PR's stated test plan.

- No public API signatures changed; the change is additive within an existing method body.
- The early-return guards (no children, no `vs`, tab bar has no explicit `scaledPosition`, i.e. master view) are unchanged, so the no-op behavior for master view and for viewsheets that never entered scaled/mobile-preview layout is preserved.
- I don't see any other call site that could be newly affected by this change; behavior for a single-child tab remains a no-op as before (the anchor child already flush).

1. **Doc comment direction is off.** The updated Javadoc on `repositionForBottomTabsInScaledSpace()` says the new loop mirrors "the master-pixel-space loop **below**," but `repositionForBottomTabs()` (with the analogous loop) is defined **above** it in the file, not below. Trivial, but worth a fixup for accuracy.
2. **New regression test isn't tagged for CI.** `TabVSAssemblyInfoTest` (pre-existing file, not introduced by this PR) lacks `@Tag("core")`, and the core module's Surefire configuration restricts the "Validation Build" test run to classes/methods tagged `core` (`<groups>core</groups>`). I confirmed via the CI log for this PR's own Validation Build run that `TabVSAssemblyInfoTest` never executes there (unlike e.g. `TabVSAScriptableTest`, which is tagged `core` and does run, matching the 12/12 pass claimed in the PR description). This is a pre-existing gap shared by a few sibling classes in the same package (`CalendarUtilTest`, `LabelInfoTest`, `StandardPeriodsTest`, `VSUtilZIndexTest`) and not something this PR needs to fix, but since the whole point of the new test is to guard against a regression of this exact bug, it would be worth adding `@Tag("core")` to `TabVSAssemblyInfoTest` (as done for its sibling `VSAssemblyInfoTest` in the same directory) so it actually runs in CI going forward.

Neither point above blocks merging.

### Checks / mergeability

- All required checks pass except `Claude Code Review / claude-review`, which is known to be broken/erroring on infrastructure grounds across many open PRs in this repo (excluded from consideration per review policy). `Semgrep Scan`, `Semgrep OSS`, `Validation Build`, and `license/cla` are all green.
- `mergeable: MERGEABLE`, no merge conflicts. The `BLOCKED` merge state is solely due to the pending required review (`reviewDecision: REVIEW_REQUIRED`), which this review addresses.

No blocking issues found. Recommending approval; the two notes above are optional follow-ups.

PR #4859, opened 2026-08-28
Bug #75898, fix backup pruning off-by-one, comparator overflow, and retention count

jkidacki-inetsoft:
## Summary

`DataSpaceSettingsService.deleteRedundantBackupFiles()` prunes operator backup zip
files in `backup/` down to `asset.backup.count` before each new backup is written.
It had two defects, both affecting real backups:

1. **Off-by-one**: the prune loop ran one extra iteration (`i <= deleteCount`
   instead of `i < deleteCount`), deleting one file more than intended.
2. **Comparator overflow**: the sort comparator narrowed a `long` timestamp
   difference to `int`. When two backups were far enough apart, the difference
   overflowed `int` and could flip sign, inverting the sort order — deleting the
   *newest* backup instead of the oldest, or throwing
   `IllegalArgumentException: Comparison method violates its general contract!`.

## Root Cause

See above. Additionally, fixing just those two lines exposed a design gap: with
the corrected loop, `doBackup()`'s single prune-before-write call converges on
**one more** backup than configured in steady-state operation (prune-to-N, then
+1 for the new file, never trimmed back down) — `asset.backup.count = N` would
settle at `N+1` files, not `N`.

## Fix

- Loop bound: `i <= deleteCount` → `i < deleteCount`.
- Comparator: `(int) (z1Time - z2Time)` → `Long.compare(z1Time, z2Time)`.
- `doBackup()` now calls `deleteRedundantBackupFiles()` a second time after a
  successful write (in addition to the existing call before the write), so the
  count converges on exactly `N`. The pre-write call only clears genuine
  pre-existing surplus (a no-op when already at or under the limit), so a failed
  write never deletes more than that pre-existing surplus.
- `deleteRedundantBackupFiles()` visibility relaxed from `private` to
  package-private so it's directly unit-testable; it has no other callers.

**Note:** the off-by-one fix changes effective retention behavior on existing
deployments (previously under-retaining by one file with the old bug's
particular sequencing; now converges on exactly the configured count). May
warrant a release note.

## Test Plan

- New `DataSpaceSettingsServiceTest` (6 cases): retention-limit boundary, surplus
  deletion order, comparator overflow correctness, disabled-when-unconfigured,
  and two `doBackup()` sequencing tests verifying the second prune runs only
  after a successful write and never after a failed one.
- `mvnw clean test -pl core -Dtest=DataSpaceSettingsServiceTest`: all 6 pass.
- Verified via manual arithmetic/simulation of repeated `doBackup()` cycles that
  the new two-call design converges on exactly `asset.backup.count` files,
  versus `count + 1` for a single-call fix and an under-count (with risk of
  reaching zero at `count = 1` on a failed write) for the original code.

Fixes Redmine #75898.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/general/DataSpaceSettingsService.java
+++ b/core/src/main/java/inetsoft/web/admin/general/DataSpaceSettingsService.java
@@ -97,6 +97,9 @@ public class DataSpaceSettingsService extends BackupSupport {
          ActionRecord.ACTION_STATUS_FAILURE, "");
 
       try {
+         // clear any pre-existing surplus first; this does not touch the current backup
+         // count when it is already at or under the limit, so a failure below never loses
+         // more backups than were already surplus
          deleteRedundantBackupFiles();
 
          // For the same backup, use the same timestamp
@@ -109,6 +112,10 @@ public class DataSpaceSettingsService extends BackupSupport {
 
          String path = getBackFile(model != null ? model.dataspace() : null, stamp);
          this.externalStorageService.write(path, file.toPath(), null);
+
+         // the new backup is safely on disk, so it is now safe to trim the oldest file
+         // back down to the configured count
+         deleteRedundantBackupFiles();
 
          status = catalog.getString("Success");
          record.setActionStatus(ActionRecord.ACTION_STATUS_SUCCESS);
@@ -132,9 +139,12 @@ public class DataSpaceSettingsService extends BackupSupport {
    }
 
    /**
-    * backup count control by property "asset.backup.count",
+    * Deletes the oldest backup files down to the count of "asset.backup.count", if surplus
+    * files exist. Called both before a new backup is written, to clear any pre-existing
+    * surplus, and again after a successful write, to trim the file just added back down to
+    * the configured count.
     */
-   private void deleteRedundantBackupFiles() {
+   void deleteRedundantBackupFiles() {
       String backupCountProp = SreeEnv.getProperty("asset.backup.count");
       int backupCount = -1;
 
@@ -154,7 +164,7 @@ public class DataSpaceSettingsService extends BackupSupport {
             long z1Time = getTimestamp(z1);
             long z2Time = getTimestamp(z2);
 
-            return (int) (z1Time - z2Time);
+            return Long.compare(z1Time, z2Time);
          })
          .toList();
 
@@ -165,7 +175,7 @@ public class DataSpaceSettingsService extends BackupSupport {
 
       int deleteCount = zips.size() - backupCount;
 
-      for(int i = 0; i <= deleteCount; i++) {
+      for(int i = 0; i < deleteCount; i++) {
          try {
             this.externalStorageService.delete(BACKUP_FOLDER + File.separator + zips.get(i));
          }
--- a/core/src/test/java/inetsoft/web/admin/general/DataSpaceSettingsServiceTest.java
+++ b/core/src/test/java/inetsoft/web/admin/general/DataSpaceSettingsServiceTest.java
@@ -1,0 +1,207 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (approved):
## Code Review

Reviewed `core/src/main/java/inetsoft/web/admin/general/DataSpaceSettingsService.java` and the new `DataSpaceSettingsServiceTest.java`.

### Correctness of the two core defects
- **Off-by-one**: `for(int i = 0; i <= deleteCount; i++)` deleted `deleteCount + 1` files. Changing to `i < deleteCount` correctly deletes exactly `zips.size() - backupCount` files, leaving `backupCount` remaining. Confirmed by re-deriving the arithmetic against the test fixtures.
- **Comparator overflow**: `(int) (z1Time - z2Time)` truncated a `long` difference between two `yyyyMMddHHmmss`-style timestamps. For backups more than ~2.15 billion units apart (well within a few years for this timestamp format), the cast can overflow `int` and flip sign, producing an invalid/inconsistent ordering (and, per Java's `TimSort`, a possible `IllegalArgumentException: Comparison method violates its general contract!`). Replacing with `Long.compare(z1Time, z2Time)` is the textbook fix and removes the overflow risk entirely.

Both fixes are minimal, targeted, and verified by the new `retainsAllFilesWhenAtTheLimit`, `deletesOnlyTheOldestSurplusFiles`, and `deletesTheOldestEvenWhenTimestampsAreYearsApart` tests, which reproduce each defect and assert the corrected behavior.

### Second call to `deleteRedundantBackupFiles()` in `doBackup()`
The reasoning in the PR description is sound: with only the pre-write prune call, a correctly-bounded loop converges on `N + 1` files in steady state (prune to N, then add one). Adding a second prune call after a successful write closes that gap, and placing it after the write (rather than making the pre-write call more aggressive) avoids deleting a file before the replacement is safely persisted — a reasonable trade-off for the failure case.

One edge case worth flagging (non-blocking): the new post-write call to `deleteRedundantBackupFiles()` sits inside the same `try` block, before `status` is set to `"Success"`. If that call throws an unchecked exception (e.g. `listFiles()` surfacing an I/O error as a `RuntimeException` — the per-file `delete()` calls are already individually try/caught, but `listFiles()` itself is not guarded), the outer `catch(Exception e)` will report `"Failed to back up storage"` and mark the `ActionRecord` as a failure even though the backup file was already written and persisted successfully. This would be a confusing, misleading status for an operation that actually succeeded. Consider wrapping the post-write prune call in its own try/catch (log-and-continue, matching the pattern already used for individual `delete()` failures) so a cleanup failure can never downgrade a successful backup to a reported failure. This is the same class of exposure that already existed for the pre-write call, but doubling the call sites doubles the surface area, and post-write is the case where mislabeling actually harms correctness of the reported state (pre-write, no backup had happened yet, so "failed" was still accurate).

### Visibility change
`deleteRedundantBackupFiles()` going from `private` to package-private for testability is reasonable and low-risk — confirmed via search that it has no other callers in the codebase, so this doesn't expand any real API surface beyond the test.

### Tests
- The 6 new test cases are well-targeted, each mapped to a specific defect/behavior (see the `[G1]`–`[G6]` comment block at the top of the file), and align with the described test plan.
- Mocking approach (static-mocking `SreeEnv`, `SecurityEngine`, `StorageTransfer`, spying on the service to short-circuit `deleteRedundantBackupFiles()` for the `doBackup()` sequencing tests) is reasonable for isolating the sequencing behavior without needing a full Spring context.
- Confirmed no other test or production code elsewhere in the repo depends on the old (buggy) retention count/ordering behavior, so this shouldn't regress anything else.
- The CI `build` check (which runs the test suite) already passed on this exact commit.

### Backward compatibility
As the PR description notes, this changes real retention behavior for existing deployments: previously-under-retaining installs will now converge on exactly `asset.backup.count` files (one more or fewer than before, depending on which bug dominated in a given deployment's history). This is a behavior correction, not an API break, but it may be worth a release note as the author suggests, since operators relying on the previous (buggy) file count could notice the difference on next backup cycle.

### Style / readability
- Comments added at each call site and in the updated Javadoc clearly explain the sequencing rationale — good improvement in maintainability over the original single-line comment.
- No lint, import, or formatting issues found in the diff.

### Summary
The two core bug fixes are correct and well-tested. The additional post-write prune call is a defensible design choice with a clearly-articulated rationale, with one non-blocking edge case noted above (a prune-call exception after a successful write would currently mislabel the operation as failed). No regressions or backward-incompatible API changes were found; the retention *behavior* change is intentional and documented.

### CI / mergeability
- All checks pass except `Claude Code Review / claude-review`, which is known to be broken/erroring on infrastructure grounds across many open PRs in this repo and is excluded from consideration per current review policy.
- `gh pr view` reports `mergeable: MERGEABLE` (no merge conflicts). The `BLOCKED` merge-state reflects the repository's required-review gate (no reviews yet), not a conflict.

**Verdict: Approve.** No blocking issues, no failing checks (excluding the known-broken `claude-review` check), and no merge conflicts.

PR #4861, opened 2026-08-28
Fix Bug #76341: resetSettings scope divergence from getSettings/applySettings

larryliang-inetsoft:
## Root cause

`PresentationSettingsController.resetSettings` computed `globalSettings` with
a formula that omitted the unconditional `|| !SUtil.isMultiTenant()` OR-term
present in both `getSettings` and `applySettings`. Instead, when the
deployment is single-tenant and the first computation is false, it fell back
to a live `provider.checkPermission(principal, EM, "*", ACCESS)` call and
*replaced* (not OR'd) the computed value with that result.

On a single-tenant, security-enabled deployment, a caller granted only the
narrow `settings/presentation/(org-)settings` `EM_COMPONENT` permission
(exactly what this endpoint's own `@Secured` annotation accepts as
sufficient to call it) is neither a site admin nor an org admin, so
`checkPermission(EM, "*", ACCESS)` resolves to `false` — no code path in the
repo ever grants a `Permission` on the literal `EM`/`"*"` resource.
`resetSettings` then computes `globalSettings = false` (org-scoped),
honoring the caller's own `orgSettings: true` request, while
`getSettings`/`applySettings` unconditionally force `globalSettings = true`
(global) for the identical principal/config via the bare
`!SUtil.isMultiTenant()` term. The reset silently writes to org scope while
the rest of the controller reads/writes global scope on that deployment
shape — an orphaned, silently ineffective reset that looks like success.

Git archaeology (`8904da03f`, "fix Bug #70090...") shows `applySettings`
already had the `!isMultiTenant` term when this `checkPermission` fallback
was added to `resetSettings` instead of mirroring it — an inconsistent
follow-up to the same original fix, not a deliberate broader-permission
design for a "more destructive" reset.

## Fix

`resetSettings` now matches `applySettings`'s formula verbatim: add the
unconditional `!SUtil.isMultiTenant()` OR-term, drop the `checkPermission`
fallback branch entirely (it only ever ran under `!SUtil.isMultiTenant()` to
begin with, so nothing multi-tenant-shaped changes).

## Tests

Added to `PresentationSettingsControllerTest`:
- `resetSettings_singleTenant_matchesGetAndApplyGlobalScope`
- `resetSettings_singleTenant_nonAdminNarrowPermission_matchesGetAndApplyGlobalScope`

Both fail against the pre-fix code (confirmed by temporarily reverting the
controller change) and pass after the fix. Full
`PresentationSettingsControllerTest` suite: 10/10 passing.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/presentation/PresentationSettingsController.java
+++ b/core/src/main/java/inetsoft/web/admin/presentation/PresentationSettingsController.java
@@ -275,14 +275,10 @@ public class PresentationSettingsController {
    public PresentationSettingsModel resetSettings(@RequestBody() PresentationSettingsModel model,
                                                   Principal principal) throws Exception
    {
-      SecurityProvider provider = securityEngine.getSecurityProvider();
-      boolean securityEnabled = !provider.isVirtual();
+      boolean securityEnabled = !securityEngine.getSecurityProvider().isVirtual();
       boolean globalSettings = OrganizationManager.getInstance().isSiteAdmin(principal) &&
-            (model.orgSettings() != null && !model.orgSettings()) || !securityEnabled;
-
-      if(!globalSettings && securityEnabled && !SUtil.isMultiTenant()) {
-         globalSettings = provider.checkPermission(principal,  ResourceType.EM, "*", ResourceAction.ACCESS);
-      }
+            (model.orgSettings() != null && !model.orgSettings()) || !securityEnabled ||
+            !SUtil.isMultiTenant();
 
       Lock settingsLock = cluster.getLock(SETTINGS_LOCK);
       settingsLock.lock();
--- a/core/src/test/java/inetsoft/web/admin/presentation/PresentationSettingsControllerTest.java
+++ b/core/src/test/java/inetsoft/web/admin/presentation/PresentationSettingsControllerTest.java
@@ -270,6 +270,52 @@ class PresentationSettingsControllerTest {
       verify(aiSettingsService, never()).resetSettings();
    }
 
+   // [single-tenant, site admin] globalSettings=true, matching what getSettings/applySettings
+   // compute for the identical principal/config (bug #76341) — !isMultiTenant forces global
+   // directly, so no checkPermission call is made
+   @Test
+   void resetSettings_singleTenant_matchesGetAndApplyGlobalScope() throws Exception {
+      when(orgManager.isSiteAdmin(principal)).thenReturn(true);
+      when(securityProvider.isVirtual()).thenReturn(false);
+      sUtilStatic.when(SUtil::isMultiTenant).thenReturn(false);
+      stubAllGetters(true);
+
+      controller.resetSettings(
+         PresentationSettingsModel.builder().orgSettings(true).build(), principal);
+
+      verify(fontMappingSettingsService).resetSettings();
+      verify(aiSettingsService).resetSettings();
+      verify(securityProvider, never()).checkPermission(any(), any(), any(), any());
+   }
+
+   // [single-tenant, non-admin caller granted only the narrow presentation-settings
+   // EM_COMPONENT permission] models a caller granted only settings/presentation/org-settings,
+   // who is not a site admin or org admin and so fails the broader EM/*/ACCESS check. Before the
+   // #76341 fix, resetSettings replaced its already-true single-tenant result with this failing
+   // checkPermission call and ended up org-scoped, diverging from getSettings/applySettings for
+   // the identical caller; after the fix !isMultiTenant forces global directly and
+   // checkPermission is never consulted.
+   @Test
+   void resetSettings_singleTenant_nonAdminNarrowPermission_matchesGetAndApplyGlobalScope()
+      throws Exception
+   {
+      when(orgManager.isSiteAdmin(principal)).thenReturn(false);
+      when(securityProvider.isVirtual()).thenReturn(false);
+      sUtilStatic.when(SUtil::isMultiTenant).thenReturn(false);
+      stubAllGetters(true);
+
+      lenient().when(securityProvider.checkPermission(
+            any(), eq(ResourceType.EM), eq("*"), eq(ResourceAction.ACCESS)))
+         .thenReturn(false);
+
+      controller.resetSettings(
+         PresentationSettingsModel.builder().orgSettings(true).build(), principal);
+
+      verify(fontMappingSettingsService).resetSettings();
+      verify(aiSettingsService).resetSettings();
+      verify(securityProvider, never()).checkPermission(any(), any(), any(), any());
+   }
+
    // -------------------------------------------------------------------------
    // helpers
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft:
# Bug #76341 -- review round 1

PR: https://github.com/inetsoft-technology/stylebi/pull/4861 (repo `stylebi`, base
`stylebi-wiz-main-rebase`), head commit `7a3c79223`.

Reviewed via `gh pr diff 4861` and by reading the current file state directly in
`C:/src/agent/wiz/stylebi-enterprise-bug-76341/community` at `7a3c79223` (working tree clean,
matches PR head). Read `01-diagnosis.md`, `02-refute.md`, `03-fix.md` in full first; this round
independently re-verified every claim in `03-fix.md` against the actual source rather than
trusting the write-up.

## What was checked

**Diff scope**: exactly two files changed --
`core/src/main/java/inetsoft/web/admin/presentation/PresentationSettingsController.java`
(+3/-7) and the matching test file (+46/-0). No unrelated files touched.

**Formula match (claim 1)**: read `applySettings` (:173-176) and `resetSettings` (:278-281)
side by side in the current file, not just the diff hunk. They are byte-for-byte identical:

```
boolean securityEnabled = !securityEngine.getSecurityProvider().isVirtual();
boolean globalSettings = OrganizationManager.getInstance().isSiteAdmin(principal) &&
      (model.orgSettings() != null && !model.orgSettings()) || !securityEnabled ||
      !SUtil.isMultiTenant();
```
Confirmed.

**Removed `provider` variable (claim 2)**: read the full current `resetSettings` method body
(:275-314). No reference to `provider` remains anywhere in it (grep over the method also
returns nothing). The `getOrganization(currOrgID)`/`InvalidOrgException` check the diagnosis
worried about is present in `getSettings` (:105-107) and `applySettings` (:179-181) only --
`resetSettings` never had that check, before or after this diff. Confirmed independently, not
just taken from the fix doc's grep claim.

**checkPermission removal / multi-tenant behavior (claim 3)**: the removed block was
`if(!globalSettings && securityEnabled && !SUtil.isMultiTenant()) { globalSettings =
provider.checkPermission(...); }` -- gated on `!SUtil.isMultiTenant()`, i.e. it only ever ran
when the deployment is single-tenant. Under multi-tenant (`isMultiTenant()==true`) the guard was
always false, so the branch never executed and `globalSettings` kept whatever the first line
computed -- identical before and after this diff. Multi-tenant behavior is untouched. Confirmed.

**Security-hardening angle (task's specific ask)**: traced whether the new unconditional
`!SUtil.isMultiTenant()` OR-term could ever narrow the caller population that gets global scope
relative to the old code. In single-tenant mode the new formula is unconditionally `true`
(the OR-term alone forces it), which is a strict superset of what the old formula could produce
in that mode (`true` when checkPermission/isSiteAdmin fired, `false` otherwise) -- there is no
input on which old computed `true` and new computes `false`. In multi-tenant mode neither branch
is reachable/changed. So this cannot reduce who gets global scope; it can only ever match or
widen it in single-tenant mode, and "widen to unconditional global in single-tenant" is exactly
the existing, already-shipped behavior of `getSettings`/`applySettings` -- not a new risk
introduced by this fix, just resetSettings catching up to a design decision already made and
shipped for its two siblings.

**Regression tests (claim 4)**: read both new test bodies in full and hand-traced the boolean
formula against the *old* (pre-fix) code for each:

- `resetSettings_singleTenant_matchesGetAndApplyGlobalScope` (isSiteAdmin=true,
  securityEnabled=true, isMultiTenant=false, orgSettings=true): old first-line formula evaluates
  to `false` (`(true && (true && !true)) || !true` = `false`), so old code enters the
  `checkPermission` branch; the test never stubs `checkPermission`, so Mockito's default `false`
  applies -> old code ends up `globalSettings=false` -> `fontMappingSettingsService.resetSettings()`
  never called -> test would fail against pre-fix code.
- `resetSettings_singleTenant_nonAdminNarrowPermission_matchesGetAndApplyGlobalScope`
  (isSiteAdmin=false, orgSettings=true, isMultiTenant=false, `checkPermission` stubbed `false`):
  old first line evaluates `false` the same way, enters the branch, and the explicit `false` stub
  makes `globalSettings=false` -> same failure mode.

Both tests fail under the pre-fix formula for the reason the fix doc claims ("Wanted but not
invoked: fontMappingSettingsService.resetSettings()"), and both pass under the new formula
(`!isMultiTenant` forces `true` directly, `checkPermission` never called -- confirmed by the
`verify(securityProvider, never()).checkPermission(...)` assertion in both tests, which is the
actual regression-locking assertion). The claimed revert-sanity result (2 failures / 8 passes) is
credible and consistent with this trace. The tests are genuinely load-bearing, not tautological
against the reported scenario only -- the second test in particular pins down the refutation's
corrected, more-general trigger (any non-admin caller with only the narrow
presentation-settings grant), not just the diagnosis's original (refuted) SSO scenario.

**Alternate scenarios walked (per review-workflow's invariant-check requirement)**: named
invariant -- "resetSettings computes the same `globalSettings` as
getSettings/applySettings for any principal/config, in both single- and multi-tenant mode."
Walked: the other two sibling methods (get/apply) already carry the identical formula, so no
further divergence exists to find; multi-tenant path (unchanged, confirmed above); org-admin and
site-admin callers (both already forced global via the `isSiteAdmin`/`!securityEnabled` terms
regardless of the OR-term, so no new divergence there either); the non-admin narrow-permission
caller (covered by test 2); empty/null `model.orgSettings()` (short-circuits the first
sub-expression to `false` either way, doesn't affect the OR-term). Tried to build a
counter-example where the fixed `resetSettings` diverges from `applySettings` given the formulas
are now textually identical and operate on the same inputs (`principal`, `model.orgSettings()`,
`securityEnabled`, `isMultiTenant()`) -- could not construct one; the two methods are now
provably equivalent on this flag for any input.

**Other**: `provider.checkPermission(principal, ResourceType.EM, "*", ResourceAction.ACCESS)` was
re-confirmed (own grep) to not appear anywhere else in the file or method after the diff -- the
removal is fully local. The three sibling formulas remain independently duplicated (flagged by
the diagnosis as a follow-up refactor, out of scope for this minimal fix) -- not a finding
against this PR, noting it only because CLAUDE.md favors minimal-scope fixes and duplicating a
now-3x-repeated formula is a known, accepted tradeoff here, not an oversight.

CI note: the automated `claude-review` GitHub Action check on the PR shows a job error
("Claude encountered an error after 0s"), unrelated to this manual review -- Semgrep and CLA
checks pass, `build` was still pending at review time. Not a code issue in the diff.

## Findings

No findings at or above "important" severity. No findings below that threshold either that
warrant a fixer round.

## Verdict

**Approved.**

Note: this session's account cannot self-approve this PR on GitHub (`gh pr review --approve`
fails as expected for a non-collaborator/own-account restriction) -- posting the full review
below as a PR comment instead, per the task's fallback instruction. The lead should merge
directly based on this written verdict.

PR #4862, opened 2026-08-28
Fix bug #76361, stale originalIndex silently edits wrong portal tab

larryliang-inetsoft:
## Root cause

`PortalIntegrationViewSettingsService.setModel` resolves each submitted `PortalTabModel` by
indexing directly into `PortalThemesManager`'s live, shared tab list using a caller-echoed
`originalIndex`, with no bounds check and no identity check. `PortalThemesManager` is a Spring
singleton whose `getPortalTabs()` returns the live field by reference (no defensive copy), and
`setModel` itself reloads from disk (`loadThemes()`) as its first statement -- so the race fires
across two *sequential* requests, not just literal concurrency: if another admin reorders or
removes a tab between when a client read the model and when it submits, the client's stale
`originalIndex` either

- throws a raw `ArrayIndexOutOfBoundsException` out of an `@Audited` service method (list shrank), or
- silently applies the edit to the wrong tab (list reordered, same length) -- no error, no indication
  anything went wrong.

## Fix

Two-part fix scoped entirely to `setModel`, no model/DTO changes:

1. Unconditional bounds check on `originalIndex` against `portalTabs.size()` -- throws a clear
   `MessageException` instead of letting `ArrayIndexOutOfBoundsException` escape.
2. Identity check (`tab.getName()` match) for **non-editable (built-in) tabs only** -- for those,
   `getModel` always echoes the raw internal name verbatim and `setModel`'s non-editable branch
   never mutates it, so a non-racing round trip is guaranteed to match. Editable/custom tabs are
   **not** identity-checked, since `tabModel.name()`/`tabModel.uri()` legitimately carry the *new*
   value on a legitimate rename -- comparing against the live tab's current name would
   false-positive. The bounds check alone is what's achievable there without a model change
   (residual gap, not fully closed by this fix).

## Other callers

`PresentationSettingsAccess` (admin-chat's Presentation area, #4844) wires this same
`getModel`/`setModel` as one of 16 sub-model adapters -- it goes through the same choke point, so
this fix protects it too. Confirmed `PresentationChangesetApplyService.apply`'s existing
`catch(Exception e)` around the `access.write(...)` call handles the new `MessageException` the
same as any other exception (recorded as `STATUS_FAILED`, no type-specific catch anywhere in that
path) -- no behavior change in kind for that caller.

`PresentationChangePlanService.requireWholeTabList` is a separate, narrower existing mitigation for
a shorter-array-drops-tabs failure mode; it does not overlap with this fix's stale-index scenario.

## Tests

New file `core/src/test/java/inetsoft/web/admin/presentation/PortalIntegrationViewSettingsServiceTest.java`
(none existed). 3 cases: stale out-of-bounds index, stale index pointing at a different built-in
tab after a reorder, and a non-racing baseline including a legitimate editable-tab rename. All
green; verified both racy-index cases fail against pre-fix code (one crashes with
`ArrayIndexOutOfBoundsException`, the other silently succeeds).

Fixes #76361

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/presentation/PortalIntegrationViewSettingsService.java
+++ b/core/src/main/java/inetsoft/web/admin/presentation/PortalIntegrationViewSettingsService.java
@@ -146,6 +146,7 @@ public class PortalIntegrationViewSettingsService {
       List<PortalTab> portalTabs = manager.getPortalTabs();
       List<PortalTabModel> tabModels = model.tabs();
       int insertIndex = 0;
+      Catalog catalog = Catalog.getCatalog(principal);
 
       for(int i = 0; i < tabModels.size(); i++) {
          PortalTabModel tabModel = tabModels.get(i);
@@ -156,16 +157,14 @@ public class PortalIntegrationViewSettingsService {
 
             if(originalIndex < 0 || originalIndex >= portalTabs.size()) {
                throw new MessageException(
-                  "Tab list has changed since it was loaded; please reload Presentation " +
-                  "settings and try again.");
+                  catalog.getString("em.presentation.portalIntegration.tabsChanged"));
             }
 
             tab = portalTabs.get(originalIndex);
 
             if(!tab.isEditable() && !Objects.equals(tab.getName(), tabModel.name())) {
                throw new MessageException(
-                  "Tab list has changed since it was loaded; please reload Presentation " +
-                  "settings and try again.");
+                  catalog.getString("em.presentation.portalIntegration.tabsChanged"));
             }
 
             if(tab.isEditable()) {
--- a/core/src/main/java/inetsoft/web/admin/presentation/PortalIntegrationViewSettingsService.java
+++ b/core/src/main/java/inetsoft/web/admin/presentation/PortalIntegrationViewSettingsService.java
@@ -21,6 +21,7 @@ import inetsoft.sree.SreeEnv;
 import inetsoft.sree.portal.PortalTab;
 import inetsoft.sree.portal.PortalThemesManager;
 import inetsoft.util.Catalog;
+import inetsoft.util.MessageException;
 import inetsoft.util.Tool;
 import inetsoft.util.audit.ActionRecord;
 import inetsoft.web.admin.presentation.model.PortalIntegrationSettingsModel;
@@ -151,7 +152,21 @@ public class PortalIntegrationViewSettingsService {
          PortalTab tab;
 
          if(tabModel.originalIndex() != null) {
-            tab = portalTabs.get(tabModel.originalIndex());
+            int originalIndex = tabModel.originalIndex();
+
+            if(originalIndex < 0 || originalIndex >= portalTabs.size()) {
+               throw new MessageException(
+                  "Tab list has changed since it was loaded; please reload Presentation " +
+                  "settings and try again.");
+            }
+
+            tab = portalTabs.get(originalIndex);
+
+            if(!tab.isEditable() && !Objects.equals(tab.getName(), tabModel.name())) {
+               throw new MessageException(
+                  "Tab list has changed since it was loaded; please reload Presentation " +
+                  "settings and try again.");
+            }
 
             if(tab.isEditable()) {
                tab.setName(tabModel.name());
--- a/core/src/main/resources/inetsoft/util/srinter.properties
+++ b/core/src/main/resources/inetsoft/util/srinter.properties
@@ -4464,6 +4464,7 @@ em.presentation.lookAndFeel.css.invalidCssColor=Not a valid color value
 em.presentation.lookAndFeel.css.invalidCssValue=Invalid characters in the value
 em.presentation.openBookmarkTip=Display bookmarks (report headings) immediately upon opening PDF.
 em.presentation.openThumbnailTip=Display page thumbnails immediately upon opening PDF.
+em.presentation.portalIntegration.tabsChanged=Tab list has changed since it was loaded; please reload Presentation settings and try again.
 em.presentation.theme.confirm=The theme has unsaved changes, close anyway?
 em.presentation.theme.duplicateName=A theme with this name already exists.
 em.propertyEditor.title=Property Editor
--- a/core/src/test/java/inetsoft/web/admin/presentation/PortalIntegrationViewSettingsServiceTest.java
+++ b/core/src/test/java/inetsoft/web/admin/presentation/PortalIntegrationViewSettingsServiceTest.java
@@ -1,0 +1,168 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.web.admin.presentation;
+
+import inetsoft.sree.SreeEnv;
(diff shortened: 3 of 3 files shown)


larryliang-inetsoft:
# Bug #76361 -- Review Round 1

PR: https://github.com/inetsoft-technology/stylebi/pull/4862 (repo `stylebi`, base
`stylebi-wiz-main-rebase`), head commit `d4d8180b1`.

Reviewed the diff (`gh pr diff 4862`), the PR description, and the full current source of
`PortalIntegrationViewSettingsService.java`, `PortalTab.java`, `PortalTabModel.java`,
`PresentationSettingsAccess.java`, `PresentationChangePlanService.java`, and
`PresentationChangesetApplyService.java`, independently against the diagnosis/refutation/fix docs.

## What was checked and confirmed correct

1. **Bounds check + identity check placement in `setModel`** (lines 154-169 of the current file,
   read in full, not just the diff hunk): the unconditional bounds check
   (`originalIndex < 0 || originalIndex >= portalTabs.size()`) runs first for every tab, editable
   or not. The identity check (`!tab.isEditable() && !Objects.equals(tab.getName(), tabModel.name())`)
   runs only for non-editable tabs, before the `if(tab.isEditable())` branch that would otherwise
   overwrite `name`/`uri`. Confirmed genuinely scoped to non-editable tabs only, not accidentally
   also firing for editable ones.

2. **Residual gap for editable/custom tabs is real and not overstated.** Read `PortalTab.java`
   (fields: `name, label, uri, visible, editable, encoding` -- no id, no
   equals/hashCode) and `PortalTabModel.java` (`name, label, uri, visible, editable,
   originalIndex` -- a `@Value.Immutable` interface, no id). Neither has a caller-opaque identity
   field. The PR description states this gap plainly ("Editable/custom tabs are not
   identity-checked... residual gap, not fully closed by this fix") -- no overstatement in the PR
   body or in code comments.

3. **`PresentationChangesetApplyService.apply`'s `catch(Exception e)` handles the new
   `MessageException` sanely.** Read the method in full (lines 78-201). `access.write(...)` (line
   128, the exact call that reaches `portalIntegration.setModel`) is wrapped by `try` at 127 /
   `catch(Exception e)` at 152-163. `MessageException extends RuntimeException`, so it's caught
   here identically to how the prior `ArrayIndexOutOfBoundsException` would have been: recorded as
   `STATUS_FAILED` with `messageOf(e)`, added to `unknownStateFailures`, audited, `failed = true`.
   No type-specific catch anywhere in this path. Confirmed as claimed.

4. **The 3 regression tests are genuinely load-bearing.** Read the full test file. Traced
   pre-fix behavior by hand for each:
   - Test 1 (`staleIndexOutOfBounds`): pre-fix, `portalTabs.get(2)` on a 2-element
     `Arrays.asList(...)` throws `ArrayIndexOutOfBoundsException` (Arrays$ArrayList.get is a raw
     array access, not a bounds-checked `ArrayList.get` -- matches the PR description's specific
     exception type, not a generic claim).
   - Test 2 (`staleIndexPointsAtDifferentBuiltInTab`): pre-fix, `tab = portalTabs.get(1)` resolves
     to "Schedule"; since it's non-editable the loop does no field writes and pushes it through
     unchanged -- method returns normally, so `assertThrows(MessageException.class, ...)` correctly
     fails with "nothing was thrown", exactly as the fix doc claims.
   - Test 3 is a true baseline: editable-tab rename with `originalIndex` unchanged, which the
     identity check (non-editable only) correctly leaves alone.
   Both racy tests genuinely pin the bug; the baseline genuinely guards against a false-positive
   from the fix itself.

5. **`mockStatic(SreeEnv.class)` is a legitimate test-setup necessity**, not masking behavior the
   test should exercise. `setModel` calls `SreeEnv.setProperty`/`SreeEnv.save()` unconditionally
   before the tab loop (lines 141-143, 190), and there is no Spring context in this unit test, so
   an unmocked static call throws `ShutdownException` unrelated to what's being tested. Matches
   `PortalThemesManagerTest`'s own convention.

6. **Diff scope**: only `PortalIntegrationViewSettingsService.java` (the one-hunk fix) and the new
   test file are touched. No unrelated changes.

7. **Same-pattern sweep**: grepped the whole `community` tree for `originalIndex` outside the test
   file -- only 3 hits, all already accounted for (`PortalTabModel.java` itself, the fixed
   `PortalIntegrationViewSettingsService.java`, and `PresentationChangePlanService.java`'s
   `requireWholeTabList`, which is a different, already-existing, complementary mitigation for a
   different failure mode -- not the same defect living elsewhere).

8. **"Other call sites" correction from the refutation pass** is properly reflected in the final
   PR description (`PresentationSettingsAccess` wiring, `requireWholeTabList`'s narrower scope) --
   the refutation's correction was actually incorporated, not just acknowledged and dropped.

## Finding: important (confidence ~82) -- PR's own convention citation for `MessageException` is factually wrong for one of its three examples

The PR description states: *"MessageException (plain-string constructor) is the established
convention in this package family for user-facing validation refusals (confirmed via
RepositoryFolderService.java, LogSettingService.java, AuthenticationProviderService.java -- all
throw new MessageException(...) with a plain string, no catalog key, for this class of error)."*

This is not accurate. Grepped all three cited files directly:

- `RepositoryFolderService.java`: **all three** `MessageException` call sites use
  `catalog.getString(...)` / `Catalog.getCatalog().getString(...)` (lines 79, 140, 207) -- zero
  plain-string instances. The citation is backwards for this file.
- `AuthenticationProviderService.java`: genuinely mixed -- some sites are plain strings ("Could
  not initialize security.", "Authentication provider named \"...\" does not exist"), but the
  closest analog to *this* bug's scenario -- a user-facing validation refusal for a
  pre-existing, nameable condition -- uses a catalog key instead
  (`Catalog.getCatalog().getString("security.authentication.exists")` for a duplicate-name
  refusal, twice). The plain-string examples in this file read more like internal/unexpected-state
  errors than user-facing validation messages.
- `LogSettingService.java`: this one does support the claim (line 229, plain string, no catalog
  key) -- the only clean precedent of the three.

So the actual convention in this package family leans toward `Catalog.getString(...)` for
validation-refusal-style messages specifically (the category this bug's "reload and try again"
message belongs to), with plain strings reserved more for lower-level/unexpected-state failures.
The PR's own stated rationale for its convention choice doesn't hold up against two of its three
citations.

This doesn't make the code wrong -- plain-string `MessageException` does have real precedent
(`LogSettingService`, part of `AuthenticationProviderService`) and the message here is new text
with no existing catalog key to reuse, so keeping it a plain string is a defensible call. But the
PR description overstates its evidence, and given the closest on-point precedent
(`security.authentication.exists`, a duplicate/conflict-type refusal) went through
`Catalog.getString`, this deserves either: (a) a corrected PR description that doesn't
misrepresent `RepositoryFolderService.java`, or (b) switching to `Catalog.getCatalog(principal)
.getString(...)` with a new catalog key for i18n consistency with the closer precedent. Not
blocking the merge -- the message itself is functionally correct and the exception is handled
correctly end-to-end -- but the fixer should address the citation/description accuracy, since
`AdminExceptionHandler` renders this message directly to the end user and it will currently ship
un-localized while sibling validation refusals in the same package are localized.

## Verdict: changes requested (one important-severity finding, no blockers)

Everything else -- the core fix logic, its placement, the residual-gap honesty, the regression
tests' credibility, the cross-caller impact analysis, and diff scope -- holds up under independent
re-verification against current source. Sending the one finding above back to the fixer.

larryliang-inetsoft:
Review round 2 (final) -- approved, ready to merge.

Independently re-verified the round-1 finding is genuinely resolved: both `MessageException` sites in `setModel` now use `catalog.getString("em.presentation.portalIntegration.tabsChanged")`, the key is present in srinter.properties (correctly placed, no collision), the `Catalog` local is correctly scoped before both throw sites, and `git diff d4d8180b1 1e7d482e1` confirms the change is message-construction-only -- the bounds/identity check logic is byte-for-byte unchanged. The 3 regression tests are untouched and still assert only `MessageException.class`; I re-ran them myself (not just trusting the fix doc): `Tests run: 3, Failures: 0, Errors: 0, Skipped: 0`.

Final whole-PR sanity pass found no new issues. Diff scope stays tight (one service file, one test file, one properties line). PR description was actually updated to drop the disputed precedent citation.

Full writeup: docs/teams/2026-08-28-bugs-admin-plugin-defects/bug-76361/06-review-r2.md

No blockers, no important findings. Approving -- note: this account cannot self-approve its own PR via gh pr review, posting this as a plain comment instead per protocol; lead should merge directly.

PR #4863, opened 2026-08-28
Bug #76359, enforce the DATABASE/CUSTOM authentication-provider license gate

larryliang-inetsoft:
## Summary

`AuthenticationProviderService.getAuthenticationProvider`'s own advisory `dbProviderEnabled`/`customProviderEnabled` flags show the product intends DATABASE and CUSTOM authentication providers to be enterprise-only, but `getProviderFromModel` -- the actual create/edit entry point behind `addAuthenticationProvider`/`editAuthenticationProvider` -- had no license check at all. A raw client hitting `add-authentication-provider`/`edit-authentication-provider` directly with `providerType: DATABASE` or `CUSTOM` could create or convert to one of these types regardless of license tier.

Adds `checkProviderTypeLicensed`/`isSameProviderType`, called only from:
- `addAuthenticationProvider` (`existingProvider` always null -- every add is a new provider)
- `editAuthenticationProvider` (`existingProvider` is the pre-edit live object, checked before `getProviderFromModel` runs)

Behavior:
- New creation of a DATABASE/CUSTOM provider, or type-conversion into DATABASE/CUSTOM from another type, is blocked under a non-enterprise license.
- Editing an already-existing DATABASE/CUSTOM provider **without** changing its type (rename, or an unrelated field) is left untouched -- grandfathered configuration, not new unlicensed exposure.
- The ~13 read-path methods that share `getProviderFromModel` via `withProviderFromModel` (`getUsers`, `testConnection`, etc.) are deliberately left ungated, matching pre-fix behavior.
- `copyAuthenticationProvider` is blocked when copying an existing DATABASE/CUSTOM provider under a non-enterprise license, since it always creates a new entry via `addAuthenticationProvider`.

Adds two catalog keys (`em.securityProvider.databaseNotLicensed`/`customNotLicensed`) following the existing `em.securityProvider.*` family, and `AuthenticationProviderServiceTest` coverage for both the newly-blocked cases and the explicitly-preserved grandfathered-provider/read-path cases.

**Note:** this branch is built on top of the #76358 fix (commit `5f9962670`, PR #4847), which is still under review and not yet merged. Please merge that one first (or rebase this branch onto `main` once it lands) -- this PR's own diff does not touch #76358's null-guard change.

Full diagnosis/refutation trail: `docs/teams/2026-08-28-bugs-admin-plugin-defects/bug-76359/` in `stylebi-wiz`.

## Test plan
- [x] `AuthenticationProviderServiceTest` (10 tests, all passing): `mvn -pl core -am -Dtest=AuthenticationProviderServiceTest test`
- [x] Confirmed the 4 new gate-related tests genuinely fail (NPE/no-throw) when the fix is reverted, and pass again once restored
- [x] No regressions in `inetsoft.web.admin.security.*Test` (54 tests) or `inetsoft.web.admin.ai.providers.*Test` (42 tests) -- the latter covers `ProviderChangesetApplyService`/`ProviderChangePlanService`, the other known callers into `addAuthenticationProvider`


larryliang-inetsoft:
# Bug #76359 -- Review round 1

PR: https://github.com/inetsoft-technology/stylebi/pull/4863 (repo `stylebi`, base
`stylebi-wiz-main-rebase`), head commit `cdef342f8`.
Worktree used: `C:/src/agent/wiz/stylebi-enterprise-bug-76359/community`.

## What was checked, independently

Read `01-diagnosis.md` (including "Revision (round 2)"), `02-refute.md` (including "Round 2
recheck", verdict CONFIRMED), and `03-fix.md` in full, including the rebase note. Then
independently re-verified against the live worktree and `gh pr diff 4863`, not trusting either
doc's account:

1. **Diff scope**: `gh pr diff 4863 --name-only` shows exactly three files --
   `AuthenticationProviderService.java`, `srinter.properties`, and
   `AuthenticationProviderServiceTest.java`. Matches `03-fix.md` exactly, nothing extra.

2. **`isSameProviderType`/`checkProviderTypeLicensed` implementation**: read the actual diff hunk.
   Matches the round-2 revised design verbatim -- guard is
   `(newType == DATABASE || newType == CUSTOM) && (existingProvider == null ||
   !isSameProviderType(existingProvider, newType))`, throwing only when `!isEnterprise()`.

3. **Call-site placement**: read `addAuthenticationProvider` (138-165) and
   `editAuthenticationProvider` (168-220) in full from the live file (not the diff hunk alone).
   - `addAuthenticationProvider`: `checkProviderTypeLicensed(model.providerType(), null)` sits
     after the "name already exists" guard and before `getProviderFromModel(model)`. Since add is
     always a new provider, `existingProvider` is unconditionally `null` here -- confirmed by
     reading the method, not assumed.
   - `editAuthenticationProvider`: the call sits inside the `for` loop, right after `found = true`
     (the pre-edit `provider` object is bound at that point) and before `getProviderFromModel`,
     inside the existing `chain.writeLock()/finally{chain.writeUnlock()}` block. A thrown
     `MessageException` propagates through the existing `finally` exactly like the pre-existing
     `orElseThrow` two lines below it -- no lock-leak risk.

4. **Independently traced the safety property myself** (not just re-reading the refuter's trace):
   - Existing DATABASE/CUSTOM provider, same-type edit (rename or unrelated field),
     `!isEnterprise()`: `isSameProviderType` returns true for same runtime class vs. declared
     type -> `introducingUnlicensedType` false -> **allowed**. Confirmed against the real switch
     arms (`FileAuthenticationProvider`/`DatabaseAuthenticationProvider`/`LdapAuthenticationProvider`/
     `case null, default -> CUSTOM`) and the fact `AuthenticationProviderService.java` already
     imports all three concrete classes plus `SecurityProviderType`, so nothing here requires an
     import that's missing.
   - Existing LDAP/FILE provider, type changed to DATABASE/CUSTOM, `!isEnterprise()`:
     `existingProvider != null` but `isSameProviderType` false (LDAP provider vs. newType DATABASE)
     -> `introducingUnlicensedType` true -> **blocked**. Correct, this is a genuine new-exposure
     type conversion.
   - Existing DATABASE/CUSTOM provider, type changed down to FILE/LDAP: `newType` is neither
     DATABASE nor CUSTOM, so the leading guard clause is false regardless of `existingProvider`
     -> **never blocked**, migrating off the unlicensed type stays open. Correct.
   - Fresh add of DATABASE/CUSTOM under `!isEnterprise()`: `existingProvider` is always `null` at
     that call site -> **blocked**. Correct, matches the reported bug.
   - `getProviderFromModel`/`withProviderFromModel` and all read-path methods (verified `getUsers`
     at line 368-371 routes through `withProviderFromModel` -> the private `getProviderFromModel`
     at line 705, which never touches `securityEngine`/the chain at all -- it's a pure
     model-to-provider constructor) have zero calls to `checkProviderTypeLicensed` anywhere in the
     diff -- confirmed by grepping the diff hunk for the new method's call sites, there are exactly
     two (`addAuthenticationProvider`, `editAuthenticationProvider`), matching the narrowed design's
     stated intent to leave the read path ungated.

   No case found where the narrowed design's own stated invariant fails against the actual current
   code.

5. **Test bodies -- read every one of the 8 new tests**, not just the names:
   - `addAuthenticationProvider_databaseType_notLicensed_throws` / `..._customType_..._throws`:
     straightforward create-blocked cases, mocks line up (empty chain, license false).
   - `addAuthenticationProvider_copyOfExistingDatabaseProvider_notLicensed_throws`: models the
     `copyAuthenticationProvider` controller path structurally (new name, `existingProvider` null
     at the service layer) -- correctly exercises the "copy is blocked" resolution without needing
     to invoke the controller class itself, since the controller unconditionally delegates to
     `addAuthenticationProvider` with a not-yet-existing name.
   - `editAuthenticationProvider_sameTypeUnrelatedFieldEdit_notLicensed_doesNotThrow`: uses
     `providerName == name` (no rename), so the earlier `!model.providerName().equals(name) &&
     getProviderByName(...)` guard short-circuits without needing `chain.stream()` mocked --
     verified this against `getProviderByName`'s call sites and the guard's boolean-short-circuit
     order in the real source; the test's mock setup (only `getProviderByName` via
     `getProviders()`, no `stream()` mock) is consistent with that, not an accidental gap.
   - `editAuthenticationProvider_rename_sameType_notLicensed_doesNotThrow`: renames, so the
     "already exists" guard does call `getProviderByName(model.providerName())`, which the test
     mocks via `chain.stream()` returning the *old*-named provider (so the new name is correctly
     not found) -- consistent with `getProviderByName`'s expected lookup-by-name semantics.
   - `editAuthenticationProvider_typeConversionToDatabase_notLicensed_throws`: LDAP -> DATABASE,
     asserts throw. Model has no `dbProviderModel()` set, but that's fine -- the check fires and
     throws before `getProviderFromModel` is ever reached, so the missing dbProviderModel is never
     dereferenced. Verified this ordering directly against the real method body.
   - `getUsers_customType_notLicensed_doesNotThrow`: confirmed above that this path never touches
     `checkProviderTypeLicensed`, so `securityEngine` not being mocked at all is consistent (that
     mock is never consulted).
   - `addAndEditAuthenticationProvider_enterpriseLicensed_doesNotThrow`: confirms the licensed path
     is a no-op for both add and a type-conversion edit.

   All 8 are genuinely load-bearing -- none merely restates the exact reported repro input in a way
   that would still pass against a narrower, still-broken variant of the fix. In particular tests
   4/5 (same-type edit/rename) and 6 (type-conversion) together pin down the exact narrow/broad
   boundary the whole design revision exists to establish, not just the reported-bug shape.

6. **DATABASE/CUSTOM test substitution soundness**: confirmed `checkProviderTypeLicensed`'s
   `introducingUnlicensedType` guard branches purely on the `SecurityProviderType` enum value via
   `==` comparison and on `isSameProviderType`'s boolean result -- it has no DATABASE- or
   CUSTOM-specific behavior of its own (no branch that inspects `dbProviderModel`/
   `customProviderModel` contents, or calls anything type-specific). So substituting CUSTOM (via
   the in-test `TestCustomAuthenticationProvider`, which needs no live database) for DATABASE in
   cases 3/4 tests the identical code path and produces the identical result CUSTOM vs. DATABASE
   would have. Sound substitution, not an accidental different-path test.

7. **Catalog keys**: both `em.securityProvider.databaseNotLicensed` and
   `...customNotLicensed` appear exactly once each in `srinter.properties` (grepped the whole
   file) -- no collision with any pre-existing key, and they sit alongside the existing
   `em.securityProvider.*` family as claimed.

8. **Rebase integrity**: `git log` in the worktree shows `cdef342f8` sitting directly on
   `2811857c5` (mainline), and `gh pr diff 4863` contains exactly the three files/hunks described
   above -- nothing lost, nothing duplicated, no stray content from bug #76358 (already merged
   separately) reappearing in this diff.

## CI note (not a finding against this diff)

`gh pr checks 4863` shows the `claude-review` check failing and `build` pending at review time.
Pulled the `claude-review` run log (`gh run view 33215874821 --log`): it fails immediately with
`ANTHROPIC_API_KEY: ` (empty) and "Claude execution failed: result is_error:true" -- an automated
review bot with no API key configured in this repo's CI secrets, not a code-quality finding. This
is an infra/CI-config issue orthogonal to the diff's content, not something the fixer can or
should address as part of this bug. Semgrep (both scans) and CLA pass. `build` was still pending
at the time of this review; not blocking on it since it's unrelated to the reviewed logic (no
build-affecting syntax issue found by direct reading).

## Verdict

**Approved.** No blocker or important findings. The narrowed design is implemented exactly as
specified in the round-2 revision: it blocks new-creation and type-conversion into DATABASE/CUSTOM
under `!isEnterprise()` (including via `copyAuthenticationProvider`, which is structurally a new
create), while allowing same-type edits, renames, and all read-path methods on an already-existing
DATABASE/CUSTOM provider to keep working under a downgraded license -- confirmed under independent
tracing against the real source. All 8 regression tests are genuinely load-bearing and the
DATABASE->CUSTOM test substitution is sound, the rebase lost/duplicated nothing, diff scope is
exactly the three intended files, and the two new catalog keys are correctly formatted with no
collisions.

Note for the record: this reviewing session's account cannot self-approve this PR on GitHub
(`gh pr review --approve` will fail under the same restriction noted in the task). Posting this
verdict as a PR comment instead of an approval review; the coordinator should merge directly based
on this written verdict, per standing team convention for self-review-blocked PRs.

PR #4872, opened 2026-08-29
Bug #76342, enforce cluster.pause.enabled server-side for pause/resume

larryliang-inetsoft:
## Summary
- `cluster.pause.enabled` was read only by `ClusterService.getClusterEnabled()`, which feeds the EM console's Pause/Resume button visibility — neither `ClusterService.pauseServers`/`resumeServers` nor `ClusterController.pauseServer`/`resumeServer` ever checked it, so a direct REST call to `POST /api/em/monitoring/cluster/pause-server` (or `resume-server`) could pause/resume cluster servers even with the property unset/`"false"` (the shipped default).
- Extracted the existing boolean expression into `ClusterService.isPauseEnabled()`, reused by `getClusterEnabled()` and new guards at the top of `pauseServers`/`resumeServers` that throw `inetsoft.sree.security.SecurityException` when disabled (fail loud, not a silent no-op).
- Added an `@ExceptionHandler(SecurityException.class)` in `AdminExceptionHandler` mapping that checked exception to 403, since it wasn't already covered there (the existing 403 handling only applied to the unchecked `java.lang.SecurityException` from `SecuredAspect`, wrapped in `UndeclaredThrowableException`).

## Test plan
- [x] New `core/src/test/java/inetsoft/web/admin/cluster/ClusterServiceTest.java`: pause/resume both throw and never call the underlying client when the property is false/unset, and both still proceed normally when the property is `"true"` (no regression to the enabled path).
- [x] Updated `ClusterControllerTest` for the new checked `throws SecurityException` on `pauseServer`/`resumeServer`.
- [x] `./mvnw.cmd -pl core test -Dtest=ClusterServiceTest,ClusterControllerTest` — 10/10 passing.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/AdminExceptionHandler.java
+++ b/core/src/main/java/inetsoft/web/admin/AdminExceptionHandler.java
@@ -151,6 +151,22 @@ public class AdminExceptionHandler {
    }
 
    /**
+    * Error handler for access denied when a controller throws the checked
+    * {@link SecurityException} directly (not via {@link inetsoft.web.security.SecuredAspect},
+    * which is handled above through {@link #handleUndeclaredThrowable}).
+    */
+   @ExceptionHandler(SecurityException.class)
+   @ResponseBody
+   @ApiResponses({
+      @ApiResponse(
+         responseCode = "403",
+         description = "Access was denied because the requested action is not enabled or permitted.")
+   })
+   public ResponseEntity<GenericError> handleAccessDenied(SecurityException e) {
+      return accessDenied(e);
+   }
+
+   /**
     * Builds a sanitized 403 response for an authorization denial. The exception is logged at
     * debug level and the client receives a generic, localized message so that no principal,
     * role, group, or organization details are exposed.
--- a/core/src/main/java/inetsoft/web/admin/AdminExceptionHandler.java
+++ b/core/src/main/java/inetsoft/web/admin/AdminExceptionHandler.java
@@ -162,7 +162,7 @@ public class AdminExceptionHandler {
          responseCode = "403",
          description = "Access was denied because the requested action is not enabled or permitted.")
    })
-   public ResponseEntity<GenericError> handleAccessDenied(SecurityException e) {
+   public ResponseEntity<GenericError> handleCheckedAccessDenied(SecurityException e) {
       return accessDenied(e);
    }
 
--- a/core/src/main/java/inetsoft/web/admin/ai/cluster/ClusterChangesetApplyService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/cluster/ClusterChangesetApplyService.java
@@ -17,6 +17,7 @@
  */
 package inetsoft.web.admin.ai.cluster;
 
+import inetsoft.sree.security.SecurityException;
 import inetsoft.util.Tool;
 import inetsoft.util.audit.*;
 import inetsoft.web.admin.ai.AdminChangesetApplyService;
@@ -142,6 +143,7 @@ public class ClusterChangesetApplyService {
 
    private void applyOne(String txId, String task, PlanChange change, ClusterChangeRequest original,
                          String reviewOutcome, Principal user, List<ClusterApplyOutcome> results)
+      throws SecurityException
    {
       String server = change.property();
       String verb = ClusterChangePlanService.requireVerb("apply", original.getVerb());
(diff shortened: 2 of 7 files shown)


larryliang-inetsoft:
## Review verdict: CHANGES REQUESTED

**Blocking - CI `build` fails with a genuine compile error, not a flake.**

```
[ERROR] core/src/main/java/inetsoft/web/admin/ai/cluster/ClusterChangesetApplyService.java:[156,37]
        unreported exception inetsoft.sree.security.SecurityException; must be caught or declared to be thrown
[ERROR] core/src/main/java/inetsoft/web/admin/ai/cluster/ClusterChangesetApplyService.java:[159,38]
        unreported exception inetsoft.sree.security.SecurityException; must be caught or declared to be thrown
```

Root cause: `ClusterChangesetApplyService` doesn't exist at this PR's recorded base SHA, but landed on `stylebi-wiz-main-rebase` via PR #4823 (merged 2026-08-27, after this branch's diagnosis/refute ran their "no other callers of pauseServers/resumeServers" check - true at the time). `ClusterChangesetApplyService.applyOne()` (lines 156/159) calls `clusterService.pauseServers(...)`/`resumeServers(...)` directly with no `throws`/catch around those calls, so this PR's new checked `SecurityException` on those two methods breaks it. PR #4823's own commit message even anticipates this ("Enforces cluster.pause.enabled Java-side ... stylebi#76342 workaround") - this is a real, expected collision between two concurrently-developed fixes, not a coincidence.

**Fix needed**: rebase onto current `stylebi-wiz-main-rebase` and add `throws SecurityException` to `ClusterChangesetApplyService.applyOne(...)`'s signature - its outer `catch(Exception e)` around the call site already handles the checked exception correctly at runtime once declared, so this is a one-line signature fix, not a logic change. Please confirm `ClusterChangesetApplyServiceTest` still passes after the rebase.

**Everything else checked out:**
- Guard placement (both pause and resume, first statement in each method, before any `ServerClusterClient` construction) - correct.
- The new `@ExceptionHandler(SecurityException.class)` in `AdminExceptionHandler` is correct and unambiguous - verified `inetsoft.sree.security.SecurityException extends Exception` directly (not `java.lang.SecurityException`), so there's no overlap with the existing unchecked-`java.lang.SecurityException` handler or the AOP-wrapped `UndeclaredThrowableException` handler. Minor non-blocking nit: the new handler reuses the method name `handleAccessDenied` as an overload of the existing one - legal and compiles, but a distinct name would read clearer.
- `ClusterServiceTest` (new, 4 tests, `@Tag("core")` present) gives real coverage: asserts the guard fires *before* `new ServerClusterClient()` is even constructed, and separately that the enabled path still calls `pauseServer`/`resumeServer` (no regression).
- No scope creep into #76343 - the per-server success/failure discard logic (`LOG.warn` block) is untouched.

(The `claude-review` CI check also shows failed, but its log shows an SDK-level crash after 1 turn/354ms before doing any actual review - tooling issue, not a code signal, not treated as blocking.)

Full write-up: `docs/teams/2026-08-29-bugs-c5-scoping-followups/bug-76342/04-review-r1.md`

larryliang-inetsoft:
## Review round 2 — APPROVED

Verified the round-1 fix resolves the blocking finding and introduces nothing new:

- **`build` is now genuinely green** (20m37s full run). The round-1 blocking compile break from concurrently-merged #4823's `ClusterChangesetApplyService.applyOne()` is fixed.
- **Checked-exception type verified correct**: pulled the actual diff — `import inetsoft.sree.security.SecurityException;` is present in `ClusterChangesetApplyService.java` immediately above the existing imports, and `applyOne(...) throws SecurityException` resolves to that checked import, not `java.lang.SecurityException`. The fixer's claim about this risk is real and correctly handled.
- **Test signature edits are purely mechanical**: diffed all 4 `ClusterChangesetApplyServiceTest` methods and the 2 `ClusterControllerTest` methods — each change is exactly an appended `throws Exception`/`throws SecurityException`, zero body/assertion changes.
- **`handleAccessDenied` → `handleCheckedAccessDenied` rename is safe**: no test file references either overload by name, and Spring `@ExceptionHandler` dispatch is by annotation + parameter type, not by method name, so the rename cannot affect resolution.
- **No scope creep**: exactly the 7 expected files changed (4 original + 3 round-1), no unrelated hunks.
- **`mergeStateStatus: UNSTABLE`** is explained entirely by `claude-review`'s failure, which is the same known SDK/tooling crash-before-review pattern seen in round 1 (aborts with `is_error:true` before analyzing the diff) — non-blocking per this program's known automated-review-bot outage.

Full writeup: `docs/teams/2026-08-29-bugs-c5-scoping-followups/bug-76342/06-review-r2.md` in stylebi-wiz.

No merge authority here — flagging as ready for the lead to merge.

PR #4873, opened 2026-08-29
Bug #76335, guard getAuthenticationProvider against unknown provider name

larryliang-inetsoft:
## Summary
- `AuthenticationProviderService.getAuthenticationProvider(name)` fed a `null` provider (returned by `getProviderByName` on a miss) into the switch's `case null, default` arm, which called `AuthenticationProviderModel.Builder.customProviderModel(null, mapper)` and threw an unhandled `NullPointerException` (`provider.writeConfiguration(mapper)`) instead of a clean, caller-friendly error.
- Add a null check immediately after resolving `selectedProvider`, throwing `MessageException` with the existing generic `em.security.providerNotFound` catalog key ("Could not find provider") — mirroring the identical guard already present in the sibling `AuthorizationProviderService.getAuthorizationProvider`.
- Root cause and fix independently diagnosed and refuted twice against real source with a reproduced NPE stack trace; see `docs/teams/2026-08-29-bugs-c5-scoping-followups/bug-76335/` in the `stylebi-wiz` repo for the full analysis.

## Test plan
- [x] Added `AuthenticationProviderServiceTest.getAuthenticationProvider_unknownName_throwsMessageExceptionNotNPE`, which mocks an empty authentication chain and asserts `getAuthenticationProvider("does-not-exist")` throws `MessageException` (fails with NPE against the pre-fix code, passes after the fix).
- [x] `./mvnw -o -pl core -am test -Dtest=AuthenticationProviderServiceTest,AuthenticationProviderControllerTest -Dsurefire.failIfNoSpecifiedTests=false` — 7 tests, 0 failures, 0 errors.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/security/AuthenticationProviderService.java
+++ b/core/src/main/java/inetsoft/web/admin/security/AuthenticationProviderService.java
@@ -108,6 +108,11 @@ public class AuthenticationProviderService extends BaseSubscribeChangeHandler {
    public AuthenticationProviderModel getAuthenticationProvider(String name) {
       boolean enterprise = LicenseManager.isEnterprise();
       AuthenticationProvider selectedProvider = getProviderByName(name);
+
+      if(selectedProvider == null) {
+         throw new MessageException(Catalog.getCatalog().getString("em.security.providerNotFound"));
+      }
+
       AuthenticationProviderModel.Builder builder = AuthenticationProviderModel.builder()
          .providerName(name)
          .oldName(name)
--- a/core/src/test/java/inetsoft/web/admin/security/AuthenticationProviderServiceTest.java
+++ b/core/src/test/java/inetsoft/web/admin/security/AuthenticationProviderServiceTest.java
@@ -1,0 +1,64 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft:
**Review verdict: approved**

Verified against the actual diff (`gh pr diff 4873`), not just the archive summary.

- **Fix is the root cause fix, correctly scoped.** The null-check on `selectedProvider` is inserted immediately after `getProviderByName(name)` resolves and before any builder work or the `switch(selectedProvider)` — this is exactly where the diagnosis/refute docs pinpointed the unguarded `null` reaching `AuthenticationProviderModel.Builder.customProviderModel(null, mapper)` → NPE at `AuthenticationProviderModel.java:147`. It mirrors the sibling `AuthorizationProviderService.getAuthorizationProvider`'s existing guard exactly, reusing the same generic, already-translated catalog key `em.security.providerNotFound` ("Could not find provider") rather than inventing a new i18n string.
- **No scope creep.** Diff touches only `getAuthenticationProvider` in `AuthenticationProviderService.java` (+5 lines) and adds one new test file. `getProviderFromModel` (bug #76336, same file, being fixed separately) is untouched. The pre-existing `case null, default ->` switch arm is left in place as a defensive default for non-null-but-unrecognized provider types, which is correct — not required to remove for this fix.
- **No import changes needed and none required** — `import inetsoft.util.*;` was already present, covering both `Catalog` and `MessageException`.
- **Test coverage is real, not shallow.** `AuthenticationProviderServiceTest.getAuthenticationProvider_unknownName_throwsMessageExceptionNotNPE` mocks an empty authentication chain (`authenticationChain.stream()` → `Stream.empty()`) and asserts the specific exception type change: `assertThrows(MessageException.class, ...)`. This directly encodes the bug (NPE → MessageException) — it was verified to fail against pre-fix code and pass after, per the fix doc's own build log (7/7 tests passing, no regression in the existing `AuthenticationProviderControllerTest`).
- **No leftover debug code.** Diff is clean — just the guard clause and the new test.
- **CI**: only `license/cla` check configured for this repo/branch, and it passes. No other automated checks are wired up to run here.

No blockers, no important findings ≥8/10 confidence. Approving.

larryliang-inetsoft:
Closing without merging: this duplicates already-merged PR #4847 ("Fix NPE in AuthenticationProviderService.getAuthenticationProvider on unknown name"), which fixes Redmine bug #76358 with an identical null-guard fix (same file, same method, same catalog-key approach, plus additional coverage of a delete-race call site in ProviderChangesetApplyService that this PR doesn't touch). #76335 and #76358 are the same underlying defect, filed independently. Confirmed via `git log` on `stylebi-wiz-main-rebase` and PR #4847's own description. No functional loss — the fix is already live on mainline.

PR #4874, opened 2026-08-29
Fix bug #76343, pauseServers/resumeServers discard per-server pause/resume result

larryliang-inetsoft:
## Root cause

`ClusterService.pauseServers`/`resumeServers` were `void`, discarding
`ServerClusterClient.pauseServer`/`resumeServer`'s per-server success boolean (only a bare
`LOG.warn`, no aggregation, no return value). `ClusterController.pauseServer`/`resumeServer`
were also `void` REST endpoints delegating straight through, so `POST
/api/em/monitoring/cluster/pause-server` (or `resume-server`) always returned HTTP 200 with an
empty body regardless of whether 0 or all N requested servers actually paused/resumed.

Rebased onto post-#76342 mainline (#4872, `isPauseEnabled()` guard) before making this change;
this fix only changes what happens after that guard passes.

## Fix

- `ClusterService.pauseServers`/`resumeServers` now return `Map<String, Boolean>` (server name
  to success), reporting `true` for a server already in the desired paused/resumed state (not
  omitted) so the caller can do exact input/output reconciliation. `LOG.warn` now names the
  server on failure.
- `ClusterController.pauseServer`/`resumeServer` now return that map instead of `void`. HTTP
  200 stays the status for "request was processed, here is the per-server outcome" -- a mixed
  true/false map is a legitimate outcome, not an exception-worthy one (that's #76342's
  `SecurityException`, for a fundamentally disallowed request).
- EM cluster monitoring page (`cluster-monitoring-page.component.ts`): the previously
  fire-and-forget `pauseServers()`/`resumeServers()` now read the response and surface any
  failed server(s) via the existing snackbar pattern.
- Added `em.cluster.pauseResume.failed` to the base `srinter.properties` bundle.

## Out of scope (left untouched)

`ClusterChangesetApplyService.applyOne`/`readBackWithRetry` -- the admin-chat wrapper (#4823)
already works around this exact gap with its own read-back-with-retry mechanism, referenced by
name in its class javadoc. That mechanism has an independent justification (its own read-back
races a ~50ms debounced status-map write, which this fix's synchronous return value doesn't
address) and may still be the more reliable choice even after this fix lands. Whether to
simplify it to consume this fix's return value is a separate design call, not resolved here.

## Test plan

- [x] `ClusterServiceTest` (8 tests: 4 pre-existing from #76342 + 4 new `[G5]`-`[G8]` covering
      the new `Map<String, Boolean>` return, including the already-satisfied `true` case)
- [x] `ClusterControllerTest` (6 tests: existing `[G4]`/`[G5]` extended to assert pass-through
      of the returned map, matching the `[G2]`/`[G3]` pattern)
- [x] `ClusterChangesetApplyServiceTest` (10 tests, unchanged -- confirmed its `doAnswer`/
      `doThrow` mocks compile and pass unmodified against the new non-void return type)
- [ ] Frontend change reviewed by hand against existing patterns in the file; not run through
      Karma/tsc (no `node_modules` in this worktree) -- flagging for reviewer

```
Tests run: 8, Failures: 0 -- ClusterServiceTest
Tests run: 6, Failures: 0 -- ClusterControllerTest
Tests run: 10, Failures: 0 -- ClusterChangesetApplyServiceTest
```

Full root-cause/diagnosis/refute/fix writeup:
`stylebi-wiz` repo, `docs/teams/2026-08-29-bugs-c5-scoping-followups/bug-76343/`.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/cluster/ClusterController.java
+++ b/core/src/main/java/inetsoft/web/admin/cluster/ClusterController.java
@@ -121,8 +121,8 @@ public class ClusterController {
       )
    )
    @PostMapping("/api/em/monitoring/cluster/pause-server")
-   public void pauseServer(@RequestBody String[] servers) throws SecurityException {
-      clusterService.pauseServers(servers);
+   public Map<String, Boolean> pauseServer(@RequestBody String[] servers) throws SecurityException {
+      return clusterService.pauseServers(servers);
    }
 
    @Secured(
@@ -133,8 +133,8 @@ public class ClusterController {
       )
    )
    @PostMapping("/api/em/monitoring/cluster/resume-server")
-   public void resumeServer(@RequestBody String[] servers) throws SecurityException {
-      clusterService.resumeServers(servers);
+   public Map<String, Boolean> resumeServer(@RequestBody String[] servers) throws SecurityException {
+      return clusterService.resumeServers(servers);
    }
 
    private final ClusterService clusterService;
--- a/core/src/main/java/inetsoft/web/admin/cluster/ClusterService.java
+++ b/core/src/main/java/inetsoft/web/admin/cluster/ClusterService.java
@@ -98,44 +98,54 @@ public class ClusterService extends MonitorLevelService implements StatusUpdater
          .build();
    }
 
-   public void pauseServers(String[] servers) throws SecurityException {
+   public Map<String, Boolean> pauseServers(String[] servers) throws SecurityException {
       if(!isPauseEnabled()) {
          throw new SecurityException("Cluster pause is not enabled");
       }
 
       ServerClusterClient client = new ServerClusterClient();
+      Map<String, Boolean> results = new LinkedHashMap<>();
 
       for(String server : servers) {
          if(client.getStatus(server).isPaused()) {
+            results.put(server, true);
             continue;
          }
 
          boolean success = client.pauseServer(server);
+         results.put(server, success);
 
          if(!success) {
-            LOG.warn("Failed to pause server");
+            LOG.warn("Failed to pause server {}", server);
          }
       }
+
+      return results;
    }
 
-   public void resumeServers(String[] servers) throws SecurityException {
+   public Map<String, Boolean> resumeServers(String[] servers) throws SecurityException {
       if(!isPauseEnabled()) {
          throw new SecurityException("Cluster pause is not enabled");
       }
 
       ServerClusterClient client = new ServerClusterClient();
+      Map<String, Boolean> results = new LinkedHashMap<>();
 
       for(String server : servers) {
          if(!client.getStatus(server).isPaused()) {
+            results.put(server, true);
             continue;
          }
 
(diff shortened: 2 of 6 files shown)


larryliang-inetsoft:
## Review: approved

Verified the actual diff (`gh pr diff 4874`), not just the fix archive's description.

**Map<String,Boolean> correctness** — confirmed every requested server appears exactly once: the `continue` branches (already-paused / already-resumed) now `results.put(server, true)` before `continue`, and the attempted-call branch always `results.put(server, success)` regardless of outcome. No server can be omitted from the returned map.

**Coexistence with #76342's guard** — confirmed directly from the diff context: `if(!isPauseEnabled()) { throw new SecurityException(...) }` appears as unchanged context (not part of this diff), sitting before the loop in both methods. This fix only touches behavior after the guard passes, as intended.

**ClusterChangesetApplyService scope** — confirmed untouched (not in the PR's file list). Read `ClusterChangesetApplyServiceTest.java` directly in the fix's worktree: every `pauseServers`/`resumeServers` stub uses `doAnswer(...)`/`doThrow(...)`, none use `doNothing()` — both styles compile and behave identically against a method whose return type changed from `void` to `Map<String,Boolean>` (a `doAnswer` lambda returning `null` is a valid stub for any reference return type). Claim verified, not just trusted.

**i18n fallback** — `em.cluster.pauseResume.failed` added to the base `srinter.properties` only. Confirmed this repo's locale bundles (`srinter_{fr_FR,ja_JP,zh_CN}.properties`) are loaded via `ResourceBundle.getBundle(resource, locale)` (`Catalog.java`), which is standard `java.util.ResourceBundle` fallback behavior — a locale bundle missing a key falls through to the base bundle. Not a real gap.

**Frontend change** — reviewed by hand against the same file's existing pattern (the `em.cluster.pause.last` snackbar a few lines above): `notifyFailedServers` correctly reuses `MatSnackBar`/`Tool.SNACKBAR_DURATION`/the `_#(js:...)` catalog-key convention, and `Record<string, boolean>` is a correct client-side shape for a Jackson-serialized `Map<String, Boolean>`. No issues found.

**Test coverage** — real. New `ClusterServiceTest` `[G5]`-`[G8]` assert actual map contents (`assertEquals(Map.of(...), result)`) and use `verify(..., never())` to confirm the already-satisfied branch skips the client call. Updated `ClusterControllerTest` `[G4]`/`[G5]` assert `assertSame(expected, result)` on the pass-through, matching the existing `[G2]`/`[G3]` pattern.

**Duplicate-fix / scope creep check** — ran `git log <mainline> -- <each touched file>` myself on all 6 touched files; no other commit changes the pause/resume return-type contract, and the touched-file set matches exactly what the fix describes.

**CI** — Semgrep OSS/Scan pass, CLA pass, `build` was pending at review time. The `claude-review` bot check failed, but its log shows an infra failure (empty `ANTHROPIC_API_KEY`, `is_error:true` on turn 1, no actual review content produced) — not a substantive finding, safe to disregard/rerun.

No blockers found.

PR #4880, opened 2026-08-29
Fix Bug #76304: set_print_layout's paperSize accepts non-canonical values silently

larryliang-inetsoft:
## Summary
- `PrintDeviceLayoutPropertyService.applyPrintLayoutPatch`'s `"paperSize"` case stored the incoming value verbatim with no lookup against `PaperSize`'s 14-entry canonical name table. A bare short name like `"Letter"` (the plugin tool's own documented example) persisted as-is instead of `"Letter [8.5x11 in]"`.
- Composer's Page Layout dialog binds its Paper Size `<select>` directly against the canonical option values (plain `[value]`-bound native select, no `ngValue`/`compareWith`), so a non-canonical stored value renders as a blank selection.
- Added `PaperSize.canonicalize(String)`: passes through an already-canonical name or the `"(Custom Size)"` sentinel unchanged, resolves a bare short name case-insensitively to its canonical form, and throws `IllegalArgumentException` naming the field and value for anything else.
- `applyPrintLayoutPatch`'s `"paperSize"` case now calls `PaperSize.canonicalize(asString(value))` instead of storing the raw string.

## Test plan
- [x] `./mvnw.cmd -pl core -am test -Dtest=PrintDeviceLayoutPropertyServiceTest -Dsurefire.failIfNoSpecifiedTests=false` → `Tests run: 19, Failures: 0, Errors: 0, Skipped: 0`, `BUILD SUCCESS`
- Updated the existing test that pinned the buggy bare-string behavior, and added 4 new tests: bare-short-name case-insensitive canonicalization, already-canonical passthrough, `"(Custom Size)"` sentinel passthrough, and unrecognized-value throws (asserting no partial write).

### DIFF
--- a/core/src/main/java/inetsoft/report/internal/PaperSize.java
+++ b/core/src/main/java/inetsoft/report/internal/PaperSize.java
@@ -129,6 +129,49 @@ public class PaperSize {
          StyleConstants.PORTRAIT;
    }
 
+   /**
+    * Composer's Page Layout dialog's own sentinel for "use customWidth/customHeight instead"
+    * (viewsheet-print-layout-dialog.component.ts). Not one of the 14 predefined sizes, so it must
+    * pass through {@link #canonicalize(String)} unchanged rather than fail the lookup below.
+    */
+   public static final String CUSTOM_SIZE = "(Custom Size)";
+
+   /**
+    * Resolves a paper size name to its canonical, predefined-list form, accepting the bare short
+    * name (e.g. "Letter") as a caller-friendly alias for the full canonical string (e.g.
+    * "Letter [8.5x11 in]") since the two are otherwise indistinguishable to a caller. An already-
+    * canonical string or {@link #CUSTOM_SIZE} is returned unchanged. Anything else is rejected
+    * rather than stored verbatim, since a non-canonical value stored via {@code setPaperSize}
+    * cannot round-trip through this class's own predefined-list lookups or Composer's Page Layout
+    * dialog, whose paper-size select is bound directly against this list.
+    *
+    * @throws IllegalArgumentException if {@code name} is null or matches neither a canonical name,
+    * a bare short name, nor {@link #CUSTOM_SIZE}.
+    */
+   public static String canonicalize(String name) {
+      if(name == null) {
+         throw new IllegalArgumentException("paperSize is required.");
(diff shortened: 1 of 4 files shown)


larryliang-inetsoft:
## Round 2: CI failure follow-up (LayoutReadServiceTest)

The full CI build for this PR failed on a test the original fix didn't touch:

```
LayoutReadServiceTest.getRoundTripsWhatSetPrintLayoutActuallyWrote:238
expected: <Letter> but was: <Letter [8.5x11 in]>
```

**Root cause**: `getRoundTripsWhatSetPrintLayoutActuallyWrote` wires the real
`PrintDeviceLayoutPropertyService` (write) against the real `LayoutReadService`
(read), writes `paperSize: "Letter"` as input, and asserts the read-back value.
Since `applyPrintLayoutPatch` now canonicalizes on write, the round-tripped
value is correctly `"Letter [8.5x11 in]"` -- the test's assertion still expected
the old pass-through value. Same class of "test pinned the bug as correct
behavior" issue the original fix already corrected once in
`PrintDeviceLayoutPropertyServiceTest`, just missed here because it's a
different test class.

**Fix**: updated the stale assertion in
`LayoutReadServiceTest.java:238` from `"Letter"` to `"Letter [8.5x11 in]"`.
No production code changed.

**Sweep**: grepped the whole `core` module for any other test asserting a
bare canonical-family paperSize (`Letter`/`Legal`/`A4`/etc.) as an expected
round-trip result. Only two files reference `PaperSize` in tests at all
(`LayoutReadServiceTest`, `PrintDeviceLayoutPropertyServiceTest`); every
other bare-string occurrence in those files is either (a) set directly on a
model object outside the patch/canonicalize path (correct, untouched-field
read-merge-write cases) or (b) already updated by the original fix. No third
hidden instance found.

**Verification**: ran the full `core` module suite (`./mvnw.cmd -pl core -am
test`, no `-Dtest` filter) -- 8066 tests, 0 failures, 0 errors, 69 skipped,
BUILD SUCCESS (~9 min). Also ran `LayoutReadServiceTest` +
`PrintDeviceLayoutPropertyServiceTest` together beforehand (26/26 passed).

Pushed as `7d1d564922` on `fix/bug-76304`.

PR #4895, opened 2026-08-30
Fix bug corpus #76350 follow-on item E: BindableColumns.require() empty-listing bypass

larryliang-inetsoft:
## Root cause

`BindableColumns.require()` had the same over-broad empty-listing bypass that PCB-002 (#4893, merged) already fixed on its sibling method `requireSource()`:

```java
if(available.isEmpty()) {
   return;
}
```

Both of `require()`'s two callers (`BindingAgentController.requireBindableColumns` and `.resolveSourceTable`) already wrap the listing read and `require()` itself in a try/catch that converts a genuine read failure into a logged skip one layer up. `resolveSourceTable` calls `require()` and `requireSource()` back-to-back in the *same* try block on the *same* `tables` value — so by the time `require()` sees `tables`, `fieldsService.list()` has already returned successfully. An empty `available` can therefore only mean a successful-but-empty listing, never an unreadable tree.

Worse than `requireSource()`'s pre-fix bug: when an assembly already has a current source but that one table's own column list came back empty (while the overall listing is non-empty), the bypass waved through *any* field name at all — including one belonging to a **different table** in the same listing. This was empirically demonstrated (binding `PRODUCT_NAME`, a real column of `Products`, to a chart whose current source `Query1` has an empty field list — accepted silently pre-fix).

## Fix

Delete the bypass outright (`core/src/main/java/inetsoft/web/wiz/binding/BindableColumns.java`). No replacement condition needed:
- `require()` already has a separate, correct "nothing being written" bypass (`fields.isEmpty()`) at the top of the method.
- Blank/absent column names are already skipped per-field inside the validation loop regardless of `available`.

## Tests

- `BindableColumnsTest.staysOutOfTheWayWhenNothingCouldBeListed` renamed to `refusesWhenNothingCouldBeListed` and flipped from `assertDoesNotThrow` to `assertThrows`, mirroring PCB-002's own `requireSource()` test shape.
- Added `refusesAColumnFromAnotherTableWhenTheCurrentTablesOwnListingIsEmpty`, a permanent regression test for the cross-table silent-bind scenario.
- Verified no other existing `require()` test depends on the old bypass (all use non-empty fixtures).

`./mvnw -pl core -o test -Dtest=BindableColumnsTest` — 22/22 passed.

## Provenance

Diagnosed and refuted in `stylebi-wiz` under `docs/teams/2026-08-30-bugs-corpus-followon/item-E/` (`01-diagnosis.md`, `02-refute.md`) — a follow-on item flagged during PCB-002's own fix (#4893). Refuter verdict: survives fully, no corrections needed.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindableColumns.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindableColumns.java
@@ -79,12 +79,6 @@ public final class BindableColumns {
          }
       }
 
-      // No listing at all means the tree could not be read — a different failure, and refusing
-      // every column on the strength of it would turn a read problem into a write problem.
-      if(available.isEmpty()) {
-         return;
-      }
-
       for(FieldRef field : fields) {
          if(field == null || field.column() == null || field.column().isBlank()) {
             continue;
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
@@ -725,9 +725,17 @@ public class BindingAgentController {
       try {
          List<BindableTable> tables =
             fieldsService.list(sessions.runtimeId(sessionToken, user), assembly, user);
+
+         // requireSource runs first: when it has something specific to say about the source
+         // itself -- a stated table the listing does not have, an ambiguous inference, or a
+         // conflicting repoint -- that is the more actionable diagnosis, and require()'s column
+         // check would otherwise win the race and report a bare "column not found" that drops
+         // the table name entirely. When requireSource does not throw (source already known, or
+         // resolved), require() still runs right after to do its own, narrower column check.
+         String source = BindableColumns.requireSource(tables, assembly, table, fields);
          BindableColumns.require(tables, assembly, fields);
 
-         return BindableColumns.requireSource(tables, assembly, table, fields);
+         return source;
       }
       catch(IllegalArgumentException e) {
          throw e;
--- a/core/src/test/java/inetsoft/web/wiz/binding/BindableColumnsTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/binding/BindableColumnsTest.java
@@ -306,13 +306,47 @@ class BindableColumnsTest {
    }
 
    /**
-    * An empty listing means the binding tree could not be read — a different failure. Refusing
-    * every column on the strength of it would turn a read problem into a write problem.
+    * An empty {@code tables} reaching {@code require} does NOT mean "the tree could not be read"
+    * (bug #76350, PCB-002's own reasoning for {@code requireSource}, which applies identically
+    * here — see {@code requireBindableColumns}/{@code resolveSourceTable}'s shared try/catch). A
+    * genuine read failure is already caught one layer up, before this method ever runs — so by the
+    * time it sees an empty list, the listing succeeded and genuinely found nothing bindable. That
+    * case must refuse, not silently accept the bind and crash later on render.
     */
    @Test
-   void staysOutOfTheWayWhenNothingCouldBeListed() {
-      assertDoesNotThrow(() -> BindableColumns.require(List.of(), "Crosstab1", dim("ANYTHING")));
-      assertDoesNotThrow(() -> BindableColumns.require(null, "Crosstab1", dim("ANYTHING")));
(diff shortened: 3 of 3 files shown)


larryliang-inetsoft:
**Review verdict: changes requested (self-review blocked, posting as comment instead).**

Blocking: the `build` check fails with 3 real test failures in `core/src/test/java/inetsoft/web/wiz/binding/BindingAgentControllerTest.java`:

- `setChartShelfRefusesATableTheListingDoesNotHaveInsteadOfSilentlyDroppingTheSource:278`
- `setChartSingleShelfRefusesATableTheListingDoesNotHave:305`
- `setAestheticFieldRefusesATableTheListingDoesNotHave:332`

(CI run: https://github.com/inetsoft-technology/stylebi/actions/runs/33318570186/job/99276379635 — `Tests run: 8165, Failures: 3, Errors: 0`, exactly these three.)

All three are PCB-002's own controller-level regression tests (already merged in the sibling fix, #4893) for the "empty listing" scenario this PR also targets. Each mocks `fields.list(...)` to return `List.of()` for a request naming source table `"Sales"`, then asserts the thrown message `.contains("Sales")` — i.e. that `requireSource()` is what fires and names the missing *source table*.

`resolveSourceTable` calls `BindableColumns.require(tables, assembly, fields)` immediately before `BindableColumns.requireSource(tables, assembly, table, fields)` in the same try block. With `require()`'s empty-listing bypass now removed, `require()` throws first for this exact case — but its message names the *column* (`"Region"`/`"Price"`), not the table (`currentSourceOf(List.of())` is null, so the `"which is bound to 'Sales'"` clause never appears). `requireSource()` never runs, so `.contains("Sales")` fails for all three.

The write is still refused either way — no silent accept survives — but which validator fires first changed, and with it the diagnostic content of the message a calling agent sees ("Region isn't bindable" vs. "Sales isn't a real source table"). PCB-002 specifically tested for the latter in this scenario. This needs a decision, not a blind test-string patch: either narrow `require()`'s fix so `requireSource()` still gets first crack at naming the missing table when nothing at all was listed, or deliberately update these three tests to assert the new message with a stated reason it's still correct — not just edit them to turn CI green.

Everything else in the diagnosis/refute/fix/verify chain checks out (bypass deleted with no replacement, new/renamed `BindableColumnsTest` cases are real and non-tautological, caller trace is accurate) — this is scoped to the one regression above.

larryliang-inetsoft:
Addressed the regression from round 1 review (3 failing \`BindingAgentControllerTest\` tests).

**Root cause**: \`resolveSourceTable\` calls \`require()\` immediately before \`requireSource()\` in the same try block. With \`require()\`'s empty-listing bypass removed, \`require()\` now throws first for an empty \`tables\` listing — but its message names the *column* (\`currentSourceOf(List.of())\` is \`null\`), not the requested source table, so the 3 tests' \`.contains("Sales")\` assertions failed. \`requireSource()\` — whose job is naming the missing/invalid table — never got to run.

**Fix (option 1 from the review, not a test patch)**: reordered the two calls in \`resolveSourceTable\` so \`requireSource()\` runs first. Traced every branch of both methods to confirm this is safe:
- already-sourced / nothing-to-establish: \`requireSource()\` returns \`null\` immediately, \`require()\` still runs right after unchanged.
- repoint conflict: \`requireSource()\` throws its own message; no test expects \`require()\` to win that race.
- resolved via requested table or via inference: \`requireSource()\`'s own \`requireHasAll\`/\`hasAll\` already guarantee the resolved table's columns satisfy every named field, so \`require()\`'s subsequent lenient check is a guaranteed-safe no-op.
- empty/lacking listing with a table requested: \`requireSource()\`'s \`matchListed\` now throws first, naming the table — this is the fix for the regression.
- \`requireBindableColumns\` (the other caller) doesn't call \`requireSource()\` at all and is untouched.

Chose this over patching the 3 tests to accept the new message because the column-only message is genuinely worse here: it drops the requested table name (\`"Sales"\`) entirely, while \`requireSource()\`'s message directly answers "why did my table request fail" with the actual bad input.

**Verified**: full \`inetsoft.web.wiz.binding\` package (495/495, \`BindingAgentControllerTest\` 11/11) and the broader \`inetsoft.web.wiz.**\` package (2709/2709, 1 pre-existing unrelated skip).

Details in \`stylebi-wiz\` repo: \`docs/teams/2026-08-30-bugs-corpus-followon/item-E/03-fix.md\`.

PR #4892, opened 2026-08-30
Fix Cluster C viewsheet/worksheet lifecycle bugs (PVA-001/002/003/004/005/010)

larryliang-inetsoft:
## Summary

Fixes for the "viewsheet assembly creation & save lifecycle" cluster (Cluster C) from the
plugin-composer bug-reports corpus (Redmine #76350), per
`docs/teams/2026-08-29-bugs-plugin-composer-corpus/cluster-C/01-diagnosis.md` and
`02-refute.md` in the stylebi-wiz repo.

Companion plugin-side PR (add_assembly alias, save_viewsheet name/scope, edit op:"refresh"):
stylebi-wiz#TBD.

**PVA-008 and PVA-009 are excluded from this PR** — see below.

## PVA-004 — `edit op:"add"` silently drops `assembly`/`width`/`height`

- `ComposerObjectService.addNewObject`'s return type widened `Void` → `String` (the new
  assembly's `getAbsoluteName()`). Precedent: `VSBindingService.commit` already uses the same
  `@ClusterProxyMethod`+`@ClusterWriteMethod` pair with a `String` return.
- `ViewsheetEditService.add()` now chains a rename (via `VSObjectPropertyService.editObjectProperty`)
  when `assembly` is supplied, and a resize (via `ComposerObjectService.resizeObject`) when
  `width`/`height` are supplied — seeding the resize event's `xOffset`/`yOffset` from
  `request.x()`/`request.y()` (the position `add()` already placed the assembly at), avoiding the
  "silently teleports to 0,0" pitfall `resize()`'s own wrapper already guards against.
- A plain `add` with no name/size hints stays a single mutate/undo step (no chained calls).

## PVA-005 — `add_table` collision-rename over-triggers unconditionally

Root cause: `AssetUtil.getNextName(ws, prefix)` (2-arg) hardcodes `preferred=null`, so the "keep
name if free" short-circuit in the 3-arg overload can never fire — every `add_table` got an
unconditional `1`-suffixed name, even with zero collision.

Fixed at **all three** call sites in `WorksheetAgentController` carrying the identical bug —
`addBoundTable`, `addLogicalModelTable`, `addTabularTable` — not just the one call site the
corpus doc's own repro happened to exercise.

## PVA-001 — no save-as for a first-time viewsheet

`ViewsheetAssemblyAgentController.save()` took no request body and refused unconditionally when
the entry was `TEMPORARY_SCOPE`. Added a `SaveRequest(name, scope)` body, mirroring
[...]

PR #4900, opened 2026-08-30
Fix Bug 76332 (PVA-007): add attach_base_worksheet for a baseless viewsheet

larryliang-inetsoft:
No `mcp__composer-chat__*` tool could attach an existing worksheet asset as a baseless viewsheet's base, or create a viewsheet asset at all. This closes the **attach** half (the **create** half is a separate, longer-term project — a `create_viewsheet`-shaped tool, out of scope here).

New endpoint on `ViewsheetAssemblyAgentController`:
`POST /api/wiz/v1/agent/viewsheet/{sessionToken}/attach-base-worksheet`

## Root cause

Full archive: `docs/teams/2026-08-30-bug-76332/` (stylebi-wiz repo) — a `/deep-debug-workflow` run with two independent, adversarial investigators.

`Viewsheet` tracks its base worksheet in two fields — `wentry` (public `setBaseEntry`/`getBaseEntry`) and `ws` (private, populated only by `update()`/`reloadBaseWorksheet()`/`repopulateWorksheet()`, read by `getBaseWorksheet()`). `list_bindable_fields`'s backing tree-builder (`VSEventUtil.refreshBaseWSTree`) reads the cached `ws` field, not `wentry`, and no code path on this tool surface ever called `setBaseEntry` for a baseless viewsheet at all, let alone the `ws`-populating follow-up a correct attach also needs.

## Fix

Resolve the caller-supplied worksheet path to a permission/existence-checked `AssetEntry` (`AssetRepository.getSheet(entry, principal, true, AssetContent.ALL, false)` — `permission=true`, actually enforcing read access, unlike the superficially similar existence-only probe `ComposerViewsheetService.checkWorksheetChanged` uses for an unrelated freshness check), refuse loudly if the viewsheet already has a base (no silent overwrite/repoint), then `vs.setBaseEntry(entry)` followed by `vs.reloadBaseWorksheet(assetRepository, principal)` — in that order, since `reloadBaseWorksheet` reads `wentry` as an instance field.

Deliberately **not** `update()`: both are traced in full in the archive; `update()`'s extra `clearCache()`/embedded-viewsheet-refresh/`layout()` work is a confirmed no-op for a freshly-attached, assembly-free viewsheet, and its catch-and-return-`false` failure mode is a worse fit for an unattended agent caller than `reloadBaseWorksheet()`'s propagate-the-exception behavior.

Does not persist — like every other mutation on this controller, only the paired session's in-memory `Viewsheet` changes; `save_viewsheet` still owns persistence.

## Tests

7 new regression tests: attach on a baseless viewsheet sets entry then reloads (in that order, verified via `InOrder`); refuses when already based; refuses on a nonexistent/unreadable path; refuses when the permission check itself throws; refuses with no path; permission check is actually enabled (`permission=true`), not skipped.

Full `inetsoft.web.wiz.viewsheet` package (564 tests) and the broader `inetsoft.web.wiz` package (2708 tests) re-run clean.

## What this PR may and may not claim

The mechanism is structurally demonstrated and unit-tested against real (unmocked) `Viewsheet`/`VSEventUtil` code (`docs/teams/2026-08-30-bug-76332/03-repro.md`), but **no live composer-chat/StyleBI session was available this run** for full end-to-end confirmation, and `AssetRepository` permission-check parity for an agent-session `Principal` vs. a live Composer session's was not independently verified. **Not claimed fixed-and-live-confirmed** — claimed fixed-and-unit-tested, live confirmation outstanding.

Companion `stylebi-wiz` PR (plugin-side `attach_base_worksheet` tool): see the linked PR in that repo.

Fixes #76332.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/ViewsheetAssemblyAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/ViewsheetAssemblyAgentController.java
@@ -30,9 +30,11 @@ import org.springframework.web.server.ResponseStatusException;
 
 import inetsoft.analytic.composition.ViewsheetService;
 import inetsoft.report.composition.RuntimeViewsheet;
+import inetsoft.uql.asset.AssetContent;
 import inetsoft.uql.asset.AssetRepository;
 import inetsoft.uql.XPrincipal;
 import inetsoft.uql.asset.AssetEntry;
+import inetsoft.uql.viewsheet.Viewsheet;
 import inetsoft.web.wiz.script.ScriptImageService;
 
 import java.security.Principal;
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
**P6 review verdict: CHANGES REQUESTED** (posting as a comment — `gh pr review -r` is blocked, self-review on this account).

Blocking finding from the /deep-debug-workflow P6 review (docs/teams/2026-08-30-bug-76332/06-review-r1.md):

**`attachBaseWorksheet`'s catch block does not roll back `vs.setBaseEntry(entry)` when `vs.reloadBaseWorksheet(rep, xp)` throws.**

`vs.setBaseEntry(entry)` mutates the live paired `Viewsheet` (plain field write) before `reloadBaseWorksheet` runs its own independent `rep.getSheet(wentry, user, true, AssetContent.ALL)` call. If that second call throws (permission changed, storage error, deserialization failure — none ruled out by the endpoint's earlier probe, which is a separate code path and separate fetch), the catch block converts it to a `PairingException` but leaves `wentry` set and `ws` still null. That is exactly this bug's own `wentry`/`ws` split, now reachable via the fix's own failure path — `list_bindable_fields` on that session will hit the same `ws == null` throw the original bug report opened on. Worse, the endpoint's own already-based guard (`vs.getBaseEntry() != null`) now refuses every subsequent `attach_base_worksheet` call on that session with a message claiming a base is already attached, even though the attach visibly failed — no retry path, no documented recovery.

Not covered by any existing test: `attachBaseWorksheetRefusesWhenPermissionCheckThrows` only exercises the endpoint's own first probe throwing, before `setBaseEntry` is called (`verify(vs, never()).setBaseEntry(any())`). No test exercises `reloadBaseWorksheet` itself throwing after `setBaseEntry` succeeded.

**Requested fix**: in the catch block, call `vs.setBaseEntry(null)` before rethrowing, so a failed attach leaves the session as it was and the caller can retry. Add a regression test asserting `vs.getBaseEntry()` is `null` again after a `reloadBaseWorksheet` failure.

Full review (permission-check re-derivation, overfit check, pane-scoped-gate decision, sibling-call-site sweep) in `docs/teams/2026-08-30-bug-76332/06-review-r1.md`.

fredaan (commented):
## Review round (code-review + attempted live verification)

**Code review: clean, round-1 blocking finding confirmed resolved.** Read the full current diff of `ViewsheetAssemblyAgentController.attachBaseWorksheet` and `ViewsheetAssemblyAgentControllerTest`, plus the full `docs/teams/2026-08-30-bug-76332/` archive (root-cause, fix, verify, and `06-review-r1.md`'s round-1 review).

- The mechanism (`setBaseEntry(entry)` then `reloadBaseWorksheet(rep, xp)`, in that order, not `update()`) matches `02-root-cause.md`'s converged design and is independently sound: `reloadBaseWorksheet` reads `wentry` as an instance field, so the ordering is required, and `update()`'s extra `clearCache()`/embedded-refresh/`layout()` work is correctly argued to be a no-op for a freshly-attached, assembly-free viewsheet.
- The permission-check correction (`getSheet(entry, xp, true, AssetContent.ALL, false)`, not the `permission=false` freshness-probe shape `checkWorksheetChanged` uses) is real and correctly applied — copying the freshness-probe's arguments literally would have silently disabled the permission gate this endpoint needs on a caller-supplied path. Confirmed the `AbstractAssetEngine.getSheet`/`checkAssetPermission` reasoning in the archive is accurate to the description (did not need to re-derive from source myself given three independent prior derivations already did, and the current diff's comment matches).
- **The round-1 blocking finding is fixed in the code currently on this PR**: the catch block around `setBaseEntry`/`reloadBaseWorksheet` now calls `vs.setBaseEntry(null)` before rethrowing, and a new test (`attachBaseWorksheetRollsBackSetBaseEntryWhenReloadBaseWorksheetFails`) asserts the 2-call `setBaseEntry` sequence (real entry, then `null`) — correctly distinct from the pre-existing `attachBaseWorksheetRefusesWhenPermissionCheckThrows` test, which covers a failure *before* `setBaseEntry` is ever reached. This closes the exact "reproduces the bug's own broken state on an unrecoverable session" defect the round-1 review found.
- Test suite otherwise covers the right shapes: already-based refusal, nonexistent/unreadable path, permission-check throwing, no-path/blank-path. No unrelated refactors, no dead code; the double-fetch redundancy the round-1 review flagged as a non-blocking nice-to-have is correctly left alone (would need a new public `Viewsheet` method — out of scope for this fix).

**Live verification: attempted, could not be completed — reporting this plainly rather than fabricating a result.** This round's instructions asked me to use a live StyleBI instance and `mcp__claude-in-chrome__*` / `mcp__plugin_stylebi-composer-chat_composer-chat__*` tools to open a genuinely baseless viewsheet and confirm `attach_base_worksheet` actually gives it a working data source end-to-end — closing the gap every phase of this PR's own archive (P3/P4/P5/P6) explicitly left open. In the actual tool set available to this review invocation, none of those MCP tools are bound (only `Bash`/`Read`/`Grep`/`Glob`), so I could not pair a browser session or exercise the real endpoint. I'm not substituting a source-level or API-level side channel for this, and I'm not claiming a live pass I didn't observe.

**Verdict this round: not merging.** No blocking code defect found, and the previously-identified blocking defect is resolved with a real regression test. What's still outstanding is exactly what every phase of this bug's own archive already discloses as not-yet-claimed: live end-to-end confirmation against a real server, and `AssetRepository` permission-check parity for an agent-session `Principal` vs. a live Composer session's. Recommend re-running this review with the browser/composer-chat MCP tools actually wired into the reviewing session before merging both PRs together (companion `stylebi-wiz`#1972 depends on this one; merge order there would be #4900 first once both are ready).

(`gh pr checks`: `build`/Semgrep OSS/Semgrep Scan/`license/cla` all pass; `claude-review` fails for the pre-existing empty-`ANTHROPIC_API_KEY` infra reason already noted in `06-review-r1.md` §7 — not treated as blocking, consistent with precedent, e.g. #4892 merged with the same failure.)

fredaan (commented):
Live end-to-end verification complete via the companion PR (stylebi-wiz#1972)'s `attach_base_worksheet` tool against a real running instance of this endpoint — see that PR for the full steps. Confirmed: a genuinely baseless viewsheet correctly gained a real data source (ORDERS table, 6 columns, correct dimension/measure typing) after the call, and the response shape matches what the plugin expects.

Code review: clean, and confirmed the round-1 blocking finding (missing `vs.setBaseEntry(null)` rollback when `reloadBaseWorksheet` throws) is fixed in the current diff, with its own regression test.

`build`, Semgrep OSS/Scan, and CLA all pass; `claude-review` fails for the known pre-existing empty-API-key infra reason (non-blocking, consistent with every other PR in this run).

Merging.

PR #4901, opened 2026-08-31
Close the L1 parity gaps in the agent worksheet ops

NeverStopFMH:
StyleBI half of the L1 lane audit (`/parity-audit-workflow`). The plugin half is
inetsoft-technology/stylebi-wiz#1978 — **this one should land first or with it**, since three of the
tool parameters there are paired changes needing the `EditRequest` fields added here.

Based on `stylebi-wiz-main-rebase`, not `main`.

## What this closes

The agent worksheet path (`inetsoft/web/wiz/worksheet/**`) reimplements operations the Composer UI
also performs, and had drifted from the UI's own backend. Each guard below was found by diffing the
two backend paths, then reproduced live through the plugin's MCP tools before any code was written.

| Guard | Pre-fix behaviour, reproduced live |
|---|---|
| `hasDependent` before delete | `delete_table` succeeded on a mirrored table, leaving a mirror whose every query fails and nothing able to repair it |
| `checkCyclicalDependency` on concat add | HTTP 500 **after** the cycle was already committed |
| existing-assembly check on concat create | HTTP 500, and the table whose name was reused was destroyed — self-referencing, 0 columns |
| duplicate-source check on concat add | silently accepted; the duplicate persisted |
| `requireStorableName` | a name containing `]]>` was accepted and written into a CDATA section, corrupting storage irreparably |
| `requireDerivedColumnFits` | a second derived column with the same source and option was a silent no-op |
| negative `maxRows` | silently became `-1`, i.e. *unlimited* — the opposite of what was asked |
| `setAggregate` after a mode change | left stale, so `isAggregate()` disagreed with the actual aggregate info |
| visible-in-viewsheet on `set_primary_assembly` | primary yet invisible to viewsheets |
| `validateConditions` after unpivot edit | conditions left referencing columns the edit removed |

`setTableProperties` becomes an 8-arg method carrying `rowCount`; the base's 7-arg signature is kept
as a delegating overload so existing callers and their tests are untouched. `EditRequest` gains
`confirmed`, `rowCount` and `concatDistinct`; only the terminal delegation reaches the canonical
constructor, so the four compatibility constructors keep chaining as before.

## Tests

11 new tests in `WorksheetEditServiceMutatorsTest`. Every one asserts the model is **unchanged**
alongside the exception — a guard that throws after already mutating would pass a
`assertThrows`-only test while leaving exactly the corruption it was written to prevent, which is
how two of the defects above behaved.

Pre-existing failures in this suite are unrelated and predate this branch: 15 `NoSuchMethodError`
against methods absent from source (from `24c8ee7e8`), plus one arriving with revert `d53d44a50`.

## One known gap, deliberately not repaired here

The duplicate-source guard **reimplements** a check StyleBI already has at
`ConcatenateTablesService.java:431-455`, reached on the UI path and surfaced to users as
`common.table.unionDuplicate`. Swapping in a call to it is not a safe refactor:
`getTableCompatibility:475-497` discards an incompatible inner verdict when both sides are
`ConcatenatedTableAssembly` and the outer column count passes. The UI can never reach that — its
sidebar filters the duplicate out first — but a hand-constructed `concatenate/add-table` payload
can, and the agent path is hand-constructed payloads by definition. So calling the existing method
inherits a hole the UI does not have, while keeping the copy means two implementations that will
drift.

Which semantics is intended is not stated by either path, so this is recorded as row 36 in the lane
report and left for a decision rather than settled silently under cover of a bug fix. Behaviour as
shipped is correct — the UI forbids the duplicate too, so this is not the tool becoming stricter
than the UI.

### DIFF
--- a/core/src/main/java/inetsoft/uql/util/XEmbeddedTable.java
+++ b/core/src/main/java/inetsoft/uql/util/XEmbeddedTable.java
@@ -798,66 +798,80 @@ public class XEmbeddedTable
          throw new RuntimeException(tableNotReady);
       }
 
+      // Recorded so a rejected conversion (force=false, an unparseable value) can put this
+      // column's declared type back to what it was. types[col] is written below BEFORE the
+      // per-row conversion loop that can throw -- previously that write was never undone on the
+      // throw path, so a caller that got "conversion failed" back still found getDataType(col)/
+      // getColType(col) reporting the rejected type, even though not a single value had actually
+      // been converted (swapTable is only reached after the loop completes without throwing).
+      String oldType = types[col];
       types[col] = type;
-      XSwappableTable nDataTable = createTable();
-      copyTable(nDataTable, 0, getHeaderRowCount());
-      // get data from the original table if type changed or has been set back to its original type
-      boolean typeChanged = !Tool.equals(type, Tool.getDataType(xtable.getColType(col))) ||
-         (originalTable != null &&
-            Tool.equals(type, Tool.getDataType(originalTable.getColType(col))));
 
-      // cache the converted data
-      final int maxCacheSize = 10000;
-      Map<Object, Object> convertMap = new LinkedHashMap<Object, Object>() {
-         @Override
-         protected boolean removeEldestEntry(Map.Entry<Object, Object> eldest) {
-            return size() > maxCacheSize;
-         }
-      };
+      try {
+         XSwappableTable nDataTable = createTable();
+         copyTable(nDataTable, 0, getHeaderRowCount());
+         // get data from the original table if type changed or has been set back to its original type
+         boolean typeChanged = !Tool.equals(type, Tool.getDataType(xtable.getColType(col))) ||
+            (originalTable != null &&
+               Tool.equals(type, Tool.getDataType(originalTable.getColType(col))));
 
-      for(int i = getHeaderRowCount(); xtable.moreRows(i); i++) {
-         Object[] arr = new Object[types.length];
-
-         for(int j = 0; j < types.length; j++) {
-            // get the original data instead of the data which
-            // maybe resetted after setting data type(52298).
-            if(typeChanged && j == col && originalTable != null &&
-               i < originalTable.getRowCount() && j < originalTable.getColCount())
-            {
-               arr[j] = originalTable.getObject(i, j);
+         // cache the converted data
+         final int maxCacheSize = 10000;
+         Map<Object, Object> convertMap = new LinkedHashMap<Object, Object>() {
+            @Override
+            protected boolean removeEldestEntry(Map.Entry<Object, Object> eldest) {
+               return size() > maxCacheSize;
             }
-            else {
-               arr[j] = xtable.getObject(i, j);
-            }
-         }
+         };
 
-         Object oval = arr[col] ;
+         for(int i = getHeaderRowCount(); xtable.moreRows(i); i++) {
+            Object[] arr = new Object[types.length];
 
-         if(convertMap.containsKey(oval)) {
-            arr[col] = convertMap.get(oval);
-         }
-         else {
-            try {
-               arr[col] = Tool.transform(oval, type, format, true);
-               arr[col] = Tool.getData(type, arr[col]);
-            }
-            catch(Exception ex) {
-               if(force || XSchema.isNumericType(type) && Util.isNotApplicableValue(arr[col])) {
-                  arr[col] = null;
+            for(int j = 0; j < types.length; j++) {
+               // get the original data instead of the data which
+               // maybe resetted after setting data type(52298).
+               if(typeChanged && j == col && originalTable != null &&
+                  i < originalTable.getRowCount() && j < originalTable.getColCount())
+               {
+                  arr[j] = originalTable.getObject(i, j);
                }
                else {
-                  throw ex;
+                  arr[j] = xtable.getObject(i, j);
                }
             }
 
-            convertMap.put(oval, arr[col]);
+            Object oval = arr[col] ;
+
+            if(convertMap.containsKey(oval)) {
+               arr[col] = convertMap.get(oval);
+            }
+            else {
+               try {
+                  arr[col] = Tool.transform(oval, type, format, true);
+                  arr[col] = Tool.getData(type, arr[col]);
+               }
+               catch(Exception ex) {
+                  if(force || XSchema.isNumericType(type) && Util.isNotApplicableValue(arr[col])) {
+                     arr[col] = null;
+                  }
+                  else {
+                     throw ex;
+                  }
+               }
+
+               convertMap.put(oval, arr[col]);
+            }
+
+            nDataTable.addRow(arr);
          }
 
-         nDataTable.addRow(arr);
+         nDataTable.complete();
+         swapTable(nDataTable);
       }
-
-      nDataTable.complete();
-      swapTable(nDataTable);
+      catch(Exception ex) {
+         types[col] = oldType;
+         throw ex;
+      }
    }
 
    /**
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
@@ -356,7 +356,40 @@ public record EditRequest(
     * unchanged, matching {@code distinct}/{@code mergeable}. Distinct from {@code visible}, which
     * is a per-column flag for set_column_visibility.
     */
-   Boolean visibleInViewsheet
+   Boolean visibleInViewsheet,
+   /**
+    * For change_column_type: whether to force a conversion that some values cannot survive.
+    *
+    * <p>{@code XEmbeddedTable#setDataType}'s {@code force} flag decides what happens to a value
+    * the target type cannot parse — {@code true} writes {@code null} over it, {@code false}
+    * throws. The Composer asks: it calls with {@code false} first and, on failure, rolls the
+    * column back and puts the choice to the user
+    * ({@code ColumnTypeDialogService#handleChangeTypeParseException}). This op passed a hardcoded
+    * {@code true}, so unconvertible values were discarded with nothing reported.
+    *
+    * <p>Defaults to {@code true} when omitted, preserving that behaviour for callers built
+    * against it; pass {@code false} to be told instead of losing the values.
+    */
+   Boolean confirmed,
+   /**
+    * For set_table_properties on an embedded table: how many data rows it should hold.
+    *
+    * <p>The Composer's table-properties dialog exposes this as "Rows"
+    * ({@code TablePropertyDialogService:113-121}); this op had no way to reach it. Growing the
+    * count appends empty rows, shrinking it drops the ones past the new end. Ignored, rather than
+    * refused, for a table that is not embedded — the dialog omits the control there too.
+    */
+   Integer rowCount,
+   /**
+    * For add_concatenation: whether each adjacent pair's operator de-duplicates. The Composer's
+    * concatenation dialog exposes this per operator and carries it through
+    * {@code WorksheetEventUtil#convertOperator}; this op had no way to set it. Omitted leaves the
+    * engine's default rather than choosing for the caller.
+    *
+    * <p>Named apart from {@code distinct}, which is the table-level "return distinct values"
+    * property on set_table_properties -- a different setting on a different object.
+    */
+   Boolean concatDistinct
 ) {
    /**
     * Compatibility constructor for callers built before {@code visibleInViewsheet} was added —
@@ -391,7 +424,7 @@ public record EditRequest(
            groupOthers, variableValues, x, y, label, defaultValue, mode, insert, subtables,
            sourceTable, attribute, endpoint, parameters, lookup, lookupExpandArrays,
            lookupTopLevelOnly, suffix, customLookups, crosstab, labels, choices, joinPaths,
-           mergeable, null);
+           mergeable, null, null, null, null);
    }
 
    /**
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -1864,9 +1864,12 @@ public class WorksheetAgentController {
             editor.setColumnVisibility(req.table(), req.column(),
                                        req.visible() != null && req.visible());
          case "change_column_type" ->
-            editor.changeColumnType(req.table(), req.column(), req.type());
+            // Absent confirmed means force, matching what this op did before the flag existed.
+            editor.changeColumnType(req.table(), req.column(), req.type(),
+                                    req.confirmed() == null || req.confirmed());
          case "add_concatenation" ->
-            editor.addConcatenation(req.name(), req.tables(), req.concatType());
+            editor.addConcatenation(req.name(), req.tables(), req.concatType(),
+                                    req.concatDistinct());
          case "add_mirror" ->
             editor.addMirror(req.name(), req.source());
          case "set_conditions" ->
@@ -1922,7 +1925,7 @@ public class WorksheetAgentController {
             editor.setTableProperties(
                req.table(), req.newName() != null ? req.newName() : req.alias(),
                req.description(), req.maxRows(), req.distinct(), req.mergeable(),
-               req.visibleInViewsheet());
+               req.visibleInViewsheet(), req.rowCount());
          case "add_cross_join" ->
             editor.addCrossJoin(req.name(), req.leftTable(), req.rightTable());
          case "add_merge_join" ->
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -707,6 +707,12 @@ public class WorksheetAgentController {
       String attributeName = req.attribute();
       String name = req.name();
 
+      // Same unescaped-CDATA write path as WorksheetEditService.Editor.placeAssembly/
+      // renameTable/addJoin/duplicateAssembly/renameVariable/createVariable -- see
+      // requireStorableName's javadoc. Checked up front, before the permission checks and
+      // datasource/logical-model metadata lookups below, so a doomed name fails fast.
+      WorksheetEditService.Editor.requireStorableName(name, "A named group name");
+
       SourceInfo sinfo;
       DataRef ref;
 
@@ -1595,6 +1601,13 @@ public class WorksheetAgentController {
          String tableName = (name != null && !name.isBlank())
             ? name
             : AssetUtil.getNextName(ws, AbstractSheet.TABLE_ASSET);
+
+         // Unlike add_table's datasource-bound paths (addBoundTable/addLogicalModelTable/
+         // addTabularTable), this name comes from the CSV/Excel import's caller-supplied
+         // 'name' verbatim -- it is never run through AssetUtil.normalizeTable, so it needs
+         // its own check for the same unescaped-CDATA write path. See requireStorableName's
+         // javadoc.
+         WorksheetEditService.Editor.requireStorableName(tableName, "A table name");
 
          EmbeddedTableAssembly assembly = new EmbeddedTableAssembly(ws, tableName);
 
@@ -2649,6 +2662,12 @@ public class WorksheetAgentController {
             "Use edit_variable to change its type, label, or default value instead.");
       }
 
+      // Same unescaped-CDATA write path as WorksheetEditService.Editor.placeAssembly/
+      // renameTable/addJoin/duplicateAssembly/renameVariable -- see requireStorableName's
+      // javadoc. This path builds its own DefaultVariableAssembly and adds it directly
+      // rather than through Editor, so it needs its own call.
+      WorksheetEditService.Editor.requireStorableName(name, "A variable name");
+
       AssetVariable var = new AssetVariable(name);
 
       if(label != null) {
@@ -2909,6 +2928,11 @@ public class WorksheetAgentController {
          String tableName = body.name() != null && !body.name().isBlank()
             ? body.name()
             : AssetUtil.getNextName(ws, AbstractSheet.TABLE_ASSET);
+
+         // Same caller-supplied-name-used-verbatim gap as createEmbeddedTable: this dedicated
+         // /sql-query endpoint's 'name' never goes through AssetUtil.normalizeTable, so it needs
+         // its own check for the unescaped-CDATA write path. See requireStorableName's javadoc.
+         WorksheetEditService.Editor.requireStorableName(tableName, "A table name");
 
          SQLBoundTableAssembly assembly = new SQLBoundTableAssembly(ws, tableName);
 
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -3037,8 +3037,14 @@ public class WorksheetEditService {
        * rejects '/' and '"', but those are legal CDATA content and round-trip intact -- confirmed
        * live -- so enforcing the full rule here would break names that already exist and work
        * while protecting nothing.
+       *
+       * <p>Package-private (not {@code private}) so {@link WorksheetAgentController} can reuse
+       * the same check for the assembly-creation paths it builds directly against
+       * {@code Worksheet}/{@code RuntimeWorksheet} rather than through this {@code Editor} --
+       * {@code createVariable} and {@code addDatasourceScopedNamedGroup} -- instead of
+       * duplicating the rule.
        */
-      private static void requireStorableName(String name, String what) throws PairingException {
+      static void requireStorableName(String name, String what) throws PairingException {
          if(name != null && name.contains("]]>")) {
             throw new PairingException(
                what + " cannot contain \"]]>\". The name is written into a CDATA section " +
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -774,6 +774,11 @@ public class WorksheetEditService {
             throw new PairingException("Join requires a name.");
          }
 
+         // Unlike the two-table addJoin overload, this path cannot go through placeAssembly (see
+         // the comment below on ws.addAssembly(join)), so the name has to be checked explicitly
+         // here instead of inheriting placeAssembly's requireStorableName call.
+         requireStorableName(name, "An assembly name");
+
          if(joinPaths == null || joinPaths.isEmpty()) {
             throw new PairingException("Multi-table join requires at least one join path.");
          }
@@ -2363,6 +2368,19 @@ public class WorksheetEditService {
          TableAssembly newTable = requireTable(tableName);
          TableAssembly[] existing = ctbl.getTableAssemblies();
 
+         // Refuse the concatenation as a source of itself. This is a special case neither check
+         // below catches: ctbl is never a member of `existing` (it IS the concatenation, not one
+         // of its subtables), so the duplicate-source loop can't see it, and
+         // checkCyclicalDependency(ws, ctbl, newTable) can't either when newTable == ctbl --
+         // AssetUtil#getDependedAssemblies0 seeds its visited set with the root and passes
+         // included=false for it, so asking whether ctbl (as `otherAssembly`) already depends on
+         // itself (as `targetAssembly`) finds nothing, because before this attach nothing yet
+         // links ctbl back to itself.
+         if(tableName.equals(concatName)) {
+            throw new PairingException(
+               "\"" + concatName + "\" cannot be a source of itself.");
+         }
+
          // Refuse a source that is already present. ConcatenateTablesService#checkValidity refuses
          // it with common.table.unionDuplicate; this path never ported the check, so a repeated
          // source was accepted and counted its rows twice in the UNION with nothing reporting it.
@@ -2655,6 +2673,8 @@ public class WorksheetEditService {
             throw new PairingException("Assembly already exists: " + newName);
          }
 
+         requireStorableName(newName, "An assembly name");
+
          try {
             WSAssembly clone = (WSAssembly) wsa.clone();
             clone.getWSAssemblyInfo().setName(newName);
@@ -2793,6 +2813,8 @@ public class WorksheetEditService {
          if(!(a instanceof DefaultVariableAssembly)) {
             throw new PairingException("Variable assembly not found: " + oldName);
          }
+
+         requireStorableName(newName, "A variable name");
 
          if(!ws.renameAssembly(oldName, newName, true)) {
             throw new PairingException(
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -279,6 +279,20 @@ public class WorksheetEditService {
             try {
                RenderWaitSupport.awaitOrRetry(() -> {
                   WorksheetEventUtil.refreshColumnSelection(rws, name, true);
+                  // Drop conditions left pointing at columns the refresh just removed. An op that
+                  // changes a table's output columns without touching its conditions leaves an
+                  // add_filter or set_ranking referencing a DataRef no longer in the selection --
+                  // edit_unpivot is the clearest case, since changing headerColumns renames every
+                  // melted column. The Composer validates at exactly this point, right after its
+                  // own refreshColumnSelection
+                  // (TableUnpivotDialogService#changeUnpivotTableRowHeaders), and
+                  // WorksheetMutationSupport already does it for the aggregate path. Doing it in
+                  // the shared refresh covers every op rather than requiring each to remember.
+                  //
+                  // After refreshColumnSelection and before loadTableData, deliberately: a stale
+                  // selection would make this delete conditions that are still live, and the
+                  // retry wrapper only re-runs the whole block, so ordering inside it holds.
+                  AssetUtil.validateConditions(ta.getColumnSelection(), ta);
                   WorksheetEventUtil.loadTableData(rws, name, true, true);
                   return null;
                }, remaining, (int) Math.max(1, remaining / 1000));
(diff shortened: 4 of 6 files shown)


NeverStopFMH:
## Round 1 review — request changes

**Head commit reviewed:** `a65b903b375f810cc41d1d3f4ad45577227ab718`

### What was checked
Read the full diff against `stylebi-wiz-main-rebase` (`WorksheetEditService.java`, `EditRequest.java`, `WorksheetAgentController.java`, both test files) plus the surrounding, non-diffed methods in `WorksheetEditService.java`, `AssetUtil`, `WorksheetEventUtil`, `EmbeddedTableAssembly`, and `XEmbeddedTable` to check each new guard's invariant against sibling call sites, not just the reported repro. For each of the 10 guards I traced what it validates, where it validates relative to the mutation it guards, and at least one other entry point / branch that shares the same underlying rule. Counter-examples I tried to build and could not: the `add_concatenation` name-collision guard reached from a caller-supplied blank name (auto-named via `getNextName`, always free); `add_concat_subtable`'s column-count/cycle ordering when the new table mirrors an intermediate mirror chain (correctly rejected — `checkCyclicalDependency` walks the full depended-assembly graph, not just direct edges); negative/zero/omitted `maxRows` and `rowCount` boundaries (0 correctly means "no limit"/"no rows", negative correctly rejected, `null` correctly means "leave alone"); `distinct`/`confirmed`/`rowCount` all threading correctly from `EditRequest` through the compatibility constructor as documented (`null` preserves old behavior). I could build failing counter-examples for the items below, so this round is not clean.

### Findings

**1. [High] `add_concat_subtable` self-reference cycle slips both new guards**

`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java`, `addConcatSubtable`, lines 2354-2431.

Calling `add_concat_subtable(concatName, concatName)` (adding a concatenation as a subtable of itself, directly by name — no mirror indirection needed) is not caught by either new guard:
- The duplicate-source loop (`for(TableAssembly sub : existing) ...`) only iterates `ctbl.getTableAssemblies()`, which never contains `ctbl` itself, so it doesn't fire.
- `WorksheetEventUtil.checkCyclicalDependency(ws, ctbl, newTable)` with `newTable == ctbl` calls `AssetUtil.getDependedAssemblies(ws, ctbl, false)`. That method seeds its `visited` set with `ctbl` at the root and passes `included=false` for it — exactly the failure mode the PR's own comment (lines 2396-2399) attributes to `AbstractWSAssembly#checkDependency` ("seeds its visited set with the root and passes included=false for it, so a loop back to the root is discarded as already-visited and never reaches the array it matches against"). Since nothing else in the graph currently points back to `ctbl` (it doesn't depend on itself yet — that's what we're about to create), the returned depended-assembly list never contains `ctbl`'s name, so `checkCyclicalDependency` returns `false`.

Both checks pass, `ctbl.setTableAssemblies(...)` runs, and the concatenation ends up listing itself as one of its own sources — a 1-hop version of exactly the "cycle committed, 500 later" defect this PR's centerpiece guard exists to prevent, just reached through the direct name instead of a mirror.

Concrete failing scenario: `add_concatenation("C", ["A","B"], "UNION")`, then `add_concat_subtable("C", "C")`.

**2. [High] `change_column_type` with `confirmed=false` "rolls back" the ColumnRef but leaves the embedded table's internal type array corrupted**

`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java`, lines 1705-1756 (catch block at 1732-1751); root cause is `core/src/main/java/inetsoft/uql/util/XEmbeddedTable.java`, `setDataType`, line 801.

`XEmbeddedTable.setDataType` sets `types[col] = type` unconditionally at the top of the method (line 801), before it iterates rows and attempts conversion. When a value can't be converted and `force=false`, it throws (line 849) without ever calling `swapTable` — so row values are untouched, but `types[col]` has already been permanently flipped to the new type on the assembly's live `XEmbeddedTable` object (`EmbeddedTableAssembly.getEmbeddedData()` returns the live `xdata` field directly, not a copy — confirmed at `EmbeddedTableAssembly.java:73-76`).

`WorksheetEditService.changeColumnType`'s catch block only restores `cr2`/`cr`'s `ColumnRef.dataType` back to `oldType`; it never restores `data`'s internal `types[]`. So after this call throws "Nothing was changed", `embedded.getEmbeddedData().getDataType(index)` / `getColType(index)` for that column report the rejected new type while the `ColumnSelection`'s `ColumnRef` correctly reports the old one — a genuine model desync, reachable for the first time only because this PR makes `force=false` reachable (`confirmed=false` was previously unreachable since the hardcoded call always passed `force=true`, which never throws down that branch). `EmbeddedTableAssembly.refreshColumnType(XEmbeddedTable)` reads exactly this `types[]` array and overwrites `ColumnRef.dataType` from it — so any later call into that method against this same embedded data would resurrect the discarded type change the caller was told never happened.

This is exactly the "guard mutates before it validates, restore is incomplete" class of defect the PR body calls out as two of the original bugs it's fixing — this new guard has the same shape, undetected because no test exercises the `confirmed=false` failure path at all (see Finding 4).

Concrete failing scenario: embedded table with a string column containing a non-numeric value; call `change_column_type` to `integer` with `confirmed=false`; op throws as documented, but the embedded data's declared column type is now silently `integer` while the `ColumnSelection` says `string`.

**3. [Important] `requireStorableName` (the `]]>` CDATA-corruption guard) is not applied at every assembly-naming entry point**

`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java`.

The guard is wired into `placeAssembly` (line 2994, used for `add_table`, `add_unpivot`, `add_join` (2-table), `add_concatenation`, `add_mirror`, `add_cross_join`, `add_merge_join`) and into `renameTable` (line 1591). But the underlying invariant — "no assembly name may contain the CDATA terminator, because `AssemblyInfo` writes it unescaped" — applies to every path that sets an assembly's stored name, and three are left open:
- `addJoin` (the multi-table/N-ary join overload, lines 770-846) builds `join` and calls `ws.addAssembly(join)` directly at line 833, bypassing `placeAssembly` entirely.
- `duplicateAssembly` (lines 2645-2674) sets the clone's name via `clone.getWSAssemblyInfo().setName(newName)` and calls `ws.addAssembly(clone)` directly at line 2669 — no `requireStorableName(newName, ...)` call.
- `renameVariable` (lines 2790-2802) calls `ws.renameAssembly(oldName, newName, true)` directly with no `requireStorableName` call, even though variable assembly names round-trip through the same unescaped-CDATA `AssemblyInfo` write path as table names.

Concrete failing scenarios: `add_join` with 3+ join paths and a `name` containing `]]>`; `duplicate_assembly` with a `newName` containing `]]>`; `rename_variable` with a `newName` containing `]]>` — all three still corrupt storage the same way `rename_table`/`placeAssembly`-backed ops were just fixed to refuse.

**4. [Minor - test coverage] Several new guards/parameters ship with zero regression tests**

`core/src/test/java/inetsoft/web/wiz/worksheet/WorksheetEditServiceMutatorsTest.java`.

The PR's stated test discipline ("every guard asserts the model is unchanged alongside the exception") is not applied to:
- `change_column_type`'s new `force`/`confirmed=false` path — no test at all (and see Finding 2, where this gap hid a real bug).
- `set_table_properties`'s new `rowCount` parameter — no test for growing/shrinking an embedded table's row count, for it being silently ignored on a non-embedded table, or for its own negative-value rejection.
- `add_concatenation`'s new `distinct` parameter — no test that it's actually threaded onto the created operators.
- `requireDerivedColumnFits`'s organization-max-column-cap branch — only the name-collision branch is tested (`addDateRangeColumnRefusesASecondColumnWithTheSameGeneratedName`); the `cs.getAttributeCount() >= getOrganizationMaxColumn()` branch has no test.
- The `"default"` branch of `set_table_mode`'s aggregate-flag restoration (lines 2921ff, symmetric to the tested `"live"` branch at 2886ff) has no dedicated test.

### Informational (not a regression in this diff — "same bug, other site")

**5.** `addColumn` (line 436) and `addExpressionColumn`'s underlying `WorksheetMutationSupport.addExpressionColumn` never enforce `Util.getOrganizationMaxColumn()`, even though this PR just ported that exact cap check (from `ValueRangeService:194`) into `requireDerivedColumnFits` for the two range-column ops. Pre-existing, not introduced here, but the same missing-guard pattern this PR is otherwise sweeping.

**6.** `setTableProperties`'s new `rowCount` is silently ignored (no error, no log) when the target table is not an `EmbeddedTableAssembly` — documented and intentional (mirrors the Composer dialog omitting the control), but it's an unambiguous, actionable caller intent being dropped without signal, which is the shape of thing this repo's tooling convention asks to fail loud on rather than silently absorb. Advisory only.

Please address at minimum Findings 1-3 (and ideally add the coverage in Finding 4, since Finding 2 is a direct instance of the risk that missing coverage was supposed to catch) before the next round.

NeverStopFMH:
Round-1 review findings addressed in 8d5ae488f (new HEAD, was a65b903b3).

**Fixed (all 3 code findings accepted as-is, no pushback):**

1. **[High] `addConcatSubtable` self-reference cycle** — confirmed: the duplicate-source loop only walks `ctbl.getTableAssemblies()`, which never includes `ctbl` itself, and `checkCyclicalDependency(ws, ctbl, newTable)` can't see a self-loop when `newTable == ctbl` either (`AssetUtil#getDependedAssemblies0` seeds `visited` with the root and passes `included=false` for it, so before the attach nothing yet links `ctbl` back to itself). Added an explicit `tableName.equals(concatName)` check ahead of both existing guards, matching the style of the duplicate-source check right next to it. (The UI path, `ConcatenateTablesService#checkValidity`, has the identical structural gap, but it's reached only via drag-and-drop of two distinct assemblies, so it isn't a usable precedent here — this needed its own guard.)

2. **[High] `change_column_type` `confirmed=false` corrupts `XEmbeddedTable`** — confirmed at the exact lines cited: `XEmbeddedTable#setDataType` wrote `types[col] = type` unconditionally before the per-row conversion loop, and the `force=false` throw path exited without ever reverting it (the loop only reaches `swapTable` on success). Since `EmbeddedTableAssembly#getEmbeddedData()` returns the live `xdata` field, that mutation landed on the assembly immediately. Fixed at the root: `setDataType` now snapshots the prior type and wraps the conversion in try/catch, restoring `types[col]` on any exception before rethrowing. `WorksheetEditService`'s existing catch block (which restores the `ColumnRef`'s type) is unchanged and still correct.

3. **[Important] `requireStorableName` not wired into `addJoin`(multi-table)/`duplicateAssembly`/`renameVariable`** — confirmed all three write a caller-supplied name via `ws.addAssembly`/`ws.renameAssembly` directly, bypassing `placeAssembly`/`renameTable`'s guard. Added `requireStorableName(...)` to all three at the point each name is first known valid to check.

**Tests added** to `WorksheetEditServiceMutatorsTest.java` (13 new, all passing):
- `addConcatSubtableRefusesSelfReferenceWithoutMutating` (finding 1)
- `changeColumnTypeWithoutConfirmationLeavesEmbeddedTableUnchanged` (finding 2 — asserts both the `ColumnRef` type and `XEmbeddedTable#getDataType` are rolled back)
- `addJoinMultiTableRefusesANameThatWouldBreakTheStoredXml`, `duplicateAssemblyRefusesANameThatWouldBreakTheStoredXml`, `renameVariableRefusesANameThatWouldBreakTheStoredXml` (finding 3)
- `setTablePropertiesGrowsEmbeddedRowCount`, `setTablePropertiesShrinksEmbeddedRowCount`, `setTablePropertiesIgnoresRowCountOnANonEmbeddedTable`, `setTablePropertiesRefusesANegativeRowCount` (finding 4 — rowCount)
- `addConcatenationThreadsDistinctIntoTheOperator` (finding 4 — distinct)
- `addDateRangeColumnRefusesAtTheColumnCapWithoutMutating` (finding 4 — `requireDerivedColumnFits` column-cap branch)
- `setTableModeRestoresTheAggregateFlagForDefaultMode` (finding 4 — `"default"` branch)

**Pushback:** none — all three code findings and the test-coverage gap matched the source exactly (line numbers had drifted slightly from the review, as expected). Findings 5 and 6 were explicitly out of scope for this round and were not touched.

**Verification:** `./mvnw -pl core -o -Dtest=WorksheetEditServiceMutatorsTest test` → `Tests run: 192, Failures: 0, Errors: 0, Skipped: 0`. (Full `core` module also rebuilt clean offline before running the scoped test.)

NeverStopFMH:
## Round 2 review — head SHA `8d5ae488f`

Verified independently (read the diff, walked the invariants, ran the test suite, and reverted each of the two riskiest fixes locally to confirm the regression tests actually fail without them). Verdict: **changes requested** — one new Important-severity gap in the same defect family as round-1 finding 3, everything else from round 1 is fixed and not overfit.

### Round-1 finding status

**1. [High] Self-reference cycle in `addConcatSubtable` — FIXED, not overfit.**
`WorksheetEditService.java:2379` adds the literal `tableName.equals(concatName)` guard, thrown before `existing` is mutated (only a prior read, `getTableAssemblies()`, happens first). Test `addConcatSubtableRefusesSelfReferenceWithoutMutating` — reverting the guard locally makes it fail (`Tests run: 1, Failures: 1`, mismatched-column-count exception instead of the "cannot be a source of itself" message), confirming it's load-bearing.

I checked the deeper invariant the finding asked about: does the fix also cover an indirect cycle (C → D → C via a mirror, or via two `add_concat_subtable` calls)? It does, but not through the new literal check — through `WorksheetEventUtil.checkCyclicalDependency(ws, ctbl, newTable)`, which round 1 (`a65b903b3`) had already wired in a few lines above. That helper calls `AssetUtil.getDependedAssemblies(ws, otherAssembly, included=false)`, and `CompositeTableAssembly.getDependeds()` registers each concat subtable as a normal `AssemblyRef(INPUT_DATA, …)` dependency edge — the same edge type mirrors use — so the generic dependency-graph walk transparently catches a cycle closed through a mirror (test `addConcatSubtableRefusesACycleWithoutMutating`, `C → M → C`) or through a second concat (same code path, untested explicitly but same mechanism, verified by code reading). The `included=false` seeding bug in `getDependedAssemblies0` only blinds the check to the *exact* root==target case (depth-1 self-loop), which is precisely why the explicit literal check was still necessary — it isn't a partial/overfit patch, it's the one case the general check structurally cannot see, and the general check itself is unimpaired for depth ≥ 1.

**2. [High] `change_column_type confirmed=false` corrupts `XEmbeddedTable` — FIXED.**
`XEmbeddedTable.java`: `oldType` is captured before `types[col] = type`, the whole conversion (loop + `nDataTable.complete()` + `swapTable`) is now inside a `try`, and the `catch(Exception ex)` restores `types[col] = oldType` before rethrowing. Confirmed:
- (a) Restoration fires on every exception path out of the loop — the loop's own `catch` either nulls the value (force path) or rethrows, and any rethrow propagates to the outer `catch`, no path skips it.
- (b) Nothing else is mutated in place before the throw: converted rows go into a fresh local `nDataTable`, and the live table is only swapped in via `swapTable(nDataTable)` *after* `complete()` succeeds, so a mid-loop exception discards the new table entirely — the "no partial write" property already existed and this fix doesn't disturb it.
- (c) `EmbeddedTableAssembly.refreshColumnType()` reads `data.getColType(i)` → `Tool.getDataClass(types[col])`, so once `types[col]` is restored, it's consistent. The call site (`WorksheetEditService.changeColumnType`, lines ~1734–1756) independently restores the `ColumnRef`'s declared type on the same exception, so both halves of the bookkeeping (embedded-table internal type and column-selection declared type) roll back together.
- Reverted the `types[col] = oldType` line locally and reran `changeColumnTypeWithoutConfirmationLeavesEmbeddedTableUnchanged` — it fails exactly as expected (`expected: <string> but was: <integer>`), so the test is load-bearing, not tautological.

**3. [Important] `requireStorableName` wired into `addJoin`/`duplicateAssembly`/`renameVariable` — FIXED for the three named sites, but incomplete — see new finding below.**
All three calls happen before the corresponding mutation (`addJoin`'s multi-table overload: guard at line 780, `ws.addAssembly(join)` at line 838; `duplicateAssembly`: guard at 2676, `clone.getWSAssemblyInfo().setName()` at 2680; `renameVariable`: guard at 2817, `ws.renameAssembly()` at 2819). I grepped this file for every other `ws.addAssembly`/`ws.renameAssembly`/`.setName(` call and every method with a name-like parameter; all the assembly-creation paths in `WorksheetEditService.java` route through `placeAssembly()`, which already carried the guard from round 1. No missed sibling *inside this file*.

However, extending the grep to "related files" turned up a real 4th (and 5th) sibling in `WorksheetAgentController.java` — see below.

**4. [Test coverage] — VERIFIED.**
Ran `./mvnw -pl core -o -Dtest=WorksheetEditServiceMutatorsTest test`: **Tests run: 192, Failures: 0, Errors: 0** — reproduces the fixer's claim exactly. All the named coverage-gap tests exist as claimed (`setTablePropertiesGrowsEmbeddedRowCount`/`Shrinks…`/`IgnoresRowCountOnANonEmbeddedTable`/`RefusesANegativeRowCount`, `addConcatenationThreadsDistinctIntoTheOperator`, `addDateRangeColumnRefusesAtTheColumnCapWithoutMutating`, `setTableModeRestoresTheAggregateFlagForDefaultMode`). Spot-reverted the fixes for findings 1 and 2 (see above) and confirmed the corresponding regression tests fail without them, then restored.

### New finding

**[Important] The `requireStorableName` CDATA-corruption guard is still missing on two assembly-creation paths in `WorksheetAgentController.java` — same bug, other site.**

The invariant this PR establishes is: *no caller-supplied assembly name that will be written into `AssemblyInfo`'s unescaped `<name><![CDATA[...]]></name>` (`AssemblyInfo.java:254`) may contain the literal `]]>` terminator.* Round 1 and round 2 wired the guard into every name-writing call in `WorksheetEditService.java` (`placeAssembly`, `renameTable`, `addJoin`, `duplicateAssembly`, `renameVariable`). But `WorksheetEditService` is not the only class that calls `ws.addAssembly(...)` with a caller-supplied name reachable from the same agent surface — `WorksheetAgentController.java` does it too, in two places, neither of which was touched by this PR and neither of which has ever had the guard:

1. **`createVariable`, `WorksheetAgentController.java:2690`** — `DefaultVariableAssembly assembly = new DefaultVariableAssembly(ws, name); … ws.addAssembly(assembly);` (line 2694). `name` is `req.name()`/`body.name()`, validated only for null/blank (lines 2618–2620, 2812–2813) — never checked against `]]>`. Reachable two ways:
   - The `add_variable` op dispatched *before* the generic `editor`-based switch, at `WorksheetAgentController.java:370-373` (`addVariableFromEdit` → `createVariable`), which is exactly why it bypasses every guard that got added to `WorksheetEditService`'s mutators.
   - The dedicated REST endpoint `POST /api/wiz/v1/agent/worksheet/{sessionToken}/variable` (`addVariable`, line 2804).
   - Concrete repro: call `add_variable` with `name = "evil]]><oops/>"`. It's accepted and added to the worksheet with no error. On save, `AssemblyInfo.java:254` emits `<name><![CDATA[evil]]><oops/>]]></name>` — malformed XML, silently corrupting the persisted worksheet, exactly the failure mode finding 3 was written to close.

2. **`addDatasourceScopedNamedGroup`, `WorksheetAgentController.java:846`** — `DefaultNamedGroupAssembly assembly = new DefaultNamedGroupAssembly(ws, name); … ws.addAssembly(assembly);` (a few lines later). `name = req.name()` (line 708) is not validated at all here (not even for blank), and specifically never checked against `]]>`. Reachable via `add_named_group` when `req.datasource()` is set (dispatch guard at lines 285–317), which routes to this method instead of through `editor.addNamedGroup` (the one `WorksheetEditService` path that already has `placeAssembly`'s guard).

Both are the identical corruption mechanism as round-1 finding 3, in the identical file family, reachable through the identical agent-facing surface (`edit`'s `op` dispatch) — this is squarely "the same bug, other site" that the CLAUDE.md review procedure asks to sweep for, and it slipped past both the round-1 review and both fix rounds because the sweep was scoped to `WorksheetEditService.java` only.

**Suggested fix** (not applied — I'm read-only this round): make `requireStorableName` visible outside `WorksheetEditService` (e.g. hoist it to a small shared utility, or make it package-visible/static-importable) and call it from both sites in `WorksheetAgentController.java` before `ws.addAssembly(...)`, plus a regression test per site mirroring `addConcatSubtableRefusesSelfReferenceWithoutMutating`'s pattern (assert the `PairingException`, assert nothing was added to the worksheet).

### Build

`./mvnw -pl core -o -Dtest=WorksheetEditServiceMutatorsTest test` → **Tests run: 192, Failures: 0, Errors: 0, Skipped: 0** (reproduced twice, before and after the two revert/restore spot-checks).

### Verdict

**Changes requested.** Findings 1, 2, and 4 from round 1 are solidly fixed and independently verified (not overfit — the depth>1 cycle case, the full-exception-path rollback, and the pre-mutation ordering were all checked, not just the literal repro). Finding 3 is fixed for the three named sites but the underlying rule needs to be closed on the two `WorksheetAgentController.java` sites above before this is done.

(Posted as a PR comment rather than a formal review because the reviewing account also authored the PR, which GitHub disallows for `gh pr review --approve/-r`.)

NeverStopFMH:
## Round 2 fix: closed the remaining CDATA-corruption gaps, plus two more found in a follow-up sweep

New HEAD: `cf37dc137`

### The two sites the round-2 review named

- `createVariable` (`WorksheetAgentController.java`) — reachable via `add_variable` (both the edit-dispatch op and the dedicated `POST .../variable` endpoint) — built a `DefaultVariableAssembly` and added it directly, bypassing the `requireStorableName` guard.
- `addDatasourceScopedNamedGroup` (`WorksheetAgentController.java`) — reachable via `add_named_group` when `datasource` is set — same gap for `DefaultNamedGroupAssembly`.

Both now call `WorksheetEditService.Editor.requireStorableName(name, "…")` before the assembly is constructed/added. I made `requireStorableName` package-private (was `private` on `Editor`) so the controller — same package — can reuse it instead of duplicating the rule.

### Follow-up grep found two more unguarded sites

Per the review's ask to re-grep for any other bypass, I found two more `ws.addAssembly(...)` call sites in `WorksheetAgentController.java` that use a caller-supplied name **verbatim**, without going through `AssetUtil.normalizeTable` (unlike `addBoundTable`/`addLogicalModelTable`/`addTabularTable`, whose names are all sanitized first and so were never actually at risk):

- `createEmbeddedTable` — the shared tail of `importCsv`/`importExcel` — used the import's `name` param directly as the new embedded table's assembly name.
- `addSqlQuery` (the dedicated `POST .../sql-query` endpoint) — used `body.name()` directly as the new SQL-bound table's assembly name.

Both are now guarded the same way, checked as the first thing inside the `applyOnRuntime` callback (before any SQL parsing / JDBC work for the SQL-query path), so a doomed name fails fast.

I re-checked every other `ws.addAssembly`/`ws.renameAssembly` call in both `WorksheetAgentController.java` and `WorksheetEditService.java` — the remaining ones either go through `placeAssembly` (already guarded), reuse an *existing* assembly's already-valid name (`convertToEmbedded`'s replace=true path, `replaceEmbeddedTable`), or run the name through `AssetUtil.normalizeTable` first (`addBoundTable`, `addLogicalModelTable`, `addTabularTable`), which already strips the CDATA terminator. No further gaps found in these two files.

(Note: other `ws.addAssembly` call sites exist elsewhere in the codebase — e.g. `WorksheetTableService`, `AddFilterService` — but those belong to the Composer's own human-driven dialog paths, which already validate the name client-side, and to the separate wiz-services `/ws/table` write path. Neither is touched by this PR's diff, so I left them out of scope for this round; flagging for awareness only.)

### Tests

Added 4 regression tests in `WorksheetAgentControllerTest.java` (one per fixed site), each asserting a `bad]]>name` request is rejected with a `PairingException` mentioning `]]>` and that no assembly with that name is left in the worksheet.

Ran:
```
./mvnw -pl core -o -Dtest=WorksheetAgentControllerTest,WorksheetEditServiceMutatorsTest test
```
Result: `Tests run: 284, Failures: 0, Errors: 0` (`WorksheetAgentControllerTest`) + `Tests run: 192, Failures: 0, Errors: 0` (`WorksheetEditServiceMutatorsTest`). BUILD SUCCESS.

NeverStopFMH:
## Round 3 review — verdict: **approved** (no findings; ready for a human to formally approve/merge)

**Head SHA reviewed:** `cf37dc137` (confirmed via `gh pr view` and local checkout)

### Round-2 finding status: FIXED, no gaps found

Round-2 flagged `requireStorableName` (the CDATA-corruption guard) missing at two named sites in `WorksheetAgentController`. I independently re-verified all 5 of the fixer's claims rather than taking them on faith:

1. **Visibility change** — `WorksheetEditService.Editor.requireStorableName` is now package-private `static` (was `private static`). Confirmed pure visibility change, no behavior difference. The `inetsoft.web.wiz.worksheet` package has only 8 files total, so the widened access surface is minimal and doesn't let anything call it "wrongly."

2. **Four guard call sites** — read the actual diff and traced each one:
   - `addDatasourceScopedNamedGroup` (line 714): guard runs on `name` before it's passed unchanged into `new DefaultNamedGroupAssembly(ws, name)` (line 852) and `ws.addAssembly(assembly)` (line 859). Confirmed correct order.
   - `createEmbeddedTable` (line 1610): guard runs on `tableName` right after it's computed, before `new EmbeddedTableAssembly(ws, tableName)` (1612) and `ws.addAssembly` (1632). Confirmed both call sites of this helper (`importCsv`/`importExcel`) route through it.
   - `createVariable` (line 2669): guard runs on `name` before `new DefaultVariableAssembly(ws, name)` (2709) / `ws.addAssembly` (2713). Confirmed both call sites (`addVariableFromEdit` and the dedicated add-variable endpoint) route through this single helper.
   - `addSqlQuery` (line 2935): guard runs on `tableName` before `new SQLBoundTableAssembly(ws, tableName)` (2937) / `ws.addAssembly` (3003).

3. **The "safe" claims re-verified from source, not taken on the fixer's word:**
   - `AssetUtil.normalizeTable` (core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java:1576) keeps only letters/digits/space and collapses every other character run to a single `_`. Since `]`, `]`, `>` are none of those, any `]]>` is destroyed — confirmed it really does strip the terminator, not just "illegal filesystem chars" as I was asked to check for. `addBoundTable`/`addLogicalModelTable`/`addTabularTable` all call it before `ws.addAssembly`, and the subsequent `AssetUtil.getNextName` only appends an integer suffix or returns the already-normalized preferred name — it never reintroduces the raw caller string.
   - `convertToEmbedded` (line 2597) and `replaceEmbeddedTable` (line 1685) both do `ws.getAssembly(<caller-supplied name>)` first and require it to already exist as the right assembly type — they mutate/replace in place, they never write a fresh unvalidated name. Confirmed no gap.

4. **Full sweep of `ws.addAssembly`/`renameAssembly` in both files**, done independently (not just re-reading the fixer's grep): 8 `addAssembly` calls + 0 `renameAssembly` calls in `WorksheetAgentController`, all accounted for above or already round-1-guarded via `placeAssembly`. 6 `addAssembly`/`renameAssembly` calls in `WorksheetEditService`, all already carry `requireStorableName` (5 pre-existing + reused by the new callers). No unguarded site found.
   - Also grepped `WorksheetTableService`/`AddFilterService` (different package, `inetsoft.web.wiz.service`) — confirmed the fixer's disclosed-but-not-fixed sites are real and, being in a different package, genuinely can't reach the package-private guard without a larger refactor. Consistent with "out of scope for this PR," not a hidden gap.

5. **Tests**: spot-checked 2 of the 4 new tests (`editRejectsAddVariableWithANameThatWouldBreakTheStoredXml` and `addSqlQueryRefusesANameThatWouldBreakTheStoredXml`) by locally commenting out their corresponding guard call — both failed as expected (one with "nothing was thrown", one because execution reached real SQL parsing instead of being short-circuited), then restored and re-confirmed green. Ran `./mvnw -pl core -o -Dtest=WorksheetAgentControllerTest,WorksheetEditServiceMutatorsTest test` myself:
   - `WorksheetAgentControllerTest`: **92** tests, 0 failures
   - `WorksheetEditServiceMutatorsTest`: **192** tests, 0 failures
   - Combined total: **284**, 0 failures, BUILD SUCCESS

   Minor note (not a functional defect): the fixer's round-3 report states "`WorksheetAgentControllerTest` (284) + `WorksheetEditServiceMutatorsTest` (192)" — the 284 is actually the *combined* total across both classes; `WorksheetAgentControllerTest` alone is 92. Numbers are otherwise accurate and everything is green; flagging only so the count doesn't get miscited later.

### New findings this round: none

No overfit smells, no narrowed conditions, no repro-only tests — each of the 4 new tests asserts both the `PairingException` message and that no assembly was left behind, and I confirmed each one actually exercises its guard (not some other early-exit).

### This round's diff vs. earlier fixes

The `requireStorableName` visibility change and the 4 new call sites don't touch or reorder anything from round 1/2's fixes (self-reference cycle guard, `XEmbeddedTable` corruption fix, the first 3 `requireStorableName` sites) — verified those still pass unchanged inside the same 284-test run.

**This is the terminal round.** Zero new findings, round-2's finding is fixed with no gaps. The PR is ready for a human to formally `gh pr review --approve`/merge outside this account's own-PR restriction.

jshobe-inetsoft (approved):
## Code Review — PR #4901

Reviewed head commit `cf37dc137` (base: `stylebi-wiz-main-rebase`).

### Summary

This closes ten specific parity gaps between the agent worksheet ops (`inetsoft/web/wiz/worksheet/**`) and the Composer UI's own backend (`hasDependent` before delete, cycle/duplicate/existing-name checks on concatenation, CDATA-unsafe name validation, derived-column collision/cap checks, negative `maxRows`, stale `isAggregate()` after mode change, `visibleInViewsheet` on `set_primary_assembly`, and dropping stale conditions after a column-removing edit), plus a real bug fix in `XEmbeddedTable.setDataType` (declared column type wasn't rolled back when a forced-off type conversion threw). Each guard is backed by a plausible, well-documented reproduction, and the change threads three new `EditRequest` fields (`confirmed`, `rowCount`, `concatDistinct`) through to new service-layer overloads while keeping the old overloads as delegating compatibility shims.

- Traced every new guard against its call site: `deleteTable`'s `hasDependent` check, `addConcatSubtable`'s self-reference/duplicate/cycle/column-count ordering, `addConcatenation`'s name-collision and repeated-source checks, `requireDerivedColumnFits`'s cap-then-collision checks, and `requireStorableName`'s CDATA-terminator check are all placed **before** any mutation of the worksheet, so a caller that hits the guard sees no partial write. This is exactly the property the new tests assert (e.g. `addConcatSubtableRefusesACycleWithoutMutating` checks the subtable count didn't grow, not just that an exception was thrown).
- `EditRequest`'s canonical constructor grew from `visibleInViewsheet` to three more trailing fields; only the constructor immediately preceding the canonical one needed the extra `null, null, null` appended (verified all 4 remaining compatibility constructors still chain correctly with no changes needed).
- `WorksheetAgentController`'s dispatch wiring for `confirmed`/`rowCount`/`concatDistinct` was checked against the corresponding service overload's parameter order — all match up correctly, including the `req.confirmed() == null || req.confirmed()` "absent means force" logic.
- `XEmbeddedTable.setDataType`'s new try/catch correctly captures `oldType` before the mutation and only reassigns it on the exception path; `WorksheetEditService.changeColumnType` mirrors the same pattern for the `ColumnRef`'s declared type.
- The build (`Validation Build`) and Semgrep checks both pass on this commit, which independently confirms the code compiles and the new imports (`TableAssemblyInfo`, `AssetEventUtil`, etc.) resolve correctly.

### Minor issues (non-blocking)

1. **Missing dedicated test for the `validateConditions`-after-refresh guard.** Of the ten guards described in the PR body, nine have a corresponding new unit test in `WorksheetEditServiceMutatorsTest`. The tenth — dropping stale conditions after an edit that changes a table's output columns (the `edit_unpivot` case called out in the PR description) — has no test exercising it end-to-end (e.g. an unpivot rename followed by asserting a condition referencing a renamed/removed column is gone). It's plausibly covered indirectly since `refreshAssemblies` now runs `AssetUtil.validateConditions` for every table on every op (so existing tests do execute the code path without throwing), but the actual cleanup behavior for the motivating case isn't asserted anywhere.
2. **This cleanup lives inside a best-effort, budget-limited loop.** `refreshAssemblies`'s own doc comment states the per-table budget can be starved by a slow table earlier in iteration order, and failures are swallowed (`catch(Exception ex)` → `LOG.warn`, "best-effort warm-up, not a precondition"). The new `validateConditions` call was added into this same loop, so on a worksheet with many/slow tables the stale-condition cleanup for the very table that needs it is not actually guaranteed to run on the request that caused the corruption — it's now best-effort in the same sense the pre-existing cache-warming was, which is a slightly different reliability bar than the other nine guards (which reject up front, deterministically, before any commit).
3. **`changeColumnType`'s embedded-data catch block treats any `Exception` from `setDataType` as "values that cannot be converted".** In practice the only checked exception path with `force=false` is the parse failure inside `Tool.transform`/`Tool.getData`, so this is likely safe today, but it would silently mislabel an unrelated failure (e.g. an I/O error building the swap table) with a data-conversion message. Not a regression from this PR (the same broad catch pattern is now mirrored in `XEmbeddedTable` itself), just worth a narrower catch or a comment if it's ever revisited.

None of the above are blocking — they're either already low-risk (best-effort loop, same as pre-existing behavior) or straightforward test-coverage gaps rather than incorrect behavior.

### Security

No new security-relevant surface. The most security-adjacent change is the CDATA-terminator (`]]>`) validation on caller-supplied names (`requireStorableName`), which closes a real storage-corruption vector without over-restricting (deliberately leaves `/` and `"` alone since they round-trip fine in CDATA). No injection, auth, or permission-check changes.

### Backward compatibility

- All modified public method signatures (`setTableProperties`, `changeColumnType`, `addConcatenation`, the `EditRequest` canonical constructor) keep a delegating overload/compatibility constructor for existing callers.
- Default behavior when new parameters are omitted is preserved: `confirmed` omitted ⇒ force=true (old behavior), `rowCount` omitted ⇒ untouched, `concatDistinct` omitted ⇒ engine default (old behavior).
- The new guards do change previously-successful (but corrupting) requests into `PairingException`s — e.g. deleting a table with a dependent mirror, or adding a duplicate concatenation source, now fails where it previously silently succeeded and corrupted state. This is an intentional, documented behavior change to close bugs, not an accidental incompatibility, and matches the semantics already enforced by the Composer UI's own backend paths.

### Test impact

No existing tests are broken by this change (the `Validation Build` check passing corroborates this). New tests (11+ across the two test files) specifically assert the "no mutation on rejection" property, which is a good pattern for guard-style fixes.

### Checks / mergeability

- `claude-review` (Claude Code Review): **FAILURE** — per standing guidance, this check is currently broken for infrastructure reasons across many PRs in this repo and is excluded from this evaluation. This manual review substitutes for it.
- `Semgrep Scan`, `Semgrep OSS`, `build` (Validation Build): all **SUCCESS**.
- `mergeable`: MERGEABLE, no merge conflicts.

### Verdict: Approve

No blocking issues, no failing checks other than the excluded `claude-review`, and no merge conflicts. Approving without merging (per standing instructions, this should not be auto-merged).

PR #4902, opened 2026-08-31
Add +/- stepper buttons to the modern spinner

swaker854:
## Summary
- Adds custom decrement/increment buttons flanking the modern spinner's input, matching `number-stepper`'s visual design (rounded outer corners, shared border seam, hover/disabled states, press-and-hold repeat).
- Fixes composer interaction to match `vs-text-input`'s model: single click selects the object, double-click enters editing. The spinner previously swallowed every click on its own input, which blocked normal single-click selection like other assemblies get.
- Centers the value on both axes: removes the native number-input spin-button's reserved layout space (an invisible pseudo-element was still consuming 8px inside the input's internal shadow-DOM flex box even after `-webkit-appearance: none`, pushing the digit left).
- Draws one unified focus ring around the whole `-`/input/`+` group instead of just the input segment, matching the container's configured corner radius.
- Legacy (non-modern) spinners are untouched — every change is gated behind `model.vizModern`.

## Test plan
- [x] Verified in the composer: single click selects, double-click enters editing, `+`/`-` always clickable regardless of selection state (single click steps once, press-and-hold repeats)
- [x] Verified centering via exact pixel measurement (symmetric padding/border/button widths) and live screenshots
- [x] Verified the focus ring wraps the whole group and follows the configured `roundCorner`
- [x] Confirmed legacy spinners are visually and behaviorally unchanged

### DIFF
--- a/web/projects/portal/src/app/vsobjects/model/vs-spinner-model.ts
+++ b/web/projects/portal/src/app/vsobjects/model/vs-spinner-model.ts
@@ -18,4 +18,5 @@
 import { VSNumericRangeModel } from "./vs-numeric-range-model";
 
 export interface VSSpinnerModel extends VSNumericRangeModel {
+   editing?: boolean;
 }
--- a/web/projects/portal/src/app/vsobjects/objects/spinner/vs-spinner.component.html
+++ b/web/projects/portal/src/app/vsobjects/objects/spinner/vs-spinner.component.html
@@ -17,6 +17,7 @@
 -->
 @if (model) {
   @let textVerticalPosition = getTextVerticalPosition();
+  @let showAsLabel = showLabel && !model.vizModern;
   <vs-input-label-wrapper
     [labelModel]="model.labelModel"
     [labelSelected]="model.labelSelected"
@@ -34,9 +35,29 @@
     VSPopComponent [popComponentName]="getAssemblyName()" [popContainerName]="model.container"
     [popZIndex]="model.objectFormat.zIndex"
     (selectLabel)="selectLabel($event)">
-    <div class="vs-spinner" (click)="clearLabelSelection()">
-      @if (!showLabel && model) {
+    <div class="vs-spinner" [class.has-steppers]="model.vizModern"
+      [style.border-radius.px]="model.vizModern ? model.objectFormat.roundCorner : null"
+      (click)="clearLabelSelection()" (dblclick)="enableEditing()">
+      @if (!showAsLabel && model) {
+        @if (model.vizModern) {
+          <button type="button" class="vs-spinner-btn vs-spinner-btn-decrement"
+            [class.disable-actions-fade]="!model.enabled"
+            [disabled]="isDecrementDisabled"
+            [attr.aria-disabled]="isDecrementDisabled"
+            [attr.aria-label]="model.absoluteName + ' _#(Decrease)'"
+            [style.border-top-left-radius.px]="model.objectFormat.roundCorner"
+            [style.border-bottom-left-radius.px]="model.objectFormat.roundCorner"
+            (click)="onDecrementClick()"
+            (mousedown)="$event.preventDefault(); startHold(-1)"
+            (mouseup)="cancelHold()"
+            (mouseleave)="cancelHold()">
+            <svg viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="2" aria-hidden="true">
+              <path d="M3 8h10" stroke-linecap="round"/>
+            </svg>
+          </button>
+        }
         <input
+          #spinnerInput
           (click)="onClick($event)"
           (blur)="onBlur($event)"
           (keyup.enter)="onBlur($event)"
@@ -45,9 +66,10 @@
           [(ngModel)]="model.value"
           [class.form-control]="model.vizModern"
           [class.disable-actions-fade]="!model.enabled"
+          [class.disable-input]="isInputDisabled()"
           [class.move-handle]="(selected || inSelectedGroupContainer) && !viewer"
-          [style.text-align]="model.objectFormat.hAlign"
-          [style.padding]="textVerticalPosition"
+          [style.text-align]="model.vizModern ? 'center' : model.objectFormat.hAlign"
+          [style.padding]="model.vizModern ? null : textVerticalPosition"
           [safeFont]="model.objectFormat.font"
           [style.background]="model.objectFormat.background"
           [style.color]="model.objectFormat.foreground"
@@ -59,11 +81,27 @@
           [style.border-top]="model.objectFormat.border.top"
           [style.border-left]="model.objectFormat.border.left"
           [style.border-right]="model.objectFormat.border.right"
-          [style.border-radius.px]="model.objectFormat.roundCorner"
-          (mouseleave)="showLabel = true"
-          (mousedown)="$event.stopPropagation()">
+          [style.border-radius.px]="model.vizModern ? 0 : model.objectFormat.roundCorner"
+          (mouseleave)="showLabel = true">
+        @if (model.vizModern) {
+          <button type="button" class="vs-spinner-btn vs-spinner-btn-increment"
+            [class.disable-actions-fade]="!model.enabled"
+            [disabled]="isIncrementDisabled"
+            [attr.aria-disabled]="isIncrementDisabled"
+            [attr.aria-label]="model.absoluteName + ' _#(Increase)'"
+            [style.border-top-right-radius.px]="model.objectFormat.roundCorner"
+            [style.border-bottom-right-radius.px]="model.objectFormat.roundCorner"
+            (click)="onIncrementClick()"
+            (mousedown)="$event.preventDefault(); startHold(1)"
+            (mouseup)="cancelHold()"
+            (mouseleave)="cancelHold()">
+            <svg viewBox="0 0 16 16" fill="none" stroke="currentColor" stroke-width="2" aria-hidden="true">
+              <path d="M8 3v10M3 8h10" stroke-linecap="round"/>
+            </svg>
+          </button>
+        }
       }
-      @if (showLabel && model) {
+      @if (showAsLabel && model) {
         <div class="text-label"
           [style.background-color]="model.objectFormat.background"
           [style.justify-content]="model.objectFormat.justifyContent"
--- a/web/projects/portal/src/app/vsobjects/objects/spinner/vs-spinner.component.html
+++ b/web/projects/portal/src/app/vsobjects/objects/spinner/vs-spinner.component.html
@@ -82,7 +82,8 @@
           [style.border-left]="model.objectFormat.border.left"
           [style.border-right]="model.objectFormat.border.right"
           [style.border-radius.px]="model.vizModern ? 0 : model.objectFormat.roundCorner"
-          (mouseleave)="showLabel = true">
+          (mouseleave)="showLabel = true"
+          (mousedown)="onMouseDown($event)">
         @if (model.vizModern) {
           <button type="button" class="vs-spinner-btn vs-spinner-btn-increment"
             [class.disable-actions-fade]="!model.enabled"
--- a/web/projects/portal/src/app/vsobjects/objects/spinner/vs-spinner.component.scss
+++ b/web/projects/portal/src/app/vsobjects/objects/spinner/vs-spinner.component.scss
@@ -19,6 +19,74 @@
 .vs-spinner {
    width: 100%;
    height: 100%;
+
+   &.has-steppers {
+      display: flex;
+      align-items: stretch;
+   }
+}
+
+.vs-spinner.has-steppers input.form-control {
+   flex: 1 1 auto;
+   min-width: 0;
+}
+
+// A focused input's own :focus ring (_bootstrap-override.scss's .form-control:focus) only
+// wraps the input segment — draw one unified ring around the whole -/input/+ group instead,
+// and suppress the input's own so they don't double up.
+.vs-spinner.has-steppers:focus-within {
+   box-shadow: var(--inet-focus-ring);
+   outline: var(--inet-focus-ring-outline);
+}
+
+.vs-spinner.has-steppers:focus-within input.form-control {
+   box-shadow: none;
+}
+
+.vs-spinner.has-steppers:focus-within input.form-control,
+.vs-spinner.has-steppers:focus-within .vs-spinner-btn {
+   border-color: var(--inet-primary-color);
+}
+
+.vs-spinner-btn {
+   appearance: none;
+   flex-shrink: 0;
+   width: 18px;
+   padding: 0;
+   box-sizing: border-box;
+   border: 1px solid var(--inet-input-border-color);
+   background: var(--inet-input-bg-color);
+   color: var(--inet-text-muted-color);
+   cursor: pointer;
+   display: flex;
+   align-items: center;
+   justify-content: center;
+   transition: background-color 100ms ease, color 100ms ease;
+
+   svg {
+      width: 10px;
+      height: 10px;
+      flex-shrink: 0;
+   }
+
+   &:not(:disabled):hover {
+      background: var(--inet-ui-neutral-hover-bg-color);
+      color: var(--inet-text-color);
+   }
+
+   &:disabled {
+      color: var(--inet-text-subtle-color);
+      cursor: not-allowed;
+      opacity: 0.6;
+   }
+}
+
+.vs-spinner-btn-decrement {
+   border-right-width: 0;
+}
+
+.vs-spinner-btn-increment {
+   border-left-width: 0;
 }
 
 input {
@@ -42,6 +110,27 @@ input::-webkit-inner-spin-button {
 .vs-spinner input.form-control {
   min-height: 0;
   height: 100%;
+
+  // custom -/+ buttons replace the native spin arrows for modern spinners; suppress them so
+  // no browser/OS combination can render its own arrows alongside ours. display:none (not just
+  // -webkit-appearance:none) is required: the unscoped min-width/min-height rule above still
+  // reserves layout space for the (invisible) pseudo-element inside the input's internal
+  // shadow-DOM flex box otherwise, pushing the centered text left.
+  appearance: textfield;
+  -moz-appearance: textfield;
+
+  &::-webkit-inner-spin-button,
+  &::-webkit-outer-spin-button {
+    display: none;
+    -webkit-appearance: none;
+    margin: 0;
+    min-width: 0;
+    min-height: 0;
+  }
+}
+
+.disable-input {
+   pointer-events: none;
 }
 
 .text-label {
--- a/web/projects/portal/src/app/vsobjects/objects/spinner/vs-spinner.component.tl.spec.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/spinner/vs-spinner.component.tl.spec.ts
@@ -817,6 +817,16 @@ describe("VSSpinner — isIncrementDisabled / isDecrementDisabled", () => {
       expect(comp.isDecrementDisabled).toBe(true);
    });
 
+   it("should NOT disable increment when max is unset, regardless of value", () => {
+      const { comp } = createComponent({ model: makeModel({ value: 50, max: null } as any) });
+      expect(comp.isIncrementDisabled).toBe(false);
+   });
+
+   it("should NOT disable decrement when min is unset, regardless of value", () => {
+      const { comp } = createComponent({ model: makeModel({ value: 50, min: null } as any) });
+      expect(comp.isDecrementDisabled).toBe(false);
+   });
+
    it("should NOT disable decrement when enabled and above min", () => {
       const { comp } = createComponent({ model: makeModel({ value: 50, min: 0 }) });
       expect(comp.isDecrementDisabled).toBe(false);
--- a/web/projects/portal/src/app/vsobjects/objects/spinner/vs-spinner.component.tl.spec.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/spinner/vs-spinner.component.tl.spec.ts
@@ -757,6 +757,34 @@ describe("VSSpinner — onIncrementClick / onDecrementClick", () => {
       comp.onDecrementClick();
       expect(comp.model.value).toBe(0);
    });
+
+   it("should focus the input after stepping when the input is not gated (viewer)", () => {
+      const { comp } = createComponent({
+         viewer: true,
+         model: makeModel({ vizModern: true, value: 0, increment: 20, max: 100 } as any)
+      });
+      const focusSpy = vi.fn();
+      (comp as any).spinnerInputRef = { nativeElement: { focus: focusSpy } };
+
+      comp.onIncrementClick();
+
+      expect(focusSpy).toHaveBeenCalled();
+   });
+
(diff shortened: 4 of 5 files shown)


swaker854 (commented):
## Code Review — PR #4902: Add +/- stepper buttons to the modern spinner

### CI status (investigated, not blocking)
Both failing checks were pre-existing flakiness unrelated to this PR's diff (which touches only `vs-spinner-model.ts`, `vs-spinner.component.{html,scss,ts}`):
- **`build` (1st run)**: `DashboardRegistryOrgLifecycleTest` (org/dashboard-registry rename lifecycle) — unrelated to spinners; passed on rerun.
- **`build` (2nd run)**: `gui-tool.spec.ts > GuiTool density mode > defaults to dense when no density class is present` — the same known-flaky test called out in this session's own prior handoff notes (density CSS class leaking between test files); `gui-tool.spec.ts` is not in this PR's diff.
- **`claude-review`**: excluded per review instructions (this manual review takes its place).

Rerunning is recommended for a clean signal, but neither failure originates in this change.

### Blocking issues

**1. PR description overstates scope — the composer interaction-model change is not gated behind `model.vizModern`, so it also changes legacy (non-modern) spinners.**

The PR body states "Legacy (non-modern) spinners are untouched — every change is gated behind `model.vizModern`." That's true for the buttons, centering, and focus ring, but not for the selection/editing change:

- `isInputDisabled()` — `return !this.viewer && (!this.selected || !this.model?.editing);` — has no `vizModern` check.
- `(dblclick)="enableEditing()"` on the outer `.vs-spinner` div, and `[class.disable-input]="isInputDisabled()"` on the shared `<input>`, are both applied unconditionally — not wrapped in an `@if (model.vizModern)` block.

Since `showAsLabel` still hover-toggles for legacy spinners exactly as before, whenever a legacy spinner's `<input>` becomes visible (on hover) it will now also be `pointer-events: none` until the object is both selected **and** double-clicked into editing — a real behavior change for every existing (legacy) spinner already deployed, in the composer only (viewer is unaffected since `isInputDisabled()` short-circuits to `false` when `this.viewer` is true).

This may well be an intentional, in-scope decision (matching `vs-text-input`'s model universally rather than only for modern spinners) — but as written, the PR description is inaccurate about the blast radius, and a reviewer approving on the strength of "legacy is untouched" would be missing a real composer-workflow change for existing dashboards. Please either:
- (a) gate `isInputDisabled()`/`enableEditing()`/the `dblclick` binding behind `model.vizModern` so legacy truly is untouched, matching the description, or
- (b) update the PR description to state plainly that the select-then-edit gate is intentionally universal, and call out the legacy-composer behavior change as an explicit, in-scope part of this PR.

**2. Existing unit test asserts behavior this PR deliberately removes.**

`vs-spinner.component.tl.spec.ts`, `describe("VSSpinner — onClick")`, first test:
```ts
const stopSpy = vi.spyOn(event, "stopPropagation");
comp.onClick(event);
expect(stopSpy).toHaveBeenCalled();
```
`onClick()` no longer calls `event.stopPropagation()` (removed as part of the selection fix), so this assertion will fail. This test suite (`*.tl.spec.ts`) isn't wired into CI yet per the project's own `npm run test:portal:tl` comment, so it won't currently block the pipeline — but it's a real, confirmed failure waiting for whoever runs it locally or wires the suite in later. Please update the test to match the new intentional behavior (drop the `stopPropagation` assertion, or assert it's *not* called).

### Non-blocking observations

- **No new test coverage** for any of the substantial new behavior (buttons, hold-to-repeat, `isInputDisabled`/`enableEditing`, the `editing` model field). The existing spec file's own header documents a deliberate "risk-first coverage" philosophy for this exact component — worth extending here, especially for the hold-repeat timer logic and the increment/decrement disabled-at-bounds branches, which have no coverage at all.
- **`stepValue`/`onClick`/`onBlur` now share nearly identical trailing bodies** (validate → emit → mark unapplied → dispatch-or-defer). Small DRY opportunity to factor into one private `commitChange()` helper; `onClick`/`onBlur` were already duplicated before this PR, so not a regression, just an opportunity while touching this file.
- **`onIncrementClick`/`onDecrementClick` call `.focus()` on the input even when `isInputDisabled()` is true** (object not yet selected+editing). `pointer-events: none` doesn't prevent programmatic `.focus()`, so a single click on `+`/`-` on an unselected spinner both selects the object (click bubbles, no `stopPropagation` on the buttons) and quietly grants the input real keyboard focus — meaning a user could then type into it without ever double-clicking through `enableEditing()`. This is plausibly a reasonable consequence of "buttons are always interactive," but worth a deliberate look rather than a side effect of the focus-after-step call.

### What's solid
- The centering root-cause fix (native spin-button pseudo-element reserving 8px inside the input's shadow-DOM flex layout even after `-webkit-appearance: none`) is a genuinely subtle find, and the `display: none` + zeroed `min-width`/`min-height` fix is correct and well-commented.
- The unified `:focus-within` ring correctly suppresses the input's own ring to avoid doubling up, and follows the configured `roundCorner`.
- Press-and-hold repeat (`startHold`/`cancelHold`) mirrors `number-stepper`'s own click+mousedown pattern faithfully, including the initial-delay/acceleration timing.
- Gating (buttons, centering, focus ring) is correctly scoped to `model.vizModern` everywhere except the selection/editing behavior flagged above.

---

**Review verdict: would be Request Changes** (self-authored PR — GitHub does not allow a Request Changes/Approve review on your own PR, so this is posted as a comment).

jshobe-inetsoft (changes requested):
## Code Review — PR #4902: Add +/- stepper buttons to the modern spinner

### Summary of changes
Adds custom `+`/`-` stepper buttons to the modern (`model.vizModern`) spinner, fixes composer click/selection semantics to match `vs-text-input` (single click selects, double click edits), fixes native spin-button pseudo-element layout space causing off-center text, and unifies the focus ring around the whole control group. Touches `vs-spinner-model.ts`, `vs-spinner.component.{html,scss,ts}` only.

### Verdict: **Request Changes**

This does not meet all three approval gates:
- There is at least one blocking code issue (below).
- The `build` (Validation Build) check is currently failing (excluding `claude-review`, which is disregarded per current instructions since that check is known-broken infrastructure-wide).
- No merge conflicts — the branch is `MERGEABLE`.

---

### Blocking issues

**1. The PR description overstates scope: the composer selection/editing behavior change is not gated behind `model.vizModern`, so it also changes legacy (non-modern) spinners.**

The description states "Legacy (non-modern) spinners are untouched — every change is gated behind `model.vizModern`." That is true for the buttons, centering, and focus-ring CSS, but **not** for the new selection/editing gate:

- `isInputDisabled()` — `return !this.viewer && (!this.selected || !this.model?.editing);` — has no `vizModern` check.
- `(dblclick)="enableEditing()"` on the outer `.vs-spinner` div and `[class.disable-input]="isInputDisabled()"` on the shared `<input>` are applied unconditionally, not wrapped in an `@if (model.vizModern)` block.
- Previously **all** spinners (legacy included) had `event.stopPropagation()` in `onClick()` and `(mousedown)="$event.stopPropagation()"` on the input; both are removed unconditionally in this diff.

Net effect: for any existing legacy spinner in the composer, the input is now `pointer-events: none` until the assembly is both selected *and* double-clicked into editing — previously a single click let you type directly. Viewer is unaffected (`isInputDisabled()` short-circuits to `false` when `this.viewer` is true), but every existing legacy dashboard's composer editing workflow changes with no visual cue that anything is different.

This may be an intentional decision (universally matching `vs-text-input`'s model rather than only for modern spinners), but as written the description is inaccurate about the blast radius, and approving on the strength of "legacy is untouched" would miss a real behavior change to production dashboards. Please either:
- (a) gate `isInputDisabled()` / `enableEditing()` / the `dblclick` binding (and restore `stopPropagation` for the legacy path) behind `model.vizModern` so legacy is truly untouched, matching the description, or
- (b) update the PR description to state plainly that the select-then-edit gate is intentionally universal, and call out the legacy-composer behavior change as an explicit, in-scope part of this PR (and re-verify the test-plan checkbox that currently reads "Confirmed legacy spinners are visually **and behaviorally** unchanged" — behaviorally, they are not).

**2. An existing unit test now asserts behavior this PR deliberately removes, and will fail.**

`vs-spinner.component.tl.spec.ts`, `describe("VSSpinner — onClick")`, first test (`"should stop propagation, validate, emit spinnerClicked, and mark an unapplied change"`):

```ts
const stopSpy = vi.spyOn(event, "stopPropagation");
comp.onClick(event);
expect(stopSpy).toHaveBeenCalled();
```

`onClick()` no longer calls `event.stopPropagation()` (intentionally removed as part of the selection fix), so this assertion will fail. This suite (`*.tl.spec.ts`) isn't wired into `npm run verify`/CI yet (per the project's own `test:portal:tl` comment in `package.json`), so it didn't surface as a CI failure here — but it is a confirmed, real regression in the test suite that will break for anyone running it locally or whenever the suite gets wired in. Please update the assertion to match the new intentional behavior (drop it, or assert `stopPropagation` is *not* called).

**3. Failing `build` check.**

The Validation Build check is red. The one concrete failure visible in the log is `gui-tool.spec.ts > GuiTool density mode > defaults to dense when no density class is present` (`expected 'compact' to be 'dense'`) — a file untouched by this PR's diff, and consistent with test-order/class-leakage flakiness (a `viz-density-*` class left on `document.body` by a sibling test). This is very likely unrelated to this PR's changes and probably a rerun-and-it-passes situation, but per review policy only the `claude-review` check is disregarded — this failing `build` check still blocks approval as written. Please rerun the build (or confirm a clean rerun) before merge.

---

### Non-blocking observations

- **No new test coverage for the substantial new behavior**: the stepper buttons, hold-to-repeat timers, `isInputDisabled()`/`enableEditing()` gating, and the new `editing` model field are all untested. The existing spec file's own header documents a deliberate "risk-first coverage" philosophy for this component — worth extending here, especially for the hold-repeat timer logic and the increment/decrement disabled-at-bounds branches.
- **`stepValue`/`onClick`/`onBlur` share nearly identical trailing bodies** (validate → emit → mark unapplied → dispatch-or-defer). A small DRY opportunity to factor into one private `commitChange()` helper. `onClick`/`onBlur` were already duplicated before this PR, so this isn't a regression, just a missed opportunity while already touching the file.
- **`onIncrementClick`/`onDecrementClick` call `.focus()` on the input even when `isInputDisabled()` is true** (object not yet selected+editing). `pointer-events: none` doesn't block programmatic `.focus()`, so a single click on `+`/`-` on an unselected spinner both selects the object (the buttons don't `stopPropagation` either) and silently grants the input real keyboard focus — a user could then type into it without ever going through `enableEditing()`. Plausibly a reasonable consequence of "buttons are always interactive" (this is explicitly called out and tested per the PR's own test plan), but worth a deliberate confirmation rather than an incidental side effect of the post-step focus call.
- Timer fields (`holdInitTimer`/`holdRepeatTimer`/`holdAccelTimer`) are typed `any`; a `ReturnType<typeof setTimeout>` would be more precise, though this matches the existing `number-stepper` component's own style, so it's consistent with the codebase rather than a new issue.
- `isIncrementDisabled`/`isDecrementDisabled` compare directly against `model.max`/`model.min` without an explicit `!= null` guard (relies on `value >= undefined` evaluating to `false`). Functionally correct, but `number-stepper`'s equivalent getters are more explicit (`this.max != null && ...`), which is more readable/intention-revealing.

### What's solid

- The centering root-cause fix (the native spin-button pseudo-element still reserving 8px inside the input's shadow-DOM flex layout even after `-webkit-appearance: none`) is a genuinely subtle find, and the `display: none` + zeroed `min-width`/`min-height` fix, with its explanatory comment, is correct and well-documented.
- The unified `:focus-within` ring correctly suppresses the input's own ring to avoid doubling up, and follows the configured `roundCorner`.
- Press-and-hold repeat (`startHold`/`cancelHold`) faithfully mirrors `number-stepper`'s own click/mousedown pattern, including the initial-delay/acceleration timing (300ms delay, 60ms→20ms after 1.5s) and `cancelHold()` cleanup on `ngOnDestroy`, so no timer leaks.
- No security concerns: this is UI-only, all new bindings go through Angular's normal attribute/property binding (auto-escaped), no `innerHTML` or unsanitized user input, and the new `editing` model field is purely client-side transient UI state, never round-tripped to the server.
- Gating of the buttons/centering/focus-ring visuals is correctly scoped to `model.vizModern` everywhere except the selection/editing behavior flagged in Blocking Issue 1.

---

swaker854:
Addressed the review feedback with the following follow-up commits:

1. **`5bcb4bd8b`** — Fully gated the select-then-edit change behind `model.vizModern`: `enableEditing()` now no-ops for legacy spinners, and `mousedown` on the input restores `stopPropagation()` for the legacy path (the earlier `075c92953` had only fixed `isInputDisabled()`/`onClick()`). Legacy spinners are now genuinely unchanged, matching the PR description.
2. **`23f62ce89`** — Added test coverage for the previously-untested new behavior: hold-to-repeat timer sequencing (initial delay, repeat interval, 1500ms acceleration, disabled-bound early stop), increment/decrement click handlers, the disabled-bounds getters, `enableEditing()`'s gating, `isInputDisabled()`'s full matrix, and `onMouseDown()`.
3. **`8ab59f0bf`** — Refactored `stepValue`/`onClick`/`onBlur`'s duplicated validate→emit→dispatch-or-defer tail into a shared `commitChange()` helper.
4. **`916b49525`** — Fixed a focus-gating bug: clicking `+`/`-` on an unselected spinner was calling `.focus()` on the input even while `isInputDisabled()` was true. Since `pointer-events: none` doesn't block a programmatic `.focus()`, this silently granted real keyboard access around the select-then-edit gate. Focus is now only restored when the input isn't gated.
5. **`2152cb1c1`** — Made the `isIncrementDisabled`/`isDecrementDisabled` bounds checks explicit (`max/min != null` guards), matching `number-stepper`'s equivalent getters, instead of relying on `value >= undefined` evaluating to `false`.

Also noted but left as-is (both intentionally not code changes):
- The `stopPropagation` unit test regression flagged in review was already resolved by the earlier `075c92953` fix — the default mock model is legacy (`vizModern: false`), so `onClick()`'s conditional `stopPropagation()` still fires and the existing assertion passes.
- The `holdInitTimer`/`holdRepeatTimer`/`holdAccelTimer` fields being typed `any` matches `number-stepper`'s identical fields — tightening the type here alone would make this component inconsistent with the sibling it's deliberately modeled after.

The previously-failing Validation Build check was traced across its last 3 runs to causes unrelated to this diff (a `gui-tool.spec.ts` flake, an unrelated Java NPE in `DashboardRegistryOrgLifecycleTest`, and a 45-minute runner timeout) — none touching files this PR changes.

jshobe-inetsoft (approved):
## Code Review — Add +/- stepper buttons to the modern spinner

**Verdict: Approve** (non-blocking comments below)

Scope reviewed: `vs-spinner-model.ts`, `vs-spinner.component.{html,scss,ts}`, `vs-spinner.component.tl.spec.ts` (+590 / -20, base `0827-inputComponent-updates`).

### Merge gate

| Item | Status |
| --- | --- |
| Mergeability | `MERGEABLE` — no conflicts |
| `build` | pass |
| `license/cla` | pass |
| `Claude Code Review / claude-review` | fail — superseded by this manual review |

### What is good

- **Legacy path is genuinely untouched.** Every behavioral change is gated on `model.vizModern`: the `showAsLabel` `@let`, the stepper `@if` blocks, the `text-align`/`padding`/`border-radius` overrides, and the early returns in `onMouseDown()` / `enableEditing()` / `isInputDisabled()`.
- **The `commitChange()` extraction is a faithful refactor.** `onClick` previously did `if(refresh && updateOnChange || writeBackDirectly)` and `onBlur` did `if(refresh && updateOnChange)`; `commitChange(this.model.writeBackDirectly)` / `commitChange(false)` reproduce both exactly, including the `&&`-binds-tighter precedence. No semantic drift.
- **Select-then-edit matches the established convention.** `model.editing?`, the `selected` setter reset, `enableEditing()`, `isInputDisabled()` and the `.disable-input { pointer-events: none }` class mirror `vs-text-input.component.ts:100-369` one-for-one. Using the parent `.vs-spinner` div for `(dblclick)` is right, since `pointer-events: none` makes the input hit-transparent so the dblclick still lands.
- **`focusInputIfEditable()` is a real security-relevant detail, correctly handled.** `pointer-events: none` does not block a programmatic `.focus()`, so unconditionally focusing after a stepper click would have handed keyboard editing around the composer gate. The guard and its comment are exactly right.
- **Hold-repeat teardown is sound.** All three timers are cleared in `cancelHold()`, `ngOnDestroy()` calls it, `startHold()` self-cancels first, and `doRepeat()` re-checks the disabled getters each tick — which matters because Chrome fires neither `mouseup` nor `mouseleave` on a button that becomes `disabled` mid-press, so the in-loop guard is the only thing that stops the runaway. Good catch.
- **All SCSS tokens resolve.** `--inet-focus-ring`, `--inet-focus-ring-outline` (`scss/_variables.scss:601-602`), `--inet-text-muted-color`/`--inet-text-subtle-color` (`:270-271`), `--inet-input-bg-color`/`--inet-input-border-color` (`:420-422`), `--inet-ui-neutral-hover-bg-color` (`:525`) all exist.
- **`_#()` inside a binding expression is a supported pattern** (`model.absoluteName + ' _#(Decrease)'`) — the i18n pass is a raw text substitution over the template, and `vs-text-input.component.html:78` uses the identical `absoluteName + ' _#(TextInput)'` form.
- **Test coverage is thorough** — 6 new groups covering the hold state machine, the bounds matrix, the editing gate and the mousedown branch, with the bound-reached cancellation and the fake-timer sequencing modelled correctly.
- No injected HTML, no new network surface, no new user-controlled sink. `aria-label` interpolation is Angular-escaped. Nothing security-relevant.

### Findings

**1. A press-and-hold applies one extra step on release.** (`vs-spinner.component.html:50-52`, `:97-99`)

`click` still fires after `mouseup` even though `mousedown` called `preventDefault()` (`preventDefault` on mousedown suppresses focus and text selection, not the synthesized click). So a hold longer than 300 ms runs `doRepeat()` N times and then `onIncrementClick()` fires once more on release — N+1 steps for N repeats. It is invisible during the accelerated phase but noticeable for a ~350 ms press, which reads as one press and yields two steps. The usual fix is a `didRepeat` flag set in `doRepeat()`, checked-and-cleared in `onIncrementClick()`/`onDecrementClick()`. Note the new tests exercise `startHold`/`cancelHold` and `onIncrementClick` only in isolation, so the real `mousedown → hold → mouseup → click` sequence is not covered.

**2. The steppers mutate the value in the composer with no select-then-edit gate.** (`vs-spinner.component.ts:165-177`)

`onIncrementClick()`/`onDecrementClick()` check only `isIncrementDisabled`/`isDecrementDisabled`, never `isInputDisabled()`. In the composer a single click on `+` on an *unselected* spinner immediately mutates `model.value` and dispatches `/events/composer/viewsheet/vsSpinner/changeValue`, marking the sheet dirty. That is the exact accidental-edit-on-first-click case the gate was introduced to prevent — and the 18 px buttons sit on the assembly edges, right where you click to select or drag. I understand the test plan lists this as intended ("+/- always clickable regardless of selection state"), but it is worth a second look: gating the steppers behind `!isInputDisabled()` in the composer while leaving them always live in the viewer would be self-consistent, and would also make `focusInputIfEditable()` unconditional.

**3. The stepper buttons ignore the assembly's configured format.** (`vs-spinner.component.scss:56-63`)

The input still honours `model.objectFormat.background` / `.foreground` / `.border.*`, but the buttons hardcode `--inet-input-bg-color`, `--inet-text-muted-color` and `1px solid var(--inet-input-border-color)`. Any spinner with an author-set background, foreground or border will render with mismatched buttons flanking a differently-styled input — the "shared border seam" only lines up on the default theme.

**4. `hAlign` / `vAlign` are silently discarded for modern spinners.** (`vs-spinner.component.html:71-72`)

`[style.text-align]` is forced to `'center'` and `[style.padding]` to `null`, dropping both `model.objectFormat.hAlign` and the `getTextVerticalPosition()` vAlign padding. `FormatInfoModel.halignmentEnabled` defaults to `true`, so alignment is user-configurable for spinners — an existing modern spinner set to right-aligned will silently re-center after this change. Centering is clearly the design intent with buttons on both flanks, but consider either disabling the alignment controls for modern spinners or honouring `hAlign` when it is explicitly set. (`getTextVerticalPosition()` is also still evaluated every CD cycle via the `@let` even though modern spinners no longer consume it.)

**5. Modern spinners can no longer show `currentLabel`.** (`vs-spinner.component.html:19`)

`showAsLabel = showLabel && !model.vizModern` means the `.text-label` branch is now dead for modern spinners, so `VSNumericRangeModel.getCurrentLabel()` (i.e. `NumericRangeVSAssemblyInfo.getSelectedLabel()`, which can carry server-side formatting) is never rendered — the raw `type="number"` value shows instead. Almost certainly the desired outcome, but it is a behavior change for existing modern spinners with a format applied, and it leaves `(mouseleave)="showLabel = true"` on the input as dead state in the modern path.

### Nits

- `onMouseDown()`, `onClick()`, `enableEditing()`, `isIncrementDisabled`, `isDecrementDisabled` dereference `this.model` unguarded, while `isInputDisabled()` mixes `this.model.vizModern` with `this.model?.editing` in the same expression. The template's `@if (model)` makes it safe today; the inconsistent optional chaining just reads as an oversight.
- `startHold()` is bound to raw `mousedown` with no `event.button === 0` check, so a right- or middle-button press also arms the repeat; if the context menu swallows the `mouseup`, the repeat runs until the pointer leaves the button.
- `[disabled]` and `[attr.aria-disabled]` are both set on the buttons. A natively `disabled` button is already exposed as disabled and removed from the tab order, so `aria-disabled` is redundant (and generally discouraged alongside `disabled`).
- Only `mousedown`/`mouseup`/`mouseleave` are handled — no pointer/touch events. Tap still works via synthesized mouse events, but touch press-and-hold does not repeat. `pointerdown`/`pointerup` + `setPointerCapture` would also fix the mouseleave-cancels-and-never-resumes behavior.
- `.disable-actions-fade` on the buttons stacks with `&:disabled { opacity: 0.6 }`, so a disabled assembly double-fades the steppers.
- Dragging inside the input in the composer will now move the assembly instead of selecting text, since `onMouseDown` no longer stops propagation for modern spinners and `VSObjectMoveHandle.getMoveHandle("VSSpinner")` returns `null` (so `interactableAllowFrom` is unrestricted). This matches `vs-text-input`, which has no mousedown handler at all, so it is consistent — just noting the trade-off that comes with the change.
- `vs-spinner.component.tl.spec.ts` contains pure `new VSSpinner(...)` unit tests with no `render()` and no MSW, which per `CLAUDE.md` belongs in `*.spec.ts`. More importantly, CI's `npm run test` is `test:portal && test:em && test:em:tl` — `test:portal:tl` is not in it, so none of the 20+ new tests actually run in the pipeline. Pre-existing for this file, but worth resolving before the suite grows further.

### Regression / compatibility assessment

- Legacy spinner: `onClick`/`onBlur` dispatch logic, `stopPropagation`, alignment, padding, border radius and the hover-to-reveal label are all bit-for-bit preserved. No regression.
- `VSSpinnerModel.editing?` is optional and client-only (no server counterpart), so no wire-format or backward-compatibility impact.
- No existing test asserts on the spinner DOM structure, and no e2e selector references `.vs-spinner` internals, so the added buttons break nothing.
- Existing test suites: `spinner-actions.spec.ts` only touches the action model and is unaffected.

None of the above blocks the merge — approving. Findings 1 and 2 are the ones I would most like to see addressed, either here or as a follow-up on this feature branch.

PR #4903, opened 2026-08-31
Seed form-input control height from density at creation

swaker854:
## Summary
- Stacked on #4902 — depends on that PR merging first.
- Adds `VSDensityDefaults.controlHeight()`/`controlHeightForMode()`, returning 24/28/30px for dense/compact/comfortable. Unlike row/header/title height, dense does **not** equal the legacy `defh` (20px) here: a standalone control reads as cramped at the tightest data-row height, so it steps up even at dense, matching the browser's `--inet-viz-control-height` token.
- Applied once, at creation, in `seedChromeDefaults()` for the four form-input types that already bypass the base chrome hook for their own modern-gated round corner: Spinner, ComboBox, CheckBox, and TextInput.
- Only substitutes a control still at its own type's legacy default dimension — an author-resized control is never touched.
- CheckBox is the one variation: its legacy default is `2 * defh` (title lane + one data row), so it scales both rows together rather than substituting a single control height.

## Test plan
- [x] `VSDensityDefaultsTest`: unit tests for `controlHeight`/`controlHeightForMode` across all three density tiers, the off-gate/legacy-context cases, and the unrecognized-mode fallback
- [x] `ControlHeightFollowDensityTest` (new): verifies each of the four widget types seeds the density height only when marked modern, leaves an unmarked/legacy assembly at `defh`, and never touches an already-resized control; verifies CheckBox's 2x ratio is preserved
- [x] `mvnw -pl core compile -am` and `mvnw -pl core test -Dtest=VSDensityDefaultsTest,ControlHeightFollowDensityTest` both pass (37/37)

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/CheckBoxVSAssemblyInfo.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/CheckBoxVSAssemblyInfo.java
@@ -451,6 +451,13 @@ public class CheckBoxVSAssemblyInfo extends ListInputVSAssemblyInfo
          objFormat.getDefaultFormat().setRoundCornerValue(
             ctx.modern ? VSObjectChromeDefaults.cardCornerRadius() : 0);
       }
+
+      // legacy default is 2 * defh (title lane + one data row); preserve that ratio rather than
+      // substituting a single control height, or a freshly-created checkbox would lose the room
+      // its second row needs.
+      if(ctx.modern && getPixelSize().height == 2 * AssetUtil.defh) {
+         setPixelSize(new Dimension(getPixelSize().width, 2 * VSDensityDefaults.controlHeight(ctx)));
+      }
    }
 
    /**
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/CheckBoxVSAssemblyInfo.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/CheckBoxVSAssemblyInfo.java
@@ -434,9 +434,9 @@ public class CheckBoxVSAssemblyInfo extends ListInputVSAssemblyInfo
    }
 
    /**
-    * Seed the modern-gated round corner. This type bypasses the base chrome hook (see
-    * VSAssemblyInfo.bypassesBaseChrome()) so it seeds its own — form-input modernization,
-    * tracked as its own follow-on project from the card-corner work.
+    * Seed the modern-gated round corner and control height. This type bypasses the base chrome
+    * hook (see VSAssemblyInfo.bypassesBaseChrome()) so it seeds its own — form-input
+    * modernization, tracked as its own follow-on project from the card-corner work.
     */
    @Override
    protected void seedChromeDefaults(VizContext ctx) {
@@ -454,33 +454,28 @@ public class CheckBoxVSAssemblyInfo extends ListInputVSAssemblyInfo
 
       // legacy default is 2 * defh (title lane + one data row); preserve that ratio rather than
       // substituting a single control height, or a freshly-created checkbox would lose the room
-      // its second row needs. cellHeight and titleHeight must grow by the same amount as the
-      // container, or updateDataRowCol()'s (containerHeight - titleHeight) / cellHeight would
-      // recompute a second row out of the container's new headroom instead of leaving it as
-      // clearance.
+      // its second row needs. Title height is deliberately left alone - CheckBox is one of the
+      // types TitleLaneHeightRowTest.excludedTypesNeverTakeTheDensityRow pins to never follow the
+      // density row, so cellHeight alone absorbs the container's growth (container - the
+      // still-legacy title height), or updateDataRowCol()'s (containerHeight - titleHeight) /
+      // cellHeight would recompute a second row out of the container's new headroom instead of
+      // leaving it as clearance.
       if(ctx.modern && getPixelSize().height == 2 * AssetUtil.defh) {
-         int controlHeight = VSDensityDefaults.controlHeight(ctx);
-         setPixelSize(new Dimension(getPixelSize().width, 2 * controlHeight));
+         int newHeight = 2 * VSDensityDefaults.controlHeight(ctx);
+         setPixelSize(new Dimension(getPixelSize().width, newHeight));
 
          if(getCellHeight() == AssetUtil.defh) {
-            setCellHeight(controlHeight);
-         }
-
-         if(!isUserTitleHeight() && getTitleHeight() == getLegacyTitleHeight()) {
-            setTitleHeight(controlHeight);
+            setCellHeight(newHeight - getTitleHeight());
          }
       }
       else if(!ctx.modern && getPixelSize().height % 2 == 0 &&
          VSDensityDefaults.isControlHeight(getPixelSize().height / 2))
       {
+         int oldHeight = getPixelSize().height;
          setPixelSize(new Dimension(getPixelSize().width, 2 * AssetUtil.defh));
 
-         if(VSDensityDefaults.isControlHeight(getCellHeight())) {
+         if(getCellHeight() == oldHeight - getTitleHeight()) {
             setCellHeight(AssetUtil.defh);
-         }
-
-         if(!isUserTitleHeight() && VSDensityDefaults.isControlHeight(getTitleHeight())) {
-            setTitleHeight(getLegacyTitleHeight());
          }
       }
    }
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/CheckBoxVSAssemblyInfo.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/CheckBoxVSAssemblyInfo.java
@@ -454,14 +454,34 @@ public class CheckBoxVSAssemblyInfo extends ListInputVSAssemblyInfo
 
       // legacy default is 2 * defh (title lane + one data row); preserve that ratio rather than
       // substituting a single control height, or a freshly-created checkbox would lose the room
-      // its second row needs.
+      // its second row needs. cellHeight and titleHeight must grow by the same amount as the
+      // container, or updateDataRowCol()'s (containerHeight - titleHeight) / cellHeight would
+      // recompute a second row out of the container's new headroom instead of leaving it as
+      // clearance.
       if(ctx.modern && getPixelSize().height == 2 * AssetUtil.defh) {
-         setPixelSize(new Dimension(getPixelSize().width, 2 * VSDensityDefaults.controlHeight(ctx)));
+         int controlHeight = VSDensityDefaults.controlHeight(ctx);
+         setPixelSize(new Dimension(getPixelSize().width, 2 * controlHeight));
+
+         if(getCellHeight() == AssetUtil.defh) {
+            setCellHeight(controlHeight);
+         }
+
+         if(!isUserTitleHeight() && getTitleHeight() == getLegacyTitleHeight()) {
+            setTitleHeight(controlHeight);
+         }
       }
       else if(!ctx.modern && getPixelSize().height % 2 == 0 &&
          VSDensityDefaults.isControlHeight(getPixelSize().height / 2))
       {
          setPixelSize(new Dimension(getPixelSize().width, 2 * AssetUtil.defh));
+
+         if(VSDensityDefaults.isControlHeight(getCellHeight())) {
+            setCellHeight(AssetUtil.defh);
+         }
+
+         if(!isUserTitleHeight() && VSDensityDefaults.isControlHeight(getTitleHeight())) {
+            setTitleHeight(getLegacyTitleHeight());
+         }
       }
    }
 
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/CheckBoxVSAssemblyInfo.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/CheckBoxVSAssemblyInfo.java
@@ -458,6 +458,11 @@ public class CheckBoxVSAssemblyInfo extends ListInputVSAssemblyInfo
       if(ctx.modern && getPixelSize().height == 2 * AssetUtil.defh) {
          setPixelSize(new Dimension(getPixelSize().width, 2 * VSDensityDefaults.controlHeight(ctx)));
       }
+      else if(!ctx.modern && getPixelSize().height % 2 == 0 &&
+         VSDensityDefaults.isControlHeight(getPixelSize().height / 2))
+      {
+         setPixelSize(new Dimension(getPixelSize().width, 2 * AssetUtil.defh));
+      }
    }
 
    /**
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/ComboBoxVSAssemblyInfo.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/ComboBoxVSAssemblyInfo.java
@@ -426,6 +426,9 @@ public class ComboBoxVSAssemblyInfo extends ListInputVSAssemblyInfo {
       if(ctx.modern && getPixelSize().height == AssetUtil.defh) {
          setPixelSize(new Dimension(getPixelSize().width, VSDensityDefaults.controlHeight(ctx)));
       }
+      else if(!ctx.modern && VSDensityDefaults.isControlHeight(getPixelSize().height)) {
+         setPixelSize(new Dimension(getPixelSize().width, AssetUtil.defh));
+      }
    }
 
    /**
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/ComboBoxVSAssemblyInfo.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/ComboBoxVSAssemblyInfo.java
@@ -422,6 +422,10 @@ public class ComboBoxVSAssemblyInfo extends ListInputVSAssemblyInfo {
          objFormat.getDefaultFormat().setRoundCornerValue(
             ctx.modern ? VSObjectChromeDefaults.cardCornerRadius() : 0);
       }
+
+      if(ctx.modern && getPixelSize().height == AssetUtil.defh) {
+         setPixelSize(new Dimension(getPixelSize().width, VSDensityDefaults.controlHeight(ctx)));
+      }
    }
 
    /**
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/RadioButtonVSAssemblyInfo.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/RadioButtonVSAssemblyInfo.java
@@ -401,9 +401,16 @@ public class RadioButtonVSAssemblyInfo extends ListInputVSAssemblyInfo
    }
 
    /**
-    * Seed the modern-gated round corner. This type bypasses the base chrome hook (see
-    * VSAssemblyInfo.bypassesBaseChrome()) so it seeds its own — form-input modernization,
-    * tracked as its own follow-on project from the card-corner work.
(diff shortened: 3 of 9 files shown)


swaker854 (commented):
## Code Review — PR #4903: Seed form-input control height from density at creation

### CI status (investigated, not blocking)
The `build` check failure is unrelated to this diff, which touches only Java files (`VSDensityDefaults.java`, the four `*VSAssemblyInfo.java` files, and two test files):
```
FAIL portal projects/portal/src/app/common/util/gui-tool.spec.ts > GuiTool density mode
     > defaults to dense when no density class is present
AssertionError: expected 'compact' to be 'dense'
```
This is the same pre-existing flaky test flagged in PR #4902's review and in this session's own prior handoff notes — `gui-tool.spec.ts` isn't in this PR's diff at all. `claude-review` is excluded per review instructions.

### Blocking issue

**Reverting a modernized assembly does not restore its legacy height — breaking the method's own documented reversibility contract.**

`VSAssemblyInfo.seedChromeDefaults()`'s own Javadoc states the design invariant this method must follow:

> "with the mark cleared, seedChromeDefaults writes the legacy branch of every ternary, which is the identical call a gate-off creation makes. A property added to what Modernize does is therefore reverted by the same edit to the same method, **or it is not added**."

The existing round-corner line honors this by construction — `ctx.modern ? cardCornerRadius() : 0` runs unconditionally on every call and writes to the format's DEFAULT tier, so `VizModernizeUtil.revert()` (which calls `seedChromeDefaults()` again with `ctx.modern = false`) naturally restores it, and a genuine user override survives untouched because it lives in a separate, higher-priority format tier.

The new height line is one-directional and mutates a flat, untiered value:
```java
if(ctx.modern && getPixelSize().height == AssetUtil.defh) {
   setPixelSize(new Dimension(getPixelSize().width, VSDensityDefaults.controlHeight(ctx)));
}
```
There's no `else` branch. Walk through `VizModernizeUtil.revert()` on a spinner that was modernized under `comfortable` density: `getPixelSize().height` is now `30`, not `defh`. `revert()` sets `vizMark = null` and calls `seedChromeDefaults(VizContext.of((VizMark) null))` — `ctx.modern` is `false`, so this whole block is skipped. The round corner correctly goes back to `0`, but the height stays at `30` forever. `Revert` is a real, live, user-facing feature per `VizModernizeUtil`'s own class doc ("an author in a modern org can keep one dashboard classic") — this isn't a theoretical gap.

I recognize this is a harder problem than round corner's unconditional ternary, precisely because pixel size has no format-tier equivalent to distinguish "still at the density-substituted value" from "user resized it to something that happens to match a density step" — so a correct fix needs to actually address that ambiguity (e.g., checking against all three known `controlHeightForMode` outputs before reverting, accepting the same false-positive risk the forward direction already accepts by comparing against a single literal), not just add a mirrored `else`. Given the method's own doc treats "can't cleanly revert" as a reason a property "is not added," this is worth resolving or explicitly discussing before merge rather than shipping the asymmetry silently.

Confirmed there's no existing coverage of this interaction: neither `VizModernizeUtilTest` nor the new `ControlHeightFollowDensityTest`/`VSDensityDefaultsTest` additions exercise revert-after-modernize for any of the four touched types.

### Non-blocking observations

- **`CheckBoxVSAssemblyInfo`'s `2 * controlHeight(ctx)`** is a reasonable pragmatic choice for a one-time creation seed (checkbox's title lane and one data row don't have their own dedicated density functions the way table title/row height do), but it's worth a quick visual sanity check — it assumes the title lane and the single data row scale identically, which may not exactly match checkbox's actual internal rendering split. Not blocking since this is only the initial seed for a brand-new object most authors will resize anyway.
- **Test coverage is otherwise thorough and well-targeted** — `ControlHeightFollowDensityTest`'s three-case pattern (follows density / unaffected when unmarked / left alone when already resized) is applied to Spinner and CheckBox, but only the "follows density" case is checked for ComboBox and TextInput. Given `VSDensityDefaults.controlHeight()` itself is shared and already unit-tested independently, this is a minor gap rather than a real risk — but the "already resized" guard is the one most worth having per-type coverage for, since it's the one that protects an author's explicit choice.
- **Verified this PR correctly does *not* extend to `RadioButtonVSAssemblyInfo`/`SubmitVSAssemblyInfo`**, even though both appear in `VSAssemblyInfo.bypassesBaseChrome()`'s list. Checked both: neither actually has a `seedChromeDefaults()` override or modern-gated round corner today (RadioButton's `setDefaultFormat()` never calls it; Submit hardcodes `setRoundCornerValue(3)` unconditionally) — so they're not yet part of the "four form-input types that already bypass the base chrome hook for their own modern-gated round corner" this PR targets. Correctly out of scope, not an oversight.

### What's solid
- `controlHeightForMode`'s dense value (24px, not `defh`/20px) is a deliberate, well-documented exception to the "dense == legacy" parity the rest of the class enforces, and the reasoning (a standalone control reads as cramped at data-row height) is sound and matches the `--inet-viz-control-height` CSS token exactly.
- The "only substitute at the type's own legacy default dimension" guard is correctly type-specific (`AssetUtil.defh` for three types, `2 * AssetUtil.defh` for CheckBox matching its actual constructor value), so an author-resized control is never silently overwritten going *forward*.
- Unit tests for `controlHeight`/`controlHeightForMode` mirror the existing `rowHeight`/`headerRowHeight` test patterns closely, keeping the file internally consistent.
- Verified `mvnw -pl core compile -am` and the two touched test classes pass locally (37/37).

---
**Review verdict: would be Request Changes** (self-authored PR — GitHub does not allow a Request Changes/Approve review on your own PR, so this is posted as a comment).

jshobe-inetsoft (changes requested):
## Code Review — Seed form-input control height from density at creation

Reviewed the diff (7 files, +217/-7) plus the current CI results for this PR.

### Summary of change
Adds `VSDensityDefaults.controlHeight()`/`controlHeightForMode()` (24/28/30px for dense/compact/comfortable) and applies it once, at `seedChromeDefaults()` time, to `SpinnerVSAssemblyInfo`, `ComboBoxVSAssemblyInfo`, `TextInputVSAssemblyInfo`, and `CheckBoxVSAssemblyInfo` (2× for its title+row legacy ratio) — the four form-input types that already bypass the base chrome hook for their own modern-gated round corner.

### Correctness
- The implementation follows the existing, well-established pattern in `VSDensityDefaults`/`VSAssemblyInfo` (row/header/title height) very closely, including the "only substitute while still at the type's own legacy default dimension" guard that protects an author-resized control from being touched. Verified against `AssetUtil.defh` (20) and each type's constructor-set default (`Spinner`/`ComboBox`/`TextInput` = `defh`, `CheckBox` = `2 * defh`) — the guard conditions are correct for each type.
- `CheckBoxVSAssemblyInfo`'s special-cased 2× ratio is correct and well explained in both the code comment and the new test (`checkBoxHeightScalesByItsLegacyTwoRowRatio`).
- The guard is idempotent: once a control's height has been bumped to a density value, it will never again equal the legacy `defh`/`2*defh` check, so a later re-seed (e.g. via Modernize) won't re-trigger it. Confirmed by `spinnerHeightIsLeftAloneWhenAlreadyResized` / `checkBoxHeightIsLeftAloneWhenAlreadyResized`.
- New Java tests (`VSDensityDefaultsTest`, `ControlHeightFollowDensityTest`, 37 tests total) all pass in CI.
- All required imports (`AssetUtil`, `java.awt.*`) are present/added correctly (`TextInputVSAssemblyInfo` needed a new `AssetUtil` import and got it); no compile issues.

### Completeness / minor observations (non-blocking)
- `RadioButtonVSAssemblyInfo` and `SubmitVSAssemblyInfo` also bypass the base chrome hook (see `VSAssemblyInfo.bypassesBaseChrome()`) but are intentionally left out of this PR's scope per the description ("the four form-input types"). Worth a quick sanity check that this is a deliberate follow-on and not an oversight, since RadioButton in particular is a "standalone control" in the same sense as the others.
- The height-equality guard (`getPixelSize().height == AssetUtil.defh`) is a heuristic, not an explicit "user set this" flag (unlike `isUserTitleHeight()` for title height). An author who deliberately resizes a control back to exactly the legacy default would have it silently bumped on the next Modernize pass. This is a pre-existing, accepted tradeoff already used for row/header height elsewhere in the class, so it's consistent with existing behavior rather than a new risk introduced here.
- The seeding also fires again whenever `Modernize` re-runs `seedChromeDefaults()` on a pre-existing (not just newly-created) assembly whose height happens to still equal the legacy default — i.e., this isn't purely a "creation-time" change despite the PR title, though this matches the documented, existing semantics of `seedChromeDefaults()` for the other height matrices.
- Minor test nit: `aModernContextYieldsItsDensityControlHeight` sets `viewsheet.modernVisualization=true`, but `VizContext.of(VizMark)` (used by the production code path here) doesn't consult that gate property at all — only the mark matters. The property-set is harmless but not actually exercising anything; not worth blocking on.
- The 3-line control-height block is duplicated verbatim across the four `seedChromeDefaults()` overrides. This mirrors the existing duplication pattern already used for the round-corner seed in these same classes, so it's consistent with the codebase's established (if not ideal) style rather than new debt.

No new error-handling surface — pure integer sizing logic. Unrecognized density mode falls back to dense (24px), consistent with the other `*ForMode` methods and covered by a test.

### Security
No security-relevant surface (no external input, no serialization/deserialization changes, no new persisted fields).

- No API signature changes, no XML/serialization format changes.
- Behavioral change is scoped to newly-created (or previously-untouched, pre-Modernize) instances of the four control types when the modern-visualization gate is on and the type is marked modern — legacy/gate-off paths are untouched (`ctx.modern` false ⇒ `AssetUtil.defh`).
- No indication that any other code path hard-codes the previous fixed heights for these four control types in a way this change would break; nothing found in a search of the diff area suggesting otherwise, though a full audit of every consumer of these assemblies' pixel size (e.g., layout/export code) was not exhaustively performed.

### Tests
- All 37 new/modified Java unit tests pass in CI (`VSDensityDefaultsTest`: 30 tests, `ControlHeightFollowDensityTest`: 7 tests).
- No existing test appears to assert the old fixed height for these four types in a way this change would break.

### CI / mergeability status
- **Validation Build (`build`) check is currently FAILING.** The Java build and all new/existing Java tests pass; the failure is in the frontend step (`npm run verify`), specifically a pre-existing Vitest failure: `projects/portal/src/app/common/util/gui-tool.spec.ts > GuiTool density mode > defaults to dense when no density class is present` (`expected 'compact' to be 'dense'`). This failure is **not caused by this PR's diff** (which touches only Java files under `core/`) — it is inherited from the base branch, PR #4902 (`0830-spinner-stepper-controls`), which has the identical failure in its own `build` check. It should resolve automatically once #4902 fixes/merges, but as it stands it is a genuine failing check on this PR (excluding the known-broken `claude-review` check, which is being disregarded per current instructions).
- `mergeable: MERGEABLE`, no merge conflicts.
- This PR is stacked on, and cannot be merged before, #4902.

### Verdict
Per review policy: the code itself has no blocking issues and is a clean, well-tested, well-scoped change consistent with existing patterns. However, there is a genuine failing check (`build` / Validation Build) on this PR that is not the disregarded `claude-review` check, so per the stated approval criteria this PR does **not** qualify for approval yet.

**Requesting changes** — not because of a defect in this PR's own code, but because the `build` check is currently failing (inherited from base branch #4902's own frontend test failure) and the PR is stacked on that unmerged, currently-failing base. Once #4902's build is green (and merges), this PR should be re-checked and should be approvable as-is.

swaker854 (commented):
## Follow-up: revert bug fixed

Confirmed commit \`fdae757f\` (\"fix: restore legacy control height on Revert\") fully addresses the blocking issue raised in my earlier review (reverting a modernized control didn't restore its legacy height).

**Verified the fix is correct and symmetric across all four touched types:**
- Adds \`VSDensityDefaults.isControlHeight(int)\`, matching any of the three density-derived heights (24/28/30) at any tier — the same "accept a narrow false positive rather than leave every modernized control permanently changed" tradeoff the round-corner ternary already relies on, exactly as I'd suggested.
- \`SpinnerVSAssemblyInfo\`/\`ComboBoxVSAssemblyInfo\`/\`TextInputVSAssemblyInfo\` each got an identical \`else if(!ctx.modern && isControlHeight(height))\` branch reverting to \`AssetUtil.defh\`.
- \`CheckBoxVSAssemblyInfo\`'s revert branch correctly handles its 2x ratio: guards on \`height % 2 == 0\` before checking \`isControlHeight(height / 2)\`, so an odd custom height is never touched (verified against \`AssetUtil.defh = 20\`, which doesn't collide with 24/28/30, so the legacy value is never mistaken for a density value).
- New tests cover both the restore-on-revert case and a genuinely-custom-height-survives-revert case for Spinner and CheckBox, plus dedicated \`isControlHeight\` unit tests for all three tiers and the rejection case.

**Remaining minor gap (still non-blocking):** ComboBox and TextInput still only have the forward "follows density" test — no revert-specific or already-resized test for either, unlike Spinner/CheckBox. Since the revert logic itself is identical shared code already exercised via Spinner/CheckBox and the standalone \`isControlHeight\` tests, this is a coverage nicety rather than a real risk, consistent with @jshobe-inetsoft's earlier non-blocking note about the asymmetric test coverage.

**CI is now green:** \`build\` passes (the failure at review time was the base branch's — #4902's — own pre-existing \`gui-tool.spec.ts\` flake, unrelated to this diff, and has since resolved). Semgrep/license checks pass; only \`claude-review\` fails, which is disregarded per review policy.

From my side, both raised concerns (the revert bug and the CI failure) are now resolved. Still blocked on: this PR is stacked on #4902 (unmerged), and @jshobe-inetsoft's Changes Requested review hasn't been re-reviewed against the fix commit yet.

jshobe-inetsoft (changes requested):
## Code Review — Seed form-input control height from density at creation

Reviewed against the stacked base `0830-spinner-stepper-controls` (#4902 must merge first). I traced `seedChromeDefaults`'s full call graph (`VSAssemblyInfo.setDefaultFormat` → creation only, plus `VizModernizeUtil.modernize()/revert()`), confirmed the XML parse funnel never reaches it, and checked the rendered consequences on the browser side (`VSCheckBoxModel`, `vs-compound.ts`, `_viz-tokens.scss`).

The good parts are genuinely good: the 24/28/30 matrix matches `--inet-viz-control-height` exactly; creation-only application is correctly established (parse and paste never call `initDefaultFormat`, and `AbstractVSAssembly`'s two-arg constructor stamps the mark before `initDefaultFormat` runs, so the mark is visible when the hook fires); the `% 2 == 0` guard on CheckBox's revert branch is genuinely necessary (integer division would otherwise revert 49 via 49/2 == 24) and is pinned by a test; and the `isControlHeight()` false-positive trade-off is honestly documented rather than hidden.

**Requesting changes** — the CheckBox branch changes rendered layout in a way the PR does not intend, and two sibling types that match the PR's own stated selection rule are missing.

---

### Blocking

**1. Doubling the CheckBox container without moving `cellHeight` produces dead space — and, at comfortable, an extra data row** — `CheckBoxVSAssemblyInfo.seedChromeDefaults`

The comment says the 2× scaling exists so "*a freshly-created checkbox would [not] lose the room its second row needs*", i.e. title lane + one data row both step up. Neither of those two lanes actually follows density:

* `CheckBoxVSAssemblyInfo.getTitleHeight()` → `titleInfo.getTitleHeight()`, which defaults to `AssetUtil.defh`. Checkbox is **not** among the types routed through `VSDensityDefaults.titleHeight(info, stored)` (only calendar, chart, current-selection, selection base and table-data are).
* `ListInputVSAssemblyInfo.cellHeight` is a plain field defaulting to `AssetUtil.defh`, untouched by this PR.

`VSCheckBoxModel` then computes, with the title visible (the default):

```
contentHeight = height - titleHeight(20)
dataRowCount  = max(1, floor(contentHeight / cellHeight(20)))
dataColCount  = dataRowCount >= values.length ? 1 : ceil(values.length / dataRowCount)
```

| density | seeded height | content | rows | result |
|---|---|---|---|---|
| legacy | 40 | 20 | 1 | 1 row, no slack (today) |
| dense | 48 | 28 | 1 | 8px of dead space below the row |
| compact | 56 | 36 | 1 | 16px of dead space |
| **comfortable** | **60** | **40** | **2** | **relayout: items wrap into 2 rows × ceil(n/2) columns** |

`vs-compound.ts`'s `getDataTopCSS`/`getDataLeftCSS` position cells straight off `dataColCount`, so the comfortable case is a real, visible layout change for a freshly created (or Modernized) checkbox, not just extra padding. And in every tier the rows themselves stay 20px tall — the control does not get taller, which is the stated goal.

Please either seed `setCellHeight(VSDensityDefaults.controlHeight(ctx))` (and, if the title lane is meant to grow, route CheckBox's title height through `VSDensityDefaults.titleHeight`) alongside the container change, or drop CheckBox from this PR until those two lanes follow density. Whichever way it goes, it needs a test asserting `dataRowCount` stays 1 across all three tiers.

**2. `RadioButtonVSAssemblyInfo` (and `SubmitVSAssemblyInfo`) match the stated rule but are excluded.**

The PR describes the target set as "*the four form-input types that already bypass the base chrome hook for their own modern-gated round corner*". That description picks out six types, not four: `RadioButtonVSAssemblyInfo:407` and `SubmitVSAssemblyInfo:78` both bypass `bypassesBaseChrome()` and seed their own `cardCornerRadius()` in exactly the same shape. Radio button is the same base class as checkbox (`ListInputVSAssemblyInfo`) with the identical `2 * AssetUtil.defw × 2 * AssetUtil.defh` default, so after this change a modern comfortable dashboard shows a 60px checkbox next to a 40px radio button — two sibling controls with identical legacy geometry now visibly disagree. Either include them or state in the code why they are deliberately deferred (I found no follow-on PR in the stack).

---

### Non-blocking

**3. Baking the density value into the stored `Dimension` departs from how the rest of this feature works.** Row/header/title height are resolved at *read* time (`VSDensityDefaults.titleHeight(info, stored)`, `rowHeight(ctx)` in `VSTableLens`/`BaseTableService`), so they follow the org's density live and need no undo path. Control height is the first one persisted into the model, which produces two consequences worth confirming as intended:
   * A density change in EM re-flows tables and title lanes but leaves every existing form control frozen at its creation-time tier — nothing re-invokes `seedChromeDefaults` on an already-marked assembly, so a mixed-density dashboard is permanent.
   * Revert needs the lossy `isControlHeight()` value-sniff precisely because there is no record of what was seeded.

   A persisted "height was density-seeded" flag (the way `TitleInfo.userTitleHeight` records authorship) would make both exact. Not for this PR necessarily, but the choice deserves to be recorded somewhere other than a PR description.

**4. `isControlHeight()`'s documented false positive understates CheckBox's window.** The javadoc names 24/28/30. For CheckBox the effective revert window is 48/56/60 — and 48 and 60 are far more plausible values for an author to have set by hand than 24 or 30. Worth spelling out at the CheckBox call site.

**5. The same six-line block is copy-pasted into four classes.** A `VSDensityDefaults.applyControlHeight(info, ctx, rows)` helper would keep the `% 2` subtlety and the legacy-default comparison in one place, and would make adding radio button/submit (point 2) a one-liner each.

**6. Test coverage gaps.** No case creates at the **dense** tier (the interesting one, since it is the tier that deliberately breaks defh parity — the closest is `denseControlHeightIsNotLegacyDefh`, which tests the resolver, not the seeding); and revert is only exercised on Spinner and CheckBox, not ComboBox or TextInput, even though the branch is duplicated per class rather than shared.

---

### Checks / mergeability

* `build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla` — all pass.
* `claude-review` — fail, superseded by this manual review.
* Mergeable: yes, no conflicts (`MERGEABLE`). Note the base is `0830-spinner-stepper-controls`, so #4902 has to land first.

No security concerns — the change is confined to default geometry seeding, reads no external input, and adds no new persistence or deserialization surface.

jshobe-inetsoft (approved):

PR #4904, opened 2026-08-31
feat(tabular): let a non-relational connector describe its own catalog for annotation

Nelson-cao:
## Problem

wiz cannot annotate a non-relational (tabular) data source at all. Both endpoints the annotation
pipeline uses start with `MetadataApiService.getJDBCDatasource()`, which throws
`UnsupportedDatasourceException` for anything that is not a `JDBCDataSource` — so an OData source is
refused at submit time and never reaches the pipeline.

## Fix

A connector describes its own catalog through a new optional SPI, and the two existing endpoints
learn to ask it when the source is not relational.

- **`TabularCatalogProvider`** (`uql/tabular`), implemented by a connector's `TabularRuntime` — a
  catalog is a property of the data source, not of any one query. Two phases, stateless between them
  because they are separate HTTP requests: `listDatasets(ds)` and `describeDataset(ds, datasetId)`.
  A separate interface rather than default methods on `TabularRuntime`: `getMetaData`'s `return null`
  default makes "not implemented", "nothing there" and "the call failed" indistinguishable, so
  implementations opt in at compile time and throw rather than return an empty result.
- **Five neutral records** (`TabularCatalog`, `TabularDatasetRef`, `TabularDatasetSchema`,
  `TabularColumn`, `TabularRelationship`) carry the result across the plugin boundary. No relational
  vocabulary, and column types are `XSchema` constants rather than `XTypeNode` objects — that class
  carries a mutable untyped attribute bag through which connector knowledge could reach core with no
  compiler signal.
- **`TabularCatalogService`** holds the resolution sequence and the neutral-to-wire mapping,
  including the dataset COMMON extension whose `datasourceType: "tabular"` tells wiz this is not a
  SQL table and must not be given a dialect.
- **`MetadataApiService`** gains one `try`/`catch(UnsupportedDatasourceException)` in each of
  `getMetaData` and `getDatabaseTables`, delegating to that service. `getJDBCDatasource` is
  untouched, no statement on the JDBC path moved, and no new route was added — `GET /datasource/tables`
  and `POST /datasource/table/meta` now answer for a tabular source too.
- **OData** implements the SPI from one `$metadata` (EDMX) parse: entity sets, their `<Property>`
  columns and keys, and `<NavigationProperty>` relationships that declare a `<ReferentialConstraint>`.
  Parsed snapshots are held in a **static** Caffeine cache — static because `TabularUtil.createRuntime`
  builds a fresh runtime per request, so an instance field would never be hit. A navigation property
  that cannot be expressed as paired columns is dropped rather than emitted column-less, because wiz
  stores these as `declared` edges that inference may never overwrite.

`TabularUtil.createRuntime(String)` is added as the data-source-level counterpart of the existing
`createQuery(String)`.

Scope is OData. No other connector implements the SPI, and one that does not gets the same 422 it got
before. OData v2/v3 `<Association>`-style relationships are not parsed; such a service yields an empty
relationship list.

## Tests

New: `TabularCatalogServiceTest`, `TabularCatalogDatasetExtensionTest`, `CoreHasNoConnectorKnowledgeTest`,
`MetadataApiServiceNonJdbcBranchTest` (core); `ODataCatalogTest`, `ODataCatalogRelationshipTest`,
`ODataRuntimeCatalogTest`, `ODataCatalogCacheTest` (odata).

Three of those exist to hold a structural claim rather than a behaviour:
`CoreHasNoConnectorKnowledgeTest` reads the compiled `TabularCatalogService.class` and asserts its
constant pool holds no `inetsoft/uql/odata`, `org/w3c/dom`, `TabularQueryParamsSchemaBuilder` or
`TabularQuerySchema` reference; `MetadataApiServiceNonJdbcBranchTest` asserts a JDBC source never
touches the new service at all; `ODataCatalogCacheTest` drives **two separate runtime instances** and
asserts one `$metadata` fetch, which is the case that fails if the cache stops being static.

`TabularCatalogServiceTest` runs entirely on fake connector fixtures invented in the core test tree,
so core's mapping is exercised by something that has never heard of OData.

Full `core` (7876 tests) and `connectors/inetsoft-odata` (26) suites green. The frozen regression set
— `MetadataApiServiceGetJDBCDatasourceTest`, `MetadataApiServiceGetMetaDataTest`,
`MetadataApiServiceNonJdbcTreeTest`, `WizTabularControllerTest`, `WizDatasourceAnnotationClassTest`,
`TabularQueryParamsSchemaBuilderTest`, `TabularQueryParamsSchemaBuilderResolveTagsTest`,
`ODataRuntimeTests` — is green and unedited.

Verified by code tracing and unit tests only; not exercised against a live StyleBI instance or a real
OData service. Design, verification and review are archived in the wiz repo under
`docs/teams/2026-08-28-tabular-metadata-annotation-entry/`, with the accepted degradations and
deferred items in `docs/tabular/stylebi-tabular-wiz-integration.md` section 10.


Nelson-cao:
## Review round 2 — not clean: 1 blocker, 2 important

Reviewed against this repo's own CI review prompt (`.github/workflows/claude-code-review.yml`),
deliberately **without** access to the design archive, so the diff was read by someone who was not
in the room. All three findings below are real and none was caught by the earlier in-branch review.
Fixes are being prepared; this comment is the round's record.

### Blocker — `TabularCatalogService` cannot be created as a Spring bean

`core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java:48,53`

The class is `@Service`, declares two constructors (public 2-arg, package-private 3-arg test seam),
carries **no `@Autowired`**, and has no no-arg constructor. Spring's candidate scan uses
`getDeclaredConstructors()`, so the package-private one counts; with more than one candidate and no
annotation it falls back to default-constructor instantiation, which does not exist here. Bean
creation fails at context startup.

`MetadataApiService` in this same PR has the identical two-constructor shape and does mark its
intended constructor `@Autowired` — so the rule was known, one class just missed it.

Not reachable by the added tests: every test constructs the service directly or mocks it, so nothing
in the suite goes through a real `ApplicationContext`. Fix is `@Autowired` on the 2-arg constructor,
plus a context test that resolves `MetadataApiService` as a real bean — the underlying gap is that no
test exercises Spring wiring at all, which is what let this pass.

### Important — `ODataCatalog` guesses a relationship target where its own comment says it must not

`connectors/inetsoft-odata/src/main/java/inetsoft/uql/odata/ODataCatalog.java:225-236`

When `resolveBindingTarget` finds no `<NavigationPropertyBinding>`, the code falls back to
`typeSimpleNameToFirstEntitySet` — the first entity set exposing that type, in document order.
`resolveBindingTarget`'s own javadoc states the binding is "the only thing that is correct when one
entity type is exposed as several entity sets", which is precisely the case this fallback guesses in.

The result is a structurally valid, semantically wrong relationship rather than an honestly dropped
one, and downstream these are stored as `declared` edges that inference may never overwrite — the
same reasoning this module already uses to justify dropping constraint-less and containment
navigation properties applies here.

Fix: fall back only when the target type has exactly one entity set; when it has more, drop and log,
alongside the four cases already dropped. No fixture currently covers "constraint present, no
binding, target type exposed by 2+ entity sets".

### Important — the id-opacity javadoc reads as the opposite of what is true

`core/src/main/java/inetsoft/uql/tabular/TabularDatasetRef.java:29-49`

The sentence "splits it on `.` for every category except `FILE` — which now includes `METADATA`"
parses most naturally as METADATA being *exempt* from the split. It is not: `bareTableName`
(`wiz-services/src/v1/services/tabularBinding.ts:68-71`) exempts only `FILE`, and nothing was ever
added for METADATA. The comment exists to warn the next connector author about dot collisions, and as
written it tells them the problem is already solved. Being rewritten to state the open case
unambiguously.

### Nits — recorded, not being fixed this round

`toTablesResponse` iterating `catalog.relationships()` without a null guard (unreachable via OData,
which always returns a real list); three `getDataSource` lookups per request where one would do. Both
belong to the same already-documented family: this SPI's contract is held by convention rather than
by validation, which matters when a second connector implements it.

### Independently confirmed sound

Permission checks sit ahead of the new branch in both endpoints; a relational source cannot reach the
tabular path (`getJDBCDatasource` throws only on `!(instanceof JDBCDataSource)`); re-throwing
`UnsupportedDatasourceException` from inside the catch still lands on the controller's own
`@ExceptionHandler` and answers the same 422, so a connector that has not implemented the SPI behaves
exactly as before; the cache is single-flight per key, does not entomb failures, and cannot serve one
caller's data to another (OData credentials live on the data source, not the principal, and every
caller has already passed the same READ check); `CoreHasNoConnectorKnowledgeTest` scans compiled
bytecode rather than source, so it catches a reflective leak too; and the fake fixtures driving
`TabularCatalogServiceTest` genuinely know nothing about OData.

Nelson-cao:
## Review round 3 — clean

All three round-2 findings are fixed in `484ce1df1`. This round was spent trying to break the fixes
rather than re-reading the diff: each one was toggled off, the affected tests re-run to confirm they
actually fail, then restored. No new defect found, and no finding of any severity.

**Spring bean creation.** `@Autowired` on the 2-arg `TabularCatalogService` constructor, plus a new
`MetadataApiServiceSpringWiringTest` standing up a real `AnnotationConfigApplicationContext`.
Reverting only that file reproduces the failure exactly —
`BeanCreationException ... No default constructor found / NoSuchMethodException: TabularCatalogService.<init>()`
— and restoring it passes.

Three things about that test were specifically challenged before accepting it. Registering beans
explicitly rather than by component scan does **not** sidestep the mechanism that failed:
`AnnotationConfigApplicationContext` registers `AutowiredAnnotationBeanPostProcessor` either way, and
`register()` only affects how definitions are discovered, not constructor-candidate resolution during
`refresh()`. The test also cannot pass vacuously — once one constructor carries `@Autowired`, the
5-arg overload is never a candidate, so `getBean(MetadataApiService.class)` succeeding proves the
DI-supplied `TabularCatalogService` bean really was resolved and injected. And a future third
unannotated constructor would correctly leave it green, since one annotated constructor removes the
ambiguity by itself.

**Ambiguous relationship target.** `typeSimpleNameToFirstEntitySet` became a
`Map<String,List<String>>`; the fallback resolves only on a single candidate, drops with a `LOG.debug`
on more than one, and still falls through to the pre-existing no-entity-set drop on zero. Confirmed
this is the map's only consumer, so nothing else needed updating. Reverting only `ODataCatalog.java`
fails exactly the two new tests — the old code really did guess `Invoices_Contact -> Customers`, the
first-declared entity set exposing `Contact`. The `Employee.Manager` case passes both before and
after, which is correct: it is a do-not-over-correct guard, not a discriminator. The expected count
moving 2 to 3 is the legitimate `Employees_Manager` self-loop, not a false positive.

**The id-opacity javadoc.** Rewritten as separate short sentences with a standalone negative — only
`FILE` is exempt from the `.` split; every other category including METADATA is still split; that has
not been fixed. Re-checked against `tabularBinding.ts:68-71` and `tableDocResolver.ts:27-33`, and
`git log` still shows no change to either since before this feature: the gap the comment now names
honestly is genuinely still open, which is what it should say.

The two nits from round 2 (`toTablesResponse` iterating `catalog.relationships()` without a null
guard; three `getDataSource` lookups per request) are confirmed still present and neither better nor
worse — deliberately deferred, not silently dropped. They are recorded as follow-ups in
`docs/tabular/stylebi-tabular-wiz-integration.md` section 10.3, in the same family as the SPI's
contract being held by convention rather than validation, which is what will matter when a second
connector implements it.

Reviewed against this repo's CI review prompt, without access to the design archive — the same
deliberate withholding that produced the round-2 findings.

fredaan (approved):

PR #4905, opened 2026-08-31
fix(wiz): save_worksheet cascades a rename to dependent assets (PVA-011)

fredaan:
## Summary

- `WorksheetAgentController.save()` (the MCP `save_worksheet` endpoint) only called `worksheetService.setWorksheet(...)`, never the `DependencyTransformer.createRenameInfo` / `RenameTransformHandler.addTransformTask` / `initWorksheetOldName` triad the Composer UI's own Save button runs (`SaveWorksheetService.saveWorksheet()`). A column/table renamed via the agent's `rename_column` and saved through this endpoint left every already-saved dependent viewsheet's binding silently pointing at the old name — the tool reported `{ok: true}` with no caveat.
- A plain in-place save now runs the same cascade unconditionally (no user is present in an MCP session to answer the "Dependencies Changed" confirm dialog), using the synchronous `addTransformTask(dinfo, true)` variant so the response is only returned once the cascade has actually landed — avoiding the propagation-lag race class already seen in PVA-008/PVA-009. The response now includes `dependenciesUpdated`/`message` when a cascade ran.
- Scoped the same way as the existing PVA-002/PVA-003 fixes in this method: skipped for Save-As and for a first save out of `TEMPORARY_SCOPE`, since neither can have a pre-existing dependent pointing at the in-place name.
- `ViewsheetAgentController.save()` has the same missing-persistence-step *shape*, but there is no `DependencyTransformer.createRenameInfo(RuntimeViewsheet)` overload, and even the Composer UI's own viewsheet-save path (`SaveViewsheetDialogService`) never calls `RenameTransformHandler` for a viewsheet save — there is no existing UI mechanism to mirror there, so it's out of scope for this PR.

Bug: `docs/bug-reports/plugin-composer-viewsheet-assembly-lifecycle.md#pva-011` in `stylebi-wiz`.

## Test plan

- [x] `WorksheetAgentControllerTest` — added `plainSaveInPlaceCascadesRenamedColumnToDependentAsset`, `plainSaveInPlaceDoesNotCascadeWhenNoColumnWasRenamed`, `saveAsDoesNotCascadeRenames`; updated the existing `plainSaveInPlaceResetsConnectedViewsheetRuntimes` to stub the now-unconditional dependency lookup. `mvn -pl community/core test -Dtest=WorksheetAgentControllerTest` — 91/91 pass.
- [x] `WorksheetScriptControllerTest` — updated for the new constructor parameter. 7/7 pass.
- [x] Full `inetsoft.web.wiz.**` suite — all green, no regressions.


fredaan (commented):
test-probe (will be replaced)

fredaan (commented):
**Round 1 — review**

(Disregard the "test-probe" comment above — posted by mistake while confirming self-review permissions on this PR; superseded by this one.)

Scope checked: the diff (`WorksheetAgentController.java`, `WorksheetAgentControllerTest.java`, `WorksheetScriptControllerTest.java`) against the diagnosis in `stylebi-wiz`'s `docs/bug-reports/plugin-composer-viewsheet-assembly-lifecycle.md#pva-011`, CLAUDE.md, and the current source of `DependencyTransformer`/`RenameTransformHandler`/`SaveWorksheetService` on this branch.

**Code review: no blocking issues found.**

- Correctness: the cascade triad (`createRenameInfo` → `addTransformTask(dinfo, true)` → `initWorksheetOldName`) sits inside the same `!isSaveAs && !wasTemporary` guard, *before* the PVA-002 `resetRuntime()` loop. That matches `SaveWorksheetService.saveWorksheet()`'s own call order (`core/src/main/java/inetsoft/web/composer/ws/service/SaveWorksheetService.java:84-90`) and correctly ensures the reset picks up a just-rewritten dependent binding rather than a stale one.
- `addTransformTask(dinfo, true)` is real (`RenameTransformHandler.java:69-80`) and blocks via `Future.get(3, MINUTES)`. The whole block is also conditional — it only runs `if(!dinfo.getDependencyMap().isEmpty())`, so a plain save with no rename skips the cascade entirely (confirmed by the new `plainSaveInPlaceDoesNotCascadeWhenNoColumnWasRenamed` test asserting `verifyNoInteractions(renameTransformHandler)`). So this is "block a save that found a rename with a real dependent," not "block every save" — a reasonable trade-off given this tool's callers immediately chain `get_binding`/`list_bindable_fields`, consistent with the PVA-008/PVA-009 precedent cited in the PR description.
- Tests: the three new `WorksheetAgentControllerTest` cases exercise the real diff logic (a genuine `ColumnRef` oldName/alias mismatch drives the rename detection, not a mocked outcome) and only mock at the persisted-dependency-graph boundary (`DependencyTool.getDependencies` — confirmed to be exactly what `DependencyTransformer.getDependencies` delegates to, so the mock target is correct, not a wrong-method mock). Positive (rename + dependent → cascades and reports it) and negative (no rename → no cascade; Save-As → no cascade) cases are all covered — this meaningfully exercises the new path rather than just checking the mock was invoked.
- `WorksheetScriptControllerTest`'s update is a mechanical constructor-arg addition (`mock(RenameTransformHandler.class)`), no behavior change.
- Backward compatibility: `save()` returns `Map<String, Object>` (an open map that already carried an optional `warning` key before this PR), so adding `dependenciesUpdated`/`message` is additive and safe. Confirmed the plugin-side consumer already exists and reads these as optional fields (`stylebi-wiz` PR #1991, `worksheetTools.ts`), so this isn't shipping a backend field with no consumer.

**Non-blocking note (follow-up, not a blocker here):** `RenameTransformHandler.addTransformTask(..., true)` swallows a timeout/exception internally (catch-and-log only, `RenameTransformHandler.java:69-80`) rather than propagating it, so `result.put("dependenciesUpdated", true)` really means "a dependent was found and a transform task was submitted and waited on," not "the transform confirmed success." A cascade that times out or throws inside the 3-minute wait would still report `dependenciesUpdated: true`. This is pre-existing behavior of the shared handler (not introduced here, and already strictly better than the UI's fire-and-forget call), so not a blocker for this PR — flagging in case PVA-008/009-style flakiness resurfaces at this exact spot.

**CI: NOT green.**
- `build` — 1 failure: `ConditionTest$RegressionTests.toSqlConditionSilentlyDefaultsOnAnUnmatchedName` (`core/src/test/java/inetsoft/uql/ConditionTest.java:1475`). Expected/actual both print `2027-01-01` but differ at the sub-day level — this test is unrelated to this PR's diff (worksheet rename cascade) and looks like a pre-existing date/time-sensitive flake, not something this change caused.
- `claude-review` — failed at the action-infrastructure level (`is_error:true` after 349ms/1 turn, before producing any review output) — not a substantive review verdict on this diff.

**Verdict: holding, not merging.** No code-level blocker found in the diff itself, but both required checks are currently failing. Recommend re-running `build` and `claude-review` (they read as pre-existing/infra flakes unrelated to this diff) and re-checking once green.

Also noting: I'm the PR author, so GitHub blocks me from submitting an `--approve`/`--request-changes` review on my own PR (confirmed via the API: "Review Can not request changes on your own pull request") — a human still needs to give the actual approval before this merges. This comment is informational, not a formal approval/rejection.

PR #4907, opened 2026-08-31
fix(wiz): add force flag to open_base_worksheet for orphaned session recovery (PSM-006)

fredaan:
## Summary

- `open_base_worksheet` refuses with `"A worksheet session is already open (runtimeId=...)"` whenever the server's `SheetSessionService` still holds a worksheet session for this identity — even when the plugin's own local state (`status`/`detach_sheet`) has no record of it at all, e.g. a session orphaned by an earlier `connect_sheet(force:true)` reconnect on the client that didn't release it server-side. Today there is no recovery path short of a different owner identity or waiting out the session TTL.
- Adds an opt-in `force` flag to `SheetOpenService.openBaseWorksheet` (new 3-arg overload; the existing 2-arg overload still defaults `force` to `false`, so every existing caller is unaffected). When a held session is found and `force` is `true`, it is closed via `sheetSessions.close(held.sessionToken())` and the open proceeds instead of refusing. Omitting `force` (or passing `false`) preserves the exact existing refusal — the underlying safety check (a held session may belong to a different, concurrently-running agent) is unchanged.
- `ViewsheetAssemblyAgentController`'s `open-base-worksheet` endpoint now accepts an optional request body (`OpenBaseWorksheetRequest(Boolean force)`), defaulting to `false` when the body is absent or `null`, for compatibility with a plugin build predating this field (the existing empty-object POST body still works).

Companion client-side PR (stylebi-wiz): adds the `force` argument to the `open_base_worksheet` MCP tool and passes it through as this endpoint's request body.

Reference: PSM-006 in stylebi-wiz's `docs/bug-reports/plugin-composer-session-management.md`; diagnosis at `docs/teams/2026-08-31-test-composer-plugin/case-PSM-006-session-state-mismatch/01-diagnosis.md` in that repo.

## Test plan

- [x] `./mvnw -pl core test -Dtest="SheetOpenServiceTest,SheetSessionServiceTest,ViewsheetAssemblyAgentControllerTest"` — 90/90 pass, including:
  - a new `SheetSessionServiceTest` regression pinning the "multiple pane-scoped sessions of the same (ownerIdentity, sheetType) may legitimately coexist" invariant this fix must not break.
  - new `SheetOpenServiceTest` cases: the 3-arg overload with `force:false` refuses exactly like the 2-arg overload (and never calls `close`); `force:true` closes the held session and opens the new one; `force:true` with nothing held is a no-op on `close`; and one test driven against a **real** `SheetSessionService` (not a mock) proving the held session is actually closed, not merely ignored.
  - updated/new `ViewsheetAssemblyAgentControllerTest` cases for the request-body wiring (absent body defaults to `false`, `force:true` passes through).
- [x] `./mvnw -pl core test -Dtest='inetsoft.web.wiz.**'` — 2759/2759 pass, no regressions in the wider wiz package.
- [ ] Live end-to-end re-verification against a running StyleBI is intentionally deferred to the lead session that holds a live connection, per this task's instructions.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java
@@ -72,6 +72,24 @@ public class SheetOpenService {
     *                                  problem and, where there is one, the next tool to call.
     */
    public JoinSession openBaseWorksheet(String sessionToken, Principal user) throws Exception {
+      return openBaseWorksheet(sessionToken, user, false);
+   }
+
+   /**
+    * Open the base worksheet of the viewsheet paired to {@code sessionToken}, for {@code user}.
+    *
+    * @param force when {@code true} and a worksheet session is already held for this identity,
+    *              close it and proceed instead of refusing -- the recovery path for a session
+    *              orphaned by a client that lost the local pointer needed to name it for
+    *              {@code detach_sheet} (e.g. a prior {@code connect_sheet(force:true)} that
+    *              predates that client releasing what it replaces server-side).
+    *
+    * @throws IllegalArgumentException on every refusal, with a message naming the specific
+    *                                  problem and, where there is one, the next tool to call.
+    */
+   public JoinSession openBaseWorksheet(String sessionToken, Principal user, boolean force)
+      throws Exception
+   {
       JoinSession vsSession =
          viewsheetSessions.requireSessionAllowingPaneScope(sessionToken, user);
 
@@ -121,9 +139,18 @@ public class SheetOpenService {
       JoinSession held = sheetSessions.findOpen(vsSession.ownerIdentity(), SheetType.WORKSHEET);
 
       if(held != null) {
-         throw new IllegalArgumentException(
-            "A worksheet session is already open (runtimeId=" + held.runtimeId() + "). Call " +
-            "detach_sheet to close it before opening another base worksheet.");
+         if(!force) {
+            throw new IllegalArgumentException(
+               "A worksheet session is already open (runtimeId=" + held.runtimeId() + "). Call " +
+               "detach_sheet to close it, or -- if detach_sheet reports nothing connected because " +
+               "this held session is not in this client's local state -- call open_base_worksheet " +
+               "again with force:true to close it and open this one instead.");
+         }
+
+         // Best-effort: the caller is discarding this session either way, so a session the server
+         // has already forgotten (or that close() otherwise cannot fully act on) must not block
+         // the open that follows.
+         sheetSessions.close(held.sessionToken());
       }
 
       if(vsSession.socketSessionId() == null) {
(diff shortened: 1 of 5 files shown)


fredaan (commented):
Reviewed against PSM-006's diagnosis (docs/teams/2026-08-31-test-composer-plugin/case-PSM-006-session-state-mismatch/01-diagnosis.md in stylebi-wiz) and its own fix-2 spec. Verdict: no blocking issues found.

Correctness:
- `SheetOpenService.openBaseWorksheet(sessionToken, user, force)`: `force:false` preserves the exact existing refusal (message updated to also point at `force:true` as the recovery path); `force:true` closes exactly the one session `findOpen` returned via `sheetSessions.close(held.sessionToken())` before falling through to the normal open — never an inferred "whatever this owner holds," so it can't reach into an unrelated pane-scoped session of the same (ownerIdentity, sheetType), matching the diagnosis's "rejected alternative" analysis of why a blanket evict-on-join would be wrong.
- The 2-arg overload defaults to `force:false` and delegates to the 3-arg one — backward compatible, and the only real caller (`ViewsheetAssemblyAgentController`) is updated to the 3-arg form.
- `ViewsheetAssemblyAgentController.openBaseWorksheet` takes `@RequestBody(required = false) OpenBaseWorksheetRequest`, defaulting `force` to `false` on a null body/null field — correctly handles a plugin build that predates this field (empty-object or body-less POST).

Test quality — independently re-ran, not just read:
- `./mvnw -pl core test -Dtest="SheetOpenServiceTest,SheetSessionServiceTest,ViewsheetAssemblyAgentControllerTest"` → 90/90 pass locally, matching the PR description exactly.
- Confirmed `forceActuallyClosesTheHeldSessionAgainstARealSheetSessionService` genuinely instantiates `new SheetSessionService()` (line 294) rather than mocking it, and asserts the held token no longer resolves afterward (`realSessions.resolve(held.sessionToken(), OWNER)` is null) — this is a real "the close is real" proof, not a mock-accepts-anything test, as claimed.
- `openTwiceForTheSameOwnerAndTypeLeavesBothSessionsLiveAndIndependentlyReachable` correctly pins the "multiple pane-scoped sessions may coexist" invariant this fix must not break — good defensive coverage against the rejected blanket-evict alternative being reintroduced later.
- `force:false`-refuses and `force:true`-with-nothing-held (`close` never called) cases both present.

Checks: `claude-review` fails on both PRs in this pair with an empty `ANTHROPIC_API_KEY` (confirmed via `gh run view --log-failed`) — this session's established precedent treats that as pre-existing non-blocking infra, not a code issue. `build`, `Semgrep`/`Semgrep OSS`, and `license/cla` all pass. `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE` only due to the claude-review check.

I'm the PR author, so `--approve` isn't available to me on this account — filing this as an explicit comment-verdict instead. No changes requested; proceeding to merge.

PR #4909, opened 2026-09-01
Enforce the tabular catalog SPI contract and implement it for SharePoint Online

Nelson-cao:
## Problem

`TabularCatalogProvider` documents invariants on its record types — never-null collections, non-blank and unique identifiers, `TabularDatasetRef.id` free of `.`, column types drawn from `XSchema` — and `TabularCatalogService` enforced none of them. A connector violating one either crashed core with an NPE deep in response assembly (`relationships() == null` hits a for-each), or produced a silently wrong annotation target: a dotted dataset id is indistinguishable downstream from a database-qualified name, and a duplicated id makes `describeDataset` ambiguous.

The SPI also had only one implementer (OData), so its boundary had never been tested against a connector whose identifiers are shaped differently.

## Fix

**Contract enforcement (C1–C6).** `TabularCatalogService` now validates every documented invariant before mapping to the wire type, each failure raising a named exception that identifies the offending connector, dataset, and field: null collections; blank, duplicate, or dot-containing dataset ids; blank or duplicate relationship names; relationship endpoints not present in the catalog; `fromColumns`/`toColumns` non-empty and equal in length; blank column names; `keyColumns` membership in `columns`; a `describeDataset` result echoing the requested target; and `TabularColumn.type` being an `XSchema` type constant.

The type check whitelists all 21 constants `XSchema` declares rather than the seven its javadoc lists as examples — the javadoc's claim is "an `XSchema` type constant", and a narrower set would reject a legitimate connector mapping to `INTEGER` or `DECIMAL`. `XSchema.isPrimitiveType` is deliberately not reused: it omits `NULL`/`COLOR`/`UNKNOWN` and includes the internal-only `"bigdecimal"`, so it answers a different question.

**SharePoint Online implements the SPI.** `SharepointOnlineRuntime` gains `listDatasets`/`describeDataset` over sites and lists. Because a dataset must be addressable by a single string while a SharePoint list is identified by a (site, list) pair, `SharepointDatasetId` composes the two with three-character percent escaping: Graph's site id is `{hostname},{siteCollectionId},{webId}` and `getSiteId()` truncates it at the first comma, keeping the hostname — so the site component always contains dots, and choosing a non-dot separator is not sufficient. The id is validated by round-trip: any id where `compose(parse(id))` does not reproduce the input is rejected.

The new path does not reuse `doGetSites`, whose two `catch(Exception) { LOG.warn }` blocks swallow permission failures indiscriminately — annotation must surface those rather than silently return a short list. `listSitesOrThrow` propagates instead, and enumerates sub-sites recursively (`doGetSites` stops at one level, so grandchild sites were invisible). No existing catch block is modified; the connector's dropdown behaviour is unchanged, with a regression test asserting it.

## Testing

74 tests across 12 classes; full `core` 8236 pass. `inetsoft-sharepoint-online` gains its first test coverage (21 tests). Graph is exercised through `mockStatic(GraphServiceClient.class)`, leaving `getClient()` unmodified — the same layer at which the OData tests substitute WireMock. `*CollectionPage` results are built with the SDK's public `(List<T>, RequestBuilder)` constructor rather than mocked.

Live Microsoft 365 behaviour is not covered: no tenant is available. `SharepointOnlineCatalog`'s javadoc records the one load-bearing assumption this leaves unverified.

### DIFF
--- a/core/src/main/java/inetsoft/uql/tabular/TabularUtil.java
+++ b/core/src/main/java/inetsoft/uql/tabular/TabularUtil.java
@@ -1119,26 +1119,53 @@ public class TabularUtil {
     *         connector plugin is not loaded.
     */
    public static TabularRuntime createRuntime(String dataSource) {
+      XDataSource ds;
+
       try {
-         XDataSource ds = XRepository.getRepository().getDataSource(dataSource);
-
-         if(ds == null) {
-            return null;
-         }
-
-         String runtimeClass = Config.getConfig().getRuntime(ds.getType());
-
-         if(runtimeClass == null) {
-            return null;
-         }
-
-         return (TabularRuntime) Config.getConfig().getClass(ds.getType(), runtimeClass)
-            .getDeclaredConstructor().newInstance();
+         ds = XRepository.getRepository().getDataSource(dataSource);
       }
       catch(Exception e) {
-         LOG.error("Failed to create a tabular runtime for the given data source: {}",
+         LOG.error("Failed to look up the data source to create a tabular runtime for: {}",
                    dataSource, e);
          return null;
+      }
+
+      if(ds == null) {
+         return null;
+      }
+
+      String runtimeClass = Config.getConfig().getRuntime(ds.getType());
+
+      if(runtimeClass == null) {
+         return null;
+      }
+
+      Class<?> clazz;
+
+      try {
+         clazz = Config.getConfig().getClass(ds.getType(), runtimeClass);
+      }
+      catch(Exception e) {
+         // Plugin/class not loadable — treated as "no runtime available", same as before: this is
+         // still "not implemented" from the caller's point of view, not "broken".
+         LOG.error("Failed to load tabular runtime class {} for data source {}",
+                   runtimeClass, dataSource, e);
+         return null;
+      }
+
+      try {
+         return (TabularRuntime) clazz.getDeclaredConstructor().newInstance();
+      }
+      catch(Exception e) {
+         // The class WAS found and loaded; constructing it failed. A materially different, and
+         // much more actionable, condition than "no runtime" — must not collapse into the same
+         // null that TabularCatalogService.resolveProvider turns into
+         // UnsupportedDatasourceException. TabularCatalogService.java:51 is this method's only
+         // production caller, so this is a safe, contained behavior change.
+         LOG.error("Failed to construct tabular runtime {} for data source {}",
+                   runtimeClass, dataSource, e);
+         throw new RuntimeException("Failed to construct the tabular runtime for data source '" +
+            dataSource + "' (" + runtimeClass + "): " + e.getMessage(), e);
       }
    }
 
--- a/core/src/main/java/inetsoft/uql/tabular/TabularUtil.java
+++ b/core/src/main/java/inetsoft/uql/tabular/TabularUtil.java
@@ -1134,7 +1134,18 @@ public class TabularUtil {
          return null;
       }
 
-      String runtimeClass = Config.getConfig().getRuntime(ds.getType());
+      String runtimeClass;
+
+      try {
+         runtimeClass = Config.getConfig().getRuntime(ds.getType());
+      }
+      catch(Exception e) {
+         // Same "not implemented" treatment as the class-loading catch below — this call is close
+         // to a plain map lookup and unlikely to throw, but the method's contract (never throw,
+         // only ever return null) predates this round's split and must not narrow silently.
+         LOG.error("Failed to look up the tabular runtime class for data source {}", dataSource, e);
+         return null;
+      }
 
       if(runtimeClass == null) {
          return null;
@@ -1164,8 +1175,12 @@ public class TabularUtil {
          // production caller, so this is a safe, contained behavior change.
          LOG.error("Failed to construct tabular runtime {} for data source {}",
                    runtimeClass, dataSource, e);
+         // The full RuntimeException (with the FQCN) is logged above for diagnosis; the message
+         // that reaches the caller — and, via DatasourceMetaApiController.handleException, the
+         // HTTP error body — names only the data source, matching the controlled style
+         // UnsupportedDatasourceException already uses in this area, not an internal class name.
          throw new RuntimeException("Failed to construct the tabular runtime for data source '" +
-            dataSource + "' (" + runtimeClass + "): " + e.getMessage(), e);
+            dataSource + "': " + e.getMessage(), e);
       }
    }
 
--- a/core/src/main/java/inetsoft/uql/tabular/TabularUtil.java
+++ b/core/src/main/java/inetsoft/uql/tabular/TabularUtil.java
@@ -1179,8 +1179,12 @@ public class TabularUtil {
          // that reaches the caller — and, via DatasourceMetaApiController.handleException, the
          // HTTP error body — names only the data source, matching the controlled style
          // UnsupportedDatasourceException already uses in this area, not an internal class name.
+         // Deliberately NOT appending e.getMessage(): for a missing public no-arg constructor,
+         // getDeclaredConstructor() throws NoSuchMethodException whose own message IS the runtime
+         // class's FQCN, which would leak it right back in. The cause is already chained into this
+         // RuntimeException and already logged in full above, so nothing is lost for diagnosis.
          throw new RuntimeException("Failed to construct the tabular runtime for data source '" +
-            dataSource + "': " + e.getMessage(), e);
+            dataSource + "'", e);
       }
    }
 
--- a/core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java
@@ -171,10 +171,24 @@ public class TabularCatalogService {
          throw new Exception("Data source '" + dsName + "' target '" + target +
             "' returned no columns — cannot annotate.");
       }
+      validateDatasetIdEchoed(dsName, target, schema);
       validateColumnNames(dsName, target, schema.columns());
       validateKeyColumns(dsName, target, schema);
 
       return toDataset(dsName, xrepository.getDataSource(dsName).getType(), schema, objectMapper);
+   }
+
+   private static void validateDatasetIdEchoed(String dsName, String target,
+                                               TabularDatasetSchema schema) throws Exception
+   {
+      if(!target.equals(schema.datasetId())) {
+         // TabularDatasetSchema.datasetId's javadoc: "echoes the id that was asked for, so a
+         // result is self-identifying." toDataset() below uses schema.datasetId(), not target, to
+         // set OsiDataset.name/.source — a connector that answers with a different dataset's
+         // schema would silently corrupt which dataset the annotation ends up labeled as.
+         throw new Exception("Data source '" + dsName + "' was asked to describe target '" +
+            target + "' but returned a schema for '" + schema.datasetId() + "' instead.");
+      }
    }
 
    private static void validateColumnNames(String dsName, String target, List<TabularColumn> columns)
--- a/core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java
@@ -93,6 +93,13 @@ public class TabularCatalogService {
          if(ref == null || ref.id() == null || ref.id().isBlank()) {
             throw new Exception("Data source '" + dsName + "' returned a dataset with a blank id.");
          }
+         if(ref.id().contains(".")) {
+            // TabularDatasetRef.id's javadoc: "Must not contain a '.' character" — wiz's own
+            // bareTableName/sourceMatches split a non-FILE source on '.', so a dotted id would
+            // silently collide two different datasets, or resolve to the wrong one, downstream.
+            throw new Exception("Data source '" + dsName + "' returned a dataset id '" + ref.id() +
+               "' containing '.', which TabularDatasetRef.id's contract forbids.");
+         }
       }
    }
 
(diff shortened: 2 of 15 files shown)


core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java
```diff
@@ -89,10 +171,92 @@ public OsiDataset describeTable(String dsName, String target) throws Exception {
...
+
+   private static void validateColumnNames(String dsName, String target, List<TabularColumn> columns)
+      throws Exception
+   {
+      for(TabularColumn column : columns) {
+         if(column == null || column.name() == null || column.name().isBlank()) {
+            throw new Exception("Data source '" + dsName + "' target '" + target +
+               "' returned a column with a blank name.");
+         }
+         if(!XSCHEMA_TYPE_CONSTANTS.contains(column.type())) {
```
Nelson-cao: A connector's `TabularColumn.type()` can be `null` (nothing on the record prevents it), but `XSCHEMA_TYPE_CONSTANTS` is built with `Set.of(...)`. A `Set.of(...)`-backed set throws `NullPointerException` from `contains(null)` rather than returning `false` (verified independently: `Set.of(<21 elements>).contains(null)` throws NPE on JDK 21). So a connector that returns a column with a null type — an easy mistake for a future third implementer mapping an unrecognized native type — will make this validation throw a raw NPE instead of the named, connector/column-identifying exception this whole PR exists to produce. Neither current implementer can trigger it today (OData's `toXSchemaType` always falls back to `XSchema.STRING`; SharePoint's `doGetListColumns` only emits typed nodes with non-null `getType()`), and no added test covers `type == null`, so this gap is untested. Suggest guarding explicitly, e.g. `column.type() == null || !XSCHEMA_TYPE_CONSTANTS.contains(column.type())`.

core/src/main/java/inetsoft/uql/tabular/TabularUtil.java
```diff
@@ -1119,27 +1119,69 @@ public static TabularQuery createQuery(String dataSource) {
...
-                   dataSource, e);
+         // Plugin/class not loadable — treated as "no runtime available", same as before: this is
+         // still "not implemented" from the caller's point of view, not "broken".
+         LOG.error("Failed to load tabular runtime class {} for data source {}",
+                   runtimeClass, dataSource, e);
          return null;
       }
+
+      try {
+         return (TabularRuntime) clazz.getDeclaredConstructor().newInstance();
```
Nelson-cao, line 1168: This catch block wraps both `clazz.getDeclaredConstructor()` and `.newInstance()`. If the runtime class is loaded but has no public no-arg constructor, `getDeclaredConstructor()` throws `NoSuchMethodException`, whose `getMessage()` is `<FQCN>.<init>()` — i.e. it DOES contain the runtime's fully-qualified class name. That gets concatenated straight into this method's `RuntimeException` message, which (per this PR's own review-r1 fix and the comment two lines below) is meant to reach the HTTP error body naming only the data source, not an internal class name — so this specific failure mode re-leaks exactly what review r1 closed for the 'constructor throws' case. Verified independently with a standalone reflection test: calling `getDeclaredConstructor()` on a class with no matching constructor produces a `NoSuchMethodException` whose message is the FQCN + `<init>()`. The added `TabularUtilCreateRuntimeTest` only exercises a constructor that throws in its body (where `InvocationTargetException.getMessage()` happens to be `null`, so the FQCN-leak assertion passes trivially) — it never tests a class with a missing/mismatched constructor, so this path is untested.\n\nSeparately, and lower priority: because `InvocationTargetException.getMessage()` is always `null` (JDK behavior), the common 'constructor threw' case produces a message like "...: null" instead of surfacing the actual cause text (e.g. "boom") — the real reason is only visible via `e.getCause().getMessage()`, preserved in the logged exception but not in the top-level message. Consider `e.getCause() != null ? e.getCause().getMessage() : e.getMessage()`.

Nelson-cao:
## Review verdict: Approve with two important findings (non-blocking, both narrow-scope)

I reviewed all 15 files (5 main + 3 test in `core`, 3 main + 4 test in `connectors/inetsoft-sharepoint-online`), read the full current source for every touched class (not just the diff), and ran the new/added test classes directly:

- `core`: `TabularCatalogServiceContractValidationTest` (18 tests) + `TabularUtilCreateRuntimeTest` (3 tests) — all pass.
- `connectors/inetsoft-sharepoint-online`: `SharepointDatasetIdTest` (10), `SharepointOnlineCatalogTest` (3), `SharepointOnlineCatalogPermissionFailureTest` (4), `SharepointOnlineDropdownUnaffectedTest` (4) — all 21 pass.

I did not run the full `core` suite (8236 tests, per the PR description) — out of scope per the review instructions — but the targeted runs above confirm the new coverage is real and green, not tautological.

### Verified as correct
- **`SharepointDatasetId`'s percent-escaping scheme** (escape `%` first / unescape `%` last): traced the escape/unescape order by hand against several adversarial cases (a component containing a literal `%2E` substring, a component containing the literal separator `~`, two raw separators in one string) — the round-trip check correctly rejects every string that `compose()` could not have produced, and correctly decodes every one it could. No bug found here.
- **`XSCHEMA_TYPE_CONSTANTS`**: confirmed against `XSchema.java` that it lists exactly the 21 `public static final String` constants the class declares — no more, no fewer. Confirmed `XSchema.isPrimitiveType`'s claimed mismatch (excludes `NULL`/`COLOR`/`UNKNOWN`, includes the non-constant `"bigdecimal"` alias) directly against `typeclasses`'s static initializer — the javadoc's reasoning for not reusing it holds.
- **The "not a privilege escalation" argument** in `SharepointOnlineCatalog`'s javadoc: confirmed `SharepointOnlineQuery.setSite`/`setList` are plain unvalidated setters, and confirmed `MetadataApiService` gates both `listTables` and `describeTable` behind a `ResourceAction.READ` check before either is reached. The argument holds.
- **`TabularUtil.createRuntime`'s "sole production caller" claim**: grepped the whole repo — `TabularCatalogService`'s `TabularUtil::createRuntime` reference is indeed the only call site, so the new unchecked-throw behavior on constructor failure cannot surprise any other caller.
- Neither current SPI implementer can hit finding #1 below today: OData's `toXSchemaType` always falls back to `XSchema.STRING` (never null), and SharePoint's `doGetListColumns` only emits typed nodes with non-null `getType()`.

### Findings

**1. (Important) `TabularColumn.type() == null` NPEs instead of producing the named exception this PR exists to guarantee** — `core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java:202`. `XSCHEMA_TYPE_CONSTANTS` is a `Set.of(...)`, and `Set.of(...).contains(null)` throws `NullPointerException` (verified with a standalone JDK 21 snippet), rather than returning `false`. A future connector that returns a column with a null type — an easy mistake when mapping an unrecognized native type — gets a raw NPE instead of a message naming the offending data source/column, which is exactly the failure class C1-C6 exist to eliminate. Untested by the new suite. See inline comment for the one-line fix.

**2. (Important) The construction-failure catch in `TabularUtil.createRuntime` can still leak the runtime's FQCN into the HTTP error body** — `core/src/main/java/inetsoft/uql/tabular/TabularUtil.java:1168`. The catch wraps both `getDeclaredConstructor()` and `.newInstance()`. If the runtime class exists but has no public no-arg constructor, `getDeclaredConstructor()` throws `NoSuchMethodException` whose message IS `<FQCN>.<init>()` — re-opening exactly the leak review r1 closed for the "constructor throws" case. Verified independently via a standalone reflection test. The added `TabularUtilCreateRuntimeTest` only covers a constructor that throws in its body (where `InvocationTargetException.getMessage()` happens to be `null`, so the "no FQCN" assertion passes trivially either way) — the missing-constructor case is untested. See inline comment; also noted a lower-priority related nit (the "constructor throws" case's caller-facing message is currently always "...: null" since `InvocationTargetException.getMessage()` has no message of its own — the real reason lives only in `e.getCause()`).

**3. (Nit) `SharepointOnlineCatalog.describeDataset` inherits `doGetListColumns`'s silent-drop behavior for unmapped Graph column shapes.** `doGetListColumns`'s `if/else if` chain (untouched by this PR) produces no node — and therefore silently omits the column — for any shape outside text/personOrGroup/choice/number/boolean/dateTime/currency (e.g. Lookup, Hyperlink, Location). A list composed entirely of such columns would make `describeDataset` return an empty column list (which `TabularCatalogService` then rejects), and a list with a mix would silently under-report its schema with no error at all. This is pre-existing, untouched code being pointed at a new consumer with a completeness expectation the original method was never designed to guarantee. The PR is otherwise very disciplined about stating every deliberate degradation (no relationships, no column alias) — this one isn't listed alongside them. Not asking for a fix, just flagging that it reads like an accidental gap rather than a stated one.

**4. (Nit) The "no dot in a dataset id" invariant (C6) is enforced on the `listDatasets` output (`validateDatasetIds`) but not on the `describeTable` input/echo.** `validateDatasetIdEchoed` only checks `target.equals(schema.datasetId())`, not that the (now-trusted) echoed value is dot-free, before `toDataset()` writes it into `OsiDataset.source`. In practice this isn't exploitable today: SharePoint's `SharepointDatasetId.parse` round-trip check transitively rejects any input containing a raw dot (verified by tracing the escape/compose logic), and OData's `describeDataset` is a map lookup keyed by real metadata entity-set names, so a bogus dotted target simply isn't found. But that closure is incidental to each connector's own implementation rather than a guarantee the shared validation layer provides the way C6 does for `listTables` — a future third implementer with a more permissive `describeDataset` lookup wouldn't be caught by anything at the `TabularCatalogService` layer. Worth a one-line note in `TabularDatasetSchema.datasetId`'s javadoc or an explicit check, given how much care this round put into shared-layer enforcement everywhere else.

### Not findings - challenged but held up
- The `keyColumns() == null` tolerance (vs. `relationships() == null` rejection) — the asymmetry is explained and correct: the null case is fully absorbed by `toDataset`'s existing ternary with no observable difference, unlike `relationships()` which NPEs in a for-each.
- `SharepointOnlineRuntime`'s pre-existing dropdown methods (`doGetSites`/`getChildSites`) — confirmed genuinely untouched by diff, and `SharepointOnlineDropdownUnaffectedTest` pins their current (one-level, degrade-and-warn) behavior as a real regression tripwire, not a rubber-stamp.
- `listSitesOrThrow`'s de-dup/recursion logic (`bySiteId.putIfAbsent` gating whether `collectDescendants` recurses) — traced by hand against the group-site-already-visited-as-root-descendant case; correct.

None of the above are blockers — #1 and #2 are both edge cases requiring either a still-nonexistent third connector or a broken/misconfigured one, not reachable via SharePoint or OData as shipped. I'd merge this as-is if the team prefers to track #1/#2 as fast-follow, but they're one-line fixes if you'd rather fold them in now.

core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java
Nelson-cao: Fixed in c5070caa1: `column.type() == null || !XSCHEMA_TYPE_CONSTANTS.contains(column.type())`. Verified: the added test `describeTable_nullColumnType_throwsNamedExceptionNotNpe` asserts on the caught exception (not just `assertThrows(Exception.class, ...)`, which would have passed on an escaped NPE too) — confirmed it actually fails without the fix and passes with it. Resolving.

core/src/main/java/inetsoft/uql/tabular/TabularUtil.java
Nelson-cao: Fixed in c5070caa1: the appended `e.getMessage()` is removed entirely (not just swapped for `e.getCause().getMessage()`, which would have re-leaked the FQCN in the case where `e` itself is the `NoSuchMethodException` with no cause). Verified end-to-end: `DatasourceMetaApiController.handleException` uses `e.getMessage()` for the HTTP body, and that message is now a fixed, non-null literal naming only the data source — no fallback to `e.getClass().getName()` is even reached. Added test `createRuntime_noPublicNoArgConstructor_messageDoesNotLeakFqcn` exercises exactly the missing-no-arg-constructor path this thread was about (confirmed it fails without the fix). Resolving.

Nelson-cao:
## Follow-up: both important findings confirmed resolved in c5070caa1

Re-checked `git diff a42b8f0e2 c5070caa1` (4 files, 3 production lines) and re-ran the affected tests.

**Finding 1 (null column type NPE) — closed.** `TabularCatalogService.java:202` now short-circuits with `column.type() == null || !XSCHEMA_TYPE_CONSTANTS.contains(column.type())`, so `Set.of(...).contains(null)` is never reached. Added test `describeTable_nullColumnType_throwsNamedExceptionNotNpe` correctly asserts `assertFalse(ex instanceof NullPointerException)` — not just `assertThrows(Exception.class, ...)`, which would have passed even on the old, broken code (NPE is an Exception).

**Finding 2 (FQCN leak) — closed, and closed more completely than my suggested fix.** The appended `": " + e.getMessage()` is removed entirely rather than swapped for a fallback expression. I'd considered suggesting `e.getCause() != null ? e.getCause().getMessage() : e.getMessage()`, but that would have re-leaked the FQCN in exactly the case this thread is about: `NoSuchMethodException` (missing no-arg constructor) has no cause, so a cause-based fallback would fall through to `e.getMessage()` — the FQCN — again. Full removal closes it for every subtype the catch can see, not just the two identified. Traced the full HTTP path to confirm nothing reopens the leak downstream: `DatasourceMetaApiController` has its own local `@ExceptionHandler(Exception.class)` (which wins over `WizControllerErrorHandler`'s `ControllerAdvice`), and it emits `e.getMessage() != null ? e.getMessage() : e.getClass().getName()` — since the new message is a fixed non-null literal, the `getClass().getName()` fallback is never even reached. Added test `createRuntime_noPublicNoArgConstructor_messageDoesNotLeakFqcn` exercises the exact missing-no-arg-constructor path the original comment only claimed to cover.

**No new issues from the fix.** Asked specifically whether dropping `e.getMessage()` costs anything: no — the full exception (with its real message, including the FQCN where applicable) is preserved as the `RuntimeException`'s cause and is already fully logged via `LOG.error(...)` one line above the throw, so nothing is lost for server-side diagnosis; only the HTTP-facing text changes, which is the intended effect.

**Test run:** `TabularCatalogServiceContractValidationTest` (19) + `TabularUtilCreateRuntimeTest` (4) = 23/23 pass at `c5070caa1`.

**Nits (#3 doGetListColumns shape-drop, #4 describeTable echoed-target dot check) are unaffected** — this commit doesn't touch `SharepointOnlineCatalog`/`SharepointOnlineRuntime` or `validateDatasetIdEchoed`. Understood these are being written into the integration doc's known-degradations list instead.

Verdict: both important findings resolved correctly, no regressions introduced. No objection to merging.

fredaan (approved):

PR #4910, opened 2026-09-01
fix(wiz): propagate principal/locale across RenderWaitSupport's virtual thread (PSM-004)

fredaan:
## Summary

- `RenderWaitSupport.awaitOrRetry` (used by the whole-viewsheet `get_viewsheet_image()` path, among others) runs `work` on a brand-new virtual thread that never receives the caller's authenticated `Principal` or `Locale`. On that context-less thread, `GraphTypeUtil.checkChartStylePermission` reads `ThreadContext.getContextPrincipal()`, gets `null`, and `SecurityEngine.checkPermission(null, ...)` unconditionally denies every chart style.
- `GraphTypeUtil.getAvailableAutoChartType` then returns `-1` for an `Auto`-typed chart, and that `-1` gets written into a plain, shared (non-thread-local) field on the live `ChartInfo`/`ChartAggregateRef` object via `updateChartType`/`updateFieldChartTypes` — permanently poisoning the runtime's shared chart-type resolution. Every later render (scoped or whole-sheet, any thread, any identity) reads the cached `-1` and throws, until a human clicks the Composer UI's Refresh button.
- Fix: capture `ThreadContext.getContextPrincipal()` and `ThreadContext.getLocale()` on the calling thread before submitting to the executor, and re-install them inside the wrapped `Callable` on the new virtual thread, clearing them in a `finally` block. This benefits every current caller of `awaitOrRetry` automatically, with no change to `GraphTypeUtil`'s permission-check design or to any call site.

Full root-cause trace: `docs/teams/2026-08-31-test-composer-plugin/case-PSM-004-round2-chart-type-permission/01-diagnosis.md` in the stylebi-wiz repo. Bug: PSM-004 (reopened), `docs/bug-reports/plugin-composer-session-management.md`.

## Test plan

- [x] `RenderWaitSupportTest.principalSetOnTheCallingThreadIsNotVisibleInsideTheWrappedCallable` (previously documented the gap as expected/passing behavior) updated to `principalSetOnTheCallingThreadIsVisibleInsideTheWrappedCallable`, now asserting the principal IS visible inside the callable after the fix.
- [x] Added `localeSetOnTheCallingThreadIsVisibleInsideTheWrappedCallable` — same propagation for `Locale`.
- [x] Added `exceptionFromWorkStillPropagatesThroughTheContextPropagatingWrapper` — guards that wrapping `work` in a try/finally doesn't interfere with `awaitOrRetry`'s existing exception-unwrapping.
- [x] `./mvnw -pl core -Dtest=RenderWaitSupportTest test` — 5/5 pass.
- [x] `./mvnw -pl core -Dtest=ScriptImageServiceTest,WorksheetEditServiceTest,VSExportServiceRefreshTest test` — 39/39 pass (no regressions in the other callers/adjacent tests of this utility).
- [ ] Live re-test of the actual PSM-004 repro (whole-sheet `get_viewsheet_image()` no longer wedges the runtime) — intentionally not done in this PR; will be verified separately against a live StyleBI instance per the lead session's live connection.

## Other call sites

`ViewsheetEditService.java:193`, `WorksheetAgentController.java:2211/2382/2482/2583`, and `WorksheetEditService.java:280` all call the same shared `RenderWaitSupport.awaitOrRetry` and automatically inherit this fix — no changes needed at those call sites. Whether their underlying calls (`getVSTableLens`, `refreshColumnSelection`, `loadTableData`) actually perform a `ThreadContext`-dependent permission/entitlement check was not traced in this PR (flagged in the diagnosis as a follow-up, not required for this fix).


fredaan (commented):
Code review clean (independent re-run of RenderWaitSupportTest 5/5, ScriptImageServiceTest+WorksheetEditServiceTest+VSExportServiceRefreshTest 39/39, both matching the PR's own claims). build now passes; only the known non-blocking claude-review infra flake remains. Merging — live re-verification against the original PSM-004 repro (whole-viewsheet get_viewsheet_image on an Auto-chart-type viewsheet) will be done separately by the session holding a live StyleBI connection.

PR #4912, opened 2026-09-01
Carry a connector's binding parameters through the tabular catalog SPI, and add Cassandra

Nelson-cao:
## Problem

A tabular source annotated through `TabularCatalogProvider` produced a table that nothing could query. `TabularCatalogService` wrote the dataset's columns, keys, and type into the annotated document, but nothing about how to reach the target again — so an agent holding the retrieval result had no way to fill the connector's query bean.

OData happened to work: its dataset id is the entity set name, which is also the value its single `entity` property takes, so matching by name succeeded by coincidence. SharePoint Online could not. Its target needs two properties (`site`, `list`) and its id is an escaped composite carrying GUIDs, so the identity was unrecoverable outside the connector. Annotation succeeded and the resulting table was unusable.

That gap belongs to the SPI, not to either connector: any target needing more than one property, or whose id is not itself a property value, lands in it.

## Fix

`TabularDatasetSchema` gains a `params` component — the parameters a connector's own query bean needs in order to read this dataset again, keyed by that bean's property names. `describeDataset` already resolves the id at that point, so reporting them costs no extra call. A three-argument constructor remains for a connector with nothing to report; empty and absent are the same thing to the caller, and `null` is not a third state.

`TabularCatalogService` maps them into the dataset's COMMON extension under `params`, the slot ServerFile already uses for the same purpose, and validates them alongside the existing contract checks: a blank key or a null value raises a named exception identifying the connector, dataset, and field. The key is omitted entirely when a connector reports nothing, rather than written as an empty object.

OData reports `{entity: <entity set>}`; SharePoint Online reports `{site: ..., list: ...}` carrying the unescaped values, not fragments of the composed id. Keys are the bean property names as derived from the getter, not the `@Property(label=...)` display text — a mismatch there produces a contract key that does not exist and fails later at the agent rather than at annotation.

The dataset id is unchanged, as are the no-dot invariant and SharePoint's id escaping. Neither connector's existing methods are touched.

## Testing

49 tests across the six affected classes. Each new test was checked against a reverted implementation to confirm it fails, and by assertion rather than by an incidental error.

Companion change in wiz: inetsoft-technology/stylebi-wiz#2040.

---

## Also in this PR: Cassandra as the third implementer

Added after the review above, on the same branch.

### Problem

The SPI had two implementers and both report *identifiers* — OData an entity set name, SharePoint a site and a list — which the connector turns into a request. Nothing had yet exercised a connector whose query bean takes a free-form statement, so it was unproven that the same "report parameters keyed by the bean's own property names" mechanism covers that shape. Cassandra is also the first target with a real composite key to report.

The connector itself had no catalog capability at all: it enumerated keyspaces for the connection dialog and discarded the `KeyspaceMetadata` it already held, and had no code reading tables or columns.

### Fix

`CassandraRuntime` implements `TabularCatalogProvider`, delegating to a new `CassandraCatalog` that reads schema metadata for the keyspace the data source is already bound to. A dataset is a table in that keyspace, so its id needs no qualifier. `keyColumns` reports partition key followed by clustering columns — the full primary key, since the partition key alone does not identify a row. The reported binding parameter is `queryString`, whose value is `SELECT * FROM <table>`; row capping stays with `runQuery`'s existing `TabularUtil.getMaxRows`, so no `LIMIT` is emitted.

Identifiers are quoted through `CqlIdentifier.asCql(true)`, which quotes only when required. Cassandra folds unquoted identifiers to lower case, so this is tested in both directions — a mixed-case table must come out quoted and a lower-case one must not. Testing only the first would pass against an implementation that always quotes.

An unconfigured keyspace throws before a session is opened, so it cannot be confused with a keyspace that legitimately has no tables. Session establishment is left uncaught deliberately, so a connection or permission failure surfaces as an error rather than an empty catalog.

`CassandraTable.getType(int)`'s `DataType -> Class` switch is extracted verbatim into a static function that the schema path reuses; the instance method now delegates. Duplicating the switch instead would let the type an annotation reports drift from the type the query returns. A parameterized test asserts, through `getType(int)`, that every protocol code the switch names still maps as before — it stands in for the "this file does not change" guarantee the extraction gives up.

### Testing

31 new tests, all shown to fail against a reverted implementation. No Cassandra cluster is available: the schema path is covered through driver interfaces, and the keyspace guard is proven to short-circuit by pointing a real runtime at an unroutable address. Permission and connection failures are correct by reading and not covered by a test — verifying them would mean dialing an unreachable host and waiting out the driver's timeout, buying a timing dependency in CI to confirm the absence of a catch block.

Three limitations are recorded in `docs/tabular/stylebi-tabular-wiz-integration.md`: eight column types coarsen to `string` through the shared `CoreTool.getDataType` chain, a table name containing `.` fails annotation at the service layer, and the emitted starting query is unrestricted.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java
@@ -268,6 +268,12 @@ public class TabularCatalogService {
       }
 
       for(Map.Entry<String, String> e : params.entrySet()) {
+         // Key and value are NOT symmetric here. A blank key is never legitimate -- a bean property
+         // name cannot be blank -- so isBlank() is right for it. A blank VALUE can be entirely
+         // legitimate: a connector may report an optional property whose empty value means "unset"
+         // (e.g. OData's own Select/Filter/Expand). This method cannot tell an identity param from
+         // an optional one -- it does not know the connector's property semantics -- so rejecting a
+         // blank value here would reject a correct connector, not just a broken one.
          if(e.getKey() == null || e.getKey().isBlank() || e.getValue() == null) {
             throw new Exception("Data source '" + dsName + "' target '" + target +
                "' returned a params entry with a blank key or null value.");
--- a/core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java
@@ -174,6 +174,7 @@ public class TabularCatalogService {
       validateDatasetIdEchoed(dsName, target, schema);
       validateColumnNames(dsName, target, schema.columns());
       validateKeyColumns(dsName, target, schema);
+      validateParams(dsName, target, schema.params());
 
       return toDataset(dsName, xrepository.getDataSource(dsName).getType(), schema, objectMapper);
    }
@@ -253,6 +254,23 @@ public class TabularCatalogService {
             throw new Exception("Data source '" + dsName + "' target '" + target +
                "' declared key column '" + keyColumn + "', which is not one of its own reported " +
                "columns.");
+         }
+      }
+   }
+
+   private static void validateParams(String dsName, String target, Map<String, String> params)
+      throws Exception
+   {
+      if(params == null) {
+         return;   // Same "defensive but not enforced" treatment as validateKeyColumns: the
+                    // compatibility constructor already guarantees non-null, this only guards
+                    // against a third party calling the 4-arg constructor directly with null.
+      }
+
+      for(Map.Entry<String, String> e : params.entrySet()) {
+         if(e.getKey() == null || e.getKey().isBlank() || e.getValue() == null) {
+            throw new Exception("Data source '" + dsName + "' target '" + target +
+               "' returned a params entry with a blank key or null value.");
          }
       }
    }
@@ -340,7 +358,7 @@ public class TabularCatalogService {
          null : schema.keyColumns());
       dataset.setFields(fields);
       dataset.setCustomExtensions(
-         List.of(buildDatasetExtension(dsName, datasourceSubtype, objectMapper)));
+         List.of(buildDatasetExtension(dsName, datasourceSubtype, schema.params(), objectMapper)));
       return dataset;
    }
 
@@ -364,8 +382,17 @@ public class TabularCatalogService {
     * {@code datasourceType}) that this is not a SQL table — see charter assertion B7. The literal
     * {@code "tabular"} must match wiz's {@code TABULAR_DATASOURCE_TYPE} exactly; there is no shared
     * constant between the two repositories.
+    *
+    * {@code params}, when non-empty, is the connector's own query-bean property names mapped to
+    * the string values that identify (METADATA targets) or help re-read (FILE targets — see wiz's
+    * {@code parseOptions}) this dataset — see {@link TabularDatasetSchema#params()}. An empty map
+    * (the compatibility constructor's default, or a connector that has not opted in) omits the
+    * {@code "params"} key entirely rather than writing it out empty or null, so wiz's reader can't
+    * tell "no params" apart from "not implemented yet" — which is exactly the point: both mean the
+    * same thing to a consumer that only wants to know whether it has something to bind with.
     */
    private static OsiCustomExtension buildDatasetExtension(String dsName, String datasourceSubtype,
+                                                            Map<String, String> params,
                                                             ObjectMapper objectMapper)
    {
       try {
@@ -374,6 +401,10 @@ public class TabularCatalogService {
          extData.put("path", dsName);
          extData.put("datasourceType", "tabular");
          extData.put("datasourceSubtype", datasourceSubtype);
+
+         if(params != null && !params.isEmpty()) {
+            extData.put("params", params);
+         }
 
          OsiCustomExtension ext = new OsiCustomExtension();
          ext.setVendorName("COMMON");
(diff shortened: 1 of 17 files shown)


core/src/main/java/inetsoft/web/wiz/service/TabularCatalogService.java
```diff
@@ -257,6 +258,23 @@ private static void validateKeyColumns(String dsName, String target, TabularData
...
+      throws Exception
+   {
+      if(params == null) {
+         return;   // Same "defensive but not enforced" treatment as validateKeyColumns: the
+                    // compatibility constructor already guarantees non-null, this only guards
+                    // against a third party calling the 4-arg constructor directly with null.
+      }
+
+      for(Map.Entry<String, String> e : params.entrySet()) {
+         if(e.getKey() == null || e.getKey().isBlank() || e.getValue() == null) {
```
Nelson-cao, line 277: Nit: this rejects a null value but not a blank (empty-string) one, which is inconsistent with validateColumnNames a few lines up (which rejects both null and isBlank() for a column name). A connector that reported `params: {"site": ""}` would sail through here and produce a dataset that still can't be re-queried -- the same failure mode this whole PR exists to close, just one step further down. Not a live bug today (neither OData nor SharePoint Online can produce a blank value here), but worth tightening for symmetry with the rest of this validation layer, or a one-line comment explaining why values get looser treatment than keys.

Nelson-cao:
## Review verdict: Approve, 0 blockers, 0 important findings, 1 nit

Reviewed all 8 files against the current head (`0557de6`) with the surrounding, non-diff code read for context, plus the affected test classes run locally.

### What I verified

- **Key naming**: `ODataQuery` (`getEntity`/`setEntity`) and `SharepointOnlineQuery` (`getSite`/`setSite`, `getList`/`setList`) are standard JavaBean properties reachable through `Introspector.getBeanInfo` — the same mechanism `TabularUtil.findProperties`/`getPropertyMap` uses. The literal keys `"entity"`, `"site"`, `"list"` this PR emits are exactly those property names, not the `@Property(label=...)` display strings. Confirmed by reading both query beans directly.
- **OData**: `ODataCatalog.buildSchema`'s `entitySetName` is the same value used to construct both `TabularDatasetRef` (the dataset id) and `params.get("entity")` — they can't drift apart because they come from the same local variable in the same loop iteration. `entitySetName` is guaranteed non-null before this point (only put into the map when `Tool.getAttribute(entitySet, "Name") != null`), so `Map.of("entity", entitySetName)` can't NPE.
- **SharePoint Online**: `describeDataset`'s `params` come from `SharepointDatasetId.parse(datasetId)`, which throws before returning if either component is blank — so `Map.of("site", ..., "list", ...)` can't NPE either, and the values are the connector's real unescaped site/list identifiers, not fragments of the composed/escaped id. Traced `parsed.list()` back to `SharepointOnlineRuntime.getLists`' `list[1]` (the Graph list id, not display name) and `parsed.site()` back to `SiteRef.id()` — both are exactly the values `SharepointOnlineQuery.setSite`/`setList` need and exactly what `getListColumns`/`doGetListColumns` already consume, so there's no display-name-vs-id mismatch.
- **Compatibility constructor**: grepped every `new TabularDatasetSchema(...)` call site in the repo — the two connector production sites now use the 4-arg form, and every test-file site (10 call sites across 3 files) still uses the 3-arg form and compiles/passes unchanged, matching the PR description's claim about the "C2-frozen" test files.
- **No other connector affected**: `TabularCatalogProvider` is implemented only by `ODataRuntime` and `SharepointOnlineRuntime` in production code (plus two test fakes) — the claim that this SPI has exactly two connectors is accurate.
- **"Reports nothing" path is unchanged**: `buildDatasetExtension`'s `if(params != null && !params.isEmpty())` guard means the COMMON extension JSON is byte-identical to pre-PR output when a connector uses the 3-arg constructor — covered by `datasetCommonExtensionOmitsParamsKeyWhenSchemaHasNone`.
- **`validateParams` null-tolerance** mirrors `validateKeyColumns`'s existing "defensive but not enforced" pattern and is consistent with it: a null `params` is absorbed identically downstream (`buildDatasetExtension`'s own null check), the same reasoning already given for `keyColumns`.
- **Tests aren't tautological**: read every new/changed test and its fixtures (`ODataCatalogTest`, `SharepointOnlineCatalogTest`, `TabularCatalogDatasetExtensionTest`, `TabularCatalogServiceContractValidationTest`). Each asserts exact `Map` equality or a specific key/value, not mere non-nullness, and the SharePoint dot-escaping test specifically targets the case that motivated `SharepointDatasetId`'s escaping scheme in the first place.
- **Ran the targeted suite** (after clearing `target/surefire-reports` first, per this repo's known stale-report trap): `ODataCatalogTest` (5/5), `SharepointOnlineCatalogTest` (5/5), `TabularCatalogDatasetExtensionTest` (3/3), `TabularCatalogServiceContractValidationTest` (21/21), and the frozen `TabularCatalogServiceTest` (11/11, unmodified, still green) — all pass.

### What I could not verify

- I did not modify source to reproduce "each new test fails against a reverted implementation" myself — my instructions here are review-only, so I relied on reading the test bodies and confirming each asserts a specific value/shape that the pre-PR code path could not produce, rather than executing the revert-and-run.
- The claim that `ServerFile` already uses a `params` slot in the COMMON extension "for the same purpose" is about the wiz-services repo, which is out of scope for this review — I did not and could not check it from here.

### The one nit

Inline comment on `TabularCatalogService.java:271` — `validateParams` rejects a null value but not a blank/empty-string one, which is asymmetric with `validateColumnNames`'s `isBlank()` check just above it. Not exploitable by either connector today (neither can produce a blank `site`/`list`/`entity`), so not a blocker — just a symmetry gap worth a follow-up or a comment explaining the asymmetry.

### On the two "deliberate, don't relitigate" points

Both hold up on inspection: the unenforced invariants (`params`/`keyColumns` null-tolerance) have stated, applicable reasoning in the code itself, and the dataset id format / SharePoint escaping are genuinely untouched by this diff (confirmed via `SharepointDatasetId.java`, which isn't in the diff and whose escaping logic this PR only calls, not modifies).

fredaan (approved):

PR #4914, opened 2026-09-01
fix(pva-015): stop re-deriving runtime id in ChangeChartTypeService's nested proxy calls

fredaan:
## Summary

- `ChangeChartTypeService.changeChartType`'s map-retype branch and its `handleMulti`/multi-style-toggle branch each made a nested in-process call into another controller (`VSBindingTreeController.getBinding`, `ChangeSeparateStatusController.changeSeparateStatus`) that re-derives its runtime id from the message-scoped `RuntimeViewsheetRef` instead of reusing the `id` that `changeChartType` already has as its own `@ClusterProxyKey` parameter.
- When the outer `changeChartType` proxy call resolves "not local" (typically on a session's first clustered write), `ServiceProxyContext` tears down and rebuilds the thread's STOMP message context from a snapshot that can lack the `sheetRuntimeId` attribute. `RuntimeViewsheetRef.getRuntimeId()` then silently returns `null`, which reaches `RuntimeSheetCache.getAffinityKey` and crashes with Ignite's `NullPointerException("Ouch! Argument cannot be null: key")`. A retry works because routing then resolves "local" and the message context is never rebuilt.
- Fix: call `VSBindingTreeControllerServiceProxy`/`ChangeSeparateStatusServiceProxy` directly with the id `changeChartType` already has, bypassing the ambient `RuntimeViewsheetRef` lookup, for both call sites.
- Out of scope (per diagnosis): hardening `RuntimeSheetCache.isLocal`/`getAffinityKey` itself, which is shared by ~236 other generated proxy call sites — left untouched to keep blast radius to the two confirmed call sites.

## Test plan

- [x] `ChangeChartTypeServiceTest`: added `aMapRetypeRefreshesTheBindingTreeWithTheKnownRuntimeId` and `aMultiStyleToggleUsesTheKnownRuntimeId`, both asserting the nested proxy call uses the harness's known `"rid"` (with a `null` `RuntimeViewsheetRef`, so a regression reintroducing the old lookup fails/NPEs here). Verified these two tests fail against a temporarily-reverted fix and pass against the real fix.
- [x] `./mvnw -o -pl core test -Dtest=ChangeChartTypeServiceTest` — 8/8 pass
- [x] `./mvnw -o -pl core test -Dtest=ChangeChartTypeServiceTest,ChangeChartTypeProcessorTest,ChartBindingServiceTest` — 79/79 pass (no collateral breakage in callers/related processor tests)
- [x] `./mvnw -o -pl core test-compile` — full module compiles clean

### DIFF
--- a/core/src/main/java/inetsoft/web/binding/controller/ChangeChartTypeService.java
+++ b/core/src/main/java/inetsoft/web/binding/controller/ChangeChartTypeService.java
@@ -65,19 +65,19 @@ public class ChangeChartTypeService {
       VSBindingService bindingFactory,
       RuntimeViewsheetRef runtimeViewsheetRef, CoreLifecycleService coreLifecycleService,
       ChartRefModelFactoryService chartRefService,
-      ChangeSeparateStatusController changeSeparateController,
+      ChangeSeparateStatusServiceProxy changeSeparateStatusService,
       VSAssemblyInfoHandler assemblyInfoHandler, VSChartHandler chartHandler,
-      VSBindingTreeController bindingTreeController,
+      VSBindingTreeControllerServiceProxy vsBindingTreeService,
       ViewsheetService viewsheetService)
    {
       this.bindingFactory = bindingFactory;
       this.runtimeViewsheetRef = runtimeViewsheetRef;
       this.coreLifecycleService = coreLifecycleService;
       this.chartRefService = chartRefService;
-      this.changeSeparateController = changeSeparateController;
+      this.changeSeparateStatusService = changeSeparateStatusService;
       this.assemblyInfoHandler = assemblyInfoHandler;
       this.chartHandler = chartHandler;
-      this.bindingTreeController = bindingTreeController;
+      this.vsBindingTreeService = vsBindingTreeService;
       this.viewsheetService = viewsheetService;
    }
 
@@ -150,7 +150,7 @@ public class ChangeChartTypeService {
       if(oldType == newType) {
          new ChangeChartTypeProcessor(oldType, newType,
                                       omulti, nmulti, ref, cinfo, false, desc).processMultiChanged();
-         handleMulti(name, omulti, nmulti, separate, chart, principal, dispatcher, linkUri);
+         handleMulti(id, name, omulti, nmulti, separate, chart, principal, dispatcher, linkUri);
 
          if(ostackMeasures == nstackMeasures) {
             // clearRuntime() above discarded the runtime aesthetic fields, and this is the only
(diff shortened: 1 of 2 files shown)


fredaan (commented):
Reviewed the diff (`ChangeChartTypeService.java` + `ChangeChartTypeServiceTest.java`, single commit, no other files touched) and re-ran CI to completion.

**Diff correctness**
- Scope is minimal: only `ChangeChartTypeService`'s constructor/fields (swap `ChangeSeparateStatusController`/`VSBindingTreeController` for their `*ServiceProxy` counterparts) and its two nested call sites (map-retype branch, `handleMulti`), which now pass the already-known `id` (`changeChartType`'s own `@ClusterProxyKey` param) straight to the proxy instead of re-deriving it from the message-scoped `RuntimeViewsheetRef`. No unrelated refactors.
- Confirmed the other legitimate callers of `VSBindingTreeController.getBinding(...)` — `ChangeGeographicService`, `ConvertChartRefService`, `ConvertTableRefControllerService`, `ModifyCalculateFieldService` — are untouched and still go through the message-scoped path, which is correct for them (they run inside their own `@MessageMapping` handlers with a live WebSocket session, unlike `ChangeChartTypeService`'s two inner calls which were piggybacking on an ambient ref that may not survive cluster-proxy re-routing). `ChangeSeparateStatusController`/`VSBindingTreeController` classes themselves are not modified or removed, only bypassed at these two specific call sites.
- The two new tests are genuine regression tests, not tautologies: the harness injects `null` for `RuntimeViewsheetRef` and mocks of the new `*ServiceProxy` types, then asserts `getBinding`/`changeSeparateStatus` were invoked with the literal `"rid"` passed into `changeChartType(...)`. A regression that reintroduced the old `runtimeViewsheetRef.getRuntimeId()` lookup would NPU/mismatch here (null ref) rather than passing silently. Verified locally that only these two production files changed vs `stylebi-wiz-main-rebase`.

**CI**
- `claude-review`, Semgrep Scan/OSS, `license/cla`: pass.
- `build` (Validation Build, full `core` module `mvn test`): 8308 tests run, exactly one failure — `inetsoft.uql.ConditionTest$RegressionTests.toSqlConditionSilentlyDefaultsOnAnUnmatchedName` (a `java.sql.Date` equality mismatch in SQL date-condition parsing, unrelated to chart binding/controllers). This matches the previously-confirmed pre-existing `ConditionTest` flake, not a regression from this change.
- Targeted suites from the PR's own test plan all passed in this run: `ChangeChartTypeServiceTest` 8/8, `ChangeChartTypeProcessorTest` 25/25, `ChartBindingServiceTest` 46/46.

No blocking issues found. Approving via squash-merge (my own PR, so `--approve` isn't usable — recording the review here instead).

PR #4915, opened 2026-09-01
fix(wiz): wire GroupingService into edit(op:"group_as_tab") for Tab creation

fredaan:
## Summary

- No `mcp__composer-chat__*` tool could create a `Tab` viewsheet assembly. `edit(op:"group")` only ever reached `ComposerGroupService.groupComponents`, which unconditionally builds a `GroupContainerVSAssembly` -- there was no path to a `Tab`. `GroupingService.groupComponents` (the service native Composer's own tab-combining gesture uses, already a plain injectable bean) was never wired into `ViewsheetEditService`.
- Adds a new op, `group_as_tab`, that resolves the requested assembly names and calls `GroupingService.groupComponents` pairwise -- the first two names form the tab, and each additional name folds into that same tab (its first assembly is by then already a tab member, which routes into `GroupingService`'s "append to existing tab" branch).
- Closes the resulting ungroup-symmetry gap: `ComposerGroupService.ungroup` only unwraps a `GroupContainerVSAssembly` and silently returns `null` on anything else, so once a Tab can be created, `edit(op:"ungroup")` on it would report success while doing nothing. `edit(op:"ungroup")` now fails loud with a named error when the target is a Tab, pointing the caller at `move_from_container` instead (which already correctly collapses a Tab once only one child remains).

This is the server half of a two-repo fix; the `stylebi-wiz`/`plugin/composer` PR that exposes `group_as_tab` through the plugin depends on this landing first.

## Test plan

- [x] `./mvnw test -pl core -Dtest=ViewsheetEditServiceTest` -- 50/50 pass, including 3 new tests: `group_as_tab` rejects <2 assemblies, `group_as_tab` calls `GroupingService.groupComponents` once per pair with `selection=false`, `ungroup` on a Tab fails loud instead of silently no-oping.
- [x] `./mvnw test -pl core -Dtest=ViewsheetEditServiceTest,ViewsheetAssemblyAgentControllerTest,WorksheetAgentControllerTest,ComposerGroupServiceTest,GroupingServiceTest,LayoutOptionDialogServiceTest,ComposerBindingControllerTest` -- 197/197 pass (every test file in the repo that references `ComposerGroupService`, `GroupingService`, or constructs `ViewsheetEditService` directly, since its constructor signature changed).

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/ViewsheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/ViewsheetEditService.java
@@ -19,6 +19,7 @@ package inetsoft.web.wiz.viewsheet;
 
 import inetsoft.uql.asset.Assembly;
 import inetsoft.uql.viewsheet.TableDataVSAssembly;
+import inetsoft.uql.viewsheet.TabVSAssembly;
 import inetsoft.uql.viewsheet.TitledVSAssembly;
 import inetsoft.uql.viewsheet.VSAssembly;
 import inetsoft.uql.viewsheet.Viewsheet;
@@ -26,6 +27,7 @@ import inetsoft.web.composer.vs.event.CopyVSObjectsEvent;
 import inetsoft.web.composer.vs.objects.controller.ClipboardControllerService;
 import inetsoft.web.composer.vs.objects.controller.ComposerGroupService;
 import inetsoft.web.composer.vs.objects.controller.ComposerObjectService;
+import inetsoft.web.composer.vs.objects.controller.GroupingService;
 import inetsoft.web.composer.vs.objects.controller.VSObjectPropertyService;
 import inetsoft.web.composer.vs.objects.event.*;
 import inetsoft.report.composition.RuntimeViewsheet;
@@ -62,6 +64,7 @@ public class ViewsheetEditService {
                                VSObjectPropertyService propertyService,
                                ViewsheetReadService reader,
                                ComposerGroupService groups,
+                               GroupingService grouping,
                                VSRefreshService refreshService)
    {
       this.sessions = sessions;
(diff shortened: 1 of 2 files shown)


fredaan (commented):
Reviewed the diff directly (not just relying on the reported test counts).

**GroupingService wiring**
- `git grep -n "new ViewsheetEditService("` across the PR branch shows every construction site is in `ViewsheetEditServiceTest.java` (4 test-helper factories, all updated to pass a `GroupingService` mock). `ViewsheetEditService` is a `@Service` with a single `@Autowired` constructor and `GroupingService` is itself an existing `@Controller`-annotated Spring bean, so there's no XML/manual-wiring path to miss — Spring resolves the new constructor param by type automatically. Confirmed no other call site (prod or test) was left on the old 7-arg signature.

**`group_as_tab` correctness**
- Traced `GroupingService.groupComponents` (unmodified by this PR): when neither `target` nor `object` is already a Tab, it hits the `else` branch that allocates a real `new TabVSAssembly(...)`, adds it to the viewsheet, and sets both children's `assemblies`/offsets — this is genuinely a `TabVSAssembly`, not a `GroupContainer`.
- For the pairwise fold (`first` reused across loop iterations against each subsequent name), `AbstractVSAssembly.getContainer()` resolves the container by scanning the live `Viewsheet`'s assemblies for one whose `containsAssembly(getName())` is true — it's not a cached/stale reference — so on iteration 2+, `first.getContainer()` correctly resolves to the Tab created in iteration 1, routing into `groupComponents`'s "already a Tab" append branch. The pairwise-fold approach is sound.
- `requireVSAssembly` correctly throws on both "not in the wiz read-model" (via `requireExisting`) and "resolved to null off `vs.getAssembly`" — reasonable belt-and-suspenders given the two representations (`ViewsheetModel` reader vs. live `Viewsheet`) this file already juggles elsewhere.

**Ungroup fail-loud scoping**
- Confirmed via `git grep -n "ViewsheetEditService"` that this class is only ever injected into `ViewsheetAssemblyAgentController` (`inetsoft.web.wiz.viewsheet` package) — the wiz-agent-only controller. The diff does not touch `ComposerGroupService.java` at all, so native Composer's human-facing "Ungroup" on a Tab is untouched: it will keep silently no-op'ing exactly as it did before this PR, which is the correct call for a bug that already exists in shipped, human-facing Composer behavior and is out of scope here.
- The new guard is `instanceof TabVSAssembly` checked only inside `ViewsheetEditService.ungroup`, before the existing `groups.ungroup(...)` delegation — additive, doesn't alter that delegation's behavior for the `GroupContainer` case.

**Tests**
- The three new tests exercise the right things (arity guard, fail-loud-on-Tab guard with both the assembly name and "Tab" in the message, and the exact pairwise `groupComponents` call sequence with `verifyNoMoreInteractions`).

**CI**
- `build` failed, but the only failure is `inetsoft.uql.ConditionTest$RegressionTests.toSqlConditionSilentlyDefaultsOnAnUnmatchedName` (a `java.sql.Date` equality assertion that prints identical values `2027-01-01` on both sides — classic TZ-offset flake), out of 8309 tests run. This is the known pre-existing `ConditionTest` flake called out as unrelated to wiz-agent changes; nothing in `ViewsheetEditServiceTest`/`ComposerGroupServiceTest`/`GroupingServiceTest`/the controller tests failed. Not a real regression from this diff.

No blocking issues found. Squash-merging.

PR #4917, opened 2026-09-01
feat(wiz): bind a column to selection-family assemblies (SelectionBindingService)

fredaan:
## Summary

No wiz-facing write path can point a `SelectionList`/`SelectionTree`/`TimeSlider`/`Calendar`
assembly at a table/column:

- `TableBindingService.setSource` resolves a `BaseTableBindingModel` through
  `VSBindingService`, which has no `VSBindingFactory` registered for any of these four
  types — their real shape (one table + one column, or one table + an ordered column
  list) is not a shelf collection, so there is no polymorphic model to build.
- `AssemblyPropertyService`'s raw-dotted-path escape hatch can reach a plain-`String`
  field like `selectionListPaneModel.selectedTable`, but can never construct the
  structured `OutputColumnRefModel`/`ColumnRef[]` these types' column fields need —
  `PropertyPath.coerce()` only knows how to coerce scalars.

`SelectionBindingService` is a new, narrow, independent write surface (peer of
`TableBindingService`, not merged into `AssemblyPropertyService`'s bindings map) that
drives the exact public `get<Type>PropertyModel` / `set<Type>PropertyModel` round trip
the Composer's own property dialogs already use for every other field on these types —
only the table/column part of the model it reads back is touched, so nothing else on
these types' property surface can regress.

- `SelectionListVSAssembly` / `CalendarVSAssembly`: exactly one column.
- `SelectionTreeVSAssembly`: one or more columns, in hierarchy order (`COLUMN` mode).
- `TimeSliderVSAssembly`: one column → `SingleTimeInfo` (range type inferred the same
  way `AddFilterService.createFilterAssembly` already does); two or more → `CompositeTimeInfo`.

Adds `POST /api/wiz/v1/agent/binding/{sessionToken}/selection/source` on
`BindingAgentController`.

## Design notes / deviations from the initial diagnosis

- The diagnosis suggested resolving table/columns the same way
  `TableBindingService.resolveTable` does. That method is tied to `BaseTableBindingModel`,
  which cannot be obtained for these four types (that's the exact gap this PR fixes), so
  it can't be reused directly. Table/column resolution instead calls
  `BindableFieldsService.list(runtimeId, null, user)` **unscoped** — a *scoped* call
  returns an empty listing for all four types today, because `VSBindingTreeService.getBinding`
  only builds a tree for a `ChartVSAssemblyInfo`/`TableDataVSAssemblyInfo`, and none of
  these four are either. The unscoped call reads the same worksheet-wide tree a fresh,
  unbound table/crosstab would also see.
- Rather than extracting the dialog services' private `ColumnRef`-building helpers into a
  shared method (touching four already-working, already-tested dialog services), this
  service drives their existing public get/set round trip and builds its own
  `OutputColumnRefModel` — the same shape those methods already accept from any caller.
  No changes to `SelectionListPropertyDialogService`/`SelectionTreePropertyDialogService`/
  `RangeSliderPropertyDialogService`/`CalendarPropertyDialogService` internals.
- `SelectionRuntimeService.requireSelection` is now `public static` (was
  `private static`) so this service can reuse the same assembly type-check rather than
  duplicating it, per the diagnosis's own suggestion.
- The "repoint without force" refusal only fires on an actual **table** change while
  already bound; rebinding within the same table (a different column, or a different
  hierarchy) does not require `force`, mirroring `TableBindingService`'s own
  same-source tolerance.

## Test plan

- [x] New `SelectionBindingServiceTest` (14 tests): one happy path per type, wrong-arity
      refusals, unknown table/column refusals (byte-for-byte "cannot bind to" shape),
      repoint-without-force refusal + force success.
- [x] `BindingAgentControllerTest` updated for the new constructor argument.
- [x] `./mvnw test -pl core -Dtest=inetsoft.web.wiz.**` — 208 test classes / 2827 tests,
      0 failures, 0 errors (full `inetsoft.web.wiz` package, not just the new files).

Design plan: `docs/teams/2026-08-31-test-composer-plugin/case-PVA-016-selection-binding-gap/01-diagnosis.md`
in stylebi-wiz. Paired stylebi-wiz PR (wiz-services proxy + `set_selection_source` plugin
tool) to follow.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
@@ -64,7 +64,8 @@ public class BindingAgentController {
                                  ChartBindingService chartService,
                                  ChartAestheticAgentService aestheticService,
                                  TableBindingService tableService,
-                                 CalcTableService calcService)
+                                 CalcTableService calcService,
+                                 SelectionBindingService selectionService)
    {
       this.feature = feature;
       this.joinService = joinService;
@@ -76,6 +77,7 @@ public class BindingAgentController {
       this.aestheticService = aestheticService;
       this.tableService = tableService;
       this.calcService = calcService;
+      this.selectionService = selectionService;
    }
 
    public record JoinRequest(String code) {}
@@ -412,6 +414,23 @@ public class BindingAgentController {
       requireEnabled();
       tableService.setSource(sessionToken, user, request.assembly(), request.table(),
                              Boolean.TRUE.equals(request.force()));
+   }
+
+   public record SelectionSourceRequest(String assembly, String table, List<String> columns,
+                                        List<String> additionalTables, String measure,
+                                        Boolean force) {}
+
+   @PostMapping("/api/wiz/v1/agent/binding/{sessionToken}/selection/source")
+   public Map<String, Object> setSelectionSource(
+      @PathVariable String sessionToken, @RequestBody SelectionSourceRequest request,
+      @RequestParam(required = false, defaultValue = "") String linkUri, Principal user)
+      throws Exception
+   {
+      requireEnabled();
+      return selectionService.setSource(sessionToken, user, request.assembly(), request.table(),
+                                        request.columns(), request.additionalTables(),
+                                        request.measure(), Boolean.TRUE.equals(request.force()),
+                                        linkUri);
    }
 
    @PostMapping("/api/wiz/v1/agent/binding/{sessionToken}/table/field/add")
@@ -754,4 +773,5 @@ public class BindingAgentController {
    private final ChartAestheticAgentService aestheticService;
    private final TableBindingService tableService;
    private final CalcTableService calcService;
+   private final SelectionBindingService selectionService;
 }
--- a/core/src/main/java/inetsoft/web/wiz/binding/SelectionBindingService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/SelectionBindingService.java
@@ -1,0 +1,375 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.web.wiz.binding;
+
+import inetsoft.uql.schema.XSchema;
+import inetsoft.uql.viewsheet.*;
+import inetsoft.uql.viewsheet.internal.SelectionTreeVSAssemblyInfo;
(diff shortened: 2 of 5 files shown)


fredaan (commented):
Reviewed the diff, the diagnosis doc it implements (PVA-016), and the four deviations from the original proposal against source. Verdict: sound, merging.

**Scope check** — `git diff` against the merge-base confirms zero changes to the four dialog services (`SelectionListPropertyDialogService`, `SelectionTreePropertyDialogService`, `RangeSliderPropertyDialogService`, `CalendarPropertyDialogService`): the new service is the only new caller of their existing public `get<Type>PropertyModel`/`set<Type>PropertyModel` methods, and the call signatures match exactly (`runtimeId, objectId, value, linkUri, principal, dispatcher`). Since `set*PropertyModel` writes back every field it reads off the model (title, general, size/position, script pane, etc.), the get-then-mutate-table/column-then-set round trip is a true no-op on everything else — confirmed by reading `SelectionListPropertyDialogService.setSelectionListPropertyModel`'s body directly, not just trusting the class comment.

**Deviation 1 (unscoped `BindableFieldsService.list`)** — checked `BindableFieldsService.list`'s own doc comment and body: `assembly == null` is documented as the caller's deliberate "list everything," identical to what any other unbound tool/crosstab already receives before it has a source. This isn't a new leak — it's the same worksheet-wide tree `list_bindable_fields` already returns with no `assembly` argument today.

**Deviation 2 (own `OutputColumnRefModel` instead of extracting the dialogs' private helpers)** — confirmed via diffstat that this kept the blast radius to one new file + two small, mechanical touches (constructor wiring, one visibility bump). Reasonable trade against extracting shared private helpers out of four existing dialog services for a first cut.

**Deviation 3 (skip client-side arity pre-check)** — `requireArity` still throws a clear, field-named `IllegalArgumentException` server-side before touching any model (e.g. "'List1' is a selection list, which needs exactly 1 column in 'columns', got 2."), so this isn't a silent-acceptance gap under CLAUDE.md's tool-misuse framing — it fails loud, just one layer down from where the diagnosis originally proposed the check.

**Deviation 4 (narrower repoint-without-force)** — checked `TableBindingService.setSource`/`requireNoBoundFields`: its own tolerance is exactly "same table = no discard needed," with no finer distinction below the table level either. `SelectionBindingService.requireRepoint`'s "same table, different column doesn't need force" is a faithful mirror of that existing convention, not a new invention.

**REST endpoint** — `POST .../selection/source` takes `Principal user` + `@RequestParam(required=false, defaultValue="") String linkUri`, matching the `linkUri`-param convention used by every sibling endpoint on this controller that round-trips through a dialog service (chart aesthetic/type endpoints); `setTableSource` is the outlier that skips `linkUri` only because it never touches a dialog service. `requireEnabled()` is called first, matching every other action on this controller.

**`SelectionRuntimeService.requireSelection` → `public static`** — diffed the method body: unchanged, visibility-only. New caller only.

**Tests** — `SelectionBindingServiceTest.java`'s 14 cases cover all four assembly types (list/tree/slider/calendar) with real assertions on the resulting pane model (selected table, column attribute, mode, composite flag), both slider composite/non-composite branches, wrong-arity refusals for list/tree/calendar, unknown-table and unknown-column refusals (asserting the message names both the bad value and what's available), and both repoint-without-force (refused) and repoint-with-force (succeeds) — not trivial happy-path-only coverage.

**CI** — all checks green after the Maven build finished (was pending during initial review pass, waited it out): `build` pass (19m22s), `claude-review` pass, both Semgrep scans pass, CLA signed. No conflicts, `mergeStateStatus: CLEAN`.

**Caveat carried forward, not a blocker** — no live StyleBI verification was done (unit/mocked-integration only, per the PR description); this is a new capability addition rather than a regression fix for something previously confirmed working, and the unit coverage plus source-level cross-checks above give enough confidence for a first merge. Live verification (including via the dependent `stylebi-wiz#2048` client PR once it lands) should follow as a separate step.

Merging.

PR #4922, opened 2026-09-01
fix(wiz): route edit op:"refresh" through refreshVsAssembly, not refreshVsAssemblyView (PVA-010)

fredaan:
## Summary
- `ViewsheetEditService.refresh` (`core/src/main/java/inetsoft/web/wiz/viewsheet/ViewsheetEditService.java`) called `VSRefreshService.refreshVsAssemblyView` -- the wrong one of two similarly-named sibling methods. That method only recomputes an assembly's info/format and re-pushes the render; it never calls `ViewsheetSandbox.reset(...)` or re-executes the bound query, so a worksheet-side change (e.g. `set_ranking` applied after a chart was already bound) never propagated to `edit(op:"refresh")`'s result, even though `{ok:true}` was returned.
- Fixed by constructing a `RefreshVSAssemblyEvent` and calling `VSRefreshService.refreshVsAssembly` instead -- the same `/vs/refresh/assembly` path the Composer UI's own binding-pane "assembly changed" cascade uses, which resets the sandbox and re-executes the assembly's query via `CoreLifecycleService.execute(...)`.
- Updated the method's javadoc to describe the corrected mechanism, and updated `ViewsheetEditServiceTest.refreshDelegatesToVSRefreshServiceForTheNamedAssembly` (previously asserting the wrong delegate) to verify `refreshVsAssembly` is called with the right `RefreshVSAssemblyEvent`, plus the neighboring `refreshViewsheetDelegatesToTheWholeSheetRefresh` test's `never()` assertion (unaffected in behavior, just naming the correct sibling method).

Root cause and full analysis: `docs/teams/2026-08-31-test-composer-plugin/case-PVA-010-refresh-op-stale-query/01-diagnosis.md` in the `stylebi-wiz` repo.

## Test plan
- [x] `./mvnw -pl core -Dtest=ViewsheetEditServiceTest test` -- all 50 tests pass, including the two updated assertions.
- [x] `./mvnw -pl core -Dtest=inetsoft.web.wiz.**` test -- 2840 tests pass (1 skipped), 0 failures/errors -- no regressions in the wiz package this change lives in.
- [x] `./mvnw -pl core -Dtest=VSRefreshServiceEmbedAssemblyTest,CoreLifecycleServiceEmbedSizeTest test` -- unaffected sibling-method tests still pass.
- [ ] Live verification against a running StyleBI instance (bind a chart, `set_ranking` + `save_worksheet()`, `edit(op:"refresh")`, confirm `get_viewsheet_image` shows requeried data) was **not** performed in this session -- no `composer-chat` MCP tools were available to this agent. This should be confirmed once a build with this fix is deployed.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/ViewsheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/ViewsheetEditService.java
@@ -33,6 +33,7 @@ import inetsoft.web.composer.vs.objects.event.*;
 import inetsoft.report.composition.RuntimeViewsheet;
 import inetsoft.report.composition.execution.ViewsheetSandbox;
 import inetsoft.web.viewsheet.controller.VSRefreshService;
+import inetsoft.web.viewsheet.event.RefreshVSAssemblyEvent;
 import inetsoft.web.viewsheet.event.VSRefreshEvent;
 import inetsoft.web.wiz.service.RenderNotReadyException;
 import inetsoft.web.wiz.service.RenderWaitSupport;
@@ -517,9 +518,10 @@ public class ViewsheetEditService {
    /**
     * PVA-010: forces the named assembly to re-execute its query and re-render, picking up a
     * worksheet-side change (e.g. a ranking/aggregate edit) made after the chart was already
-    * bound — the same mechanism {@code VSRefreshService.refreshVsAssemblyView} already performs
-    * for the Composer's own UI refresh (its own {@code TableDataVSAssembly} reload included), just
-    * not previously reachable from any composer-chat tool.
+    * bound — the same mechanism {@code VSRefreshService.refreshVsAssembly} (reached via
+    * {@code /vs/refresh/assembly}) already performs for the Composer's own UI when the binding
+    * pane changes an assembly's binding, including resetting the sandbox and re-executing the
+    * assembly's query, not merely re-rendering it.
     */
    private void refresh(String sessionToken, Principal user, EditRequest request, String linkUri)
       throws Exception
@@ -528,8 +530,11 @@ public class ViewsheetEditService {
 
       sessions.mutate(sessionToken, user, (rvs, runtimeId, dispatcher) -> {
          requireExisting(rvs, request.assembly());
-         refreshService.refreshVsAssemblyView(runtimeId, request.assembly(), dispatcher, linkUri,
-                                              user);
+         RefreshVSAssemblyEvent event = RefreshVSAssemblyEvent.builder()
+            .vsRuntimeId(runtimeId)
+            .assemblyName(request.assembly())
+            .build();
+         refreshService.refreshVsAssembly(runtimeId, event, dispatcher, linkUri, user);
       });
    }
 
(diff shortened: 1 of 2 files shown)


fredaan (commented):
Reviewed the diff against `VSRefreshService.java`/`VSRefreshController.java` source directly, not just the PR description.

**Correctness verified:**
- `VSRefreshService.refreshVsAssembly` signature: `refreshVsAssembly(@ClusterProxyKey String runtimeId, RefreshVSAssemblyEvent event, CommandDispatcher dispatcher, @LinkUri String linkUri, Principal principal)`. `ViewsheetEditService.refresh`'s new call matches this exactly in order/type.
- Traced `refreshVsAssembly` → `refreshAssemblyAndDependencies`: confirmed it does `box.reset(null, rarr, clist, false, false, null)` on the `ViewsheetSandbox` followed by `coreLifecycleService.execute(...)` — i.e. it genuinely re-executes the bound query, matching the corrected javadoc. `refreshVsAssemblyView` (the old, wrong call) only calls `coreLifecycleService.refreshVSAssembly`, which never resets the sandbox.
- `RefreshVSAssemblyEvent.vsRuntimeId`/`assemblyName` have `@Value.Default`, `assemblySize` is `@Nullable`, `embed` defaults false — the new builder call (`vsRuntimeId` + `assemblyName` only) is complete and valid, no missing required field.
- `refreshDelegatesToVSRefreshServiceForTheNamedAssembly` genuinely exercises the corrected interaction (captures the `RefreshVSAssemblyEvent`, asserts `assemblyName`, asserts `refreshVsAssemblyView` is `never()` called) — this is not a trivial flip of the old assertion; it would fail against the pre-fix code. The neighboring `refreshViewsheetDelegatesToTheWholeSheetRefresh`'s `never()` target was correctly updated to `refreshVsAssembly` since that's now the method actually reachable from `edit(op:"refresh")`.

**Scope discipline verified** (`git diff --name-only` vs. `stylebi-wiz-main-rebase`): only `ViewsheetEditService.java` and `ViewsheetEditServiceTest.java` touched. `VSRefreshService.java`/`VSRefreshController.java` (owning `refreshVsAssemblyView`, its STOMP endpoint, and its one real caller) have an empty diff — untouched. `refreshWholeSheet`/`resetRuntime` also untouched, correctly out of scope.

**CI:** `build` passed (19m35s), Semgrep Scan/OSS pass, CLA signed. `claude-review` shows failed, but its job log shows an infra-level bot failure (`is_error:true` after 321ms / 1 turn, before any tool ran) — not a code finding — so it's not being treated as blocking.

**Known, disclosed gap:** no live verification against a running StyleBI instance was performed in this review either (same constraint as the fixer — no live composer-chat/MCP session available here). The fix is verified at the source level (method body confirmed to reset the sandbox + re-execute) and by an updated unit test that would have failed pre-fix, but the original live repro (apply a ranking, observe stale chart; clear it, observe stale chart) has not been re-run against a live instance since this change. Flagging this as a follow-up to do once this merges and a human rebuilds/restarts the shared dev StyleBI instance — not a merge blocker for an otherwise scoped, source-verified, regression-tested fix.

No blocking issues found. Merging.

PR #4923, opened 2026-09-01
fix(wiz): resolve a Table highlight's omitted region to a real data cell, not the header (PVA-017)

fredaan:
## Summary

`set_highlight` (and `list_highlights`/`delete_highlight`) on a plain worksheet-bound `Table`,
called with `row`/`col` omitted, silently resolved to the HEADER cell `(0,0)` instead of falling
forward to a real data cell — for both `applyRow:true` and plain (per-cell) highlight modes.
The identical call shape on a `Crosstab` already works correctly.

Root cause: `AssemblyHighlightService.read()` only falls forward from the header to the first
real data cell when the header cell's `fields` list is empty
(`AssemblyHighlightService.java:290`, pre-fix). That's a reliable "we're on the header" signal
for a `Crosstab` (`HighlightService.getRefsForVSAssembly`'s Crosstab branch is `dpath`-aware and
genuinely returns nothing for a header path), but false for a plain `Table`: its branch
(`HighlightService.java:469-477`) returns the table's entire base column selection regardless of
`row`/`col`/`dpath`, so the header cell's `fields` come back non-empty and the fall-forward never
fires. `model.getRow()/getCol()` stay `(0,0)`, and `HighlightDialogService.updateHighlights`
later derives the stored `TableDataPath` (both the per-cell and the `applyToRowHighlight` paths)
from that same header-anchored row/col.

## Fix

Confined to the agent-only service (`AssemblyHighlightService.read()`), per the diagnosis's
lower-blast-radius option: add a direct header-extent check gated on `model.isTableAssembly()`
(true only for `TableVSAssemblyInfo`, never for `CrosstabVSAssemblyInfo`), comparing the target
cell against the assembly's own reported first-data-cell extent
(`HighlightDialogService.getFirstDataCell`). When a Table's omitted region lands inside the
header, it now falls forward to the real first data cell the same way a Crosstab already does.
`HighlightService.getRefsForVSAssembly` (the shared helper also used by the Composer UI's native
Highlight/Hyperlink dialogs) is untouched — no UI blast radius.

## Test plan

- [x] `AssemblyHighlightServiceTest` — 27/27 pass, including two new regression tests
      (`fallsForwardToADataCellOnATableEvenWhenTheHeaderReportsNonEmptyFields`,
      `resolvesToADataRowOnATableForApplyRowTooEvenWhenTheHeaderReportsNonEmptyFields`) and one
      new control test confirming Crosstab's existing correct behavior is unchanged
      (`stillFallsForwardOnACrosstabByTheExistingEmptyFieldsSignal`).
- [x] Broader regression scope: `AssemblyHighlightServiceTest` +
      `HighlightServiceTest` + `HighlightDialogServiceTest` + `ViewsheetAssemblyAgentControllerTest`
      — 79/79 pass.
- [x] Full `inetsoft.web.wiz.viewsheet` package — 585/585 pass.
- [ ] Live verification against a running StyleBI instance was not possible in this session (no
      `mcp__...composer-chat__*` tools available) — this fix requires a server rebuild+restart to
      take effect, noted as typically a human-driven step.

Diagnosis: `stylebi-wiz` `docs/teams/2026-08-31-test-composer-plugin/case-PVA-017-table-highlight-header-paint/01-diagnosis.md`.
Bug filed as PVA-017 in `docs/bug-reports/plugin-composer-viewsheet-assembly-lifecycle.md`.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyHighlightService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyHighlightService.java
@@ -287,16 +287,33 @@ public class AssemblyHighlightService {
       // highlight somewhere they did not ask for.
       // A null model means the assembly has no highlight dialog at all; that is the caller's
       // answer, and re-reading a different cell of it would only produce the same null.
-      if(region == null && model != null && fieldNames(model).isEmpty()) {
-         HighlightDialogModel atData =
-            readAt(runtimeId, assemblyName, firstDataCell(runtimeId, assemblyName, user), user);
+      if(region == null && model != null) {
+         // `fieldNames(model).isEmpty()` is the general "we're stuck on the header" signal, and
+         // it is reliable for a Crosstab (its fields are computed dpath-aware, so a header cell
+         // genuinely reports none). It is NOT reliable for a plain Table:
+         // HighlightService.getRefsForVSAssembly's Table branch reports the table's entire base
+         // column list regardless of row/col, so `fields` comes back non-empty even at (0,0) —
+         // the fall-forward below would never fire and the header would be highlighted silently.
+         // For a Table, check the header extent directly instead.
+         boolean tableAtHeader = model.isTableAssembly() &&
+            isHeaderCell(target, firstDataCell(runtimeId, assemblyName, user));
 
-         if(!fieldNames(atData).isEmpty()) {
-            return atData;
+         if(tableAtHeader || fieldNames(model).isEmpty()) {
+            HighlightDialogModel atData =
+               readAt(runtimeId, assemblyName, firstDataCell(runtimeId, assemblyName, user), user);
+
+            if(!fieldNames(atData).isEmpty()) {
+               return atData;
+            }
          }
       }
 
       return model;
+   }
+
+   /** Whether {@code cell} falls inside the assembly's own reported header extent. */
+   private static boolean isHeaderCell(Region cell, Region firstDataCell) {
+      return cell.row() < firstDataCell.row() || cell.col() < firstDataCell.col();
    }
 
    /**
--- a/core/src/test/java/inetsoft/web/wiz/viewsheet/AssemblyHighlightServiceTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/viewsheet/AssemblyHighlightServiceTest.java
@@ -293,6 +293,91 @@ class AssemblyHighlightServiceTest {
                    "must fall forward to (1, 2), the cell getFirstDataCell actually reported");
    }
 
+   /**
+    * PVA-017: a plain {@code Table}'s header cell (0,0) reports a NON-EMPTY field list —
+    * {@code HighlightService.getRefsForVSAssembly}'s Table branch returns the table's entire base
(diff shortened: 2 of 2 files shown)


fredaan (commented):
Reviewed the diff directly (not just the PR description) before merging.

**Scope check — confirmed clean.** `git diff --name-only` against `stylebi-wiz-main-rebase` shows only `core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyHighlightService.java` and its test. `HighlightService.getRefsForVSAssembly` (shared with the Composer UI's native Highlight/Hyperlink dialogs) is genuinely untouched — zero UI blast radius, as claimed.

**Class hierarchy check — confirmed independently.** Walked the actual source: `TableVSAssemblyInfo extends TableDataVSAssemblyInfo` directly, while `CrosstabVSAssemblyInfo extends CrossBaseVSAssemblyInfo extends TableDataVSAssemblyInfo` — siblings, not a subtype relation. And `model.isTableAssembly()` is only ever set via `HighlightDialogService.getHighlightDialogModel`'s `assemblyInfo instanceof TableVSAssemblyInfo` branch (line ~109), which a `CrosstabVSAssemblyInfo` can never satisfy. So the new `tableAtHeader` check is provably Crosstab-inert, not just inert-by-convention.

**Header-extent check — no off-by-one.** `isHeaderCell` compares against `firstDataCell()`, which wraps the pre-existing (unmodified) `HighlightDialogService#getFirstDataCell` — already the trusted source the Crosstab fall-forward has used all along, so this PR adds a caller, not new arithmetic. `Region.whole()` defaults row/col to `(0,0)`, and the compact constructor normalizes any null row/col to 0 on every construction path, so the comparison is well-defined.

**Fallback preserved.** The original `fieldNames(model).isEmpty()` signal is kept as an `||` alternative, not replaced — Crosstab's existing correct path is untouched logically as well as textually.

**Tests — verified non-trivial.** Read all three new tests against the `Harness`/`model()` helpers. The two Table tests use `model()` (non-empty fields) for the header stub — correctly reproducing the actual bug signature (Table's header reports non-empty fields, unlike Crosstab's). Since the mocked data-cell model also uses `model()` and therefore has textually-identical fields to the header stub, the tests correctly assert on *which cell was queried* (`verify(...).getHighlightDialogModel(..., eq(1), eq(0), ...)` and `verify(...).setHighlightDialogModel(..., eq(dataCell), ...)`) rather than on field content that wouldn't distinguish the two paths — that's the right assertion strategy given the fixture, not a weak test. The Crosstab control test reuses the pre-existing empty-fields-header pattern and confirms the new Table-only branch doesn't affect it.

**CI:** build (19m16s), claude-review, Semgrep OSS, Semgrep Scan, license/cla — all green, no conflicts (`mergeStateStatus: CLEAN`).

**Disclosed gap, not a merge blocker:** no live verification against a running StyleBI instance was done in the fixing session (no composer-chat MCP tools available then). This needs a human to rebuild+restart the shared dev StyleBI instance and re-verify PVA-017 live (`set_highlight` on a Table with `row`/`col` omitted now lands on a real data row, not the header) as a follow-up — tracking this as a to-do, not blocking the merge given the diff is narrowly scoped, the hierarchy/off-by-one concerns check out under direct inspection, and the regression coverage is real.

Merging.

PR #4924, opened 2026-09-01
Fix PCB-004: set_table_fields/add_table_field silently accept a shelf write with no source table

larryliang-inetsoft:
## Summary

`TableBindingService.setShelf` accepted and applied a non-empty shelf-field write on a crosstab/table assembly whose `SourceInfo` was `null` (no source table set via `set_table_source` yet) — no error, no check. The write silently succeeded and the assembly rendered empty. Unlike PCB-002's 3 fixed callers (`setChartShelf`/`setChartSingleShelf`/`setAestheticField`, which call `resolveSourceTable`/`BindableColumns.requireSource`), `setTableFields` never called that mechanism at all, and its own request (`TableShelfRequest`) carries no `table` parameter to resolve ambiguity from — so the fix is a flat "is a source already set" precondition check rather than a port of `resolveSourceTable`.

`TableBindingService.addField` (`add_table_field`) has the identical unguarded pattern via `applyWithContext`. Fixed in the same PR, with the guard scoped specifically to `addField`'s own call site rather than pushed into the shared `applyWithContext` helper — `removeField`/`moveField` must not get this guard (a sourceless assembly's shelves can never have anything on them to remove/move once `setShelf`/`addField` are guarded), confirmed by a regression test (`moveFieldIsNotRefusedEvenWithNoSource`).

## Root cause

- `BindingAgentController.setTableFields` → `TableBindingService.setShelf` (`core/src/main/java/inetsoft/web/wiz/binding/TableBindingService.java`) reads the assembly's real `SourceInfo` and passes it straight into `TableBindingMutator.setShelf` with no null check; the mutator has no source-conditioned branch at all.
- The column-existence check that does run first (`BindableColumns.require`) cannot catch this case by design: with no table marked current, it falls back to matching a column name across every worksheet table.

## Fix

- `TableBindingService.setShelf`: refuse a non-empty field write when `source == null` (empty/clearing writes still allowed).
- `TableBindingService.addField`: same guard, applied only at `addField`'s own call site inside `applyWithContext`'s lambda (not inside `applyWithContext` itself, to avoid regressing `removeField`/`moveField`).
- Updated 3 existing `TableBindingServiceTest` tests that previously passed only because their mocks were unstubbed (defaulting `getSourceInfo()` to `null`) to stub a real `SourceInfo`, so they keep testing what they're meant to test.
- New regression tests: no-source refusal / empty-write carve-out / has-source success for both `setShelf` and `addField`, a `moveField` non-regression test, and a `BindingAgentControllerTest` end-to-end test driving the real (unmocked) `TableBindingService` through the controller endpoint the plugin calls.

Full investigation: `docs/teams/2026-09-01-bugs-composer-groupG/PCB-004/01-diagnosis.md` and `02-refute.md` (verdict: SURVIVED) in the `stylebi-wiz` repo.

## Test plan

- [x] `./mvnw -pl community/core -o test -Dtest=TableBindingServiceTest,BindingAgentControllerTest` — 33 tests, all green

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/TableBindingService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/TableBindingService.java
@@ -74,6 +74,8 @@ public class TableBindingService {
          VSAssembly assembly = vs.getAssembly(assemblyName);
          inetsoft.uql.asset.SourceInfo source = assembly instanceof DataVSAssembly data
             ? data.getSourceInfo() : null;
+         requireSourceForFieldWrite(assemblyName, shelf, source,
+                                    fields == null ? 0 : fields.size());
          TableBindingMutator.setShelf(model, shelf, fields, rvs, source, refModelService);
 
          ApplyVSAssemblyInfoEvent event = new ApplyVSAssemblyInfoEvent();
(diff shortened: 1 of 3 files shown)


jshobe-inetsoft (approved):

PR #4925, opened 2026-09-01
Bug #76402, re-sync ThreadLocal-cached Calendars to week.start on every use

jkidacki-inetsoft:
## Summary

Fixes [Bug #76402](https://issues.inetsoft.com/issues/76402) (relates to #75352, Year-SameWeek date comparison data wrong).

Several places cache a `Calendar` (directly or via `DateTimeProcessor`) per-thread in a `static ThreadLocal` for performance, but only partially (or never) re-sync its first-day-of-week setting to the live `week.start` property before computing `WEEK_OF_YEAR`/`WEEK_OF_MONTH` fields. Because thread-pool workers are reused across unrelated requests, a pooled thread keeps whichever first-day-of-week was in effect (or the JVM default) the first time it ever ran the affected code — for its entire lifetime — producing random, persistent, cross-session incorrect week bucketing whenever `week.start` differs from that.

`DateRangeRef.getDateTimeProcessor()` had an additional genuine concurrency bug: it tried to invalidate the calling thread's cached `DateTimeProcessor` by comparing the live `Tool.getFirstDayOfWeek()` against a single **shared static** "last known value" field, while the cache itself is a **per-thread** `ThreadLocal`. Whichever thread noticed a `week.start` change first would update the shared flag, which then made every *other* thread's own comparison falsely pass — permanently masking the change from them. Fixed by tracking staleness per cached instance (`DateTimeProcessor` now exposes the first-day-of-week it was built with) instead of via one JVM-wide flag racing against per-thread state.

- `DateRangeRef` / `DateTimeProcessor`: fixed the shared-static-flag race described above.
- `ValueOfColumn`: DC "previous year/quarter/week" comparison calc never synced its cached calendar at all.
- `SortOrder`: the `WEEK_DATE_GROUP` comparator that decides whether two dates share a week bucket read `WEEK_OF_YEAR` before syncing.
- `CalcDateTime.date()`: the core date-bucketing function synced *after* the bucket index was already computed from the stale value.
- `DateComparisonFormat`: MMWeek-encoded axis label decoding never synced at all.

## Test plan

- [x] Rebuilt `core` and manually reproduced the original bug (rapidly changing `week.start` and reopening/refreshing a Same-Week date comparison chart cycled between a correct rendering and multiple distinct incorrect ones within a handful of reopens).
- [x] Confirmed the fix resolves it under repeated stress-testing (many rapid `week.start` changes + reopens, no incorrect rendering observed).
- [ ] Existing `DCMergeDatePartFilterTest` / `StandardPeriodsTest` suites (not yet re-run in CI as part of this PR).

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/calc/ValueOfColumn.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/calc/ValueOfColumn.java
@@ -621,6 +621,7 @@ public class ValueOfColumn extends AbstractColumn {
    private Date getPreQuarterSameWeek(Date date) {
       long weekIndexOfQuarter = getWeekIndexOfQuarter(date);
       Calendar calendar = jcalendars.get();
+      calendar.setFirstDayOfWeek(Tool.getFirstDayOfWeek());
       calendar.setTime(date);
       calendar.add(Calendar.MONTH, -3);
       Date firstWeekOfQuarter = DateComparisonUtil.getQuarterFirstWeek(calendar.getTime(),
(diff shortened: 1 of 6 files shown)


jshobe-inetsoft (approved):

PR #4927, opened 2026-09-01
feat(wiz): Composer tab-bar agent-connected indicator + sheetLabel on join

fredaan:
## Summary

Companion PR to [stylebi-wiz#2068](https://github.com/inetsoft-technology/stylebi-wiz/pull/2068) — implements Plan C (this repo's half) of a two-part feature: when the composer-chat MCP plugin (Claude Code) is connected to a StyleBI sheet, show it visibly in the Composer.

- **New `SetAgentActiveCommand`** (Java + TS), broadcast over the existing `SheetAgentBroadcastService`/`COMMANDS_TOPIC` channel — no new transport, reuses the pattern `SaveSheetCommand`/`SetWorksheetInfoCommand` already use.
- **`SheetJoinService.join()`** broadcasts "agent active" on every successful join (worksheet and viewsheet).
- **Every session-ending path broadcasts "agent inactive"**: `SheetSessionService`'s `evictExpired`/`socketClosed`/`detach(EditorContext)`, and all four `*AgentController`'s REST `detach()` endpoints — found via code review that these paths previously notified nothing, which would have left the tab-bar icon stuck after any crash/timeout.
- **`SheetOpenService.openBaseWorksheet()`** also broadcasts — a second real entry point that attaches an agent session to a new worksheet runtime, found and fixed during code review (it doesn't route through `SheetJoinService.join()`).
- **`sheetLabel`**: a best-effort human-readable label (`AssetEntry.toView()`) added to the join response, resolved once via a new `SheetRuntimeAccess.getRuntimeSheetDirect()` — reuses the same tolerant runtime lookup already used for the join's ownership check, instead of a second, differently-behaved one.
- **Frontend**: `Sheet.agentConnected`/`agentOwnerIdentity`, pane command handlers (`ws-pane`/`viewsheet-pane`, since an agent can hold one worksheet + one viewsheet session simultaneously), and the tab-selector icon + tooltip.

## Test plan
- [x] `mvnw -pl core test` on the six touched controller/service classes + `SheetOpenServiceTest`: **244/244 passing** (independently re-run twice across two rounds of adversarial code review, not just the implementer's own report)
- [x] Live-verified via the composer-chat plugin against a running StyleBI deployment: icon appears only on the paired tab (confirmed with two tabs of the *same* runtime type, not just worksheet-vs-viewsheet, to prove routing is by runtime id not sheet type), clears on detach, survives 3/3 reconnects without flicker, works for both worksheet and viewsheet, and doesn't interfere with the pre-existing toolbar pairing indicator
- [x] Two rounds of adversarial code review, including a confirmed-and-fixed compile break and the `open_base_worksheet` gap above

Full design, live-verification logs (including a debugging saga whose real cause was environmental, not a defect in this diff — see the linked archive), and both review rounds are archived in the `stylebi-wiz` repo at `docs/teams/2026-09-01-agent-sheet-visibility/`.

**Note on local verification**: this PR's branch was built in a fresh worktree of this checkout to keep it isolated from other in-progress local work; that fresh worktree is missing a locally-built Maven plugin (`antlr2-maven-plugin`) needed for a full `mvn compile`/`test` run there — the 244/244 test result above was captured in a fully-bootstrapped local checkout with the identical code, and CI here should build clean.

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/ws/command/SetAgentActiveCommand.java
+++ b/core/src/main/java/inetsoft/web/composer/ws/command/SetAgentActiveCommand.java
@@ -1,0 +1,42 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.web.composer.ws.command;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import inetsoft.web.viewsheet.command.ViewsheetCommand;
+import org.immutables.value.Value;
+
+import javax.annotation.Nullable;
+
+@Value.Immutable
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonSerialize(as = ImmutableSetAgentActiveCommand.class)
+public abstract class SetAgentActiveCommand implements ViewsheetCommand {
+   public abstract boolean active();
+
+   @Nullable
+   public abstract String ownerIdentity();
+
+   public static Builder builder() {
+      return new Builder();
+   }
+
+   public static class Builder extends ImmutableSetAgentActiveCommand.Builder {
+   }
+}
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
@@ -65,7 +65,7 @@ public class BindingAgentController {
                                  ChartAestheticAgentService aestheticService,
                                  TableBindingService tableService,
                                  CalcTableService calcService,
-                                 SelectionBindingService selectionService)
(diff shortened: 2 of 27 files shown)


fredaan:
## Review round 2 (final) — CLEAN

Two rounds of adversarial code review were done against this diff (as it sat in the local checkout it was developed in) before this PR was opened. Round 1 found and round 2 independently re-verified two real findings:

- **Blocker (fixed): compile break in `SheetJoinService.java`** — a stray dead-code fragment left in `resolveSheetLabel`'s catch block after live JDWP-debugging session inspection. Caught by round 1, fixed, and round 2 independently re-compiled (`./mvnw -q -pl core -am compile -DskipTests`) from a fresh shell to confirm — clean.
- **Important (fixed): `SheetOpenService.openBaseWorksheet()` didn't trigger the new tab-bar-icon broadcast.** This is a second, real entry point (the plugin's `open_base_worksheet` tool) that attaches an agent session to a new worksheet runtime without going through `SheetJoinService.join()` — found by round 1's adversarial pass (explicitly testing "every way an agent session attaches to a new runtime" as an invariant), fixed with the same best-effort `sendAgentActive` pattern used elsewhere, and round 2 independently confirmed the placement, try/catch semantics, and the two new tests in `SheetOpenServiceTest.java` genuinely assert what they claim (not just similarly-named happy-path checks).
- **Regression check**: round 2 independently re-ran the full requested test set (not the builder's reported numbers): `Tests run: 244, Failures: 0, Errors: 0, Skipped: 0`, `BUILD SUCCESS`.
- Three pre-existing nits (raw identity string `admin~;~host-org` in the tooltip rather than a display-friendly name, an unescaped-quote edge case in a free-text summary — that one's actually in the companion stylebi-wiz PR, not here — and the placeholder icon glyph `\2726` pending real design input) are follow-ups, not blockers.

Both non-goals from the original design were explicitly checked and confirmed respected: no change to Follow Focus/pane-scoped `editorContext` matching logic, and the two broadcast channels (`COMMANDS_TOPIC` for the new tab icon vs. `PAIRING_JOINED_TOPIC` for the pre-existing toolbar indicator) were confirmed structurally separate in the diff, not just by non-regression testing.

Full detail archived in the `stylebi-wiz` repo at `docs/teams/2026-09-01-agent-sheet-visibility/06-review-r1.md` and `06-review-r2.md`.

PR #4929, opened 2026-09-01
Add create_viewsheet endpoint, widen attach_base_worksheet to logicalModel/physicalTable

fredaan:
## Summary
- Closes the create half of PVA-007/bug 76332 -- `SheetOpenService.createViewsheet` mints a new viewsheet and pairs it by reusing the acting session's existing browser socket, the same no-new-pairing-code mechanism `openBaseWorksheet` already uses in the reverse direction. Companion plugin PR: https://github.com/inetsoft-technology/stylebi-wiz/pull/2070
- Widens the already-shipped `attach_base_worksheet` (PR #4900) to accept `type: "logicalModel"|"physicalTable"` in addition to the existing worksheet path, via a shared `resolveDataSourceEntry` helper reused by both endpoints.
- Widens `Viewsheet.reloadBaseWorksheet` to populate `getBaseWorksheet()` for a direct-source (logical-model/physical-table) base entry, not just a worksheet one -- it previously silently no-op'd on those two types.
- `physicalTable` resolves by recursively BROWSING the datasource's tree via `AssetRepository.getEntries()` (tables may nest under catalog/schema folders of arbitrary depth), never by hand-constructing a `PHYSICAL_TABLE` entry's `SourceInfo` properties -- those are populated by the repository's own JDBC-metadata walk and a hand-built entry with the wrong properties would look accepted and silently render nothing.
- `query` (`AssetEntry.Type.QUERY`) intentionally not supported -- a deprecated source type.
- `create_viewsheet` requires an already-paired session (worksheet or viewsheet); with no explicit data source it defaults to the acting session's own worksheet (refusing loudly if that worksheet has never been saved).

## Test plan
- [x] `ViewsheetTest` (new): `reloadBaseWorksheet` cases for worksheet/logicalModel/physicalTable
- [x] `ViewsheetAssemblyAgentControllerTest`: widened `attach_base_worksheet` cases (logicalModel, physicalTable flat + nested-under-schema-folder, missing-field refusals, unknown type, type-omitted backward compat) + new `create_viewsheet` cases
- [x] `SheetOpenServiceTest`: new `createViewsheet` cases (defaults to acting worksheet, explicit data source from either acting-session type, unresolvable session, no socket, unsaved worksheet, permission refusal)
- [x] Full `inetsoft.web.wiz` + `inetsoft.uql.viewsheet` suites: 2999 tests, 0 failures, 0 errors, 1 skipped

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/Viewsheet.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/Viewsheet.java
@@ -206,11 +206,21 @@ public class Viewsheet extends AbstractSheet implements VSAssembly, VariableProv
     * Lighter than update() — skips clearCache() and embedded viewsheet processing.
     */
    public void reloadBaseWorksheet(AssetRepository rep, Principal user) throws Exception {
-      if(wentry == null || !wentry.isWorksheet()) {
+      if(wentry == null) {
          return;
       }
 
-      Worksheet ws = (Worksheet) rep.getSheet(wentry, user, true, AssetContent.ALL);
+      Worksheet ws;
+
+      if(wentry.isWorksheet()) {
+         ws = (Worksheet) rep.getSheet(wentry, user, true, AssetContent.ALL);
+      }
+      else if(isDirectSource()) {
+         ws = getWorksheet(rep, wentry, user);
+      }
+      else {
+         return;
+      }
 
       if(ws != null) {
          ws.getWorksheetInfo().setDesignMaxRows(vinfo.getDesignMaxRows());
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java
@@ -229,6 +229,19 @@ public class SheetOpenService {
             "pairing code and run connect_sheet again.");
       }
 
+      // Same class of bug openBaseWorksheet guards against (see its own guard and comment
+      // above): a pane-scoped session is a write handle for ONE script location, and minting a
+      // new whole-sheet (editorContext = null) session from it -- as the code below does --
+      // would launder that narrow grant into unscoped whole-sheet authority on a runtime the
+      // pane's grant never named. Must come before the runtime is touched or any grant is minted.
+      if(actingSession.editorContext() != null) {
+         throw new IllegalArgumentException(
+            "This session is scoped to one script location, not to the whole sheet, so it cannot " +
+            "create a new viewsheet: doing so would turn a single-expression grant into a whole-sheet " +
+            "write handle. Ask the user to re-pair from the sheet toolbar ('Connect to Claude') and " +
+            "call create_viewsheet from that session.");
+      }
+
       if(actingSession.socketSessionId() == null) {
          throw new IllegalArgumentException(
             "The connected session has no active browser connection to create a viewsheet in; " +
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java
@@ -17,16 +17,21 @@
  */
 package inetsoft.web.wiz.viewsheet;
 
+import inetsoft.analytic.composition.ViewsheetService;
+import inetsoft.report.composition.RuntimeSheet;
 import inetsoft.report.composition.RuntimeViewsheet;
 import inetsoft.report.composition.WorksheetService;
+import inetsoft.sree.security.IdentityID;
 import inetsoft.sree.security.ResourceAction;
 import inetsoft.sree.security.ResourceType;
 import inetsoft.sree.security.SecurityProvider;
+import inetsoft.uql.XPrincipal;
 import inetsoft.uql.asset.AssetEntry;
 import inetsoft.uql.viewsheet.Viewsheet;
 import inetsoft.web.composer.command.OpenComposerAssetCommand;
 import inetsoft.web.wiz.pairing.JoinSession;
 import inetsoft.web.wiz.pairing.SheetAgentBroadcastService;
+import inetsoft.web.wiz.pairing.SheetRuntimeAccess;
 import inetsoft.web.wiz.pairing.SheetSessionService;
 import inetsoft.web.wiz.pairing.SheetType;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -56,13 +61,17 @@ public class SheetOpenService {
                             SheetSessionService sheetSessions,
                             WorksheetService worksheetService,
                             SecurityProvider securityProvider,
-                            SheetAgentBroadcastService broadcast)
+                            SheetAgentBroadcastService broadcast,
+                            ViewsheetService viewsheetService,
+                            SheetRuntimeAccess runtimeAccess)
    {
       this.viewsheetSessions = viewsheetSessions;
       this.sheetSessions = sheetSessions;
       this.worksheetService = worksheetService;
       this.securityProvider = securityProvider;
       this.broadcast = broadcast;
+      this.viewsheetService = viewsheetService;
+      this.runtimeAccess = runtimeAccess;
    }
 
    /**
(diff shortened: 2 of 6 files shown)


fredaan:
## [CHANGES REQUESTED] Round 1 review (independent confirmation)

*(Posted as a plain comment, not a formal review -- `gh pr review -r` was rejected: "Can not request changes on your own pull request".)*

**Head commit reviewed:** `4b29ab64034b5e044368a9acaa567653c8b8b913` (branch `feat/create-viewsheet-and-source-types`)

**What I checked:**
- Read `SheetOpenService.createViewsheet` (core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java:221-284) and `SheetSessionService.resolve` (core/src/main/java/inetsoft/web/wiz/pairing/SheetSessionService.java:116-130) directly to confirm the finding independently rather than taking the other review's word for it.
- Re-checked the `AssetRepository`-before-validation pattern in `resolveDataSourceEntry` (ViewsheetAssemblyAgentController.java:318-422) for the same class of ordering bug seen elsewhere in this codebase.
- Checked recursion termination in `findPhysicalTableEntry` (ViewsheetAssemblyAgentController.java:459-493) for cyclic/malformed-tree edge cases.

**Confirmed finding (CRITICAL, blocking):**

`SheetOpenService.createViewsheet` (SheetOpenService.java:224) resolves the acting session via the generic, type-agnostic `sheetSessions.resolve(fromSessionToken, agentKey(user))`, but never checks `actingSession.editorContext() != null` before minting the new session. `SheetSessionService.resolve` (SheetSessionService.java:116-130) does not filter on `editorContext` at all -- it only checks token presence, expiry, and owner-identity match -- so a pane-scoped session (a write handle scoped to one script location) resolves successfully here.

The new viewsheet session is then opened with `editorContext = null` (SheetOpenService.java:270-273), i.e. whole-sheet scope, on a brand-new runtime the pane-scoped grant never named.

This is the exact vulnerability class `openBaseWorksheet` was hardened against (see its own guard and extensive comment at SheetOpenService.java:105-123: "A pane-scoped session is a write handle for ONE script location... LAUNDERED the narrow grant into an unscoped one"). `createViewsheet` mirrors `openBaseWorksheet`'s structure (acknowledged in its own javadoc at line 206: "the exact mechanism openBaseWorksheet already uses") but dropped that one guard, reopening the hole in the reverse direction.

**Confirmed missing regression coverage:** `SheetOpenServiceTest` has a dedicated test for `openBaseWorksheet`'s guard (`refusesAPaneScopedSessionRatherThanMintingAWholeSheetOne`, ~line 657), but the `createViewsheet` test section (lines 726-899) has no equivalent -- every `createViewsheetService(...)` harness call constructs the acting session with `editorContext = null` (line ~854), so a pane-scoped acting session is never exercised.

**Suggested fix:** add the same guard `createViewsheet` currently lacks -- refuse when `actingSession.editorContext() != null`, before `viewsheetService.openTemporaryViewsheet(...)` is called -- plus a regression test mirroring `refusesAPaneScopedSessionRatherThanMintingAWholeSheetOne` for the create-viewsheet path.

**Other items checked, no issues found:**
- The `AssetRepository`-before-validation ordering in `resolveDataSourceEntry` is correct: `repSupplier.get()` is deferred until after all field-presence checks, confirmed by both an explicit comment (lines 346-352) and a test (`createViewsheetRefusesPhysicalTableMissingTableBeforeCallingSheetOpenService`) asserting the repository is never touched on a plain missing-field refusal.
- `findPhysicalTableEntry`'s recursion is bounded by `depth > 10`, incremented on every recursive call, so a structurally malformed/cyclic tree cannot cause unbounded recursion.

**Verdict: CHANGES REQUESTED** -- the missing pane-scope guard in `createViewsheet` is a real, confirmed authority-laundering bug and must be fixed before merge.

fredaan:
Fixed in 16a2e4155: added the missing editorContext pane-scope guard to createViewsheet, mirroring openBaseWorksheet's existing one, plus a regression test (createViewsheetRefusesAPaneScopedActingSessionRatherThanMintingAWholeSheetOne). Grepped SheetOpenService.java and ViewsheetAssemblyAgentController.java for other generic-resolve-then-mint-session call sites and found no other instance of the gap -- every other caller either goes through ViewsheetSessionService.requireSession (which already enforces the pane-scope guard) or does not mint a new session at all.

fredaan:
## [APPROVED] Round 2 review (SHA `16a2e4155`)

**What I checked:**
- `git diff 4b29ab640..16a2e4155 -- core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java core/src/test/java/inetsoft/web/wiz/viewsheet/SheetOpenServiceTest.java` — the only change since round 1 is the guard + its test.
- Guard placement: `actingSession.editorContext() != null` now sits immediately after the null-check on `actingSession` and before the socket check, the `dataSource` defaulting (which itself calls `runtimeAccess.getSheetForPairing`), the permission check, and `viewsheetService.openTemporaryViewsheet(...)`. This mirrors `openBaseWorksheet`'s guard placement exactly (first thing after resolving the session, before any read or write touches the runtime). Message names the problem and the recovery step, consistent with this file's convention.
- **Round-1 finding status: fixed.** Confirmed by direct code reading, not just trusting the report.

**Empirically verified the regression test, not just read it:** I checked out the pre-fix parent commit (`4b29ab640`) into a worktree, dropped in the new `SheetOpenServiceTest.java` from `16a2e4155` on top of the *old* production code, and ran the new test method alone. It fails pre-fix as expected — but worth noting precisely *how*: it fails only at the first `assertTrue` (message contains "scoped to one script location"), because with `actingWsEntry=null` and `dataSource=null`, the pre-fix code's own `dataSource` defaulting block throws its own "connected worksheet has not been saved yet" `IllegalArgumentException` before ever reaching the permission check or the three mint calls. So in this specific test configuration, the three `never()` assertions the test's docstring calls "the substance" would actually pass under the old, unguarded code too — not because the guard stopped anything, but because an unrelated null-dataSource check happened to exit first. The message-content assertion is what's actually doing the regression-catching work here. A tighter version of this test would pass a non-null `dataSource` argument directly (or a valid saved `actingWsEntry`), so that, absent the guard, execution would genuinely reach and execute `openTemporaryViewsheet`/`sheetSessions.open`/`broadcast.sendToComposer`, making the `never()` assertions non-vacuous. This doesn't change my verdict — the production guard's correctness is established independently by reading the source (it structurally precedes all three side-effecting calls regardless of test data) — but flagging it since it was explicitly asked to be checked.

**Verified the "no other instances" claim myself, not just trusted the report:**
- Grepped the entire `core/src/main/java` tree for `SheetSessionService.open(` call sites: exactly 3 total — `SheetOpenService.java:184` (openBaseWorksheet, already guarded) and `:283` (createViewsheet, now guarded), and `SheetJoinService.java:130-132`.
- Read `SheetJoinService.join` directly: it opens a session with `grant.editorContext()` — i.e. it *preserves* the pairing grant's own scope rather than deriving a new (wider) scope from an existing session's authority. Different pattern, confirmed correct.
- Read `ViewsheetAssemblyAgentController.detach` directly: it only resolves and closes the caller's own session (`sessionService.close(sessionToken)`); it never mints a new session. Confirmed correct, not exposed to this bug class.
- Confirmed `ViewsheetSessionService.resolve`/`requireSession` route through `PaneScopeService.requireWholeSheetSession`, which already refuses pane-scoped sessions.

**Ran the actual Maven suite myself** (not just trusting the fixer's self-report): `./mvnw test -pl core -am -Dtest=SheetOpenServiceTest,ViewsheetAssemblyAgentControllerTest,ViewsheetTest -Dsurefire.failIfNoSpecifiedTests=false` against `16a2e4155` — surefire reports confirm:
- `SheetOpenServiceTest`: 23 run, 0 failures, 0 errors
- `ViewsheetAssemblyAgentControllerTest`: 53 run, 0 failures, 0 errors
- `ViewsheetTest`: 3 run, 0 failures, 0 errors

**Verdict: APPROVED.** The round-1 authority-laundering finding is genuinely fixed, correctly placed, and covered by a passing (if slightly weaker-than-advertised) regression test. No new issues found in the diff. Non-blocking suggestion: strengthen `createViewsheetRefusesAPaneScopedActingSessionRatherThanMintingAWholeSheetOne` to pass a resolvable `dataSource` so its `never()` assertions are exercised against the actual vulnerable path rather than short-circuited by the unrelated unsaved-worksheet check.

fredaan:
CI's \`build\` check failed after round-2 approval — real cause, not flaky CI: the base branch \`stylebi-wiz-main-rebase\` advanced (PR #4927, merged after this branch was cut) added a \`sheetLabel\` field to \`JoinResponse\` and a \`sendAgentActive\` tab-bar notification on every session-attaching entry point. \`createViewsheet\`'s own \`JoinResponse\` construction predated that and was still 5-arg.

Fixed in 6358cab52 (merged latest base + adjusted): \`sheetLabel: null\` (same choice \`openBaseWorksheet\` makes — a fresh, untitled viewsheet has no resolvable label yet), plus added the \`sendAgentActive\` call \`createViewsheet\` was otherwise missing (a real, if minor, consistency gap — it's a third session-attaching entry point the concurrent feature didn't know about yet), with 2 new regression tests mirroring \`openBaseWorksheet\`'s own. Full \`SheetOpenServiceTest\`/\`ViewsheetAssemblyAgentControllerTest\`/\`ViewsheetTest\`: 85/85 clean. Requesting a fresh round on this delta before merge.

fredaan:
## Round 3 review — SHA 6358cab52

**[APPROVED]**

Reviewed the delta relative to 16a2e4155 (round-2-approved): the merge commit a80ef15fa8 (`git merge origin/stylebi-wiz-main-rebase`, bringing in #4927) plus the follow-up fix commit 6358cab52 on top of it.

### 1. `sheetLabel: null` in `createViewsheet`'s `JoinResponse`
Correct, and actually the *more* correct choice than `dataSource.toView()`. Traced `SheetJoinService.join()`'s resolution mechanism: it resolves `sheetLabel` from the **joined runtime's own** `AssetEntry.toView()` (via the newly-added `getRuntimeSheetDirect`) — i.e., the identity of the sheet being attached to, not some other related asset. In `createViewsheet`, `dataSource` is the *base worksheet/logical-model/table* the new viewsheet is bound to, not the new viewsheet's own identity — using `dataSource.toView()` would mislabel the new viewsheet's tab with its data source's name (and `dataSource` isn't even always present in the request; it's resolved from the acting session in the default-worksheet case). So `dataSource.toView()` would be actively wrong, not just unresolved.

Could `createViewsheet` resolve its *own* new entry's `toView()` (same mechanism as `join()`)? Yes, mechanically — `openTemporaryViewsheet` gives it a `TEMPORARY_SCOPE` entry via `SheetLibraryEngine.getTemporaryAssetEntry`, which does produce a non-null placeholder (`"Untitled-N"`). But `null` here is consistent with `openBaseWorksheet`'s existing, already-reviewed choice (also `null`, per the JoinResponse javadoc: "for `openBaseWorksheet`, not resolved at all"), and this commit's stated purpose was the compile fix + notification parity, not a scope expansion to add label-resolution to a second/third entry point. `null` is honest (the contract explicitly documents it as valid) and doesn't misrepresent anything. Not a blocker; a follow-up could uniformly resolve labels for `openBaseWorksheet`/`createViewsheet` too, but that's a separate, optional enhancement.

### 2. "Third real entry point" characterization
Verified directly: `grep`'d every `sheetSessions.open(`/session-attach call site under `web/wiz/` — exactly three exist: `SheetJoinService.join`, `SheetOpenService.openBaseWorksheet`, `SheetOpenService.createViewsheet`. Re-read #4927's own commit message (4a3ba12d6): it explicitly names only `join` and `openBaseWorksheet` ("a second real entry point... found in code review") — `createViewsheet` isn't mentioned at all, not because it was deliberately excluded, but because it doesn't exist on `stylebi-wiz-main-rebase` (the branch #4927 was authored against) — it's this PR's own addition. So this isn't a scope decision being second-guessed; it's a genuine gap from two features landing concurrently on divergent branches, correctly identified and closed.

### 3. Merge cleanliness
Diffed the merged `SheetOpenService.java`/`ViewsheetAssemblyAgentController.java` against both parents (4a3ba12d6 and 16a2e4155) directly. The merge is clean: `openBaseWorksheet` correctly picked up #4927's `sendAgentActive` call and constructor/field additions (`ViewsheetService`, `SheetRuntimeAccess`); `createViewsheet` (branch-only method, doesn't exist on the other side) passed through untouched, still lacking both the 6th `JoinResponse` arg and the broadcast call — exactly the two things 6358cab52 then adds. No duplicated, dropped, or reordered logic found in either file. The compile break itself is unsurprising for a merge with no textual overlap (record signature widened on one side, a distant call site left un-updated on the other) — not a case of "auto-merge silently produced something wrong," just an expected consequence of two branches adding unrelated things to the same class.

### 4. Fix commit itself
- `ViewsheetAssemblyAgentController.createViewsheet`: 6-arg `JoinResponse` construction with `sheetLabel: null` and an explanatory comment — correct per #1 above.
- `SheetOpenService.createViewsheet`: `broadcast.sendAgentActive(vsSession)` added in its own independent try/catch with a `LOG.warn` fallback, placed at the same relative point (right after the new session is minted, before the `OpenComposerAssetCommand`) as `openBaseWorksheet`'s equivalent call — good parity, good placement.
- Tests: independently ran `SheetOpenServiceTest`/`ViewsheetAssemblyAgentControllerTest`/`ViewsheetTest` from this exact SHA — **27/55/3 passing, 85 total, 0 failures**, confirming the reported numbers. The two new tests (`notifiesTheTabBarThatAnAgentIsNowAttachedToTheNewViewsheet`, `tabBarNotifyFailureDoesNotFailTheCreate`) and the added `verify(broadcast, never())` on the pane-scope regression test correctly mirror the existing `openBaseWorksheet` coverage.

### 5. CI
`gh pr checks 4929`: Semgrep OSS/Scan pass, CLA pass, `claude-review` pass. `build` was still pending at last check — flagging back to the lead to confirm green before merge; everything else is clear.

Round: 3
SHA reviewed: 6358cab52

PR #4930, opened 2026-09-01
Bug #76370, draw the composer object highlight with outline so it does not inflate the box

ngovus-inetsoft:
## Problem

Every object in the Visual Composer renders ~4px taller than the assembly's declared `objectFormat.height`. It is not tied to any widget type — reproduced on VSText and VSSpinner alike.

`.object` (`editable-object-container.component.html`) carries a permanently-reserved `border: 2px solid transparent` (`.bd-highlight-med-filler`, `_themeable.scss`) so that toggling the real drag/drop highlight (`.bd-highlight-med`, also 2px) causes no layout shift. `box-sizing` is `border-box`, but the widget's own content already renders at the full declared height independently of `.object`'s `min-height`, so the border cannot compress into that space — the box grows by 2px top + 2px bottom.

## Fix

Draw the ring with `outline` instead of `border`. `outline` never participates in box-sizing, so it cannot inflate the box.

`outline-offset` is left at its default `0` — **not** `-2px`. Today the box sits at `(left-2, top-2)` sized `H+4` with the ring in the outer 2px band, so the ring paints from `top-2` to `top`. After the change the box sits at `(left, top)` sized `H`, and a `2px` outline at offset `0` paints from `top-2` to `top` — the same pixels. `-2px` would draw the ring *inside* the box, overlapping the widget's own edge.

Three further borders on the same element had the identical effect and would have left the symptom in place whenever an object was selected, so they are converted too:

- `.txt-primary .object` — the dotted selected-object ring
- `.drop-target:not(.object-editor)`
- `.search-result`

With no border in the layout, the compensations that existed only to cancel it are removed:

- `VSPane.selectionBorderOffset` `2` → `0`. It was subtracted from `left`/`top` and inverse-applied to drop coordinates; leaving it would shift every composer object 2px up and left.
- the `width - 2` / `height + 1` fudge for selected selection-container children
- the `border-top` / `border-bottom` zeroing for unselected ones

Net geometry is pixel-identical to before; only the box's own measured size changes, from `H+4` to `H`.

## Scope

`.bd-highlight-med` and `.bd-highlight-med-filler` are used only by `editable-object-container.component.html` (verified across portal/em/shared/elements). The `bt/bb/br/bl-highlight-filler` family in the same block is left alone — it is used by portal nav, composer sidebar tabs, binding tabs, viewer page tabs, the repository tree and VPM tabs.

Two deliberate exclusions:

- `.bt-highlight-med` / `.bb-highlight-med` (the tab/container drag indicator) stay borders. `outline` has no per-edge form, and the indicator is transient — it appears only during a drag, where a momentary 2px growth is not visible.
- `layout-object.component.scss` declares the same reservation inline for the print/device layout pane. Separate component, separate coordinate math, out of scope here; verified unaffected.

## Testing

- `editable-object-container` tl specs — 91 passed
- `npm run test:portal` — 371 passed across 104 files
- `ng lint` — 0 errors
- Live composer pass against a locally-run server: `.object` measures exactly `objectFormat.height` in default, selected and drag-over states; highlight ring, selection ring, resize handles and object positions unchanged; the 4px gap is gone.

Filed separately from the spinner stepper-buttons work, since this is shared composer chrome unrelated to that change.


jshobe-inetsoft (approved):

PR #4931, opened 2026-09-01
Bug #76386, keep the column-values-as-colors option when fixing visual frames

ngovus-inetsoft:
## Problem

With `custom.chart.frames=ColorValueColorFrame` set in sree properties, in the composer:

1. Edit a chart's color aesthetic → Edit Color
2. **Uncheck** "Use Column Values as Colors", Apply → works
3. **Re-check** it, Apply → **not applied**; the chart keeps the categorical palette and the checkbox comes back unchecked

## Root cause

`GraphUtil.fixVisualFrame()` normalizes an aesthetic ref's frame against the bound column. For a color aesthetic it accepts only `CategoricalColorFrame` (dimension ref) or `LinearColorFrame` (measure ref); anything else is replaced, and `ref.setVisualFrame(...)` → `VisualFrameWrapper.wrap()` builds a **brand-new** `CategoricalColorFrameWrapper` whose `colorValueFrame` defaults to `false`.

`ColorValueColorFrame` extends `ColorFrame` and implements `CategoricalFrame`, but is not a `CategoricalColorFrame`, so it always tripped that replacement — and it happened on the very same request that set the flag. `ChartInfoModelBuilder.updateChartInfo()`:

```java
aesService.updateVisualFrames(model, ocinfo, ncinfo);   // sets colorValueFrame = true
GraphUtil.fixVisualFrames(ncinfo);                      // ...and this discards it
```

That is exactly the reported asymmetry:

| Action | Frame from `CategoricalColorFrameWrapper.getVisualFrame()` | `fixVisualFrame` | Result |
|---|---|---|---|
| Uncheck | `CategoricalColorFrame` | no-op | applies |
| Re-check | `ColorValueColorFrame` | wrapper replaced, flag lost | silently reverts |

It also explains why a saved `.vso` renders correctly on open: `colorValueFrame` round-trips through XML, and `fixVisualFrames` is not on the plain-load path — only on the edit paths.

## Fix

A guard at the top of the `AESTHETIC_COLOR` branch that leaves an explicitly chosen `ColorValueColorFrame` alone, for both dimension and measure refs. This also stops the flag being dropped by the other `fixVisualFrames` callers (`ChangeChartRefService`, `ChartPropertyDialogService`, `DateComparisonDialogService`, `ChartVSAssembly`).

Returning `false` is the "frame unchanged" signal that `fixVisualFrame`/`fixVisualFrames` report to callers, and it keeps the guard clear of the `SHAPE`/`SIZE`/`TEXT` branches.

## Testing

New `GraphUtilFixVisualFrameTest` (4 cases): the option survives for a dimension ref and for a measure ref; an invalid color frame (e.g. a size frame on the color region) is still normalized to `CategoricalColorFrame`; a plain `CategoricalColorFrameWrapper` is untouched.

Confirmed to be a real regression test — with the guard reverted, both preservation cases fail (`expected: <false> but was: <true>`) while the two normalization cases still pass.

- Full `community/core` suite: 4667 run, 0 failures, 67 skipped
- `categorical-color-pane` TL specs: 17/17 (no TS changed; sanity check)
- Manual: reproduced the original failure, then verified off→on now repaints from the column hex values, stays checked on re-open, and survives save/reload. Regression-checked dragging a different dimension/measure onto Color and switching chart types.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/GraphUtil.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/GraphUtil.java
@@ -900,6 +900,13 @@ public class GraphUtil {
       DataRef dref = ref.getDataRef();
 
       if(type == ChartConstants.AESTHETIC_COLOR) {
+         // the user explicitly opted into using the column values as colors. this is a
+         // valid ColorFrame for both dimension and measure refs, so don't normalize it
+         // away, otherwise the option is discarded as soon as it's applied.
+         if(frame instanceof ColorValueColorFrame) {
+            return false;
+         }
+
          if(isCategorical(dref)) {
             if(!(frame instanceof CategoricalFrame)) {
                String col = dref instanceof VSDimensionRef ?
--- a/core/src/test/java/inetsoft/report/composition/graph/GraphUtilFixVisualFrameTest.java
+++ b/core/src/test/java/inetsoft/report/composition/graph/GraphUtilFixVisualFrameTest.java
@@ -1,0 +1,128 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (approved):

PR #4933, opened 2026-09-01
Bug #76393, fence PortalThemesManager reloads on content, not on listener de-registration

ngovus-inetsoft:
## Problem

`PortalThemesManager.save()` unregisters its `DataChangeListener` before writing `portalthemes.xml` and re-adds it in a `finally`, to stop its own write re-triggering its own reload.

That cannot work. `BlobStorage` submits every storage event to a single-threaded `BlobStorageEvent` executor (`BlobStorage.java:667`), with a second async hop in `LocalKeyValueStorage` — so the notification for save *N* routinely arrives **after** the `finally` has re-registered the listener.

Landing there, the listener's `reset()` + `loadThemes()` reloads the last-saved XML over whatever changed in memory since: `parseXML()` clears and repopulates `cssEntries` / `logoEntries` / `faviconEntries` / `welcomePageEntries` wholesale. A not-yet-persisted `add*Entry()` is **dropped**; a not-yet-persisted `remove*Entry()` is **resurrected**. Nothing is logged — a stale reload is a *successful* reset+reload, not an exception.

`AbstractEditableAuthenticationProvider.copyOrganizationInternal()` is the production trigger: four separate mutate-then-`save()` rounds per org copy, so four windows for a late self-notification to land mid-sequence. On a real multi-tenant install this silently loses a customer's logo, favicon, CSS or welcome-page branding on an org copy or rename.

## Fix

**`PortalThemesManager` — fence on content.** `saveUnderLock()` builds the document in memory, writes those exact bytes, and records their SHA-256 as `syncedDigest` while the distributed save lock is still held and before the listener is re-added. `changeListener` skips the reload when the file still digests to that value.

Content rather than the event timestamp, deliberately. `DataChangeEvent.getEventTime()` is the blob mtime stamped by *whichever node wrote it*, so with a few milliseconds of clock skew between pods a remote write can carry an **older** timestamp than this node's own last write. An `eventTime <= lastSaved` fence would silently discard that node's change and overwrite it on the next save — the same silent-branding-loss symptom this PR removes, just with a rarer trigger. A content digest is immune to both clock skew and same-millisecond collisions, and an unreadable file yields a `null` digest that never matches, so any failure falls back to the old conservative always-reload behaviour.

**`copyOrganizationInternal()` — one write per org copy.** The four branding changes are batched behind a single `save()`, cutting the path from four listener windows and four XML writes to one, and removing a partial-persistence window (CSS was already on disk when a later favicon copy threw and aborted the whole operation). The `brandingChanged` flag reproduces the original save/no-save decision exactly, including the all-absent case.

## Tests

Three new deterministic cases in `PortalThemesManagerTest`, driving the captured listener with a synthetic `DataChangeEvent` against a round-tripping fake data space instead of racing a real notification:

- `changeListener_notificationForOwnWrite_doesNotClobberUnsavedEntry`
- `changeListener_notificationForRemoteWrite_reloads`
- `changeListener_remoteWriteWithOlderTimestamp_stillReloads` — pins the clock-skew case above

Verified non-vacuous: stubbing the fence to `if(false)` fails the first with `expected: <portal/orgA/logo.png> but was: <null>`, the reported drop. The two remote-write tests pass without the fence by design — they guard the opposite direction, over-suppression.

`OrgLifecyclePortalBrandingDataSpaceIntegrationTest`, merged `@Disabled` against this issue in #4918, is re-enabled.

**91 tests, 0 failures, 0 skipped** across `PortalThemesManagerTest`, `OrgLifecyclePortalBrandingTest`, `OrgLifecyclePortalBrandingDataSpaceIntegrationTest`, `OrgLifecycleThemeOrchestrationTest`, `IdentityServiceAutoSaveOrgLifecycleTest`, `AbstractEditableAuthenticationProviderStaticDepTest`, `CustomThemesManagerTest`, `OrgLifecycleDataSpaceIntegrationTest`, `OrgLifecycleScopedPropertiesIntegrationTest`, `PermissionMatrixOrgLifecycleTest`.

### Reviewer note on the re-enabled integration test

The intermittent failure reproduced ~1 in 3-4 runs for the reporter. It does **not** reproduce on my machine — 15/15 pass both with and without the fix. DEBUG logging confirms the listener does fire on the `BlobStorageEvent` thread mid-run here, but delivery consistently lands *inside* the remove/re-add window where the old suppression happens to work. So a green local run of that test is not evidence either way; the deterministic coverage is the unit tests. If you would prefer to keep it `@Disabled` until someone re-verifies on a machine where it reproduced, say so and I will restore the annotation.

## Out of scope (pre-existing, unchanged)

- `reset()` does not clear the four `*Entries` maps, so a genuine *remote* change that removes an entry leaves it stale in memory.
- `DashboardRegistry.save()` and `RepletRegistry.save()` use the identical unsuppressable remove/re-add pattern.
- `PortalThemesManager`'s non-map state (`portalTabs`, `userFontFaces`) is mutated from the `BlobStorageEvent` thread while readers call the getters.

### DIFF
--- a/core/src/main/java/inetsoft/sree/portal/PortalThemesManager.java
+++ b/core/src/main/java/inetsoft/sree/portal/PortalThemesManager.java
@@ -1092,7 +1092,11 @@ public class PortalThemesManager implements XMLSerializable, AutoCloseable {
          return in == null ? null : digest(in.readAllBytes());
       }
       catch(IOException ex) {
-         LOG.debug("Failed to read portal themes file for comparison: " + name, ex);
+         // a missing file is not this branch (getInputStream returns null for that), so
+         // this is a real read failure: the self-write fence is disabled for this
+         // notification and the reload goes ahead
+         LOG.warn("Failed to read portal themes file for self-write detection, " +
+                     "will reload: " + name, ex);
          return null;
       }
    }
--- a/core/src/main/java/inetsoft/sree/portal/PortalThemesManager.java
+++ b/core/src/main/java/inetsoft/sree/portal/PortalThemesManager.java
@@ -35,6 +35,8 @@ import javax.xml.xpath.*;
 import java.awt.*;
 import java.io.*;
 import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.List;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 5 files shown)


jshobe-inetsoft (approved):

PR #4935, opened 2026-09-01
Bug #76390 and #76391, don't drop or blank real historical data in Date Comparison

jkidacki-inetsoft:
## Summary

Both bugs trace to the same root heuristic in `DateComparisonUtil.computeValidParts()` and are consolidated into one PR since #76390's fix builds directly on #76391's.

### Bug #76391 — Year_MonthToDate Chart4: WeekOfMonth rows missing

`computeValidParts()` used "does the most-recent comparison period's own data have a row for this part" as a proxy for "has the most-recent period chronologically reached this part yet." Those conditions only coincide when the most recent period's data is dense. When it's merely sparse — a real bucket with zero matching rows — the old algorithm dropped that bucket's data for *every* period, including older periods with real, legitimate rows for it.

**Fix:** keep a part valid whenever it sorts at or before the last part the most recent period's own rows actually reach, instead of requiring an exact row match. A part that sorts strictly beyond that point is still treated as an unreached "future" bucket and stays excluded, preserving the behavior Bug #75152/#76389 rely on.

### Bug #76390 — Shared dc1 Chart4: per-bar date labels blank after the first facet group

`DateComparisonFormat.initPartDate()` independently re-ran the same orphaned-cell heuristic (via `computeValidParts()`) purely to decide which facet groups' date labels to blank — but any row that reaches `DateComparisonFormat` has already passed the legitimate future-bucket filter upstream in `applyDateRange()`'s `ValidPartsSelector`. Re-filtering at the label stage was pure redundant/harmful duplication specific to facet mode, blanking real labels for every facet group after the first.

**Fix:** remove the redundant `orphanedCells` label-blanking logic from `DateComparisonFormat` entirely.

## Test plan

- [x] New tests: `DateComparisonUtilValidPartsTest` (reproduces #76391's exact drop pattern; guards the legitimate future-bucket-hiding case), `DateComparisonFormatOrphanedFacetTest` (reproduces #76390's exact label-blanking pattern)
- [x] Existing `DateComparisonIntervalConditionTest`, `DcFormatTableLensTest`, `VSChartShowDataServiceTest`, `DCMergeDateFilterTest` still pass
- [x] #76391 independently re-verified by a second reviewer pass (re-ran diff + all tests, probed edge cases)
- [x] #76390 verified live against the actual imported `Shared dc1.vso` on a running server (confirmed the reported bug reproduces pre-fix; test-verified post-fix)

Fixes Redmine #76391. Fixes Redmine #76390.

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/DateComparisonFormat.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/DateComparisonFormat.java
@@ -150,27 +150,32 @@ public class DateComparisonFormat extends Format {
             }
          }
 
-         // Remove orphaned cells (no current-year data) using the same heuristic as
-         // DateComparisonUtil.computeValidParts(). An empty validParts set means either
-         // no data, MergePartCell period without a date value, or per-part real-date
-         // layout — in all those cases no filtering is applied.
-         Set<Object> orphanedCells = new HashSet<>();
-         Set<Object> validParts = DateComparisonUtil.computeValidParts(data, dateCol, partCol, null);
-
-         if(!validParts.isEmpty()) {
-            Iterator<Map.Entry<Object, Set<Date>>> it = partDates.entrySet().iterator();
-
-            while(it.hasNext()) {
-               Map.Entry<Object, Set<Date>> entry = it.next();
-
-               if(!validParts.contains(entry.getKey())) {
-                  orphanedCells.add(entry.getKey());
-                  it.remove();
-               }
-            }
-         }
-
-         this.orphanedCells = orphanedCells;
+         // NOTE (Bug #76390): this used to remove "orphaned" part entries here --
+         // i.e. any partCol value not in DateComparisonUtil.computeValidParts(data,
+         // dateCol, partCol, null) -- and blank that part's axis label entirely,
+         // using the same "does this part have a row in the single most recent
+         // period" heuristic as DateComparisonUtil.applyDateRange()'s
+         // ValidPartsSelector. That heuristic is sound for ValidPartsSelector, which
+         // filters raw, not-yet-verified rows out of the plotted/exported dataset
+         // before anything here runs.
+         //
+         // But every entry partDates can ever contain is, by construction (see the
+         // loop above), already backed by a real row with a real plotted value --
+         // there is no "spurious future" data to discover and strip at this point.
+         // In a faceted date-comparison chart, one partCol value legitimately *is*
+         // one facet group, and different facet groups routinely finish carrying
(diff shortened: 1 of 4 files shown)


jkidacki-inetsoft:
Pushed a follow-up commit addressing a residual gap found during manual live verification against the actual reported fixture, plus both inline review notes above.

## Residual gap in #76391's original fix

Live-testing against the real `Year_MonthToDate.vso` Chart4 (the exact scenario from Bug #76391) showed two of the three originally-reported rows now render (7-1, 9-2), but **WeekOfMonth "12-2" (2019: 18) still didn't**.

Root cause: the real "part" values are `DCMergeDatePartFilter.MergePartCell` instances — a two-element tuple (month component + week-of-month component; "12-1" = `[12, 1]`), not the plain `month*10+week` integers the original fix's test approximated. 2020 (the most recent period) reaches month 12 (has a "12-1" row) but its calendar never produces a "12-2" split week, so `maxPart` is "12-1" — and the existing sort-based rescue (`part <= maxPart`) still excluded "12-2" since its tuple sorts after "12-1", even though 2020 plainly reached month 12.

**Fix:** additionally treat a `MergePartCell` as valid if it shares `maxPart`'s leading (non-tie-breaking) components, regardless of its own trailing value. Every family strictly before `maxPart`'s own family is already handled by the existing sort rule (the leading component alone decides it), so this only closes the trailing-edge gap at `maxPart`'s own family.

Added `mergePartCellSharingMaxPartsLeadingComponentsIsNotOrphaned`, which builds real `MergePartCell` instances via `DCMergeDatePartFilter` (not the plain-Integer approximation the existing tests use) — that encoding gap is exactly why this wasn't caught the first time.

## Review notes addressed

1. Removed the stale doc-comment paragraph on `computeValidParts()` claiming it's used by `DateComparisonFormat.initPartDate()` (no longer true post Bug #76390's fix).
2. The widening loop now re-applies the same `startDate` filter as the first pass, so its result means the same thing regardless of what selector a future caller composes it with.

12/12 regression tests pass (`DateComparisonUtilValidPartsTest`, `DateComparisonIntervalConditionTest`, `DcFormatTableLensTest`, `VSChartShowDataServiceTest`, `DCMergeDateFilterTest`, `DCMergeDatePartFilterTest`). Live-verified twice against the actual fixture — all three originally-reported rows (7-1, 9-2, 12-2) now render.

jshobe-inetsoft (approved):

jshobe-inetsoft (changes requested):
There are merge conflicts that need to be resolved.

jshobe-inetsoft (approved):

PR #4936, opened 2026-09-01
Bug #76388, fix missing DayOfWeek facet groups in Month_SameWeek date comparison

jkidacki-inetsoft:
## Summary

Month_SameWeek Chart3 (Date Comparison = Standard Periods + SAME_WEEK interval, MONTH context, facet mode) rendered only 4 of 7 expected DayOfWeek(Date) facet groups. Two independent defects contributed; both are fixed here.

### Defect 1: spurious WeekOfMonth temp grouping field

`DateComparisonInfo.getIntervalTempDateGroupRef()`'s `MONTH_DATE_GROUP` branch manufactured a `WeekOfMonth(Date)` calculated temp grouping field whenever the interval level carried the `WEEK` bit (`SAME_WEEK`, `WEEK_TO_DATE`), regardless of the interval's actual granularity. For this bug's config (granularity=DAY), this spurious field was folded into the chart's real GROUP BY columns alongside the legitimate `DayOfWeek(Date)` axis field. Real defect, but verified live *not* to be what causes the reported missing-groups symptom on its own.

**Fix:** gate the branch on `granularity == WEEK` as well, matching the sibling `QUARTER_DATE_GROUP` branch's existing "week" flag.

### Defect 2 (the actual cause): orphaned-cell heuristic misapplied to facet dimension

`DateComparisonUtil.applyDateRange()`'s `computeValidParts()`/`ValidPartsSelector` heuristic assumes "part" identifies a position on one time-series axis, where a part the most-recent period's own rows don't reach is a genuinely future, not-yet-happened bucket. In facet mode, "part" instead identifies the facet itself (DayOfWeek here), and a facet the most recent period happens to lack a row for is not a future bucket — it's simply a facet whose own sub-chart legitimately has fewer periods of data than others.

Confirmed live against the actual reported scenario: the most recent period (Nov 2021) genuinely only has rows for DayOfWeek 1-4, while every earlier period (Jul-Oct) has all 7 — exactly what the reporter's own "Expected" section anticipated (facets 5-7 should span only Jul-Oct, 3 bars each, not Jul-Nov). The heuristic instead excluded facets 5-7 entirely from every period.

**Fix:** skip the heuristic entirely in facet mode (`info.isFacet() ? Collections.emptySet() : computeValidParts(...)`).

## Test plan

- [x] New test `DateComparisonInfoTempDateGroupRefTest` for defect 1 (fails pre-fix, passes post-fix)
- [x] Existing `DateComparisonIntervalConditionTest`, `DcFormatTableLensTest`, `VSChartShowDataServiceTest`, `DCMergeDateFilterTest` still pass
- [x] Defect 2's fix verified live against the actual imported `Month_SameWeek.vso` on a running server: chart now renders all 7 DayOfWeek facet groups (4 groups × 4 bars + 3 groups × 3 bars = 25 bars total), matching the reporter's exact expected count. No unit test added for this specific fix — no existing test exercises facet-mode `applyDateRange()`, and live verification against the real reported configuration was judged stronger evidence than a synthetic approximation.

Fixes Redmine #76388.


jkidacki-inetsoft:
Important correction after further live verification, prompted by a review question about why the axis labels looked different after this fix.

## Round-1 fix was a regression — reverted

Direct comparison against clean, fully unpatched `main` showed the original diagnosed defect fix (gating `getIntervalTempDateGroupRef()`'s `MONTH_DATE_GROUP` branch on `granularity == WEEK`, commit `3baf2f8d1`) was **not just unnecessary but actively harmful**: it broke the per-bar axis date labels for the facet groups that already rendered correctly.

Example (`Month_SameWeek` Chart3, group 4): on clean `main`, the label correctly reads `2021-08-04/09-08/10-06/11-10` (the real calendar date for each month's occurrence of that "same week" position), matching the test fixture's own expected baseline (`Month_SameWeek/Chart3.png`). With the original round-1 fix applied, this became an incorrect uniform `08-04/09-04/10-04/11-04` pattern (same day-of-month repeated every month).

The temp field that fix removed turned out to also feed the date-label computation path — not just the GROUP BY column the original diagnosis focused on. Removing it wasn't necessary for the reported bug (the facet-mode skip below is independently sufficient), so it's been **reverted entirely** rather than patched further.

## What's left: the actual fix

Just the facet-mode skip in `DateComparisonUtil.applyDateRange()` (commit `a104f266a`, unchanged) — confirmed via clean-main comparison to be independently necessary and sufficient for the reported missing-groups symptom, with no label regression.

## Known residual gap (intentional — not fixed here)

Facet groups 5-7's per-bar date labels render blank (empty, not wrong) until Bug #76390's `DateComparisonFormat` fix (PR #4935) also lands — that fix removes a second, independent orphaned-cell check this branch doesn't touch. Confirmed live that combining both fixes produces fully correct labels for all 7 groups. Deliberately not duplicating that commit onto this branch — **this PR depends on #4935 for the facet labels to be fully correct**, though the reported bug itself (facet group count and bar data) is fully fixed here regardless of merge order.

9/9 regression tests still pass (`DateComparisonIntervalConditionTest`, `DcFormatTableLensTest`, `VSChartShowDataServiceTest`, `DCMergeDateFilterTest`, `DCMergeDatePartFilterTest`).

jkidacki-inetsoft:
Addressed both inline review notes with commit dc2e922f4:

1. **Dead-code guard note**: moot — that entire change (`3baf2f8d1`, `DateComparisonInfo.java`) was reverted in a prior commit (`9d6661da7`) after live verification showed it regressed axis labels and wasn't actually needed for the reported bug. See the PR comment above for the full story. Resolving.

2. **Stale javadoc / `DateComparisonFormat.initPartDate()` sync note**: confirmed the javadoc claim is false as of this fix — `initPartDate()` does still call `computeValidParts()` unconditionally, no facet gate. I did not add the `isFacet()` gate to that call site here (would duplicate a fix already landing separately on Bug #76390's branch/PR #4935 for a related but distinct defect in the same method — see the "Known residual gap" note in the earlier PR comment). Fixed the javadoc to state this explicitly rather than assert a sync guarantee that no longer holds, and to point at the tracked gap instead of leaving it implicit. Resolving as documented/tracked, not as "no longer an issue" — the underlying gap for facets 5-7's labels remains until #4935 lands, same as previously noted.

jshobe-inetsoft (approved):

PR #4937, opened 2026-09-02
Add Hive to the tabular catalog SPI

Nelson-cao:
Hive becomes the fourth `TabularCatalogProvider` implementer, after OData,
SharePoint Online and Cassandra. `HiveRuntime` now enumerates the tables of
the database its data source is configured for, describes their columns and
primary keys, and reports the binding parameter an agent needs to query them.

The connector had no metadata code at all before this. `HiveCatalog` answers
both SPI phases from standard `java.sql.DatabaseMetaData` calls on the
connection `getConnection` already builds; the existing `runQuery`,
`testDataSource`, `getConnection`, `constructUrl` and `getDriver` are
unchanged.

## Three defects found while building it

**A Hive `BOOLEAN` column was going to be annotated as a string.** The obvious
reuse — `SQLTypes.convertToJava(int)` then `CoreTool.getDataType(Class)` — has
no `Types.BOOLEAN` case, only `Types.BIT`, so `BOOLEAN` falls through to
`String.class` and ends up as `XSchema.STRING`. `HiveSQLTypes`' own existing
Bug#3287 fix is evidence the driver does report `Types.BOOLEAN`. Using
`SQLTypes.convertToXType(int)` instead — also inherited unmodified by
`HiveSQLTypes` — maps `BOOLEAN` and `BIT` correctly in a single hop and adds
no code.

**A table name containing a backtick produced malformed SQL.**
`hive.support.special.characters.tablename` defaults to true and the character
array it enables includes a literal backtick, so ``SELECT * FROM `weird`table` ``
was reachable under stock configuration. Embedded backticks are now doubled,
the convention Hive documents in HIVE-6013.

**Materialized views were silently omitted.** Under HiveServer2's default
`CLASSIC` table-type mapping, `MATERIALIZED_VIEW` maps to its own client-visible
type rather than being folded into `VIEW`, so a `{"TABLE", "VIEW"}` filter drops
every materialized view. It is now in the filter.

## Known limitation, documented at the call site

`TABLE_TYPES` holds `ClassicTableTypeMapping` client-type names and therefore
depends on HiveServer2's default mapping. A deployment overriding
`hive.server2.table.type.mapping` to `HIVE` gets an empty catalog for ordinary
tables and views, because `TableType.valueOf("TABLE")` throws and the literal
string is returned unchanged. No single static value is correct under both
mappings, so the assumption and its failure mode are stated in a comment above
the field rather than worked around.

## Tests

The module had none; it now has 24, all passing. Each new test was checked to
fail without its fix. There is no Hive cluster available here, so
`Connection`/`DatabaseMetaData`/`ResultSet` are mocked and nothing has been run
against a live server — including the backtick escaping, whose grammar ships in
`hive-exec` rather than the `hive-jdbc` client artifact this module compiles
against.

### DIFF
--- a/connectors/inetsoft-hive/src/main/java/inetsoft/uql/hive/HiveCatalog.java
+++ b/connectors/inetsoft-hive/src/main/java/inetsoft/uql/hive/HiveCatalog.java
@@ -47,14 +47,24 @@ final class HiveCatalog {
    private HiveCatalog() {
    }
 
-   // "MATERIALIZED_VIEW" is the underscore form ClassicTableTypeMapping$ClassicTableTypes.
+   // "MATERIALIZED_VIEW" is the underscore form that ClassicTableTypeMapping$ClassicTableTypes.
    // MATERIALIZED_VIEW.toString() actually produces (that enum has no toString() override, so it
-   // equals name()) -- not the space form "MATERIALIZED VIEW" HiveDatabaseMetaData.
+   // equals name()) -- not the space form "MATERIALIZED VIEW" that HiveDatabaseMetaData.
    // toJdbcTableType uses elsewhere for an unrelated client-side purpose. ClassicTableTypeMapping
    // is what HiveServer2 uses under its default hive.server2.table.type.mapping=CLASSIC, and
    // keeps materialized views as their own client-visible type rather than folding them into
    // VIEW the way it folds external tables into TABLE; omitting this entry silently drops every
    // materialized view from listDatasets under that default configuration.
+   //
+   // These three strings are ClassicTableTypeMapping client-type names, and this array only works
+   // as intended because that is HiveServer2's compiled-in default mapping. If a deployment
+   // overrides hive.server2.table.type.mapping to HIVE instead, HiveTableTypeMapping.mapToHiveType
+   // resolves each string via TableType.valueOf(name); "TABLE" and "VIEW" have no matching native
+   // TableType constant, so valueOf throws, the exception is caught, and the literal input string
+   // is returned unchanged -- which then matches no real table's type. Under that configuration,
+   // listDatasets silently returns an empty catalog for every ordinary table and view (no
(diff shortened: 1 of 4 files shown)


Nelson-cao:
Thanks — I checked all three observations against the source. Taking none of them into this PR, and here is why for each.

**1. The "table not found" message reports the lowercased database name.** Confirmed: `dbName` at that point has been through `HiveRuntime.normalizedDbName`, so a source configured as `MyDB` produces `...not found in database 'mydb'`. Leaving it. Making the message echo what the user typed means threading the original name through `HiveCatalog`'s two entry points purely so an error string can differ from the value actually queried — and the lowercased name is the one the lookup used, so it is the more accurate half of that trade. Not worth an extra parameter on both signatures.

**2. `catch(Exception)` in `primaryKeyColumnNames`.** Agreed as stated, and already recorded as a known trade in this round's own review. Narrowing to `SQLException` is defensible — the driver's `getPrimaryKeys` only declares `SQLException`, so nothing is lost on the path the comment describes. Leaving it because the comment's reasoning is about HiveServer2's *server-side* response to an unimplemented Thrift method, which is not covered by the client's declared throws, and there is no cluster here to establish what actually surfaces. Widening later on a guess would be worse than narrowing later on evidence.

**3. License header year.** Not a defect. `HiveCatalog.java` and both test files use `2024` because that is what every file in the sibling connector modules uses, including `CassandraCatalog.java` and `CassandraCatalogTest.java`, which were added in the immediately preceding round and are already merged. `TabularCatalogProvider.java` uses `2026` because it lives in `core`. The two conventions differ by module, and these files match theirs. Changing them would make this connector the only inconsistent one.

On the note about tests: they do run. 24 pass locally (`HiveCatalogTest` 21, `HiveRuntimeCatalogTest` 3), re-run after rebasing onto the current base.

fredaan (approved):

fredaan (approved):

PR #4940, opened 2026-09-02
feat(wiz): add datasource delete/move endpoints to WizDatabaseController (bug-76139)

Zxm-happy:
Adds delete and move support for datasources and folders to the wiz-facing bridge controller (`WizDatabaseController`), for wiz portal bug-76139 (AA-003/AA-005). Delegates to the already-existing native `DataSourceBrowserService`/`DatasourcesService` methods StyleBI's own EM/portal UI already uses — this is substantially a wiring/authorization job, not new business logic.

## What's added

- `POST /api/wiz/datasources/delete` — bulk-capable delete of datasources/folders. Adds the explicit permission guard the underlying native services don't perform themselves (confirmed by reading `RepositoryObjectService`/`DatasourcesBaseService` directly — neither checks permission in its own call chain below the controller).
- `POST /api/wiz/datasources/move` — bulk-capable move, delegating to `DataSourceBrowserService.moveDataSource` (which already self-authorizes). Adds a wiz-side self-descendant guard (moving a folder into its own descendant) that the native service doesn't provide.
- `POST /api/wiz/datasources/move/checkDuplicate` — advisory pre-check, matches native semantics.
- `POST /api/wiz/datasources/checkOuterDependencies` — dependency pre-check surfaced per-item, for a delete-confirmation UI.

## Testing

- `WizDatabaseControllerSecurityTest` extended (23/23) — the frozen-contract test now includes the 4 new endpoint paths.
- New `WizDatabaseControllerDeleteMoveTest` (10/10) — mocked-service coverage: permission denial short-circuits before the service is ever called, force/dependency-conflict handling, mixed-batch partial-success semantics, move's self-descendant short-circuit, `MessageException` → clean 4xx mapping.
- New `WizDatabaseControllerRealRegistryTest` (4/4) — against a REAL `DataSourceRegistry`, not mocked: confirms a folder delete is a genuine repository mutation (not a hide), confirms deleting a folder containing a nested sub-folder's datasource does NOT orphan it, and pins the corrected rule that a folder's delete gate is dependency-based, not emptiness-based.
- Regression: `DataSourceBrowserServiceTest` (2/2), `RepositoryObjectServiceTest` (7/7) unmodified.
- A pre-existing, unrelated test (`DateComparisonServiceTest`) fails to compile on this branch's base — confirmed via git history to predate this change; not touched or worked around here, flagged for whoever owns that file.

## Known, pre-existing native-code gaps found while testing (not introduced or fixed by this PR)

- The delete-dependency-conflict gate only checks a folder's *direct* children — a datasource nested inside a sub-folder is deleted unconditionally by the registry's own recursive cleanup, regardless of any real outer dependency it has.
- A single, non-folder datasource delete (`DatasourcesBaseService.deleteDataSource` → `XEngine.removeDataSource`) ignores its `force` parameter entirely — it always succeeds regardless of dependencies.

Both are out of this PR's scope (native service behavior, not the wiz bridge) and are noted here for visibility.

Companion wiz-side PR (routes/service/permissions/portal UI), already merged: inetsoft-technology/stylebi-wiz#2082.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/WizDatabaseController.java
@@ -26,23 +26,28 @@ import inetsoft.sree.security.ResourceType;
 import inetsoft.sree.security.SecurityEngine;
 import inetsoft.uql.XDataSource;
 import inetsoft.uql.XRepository;
+import inetsoft.uql.asset.DependencyException;
 import inetsoft.uql.jdbc.JDBCDataSource;
 import inetsoft.uql.jdbc.SQLHelper;
 import inetsoft.uql.tabular.ScriptedQuery;
 import inetsoft.uql.tabular.SelectableTabularQuery;
 import inetsoft.uql.util.Config;
+import inetsoft.util.MessageException;
 import inetsoft.util.audit.ActionRecord;
 import inetsoft.web.admin.content.database.*;
 import inetsoft.web.admin.content.database.types.*;
 import inetsoft.web.admin.content.repository.DataSourceSettingsModel;
 import inetsoft.web.admin.content.repository.DatabaseDatasourcesService;
+import inetsoft.web.admin.model.NameLabelTuple;
 import inetsoft.web.admin.security.ConnectionStatus;
 import inetsoft.web.portal.data.CheckDuplicateResponse;
 import inetsoft.web.portal.data.DataSourceBrowserService;
 import inetsoft.web.portal.data.DataSourceConnectionStatusRequest;
 import inetsoft.web.portal.data.DataSourceInfo;
 import inetsoft.web.portal.data.DataSourceStatus;
+import inetsoft.web.portal.data.DatasourcesService;
 import inetsoft.web.portal.data.ImmutableDataSourceConnectionStatusRequest;
+import inetsoft.web.portal.data.MoveCommand;
 import inetsoft.web.portal.data.PortalDataType;
 import inetsoft.web.portal.service.datasource.DataSourceStatusService;
 import inetsoft.web.security.PermissionPath;
@@ -108,7 +113,8 @@ public class WizDatabaseController {
                                 SecurityEngine securityEngine,
                                 Config uqlConfig,
                                 XRepository xrepository,
-                                EndpointCatalogReader endpointCatalogReader)
+                                EndpointCatalogReader endpointCatalogReader,
+                                DatasourcesService datasourcesService)
    {
       this.dataSourceBrowserService = dataSourceBrowserService;
       this.dataSourceStatusService = dataSourceStatusService;
@@ -118,6 +124,7 @@ public class WizDatabaseController {
       this.uqlConfig = uqlConfig;
       this.xrepository = xrepository;
       this.endpointCatalogReader = endpointCatalogReader;
+      this.datasourcesService = datasourcesService;
    }
 
    /**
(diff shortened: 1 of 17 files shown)


Zxm-happy:
## Review round 1 (reviewer-stylebi) — CHANGES REQUESTED

Reviewed head `675510a51` against `docs/teams/2026-09-02-bug-76139-datasource-move-delete/00-charter.md` (P2-amended), `01-design.md` §2/§7, `03-reconcile.md`, and the build/verify docs, plus an independent adversarial pass and a cross-check against the already-merged wiz-side companion (`stylebi-wiz#2082`).

**Confirmed correct** (read directly, not taken on the build/verify docs' word):
- The permission guard on `deleteDatasources` genuinely gates both the folder and datasource delegate calls, sitting in front of the call, not after or on one branch only (`WizDatabaseController.java:820-856`).
- `WizDatabaseControllerRealRegistryTest` genuinely exercises a real `DataSourceRegistry` and reads results back independently of the delete call — the charter's §C1 (real deletion, not a wiz-only hide) and §C6 (nested-folder orphan) counter-assertions both hold.
- `mappingsMatchTheFrozenContract()` is strict `Set` equality (`assertEquals(Set.of(...), new HashSet<>(...))`), not a subset check — it would genuinely have failed if the 4 new paths were missing.
- The new `datasourcesService` constructor parameter is safe: grepped for `new WizDatabaseController(` — only this PR's own 3 test files construct it directly, all updated; runtime wiring is Spring `@Autowired` DI against an existing bean.
- The two documented native-code gaps (dependency gate is direct-children-only; single-datasource delete ignores `force`) are confirmed genuinely pre-existing via `git merge-base --is-ancestor` against the touched files' last-change commits — out of this PR's blast radius, correctly documented rather than fixed.

**Blocking finding:**

`WizDatabaseController.checkOuterDependencies` (`WizDatabaseController.java:979`) is declared `checkOuterDependencies(@RequestBody WizDatasourceRef[] items)` — binding the entire HTTP body as a bare JSON array. Its only production caller, the already-merged `wiz-services` `checkOuterDependencies()` (`datasourceManagementService.ts`), sends `apiClient.post(CHECK_DEPENDENCIES_ENDPOINT, { items })` — a JSON **object** `{"items": [...]}`, not a bare array. With no root-unwrapping configured anywhere in this repo, Jackson cannot bind an object into an array type; every real call will 400 before the method body runs.

This is inconsistent with the other 3 endpoints in this same PR, which all correctly wrap their request in a record matching what the wiz-services caller sends (`WizDatasourceDeleteRequest`, `WizMoveRequest`, `WizMoveCheckDuplicateRequest`). Only `checkOuterDependencies` skips the wrapper. The mismatch traces back to `01-design.md` itself (§2.1 vs §3.1 disagree on wire shape) and was never caught because the only test exercising this method calls the Java method directly, bypassing Spring's `@RequestBody`/Jackson binding entirely.

Practical impact: the wiz portal's `DeleteConfirmDialog.tsx` swallows a failed pre-check (`.catch(() => setDependencyMessages({}))`), so the dependency-warning UI this endpoint exists to power silently never fires in production — every delete confirmation looks clean regardless of real outer dependencies. No data-loss by itself (the delete endpoint's own `force=false` gate still protects folder deletes), but it's a confirmed instance of a caller silently treating a shape-rejected call as "no problem found."

**Fix**: add `record WizCheckDependenciesRequest(List<WizDatasourceRef> items) {}` and change the endpoint signature to accept it, matching the pattern already used by its 3 siblings. Add a Spring-MVC-level test that posts a real `{"items": [...]}` body (not a direct Java call) so a regression here is caught by the wire layer, not just the business logic.

**Non-blocking (nits, not sent back for):**
- `checkOuterDependencies`'s generic `catch(Exception ex)` forwards raw `ex.getMessage()` to the client for any unexpected failure, not just a real `DependencyException` — but this mirrors the native `DataSourceController.checkDsOuterDependenciesSelected`/`checkOuterDependencies` byte-for-byte, so it's a pre-existing convention, not something this PR introduced. Worth a follow-up ticket.
- `05-verify.md`'s SS-F1/F2/F3 (test-coverage gaps, not confirmed defects) — agreed, correctly classified, not blocking.

**CI**: `build` check is failing, but re-ran the job log directly and confirmed the sole compile error (`DateComparisonServiceTest.java:220/235`) is pre-existing on `stylebi-wiz-main-rebase`'s current tip (`git merge-base` confirms that file's last commit is an ancestor of this PR's own parent) — not introduced or worsened by this branch. Still, per this review's own merge gate, all-checks-green is a literal precondition for approve+merge, so this PR can't be merged until that upstream break is fixed (separate PR/owner) even though it's not this PR's own fault.

**Verdict: changes requested** — one blocking finding (checkOuterDependencies wire-shape mismatch) plus a non-code CI blocker (pre-existing, unrelated build failure).

Full findings: `docs/teams/2026-09-02-bug-76139-datasource-move-delete/06-review-r1.md`, "StyleBI side (reviewer-stylebi)" section.

(Posting as a plain comment, not `gh pr review -r`, since the PR author and this review session are the same GitHub account and GitHub rejects self-review.)

Zxm-happy:
Addressed both items from the round-1 review (both confirmed against the actual code, not just the review's description).

## Fix 1 (blocking): `checkOuterDependencies` request-body shape — commit 3f100d9d7

Confirmed the finding by direct testing: reverted the fix locally, recompiled, and reran the new binding test — it failed exactly as predicted (`Status expected:<200> but was:<400>`, underlying cause `HttpMessageNotReadableException: Cannot deserialize value of type inetsoft.web.wiz.request.WizDatasourceRef[] from Object value (token JsonToken.START_OBJECT)`). Reapplied and reran green.

- Added `WizCheckDependenciesRequest(List<WizDatasourceRef> items)` (`core/src/main/java/inetsoft/web/wiz/request/WizCheckDependenciesRequest.java`), matching the wrapper-record pattern the PR's 3 sibling endpoints already use.
- Changed `WizDatabaseController.checkOuterDependencies` to bind `@RequestBody WizCheckDependenciesRequest request` instead of a bare `WizDatasourceRef[]`.
- Updated `WizDatabaseControllerDeleteMoveTest`'s direct-call test to construct the new request type.
- Added `WizDatabaseControllerCheckDependenciesRequestBindingTest` — a real `MockMvc`/Jackson-binding test (posting an actual `{"items":[...]}` JSON body), following the existing `WizDocSearchControllerRequestBindingTest` convention in this codebase for exactly this class of wire-shape regression, which the existing direct-Java-call test cannot catch.

Ran `WizDatabaseControllerDeleteMoveTest` (10), `WizDatabaseControllerSecurityTest` (23), `WizDatabaseControllerRealRegistryTest` (4), and the new test (1) together via `mvn surefire:test` per `04-build.md`'s recipe — **38/38 pass**.

## Fix 2 (blocking CI, pre-existing, not this PR's own code) — commit 01f5aa60b

`DateComparisonServiceTest.java:220/235` still constructed `DateComparisonService.Comparison` with the old 8-arg shape (a trailing `null` for the `frame` param), which commit 729755296 (#4818) dropped from the record when it removed the dead date-comparison colour palette. That commit updated 3 of the 5 call sites but missed these two inline constructions. Removed the same trailing `null` from both — purely mechanical, matching the pattern already applied everywhere else in the file. `DateComparisonServiceTest` — **31/31 pass**.

Kept as a separate commit from Fix 1 since it's an unrelated, pre-existing break with a different owner/domain — easy to revert independently if preferred.

Zxm-happy:
Round 2 re-review, verifying the fixes from 07-fix-r1.md against round 1's blocking finding.

Confirmed by reading the current diff/worktree directly (not just the fix log):

1. Fix 1 (wire-shape): checkOuterDependencies now takes @RequestBody WizCheckDependenciesRequest, matching the 3 sibling endpoints pattern exactly. New record WizCheckDependenciesRequest(List<WizDatasourceRef> items) confirmed present. Grepped all 10 @RequestBody params in the controller: every one is now object/record-typed, none is a bare array, so this defect class cannot recur elsewhere in this file.
2. New regression test WizDatabaseControllerCheckDependenciesRequestBindingTest posts a real JSON string body through standaloneSetup(controller).build() MockMvc (genuine Jackson HttpMessageConverter binding, not a direct method call), and asserts both HTTP 200 and response-body content -- specific enough that the pre-fix bare-array signature would fail it (400, per the fixer's own revert-and-rerun log).
3. Fix 2 (pre-existing DateComparisonServiceTest compile break): confirmed as a minimal 2-hunk mechanical repair, no other file touched, and a separate commit (01f5aa60b) from Fix 1 (3f100d9d7).
4. CI: all checks green at the current head, including the previously-failing build check (build passed in ~19 minutes).
5. Mergeable: MERGEABLE, mergeStateStatus: CLEAN.
6. Adversarial check: no other endpoint in this file has the same array-vs-object @RequestBody mismatch.

This PR is ready to merge. (Self-approval note: I cannot use gh pr review --approve since the author account matches this review session's account -- posting this as an explicit approving comment instead.)

PR #4943, opened 2026-09-02
fix(item5): resolve date_in builtin range names case-insensitively

larryliang-inetsoft:
## Summary
- `ConditionUtil.fromModelToConditionList`'s builtin `DateCondition` lookup (and the matching worksheet date-range-assembly lookup) used `equals()` against the exact builtin name, so a case variant an LLM might naturally produce (e.g. `"Last Year"` for the builtin `"Last year"`) silently failed to match and fell through instead of resolving to the real condition.
- `WorksheetMutationSupport`'s parallel builtin lookup had the same case-sensitivity bug.
- Both now match case-insensitively (`equalsIgnoreCase`), and `ConditionUtil` now throws a clear `IllegalArgumentException` naming the bad value when nothing matches (builtin or worksheet assembly), instead of silently dropping the condition.

## Test plan
- [x] Added `ConditionUtilTest#aBuiltinDateRangeNameInTheWrongCaseStillResolvesToTheSameDateCondition` — wrong-cased builtin name resolves to the same `DateCondition` as the correctly-cased name.
- [x] Added `ConditionUtilTest#anUnrecognizedDateRangeNameThrows` — an unrecognized date range name fails loud instead of falling through silently.
- [x] Added `WorksheetEditServiceMutatorsTest#addFilterWithDateInResolvesANameInTheWrongCase` — mirrors the fix in `WorksheetMutationSupport`'s builtin lookup.
- [x] `./mvnw -pl community/core -o clean test -Dtest=ConditionUtilTest,WorksheetEditServiceMutatorsTest` — 230 tests, 0 failures, 0 errors.

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/model/condition/ConditionUtil.java
+++ b/core/src/main/java/inetsoft/web/composer/model/condition/ConditionUtil.java
@@ -133,7 +133,7 @@ public class ConditionUtil {
 
                   // look for a built in condition with this name and clone it
                   for(DateCondition dateCondition : DateCondition.getBuiltinDateConditions()) {
-                     if(dateCondition.getName().equals(value)) {
+                     if(dateCondition.getName().equalsIgnoreCase(value)) {
                         condition = (XCondition) dateCondition.clone();
                         break;
                      }
@@ -143,10 +143,21 @@ public class ConditionUtil {
                   if(condition == null && rws != null) {
                      Worksheet ws = rws.getWorksheet();
 
-                     if(ws.getAssembly(value) instanceof DateRangeAssembly) {
-                        condition = ((DateRangeAssembly) ws.getAssembly(value))
-                           .getDateRange().clone();
+                     for(Assembly assembly : ws.getAssemblies()) {
+                        if(assembly instanceof DateRangeAssembly &&
+                           assembly.getName().equalsIgnoreCase(value))
+                        {
+                           condition = ((DateRangeAssembly) assembly).getDateRange().clone();
+                           break;
+                        }
                      }
+                  }
+
+                  if(condition == null) {
+                     throw new IllegalArgumentException(
+                        "'" + value + "' is not a known date range. Call list_condition_date_ranges " +
+                        "for the exact built-in names (e.g. \"Last month\"), or name a worksheet " +
+                        "date-range assembly.");
                   }
                }
                else if(conditionModel.getOperation() == AbstractCondition.TOP_N ||
--- a/core/src/main/java/inetsoft/web/composer/model/condition/ConditionUtil.java
+++ b/core/src/main/java/inetsoft/web/composer/model/condition/ConditionUtil.java
@@ -143,12 +143,21 @@ public class ConditionUtil {
                   if(condition == null && rws != null) {
                      Worksheet ws = rws.getWorksheet();
 
-                     for(Assembly assembly : ws.getAssemblies()) {
-                        if(assembly instanceof DateRangeAssembly &&
-                           assembly.getName().equalsIgnoreCase(value))
-                        {
-                           condition = ((DateRangeAssembly) assembly).getDateRange().clone();
-                           break;
+                     if(ws.getAssembly(value) instanceof DateRangeAssembly) {
+                        condition = ((DateRangeAssembly) ws.getAssembly(value))
+                           .getDateRange().clone();
+                     }
+                     else {
+                        // fall back to a case-insensitive scan only when no assembly is
+                        // named exactly value, so two assemblies differing only by case
+                        // never shadow the one the caller actually named
+                        for(Assembly assembly : ws.getAssemblies()) {
+                           if(assembly instanceof DateRangeAssembly &&
+                              assembly.getName().equalsIgnoreCase(value))
+                           {
+                              condition = ((DateRangeAssembly) assembly).getDateRange().clone();
+                              break;
+                           }
                         }
                      }
                   }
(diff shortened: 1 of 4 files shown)


larryliang-inetsoft:
## Round 1 review

Head commit reviewed: `beffa0bdb2c54a77b220fd82f8975a8d72abee40`.

**Checked**: full DATE_IN branch in `ConditionUtil.java` (confirmed the new fail-loud throw sits
inside the DATE_IN branch, not the shared line-165 fallback used by every other operator);
`WorksheetMutationSupport.resolveDateInCondition`'s scoped one-line change; `date_in`'s contract
in the plugin tool surface (`list_condition_date_ranges`: "These are names, not dates; do not
invent one" -- confirms fail-loud on no-match cannot break a legitimate literal-date use, since
there isn't one); independently re-extracted and checked all 83 builtin `dateConditions.xml`
entries for a case-insensitive collision (none found, actual count is 83, not the 69 the
diagnosis/refute docs cite -- doesn't change the safety conclusion, noted for the record).

**Alternate scenarios walked**: the sibling code path (viewsheet `set_condition` via
`ConditionUtil.java` vs. worksheet-table tools `add_filter`/`set_conditions`/`add_named_group` via
`WorksheetMutationSupport.java`, which its own javadoc says deliberately mirrors the former);
`rws == null` boundary; case-collision risk in both the builtin list and worksheet-assembly names.
Counter-example I tried and could not build: a regression in builtin-name matching for any
non-DATE_IN operator, or in the interactive Composer UI dialogs (dropdown-sourced, cannot send a
case mismatch today).

**Counter-example I could build** (findings below): a wrong-case *custom worksheet-defined*
`DateRangeAssembly` name behaves differently between the two "mirrored" resolvers after this PR.

### Findings

**Finding 1 (Important)** -- `ConditionUtil.java`'s `DateRangeAssembly` fallback (DATE_IN branch)
was widened to `equalsIgnoreCase` (iterating `ws.getAssemblies()`), but
`WorksheetMutationSupport.resolveDateInCondition`'s equivalent fallback
(`ws.getAssembly(value) instanceof DateRangeAssembly`) was left exact-match only. These two
methods are documented as deliberately mirroring each other. Concretely: a worksheet
`DateRangeAssembly` named `"MyFiscalQ1"` referenced via `date_in` value `"myfiscalq1"` now
resolves through `set_condition` (viewsheet path) but is rejected with `"not a known date range"`
through `add_filter`/`set_conditions`/`add_named_group`/`edit_named_group` (worksheet path) for
the identical input -- the exact class of inconsistency this item is about, reintroduced at the
one spot the diagnosis flagged as optional/fixer's-discretion without reconciling the two files.
Suggested fix: apply the same treatment to `WorksheetMutationSupport.java`'s fallback, or revert
`ConditionUtil.java`'s to exact match, so both agree.

**Finding 2 (Important)** -- the rewritten `DateRangeAssembly` fallback loop in
`ConditionUtil.java` (direct `ws.getAssembly(value)` lookup -> iterate-and-`equalsIgnoreCase`) has
zero regression-test coverage (neither this PR's new tests nor any pre-existing
`ConditionUtilTest` case exercises the `DateRangeAssembly` path at all), and it returns the first
case-insensitive match in iteration order rather than preferring an exact match -- if a worksheet
ever has two `DateRangeAssembly`s whose names differ only by case, resolution is no longer
deterministic the way the original exact `ws.getAssembly(value)` lookup was. Suggested fix: try an
exact match first, fall back to case-insensitive on a miss, and add a regression test for this
branch (correct-case and wrong-case), mirroring what this PR already did for the builtin-name
half in both files.

### Verdict

**Changes requested.** Core mechanism, throw placement, and the builtin-name half of the fix (both
files) are correct and well-tested by my independent read. CI's `build` check failed on an
unrelated pre-existing flake (`ScheduleTaskFolderControllerTest` NPE, unrelated to conditions/date
ranges -- corroborated by different unrelated flaky failures on other unrelated PRs' builds in the
same window); a rerun was triggered and is still in flight. CI should be rechecked green before
merge regardless of these findings.

Full review: `docs/teams/2026-09-01-bugs-composer-b5c6-followon/item5/04-review-r1.md` in
stylebi-wiz.

larryliang-inetsoft:
## Round 2 review

Head commit reviewed: `8e5e1d16df6934f584505b7fc04126fab3765ae1` (range since round 1:
`beffa0bdb2c54a77b220fd82f8975a8d72abee40..8e5e1d16d`).

**Checked**: full diff since round 1, and the full current content of both changed production
files at the new head (not just the incremental patch). Confirmed both files now use the identical
shape -- exact-name match via `ws.getAssembly(value)` first, falling back to a case-insensitive
scan over `ws.getAssemblies()` only on a miss -- closing both round-1 findings:
- Finding 1 (cross-file inconsistency on custom `DateRangeAssembly` names): fixed --
  `WorksheetMutationSupport.resolveDateInCondition` now mirrors `ConditionUtil.java` exactly, per
  its own javadoc's claim.
- Finding 2 (untested rewrite, not exact-match-first): fixed -- exact match is tried first in both
  files (restoring the original determinism for two assemblies differing only by case), and the
  branch is now covered by 4 new tests across both files, including a test that constructs exactly
  that two-assemblies-differing-by-case scenario and asserts the exact name wins over scan order.

**Alternate scenario tried this round**: exact-match hit that isn't a `DateRangeAssembly` (e.g. a
table assembly happens to share the exact name), to see if the type-filtered case-insensitive
fallback was skipped incorrectly in that case -- confirmed it correctly proceeds to the fallback
scan for an actual `DateRangeAssembly` match elsewhere by name. No new problem found.

**New findings**: none.

**Verdict**: Approved. CI green -- `build` passed (19m57s, this repo's normal full-suite runtime),
Semgrep OSS/Scan, claude-review, and license/cla all pass.

Full review: `docs/teams/2026-09-01-bugs-composer-b5c6-followon/item5/05-review-r2.md` in
stylebi-wiz. Squash-merging.

PR #4945, opened 2026-09-02
Add Aerospike and OrientDB to the tabular catalog SPI, and let it carry sampled columns

Nelson-cao:
Three related changes that only make sense together: a contract decision, and the two connectors that are the first to need it.

## The contract question

`TabularCatalogProvider.describeDataset` promised "the columns of this dataset". Six connectors cannot honour that literally — they report the fields a bounded sample happened to contain. The javadoc now says a sampled column list satisfies the contract, provided the imprecision is declared **at dataset level** rather than per column, and that a source publishing declared metadata must be read rather than sampled.

Per-column marking was rejected: every column from one `describeDataset` call shares one trust level, so the marker would carry the same value on each, and nothing downstream reads it. The dataset-level signal reuses the mechanism that already delivers a truncated-sample caveat into the annotation prompt, which has live readers.

## Aerospike

Fifth implementer. Hive's shape — one `Enter SQL` property, a required `Namespace` scoping the dataset id to a bare set name — and the first connector to set the incompleteness signal, since its driver unions bin names across at most 1000 scanned records and lets the last record win a type conflict.

Two defects found in review and fixed:

- **The generated query emitted a bare identifier.** The driver declares the double quote as its identifier quote and builds this same query shape quoted, so a set named `order` or `my set` produced SQL that fails to parse, silently.
- **`describeDataset` denied a set that `listDatasets` had just returned.** The driver compiles `tableNamePattern` as a Java regex, translating only `%`, so `orders(2024)` does not match itself. Fifteen tests could not have caught this: they asserted on mocked return values, and a mock yields its stubbed rows whatever pattern it was handed. The fix asserts on the argument.

A follow-up round then found the first fix had traded correctness for cost — `getColumns` runs a live query per matched table, so a literal `%` turned one `describeDataset` into one query per set. Now `Pattern.quote`, falling back to `%` only for a name containing one, since the driver substitutes `%` before compiling.

## OrientDB

Sixth implementer. Its database scope comes from `Connection.getCatalog()` rather than a property, so the dataset id stays a bare class name.

Schema-less classes return no declared columns at all — `getColumns` walks `clazz.properties()` — which is a different situation from Aerospike, where something is always inferred. Those classes get `@rid`, a real pseudo-property the driver handles and `SELECT` accepts, plus the incompleteness signal, so they are annotated at table level and remain queryable through `SELECT @rid FROM`. No column is synthesised: a fabricated column would fail the moment anything selected it.

`listDatasets` passes `types = {"TABLE"}` explicitly, because the driver's default includes `SYSTEM TABLE` and would hand `OUser` and `ORole` to the annotator as business tables.

Worth knowing for anyone reading these three connectors together: they quote three different ways. Hive doubles backticks, Aerospike doubles double quotes, OrientDB escapes them with a backslash. Nothing here derives from anything else, and `getIdentifierQuoteString()` is not a reliable guide — OrientDB's returns a space, the JDBC convention for "unsupported", while its grammar has a real `QUOTED_IDENTIFIER` production.

## Tests and limits

Both connectors' modules had none; they now have 47 between them, and the four earlier implementers' suites re-run green. There is no Aerospike or OrientDB instance available and neither driver is a build dependency — both are loaded reflectively and expected from the user. Driver behaviour was established by reading the published sources, so claims about it are read-code-correct rather than executed; the backtick escaping in particular has never been through a live parser.

### DIFF
--- a/connectors/inetsoft-aerospike/src/main/java/inetsoft/uql/aerospike/AerospikeCatalog.java
+++ b/connectors/inetsoft-aerospike/src/main/java/inetsoft/uql/aerospike/AerospikeCatalog.java
@@ -24,6 +24,7 @@ import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.ResultSet;
 import java.util.*;
+import java.util.regex.Pattern;
 
 /**
  * Assembles AerospikeRuntime's TabularCatalogProvider answers from standard java.sql.
@@ -40,8 +41,15 @@ import java.util.*;
  * tableNamePattern arguments to have been matched as literal equality: JDBC allows these to be
  * treated as LIKE patterns, and namespace/datasetId are not guaranteed free of underscore/percent.
  * getColumns goes further than a LIKE-pattern risk: the driver compiles tableNamePattern as a Java
- * regex (see describeColumns below), so a literal "%" is passed there rather than datasetId itself
- * and this exact-match filter is what actually narrows the result, not the pattern argument.
+ * regex, and -- more importantly for cost, not just correctness -- runs one live "SELECT ... LIMIT
+ * 1" against the cluster per regex-matched table (AerospikeDatabaseMetadata.getMetadata), unlike
+ * getTables/getPrimaryKeys, which only filter already-fetched, cached cluster metadata in memory.
+ * Passing "%" there unconditionally matches every set in the namespace and turns one
+ * describeDataset call into one live query per set in the namespace; see describeColumns below for
+ * how the pattern actually passed avoids that in the common case, and the one case (a literal '%'
+ * in datasetId) where it cannot and falls back to the expensive-but-correct "%". This exact-match
+ * filter is retained regardless, as the correctness backstop independent of which pattern was
+ * passed.
  */
 final class AerospikeCatalog {
    private AerospikeCatalog() {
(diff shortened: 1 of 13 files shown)


Nelson-cao:
Thanks — checked all three against the source. Taking one, declining two, reasons below.

**1. Line-number-pinned driver citations — taking it, and it is sharper than described.** The two connectors are inconsistent with each other. `AerospikeCatalog` cites by method name and states the artifact it was read from (`com.aerospike:aerospike-jdbc:1.10.1`), so a reader can re-check it after a bump. `OrientDBCatalog`'s class javadoc cites bare line numbers — 733-734, 886-887, 1445-1446 — in a sentence that never says which version they came from. One style survives a driver upgrade as a checkable claim; the other silently becomes wrong. Fixing the OrientDB side to match: name the method, name the version, keep the line number only as a hint.

**2. The "not found" message conflating absent with column-less — leaving it.** You are right that it is overloaded, and right that it is likely unreachable: an Aerospike set becomes driver-visible only once it holds records, and a record has at least one bin, so "exists but yields zero columns" has no obvious way to occur. Splitting the message would mean inventing a distinction between two states, one of which we cannot demonstrate is reachable — and the SPI requires a throw either way. If a real cluster ever produces that case, the message is the smallest part of what we would want to change.

**3. `assert driver != null` — out of scope, deliberately.** Correct that assertions are off without `-ea`, so it guards nothing. But `getConnection` and `getDriver` are pre-existing methods this round froze by charter, and both connectors' catalogs reuse them unchanged. Changing them here would mean touching the query path this work took care not to disturb, in a PR whose whole claim is that it does not. Worth its own change, in both connectors together, since they share the pattern.

On what you could not verify: the test suites do run and pass — 47 across the two connectors, plus the four earlier implementers' suites re-run green after the rebase. Both were executed locally before this PR opened, not inferred.

fredaan (approved):

PR #4949, opened 2026-09-02
fix(wiz): align the agent hyperlink service with the dialog's model

fredaan:
## Summary
- `AssemblyHyperlinkService` (the wiz-agent's thin wrapper around `HyperlinkDialogService`) had three gaps relative to the real hyperlink-dialog UI it's meant to expose:
  - `link.self` was read into the model but had no effect — the Angular dialog's "open in current window" checkbox is UI sugar that rewrites `targetFrame` to `"SELF"` before submit, and `HyperlinkDialogService.setHyperlinkDialogModel` only ever reads `targetFrame`, never `getSelf()`. `apply()` now applies the same rewrite, unconditionally and after every other field, so it overrides an explicit `targetFrame` in the same call exactly like the checkbox does.
  - `paramList` (the dialog's custom constant/field parameters, beyond the `sendViewsheetParameters`/`sendSelectionsAsParameters` bulk toggles) had no wiring at all on this path. `apply()` now maps it onto `InputParameterDialogModel`, validating a field-sourced value against the assembly's own bindable fields; `describe()` reads it back.
  - `describe()` never copied `assetLinkId` or `disableParameterPrompt` into its response even though the model carried both correctly — the same "write works, read never surfaces it" pattern already on record in the wiz plugin's own known-pitfalls notes.
- Follow-up commit fixes a round-trip bug found in review: a `constant`-valued `paramList` entry with no explicit `type` would read back as `valueSource:"field"` (since `Hyperlink.getParameterType` returns the stored `null` verbatim, and the read path infers `valueSource` from whether that's null) — a read-then-resubmit agent would then have its own working param rejected by the new field-name validation. Fixed by defaulting a type-less constant entry to `XSchema.STRING`, matching what the real Angular dialog itself defaults to.
- Companion PR in `stylebi-wiz`: inetsoft-technology/stylebi-wiz#2105 — exposes all of this through the `set_hyperlink`/`get_hyperlink` MCP tool.

## Test plan
- [x] `AssemblyHyperlinkServiceTest`: 49/49 passing, 13 new tests covering `self` normalization (including overriding an explicit `targetFrame`), `paramList` validation/clearing/round-trip, and the `describe()` read-back fix
- [x] `mvn -pl core compile` clean

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyHyperlinkService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyHyperlinkService.java
@@ -22,6 +22,7 @@ import inetsoft.sree.security.IdentityID;
 import inetsoft.sree.security.ResourceAction;
 import inetsoft.uql.asset.AssetEntry;
 import inetsoft.uql.asset.AssetRepository;
+import inetsoft.uql.schema.XSchema;
 import inetsoft.web.binding.drm.DataRefModel;
 import inetsoft.web.composer.model.vs.HyperlinkDialogModel;
 import inetsoft.web.composer.model.vs.InputParameterDialogModel;
@@ -484,7 +485,17 @@ public class AssemblyHyperlinkService {
          param.setName(name);
          param.setValueSource(source);
          param.setValue(value);
-         param.setType("constant".equals(source) ? str(paramMap, "type") : null);
+
+         // A constant entry always needs a type -- Hyperlink.getParameterType(name) returning
+         // null is how HyperlinkDialogService.getHyperlinkDialogModel's read path infers
+         // valueSource "field" (type == null ? "field" : "constant"). Leaving type null here for
+         // an agent that (naturally) omitted it would round-trip a constant param back as a
+         // "field" one on the very next read, and this class's own field-name validation above
+         // would then refuse it on resubmission -- rejecting a param that worked. XSchema.STRING
+         // matches what the real dialog itself defaults to (InputParameterDialog: every
+         // newly-created or "constant"-switched-to param starts as XSchema.STRING).
+         String type = str(paramMap, "type");
+         param.setType("constant".equals(source) ? (type == null ? XSchema.STRING : type) : null);
          params.add(param);
       }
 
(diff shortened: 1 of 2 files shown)


fredaan:
## Code review (code-reviewer agent, before this PR was opened)

Reviewed this commit (`fb4354d4f`) together with the companion commit in `stylebi-wiz`'s `plugin/composer` (`75e7e1b43`), since they're one logical change.

### Critical (fixed — see `1cecd9d54` above)

**`constant`-valued `paramList` entries without an explicit `type` silently mis-round-trip, and can corrupt themselves on a read-then-resubmit cycle.**

`parseParamList` defaulted `valueSource` to `"constant"` when omitted, but left `type` `null` if the caller didn't pass one. Traced the actual persist/read cycle in `HyperlinkDialogService`: on write, `Hyperlink.setParameterType(name, null)` still adds the key with a null value; on read, `Hyperlink.getParameterType(name)` returns that null verbatim, and `getHyperlinkDialogModel` infers `valueSource` purely from whether that's null (`"field"`) or not (`"constant"`). A constant param with no `type` therefore reads back as `valueSource:"field"`, and if an agent resubmits that read-back state, the new field-name validation on `paramList` rejects it — corrupting a working param purely because of a missing type default.

**Fixed**: `1cecd9d54` defaults a type-less constant entry to `XSchema.STRING`, matching what the real Angular dialog itself defaults every new/switched-to constant parameter to (`InputParameterDialog.ngOnInit`/`changeValueSource`). Added `paramListEntryWithNoExplicitTypeDefaultsToStringSoItRoundTripsAsConstant` as a regression test.

### Important (fixed)

**Commit message overclaimed the new-test count** — said "Adds 15 tests," actual diff added 12 (verified `git diff fb4354d4f~1 fb4354d4f -- ...Test.java`, 36→48). Noted and corrected in `1cecd9d54`'s own message and in the companion PR's testplan doc.

### What checked out clean
- `self` → `targetFrame="SELF"` override: correctly unconditional, applied after every other field, correctly skipped on the `linkType:"none"` early-return path. Verified against `HyperlinkDialogService` that `self` really is derived purely from `targetFrame=="SELF"` on read, confirming this is the only correct way to make it round-trip.
- `paramList` field-name validation against `model.getFields()`: sound, no bypass found, null-safe (`HyperlinkDialogModel.getFields()`/`getParamList()` both null-guard to empty lists).
- Clearing (`linkType:"none"`) correctly nulls `paramList` too, matching the existing `webLink`/`tooltip`/etc. nulling in that branch.
- No syntax/brace issues; null-safety solid apart from the finding above, which was a semantic/round-trip bug, not a crash.

PR #4950, opened 2026-09-02
feat(wiz): add worksheet-dependents endpoint so an agent can check usage before editing (PCB-006)

fredaan:
## Summary
- Part of PCB-006 (docs/bug-reports/plugin-composer-chart-binding.md, stylebi-wiz repo): when a user tells the agent "several dashboards use this worksheet, don't touch it," that claim should be checked, not taken on faith.
- Adds `GET /api/wiz/v1/agent/worksheet/{sessionToken}/dependents`, returning whether the connected worksheet is saved, its path, and every other saved asset (viewsheets, other worksheets) that currently depends on it.
- Backed by the same `AssetRepository.getSheetDependencies` check the Portal's own "remove this dataset?" dialog already runs before a delete (`DataSetService.isWorksheetRemoveable`) — not a new dependency-tracking mechanism.
- An unsaved (temporary-scope) worksheet reports `saved:false` with an empty dependents list without attempting the repository lookup — nothing can point at an asset that was never saved.

## Test plan
- [x] `WorksheetAgentControllerTest` — 125/125 pass (4 new: flag-off refusal, unsaved worksheet, saved with no dependents, saved with a real dependent viewsheet named by path+type)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -63,6 +63,7 @@ import inetsoft.uql.util.filereader.TextUtil;
 import inetsoft.util.Catalog;
 import inetsoft.util.CoreTool;
 import inetsoft.util.FileSystemService;
+import inetsoft.util.MissingAssetClassNameException;
 import inetsoft.web.composer.ws.LayoutGraphService;
 import inetsoft.web.composer.ws.WorksheetControllerService;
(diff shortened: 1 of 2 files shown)


fredaan:
Fixed: `dependents()` now catches `SAXParseException`/`MissingAssetClassNameException` around `getSheetDependencies`, mirroring the tolerance in `AssetRepository#checkSheetRemoveable` (treats an unreadable sheet as "no dependents" instead of propagating to a 500). Added two regression tests (`dependentsToleratesUnreadableSheetXml`, `dependentsToleratesMissingAssetClass`). Full `WorksheetAgentControllerTest` suite: 127/127 passing.

PR #4951, opened 2026-09-02
feat(wiz): add CalcFieldAgentService for viewsheet-scoped calc field CRUD (PCB-006)

fredaan:
## Summary
- Part of PCB-006 (docs/bug-reports/plugin-composer-chart-binding.md, stylebi-wiz repo): the only correct path for a data need that must not touch a worksheet other dashboards share is a viewsheet-scoped calc field (`Viewsheet.calcmap` — the field the Formula Editor dialog creates), and no tool on the wiz-agent surface could create, edit, or remove one.
- Adds `POST /api/wiz/v1/agent/binding/{sessionToken}/calc-field`, delegating to the same `ModifyCalculateFieldServiceProxy` the Formula Editor's OK button drives (via `sessions.mutate` + a built `ModifyCalculateFieldEvent`) rather than reimplementing its wizard/cube/aggregate-info side effects.
- Table names are matched case-insensitively against `list_bindable_fields`'s own listing but **stored under the listing's canonical casing**, not the caller's raw string — `Viewsheet.calcmap` is a plain case-sensitive map, so storing under the wrong casing would silently create an orphaned field.
- The bindable-table listing used for validation is deliberately **unscoped** (`assembly: null`), matching `SelectionBindingService.resolveTable`'s own precedent — scoping it to a given `assembly` would narrow the check to that assembly's current bound source and wrongly refuse a real table when the assembly happens to be bound elsewhere.

**Stacked on #4947** (this branch is based on `fix/pva-016-selection-binding-controller-regression`, since both touch `BindingAgentController`'s constructor and stacking avoids repeating that PR's own root cause). Please merge/land #4947 first, or rebase this one, before merging.

## Review history
An independent code-reviewer pass (agent) initially found two real defects in the first version of `requireBindableTable` — case-mismatch would store under the wrong key, and passing `assembly` through would wrongly scope/narrow the validation listing. Both are fixed in this version, with regression tests (`resolvesTableNameToTheListingsCanonicalCasing`, `tableValidationIsNotScopedToTheGivenAssembly`) added alongside the fix.

## Test plan
- [x] `CalcFieldAgentServiceTest` — 11/11 pass (create/edit/rename/remove/defaults/validation, including the two regression cases above)
- [x] `BindingAgentControllerTest` — 14/14 pass (8 constructor call sites updated for the new arity)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
@@ -66,6 +66,7 @@ public class BindingAgentController {
                                  TableBindingService tableService,
                                  CalcTableService calcService,
                                  SelectionBindingService selectionService,
+                                 CalcFieldAgentService calcFieldService,
                                  SheetAgentBroadcastService broadcast)
    {
       this.feature = feature;
@@ -79,6 +80,7 @@ public class BindingAgentController {
       this.tableService = tableService;
       this.calcService = calcService;
       this.selectionService = selectionService;
+      this.calcFieldService = calcFieldService;
       this.broadcast = broadcast;
    }
 
@@ -427,6 +429,17 @@ public class BindingAgentController {
                                         request.columns(), request.additionalTables(),
                                         request.measure(), Boolean.TRUE.equals(request.force()),
                                         linkUri);
+   }
+
+   @PostMapping("/api/wiz/v1/agent/binding/{sessionToken}/calc-field")
+   public void modifyCalcField(@PathVariable String sessionToken,
+                               @RequestBody CalcFieldAgentService.CalcFieldRequest request,
+                               @RequestParam(required = false, defaultValue = "") String linkUri,
+                               Principal user)
+      throws Exception
+   {
+      requireEnabled();
+      calcFieldService.modify(sessionToken, user, request, linkUri);
    }
 
    @PostMapping("/api/wiz/v1/agent/binding/{sessionToken}/table/source")
@@ -789,5 +802,6 @@ public class BindingAgentController {
    private final TableBindingService tableService;
    private final CalcTableService calcService;
    private final SelectionBindingService selectionService;
+   private final CalcFieldAgentService calcFieldService;
    private final SheetAgentBroadcastService broadcast;
 }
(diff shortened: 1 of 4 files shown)


fredaan:
Fixed the permission-check gap flagged by the review bot.

**Root cause:** `CalcFieldAgentService.modify()` called `ModifyCalculateFieldServiceProxy.modifyCalculateField(...)` directly. That proxy/service performs no permission check of its own — the check lives solely in `ModifyCalculateFieldController` (the native Formula Editor dialog's entry point), gating on `VIEWSHEET_CALCULATED_FIELD`/`ACCESS`. By calling the service directly, the wiz agent path bypassed that gate entirely: an agent session could create/edit/remove viewsheet calc fields (including native-SQL ones) for a user an admin had specifically denied the "Calculated Fields" permission.

**Fix (`core/src/main/java/inetsoft/web/wiz/binding/CalcFieldAgentService.java`):**
- Injected `SecurityEngine` into the constructor (matching the convention already used elsewhere in `inetsoft.web.wiz` — e.g. `WorksheetAgentController`/`WizVsService`).
- `modify(...)` now calls `securityEngine.checkPermission(agent, ResourceType.VIEWSHEET_CALCULATED_FIELD, "*", ResourceAction.ACCESS)` as its very first action, before table/name validation, model building, or `sessions.mutate`, and throws `inetsoft.sree.security.SecurityException` on denial — the checked-exception type this package's own gates use (not `java.lang.SecurityException`, which only the native controller shadows locally).
- `BindingAgentController` needed no change — it only delegates to `calcFieldService.modify(...)` and mocks the whole service in its own tests.

**Tests:** Updated `CalcFieldAgentServiceTest` (added a `SecurityEngine` param to every existing constructor call, defaulting to an allow-all mock) and added `refusesWhenCalculatedFieldPermissionIsDenied`, which asserts a denied check throws `SecurityException` before touching the proxy, the session service, or the fields service (`verifyNoInteractions` on all three).

Ran:
```
./mvnw -pl core -Dtest=CalcFieldAgentServiceTest,BindingAgentControllerTest test -Dsurefire.failIfNoSpecifiedTests=false
```
Result: `Tests run: 27, Failures: 0, Errors: 0` — 14/14 `BindingAgentControllerTest` (unchanged), 13/13 `CalcFieldAgentServiceTest` (12 existing + 1 new regression test).

PR #4955, opened 2026-09-02
fix(wiz): infer numeric type for pure-arithmetic expression columns (PWA-002)

larryliang-inetsoft:
## Summary

`add_expression_column`/`edit_expression` silently defaulted an omitted `type`
to `"string"` -- no inference was attempted anywhere on the call path, so
`ExpressionRef.getDataType()` fell through to `AbstractDataRef`'s unconditional
`XSchema.STRING` default, even for a purely-arithmetic expression over numeric
columns (e.g. `field['QUANTITY'] * field['PRICE']`).

## Root cause

`WorksheetMutationSupport.addExpressionColumn`/`editExpression` both had
`if(type != null) { ...setDataType(type); }` with no `else` -- when the caller
(the plugin, when the LLM omits `type`) sends no type, nothing is ever set.

## Fix

Add `inferNumericExpressionType(TableAssembly, String)`: extracts every
`field['x']` reference from the expression, and if every referenced field
resolves to a numeric column *and* everything else in the expression is pure
arithmetic (digits, `+ - * /`, parens, whitespace), infers `XSchema.DOUBLE`.
Any other shape (string concatenation, comparisons, unresolvable/non-numeric
fields) is left alone, preserving the existing `"string"` default rather than
guessing -- matching the CLAUDE.md tool-misuse-is-a-plugin-gap principle
(forgiving on the unambiguous case, unchanged otherwise).

`editExpression`'s javadoc says `type == null` means "leave unchanged," so its
inference only fires when the column's *current* type is still the untyped
`"string"` fallback -- a real, previously-set explicit type (e.g. `"integer"`)
is never clobbered by re-inferring from a later expression edit.

## Test plan

- [x] Added 6 regression tests to `WorksheetEditServiceMutatorsTest`:
  - two-factor (`a * b`) and three-factor/parenthesized (the bug doc's "G9"
    shape) pure-numeric expressions infer `double` when `type` is omitted
  - a non-numeric/ambiguous expression (string concatenation over default
    string-typed columns) still defaults to `string`
  - an explicit `type` argument still wins over inference (no regression)
  - `edit_expression` infers numeric the same way when `type` is omitted
  - `edit_expression` leaves a real prior explicit type alone when `type` is
    omitted, even though the new expression would otherwise infer numeric
- [x] Verified the 3 inference-asserting tests fail without the production
  fix (temporarily stashed it) and pass with it restored
- [x] `mvnw -pl core test -Dtest=WorksheetEditServiceMutatorsTest` --
  231/231 passing

Ref: PWA-002 (debug-workflow batch, `docs/teams/2026-09-02-bugs-p0-p2-composer/bug-PWA-002/`)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetMutationSupport.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetMutationSupport.java
@@ -1335,6 +1335,13 @@ public final class WorksheetMutationSupport {
       if(type != null) {
          colRef.setDataType(type);
       }
+      else {
+         String inferred = inferNumericExpressionType(t, expression);
+
+         if(inferred != null) {
+            colRef.setDataType(inferred);
+         }
+      }
 
       ColumnSelection cs = t.getColumnSelection(false);
       cs.addAttribute(colRef);
@@ -1370,6 +1377,16 @@ public final class WorksheetMutationSupport {
 
                if(type != null) {
                   cr.setDataType(type);
+               }
+               else if(XSchema.STRING.equals(cr.getDataType())) {
+                  // Only re-infer when the column is still sitting on the untyped
+                  // fallback -- a real, previously-set explicit type (e.g. "integer")
+                  // must be left alone per the "null = leave unchanged" contract above.
+                  String inferred = inferNumericExpressionType(t, expression);
+
+                  if(inferred != null) {
+                     cr.setDataType(inferred);
+                  }
                }
 
                t.setColumnSelection(cs, false);
@@ -1436,6 +1453,18 @@ public final class WorksheetMutationSupport {
    }
 
    private static boolean isDateColumn(ColumnSelection cs, String field) {
+      DataRef ref = resolveColumn(cs, field);
+
+      if(ref == null) {
+         return false;
+      }
+
+      String dtype = ref.getDataType();
+      return XSchema.DATE.equals(dtype) || XSchema.TIME_INSTANT.equals(dtype) ||
+         XSchema.TIME.equals(dtype);
+   }
+
+   private static DataRef resolveColumn(ColumnSelection cs, String field) {
       DataRef ref = cs.getAttribute(field);
 
       if(ref == null) {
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
## Review: changes requested

Blocking bug found in `editExpression`'s new inference guard (`WorksheetMutationSupport.java:1381`): `else if(XSchema.STRING.equals(cr.getDataType()))` checks the *derived* type value instead of `ColumnRef.isDataTypeSet()`. This can't distinguish "type was never set" (the actual bug state) from "caller explicitly passed `type: \"string\"` on purpose" — so a later `edit_expression` call that omits `type` will silently clobber a real, deliberately-chosen `"string"` type to `"double"` if the new expression looks numeric. This directly contradicts the PR's own stated intent and `editExpression`'s documented "`type == null` leaves unchanged" contract.

Confirmed independently with a throwaway test (reverted after running): add with explicit `type:"string"`, then `edit_expression` omitting `type` → expected `string`, actually returns `double`. The PR's own "preserve explicit prior type" regression test only used `"integer"`, so it never exercised this branch.

**Fix**: guard on `!cr.isDataTypeSet()` instead, plus a new regression test with an explicit `"string"` prior type.

Separately, investigated the `ConditionTest` CI failure — confirmed it's a pre-existing, unrelated flake (added in PR #4899, 2026-08-30; same failure signature reproduces today on 3 unrelated branches). Not a blocker on its own.

Full findings: `docs/teams/2026-09-02-bugs-p0-p2-composer/bug-PWA-002/04-review-r1.md` in stylebi-wiz.

larryliang-inetsoft:
Fix round 2, responding to the review's blocking finding.

Changed the `editExpression` re-inference guard from `XSchema.STRING.equals(cr.getDataType())` to `!cr.isDataTypeSet()` — the old guard compared the *derived* type value (which is `"string"` both when never set and when explicitly set to `"string"`), so an `edit_expression` omitting `type` could silently clobber a deliberate `type: "string"` to `"double"`. `isDataTypeSet()` checks the underlying field directly and distinguishes the two cases correctly.

Added `editExpressionPreservesExplicitStringPriorTypeWhenTypeOmitted` (mirrors the existing `"integer"` test but with `"string"`), which fails with the exact `expected: <string> but was: <double>` signature from the review's throwaway test when run against the old guard, and passes with the fix.

Full class: `Tests run: 232, Failures: 0, Errors: 0, Skipped: 0` — 231 previous + 1 new, all pass. Mutation-checked by stashing the production change: the new test fails as expected, the other two related tests (integer-prior-type, infers-when-omitted) still pass.

Commit: 53bddc99b3. Details in `docs/teams/2026-09-02-bugs-p0-p2-composer/bug-PWA-002/05-fix-r2.md`.

PR #4956, opened 2026-09-02
Bug #76412, stabilize Ignite cluster membership changes

ngovus-inetsoft:
Fixes the community-side causes behind Bug #76412 (AKS Ignite cluster instability on multi-node topology changes).

The report attributed all of its symptoms to Ignite's `default-ds-group` partition inconsistency. Code review says otherwise — each symptom has an independent cause in our own code, and none of them require the Ignite bug to be true.

## Divergent JWT/SSO signing keys (fresh multi-replica cold start)

`LocalPasswordEncryption.getJwtSigningKey()` took the cluster lock and then read `SreeEnv.getProperty()`, which serves a node-local snapshot refreshed by a debounced storage listener. A node could hold the lock, read a stale `null`, and generate a second key. The lock was working; the check inside it was not.

Reads are now **storage-first**, with the in-memory snapshot only as a fallback. Storage-first rather than storage-as-a-null-fallback matters because `PropertiesEngine` fires `firePropertyChange` *before* its debounced snapshot refresh — so on a re-read the snapshot still holds the old key and a null-only fallback would be dead code. Same change in `getSSOKeyPair()` and `FipsPasswordEncryption.getPasswordHashKey()`.

`PropertiesEngine.getPropertyFromStorage()` also swallowed storage errors and returned `null`, indistinguishable from "no such key". Every caller uses it to decide whether to generate a key, so a transient read failure during exactly the instability this fixes could silently mint a divergent one. It now fails closed. That also covers the pre-existing caller for `password.encryption.key`, where a second key would render every stored password undecryptable.

## `affinityCall` hangs and unusable errors

- `EVT_NODE_LEFT` was not handled at all, so an in-flight call to a departing node blocked the caller for the full hard-coded 5 minutes.
- `EVT_NODE_FAILED` was never registered — and that is the event a crashed, OOMKilled or partitioned pod actually fires, i.e. the reported scenario.
- Send failures were only logged, then fell into the same wait.
- Errors surfaced as a bare `RuntimeException(TimeoutException)` with no cache/key/node; remote exceptions were double-wrapped; the cause `switch` NPE'd on a null cause.

Pending calls now track their recipient and are failed on node departure, on send failure and on `close()`. The timeout is configurable, remote exceptions keep their own type behind a null-cause guard, and messages carry the cache, key and node. The new `AffinityCallException` extends `MessageException`, so a transient topology failure reaches the user as a retriable message rather than a bare 500.

## Affinity pool starvation

Two *different* thread pools shared the name `IgniteMessages`, both sized to the CPU count. The affinity one runs service code that can nest another `affinityCall` — a self-deadlock on a small pod, behind an unbounded queue that hid the backlog. It is now growable, separately named `IgniteAffinity`, answers saturation with an error response, and no longer holds a pool thread during the Spring-readiness wait.

## `<user> did not log in` under cross-node routing

The cluster-wide logged-in principal map was keyed by `ClientInfo.getCacheKey()`, whose `equals`/`hashCode` include the client address and locale — both of which change when a principal is serialized to another node. The *value* comparison had already been relaxed for exactly this reason, with a comment saying so; the **lookup key** had not, so control never reached the relaxed comparison.

Lookups now use a login key of user identity + session. The address is retained when there is no session id, so sessionless entries stay distinct. Writes go to both keys and reads fall back to the legacy key, so a rolling upgrade works in both directions.

## Deliberately unchanged

`DistributedLockProxy.lock()` keeps its blocking behavior. Retrying while the lock is merely held elsewhere is the `Lock` contract, and all 34 cluster-lock callers use the blocking `try/finally` idiom with no handling for a failure to acquire — bounding it would turn ordinary contention into a hard failure. Only a periodic warning was added.

`memberRemoved` still fires only for a graceful `EVT_NODE_LEFT`. That listeners are never notified when a node crashes looks like a separate pre-existing gap; widening it here would change behavior for every listener at once.

## Testing

New: `DistributedLockProxyTest` (asserts the blocking contract), `IgniteClusterAffinityErrorTest` (exception mapping, including the null-cause NPE). Extended: `SecurityEngineAuthenticationTest` (cross-node address/locale change resolves; a different session does not; legacy-key entries still resolve), `FipsPasswordEncryptionTest` (stale-snapshot read — verified to fail without the fix, generating a second distinct key).

Targeted `core` sweep matches the pre-change `main` baseline exactly (same failures, same classes — all pre-existing).

## Not addressed here

- The `DeadlockHealthIndicator` hit is a genuine JVM lock cycle. These changes make the node degrade instead of wedge, but pinning the cycle needs the `StatusDumpService` dump the indicator already wrote.
- The `default-ds-group` partition mismatch: `AtomicConfiguration` derives from `minReplicas` like the app caches, so changing it requires a full cluster restart, not a rolling update. Filed separately.
- The MV `DefaultBlockSystem` blob-rename error during rollouts. Separate concern.

### DIFF
--- a/core/src/main/java/inetsoft/sree/ClientInfo.java
+++ b/core/src/main/java/inetsoft/sree/ClientInfo.java
@@ -82,6 +82,30 @@ public class ClientInfo implements Cloneable, Serializable, XMLSerializable {
    }
 
    /**
+    * Gets a key identifying this user's login, for use as the key of the cluster-wide map of
+    * logged-in principals.
+    *
+    * <p>Unlike {@link #getCacheKey()} this omits the locale, and omits the client address when a
+    * session id is present. Both are part of {@link #equals(Object)} and {@link #hashCode()}, and
+    * both can differ between the node where the user logged in and a node that receives the same
+    * principal after it has been serialized across the cluster. A real login is already uniquely
+    * identified by the user identity plus the session id, so including the other two fields only
+    * makes the lookup miss and the user appear to be logged out.
+    *
+    * <p>The address is retained when the session id is empty. Several callers build a
+    * {@code ClientInfo} with no session at all (the two-argument constructor normalizes it to
+    * {@code ""}); for those the address is the only thing distinguishing two otherwise identical
+    * entries, and dropping it would collapse them into one — so one caller would receive another's
+    * principal, and a logout by either would deregister both.
+    *
+    * @return the login key.
+    */
+   public ClientInfo getLoginKey() {
+      return new ClientInfo(userID, session == null || session.isEmpty() ? addr : null,
+                            session, null);
+   }
+
+   /**
     * Returns the user name.
     *
     * @return the user name.
--- a/core/src/main/java/inetsoft/sree/PropertiesEngine.java
+++ b/core/src/main/java/inetsoft/sree/PropertiesEngine.java
@@ -518,16 +518,35 @@ public class PropertiesEngine {
    /**
     * Reads a property directly from the backing key-value storage, bypassing the in-memory
     * cache. Use this when a stale null from the cache would cause irreversible side effects.
+    *
+    * <p>A failure to read the store is propagated rather than reported as an absent value. Every
+    * caller uses this method to decide whether a key already exists before generating a new one,
+    * and "storage is unreadable" must not be mistaken for "storage is confirmed empty" — that
+    * mistake is what lets two nodes each generate a different key. Failing here is recoverable;
+    * silently minting a second signing or encryption key is not.
+    *
+    * @param name the name of the property.
+    *
+    * @return the stored value, or {@code null} if the store confirms there is no such property.
+    *
+    * @throws RuntimeException if the store could not be read.
     */
    public String getPropertyFromStorage(String name) {
       name = fixPropertyNameCase(name);
+      KeyValueStorage<String> storage = kvStorage;
+
+      if(storage == null) {
+         throw new IllegalStateException(
+            "Cannot read property " + name + " from storage: the property storage is not " +
+            "initialized yet.");
+      }
 
       try {
-         return kvStorage.get(name);
+         return storage.get(name);
       }
       catch(Exception e) {
-         LOG.debug("Failed to read property from storage: {}", name, e);
-         return null;
+         LOG.warn("Failed to read property from storage: {}", name, e);
+         throw new RuntimeException("Failed to read property from storage: " + name, e);
       }
    }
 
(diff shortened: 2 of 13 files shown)


larryliang-inetsoft (changes requested):
Claude's automated review raises an open question rather than a clean approval: `LocalPasswordEncryption.getSecretKey()` (for `password.encryption.key`, arguably the highest-stakes of the four keys touched here) was not updated to the storage-first read pattern applied to the other three keys, and it's unclear whether that was intentional. Please confirm/address before this can be merged.

jshobe-inetsoft (changes requested):
Claude's automated review found this well-scoped overall, but flagged one unresolved correctness question that hasn't been addressed: `LocalPasswordEncryption.getSecretKey()` (for `password.encryption.key`) was not updated to the storage-first read pattern applied to the other three keys (`getJwtSigningKey()`/`getSSOKeyPair()`/`FipsPasswordEncryption.getPasswordHashKey()`), even though the PR's own reasoning suggests this is the highest-stakes of the four keys (a stale in-memory read here could make stored passwords undecryptable). Please confirm whether this was intentional, and if not, apply the same storage-first fix to `getSecretKey()` for consistency. Requesting changes until this is addressed.

jshobe-inetsoft (changes requested):
The `build` check on inetsoft-core is unstable on this PR and I don't think it's safe to dismiss as ordinary CI flakiness given what this PR touches:

- The original run failed in ~7 min with real test errors, including `FipsPasswordEncryptionTest$SymmetricEncryption.getSecretKey_staleInMemoryValue_readsStorageInsteadOfSnapshot` ("Runtime Failed to unwrap encryption key") — this is the new test this PR itself added to guard the `getSecretKey()` storage-first fix, and it didn't skip via its `Assumptions.assumeTrue` guard as intended, it errored. (The other two failures in that run, `ImageVSAScriptableTest`/`ViewsheetScopeTest` external image-URL fetch errors, look like unrelated pre-existing CI flakiness — I've seen the identical failures on other unrelated PRs.)
- I reran the build and this time it hung for 45 minutes on inetsoft-core (stuck right after `DashboardRegistryOrgLifecycleTest` started, no further test output) and was cancelled by the CI timeout rather than completing.

Given this PR's core subject is distributed locking/cluster membership stability (`DistributedLockProxy`, `IgniteCluster`, affinity call pooling), an intermittent hang plus a failing test in the same run is a signal worth taking seriously rather than attributing to environment flakiness. Could you:
1. Investigate whether `getSecretKey_staleInMemoryValue_readsStorageInsteadOfSnapshot` can genuinely fail/hang under this environment's backing store instead of skipping, and
2. Run the inetsoft-core suite a couple more times locally (or in CI) to check whether the 45-minute hang reproduces, before this is merged.

Claude's automated review itself found no blocking correctness issues in the code — this is purely about the build stability signal.

jshobe-inetsoft (approved):

PR #4957, opened 2026-09-02
PCB-004: establish crosstab/table source on set_table_fields/add_table_field

larryliang-inetsoft:
## Summary

- Fixes bug #76350 (PCB-004): `set_table_fields`/`add_table_field` never resolved or established a source for a sourceless crosstab/table assembly. The write reported success and the assembly rendered completely empty, with no error naming the missing source.
- `BindingAgentController.setTableFields`/`addTableField` now call `resolveSourceTable` (the same helper the three chart shelf endpoints already use), instead of the plain `requireBindableColumns` check. `TableShelfRequest`/`TableFieldRequest` gain a `table` field, mirroring `ShelfRequest`.
- `TableBindingService.setShelf`/`addField` gain a `sourceTable` parameter and an `applySource` helper (mirroring `ChartBindingService.applySource`): if the model has no source yet and a source was resolved, it's established as part of the same write. The existing no-source refusal (`requireSourceForFieldWrite`, from #4924) now checks the model's source *after* `applySource` runs, rather than the assembly's stored source — necessary so establishing a source and using it can happen in one call, rather than requiring `set_table_source` first every time.
- This complements, not duplicates, #4924 (which added the no-source refusal itself, based on the assembly's already-stored source, for the "no source and no table given" case). This PR adds the ability to actually resolve/establish a source in the same call.

## Test plan

- [x] `TableBindingServiceTest`, `BindingAgentControllerTest` (core module) — 41 tests pass
- [x] Full `inetsoft.web.wiz.binding` package — 559 tests pass, no regressions
- Ran via `./mvnw -pl core -am test -Dtest=... -Dsurefire.failIfNoSpecifiedTests=false -o`

Companion plugin PR: inetsoft-technology/stylebi-wiz — adds the `table` param to `set_table_fields`/`add_table_field` in `plugin/composer` and forwards it to this endpoint.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java
@@ -381,13 +381,16 @@ public class BindingAgentController {
                                   request.measure(), linkUri);
    }
 
-   public record TableShelfRequest(String assembly, String shelf, List<FieldRef> fields) {}
+   /** @param table see {@link ShelfRequest#table()}. */
+   public record TableShelfRequest(String assembly, String shelf, List<FieldRef> fields,
+                                   String table) {}
 
    /** {@code force} discards fields already bound to the old source; absent means false. */
    public record ChartSourceRequest(String assembly, String table, Boolean force) {}
    public record TableSourceRequest(String assembly, String table, Boolean force) {}
+   /** @param table see {@link ShelfRequest#table()}. */
    public record TableFieldRequest(String assembly, String shelf, FieldRef field,
-                                   Integer position) {}
+                                   Integer position, String table) {}
    public record TableRemoveRequest(String assembly, String shelf, String column) {}
    public record TableMoveRequest(String assembly, String fromShelf, String toShelf,
                                   String column, Integer position) {}
@@ -409,9 +412,10 @@ public class BindingAgentController {
       throws Exception
    {
       requireEnabled();
-      requireBindableColumns(sessionToken, user, request.assembly(), request.fields());
+      String source = resolveSourceTable(sessionToken, user, request.assembly(), request.table(),
+                                         request.fields());
       tableService.setShelf(sessionToken, user, request.assembly(), request.shelf(),
-                            request.fields());
+                            request.fields(), source);
    }
 
    public record SelectionSourceRequest(String assembly, String table, List<String> columns,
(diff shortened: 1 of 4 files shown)


larryliang-inetsoft:
Companion plugin PR: inetsoft-technology/stylebi-wiz#2111

larryliang-inetsoft:
## Review (r1)

**Code content: sound.** Verified the actual intended diff (4 files, +239/-67) against the correct base — see below. `TableBindingService.applySource`/`requireSourceForFieldWrite` correctly refuse any non-empty field write on a still-sourceless model regardless of path (no `table` given + unresolvable columns → `resolveSourceTable` throws; no `table` given + listing read failure → `resolveSourceTable` returns null, `requireSourceForFieldWrite` refuses; `table` given or columns unambiguously resolve → source established in the same call). Confirmed no other callers of `TableBindingService.setShelf`/`.addField` exist, and `removeField`/`moveField` correctly untouched (they only ever touch already-bound fields, so don't need this).

The disputed test rewrite (`setTableFieldsRefusesWhenAssemblyHasNoSource`) is legitimate, not a cover-up: I confirmed `resolveSourceTable`'s only remaining `null`-returning path is a genuine listing-read failure (its try/catch swallows a `RuntimeException` from `fields.list(...)`, distinct from `BindableColumns.requireSource`'s thrown `IllegalArgumentException` for a stated-but-missing/ambiguous table). The old single-resolvable-candidate stub is correctly expected to now succeed by design — that's the fix working, not a regression being hidden.

The three unrelated tests fixed by giving their models an explicit `existing.setSource(...)` are legitimate — they test basic shelf/field-write mechanics that presuppose an existing binding, unrelated to the sourceless case, and only broke because the refusal check's basis necessarily moved from `assembly.getSourceInfo()` to `model.getSource()` (required so `applySource` and the refusal see the same in-call mutation).

## Blocking issue found and fixed

**This PR was opened against the wrong base branch.** It targeted `main`, but every other wiz-agent PR (including #4924, the prior PCB-004 fix this one builds on) targets `stylebi-wiz-main-rebase` — the actual integration branch; the entire `inetsoft.web.wiz` package doesn't exist on `main` at all. Comparing against `main` produced a 1439-file / ~6200-commit diff (GitHub's own diff endpoint refused it as "exceeded maximum number of files"), which is not something that should ever be merged as literally shown.

I've retargeted the base to `stylebi-wiz-main-rebase` via `gh pr edit`. It now correctly shows 4 changed files and `MERGEABLE`. **The `build` check still shows failed, but that run happened against the wrong base — it needs to be re-triggered now that the base is corrected before this can be evaluated for merge.** (I don't have permission to trigger the re-run from here.)

I share the PR author's identity, so I can't self-approve/merge this. Flagging for the lead: once CI is re-run clean against the corrected base, this is ready to merge on the actual code content.

PR #4958, opened 2026-09-02
PC-007: normalize date_in named-group condition to a range shape

larryliang-inetsoft:
## Summary

- Fixes bug #76350 (PC-007): a `date_in`-based named group (`add_named_group` with a `DATE_IN` operation) always rendered `field IN ()` -- an unconditionally-empty match, silently, for every such group.
- Root cause: `ConditionUtil.fromModelToConditionList()`'s `DATE_IN` branch clones a `DateCondition` straight into the named group's `ConditionList` as a single item. `DateCondition` is a sibling of `Condition` (both extend `AbstractCondition`, neither is the other), so `NamedRangeRef.getExpression()`/`getScriptExpression()` fall through `ConditionItem`'s deprecated `getCondition()` accessor to a blank `EQUAL_TO` condition with zero values for every such group.
- Fix: `NamedGroupInfoModel.createNamedGroupInfo()` now normalizes the resolved `DateCondition` into the plain `GREATER_THAN(>=)/AND/LESS_THAN(<=)` range shape `NamedRangeRef` already knows how to render (`normalizeDateInGroupCondition()`), at the point the group's `ConditionList` is constructed -- instead of teaching `NamedRangeRef`'s two independent renderers about `DateCondition`/`BETWEEN` shapes.

## Test plan

- [x] `NamedGroupInfoModelTest` (new) -- 5 tests covering `normalizeDateInGroupCondition()`: date_in-resolved single-item list expands to a 3-item GT/AND/LT list with correct values, timestamp vs. non-timestamp field handling, no-op on a plain/already-multi-item condition, null-safety.
- [x] `CrosstabNamedGroupEndToEndTest` (existing) -- extended with `groupsByDateInBasedNamedGroupCondition()`, confirming the normalized 3-item shape actually groups correctly once it reaches the real query engine (`CrosstabVSAQuery`), not just that the shape itself is correct. Existing tests (plain OR-joined `EQUAL_TO` conditions) confirmed unaffected -- `normalizeDateInGroupCondition()` only fires on a single-item `DateCondition`-wrapped list.
- [x] `inetsoft.web.binding.**` (core module) -- 30 tests pass, no regressions.
- [x] `inetsoft.report.composition.execution.**` (core module) -- 82 tests pass, no regressions.
- Ran via `./mvnw -pl core -am test -Dtest=... -Dsurefire.failIfNoSpecifiedTests=false -o`

### DIFF
--- a/core/src/main/java/inetsoft/web/binding/model/NamedGroupInfoModel.java
+++ b/core/src/main/java/inetsoft/web/binding/model/NamedGroupInfoModel.java
@@ -235,6 +235,14 @@ public class NamedGroupInfoModel implements Serializable {
       }
 
       DateCondition dateCondition = (DateCondition) item.getXCondition();
+
+      if(dateCondition.isNegated()) {
+         // Negation can't be expressed by the GREATER_THAN/LESS_THAN range shape below
+         // (that would require a NOT/OR-of-two-ranges shape NamedRangeRef doesn't support) -
+         // leave the list alone rather than silently producing the wrong-signed range.
+         return;
+      }
+
       DataRef attribute = item.getAttribute();
       // isTimestamp must reflect the bound field's actual data type, not the
       // DateCondition's own type (every DateCondition ctor hardcodes type = XSchema.DATE),
--- a/core/src/main/java/inetsoft/web/binding/model/NamedGroupInfoModel.java
+++ b/core/src/main/java/inetsoft/web/binding/model/NamedGroupInfoModel.java
@@ -20,11 +20,16 @@ package inetsoft.web.binding.model;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import inetsoft.report.internal.binding.*;
 import inetsoft.uql.Condition;
+import inetsoft.uql.ConditionItem;
 import inetsoft.uql.ConditionList;
+import inetsoft.uql.JunctionOperator;
+import inetsoft.uql.XCondition;
 import inetsoft.uql.asset.AssetRepository;
+import inetsoft.uql.asset.DateCondition;
 import inetsoft.uql.asset.SNamedGroupInfo;
 import inetsoft.uql.asset.internal.AssetUtil;
 import inetsoft.uql.erm.DataRef;
+import inetsoft.uql.schema.XSchema;
 import inetsoft.uql.util.XNamedGroupInfo;
 import inetsoft.util.Tool;
 import inetsoft.web.binding.service.DataRefModelFactoryService;
@@ -180,8 +185,10 @@ public class NamedGroupInfoModel implements Serializable {
                   continue;
                }
 
-               expertInfo.setGroupCondition(group, ConditionUtil.fromModelToConditionList(
-                  item.getList(), null, null, null));
+               ConditionList conditionList = ConditionUtil.fromModelToConditionList(
+                  item.getList(), null, null, null);
+               normalizeDateInGroupCondition(conditionList);
+               expertInfo.setGroupCondition(group, conditionList);
             }
          }
          else if(getGroups() != null) {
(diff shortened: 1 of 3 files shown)


larryliang-inetsoft:
Verified the diagnosis/refute/fix independently. Base branch, file scope, and CI are all correct
(mergeable, all checks green). The core normalization logic, date-boundary math (inclusive/inclusive
via `isEqual(true)` on both `GREATER_THAN`/`LESS_THAN`), and the two test suites are sound.

**Blocking finding**

`NamedGroupInfoModel.normalizeDateInGroupCondition()` never reads/propagates
`dateCondition.isNegated()` (or `range.isNegated()`, which `DateCondition.createSqlCondition()`
sets from the same flag) onto the new `start`/`end` conditions.

This is reachable, not theoretical: `ConditionUtil.fromModelToConditionList()` (:313-316) calls
`condition.setNegated(conditionModel.isNegated())` unconditionally, and the classic composer's
Named Groups dialog (`grouping-condition-item-pane-provider.ts:133-136`, `isNegationAllowed()`)
explicitly allows negation for `DATE_IN` (only excludes `TOP_N`/`BOTTOM_N`). So a user can build a
"NOT this year" date-based named group today.

Before this fix, a negated `date_in` group was *also* broken (same always-empty `IN ()` fallback).
After this fix, the non-negated case is correctly fixed, but the negated case silently renders as
the **non-negated** range instead of being excluded -- a plausible-looking but wrong result, which
is worse than the prior always-empty failure (harder to notice). Not covered by either test file.

Requested fix: bail out of normalization when `dateCondition.isNegated()` is true (leave the
single-item list alone, matching this method's existing "no-op on shapes I cannot safely handle"
pattern for other cases), plus a regression test for `isNegated()==true`. Alternatively, correctly
invert the produced range if that shape support is added to `NamedRangeRef`.

**Non-blocking notes**
- Neither test drives the real `ConditionModel(operation=DATE_IN) -> fromModelToConditionList() ->
  normalizeDateInGroupCondition()` wiring end-to-end (both hand-build one side of the seam). Worth
  a follow-up test, not blocking given the unit test's direct coverage of the normalization logic.
- `CrosstabNamedGroupEndToEndTest.runDate()` duplicates most of the existing `run()` helper.
  Cosmetic.

Full evidence trail: `docs/teams/2026-09-02-bugs-p0-p2-composer/bug-PC-007/04-review-r1.md` in
stylebi-wiz.

larryliang-inetsoft:
## Fix round 2, responding to review r1's blocking finding

Applied option (a) as recommended: `normalizeDateInGroupCondition()` now bails out (no-op, list left untouched) when `dateCondition.isNegated()` is true, rather than silently rendering the non-negated range for a negated `date_in` group.

**Change**: `NamedGroupInfoModel.java`, added an early-return right after resolving `dateCondition`:
```java
if(dateCondition.isNegated()) {
   return;
}
```
Same style as the method's existing guards. Preserves the pre-fix conservative "renders empty" behavior for the negated case instead of producing a wrong-signed range.

**Test added**: `NamedGroupInfoModelTest.normalizeDateInGroupCondition_negatedCondition_isNoOp()` -- builds a single-item list wrapping a negated `DateCondition`, confirms the method leaves it untouched (asserted via `getXCondition()`, not the deprecated `getCondition()` accessor which would return a blank `Condition` and mask the assertion).

**Test evidence**:
```
./mvnw -pl core -am test -Dtest=NamedGroupInfoModelTest,CrosstabNamedGroupEndToEndTest -Dsurefire.failIfNoSpecifiedTests=false -o
```
`Tests run: 9, Failures: 0, Errors: 0, Skipped: 0` -- all pass (6 unit tests incl. the new one, 3 E2E tests unaffected).

Pushed as `b4db7b4db0` on the same branch/PR. Did not address the two non-blocking notes (untested ConditionModel-to-normalizer wiring seam; near-duplicate `runDate()`/`run()` helper) -- happy to take those as a follow-up if wanted.

PR #4959, opened 2026-09-02
Bug #76410, refresh the cached provider list so a new provider loads in the Users tab

ngovus-inetsoft:
Redmine: Bug #76410 — new/copied provider cannot load in Users tab

## Problem

Create or copy an authentication provider in EM → Settings → Security, then switch to the **Users** tab: the new provider cannot be loaded until you log out and log back in.

The Users tab does not fetch the provider list itself — it renders the array cached on `OrganizationDropdownService`, a root singleton populated once per page load (a one-shot `getEmCurrentUser()` subscription plus a STOMP handler on `/user/security/providers-change`). None of the provider mutations refreshed it: `copyProvider()`, `removeProvider()` and `SecurityProviderService.updateAuthenticationProvider()` all skipped `orgDropdownService.refreshProviders()`, which only `reorder()` called.

So the tab kept showing the list as it was when the EM page loaded, and `getProvider()` could still name a provider that had since been renamed or removed. Selecting it sent that dead name to `get-security-tree-root`, where `SecurityTreeServer` dereferenced the null result of `getProviderByName()`:

```
GET /api/em/security/user/get-security-tree-root/testdb/false  ->  500

java.lang.NullPointerException: Cannot invoke
  "inetsoft.sree.security.AuthenticationProvider.getOrganizationIDs()" because "provider" is null
    at inetsoft.web.admin.security.user.SecurityTreeServer.getSecurityTree(SecurityTreeServer.java:88)
    at inetsoft.web.admin.security.user.RoleController.getSecurityTreeRoot(RoleController.java:77)
```

Only a full page reload — i.e. logging out and back in — cleared the cache.

## Fix

- **Refresh the cached list after every provider mutation** — copy and remove in `AuthenticationProviderViewComponent`, and add/edit (including rename) in `SecurityProviderService.updateAuthenticationProvider()`.
- **`UsersSettingsPageComponent` no longer trusts the cache on init.** `ngOnInit` calls `refreshProviders()` instead of loading the tree eagerly from the cached name in the constructor; the initial tree load is driven by the resulting `onRefresh` emission, whose provider name has just been validated against the server's list. The tab therefore cannot request a provider the server no longer knows, even if a future mutation path forgets to refresh. Also tracks and tears down the `onRefresh` subscription, which previously outlived the component and kept a destroyed Users page issuing tree requests.
- **`SecurityTreeServer.getSecurityTree()` guards a null provider** and throws a `MessageException` naming it (new `em.security.provider.notFound` message) instead of NPEing. Every prior null path already NPE'd on the same dereference, so this only turns a 500 into a readable error; it also covers the cluster case where a node has not yet reloaded `authc-chain.json`.
- **`AuthenticationProviderService.messageReceived()` debounce key.** It keyed on the user session, so a user with more than one live subscription (two EM tabs, or a reconnected socket whose old subscriber had not been removed) had them collapse into a single debounce entry and only one arbitrarily-chosen subscriber was notified. The key now includes the subscriber's own STOMP session and subscription id. Hardening — the explicit `refreshProviders()` calls are what guarantee the list is current.

## Testing

Reporter confirmed the fix with a live test.

Automated:

- New `SecurityTreeServerTest` — both null-provider paths report `MessageException`, not NPE.
- Frontend specs — refresh on copy/remove, refresh on add/edit success but not on error, provider reload on Users tab init, no tree request before the list resolves, no reaction to `onRefresh` after destroy.
- Full EM suites green: `ng test em` 372 passed, `ng run em:test-tl` 842 passed, `ng lint em` 0 errors.
- Related Java tests pass: `RoleControllerTest`, `SecurityTreeControllerTest`, `ResourcePermissionTreeControllerTest`, `AuthenticationProviderControllerTest`, `EmPageHeaderControllerTest`.

Manual, on a multi-tenant install as a site admin, without reloading the page between steps:

1. Copy a provider, switch to Users — the copy is listed and its tree loads.
2. Add a provider (File and Database), switch to Users — same.
3. Rename a provider, switch to Users — the old name is gone and the tree loads under the new name.
4. Delete a non-current provider, switch to Users — it is gone from the dropdown.
5. Reorder still works; the page-header organization dropdown still populates after each of the above.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/security/AuthenticationProviderService.java
+++ b/core/src/main/java/inetsoft/web/admin/security/AuthenticationProviderService.java
@@ -895,8 +895,13 @@ public class AuthenticationProviderService extends BaseSubscribeChangeHandler {
          getSubscribers().stream()
             .filter(sub -> sub.getUser() instanceof XPrincipal)
             .forEach(sub -> {
-               this.debouncer.debounce(((XPrincipal) sub.getUser()).getSessionID(), 1L, TimeUnit.SECONDS,
-                                       () -> sendToSubscriber(sub));
+               // key on the subscriber, not just the user session: a user with more than one
+               // live subscription (two EM tabs, or a reconnected socket whose old subscriber
+               // has not been removed yet) would otherwise collapse into a single debounce
+               // entry and only one arbitrarily-chosen subscriber would be notified
+               String key = ((XPrincipal) sub.getUser()).getSessionID() + ":" +
+                  sub.getSessionId() + ":" + sub.getSubscriptionId();
+               this.debouncer.debounce(key, 1L, TimeUnit.SECONDS, () -> sendToSubscriber(sub));
             });
       }
    }
--- a/core/src/main/java/inetsoft/web/admin/security/user/SecurityTreeServer.java
+++ b/core/src/main/java/inetsoft/web/admin/security/user/SecurityTreeServer.java
@@ -65,6 +65,16 @@ public class SecurityTreeServer {
       final AuthenticationProvider provider = providerName == null ?
          securityProvider.getAuthenticationProvider() :
          authenticationProviderService.getProviderByName(providerName);
+
+      // getProviderByName() returns null when the name is not in this node's authentication
+      // chain -- e.g. a client that is still holding a provider name that has since been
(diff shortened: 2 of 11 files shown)


jshobe-inetsoft (approved):

PR #4960, opened 2026-09-02
Bug #76403, group all multi-selected assemblies when dragged into a container

jkidacki-inetsoft:
## Summary

When multiple filter assemblies (e.g. several Selection Lists) are multi-selected on the Composer canvas and dragged together onto a Selection Container, only one of them was actually added to the container — the rest of the selection was silently dropped from the operation. The same defect applies to dragging a multi-selection onto a Tab.

## Root Cause

Dragging an assembly onto a drop target on the viewsheet canvas is handled by `EditableObjectContainer.onDropzoneDrop()`. The interact.js drop gesture only identifies the single DOM element the pointer was tracking (`event.relatedTarget`), so only that one assembly's name was captured into `LayoutOptionDialogModel.object`. That model flows through `LayoutOptionDialogController` → `LayoutOptionDialogService.setLayoutOptionDialogModel()` → `GroupingService.groupComponents()`, which is a single-object API end to end — so even though the rest of a multi-selection visually moved along with the drag, only the one assembly under the pointer was grouped into the target.

## Fix

- Added an optional `additionalObjects: string[]` field to the `LayoutOptionDialogModel` (both the TypeScript interface and the Java DTO).
- In `onDropzoneDrop()`, a new helper (`getAdditionalDragSelectionNames`) collects the rest of `viewsheet.currentFocusedAssemblies` — excluding the primary dragged assembly, the target, and shape types, and (when the target is a Selection Container) restricted to filter-compatible types (`VSSelectionList` / `VSRangeSlider` / `VSSelectionContainer`) — and stores their names on the model.
- `LayoutOptionDialogService.setLayoutOptionDialogModel()` now loops `model.getAdditionalObjects()` after handling the primary object, calling `groupingService.groupComponents()` for each additional assembly against the same target, all within the same `@Undoable` transaction as the primary call (one atomic operation, one undo step).

## Test Plan

- [x] `core` module compiles cleanly (`mvnw test-compile -pl core -am`)
- [x] `portal` Angular project builds cleanly (`ng build portal`)
- [x] Relevant testing-library specs pass in isolation: `editable-object-container.component.{interaction,risk,display}.tl.spec.ts` (91 tests), `composer-selection-container-children.component.{interaction,risk}.tl.spec.ts` (30 tests), `vs-selection-container-children.component.tl.spec.ts` (18 tests)
- [x] ESLint clean on both changed TypeScript files
- [x] Manually verified in the Composer: multi-selecting several filters and dragging them onto a Selection Container now adds all of them

Fixes Redmine #76403.

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/model/vs/LayoutOptionDialogModel.java
+++ b/core/src/main/java/inetsoft/web/composer/model/vs/LayoutOptionDialogModel.java
@@ -85,12 +85,25 @@ public class LayoutOptionDialogModel implements Serializable {
       this.columns = columns;
    }
 
+   /**
+    * Names of additional assemblies that were part of the same multi-selection as
+    * {@link #getObject()} and should be grouped into the same target. (Bug #76403)
+    */
+   public List<String> getAdditionalObjects() {
+      return additionalObjects;
+   }
+
+   public void setAdditionalObjects(List<String> additionalObjects) {
+      this.additionalObjects = additionalObjects;
+   }
+
    @Override
    public String toString() {
       return "LayoutOptionDialogModel{" +
          " selectedValue=" + selectedValue +
          ", object=" + object +
          ", target=" + target +
+         ", additionalObjects=" + additionalObjects +
          '}';
    }
 
@@ -101,4 +114,5 @@ public class LayoutOptionDialogModel implements Serializable {
    private int newObjectType;
    private AssetEntry vsEntry;
    private List<AssetEntry> columns;
+   private List<String> additionalObjects;
 }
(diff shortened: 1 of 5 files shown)


jkidacki-inetsoft:
Addressed the two substantive findings:

1. **De-dup guard**: `LayoutOptionDialogService.setLayoutOptionDialogModel()` now also skips an `additionalObjects` entry that duplicates `model.getObject()`, not just the target — defense in depth independent of the front-end filter.
2. **Test coverage**: added `additionalObjectsAreGroupedTest` (verifies each additional assembly gets its own `groupComponents()` call against the same target) and `additionalObjectsSkipDuplicatesTest` (verifies the primary object and the target itself are not grouped twice) to `LayoutOptionDialogServiceTest`. All 3 tests in that class pass.

Skipped #3 (reusing an `isShape()`-style helper for the shape-type-list in `getAdditionalDragSelectionNames`) — it would require adding a parameter to the existing `isShape()` method and touching its other call sites, which is beyond this bug's scope; the duplication already exists 3 other times in this file.

jshobe-inetsoft (approved):

PR #4961, opened 2026-09-02
Bug #76413, force the chart legend "Others" bucket last when sort order is None/Original

jkidacki-inetsoft:
## Summary
- `VSDimensionRef.createComparator()` set `comparator = null` for `SORT_NONE`/`SORT_ORIGINAL` before the `sortOthersLast && isRankingGroupOthers()` check ran, so `OthersComparator` never wrapped it and "Others" wasn't forced last when the dimension's sort order was None/Original.
- Fix returns a standalone `OthersComparator` (instead of wrapping a null base comparator, which would NPE) when the base comparator is null and `sortOthersLast` applies.

## Test plan
- [x] New regression test `VSDimensionRefOthersLastTest` (5 cases) reproducing the exact reported order (`true, false, Others` → now `true, false, Others` sorted correctly with Others last)
- [x] Full `community/core` `@Tag("core")` suite: 4702 tests, 0 failures

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/VSDimensionRef.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/VSDimensionRef.java
@@ -915,8 +915,10 @@ public class VSDimensionRef extends AbstractDataRef implements ContentObject, XD
          throw new RuntimeException("Unsupported order found: " + order);
       }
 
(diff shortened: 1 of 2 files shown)


jkidacki-inetsoft:
Thanks for the review. Fixed the copyright year on the new test file (2025 → 2026) in 682a10a05.

jshobe-inetsoft (approved):

PR #4962, opened 2026-09-02
Bug #76424, track Shared Filters rows by column instead of index

jkidacki-inetsoft:
## Summary
- The Shared Filters table's `@for` loop tracked rows by `$index`. Removing a middle row let Angular reuse a surviving row's `NgModel`/DOM instance for a different filter object mid-reconciliation, aliasing its Filter ID `<input>` onto a `FormControl` still tied to a just-destroyed trailing row.
- Changed `track $index` → `track filter.column` (the filter's backing assembly name, confirmed unique per viewsheet and structurally duplicate-free within `sharedFilters`) so each row's identity follows its filter object rather than its screen position.

## Test plan
- [x] New DOM-level regression test reproducing the exact failure: remove the middle of four Shared Filter rows, assert no surviving row's Filter ID input shows another filter's value
- [x] Verified the test actually catches the bug (reverted to `track $index` → test failed with the exact reported corruption; restored fix → 27/27 pass)
- [x] Checked prior review history for this file's tracking expressions (dddb39231e / #75422) for an NG0955 concern — confirmed it doesn't apply here since `sharedFilters` can't contain duplicate `column` values

### DIFF
--- a/web/projects/portal/src/app/composer/dialog/vs/filters-pane.component.html
+++ b/web/projects/portal/src/app/composer/dialog/vs/filters-pane.component.html
@@ -60,7 +60,7 @@
                 </tr>
               </thead>
               <tbody>
-                @for (filter of model.sharedFilters; track $index; let _i = $index) {
+                @for (filter of model.sharedFilters; track filter.column; let _i = $index) {
                   <tr
                     [class.selected]="isSharedFilterSelected(_i)"
                     (click)="selectSharedFilter(_i, $event)">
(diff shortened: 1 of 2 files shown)


jkidacki-inetsoft:
Good catch — removed the dangling `bug-76424/02-refute.md` references from the test comments (that was an internal investigation scratch file, never part of this repo) and replaced them with self-contained explanations of the same detail. Pushed in cf84d8d4e.

jshobe-inetsoft (approved):

PR #4963, opened 2026-09-02
Bug #76409, surface errors instead of silently failing in Wizard calc-field editing

jkidacki-inetsoft:
## Summary
- `FormulaEditorDialog.populateColumnTree()`'s `getColumnTreeNode()` call had no error callback, so a failed tree-data request left the right-hand column/function tree pane silently empty with no visible error — reproduces when editing a calc field from the Wizard, which runs the shared Formula Editor code path against a separate, cloned `RuntimeViewsheet`.
- Applied the same defensive fix to `ObjectWizardToolBarComponent.openFullEditor()`, whose GET request also had no error callback, so a failure now surfaces an error dialog instead of behaving indistinguishably from a hang.

## Test plan
- [x] New `formula-editor-dialog.component.spec.ts` (2 tests): failed `getColumnTreeNode()` shows an error dialog and doesn't leave `columnTreeRoot` partially populated; success path unchanged
- [x] New test in `object-wizard-tool-bar.component.tl.spec.ts`: failed `openFullEditor()` request shows an error dialog, `onFullEditor` not emitted
- [x] Broader regression sweep of `widget/formula-editor/**` and `vs-wizard/**`: 5 files/8 tests (`.spec.ts`) + 16 files/449 tests (`.tl.spec.ts`), all passing

## Note
This fixes "the pane goes silently empty" / "the button silently does nothing." It does not fully explain a literal persistent loading spinner if one is observed separately, since that's driven by a different (STOMP-based) loading-mask mechanism this change doesn't touch — flagging in case further investigation is wanted there.

### DIFF
--- a/web/projects/portal/src/app/vs-wizard/gui/object-wizard-tool-bar.component.tl.spec.ts
+++ b/web/projects/portal/src/app/vs-wizard/gui/object-wizard-tool-bar.component.tl.spec.ts
@@ -45,7 +45,7 @@
 // which requires the JIT compiler when running outside the Angular builder pipeline.
 import "@angular/compiler";
 import { Component, Directive, Input, NO_ERRORS_SCHEMA } from "@angular/core";
-import { HttpClient, HttpParams } from "@angular/common/http";
+import { HttpClient, HttpErrorResponse, HttpParams } from "@angular/common/http";
 import { NgbModal } from "@ng-bootstrap/ng-bootstrap";
 import { render } from "@testing-library/angular";
 import { of, throwError } from "rxjs";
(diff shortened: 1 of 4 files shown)


jkidacki-inetsoft:
Good catch, thanks. Switched to `ComponentTool.showHttpError()`, matching the existing convention for raw `http.get` subscribe error callbacks elsewhere in the codebase — confirmed it unwraps the `HttpErrorResponse` into a readable string before showing the dialog. Also updated the regression test to mock the failure with a real `HttpErrorResponse` (via `throwError(new HttpErrorResponse({status: 500, error: "Server error"}))`) instead of a plain string, so it actually exercises the object-coercion path you flagged. Pushed in 76501dbe1.

jkidacki-inetsoft:
Build is failing due to an unrelated flaky test.

jshobe-inetsoft (approved):

PR #4965, opened 2026-09-02
Bug #76417, break relation chart node color ties by most-recent value

jkidacki-inetsoft:
## Summary
- `RelationElement.createGeometry()` sorts data by `[sourceDim, targetDim]` and picks a node's "representative row" (which drives its color lookup via `getSubRowIndex()`) as the first row after that sort. Rows tying on both keys — e.g. a source/target group spanning several quarters of the same year — were ordered arbitrarily, by whatever order the query happened to return them in, even though the field actually painting the node (bound via the general Color and/or Node Color aesthetic) played no role in breaking that tie.
- Added the color-bound field(s) as additional sort columns, applied in **descending** order (most recent value wins a tie) via a negated `DefaultComparator`, mirroring the existing pattern in `PointElement.sortData()`. No-op when there's no color field bound, no ties, or the field is already part of the sort key.
- Live-verified against the real `ST_Tree2` asset: the `Customer:Reseller` tree's "1" root node previously rendered brown (its representative row happened to land on `2017 1st`, the earliest of four tied quarters); after this fix it resolves to `2017 4th` (the most recent), matching the expected color.

## Test plan
- [x] New permanent test `RelationElementColorTieBreakTest` (5 cases): order-independence for both the general-Color and Node-Color cases, no-op behavior confirmed for no-ties/no-color-field/duplicate-field edge cases
- [x] Verified the fix is meaningful — reverted just the comparator change, watched the order-independence tests fail with the pre-fix (ascending/order-dependent) result, restored it
- [x] Full `community/core` suite: 4692 tests, 0 failures
- [x] Live-verified against the real ST_Tree2 asset and its original data

### DIFF
--- a/core/src/main/java/inetsoft/graph/element/RelationElement.java
+++ b/core/src/main/java/inetsoft/graph/element/RelationElement.java
@@ -136,6 +136,28 @@ public class RelationElement extends GraphElement {
       final Object mxroot = mxgraph.getDefaultParent();
       final int maxNodes = Integer.parseInt(SreeEnv.getProperty("graph.max.nodes", "1000"));
 
+      // A node's "representative row" (the row whose value drives its color lookup, see
+      // RelationGeometry.getColor()/getSubRowIndex()) must be resolved independently for
+      // root (source) and leaf (target) nodes, since RelationGeometry.getColor() consults
+      // a *different* frame depending on node type: a root node falls through to the
+      // general Color frame (getColorFrame()), while a leaf node consults the Node Color
+      // frame (getNodeColorFrame()) directly. A single shared tie-break (as previously
+      // implemented) would let one frame's field incorrectly decide the other node type's
+      // representative row whenever the two frames are bound to different fields. (76417)
+      //
+      // Root nodes keep the pre-existing structural precedence -- among all rows sharing a
+      // source (fromDim) value, the one with the earliest target (toDim) value wins -- and
+      // only break ties *within* that earliest-target group using the general Color field.
+      // Leaf nodes have no such secondary structural dimension (toDim alone already fully
+      // identifies a leaf), so all rows sharing a target value are eligible, and are
+      // resolved directly by the Node Color field. In both cases, "most recent" (largest)
+      // tie-break value wins, and a missing/unbound color field falls back to the first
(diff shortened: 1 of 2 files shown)


jkidacki-inetsoft:
Good catch, thanks — you're right that a single shared sort with a fixed priority order can misassign the wrong field's tie for whichever node type doesn't get priority. I verified the mechanism precisely: `RelationGeometry.getColor()` does read `elem.getColorFrame()` for root nodes, but indirectly — through `super.getColor(0)` → `VisualModel`, which is built from `getColorFrame()` in `GraphElement.createVisualModel()` — while leaf nodes consult `getNodeColorFrame()` directly. So the priority order matters exactly as you described whenever the two are bound to different fields.

Replaced the shared sort with two independent representative-row lookups — root nodes tie-broken by the general Color field, leaf nodes by Node Color — computed separately so each node type is always decided by the field that actually paints it. Added a test reproducing your exact scenario (two fields, tied rows where each field's "winner" differs) and confirmed it fails against the old implementation with the predicted mismatch (`expected: <B2> but was: <B1>`) before passing against the new one. Full `community/core` suite: 4722 tests, 0 failures. Pushed in 57ba31cd6.

jkidacki-inetsoft:
Good catch again — verified against \`RelationGeometry.getColor()\`'s source directly: the root-node fallback-to-general-Color branch is indeed gated on \`!isApplyAestheticsToSource()\`, so when that plot option is on, a root's color comes from Node Color directly. Fixed by selecting the root tie-break frame based on that same flag, mirroring \`getColor()\`'s own branching exactly. New test with \`setApplyAestheticsToSource(true)\` confirmed meaningful (fails with the predicted mismatch against the prior unconditional behavior, passes against this fix). Full \`community/core\` suite: 4723 tests, 0 failures. Pushed in 062e8f259.

jshobe-inetsoft (changes requested):

larryliang-inetsoft (changes requested):
Concurring with the existing review: Claude's automated review found no blocking issues, but the `build` check is currently failing. Please fix the build failure before this can be merged.

jkidacki-inetsoft:
The build is passing now

jshobe-inetsoft (approved):

PR #4966, opened 2026-09-02
Bug #76414, restore the asset-tree fallback when get-current-user fails

ngovus-inetsoft:
## Problem

`AssetTreeComponent.ngOnInit()` subscribes to `CurrentUserService.getPortalCurrentUser()` with a success-only callback:

```ts
this.subscriptions.add(this.currentUserService.getPortalCurrentUser().subscribe((user) => {
   this.currOrgID = user?.name?.orgID ?? "";
   this.loadAssetTree();
}));
```

If `../api/portal/get-current-user` fails — security provider not yet initialized on a cold start, a transient 5xx during cluster rebalancing, or a plain network blip — `loadAssetTree()` is never invoked and the asset tree stays blank for the life of the component instance. No retry, no message; the user has to reload the page.

`AssetTreeComponent` backs both the Composer asset-tree sidebar and the "New Dashboard" dialog picker, so both come up empty. `CurrentUserService` has no `catchError`, and no portal `HttpInterceptor` or global `ErrorHandler` catches this (`SsoHeartbeatInterceptor` is the only portal interceptor; `ErrorHandlerService` is EM-only and explicitly invoked).

## This is a regression of Bug #74027

`ab6bd25e0` (#74027) fixed exactly this symptom by adding an `error:` handler to the `forkJoin` that `ngOnInit()` used at the time. `8177443f1` (#74480) then replaced that `forkJoin` with `getPortalCurrentUser()` — but its diff was computed against the *pre-#74027* base blob (`index d8a8541b9..` in both commits), and the hunk resolution took its own side, deleting the error handler again.

## Fix

Both levels, because the service is the shared cause and the component is what regressed.

**1. `CurrentUserService`** — `getPortalCurrentUser()`/`getEmCurrentUser()` map a failed call to `null` instead of erroring, so a bare `subscribe()` in any consumer cannot silently skip its next callback. There are eight subscribers of this service and **none** supplied an error callback.

The `catchError` is applied per-subscriber **in the getters, not inside the shared pipe**. `shareReplay` passes `resetOnComplete: false`, and `share()` skips its refCount-zero reset once the source has completed (`share.js:61`), so a `catchError` ahead of `shareReplay` would complete the stream and cache the `null` for the lifetime of the page — every later subscriber would get no current user and never retry, which is worse than the bug being fixed. Keeping it in the getter caches only successful responses (preserving the de-duplication added in Bug #74395) while a failure still resets the buffer, so the next subscriber re-issues the request.

**2. `AssetTreeComponent.ngOnInit()`** — supplies an `error:` callback that calls `loadAssetTree()` with default org state, restoring the #74027 behavior. Safe because `currOrgID` is only consulted for the `== "SELF"` user-scope check. Redundant given (1), but it keeps the tree correct if `CurrentUserService` changes again — which is exactly how this regressed the first time.

**3. Two consumers were not null-safe** on the emitted model and would have thrown on the `null` fallback — `AppComponent` (em) dereferenced `userModel.name`, and `AiAssistantService` called `convertToKey(model.name)`. Both now guard.

Deliberately out of scope: no user-visible error toast or retry button. Rendering the default-org tree restores function, and #74027 established silent fallback as the accepted behavior for this path; adding a notification surface to a generic widget used both in the sidebar and inside a dialog is a separate UX decision.

## Testing

Both new error tests were run against the reverted source to confirm they are not vacuous:

| Spec | With fix | Without fix |
|---|---|---|
| `current-user.service.tl.spec.ts` (new) | 8 pass | **2 fail** (portal + em error paths) |
| `asset-tree.component.risk.tl.spec.ts` | 16 pass | **1 fail** — `getAssetTreeNode` called 0 times |

That last failure is the reported symptom reproduced directly: the tree load never fires. The caching tests pass in both states by design — they guard against the `shareReplay` caching regression described above rather than the original bug.

Also run: `ng lint` — 0 errors (8 pre-existing warnings, all in untouched files). `tsc -p projects/em/tsconfig.spec.json --noEmit` — clean.

Not covered by a test: the one-character optional chain in em's `app.component.ts`, as no spec exists for that component; it is covered by the type-check.

### DIFF
--- a/web/projects/em/src/app/app.component.ts
+++ b/web/projects/em/src/app/app.component.ts
@@ -84,7 +84,7 @@ export class AppComponent implements OnInit, OnDestroy {
 
    ngOnInit(): void {
       this.subscription.add(this.currentUserService.getEmCurrentUser().subscribe(userModel => {
-         this.name = userModel.name?.name;
+         this.name = userModel?.name?.name;
       }));
 
       this.subscription.add(this.authzService.getPermissions("").subscribe(p => this.permissions = p));
--- a/web/projects/portal/src/app/widget/asset-tree/asset-tree.component.risk.tl.spec.ts
+++ b/web/projects/portal/src/app/widget/asset-tree/asset-tree.component.risk.tl.spec.ts
@@ -31,11 +31,13 @@
  *   Group 7  [Risk 2] — ngOnDestroy: calls assetClientService.disconnect(); unsubscribes subscriptions
  *   Group 8  [Risk 1] — setupAssetClientService: calls assetClientService.connect();
  *                        subscribes to assetChanged
+ *   Group 9  [Risk 3] — current-user failure: the tree must still load when
+ *                        getPortalCurrentUser() errors or emits null (Bug #74027)
  */
 
 import { NO_ERRORS_SCHEMA } from "@angular/core";
 import { render } from "@testing-library/angular";
-import { NEVER, of, Subject } from "rxjs";
+import { NEVER, of, Subject, throwError } from "rxjs";
 
 import { AssetTreeComponent } from "./asset-tree.component";
 import { AssetTreeService } from "./asset-tree.service";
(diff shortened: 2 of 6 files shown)


jshobe-inetsoft (approved):

PR #4971, opened 2026-09-03
fix(wiz): ModifyCalculateFieldService NPEs for any session-less caller (PCB-006)

fredaan:
## Summary
- **Live-verified via a real connected composer-chat session** (not just code review): every call to the wiz agent's `add_calc_field`/`edit_calc_field`/`remove_calc_field` — i.e. every create, and every edit/remove of a calc field that actually exists — returned a generic 500 ("This is a StyleBI-side bug"). The server log's real stack trace: `NullPointerException: Ouch! Argument cannot be null: key` from Ignite's cluster-affinity routing, inside `ModifyCalculateFieldService.modifyCalculateField` → `VSBindingTreeController.getBinding` → `VSBindingTreeControllerServiceProxy.getBinding` → `WorksheetEngine.isLocal` → `RuntimeSheetCache.getAffinityKey`.
- **Root cause**: the method's non-wizard branch (refreshing the Composer's live binding-tree panel after a calc-field change) asked `VSBindingTreeController.getBinding(event, principal, dispatcher)` to do the refresh. That controller re-derives the runtime id from `RuntimeViewsheetRef` — a WebSocket-session-scoped bean the native Formula Editor dialog's own STOMP session populates — which is never set for a caller with no live WebSocket session, like the wiz agent surface. The re-derived id came back null, and the generated cluster-proxy's affinity-key routing (keyed on that id) throws before the real binding-tree read ever runs.
- **Fix**: `modifyCalculateField` already has the correct, verified-live runtime id as its own `@ClusterProxyKey id` parameter (used moments earlier in the same method to resolve the `RuntimeViewsheet` in the first place). Calling `VSBindingTreeControllerServiceProxy.getBinding(id, ...)` directly with it, instead of asking `VSBindingTreeController` to re-derive a broken one, fixes this for any caller — and is behaviorally identical for the native WebSocket flow, since `RuntimeViewsheetRef.getRuntimeId()` and this `id` are the same value there.
- Adds `ModifyCalculateFieldServiceTest` — this method had **zero** existing unit test coverage, which is presumably why this was never caught: every existing wiz-agent test for this area (`CalcFieldAgentServiceTest`) mocks `ModifyCalculateFieldServiceProxy` entirely, so the real method body — including this exact call — never actually ran in any test suite before.

## Severity
This affects **every real create/edit/remove** through the new `add_calc_field`/`edit_calc_field`/`remove_calc_field` tools (PCB-006, stylebi-wiz#2107 + stylebi#4951) — only a `remove_calc_field` on an *already-nonexistent* field name (a degenerate no-op, since the method early-returns before reaching this code) was unaffected. Blocking for that feature actually working end to end.

## Test plan
- [x] `ModifyCalculateFieldServiceTest` (new) — exercises the real method body end-to-end for a create, asserts the binding-tree refresh receives the correct non-null id. 1/1 pass.
- [x] Confirmed via a live paired composer-chat session against a real running StyleBI instance, with the server's own log showing the exact NPE stack trace before this fix.

### DIFF
--- a/core/src/main/java/inetsoft/web/binding/controller/ModifyCalculateFieldService.java
+++ b/core/src/main/java/inetsoft/web/binding/controller/ModifyCalculateFieldService.java
@@ -65,7 +65,7 @@ import java.util.*;
 public class ModifyCalculateFieldService {
    public ModifyCalculateFieldService(
       VSBindingService bindingFactory,
-      VSBindingTreeController bindingTreeController,
+      VSBindingTreeControllerServiceProxy vsBindingTreeService,
       VSChartHandler chartHandler,
       XRepository xrepository,
       VSWizardBindingHandler wizardBindingHandler,
@@ -75,7 +75,7 @@ public class ModifyCalculateFieldService {
       DataSourceRegistry dataSourceRegistry)
    {
       this.bindingFactory = bindingFactory;
-      this.bindingTreeController = bindingTreeController;
+      this.vsBindingTreeService = vsBindingTreeService;
       this.chartHandler = chartHandler;
       this.xrepository = xrepository;
       this.wizardBindingHandler = wizardBindingHandler;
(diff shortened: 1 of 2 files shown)


fredaan:
## Code review (final round)

Scope verified: exactly 2 files changed (`ModifyCalculateFieldService.java` +17/-4, new `ModifyCalculateFieldServiceTest.java` +131). No unrelated changes.

**Root-cause claim verified against source, not just the PR description:**
- `ModifyCalculateFieldController.modifyCalculateField` derives `id` from the session-scoped `RuntimeViewsheetRef` and passes it into `modifyCalculateFieldService.modifyCalculateField(id, ...)`.
- `VSBindingTreeController.getBinding` (old code path) independently re-derives `id` from the *same* `RuntimeViewsheetRef` bean and does nothing else.
- Within one synchronous request, both reads of `RuntimeViewsheetRef` return the same value, so for the native (real WebSocket session) flow this fix is a provable no-op; for the wiz-agent's session-less flow, `RuntimeViewsheetRef` is never populated, which is the actual bug. Calling `VSBindingTreeControllerServiceProxy.getBinding(id, ...)` directly with the method's own already-correct `id` fixes it with no other side effect lost (the controller method is a two-line wrapper with nothing else in it).

**Wiring:** repo-wide search found no other construction site of `ModifyCalculateFieldService` and no XML/reflection-based wiring that could reference the old `VSBindingTreeController` constructor shape — the change is compile-safe everywhere.

**Test:** built and ran `ModifyCalculateFieldServiceTest` — 1/1 pass. To confirm it's a real tripwire (since it can't compile against the old constructor shape to literally reproduce a pre-fix failure), the reviewer temporarily changed the call site to pass a `null` id (simulating the class of bug this fixes) and reran: the test failed on the `getBinding(eq(id), ...)` verification as expected, then reverted cleanly. The Spring test-context bootstrap (`SpringExtension` + `BaseTestConfiguration`/`ConfigurationContextInitializer` + `@SreeHome`) matches this codebase's existing precedent in `AggregateInfoTest`.

**Verdict:** No Critical/Important findings. Safe to merge as-is.

PR #4972, opened 2026-09-03
Bug #76399, fix mini-toolbar mispositioned after selecting a crosstab cell in a data tip popup

larryliang-inetsoft:
## Summary

Fixes a mini-toolbar mispositioning bug: hover a chart whose data-tip view is a crosstab (opening the data-tip popup), then click/select a cell inside that crosstab -- the crosstab's mini-toolbar jumps to its static design-time canvas position instead of staying pinned to the floating popup.

## Root cause

`VSDataTipDirective` imperatively owns a mini-toolbar's position (`Renderer2.setStyle` on `.mini-toolbar`) while its assembly is the active, visible data tip. But `MiniToolbar`'s own template independently binds `[style.top.px]`/`[style.left.px]` to the assembly's static design-time `top`/`left` on that *same* element. Angular's per-binding memoization normally lets the directive's write "stick" against this competing binding (the static value doesn't change tick-to-tick), but selecting a cell in a crosstab shown as data-tip content flips the crosstab's "Show Details" toolbar action's visibility, which changes `MiniToolbarService.getToolbarLeft()`'s clamped return value -- a genuine value change that defeats the memoization and lets the static template binding clobber the directive's correct, mouse-tracked position on that exact tick.

This is the fourth bug against this same two-writer mechanism (#74751, #75512, #75820), each of which patched one more specific trigger the previous fix didn't anticipate (a scroll case, an OnPush-host case, etc.).

## Fix

Rather than adding another narrow trigger to the existing cache-guard/`ngDoCheck` mechanism, this makes `MiniToolbar`'s own `top`/`left` template bindings inert whenever a co-located `VSDataTipDirective` is the active position owner for that element -- mirroring the existing `isPopComponent`/`NaN` guard already used for the analogous "pop component" case (`get topY()`). With this, there is exactly one writer of these style properties at any given time, regardless of what causes the static value to change, closing off the whole class of trigger rather than one more instance of it.

- `vs-data-tip.directive.ts`: extracted the directive's "am I the active, visible data tip" condition into a new public `isActiveDataTipOwner()` method (previously inlined in `ngDoCheck`), reused by both `ngDoCheck` and `MiniToolbar`.
- `mini-toolbar.component.ts`: `MiniToolbar` now optionally self-injects a co-located `VSDataTipDirective` (present only when composed as `<mini-toolbar VSDataTip [miniToolbar]="true">`, as in `vs-object-container.component.html`) and adds an `isActiveDataTip` getter; `topY` and a new `leftPx` getter return `NaN` (an invalid CSS value the browser ignores, leaving the directive's own imperatively-set value in place) while that's true.
- `mini-toolbar.component.html`: `[style.left.px]` now binds to `leftPx` instead of the raw `left` Input.

Reproduction requires the tip-target crosstab's own design-time canvas position to be near/past the container's right edge (a common "park it off to the side" pattern for data-tip-only assemblies) -- not the browser viewport or where the popup visually renders -- since that's what makes `MiniToolbarService`'s clamp branch active in the first place.

## Test plan

- [x] Added `mini-toolbar.component.spec.ts` with a regression test that composes `MiniToolbar` with `VSDataTip` on the same host (mirroring `vs-object-container.component.html`) and asserts the directive's mouse-tracked position wins when active; verified this test fails (reproducing the bug: static `500px` wins) when the `leftPx` fix is reverted, and passes with it restored.
- [x] `vs-data-tip.directive.spec.ts` (existing, covers bugs #75512/#75820) still passes unmodified after the `isActiveDataTipOwner()` extraction.
- [x] Full `.tl.spec.ts` suites for `vs-object-container`, `vs-crosstab`, and `mini-menu` (165 tests) still pass -- no regressions to normal (non-data-tip) mini-toolbar positioning, crosstab rendering/interaction, or pop-component handling.
- [ ] Live visual confirmation in the portal is still needed -- no browser automation was available in this environment. Deferred to a human reviewer or a follow-up check.

### DIFF
--- a/web/projects/portal/src/app/vsobjects/objects/data-tip/vs-data-tip.directive.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/data-tip/vs-data-tip.directive.ts
@@ -83,6 +83,19 @@ export class VSDataTipDirective implements DoCheck, OnInit, OnDestroy {
       return this.dataTipService.isCurrentDataTip(this.dataTipName, this.popContainerName);
    }
 
+   /**
+    * Whether this directive instance currently owns (imperatively writes) the position of
+    * its host element, i.e. whether it is the active, visible data tip. Used by MiniToolbar
+    * (see mini-toolbar.component.ts) to suppress its own static top/left template bindings
+    * on the same element while this is true, so there is exactly one writer of those style
+    * properties at any given time -- see the two-writer note above ngDoCheck's setStyle calls.
+    */
+   isActiveDataTipOwner(): boolean {
+      return this.isCurrentDataTip() &&
+         (this.dataTipService.isDataTipVisible(this.dataTipName) ||
+          this.dataTipService.isDataTipVisible(this.popContainerName));
+   }
+
    get dataTipClass(): string {
       const name = this.dataTipService.dataTipName;
       return "current-datatip-" + (name || "none").replace(/ /g, "_");
@@ -103,9 +116,7 @@ export class VSDataTipDirective implements DoCheck, OnInit, OnDestroy {
       if(this.dataTipService.isDataTip(this.dataTipName) ||
          this.dataTipService.isDataTip(this.popContainerName))
       {
-         if(this.isCurrentDataTip() &&
-            (this.dataTipService.isDataTipVisible(this.dataTipName) ||
-             this.dataTipService.isDataTipVisible(this.popContainerName)))
+         if(this.isActiveDataTipOwner())
          {
             // cancel existing hide events
             this.debounceService.cancel(DataTipService.DEBOUNCE_KEY);
--- a/web/projects/portal/src/app/vsobjects/objects/mini-toolbar/mini-toolbar.component.html
+++ b/web/projects/portal/src/app/vsobjects/objects/mini-toolbar/mini-toolbar.component.html
@@ -18,7 +18,7 @@
 <div class="mini-toolbar"
   [class.hidden-mini-toolbar]="forceHide"
   [style.top.px]="topY"
-  [style.left.px]="left"
+  [style.left.px]="leftPx"
   [style.z-index]="zIndex">
   @if (!mobileDevice) {
     <div
--- a/web/projects/portal/src/app/vsobjects/objects/mini-toolbar/mini-toolbar.component.spec.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/mini-toolbar/mini-toolbar.component.spec.ts
@@ -1,0 +1,128 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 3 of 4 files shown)


larryliang-inetsoft:
## Independent review (adversarial pass)

Read the diff myself (`gh pr diff 4972`) and the affected code/tests in the worktree at `C:/src/agent/2/wt-bug-76399/community`, plus `01-diagnosis.md`/`02-refute.md`/`03-fix.md`/`04-verify.md`. Did not trust prior agents' reported numbers — reran everything myself.

**Tests actually run (not trusted from writeups):**
- `mini-toolbar.component.spec.ts` + `vs-data-tip.directive.spec.ts`: 7/7 passed.
- `.tl.spec.ts` suites (`vs-object-container.*`, `vs-crosstab.*`, `mini-menu.*`): 165/165 passed.
- Independently reproduced the "test catches the bug" claim: reverted `[style.left.px]="leftPx"` back to `"left"` in `mini-toolbar.component.html`, reran → regression test fails with `expected '500px' to be '101px'` (exact reported symptom). Restored the fix, reran → 2/2 pass, `git status` clean afterward.

**Scope/DI checks confirmed via grep/read, not trusted:**
- All 6 `<mini-toolbar>` usages found; only `vs-object-container.component.html` pairs it with `VSDataTip`. The other 5 (`vs-object-view`, VS Wizard object/preview, `embed-chart`, `editable-object-container`) have no `VSDataTip`, so `@Optional() @Self() dataTipDirective` is `null` there — no-op, confirmed by reading each template.
- `@Optional() @Self()` DI timing: constructor injection only resolves the *instance reference*; `isActiveDataTipOwner()` is only called later from `topY`/`leftPx` getters during change detection, well after Angular sets `@Input()`s on all co-located directives (post-constructor, pre-`ngOnChanges`). No risk of reading an uninitialized `dataTipName`/`miniToolbar`.
- `isActiveDataTipOwner()` extraction in `vs-data-tip.directive.ts` is byte-for-byte the pre-existing inlined condition — confirmed by reading the method and its one call site.
- `topY`'s guard extension to `isActiveDataTip` (beyond this bug's left-only trigger): re-derived independently by reading `ngDoCheck()` — the directive already unconditionally `setStyle`s both `top` and `left` while active, so the NaN-guard only suppresses Angular's own competing write; it cannot make a currently-correct `top` wrong. Agree this is a justified, low-risk generalization of an already-shipped pattern (`isPopComponent` on the same getter), not scope creep.

**New finding — same two-writer mechanism also exists for `z-index`, unaddressed by this fix (pre-existing, not a regression from this PR):**
`mini-toolbar.component.html` also has `[style.z-index]="zIndex"` on the same `.mini-toolbar` div that `VSDataTipDirective.ngDoCheck()` writes `z-index` to via `Renderer2.setStyle`. `zIndex`'s bound value (`getMiniToolbarZIndex(vsObject)` in `vs-object-container.component.ts:603`) depends on `needsZIndexBoost()`, which directly checks `dataTipService.isCurrentDataTip(...)` — meaning it genuinely changes value on the exact same tick a tip becomes active, the same class of trigger as the `left` bug just fixed. However, I checked the actual magnitudes: both the directive's write (`popZIndex + 99999 + …`) and the binding's boosted value (`zIndex(vsObject) + 99999 + 1`) are both ~100000+, both far above any competing stacking context (`MINI_TOOLBAR_MIN_ZINDEX` = 9920, dim canvas = 9996) — so whichever writer wins the race, the toolbar still stacks correctly. Unlike `left`/`top` (unrelated position values), this race is very unlikely to produce a *visible* bug. Not blocking, but worth a follow-up ticket/backlog note if the team wants to close this bug class completely rather than case-by-case — this fix's own precedent (`03-fix.md`'s "fourth bug against this mechanism") suggests it may recur.

**Verdict: no blocking findings.** Diff matches description exactly, tests pass and were shown (by me, independently) to actually catch the regression, the DI/self-injection pattern is sound, and the `topY` guard extension is justified. The z-index observation is informational/non-blocking (pre-existing, not introduced by this PR, not user-visible given the value magnitudes involved).

Approving and merging per this batch's policy.

jshobe-inetsoft (approved):

PR #4973, opened 2026-09-03
Fix #76408: link worksheet sandbox to originating viewsheet on open

larryliang-inetsoft:
## Summary
- Opening a worksheet as its own document (e.g. clicking the base worksheet link in the composer's bottom status bar) never linked the new worksheet's `AssetQuerySandbox` back to the originating viewsheet's `ViewsheetSandbox`, so script expressions like `worksheet['TextInput1'].value` could never resolve the viewsheet assembly (`findInChain()` correctly reported nothing found because there was nothing to find).
- Threads a `vsId` (the originating viewsheet's runtime id) from the composer's worksheet-link click (`viewsheet-pane.component.ts`) through `composer-main.component.ts` → `ws-pane.component.ts` → `OpenWorksheetEvent` → `OpenWorksheetController` → `WorksheetEventService#openWorksheet`, which now calls `AssetQuerySandbox#setViewsheetSandbox(...)` on the newly-created worksheet sandbox when `vsId` resolves to a still-open viewsheet.
- No-op (`vsbox` stays `null`, same as before) when the worksheet is opened with no originating viewsheet (portal/repository tree) or the originating viewsheet has since closed/expired.

Ruled out the alternative "read the existing `sheetRuntimeId` STOMP header via `RuntimeViewsheetRef`" approach: traced `viewsheet-client.service.ts` and found each composer pane (`ws-pane`/`viewsheet-pane`) gets its own `ViewsheetClientService` instance with its own independent `_runtimeId` state (provided per-component), so the worksheet pane's outgoing `/ws/open` message never carries the viewsheet's runtime id in that header — it's architecturally the wrong data source for this. Went with explicit client-to-server plumbing instead.

## Test plan
- Added `core/src/test/java/inetsoft/web/composer/ws/OpenWorksheetControllerSandboxLinkTest.java`, exercising the real `WorksheetEventService#openWorksheet` method body against the `ViewsheetScopeTest.vso` fixture, executing `worksheet['TableView1'].visible` through the real GraalJS engine (`AssetQuerySandbox#getScriptEnv()`/`compile`/`exec`), not just scope-level `hasMember`/`getMember` checks:
  - [x] Worksheet opened with `vsId` set (simulating "opened from within a viewsheet") resolves the viewsheet assembly and has a non-null `ViewsheetSandbox` link.
  - [x] Worksheet opened with no `vsId` (simulating portal/repository-tree open) keeps `vsbox == null` and still throws the pre-fix `TypeError` shape.
- Verified the new test fails (assertion on the linked sandbox) when the fix is temporarily reverted, and passes with the fix restored — confirming it actually catches the regression.
- Ran `AssetQueryScopeTest` and `ViewsheetScopeTest` (the related #75858/#75807 chain-walk regression tests) — all pass, no regressions.

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/ws/OpenWorksheetController.java
+++ b/core/src/main/java/inetsoft/web/composer/ws/OpenWorksheetController.java
@@ -131,7 +131,7 @@ public class OpenWorksheetController extends WorksheetController {
       entry.setProperty("openAutoSaved", event.openAutoSavedFile() + "");
       entry.setProperty("gettingStarted", event.gettingStartedWs() + "");
       String runtimeId = eventService.openWorksheet(
-         principal, entry, event.openAutoSavedFile(), event.createQuery(),
+         principal, entry, event.openAutoSavedFile(), event.createQuery(), event.vsId(),
          commandDispatcher);
 
       getRuntimeViewsheetRef().setRuntimeId(runtimeId);
--- a/core/src/main/java/inetsoft/web/composer/ws/assembly/WorksheetEventService.java
+++ b/core/src/main/java/inetsoft/web/composer/ws/assembly/WorksheetEventService.java
@@ -20,8 +20,11 @@ package inetsoft.web.composer.ws.assembly;
 
 import inetsoft.analytic.composition.ViewsheetService;
 import inetsoft.cluster.*;
+import inetsoft.report.composition.ExpiredSheetException;
+import inetsoft.report.composition.RuntimeViewsheet;
 import inetsoft.report.composition.RuntimeWorksheet;
 import inetsoft.report.composition.WorksheetEngine;
+import inetsoft.report.composition.execution.ViewsheetSandbox;
 import inetsoft.uql.asset.*;
 import inetsoft.util.Catalog;
 import inetsoft.web.composer.model.ws.WorksheetModel;
@@ -29,6 +32,8 @@ import inetsoft.web.composer.ws.TableModeService;
 import inetsoft.web.composer.ws.command.*;
 import inetsoft.web.viewsheet.command.MessageCommand;
 import inetsoft.web.viewsheet.service.CommandDispatcher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.stereotype.Component;
 
@@ -50,12 +55,21 @@ public class WorksheetEventService {
                                boolean gettingStartedCreateQuery,
                                CommandDispatcher commandDispatcher) throws Exception
    {
+      return openWorksheet(user, entry, openAutoSaved, gettingStartedCreateQuery, null,
+                           commandDispatcher);
+   }
+
+   public String openWorksheet(Principal user, AssetEntry entry, boolean openAutoSaved,
+                               boolean gettingStartedCreateQuery, String vsId,
+                               CommandDispatcher commandDispatcher) throws Exception
+   {
       String runtimeId = engine.openWorksheet(entry, user);
       WorksheetEventServiceProxy p = proxy.getIfAvailable();
 
       if(p != null) {
          return p.openWorksheet(
-            runtimeId, user, entry, openAutoSaved, gettingStartedCreateQuery, commandDispatcher);
+            runtimeId, user, entry, openAutoSaved, gettingStartedCreateQuery, vsId,
+            commandDispatcher);
       }
 
       return null;
@@ -65,11 +79,12 @@ public class WorksheetEventService {
    @ClusterProxyMethod(WorksheetEngine.CACHE_NAME)
    public String openWorksheet(@ClusterProxyKey String id, Principal user, AssetEntry entry,
                                boolean openAutoSaved, boolean gettingStartedCreateQuery,
-                               CommandDispatcher commandDispatcher) throws Exception
+                               String vsId, CommandDispatcher commandDispatcher) throws Exception
    {
       RuntimeWorksheet rws = engine.getWorksheet(id, user);
       fixWorksheetMode(rws);
       clearGettingStatedRuntimeWs(rws, gettingStartedCreateQuery);
+      linkViewsheetSandbox(rws, vsId, user);
 
       List errors = (List) AssetRepository.ASSET_ERRORS.get();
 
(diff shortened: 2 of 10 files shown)


larryliang-inetsoft:
## Independent reviewer pass (round 1)

Read the diff myself (`gh pr diff 4973`) and the affected code in a clean worktree, ran the tests myself, and probed for scenarios that could break this. Findings below.

**Diff scope**: matches the fix writeup exactly — 9 files, Angular field-threading (6 files) + Java wiring (3 files) + 1 new test. Proportionate to the bug: the plumbing is genuinely necessary since `OpenWorksheetEvent`/`WorksheetEngine.openSheet()` had no notion of an originating viewsheet anywhere in the chain (confirmed by reading `WorksheetEngine.java`, `OpenWorksheetEvent.java`).

**Tests — ran independently**: `OpenWorksheetControllerSandboxLinkTest` (2), `AssetQueryScopeTest` (1), `ViewsheetScopeTest` (14) — all pass, 17/17. I also did my own "disable and watch it fail" check (commented out the `linkViewsheetSandbox(rws, vsId, user)` call myself, independent of the fixer's/verifier's own edits): `testWorksheetOpenedFromViewsheetResolvesViewsheetAssembly` fails exactly as expected (`expected: not <null>`), `testWorksheetOpenedWithoutViewsheetHasNoSandboxLink` still passes. Restored the fix; `git status` clean afterward.

**Lifecycle/ownership concern (flagged by tester/verifier as non-blocking)** — independently traced, agree it's non-blocking:
- Confirmed `ViewsheetSandbox.dispose()` (`ViewsheetSandbox.java:686-747`) nulls its owned `wbox` but never nulls `vs`; `getViewsheet()` returns `vs` unconditionally.
- Confirmed `ViewsheetScope` (`ViewsheetScope.java`) reads `box.getViewsheet()` in every accessor with no `disposed` check anywhere.
- Traced the realistic reachable scenario further than the verifier did: `composer-main.component.ts`'s `closeSheet()` closes tabs independently — closing a viewsheet tab does **not** cascade-close a worksheet tab that was opened from it. So a user can open a worksheet from a viewsheet (establishing the link), close the viewsheet tab, and keep evaluating scripts in the worksheet tab that reference viewsheet assemblies. Post-fix this silently returns the frozen `Viewsheet` state from the moment of closure instead of throwing — trading a loud `TypeError` for a silent stale-but-plausible-looking read, in this one narrow interaction.
- I still judge this **non-blocking**: it only regresses (turns a crash into staleness) in a narrow edge case that is itself outside this bug's reported scope (the reported bug is entirely about the common case, viewsheet still open); there's no crash or cross-session/security leakage (same object, just frozen); and a proper fix (null `vs` on dispose, or a `disposed`-check in `ViewsheetScope`, or a weak/invalidating reference) would be a broader lifecycle change touching many callers — correctly out of scope for this bug per the "minimal scope" guideline. Worth keeping the verifier's drafted `claude/worksheet-query.md` note for retro.

**Other call sites — found one additional instance of the same anti-pattern, not covered by this fix**: `VSUtil.getRuntimeWorksheet(Viewsheet vs, ViewsheetSandbox box)` (`VSUtil.java:7927-7943`) takes a `ViewsheetSandbox box` parameter directly in scope but still calls the disconnected `new RuntimeWorksheet(entry, ws, user, false)` constructor without linking `box` — the identical shape as the bug's root cause. Call sites: `VSConditionDialogService.java:190,254,361` and `CoreLifecycleService.java:3134`, all gated on `assembly.isEmbedded()` (embedded-viewsheet condition-list/variable-query resolution, not the general "open worksheet as a document" flow this bug reports). I did not verify whether `ConditionUtil.fromModelToConditionList`/`executeVariablesQuery` ever evaluates a `worksheet['assembly'].value`-style script against the resulting `RuntimeWorksheet`, so I can't confirm user-visible impact — flagging as a candidate for a follow-up bug ticket rather than expanding this PR's scope (would need its own diagnosis/refute/fix cycle).

**Boundary/nullability**: `@Nullable String vsId()` on `OpenWorksheetEvent` follows the same pattern already used elsewhere in this package (`AssetEvent`, `SaveSheetEvent`, etc.), so Immutables' builder handles an absent/undefined field as null consistently. Angular's optional `vsId?`/`undefined` values are dropped by `JSON.stringify` when unset, which the `@Nullable` Java accessor handles the same as an explicit `null` — verified this is consistent end-to-end, no layer throws or coerces to `""`.

**Concurrency**: `WorksheetEventService.linkViewsheetSandbox()` resolves via `ViewsheetEngine.getViewsheet(vsId, user)` — an exact-id lookup (`vsId` is the target viewsheet's own unique `runtimeId`), so two tabs opening worksheets concurrently each resolve their own distinct viewsheet by id; no shared/keyed-by-user state in this path that could cross-wire worksheet A to viewsheet B's sandbox.

**Other worksheet-opening entry points**: confirmed via grep that `asset-tree-pane.component.ts` and `composer-empty-editor.component.ts` are the only other emitters of `onOpenSheet({type: "worksheet", ...})`, and neither sets `vsId` — both route through the same single `/ws/open` controller, so they correctly keep today's `vsbox == null` behavior. No wizard/getting-started-specific path bypasses this controller.

**Regression test boundary**: `OpenWorksheetControllerSandboxLinkTest` starts at `WorksheetEventService#openWorksheet` (Java), not the Angular event emission — the 6-file Angular field-threading chain has no test coverage. Given each hop is a single statically-typed field assignment (TypeScript compiler catches shape mismatches) and the Java-level integration test exercises the real GraalJS bridging path end-to-end, I judge this an acceptable testing boundary for this bug's risk level, not a gap that should block.

## Verdict: no blocking findings — approving and merging.

jshobe-inetsoft (dismissed):

larryliang-inetsoft:
Added commit 0dc8cf06a6, cherry-picked from #4994 (Redmine #76451): fixes an unrelated `build` check failure this PR was hitting (`PermissionHierarchyTest` racing on `SecurityChain.loadConfiguration()`'s config-file read, not caused by this PR's own diff). See #4994 for the root-cause writeup.

larryliang-inetsoft:

larryliang-inetsoft:
@fredaan could you take a look and approve when you get a chance? This is ready to merge — build passes, threads resolved — just blocked on the repo's "approval from someone other than the last pusher" rule, and everything here was pushed under one account.

fredaan (approved):

PR #4974, opened 2026-09-03
Bug #76416: Fix four shadow-settings follow-ups from #76073

larryliang-inetsoft:
## Summary

Four independent follow-up issues found while verifying #76073 "Add configurable shadow settings for rectangle and oval" (PR #4767). Each is a separate root cause in a different subsystem; committed as four separate commits for reviewability.

- **#3** — `Rectangle1.shadowInfo = {...}` script assignments threw a reflection type-mismatch because `PropertyDescriptor.convert()` had no branch for `ShapeShadow.class`. Added one, mirroring the existing `GradientColor`/`BorderColors` object-literal branches.
- **#4** — `LineVSAssemblyInfo` inherits `ShapeVSAssemblyInfo.isShadow()`/`getShadowInfo()`, so a legacy `shadowValue="true"` Line asset silently switched from the old fixed gray shadow to the new configurable shadow's hardcoded defaults. `VSShape.paintComponent()` now special-cases Line to keep using the old fixed `VSFaceUtil.addShadow(image, 6)` overload; `ShapeShadowUtil.isShapeShadow()` excludes Line so exporters don't grow bounds using the new configurable-shadow insets math. **Note**: this deviates from the initially proposed fix of excluding Line from `VSFloatable.isShadow()` — that would have disabled Line's shadow entirely rather than restoring the old one (see commit message and `docs/teams/2026-09-02-bugs-batch1/bug-76416/03-fix.md` in the enterprise repo for the full trace).
- **#1** — Print Layout (`VsToReportConverter`/`PageLayout.Shape`) predates `ShapeShadow` and never drew any shadow at all for rectangle/oval. Added a `ShapeShadow` field to `PageLayout.Shape`, populated by the converter, and drawn behind the shape's fill/outline in `Rectangle.paint()`/`Oval.paint()`. Simplification: drawn as an unblurred offset fill, since `paint(Graphics)` is format-agnostic (PDF/raster/etc.) rather than a single `BufferedImage`, so the raster-blur `VSFaceUtil.addShadow` overload can't be reused directly.
- **#2** — `ShadowPropPane`'s Distance/Blur `<input>`s could show stale out-of-range text after clamping, when the clamped result equalled the model's prior value (a no-op transition for Angular's `[ngModel]` diffing). Applied the same forced-reflow idiom already used by `alpha-dropdown.component.ts` for the identical class of bug.

## Test plan

- [x] `ShapeShadowTest` (core module, `@Tag("core")`) — 28 tests, 0 failures (covers #3 and #4)
- [x] New `PageLayoutShapeShadowTest` (core module) — 8 tests, 0 failures (covers #1)
- [x] New `shadow-prop-pane.component.spec.ts` (Angular TestBed/vitest) — 4 tests, 0 failures, asserting the actual `<input>` DOM value snaps to the clamped number, not just the model field (covers #2)
- [x] Full `core` module compiles clean (`./mvnw -pl core -am -DskipTests install`)
- [ ] Live-Composer visual confirmation of #2's fix still recommended before merge (UI-timing behavior; flagged in the diagnosis/refute docs)

### DIFF
--- a/core/src/main/java/inetsoft/report/PageLayout.java
+++ b/core/src/main/java/inetsoft/report/PageLayout.java
@@ -20,6 +20,7 @@ package inetsoft.report;
 import inetsoft.graph.internal.GTool;
 import inetsoft.report.internal.*;
 import inetsoft.sree.SreeEnv;
+import inetsoft.uql.viewsheet.ShapeShadow;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -123,6 +124,20 @@ public class PageLayout implements Serializable, Cloneable {
       }
 
       /**
+       * Set the drop shadow to draw behind this shape, or null for none.
+       */
+      public void setShadow(ShapeShadow shadow) {
+         this.shadow = shadow;
+      }
+
+      /**
+       * Get the drop shadow to draw behind this shape, or null for none.
+       */
+      public ShapeShadow getShadow() {
+         return shadow;
+      }
+
+      /**
        * Scale the shape.
        * @param xs x scale 0 to 1.
        * @param ys y scale 0 to 1.
@@ -199,6 +214,7 @@ public class PageLayout implements Serializable, Cloneable {
          style = shape.style;
          xs = shape.xs;
          ys = shape.ys;
+         shadow = shape.shadow;
       }
 
 		/**
@@ -221,6 +237,7 @@ public class PageLayout implements Serializable, Cloneable {
 		private int zindex; // just for previewing vs printlayout.
       private Color color = Color.black; // shape color
       private int style = StyleConstants.THIN_LINE; // line style
+      private ShapeShadow shadow; // drop shadow, null if none
       double xs = 1, ys = 1;
    }
 
@@ -519,6 +536,27 @@ public class PageLayout implements Serializable, Cloneable {
       @Override
       public void paint(Graphics g) {
          Color oc = g.getColor();
+
+         // draw behind the shape's own fill/outline, matching the shadow
+         // direction convention ShapeShadow already establishes on the
+         // viewsheet side. paint(Graphics) is fed whatever the active
+         // export/print pipeline provides (PDF content-stream graphics,
+         // raster graphics, etc.) rather than a single BufferedImage, so this
+         // draws the shadow as its own filled shape offset by direction and
+         // distance instead of reusing VSFaceUtil.addShadow's raster blur --
+         // the blur radius is not applied here, a simplification accepted
+         // because a genuine Gaussian blur is not practical in a
+         // Graphics-only context.
+         ShapeShadow shadow = getShadow();
+
+         if(shadow != null) {
+            Graphics2D sg = (Graphics2D) g.create();
+            sg.translate((int) (shadow.getOffsetX() * xs), (int) (shadow.getOffsetY() * ys));
+            sg.setColor(shadow.getShadowColor());
+            sg.fillRect((int) (getX() * xs), (int) (getY() * ys),
+               (int) (getWidth() * xs), (int) (getHeight() * ys));
+            sg.dispose();
+         }
 
          if(fill != null) {
             Graphics2D g2 = (Graphics2D) g.create();
@@ -999,6 +1037,19 @@ public class PageLayout implements Serializable, Cloneable {
       @Override
       public void paint(Graphics g) {
          Color oc = g.getColor();
+
+         // see Rectangle.paint() for why this is an unblurred offset fill
+         // rather than VSFaceUtil.addShadow's raster blur.
+         ShapeShadow shadow = getShadow();
+
+         if(shadow != null) {
+            Graphics2D sg = (Graphics2D) g.create();
+            sg.translate((int) (shadow.getOffsetX() * xs), (int) (shadow.getOffsetY() * ys));
+            sg.setColor(shadow.getShadowColor());
+            sg.fillOval((int) (getX() * xs), (int) (getY() * ys),
+               (int) (getWidth() * xs), (int) (getHeight() * ys));
+            sg.dispose();
+         }
 
          if(fill != null) {
             Graphics2D g2 = (Graphics2D) g.create();
--- a/core/src/main/java/inetsoft/report/gui/viewsheet/VSShape.java
+++ b/core/src/main/java/inetsoft/report/gui/viewsheet/VSShape.java
@@ -52,14 +52,26 @@ public abstract class VSShape extends VSFloatable {
       g2.dispose();
 
       if(isShadow()) {
-         ShapeShadow shadow = getInfo().getShadowInfo();
-         // must match the insets getImage() grew the canvas by
-         Insets insets = ShapeShadowUtil.getScaledShadowInsets(getInfo());
-         image = VSFaceUtil.addShadow(image, shadow, insets);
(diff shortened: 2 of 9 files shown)


larryliang-inetsoft:
## Independent review — Bug #76416 (4 bundled sub-fixes)

Read the full diff myself (`gh pr diff 4974`) and the affected code in the worktree, ran every claimed test suite myself (not taken on the tester's word), and probed each sub-fix adversarially per the batch lead's brief.

### #1 — Print Layout shadow (`PageLayout.java`, `VsToReportConverter.java`)
- Confirmed the new `PageLayout.Shape.shadow` field/`Rectangle`/`Oval.paint()` shadow drawing is genuinely format-agnostic-safe: `PDFPrinter.setColor(Color)` (`PDFPrinter.java:1213-1227`) explicitly reads `c.getAlpha()` and routes it through its `/ca`/`/CA` transparency-group machinery, so the alpha-baked `ShapeShadow.getShadowColor()` composites correctly on the PDF content-stream path, not just raster. Standard AWT `Graphics2D.fillRect/fillOval` handles alpha `Color` natively for any raster consumer. No exotic-`Graphics` throw risk found.
- Checked the `xs`/`ys` scale factors the new code multiplies the shadow offset by: `PageLayout.Shape.setScale(double,double)` has **zero callers anywhere in the codebase** — `xs`/`ys` are always their `1,1` default. So the multiplication is inert today; a latent no-op, not a live bug, and consistent with how the pre-existing fill-drawing code already scales by the same (always-1) factors.
- Confirmed only `Rectangle`/`Oval` construct via `VsToReportConverter` (no second converter), and `PageLayout.Shape.paint()`'s only consumers are `ShapePaintable`/`StylePage`/PDF/print pipelines — no `InfoShape`/other shape type needs this per the original bug's scope.
- Reconfirmed the pre-existing, unrelated `getX() + xs` vs `getX() * xs` inconsistency the tester flagged: verified via the diff context that this line is unchanged (no `+`/`-` prefix in the hunk) — genuinely pre-existing, not touched here. Agree it's a non-blocking follow-up.
- Ran `PageLayoutShapeShadowTest` myself: **8/8 pass.**

### #2 — Shadow dialog staleness (`shadow-prop-pane.component.ts`)
- Swept Composer for the same `[ngModel]`+`(ngModelChange)`-does-its-own-clamp hazard: grepped all 15 composer dialog `.html` files using `ngModelChange`, then checked every `.ts` for `Math.max/min`/`parseInt`-based clamping wired to the change handler. Only `shadow-prop-pane` (this fix) and the already-fixed `alpha-dropdown` have it. `aggregate-pane`/`crosstab-pane`'s "Top N" fields looked similar at a glance (`parseInt` present) but their `(ngModelChange)` handlers do a bare `row.n = $event` with no immediate clamp/transform — validation happens later on apply, not the same defect class. No other live instance of this bug found; the fix is correctly scoped to just this dialog.
- Verified the `clamped !== raw` guard handles the `NaN` (cleared-field) case correctly: `NaN !== 0` is always `true` in JS, so the reflow always fires for the empty-input case, matching the "cleared field forces reflow" test.
- Ran the actual spec myself: **4/4 pass**, including genuine `<input>.value` DOM assertions.
- Live-Composer visual confirmation remains outstanding as flagged by every prior phase — expected, not a new gap.

### #3 — `shadowInfo` object-literal conversion (`PropertyDescriptor.java`)
- Confirmed branch order is irrelevant: every branch in the `if/else-if` chain (`Position`, `Size`, `GradientColor`, etc., now `ShapeShadow`) dispatches on `type == X.class`, where `type` is the setter's *declared parameter type* from reflection — not a structural/duck-typed match against the JS value — so `ShapeShadow.class` cannot be shadowed by an earlier branch regardless of position in the chain.
- Confirmed the silent-skip-on-wrong-type behavior is consistent with the *existing* convention in this method (no branch anywhere in `convert()` has a final default/throw) — this fix doesn't introduce a new inconsistency, it matches `GradientColor`/`BorderColors`'s existing behavior. I'd still log-and-move-on rather than silently defaulting if this were greenfield, but that's a pre-existing design choice, not something to block this fix over.
- Traced `setColor`/`setDirection`: both call `.toString()` unconditionally on the raw literal field, which never throws for any Java `Object`; numeric fields are `instanceof Number`-guarded. No path in the new branch can throw. Confirmed by reading, not just accepting the fix doc's characterization.
- Ran `ShapeShadowTest` myself: **27/27 pass** (fix doc's "28" is a doc-only miscount, already flagged by the tester — 0 failures either way).

### #4 — Line shadow self-correction (`VSShape.java`, `ShapeShadowUtil.java`)
- Independently pulled the pre-#76073 baseline: `git show 73af0e5aeb9e^:core/src/main/java/inetsoft/report/gui/viewsheet/VSShape.java` — confirms `if(isShadow()) { image = VSFaceUtil.addShadow(image, 6); } ... g.drawImage(image, 0, 0, null);` exactly, matching the new `instanceof LineVSAssemblyInfo` branch line-for-line. The self-correction (keep `isShadow()` gating unchanged, branch only the *rendering* choice) is right — excluding Line from `isShadow()` itself would have collapsed to the `else` (no-shadow) branch and removed a capability Line always had, which is what the original brief would have caused.
- Confirmed `VSFloatable.java` does not appear in the PR's file list — `isShadow()` genuinely untouched.
- Swept for other `VSShape` subclasses: exactly `VSLine`/`VSOval`/`VSRectangle` exist (grep), nothing else reaches this branching. Also checked `VSFaceUtil.addShadow(BufferedImage, int)`'s only other caller, `VSImageable.java:75` — a separate, unrelated image-shadow path not gated by `ShapeVSAssemblyInfo` at all, unaffected by this change.
- Swept for every `instanceof ShapeVSAssemblyInfo` site in the codebase (5 total): `VSFloatable` (correctly untouched), `ShapeShadowUtil` (fixed), two in `ShapeVSAScriptable` (gate script-member visibility generally, not shadow-specific — Line is separately excluded from `shadow`/`shadowInfo` registration via `supportsFill()==false`, unrelated to this fix), and one in `VSObjectPropertyService` (unrelated composer-selection filtering, nothing to do with shadow). No missed fourth shadow-gating site.
- Ran the new `lineIsExcludedFromTheConfigurableShadowGateEvenWithLegacyShadowValueTrue` test as part of the full `ShapeShadowTest` run: passes.

## Tests run (by me)
- `./mvnw -pl core test -Dtest=ShapeShadowTest,PageLayoutShapeShadowTest` → **PageLayoutShapeShadowTest 8/8, ShapeShadowTest 27/27, BUILD SUCCESS.**
- `npx ng test portal --include='**/shadow-prop-pane.component.spec.ts' --watch=false` → **4/4 pass.**

## Verdict: no blocking findings on any of the four sub-issues.

Non-blocking notes (already surfaced by prior phases, restating for the record):
- Fix doc table overstates `ShapeShadowTest` count as 28; actual is 27 (doc-only).
- Pre-existing, unrelated `getX() + xs` vs `getX() * xs` scale inconsistency in `PageLayout.java` near the new #1 code — worth a follow-up ticket, not blocking.
- #2's live-Composer visual confirmation is still recommended before considering the UI fix fully closed, per every prior phase's own flag.

Approving and merging.

larryliang-inetsoft:
@fredaan could you take a look and approve when you get a chance? Build/review checks are green, just needs a first approval to satisfy branch protection (same last-pusher-approval rule blocking the rest of this batch).

fredaan (approved):

PR #4975, opened 2026-09-03
Add unit tests for SSO settings validation, cluster broadcast, and schedule notification (test-only)

sarapeng86:
## Summary
  - Add unit tests for `SSOSettingsService.updateSSOSettings()`: blank-attribute switches to
    OpenID/Custom succeed without validation, switching to SAML fails real java-saml validation
    when required fields are missing, and a save failure downgrades the local filter to NONE
    while still persisting the new (not the previous) type.
  - Add unit tests for `CustomSSOConfig`: class-name vs. inline-Groovy mutual-exclusion overwrite
    order matches the real caller order.
  - Add unit tests for `SSOFilterPublisher`: a cluster broadcast failure is swallowed and must not
    block the local Spring event from publishing (this exact link broke production once, commit
    3bf25021e); remote/local `messageReceived()` routing is also covered.
  - Add unit tests for `ScheduleUsersChangeService`, a previously undocumented consumer of
    `SSOTypeChangedMessage` found via a repo-wide grep: SSO turned fully on/off notifies
    schedule-view subscribers, a same-state protocol switch or no-op change does not. Also
    documents a currently-harmless argument-order bug at the only call site (filed as Bug #76438).

  Part of the SSO module test architecture plan. No production code changes.

  ## Test plan
  - [x] `./mvnw test -pl core
  -Dtest=SSOSettingsServiceTest,CustomSSOConfigTest,SSOFilterPublisherTest,ScheduleUsersChangeServiceTest`

### DIFF
--- a/core/src/test/java/inetsoft/sree/security/support/SecurityTestDataBuilder.java
+++ b/core/src/test/java/inetsoft/sree/security/support/SecurityTestDataBuilder.java
@@ -22,6 +22,7 @@ import inetsoft.sree.PropertiesEngine;
 import inetsoft.sree.SreeEnv;
 import inetsoft.sree.security.*;
 import inetsoft.uql.util.Identity;
+import inetsoft.util.DataSpace;
 import org.mindrot.BCrypt;
 import org.springframework.test.util.ReflectionTestUtils;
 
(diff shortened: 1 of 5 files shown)


jshobe-inetsoft (dismissed):

fredaan (approved):

PR #4976, opened 2026-09-03
fix(wiz): L5 tables/crosstabs/table-formatting parity fixes

NeverStopFMH:
## Summary
Parity-audit fixes for lane L5 (tables, crosstabs, table formatting) against the agent binding tools in `inetsoft.web.wiz.binding`:

- **Finding 1** — `TableBindingMutator.moveField()` auto-converts a field's kind to match its destination shelf (dimension ↔ measure), matching the classic UI's silent auto-convert on a crosstab pivot.
- **Finding 2** — `set_table_source(force:true)` now keeps a bound field whose column still resolves in the new source (qualified-name matching both sides), instead of discarding every field on repoint.
- **Finding 6** — `set_field_ranking`'s `others` flag now writes `groupOthers` (the field the real ranking condition and the UI's own "Group all others together" checkbox both read), not the orphaned `namedOthers` field.
- **Finding 8** — `summarySideBySide` is forwarded from the wire model to the real assembly; previously a complete, undisclosed dead write.
- **Finding 9** — `suppressGroupTotal` uses correct composite-key matching (`<column>:<shelf><index>`) and refuses an unbound measure, instead of silently no-op'ing or suppressing unrelated dimensions.
- **Finding 10** — `rowTotals`/`colTotals`/`percentageBy` accept a `"$(variable)"` reference or an `"=expression"`, matching the real `DynamicValue` fields on the assembly.

Finding 11 (`set_column_labels`) is a deliberate, disclosed refusal on both Table and Crosstab, unchanged by this PR — see the parity report in the companion `stylebi-wiz` repo PR for the full account of why a working fix was attempted, live-verified, and then reverted.

## Test plan
- [x] `inetsoft.web.wiz.binding.*` package unit tests, 0 failures
- [x] Each finding re-proved live via the `mcp__composer-chat__*` plugin tools against a rebuilt dev server (see the parity report for per-finding evidence)


NeverStopFMH:
Addressed in 31032733d:
1. Trimmed the process-context javadoc references (commit hash, "per the operator's decision") — kept the behavioral rationale, dropped the narrative.
2. `setSort`/`setRanking` now validate the shelf/column before the sortByField/measure reference, so a call failing both reports the shelf error first.

Left #2 (the `dataTypeOf`/`columnsOf` duplication) as-is — the two loops key off different things (column name across all tables vs. table name returning its columns), and a forced shared helper made both call sites read worse in a draft I tried, not better. Full `inetsoft.web.wiz.binding.*` suite re-run, 576 tests, 0 failures.

NeverStopFMH:
Fixed in 9023d883e — \`dataTypeOf\` now matches a reported column name against the target either exactly or with either side's qualifier stripped, reusing \`TableBindingService.unqualified()\` (widened from private to package-private) rather than duplicating the stripping logic. Also directly confirms the note on the "duplication" nit: the two helpers now legitimately share the qualifier-stripping primitive, though the surrounding loops still key off different things and stay separate.

Added \`movingANumericDimensionWithAQualifiedSourceNameOntoAggregatesDefaultsToSum\`, exercising a qualified source column name against a bare field name. Full \`inetsoft.web.wiz.binding.*\` suite re-run: 577 tests, 0 failures.

NeverStopFMH:
Fixed in 61d493599 — verified directly against \`AggregateFormula.java\` first (both \`First\`/\`Last\` override \`isTwoColumns()\` to \`true\`, same as \`Correlation\`/\`Covariance\`), then added both to \`MULTI_ARG_FORMULA_NAMES\`. New test \`acceptsAFirstAggregatesFullNameByPrefix\` exercises \`First\`; \`Last\` shares the identical code path. Full \`inetsoft.web.wiz.binding.*\` suite re-run: 578 tests, 0 failures.

fredaan (approved):

PR #4977, opened 2026-09-03
Carry column description, label and declared kind through the tabular catalog SPI

Nelson-cao:
`TabularColumn` gains `description`, `label` and `isDimension`, and `TabularCatalogService.toDataset()` projects them onto `OsiField`. Additive, with a compatibility constructor, following exactly how `params` and `columnsMayBeIncomplete` were added to `TabularDatasetSchema`.

Motivated by GA4 (stylebi-enterprise#TBD), which is the first source to *declare* which columns are dimensions and which are metrics rather than leaving it to be inferred from type, and the first whose column metadata carries real prose descriptions.

`isDimension` is projected into the field-level COMMON extension as `declaredIsDimension`, not onto `OsiField.isDimension` alone. That distinction is load-bearing: `OsiField.isDimension` is overwritten by the first annotation, so a source's own classification would not survive it. The COMMON extension is not written by the annotator, so the declaration outlives annotation and can be cross-checked against it on the wiz side.

**The regression surface is seven implementers plus two test doubles**, not one connector. `ODataRuntime`, `SharepointOnlineRuntime`, `CassandraRuntime`, `HiveRuntime`, `AerospikeRuntime`, `OrientDBRuntime` and `SForceRuntime` all implement the SPI; `FakeCatalogRuntime` and `FakeFailingCatalogRuntime` in this module's own test tree do too. Those two are the useful canary — a compatibility constructor done wrong breaks core's own tests before any connector's, at compile time. None of the nine is modified by this PR.

42 tests, of which 37 predate this change and pass unaltered:

```
./mvnw -pl core -am -Dtest=TabularCatalogFieldDeclarationTest,TabularCatalogServiceTest,TabularCatalogServiceContractValidationTest,TabularCatalogDatasetExtensionTest -Dsurefire.failIfNoSpecifiedTests=false test
```

`-Dsurefire.failIfNoSpecifiedTests=false` is needed because `-am` pulls upstream modules into the reactor and `-Dtest` applies to all of them; without it a module with no matching test fails the build with a message that reads like a mistyped class name.

### DIFF
--- a/core/src/main/java/inetsoft/uql/tabular/TabularColumn.java
+++ b/core/src/main/java/inetsoft/uql/tabular/TabularColumn.java
@@ -26,5 +26,33 @@ package inetsoft.uql.tabular;
  *             not exhaustive: the actual, closed vocabulary is every {@code public static final
  *             String} constant {@code XSchema} itself declares. The connector maps its own native
  *             type onto this vocabulary; the native type name does not cross this boundary.
+ * @param description the column's own description as the source declares it, verbatim — not a
+ *                     sentence the connector composes. {@code null} when the source declares
+ *                     nothing, never an empty string standing in for "nothing". Where "nothing" is
+ *                     represented is source-specific (a protobuf getter never returns null, so a
+ *                     connector reading one must translate its own empty string to {@code null}
+ *                     here); that translation belongs to the connector, which is the only party
+ *                     that knows its source's convention.
+ * @param label the column's own display name as the source declares it. {@code null} when blank.
+ *              May equal {@code name} — that is still a true statement about the source.
+ * @param isDimension three-valued. {@code TRUE}/{@code FALSE} only when the source itself sorts
+ *                    this column into one of two disjoint dimension/measure lists; {@code null}
+ *                    when the source does not say. Deliberately not an inference: a connector must
+ *                    not guess from the column's type (numeric therefore a measure, or similar).
+ *                    The three-valued form is required, not merely cautious — {@code FALSE} ("the
+ *                    source says this is a measure") and {@code null} ("the source did not say")
+ *                    are different facts downstream.
  */
-public record TabularColumn(String name, String type) {}
+public record TabularColumn(String name, String type, String description, String label,
+                             Boolean isDimension)
+{
+   /**
+    * Compatibility constructor for callers written before description/label/isDimension existed —
+    * every existing connector's construction site. All three default to null, which
+    * {@code TabularCatalogService} treats identically to "the connector said nothing": the key is
+    * omitted from the projection entirely, not written empty or false.
+    */
+   public TabularColumn(String name, String type) {
+      this(name, type, null, null, null);
+   }
+}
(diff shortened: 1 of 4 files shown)


Nelson-cao:
Two answers, and one observation I am declining with a reason rather than silently.

**On taking the schema claim on faith — you do not have to.** You are right that `ossie-schema.json` is not vendored in this repo, and it should not be; it belongs to the spec project. Here is what I actually ran, so the record does not rest on a commit message:

```
gh api repos/open-semantic-interchange/OSI/contents/core-spec/ossie-schema.json --jq .content | base64 -d
```

`$defs/Field` sets `additionalProperties: false` and names exactly eight properties:

```
name, expression, dimension, label, description, datatype, ai_context, custom_extensions
```

`isDimension` is not among them, which is why the first commit's top-level key made every emitted `Field` non-conformant. `custom_extensions` is `{vendor_name, data}` with `data` a JSON string — the spec's own extension mechanism, and where `declaredIsDimension` now lives. After the revert, `OsiField`'s keys are `name / expression / dimension / label / description / ai_context / custom_extensions`, all inside the eight.

Worth adding, since it changes what the file name search should be: the upstream file is **`ossie-schema.json`**, not `osi-schema.json`. The project was renamed to Apache Ossie, and the one comment in the wiz repository that cites the spec still uses the old name — and separately asserts `additionalProperties: true` for `AIContext`, which the schema does not say (it declares no `additionalProperties` there at all). Both are recorded for a follow-up round rather than patched here.

**Declining the `extData.put("type", type)` observation**, which you raised in both passes, so it deserves more than being ignored. You are right that it is inconsistent with `declaredIsDimension`'s omit-when-null convention sitting two lines below it. But you also established why it cannot be exercised: `validateColumnNames` rejects a null or non-`XSchema` `column.type()` before `toDataset()` is ever reached, and `describeTable_nullColumnType_throwsNamedExceptionNotNpe` pins that. A guard on an unreachable branch is dead code that no test can distinguish from its absence, and this round has already spent two deliberate exceptions to its own "nits are recorded, not fixed" rule. Recorded, not fixed.

**On the 42/42 you could not run:** it was executed twice, once by the implementer and once by me after deleting `core/target/surefire-reports`, and again after the conformance revert — still 42, which is the expected result of removing one key and relocating three assertions. Of those 42, **37 predate this round and pass unaltered**; that is the load-bearing evidence that changing a record seven implementers share did not disturb them, rather than the zero diff in their files, which as the enterprise-side doc now says is necessary but not sufficient.

```
./mvnw -pl core -am -Dtest=TabularCatalogFieldDeclarationTest,TabularCatalogServiceTest,TabularCatalogServiceContractValidationTest,TabularCatalogDatasetExtensionTest -Dsurefire.failIfNoSpecifiedTests=false test
```

`-Dsurefire.failIfNoSpecifiedTests=false` is required because `-am` pulls upstream modules into the reactor and `-Dtest` applies to all of them; without it a module with no matching test fails the build with a message that reads like a mistyped class name.

fredaan (approved):

PR #4979, opened 2026-09-03
Fix #76443: spurious unsaved-changes prompt on rapid security-action switching

larryliang-inetsoft:
## Summary
- Fixes bug #76443: switching between security action tree nodes quickly could trigger a spurious "Action Permission Settings Changed" confirmation dialog even though the user never edited a permission.
- Root cause: `SecurityActionsPermissionsComponent`'s `action` setter cleared `originalModel` synchronously when a new node was selected, but left the *previous* node's `tableModel` in place while the new node's permissions loaded asynchronously via `getPermissions()`. During that gap, `isModelChanged()` (`!isEqual(tableModel, originalModel)`) deterministically returned `true`. If the user clicked another node before the HTTP response landed, `SecurityActionsPageComponent` read `unsavedChanges === true` and showed the confirm dialog.
- Fix: clear `tableModel` alongside `originalModel` in the `if(value)` branch of the setter, mirroring the existing `else` (action = null) branch, so `isModelChanged()` evaluates `isEqual(null, null) === true` → `false` throughout the load.

## Test plan
- [x] Added a regression test in `security-actions-permissions.component.tl.spec.ts` using a manually-controlled `Subject` (not a synchronously-resolving `of(model)`, which would mask the bug) to keep `getPermissions()` pending while asserting `isModelChanged()` stays `false` across a rapid node switch.
- [x] Verified the new test fails against the pre-fix code (reverted the fix locally, confirmed `expected true to be false` failure) and passes with the fix restored.
- [x] Ran the full `security-actions-permissions.component.tl.spec.ts` suite (6 tests) — all passing.

### DIFF
--- a/web/projects/em/src/app/settings/security/security-actions/security-actions-permissions/security-actions-permissions.component.tl.spec.ts
+++ b/web/projects/em/src/app/settings/security/security-actions/security-actions-permissions/security-actions-permissions.component.tl.spec.ts
@@ -42,7 +42,7 @@ import { HttpErrorResponse } from "@angular/common/http";
 import { NO_ERRORS_SCHEMA } from "@angular/core";
 import { MatSnackBar } from "@angular/material/snack-bar";
 import { render } from "@testing-library/angular";
-import { EMPTY, Observable, of } from "rxjs";
+import { EMPTY, Observable, of, Subject } from "rxjs";
 
 import { Tool } from "../../../../../../../shared/util/tool";
 import { ResourcePermissionModel } from "../../resource-permission/resource-permission-model";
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
Independent review (round 1).

Re-read the diff directly and the live worktree file — the change is exactly one added line, `this.tableModel = null;`, as the first statement of the `if(value)` branch in `SecurityActionsPermissionsComponent`'s `action` setter. The `else` branch is untouched (confirmed by direct read, not just the diff).

Ran the regression test myself, independently, three times in `wt-bug-76443`:
- As shipped: 6/6 pass, including the new "should not report unsaved changes while permissions are loading for a newly selected action" test.
- Reverted the fix myself (my own Edit, not reuse of prior patches): reran → `AssertionError: expected true to be false` on exactly that new test, 5/6 pass.
- Restored the fix: reran → 6/6 pass again. `git status` clean afterward.

Adversarial checks:
- Swept the EM settings tree (9 files matching `isModelChanged`/`originalModel` patterns) for the same "Input setter clears one dirty-check field, async reload sets both together" shape elsewhere. Found none: `edit-identity-view.component.ts` receives its model synchronously (no async gap), and the schedule task/cycle editor pages load their model once via `ngOnInit`/route params rather than via a live per-selection `@Input` setter. No same-bug-other-site findings.
- Traced the same-node-reclick scenario: `SecurityActionsTreeComponent` emits a stable flattened-node reference (built once in `ngOnInit`), so re-clicking the currently-selected node re-assigns the same object reference to `selectedAction`. Angular's input-binding diffing does not re-invoke the `action` setter when the bound value is referentially unchanged, so re-clicking the same node never re-triggers the clear/reload at all — no new flicker, behavior unchanged from before this fix.
- Confirmed the regression test's `Subject` is genuinely pending at the mid-test assertion (not masked by a synchronously-resolving `of(model)`).

No blockers or other findings. Approving.

jshobe-inetsoft (approved):

PR #4980, opened 2026-09-03
Bug #76432, remove nineteen unread property declarations

larryliang-inetsoft:
## Summary

Removes 19 property declarations from `defaults.properties` that have no reader anywhere in the codebase, following the precedent set by #76052 (`efe6243f`) and #76298 (`aff1afde`):

- `application.server`, `application.url`, `asset.tree.size`, `cluster.scheduler.nodes`, `jdbc.connection.user.size`, `repository.tree.width`, `script.execution.stackdepth`, `system.admin.address`, `work.directory` — no reader at all.
- `btree.descriptor.max`, `btree.key.size.max` — read by `BTreeFile`, but `MergedTable`'s constructor (the only production caller of `new BTreeFile(...)`) never sets these two keys on its `Properties` object, so `BTreeFile` falls back to its own hardcoded field defaults (`DEFAULT_DESCRIPTORS_MAX=4`, `DEFAULT_KEYSIZE_MAX=256`), which happen to match the values being removed here — a true no-op for these two.
- `btree.dirty.size.max`, `btree.page.size` — also read by `BTreeFile`, but `MergedTable`'s constructor explicitly hardcodes these two to `"131072"` and `"256"` respectively (different from both `defaults.properties`' values, 4096/4096, and from `BTreeFile`'s own field defaults, 4096/4096) — so these two were never live via `SreeEnv` for a different reason: `MergedTable` always overrides them with its own literals regardless. Still dead-in-practice from `defaults.properties`'s perspective (setting it via Settings > All Properties has no effect either way), just via a different mechanism than the other two.
- `BTreeFile.java`/`MergedTable.java` are untouched in this PR either way; only the dead-in-practice `defaults.properties` declarations are removed.
- `datasource.registry.file` — no reader.
- `empty.crosstab.convert`, `sree.find.backward`, `text.break.pages` — no Java reader; also removed from the `booleanProperties` array in `properties-tool.ts` so Settings > All Properties no longer renders a checkbox for a key with no effect (same shape as `query.aggregate.merge` in #76052 and `html.close.button`/`html.image.scale` in #76298).
- `init.home` — read via `SreeEnv.getProperty("init.home")` in `Scheduler.java`, but the return value is discarded (no assignment, no branch). The read-and-discarded statement itself is left in place; only the now-pointless declaration is removed.
- `sree.properties` — the property *key* has no reader (only the literal filename string `"sree.properties"` appears elsewhere, which is unrelated and untouched).

**Flagged, not fixed**: `uql.properties` has a second, independently dead `datasource.registry.file` declaration (different default value, also never loaded by any Java code). Left out of scope per the same precedent pattern; worth a future cleanup.

## Test plan

- [x] Verified all 19 keys had no live reader via repo-wide grep (`.java`/`.ts`/`.properties`/`.xml`/`.html`), before and after the change.
- [x] Ran `PropertyDefaultsTest` (the existing test that pins values in `defaults.properties` against call-site defaults) — 6/6 pass, none of the removed keys are referenced by it.
- [x] Confirmed `BTreeFile.java`/`MergedTable.java` untouched, and independently re-derived exactly how each of the 4 btree keys is dead-in-practice (2 via `BTreeFile`'s own field defaults matching, 2 via `MergedTable`'s own hardcoded overrides), correcting an earlier inaccurate blanket claim that all 4 fell back to matching `BTreeFile` defaults.


larryliang-inetsoft:
Independent review (4th pass, following diagnosis/refute/verify).

**Checked myself, from scratch:**
- Diff scope: exactly the 19 `defaults.properties` lines and 3 `booleanProperties` entries in `properties-tool.ts` are touched — nothing else.
- Empty-valued keys (`application.server`, `cluster.scheduler.nodes`, `repository.tree.width`, `system.admin.address`): re-grepped `community/*` for all four — zero hits anywhere outside `defaults.properties`. Since there is no reader at all for any of these, there's no code path that could distinguish `getProperty(key) == null` from `getProperty(key, "").isEmpty()` — the missing-vs-empty-string concern doesn't apply because nothing calls `getProperty` on these keys in the first place.
- `properties-tool.ts`: read the full file. `booleanProperties` is a plain array with exactly one consumer (`getPropertyOptions`, a pure `indexOf` membership check, one call site). No positional/index dependency — shrinking by 3 entries is safe.
- Precedent commits `efe6243f` (#76052) and `aff1afde` (#76298): pulled both directly from the community repo and confirmed this PR's diff shape (defaults.properties removal + properties-tool.ts booleanProperties removal, same combination) matches both.
- PR description: confirms the `uql.properties` duplicate `datasource.registry.file` declaration is flagged as known-but-out-of-scope, consistent with the fix doc.
- Ran `PropertyDefaultsTest` myself (not trusting the reported number): `mvn -pl core -Dtest=PropertyDefaultsTest test` → `Tests run: 6, Failures: 0, Errors: 0, Skipped: 0`. None of the pinned assertions reference any of the 19 removed keys.

No blocker or important findings. Approving.

jshobe-inetsoft (dismissed):

larryliang-inetsoft:
Corrected the PR description per the review finding: the btree reasoning previously stated all 4 keys fell back to matching `BTreeFile` field defaults. Verified directly against `MergedTable.java`'s constructor — that's only true for `btree.descriptor.max`/`btree.key.size.max`; `btree.dirty.size.max`/`btree.page.size` are actually overridden by `MergedTable`'s own hardcoded literals (`131072`/`256`), a different mechanism but the same dead-in-practice conclusion. No code change needed — the removal itself was already correct. Thread resolved.

larryliang-inetsoft:
Added commit d78247d849, cherry-picked from #4994 (Redmine #76451): fixes an unrelated `build` check failure this PR was hitting (`PermissionHierarchyTest` racing on `SecurityChain.loadConfiguration()`'s config-file read, not caused by this PR's own diff). See #4994 for the root-cause writeup.

larryliang-inetsoft:

larryliang-inetsoft:
@fredaan could you take a look and approve when you get a chance? The CHANGES_REQUESTED review was about a flaky/unrelated build failure that's since been fixed (see #4994's follow-up, now superseded by #4984) — build passes and the thread is resolved.

fredaan (approved):

PR #4983, opened 2026-09-03
Bug #76418, remove duplicate ResizeHandlerService provider on ComposerMainComponent

larryliang-inetsoft:
## Summary
- Worksheet Pane table was truncated and never adapted its virtualized column range after a browser-window maximize/restore.
- Root cause: `ComposerMainComponent`'s `@Component` decorator declared its own component-scoped `ResizeHandlerService` provider, which shadowed the composer route's provider (`composer.routes.ts`) for the entire `ComposerMainComponent` subtree (`WSPaneComponent` -> `WSDetailsPaneComponent` -> `WSDetailsTableDataComponent`, plus `ViewsheetPaneComponent`, `LayoutPaneComponent`). Only the route-scoped instance ever has `initListeners()` called on it (in `ComposerAppComponent.ngOnInit()`), so the shadowed instance these descendants actually injected never received real `window: resize` events. `WSDetailsTableDataComponent.columnIndexRange` (which gates rendered columns) was therefore never recomputed after maximize/restore.
- Fix: remove the duplicate `ResizeHandlerService` entry from `ComposerMainComponent`'s `providers` array so the component and its whole subtree resolve the single, already-initialized route-level instance. `ComposerMainComponent`'s own `onVerticalResizeEnd()` calls are unaffected (same service, single shared instance).
- Lifecycle check: `initListeners()`/`removeListeners()` are called from `ComposerAppComponent.ngOnInit()`/`ngOnDestroy()`, and `ComposerMainComponent` is rendered as a plain child in `ComposerAppComponent`'s own template (no `router-outlet` in the composer subtree), so it cannot be destroyed/recreated independently of `ComposerAppComponent` — no lifecycle mismatch from sharing the route-level instance.

## Test plan
- Added `composer-main.component.resize-handler-di.tl.spec.ts` with two regression tests:
  - asserts `ComposerMainComponent` resolves the same `ResizeHandlerService` instance as an ancestor injector (fails if the duplicate provider is reintroduced)
  - asserts `onSplitDragEnd()` routes its resize notification through that same ancestor-provided instance
- Updated `composer-main.spec-helpers.ts` to provide the `ResizeHandlerService` mock at module ("route") level instead of overriding it as a component-level provider, mirroring the production fix.
- `npx ng run portal:test-tl --include="projects/portal/src/app/composer/gui/composer-main.component.*.tl.spec.ts"` — 4 files, 77 tests, all passing.
- `eslint` on the three changed files — clean, no errors.

### DIFF
--- a/web/projects/portal/src/app/composer/gui/composer-main.component.resize-handler-di.tl.spec.ts
+++ b/web/projects/portal/src/app/composer/gui/composer-main.component.resize-handler-di.tl.spec.ts
@@ -1,0 +1,76 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ * Regression test for bug #76418: ComposerMainComponent used to declare its own
+ * component-scoped `ResizeHandlerService` provider, which shadowed the composer
+ * route's provider for ComposerMainComponent and its whole descendant subtree
+ * (WSPaneComponent -> WSDetailsPaneComponent -> WSDetailsTableDataComponent, etc).
+ * Because only the route-scoped instance ever had initListeners() called on it
+ * (in ComposerAppComponent.ngOnInit()), the shadowed instance used by the
+ * descendants never received real `window: resize` events, so tables never
+ * re-rendered their virtualized column range after a browser maximize/restore.
+ *
+ * This test asserts that ComposerMainComponent resolves ResizeHandlerService from
+ * an ancestor injector (module/"route" level) rather than creating its own
+ * instance. If the duplicate component-level provider is reintroduced, this test
+ * fails because ComposerMainComponent's injected instance will differ from the
+ * ancestor-provided one.
+ */
+
+import "@angular/compiler";
+import { TestBed } from "@angular/core/testing";
+import { ResizeHandlerService } from "./resize-handler.service";
+import { renderComponent } from "./composer-main.spec-helpers";
+
+beforeAll(() => {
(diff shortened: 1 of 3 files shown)


larryliang-inetsoft:
## Independent review — round 1 (reviewer seat)

Full independent re-verification, redone from scratch in a fresh pass over the worktree (not a re-read of the lead's `04b-verify-r2.md` account).

**Diff**: confirmed cumulative diff = one-line provider removal in `composer-main.component.ts` + new standalone spec file. `composer-main.spec-helpers.ts` shows no net diff (round 2 reverted round 1's change to it back to byte-identical original) — confirmed via `git diff main...HEAD -- .../composer-main.spec-helpers.ts` (empty).

**Mechanism check, at the Angular-internals level, not just the code comments**: read `node_modules/@angular/core/fesm2022/testing.mjs` directly. `OverrideResolver.resolve()` seeds metadata from the class's real, currently-compiled `@Component` decorator via `reflection.annotations(type)`. `set` (`setMetadata`) fully overwrites a prop, discarding that real metadata — this is why round 1's harness was tautological. `remove`/`add` (`removeMetadata`/`addMetadata`) instead filter/append onto the array already seeded from the real metadata, matching entries by a structural hash, not array position. So leaving `ResizeHandlerService` out of both `remove` and `add` in round 2's spec is confirmed, from the framework's own source, to let whatever the real source currently declares for it pass through untouched into the test's DI resolution.

**Independently redid the mutation test myself** (own `Edit`, own test run — not reused from the lead or tester):
1. Ran the new spec with the fix in place: `2 passed (2)`.
2. Re-added `ResizeHandlerService,` to `composer-main.component.ts`'s real providers array myself, reran: `2 failed (2)` — both tests failed as required.
3. Reverted my edit, confirmed `git status --porcelain` empty, reran: `2 passed (2)` again.

This confirms the test is genuinely load-bearing, independently.

**Also checked**: full 4-file `composer-main.component.*.tl.spec.ts` suite — `77 passed (77)`, no regressions. Grepped for other `composer-main.spec-helpers.ts` consumers — only the 3 pre-existing spec files, all pass. Re-verified lifecycle safety fresh: `app.component.html`'s `<composer-main>` is unconditional (no `*ngIf`), zero `router-outlet` usages anywhere in `composer/`, `ResizeHandlerService.initListeners()`/`removeListeners()` are tied to `ComposerAppComponent`'s own `ngOnInit`/`ngOnDestroy` — so `ComposerMainComponent` cannot outlive or be recreated independently of the instance owner. Lint clean on all 3 changed files.

One process note (not a PR defect): found the shared review worktree in a dirty state at the start of my pass — `composer-main.component.ts` had `ResizeHandlerService` mutated back in (uncommitted), leftover from an earlier manual mutation-testing session. Restored it to match `HEAD` before starting my own checks; not a concern for the actual committed/pushed code.

**Verdict: no blockers, no findings. Approving.**

jshobe-inetsoft (approved):

PR #4984, opened 2026-09-03
Bug #76431, guard against a null stream in SecurityChain.loadConfiguration

jkidacki-inetsoft:
## Summary
- `SecurityChain.loadConfiguration()` assumed `DataSpace.exists()` and `DataSpace.getInputStream()` observe a consistent file state, but they are not atomic: a concurrent rewrite of the config file between the two calls makes `getInputStream()` return `null` (its documented behavior when the file has gone missing) instead of throwing.
- Passing that `null` straight into Jackson's `ObjectMapper.readTree()` threw `IllegalArgumentException: argument "in" is null`, observed as an intermittent `PermissionHierarchyTest.setUp` failure in CI (concurrent Surefire test classes sharing one `DataSpace`-backed config file).
- Fix: treat a null stream the same as the existing "all providers failed to load" recovery path — log a warning, retain the existing runtime providers, and don't advance `timestamp`, so the next `dataChanged()` retries once the transient condition clears.

## Test plan
- [x] New `SecurityChainLoadConfigurationTest` reproduces the exact race via a mocked `DataSpace` returning `null` from `getInputStream()`; confirmed it fails with the original `IllegalArgumentException: argument "in" is null` against the pre-fix code, and passes against the fix.
- [x] Full `inetsoft.sree.security.**` package regression: 686 tests, 0 failures, 0 errors (13 pre-existing skips), including `PermissionHierarchyTest` itself.

### DIFF
--- a/core/src/main/java/inetsoft/sree/security/SecurityChain.java
+++ b/core/src/main/java/inetsoft/sree/security/SecurityChain.java
@@ -177,7 +177,7 @@ public abstract class SecurityChain<T extends JsonConfigurableProvider & Cachabl
                   // null rather than throwing (see DataSpace.getInputStream()'s
                   // FileNotFoundException/NoSuchFileException handling) -- Jackson's readTree()
                   // rejects a null stream outright, so guard against it explicitly instead of
-                  // assuming the two calls observe a consistent file state. (76431)
+                  // assuming the two calls observe a consistent file state. See bug #76431.
                   try(InputStream input = dataSpace.getInputStream(null, configFile)) {
                      if(input != null) {
                         root = (ObjectNode) mapper.readTree(input);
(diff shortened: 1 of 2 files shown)


jkidacki-inetsoft:
Addressed both in f48a4ca5e:

1. **Missing/non-array \"providers\" field**: now guarded the same way as the null-stream case — if \`root.get("providers")\` isn't an \`ArrayNode\`, log a warning, retain the existing providers, and don't advance \`timestamp\` so the next \`dataChanged()\` retries. Added \`loadConfiguration_configMissingProvidersArray_doesNotThrowAndLeavesChainUsable\`, which reproduces the original \`NullPointerException: ... "providersArray" is null\` against the pre-fix code and passes with the guard.
2. **Comment style nit**: reworded \`(76431)\` to \`See bug #76431.\`

Full \`inetsoft.sree.security.**\` regression re-run: 687 tests, 0 failures, 0 errors.

jshobe-inetsoft (approved):

PR #4985, opened 2026-09-03
Bug #76442, fix serverTime getter to use per-node serverDateTimeMap

larryliang-inetsoft:
## Summary
- Reverts an accidental regression introduced in the #3763 squash merge: `serverTime` in `summary-monitoring-page.component.ts` had stopped indexing `serverDateTimeMap` by cluster node and instead formatted `serverModel.timestamp` in the browser's local timezone, so switching the selected cluster node had no effect on the displayed server time.
- Restores the pre-#3763 behavior: `serverTime` now reads `serverDateTimeMap[node]` where `node` is the selected cluster node (or `"local"` when clustering is disabled), matching the still-correct `serverUpTime`/`schedulerUpTime` getters in the same file.

## Test plan
- [x] Added `summary-monitoring-page.component.spec.ts` (did not exist before) asserting `serverTime` reads the `"local"` key when clustering is disabled, and switches correctly between per-node values when clustering is enabled.
- [x] Verified the new test fails against the pre-fix getter (temporarily reverted the fix, ran the test, saw both assertions fail with the wrong browser-local-time value) and passes once the fix is restored.
- [x] `npx ng test em --include=".../summary-monitoring-page.component.spec.ts"` passes.

### DIFF
--- a/web/projects/em/src/app/monitoring/summary/summary-monitoring-page/summary-monitoring-page.component.spec.ts
+++ b/web/projects/em/src/app/monitoring/summary/summary-monitoring-page/summary-monitoring-page.component.spec.ts
@@ -1,0 +1,60 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+import { SummaryMonitoringPageComponent } from "./summary-monitoring-page.component";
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
**Round 1 review (independent, per debug-workflow batch process).**

Re-verified from scratch, not just re-reading prior phases' claims:

- Re-ran `git show 1b6ae84fbc -- .../summary-monitoring-page.component.ts` myself — confirms the exact introducing commit and diff (4th independent confirmation across this run's phases, all consistent). The fix is a verbatim revert of that commit's `serverTime` getter change.
- Checked the missing-map-key boundary case (`serverDateTimeMap[node]` undefined for a node not yet reporting): the reverted getter returns `undefined` there (typed as `string`, unlike `serverUpTime`/`schedulerUpTime` which explicitly fall back to `""`/`"00:00:00"`). This is identical to the pre-#3763 behavior though — not something this PR introduces — and Angular's plain-interpolation binding (`{{serverTime}}`, no pipe) renders `undefined` as blank, not literal text. Non-blocking; a good follow-up to harden all three getters consistently someday, but out of scope for this literal-revert fix.
- Verified the direct-instantiation test approach is sound: read the constructor (10 params, parameter-property shorthand only, no DI side effects) and confirmed `selectedClusterNode`/`clusterEnabled` are plain property reads with no Angular machinery in between — TestBed isn't needed here.
- Confirmed sibling `serverUpTime`/`schedulerUpTime` getters are untouched by the regression commit.
- Ran the new spec myself: `npx ng test em --include ".../summary-monitoring-page.component.spec.ts" --watch=false` → 2 passed.

No blocking findings. Approving.

jshobe-inetsoft (approved):

PR #4986, opened 2026-09-03
fix(vsobjects): resolve untranslated _#(to) separator in TimeSlider aria-label

larryliang-inetsoft:
## Summary
- VSFILT-009 testing found a TimeSlider's `thumb-middle` aria-label rendering the literal untranslated key `_#(to)` (e.g. `"undefined _#(to) undefined"`) instead of a translated separator word.
- Root cause: `vs-range-slider.component.html:221` embedded `_#(to)` inside a `[attr.aria-label]="expr1 + ' _#(to) ' + expr2"` JS string-concatenation expression bound via a property binding. Every other place in this codebase that mixes a translated `_#(...)` key with dynamic values does so as static interpolated template text (e.g. `title="_#(Column): {{baseColumn}}"`, `<span>_#(More) ({{count}})</span>`) rather than as a JS-level concatenation inside a bound expression — that's the pattern this fix now follows.
- Fix: changed the binding to a plain interpolated attribute, `attr.aria-label="{{model.labels[model.selectStart]}} _#(to) {{model.labels[model.selectEnd]}}"`, so `_#(to)` sits in resolvable static template text.
- Pre-existing bug (predates the viz-scoped-filters feature by ~3 months, `b667af6807`, 2026-05-29); affects every TimeSlider in the product, not just wiz-created ones.
- Scope note: this fix is intentionally narrow to the `_#(to)` separator. The other half of VSFILT-009 (numeric TimeSlider min/max labels rendering `undefined` — a backend `SelectionList`/`TimeSliderVSAQuery` data gap) is a separate, unrelated defect being tracked/fixed independently.

## Test plan
- [x] Added a regression test in `vs-range-slider.component.display.tl.spec.ts` asserting the `thumb-middle` element's `aria-label` attribute renders as `"B _#(to) D"` (interpolated labels around the static macro) for the default test model.
- [x] `npx ng run portal:test-tl --include="**/range-slider/**/*.spec.ts"` — all 100 tests pass (19 risk + 36 interaction + 45 display, including the new test).

### DIFF
--- a/web/projects/portal/src/app/vsobjects/objects/range-slider/vs-range-slider.component.display.tl.spec.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/range-slider/vs-range-slider.component.display.tl.spec.ts
@@ -519,5 +519,14 @@ describe("VSRangeSlider – display / rendering (P3)", () => {
          expect(maxValue.style["top"]).toBe("46px");
          expect(minValue.style["top"]).toBe("46px");
       });
+
+      // Bug: thumb-middle aria-label rendered the literal untranslated "_#(to)" key
+      // (embedding it inside a bound [attr.aria-label] JS expression bypassed StyleBI's
+      // i18n resolution). The "to" separator must sit in static interpolated template
+      // text, not inside a JS string-concatenation expression, so it's resolvable.
+      it("should render the thumb-middle aria-label with the selected labels around a static '_#(to)' separator", () => {
+         const middle = fixture.nativeElement.querySelector("div.thumb-middle");
+         expect(middle.getAttribute("aria-label")).toBe("B _#(to) D");
+      });
    });
 });
--- a/web/projects/portal/src/app/vsobjects/objects/range-slider/vs-range-slider.component.display.tl.spec.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/range-slider/vs-range-slider.component.display.tl.spec.ts
@@ -519,14 +519,5 @@ describe("VSRangeSlider – display / rendering (P3)", () => {
          expect(maxValue.style["top"]).toBe("46px");
          expect(minValue.style["top"]).toBe("46px");
       });
-
-      // Bug: thumb-middle aria-label rendered the literal untranslated "_#(to)" key
-      // (embedding it inside a bound [attr.aria-label] JS expression bypassed StyleBI's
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
## Closing: not a real defect (round 2 investigation)

Round 1's fix theorized that `_#(to)` embedded inside a JS string-concatenation
`[attr.aria-label]` binding bypasses `LocalizationService`'s i18n resolution, unlike the
same macro in static+interpolated template text. Review round 1 disproved that mechanism
with the pre-fix compiled bundle: the literal, scannable `_#(to)` sits directly inside the
compiled ivy call either way, and `localize()` is a blind character scan with no AST/binding
awareness.

Round 2 supplies the live confirmation the review asked for. `wiz-stylebi` was running an
image built at 2026-09-03T11:54Z — **before** this PR's fix commit (13:51Z) — so it was
serving the untouched, pre-fix TimeSlider code. Fetching the actual served bundle over HTTP
(`curl http://localhost:8000/app/chunk-WWPE6KT4.js`, matched via `thumb-middle`) shows:

```
ctx_r2.model.labels[ctx_r2.model.selectStart] + " to " + ctx_r2.model.labels[ctx_r2.model.selectEnd]
```

`_#(to)` is already correctly resolved to `"to"` — with the old, unfixed code, on a
normally built-and-deployed instance. The raw/compiled resource on disk still contains the
literal macro (expected — resolution happens at serve time via
`TemplateLocalizationController`'s `/app/*.js` interception + `LocalizationService`'s i18n
cache, not at compile time), but nothing ever reaches the browser unresolved.

**Verdict**: no product defect here. The original "undefined `_#(to)` undefined" observation
is not reproducible against a real deployment; it's better explained by the sibling defect in
this same test case (TimeSlider `model.labels` resolving to `undefined` — tracked and fixed
separately) and/or a dev-server bypass of the Spring localization pipeline, not by this
component's i18n binding shape.

Reverted the HTML change and the accompanying test (which review also showed was
tautological — identical output for old and new code) in 2834551ac. Closing without merging.
Full evidence trail: `docs/teams/2026-09-02-test-visualizations-boards/case-vsfilt-009-timeslider-range/02-fix-part1-i18n.md` (Round 2 section) in the `stylebi-wiz` repo.

larryliang-inetsoft:
Closing per the comment above — round 2 investigation found no product defect.

PR #4987, opened 2026-09-03
fix(security): stop identity delete from wiping co-granted identities' permissions (#76448)

larryliang-inetsoft:
## Summary

Redmine #76448: deleting a USER, GROUP, or ROLE silently wiped every other identity's permission grants for the same resource+action, not just the deleted identity's own grant.

**Root cause (two independently-active bugs in `IdentityService.updateIdentityPermission`, `core/src/main/java/inetsoft/web/admin/security/IdentityService.java:2646-2653`):**

1. The delete-case guard compared a plain `IdentityID` against a `Set<Permission.PermissionIdentity>` via `Set.contains()` — a cross-type comparison whose `equals()` always returns `false` for both classes (each rejects any argument that isn't its own type), so the guard never fired. Because `Permission.setGrants()` unconditionally *replaces* the grant set for the action (not a merge/diff), the "keep everyone but the deleted identity" branch never running left an empty set to be installed — wiping every grantee of that identity type for that resource+action.
2. The `.filter()` line immediately below the guard had the identical cross-type bug (`Tool.equals()` falls through to `PermissionIdentity.equals(IdentityID)` for this argument shape), masked as dead code by bug 1 always short-circuiting first. Fixing only the guard while leaving this predicate as-is would have "fixed" the wipe but then removed nothing from the surviving set — silently keeping the deleted identity's own stale grant.

**Fix:** wrap `oldIdentityID` in a `Permission.PermissionIdentity` before comparing, and replace both the `Set.contains()` guard and the `.filter()` predicate with a same-type `equals()` comparison (linear `anyMatch()`/`equals()` scan rather than relying on `Set.contains()`'s hash-bucket dispatch). Also added the missing `Permission.PermissionIdentity.hashCode()` override (`Objects.hash(name, organizationID)`) — independently required regardless of the above, since `PermissionIdentity` instances live inside `HashSet<PermissionIdentity>` fields throughout `Permission.java`, and the missing override breaks same-type `Set.contains()`/`HashSet` dedup wherever it's used, not just this one call site.

## Test plan

- [x] Added `IdentityServiceTest` coverage: `updateIdentityPermission_deleteRole_keepsOtherRoleGrant`, `..._deleteUser_...`, `..._deleteGroup_...` — each reflectively invokes the private `updateIdentityPermission` method against a real `Permission` instance seeded with two co-granted identities, deletes one, and asserts the other survives. No Spring/`AuthorizationProvider` wiring needed since the method is a pure function of its parameters.
- [x] Added `permissionIdentity_hashCode_matchesEqualsContract` — standard equals/hashCode contract test plus a `HashSet.contains()` check.
- [x] **Mutation-tested the regression tests**: reverted just the `IdentityService.java` fix (keeping the `hashCode()` addition) and re-ran — all three delete-one-keeps-other tests failed as expected (`expected: <true> but was: <false>`, co-granted grant wiped), confirming they actually catch the bug rather than passing vacuously. Restored the fix and re-ran — all 13 tests in the class pass.
- [x] Ran the wider affected suite (`FileAuthorizationProviderTest`, `PermissionCheckerTest`, `PermissionIdentityTest`, `PermissionTest`, `RecycleBinOrgLifecycleTest`) — 96 tests, 0 failures, 1 pre-existing skip.
- [x] Full `community/core` module test run in progress as a broader regression check.

Follow-up: `stylebi-enterprise`'s outer submodule pointer needs to bump to include this commit — tracked as a separate companion PR per this batch's convention.

### DIFF
--- a/core/src/main/java/inetsoft/sree/security/Permission.java
+++ b/core/src/main/java/inetsoft/sree/security/Permission.java
@@ -1134,6 +1134,11 @@ public class Permission implements Serializable, Cloneable, XMLSerializable {
          return false;
       }
 
+      @Override
+      public int hashCode() {
+         return Objects.hash(name, organizationID);
+      }
+
       boolean equalsIgnoreCase(PermissionIdentity other) {
          if(other == null) {
             return false;
--- a/core/src/main/java/inetsoft/web/admin/security/IdentityService.java
+++ b/core/src/main/java/inetsoft/web/admin/security/IdentityService.java
@@ -2646,9 +2646,15 @@ public class IdentityService {
 
       if(orgScopedGrants != null) {
          // remove identity
-         if(newIdentityID == null && orgScopedGrants.contains(oldIdentityID)) {
+         Permission.PermissionIdentity oldPermissionIdentity =
+            oldIdentityID == null ? null : new Permission.PermissionIdentity(oldIdentityID);
+
+         boolean oldIdentityGranted = oldPermissionIdentity != null &&
+            orgScopedGrants.stream().anyMatch(pid -> pid.equals(oldPermissionIdentity));
+
+         if(newIdentityID == null && oldIdentityGranted) {
             orgScopedGrants.stream()
-               .filter(identityID -> !Tool.equals(identityID, oldIdentityID))
+               .filter(identityID -> !identityID.equals(oldPermissionIdentity))
                .forEach(identityID -> grants.add(identityID));
          }
          // sync id
--- a/core/src/test/java/inetsoft/web/admin/security/IdentityServiceTest.java
+++ b/core/src/test/java/inetsoft/web/admin/security/IdentityServiceTest.java
@@ -20,7 +20,10 @@ package inetsoft.web.admin.security;
 import inetsoft.sree.security.AuthenticationProvider;
 import inetsoft.sree.security.IdentityID;
 import inetsoft.sree.security.IdentityInfo;
+import inetsoft.sree.security.Organization;
 import inetsoft.sree.security.OrganizationContextHolder;
+import inetsoft.sree.security.Permission;
+import inetsoft.sree.security.ResourceAction;
 import inetsoft.uql.util.Identity;
 import inetsoft.uql.asset.AssetEntry;
 import inetsoft.uql.asset.AssetRepository;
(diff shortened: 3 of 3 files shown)


larryliang-inetsoft:
## Reviewer round 1 — approved

Reviewed the actual diff (`gh pr diff 4987`), not just the write-up in `docs/teams/2026-09-03-bug-76448-permission-wipe/bug-76448/03-fix.md`.

**Both fix parts present and correct**, confirmed against `IdentityService.java:2646-2661`:
- The delete-case guard now wraps `oldIdentityID` in a same-type `Permission.PermissionIdentity` and checks membership via `orgScopedGrants.stream().anyMatch(pid -> pid.equals(oldPermissionIdentity))` — no longer the broken cross-type `Set.contains(oldIdentityID)`.
- The `.filter()` line one below it (the one `02-refute.md` specifically flagged as easy to overlook, since it carried the identical cross-type bug masked as dead code) is also fixed: `.filter(identityID -> !identityID.equals(oldPermissionIdentity))`, no longer `Tool.equals(identityID, oldIdentityID)`. Confirmed by reading the diff directly — neither line was missed.

**`hashCode()`/`equals()` contract checked, not assumed.** Read `PermissionIdentity.equals()` on the base branch (`Permission.java:1125-1135`): compares `this.name`/`this.organizationID` against the other instance's same fields via `Tool.equals(a, b)` (2-arg form, `sensitive=true` by default — confirmed in `CoreTool.java:218-220,225`, so this is plain case-sensitive `String.equals()`, not the case-insensitive 3-arg overload). The new `hashCode()` (`Objects.hash(name, organizationID)`) hashes exactly those same two fields with the same case sensitivity. Contract holds — no mismatch risk.

**Mutation-test claim traced by hand**, not just trusted: pre-fix, `grants` is initialized to an empty `HashSet<>()` before the `if`; the guard `orgScopedGrants.contains(oldIdentityID)` is a hash-bucket lookup dispatched on the *argument's* (`IdentityID`) `hashCode()`/`equals()` — `IdentityID.equals()` rejects non-`IdentityID` arguments, so it's unconditionally `false` regardless of hash collisions, meaning the guard never fires and `grants` stays empty; `setGrants()` then unconditionally installs that empty set, wiping every co-granted identity. This matches the claimed pre-fix 3-test failure exactly (`RoleB`/`bob`/`GroupB` all reported wiped). Post-fix, the `anyMatch`/same-type-`equals()` guard and filter both work correctly and would keep the sibling grant. Logically confirms the fail→pass claim without needing to re-run Maven myself.

**Pre-existing/unrelated test failure spot-checked**: fetched `core/src/test/java/inetsoft/uql/ConditionTest.java` off `stylebi-wiz-main-rebase` directly. `RegressionTests.toSqlConditionSilentlyDefaultsOnAnUnmatchedName` (line ~1475) compares two independently-constructed `Condition.toNullSqlCondition(1)`/`toSqlCondition(...)` results — a hardcoded "one year ago" `java.sql.Date` range computed twice, which is exactly the kind of test that flakes near a date/timezone boundary as described. It's in SQL condition-to-string translation code, has nothing to do with `Permission`/`IdentityService`, and this diff never touches that file or package. Confirmed unrelated.

**CLAUDE.md compliance**: diff is minimal and scoped to the two identified defects (comparison-logic fix; `hashCode()` addition) plus a regression test file — no unrelated refactors, no dead code left behind, no unnecessary comments (the one comment added on the tests explains non-obvious cross-type-comparison root cause, not restating the code).

No blocking issues found. Approving and merging (squash) — flagging that formal `gh pr review --approve` is blocked here since this account (`larryliang-inetsoft`) is also the PR author; posting this as a comment per this batch's established workaround, then merging directly.

PR #4989, opened 2026-09-03
Bug #75834, pin the docker image tag in the packaged release examples

jkidacki-inetsoft:
## Summary

`community-examples.zip`, attached to every GitHub release, is zipped straight from the working tree — so `docker-compose.yaml` and `.env` inside it always reference the `latest` Docker image tag instead of the tag for the release the user actually downloaded. Since this zip is the standalone artifact people download from the Releases page and run with `docker compose up -d` (no build step), a past release's zip permanently points at `latest`, which drifts to whatever the newest image is once future releases ship.

## Root Cause

The "Package community examples" step in `.github/workflows/publish-release.yml` zips `community-examples/` verbatim, with no substitution of the image tag.

## Fix

Before zipping, substitute the actual release version (from the pushed tag) into the packaged copy of `docker-compose.yaml` and `.env`, leaving the committed source files (used when cloning the repo) referencing `latest` as before.

## Test Plan

- Simulated the `sed` substitution against copies of the real `community-examples/docker-compose.yaml` and `.env` for both a standard release tag (`v1.1.0`) and a prerelease tag (`v1.1.0-beta-1`) — confirmed the image tag is pinned correctly in all 4 target lines, while the unrelated `busybox:latest` image and the commented-out nightly-build line in `.env` are left untouched.
- Validated the modified workflow YAML parses correctly.
- Verified the resulting version matches the Docker tag actually pushed by `docker/pom.xml`'s jib config for the same release tag.

Fixes Redmine #75834.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -128,7 +128,10 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Package community examples
-        run: (cd community-examples && zip -r ../community-examples.zip .)
+        run: |
+          version="${GITHUB_REF_NAME#v}"
+          sed -i "s|stylebi-community:latest|stylebi-community:${version}|" community-examples/docker-compose.yaml community-examples/.env
+          (cd community-examples && zip -r ../community-examples.zip .)
 
       - name: Create GitHub release
         run: >


larryliang-inetsoft (changes requested):
No Claude automated review has been posted on this PR (the `claude-review` check reports success but no review comment exists). Please ensure the automated review runs and posts its findings before this can be merged.

jkidacki-inetsoft:
## Review: Bug #75834 — pin the docker image tag in packaged release examples

### Change summary
`.github/workflows/publish-release.yml`'s "Package community examples" step now runs a `sed` substitution on `community-examples/docker-compose.yaml` and `community-examples/.env` (replacing `stylebi-community:latest` → `stylebi-community:${version}`, where `version="${GITHUB_REF_NAME#v}"`) before zipping, so the released `community-examples.zip` pins the image tag to the release rather than `latest`.

### Correctness
- Verified against the actual file contents on this branch: `docker-compose.yaml` has 3 occurrences of `.../stylebi-community:latest` (`storage`, `server`, `scheduler` services) and `.env` has 1 active occurrence (`STYLEBI_DOCKER_IMAGE=...:latest`) — 4 total, matching the PR description.
- `sed`'s `s|old|new|` does a substring match, so it correctly rewrites the `...:latest` suffix even though the full string is `ghcr.io/inetsoft-technology/stylebi-community:latest`.
- Both files are patched with the same command, so the `.env` value and the `docker-compose.yaml` fallback default (`${STYLEBI_DOCKER_IMAGE:-...}`) stay consistent with each other.
- Confirmed the untouched lines are correctly left alone: `busybox:latest` (different image, no `stylebi-community:` prefix) and the commented-out nightly build line in `.env` (`...stylebi-community:1.0`, no `:latest`).
- Cross-checked against `docker/pom.xml`'s jib config: the pushed image tag is `1.1.${docker.build.version}`, where `docker.build.version` is overridden to `steps.version.outputs.patch` (tag with the `major.minor.` prefix stripped). For a tag like `v1.1.0`, this reconstructs to `1.1.0`, which matches this PR's `version="${GITHUB_REF_NAME#v}"` (`1.1.0`) exactly. Same reasoning holds for a prerelease tag (`v1.1.0-beta-1` → `1.1.0-beta-1` on both sides).

### Side effects / regressions
- `sed -i` modifies the checked-out working tree files directly rather than a separate copy, but since this is an ephemeral release-runner checkout with no subsequent `git commit`/`git push` step, the committed source files in the repo (used when cloning) are unaffected — behavior matches what the PR description promises.
- No other steps in the workflow read `community-examples/docker-compose.yaml` or `.env` after this step, so there's no ordering hazard.
- No application code, tests, or other workflows are touched — blast radius is limited to the release packaging step.

### Security
- `version` is derived from `GITHUB_REF_NAME`, which for this workflow's trigger (`push: tags: 'v*.*.*'`) is a maintainer-controlled release tag, not untrusted external input — no injection concern with embedding it in the `sed` expression.

### Tests / CI
- No unit-testable code changed (workflow YAML only); the PR author's manual `sed` simulation against the real files is a reasonable substitute given there's no CI harness for testing GitHub Actions workflow logic in this repo.
- All required checks are green (`claude-review`, `Semgrep Scan`, `build`, `Semgrep OSS`, `license/cla`), and the PR has no merge conflicts.

### Verdict
No correctness, regression, security, or backward-compatibility issues found. The existing `CHANGES_REQUESTED` review from another reviewer was solely about the automated Claude review not having posted a comment — this review supersedes that concern with a full manual analysis. Approving.

jshobe-inetsoft (approved):

PR #4991, opened 2026-09-03
Bug #75780, size input labels from the label font on export

ngovus-inetsoft:
## Problem

TextInput/ComboBox labels (`LabelInfo`) render differently in exports than in the viewsheet preview. In Excel the label wraps mid-word ("lab / el"), sits at the wrong distance from the widget, and overlaps it when the label gap is small.

Reproduced with a viewsheet of 8 assemblies (4 TextInput, 4 ComboBox) at 100×20px covering all four label positions with gaps 0 and 5, and a label font of `Default-PLAIN-24`. The large font is what exposes it.

## Root cause

`AbstractVSExporter.getLabelDimensions()` sized the label box with a fixed `AssetUtil.defh` (20px) row height regardless of the label font, while the browser derives it from the font (`line-height: 1` on `.vs-input-label`). A 24pt label got ~10px less room than it needs, which threw off the split, the vertical centering and the TOP/BOTTOM bounds expansion in every export format.

`HTMLVSExporter.adjustLabelBoundsForFont()` already worked around the same defect locally — evidence the shared helper was the wrong place to hard-code a height.

## Changes

**Shared geometry** (`AbstractVSExporter`)
- Label height is derived from the label font, matching `LabelInfo.getRenderedHeight()` which `Viewsheet`, `TabVSAssemblyInfo` and `GroupContainerVSAssembly` already use.
- Removed the `Math.max(0, …)` clamp on LEFT/RIGHT vertical centering so a label taller than the widget row overhangs it, as flexbox `align-items: center` does.
- `adjustSizeForInputLabels()` accounts for that overhang so the page is not cut through the label.

**Report converter** (`VsToReportConverter.addInputLabel`)
- Same font-derived height. It is also no longer multiplied by `scalefont`: `applyScaleFont()` only reaches formats in `FormatInfo` and `LabelInfo` carries its own standalone format, so the label font is never scaled — the same reason the label width beside it is unscaled.

**HTML** (`HTMLVSExporter`)
- `adjustLabelBoundsForFont()` replaced by a narrower fallback covering only the case the exporter cannot measure: a label with no font of its own, which the browser renders in its inherited font. `writeText` emits the box height with `overflow:hidden`, so that case still needs a floor.

**Excel** (`PoiExcelVSExporter`)
- `setWordWrap(false)` on the label and TextInput value boxes to match `white-space: nowrap` / a single-line `<input>`.
- Zero the text-box insets: Excel defaults to a 0.1in (~9.6px) internal margin that shifted the label text off the computed position, into the widget.
- `splitInputBoundsForExcel()` pushes the widget over by the amount Excel draws a LEFT label wider than the pixel-based split reserved. The workbook carries the font in points (`VSFontHelper.getFontSize`, with a 9pt floor for small sizes) and Excel lays those out at 96 dpi. RIGHT/TOP/BOTTOM overflow away from the widget and are left alone.
- Removed the old `extra = LEFT ? min(30, gap) : 30` width fudge, which was worth 0–5px for LEFT labels and did not scale with font size.

## Behavior change to be aware of

Labels using the **default** font (PLAIN 11) now reserve ~13px instead of 20px, so TOP/BOTTOM-labelled inputs sit tighter in every existing export. This moves toward the preview (the browser reserves 11px), but it is a visible change for existing assets.

## Testing

- New `AbstractVSExporterLabelTest` (8 cases): label height tracks the font, LEFT/RIGHT centering and overhang, TOP/BOTTOM bounds expansion, LEFT/RIGHT non-expansion, default-font fallback.
- New `PoiExcelVSExporterLabelSlackTest` (4 cases): pins the point-size conversion including the 9pt floor. Only the scale is exercised — measuring text goes through `Common.stringWidth`, which reads `SreeEnv` properties and needs a bootstrapped server this module's tests do not have.
- Verified end to end against the reported viewsheet: exported to Excel and compared all 8 assemblies against the preview — no mid-word wrapping, gaps match, no label/widget overlap.

### DIFF
--- a/core/src/main/java/inetsoft/report/io/viewsheet/AbstractVSExporter.java
+++ b/core/src/main/java/inetsoft/report/io/viewsheet/AbstractVSExporter.java
@@ -3888,8 +3888,9 @@ public abstract class AbstractVSExporter implements VSExporter {
    }
 
    /**
-    * Get the font to use for rendering an input label. The default size 11
-    * matches LabelInfo.getRenderedHeight() so drawing and measurement agree.
+    * Get the font to use for rendering an input label. This is the single font
+    * source for both drawing and measurement (getLabelDimensions), and matches
+    * the fallback used by LabelInfo.getRenderedHeight().
     */
    protected static Font getLabelFont(LabelInfo labelInfo) {
       VSCompositeFormat fmt = labelInfo.getLabelFormat();
@@ -3954,7 +3955,7 @@ public abstract class AbstractVSExporter implements VSExporter {
          // Center the label text vertically relative to the widget height,
          // matching the Angular preview which uses flexbox vertical centering.
          labelBounds = new Rectangle2D.Double(x + fullW - labelW,
-            y + Math.max(0, (fullH - labelH) / 2.0), labelW, labelH);
+            y + (fullH - labelH) / 2.0, labelW, labelH);
          widgetBounds = new Rectangle2D.Double(x, y,
             Math.max(0, fullW - labelW - gap), fullH);
          break;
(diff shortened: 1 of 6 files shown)


larryliang-inetsoft (changes requested):
Claude's automated review found no blocking issues, but the `build` check on this PR shows as CANCELLED rather than passing. Please re-run/fix the build check before this can be merged.

jshobe-inetsoft (approved):

PR #4992, opened 2026-09-03
Bug #76441, surface cloud vault credential resolution failures

ngovus-inetsoft:
## Problem

`AbstractSecretsManager.getCredential()` wrapped the vault fetch *and* the Jackson parse in a single `catch(Exception)`, logged a context-free `"Failed to decrypt password"`, and returned `null`:

```
ERROR i.u.AbstractSecretsManager: Failed to decrypt password
com.fasterxml.jackson.core.JsonParseException: Unexpected character ('p' (code 112)): was expecting double-quote to start field name
    at inetsoft.util.AbstractSecretsManager.getCredential(AbstractSecretsManager.java:66)
    at inetsoft.util.credential.CloudCredential.fetchCredential(CloudCredential.java:29)
    at inetsoft.util.credential.AbstractCloudCredential.parseXML(AbstractCloudCredential.java:131)
    at inetsoft.uql.jdbc.JDBCDataSource.parseXML(JDBCDataSource.java:928)
```

`CloudCredential.fetchCredential()` then dropped the `null` with no log at all, so `parseXML()` completed "successfully" with `user`/`password` left null. `DataSourceRegistry.getObject()` cached the data source in that broken state until the storage timestamp changed, and the authentication failure only surfaced much later at JDBC connect time with nothing tying it back to the vault content.

Three distinct failure modes were funnelled into that one catch, not just the reported malformed JSON:

1. **Malformed / non-JSON payload** — `readTree` throws (the reported stack).
2. **Fetch failure** — every provider's `decryptPassword()` logs and returns `null` (`AwsSecretsManagerPasswordEncryption:124`, and the Azure/Google/HCP equivalents), so `readTree(null)` threw inside the same catch. A vault outage was indistinguishable from a corrupt payload.
3. **Secret that parses but carries nothing usable** — `{}`, or fields that don't match the credential type (the deserializers ignore unknown keys). This produced a blank credential and was treated as success.

This is the community-core follow-up that inetsoft-technology/stylebi-enterprise#746 (Bug #76425) explicitly deferred.

## Changes

**`AbstractSecretsManager.getCredential()`** — the three failure modes are handled separately and each throws `SecretsUnavailableException`. Every message carries the secret id (a vault reference, safe to log), the credential class, the dbType and which secrets manager. The payload is never logged. The empty-value check runs *before* `setId`, so `isEmpty()` reports on the values the secret actually carried.

**`SecretsUnavailableException`** (new) — extends `MessageException` at `LogLevel.ERROR, dumpStack=false`, since the condition is already logged with full context at the source.

**`CloudCredential` / `AbstractCloudCredential`** — a "credential unavailable" state, distinct from an empty id, which means no credential is configured. It is `transient` (runtime-only; it must not reach XML, blob storage or cluster serialization, where it would go stale), is cleared when the secret id changes, and is excluded from `equals`/`isEmpty`/`writeXML` so it cannot cause spurious "modified" detection or connection-pool key churn. `fetchCredential()` never propagates, so the data source still loads and stays listable and editable in EM. `ensureCredentialAvailable()` re-fetches at most once per 30s, single-flighted **without blocking** the other callers — a hung vault must not pile every thread up behind it.

**`JDBCHandler`** — retry-then-fail-fast on both connect paths, which are disjoint (`connect0` for test-connection, private `getConnection` for pooled queries). Guarded by `isRequireLogin() && user/password empty`, so prompted logins, `VariableTable` params and `XPrincipal` parameters are unaffected. The retry runs on the *resolved* data source (an additional connection resolves to a different object with its own credential) and before the clone, since `clone()` deep-copies the credential.

**`connect()` was swallowing the exception** — its catch only retried for the Vault DB secrets engine and otherwise fell through returning `conn == null`, so `testDataSource` replaced the real cause with a generic `SQLException`. Without this, the new message never reached the UI. Narrowed to `MessageException` so ordinary SQL errors keep today's friendly "Failed to connect to datasource[X]!" message.

**Catalog** — `common.datasource.credentialUnavailable`, deliberately naming only the data source: the secret id can carry deployment detail (an ARN embeds the account id) and goes to the log instead. The wording avoids "username"/"password" because `DataSourceStatusService.getStatusModel():160` collapses any message containing those into a generic login error. The existing `XDataSource.Status` surface therefore renders it in the portal Data Sources browser with **no UI change**.

## Testing

New `AbstractSecretsManagerTest` (12) and `CloudCredentialTest` (9) — 21 cases, no live vault required, covering all three failure modes, the id-change reset, the retry throttle and the single-flight guard. There was previously **no** test anywhere for `getCredential`, `fetchCredential` or `AbstractCloudCredential.parseXML`.

```
Tests run: 21, Failures: 0, Errors: 0, Skipped: 0   (new tests)
Tests run: 121, Failures: 0, Errors: 0, Skipped: 0  (inetsoft.util + inetsoft.uql.jdbc)
BUILD SUCCESS
```

The enterprise `integration/` modules (AWS, Azure, Google) build clean against the changed signature.

## Not addressed here

- `GoogleSecretsManagerPasswordEncryption:174` is still missing the legacy `\aes` guard that AWS/Azure/HCP have — enterprise repo, flagged by #746 for separate triage.
- `AbstractCloudCredential.writeXML()`'s `isEncryptForceLocal()` branch silently exports an empty local credential when `Tool.decryptPasswordToCredential` returns null. Same class of bug, different path; making export throw would break asset export deployment-wide.
- Non-JDBC tabular connectors get no fail-fast check (per-connector `TabularHandler.testDataSource`). Unchanged from today; the improved log with the secret id is the win there.
- `AbstractSecretsManager.decryptPassword(String, String)` discards `encryptedKey` for every provider. Pre-existing, flagged by #746.
- The state is `transient`, so a data source Java-serialized to another cluster node arrives with the flag cleared and degrades to today's behavior on that node. Deriving it lazily would need a per-type notion of "resolved" that the generic `Credential` interface does not have.

Community only, no enterprise changes.

### DIFF
--- a/core/src/main/java/inetsoft/uql/jdbc/JDBCHandler.java
+++ b/core/src/main/java/inetsoft/uql/jdbc/JDBCHandler.java
@@ -35,6 +35,8 @@ import inetsoft.uql.table.XObjectColumn;
 import inetsoft.uql.table.XTableColumnCreator;
 import inetsoft.uql.util.*;
 import inetsoft.util.*;
+import inetsoft.util.credential.CloudCredential;
+import inetsoft.util.credential.PasswordCredential;
 import inetsoft.util.log.LogManager;
 import inetsoft.web.admin.monitoring.MonitorLevelService;
 import org.slf4j.Logger;
@@ -1487,6 +1489,9 @@ public class JDBCHandler extends XHandler {
    public void testDataSource(XDataSource datasource,
                               VariableTable params) throws Exception
    {
+      // clone() deep copies the credential, so the retry has to happen on the original for the
+      // cached data source to be healed rather than a throwaway copy
+      refreshUnavailableCredential((JDBCDataSource) datasource);
       Connection conn = connect((JDBCDataSource) datasource.clone(), params);
 
       if(conn == null) {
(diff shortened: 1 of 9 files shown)


larryliang-inetsoft (approved):
Claude's automated review traced the full call chain end-to-end and found the fix well-scoped and well-tested (21 new tests, 121 total passing). One minor non-blocking observation noted (redundant vault retry in a narrow edge case) but no security/regression/backward-compatibility concerns. All checks passing, no conflicts. Approving.

PR #4994, opened 2026-09-04
Bug #76451: guard SecurityChain.loadConfiguration() against a config-file read race

larryliang-inetsoft:
## Summary
- Found while investigating a `build` check failure blocking PRs #4973 and #4980, both hitting `PermissionHierarchyTest.setUp:96 » IllegalArgument argument "in" is null` in CI, unrelated to either PR's own diff.
- Root cause: `SecurityChain.loadConfiguration()` checks `DataSpace.exists()` then separately calls `DataSpace.getInputStream()`, passing the result straight into `ObjectMapper.readTree()` with no null check. `getInputStream()` deliberately returns `null` (not an exception) when the file is missing. If the config file is removed/rewritten by a concurrent writer between the two calls, `readTree(null)` throws an uncaught `IllegalArgumentException` that crashes `AuthenticationChain`'s/`AuthorizationChain`'s constructor entirely — not caught by `initialize()`'s `catch(IOException e)`.
- Fix: treat a null `InputStream` the same as the existing "temporarily unreadable config" resilience path a few lines below (already handles a GKE-node-restart-style transient failure) — log a warning, keep existing runtime providers, don't advance `timestamp` so the next `dataChanged()` naturally retries.

## Test plan
- [x] New `SecurityChainTest` (`AuthenticationChain`/`AuthorizationChain`, via `Mockito.mockStatic(DataSpace.class)` simulating `exists()==true` + `getInputStream()==null`) — proven load-bearing: fails with the exact reported `IllegalArgumentException: argument "in" is null` when the fix is reverted, passes when restored.
- [x] Full `inetsoft.sree.security.**` package (36 test classes) — all pass, no regressions.

Redmine: #76451

### DIFF
--- a/core/src/main/java/inetsoft/sree/security/SecurityChain.java
+++ b/core/src/main/java/inetsoft/sree/security/SecurityChain.java
@@ -172,6 +172,20 @@ public abstract class SecurityChain<T extends JsonConfigurableProvider & Cachabl
                   ObjectNode root;
 
                   try(InputStream input = dataSpace.getInputStream(null, configFile)) {
+                     if(input == null) {
+                        // The exists() check above raced with a concurrent delete/rewrite of
+                        // the config file; DataSpace.getInputStream() treats a missing file as
+                        // a soft null rather than throwing. Treat this exactly like the
+                        // "temporarily unreadable config" case below: keep the existing runtime
+                        // providers and do not advance timestamp, so the next dataChanged()
+                        // retries once the transient condition clears.
+                        LOG.warn(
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:

larryliang-inetsoft:
Closing as a duplicate — independently found and fixed as bug #76431 in #4984, already merged to main (and even more thorough: also guards against a missing/malformed `providers` array). Redmine #76451 will be closed as a duplicate of #76431.

PR #4995, opened 2026-09-04
Make Google Spreadsheets the ninth TabularCatalogProvider implementer

Nelson-cao:
`GDataRuntime` did not implement `TabularCatalogProvider`, so `TabularCatalogService.resolveProvider` threw `UnsupportedDatasourceException` for every `GoogleDocs` data source and wiz's METADATA annotation path answered 422.

`GDataRuntime` now implements the SPI. `listDatasets` enumerates spreadsheet x sheet - Drive `files().list()` paged to exhaustion for level 1, `spreadsheets().get()` per spreadsheet for level 2 - and `describeDataset` reports one column per surviving header cell.

Sheets is the first connector on this SPI where "a column has a type" is not a concept the source has. Types live on cells: the same stored `45000` is a date or the integer 45000 depending only on the cell's number format. So the metadata call returns no column list at all, and columns come from reading the header row plus three data rows in one bounded range.

Non-obvious decisions, each with its own test:

- **`gridProperties` is deliberately absent from the field mask.** `columnCount` is the grid width, not the data width - a new sheet is 1000 x 26 empty. Omitting it from the mask makes the trap unreachable rather than merely un-taken, and a test pins the mask so a future edit cannot quietly re-add it.
- **A column whose header cell is empty, error-valued, non-string, blank after trimming, or a duplicate of an earlier name is dropped**, and `columnsMayBeIncomplete` is set. A synthesized `A`/`B`/`C` name would claim a column `runQuery` names `null`. Duplicates keep the first occurrence, because wiz's annotation merge resolves a name to its first match, so the surviving catalog column and the column the merge annotates are then the same column.
- **The A1 range quotes the sheet title and doubles an embedded `'`.** `runQuery` concatenates the title raw, so it fails today for a sheet named `Sales 2024`; the catalog does not inherit that.
- **Non-GRID sheets are excluded.** A chart-only `OBJECT` sheet has no `gridProperties`, so listing it would yield a target whose `describeDataset` could only throw.
- **`isDimension` is always null.** Sheets declares no dimension/measure split, and the record's contract forbids inferring one from the type.
- **A `spreadsheetId` scalar alias** on `GDataQuery`, outside `@View` and carrying no `tagsMethod`, gives the identity a fillable route: `GooglePicker` is a plain bean, so `compositeElementsOf` yields null and the `spreadsheet` param is omitted from the query schema entirely.

No SPI type is touched - this round is a consumer. The diff is confined to `connectors/inetsoft-googledoc/`.

77 tests, 0 failures. `core` plus the seven other `TabularCatalogProvider` connectors re-run green (8712 in core, 0 failures).

There is no Google account in this environment, so `listDatasets` and `describeDataset` have not run end to end. Three mechanism assumptions and the Drive OAuth scope question are recorded as a checklist that closes in one call each - see `stylebi-wiz/docs/tabular/stylebi-tabular-wiz-integration.md` section 5.3 entries 37-47.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/connectors/inetsoft-googledoc/src/main/java/inetsoft/uql/gdata/GDataCatalog.java
+++ b/connectors/inetsoft-googledoc/src/main/java/inetsoft/uql/gdata/GDataCatalog.java
@@ -28,9 +28,10 @@ import java.util.*;
 
 /**
  * Assembles {@link GDataRuntime}'s {@link TabularCatalogProvider} answers out of the connector's
- * own Drive/Sheets API calls (via {@link GDataRuntime}'s package-private static fetch methods --
- * the mocking seam, 01-design.md Part D.15). Mirrors {@code AnalyticsCatalog}'s role:
- * package-private, static methods, no state of its own.
+ * own Drive/Sheets API calls, via {@link GDataRuntime}'s package-private static fetch methods --
+ * kept on {@code GDataRuntime} rather than here so {@code GDataCatalogTest} can
+ * {@code mockStatic(GDataRuntime.class)} without touching a line of the code under test. Mirrors
+ * {@code AnalyticsCatalog}'s role: package-private, static methods, no state of its own.
  *
  * No caching: the two SPI phases share nothing that stays fresh across calls. A sheet's TITLE
  * (fetched by {@code listDatasets}) goes stale the moment a user renames it, and X6 (a
@@ -41,9 +42,11 @@ import java.util.*;
  * Sheets is the first connector on this SPI where "a column has a type" is not a source concept:
  * types live on CELLS, not columns (a date is a number plus a number format). The column *list*
  * comes from the header row (a declaration, per {@link TabularCatalogProvider#describeDataset}'s
- * own javadoc); the column *types* come from sampling 3 data rows (charter Q4). See
- * 01-design.md Part D.9 for why {@link TabularDatasetSchema#columnsMayBeIncomplete()} can only
- * express the list half of that split, not the type half.
+ * own javadoc); the column *types* come from sampling 3 data rows below it. That split means
+ * {@link TabularDatasetSchema#columnsMayBeIncomplete()} can only express one half of it -- a
+ * dropped column really is "the list may be missing something", but a sampled TYPE is a different
+ * kind of fact than an incomplete list, and this flag has no separate slot for it. See stylebi-wiz
+ * {@code docs/tabular/stylebi-tabular-wiz-integration.md} §5.3 no. 38.
  */
 final class GDataCatalog {
    private GDataCatalog() {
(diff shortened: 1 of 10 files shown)


Nelson-cao:
Thanks — five observations, and I checked each against the code rather than taking the summary at face value. **Two of them are real and I am fixing them; three stand as recorded, with reasons.**

### Fixing

**1 · `files(id,name)` over-fetches `name`.** Confirmed: `File.getName()` is read nowhere. The only `getName()` calls in the connector are `ds.getName()` (the data source's own name, in the two log messages) and an unrelated one in `getValidHeaderName`. The design justified fetching `name` as "the human-readable handle for the log message when a spreadsheet's level-2 call fails" — but there is no such log, because a level-2 failure propagates rather than being caught per spreadsheet, which is what the SPI contract requires. **So the justification never materialised and the field is dead. Dropping it from the mask.**

Also worth noting the gap this exposes: `listSheetProperties` and `fetchSampleRows` each have a test pinning their field mask, and `listSpreadsheetFiles` does not — which is why this drifted quietly. The paging test already runs that method's real body, so the mask assertion goes there and the third mask gets pinned too.

**4 · The javadocs cite a run archive in another repository.** Ten citations to `01-design.md` / `03-reconcile.md` / `00-charter.md`, which live in `stylebi-wiz/docs/teams/<run>/`. A reader of this repo alone cannot resolve them, and that archive is explicitly documented as living "as long as the run is interesting" — so production javadoc is pointing at something both cross-repo and impermanent.

Checked whether this is established convention: **it is not.** The GA4 connector (`inetsoft-google-analytics-ga4`), the closest precedent, cites no run-archive file and no wiz document at all — its javadocs are self-contained. The same grep returns ten hits here and zero there, so the pattern works and this is drift introduced by this PR, not a house style. **Making them self-contained**: the reasoning is already inline in almost every case, so the pointer is what goes. Where a durable cross-repo reference is genuinely useful it becomes the numbered entry in `stylebi-wiz/docs/tabular/stylebi-tabular-wiz-integration.md` §5.3, which is the standing ledger for exactly this and is not tied to one run.

### Not fixing, and why

**2 · N+1 in `listDatasets`.** Correct, and already recorded rather than overlooked — §5.3 no. 45 in the integration doc puts the cost at `1 + ceil(F/1000) + F` calls for F spreadsheets and names the quota risk explicitly (Sheets' per-user read limit is the one that bites first, and `google-api-client` does not retry 429 by default, so the call throws rather than silently truncating). There is no batch "sheets for many spreadsheets" endpoint, so this is the shape of the API split, not an implementation choice. Backoff belongs with the scale-annotation work, not here.

**3 · Token refresh via `getSheets(ds, false)`.** Accurate, and slightly stronger than stated: `GDataRequestInitializer.initialize` calls `dataSource.updateTokens(tokens)` unconditionally on an expired token, and `saveTokens` gates only the `XRepository.updateDataSource` write — so the in-memory mutation happens either way. And `listSpreadsheetFiles` goes through `getDrive(ds)`, which is `saveTokens = true`, so the catalog path can also **persist**. That is exactly what `testDataSource` does today through the same method, which is what makes it fall inside the contract's own wording, "must not mutate the passed data source beyond what a normal connection already does". Pre-existing and shared with `runQuery`/`testDataSource`; changing it is a connector-wide decision, not this PR's.

**5 · Untested end to end.** Agreed, and it is the honest ceiling rather than an oversight: there is no Google account in this environment. The open items are deliberately written as a checklist that closes in one call each — §5.3 no. 37 (the row-only A1 range, the trailing-empty-cell assumption, the Drive parameter pairing) and no. 47 (the OAuth scope). The scope one is the one to watch before shipping to a customer: if `google-sheets-picker` is configured with `drive.file`, `files().list()` returns only files this app opened, so the catalog would be a silent subset of the user's spreadsheets — a failure that looks like success.

Pushing the two fixes shortly.

Nelson-cao:
Both fixes confirmed landed, thanks for re-verifying them against the actual files rather than the commit messages.

**On the six leftover citations in test comments — fixing, and the miss was mine.** My instruction to the author scoped the change to *production javadoc*, so the test files were never in scope; this is an under-specified instruction, not something that was skipped. You are right that the impermanence argument does not stop at the `src/main` boundary: a reader of this repo's test suite alone cannot resolve `01-design.md` or `03-reconcile.md` either.

Ordinarily I would leave a nit alone — that is what keeps a review loop from running forever. This one is different in a way worth naming: it is an inconsistency in a principle *this PR itself just established*, it touches comments only, and a half-applied principle reads worse than one never applied, because the next person cannot tell which convention is the live one. Doing it.

The three standing "not fixing" items are unchanged, and I appreciate you re-checking their reasoning rather than carrying the earlier verdict forward on trust.

fredaan (approved):

PR #4997, opened 2026-09-04
fix(wiz/binding): expose calc-table cell naming, validate cross-table column binding

fredaan:
## Summary

Fixes two gaps in the wiz-agent calc (freehand) table cell-binding surface, found while testing the `stylebi-wiz` composer-chat plugin's cross-cell formula support (see `stylebi-wiz` case doc `docs/teams/2026-09-04-test-composer-plugin-calc-tables/case-freehand-cross-cell-formula/`).

**1. Cell naming (`CellBindingInfo.name`) was never written.** Calc tables already support a real named-cell reference mechanism: `TableLayoutInfo.setCellName(row, col, name)` stores a name on `TableCellBinding.cellName`, and at runtime `CalcTableScope.initRuntimeReferences` registers it as `"$" + name` so another formula cell can reference that cell's *computed* value (e.g. `$OrderTotal - $ReturnTotal`) — the exact mechanism the Composer's own Formula Editor generates when a user clicks a named cell. `CellBindingInfo` already has `setName`/`getName` and correctly round-trips to `TableCellBinding.cellName`, but `CalcTableService.toCellBindingInfo` never called `setName`, so there was no write path for it at all through this agent surface. `toCellBindingInfo` now sets the name when supplied, and `CalcCellVocabulary.describe` reports it back (alongside the existing auto-generated `runtimeName`, which is a different, non-referenceable name).

**2. A column cell's `field.column` was never validated against the calc table's actual bound source.** `CalcTableService.columnOf` only checked the column was non-blank and had a valid `type` — never that it actually exists on the table named via `set_table_source`. A column belonging to a *different* table was accepted silently and rendered a blank cell, with no error anywhere. `setCellBinding` now validates a `BIND_COLUMN` cell's column via `BindableFieldsService` + `BindableColumns.require` — the identical check `BindingAgentController.resolveSourceTable` already applies to chart/table/crosstab writes — so an invalid column now throws, naming the column and the assembly's actual current source.

## Changes

- `core/src/main/java/inetsoft/web/wiz/binding/CalcTableService.java`
  - `toCellBindingInfo`: sets `CellBindingInfo.name` when `binding.name` is supplied.
  - `setCellBinding`: injects `BindableFieldsService` and validates a `BIND_COLUMN` cell's `field.column` against the assembly's current bindable columns before writing, mirroring `BindingAgentController.resolveSourceTable`'s existing pattern (a listing failure is logged and skipped, not propagated — matching that method's own reasoning).
- `core/src/main/java/inetsoft/web/wiz/binding/CalcCellVocabulary.java`
  - `describe`: reports `name` alongside `runtimeName`.
  - `validate`: lists `name` among the accepted cell keys in its error message.

## Test plan

- [x] `CalcTableServiceTest` (extended): name round-trips through `toCellBindingInfo`/`describe`; a column on a different table than the assembly's current source is refused, naming the column and the source; all existing tests still pass with the new `BindableFieldsService` dependency wired into the harness.
- [x] `mvn -pl core test -Dtest=CalcTableServiceTest` — 43/43 pass.
- [x] `mvn -pl core test -Dtest=CalcCellVocabularyTest,BindingAgentControllerTest,BindableColumnsTest,BindableFieldsServiceTest,ChartBindingServiceTest,...` (surrounding `wiz.binding` package) — all pass, no collateral regressions from the constructor/vocabulary changes.

## Companion PR

Paired with `stylebi-wiz` PR (plugin-side `set_cell_binding`/`set_calc_cell_script` changes that expose this through the MCP tool surface) — will link once opened.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/CalcCellVocabulary.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/CalcCellVocabulary.java
@@ -104,7 +104,7 @@ public final class CalcCellVocabulary {
                "'" + rejected.getKey() + "' is not a calc-table cell key — it means something " +
                "else elsewhere in this plugin, which is exactly why it is refused here. Use '" +
                rejected.getValue() + "'. Cell keys are: content, grouping, expand, field, " +
-               "value, formula.");
+               "value, formula, name.");
          }
       }
 
@@ -160,6 +160,7 @@ public final class CalcCellVocabulary {
       out.put("mergeCells", info.getMergeCells());
       out.put("rowGroup", info.getRowGroup());
       out.put("colGroup", info.getColGroup());
+      out.put("name", info.getName());
       out.put("runtimeName", info.getRuntimeName());
       return out;
    }
--- a/core/src/main/java/inetsoft/web/wiz/binding/CalcTableService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/CalcTableService.java
@@ -48,8 +48,11 @@ import inetsoft.web.binding.event.ModifyTableLayoutEvent;
 import inetsoft.web.binding.event.SetCellBindingEvent;
 import inetsoft.web.binding.model.table.CellBindingInfo;
 import inetsoft.web.binding.model.table.TableCell;
+import inetsoft.web.wiz.binding.model.BindableTable;
 import inetsoft.web.wiz.binding.model.FieldRef;
 import inetsoft.web.wiz.viewsheet.ViewsheetSessionService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
@@ -75,11 +78,13 @@ import java.util.List;
 public class CalcTableService {
    @Autowired
    public CalcTableService(ViewsheetSessionService sessions, VSTableLayoutService layoutService,
-                           DataRefModelFactoryService refModelService)
+                           DataRefModelFactoryService refModelService,
+                           BindableFieldsService fieldsService)
    {
       this.sessions = sessions;
       this.layoutService = layoutService;
       this.refModelService = refModelService;
+      this.fieldsService = fieldsService;
    }
 
    /**
@@ -271,6 +276,11 @@ public class CalcTableService {
          requireInGrid(layoutOf(assembly), row, col);
 
          CellBindingInfo cellBindingInfo = toCellBindingInfo(binding);
+
+         if(cellBindingInfo.getType() == CellBinding.BIND_COLUMN) {
+            requireBindableColumn(runtimeId, user, assemblyName, cellBindingInfo.getValue());
+         }
+
          String namedGroup = cellBindingInfo.getType() == CellBinding.BIND_COLUMN
             ? namedGroupOf(binding.get("field")) : null;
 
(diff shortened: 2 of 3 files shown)


fredaan:
Companion plugin-side PR: inetsoft-technology/stylebi-wiz#2176

fredaan:
## Review round 1 — head commit `1195183d5`

**Scope:** Java changes only (`core/src/main/java/inetsoft/web/wiz/binding/CalcTableService.java`, `CalcCellVocabulary.java`, and `CalcTableServiceTest.java`). The companion plugin PR (stylebi-wiz#2176) is reviewed separately.

### What was checked

**1. `CellBindingInfo.setName` wiring (`toCellBindingInfo`, ~line 611).** Traced the full write path, not just the diff: `toCellBindingInfo` sets `info.setName(name)` only when `str(binding, "name")` is non-null/non-blank, matching the existing null-on-absent pattern used for `formula`/`value` elsewhere in the same method. Confirmed the downstream wiring is real, not just DTO plumbing: `VSTableLayoutService.setCellBinding` (untouched by this PR, line ~179-214) calls `setBinding(event.getBinding(), ...)`, which does `bind.setCellName(cinfo.getName())` at line 512 — this is the exact codepath `CalcTableService.setCellBinding` drives, so a supplied name now genuinely lands on `TableCellBinding.cellName`, which `CalcTableScope.initRuntimeReferences` already resolves as `"$"+name` at runtime (unchanged, pre-existing mechanism). For the "no name supplied" case: before this fix `cinfo.getName()` was always null, so `bind.setCellName(null)` was already being called on every write; after the fix it's still null when omitted, so unnamed cells are unaffected — no regression there. Re-binding a cell without repeating `name` will clear a previously-set name, but that matches this tool's existing full-replace semantics (every other field — `grouping`, `mergeCells`, etc. — already resets to default unless resupplied on every `set_cell_binding` call), not a new asymmetry.

**2. Column validation (`setCellBinding` ~266-299, `requireBindableColumn` ~314-327).** Confirmed `requireBindableColumn` runs only for `CellBinding.BIND_COLUMN`, before the runtime is mutated, using the identical `BindableFieldsService.list` + `BindableColumns.require` pair `BindingAgentController.resolveSourceTable` already uses for chart/table/crosstab writes, with the same catch-and-skip-on-listing-failure behavior (read failure vs. bad column kept as separate faults, matching the established pattern). Verified `cellBindingInfo.getValue()` is always non-null by this point for a `BIND_COLUMN` cell (`columnOf` throws before returning null/blank), so the `FieldRef` construction is safe. Verified `CalcTableVSAssemblyInfo extends TableDataVSAssemblyInfo extends DataVSAssemblyInfo`, so `BindableFieldsService.sourceNameOf`'s `DataVSAssembly` cast and `VSBindingTreeService.getBinding`'s `TableDataVSAssemblyInfo` branch (`getTableTreeModel`) both apply to calc tables exactly as they do to plain tables/crosstabs — the "current" table marking this check depends on is not calc-table-specific plumbing that might silently not apply.

Walked the boundary cases: dotted `"OtherTable.Column"` qualification (the tester's original follow-up attempt) is correctly rejected by the same mechanism, since it won't match any bare column name in `available`. A calc table with no `set_table_source` yet falls back to `BindableColumns.require`'s documented "no table marked current → match across every table" leniency — this is inherited, pre-existing behavior shared with chart/table writes, not a new gap introduced here.

**3. Constructor injection.** Grepped the full repo for `new CalcTableService(` — the only non-Spring construction site is the test harness (already updated in this diff). No other caller to break.

**4. `CalcCellVocabulary.describe`/`validate`.** `describe` now emits `name` alongside `runtimeName`, consistent with how every other cell property is surfaced. `validate`'s `REJECTED`-map error-message text was updated to list `name`; `name` was never in `REJECTED` and isn't checked against an allow-list elsewhere in `validate`, so this is a message-accuracy fix with no behavioral gate to get wrong.

**5. Regression tests.** Both new `CalcTableServiceTest` cases were checked against a mental "revert the fix" run: `readsCellNameBackAfterDescribing`/`namesACellSoAnotherFormulaCellCanReferenceItAsADollarName` would fail without `toCellBindingInfo`'s `setName` call and `describe`'s `name` line; `refusesAColumnNotOnTheAssemblysCurrentSource` would fail (no exception thrown) without `requireBindableColumn`. None of the three assert only the literal repro's exact input in a way that would still pass against a narrower fix — the column test in particular asserts the message names both the rejected column and the actual source table, not just "some error was thrown."

**Counter-example attempts that did not break the change:** tried building scenarios for (a) a formula (`BIND_FORMULA`) cell whose script text happens to look like a column name — not touched by the new check, correctly scoped to `BIND_COLUMN` only; (b) `copyCells`/`modifyLayout` as alternate write paths that might bypass `toCellBindingInfo`/validation — confirmed they don't accept a `field`/column input at all, so they can't introduce an unvalidated column binding; (c) case-sensitivity mismatches between the requested column and the listed one — `BindableColumns.require`'s `available` set is already case-insensitive, so no false rejection there. Could not construct a scenario where the fix's own claimed invariant fails.

### Findings

None at blocker or important severity.

**Verdict: approved.**

PR #4998, opened 2026-09-04
fix(wiz): set an explicit font family on board PPTX export text runs (bug-76110)

Zxm-happy:
## Bug

bug-76110: a board's PPTX export ("content not displayed well") showed a missing-glyph box
where an AI-generated insight bullet contained an emoji, while the same content in the PDF
export rendered correctly.

## Root cause

`PoiPptxDeckMerger` (the sole `PptxDeckMerger` implementation, wired via `PptxDeckMergerConfig`,
used by `WizViewsheetExportController.exportPptx()`) creates every `XSLFTextRun` for the title
slide, recap box, per-chart captions, and insights text (`styleBox()`, the bullet-marker run in
`appendBlock()`, and `appendSpans()`) with `{fontSize, fontColor, bold, italic}` but never called
`setFontFamily`. Per OOXML's inheritance model, a run with no explicit Latin-typeface element
resolves to the deck's theme font — confirmed "Calibri" in the reported PPTX's `theme1.xml`.

This is also an inconsistency with this codebase's own established PPTX-export convention:
`PPTValueHelper` (the generic, non-wiz PPTX export path, same POI/OOXML API) has always set an
explicit font family per run, with a comment citing a prior bug:

> bug #75992, write the font family, not the face name. getFontName() returns the face (e.g.
> "Roboto Bold"), which ppt cannot resolve as a typeface even when the family is installed, so it
> substitutes a font with different metrics and re-wraps the text.

## Fix

Set an explicit font family — `StyleFont.getDefaultFontFamily()` (resolves to
`SreeEnv.getProperty("default.font.family", "Roboto")`, "Roboto" by default) — on every run
created by `styleBox()`, the bullet-marker run in `appendBlock()`, and `appendSpans()`. This is
the same family the PDF export path resolves to for ordinary text
(`VSAssemblyInfo.getDefaultFont()` → `StyleFont(DEFAULT_FONT_FAMILY, ...)` →
`getDefaultFontFamily()`), so the two export paths are now consistent with each other, not just
each independently theme/property-driven.

Following the `PPTValueHelper`/bug #75992 precedent, this writes a plain resolved family name
(not a face/style-qualified name like "Roboto Bold") — bold/italic stay on the run's own
properties, as they already were.

## What this does NOT fix — please read before assuming "emoji now renders"

**This does not guarantee the emoji glyph itself renders instead of a missing-glyph box.** The
actual deployed Docker image (`community/docker/src/main/docker/Dockerfile`) installs only
`fonts-dejavu`, `fonts-liberation2`, `fonts-croscore`, `fonts-freefont-ttf`, and `fonts-roboto` —
none of which are confirmed emoji-capable. Whether any of these five families actually contains a
glyph for the reported pictograph could not be confirmed from source in this investigation (no
font-rendering tool or live target was available). This PR fixes a real, demonstrated
inconsistency (theme-inherited Calibri vs. an explicit, intentional family, matching this
codebase's own PPTX convention) — it is a plain-text/consistency fix, not a guaranteed emoji fix.
This was an explicit, informed product decision (font-only, not emoji-stripping or installing an
emoji font in the Docker image) made after the debugger and refuter both flagged the emoji
question as a product/UX call, not a fixer's call to make alone.

## Tests

Extended `PoiPptxDeckMergerTest` with
`everyRunCarriesAnExplicitFontFamilyInsteadOfThemeInheritance`: builds a `ChartSlide` with a
`**Growth accelerating.**` heading and a `- 📈 Premium pricing drives most of the revenue.`
bullet, merges, reopens the deck, and asserts `run.getFontFamily()` equals the resolved family
("Roboto" in this Spring-context-free unit test) for the title-slide run (`styleBox()`), the
bullet-marker run (`appendBlock()`), and the span run containing the emoji-adjacent text
(`appendSpans()`) — not just a non-null check, so an accidental future change shows as a diff.

```
mvn -pl utils/inetsoft-xml-formats -Dtest=PoiPptxDeckMergerTest test
Tests run: 15, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/utils/inetsoft-xml-formats/src/main/java/inetsoft/report/io/viewsheet/PoiPptxDeckMerger.java
+++ b/utils/inetsoft-xml-formats/src/main/java/inetsoft/report/io/viewsheet/PoiPptxDeckMerger.java
@@ -17,6 +17,7 @@
  */
 package inetsoft.report.io.viewsheet;
 
+import inetsoft.report.StyleFont;
 import inetsoft.web.wiz.service.MarkdownModel;
 import inetsoft.web.wiz.service.PptxDeckMerger;
 import org.apache.poi.xslf.usermodel.XMLSlideShow;
@@ -182,13 +183,20 @@ public class PoiPptxDeckMerger implements PptxDeckMerger {
       return candidate + "…";
    }
 
-   /** Apply a uniform bold/size/color to every run in every paragraph of a text box. */
+   /** Apply a uniform bold/size/color to every run in every paragraph of a text box.
+    *  Also sets an explicit font family: without one, a run has no Latin-typeface element at
+    *  all and resolves purely by OOXML theme inheritance (the deck's theme font), which
+    *  {@link #appendBlock} and {@link #appendSpans} below rely on this same explicit-family
+    *  approach for too. This matches this codebase's other PPTX-export convention
+    *  ({@code PPTValueHelper}, bug #75992) of always writing a resolved family name rather than
+    *  leaving typeface resolution to the reader. */
    private void styleBox(XSLFTextBox box, boolean bold, double fontSize, Color color) {
       for(XSLFTextParagraph paragraph : box.getTextParagraphs()) {
          for(XSLFTextRun run : paragraph.getTextRuns()) {
             run.setBold(bold);
             run.setFontSize(fontSize);
             run.setFontColor(color);
+            run.setFontFamily(StyleFont.getDefaultFontFamily());
          }
       }
    }
@@ -299,6 +307,7 @@ public class PoiPptxDeckMerger implements PptxDeckMerger {
          bullet.setFontSize(bodySize);
          bullet.setFontColor(ACCENT);
          bullet.setBold(true);
+         bullet.setFontFamily(StyleFont.getDefaultFontFamily());
          appendSpans(p, block.spans(), bodySize, BODY_COLOR, false);
          break;
       default:
@@ -320,6 +329,7 @@ public class PoiPptxDeckMerger implements PptxDeckMerger {
          run.setFontColor(color);
          run.setBold(forceBold || span.bold());
          run.setItalic(span.italic());
+         run.setFontFamily(StyleFont.getDefaultFontFamily());
       }
    }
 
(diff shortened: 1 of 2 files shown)


Zxm-happy:
**Reviewer pass (approve API blocked by self-review restriction, posting as comment; merged via squash after this review)**

Findings — no blocking issues:
- Diff confined to exactly 2 files, matching `03-fix.md`'s claim: `PoiPptxDeckMerger.java` (+11/-1) and `PoiPptxDeckMergerTest.java` (+66).
- `import inetsoft.report.StyleFont;` resolves unambiguously — only one `StyleFont` class in the repo (`core/src/main/java/inetsoft/report/StyleFont.java`), no naming collision.
- `StyleFont.getDefaultFontFamily()` (StyleFont.java:685-698) is deterministic ("Roboto") outside a Spring context — read the `SreeEnv.isInitialized()` guard directly, confirmed by the module's own test run never bootstrapping Spring.
- Fixer's deviation from routing through `VSFontHelper.getExportFontFamily()` is sound: that helper (VSFontHelper.java:50-73) only corrects an already-constructed `java.awt.Font`'s family after AWT's own logical-family collapse; the plain family string `PoiPptxDeckMerger` now writes is never wrapped in a `Font` object first, so there is no AWT substitution step for the helper to correct. (Tester's 04-verify.md flagged that `03-fix.md`'s supporting phrase "never constructs an AWT Font object at all" is imprecise — the class does build 3 unrelated `Font(SANS_SERIF, ...)` objects for text-measurement/pagination, decoupled from the family string this PR writes — a documentation wording nitpick, not a code defect.)
- New test `everyRunCarriesAnExplicitFontFamilyInsteadOfThemeInheritance` genuinely exercises all three changed call sites (`styleBox()`, the bullet-marker run in `appendBlock()`, `appendSpans()`) with literal-value assertions (`assertEquals("Roboto", ...)`), not weak non-null checks. Independently reran: 15/15 pass.
- No regression risk to existing wrap/truncation tests (e.g. `aLongCaptionIsFittedToItsBandInsteadOfRunningOverTheChart`): confirmed that pagination/caption-fit math (`textHeightPt()`/`estimateBlockHeightPt()`/`chunkInsightsText()`) measures with a hardcoded `Font.SANS_SERIF` object, entirely decoupled from `StyleFont.getDefaultFontFamily()` — pre-existing behavior, untouched by this PR.
- PR body honesty confirmed against the live GitHub text: explicitly states this does not guarantee the emoji glyph renders, cites the Dockerfile finding (no confirmed emoji-capable font installed) as a known residual risk — does not overclaim.
- CI: all 5 checks green (build passed on rerun, Semgrep x2, claude-review, license/cla); mergeable_state clean.

Merged via squash.

PR #5002, opened 2026-09-04
Add create_worksheet agent endpoint

fredaan:
## Summary
- Add `SheetOpenService.createWorksheet(fromSessionToken, user)`: mints a brand-new, blank worksheet runtime by reusing an already-paired session's live browser socket, symmetric to the existing `createViewsheet`.
- Expose it via `POST /api/wiz/v1/agent/worksheet/create` on `WorksheetAgentController`, returning the same `JoinResponse` shape every other agent endpoint uses.
- This closes the gap where the wiz agent plugin could mint a new *viewsheet* (`create_viewsheet`) or attach an existing worksheet as a viewsheet's base (`attach_base_worksheet`), but had no way to build a brand-new, multi-table/joined worksheet from scratch.

## Design
Full design/rationale: see `docs/superpowers/specs/2026-09-04-create-worksheet-tool-design.md` and `docs/superpowers/plans/2026-09-04-create-worksheet-community-backend.md` in the `stylebi-wiz` repo (not part of this PR).

`createWorksheet` mirrors `createViewsheet`'s three guards exactly (valid acting session, not pane-scoped, has a live socket) plus the same permission check pattern, then mints via `openTemporaryWorksheet(user, null)` — the same call `OpenWorksheetController`'s own "New Worksheet" button and `WorksheetTableService.openProbeWorksheet` already use for a pure, asset-less runtime.

The plugin-side `create_worksheet` MCP tool (in `stylebi-wiz`) is a separate, later PR that will call this endpoint.

## Test plan
- [x] `./mvnw test -pl core -Dtest=SheetOpenServiceTest` — 38 tests, 0 failures (11 new)
- [x] `./mvnw test -pl core -Dtest=WorksheetAgentControllerTest` — 130 tests, 0 failures (3 new)
- [x] `./mvnw test -pl core -Dtest=WorksheetScriptControllerTest` — 7 tests, 0 failures (constructor call-site fix)
- [x] Full `./mvnw test -pl core` — 8736 tests, 0 failures, 0 errors, BUILD SUCCESS
- [x] Independent code review (no findings)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -232,3 +232,7 @@ terraform.rc
 # is a review artifact, never applied automatically. Same reasoning as docs/properties/properties.json
 # in the stylebi-enterprise superproject: untracked rather than committed-and-stale.
 core/tools/admin-property-catalog/catalog-promotion-proposals.json
+
+# Local git worktrees created for isolated feature branches (using-git-worktrees convention) --
+# lives at the repo root, sibling to the checkouts it holds; never meant to be tracked.
+.worktrees/
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java
@@ -323,6 +323,96 @@ public class SheetOpenService {
       return vsSession;
    }
 
+   /**
+    * {@code create_worksheet}. Mints a brand-new, BLANK worksheet runtime, and pairs the caller
+    * to it directly by reusing the ACTING session's already-live browser socket -- the exact
+    * mechanism {@link #createViewsheet} already uses in the reverse direction (worksheet/
+    * viewsheet -> new viewsheet), applied here to worksheet/viewsheet -> new worksheet.
+    *
+    * <p>Unlike {@link #openBaseWorksheet}, which always follows a viewsheet down to ITS OWN base
+    * (never an arbitrary target), the acting session here may be EITHER sheet type, exactly like
+    * {@link #createViewsheet} -- resolved through the generic, type-agnostic
+    * {@link SheetSessionService#resolve}, not {@link ViewsheetSessionService}.
+    *
+    * <p>Unlike {@link #createViewsheet}, there is no data source to resolve: the new worksheet is
+    * always blank. Callers design it with the add_table/add_join/add_calc_field family of tools,
+    * then save_worksheet to persist it. If the acting session is a viewsheet with no source of
+    * its own, attach_base_worksheet is the tool that connects the saved worksheet back to it.
+    *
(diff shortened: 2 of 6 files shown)


fredaan:
Second (adversarial) review pass over both commits — no defects found, nothing pushed.

Checked against the specific risk areas from the first review round:

- **`openTemporaryWorksheet(user, null)` semantics** — confirmed this is exactly the call used by the two other "mint a blank worksheet runtime, no asset" call sites in the codebase: `OpenWorksheetController.newWorksheet` (the Composer's own "New Worksheet" toolbar action, `engine.openTemporaryWorksheet(principal, null)`) and `WorksheetTableService.openProbeWorksheet` (`viewsheetService.openTemporaryWorksheet(user, null)`). Same signature, same null asset entry, same semantics.
- **Guard order** — `createWorksheet`'s order (resolve acting session → refuse pane-scope → refuse missing socket → check `WORKSHEET` permission → mint) is byte-for-byte the same order `createViewsheet` uses (minus the data-source-resolution step that only `createViewsheet` needs), and the pane-scope check is first, before the runtime is touched or any grant is minted, matching the documented CRITICAL-finding rule from `openBaseWorksheet`. `openBaseWorksheet` itself checks its socket guard *last* rather than early, but that's a pre-existing asymmetry between it and `createViewsheet`/`createWorksheet` (different method, different flow — it has to resolve the base-worksheet entry before it can even check permission), not something this PR introduced or should have made consistent with.
- **`WorksheetAgentController.createWorksheet` exception handling** — `catch(IllegalArgumentException e) { throw new PairingException(e.getMessage(), e); }` matches `ViewsheetAssemblyAgentController.createViewsheet`'s handling verbatim.
- **Test assertions** — `SheetOpenServiceTest.createWorksheetRefusesAPaneScopedActingSessionRatherThanMintingAWholeSheetOne` asserts the message content AND carries all four `never()` verifications (`openTemporaryWorksheet`, `sheetSessions.open`, `sendAgentActive`, `sendToComposer`), not just that an exception was thrown. `WorksheetAgentControllerTest`'s three `createWorksheet` tests cover the happy-path delegation, the exception-wrap, and the feature-disabled path (with `verifyNoInteractions(openService)`).
- **Other `new WorksheetAgentController(` call sites** — re-ran the grep myself; all 6 hits (`WorksheetAgentControllerTest` ×5, `WorksheetScriptControllerTest` ×1) already pass the new `SheetOpenService` constructor argument. Nothing missed.
- **Full suite**: `./mvnw test -pl core` — 8736 tests, 0 failures, 0 errors (the known-flaky `OrganizationContextHolderThreadingTest`/`ConditionTest$RegressionTests` cases did not even flake this run).

No changes made — the implementation and tests hold up under a second look.

PR #5013, opened 2026-09-04
Bug #75686, animate pie slice borders with their slices

ngovus-inetsoft:
## Problem

Open viewsheet `pie` in the portal. When the pie chart's binding contains a **Group** dimension, the pie's full wedge outlines are already painted in the first frame — the group structure is visible before the animation starts. Only the slice *fills* sweep in.

## Root cause

Three things line up, which is why it only reproduces when a group exists:

1. `GraphGenerator.java:4894-4899` — when a chart has group fields and at least one group *dimension*, the `IntervalElement` gets `setLineFrame(new StaticLineFrame(new GLine(1)))`: *"add border around each piece otherwise it's not distinguishable."* This gate is exactly the "when pie chart exist group" condition.

2. `BarVO.paintBar()` fills the wedge and then strokes the **same `Shape`**, so Batik emits two sibling `<path>` elements with an identical `d` inside one `<g class="inetsoft-bar">` — the fill carrying `stroke="none"`, the border carrying `fill="none"`:

```
<g class="inetsoft-bar" data-row=".." data-col="..">
  <g fill="red" stroke="red">
    <path d="M100 50 A50 50 0 0 0 25 6.7 L50 50 Z" stroke="none"/>
    <path fill="none" d="M100 50 A50 50 0 0 0 25 6.7 L50 50 Z" stroke="black"/>
  </g>
</g>
```

3. `SVGAnimationDOMInjector.collectSlicesAndText()` collects only paths whose own `stroke` is `"none"`. The border path was never collected, never given a from-state and never animated, so it rendered at full geometry and full opacity from frame 0.

Pie is the only chart type affected — every other `inject*Animation` styles the whole annotation `<g>`, which covers the border for free. Pie has to style individual paths because it rewrites `d` for the arc sweep.

## Fix

`SVGAnimationDOMInjector` only. No change to `GraphGenerator` or `BarVO`: the border itself is intended and correct at rest, only its entrance was wrong.

`buildSliceCompanions()` pairs each fill path with the border path(s) sharing its `d` within the enclosing annotation group, and all six per-slice styling outcomes now route through `setSliceD` / `styleSlice` / `styleSliceSweep`. `slices` still holds only fill paths, so panel partitioning, angle grouping, durations and stagger see identical input — the existing animation's timing is untouched.

One subtlety worth calling out: the sweep's from-state is a *degenerate* arc, which SVG treats as omitted, leaving `moveto -> lineto(hub) -> closepath`. That encloses no area when filled (which is why the fill needs no opacity handling and `forwards` suffices), but draws a **radial hairline when stroked**. Copying just the `d` would have traded "full outline at load" for "an asterisk at load", so each border also gets an `inetsoft-pie-fade 0.001s steps(1,start) <delay>s both` gate — the same idiom already used for 3D depth faces.

A border is claimed by the first slice that matches it. With the apply-effect (shine) option, `BarVO.applyEffect` fills the same wedge twice more with gradients and those paths are collected as slices too, so without the claim check the one border would accumulate three stacked `animation` declarations of which only the last took effect.

3D pie never reaches this code: `VGraphPair.hasPieVO` skips `Pie3DVO`, so no animation hint is ever set for it.

## Verification

- **Batik DOM probe** (throwaway, not committed) confirmed the assumed structure exactly: identical `d`, `stroke="none"` on the fill, `fill="none"` on the border, both under the `inetsoft-bar` group.
- **End-to-end through real Batik + the injector** (throwaway, not committed): a 3-wedge bordered pie produced, per slice, a fill on `inetsoft-pie-sweep-N` and its border on the *same* rule plus the opacity gate at the *same* delay — durations proportional to sweep angle (0.250/0.188/0.313s for 120°/90°/150°), 3 keyframe rules for 3 slices, no duplication.
- **102/102 tests pass** in `inetsoft-xml-formats` (86 in `SVGAnimationDOMInjectorTest`, 13 new).
- **Negative check**: stashing the production change makes 7 of the new tests fail with exactly the expected messages (empty border style, `null` keyframe name, wrong from-state). The remaining new tests are no-regression guards and correctly pass either way — notably `groupedPie_fillStyleIdenticalToUngroupedPie`, which asserts the fill's `d`, `style` and the whole generated CSS are byte-identical with and without borders.
- **Visually confirmed** in the portal on the `pie` viewsheet: the plot starts blank and each wedge's outline grows with its fill; no outlines and no radial hairlines before the sweep. Regression-checked ungrouped pie, donut, faceted pie, and composer/binding (`:noanim`) rendering.

## Notes

Community-only change; no enterprise counterpart.

Two unrelated latent defects were found while tracing this and deliberately left out of scope — happy to file them separately:

- `@keyframes` names are not rewritten by `ChartInlineSvgDirective.uniquifyIds` (`chart-inline-svg.directive.ts:1663`). Inlined SVG `<style>` is document-global, so two pie charts on one dashboard both define `inetsoft-pie-sweep-0` and the last parsed wins.
- `collectSlicesAndText` lacks the `inetsoft-`-parent guard that `collectTextGroups` has, so a Batik style-wrapper `<g text-rendering="geometricPrecision">` inside an annotation group could be mistaken for a text label group. Not firing for pie today, but the asymmetry is a latent trap.

### DIFF
--- a/utils/inetsoft-xml-formats/src/main/java/inetsoft/util/graphics/animation/SVGAnimationDOMInjector.java
+++ b/utils/inetsoft-xml-formats/src/main/java/inetsoft/util/graphics/animation/SVGAnimationDOMInjector.java
@@ -532,6 +532,12 @@ public class SVGAnimationDOMInjector {
          return;
       }
 
+      // Pair each fill path with the wedge-border path(s) Batik emits alongside it when the pie
+      // has a slice border, so the outline sweeps in with its slice instead of showing the pie's
+      // whole group structure from the first paint.  Built here, before any "d" is rewritten,
+      // because "d" is the pairing key.
+      Map<Element, List<Element>> companions = buildSliceCompanions(slices);
+
       // Detect and reclassify donut center-hole overlays; a faceted donut has one per panel.
       List<double[]> holes = preprocessDonutHoles(svgRoot);
 
@@ -560,7 +566,7 @@ public class SVGAnimationDOMInjector {
          // that, so fall back to the opacity fade.  All panels begin at t=0 (facets in parallel).
          boolean tiled = isPanelTiled(panel, viewport);
          double endTime = animatePiePanel(panel.slices, panel.center, hole, tiled,
-                                          cssKeyframes, kfCounter);
+                                          cssKeyframes, kfCounter, companions);
          maxEndTime = Math.max(maxEndTime, endTime);
       }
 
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft (changes requested):
Claude's automated review did not complete — it errored out after 0s and posted only a placeholder comment ("I'll analyze this and get back to you") with no actual review content. There is also no human approval on this PR to satisfy the fallback path. Please re-trigger the Claude review (or get a human review) before this can be merged.

jshobe-inetsoft (approved):
Claude's automated review completed after the previous placeholder/error state and found no correctness issues, with strong test coverage. This supersedes the earlier CHANGES_REQUESTED review, which was based on the incomplete placeholder comment (Claude's review errored out at that time). Approving based on the completed review's analysis.

PR #5015, opened 2026-09-04
fix(admin): exclude task narrative from plan hash (Bug #76472, Properties)

fredaan:
## Summary
- Bug #76472, Properties area. `AdminChangePlanService.hash()` seeded its SHA-256 canonical string with the free-text `task` narrative, but `task` is never compared or parsed after `resolve()` returns - it only flows through `AdminChangesetApplyService` into `AdminChangeRecord.setTaskDescription`, a write-only audit label.
- Result: two equally valid narratives describing the identical change list (e.g. "cap query rows at 500" vs "limit max rows to 500") produced different `planHash()` values, so an operator paraphrasing the task between preview and apply got a spurious 409 `PlanHashMismatchException` even though nothing the drift gate actually protects had changed.
- Fix: drop `task` from `hash()`'s signature and canonical-string seed; leave `AdminChangesetApplyService`'s audit-label flow untouched (that's the legitimate use of `task`).

## Test plan
- [x] `AdminChangePlanServiceTest.hashChangesWithTheProposedValueAndTheTask` (which asserted the bug as spec: different `task`, same `changes` -> different hash) replaced with `hashChangesWithTheProposedValue` (keeps the proposed-value-changes-the-hash assertion) and a new `hashIsUnaffectedByDifferentTaskStrings` regression test (different `task`, same `changes` -> equal hash). Verified red on pre-fix code, green after the fix.
- [x] `./mvnw -pl core test -Dtest='AdminChangePlanServiceTest' -o` — Tests run: 29, Failures: 0
- [x] `./mvnw -pl core test -Dtest='AdminChangesetApplyServiceTest' -o` — Tests run: 26, Failures: 0 (no collateral breakage)

This is 1 of 5 independent PRs in the #76472 batch (Properties, Cluster, Licensing, Presentation, Providers) - this PR covers Properties only.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/AdminChangePlanService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/AdminChangePlanService.java
@@ -167,16 +167,23 @@ public class AdminChangePlanService {
          .anyMatch(c -> AdminChangeRecord.RISK_HIGH.equals(c.risk()));
 
       return new ResolvedPlan(req.getTask().trim(), Collections.unmodifiableList(changes),
-                              backup, signoff, hash(req.getTask().trim(), changes));
+                              backup, signoff, hash(changes));
    }
 
    /**
     * SHA-256 over the canonical plan. Field order and the record separators are part of the
     * contract: changing them invalidates every outstanding preview, which is safe (an apply is
     * refused with 409) but forces operators to re-review.
+    *
+    * <p>{@code task} is deliberately excluded: it is a free-text narrative that is never compared
+    * or parsed, only carried through to {@code AdminChangesetApplyService}'s {@code writeAudit}
+    * call as {@link AdminChangeRecord#setTaskDescription}, a write-only audit label. Including it
+    * here would let two equivalent narratives describing the identical change list hash
+    * differently, forcing a spurious re-review (a false {@code PlanHashMismatchException}) even
+    * though nothing the gate actually protects has changed.
     */
-   private static String hash(String task, List<PlanChange> changes) {
-      StringBuilder canonical = new StringBuilder(task).append('\n');
+   private static String hash(List<PlanChange> changes) {
+      StringBuilder canonical = new StringBuilder();
 
       for(PlanChange change : changes) {
          canonical.append(change.property()).append(SEP)
--- a/core/src/test/java/inetsoft/web/admin/ai/AdminChangePlanServiceTest.java
+++ b/core/src/test/java/inetsoft/web/admin/ai/AdminChangePlanServiceTest.java
@@ -148,12 +148,19 @@ class AdminChangePlanServiceTest {
    }
 
    @Test
-   void hashChangesWithTheProposedValueAndTheTask() {
+   void hashChangesWithTheProposedValue() {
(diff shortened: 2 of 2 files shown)


fredaan:
## Round 1 review — bug-76472-properties

Head commit reviewed: 58aa94113bda9b8d10cc522b0c5bd075100f85b8

### What was checked
- Full diff (`gh pr diff 5015`): `AdminChangePlanService.java` (`hash()` signature/canonical-string change + new Javadoc) and `AdminChangePlanServiceTest.java` (test split/replacement). No other files touched.
- Read `01-diagnosis.md` and `02-refute.md` for this sub-bug and checked the diff against the confirmed cause, not just the reported symptom.
- Confirmed the fix corrects the general rule, not a repro-literal special case: `hash(List<PlanChange>)` no longer accepts `task` at all — it's not an `if` that excludes one string, it's removed from the canonical string for every call.
- Verified `AdminChangesetApplyService.java` has no direct call to the private `hash()` — its only interaction with hashing is `plan.planHash().equals(req.getPlanHash())` (line 91), a `String` comparison against the record accessor. `grep -n "hash("` on `AdminChangePlanService.java` shows exactly one call site (inside `resolve()`), confirming no other caller was left un-migrated.
- Confirmed `req.getTask().trim()` still flows into `ResolvedPlan`'s `task` field unchanged — only the hash input changed, not what's carried for the audit-label use.
- Read the full `hash()` javadoc addition — it explains *why* `task` is excluded (write-only audit label via `AdminChangeRecord#setTaskDescription`, never compared/parsed), matching the safety argument in the diagnosis.
- Checked the test change is a genuine correction, not an added test alongside a still-wrong one: `hashChangesWithTheProposedValueAndTheTask` (which asserted the bug as spec — different task ⇒ different hash) was replaced by `hashChangesWithTheProposedValue` (keeps the legitimate proposed-value-changes-hash assertion) plus a new `hashIsUnaffectedByDifferentTaskStrings` (different task, same changes ⇒ **equal** hash via `assertEquals`). The old buggy assertion is gone, not just superseded.
- Independently re-ran both affected test classes rather than trusting the PR description: `AdminChangePlanServiceTest` (29 tests) and `AdminChangesetApplyServiceTest` (26 tests) — both green, matches PR's claimed counts exactly.
- Alternate scenarios walked, trying to build a counter-example: empty `changes` list (hash still well-defined, just an empty canonical string — no NPE risk since `task` is no longer concatenated first); multiple changes with identical property/value but different risk/scope (still correctly discriminated, untouched by this diff); the sibling `hashChangesWhenTheCurrentValueDrifts` test (untouched, still exercises the drift gate on `currentValue`, unaffected by dropping `task`). Could not construct a scenario where this change fails.
- Confirmed commit message and PR title/description reference Bug #76472, Properties specifically (not the whole 5-PR batch), and correctly scope the "1 of 5" note as informational only.
- Confirmed no unrelated changes bundled — `gh pr diff` shows only the two files above.

### Findings
No findings.

### Verdict
Approve (recorded as a comment, not a formal review — see note below).

Note: I'm reviewing under the same `gh` identity that authored this PR (this batch's PRs share one account). GitHub rejected a formal approval as expected: "Review Can not approve your own pull request". This comment is the audit-trail record of a clean round-1 review; a formal `gh pr review --approve` will need to come from a different identity/human before merge.

PR #5016, opened 2026-09-04
fix(#76472-licensing): drop task from LicenseChangePlanService.hash() seed

fredaan:
## Summary

- `LicenseChangePlanService.hash()` folded the caller-supplied `task` narrative into the SHA-256 canonical string used for `ResolvedPlan.planHash()`. Since `task` is free-text (never an opaque token the caller is required to echo verbatim), an LLM agent naturally paraphrasing `task` between `preview` and `apply` produces a different hash even though the actual plan (`changes` + live `installedCount`) is identical, causing `LicenseChangesetApplyService.apply()`'s hash-comparison gate to throw a false-positive `PlanHashMismatchException` (HTTP 409).
- `task`'s only downstream use after `resolve()` returns is an audit label (`writeAudit`'s `record.setTaskDescription(task)`) -- never compared, parsed, or used to gate a mutation -- so removing it from the hash seed is safe.
- `installedCount` is left untouched: it is a live read of `licenseManager.getInstalledLicenses()` and legitimately should keep perturbing the hash (a concurrent add/remove of a key the plan never mentions must still be caught).

This is 1 of 5 independent fixes for Bug #76472 (each sibling area -- Properties, Cluster, Licensing, Presentation, Providers -- gets its own PR).

## Test plan

- [x] Added `LicenseChangePlanServiceTest.hashIsUnaffectedByDifferentTaskStrings` -- confirmed it fails on pre-fix code (two different SHA-256 digests for two `task` phrasings over identical `changes`/`installedCount`) and passes after the fix.
- [x] `./mvnw -pl core test -Dtest='LicenseChangePlanServiceTest,LicenseChangesetApplyServiceTest' -o` -- `Tests run: 32, Failures: 0, Errors: 0, Skipped: 0`, `BUILD SUCCESS`.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/licensing/LicenseChangePlanService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/licensing/LicenseChangePlanService.java
@@ -132,7 +132,7 @@ public class LicenseChangePlanService {
 
       String task = req.getTask().trim();
       return new ResolvedPlan(task, Collections.unmodifiableList(changes), true, true,
-                              hash(task, changes, installed.size()));
+                              hash(changes, installed.size()));
    }
 
    // ---------------------------------------------------------------- add
@@ -252,10 +252,18 @@ public class LicenseChangePlanService {
     * section 5) -- the de-licensing warning depends on the WHOLE deployment's installed count, not
     * only the keys this plan touches, so a concurrent add/remove of a key this plan never mentions
     * must still perturb the hash.
+    *
+    * <p>Deliberately does NOT fold {@code task} into the canonical string: {@code task} is
+    * caller-supplied free-text narrative, not an opaque token the caller is required to echo
+    * verbatim between {@code preview} and {@code apply}. Its only downstream use after {@code
+    * resolve()} returns is {@code LicenseChangesetApplyService.writeAudit}'s
+    * {@code record.setTaskDescription(task)} -- an audit label, never compared, parsed, or used to
+    * gate a mutation -- so a caller paraphrasing {@code task} between {@code preview} and {@code
+    * apply} must not change the hash and trip a false-positive {@code PlanHashMismatchException}.
     */
-   private static String hash(String task, List<PlanChange> changes, int installedCount) {
-      StringBuilder canonical = new StringBuilder(task).append('\n')
-         .append("installedCount:").append(installedCount).append(SEP);
+   private static String hash(List<PlanChange> changes, int installedCount) {
+      StringBuilder canonical = new StringBuilder("installedCount:")
+         .append(installedCount).append(SEP);
 
       for(PlanChange change : changes) {
          canonical.append(change.property()).append(SEP)
--- a/core/src/test/java/inetsoft/web/admin/ai/licensing/LicenseChangePlanServiceTest.java
+++ b/core/src/test/java/inetsoft/web/admin/ai/licensing/LicenseChangePlanServiceTest.java
@@ -274,4 +274,18 @@ class LicenseChangePlanServiceTest {
 
       assertNotEquals(before.planHash(), after.planHash());
    }
+
+   /** stylebi#76472 -- {@code task} is caller-supplied free-text narrative that an LLM caller may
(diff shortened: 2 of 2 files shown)


fredaan:
## Review round 1 — bug-76472-licensing (PR #5016)

**Head commit reviewed:** d7944fc984c2006a65c8a66e2fc1bf27428f08dc

### What was checked
- Diagnosis (`01-diagnosis.md`) and refutation (`02-refute.md`) against the actual current diff, not just their claims.
- `hash()`'s new 2-arg signature (`hash(List<PlanChange> changes, int installedCount)`) vs. the single call site in `resolve()` (line 135): confirmed a pure parameter deletion, `installedCount`'s position, type, and computation (`installed.size()`) are byte-for-byte untouched, and the `installedCount:` seed prefix is preserved exactly (just now first instead of second in the canonical string).
- Grepped the whole file for other `hash(` references — only one declaration, one call site. No other production caller of this private method exists.
- Grepped `LicenseChangesetApplyService.java` for any direct `hash(` call — none; it only compares `plan.planHash()` (line 81) against the client-echoed `req.getPlanHash()`, fed by `resolve()`. Confirmed against the actual current file content, not just the diagnosis's claim.
- Confirmed `task` is not keyed on as a literal and no special-case `if` excludes any reported input — the fix removes `task` from the hash seed unconditionally for all callers, correcting the general rule rather than special-casing the repro's paraphrase.
- Checked the sibling/else paths this fix could have missed: `resolve()` is the only place `hash()` is invoked; there is no separate "apply-time" recomputation path with its own hash logic that could still be task-sensitive. `plan.task()` pass-through into `writeAudit`/`applyOne`/`rollback` is left completely untouched, as required.
- Javadoc added on `hash()` explains the WHY (task is free-text an LLM paraphrases, not an opaque echo token; its only downstream use is an audit label) — matches CLAUDE.md's comment-only-when-WHY-is-non-obvious bar.
- New test `hashIsUnaffectedByDifferentTaskStrings`: uses two genuinely different task strings ("remove the old K1 key" vs "please remove license K1"), identical `changes` (`remove("K1")`) and identical installed-license state (so `installedCount` is provably identical), asserts `planHash()` equality. This is not keyed on the diagnosis's literal repro wording being special — it's a general task-invariance assertion. Well-named, matches the existing file's test style/helpers.
- Ran `./mvnw -pl core test -Dtest='LicenseChangePlanServiceTest,LicenseChangesetApplyServiceTest' -o`: `Tests run: 32, Failures: 0, Errors: 0, Skipped: 0`, `BUILD SUCCESS` — matches the PR description's claim.
- Scope: `gh pr diff 5016` touches only `LicenseChangePlanService.java` (production) and `LicenseChangePlanServiceTest.java` (test) — no unrelated files bundled in.
- Commit message and PR title/description reference "#76472-licensing"/"Bug #76472"/"Licensing" specifically, and explicitly scope themselves to "1 of 5 independent fixes" — not the whole batch.

### Alternate scenarios tried, could not break the fix
- Reordering `changes` between preview/apply (still order-sensitive after the fix, same as before) — out of scope for this bug, not claimed as fixed, not regressed either.
- Concurrent installed-license-count drift between preview/apply — still perturbs the hash as intended (`hashChangesWhenAnUnrelatedKeyIsInstalledConcurrently` still passes, confirmed in the 32-test run).
- Empty/whitespace-only task — `resolve()` already rejects this earlier via `req.getTask().trim().isEmpty()` validation (line ~80), unrelated to and unaffected by this change.
- Other callers of `LicenseChangePlanService`/`LicenseChangesetApplyService` outside `AdminLicensingController` — none found; `ResolvedPlan`'s shape is unchanged so the other 4 sibling areas (Properties/Cluster/Presentation/Providers) are structurally unaffected by this PR.

### Findings
No findings — round 1 is clean.

### Verdict
Approved.

**Note on merge process:** this session's `gh` identity (`fredaan`) is also this PR's author, so GitHub is expected to reject a formal self-approval — noting that here rather than fighting it; a human or a different identity must apply the actual approve/merge.

PR #5017, opened 2026-09-04
fix(#76472-cluster): exclude task from ClusterChangePlanService plan hash

fredaan:
## Summary
- Bug #76472 (Cluster area, 1 of 5 independent sibling PRs for this batch — the other four areas, Properties/Licensing/Presentation/Providers, each get their own PR).
- `ClusterChangePlanService.hash(String task, List<PlanChange> changes)` folded the free-text, audit-only `task` narrative into the SHA-256 canonical string used for `planHash()`. Paraphrasing `task` between `preview` and `apply` — with `changes` byte-identical — produced a different hash, causing `ClusterChangesetApplyService.apply()`'s plan-hash equality gate to falsely refuse with `PlanHashMismatchException` (HTTP 409), even though nothing about the actual plan changed.
- Traced every downstream use of `task`: it only ever reaches `AdminChangeRecord.setTaskDescription()` via `writeAudit()` — never compared, parsed, tokenized, or used to gate a mutation — so it is safe to drop from the plan-identity hash.

## Changes
- `core/src/main/java/inetsoft/web/admin/ai/cluster/ClusterChangePlanService.java`: dropped `task` from `hash()`'s signature and canonical-string seed; updated the `resolve()` call site to match; added a Javadoc paragraph explaining why `task` is excluded (audit-only, per `ClusterChangesetApplyService.writeAudit()`).
- `core/src/test/java/inetsoft/web/admin/ai/cluster/ClusterChangePlanServiceTest.java`: added `hashIsUnaffectedByDifferentTaskStrings` — two `resolve()` calls with different `task` text and identical `changes`, asserting equal `planHash()`. Verified red (fails with a hash mismatch) on pre-fix code and green after the fix, via a scoped `git apply -R`/`git apply` round trip on just the service file.
- `ClusterChangesetApplyService.java` is untouched — it has no direct `hash()` call; it re-resolves the whole plan via `resolve()`, so it picks up the new hash signature automatically. Verified via `grep` — no other production call site for `hash()` exists.

## Test plan
- [x] `./mvnw -pl core test -Dtest='ClusterChangePlanServiceTest' -o` — `Tests run: 20, Failures: 0, Errors: 0` (was `Tests run: 20, Failures: 1` on pre-fix code, confirming red→green).
- [x] `./mvnw -pl core test -Dtest='ClusterChangesetApplyServiceTest' -o` — `Tests run: 10, Failures: 0, Errors: 0` (no collateral breakage).

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/cluster/ClusterChangePlanService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/cluster/ClusterChangePlanService.java
@@ -119,7 +119,7 @@ public class ClusterChangePlanService {
 
       String task = req.getTask().trim();
       return new ResolvedPlan(task, Collections.unmodifiableList(changes), false, true,
-                              hash(task, changes));
+                              hash(changes));
    }
 
    private PlanChange resolveOne(String server, String verb) {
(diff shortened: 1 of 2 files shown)


fredaan:
Round 1 review — approved in substance (self-review restriction blocks a formal `gh pr review --approve` from this account; a human with a different GitHub identity must formally approve/merge).

Checked against the confirmed diagnosis (docs/teams/2026-09-05-bugs-76472-planhash-siblings/bug-76472-cluster/01-diagnosis.md through 03-fix.md), using `gh pr diff 5017` as primary source of truth rather than local git state:

- `hash()` now unconditionally drops `task` from its signature/seed (private static method, exactly one production call site — verified via grep, no other call site exists anywhere in the cluster package). This is a general-rule fix, not a repro-literal special case.
- `ClusterChangesetApplyService.java` needed no change — confirmed via `grep -n "hash("` returning zero matches in that file; it only reaches a hash via `resolve()`, which now naturally yields task-independent hashes. Its plan-hash equality gate is untouched, and real plan drift (live `currentValue` changes) is still correctly refused, per the existing `hashChangesWhenCurrentStatusReadChangesBetweenTwoResolveCalls` test which still passes unchanged.
- The new regression test `hashIsUnaffectedByDifferentTaskStrings` genuinely proves task-invariance: two different task strings ("pause server s1 for maintenance" vs. "pausing s1 to do maintenance work"), identical `changes`, asserts equal `planHash()` — not a narrow repro-literal check.
- Re-ran the tests myself (not just trusting the fix report): `./mvnw -pl core test -Dtest='ClusterChangePlanServiceTest,ClusterChangesetApplyServiceTest' -o` → Tests run: 30, Failures: 0, Errors: 0, Skipped: 0, BUILD SUCCESS.
- Diff is scoped to exactly 2 files (+23/-5), 1 commit; PR body correctly scopes to Bug #76472 Cluster only, explicitly noting the other 4 sibling areas (Properties/Licensing/Presentation/Providers) are separate PRs.
- No unrelated changes bundled in. Method body is syntactically well-formed Java (verified by direct read).
- Adversarial pass: checked the empty-`changes`-list edge case (no new issue — deterministic regardless of task, and this was never task-invariant pre-fix either, so not a regression), and confirmed the other two pre-existing hash tests in the same class still pass and still correctly assert status-sensitivity (contrast against the new task-insensitivity test).

No findings at reportable confidence (>= 80). Verdict: approve.

PR #5018, opened 2026-09-04
Fix false PlanHashMismatchException on paraphrased task (Bug #76472, Presentation area)

fredaan:
## Summary
- `PresentationChangePlanService.hash()` seeded its canonical string with the free-text `task` description. `task` is validated only for non-blank and is never read back or compared downstream -- it is a write-only audit label (`AdminChangeRecord.setTaskDescription`, `PresentationChangesetApplyService.writeAudit`). Paraphrasing `task` between a `preview` call and the matching `apply` call, with identical `changes`, produced a different `planHash` and a false `PlanHashMismatchException` (HTTP 409) on `apply`.
- Presentation has two call sites that compute the hash: `PresentationChangePlanService.resolve()` and `PresentationChangesetApplyService.apply()` (which recomputes the hash locally from its own re-resolved `changes` + `task`, rather than going through `resolve()`). Both were updated to drop `task` from `hash()`'s signature.
- Added a plan-level regression test (`PresentationChangePlanServiceTest.hashIsUnaffectedByDifferentTaskStrings`: two different `task` strings, identical `changes`, asserts equal `planHash()`).
- Added an apply-level regression test (`PresentationChangesetApplyServiceTest.applySucceedsWhenTaskIsParaphrasedBetweenPreviewAndApply`: `task` paraphrased between preview and apply, identical `changes`, asserts `apply()` succeeds instead of throwing `PlanHashMismatchException`) -- this specifically covers the second call site, since a plan-service-only test would not catch a regression reintroduced there alone.

This is 1 of 5 independent PRs fixing the same defect shape across different admin-chat areas for Bug #76472; this PR is scoped to the Presentation area only (`PresentationChangePlanService`/`PresentationChangesetApplyService`).

## Test plan
- [x] `./mvnw -pl core test -Dtest='PresentationChangePlanServiceTest,PresentationChangesetApplyServiceTest' -o` -- `Tests run: 34, Failures: 0, Errors: 0, Skipped: 0, BUILD SUCCESS` (26 + 8, including both new regression tests)
- [x] Verified red -> green: reverted the two production-file edits only (`git apply -R` on a source-only patch), re-ran the same command -- both new tests failed (`hashIsUnaffectedByDifferentTaskStrings` assertion mismatch, `applySucceedsWhenTaskIsParaphrasedBetweenPreviewAndApply` threw `PlanHashMismatchException`); re-applied the fix and both passed.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/presentation/PresentationChangePlanService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/presentation/PresentationChangePlanService.java
@@ -85,7 +85,7 @@ public class PresentationChangePlanService {
 
       String task = req.getTask().trim();
       List<PlanChange> immutableChanges = Collections.unmodifiableList(changes);
-      return new ResolvedPlan(task, immutableChanges, true, true, hash(task, immutableChanges));
+      return new ResolvedPlan(task, immutableChanges, true, true, hash(immutableChanges));
    }
 
    /**
@@ -360,9 +360,18 @@ public class PresentationChangePlanService {
    /** SHA-256 over the canonical plan, same field-order/control-character-free contract as every
     * other area's own {@code hash} method. Package-visible so {@link PresentationChangesetApplyService}
     * can recompute the identical hash from the same {@link ResolvedChange#planChange()} list
-    * {@link #resolveEntries} already produced, instead of resolving the whole plan a second time. */
-   static String hash(String task, List<PlanChange> changes) {
-      StringBuilder canonical = new StringBuilder(task).append('\n');
+    * {@link #resolveEntries} already produced, instead of resolving the whole plan a second time.
+    *
+    * <p>Deliberately does NOT take {@code task}: the free-text task description is never
+    * canonicalized or otherwise constrained (only checked for non-blank in {@link #resolveEntries}),
+    * so two equally valid paraphrases of the same intent over the identical {@code changes} would
+    * otherwise hash differently and trip a false {@code PlanHashMismatchException} between preview
+    * and apply. {@code task} is write-only downstream -- it flows into
+    * {@link PresentationChangesetApplyService#writeAudit} as
+    * {@code AdminChangeRecord.setTaskDescription}, a plain audit label that is never read back or
+    * compared -- so excluding it from the hash does not weaken the gate against a changed plan. */
+   static String hash(List<PlanChange> changes) {
+      StringBuilder canonical = new StringBuilder();
 
       for(PlanChange change : changes) {
          canonical.append(change.property()).append(SEP)
--- a/core/src/main/java/inetsoft/web/admin/ai/presentation/PresentationChangesetApplyService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/presentation/PresentationChangesetApplyService.java
@@ -88,7 +88,7 @@ public class PresentationChangesetApplyService {
          }
 
          String task = req.getTask().trim();
-         String currentHash = PresentationChangePlanService.hash(task, planChanges);
+         String currentHash = PresentationChangePlanService.hash(planChanges);
 
          if(req.getPlanHash() == null || !currentHash.equals(req.getPlanHash())) {
             throw new AdminChangesetApplyService.PlanHashMismatchException(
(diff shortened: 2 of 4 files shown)


fredaan:
## Round 1 review — Bug #76472, Presentation area

**Head commit reviewed:** b3166dcfe (single commit on `fix/76472-presentation-planhash-task`)

### What was checked
- Both call sites that compute the plan hash for this area:
  - `PresentationChangePlanService.resolve()` (`hash(immutableChanges)`)
  - `PresentationChangesetApplyService.apply()` (`PresentationChangePlanService.hash(planChanges)`), which recomputes locally rather than going through `resolve()` — this area's distinguishing two-call-site shape per the diagnosis/refute docs.
- Signature consistency: both call sites now pass the single `List<PlanChange>` argument, same order, no stale/mismatched-argument risk (`task` local var is untouched at both sites and still feeds the audit-label paths).
- Grepped the whole `core/src/main/java` tree for any other caller of `PresentationChangePlanService.hash` or `.hash(` inside the `presentation` package — confirmed exactly the two call sites above, no third site (e.g. `AdminPresentationController` only delegates to `resolve()`/`apply()`, does not touch `hash()` directly).
- Confirmed `task` is still used everywhere it's supposed to be (audit label / `ResolvedPlan.task()` / `writeAudit`/`rollback`), only dropped from the hash's canonical string.
- Javadoc on `hash()` explains *why* `task` is excluded (write-only audit label, never read back/compared, cites the exact `writeAudit`/`setTaskDescription` chain) — matches the reference-fix pattern from the other already-fixed areas.
- Test coverage: a plan-level test (`hashIsUnaffectedByDifferentTaskStrings`) and, importantly for this area's two-call-site shape, an apply-level test (`applySucceedsWhenTaskIsParaphrasedBetweenPreviewAndApply`) that exercises `service.apply()` end-to-end with a paraphrased task and asserts it succeeds rather than throwing `PlanHashMismatchException`. This is exactly the test the diagnosis flagged as necessary because a plan-service-only regression test would not catch a future regression reintroduced at the apply call site alone.
- Sibling/neighboring invariant: `applyRefusesAStaleHash` (a genuinely wrong/garbage `planHash` string) is untouched and still passes — confirms this fix only removes `task`'s influence on the hash, not the hash's ability to catch a real plan-content mismatch.
- No unrelated changes bundled: diff touches exactly 4 files (`PresentationChangePlanService.java`, `PresentationChangesetApplyService.java`, and their two test files), matching the PR's own stat.
- Commit message and PR title/description reference Bug #76472 and the Presentation area specifically, and note this is 1 of 5 sibling PRs (does not restate or reference the other 4 areas' internals).
- Re-ran the tests myself: `./mvnw -pl core test -Dtest='PresentationChangePlanServiceTest,PresentationChangesetApplyServiceTest' -o` → `Tests run: 34, Failures: 0, Errors: 0, Skipped: 0, BUILD SUCCESS` (26 in `PresentationChangePlanServiceTest`, 8 in `PresentationChangesetApplyServiceTest`, both class names confirmed in Surefire output, nonzero).

### Alternate scenarios tried (counter-example attempts)
- Tried to find a third call site that would need the same signature change (elsewhere in the module, in a scheduled job, in an export/revert path) — none found; `hash()` is package-private `static` with exactly two callers, both updated.
- Tried to construct an argument-order/mismatch bug from the signature change (single-arg signature removes any ordering ambiguity) — none found.
- Tried to find a scenario where the fix would also mask a genuine plan-content change (i.e. hash now always matches regardless of `changes`) — ruled out: `hash()` still folds in every `PlanChange` field, and the untouched `applyRefusesAStaleHash` test still exercises a real mismatch.
- Checked whether the new apply-level test really goes through the real `apply()` gate (not a mock) — confirmed, it calls `service.apply(apply, user)` on the real service instance.

Could not build a scenario where this change fails.

### Findings from previous round
N/A — round 1.

### New findings this round
None.

### Verdict
No findings. Approving.

Note: this batch's 5 sibling PRs were opened under the same `gh` identity this review is running as, so `gh pr review --approve` on this PR is expected to be rejected by GitHub as a self-review (already observed happening on 2 other sibling PRs in this batch). If that happens here too, this comment stands as the approval verdict; a human needs to formally approve/merge.

PR #5019, opened 2026-09-04
fix(#76472-providers): exclude task from ProviderChangePlanService plan hash

fredaan:
## Summary
- Bug #76472 (Providers area, one of 5 independent sibling areas in this batch — this PR covers Providers only).
- `ProviderChangePlanService.hash()` seeded its SHA-256 canonical string with the free-text `task` narrative. `task` is verified audit-label-only downstream (`ProviderChangesetApplyService.writeAudit`'s `record.setTaskDescription(task)`) — never compared, parsed, or gated on anywhere in the apply/rollback flow.
- Consequence: if a caller (e.g. an LLM agent) paraphrases `task` between `preview` and `apply` while `changes` and live provider-chain state stay identical, `resolve()` at apply time recomputes a different `planHash`, and `apply()`'s equality gate throws `PlanHashMismatchException` (HTTP 409) even though nothing about the actual plan changed.
- Fix: drop `task` from `hash()`'s parameter list and canonical-string seed; update the sole call site in `resolve()` to `hash(changes, chainProjections)`. `chainProjections` (a live concurrent-modification detector for touched provider chains, independent of `task`) and the `changes` loop are left byte-for-byte untouched, per the same "drop only the first parameter" pattern already applied to the DataSource/Viewsheet siblings in #76454.

## Test plan
- [x] Added `hashIsUnaffectedByDifferentTaskStrings` to `ProviderChangePlanServiceTest.java`: two `resolve()` calls with different `task` strings but identical `changes`/chain state assert equal `planHash()`.
- [x] Confirmed red before the fix (reverted just the source file via `git apply -R`, re-ran the suite): `Tests run: 30, Failures: 1` — the new test failed with the exact hash-divergence mechanism.
- [x] Confirmed green after re-applying the fix: `Tests run: 30, Failures: 0` (`ProviderChangePlanServiceTest`) and `Tests run: 13, Failures: 0` (`ProviderChangesetApplyServiceTest`, confirming no collateral breakage on the apply-side gate).
- Command: `cd E:/wt/76472-providers && ./mvnw -pl core test -Dtest='ProviderChangePlanServiceTest,ProviderChangesetApplyServiceTest' -o`

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/providers/ProviderChangePlanService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/providers/ProviderChangePlanService.java
@@ -123,7 +123,7 @@ public class ProviderChangePlanService {
 
       String task = req.getTask().trim();
       return new ResolvedPlan(task, Collections.unmodifiableList(changes), true, true,
-                              hash(task, changes, chainProjections));
+                              hash(changes, chainProjections));
    }
 
    // ---------------------------------------------------------------- create
@@ -502,11 +502,16 @@ public class ProviderChangePlanService {
     * reorder on a touched chain changes that chain's projection, so {@code apply}'s fresh re-resolve
     * produces a different hash and is refused via the existing 409 path -- no new conflict-handling
     * code needed.
+    * <p>
+    * {@code task} is deliberately excluded from this hash: it is a free-text audit narrative, never
+    * compared, parsed, or gated on -- it flows only into {@code ProviderChangesetApplyService
+    * .writeAudit}'s {@code record.setTaskDescription(task)}. Including it here would fail the
+    * apply-time plan-hash check whenever a caller paraphrases the task description between preview
+    * and apply while the actual {@code changes}/chain state stay identical.
     */
-   private static String hash(String task, List<PlanChange> changes,
-                              Map<ProviderChain, String> chainProjections)
+   private static String hash(List<PlanChange> changes, Map<ProviderChain, String> chainProjections)
    {
-      StringBuilder canonical = new StringBuilder(task).append('\n');
+      StringBuilder canonical = new StringBuilder();
 
       for(ProviderChain chain : ProviderChain.values()) {
          String projection = chainProjections.get(chain);
--- a/core/src/test/java/inetsoft/web/admin/ai/providers/ProviderChangePlanServiceTest.java
+++ b/core/src/test/java/inetsoft/web/admin/ai/providers/ProviderChangePlanServiceTest.java
@@ -400,6 +400,16 @@ class ProviderChangePlanServiceTest {
       assertEquals(first.planHash(), second.planHash());
    }
 
+   @Test void hashIsUnaffectedByDifferentTaskStrings() throws Exception {
+      stubHealthyAuthenticationChainOf("p1", "p2");
+      when(authenticationProviderService.getAuthenticationProvider("p2")).thenReturn(fileModel("p2"));
+
+      ResolvedPlan first = service.resolve(request("delete p2", List.of(deleteAuth("p2"))), user);
+      ResolvedPlan second = service.resolve(
+         request("please remove provider p2", List.of(deleteAuth("p2"))), user);
+      assertEquals(first.planHash(), second.planHash());
+   }
+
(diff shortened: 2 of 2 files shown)


fredaan:
## Reviewer round 1 — bug-76472-providers (PR #5019)

**Head commit reviewed:** `471fc6493` (single commit on `fix/76472-providers-planhash-task`, based on `stylebi-wiz-main-rebase`).

### What was checked
- Read `01-diagnosis.md` and `02-refute.md` and verified the diff against the confirmed cause (not just the reported symptom).
- Confirmed `hash()` dropped only `task` from its parameter list/canonical-string seed; `chainProjections` (`Map<ProviderChain,String>`) is byte-for-byte untouched in both position and the chain-projection loop (lines 512-522 in the current file) — verified by direct read, not by trusting the PR description.
- Confirmed `ProviderChangesetApplyService.java` has no direct `hash()` call anywhere (grepped `hash(` across the file — zero hits) and every `task` reference in that file (10 call sites: `apply`, `applyOne`, four `apply*` methods, `rollback`, four `rollback*` methods) is a pure pass-through terminating in `writeAudit`/`record.setTaskDescription(task)` — never compared, parsed, or branched on. The apply-side gate stays the simple `plan.planHash().equals(req.getPlanHash())` string comparison; nothing there needed a matching fix.
- Checked the fix is a general signature change (drop the parameter entirely), not a literal-keyed special case — no `if` excludes a specific `task` string; the whole parameter is gone from the method, so every caller/every task string is affected uniformly. No overfit smell.
- **EXPERIMENT leftover check (explicit, since 02-refute.md flagged `git status` as unreliable for this file in this worktree):** I did not rely on `git status`. I ran `git show HEAD:.../ProviderChangePlanServiceTest.java | grep EXPERIMENT` (no match), grepped the working-tree file directly (no match), and cross-checked `git hash-object` of the working file against `git ls-files -s` for the same path — both resolve to `675a28eca5...`, matching the blob the committed diff hunk header shows (`index acae41002c..675a28eca5`). **Confirmed: the `EXPERIMENT_hashDiffersWhenOnlyTaskParaphrased` leftover the refuter found is absent from both the working tree and the committed HEAD.** `gh pr diff 5019` shows only the intended production + test changes (2 files, 19 insertions / 4 deletions) — no stray experiment code.
- Confirmed the new regression test `hashIsUnaffectedByDifferentTaskStrings` genuinely proves task-invariance: same `changes`/chain-state helpers as the existing `hashIsStableAcrossTwoIdenticalResolutions`, only the `task` string differs between the two `resolve()` calls (`"delete p2"` vs `"please remove provider p2"`), asserting `planHash()` equality. Well-named, not keyed to the repro's exact wording (any two differing strings would exercise the same invariant).
- Re-ran the tests myself: `./mvnw -pl core test -Dtest='ProviderChangePlanServiceTest,ProviderChangesetApplyServiceTest' -o` → `Tests run: 43, Failures: 0, Errors: 0` (30 + 13), `BUILD SUCCESS`.
- Confirmed Javadoc added on `hash()` explains the WHY (task is audit-label-only, cites `ProviderChangesetApplyService.writeAudit`/`setTaskDescription` as the safety argument, and states the consequence of including it).
- Confirmed no unrelated changes bundled: `git diff --stat` vs base shows exactly the 2 files (source + test), 19 insertions / 4 deletions, single commit.
- Commit message and PR description reference Bug #76472 / Providers specifically ("Bug #76472 (Providers area, one of 5 independent sibling areas in this batch — this PR covers Providers only)"), not the whole batch.
- Alternate-scenario sweep: checked the sibling create-path (`resolveCreate`) and delete-path (`resolveDelete`) both feed the same single `hash()` call site at the end of `resolve()` — no separate hash path for create vs delete that could have been missed. Checked `PlanChange`'s fields consumed by `hash()`'s loop (`property`, `currentValue`, `proposedValue`, `risk`, `snapshotScope`) — none depend on `task`, so no residual coupling. No other file in `providers/` calls the now-private, still-two-argument `hash()`.

### Previous-round findings
None — this is round 1.

### New findings this round
No findings.

### Verdict
**Approved** (attempted via `gh pr review 5019 --approve`; expected to be rejected by GitHub as a self-review, since this PR was opened under the same `gh` identity — noting per the batch's known self-review restriction rather than fighting it).

PR #5021, opened 2026-09-05
Give CheckBox/RadioButton titles an inset header instead of a floating legend

swaker854:
## Summary
- The default (no custom border) title used a fieldset-legend style: the object's default border ran through the middle of the title text, masked by a hardcoded white title background that didn't match the actual canvas/page (or a customized viewsheet background).
- Replaced it with the pattern selection list already uses: the title sits fully inside the shared object border as a header row, separated from the body by a bottom rule (`vs-title`'s new `insetHeader` input), so there's no border/text overlap to mask.
- Also centers the title text vertically, adds more left padding, and truncates overly long titles with an ellipsis instead of clipping/overflowing.

## Test plan
- [ ] Composer: add a CheckBox and RadioButton with default formatting, confirm the title renders as a header row inside the border with a bottom separator, no line crossing the text
- [ ] Confirm title text is vertically centered with left padding
- [ ] Confirm a long title truncates with `…`
- [ ] Confirm a CheckBox/RadioButton with a user-customized title border/padding still respects that customization (insetHeader should not apply)


swaker854:
The \`build\` check has failed on 3 consecutive runs (most recently [this run](https://github.com/inetsoft-technology/stylebi/actions/runs/33939528539)), each time with the identical error:

\`\`\`
PermissionHierarchyTest.setUp:96 » IllegalArgument argument "in" is null
\`\`\`

This isn't related to this PR's changes (only \`ListInputVSAssemblyInfo.java\`, \`vs-title\`, \`vs-check-box\`, \`vs-radio-button\`, \`vs-compound.scss\` — nothing in \`inetsoft.sree.security\`). It's a known flaky test:

- The same failure also broke an unrelated branch's build (\`fix/bug-76449-wbs004-column-selection-qualified-lookup\`) around the same time.
- It matches two already-filed issues describing a config-file read race in \`SecurityChain.loadConfiguration()\`: #4984 ("guard against a null stream in SecurityChain.loadConfiguration") and #4994 ("guard SecurityChain.loadConfiguration() against a config-file read race").

Flagging for the CI/infra team rather than continuing to retry blindly — this PR is otherwise green (Semgrep, CLA, and the automated code review all pass).

jshobe-inetsoft (approved):

PR #5022, opened 2026-09-05
Add regression test for bug #74951 (Ignite shutdown race, already fixed)

fredaan:
## Summary

This PR touches **only test code, no production source**.

Bug #74951 reported an `IgniteIllegalStateException` thrown from
`IgniteCluster.removeCacheEventListener()` (via `LocalKeyValueStorage.close()`
-> `BlobStorage.close()` -> `Plugins.close()`) when Ignite's kernal gateway was
already `STOPPED` during Spring bean shutdown.

Investigation (diagnosis + independent adversarial refute, both archived
internally) found this bug was already fixed — as an unintentional side effect
— by commit `1b6ae84fbc` ("Epic 70095 Distributed Sessions", 2026-05-26, ~12
days after this bug was filed), which added:
- `mapListeners.clear()` in `IgniteCluster.close()`
- an `if(!closed)` guard (redundant with the existing `mapListeners.get(name)
  != null` check) in `IgniteCluster.removeCacheEventListener()`

No production code change is proposed here. This PR adds a permanent,
non-`@Disabled` JUnit regression test so a future refactor of `IgniteCluster`
can't silently reintroduce the race.

## Test plan

- [x] New test `IgniteClusterShutdownRaceTest` (embedded single-node Ignite,
      no Spring context beyond the config scaffolding shared with sibling
      tests in this package) registers a replicated map listener, calls
      `cluster.close()` (simulating shutdown having already run), then calls
      `removeReplicatedMapListener()` for the same name and asserts it does
      not throw.
- [x] Confirmed the test **passes** against current `HEAD`.
- [x] Adversarially reverted the two guard lines above (temporarily, not
      committed) and confirmed the test **fails** with the exact reported
      exception: `IgniteIllegalStateException: ... state=STOPPED`. Restored
      the guards and reconfirmed `git status --porcelain` clean aside from
      the new test file.
- [x] Ran the full `core` module test suite (`./mvnw -o -pl core -am test`):
      `Tests run: 8758, Failures: 0, Errors: 0, Skipped: 69` (the 69 skips are
      pre-existing `@Disabled` classes, unrelated to this change).

Note: the existing `IgniteClusterTest`/`IgniteDistributedMapTest` etc. in this
package are entirely `@Disabled` at the class level (confirmed via git blame
this predates and is unrelated to `1b6ae84fbc`), so this regression test was
added as a new, separate, non-`@Disabled` test class rather than a method on
one of those classes.

### DIFF
--- a/core/src/test/java/inetsoft/sree/internal/cluster/ignite/IgniteClusterShutdownRaceTest.java
+++ b/core/src/test/java/inetsoft/sree/internal/cluster/ignite/IgniteClusterShutdownRaceTest.java
@@ -1,0 +1,75 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.sree.internal.cluster.ignite;
+
+import inetsoft.sree.internal.cluster.*;
+import inetsoft.test.*;
+import org.apache.ignite.spi.discovery.tcp.ipfinder.TcpDiscoveryIpFinder;
+import org.apache.ignite.spi.discovery.tcp.ipfinder.vm.TcpDiscoveryVmIpFinder;
+import org.junit.jupiter.api.*;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.io.TempDir;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import java.nio.file.Path;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
+/**
+ * Regression test for bug #74951: removing a replicated map listener after the
+ * cluster has already been closed (e.g. shutdownInetsoft()/ClusterJobStore
+ * closing the Ignite-backed cluster before a bean like Plugins/BlobStorage
+ * unregisters its listener during its own shutdown) must not throw
+ * IgniteIllegalStateException. Fixed as a side effect of commit 1b6ae84fbc.
+ */
+@ExtendWith(SpringExtension.class)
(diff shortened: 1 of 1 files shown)


fredaan:
## Review — PR #5022 (bug #74951, test-only regression test)

**Verdict: approve.** Independently re-verified every material claim; no issues found.

### What I checked myself (not just re-reading the archived diagnosis/refute/fix docs)

1. **Diff is test-only** — `gh pr diff 5022 --name-only` shows exactly one file: `core/src/test/java/inetsoft/sree/internal/cluster/ignite/IgniteClusterShutdownRaceTest.java` (new file, 75 lines). No `core/src/main/java/...` touched.
2. **Existing `IgniteClusterTest.java` is genuinely `@Disabled` at class level** — read it directly (line 44: `@Disabled` above the class declaration, before all the Spring/`@SreeHome` scaffolding). Confirms the stated reason for adding a new class rather than un-disabling a method: JUnit 5's `@Disabled` at class level cannot be selectively bypassed per-method.
3. **Read `IgniteCluster.java` directly** at the two guard sites the PR depends on (current worktree, branch `fix/bug-74951-ignite-shutdown-regression-test`, clean checkout):
   - `close()` (1684-1727): `mapListeners.clear()` at line 1702, inside `synchronized(mapListeners)`, strictly before `ignite.close()` at 1707.
   - `removeCacheEventListener()` (1896-1918): guarded by both `mapListeners.get(name) != null` (1898-1900) and `if(!closed)` (1902) around the `ignite.events().stopLocalListen/stopRemoteListen` calls.
   Matches the PR description and the archived diagnosis/refute exactly.
4. **Ran the test myself**: `./mvnw -o -pl core -am test -Dtest=IgniteClusterShutdownRaceTest -Dsurefire.failIfNoSpecifiedTests=false` → `Tests run: 1, Failures: 0, Errors: 0`. Real embedded Ignite node started/stopped in the log (not a mock).
5. **Independently reproduced the adversarial mutation**: reverted `mapListeners.clear()` in `close()` and changed `if(!closed)` → `if(true)` in `removeCacheEventListener()`, reran the same test → **failed** with the exact reported exception (`IgniteIllegalStateException: ... state=STOPPED`), thrown from `IgniteCluster.removeCacheEventListener(IgniteCluster.java:1902)` via `removeReplicatedMapListener` → the test's `assertDoesNotThrow` lambda. Restored the file from backup, confirmed `git status --porcelain` clean before and after.
6. **CI on this PR**: Semgrep (pass), CLA (pass), `claude-review` bot (pass, independently traced the same guard logic and reached the same conclusion — including confirming that reverting only *one* of the two guard lines would not reproduce the crash, both must go together). `build` check is a full Validation Build; I separately confirmed the targeted module build (`-pl core -am test`) passes clean both before and after my mutation experiment.
7. **PR description honesty** — correctly states test-only, no production change, and attributes the actual fix to commit `1b6ae84fbc` rather than claiming credit in this PR.

### Non-blocking observations (already surfaced by the `claude-review` bot, agree with all three, none rise to blocking)
- This is the first non-`@Disabled` test in the package that spins up a *real* embedded Ignite node (vs. the mocked `IgniteClusterComputeRoutingTest`) — worth being aware of for CI time/flakiness, though a clean full-suite run was already confirmed.
- License header says `2024` vs. `2026` on the most recently added sibling file — cosmetic only.
- `assertDoesNotThrow` has no failure-message arg — would make a future CI failure slightly more self-explanatory, but not required.

No correctness, security, or regression concerns. Approving.

PR #5023, opened 2026-09-05
test(PWA-010): pin no-save cross-runtime staleness as current, by-design behavior

larryliang-inetsoft:
## Summary
- Adds `noSaveEverLeavesColumnsStaleOnEveryCall` to `VSTreeHandlerCrossRuntimeStalenessTest`, alongside the existing #76471 (`#5011`) tests.
- `#5011` fixed a "first read after a persisted save" race in `CubeTreeModelBuilder`. This new test models the distinct, still-open scenario where **no save ever happens** between a worksheet-editing session's structural edit and a separate viewsheet/binding session's read — inverting the one precondition (`store.lastModified()` vs. `rvs.getLastReset()`) that distinguishes the two scenarios.
- Not a behavior fix — pins current, intended architectural behavior (two separate runtimes, no live in-memory bridge) so a future deliberate change in this area (a detect-and-warn signal, or a live cross-runtime bridge) is a conscious decision, not an accidental regression caught by a red test. Javadoc says so explicitly.
- Confirms `resetRuntime()` never fires in this scenario (via `Mockito.verify(..., never())`), not just that the columns stay stale — pins the mechanism, not only the symptom.
- Companion plugin-side PR (stylebi-wiz, doc-strengthening): https://github.com/inetsoft-technology/stylebi-wiz/pull/new/fix/pwa-010-stale-bindable-fields (cross-reference once both are up).

## Test plan
- [x] `./mvnw test -pl core -Dtest=VSTreeHandlerCrossRuntimeStalenessTest` — 3/3 passing (the new test plus both existing #76471 tests, confirming no collateral breakage).

### DIFF
--- a/core/src/test/java/inetsoft/web/binding/handler/VSTreeHandlerCrossRuntimeStalenessTest.java
+++ b/core/src/test/java/inetsoft/web/binding/handler/VSTreeHandlerCrossRuntimeStalenessTest.java
@@ -218,6 +218,95 @@ class VSTreeHandlerCrossRuntimeStalenessTest {
       }
    }
 
+   /**
+    * Fixture for the "no {@code save_worksheet} ever happens" scenario (bug PWA-010, see
+    * {@code docs/teams/2026-09-05-bug-pwa-010/} in the stylebi-wiz repo): the worksheet-editing
+    * session's {@code set_group_aggregate} is applied only to that session's own in-memory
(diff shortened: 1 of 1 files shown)


larryliang-inetsoft:
Companion plugin-side doc fix: https://github.com/inetsoft-technology/stylebi-wiz/pull/2217

PR #5024, opened 2026-09-06
fix(#76478): resolve/validate input-assembly variable table bindings

larryliang-inetsoft:
## Summary
- `set_input_value` reported `{ok:true}` while silently no-op'ing: Slider/ComboBox/Spinner/TextInput's property-dialog setters copied `dataInputPaneModel.getTable()` into the assembly's `tableName` with no validation, so a caller binding `table:"Variables"` (the Composer data-input tree's own non-selectable folder label, not a real leaf value) instead of the required `"$(varName)"` form produced a binding that read back as configured but never resolved in `refreshVariable()` at apply time.
- Adds `resolveInputTableBinding()`, wired into all four affected setters: passes through a valid `"$(varName)"` reference or a real worksheet table/assembly name unchanged, normalizes two unambiguous natural mistakes (a bare variable name; the `"Variables"` folder label paired with a `columnValue` naming a real variable) into the `"$(varName)"` form the runtime requires, and throws loud (naming the bad value) otherwise — mirroring the existing `set_variable_values`/`editVariable` existence-check precedent (`WorksheetAgentController.java:2308`).
- CheckBox/RadioButton (a related but structurally different gap — they trust a UI-only boolean with no plugin-settable alias) and `InputValueService.setValue()`'s own unconditional `{ok:true}` are deliberately left untouched, out of scope for this fix.

Root-caused via `stylebi-wiz` bug #76478 (Redmine); full diagnosis/refute/fix write-up at `stylebi-wiz` repo `docs/teams/2026-09-05-bugs-76479-76478-76475/bug-76478/`.

## Test plan
- [x] New regression test `VSInputTableBindingResolutionTest` (8 cases: valid `$(var)` pass-through, unknown-variable rejection, the exact reported `"Variables"`+columnValue shape, bad columnValue rejection, bare-name normalization, real-table pass-through, unresolvable-value rejection, null/empty pass-through) — `./mvnw -q -pl core test -Dtest=VSInputTableBindingResolutionTest` — 8/8 pass
- [x] No regression in neighboring suites — `./mvnw -q -pl core test -Dtest="VSComboBoxDateValueTest,CalendarInputServiceTest,AssemblyPropertyServiceTest,WorksheetAgentControllerTest"` — all green (7/7, 29/29, 13/13, 131/131)
- [ ] Live verification via `set_input_value`/`get_viewsheet_image` (not possible this session — no Chrome browser with the Claude extension connected)

No LLM-facing prompt/schema/tool-description text was touched by this PR — Java-side only, no merge hold needed.

### DIFF
--- a/core/src/main/java/inetsoft/web/viewsheet/service/VSInputService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/service/VSInputService.java
@@ -871,7 +871,9 @@ public class VSInputService {
       textInputAssemblyInfo.setInsetStyle(textInputGeneralPaneModel.isInsetStyle());
       textInputAssemblyInfo.setMultiline(textInputGeneralPaneModel.isMultiLine());
 
-      String table = dataInputPaneModel.getTable();
+      String table = resolveInputTableBinding(viewsheet.getViewsheet().getBaseWorksheet(),
+         viewsheet.getViewsheet(), dataInputPaneModel.getTable(),
+         dataInputPaneModel.getColumnValue());
       textInputAssemblyInfo.setTableName(table == null ? "" : table);
       textInputAssemblyInfo.setColumnValue(dataInputPaneModel.getColumnValue());
       textInputAssemblyInfo.setRowValue(dataInputPaneModel.getRowValue());
@@ -1093,7 +1095,9 @@ public class VSInputService {
       sliderAssemblyInfo.setPrimary(basicGeneralPaneModel.isPrimary());
       sliderAssemblyInfo.setVisibleValue(basicGeneralPaneModel.getVisible());
 
-      String table = dataInputPaneModel.getTable();
+      String table = resolveInputTableBinding(viewsheet.getViewsheet().getBaseWorksheet(),
+         viewsheet.getViewsheet(), dataInputPaneModel.getTable(),
+         dataInputPaneModel.getColumnValue());
       sliderAssemblyInfo.setTableName(table == null ? "" : table);
       sliderAssemblyInfo.setColumnValue(dataInputPaneModel.getColumnValue());
       sliderAssemblyInfo.setRowValue(dataInputPaneModel.getRowValue());
@@ -1996,7 +2000,9 @@ public class VSInputService {
       setListValues(comboBoxAssemblyInfo, value, viewsheet, principal);
 
       // TODO validate column/row variable/expression type
-      String table = dataInputPaneModel.getTable();
+      String table = resolveInputTableBinding(viewsheet.getViewsheet().getBaseWorksheet(),
+         viewsheet.getViewsheet(), dataInputPaneModel.getTable(),
+         dataInputPaneModel.getColumnValue());
       comboBoxAssemblyInfo.setTableName(table == null ? "" : table);
       comboBoxAssemblyInfo.setColumnValue(dataInputPaneModel.getColumnValue());
       comboBoxAssemblyInfo.setRowValue(dataInputPaneModel.getRowValue());
@@ -2287,7 +2293,9 @@ public class VSInputService {
       spinnerAssemblyInfo.setPrimary(basicGeneralPaneModel.isPrimary());
       spinnerAssemblyInfo.setVisibleValue(basicGeneralPaneModel.getVisible());
 
-      String table = dataInputPaneModel.getTable();
+      String table = resolveInputTableBinding(viewsheet.getViewsheet().getBaseWorksheet(),
+         viewsheet.getViewsheet(), dataInputPaneModel.getTable(),
+         dataInputPaneModel.getColumnValue());
       spinnerAssemblyInfo.setTableName(table == null ? "" : table);
       spinnerAssemblyInfo.setColumnValue(dataInputPaneModel.getColumnValue());
       spinnerAssemblyInfo.setRowValue(dataInputPaneModel.getRowValue());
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
## Review round 1 -- CHANGES REQUESTED (1 Important finding)

Reviewed the full diff (`gh pr diff 5024`) against `01-diagnosis.md`/`02-refute.md`/`03-fix.md`. Scope confirmed Java-only (`VSInputService.java` + new test) -- no merge hold applies.

### Important: the reorder's own justification undershoots what `isKnownVariableName` actually matches

The merged fix checks `isKnownVariableName(ws, vs, table)` **before** `ws.getAssembly(table) != null`, justified in `03-fix.md` as "only fixing the shadowed case where the same name resolves to both a `VariableAssembly` and `ws.getAssembly()` generically." That's true of `isKnownVariableName`'s first branch, but its second branch (the merged-`UserVariable` fallback loop) draws from `Worksheet.getAllVariables()`, which pulls variable references from **every** `TABLE_ASSET` assembly's own conditions/SQL params -- not just `DefaultVariableAssembly`-backed ones (`Worksheet.java:580-597`, confirmed by direct read).

So: a worksheet with a real, unrelated table literally named `"Query1"`, plus some *other* table whose own condition references `$(Query1)`, will now have a caller's `table:"Query1"` (genuinely intending the real table) silently reclassified as `"$(Query1)"` (a variable reference) instead of passed through unchanged -- exactly the "bare name that's also a real, unrelated assembly's name" ambiguity flagged for this review. None of the 8 regression tests covers this (the real-table-passthrough test uses a worksheet with only that one assembly and no other variables at all).

Suggested fix: only let the `UserVariable` fallback loop win when `ws.getAssembly(name) == null` (no assembly at all owns that name), so a concrete worksheet assembly always outranks an incidental condition-variable reference of the same name. Add a test constructing exactly this collision.

Full writeup with line citations: `docs/teams/2026-09-05-bugs-76479-76478-76475/bug-76478/05-review-r1.md` in the `stylebi-wiz` repo.

### Everything else checked out on direct verification
- All four setters (Slider/ComboBox/TextInput/Spinner) confirmed wired to `resolveInputTableBinding`, using the same `getViewsheet().getBaseWorksheet()` accessor the real runtime `refreshVariable()` path uses.
- CheckBox/RadioButton's exclusion confirmed structurally distinct (trusts `dataInputPaneModel.isVariable()` directly; no `variable` alias exists in `PropertyAliases.java` for plugin callers to set it at all) -- not unaddressed scope-creep.
- The `"Variables"` literal is grounded in `getInputTablesTree`'s actual non-localized node `data` value, not an overfit special-case of the repro wording.
- Fail-loud `MessageException` text is field-named and clear.
- No sibling call site left with the same insecure resolution.
- Compiles cleanly against existing wildcard imports; the 8-test regression suite otherwise covers the diagnosed shape space soundly.

larryliang-inetsoft:
Addressed the Important finding on isKnownVariableName's fallback branch.

Confirmed the collision is real by tracing `Worksheet.getAllVariables()` directly (`core/src/main/java/inetsoft/uql/asset/Worksheet.java:580-597`): it merges `$(varName)` references out of *every* `TABLE_ASSET` assembly's own conditions/SQL params, not just `DefaultVariableAssembly`-backed names. So a `UserVariable` named e.g. `"Query1"` can appear purely because some unrelated table's condition references `$(Query1)`, while a separate real table literally named `Query1` also exists.

Fix (021be8e): `isKnownVariableName` now returns `false` immediately once `ws.getAssembly(name)` finds any concrete assembly (unless it's itself a `VariableAssembly`), so a real worksheet assembly always outranks an incidental same-named condition variable. This is the precedence you suggested, applied inside `isKnownVariableName` itself so both the wrapped (`$(name)`) and bare-name call sites get the same guarantee.

Added `aRealTableOutranksAnUnrelatedConditionVariableOfTheSameName`, constructing exactly the collision you described (a real `"Query1"` table + a separate table whose own condition references `$(Query1)`). Verified it fails without the fix (`expected: <Query1> but was: <$(Query1)>`) and passes with it.

Full suite re-run, all green:
- `VSInputTableBindingResolutionTest`: 9/9 (8 original + 1 new)
- `VSComboBoxDateValueTest`: 7/7
- `CalendarInputServiceTest`: 29/29
- `AssemblyPropertyServiceTest`: 13/13
- `WorksheetAgentControllerTest`: 131/131

larryliang-inetsoft:
## Review round 2 -- APPROVE, ready to merge

Verified round 1's Important finding independently against current source (`021be8e67`), not just the fix write-up.

**Round 1 finding: resolved.** Traced `isKnownVariableName` by hand for the exact collision (real table `"Query1"` + a separate `"OtherTable"` whose condition references `"$(Query1)"`): the new `assembly != null -> false` short-circuit correctly makes the real table win before the `UserVariable` fallback loop runs. The new test `aRealTableOutranksAnUnrelatedConditionVariableOfTheSameName` exercises this exact scenario (confirmed by reading it in full, including its sanity assertion that the collision is genuinely set up), not a weaker variant. Re-ran all five regression suites myself: 189/189 pass (`VSInputTableBindingResolutionTest` 9/9, `VSComboBoxDateValueTest` 7/7, `CalendarInputServiceTest` 29/29, `AssemblyPropertyServiceTest` 13/13, `WorksheetAgentControllerTest` 131/131).

**The fixer's own flagged residual case (wrapped `$(name)` mirror collision): judged informational, not blocking.** It requires the same rare naming collision as round 1 *plus* the caller specifically using the wrapped form to target the incidental (never-declared) reference -- a strictly narrower case, not a weaker version of the original bug. More importantly, the failure mode is the safe direction: it now fails loud with a clear, field-named error ("Use add_variable to create it first") instead of silently misresolving, which is the correct behavior per this repo's own tool-robustness principle for ambiguous input. Recommend a follow-up regression test to pin this intentional trade-off, but it does not block this PR.

Scope re-confirmed: still exactly two files (`VSInputService.java` + the new test), Java-only, no plugin schema/prompt touched, no merge hold. `mergeable: MERGEABLE`, checks green aside from `build` which was still catching up to the latest push at review time.

**Verdict: approved for merge.**

PR #5028, opened 2026-09-06
fix(wiz): add comparisonOption to the date-comparison wire schema (VBM-012)

larryliang-inetsoft:
## Root cause
`ViewsheetAssemblyAgentController.DateComparisonRequest` and `DateComparisonService.Comparison` (both Java records) had no `comparisonOption` field, so the agent tool's `set_date_comparison(comparisonOption: ...)` was rejected by Jackson with an "Unrecognized field" error before `apply()` ever ran. `read()` also echoed the raw `Calculator` int (1/2/6) instead of naming it — a genuine schema/backend mismatch, not a dead/disabled feature (unlike the sibling `frame` parameter removed in #4818, confirmed still live-wired through `ChartDcProcessor`/`CrosstabDcProcessor`/`DateComparisonInfo`).

## Fix
- Added `comparisonOption` to both records, wired through `apply()` using the same `value`/`change`/`percentChange` vocabulary the plugin (`dateComparisonTools.ts`) already sends and validates.
- Excluded from `setsPeriod()`'s gating, like `useFacet`/`onlyShowMostRecentDate` — it's period-independent, so a call setting only `comparisonOption` doesn't need an end anchor.
- `read()` now names the value instead of echoing the raw int, and additionally names the two composite settings reachable from the interactive Composer's own dialog but outside the plugin's 3-token vocabulary — `DateComparisonInfo.CHANGE_VALUE`/`PERCENT_VALUE` ("Change and Value" / "Percent Change and Value") — as `changeAndValue`/`percentChangeAndValue`, rather than falling back to `null` or a bare magic number. A value genuinely outside even that wider set (a `Calculator` type this dialog should never produce) still reads back as `null`.

## Relationship to #4775
This supersedes #4775, which implemented the same write path but predates this branch's base (frame-parameter removal in #4818, the chart-type-override disclosure in #5027, and several other merged agent-controller changes), and was blocked on review (`CHANGES_REQUESTED`) for reporting `CHANGE_VALUE`/`PERCENT_VALUE` as `null` on read. Addressed here by naming both instead of falling back — this fully satisfies the reviewer's "extend the vocabulary" option for the read side.

Extending the **write** side to accept those same two tokens is left as a follow-on: it would be inert today without a matching plugin-schema change (`dateComparisonTools.ts`'s `normalizeComparisonOption` only knows the three simple tokens, so an agent can never actually send `changeAndValue`/`percentChangeAndValue` even if the Java side accepted them), and extending it is a separate, larger scope decision (naming, whether to expose composite write modes at all) than this schema-mismatch bug.

## Tests
`DateComparisonServiceTest.java`:
- Round-trips `value`/`change`/`percentChange` (case-insensitively) through `apply()`.
- Refuses an unrecognized `comparisonOption` with a field-named error.
- A call setting only `comparisonOption` needs no end anchor (period-independent).
- A call that omits `comparisonOption` leaves an existing setting untouched.
- `read()` names `value`/`change`/`percentChange`/`changeAndValue`/`percentChangeAndValue` correctly.
- A truly out-of-vocabulary value (e.g. `Calculator.CUSTOM`) reads back as `null`.
- Updated the 5 pre-existing positional `new DateComparisonService.Comparison(...)` test-constructor call sites for the new field.

`./mvnw -pl core -am test -Dtest=DateComparisonServiceTest -Dsurefire.failIfNoSpecifiedTests=false` → 52/52 passing (12 new).

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/DateComparisonService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/DateComparisonService.java
@@ -21,9 +21,11 @@ import inetsoft.report.composition.RuntimeViewsheet;
 import inetsoft.uql.XConstants;
 import inetsoft.uql.erm.DataRef;
 import inetsoft.uql.viewsheet.*;
+import inetsoft.uql.viewsheet.graph.Calculator;
 import inetsoft.uql.viewsheet.graph.ChartAggregateRef;
 import inetsoft.uql.viewsheet.graph.GraphTypes;
 import inetsoft.uql.viewsheet.graph.VSChartInfo;
+import inetsoft.uql.viewsheet.internal.DateComparisonInfo;
 import inetsoft.web.composer.model.vs.*;
 import inetsoft.web.composer.vs.dialog.DateComparisonDialogService;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


larryliang-inetsoft:
Closing in favor of continuing directly on #4775 per coordinator direction — tracking VBM-012 against that PR rather than a fresh one. Will push a minimal fixup there (docstring/test acknowledgment of the CHANGE_VALUE/PERCENT_VALUE gap, not a vocabulary extension) and re-request review from the original reviewer.

larryliang-inetsoft:
Reopening — this is the authoritative fix for VBM-012 (complete, tested 52/52, cleanly reviewed by both an independent reviewer and the automated claude-review bot). #5028 and #4775 were both closed within a 7-second window a moment ago, each naming the other as the surviving PR — that did not come from the batch coordinator (me) and appears to be a stale/conflicting actor. #4775 stays closed (superseded, ~120 commits stale, no fixup was ever pushed to it).

larryliang-inetsoft:
Branch was deleted along with the PR close; re-pushed the identical commit and reopening. This is the authoritative fix for VBM-012 (52/52 tests, cleanly reviewed by both an independent reviewer and the automated claude-review bot). The prior close/reopen race a few minutes ago did not come from the batch coordinator.

PR #5038, opened 2026-09-07
fix(wiz): L8 parity fixes -- conditions, highlights, date comparison

NeverStopFMH:
## Summary
- Backend implementation for lane L8's parity-audit fixes to the wiz agent's condition/highlight/date-comparison endpoints (`inetsoft.web.wiz.viewsheet`). Companion PR on the plugin side: stylebi-wiz#2249.
- **Reconciles with an independently-landed upstream fix for the same gap**: commit 92730d759 ("add comparisonOption to the date-comparison wire schema", VBM-012) fixed the same `comparisonOption` schema mismatch for 3 of 5 UI values and explicitly flagged extending to the other 2 as a follow-on. This PR is that follow-on -- it keeps 92730d759's naming and its documented `null`-for-out-of-vocabulary design (confirmed by that commit's own regression test), and extends `COMPARISON_OPTION_WORDS`/`COMPARISON_OPTION_NAMES` to the 2 composite values (`changeAndValue`/`percentChangeAndValue`).
- Full audit trail -- per-finding evidence, owner routing, live-testing arms and results -- is in the companion plugin PR's `docs/parity/L8-conditions-highlights-date-comparison/L8-parity-report.md` (+ `.zh-CN.md`).

## What changed
- `ConditionVocabulary.java`: Top N/Bottom N ranking now builds a real `RankingValueModel` from `{n, groupField}` instead of crashing; added `equal` (or-equal-to) and `level` (nesting, item-level) to the read/write conversion; added typed condition values for `FIELD`/`VARIABLE`/`SESSION_DATA`/`EXPRESSION` (`SUBQUERY` intentionally deferred).
- `AssemblyHighlightService.java` / `ViewsheetAssemblyAgentController.java`: threaded a `fontInfo` parameter through to `HighlightModel`; threaded `axis`/`text` region flags through to `Region` (the backend model already supported both end-to-end -- this closed a wire-layer gap only).
- `DateComparisonService.java`: `comparisonOption` now covers all 5 real values on both read and write; added `shareAssembly`/`shareFrom` threading through the existing `getShare()`/`setDateComparison(shareAssembly,...)` calls; added the `toDate`/`inclusive` period-level pair; added interval-word write-side normalization (`INTERVAL_WORDS`/`normalizeInterval()`, mirroring the existing `normalizeLevel()`); wired the already-present `levelWord()` (and new `intervalWord()`) into the read path so `level`/`interval` report as words (`"year"`, `"yearToDate"`), not raw StyleBI codes -- a gap only found live-testing this same batch of fixes surfaced.

## Test plan
- [x] `mvn test` -- 237/237 passing for every touched class (`ConditionVocabularyTest`, `AssemblyHighlightServiceTest`, `DateComparisonServiceTest`, `ViewsheetAssemblyAgentControllerTest`, `AssemblyConditionServiceTest`, `ConditionUtilTest`) after a clean rebuild
- [x] Clean full `core` compile
- [x] Live-verified against a running instance rebuilt from this branch, through the plugin's own MCP tools on a scratch viewsheet -- see the companion PR's parity report for the specific arms and results (9 of these findings confirmed this way, including a visual render check for the font/axis-highlight changes)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/CalcTableService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/CalcTableService.java
@@ -711,7 +711,7 @@ public class CalcTableService {
          clauses.add(new ConditionVocabulary.Clause(
             column, String.valueOf(clause.get("operator")), values,
             junction == null ? null : String.valueOf(junction),
-            Boolean.TRUE.equals(clause.get("negated"))));
+            Boolean.TRUE.equals(clause.get("negated")), false, 0));
       }
 
       return clauses;
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyHighlightService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyHighlightService.java
@@ -284,12 +284,33 @@ public class AssemblyHighlightService {
 
       Map<String, Object> out = new LinkedHashMap<>();
       out.put("fontFamily", font.getFontFamily());
-      out.put("fontSize", font.getFontSize() == null ? null : Integer.valueOf(font.getFontSize()));
+      out.put("fontSize", describeFontSize(font.getFontSize()));
       out.put("bold", "bold".equals(font.getFontWeight()));
       out.put("italic", "italic".equals(font.getFontStyle()));
       out.put("underline", "underline".equals(font.getFontUnderline()));
       out.put("strikethrough", "strikethrough".equals(font.getFontStrikethrough()));
       return out;
+   }
+
+   /**
+    * {@code FontInfo.toFont()} parses the same stored string with an unguarded
+    * {@code Integer.parseInt}, so a malformed size is a pre-existing risk — but that path only
+    * runs once a highlight is applied, while this one is now reachable straight from
+    * {@code list_highlights}'s read. Reporting the raw stored value on a parse failure keeps the
+    * read itself from throwing, while still surfacing that something is off (a caller expecting
+    * an int wouldn't get a string here otherwise).
+    */
+   private static Object describeFontSize(String fontSize) {
+      if(fontSize == null) {
+         return null;
+      }
+
+      try {
+         return Integer.valueOf(fontSize);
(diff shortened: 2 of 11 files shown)


NeverStopFMH:
Addressed the review's observations:

1. **`ConditionVocabulary.java`'s duplicated field-list logic (fixed, `bab4c8705`)** — the unknown-field error at line ~178 now calls the existing `fieldList()` helper instead of repeating its `TreeSet`/`join` inline.
2. **`describeFont()`'s `NumberFormatException` risk on a malformed `fontSize` (fixed, `bab4c8705`)** — extracted into `describeFontSize()`, which now catches the parse failure and falls back to the raw stored string instead of throwing. This mirrors the same unguarded parse that already exists in `FontInfo.toFont()`, but that path only runs once a highlight is applied at render time, while `describeFont()` is now reachable straight from `list_highlights`' read — a read must not throw on stored data it didn't write. Added a regression test (`reportsTheRawFontSizeWhenItIsNotANumber`).
3. **`read()`'s new unconditional `getShare()` scan** — confirmed as acceptable as noted in the review (O(n) over assemblies, same cost `DateComparisonDialogService.getShare()` already pays for the interactive dialog's own share-lookup; no separate mitigation needed for typical viewsheet sizes).
4. **`expression`-typed condition value's auth scope** — confirmed at parity with the interactive Composer. `ConditionVocabulary.value()`'s `expression` case builds the same `ExpressionValueModel`/`ExpressionValue` that `ConditionUtil.java` (the interactive condition dialog's own conversion path) already builds when a user types a free-form JS/SQL expression — there is no separate script-execution permission gating either path today; both are gated by whatever asset write-permission already applies to writing a condition on that viewsheet. The agent path is not exposing a broader write surface than the interactive one.

larryliang-inetsoft (changes requested):

larryliang-inetsoft (changes requested):

fredaan (approved):

fredaan (approved):

PR #5041, opened 2026-09-07
feat(elastic): add Elasticsearch as tabular catalog connector; make dataset ids opaque

Nelson-cao:
## Summary

- `ElasticRestRuntime` implements `TabularCatalogProvider` via a new `ElasticCatalog`: `listDatasets` enumerates open, non-internal indexes (`_cat/indices`, sorted by name); `describeDataset` reads `_mapping`/`_count`, flattens declared objects, keeps a `nested` field as one column, excludes fields whose `_source` shape the mapping does not commit to (`alias`, `.keyword` multi-fields, geo/range/`join`/`percolator`/`completion`), and reports an unrecognized mapping type as a string — flagged in the **dataset** description (never the column's own, so it can't collide with a verbatim `meta.description` on the same field) plus a WARN log.
- All metadata reads go through a dedicated GET-only helper (`ElasticRestRuntime.getMetadata`), distinct from the query path, so a catalog read can never fall through to Elasticsearch's mapping-**update** API.
- Fixed a test-only bug: the mocked `ApplicationContext` in both test classes didn't stub `PropertiesEngine`/`XSwapper`, so any test that called `runQuery` threw a `NullPointerException` out of a static initializer (Caffeine's bean cache rejects a `null` value), permanently breaking `inetsoft.report.internal.Util` for the rest of the JVM.
- `TabularCatalogService.validateDatasetIds` no longer rejects a dataset id for containing `.`. The check was a producer-side guard against wiz's dot-splitting of a `TabularDatasetRef.id`; wiz now treats a METADATA source's id as opaque directly instead (companion PR, see below), which is the real fix. Elasticsearch's own ILM/rollover index naming convention (`logs-2026.09.04`) is exactly the case the old check blocked outright — one such index aborted an entire data source's catalog, not just that one index. `TabularDatasetRef.id`'s javadoc is rewritten to say ids are opaque and a dot is permitted, not forbidden.
- Checked all ten pre-existing `TabularCatalogProvider` implementers' `listDatasets`: none emits a dotted id today (SharePoint Online and Google Sheets both already percent-escape a dot out of their composed ids defensively; Cassandra/Aerospike/Hive/OrientDB read an unqualified name from JDBC metadata; OData/Salesforce/GA4 read names whose own grammar excludes a dot). Removing the check is not a silent behavior change for any connector but this one.

## Test plan

- `./mvnw -pl connectors/inetsoft-elastic -am -Delastic.tests=true -Dtest=ElasticCatalogTests,ElasticRestRuntimeTests test` — 20/20 pass against a live `elasticsearch:7.9.2` container.
- `./mvnw -pl connectors/inetsoft-elastic -am -Dtest=ElasticCatalogTests,ElasticRestRuntimeTests test` (no property) — both classes skip cleanly, no Docker required.
- `./mvnw -pl core -am -Dtest=TabularCatalogServiceContractValidationTest,TabularCatalogServiceTest test` — 21 + 11 pass, including the new `listTables_dottedDatasetId_isAccepted`.

## Merge order

**This PR must merge after `feat/tabular-opaque-metadata-ids` in `stylebi-wiz`.** Merging this one first would start emitting dotted METADATA ids into a wiz that still splits a source on `.`, reintroducing the exact false-duplicate/wrong-resolution bug this round is fixing on the wiz side.

### DIFF
--- a/connectors/inetsoft-elastic/src/main/java/inetsoft/uql/elasticrest/ElasticCatalog.java
+++ b/connectors/inetsoft-elastic/src/main/java/inetsoft/uql/elasticrest/ElasticCatalog.java
@@ -25,6 +25,7 @@ import inetsoft.uql.tabular.*;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.nio.charset.StandardCharsets;
 import java.util.*;
 
 /**
@@ -203,6 +204,8 @@ final class ElasticCatalog {
          throw new Exception("Elasticsearch was asked to describe a blank dataset id.");
       }
 
+      validateIndexName(datasetId);
+
       Map<?, ?> byIndex = asMap(parse(ElasticRestRuntime.getMetadata(
          ds, "/" + datasetId + "/_mapping")));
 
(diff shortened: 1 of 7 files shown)


Nelson-cao:
Round 2 (confirmation only): checked the I1 fix (commit `2a7e155d8`) by reading `ElasticCatalog.java`/`ElasticRestRuntime.java` directly.

Confirmed `validateIndexName` runs before every URL-building call in `describeDataset` (`/_mapping`, `/_count` via `documentCount`, and `searchSuffix` both in `params` and in the dataset description text), and that `listDatasets` never needs it — its ids come only from the server's own `_cat/indices` response.

Cross-checked the forbidden-character set character-by-character against the fix doc's live 7.9.2 probe transcript: matches exactly. The exact-string `.`/`..` check deliberately does not reject a legitimate leading dot or a run of three or more dots, which is correct rather than a gap — Elasticsearch's own rule forbids only those two literal strings. The 255-byte (not -character) cap is right, since the probe measured a byte limit.

Attacked two angles the fix does not explicitly address:

- **Percent-encoded input** — not a bypass. Nothing in this path percent-decodes the string, so `logs%2e%2e` reaches the server as literal text, never as a decoded `..`.
- **Raw newline / control characters** — `validateIndexName` does not reject them, but a throwaway JDK test confirmed `HttpURLConnection` itself throws at `connect()` before any bytes reach a socket, so it fails closed rather than enabling request-splitting. Worth an explicit javadoc note if we want to stop relying on unstated JDK behaviour; not sent back as a blocker.

Independently opened `SharepointOnlineCatalog.java` and confirmed the new javadoc's citation of it is accurate, not fabricated.

Confirmed, no new findings. Clean.

Nelson-cao:
The alias-accounting finding is fixed in `2c1cfe1`.

Two notes on how, since the fix is not the one suggested.

**The suggested fix would have broken something deliberate.** `unaddressable` is not a general skipped-field register — it is the roster written into the dataset *description*, naming fields that exist in the data but cannot be addressed by a single column name. A field alias is not in `_source` at all, and `fieldAliasIsNotAColumn` asserts on purpose that the description does not name one. So the error message got its own accounting instead: a separate `aliases` accumulator, used only when building the `columns.isEmpty()` exception. `datasetDescription` is unchanged.

**The illustration turned out not to be constructible, and the real case is broader.** Checked against a live 7.9.2 rather than assumed:

```
PUT /t1 {"a": {"type":"alias","path":"nonexistent"}}                    -> 400 "an alias must refer to an existing field in the mappings"
PUT /t2 {"a":{"type":"alias","path":"b"}, "b":{"type":"alias","path":"a"}} -> 400 same
```

Elasticsearch requires every alias's `path` to name an already-declared, non-alias field in the same mapping, so an index whose fields are *all* aliases cannot exist. `collect()` visits every declared field, so the backing field is always visited too, and for `columns` to stay empty that backing field must itself be unaddressable — meaning the list was never literally empty. The defect that *is* constructible is the more general one the finding is really about: an alias silently missing from the explanation whether the surrounding list is empty or merely incomplete. That is what the new fixture (`geo_point` plus two aliases pointing at it) and `noAddressableColumnsExceptionNamesTheAliasesTooNotJustTheBackingField` cover.

Also took the "no action needed" observation on the same message: it quoted `properties.size()`, a top-level count that could under- or over-state what the sentence enumerates once a declared object was in the mix. It now quotes `reasons.size()` — the number of items it actually lists.

23/23 green with `-Delastic.tests=true`; all 23 still skip cleanly without it.

On merge order: still after `feat/tabular-opaque-metadata-ids` (stylebi-wiz#2252). That PR now also carries the `dedupeAnnotationTableDocs` scoping fix, which was briefly a separate PR and was folded back in, so #2252 is the single prerequisite rather than two.

fredaan (approved):

PR #5043, opened 2026-09-07
L9 Scripting parity: fix Group A findings (G1-1, G1-2, G3-1, G3-2)

NeverStopFMH:
## Summary

Fixes the "Group A" clear-cut findings from the wiz-agent L9 (Scripting)
parity audit (see `stylebi-wiz`'s `docs/parity/L9-scripting/L9-parity-report.md`),
all live-verified against a real running instance built from this branch:

- **G1-1** — `CalcFieldService.write` never validated a calc field's new
  expression before storing it, unlike the native Formula Editor dialog's OK
  button (`ModifyCalculateFieldService.checkScriptValid`). Now runs a real
  syntax check (SQL: `XUtil.isSQLExpressionValid`; JS: `ScriptEnv.checkFunction`)
  before mutating, raising a clear error with the write left untouched on
  failure.
- **G1-2** — `WorksheetEditService.Editor.editExpression` had the identical
  gap for a worksheet expression column. Same fix shape; SQL mode delegates to
  `ExpressionDialogService.check()` (its SQL branch already works correctly),
  JS mode calls `ScriptEnv.checkFunction()` directly.
- **G3-1 / G3-2** — `get_function_signature`'s backing `ScriptApiService`
  couldn't resolve any CALC-namespace function (`day`, `eomonth`, ...) and
  leaked 11 report-only `sreeOnly` functions a viewsheet script should never
  see. `load()` now mirrors `VSScriptableService.createStaticDefinitions()`'s
  CALC-unwrap and `sreeOnly` filter.

**A real bug found only by live-testing, not unit tests**: the first attempt
at G1-1/G1-2 used `ScriptEnv.compile()` for JS validation, mirroring what
`ModifyCalculateFieldService.checkScriptValid()`/`ExpressionDialogService.check()`
already do. Live testing showed `compile()` does not actually parse anything
under the Graal engine (`GraalJavaScriptEngine.compile()` only builds a lazy
`Source` literal) — a real syntax error was silently accepted and only
surfaced later at render time. **This is a pre-existing platform gap in
`checkScriptValid()`/`check()` themselves, reproduced live in the native
Formula Editor/Edit Expression dialogs too** (confirmed: the native UI
accepts the same broken input, then `FormulaTableLens` logs a `SyntaxError`
only at data-refresh time). This PR fixes only the two wiz-agent call sites
(using `ScriptEnv.checkFunction()`, which parses eagerly via
`context.parse("js", cmd)` — the same primitive `OpenScriptController`'s own
"check script" endpoint already uses) — it does **not** touch
`ExpressionDialogService.check()`/`checkScriptValid()` itself, so the native
UI's own gap remains open as a separate, out-of-scope platform issue.

## Test plan
- [x] `CalcFieldServiceTest`, `WorksheetEditServiceMutatorsTest`,
      `WorksheetEditServiceTest`, `WorksheetAgentControllerTest`,
      `WorksheetScriptControllerTest`, `WorksheetScriptServiceTest`,
      `ScriptEditServiceTest`, `ScriptApiServiceTest` all pass (451 tests)
- [x] Two new regression tests exercise the real `ScriptEnvRepository`
      fallback engine (not a mocked `ScriptEnv`) to prove the fallback path
      itself catches a real syntax error
- [x] Live-verified both arms on a real running instance (reject broken
      input + read-back unchanged; accept valid input + read-back applied)
      for all four findings — see `stylebi-wiz`'s
      `docs/parity/L9-scripting/L9-parity-report.md` for the recorded arms

Paired plugin PR: inetsoft-technology/stylebi-wiz#2255

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/script/CalcFieldService.java
+++ b/core/src/main/java/inetsoft/web/wiz/script/CalcFieldService.java
@@ -28,6 +28,7 @@ import inetsoft.uql.viewsheet.VSAssembly;
 import inetsoft.uql.viewsheet.Viewsheet;
 import inetsoft.uql.viewsheet.internal.DataVSAssemblyInfo;
 import inetsoft.util.script.ScriptEnv;
+import inetsoft.util.script.ScriptEnvRepository;
 import inetsoft.web.wiz.pairing.PairingException;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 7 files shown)


NeverStopFMH:
Addressed in commit e9f5a59c2:

1. **Fixed** — `Editor.addExpressionColumn` had the same missing-validation gap as `editExpression`. Now calls `validateExpressionSyntax` too, with a new regression test (`addExpressionColumnRejectsAnUncompilableScriptRatherThanCreatingIt`) exercising the real fallback engine, same pattern as the existing `editExpression` test.
2. **Fixed** — `ScriptEnv` is now resolved lazily via a `Supplier<ScriptEnv>`, only inside `validateExpressionSyntax` when it's actually needed, instead of unconditionally on every `apply()` call.
3. **Left as follow-up** (per the review's own framing, "not asking for a refactor in this PR necessarily") — the `sreeOnly` deny-list duplication between `ScriptApiService` and `VSScriptableService`.

All 457 tests in the affected suites pass, including one collateral fix: `editExpressionValidatesSqlModeTextTooNotJustJavaScript`'s blanket mock stub now also matched its own `addExpressionColumn` setup call (since that op validates too), scoped the stub to the specific failing text instead.

larryliang-inetsoft (changes requested):

fredaan (approved):

PR #5044, opened 2026-09-07
feat(tabular): scaled annotation target selection -- contract, threshold, paging route

Nelson-cao:
Steps 1-3 and 5 of `docs/superpowers/specs/2026-09-07-scaled-annotation-target-selection-design.md`, on the community side. The SAP connector is a separate PR in the enterprise repo; the portal dialog, the wiz-services forwarding route and selection persistence are in stylebi-wiz.

## What this contains

**The SPI contract (§3.1 / §3.2).** Two `default` methods on `TabularCatalogProvider` — a filtered, paged `listDatasets` returning `TabularCatalogPage`, and `listRelationships(dataSource, datasetIds)` returning only edges with **both** endpoints inside the given set. Both are `default`, so all 12 production implementers are unchanged and none had to be touched. Plus two neutral records, and a javadoc addition making `TabularCatalog.datasets()`'s ordering guarantee explicit across calls — something paging silently depended on that had never been written down.

The default cursor is a decimal index by design: the opaque-cursor contract exists so an overriding connector is not bound to an offset's weakness, not to force the in-memory default to be keyset. Its residual — an insert or delete between two page calls can skip or repeat an entry — is stated in the method's own javadoc rather than engineered away.

**The threshold channel (§3.3).** `WizDatasourceEntry` carries `annotationTargetThreshold` (500, global). It has to live there rather than in a list response, because the first request is sent with `limit = threshold + 1` and must know the threshold before making it. The first page *is* the over-the-line probe, which is why no count operation was added.

**The paging route (§7 backend).** `GET /datasource/tables` gains optional `nameContains` / `limit` / `cursor`. Omitting all three is byte-identical to its previous behaviour, which matters because it has an existing caller. The presence of `limit` is what selects paged mode.

`TabularCatalogService`'s relationship-endpoint validation is **unchanged**. A paged response carries no relationships — not because the check is bypassed, but because `TabularCatalogPage` has no relationships field at all, so that path never produces an input for it.

A paging request against a JDBC source fails with a named error rather than quietly returning an unpaged listing.

## Review

The contract went through two review rounds plus two independent CI reviews. Round one found an integer overflow in the paging slice — `start + limit` wrapping negative for a caller that raises `limit` after the first page, throwing out of `subList` — and that neither documented `decodeCursor` behaviour had a test. Both fixed; the overflow regression test was confirmed red against the unfixed line before the fix was restored. Round two was clean.

The scope extension had its own review: unescaped single quotes in the SAP RFC conditions (fixed in the enterprise PR), `SAPTableRuntime` missing from the catalog-provider canary list (fixed here), and a missing empty-page test (fixed there).

## What this does not do

**JDBC paging is out of scope**, and not for lack of a UI. `MetadataApiService.getDatabaseTables` runs `buildRelationships` and `populatePrimaryKeys` once per table on every full listing, so a 600-table schema issues roughly 1200 catalog queries on a single click today — the same cost shape as SAP's per-table describe, never previously recognised as the same problem. It needs the same surgery, and judging its priority needs timing against a real database rather than H2. Recorded for a separate round.

The argument that no implementer overrides the new defaults is **constructive, not observed**: the methods are new, so nothing could have overridden them, and no connector file changed. `core`'s test classpath reaches none of the 12 implementers — connectors depend on `inetsoft-core`, never the reverse, and the enterprise connectors are a separate Maven reactor — so the canary test verifies zero of them today, and says so in its own javadoc.

Full run archive: `docs/teams/2026-09-07-tabular-catalog-paging-contract/` in stylebi-wiz.

### DIFF
--- a/core/src/main/java/inetsoft/uql/tabular/TabularCatalog.java
+++ b/core/src/main/java/inetsoft/uql/tabular/TabularCatalog.java
@@ -22,7 +22,15 @@ import java.util.List;
 /**
  * A tabular data source's whole catalog.
  *
- * @param datasets      every dataset in the source; never null, in the source's own order.
+ * @param datasets      every dataset in the source; never null, in the source's own order. This
+ *                      order MUST be stable across repeated calls against an unchanged source --
+ *                      not merely stable within one call -- so that
+ *                      {@link TabularCatalogProvider}'s default paging (and any override built the
+ *                      same way) does not skip or repeat a dataset while a caller pages through an
+ *                      unchanged catalog. A connector whose native order can vary between two calls
+ *                      (e.g. hash-based iteration) MUST impose its own stable secondary order (such
+ *                      as sorting by {@link TabularDatasetRef#id()}) before returning -- there is no
+ *                      separate ordering hook for the paged method to use instead.
  * @param relationships edges the source declares between datasets in {@code datasets}. Never null;
  *                      empty when the source declares none, or declares none this SPI can express.
  *                      Every fromDataset/toDataset MUST be an id in {@code datasets}.
--- a/core/src/main/java/inetsoft/uql/tabular/TabularCatalogPage.java
+++ b/core/src/main/java/inetsoft/uql/tabular/TabularCatalogPage.java
@@ -1,0 +1,30 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.uql.tabular;
+
+import java.util.List;
+
+/**
+ * One page of {@link TabularDatasetRef}s, in answer to
+ * {@link TabularCatalogProvider#listDatasets(TabularDataSource, TabularCatalogRequest)}.
+ *
+ * @param datasets   this page's datasets. Never null; may be empty. In a stable relative order.
+ * @param nextCursor the cursor to pass back as {@link TabularCatalogRequest#cursor()} to fetch the
+ *                   following page. {@code null} means this was the LAST page. Opaque.
+ */
+public record TabularCatalogPage(List<TabularDatasetRef> datasets, String nextCursor) {}
--- a/core/src/main/java/inetsoft/uql/tabular/TabularCatalogProvider.java
+++ b/core/src/main/java/inetsoft/uql/tabular/TabularCatalogProvider.java
@@ -17,6 +17,13 @@
  */
 package inetsoft.uql.tabular;
 
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+import java.util.stream.Collectors;
+
 /**
  * A tabular connector's description of its own catalog: which datasets a data source holds, and
  * what one dataset looks like.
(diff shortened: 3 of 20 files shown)


Nelson-cao:
## Review round 1 (P5)

Reviewed against `00-charter.md` (A1-A11 / B1-B9), `03-reconcile.md`'s rulings, and `05-verify.md`'s P4 judgment, plus a fresh adversarial pass over the actual diff.

**Result: not clean. Two important findings, no blockers, no new nits.**

### Important #1 — integer overflow in the default paged `listDatasets`

In `TabularCatalogProvider.java`, the default `listDatasets(dataSource, request)` computes:

```java
int end = Math.min(start + request.limit(), filtered.size());
```

`request.limit()` has no upper bound (only `limit > 0` is enforced). When `start > 0` and `limit` is large enough, `start + request.limit()` overflows 32-bit `int` and wraps negative, so `filtered.subList(start, end)` throws an undocumented `IndexOutOfBoundsException` instead of any behavior the method's `@throws` javadoc promises.

This is reachable with a fully compliant caller, not a crafted cursor: the contract only requires replaying a cursor with the *same* `nameContains` it was minted under — nothing pins `limit` across calls. A caller that pages a few times with `limit=3` and then asks for "everything left" with `limit=Integer.MAX_VALUE` on the next call (reusing the last valid, in-range cursor) hits this immediately.

Suggested fix: widen the intermediate computation to `long` before clamping:

```java
long endLong = Math.min((long) start + request.limit(), filtered.size());
int end = (int) endLong;
```

### Important #2 — zero test coverage for two documented cursor-error paths

`decodeCursor`'s two contractual behaviors are untested anywhere in this diff:

1. An unparseable cursor throws `IllegalArgumentException` (the method's own javadoc promises this).
2. A syntactically valid cursor that lands past the end of the (possibly re-filtered) list returns an empty, exhausted page rather than throwing (charter decision D8).

The existing paging loop helper (`collectAllPages`) structurally never drives `start >= filtered.size()` in normal use, since it stops the moment `nextCursor == null`. A wrong implementation of either behavior (e.g., silently defaulting to page 0 on a garbage cursor, or throwing instead of returning empty on a stale one) would not turn any current test red.

Suggested tests to add: one negative-cursor-string case asserting `IllegalArgumentException`, and one stale-but-syntactically-valid-cursor case asserting an empty page with `nextCursor == null`.

### What I tried and could not break

- Overload/method-reference ambiguity from the new `listDatasets` overload (no `::listDatasets` references anywhere; reflection dispatch is by exact parameter types).
- Whether the both-endpoints relationship filter could disturb `TabularCatalogService.validateRelationshipEndpoints`'s duplicate-name or `fromColumns`/`toColumns`-size checks (it's a subset operation — cannot introduce new violations).
- Cross-module classpath reachability for the canary test (re-verified: zero of 12 implementers loadable from `core`).
- Thread safety of the two new default methods (stateless, no fields, delegate to an existing method whose own thread-safety contract is unchanged).

Full review: `docs/teams/2026-09-07-tabular-catalog-paging-contract/06-review-r1.md` (stylebi-wiz repo).

Nelson-cao:
Round 2 review posting - connectivity test

Nelson-cao:
## Round 2 review — fix commit 8c3f95c3d

Scope: verify the two round-1 important findings were actually fixed, not just that a commit landed.

**Finding 1 (integer overflow in `start + request.limit()`) — fixed correctly.**
- `long end = Math.min((long) start + request.limit(), filtered.size())` cannot overflow: the widened addition tops out around 2*Integer.MAX_VALUE, far under `Long.MAX_VALUE`.
- The narrowing cast `(int) end` is provably safe, not just empirically safe: the pre-existing bounds check clamps `start` to `[0, filtered.size())` before this line runs, and `limit >= 1`, so `start < end <= filtered.size()` always holds — `end` is always representable as `int`.
- `end < filtered.size()` (long vs. int) promotes correctly and preserves the original null-cursor-on-last-page semantics.
- Walked the `start == Integer.MAX_VALUE` case: it's caught by the earlier `start >= filtered.size()` bounds check and returns an empty exhausted page before ever reaching the overflow-prone line — no path reaches it un-clamped.
- The regression test (`limitLargerThanRemainingCount_doesNotOverflowStartPlusLimit`) asserts a *successful* result (no `assertThrows`), so it fails on any exception regardless of type — it can't pass for the wrong reason if the cursor path started throwing instead of overflowing. Independently re-derived the pre-fix arithmetic (`2 + Integer.MAX_VALUE` wraps to `-2147483647`) and it matches the pasted pre-fix stack trace exactly.
- One correction worth noting for the record: the exception the overflow actually throws is `IllegalArgumentException` (from `subListRangeCheck`'s `fromIndex > toIndex` branch), not `IndexOutOfBoundsException` as round 1's prose said — the round-1 report's own pasted stack trace already showed this. Doesn't change the finding's validity or the test's validity, since the new regression test doesn't assert on exception type at all.

**Finding 2 (`decodeCursor`'s two documented branches, previously untested) — both new tests hit the intended branch.**
- `cursorNotMintedByThisImplementation_throwsIllegalArgumentException` (parameterized over `"not-a-cursor"` and `"-1"`) hits the `NumberFormatException`-catch-and-rethrow path from two different throw sites inside `decodeCursor`.
- `cursorPastTheEnd_returnsEmptyExhaustedPage_notException` hits the separate "syntactically valid but out of range" branch in `listDatasets` itself, not `decodeCursor`.
- Both would go red against the implementations named in the review brief (return 0 on unparseable, or throw on past-the-end).

**Nit ride-along (`matchesNameContains` refactor)** — verified `null`/blank/`"   "` still mean "no filter" (existing test covers all three), `null` id handling is byte-for-byte preserved, normalization now runs once per call instead of once per element, and none of the A4/B8 assertions became vacuous — they assert on filtered id sets, not on the helper's signature.

Ran the four targeted test classes with `-am` and independently reproduced the reported result:
```
Tests run: 30, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```
Also confirmed `TabularCatalogService.java` is still absent from `git diff stylebi-wiz-main-rebase..HEAD --stat` — B1 hasn't regressed.

**Verdict: clean round. No blockers, no important findings, no new nits.** Both round-1 important findings are confirmed fixed, not just superficially patched.

larryliang-inetsoft (changes requested):

fredaan (approved):

PR #5046, opened 2026-09-07
fix(76489): ConnectToClaudeComponent surfaces connection failures instead of hanging

larryliang-inetsoft:
## Summary
- `ConnectToClaudeComponent.requestCode()` only listened to `ViewsheetClientService.whenConnected()`, which emits solely on a successful connect. A socket that fails or keeps failing to (re)connect never pushes to it, so the "Generating..." spinner hung indefinitely with no error surfaced to the user.
- This is the client-side gap identified in `docs/teams/2026-09-07-bugs-76489-76490-env/bug-76489/02-refute.md` (stylebi-wiz repo) as the piece that turns any server-side connectivity blip -- nginx dual-stack burst, a stale-JWT hard-401, or anything else that breaks the underlying vs-events socket -- into an unrecoverable silent hang, regardless of which server-side cause is at fault on a given occasion.
- Fix: subscribe to the sibling `connectionError()` stream (already being pushed to, just never listened to by this component) and surface a real error message through the component's *existing* error/retry UI -- the `error` banner plus the "Connect to Claude" button, which re-enables itself once `loading` clears. No new UI was needed.
- Added an independent `timeout()` backstop (30s) on the `whenConnected()` wait itself, in case `connectionError()` never fires for some reason.
- Deliberately narrow: does **not** touch `stomp-client.ts`'s `reloadOnFailure` (still wallboard-only). A dialog-local fix is lower risk and sufficient here -- broadening full-page auto-reload to the Composer socket was considered and set aside per the investigation notes linked above.

## Root cause chain (for reviewer context)
`ViewsheetClientService.whenConnected()` returns `this.connected.asObservable()`, a `ReplaySubject` that only ever gets `.next()` on a successful connect. On any failure it instead pushes to a *different* subject, `connectionErrorSubject` (exposed as `connectionError()`), which this component never subscribed to. The underlying `StompClient.reconnect()` (`stomp-client.ts`) retries every 10s for up to 30 attempts (~5 min) before giving up silently for any non-wallboard client.

Related work in this ticket (separate repos/PRs, not part of this change):
- nginx dual-stack fix: `stylebi-wiz` repo, PR #2266
- Stale-JWT hard-401: confirmed real, explicitly NOT auto-fixed here -- flagged for StyleBI security review

Redmine #76489.

## Test plan
- [x] `cd web && npx ng test portal --include="**/connect-to-claude.component.spec.ts"` -- 44/44 passing (40 pre-existing + 4 new regression tests covering: connection error surfaces a visible message, a `null` connectionError emission is correctly ignored, the button re-enables for retry, and the independent `timeout()` backstop fires when neither stream ever emits)
- [x] `npx ng lint portal` -- 0 errors (pre-existing unrelated warnings only, none in touched files)

### DIFF
--- a/web/projects/portal/src/app/composer/gui/wiz/connect-to-claude.component.spec.ts
+++ b/web/projects/portal/src/app/composer/gui/wiz/connect-to-claude.component.spec.ts
@@ -17,7 +17,7 @@
  */
 import { NO_ERRORS_SCHEMA } from "@angular/core";
 import { ComponentFixture, TestBed, waitForAsync } from "@angular/core/testing";
-import { of } from "rxjs";
+import { NEVER, of, Subject } from "rxjs";
 import { ViewsheetClientService } from "../../../common/viewsheet-client";
 import { ConnectToClaudeComponent } from "./connect-to-claude.component";
 
@@ -26,6 +26,7 @@ describe("ConnectToClaudeComponent", () => {
    let component: ConnectToClaudeComponent;
    let mockSocketConnection: any;
    let mockStompConnection: { subscribe: ReturnType<typeof vi.fn>; send: ReturnType<typeof vi.fn> };
+   let mockConnectionErrorSubject: Subject<string>;
 
    beforeEach(waitForAsync(() => {
       mockStompConnection = {
@@ -33,8 +34,11 @@ describe("ConnectToClaudeComponent", () => {
          send: vi.fn()
       };
 
+      mockConnectionErrorSubject = new Subject<string>();
+
       mockSocketConnection = {
-         whenConnected: vi.fn(() => of(mockStompConnection))
+         whenConnected: vi.fn(() => of(mockStompConnection)),
+         connectionError: vi.fn(() => mockConnectionErrorSubject.asObservable())
       };
 
       // Default: subscribe returns a Subscription-like object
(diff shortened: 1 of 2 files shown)


fredaan (approved):
## Summary

Fixes bug 76489: `ConnectToClaudeComponent.requestCode()` only listened to `whenConnected()`, so a socket that failed/kept failing to reconnect left the "Generating..." spinner hanging forever with no error. This PR subscribes to the existing `connectionError()` stream and adds an independent 30s `timeout()` backstop, surfacing failures through the component's existing error banner + retry button.

## Correctness

- The core fix is sound. `ViewsheetClientService.connectionError()` is backed by a `ReplaySubject<string>(1)` that is pushed `null` on every successful connect (`viewsheet-client.service.ts:100`) right before `connected.next(...)`, so a late subscriber always gets the current connection status, not a stale one. The `filter(message => message != null)` correctly ignores the "healthy" marker.
- This mirrors the existing `connectionError()` consumption pattern already used elsewhere (e.g. `viewsheet-pane.component.ts:535-539` uses the equivalent `filter(err => !!err)`), so the approach is consistent with established conventions rather than inventing a new one.
- The three commits show real, well-targeted iteration: round 1 added the fix, round 2 fixed a subscription leak (`connectErrorSubscription` was nulled without `unsubscribe()`), round 3 closed a race where a `connectionError()` that fires first didn't cancel the still-pending `whenConnected()`/`timeout()` wait, which could otherwise let a self-healing socket mint a code next to a stale error message. Good defense-in-depth thinking, and each fix is backed by a regression test.

**Minor (non-blocking) nit** — in `requestCode()`:
```ts
this.connectWaitSubscription = this.socketConnection.whenConnected().pipe(
   take(1), timeout(MINT_CONNECT_TIMEOUT_MS)
).subscribe({
   next: (conn) => {
      this.connectWaitSubscription = null;   // <-- see below
      ...
```
`whenConnected()` returns `connected.asObservable()`, also a `ReplaySubject(1)`. In the common case where the socket is already connected when `requestCode()` runs (the typical state once the Composer is open), `.subscribe()` emits `next` **synchronously**, before the outer assignment on the left-hand side executes. That means `this.connectWaitSubscription = null` inside the callback gets immediately clobbered by the outer `this.connectWaitSubscription = <subscription>` once `.subscribe()` returns (the returned subscription is already closed due to `take(1)`, but non-null). Every place that reads `connectWaitSubscription` afterward only uses it to decide whether to call `.unsubscribe()`, which is idempotent, so this doesn't cause a user-visible bug today — but it's fragile: if this field is ever used for a real null-check later, it will misbehave silently. Worth a follow-up to null it out after the `subscribe()` call returns instead of inside the synchronous callback (or capture the subscription in a local var first).

## Test coverage

The 6 new tests in the `spec.ts` are genuine regression tests, not padding — they cover: error surfacing, ignoring the `null` "healthy" marker, button re-enable for retry, the connectionError-subscription leak fix, the independent timeout backstop, the two-way race-cancellation fix, and teardown on destroy. This is good coverage for the bug and its two follow-up fixes.

## Regression / backward-compatibility risk

Low. The change is additive (two new private subscriptions + one new operator import) and scoped to a single component; no public API, template, or shared service surface changes. `stomp-client.ts`'s `reloadOnFailure` is deliberately untouched, keeping the blast radius to this dialog only, as called out in the PR description.

## CI status

- `Semgrep`, `Semgrep OSS`, `license/cla`: passing.
- `claude-review`: failed, but with an internal bot execution error (`Claude execution failed: result is_error:true`) before it produced any actual review output — not a finding about this diff. Per review process this check is superseded by this manual review anyway.
- `build` (Validation Build): failed, reproduced on a rerun, but in **`inetsoft.sree.security.PermissionHierarchyTest`** — `IllegalArgumentException: argument "in" is null` inside `SecurityChain.loadConfiguration` → Jackson `ObjectMapper.readTree`. This is a Java backend security test with no possible relationship to this PR's diff (2 files, both Angular/TypeScript under `web/projects/portal/.../wiz/`). The identical failure was also observed on an unrelated, unmerged PR (`feature/76493-task-audit-token`) against the same base branch, confirming this is a pre-existing base-branch issue (consistent with this repo's documented "trunk-based development: some tests on main may be unstable" note), not something introduced or fixable by this change.

## Merge state

`MERGEABLE`, no conflicts.

## Verdict

No blocking issues found in the diff itself. The one nit above is cosmetic/defensive and doesn't need to hold up this PR. Approving — the failing `build` check is unrelated pre-existing flakiness in a Java security test, and `claude-review` failed on bot infrastructure, not on this code.

fredaan (approved):

PR #5047, opened 2026-09-07
fix(WSQ-002): editSqlQuery now reports undeclaredVariables

fredaan:
## Summary
- `editSqlQuery` ended its `applyOnRuntime` lambda with a bare `return null;` and never called `detectUndeclaredVariables` the way `addSqlQuery` does — an unresolved `$(name)` placeholder introduced by an edit was silently accepted with no signal.
- Since `edit_sql_query` was already dispatched to `editSqlQuery` via its own early-return branch in `editOp`, outside the generic ~50-op Editor/dispatch pipeline, this gives it its own bespoke `PUT .../sql-query` endpoint (mirroring `addSqlQuery`'s shape: own typed `SqlQueryResponse`, own `@PutMapping`) rather than changing the shared void `/edit` architecture. No risk to the other ops still served by `/edit`.

Root-caused and fix direction confirmed via `docs/teams/2026-09-07-bugs-76500-wsq-pcb/bug-WSQ-002/01-diagnosis.md` and `02-refute.md` in the companion `stylebi-wiz` repo (Redmine #76500).

## Test plan
- [x] `WorksheetAgentControllerTest` — new `editSqlQueryEndpointReturnsUndeclaredVariables` test (mocks `WorksheetEventUtil.refreshVariables` to return an undeclared-variable command, asserts the new bespoke endpoint's response carries it)
- [x] Full `WorksheetAgentControllerTest` suite: 133 tests passed (`./mvnw -pl community/core test -Dtest=WorksheetAgentControllerTest`)
- [ ] Live StyleBI verification — not done this session (stack was down); pending separately

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -395,7 +395,7 @@ public class WorksheetAgentController {
 
       // edit_sql_query needs RuntimeWorksheet for SQL parsing and column re-init.
       if("edit_sql_query".equals(req.op())) {
-         editSqlQuery(sessionToken, req, user);
+         editSqlQuery(sessionToken, req.table(), req.expression(), user);
          return;
       }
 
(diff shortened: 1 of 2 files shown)


fredaan (commented):
Reviewed (adversarial pass per team workflow, comparing diff against 01-diagnosis.md / 02-refute.md's confirmed narrow bespoke-route fix). No blocker or important findings.

Verified:
- Matches the refuter's recommended shape exactly: `editSqlQuery`'s private worker now returns `SqlQueryResponse` (calling `detectUndeclaredVariables`, mirroring `addSqlQuery`), and a new bespoke `@PutMapping(.../sql-query)` endpoint exposes it — not a broad `/edit` architecture change.
- The shared `/edit`/`editOp` dispatcher for the other ~50 ops is untouched: `editOp`'s `edit_sql_query` branch just changed its call arguments (`req.table()`/`req.expression()` instead of `req`) and still discards the return value, still `return;`s immediately.
- The private `editSqlQuery` method is `private`, so the only two call sites are the ones in this diff (compiler-enforced — no other caller could be broken by the signature change).
- New test (`editSqlQueryEndpointReturnsUndeclaredVariables`) exercises the actual new endpoint end-to-end with a mocked undeclared-variable command and asserts the response carries it through — not a trivial assertion.
- Does not touch any prompt/LLM-facing template — this is REST controller code only.

Clean review, no changes requested.

PR #5048, opened 2026-09-07
fix(WSQ-001): surface referenced variables in read_worksheet_model

fredaan:
## Summary
- `read_worksheet_model`'s `TableModel` had no field showing which worksheet variables a table's SQL/conditions reference, so an agent that already declared and set a variable via `add_variable`/`set_variable_values` had no way to confirm a SQL-bound table actually picked it up, short of re-submitting the SQL text or watching for a behavior change.
- Adds `TableModel.referencedVariables` (a `List<String>`), populated in `WorksheetReadService.readTable` via `TableAssembly.getAllVariables()` -- the same mechanism `SQLBoundTableAssembly` already uses (via `XDataService.getQueryParameters`) to resolve a SQL-bound table's actual name-placeholder tokens for the native Composer SQL Query dialog, not a regex over the raw SQL string.
- Intentionally reports every referenced variable, declared or undeclared -- broader than `detectUndeclaredVariables`, which filters out anything already present in the live `VariableTable` and would report nothing for this bug's own repro (variable declared and set before `read_worksheet_model` is called).
- Non-SQL-bound table types are not guaranteed to report an empty list: `AbstractTableAssembly.getAllVariables()` resolves whatever their own filter/ranking/TopN conditions reference, recursively aggregated across upstream sources for a join/concat/mirror table. This is documented on the new field and pinned down by a dedicated regression test rather than assumed.

Root-caused and fix direction confirmed via `docs/teams/2026-09-07-bugs-76500-wsq-pcb/bug-WSQ-001/01-diagnosis.md` and `02-refute.md` in the companion `stylebi-wiz` repo (Redmine #76500).

## Test plan
- [x] `WorksheetReadServiceTest` — 5 new cases: a SQL-bound table's placeholder resolved via mocked `XDataService.getQueryParameters`; the same variable still reported when it is already declared as a worksheet variable assembly (proving this isn't reusing `detectUndeclaredVariables`'s undeclared-only filter); no placeholder tokens reports an empty list, not null; a non-SQL-bound (`EmbeddedTableAssembly`) table with its own `$(name)`-valued filter condition reports it too
- [x] Full `WorksheetReadServiceTest` suite: 38 tests passed (`./mvnw.cmd -pl community/core test -Dtest=WorksheetReadServiceTest`)
- [x] Full `WorksheetAgentControllerTest` suite (exercises the `model` endpoint end-to-end): 132 tests passed, no regressions (`./mvnw.cmd -pl community/core test -Dtest=WorksheetAgentControllerTest`)
- [ ] Live StyleBI verification — not done this session (stack was down, handled separately); pending

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
@@ -133,7 +133,36 @@ public class WorksheetReadService {
          // Note this is the effective limit, capped by query.runtime.maxrow; see TableModel.
          maxRows <= 0 ? null : maxRows,
          t.isDistinct(), t.isSQLMergeable(), t.isVisibleTable(), tableMode(t),
-         offset == null ? null : offset.x, offset == null ? null : offset.y);
+         offset == null ? null : offset.x, offset == null ? null : offset.y,
+         readReferencedVariables(t));
+   }
+
+   /**
+    * The worksheet variables this table's SQL or conditions reference, declared and undeclared
+    * alike. {@link TableAssembly#getAllVariables()} is overridden by {@link SQLBoundTableAssembly}
+    * to resolve a SQL-bound table's actual name-placeholder tokens (via
+    * {@code XDataService.getQueryParameters} plus its WHERE/HAVING scan); every other table type
+    * falls back to {@link AbstractTableAssembly}'s condition-derived variables, aggregated
+    * recursively across upstream sources for a join, concat or mirror table -- neither is a regex
+    * over the SQL text, so this can never disagree with what {@code add_sql_query}'s own
+    * undeclared-variable detection reports for the same query.
+    */
+   private List<String> readReferencedVariables(TableAssembly t) {
+      UserVariable[] vars = t.getAllVariables();
+
+      if(vars == null || vars.length == 0) {
+         return Collections.emptyList();
+      }
+
+      LinkedHashSet<String> names = new LinkedHashSet<>();
+
+      for(UserVariable var : vars) {
+         if(var != null && var.getName() != null) {
+            names.add(var.getName());
+         }
+      }
+
+      return new ArrayList<>(names);
    }
 
    // -------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


fredaan (commented):
## Review: WSQ-001 (referencedVariables in read_worksheet_model / TableModel)

Reviewed against `docs/teams/2026-09-07-bugs-76500-wsq-pcb/bug-WSQ-001/01-diagnosis.md` (revision round 1) and `02-refute.md` (recheck round 2, verdict SURVIVES).

**Mechanism matches the final spec exactly**: `WorksheetReadService.readTable` populates the new `TableModel.referencedVariables` field via `TableAssembly.getAllVariables()` — not `detectUndeclaredVariables` (correctly avoided; that mechanism filters out anything already present in the live `VariableTable` and would report nothing for this bug's own repro, since `set_variable_values` writes into that same table before `read_worksheet_model` runs).

**Refute's correction respected**: the new test `nonSqlBoundTableReportsVariableReferencedByItsOwnFilterCondition` pins down that a non-SQL-bound table (`EmbeddedTableAssembly`) with its own `$(name)`-valued filter condition correctly reports a non-empty `referencedVariables` — nothing in the code or tests assumes non-SQL-bound tables are always empty.

**Constructor/record ordering**: `readReferencedVariables(t)` is appended as the new record's last field and passed as the last constructor argument in `readTable` — consistent, verified both sides.

**Test coverage matches the brief**: declared variable (still listed, proving it's not the undeclared-only filter), undeclared variable, no-token case (empty list, not null), and the non-SQL-bound condition-derived case — all four present.

**No prompt-file changes.** Diff is clean and scoped to this bug only (`WorksheetReadService.java`, `WorksheetModel.java`, `WorksheetReadServiceTest.java`).

Verified test helper APIs used in the new tests (`WorksheetMutationSupport.addFilter`, `TestWorksheets.tableWithColumns`, `TableAssembly.getAllVariables()`) exist in the current source with matching signatures.

No blocker or important findings. LGTM.

fredaan:
Rebased onto `stylebi-wiz-main-rebase` (which had moved ~15 commits ahead, including #5047's merge) to resolve the merge conflict. Only conflict was in `WorksheetReadServiceTest.java` — both sides had appended new tests at the end of the file with no logical overlap; resolved by keeping both sets in sequence. Re-ran `WorksheetReadServiceTest` after rebase: 40/40 pass. Force-pushed.

PR #5050, opened 2026-09-07
fix(WSQ-006): differentiate parse-failure vs unresolved-column messages in add/edit_sql_query

fredaan:
## Summary

Redmine #76500 (consolidated), sub-item WSQ-006. `addSqlQuery`/`editSqlQuery` threw the same generic message ("SQL could not be parsed or no columns detected -- check syntax and table references.") whether the SQL was genuinely malformed (grammar failure) or syntactically valid but referencing a table/column that couldn't be resolved against the datasource. Both signals needed to tell these apart -- `sql.getParseResult()` and `query`/`scratchQuery.getLastQueryError()` -- were already computed and in local scope at the throw site, but never read.

## Fix

At both throw sites (`addSqlQuery`, `editSqlQuery`), read the two signals before falling back to the generic message:
1. `sql.getParseResult() == UniformSQL.PARSE_FAILED` -> a syntax-specific message.
2. Otherwise, a non-null `getLastQueryError()` with a message -> a resolution-specific message including the driver's own error text.
3. Otherwise -> the original generic message (covers the "driver call succeeded but returned nothing" case).

## Review history and an open question this PR does not resolve

This fix went through two rounds of adversarial review (diagnosis + refute, `stylebi-wiz` repo, `docs/teams/2026-09-07-bugs-76500-wsq-pcb/bug-WSQ-006/`). Both rounds unanimously agree the fix code itself is correct and sound. What was contested across both rounds was narrower: which hand-crafted SQL string's real ANTLR2 grammar parse would actually exercise the fallback/`lastQueryError` branch for a live repro. The refuter traced the code precisely and found that **no bad-column-name-only repro (in an ON clause or in the SELECT list) reliably reaches the fallback branch** -- reaching it requires either a genuine grammar failure or a bad *table* reference, not a bad column name against an otherwise-valid table. This means the bug's own original historical repro (WSQ-006) most likely hit this guard for a reason not fully traced by either round (see `01-diagnosis.md`'s "wrinkle" section and `02-refute.md`'s two recheck rounds for the full trace).

This is a live-verification-pending question, not a defect in the fix: the two signals (`parseResult`, `lastQueryError`) are real, in-scope, and were being discarded regardless of which path any particular SQL shape takes. Per the lead's decision, this PR's regression tests mock `parseResult`/`lastQueryError` directly (via `mockConstruction<UniformSQL>` and a `queryManagerService` stub that sets `lastQueryError` on the real query instance) rather than relying on a specific SQL string to naturally trigger a given branch, sidestepping that open question entirely.

Full context: `docs/teams/2026-09-07-bugs-76500-wsq-pcb/bug-WSQ-006/01-diagnosis.md`, `02-refute.md`, `03-fix.md` in the `stylebi-wiz` repo.

## Test plan

- [x] `WorksheetAgentControllerTest` -- 6 new tests (3 per throw site): `PARSE_FAILED` -> syntax message, `lastQueryError` set -> resolution message including driver text, neither signal set -> generic fallback message.
- [x] Full `WorksheetAgentControllerTest` suite green: `./mvnw.cmd -pl community/core test -Dtest=WorksheetAgentControllerTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false` -- 138 tests, 0 failures, 0 errors.
- [ ] Live StyleBI verification -- not done, out of scope per task instructions (handled separately).

Not merging -- awaiting review.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -2651,6 +2651,19 @@ public class WorksheetAgentController {
             scratchQuery, new VariableTable(), sqlTable, metaSession, null);
 
          if(columns == null || columns.getAttributeCount() == 0) {
+            if(sql.getParseResult() == UniformSQL.PARSE_FAILED) {
+               throw new PairingException(
+                  "SQL could not be parsed — check syntax (punctuation, keywords, clause structure).");
+            }
+
+            Exception driverError = scratchQuery.getLastQueryError();
+
+            if(driverError != null && driverError.getMessage() != null) {
+               throw new PairingException(
+                  "SQL is syntactically valid but could not be resolved against the datasource: "
+                  + driverError.getMessage() + " — check table/column references.");
+            }
+
             throw new PairingException(
                "SQL could not be parsed or no columns detected — check syntax and table references.");
          }
@@ -3539,6 +3552,19 @@ public class WorksheetAgentController {
             query, new VariableTable(), assembly, metaSession, null);
 
          if(columns == null || columns.getAttributeCount() == 0) {
+            if(sql.getParseResult() == UniformSQL.PARSE_FAILED) {
+               throw new PairingException(
+                  "SQL could not be parsed — check syntax (punctuation, keywords, clause structure).");
+            }
+
+            Exception driverError = query.getLastQueryError();
+
+            if(driverError != null && driverError.getMessage() != null) {
+               throw new PairingException(
+                  "SQL is syntactically valid but could not be resolved against the datasource: "
+                  + driverError.getMessage() + " — check table/column references.");
+            }
+
             throw new PairingException(
                "SQL could not be parsed or no columns detected — check syntax and table references.");
          }
--- a/core/src/test/java/inetsoft/web/wiz/worksheet/WorksheetAgentControllerTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/worksheet/WorksheetAgentControllerTest.java
@@ -47,6 +47,8 @@ import inetsoft.uql.erm.XEntity;
 import inetsoft.uql.erm.XLogicalModel;
 import inetsoft.uql.jdbc.JDBCDataSource;
 import inetsoft.uql.jdbc.JDBCQuery;
+import inetsoft.uql.jdbc.JDBCSelection;
+import inetsoft.uql.jdbc.UniformSQL;
 import inetsoft.uql.jdbc.util.SQLTypes;
 import inetsoft.web.wiz.model.DatabaseTableMeta;
 import inetsoft.uql.schema.UserVariable;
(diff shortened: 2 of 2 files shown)


fredaan (commented):
## Review — WSQ-006 (PR #5050)

**Diff vs. confirmed root cause**: Verified directly against the live source in `E:\wt\76500-006-parseerr` (`WorksheetAgentController.java:2653-2669` for `editSqlQuery`, `:3554-3567` for `addSqlQuery`). Both throw sites implement the exact three-way branch from `01-diagnosis.md`'s falsifiable claim: `sql.getParseResult() == UniformSQL.PARSE_FAILED` → syntax message, else `driverError != null && driverError.getMessage() != null` (`scratchQuery`/`query` respectively — correct per-method local var, confirmed by reading both method bodies) → resolution message including the driver text, else the original generic fallback. `driverError.getMessage()` is null-checked before use, so no NPE / no `": null"` wording risk. Message wording matches the diagnosis's proposed text (using the codebase's existing em-dash style rather than the diagnosis markdown's literal `--`, consistent with the unchanged generic message already using an em-dash — a sensible, non-issue implementation choice).

**Test quality (the point you flagged for scrutiny)**: The `mockConstruction<UniformSQL>` + real-`JDBCQuery`-instance approach is sound, not just compile-and-pass theater. Traced it end to end:
- `sql` in both production methods is a local var from a real `new UniformSQL()` call (confirmed at `WorksheetAgentController.java:2610` and `:3504`) — `mockConstruction` correctly intercepts exactly that call, so `sql.getParseResult()` in production code genuinely reads the test's stubbed value, not a bypassed path.
- For `editSqlQuery`'s test, `sql` is only freshly constructed when `scratchQuery.getSQLDefinition()` is null — the test's `JDBCQuery()` fixture has no SQLDefinition set, so the `new UniformSQL()` branch is correctly taken (verified this isn't a false setup that would skip the mock).
- `getLastQueryError()` is exercised via a real (non-mocked) `JDBCQuery`/clone instance, with the mocked `queryManagerService.getColumnSelection` answer calling `q.setLastQueryError(...)` on the actual argument passed in — this mirrors the real `QueryManagerService.getColumnSelection`'s side effect (`QueryManagerService.java:1955`) precisely, and the controller's subsequent `query.getLastQueryError()` read genuinely observes it.
- Since `queryManagerService` is already mocked in every other pre-existing test in this file (132 pre-existing tests), this is consistent with the file's existing scope: these are controller-unit tests of the throw-site branching logic, not integration tests of which real SQL string reaches which branch. That narrower scope is honestly disclosed rather than implied to be more than it is.

**Open question disclosure**: Confirmed the PR description explicitly states the fallback-branch-reachability question is "live-verification-pending" and not resolved, with an accurate summary of both refute rounds (including round 2's correction that no bad-column-only repro reaches the fallback branch). Not presented as fully settled.

**Scope**: Only `WorksheetAgentController.java` (prod) and `WorksheetAgentControllerTest.java` (test) touched, additions-only. No prompt-file changes.

No blocker or important findings. Clean.

PR #5052, opened 2026-09-07
fix(wiz): get_table_details throws on non-catalog table names instead of silently returning empty columns (#76500 WSQ-003)

fredaan:
## Summary
- Redmine #76500 WSQ-003: `get_table_details` returned HTTP 200 with `columns: []` for a name that isn't a real datasource table (e.g. a worksheet-internal assembly created by `add_table`/`add_sql_query`), instead of erroring.
- Root cause: `getTableColumns` (`JDBCHandler.java`) always attaches an empty "Result" child node before it knows whether any columns exist, so `MetadataApiService.getTableDetails`'s only existing not-found guard (`tableNode == null`) never fires for this trigger.
- Fix: add an AND-gated check in `getTableDetails` -- throw only when the column fetch found nothing **and** `DefaultMetaDataProvider.getTable(catalog, schema, tableName, false)` also can't find the name in the JDBC catalog listing. This keeps a real table with genuinely zero columns (e.g. a permission-filtered view) succeeding as long as `getTable` resolves it, while correctly rejecting names that aren't real datasource tables at all.

## Test plan
- [x] Added `MetadataApiServiceGetTableDetailsTest` with 3 cases: real table w/ columns succeeds, worksheet-internal name throws with a field-named message, real table w/ zero columns but catalog match does not throw (proves the AND-gate, not a standalone check).
- [x] `./mvnw -pl core -Dtest=MetadataApiServiceGetTableDetailsTest test` -- 3/3 pass.
- [x] Confirmed the bug-reproducing case (`worksheetInternalNameWithNoColumnsAndNoCatalogMatchThrows`) fails without the fix ("Expected Exception to be thrown, but nothing was thrown") and passes with it.
- [ ] Live StyleBI verification pending (not this PR's scope -- being handled separately per the batch coordination).

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
@@ -1937,7 +1937,9 @@ public class MetadataApiService {
       }
 
       if(tableNode.getChildCount() == 0 &&
-         metaDataProvider.getTable(catalog, schema, tableName, false) == null)
+         metaDataProvider.getTable(Tool.isEmptyString(catalog) ? null : catalog,
+                                    Tool.isEmptyString(schema) ? null : schema,
+                                    node.getName(), false) == null)
       {
          throw new Exception(Tool.buildString(
             "Table '", tableName, "' was not found as a datasource table in '", dsName, "'. ",
--- a/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
@@ -1936,6 +1936,16 @@ public class MetadataApiService {
             Tool.buildString("Table ", tableName, " not found in data source ", dsName));
       }
 
+      if(tableNode.getChildCount() == 0 &&
+         metaDataProvider.getTable(catalog, schema, tableName, false) == null)
+      {
+         throw new Exception(Tool.buildString(
+            "Table '", tableName, "' was not found as a datasource table in '", dsName, "'. ",
+            "get_table_details only resolves real datasource tables from the JDBC catalog -- ",
+            "it cannot look up worksheet-internal assemblies/tables (e.g. ones created by ",
+            "add_table or add_sql_query). Use read_worksheet_model to inspect those instead."));
+      }
+
       DatabaseTableMeta meta = new DatabaseTableMeta();
       meta.setName(tableName);
       meta.setCatalog(Tool.isEmptyString(catalog) ? null : catalog);
--- a/core/src/test/java/inetsoft/web/wiz/service/MetadataApiServiceGetTableDetailsTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/service/MetadataApiServiceGetTableDetailsTest.java
@@ -160,4 +160,66 @@ class MetadataApiServiceGetTableDetailsTest {
 
       assertTrue(meta.getColumns().isEmpty());
    }
+
+   /**
+    * GitHub code review on stylebi#5052 caught this: {@code catalog}/{@code schema} being an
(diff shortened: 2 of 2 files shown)


fredaan (commented):
## Review: WSQ-003 (`get_table_details` silent empty-columns fix)

Reviewed against `01-diagnosis.md` (incl. Revision round 1) and `02-refute.md` (incl. Recheck round 2, verdict SURVIVES).

**Diff matches the confirmed spec exactly.** The AND-gate is implemented verbatim as specified:
```java
if(tableNode.getChildCount() == 0 &&
   metaDataProvider.getTable(catalog, schema, tableName, false) == null)
```
— not the weaker standalone `getTable(...) == null` form that was refuted in round 1. Placement is correct: inserted after the existing `tableNode == null` guard (kept unchanged, lines 1934-1937) and before `DatabaseTableMeta meta = new DatabaseTableMeta();`. No special-case `if` replacing the underlying rule — this is a genuine second existence check, not a symptom patch.

**Regression tests**: all 3 cases from the diagnosis's revision landed in `MetadataApiServiceGetTableDetailsTest.java` and are meaningful, not just happy-path — including the AND-discriminating case 3 (`realTableWithZeroColumnsButCatalogMatchDoesNotThrow`), which is the one that would fail if a fixer had accidentally written OR instead of AND. CI confirms all 3 pass (`Tests run: 3, Failures: 0, Errors: 0`). The claimed git-stash before/after verification (test 2 fails without the fix with the exact original symptom, passes with it) is credible given the test's structure and the PR body's stated command.

**Deliberately-skipped case-insensitivity hardening**: reasonable engineering call. `findTable`/`getTable` is a shared, deeply-used private method with 5+ overloads; hardening it for one caller's edge case is out of proportion to a narrow, compound (case-mismatch + genuinely-zero-columns) residual risk that both refute rounds agreed is non-blocking and no worse than the original standalone-check proposal would have been.

**Code quality**: error message names both the table and the datasource, and points at the correct alternative tool (`read_worksheet_model`), matching CLAUDE.md's fail-loud-not-silent guidance. No CLAUDE.md violations — no unused code, no scope creep, no comments beyond what's needed.

**Scope check**: confirmed via `gh pr diff --name-only` — only `MetadataApiService.java` and the new test file changed. No prompt-related files touched.

**CI note (not a defect in this diff)**: the `build` check is currently failing, but the failure is `PermissionHierarchyTest.setUp` — a `Jackson ObjectMapper.readTree(null)` NPE inside `SecurityChain.loadConfiguration`, unrelated to table-details/metadata-service code. This PR's own new test (`MetadataApiServiceGetTableDetailsTest`) ran in the same CI run and passed 3/3. Flagging so whoever merges knows the red CI isn't caused by this change — but this needs resolving (or the run re-triggered against a green base) before merge either way.

**Verdict: no blocker or important findings against the diff itself.** Clean, matches the AND-gated spec precisely, tests are genuine regression tests. Only actionable item is the unrelated CI failure noted above, which blocks a green merge but is not this PR's defect.

fredaan:
Both gaps from the automated review are fixed (commit 9ddbfb1e5, force-pushed after rebasing onto the updated base branch):

1. `catalog`/`schema` of `""` are now normalized to `null` before the `getTable` call, matching the `getTableMetaData` precedent in this same file (`Tool.isEmptyString(catalog) ? null : catalog`).
2. The `getTable` lookup now reuses `node.getName()` (the already-stripped, unqualified name `getQualifiedTableNode` produces a few lines above, for the column-fetch lookup) instead of the raw, possibly-still-qualified `tableName` parameter.

Added two new regression tests (`emptyStringCatalogAndSchemaAreNormalizedBeforeGetTableLookup`, `qualifiedTableNameIsStrippedBeforeGetTableLookup`) covering exactly the two false-positive scenarios you flagged — both confirmed failing against the prior code (via `git stash`) before confirming the fix. Full suite: 5/5 pass.

PR #5055, opened 2026-09-07
Bug #76497: fix null DataRef in Filter wizard recommendation

larryliang-inetsoft:
## Summary
- `VSFilterRecommendationFactory.createDataRefs()` built one `DataRef` slot per X/Y field currently bound on the temp chart, positionally matched against the selected tree entries. When a chart already had 2+ bound fields (e.g. an X dimension and a Y measure — what "bind a measure column to a chart" sets up) but only one entry was selected in the Filter step, unmatched slots came back `null`, and `refs[0]` could be null.
- `addFilterVSAssembly()` dereferences `refs[0]` unconditionally for Range Slider (`updateTimeSliderBinding` → `ref.getRefType()`) and Selection List (`getTitle` → `ref.getRefType()`) — an uncaught NPE before any assembly was added or command sent. It's silently swallowed by `WizardEventAspect.handleWizardExceptions()` in metadata mode, leaving the wizard preview canvas blank while the toolbar still shows the type as selected.
- Fix: build the `DataRef[]` directly from `wizardData.getSelectedEntries()` instead of iterating the chart's bound X/Y fields, so there's exactly one non-null ref per selection. This also fixes the same unguarded-null pattern in the multi-entry Selection Tree path (`getFilterDataRefs()`).

Fixes Bug #76497.

## Test plan
- [x] Added `VSFilterRecommendationFactoryTest` with two cases: a chart with X+Y bound and one selected entry (repro shape), and a multi-entry Selection Tree case. Both assert a fully non-null `DataRef[]` of the expected length.
- [x] Confirmed both new tests fail against the pre-fix code (temporarily stashed the fix) with the exact documented symptom (`refs` containing nulls / wrong length), then pass with the fix restored.
- [x] `./mvnw.cmd -pl core -Dtest=VSFilterRecommendationFactoryTest test` → `Tests run: 2, Failures: 0, Errors: 0`.
- [x] Full `core` module build (`-pl core -am -DskipTests install`) compiles cleanly.
- [ ] Live-browser confirmation of the wizard preview canvas rendering (Portal and Composer entry paths) — not done in this pass, flagged for follow-up.

### DIFF
--- a/core/src/main/java/inetsoft/web/vswizard/recommender/object/VSFilterRecommendationFactory.java
+++ b/core/src/main/java/inetsoft/web/vswizard/recommender/object/VSFilterRecommendationFactory.java
@@ -20,13 +20,9 @@ package inetsoft.web.vswizard.recommender.object;
 import inetsoft.uql.asset.AssetEntry;
 import inetsoft.uql.erm.DataRef;
 import inetsoft.uql.viewsheet.ChartVSAssembly;
-import inetsoft.uql.viewsheet.VSAggregateRef;
-import inetsoft.uql.viewsheet.graph.ChartRef;
-import inetsoft.uql.viewsheet.graph.VSChartInfo;
 import inetsoft.web.vswizard.model.VSWizardData;
 import inetsoft.web.vswizard.model.recommender.*;
 import inetsoft.web.vswizard.recommender.WizardRecommenderUtil;
-import org.apache.commons.lang3.ArrayUtils;
 import org.springframework.stereotype.Component;
 
 import java.security.Principal;
@@ -87,36 +83,19 @@ public class VSFilterRecommendationFactory
          return null;
       }
 
-      VSChartInfo tempInfo = temp.getVSChartInfo();
       AssetEntry[] entries = wizardData.getSelectedEntries();
-      ChartRef[] fields =
-         (ChartRef[]) ArrayUtils.addAll(tempInfo.getXFields(), tempInfo.getYFields());
 
       if(entries == null || entries.length < 1) {
          return null;
       }
 
       List<DataRef> refs = new ArrayList<>();
-      Arrays.asList(fields).forEach(field ->
-         refs.add(WizardRecommenderUtil.createColumnRef(getField(field, entries))));
-      return refs.toArray(new DataRef[0]);
-   }
 
-   private static AssetEntry getField(DataRef field, AssetEntry[] entries) {
       for(AssetEntry entry : entries) {
-         String entryName = WizardRecommenderUtil.getFieldName(entry);
-         String vsName = null;
-
-         if(field instanceof VSAggregateRef) {
-            vsName = ((VSAggregateRef) field).getVSName();
-         }
-
-         if(entryName != null && (entryName.equals(field.getName()) || entryName.equals(vsName))) {
-            return entry;
-         }
+         refs.add(WizardRecommenderUtil.createColumnRef(entry));
       }
 
-      return null;
+      return refs.toArray(new DataRef[0]);
    }
 
    private static final List<VSSubType> NUMERIC_RECOMMENDATION = Arrays.asList(
--- a/core/src/test/java/inetsoft/web/vswizard/recommender/object/VSFilterRecommendationFactoryTest.java
+++ b/core/src/test/java/inetsoft/web/vswizard/recommender/object/VSFilterRecommendationFactoryTest.java
@@ -1,0 +1,128 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.web.vswizard.recommender.object;
+
+import inetsoft.uql.asset.AssetEntry;
+import inetsoft.uql.asset.AssetRepository;
+import inetsoft.uql.erm.DataRef;
+import inetsoft.uql.schema.XSchema;
+import inetsoft.uql.viewsheet.ChartVSAssembly;
+import inetsoft.uql.viewsheet.graph.VSChartAggregateRef;
+import inetsoft.uql.viewsheet.graph.VSChartDimensionRef;
+import inetsoft.uql.viewsheet.graph.VSChartInfo;
+import inetsoft.web.vswizard.model.VSWizardData;
+import inetsoft.web.vswizard.model.recommender.VSFilterRecommendation;
+import inetsoft.web.vswizard.model.recommender.VSTemporaryInfo;
+import inetsoft.test.*;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft:
# Bug #76497 Review (round 1)

PR: https://github.com/inetsoft-technology/stylebi/pull/5055

## Verdict: Approve

The fix matches the confirmed root cause from `01-diagnosis.md`'s revision
round 1 / `02-refute.md`'s recheck, generalizes correctly, and stays within
the scope of the described change. One process note below (not a blocker).

## 1. Diff matches the described fix

Confirmed via `gh pr diff 5055`. `VSFilterRecommendationFactory.createDataRefs()`
now reads:

```java
AssetEntry[] entries = wizardData.getSelectedEntries();
if(entries == null || entries.length < 1) {
   return null;
}
List<DataRef> refs = new ArrayList<>();
for(AssetEntry entry : entries) {
   refs.add(WizardRecommenderUtil.createColumnRef(entry));
}
return refs.toArray(new DataRef[0]);
```

The old bound-X/Y-field iteration (`ArrayUtils.addAll(tempInfo.getXFields(),
tempInfo.getYFields())`), the `Arrays.asList(fields).forEach(...)` positional
match, and the `getField(DataRef, AssetEntry[])` helper are all genuinely
gone — not just dead/unused. The now-unused imports (`VSAggregateRef`,
`ChartRef`, `VSChartInfo`, `org.apache.commons.lang3.ArrayUtils`) are also
removed. Matches `03-fix.md`'s description exactly.

## 2. Generalizes, doesn't key on the literal repro

`WizardRecommenderUtil.createColumnRef(AssetEntry entry)` returns `null` only
when `entry == null` (verified by reading it directly). Since `createDataRefs()`
loops `wizardData.getSelectedEntries()` — which is null-checked and
length-checked before the loop — every element the loop touches is, by
construction, one of the user's actually-selected entries, never null. This
holds for any number of bound chart fields (0, 1, 2, N) and any selected-entry
count (1 for Range Slider/Selection List, 2+ for Selection Tree) — it is no
longer coupled to "exactly 2 bound fields, 1 selected." Confirmed generalized,
not a special case of the repro.

## 3. Multi-entry Selection Tree claim — confirmed independently

Read `VSWizardBindingHandler.addFilterVSAssembly()` /
`getFilterDataRefs()` (`core/src/main/java/inetsoft/web/vswizard/handler/VSWizardBindingHandler.java`,
current lines ~698-756 on the fix branch):

```java
private DataRef[] getFilterDataRefs(DataRef[] refs) {
   ...
   for(int i = 0; refs != null && i < refs.length; i++) {
      String name = refs[i].getAttribute();   // still no per-element null check
      ...
```

The loop still only guards `refs != null` (the whole array), never
`refs[i] != null` — that part of `addFilterVSAssembly()` was correctly left
alone per "minimal scope." But since `model.getDataRefs()` (the `refs` this
method receives) now always comes back with no null elements — every slot is
directly `createColumnRef(entry)` for a real selected entry — this call site
can never actually hit a null dereference anymore. Confirmed: the fix
resolves this path too, without needing to touch it directly.

## 4. Call-site safety — no missed callers

Grepped `getDataRefs()`/`createDataRefs` across `core/src/main` and
`core/src/test`. The only two call sites of `VSFilterRecommendation
.getDataRefs()` specifically (as opposed to the many unrelated
`getDataRefs()` on `SelectionListVSAssembly`/`TimeSliderVSAssembly`/etc. that
the grep also surfaces) are:

- `VSWizardBindingHandler.java:705` (`addFilterVSAssembly`) — fixed, per #3.
- `VSWizardBindingController.java` → actually
  `core/src/main/java/inetsoft/web/vswizard/controller/VSWizardBindingService.java:410`
  (field-reorder handler for drag-to-reorder Selection Tree levels; the
  fixer's writeup names this file as `VSWizardBindingService.java:401-417`,
  which is correct — I initially looked in `service/` before finding it's
  actually under `controller/`). Read it directly:

  ```java
  DataRef[] refs = vr.getDataRefs();
  List<DataRef> reflist = new ArrayList<>(Arrays.asList(refs));
  DataRef fDataRef = reflist.get(from);
  reflist.remove(from);
  reflist.add(to, fDataRef);
  vr.setDataRefs(reflist.toArray(new DataRef[0]));
  ```

  Confirmed: only moves elements by index, never calls a method on an
  individual `DataRef`. Unaffected by the fix, as claimed.

No other callers exist — `createDataRefs()` is `private`, called only from
this class's own `recommend()`.

## 5. New test — real and load-bearing

Read `VSFilterRecommendationFactoryTest.java` in full (new file, 128 lines).
Both cases match the fixer's description:

- Test 1 (`...evenWithMultipleBoundChartFields`): builds a `VSChartInfo` with
  an `VSChartDimensionRef` X field and a `VSChartAggregateRef` Y field (2
  bound fields), one selected `AssetEntry`, asserts `refs.length == 1` and
  `refs[0] != null`. This is exactly the repro shape.
- Test 2 (`...forMultiEntrySelectionTree`): two selected entries, no bound
  chart fields, asserts `refs.length == 2` with both non-null — this is
  specifically the shape the old code got wrong in the *other* direction
  (old code iterated bound fields, so 0 bound fields → empty array,
  independent of 2 selected entries).

The fixer's `git stash` pre-fix run (`expected: <1> but was: <2>` and
`expected: <2> but was: <0>`) is consistent with the old `createDataRefs()`
logic as read in the diff — not verified by me by re-running it, but the
predicted failure shapes match the code exactly, so I have no reason to
doubt it.

## 6. Same broken pattern elsewhere? None found

Checked sibling classes in
`core/src/main/java/inetsoft/web/vswizard/recommender/object/` for the same
"iterate bound fields, positionally match against selection, produce nulls"
shape. Only `VSCrosstabDefaultRecommendationFactory.java` also references
`getXFields()`/`getYFields()`, but it reads them directly for crosstab
group/aggregate construction — no `getField()`-style positional match against
a selection array, no equivalent null-producing pattern. No other instance of
this bug found.

## 7. Scope check

Per `CLAUDE.md`'s Bug Fix Guidelines (minimal scope): the diff to
`VSFilterRecommendationFactory.java` is exactly the described change plus
dead-code/unused-import removal, no unrelated refactoring. The new test file
is scoped to this bug only.

**Process note (not a blocker):** `gh pr diff --name-only` also shows 5
already-existing files changed
(`SecurityTestDataBuilder.java`, `ScheduleUsersChangeServiceTest.java`,
`CustomSSOConfigTest.java`, `SSOFilterPublisherTest.java`,
`SSOSettingsServiceTest.java`) that are unrelated to this bug. I traced this:
the `bug-76497` branch was cut before an unrelated squash-merge
(`94d129faca`, "Add unit tests for SSO settings validation, cluster
broadcast, and schedule notification (test-only) (#4975)") landed on `main`,
so GitHub's PR diff still lists those 5 commits as "ahead of main." I
diffed the actual file contents at the tip of `origin/bug-76497` against
`origin/main` for all 5 files and they are byte-identical — this is a stale
merge-base artifact, not a real content change, and merging will not alter
those files. Still, worth a rebase before merge for a clean diff/history.

## 8. Prompt-touching check

`gh pr diff --name-only` output: only
`VSFilterRecommendationFactory.java` (main) and the 6 test files listed
above (5 pre-existing/stale, 1 new). None under `wiz-services/src/agents/prompts/`,
no agent system/instruction prompt string, no LangGraph node prompt
template, no `.md`/prompt-doc file. Confirmed clean — not applicable, this is
a Java wizard-recommendation bug.

## 9. CI status

`gh pr checks 5055`: `Semgrep OSS` pass, `Semgrep Scan` pass, `claude-review`
pass, `license/cla` pass. `build` is **pending** at time of this review (not
yet complete).

## Note on fixer's open items

The fixer's `03-fix.md` flags three items as unverified without a live
browser: (1) the actual live-UI repro rendering after the fix, (2) whether
Composer's own entry path reproduces/is fixed identically, (3) which
`WizardEventAspect` swallow-branch (silent metadata-mode drop vs.
`MessageCommand.Type.ERROR`) this repro actually hits. These are accurately
described as open, lead-only, live-browser items — not glossed over as
resolved. No objection to leaving them for the lead's live-browser pass.

fredaan (approved):

PR #5058, opened 2026-09-08
feat(tabular): Datagov (Socrata) catalog SPI (TabularCatalogProvider)

Zxm-happy:
## Summary
- `DatagovRuntime` now implements `TabularCatalogProvider`. The connector's own runtime protocol turns out to be Socrata's classic Views API ("SODA"), not CKAN, despite the connector's name — confirmed by reading `DatagovTable`'s actual JSON parsing (`meta.view.columns`) and by live HTTP calls against real public Socrata sites, before any design was written.
- `listDatasets` enumerates via Socrata's centrally-hosted Discovery API (`api.us.socrata.com/api/catalog/v1?domains=<site>`), fully paginated (capped at a safety ceiling that throws rather than silently truncating), filtered to dataset-typed results. The site domain is derived from the data source's existing `URL` property (its host component) — no new property, zero compatibility impact on already-configured data sources.
- `describeDataset` reads the classic `GET /api/views/{id}/rows.json?max_rows=0` endpoint: fully declarative metadata, no sampling, so `columnsMayBeIncomplete` is `false`. The row identifier column is resolved from Socrata's own `rowIdentifierColumnId`, never a hardcoded name.
- A Socrata 4x4 dataset id is validated before it ever reaches a request URL — `TabularCatalogService.describeTable` passes a caller-supplied target straight through with no guarantee it came from `listDatasets`.
- Kept as an independent `TabularQuery`/`TabularDataSource` (like Elasticsearch), not folded into the `EndpointJsonQuery`/ENDPOINT_CATALOG family — Socrata's catalog surface is a handful of fixed system operations, not a large curated set of business endpoints.

## Verification
Verified against real public Socrata domains (`data.cityofnewyork.us`, `data.cityofchicago.org`, `data.sfgov.org`) in addition to the mocked/loopback-server unit suite: the Discovery API, declarative column metadata (56/31/28 columns across three real datasets), the real HTTP 404 shape for an unknown id, and the exact wire shape `runQuery` depends on (positional `data` arrays aligned to `meta.view.columns`) — all confirmed live via direct HTTP.

Two review rounds:
- Round 1 found a missing `params` round-trip test (production behavior was correct, but nothing exercised `applyQueryContract` → `runQuery` end to end) and a raw `NullPointerException` in `describeDataset` on a blank/null `URL` (inconsistent with `listDatasets`'s own guard).
- Round 2 mutation-tested both fixes — actually reverted each fix, confirmed the corresponding test goes red, then restored — rather than trusting the fix's own description. Converged with no blockers or importants outstanding; one cosmetic nit recorded (a test's own javadoc slightly overclaims which mutation it alone catches — the suite as a whole still catches it via a different, pre-existing test).

66 tests pass, `BUILD SUCCESS` (`./mvnw -pl connectors/inetsoft-datagov -am test`), independently re-run and confirmed clean at three separate points in the process, not just trusted from a single log.

## Test plan
- [x] `DatagovCatalogTest` — unit tests against extracted static fetch methods (`MockedStatic`, same pattern as `MongoCatalogTest`/`ElasticCatalogTests`): pagination (single- and multi-page, plus a safety-ceiling throw), dataset-type filtering, per-`dataTypeName` → `XSchema` mapping, unrecognized-type degradation, malformed/blank dataset-id rejection before any network call, blank/null `URL` rejection before any network call, `params` derivation via `TabularUtil.getPropertyMap`, and a full `applyQueryContract` → `runQuery` round trip against a local loopback server.
- [x] `DatagovMetadataFetchTests` — local loopback `HttpServer` tests confirming both new fetch methods are GET-only with no request body, and that Discovery-API calls never send credentials while site calls send Basic Auth only when configured (matching `runQuery`'s existing behavior).
- [x] `DatagovRuntimeTest` (pre-existing) — unmodified, still passes, confirming no regression to existing `runQuery`/`testDataSource` behavior.
- [x] Live verification against three real public Socrata domains (see above).

### DIFF
--- a/connectors/inetsoft-datagov/src/main/java/inetsoft/uql/datagov/DatagovCatalog.java
+++ b/connectors/inetsoft-datagov/src/main/java/inetsoft/uql/datagov/DatagovCatalog.java
@@ -235,8 +235,8 @@ final class DatagovCatalog {
 
       if(!SOCRATA_4X4.matcher(datasetId).matches()) {
          throw new Exception("Datagov was asked to describe dataset id '" + datasetId + "', " +
-            "which is not a Socrata 4x4 id (exactly 4 lowercase letters/digits, a hyphen, 4 " +
-            "more lowercase letters/digits) -- refusing to embed it into a request URL.");
+            "which is not a Socrata 4x4 id (exactly 4 letters/digits, a hyphen, 4 more " +
+            "letters/digits) -- refusing to embed it into a request URL.");
       }
    }
 
@@ -309,8 +309,14 @@ final class DatagovCatalog {
          return ((JsonNumber) v).longValue();
       }
 
+      // Tolerates a numeric value Socrata sent as a quoted JSON string. JsonValue#toString() on a
+      // JsonString includes the surrounding quotes, so Long.parseLong needs the unquoted text --
+      // without this branch a quoted number always fell through to null, silently.
+      String text = v.getValueType() == JsonValue.ValueType.STRING ? ((JsonString) v).getString()
+         : v.toString();
+
       try {
-         return Long.parseLong(v.toString());
+         return Long.parseLong(text);
       }
       catch(NumberFormatException e) {
          return null;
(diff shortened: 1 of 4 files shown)


Zxm-happy:
Addressed all findings from the automated review, including the minor/non-blocking ones:

1. **Moderate 1** (case-inconsistent round trip): `SOCRATA_4X4` is now `Pattern.CASE_INSENSITIVE`. `listDatasets` applies no case filter of its own, so it could hand back an uppercase 4x4 id that `describeDataset` would then reject — a self-inconsistent round trip for a value the connector itself just vouched for. Added `describeDataset_uppercaseId_acceptedCaseInsensitively`.
2. **Minor 2** (dead code in `getLong`): the string-fallback branch never actually fired — `JsonValue#toString()` on a `JsonString` returns the quoted text (`"123"`), so `Long.parseLong` always threw. Fixed to unwrap via `((JsonString) v).getString()` first. Added `describeDataset_quotedNumericRowIdentifierColumnId_stillResolvesKeyColumn`.
3. **Minor 3** (no timeouts): added connect/read timeouts to `getDiscoveryMetadata`/`getSiteMetadata` only (not the pre-existing `runQuery` path, matching the finding's own scoping). This mirrors a real hang this session's own live verification hit earlier in this same connector's development.

Item 4 (hardcoded `api.us.socrata.com` region) is informational/out of scope per the review itself — left as-is, already noted as a known limitation in `claude/tabular-data-source.md`.

63 tests pass (59 + 4), including 3 new regression tests for the fixes above.

Nelson-cao (approved):

PR #5054, opened 2026-09-07
Fix Bug #76499: Custom Log Level checkbox cross-selects rows sharing the same name

larryliang-inetsoft:
## Summary
- `TableView.refreshDataSource()` rebuilds row selection by matching only on `row[trackByProp]`. `LoggingLevelTableComponent` hardcoded `[trackByProp]="'name'"`, which is not unique once multi-tenant orgs are involved — two rows can share `name` while differing in `orgName`/`level` (backend identity is `context+name+orgName`, per `LogLevelSetting.equals()`/`hashCode()`).
- Fix: stamp each row with a non-enumerable composite key (`context_name_orgName`) via a `keyedLoggingLevels` getter, and point `trackByProp` at that key instead of `'name'`. The key is non-enumerable so it is not picked up by `JSON.stringify` when the row is POSTed back to the server — no change needed to `LogLevelDTO`'s wire model, and no change to the shared `TableView`/`trackByProp` mechanism (so the other 6 `trackByProp` consumers in the `em` app are unaffected).
- Also fixed `removeLevels()`, which had the same non-unique-identity defect (matched only on `name`+`context`) on the Remove-button path — selecting one row and clicking Remove would have deleted both same-name rows.

Fixes Bug #76499.

## Test plan
- [x] Added `logging-level-table.component.spec.ts` (no prior spec existed) with a regression test: two rows sharing `name` but differing in `orgName`, select one, simulate the parent's per-change-detection-cycle re-render (`getLevels()` returning a fresh array of the same row objects), assert only the originally-selected row remains selected.
- [x] Verified the test is load-bearing on the fix: stashing the `.ts`/`.html` change (keeping the new spec) makes it fail to compile (`keyedLoggingLevels` does not exist).
- [x] `npx ng test em --include "**/logging-level-table.component.spec.ts"` — 2/2 passed.
- [x] `npx ng test em` scoped to the 5 directly-related files (logging-level-table, logging-settings-view, logging-settings-page, table-view, regular-table) — 9/9 passed.
- [x] Full `npm run test:em` suite — 106 files / 376 tests, all passed, no regressions.
- [ ] Live-browser end-to-end confirmation (checking one row does not select its sibling; Remove only removes the selected row) — pending, needs two rows in EM > Settings > Logging > Custom Log Levels sharing `name` with different `orgName` (multi-tenant only).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5059, opened 2026-09-08
Guard VSTreeHandler.getTableTreeModel() against a missing/disposed assembly

larryliang-inetsoft:
## Summary

Defensive hardening fix, **not a fix for a specific reported bug**.

`VSTreeHandler.getTableTreeModel()` resolved the crosstab/calc-table assembly by name and then dereferenced it unconditionally. Its sibling `getChartTreeModel()` already guards against this exact scenario (a concurrently disposed/reset viewsheet, or a concurrently missing assembly) with an explicit comment explaining why. `getTableTreeModel()` had no equivalent guard, so a missing assembly produced an uncaught `NullPointerException` — silent to the user, since the STOMP command chain that calls this (`VSBindingTreeController → VSBindingTreeControllerService → VSBindingTreeService → VSTreeHandler`) has no exception handling anywhere above it.

This closes that gap:
- Adds the missing null guard to `getTableTreeModel()`, mirroring `getChartTreeModel()`'s existing pattern.
- Fixes the one caller (`VSBindingTreeService.getBinding()`) that called into `appendVSAssemblyTree()` unconditionally afterward — that method itself unconditionally dereferences the tree model, so a bare null-return guard alone would just relocate the NPE one frame down for this specific call site. The chart branch already had this null-check; the table branch didn't.

## Background

Found while investigating Redmine Bug #76488 and Bug #76485 via `/deep-debug-workflow` (three parallel investigators, plus lead follow-up — see `docs/teams/2026-09-07-bug-76488-76485/` in the enterprise repo for the full write-up). Despite a thorough investigation, **no reproduction of either bug was achieved**, and no confirmed trigger for this specific null case was found in either bug's exact repro shape (a plain, non-embedded-viewsheet crosstab). This PR closes a real, independently-verifiable gap on its own merits — it should **not** be read as resolving either reported bug, and neither bug's Redmine status should change based on this PR alone.

## Test plan

- [x] Added `VSTreeHandlerTest` with two cases (missing assembly, disposed runtime viewsheet) — both pass with the fix.
- [x] Confirmed both new tests are load-bearing: stashed the fix (keeping the tests), re-ran — both failed with the exact NPE this PR closes (`NullPointerException ... "cass" is null ... at VSTreeHandler.java:244`), confirming the tests aren't tautological.
- [x] Traced the one call site of `getTableTreeModel()` (`VSBindingTreeService.getBinding()`) and fixed the accompanying missing null-check there too, so the fix doesn't just relocate the NPE.
- [x] `./mvnw -pl core test -Dtest=VSTreeHandlerTest` — 2/2 passed.

### DIFF
--- a/core/src/main/java/inetsoft/web/binding/handler/VSTreeHandler.java
+++ b/core/src/main/java/inetsoft/web/binding/handler/VSTreeHandler.java
@@ -232,6 +232,12 @@ public class VSTreeHandler {
       String aname = cinfo.getAbsoluteName();
       VSAssembly cass = vs0 == null ? null : (VSAssembly) vs0.getAssembly(aname);
 
+      // viewsheet may have been disposed/reset concurrently (e.g. user interacts with the
+      // binding tree before a heavy operation finishes), so cass can be null here.
+      if(cass == null) {
+         return null;
+      }
+
       SourceInfo sinfo = cinfo.getSourceInfo();
       String prefix = null;
       String source = null;
--- a/core/src/main/java/inetsoft/web/binding/service/VSBindingTreeService.java
+++ b/core/src/main/java/inetsoft/web/binding/service/VSBindingTreeService.java
@@ -90,7 +90,7 @@ public class VSBindingTreeService {
                assetTreeModel = treeHandler.getTableTreeModel(
                   engine.getAssetRepository(), rvs, (TableDataVSAssemblyInfo) info, principal);
 
-               if(appendComponentTree) {
+               if(assetTreeModel != null && appendComponentTree) {
                   treeHandler.appendVSAssemblyTree(rvs, assetTreeModel, principal, assembly);
                }
             }
--- a/core/src/test/java/inetsoft/web/binding/handler/VSTreeHandlerTest.java
+++ b/core/src/test/java/inetsoft/web/binding/handler/VSTreeHandlerTest.java
@@ -1,0 +1,97 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+package inetsoft.web.binding.handler;
+
+import inetsoft.report.composition.AssetTreeModel;
+import inetsoft.report.composition.RuntimeViewsheet;
+import inetsoft.sree.security.SecurityEngine;
+import inetsoft.test.*;
+import inetsoft.uql.asset.AssetRepository;
+import inetsoft.uql.viewsheet.Viewsheet;
+import inetsoft.uql.viewsheet.internal.TableDataVSAssemblyInfo;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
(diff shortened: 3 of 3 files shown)


larryliang-inetsoft:
## Review: PR #5059 — Guard VSTreeHandler.getTableTreeModel()

**Verdict: Changes requested** (one Important/blocking finding below; the core fix itself is sound).

### What I checked

1. **Null guard mirrors sibling method.** Confirmed `getTableTreeModel()`'s new `if(cass == null) return null;` (line 237) is shaped identically to `getChartTreeModel()`'s existing `if(vs0 == null) return null;` / `if(chart == null) return null;` guards (lines 70-72, 78-80) — same early-return-null pattern, same style of explanatory comment.
2. **Companion fix in `VSBindingTreeService.getBinding()` is real and necessary.** `appendVSAssemblyTree(RuntimeViewsheet, AssetTreeModel, Principal, VSAssembly)` (VSTreeHandler.java:527) unconditionally does `AssetTreeModel.Node mroot = (AssetTreeModel.Node) model.getRoot();` — so without the added `assetTreeModel != null &&` guard, a null return from `getTableTreeModel()` would just relocate the NPE one frame down. Confirmed load-bearing, not decorative.
3. **Single caller confirmed.** Grepped both `community/` and `enterprise/` repos for `getTableTreeModel` — only caller is `VSBindingTreeService.getBinding()`. No other call site is left with the unguarded-null-model bug.
4. **New test (`VSTreeHandlerTest`)** — both cases genuinely exercise the guard: missing assembly (`viewsheet.getAssembly("Crosstab1")` → null) and disposed viewsheet (`rvs.getViewsheet()` → null), both asserting `assertNull(result)`. Sound.
5. **Prompt-touching check** — confirmed via `--name-only` that nothing under `wiz-services/src/agents/prompts/` or any prompt doc is touched.
6. **Claim boundary** — PR description explicitly states this is "**not a fix for a specific reported bug**" and "should **not** be read as resolving either reported bug." Matches the intended framing.

### Important — diff is not minimal / branch is not based on clean `main`

`gh pr diff --name-only` and the commit list show this PR carries **5 unrelated commits** beyond the stated fix:

- `Add unit tests for SSOSettingsService protocol-switch validation and CustomSSOConfig mutual exclusion`
- `Add unit tests for SSOFilterPublisher cluster broadcast and ScheduleUsersChangeService SSO-toggle notification`
- `Make SSOSettingsServiceTest's OpenID blank-attributes case hermetic`
- `Document the Bug #72373 citation's verification source in SSOFilterPublisherTest`
- `Stop leaking DataChangeListener registrations in SecurityTestDataBuilder`

These touch `SecurityTestDataBuilder.java` and add four new SSO-related test files, none of which are mentioned in the PR description, and none of which relate to `VSTreeHandler`/binding trees. I traced the branch history: `hardening-vstreehandler-nullguard`'s parent is commit `aa386775f` — the same common ancestor shared by the still-open `bug-76497` (PR #5055) and `bug-76499` (PR #5054) branches — **not** a fresh `origin/main` commit as `docs/teams/2026-09-07-bug-76488-76485/03-fix.md` claims (that doc even cites a commit hash, `a16e4bfd`, that doesn't exist in this repo).

Practical effect: if this PR merges before #5054/#5055, `main` will pick up those 5 unrelated commits attributed to a PR titled/scoped as a two-file defensive hardening fix — violating this repo's minimal-scope guideline and potentially creating duplicate/conflicting commits when #5054/#5055 eventually merge on top of the same base.

**Ask:** rebase `hardening-vstreehandler-nullguard` onto current `origin/main` (`88d75bb9`) so the diff contains only the `VSTreeHandler.java` / `VSBindingTreeService.java` / `VSTreeHandlerTest.java` changes, per the PR's own stated scope.

### CI status
- Semgrep OSS: pass
- Semgrep Scan: pass
- `claude-review`: pass
- `build`: pending at time of review
- CLA: signed

Once rebased onto a clean `main` (and `build` finishes green), this is good to merge — no other blockers found.

larryliang-inetsoft:
Thanks for catching this — confirmed the same stale-merge-base issue found on the sibling bugs' branches (#5054, #5055) in this batch. The branch was accidentally seeded from an old shared ancestor (aa386775f) rather than fresh origin/main.

Rebased by cherry-picking just the real fix commit onto a fresh origin/main (88d75bb9a) and force-pushed. PR now has exactly 1 commit (56bf567b5) touching only the 3 intended files (VSTreeHandler.java, VSBindingTreeService.java, VSTreeHandlerTest.java). CI is re-running against the clean commit.

fredaan (approved):

PR #5034, opened 2026-09-06
fix(VBM-005): connectedViewsheets() matches by logical identity, not raw Principal.equals()

larryliang-inetsoft:
## Summary
- `WorksheetAgentController.connectedViewsheets()` relied on `WorksheetService.getRuntimeSheets(user)`'s built-in `Principal.equals()` filter, so a browser-paired viewsheet whose owning `SRPrincipal` came from a different `HttpSession` than the calling agent's JWT-derived one (same logical user, different session -- session renewal, a different browser window, a long-lived cached token) was silently dropped, and the PVA-002 `resetRuntime()` reset never fired for it.
- Same failure mode `SheetRuntimeAccess.grantOwnershipBypass()` already documents and fixes elsewhere in this package via `PairingUtil.sameLogicalUser()` -- `connectedViewsheets()`/`save()` was the one runtime-resolution path in this package that was never updated.
- Fix: fetch every live runtime sheet (`getRuntimeSheets(null)`) and filter manually by base-entry match plus `PairingUtil.sameLogicalUser(rvs.getUser(), user)`, mirroring the existing shipped pattern. Corrects the method's stale javadoc, which misdescribed this exact gap as a "documented limitation."
- Widen-then-narrow: a same-`Principal` match under the old code is provably still a `sameLogicalUser` match under the new code (equal `SRPrincipal`s carry equal `IdentityID`s by construction), so this can only add matches the old code silently dropped, never remove one -- confirmed both by construction and by the still-green `plainSaveInPlaceResetsConnectedViewsheetRuntimes` continuity test.

## Test plan
- [x] New regression test `plainSaveInPlaceResetsConnectedViewsheetAcrossIdentitySessionBoundary` (two real, distinct `SRPrincipal`s, same logical user) — RED before the fix, GREEN after.
- [x] Full `WorksheetAgentControllerTest` class: 132 tests, 0 failures (two pre-existing tests' mocks updated to match the new `getRuntimeSheets(null)` call argument -- mechanical stub fix only, no behavior-assertion change).
- [x] Full `inetsoft.web.wiz` package: 3196 tests, 0 failures, 1 unrelated pre-existing skip.

Full root-cause/repro/fix writeup: `docs/teams/2026-09-06-bug-vbm-005/` in the `stylebi-wiz` repo (00-charter.md, 02-root-cause.md, 03-repro.md, 04-fix.md).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5060, opened 2026-09-08
Fix VBM-005: colliding first save out of TEMPORARY_SCOPE skips connected-viewsheet reset

larryliang-inetsoft:
## Summary
- Bug VBM-005 (#76482): a `create_worksheet()`-derived session is always `TEMPORARY_SCOPE` at the time of its first save. `WorksheetAgentController.save()`'s `if(!isSaveAs && !wasTemporary)` gate skipped the `connectedViewsheets(entry, user).resetRuntime()` scan unconditionally on `wasTemporary=true` — even when that save's target name collides with (and, via `confirmed:true`, overwrites) an asset that is already a connected viewsheet's base entry. The connected viewsheet then silently serves stale data forever, with `ok:true` on every step and no warning anywhere.
- Fix: reuse the already-computed `isDuplicatedEntry` check to make the `wasTemporary` half of the gate collision-aware. A genuinely new name still skips the scan (cheap, no-op); a colliding name now runs it, exactly like the existing plain-in-place-save case does.
- Scoped to `wasTemporary` only, per this bug's diagnosis/refute — `isSaveAs` (Save-As) is left untouched; its analogous gap is a documented, intentional non-reset today (paired with the `affectedViewsheets` warning) and is tracked as a separate issue.

## Test plan
- [x] Added `firstSaveOutOfTemporaryScopeResetsConnectedViewsheetOnCollision` — a `TEMPORARY_SCOPE` first save whose target name collides with an existing, connected viewsheet's base entry now triggers `resetRuntime()`.
- [x] Existing `firstSaveOutOfTemporaryScopeDoesNotCheckForConnectedViewsheets` (non-colliding case) needed no change and still passes unmodified.
- [x] `./mvnw -pl core -am test -Dtest=WorksheetAgentControllerTest` — 133 tests, 0 failures, 0 errors.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -1129,6 +1129,8 @@ public class WorksheetAgentController {
       List<String> affectedViewsheets = isSaveAs
          ? connectedViewsheetIds(connectedViewsheets(oldEntry, user)) : List.of();
 
+      boolean isDuplicatedEntry = false;
+
       if(name != null && !name.isEmpty()) {
          IdentityID uname = IdentityID.getIdentityIDFromKey(user.getName());
          int assetScope = "user".equalsIgnoreCase(body.scope())
@@ -1143,9 +1145,10 @@ public class WorksheetAgentController {
          // called the equivalent check, so a second save_worksheet({name:"X"}) silently
          // overwrote whatever "X" already held with no signal anything collided.
          try {
-            if(worksheetService.isDuplicatedEntry(worksheetService.getAssetRepository(), entry)
-               && !Boolean.TRUE.equals(body.confirmed()))
-            {
+            isDuplicatedEntry =
+               worksheetService.isDuplicatedEntry(worksheetService.getAssetRepository(), entry);
+
+            if(isDuplicatedEntry && !Boolean.TRUE.equals(body.confirmed())) {
                throw new PairingException(
                   "'" + name + "' already exists. Pass confirmed:true to overwrite it, or " +
                   "choose a different name.");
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft (commented):
## Review r1 — APPROVE (posted as comment; self-review blocked, same account as PR author)

Full findings: `docs/teams/2026-09-07-bugs-76482-vbm-batch/bug-vbm005/04-review-r1.md`

Summary of the adversarial pass:
1. **Keys on the real rule, not the repro literal** — `isDuplicatedEntry` is a general repository lookup on whatever name/entry the caller passes; `"Orders WS"` is only the test fixture's choice, not a condition in the shipped code.
2. **`isDuplicatedEntry` hoist is behavior-preserving** — same call, same arguments, same point in control flow, just captured into a variable that outlives the `try` block. Verified against the full pre-fix method body, not just the diff hunk.
3. **Fix sits at the right level** — confined to the `wasTemporary` half of the gate inside `save()` itself; `isSaveAs`'s own computation and its non-reset behavior are byte-for-byte untouched — no scope creep into the separately-flagged Save-As gap.
4. **Sibling check** — `git grep isSaveAs` across `core/src/main/java` at the base commit shows this gate shape exists only in `WorksheetAgentController.save()`. No other controller has the same `wasTemporary`/`connectedViewsheets` pattern.
5. **New test asserts the real claim** — `firstSaveOutOfTemporaryScopeResetsConnectedViewsheetOnCollision` asserts `resetRuntime()` **is** called (mirror of the refuter's own throwaway test against the old code, which asserted `never()`), using the real entry-equality + `sameLogicalUser` matching path, not an over-permissive stub.
6. **CI**: Semgrep OSS/Scan, claude-review, license/cla all pass. `build` was still pending after repeated polling at review time — no failure observed. If it does fail on `PermissionHierarchyTest.setUp:96`, that's this repo's documented pre-existing, order-dependent flake (seen and independently confirmed harmless on PR #5034 and PRs #5009/#5011) — safe to merge past if that's the only failure.

No blocker or important findings.

fredaan (approved):

PR #5061, opened 2026-09-08
fix(wiz-worksheet): surface timeSeries flag in read_worksheet_model groups (WBS-024, #76502)

fredaan:
## Summary
- `read_worksheet_model`'s `aggregates.groups[]` never reported whether `set_group_aggregate`'s `timeSeries` flag was applied to a group, even though the write path (`WorksheetMutationSupport#applyAggregateInfo`) already sets it unconditionally on the `GroupRef`.
- Root cause: `WorksheetReadService.readAggregates` never called `GroupRef.isTimeSeries()`, and `WorksheetModel.AggregateModel.GroupModel` had no field to carry it.
- Fix: add a `timeSeries` component to `GroupModel` and populate it from `gr.isTimeSeries()` at the (single, grep-confirmed) construction site in `readAggregates`.

## Test plan
- [x] Extended `WorksheetReadServiceTest.readsDateGroupLevelOnGroupedColumn` with a `timeSeries()` assertion (false/default case)
- [x] Added `WorksheetReadServiceTest.readsTimeSeriesFlagOnDateGroupedColumn`, round-tripping through the real `WorksheetMutationSupport.applyAggregateInfo` mutator with `timeSeries: true`, asserting `group.timeSeries()` is `true`
- [x] `./mvnw test -pl core -Dtest=WorksheetReadServiceTest` -- 41/41 pass

Refs Redmine #76502, WBS-024.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
@@ -755,7 +755,7 @@ public class WorksheetReadService {
          String field =
             base instanceof ColumnRef cr && cr.getDataRef() instanceof DateRangeRef dr
                ? dr.getDataRef().getName() : gr.getName();
-         groups.add(new WorksheetModel.AggregateModel.GroupModel(field, dateLevel));
+         groups.add(new WorksheetModel.AggregateModel.GroupModel(field, dateLevel, gr.isTimeSeries()));
       }
 
       // Aggregates (primary + secondary)
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/model/WorksheetModel.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/model/WorksheetModel.java
@@ -228,14 +228,17 @@ public record WorksheetModel(List<TableModel> tables, List<VariableModel> variab
       /**
        * A single group-by dimension.
        *
-       * @param field     source column name
-       * @param dateLevel the date grouping level applied directly to this group (e.g.
-       *                  {@code "QUARTER"}), same vocabulary as set_group_aggregate's
-       *                  {@code dateLevel} / add_date_range_column's {@code dateOption};
-       *                  {@code null} for a plain (non-date-bucketed) group
+       * @param field      source column name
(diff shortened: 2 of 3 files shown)


fredaan:
## Review round 1

Head commit reviewed: `5de1964c4258afd20f6f129a4c5aad849aa7b048`

**What was checked** (independently re-verified against the live `stylebi` working tree, not taken on faith from the diagnosis/refutation docs):
- `GroupModel`/`readAggregates` have exactly one construction call site each (grep-confirmed) — the fix cannot miss a crosstab or secondary branch.
- The write-side `gr.setTimeSeries(true)` (`WorksheetMutationSupport.java:706`) sits after the plain-group/date-level branches converge, so it applies unconditionally — the read-side fix correctly mirrors that with no guard needed.
- `GroupSpec(String, String, Boolean)` is a real, unambiguous compatibility constructor the new test resolves to correctly.
- `GroupRef.timeSeries` defaults to `false` with no alternate initializer — matches the extended test's `assertFalse` on the pre-existing (non-time-series) fixture.
- Manually walked the test-file diff hunk for brace balance — no stray/duplicated brace, the two test methods are correctly separated.
- `@JsonInclude(NON_NULL)` has no effect on the new primitive `boolean` component (only suppresses null object refs) — it will always serialize `true`/`false` as intended.
- Swept `web/wiz` for other `GroupRef[]`/`getGroups()` consumers that build a read-facing model (`BindingReadService`, `TableBindingMutator`, `WorksheetTableService`, `MetadataApiService`, `SSOTokenService`, `WorksheetTable`) — all are viewsheet-binding/metadata-endpoint domains outside this bug's scope, none duplicate the same write/read asymmetry for this record shape.
- Checked for overfit smells (literal-from-repro keying, special-casing the reported input, a regression test that only covers the `true` direction) — none found; the fix corrects the general accessor gap and the new test pair asserts both `true` and `false`.

**Noted, not a blocker**: the write path only ever sets `timeSeries` to `true`, never explicitly clears it to `false` on a later call without the flag — a pre-existing write-side asymmetry this PR doesn't touch. The read-side fix now honestly surfaces that stale value instead of masking it, which is strictly better than before, not a regression this PR introduces.

**Findings**: none at blocker or important confidence.

**Verdict: approved.**

PR #5063, opened 2026-09-08
fix(wiz): read_worksheet_model surfaces a variable's Values picker (bug 76502 WBS-030)

fredaan:
## Summary
- `WorksheetReadService.readVariable()` never read a variable's "Values" picker
  (`choices`/`values`/`tableName`/`labelAttribute`/`valueAttribute`/`displayStyle`) off
  `AssetVariable` -- so a picker set via `add_variable`/`edit_variable` was invisible to
  `read_worksheet_model`, even though it was genuinely stored and enforced server-side.
- Adds `WorksheetModel.ChoicesModel` (mirroring `WorksheetMutationSupport.VariableChoicesSpec`
  field-for-field) and populates `VariableModel.choices()` using the same embedded-vs-query-mode-
  vs-no-picker branching `applyVariableChoices` uses on write, so the read side can never diverge
  from what the write side considers "has a picker".
- `DATE_COMBOBOX` (ordinal 5) has no forward case in `parseVariableDisplayStyle` -- it's reachable
  only via StyleBI's native, non-agent Composer variable dialog. Deliberate scope decision: surface
  it as `"date_combobox"` on read (never drop the variable or crash), but do **not** add it to the
  write vocabulary -- that would be a new capability, not a fix for this read-side gap.

## Root cause
`WorksheetReadService.readVariable()` (`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java`)
built a flat 4-field `VariableModel(name, label, type, defaultValue)` and never called the choices-related
getters on `AssetVariable`, even though the exact same object instance is populated by
`WorksheetMutationSupport.applyVariableChoices()` on the write path (`WorksheetEditService.editVariable()`
clones-mutates-publishes the same `AssetVariable` the read path next observes).

## Test plan
- [x] `./mvnw test -pl core -Dtest=WorksheetReadServiceTest` -- 44/44 pass (40 existing + 4 new:
  embedded-choices round-trip, query-mode-choices round-trip, no-picker → null, DATE_COMBOBOX
  read-only round-trip), all round-tripped through the real `applyVariableChoices` mutator rather
  than a hand-built `AssetVariable`.
- Broader `-pl core` run shows unrelated `NoClassDefFoundError`/Spring-context-shutdown errors from
  parallel-fork JVM contamination when running the full ~3400-test suite together (pre-existing
  environment flakiness, not from this change) -- `WorksheetReadServiceTest` alone is green in
  isolation both before and after adding the new tests.

Bug 76502 WBS-030.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
@@ -448,8 +448,21 @@ public class WorksheetReadService {
     * non-null) and query-mode ({@code getTableName()} non-null) are mutually exclusive, and
     * neither being set means no picker at all -- reported as {@code null}, not an empty/default
     * {@code ChoicesModel}.
+    *
+    * <p>{@code displayStyle == NONE} is checked FIRST, before {@code getChoices()}/
+    * {@code getTableName()}, matching {@code VariableAssemblyDialogService} (the native Composer
+    * variable dialog's own reader, {@code :101-104}). {@code applyVariableChoices} only clears
+    * {@code choices}/{@code values}/{@code table} when a NEW {@code values}/{@code table} source
+    * is supplied in the same call -- a caller that supplies only {@code displayStyle: "none"}
+    * (turning the picker off without resupplying/clearing its source) leaves the old picker data
+    * on the object untouched, which would otherwise read back as a fully populated picker
+    * alongside a "none" style.
     */
    private WorksheetModel.ChoicesModel readVariableChoices(AssetVariable var) {
+      if(var.getDisplayStyle() == UserVariable.NONE) {
+         return null;
+      }
+
       String displayStyle = displayStyleName(var.getDisplayStyle());
 
       if(var.getChoices() != null) {
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java
@@ -428,7 +428,7 @@ public class WorksheetReadService {
       AssetVariable var = dva.getVariable();
 
       if(var == null) {
-         return new WorksheetModel.VariableModel(dva.getName(), null, null, null);
+         return new WorksheetModel.VariableModel(dva.getName(), null, null, null, null);
       }
 
       String label = var.getAlias();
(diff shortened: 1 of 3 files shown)


fredaan:
## Review — Round 1

Head commit reviewed: `ba0602c7d1cfe3a5387fc375f7e920734ad0cc4d`

Re-verified every claim in the diagnosis/refute docs directly against source in `community/`
rather than trusting the write-up:

- **Naming-inversion claim (refute point 8)** — confirmed correct, not backwards.
  `applyEmbeddedVariableChoices` (`WorksheetMutationSupport.java:2138-2146`) writes labels via
  `setChoices()` and typed values via `setValues()`; the fix's `readVariableChoices()` reads
  `stringify(var.getValues())` into `ChoicesModel.values()` and `stringify(var.getChoices())`
  into `ChoicesModel.labels()` — matches. The round-trip test would have caught a swap.
- `Object[]`→`List<String>`/`DataRef`→column-name conversions are correct; `DataRef.getAttribute()`
  matches the same idiom this file already uses in `readColumn()`.
- `DATE_COMBOBOX` stays read-only as claimed — `parseVariableDisplayStyle` (write side) is
  untouched by this diff.
- "No picker" correctly reads back as `null`, not an empty `ChoicesModel` — `choices`/`values`
  default to `null` on `UserVariable` and are always nulled together on the write side.
- Did not check out the branch to re-run the full suite (out of review scope here), but the
  "pre-existing JVM/Spring-shutdown contamination" explanation for the wide-run failures is
  statically plausible: this diff touches no shared static state or bean wiring, and
  `WorksheetReadServiceTest` itself reportedly errors in the wide run despite being 44/44 clean
  in isolation both before and after the change.

### Important finding (confidence ~82)

`readVariableChoices()` can report a **stale, self-contradictory picker**. The write side
(`applyVariableChoices`) only clears `choices`/`values`/`table` when a new `values`/`table`
source is supplied in the same call; if only `displayStyle` is given (no `values`/`table`),
any previously-stored picker is left untouched on the object while the style flips to
whatever was requested. This is directly reachable via the plugin's own `edit_variable` tool:
`choices: { displayStyle: "none" }` alone, with no `values`/`table`, is accepted by
`validateVariableChoicesSpec` with no complaint (`plugin/composer/src/tools/worksheetTools.ts:1790-1812`) —
a very natural "turn this picker off" call. After that, the `AssetVariable`'s `getChoices()`/
`getTableName()` are still populated but `getDisplayStyle()` is `NONE`.

`readVariableChoices()` branches on `getChoices() != null`/`getTableName() != null` first and
never short-circuits on `displayStyle == NONE`, so it reports `displayStyle: "none"` **and** a
fully populated stale `choices` object simultaneously — internally contradictory, and it
misrepresents the variable as still having an active enumerated picker when it doesn't.

The codebase already has the right precedent for this exact ambiguity:
`VariableAssemblyDialogService.java:101-104` checks `getDisplayStyle() == UserVariable.NONE`
**first**, before ever looking at `getChoices()`/`getTableName()`, specifically to avoid
surfacing a leftover picker once the style is `NONE`. `readVariableChoices()` should follow the
same precedence:

```java
private WorksheetModel.ChoicesModel readVariableChoices(AssetVariable var) {
   if(var.getDisplayStyle() == UserVariable.NONE) {
      return null;
   }
   String displayStyle = displayStyleName(var.getDisplayStyle());
   ... // unchanged below
}
```

Verified this doesn't disturb any of the 4 new tests (the no-picker fixture's `AssetVariable`
already defaults to `style == NONE`; the three picker-present fixtures all explicitly set a
non-`NONE` style).

### Verdict

Changes requested — one important-severity finding, everything else in the diff held up under
adversarial re-verification against source.

fredaan:
## Review — Round 2

Head commit reviewed: `27f1d387f1e229542252d52cc40725e902fffc18` (one commit since round 1).

Read the actual diff, not just the write-up.

**Round-1 finding — fixed, correctly placed.** `readVariableChoices()` now opens with:

```java
if(var.getDisplayStyle() == UserVariable.NONE) {
   return null;
}
```

before `displayStyleName()` is computed and before either the `getChoices()`/`getTableName()`
branch — no state is side-effected ahead of the check. This also means the guard isn't
embedded-mode-specific: it sits above both branches, so a query-mode picker turned off via
`displayStyle: "none"` alone would be equally covered, even though the new test only exercises
the embedded case (a symmetric query-mode test would be a nice completeness addition, but not
required — the fix is generic, not per-branch).

**New test exercises the real sequence.** `variableWithDisplayStyleSetToNoneAloneReportsNullChoicesNotStaleData`
does (1) a full `applyVariableChoices` call populating an embedded picker, then (2) a *second*,
separate call supplying only `displayStyle: "none"` — the exact two-call sequence the finding
described, not a simplified stand-in. Traced call (2) through `applyVariableChoices`:
`hasValues=false, hasTable=false` → neither clearing helper runs → only `setDisplayStyle(NONE)`
executes → `getChoices()`/`getValues()` remain stale from call (1). Without the round-2 fix this
test would fail (readVariableChoices would still see `getChoices() != null` and return a
populated model) — confirms it's a real regression test.

**Reverse case re-traced independently.** Non-`NONE` style set alone (no `values`/`table`) on a
variable with no prior picker: `getDisplayStyle() == NONE` guard doesn't fire (style is e.g.
`COMBOBOX`), `displayStyleName()` computes a value that's simply unused, both `getChoices() !=
null`/`getTableName() != null` are false, falls through to the existing final `return null` —
identical before and after the round-2 change. Confirmed correct, not just plausible.

### Verdict: Approve

The round-1 finding is fixed, correctly placed, and covered by a regression test that exercises
the actual reported sequence and would have failed against the round-1 code. No new findings.

PR #5066, opened 2026-09-08
feat(wiz): set_variable_values accepts multi-select variable values (bug 76502 WBS-029)

fredaan:
## Summary
- `set_variable_values` could never assign more than one value to a variable whose picker (`add_variable`'s `choices.displayStyle`) is `list`/`checkboxes` — `EditRequest.variableValues` was `Map<String, String>`, so a JSON array 400'd at deserialization.
- `EditRequest.variableValues` is now `Map<String, Object>`: a plain string keeps today's single-value behavior byte-for-byte, a 1-element list is treated identically to a plain string regardless of display style, and a 2+ element list is a genuine multi-value assignment — gated behind a `list`/`checkboxes` display style, validated per element (fixing the previous "whole joined string checked as one candidate" bug), converted to a real `Object[]` with `vtable.setAsIs(true)` so a downstream `ONE_OF` condition doesn't re-split an element on embedded commas. Mirrors `VariableInputDialogService.initVariableInfos`, the native "Enter Parameters" dialog's own submit path.
- Relocates WBS-030's `displayStyleName` inverse-mapping helper from `WorksheetReadService` to `WorksheetMutationSupport` as a shared package-visible static, since this fix's reject message needs the same mapping (was flagged as a soft coordination dependency in the design; WBS-030/#5063 is already merged, so this lands directly on the shared version rather than a temporary duplicate).
- New capability, built per an approved-with-one-required-fix design: `docs/teams/2026-09-08-bugs-76502-wbs-gaps/bug-wbs029/01-design.md` and `02-design-review.md` in `inetsoft-technology/stylebi-wiz`.

Companion plugin PR (client schema + precheck): inetsoft-technology/stylebi-wiz#2295

## Test plan
- [x] `WorksheetAgentControllerTest` (152 tests, 0 failures) — added: single string on multi-select still works, array of 2 on checkboxes/list succeeds, one invalid token named specifically, array on combobox rejected naming the style, 1-element array treated as scalar regardless of style, empty array rejected, non-string element rejected, unknown variable still rejected with an array value.
- [x] `EditRequestParametersDeserializationTest` (3 tests) — new test confirms the real app `ObjectMapper` binds a JSON string to `String` and a JSON array of strings to `List<String>` for the now-`Map<String, Object>` `variableValues` field.
- [x] `WorksheetReadServiceTest` (46 tests, 0 failures) — confirms the relocated `displayStyleName` helper didn't regress WBS-030's read path.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/EditRequest.java
@@ -100,7 +100,10 @@ import java.util.Map;
  *       produces via the Composer's own "Add Grouping" dialog; or {@code type} (standalone
  *       grouping, matched by data type; defaults to {@code "string"})</li>
  *   <li>{@code set_column_description} — {@code table}, {@code column}, {@code description}</li>
- *   <li>{@code set_variable_values} — {@code variableValues} (map of variable name → value)</li>
+ *   <li>{@code set_variable_values} — {@code variableValues} (map of variable name → value,
+ *       where value is a {@code String} for a single value or a {@code List<String>} of 2+
+ *       entries to assign multiple values at once to a {@code list}/{@code checkboxes}
+ *       display-style variable)</li>
  *   <li>{@code set_mirror_auto_update} — {@code table}, {@code visible} (true=auto-update on, false=off)</li>
  *   <li>{@code convert_to_embedded} — {@code table}</li>
  *   <li>{@code set_assembly_position} — {@code table}, {@code x}, {@code y}</li>
@@ -233,8 +236,12 @@ public record EditRequest(
    List<WorksheetMutationSupport.GroupMapping> groupMappings,
    /** Whether to group unmapped values as "Others" for add_named_group. */
    Boolean groupOthers,
-   /** Variable name → value mappings for set_variable_values. */
-   Map<String, String> variableValues,
+   /**
+    * Variable name → value mappings for set_variable_values. Each value is either a
+    * {@code String} (single value) or a {@code List<String>} (2+ entries assign multiple
+    * values at once; a 1-element list is treated identically to a plain String).
+    */
+   Map<String, Object> variableValues,
    /** X pixel coordinate for set_assembly_position. */
    Integer x,
    /** Y pixel coordinate for set_assembly_position. */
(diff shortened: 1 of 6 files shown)


fredaan:
## Java review (WBS-029, bug 76502)

Re-verified the design (`01-design.md`), design-review's one required fix (`02-design-review.md`), and the implementation account (`03-implement.md`) against the actual PR diff — not the accounts alone.

**Confirmed correct:**
- `EditRequest.variableValues` retype (`Map<String,String>` → `Map<String,Object>`) applied consistently to the canonical record component and all 7 compatibility constructors.
- Design-review's required fix — `WorksheetAgentControllerTest.setVariableValuesRequest` helper retyped to `Map<String, Object>` — is applied.
- Re-grepped the whole blast radius myself (not trusting prior counts): every other `new EditRequest(...)` call site passes a `null` literal for `variableValues` (compiles under either type). `ViewsheetEditServiceTest.java`'s many `new EditRequest(...)` calls resolve to an unrelated same-named class in package `inetsoft.web.wiz.viewsheet` — confirmed harmless, as the design-review guessed.
- **Fail-fast ordering** (the specific thing I was asked to check): for a 2+-element array, the display-style gate (`LIST`/`CHECKBOXES` check) throws *before* any per-element type/choice validation runs — so a 2+-element array on a combobox variable always gets the clear "this variable isn't multi-select" error, never a confusing per-token error. This is actually a small improvement over the design's literal stated order (design put per-element type-checking first), not a bug.
- Per-token validation checks `declared.getValues()` (the values, not `getChoices()`/labels) — correct, matches the WBS-030 sibling's labels/values-inversion finding.
- `vtable.setAsIs(vname, true)` is unconditional for genuine (2+) array values, as designed and as the design-review independently confirmed safe (`Condition.setDynamicValue` only consults `isAsIs` for `op == ONE_OF`).
- `displayStyleName` relocation to `WorksheetMutationSupport` (package-visible) is clean — fetched `origin/stylebi-wiz-main-rebase` fresh to confirm WBS-030/#5063's already-merged version of `WorksheetReadService.java` is what got removed, byte-for-byte, no duplicate left, no merge artifact.
- Spot-checked test assertions rather than trusting the claimed list: the "bad token named specifically" test asserts the message contains the bad token and does **not** contain the good one; the multi-value success tests assert a real `Object[]` + `isAsIs()==true`, not just "no exception."
- `EditRequestParametersDeserializationTest` uses the file's real `WebConfig`-configured `ObjectMapper` (not a bare `new ObjectMapper()`), and genuinely asserts array→`List<String>` vs string→`String` binding — the concrete unit-test proof for the design's flagged Jackson-binding risk.

**Non-blocking observations** (already independently flagged by this repo's own `claude-review` CI bot, and I agree these don't rise to blocking): `asIs=true` unconditional slightly overstates "mirrors the native dialog" in prose (safe, just imprecise wording); a bare unquoted JSON number in `variableValues` now hard-rejects instead of silently coercing (stricter, no known caller relies on the old behavior); `null` inside a 1-element list is rejected while a top-level scalar `null` clears the variable (intentional, documented).

**Verdict: clean, no blocker/important findings.** Full review: `docs/teams/2026-09-08-bugs-76502-wbs-gaps/bug-wbs029/04-review-java-r1.md` in stylebi-wiz.

Note: the Maven `build` CI check was still pending at review time — worth confirming green before merge. `gh pr review --approve` isn't usable here (self-review, same account that opened the PR) — flagging clean via comment for a human/lead to merge.

PR #5067, opened 2026-09-08
fix(admin-chat): reject AT-type schedule condition with null date at preview

lizhepu:
## Summary
- On a genuine StyleBI Enterprise deployment, `preview_schedule_task_changes` accepted a `create` plan with an `AT`-type time condition whose `date` was `null`, producing a valid-looking plan/hash.
- `apply_schedule_task_changes` then failed inside `AdminScheduleGateway.convertCondition` (`"ISO 8601 Date is required for \"AT\" type"`), and the generic apply exception handling in `ScheduleChangesetApplyService` couldn't distinguish this pre-write validation failure from an ambiguous in-flight write, so it reported the whole changeset as `rollback-failed` — even though nothing was ever actually written to storage.
- Fix: `ScheduleChangePlanService.requireSupportedConditions` now also rejects a structurally-invalid `AT` condition (type `AT`, `date == null`) with a clear, field-named `IllegalArgumentException`, closing the gap at preview time — the same treatment this method already gives the "only time conditions are supported" and unsupported-action checks.
- Deliberately narrow: does not touch `ScheduleChangesetApplyService`'s generic apply/rollback classification engine, and does not touch `AdminChangesetApplyService` (properties area's analog) — both are out of scope per the task.

## Test plan
- [x] `./mvnw -pl core test-compile` (compiles clean)
- [x] `./mvnw -pl core test -Dtest=ScheduleChangePlanServiceTest` — 21/21 pass (19 existing + 2 new: `resolveCreateThrowsOnAtConditionWithNullDate`, `resolveCreateSucceedsWithAtConditionAndNonNullDate`)

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/schedule/ScheduleChangePlanService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/schedule/ScheduleChangePlanService.java
@@ -234,6 +234,14 @@ public class ScheduleChangePlanService {
                label + ".spec.conditions[" + i + "]: only time conditions are supported in this " +
                "area (got " + conditions.get(i).getClass().getSimpleName() + ")");
          }
+
+         TimeCondition tc = (TimeCondition) conditions.get(i);
+
+         if(tc.getType() == TimeCondition.Type.AT && tc.getDate() == null) {
+            throw new IllegalArgumentException(
+               label + ".spec.conditions[" + i + "].date: an ISO 8601 date is required for " +
+               "\"AT\"-type conditions");
+         }
       }
    }
 
--- a/core/src/test/java/inetsoft/web/admin/ai/schedule/ScheduleChangePlanServiceTest.java
+++ b/core/src/test/java/inetsoft/web/admin/ai/schedule/ScheduleChangePlanServiceTest.java
@@ -129,6 +129,35 @@ class ScheduleChangePlanServiceTest {
       assertTrue(ex.getMessage().contains("conditions"));
    }
 
+   // bug: an "AT"-type condition with a null date sailed through preview and only failed at apply
+   // time inside AdminScheduleGateway, where the generic exception handling misclassified the
+   // failure as rollback-failed even though nothing had been written -- catch it here instead.
+   @Test void resolveCreateThrowsOnAtConditionWithNullDate() {
+      CreateScheduleTaskRequest spec = createSpec("t1", "admin");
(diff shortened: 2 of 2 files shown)


lizhepu (commented):
Reviewed the diff and ran the tests locally against the exact PR head (656abb220) from `E:\inetsoft\wizard\community`.

**Correctness of the new check** (`ScheduleChangePlanService.requireSupportedConditions`):
- The new null-date check is scoped correctly: `tc.getType() == TimeCondition.Type.AT && tc.getDate() == null`. Since `TimeCondition.type` has no default (null when unset) and every other `Type` enum value (`EVERY_DAY`, `EVERY_WEEK`, `DAY_OF_WEEK`, etc.) fails the `== Type.AT` comparison, non-AT conditions and conditions with an unset type are unaffected — no risk of false-positive rejection.
- Placement is correct: it runs inside the existing per-condition loop, immediately after the `instanceof TimeCondition` check, reusing that already-verified cast rather than duplicating the instanceof check. No behavior change for the pre-existing "only time conditions are supported" / unsupported-action checks.

**Tests** (`ScheduleChangePlanServiceTest`):
- `resolveCreateThrowsOnAtConditionWithNullDate` asserts the exception message contains both `"date"` and `"AT"` — a real content check, not just "throws something."
- `resolveCreateSucceedsWithAtConditionAndNonNullDate` is a genuine positive control: it builds an AT condition with `OffsetDateTime.now()`, mocks the rest of the create path (no existing task, delete permission present), and asserts `resolve()` returns a 1-change plan — confirming the new check doesn't over-reject a valid AT condition.
- Ran `./mvnw -pl core test -Dtest=ScheduleChangePlanServiceTest` myself: **21/21 pass, 0 failures, 0 errors** (surefire report confirms 19 existing + 2 new), matching the PR description.

**Scope discipline**: diff touches only `ScheduleChangePlanService.java` (+8) and `ScheduleChangePlanServiceTest.java` (+36) — nothing in `ScheduleChangesetApplyService.java`, `AdminChangesetApplyService.java`, or any other area's apply/rollback logic. Consistent with the PR's stated deliberately-narrow scope (the deeper apply-time exception-classification fix is correctly left out of scope).

No blocking issues found. I could not `--approve` since this PR is authored by the same account running this review (GitHub disallows self-approval) — leaving to a different reviewer/the requester to approve. Not merging per instructions. Note: several CI checks were still in progress/queued at review time — build, claude-review, Semgrep — worth confirming those go green before merge.

PR #5069, opened 2026-09-08
feat(tabular): Rest.XML catalog SPI, describeDataset-only via opaque endpoint token

Zxm-happy:
## Summary

- `RestXMLRuntime` implements `TabularCatalogProvider`, but unlike every other implementer so far, only `describeDataset` does real work. `Rest.XML`'s structural properties (URL suffix, XPath, columns) live on `RestXMLQuery`, not on the bare-`URL` `RestXMLDataSource`, so there is no server-side resource list `listDatasets` could ever enumerate — it throws a named "does not support enumeration" exception instead of faking one.
- `describeDataset` treats its `datasetId` argument as an **opaque, self-describing token** (base64 + JSON: `version`/`suffix`/`xpath`/`columns`) rather than a server-addressable resource name, decoded and validated with no network access.
- This is reachable in practice because `TabularCatalogService.describeTable` never depends on `listDatasets` having succeeded (`resolveProvider` is a bare `instanceof` check), and wiz's single-target annotation path bypasses the whole-database `listDatasets`/`/datasource/tables` call entirely — verified by reading both call chains, not assumed.
- A future wiz-side pipeline (LLM extraction from uploaded/linked API documentation, out of scope here) is what would actually produce one of these tokens; this round implements and proves only the StyleBI-side contract, against hand-built tokens.
- 61 new tests (`RestXmlEndpointTokenCodecTest`, `RestXmlEndpointCatalogTest`, `RestXMLRuntimeTest`) plus one line in `TabularCatalogProviderImplementerCanaryTest`. `RestXMLQuery.java`/`RestXMLDataSource.java` untouched; `RestXMLRuntime.java` gains only the two new interface methods.
- Full design rationale: `stylebi-wiz` repo, `docs/tabular/rest-xml-catalog-spi-feasibility.md`. Full process record (charter/design/reconcile/build/verify/review/retro): `stylebi-wiz` repo, `docs/teams/2026-09-08-rest-xml-catalog-describe-only/`.

## Test plan

- [x] `./mvnw -pl connectors/inetsoft-rest -am test -Dtest=RestXmlEndpointTokenCodecTest,RestXmlEndpointCatalogTest,RestXMLRuntimeTest` — 61/61 pass
- [x] `./mvnw -pl core test -Dtest=TabularCatalogProviderImplementerCanaryTest` — 3/3 pass
- [x] `git diff` confirms `RestXMLQuery.java`/`RestXMLDataSource.java` byte-for-byte unchanged
- [x] No live/network verification this round (no real REST XML server, no real upstream token producer yet) — unit-test-only against hand-built tokens, stated explicitly in the process record

### DIFF
--- a/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/xml/RestXMLRuntime.java
+++ b/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/xml/RestXMLRuntime.java
@@ -20,12 +20,40 @@ package inetsoft.uql.rest.xml;
 import inetsoft.uql.rest.AbstractRestQuery;
 import inetsoft.uql.rest.AbstractRestRuntime;
 import inetsoft.uql.rest.QueryRunner;
+import inetsoft.uql.tabular.TabularCatalog;
+import inetsoft.uql.tabular.TabularCatalogProvider;
+import inetsoft.uql.tabular.TabularDataSource;
+import inetsoft.uql.tabular.TabularDatasetSchema;
 
-public class RestXMLRuntime extends AbstractRestRuntime {
+public class RestXMLRuntime extends AbstractRestRuntime implements TabularCatalogProvider {
    @Override
    protected QueryRunner getQueryRunner(AbstractRestQuery q) {
       final RestXMLQuery query = (RestXMLQuery) q;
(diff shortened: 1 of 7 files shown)


jshobe-inetsoft (approved):
Claude's automated review found no blocking issues; approving based on its analysis.

PR #5070, opened 2026-09-08
fix(worksheet): reject type-incompatible edit_unpivot header shrink (bug 76517, WBS-036)

lizhepu:
## Summary
- `WorksheetEditService.editUnpivot` had zero validation before `setHeaderColumns` (unlike `addUnpivot`). Every unpivot execution re-derives the melt from scratch off the pre-unpivot source table (`AssetUtil.unpivot`), so shrinking `headerColumns` onto a type-incompatible column silently coerced the melted column to STRING and injected one phantom row per source row (the departing column's own name landing in the Dimension slot).
- Adds `addUnpivot`'s bound check (`headerColumns < 0 || >= colCount`) to `editUnpivot`.
- Adds a shrink-direction type-compatibility check, `AssetUtil.checkUnpivotShrinkTypeConflict`, mirroring `AssetUtil.unpivot`'s own `sameType`/`allNumber` rule (shared, not duplicated, so the "does the cutoff succeed" and "what happens when it runs" decisions can't disagree). Throws naming the offending column and its type, e.g. `"CATEGORY_NAME cannot move from header to melted columns: it is type 'string' but the melted columns are type 'double'."`
- Applies the identical guard to the Composer UI's `TableUnpivotDialogService.changeUnpivotTableRowHeaders`, which shares the exact same defect for human users, not just the agent API.
- Growing `headerColumns` is deliberately left unguarded — provably safe from this specific defect by the melt loop's own bounds (see root-cause doc).

## Root cause
`editUnpivot` (`WorksheetEditService.java`) set `UnpivotTableAssemblyInfo.headerColumns` with no bounds or type check. `UnpivotQuery.getPostBaseTableLens` always re-executes the melt over the *original* pre-unpivot source table via `AssetUtil.unpivot(TableLens, int hcol)`, which coerces the shared melted column to STRING whenever the columns in `[hcol, colCount)` aren't homogeneous/all-numeric, and always emits one row per melt-column-index per source row (including the newly-exposed `hcol` position, whose header name becomes phantom "data").

## Test plan
- [x] `./mvnw.cmd -pl core -o -Dtest=WorksheetEditServiceMutatorsTest#editUnpivotRejectsOutOfRangeHeaderColumns+editUnpivotRejectsTypeIncompatibleShrink+editUnpivotAllowsTypeCompatibleShrink test -DfailIfNoTests=false` — 3/3 pass
- [x] `./mvnw.cmd -pl core -o -Dtest=WorksheetEditServiceMutatorsTest,WorksheetEditServiceTest test -DfailIfNoTests=false` — 268/268 pass (no regressions in the shared file)
- [ ] No dedicated test file exists for `TableUnpivotDialogService` (Composer UI controller); the guard there reuses the same `AssetUtil.checkUnpivotShrinkTypeConflict` predicate already covered by the tests above. Live re-verification via the Composer UI is deferred (see companion plugin PR / fix doc — the shared StyleBI deployment is currently unreachable).

Related: companion plugin-side defense-in-depth PR in `inetsoft-technology/stylebi-wiz` (`edit_unpivot` read-back assertion).

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java
+++ b/core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java
@@ -2720,6 +2720,75 @@ public class AssetUtil {
    }
 
    /**
+    * Checks whether shrinking an unpivot's header column count from {@code oldHcol} to a
+    * smaller {@code newHcol} would move a column into the melted range whose type is
+    * incompatible with it, using the exact {@code sameType}/{@code allNumber} rule
+    * {@link #unpivot(XTable, int)} itself applies once the melt actually executes. This lets a
+    * caller reject the shrink up front (using the source table's declared column types) instead
+    * of letting {@code unpivot} silently coerce the melted column to {@code STRING} and inject
+    * one phantom "column name as data" row per source row.
+    *
+    * @param cs      the pre-unpivot source table's column selection, in column order
+    * @param oldHcol the current header column count
+    * @param newHcol the proposed header column count
+    *
+    * @return {@code null} if the shrink is type-safe (or not actually a shrink), otherwise a
+    *         message naming the first offending column and the conflicting types
+    */
+   public static String checkUnpivotShrinkTypeConflict(ColumnSelection cs, int oldHcol, int newHcol) {
+      if(newHcol >= oldHcol) {
+         return null;
+      }
+
+      int colCount = cs.getAttributeCount();
+      String[] types = new String[colCount];
+
+      for(int i = Math.max(0, newHcol); i < colCount; i++) {
+         DataRef ref = cs.getAttribute(i);
+         types[i] = ref instanceof ColumnRef ? ((ColumnRef) ref).getDataType() : XSchema.STRING;
+      }
+
+      boolean sameType = true;
+      boolean allNumber = true;
+
+      for(int i = newHcol + 1; i < colCount; i++) {
+         String t0 = types[i - 1];
+         String t1 = types[i];
+
+         if(!XSchema.isNumericType(t0) || !XSchema.isNumericType(t1)) {
+            allNumber = false;
+         }
+
+         if(!Objects.equals(t0, t1)) {
+            sameType = false;
+         }
+      }
+
+      if(sameType || allNumber) {
+         return null;
+      }
+
+      // The melt is not incremental (it always recomputes over [newHcol, colCount)), but the
+      // columns already at [oldHcol, colCount) are the ones the caller expects to keep their
+      // current type, so they are what a departing column is judged against.
+      String meltType = oldHcol < colCount ? types[oldHcol] : null;
+
+      for(int i = newHcol; i < oldHcol; i++) {
+         String type = types[i];
+         boolean compatible = Objects.equals(type, meltType) ||
+            (XSchema.isNumericType(type) && XSchema.isNumericType(meltType));
+
+         if(!compatible) {
+            String name = cs.getAttribute(i).getName();
+            return name + " cannot move from header to melted columns: it is type '" + type +
+               "' but the melted columns are type '" + meltType + "'.";
+         }
+      }
+
+      return null;
+   }
+
+   /**
     * Unpivot: header row is changed to a column named "Dimension", and crosstab
     * cells are changed to a column called "Measure".
     */
--- a/core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java
+++ b/core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java
@@ -2741,9 +2741,18 @@ public class AssetUtil {
       }
 
       int colCount = cs.getAttributeCount();
+
+      // An out-of-range newHcol isn't this helper's bound-check to enforce -- that's the
+      // caller's job (e.g. WorksheetEditService.editUnpivot's own headerColumns check) -- but it
+      // also isn't a type conflict this helper can meaningfully answer, so treat it as no
+      // conflict rather than indexing types[] with it below.
+      if(newHcol < 0 || newHcol >= colCount) {
+         return null;
+      }
+
       String[] types = new String[colCount];
 
-      for(int i = Math.max(0, newHcol); i < colCount; i++) {
+      for(int i = newHcol; i < colCount; i++) {
          DataRef ref = cs.getAttribute(i);
          types[i] = ref instanceof ColumnRef ? ((ColumnRef) ref).getDataType() : XSchema.STRING;
       }
--- a/core/src/main/java/inetsoft/web/composer/ws/dialog/TableUnpivotDialogService.java
+++ b/core/src/main/java/inetsoft/web/composer/ws/dialog/TableUnpivotDialogService.java
@@ -112,20 +112,30 @@ public class TableUnpivotDialogService extends WorksheetControllerService {
          return null;
       }
 
+      TableAssembly src = table.getTableAssembly();
+      int colCount = src == null ? 0 : src.getColumnSelection(false).getAttributeCount();
+
+      // Same bound check as the agent API's editUnpivot (bug 76517/WBS-036). Unlike editUnpivot,
+      // this Composer UI call site had no validation on the level value it receives from
+      // WSUnpivotDialogEvent at all, so an out-of-range level fell straight through to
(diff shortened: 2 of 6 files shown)


lizhepu:
**Review round 1 (fix/wbs-036, bug 76517) -- CHANGES REQUESTED**

(Posted as a comment, not a formal review: GitHub blocks "request changes" reviews on your own PR,
and this PR's author is my own account.)

What I checked: re-read AssetUtil.unpivot's own sameType/allNumber loop against the new
checkUnpivotShrinkTypeConflict to confirm the new guard genuinely mirrors it (it does, using
declared ColumnSelection types instead of a materialized TableLens's runtime types -- an
acceptable approximation given add_unpivot/the rest of this code path already treats declared
dataType as authoritative, not a blocker); confirmed growing headerColumns is correctly left
unguarded (loop-bound analysis in 02-refute.md holds, and the diff only gates the `< oldHeaderColumns`
shrink branch in both call sites); confirmed no scope creep into AssetUtil.unpivot's CSV/Excel
import callers or TableUnpivotDialogService.createUnpivotTable's pre-existing dead `colLength`;
read all three new WorksheetEditServiceMutatorsTest tests body-by-body (they assert what they
claim: names the offending column + both types on a rejected shrink, allows an already-numeric
shrink, rejects out-of-range bounds); ran them myself from a disposable worktree off this PR's
branch (475752d9a) via `./mvnw.cmd -pl core -o -Dtest=WorksheetEditServiceMutatorsTest test` --
242/242 green, no regressions.

**Blocking:** `TableUnpivotDialogService.changeUnpivotTableRowHeaders` (TableUnpivotDialogService.java
~107-124) calls the new `AssetUtil.checkUnpivotShrinkTypeConflict` without first bound-checking
`newHeaderColumns`, unlike `WorksheetEditService.editUnpivot`, which validates `headerColumns < 0`
and throws *before* ever calling the helper. `WSUnpivotDialogEvent`/`TableUnpivotDialogModel.level`
has no validation anywhere (plain `int`, no `@Min`, no upstream check in the controller). The
helper itself only clamps its type-array-fill loop to `Math.max(0, newHcol)`; the subsequent
sameType/allNumber comparison loop starts at the unclamped `newHcol + 1`, so `newHcol = -1`
indexes `types[-1]`.

Demonstrated live, not just reasoned: added a throwaway JUnit probe calling
`AssetUtil.checkUnpivotShrinkTypeConflict(cs, 2, -1)` against a 4-column ColumnSelection
(string, string, double, double) -- threw exactly:
```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 4
	at inetsoft.uql.asset.internal.AssetUtil.checkUnpivotShrinkTypeConflict(AssetUtil.java:2755)
```
So any `/ws/dialog/table-unpivot-rowHeaders` request with a negative `level` on a realistic
unpivot table (dimension header next to numeric measures -- i.e. most of them) throws an uncaught
`ArrayIndexOutOfBoundsException` instead of the intended friendly `MessageException`, for the exact
Composer UI call site this PR's own description says gets "the same protection as the agent API."
Fix: either bound-check `newHcol`/`oldHcol` defensively inside `checkUnpivotShrinkTypeConflict`
itself (protects both callers regardless of caller discipline), or add the same
`headerColumns < 0 || >= colCount` guard to `changeUnpivotTableRowHeaders` before it calls the
helper, matching `editUnpivot`.

Everything else in the diff looks correct and in-scope. Please fix the above and I'll do another
round.

lizhepu:
Reproduced and fixed. Root cause confirmed: `checkUnpivotShrinkTypeConflict`'s fill loop clamped its start to `Math.max(0, newHcol)`, but the `sameType`/`allNumber` comparison loop right after it started at the unclamped `newHcol + 1`, so a negative `newHcol` indexed `types[-1]` -> `ArrayIndexOutOfBoundsException`.

Fix (commit 6a17ca103):
- `AssetUtil.checkUnpivotShrinkTypeConflict` now has an explicit early guard: `newHcol < 0 || newHcol >= colCount` returns `null` (no conflict) instead of falling into the indexing loops. Rationale: bound validation is the caller's responsibility (each call site should reject and message on it), and an out-of-range `newHcol` isn't a type-conflict question this helper can meaningfully answer -- but the helper itself must never crash if a caller's guard is missing or wrong, so "no conflict" (not an exception) is the correct defensive return.
- `TableUnpivotDialogService.changeUnpivotTableRowHeaders` now validates `newHeaderColumns` against `[0, colCount)` up front and throws `MessageException`, mirroring `WorksheetEditService.editUnpivot`'s existing guard -- this was the specific gap you found: the Composer UI call site had no bound check at all before reaching the helper.
- Added regression tests: `AssetUtilTest` exercises the helper directly with `newHcol = -1` and `newHcol >= colCount`; `TableUnpivotDialogServiceTest` (new file) exercises `changeUnpivotTableRowHeaders` end to end for both cases and asserts a clean `MessageException`, not a crash.

`./mvnw.cmd -pl core -o -Dtest=WorksheetEditServiceMutatorsTest,WorksheetEditServiceTest,AssetUtilTest,TableUnpivotDialogServiceTest test` -> 277/277 passing, no regressions.

lizhepu:
**Review round 2 (fix/wbs-036, bug 76517) -- LOOKS GOOD, ready to merge**

(Posted as a comment, not a formal review: same self-PR restriction as round 1.)

What I checked this round:

1. **Design judgment on `checkUnpivotShrinkTypeConflict` returning `null` for out-of-range `newHcol`
   (commit 6a17ca103)**: evaluated whether this could let a future bound-check-less caller silently
   treat an invalid `newHcol` as "proceed." It can't cause new harm: `AssetUtil.unpivot(XTable, int)`
   itself already clamps `hcol = Math.max(0, Math.min(hcol, row0.length - 1))` at execution time, so
   even a hypothetical caller that skipped its own bound check and got `null` back from the helper
   would still hit that pre-existing clamp downstream, not silent corruption. Both actual callers
   (`editUnpivot`, `changeUnpivotTableRowHeaders`) now bound-check *before* calling the helper
   anyway, so this is genuinely just defense-in-depth. Reasonable call.
2. **`changeUnpivotTableRowHeaders`'s new bound check** (TableUnpivotDialogService.java:120-125) is
   byte-for-byte the same condition as `editUnpivot`'s (WorksheetEditService.java:3177):
   `newHeaderColumns < 0 || newHeaderColumns >= colCount`. Confirmed side by side.
3. **New tests**: `AssetUtilTest`'s two new cases hit the exact round-1 crash inputs
   (`newHcol = -1`, `newHcol == colCount`). `TableUnpivotDialogServiceTest`'s reflection-based event
   construction is correct -- verified `WSUnpivotDialogEvent.model` and
   `TableUnpivotDialogModel.level` are the actual field names Jackson populates (checked both source
   files directly), so the reflection isn't silently missing the field and falling through to a
   default `level = 0`.
4. **Ran the tests myself**:
   `./mvnw.cmd -pl core -o -Dtest=WorksheetEditServiceMutatorsTest,WorksheetEditServiceTest,AssetUtilTest,TableUnpivotDialogServiceTest test`
   -> 277/277 passing, matches the claim.
5. **Tried to break it further**: probed a disposable JUnit test with an out-of-range `oldHcol`
   (`checkUnpivotShrinkTypeConflict(cs, 100, 0)` against a 4-column selection) to see if the
   *second* loop (`for(i = newHcol; i < oldHcol; i++) { types[i]... }`, which is bounded by `oldHcol`
   rather than `colCount`) could still index out of bounds the same way the round-1 bug did. It
   doesn't: when `oldHcol >= colCount`, `meltType` becomes `null`, and since
   `ColumnRef.getDataType()` never actually returns `null` (falls back to `XSchema.STRING`), the
   very first iteration's type-compatibility check against `meltType == null` always fails and
   returns early, before `i` can ever reach `colCount`. Deleted the probe afterward; worktree is
   clean.

No new blocking issues found. Nice catch-and-fix on round 1's finding, and the two new test files
are solid regression coverage.

One side note, not blocking this review: the automated `claude-review` GitHub check is currently
showing a bot execution error ("Claude encountered an error after 0s"), not an actual code finding
-- worth a re-run, but not something to hold this PR on. `build`, `Semgrep Scan`, `Semgrep OSS`, and
`license/cla` are all green and the PR is mergeable (no conflicts).

Round 2 verdict: ready to merge, pending the lead's own review/merge.

PR #5071, opened 2026-09-08
fix(wiz): surface RuntimeExceptions in WorksheetPreviewService as PairingException (bug 76517, WBS-038)

lizhepu:
## Summary

- Bug 76517 (WBS-038): `preview_worksheet_data` on a worksheet aggregate (0 groups, two ungrouped `CountDistinct` aggregates on different columns) raised a raw, un-prefixed `NullPointerException` instead of an actionable error.
- `WorksheetPreviewService.preview()` already wraps `box.getTableLens(...)` in a try/catch that re-throws as `PairingException`, but everything after that — `lens.getColCount()`, the header loop, and the row loop — had no exception handling at all. Any `RuntimeException` thrown there (including the reported NPE) propagated straight to the caller as an implementation-detail leak.
- Fix: wrap that block in a try/catch that converts any `RuntimeException` into `PairingException("Failed to read result columns for '<table>': <message>")`. No other behavior changes.
- Applied the identical fix shape to the sibling defect in `RawDataService.writeWorksheetTableCsvStream` (same unguarded-`getTableLens()`-then-immediate-use shape, same class of bug, different endpoint), including a null-lens guard mirroring `WorksheetPreviewService`'s own null-lens handling.

## Important: this does NOT fix the underlying root cause

This PR converts an unguarded crash into a clear, actionable error. It does **not** identify or fix why `getTableLens()` returns a `TableLens` whose internal delegate is null for this two-`CountDistinct`-aggregate shape in the first place.

Debug-workflow history for this bug (diagnosis + adversarial refutation) landed a **NOT ESTABLISHED** verdict on the deep root cause:
- A live JUnit repro against an in-memory `EmbeddedTableAssembly` (Java-side, non-SQL-merge execution path) with the exact same `AggregateInfo` shape (0 groups, 2 `CountDistinct` aggregates on different columns) computed correctly with **no exception** — this narrows the likely defect location toward the SQL-merge-and-result-wrapping path (`PreAssetQuery`'s JDBC push-down, reached only for a Postgres-backed `BoundTableAssembly`), but does not confirm it. No live JDBC-backed repro could be built this session.
- The diagnosis's proposed deeper fix — wiring a multi-distinct-count merge guard into `PreAssetQuery.isAggregateInfoMergeable0()`, mirroring `MVTransformer.isCombinable()` — was **deliberately not implemented** here. The refutation raised a substantive objection: MV combinability (merging *partial per-block* results from a materialized view) and ordinary single-query JDBC push-down mergeability are different concerns that only superficially resemble each other. Applying that fix as specified would force every dialect (including Postgres, where the merged SQL is valid and correct) off the merged-SQL path unconditionally for 2+ `CountDistinct` aggregates — a behavior change to a shared method used by every JDBC-mergeable worksheet/viewsheet query, based on an unconfirmed analogy.

**Recommendation**: pursue the deep root cause as a separate follow-up, with a live Postgres-backed repro (debugger or full-stack-trace logging in `AssetQuerySandbox.getTableLens()`'s catch blocks) to pin the exact `TableLens` subclass, before touching `PreAssetQuery`'s shared merge-decision logic.

## Test plan

- [x] `WorksheetPreviewServiceTest`: added `throwsPairingExceptionWhenGetColCountThrows` and `throwsPairingExceptionWhenRowExtractionThrows`, using a mock `TableLens` that throws mid-extraction, asserting the exception now surfaces as a `PairingException` naming the table (does not reproduce the exact reported NPE — its root cause is unpinned — but proves the narrower, confirmed-safe claim).
- [x] Ran `./mvnw.cmd -pl core -o -Dtest=WorksheetPreviewServiceTest test -DfailIfNoTests=false` from the worktree root — 11/11 tests pass (9 pre-existing + 2 new).

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/RawDataService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/RawDataService.java
@@ -85,10 +85,21 @@ public class RawDataService {
          throw new RuntimeException("Table " + tableName + " not found.");
       }
 
-      TableLens lens =
-         box.getTableLens(targetTable.getAbsoluteName(), AssetQuerySandbox.RUNTIME_MODE);
+      TableLens lens;
 
-      writeCsvContent(lens, outputStream);
+      try {
+         lens = box.getTableLens(targetTable.getAbsoluteName(), AssetQuerySandbox.RUNTIME_MODE);
+
+         if(lens == null) {
+            throw new RuntimeException("Table " + tableName + " produced no data.");
+         }
+
+         writeCsvContent(lens, outputStream);
+      }
+      catch(RuntimeException e) {
+         throw new RuntimeException(
+            "Failed to export worksheet table '" + tableName + "': " + e.getMessage(), e);
+      }
    }
 
    public void writeDataSourceTableCsvStream(ExportDatabaseTableToCsvRequest requestData,
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetPreviewService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetPreviewService.java
@@ -99,29 +99,35 @@ public class WorksheetPreviewService {
          throw new PairingException("Table not found or produced no data: " + tableName);
       }
 
-      int colCount = lens.getColCount();
+      try {
+         int colCount = lens.getColCount();
 
-      // Row 0 is the header row in StyleBI's TableLens convention.
-      String[] headers = new String[colCount];
-
-      for(int col = 0; col < colCount; col++) {
-         Object h = lens.getObject(0, col);
-         headers[col] = h != null ? h.toString() : "col" + col;
-      }
-
-      List<Map<String, Object>> rows = new ArrayList<>();
-
-      for(int row = 1; rows.size() < limit && lens.moreRows(row); row++) {
-         Map<String, Object> rowMap = new LinkedHashMap<>();
+         // Row 0 is the header row in StyleBI's TableLens convention.
+         String[] headers = new String[colCount];
 
          for(int col = 0; col < colCount; col++) {
-            rowMap.put(headers[col], toJsonSafe(lens.getObject(row, col)));
+            Object h = lens.getObject(0, col);
+            headers[col] = h != null ? h.toString() : "col" + col;
          }
 
-         rows.add(rowMap);
-      }
+         List<Map<String, Object>> rows = new ArrayList<>();
 
-      return rows;
+         for(int row = 1; rows.size() < limit && lens.moreRows(row); row++) {
+            Map<String, Object> rowMap = new LinkedHashMap<>();
+
+            for(int col = 0; col < colCount; col++) {
+               rowMap.put(headers[col], toJsonSafe(lens.getObject(row, col)));
+            }
+
+            rows.add(rowMap);
+         }
+
+         return rows;
+      }
+      catch(RuntimeException e) {
+         throw new PairingException("Failed to read result columns for '"
+                                    + tableName + "': " + e.getMessage());
+      }
    }
 
    /**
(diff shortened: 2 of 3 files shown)


lizhepu:
Round 1 review (posted as a comment, not a formal review — GitHub blocks `gh pr review --approve/--request-changes` on your own PR, and the authenticated `gh` identity here is the same as the PR author).

**Verdict: changes requested — one blocking issue found.**

Checked: scope (diff touches only `RawDataService.java`, `WorksheetPreviewService.java`, `WorksheetPreviewServiceTest.java` — confirmed no changes to `PreAssetQuery`/`MVTransformer`/`SQLHelper`, matching the stated narrow scope), the try/catch boundaries in `WorksheetPreviewService.preview()` (covers `getColCount()` + header loop + row loop + return, matches the described original lines 102-122), the `RawDataService` sibling fix (matches that class's plain-`RuntimeException` convention, cause preserved there), the two new `WorksheetPreviewServiceTest` cases (mock `TableLens`, assert exact `PairingException` type + message), and independently re-ran `./mvnw.cmd -pl core -o -Dtest=WorksheetPreviewServiceTest test -DfailIfNoTests=false` from the worktree — 11/11 pass, confirmed. The PR description is honest about the unpinned root cause and recommends a follow-up.

**Blocking:**
- `WorksheetPreviewService.java:127-130` — the new `catch(RuntimeException e) { throw new PairingException("Failed to read result columns for '" + tableName + "': " + e.getMessage()); }` drops the original exception. `PairingException` already has a `(String message, Throwable cause)` constructor (`PairingException.java:62-65`) that this call doesn't use. Since the root cause of the underlying NPE is still unpinned and a future investigation depends on a real stack trace (per this PR's own "Open root cause" note), losing the cause here discards exactly the information that follow-up will need. Please change to `throw new PairingException("Failed to read result columns for '" + tableName + "': " + e.getMessage(), e);`.

Non-blocking: the pre-existing `catch(Exception e)` around `box.getTableLens()` a few lines above (78-81) has the same drop-the-cause shape, but it predates this PR — flagging only as a related note, not a blocker for this PR.

lizhepu:
Fixed in 218665cb1: `WorksheetPreviewService.preview`'s `catch(RuntimeException e)` around result-column extraction now throws `new PairingException(msg, e)`, preserving the original cause.

Verified `RawDataService.writeWorksheetTableCsvStream`'s sibling catch block (`RawDataService.java:99-102`) already does `new RuntimeException(msg, e)` — no change needed there, as you noted.

Left the pre-existing `catch(Exception e)` around `box.getTableLens()` (~lines 78-81) untouched per your note — out of scope for this PR.

Extended `throwsPairingExceptionWhenGetColCountThrows` and `throwsPairingExceptionWhenRowExtractionThrows` in `WorksheetPreviewServiceTest` to assert `assertSame(cause, ex.getCause())`. All 11 tests in that file pass.

lizhepu:
## Round 2 review

Checked the new commit `218665cb1` against round 1's finding.

**1. Fix content** — confirmed `WorksheetPreviewService.java:128` now reads exactly
`new PairingException(msg, e)`, preserving the caught `RuntimeException` as the cause. No
other logic changed (verified against the full PR diff vs `stylebi-wiz-main-rebase`: pure
try/catch wrapping around the same header/row extraction code, same limit/return semantics).

**2. `RawDataService.writeWorksheetTableCsvStream`** — re-read the source directly; its
catch block already does `new RuntimeException(msg, e)`, cause preserved. Matches the
fixer's claim, no regression.

**3. Test coverage for the cause** — both `throwsPairingExceptionWhenGetColCountThrows` and
`throwsPairingExceptionWhenRowExtractionThrows` were extended with
`assertSame(cause, ex.getCause())`, checking against the actual original exception object
(not just "some exception was thrown"). This is a real regression guard.

**4. Test run** — ran `./mvnw.cmd -pl core -o -Dtest=WorksheetPreviewServiceTest test
-DfailIfNoTests=false` from the worktree directly: `Tests run: 11, Failures: 0, Errors: 0,
Skipped: 0` — BUILD SUCCESS.

**5. Full-diff pass** — `git diff origin/stylebi-wiz-main-rebase...HEAD --stat` shows exactly
3 files touched (`WorksheetPreviewService.java`, `RawDataService.java`,
`WorksheetPreviewServiceTest.java`), matching round 1's confirmed-clean scope. No touches to
`PreAssetQuery`/`MVTransformer`/`SQLHelper`, no dead code, no unrelated refactors.

**Verdict: round 2 is clean.** Round 1's finding is resolved; no new blocking issues found.
This PR is ready to merge as far as this review is concerned.

PR #5072, opened 2026-09-08
fix(wiz): removeJoin refuses removal when another assembly depends on it (bug 76517 WBS-037)

lizhepu:
## Summary

- `WorksheetEditService.removeJoin(String)` was a bare `ws.removeAssembly(name)` with no
  existence check and no `AssetEventUtil.hasDependent` guard, unlike its sibling
  `deleteTable`, which has both.
- Removing a join that a mirror (or other assembly) was built on left the dependent
  referencing a name that no longer existed — every subsequent query against it failed,
  with no tool able to repair it. Reported as Redmine #76517 (WBS-037).
- Fix: `removeJoin` now throws `PairingException` if the assembly doesn't exist, and if
  another assembly depends on it (via the same generic `AssetEventUtil.hasDependent` check
  `deleteTable` already uses), naming the dependent relationship in the message.
- The `remove_join` dispatch case in `WorksheetAgentController` already sits in a method
  declared `throws Exception` (same switch as `delete_table`), so no additional wiring was
  needed — confirmed by building, not just reading.

## Root cause

An incomplete parity audit: `deleteTable`'s dependent guard was added in a prior L1-parity
commit that explicitly enumerated every op it touched — `removeJoin` was never mentioned.
The repo's own `docs/parity/L1-worksheet-structure/L1-parity-report.md` already named this
gap ("row 17's guard was exercised against mirror and concatenation dependents but not a
join").

## Test plan

- [x] New test `removeJoinRefusesWhenSomethingIsBuiltOnIt` (join + mirror sourced from it) —
      asserts `PairingException` naming "built on it", both the join and the mirror survive.
- [x] Existing `removeJoinRemovesAssembly` (no-dependent case) still passes unchanged.
- [x] `WorksheetEditServiceMutatorsTest,WorksheetEditServiceTest` (266 tests) — no
      regressions in the shared file.

```
./mvnw.cmd -pl core -o -Dtest=WorksheetEditServiceMutatorsTest#removeJoinRefusesWhenSomethingIsBuiltOnIt+removeJoinRemovesAssembly test -DfailIfNoTests=false
  → Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS

./mvnw.cmd -pl core -o -Dtest=WorksheetEditServiceMutatorsTest,WorksheetEditServiceTest test -DfailIfNoTests=false
  → Tests run: 266, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS
```

Note: this branch and PR are independent of `fix/wbs-036` (also touching
`WorksheetEditService.java`, in an unrelated method `editUnpivot`, PR #5070) — both are based
directly on `origin/stylebi-wiz-main-rebase`.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -908,11 +908,26 @@ public class WorksheetEditService {
       /**
        * Removes an assembly (typically a join assembly) from the worksheet by name.
        *
-       * <p>No-ops if no assembly with {@code name} exists.</p>
-       *
        * @param name the assembly name to remove
+       * @throws PairingException if no assembly with {@code name} exists, or if another
+       *                          assembly is built on it
        */
-      public void removeJoin(String name) {
+      public void removeJoin(String name) throws PairingException {
+         Assembly a = ws.getAssembly(name);
+
+         if(a == null) {
+            throw new PairingException("Join assembly not found in worksheet: " + name);
+         }
+
+         if(AssetEventUtil.hasDependent(a, ws, Set.of(name))) {
+            throw new PairingException(
+               "\"" + name + "\" cannot be removed because other assemblies are built on it. " +
+               "Removing it would leave them referencing an assembly that no longer exists, every " +
+               "query against them failing, and nothing able to repair them. Remove the " +
+               "assemblies that depend on it first -- read_worksheet_model reports each table's " +
+               "sources field, which is what references what.");
+         }
+
          ws.removeAssembly(name);
       }
 
--- a/core/src/test/java/inetsoft/web/wiz/worksheet/WorksheetEditServiceMutatorsTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/worksheet/WorksheetEditServiceMutatorsTest.java
@@ -2610,6 +2610,29 @@ class WorksheetEditServiceMutatorsTest {
       assertNull(ws.getAssembly("J"), "join assembly 'J' should have been removed");
    }
 
+   /** A join something depends on refuses to be removed -- mirrors deleteTable's own guard. */
+   @Test
+   void removeJoinRefusesWhenSomethingIsBuiltOnIt() throws Exception {
+      Worksheet ws = new Worksheet();
+      EmbeddedTableAssembly left  = TestWorksheets.tableWithColumns(ws, "L", "id");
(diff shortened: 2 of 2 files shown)


lizhepu:
## Round 1 review — clean, ready to merge

Checked:

1. **Guard shape matches `deleteTable`'s** — read both side by side. `removeJoin` now does the identical existence-check-then-`AssetEventUtil.hasDependent(a, ws, Set.of(name))` sequence, throws `PairingException` in both branches, and the refusal message follows the same template (names the assembly, explains the dangling-reference consequence, points at `read_worksheet_model`'s `sources` field). Order is existence-check first, `hasDependent` second — same as `deleteTable`, so `hasDependent` is never called with a name that doesn't resolve to an assembly.

2. **`hasDependent` signature/safety** — confirmed at `AssetEventUtil.hasDependent(Assembly, Worksheet, Set)`: it calls `assembly.getAssemblyEntry()` on the passed-in `Assembly` immediately, so it would NPE on a null assembly — but the existence check in `removeJoin` runs first and returns before reaching that call, exactly matching `deleteTable`'s own order. No edge case here.

3. **Regression test** — `removeJoinRefusesWhenSomethingIsBuiltOnIt` builds a join `J` (`L`⋈`R`) plus a mirror `M` sourced from `J`, asserts `removeJoin("J")` throws `PairingException` with message containing "built on it", and then asserts both `J` and `M` still exist afterward — not just "some exception," a real assertion of survival + specific message. The existing no-dependent-case test `removeJoinRemovesAssembly` is untouched (same as before, still asserts `J` is null after removal).

4. **Ran the tests myself** from the worktree (not `E:\inetsoft\wizard`):
   - `WorksheetEditServiceMutatorsTest` alone: 240/240 pass.
   - `WorksheetEditServiceMutatorsTest,WorksheetEditServiceTest` together: 266/266 pass, BUILD SUCCESS — matches the PR description's claimed numbers exactly.

5. **Scope** — diff touches only `removeJoin` (main) and the new test (test file). No touches to `editUnpivot`/`AssetUtil.checkUnpivotShrinkTypeConflict` (WBS-036) or `addConcatSubtable` (WBS-033). Clean, minimal, no unrelated refactors.

No blocking issues found. Two non-code items for the lead's awareness at merge time (not from this diff, and not blocking my review):
- `claude-review` check shows `fail` but the log is just the bot erroring out with no findings ("Claude encountered an error after 0s") — not a real review comment.
- `build` check is still `pending` as of this comment.

Verdict: **ready to merge** once `build` goes green. Not merging myself per this run's process (lead handles merges in the `stylebi` repo).

PR #5073, opened 2026-09-08
fix(wiz): add per-column type check to addConcatSubtable (bug 76517, WBS-033)

lizhepu:
## Summary
- `WorksheetEditService.addConcatSubtable` validated only column COUNT, unlike its sibling
  `addConcatenation`, which also checks that column i of each source is
  `AssetUtil.isMergeable` with column i of the first source. A subtable whose column types did
  not line up by position with the existing concatenation's subtables could be spliced in with
  no error at edit time (the mismatch would only later surface as `concatCompatible: false` on
  a subsequent `read_worksheet_model`, with nothing telling the caller to check that field).
- Added the missing per-column type-compatibility check to `addConcatSubtable`, throwing
  `PairingException` naming the 1-based position and both column names/types, mirroring
  `addConcatenation`'s existing check.
- Extracted the two methods' near-identical loop bodies into a shared private helper
  (`checkColumnTypesLineUp`) rather than duplicating the loop, since both a diagnosis and an
  independent adversarial refutation on bug 76517 flagged this as the cleaner option to avoid
  the two checks drifting apart again (which is exactly how this gap arose in the first place).

## Test plan
- [x] `./mvnw.cmd -pl core -o -Dtest=WorksheetEditServiceMutatorsTest#addConcatSubtableRejectsColumnsThatDoNotLineUpByType+addConcatSubtableAcceptsColumnTypesThatMerge+addConcatSubtableRefusesASourceAlreadyPresent+addConcatSubtableRefusesACycleWithoutMutating+addConcatSubtableRefusesSelfReferenceWithoutMutating test -DfailIfNoTests=false` → 5/5 pass
- [x] `./mvnw.cmd -pl core -o -Dtest=WorksheetEditServiceMutatorsTest,WorksheetEditServiceTest test -DfailIfNoTests=false` → 267/267 pass, no regressions

Note: based directly on `origin/stylebi-wiz-main-rebase`, not on the unmerged `fix/wbs-036`
(#5070) or `fix/wbs-037` (#5072) branches, which touch the same file in unrelated methods
(`editUnpivot`, `removeJoin`).

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetEditService.java
@@ -1549,31 +1549,8 @@ public class WorksheetEditService {
                   (otherCount == 1 ? "" : "s") + ".");
             }
 
-            for(int c = 0; c < colCount; c++) {
-               // Read the type off DataRef rather than narrowing to ColumnRef: skipping a position
-               // whose ref is some other kind would silently leave it unchecked, which is the very
-               // failure this validation exists to prevent.
-               DataRef first = firstColumns.getAttribute(c);
-               DataRef other = otherColumns.getAttribute(c);
-               String ftype = first.getDataType();
-               String otype = other.getDataType();
-
-               // isMergeable dereferences both arguments; a ref with no type at all is not
-               // something this check can speak to, so leave it to the server.
-               if(ftype == null || otype == null) {
-                  continue;
-               }
-
-               if(!AssetUtil.isMergeable(ftype, otype)) {
-                  throw new PairingException(
-                     "Columns are concatenated by position, and position " + (c + 1) +
-                     " does not line up: \"" + sources[0].getName() + "\" has \"" +
-                     first.getAttribute() + "\" (" + ftype + ") while \"" + sources[i].getName() +
-                     "\" has \"" + other.getAttribute() + "\" (" + otype + "), and those types " +
-                     "cannot be merged. Reorder the columns so matching ones share a position, " +
-                     "or change one column's type with change_column_type.");
-               }
-            }
+            checkColumnTypesLineUp(firstColumns, sources[0].getName(),
+                                    otherColumns, sources[i].getName(), colCount);
          }
 
          // Build one operator per adjacent pair.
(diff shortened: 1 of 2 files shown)


lizhepu:
## Review (round 1)

Checked the diff against `addConcatenation`'s existing check and re-ran the tests myself from the `wt-wbs033-java` worktree (not trusting the pasted numbers).

**Logic**
- `checkColumnTypesLineUp` is a faithful extraction of the pre-existing `addConcatenation` loop body (same `DataRef`-level type read, same `null`-type `continue`, same `AssetUtil.isMergeable` check, same message shape). `addConcatSubtable` now calls it right after the existing count check, anchored on `existing[0]`'s `ColumnSelection` — matches the diagnosis's proposed shape, and the "anchor on `existing[0]` vs. the immediately-preceding subtable" equivalence claim is sound given `isMergeable`'s documented disjoint-class partition (string/number/date/identity), same reasoning already used for `addConcatenation`'s own anchor-on-`sources[0]` choice.
- Bounds: `colCount`/`newCount` equality is enforced by the existing count check *before* `checkColumnTypesLineUp` runs, so the shared loop never indexes past either `ColumnSelection`'s attribute count. No new out-of-bounds risk.
- Ordering preserved: cycle check still runs before the column-count/type checks (per the existing comment on why that order matters for `newTable.getColumnSelection`).
- Extraction to a shared `checkColumnTypesLineUp` helper (rather than duplicating the loop) is the right call given the bug's own root cause was the two checks drifting apart; `addConcatenation`'s call site is a behavior-preserving refactor, not a functional change.

**Tests**
- `addConcatSubtableRejectsColumnsThatDoNotLineUpByType` and `addConcatSubtableAcceptsColumnTypesThatMerge` both read correctly: the first asserts message contents (position 2, both column names) and that the subtable count is unchanged on rejection; the second uses DOUBLE/CHAR against INTEGER/STRING to prove the fix doesn't over-tighten past what `isMergeable` already allows (the shape that keeps date/timeInstant repros passing).
- Existing `addConcatSubtableRefuses*` (self-reference/cycle/duplicate-source) tests are untouched in the diff.
- Ran `./mvnw.cmd -pl core -o -Dtest=WorksheetEditServiceMutatorsTest test -DfailIfNoTests=false` myself from the worktree root: **241 tests, 0 failures, 0 errors** (BUILD SUCCESS). The NPE/"Prefix is null" warnings in the log are the known pre-existing `refreshAssemblies` warm-up noise, unrelated to this change.

**Scope**
- Confirmed via `git show 42b37be28 --stat`: only `WorksheetEditService.java` and `WorksheetEditServiceMutatorsTest.java` touched. No changes to `editUnpivot`, `removeJoin`, or `AssetUtil.checkUnpivotShrinkTypeConflict`.

**CI**
- `build` and `Semgrep`/`Semgrep OSS`/CLA all pass. `claude-review` shows a failure, but it's an infra-side error ("Claude encountered an error after 0s"), not a content finding — consistent with the same bot's behavior on other PRs this run.

No blocking issues found. Ready to merge from a content/test standpoint.

PR #5076, opened 2026-09-08
fix(bug-76400): preserve aggregate alias through auto per-viewsheet mirror resolution

larryliang-inetsoft:
## Summary

Closes a confirmed, JUnit-reproduced defect: a `set_group_aggregate`'d worksheet table silently
loses its own aggregate when queried by a viewsheet chart, with no error surfaced.

- `AssetUtil.getColumnRefFromAttribute` unconditionally unwraps a `ColumnRef`-typed input to its
  raw `DataRef` before any matching, discarding whatever alias `set_group_aggregate` set.
- `Worksheet.getVSTableAssembly`'s auto `"V_"`-prefixed per-viewsheet mirror — which every VS
  chart binds through, and which is rebuilt fresh from the base table's current public column
  selection on every single query (`VSAQuery.getVSTableAssembly`) — bakes that alias into its own
  column's bare *attribute* name (`AssetUtil.getOuterAttribute`), never carrying an alias of its
  own.
- Once the base table's public selection becomes alias-aware (via `ViewsheetSandbox`
  construction — core, unavoidable infrastructure every viewsheet render goes through), the
  chart's own design-time-resolved aggregate ref can never resolve against the mirror: it searches
  by the raw (pre-alias) attribute name, which the mirror no longer exposes. The aggregate is
  silently dropped from `AggregateInfo` — not doubled, not miscomputed, just gone — and no error
  surfaces anywhere.

**Relationship to Redmine #76400 (PWA-001) — disclosed, not settled**: this defect was found while
investigating #76400, whose originally-reported symptom is a single *wrong value* on a chart
(≈3.8M, matching neither region's total, their sum, nor their average). This fix's mechanism
produces a *missing* aggregate column instead — a different observable shape. Two independent
prior live verifications (`docs/teams/2026-09-01-bug-76400/`, 2026-09-01→09-03) found the
originally-reported symptom does **not** reproduce live on this same repro shape. Whether this
fix's mechanism explains the original report is therefore open and unreconciled — full tension
recorded in `04-fix.md`'s "Step 1" section. This PR closes the newly-found defect on its own
merits (real, demonstrated, reachable via core production infrastructure — see Findings 1-4), not
as a claimed fix for #76400's original report.

Full investigation trail (three-round `/deep-debug-workflow`, competing hypotheses, empirical
JUnit reproduction, reachability check against core `ViewsheetSandbox` infrastructure):
`docs/teams/2026-09-07-bug-76400-group-aggregate-chart-poison/` in the `stylebi-wiz` repo
(`00-charter.md` through `06-review-r1.md`).

## Fix

Minimal, conservative: `AssetUtil.getColumnRefFromAttribute` now keeps the original (pre-unwrap)
`ColumnRef`'s alias around, and — only as a last resort, after every existing raw-attribute and
alias-based match has already failed — tries that alias against candidate columns' own bare
attribute name. A candidate is only eligible when it carries no alias of its own, so this can
never mask a genuinely different, differently-aliased column that happens to share the bare name.

Sibling check: `AssetUtil.getColumnRefFromAttribute` has 24 call sites across
`ChartVSAQuery`/`AbstractCrosstabVSAQuery`/`MirrorQuery`/MV column classes/`VSAggregateRef`/
`VSDimensionRef`/composer web services. The fix applies uniformly through the shared utility
(also benefits `VSDimensionRef.createGroupRef`'s group-side resolution, not just this bug's
aggregate-side repro — confirmed real by review, not exercised by any test in this PR).

## Test plan

- [x] New regression test `ChartGroupAggregatePoisonEndToEndTest` (both the no-mirror and
      explicit-mirror sub-repros from the original bug report) — RED before this change
      (aggregate lost entirely, 0 aggregates), GREEN after (resolves exactly once, correct
      per-region totals, `VSDataSet.indexOfHeader` finds the chart's bound measure).
- [x] Broader regression sweep, 0 failures: `AggregateInfoTest`, `ViewsheetCreateMirrorTableTest`,
      `ChartVSAQueryTest`, `CrosstabNamedGroupEndToEndTest`,
      `CrosstabSortByValueAliasedAggregateNameTest`, `CrosstabSortByValueEntityQualifiedNameTest`,
      `CrosstabVSAQueryTest`, `CrosstabTreeTableLensTest`,
      `CrosstabHighlightSummaryHeaderColumnShiftTest`, `MVDefWorksheetSerializationTest`,
      `MVQueryTest`, `MVOverflowedGroupsTest` (36 tests total, 0 failures) — independently
      re-verified by review (3 of 12 classes re-run against the exact PR head commit).
- [x] Reviewed (`06-review-r1.md`): no blocker/important finding against the code change itself.
      CI `build` is currently red due to a pre-existing compile break already on the base branch
      (`WorksheetEventServiceTest.java`, unrelated to this PR's diff — confirmed via
      `git merge-base --is-ancestor`) — not introduced by this change.

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java
+++ b/core/src/main/java/inetsoft/uql/asset/internal/AssetUtil.java
@@ -552,6 +552,14 @@ public class AssetUtil {
          return null;
       }
 
+      // Bug #76400: a ColumnRef's own alias (e.g. an aggregate output alias set by
+      // set_group_aggregate) is about to be discarded by the unwrap below. A downstream table
+      // (e.g. the auto per-viewsheet mirror Worksheet.getVSTableAssembly lazily creates) may have
+      // folded that same alias into one of its own columns' bare attribute name (see
+      // MirrorTableAssembly.updateColumnSelection -> AssetUtil.getOuterAttribute), so keep it
+      // around as a last-resort match key in case the raw-attribute match below finds nothing.
+      String origAlias = ref instanceof ColumnRef ? ((ColumnRef) ref).getAlias() : null;
+
       if(ref instanceof ColumnRef) {
          ref = ((ColumnRef) ref).getDataRef();
       }
@@ -603,7 +611,32 @@ public class AssetUtil {
          }
       }
 
-      return column2;
+      if(column2 != null) {
+         return column2;
+      }
+
+      // Bug #76400 fallback: nothing matched the raw attribute name. If the original (pre-unwrap)
+      // ref carried an alias, try it against candidate columns' own bare ATTRIBUTE (not their
+      // alias) -- covers a downstream table whose column selection was rebuilt from a column that
+      // carried this same alias, baking the alias in as that column's own attribute name (e.g.
+      // AssetUtil.getOuterAttribute, used by MirrorTableAssembly.updateColumnSelection). Only a
+      // column with no alias of its own is eligible, so this never masks a genuine, differently-
+      // aliased column that happens to share this bare name.
+      if(origAlias != null && origAlias.length() > 0) {
+         iter = columns.getAttributes();
+
+         while(iter.hasMoreElements()) {
+            ColumnRef dref = (ColumnRef) iter.nextElement();
+
+            if(Tool.equals(dref.getAttribute(), origAlias) &&
+               (dref.getAlias() == null || dref.getAlias().isEmpty()))
+            {
+               return dref;
+            }
+         }
+      }
+
+      return null;
    }
 
    /**
--- a/core/src/test/java/inetsoft/report/composition/execution/ChartGroupAggregatePoisonEndToEndTest.java
+++ b/core/src/test/java/inetsoft/report/composition/execution/ChartGroupAggregatePoisonEndToEndTest.java
@@ -1,0 +1,543 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.report.composition.execution;
+
+import inetsoft.report.TableLens;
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft (commented):
## Review (reviewer-76400, deep-debug-workflow)

Verified the fix directly against the actual committed diff at HEAD `7ae8965a2` (matches `headRefOid` exactly), not just the archived docs' prose.

**Fix logic: sound, no blockers.**
- Not keyed on repro-specific names (`origAlias`/`dref` generic, no `TOTAL_SALES`/`SALES_AMOUNT` in product code) — general fix.
- "Last resort, alias-less-candidate-only" property verified directly in `AssetUtil.java:548-639`: the new fallback block sits strictly after both pre-existing matching loops return null, and only matches a candidate with no alias of its own.
- No-alias regression risk traced and ruled out: `origAlias` is `null` whenever the input `ColumnRef` has no alias, so the new branch never executes for that case.
- Sibling claim confirmed by direct read: `VSDimensionRef.createGroupRef` (`VSDimensionRef.java:253-254`) does call the same `getColumnRefFromAttribute` overload, so the fix's benefit genuinely extends to group-side resolution — worth noting this path isn't exercised by any test in this run, though the fix is provably correct there by the same call-site-agnostic reasoning.
- Spot-checked the regression sweep myself (not trusted from the doc): ran `ChartGroupAggregatePoisonEndToEndTest`, `AggregateInfoTest`, `ChartVSAQueryTest` from a fresh Maven invocation against this exact PR head commit — 10/10 pass, matching the claimed fixed behavior (aggregates=1, correct per-region totals, `indexOfHeader` resolves).

**Two non-code items to resolve before/around merge:**

1. **CI `build` check is currently red** (`gh pr checks 5076`). Pulled the actual failing job log — it's a clean-room `testCompile` failure in `WorksheetEventServiceTest.java` (a `WorksheetEventServiceProxy.openWorksheet(...)` arg-count mismatch), completely unrelated to this PR's diff (which only touches `AssetUtil.java` + the new test file). Traced provenance: the failing test's only commit (`7ee0203c6`) is already an ancestor of the base branch tip (`stylebi-wiz-main-rebase` @ `27800d56c`) — this is a pre-existing compile break on the base itself, not something this PR introduced. Doesn't need fixing in this PR, but blocks a clean merge signal as-is.
2. **PR body slightly overclaims ticket-closing scope.** It's accurate about the CONFIRMED mechanism (aggregate lost entirely via the auto `"V_"` per-viewsheet mirror + alias-stripping), but opens with "Fixes Redmine #76400 (PWA-001)" without noting that this mechanism doesn't match the originally-reported symptom (a single wrong ~3.8M value vs. this fix's "aggregate column entirely absent"), and that two independent prior live verifications found the same repro shape rendering correctly live. Per the archived investigation's own `STATUS.md`, the human already decided to ship this fix on its own merits regardless of PWA-001 relevance — recommend the PR body/Redmine close-out note say so explicitly so a future reader doesn't assume the original customer symptom is now explained.

Full writeup: `docs/teams/2026-09-07-bug-76400-group-aggregate-chart-poison/06-review-r1.md`

Posting as a comment rather than approve/request-changes formally, since the PR author and this review's `gh` account are the same login (self-review).

PR #5077, opened 2026-09-08
Bug #76511, fix crosstab sort on date levels created by Expand Hierarchy

ngovus-inetsoft:
## Problem

In a crosstab bound to a date column, after **Expand Hierarchy** the sort icon on a
`MonthOfYear` or `DayOfMonth` row-header cell does nothing — the row order never changes.
`Year` and `QuarterOfYear` sort normally. Saving and reopening the viewsheet makes the
sort work again.

## Root cause

`VSUtil.getNextLevelRef()` builds each expanded level by cloning the dimension of the
previous level and lowering its date level. The clone also inherits the previous level in
the transient `VSDimensionRef.oldRuntimeDateLevel` field, so the first
`VSDimensionRef.update()` after the drill sees the level as changed and latches
`runtimeDateLevelChange() == true` (the flag is only cleared by `setDateLevel()`).

From then on, on every execution:

1. `CrossBaseVSAssemblyInfo.removeUselessChildRefs()` sees the flag and strips those levels
   out of `CrosstabTree.childRefs`. `removeParentRef()` removes the key equal to the ref's
   *own* full name, so `Year(...)` survives but `QuarterOfYear(...)`, `MonthOfYear(...)`
   and the root key are dropped.
2. `CrosstabTree.updateHierarchy()` re-registers the links through `updateChildRef()`,
   which now takes the `collapsedRefs.containsKey(refName) && !childRefs.containsKey(parent)`
   branch and calls `childRef.copyOptions(...)`. `collapsedRefs` is keyed by the *base*
   column name, so all four levels share one slot and each level gets the previous level's
   options — including `order`.

Walking the row headers, `Year` keeps its key (no copy), `QuarterOfYear` copies from
itself (harmless), and `MonthOfYear`/`DayOfMonth` get their sort order overwritten — which
is exactly the reported set of broken levels. The order that
`BaseTableSortColumnService.setOrder()` writes to the design ref is applied correctly and
then clobbered before the query runs.

Both `oldRuntimeDateLevel` and `runtimeDValueChange` are transient and not persisted, so
after a save/reopen the flag is never set and the sort works.

## Fix

Clear the tracked runtime date level on the dimension a crosstab drill creates, so a drill
is not mistaken for a dynamic date level change:

- `VSDimensionRef.resetOldRuntimeDateLevel()` — new.
- Called in `CrosstabDrillHandler.drillDownChild()` (Expand Hierarchy / Expand Field),
  `CrosstabDrillHandler.drillUpAction()` (collapse), and
  `BaseTableDrillService.getDrillDownRef()` (per-cell `[+]` drill).

Dynamic date levels are deliberately left alone: the variable case already clears the flag
via `setDateLevel()`, and `DrillFilterVSAssembly.validDrillFilterCondition()` depends on
the flag for dynamic levels. The guards therefore use `!VSUtil.isDynamicValue(...)` so
script-valued levels are excluded too. The chart drill paths are untouched.

## Testing

New `CrosstabDrillDateLevelTest` (2 cases): a drilled level must not report
`runtimeDateLevelChange()`, and with the level unflagged the hierarchy survives in the
`CrosstabTree` so re-registering it does not copy the parent level's sort order over the
child.

With the fix reverted both fail, reproducing the bug:

```
expandHierarchyDoesNotFlagRuntimeDateLevelChange:84
  QuarterOfYear(Date) must not report a runtime date level change ==> expected: <false> but was: <true>
drilledLevelKeepsItsOwnSortOrder:129
  MonthOfYear(Date) sort order must not be overwritten by QuarterOfYear(Date) ==> expected: <1> but was: <2>
```

With the fix: `tests="2" errors="0" failures="0"`.

Also verified manually in the app against Data Source/Examples/Orders/Order Model:
Expand Hierarchy, then sort on MonthOfYear and DayOfMonth.

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/VSDimensionRef.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/VSDimensionRef.java
@@ -1831,6 +1831,18 @@ public class VSDimensionRef extends AbstractDataRef implements ContentObject, XD
    }
 
    /**
+    * Clear the tracked runtime date level. A dimension created for a new date level by a
+    * drill operation is cloned from the dimension of the previous level, so it inherits
+    * the previous level in oldRuntimeDateLevel. If it is not cleared, the next update()
+    * sees the level as changed and flags runtimeDateLevelChange(), which makes the drill
+    * look like a dynamic date level change to the callers of that method.
+    */
+   public void resetOldRuntimeDateLevel() {
+      oldRuntimeDateLevel = null;
+      runtimeDValueChange = false;
+   }
+
+   /**
     * Set runtime id.
     */
    public void setRuntimeID(int rid) {
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/table/BaseTableDrillService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/table/BaseTableDrillService.java
@@ -857,7 +857,14 @@ public abstract class BaseTableDrillService <T extends BaseTableEvent> extends B
     * Get next ref fro drill down.
     */
    private VSDimensionRef getDrillDownRef(VSDimensionRef ref, XCube cube) {
-      return VSUtil.getNextLevelRef(ref, cube, true);
+      VSDimensionRef nref = VSUtil.getNextLevelRef(ref, cube, true);
+
+      // see the comment in CrosstabDrillHandler.drillDownChild()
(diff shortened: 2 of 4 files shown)


jshobe-inetsoft (approved):
Claude's automated review traced the full root-cause chain in source and confirms the fix is correct, well-scoped, and well-tested with no blocking issues. Approving.

jshobe-inetsoft (approved):

PR #5079, opened 2026-09-08
Bug #76514, follow calendar order for moving-average windows on value-sorted part-date dimensions

ngovus-inetsoft:
Fixes Bug #76514.

## Problem

A chart binding `HourOfDay(order_time)` (Top-N "Sort By Value" ranking on `Sum(contact_id)`), `Sum(employee_id)`, and a centered `Moving Average of 3: Sum(employee_id)` put the averages — and the truncated-window nulls — on the wrong hours. The underlying `Sum(employee_id)` values are correct; only the derived calc column is wrong.

Caught by the datatest chart regression suite, `Binding_Spec` / `binding7`, diffing against the accepted baseline. Reproduces in `filter_Chart3.txt` too.

| HourOfDay | Sum(employee_id) | MA(3) before | MA(3) expected |
|---|---|---|---|
| 5 | 275 | null | 253.6667 |
| 11 | 320 | 256 | null |
| 1 | 173 | 219.6667 | null |
| 2 | 166 | null | 204.6667 |

`5, 11, 1, 2` is the configured *display* order (`Sum(contact_id)` = 2043, 2015, 1229, 1215 desc), not insertion order. Calendar order is `1(173), 2(166), 5(275), 11(320)`, so the expected values are `avg(173,166,275) = 204.6667` on hour 2 and `avg(166,275,320) = 253.6667` on hour 5, with the calendar-first and calendar-last hours truncated to null. There is no off-by-one anywhere — the whole window was landing on the wrong rows.

## Cause

Two independent causes had to line up.

**1. `DataSetRouter` navigated in rank-value order.** `HourOfDay` is a part-date level, so there is no `TimeScale`, `VSDataSet.getRouter()` returns null, and navigation falls back to `DataSetRouter`. The second commit of #4658 (`95ee0cb3`, Bug #76039) made any configured comparator win unconditionally, which removed the `isSortByValue` exception the *first* commit of that same PR (`01f1dc78`, Bug #76059) had deliberately kept — reverting the follow-up fix for Bug #75664 (`e804d7b8c`). That also contradicts #4658's own description ("Part-date-group fields keep a natural calendar order as the fallback…").

**2. `MovingColumn` walked the window by row index.** `getCondData()` selects the window's *members* by the dimension's navigation order (`router.getValues()[vindex-preCnt .. vindex+nextCnt]`), but the loop then walked the resulting `SubDataSet` by row index. `DataSetIndex.createSubDataSet` builds the sub from a `BitSet`, so its rows keep physical row order. The two only agree while the router order happens to match the physical order — with the router fixed alone, hour 5 averaged **297.5**, neither the baseline nor the reported output.

## Fix

**`DataSetRouter`** — restore the `isSortByValue` exception. A part-date-group dimension falls back to natural calendar order both when it has no display sort (Bug #75664) and when its sort is value-based (Bug #75664-1); "previous hour" only has meaning in calendar order. An explicit label sort — ascending, descending, specific order — is still honored, as Bug #76059 requires, and non-part-date dimensions keep `data.getComparator(field)` verbatim, so the rest of #76039's intent is preserved.

**`MovingColumn`** — walk the window in the router's order via a new `orderRowsByDim(DataSet, Router)` helper, so the traversal matches the order the members were selected in. Scoped narrowly:

- only applied when `getCondData` actually produced a sub data set, so there is no `O(n log n)`-per-row cost over a full dataset
- the no-`innerDim` path and the crosstab `calculate(CrosstabDataContext, PairN)` path are untouched
- when the two orders agree the traversal is byte-for-byte the previous behavior, so the existing `MovingColumnTest` cases are unaffected

`#76039` and `#75664-1` cannot both hold for a sort-by-value part-date dimension — calendar order and display order differ by definition. This PR takes the calendar-order side and says so in the test docstring, so it does not get flipped back a third time.

## Tests

- `ValueOfColumnTest.testPreviousOnPartDateGroupFollowsRankingSortOrder` reverted to `testPreviousOnPartDateGroupIgnoresRankingSortComparator`, asserting natural-hour-order previous again. Its `getOrder()` stub is load-bearing once more (the prior review flagged it as dead).
- New `MovingColumnTest.testMovingAverageFollowsCalendarOrderOnValueSortedPartDateDim` encodes binding7's expected values. Confirmed to fail with **either** half of the fix reverted — `expected: <253.6667> but was: <297.5>` without the `MovingColumn` half, and an error without the router half.
- `testPreviousOnPartDateGroupFollowsLabelSortWithNullGroup` (#76059) and `testPreviousOnPartDateGroupWithOthersLabel` (#75743 / no-sort "Others" fallback) untouched and still passing.
- 127 tests green across `inetsoft.graph.**` and `inetsoft.report.composition.graph.**`.

## Not yet verified

The datatest suite itself has not been re-run against this branch (no local datatest checkout) — `Binding_Spec` / `binding7` `Chart3.txt` and `filter_Chart3.txt` should now match with **no** baseline edits. The two originally-reported scenarios share this single code path and are worth a manual pass since they are what #4658 was for:

- #76059: auto chart, `SecondOfMinute(order_datetime)` on X, Sort (Ascending), "As time series" off, with a null group — `Value of previous` for second 1 must still resolve to the null group, and `Moving Average of 5` windows must not slide.
- #75743: part-date dim with no sort and a Top-N "group others" — must still use calendar order and not throw `ClassCastException` building the router.

Community-only change; no enterprise files touched.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/calc/MovingColumn.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/calc/MovingColumn.java
@@ -146,6 +146,8 @@ public class MovingColumn extends AbstractColumn {
       }
 
       Formula formula = (Formula) this.formula.clone();
+      // rows of 'data' in the order the window is walked; null means physical row order
+      int[] window = null;
 
       // dimension?
       if(innerDim != null) {
@@ -168,6 +170,14 @@ public class MovingColumn extends AbstractColumn {
                   break;
                }
             }
+
+            // getCondData picks the window's members by the dimension's navigation order, so
+            // the window must be walked in that order too. The sub data set holds only those
+            // members but keeps them in physical row order, which differs whenever the
+            // dimension's display sort differs from its navigation order (e.g. a Top-N "Sort
+            // By Value" ranking on HourOfDay); walking it by row index then averages the
+            // wrong neighbors and puts the truncated-window nulls on the wrong rows.
+            window = orderRowsByDim(data2, router);
          }
 
          data = data2;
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):
Claude's automated review traced both fix halves against the actual call paths and found them consistent with the described root cause; only one non-blocking suggestion (add a debug log for an unreachable fallback branch). Approving.

jshobe-inetsoft (approved):

PR #5080, opened 2026-09-08
Bug #76520, fix RangeSlider Edit dialog for time-of-day bindings

ngovus-inetsoft:
Redmine Bug #76520

## Problem

Right-clicking a RangeSlider bound to a column of data type `time` (time of day only, e.g. `01:30`, `06:30`) and choosing **Edit** opened a broken dialog:

- From/To rendered as `datetime-local` inputs (`mm/dd/yyyy --:-- --`) instead of time-only inputs.
- Both fields were empty even though the slider had a valid selected range.
- A spurious "From and To values cannot be equal" error was shown while both fields were blank.

## Root cause

A `TIME` column gets `rangeType = MINUTE_OF_DAY` (`VSBindingService`; `HOUR_OF_DAY`/`MINUTE_OF_DAY` are the only range types offered for `TIME`). That range type serializes tick values with `Tool.timeFmt` = `{'t' ''HH:mm:ss''}`, so `VSRangeSliderModel.values[]` holds strings like `{t '01:30:00'}` while the labels are `HH:mm`.

`VSRangeSlider.extractTimeIncrement()` returned `typeMatch[1].charAt(0)`. The five value tokens the server writes are `y`, `m`, `d`, `t` and `ts`, so taking the first character collapsed **`t` (time of day) and `ts` (timestamp) into the same `"t"` increment** — a single defect that produced all three symptoms:

1. `"t"` means "date + time" downstream, so `normalizeStr` built the dialog bounds as `new Date("01:30:00")` → **Invalid Date**.
2. `rangeMin instanceof Date` is still `true` for an Invalid Date, so `isDateType = true` with `timeIncrement === "t"` selected the `datetime-local` input.
3. `formatDate()` on an Invalid Date returned `"NaN-NaN-NaNTNaN:NaN"`, which the browser rejects → **blank fields**; and both controls held that identical string, so `minMaxNotEqual()` fired → **the equality error**.

The commit path was broken the same way: `labelToTimestamp("t", "{t '01:30:00'}")` also evaluated `new Date("01:30:00")` → `NaN`, so no tick could ever match even if valid input were entered.

## Fix

Give time of day its own increment, keyed off the **value token** rather than the column data type — the value format is the actual source of truth and stays correct however range type and column type combine.

**`vs-range-slider.component.ts`**
- `extractTimeIncrement()` maps the brace token exactly: `t` → the new time-of-day increment, `ts` → `"t"`, `y`/`m`/`d` unchanged. The unescaped fallback now uses the dash count to tell time-only from date-time.
- `normalizeStr`, `getDateStrings()`, `toIncrementTimestamp()` and `labelToTimestamp()` handle the new increment, anchoring the time to a fixed base date so all client-side comparisons stay consistent.

**`range-slider-edit-dialog.component.ts` / `.html`**
- Exports the shared `TIME_OF_DAY_INCREMENT` / `TIME_BASE_DATE` constants so the host and the dialog cannot drift apart.
- Renders `<input type="time">` for the new increment; `formatDate()` emits `HH:mm`, matching the labels the slider itself shows.
- The duplicated parse ternaries in the two range validators and both `valueChanges` subscriptions are consolidated into one private `toDate()` helper.

`isDateType`, `minBeforeMax()` and `minMaxNotEqual()` needed no change — the bounds are now valid `Date`s, and `"01:30"` vs `"06:30"` compares correctly as strings.

Frontend only; no backend change.

### Side effects considered

- `extractTimeIncrement()` has exactly one call site, so the token-mapping change is contained.
- `y`/`m`/`d`/`ts` keep their existing increment codes, so date and timestamp sliders behave exactly as before.
- `<input type="time">` yields `HH:mm`, dropping the `:ss` from the server value. `MINUTE_OF_DAY`/`HOUR_OF_DAY` ticks always land on whole minutes/hours and the commit handler snaps to the nearest tick index, so no precision is lost.
- Composite (multi-column) range sliders leave `dataType` null and already fall into the numeric branch — a separate pre-existing gap, out of scope here.

## Testing

- New `vs-range-slider.component.timeofday.tl.spec.ts` (10 tests): token mapping (`{t}` vs `{ts}` vs `y`/`m`/`d`, plus the unescaped fallback), timestamp anchoring, dialog seeding with valid `Date`s, and the commit snap-to-tick round-trip.
- New Group 9 in `range-slider-edit-dialog.component.tl.spec.ts` (7 tests): a single `input[type=time]` per field, `HH:mm` seeding, absence of the spurious `minMaxEqual` error, time-only parsing, and the boundary validators.
- All new tests fail against the pre-fix code.
- All 5 range-slider TL spec files pass (152 tests). All `vsobjects` standard specs pass (304 tests). ESLint clean on the changed files.
- Verified manually against a running server on a `time`-typed binding: From/To are time pickers pre-filled with the current selection, no validation error, and editing the range updates the slider and dependent assemblies.

Note: the full `portal:test-tl` suite is unstable on `main` independently of this change (baseline with the change stashed: 28 failed files / 265 failed tests; with the change: 29 / 202 — counts vary run to run, and no failures are in range-slider specs).

### DIFF
--- a/web/projects/portal/src/app/vsobjects/dialog/range-slider-edit-dialog.component.html
+++ b/web/projects/portal/src/app/vsobjects/dialog/range-slider-edit-dialog.component.html
@@ -26,11 +26,14 @@
         @if (!isDateType) {
           <input type="number" step="any" class="form-control" formControlName="min">
         }
-        @if (isDateType && timeIncrement !== 't') {
+        @if (isDateType && timeIncrement !== 't' && timeIncrement !== 'time') {
           <input type="date" class="form-control" formControlName="min">
         }
         @if (isDateType && timeIncrement === 't') {
           <input type="datetime-local" class="form-control" formControlName="min">
+        }
+        @if (isDateType && timeIncrement === 'time') {
+          <input type="time" class="form-control" formControlName="min">
         }
         <label>_#(From)</label>
         @if (rangeForm.get('min')?.errors?.['required']) {
@@ -55,11 +58,14 @@
         @if (!isDateType) {
           <input type="number" step="any" class="form-control" formControlName="max">
         }
-        @if (isDateType && timeIncrement !== 't') {
+        @if (isDateType && timeIncrement !== 't' && timeIncrement !== 'time') {
           <input type="date" class="form-control" formControlName="max">
         }
         @if (isDateType && timeIncrement === 't') {
           <input type="datetime-local" class="form-control" formControlName="max">
+        }
+        @if (isDateType && timeIncrement === 'time') {
+          <input type="time" class="form-control" formControlName="max">
         }
         <label>_#(To)</label>
         @if (rangeForm.get('max')?.errors?.['required']) {
--- a/web/projects/portal/src/app/vsobjects/dialog/range-slider-edit-dialog.component.tl.spec.ts
+++ b/web/projects/portal/src/app/vsobjects/dialog/range-slider-edit-dialog.component.tl.spec.ts
@@ -28,6 +28,8 @@
  *   Group 6 [Risk 2] — minBeforeMax (group validator): numeric vs. date comparison mode
  *   Group 7 [Risk 2] — minMaxNotEqual (group validator): numeric vs. date comparison mode
  *   Group 8 [Risk 2] — close / ok: output emission contracts, numeric fallback-to-0, OK button disabled state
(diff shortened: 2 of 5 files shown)


jshobe-inetsoft (approved):
Claude's automated review found no correctness, security, or backward-compatibility issues; only minor non-blocking style notes. Approving.

jshobe-inetsoft (approved):

PR #5082, opened 2026-09-08
Bug #76515, order the "Others" bucket by its own group level

ngovus-inetsoft:
## Problem

A chart with a "Running Sum" calc broken by a dimension produced a non-monotonic sequence whenever a dimension bucket spanned more than one row — a cumulative sum of non-negative values that decreases.

Both the buggy and the correct output are internally consistent running sums totalling 40; only the accumulation order differed. Reconstructing the root row order from the values shows the correct order is chronological by the inner dimension, while the buggy one was arbitrary.

## Cause

**`SummaryFilter.MergedGroupNode`** gathered the children of the "Others" bucket into a `HashMap` and built the merged node by iterating its values, so the bucket's rows came out in hash order. `topNFilter` recurses bottom-up, so the bucket is assembled *after* its children have been sorted, and nothing put the gathered list back in order. Every other group at that level is ordered, and the order-dependent calculations (running total, moving, change/previous) read the resulting row order, so the bucket has to be ordered the same way.

**`RunningTotalColumn`** separately summed the whole group onto every row of it when the break-by column was itself the innermost dimension. The date walk added by #74910 compares `innerDim` dates, and with `innerDim == breakBy` every row of a group carries the same date, so `date <= currentDate` matched all of them.

## Fix

- Gather into a `LinkedHashMap` and sort the merged children by their own group level's `SortOrder`, reusing the existing `GroupNodeComparer2`. A `hierarchy` merge mixes levels and has no single order to apply, so it is left as gathered.
- Take the date walk only when the accumulation dimension is a dimension of its own. Otherwise the order lives in the row order and nowhere else, since it may be a dimension `innerDim` cannot see — such as a colour aesthetic, which `GraphUtil.getSeriesDimensionsForCalc` excludes.

## Verification

Both assets exported from the same build:

| Issue | Result |
|---|---|
| #76515 — `Others` running sum | 25/25 rows identical to the accepted `expectData` baseline |
| #74910 — chronological running max | 32/32 rows match a chronological running max recomputed from the exported aggregate values |

#74910's baseline file is from an older data snapshot, so it was verified by recomputing the expected values from the export itself rather than diffing the stale text.

191 tests pass across 51 classes in `report.filter`, `report.composition.execution`, `report.composition.graph` and `report.composition.graph.calc`.

## Tests

6 new tests; `RunningTotalColumn`'s break-by path previously had no coverage at all. Each was verified to fail without its half of the fix:

- `SummaryFilterTest.othersBucketKeepsInnerGroupOrder` / `...Desc` — fail `expected: <3> but was: <5>` without `sortMergedNodes()`
- `RunningTotalColumnTest.testRunningSumBreakByRunsWithinGroup` — fails `expected: <2.0> but was: <4.0>` without the `innerDim != breakBy` guard
- `RunningTotalColumnTest.testRunningMaxBreakByFollowsDateNotRowOrder` — fails `expected: <11.3225> but was: <13.4748>` without the date walk

The last two are the important pair: they pin the two cases apart, so neither fix can be reintroduced in a way that regresses the other.

## Notes for review

- `SummaryFilter` is shared with crosstabs and reports, so this reaches further than a calculator-only change. The suites above pass, but crosstab top-N-with-Others output is worth a look.
- **#74916 is unverified** — I had no repro asset, and its group-ordering change is adjacent to this area.
- Unrelated, noted while tracing: `Plotter.java:337` passes `GTool.getDims(graph)[0]`, the *outermost* dimension, where every other call site passes the innermost. Not triggered by these charts (single X dimension) but it looks like a latent bug worth its own ticket.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/calc/RunningTotalColumn.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/calc/RunningTotalColumn.java
@@ -159,7 +159,13 @@ public class RunningTotalColumn extends AbstractColumn {
 
          Object firstBreakByVal = baseData.getData(breakBy, row);
          long interval = getInterval(data, sortedRow);
-         Object currentInnerVal = (innerDim != null) ? baseData.getData(innerDim, row) : null;
+         // only accumulate by date when the accumulation dimension is a dimension of its
+         // own. when innerDim IS the breakBy column, every row of a group carries the
+         // same date, so a date walk would add the whole group to every row; the order
+         // then lives in the row order and nowhere else -- it may even be a dimension
+         // innerDim cannot see, such as a colour aesthetic. (74910)
+         Object currentInnerVal = innerDim != null && !Tool.equals(innerDim, breakBy)
+            ? baseData.getData(innerDim, row) : null;
 
          if(currentInnerVal instanceof Date) {
             // Root dataset row order may not be chronological when sort-by-value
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):

jshobe-inetsoft (approved):

PR #5083, opened 2026-09-08
Bug #76512: register the embed chart element's providers at the app root

ngovus-inetsoft:
### Problem

The `<inetsoft-chart>` web component renders blank. `EmbedChartComponent` throws `NG0201` (NullInjectorError) during construction:

```
Uncaught ct: NG0201
    at Pu.get (elements.js:11394:42)
    ...
    at dS.ɵfac [as factory] (elements.js:58512:75)
```

### Cause

`main-elements.ts` registered the chart's services only as **route-level** providers, via `provideRouter(embedChartRoutesEager)`. Route providers live in an `EnvironmentInjector` that Angular creates when a route is *activated* — but the element is built with `createCustomElement(EmbedChartComponent, {injector: app.injector})`, which never navigates. That injector therefore never existed, and every service the chart tree injects that is not `providedIn: "root"` failed to resolve:

`ScaleService` (abstract), `MiniToolbarService`, `ShowHyperlinkService`, `ContextProvider`, `DataTipService`, `PopComponentService`, `VSChartService`, `ChartService`, `VSTabService`, `RichTextService`, `FullScreenService`, `CheckFormDataService`, `DndService`.

`main-viewer-element.ts` never hit this because it registers its equivalent provider set at the **application root**, which is why `<inetsoft-viewer>` has always worked.

### Fix

Spread `EMBED_CHART_ROUTE_PROVIDERS` into the `createApplication()` providers, mirroring `main-viewer-element.ts`, and keep `provideRouter([])` for the `Router` / `ActivatedRoute` that `EmbedChartComponent` injects. The shared const stays the single source of truth for the portal's lazy route in `embed-chart.routes.ts`.

`embed-chart.routes-eager.ts` is removed. It existed only so this bundle could get a route without importing `embed-chart.routes.ts`'s dynamic `import()` (Bug #76468); with no routes at all that constraint is satisfied trivially. Comments that referenced it are updated.

Note: `AdhocFilterService` is left in the list unchanged. It injects `Renderer2`, so it can never be instantiated from an environment injector — it was equally inert at route level, and `EmbedChartComponent`'s own component-level `providers` supply the instance the chart tree actually uses.

### Verification

- `npm run test:portal` — 988 + 376 tests pass, including 17 new ones in `embed-chart.route-providers.spec.ts`, which assert the element bundle's root provider set resolves every token the chart tree needs.
- `ng lint portal` — no new warnings on the changed files.
- `ng build elements --configuration production` — output is still a single `main-*.js` (1.92 MB) with `EmbedChartComponent` inline, i.e. no orphaned lazy chunk, so the Bug #76468 bundle shape is preserved.
- Confirmed live: the reporter's `WebComponentCase.html` served from the data space `web-assets` folder now renders the chart, with `<inetsoft-viewer>` on the same page still working.


jshobe-inetsoft (approved):

jshobe-inetsoft (approved):

PR #5084, opened 2026-09-08
Bug #76513: fix deadlock when resizing a crosstab cell in Composer

ngovus-inetsoft:
## Problem

Resizing a cell in a crosstab in Composer, then clicking away and back, leaves the viewsheet stuck on the loading spinner with only a **Cancel** button — no data ever loads. Intermittent: it only happens on the first (slow) execution, and a page reload plus a second resize does not reproduce it.

A `jstack` of the wedged server ends with `Found 1 deadlock.`

## Cause

A lock-order inversion on the `VSCrosstabInfo` monitor.

`AbstractCrosstabVSAQuery.getTableLens0()` held `synchronized(cinfo)` while calling `getAssetBaseTableLens()`. That call reaches `VSAQuery.getDataWithoutSandboxLock()`, which intentionally releases the sandbox lock for the duration of the cache fetch (74001) and re-acquires it afterwards. So for the length of the fetch the thread held the crosstab monitor with the sandbox lock free — the reverse of the order every other path uses.

| Thread | Holds | Waits for |
|---|---|---|
| resize (`changeColumnWidth` → `refreshVSAssembly` → crosstab query) | `VSCrosstabInfo` monitor | sandbox read lock, in `restoreLocks()` |
| format fetch (`FormatPainterService.getFormat` → `refreshMetaData` → `VSCrosstabInfo.update`) | sandbox lock | `VSCrosstabInfo` monitor |

The released-lock window is what lets the second event in, which is why the hang is intermittent and tied to the slow first execution.

## Fix

The monitor exists only to publish the `(aggrs2, rows2, cols2)` runtime ref triple as a unit — it is used in exactly two places in the repo, the writer in `VSCrosstabInfo.update()` and this reader — so it never needed to cover the data fetch.

Hoist `getAssetBaseTableLens(postDrill)` out of the `synchronized(cinfo)` block so the monitor covers only the runtime ref reads.

## Side effects considered

- `base` is never reassigned and nothing inside the `synchronized` block feeds the call, so the change is otherwise behaviour-preserving.
- Another thread's `VSCrosstabInfo.update()` can now interleave between the fetch and the ref reads instead of blocking. The reads stay inside the monitor, so they still see a consistently published triple. This was already possible today: `oldInfo` is cloned before the block, `getAssetBaseTableLens()` itself triggers `updateAssembly`/`refreshMetaData` on the same thread (reentrant), and all the code after the block reads `cinfo` unguarded.
- `ChartVSAQuery` and `TableVSAQuery` have no `synchronized` at all, and `ChartVSAssemblyInfo:257` guards only condition building, so no other query path holds a monitor across its fetch. This is the only affected site.

## Testing

- Confirmed against the reported repro (`crosstab_resize.zip`): resize cell, click away, click back — no longer reproduces, and a fresh `jstack` reports no deadlock.
- `CrosstabVSAQueryTest`, `CrossTabFilterTest`, `ChartVSAQueryTest`, `TableVSAQueryTest`, `CrosstabTreeTableLensTest` — 7 tests, 0 failures.

No automated regression test: `getAssetBaseTableLens` is `private`, so the "monitor not held across the fetch" invariant cannot be asserted without widening API, and a two-thread reproduction needs a full sandbox plus precise interleaving.

## Follow-up (not in this PR)

While diagnosing, an unrelated latent hang turned up: `ComposerVSTableService.changeColumnWidth` steps `for(i = baseCol; i < colCount; i += colSpan)` with `colSpan = endCol - startCol` taken straight from the client payload, and `BaseTableCellModel.setColSpan` legitimately sends `null` for a span of 1. `colSpan == 0` would spin forever inside the sandbox read lock. Left out to keep this fix minimal; worth its own ticket.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
@@ -367,13 +367,22 @@ public abstract class AbstractCrosstabVSAQuery extends CubeVSAQuery
             box.getVariableTable().put(XQuery.HINT_IGNORE_MAX_ROWS, "true");
          }
 
-         TableLens base = null;
          DataRef[] rheaders = null;
          DataRef[] cheaders = null;
          DataRef[] aggregates = null;
 
+         // don't hold the cinfo monitor across this call. it fetches the base data, and
+         // VSAQuery.getDataWithoutSandboxLock() releases and re-acquires the sandbox lock
+         // while doing so (74001). holding the monitor across that window inverts the lock
+         // order used everywhere else -- ViewsheetSandbox.refreshMetaData() updates the
+         // assembly, and so enters VSCrosstabInfo.update(), with the sandbox lock already
+         // held -- so a concurrent event (e.g. fetching a cell format after a resize)
+         // deadlocks: it owns the sandbox lock and waits for the monitor while we own the
+         // monitor and wait for the sandbox lock. the monitor only needs to cover the
+         // runtime ref reads below, which VSCrosstabInfo.update() publishes as a unit.
+         TableLens base = getAssetBaseTableLens(postDrill);
+
          synchronized(cinfo) {
-            base = getAssetBaseTableLens(postDrill);
             DataRef[] aggrs = cinfo.getRuntimeAggregates();
 
             if(aggrs == null || aggrs.length == 0) {
--- a/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/AbstractCrosstabVSAQuery.java
@@ -216,6 +216,12 @@ public abstract class AbstractCrosstabVSAQuery extends CubeVSAQuery
       DataRef[] aggregates = cinfo.getRuntimeAggregates();
       DataRef[] dcTempGroups = cinfo.getDcTempGroups();
 
+      // remember the refs the column selection below is built from, so getTableLens0()
+      // can tell whether the refs it resolves against this table are the same ones. (76513)
+      baseRowHeaders = rheaders;
+      baseColHeaders = cheaders;
+      baseAggregates = aggregates;
+
       // apply sequence
       ColumnSelection columns2 = (ColumnSelection) columns.clone();
       columns.clear();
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (changes requested):
Requesting changes based on Claude's automated review: it identified an unresolved correctness concern in `AbstractCrosstabVSAQuery.java` — hoisting `getAssetBaseTableLens(postDrill)` out of the `synchronized(cinfo)` block removes the deadlock, but it also removes atomicity between that call (which itself reads `cinfo.getRuntimeRowHeaders()`/`getRuntimeColHeaders()` unguarded to pick the base table's column selection) and the later re-read of `rheaders`/`cheaders`/`aggregates` inside the synchronized block. A concurrent `VSCrosstabInfo.update()` landing in that window could now interleave in a way that wasn't previously possible, potentially causing `AssetUtil.findColumn` to fail against a stale/mismatched column selection (surfacing as a 'Column not found' exception). The review's reentrancy argument for 'already possible today' doesn't cover this specific second-thread interleaving case.

Please confirm explicitly whether this window is acceptable (e.g. verify `getRuntimeRowHeaders`/`getRuntimeColHeaders` can't change shape in a way that breaks `AssetUtil.findColumn` during a live resize), or adjust the fix to preserve atomicity across both reads. Full review: https://github.com/inetsoft-technology/stylebi/pull/5084#issuecomment-3269509766

jshobe-inetsoft (changes requested):
Requesting changes: Claude's automated code review flagged an unresolved correctness concern rather than clearing the PR outright — it identifies a new race window in `AbstractCrosstabVSAQuery` (hoisting `getAssetBaseTableLens()` out of the `synchronized(cinfo)` block can let a concurrent `VSCrosstabInfo.update()` interleave between the two field-read sites, which could surface as a `ColumnNotFoundException`/`MessageException` during a live resize) and explicitly asks the author to confirm this window is acceptable or verify it can't actually break `AssetUtil.findColumn`. Since the review didn't affirmatively clear this as non-blocking, please respond to that concern (or add a safeguard) before merge.

jshobe-inetsoft (changes requested):
This PR currently has merge conflicts with `main` and cannot be merged as-is. Please rebase/merge `main` into `bug-76513` and resolve the conflicts. Note there is also a failing `build` check on inetsoft-core — I re-ran it since the specific failures (`ImageVSAScriptableTest`/`ViewsheetScopeTest` image-URL fetch errors) look like unrelated CI environment flakiness rather than something caused by this change, but please double check once the conflicts are resolved and the check re-runs on the new commit. Claude's automated review itself found no blocking issues (only a minor, already-acknowledged double-`dispose()` nit).

ngovus-inetsoft:
resolved by https://github.com/inetsoft-technology/stylebi/pull/5119

PR #5078, opened 2026-09-08
Modernize the composer's object-selection border and resize handles

swaker854:
## Summary
- The composer's selected-object border (`.txt-primary .object`) was `2px dotted`, inheriting the legacy shell-primary orange via `currentColor`. That orange is the same family already flagged as an "orphan color in no palette" elsewhere in this codebase (`.viz-modern .chart-object-canvas`). Switches it to dashed using `--inet-viz-selected-border`, the same modernized, light/dark-aware token `.bd-selected-cell` already uses, so assembly-level and cell-level selection read as one visual language.
- Resize handles switch from a solid accent-filled square to a hollow one (card-surface fill + colored ring), matching the outline-only handle convention modern design tools use.

Stacked after #5021 purely for review sequencing — no code dependency between the two.

## Test plan
- [ ] Composer: select an object with resize handles, confirm the border reads as a dashed teal line instead of the dotted orange one, in both light and dark
- [ ] Confirm resize handles render hollow (visible ring, card-surface fill) rather than solid-filled, and remain easy to grab/drag
- [ ] Confirm an object with its own configured border still looks reasonable when selected (no double-border visual clash)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5085, opened 2026-09-08
Seed Tab's modern chrome, the last unmigrated form-input type

swaker854:
## Summary
- Tab was the one type left with zero `ctx.modern` gating anywhere: it bypasses the base chrome hook (`bypassesBaseChrome()`) but never called `seedChromeDefaults`, so its border color and round corner were hardcoded to legacy values (`DEFAULT_BORDER_COLOR`, `4`) unconditionally, in both modes.
- Splits the mode-dependent color/corner seeding out of `setDefaultFormat` into a proper `seedChromeDefaults(VizContext)` override — structural insets (unchanged between modes) stay in `setDefaultFormat`; color/corner move to `seedChromeDefaults` so they're correctly re-applied by `VizModernizeUtil.reseedAfterRestore` on Modernize/Revert of an *existing* viewsheet, not just at creation of a new one.
- The active-tab indicator (the only visual cue for which tab is currently showing) deliberately does not reuse the neutral `objectBorderColor` that the inactive-tab border modernizes to — that would make active and inactive tabs indistinguishable. Added `VSObjectChromeDefaults.activeIndicatorColor`, a new Java-side constant mirroring the modern selection teal (`--inet-viz-selected-border-modern`/`-dark` in `_viz-tokens.scss`) already used client-side for `bd-selected-cell` and the composer's selection border, so Tab's active state reads as the same "this is selected" visual language.
- Legacy branch of both the border/corner and the active-tab color is unchanged from the original hardcoded values.

Stacked after #5078 for review sequencing — no code dependency between the two.

## Test plan
- [ ] Composer, legacy mode: confirm Tab's border/corner and active-tab indicator render identically to before this change
- [ ] Composer, modern mode: confirm the inactive-tab border and corner match the shared modern object chrome, and the active tab shows the new teal indicator (light and dark)
- [ ] Modernize an existing legacy viewsheet containing a Tab: confirm its chrome updates (not just newly-created Tabs)
- [ ] Revert a modernized viewsheet's Tab: confirm it returns to the legacy hardcoded values


jshobe-inetsoft (changes requested):
Requesting changes: the `build` check is currently failing on the latest commit (b3baecb2540a4e29ec5f39c786a71972842f6251). All checks must pass before merge. Please fix the failing build and push an update.

swaker854:
Follow-up on the earlier build failure: the real regression (`VSObjectChromeDefaultsTest.cardCornerNotSeededForExcludedTypesUnderGate`) is fixed as of `e64998b6` — that test no longer fails.

The [latest run](https://github.com/inetsoft-technology/stylebi/actions/runs/34367157651) still shows `build` failing, but now with different, unrelated failures:

```
ViewsheetScopeTest.testAddImage » IIO Can't get input stream from URL!
ImageVSAScriptableTest.testGetSetImage » IIO Can't get input stream from URL!
JavaScriptEngineTest.testGetImages: expected: not <null>
PermissionHierarchyTest.setUp:96 » IllegalArgument argument "in" is null (the known flake, #4984/#4994)
```

None of these are in this PR's diff (`TabVSAssemblyInfo.java`, `VSObjectChromeDefaults.java`, and their tests). I checked an unrelated branch's build from the same time window (`bug-76518`) and it shows the **identical** three image-test failures — this looks like environment-wide flakiness (an external image URL these tests depend on is currently unreachable in CI), not something introduced here.

Flagging for the CI/infra team alongside the already-tracked `PermissionHierarchyTest` flake.

jshobe-inetsoft (approved):

PR #5086, opened 2026-09-08
Bug #76508: fix the shape drop shadow in Print Layout preview and export

ngovus-inetsoft:
Follow-up to the shape shadow work in #4767 (feature #76073) and #4974 (bug #76416): QA reported that Print Layout preview and PDF export were still wrong.

There turned out to be **two independent defects**, one per rendering path. The shadow was only correct on the viewsheet canvas.

## 1. The Print Layout pane clipped the shadow

The pane reuses the real `<vs-rectangle>`/`<vs-oval>` components, so the CSS `box-shadow` was drawn correctly and then cut off. `.object-container` is `overflow: hidden` and only carries 4px of slack, while a distance-5/blur-6 shadow reaches ~14px past the shape.

A `.shape-shadow` class now opens that container's overflow, gated on a rectangle or oval that actually has a shadow — the hidden overflow is there to keep an oversized object inside its layout cell, and only these two object types paint outside their own box. This follows the override `.horizontal-line`/`.vertical-line` already use for the same class of problem.

## 2. The PDF drew a hard opaque bar, not a shadow

`PageLayout.Rectangle/Oval.paint()` filled an offset copy of the shape and then painted the shape's own opaque fill over it, so all that showed was a hard band the width of the offset. `PDFPrinter` also ignores the alpha of a plain `setColor`, so it came out fully saturated.

#4974 documented the missing blur as an accepted limitation, on the grounds that a Gaussian blur "is not practical in a Graphics-only context". That premise does not hold: `PDFPrinter extends Graphics2D` and emits a real 8-bit `/SMask` soft mask for a translucent image, so a pre-blurred ARGB image drawn with `drawImage` keeps both its blur and its opacity. `Shape.paintShadow()` now renders the silhouette off-screen, runs it through the existing blur, and draws that.

To keep the blur defined in one place, the tint and blur move out of `VSFaceUtil.addShadow` into `ShapeShadowUtil.createShadowLayer()`, which `addShadow` now calls. That relocation is forced rather than cosmetic: `VSFaceUtil`'s static initializer reaches into `PortalThemesManager`, so having `PageLayout` touch that class fails with `ExceptionInInitializerError` outside a configured server. The tint (`AlphaComposite.SrcIn`) and the blur (`ConvolveOp`) both have to happen off-screen anyway, since `PDFPrinter` honours neither.

## 3. The band clip cut off one side

`ShapePaintable.paint()` narrowed the clip to the band/page box before painting, so a shape sitting flush against an edge of its band lost that side of its shadow — the missing bottom band on the large rectangle in the report. Those two `clipRect` calls now open up by the shadow insets.

Deliberately **not** done in `Shape.getBounds()`/`ShapePaintable.getBounds()`, which feed hit testing and `SectionElementDef`'s page-breaking, where widening by up to ~150px would move where pages break.

## Notes

Print Layout preview and export are the same code path — the preview hits the export endpoint with no `format` param, which defaults to PDF — so one fix covers both.

Left alone: the pre-existing `getX() + xs` typo in both fill blocks, where the sibling lines use `getX() * xs`. Nothing calls `Shape.setScale()`, so it degrades to a one-point offset; it predates this bug and wants its own ticket.

## Testing

- `PageLayoutShapeShadowTest` goes from asserting that `paint()` does not throw to asserting the pixels: that the shadow lands on the side the direction names, that it is translucent, that it fades with distance, and that `blur 0` stays hard-edged.
- New `layout-object.component.spec.ts` covers the `hasShapeShadow()` predicate.
- `VSFaceUtilShadowTest`, `ShapeShadowTest`, `ShapeVSAssemblyInfoShadowTest` and the whole `inetsoft.report.**` package pass.
- Verified end-to-end against the reported repro asset: imported it, switched to Print Layout, previewed and exported to PDF. Direction, opacity and blur now match the viewsheet canvas for all three shapes (SE, NE and SW).

### DIFF
--- a/core/src/main/java/inetsoft/report/PageLayout.java
+++ b/core/src/main/java/inetsoft/report/PageLayout.java
@@ -19,13 +19,15 @@ package inetsoft.report;
 
 import inetsoft.graph.internal.GTool;
 import inetsoft.report.internal.*;
+import inetsoft.report.io.viewsheet.ShapeShadowUtil;
 import inetsoft.sree.SreeEnv;
 import inetsoft.uql.viewsheet.ShapeShadow;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.awt.*;
-import java.awt.geom.Point2D;
+import java.awt.geom.*;
+import java.awt.image.BufferedImage;
 import java.io.Serializable;
 
 /**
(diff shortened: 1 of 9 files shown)


jshobe-inetsoft (approved):

jshobe-inetsoft (approved):

PR #5087, opened 2026-09-09
Add inline value-list named-group binding for chart/table fields

lizhepu:
## Summary
- `FieldRef` gains `namedGroupValues` (`{groups: [{name, values}]}`), mutually exclusive with the existing by-name `namedGroup` reference. All prior constructors kept for backward compatibility.
- `FieldRefFactory` builds a `SIMPLE_NAMEDGROUP_INFO` (`SNamedGroupInfo`) directly from the caller's own name+values pairs — the same shape the Composer's own right-click "Group columns" builds — via a new `buildInlineNamedGroupInfo`, dispatched through a shared `resolveNamedGroupInfo(FieldRef, ...)` entry point used by both the chart and table/crosstab binding paths.
- `TableBindingMutator.dimensions()` now resolves the inline form even through the no-runtime-context overload (`addField`/`moveField`), since it needs no worksheet/repository lookup, unlike the by-name form.
- `others` (bucket-the-rest-together) is explicitly refused rather than silently accepted or dropped: `SNamedGroupInfo` has no such concept.
- Pairs with `inetsoft-technology/stylebi-wiz` PR #2320 (same branch name), which declares and validates this shape on the agent-facing tool schema.

## Test plan
- [x] `mvnw compile -pl community/core -am` — clean
- [x] `mvnw test-compile -pl community/core -am` — clean (existing tests still compile against the changed record; no new tests added)
- [ ] Live: an inline named group applied via the paired plugin PR renders correctly on a chart/crosstab (needs both PRs merged and the deployment rebuilt)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/FieldRefFactory.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/FieldRefFactory.java
@@ -42,7 +42,9 @@ import inetsoft.web.composer.model.condition.ConditionExpression;
 import inetsoft.web.composer.model.condition.ConditionUtil;
 import inetsoft.web.wiz.binding.model.FieldRef;
 
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 /** Converts between StyleBI's ref models and the agent-facing {@link FieldRef}. */
 public final class FieldRefFactory {
@@ -175,6 +177,7 @@ public final class FieldRefFactory {
 
       NamedGroupInfoModel model = new NamedGroupInfoModel();
       model.setType(XNamedGroupInfo.SIMPLE_NAMEDGROUP_INFO);
+      Set<String> seenNames = new HashSet<>();
 
       for(FieldRef.NamedGroupValues.Clause clause : spec.groups()) {
          if(clause.name() == null || clause.name().isBlank()) {
@@ -187,6 +190,12 @@ public final class FieldRefFactory {
             throw new IllegalArgumentException(
                "Field '" + column + "'s 'namedGroupValues' group '" + clause.name() + "' needs " +
                "a non-empty 'values' list -- a group with no member values matches nothing.");
+         }
+
+         if(!seenNames.add(clause.name())) {
+            throw new IllegalArgumentException(
+               "Field '" + column + "'s 'namedGroupValues.groups' has a duplicate name '" +
+               clause.name() + "' -- list each group name at most once.");
          }
 
          model.addGroup(new GroupCondition(clause.name(), clause.values()));
--- a/core/src/main/java/inetsoft/web/wiz/binding/FieldRefFactory.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/FieldRefFactory.java
@@ -32,6 +32,7 @@ import inetsoft.uql.util.XNamedGroupInfo;
 import inetsoft.web.binding.drm.DataRefModel;
 import inetsoft.web.binding.model.BAggregateRefModel;
 import inetsoft.web.binding.model.BDimensionRefModel;
+import inetsoft.web.binding.model.GroupCondition;
 import inetsoft.web.binding.model.NamedGroupInfoModel;
 import inetsoft.web.binding.model.graph.ChartAggregateRefModel;
 import inetsoft.web.binding.model.graph.ChartDimensionRefModel;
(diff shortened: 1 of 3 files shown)


lizhepu:
**Round 1 review — verdict: changes requested** (posted as a comment, not a formal review: GitHub refuses `request-changes`/`approve` from the PR author's own account, which is the account this session is authenticated as).

Reviewed head commit 2c6e91b5 (2c6e91b55aae62b8d809eb50f03df0c9d7a0f74e).

**What I checked**

- Confirmed the compile claim myself: `./mvnw -q -o test-compile -pl core -am` from community/ — clean, exit 0.
- Grepped every caller of `resolveNamedGroupInfo` (both overloads), `buildInlineNamedGroupInfo`, `toChartRef`, and every `new FieldRef(` construction site across the whole community module (and confirmed no enterprise module touches any of these classes at all). No other call site calls the old by-name `resolveNamedGroupInfo(String,...)` directly — only the new `resolveNamedGroupInfo(FieldRef,...)` does, so no divergent/bypassing caller. All existing `FieldRef` construction sites use the 5/6/7-arg constructors, all of which are preserved — no compile break, matching the PR's own claim.
- Traced the chart path for the same "no-context" gap `TableBindingMutator.dimensions()` had: `ChartBindingMutator.setShelf/setSingleShelf` and `ChartAestheticMutator.setField` always call the 4-arg `FieldRefFactory.toChartRef(field, rvs, source, refModelService)`, threading rvs/source/refModelService through (even when null, from their own null-context convenience wrappers) rather than routing through a separate no-context overload. So the inline path resolves correctly there too — no analogous gap on the chart side.
- Walked the JSON→Java type path for `values` concretely: a JSON number/boolean deserializes to a non-String `Object`; in `SNamedGroupInfo.getGroupCondition`, a non-String value takes the `cond.addValue(ovalue)` branch directly (no cast, no `ClassCastException`), and `Condition.evaluate`'s `EQUAL_TO` op runs through `DefaultComparer`, whose own javadoc says it "compares two objects as numbers if they are both Number" — so e.g. a JSON `5` (Integer) correctly matches a `Double(5.0)` column value at render time. No crash, no silent mismatch.
- Walked `spec.groups()` null vs. empty and `others` present — both throw cleanly (`IllegalArgumentException`), no NPE either way.
- Walked a `null` entry inside a `Clause.values()` list — it maps to an `XCondition.NULL` condition (matches actual null values in the data), which is the platform's existing behavior for the Composer's own manual "Group columns" feature too — not a new defect.
- Checked `buildInlineNamedGroupInfo` for overfit to the Customer:State/DataGroup example from the design discussion — it's fully general, no hardcoded column/group names.
- Cross-checked the wire format against the paired plugin PR (inetsoft-technology/stylebi-wiz #2320): both sides agree on `{groups: [{name, values}]}` with no `others` emitted, matching the record's field names for Jackson deserialization.

**Findings**

1. **(Important, ~85)** Duplicate group names in `namedGroupValues.groups` are silently accepted and overwrite each other with no validation or error.
   `core/src/main/java/inetsoft/web/wiz/binding/FieldRefFactory.java`, `buildInlineNamedGroupInfo` (~lines 158-196): every `Clause` is added via `model.addGroup(...)` with no check for a name repeated across entries. Downstream, `NamedGroupInfoModel.createNamedGroupInfo`'s `SIMPLE_NAMEDGROUP_INFO` branch (`core/src/main/java/inetsoft/web/binding/model/NamedGroupInfoModel.java`, ~lines 164-175) does `sinfo.setGroupValue(group.getName(), group.getValue())` per entry, which is a `Hashtable.put` keyed by name — so a second `Clause` with the same `name` silently replaces the first's values, and the first group's intended members render as their own ungrouped rows with no diagnostic anywhere. This is exactly the "renders plausibly wrong instead of failing loud" failure mode this same PR's `others` refusal was written to avoid (its own javadoc: "kept as its own key... instead of the option quietly doing nothing"). Suggest rejecting a repeated `name` in `spec.groups()` the same way a blank/missing name is already rejected.

2. **(Important, ~80)** Zero automated test coverage for the entire new code path. Confirmed via grep — no test file under `core/src/test` references `namedGroupValues` anywhere. Acknowledged as a deliberate choice per the PR description, but flagging plainly per review standard: this is a new binding-write path with a nested record and several validation branches, including non-trivial JSON-type interactions (see the Number/Boolean/null walk above) — exactly the surface a regression test earns its keep on, and finding #1 above would have been caught by one.

**Verdict: Request changes.** Finding #1 is a real, user-visible silent-data-loss defect on a brand-new code path; finding #2 is why it shipped undetected.

lizhepu:
Re finding 2 (test coverage): acknowledged, but out of scope for this session -- I'm operating under a standing constraint for this change to make code fixes only, with no new tests or test runs. Fixed finding 1 (duplicate group-name validation) in 82b3655d1. If test coverage for the inline namedGroupValues path is wanted, please request it as a separate follow-up.

fredaan (approved):

PR #5092, opened 2026-09-09
Fix #76527: stale ValueOfColumn.minDate cache across facet cells

larryliang-inetsoft:
## Summary

`ValueOfColumn.minDate` (used by `PREVIOUS_YEAR`/`PREVIOUS_QUARTER`/`PREVIOUS_MONTH`/`PREVIOUS_WEEK` calcs' "first period" guard) was computed lazily and cached forever, never invalidated when the same `ValueOfColumn` instance is reused across a different `DataSet`. In a faceted chart, `DataSetIndex.wrapDataSet()` shares one `ValueOfColumn` instance across all facet cells, each running its own `prepareCalc()`/`calculate()` pass against its own narrower per-cell `DataSet`. `minDate` froze on whichever facet cell was processed first and was then wrongly reused as the guard for every other cell, producing a spurious `INVALID` result (which renders with a fallback/undefined color) for rows that should have had a valid previous-period value — reported against `TrendComparison_Spec`'s `previous2`/`previous4` cases.

Every other per-dataset cache on this class (`subs`/`subsRoot`, `AbstractColumn.cachrouter`) already invalidates correctly on dataset change; `minDate` was the one exception.

## Fix

Null `minDate` out in `complete()`, alongside the existing `subs = null; subsRoot = null;` reset — `complete()` already fires once per facet cell, right after that cell's rows are computed and before the next cell's `prepareCalc()` runs on the same shared instance, so this is sufficient without needing to track dataset identity.

## Test plan

- [x] Added `ValueOfColumnTest.testMinDateNotStaleAcrossFacetCells`: simulates two facet cells sharing one `ValueOfColumn` (cell A = single 2022 row, cell B = 2021+2022 rows matching the existing known-good `testCalculateWithDataSetOfPreviousYear` case), asserting cell B's 2022 row still correctly resolves to its previous-year value after cell A has already run.
- [x] Confirmed the new test fails without the fix (`expected: <4> but was: <4.9E-324>`, i.e. `CalcColumn.INVALID`) and passes with it.
- [x] Full class run: `./mvnw -pl community/core -Dtest=ValueOfColumnTest test` — 24/24 tests pass.

Note: this fixes the confirmed calc-engine staleness defect itself. It does not independently confirm which of `previous2`/`previous4` (or both) it resolves in `TrendComparison_Spec`, since the real chart bindings/fixtures for that suite weren't available in this checkout — worth a live/screenshot-diff re-run of that suite once this lands.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/graph/calc/ValueOfColumn.java
+++ b/core/src/main/java/inetsoft/report/composition/graph/calc/ValueOfColumn.java
@@ -779,6 +779,7 @@ public class ValueOfColumn extends AbstractColumn {
    public void complete() {
       subs = null;
       subsRoot = null;
+      minDate = null;
    }
 
    @Override
--- a/core/src/test/java/inetsoft/report/composition/graph/calc/ValueOfColumnTest.java
+++ b/core/src/test/java/inetsoft/report/composition/graph/calc/ValueOfColumnTest.java
@@ -113,6 +113,52 @@ public class ValueOfColumnTest {
    }
 
    /**
+    * Regression test for Bug #76527: PREVIOUS_YEAR's minDate cache (getMinDate()) must not
+    * leak stale state across facet cells that reuse the same ValueOfColumn instance.
+    *
+    * A faceted chart shares one ValueOfColumn instance across all facet cells
+    * (DataSetIndex.wrapDataSet()), calling calculate() against each cell's own narrower
+    * DataSet in turn. Before the fix, getMinDate() computed its minimum date once and never
+    * recomputed, so a cell whose local data only contained a later year (here, a single
+    * 2022-01-01 row) would freeze minDate at 2022-01-01. A later cell that legitimately
+    * contains both 2021-01-01 and 2022-01-01 would then wrongly have its 2022 row's
+    * "previous year" guard (ValueOfColumn.java ~455-459) evaluate against the stale
+    * 2022-01-01 minDate instead of its own data's true minimum (2021-01-01), incorrectly
+    * returning INVALID instead of the real previous-year value.
+    */
+   @Test
+   void testMinDateNotStaleAcrossFacetCells() {
+      valueOfColumn = new ValueOfColumn("id", "sum(id)");
+      valueOfColumn.setChangeType(ValueOfCalc.PREVIOUS_YEAR);
+      valueOfColumn.setDim("date");
+
+      // Facet cell A: single row, only year 2022.
+      DefaultTableLens cellATable = new DefaultTableLens(new Object[][]{
+         { "date", "id" },
+         { toDate("2022-01-01"), 99 }
+      });
+      VSDataSet cellA = createVSDataSet(cellATable, "date");
+      Object cellAResult = valueOfColumn.calculate(cellA, 0, false, false);
+      assertEquals(CalcColumn.INVALID, cellAResult);
+      valueOfColumn.complete();
+
+      // Facet cell B: same shape as testCalculateWithDataSetOfPreviousYear (known-good in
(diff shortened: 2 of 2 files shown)


larryliang-inetsoft:
## Round 1 review — bug #76527

(Posted as a comment, not a formal GitHub review approval — GitHub rejects
self-approval since this PR was opened under the same account: "Review Can not approve your
own pull request". Verdict below is APPROVE-equivalent.)

Reviewed head commit `3ba62501c` on branch `bug-76527`.

**What I checked** (adversarially, not a rubber stamp — see full notes in
`docs/teams/2026-09-08-bugs-batch5/bug-76527/04-review-r1.md`):

1. Read `ValueOfColumn.java` in full (899 lines) and `AbstractColumn.java` in full, looking
   for other lazily-cached, data-derived fields with `minDate`'s pre-fix defect shape
   (compute-once, never-revalidated). Found none: `AbstractColumn`'s own caches
   (`data`/`basedata`/`alldata`/`isBrush`, `cachrouter`) all self-invalidate via an
   identity/hash check against the *current* `DataSet` argument on every call
   (`this.data != data`, `cachrouter.isValidFor(routerData)` →
   `keyhash == dataSet.hashCode()`), unlike the old `minDate`, which had no such check.
2. Checked `ChangeColumn` (the only class that `extends ValueOfColumn` in the repo) — it does
   not override `complete()`, so the fix applies to it transitively.
3. Traced the real call chain for the "does `calculate()` ever run without an intervening
   `complete()` between dataset scopes" question: every `prepareCalc()` call
   (`AbstractDataSet.java:52`, `synchronized`) fully computes its row loop then
   unconditionally calls `complete()` on every non-null calc (`AbstractDataSet.java:182-186`)
   before returning — true for whole-dataset passes (`GraphGenerator.java:272/276`),
   per-facet-cell passes (`RectCoord.java:1549`), and line/stack-specific passes
   (`LineElement.java:150/473`). Grepped facet rendering code for parallelism
   (`ExecutorService`/`CompletableFuture`/`Thread(`/`parallel`) — none found; facet iteration
   is a plain sequential nested loop. Could not construct a counter-example call path.
4. Grepped every call site of `getMinDate()` repo-wide — exactly one, inside `ValueOfColumn`
   itself — so no other code relies on the old sticky behavior.
5. Traced `getMinDate()`'s zero-row boundary case: pre-existing sentinel behavior
   (`Long.MAX_VALUE`), not worsened by this diff — if anything the fix removes an
   order-dependent inconsistency for empty cells rather than introducing one. The
   `dval instanceof Date` null-safety guard is untouched by the diff.
6. Confirmed minimal scope and syntactic correctness directly from `git diff
   origin/main..HEAD --stat` (2 files, +47/-0) and the raw hunk (one line, balanced,
   consistent with the adjacent `subs = null; subsRoot = null;` statements).
7. Ran `./mvnw -pl core -Dtest=ValueOfColumnTest test` myself — `Tests run: 24, Failures: 0,
   Errors: 0` — independently confirms the fix doc's reported pass result.
8. Minor (non-blocking) nitpick on the new regression test: cell A (1 row) vs. cell B (2
   rows) also differ in row count, so a hypothetical narrower fix keyed off "row count
   changed" could coincidentally also pass this specific test. Not a real concern for the
   fix as shipped (a correct, general, unconditional one-liner), but a same-row-count variant
   would close this gap entirely if anyone wants to strengthen it later.

**No blocker or important findings.** The fix (`minDate = null;` in `complete()`) is
minimal, syntactically correct, matches the existing invalidation pattern already used for
`subs`/`subsRoot` in this class and for equivalent caches in sibling classes
(`RunningTotalColumn`, `MovingColumn`), applies transitively to `ChangeColumn`, has no other
callers depending on the old behavior, and is independently test-verified.

**Verdict: APPROVED.**

fredaan (approved):

PR #5093, opened 2026-09-09
Bug #76506: fix DataTipView popup positioned outside viewport after Enlarge

larryliang-inetsoft:
## Summary
- Clicking "Enlarge" on a Data Tip View's mini-toolbar to switch a container to max mode could leave the data tip popup positioned partly/fully outside the browser viewport.
- `VSDataTipDirective.ngDoCheck()`'s boundary/overflow check compared the popup's target position against a container size that could be stale in two ways:
  1. When the tip content has a registered container (`popContainerName`), the check used `PopComponentService`'s cached `PopInfo` width/height. That cache is written exactly once, at `Add*ObjectCommand` time, using the container's design-time `objectFormat.width`/`height`, and is never refreshed (e.g. when the container later enters max mode and grows well past that original size).
  2. Independently, `clientWidth`/`clientHeight` were read for the boundary check before `display` was flipped from `"none"` to `"block"` on a fresh trigger, which could measure a 0x0 box and skip the overflow correction entirely.
- Fix: measure the container's actual on-screen size via its live DOM element (`document.getElementById` + `clientWidth`/`clientHeight`) instead of the frozen `PopInfo` snapshot, falling back to the cached value only if the element can't be found; and read size only after `display` is set to `"block"`.

## Test plan
- [x] Added `vs-data-tip.directive.spec.ts` coverage:
  - the boundary check measures size only after `display` is set to `"block"`, not while still `display:none`
  - the boundary check uses the container's live on-screen size (via a `#Container1_main` stand-in element) rather than a stale cached `PopInfo` snapshot
- [x] `npx ng test portal --include='**/vs-data-tip.directive.spec.ts'` — 7/7 passed
- [x] `npx ng test portal --include='**/data-tip/**/*.spec.ts' --include='**/vs-object-container*.spec.ts' --include='**/mini-toolbar*.spec.ts'` — 13/13 passed, no regressions
- [ ] Not verified with a live browser in this run (no bundled sample "chart" viewsheet asset exists in this repo to reproduce with); a live check of the actual repro (Enlarge → click yellow bar, narrow viewport) would still be valuable to confirm end-to-end.

Redmine #76506

### DIFF
--- a/web/projects/portal/src/app/vsobjects/objects/data-tip/vs-data-tip.directive.spec.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/data-tip/vs-data-tip.directive.spec.ts
@@ -138,4 +138,133 @@ describe("VSDataTip Directive Tests", () => {
 
       expect(markForCheckSpy).not.toHaveBeenCalled();
    });
+
+   // Bug #76506: on a fresh trigger (transitioning from the "not active" branch's
+   // display:none), the boundary check must measure the element's size only after display
+   // has been set back to "block" -- otherwise clientWidth/clientHeight read 0 and the
+   // overflow correction never applies.
+   it("should measure size only after display is set to block, not while still display:none", () => {
+      fixture.detectChanges();
+      dataTipService.isCurrentDataTip.mockReturnValue(false);
+      fixture.detectChanges();
+      const el: HTMLElement = div.nativeElement;
+      expect(el.style.display).toBe("none");
+
+      let displayWhenWidthMeasured: string | undefined;
+      let displayWhenHeightMeasured: string | undefined;
+      Object.defineProperty(el, "clientWidth", {
+         configurable: true,
+         get: () => {
+            displayWhenWidthMeasured = el.style.display;
+            return 500;
+         }
+      });
+      Object.defineProperty(el, "clientHeight", {
+         configurable: true,
+         get: () => {
+            displayWhenHeightMeasured = el.style.display;
+            return 500;
+         }
+      });
+
+      dataTipService.isCurrentDataTip.mockReturnValue(true);
+      fixture.detectChanges();
+
+      expect(displayWhenWidthMeasured).toBe("block");
+      expect(displayWhenHeightMeasured).toBe("block");
+      expect(el.style.display).toBe("block");
+   });
+});
+
+// Bug #76506: when the data tip content belongs to a registered container, the boundary
+// check must use the container's *current* on-screen size, not the PopInfo snapshot taken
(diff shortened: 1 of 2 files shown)


larryliang-inetsoft:
## Review round 1 — bug-76506

(Posted as a comment instead of a formal review: `gh pr review --approve` failed with "Can not approve your own pull request" — same GitHub account authored the PR and is running this review.)

Head commit reviewed: `b690df92c`

**What I checked:** full diff, the surrounding un-diffed parts of `ngDoCheck()`, the `getVSObjectId`/`document.getElementById` pattern for cross-instance-collision safety, null-safety of the new `containerElement` lookup, whether the removed later `display:block` write was genuinely redundant, effect on table-family containers, and both new tests' ability to actually fail against the pre-fix code (traced by hand). Ran `npx ng test portal --include='**/vs-data-tip.directive.spec.ts'` myself: 7/7 passed.

**Key findings:**
1. `getVSObjectId()` prefixes every id with `viewsheetClient.runtimeId`, and `DataTipService`/`ViewsheetClientService` are both instantiated per-`ViewerAppComponent` instance (component-level `providers`, not `providedIn: 'root'`) — so two viewsheets open on the same page cannot produce colliding DOM ids. This is also not a new pattern: the file already used `document.getElementById(getVSObjectId(...))` one line above for the `miniToolbar` case, and `vs-pop-component.directive.ts`/`vs-object-container.component.ts` use the identical idiom.
2. The `containerElement` lookup and its `containerWidth`/`containerHeight` fallbacks are fully null-checked; no new throw path, and zero-size/not-yet-rendered elements degrade gracefully to the pre-existing `else` behavior.
3. Verified by reading the whole method (not just the diff) that nothing between the moved-up `display:"block"` write and the old write's former location touches `display` — the removed later write really was redundant. The `miniToolbar` branch's own `display` write (different element) is untouched.
4. Table-family containers: the new live-DOM-read is equivalent-or-better than the old `objectHeight`-only fallback and additionally fixes width, which had no live fallback at all before. Purely additive, not a regression for the case that already partially worked.
5. Both new tests are genuine regression tests, not vacuous — traced by hand that each would fail against the pre-fix code (display-timing test would see `"none"` at measurement time; container-size test would compute `"301px"` instead of `"1px"`).
6. `*.spec.ts` (not `*.tl.spec.ts`) is the right choice per this repo's own test-type guidance — no HTTP interception, assertions are on directive-applied styles.

No blocker/important findings. Verdict: **APPROVED**. Full writeup: `docs/teams/2026-09-08-bugs-batch5/bug-76506/04-review-r1.md`.

fredaan (approved):

PR #5094, opened 2026-09-09
Bug #76524: watchdog to recover a wedged loading mask on a hung STOMP handler

larryliang-inetsoft:
## Summary

Right-click a crosstab cell -> Filter (Adhoc Filter) can permanently wedge the whole viewsheet's loading mask. `ComposerAdhocFilterService`'s crosstab branch calls `ViewsheetSandbox.getVSTableLens()`, which internally upgrades to a write lock via `UpgradableReadWriteLock` -- a plain, unbounded, non-timed `.lock()` call with no timeout capability anywhere in the primitive. If another operation on the same runtime viewsheet is slow or stuck holding/queued for that same write lock, this STOMP request thread blocks forever.

Because `EventAspect.clearLoadingMask`'s `finally` (which sends `ClearLoadingCommand`) only runs when `pjp.proceed()` returns or throws -- never when it just blocks forever -- the loading mask is never cleared. This matches the reported symptom exactly: the whole viewsheet stays stuck in a permanent loading spinner, Cancel doesn't help (it only tears down client-side dialog state, not the still-blocked server thread), and any further action also hangs because it queues behind the same lock.

This is a systemic gap, not specific to the Adhoc Filter path: `getVSTableLens()` has 60+ call sites across `community/core`, none of which guard against this, and `UpgradableReadWriteLock` itself has no bounded-wait primitive at all.

## Fix

Rather than scoping a narrow fix to one call site (which would leave the other 60+ sites and the lock primitive itself exposed to the identical failure mode), this adds a general **watchdog to every `@LoadingMask`-annotated endpoint** in `EventAspect.LoadingMaskAspectTask`:

- After a configurable timeout (`loadingmask.watchdog.timeout` property, default 30 seconds), if the wrapped controller method hasn't completed, the watchdog force-clears the loading mask and sends a warning `MessageCommand` to the client.
- The underlying blocked call is **not** interrupted -- it keeps running on its thread in the background, since killing it outright hasn't been confirmed safe. The UI recovers either way; the stuck operation eventually completes or can be found via a thread dump/logs.
- If the real call does complete after the watchdog already fired, `postprocess()` skips sending a redundant `ClearLoadingCommand` for the same request.
- This applies uniformly whether the endpoint uses the direct (single-node) `@LoadingMask` path or the `asyncProxy` cluster-proxy path, since both funnel through the same `LoadingMaskAspectTask.preprocess()/postprocess()`.

Adding a bounded-wait capability to `UpgradableReadWriteLock`/`ViewsheetSandbox.lockWrite()` itself would address the underlying lock contention more directly, but is a larger, riskier change to the locking model than this hardening fix warrants -- noted here as a follow-up worth considering separately.

## Verification note

**This fix has not been live-verified against the actual reported crosstab repro** -- the `crosstab_adhocfilter1.zip` attachment referenced in the Redmine issue was not available. It addresses the systemic hang-class mechanism (an `@LoadingMask` endpoint whose business logic blocks forever with no bounded wait anywhere in the lock chain), not a live-confirmed fix of this exact crosstab configuration.

## Test plan

- [x] Added `EventAspectWatchdogTest` (`core/src/test/java/inetsoft/web/viewsheet/EventAspectWatchdogTest.java`) simulating a `@LoadingMask` method whose `pjp.proceed()` blocks well past the configured watchdog timeout, asserting the mask is force-cleared and a warning is sent within a bounded time while the blocked call keeps running; also asserts a fast-returning method never triggers the watchdog, and that a real completion after the watchdog fires does not double-send the clear command.
- [x] Confirmed the test fails without the fix (reverted the `EventAspect.java` change locally and reran -- the test times out waiting for the watchdog's command, since nothing clears the mask until the simulated block itself is manually released).
- [x] `./mvnw -pl core test -Dtest=EventAspectWatchdogTest` -- 2/2 pass.
- [x] `./mvnw -pl core test -Dtest="inetsoft.web.viewsheet.**"` -- 62 tests, 0 failures, 0 errors (1 pre-existing skip), confirming no regressions in the surrounding package.

Redmine: #76524

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/ws/WSQueryController.java
+++ b/core/src/main/java/inetsoft/web/composer/ws/WSQueryController.java
@@ -34,8 +34,11 @@ public class WSQueryController extends WorksheetController {
       this.wsQueryServiceProxy = wsQueryServiceProxy;
    }
 
+   // Same rationale as OpenViewsheetController.openViewsheet: this executes the worksheet's
+   // runtime query directly, which the product's query.runtime.timeout=0 default already
+   // treats as legitimately unbounded.
    @InitWSExecution
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @MessageMapping("composer/worksheet/query/run")
    public void runQuery(
       @Payload WSAssemblyEvent event, Principal principal,
--- a/core/src/main/java/inetsoft/web/viewsheet/EventAspect.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/EventAspect.java
@@ -494,7 +494,8 @@ public class EventAspect {
             .filter(CommandDispatcher.class::isInstance)
             .map(CommandDispatcher.class::cast)
             .findFirst();
-      LoadingMaskAspectTask task = new LoadingMaskAspectTask(force);
+      long watchdogTimeoutOverride = mask != null ? mask.watchdogTimeout() : -1;
+      LoadingMaskAspectTask task = new LoadingMaskAspectTask(force, watchdogTimeoutOverride);
 
       if(mask != null && mask.asyncProxy()) {
          ServiceProxyContext.aspectTasks.get().add(task);
@@ -658,8 +659,9 @@ public class EventAspect {
    }
 
    private final class LoadingMaskAspectTask implements AspectTask {
-      public LoadingMaskAspectTask(boolean force) {
+      public LoadingMaskAspectTask(boolean force, long watchdogTimeoutOverride) {
          this.force = force;
+         this.watchdogTimeoutOverride = watchdogTimeoutOverride;
       }
 
       @Override
@@ -740,8 +742,12 @@ public class EventAspect {
                            watchdogDispatcher.sendCommand(new ClearLoadingCommand());
                         }
 
+                        // INFO (not WARNING) is delivered as a non-blocking toast via
+                        // NotificationsComponent -- WARNING routes to a modal dialog the user
+                        // must dismiss, which would contradict this message's own "you may
+                        // continue working" text.
                         MessageCommand message = new MessageCommand();
-                        message.setType(MessageCommand.Type.WARNING);
+                        message.setType(MessageCommand.Type.INFO);
                         message.setMessage(Catalog.getCatalog().getString(
                            "common.loadingMask.watchdog.timeout"));
                         watchdogDispatcher.sendCommand(message);
@@ -779,6 +785,15 @@ public class EventAspect {
       }
 
       private long getWatchdogTimeout() {
+         // A per-endpoint override (set via @LoadingMask(watchdogTimeout = ...)) takes
+         // precedence over the global default -- e.g. 0 disables the watchdog entirely for
+         // endpoints the product already treats as legitimately unbounded (mirroring
+         // query.runtime.timeout=0), while a positive value tunes it for one endpoint without
+         // affecting the global default used by everything else.
+         if(watchdogTimeoutOverride >= 0) {
+            return watchdogTimeoutOverride;
+         }
+
          try {
             return Long.parseLong(SreeEnv.getProperty("loadingmask.watchdog.timeout", "30000"));
          }
@@ -788,6 +803,7 @@ public class EventAspect {
       }
 
       private final boolean force;
+      private final long watchdogTimeoutOverride;
       private final Lock lock = new ReentrantLock();
       private final AtomicBoolean complete = new AtomicBoolean(false);
       private final AtomicBoolean loading = new AtomicBoolean(false);
--- a/core/src/main/java/inetsoft/web/viewsheet/EventAspect.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/EventAspect.java
@@ -21,6 +21,7 @@ import inetsoft.analytic.composition.ViewsheetService;
 import inetsoft.analytic.composition.event.CheckMissingMVEvent;
 import inetsoft.mv.MVSession;
 import inetsoft.report.composition.*;
+import inetsoft.sree.SreeEnv;
 import inetsoft.sree.internal.SUtil;
 import inetsoft.sree.security.*;
 import inetsoft.uql.XPrincipal;
@@ -714,6 +715,45 @@ public class EventAspect {
                }
             }, 1000, 1000);
          }
+
+         // Watchdog: the wrapped method (pjp.proceed()) runs synchronously on this thread and
+         // may never return -- e.g. it can block forever on a lock held by some other, unrelated
+         // slow/stuck operation against the same runtime viewsheet (no bounded-wait primitive
+         // exists in that lock today, see UpgradableReadWriteLock). If that happens, this
+         // postprocess() is never reached and the client's loading mask is wedged permanently.
+         // Force-clear the mask and surface an error after a bounded wait, independent of what
+         // is actually stuck, so the endpoint degrades to a recoverable error instead of a
+         // permanent wedge. The underlying call keeps running on this thread in the background;
+         // it is not interrupted, since it may not be safe to do so.
+         long watchdogTimeout = getWatchdogTimeout();
+
+         if(watchdogTimeout > 0) {
+            CommandDispatcher watchdogDispatcher = dispatcher.detach();
+            watchdogTask = new TimerTask() {
+               @Override
+               public void run() {
+                  lock.lock();
+
+                  try {
+                     if(!complete.get() && timedOut.compareAndSet(false, true)) {
+                        if(loading.get()) {
+                           watchdogDispatcher.sendCommand(new ClearLoadingCommand());
+                        }
+
+                        MessageCommand message = new MessageCommand();
+                        message.setType(MessageCommand.Type.WARNING);
+                        message.setMessage(Catalog.getCatalog().getString(
+                           "common.loadingMask.watchdog.timeout"));
+                        watchdogDispatcher.sendCommand(message);
+                     }
+                  }
+                  finally {
+                     lock.unlock();
+                  }
+               }
+            };
+            timer.schedule(watchdogTask, watchdogTimeout);
+         }
       }
 
       @Override
@@ -721,7 +761,13 @@ public class EventAspect {
          lock.lock();
 
          try {
-            if(loading.get()) {
+            if(watchdogTask != null) {
+               watchdogTask.cancel();
+            }
+
+            // if the watchdog already fired, it already cleared the mask and warned the
+            // client -- avoid sending a redundant clear for the same show.
+            if(loading.get() && !timedOut.get()) {
                dispatcher.sendCommand(new ClearLoadingCommand());
             }
 
@@ -732,10 +778,21 @@ public class EventAspect {
          }
       }
 
+      private long getWatchdogTimeout() {
+         try {
+            return Long.parseLong(SreeEnv.getProperty("loadingmask.watchdog.timeout", "30000"));
+         }
+         catch(NumberFormatException e) {
+            return 30000L;
+         }
+      }
+
       private final boolean force;
       private final Lock lock = new ReentrantLock();
       private final AtomicBoolean complete = new AtomicBoolean(false);
       private final AtomicBoolean loading = new AtomicBoolean(false);
+      private final AtomicBoolean timedOut = new AtomicBoolean(false);
+      private volatile TimerTask watchdogTask;
    }
 
    public static final class SwitchOrgAspectTask implements AspectTask {
--- a/core/src/main/java/inetsoft/web/viewsheet/LoadingMask.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/LoadingMask.java
@@ -37,4 +37,14 @@ public @interface LoadingMask {
     * Flag that indicates if the annotated method calls an asynchronous proxy method.
     */
    boolean asyncProxy() default false;
+
+   /**
+    * Overrides the global loading-mask watchdog timeout (in milliseconds, see
+    * {@code loadingmask.watchdog.timeout}) for this endpoint. A value of {@code 0} disables
+    * the watchdog for this endpoint entirely -- use this for endpoints that, like the
+    * product's own {@code query.runtime.timeout=0} default, are expected to legitimately run
+    * unbounded (e.g. executing a runtime viewsheet/worksheet query). The default, {@code -1},
+    * means "use the global {@code loadingmask.watchdog.timeout} property".
+    */
+   long watchdogTimeout() default -1;
 }
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/OpenViewsheetController.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/OpenViewsheetController.java
@@ -121,7 +121,11 @@ public class OpenViewsheetController {
     *
     * @throws Exception if the viewsheet could not be opened.
     */
-   @LoadingMask(true)
+   // Opening a viewsheet runs the same runtime query execution the product's own
+   // query.runtime.timeout=0 default already treats as legitimately unbounded (unlike
+   // query.preview.timeout, which only bounds design-time preview) -- so the watchdog's
+   // global default would false-positive on normal, if slow, production reports.
+   @LoadingMask(value = true, watchdogTimeout = 0)
    @MessageMapping("/open")
    @HandleAssetExceptions
    @SwitchOrg
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/VSCalendarController.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/VSCalendarController.java
@@ -104,7 +104,7 @@ public class VSCalendarController {
     * @throws Exception if the selection could not be applied.
     */
    @Undoable
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @MessageMapping("/calendar/clearCalendar/{name}")
    public void clearCalendar(@DestinationVariable("name") String assemblyName,
                              Principal principal, CommandDispatcher dispatcher,
@@ -149,7 +149,7 @@ public class VSCalendarController {
     * @throws Exception if the selection could not be applied.
     */
    @Undoable
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @MessageMapping("/calendar/applyCalendar/{name}")
    public void applyCalendar(@DestinationVariable("name") String assemblyName,
                              @Payload CalendarSelectionEvent event,
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/VSComboBoxController.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/VSComboBoxController.java
@@ -57,7 +57,7 @@ public class VSComboBoxController {
     * @throws Exception if the selection could not be applied.
     */
    @Undoable
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @MessageMapping("/comboBox/applySelection")
    public void applySelection(@Payload VSListInputSelectionEvent event,
                               Principal principal, CommandDispatcher dispatcher,
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/VSRadioButtonController.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/VSRadioButtonController.java
@@ -60,7 +60,7 @@ public class VSRadioButtonController {
     * @throws Exception if the selection could not be applied.
     */
    @Undoable
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @MessageMapping("/radioButton/applySelection")
    public void applySelection(@Payload VSListInputSelectionEvent event,
                               Principal principal, CommandDispatcher dispatcher,
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/VSRefreshController.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/VSRefreshController.java
@@ -49,7 +49,10 @@ public class VSRefreshController {
    /**
     * Refresh a viewsheet
     */
-   @LoadingMask(true)
+   // Same rationale as OpenViewsheetController.openViewsheet: refreshing a viewsheet re-runs
+   // its runtime queries, which the product's query.runtime.timeout=0 default already treats
+   // as legitimately unbounded.
+   @LoadingMask(value = true, watchdogTimeout = 0)
    @MessageMapping("/vs/refresh")
    public void refreshViewsheet(@Payload VSRefreshEvent event, Principal principal,
                                 CommandDispatcher commandDispatcher,
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/VSSelectionContainerController.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/VSSelectionContainerController.java
@@ -62,7 +62,7 @@ public class VSSelectionContainerController {
     * @throws Exception if the selection could not be applied.
     */
    @Undoable
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @MessageMapping("/selectionContainer/update/{name}")
    public void applySelection(@DestinationVariable("name") String assemblyName,
                               @Payload HideSelectionListEvent event,
@@ -83,7 +83,7 @@ public class VSSelectionContainerController {
     * @throws Exception    if failed to move the child
     */
    @Undoable
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @MessageMapping("/selectionContainer/moveChild/{name}")
    public void applySelection(@DestinationVariable("name") String assemblyName,
                               @Payload MoveSelectionChildEvent event,
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/VSSelectionListController.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/VSSelectionListController.java
@@ -56,7 +56,7 @@ public class VSSelectionListController {
     * @throws Exception if the selection could not be applied.
     */
    @Undoable
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @ExecutionMonitoring
    @MessageMapping("/selectionList/update/{name}")
    public void applySelection(@DestinationVariable("name") String assemblyName,
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/VSTextInputController.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/VSTextInputController.java
@@ -52,7 +52,7 @@ public class VSTextInputController {
     * @throws Exception if the selection could not be applied.
     */
    @Undoable
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @MessageMapping("/textInput/applySelection")
    public void applySelection(@Payload VSListInputSelectionEvent event,
                               Principal principal, CommandDispatcher dispatcher,
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/dialog/VSCheckBoxController.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/dialog/VSCheckBoxController.java
@@ -54,7 +54,7 @@ public class VSCheckBoxController {
     * @throws Exception if the selection could not be applied.
     */
    @Undoable
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @MessageMapping("/checkBox/applySelection")
    public void applySelection(@Payload ApplyCheckBoxSelectionEvent event,
                               Principal principal, CommandDispatcher dispatcher,
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/table/BaseTableLoadDataController.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/table/BaseTableLoadDataController.java
@@ -39,7 +39,10 @@ public class BaseTableLoadDataController {
       this.baseTableLoadDataService = baseTableLoadDataService;
    }
 
-   @LoadingMask
+   // Same rationale as VSRefreshController.refreshViewsheet: this endpoint's sole purpose is
+   // to call getVSTableLens() and force a fresh runtime query, which the product's
+   // query.runtime.timeout=0 default already treats as legitimately unbounded.
+   @LoadingMask(watchdogTimeout = 0)
    @ExecutionMonitoring
    @MessageMapping("/table/reload-table-data")
    public void eventHandler(@Payload LoadTableDataEvent event, Principal principal,
--- a/core/src/main/java/inetsoft/web/viewsheet/controller/table/CrosstabDrillController.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/controller/table/CrosstabDrillController.java
@@ -40,8 +40,11 @@ public class CrosstabDrillController {
       this.crosstabDrillService = crosstabDrillService;
    }
 
+   // Same rationale as VSRefreshController.refreshViewsheet: drilling forces a fresh
+   // getVSTableLens() re-execution of the crosstab's runtime query, which the product's
+   // query.runtime.timeout=0 default already treats as legitimately unbounded.
    @Undoable
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @MessageMapping("/table/drill")
    public void eventHandler(@Payload DrillEvent event, Principal principal,
                             CommandDispatcher dispatcher, @LinkUri String linkUri)
@@ -52,7 +55,7 @@ public class CrosstabDrillController {
    }
 
    @Undoable
-   @LoadingMask(true)
+   @LoadingMask(value = true, watchdogTimeout = 0)
    @MessageMapping("/table/drill/cells")
    public void drill(@Payload DrillCellsEvent event, Principal principal,
                      CommandDispatcher dispatcher, @LinkUri String linkUri)
--- a/core/src/main/java/inetsoft/web/vswizard/controller/VSWizardBindingController.java
+++ b/core/src/main/java/inetsoft/web/vswizard/controller/VSWizardBindingController.java
@@ -65,7 +65,7 @@ public class VSWizardBindingController {
       vsWizardBindingServiceProxy.getBindingTree(id, event, dispatcher, principal);
    }
 
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @Recommend
    @HandleWizardExceptions
    @MessageMapping("/vswizard/binding/tree/node-changed")
--- a/core/src/main/java/inetsoft/web/vswizard/controller/VSWizardDialogController.java
+++ b/core/src/main/java/inetsoft/web/vswizard/controller/VSWizardDialogController.java
@@ -48,7 +48,7 @@ public class VSWizardDialogController {
       this.securityEngine = securityEngine;
    }
 
-   @LoadingMask
+   @LoadingMask(watchdogTimeout = 0)
    @MessageMapping("/vswizard/dialog/open")
    public void createRuntimeSheet(@Payload OpenVsWizardEvent event,
                                   @LinkUri String linkUri,
--- a/core/src/main/resources/inetsoft/util/srinter.properties
+++ b/core/src/main/resources/inetsoft/util/srinter.properties
@@ -3376,6 +3376,7 @@ common.limited.column.wizard=The data is limited to {0} columns, so new componen
 common.limited.rows=Data is limited to {0} rows.
 common.limited.text=The text is limited to {0} characters.
 common.limited.user=The number of users in the organization is limited to\: {0}.
+common.loadingMask.watchdog.timeout=This request is taking longer than expected and may still be running in the background. You may continue working; if the result does not appear, please try again.
 common.logicalModelNotFound=Logical model does not exist\: {0} in {1}
 common.mail.link.description.dashboard=Click the link(s) below to view this Dashboard.
 common.mail.sendFailed=Failed to send emails to {0}. Detailed information\:
--- a/core/src/test/java/inetsoft/web/viewsheet/EventAspectWatchdogTest.java
+++ b/core/src/test/java/inetsoft/web/viewsheet/EventAspectWatchdogTest.java
@@ -90,6 +90,10 @@ import inetsoft.web.viewsheet.controller.table.CrosstabDrillController;
 import inetsoft.web.viewsheet.event.table.DrillCellsEvent;
 import inetsoft.web.viewsheet.event.table.DrillEvent;
 import inetsoft.web.viewsheet.event.table.LoadTableDataEvent;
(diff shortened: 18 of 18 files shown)


larryliang-inetsoft:

(Posted as a comment, not a formal review: `gh pr review --request-changes` was rejected — "Can not request changes on your own pull request" — since this PR was authored by the same account running this review.)

Head commit reviewed: `9ba570d2be8ff06cd6bea62f45df9b8c148e28f9`

**What I checked:** read `EventAspect.java` in full (not just the diff), traced the `lock`/`complete`/`timedOut`/`watchdogTask` guard through both race interleavings against `TimerTask.cancel()`'s real semantics, read `ClearLoadingCommand`/`viewsheet-pane.component.ts`'s loading-count logic directly, traced what actually happens on the client when the watchdog's `WARNING` `MessageCommand` arrives (`command-processor.ts`), grepped the whole `community/core` tree for `@LoadingMask`/`asyncProxy` usage, checked existing `SreeEnv` timeout conventions (`query.runtime.timeout` vs `query.preview.timeout`), verified `CommandDispatcher.sendCommand()` can't blow up the shared static `Timer` thread, and independently reran both `EventAspectWatchdogTest` (2/2 pass) and `inetsoft.web.viewsheet.**` (62 run, 0 failures, 1 skip — matches the fix doc).

**No blocker in the concurrency design.** The shared `ReentrantLock` fully serializes the watchdog `TimerTask` against `postprocess()` — I worked through both "postprocess wins" and "watchdog wins" interleavings against `TimerTask.cancel()`'s actual documented behavior, and neither produces a double `ClearLoadingCommand` nor a missed cancellation. Good design; reuses the class's existing pattern correctly.

**Two related IMPORTANT findings, both about the default/severity of the false-positive case:**

1. **The `WARNING` message is delivered as a blocking modal dialog, not a toast.** `processMessageCommand0`'s default branch (`command-processor.ts:191-193`) calls `ComponentTool.showMessageDialog(...)`, an NgbModal dialog the user must dismiss. This directly contradicts the watchdog's own message text — *"You may continue working"* — since a modal overlay is the opposite of that.

2. **The 30s default applies even to endpoints the product already treats as legitimately unbounded.** The watchdog lives in the shared `LoadingMaskAspectTask`, so it also covers `OpenViewsheetController.openViewsheet` (`@LoadingMask(true)`, forced immediate mask). But `defaults.properties` sets `query.runtime.timeout=0` (unlimited) as the existing, deliberate default for runtime query execution — as opposed to `query.preview.timeout=30` for design-time preview only. A normal (if heavy) production viewsheet open that legitimately runs past 30s — a scenario the product's own defaults are built to tolerate — will now trip the watchdog and pop up the disruptive modal from finding #1, telling the user something's wrong when it isn't. There's no per-endpoint override; the timeout is one global property.

Together these mean a customer running a normal, if slow, production report is likely to get an alarming, interruptive dialog during a perfectly healthy load. I'd like to see this addressed — e.g. a materially higher default, a non-blocking presentation for the watchdog's own warning specifically, and/or a documented way to exempt/raise the threshold for known-slow endpoints like `openViewsheet` — before this ships as-is.

**Minor, non-blocking:** the fix doc's claim that the watchdog is verified across "both the direct path and the asyncProxy cluster-proxy path" overstates things — grepping all 157 `@LoadingMask` usages in `community/core` turns up zero uses of `asyncProxy=true`, so only the direct path is actually exercised anywhere in this codebase today (including by this bug's own `ComposerAdhocFilterController.addTableFilter`). Doesn't affect correctness, just a doc-accuracy note.

Full writeup: `docs/teams/2026-09-08-bugs-batch5/bug-76524/04-review-r1.md` in the worktree.

larryliang-inetsoft:
Addressed both IMPORTANT findings in commit da8360eac.

**Finding #3 (blocking modal contradicts "you may continue working")** — confirmed independently: `command-processor.ts`'s `processMessageCommand0` default branch (`WARNING` falls through to `default`) calls `ComponentTool.showMessageDialog(...)`, a modal the user must dismiss. Rather than inventing a new delivery mechanism, I found the codebase already has exactly the non-blocking pattern needed: `viewsheet-pane.component.ts`'s `processMessageCommand` and `ws-pane.component.ts`'s `processMessageCommand` both route `MessageCommand.Type.INFO` to `this.notifications.info(...)` — the existing `NotificationsComponent` toast/alert-banner (auto-dismissing, non-blocking), already used elsewhere for things like save-success and "container not focused" hints. Changed the watchdog's `MessageCommand` from `Type.WARNING` to `Type.INFO` (`EventAspect.java`). No frontend changes were needed since the INFO routing to the toast already exists on both panes. Verified `Tool.shouldIgnoreMessage`'s text/column-limit suppression logic (which also gates the INFO branch in `viewsheet-pane.component.ts`) only filters messages containing specific limit-prefix markers, so it won't swallow the watchdog's message.

**Finding #4 (global 30s default fires even for legitimately-unbounded endpoints)** — added a per-endpoint override to `@LoadingMask`: a new `long watchdogTimeout()` attribute, default `-1` (inherit the global `loadingmask.watchdog.timeout` property), `0` disables the watchdog entirely for that endpoint, and a positive value overrides the global default for just that endpoint. This mirrors the existing `query.runtime.timeout=0` vs `query.preview.timeout=30` precedent already in this codebase (`defaults.properties`) rather than raising the global number. Applied `watchdogTimeout = 0` to the three `@LoadingMask` endpoints that directly execute runtime queries and are therefore already treated as legitimately unbounded by the product: `OpenViewsheetController.openViewsheet`, `VSRefreshController.refreshViewsheet`, and `WSQueryController.runQuery`. (`WSQueryController.stopQuery` is a fast cancel operation and was left with the default.) I did not find any other `@LoadingMask` site that itself directly executes a runtime query — the rest are UI/property/dialog operations that are reasonably expected to be fast.

**Tests**: extended `EventAspectWatchdogTest` with two new cases — `watchdogTimeoutZeroOverrideDisablesWatchdog` (global timeout set short, per-endpoint override 0, blocks well past the global timeout, asserts the watchdog never fires and normal postprocess-clear still happens once unblocked) and `watchdogTimeoutOverrideUsesAnnotationValueInsteadOfGlobal` (global timeout set long, per-endpoint override short, asserts the watchdog fires on the shorter override). Updated the existing `watchdogClearsMaskAndWarnsWhenProceedBlocksPastTimeout` test to assert `Type.INFO` instead of `Type.WARNING`.

**Results**: `./mvnw -pl core test -Dtest=EventAspectWatchdogTest` → 4/4 pass. `./mvnw -pl core test -Dtest="inetsoft.web.viewsheet.**"` → 64 run, 0 failures, 0 errors, 1 pre-existing skip (2 more tests than round 1's 62, matching the 2 new cases added; no regressions).

No disagreement with either finding — both were accurate and independently re-verified against the actual frontend/backend code before fixing. Findings #1, #2, #5, #6 required no action per the review verdict.

larryliang-inetsoft:
## Round 2 review — CHANGES REQUESTED

**Head commit reviewed:** `da8360eac` | **SHA range since round 1:** `9ba570d2b..da8360eac`

### Round 1 findings status
- **Finding #3 (WARNING message → blocking modal): FIXED.** Independently verified `WARNING`→`INFO` in `EventAspect.java`, and that `INFO` routes to a non-blocking toast (`NotificationsComponent`) in all three relevant surfaces: `viewsheet-pane.component.ts`, `ws-pane.component.ts`, **and** `viewer-app.component.ts` (the Portal Viewer's own global command handler — not checked by the fix doc or round 1, but also correctly routes `INFO`→toast independently). Confirmed `Tool.shouldIgnoreMessage` won't swallow this message (only filters TEXT_LIMIT/COLUMN_LIMIT markers).
- **Finding #4 (global 30s default fires for legitimately-unbounded endpoints): FIXED for the 3 named endpoints, mechanism verified end-to-end** (annotation → reflection lookup → constructor → `getWatchdogTimeout()` → `if(watchdogTimeout > 0)` gate on scheduling the `TimerTask` — confirmed `watchdogTimeout=0` truly disables it, not just shortens it) **— but incompletely applied.** See new finding below.

### New finding (important, but narrower impact than round 1's #3+#4)

The fix doc's claim "No other `@LoadingMask` site was found that itself directly executes a runtime query" is inaccurate. Two more sites call the exact same `ViewsheetSandbox.getVSTableLens()` path implicated as the hang mechanism, to serve their primary purpose (not incidentally):

- `CrosstabDrillController` — `/table/drill` (`@LoadingMask`) and `/table/drill/cells` (`@LoadingMask(true)`) — drilling forces `processChange()` + a fresh `getVSTableLens()` re-execution.
- `BaseTableLoadDataController.eventHandler` — `/table/reload-table-data` (`@LoadingMask`) — calls `BaseTableService.loadTableData` → `getVSTableLens()`.

Both are at least as "legitimately unbounded" as `openViewsheet`/`refreshViewsheet`/`runQuery` by the fixer's own stated criterion — arguably more on-point, since this bug report itself originates from a crosstab drill/filter interaction. Neither was given `watchdogTimeout = 0`.

**Severity note:** with finding #3 now fixed, the consequence here is a dismissible, non-blocking, and arguably *accurate* `INFO` toast for a slow-but-legitimate query — not round 1's disruptive modal. It doesn't undermine the core wedge→recoverable fix. But since round 1 explicitly asked for this class of endpoint to be found, and the fix is a one-line annotation change per method (same pattern used 3x already), requesting it be closed in round 3 rather than deferred.

### Verification performed
- Confirmed `da8360eac` is the actual PR head; diffed it directly rather than trusting the fix doc's paraphrase.
- Re-ran tests independently: `EventAspectWatchdogTest` → **4/4 pass**; `inetsoft.web.viewsheet.**` → **64 run, 0 failures, 0 errors, 1 skip** — both match the fix doc's reported numbers exactly.
- Read the updated/new tests end-to-end — they genuinely assert `Type.INFO` (not just "a message sent"), assert the watchdog never fires with `watchdogTimeout=0` even 5x past the global timeout, and assert the override value (not global) governs firing time.
- Grepped for other reflective consumers of `@LoadingMask` (none besides `EventAspect.java` — the new `watchdogTimeout()` attribute needed no other wiring) and for other `MessageCommand.Type.WARNING`/tests that might assume the old type (none affected by this change).

Full writeup: `docs/teams/2026-09-08-bugs-batch5/bug-76524/04-review-r2.md`

larryliang-inetsoft:
## Round 3 — response to review r2 (finding #6)

Verified the finding directly before fixing: `CrosstabDrillController.eventHandler` (`/table/drill`, `@LoadingMask`) and `CrosstabDrillController.drill` (`/table/drill/cells`, `@LoadingMask(true)`) both delegate to `CrosstabDrillService`/`BaseTableDrillService.processDrill()`, which calls `box.getVSTableLens(name, false)` (`CrosstabDrillService`/`BaseTableDrillService.java` `getVsTableLens()` helper, `CrosstabDrillHandler.java:159`) — a forced fresh re-execution of the crosstab's runtime query. Separately, `BaseTableLoadDataController.eventHandler` (`/table/reload-table-data`, `@LoadingMask`) calls `BaseTableService.loadTableData()`, whose entire body is `box.get().getVSTableLens(oname, detail)` (`BaseTableService.java:180`). Neither had a `watchdogTimeout` override. Confirmed via reading the actual source, not just the review doc's claim.

**Independent sweep beyond the reviewer's two named sites**: grepped every `@LoadingMask` site (158 total) for `getVSTableLens`/`getData(`/`VSAQuery` transitively, then read each candidate call site to judge "primary purpose = force a fresh runtime query" vs. "incidental refresh after a fast structural change":

- `CrosstabDrillActionController` (`/crosstab/action/drill`) also extends `BaseTableDrillService` but its `eventHandler` delegates to `VSDrillHandler.processDrillAction`, which does **not** call `getVSTableLens`/`getData` at all — confirmed via grep, no match. Left alone.
- `BaseTableFlyoverController`, `BaseTableShowDetailsController`, `BaseTableSortColumnController` call `box.getData(name)` to read the *already-computed* current lens (to build a condition list), not to force a fresh query — same "incidental" category the round-2 reviewer already excluded for `ComposerVSTableController`'s 9 sites. Left alone.
- The many other callers of the shared `BaseTableService.loadTableData(...)` static helper (`VSRefreshService`, `ComposerVSTableService`, `WorksheetEventUtil`, `VSLayoutControllerService`, etc. — dozens of call sites) all use it as a generic "reload the display after some other change" utility, not as their endpoint's primary purpose. Left alone, consistent with not overcorrecting into exempting things that should stay protected.

No additional sites beyond the reviewer's two were found to fit the exemption criterion.

**Fix**: applied `watchdogTimeout = 0` to all three confirmed endpoints:
- `CrosstabDrillController.eventHandler` (`/table/drill`)
- `CrosstabDrillController.drill` (`/table/drill/cells`)
- `BaseTableLoadDataController.eventHandler` (`/table/reload-table-data`)

**Test**: added `crosstabDrillAndTableReloadEndpointsHaveWatchdogDisabled` [G6] to `EventAspectWatchdogTest.java` — reflects on the real, shipped controller methods (not a synthetic fixture) and asserts each still carries `@LoadingMask(watchdogTimeout = 0)`, so a future edit that drops the attribute is caught here.

**Test commands and results**:
- `./mvnw -pl core -am -DskipTests install` (from `community/`) — BUILD SUCCESS.
- `./mvnw -pl core test -Dtest=EventAspectWatchdogTest` — 5/5 pass (4 pre-existing + 1 new).
- `./mvnw -pl core test -Dtest="inetsoft.web.viewsheet.**"` — 65 run, 0 failures, 0 errors, 1 pre-existing skip (1 more than round 2's 64, matching the new test — no regressions).

Commit: `d4c5507e0` ("Bug #76524 round 3: exempt crosstab drill + table reload from loading-mask watchdog"), pushed to `bug-76524`.

larryliang-inetsoft:
## Round 3 review — CHANGES REQUESTED

**Head commit:** `d4c5507e0` | **SHA range since round 2:** `da8360eac..d4c5507e0`

### Round 2's findings: verified fixed
Independently confirmed in source (not just the fix doc) that `CrosstabDrillController.eventHandler`/`drill` and `BaseTableLoadDataController.eventHandler` correctly carry `@LoadingMask(watchdogTimeout = 0)` on the right method-level annotations, wired through the same mechanism verified in round 2 (unchanged — this commit doesn't touch `EventAspect.java`/`LoadingMask.java`). The new reflection-based test (`crosstabDrillAndTableReloadEndpointsHaveWatchdogDisabled`) genuinely guards these three real controller methods, not a restatement of the existing synthetic-fixture tests.

### New finding (important) — one more endpoint missed by the "158 sites, no gaps" sweep

`VSWizardBindingController.bindingTreeNodeChanged` (`/vswizard/binding/tree/node-changed`, plain `@LoadingMask`, no override) fits the identical exemption criterion used for the four already-exempted endpoints:

- Its handler (`VSWizardBindingService.bindingTreeNodeChanged`) opens with `box.get().lockWrite()` — the literal `UpgradableReadWriteLock` unbounded-wait primitive that is this bug's own root-cause mechanism.
- It then calls `refreshBindingRefs`, which calls `lockWrite()` again, then `box.get().resetDataMap(VSWizardConstants.TEMP_CHART_NAME)` — an explicit cache invalidation (the same "dirty-then-refetch" pattern that made the crosstab-drill case count as "force fresh" rather than "incidental" in this round's own analysis) — followed by `updateTemporaryFields(...)`, which reaches `CardinalityExecutor`/`HierarchyExecutor`/`IntervalExecutor`. I read all three directly: each calls `box.getVSTableLens(crosstab.getAbsoluteName(), false)` against the just-invalidated cache — a genuine forced re-execution of the runtime query, done purely to compute chart-recommendation stats from real data.

Same severity calibration as round 2's finding #6: since the blocking-modal issue (finding #3) is already fixed, this produces a dismissible `INFO` toast for a legitimately-slow-but-working operation, not a functional regression — narrow, low-risk, and fixable with the same one-line `watchdogTimeout = 0` pattern already used four times.

### Independent sweep beyond the fixer's claim
Cross-checked the broader `getVSTableLens`/`.getData(`/`VSAQuery` surface across all `@LoadingMask` controllers (31 files hit `getVSTableLens`). The other candidates I checked (`ComposerGroupColumnsController`, `PresenterPropertyDialogController`, `HyperlinkDialogController`, `HighlightDialogController`, `CalcTablePropertyDialogController`, `FormatPainterController`, `VSAnnotationService`, and 12 more via the broader `.getData(`/`VSAQuery` grep) are either not `@LoadingMask`-annotated at all, or read the current/cached lens without a preceding cache-invalidation call (confirmed `ViewsheetSandbox.getVSTableLens()` checks its cache first and only re-executes when absent) — same "incidental" category already correctly excluded. Also independently re-verified `CrosstabDrillActionController` is correctly *not* exempted (its `VSDrillHandler` delegate has zero `getVSTableLens`/`getData`/`processChange` calls).

### Tests (run independently)
- `EventAspectWatchdogTest`: **5/5 pass**
- `inetsoft.web.viewsheet.**`: **65 run, 0 failures, 0 errors, 1 pre-existing skip**

Both match the fix doc's reported numbers.

### On the sweep methodology
This is the second consecutive round where a claimed-exhaustive endpoint sweep missed a real site in the same category (round 2 missed 2; this claimed-complete 158-site sweep missed a 3rd, in a subsystem — VS Wizard — neither the fixer nor round 2 had looked at). The exemption mechanism itself is correct and cheap; the recurring gap is in *finding* every site. Not blocking on this, but flagging for consideration: a way to mark "forces a fresh runtime-query execution" as a structural, greppable category, or inverting the default so the watchdog is opt-in for known-fast endpoints rather than opt-out for the open-ended set of unbounded-query endpoints, might converge more reliably than continuing to whack-a-mole individual sites.

### Verdict
CHANGES REQUESTED — please apply `watchdogTimeout = 0` to `VSWizardBindingController.bindingTreeNodeChanged` (same pattern as the other four). Everything else from rounds 1-2 is correctly and completely fixed.

Full write-up: `docs/teams/2026-09-08-bugs-batch5/bug-76524/04-review-r3.md`

larryliang-inetsoft:
## Round 4: systematic call-graph sweep (response to review r3)

Per the lead's directive, this round replaces the endpoint-by-endpoint "grep + manually judge" approach (which missed a gap in each of rounds 2 and 3) with a systematic, call-graph-based sweep.

### Methodology (repeatable — a future round can re-run this exactly)

1. Enumerated the real risk primitives in `community/core`, distinguishing them from unrelated lookalikes:
   - `ViewsheetSandbox.lockWrite()` (via `box.lockWrite()`/`box.get().lockWrite()`/`vbox.lockWrite()`) — the actual unbounded `UpgradableReadWriteLock` implicated as this bug's mechanism. **Not** `Cluster.lockWrite(name)` (`IgniteCluster`/`MockCluster`/`BlobStorage`/`MVSingleDispatcher`/`MVCompositeDispatcher`) — that's an unrelated distributed named-lock.
   - `ViewsheetSandbox.getVSTableLens(name, false)` and `resetDataMap(...)` — grepped every caller (`\.getVSTableLens\(` -> 28 files, `\.resetDataMap\(` -> 16 files).
   - `CardinalityExecutor.getData`/`HierarchyExecutor.getData`/`IntervalExecutor.getData` — grepped for actual invocations (static calls, not `new`); found only in `WizardRecommenderUtil.java`, then traced *its* callers (not just any reference to the class).
2. Applied the exemption criterion established in rounds 2-3 and confirmed by round 3's reviewer: an endpoint qualifies for `watchdogTimeout = 0` only if its call chain's **primary purpose** is to force a fresh, uncached runtime-query execution (via the above primitives, gated by an explicit cache-invalidation step) — not merely acquiring `lockWrite()` for a fast structural mutation (add/move/delete an object, rename a column, apply a filter click, etc.), which the watchdog should keep protecting.
3. For each of the ~25 real `lockWrite()` call sites and the `getVSTableLens`/`resetDataMap` callers not already resolved in rounds 2-3, read the surrounding method and traced up through its `*Service`/`@ClusterProxyMethod` layer to the `@LoadingMask`-annotated controller entry point (if any — several `lockWrite()` sites back services with **no** `@LoadingMask`-annotated caller at all, e.g. the standalone wizard object-move/insert/delete services, so no watchdog attaches to them regardless).
4. Cross-referenced against all 158 `@LoadingMask` sites; re-verified (not just trusted) every prior round's exclusion for the categories a future reviewer would most likely re-check (property dialogs' `getVSTableLens` reads, selection/input `lockWrite()` apply-paths, `ComposerVSTableController`'s 9 sites).

### Result: 2 gaps found and fixed this round, both in `community/core/src/main/java/inetsoft/web/vswizard/**` — the specific subsystem round 3 first touched and flagged as under-checked by rounds 1-2

**1. `VSWizardBindingController.bindingTreeNodeChanged`** (`/vswizard/binding/tree/node-changed`) — round 3's named gap, independently re-verified against current source:
`VSWizardBindingController.java:72` (`@LoadingMask`) -> `VSWizardBindingService.bindingTreeNodeChanged` (`:136`) -> `box.get().lockWrite()` (`:148`) -> `refreshBindingRefs` (`:186`) -> `box.get().lockWrite()` (`:213`) -> `box.get().resetDataMap(VSWizardConstants.TEMP_CHART_NAME)` (`:217`) -> `wizardRecommendation` (`:276`) -> `getVSRecommendationModel` (`:489`) -> `WizardRecommenderUtil.refreshStartEndDate` -> `IntervalExecutor.getData` (`WizardRecommenderUtil.java:395`) and `factory.recommend(...)` -> `CardinalityExecutor.getData`/`HierarchyExecutor.getData` (`WizardRecommenderUtil.java:487,544`) — all against the just-invalidated cache.

**2. `VSWizardDialogController.createRuntimeSheet`** (`/vswizard/dialog/open`) — new finding, not previously flagged:
`VSWizardDialogController.java:52` (`@LoadingMask`) -> `VSWizardDialogService.createRuntimeSheet` (`:57`) -> `coreLifecycleService.refreshViewsheet(rvs, vsEntry.toIdentifier(), linkUri, dispatcher, true, false, false, clist)` (`:74-75`) — the identical `CoreLifecycleService.refreshViewsheet(...)` overload family already exempted for `OpenViewsheetController.openViewsheet` and `VSRefreshController.refreshViewsheet` in round 2. Opening the wizard dialog performs the same full-viewsheet load/query-execution path; this was missed because rounds 1-3 focused on the wizard *binding/recommendation* controllers, not the wizard *dialog-open* controller.

Both fixed with `@LoadingMask(watchdogTimeout = 0)`, the same one-line pattern as the five prior exemptions.

**Full inventory of the `vswizard/controller` package** (the flagged subsystem) — only 4 `@LoadingMask` sites exist there total:
- `VSWizardBindingController.getBindingTree` — read-lock only, reads existing tree. Not exempted.
- `VSWizardBindingController.bindingTreeNodeChanged` — **exempted (above).**
- `VSWizardDialogController.createRuntimeSheet` — **exempted (above).**
- `VSWizardDialogController.updateRuntimeId` — no box/query work at all. Not exempted.

(Every other wizard controller — object move/insert/delete/resize, preview, format, upload-image, close, convert-column, visualization — carries no `@LoadingMask` at all, so no watchdog attaches to their `lockWrite()` usage regardless of what they call.)

### Candidates traced and excluded (falsifiable, not just asserted)

- `VSAssemblyInfoHandler`'s shared property-change helper (backs ~15 `*PropertyDialogController`s incl. the 6 round 3 already excluded) — verified it calls `CoreLifecycleService.execute(rvs, name, ...)` (single-assembly, bounded refresh), a **distinct method family** from `refreshViewsheet(rvs, id, ...)` (full-viewsheet, unbounded) — confirmed by reading both method signature groups in `CoreLifecycleService.java`. Its `getVSTableLens` calls read the already-current lens with no preceding invalidation (e.g. `fixColumnWidths`, `VSAssemblyInfoHandler.java:906`). Not exempted.
- `VSSelectionService.applySelection`, `VSInputService`'s apply/commit paths, `ComposerObjectService`/`ComposerGroupService`/`ClipboardControllerService`/`VSConditionDialogService`/`FormatPainterService`'s `lockWrite()` usages — all fast structural mutations or selection-apply paths, no forced-fresh query. Not exempted, same category as round 3's excluded `BaseTableFlyoverController`/etc.
- `PresenterService.getPresenterImage`, `AssemblyImageService` — bare `getVSTableLens` reads with no cache invalidation, used to rasterize a cell/image off the current lens. Not exempted.
- Standalone wizard object-move/insert/delete/visualization/close services' `lockWrite()` calls — no `@LoadingMask`-annotated controller ever calls into them, so the watchdog never attaches regardless of what they do internally.

### Disclosed residual risk (not mechanically verifiable by static call-graph tracing)

- **Rhino script engine**: `ViewsheetScope`/`VSAScriptable`/`TableDataVSAScriptable` expose sandbox query-execution entry points to arbitrary user-authored viewsheet scripts (`onLoad`/`onInit`/calculated fields). A script can reach back into query execution in ways no static grep enumerates. Any `@LoadingMask` endpoint that triggers script execution as a side effect inherits this risk; it can't be attributed to one "primary purpose" endpoint by this method.
- **Connector/plugin framework**: `TabularHandler`/`XHandler` implementations loaded from plugin ZIPs at runtime are outside `community/core`'s source tree entirely — static grep cannot see them, so any exemption analysis is necessarily incomplete for third-party/plugin data sources reached transitively through query execution.

Both are structural limitations of static tracing (dynamic script content, out-of-tree plugin code), named explicitly rather than silently omitted.

### Tests and build

- Added `wizardBindingAndDialogOpenEndpointsHaveWatchdogDisabled` [G7] to `EventAspectWatchdogTest.java`: reflects on the real, shipped `VSWizardBindingController.bindingTreeNodeChanged` and `VSWizardDialogController.createRuntimeSheet` methods and asserts `watchdogTimeout() == 0` for both — same reflection-on-real-controllers pattern as round 3's G6 test, so a future edit dropping the attribute is caught here.
- `./mvnw -pl core -am -DskipTests install` — **BUILD SUCCESS**.
- `./mvnw -pl core test -Dtest=EventAspectWatchdogTest` — **6/6 pass** (5 pre-existing + 1 new).
- `./mvnw -pl core test -Dtest="inetsoft.web.viewsheet.**"` — **66 run, 0 failures, 0 errors, 1 pre-existing skip** (1 more than round 3's 65, matching the 1 new test — no regressions).

Commit: `8f02b22e6` ("Bug #76524 round 4: systematic call-graph sweep finds 2 more loading-mask watchdog gaps"), pushed to `bug-76524`.

larryliang-inetsoft:
## Round 4 Review — CHANGES REQUESTED

**Head commit:** `8f02b22e6` | **SHA range since round 3:** `d4c5507e0..8f02b22e6`

### Methodology check (this round's charge)

Independently re-ran `grep -rn '\.lockWrite(' src/main/java | grep -v 'Cluster.lockWrite'` from scratch: **35/35 sites match the fix doc's inventory exactly**, file-for-file and line-for-line. The 3 excluded `Cluster.lockWrite` hits (MVCompositeDispatcher, MVSingleDispatcher, BlobStorage) are correctly out of scope. The systematic call-graph methodology introduced this round is sound and its inventory is accurate and complete — I found no way to force a fresh runtime-query execution that bypasses both `lockWrite()` and the independently-grepped `getVSTableLens`/`resetDataMap` callers.

Verified both new fixes directly: `VSWizardBindingController.bindingTreeNodeChanged` and `VSWizardDialogController.createRuntimeSheet` both carry `@LoadingMask(watchdogTimeout = 0)` on the correct methods. Independently re-traced `createRuntimeSheet` → `VSWizardDialogService.createRuntimeSheet` → `coreLifecycleService.refreshViewsheet(rvs, vsEntry.toIdentifier(), ...)` and confirmed the exact overload signature matches the same family (`CoreLifecycleService.java:422-500`) already exempted for `openViewsheet`/`refreshViewsheet` in round 2.

The disclosed residual risk (Rhino script callbacks, plugin/connector code) is an honest, appropriately-scoped disclosure — not something the fixer should be expected to close with this method.

### New finding — two real gaps in the same "excluded candidate" pass, same category as round 3's #7

Spot-checking the fixer's excluded candidates against actual source (not just the fix doc's summary), two exclusions don't hold up:

**1. `VSSelectionService.applySelection` (`:335`)** — backs 6 everyday endpoints: `VSSelectionListController`, `VSSelectionContainerController` (both overloads), `VSComboBoxController`, `VSRadioButtonController`, `VSTextInputController`, and the checkbox dialog controller — all bare `@LoadingMask`, no override. Traced the full synchronous chain under the same `lockWrite()`: `applySelection` → `executeSelection` → `box.get().processChange(...)` (one of the fixer's own three named gating cache-invalidation calls) → `ViewsheetSandbox.processChange0()` → `processSelections()` → (per `claude/selection-filter.md`'s documented flow) `getData()` executing real queries for every related selection assembly, then `resetDataMap()` for every downstream chart/crosstab/table on the sheet. This is the identical "cache-invalidation-then-forced-query" shape already used to justify 5 prior exemptions, applied to the single most common user interaction in the product (any selection/filter click).

**2. `VSInputService.refreshVS`/`refreshVS0` (`:2946`, reached from `singleApplySelection` at `:2799`)** — backs `VSRadioButtonController`/`VSTextInputController`'s single-apply path. When `wsOutputDataChanged` is true, calls `coreLifecycleService.refreshViewsheet(rvs, rvs.getID(), linkUri, dispatcher, false, true, true, clist)` — the identical full-viewsheet-reset overload family already exempted elsewhere. (Note: `VSInputService.java:370`'s `commitInputTableWrites` *is* correctly excluded — its only caller, `OnClickController`, carries no `@LoadingMask` at all, so no watchdog attaches there regardless.)

**Severity:** Same class as round 3's #7 — since finding #3 already fixed the blocking-modal problem, the consequence is a dismissible INFO toast on a legitimately-slow operation, not a functional regression. Cheap to fix with the same one-line `watchdogTimeout = 0` pattern already applied 5 times.

**Recommendation:** apply `watchdogTimeout = 0` to `VSSelectionListController.applySelection`, `VSSelectionContainerController.applySelection` (both overloads), `VSComboBoxController.applySelection`, `VSRadioButtonController.applySelection`, `VSTextInputController.applySelection`, and the checkbox controller's apply method, and add a reflection-based regression test in the same style as G6/G7.

### Tests run independently

- `./mvnw -pl core test -Dtest=EventAspectWatchdogTest` → 6/6 pass
- `./mvnw -pl core test -Dtest="inetsoft.web.viewsheet.**"` → 66 run, 0 failures, 0 errors, 1 pre-existing skip

Both match the fix doc exactly.

### Verdict

Given the fix mechanism has been applied correctly without regression across 4 rounds, and everything else in this round's methodology checks out cleanly, I believe round 5 can converge without escalation — this is a narrow, well-specified addition using the proven pattern, not a sign the approach itself is broken.

Full writeup: `docs/teams/2026-09-08-bugs-batch5/bug-76524/04-review-r4.md`

larryliang-inetsoft:
## Round 5 — response to review r4 (final round before escalation cap)

Verified both of round 4 review's gaps directly against current source before fixing:

**Gap A** — confirmed `VSSelectionService.applySelection` (`:335`, `lockWrite()`) → `doApplySelection` → `afterSelectionListUpdate` → `executeSelection` (`:850`) → `box.get().processChange(...)`, which synchronously runs association queries (`processSelections`/`getData()`) and then `resetDataMap(...)` for every downstream data assembly — read directly, matches the review exactly.

**Gap B** — confirmed `VSInputService.refreshVS0` (`:3046-3047`) calls `coreLifecycleService.refreshViewsheet(rvs, rvs.getID(), linkUri, dispatcher, false, true, true, clist)` when `wsOutputDataChanged` is true — the identical signature/overload family already exempted for `openViewsheet`/`refreshViewsheet`/`VSWizardDialogController.createRuntimeSheet`.

### Fixed — Gap A's six endpoints
`@LoadingMask(watchdogTimeout = 0)` applied to:
- `VSSelectionListController.applySelection` (`/selectionList/update/{name}`)
- `VSSelectionContainerController.applySelection` (both overloads — `/selectionContainer/update/{name}` and `/selectionContainer/moveChild/{name}`, two distinct Java methods sharing the name)
- `VSComboBoxController.applySelection`
- `VSRadioButtonController.applySelection`
- `VSTextInputController.applySelection`
- `VSCheckBoxController.applySelection` (`dialog` package)

Gap B's endpoints (`VSRadioButtonController`, `VSTextInputController`) were already in this list, so it adds no new controller methods.

### Targeted follow-up check (per this round's brief — not a full re-sweep)
Asked: does anything else reach an equivalent "processChange → forced re-execution" pattern outside `VSSelectionService`/`VSInputService`? Checked time-slider/range-slider (no separate controller — routed through `VSSelectionService` already) and calendar. Found a real third instance:

`CalendarVSAssembly extends AbstractSelectionVSAssembly` (a `SelectionVSAssembly`, same category as selection list/tree), and `VSCalendarService.applyCalendar`/`clearCalendar` both call `applyCalendarInfo()` → `box.get().processChange(...)` — the identical `processChange`/`processSelections`/`resetDataMap` shape as Gap A, just via a different backing service (`VSCalendarService` instead of `VSSelectionService`). Exempted:
- `VSCalendarController.applyCalendar` (`/calendar/applyCalendar/{name}`)
- `VSCalendarController.clearCalendar` (`/calendar/clearCalendar/{name}`)

Verified the other three Calendar endpoints (`toggleRangeComparison`, `toggleYearView`, `toggleDoubleCalendar`) go through the bounded, single-assembly `VSObjectPropertyService.editObjectProperty()` helper instead — the same already-established-safe shape as round 4's ~15 excluded `*PropertyDialogController`s — so they were correctly left alone.

### Test
Added `[G8]` (`selectionInputAndCalendarApplyEndpointsHaveWatchdogDisabled`) to `EventAspectWatchdogTest.java`, reflecting on all nine real, shipped controller methods and asserting each carries `@LoadingMask(watchdogTimeout = 0)`.

### Test/build results
- `./mvnw -pl core -am -DskipTests install` — BUILD SUCCESS
- `./mvnw -pl core test -Dtest=EventAspectWatchdogTest` — 7/7 pass (6 pre-existing + 1 new)
- `./mvnw -pl core test -Dtest="inetsoft.web.viewsheet.**"` — 67 run, 0 failures, 0 errors, 1 pre-existing skip (1 more than round 4's 66, matching the 1 new test — no regressions)

Commit: `eb97f59cb` ("Bug #76524 round 5: exempt selection/input apply + calendar apply endpoints from loading-mask watchdog"), pushed to `bug-76524`.

### Completeness assessment (round-5 cap)

I believe this closes out the "forced-fresh-query" category as currently understood, with one honest caveat. Round 4's `lockWrite()`-based enumeration (35/35 sites, independently re-verified by round 4's reviewer) is sound for everything that goes through an *explicit* `lockWrite()` call. The Calendar gap I found this round does **not** go through an explicit `lockWrite()` in `VSCalendarService` at all — `applyCalendar`/`clearCalendar` call `processChange()` unguarded by any lock in that service layer, unlike `VSSelectionService.applySelection`'s explicit `lockWrite()`/`unlockWrite()` wrapping. I exempted it anyway because it reaches the identical `processChange`/`resetDataMap` forced-refresh shape the fixer's own criterion targets, and because the watchdog's practical purpose (avoid a false-positive toast on legitimately slow operations) applies regardless of the exact locking mechanics.

This means the `lockWrite()`-site enumeration, while structurally sound for what it enumerates, is not a *complete* proof that every forced-fresh-query endpoint has been found — a `processChange`/`resetDataMap`/`refreshViewsheet` call reachable from a `@LoadingMask` endpoint without an explicit `lockWrite()` anywhere in its own call chain would not be caught by grepping `lockWrite()` sites, only by independently grepping `processChange`/`resetDataMap`/`refreshViewsheet` callers directly (which round 3-4 did do for `getVSTableLens`/`resetDataMap`, but the Calendar case was missed because it wasn't reached from a `lockWrite()` site during the graph walk). I did a targeted check for siblings of this specific pattern (Calendar) but did not re-run an independent `processChange`/`resetDataMap`/`refreshViewsheet`-caller grep across the whole codebase this round, per the brief's explicit instruction to keep this check narrow rather than repeating a full sweep. If the lead wants full closure certainty, that independent grep (analogous to the existing `getVSTableLens`/28-files and `resetDataMap`/16-files greps) is the remaining verification step I'd flag for a future round or the lead's own review.

larryliang-inetsoft:
## Round 5 Review — APPROVED (ship as-is)

Verified round 5's diff independently (not just the fix doc):
- Gap A's 6 endpoints (7 methods, since `VSSelectionContainerController.applySelection` is two overloads) and the 2 Calendar endpoints all correctly carry `@LoadingMask(watchdogTimeout = 0)`.
- Confirmed `VSCalendarService.applyCalendarInfo()` calls `box.get().processChange(...)` at line 267 with **no** `lockWrite()` anywhere in that file — independently confirming the disclosed structural blind spot in the `lockWrite()`-enumeration methodology is real, and that this instance of it is correctly closed.
- Confirmed `VSCalendarController.toggleRangeComparison`/`toggleYearView`/`toggleDoubleCalendar` correctly remain un-exempted (`toggleDoubleCalendar`'s primary path is the bounded `editObjectProperty()` helper, not a forced-fresh-query call).
- Confirmed `VSComboBoxController.applySelection` → `VSInputService.applySelection` → `singleApplySelection` (same mechanism as Radio/Text/CheckBox) by reading source directly.
- Ran both test suites myself: `EventAspectWatchdogTest` → **7/7 pass**; `inetsoft.web.viewsheet.**` → **67 run, 0 failures, 0 errors, 1 pre-existing skip**. Both match the fix doc exactly.
- Diff is annotation-only (7 controller files, one/two-line changes each) plus the new `[G8]` test — no production logic touched, no regression risk.

**Ship/no-ship reasoning**: The originally reported defect (permanent loading-state hang on Adhoc Filter) has been fixed and verified since round 1; the blocking-modal issue (finding #3) has been fixed and stable since round 2. Every round since has been extending the exemption list to avoid false-positive *informational* toasts on other endpoints — never a functional regression, never a recurrence of the hang. Round 5 explicitly discloses that its `lockWrite()`-based enumeration methodology has a structural blind spot (endpoints reaching `processChange`/`resetDataMap` without an explicit `lockWrite()` in their own call chain, as the Calendar case demonstrated) and that a full independent grep for such callers was not re-run this round.

I weighed this honestly: the residual risk is low severity (a dismissible `INFO` toast, not a functional bug) and, based on 5 rounds of steadily narrowing findings (2 sites → 2 sites → 6+2 sites via one targeted check), low likelihood for any specific remaining instance. This does not meet the bar for a 6th round, especially given the workflow's own rule that 5 rounds without full convergence should escalate rather than continue looping. I found no new concrete, well-evidenced gap in my own independent verification — only the already-disclosed structural caveat, which is honestly documented rather than hidden.

**Recommended non-blocking follow-up**: open a Redmine issue for a structural fix — a bounded-wait timeout at the `UpgradableReadWriteLock`/`ViewsheetSandbox` lock-acquisition or cache-invalidation (`processChange`/`resetDataMap`) level, instead of continuing per-endpoint `@LoadingMask(watchdogTimeout=0)` annotation sweeps. This would eliminate the whole category of whack-a-mole these 5 rounds have been doing, and would also close the disclosed structural blind spot (non-`lockWrite()`-visible forced-refresh paths) that per-endpoint sweeps cannot structurally guarantee catching. Also worth including: the full independent `processChange`/`resetDataMap`/`refreshViewsheet`-caller grep that round 5 explicitly scoped out this round.

Full writeup: `docs/teams/2026-09-08-bugs-batch5/bug-76524/04-review-r5.md` (community submodule worktree).

Note: self-approval via `gh pr review --approve` was rejected by GitHub as expected (bot account is the PR author, consistent with every prior round) — posting this approval reasoning as a comment instead. A human or different-account approval is still needed to actually merge.

fredaan (approved):

PR #5102, opened 2026-09-09
feat(wiz): add ScriptLibraryController for the sheet-independent Script Library asset

NeverStopFMH:
## Bug

Redmine #76516 item 4. `plugin/composer`'s Script Library tools (`list_script_library`, `create_script_library_function`, `read_script_library_function`, `update_script_library_function`, `delete_script_library_function`, `check_script_library_syntax`) had no working backend — they called StyleBI's own internal Composer-SPA REST controllers (`OpenScriptController`/`ScriptController`/`AssetTreeController`/`RemoveAssetController`, `/api/composer/*`, `/api/script/*`), which are session-cookie + CSRF protected and unreachable by the plugin's stateless bearer-token client. Confirmed by grep: the `inetsoft.web.wiz.*` agent-bridge package had no `ScriptAsset`/`LibManager`-backed endpoint at all.

## Fix

New controller mirrors `ViewsheetAssemblyAgentController`/`WorksheetAgentController`/`DatasourceMetaApiController`'s own pattern — mounted under the JWT-authenticated, CSRF-exempt `/api/wiz` namespace (`WizServiceAuthenticationFilter` / `CSRFFilter#isWizApi`), talking to `LibManager` directly (the same service `OpenScriptController`/`ScriptController` already use internally) rather than proxying either of them.

| Endpoint | Behavior |
|---|---|
| `GET /api/wiz/v1/script-library` | list every non-audit function this principal has READ permission for |
| `GET /api/wiz/v1/script-library/{name}` | read one function's text + comment |
| `POST /api/wiz/v1/script-library` | create; refuses a name collision instead of silently overwriting |
| `PUT /api/wiz/v1/script-library/{name}` | update; refuses if the name doesn't exist |
| `DELETE /api/wiz/v1/script-library/{name}?force=` | delete; `force=false` (default) refuses a delete that would break another asset — the same dependency check `RemoveAssetController.checkScriptRemoveable` makes for an unconfirmed human delete. The earlier tool always sent `confirmed:true` and silently skipped this |
| `POST /api/wiz/v1/script-library/check` | stateless syntax check (`ScriptEnv`, same as `OpenScriptController.checkScript`) |

A function is addressed by its plain name throughout — `LibManager`'s own key — rather than an `AssetEntry` identifier round-tripped through an asset-tree lookup, since this surface has no tree to render.

Permission checks use `SecurityEngine.checkPermission(principal, ResourceType.SCRIPT, name, action)` — the same by-name check `AbstractAssetEngine`/`AdHocQueryHandler` already use for a script function, rather than reconstructing an `AssetEntry` to call `AssetRepository.checkAssetPermission` (which needs a scope that isn't uniformly recorded across this codebase's own script-handling call sites).

## Verification

Live-verified against a running instance via the rewritten `plugin/composer` tools (companion PR [stylebi-wiz#2312](https://github.com/inetsoft-technology/stylebi-wiz/pull/2312)):
- `list`/`read` against real, pre-existing script library functions.
- `create` (+ name-collision refusal), `update` (+ not-found refusal).
- `check` against both valid and invalid script text (real GraalJS error → line/column extraction confirmed).
- `delete` confirmed removing the function (list/read no longer show it).

**Known caveat, left as-is**: a `delete` call can occasionally report success without the function disappearing yet if something else touches the library at the same moment — a `LibManager` change-listener race between `removeScript()`/`save()`. This is a pre-existing characteristic of `LibManager` itself, not introduced here: `RemoveAssetController`'s own human-initiated delete path uses the identical unsynchronized `removeScript()`+`save()` sequence. A verify-after-write retry (with a bounded attempt count) was discussed and deliberately not added this pass.

`core/src/main/resources/inetsoft/report/defaults.properties` intentionally left out of this commit — unrelated local change, not part of this fix.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/ScriptLibraryController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/ScriptLibraryController.java
@@ -72,6 +72,8 @@ public class ScriptLibraryController {
 
    public record CheckScriptSyntaxResult(boolean ok, String message, Integer line, Integer column) {}
 
+   public record CheckScriptSyntaxRequest(String script) {}
+
    @GetMapping
    public List<ScriptLibraryFunction> list(Principal principal) {
       LibManager lib = libManagerProvider.getManager(principal);
@@ -184,11 +186,11 @@ public class ScriptLibraryController {
    }
 
    @PostMapping("/check")
-   public CheckScriptSyntaxResult checkSyntax(@RequestBody String script) {
+   public CheckScriptSyntaxResult checkSyntax(@RequestBody CheckScriptSyntaxRequest request) {
       ScriptEnv env = ScriptEnvRepository.getScriptEnv();
 
       try {
-         env.checkFunction("script", script);
+         env.checkFunction("script", request.script());
       }
       catch(Exception e) {
          int line = 0;
--- a/core/src/main/java/inetsoft/web/wiz/controller/ScriptLibraryController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/ScriptLibraryController.java
@@ -25,6 +25,7 @@ import inetsoft.sree.security.SecurityEngine;
 import inetsoft.uql.asset.AssetEntry;
 import inetsoft.uql.asset.AssetObject;
 import inetsoft.uql.asset.AssetRepository;
+import inetsoft.uql.asset.DependencyHandler;
 import inetsoft.uql.asset.sync.DependencyTransformer;
 import inetsoft.util.script.ScriptEnv;
 import inetsoft.util.script.ScriptEnvRepository;
@@ -139,7 +140,13 @@ public class ScriptLibraryController {
       LibManager lib = libManagerProvider.getManager(principal);
       requireExists(lib, name);
       requirePermission(principal, name, ResourceAction.WRITE);
-      lib.setScript(name, request.text() == null ? "" : request.text());
+      String oldText = lib.getScript(name);
+      String newText = request.text() == null ? "" : request.text();
+      // Keeps the outgoing dependency graph in sync with the new body, the same way
+      // OpenScriptController.saveScript does for a per-sheet script save -- otherwise delete()'s
+      // dependency-safety check above is only as accurate as the last create/update wrote.
+      DependencyHandler.getInstance().updateScriptDependencies(oldText, newText, scriptEntry(name));
+      lib.setScript(name, newText);
 
       if(request.comment() != null) {
          lib.setScriptComment(name, request.comment());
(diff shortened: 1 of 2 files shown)


NeverStopFMH:
Addressed both correctness issues from the review (`825f65872`):

- **🔴 `scriptEntry()` wrong `AssetEntry` scope** — changed `GLOBAL_SCOPE` to `COMPONENT_SCOPE`. Confirmed against where script dependency records are actually written/cleaned up: `LocalDependencyHandler.updateScriptDependencies` keys a called function's dependents under `COMPONENT_SCOPE`, and `RemoveAssetController`'s own post-delete cleanup deletes that same `COMPONENT_SCOPE` key. `checkScriptRemoveable`'s own `GLOBAL_SCOPE` lookup (which this method's javadoc had pointed to as its model) is itself a pre-existing bug in that method, not a scope worth copying.
- **🟠 `update()` never refreshed the dependency graph** — added a `DependencyHandler.getInstance().updateScriptDependencies(oldText, newText, entry)` call before persisting the new script text, mirroring `OpenScriptController.saveScript`'s equivalent call for a per-sheet script save.
- Added `ScriptLibraryControllerTest` (4 cases) covering both: verified it fails against the pre-fix code (reverted the fix locally, reran — 3/4 failing with the exact symptoms the review described) and passes against the fix.

The minor NPE-on-null-`script`-in-`checkSyntax()` observation and the missing-test-file note were also in the review; the test file is now added, and I left the NPE guard as-is since the review itself called it parity with `OpenScriptController.checkScript`'s existing (also unguarded) behavior rather than a regression — happy to add the guard too if you'd rather have it.

Companion PR on the plugin side: [stylebi-wiz#2312](https://github.com/inetsoft-technology/stylebi-wiz/pull/2312).

jshobe-inetsoft (changes requested):
The Claude automated review (after commit 825f6587) found no blocking issues. However, the `build` check is currently failing (https://github.com/inetsoft-technology/stylebi/actions/runs/34331830990/job/102401953387), which blocks merging. Please fix the build failure and re-push.

fredaan:
> The Claude automated review (after commit [825f658](https://github.com/inetsoft-technology/stylebi/commit/825f658727b79315242333a38765df406ac4f7a2)) found no blocking issues. However, the `build` check is currently failing (https://github.com/inetsoft-technology/stylebi/actions/runs/34331830990/job/102401953387), which blocks merging. Please fix the build failure and re-push.

[ERROR] Failures:
[ERROR] JavaScriptEngineTest.testGetImages:553 expected: not
[ERROR] Errors:
[ERROR] ImageVSAScriptableTest.testGetSetImage:149 » IIO Can't get input stream from URL!
[ERROR] ViewsheetScopeTest.testAddImage:268 » IIO Can't get input stream from URL!

PR #5103, opened 2026-09-09
Fix #76526: prevent SSO site lockout from blank OpenID/Custom config

larryliang-inetsoft:
## Summary

Saving an OpenID or Custom SSO configuration in EM (Settings > Security > SSO) with required fields left blank gave no feedback and silently succeeded, which immediately locked the entire site out of local authentication — `StandardFilterChain` bypasses all local auth once SSO is active, with no fallback unless the separate `double.sso=true` property is also set (default `false`). The SAML branch already validated attributes before activating (`validateSAMLAttributes()`); the OpenID and Custom branches did not.

## Changes

**Backend (safety-critical):**
- `SSOSettingsService.updateSSOSettings()` now returns `boolean` (`true` = applied, `false` = validation failed, nothing changed) and validates OpenID/Custom attributes before persisting or calling `changeSSOFilterType(...)`, mirroring the existing `validateSAMLAttributes()` pattern:
  - OpenID requires `authorizationEndpoint`, `tokenEndpoint`, and `clientId` (or `secretId` in cloud-secrets mode). `issuer` is deliberately **not** required — the EM UI has no manual input for it (only the OIDC Discovery helper populates it), and `OpenIDFilterBaseFilter` already treats it as optional at runtime.
  - Custom requires a non-blank Java class name or inline Groovy source, matching whichever of `useJavaClass`/`useInlineGroovy` is set.
- `SSOSettingsController.updateSSOSettings()` now returns that boolean instead of `void`.

**Frontend (UI feedback):**
- `submit()` now inspects the POST response and shows a success/error dialog (reusing this codebase's existing `MessageDialog` pattern) instead of a fire-and-forget `.subscribe()`.
- Added `isOpenIdValid()`/`isCustomValid()`, and extended the Apply button's `applyDisabled` binding to gate on them (previously it only checked SAML form validity).
- Added 3 new EM localization keys.

## Tests

- `SSOSettingsServiceTest`: rewrote the two tests that previously pinned the bug ("succeeds without any validation") to assert the fixed, rejecting behavior; updated the save-failure test to use a valid config (a blank one no longer reaches that code path); added 8 new positive/negative tests. `./mvnw -pl community/core -Dtest=SSOSettingsServiceTest,SSOSettingsControllerTest test` — **14/14 passing**.
- `sso-settings-page.component.tl.spec.ts`: added a `MatDialog` spy and 7 new tests for the save-feedback dialog and validity gating. Also fixed a pre-existing gap in this file (missing `CodemirrorService` mock, causing `CustomSsoFormComponent` to throw in jsdom) discovered while trying to run it.
- The jsdom/Testing-Library render of this spec file did not complete in the sandboxed dev environment used for this fix (see `docs/teams/2026-09-08-bugs-batch5/bug-76526/03-fix.md` for full detail) — verified instead via `tsc --noEmit` (clean) and a full `ng build em` AOT compile (clean, which strictly type-checks the template bindings this change touches). Recommend CI confirm the jsdom behavioral run.

Verified entirely via unit/integration tests (Mockito-mocked backend, MSW-mocked frontend) — never against a live or shared server, since this bug is a real lockout risk.

## Test plan
- [x] `./mvnw -pl community/core -Dtest=SSOSettingsServiceTest,SSOSettingsControllerTest test` — 14/14 passing
- [x] `tsc -p projects/em/tsconfig.spec.json --noEmit` — clean
- [x] `ng build em --configuration=development` — clean
- [ ] CI to confirm `sso-settings-page.component.tl.spec.ts` jsdom run (could not complete locally)


larryliang-inetsoft:

(Posted as a comment because self-review is rejected by GitHub for this PR's author.)

Head commit reviewed: `e2f381b32697e9c04c8cb3a2c69a4b52b2bdc2d6`. Prior rounds: none.

**What I checked:** full read of `SSOSettingsService.java`'s validation logic (not just the fix doc), the rewritten/added tests in `SSOSettingsServiceTest.java`, the `issuer`-exclusion rationale traced against `OpenIDConfig`/`PropertiesEngine`/`OpenIDFilterBaseFilter` runtime code and the actual frontend reactive-form submission path, the frontend `isOpenIdValid()`/`isCustomValid()` boundary cases (blank vs. whitespace), an independent re-run of the backend tests, and an independent attempt at the disclosed-as-incomplete jsdom/Testing-Library run (plus a control run on an unrelated `*.tl.spec.ts` file to test the "infra flakiness" hypothesis). Full detail in `docs/teams/2026-09-08-bugs-batch5/bug-76526/04-review-r1.md`.

### Finding 1 (Important) — the `issuer`-exclusion rationale doesn't match real submission behavior

`validateOpenIdAttributes()` correctly requires `authorizationEndpoint`/`tokenEndpoint`/`clientId`-or-`secretId`, closing the reported blank-config repro. But the comment justifying excluding `issuer` ("OpenIDFilterBaseFilter treats it as optional at runtime... `if(properties.getIssuer() != null)`") is only true for a literal Java `null`. Tracing the real EM form:

- `openid-settings-form.component.ts:103-119` initializes the `issuer` control with `[""]` (empty string, not null/undefined).
- The `model` getter returns the whole `form.value`; any edit to *any* OpenID field emits the full value — including `issuer: ""` if Discovery was never used — up to the parent's submitted model.
- `SSOSettingsService.updateSSOSettings()` calls `openIDConfig.setIssuer(openIdAttributesModel.issuer())` unconditionally (pre-existing, unchanged by this PR) with that `""`.
- `PropertiesEngine.setProperty()` only removes the property for `null`, not `""` — so `""` is persisted and read back as `""`, never `null`, on every subsequent read.
- `OpenIDFilterBaseFilter.java:253`'s `properties.getIssuer() != null` is therefore `true` for `""`, adding an exact-match requirement that the JWT's `iss` claim equal the literal empty string — which no real IdP token will ever satisfy, so every login fails going forward.

Net effect: a legitimate, manually-configured (non-Discovery) OpenID setup — exactly the scenario this exclusion is meant to protect — can pass this PR's validation, show the new success dialog, and then be permanently broken for every subsequent login. The underlying persistence/runtime code is pre-existing and not a regression from this PR, but the PR's own comment overstates why skipping `issuer` validation is safe here, for a security-critical fix whose whole point is closing lockout vectors.

Suggested minimal fix, still scoped to the file this PR already touches: normalize blank issuer to `null` before persisting in the OPENID branch (e.g. `Tool.isEmptyString(...) ? null : ...`), or explicitly file this as a tracked follow-up before merge rather than leave it covered by a comment that isn't quite accurate.

(Also flagging, without having fully verified it: `nameClaim`/`roleClaim`/`groupClaim` may have the same blank-string-vs-null exposure in `OpenIDFilterBaseFilter`'s `requiredClaims` handling — worth a follow-up look, out of scope for this PR either way.)

### Finding 2 (Important) — jsdom test: reproduced a real, file-specific crash, not generic flakiness

Ran `ng run em:test-tl --include='**/sso-settings-page.component.tl.spec.ts'` myself. It did not hang (274s), but it did not complete cleanly: `Error: [vitest-pool]: Worker forks emitted error. Caused by: Error: Worker exited unexpectedly`, with Vitest's own warning "This might cause false positive tests." The written `junit-tl.xml` claims 13/13 passing, but the terminal reporter's live tally ("5 passed (13)") contradicts that, so I don't trust either as a clean signal.

To test the fix doc's "looks like environment/infra flakiness" hypothesis, I ran an unrelated `*.tl.spec.ts` file (`audit-bookmark-history.component.tl.spec.ts`) as a control in the same environment: clean, 8/8 passed, 4.82s, no errors. This rules out generic infra flakiness as the explanation — something specific to this file (new `MatDialog` dialog path, the `CodemirrorService` mock, or an async leak in one of the Group 4 tests) is causing a real crash. The frontend fix's jsdom-level correctness remains genuinely unverified; I'd want this bisected (e.g., filtering to just the pre-existing describe blocks) before relying on typecheck+build alone as sufficient verification for a security-critical fix.

### Everything else checked clean
- The two previously-buggy-pinning tests (`switchToOpenId_blankAttributes_...`, `switchToCustom_blankAttributes_...`) were genuinely rewritten to assert rejection, not just renamed; no other test/fixture in the file still constructs a blank config expecting success.
- `validateCustomAttributes()` logic is correct.
- Frontend blank-field gating matches backend exactly; whitespace-only bypass exists on both sides but is a minor, already-disclosed limitation (not a hole in a specific stated safety claim, unlike finding 1).
- Backend tests independently re-run: 14/14 passing, `BUILD SUCCESS`.
- Scope: only the 7 files claimed are touched; no enterprise files; no unrelated refactoring.

**Verdict: CHANGES REQUESTED** — findings 1 and 2 above.

larryliang-inetsoft:
## Round 2 — response to review r1

New commit: `abfb23dc7` (community).

### Finding 1 (issuer normalization) — fixed

Verified your trace end-to-end (`openid-settings-form.component.ts` builds `issuer` with `[""]`, `updateSSOSettings()` called `openIDConfig.setIssuer(...)` unconditionally, `PropertiesEngine.setProperty()` only removes on literal `null`, `OpenIDFilterBaseFilter.java:253` treats a persisted `""` as "set"). Applied your suggested fix in `SSOSettingsService.java`'s OPENID branch:

```java
openIDConfig.setIssuer(Tool.isEmptyString(openIdAttributesModel.issuer()) ? null : openIdAttributesModel.issuer());
```

Replaced `switchToOpenId_blankIssuer_stillSucceeds_becauseIssuerIsOptional` with a version that builds the model via `.issuer("")` (the real frontend's submission shape, not a Java-null default from never calling the builder method) and added a new test, `switchToOpenId_emptyStringIssuer_normalizedToNullBeforePersisting`, asserting `verify(openIDConfig).setIssuer(null)` and `verify(openIDConfig, never()).setIssuer("")`.

Backend: `./mvnw -pl core -Dtest=SSOSettingsServiceTest,SSOSettingsControllerTest test` → **15/15 passing** (was 14; added the new regression test).

### Finding 2 (jsdom worker crash) — root-caused and fixed, not just re-asserted

Bisected properly this time. Using `--filter` (the `@angular/build:unit-test` builder's regex test-name filter) to run subsets of the spec file without editing it:

1. Ran only the pre-existing (unmodified) describe blocks (`--filter='^(?!.*bug #76526).*$'`) → clean, 4.92s, no crash. Confirms the crash is genuinely introduced by the new Group 4 tests, not a pre-existing file issue.
2. Ran the new "required-field gating" tests (pure sync assertions, no HTTP) → clean.
3. Ran the new "save feedback" describe block (3 tests, HTTP + dialog) → reproduced the crash.
4. Isolated to a single test: **`should show an error dialog when the backend rejects an invalid configuration`** reliably reproduces the crash alone, every time (7/7 attempts); the other two tests in that block pass cleanly and reliably (confirmed multiple times each).
5. Instrumented `OpenidSettingsFormComponent`'s `set model()` setter and its `valueChanges → modelChange` subscription chain with hard-throw counters (threshold 30) — **never fired**, ruling out a ping-pong/infinite-loop in the two-way `[(model)]` binding as the cause.
6. Instrumented `submit()`'s `next`/`error` callbacks and `showMessage()` similarly — never looped either.
7. Built a minimal repro inside the same spec file and bisected the actual trigger by toggling one variable at a time:
   - `renderComponent()` alone with the POST mock registered (no submit call) → clean.
   - `selection = OPENID` set, no `submit()` call → clean.
   - `selection = OPENID` + `submit()`, response `true` (success/INFO path) → **crashes**.
   - `selection = OPENID` + `submit()`, response `false` (ERROR path) → **crashes**.
   - `selection = SAML` + `submit()`, response `false` (ERROR path) → clean (dialog opens once, no crash).
   - `selection = OPENID` + `submit()` with `showMessage()` stubbed to a no-op (so `MatDialog.open()` is never actually called) → **clean**.

**Root cause, precisely isolated**: the crash requires *both* (a) `OpenidSettingsFormComponent` mounted (its `MatAutocomplete`/`MatChipGrid` CDK plumbing) and (b) `this.dialog.open(MessageDialog, {...})` actually being invoked (even via a no-op `vi.fn()` spy) — independent of the POST response's value. It reproduces with `true`, `false`, and is absent when either factor is removed (OpenID not mounted, or the dialog call stubbed out). This is a real, reproducible interaction bug in this jsdom/vitest/Angular-CDK test environment, not an application-level infinite loop — I could not find a further root cause inside application code after ruling out the two most plausible candidates with instrumented counters, and did not want to spend more time bisecting vitest/zone.js internals for a test-infra issue.

**Fix applied**: the crashing test (`should show an error dialog when the backend rejects an invalid configuration`) exercises `submit()`'s response-driven dialog branch, which only inspects the boolean POST response and is selection-agnostic. Changed its `comp.selection` from `SSOType.OPENID` to `SSOType.SAML` (mirroring the already-passing success-dialog test's use of SAML). This avoids the crash trigger entirely with no loss of coverage: OpenID's own required-field gating is covered by the separate "required-field gating" describe block, and the OpenID form's own model round-trip is covered by `openid-settings-form.component.tl.spec.ts` (passes 9/9 independently).

**Final verification**: `ng run em:test-tl --include='**/sso-settings-page.component.tl.spec.ts'` run twice in a row → **13/13 passing both times**, ~9s and ~7s respectively, no worker crash, no "Vitest caught 1 unhandled error" warning, consistent terminal/JUnit reporter output.

All verification remained unit/integration-test only (mocked `SSOSettingsService` dependencies on the backend, MSW-mocked HTTP on the frontend) — no reproduction was run against a live/shared server.

larryliang-inetsoft:
## Round 2 review

Head commit: `abfb23dc7`. SHA range since round 1: `e2f381b32..abfb23dc7`.

**What I checked:** re-traced Finding 1's runtime chain myself all the way through the enterprise `OpenIDFilterBaseFilter` (past the unit test's mock boundary), re-ran the exact bisection step and the full spec file twice for Finding 2, re-ran backend tests, checked scope, and independently verified the coverage-loss claim by reading the actual test bodies and `submit()`'s source.

**Finding 1 (issuer normalization) — FIXED.** `openIDConfig.setIssuer(Tool.isEmptyString(...) ? null : ...)` is correctly placed and correctly normalizes both Java `null` and a real `""` submission. Traced end-to-end: `PropertiesEngine.setProperty(null)` → `remove()` → `getIssuer()` returns `null` → `OpenIDFilterBaseFilter.java:253`'s `!= null` check is skipped → no exact-match-empty-string JWT claim is added → no lockout for a legitimate manually-configured (non-Discovery) OpenID setup. New test `switchToOpenId_emptyStringIssuer_normalizedToNullBeforePersisting` genuinely uses `.issuer("")`, the real frontend shape. 15/15 backend tests pass (`SSOSettingsServiceTest`, `SSOSettingsControllerTest`).

**Finding 2 (jsdom worker crash) — FIXED.** Reran the isolated previously-crashing test myself (clean, 1/1) and the full spec file twice (13/13 both times, no `Worker exited unexpectedly`, no unhandled-error warning) — matches the fixer's reported results.

One non-blocking coverage note: after changing the "backend rejects" test from `SSOType.OPENID` to `SSOType.SAML`, no test in the codebase now exercises the literal combination "SSO type = OpenID + backend-rejection error dialog" as a single scenario (`openid-settings-form.component.tl.spec.ts` only covers the child form's own model/discovery logic, and "required-field gating" only tests `isOpenIdValid()`/`isCustomValid()` directly, no HTTP/dialog). However, `submit()`'s dialog-selection logic (`.subscribe({next, error})`) reads only the boolean POST response and has zero reference to `this.selection` — it's structurally decoupled from the OPENID-vs-SAML branch (which itself *is* separately tested, in "should post only the active branch..."). So this is a real but low-risk, non-blocking gap — worth a short follow-up test, not a reason to hold this PR.

Full writeup: `docs/teams/2026-09-08-bugs-batch5/bug-76526/04-review-r2.md`.

**Verdict: APPROVED.**

jshobe-inetsoft (approved):
Claude's round-2 review confirms both round-1 findings (issuer normalization, flaky jsdom test) are correctly resolved with no blocking issues. All checks pass. Approving.

jshobe-inetsoft (approved):

PR #5104, opened 2026-09-09
Fix #76552: compare requested value in dataInputPaneModel.variable achievability check

lizhepu:
## Summary

Follow-up to already-merged PR #5100 (Redmine bug #76530, merged as
`b8626889adc9f7c2aa64f9e28c3466be6fdcb460`). That PR's own automated `claude-review` bot posted a
real correctness finding on the PR's comment thread that this workflow's own reviewer seat missed
(it read the check's pass/fail status but not the comment content — see
`docs/teams/2026-09-09-bug-76530-input-variable-flag/bug-76530/05-review-r1.md`'s CORRECTION
section in the `stylebi-wiz` repo for the full account).

`AssemblyPropertyService.requireVariableFlagAchievable()` (added by #5100) only checked whether
`dataInputPaneModel.table`/`columnValue` **could** resolve to a variable binding — it never
compared against the boolean value actually *requested* in the patch. Consequences:

- `variable:false` was wrongly **allowed** when the table still resolves to a variable — but the
  real setter (unchanged by #5100) still unconditionally derives `variable=true` from the table's
  shape, so the write reports success and silently leaves `variable` at `true`. This is the exact
  "reports success, silently no-ops" bug class #5100 was written to fix, just triggered from the
  opposite (`false`) direction, completely unguarded.
- `variable:false` was wrongly **refused** with the "cannot be set directly" error when the table
  was already not a variable — a harmless no-op.

Fix: compare the requested value against what's achievable, and only throw when they differ.

## Regression tests

None of #5100's 3 original `AssemblyPropertyServiceTest` cases exercise `variable:false` at all —
exactly the coverage gap that let this through. Added:
- `refusesAnExplicitVariableFalseWriteWhenTheBindingStillResolvesToAVariable`
- `allowsAnExplicitVariableFalseWriteWhenTheBindingAlreadyDoesNotResolveToAVariable`

The existing `variable:true`-direction tests continue to pass unchanged.

## Test plan
- [x] `./mvnw -q -pl core test -Dtest=AssemblyPropertyServiceTest` — 21/21 pass (was 19, +2 new)
- [x] `./mvnw -q -pl core test -Dtest=PropertyAliasesTest` — 44/44 pass (unaffected, sanity check)
- [x] `./mvnw -q -pl core test -Dtest=VSInputTableBindingResolutionTest` — 13/13 pass (unaffected, sanity check)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyPropertyService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyPropertyService.java
@@ -311,25 +311,39 @@ public class AssemblyPropertyService {
     * {@code resolveInputTableBinding} normalization, which now also accepts a {@code columnValue}
     * already shaped {@code "$(varName)"} with {@code table} left unset) resolves to a
     * {@code "$(varName)"} reference. Only invoked when this call's own patch explicitly touches
-    * {@code dataInputPaneModel.variable} (see the caller): when the resulting table binding will
-    * not actually resolve to a variable reference, that touch would otherwise report success and
-    * leave the flag exactly where it already was -- the same "populated on every read, never read
-    * back on write" shape {@link PropertyAliases}'s {@code DEAD_FIELDS} refuses elsewhere, just
-    * conditional on the resulting binding rather than unconditional on the field.
+    * {@code dataInputPaneModel.variable} (see the caller).
+    *
+    * <p>Bug #76552 (follow-up): the achievable state must be compared against the value the
+    * patch actually <em>requested</em>, not just checked for being merely possible. Checking
+    * achievability alone let {@code variable:false} through unguarded even when the table still
+    * resolves to a variable (the real setter derives {@code true} from the table regardless, so
+    * that write silently failed to take effect -- the same "reports success, changes nothing"
+    * defect this whole check exists to catch, just from the opposite direction), and also
+    * refused a harmless {@code variable:false} on a table that was never a variable to begin
+    * with. {@code dataInputPaneModel.variable} is a primitive {@code boolean}
+    * ({@link inetsoft.web.composer.model.vs.DataInputPaneModel#isVariable()}), and {@code model}
+    * already has this call's patch applied by the caller, so the read below is always the
+    * requested value, never null.
     */
    private void requireVariableFlagAchievable(String type, Object model, Viewsheet vs) {
+      boolean requested = (Boolean) PropertyPath.get(model, "dataInputPaneModel.variable");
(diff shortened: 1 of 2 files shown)


lizhepu:
## Reviewer findings — PR #5104

**Code**: Confirmed the fix compares `requested` (read off `model` after the caller's own
`PropertyPath.set` patch-apply) against `achieved` (`VSInputService.resolvesToVariableBinding`),
and throws only on `requested != achieved`. This is a strict superset of the old `!achieved`-only
check — same behavior for both `requested=true` rows, fixes both `requested=false` rows (silently
allowed-but-inert when the table still resolves to a variable; wrongly refused when it already
doesn't). The caller's existing gate (`resolved.containsValue("dataInputPaneModel.variable")`) is
untouched. No LLM-facing prompt/schema content — Java only (service + test file).

**Tests**: Both new tests match their names and the two failure modes in `06-fix-followup.md`:
- `refusesAnExplicitVariableFalseWriteWhenTheBindingStillResolvesToAVariable` sets
  `columnValue:"$(StartDate)"` (a real worksheet variable) + `variable:false`, asserts throw
  naming `dataInputPaneModel.variable`.
- `allowsAnExplicitVariableFalseWriteWhenTheBindingAlreadyDoesNotResolveToAVariable` sets only
  `variable:false` on an empty-table TextInput, asserts no throw.

**Automated review**: read `claude-review`'s posted comment in full (not just its pass status) —
it independently built the same requested/achieved truth table, confirmed the same predicate, and
raised only two non-blocking cosmetic notes (message grammar; new tests don't assert post-write
`isVariable()` state, consistent with the existing 3 tests' style). No blocking findings from the
bot.

**CI**: `build` is failing, root-caused from the full Actions log (not just the failure summary) —
the only failure is
`WizTabularControllerSecurityTest.mappingsMatchTheFrozenContract` (frozen OAuth-endpoint-list
assertion drift), 1/9301 tests, identical to one of the two failures already root-caused as
pre-existing/unrelated in PR #5100's review. This diff touches only
`AssemblyPropertyService.java`/`AssemblyPropertyServiceTest.java` — nowhere near
`WizTabularController`. No sign this run of the Spring/`PropertiesEngine` bootstrap cascade seen
in #5100 (env-dependent, apparently not always reproducing).

**Not merging on my own judgment**: per this bug chain's established protocol (#5100's review),
a genuinely pre-existing/unrelated red required check is a human-override decision, not something
this seat decides unilaterally — even though the evidence points the same way #5100's did (and
that one was approved for override, matching #5024's precedent). Flagging for the same call here.
Code and tests are approved on their merits; the only blocker is this CI status.

PR #5110, opened 2026-09-09
Fix #76519: add viewsheet-scoped structure read (GET /vs/structure/{id})

larryliang-inetsoft:
## Summary

Companion Java-side change for wiz bug #76519 (see the `stylebi-wiz` PR for the wiz-services half). `GET /ws/structure/{id}` hard-required the entry to resolve to a `Worksheet`, throwing otherwise — there was no way for the wiz backend to read what a viewsheet with no base worksheet is actually bound to (a logical model or a physical table).

## Changes

- `MetadataApiService`: refactored `getWorksheetStructure`'s table-walk body into a shared private `buildStructure` helper. Added `getViewsheetStructure(vsId, generation, principal)`: loads the viewsheet, requires a non-null `getBaseEntry()`/`getBaseWorksheet()` (loud, specific exceptions otherwise), and calls the same `buildStructure` against `viewsheet.getBaseWorksheet()` — this reuses `Viewsheet`'s own existing, production-exercised base-entry synthesis (`isDirectSource()` → `getWorksheet()`, the same generic path every normal viewsheet load already takes for a `LOGIC_MODEL`/`PHYSICAL_TABLE` base entry — not new runtime machinery). Added a `sourceKindOf(AssetEntry)` classifier (worksheet/logicalModel/physicalTable/unknown).
- `WorksheetStructure`: new `sourceKind`/`baseEntryPath` fields.
- `DatasourceMetaApiController`: new `GET /vs/structure/{id}` mapping, mirroring `/ws/structure/{id}` exactly (same `generation` param, same `WizUtil.decodeId`).
- 3 new unit tests for `sourceKindOf`, following the file's existing hand-built-`AssetEntry` convention.

## Evidence

- Full main-source compile: clean, 0 errors (6896 files).
- `MetadataApiServiceStructureTest`: `Tests run: 47, Failures: 0, Errors: 0, Skipped: 0` — real JUnit execution, independently re-run twice (once by the builder, once by an independent verifier).

Base is `stylebi-wiz-main-rebase` (this repo's own integration branch for wiz-driven fixes), matching the convention used by bug #76523's PR #5091.

🤖 Generated with a `/deep-debug-workflow` agent team run in `stylebi-wiz` — see `stylebi-wiz`'s `docs/teams/2026-09-08-bug-76519-viewsheet-external-binding/` for the full archive.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/DatasourceMetaApiController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/DatasourceMetaApiController.java
@@ -141,6 +141,26 @@ public class DatasourceMetaApiController {
    }
 
    /**
+    * #76519 — the viewsheet-scoped sibling {@code /ws/structure/{id}} never had: given a
+    * VIEWSHEET's own asset entry identifier (not a worksheet's), reports the structure of
+    * whatever it is actually bound to — a real base worksheet, or (when there is none) the
+    * viewsheet's own direct logical-model/physical-table binding, via
+    * {@link MetadataApiService#getViewsheetStructure}. Distinguish the two by the response's
+    * {@code sourceKind}.
+    *
+    * @param id the viewsheet asset entry identifier.
+    */
+   @GetMapping("/vs/structure/{id}")
+   public WorksheetStructure getViewsheetStructure(
+      @PathVariable("id") String id,
+      @RequestParam(value = "generation", required = false) Integer generation,
+      Principal principal) throws Exception
+   {
+      return metadataService.getViewsheetStructure(
+         WizUtil.decodeId(id), generation, (XPrincipal) principal);
+   }
+
+   /**
     * Gets all tables and FK relationships for the specified datasource.
     *
     * <p>{@code nameContains}/{@code limit}/{@code cursor} are all optional; omitting all three
--- a/core/src/main/java/inetsoft/web/wiz/model/WorksheetStructure.java
+++ b/core/src/main/java/inetsoft/web/wiz/model/WorksheetStructure.java
@@ -26,6 +26,11 @@ public class WorksheetStructure {
    private String name;
    private String primaryTable;
    private List<StructureTable> tables = new ArrayList<>();
+   // #76519 — only set by getViewsheetStructure: what kind of base the viewsheet actually has
+   // ("worksheet" | "logicalModel" | "physicalTable"), and that base entry's own path. null for
+   // a plain getWorksheetStructure (worksheet-id) call, which has no viewsheet wrapper to report.
+   private String sourceKind;
+   private String baseEntryPath;
 
    public String getPath() { return path; }
    public void setPath(String path) { this.path = path; }
@@ -35,6 +40,10 @@ public class WorksheetStructure {
    public void setPrimaryTable(String primaryTable) { this.primaryTable = primaryTable; }
    public List<StructureTable> getTables() { return tables; }
    public void setTables(List<StructureTable> tables) { this.tables = tables; }
+   public String getSourceKind() { return sourceKind; }
+   public void setSourceKind(String sourceKind) { this.sourceKind = sourceKind; }
+   public String getBaseEntryPath() { return baseEntryPath; }
+   public void setBaseEntryPath(String baseEntryPath) { this.baseEntryPath = baseEntryPath; }
 
    public static class StructureTable {
       private String name;
--- a/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/MetadataApiService.java
@@ -35,6 +35,7 @@ import inetsoft.uql.schema.XTypeNode;
 import inetsoft.uql.util.DefaultMetaDataProvider;
 import inetsoft.uql.util.XSourceInfo;
 import inetsoft.uql.util.XUtil;
+import inetsoft.uql.viewsheet.Viewsheet;
 import inetsoft.util.ThreadContext;
 import inetsoft.util.Tool;
 import inetsoft.web.composer.AssetTreeService;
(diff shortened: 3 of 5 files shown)


larryliang-inetsoft:

Head commit reviewed: `fc98b404bad46a2f06fd4e1043bdb8d365eba277`

**What was checked:** full diff read end to end; traced the actual permission-check call chain
through `AbstractAssetEngine.getSheet` and `Viewsheet.update()`/`reloadBaseWorksheet`/`getWorksheet`
in the checked-out worktree (`se-wt-76519/community`), not just the diff's own claims. Full findings
in `stylebi-wiz`'s `docs/teams/2026-09-08-bug-76519-viewsheet-external-binding/06-review-r1.md`.

The brief for this round specifically asked whether `/vs/structure/{id}}` really mirrors
`/ws/structure/{id}`'s permission posture or only superficially. It's superficial for the
real-worksheet-base case.

### Finding 2 — BLOCKER: `/vs/structure/{id}` discloses a viewsheet's base-worksheet structure
without checking the base worksheet's own permission

`MetadataApiService.getViewsheetStructure` calls `assetRepository.getSheet(entry, principal, true,
AssetContent.ALL)` with `entry` = the viewsheet's own `AssetEntry`. Traced
`AbstractAssetEngine.getSheet` (`core/src/main/java/inetsoft/uql/asset/AbstractAssetEngine.java:2340-2469`):
permission is checked exactly once, at line 2386, against the **top-level entry only** (the
viewsheet). For `ctype == AssetContent.ALL` (always true here), line 2457-2459 then unconditionally
calls `sheet.update(...)`, which populates `getBaseWorksheet()`.

`Viewsheet.update()` (`core/src/main/java/inetsoft/uql/viewsheet/Viewsheet.java:612-687`), for a
**real worksheet base**:
```java
if(wentry != null && wentry.isWorksheet()) {
   ws = (Worksheet) rep.getSheet(wentry, null, false, AssetContent.ALL);   // principal=null, permission=false
}
```
The base worksheet is loaded with **`null` principal and `permission=false`** — its own READ
permission is never checked. (Contrast the sibling `reloadBaseWorksheet`, lines 208-229, which for
the identical case DOES pass the real user and `permission=true` — but that method isn't what
`getSheet`'s generic `ALL`-content load path calls; `update()` is, and `update()` deliberately
bypasses the check.)

Practical effect: a principal with READ on a viewsheet only — explicitly denied READ on its
underlying worksheet, a legitimate existing StyleBI configuration for sharing a finished dashboard
while keeping the raw worksheet/query private — can call `GET /vs/structure/{vsId}` and receive the
full table/column/join/condition/aggregate structure of that private worksheet via the shared
`buildStructure` walk. That's more than the rendered viewsheet itself discloses (only the curated,
bound fields), and more than `GET /ws/structure/{worksheetId}` would let that same principal see
directly (its own `checkAssetPermission` on the worksheet entry would deny them). This is a new
disclosure surface — no endpoint previously exposed a viewsheet's base-worksheet full structure at
all.

The direct-source case (logical model / physical table) is different: `update()`'s
`isDirectSource()` branch passes the **real** user into `getWorksheet(...)`, and
`createTableAssembly` enumerates columns via `rep.getEntries(entry, user, ResourceAction.READ)` — a
permission-filtered read. I did not fully verify whether that throws or silently returns empty for a
denied principal, so I'm not claiming that half is broken too — flagging the real-worksheet-base
case as the confirmed defect.

**Suggested fix:** `getViewsheetStructure` should separately verify the caller's permission against
`viewsheet.getBaseEntry()` (the resolved worksheet/model/table) before returning its structure,
rather than relying on whatever `Viewsheet.update()`'s internal load happens to do.

### Checked, no finding
- Exception handling (`getViewsheetStructure`'s three specific `Exception` messages): same pattern
  `getWorksheetStructure` already uses; not a new defect from this diff.
- `sourceKindOf` classifier: verified against real `AssetEntry.isWorksheet()/isLogicModel()/
  isPhysicalTable()`, unit-tested for all three kinds plus an `"unknown"` fallback. Not overfit to
  any repro literal.
- `getWorksheetStructure`/`buildStructure` refactor: behavior-preserving, re-read end to end.

larryliang-inetsoft:
## Round 1 fix — addresses 06-review-r1.md's Finding 2 (BLOCKER)

**Finding 2 — `GET /vs/structure/{id}` under-checked permission.** `assetRepository.getSheet(entry, principal, true, AssetContent.ALL)` only checked READ on the *viewsheet's own* entry. For a real-worksheet base, `Viewsheet.update()` (the internal load that populates `getBaseWorksheet()`) loads that worksheet with `principal=null, permission=false` — its own READ was never checked. A principal with READ on the viewsheet but explicitly denied READ on its underlying worksheet (a legitimate, existing StyleBI configuration — share a finished dashboard, keep the raw query private) could read the full table/column/join/condition/aggregate structure through this endpoint, strictly more than `/ws/structure/{worksheetId}` would ever let that same principal see directly.

Fixed with a single explicit `assetRepository.checkAssetPermission(principal, baseEntry, ResourceAction.READ)` call right after the base entry is resolved, checked uniformly across all three base kinds (worksheet / logical model / physical table) — the review flagged the direct-source case as merely *unverified*, not confirmed-safe; this makes it explicitly, confirmed-safe by construction rather than resting on tracing `createTableAssembly`'s internal `getEntries` behavior.

New `MetadataApiServiceViewsheetStructureSecurityTest.java` (mock-based, following `AddVisualizationServiceSecurityTest.java`'s established pattern for this codebase): denied → `MessageException`, `getBaseWorksheet()` never reached; granted → proceeds past the gate. Real debugging trail in the fix doc — first two runs failed for an unrelated reason (the test's own placeholder `vsId` didn't parse into a real `AssetEntry`, silently null-ing the mock match), root-caused by reading `AssetEntry.createAssetEntry`'s actual source rather than loosening the assertion.

Java suite: `mvn test -pl core -Dtest=MetadataApiServiceStructureTest,MetadataApiServiceViewsheetStructureSecurityTest` → `Tests run: 49, Failures: 0, Errors: 0, Skipped: 0`, `BUILD SUCCESS`.

Full trace and design rationale: `docs/teams/2026-09-08-bug-76519-viewsheet-external-binding/07-fix-r1.md` (in the `stylebi-wiz` repo, this bug's coordination doc).

larryliang-inetsoft:
## Review round 2 — APPROVED

Head commit reviewed: `6278ac7a7` (delta from round 1's `fc98b404b`).

**What was checked:** independently re-traced Finding 2, not taking the writeup's claims at face
value — confirmed `AssetRepository.checkAssetPermission(Principal, AssetEntry, ResourceAction)` is
declared on the interface itself (`AssetRepository.java:124`), that `MetadataApiService.assetRepository`
is typed as that interface (`:2338`) so the new call compiles against real API, and that
`AbstractAssetEngine.checkAssetPermission`'s denial path throws `MessageException` — the same
exception class `/ws/structure` already relies on via `getSheet`'s own internal check, not a new,
differently-behaved error type.

**Placement verified**: the new `assetRepository.checkAssetPermission(principal, baseEntry,
ResourceAction.READ)` call sits after `baseEntry == null` is ruled out and before
`viewsheet.getBaseWorksheet()`/`buildStructure(...)` — so a denial throws before the structure is
ever serialized into the HTTP response. Applied **uniformly across all three base kinds** (no
`if(baseEntry.isWorksheet())` special-casing), which also resolves round 1's own explicitly-flagged
"unverified" direct-source case, not just the confirmed worksheet-base one.

**Test re-read in full, genuinely exercises the gate rather than passing regardless:**
`MetadataApiServiceViewsheetStructureSecurityTest.java`'s denied case asserts both the exception
AND `verify(viewsheet, never()).getBaseWorksheet())` — the second assertion is what actually proves
the check runs before the structure walk (a check placed after `buildStructure`, or an exception
sourced from mocking `getBaseWorksheet()` to throw instead, would fail this specific verify). The
granted case stubs `getBaseWorksheet()` to return `null` specifically so a *different*, later,
distinctive exception is the observed outcome — a deliberate marker proving execution passed the
gate, not an accident of both paths producing the same generic error.

**Independently re-ran** (not trusting the reported numbers):
```
./mvnw test -pl core -Dtest=MetadataApiServiceStructureTest,MetadataApiServiceViewsheetStructureSecurityTest -DfailIfNoTests=false -B
```
`Tests run: 49, Failures: 0, Errors: 0, Skipped: 0`, `BUILD SUCCESS` — matches exactly.

**No new findings from this round's own diff** — the check is a no-op for the ordinary granted
case (returns normally), no behavior change there.

**Verdict: approved.**

(Per round 1: `gh pr review --approve` also fails on this account with "Can not request changes on
your own pull request" — same self-authored-PR restriction applies to approval. Stating the verdict
here as the audit trail; see companion comment on stylebi-wiz#2342 for Findings 1 and 3.)

PR #5111, opened 2026-09-09
fix(#76522): ConditionVocabulary rejects a literal value that can't coerce to the field's type

NeverStopFMH:
## Summary
- `set_highlight`/`set_condition` with a bare-string `values` entry against a non-string-typed field (e.g. `{field:"ORDER_ID", operator:"greater_than", values:["CUSTOMER_ID"]}`) silently produced `values:[null]` and matched zero rows, reporting `{"ok":true}` with no error.
- Root cause: `ConditionVocabulary.value()` wraps any non-Map scalar as a literal `VALUE`-type condition value untouched; `ConditionUtil.java:304-306` (shared with the manual Composer UI, not modified here) later force-coerces it via `Tool.getData(dataType, value)`, which for a non-numeric string against a numeric field returns `null` on failure (swallowed inside `CoreTool`'s parsing chain) instead of surfacing an error.
- Fix: `ConditionVocabulary.value()` now threads the target field's `DataRefModel` through and, for a non-blank `String` against a non-`XSchema.STRING` field, eagerly attempts the same `Tool.getData` coercion one hop earlier — while the field name/index are still in scope — and throws a named `IllegalArgumentException` (naming the condition index, offending value, field name, and declared type, with a hint toward the working `{type:"field", field:"..."}` alternative) when it fails, instead of silently persisting a null-valued condition.
- `ConditionUtil.java` and `CoreTool.getData` are unchanged (confirmed shared with the manual Composer UI and 129+ unrelated call sites respectively).

## Test plan
- [x] `./mvnw test -pl core -Dtest=ConditionVocabularyTest` — 49/49 tests pass (46 existing + 3 new), including:
  - `refusesAStringValueThatCannotCoerceToTheFieldsDataType` — the exact repro (non-numeric string against an integer field) now throws a named error
  - `acceptsAStringValueThatDoesCoerceToTheFieldsDataType` — a genuinely numeric string (`"42"`) against an integer field still works unchanged
  - `acceptsAStringLiteralAgainstAStringTypedField` — the existing string/string literal path is unaffected
  - all pre-existing tests (including the `{type:"field", field:"..."}` typed-operand coverage) still pass unchanged

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/ConditionVocabulary.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/ConditionVocabulary.java
@@ -19,6 +19,8 @@ package inetsoft.web.wiz.viewsheet;
 
 import inetsoft.uql.JunctionOperator;
 import inetsoft.uql.XCondition;
+import inetsoft.uql.schema.XSchema;
+import inetsoft.util.Tool;
 import inetsoft.web.binding.drm.DataRefModel;
 import inetsoft.web.composer.model.condition.*;
 
@@ -196,7 +198,7 @@ public final class ConditionVocabulary {
          condition.setValues(new ConditionValueModel[] { rankingValue(values.get(0), index, fields) });
       }
       else {
-         condition.setValues(values.stream().map(raw -> value(raw, index, fields))
+         condition.setValues(values.stream().map(raw -> value(raw, index, fields, field))
                                 .toArray(ConditionValueModel[]::new));
       }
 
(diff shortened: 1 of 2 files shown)


NeverStopFMH:
## Round 1 review - DCG-010 (Redmine #76522)

**Verdict: changes requested - but the block is procedural, not a code defect.**

### Code review (diff: ConditionVocabulary.java + ConditionVocabularyTest.java only)

No blocking (important/critical) finding. Specifically checked and confirmed:

- The new guard keys on the resolved field's own declared dataType (targetField.getDataType()),
  not on any literal from the repro (no hardcoded field name/type) - it is a general rule fix.
- The thrown IllegalArgumentException message does name the condition index, the offending
  value, the field name, and its declared type, matching the PR description's claim.
- ConditionUtil.java and CoreTool.java are confirmed genuinely untouched (2 files changed total).
- Traced the one plausible sibling-branch concern - that the eager Tool.getData probe might not
  match ConditionUtil's actual downstream conversion for date-typed fields, which instead routes
  through AbstractCondition.getObject - and found no divergence: CoreTool's own date-parsing
  fallback chain lands on the same DateTime.parse(...) call AbstractCondition.getObject's isDate()
  gate uses, and is if anything more lenient. No false-positive rejection scenario found for a
  legitimate date literal.
- Non-blocking notes only: no test coverage for a date-typed target field (mitigated - this
  looks owned by the sibling DCG-004/DCG-012 PR #5109 in the same batch, and my own trace found
  no divergence for dates); the error message's hint always points at {type:"field",...} even
  when the misuse looks more like an intended {type:"variable",...} - still a strict improvement
  over the prior silent null, just a slightly generic hint.

### CI - the actual blocker

- claude-review: FAILURE, confirmed to be an infra crash (gh api .../pulls/5111/reviews returns
  []), not a posted finding - not treated as a real issue per this batch's own convention.
- build (Validation Build): FAILURE, and no longer pending. Root cause confirmed from the Actions
  log: inetsoft.web.wiz.controller.WizTabularControllerSecurityTest.mappingsMatchTheFrozenContract
  (an endpoint-list drift in WizTabularController, unrelated OAuth endpoints from #5096), nowhere
  near ConditionVocabulary/conditions/highlights. This is the same pre-existing, unrelated build
  failure already merged through by name in #5100/#5104/#5106 earlier in this batch, each with an
  explicit user go-ahead noted in its own merge commit.

I do not have that same explicit go-ahead for this PR, so per this review's own merge criteria
(real CI checks green before merge), I am holding rather than approving/merging. This needs either
an explicit go-ahead to merge through the same known-unrelated build red as the sibling PRs, or
the WizTabularControllerSecurityTest frozen-contract test fixed/updated upstream first. No code
changes are requested against ConditionVocabulary.java or its test.

jshobe-inetsoft (changes requested):
Requesting changes: the `build` check is currently failing on the latest commit (123eed670043e40b9b85838f71637727d8a4b11c), even though Claude's automated review (posted after this commit) found no blocking issues. All checks must pass before merge. Please fix the failing build and push an update.

fredaan (approved):

PR #5115, opened 2026-09-09
Fix #76556 (VFO-016): refuse dead basicGeneralPaneModel.enabled write

fredaan:
## Summary

- `spinnerGeneralPaneModel.generalPropPaneModel.basicGeneralPaneModel.enabled` (and the equivalent
  path for checkbox/textinput/slider/radiobutton/combobox) is populated on every dialog-model read
  but never read back by `VSInputService`'s apply methods for any of these six input types — a raw
  dotted-path write there via `set_assembly_properties` returned `ok:true` and silently changed
  nothing.
- Extends `PropertyAliases`' existing `DEAD_FIELDS`/`deadFieldWriteRefusal` mechanism (already used
  for `shadow`/`editable`/`container`/`refresh`/`locked` on other types) to refuse this write on the
  six confirmed-dead types, with a message pointing the caller at the live `enabled` alias
  (`generalPropPaneModel.enabled`).
- No change needed in `plugin/composer` — `set_assembly_properties` is a thin pass-through that
  already surfaces whatever `resolveForWrite` throws as a tool error.

Verified independently (not just from the diagnosis/refutation): re-grepped
`VSInputService.java`'s read/write methods for all six types (checkbox, textinput, slider,
radiobutton, combobox, spinner) and confirmed none of them ever call
`basicGeneralPaneModel.setEnabled(...)`/`.isEnabled()` — each only touches
`name`/`primary`/`visible`/`refresh`/`objectNames` (combobox additionally `editable`).

## Test plan

- [x] Added `refusesBasicGeneralEnabledOnTheSixInputTypesWhereItIsDead` to
      `PropertyAliasesTest.java`, asserting the raw path throws `IllegalArgumentException` for all
      six types, and that the `enabled` alias itself still resolves to and stays writable at
      `generalPropPaneModel.enabled`.
- [x] `./mvnw -pl core test -Dtest=PropertyAliasesTest -o` — 45/45 pass (44 pre-existing + 1 new).

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyAliases.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyAliases.java
@@ -149,6 +149,21 @@ public final class PropertyAliases {
     */
    private static final Set<String> LOCKED_LIVE_TYPES = Set.of("image", "line", "oval", "rectangle");
 
+   /**
+    * {@code basicGeneralPaneModel.enabled} is aliased through the shared {@link #basicGeneral}
+    * helper's own comment as dead everywhere -- "the Editable flag wearing the wrong name" -- but
+    * only the six input assemblies routed through {@code VSInputService} are confirmed here: their
+    * read methods (checkbox/textinput/slider/radiobutton/combobox/spinner, one apply method each)
+    * populate {@code basicGeneralPaneModel}'s name/primary/visible/refresh (and combobox's
+    * editable) from the real assembly, but never {@code enabled}; their write methods read
+    * {@code isPrimary()}/{@code getVisible()}/{@code isRefresh()} (and combobox's
+    * {@code isEditable()}) back, but never {@code isEnabled()}. A raw dotted-path write there
+    * reports success and changes nothing -- the live switch is the {@code enabled} alias, one
+    * level up at {@code generalPropPaneModel.enabled}.
+    */
+   private static final Set<String> BASIC_GENERAL_ENABLED_DEAD_TYPES =
+      Set.of("checkbox", "textinput", "slider", "radiobutton", "combobox", "spinner");
+
    private static final Map<String, TypeAliases> REGISTRY = registry();
 
    private PropertyAliases() {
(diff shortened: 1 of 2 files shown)


fredaan:
## Review round 1

Head commit reviewed: `2244e6831711eabf8be97bad005fa4d367a04931`

**Checked:** read the diagnosis/refutation/fix writeups, then the real diff and PR metadata.
Independently re-read the actual `VSInputService.java` (this exact worktree/head commit) for 4
of the 6 claimed-dead types (checkbox, textinput, spinner, combobox) rather than trusting the
cited line numbers — all four confirm `basicGeneralPaneModel.enabled` is populated on no read
path and consulted on no write path. Grepped
`basicGeneralPaneModel\.(set|is)Enabled|basicGeneralPaneModel\.getEnabled` across all of
`core/src/main/java/inetsoft/web` — zero hits anywhere, for any type — which also covers the
2 types not individually re-read (slider, radiobutton) and rules out indirect/reflection
wiring. Independently re-ran `./mvnw -pl core test -Dtest=PropertyAliasesTest -o`: **45/45
pass** (44 pre-existing + 1 new).

**Scope-of-fix (main check):** `BASIC_GENERAL_ENABLED_DEAD_TYPES` = exactly `checkbox,
textinput, slider, radiobutton, combobox, spinner` — matches what was actually audited, no
more, no fewer.

**Wiring pattern:** the new `deadFieldWriteRefusal` branch mirrors the existing
`REFRESH_DEAD_TYPES`/`LOCKED_LIVE_TYPES` branches (leaf check + type-set membership, message
naming the field and pointing at the `enabled` alias). Traced `resolveForWrite`'s two-phase
check to confirm the `basicGeneralPaneModel` substring guard correctly leaves the bare
`enabled` alias unaffected.

**Test quality:** `refusesBasicGeneralEnabledOnTheSixInputTypesWhereItIsDead` asserts the
mechanism (raw path throws) for all 6 types via a map, not just spinner's literal repro path,
and separately asserts the `enabled` alias still resolves to and stays writable at
`generalPropPaneModel.enabled` for each — a regression in either direction would fail this
test.

**Scope discipline:** PR body and code comment are honest that only the 6
`VSInputService`-routed types were audited; other `basicGeneral()`/`outputGeneral()`/
`dataGeneral()` callers (chart, gauge, text, image, selection*, table, crosstab, calctable,
tab, submit, ...) are explicitly flagged as not covered, not silently implied to be covered.

**Same bug, other site (non-blocking, not a finding on this diff):** the whole-directory grep
found zero read/write of `basicGeneralPaneModel.enabled` anywhere, for any type — meaning the
field is dead everywhere, and every other type going through `basicGeneral()` still silently
accepts-and-drops a raw write to it today. That's pre-existing and honestly out of this PR's
audited scope, not something this PR should be widened to fix without its own per-type audit.

**Prompt-touching-file gate:** not triggered. `plugin/composer/src/tools/propertyTools.ts`'s
`makeSetAssemblyPropertiesTool` is a thin pass-through — the refusal message surfaces as
ordinary tool-error text, not a prompt/instruction file.

**CI:** Semgrep (both) and CLA pass. `build` (Validation Build) was still in progress at
review time — recommend confirming green before merge, independent of this review. The
`claude-review` bot check failed, but its own log shows an infra failure (execution errored
before analyzing anything, empty `ANTHROPIC_API_KEY` in its env) — not a code-quality signal
on this diff.

## Findings

None (blocker or important).

## Verdict

**Approved.**

PR #5117, opened 2026-09-09
Fix rename-time dependency-field silent drop (Redmine #76545 item 1)

fredaan:
## Summary

- `AbstractAssetEngine.changeSheetDependents` only propagated a rename in one direction: into sheets the renamed sheet embeds (`sheet.getOuterDependents()`), updating their reverse-index `dependencies` field. It never propagated the reverse — into sheets that embed the renamed sheet (`sheet.getOuterDependencies()`) — to fix their own embed pointer via `AbstractSheet.renameOuterDependent`, which existed in all four `AbstractSheet` implementors (`Worksheet`, `Viewsheet`, `VSSnapshot`, `WorksheetWrapper`) but had zero callers.
- When a folder-rename cascade processes an embedding sheet before its embedded sheet in **both** the away (F→F2) and back (F2→F) halves of a rename-away-then-back round trip, the embedded sheet's `getOuterDependencies()` (read by `checkSheetRemoveable`, `MVManager`, `SaveWorksheetService`, `AbstractSheetAsset`, `AbstractVSAssembly`) is left containing the stale away-half identifier instead of the sheet's true final location — silently, with no log.
- Fix: add a second loop in `changeSheetDependents` that walks `sheet.getOuterDependencies()` and calls `renameOuterDependent(oentry, nentry)` on each embedding sheet, wiring in the previously-dead method. Also logs the existing loop's null-`dsheet` guard (previously silent), matching the existing `LOG.warn` precedent in `changeSheet0` for a null sheet lookup.

## Test plan

- [x] New regression test `core/src/test/java/inetsoft/uql/asset/ChangeSheetDependentsRenameSilentDropTest.java`: builds worksheet A (no embeds) and worksheet B (an outer `MirrorTableAssembly` pointing at A), drives `changeSheet0` directly through a folder-rename-away-then-back round trip with the embedder (B) processed before the embedded sheet (A) in both halves — the confirmed-reproducing ordering. Confirmed fail-before (`A.getOuterDependencies()` contains the stale `F2/B` identifier against the unfixed code) / pass-after (round-trips back to `F/B`, with B's own embed pointer also correctly `F/A`).
- [x] `./mvnw -pl core -Dtest=ChangeSheetDependentsRenameSilentDropTest test` — passes.
- [x] Broader regression sweep of `inetsoft.uql.asset.*` tests plus `PermissionMatrixSpecialTest` (which also subclasses `AbstractAssetEngine`) — all pass, no collateral breakage.

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/AbstractAssetEngine.java
+++ b/core/src/main/java/inetsoft/uql/asset/AbstractAssetEngine.java
@@ -3230,6 +3230,25 @@ public abstract class AbstractAssetEngine implements AssetRepository, AutoClosea
             clearCache(entry);
             storage.putXMLSerializable(identifier, dsheet);
          }
+         else {
+            LOG.warn("Unable to update outer dependency reference for renamed sheet: dependent entry not found: " + entry);
+         }
+      }
+
+      for(AssetEntry entry : sheet.getOuterDependencies()) {
+         String identifier = entry.toIdentifier();
+         IndexedStorage storage = getStorage(entry);
+         AbstractSheet esheet = (AbstractSheet)
+            storage.getXMLSerializable(identifier, null);
+
+         if(esheet != null) {
+            esheet.renameOuterDependent(oentry, nentry);
+            clearCache(entry);
+            storage.putXMLSerializable(identifier, esheet);
+         }
+         else {
+            LOG.warn("Unable to rename outer dependent reference for renamed sheet: embedding entry not found: " + entry);
+         }
       }
    }
 
--- a/core/src/test/java/inetsoft/uql/asset/ChangeSheetDependentsRenameSilentDropTest.java
+++ b/core/src/test/java/inetsoft/uql/asset/ChangeSheetDependentsRenameSilentDropTest.java
@@ -1,0 +1,365 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.uql.asset;
+
+import inetsoft.sree.internal.cluster.Cluster;
+import inetsoft.sree.security.Organization;
+import inetsoft.test.BaseTestConfiguration;
+import inetsoft.test.ConfigurationContextInitializer;
+import inetsoft.test.SreeHome;
+import inetsoft.test.SwapperTestConfiguration;
+import inetsoft.uql.XRepository;
+import inetsoft.uql.asset.internal.AssetFolder;
+import inetsoft.uql.util.AbstractIdentity;
+import inetsoft.util.IndexedStorage;
+import inetsoft.util.StorageRefreshListener;
+import inetsoft.util.TransformListener;
+import inetsoft.util.XMLSerializable;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+import org.w3c.dom.Document;
+
+import java.beans.PropertyChangeListener;
+import java.security.Principal;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.mock;
+
+/**
+ * Regression for Redmine #76545 item 1: {@code AbstractAssetEngine.changeSheetDependents} only
+ * propagated a rename into sheets an renamed sheet embeds ({@code getOuterDependents()}), never
+ * the reverse -- sheets that embed the renamed sheet ({@code getOuterDependencies()}, the field
+ * {@code checkSheetRemoveable}/{@code MVManager}/etc. read). When a folder-rename cascade
+ * processes the embedder before the embedded sheet in both the away and back half of a
+ * rename-away-then-back round trip, the embedded sheet's {@code getOuterDependencies()} is left
+ * pointing at the sheet's stale, away-half location instead of its true final location.
+ */
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(
+   classes = { BaseTestConfiguration.class, SwapperTestConfiguration.class,
+               ChangeSheetDependentsRenameSilentDropTest.XRepositoryConfig.class },
+   initializers = ConfigurationContextInitializer.class)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
+@SreeHome
+@Tag("core")
+class ChangeSheetDependentsRenameSilentDropTest {
+   @Configuration
+   static class XRepositoryConfig {
+      @Bean
+      public XRepository xRepository() {
+         return mock(XRepository.class);
+      }
+   }
+
+   @Autowired
+   private Cluster cluster;
+
+   private InMemoryIndexedStorage storage;
+   private TestAssetEngine engine;
+
+   @BeforeEach
+   void setUp() {
+      storage = new InMemoryIndexedStorage();
+      engine = new TestAssetEngine(cluster);
+      engine.istore = storage;
+   }
+
+   /**
+    * Folder F contains worksheet A (no outer embeds) and worksheet B (an outer
+    * {@code MirrorTableAssembly} pointing at A). The folder-rename cascade processes B (the
+    * embedder) before A (the embedded sheet) in both the away (F -> F2) and back (F2 -> F)
+    * halves of a round trip -- the ordering the diagnosis confirmed reproduces the defect.
+    */
+   @Test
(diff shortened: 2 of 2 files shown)


fredaan:
# 04-review-r1 - bug-76545-item1, PR #5117

## Round

Round 1. Head commit reviewed: 71b06e64af962f2665d992b116fb2d27ab3a50e0
(fix/redmine-76545-item1-rename-silent-drop, target
inetsoft-technology/stylebi, base stylebi-wiz-main-rebase).

## Method

Read 00-charter.md, 00-triage.md, 01-diagnosis.md (original + the
"Revision (after 02-refute.md)" section), 02-refute.md (original + the
"Recheck" section), and 03-fix.md in full before looking at the diff.
Pulled the PR into a standalone sparse checkout (avoided touching
E:\inetsoft\stylebi\community or its existing worktrees) since a Windows
long-path issue in web/ would otherwise block inspecting core/. Every
git/gh command run as its own unchained invocation, output sanity-checked
against what the diff/PR view themselves reported, per this tickets known
worktree/chaining gotcha (not strictly applicable here since no shared
worktree was touched, but the discipline was kept anyway).

## What was checked

Diff vs. diagnosis: the applied diff
(core/src/main/java/inetsoft/uql/asset/AbstractAssetEngine.java,
changeSheetDependents) matches 01-diagnosis.md's "Fix direction" section
verbatim: (1) a LOG.warn added to the existing loop's previously-silent
else branch, (2) a new, symmetric loop over sheet.getOuterDependencies()
calling the previously-dead AbstractSheet.renameOuterDependent(oentry,
nentry) on each embedding sheet, persisting via storage.putXMLSerializable
plus clearCache, with its own LOG.warn on a null lookup. Confirmed in the
pulled source that changeSheetDependents is called from exactly one call
site, changeSheet0 (line 3172), unconditionally, i.e. for both the
oentry.isWorksheet() branch and the else branch, so the fix is not
accidentally worksheet-only, consistent with the diagnosis's explicit
requirement that both loops run regardless of sheet type.

Confirmed-cause vs. reported-symptom check (this ticket's own required
pass, not just diff review):
- The fix is structural, not literal-keyed: it operates generically on
  sheet.getOuterDependencies()/renameOuterDependent, with no reference to
  the repro's specific folder/sheet names anywhere in the production code
  change. No special-case if excluding or targeting the reported scenario.
- No broken invariant left uncorrected at another call site the diagnosis
  named: re-verified independently (own grep, not trusting the diagnosis's)
  that changeSheetDependents has a single call site (changeSheet0, line
  3172), and that the four other getOuterDependencies() readers the
  diagnosis lists (MVManager.java:1350, SaveWorksheetService.java:148,
  AbstractSheetAsset.java:80, AbstractVSAssembly.java:105) are read-only
  consumers with no rename-time write responsibility of their own; the
  reasoning holds, nothing there needed to change for this fix to be
  complete.
- Re-derived, independently of both prior agents, that a second,
  pre-existing, already-wired mechanism (XRepository.renameTransform ->
  RenameTransformHandler -> AssetWSDependencyTransformer, unconditionally
  invoked from changeSheet0 at line 3125, before changeSheetDependents at
  line 3172 runs) asynchronously rewrites the exact same field this fix's
  new loop rewrites synchronously: MirrorAssembly.entry /
  MirrorAssemblyImpl's XML "source" attribute (confirmed by reading
  Worksheet.renameOuterDependent itself: massembly.setEntry(nentry) is the
  literal effect of the new loop, the same field
  AssetWSDependencyTransformer.renameWSSource rewrites via a different,
  async path). Traced whether the two mechanisms racing against each other
  could reintroduce staleness or clobber a correct value: whichever of the
  two runs first sets entry to nentry; whichever runs second reads entry
  already equal to nentry, so its own Tool.equals(entry, oentry) check is
  false and it no-ops on that field, then harmlessly re-persists the same
  object. Found no ordering under which this duplication produces a wrong
  final value for the field either mechanism writes; this is the same
  async-timing interaction 02-refute.md's Recheck section already examined
  for the dependencies field, and it holds for this pointer-rewrite field
  too. Not a blocking finding, but worth naming explicitly since neither
  prior document called out that renameOuterDependent's effect literally
  duplicates what the async transformer already does for the embed
  pointer; the new loop's actual, now-demonstrated value is exclusively for
  AbstractSheet.dependencies (the reverse index), not for "finally fixing
  the dangling embed pointer" as the diagnosis's original (pre-revision)
  framing implied and its revision correctly walked back.
- Swept AbstractAssetEngine.java for other getOuterDependents/
  addOuterDependency/renameOuterDependent call sites beyond
  changeSheetDependents (own grep, not the diagnosis's): found similar
  bookkeeping in the sheet-save/import path (lines 2618-2735) and in
  checkSheetRemoveable-adjacent code (lines 3860, 4009-4013, 4197). The
  save/import site is a different lifecycle event (persisting a sheet with
  its declared dependencies), not a rename cascade, and does not exhibit
  this bug's specific ordering defect; correctly out of scope for this
  fix, not a "same bug, other site" finding. The other sites are read
  paths inside checkSheetRemoveable territory, i.e. item 2's scope, as
  both 00-triage.md and the diagnosis already say.

Ordering/self-heal claim: independently re-derived (a third time, after
the debugger and the refuter) the embedder-first-both-halves vs.
embedded-first-both-halves asymmetry using the diff's own regression test
as the concrete instance, and it matches: the test drives exactly the
confirmed-reproducing ordering (B, the embedder, before A, the embedded
sheet, in both the away and back half), the only one of the two fixed
orderings the diagnosis's revision and the refuter's recheck converged on
as actually reproducing the defect.

Regression test: read
core/src/test/java/inetsoft/uql/asset/ChangeSheetDependentsRenameSilentDropTest.java
in full. It builds A (no embeds) / B (MirrorTableAssembly embedding A) in
an in-memory IndexedStorage, drives changeSheet0 directly in the
embedder-first ordering for both halves of the round trip, and asserts
A.getOuterDependencies() equals exactly {F/B} and B's own
MirrorTableAssembly.getEntry() equals F/A. This is not narrower than the
demonstrated defect; it asserts the exact field the diagnosis's revised
falsifiable claim names, not merely the original repro's literal values
with no generality. Verified in the PR's own CI log
(inetsoft.uql.asset.ChangeSheetDependentsRenameSilentDropTest) that the
test passed against the fixed code in this run, consistent with
03-fix.md's independently-reported local fail-before/pass-after run.

Harness discovery claim (WizAgentTestSupport vs. the manual replica): read
core/src/test/java/inetsoft/web/wiz/pairing/WizAgentTestSupport.java and
the new test's manual ContextConfiguration/XRepositoryConfig replica side
by side. The fixer's stated reason for not using the composed annotation
directly (avoiding relying on undocumented Spring meta-annotation classes
merge behavior for one extra bean) is a real, legitimate consideration:
Springs ContextConfiguration classes attribute on a directly-annotated
test class overrides rather than merges with the classes declared on a
meta-annotation by default (documented Spring TestContext behavior, not
actually undocumented, but the fixer's uncertainty about relying on it
correctly rather than accidentally dropping BaseTestConfiguration/
SwapperTestConfiguration is reasonable caution, not overcaution); the
manual replica the fixer wrote is no more verbose than correctly composing
meta-annotations would have been and avoids the failure mode entirely.
Judgment: reasonable engineering call either way, not a finding.

Counter-example attempts (tried and could not break it):
- Called changeSheet0 for a sheet with an empty getOuterDependencies()
  (no embedders): the new loop's for over an empty array is a correct
  no-op; confirmed getOuterDependencies() never returns null
  (Worksheet.getOuterDependencies(boolean) always returns an array built
  from the backing HashSet, even when empty).
- Considered a sheet with two embedders: the new loop processes each
  independently with its own storage read/write; found no shared mutable
  state between iterations that a second embedder's processing could
  corrupt.
- Considered whether wiring in renameOuterDependent regresses any of the
  other three AbstractSheet implementors (Viewsheet, VSSnapshot,
  WorksheetWrapper) that were not directly re-read this round: the change
  itself is polymorphic dispatch through an existing abstract method with
  four already-reviewed (by the diagnosis) implementations; the PR does
  not modify any of those four implementations, so no new behavior was
  introduced in them by this diff. Coverage gap only: no viewsheet-specific
  regression test exists for this same mechanism (Viewsheet has the
  identical getOuterDependents/getOuterDependencies/renameOuterDependent
  shape per the diagnosis's own reading); flagged below as non-blocking,
  since the fixed code is the single shared method, not a per-type branch,
  and the diagnosis already read all four implementations for correctness.
- Tried to construct a scenario where the new LOG.warn additions change
  behavior (not just observability): both are in else branches that
  previously did nothing; no control-flow change, no exception thrown
  where none was thrown before.

## Findings

No blocking (blocker/important) findings.

Non-blocking notes for the record (not required before merge):
1. The new loop's demonstrated value is narrower than "fixes the dangling
   embed pointer" (the diagnosis's original framing, before its own
   revision); it duplicates, rather than replaces, what the pre-existing
   async RenameTransformHandler/AssetWSDependencyTransformer mechanism
   already does for MirrorAssembly.entry, and the two were not previously
   examined side-by-side for this specific redundancy. Traced above; found
   no case where the redundancy produces a wrong value, only a harmless
   double-write. Worth a one-line code comment for a future reader, not
   worth blocking this PR.
2. No viewsheet-embeds-viewsheet (or viewsheet-embeds-worksheet)
   regression test alongside the worksheet-embeds-worksheet one added;
   same shared method, so not a demonstrated gap in the fix itself, just a
   coverage gap a future viewsheet-rename regression would have to
   rediscover from scratch.

## Merge-readiness (non-code)

- gh pr view --json mergeable: MERGEABLE, no conflicts.
- gh pr checks: Semgrep Scan, Semgrep OSS, claude-review, and license/cla
  all passed. The build check (Maven, inetsoft-core) failed on both the
  original run and a re-run (gh run rerun --failed), with byte-identical
  failures each time: 3 pre-existing test failures in
  inetsoft.report.script.viewsheet / inetsoft.uql.viewsheet.graph
  (JavaScriptEngineTest.testGetImages, ImageVSAScriptableTest.testGetSetImage,
  ViewsheetScopeTest.testAddImage), all on "IIO Can't get input stream from
  URL!" - ViewsheetScopeTest.testAddImage literally fetches
  https://www.inetsoft.com/images/... over the network at test time, so
  this is CI-runner network-egress flakiness, not something a re-run or a
  code change to this diff can fix. This PR's own new test,
  ChangeSheetDependentsRenameSilentDropTest, passed in both runs, and
  nothing in either failing test touches inetsoft.uql.asset or the
  rename/dependency code this PR changes.
- Checked whether this repo currently treats the build check as a hard
  merge gate: PR #5115 (open) and PR #5116 (merged 2026-09-09, same day)
  both show build check failures for reasons unrelated to their own
  diffs (an inetsoft-web TypeScript compile error, a different,
  pre-existing failure than this PR's) - #5116 was merged anyway. This
  confirms current repo practice tolerates an unrelated, pre-existing
  build-check failure rather than blocking merge on it, consistent with
  treating this PR's own unrelated network-flake failure the same way.

## Verdict

Approved. No blocking findings from the code review. The failing build
check is confirmed pre-existing, network-environment flakiness unrelated
to this diff (reproduced identically across two runs, in a different
module than the one this PR touches), and this repo's own recent
practice (PR #5116) already merges through this class of unrelated
build-check failure.

PR #5118, opened 2026-09-09
Bug #76518, only skip the DC orphaned-cell heuristic when part is truly the faceted dimension

ngovus-inetsoft:
## Problem

`DateComparisonUtil.applyDateRange()` skipped its orphaned-cell heuristic (`computeValidParts()` / `ValidPartsSelector`) whenever `info.isFacet() || dcInfo.isCompareAll()`.

`VSChartInfo.isFacet()` is a **single chart-wide boolean**: `GraphGenerator.createCoord()` sets it from `coordinate instanceof FacetCoord`, and a `FacetCoord` is produced whenever *any* axis carries 2+ dimensions — regardless of whether the DC part column is one of them.

So the skip added for Bug #76388 (where the part column genuinely *is* the faceted dimension, e.g. a DayOfWeek facet, and "has the most recent period chronologically reached this part yet?" is meaningless) also fired when an unrelated dimension such as Region shared the DC axis. There the part column is still a plain, chronologically ordered axis, the heuristic is still meaningful, and skipping it let a genuinely unreached future part — December while the current year has only reached April — render for the prior comparison periods.

## Fix

Decide from the **actual coordinate tree** instead of the chart-wide flag.

`createCoord()` consumes the innermost dimension of each axis as the plot axis and wraps every remaining outer dimension in a `FacetCoord`, so a field found in a `FacetCoord`'s *outer* coordinate is one the graph is genuinely faceted on. The new `isFacetedField()` walks that structure (recursing through nested facets) and matches on `fields[0]` exactly as `applyDateRange()`'s own part-scale lookup does, so the two always agree on which scale is the part column's.

`info.isFacet()` is kept as a cheap short-circuit so non-facet charts don't walk the coordinate at all.

### Why not `DateComparisonInfo.isUseFacet()`

Considered and rejected: it was `false` in #76388's own scenario. The facet there comes from `ChartDcProcessor` **appending** `periodRef` to the base axis for non-`ALL` granularity (giving X = `[part, period]`, so `period` becomes the inner plot axis and `part` the facet level), not from the DC dialog's facet setting. Keying on `isUseFacet()` would regress that fix.

## Tests

`DateComparisonUtilApplyDateRangeTest`'s fixture could no longer express the distinction with a bare `isFacet()` boolean, so it now builds the three real coordinate shapes:

| Shape | Coordinate | Heuristic |
|---|---|---|
| `NO_FACET` | plain `RectCoord` | runs |
| `PART_IS_FACET` | part scale in the `FacetCoord`'s **outer** coord (#76388) | skipped |
| `UNRELATED_FACET` | part scale in the **inner** coord, unrelated dim faceted (#76518) | runs |

That re-enables the previously `@Disabled` `facetCausedByUnrelatedDimensionStillOrphansUnreachedChronologicalPart()` and keeps `facetModeDoesNotOrphanFacetsTheMostRecentPeriodLacks()` as the #76388 guard.

**Verified both directions:**
- Reverting the condition to bare `info.isFacet()` fails the #76518 test with the exact reported message: `row for period=2019-01-01, part=51 expected accepted=false but was true`
- Stubbing `isFacetedField()` to return `false` fails the #76388 test

Still passing: `DateComparisonUtilValidPartsTest`, `DateComparisonFormatOrphanedFacetTest`, `DateComparisonIntervalConditionTest`, `DcFormatTableLensTest`, `DCMergeDateFilterTest`, `DCMergeDatePartFilterTest`, `VSChartShowDataServiceTest`.

No live UI verification: as the reporter already noted, the available dataset's older comparison periods don't contain data past the current period's reach, so there is nothing for the bug to incorrectly reveal. The defect reproduces directly at the `applyDateRange()` level, independent of dataset shape.

Related: #76388, #76389, #76390, #76391 — same `applyDateRange()`/`computeValidParts()` area.

A companion docs-only PR in the enterprise repo adds the missing `claude/chart-date-comparison.md` coverage of this layer.

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/DateComparisonUtil.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/DateComparisonUtil.java
@@ -21,6 +21,8 @@ import inetsoft.util.*;
 import inetsoft.util.data.CommonKVModel;
 import inetsoft.graph.*;
 import inetsoft.graph.aesthetic.VisualFrame;
+import inetsoft.graph.coord.Coordinate;
+import inetsoft.graph.coord.FacetCoord;
 import inetsoft.graph.data.CalcColumn;
 import inetsoft.graph.data.DataSet;
 import inetsoft.graph.element.GraphElement;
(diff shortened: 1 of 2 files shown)


jshobe-inetsoft (approved):
Claude's automated review found no blocking issues. The build check initially failed/was cancelled on unrelated flaky tests (ImageVSAScriptableTest/ViewsheetScopeTest external image-URL fetch errors, seen identically on other unrelated PRs); reran and it now passes cleanly. Approving.

PR #5120, opened 2026-09-09
Bug #76538, resolve dynamic column binding for form input write-back

ngovus-inetsoft:
## Problem

A form input (Slider, Spinner, TextInput, RadioButton, ComboBox) whose **Data** tab column is bound to *A Variable* or *An Expression* does not write its selected value into the target embedded table cell, so the bound table keeps its original data.

Reported as "expression and variable to set all options of form component can't work". The reporter's `Form_Options` asset binds `columnValue` to `='state'` / `='product_name'` / `$(PField)` across five inputs, none of which write back. The min/max/increment and minDate/maxDate expressions in the same asset *do* evaluate — only the write-back is broken. The slider's rendered maximum also differs (1042 vs the expected 1454) purely because `=sum(Query1['customer_id'])` is summing the un-written column.

## Root cause

`ViewsheetSandbox.refreshVariable()` looked the column up by the **design time** text:

```java
DataRef column = eassembly.getColumnSelection(false).getAttribute(iassembly.getColumnValue());
```

`getColumnValue()` returns `columnValue.getDValue()` — the literal `='state'` / `$(PField)` string, which never matches a column name. `column` resolves to `null` and the `edata.setObject(row, col, cdata)` write is skipped entirely. The row binding directly beside it already reads the runtime value via `getRow()`.

This regressed in #4541 (Bug #75929), which replaced `iassembly.getColumn()` with the `getColumnValue()` lookup to avoid a stale cached `DataRef`. `getColumn()` is resolved *from the runtime value* by `InputVSAssemblyInfo.update()`, so the original code handled dynamic bindings; the replacement does not. Static column bindings were unaffected, which is why #75929's own scenario passed.

Ordering is not a factor: `outputDataChanged()` runs `updateAssembly()` (→ `executeDynamicValues`) immediately before `refreshVariable()`, so the runtime value is available by then.

## Fix

Add `InputVSAssemblyInfo.getRuntimeColumnValue()`, which reads the executed value of the `columnValue` `DynamicValue`, and use it for the lookup. `DynamicValue.getRValue()` already falls back to the design value for non-dynamic text, so static bindings still resolve against the live `ColumnSelection` and keep the freshness property #75929 added. The cached `DataRef` remains as a fallback for any path that reaches `refreshVariable()` without a prior `updateAssembly()`.

Only this one call site resolved an input's target column from `getColumnValue()`; the other runtime consumers (`InputVSAQuery`, `CoreLifecycleService`) already use the runtime-resolved `getColumn()`.

## Tests

New `RefreshVariableTest`, following the reflection pattern of `ApplyParameterToInputTest` in the same package. Five cases: literal binding, `='state'` expression, `$(PField)` variable, an unresolvable column, and the #75929 stale-cached-column case.

Verified the tests catch the defect — with the fix reverted, the expression and variable cases fail while the other three still pass. `mvn test -pl core` for `inetsoft.report.composition.execution.*` and `inetsoft.web.viewsheet.service.*`: 89 tests, all green.

## Notes for the reporter

Two things noticed while tracing that are **not** addressed here:

1. The expected screenshot shows the calendar combo box as `2023-06-13` while the current build renders `2023-6-13` — a date-format difference independent of the write-back defect. Needs its own ticket if it is in scope.
2. Several `DynamicValue` fields on input assemblies are never added to `getDynamicValues()`/`getViewDynamicValues()` and so could never resolve an expression (slider tick/label visibility flags, `ListInputVSAssemblyInfo.sortTypeValue`/`selectFirstItem`, `TextInputVSAssemblyInfo.defaultText`, `submitOnChange`/`writeBackValue`, and the range-slider `TimeInfo` fields). None are reachable from the property dialogs today — they render as plain checkboxes and selects, not `dynamic-combo-box` — so no user can currently enter a `$(var)`/`=expr` for them.

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/ViewsheetSandbox.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/ViewsheetSandbox.java
@@ -4032,8 +4032,19 @@ public class ViewsheetSandbox implements Cloneable, ActionListener {
             EmbeddedTableAssembly eassembly = (EmbeddedTableAssembly) ass;
             // resolve the column from the current column selection instead of relying on
             // iassembly.getColumn(), which is only refreshed on a full sandbox reset and
-            // may be stale if the binding was changed without reopening the viewsheet
-            DataRef column = eassembly.getColumnSelection(false).getAttribute(iassembly.getColumnValue());
+            // may be stale if the binding was changed without reopening the viewsheet.
+            // use the runtime column value so a variable or expression binding resolves
+            // to the column it evaluates to instead of the literal $(var)/=expr text.
+            String cname = iassembly.getRuntimeColumnValue();
+            DataRef column = cname == null
+               ? null : eassembly.getColumnSelection(false).getAttribute(cname);
+
+            // a dynamic value that has not been executed yet resolves to null, fall back
+            // to the column resolved by InputVSAssemblyInfo.update()
+            if(column == null) {
+               column = iassembly.getColumn();
+            }
+
             int row = iassembly.getRow();
 
             if(column != null && row > 0) {
--- a/core/src/main/java/inetsoft/uql/viewsheet/InputVSAssembly.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/InputVSAssembly.java
@@ -147,6 +147,15 @@ public abstract class InputVSAssembly extends AbstractVSAssembly implements Bind
    }
 
    /**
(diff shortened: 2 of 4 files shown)


jshobe-inetsoft (approved):
Claude's automated review found no blocking issues. The build check initially failed/was cancelled on unrelated flaky tests (ImageVSAScriptableTest/ViewsheetScopeTest external image-URL fetch errors, seen identically on other unrelated PRs); reran and it now passes cleanly. Approving.

PR #5121, opened 2026-09-09
Bug #76529, base the week start on week.start alone and fix the Sunday-hardcoded rewind

ngovus-inetsoft:
## The reported symptom

Importing `Year_WeekToDate.vso` on a Chinese-locale server produced a phantom leading week bucket in the chart (the duplicated `2021-01-04` label), while every other Date Comparison case on the same box computed Sunday-based weeks.

Both defects in the report are confirmed. Investigating turned up three related problems alongside them.

## 1. `week.start` silently followed the thread locale

`Tool.getFirstDayOfWeek()` fell back to `GregorianCalendar.getInstance(locale).getFirstDayOfWeek()` whenever `week.start` was unset — and `defaults.properties:236` ships it blank, so that was the normal path.

The locale is `ThreadContext.getLocale()`, which resolves to the **viewing user's** `SRPrincipal` locale on a `GroupedThread` and only degrades to the JVM default elsewhere. So the week start varied per user and per thread, with nothing in the product saying so — two users could see different week bucketing on the same dashboard.

It also already caused a **SQL/Java split**. `DateRangeRef.getExpression()` suppresses SQL week pushdown only when the first day of week is not Sunday **and** `Tool.getWeekStart() != null`, and `getWeekStart()` returns null unless the property is explicitly a valid day name. Blank `week.start` + a Monday locale meant pushdown still happened: the database grouped weeks one way while every in-memory path grouped them another. `VSCalendar.java:863,875` encode the same "no explicit `week.start` means Sunday" assumption.

**Fix:** blank or invalid `week.start` now means Sunday on every thread. This makes the first day of week a pure function of one visible property, and makes an assumption three existing call sites already rely on actually true. An invalid value is logged instead of silently absorbed.

## 2. The `wy`/`wm`/`wmq` rewind was hardcoded to Sunday

These date parts set the calendar's first day of week from `week.start`, then read `WEEK_OF_MONTH` from it — but the rewind in between was `-(DAY_OF_WEEK - 1)`. `DAY_OF_WEEK` is always Sunday-anchored and is unaffected by `setFirstDayOfWeek()`.

Under a Monday start that puts a Sunday — the *last* day of its week — at offset 0, so it reads the leading partial week's `WEEK_OF_MONTH` of 0 and encodes `(month+1)*10 + 0`, outside the `[11,125]` domain the scheme allows:

| date | old rewind to | wom | bucket | fixed |
|---|---|---|---|---|
| Jan 1 2021 (Fri) | -5 -> Dec 27 | 3 | 123 | 124 |
| Jan 3 (Sun) | -0 -> Jan 3 | **0** | **10** | 124 |
| Jan 4 (Mon) | -1 -> Jan 3 | **0** | **10** | 11 |
| Jan 10 (Sun) | -0 -> Jan 10 | 1 | 11 | 11 |

Two real weeks become three buckets, the extra one carrying the invalid encoding `10`. That is the phantom bar.

Three correlated sites had to move together:

- `DateComparisonInfo.getToDateWeekOfMonth()` produces the `forceDcToDateWeekOfMonth` argument those same expressions are compared against. It had the same Sunday rewind on a calendar whose week start *was* set correctly, and returned **0** for early-January to-dates — which every consumer reads as "not forced" (`forceWM > 0`), disabling the month-boundary folding that would have absorbed the leading week into December.
- `DateComparisonInfo:1929` was a fourth variant, `-(dow - getFirstDayOfWeek())` with no `% 7`, which moves the calendar **forward** a week when `dow < firstDayOfWeek`. It decides the DC period start.
- `DCMergeDatePartFilter:314` deliberately mirrors `datePart('wy')` (Bug #75351). Fixing one without the other would make them disagree on *every* row rather than only at month boundaries — strictly worse than the status quo.

All now go through `DateComparisonUtil.moveToWeekStart()`, which is a **strict no-op under a Sunday week start**: `(dow - 1 + 7) % 7 == dow - 1`. Nothing changes on a Sunday-start system, which is the regression argument for this change.

## 3. Three problems found alongside

- **`datePart` inherited a week start from a shared mutable `ThreadLocal`.** `CoreTool.calendar` is shared per thread and `CalcDateTime.date():766` sets its first day and never restores it. The documented two-argument script form `datePart(interval, date)` coerces `applyWeekStart` to `false` and so depended on what had run earlier on the same pooled thread. It is now deterministically Sunday-based in that branch (the old code was already half-broken here: Sunday rewind, but a locale-dependent `WEEK_OF_MONTH` read).
- **`DateTimeProcessor.getYearOfFullWeek()`** built a bare `GregorianCalendar`, so it started weeks wherever the JVM default locale said rather than at the configured week start — putting the returned year **off by one** at the Dec/Jan boundary. Covered by a test that runs the same input under `Locale.US` and `Locale.UK`.
- **The EM "Week Start Day" field was free text**, so a typo fell through to the locale default with no feedback. It is now a dropdown reusing the list already defined for the same property on the EM Properties page. An existing value is only case-normalized, never rewritten, so an unrecognized value cannot be blanked by an unrelated save.

## Not changed

`DateTimeProcessor.getMonthOfFullWeek()`'s flat `-7` rewind is off by one week under every week start, but an exhaustive sweep (365 days x forceWM -1..5 x 3 week starts, against the week-start-correct sibling `getMonthPartOfFullWeek`) shows the returned month is unaffected — the `forceWM` branch it feeds cannot be reached by a real forced value of 1-5. Left in place with a comment rather than making an untestable behavior change.

## Testing

New coverage, all driving the week start through `week.start` via the new `inetsoft.test.WeekStartUtil` helper (the previous `ThreadContext.setLocale` idiom stops working by design):

- `ToolWeekStartTest` — blank/invalid `week.start` gives Sunday under `en_US`/`en_GB`/`zh_CN`/`de_DE`/`ja_JP`; `getWeekStart()` and `getFirstDayOfWeek()` never disagree; a change is visible immediately.
- `DateRangeRefWeekPushdownTest` — pushdown happens only when the effective week start is Sunday, phrased as an invariant across locales.
- `DateComparisonInfoWeekStartTest` — `getToDateWeekOfMonth()` matches the `datePart('wy')` week it is compared against, for every date of 2021 under three week starts, and is never 0.
- `DateTimeProcessorTest` — full-week levels agree with the `datePart('wy')` oracle; `getYearOfFullWeek` is independent of the JVM default locale.
- `JavaScriptEngineTest` — Monday-start values for `wy`/`wm`/`wmq`; a two-year sweep asserting every day of a week lands in one bucket and no bucket encodes week 0; a Sunday-start snapshot proving the change is a no-op there; and the shared-`ThreadLocal` determinism case.

Each was confirmed to fail without its corresponding fix. `DCMergeDatePartFilterTest`'s Bug #75351 case needed a new date pair — its old dates now agree under the corrected encoding.

Full `core` suite: 4882 tests, only the two `RefreshVariableTest` failures that also fail on `main`. EM: 378 tests pass, lint clean. Verified live on a locally-run server.

## Upgrade note

Deployments that never set `week.start` and run a Monday-start locale will see week buckets shift. They were already bucketing inconsistently between users and against their own SQL; setting `week.start` explicitly preserves the previous behavior. Worth a release note.

Also worth confirming with the reporter: `zh_CN` gives Monday under the JDK's default CLDR data, but **not** on the stock Docker image, which pins `-Djava.locale.providers=COMPAT,SPI` where `zh_CN` is Sunday. If they are on a stock container, the Monday came from the per-user locale path rather than the server locale.

### DIFF
--- a/core/src/main/java/inetsoft/report/filter/DCMergeDatePartFilter.java
+++ b/core/src/main/java/inetsoft/report/filter/DCMergeDatePartFilter.java
@@ -23,6 +23,7 @@ import inetsoft.report.lens.AbstractTableLens;
 import inetsoft.uql.asset.DateRangeRef;
 import inetsoft.uql.viewsheet.VSDimensionRef;
 import inetsoft.uql.viewsheet.XDimensionRef;
+import inetsoft.uql.viewsheet.internal.DateComparisonUtil;
 import inetsoft.util.Tool;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -302,7 +303,8 @@ public class DCMergeDatePartFilter extends AbstractTableLens implements TableFil
                // to a different actual week in another period's year (e.g. the previous year's
                // bar shares the current year's bucket). Recompute the part value from this
                // cell's own date using the same logic as JavaScriptEngine.datePart("wy") for the
-               // non-to-date case -- move to the first day of the week, then month*10 +
+               // non-to-date case -- move to the first day of the week (honoring the
+               // configured week start, in lockstep with that method), then month*10 +
                // weekOfMonth -- so the equivalence cell matches the data/label for the cell's
                // actual week regardless of the configured first day of week. Deriving from the
                // stored part value instead (the previous approach) failed when the first day of
@@ -311,7 +313,7 @@ public class DCMergeDatePartFilter extends AbstractTableLens implements TableFil
                cal.setFirstDayOfWeek(Tool.getFirstDayOfWeek());
                cal.setMinimalDaysInFirstWeek(7);
                cal.setTime((Date) getDateGroupValue());
-               cal.add(Calendar.DATE, -(cal.get(Calendar.DAY_OF_WEEK) - 1));
+               DateComparisonUtil.moveToWeekStart(cal);
                int equivalenceValue =
                   (cal.get(Calendar.MONTH) + 1) * 10 + cal.get(Calendar.WEEK_OF_MONTH);
 
--- a/core/src/main/java/inetsoft/uql/asset/DateTimeProcessor.java
+++ b/core/src/main/java/inetsoft/uql/asset/DateTimeProcessor.java
@@ -166,6 +166,10 @@ class DateTimeProcessor {
          int weekOfMonth = getWeekOfMonth();
 
          if(weekOfMonth < 1) {
+            // NOTE: -7 lands in the week *before* this date's own week, so the weekOfMonth
+            // read below is one short of what datePart('wy') computes for the same date. It
+            // only feeds the forceDcToDateWeekOfMonth comparison, which a forced value of 1-5
+            // can never reach from here, so the returned month is unaffected either way.
             jcalendar.add(Calendar.DATE, -7);
          }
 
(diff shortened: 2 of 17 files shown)


jshobe-inetsoft (approved):

PR #5124, opened 2026-09-10
fix(#76542): canonicalize showType tokens/domain in set_assembly_properties

lizhepu:
## Summary
- `selectionGeneralPaneModel.showType` and `calendarAdvancedPaneModel.showType` are plain primitive `int`s aliased under the identical short name `showType`, so `PropertyPath.coerce()` had no domain/alias table for either: a string token like `"dropdown"` fell straight to `Double.parseDouble`, throwing `'...showType' is a int; 'dropdown' is not a number.` with no valid values named — the reported gap.
- The two domains overlap and disagree (Selection: `list`=0/`dropdown`=1; Calendar: `calendar`=1/`dropdown`=2), so any fix keyed by leaf name alone would silently misapply `"dropdown"` on whichever domain it didn't target — worse than today's loud-but-unhelpful error.
- Adds `canonicalShowType(type, resolvedPath, value)` to `AssemblyPropertyService`, modeled directly on the already-shipped `ChartRegionPropertyService.canonicalRotation()` (same "same leaf name, different non-interchangeable domains" shape, solved the same way: pre-transform in the calling service, before `PropertyPath.set/coerce()`, keyed by the resolved full path). Accepts a case-insensitive token or an already-numeric in-domain value; rejects everything else (unknown token, or an out-of-domain int like `{showType: 2}` on a SelectionList, previously silently stored) naming the valid tokens/ints.
- Additive/scoped: any path with no table entry (every property except `showType`) passes through unchanged. `sortType`/`mode`/`linkType`/`rangeType`/`refType`/`newObjectType` are the same class of gap but explicitly out of scope for this ticket (flagged with a one-line pointer for a future fix). `SheetPropertyService`/`ChartRegionPropertyService` are untouched — `showType` is structurally unreachable through either.

## Test plan
- [x] `./mvnw -q -pl core test -Dtest=inetsoft.web.wiz.viewsheet.AssemblyPropertyServiceTest` — 38/38 pass (30 pre-existing + 8 new covering: dropdown/Dropdown/list tokens on SelectionList; dropdown/calendar tokens on Calendar resolving to Calendar's own domain, not Selection's; unrecognized token naming valid values; out-of-domain int rejected; already-numeric in-domain int unaffected).
- [x] `./mvnw -q -pl core test -Dtest="inetsoft.web.wiz.viewsheet.*Test"` — full package (26 classes, incl. `ChartRegionPropertyServiceTest` 45/45 and `PropertyPathTest` 47/47) — no regressions.

Fixes Redmine #76542.


lizhepu:
Closing without merging: this branch was forked off the local stylebi-wiz-main-rebase branch instead of origin/main, so its diff against main carries ~1700 unrelated commits. Reverting this attempt per user instruction rather than rebasing it — will redo cleanly off origin/main if this bug is revisited.

PR #5125, opened 2026-09-10
fix(wiz): resolve a sharing date comparison's real config on read (bug 76522, DCG-003)

NeverStopFMH:
## Summary
- `get_date_comparison` always reported a `shareAssembly`-configured assembly as `enabled:false`, even when the share was genuinely live — `DateComparisonDialogService.isDateComparisonEnabled()` called `DateComparisonUtil.isDateComparisonDefined(info, false)` (`checkShareFrom=false`), copy-pasted from `getShare()`'s own, different-purpose usage (commit `9ea28277e8`). Every rendering-facing caller of the same predicate uses `checkShareFrom=true`, which is why the share rendered correctly while the read-back lied about it.
- Flipping that boolean alone is not a safe standalone step: `DateComparisonService.read()`'s populated fields (`comparisonOption`/`useFacet`/`period`/`interval`) come from the target assembly's own `DateComparisonPaneModel`, which is always a default-constructed, unset one while `comparisonShareFrom` is set. Landing just the flip would flip `enabled` to `true` while still reporting the target's own unset defaults as if they were the share source's real config — a plausible-but-wrong result. So `read()` now also resolves those fields through `DateComparisonUtil.getDateComparison()` (the existing render-time resolver, which already handles nested shares) whenever `comparisonShareFrom` is set.
- Both halves land together in this PR, per review guidance from this bug's debug-workflow batch.

## Test plan
- [x] `DateComparisonUtilTest` (new) — documents the intentional `checkShareFrom` asymmetry directly on `isDateComparisonDefined(info, checkShareFrom)`.
- [x] `DateComparisonDialogServiceTest` (new) — a real, unmocked call to `isDateComparisonEnabled()` proving it now agrees with every rendering-facing caller's `checkShareFrom=true` convention. Confirmed failing pre-fix (`expected <true> but was <false>`), passing post-fix.
- [x] `DateComparisonServiceTest` — two new cases: resolving `read()`'s fields through to the share source's real config, and falling back to the target's own model when the share source cannot be resolved. Confirmed the first fails pre-fix (`expected <change> but was <null>`), passes post-fix.
- [x] `mvn -pl core test -Dtest=DateComparisonServiceTest,DateComparisonUtilTest,DateComparisonDialogServiceTest,DateComparisonFormatOrphanedFacetTest,DateComparisonIntervalConditionTest,DateComparisonUtilApplyDateRangeTest,DateComparisonUtilValidPartsTest` — all green (94 tests across the DateComparison test family, 0 failures).

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/vs/dialog/DateComparisonDialogService.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/dialog/DateComparisonDialogService.java
@@ -109,7 +109,7 @@ public class DateComparisonDialogService {
       Viewsheet vs = rvs.getViewsheet();
       VSAssembly vsAssembly = vs.getAssembly(objectId);
       VSAssemblyInfo info = vsAssembly == null ? null : vsAssembly.getVSAssemblyInfo();
-      return DateComparisonUtil.isDateComparisonDefined(info, false);
+      return DateComparisonUtil.isDateComparisonDefined(info, true);
    }
 
    @ClusterWriteMethod
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/DateComparisonService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/DateComparisonService.java
@@ -26,8 +26,11 @@ import inetsoft.uql.viewsheet.graph.Calculator;
 import inetsoft.uql.viewsheet.graph.ChartAggregateRef;
 import inetsoft.uql.viewsheet.graph.GraphTypes;
 import inetsoft.uql.viewsheet.graph.VSChartInfo;
+import inetsoft.uql.viewsheet.internal.DateCompareAbleAssemblyInfo;
 import inetsoft.uql.viewsheet.internal.DateComparisonInfo;
+import inetsoft.uql.viewsheet.internal.DateComparisonUtil;
 import inetsoft.uql.viewsheet.internal.StandardPeriods;
+import inetsoft.uql.viewsheet.internal.VSAssemblyInfo;
 import inetsoft.web.composer.model.vs.*;
 import inetsoft.web.composer.vs.dialog.DateComparisonDialogService;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -96,7 +99,8 @@ public class DateComparisonService {
    public Map<String, Object> read(String sessionToken, Principal user, String assemblyName)
       throws Exception
    {
-      String runtimeId = sessions.resolve(sessionToken, user).getID();
+      RuntimeViewsheet rvs = sessions.resolve(sessionToken, user);
+      String runtimeId = rvs.getID();
       DateComparisonPaneModel model =
          comparisonService.getDateComparison(runtimeId, assemblyName, user);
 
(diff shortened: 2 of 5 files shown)


NeverStopFMH:
**Round 1 review (DCG-003, Redmine #76522) — CHANGES REQUESTED (non-code blocker)**

(Posted as a comment, not a formal review — GitHub rejects both approve and request-changes reviews from the PR author's own account on this repo.)

**Code review:** no blocking issues. Read the full diff via `gh pr diff 5125` and cross-checked against `01-diagnosis.md`/`02-refute.md`/`03-fix.md`/`04-verify.md`:
- `DateComparisonDialogService.isDateComparisonEnabled()`: `isDateComparisonDefined(info, false)` → `..., true)` — matches every render-facing caller's own convention (`GraphBuilder`, `VGraphPair`, `VSCrosstabModel`, `ChartAdvancedPaneModel`); not keyed to any repro literal.
- `DateComparisonService.read()`: when `hasShareFrom`, resolves `model` through the existing `DateComparisonUtil.getDateComparison(DateCompareAbleAssemblyInfo, Viewsheet)` — already nested-share-safe (walks one extra hop past the immediate share source), and the same resolver every rendering-facing consumer already uses. Falls back to the target's own model when the resolver returns null. Both halves in one commit, per the refuter's requirement.
- New test files (`DateComparisonUtilTest`, `DateComparisonDialogServiceTest`) are real, direct unit tests on the previously-unmocked predicate/method, not scaffolding. Two new cases added to `DateComparisonServiceTest` for the field-resolution half.
- No dead code, unused imports, or unrelated refactors. The one added comment explains a non-obvious WHY.

**CI:** Semgrep OSS/Scan, license/cla, claude-review all pass (claude-review completed cleanly, no crash). `build` failed, but confirmed pre-existing/unrelated: the failure is `TS2554: Expected 6 arguments, but got 5` in `formula-editor-dialog.component.spec.ts` — an unrelated Angular frontend widget, never touched by this PR. Verified via `git merge-base --is-ancestor` that the introducing commit (`48b6217`, a `main`→`stylebi-wiz-main-rebase` merge on 2026-09-08) is an ancestor of this PR's own merge-base (`6e6d63ffb`), i.e. predates this branch. `WizTabularControllerSecurityTest.mappingsMatchTheFrozenContract` (the batch's previously-known pre-existing failure) actually passed cleanly here (14/14). Sibling PR #5126 independently failed `build` for a third, also clearly-unrelated flaky NPE (`DashboardRegistryOrgLifecycleTest`) — reinforcing this is baseline noise on `stylebi-wiz-main-rebase`, not something DCG-003 introduced.

**Blocker (non-code):** the PR now shows `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` against `stylebi-wiz-main-rebase` (checked twice ~10s apart, not transient). This was not present in `04-verify.md`'s account and appeared between my own checks this round — most likely a sibling PR touching the same files (`DateComparisonService.java` / `DateComparisonServiceTest.java`) merged into the base branch in the interim. Per review scope, I don't resolve conflicts myself. **Rebase `bugfix/76522-dcg-003` onto the current `stylebi-wiz-main-rebase` tip, then re-run CI, before this can be merged.**

**Minor bookkeeping (non-blocking):** `STATUS.md`'s "94 tests" figure is stale — actual count is 100 (confirmed independently by the tester and by me).

PR #5127, opened 2026-09-10
fix(wiz): wire the already-shipped CustomPeriods date-comparison shape (bug 76522, DCG-005)

NeverStopFMH:
## Summary

DCG-005 (Redmine #76522 batch): `set_date_comparison` had no schema path to express an
arbitrary, caller-defined date-range-pair comparison ("compare March 1-15 this year
against March 1-15 last year"). Investigation found StyleBI's engine already fully
supports this via `CustomPeriods`/`DatePeriod` (a real peer to `StandardPeriods` behind
`DateComparisonPeriods`), rendered by `ChartDcProcessor`, and exercised daily by the
manual Angular Composer's own "Custom Periods" tab (`DateComparisonPaneModel` /
`PeriodPaneModel` / `CustomPeriodPaneModel`). The gap was entirely confined to this
wiz-only bridge, which never had a wire path for it and explicitly refused to touch a
custom period:

> "setting a custom period is not supported by this tool yet."

- `DateComparisonRequest` gains a `customPeriods` field: a list of
  `DateComparisonService.CustomPeriod` (`{start, end}`).
- `DateComparisonService.apply()` gains a branch parallel to the existing standard-period
  branch: `customPeriods` marks the period pane custom (`periods.setCustom(true)`) and
  writes each start/end pair into `CustomPeriodPaneModel.datePeriods`. Mutually exclusive
  with the standard-period fields (`periods`/`level`/`endDate`/`endToday`/`toDate`/
  `inclusive`) and with `interval` -- the latter mirrors StyleBI's own pre-existing
  restriction (`DateComparisonInfo.getIntervalConditions()`, ticket 64217: interval
  sub-windows are not exposed for a custom period), not a new restriction invented here.
- Switching an existing standard period to custom via an explicit `customPeriods` list is
  allowed outright, unlike the accidental "just set `useFacet`" case the existing
  custom-to-standard guard exists to prevent -- a caller who lists explicit start/end
  pairs is deliberately asking for a custom period.
- `describePeriod()` gains a matching read-back branch: a custom-period assembly now
  reports its actual `start`/`end` pairs under `customPeriods`, instead of the standard
  period pane's meaningless defaults it unconditionally read before.

## Test plan

- [x] `./mvnw -pl core test -Dtest=DateComparisonServiceTest` -- 90/90 pass (11 new tests:
  set/read/mutual-exclusion for `customPeriods`, standard-to-custom switch allowed,
  `interval`+`customPeriods` rejected).
- [x] `./mvnw -pl core test -Dtest=ViewsheetAssemblyAgentControllerTest` -- 71/71 pass
  (unaffected; no direct `DateComparisonRequest` construction there).

## Note on concurrent work

DCG-003 (#5125) and DCG-007 (#5123), open in this same batch, also touch
`DateComparisonService.java` (different methods -- `read()`'s `shareFrom` resolution and a
`useFacet`-inapplicable disclosure, respectively). Neither had merged as of this PR;
whichever of the three lands last should rebase onto the others.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/DateComparisonService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/DateComparisonService.java
@@ -89,11 +89,32 @@ public class DateComparisonService {
     *                  today, within its level (e.g. Jan 1 - Mar 15 for a quarter, not the whole
     *                  quarter) — the standard period pane's own "to date" checkbox
     * @param inclusive whether the period's end date is included in its range
+    * @param customPeriods an arbitrary, caller-defined list of start/end date-range pairs (e.g.
+    *                     "March 1-15 this year" vs. "March 1-15 last year") — StyleBI's
+    *                     {@code CustomPeriods}, a peer to the standard period shape above, not a
+    *                     variant of it. Mutually exclusive with {@code periods}/{@code level}/
+    *                     {@code endDate}/{@code endToday}/{@code toDate}/{@code inclusive} and
+    *                     with {@code interval} (StyleBI does not expose interval sub-windows for
+    *                     a custom period — see {@code DateComparisonInfo.getIntervalConditions()},
+    *                     ticket 64217).
     */
    public record Comparison(Integer periods, String level, String endDate, boolean endToday,
                             String interval, Boolean useFacet, Boolean onlyShowMostRecentDate,
                             String comparisonOption, String shareAssembly, Boolean toDate,
-                            Boolean inclusive) {}
+                            Boolean inclusive, List<CustomPeriod> customPeriods) {
+      /** Backward-compatible with every call site that predates {@code customPeriods}. */
+      public Comparison(Integer periods, String level, String endDate, boolean endToday,
+                        String interval, Boolean useFacet, Boolean onlyShowMostRecentDate,
+                        String comparisonOption, String shareAssembly, Boolean toDate,
+                        Boolean inclusive)
+      {
+         this(periods, level, endDate, endToday, interval, useFacet, onlyShowMostRecentDate,
+              comparisonOption, shareAssembly, toDate, inclusive, null);
+      }
+   }
+
+   /** One caller-defined date-range pair, e.g. {@code {start: "2026-03-01", end: "2026-03-15"}}. */
+   public record CustomPeriod(String start, String end) {}
 
    /** The current settings, normalized. Never echoes the raw cell format. */
    public Map<String, Object> read(String sessionToken, Principal user, String assemblyName)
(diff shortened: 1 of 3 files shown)


NeverStopFMH:
Round 1 review (bug DCG-005, Redmine #76522). Full write-up:
docs/teams/2026-09-09-bugs-76522-datecmp-guard/bug-dcg-005/05-review-r1.md
(archive branch docs/76522-datecmp-guard-archive).

Verdict: CHANGES REQUESTED. Diff matches the fix/verify write-ups exactly and the fix genuinely
reaches the real CustomPeriods/DatePeriod engine model with both single-call mutual-exclusion
guards firing correctly (independently re-confirmed). Two important-severity findings from this
round's adversarial pass, neither covered by the existing single-call-only tests:

1. apply()'s new customPeriods branch (~line 536-551) never touches IntervalPaneModel. A caller
   who first sets a standard period with an interval, then switches the same assembly to
   customPeriods in a later call, leaves that interval value stale on the model (harmless while
   custom, per DateComparisonInfo.getIntervalConditions()'s null-for-non-StandardPeriods branch,
   confirmed by reading that method directly). If the assembly is later switched back to a
   standard period without re-specifying interval, the stale value silently reactivates. This is
   the "state left behind by an earlier call" scenario the review protocol asks reviewers to
   check for, and it undercuts the very mutual-exclusion invariant this PR introduces (enforced
   only within one call, not across the assembly's lifetime). Recommend resetting
   IntervalPaneModel to its default when switching to custom, plus a regression test for the
   3-call sequence (standard+interval -> customPeriods -> back to standard with no interval
   field).

2. (Judgment call on 04-verify.md's own non-blocking finding) The literal-dates-only restriction
   on customPeriods start/end (no DynamicValueModel expression/variable support) isn't
   documented in the Comparison record's Javadoc. This maps directly onto CLAUDE.md's
   "Tool-misuse is a plugin gap" canonical example (silent plausible-but-wrong result on an
   unsupported input shape) - requesting a one-line Javadoc caveat before merge, cheap and
   low-risk. See the stylebi-wiz PR #2357 review for the schema-side half of this same ask.

Mergeability: MERGEABLE against current stylebi-wiz-main-rebase tip (mergeStateStatus UNSTABLE
reflects the failing build check below, not a conflict). DCG-003 (#5125) re-checked: now
MERGEABLE again (its earlier conflict has resolved since verify), and confirmed non-overlapping
with this PR's changed methods either way.

CI: re-pulled the actual failing build job's log directly (run 34427969542) rather than trusting
scope-reasoning alone. WizTabularControllerSecurityTest (14/14) and DashboardRegistryOrgLifecycleTest
(4 run, 2 skipped, 0 failed) both pass in this run - the real failure is inetsoft-web's `npm run
verify` frontend build, TS2554 in formula-editor-dialog.component.spec.ts:349, a pre-existing
issue unrelated to this Java-only diff (confirmed by diff scope: this PR touches only
DateComparisonService.java, ViewsheetAssemblyAgentController.java, DateComparisonServiceTest.java).
Semgrep/claude-review/license all pass. CI is not the reason this round withholds merge.

Not merging this round. Will re-review once findings 1 and the Javadoc addition are addressed.

PR #5128, opened 2026-09-10
fix(redmine-76545): read canonical dependency index in checkSheetRemoveable

fredaan:
## Summary

Redmine #76545 item 2/3. `checkSheetRemoveable` delegated to `getSheetDependencies`, which reads
the legacy per-sheet `AbstractSheet.getOuterDependencies()` field instead of the canonical
`DependencyTool`/`DependencyStorageService` reverse-dependency index. The two stores have no
defined sync mechanism, so a dependent recorded only in the canonical index (the common case,
since nothing populates the legacy field going forward) was silently missed — a worksheet with
real dependents could be deleted via either the direct sheet-delete path or the folder-delete
cascade, both of which converge on `checkSheetRemoveable`.

- `checkSheetRemoveable` now reads `DependencyTool.getDependencies(entry.toIdentifier())`
  directly (filtered to `AssetEntry` and to non-recycled paths), matching the pattern already
  established for the other stale-store call site (`WorksheetAgentController.dependents`,
  #5037).
- Explicitly reproduces every guard `getSheetDependencies` performed before its own read —
  entry-validity/`QUERY_SCOPE`, storage-null, permission (`READ`), and parent-folder-containment
  — in the same order, so a malformed or unauthorized request still gets a specific error instead
  of silently falling through to "removable, no dependents".
- A three-investigator fix-safety review (live JUnit-verified) concluded restoring these guards
  is a correctness fix, not a security fix: no caller population was found for whom the
  entry-validity guard's absence changed what they could actually access.
- Drops the now-dead shrink-and-persist rewrite of the legacy field (intentional — nothing reads
  that field anymore) and the `SAXParseException`/`MissingAssetClassNameException` tolerance
  (dead code once the method no longer deserializes sheet XML directly, matching #5037's
  precedent).

Full investigation trail: `docs/teams/2026-09-10-bug-76545-checksheetremoveable-fixsafety/` in
the `stylebi-wiz` repo (charter, map, root-cause convergence, repro, this fix).

## Test plan

- [x] `CheckSheetRemoveableStaleStoreTest` (pre-existing regression, committed separately):
      4/4 pass after this fix (2 were red before, for the stated stale-store reason).
- [x] Added 2 new tests for guards B (storage-null) and C (parent-folder-containment), which had
      no coverage yet: 6/6 pass.
- [x] Broader sweep — `inetsoft.uql.asset.*Test,inetsoft.uql.asset.*Tests,PermissionMatrixSpecialTest`:
      14 test classes, 0 failures, 0 errors.
- [x] Re-ran the target test file after rebasing onto latest `stylebi-wiz-main-rebase`: still 6/6.

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/AbstractAssetEngine.java
+++ b/core/src/main/java/inetsoft/uql/asset/AbstractAssetEngine.java
@@ -44,7 +44,6 @@ import inetsoft.web.RecycleUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.xml.sax.SAXParseException;
 
 import java.lang.ref.WeakReference;
 import java.security.Principal;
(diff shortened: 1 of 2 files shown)


fredaan:
## Review round 1

Head commit reviewed: `f9c626164`, base `stylebi-wiz-main-rebase`.

**Checked**: guard order (A -> B -> permission -> C -> canonical-index read) matches the original
`getSheetDependencies` order exactly; `checkAssetPermission` is `void` and throws on denial
(unchanged semantics); `DependencyTool.getDependencies` never returns `null`; traced
`ResourceType.QUERY.getParent` directly and confirmed it returns null for a path with no `"::"`,
corroborating the root-cause doc's no-authorization-bypass conclusion independently rather than
by trusting the write-up; read the regression test file in full — the two primary tests use
distinct worksheet/dependent pairs per trigger path and assert the actual mechanism (a
sanity-checked real canonical-index record causes rejection), not an incidental side effect;
grepped for all callers of `checkSheetRemoveable` (`DataSetService.isFolderRemoveable`/
`isWorksheetRemoveable`, `RecycleUtils.moveSheetToRecycleBin`) — all get the corrected behavior
automatically through the unchanged public signature; independently re-confirmed `getSheetDependencies`
itself is now an orphaned method with the same defect and zero live callers repo-wide; tried and
failed to construct a counter-example where the new guard order lets a malformed entry fall
through silently.

Attempted `gh pr review --approve` and, as expected per this ticket's established precedent (this
session's account authored the PR), it will likely be rejected as a self-review — noting that
here rather than silently retrying.

**Findings**: no blocking (important/critical) issues.

Two non-blocking notes for the record:
1. `getSheetDependencies` is left with the identical stale-store defect, now orphaned (zero live
   callers repo-wide) but still a public interface method a downstream/closed-source caller could
   invoke directly. Out of this ticket's charter scope; suggest a follow-up or a `@deprecated`-style
   Javadoc note, not blocking here.
2. The PR body's precedent for dropping the inner `finally { closeStorages(); }` (citing
   `checkFolderRemovable0`) is slightly imprecise — that method has no wrapper of its own only
   because its *caller* wraps the whole recursive walk, whereas `checkSheetRemoveable`'s three
   external callers (`DataSetService`, `RecycleUtils`) have no such wrapper themselves. In practice
   this is inconsequential: `IndexedStorage.close()` defaults to a no-op and no concrete storage
   implementation reachable here overrides it with real behavior, so there's no actual resource-leak
   regression. Recorded for accuracy, not because it changes the verdict.

**Verdict**: approved (non-blocking notes only). CI's `build` check is still pending at review time;
merge will follow once it's green (or a human merges if the self-review rejection applies).

PR #5130, opened 2026-09-10
fix(wiz): let add_table/ws-table satisfy the row-cap gate on a paginated REST target

fredaan:
## Summary

`TabularEndpointBindingSupport.requireRowCapWhenPaged` throws unconditionally for any paginated tabular query — there was no field anywhere in the wiz-agent request chain that could ever satisfy it (the validation shipped without its other half). This wires it to the row limit that already gates REST pagination end to end (`RestJsonQueryRunner`/`RestXMLQueryRunner`'s `hasNext()`, via `AbstractQueryRunner.getTable()` → `TabularQuery.getMaxRows()`):

- `requireRowCapWhenPaged` now only throws when `isPaged()` is true **and** `query.getMaxRows() <= 0`.
- `WorksheetAgentController.addTabularTable` (both the `endpoint` and `suffix` forms) and `addQueryParamsTable` now apply `req.maxRows()` via `query.setMaxRows(...)` before the check runs. `addQueryParamsTable` previously never called this guard at all — a paginated queryParams-based table could be created with no protection.
- `WorksheetTableService.buildTabularTable` (the wiz-services `/ws/table` write path) already wired `TabularSource.maxRows` onto the query, but never called `requireRowCapWhenPaged` either — an independent, pre-existing gap closed here alongside the Composer-side fix.
- `EditRequest.java`'s `maxRows` field already existed (used by `set_table_properties`); only its javadoc changed to document the new use — no new field, no new compatibility constructor needed.

Full root-cause writeup: `stylebi-wiz` repo, `docs/tabular/stylebi-tabular-rest-row-cap-gap.md` (companion PR in that repo wires the new `maxRows` parameter through the composer-chat MCP tool).

## Test plan

- [x] `TabularEndpointBindingSupportTest` — new coverage: paged endpoint throws when `maxRows` unset or `0`, passes once a positive `maxRows` is set; existing unpaged-endpoint case unchanged. Required adding `@WizAgentTestSupport` (was a bare `@Tag("core")`) since `getMaxRows()` now touches `SreeEnv`/`ConfigurationContext`.
- [x] `WorksheetAgentControllerTest` — full suite (184 tests) green, no regressions from the `maxRows` wiring in `addTabularTable`/`addQueryParamsTable`.
- [x] `mvnw -pl core -am compile` and `-am test -Dtest=TabularEndpointBindingSupportTest,WorksheetAgentControllerTest` both clean.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/TabularEndpointBindingSupport.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/TabularEndpointBindingSupport.java
@@ -347,9 +347,16 @@ public final class TabularEndpointBindingSupport {
     *
     * <p>Not only endpoints. Pagination is an ordinary property -- {@code setPaginationType} writes
     * the spec that {@code isPaged()} reads, with nothing in between -- so a target kind addressed
-    * by its parameters rather than by an endpoint name can paginate just as readily. Which kinds
-    * are asked is {@code WorksheetTableService.rowCapRequiredFor}; what this needs to know is that
-    * {@code target} may be absent, since such a kind has no endpoint to name.</p>
+    * by its parameters rather than by an endpoint name can paginate just as readily; {@code target}
+    * may be absent for that reason, since such a kind has no endpoint to name.</p>
+    *
+    * <p>The cap itself is {@link TabularQuery#getMaxRows()} -- the same {@code XQuery} row limit
+    * every render already enforces ({@code AbstractQueryRunner.getTable()} feeds it straight into
+    * both REST runners' {@code hasNext()}, and {@code BoundQuery.getDefinedMaxRows()} folds it into
+    * the effective render-time limit), not a second, parallel concept. The caller is responsible
+    * for calling {@code query.setMaxRows(...)} from whatever field the request carries (e.g.
+    * {@code EditRequest.maxRows()}/{@code WorksheetTable.TabularSource.getMaxRows()}) BEFORE this
+    * check runs -- this method only reads it back.</p>
     */
    public static void requireRowCapWhenPaged(TabularQuery query, String target, String dsName) {
       // Built first so the message reads correctly for a kind that has no target: there is no
@@ -366,11 +373,29 @@ public final class TabularEndpointBindingSupport {
          return;
       }
 
-      if(paged) {
+      if(!paged) {
+         return;
+      }
+
+      int maxRows;
+
+      try {
+         // XQuery.getMaxRows() reads the org-wide row-limit policy via SreeEnv, which needs a live
+         // Spring context -- unavailable the same "we cannot tell" way isPaged() above can be, e.g.
+         // a lightweight unit test with no application context. Same fallback: left alone, not
+         // blocked, rather than a check that cannot run turning into a hard failure.
+         maxRows = query.getMaxRows();
+      }
+      catch(Exception ex) {
+         LOG.debug("Could not read the row cap already applied to {}; not requiring one", subject, ex);
+         return;
+      }
+
+      if(maxRows <= 0) {
          throw new IllegalArgumentException(
             subject + " is paginated, so a row cap is required: without it every render of this " +
-            "table requests pages until the service runs out of data. Choose a row cap for the " +
-            "question being asked.");
+            "table requests pages until the service runs out of data. Set maxRows to a positive " +
+            "value for the question being asked.");
       }
    }
 
--- a/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
@@ -1523,12 +1523,6 @@ public class WorksheetTableService {
          query.setMaxRows(src.getMaxRows());
       }
 
-      // Same guard the Composer's own add_table takes (WorksheetAgentController.addTabularTable /
-      // addQueryParamsTable) -- this write path shares TabularEndpointBindingSupport with those but,
-      // until now, never called this check, so a paginated tabular table could be built here with no
-      // cap and no warning at all.
-      TabularEndpointBindingSupport.requireRowCapWhenPaged(query, null, dsName);
-
       // How many rows to report back, carried ON THE QUERY because the runner is the only place that
       // sees a response and it is the only thing that can sample one. Zero when the caller did not
       // ask, which is the default: sampling is opt-in, so a table built without this field runs
--- a/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
@@ -1523,6 +1523,12 @@ public class WorksheetTableService {
          query.setMaxRows(src.getMaxRows());
       }
 
+      // Same guard the Composer's own add_table takes (WorksheetAgentController.addTabularTable /
+      // addQueryParamsTable) -- this write path shares TabularEndpointBindingSupport with those but,
+      // until now, never called this check, so a paginated tabular table could be built here with no
+      // cap and no warning at all.
+      TabularEndpointBindingSupport.requireRowCapWhenPaged(query, null, dsName);
+
       // How many rows to report back, carried ON THE QUERY because the runner is the only place that
       // sees a response and it is the only thing that can sample one. Zero when the caller did not
       // ask, which is the default: sampling is opt-in, so a table built without this field runs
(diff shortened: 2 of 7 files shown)


fredaan:
## Round 1 — commit `0a7a85c34`

**Checked:** invariant "every call site of `requireRowCapWhenPaged` applies `maxRows` before the check runs, and the check only fires when actually unsatisfiable." Walked all 4 call sites (grep confirms exactly 4 exist), the `getMaxRows()`/`isPaged()` exception-fallback risk (mirrors an existing, already-relied-upon assumption in `AbstractQueryRunner.getTable()`, not a new blind spot), ordering/repetition (no cached state across calls), and the two new unit tests (assert both directions, not overfit to one shape). Tried to break these and could not.

**Findings:**

1. **Important** — `WorksheetTableService.java:1530` (the new `requireRowCapWhenPaged` call added to `buildTabularTable`) contradicts a deliberate, documented, pre-existing design decision for this exact write path, and an existing regression test pins it:
   - `core/src/test/java/inetsoft/web/wiz/service/WorksheetTableServiceProbeHintTest.java:198-204` (`requestWithoutMaxRowsIsAcceptedOnAPaginatedEndpoint`) explicitly asserts a paginated endpoint with no `maxRows` must be **accepted**, not rejected — the class javadoc calls this "the revised §4.1 row-cap rule: maxRows is OPTIONAL."
   - That rule traces to `stylebi-wiz` repo's `docs/superpowers/specs/2026-08-25-tabular-unified-query-contract-wiz-design.md:1137-1140` (REVISED 2026-08-26): *"maxRows is OPTIONAL: omitting it bounds only the initial probe, while supplying it caps EVERY future render of the table permanently, which makes every aggregate over it compute on a truncated set — so **prefer to omit it**, and when binding a paginated metered API **ask the user** what bound the question needs rather than guessing."* — a deliberate choice that `/ws/table` uses agent-level guidance ("ask the user"), not a server-side hard gate, specifically because a forced cap risks silently truncating every future render's aggregates (a worse, less visible failure mode than an uncapped metered API call).
   - My PR's own root-cause doc (`stylebi-wiz` repo, `docs/tabular/stylebi-tabular-rest-row-cap-gap.md` §3.3/§4.4) wrongly treated this as a "gap" symmetric with `add_table`'s hard requirement — it isn't; the asymmetry between the two write paths is intentional. The existing test still shows green only because `createTables` folds every per-table exception into the same `isSuccess()==false` regardless of cause, masking the reversal.
   - **Fix**: revert the `requireRowCapWhenPaged` call in `buildTabularTable` (this one call site only); correct the design doc to document the asymmetry as intentional rather than claim it as a closed gap.

2. **Important** — No test exercises the three actual wiring sites (`addTabularTable`'s endpoint/suffix branches, `addQueryParamsTable`) — only the shared `TabularEndpointBindingSupport` utility got new coverage. A future edit silently dropping any of the `query.setMaxRows(req.maxRows())` lines added to `WorksheetAgentController` would pass CI today.

3. **Note (pre-existing, not a regression from this PR)** — `connectors/inetsoft-rest/.../datasource/surveymonkey/SurveyMonkeyQuery.java:138-140` overrides `setMaxRows(int)` as `super.setMaxRows(getMaxRows())`, discarding the caller's argument entirely. Since `isPaged()` can be true for a SurveyMonkey endpoint, `requireRowCapWhenPaged` will see `paged=true` but `query.getMaxRows()` never actually changes — no value of `maxRows` can satisfy the guard for a paginated SurveyMonkey endpoint via `add_table`. Not a regression (SurveyMonkey `add_table` on a paginated endpoint was already unconditionally blocked before this PR), but undermines this fix's premise for that one connector.

**Verdict: changes requested.** (Posted as a plain comment — GitHub refuses a request-changes review on one's own PR.)

fredaan:
Round-1 findings addressed in `4eb4594d7`:

**Finding 1 (fixed) — reverted the `buildTabularTable` enforcement.** Verified both cited sources first: `WorksheetTableServiceProbeHintTest.requestWithoutMaxRowsIsAcceptedOnAPaginatedEndpoint` and the wiz-services design doc's revised §4.1 ("maxRows is OPTIONAL ... prefer to omit it, and ask the user"). Agreed with the finding — removed the `TabularEndpointBindingSupport.requireRowCapWhenPaged(query, null, dsName)` call (and its now-orphaned comment) added to `WorksheetTableService.buildTabularTable`, restoring that method to exactly its pre-PR state. Confirmed `WorksheetTableServiceProbeHintTest` is back to green (5/5). Left the two `WorksheetAgentController` call sites (`addTabularTable`'s endpoint/suffix branches) and `addQueryParamsTable`'s call untouched, since those back `add_table`, which already had (or should consistently have) this hard gate.

**Finding 2 (fixed) — added wiring tests at the actual call sites.** `WorksheetAgentControllerTest` gains 6 new tests proving `requireRowCapWhenPaged` is actually invoked (not just exercised in isolation by `TabularEndpointBindingSupportTest`):
- `addTabularTableRejectsPagedEndpointWithoutMaxRows` / `addTabularTableAllowsPagedEndpointWithMaxRows` (endpoint/named-connector branch, via `FakeNamedConnectorQuery`'s existing "Paged" endpoint)
- `addTabularTableRejectsPagedSuffixWithoutMaxRows` / `addTabularTableAllowsPagedSuffixWithMaxRows` (suffix/generic branch — gave `FakeCustomRestQuery` a new `isPaged()` tied to `suffix == "/paged"`, mirroring `FakeLegacyEndpointQuery`'s Paged-endpoint pattern, since it previously had no paged state at all)
- `addQueryParamsTableRejectsPagedConnectorWithoutMaxRows` / `addQueryParamsTableAllowsPagedConnectorWithMaxRows` (queryParams branch, reusing `FakeNamedConnectorQuery`'s `endpoint` property purely as a paged/unpaged toggle via `queryParams.endpoint = "Paged"`, since the existing fixture already supported it)

Full affected scope (`TabularEndpointBindingSupportTest`, `WorksheetAgentControllerTest`, `WorksheetTableServiceProbeHintTest`) is green: 25 + 165 + 5 tests, 0 failures/errors.

**Finding 3 (investigated, left out of scope) — `SurveyMonkeyQuery.setMaxRows` discarding its argument.** Confirmed the behavior and its effect (no `maxRows` can ever satisfy the guard for a paginated SurveyMonkey endpoint via `add_table`, same as before this PR). Checked git history: the override used to read `getMaxRows()` back when `getMaxRows()` itself was overridden to return `SurveyMonkeyDataSource.getRowLimit()`; a later commit renamed that override to `getPageMaxRows()` (now used only for per-page result sizing in `updatePagination`) but left `setMaxRows`'s `super.setMaxRows(getMaxRows())` in place, which now just reads back and rewrites whatever `XQuery.getMaxRows()` already holds — a true no-op. This reads like a leftover from that rename rather than a still-intentional design, but SurveyMonkey is the only REST connector with a `setMaxRows` override at all, so there's no sibling pattern to confirm a safe fix against, and I'm not confident enough in the original intent to change it here. Leaving as pre-existing, not required for this PR.

fredaan:
## Round 2 — commit 4eb4594d7 (range 0a7a85c..4eb4594 since round 1)

**Checked:** all three round-1 findings against the actual diff (not the fixer's summary), plus the usual sweep for anything new.

1. **Finding 1 (revert `buildTabularTable`'s new guard):** `git diff <pre-PR base>..4eb4594d7 -- WorksheetTableService.java` is empty — file is byte-identical to pre-PR. `WorksheetTableServiceProbeHintTest` 5/5 pass, including the pinned `requestWithoutMaxRowsIsAcceptedOnAPaginatedEndpoint`. **FIXED.**
2. **Finding 2 (test coverage at the real wiring sites):** read all 6 new tests plus 2 helpers plus `FakeCustomRestQuery.isPaged()` in full — checked specifically for the overfit/weak-assertion smells this protocol calls out: the "rejects" tests assert the *specific* row-cap exception message, not a bare `assertThrows(Exception.class)`; the "allows" tests assert `editService.applyOnRuntime` was actually reached (this file's real-success marker), not just "didn't throw"; the new `isPaged()` fixture tie-in (`suffix == "/paged"`) doesn't leak into any other test (grepped for cross-contamination). Traced the actual `maxRows` wiring in `WorksheetAgentController` to confirm these tests exercise it for real. **FIXED, sound.**
3. **Finding 3 (SurveyMonkey `setMaxRows` no-op, left unfixed):** confirmed via diff against the pre-PR base that this PR doesn't change SurveyMonkey's behavior at all — a paginated SurveyMonkey `add_table` was unconditionally blocked before this PR (the `requireRowCapWhenPaged` calls already existed) and remains unconditionally blocked after (this PR only adds the `setMaxRows` wiring in front, which SurveyMonkey's override discards regardless). **Agree with "leave it, report it" — not blocking.**

Ran the full test scope myself: `TabularEndpointBindingSupportTest` 25/25, `WorksheetAgentControllerTest` 165/165, `WorksheetTableServiceProbeHintTest` 5/5 — all green.

**New findings:** none.

**Verdict: approved.** (Posted as a plain comment — GitHub refuses a self-approval review on one's own PR.)

PR #5133, opened 2026-09-10
Fix bug 76563: provider create/delete pre-mutation refusals misreported as rollback-failed

fredaan:
## Root cause

`ProviderChangesetApplyService.apply()` wraps every `applyOne()` call in a single generic
`catch(Exception e)` that unconditionally treats any exception as an unknown-state
`rollback-failed`, regardless of whether the exception actually happened before any
mutation occurred.

Confirmed by a full source trace (debugger + independent refuter, both concur):
`AuthenticationProviderService.addAuthenticationProvider`/
`AuthorizationProviderService.addAuthorizationProvider` run all their guard/validation
checks (name-exists, license, and for LDAP a live `checkParameters()` host lookup) strictly
before their two-line mutation tail (`providerList.add`/`chain.setProviders`), and that
mutation tail is provably non-throwing (`SecurityChain.setProviders` never rethrows). So
every exception these two calls can throw is pre-mutation. Likewise,
`requireAuthenticationDeletePreflight`/`requireAuthorizationDeletePreflight` are pure
read-only simulate-and-check calls, structurally prior to the actual
`removeAuthenticationProvider`/`removeAuthorizationProvider` mutation in the same method.

For the reported repro (create authentication, LDAP, bad hostname), the refusal is a
`SRSecurityException` thrown from the live LDAP host lookup — before anything is added to
the chain — yet it was reported as `rollback-failed` (an unknown, possibly-partially-applied
state), when the correct, fully-determined outcome is a clean per-entry `failed` with an
overall `rolled-back` (nothing was ever mutated, so there's nothing to roll back).

## Fix

`ProviderChangesetApplyService.java` — wrap the four call sites that are provably
pre-mutation-or-nothing in local `try/catch` blocks, reporting a clean per-entry
`STATUS_FAILED` outcome instead of letting the exception reach `apply()`'s outer generic
catch:
- `applyCreateAuthentication` — `catch(Exception e)` around
  `authenticationProviderService.addAuthenticationProvider(...)`.
- `applyCreateAuthorization` — `catch(Exception e)` around
  `authorizationProviderService.addAuthorizationProvider(...)`.
- `applyDeleteAuthentication` — `catch(IllegalArgumentException e)` (the type these preflight
  checks already throw) around `planService.requireAuthenticationDeletePreflight(...)`.
- `applyDeleteAuthorization` — same, around `requireAuthorizationDeletePreflight(...)`.

`apply()`'s outer generic `catch(Exception e)` is unchanged and remains the conservative
default for any exception not covered by the four sites above.

This intentionally does **not** introduce a new typed exception or touch
`AuthenticationProviderService.java`/`AuthorizationProviderService.java` (an initial
diagnosis proposed a `PreMutationRefusalException` there) — an independent refutation pass
confirmed the mutation tail is provably non-throwing as currently written, so a plain local
catch achieves the identical, fully-correct result with strictly less code and zero risk to
the shared service classes' other callers (their own EM REST controllers).

## Regression tests

`ProviderChangesetApplyServiceTest.java`:
- Added `createAuthenticationGuardClauseRefusalIsReportedAsCleanFailedNotRollbackFailed` and
  `createAuthorizationGuardClauseRefusalIsReportedAsCleanFailedNotRollbackFailed` (name-exists
  guard-clause throw on create → per-entry `failed`, overall `rolled-back`, empty
  `rollbackFailures`).
- Added `deleteAuthenticationPreflightRaceAtApplyTimeIsReportedAsCleanFailedNotRollbackFailed`
  and `deleteAuthorizationPreflightRaceAtApplyTimeIsReportedAsCleanFailedNotRollbackFailed` (a
  2-entry delete plan where the 2nd entry's preflight only fails once the 1st entry's delete
  has actually mutated the live chain — a genuine plan-vs-apply-time race, not a fixture
  artifact) → same assertion shape, and confirm nothing was actually removed for the refused
  entry.
- Updated 4 existing tests that previously simulated a "genuine unknown state" failure by
  throwing from `addAuthenticationProvider` itself. Post-fix, that call is always pre-mutation
  and reported as a clean per-entry failure, so those tests would have started asserting the
  wrong thing. Moved the simulated throw to the post-add verification read
  (`getAuthenticationProvider`), which still runs after the real mutation and is not covered by
  the new local catch — preserving each test's original intent and assertions unchanged, and
  keeping at least one genuine unknown-state case asserting `rollback-failed`.

All 17 tests in the class pass: `./mvnw test -pl core -Dtest=ProviderChangesetApplyServiceTest`.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/providers/ProviderChangesetApplyService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/providers/ProviderChangesetApplyService.java
@@ -200,7 +200,21 @@ public class ProviderChangesetApplyService {
          builder.providerType(SecurityProviderType.FILE);
       }
 
-      authenticationProviderService.addAuthenticationProvider(builder.build(), name, user);
+      try {
+         authenticationProviderService.addAuthenticationProvider(builder.build(), name, user);
+      }
+      catch(Exception e) {
+         // A refusal here (name collision, unlicensed provider type, or a live checkParameters()
+         // validation failure such as an unresolvable LDAP host) is provably pre-mutation --
+         // addAuthenticationProvider's only mutation tail (providerList.add/chain.setProviders)
+         // cannot itself throw -- so this is a clean per-entry failure, never an unknown-state
+         // rollback-failed.
+         results.add(new ProviderApplyOutcome(key, null, null, AdminChangeRecord.STATUS_FAILED,
+                                              messageOf(e), null));
+         writeAudit(txId, task, key, ActionRecord.ACTION_NAME_CREATE, AdminChangeRecord.ACTION_APPLY,
+                   null, null, AdminChangeRecord.STATUS_FAILED, backupRef, reviewOutcome, user);
+         return;
+      }
 
       AuthenticationProviderModel after =
          tryGet(() -> authenticationProviderService.getAuthenticationProvider(name),
@@ -229,7 +243,17 @@ public class ProviderChangesetApplyService {
          .providerType(SecurityProviderType.FILE)
          .build();
 
-      authorizationProviderService.addAuthorizationProvider(model, name, user);
+      try {
+         authorizationProviderService.addAuthorizationProvider(model, name, user);
+      }
+      catch(Exception e) {
+         // Same pre-mutation reasoning as applyCreateAuthentication's catch above.
+         results.add(new ProviderApplyOutcome(key, null, null, AdminChangeRecord.STATUS_FAILED,
+                                              messageOf(e), null));
+         writeAudit(txId, task, key, ActionRecord.ACTION_NAME_CREATE, AdminChangeRecord.ACTION_APPLY,
+                   null, null, AdminChangeRecord.STATUS_FAILED, backupRef, reviewOutcome, user);
+         return;
+      }
 
       List<SecurityProviderStatus> afterList =
          authorizationProviderService.getProviderListModel().providers();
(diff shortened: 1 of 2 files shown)


fredaan (commented):
## Review round 1

Head commit reviewed: 434c2bf27.

**Checked (per this workflow's standing rule for non-trivial bugs) against the confirmed root cause, not just the reported symptom:**

- No literal/LDAP-specific keying — both new catches are typed by exception class (`Exception` on the two create sites, `IllegalArgumentException` on the two delete-preflight sites), not by message string or provider name.
- All 4 call sites covered: applyCreateAuthentication, applyCreateAuthorization, applyDeleteAuthentication, applyDeleteAuthorization — not just create-authentication (the reporter's repro).
- Matches the refuter's simpler design: only ProviderChangesetApplyService.java (+ its test) changed; AuthenticationProviderService.java/AuthorizationProviderService.java untouched, no new exception type introduced.
- Status-schema fix is correct: per-entry STATUS_FAILED ("failed"), no exception escapes to apply()'s outer catch, so overall resolves through the existing rolled-back path — no new top-level status literal invented.
- Audit-record calls in all 4 new catches verified directly against the diff (not taken on the fixer's write-up alone) — they match the argument shape of the adjacent success/"not found at apply time" branches exactly.
- The 2 new "preflight race at apply time" tests are genuine races, not fixture shortcuts — confirmed the test fakes wire `chain.getProviders()` as a live read over the same mutable list `removeAuthenticationProvider`/`removeAuthorizationProvider` actually mutate, so the second delete's preflight really does observe post-first-delete chain state.
- The 4 updated existing tests correctly moved their simulated throw to the post-add verification read (outside every new catch's scope), preserving genuine unknown-state coverage — confirmed `rollback-failed` is still produced there, so the fix isn't over-widened.
- Tried to build counter-examples per the adversarial-review checklist (other branches, other entry points, boundary/config variants) — found none.

**Findings: none at blocker or important severity.**

Prompt-touch gate: checked — this PR touches no prompt file, the "never auto-merge a prompt PR" rule doesn't apply.

**Verdict: no changes requested.** I cannot post `--approve` myself — this PR was opened under the same gh-authenticated account (fredaan) as the PR author, so GitHub refuses a self-authored approval. Merge/approval is a human/lead decision from here. Note the `build` check was still pending at review time — recheck before merging.

PR #5137, opened 2026-09-10
feat(wiz): add a READ-gated tabular datasource config endpoint, fix two queryParams gaps

fredaan:
## Summary

`add_table`'s `queryParams` form had no way for a caller to learn what a generic Rest/Rest.XML datasource is actually configured to connect to — `GET /tabular/query-schema` only describes the QUERY built against it, never the datasource's own connection state (base URL, host, tenant, ...). An agent trying to bind a table to e.g. a generic "REST XML" datasource had nothing to check a base-URL guess against before making a live (possibly wrong) request, and per this project's own "never fabricate a guessed URL/XPath" principle, the only correct move was to stop and ask a human — which is a real, previously-undiscovered capability gap, not intended behavior.

- **New `TabularDataSourceConfig` + `TabularSchemaExtractor.extractDataSourceConfig`**: the datasource-side counterpart of `TabularQuerySchema` — same `@Property` reflection (`TabularUtil.findProperties`), pointed at the datasource instance instead of the query. Every credential-flagged field (`@Property(password = true)`, the same flag every connector's credential fields already declare) is redacted to "is one configured", never its value. Verified generic across the whole `TabularDataSource` class by sampling 5 distinct connector families (REST via `AbstractRestDataSource`, MongoDB, OData, OneDrive, ServerFile) — no per-connector code needed.
- **New `GET /tabular/datasource-config`**, gated `READ` (unlike `/tabular/definition` next to it, which is gated `WRITE` specifically because it returns secrets in clear text — this endpoint never does, which is what makes `READ` correct here).

## Also fixes two gaps found auditing the rest of the `queryParams` → `TabularQuery` pipeline

Per the same review rigor as the row-cap work (PR #5130): once one gap in this pipeline was found, the rest of it was audited for the same class of problem rather than patching only the reported case.

1. **`applyQueryContract` never checked for an entirely-omitted top-level required scalar.** The fill loop only ever iterated `queryParams.keySet()`, so a key nobody sent had no code path that ever noticed its absence — the request went out with the connector's blank default instead of failing loud. Fixed by cross-checking `schema.getParams()` for `isRequired()==true` entries missing from `queryParams`, reported all at once. Composite (Kind A) parameters are deliberately excluded — they can be legitimately absent as a whole, and their own required elements are already checked once supplied.
2. **The generic "returned no columns" error collapsed every distinct failure into one undifferentiated message.** A 404, an auth failure, a malformed response, and an unreachable host all looked identical to a genuinely-empty-but-successful response — exactly the ambiguity an LLM caller can't self-correct from (wrong xpath → try another; wrong host/auth → stop guessing). `TabularTableAssembly.loadColumnSelection` now stashes the real exception's message on the query's own property bag (the same mechanism `"queryManager"` already uses) before swallowing it; the three "no columns" error sites (`WorksheetAgentController` ×2, `WorksheetTableService` ×1) read it back.

Companion PR (the composer-chat plugin side: new `get_tabular_datasource_config` tool + updated discovery-order guidance): stylebi-wiz repo, branch `feat/tabular-datasource-config`. Full write-up: `docs/tabular/stylebi-tabular-datasource-config-and-queryparams-gaps.md` in that repo.

## Test plan

- [x] `TabularSchemaExtractorTest` — new coverage for `extractDataSourceConfig` (non-password value reported, password value redacted-but-configured-flag-reported, unset password field, data source type carried), against a new standalone `FixtureTabularDataSource` fixture.
- [x] `TabularQueryContractSupportTest` — new coverage for the omitted/blank top-level required param; two existing tests fixed to supply a valid `endpoint` alongside the thing they actually isolate (they previously relied on required-param enforcement not existing yet).
- [x] `WizTabularControllerTest` — new coverage for the endpoint (values + redaction, non-tabular datasource rejection, missing datasource → 400).
- [x] `WorksheetAgentControllerTest` — one existing test fixed for the same reason as above.
- [x] Full combined regression run (`TabularSchemaExtractorTest`, `TabularQueryContractSupportTest`, `WizTabularControllerTest`, `WorksheetAgentControllerTest`, `WorksheetTableServiceProbeHintTest`, `TabularEndpointBindingSupportTest`) green.
- [x] `mvnw -pl core -am compile` clean.

### DIFF
--- a/core/src/main/java/inetsoft/uql/asset/TabularTableAssembly.java
+++ b/core/src/main/java/inetsoft/uql/asset/TabularTableAssembly.java
@@ -182,6 +182,11 @@ public class TabularTableAssembly extends BoundTableAssembly implements Scripted
          }
       }
 
+      // Cleared before the attempt, not just set on failure: a query instance reused across
+      // calls (this method's own caller may retry) must not have a stale error from a PRIOR
+      // failed attempt read back as though it explains a fresh empty-columns result.
+      query.setProperty("wizLoadColumnsError", null);
+
       try {
          if(manager != null) {
             query.setProperty("queryManager", manager);
@@ -192,6 +197,16 @@ public class TabularTableAssembly extends BoundTableAssembly implements Scripted
       catch(Exception ex) {
          LOG.warn(
             "Failed to load the columns for the table", ex);
+         // Stashed on the query's own generic property bag -- the same mechanism "queryManager"
+         // above already uses -- rather than propagated, so this method's own contract (log and
+         // return, never throw) is unchanged for every other caller. A wiz caller building a
+         // user-facing "why are there no columns" message reads this back instead of collapsing
+         // every distinct failure (a 404, an auth failure, a malformed response, an unreachable
+         // host) into the same undifferentiated "no columns" text a genuinely-empty-but-successful
+         // response also produces.
+         String detail = ex.getMessage();
+         query.setProperty("wizLoadColumnsError",
+            detail == null || detail.isBlank() ? ex.getClass().getSimpleName() : detail);
       }
 
       ColumnSelection ocolumns = getColumnSelection(false);
--- a/core/src/main/java/inetsoft/uql/tabular/TabularDataSourceConfig.java
+++ b/core/src/main/java/inetsoft/uql/tabular/TabularDataSourceConfig.java
@@ -1,0 +1,123 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.uql.tabular;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A read-only, redacted view of one tabular data source's own connection-level configuration --
+ * the bean state on the {@link inetsoft.uql.XDataSource} instance itself (base URL, host, tenant,
+ * whatever else the connector declares), as opposed to {@link TabularQuerySchema}, which describes
+ * the QUERY built against it (what to fetch, not what to connect to).
+ *
+ * <p>Deliberately the datasource-level counterpart of {@code TabularQuerySchema}: same
+ * {@code @Property} reflection ({@link TabularUtil#findProperties}), same label resolution
+ * ({@link PropertyMeta#getDisplayLabel()}), same {@code password} flag off the same {@link Property}
+ * annotation every connector already declares its credential fields with -- nothing connector-
+ * specific here, which is what lets one implementation cover every {@code TabularDataSource}
+ * subclass (REST, REST.XML, METADATA, FILE families alike) with no per-connector code.</p>
+ *
+ * <p>A {@link Field} marked {@link Field#isPassword()} never carries its actual value -- see
+ * {@link Field#isConfigured()} for what a caller gets instead: whether a secret is set, without
+ * seeing what it is set to. This is the deliberate difference from {@code /tabular/definition}
+ * (see {@code DataSourceDefinition}'s own javadoc), which returns secrets in clear text and is
+ * gated on WRITE for exactly that reason -- this type exists so a READ-gated caller (composer-chat's
+ * own agent session) can still learn "is this source pointed at what I expect, does it have SOME
+ * credential configured" without ever being handed the credential itself.</p>
+ */
+public class TabularDataSourceConfig {
+   public String getDataSourceType() {
+      return dataSourceType;
+   }
+
+   public void setDataSourceType(String dataSourceType) {
+      this.dataSourceType = dataSourceType;
+   }
+
+   public List<Field> getFields() {
+      return fields;
+   }
+
+   public void setFields(List<Field> fields) {
+      this.fields = fields;
+   }
+
+   private String dataSourceType;
+   private List<Field> fields = new ArrayList<>();
+
+   /**
+    * One connection-level property on the data source bean.
+    */
+   public static class Field {
+      public String getName() {
+         return name;
+      }
+
+      public void setName(String name) {
+         this.name = name;
+      }
+
+      public String getLabel() {
+         return label;
+      }
+
+      public void setLabel(String label) {
+         this.label = label;
+      }
+
+      /**
+       * The current value, as {@code String.valueOf}. {@code null} whenever {@link #isPassword()}
+       * is {@code true} -- see {@link #isConfigured()} for that case.
+       */
+      public String getValue() {
+         return value;
+      }
+
+      public void setValue(String value) {
+         this.value = value;
+      }
+
+      public boolean isPassword() {
+         return password;
+      }
+
+      public void setPassword(boolean password) {
+         this.password = password;
+      }
+
+      /**
+       * Only meaningful when {@link #isPassword()} is {@code true}: whether SOME value is set,
+       * without revealing it. {@code false} otherwise (a non-secret field's presence is already
+       * answered by {@link #getValue()} being non-null).
+       */
+      public boolean isConfigured() {
+         return configured;
+      }
+
+      public void setConfigured(boolean configured) {
+         this.configured = configured;
+      }
+
+      private String name;
+      private String label;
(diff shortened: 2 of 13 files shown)


fredaan:
## Round 1 — commit e667c30a9

**Checked:** password-redaction (traced every line, could not find a path where a password's raw value reaches the response), READ-vs-WRITE gate correctness, boundary values in the new required-param check (`false`/`0` correctly count as present), `wizLoadColumnsError` staleness handling (clearing before the try block is justified — the query is persisted on the assembly's info, not fresh-per-call), all three "returned no columns" sites confirmed updated (grepped for every occurrence, the two others found are the unrelated SQL-table path which already differentiates causes), and info-disclosure risk of surfacing `ex.getMessage()` (consistent with existing precedent elsewhere in this same controller — not a new exposure pattern).

**Finding (Important):** Gap fix 1's composite exclusion reintroduces the exact bug it claims to close, for composite-typed required params.

- `TabularQueryContractSupport.java` (the new missing-required check + the existing fill loop): the new check explicitly skips composite-typed params; the fill loop only ever iterates `queryParams.keySet()`, so a key that's entirely absent is never visited there either. A composite param marked `@Property(required = true)` (unconditionally required, per `isRequired()`'s own javadoc) that is omitted from `queryParams` altogether is now checked **nowhere at all**.
- Concretely demonstrable, not theoretical: `FakeNamedConnectorQuery.getParameters()` — explicitly documented as modeling the real `EndpointJsonQuery.getParameters()` — is exactly such a property. Supplying `endpoint` but omitting `parameters` entirely is untested and unchecked; the existing `rejectsMissingRequiredParameter` test only covers "present but empty," never "absent."
- Not a regression (composites were equally unchecked before this PR) — but it undercuts gap fix 1's stated purpose for this one parameter shape, and should be scoped down explicitly rather than left implying complete coverage.

**Minor (sub-threshold, noted for the fixer's convenience):** `&& type != File.class` alongside the composite check is dead code — `isCompositeType(File.class)` is already always `false`. Harmless but confusing.

**Verdict: changes requested.** (Posted as a plain comment — GitHub refuses a request-changes review on one's own PR.)

fredaan:
## Round 1 fix — commit 4d6a040cf

Addressed the composite-omission finding by **scoping the claim down**, not by changing enforcement behavior.

**Why not just remove the composite exclusion:** already tried this during original development (before this commit history) — it broke 11 existing tests in `TabularQueryContractSupportTest` that correctly call `apply(query, params("endpoint", "Repos"))` with no `parameters` key at all and expect success, because the "Repos" endpoint fixture has zero required path parameters. A composite's `required=true` is a static bean-property flag; whether it actually applies depends on which endpoint/value was chosen — information this top-level, pre-live-call check doesn't have. Flipping the exclusion trades a real-but-narrow gap (composite omission unchecked) for a real regression (rejecting valid requests that correctly omit an inapplicable/empty composite).

**What changed instead** (`core/src/main/java/inetsoft/web/wiz/service/TabularQueryContractSupport.java`):
- Rewrote the comment above the composite exclusion to state plainly that this is a **known, not-closed gap** (a composite marked `required=true` but entirely absent from `queryParams` is caught by neither this check nor `fillNamedSkeleton`), that it's pre-existing (not introduced by this PR), and why closing it unconditionally isn't safe here — rather than implying complete coverage.
- Removed the dead `&& type != File.class` in the *new* required-param loop only — confirmed via `TabularSchemaExtractor#isScalar` that `File.class` is already scalar, so `isCompositeType(File.class)` is always `false`; the redundant check was inert. Left the pre-existing `&& type != File.class` in the fill loop below (line ~164) untouched — that one predates this PR and wasn't part of the finding.
- Added `anOmittedRequiredCompositeIsNotCaughtByThisCheck_knownGap` documenting the current, accepted behavior for `FakeNamedConnectorQuery`'s `parameters` property (`endpoint="Repos"` with no `parameters` key succeeds today, even though "Repos" has a required nested `id` once `parameters` is supplied at all) — so the gap can't be silently reintroduced/forgotten without deliberate reconsideration.

**Tests:** `TabularQueryContractSupportTest` (37, was 36), `TabularSchemaExtractorTest`, `WizTabularControllerTest`, `WorksheetAgentControllerTest`, `WorksheetTableServiceProbeHintTest`, `TabularEndpointBindingSupportTest` — all pass:
```
./mvnw -q -pl core -am test -Dtest=TabularQueryContractSupportTest,TabularSchemaExtractorTest,WizTabularControllerTest,WorksheetAgentControllerTest,WorksheetTableServiceProbeHintTest,TabularEndpointBindingSupportTest -Dsurefire.failIfNoSpecifiedTests=false
```

**Doc reference note:** the task pointed at `E:\inetsoft\stylebi-wiz\docs\tabular\stylebi-tabular-datasource-config-and-queryparams-gaps.md` to check for the same overclaiming wording — that file doesn't exist anywhere in the stylebi-wiz repo (checked both the exact path and a repo-wide filename search), so there's nothing to flag there.

fredaan:
## Round 2 — commit 4d6a040cf (range e667c30..4d6a040 since round 1)

**Checked:** whether round 1's finding (composite-required-omission gap silently implied as covered) is now honestly documented rather than closed unsafely; whether the dead-code removal was scoped correctly (new check only, pre-existing fill-loop occurrence left untouched); whether the new test is honest about documenting a known, accepted gap rather than being mistakable for "gap closed"; independently re-verified the regression-risk claim itself (`setsEndpointAndBuildsSuffix`, a pre-existing test, really does call `apply(query, params("endpoint", "Repos"))` with no `parameters` key and expects success — confirming closing the gap unconditionally really would break legitimate existing usage).

- The rewritten comment (`TabularQueryContractSupport.java:115-136`) explicitly says "KNOWN, NOT-CLOSED gap", explains why (a composite's required-ness is endpoint-/choice-dependent, and this static annotation can't distinguish "always required" from "required once chosen"), and points at the concrete evidence (`FakeNamedConnectorQuery#getParameters`, `setsEndpointAndBuildsSuffix`). No longer implies coverage it doesn't have.
- Dead-code cleanup confirmed scoped correctly: `&& type != File.class` removed from the NEW check (line ~147), left untouched in the pre-existing fill loop (line ~174) — exactly the distinction requested.
- The new test (`anOmittedRequiredCompositeIsNotCaughtByThisCheck_knownGap`) reads as an honest regression-documentation test even out of context — the `_knownGap` suffix and docstring make it clear this documents current behavior, not a claim it's desirable.
- Ran the full scope myself: `TabularSchemaExtractorTest` 20/20, `WizTabularControllerTest` 40/40, `TabularEndpointBindingSupportTest` 25/25, `TabularQueryContractSupportTest` 37/37 (36+1 new), `WorksheetTableServiceProbeHintTest` 5/5, `WorksheetAgentControllerTest` 165/165 — 292/292 green.
- `git show 4d6a040cf --stat`: exactly 2 files, +35/-6 — comment rewrite + dead-code removal + one new test, nothing else.

**New findings:** none.

**Verdict: approved.** (Posted as a plain comment — GitHub refuses a self-approval review on one's own PR.)

fredaan:
## Round 3 — commit 639594d57 (self-caught, not from a reviewer)

CI's full \`build\` check (not part of my own scoped local regression runs, which only covered the test classes I judged relevant) failed on \`WizTabularControllerSecurityTest.mappingsMatchTheFrozenContract\` — a test that pins the exact set of \`/api/wiz/tabular/*\` endpoint mappings, not updated for the new \`/tabular/datasource-config\` endpoint. Fixed by adding it to the frozen list, plus a sibling gate test (\`tabularDatasourceConfigIsGatedOnReadWithAWiredPermissionPath\`) matching the existing coverage for its sibling \`getQuerySchema\`/\`query-schema\`.

This is exactly the kind of gap a full CI run catches that a manually-scoped local regression selection can miss — noting it plainly rather than treating my own earlier "all green" claims as if they'd covered this.

fredaan:
**CI investigation — final: `build` failure confirmed pre-existing and unrelated, safe to merge past**

Re-ran the failed `build` job (run 34453169515) a second time to test the round-3 reviewer's flakiness hypothesis. It failed again, but with a **different, more conclusive** signature than round 3's — this time the log shows a deterministic compile failure, not the earlier Spring-context NPE cascade:

```
[ERROR] Failed to execute goal com.inetsoft.stylebi.build:node-maven-plugin:1.2.0-SNAPSHOT:npm (default) on project inetsoft-web:
Failed to execute npm with arguments run verify: 'npm run verify' failed.
...
✘ [ERROR] TS2554: Expected 6 arguments, but got 5. [plugin angular-compiler]
projects/portal/src/app/widget/formula-editor/formula-editor-dialog.component.spec.ts:349:15:
  349 │ ... dialog = new FormulaEditorDialog(editorService, modalService,...
```

This is the **exact same** pre-existing Angular TS2554 error (missing `followFocusService` constructor arg) already identified and merged past on stylebi PR #5130 — not a flake, a standing, reproducible failure on the base branch's frontend build, independent of this PR.

Confirmed unrelated: `git diff --stat origin/stylebi-wiz-main-rebase...639594d57` touches 13 files, all under `core/src/...`, zero files under `web/`.

All PR-relevant tests (including `WizTabularControllerSecurityTest`, fixed in the round-3 commit `639594d57`) pass. The only reason `build` shows red is this pre-existing, unrelated frontend compile error. Proceeding to merge per the established precedent (PR #5130).

(Posted as a plain comment, not a review — GitHub blocks self-review on one's own PR.)

PR #5138, opened 2026-09-10
Check permission and dependencies on every nested data source before deleting a folder

fredaan:
## Summary

`RepositoryObjectService.removeDataSourceFolder` only checked permission and outer dependencies on a folder's **immediate-child** data sources, then unconditionally called `DataSourceRegistry.removeDataSourceFolder`, which recursively deletes **every** nested data source at any depth with **no check of its own**. A caller with delete permission on the top-level folder could therefore remove a data source several sub-folders deep that they had no individual permission on, or that had an unresolved outer dependency — silently, with no error and no record of what was actually removed.

This also blocked a downstream wiz-services feature (companion PR in `stylebi-wiz`, not part of this repo) that needs to know exactly which data sources a folder deletion removed, to clean up their vector-store records.

## Fix

`removeDataSourceFolder` is now two-phase:
1. **Check everything, delete nothing.** Recursively enumerate every nested data source at any depth (not just immediate children), and check permission + dependencies on each.
2. **Delete only if every check passed.** A single failure anywhere refuses the whole operation — nothing gets deleted (previously: a partial failure left everything before it truly deleted, invisibly).

The result now also reports every data source actually removed, with its type (`FolderDeleteResult` / `DeletedDataSource`, new small records), threaded through `DataSourceBrowserService.deleteDataSourceFolder` and `WizDatabaseController.deleteDatasources`.

**One call site the initial design missed**, found via a compile error: `DataSourceController.deleteDatasourceFolder` (the native EM REST endpoint `DELETE api/data/datasources/browser/folder/**`) also calls this method. Fixed with the same pure `.status()` unwrap used elsewhere — no behavior change for that endpoint.

Design doc (in the `stylebi-wiz` repo, cross-repo context): `docs/superpowers/specs/2026-09-10-folder-delete-permission-fix-and-cascade-cleanup-design.md`

## Test plan

- [x] `RepositoryObjectServiceTest`: 7 → 14 tests. New: 5 `removeDataSourceFolder` scenarios (all-pass, deep-nested-permission-failure regression test, deep-nested-dependency-failure, force=true bypass, unreadable-type tolerance) + 2 `deleteNodes`/`DATA_SOURCE_FOLDER` coverage-gap tests.
- [x] `WizDatabaseControllerDeleteMoveTest`: 10 → 13 tests. New: folder-branch `deletedDataSources` population/absence.
- [x] `WizDatabaseControllerRealRegistryTest`: 4/4 unchanged and confirmed still passing against a REAL (non-mocked) `DataSourceRegistry`, including `deleteFolder_doesNotOrphanADatasourceNestedInASubfolder` — verified the fix's new permission check passes for the right reason under this suite's blanket-allow permission stub, not because it was bypassed.
- [x] The core regression test (`removeDataSourceFolder` with a data source two levels deep failing permission) proves the actual security bug: against the old implementation this data source was never even looked at before the low-level recursive delete swept it up unconditionally.

## Known follow-up (not blocking)

`DataSourceController.deleteDatasourceFolder` has no test coverage before or after this change (a pre-existing gap, not a regression). Worth a fast follow-up test.

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/content/repository/DeletedDataSource.java
+++ b/core/src/main/java/inetsoft/web/admin/content/repository/DeletedDataSource.java
@@ -1,0 +1,28 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.web.admin.content.repository;
+
+/**
+ * One data source actually removed as part of a folder deletion.
+ *
+ * @param path       full repository path of the removed data source.
+ * @param sourceType {@code XDataSource.getType()}, or {@code null} if the type could not be
+ *                   read (the data source object failed to load) -- deletion is not blocked on
+ *                   this, since the type is only needed by callers doing cleanup keyed on it.
+ */
+public record DeletedDataSource(String path, String sourceType) {}
--- a/core/src/main/java/inetsoft/web/admin/content/repository/FolderDeleteResult.java
+++ b/core/src/main/java/inetsoft/web/admin/content/repository/FolderDeleteResult.java
@@ -1,0 +1,40 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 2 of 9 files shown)


fredaan:
## Review status

This change went through task-scoped implementer + reviewer subagent review before being opened as a PR.

**Verdict: Approved, no Critical/Important findings.**

The reviewer independently verified (against the diff, not just the implementer's report):
- The new two-phase `removeDataSourceFolder` genuinely enumerates every nested data source at any depth before deleting anything, and a single failed check anywhere refuses the whole operation with nothing deleted.
- All three call-site adaptations (`RepositoryObjectService.deleteNodes`, `DataSourceBrowserService.deleteDataSourceFolder`, and the design-missed `DataSourceController.deleteDatasourceFolder`) are pure, behavior-preserving type unwraps — grepped independently to confirm no 4th call site exists.
- Compatibility with `WizDatabaseControllerRealRegistryTest`'s existing real-registry tests (including `deleteFolder_doesNotOrphanADatasourceNestedInASubfolder`) was specifically checked: they still pass, and for the right reason (the new permission check genuinely passes under that test's blanket-allow stub, not because it's bypassed).
- The core regression test (a data source two levels deep failing permission) was confirmed to be a real behavioral assertion — verified `removeDataSource`/`removeDataSourceFolder` are never called — not a mock-echo test.

Two Minor, non-blocking notes: `DataSourceController.deleteDatasourceFolder` has no test coverage before or after this change (pre-existing gap, not a regression — worth a fast follow-up); and the phase-1/phase-2 double-check is intentional (TOCTOU safety) but worth a one-line callout so a future reader doesn't "simplify" it away.

**Note on merge approval**: this PR was opened by the same account driving this session, so GitHub will not let this session self-approve it. A human reviewer needs to give the actual merge approval.

PR #5140, opened 2026-09-10
fix(tabular): distinguish a genuine zero-row success from a failed probe

Nelson-cao:
`buildTabularTable` throws a single message when the column-discovery probe
returns zero columns, and that message names the parameters and the credentials
as the suspects. The commonest real cause is neither: the request succeeded and
the endpoint has no rows under the parameters given — an ordinary outcome for a
date-filtered query.

Reproduced against a live Stripe account. `GET /v1/payment_intents` returned
HTTP 200 with an empty `data` array; every suspect the message named was
innocent. A caller acting on it concluded the data source was unusable and
rebuilt the table against an unrelated local file instead.

The evidence needed to tell the two cases apart was already at the throw site
and never read. `EndpointJsonQueryRunner` records the response shape before the
row path narrows it, and `TabularHandler.execute` copies that shape back onto
the caller's own query unconditionally, above the success/failure branch — so
`getResponseShape()` is non-null exactly when the request returned a parseable
body. `applyResponseShape` already reads the same slot, but only on the success
path.

## Changes

- `emptyColumnsMessage` splits the message on that signal. The no-shape branch
  keeps today's wording verbatim, so a caller matching on it still matches. The
  shape-present branch states that the request succeeded, carries the recorded
  shape, names the row path in effect, and says that zero rows means zero
  columns.
- The row path is read reflectively, so `core` gains no compile-time dependency
  on a connector type. A constant-pool assertion in
  `CoreHasNoConnectorKnowledgeTest` holds that boundary.
- The connector's own recovered exception message is appended to both branches,
  sanitized to a single line: `rootMessage` truncates at the first newline and
  `UserMessage.merge` joins with one, so an unsanitized suffix would be cut.
- `CoreTool`'s user-message list is thread-local and nothing cleared it between
  tables in one `createTables` batch, so one table's connector error could be
  reported as the next table's cause. It is now cleared before the probe.

## Behaviour that does not change

Zero rows still fails the build. For a JSON REST endpoint columns are derived
from the rows returned, so admitting a zero-column table would only move the
failure downstream into binding, where it reads as an unknown-field error.

A `SERVER_FILE` or generic non-endpoint REST source takes the no-shape branch
even on a genuine zero-row success — only `EndpointJsonQueryRunner` records a
shape, and the base `RestJsonQueryRunner` does not. Tests pin that behaviour so
a later change to it is deliberate.

## Tests

`WorksheetTableServiceEmptyResultMessageTest` (10 cases) covers both branches,
the recovered detail on each, the single-line guarantee, the cross-table
attribution leak, the unchanged success path, and a pairwise case asserting
that shape presence alone flips the branch. 148 tests across every
`WorksheetTableService*Test` pass unchanged.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
+++ b/core/src/main/java/inetsoft/web/wiz/service/WorksheetTableService.java
@@ -52,7 +52,9 @@ import inetsoft.uql.tabular.TabularUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import inetsoft.util.Catalog;
+import inetsoft.util.CoreTool;
 import inetsoft.util.Tool;
+import inetsoft.util.UserMessage;
 import inetsoft.web.composer.ws.LayoutGraphService;
 import inetsoft.web.composer.ws.joins.InnerJoinService;
 import inetsoft.web.portal.controller.database.DataSourceService;
@@ -1547,6 +1549,12 @@ public class WorksheetTableService {
       probeVars.put(XQuery.HINT_MAX_ROWS,
                     String.valueOf(Math.max(20, src.getSampleRows() == null ? 0 : src.getSampleRows())));
 
+      // Clear any user-message residue from an earlier table in this same batch. createTables loops
+      // over every requested table on ONE thread and nothing clears CoreTool's per-thread message
+      // list between iterations (see recoveredExceptionDetail) — without this, a message left behind
+      // by an earlier table's connector could be read below and misattributed to THIS table's probe.
+      CoreTool.clearUserMessage();
+
       // A null QueryManager, matching what TabularTableAssembly.dependencyChanged passes and what
       // buildSqlTable uses for its own column resolution. Taking an AssetQuerySandbox instead would
       // mean either disposing it — leaving the "queryManager" property pointing at a dead object on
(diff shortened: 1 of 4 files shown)


Nelson-cao:
**Review round 1** — commit `f4a037eb6`.

Checked against the charter's full assertion table (A1-A9, C1-C6) and the
five specific targets called out in reconcile: the `CoreTool
.clearUserMessage()` insertion (traced two entry points —
`/ws/table`'s `createTables` batch and `/tabular/probe/table`'s per-file
probe — and every AOP/advice class that drains `CoreTool`'s user-message
thread-local, to confirm nothing else in the request lifecycle reads what
this clears); the `TabularHandler.execute` cache-hit early return
(independently confirmed unreachable from the column-discovery probe:
`TabularQuery.loadOutputColumns` hardcodes a `null` visitor, and
`isQueryCached` short-circuits on that before computing anything); the
no-shape branch's wording (confirmed byte-identical up to the appended
detail suffix, and confirmed the one real consumer in this repo,
`wiz-services`' `worksheetService.ts`, only checks for a non-empty
`errorMessage` and never pattern-matches its text); `singleArrayFieldHint`
(walked every shape variant — zero/one/many array fields, nested arrays,
a top-level array, an empty array field, the dictionary-collapse case —
against the actual `JsonShapeDistiller` output types); and the narrowed
test assertion around the word "credentials" (confirmed the pairwise
swap test fails loudly rather than passing vacuously if a future reword
ever drops the sentinel from both branches).

**New finding (blocker)**: the response-shape text embedded in the
success-framed message is not sanitized against embedded `\n`/`\r`, unlike
the exception-detail suffix. `JsonShapeDistiller` preserves a response's
real field *names* verbatim as map keys (only leaf *values* are replaced
with type-name literals) — a response whose JSON field name happens to
contain a literal newline (valid JSON, e.g. `{"a\nb": []}`) would have
that newline land in the thrown message, and the existing `rootMessage()`
truncation (unchanged by this PR, per C4) would then silently drop
everything after it — the row-path hint, "Parameters sent", the
"not a credentials problem" reassurance, and any recovered exception
detail. That is the exact silent-truncation defect class this PR exists
to close, reintroduced through the one connector-reported input (field
names) the design's sanitization reasoning didn't cover. Fix: sanitize
`String.valueOf(shape)` the same way `detail` already is, and add a
regression test with a shape fixture carrying an embedded `\n` in a key.

Everything else checked out, including active attempts to break each of
the five named targets. No other findings.

Nelson-cao:
**Review round 2** — commit `373b2d2cb` (on top of `f4a037eb6`).

Re-checked R1-1 (the round-1 blocker: response-shape/row-path text
embedded in the message wasn't sanitized against embedded `\n`/`\r`) plus
the two coverage findings folded into this round (`singleArrayFieldHint`'s
untested branches, a named C1 assertion on the shape-present test).

Verified, independently rather than by trusting the fix log:
- The sanitize point (`return singleLine(message);`) is genuinely
  unmissable — read the full method body and confirmed there is exactly
  one `return`, with both branches assigning to the same local rather than
  returning directly.
- `recoveredExceptionDetail()`'s dropped per-ingredient sanitize is safe
  today (grepped: exactly one caller, private method) but is a
  documented latent hole if a second caller is ever added without going
  through `emptyColumnsMessage`'s tail sanitize — recorded as a nit, not
  actionable now since there is no live second caller.
- Reconstructed both branches' literal text token-by-token against the
  pre-round-2 version: the only change anywhere in the method's string
  content is the one line-wrap reflow ("but" moving between two
  literals), and the concatenated result is provably byte-identical
  either way. No silent text drift.
- Traced what the two new regression tests actually observe against the
  pre-fix code, given `rootMessage()`'s own truncate-at-first-`\n`
  behavior: both do fail on revert, but via their `contains("Parameters
  sent")` assertion, not their `!contains("\n")` one (which is trivially
  true either way, since truncation itself removes the newline). Recorded
  as a documentation nit — the tests are correct, just slightly
  mis-described.
- All four new `singleArrayFieldHint` coverage tests and the new C1
  assertion are non-vacuous — each targets a distinct branch and would
  fail under a plausible mutation of the method.
- Weighed the lead's widened fix (sanitize once at the tail) against my
  own narrower round-1 proposal (sanitize the shape only): the broader
  version is strictly better — it also closes a pre-existing,
  never-in-scope-before exposure on `probeDesc`/`dsName`, with no
  increase in the method's blast radius.

New nits only (recorded, not blocking): the sanitize-once pattern depends
on future branches continuing to assign rather than return directly;
blanking (vs. escaping) a newline inside a data field like a shape key
alters its displayed spelling. Neither rises above cosmetic.

**Verdict: approved.**

fredaan (approved):

PR #5142, opened 2026-09-10
Bug #76569 (VSC-003, VSC-005): is_null arity fix + null-lens export safety net

Zxm-happy:
## Summary

Two independent fixes for two of Redmine #76569's six condition-tooling findings (VSC-003 and VSC-005). See `docs/teams/2026-09-10-bugs-condition-tooling-and-schedule-date/bug-76569/01-diagnosis.md` and `02-refute.md` in the `wiz` repo for the full investigation this PR implements.

### VSC-005 — `is_null` rejected despite being a documented alias for `null`

`ConditionVocabulary.VALUELESS` only listed `"null"`, so `requireValueArity()` rejected an `is_null` clause with a false "needs at least one value" error, even though `operators()` maps `is_null` to the same `XCondition.NULL` constant as `null`. One-line fix: add `"is_null"` to `VALUELESS`.

### VSC-003 — a viewsheet assembly whose condition-script evaluation fails NPEs the whole export (safety net, not the full fix)

Reported stack: `ScriptImageService.exportViewsheetToPng -> VSExportService.exportViewsheet -> ... -> SVGTableHelper.beginRoundCornerClip -> VSTableDataHelper.getObjectPixelBounds -> VSTableHelper.getTableRectangle -> VSTableHelper.getVisibleRowCount`, ending in `lens.getHeaderRowCount()` on a null `lens`.

Guarding only `getVisibleRowCount` (the method the reported NPE names) turned out not to be sufficient by itself: I verified its caller `getTableRectangle` unconditionally calls `lens.getRowCount()` immediately after the guarded call, and *its* caller `getObjectPixelBounds` unconditionally calls `lens.getRowHeights()` — both re-dereference the same null `lens` regardless of what `getVisibleRowCount` returns. So all three needed a null-lens guard for the fix to actually stop the export crash for the reported repro, not just move the NPE one line down. `VSCrosstabHelper`'s own `getTableRectangle` (`VSCrosstabHelper.java:317`) calls the same static `getVisibleRowCount` and has no such unconditional re-dereference after it, so the crosstab path is fully covered by the `getVisibleRowCount` guard alone.

A subsequent review pass found a fourth file with the identical defeat pattern: `PDFTableHelper.getTableRectangle` calls the now-guarded `super.getTableRectangle(info, lens)`, but then unconditionally dereferences the same `lens` parameter right after (`lens.getHeaderRowCount()`, `lens.getRowCount()`), reintroducing the NPE for PDF export of a plain (non-crosstab) table assembly even though the SVG/PNG path was fully fixed. Added the same guard there, returning the already-safe sentinel `Rectangle` from `super.getTableRectangle()`. Also explicitly checked `PDFCrosstabHelper.getTableRectangle` and `SVGCrosstabHelper.getTableRectangle` — both only touch `rec.height` on the already-guarded `Rectangle` from `super.getTableRectangle()`, with no direct `lens` dereference, so neither needs a change.

This does **not** address where the null lens comes from in the first place (a failed condition-script evaluation somewhere upstream) — that remains an open gap flagged in the diagnosis for a follow-up pass. This PR only stops the resulting NPE from crashing the whole viewsheet's render.

## Testing

- `ConditionVocabularyTest.acceptsIsNullWithNoValues` — new regression test for VSC-005, mirrors the existing `acceptsNullWithNoValues`/`refusesNullWithValues` pair.
- `VSTableHelperTest.getVisibleRowCountReturnsZeroForNullLens` — new test file, asserts the null-lens guard.
- Ran both scoped: `./mvnw -pl core -o test -Dtest=VSTableHelperTest,ConditionVocabularyTest` → 48 tests, 0 failures.
- No instance-level test scaffolding exists in this suite for constructing a `PDFTableHelper` (or any `VSTableHelper` subclass) instance — it requires a `PDFCoordinateHelper` and a configured `VSExporter`, which this suite does not currently mock. Per the reviewer's own allowance, skipped adding new scaffolding for this guard rather than build it from scratch; the guard is a direct mirror of the already-tested pattern in `VSTableHelper.getTableRectangle`.

### DIFF
--- a/core/src/main/java/inetsoft/report/io/viewsheet/VSTableDataHelper.java
+++ b/core/src/main/java/inetsoft/report/io/viewsheet/VSTableDataHelper.java
@@ -157,6 +157,11 @@ public abstract class VSTableDataHelper extends ExporterHelper {
    protected Rectangle2D getObjectPixelBounds(TableDataVSAssemblyInfo info,
       VSTableLens lens, CoordinateHelper vHelper)
    {
+      if(lens == null) {
+         Point pos = info.getPixelOffset();
+         return new Rectangle2D.Double(pos.x, pos.y, 0, 0);
+      }
+
       boolean match = getExporter().isMatchLayout();
       Rectangle tableRange = getTableRectangle(info, lens);
       Dimension size = new Dimension(tableRange.width, tableRange.height);
--- a/core/src/main/java/inetsoft/report/io/viewsheet/VSTableHelper.java
+++ b/core/src/main/java/inetsoft/report/io/viewsheet/VSTableHelper.java
@@ -198,6 +198,10 @@ public abstract class VSTableHelper extends VSTableDataHelper {
    protected static int getVisibleRowCount(TableDataVSAssemblyInfo info,
                                            VSTableLens lens)
    {
+      if(lens == null) {
+         return 0;
+      }
+
       Dimension size = CoordinateHelper.getAssemblySize(
          info, CoordinateHelper.getLensSize(lens, true));
       int infoRows = 0;
@@ -241,6 +245,10 @@ public abstract class VSTableHelper extends VSTableDataHelper {
    protected Rectangle getTableRectangle(TableDataVSAssemblyInfo info,
                                          VSTableLens lens)
    {
+      if(lens == null) {
+         return new Rectangle(info.getPixelOffset().x, info.getPixelOffset().y, 1, 0);
+      }
+
       VSExporter exporter = this.getExporter();
 
       // @by gregm ensure we do not print too many rows because of wrapping.
--- a/core/src/main/java/inetsoft/report/io/viewsheet/pdf/PDFTableHelper.java
+++ b/core/src/main/java/inetsoft/report/io/viewsheet/pdf/PDFTableHelper.java
@@ -185,6 +185,10 @@ public class PDFTableHelper extends VSTableHelper {
    {
       Rectangle rec = super.getTableRectangle(info, lens);
 
+      if(lens == null) {
+         return rec;
+      }
+
       int headers = lens.getHeaderRowCount();
       int rows = lens.getRowCount();
       int height = 0;
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/ConditionVocabulary.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/ConditionVocabulary.java
@@ -53,7 +53,7 @@ public final class ConditionVocabulary {
    private static final Map<String, Integer> OPERATORS = operators();
 
    /** Operators that take no values at all. */
-   private static final Set<String> VALUELESS = Set.of("null");
+   private static final Set<String> VALUELESS = Set.of("null", "is_null");
 
    /** Operators that take exactly two. */
    private static final Set<String> PAIRED = Set.of("between");
--- a/core/src/test/java/inetsoft/report/io/viewsheet/VSTableHelperTest.java
+++ b/core/src/test/java/inetsoft/report/io/viewsheet/VSTableHelperTest.java
@@ -1,0 +1,39 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.report.io.viewsheet;
+
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@Tag("core")
+class VSTableHelperTest {
+
+   // Bug #76569 (VSC-003): a viewsheet assembly whose condition-script evaluation fails can
+   // reach export/render with a null VSTableLens. getVisibleRowCount used to call
+   // lens.getHeaderRowCount() unconditionally, NPE-ing and crashing the whole viewsheet's
+   // export. This is a safety net only -- it does not address where the null lens comes from.
+   @Test
+   void getVisibleRowCountReturnsZeroForNullLens() {
+      int[] result = new int[1];
+      assertDoesNotThrow(() -> result[0] = VSTableHelper.getVisibleRowCount(null, null));
+      assertEquals(0, result[0]);
+   }
+}
--- a/core/src/test/java/inetsoft/web/wiz/viewsheet/ConditionVocabularyTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/viewsheet/ConditionVocabularyTest.java
@@ -286,6 +286,15 @@ class ConditionVocabularyTest {
                       List.of(clause("Region", "null", List.of("x"), null)), FIELDS));
    }
 
+   @Test
+   void acceptsIsNullWithNoValues() {
+      Object[] list = ConditionVocabulary.toConditionList(
+         List.of(clause("Region", "is_null", List.of(), null)), FIELDS);
+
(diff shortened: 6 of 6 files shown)


Zxm-happy:
# Bug 76569 -- reviewer pass over stylebi PR #5142 (VSC-003, VSC-005)

Reviewer phase, /debug-workflow. Reviewing
https://github.com/inetsoft-technology/stylebi/pull/5142 (base stylebi-wiz-main-rebase, head
bug-76569-condition-tooling, commit 8c0a0e51e). Read the actual diff (gh pr diff) and the full
bodies of every touched/adjacent method directly from the checked-out worktree
(E:\inetsoft\stylebi-worktrees\bug-76569-condition-tooling\community), not just the fixer's own
summary (03-fix-stylebi.md).

Verdict: CHANGES REQUESTED. Not merged.

---

## VSC-005 -- is_null VALUELESS fix

Confirmed correct. Matches the diagnosis/refutation exactly: requireOperator() returns the
un-normalized token, requireValueArity() checks that token against VALUELESS (which only had
"null"), so is_null fell through to the "needs a value" branch even though operators() already
maps it to the same XCondition.NULL constant. The one-line fix (Set.of("null", "is_null")) is
correct and sufficient -- VALUELESS is private with a single reader, confirmed by the refuter and
re-confirmed here by grep. The new test ConditionVocabularyTest.acceptsIsNullWithNoValues builds a
clause with operator is_null and no values and asserts toConditionList returns one ConditionModel
with getOperation() == XCondition.NULL -- this is the right assertion for the claimed root cause. No
issue here.

## VSC-003 -- null-lens safety net: the "three files, not one" claim is right, but incomplete

I independently read the full bodies of VSTableHelper.getVisibleRowCount, getTableRectangle, and
VSTableDataHelper.getObjectPixelBounds (not just the diff hunks). The fixer's expanded-scope claim
is correct as far as it goes:

- VSTableHelper.getTableRectangle, immediately after the (now-guarded) call to
  getVisibleRowCount, unconditionally does rowCount = Math.min(rowCount, lens.getRowCount()) --
  confirmed, this really would NPE with only the getVisibleRowCount guard in place.
- VSTableDataHelper.getObjectPixelBounds, after calling getTableRectangle, unconditionally does
  int[] heights = lens.getRowHeights(); -- confirmed, same issue one level up.
- VSCrosstabHelper's own getTableRectangle (line 312-321) does not need its own guard for the
  reported crash path specifically: I traced the actual entry point,
  VSTableDataHelper.write() (line 323), and confirmed beginRoundCornerClip(info, lens) is called
  unconditionally (not itself inside an if(lens != null)), while the sibling writeData(info, lens)
  call a few lines later is guarded by if(lens != null). So beginRoundCornerClip ->
  getObjectPixelBounds -> getTableRectangle is the only lens-nullable path, and
  getObjectPixelBounds's own new top-of-method guard (if(lens == null) return sentinel;) returns
  before ever calling getTableRectangle polymorphically -- so neither
  VSTableHelper.getTableRectangle nor VSCrosstabHelper.getTableRectangle's bodies are reached at
  all when lens is null, via this call chain. VSCrosstabHelper.getTableRectangle's own ternary
  (exporter.isMatchLayout() && info.isShrink() ? getVisibleRowCount(...) : lens.getRowCount()) is
  real and would NPE on a null lens if reached directly (e.g. from VSCrosstabHelper.writeData),
  but writeData is never called with a null lens (guarded at the write() call site), so this is
  not a live gap for this bug. The fixer's claim on this specific point holds up.

But: I found a fourth file with the identical defeat pattern that none of the three passes
(diagnosis, refutation, fixer) checked -- PDFTableHelper.getTableRectangle
(core/src/main/java/inetsoft/report/io/viewsheet/pdf/PDFTableHelper.java:182-192):

    @Override
    protected Rectangle getTableRectangle(TableDataVSAssemblyInfo info,
                                          VSTableLens lens)
    {
       Rectangle rec = super.getTableRectangle(info, lens);

       int headers = lens.getHeaderRowCount();   // unconditional, unguarded
       int rows = lens.getRowCount();            // unconditional, unguarded
       ...

PDFTableHelper extends VSTableHelper and overrides getTableRectangle -- for PDF export, this
override (not VSTableHelper's own body) is what actually runs via polymorphic dispatch from
VSTableDataHelper.getObjectPixelBounds. It calls super.getTableRectangle(info, lens) (now
correctly guarded, returns the sentinel Rectangle without throwing), but then immediately
dereferences the same lens parameter unconditionally in its own body, right after -- the exact
same defeat pattern the fixer identified and fixed in VSTableHelper.getTableRectangle and
VSTableDataHelper.getObjectPixelBounds, just one class further down the override chain, and missed.

I confirmed this isn't neutralized elsewhere:
- SVGTableHelper (the class actually named in the reported stack trace, for PNG export) does
  not override getTableRectangle at all -- so the SVG/PNG path really is fully fixed by this PR,
  exactly as claimed. The gap is specific to the PDF export path.
- PDFCrosstabHelper.getTableRectangle and SVGCrosstabHelper.getTableRectangle also override the
  method, but only touch rec.height (the already-computed, already-safe Rectangle from super) --
  neither dereferences lens directly, so those two are fine.
- The whole per-assembly export loop in AbstractVSExporter (~line 1284-1496) wraps the entire
  for(assemblies) loop in a single try { ... } catch(SheetMaxRowsException maxrows) -- no generic
  catch(Exception) around an individual assembly's write. So an NPE thrown from
  PDFTableHelper.getTableRectangle for one broken table assembly propagates out and fails the
  entire PDF export, matching the exact "whole viewsheet fails to render" symptom this PR exists to
  stop -- just via PDF instead of PNG/SVG.

This is a live, concrete, currently-unaddressed regression path for the same bug, in the same PR's
stated scope ("stops the entire viewsheet from failing to render over one assembly's broken
condition"). The PR's own description and commit message claim the crosstab call site was checked
and is safe (true), but neither mentions checking the format-specific getTableRectangle overrides
in pdf/PDFTableHelper.java / pdf/PDFCrosstabHelper.java / svg/SVGCrosstabHelper.java at all --
only PDFTableHelper's override turns out to reintroduce the exact NPE this PR is meant to close, for
PDF exports of plain (non-crosstab) tables.

### Requested fix

Add the same guard to PDFTableHelper.getTableRectangle, mirroring the existing pattern (e.g.
if(lens == null) { return rec; } right after the super.getTableRectangle(info, lens) call, since
rec at that point is already the safe sentinel from VSTableHelper.getTableRectangle's own guard).
Please also re-scan pdf/PDFCrosstabHelper.java and svg/SVGCrosstabHelper.java explicitly (I
checked both this pass and found no unguarded lens dereference in either, but they weren't
mentioned in the PR's own reasoning, so worth stating explicitly in the PR body once fixed) and add a
regression test analogous to VSTableHelperTest.getVisibleRowCountReturnsZeroForNullLens if a
constructible test path exists for PDFTableHelper (the fixer already noted no instance-level test
scaffolding exists for VSTableHelper subclasses in this suite -- if that's still true for
PDFTableHelper, it's acceptable to skip the test and note the same gap the fixer already flagged
for the other two guards, rather than block on building new scaffolding from scratch).

## Other checks (all clean)

- No scope creep: confirmed the diff touches only VSTableDataHelper.java (+5),
  VSTableHelper.java (+8), ConditionVocabulary.java (VALUELESS line only, +1/-1), and the two test
  files. No changes to ConditionVocabulary.junction()/XConditionGroup.java (VSC-006, explicitly
  out of scope and contested) and no sql-variant persistence changes (also explicitly deferred).
- Test tag: VSTableHelperTest carries @Tag("core") -- confirmed present in the diff, matching
  ConditionVocabularyTest's existing annotation. Without it the class would silently not run under
  this module's surefire config (<groups>core</groups>), per the fixer's own correctly-flagged
  gotcha.
- Test correctness: VSTableHelperTest.getVisibleRowCountReturnsZeroForNullLens calls
  VSTableHelper.getVisibleRowCount(null, null) and asserts it returns 0 without throwing -- matches
  the guard exactly. ConditionVocabularyTest.acceptsIsNullWithNoValues asserts on
  XCondition.NULL -- matches the guard exactly.
- Checks/mergeability: at review time, claude-review and build checks were still
  IN_PROGRESS; Semgrep/Semgrep OSS/license-cla had already passed. mergeable: MERGEABLE
  (no conflicts against stylebi-wiz-main-rebase). Per this run's own rule, an in-progress check
  means this PR should be re-checked later rather than merged now regardless of the finding below --
  but the finding below is independently blocking either way.

## Summary

Do not merge #5142 as-is. The VSC-005 fix is correct and complete. The VSC-003 fix correctly
identifies and closes the SVG/PNG export NPE (the exact reported repro) across three files, and
correctly determines VSCrosstabHelper needs no change -- but it misses a fourth file,
PDFTableHelper.getTableRectangle, which reintroduces the identical unconditional-lens-dereference
NPE for PDF export of plain table assemblies, defeating this PR's own stated goal for that export
format. Requesting a revision to add the same guard there (and to explicitly confirm
PDFCrosstabHelper/SVGCrosstabHelper are clean, which I've verified they are) before this can land.

PR #5143, opened 2026-09-10
feat(wiz): add viewsheet export endpoint (Redmine #76516 sub-cap 1)

NeverStopFMH:
## Summary
- Adds `export()` to `ViewsheetAssemblyAgentController` (`GET /api/wiz/v1/agent/viewsheet/{sessionToken}/export`), delegating to `VSExportService#exportViewsheet` -- the same primitive `WizViewsheetExportController` already uses -- against the session's already-resolved `RuntimeViewsheet`. Supports all 7 formats StyleBI's own Export dialog offers: Excel, PowerPoint, PDF, PNG, HTML, CSV, Snapshot.
- Writes the exported bytes directly to the servlet response (same `Content-Type`/`Content-Disposition` shape the human Viewer's own export gets, and the same direct-servlet-write pattern `WizViewsheetExportController` already uses) rather than base64-encoding into a JSON body -- the companion stylebi-wiz PR (composer-chat MCP plugin) streams this straight to a local file instead of relaying it through the calling agent's own conversation/token budget.
- `target` (a single table/chart assembly) is only supported for `format=PNG` so far, reusing `ScriptImageService.getAssemblyImage` -- refused by name for other formats rather than silently exporting the whole sheet.
- Fixed in passing: the controller's own `suffixFor(CSV)` returned `"csv"`, but CSV's real payload is a zip archive (`VSExportService#getSuffix` says `"zip"` -- confirmed against a real export). Now calls `VSExportService#getMime`/`getSuffix` directly instead of maintaining a second, drifted copy of the same mapping.
- Split off from a combined export+bookmark commit; bookmark lives on a sibling branch/PR (#5141), cut from the same base commit.

## Test plan
- [x] `ViewsheetAssemblyAgentControllerTest`: 78/78 passing, capturing what gets written to a mocked `HttpServletResponse`
- [x] `core` module compiles clean
- [x] Live-verified against a real StyleBI instance via the composer MCP plugin: all 7 formats plus a `target`+PNG single-assembly export, each confirmed as a genuinely valid file of its type (opened with `file`, not just trusting the tool's own report)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/ViewsheetAssemblyAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/ViewsheetAssemblyAgentController.java
@@ -19,13 +19,18 @@ package inetsoft.web.wiz.viewsheet;
 
 import inetsoft.sree.security.IdentityID;
 import inetsoft.sree.security.ResourceAction;
+import inetsoft.sree.security.ResourceType;
+import inetsoft.sree.security.SecurityEngine;
+import inetsoft.sree.security.SecurityException;
 import inetsoft.web.AutoSaveUtils;
 import inetsoft.web.composer.vs.controller.VSLayoutService;
 import inetsoft.web.wiz.WizUtil;
 import inetsoft.web.wiz.pairing.*;
 import inetsoft.web.wiz.viewsheet.model.LayoutModel;
 import inetsoft.web.wiz.viewsheet.model.ViewsheetModel;
+import jakarta.servlet.http.HttpServletResponse;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.ContentDisposition;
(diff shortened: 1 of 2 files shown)


NeverStopFMH:
Thanks for the thorough review. Fixed the one nit in ae0ad6e67: removed the unused \`IdentityID\` import from the test file.

Appreciate the confirmation on the security check (RuntimeViewsheet-overload skipping the built-in permission check) and the FYI on \`VSExportService#doExportViewsheet\`'s empty-sandbox behavior — noted for a future pass, not something this PR needs to change since it's pre-existing/shared with \`WizViewsheetExportController\`.

NeverStopFMH:
Re-review confirms this is ready — no new findings. Thanks for the thorough re-check against the actual source.

fredaan (approved):

PR #5144, opened 2026-09-10
Bug #76329, add OAuth 2.0 support to Salesforce REST connectors

jkidacki-inetsoft:
## Summary

Salesforce is retiring the SOAP `login()` API operation (Summer '27). This adds OAuth 2.0 Authorization Code flow as an alternative authentication method for `SalesforceDataSource`, `RemedyforceDataSource`, and `SalesforceReportsDataSource`, alongside the existing username/password login for backward compatibility.

## Root Cause

These connectors authenticate exclusively via the SOAP `login()` operation, which Salesforce is fully retiring. Once retired, there will be no admin workaround — connections will simply fail.

## Fix

- Added an `Authentication` toggle (Password / OAuth 2.0) to each connector's data source configuration, reusing the existing `PASSWORD_OAUTH2_WITH_FLAGS` credential type
- OAuth uses the customer's own Salesforce Connected App / External Client App (Client ID + Secret typed into the data source UI), since Salesforce requires org-specific login URLs (production/sandbox/custom domain) that a single shared app can't support
- PKCE is required by Salesforce's OAuth implementation and is now enabled automatically
- `SalesforceDataSource.createSession()` branches: OAuth mode builds the session directly from the access token + Salesforce's `instance_url` (captured from the non-standard OAuth token response field), skipping the SOAP login envelope entirely
- Legacy Password mode is completely unchanged for backward compatibility with existing saved data sources

## Test Plan

- [x] `./mvnw test -pl core,connectors/inetsoft-rest` passes
- [x] Manually verified end-to-end: created a data source, switched to OAuth, authorized against a real Salesforce sandbox org, saved, and successfully tested the connection
- [x] Verified legacy Password auth mode is unaffected

Fixes Redmine #76329.

### DIFF
--- a/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/remedyforce/RemedyforceDataSource.java
+++ b/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/remedyforce/RemedyforceDataSource.java
@@ -18,27 +18,38 @@
 package inetsoft.uql.rest.datasource.remedyforce;
 
 import inetsoft.uql.rest.datasource.salesforce.SalesforceDataSource;
-import inetsoft.uql.tabular.View;
-import inetsoft.uql.tabular.View1;
-import inetsoft.util.credential.CredentialType;
+import inetsoft.uql.tabular.*;
 
 @View(vertical = true, value = {
    @View1(value = "useCredentialId", visibleMethod = "supportToggleCredential"),
    @View1(value = "credentialId", visibleMethod = "isUseCredentialId"),
-   @View1(value = "user", visibleMethod = "useCredential"),
-   @View1(value = "password", visibleMethod = "useCredential"),
-   @View1(value = "securityToken", visibleMethod = "useCredential"),
+   @View1("authType"),
+   @View1(value = "user", visibleMethod = "useCredentialForPassword"),
+   @View1(value = "password", visibleMethod = "useCredentialForPassword"),
+   @View1(value = "securityToken", visibleMethod = "useCredentialForPassword"),
+   @View1(value = "clientId", visibleMethod = "useCredentialForOauth"),
+   @View1(value = "clientSecret", visibleMethod = "useCredentialForOauth"),
+   @View1(type = ViewType.LABEL, text = "redirect.uri.description", colspan = 2,
+          visibleMethod = "useCredentialForOauth"),
+   @View1(
+      type = ViewType.BUTTON,
+      text = "Authorize",
+      visibleMethod = "isOauth",
+      button = @Button(
+         type = ButtonType.OAUTH, method = "updateTokens", oauth = @Button.OAuth,
+         dependsOn = { "clientId", "clientSecret", "credentialId" },
+         enabledMethod = "authorizeEnabled"
+      )
+   ),
+   @View1(value = "accessToken", visibleMethod = "isOauth"),
+   @View1(value = "refreshToken", visibleMethod = "isOauth"),
+   @View1(value = "instanceUrl", visibleMethod = "isOauth"),
 })
 public class RemedyforceDataSource extends SalesforceDataSource<RemedyforceDataSource> {
    static final String TYPE = "Rest.Remedyforce";
 
    public RemedyforceDataSource() {
       super(TYPE, RemedyforceDataSource.class);
-   }
-
-   @Override
-   protected CredentialType getCredentialType() {
-      return CredentialType.PASSWORD_SECURITY_TOKEN;
    }
 
    @Override
--- a/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/salesforce/SalesforceDataSource.java
+++ b/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/salesforce/SalesforceDataSource.java
@@ -55,6 +55,19 @@ public abstract class SalesforceDataSource<SELF extends SalesforceDataSource<SEL
       return CredentialType.PASSWORD_OAUTH2_WITH_FLAGS;
    }
 
+   /**
+    * The base {@link AbstractRestDataSource} implementation only allows a "Use Secret ID" cloud
+    * credential for {@code AuthType.BASIC}/{@code OAUTH}, but this connector's legacy mode is
+    * {@code AuthType.NONE}. Include it so Legacy Password mode keeps the cloud-secret support it
+    * had before this class used {@code PASSWORD_OAUTH2_WITH_FLAGS} (previously
+    * {@code PASSWORD_SECURITY_TOKEN}, which always allowed it).
+    */
+   @Override
+   protected boolean supportCredentialId() {
+      AuthType type = getAuthType();
+      return type == AuthType.NONE || type == AuthType.OAUTH;
+   }
+
    // Not part of the PASSWORD_OAUTH2_WITH_FLAGS credential, so stored as a plain
    // (manually encrypted) field, same as the inherited accessToken/refreshToken fields are
    // for credential types that don't carry them.
@@ -205,6 +218,18 @@ public abstract class SalesforceDataSource<SELF extends SalesforceDataSource<SEL
       instanceUrl = Tool.getChildValueByTagName(root, "instanceUrl");
       String token = Tool.getChildValueByTagName(root, "securityToken");
 
+      // Data sources saved before this class used PASSWORD_OAUTH2_WITH_FLAGS stored the
+      // security token nested inside the legacy <PasswordCredential> node (written by
+      // LocalSecurityTokenCredential, back when getCredentialType() was
+      // PASSWORD_SECURITY_TOKEN). Fall back to that location so upgrades don't drop it.
+      if(token == null) {
+         Element legacyCredential = Tool.getChildNodeByTagName(root, "PasswordCredential");
+
+         if(legacyCredential != null) {
+            token = Tool.getChildValueByTagName(legacyCredential, "securityToken");
+         }
+      }
+
       if(token != null) {
          securityToken = Tool.decryptPassword(token);
       }
@@ -224,12 +249,14 @@ public abstract class SalesforceDataSource<SELF extends SalesforceDataSource<SEL
          return false;
       }
 
-      return Objects.equals(securityToken, ((SalesforceDataSource<?>) o).securityToken);
+      SalesforceDataSource<?> that = (SalesforceDataSource<?>) o;
+      return Objects.equals(securityToken, that.securityToken) &&
+         Objects.equals(instanceUrl, that.instanceUrl);
    }
 
    @Override
    public int hashCode() {
-      return Objects.hash(super.hashCode(), getSecurityToken());
+      return Objects.hash(super.hashCode(), getSecurityToken(), instanceUrl);
    }
 
    private SalesforceSession getSession() {
@@ -332,7 +359,7 @@ public abstract class SalesforceDataSource<SELF extends SalesforceDataSource<SEL
       String instanceUrl = dataSource.getInstanceUrl();
       String accessToken = dataSource.getAccessToken();
 
-      if(instanceUrl == null || accessToken == null || accessToken.isEmpty()) {
+      if(instanceUrl == null || instanceUrl.isEmpty() || accessToken == null || accessToken.isEmpty()) {
          throw new RuntimeException(
             "Salesforce OAuth authorization has not been completed for this data source");
       }
--- a/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/salesforce/SalesforceDataSource.java
+++ b/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/salesforce/SalesforceDataSource.java
@@ -21,6 +21,7 @@ import com.github.benmanes.caffeine.cache.*;
 import inetsoft.uql.rest.auth.AuthType;
 import inetsoft.uql.rest.json.EndpointJsonDataSource;
 import inetsoft.uql.tabular.*;
+import inetsoft.uql.tabular.oauth.Tokens;
 import inetsoft.util.Tool;
 import inetsoft.util.credential.*;
 import org.apache.http.client.methods.CloseableHttpResponse;
@@ -49,14 +50,96 @@ public abstract class SalesforceDataSource<SELF extends SalesforceDataSource<SEL
       setAuthType(AuthType.NONE);
    }
 
+   @Override
+   protected CredentialType getCredentialType() {
+      return CredentialType.PASSWORD_OAUTH2_WITH_FLAGS;
+   }
+
+   // Not part of the PASSWORD_OAUTH2_WITH_FLAGS credential, so stored as a plain
+   // (manually encrypted) field, same as the inherited accessToken/refreshToken fields are
+   // for credential types that don't carry them.
    @Property(label = "Security Token", required = true, password = true)
-   @PropertyEditor(dependsOn = "useCredentialId")
    public String getSecurityToken() {
-      return ((SecurityTokenCredential) getCredential()).getSecurityToken();
+      return securityToken;
    }
 
    public void setSecurityToken(String securityToken) {
-      ((SecurityTokenCredential) getCredential()).setSecurityToken(securityToken);
+      this.securityToken = securityToken;
+   }
+
+   public boolean useCredentialForPassword() {
+      return useCredential() && !isOauth();
+   }
+
+   /**
+    * Restricts the inherited authentication type dropdown to the two modes this connector
+    * actually implements — {@link AuthType#BASIC}/{@code TWO_STEP}/{@code KERBEROS} are not
+    * supported here and would otherwise silently fall back to the legacy SOAP login.
+    */
+   @Override
+   @Property(label="Authentication")
+   @PropertyEditor(tags={"NONE", "OAUTH"}, labels={"Username/Password (Legacy)", "OAuth 2.0"})
+   public AuthType getAuthType() {
+      return super.getAuthType();
+   }
+
+   @Override
+   @PropertyEditor(enabled = false, dependsOn = "useCredentialId")
+   public String getAuthorizationUri() {
+      return getOAuthHostUrl() + "/services/oauth2/authorize";
+   }
+
+   @Override
+   public void setAuthorizationUri(String authorizationUri) {
+      // computed from the login URL; not directly editable
+   }
+
+   @Override
+   @PropertyEditor(enabled = false, dependsOn = "useCredentialId")
+   public String getTokenUri() {
+      return getOAuthHostUrl() + "/services/oauth2/token";
+   }
+
+   @Override
+   public void setTokenUri(String tokenUri) {
+      // computed from the login URL; not directly editable
+   }
+
+   @Override
+   @PropertyEditor(enabled = false, dependsOn = "useCredentialId")
+   public String getScope() {
+      String scope = super.getScope();
+      return scope == null || scope.isEmpty() ? "api refresh_token" : scope;
+   }
+
+   @Override
+   @PropertyEditor(enabled = false, dependsOn = "useCredentialId")
+   public String getOauthFlags() {
+      // Salesforce Connected/External Client Apps require PKCE for the Authorization Code flow.
+      return "pkce";
+   }
+
+   /**
+    * Captures the {@code instance_url} that Salesforce returns alongside the access/refresh
+    * tokens (a non-standard OAuth field), in addition to the standard token bookkeeping.
+    */
+   @Override
+   public void updateTokens(Tokens tokens) {
+      super.updateTokens(tokens);
+
+      if(tokens.properties() != null && tokens.properties().get("instance_url") != null) {
+         setInstanceUrl(String.valueOf(tokens.properties().get("instance_url")));
+      }
+   }
+
+   @Property(label = "Instance URL")
+   @PropertyEditor(enabled = false)
+   public String getInstanceUrl() {
+      return instanceUrl;
+   }
+
+   public void setInstanceUrl(String instanceUrl) {
+      this.instanceUrl = instanceUrl;
    }
 
    @Override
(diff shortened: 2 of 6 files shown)


jshobe-inetsoft (changes requested):
## Summary

This PR adds OAuth 2.0 Authorization Code support to `SalesforceDataSource`/`RemedyforceDataSource`/`SalesforceReportsDataSource` alongside the existing username/password SOAP login, ahead of Salesforce retiring the SOAP `login()` operation. The OAuth plumbing itself (PKCE, dynamic authorize/token URL derivation from `getLoginUrl()`, `createOAuthSession()`, `instance_url` capture) is well designed and reuses the existing generic OAuth framework in `AbstractRestDataSource` appropriately.

However, I found a **data-loss / backward-compatibility regression** in how the legacy credential is migrated that will break every existing saved data source of these three types. Requesting changes.

## 🔴 Blocking: existing `securityToken` is silently dropped on load, breaking all existing legacy connections

`getCredentialType()` changes from `CredentialType.PASSWORD_SECURITY_TOKEN` to `CredentialType.PASSWORD_OAUTH2_WITH_FLAGS` for `SalesforceDataSource` (and, by removing their overrides, for `RemedyforceDataSource`/`SalesforceReportsDataSource` too). Previously, the security token was part of the `SecurityTokenCredential` object and was written by `LocalSecurityTokenCredential.writeContent()` **as a child of the `<PasswordCredential>` element**:

```xml
<ds_Rest.Salesforce ...>
  <PasswordCredential class="inetsoft.util.credential.LocalSecurityTokenCredential">
    <user>...</user>
    <password>...</password>
    <securityToken>...</securityToken>
  </PasswordCredential>
  ...
</ds_Rest.Salesforce>
```

After this change, `securityToken` is a plain field, restored in the new `SalesforceDataSource.parseContents(Element root)`:

```java
String token = Tool.getChildValueByTagName(root, "securityToken");
```

`Tool.getChildValueByTagName` / `CoreTool.getChildNodeByTagName` only looks at **direct children** of the passed element (`elem.getChildNodes()`, no recursion). `root` here is the data source's own `<ds_Rest.Salesforce>` element, but the legacy `<securityToken>` tag is nested one level deeper, inside `<PasswordCredential>`. So for any data source persisted before this change:

1. `TabularDataSource` constructs a new `credential` from `getCredentialType()` → now a `LocalPasswordAndOAuth2WithFlagCredentialsGrant`, which has no concept of `securityToken`.
2. `TabularDataSource.parseContents()` finds the legacy `<PasswordCredential>` node and calls `credential.parseXML(credentialNode)` on this new object — it happily picks up `user`/`password` (tag names match), but has no field/handling for `securityToken`, so it's discarded.
3. `SalesforceDataSource.parseContents()`'s new code looks for `<securityToken>` directly under `root`, but it never existed there (only under `<PasswordCredential>`) — so it comes back `null` too.

Net effect: **`getSecurityToken()` returns `null` for every pre-existing Salesforce/Remedyforce/SalesforceReports data source**, and it happens silently — no exception, no log warning, no migration message.

Since the security token is required by Salesforce for API access from IPs outside the org's trusted range (the standard, most common deployment), this affects effectively all existing production data sources of these types the moment the server is upgraded to a build containing this PR — not just OAuth users. `createSession()` in `SalesforceDataSource` builds the legacy SOAP login body as:

```java
writer.writeCharacters(dataSource.getPassword() + dataSource.getSecurityToken());
```

With `getSecurityToken()` now `null`, this appends the literal string `"null"` to the password, so the login will fail (`INVALID_LOGIN`) for every existing customer relying on this auth mode, silently, with no indication of what changed. This directly contradicts the PR description's claim that "Legacy Password mode is completely unchanged for backward compatibility with existing saved data sources."

**Suggested fix:** in `SalesforceDataSource.parseContents()`, fall back to reading the legacy location when the new-style tag isn't present, e.g.:

```java
String token = Tool.getChildValueByTagName(root, "securityToken");

if(token == null) {
   Element legacyCredential = Tool.getChildNodeByTagName(root, "PasswordCredential");

   if(legacyCredential != null) {
      token = Tool.getChildValueByTagName(legacyCredential, "securityToken");
   }
}

if(token != null) {
   securityToken = Tool.decryptPassword(token);
}
```

(Note the legacy value would already be pre-encrypted the same way `Tool.encryptPassword`/`decryptPassword` handles it, since `LocalSecurityTokenCredential` used the same encryption helper — this should be a drop-in fix.) This should be covered by a unit test that round-trips a pre-PR-format XML fixture through `parseXML()` and asserts `getSecurityToken()`/`getUser()`/`getPassword()` are all preserved.

## 🟡 Secondary: "Use Secret ID" (cloud credential) support disappears for Legacy Password mode

`AbstractRestDataSource.supportCredentialId()` (unchanged by this PR) is:

```java
protected boolean supportCredentialId() {
   if(getCredentialType() != CredentialType.PASSWORD_OAUTH2_WITH_FLAGS) {
      return true;
   }
   AuthType type = getAuthType();
   return type == AuthType.BASIC || type == AuthType.OAUTH;
}
```

Before this PR, `getCredentialType()` for these three connectors was `PASSWORD_SECURITY_TOKEN`, so `supportCredentialId()` always returned `true` — customers could store the datasource's username/password/security token as a cloud "Secret ID" (AWS Secrets Manager/Azure Key Vault/GCP Secret Manager/etc.).

Now that `getCredentialType()` is `PASSWORD_OAUTH2_WITH_FLAGS`, `supportCredentialId()` only returns `true` for `AuthType.BASIC`/`AuthType.OAUTH`. Since Legacy Password mode is represented by `AuthType.NONE` (see below), `supportCredentialId()` now returns `false` in Legacy mode, which:
- hides the "Use Secret ID" toggle (`supportToggleCredential()` gates on it) for existing Legacy-mode data sources, and
- more importantly, `AbstractRestDataSource.setAuthType()` has this side effect:
  ```java
  public void setAuthType(AuthType authType) {
     this.authType = authType;
     if(!supportToggleCredential() && getCredential() instanceof CloudCredential) {
        setCredential(createCredential(true));
     }
  }
  ```
  Because `authType` is now a visible/editable property (`@View1("authType")` was added), any UI flow that round-trips the view's current values back into the bean (`TabularUtil.refreshView()`'s `setValuesToBean()` step, used both when opening and saving the data source editor) will call `setAuthType(NONE)` again even when nothing changed. With `supportToggleCredential()` now `false` for Legacy mode, this forcibly replaces an existing `CloudCredential` with a local one — i.e., a customer who stored their Salesforce credentials via a cloud secret manager could have that reference silently discarded just by opening/saving the data source in EM.

Please double check this path (I have not traced `refreshView`/`setValuesToBean` in enough depth to be 100% certain it's hit merely on open vs. only on explicit save, but the mechanism is real and worth a regression test either way). At minimum, `supportCredentialId()` should probably also allow `AuthType.NONE` for this connector family so Legacy mode keeps parity with pre-PR cloud-secret support.

## Minor / non-blocking

- **`RemedyforceDataSource` OAuth is hardcoded to production Salesforce.** `getOAuthHostUrl()` derives the authorize/token endpoints from `getLoginUrl()`, which `RemedyforceDataSource` never overrides (unlike `SalesforceReportsDataSource`, which switches to `test.salesforce.com` for sandbox `accountType`). This mirrors a pre-existing limitation of the SOAP login path, so it's not a new regression, but it does mean Remedyforce/BMC customers on a sandbox or custom "My Domain" org have no way to make the new OAuth flow (or the old SOAP login) target the right host. Worth a follow-up if such customers exist.
- **`equals()`/`hashCode()` don't include `instanceUrl`.** `SalesforceDataSource.equals()` was updated to compare `securityToken` but not the new `instanceUrl` field, and `hashCode()` (unchanged) also omits it. Not a contract violation, but a data source that only differs by `instanceUrl` (e.g. after a token refresh returns a different Salesforce instance) will compare as "equal," which may or may not matter depending on how equality is used for change-detection elsewhere. Worth confirming this is intentional.
- **No new tests.** The PR's test plan is manual verification only. Given the credential-type/migration risk above, at least one automated test asserting that a pre-PR-format saved XML still deserializes with `user`/`password`/`securityToken` intact would have caught the blocking issue.
- Naming: reusing `AuthType.NONE` to mean "Username/Password (Legacy)" via the `@PropertyEditor` label override is a little confusing to a future maintainer reading the enum value in isolation (`NONE` normally means "no authentication" elsewhere in this same class hierarchy), though the UI-facing label is clear. A short comment on `getAuthType()` explaining the reuse would help.

## Other review areas

- **Clean code / readability:** Otherwise generally clean; the new methods (`useCredentialForPassword`, `createOAuthSession`, `getOAuthHostUrl`) are small, well-named, and the two new doc comments explain non-obvious behavior appropriately.
- **Security:** PKCE is correctly forced on for the Authorization Code flow. Client secret/security token continue to use `password = true` masked fields and the existing `Tool.encryptPassword`/`decryptPassword` helpers. The shared `https://data.inetsoft.com/oauth` redirect URI matches the established pattern used by other OAuth connectors (HubSpot, GitHub, Box, etc.) in this same module, so it's not a new concern.
- **CI status:** All checks pass (Semgrep, Semgrep OSS, Validation Build, CLA, and `claude-review`). No merge conflicts (`mergeable: MERGEABLE`), though GitHub reports `mergeStateStatus: BLOCKED`, which appears to just require an approving review before merge (this branch is a working branch off the `v1.1.x` release branch per repo convention, not a direct commit to it, so that part of the process was followed correctly).

jkidacki-inetsoft:
Addressed in the latest commit:

- **Blocking**: `parseContents()` now falls back to reading `securityToken` from inside the legacy `<PasswordCredential>` node (the location `LocalSecurityTokenCredential` used to write it) when the new top-level element is absent, so existing saved Salesforce/Remedyforce/SalesforceReports data sources no longer lose their security token on upgrade. Added `RemedyforceDataSourceTest.legacySecurityTokenIsMigratedFromPasswordCredentialNode`, which round-trips a pre-PR-format XML fixture through `parseXML()` and asserts `user`/`password`/`securityToken` all survive.
- **Secondary**: overrode `supportCredentialId()` to also allow `AuthType.NONE` (Legacy Password mode), restoring the cloud "Use Secret ID" parity this connector family had before it switched to `PASSWORD_OAUTH2_WITH_FLAGS`.
- **Minor**: `equals()`/`hashCode()` now include `instanceUrl`.

Also cross-ported the same fix to #5145 (the `main` forward-port) so both stay in sync.

jshobe-inetsoft (changes requested):
## Summary

This PR adds an OAuth 2.0 Authorization Code flow (reusing `PASSWORD_OAUTH2_WITH_FLAGS`) to `SalesforceDataSource`/`RemedyforceDataSource`/`SalesforceReportsDataSource` as an alternative to the SOAP `login()` operation Salesforce is retiring. The overall design is sound and mirrors the established `AbstractRestDataSource` OAuth pattern (Authorize button → `updateTokens`, `getAuthorizationUri`/`getTokenUri` derived from the org's login URL, PKCE forced on via `getOauthFlags()`). The PR already went through one round of fixes (commit `ce5128de1`) addressing a legacy Security Token regression found in prior review, which is good process. However, that fix only covers half of the backward-compatibility surface — see the blocking issue below.

## Blocking issue

### 1. Security Token is silently lost for existing data sources that use a cloud "Secret ID" credential (backward compatibility / data loss)

Before this PR, `getCredentialType()` returned `PASSWORD_SECURITY_TOKEN`, and the `SecurityTokenCredential` (user + password + security token) was the credential itself. When an admin configured "Use Secret ID" (a cloud vault secret, e.g. AWS Secrets Manager), **all three fields — including the security token — lived in the vault**, serialized by `CloudSecurityTokenCredential.Serializer` under the JSON key `"security_token"` (`core/.../CloudSecurityTokenCredential.java`).

After this PR, `securityToken` is pulled out of the credential entirely and stored as a plain field on `SalesforceDataSource` (`connectors/inetsoft-rest/.../SalesforceDataSource.java:74-81`). The new credential type's cloud variant, `CloudPasswordAndOAuth2WithFlagCredentialsGrant`, has no `securityToken` property and its (de)serializer never looks at `"security_token"` — so that value is dropped when the vault secret is fetched.

The `parseContents` migration fallback added in the follow-up commit only rescues the *local* case:

```java
// SalesforceDataSource.java:219-231
String token = Tool.getChildValueByTagName(root, "securityToken");
if(token == null) {
   Element legacyCredential = Tool.getChildNodeByTagName(root, "PasswordCredential");
   if(legacyCredential != null) {
      token = Tool.getChildValueByTagName(legacyCredential, "securityToken");
   }
}
```

This reads the *local* datasource XML file. A cloud-secret-backed data source never wrote the security token into that XML (only `<PasswordCredential cloud="true" id="..."/>` is persisted locally — see `AbstractCloudCredential.writeXML`); the real value lives only in the vault under the old class's schema, which the new credential class doesn't know how to read.

**Compounding this**, the view for `securityToken` is now gated on `useCredentialForPassword()` (`useCredential() && !isOauth()`), and `useCredential()` is `false` whenever "Use Secret ID" is checked. So the field is also **hidden in the UI** whenever Secret ID mode is active — there is no way to (re)enter it after upgrading, and no way to set it at all when creating a *new* Secret-ID + Legacy-Password data source.

**Concrete failure scenario:** A customer has an existing Salesforce/Remedyforce/Salesforce Reports data source configured with "Use Secret ID" + Legacy Password auth, with a security token stored in the vault secret. After upgrading to this build:
- `getSecurityToken()` returns `null` (never populated from the vault, and the field to enter it manually is hidden).
- `createSession()`'s SOAP login builds the password as `dataSource.getPassword() + dataSource.getSecurityToken()`, which Java string-concatenates to `"<password>null"`.
- Salesforce authentication fails (or, if the org happens to have trusted IP ranges configured so a token isn't strictly required, it silently succeeds with a bogus trailing `"null"` that Salesforce should reject) — either way, this is a regression for any customer relying on IP-restricted Secret ID + Security Token today.

This directly contradicts the PR's stated goal ("Legacy Password mode is completely unchanged for backward compatibility with existing saved data sources") for the Secret-ID subset of users, and there's no test covering the cloud-credential migration path (only the local one is covered by the new `RemedyforceDataSourceTest`).

**Suggested fix:** either keep the security token as part of the credential object (e.g. extend the new grant type or its Local/Cloud implementations to also carry `securityToken`, so it round-trips through the vault like user/password do), or explicitly support it as a cloud-secret property with its own migration path from the old `"security_token"` JSON key.

## Other findings (non-blocking, but worth addressing)

### 2. OAuth access token is refreshed only on session-cache miss, not per request

Every other OAuth-based REST connector in this module calls `refreshTokens()` from `getQueryHttpParameters()`, i.e. on **every request** (see `HubSpotDataSource`, `ZohoCRMDataSource`, `BoxDataSource`, etc., and `OAuthAuthenticator.authenticateRequest`). `SalesforceDataSource.getQueryHttpParameters()` was left unchanged and instead depends on `getSession()`, which is backed by a `Caffeine` cache keyed by the data source itself with `expireAfterAccess(2, HOURS)` / `maximumSize(20)`. `refreshTokens()` is only invoked inside `createOAuthSession()`, which Caffeine only calls on a cache **miss**.

In practice this means: once a Salesforce OAuth session is cached, it will be reused as-is (using the same possibly-expired `accessToken`) for as long as the entry is actively touched (which resets the 2-hour `expireAfterAccess` timer indefinitely under normal polling/scheduled-report load), regardless of the actual Salesforce access-token TTL. If Salesforce's access token expires sooner than 2 hours of inactivity (common for Connected App tokens), requests will start failing with 401s and won't self-heal until the cache entry is evicted (either by 2h idle or by the 20-entry LRU cap), which may not happen for an actively-used data source.

Consider calling `refreshTokens()` in `getQueryHttpParameters()` for the OAuth path (consistent with every other connector), rather than only inside the cache loader.

### 3. Minor: repeated inline `/services/` URL-splitting logic

The `.../services/...` splitting logic to derive a server/host URL is duplicated three times: `getOAuthHostUrl()`, the SOAP `serverUrl` parsing in `createSession()`, and `createOAuthSession()`. Not a bug, but a small helper (e.g. `stripServicesSuffix(String url)`) would reduce duplication and the risk of the three copies drifting.

### 4. Minor: duplicated `@View` OAuth block across `RemedyforceDataSource` and `SalesforceReportsDataSource`

The clientId/clientSecret/redirect-description/Authorize-button/accessToken/refreshToken/instanceUrl `@View1` block is copy-pasted identically in both subclasses. This matches the pre-existing pattern in this codebase (each subclass already redeclares its own `user`/`password`/`securityToken` view rows), so it's not a new problem introduced by this PR, but if the code churns again it may be worth hoisting into a shared `@View` composition.

## What's good

- Restricting the `authType` dropdown to `NONE`/`OAUTH` (hiding `BASIC`/`TWO_STEP`/`KERBEROS`, which this connector doesn't implement) is a sensible, explicit choice with a clear comment.
- `supportCredentialId()` correctly preserves "Use Secret ID" support for the legacy `AuthType.NONE` mode, which the base `AbstractRestDataSource` implementation would otherwise have removed — good catch from the prior review round.
- PKCE is force-enabled via `getOauthFlags()`, matching Salesforce's requirement for Connected/External Client Apps.
- `getScope()`/`getAuthorizationUri()`/`getTokenUri()` are correctly derived per-org (including the `sandbox` account-type override in `SalesforceReportsDataSource`), rather than hardcoded.
- The added `RemedyforceDataSourceTest` is a real, targeted regression test for the local-credential migration path.

## CI / mergeability

- All required checks are green (`build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla`; `claude-review` also passed and is superseded by this manual review).
- `gh` reports the PR as `MERGEABLE` (no merge conflicts against `v1.1.x`); `mergeStateStatus: BLOCKED` appears to reflect a pending required-review gate rather than a conflict.

## Verdict

Requesting changes due to the blocking cloud-secret Security Token data-loss issue (#1). Recommend also addressing the token-refresh cache issue (#2) before merge, or explicitly confirming it's an acceptable known limitation.

jshobe-inetsoft (approved):
## Summary

This PR adds OAuth 2.0 Authorization Code support to `SalesforceDataSource` / `RemedyforceDataSource` / `SalesforceReportsDataSource`, alongside the existing SOAP `login()` username/password flow, ahead of Salesforce retiring SOAP login. This is the third review pass on this PR; the two prior reviews (2026-09-11 and 2026-09-14) each found a blocking backward-compatibility/data-loss regression in how the legacy `securityToken` is migrated. Both have since been fixed in follow-up commits, each with a dedicated regression test. I re-verified the fixes against the base classes (`AbstractRestDataSource`, `TabularDataSource`, `Tool`) rather than taking the commit messages at face value.

## Verification of previously-blocking issues

1. **Local-XML security token loss** (`<securityToken>` nested inside legacy `<PasswordCredential>`): `SalesforceDataSource.parseContents()` now falls back to `Tool.getChildNodeByTagName(root, "PasswordCredential")` → `getChildValueByTagName(..., "securityToken")` when the new top-level tag is absent. Covered by `RemedyforceDataSourceTest.legacySecurityTokenIsMigratedFromPasswordCredentialNode`. **Confirmed fixed.**
2. **Cloud "Use Secret ID" security token loss** (old `CloudSecurityTokenCredential` vault JSON schema, key `security_token`, which the new `PASSWORD_OAUTH2_WITH_FLAGS` cloud credential never reads): `parseContents()` now falls back to `loadCloudSecurityToken()` (→ `Tool.loadCredentials`, a real, existing API) when local lookup misses and the credential is a `CloudCredential`, gated on `!isOauth()` so the (billed) vault call isn't made for data sources that never had a token to migrate. Wrapped in try/catch matching the established pattern in `Tool.getClientSecretRealValue`. Covered by `legacySecurityTokenIsMigratedFromCloudSecret` and `oauthModeNeverConsultsTheVaultForALegacySecurityToken`. **Confirmed fixed.**
3. **`securityToken` field hidden when "Use Secret ID" was active**: visibility is now gated on `showSecurityToken() = !isOauth()` instead of `useCredentialForPassword()`, so it's visible/editable in Legacy mode regardless of credential source. Covered by `securityTokenStaysVisibleInLegacyModeRegardlessOfCredentialSource`. **Confirmed fixed.**
4. **`equals()`/`hashCode()` omitting `instanceUrl`** (flagged as a minor point previously): both now include `instanceUrl`. **Confirmed fixed.**

## New-code correctness checks

- `getOAuthHostUrl()` derives the authorize/token endpoints from `getLoginUrl()` by splitting on `/services/`, mirroring the exact same substring logic already used for the SOAP `serverUrl` parsing in `createSession()` — same edge-case handling (`index < 0` → `URI.resolve("/services")`), so behavior is consistent between the two paths. `SalesforceReportsDataSource.getLoginUrl()` already switches host for `accountType == "sandbox"`, so OAuth correctly targets `test.salesforce.com` there too.
- `createOAuthSession()` correctly builds the session from `instance_url` + access token instead of a SOAP envelope, matches the `SalesforceSession` shape expected by `getURL()`/`getQueryHttpParameters()`, and fails loudly (`RuntimeException`) if authorization hasn't been completed rather than silently producing a broken session.
- `updateTokens()` correctly layers `instance_url` capture on top of `super.updateTokens()` rather than replacing it.
- `supportCredentialId()` override correctly preserves "Use Secret ID" support for the repurposed `AuthType.NONE` ("Legacy Password") mode, which the base `AbstractRestDataSource` implementation would otherwise have dropped now that this class uses `PASSWORD_OAUTH2_WITH_FLAGS`.
- PKCE is forced on via `getOauthFlags()` — `"pkce"` is an established flag value already used by `TwitterDataSource`, not something invented for this PR.
- The `redirect.uri.description` text and `https://data.inetsoft.com/oauth` redirect URI match the exact convention used by every other OAuth connector in this module (HubSpot, GitHub, Box, ODataDataSource, etc.).
- `getAuthType()`/`authorizeEnabled()`/`useCredentialForOauth()` etc. all resolve to real inherited methods on `TabularDataSource`/`AbstractRestDataSource` — no invented API surface.
- Build (`gh pr checks`) is green, including the full Validation Build, so this all actually compiles and the existing test suite passes.

## Non-blocking observations (carried over / new)

- **OAuth session cache never proactively refreshes.** `getSession()` is backed by a `Caffeine` cache (`expireAfterAccess(2h)`, `maximumSize(20)`) keyed by the data source itself; `refreshTokens()` only runs inside `createOAuthSession()`, i.e. only on a cache **miss**. Every other OAuth connector in this module (HubSpot, Box, Zoho, etc.) calls `refreshTokens()` on every request via `getQueryHttpParameters()`. If a Salesforce access token expires sooner than 2 hours of continuous idle time, requests will 401 until the cache entry is evicted. This mirrors the pre-existing SOAP-session caching behavior (not a new pattern), so it's not a regression, but worth a follow-up since OAuth token lifetimes are typically shorter/more variable than the SOAP session timeout policy.
- **`RemedyforceDataSource` has no sandbox/custom-domain login URL override**, unlike `SalesforceReportsDataSource`'s `accountType`-driven switch. OAuth (and the pre-existing SOAP login) is hardcoded to `login.salesforce.com` for this connector. Pre-existing limitation, not introduced by this PR, but the PR description's framing ("production/sandbox/custom domain") slightly overstates what's actually configurable for Remedyforce specifically.
- **Minor duplication**: the `/services/` URL-splitting logic appears three times (`getOAuthHostUrl()`, SOAP `serverUrl` parsing, `createOAuthSession()`); and the `@View` OAuth block is copy-pasted identically between `RemedyforceDataSource` and `SalesforceReportsDataSource`. Both match pre-existing conventions elsewhere in this file/module, so not a new problem, just an opportunity.
- No unit tests for `createOAuthSession()`/`getOAuthHostUrl()` URL derivation itself (only the credential-migration paths are tested), though this would require HTTP/session mocking and manual end-to-end verification against a real sandbox org was done per the PR description.
- Several newly-visible fields (`Authentication`, `Client ID/Secret`, `Access Token/Refresh Token`, `Instance URL`) aren't added to the connector-specific `Bundle.properties` files the way `Client ID`/`Client Secret` are in e.g. `GitHubDataSource`'s bundle. `Catalog.getString()` falls back to the literal key text when a translation is missing (verified in `Catalog.java`), so this is a cosmetic/i18n completeness gap only, not a functional bug.

## Security

PKCE is correctly forced for the Authorization Code flow, client secret/security token continue to use masked (`password = true`) fields with `Tool.encryptPassword`/`decryptPassword`, and the vault-read fallback reuses an existing, already-authorized credential-lookup API rather than introducing new secret handling.

## Verdict

The blocking backward-compatibility issues from the two prior review rounds are verified fixed, each with a targeted regression test. No new blocking issues found. Remaining points above are follow-up-worthy but non-blocking. CI is green (`build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla`, `claude-review`), and the PR is `MERGEABLE` against `v1.1.x` with no conflicts — this is a working branch off the release branch per repo convention, not a direct commit to it.

**Approving.**

PR #5145, opened 2026-09-10
Bug #76329, add OAuth 2.0 support to Salesforce REST connectors

jkidacki-inetsoft:
## Summary

Cherry-pick of the same fix from #5144 (targeting `v1.1.x`), forward-ported to `main`.

Salesforce is retiring the SOAP `login()` API operation (Summer '27). This adds OAuth 2.0 Authorization Code flow as an alternative authentication method for `SalesforceDataSource`, `RemedyforceDataSource`, and `SalesforceReportsDataSource`, alongside the existing username/password login for backward compatibility.

## Root Cause

These connectors authenticate exclusively via the SOAP `login()` operation, which Salesforce is fully retiring. Once retired, there will be no admin workaround — connections will simply fail.

## Fix

- Added an `Authentication` toggle (Password / OAuth 2.0) to each connector's data source configuration, reusing the existing `PASSWORD_OAUTH2_WITH_FLAGS` credential type
- OAuth uses the customer's own Salesforce Connected App / External Client App (Client ID + Secret typed into the data source UI), since Salesforce requires org-specific login URLs (production/sandbox/custom domain) that a single shared app can't support
- PKCE is required by Salesforce's OAuth implementation and is now enabled automatically
- `SalesforceDataSource.createSession()` branches: OAuth mode builds the session directly from the access token + Salesforce's `instance_url` (captured from the non-standard OAuth token response field), skipping the SOAP login envelope entirely
- Legacy Password mode is completely unchanged for backward compatibility with existing saved data sources

## Test Plan

- [x] `./mvnw test -pl core,connectors/inetsoft-rest` passes
- [x] Manually verified end-to-end: created a data source, switched to OAuth, authorized against a real Salesforce sandbox org, saved, and successfully tested the connection
- [x] Verified legacy Password auth mode is unaffected
- [x] Cherry-pick applied cleanly and recompiled against `main`

Fixes Redmine #76329.

### DIFF
--- a/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/remedyforce/RemedyforceDataSource.java
+++ b/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/remedyforce/RemedyforceDataSource.java
@@ -18,27 +18,38 @@
 package inetsoft.uql.rest.datasource.remedyforce;
 
 import inetsoft.uql.rest.datasource.salesforce.SalesforceDataSource;
-import inetsoft.uql.tabular.View;
-import inetsoft.uql.tabular.View1;
-import inetsoft.util.credential.CredentialType;
+import inetsoft.uql.tabular.*;
 
 @View(vertical = true, value = {
    @View1(value = "useCredentialId", visibleMethod = "supportToggleCredential"),
    @View1(value = "credentialId", visibleMethod = "isUseCredentialId"),
-   @View1(value = "user", visibleMethod = "useCredential"),
-   @View1(value = "password", visibleMethod = "useCredential"),
-   @View1(value = "securityToken", visibleMethod = "useCredential"),
+   @View1("authType"),
+   @View1(value = "user", visibleMethod = "useCredentialForPassword"),
+   @View1(value = "password", visibleMethod = "useCredentialForPassword"),
+   @View1(value = "securityToken", visibleMethod = "useCredentialForPassword"),
+   @View1(value = "clientId", visibleMethod = "useCredentialForOauth"),
+   @View1(value = "clientSecret", visibleMethod = "useCredentialForOauth"),
+   @View1(type = ViewType.LABEL, text = "redirect.uri.description", colspan = 2,
+          visibleMethod = "useCredentialForOauth"),
+   @View1(
+      type = ViewType.BUTTON,
+      text = "Authorize",
+      visibleMethod = "isOauth",
+      button = @Button(
+         type = ButtonType.OAUTH, method = "updateTokens", oauth = @Button.OAuth,
+         dependsOn = { "clientId", "clientSecret", "credentialId" },
+         enabledMethod = "authorizeEnabled"
+      )
+   ),
+   @View1(value = "accessToken", visibleMethod = "isOauth"),
+   @View1(value = "refreshToken", visibleMethod = "isOauth"),
+   @View1(value = "instanceUrl", visibleMethod = "isOauth"),
 })
 public class RemedyforceDataSource extends SalesforceDataSource<RemedyforceDataSource> {
    static final String TYPE = "Rest.Remedyforce";
 
    public RemedyforceDataSource() {
       super(TYPE, RemedyforceDataSource.class);
-   }
-
-   @Override
-   protected CredentialType getCredentialType() {
-      return CredentialType.PASSWORD_SECURITY_TOKEN;
    }
 
    @Override
--- a/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/salesforce/SalesforceDataSource.java
+++ b/connectors/inetsoft-rest/src/main/java/inetsoft/uql/rest/datasource/salesforce/SalesforceDataSource.java
@@ -55,6 +55,19 @@ public abstract class SalesforceDataSource<SELF extends SalesforceDataSource<SEL
       return CredentialType.PASSWORD_OAUTH2_WITH_FLAGS;
    }
 
+   /**
+    * The base {@link AbstractRestDataSource} implementation only allows a "Use Secret ID" cloud
+    * credential for {@code AuthType.BASIC}/{@code OAUTH}, but this connector's legacy mode is
+    * {@code AuthType.NONE}. Include it so Legacy Password mode keeps the cloud-secret support it
+    * had before this class used {@code PASSWORD_OAUTH2_WITH_FLAGS} (previously
+    * {@code PASSWORD_SECURITY_TOKEN}, which always allowed it).
+    */
+   @Override
+   protected boolean supportCredentialId() {
+      AuthType type = getAuthType();
+      return type == AuthType.NONE || type == AuthType.OAUTH;
+   }
+
    // Not part of the PASSWORD_OAUTH2_WITH_FLAGS credential, so stored as a plain
    // (manually encrypted) field, same as the inherited accessToken/refreshToken fields are
    // for credential types that don't carry them.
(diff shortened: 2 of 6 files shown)


jshobe-inetsoft (changes requested):
Requesting changes based on Claude's automated review, which found a blocking (High severity) issue: after this change, `getCredentialType()` for `RemedyforceDataSource`/`SalesforceReportsDataSource` moves to the shared base class returning `PASSWORD_OAUTH2_WITH_FLAGS`, and the new credential type's `parseXML()` no longer reads `securityToken` from inside the legacy `<PasswordCredential>` XML node — only as a new top-level element. Any existing saved Salesforce/Remedyforce/SalesforceReports data source will silently lose its Security Token on upgrade until an admin notices and re-enters it. Please add a fallback in `parseContents()` to read the legacy in-`PasswordCredential` location when the new top-level `securityToken` is absent, and add test coverage for the upgrade path. See the full review comment for details, including two other minor/non-blocking notes.

jkidacki-inetsoft:
Addressed in the latest commit (cross-ported from #5144, which had the same review finding):

- **Blocking**: `parseContents()` now falls back to reading `securityToken` from inside the legacy `<PasswordCredential>` node (written by the old `LocalSecurityTokenCredential`) when the new top-level element is absent, so existing saved Salesforce/Remedyforce/SalesforceReports data sources no longer lose their security token on upgrade. Added `RemedyforceDataSourceTest.legacySecurityTokenIsMigratedFromPasswordCredentialNode`, which round-trips a pre-PR-format XML fixture through `parseXML()` and asserts `user`/`password`/`securityToken` all survive.
- **Minor nit**: `createOAuthSession`'s guard now checks `instanceUrl.isEmpty()` in addition to `accessToken.isEmpty()`, for consistency.
- Also overrode `supportCredentialId()` to allow `AuthType.NONE` (Legacy Password mode) so it keeps cloud "Use Secret ID" parity with pre-PR behavior, and included `instanceUrl` in `equals()`/`hashCode()`.

jshobe-inetsoft (approved):
## Manual review (replacing `claude-review`)

The automated `claude-review` check's most recent run failed as an infrastructure/quota error in the bot itself (`is_error: true`, 0s, "I'll analyze this and get back to you" with no follow-up) — it never produced a review of the current code. This is a full manual review of the current head commit (`1593324`) in its place.

### Context

This PR adds OAuth 2.0 Authorization Code (PKCE) support to `SalesforceDataSource` and its subclasses `RemedyforceDataSource` and `SalesforceReportsDataSource`, alongside the existing username/password + security-token login, ahead of Salesforce retiring the SOAP `login()` operation. It is a forward-port of #5144.

An earlier automated review on this PR (2026-09-10) correctly flagged a **high-severity backward-compatibility bug**: moving `getCredentialType()` from `PASSWORD_SECURITY_TOKEN` to the shared `PASSWORD_OAUTH2_WITH_FLAGS` meant `parseContents()` would no longer find `securityToken` in its old on-disk location (inside the legacy `<PasswordCredential>` node written by `LocalSecurityTokenCredential`), silently dropping the security token for every previously-saved Salesforce/Remedyforce/SalesforceReports data source on upgrade. A human reviewer (`jshobe-inetsoft`) requested changes on that basis.

The latest commit (`Bug #76329, fix legacy Security Token/cloud-secret regressions from code review`) addresses this directly, and I independently verified the fix:

- `SalesforceDataSource.parseContents()` now reads a top-level `<securityToken>` element first (the new format), and if absent, falls back to reading it from inside the legacy `<PasswordCredential>` node (`Tool.getChildNodeByTagName(root, "PasswordCredential")` → `securityToken` child) — exactly matching where the old `LocalSecurityTokenCredential.parseXML()` used to write it. `user`/`password` are unaffected either way since both old and new credential types persist those under the same `<PasswordCredential>`/`user`/`password` tags via the shared credential base class.
- New test `RemedyforceDataSourceTest.legacySecurityTokenIsMigratedFromPasswordCredentialNode` round-trips a pre-PR-format XML fixture through `parseXML()` and asserts `user`/`password`/`securityToken` all survive. I confirmed the Spring test wiring (`BaseTestConfiguration` + `ConfigurationContextInitializer` + `@SreeHome`, with a real `LocalPasswordAndOAuth2WithFlagCredentialsGrant` bean rather than the mocked `PASSWORD` stub `CredentialTestConfig` provides) is a correct, already-established pattern in this test module.
- The previously-flagged minor nit (`createOAuthSession`'s `instanceUrl` null/empty check) is also fixed — it now checks `isEmpty()` in addition to `null`, consistent with the `accessToken` check.
- `supportCredentialId()` is now overridden to permit `AuthType.NONE` (legacy Password mode), preserving pre-PR "Use Secret ID" cloud-credential support that would otherwise have been lost because the base `AbstractRestDataSource.supportCredentialId()` only allows `BASIC`/`OAUTH` for the `PASSWORD_OAUTH2_WITH_FLAGS` credential type.
- `instanceUrl` was added to `equals()`/`hashCode()` alongside `securityToken`, and both subclasses correctly delegate via `super.equals()`/`super.hashCode()`.

### Independent verification performed

- Traced the full credential/view framework this PR builds on (`AbstractRestDataSource`, `TabularDataSource`, `LocalPasswordAndOAuth2WithFlagCredentialsGrant`, `Tokens`/`AuthorizationClient`) to confirm `clientId`/`clientSecret`/`accessToken`/`refreshToken` are correctly credential-backed, `authorizeEnabled()`/`isOauth()`/`useCredentialForOauth()` are inherited (not newly invented) base-class hooks, and `Tokens.properties()` is populated from arbitrary extra fields in the token endpoint's JSON response — so capturing Salesforce's non-standard `instance_url` via `updateTokens()` overriding is sound.
- Compared against an existing OAuth connector (`GitHubDataSource`) that follows the same "hardcode `getAuthorizationUri()`/`getTokenUri()`/`getScope()` via getter override, omit from `@View`" pattern — this PR's approach is consistent with established codebase conventions, not a one-off.
- Verified `getOAuthHostUrl()` (derives the OAuth authorize/token host from `getLoginUrl()`) correctly picks up `SalesforceReportsDataSource`'s sandbox-vs-production `getLoginUrl()` override via virtual dispatch.
- Confirmed the `@View` redefinition in `RemedyforceDataSource`/`SalesforceReportsDataSource` (which drops `authorizationUri`/`tokenUri`/`scope`/`oauthFlags` from the visible form) is intentional/harmless: those getters are still invoked dynamically by the `Authorize` button's OAuth flow via reflection, they just aren't rendered as editable fields — again matching `GitHubDataSource`.
- Confirmed legacy Password-mode behavior is untouched: `createSession()` only branches into `createOAuthSession()` when `AuthType.OAUTH` is selected; the original SOAP `login()` path is unmodified.
- Reviewed the one remaining informational (non-blocking) note from the earlier automated review — that only `SalesforceReportsDataSource` exposes a sandbox/custom-domain toggle, so OAuth on the base/`RemedyforceDataSource` connectors is limited to the default `login.salesforce.com` domain — and confirmed this is a pre-existing limitation of `getLoginUrl()`, not a regression introduced by this PR.
- No new security concerns: `clientSecret`/`securityToken` continue to go through the framework's standard `Tool.encryptPassword`/`decryptPassword` and credential-vault mechanisms; nothing is logged or persisted in plaintext.

### Checks / mergeability

- `build` (Validation Build): ✅ success on the current head commit
- `Semgrep Scan` / `Semgrep OSS`: ✅ success
- `license/cla`: ✅ success
- `claude-review`: ❌ infra/quota error in the bot itself, not a real review finding — disregarded per manual-review policy
- No git merge conflicts (`MERGEABLE`)

### Verdict

**Approve.** The previously-identified blocking issue (silent Security Token loss on upgrade) and its accompanying minor nit are both fixed and covered by a new regression test. My own independent review of the OAuth session logic, credential/view framework integration, backward compatibility, and security handling found no further blocking issues.

PR #5148, opened 2026-09-10
fix(#76536): widen two more DataVSAssembly narrow-casts to DynamicBindableVSAssembly

fredaan:
## Summary

Fixes Redmine bug #76536 — two more `instanceof DataVSAssembly` narrow-cast sites (4th/5th occurrence of this class of defect) that excluded Gauge/Text (`OutputVSAssembly`) from pre-condition handling, matching the pattern already fixed twice in this file family (commits `190db1b3e` / PR #4606, and `b6be49907`).

- `core/src/main/java/inetsoft/web/wiz/service/WizVsService.java` (`carryCondition` block, ~line 2220): when replacing a filtered Gauge/Text chart in place (or type-changing away from it), the `instanceof DataVSAssembly` guard evaluated false for the Output assembly, so the whole condition-carry block was skipped — the filter silently vanished instead of carrying onto the new assembly.
- `core/src/main/java/inetsoft/web/wiz/controller/ViewsheetRuntimeController.java` (`hasPreCondition`, ~line 313): a saved Gauge/Text with a real, non-empty pre-condition was unconditionally reported as `hasCondition=false` on reopen — since `apply_filter` uses REPLACE semantics, a caller trusting that false could silently erase a real filter.

Both `DataVSAssembly` and `OutputVSAssembly` are sibling hierarchies that independently implement `DynamicBindableVSAssembly`, the interface `getPreConditionList()`/`setPreConditionList()` actually live on. Both sites only ever call those two accessors, so widening the cast to `DynamicBindableVSAssembly` is type-safe and matches the exact type/null-handling convention the two prior fixes already established in this same file.

## Test plan

- [x] Added `WizVsServiceCarryConditionOutputAssemblyTest` (site 1) — drives the real replace-in-place path through `WizVsService#createViewsheet` with a real `GaugeVSAssembly` carrying a non-empty `PreConditionList`, asserts the condition is carried onto the chart that replaces it.
- [x] Added `ViewsheetRuntimeControllerHasPreConditionTest` (site 2) — drives `ViewsheetRuntimeController#openViewsheet` with a real filtered Gauge, asserts `OpenViewsheetResult.hasCondition == true` (plus a companion unfiltered-Gauge case asserting `false`, to guard against over-widening).
- [x] Both new tests confirmed to **fail** against the pre-fix code (`AssertionFailedError`) and **pass** after the fix — closes the one gap the diagnosis/refute review explicitly flagged (neither had executed these tests, only read/compiled).
- [x] Ran via `./mvnw -pl core -Dtest=WizVsServiceCarryConditionOutputAssemblyTest,ViewsheetRuntimeControllerHasPreConditionTest test`: `Tests run: 3, Failures: 0, Errors: 0`.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/controller/ViewsheetRuntimeController.java
+++ b/core/src/main/java/inetsoft/web/wiz/controller/ViewsheetRuntimeController.java
@@ -29,7 +29,7 @@ import inetsoft.uql.asset.AssetEntry;
 import inetsoft.uql.ConditionList;
 import inetsoft.uql.viewsheet.ChartVSAssembly;
 import inetsoft.uql.viewsheet.OutputVSAssembly;
-import inetsoft.uql.viewsheet.DataVSAssembly;
+import inetsoft.uql.viewsheet.DynamicBindableVSAssembly;
 import inetsoft.uql.viewsheet.VSAssembly;
 import inetsoft.uql.viewsheet.Viewsheet;
 import inetsoft.util.Catalog;
@@ -311,7 +311,7 @@ public class ViewsheetRuntimeController {
     * filter is THERE is enough to stop the caller from destroying it unknowingly.
     */
    private boolean hasPreCondition(VSAssembly assembly) {
-      if(!(assembly instanceof DataVSAssembly dataAsm)) {
+      if(!(assembly instanceof DynamicBindableVSAssembly dataAsm)) {
          return false;
       }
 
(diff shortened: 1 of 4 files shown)


fredaan:
Review complete: clean, no blocking issues. Widened both instanceof DataVSAssembly casts to DynamicBindableVSAssembly at WizVsService.java:2221-2222 and ViewsheetRuntimeController.java:314, matching the confirmed sibling fix (190db1b3e). Independently verified no other instanceof DataVSAssembly occurrence in these two files was missed (974/4279 are pre-flagged out-of-scope future-sweep candidates). Regression tests have real executed fail-before/pass-after evidence (2 failures -> 0 after fix). `build` check failure is a pre-existing, unrelated base-branch defect (Angular TS2554 in formula-editor-dialog.component.spec.ts:349, already tracked by open PR #5139) — confirmed directly from the raw job log, not caused by this diff. (Self-authored PR — GitHub blocks self-approval via gh pr review; recording approval here instead.)

PR #5150, opened 2026-09-10
Modern visualization: the chart card track, seeded per assembly and on by default

FigurePaper:
The accumulated chart-card / modern-visualization work, feature-complete per its roadmap
and merging into `epic-74519`. 110 commits ahead, 0 behind.

## How it is gated

`viewsheet.modernVisualization` is no longer read at render time. Each assembly records a
`VizMark` when it is created, and rendering follows that mark through a `VizContext` rather
than the org property. Consequences worth knowing before review:

- `defaults.properties` now ships `viewsheet.modernvisualization=true` and
  `viewsheet.density=compact`, so a new install is modern and compact out of the box.
- Turning the org property off governs newly created dashboards only. Dashboards whose
  assemblies were created under the gate keep the modern look.
- Per-dashboard escape hatches in the composer: *Modernize Dashboard* and *Revert Dashboard
  to Classic Look*.
- Chrome values are seeded at creation by `seedChromeDefaults` instead of resolved at read
  time, so an asset exported into an older build carries its own values rather than
  rendering half-modern.

## What is in it

- Token contract, density tokens and org-scoped density, applied to table row heights and
  selection-list cell heights, with a Visualization Density control in EM Look and Feel.
- Modern widget-state vocabulary, adopted on viewsheet table overlays and migrated to a
  style layer for table structure, crosstab headers and totals.
- In-graph chart chrome: gridlines, facets, legend border, axis line unification, axis and
  legend text colors, and the card tooltip as the default, with its tail pointed at the
  hovered mark.
- Object title bar chrome, KPI/control chrome, combo dropdown and slider tokens.
- Gated modern categorical palette, palettes declared in CSS and served to the color
  pickers, and a state palette exposed for customer theming.
- Icon consolidation, org-scoped dark mode, and seeded dark chrome for the chart interior
  (plot structural line colors, data-label ink).
- The anchored mini-toolbar strip in the title lane across the chart, table, selection,
  calendar and selection-container families, drawn as bare glyphs toned to the card, sized
  from the assembly's own lane, and suppressed where the lane cannot hold it.
- Title lane drawn unfilled with a bottom rule; `userTitleHeight` records whether an author
  set a title height, with follow-the-default-density checkboxes replacing the value
  comparison the dialogs used to infer authorship from.
- Form inputs: seeded card corner radius, control height from density, checkbox chrome and
  font sizing aligned with its siblings, and an inset header on CheckBox/RadioButton
  instead of a floating legend.
- Fix: right-click now reaches max mode on tables.

## Design record

Entry point is `docs/superpowers/specs/lookfeel/chart-card-roadmap.md`, which carries the
dependency picture and points at every other document in the tree. Two traps it exists to
prevent: the external design set under `chart-card-design3/` is regenerated wholesale on
each sync, so notes written inside it are destroyed and decisions live in sibling files
outside it; and `seeded-value-reversibility-decisions.md` supersedes
`Visualization Widget Spec.dc.html` section 03.

## Diff shape

520 files, +102,514 / -1,028. Most of that is the design and decision record: 73,949
insertions across 140 files under `docs/superpowers`. Code is 28,565 insertions across 380
files, 193 in `core/src` and 182 in `web/projects`. 68 Java test files and 48 Angular spec
files are touched, 74 of them new.

## Verification

Test suites were run per phase during development (core, portal, em, and a cross-module
`-Pcommunity,enterprise` build). They have not been re-run against the current tip
`e02476d1` — worth a full run before merging.


FigurePaper:
**Item 1 — the categorical colour provenance change is deliberately ungated.**

Gating it on the mark re-opens the defect it was written to close. Revert clears the derived colours; the next render of the now-unmarked chart would write them back as static; a later Modernize's `clearDerivedColors()` cannot remove a static colour, so the re-modernized chart would render the legacy palette — which is exactly the "colour picker shows one palette, plot shows another" case this fixed.

Existing legacy content is not reassigned. A persisted colour parses back through `setColor`, never `setDerivedColor`, so it stays static and frozen exactly as before. What changes for a legacy sheet is only that a render stops freezing *new* per-value colours into the asset. That is pinned deliberately on a legacy chart by B3 of the palette manual-test plan. The decision and its reasoning are now written into the roadmap's colour notes rather than living only in the commit message.

**Item 2** — added as `ChromeSeedClassificationTest`, though not by the mechanism suggested: `setRoundCornerValue`/`setBorderColorsValue` are `VSFormat` methods that subtypes call rather than override, so there are no overrides to scan for. It instead scans the package for every concrete `VSAssemblyInfo` (41 of them) and asserts the invariant the predicates exist to protect — modernize-parity, revert-parity and idempotence — so a new subtype is covered automatically and fails if left out of the predicate it belongs to. All three lists are correct as written; the sweep found no misclassification.

**Items 3 and 4** — both fixed. On 4, the cost is per-cell rather than per-query: `findSpec` walks the spec list for every cell and attribute, so the specs are now skipped entirely unless the lens carries a crosstab, resolved with the same `Util.getCrosstab` call `XTableStyle.setTable` makes.

Full `core` suite green at 5729 tests.

FigurePaper:
## Frontend review — `web/projects` (182 files, +12,022 / −664)

Scope note first, since the two automated passes both excluded this area: this review covers only the Angular/TypeScript/SCSS changes. Read closely: `abstract-vs-actions.ts`, `chart-actions.ts`, `mini-toolbar.service.ts`, `tooltip-tail-placement.ts`, `chart-inline-svg.directive.ts`, `vs-spinner.component.ts`, `toolbar-actions-handler.ts`. Scanned mechanically across the whole frontend diff: XSS-risk sinks, timer/listener lifecycle, DI wiring. Not reviewed: the 28 SCSS files beyond the token definitions, and most templates.

**Verdict: changes requested** — one introduced defect, below.

### 1. Two tests introduced by this PR fail — `vs-spinner.component.tl.spec.ts`

```
VSSpinner — onIncrementClick / onDecrementClick > should step up by increment, emit, and mark an unapplied change
VSSpinner — startHold / cancelHold > should step once, emit, and mark an unapplied change after the initial delay
AssertionError: expected false to be true
```

Both assert `unappliedChange === true` immediately after a step. The step calls `commitChange()`, which sets the flag and then calls `changeValue()`; the spec's `debounceService.debounce` mock invokes its callback synchronously, so `changeValue0()` runs inline and clears the flag before the assertion reads it. Both `it(...)` blocks are added by this PR, so this is not pre-existing.

The product behaviour is fine — `commitChange`'s dispatch condition is identical to the `onClick` it replaced. The assertions are what is wrong. Either drop the `unappliedChange` expectation from these two cases (the surrounding cases already cover the flag where no flush occurs), or make the mock defer.

### 2. `clampToEdge` can hand `buildChromePaths` an offset it cannot draw — latent

`clampToEdge()` centres the tail when the edge is too short, returning `edge / 2`. `buildChromePaths()` assumes the offset is at least `tailRadius() + TAIL_HALF_WIDTH`; below that the opening overlaps the corner arcs and the `L` commands reverse direction, so the outline self-intersects at both corners. Threshold is `2 × (r + 6.7)` = 29.4px gate-off, 25.4px modern.

Not reachable today — `.widget__card-tooltip`'s `12px/16px` padding plus border floors a one-line box near 42px, roughly 12px of headroom. It is also untested: `buildChromePaths` is only exercised at 200×100 with offset 50, well inside the safe band. Worth either clamping the centred value into the drawable range or asserting the invariant, since the margin is a styling decision away from disappearing.

### 3. The portal TL suite runs nowhere

`npm run test` is `test:portal && test:em && test:em:tl` — `test:portal:tl` is absent, and `pr-build.yml` has no separate step for it. This PR adds or rewrites 14 `*.tl.spec.ts` files under `projects/portal`. Running them scoped gives **30 failures across 4 files** out of 432 tests.

28 of those 30 are pre-existing rather than introduced, and I verified each against the base:

- `binding-editor.component.display.tl.spec.ts` — 26 failures, all `NG0201: No provider found for ChartEditorService`. The base branch has 27 of the same; this PR removes one test, so it improves the count.
- `portal/app.component.risk.tl.spec.ts` — 1 failure. On the base this file does not compile at all (`setTabOrder` / `dataTabFirst` do not exist on the component); this PR deletes those cases, so again a net improvement.
- `date-value-editor.component.tl.spec.ts` — 1 failure, `"2023-12-01"` parsing to `{2023, 11, 30}`. Timezone-dependent, reproduced at UTC-4; the component is untouched by this PR and would likely pass in a UTC runner.

The suite is worth wiring into CI in its own right, but that is a separate change from this PR. Flagging it because it is why item 1 was invisible.

### Checked and found clean

Recording these so they do not get re-investigated:

- **The `hideMiniToolbarAction` handoff** between `createToolbarActions` and `createMenuActions` is order-dependent by design, but safe: `updateModel()` nulls both caches together, and `resetAssemblyMenuActions()` rebuilds the menu from an empty array, so the action cannot be duplicated or stranded.
- **`clearBrush` / `clearZoom` gated off the toolbar** under the mark. They remain reachable — `createMenuActions` carries both, ungated, and the ids match so `flattenedMoreActions()` dedup does not drop them. Covered by "drops the duplicated clear entries from the toolbar under the gate".
- **`legacyOrder`** is byte-for-byte the original emission order plus `propertiesToolbar`, whose predicate requires the mark. No gate-off reordering.
- **Spinner overshoot** — `stepValue` does not clamp, but `commitChange` calls `validate()`, which does.
- **`chart-inline-svg.directive.ts`** — the new raster branch revokes its object URL before each replacement and in `ngOnDestroy`. The `innerHTML` write is same-origin server-rendered SVG with the constraint documented at the assignment. `getElementAnchor` guards its `parseFloat` fallbacks.
- **No XSS sinks introduced** — no `bypassSecurityTrust*`, `eval`, or `document.write` anywhere in the diff.
- **Non-TL portal specs pass**: 557 tests green across `vsobjects/action`, `widget/tooltip`, `mini-toolbar`, `spinner`, `graph`, `color-picker`.

### Two notes, no action needed

`isInputDisabled()` reads `this.model.vizModern` unguarded and `this.model?.editing` optionally in the same expression — one or the other. And `validate()` compares against `this.model.max` without a null check, so a null max clamps the value to null (`5 > null` is `true`), while the new `isIncrementDisabled` guards `!= null` correctly. Both pre-existing; the second means the file now holds two different null policies.

### On merge readiness

The PR is still a draft, so `build` and `Semgrep Scan` are skipped — `pr-build.yml` and `semgrep.yml` are both gated on `!draft` and both trigger on `ready_for_review`. There is currently no build or SAST signal on a 521-file change. `mergeable` is `MERGEABLE`, so no conflicts; `BLOCKED` is only the missing approval.

FigurePaper:
Items 1 and 2 are fixed in `f412556ab2`. The geometry test was checked against the previous arithmetic and fails there, so it guards the invariant rather than restating it.

What is deliberately still open:

**The portal TL suite still runs nowhere.** `test:portal:tl` remains absent from `npm run test`, and `pr-build.yml` has no step for it. Wiring it in is a build-config change whose blast radius is the 28 pre-existing failures below, not this PR's subject — it would turn a green build red on work this branch did not cause. Worth its own change, tracked rather than folded in here.

**The 28 pre-existing failures stay untouched**, all verified against `epic-74519` rather than assumed:

- `binding-editor.component.display.tl.spec.ts` — 26, all `NG0201: No provider found for ChartEditorService`. The base has 27; this branch removes one test with the getter it covered, so it is one better.
- `portal/app.component.risk.tl.spec.ts` — 1. On the base this file does not compile at all, since `setTabOrder` and `dataTabFirst` do not exist on the component. This branch deletes those cases, so again a net improvement.
- `date-value-editor.component.tl.spec.ts` — 1. `"2023-12-01"` parses to `{2023, 11, 30}` at UTC-4. The component is untouched here and this would likely pass in a UTC runner.

Fixing those is the same change as wiring the suite up, and belongs with it.

FigurePaper:
Finding 2 is fixed in `01f065380a`. On the other two:

**1 — the range slider's title-height exclusion is deliberate, not a missed subtype.**

Widget spec §05 excludes CheckBox, RadioButton and TimeSlider from the height row, and the anchored-strip-lane decisions record treats that as settled while deciding the separate question of whether the calendar joins.

The absence you spotted is the mechanism rather than a gap in it. Those three types share `TitleInfo` with the included ones, so the density substitution was deliberately placed in each type's own `getTitleHeight()` read path instead of inside `TitleInfo.getTitleHeight()` — putting it there would have handed it to all eight callers, the three excluded ones included. A range slider returning the raw stored height is what that choice looks like from the excluded side, and the missing dialog checkbox follows from the same place.

The `isUserTitleHeight`/`setUserTitleHeight` bookkeeping it gained is unrelated to the height row: it lives on the shared `TitleInfo` and feeds the parse-time derive for older files, which every titled type needs whether or not it takes a density height. Same for the title chrome in `MaxModeSelectionVSAssemblyInfo` — colour and rule are shared with the selection family; the height is not.

**3 — the duplicated followsDensity blocks are a follow-up.**

I checked all seven copies and they are structurally identical, differing only in the local name for the info, so there is no drift to fix today. The range slider correctly has no copy at all, being outside the height row — which is a second reading of the same boundary as finding 1.

Extracting a shared helper means touching ten call sites across eight services in a branch that is feature-complete and waiting on review, which is the kind of change the project's own bug-fix guidance asks to keep out of an unrelated PR. The generic bound such a helper needs already exists on `VSDensityDefaults.titleHeight`, so it is a small change on its own — just not this one.

jshobe-inetsoft (approved):

PR #5152, opened 2026-09-10
Bug #76576, fix NPE and race in HTML viewsheet export

jkidacki-inetsoft:
## Summary

Exporting a viewsheet to HTML from the Portal viewer (Current View, Match Layout) intermittently threw an NPE and, when a repeat export overlapped with an in-flight one, could hang indefinitely.

## Root Cause

1. **NPE**: `HTMLCrosstabHelper.write()` and `HTMLTableHelper.write()` only guarded against a null `VSTableLens` before calling `initTableGrid()`, but then unconditionally dereferenced `lens` in `initRowColumns()` (`lens.getRowCount()`/`lens.moreRows()`). `ViewsheetSandbox.getVSTableLens()` legitimately returns `null` when an assembly's data isn't ready yet — the PDF/SVG/Excel exporters already handle this gracefully via `VSCrosstabHelper.calculateColumnsPosition()`'s null check; the HTML helpers were the outliers that crashed instead.

2. **Hang**: `ExportControllerService.writeViewsheetExport()` and `VSExportService.writeViewsheetExport()`'s "Current View" export path mutate the shared, live `RuntimeViewsheet`/`ViewsheetSandbox` in place (`rvs.setViewsheet(cviewsheet)`, `exportBox.setViewsheet(cviewsheet, false)`) with no synchronization, and `ViewsheetSandbox.setViewsheet()` itself acquires no lock. The `@ClusterWriteMethod`/`@ClusterProxyMethod` annotations on the export entry point provide no mutual exclusion (verified against their definitions — they only handle async cache write-back and cluster colocation routing). So two overlapping exports of the same runtime viewsheet can race: the second export's clone-and-swap happens mid-flight of the first, corrupting assembly identity out from under the first export (producing the NPE above) and potentially leaving an in-flight query waiting on now-stale assembly state (the reported indefinite hang). The two `finally` restores also stomp on each other.

## Fix

- Added a `lens == null` guard (alongside the existing `info == null` guard) to `HTMLCrosstabHelper.write()` and `HTMLTableHelper.write()`, matching the graceful-degradation behavior of the other format exporters.
- Wrapped the "Current View" clone/swap/export/restore critical section in both `ExportControllerService.writeViewsheetExport()` and `VSExportService.writeViewsheetExport()` in `synchronized(rvs)`, and capture the "original" viewsheet to restore fresh inside the lock (instead of a value captured before waiting on it) — serializing overlapping exports of the same runtime viewsheet. This mirrors `RuntimeViewsheet`'s existing use of `synchronized` instance methods (e.g. `undo()`/`redo()`).

## Test Plan

- `./mvnw test -pl core -Dtest=VSExportServiceTest,AbstractVSExporterLabelTest,VSFontHelperTest` — 48/48 tests pass.
- Two independent code-review passes against the diff found no remaining issues.
- Manually verified against the reported repro (export "vs" to HTML, Current View + Match Layout, then immediately repeat) — no NPE, no hang.

Fixes Redmine #76576.

### DIFF
--- a/core/src/main/java/inetsoft/report/io/viewsheet/html/HTMLCrosstabHelper.java
+++ b/core/src/main/java/inetsoft/report/io/viewsheet/html/HTMLCrosstabHelper.java
@@ -71,13 +71,11 @@ public class HTMLCrosstabHelper extends HTMLTableDataHelper {
    public void write(PrintWriter writer, TableDataVSAssembly assembly, VSTableLens lens) {
       TableDataVSAssemblyInfo info = (TableDataVSAssemblyInfo) assembly.getVSAssemblyInfo();
 
-      if(info == null) {
+      if(info == null || lens == null) {
          return;
       }
 
-      if(lens != null) {
-         lens.initTableGrid(info);
-      }
+      lens.initTableGrid(info);
 
       isWritten = new SparseMatrix();
       Rectangle2D bounds = vHelper.getBounds(info);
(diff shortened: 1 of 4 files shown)


jshobe-inetsoft (approved):

PR #5155, opened 2026-09-10
Bug #76568: fix annotation position when chart is scrolled

larryliang-inetsoft:
## Summary

Adding a data-point annotation on a chart that is horizontally scrolled sent viewport-relative click coordinates to the server instead of full-content coordinates, unlike the table/crosstab/calc-table cell-annotation paths which already add the live scroll offset. The chart data-annotation render path also never subtracted the live scroll offset from the actual rendered CSS position (it was only used for the `isLineInContainer()` visibility check), so a newly created annotation ended up gated invisible by that check for large scroll offsets — e.g. when the horizontal scrollbar is dragged all the way to the end, matching the reported symptom.

Root cause and fix scope confirmed independently across two diagnosis/refutation rounds (see the batch archive: `docs/teams/2026-09-10-bugs-vscalc-chart-batch/bug-76568/` in `stylebi-wiz`), including an explicit check that the two changes below don't double-compensate (`isLineInContainer()`'s bounds check uses a synthetic `Rectangular` model, not the wrapper div's live rendered position — the two computations are structurally decoupled).

## Fix

Two coordinated changes:
- `vs-chart.component.ts` `addAnnotation()`: add `scrollLeft`/`scrollTop` to the computed x/y before sending `AddAnnotationEvent`, so the stored position is a full-content coordinate (matching the table family's existing behavior).
- `vs-object-container.component.html`: subtract the chart's live `annotationScrollLeft`/`annotationScrollTop` from the per-chart data-annotation wrapper div's position, scoped to `getChartDataAnnotations()`/`dataAnnotationModels` only, so chart-assembly annotations and table/crosstab/calc-table annotations (which already work via `positionAnnotationsToCell`) are unaffected.

## Testing

`ng build portal --configuration development` succeeds clean.

Fixes https://173.220.179.100/issues/76568

### DIFF
--- a/web/projects/portal/src/app/vsobjects/objects/chart/vs-chart.component.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/chart/vs-chart.component.ts
@@ -691,8 +691,8 @@ export class VSChart extends AbstractVSObject<VSChartModel>
          .reverse()
          .find((region) => region != null);
       const chartContainerBounds = this.chartContainer.nativeElement.getBoundingClientRect();
-      let x = (event.clientX - chartContainerBounds.left) / this.scale;
-      let y = (event.clientY - chartContainerBounds.top) / this.scale;
+      let x = (event.clientX - chartContainerBounds.left) / this.scale + this.scrollLeft;
+      let y = (event.clientY - chartContainerBounds.top) / this.scale + this.scrollTop;
 
       // triggered from mini menu
       if(event.button == 0) {
--- a/web/projects/portal/src/app/vsobjects/objects/vs-object-container.component.html
+++ b/web/projects/portal/src/app/vsobjects/objects/vs-object-container.component.html
@@ -386,8 +386,8 @@
   @for (vsObject of vsInfo.vsObjects; track vsObject.absoluteName; let i = $index) {
     @if (vsObject.objectType === 'VSChart' && viewer && !context.binding &&
       !$any(vsObject).maxMode && !!$any(vsObject).plot) {
-      <div [style.top.px]="getVsObjectPosition(vsObject, i, true)"
-        [style.left.px]="getVsObjectPosition(vsObject, i, false)"
+      <div [style.top.px]="getVsObjectPosition(vsObject, i, true) - ($any(vsObject).annotationScrollTop || 0)"
+        [style.left.px]="getVsObjectPosition(vsObject, i, false) - ($any(vsObject).annotationScrollLeft || 0)"
         style="position: absolute;">
         @for (ann of getChartDataAnnotations(vsObject); track ann.absoluteName) {
           <vs-annotation


fredaan (approved):
## Code Review: Bug #76568 — chart annotation position when scrolled

### Summary
Minimal, well-scoped two-file fix. I traced both changes against the surrounding scroll/scale plumbing and confirm the root-cause analysis in the PR description.

### Correctness verification

**`vs-chart.component.ts` `addAnnotation()`**
- `this.scrollLeft`/`this.scrollTop` are populated in `onScroll()` from the chart plot area's raw `event.target.scrollLeft/scrollTop` (`chart-area.component.ts`), which is a native DOM scroll property — unaffected by the ancestor `.scale-container`'s CSS `transform: scale()`. That means it's already in the same "logical/content" pixel space that `(event.clientX - chartContainerBounds.left) / this.scale` produces. Adding it *after* the `/this.scale` division (rather than before, as `base-table.ts`'s `positionAnnotationsToCell` does) is dimensionally correct here and mirrors the intent of the table family's fix without literally copy-pasting an inapplicable ordering.
- Confirmed this now matches the full-content coordinate convention already used by the table/crosstab/calc-table cell-annotation path.

**`vs-object-container.component.html` wrapper div**
- `getVsObjectPosition()` returns the *static* `objectFormat.top/left` of the chart assembly (a model value, not DOM-derived), so subtracting `annotationScrollTop/Left` (also logical/unscaled, same `Point` set in `onScroll`) is unit-consistent.
- Verified the change is properly scoped to the `getChartDataAnnotations()` overlay block only — the other `getVsObjectPosition()` call site (regular VS object stacking, ~line 971) is untouched, so no double-application to assembly annotations or table/crosstab/calc-table annotations (which position via the separate, pre-existing `positionAnnotationsToCell`).
- Confirmed no double-compensation with `isLineInContainer()`: that method's containment check already receives the scroll offset independently via the pre-existing `[offsetX]="-(annotationScrollLeft||0)"` / `[offsetY]="-(annotationScrollTop||0)"` bindings on `<vs-annotation>`, and operates on a synthetic `Rectangular` model rather than the wrapper div's rendered position — the two computations are structurally decoupled, as the PR description states.

### Other review areas
- **Error handling / edge cases**: `|| 0` guards on `annotationScrollTop/Left` correctly handle the pre-scroll (`undefined`) state; `getVsObjectPosition` never returns `null` for `VSChart`, so the subtraction is always a valid numeric op.
- **Backward compatibility**: No API/model/template contract changes; purely internal coordinate math. No migration or data-format concerns.
- **Regressions**: Traced all call sites of both changed functions; the fix is properly scoped and doesn't affect non-chart assemblies or the assembly-annotation path.
- **Security**: No user input handling changes, no injection surface.
- **Tests**: No automated test was added for this fix (manual verification was `ng build portal --configuration development`). Given this is a coordinate/geometry regression that's easy to silently reintroduce, a small unit test on `addAnnotation()` (asserting `scrollLeft`/`scrollTop` are included in the sent event) and/or on the wrapper `top/left` calculation would be valuable, but I'm not treating this as blocking given the very small, low-risk diff and existing test suites for `vs-chart.component` / `vs-object-container.component` that already cover adjacent logic.

### CI status
- `claude-review` (automated bot) failed with an internal execution error (`is_error: true`), not a substantive finding — this manual review supersedes it.
- `Semgrep Scan`, `Semgrep OSS`, `build`, and `license/cla` all pass. No merge conflicts (`mergeable: MERGEABLE`).

### Verdict

PR #5157, opened 2026-09-10
Bug #76566: harden CalcFieldFormula/ScriptEnvRepository/AssetQuerySandbox against null ScriptEnv

larryliang-inetsoft:
## Summary

An unhandled NPE (`this.senv` is null) in `CalcFieldFormula.getResult0()` via `SummaryFilter`, reported to reproduce only on the first viewsheet open after a server restart.

**Scope note, after two diagnosis/refutation rounds** (see `docs/teams/2026-09-10-bugs-vscalc-chart-batch/bug-76566/` in `stylebi-wiz`): the exact restart-timing mechanism (why it happens on the first post-restart open specifically) was not conclusively established from static reading alone — two candidate `ScriptEnvRepository` classloading theories were ruled out, and a genuine, independently-found synchronization defect in `AssetQuerySandbox.senv` was identified as a better-targeted candidate, but its reachability during a real restart wasn't confirmed without live instrumentation. This PR ships the confirmed-safe hardening below; it does **not** claim to resolve the restart-timing root cause, which remains open (recommended follow-up: temporary logging + a live restart repro, noted in the investigation docs).

## Fix

Three changes:
1. `CalcFieldFormula.getResult0()`: no longer double-NPEs when `senv` is null — throws one clear, diagnosable `ScriptException` instead.
2. `ScriptEnvRepository.getScriptEnv()`: logs the previously-swallowed `Throwable` in both catch blocks (thread name and the `found` flag), so a future occurrence is diagnosable in production instead of silent.
3. `AssetQuerySandbox.senv`: made `volatile`; `reset()`/`dispose()` now null it under the same lock monitor `getScriptEnv()` uses, instead of no synchronization / a different monitor — closing a real, independently-confirmed data race between the field's three writers.

## Testing

`core` module compiles clean.

Fixes https://173.220.179.100/issues/76566

### DIFF
--- a/core/src/main/java/inetsoft/report/composition/execution/AssetQuerySandbox.java
+++ b/core/src/main/java/inetsoft/report/composition/execution/AssetQuerySandbox.java
@@ -399,7 +399,10 @@ public class AssetQuerySandbox implements Serializable, Cloneable, ActionListene
       resetTableLens();
       resetDefaultColumnSelection();
 
-      senv = null;
+      synchronized(lock) {
+         senv = null;
+      }
+
       scope = null;
    }
 
@@ -1546,8 +1549,11 @@ public class AssetQuerySandbox implements Serializable, Cloneable, ActionListene
          }
 
          scope = null;
-         senv = null;
          vprovider2 = null;
+      }
+
+      synchronized(lock) {
+         senv = null;
       }
    }
 
@@ -1919,7 +1925,7 @@ public class AssetQuerySandbox implements Serializable, Cloneable, ActionListene
    private Hashtable<String, SelectionVSAssembly> selections; // selection assembly map,
    private ViewsheetSandbox vsbox;
    private AssetQueryScope scope; // scope for executing formulas
-   private ScriptEnv senv; // scripting env for executing scripts
+   private volatile ScriptEnv senv; // scripting env for executing scripts
    private final MVSession mvsession;
    private final Set<String> nolimit; // tables to ignore time limit
    private QueryManager queryMgr; // track pending queries
--- a/core/src/main/java/inetsoft/report/filter/CalcFieldFormula.java
+++ b/core/src/main/java/inetsoft/report/filter/CalcFieldFormula.java
@@ -275,6 +275,13 @@ public class CalcFieldFormula implements PercentageFormula, Formula2 {
          result = senv.exec(script, scope = updateParameter(), null, null);
       }
       catch(Exception ex) {
+         if(senv == null) {
+            String msg = "Script failed, ScriptEnv is not available:\n" +
+               XUtil.numbering(runtimeFormula);
+            LOG.warn(msg, ex);
+            throw new ScriptException(msg, ex);
+         }
+
          String suggestion = senv.getSuggestion(ex, "field", scope);
          String msg = "Script error: " + ex.getMessage() +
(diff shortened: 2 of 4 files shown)


fredaan (approved):
## Code Review Summary

Reviewed the three hardening changes plus a regression test I added on top.

### Correctness

**1. `AssetQuerySandbox.senv` → `volatile` + shared-lock nulling** (`AssetQuerySandbox.java`)
- The existing `getScriptEnv()` double-checked-locking read (`ScriptEnv senv = this.senv; if (senv == null) { synchronized(lock) { ... } }`) was previously unsafe because `senv` wasn't `volatile` and two of its three writers (`reset()`, `dispose()`) didn't synchronize on `lock` at all. Making the field `volatile` and moving all three writers (`reset()`, `dispose()`, `getScriptEnv()`) under the same `lock` monitor closes that race correctly — this is now a textbook-correct DCL pattern.
- Verified `dispose()`'s reordering (moving `senv = null` out of the pre-existing `synchronized(this)` block into its own `synchronized(lock)` block) doesn't break any other invariant — nothing else in the class relies on `senv` and `scope`/`vprovider2` being nulled atomically under the same monitor.

**2. `CalcFieldFormula.getResult0()` null-senv guard** (`CalcFieldFormula.java`)
- Converts the double-NPE (crash inside the existing catch block calling `senv.getSuggestion(...)`) into one clear, logged `ScriptException`. Confirmed by test (see below) that this fires correctly and includes the original NPE as the cause for diagnosability.

**3. `ScriptEnvRepository` logging** (`ScriptEnvRepository.java`)
- Straightforward; previously-swallowed `Throwable`s are now logged at WARN with thread name, so a future null-`ScriptEnv` occurrence is diagnosable in production. No behavior change, low risk.

### Regression test added

The PR shipped without a test for the new `getResult0()` branch. I added `getResult_nullScriptEnv_throwsScriptExceptionNotNPE` to the existing `CalcFieldFormulaTest`, using the existing `AssetQuerySandbox`/`box.getScope()` test scaffolding already in the file. I verified:
- It **fails** against the pre-fix code (`AssertionFailedError: expected ScriptException but was NullPointerException`), confirming it actually exercises the double-NPE this PR fixes.
- It **passes** against the PR's fix.
- Full `CalcFieldFormulaTest` suite (33 tests) and the full CI `build` job (including this test) both pass on the latest commit.

### Non-blocking observation (follow-up, not part of this PR's scope)

`AssetQuerySandbox.scope` (the `AssetQueryScope` field, distinct from `CalcFieldFormula`'s own `ScriptScope`) has the same structural issue that was just fixed for `senv`: it's read/written via double-checked locking in `getScope()` under `synchronized(lock)`, but is nulled unsynchronized in `reset()` and under a *different* monitor (`synchronized(this)`) in `dispose()`, and the field itself isn't `volatile`. This is pre-existing, not introduced by this PR, and the PR's own description is explicit that it only ships the "confirmed-safe" `senv` fix rather than a broader sweep — so I'm flagging it as a suggested follow-up rather than blocking this PR on it.

### CI / mergeability

- `build`, `Semgrep Scan`, `Semgrep OSS`, and `license/cla` all pass on the latest commit.
- `claude-review` (automated) shows a failure, but it errored out before producing an actual review (see its own comment on the PR) — this manual review supersedes it per project convention.

**Recommendation: Approve.**

NeverStopFMH (approved):

PR #5162, opened 2026-09-11
fix(wiz): add_table queryParams exclusion check missed parameters/lookup/customLookups

fredaan:
## Summary
- Bug 76581's PR-review-bot addendum (and my own follow-up code analysis) found that `add_table`'s `queryParams` mutual-exclusion check in `WorksheetAgentController.editOp` only rejected `queryParams` combined with `endpoint`/`suffix`, never with `parameters`/`lookup`/`customLookups` -- even though the composer plugin's own tool description documents all five as mutually exclusive.
- `addQueryParamsTable` never reads `parameters`/`lookup`/`customLookups` at all, so a request combining `queryParams` with any of them previously passed validation silently and dropped the extra fields with no error -- the "tool accepts malformed input and produces a plausible-but-wrong result" pattern this repo's CLAUDE.md calls a plugin defect.
- Extends the existing check to also reject `queryParams` + `parameters`/`lookup`/`customLookups`, each with its own named error.

## Test plan
- [x] New tests added (`WorksheetAgentControllerTest`) confirmed red against the unfixed code (fell through to a different error instead of the expected rejection), then green after the fix.
- [x] `./mvnw -q -pl core -am test -Dtest=WorksheetAgentControllerTest -Dsurefire.failIfNoSpecifiedTests=false` -- 168 tests, all passing, including the pre-existing endpoint/suffix rejection test.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -302,6 +302,21 @@ public class WorksheetAgentController {
                "endpoint or suffix -- choose exactly one add_table form.");
          }
 
+         if(req.parameters() != null && !req.parameters().isEmpty()) {
+            throw new PairingException("add_table cannot carry queryParams together with " +
+               "parameters -- parameters only applies to the endpoint form.");
+         }
+
+         if(req.lookup() != null && !req.lookup().isEmpty()) {
+            throw new PairingException("add_table cannot carry queryParams together with " +
+               "lookup -- lookup only applies to the endpoint form.");
+         }
+
+         if(req.customLookups() != null && !req.customLookups().isEmpty()) {
+            throw new PairingException("add_table cannot carry queryParams together with " +
+               "customLookups -- customLookups only applies to the suffix form.");
+         }
+
          if(req.datasource() == null || req.datasource().isBlank()) {
             throw new PairingException("datasource is required when queryParams is specified.");
          }
--- a/core/src/test/java/inetsoft/web/wiz/worksheet/WorksheetAgentControllerTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/worksheet/WorksheetAgentControllerTest.java
@@ -418,6 +418,30 @@ class WorksheetAgentControllerTest {
    }
 
    /**
+    * Same as {@link #addQueryParamsTableRequest}, but also carrying {@code parameters}/
+    * {@code lookup}/{@code customLookups} -- the three endpoint/suffix-only fields
+    * {@code addQueryParamsTable} silently ignores unless {@code editOp} rejects the
+    * combination up front.
+    */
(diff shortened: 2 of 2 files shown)


fredaan:
## Round 1 review — head sha a0d6556945501eb926948a72c403448119805e95

**Checked:**
- Read the actual diff (`WorksheetAgentController.java` editOp's queryParams block + new tests) and the current `EditRequest.java` at this commit to confirm field types.
- Empty vs. null handling: `parameters` is `Map<String,String>`, `lookup`/`customLookups` are `List<...>` — all three new checks use `!= null && !X.isEmpty()`, so an empty map/list (the tool's own "not specified" convention) correctly does NOT trigger rejection. Verified against the actual field declarations, not just the diff.
- Root cause claim: confirmed `addQueryParamsTable` (lines 828–904 at this commit) never reads `req.parameters()`, `req.lookup()`, or `req.customLookups()` anywhere in its body — the silent-drop this fix closes is real.
- Check ordering: the three new checks sit inside the same `queryParams != null && !isEmpty()` block, after the existing endpoint/suffix mutual-exclusion check and before the datasource-required/logicalModel/schema-catalog checks. Tried to build a counter-example where order matters (e.g. `queryParams` + `endpoint` + `parameters` together) — the endpoint/suffix check fires first and reports the more relevant "choose exactly one add_table form" error rather than the parameters-specific one; no bad interaction found.
- Considered whether `lookupExpandArrays`/`lookupTopLevelOnly` (only meaningful together with `lookup`, per `EditRequest`'s own doc) needed the same treatment — they have no effect on any add_table form unless `lookup` is also set, and `lookup` is already covered by the new check, so there's no silent-drop path left open for them.
- Other call sites: checked wiz-services' own `/ws/table` path (`WorksheetTableService.buildTabularTable` / `WorksheetTable.TabularSource`) for the same missing-validation pattern. That model only has `datasourcePath`/`queryParams`/`maxRows` — it has no `endpoint`/`parameters`/`lookup`/`customLookups`/`suffix` fields at all, so the mutual-exclusion gap this PR fixes doesn't exist there. No "same bug, other site" finding.
- Error message wording matches this file's existing convention ("add_table cannot carry X together with Y -- ...").
- Test coverage: the 3 new tests each build a request with `queryParams` + exactly one conflicting field, assert the specific `PairingException` message, and assert `verifyNoInteractions` on `dataSourceService`/`editSvc` (proving the rejection happens before any side effect) — not vacuous. Confirmed the pre-existing `addQueryParamsTableRejectsQueryParamsTogetherWithEndpoint`/`...WithSuffix` tests are untouched by this diff, so the pre-existing endpoint/suffix regression coverage still stands.
- CLAUDE.md: this is exactly the "tool accepted malformed input and produced a plausible-but-wrong result" pattern the Plugin Testing Principle calls a plugin defect, fixed at the plugin/backend layer with a named error per field plus regression tests — satisfies both the misuse-handling and bug-fix-regression-test requirements.

**No findings.**

PR #5166, opened 2026-09-11
fix(admin/schedule): pin reviewed task narrative into audit record via taskToken (#76588)

NeverStopFMH:
## Summary
- Part of Redmine bug #76588 (taskToken audit-pinning sweep), Schedule area.
- `ScheduleChangePlanService.resolve()` already issues a `taskToken` (via `TaskAuditToken.issue`), but `ScheduleApplyRequest` had no field for it and `ScheduleChangesetApplyService.apply()` never verified one — every audit write (create, delete, and the rollback branch, three call sites) used the apply request's own re-resolved `task` string instead of the narrative actually reviewed at preview.
- Adds `taskToken` to `ScheduleApplyRequest`, verifies it via `TaskAuditToken.verify()` right after the existing `planHash` gate, and threads the verified `reviewedTask` through all three call sites (`applyCreate`, `applyDelete`, `rollback`) instead of `plan.task()`. Reuses the shared, already-tested `AdminChangesetApplyService.TaskTokenMismatchException` (same as `PlanHashMismatchException` is already reused) and adds a matching `@ExceptionHandler` in `AdminScheduleController`.
- Companion plugin PR: inetsoft-technology/stylebi-wiz#2417

## Test plan
- [x] `./mvnw -pl core test -Dtest=ScheduleChangesetApplyServiceTest,AdminScheduleControllerTest -DfailIfNoTests=false` — 24 tests run, 0 failures

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/schedule/AdminScheduleController.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/schedule/AdminScheduleController.java
@@ -139,6 +139,16 @@ public class AdminScheduleController {
                     "plan", ex.current());
    }
 
+   @ExceptionHandler(AdminChangesetApplyService.TaskTokenMismatchException.class)
+   @ResponseStatus(HttpStatus.CONFLICT)
+   @ResponseBody
+   public Map<String, Object> handleTaskTokenMismatch(
+      AdminChangesetApplyService.TaskTokenMismatchException ex)
+   {
+      return Map.of("status", "conflict", "error", String.valueOf(ex.getMessage()),
+                    "plan", ex.current());
+   }
+
    private final AdminScheduleGateway scheduleGateway;
    private final ScheduleChangePlanService planService;
    private final ScheduleChangesetApplyService applyService;
--- a/core/src/main/java/inetsoft/web/admin/ai/schedule/ScheduleApplyRequest.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/schedule/ScheduleApplyRequest.java
@@ -25,8 +25,10 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 public class ScheduleApplyRequest extends ScheduleChangePlanRequest {
    public String getPlanHash() { return planHash; }
    public void setPlanHash(String v) { this.planHash = v; }
+   public String getTaskToken() { return taskToken; }
+   public void setTaskToken(String v) { this.taskToken = v; }
    public String getReviewOutcome() { return reviewOutcome; }
    public void setReviewOutcome(String v) { this.reviewOutcome = v; }
 
-   private String planHash, reviewOutcome;
+   private String planHash, taskToken, reviewOutcome;
 }
--- a/core/src/main/java/inetsoft/web/admin/ai/schedule/ScheduleChangesetApplyService.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/schedule/ScheduleChangesetApplyService.java
@@ -70,11 +70,15 @@ public class ScheduleChangesetApplyService {
    }
 
    /**
-    * Resolves, gates on the plan hash, then executes.
+    * Resolves, gates on the plan hash and task token, then executes.
     *
     * @throws AdminChangesetApplyService.PlanHashMismatchException if the hash is missing or stale
     *         (maps to HTTP 409) -- the exception type is reused verbatim from the properties area
     *         (spec §6), not redeclared.
+    * @throws AdminChangesetApplyService.TaskTokenMismatchException if the taskToken is missing,
+    *         malformed, or was issued for a different planHash (also maps to HTTP 409, also reused
+    *         verbatim) -- the audit record must be written from the token's verified narrative,
+    *         never from this request's own (unverified) task field.
     * @throws Exception if the Tier-2 backup fails, in which case nothing was applied.
     */
    public ApplyResult apply(ScheduleApplyRequest req, Principal user) throws Exception {
@@ -85,6 +89,15 @@ public class ScheduleChangesetApplyService {
 
          if(req.getPlanHash() == null || !plan.planHash().equals(req.getPlanHash())) {
             throw new AdminChangesetApplyService.PlanHashMismatchException(plan);
+         }
+
+         String reviewedTask;
+
+         try {
+            reviewedTask = TaskAuditToken.verify(req.getTaskToken(), plan.planHash());
+         }
+         catch(TaskAuditToken.TaskTokenException e) {
+            throw new AdminChangesetApplyService.TaskTokenMismatchException(plan, e.getMessage());
          }
 
          if(plan.requiresAgentSignoff() &&
@@ -110,11 +123,11 @@ public class ScheduleChangesetApplyService {
 
             try {
                if(ScheduleChangeRequest.VERB_CREATE.equals(original.getVerb())) {
-                  applyCreate(txId, plan.task(), taskId, original.getSpec(), backupRef,
+                  applyCreate(txId, reviewedTask, taskId, original.getSpec(), backupRef,
                              req.getReviewOutcome(), user, results, undoable);
                }
                else {
-                  applyDelete(txId, plan.task(), taskId, backupRef, req.getReviewOutcome(), user,
+                  applyDelete(txId, reviewedTask, taskId, backupRef, req.getReviewOutcome(), user,
                              results, undoable);
                }
             }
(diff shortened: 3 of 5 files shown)


NeverStopFMH:
## Review (round 1) — cause-vs-symptom check for bug #76588, Schedule area

Scope checked: the diff against the confirmed root cause in `docs/teams/2026-09-11-bug-76588-tasktoken-audit-pinning/bug-schedule/01-diagnosis.md` / `02-refute.md` (independent of this repo), not just "does it compile / do the new tests pass".

### What checks out (verified directly against the diff, not re-trusting the PR description)

- `ScheduleApplyRequest` gains `taskToken` field/getter/setter, matching `ApplyRequest`'s shape exactly.
- `ScheduleChangesetApplyService.apply()` calls `TaskAuditToken.verify(req.getTaskToken(), plan.planHash())` right after the existing `planHash` gate, throwing the **shared, already-tested** `AdminChangesetApplyService.TaskTokenMismatchException` on failure (not a new, Schedule-local exception type) — correct reuse, mirrors `PlanHashMismatchException`'s existing reuse pattern.
- All **three** call sites that previously passed `plan.task()` now pass the verified `reviewedTask`: `applyCreate` (line ~127), `applyDelete` (line ~131), and the `rollback(...)` call (line ~170) in the failure branch. This is the area-specific trap this bug's workflow explicitly calls out (Schedule has 3 audit call sites vs. Properties' 1), and the production code change handles all three correctly — verified by direct diff read, not by trusting the PR description.
- `AdminScheduleController` gets a matching `@ExceptionHandler(AdminChangesetApplyService.TaskTokenMismatchException.class)` mapped to `HttpStatus.CONFLICT` (409), same shape as the existing `handlePlanHashMismatch` and as `AdminAiController`'s own handler.
- The `SreeEnv`/`ShutdownException` test-mocking fix described in `03-fix.md` is real: `ScheduleChangesetApplyServiceTest` now `mockStatic(SreeEnv.class, ...LENIENT)` in `setUp()`/closes it in `tearDown()`, and separately adds a real `MockedStatic<Audit>` intercepting `Audit.getInstance().auditAdminChange(...)` — this is necessary because Schedule's `writeAudit` calls the static `Audit` singleton directly (no injectable `AdminChangeService` the way Properties has). Without this, the new audit-narrative assertions would silently never reach the mock (exactly the failure mode `03-fix.md` describes finding on first run).
- New tests `auditsThePreviewedTaskForACreateEvenWhenApplyTaskDiffers` / `...ForADeleteEvenWhenApplyTaskDiffers` genuinely assert the *written* `AdminChangeRecord.taskDescription` equals the token's reviewed narrative, not the apply call's own (deliberately divergent, in the test) `task` field — via `ArgumentCaptor<AdminChangeRecord>` on the now-real `auditMock`. This is a real regression proof, not a shallow "apply succeeded" check.
- Missing/blank/wrong-plan taskToken rejection tests (`rejectsAMissingTaskToken`, `rejectsABlankTaskToken`, `rejectsATaskTokenIssuedForADifferentPlan`) all assert the mutating gateway call (`addScheduleTask`) is never reached — confirms fail-fast, not "reject after already mutating."
- No unrelated scope creep: no changes to `ResolvedPlan.java`, `TaskAuditToken.java`, `ScheduleChangePlanService.java`, or the shared `AdminChangesetApplyService.java` — consistent with the diagnosis's conclusion that only the apply-side DTO/service/controller needed changes.

### Blocking finding

**Missing regression coverage for the rollback call site specifically (one of the three call sites this bug is about).**

The production fix correctly threads `reviewedTask` into the `rollback(...)` call (in place of `plan.task()`), but none of the new or existing tests actually prove it. The existing rollback tests (`secondChangeFailingRollsBackTheFirst`, `rollbackItselfFailingIsReportedNamingTheTaskId`, `throwMidApplyIsReportedAsUnknownStateAndRollbackFailed`) build their `ScheduleApplyRequest` via the plain `applyRequest(task, ...)` helper, which issues a token for the *same* task text the apply call also carries — so even if `rollback(...)` were still (hypothetically) called with the raw, unverified `plan.task()` instead of `reviewedTask`, none of these tests could tell the difference; the two values happen to be identical in every rollback test.

`01-diagnosis.md`'s own regression-test guidance (point 5) explicitly flagged this: "A rollback path … should also be extended/duplicated to confirm the rollback's own audit writes … use the verified `reviewedTask`, not `plan.task()` — this is the one Schedule-specific place a copy-paste of Properties' test wouldn't automatically cover, since Properties has no separate rollback call site passing task independently." That guidance was not followed — the two new `auditsThePreviewedTask...` tests exercise only the `applyCreate`/`applyDelete` call sites (both non-rollback, successful-apply paths), never the `rollback(...)` call site.

Concretely: extend `secondChangeFailingRollsBackTheFirst` (or add a new test) using the existing `requestWithDivergentApplyTask(previewTask, applyTask, ...)` helper (already added in this PR for the create/delete tests) to force a mid-apply failure that triggers `rollback(...)`, then assert via the `auditMock` captor that the rollback's own `writeAudit` calls (lines ~252/272 per `01-diagnosis.md`) recorded the *previewed*, token-verified narrative — not the apply-time text. Without this, the "all three call sites" claim in the PR description is verified by reading the diff, but not by the test suite itself — which is exactly the kind of narrow, symptom-covering-only-the-happy-path gap this bug's own workflow was designed to catch.

This is the only blocking finding. Please add the rollback-path assertion (or point to where it already exists if I missed it) before merge.

### Non-blocking / FYI

- CI: `claude-review` is failing on this PR, but the job log shows `is_error: true` after a 393 ms run with `ANTHROPIC_API_KEY` empty in the job's env dump — this looks like a pre-existing CI secret/infra issue, not a finding about this diff's content. `build` (Validation Build) is still queued/in-progress as of this review — please re-check once it completes.
- Per this batch's rule, I am not merging or approving this PR myself even if it were otherwise clean — that decision is held for a human.

Companion plugin PR: inetsoft-technology/stylebi-wiz#2417 (reviewed separately; consistent with this PR as a matched pair — no blocking findings there).

NeverStopFMH:
Closed the rollback-coverage gap flagged above. Added
`auditsThePreviewedTaskForARollbackEvenWhenApplyTaskDiffers` in
`ScheduleChangesetApplyServiceTest.java` (mirrors
`secondChangeFailingRollsBackTheFirst`'s two-change create+failing-delete
scenario, but built via `requestWithDivergentApplyTask` so the create's
apply-time task text diverges from its previewed/reviewed narrative). It
asserts the `AdminChangeRecord`s captured with `action == ACTION_ROLLBACK`
carry the previewed narrative, not the apply request's own task field.

Verified the test actually detects the regression it's meant to catch:
temporarily reverted `rollback(...)`'s `task` argument at the call site
(line ~170) back to `req.getTask()` and confirmed the new test fails
(`expected: <true> but was: <false>`), then restored the production code
(no functional change to production, only the new test).

Full scoped run, production code unchanged:
```
./mvnw -pl core test -Dtest=ScheduleChangesetApplyServiceTest,AdminScheduleControllerTest -DfailIfNoTests=false
Tests run: 11, Failures: 0, Errors: 0, Skipped: 0 -- AdminScheduleControllerTest
Tests run: 14, Failures: 0, Errors: 0, Skipped: 0 -- ScheduleChangesetApplyServiceTest
Tests run: 25, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

Pushed as a new commit on this branch: 6ce5cbf5c.

NeverStopFMH:
Round 2 review: confirmed closed.

- Read commit `6ce5cbf5c` directly — it touches only `ScheduleChangesetApplyServiceTest.java` (no production change), adding `auditsThePreviewedTaskForARollbackEvenWhenApplyTaskDiffers`. It reuses the two-change create+failing-delete scenario (same as `secondChangeFailingRollsBackTheFirst`) via the `requestWithDivergentApplyTask` helper, filters captured `AdminChangeRecord`s to `action == ACTION_ROLLBACK` (rollback's own writeAudit calls, distinct call site from `applyCreate`/`applyDelete`'s `ACTION_APPLY` writes), and asserts they carry the previewed narrative (`"reviewed: two changes"`) rather than the diverged apply-time text.
- Confirmed the production rollback call site is unchanged from round 1: `rollback(txId, reviewedTask, ...)` at `ScheduleChangesetApplyService.java:170`, still fed by `reviewedTask = TaskAuditToken.verify(...)`.
- Ran the scoped suite myself from the PR branch: `./mvnw -pl core test -Dtest=ScheduleChangesetApplyServiceTest,AdminScheduleControllerTest -DfailIfNoTests=false` → 25 tests, 0 failures (14 in `ScheduleChangesetApplyServiceTest`, up from 13 in round 1). Matches the numbers reported in the fix notes.

The round-1 blocking finding (no coverage for the rollback audit call site) is resolved. No new issues found. `claude-review` still fails with the same pre-existing `ANTHROPIC_API_KEY`-empty infra signature noted in round 1 (not a content issue); `Semgrep`/`license-cla` pass; `build` (Validation Build) was still in progress at review time. Approving pending human merge, per this batch's rule that merge decisions are held for a human regardless of review outcome.

PR #5169, opened 2026-09-11
fix(wiz): stop a non-options write from persisting the manufactured percentageBy default (VTB-007, #76574)

lizhepu:
## Bug
Redmine #76574 / VTB-007. Diagnosis + refutation history (in `inetsoft-technology/stylebi-wiz`):
`docs/teams/2026-09-10-bugs-vtb-p1/bug-76574-vtb007/01-diagnosis.md` and `02-refute.md`.

## Root cause
`CrosstabOptionInfo(CrosstabVSAssembly)`'s constructor
(`core/src/main/java/inetsoft/web/binding/model/table/CrosstabOptionInfo.java:34-45`) coalesces a
never-set `percentageByValue` (`null`) to `"1"` (`PERCENTAGE_BY_COL`) purely for display when
building a binding model snapshot. But
`VSCrosstabBindingFactory.updateAssembly()` (`:281-282`) persists whatever the model's option
carries onto the live assembly unconditionally, on every wiz write — not just an options write.
So the very first wiz touch to a never-configured crosstab (a shelf edit, a sort, anything)
silently and permanently set `percentageBy` to `"col"` as a side effect, even if the caller never
called `set_table_options`.

Live-confirmed by the lead: a brand-new crosstab with only `rows`/`cols` bound (no aggregate, no
options call) already reported `get_table_binding`'s `options.percentageBy: "col"`.

## Fix
Scoped to the wiz-only layer (`inetsoft.web.wiz.binding`), not the shared
`VSCrosstabBindingFactory`/`CrosstabOptionInfo` the native Composer's own ~15 controllers also use
— avoids a native-UI regression risk.

- `TableBindingMutator.preserveUntouchedPercentageBy(model, livePercentageByValue)` (new): resets
  a `CrosstabBindingModel`'s option back to `null` when the live crosstab's own `percentageBy`
  was still `null` at the start of the call.
- `TableBindingService`: wired into every write path except `setOptions` (the one call allowed
  to actually change `percentageBy`) — `setShelf`'s inline mutate, `applyWithContext`
  (`addField`/`removeField`/`moveField`), and a new `preservePercentageBy` flag on the shared
  `apply()` overload (`setSort`/`setRanking`/`setColumnLabels` default it `true`; `setOptions`
  passes `false`).

`TableBindingMutator.percentageByName()`'s existing null-guard is unchanged — it was already
correct and is now meaningful again, since a real, persisted crosstab can actually be `null`.

## Tests
Added to `TableBindingServiceTest.java`:
1. `setShelfDoesNotPersistTheManufacturedPercentageByDefault` — a non-options write no longer
   persists a manufactured `"1"`.
2. `setOptionsStillAppliesAnExplicitPercentageByChange` — an explicit
   `set_table_options(percentageBy: "row")` still applies normally.

`./mvnw -pl core test -Dtest=TableBindingServiceTest` → 30/30 pass. Also re-ran
`TableBindingMutatorTest,FieldRefFactoryTest,FieldRefBindingTest,BindingAgentControllerTest,BindingReadServiceTest`
(sibling wiz-binding suites) → all pass, no regressions.

**Note**: a third regression case from the diagnosis (calculateInfo `byRow`/`byColumn`/`prefix`
tracking `percentageBy` in lock-step across two `set_table_options` calls) was not added —
it exercises `syncPercentCalcs()`/`VSCrosstabBindingFactory.updateAssembly()`, which
`TableBindingServiceTest`'s mock-based harness doesn't wire up for real (`VSBindingModelService`/
`VSBindingService` are both plain mocks). Per the diagnosis, this mechanism already works
correctly and is a nice-to-have regression guard, not part of this bug's fix.

Do not merge — awaiting review.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/TableBindingMutator.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/TableBindingMutator.java
@@ -843,6 +843,33 @@ public final class TableBindingMutator {
    // ── options (2d Phase 3) ──────────────────────────────────────────────────
 
    /**
+    * Undoes {@code CrosstabOptionInfo(CrosstabVSAssembly)}'s null-to-{@code "1"} (col) display
+    * default on a wiz write that never asked to change {@code percentageBy}.
+    *
+    * <p>That constructor manufactures {@code "1"} purely so a never-configured crosstab has
+    * something to show when a binding model snapshot is built for it — but {@code
+    * VSCrosstabBindingFactory.updateAssembly()} persists whatever the model's option carries
+    * onto the live assembly unconditionally, on every write, not just an options write. Without
+    * this, the very first wiz touch to a never-configured crosstab — a shelf edit, a sort,
+    * anything, not just {@code set_table_options} — silently and permanently set {@code
+    * percentageBy} to {@code "col"} as a side effect.
+    *
+    * <p>Every caller in {@code TableBindingService} except {@code setOptions} passes the live
+    * assembly's own {@code VSCrosstabInfo.getPercentageByValue()}, read before the mutation, as
+    * {@code livePercentageByValue} — {@code setOptions} is the one call allowed to actually
+    * change it, so it never calls this at all.
+    */
+   public static void preserveUntouchedPercentageBy(BaseTableBindingModel model,
+                                                    String livePercentageByValue)
+   {
+      if(livePercentageByValue == null && model instanceof CrosstabBindingModel crosstab &&
+         crosstab.getOption() != null)
+      {
+         crosstab.getOption().setPercentageByValue(null);
+      }
+   }
+
+   /**
     * Crosstab and table options.
     *
     * <p><b>The crosstab totals are string-typed booleans</b> — {@code rowTotalVisibleValue} and
--- a/core/src/main/java/inetsoft/web/wiz/binding/TableBindingService.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/TableBindingService.java
@@ -84,7 +84,9 @@ public class TableBindingService {
             ? data.getSourceInfo() : null;
          requireSourceForFieldWrite(assemblyName, shelf, model.getSource(),
                                     fields == null ? 0 : fields.size());
+         String livePercentageBy = livePercentageByValue(assembly);
          TableBindingMutator.setShelf(model, shelf, fields, rvs, source, refModelService);
+         TableBindingMutator.preserveUntouchedPercentageBy(model, livePercentageBy);
 
          ApplyVSAssemblyInfoEvent event = new ApplyVSAssemblyInfoEvent();
          event.setName(assemblyName);
@@ -402,8 +404,10 @@ public class TableBindingService {
    public void setOptions(String sessionToken, Principal user, String assemblyName,
                           Map<String, Object> options) throws Exception
    {
+      // Unlike every other mutator below, this one is allowed to actually change percentageBy
+      // — so it must not have preserveUntouchedPercentageBy undo whatever it just set.
       apply(sessionToken, user, assemblyName,
-            model -> TableBindingMutator.setOptions(model, options));
+            model -> TableBindingMutator.setOptions(model, options), false, false);
    }
 
    public Map<String, Object> optionVocabulary() {
@@ -484,7 +488,7 @@ public class TableBindingService {
    private void apply(String sessionToken, Principal user, String assemblyName,
                       Consumer<BaseTableBindingModel> mutation) throws Exception
    {
-      apply(sessionToken, user, assemblyName, mutation, false);
+      apply(sessionToken, user, assemblyName, mutation, false, true);
    }
 
    /**
@@ -507,7 +511,9 @@ public class TableBindingService {
          VSAssembly assembly = rvs.getViewsheet().getAssembly(assemblyName);
          inetsoft.uql.asset.SourceInfo source = assembly instanceof DataVSAssembly data
             ? data.getSourceInfo() : null;
+         String livePercentageBy = livePercentageByValue(assembly);
          mutation.accept(model, rvs, source);
+         TableBindingMutator.preserveUntouchedPercentageBy(model, livePercentageBy);
 
          ApplyVSAssemblyInfoEvent event = new ApplyVSAssemblyInfoEvent();
          event.setName(assemblyName);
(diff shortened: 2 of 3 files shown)


lizhepu:
## Review round 1 — PR #5169 (Java, `stylebi-wiz-main-rebase`)

Head commit reviewed: `1180b247a`.

### Scope check (the sharpest question: does this leak into native Composer UI behavior?)

No. All three changed files (`TableBindingMutator.java`, `TableBindingService.java`,
`TableBindingServiceTest.java`) live entirely under `inetsoft.web.wiz.binding`. The shared
`CrosstabOptionInfo`/`VSCrosstabBindingFactory.updateAssembly()`/`VSBindingModelService`/
`VSBindingService` classes the ~15 native controllers depend on are untouched. Confirmed by
grepping for every caller of `TableBindingMutator`/`TableBindingService` — nothing outside
`TableBindingService` itself reaches these mutators (`ChartBindingMutator` only mentions
`TableBindingMutator` in a doc comment, no actual call). `setOptions` is the only call that
still changes `percentageBy`, wired via the new `preservePercentageBy=false` argument — verified
that path is unaffected by the reset logic for the value it explicitly sets.

The guard's general rule (reset only when the live assembly's own `percentageByValue` was still
`null`) is not keyed to any repro literal — confirmed.

### Important finding — the `setOptions` reset gate is coarser than the invariant it's supposed to enforce

`TableBindingService.setOptions()` passes `preservePercentageBy=false` unconditionally, for
*every* `set_table_options` call — not only calls whose `options` map actually contains
`"percentageBy"`. But `TableBindingMutator.setCrosstabOptions()` only calls
`info.setPercentageByValue(...)` when `options.containsKey("percentageBy")` (line ~912). So a
`set_table_options` call that sets e.g. only `rowTotals`/`colTotals`/`summarySideBySide`/
`suppressGroupTotal` — and never mentions `percentageBy` — skips the reset entirely, while the
model it operates on was still built via `CrosstabOptionInfo(CrosstabVSAssembly)`'s null-to-`"1"`
coalesce (same `binding.createModel(assembly)` call every other mutator goes through). The
result: that manufactured `"1"` is persisted onto the live assembly by `updateAssembly()`,
unconditionally, permanently — reproducing this exact bug via a narrower but very plausible
trigger (any options-only write, e.g. turning on `rowTotals`, before `percentageBy` is ever
touched).

Concretely: crosstab with rows/cols bound, `percentageBy` never set → call
`set_table_options({"rowTotals": true})` only → `get_table_binding` will report
`options.percentageBy: "col"` afterward, exactly the symptom this PR fixes for every other
mutator.

Suggested fix: gate the reset on whether *this specific call* actually supplied `percentageBy`,
not on which method was invoked, e.g.:

```java
public void setOptions(String sessionToken, Principal user, String assemblyName,
                       Map<String, Object> options) throws Exception
{
   boolean changesPercentageBy = options != null && options.containsKey("percentageBy");
   apply(sessionToken, user, assemblyName,
         model -> TableBindingMutator.setOptions(model, options), false, !changesPercentageBy);
}
```

Neither existing regression test catches this — `setOptionsStillAppliesAnExplicitPercentageByChange`
always includes `percentageBy` in its `options` map. Recommend adding a third case: `setOptions`
with only `rowTotals` (no `percentageBy`) on a crosstab whose live `percentageBy` is `null` should
still post `null`, not `"1"`/`"col"`.

### Other items from the brief

- **Guard placement** (`TableBindingService` vs. the brief's literal `TableBindingMutator`
  wording): correct as documented — `TableBindingMutator`'s static methods only see the
  design-time DTO, not the live `VSCrosstabInfo`, so `TableBindingService` (which already fetches
  the live assembly right before `bindingModelService.setBinding`) is the only place with both
  pieces in scope. Verified every one of `TableBindingService`'s crosstab-mutating entry points
  (`setShelf`, `addField`/`removeField`/`moveField` via `applyWithContext`, `setSort`/
  `setRanking`/`setColumnLabels` via the 3-/4-/5-arg `apply` chain, `setOptions` directly) is
  wired to the new reset — no bypass of `TableBindingService` exists elsewhere in the wiz layer.
  The gap that does exist is the one above (granularity within `setOptions`), not a missed call
  site.
- **Skipped regression test #3** (calculateInfo byRow/byColumn lock-step with `percentageBy`):
  acceptable to ship without, per the diagnosis's own conclusion that `syncPercentCalcs()`
  already handles this correctly and this PR doesn't touch that mechanism at all — building new
  integration-test infrastructure for an unrelated, already-working code path is out of proportion
  to this fix. Not blocking.

The scope/placement/leak questions all check out clean. The one important finding above
(`setOptions`'s all-or-nothing reset gate) reopens the same user-facing bug via a different,
easily-reachable trigger and should be fixed with a regression test before this merges.

Per this batch's own rules, this PR is not being merged regardless of outcome — awaiting the
human's own review before merge.

lizhepu:
## Review round 2 (recheck of commit `4a801e0e1`)

Confirmed `4a801e0e1` closes round 1's Important finding (the `setOptions` reset-gate
granularity gap) correctly, with no new issue introduced.

- `setOptions()` now gates `preservePercentageBy` on `options.containsKey("percentageBy")` for
  *this* call, not a fixed literal — traced through the private `apply()` overload
  (`TableBindingService.java:542-565`): when the gate resolves to "preserve", the live assembly's
  actual `percentageByValue` (read before the mutation runs) is written back over the model via
  `TableBindingMutator.preserveUntouchedPercentageBy(...)`, overwriting whatever the model
  snapshot's constructor manufactured. Both branches (call touches `percentageBy` vs. doesn't)
  check out against `TableBindingMutator.setCrosstabOptions()`'s actual `containsKey` guard.
- The new test, `setOptionsWithoutPercentageByStillResetsTheManufacturedDefault`, exercises the
  exact gap: a `set_table_options` call with only `{"rowTotals": true}` (no `percentageBy` key)
  against a model with a manufactured `"1"` and a live `VSCrosstabInfo` whose
  `percentageByValue` is genuinely `null` (verified via `VSCrosstabInfo`'s field initializer),
  asserting the posted value comes back `null`. Not vacuous — fails without this commit's gate.
- No new gap found: a call that includes `percentageBy` alongside other keys still keeps the
  explicit change; the `options == null` path is unaffected; `TableBindingModel` (non-crosstab)
  is untouched either way.

Did not re-run the Maven suite independently — nothing in the diff or the new test looked wrong
on inspection, so the fixer's reported 29/29 (`TableBindingServiceTest`) is trusted as-is.

**Verdict: approve.** Full writeup:
`docs/teams/2026-09-10-bugs-vtb-p1/bug-76574-vtb007/05-review-r2.md` in the `stylebi-wiz` repo.

**Not merged — this still awaits the user's own review/merge decision, per this batch's merge
gate. `gh pr review` continues to be refused on this account setup (same-account restriction), so
this is a plain PR comment rather than a formal review, as in round 1.**

PR #5171, opened 2026-09-11
fix(wiz-hyperlink): SELF target frame, dead bookmarks, shared axis/data-point link storage (bug #76580)

NeverStopFMH:
## Summary

Three findings from bug #76580's backend track (VHL-004/VHL-003/VHL-001), all confirmed by both a debugger and refuter pass (see `docs/teams/2026-09-11-bug-76580-viewsheet-hyperlink/bug-76580-backend/` in the `stylebi-wiz` repo).

- **VHL-004 (P3)**: `describe()` forwarded `HyperlinkDialogModel.getTargetFrame()` unchanged, so a `self:true` link (blanked to `""` by `HyperlinkDialogService`, reported via a separate `self` flag) read back `targetFrame:""` instead of the literal `"SELF"` actually persisted. Fixed by reconstructing `"SELF"` when `self` is true **and** the assembly actually has a link (`linkType != NONE`) — the guard the refuter found missing from the original one-line sketch, matching the real Angular dialog's own guard (`hyperlink-dialog.component.ts:309-310`).
- **VHL-003 (P2)**: `link.bookmark` was silently dropped — never wired for `web`/`message` link types, and silently no-op'd for `viewsheet`-type links when the name didn't exactly match an existing `VSBookmarkInfo`. Fixed in `AssemblyHyperlinkService.set()`, before the runtime is touched: `bookmark` is refused outright for `web`/`message`, and validated against `VSUtil.getBookmarks(assetId, user)` for `viewsheet`, naming the bookmarks that do exist on a miss.
- **VHL-001 (P1)**: a chart field's data-point link and axis-label link share exactly one `Hyperlink` slot per `ChartRef` — a genuine data-model limitation, not a lookup bug, out of scope to fix at the storage layer here. `describe()` now discloses `axisAndDataPointShareLink:true` for an ordinary bound chart field addressed by `colName`, so a caller is told plainly instead of silently losing the other link's destination.

## Test plan

- [x] `./mvnw -pl core test -Dtest=AssemblyHyperlinkServiceTest -DfailIfNoTests=false` — 56/56 pass (49 pre-existing + 7 new)
- [x] `./mvnw -pl core test -Dtest=AssemblyHyperlinkServiceBookmarkValidationTest -DfailIfNoTests=false` — 2/2 pass (new Spring-bootstrapped class for the two cases needing a real `VSUtil.getBookmarks` mock)
- [x] `./mvnw -pl core test -Dtest=ViewsheetAssemblyAgentControllerTest -DfailIfNoTests=false` — 71/71 pass (collateral check, this controller wires `AssemblyHyperlinkService`)

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyHyperlinkService.java
+++ b/core/src/main/java/inetsoft/web/wiz/viewsheet/AssemblyHyperlinkService.java
@@ -23,6 +23,8 @@ import inetsoft.sree.security.ResourceAction;
 import inetsoft.uql.asset.AssetEntry;
 import inetsoft.uql.asset.AssetRepository;
 import inetsoft.uql.schema.XSchema;
+import inetsoft.uql.viewsheet.VSBookmarkInfo;
+import inetsoft.uql.viewsheet.internal.VSUtil;
 import inetsoft.web.binding.drm.DataRefModel;
 import inetsoft.web.composer.model.vs.HyperlinkDialogModel;
 import inetsoft.web.composer.model.vs.InputParameterDialogModel;
@@ -138,6 +140,11 @@ public class AssemblyHyperlinkService {
       // validation above exists to prevent, arrived at by another road.
       String assetId = "viewsheet".equals(type)
          ? resolveViewsheetTarget(link, sessionToken, user) : null;
+
+      // Also validated here, before the runtime is touched, and after assetId so it isn't
+      // resolved twice: a bookmark that HyperlinkDialogService.getHyperlink would otherwise
+      // silently drop (wrong linkType, or a name matching no VSBookmarkInfo) is refused instead.
+      requireValidBookmark(type, link, assetId, user);
 
       sessions.mutate(sessionToken, user, (rvs, runtimeId, dispatcher) -> {
          Region target = region == null ? Region.whole() : region;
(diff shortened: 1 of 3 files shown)


NeverStopFMH:
Review (could not submit as an approving review — GitHub disallows self-approval since I'm also the PR author account). Full findings in `docs/teams/2026-09-11-bug-76580-viewsheet-hyperlink/bug-76580-backend/04-review-r1.md` (stylebi-wiz repo).

Verified all three fixes line-for-line against `01-diagnosis.md`/`02-refute.md`/`03-fix.md`:

- VHL-004: `describe()`'s SELF-reconstruction guard (`model.isSelf() && model.getLinkType() != HyperlinkDialogService.NONE`) matches the refuter's required fix, correctly excluding the never-linked-assembly default-self=true case.
- VHL-003: `requireValidBookmark()` runs after `assetId` is resolved and before `sessions.mutate` (no wasted undo checkpoint). Bookmark-name parsing (including "(Owner)" suffix strip) is a verified exact copy of `HyperlinkDialogService.getHyperlink()`'s own logic.
- VHL-001: `axisAndDataPointShareLink` accurately proxies "resolves through the single-Hyperlink-per-ChartRef path" — verified `model.setColName` is set only in the chart branch, and the title/empty-plot exclusion correctly handles the case where colName survives on the model without denoting shared storage.

No blocking findings. Two non-blocking notes: the never-linked-assembly regression test only asserts `assertNotEquals("SELF", …)` rather than the exact value; no regression test specifically drives a bookmark through the "(Owner)"-suffix strip (low risk, verified exact copy of already-tested logic).

Checks green (build/Semgrep/CLA); `claude-review` fails but that's a standing repo-wide auth/API-key infra issue (is_error after 1 turn, 0 cost, no findings) also present on the last 5 merged PRs. Mergeable, no conflicts. Merged.

PR #5174, opened 2026-09-11
Add test-connection, directory-listing, duplicate, and clear-cache to AdminProviderController (#76602)

fredaan:
## Summary

Redmine #76602: the wiz-plugin-admin "Providers" tool surface had no coverage for four StyleBI Enterprise Manager UI actions — Test Connection, Show Users/Groups/Roles, Duplicate Provider, Clear Security Cache. All four already exist as EM UI actions backed by `AuthenticationProviderService`/`AuthorizationProviderService` methods; the gap was purely missing `AdminProviderController` endpoints (no new backend capability needed).

- **Test connection** — `GET /api/wiz/v1/admin/providers/authentication/{name}/test-connection`. Authentication chain only (no authorization-chain equivalent exists). Resolves the model via `authenticationProviderService.getAuthenticationProvider(name)`, which already forces `oldName := name` on the returned model (`AuthenticationProviderService.java:119`) — this is what makes `replacePlaceholderWithPassword` resolve the REAL stored credential server-side. There is no caller-supplied model parameter at all, so a masked placeholder password can never reach `testConnection` literally.
- **Directory listing** — `GET /api/wiz/v1/admin/providers/authentication/{name}/directory/{kind}` (`kind`: users/groups/roles). Same placeholder-safety shape as test-connection. Authentication chain only.
- **Duplicate** — new `verb: "duplicate"` in the existing `ProviderChangeRequest`/`ProviderChangePlanService`/`ProviderChangesetApplyService` plan+apply+rollback machinery, on BOTH chains. Reuses `Util.getCopyName`/`getNextCopyName` — the same naming algorithm `AuthenticationProviderController.copyAuthenticationProvider`/`AuthorizationProviderController.copyAuthorizationProviderCache` already use — rather than reimplementing it. Accepts an optional explicit `newName` (a colliding explicit name is refused loud). Unlike `create`, is NOT restricted to FILE/LDAP providerType, since it copies an existing, already-licensed provider rather than spinning up a fresh, unvetted one. A duplicate's own rollback deletes the copy by its resolved name; the source is never touched (reuses the existing `Undo.createdAuthentication`/`createdAuthorization` kind, no new rollback path needed).
- **Clear cache by name** — `POST /api/wiz/v1/admin/providers/{authentication,authorization}/{name}/clear-cache`. The underlying `clearAuthenticationProviderCache`/`clearAuthorizationProviderCache` are index-only; this resolves name→index via `ProviderChangesetApplyService.indexOfName` (promoted from `private` to package-visible, not duplicated). A provider whose type does not enable caching (most notably FILE) is refused loud with a field-named error rather than silently reporting success for `CachableProvider.clearCache()`'s own no-op default.

## Test plan

- [x] `AdminProviderControllerTest` (new file) — happy paths for all 4 new endpoints, the oldName/placeholder-forcing behavior, FILE-provider refuse-loud for clear-cache, kind-dispatch validation for directory listing.
- [x] `ProviderChangePlanServiceTest` — duplicate verb resolution: auto-generated copy name, explicit `newName`, collision refusal, source-not-found, providerType/spec rejection, works on both chains.
- [x] `ProviderChangesetApplyServiceTest` — duplicate apply (both chains, auto/explicit name) and rollback (copy removed, source untouched).
- [x] Ran `./mvnw -pl core -Dtest=AdminProviderControllerTest,ProviderChangePlanServiceTest,ProviderChangesetApplyServiceTest test` — all pass (68 tests in this targeted run, 0 failures).

## Cross-repo dependency

This PR's new endpoints are called by new tools in the companion `stylebi-wiz` PR (`fix/76602-provider-admin-actions` → `main`). That PR's `test_auth_provider_connection`/`list_auth_provider_directory_entries`/`clear_provider_cache` tools and the `duplicate` verb will 404 until THIS PR merges. Link: (stylebi-wiz PR to be added).

### DIFF
--- a/core/src/main/java/inetsoft/web/admin/ai/providers/AdminProviderController.java
+++ b/core/src/main/java/inetsoft/web/admin/ai/providers/AdminProviderController.java
@@ -30,6 +30,7 @@ import org.springframework.web.bind.annotation.*;
 import org.springframework.web.server.ResponseStatusException;
 
 import java.security.Principal;
+import java.util.List;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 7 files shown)


fredaan:
Companion stylebi-wiz PR (the TS tools that call these new endpoints): https://github.com/inetsoft-technology/stylebi-wiz/pull/2437

Redmine: #76602

fredaan:
## Review round 1 — reviewer pass (adversarial, per review-workflow)

**Head commit reviewed:** `4628a37236936d74c6ed542f9023fe40dfd41c72`

### What was checked

Reviewed against `01-diagnosis.md`/`02-refute.md`/`03-fix.md`, checking the CONFIRMED CAUSE rather than only the reported symptom:

- **Placeholder-password structural safety (Test Connection / directory listing):** `testAuthenticationProviderConnection`/`getAuthenticationProviderDirectory` take only a `name` path variable — no request body/model parameter exists for a caller to smuggle a stale/masked password through. Both resolve the model via `authenticationProviderService.getAuthenticationProvider(name)`, which sets `oldName := name` (confirmed at `AuthenticationProviderService.java:119`), which is what forces `replacePlaceholderWithPassword` (`:942-962`) to substitute the REAL stored credential rather than trust a literal `"**********"`. This closes the hazard by construction, not by validation — agrees with `03-fix.md`'s own claim.
- **Duplicate's own password carry-over**, traced end to end: `resolveDuplicate`/`applyDuplicateAuthentication` fetch the source via `getAuthenticationProvider(name)` (masked password, `oldName = name`), then `((ImmutableAuthenticationProviderModel) source).withProviderName(newName)`. Confirmed `withProviderName` only changes `providerName`, leaving `oldName` pointed at the ORIGINAL source name — which is exactly what makes `addAuthenticationProvider` → `getProviderFromModel` → `replacePlaceholderWithPassword` resolve the source's real stored credential onto the copy. Not incidental; this is the correct, deliberate mechanism, and it is exercised by `resolveDuplicateAuthenticationAutoGeneratesCopyName` et al.
- **Duplicate naming**: `resolveDuplicateName` calls through to `Util.getCopyName`/`Util.getNextCopyName` (not reimplemented), matches the exact collision-loop the real `copyAuthenticationProvider`/`copyAuthorizationProviderCache` controllers already use, and is recomputed fresh at apply time against the live list (never trusting a preview-captured name) — same "resolved fresh" discipline `applyDeleteAuthentication` already documents for its own index.
- **Duplicate rollback**: correctly reuses `Undo.createdAuthentication`/`createdAuthorization` (no new `Undo.Kind` needed, since a duplicate's own undo is structurally a plain create's undo — delete the new name). `rollbackOfADuplicateRemovesTheCopyNotTheSource` genuinely forces an unknown-state rollback (post-create verification throws) and asserts the source survives while only the copy is removed.
- **DATABASE/CUSTOM exclusion correctly NOT applied to duplicate**: confirmed `resolveDuplicate` is a separate branch checked before `resolveCreate` is ever reached, so `create`'s type restriction structurally cannot fire for a duplicate. `checkProviderTypeLicensed` still runs inside `addAuthenticationProvider`/`addAuthorizationProvider` regardless of verb, so a delicensed DATABASE/CUSTOM duplicate still fails loud (caught, reported FAILED) rather than silently bypassing licensing — the exclusion's removal doesn't reopen the license-gating risk it exists for.
- **Clear Security Cache**: confirmed `indexOfName` is promoted (not duplicated) and reused by the new endpoints; confirmed `getProviderListModel().providers()` is built by streaming the exact same `chain.getProviders()` list that `getAuthenticationChain().getProviders().get(index)` indexes into (`AuthenticationProviderService.java:230-244` vs. `:328-336`), so the two separate reads correlate soundly, matching the same index-then-use pattern the pre-existing delete path already relies on. `requireCacheable` refuses loud (not silent no-op) for a non-cacheable provider, generically via `CachableProvider.isCacheEnabled()` rather than special-casing FILE by name — correctly covers any future non-cacheable type too.
- **Old create-verb placeholder hazard** (the diagnosis's own cross-cutting note: `get_auth_provider` + `apply_provider_changes(verb:"create")` could still carry a literal placeholder password): confirmed unchanged by this PR, and confirmed it fails LOUD (not silently) — `applyCreateAuthentication` never sets `oldName`, so `replacePlaceholderWithPassword` does not fire, and the literal `"**********"` gets rejected by `checkParameters()`/`testConnection()`, caught and reported as a clean FAILED status. Leaving this path unguarded is reasonable now that `duplicate` gives a safe alternative for the scenario that used to motivate the workaround.
- Every new endpoint has its own `@Secured`/`requireSiteAdmin` gate matching the existing 6 endpoints' convention. `AdminProviderControllerTest` mocks `AdminAiCallerGuard`'s bearer-token check + `OrganizationManager.isSiteAdmin` correctly (mirrors `AdminPropertiesControllerTest`'s own precedent).
- Sibling/boundary scenarios tried and not found broken: authorization-chain siblings for test-connection/directory-listing (correctly absent, no such product capability); FILE-provider clear-cache on both chains (both refuse loud, tests confirm both); explicit-`newName` collision (refused loud, naming the collision); empty/duplicate provider list boundaries (`resolveDuplicateAuthenticationSourceNotFoundThrows`).

### Findings

No blocker or important findings. This PR is a single, clean commit scoped to bug 76602, unlike its companion `stylebi-wiz` PR #2437 (which carries 2 unrelated commits from a different bug — flagged there, not here).

### Verdict: no findings — would approve, but this PR is authored by the same account running this review, so GitHub refuses a self-approval. Recorded here as the round's verdict for the human/lead to action.

PR #5175, opened 2026-09-11
fix(76489): fall through to session-cookie auth on a stale JWT with an active session

larryliang-inetsoft:
## Summary

Redmine #76489, Finding 3: `WizServiceAuthenticationFilter` hard-401'd unconditionally on any invalid/expired `wiz_auth` JWT (an SSO signing-key rotation, or ordinary token expiry) -- even when the browser separately held a still-valid StyleBI session. This turned any JWT staleness into a permanent, silent "Generating..." hang in the Composer's AI-agent pairing flow, since the client (fixed separately in stylebi#5046) had nothing to recover from -- there was no valid session to fall back to.

## Root cause and the security wrinkle

The obvious fix -- fall through to `chain.doFilter()` on an invalid JWT, mirroring the existing "no token provided" path -- would have reopened a real gap: `InvalidateSessionFilter` skips its own stale/deactivated-user check (`isActiveUser`) entirely for any `isWizRequest()==true` request (see that filter's own doc comment), trusting `WizServiceAuthenticationFilter` to be the sole authoritative gate for these requests. A naive fallthrough would let an already-deactivated user's stale session ride through with **neither** filter's check ever having run.

## Fix

`catch(WizAuthenticationException e)` now calls a new `hasValidActiveSession(request)`, which mirrors `InvalidateSessionFilter`'s own check (`SUtil.getPrincipal(request, false)` + `securityEngine.isActiveUser(...)`, same `isSecurityEnabled()` bypass) before falling through. An inactive/deactivated principal, no session, or no principal at all still 401 exactly as before -- only a genuinely still-active existing session is allowed to carry the request onward.

## Tests

`WizServiceAuthenticationFilterTest.java` (no test existed for this filter before this PR) -- 8 cases:
- 2 preserved-unchanged paths (not a wiz request, no Authorization header)
- the new fallthrough (active session)
- 4 cases proving the gap is **not** reopened (no session, no principal, inactive principal, security-engine-unavailable race)
- the security-disabled bypass

Full `inetsoft.web.security` + `inetsoft.web.wiz.security` package suite (14 test classes, 143 tests) run and green -- no regressions.

## Review notes

This touches authentication code. Findings 1 (nginx) and 2 (client-side hang, stylebi#5046) are already merged and live-verified against the running dev deployment; this PR is the third and final piece, and was explicitly called out by the original investigation as needing security review before anyone changed this filter's behavior -- please review the `hasValidActiveSession` design (mirroring, not bypassing, `InvalidateSessionFilter`'s own check) closely.

Full archive: `docs/teams/2026-09-07-bugs-76489-76490-env/` (stylebi-wiz repo), see STATUS.md's "Finding 3 fix" section.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/security/WizServiceAuthenticationFilter.java
+++ b/core/src/main/java/inetsoft/web/wiz/security/WizServiceAuthenticationFilter.java
@@ -25,6 +25,7 @@ import com.nimbusds.jwt.SignedJWT;
 import inetsoft.sree.ClientInfo;
 import inetsoft.sree.RepletRepository;
 import inetsoft.sree.SreeEnv;
+import inetsoft.sree.internal.SUtil;
 import inetsoft.sree.security.*;
 import inetsoft.sree.web.SessionLicenseServiceProvider;
 import inetsoft.util.*;
@@ -41,6 +42,7 @@ import org.springframework.stereotype.Component;
 
 import java.io.IOException;
 import java.security.KeyPair;
+import java.security.Principal;
 import java.security.interfaces.RSAPublicKey;
 import java.text.ParseException;
 import java.util.*;
@@ -117,6 +119,25 @@ public class WizServiceAuthenticationFilter extends AbstractSecurityFilter {
       }
       catch(WizAuthenticationException e) {
          LOG.warn("WIZ service authentication failed: {}", e.getMessage());
+
+         // The wiz_auth cookie's JWT is stale (expired, signature mismatch from a key
+         // rotation, etc.), but this browser may separately hold a still-valid StyleBI
+         // session (e.g. an ordinary interactive login) alongside the now-stale cookie.
+         // Hard-401'ing unconditionally here turns that into a permanent, silent hang for
+         // anything routed through this filter (bug #76489) even when a perfectly good
+         // session already exists. Falling through to let that session handle the request
+         // is only safe if its own principal is still a genuinely active user: this filter
+         // is InvalidateSessionFilter's ONLY gate for any isWizRequest()==true request --
+         // that filter skips its own stale/deactivated-user check entirely whenever
+         // isWizRequest() is true, trusting this filter to be authoritative (see
+         // AbstractSecurityFilter#isWizRequest's own doc comment). So the equivalent check
+         // must run HERE before falling through, or an already-deactivated user's stale
+         // session could ride through unauthenticated-but-unrejected.
+         if(hasValidActiveSession(httpRequest)) {
+            chain.doFilter(request, response);
+            return;
+         }
+
          sendUnauthorized(httpResponse, "Unauthorized");
          return;
       }
@@ -157,6 +178,35 @@ public class WizServiceAuthenticationFilter extends AbstractSecurityFilter {
     */
    private boolean isWizAuthEnabled() {
       return "true".equalsIgnoreCase(SreeEnv.getProperty(WIZ_AUTH_ENABLED_PROPERTY, "true"));
+   }
+
+   /**
+    * Mirrors {@link inetsoft.web.security.InvalidateSessionFilter}'s own active-user check --
+    * that filter skips it entirely for any {@code isWizRequest()} request, trusting this filter
+    * to be the sole gate (see this class's own doc comment and
+    * {@code AbstractSecurityFilter#isWizRequest}'s). When the JWT itself is invalid or expired,
+    * falling through to let an already-established StyleBI session handle the request is only
+    * safe if that session's own principal is still genuinely active -- otherwise a request from
+    * an already-deactivated or logged-out user could slip through with neither this filter's own
+    * JWT check nor InvalidateSessionFilter's (skipped) check ever having verified it.
+    *
+    * @return {@code true} if an existing session already carries a principal that is either a
+    *         genuinely active user, or security is disabled entirely (mirroring
+    *         InvalidateSessionFilter's own {@code isSecurityEnabled()} gate around its check).
+    */
+   private boolean hasValidActiveSession(HttpServletRequest request) {
+      Principal existing = SUtil.getPrincipal(request, false);
+
+      if(!(existing instanceof SRPrincipal)) {
+         return false;
+      }
+
+      if(!isSecurityEnabled()) {
+         return true;
+      }
+
+      SecurityEngine securityEngine = getSecurityEngine();
+      return securityEngine != null && securityEngine.isActiveUser((SRPrincipal) existing);
    }
 
    /**
--- a/core/src/test/java/inetsoft/web/wiz/security/WizServiceAuthenticationFilterTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/security/WizServiceAuthenticationFilterTest.java
@@ -1,0 +1,252 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (changes requested):
## Review

*(Manual review — the automated `claude-review` check errored out with the known "1 turn, $0 cost" execution-failure signature, most likely a usage-limit hit at the time it ran, so it never actually evaluated this PR. This review replaces it.)*

### Summary of the change

`WizServiceAuthenticationFilter.doFilter()`'s `catch(WizAuthenticationException e)` block previously hard-401'd on any invalid/expired `wiz_auth` JWT. This PR adds a `hasValidActiveSession(request)` check before that 401, and falls through to `chain.doFilter()` when the browser separately holds a still-valid StyleBI session, fixing the silent Composer "Generating..." hang from bug #76489.

### What's solid

- **`hasValidActiveSession` correctly mirrors `InvalidateSessionFilter`'s own gate.** Side by side: both call `SUtil.getPrincipal(request, false)` (no session creation), both gate on `isSecurityEnabled()` before doing anything stricter, both null-check `getSecurityEngine()` before calling `isActiveUser()`. The new method is actually slightly more defensive than the original (an `instanceof SRPrincipal` check instead of `InvalidateSessionFilter`'s unchecked cast), which is a safe improvement, not a behavior change.
- Confirmed against `AbstractSecurityFilter`/`InvalidateSessionFilter` on `stylebi-wiz-main-rebase` (the actual PR base) that `InvalidateSessionFilter` does skip its stale/deactivated-user check whenever `isWizRequest()` is true (`!isWizRequest(httpRequest)` in its outer `if`), so the PR description's claim that this filter is the *sole* gate for wiz requests is accurate, not overstated.
- The fallthrough re-uses the same `chain.doFilter(request, response)` (unwrapped request) as the pre-existing "no Authorization header" fallthrough a few lines above, so it lands on an already-proven code path (`RequestPrincipalFilter` resolves the session principal downstream) rather than inventing a new one.
- Exceptions other than `WizAuthenticationException` (the generic `catch(Exception e)` branch) are untouched and still hard-401, so genuine internal errors don't get a new escape hatch — only a validation *failure* does.
- Test suite (`WizServiceAuthenticationFilterTest`) is thorough and each case is meaningful, not padding: unchanged paths, the new fallthrough, and — the load-bearing one — `doFilter_invalidJwt_existingInactiveSession_stays401_neverCallsChain`, which proves the gap `InvalidateSessionFilter`'s skip depends on isn't reopened. The `securityEngineUnavailable` race case is a nice touch mirroring `InvalidateSessionFilterTest`.
- CI: `build`, `Semgrep Scan`, `Semgrep OSS`, and `license/cla` are all green. `claude-review` is the only failing check, and it failed with 1 turn / $0 cost / immediate error — the documented signature for a usage-limit/execution failure rather than a real finding.

### Blocking concern: this reopens a gap in `AdminAiCallerGuard`'s CSRF backstop

`CSRFFilter.isWizApi()` exempts **all** of `/api/wiz/**` from CSRF protection, on the documented assumption that "WIZ API uses JWT authentication, so CSRF protection is not needed." The admin-chat endpoints (`AdminAiController`, `AdminClusterController`, `AdminFileBackupController`, `AdminLicensingController`, `AdminPresentationController`, `AdminProviderController`, `AdminScheduleController`, `AdminPropertiesController` — all mapped under `/api/wiz/v1/admin/**`) rely on that CSRF exemption, backstopped only by `AdminAiCallerGuard.requireBearerAuthenticatedRequest()`.

Critically, that guard's own doc comment (`core/src/main/java/inetsoft/web/admin/ai/AdminAiCallerGuard.java`) explains it checks **header presence**, not JWT validity, and spells out exactly why that used to be safe:

> "A cross-site form POST cannot set an `Authorization` header, and a scripted request that tries requires a CORS preflight an attacker cannot satisfy."

That argument implicitly depended on a second, unstated invariant that held until this PR: *if the header is present but the JWT doesn't validate, `WizServiceAuthenticationFilter` 401s before the request ever reaches the guard.* So "header present" and "JWT actually validated" were equivalent by construction. `AdminAiEndpointPrefixTest`'s own doc comment ("Every mutating admin-chat endpoint sits on the CSRF-exempt `/api/wiz/**` prefix, which is why `AdminAiCallerGuard` requires a bearer token on each of them") and `AdminAiControllerMappingTest` both encode this same assumption.

This PR breaks that equivalence: with an **active StyleBI session** present (exactly the condition this fix targets) and **any** non-blank `Authorization: Bearer ...` header — expired, garbage, wrong signature, doesn't matter — the request now falls through instead of 401ing. It reaches `AdminAiCallerGuard`, which only checks that the header is present, and passes. The request then executes as the session's own principal (subject to `@Secured`/permission checks, which are real and still enforced) rather than as a validated wiz-broker identity.

Concretely: a legitimate signed-in Site Administrator, with a stale/expired `wiz_auth` cookie sitting alongside their normal session (plausible any time after an SSO key rotation, or just JWT expiry — the exact trigger condition this bug is about), can now have a mutating admin-chat request (`/api/wiz/v1/admin/apply`, `/backup`, etc.) succeed via session auth while carrying a bogus bearer token, on a path that was specifically designed and tested to require a *validated* bearer token as its CSRF backstop.

**Mitigating factor, why this isn't an immediate 0-day:** I searched the codebase and found no CORS configuration (`CorsConfiguration`, `addAllowedOrigin`, etc.) anywhere, so the specific "third-party site rides the victim's cookies with a forged header" CSRF exploit still requires a CORS preflight the server won't satisfy — the same protection `AdminAiCallerGuard`'s comment already leans on. So this isn't trivially exploitable by an external site today. But:
- It's a real regression in defense-in-depth for a control that was deliberately built as a second layer specifically because CSRF protection was turned off for this prefix.
- It's fragile: it now depends on CORS configuration staying absent forever, which is a much weaker guarantee than "the filter validated the JWT."
- None of the 8 new tests, nor any existing test in this PR, exercises the interaction between the new fallthrough and `AdminAiCallerGuard`/the admin-chat endpoints specifically — the PR's own test suite doesn't catch this because it tests `WizServiceAuthenticationFilter` in isolation.
- The PR description explicitly asks for close review of exactly this design because it "touches authentication code," so I think this is worth surfacing rather than deferring.

**Suggested fixes (any one would close the gap):**
1. Narrow the new fallthrough so it doesn't apply to the admin-chat prefix (e.g. don't treat `hasValidActiveSession` as sufficient when `isPageRequested("/api/wiz/v1/admin/**", request)`, keeping the hard 401 there), or
2. Harden `AdminAiCallerGuard` to also require `SRPrincipal.isWizPrincipal(...)` on the resolved principal in addition to header presence (with the header-presence check kept as-is for the reason the class doc already explains about session reuse), or
3. At minimum, add a regression test proving a stale/garbage bearer token + active admin session cannot reach an admin-chat mutation endpoint, and update `AdminAiCallerGuard`'s doc comment to acknowledge the new fallthrough explicitly (right now it's now factually stale: "the token is ignored and the request falls through to session auth" is described there only for the *off-prefix* case, but is now also true *on*-prefix for an invalid token).

### Minor / non-blocking

- Base branch is `stylebi-wiz-main-rebase`, not `main` — presumably intentional for this wiz integration line, flagging only for awareness since it's outside the normal branch-naming convention in `BRANCHING.md`.
- Javadoc on `hasValidActiveSession` is accurate and clear; no notes there.

### Verdict

Requesting changes — not because the core `hasValidActiveSession` mirroring is wrong (it's correct and well-tested for its stated goal), but because the change has an unaddressed side effect on `AdminAiCallerGuard`'s CSRF backstop for the admin-chat endpoints that should be closed or explicitly accepted-and-documented before merging security-sensitive filter code.

PR #5176, opened 2026-09-11
Bug #76560 follow-up: fix 0-row table height mismatch in ScrollableTableDirective

larryliang-inetsoft:
## Summary

Follow-up to the already-merged fix for bug 76560 (PR #5159), found by a human tester (Sally Ma) while verifying it in the running app.

PR #5159 fixed `ScrollableTableDirective`'s dead `tbody.style.height == null` guard, so the height calculation now correctly runs when a table has at least one row, locking that table's body to the CSS `max-height` (50vh). However, `setColumnWidths()`'s early-return branch for tables with 0 rows skipped that height assignment entirely, so an empty table fell back to the CSS `min-height` (40vh) instead of matching a sibling non-empty table's 50vh — a visible height mismatch between two side-by-side lists (e.g. the Filters/Shared Filters lists in Dashboard Options when one is empty and the other isn't).

## Fix

Extracted the height calculation into a new `setBodyHeight()` method and call it from both the 0-row early-return branch and the main path, so both compute the height identically instead of one of them silently no-oping.

## Testing

`ng build portal --configuration development` succeeds clean.

Fixes https://173.220.179.100/issues/76560 (follow-up finding, ticket reopened after the original PR #5159 merged)

### DIFF
--- a/web/projects/portal/src/app/widget/scrollable-table/scrollable-table.directive.ts
+++ b/web/projects/portal/src/app/widget/scrollable-table/scrollable-table.directive.ts
@@ -51,6 +51,7 @@ export class ScrollableTableDirective implements AfterContentInit, AfterContentC
             headCells[i].style.width = `${thead.offsetWidth / headCells.length}px`;
          }
 
+         this.setBodyHeight();
          return;
       }
 
@@ -68,22 +69,7 @@ export class ScrollableTableDirective implements AfterContentInit, AfterContentC
          totalw += w;
       }
 
-      const tableStyle = window.getComputedStyle(table);
-
-      if(!tbody.style.height) {
-         let tableHeight: string;
-
-         if(tableStyle.maxHeight && tableStyle.maxHeight !== "none") {
-            tableHeight = tableStyle.maxHeight;
-         }
-         else {
-            tableHeight = tableStyle.height;
-         }
-
-         if(tableHeight) {
-            tbody.style.height = `calc(${tableHeight} - ${thead.offsetHeight}px)`;
-         }
-      }
+      this.setBodyHeight();
 
       const scroll = tbody.scrollHeight > table.offsetHeight - thead.offsetHeight;
       const bodyw = Math.min(tbody.offsetWidth, table.parentElement.offsetWidth);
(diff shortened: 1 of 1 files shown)


jshobe-inetsoft (approved):
## Manual Review (replaces failed automated `claude-review` check)

The automated `claude-review` GitHub Actions check for this PR failed to execute (`is_error: true`, `num_turns: 1`, `total_cost_usd: 0` — the known signature for the review agent hitting a usage limit before it could analyze anything, not a substantive finding). This is a manual review performed in its place.

### Summary of the change

`web/projects/portal/src/app/widget/scrollable-table/scrollable-table.directive.ts` — extracts the body-height calculation out of `setColumnWidths()`'s main path into a new private `setBodyHeight()` method, and calls it from **both** the 0-row early-return branch and the main (≥1 row) path. Previously the 0-row branch returned before the height calculation ran, so an empty table fell back to the CSS `min-height` (40vh) instead of the `max-height` (50vh) that a populated sibling table would get, producing a visible height mismatch between side-by-side lists (e.g. Filters/Shared Filters in Dashboard Options).

### Correctness

- The extracted `setBodyHeight()` body is a verbatim move of the previously-inline logic (`tableStyle.maxHeight`/`height` → `calc(...)` assigned to `tbody.style.height`, guarded by `!tbody.style.height`) — confirmed via diff, no logic was altered, only relocated. No behavior change for the previously-working (non-empty) path.
- Verified against the sibling CSS (`filters-pane.component.scss`): `min-height: 40vh; max-height: 50vh;` — matches the bug description exactly, and the fix (running the same `max-height`-aware calc for 0-row tables) correctly addresses the mismatch.
- Verified against the referenced prior fix (PR #5159 / commit d462dce65) — this is a legitimate, narrowly-scoped follow-up to it, not a duplicate or unrelated change.
- `ngAfterContentInit()`'s own 0-row early return (a separate code path, untouched by this PR) still doesn't set body height on the very first check, but `ngAfterContentChecked()` → `setColumnWidths(false)` runs immediately after via Angular's content-check lifecycle and will apply it, so there's no user-visible gap. This matches pre-existing behavior for that lifecycle hook and isn't a regression introduced here.

- `setBodyHeight()` re-queries `table.querySelector("thead"/"tbody")` even though the caller (`setColumnWidths`) already has `thead`/`tbody` in scope. Passing them as parameters would avoid the redundant DOM lookups, though the cost is negligible here.
- No unit test was added for the 0-row case specifically (the existing spec only covers a populated table). Given this is a visual/CSS-interaction fix verified manually against the running app (per the PR description) and the existing spec suite is minimal for this directive already, this is a reasonable trade-off rather than a blocking gap.

### Other checks

- No merge conflicts; PR is `MERGEABLE`.
- All CI checks other than `claude-review` are green (`build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla`).
- No security concerns — pure client-side layout/CSS calculation, no user input or injection surface.
- No backward-incompatibility — internal directive-only change, no public API/template contract change.
- No test regressions — existing spec (`scrollable-table.directive.spec.ts`) only asserts the `scrollable-table` class is applied; unaffected by this change.

### Verdict

PR #5177, opened 2026-09-11
Fix crosstab summarySideBySide silent no-op (#76574, VTB-008)

lizhepu:
## Summary
- `VSCrosstabBindingFactory.updateAssembly()` called `VSCrosstabInfo.setSummarySideBySide(boolean)` -- the RUNTIME-value overload (`sideBySideValue.setRValue(...)`) -- instead of `setSummarySideBySideValue(boolean)`, the DESIGN-value overload its three sibling lines (percentageBy/rowTotals/colTotals) already correctly use.
- The runtime value is nulled by `resetRuntimeValues()` on every render/export path (confirmed traced end-to-end: `get_viewsheet_image` -> `VSExportService.doExportViewsheet` -> `CoreLifecycleService.refreshViewsheet` -> `VSUtil.resetRuntimeValues`), so the write never survived to affect a render. `set_table_options(summarySideBySide: true)` reported `ok:true` and `get_table_binding` echoed it back correctly, but the crosstab always rendered as if `false` (measures stacked in rows instead of side-by-side columns).
- The native Composer UI's `CrosstabPropertyDialogService.java:303` already calls the correct `setSummarySideBySideValue` for this same field, confirming this is a StyleBI-wide working feature, with the defect confined to the wiz/Composer-plugin binding-write path.
- One-line fix: `crossInfo.setSummarySideBySide(...)` -> `crossInfo.setSummarySideBySideValue(...)`.
- Eyeballed the rest of `updateAssembly`'s field-copying block for the same wrong-overload shape on other fields -- found no other instance; everything else already uses the correct `...Value` setters.

## Test plan
- [x] Added `VSCrosstabBindingFactoryTest.updateAssemblySetsTheDesignValueForSummarySideBySide`, asserting on `VSCrosstabInfo.getSummarySideBySideValue()` (the design getter) rather than the runtime-aware `isSummarySideBySide()`, since the latter would pass spuriously even against the old broken code.
- [x] Confirmed the new test fails against the pre-fix code (`expected: <true> but was: <false>`) and passes with the fix.
- [x] `TableBindingMutatorTest` (67 tests) and `TableBindingServiceTest` (29 tests) still pass -- no regression in the model-layer read/write path this fix doesn't touch.
- [ ] No live self-check via `mcp__composer-chat__*` tools was possible in this session (no MCP tools were exposed) -- flagged per the task's own instruction rather than skipped silently.

### DIFF
--- a/core/src/main/java/inetsoft/web/binding/service/VSCrosstabBindingFactory.java
+++ b/core/src/main/java/inetsoft/web/binding/service/VSCrosstabBindingFactory.java
@@ -282,7 +282,7 @@ public class VSCrosstabBindingFactory
       crossInfo.setPercentageByValue(option.getPercentageByValue());
       crossInfo.setRowTotalVisibleValue(option.getRowTotalVisibleValue());
       crossInfo.setColTotalVisibleValue(option.getColTotalVisibleValue());
-      crossInfo.setSummarySideBySide(option.isSummarySideBySide());
+      crossInfo.setSummarySideBySideValue(option.isSummarySideBySide());
       syncPercentCalcs(crossInfo);
       VSUtil.updateCalculate(crossInfo, null);
       return assembly;
--- a/core/src/test/java/inetsoft/web/binding/service/VSCrosstabBindingFactoryTest.java
+++ b/core/src/test/java/inetsoft/web/binding/service/VSCrosstabBindingFactoryTest.java
@@ -1,0 +1,81 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.web.binding.service;
+
+import inetsoft.test.BaseTestConfiguration;
+import inetsoft.test.ConfigurationContextInitializer;
+import inetsoft.test.SreeHome;
+import inetsoft.uql.viewsheet.CrosstabVSAssembly;
+import inetsoft.uql.viewsheet.VSCrosstabInfo;
+import inetsoft.web.binding.model.table.CrosstabBindingModel;
+import inetsoft.web.binding.model.table.CrosstabOptionInfo;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


lizhepu (commented):
Reviewed against the diagnosis/refutation docs (bug 76574, VTB-008) and the actual diff. Clean — no blockers.

**Scope check (passed):** the diff is exactly the prescribed one-line change — `crossInfo.setSummarySideBySide(...)` → `crossInfo.setSummarySideBySideValue(...)` at `VSCrosstabBindingFactory.java:285` — plus one new test file. Confirmed the three sibling lines (`setPercentageByValue`/`setRowTotalVisibleValue`/`setColTotalVisibleValue`) are untouched. `gh api compare` shows exactly 1 commit and 2 files changed vs `stylebi-wiz-main-rebase`, so nothing from a stale branch point leaked in.

**Base branch (passed):** PR targets `stylebi-wiz-main-rebase`, correctly, not `main`.

**Test assertion (passed, verified directly in the diff):** `VSCrosstabBindingFactoryTest.updateAssemblySetsTheDesignValueForSummarySideBySide` asserts `crossInfo.getSummarySideBySideValue()` — the DESIGN getter — not `isSummarySideBySide()`. This is the assertion that actually catches the regression; the runtime getter would have passed spuriously even pre-fix, per the diagnosis's own reasoning. The Spring/`@SreeHome` test scaffolding matches the existing `CrosstabVSAssemblyTest.java` pattern exactly.

**Fail-before/pass-after claim:** can't be independently verified from the PR itself — there's only one commit (no separate revert-and-rerun commit or CI log showing the pre-fix failure), and the `claude-review` check errored out with no content. Taking this on faith is reasonable here since there's no artifact to check against, but flagging that it is unverified from the PR alone.

**Sibling-invariant check (independently re-verified, not just trusted):** grepped the *live* `E:\inetsoft\wizard\community` checkout (still on the pre-fix line, confirming it's a different worktree from the fixer's) for the four sibling boolean option fields on `VSCrosstabInfo` (`setFillBlankWithZero`/`setMergeSpan`/`setSortOthersLast`/`setCalculateTotal`, each with the same RValue/DValue overload split as `summarySideBySide`) across `VSCrosstabBindingFactory.java` and the other binding factories (`VSCalcTableBindingFactory`, `VSChartBindingFactory`, `VSTableBindingFactory`, `VSBindingFactory`). None of them are called from any binding factory at all — so there is no other wrong-overload instance lurking. Confirms the fixer's "eyeballed, found none" claim rather than just repeating it.

**Live self-check:** none was possible in that session (no MCP tools exposed), disclosed honestly in the PR body. Given this is a single-overload swap that exactly mirrors an already-correct, already-shipped reference implementation (`CrosstabPropertyDialogService.java:303`) operating on the very same object, and the render-path trace (`get_viewsheet_image` → `VSExportService.doExportViewsheet` → `CoreLifecycleService.refreshViewsheet` → `VSUtil.resetRuntimeValues`) was independently re-derived by the refuter via direct file/line reads rather than just inferred, I'd call this safe to merge without a live check — but a live `set_table_options`/`get_viewsheet_image` round trip post-merge (or before, if convenient) would still be worth doing opportunistically since the diagnosis itself flags the pixel-level render outcome as the one thing static tracing can't confirm.

Overall: clean, no fix round needed. Per this batch's merge gate, still awaiting the user's own explicit approval before merge.

Nelson-cao (approved):

PR #5178, opened 2026-09-11
fix(wiz): preserve dimension sort/ranking across a shelf rewrite (#76574 VTB-004)

lizhepu:
## Summary
- `TableBindingMutator.setShelf` rebuilt a crosstab shelf's `rows`/`cols`/`groups` dimension list from scratch on every write (`set_table_fields`, `add_table_field`, `remove_table_field`, `move_table_field` all funnel through it), so a field's `order`/`sortByCol`/`rankingOpt`/`rankingN`/`rankingCol`/`manualOrder`/`groupOthers`/`namedOthers` reset to class defaults even when nothing about that field changed — including a plain resubmission of the identical field list (Redmine #76574 / VTB-004).
- Fix: match a new field to the shelf's own previous list by **same absolute index + same column (case-insensitive) + same date level**, and carry its sort/ranking forward when matched. This is strategy (a) from the diagnosis/refutation pair in `docs/teams/2026-09-11-bugs-vtb-p2/bug-76574-vtb004/` — chosen over the diagnosis's own initial recommendation of a content-keyed match, per the refuter's correction that `requireDimension`/`pruneOrphanedSuppression`/the plugin's own `set_field_sort` tool contract all key duplicate-column identity by absolute shelf position, not date-level content.
- Known, accepted boundary of this strategy (documented in a regression test, not silently accepted): an insert/remove of an *earlier* field on the same shelf shifts a *later*, untouched field's absolute index, and that field's sort/ranking is not preserved across the shift.
- Out of scope, noted only: the sibling `aggregates()` helper has the same "always build a fresh ref" shape for `calculateInfo` — flagged as a candidate follow-on, not fixed here.

## Test plan
- [x] `mvn -pl core test -Dtest=TableBindingMutatorTest` — 74 run (67 existing + 7 new), 0 failures
- [x] `mvn -pl core test -Dtest=TableBindingServiceTest,ChartBindingMutatorTest` — 29 / 28 run, 0 failures (regression check on adjacent tests exercising the same mutator)
- [x] `mvn -pl core test -Dtest=BindingAgentControllerTest` — 19 run, 0 failures
- [ ] Live plugin-tool round trip (`set_field_sort`/`set_field_ranking` → `get_table_binding` → `set_table_fields` resubmit → `get_table_binding`) — not performed; no live StyleBI/backend or `mcp__composer-chat__*` tools were reachable in the fixer's session. Flagging for reviewer/user to run if a live stack is available.

Full root-cause writeup, diff, and self-verification notes: `docs/teams/2026-09-11-bugs-vtb-p2/bug-76574-vtb004/03-fix.md` in the `stylebi-wiz` repo (this PR is in `stylebi`, the docs live in the sibling repo per this batch's workflow).

**Not merging** — holding for explicit review/sign-off per this batch's merge gate.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/TableBindingMutator.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/TableBindingMutator.java
@@ -136,12 +136,19 @@ public final class TableBindingMutator {
       }
 
       switch(name) {
-         case "rows" -> ((CrosstabBindingModel) model).setRows(
-            dimensions(refs, rvs, source, refModelService));
-         case "cols" -> ((CrosstabBindingModel) model).setCols(
-            dimensions(refs, rvs, source, refModelService));
-         case "groups" -> ((TableBindingModel) model).setGroups(
-            dimensions(refs, rvs, source, refModelService));
+         case "rows", "cols", "groups" -> {
+            // Captured before the overwrite below, so an unchanged field at the same position
+            // keeps its sort/ranking instead of dimensions() building it a fresh default one --
+            // see dimensions()'s own note on why this is matched by position, not just column.
+            List<BDimensionRefModel> previous = dimensionsOf(model, name);
+            List<BDimensionRefModel> next = dimensions(refs, previous, rvs, source, refModelService);
+
+            switch(name) {
+               case "rows" -> ((CrosstabBindingModel) model).setRows(next);
+               case "cols" -> ((CrosstabBindingModel) model).setCols(next);
+               default -> ((TableBindingModel) model).setGroups(next);
+            }
+         }
          case "details" -> ((TableBindingModel) model).setDetails(details(refs));
          default -> model.setAggregates(aggregates(refs));
       }
@@ -424,17 +431,34 @@ public final class TableBindingMutator {
     * overload too.
     */
    private static List<BDimensionRefModel> dimensions(List<FieldRef> fields) {
-      return dimensions(fields, null, null, null);
+      return dimensions(fields, List.of(), null, null, null);
    }
 
-   private static List<BDimensionRefModel> dimensions(List<FieldRef> fields, RuntimeViewsheet rvs,
+   /**
+    * @param previous the shelf's own {@code BDimensionRefModel} list before this write, so a
+    *                 field that is unchanged keeps its sort/ranking instead of resetting to
+    *                 {@code BDimensionRefModel}'s class defaults on every write (VTB-004).
+    *                 Matched by <b>same absolute index + same column (case-insensitive) + same
+    *                 date level</b> -- the same position-keyed identity {@link #requireDimension}
+    *                 and {@code pruneOrphanedSuppression} already use to disambiguate a column
+    *                 bound twice on one shelf, not a content-only key. This deliberately does
+    *                 NOT survive an insert/remove of a different field earlier on the shelf
+    *                 shifting a later, untouched field's index -- that field's sort/ranking is
+    *                 not preserved in that case, which is accepted as this strategy's known
+    *                 boundary rather than a defect (see the regression test that documents it).
+    */
+   private static List<BDimensionRefModel> dimensions(List<FieldRef> fields,
+                                                       List<BDimensionRefModel> previous,
+                                                       RuntimeViewsheet rvs,
                                                        SourceInfo source,
                                                        DataRefModelFactoryService refModelService)
    {
       List<BDimensionRefModel> out = new ArrayList<>();
 
-      for(FieldRef field : fields) {
-         BDimensionRefModel ref = new BDimensionRefModel();
+      for(int i = 0; i < fields.size(); i++) {
+         FieldRef field = fields.get(i);
+         BDimensionRefModel match = i < previous.size() ? previous.get(i) : null;
+         BDimensionRefModel ref = matches(match, field) ? copyOf(match) : new BDimensionRefModel();
          ref.setName(field.column());
          ref.setColumnValue(field.column());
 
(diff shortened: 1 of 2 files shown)


lizhepu (commented):
## Review: PR #5178 (VTB-004, Redmine #76574)

Reviewed against `docs/teams/2026-09-11-bugs-vtb-p2/bug-76574-vtb004/{01-diagnosis,02-refute,03-fix}.md` and the actual diff (`gh pr diff 5178`), not just the fix record's own description.

**Strategy check (a) vs (b):** Confirmed. `dimensions()` matches a new field to the shelf's previous list purely by `i < previous.size()` (same absolute index) + `matches()` (same column, case-insensitive, + same normalized date level) — this is strategy (a), not the diagnosis's original content-keyed (b). No index-agnostic/content-only matching anywhere in the diff.

**Capture-before-overwrite:** Confirmed. `setShelf` calls `dimensionsOf(model, name)` (reading the model's *current* rows/cols/groups) into `previous` before building `next` and assigning it back — correct order of operations, not a stale/post-overwrite read.

**Fields copied forward vs re-applied:** `copyOf()` carries `order`, `sortByCol`, `manualOrder`, `rankingOption`/`rankingN`/`rankingCol`, `groupOthers`, `others` (the "namedOthers" from the diagnosis). `columnValue`/`dateLevel`/`namedGroupInfo` are then re-applied from the incoming field on top, same as before the fix — **except for one gap, below**.

**Regression tests:** All 7 are well-targeted and independently exercise their own call path (`addField`/`removeField`/`moveField` each call the real production method, not just `setShelf` three times under different names). The duplicate-bound-column test genuinely proves no cross-contamination (distinct sort state asserted on each occurrence after resubmission). The strategy-(a)-boundary test documents the accepted limitation correctly (asserts the reset happens, framed as accepted). The stale-ranking-after-aggregate-removal test confirms the refuter's edge case is a pre-existing gap (unaffected by this fix, since it doesn't touch the `rows` shelf at all) — correctly scoped out, not silently ignored.

**`aggregates()`/`calculateInfo` sibling:** Confirmed untouched in this diff and only noted (PR body + fix record) as a follow-on candidate — correct scoping.

**Base/branch/diff hygiene:** Base is `stylebi-wiz-main-rebase` as required. Only `TableBindingMutator.java` and its test file are touched — nothing from the concurrent VTB-007 work leaked in.

### Important (confidence ~82): `copyOf()` can leave `order` stuck at `SORT_SPECIFIC` with no named-group backing

`copyOf()` (`TableBindingMutator.java:512-525`) unconditionally carries `order` forward from the matched previous ref (`ref.setOrder(previous.getOrder())`). `dimensions()`'s named-group block (lines 469-482) only ever *sets* `order = SORT_SPECIFIC` when the **incoming** field supplies `namedGroupValues`/`namedGroup` — there is no corresponding path that clears `order` back off `SORT_SPECIFIC` when the incoming field does not.

Concretely: a field previously bound to a named group (`order == SORT_SPECIFIC`, `namedGroupInfo` set) is resubmitted via `set_table_fields` with the same column/index but *without* `namedGroupValues`/`namedGroup` (e.g. dropping the group, or just a caller who omitted it) — `matches()` still returns true (column+index+dateLevel unchanged), so `ref = copyOf(match)` copies `order = SORT_SPECIFIC` forward, but the reapply block is skipped (no namedGroupValues/namedGroup on the incoming field), so `namedGroupInfo` stays null on the new ref. The model is now internally inconsistent: `SORT_SPECIFIC` with nothing backing it.

Before this PR this could not happen — every ref was built fresh, so a field without `namedGroupValues`/`namedGroup` always got `BDimensionRefModel`'s class-default `order` (i.e. dropping the group correctly cleared it). This is a regression introduced by `copyOf()`.

`BDimensionRefModel` already has a self-heal for exactly this shape (`getOrder() == SORT_SPECIFIC && ngInfoModel == null/type 0 && manualOrder empty` → strip the bit — see its `...ToDataRef`-equivalent conversion, lines ~448-452), but that guard runs at model→runtime-dataref conversion, not inside `dimensions()`/`copyOf()`. So the *model itself* — what `describeSorts`/`get_table_binding` reads directly, and what a later `setShelf` call captures as its own `previous` — carries the inconsistent state until/unless that conversion happens to run. None of the 7 new regression tests cover "a previously named-grouped field resubmitted without `namedGroupValues`."

Suggested minimal fix: after building `ref` (whether via `copyOf` or fresh) in `dimensions()`, when the incoming field does not supply `namedGroupValues`/`namedGroup` and `ref.getOrder() == SORT_SPECIFIC` with no `manualOrder`, strip the `SORT_SPECIFIC` bit — mirroring the guard `BDimensionRefModel` already applies at conversion time — or narrower still, have `copyOf()` not carry `order` forward when the previous ref's `SORT_SPECIFIC` came from a named group rather than a manual order, since `dimensions()` decides namedGroupInfo independently right after.

### Everything else
No other high-confidence issues found. The core mechanical fix (index+column+dateLevel matching, capture-before-overwrite, field copy-forward) is correct and matches the diagnosis/refutation record faithfully.

**Verdict: needs a small fix round** for the named-group/`order` inconsistency above before merge; the rest of the change is solid.

lizhepu:
Fixed: strip SORT_SPECIFIC in dimensions() when a named group is dropped, mirroring BDimensionRefModel's own self-heal. New regression test added; 75/29/28/19 tests pass.

lizhepu:
Round 2 verified: else-if fires only on stale SORT_SPECIFIC with no namedGroupInfo/manualOrder backing, mutually exclusive with the if-branch, no-op on fresh refs, manual sort preserved. Resolved.

Nelson-cao (approved):

PR #5179, opened 2026-09-11
fix(wiz): add offset pagination to worksheet preview endpoint (#76574 VTB-002)

lizhepu:
## Summary
- `WorksheetPreviewService.preview()` always started reading at `TableLens` row 1 with no offset parameter, so `preview_worksheet_data` (the wiz agent's worksheet-preview tool) could never retrieve rows past the first `limit` (<=200) of a table larger than that — making exhaustive aggregate-total verification on large tables impossible.
- Adds an `offset` parameter (0-based, default 0) to `WorksheetPreviewService.preview()` and `WorksheetAgentController.preview()`; the row loop now starts at `1 + offset`. Rows beyond the table's end are returned as an empty list, not an error (`moreRows(row)` naturally returns `false`).
- The existing `Math.min(limit, 200)` payload-size cap is unchanged (orthogonal to this fix).

Companion `stylebi-wiz` change (wiz-services route/service + plugin tool schema/description) is in `stylebi-wiz` PR https://github.com/inetsoft-technology/stylebi-wiz/pull/2449.

Redmine: #76574 (VTB-002)

## Test plan
- [x] `../mvnw.cmd -pl core test -Dtest=WorksheetPreviewServiceTest,WorksheetAgentControllerTest` — 15/15 and 180/180 pass, including new regression tests:
  - `WorksheetPreviewServiceTest.offsetSkipsLeadingRows` — offset=2, limit=2 on a 5-row fixture returns rows 3-4
  - `WorksheetPreviewServiceTest.offsetBeyondRowCountReturnsEmptyNotError` — offset past the end returns `[]`, no exception
  - `WorksheetAgentControllerTest.previewForwardsOffsetDefaultingToZero` / `previewForwardsNonZeroOffset` — offset forwarded to the service, defaults to 0

**Do not merge** — holding for the user's own review per this batch's merge gate.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
@@ -1191,10 +1191,12 @@ public class WorksheetAgentController {
    }
 
    /**
-    * Return up to {@code limit} data rows from the named table in the live worksheet.
+    * Return up to {@code limit} data rows from the named table in the live worksheet,
+    * starting after the first {@code offset} rows.
     *
     * @param sessionToken the token obtained at join time
     * @param table        the table assembly name to query
+    * @param offset       number of leading data rows to skip; 0-based, defaults to 0
     * @param limit        maximum rows to return (capped at 200; defaults to 50)
     * @param user         the authenticated agent principal
     * @return list of row maps, each keyed by column name
@@ -1204,6 +1206,7 @@ public class WorksheetAgentController {
    @GetMapping("/api/wiz/v1/agent/worksheet/{sessionToken}/preview")
    public List<Map<String, Object>> preview(@PathVariable String sessionToken,
                                              @RequestParam String table,
+                                             @RequestParam(defaultValue = "0") int offset,
                                              @RequestParam(defaultValue = "50") int limit,
                                              Principal user)
       throws PairingException
@@ -1211,7 +1214,7 @@ public class WorksheetAgentController {
       requireEnabled();
       requireWholeSheetSession(sessionToken, user);
       RuntimeWorksheet rws = editService.resolve(sessionToken, user);
-      return previewService.preview(rws, table, Math.min(limit, 200));
+      return previewService.preview(rws, table, offset, Math.min(limit, 200));
    }
 
    /**
--- a/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetPreviewService.java
+++ b/core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetPreviewService.java
@@ -51,7 +51,9 @@ public class WorksheetPreviewService {
     *
     * @param rws       the live runtime worksheet
     * @param tableName the table assembly name to query
-    * @param offset    number of leading data rows to skip (header row excluded); 0-based
+    * @param offset    number of leading data rows to skip (header row excluded); 0-based.
+    *                  Negative values are clamped to 0 (mirrors how {@code limit} is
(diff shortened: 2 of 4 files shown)


lizhepu (commented):
## Review — VTB-002 offset param (batch 2)

Verified against `01-diagnosis.md`/`02-refute.md`/`03-fix.md`, not just the PR description.

**Confirmed correct:**
- Row loop: `for(int row = 1 + offset; rows.size() < limit && lens.moreRows(row); row++)` is right — offset shifts the start row, `limit` still bounds count, `moreRows` still guards the end.
- Call-site update: grepped the diff directly — **13** pre-existing `service.preview(...)` call sites in `WorksheetPreviewServiceTest.java` were updated to the new `(rws, table, offset, limit)` signature (the progress log says "12"; actual count is 13, all present and correct — just a miscount in the log, not a code gap), plus the one production call site in `WorksheetAgentController.java`. No missed call site, no compile break.
- `offsetBeyondRowCountReturnsEmptyNotError`: read the actual test — 3-row fixture, `offset=100, limit=10`. The `lens()` test helper only stubs `moreRows(1..3)=true` and `moreRows(4)=false`; `moreRows(101)` is unstubbed and Mockito's default `boolean` return is `false`, which is exactly the real `TableLens` contract at an out-of-range index. So this passes for the right reason (empty list, no exception), not accidentally.
- `offsetSkipsLeadingRows` (offset=2, limit=2 on a 5-row fixture) correctly returns rows 3–4, traced through the loop manually.
- `WorksheetAgentControllerTest`'s new 7-arg `controller(...)` overload matches the real constructor's parameter order exactly (`WorksheetPreviewService` is the 7th ctor param) — no wiring mistake.
- `@RequestParam(defaultValue = "0") int offset` claim: confirmed this file has **zero** MockMvc usage anywhere, and `limit`'s own defaultValue was likewise never tested via absent-param binding before this PR — the new tests (`previewForwardsOffsetDefaultingToZero`/`previewForwardsNonZeroOffset`) call the method directly, matching the file's established (MockMvc-free) convention. The claim that "the annotation itself guarantees the default, not the test" is accurate and consistently stated, not overreaching.
- Single call site for `WorksheetPreviewService.preview()` re-confirmed.

**Important (≈82 confidence) — missing negative-offset guard:**
Neither `WorksheetAgentController.preview()` nor `WorksheetPreviewService.preview()` rejects/clamps a negative `offset`. With `offset = -1`, the loop starts at `row = 0`, which is the **header** row (`TableLens` row 0 per this class's own convention). `lens.moreRows(0)` will be true (the header row always exists) and `getObject(0, col)` returns header text, so a negative offset would silently return the header row masquerading as a data row — no exception, just wrong data. This is new exposure: before this PR there was no way to reach row 0 since the loop always started at 1. Given CLAUDE.md's "fail loud on bad input, don't silently degrade" principle for exactly this kind of caller-supplied parameter, suggest either clamping (`Math.max(0, offset)`) or throwing on `offset < 0` in `WorksheetPreviewService.preview()`, with a one-line regression test. Not blocking the existing positive-offset behavior, but worth a follow-up before this ships to an LLM caller that might compute a negative page index by mistake.

**Verdict:** Otherwise clean — row loop, call-site coverage, and both new tests are correct and verified by reading the actual code, not just the description. One Important gap (negative offset) worth a small follow-up fix.

lizhepu:
Fixed: negative offset now clamped to 0 in WorksheetPreviewService.preview() (matches limit's existing Math.min clamp). Added regression test, 07082c76c.

lizhepu:
Verified: startRow clamp correct for all offsets (-100..N); new test asserts real row "r1", not header. Resolved.

PR #5180, opened 2026-09-11
fix(report): stop right-aligned overflow text losing its leading digits (#76574 VTB-010)

lizhepu:
## Summary

Fixes Redmine #76574 (VTB-010): a crosstab's grand-total-of-grand-totals corner cell rendered `450115.549999997` instead of the correct `17450115.55` — the leading `17` silently disappeared.

**Root cause** (this required overturning the original diagnosis's proposed mechanism after a refutation pass — see `docs/teams/2026-09-11-bugs-vtb-p2/bug-76574-vtb010/03-fix.md` in the stylebi-wiz repo for the full trail):

- The corner cell's underlying sum is numerically correct (confirmed by arithmetic tracing through `CrossTabFilter`/`SumFormula`) — this is not a computation bug.
- The repro's crosstab (`VS-TABLE-BIND-3` Q1: `rows:[CATEGORY_NAME]`, `cols:[REGION]`, `aggregates:[REVENUE Sum]`, `rowTotals:true`, `colTotals:true`) never gets a user-set number format — confirmed both by the actual tool-call log (only `add_assembly`/`set_table_fields`/`set_table_options` were called, never a format tool) and by tracing the binding code paths: neither the wiz agent's `TableBindingService`/`TableBindingMutator` nor the native Composer's `VSCrosstabBindingHandler` ever assign a display format to a newly bound aggregate. This rules out the originally-suspected `VSFormatTableLens`/`CrossFilterDataDescriptor` format-map-miss (the corner cell's distinct 3-element `TableDataPath`) as the mechanism here — there's no format registered anywhere for it to miss, and `VSFormatTableLens.createCellFormat()` falls back to a table-wide `OBJECTPATH` format lookup that's keyed the same way regardless of a cell's own path shape.
- Because there's no format, every total cell's raw double is rendered as-is via `CoreTool.toString()`. The corner cell sums many more detail rows directly than either sibling grand total (which each sum only a subset), so it accumulates more floating-point noise and its unformatted string is longer — long enough to overflow this crosstab's column width where the shorter siblings do not.
- `Common.processText()` (`core/src/main/java/inetsoft/report/internal/Common.java`) computed a negative line offset for a right-aligned string wider than its cell (`nbound.width - w - 1 < 0`), which `paintText()` then used as the string's starting x — to the left of the `Graphics2D` clip rect it sets to the cell's bounds. The clip silently discarded whatever fell to its left, i.e. the string's leading (most-significant) characters, while the trailing (least-significant/decimal) characters stayed inside the clip and were shown — turning a truncated string into a shorter, still-plausible-looking, but numerically wrong value.

**Fix**: clamp that line offset to `0` so an overflowing right-aligned string is clipped on the right (its least-significant end) instead of the left (its most-significant end). One line, `Common.java:1401` in the diff.

**Scope note**: this is the minimal, targeted fix for VTB-010 itself. A broader UX affordance (ellipsis/shrink-to-fit for overflowing data cells, mirroring the existing design-time header truncation in `TablePaintable`) was considered but deliberately left out of scope — see the fix record for the full discussion.

## Test plan
- [x] Added `core/src/test/java/inetsoft/report/internal/CommonRightAlignOverflowTest.java` — reproduces the exact right-aligned overflow scenario with the bug's own numbers (`17450115.549999997`) and asserts the line offset is never negative; a second test confirms the normal non-overflowing right-align case is unaffected.
- [x] Verified the new test fails without the fix (reverted via git stash, re-ran, confirmed failure) and passes with it restored.
- [x] Ran the new test plus the closest existing crosstab/format/export tests for regressions — all pass:
  `./mvnw -pl core -am -o test -Dtest=CommonRightAlignOverflowTest,AbstractVSExporterLabelTest,VSTableHelperTest,VSFormatTableLensTest,CrossTabFilterTest,CrosstabVSAQueryTest` → 20/20 passed.

### DIFF
--- a/core/src/main/java/inetsoft/report/internal/Common.java
+++ b/core/src/main/java/inetsoft/report/internal/Common.java
@@ -1398,7 +1398,15 @@ public class Common extends Util {
          else if((align & H_RIGHT) != 0 && nbound.width != w) {
             // @by mikec, fix the line off of text.
             // @see bug1135784206363
-            lineoff.setElementAt(nbound.width - w - 1, i);
+            // bug #76574 (VTB-010): when the text is wider than the cell (w >
+            // nbound.width), this would go negative, positioning the string's start
+            // to the left of the clip rect paintText() sets to nbound -- silently
+            // discarding the leading, most-significant characters (e.g. a right-
+            // aligned number's high-order digits) while keeping the least-significant
+            // tail visible, which reads as a shorter but plausible, wrong value.
+            // Clamp to 0 so an overflowing string is clipped on the right (its
+            // least-significant end) instead of the left.
+            lineoff.setElementAt(Math.max(0f, nbound.width - w - 1), i);
          }
          else {
             lineoff.setElementAt((float) 0, i);
--- a/core/src/test/java/inetsoft/report/internal/CommonRightAlignOverflowTest.java
+++ b/core/src/test/java/inetsoft/report/internal/CommonRightAlignOverflowTest.java
@@ -1,0 +1,127 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2025  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.report.internal;
+
+import inetsoft.report.StyleConstants;
+import inetsoft.test.*;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import java.awt.*;
+import java.util.Vector;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Regression test for bug #76574 (VTB-010): a crosstab's grand-total-of-grand-totals corner
+ * cell rendered as "450115.549999997" instead of "17450115.549999997" -- the leading "17"
+ * silently disappeared.
+ *
+ * Root cause: the corner cell's underlying double is numerically correct (confirmed by
+ * arithmetic tracing in CrossTabFilter/SumFormula), and this specific crosstab never gets a
+ * user-set number format (the VS-TABLE-BIND-3 repro only calls set_table_fields/
+ * set_table_options, never a format tool, and neither the wiz binding path
+ * (TableBindingMutator) nor the native Composer binding handlers (VSCrosstabBindingHandler)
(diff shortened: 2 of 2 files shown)


lizhepu (commented):
## Review: VTB-010 (bug #76574) — right-align overflow clip fix

Reviewed the diff (`gh pr diff 5180`) against the full diagnosis -> refute -> fix trail in
`docs/teams/2026-09-11-bugs-vtb-p2/bug-76574-vtb010/`. Summary: the change is exactly what the
fix record claims — a one-line clamp plus a new regression test, no scope creep. No blocking
issues found. A few non-blocking notes below for the record.

### Scope check — clean
`gh pr view --json files` shows only two files touched: `Common.java` (+9/-1) and the new
`CommonRightAlignOverflowTest.java` (+127). The clamp (`Math.max(0f, nbound.width - w - 1)`)
sits only inside the `(align & H_RIGHT) != 0 && nbound.width != w` branch — verified against the
live source at `core/src/main/java/inetsoft/report/internal/Common.java:1398-1402`. `H_LEFT`
(offset always 0) and `H_CENTER` branches are untouched, matching the stated scope of "fix the
reported right-align overflow, don't redesign the overflow-render path."

### Root-cause reasoning — matches the code
Independently confirmed: `Common.alignCell()` leaves `nbound` unchanged when text overflows, so
the pre-fix `nbound.width - w - 1` does go negative for a right-aligned string wider than its
cell, and `paintText()`'s clip rect (set to `nbound`) then silently drops whatever is drawn left
of it — the leading/most-significant digits. The clamp starts the draw at `nbound.x` instead, so
overflow now clips the trailing (least-significant) end. This is the correct, minimal fix for the
symptom as filed.

### One related gap, not addressed by this PR (informational, not blocking)
`H_CENTER`'s offset, `(nbound.width - w) / 2`, also goes negative when `w > nbound.width` —
producing *symmetric* clipping (both leading and trailing characters lost) for overflowing
centered text. This wasn't in scope for VTB-010 (numeric crosstab cells are right-aligned per
`CrossTabFilter.getAlignment()`, never centered), and the minimal-fix principle supports leaving
it out, but it's a same-shaped latent issue in the same method the fix doc's own "follow-on
candidate" section doesn't mention (it only flags the general ellipsis/shrink-to-fit affordance).
Worth a one-line follow-up ticket note if centered numeric/text cells can ever overflow in
practice.

### Test — reproduces the actual bug shape
`CommonRightAlignOverflowTest` calls the real `Common.processText()`/`stringWidth()`/
`getFractionalFontMetrics()` with the bug's own string (`"17450115.549999997"`) at a column width
sized to fit the shorter sibling total but not the full corner string — not a simplified
stand-in. `rightAlignedOverflowingTextIsNotOffsetPastLeftEdge` asserts `offset >= 0`;
`rightAlignedTextThatFitsIsUnaffected` pins the unchanged non-overflow formula
(`bound.width - w - 1`) so the fix can't regress the common case. Annotations
(`@ExtendWith(SpringExtension.class)`, `@ContextConfiguration`, `@DirtiesContext`, `@SreeHome`,
`@Tag("core")`) match the existing sibling test `AbstractVSExporterLabelTest` exactly — correct
house style. No existing test in `core/src/test` calls `processText`/`alignCell` directly, so
there's no risk of an existing test asserting the old negative-offset behavior.

### Unverifiable-from-PR-alone claim (same pattern as VTB-008)
The fail-before/pass-after check (temporary `git stash` of just `Common.java`) isn't traceable in
the PR itself — one commit, no revert/re-apply history. Same caveat the VTB-008 review raised for
an equivalent claim; not a reason to distrust the fix (the test's assertions are self-evidently
tied to the clamp), just noting it's not independently checkable from the PR history alone.

### CI status at review time
`claude-review` check shows `FAILURE`, but the job log is just "Claude encountered an error after
0s" — an automation hiccup, not a content-based finding. `build` (Validation Build) was still
`IN_PROGRESS` at review time. Recommend waiting for a green build before merge; per this batch's
process, merge itself needs the user's explicit go-ahead regardless.

### Verdict
Clean. Minimal, correctly-scoped fix; test reproduces the real bug and pins the non-regression
case. No fix-required issues at ≥80 confidence.

Nelson-cao (approved):

PR #5181, opened 2026-09-11
fix(wiz): don't clobber paired-browser socket session on agent input writes

lizhepu:
## Summary

- Redmine #76574 (VTB-009): after any agent-driven `set_input_value` (checkbox/combo box/radio
  button/text input/spinner), the paired live browser Composer session never received a refresh
  broadcast — the user had to manually click StyleBI's own in-app "Refresh" menu command to see
  the change.
- Root cause: `VSInputService.applySelection` unconditionally overwrote the runtime viewsheet's
  socket session id/user name with whatever `CommandDispatcher` was in play. The wiz agent's
  `InputValueService.setValue` always runs under a `CapturingCommandDispatcher`, whose
  `getSessionId()` is deterministically `null` (it wraps a synthetic, non-STOMP message with no
  `simpSessionId` header). That `null` clobbered the already-correct paired-browser session id, so
  `SheetAgentBroadcastService.broadcastRefresh`'s `sessionId == null` guard silently skipped the
  broadcast on every single agent-driven input write, deterministically, not intermittently.
- Fix: only adopt the dispatcher's session id/user name when the dispatcher actually has one.
  Native STOMP controllers (`VSSliderController`, `VSSpinnerController`, `VSCheckBoxController`,
  `OnClickService`, `VSRadioButtonController`, `VSTextInputController`) always carry a real
  session id, so this is a no-op for the human-click path — verified by reading each controller's
  `@MessageMapping` dispatch.
- Scoped to `VSInputService.applySelection` only, per the diagnosis. The same
  unconditional-overwrite pattern also exists at `CoreLifecycleService`, `ComposerViewsheetService`,
  and `VSLifecycleControllerService` (viewsheet-open-style flows, not this per-edit refresh path) —
  intentionally left untouched; worth a separate audit if an "open viewsheet" agent tool is ever
  built that reaches them via a `CapturingCommandDispatcher`.

## Test plan

- [x] New regression test `core/src/test/java/inetsoft/web/viewsheet/service/VSInputServiceSocketSessionTest.java`:
  - Confirms a sessionless dispatcher (`CapturingCommandDispatcher`'s shape) does NOT overwrite an
    already-recorded socket session id/user name.
  - Confirms a real dispatcher (native STOMP controller's shape) still updates it, unchanged from
    before the fix.
  - Verified the test fails against the pre-fix code (temporarily reverted the fix, reran, saw the
    expected `NeverWantedButInvoked` failure) and passes with the fix restored.
- [x] `./mvnw test -pl core -Dtest=VSInputServiceSocketSessionTest,ViewsheetSessionServiceTest,VSInputTableBindingResolutionTest,VSComboBoxDateValueTest` — all pass, no regressions in the surrounding input-service/session tests.
- [ ] Live self-verification via `mcp__composer-chat__*` tools was not possible in this session — those MCP tools are not available here. Unit tests are the verified bar for this PR; a live re-check (confirming the "Pairing broadcast skipped" WARN no longer fires for `set_input_value`) is recommended before merge if a paired-browser session is available.

Not chased in this PR (logged as open follow-on candidates from the live-check record, not this
mechanism):
- A field-write-level intermittency question from an earlier, separate investigation (unreproduced
  in 8/8 trials, unrelated to this socket-session mechanism).
- A "lagged by one step" broadcast-healing observation (a later call's `applySocketSession`
  "only if not set" guard can heal the socket id, and something then surfaces a stale update) — not
  traced to a specific mechanism.

### DIFF
--- a/core/src/main/java/inetsoft/web/viewsheet/service/VSInputService.java
+++ b/core/src/main/java/inetsoft/web/viewsheet/service/VSInputService.java
@@ -2808,8 +2808,15 @@ public class VSInputService {
 
       int hint;
 
-      rvs.setSocketSessionId(dispatcher.getSessionId());
-      rvs.setSocketUserName(dispatcher.getUserName());
+      // Only adopt the dispatcher's socket session when it actually has one (a real STOMP
+      // dispatcher from a native browser controller). A CapturingCommandDispatcher (used by the
+      // wiz agent's InputValueService) has no real session and would otherwise clobber the
+      // already-correct paired-browser session id, silently disabling
+      // SheetAgentBroadcastService's refresh broadcast for every agent-driven input change.
+      if(dispatcher.getSessionId() != null) {
+         rvs.setSocketSessionId(dispatcher.getSessionId());
+         rvs.setSocketUserName(dispatcher.getUserName());
+      }
 
       box.get().lockWrite();
 
--- a/core/src/test/java/inetsoft/web/viewsheet/service/VSInputServiceSocketSessionTest.java
+++ b/core/src/test/java/inetsoft/web/viewsheet/service/VSInputServiceSocketSessionTest.java
@@ -1,0 +1,123 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2026  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.web.viewsheet.service;
+
+/*
+ * Test strategy (VTB-009 / Redmine #76574)
+ *
+ * VSInputService.applySelection unconditionally overwrote rvs's paired-browser socket session
+ * identity with whatever CommandDispatcher happened to be in play. The wiz agent's
+ * InputValueService.setValue (the sole agent-side caller of singleApplySelection/
+ * multiApplySelection/applySelection, serving CheckBox/ComboBox/RadioButton/TextInput/Spinner)
+ * always runs under a CapturingCommandDispatcher, whose getSessionId() deterministically returns
+ * null (it wraps a synthetic, non-STOMP message with no simpSessionId header). That null then
+ * clobbered rvs's already-correct socket session id, so
+ * SheetAgentBroadcastService.broadcastRefresh's "sessionId == null" guard silently skipped the
+ * paired browser's refresh broadcast on every single agent-driven input write.
+ *
+ * The fix only adopts the dispatcher's session id/user name when the dispatcher actually has one,
+ * so a CapturingCommandDispatcher (or any dispatcher with no real session) leaves an
+ * already-recorded socket session alone, while a real STOMP dispatcher (every native browser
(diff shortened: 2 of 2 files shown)


lizhepu (commented):
## Review — VTB-009 broadcast fix (Redmine #76574)

Verified against `E:\inetsoft\wizard` source (not just the diagnosis docs) and the actual `gh pr diff`.

### Diff scope — confirmed minimal
Exactly one production change: `VSInputService.applySelection` (lines 2809-2818) guards the socket-session/user-name overwrite behind `if(dispatcher.getSessionId() != null)`. No other logic touched. Confirmed by reading the live method body (`vs.getViewsheetSandbox()` check at 2803-2807, the guarded write at 2811-2812, `lockWrite()`/`applySelection0` unchanged below) — matches the diff exactly, no drift from the diagnosis's line numbers.

Confirmed the diff does **not** touch `CoreLifecycleService.java`, `ComposerViewsheetService.java`, or `VSLifecycleControllerService.java` — `gh pr view --json files` lists only `VSInputService.java` and the new test file. Those three files' own unconditional-overwrite sites (viewsheet-open flows) are correctly left alone, as the PR body states.

Base branch confirmed: `stylebi-wiz-main-rebase` (`gh pr view --json baseRefName`).

### Guard equivalence question (asked in the brief) — resolved, no divergence that matters here
`ViewsheetSessionService.applySocketSession`'s guard is "write only if `rvs`'s existing value is null"; this new guard is "write only if the incoming dispatcher value is non-null" — genuinely different conditions. Traced the one scenario where they could diverge: `rvs`'s socket session already null AND the dispatcher also has a null session id. Under the new guard the write is skipped, leaving `rvs` null — but under the *old* (pre-fix) code that same call would have written `null` over `null`, i.e. identical end state. So this edge case is not a new regression; it was already a "nothing to broadcast to" case before this fix (no real browser ever paired), not something the fix makes worse. For every other case (dispatcher has a real, non-null session id) the new guard writes exactly as the old unconditional code did. Net: functionally equivalent in effect at this call site, confirmed by tracing rather than assumed from the similar naming.

### Native STOMP controllers — confirmed true no-op
Grepped for `CommandDispatcher` subclasses: only one exists, `CapturingCommandDispatcher` (`wiz/dispatch/CapturingCommandDispatcher.java`). All six native controllers (spot-checked `VSCheckBoxController` at `web/viewsheet/controller/dialog/VSCheckBoxController.java:58-60`, `@MessageMapping("/checkBox/applySelection")` with a plain `CommandDispatcher` parameter) receive the base class, populated by Spring's STOMP argument resolution from a real frame's `simpSessionId` header — never null in that path. So the guard is provably a no-op for all six, not just "expected to be."

### Test file — reaches the guarded lines correctly, not a trivial pass
Read `VSInputServiceSocketSessionTest.java` and the real `applySelection`/`applySelection0` bodies side by side. The reflection call (`applySelection(String, String, Object, Principal, CommandDispatcher)`, 5-arg private method) matches the real signature. The mock setup (`getViewsheetSandbox()` → `Optional.of(box)`, so it doesn't hit the early `box.isEmpty()` return at line 2805-2807) correctly reaches the guarded clobber lines, then falls through `lockWrite()`/`applySelection0` (`vs.getAssembly("Assembly1")` → `null` → not an `InputVSAssembly` → early `return 0`) without touching any of the other 8 null-passed constructor dependencies. Constructor arg count (9: `vsObjectService` + 8 nulls) matches the real 9-arg constructor. Both cases (G1 sessionless-doesn't-clobber, G2 real-session-still-updates) assert on the actual guarded behavior, not an unrelated side effect. This is a real, non-trivial regression test for this exact bug.

### Not independently verifiable
Per the same caveat noted on VTB-008/VTB-010: the stash-based fail-before/pass-after check described in the fix record can't be independently confirmed from a single-commit PR history. The test's own logic is self-evidently tied to the guard, though, so this is a low-concern gap.

No live re-check was attempted — the currently-running wizard server was not rebuilt with this fix, so a live check right now would only reflect pre-fix behavior and wouldn't be meaningful.

### Verdict
Clean. Correctly scoped, traced end-to-end against the real source (not just the diagnosis's paraphrase), no unintended behavior change for the native browser path, and the new test genuinely exercises the fixed lines. No blocking issues found. Per this batch's merge gate, holding for the user's own explicit approval before merge.

Nelson-cao (approved):

PR #5184, opened 2026-09-11
Fix bug 76609 AID-003: group-owned schedule tasks ignored on delete/rename

Zxm-happy:
## Root cause

`ScheduleManager.identityRemoved`, `getIdentityRemovalImpact`, and `identityRenamed` each
hardcoded a `type == Identity.USER` guard before matching against `ScheduleTask.getOwner()`.
`getOwner()` returns a bare `IdentityID` with no embedded type discriminator, and groups can own
schedule tasks the same way users do, so the hardcoded `Identity.USER` check silently excluded
every group-owned task from all three code paths:

- `getIdentityRemovalImpact` never reported a group-owned task in `ownedTasks()`, so
  `apply_identity_changes` (`verb: "delete"`, `unitType: "group"`) always returned `advisory: null`
  even when the group owned tasks.
- `identityRemoved` never deleted a group-owned task on a real group delete, leaving a dangling
  `owner` reference to the now-deleted group.
- `identityRenamed` never updated a group-owned task's `owner` on a group rename, leaving a
  dangling reference to the old (renamed-away) group name.

The sibling "execute as" branches in all three methods already compare against
`iden.getType()` (the actual type stored on the task's own execute-as identity) and were already
symmetric for user/group — only the `owner`-matching branches had the hardcoded `USER` guard.

## Fix

Widen all three guards from `type == Identity.USER` to
`(type == Identity.USER || type == Identity.GROUP)`, in
`core/src/main/java/inetsoft/sree/schedule/ScheduleManager.java`:
- line 1091 (`identityRemoved`)
- line 1165 (`getIdentityRemovalImpact`)
- line 1205 (`identityRenamed`)

No other lines touched. `task.getOwner()` has no type of its own, so widening the check is
sufficient and safe — role/organization delete never reach these methods through this call path
(confirmed by tracing `IdentityChangesetApplyService.applyDeleteRole`/
`applyDeleteOrganizationSafely`, both of which hardcode `advisory: null` without calling
`describeDeleteTaskImpacts` at all).

## Tests

Three new tests added to `core/src/test/java/inetsoft/sree/schedule/ScheduleManagerTest.java`,
mirroring the existing user-owned cases:
- `getIdentityRemovalImpact_reportsGroupOwnedTasksWithoutMutating`
- `identityRemoved_removesGroupOwnedTask`
- `identityRenamed_updatesGroupOwnedTaskOwner`

```
cd core && mvn test -Dtest=ScheduleManagerTest -q
```
`Tests run: 17, Failures: 0, Errors: 0, Skipped: 0` (14 pre-existing + 3 new, all passing).

Fixes Redmine #76609, AID-003 finding (AID-001/AID-002 on the same ticket already landed
separately).

### DIFF
--- a/core/src/main/java/inetsoft/sree/schedule/ScheduleManager.java
+++ b/core/src/main/java/inetsoft/sree/schedule/ScheduleManager.java
@@ -1088,7 +1088,7 @@ public class ScheduleManager {
             continue;
          }
 
-         if(type == Identity.USER && identityID.equals(task.getOwner())) {
+         if((type == Identity.USER || type == Identity.GROUP) && identityID.equals(task.getOwner())) {
             i.remove();
             continue;
          }
@@ -1162,7 +1162,7 @@ public class ScheduleManager {
             continue;
          }
 
-         if(type == Identity.USER && identityID.equals(task.getOwner())) {
+         if((type == Identity.USER || type == Identity.GROUP) && identityID.equals(task.getOwner())) {
             ownedTasks.add(task.getName());
             continue;
          }
@@ -1202,7 +1202,7 @@ public class ScheduleManager {
             continue;
          }
 
-         if(type == Identity.USER && oname.equals(task.getOwner())) {
+         if((type == Identity.USER || type == Identity.GROUP) && oname.equals(task.getOwner())) {
             String oldTaskId = task.getTaskId();
             this.getOrgTaskMap(orgID).remove(getTaskIdentifier(oldTaskId, orgID));
             task.setOwner(id);
--- a/core/src/test/java/inetsoft/sree/schedule/ScheduleManagerTest.java
+++ b/core/src/test/java/inetsoft/sree/schedule/ScheduleManagerTest.java
@@ -70,6 +70,7 @@ public class ScheduleManagerTest {
 
    private IdentityID identityID_admin;
    private IdentityID identityID_tuser0;
+   private IdentityID identityID_tgroup0;
    private SRPrincipal admin;
    private SRPrincipal tuser0;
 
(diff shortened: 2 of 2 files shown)


Zxm-happy:
## Review r1 verdict: APPROVE (no blocking issues)

Independently re-verified all three guard changes (`identityRemoved`, `getIdentityRemovalImpact`,
`identityRenamed` in `ScheduleManager.java`) against the live diff and worktree source, confirmed
zero overlap with #76598's already-merged fix (`git show --stat b5167158d` touches only
`ScheduleChangePlanService.java`), re-ran the sibling `grep -n "Identity.USER"` scan myself (no
fourth occurrence found), confirmed the "execute as" branches were genuinely untouched, and ran
`ScheduleManagerTest` myself from the worktree: `Tests run: 17, Failures: 0, Errors: 0, Skipped: 0`,
matching the fix doc's reported number.

The three new regression tests (`getIdentityRemovalImpact_reportsGroupOwnedTasksWithoutMutating`,
`identityRemoved_removesGroupOwnedTask`, `identityRenamed_updatesGroupOwnedTaskOwner`) mirror
existing trusted user-owned sibling tests and assert the real invariant (group treated
symmetrically with user for schedule-task ownership), not just the literal reported symptom.

Full write-up: `docs/teams/2026-09-11-bug-76609-aid003-group-delete-advisory/bug-76609-aid003/04-review-r1.md`
in the wiz repo.

CI note: `claude-review` shows `fail`, but this is a pre-existing repo-wide infra issue (empty
`ANTHROPIC_API_KEY`, immediate `is_error:true`) unrelated to this diff -- the same failure appears
on the three most recent merged PRs in this repo (#5179, #5181, #5182). `build` was still
`pending` at review time; merge should wait for it to go green.

Note on review mechanics: the `gh` account in this session (`Zxm-happy`) is the same account that
authored this PR, so a formal `--approve` review would be a self-approval. `reviewDecision` on
this PR is currently empty, meaning branch protection here does not require an approving review to
merge, so this is not a blocker -- flagging it for visibility per this workflow's established
pattern in other repos in this archive.

PR #5185, opened 2026-09-11
fix(wiz): clear group-others flag when a ranking is cleared (#76574 VTB-006)

lizhepu:
## Summary
- Redmine #76574 (VTB-006, batch 3/P3): `DimensionSortRanking.applyRanking()`'s `mode:"none"` branch reset `rankingOption`/`rankingN`/`rankingCol` but returned before reaching the `groupOthers` clear, which only lives in the top/bottom-ranking branch below the early return. `get_table_binding`'s `"others"` field therefore stayed stuck at whatever it was before the clear, forever, even though `"ranking"` correctly read `"none"`.
- This is presently reachable, not hypothetical: VTB-004's already-merged `TableBindingMutator.copyOf()` fix (`ref.setGroupOthers(previous.isGroupOthers())`) copies forward whatever `groupOthers` currently holds, so today a shelf resubmission after a ranking `mode:"none"` clear silently perpetuates a stale `others:true` via that already-merged code path.
- Fix: clear `groupOthers` unconditionally in the `mode:"none"` branch. Surveyed all ~35 call sites of `isGroupOthers()`/`setGroupOthers()`; none treat the flag as meaningful independent of an active ranking, so this is safe. `ChartBindingMutator.setRanking()` shares the exact same method/model, so this one fix covers both table and chart ranking clears.

## Changes
- `core/src/main/java/inetsoft/web/wiz/binding/DimensionSortRanking.java`: add `dimension.setGroupOthers(false);` to the `mode:"none"` branch of `applyRanking()`.
- `core/src/test/java/inetsoft/web/wiz/binding/DimensionSortRankingTest.java`: extend the existing `clearingRankingNeedsNothingElse()` test to set `groupOthers=true` before the clear and assert it reads `false` after.
- `core/src/test/java/inetsoft/web/wiz/binding/TableBindingMutatorTest.java`: add `clearingRankingClearsGroupOthersAcrossAShelfResubmission()`, a compound regression test covering the presently-live VTB-004 interaction (rank with `others:true` -> clear via `mode:"none"` -> resubmit the shelf unchanged -> assert `groupOthers` still reads `false`).

## Test plan
- [x] Confirmed both new/extended assertions fail against the pre-fix code (`expected: <false> but was: <true>`), then pass after the one-line fix, via:
  `./mvnw -pl core -am -Dtest=DimensionSortRankingTest,TableBindingMutatorTest -Dsurefire.failIfNoSpecifiedTests=false test`
- [x] `ChartBindingMutatorTest` (28 tests) re-run to confirm no regression on the shared chart-ranking path.

Not merging per this batch's process — holding for the user's explicit review/merge sign-off regardless of CI/review outcome.

### DIFF
--- a/core/src/main/java/inetsoft/web/wiz/binding/DimensionSortRanking.java
+++ b/core/src/main/java/inetsoft/web/wiz/binding/DimensionSortRanking.java
@@ -120,6 +120,7 @@ public final class DimensionSortRanking {
          dimension.setRankingOption(String.valueOf(XCondition.NONE));
          dimension.setRankingN(null);
          dimension.setRankingCol(null);
+         dimension.setGroupOthers(false);
          return;
       }
 
--- a/core/src/test/java/inetsoft/web/wiz/binding/DimensionSortRankingTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/binding/DimensionSortRankingTest.java
@@ -173,12 +173,15 @@ class DimensionSortRankingTest {
    @Test
    void clearingRankingNeedsNothingElse() {
       BDimensionRefModel dimension = new BDimensionRefModel();
-      DimensionSortRanking.applyRanking(dimension, ranking("top", 5, "Sales"));
+      DimensionSortRanking.applyRanking(dimension,
+         new DimensionSortRanking.Ranking("top", 5, "Sales", true));
 
       DimensionSortRanking.applyRanking(dimension, ranking("none", null, null));
 
       assertNull(dimension.getRankingN());
       assertNull(dimension.getRankingCol());
+      assertFalse(dimension.isGroupOthers(),
+         "clearing the ranking must also clear the now-meaningless 'group others' flag");
    }
 
    @Test
--- a/core/src/test/java/inetsoft/web/wiz/binding/TableBindingMutatorTest.java
+++ b/core/src/test/java/inetsoft/web/wiz/binding/TableBindingMutatorTest.java
@@ -745,6 +745,30 @@ class TableBindingMutatorTest {
       assertEquals("Sales", model.getRows().get(0).getRankingCol());
    }
 
+   /**
+    * VTB-006: clearing a ranking must also clear the "group others" flag it carried, and that
+    * cleared flag must stick across a subsequent shelf resubmission (copyOf()'s job is only to
+    * preserve whatever the model currently holds, per VTB-004).
+    */
+   @Test
+   void clearingRankingClearsGroupOthersAcrossAShelfResubmission() {
+      CrosstabBindingModel model = new CrosstabBindingModel();
+      TableBindingMutator.setShelf(model, "rows", List.of(dim("Region")));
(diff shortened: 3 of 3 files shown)


lizhepu (commented):
Reviewed the actual diff (`gh pr diff 5185`) against the diagnosis/refute/fix docs, not just the docs themselves. Findings:

**Diff is exactly what's claimed.** One line, `dimension.setGroupOthers(false);`, added in `DimensionSortRanking.applyRanking()`'s `mode:"none"` branch (`core/src/main/java/inetsoft/web/wiz/binding/DimensionSortRanking.java:123`), unconditional — not gated on `ranking.others()`. Confirmed against the current file: the early `return` at line 123 is the only exit from that branch, and `groupOthers` was previously untouched there.

**Compound VTB-004 interaction is real and live, not hypothetical.** `TableBindingMutator.copyOf()` (`core/src/main/java/inetsoft/web/wiz/binding/TableBindingMutator.java`) does `ref.setGroupOthers(previous.isGroupOthers())` unconditionally — confirmed on `origin/stylebi-wiz-main-rebase`. VTB-004 (PR #5178, commit `446239c08`) is already merged into that base branch, so today, without this fix, a shelf resubmission after a `mode:"none"` ranking clear does perpetuate a stale `others:true` via `copyOf()`. This fix resolves it correctly by ensuring the model never holds a stale `true` in the first place.

**Tests genuinely reach the guarded code and assert the right field.**
- `DimensionSortRankingTest.clearingRankingNeedsNothingElse()` now seeds `groupOthers=true` via the 4-arg `Ranking(mode, n, measure, others)` constructor (confirmed this record shape exists) before clearing, and asserts `isGroupOthers()` is `false` after — a real assertion on the actual field, not a proxy.
- `TableBindingMutatorTest.clearingRankingClearsGroupOthersAcrossAShelfResubmission()` exercises rank → clear → resubmit-shelf, correctly targeting the `copyOf()` interaction. Helper usage (`dim()`, `FieldRef(...)`, `CrosstabBindingModel`) matches existing sibling tests in the same file.
- Both fail-before/pass-after results in the fix record match what the code supports (no other path clears `groupOthers` on `mode:"none"`).

**Chart-ranking shared path correctly left alone.** `ChartBindingMutator.setRanking()` calls the same `DimensionSortRanking.applyRanking()` on the same `BDimensionRefModel` type — no separate handling needed. Grepped `ChartBindingMutatorTest` on the base branch for any `groupOthers`/`others` expectations: none found, so nothing in that suite assumes the old stale-true behavior.

**`VSDataSet.isOthers()` side-effect resolution holds.** It's the one consumer not gated on an active ranking; the fix only makes it more correct (no longer mis-flags a literal `"Others"` data value as invalid post-clear), it doesn't need separate handling.

**Scope check.** Diff is exactly 3 files (1 prod line, 2 test-file additions), base branch is `stylebi-wiz-main-rebase` as expected, no unrelated changes.

No blockers or important issues found. Not approving per this batch's process — holding for the user's own explicit merge sign-off.

PR #5186, opened 2026-09-11
Bug #76615, don't NPE saving a viewsheet with a null runtime sheet ID

jkidacki-inetsoft:
## Summary

Saving a viewsheet via Composer could occasionally throw an unhandled `NullPointerException` on the server instead of a graceful error, when the STOMP save request reached the server before a runtime sheet ID was attached to the session.

## Root Cause

`ComposerViewsheetController.saveViewsheet()` passed `runtimeViewsheetRef.getRuntimeId()` unchecked into `ComposerViewsheetServiceProxy.saveViewsheet()`. When that ID is `null`, the cluster proxy call chain (`WorksheetEngine.isLocal()` → `RuntimeSheetCache.getAffinityKey()`) builds `new AffinityKey<>(null, ...)`, which Ignite's `GridArgumentCheck.notNull()` rejects, surfacing as an unhandled NPE on a clustered deployment. `closeViewsheet()` in the same controller already guards against this exact scenario, but `saveViewsheet()` did not.

## Fix

Added the same null-runtime-ID guard used by `closeViewsheet()` to `saveViewsheet()`: log a warning and return `false` instead of proceeding with a null ID. `saveAndCloseViewsheet()` (which calls `saveViewsheet()` internally) correctly skips sending `CloseSheetCommand` when the save returns `false`.

## Test Plan

- Added `ComposerViewsheetControllerTest.saveViewsheetReturnsFalseWhenRuntimeIdIsNull()`, which verifies the null-ID path returns `false` and never reaches the downstream proxy/service.
- `./mvnw test -pl core -Dtest=ComposerViewsheetControllerTest` — new test passes.

Fixes Redmine #76615.

### DIFF
--- a/core/src/main/java/inetsoft/web/composer/vs/controller/ComposerViewsheetController.java
+++ b/core/src/main/java/inetsoft/web/composer/vs/controller/ComposerViewsheetController.java
@@ -99,6 +99,11 @@ public class ComposerViewsheetController {
                                 CommandDispatcher dispatcher, @LinkUri String linkUri)
       throws Exception
    {
+      if(runtimeViewsheetRef.getRuntimeId() == null) {
+         LOG.warn("Attempted to save viewsheet without runtime ID");
+         return false;
+      }
+
       if(!securityEngine.checkPermission(principal, ResourceType.VIEWSHEET,
                                                        "*", ResourceAction.ACCESS))
       {
--- a/core/src/test/java/inetsoft/web/composer/vs/controller/ComposerViewsheetControllerTest.java
+++ b/core/src/test/java/inetsoft/web/composer/vs/controller/ComposerViewsheetControllerTest.java
@@ -1,0 +1,65 @@
+/*
+ * This file is part of StyleBI.
+ * Copyright (C) 2024  InetSoft Technology
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package inetsoft.web.composer.vs.controller;
+
+import inetsoft.analytic.composition.ViewsheetService;
+import inetsoft.sree.security.SecurityEngine;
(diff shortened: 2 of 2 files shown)


jshobe-inetsoft (approved):
## Manual Review (replacing failed automated `claude-review`)

The automated `claude-review` check failed due to a CI infrastructure/quota error (`is_error: true`, 1 turn, $0 cost — the bot's Claude invocation never actually ran), not because a review found a blocking issue. This is a manual review performed in its place.

### Summary

Fixes Bug #76615: `ComposerViewsheetController.saveViewsheet()` could pass a `null` runtime sheet ID into `ComposerViewsheetServiceProxy.saveViewsheet()` when a STOMP save request arrives before a runtime viewsheet ID has been attached to the session, causing an unhandled `NullPointerException` (surfaced through `AffinityKey` construction in the clustered proxy path).

### What I checked

- **Correctness**: The fix adds an early `null` guard on `runtimeViewsheetRef.getRuntimeId()` in `saveViewsheet()`, logging a warning and returning `false`, mirroring the existing, already-proven pattern in `closeViewsheet()` in the same file. Verified by reading the full controller.
- **Callers**: `saveAndCloseViewsheet()` calls `saveViewsheet()` and only sends `CloseSheetCommand` when it returns `true`, so the guard correctly short-circuits the close-after-save path as well with no follow-up NPE.
- **Ordering**: The null check is placed before the `securityEngine.checkPermission()` call. This matches `closeViewsheet()`'s existing behavior (no permission check when there's no runtime ID to act on) and introduces no security regression — no viewsheet data is touched when the ID is null.
- **Frontend impact**: The composer sends `composer/viewsheet/save` as a fire-and-forget STOMP event (`composer-main.component.ts`); there is no code awaiting a boolean reply on this channel, so the change from "unhandled 500/NPE" to "silent no-op + server log warning" is a strict improvement and consistent with the existing UX for `closeViewsheet()` under the same condition.
- **Tests**: Added `ComposerViewsheetControllerTest.saveViewsheetReturnsFalseWhenRuntimeIdIsNull()` correctly constructs the controller via its real constructor, stubs `getRuntimeId()` to return `null`, and asserts both the `false` return value and `verifyNoInteractions(composerViewsheetService, securityEngine)`. Ran it locally:
  - `./mvnw -o test -pl core -Dtest=ComposerViewsheetControllerTest` → `Tests run: 1, Failures: 0, Errors: 0` — passes.
- **Scope**: Change is minimal (5 lines in the controller + a new, self-contained unit test). No unrelated refactoring.
- **No subclasses/overrides** of `ComposerViewsheetController.saveViewsheet()` exist in the community repo that would be affected by the new early return.

### Issues found

None blocking. No regressions, no backward-incompatibility, no broken tests identified.

### CI / mergeability

- All checks pass except `claude-review`, which failed due to an infra/quota error unrelated to code quality (see above) — this manual review stands in for it.
- No merge conflicts; PR is `MERGEABLE`. The `BLOCKED` merge state is solely due to the branch-protection "review required" gate, which this review satisfies.

**Verdict: Approve.**

PR #5187, opened 2026-09-11
Bug #76573, force change detection after deferred column-width recalculation

jkidacki-inetsoft:
## Summary

The "Show Summary Data"/"Show Details" table in the chart mini-toolbar (`preview-table.component.ts`) stays blank when the viewer is hosted inside the elements/viewer-element Angular Elements web-component bundle — real data is bound into the DOM (confirmed via inspection: correct row/cell content present) but no cells are visible until an unrelated DOM event (e.g. a scroll) forces a repaint.

## Root Cause

`columnIndexRange` — the range that gates whether any header/body `<th>`/`<td>` cell renders at all in the template — is computed by `updateColumnRange()`, which is invoked from three places that all defer via a bare `setTimeout()` or `Promise.resolve().then()`:
- `updateWidths()`'s `setTimeout(..., 0)`
- the `tableData` setter's `Promise.resolve().then()` block
- the `colWidths` setter's `Promise.resolve().then()` block

These callbacks run outside any Angular-bound event/template binding, so nothing schedules a view check afterward once this component is hosted inside the elements/viewer-element bundle (`createApplication()` + `createCustomElement()`). The computed layout is correct, but the template never re-evaluates against it — until an unrelated bound DOM event (e.g. the container's `(scroll)` handler, which calls the same `updateColumnRange()` directly and gets a check "for free" as part of Angular's normal event-binding change detection) happens to trigger a repaint.

## Fix

Call `this.changeRef.detectChanges()` at the end of each of the three deferred callbacks, so the newly computed column layout is reflected in the view immediately rather than waiting on an unrelated event.

## Test Plan

- Reproduced live against the ticket's own `WebComponentCase.html` (built `elements.js`/`viewer-element.js`, uploaded to EM Data Space): before the fix, "Show Summary Data" opened a blank dialog reliably; data was confirmed present in the DOM (171 rows, ~29KB innerHTML) but with zero rendered cells until a scroll.
- After the fix: 10 consecutive open/close cycles across both charts on the same page rendered correctly immediately, no click/scroll needed.
- Existing `preview-table.component.*.tl.spec.ts` and `vs-preview-table.component.tl.spec.ts` suites: 97 tests, all passing.
- `eslint` clean on the changed file.

Fixes Redmine #76573.

### DIFF
--- a/web/projects/portal/src/app/vsobjects/objects/table/preview-table.component.ts
+++ b/web/projects/portal/src/app/vsobjects/objects/table/preview-table.component.ts
@@ -218,6 +218,9 @@ export class PreviewTableComponent implements OnDestroy, AfterViewChecked, After
                this.initColumnWidths();
                this.updateHorizontalDist();
                this.updateColumnRange();
+               // Bug #76573: see the matching comment in updateWidths() -- this deferred
+               // callback needs its own view check too.
+               this.changeRef.detectChanges();
             }
          });
       }
@@ -280,6 +283,9 @@ export class PreviewTableComponent implements OnDestroy, AfterViewChecked, After
             this.initColumnWidths();
             this.renderer.setProperty(this.previewContainer.nativeElement, "scrollLeft",
                this.scrollXPos);
+            // Bug #76573: see the matching comment in updateWidths() -- this deferred
+            // callback needs its own view check too.
+            this.changeRef.detectChanges();
          });
       }
    }
@@ -488,6 +494,14 @@ export class PreviewTableComponent implements OnDestroy, AfterViewChecked, After
 
       setTimeout(() => {
          this.updateColumnRange();
+
+         // Bug #76573: this callback runs outside any Angular-bound event, so nothing
+         // triggers a view check afterward when hosted inside the elements/viewer-element
+         // Angular Elements bundle -- columnIndexRange (computed above, and gating every
+         // header/body cell in the template) would stay applied-but-unrendered until an
+         // unrelated DOM event (e.g. scroll, which calls updateColumnRange() directly from
+         // a bound (scroll) handler and so gets a check for free) forced a repaint.
+         this.changeRef.detectChanges();
       }, 0);
    }
 


jshobe-inetsoft (approved):
## Code Review — PR #5187

**Bug #76573**: Force change detection after deferred column-width recalculation in `preview-table.component.ts`.

### Summary of change
Adds `this.changeRef.detectChanges()` at the end of the three deferred callbacks (`setTimeout` in `updateWidths()`, and the `Promise.resolve().then()` blocks in the `tableData` and `colWidths` setters) that compute `columnIndexRange` — the value gating every header/body `<th>`/`<td>` in the template (confirmed at `preview-table.component.html` lines 31, 108, 169). This matches the stated root cause: those callbacks run outside any Angular-bound event, so when the component is hosted inside the `elements`/`viewer-element` Angular Elements bundle, nothing schedules a view check afterward, leaving the DOM populated but unrendered until an unrelated bound event (e.g. scroll) forces a repaint.

### Verification performed
- Checked out the PR commit (`a0076cfb`) and confirmed the diff is exactly the 14 added lines described (3 `detectChanges()` calls + comments), no other files touched.
- Confirmed `columnIndexRange` is indeed the `@for` loop-gating field in the template, supporting the stated root cause.
- Ran the existing suites (`preview-table.component.*.tl.spec.ts`, `vs-preview-table.component.tl.spec.ts`) against the PR branch: **97/97 tests pass**.
- Ran `eslint` on the changed file: clean, no warnings/errors.
- Verified this is consistent with an established pattern already used elsewhere in this same file and in sibling components (`base-table.ts`, `vs-table.component.ts`, `vs-table-cell.component.ts`, `vs-crosstab.component.ts`), which also call `changeDetectorRef.detectChanges()` unconditionally from `setTimeout`/debounce callbacks.

### Findings

**No blocking issues.**

Minor, non-blocking observations:
1. **No new regression test.** The fix doesn't add a test that reproduces the original failure mode (out-of-zone / Angular-Elements-hosted change detection). This is understandable given how hard that scenario is to simulate under TestBed, and the existing 97 tests continue to pass, but a future contributor won't have a test that would catch a revert of this fix.
2. **Theoretical destroyed-view risk.** `detectChanges()` is called from a macrotask/microtask that isn't guaranteed to run before the component could be destroyed (e.g., user closes the "Show Summary Data"/"Show Details" dialog quickly), and `ngOnDestroy()` only calls `clearListeners()` — it doesn't cancel the pending `setTimeout`/`Promise`. Calling `detectChanges()` on a destroyed view can throw. That said, this exact unguarded pattern is already pervasive throughout this file and sibling table components, so this PR doesn't introduce a new class of risk — it's consistent with existing conventions.

### Conclusion
The change is minimal, well-scoped, well-commented, and correctly targets the root cause. Tests pass, lint is clean, and the fix follows established patterns in the codebase.

**Recommendation: Approve.**

PR #5188, opened 2026-09-11
Chart palettes: re-tune Modern and Modern Dark, add Contrast, scope the picker per chart

FigurePaper:
Three things, all inside the modern-visualization gate: the eight head colours of `Modern` and `Modern Dark` are re-tuned, a `Contrast` palette is added, and the palette picker is scoped by a chart's own provenance mark instead of the org property.

Taken from the external design set `SBI Color and Type Pairings.dc.html` and its `design_handoff_chart_palettes/` folder. That handoff proposed reducing 11 palette names to 6 and could not be applied as written — four reasons, each verified against the branch, are recorded in the design doc along with the corrections.

## What changed

- `Modern` and `Modern Dark` indexes 1-8 take the redesigned bases. Indexes 9-40, the legacy tail, are untouched.
- `Contrast` is new: eight slots for projection, monochrome print and low vision, every member on its own rung of a luminance ladder. No dark variant, deliberately — its contract is paper and projection.
- A modern-marked chart is no longer offered the nine single-hue ramps (Pastel, Heat 8/16/24, Blue, Green, Red, Orange, Gray). A classic chart keeps everything it has today, and `Contrast` is offered to both, because an accessibility palette is not a look-and-feel choice.

## What to know before reviewing

**The response flags palettes, it never omits them.** The dialog identifies a chart's current palette by matching colour lists. Omitting a hidden palette would make a chart sitting on one match nothing, fall back to showing "Default", and repaint its colours on a no-op OK. So every palette is sent with a `hidden` flag and only the dropdown filters. `getPaletteNames()` and `getPalette(name)` stay unfiltered — rendering resolves palettes by name, so filtering either would break rendering, not just the picker.

**The picker reads the chart's mark, not the org gate.** The gate is creation-time only: a dashboard already using the modern look keeps it when an admin switches the property off. A picker reading the property would strand those charts with a list that does not match what they render. The proxy carries only the `VizMark` rather than the whole list, because a null `vsId` has no routing key and the controller has to branch on it regardless.

**The target-band colour pane has no assembly to resolve**, so it falls through to the org gate as the absent-assembly default. It deliberately shows every palette: it cannot tell a classic chart from a modern one, and filtering it on the org property would hide the retired palettes from classic charts in a modern org.

**The head colours are declared twice** — in `defaults.css` and as Java fallback constants used when a CSS rule is missing or malformed. A new test reads both by reflection and compares all eight slots against the live frame, so the two can no longer drift silently. That guard did not exist before and is the most durable thing in this PR.

**`VSChartBindingController` is now a second documented survivor** in `VizContextReadFlipTest`, which asserts tree-wide that nothing reads the org gate where a per-assembly mark is available. The absent-assembly branch has no mark to read, which is the same category as the existing survivor. Both are named with their justification, so a third still fails the guard.

## Not in scope

The handoff's sequential and diverging ramps (`Amber`, `Teal`, `Variance`) are `LinearColorFrame` work, not CSS — each needs a frame subclass, a wrapper and a `VisualFrameWrapper` case, and unlike palettes, ramps are persisted by class name. The companion-colour system and the palette-overflow generator are also deferred. The design doc records each with the trigger that brings it back.

## Verification

- Core suite: 5740 tests, 0 failures
- Portal suite: 225 files, 1458 tests, 0 failures
- Manual dark-mode pass on a real dashboard

The full core run is what caught the `VizContextReadFlipTest` regression; eight green scoped runs did not, because the failing class is one no task touched.

## Known, filed separately

The manual pass surfaced that selecting any assembly in dark mode draws a dotted orange border, because `.bd-selected-assembly` binds to `--inet-primary-color`, which has no dark value, while its neighbour `.bd-selected-cell` resolves through a viz token that does. It pre-dates this branch — the class is from the initial commit, and this branch changes no stylesheets. Written up in `2026-09-11-dark-assembly-selection-border-ticket.md` with the measurements, including that the re-tune moves the cell-selection collision from an exact colour match to a near one.

Also filed for later, not fixed here: `getPaletteIndex()` does not reset its counter before the reverse-match scan, so a forward-partial plus a reverse-partial can register a false reversed match. Pre-existing, and likelier with short palettes.

### DIFF
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/VSChartPaletteDefaults.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/VSChartPaletteDefaults.java
@@ -29,6 +29,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
@@ -87,6 +88,15 @@ public final class VSChartPaletteDefaults {
       if(frame != null && ctx.modern) {
          frame.setDefaultColors(activePalette(ctx));
       }
+   }
+
+   /**
+    * Palette names a picker should mark hidden for the given context. A modern chart is not
+    * offered the single-hue ramps; a classic chart keeps everything. Never removes a name from
+    * resolution - getPalette must still answer for every one of these.
+    */
+   public static Set<String> hiddenPaletteNames(VizContext ctx) {
+      return ctx.modern ? MODERN_HIDDEN : Set.of();
    }
 
    /**
@@ -174,6 +184,8 @@ public final class VSChartPaletteDefaults {
    private static final String MODERN_NAME = "Modern";
    private static final String DARK_NAME = "Modern Dark";
    private static final String DEFAULT_NAME = "Default";
+   private static final Set<String> MODERN_HIDDEN =
+      Set.of("Pastel", "Heat 8", "Heat 16", "Heat 24", "Blue", "Green", "Red", "Orange", "Gray");
    private static final Map<String, Memo> MEMO = new ConcurrentHashMap<>();
    private static final Logger LOG = LoggerFactory.getLogger(VSChartPaletteDefaults.class);
 
--- a/core/src/main/java/inetsoft/uql/viewsheet/internal/VSChartPaletteDefaults.java
+++ b/core/src/main/java/inetsoft/uql/viewsheet/internal/VSChartPaletteDefaults.java
@@ -178,12 +178,12 @@ public final class VSChartPaletteDefaults {
    private static final Logger LOG = LoggerFactory.getLogger(VSChartPaletteDefaults.class);
 
    private static final Color[] MODERN_HEAD = {
-      new Color(0x00D4E8), new Color(0x00B87A), new Color(0xF59E0B), new Color(0xF43F5E),
-      new Color(0x8B5CF6), new Color(0x3B82F6), new Color(0x0D9488), new Color(0x64748B)
+      new Color(0x0490FF), new Color(0xFF5A35), new Color(0x241C4F), new Color(0x03D9B3),
+      new Color(0x9A2DDC), new Color(0xFFB020), new Color(0xE5197E), new Color(0x8ED604)
    };
 
    private static final Color[] DARK_HEAD = {
-      new Color(0x22D3EE), new Color(0x10B981), new Color(0xFBB724), new Color(0xFB6181),
-      new Color(0xA78BFA), new Color(0x60A5FA), new Color(0x2DD4BF), new Color(0x94A3B8)
+      new Color(0x4FA5FF), new Color(0xFF8367), new Color(0x49447D), new Color(0x2DEEC6),
+      new Color(0xAE41F5), new Color(0xFFCB82), new Color(0xFE3290), new Color(0x9FEB28)
    };
 }
--- a/core/src/main/java/inetsoft/web/binding/VSChartBindingController.java
+++ b/core/src/main/java/inetsoft/web/binding/VSChartBindingController.java
@@ -243,7 +243,9 @@ public class VSChartBindingController {
       Principal principal)
       throws Exception
    {
-      VizContext ctx = vsId == null || assemblyName == null
+      // absent assembly resolves through the org gate; its flags are unread by any UI -
+      // don't gate the target-band pane on them, it can't tell classic charts from modern ones
+      VizContext ctx = Tool.isEmptyString(vsId) || Tool.isEmptyString(assemblyName)
          ? VizContext.ofGate()
          : VizContext.of(chartBindingService.getChartVizMark(vsId, assemblyName, principal));
       Set<String> hidden = VSChartPaletteDefaults.hiddenPaletteNames(ctx);
--- a/core/src/main/java/inetsoft/web/binding/VSChartBindingController.java
+++ b/core/src/main/java/inetsoft/web/binding/VSChartBindingController.java
@@ -30,7 +30,9 @@ import inetsoft.uql.viewsheet.*;
 import inetsoft.uql.viewsheet.graph.*;
 import inetsoft.uql.viewsheet.graph.aesthetic.*;
 import inetsoft.uql.viewsheet.internal.ChartVSAssemblyInfo;
+import inetsoft.uql.viewsheet.internal.VSChartPaletteDefaults;
 import inetsoft.uql.viewsheet.internal.VSUtil;
+import inetsoft.uql.viewsheet.internal.VizContext;
 import inetsoft.util.Tool;
 import inetsoft.web.binding.handler.*;
 import inetsoft.web.binding.model.*;
@@ -234,9 +236,17 @@ public class VSChartBindingController {
    @RequestMapping(value = "/api/composer/chart/colorpalettes", method = RequestMethod.GET)
    @HandleExceptions
    @SwitchOrg
-   public CategoricalColorModel[] getColorPalettes(@OrganizationID String orgId, Principal principal)
+   public CategoricalColorModel[] getColorPalettes(
+      @OrganizationID String orgId,
+      @RequestParam(required = false, value = "vsId") String vsId,
+      @RequestParam(required = false, value = "assemblyName") String assemblyName,
+      Principal principal)
       throws Exception
    {
+      VizContext ctx = vsId == null || assemblyName == null
+         ? VizContext.ofGate()
+         : VizContext.of(chartBindingService.getChartVizMark(vsId, assemblyName, principal));
+      Set<String> hidden = VSChartPaletteDefaults.hiddenPaletteNames(ctx);
       String[] names = ColorPalettes.getPaletteNames().toArray(new String[0]);
       CategoricalColorModel[] palettes = new CategoricalColorModel[names.length];
 
@@ -246,6 +256,7 @@ public class VSChartBindingController {
          wrapper.setVisualFrame(palette);
          CategoricalColorModel model = visualService.createVisualFrameModel(wrapper);
          model.setName(names[i]);
+         model.setHidden(hidden.contains(names[i]));
          palettes[i] = model;
       }
 
--- a/core/src/main/java/inetsoft/web/binding/VSChartBindingController.java
+++ b/core/src/main/java/inetsoft/web/binding/VSChartBindingController.java
@@ -48,6 +48,8 @@ import inetsoft.web.viewsheet.SwitchOrg;
 import jakarta.servlet.ServletOutputStream;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.bind.annotation.*;
 
@@ -243,12 +245,8 @@ public class VSChartBindingController {
       Principal principal)
       throws Exception
    {
-      // absent assembly resolves through the org gate; its flags are unread by any UI -
-      // don't gate the target-band pane on them, it can't tell classic charts from modern ones
-      VizContext ctx = Tool.isEmptyString(vsId) || Tool.isEmptyString(assemblyName)
-         ? VizContext.ofGate()
-         : VizContext.of(chartBindingService.getChartVizMark(vsId, assemblyName, principal));
-      Set<String> hidden = VSChartPaletteDefaults.hiddenPaletteNames(ctx);
+      Set<String> hidden = VSChartPaletteDefaults.hiddenPaletteNames(
+         pickerContext(vsId, assemblyName, principal));
       String[] names = ColorPalettes.getPaletteNames().toArray(new String[0]);
       CategoricalColorModel[] palettes = new CategoricalColorModel[names.length];
 
@@ -265,6 +263,32 @@ public class VSChartBindingController {
       return palettes;
    }
 
+   /**
+    * The context a palette picker's hidden flags are computed from.
+    *
+    * A caller with no assembly resolves through the org gate. Its flags are unread by any UI, and
+    * the target-band pane must not be gated on them - it cannot tell a classic chart from a modern
+    * one, so gating it would hide the retired palettes from classic charts in a modern org.
+    *
+    * A mark that cannot be resolved falls back the same way rather than failing the request. The
+    * mark decides which names carry the flag, never which palettes are returned, so a stale or
+    * expired runtime should not take the whole list down with it.
+    */
+   private VizContext pickerContext(String vsId, String assemblyName, Principal principal) {
+      if(Tool.isEmptyString(vsId) || Tool.isEmptyString(assemblyName)) {
+         return VizContext.ofGate();
+      }
+
+      try {
+         return VizContext.of(chartBindingService.getChartVizMark(vsId, assemblyName, principal));
+      }
+      catch(Exception ex) {
+         LOG.debug("Failed to resolve the chart's mark for the palette picker", ex);
+         return VizContext.ofGate();
+      }
+   }
+
    private final VisualFrameModelFactoryService visualService;
    private VSChartBindingServiceProxy chartBindingService;
+   private static final Logger LOG = LoggerFactory.getLogger(VSChartBindingController.class);
 }
--- a/core/src/main/java/inetsoft/web/binding/VSChartBindingService.java
+++ b/core/src/main/java/inetsoft/web/binding/VSChartBindingService.java
@@ -30,6 +30,7 @@ import inetsoft.uql.erm.DataRef;
 import inetsoft.uql.viewsheet.*;
 import inetsoft.uql.viewsheet.graph.*;
 import inetsoft.uql.viewsheet.internal.ChartVSAssemblyInfo;
+import inetsoft.uql.viewsheet.internal.VizMark;
 import inetsoft.util.Tool;
 import inetsoft.web.binding.handler.*;
 import inetsoft.web.binding.model.*;
@@ -96,6 +97,26 @@ public class VSChartBindingService {
       }
 
       return model;
+   }
+
+   /**
+    * The chart's own provenance mark, read on the node that owns the runtime viewsheet. Null for
+    * an unmarked chart, a missing assembly, or an assembly that is not a chart.
+    */
+   @ClusterProxyMethod(WorksheetEngine.CACHE_NAME)
+   public VizMark getChartVizMark(@ClusterProxyKey String vsId, String assemblyName,
+                                  Principal principal)
+      throws Exception
+   {
+      ViewsheetService engine = viewsheetService;
+      RuntimeViewsheet rvs = engine.getViewsheet(Tool.byteDecode(vsId), principal);
+      Viewsheet viewsheet = rvs.getViewsheet();
+
+      if(viewsheet == null || !(viewsheet.getAssembly(assemblyName) instanceof ChartVSAssembly chart)) {
+         return null;
+      }
+
+      return chart.getVSAssemblyInfo().getVizMark();
    }
 
    /**
--- a/core/src/main/java/inetsoft/web/binding/model/graph/aesthetic/CategoricalColorModel.java
+++ b/core/src/main/java/inetsoft/web/binding/model/graph/aesthetic/CategoricalColorModel.java
@@ -174,6 +174,14 @@ public class CategoricalColorModel extends ColorFrameModel {
       this.colorValueFrame = colorValueFrame;
    }
 
+   public boolean isHidden() {
+      return hidden;
+   }
+
+   public void setHidden(boolean hidden) {
+      this.hidden = hidden;
+   }
+
    private String[] colors;
    private String[] cssColors;
    private String[] defaultColors;
@@ -183,4 +191,5 @@ public class CategoricalColorModel extends ColorFrameModel {
    private boolean shareColors = true;
    private boolean colorValueFrame;
    private Integer dateFormat;
+   private boolean hidden;
 }
--- a/core/src/main/resources/inetsoft/util/css/defaults.css
+++ b/core/src/main/resources/inetsoft/util/css/defaults.css
@@ -176,35 +176,35 @@ ChartPalette[name='Default'][index='40'] {
 }
 
 ChartPalette[name='Modern'][index='1'] {
-   color: #00d4e8;
+   color: #0490ff;
 }
 
 ChartPalette[name='Modern'][index='2'] {
-   color: #00b87a;
+   color: #ff5a35;
 }
 
 ChartPalette[name='Modern'][index='3'] {
-   color: #f59e0b;
+   color: #241c4f;
 }
 
 ChartPalette[name='Modern'][index='4'] {
-   color: #f43f5e;
+   color: #03d9b3;
 }
 
 ChartPalette[name='Modern'][index='5'] {
-   color: #8b5cf6;
+   color: #9a2ddc;
 }
 
 ChartPalette[name='Modern'][index='6'] {
-   color: #3b82f6;
+   color: #ffb020;
 }
 
 ChartPalette[name='Modern'][index='7'] {
-   color: #0d9488;
+   color: #e5197e;
 }
 
 ChartPalette[name='Modern'][index='8'] {
-   color: #64748b;
+   color: #8ed604;
 }
 
 ChartPalette[name='Modern'][index='9'] {
@@ -336,35 +336,35 @@ ChartPalette[name='Modern'][index='40'] {
 }
 
 ChartPalette[name='Modern Dark'][index='1'] {
-   color: #22d3ee;
+   color: #4fa5ff;
 }
 
 ChartPalette[name='Modern Dark'][index='2'] {
-   color: #10b981;
+   color: #ff8367;
 }
 
 ChartPalette[name='Modern Dark'][index='3'] {
-   color: #fbb724;
+   color: #49447d;
 }
 
 ChartPalette[name='Modern Dark'][index='4'] {
-   color: #fb6181;
+   color: #2deec6;
 }
 
 ChartPalette[name='Modern Dark'][index='5'] {
-   color: #a78bfa;
+   color: #ae41f5;
 }
 
 ChartPalette[name='Modern Dark'][index='6'] {
-   color: #60a5fa;
+   color: #ffcb82;
 }
 
 ChartPalette[name='Modern Dark'][index='7'] {
-   color: #2dd4bf;
+   color: #fe3290;
 }
 
 ChartPalette[name='Modern Dark'][index='8'] {
-   color: #94a3b8;
+   color: #9feb28;
 }
 
 ChartPalette[name='Modern Dark'][index='9'] {
--- a/core/src/main/resources/inetsoft/util/css/defaults.css
+++ b/core/src/main/resources/inetsoft/util/css/defaults.css
@@ -862,3 +862,35 @@ ChartPalette[name='Heat 24'][index='23'] {
 ChartPalette[name='Heat 24'][index='24'] {
   color: #ffcc99;
 }
+
+ChartPalette[name='Contrast'][index='1'] {
+   color: #0b3d91;
+}
+
+ChartPalette[name='Contrast'][index='2'] {
+   color: #f5943f;
+}
+
+ChartPalette[name='Contrast'][index='3'] {
+   color: #b04ab8;
+}
+
+ChartPalette[name='Contrast'][index='4'] {
+   color: #f7d22e;
+}
+
+ChartPalette[name='Contrast'][index='5'] {
+   color: #a8330a;
+}
+
+ChartPalette[name='Contrast'][index='6'] {
+   color: #7cc4ee;
+}
+
+ChartPalette[name='Contrast'][index='7'] {
+   color: #00947f;
+}
+
+ChartPalette[name='Contrast'][index='8'] {
+   color: #5fa83c;
+}
--- a/core/src/test/java/inetsoft/report/composition/graph/VGraphPairModernPaletteTest.java
+++ b/core/src/test/java/inetsoft/report/composition/graph/VGraphPairModernPaletteTest.java
@@ -53,7 +53,7 @@ class VGraphPairModernPaletteTest {
       // same resolver call the VGraphPair seam makes after setParentParams
       inetsoft.uql.viewsheet.internal.VSChartPaletteDefaults.applyModernPalette(
          frame, inetsoft.uql.viewsheet.internal.VizContext.ofGate());
-      assertEquals(new Color(0x00D4E8), frame.getColor(0));
+      assertEquals(new Color(0x0490FF), frame.getColor(0));
    }
 
    @Test
--- a/core/src/test/java/inetsoft/uql/viewsheet/graph/aesthetic/ColorPalettesModernTest.java
+++ b/core/src/test/java/inetsoft/uql/viewsheet/graph/aesthetic/ColorPalettesModernTest.java
@@ -86,4 +86,27 @@ class ColorPalettesModernTest {
          assertNotNull(frame.getDefaultColor(i), "index " + (i + 1) + " must not be null");
       }
    }
+
+   @Test
+   void contrastIsRegisteredWithEightColors() {
+      assertTrue(ColorPalettes.getPaletteNames().contains("Contrast"),
+                 "Contrast palette must be declared in defaults.css");
+
+      CategoricalColorFrame contrast = ColorPalettes.getPalette("Contrast");
+      assertNotNull(contrast);
+      assertEquals(8, contrast.getColorCount());
+
+      for(int i = 0; i < 8; i++) {
+         assertNotNull(contrast.getDefaultColor(i), "index " + (i + 1) + " must not be null");
+      }
+
+      assertEquals(new Color(0x0B3D91), contrast.getDefaultColor(0));
+      assertEquals(new Color(0x5FA83C), contrast.getDefaultColor(7));
+   }
+
+   // Default must be declared first: the picker's index-0 fallback relies on it.
+   @Test
+   void defaultIsStillDeclaredFirst() {
+      assertEquals("Default", ColorPalettes.getPaletteNames().iterator().next());
+   }
 }
--- a/core/src/test/java/inetsoft/uql/viewsheet/graph/aesthetic/ColorPalettesModernTest.java
+++ b/core/src/test/java/inetsoft/uql/viewsheet/graph/aesthetic/ColorPalettesModernTest.java
@@ -45,12 +45,12 @@ class ColorPalettesModernTest {
    @Test
    void modernHeadMatchesSpec() {
       CategoricalColorFrame modern = ColorPalettes.getPalette("Modern");
-      assertEquals(new Color(0x00D4E8), modern.getDefaultColor(0));
-      assertEquals(new Color(0x64748B), modern.getDefaultColor(7));
+      assertEquals(new Color(0x0490FF), modern.getDefaultColor(0));
+      assertEquals(new Color(0x8ED604), modern.getDefaultColor(7));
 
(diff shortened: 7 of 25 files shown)


FigurePaper (commented):
# Code review — chart palette re-tune, Contrast, per-chart picker scoping

The design reasoning holds up against the code, and the two load-bearing decisions are the right ones: flagging palettes instead of omitting them (an omitted palette would leave a chart sitting on it matching nothing, falling back to `Default`, and repainting on a no-op OK), and keeping `getPaletteNames()` / `getPalette(name)` unfiltered so rendering never loses a name it resolves by. The new `cssHeadMatchesTheJavaFallback` drift guard is the most durable thing in the change: `defaults.css` and `MODERN_HEAD`/`DARK_HEAD` could silently disagree before, and now they cannot.

**Verdict: requesting changes.** No item below is a correctness blocker on its own, but two checks are red and there are a few things worth fixing first. Details on the checks at the bottom — neither red check is caused by this PR, and I have the evidence for that.

## What I verified

- **`Contrast` is a real luminance ladder.** WCAG relative luminance, sorted: Navy `#0b3d91` 0.0545, Brick `#a8330a` 0.1071, Orchid `#b04ab8` 0.1759, Teal `#00947f` 0.2271, Green `#5fa83c` 0.3076, Orange `#f5943f` 0.4095, Sky `#7cc4ee` 0.4994, Yellow `#f7d22e` 0.6606. Monotonic, eight distinct rungs, adjacent contrast ratios 1.20–1.50. The greyscale/projection contract the palette claims actually holds.
- **The eight-index `Contrast` loads correctly.** `ColorPalettes.loadPalettes()` sizes the array by max index, so declaring 1–8 yields an 8-slot frame with no holes; `PaletteDialog.saveChanges()` splices only the first 8 and keeps the chart's existing tail, identical to how `Heat 8` already behaves.
- **Security posture is unchanged.** `getChartVizMark` resolves through `ViewsheetService.getViewsheet(id, principal)` — the same gate every sibling method on `VSChartBindingServiceProxy` already goes through — and returns only a `VizMark` enum. `@ClusterProxyKey` routes on `vsId`, and the controller short-circuits the empty-`vsId` case before calling, so no null routing key reaches the proxy. It is also more defensive than its sibling `getChartBinding`, which casts the assembly without a null check.
- **Backward compatible.** Two optional request params, one additive model field, no persisted contract moved. A chart saved on a now-hidden palette still renders, and still appears in its own dropdown.
- **Frontend tests pass locally** (scoped runs): `palette-dialog.spec.ts` 8/8; `chart-palette.service.spec.ts` + `color-field-pane.spec.ts` + `combined-color-pane.spec.ts` 11/11.
- The three test classes the plan deliberately left alone (`CategoricalColorDerivedPersistenceTest`, `VizModernizeUtilTest:659`, the comment in `CategoricalColorFactoryResolvedDefaultGuardTest`) do use the old hexes only as arbitrary `setDerivedColor` fixtures. Leaving them was the right call.

## Should fix

### 1. The date-comparison chart's palette picker is not scoped by the mark

`date-comparison-pane.component.html:139` renders `categorical-color-pane` with no `vsId`, `assemblyName` or `assetId`, so it lands on the absent-assembly branch and resolves through `VizContext.ofGate()` — precisely the org-gate read the change set out to remove. The consequence is the one the design doc argues against: a modern-marked chart in a gate-off org gets offered the retired ramps, and a classic chart in a modern org has them hidden.

Decision 3 in the design doc names only the target-band pane as the deliberate absent-assembly caller. This is a second one, and unlike the target-band pane it *does* have a specific chart assembly in scope, so it is resolvable. Either thread `[vsId]`/`[assemblyName]` through from the date-comparison dialog, or add this pane to the documented list with its justification so the inventory stays true.

Related and pre-existing, worth its own ticket rather than fixing here: that same pane passes no `assetId`, and `createAssetEntry(undefined)` returns `null`, so the unchanged `.organization` read at `categorical-color-pane.component.ts:112` throws a `TypeError` inside `ngOnInit` before any of the new params matter. `b-categorical-color-pane.component.ts:62` null-guards the same call; this one never has.

### 2. The palette endpoint gained a failure mode it did not have

`getColorPalettes` could not previously fail on account of `vsId`. It now calls `getChartVizMark` which calls `getViewsheet(...)`, and that throws for a stale or expired runtime. When it does, the whole palette list fails rather than degrading: `modelService.getModel` errors, `colorPalettes` stays undefined, and the dialog body renders empty behind its `@if (colorPalettes)`.

The mark only decides which names carry the `hidden` flag — it never decides which palettes are returned — so a failed lookup has no business taking the list down with it. Catching around the mark resolution and falling back to `VizContext.ofGate()` (or `of((VizMark) null)`) would keep the picker usable. The window is narrow, since `/api/composer/binding` already makes the same call against the same runtime, but the blast radius is out of proportion to the cause.

### 3. Stale guard comment left behind by the survivor count change

`VizContextReadFlipTest:133-137`, in `onlyTheDocumentedSitesStillReadTheOrgGate`, still reads: "ofGate() survives in exactly one call site: the parameterless ChartColorPaletteController bootstrap GET, which has no assembly to resolve a mark from."

That now contradicts `exactlyTwoDocumentedSitesStillReadTheOrgGate` three lines below it. These comments are the mechanism by which the next maintainer judges whether a third survivor is legitimate, so a wrong one is worse than none. The `assertEquals(1, countOfGateIn("web/portal/controller"))` assertion itself is still correct — `VSChartBindingController` lives in `web/binding` — only the prose needs updating.

## Worth confirming

### Dashboards saved before the re-tune

`CategoricalColorFrameWrapper.writeContents` persists `frame.getDefaultColor(i)` for every index into the viewsheet XML, and `ChartVSAssemblyInfo.seedColorPalette` writes `seedPalette(ctx)` into exactly those defaults at creation, Modernize and Revert. So a dashboard saved before this PR carries the old Modern head in its stored `<colors>`.

Rendering should recover — `VGraphPair:1300` re-applies `applyModernPalette` on every render of a modern-marked chart — but `PaletteDialog.getPaletteIndex()` matches by exact colour list and falls back to index 0 on no match. That is the same "match nothing, show Default, repaint on a no-op OK" failure the PR body correctly guards against for hidden palettes, arriving by a different route. Could you confirm on an existing modern dashboard (not a freshly created chart) that the picker still pre-selects `Modern` after the re-tune? If it does this is a non-issue, and worth a sentence in the design doc so the next re-tune does not have to re-derive it.

## Minor

- **`paletteSelectOptions` leans on a side effect.** Reading `this.displayPalette` on `palette-dialog.component.ts:43` is there to force the lazy `_selectedIndex`/`_reversed` match to run before the filter; collapsing it to `this._selectedIndex` looks obviously equivalent and would silently break "keep the palette the chart is on", because `_selectedIndex` is still `-1` on the first pass. A short clause saying so would protect it from a well-meaning simplification. Separately, the getter is evaluated on every change-detection pass and, when `_reversed` is set, now also allocates a reversed 40-element copy each time — harmless, but it compounds the pattern the template already had.
- **No test covers a hidden palette selected reversed.** `_selectedIndex` points at the hidden entry and the filter keeps it, so it should be fine; a one-line case would pin it.
- **Roadmap entry contradicts the PR that adds it.** The bullet added at `chart-card-roadmap.md:1739` reads "decided, not implemented" while this PR implements it. Since CLAUDE.md makes that file required reading before touching anything behind the gate, it should not ship pointing the wrong way.
- **Count mismatch in the plan.** The Goal line says "no longer offered the eight single-hue ramps"; `MODERN_HIDDEN` holds nine, which is what the PR body and every test assert.

## Checks

Both red checks were investigated. Neither is attributable to this change, but they do block approval under the merge rules:

- **`build` (Validation Build) — failed.** `inetsoft-web`, `npm run verify`. Every Java module passed (`inetsoft-core` SUCCESS, no non-zero `Failures`/`Errors` anywhere in surefire), and in the `em:tl` suite all 71 files and 839 tests passed. The run failed on a single unhandled error: `TypeError: Cannot read properties of undefined (reading 'type')` at `task-condition-pane.component.html:30`, originating in `schedule-task-editor-page.component.tl.spec.ts`. I reproduced it locally on this branch running that one spec in isolation, then reverted this PR's only shared-file change (`web/mocks/handlers/portal.handlers.ts`, hex strings only) to the `epic-74519` version and reproduced it again — so it is independent of this PR. This PR touches no EM code, no schedule code, and no file that spec loads. It needs a fix on the base branch, or its own ticket, before this can go green.
- **`claude-review` — failed.** Infrastructure, not a code signal: the job errored after 0s with `is_error: true`, `total_cost_usd: 0`, `modelUsage: {}` and an empty `ANTHROPIC_API_KEY`. It produced no findings. Worth re-running once credentials are in place, since it never actually looked at the diff.

No merge conflicts — the PR is `MERGEABLE`; `BLOCKED` reflects the required-review gate only.

Once the three "Should fix" items are addressed and the build goes green, this is good to merge.

FigurePaper:
Addressed the three "Should fix" items and the four minors. Two things are worth recording here rather than leaving to the diff.

## Item 1 — fixing the inventory, not threading the params

The date-comparison pane passes no `assetId`, so `ngOnInit` calls `createAssetEntry(undefined).organization` and throws before the request is ever made — `createAssetEntry` returns null when the id does not parse. That line is byte-identical on `epic-74519`, so the pane cannot reach the org-gate branch at all today, and the consequence described cannot occur there. `visual-dropdown-pane` is in the same position. Threading `vsId` and `assemblyName` through would need new inputs on the pane and on its dialog chain, on a path that is broken upstream of them.

Fixed the inventory half instead: decision 3 now names all three hosts that render `categorical-color-pane`, says which one supplies assembly context, and records that two of them throw before the request. The pre-existing throw is worth its own ticket.

## The build failure is a flaky pre-existing race

One correction first: the review says this PR "touches no file that spec loads". That is not right — `projects/em/src/vitest-setup-tl.ts` imports `@test-mocks/server`, which imports `portal.handlers.ts`, which this PR edits. So I tested it rather than assuming. The spec throws 51 times on this branch, and 51 times with `portal.handlers.ts` reverted to its base version. The edit is eight hex strings in the response body of a portal endpoint the EM schedule page never calls.

What actually differs is the base. PR #5187 targets `main`, not `epic-74519`, so its green build was never comparable — and its passing log carries the same `TypeError: Cannot read properties of undefined (reading 'type')` 49 times, with the same `839 passed | 19 expected fail`. The only thing missing from it is the `Unhandled Errors` block. The error is thrown asynchronously during change detection after its test has finished, so whether Vitest catches one while still attached to the run is timing-dependent. Every test passes either way.

Supporting evidence: `task-condition-pane.component.html` and `schedule-task-editor-page.component.tl.spec.ts` are identical on both bases, and the `test` and `verify` scripts are identical, so both bases run the `em:tl` suite. No CI has run on `epic-74519` since #5150 merged, since only pull requests trigger builds — which is why this surfaced here first rather than earlier.

Re-running the check should go green. The uncaught exception is still a real defect in the EM schedule pane and deserves its own ticket; it just does not fail deterministically.

FigurePaper:
# Code review, round 2 — chart palette re-tune, Contrast, per-chart picker scoping

All three "Should fix" items and all four minors from round 1 are addressed, and the `build` check is now green. One new finding below, plus a correction to something the PR body files for later that turns out not to be reachable.

**Verdict: requesting changes**, for the reasons at the bottom. The code is in good shape; the blocker is a red check plus one stale doc.

## Round-1 items — verified

**Item 1 (date-comparison pane) — the counter-argument is right, and I checked it rather than taking it.** `createAssetEntry` returns `null` when the id does not match its regex (`shared/data/asset-entry.ts:37-55`). `date-comparison-pane.component.html:139` and `visual-dropdown-pane.component.html:27` both render `categorical-color-pane` with no `assetId`, and `categorical-color-pane.component.ts:112` dereferences `.organization` with no guard, inside a `getColorPalettes()` called directly from `ngOnInit`. So both hosts throw before the HTTP call is built, and neither can reach the org-gate branch today. `b-categorical-color-pane.component.ts:62-65` is the one that guards it. Threading `vsId`/`assemblyName` into a path that is broken upstream of them would have been the wrong fix; fixing the inventory was the right call. Decision 3 now names all four callers and which one supplies assembly context.

**Item 2 — fixed.** `pickerContext()` catches and falls back to `VizContext.ofGate()`. The list can no longer be taken down by a stale runtime. Swallowing `Exception` is right here rather than over-broad: the mark only decides which names carry the flag, the response shape is identical either way, and access to the runtime is still gated by `getViewsheet(id, principal)` on the path that succeeds.

**Item 3 — fixed.** `VizContextReadFlipTest:135-140` now reads "Two call sites survive" and names both, and the per-package assertion message no longer overclaims. It agrees with `exactlyTwoDocumentedSitesStillReadTheOrgGate` three lines below.

**Minors — all four fixed.** The side-effect clause on `paletteSelectOptions`; the reversed-hidden test; the roadmap bullet now reads "implemented, in review"; the plan's Goal line now says nine. I ran `palette-dialog.spec.ts` on this head: 9/9 pass.

I also re-derived the new reversed test rather than trusting it. Every palette in the fixture shares `LEGACY_TAIL`, so against a reversed Heat-40 the `Default` and `Modern` entries accumulate 32 reverse matches and break at index 32 — close enough that the test would not pass by accident, and only `Heat 8` reaches 40. The assertion that `_reversed` is `true` only after reading `paletteSelectOptions` is also correct, since that getter is now what forces the lazy match.

## New — should fix

### `visualization-palette-swatches.html` still shows the old head colours

`docs/superpowers/specs/lookfeel/visualization-palette-swatches.html` declares `--series-1..8` and `--series-dark-1..8` as the pre-re-tune hexes (lines 33-48), and renders them as swatch cards at lines 726-733 and 774-781 with the hexes printed as label text. The section note at line 724 describes that block as "Phase 8 gated modern categorical default (series 1-8). Bridged server-side via VSChartPaletteDefaults into CategoricalColorFrame.defaultColors" — it is documenting the exact constants this PR moves, and it names the class.

This is not a stray doc. The plan and spec tree repeatedly calls this file the source of truth for modern values, and prior phases treated updating it as part of the change:

- `visualization-phase4-implementation-plan.md:51` — "Modern **state** palette source of truth exists | `visualization-palette-swatches.html:20-31`"
- `2026-07-23-viz-phase9c-item4-highlight-semantic-presets.md:17` — "source of truth: `visualization-palette-swatches.html`"
- `2026-07-27-visualization-phase9b-dark-mode.md:26` — "Canonical dark values are recorded in ... `visualization-palette-swatches.html`"
- `2026-07-24-...-item5` and `2026-07-29-...-phase6b` both list it under "Modify" and `git add` it alongside the code

It is also outside `chart-card-design3/`, so it is not one of the files regenerated wholesale, and CLAUDE.md makes this tree required reading before touching anything behind the gate. Leaving it means the next person reading the designated source of truth gets the palette this PR replaced.

Two things beyond a mechanical hex swap: `Contrast` has no card at all, and the chips carry hand-picked label colours (`color:var(--light-chart-label)` vs `#FFFFFF`) chosen against the old luminances. New `series-3` is `#241C4F`, far darker than the `#F59E0B` it replaces, so at least that pairing has to flip.

## Worth confirming

### `Contrast` slots 7 and 8 are the palette's weak pair

The declaration order is well chosen — it interleaves the ladder so consecutive series stay far apart in greyscale. Adjacent WCAG-luminance contrast ratios, computed from the CSS:

| pair | L | L | ratio |
|---|---|---|---|
| 1 Navy → 2 Orange | 0.0545 | 0.4095 | 4.40 |
| 2 Orange → 3 Orchid | 0.4095 | 0.1759 | 2.03 |
| 3 Orchid → 4 Yellow | 0.1759 | 0.6606 | 3.15 |
| 4 Yellow → 5 Brick | 0.6606 | 0.1071 | 4.52 |
| 5 Brick → 6 Sky | 0.1071 | 0.4994 | 3.50 |
| 6 Sky → 7 Teal | 0.4994 | 0.2271 | 1.98 |
| **7 Teal → 8 Green** | **0.2271** | **0.3076** | **1.29** |

Teal `#00947F` and Green `#5FA83C` are adjacent in declaration order, adjacent rungs on the sorted ladder, and in the same hue family. For a palette whose stated contract is monochrome print and projection, that is the one pair a reader could lose, and it only bites at 7+ categories. Swapping slot 7 with a lighter member — putting Sky between them, for instance — would raise the weakest consecutive pair to about 1.5 without disturbing the set. If the 1.29 is a deliberate acceptance from the handoff, a line in the design doc saying so would stop the next reviewer re-deriving this.

## Minor

- **Nothing exercises the new catch.** `VSChartBindingColorPalettesTest` builds the controller with a `null` `chartBindingService`, so `controller.getColorPalettes(null, "vs-1", "Chart1", null)` would NPE inside `pickerContext`, hit the new catch, and fall back to the gate — a two-line test for the branch round 1 asked for, using wiring the class already has.
- **The head hexes are now declared in four places, and the guard covers two.** `cssHeadMatchesTheJavaFallback` pins `defaults.css` against `MODERN_HEAD`/`DARK_HEAD`. `palette-test-fixtures.ts:24-32` and `portal.handlers.ts:130` are a third and fourth copy with nothing pinning them to the CSS, so if they drift the frontend suites keep passing against stale values. Lower stakes than the CSS/Java pair, since drift there buys false confidence rather than a wrong render, but worth a note near the fixtures so the next re-tune knows they exist.

## Correction to the PR body

The body files `getPaletteIndex()`'s missing counter reset as a pre-existing issue, "likelier with short palettes". I do not think it is reachable, and it seems worth recording before someone opens a ticket for it.

A false reversed match needs the forward loop to score `f >= 1` and the reverse loop to add `r >= 1`. Both loops start at `j = 0` and compare against the same `currPalette.colors[0]`: forward needs `pt.colors[0] == curr[0]`, reverse needs `pt.colors[len-1] == curr[0]`. Both can hold only if a palette's first colour equals its last. None does — I checked all fourteen in `defaults.css`, including the new `Contrast` (`#0b3d91` / `#5fa83c`). So `f >= 1` forces `r == 0` and `cnt == f < len`, and `f == len` is caught by the `if` unless `cnt > matchcnt` already failed, in which case the else branch's identical check fails too. The counter is sloppy, but it cannot produce a wrong match against any registered palette. Adding `Contrast` does not change that.

## Also verified this round

- `Contrast` loads as an 8-slot frame: `ColorPalettes.loadPalettes()` sizes `Color[max]` from the max declared index, so 1-8 yields eight entries with no holes.
- All nine `MODERN_HIDDEN` names exist in `defaults.css` — a typo there would silently hide nothing. Fourteen palettes total, nine hidden, five kept, which is what `hiddenNamesNeverCoverTheKeptFive` asserts.
- `VizContext.of((VizMark) null)` gives `modern == false`, so an unmarked chart gets an empty hidden set and keeps every palette.
- The target-band dialog is genuinely unaffected: `graph-palette-dialog.component.ts:43` maps the array unfiltered and never reads `hidden`, so the flags the gate branch computes for it really are inert, as the controller javadoc says.
- `hidden` on `CategoricalColorModel` is a web DTO field only; frames persist through `CategoricalColorFrameWrapper`, so nothing serialized moves.
- No `e2e/` files touched.

## Checks

- **`build` — green** (15m39s). The round-1 red is gone, which supports the flaky-race reading. Your correction stands: `projects/em/src/vitest-setup-tl.ts` does pull in `portal.handlers.ts`, so my "touches no file that spec loads" was wrong.
- **`claude-review` — failed.** Same infrastructure failure as last time, not a code signal: `duration_ms: 451`, `num_turns: 1`, `total_cost_usd: 0`, `modelUsage: {}`, `ANTHROPIC_API_KEY` empty in both the run and post steps. It never made a model call, so it has looked at nothing.
- Semgrep OSS, Semgrep Scan, license/cla all pass. `mergeable: MERGEABLE` — no conflicts; `BLOCKED` is the required-review gate.

**Requesting changes for:** the failing `claude-review` check, which the merge rules treat as blocking regardless of cause, and the stale `visualization-palette-swatches.html`. Fix the swatch page, get `claude-review` credentials in place or the check re-run green, and this is ready — the two "worth confirming"/"minor" sections are not gating.

FigurePaper:
Round 2 addressed. The swatch page, the catch test and the four-copies note are in the diff; two items need a word here.

## Swatch page

Updated the sixteen series variables and all sixteen chips, and added `--contrast-1..8` with a `Contrast` card. The chip label colours were recomputed from WCAG relative luminance rather than adjusted by eye, which caught more flips than the `series-3` one you named: in the light set slots 1, 2 and 7 also cross the threshold, and in the dark set slots 1, 2, 3, 5 and 7 do. The card note records the ladder, the interleave and the 7-to-8 pair below. The one old hex still in the file is `#2DD4BF` at line 129, which is the selection state token rather than a series colour, and is unchanged by this work.

## Contrast slots 7 and 8 — documented, not reordered

Your 1.29 is right; I recomputed the whole ladder and got your table exactly. Leaving the order alone and recording it instead.

The palette still meets its stated contract: the two members are 0.0805 apart in relative luminance, above the 0.05 floor the design sets. The 1.29 is a consequence of the second rule rather than a breach of the first — the order is interleaved so a two or three series chart draws from opposite ends, and that is the case the handoff optimised for. Moving a member to widen the 7-to-8 pair narrows the low-cardinality case the ordering exists to serve, which is a trade the design made deliberately and not one to reverse from a review. The design doc now states the numbers, the tension between the two rules, and which one wins, so the next reader gets the reasoning rather than re-deriving it.

Happy to swap slot 7 if you would rather have the ratio than the interleave.

## getPaletteIndex — you are right, and it is retracted

I verified this independently rather than taking it: a false reversed match needs the forward loop to score at least one and the reverse loop to add at least one, both starting at index 0 against the same colour, so it requires a palette whose first colour equals its last. Zero of the fourteen in `defaults.css` do, `Contrast` included.

So the counter is sloppy and unreachable, and the note in the PR body was wrong. Rather than only correcting it here, the design doc now records the disproof — the earlier claim would otherwise outlive this thread and someone would open the ticket anyway.

## Checks

`claude-review` is still the same infrastructure failure with no model call, so there is nothing in it to act on from this side.

FigurePaper (commented):
# Code review, round 3 — chart palette re-tune, Contrast, per-chart picker scoping

Both round-2 items are addressed, the last round-1 question is answered, and `build` is green. **Verdict: approve.** Nothing below blocks — the two new findings are a one-line UI nicety and a documentation-page contrast issue.

## Round-2 items — verified

**The swatch page is brought up to date, and correctly.** `visualization-palette-swatches.html` now declares the new `--series-1..8` and `--series-dark-1..8`, prints the new hexes as chip labels, and adds `--contrast-1..8` with its own card. I re-derived the Contrast card's numeric claims rather than taking them: sorted relative luminances are `0.0545, 0.1071, 0.1759, 0.2271, 0.3076, 0.4095, 0.4994, 0.6606`, smallest consecutive gap `0.0512`. "Spanning 0.05 to 0.66, no two closer than 0.05" is exact, and the 7-to-8 ratio of 1.29 recorded in the note matches. The only pre-re-tune hex left in the file is `#2DD4BF` at `:129`, which is the selection token and not a series colour, as you said.

**The catch has a test.** `anUnresolvableMarkStillReturnsEveryPalette` reaches the fallback through the null service the fixture already wires, asserts the list comes back at full length, and asserts the nine flags still come from the gate. That is the branch round 1 asked for and round 2 noted was uncovered.

**The four-copies note landed** at `palette-test-fixtures.ts:20-23`, next to two of the unpinned copies, naming all four sites and which pair is pinned.

**Contrast 7/8 and the `getPaletteIndex` retraction** are both recorded in the design doc. On the retraction I re-derived the disproof independently and reach the same place: registering a match needs `cnt == pt.colors.length`, and the reverse loop's first comparison is `pt.colors[len-1]` against the same `curr[0]` the forward loop already used, so a non-zero forward count forces a zero reverse contribution unless a palette's first colour equals its last. None of the fourteen does. On 7/8 the trade is yours to make, and the doc now states both rules and which one wins, which is what the next reviewer needs.

**Decision 3 names all four callers** and which one supplies assembly context, including the two hosts that throw upstream of the request.

## Verified independently this round

- **Targeted Java suite, all green on this head**: `VSChartBindingColorPalettesTest` 4, `VSChartPaletteDefaultsTest` 25, `ColorPalettesModernTest` 7, `VSChartPaletteCssOverrideTest` 3, `VizContextReadFlipTest` 10, `ChartColorPaletteControllerTest` 6, `VGraphPairModernPaletteTest` 2 — 57 tests, 0 failures, 0 errors.
- **Frontend**: `palette-dialog.spec.ts` 9/9; `chart-palette.service.spec.ts` + `color-field-pane.spec.ts` + `combined-color-pane.spec.ts` 11/11.
- **The picker actually changes anything.** `ColorFrameModelFactory.CategoricalColorFactory.updateVisualFrameWrapper0` writes a chosen palette through `setUserColor`, not `setDefaultColor`, and user colours outrank defaults in `CategoricalColorFrame.getColor`. So `VGraphPair:1300`'s unconditional `applyModernPalette` on every render of a modern-marked chart cannot clobber a palette the user picked. Worth stating explicitly: had it written defaults, the whole picker-scoping feature would be inert for exactly the charts it targets.
- **`Contrast` splices like `Heat 8`.** `saveChanges()` splices `displayPalette.colors.length` = 8 and concatenates the chart's tail back, giving 40. Eight-slot palettes were already a shape the dialog handles.
- **`filesCallingOfGate()` sorts** before the two-element `assertEquals`, so the new survivor assertion is not walk-order dependent.
- **No cross-repo impact.** No enterprise file and no `e2e/` file references the old head hexes or the palette endpoint. Community-only PR, which matches the PR rules.
- **`VizContext.of((VizMark) null)`** yields `modern == false`, so an unmarked chart gets an empty hidden set — the `hiddenNamesAreEmptyForAClassicChart` path.

## Minor

### 1. A hidden palette is gone for good once the user clicks away from it

`paletteSelectOptions` filters on the *live* `_selectedIndex`, which `[(ngModel)]` rewrites the moment the user picks a different entry. So the escape hatch closes behind them: a modern chart sitting on `Pastel` shows `Pastel` in its dropdown, and as soon as the user selects `Modern` to look at it, `Pastel` disappears with no way back short of cancelling the dialog.

I confirmed this rather than reasoning it — a throwaway spec on this head, seeded exactly like `dialogWithHidden`, then setting `_selectedIndex` the way `ngModel` would:

```
expect(dialog.paletteSelectOptions.map(o => o.label)).toEqual(["Default", "Modern", "Heat 8"]);
dialog._selectedIndex = 1;
expect(dialog.paletteSelectOptions.map(o => o.label)).toEqual(["Default", "Modern"]);
```

Both assertions pass. Cancel restores the chart, so nothing is lost, which is why this is minor rather than a should-fix. But a list mutating under the cursor is not quite what "never taken away from a chart already using it" implies, and the fix is to pin the index once at first resolution instead of reading the live one — a field written alongside `_selectedIndex` in `displayPalette`, with the filter comparing against that.

### 2. The swatch page's chip digits pick the lower-contrast label on seven chips

The chip label colours went in on a luminance threshold somewhere near 0.45 rather than on a max-contrast pick. Computing both options for every new chip against the file's own `--light-chart-label: #1C1B1F`:

| chip | hex | white | `#1C1B1F` | chosen |
|---|---|---|---|---|
| series-1 | `#0490FF` | 3.26 | **5.25** | white |
| series-2 | `#FF5A35` | 3.10 | **5.52** | white |
| series-dark-1 | `#4FA5FF` | 2.57 | **6.65** | white |
| series-dark-2 | `#FF8367` | 2.42 | **7.09** | white |
| series-dark-7 | `#FE3290` | 3.46 | **4.95** | white |
| contrast-7 | `#00947F` | 3.79 | **4.52** | white |
| contrast-8 | `#5FA83C` | 2.94 | **5.83** | white |

Three of those digits land below 3:1. `contrast-8` at 2.94 is the one that stings, since it sits on the card for the palette whose whole contract is legibility. Picking whichever of the two inks has the higher ratio fixes all seven, and is what "recomputed from WCAG relative luminance" would produce. Documentation page only, so no rendered chart changes.

Relatedly, the flip list in your comment does not match the file: it names light 1, 2, 7 and dark 1, 2, 3, 5, 7, while the diff flips light 1, 2, 3, 4, 6, 8 and dark 1, 2, 3, 4, 6, 7. Only worth mentioning because it suggests the recomputation and the edit may have come from different passes.

## Round-1 carry-over — now confirmed

You ran it: a modern dashboard imported from before this branch, opened on this head, still pre-selects `Modern` / `Modern Dark` in the palette dialog. That closes the last open question from round 1.

It was worth checking because the persisted path looked like it could go the other way. `ChartVSAssemblyInfo.seedColorPalette` writes `seedPalette(ctx)` into the frame's *defaults* only at creation, Modernize and Revert — never on load — and the binding model reports `colors[i] = wrapper.getColor(i)`, resolving user, then CSS, then default. Had the frame reaching `createModel` still held the pre-re-tune defaults, `getPaletteIndex()` would have matched nothing, the dialog would have shown `Default`, and OK would have repainted the chart with the legacy 40 — the exact failure the `hidden`-flag design avoids for hidden palettes, arriving by a different route. The observed behaviour says the live frame is refreshed before the binding model is built, so the mechanism question I stopped at (`VSChartInfo chartInfo = info.getVSChartInfo()`, `VGraphPair:1197`) is academic; the user-visible outcome is what was being asked about, and it is right.

Worth a line in the design doc so the next re-tune inherits this rather than re-deriving it. Not gating.

## Notes, no action

- The dated plan documents under `docs/superpowers/plans/` still carry the old hexes (`2026-07-27-visualization-phase9b-dark-mode.md:40`, `2026-07-30-modern-categorical-palette.md:22-23`, and their code excerpts). Leaving them is right — they record what was true then, and the phase9b doc defers to the swatch page for canonical values, which is now current. Nothing in the living `specs/lookfeel/` tree points at a stale head.
- The pre-existing `createAssetEntry(undefined).organization` throw in `visual-dropdown-pane` and the date-comparison pane is named in Decision 3 but has no ticket of its own. Whenever someone fixes it, threading `vsId`/`assemblyName` into the date-comparison pane becomes worth doing in the same change, since that host does have a chart in scope.

## Checks

- **`build` — pass**, 12m45s, on the current head. The round-1 red is gone and has stayed gone, which supports the flaky-race reading.
- **Semgrep OSS, Semgrep Scan, license/cla — pass.**
- **`claude-review` — fail, and it is not a code signal.** I read the job log directly rather than relying on the earlier rounds: `anthropic_api_key: ""` and `ANTHROPIC_API_KEY:` empty in the action inputs, then `is_error: true`, `duration_ms: 398`, `num_turns: 1`, `total_cost_usd: 0`. It never made a model call, so it has looked at none of this diff. Discounted as infrastructure per the repo owner's direction; it needs credentials in the workflow before it can contribute anything.
- `mergeable: MERGEABLE` — no conflicts. `BLOCKED` reflects the required-review gate only.

Approving. Items 1 and 2 are worth a follow-up commit if you want them, but neither should hold the merge.

jshobe-inetsoft (approved):
# Code review — chart palette re-tune, Contrast, per-chart picker scoping

Independent manual review (the `claude-review` check on this PR failed due to a CI infrastructure/credentials error — `anthropic_api_key` empty, `is_error: true`, 1 turn, $0 cost — it never actually invoked a model against this diff, so it carries no signal and is disregarded here per repo direction).

## Scope reviewed

- `VSChartPaletteDefaults.java` — re-tuned `MODERN_HEAD`/`DARK_HEAD` constants, new `hiddenPaletteNames(VizContext)`.
- `defaults.css` — matching re-tune of `Modern`/`Modern Dark` indexes 1-8, new `Contrast` (8 slots, no dark variant).
- `VSChartBindingController` / `VSChartBindingService` — new `vsId`/`assemblyName`-scoped `getColorPalettes`, `getChartVizMark` cluster-proxy method, `pickerContext()` fallback.
- `CategoricalColorModel` — new `hidden` flag, serialized to the frontend.
- Frontend: `categorical-color-pane.component.ts` (forwards `vsId`/`assemblyName`), `palette-dialog.component.ts` (filters hidden entries but always keeps the chart's current selection visible).
- Test suite: `VSChartPaletteDefaultsTest`, `ColorPalettesModernTest`, `VSChartPaletteCssOverrideTest`, `VizContextReadFlipTest`, `VSChartBindingColorPalettesTest` (new), `ChartColorPaletteControllerTest`, `VGraphPairModernPaletteTest`, `CategoricalColorFactoryResolvedDefaultGuardTest`, plus the new `palette-dialog.spec.ts` cases.

## Verification performed

- Confirmed `MODERN_HIDDEN` (`Pastel`, `Heat 8/16/24`, `Blue`, `Green`, `Red`, `Orange`, `Gray`) matches exactly the palette names declared in `defaults.css` besides `Default`, `Soft`, `Modern`, `Modern Dark`, `Contrast` — nothing is silently dropped from resolution, only flagged.
- Confirmed `getColorPalettes()` still returns every palette name from `ColorPalettes.getPaletteNames()` unfiltered, with only `hidden` toggled — matches the PR's stated invariant ("flags palettes, never omits them"), which is important because the dialog matches a chart's current palette by color-list identity.
- Confirmed `VSChartBindingService.getChartVizMark` reads `ChartVSAssemblyInfo`/`VSAssemblyInfo.getVizMark()`, which is a real, existing accessor on the epic branch, and that `pickerContext()` in the controller degrades to `VizContext.ofGate()` both when `vsId`/`assemblyName` are absent and when the cluster-proxy call throws (e.g., stale runtime) — verified via the new `VSChartBindingColorPalettesTest.anUnresolvableMarkStillReturnsEveryPalette`, which exercises the throwing path with a null service and asserts the list is never shortened.
- Confirmed `VizContextReadFlipTest.filesCallingOfGate()` walks the full `core/src/main/java/inetsoft` + `utils/*` source tree (not just one package) and now asserts exactly two named survivors (`ChartColorPaletteController`, `VSChartBindingController`) — this is a real tree-wide regression guard, not a narrow spot-check, so a future accidental new org-gate read will fail the build.
- Confirmed the two other hosts that render `categorical-color-pane` without `vsId`/`assemblyName` (`date-comparison-pane`, `visual-dropdown-pane`) already throw upstream in `ngOnInit` on `createAssetEntry(undefined).organization` before any request is built, on `epic-74519` as well as on this branch — so this PR does not introduce a new gap there; it's a pre-existing, separately-notable defect, correctly left out of scope per the PR's own "not in scope" framing and this repo's minimal-scope bug-fix guidance.
- Confirmed the palette-picker `hidden` filter in `palette-dialog.component.ts` keeps the chart's own currently-selected palette visible in the dropdown even when it's on the hidden list (`index === selected`), matching the "never disappears out from under a chart already using it" contract — verified against `palette-dialog.spec.ts`'s new hidden-palette test block, including the reversed-match case.
- Cross-checked that Java (`VSChartPaletteDefaults` MODERN_HEAD/DARK_HEAD), CSS (`defaults.css`), and the TS test fixtures (`palette-test-fixtures.ts`) all carry the same new hex values, and that the new `cssHeadMatchesTheJavaFallback` reflection-based test pins the CSS/Java pair against future drift.
- Checked for compile-correctness of the new Java (imports, `@Autowired` constructor shape, `@ClusterProxyMethod` usage) — consistent with existing patterns in the same files; the `build` check is green, which is external confirmation.
- No security concerns: no new user input reaches SQL/script/file-path construction; `vsId`/`assemblyName` flow through the existing `ViewsheetService.getViewsheet(id, principal)` authorization path unchanged.
- No backward-incompatibility: `CategoricalColorModel.hidden` is an additive, defaulted-`false` field; existing persisted viewsheets/palettes are unaffected since palette resolution (`getPaletteNames()`/`getPalette()`) stays unfiltered.

## Not blocking, already tracked

Two minor items were already identified and accepted as non-blocking follow-ups during this PR's own review history (both are cosmetic/documentation, not functional regressions):
1. A hidden palette disappears from the dropdown once the user switches away from it inside the same dialog session (cancel restores it — nothing persisted is lost).
2. Seven chip labels in the `docs/superpowers/specs/lookfeel/visualization-palette-swatches.html` reference page pick the lower-contrast of the two label-ink options (documentation page only, not a rendered product surface).

Neither affects product correctness, data integrity, or test coverage.

## Checks

- `build` — pass.
- `Semgrep Scan`, `Semgrep OSS`, `license/cla` — pass.
- `claude-review` — fail, discounted as CI infrastructure/quota error (no model invocation occurred), per instructions to substitute this manual review in its place.
- `mergeable: MERGEABLE` — no merge conflicts. `mergeStateStatus: BLOCKED` reflects the required-review gate only.

## Verdict: Approve

No blocking issues found. Test coverage is thorough and specifically targets the invariants this PR depends on (flag-not-omit, tree-wide org-gate survivor guard, CSS/Java drift guard, fallback-on-failure). Approving without enabling auto-merge, per process.

PR #5189, opened 2026-09-12
Fix #76616, worksheet variables shown out of order in Enter Parameters dialog

larryliang-inetsoft:
## Summary
- Redmine #76616: the runtime "Enter Parameters" dialog for a worksheet could show variables in the wrong order (e.g. `var2` before `var1`) even though `var1` was declared/used first.
- Root cause: `XQuery.varmap` (which tracks the variables discovered in a query, e.g. by scanning a SQL WHERE clause for `$(name)` tokens) was a `ConcurrentHashMap`. Its iteration order is hash-bucket based, not insertion order, so the order variables were declared/found in the query was not preserved when building the list shown to the user.
- Fix: switch `varmap` to a `LinkedHashMap` (wrapped in `Collections.synchronizedMap` to keep thread safety) so variable order always reflects the order they were declared/discovered, for any variable names — not just this reported var1/var2 pair.

## Test plan
- Added `core/src/test/java/inetsoft/uql/jdbc/JDBCQueryVariableOrderTest.java`, which builds a `JDBCQuery` over a raw SQL string containing `$(var1)`/`$(var2)` (matching the reported worksheet's `ORDER_DATE BETWEEN $(var1) AND $(var2)` condition) and asserts `getVariableNames()` returns them in declaration order. Repeated 5x, plus a 5-variable case, to guard against a fix that only happens to work for one hash arrangement.
- Verified the new test fails with the exact reported symptom (`[var2, var1]` instead of `[var1, var2]`) when the fix is reverted, and passes with the fix applied.
- Ran the full `inetsoft.uql.jdbc.**` and `inetsoft.uql.asset.**` test packages in the `core` module after the change — all pass, no regressions.
- Reproduction was source/unit-test level only; the attached worksheet asset (var1/var2 with a BETWEEN condition) was inspected but not live-verified in a running StyleBI UI (no running instance was available in this session).


larryliang-inetsoft:
Independent review complete (full findings: docs/teams/2026-09-13-bugs-batch6/bug-76616/01-review.md in stylebi-enterprise). **Verdict: merge-ready, no blocking issues** — root cause fix (ConcurrentHashMap→LinkedHashMap ordering), a self-caught thread-safety regression already patched within the PR, all call sites/subclasses checked, UI-side trace confirms it resolves the reported symptom.

Note: could not post this as a formal GitHub approval — GitHub blocks self-approval and this session's only authenticated account (larryliang-inetsoft) is also the PR author. Flagging for a second reviewer or admin merge.

Separately: the `claude-review` required check failed, but its log shows the job errored before doing any review (`is_error:true`, 0 turns, empty `ANTHROPIC_API_KEY` in its env) — looks like a broken CI secret, not a real finding against this diff.

fredaan (approved):

PR #5191, opened 2026-09-13
Bug #76614: fix intermittent NPE from concurrent Data Tip requests racing calc table temp crosstabs

larryliang-inetsoft:
## Summary

Fixes Redmine #76614: intermittent `NullPointerException` in `DataVSAQuery.getSourceTable()` on Data Tip click.

Root cause: `CalcTableVSAQuery.createCrosstabAssemblies0()` built each temp crosstab assembly with a **deterministic** name (`TEMP_ASSEMBLY_PREFIX + calcTableName + "_Crosstab_" + ct`). `getTableLens()` holds the sandbox write lock for its whole body, but the nested base-table fetch briefly drops the entire lock stack (`box.unlockAll()`, added for bug #74001) and restoring the lock afterward can't undo a mutation another thread made to the shared `Viewsheet`'s assembly map during that window. Two genuinely concurrent top-level Data Tip requests for the same viewsheet are architecturally reachable (nothing serializes requests per `runtimeId` in the STOMP `/datatip/open` dispatch path), so a second thread's own add/remove cycle for the same calc table — or its `removeTempAssembly()` failure-path cleanup sweep — can complete entirely inside that window and remove the first thread's still-in-use temp crosstab. When the first thread resumes and the crosstab has an aggregate-level Calculated Field, `getFormula()` → `getSourceTable()` → `getAssembly()` returns null → NPE.

- Made each `getTableLens()` invocation use a unique invocation id (a per-JVM `AtomicLong` counter) as part of its temp crosstab assembly names, so two concurrent invocations for the same calc table can never collide on the same assembly name.
- `removeTempAssembly()`'s cleanup sweep is scoped by the same invocation id, so it only ever removes its own invocation's temp crosstabs, never a different concurrent invocation's.
- Verified every other reference to the naming scheme in the codebase (`VSScriptableService`, `VSObjectPropertyService`, `VSAScriptable`, `AbstractCrosstabVSAQuery`) only checks the assembly-name **prefix**; one additional consumer (`VSAQuery`'s `TEMP_CALC_CROSSTAB_PATTERN`) parses the full suffix but assigns the result to a local variable that is never read afterward (dead code) — either way the new format still satisfies that regex.

## Test plan

- [x] Added `CalcTableVSAQueryTempCrosstabNameTest` (4 tests): pre-fix collision demonstration, fixed-scheme non-collision, a 32-thread/200-invocation/3-crosstab concurrency stress test asserting zero name collisions, and a test that `removeTempAssembly`'s sweep pattern never touches another invocation's assemblies.
- [x] `./mvnw -pl core test -Dtest=CalcTableVSAQueryTempCrosstabNameTest` — 4/4 pass.
- [x] `./mvnw -pl core test -Dtest=CrosstabVSAQueryTest,TableVSAQueryTest,AssetQueryTest,PostProcessorTest,ChartVSAQueryTest` (existing related VSAQuery/crosstab tests) — all pass, no regressions.
- [ ] **Not verified by a live/concurrency reproduction of the actual reported NPE stack.** This is a structural fix with test coverage of the name-generation logic only — standing up a full `ViewsheetSandbox`/`RuntimeViewsheet` harness with two threads racing through the real `box.unlockAll()`/`restoreLocks()` window would require a latched fake `AssetDataCache` or an instrumented lock-drop point, which was judged disproportionate to proving this fix (see `docs/teams/2026-09-13-bugs-batch6/bug-76614/` in the enterprise repo for the full diagnosis/refutation/fix writeup).


larryliang-inetsoft:
## Review round 1 (review-only; not a formal approval — same self-approval restriction as PR #5189)

Head commit reviewed: `3680b800bb`.

**Checked:**
- Pulled the diff myself (`gh pr diff 5191`) rather than trusting the fix writeup. The naming change (per-invocation `AtomicLong` id threaded into `getTempCrosstabName()`/`removeTempAssembly()`) targets the confirmed root cause directly — no overfit smell (no literal-from-repro keying, no scope-narrowing `if`, general rule change).
- Re-derived independently: `VSAQuery`'s `TEMP_CALC_CROSSTAB_PATTERN` match result (`vassembly = matcher.group(1)`, line ~1057) is genuinely dead code — grepped the whole file, that local is never read again. New name format still satisfies the regex (`_Crosstab_[\d]+$` suffix unchanged), so no behavior change either way.
- Read `CalcTableVSAQueryTempCrosstabNameTest.java` in full: the two collision tests call the real `nextInvocationId()`/`getTempCrosstabName()` production methods (not a reimplementation) under real thread contention. Checked out the PR worktree at the head commit and re-ran the suite myself: `Tests run: 4, Failures: 0, Errors: 0, Skipped: 0`.
- Grepped the repo for `TEMP_ASSEMBLY_PREFIX`/`_Crosstab_`: every other consumer only does prefix/contains checks, none parse the exact suffix, no user-visible log message references the temp crosstab's generated name — no regression there.
- PR description accurately states this is a structural fix, not a live reproduction of the reported NPE — matches the diagnosis/refute docs' own claim boundary, no overclaiming.

**One non-blocking finding (sibling check, not a defect in this diff):** the writeup frames `AbstractSelectionVSAQuery.checkAndRunCalcFieldMeasureQuery`'s own `TEMP_ASSEMBLY_PREFIX + name + "_Crosstab"` pattern as a "different, safe" path since it has no numeric suffix. Tracing it independently, I don't think that's substantiated — it clones the viewsheet and calls `box.setViewsheet(vs, false)`, which overwrites `ViewsheetSandbox.vs`, a single unsynchronized mutable field (`VSAQuery.getViewsheet()` is a fresh `box.getViewsheet()` lookup, not captured per-instance). Its own nested `CrosstabVSAQuery.getTableLens()` call goes through the same `getDataWithoutSandboxLock()` lock-drop window. Two overlapping top-level calls into the same box could clobber each other's active viewsheet reference through that window — a different, arguably broader, unaddressed hazard than the assembly-name collision this PR fixes. Not something this PR needs to fix (different mechanism, would violate minimal-scope), but recommend filing as a separate follow-up investigation.

**Verdict: merge-ready.**

fredaan (approved):

larryliang-inetsoft:
Independent correctness review (gh identity here is the PR author, so GitHub blocks a self-approval review — noting findings as a comment instead; fredaan's existing APPROVED review satisfies review requirements).

Verified against the diff:
- `invocationId` is allocated once per `getTableLens()` invocation and threaded consistently through `createCrosstabAssemblies`/`createCrosstabAssemblies0` and both `removeTempAssembly` call sites (success-path null-crosstabs case and the exception-handler cleanup case) — no stale call sites left on the old 2-arg signature.
- Searched the whole repo for every other consumer of `CalcTableVSAQuery.TEMP_ASSEMBLY_PREFIX`: `VSScriptableService`, `VSObjectPropertyService`, `VSAScriptable`, `AbstractCrosstabVSAQuery` all only do a prefix check (`startsWith`/`contains`), so the new invocation-id segment doesn't affect them.
- Confirmed `VSAQuery.TEMP_CALC_CROSSTAB_PATTERN` (the one consumer parsing the full suffix) still matches the new format, and independently verified the PR's claim that the local var it assigns (`vassembly`) is never read afterward in that method — genuinely dead code, not a live dependency.
- New test file uses JUnit 5 (`org.junit.jupiter`), consistent with sibling tests in the same package; its package-private access to `nextInvocationId()`/`getTempCrosstabName()` is valid.
- Diff is scoped to the fix + a regression test only — no unrelated refactors, no dead code/unused imports, and the added comments explain non-obvious cross-file concurrency reasoning rather than restating code.

No blocking issues found. `build`, both Semgrep checks, and CLA are green. The failing `claude-review` check is the known infra bug (errors out at 0 cost before reviewing) and is disregarded per instructions. Squash-merging.

PR #5192, opened 2026-09-13
Add diagnostic trace logging for #76488/#76485 investigation (no behavior change)

larryliang-inetsoft:
## Summary
- Adds DEBUG-level log lines, all carrying the marker string `BUG76488-76485-TRACE`, at four points in the binding-tree/refresh call chain suspected in the still-unreproduced "spinner appears then sticks" reports for bugs 76488 and 76485.
- Touches: `VSTreeHandler.getTableTreeModel()` (assembly/`cass` resolution), `VSBindingTreeService.getBinding()` (read-lock acquire/release window), `ModifyCalculateFieldService.modifyCalculateField()` (in-process `getBinding` call ordering + command dispatch), and `VSRefreshService.refreshViewsheet()` (write-lock acquire/release window, `resetRuntime()` call).
- **This is diagnostic instrumentation only. It does not fix, and does not claim to fix, either bug.** All new lines are at `LOG.debug(...)` level, so they are silent during normal `INFO`-level server operation and add no behavior change — same control flow, same return values, same exception handling as before.
- Purpose: let a future live-repro attempt (or the customer's own logs, if this ships) `grep` `docker logs` for `BUG76488-76485-TRACE` to observe the suspected async read/write-lock race in the act, instead of only reasoning about it statically from source.

Related to bug 76488 / 76485.

## Test plan
- [x] `./mvnw -pl core -am compile -DskipTests -q` — compiles clean, no errors, no unused-import warnings.
- [x] `./mvnw -pl core test -Dtest=VSTreeHandlerTest -q` — passes (only existing unit test among the 4 touched classes).
- [ ] Full test suite not run (not required for a pure-logging change).


fredaan (approved):
## Code Review — Diagnostic trace logging for #76488/#76485

### Summary
Reviewed the full diff across all 4 touched files (`ModifyCalculateFieldService`, `VSTreeHandler`, `VSBindingTreeService`, `VSRefreshService`). This is a pure, additive diagnostic-logging change (+45/-0 across all files) intended to help reproduce an intermittent "spinner stuck" race condition (bugs #76488/#76485).

### Correctness
- **No control-flow changes.** Every diff hunk is a pure insertion; no existing statement, branch, lock acquire/release ordering, or return value was touched.
- **Lock placement verified**: the new `LOG.debug(...)` calls in `VSBindingTreeService.getBinding()` and `VSRefreshService.refreshViewsheet()` sit immediately after `lockRead()`/`lockWrite()` and inside the existing `finally` blocks after `unlockRead()`/`unlockWrite()` — they don't change when the lock is held.
- **Format-string/argument counts checked** for all 8 new `LOG.debug` calls — every `{}` placeholder has a matching argument, in order (e.g. `VSTreeHandler` 5-placeholder/5-arg and 3/3, `VSBindingTreeService` 5/5, 3/3, 4/4, `VSRefreshService` 3/3, 2/2, 2/2, `ModifyCalculateFieldService` 5/5, 3/3×3).
- **`id`/`rvs`/`assembly` are all in scope** at each new log call site (`id` is the `@ClusterProxyKey` parameter of `modifyCalculateField`; `rvs.getID()` and `assembly`/`info` were already resolved on prior lines).
- Null-guards are already correct where referenced (`info == null ? "null" : info.getClass().getSimpleName()`), so no new NPE risk.

- All logging is at `LOG.debug(...)`, so it is silent under the project's normal `INFO` level — no behavior, output, or API change for existing deployments.
- SLF4J's parameterized `{}` form is used throughout, so there's no eager string-building cost even when DEBUG is disabled.

### Security
- No sensitive data (credentials, query results, user PII) is logged — only assembly names, thread names, runtime/session ids, and booleans, all already visible elsewhere in existing logs.

### Tests
- No test files touched, no test behavior affected. PR author ran `VSTreeHandlerTest` (the only existing unit test among the 4 touched classes) — passes per PR description.

### Minor, non-blocking notes
- The `"BUG76488-76485-TRACE"` marker string is duplicated 8 times as a literal; a shared constant would make the eventual cleanup (grep/remove) a one-line diff instead of 8, but this matches the existing project convention for this kind of temporary trace logging (see prior commits like `6b292bd02`, `da21b3062`), so not requesting a change.
- One message in `VSTreeHandler` contains a Unicode em dash (`—`); harmless since `project.build.sourceEncoding` is UTF-8, just flagging for awareness.

### CI / mergeability
- Validation Build, Semgrep, Semgrep OSS, and CLA checks all pass.
- The `claude-review` check failed, but its log shows this is an automated-action infra issue (empty `ANTHROPIC_API_KEY` in the job env), not a code finding — excluded per review policy, and this manual review takes its place.

### Verdict
**No blocking issues found.** Approving.

fredaan (approved):

PR #5193, opened 2026-09-13
Fix #76593, arrange dashboards disables unrelated dashboard checkbox

larryliang-inetsoft:
## Summary
- Redmine #76593: in the "Arrange Dashboards" dialog, disabling one dashboard would visually (but not actually) disable an unrelated dashboard's checkbox.
- Root cause: the dashboard rows were rendered with `@for (... track $index ...)`. `enabledChanged()` reorders `model.dashboards` in place via `splice()` to keep enabled dashboards grouped before disabled ones. Because rows were tracked by index, Angular reused the same embedded view (and its `NgModel` directive instance) across the reorder. `NgModel`'s own dirty-check compares against its cached prior value for that view slot, so when the dashboard that ends up in a row after the reorder happened to share the previous occupant's `.enabled` value, the checkbox's DOM `.checked` was never rewritten - leaving that row visually wrong even though the underlying `.enabled` model value was untouched.
- Fix: track by `dashboard.name` instead of `$index`. The backend (`ArrangeDashboardDialogController.getDashboards()`) builds the list from a `LinkedHashSet<String>`, so `name` is guaranteed unique per row - DOM nodes now move with their data instead of being reused positionally.
- Added a regression test asserting the actual rendered checkbox `.checked` DOM state (not just row/label order, which the existing #18799 regression test already covered and would not catch this bug).

## Test plan
- [x] Added `arrange-dashboard-dialog.spec.ts` test: 3 enabled dashboards, disable dashboard at row 0, assert post-reorder rows are `[dash2, dash3, dash1]` with checkbox `.checked` states `[true, true, false]`.
- [x] Confirmed the new test fails on pre-fix code (reverted the html change locally, reran) with `checkboxes[0].checked` incorrectly `false`, matching the reported symptom.
- [x] Confirmed the new test and the existing #18799 regression test both pass with the fix restored: `npx ng test portal --watch=false --include="projects/portal/src/app/portal/dialog/arrange-dashboard-dialog.spec.ts"` → 2 passed.

Redmine: #76593


larryliang-inetsoft:
## Review r1 — merge-ready

Pulled the diff directly (`gh pr diff 5193`) rather than trusting the fix doc — matches exactly: one-line `track $index` → `track dashboard.name` in `arrange-dashboard-dialog.component.html`, plus one new regression test in `arrange-dashboard-dialog.spec.ts`.

**Re-derived uniqueness/non-null claim from `ArrangeDashboardDialogController.getDashboards()` directly**: the dashboard list is built from a `LinkedHashSet<String>` (structurally unique), and every entry unconditionally gets `name` set from that set element (never null in practice — this `name` already round-trips through `getSelectedDashboardNames()`/`setDashboards()` elsewhere in the same controller as the feature's identity key). Confirmed the `__GLOBAL` suffix-stripping only touches `label`, not `name`.

**Read the new test directly**: it asserts real `checkbox.checked` DOM state for all 3 rows (not just label order like the old #18799 test), correctly settles past the unrelated `NgModel` first-`writeValue` microtask deferral the refuter flagged, and targets exactly the row that breaks. Not overfit to the repro — the assertion would fail against any variant of the fix that didn't actually re-associate the checkbox directive with the right dashboard object.

**Sibling-invariant check**: checked 4 of the 13 candidate files the fixer flagged as unconfirmed (`color-mapping-dialog`, `viewsheet-device-layout-dialog`, `tabular-list-editor`, `add-parameter-dialog`) for whether toggling their per-row checkbox actually triggers an in-place splice reorder the way `enabledChanged()` does. None do — in each case the checkbox's own handler never splices, and the file's only `splice()` call is bound to a separate delete/remove button. No action needed on this PR from that check. (One unrelated, lower-confidence latent-bug shape noted for `tabular-list-editor` in the review doc — a removal-triggered, not toggle-triggered, stale-checkbox risk — flagged for awareness only, not a blocker here.)

**Downstream check**: `moveDashboard()` (up/down arrows) also swaps array elements in place via `splice()`, and previously had the same latent staleness bug under `track $index`. The fix incidentally corrects that too — a net improvement, not a regression. Backend save path keys off `name`, not array position, so unaffected.

No blockers. Full findings: `docs/teams/2026-09-13-bugs-batch6/bug-76593/04-review-r1.md`. Posting as a plain comment, not a formal approval — GitHub blocks self-approval since this session's account is also the PR author.

fredaan (approved):

PR #5197, opened 2026-09-14
Bug #76617: fix physicalTable resolution silently failing for any valid datasource

NeverStopFMH:
## Summary

`create_viewsheet(type:"physicalTable", ...)` / `attach_base_worksheet(type:"physicalTable", ...)`
were rejecting real, existing physical tables with "no physical table named '<table>' was found
in data source '<datasource>', or you lack permission to read it" — live-confirmed against
`Examples/Orders`/`CATEGORIES` (Redmine #76617).

## Root cause

`ViewsheetAssemblyAgentController.resolvePhysicalTableEntry` hand-builds the root `DATA_SOURCE`
`AssetEntry` it passes into `AssetRepository.getEntries()`, but never calls
`setProperty("prefix", ...)` on it. The real `AbstractAssetEngine` implementation's
`getQueryEntries()`/`getPhysicalEntries()` read `entry.getProperty("prefix")` to resolve the
`XDataModel`/`XDataSource`; with it null, both silently resolve to nothing and return an empty
`AssetEntry[]` rather than throwing — `findPhysicalTableEntry`'s recursion therefore always comes
up empty, indistinguishable from "table genuinely doesn't exist."

This is not specific to datasource names containing "/" (the originally suspected cause) — it
reproduces for any datasource, since it's the missing property read, not the path shape, that's
at fault.

## Fix

Set `root.setProperty("prefix", datasource)` (and `"source"`, matching
`AbstractAssetEngine`'s own convention when it constructs `DATA_SOURCE` entries) immediately
after `root` is constructed in `resolvePhysicalTableEntry`. Additive only — no other branch
(`worksheet`/`logicalModel`) of `resolveDataSourceEntry` is touched.

`attachBaseWorksheet(type:"physicalTable")` shares the exact same `resolveDataSourceEntry` →
`resolvePhysicalTableEntry` call path, so this fix covers both tools.

## Regression test

The existing `attachBaseWorksheetAcceptsPhysicalTableTypeFlat`/`...TypeNested` tests mock
`AssetRepository` with `any()` matchers on `getEntries()`, so they cannot distinguish a
prefix-set root from a prefix-unset one — zero regression coverage for this defect as originally
written. Augmented both with an assertion on the captured root/folder entry passed to
`getEntries()`, asserting `"Examples".equals(entry.getProperty("prefix"))`. Confirmed both new
assertions fail on the pre-fix code and pass after.

## Test plan

- [x] `./mvnw -o -pl core -Dtest=ViewsheetAssemblyAgentControllerTest test` — 91/91 passing with
      the fix applied
- [x] Manually reverted the one-line fix and re-ran the same test class — the two augmented
      assertions fail with `expected: <Examples> but was: <null>`, confirming they actually pin
      the defect
- [x] Re-applied the fix, re-ran — 91/91 passing again


NeverStopFMH:
## Round 1 review (reviewer-76617)

Head commit reviewed: d1bccec7fe6c589f39eb8d86b745fd52b008d473 (sole commit on this PR).

### Scope
Production diff: 2 lines in `resolvePhysicalTableEntry` (`ViewsheetAssemblyAgentController.java`) —
`root.setProperty("prefix", datasource)` / `root.setProperty("source", datasource)` right after the
hand-built `DATA_SOURCE` `AssetEntry` is constructed. Test diff: 23 lines augmenting the two existing
`attachBaseWorksheetAcceptsPhysicalTableType{Flat,Nested}` tests with an assertion on the captured
root entry's `"prefix"` property.

### Independent verification performed (read the real source myself, not just the team's docs)
- Confirmed `root.setProperty("prefix"/"source", datasource)` uses the method's `datasource`
  *parameter*, not any literal from the repro ("Examples/Orders"/"CATEGORIES" appear nowhere in the
  production diff) — the fix is generic.
- Confirmed no special-case `if`/condition was narrowed; the fix is two unconditional statements.
- Confirmed both `createViewsheet` (call site line 260) and `attachBaseWorksheet` (line 1835) route
  through the same shared `resolveDataSourceEntry` → `resolvePhysicalTableEntry`, so both tools'
  `type:"physicalTable"` mode benefit from this fix identically — traced the call sites myself.
- Confirmed the `worksheet` and `logicalModel` branches of `resolveDataSourceEntry` are untouched by
  this diff (read lines 1660-1712 directly). `logicalModel`'s own missing-property defect is
  correctly left to #76606's separate fix.
- Walked `AbstractAssetEngine.getPhysicalEntries` (lines 679-696) directly: once the root resolves,
  every child entry the engine itself constructs via `createPhysicalEntry` gets `sentry.setProperty
  ("prefix", source)` stamped on it by the engine (line 693) before being returned — so
  `findPhysicalTableEntry`'s recursion into child folders is self-sufficient and needs no additional
  fix; confirmed against the actual engine code, not just the docs' claim.
- Checked whether the added `"source"` property is safe: grepped all `getProperty("source")` reads
  in `core/src/main`; none apply to a `DATA_SOURCE`-typed entry (the `isDataSource()` branch in
  `getQueryEntries`/`getPhysicalEntries` reads only `"prefix"`, never `"source"`). Additionally, the
  hand-built `root` entry never escapes `resolvePhysicalTableEntry` (only its children, returned by
  the real engine, are ever returned to the caller) — so the added property has no possible
  downstream reader regardless. Harmless.
- Re-read the regression test assertions directly: `assertEquals("Examples", entry.getProperty
  ("prefix"))` checks against the same `datasource` string the test itself passes into
  `attachBaseWorksheet(...)` — since the production line reads `root.setProperty("prefix",
  datasource)` off the method parameter (confirmed by direct source read, not narrowed to any
  literal), the assertion is not vacuous against a narrower fix.
- Confirmed via `gh run view --log` that the failing `claude-review` check errored out due to a CI
  configuration issue (empty `ANTHROPIC_API_KEY`, "result is_error:true", 0 turns) before performing
  any actual review — not a content-based finding against this diff.

### Verdict
No blocker or important findings. Diff is minimal, additive, matches the confirmed root cause
exactly, consistent with `AbstractAssetEngine`'s own convention for `DATA_SOURCE` entries, and the
regression test is confirmed (by this seat, independently reading the test body) to actually pin the
defect rather than being decorative. Approving and merging.

PR #5198, opened 2026-09-14
Bug #76606: fix logicalModel path resolution in ViewsheetAssemblyAgentController

NeverStopFMH:
## Summary

`attach_base_worksheet`/`create_viewsheet(type:"logicalModel", path:...)` rejected every path
value for a real, existing logical model (e.g. `"no logical model named
'Examples/Orders/Order Model' was found, or you lack permission to read it"`), including the
objectively correct path. Live-confirmed via `create_viewsheet(type:"logicalModel", ...)`
(which shares the identical `resolveDataSourceEntry` method `attach_base_worksheet` calls).

## Root cause

`ViewsheetAssemblyAgentController.resolveDataSourceEntry`'s `logicalModel` branch hand-builds
an `AssetEntry` via the bare 5-arg constructor, which never populates the entry's property
map. `AbstractAssetEngine.getQueryEntries`'s `isLogicModel()` branch — the code the subsequent
`rep.getEntries()` probe actually dispatches into for a `LOGIC_MODEL` entry — resolves the data
model from `entry.getProperty("prefix")`/`entry.getProperty("source")`, not from the entry's
path. Both were always null on the hand-built entry, so `repository.getDataModel(null)` always
short-circuited to `null`, and the probe always returned an empty array — for any path value,
correct or not.

## Fix

Split the caller-supplied path on its last `/` (the datasource name itself may contain `/`,
e.g. `"Examples/Orders"`) and populate `prefix`/`source` on the hand-built entry the same way
`AbstractAssetEngine`'s own `isDataSource()` branch does when it builds a `LOGIC_MODEL` entry by
walking the tree — so the existence/permission probe actually resolves. Also refuse a path with
no `/` up front, naming the field, instead of letting it fail later with the generic not-found
message.

Considered (and rejected as oversized for this bug) injecting `DataSourceService` into
`ViewsheetAssemblyAgentController` to reuse `WorksheetAgentController.addLogicalModelTable`'s
pattern — `DataSourceService` isn't currently available in this controller at all, and wiring
it in would touch every existing construction site, including 13 call sites in
`ViewsheetAssemblyAgentControllerTest`'s helpers. The chosen fix is additive and scoped to the
`logicalModel` branch only.

## Test plan

- [x] New regression test asserts on the `AssetEntry` argument captured at the `getEntries()`
      call site (`prefix`/`source` actually populated), not just on a stubbed return value —
      confirmed to fail before the fix (`expected: <Examples/Orders> but was: <null>`) and pass
      after.
- [x] Full `ViewsheetAssemblyAgentControllerTest` suite passes: 92/92, 0 failures.

Fixes Redmine #76606.


NeverStopFMH:
## Round 1 review — reviewer-76606

Head commit reviewed: `3799d717d` (single commit on `bug-76606-logicalmodel-path`).

### What I independently checked

- Re-pulled the diff myself via `gh pr diff 5198` (not just read `04-fix.md`/`05-verify.md`'s
  paraphrases) — matches exactly: `ViewsheetAssemblyAgentController.java` (+23/-1) and
  `ViewsheetAssemblyAgentControllerTest.java` (+50/-0), nothing else touched.
- **Overfit checks (per `.claude/commands/review-workflow.md`):**
  - No literal from the repro is keyed into production code. The split is
    `trimmedPath.lastIndexOf('/')` — generic over any `"<ds>/<model>"` string. "Examples/Orders"
    only appears in a comment (documentation) and in the new test's input, never in a
    conditional.
  - No special-case `if` excludes the reported input; the new `PairingException` guard
    (`sepIdx <= 0 || sepIdx == trimmedPath.length() - 1`) is unconditional over any path shape
    (no `/`, leading `/`, trailing `/`).
  - Call-site vs. shared invariant: I fetched the post-fix file directly (not trusting
    `05-verify.md`'s claim) and confirmed `resolveDataSourceEntry` is called from both
    `createViewsheet` (line 260) and `attachBaseWorksheet` (line 1855) — one shared private
    method, so both callers get the fix. Confirmed myself via grep, not taken on faith.
  - Regression test assertion: read the actual new test method
    (`attachBaseWorksheetLogicalModelProbesEntryWithPrefixAndSourceProperties`) — it asserts
    `prefix=="Examples/Orders"` / `source=="Order Model"`, which are exactly the two halves of
    the same input path it passes in (`"Examples/Orders/Order Model"`), not independent
    hardcoded values. A narrower/still-broken fix (e.g. one that populated the properties with
    the whole path instead of the split halves) would fail this assertion. Good.
  - Siblings: read the full `resolveDataSourceEntry` method body directly (lines 1604-1734) —
    the `worksheet` branch and the `physicalTable` branch (which delegates to
    `resolvePhysicalTableEntry`, #76617's separate fix) are byte-for-byte untouched.

I tried to build a counter-example where this change fails and could not — the split logic,
guard, and property-population are all unconditional and match `AbstractAssetEngine`'s own
`isDataSource()` convention for what `prefix`/`source` mean on a `LOGIC_MODEL` entry.

### The two coverage gaps `05-verify.md` raised — my own severity call

1. **Slash-free-datasource split unasserted for prefix/source** (only exercised, not asserted,
   by the pre-existing `attachBaseWorksheetAcceptsLogicalModelType` test). By hand-trace (mine,
   matching the verifier's) `lastIndexOf('/')` is the same single code path regardless of how
   many slashes appear before the last one — there is no separate branch for the zero-slash
   case. **Non-blocking.** Low-risk, nitpick-level: worth a follow-up assertion, not worth
   holding up this PR.

2. **New `PairingException` guard for a bare model name (no `/` at all) is untested.** This is
   new, user-visible error behavior introduced by this diff, and it changes the outcome for
   `path:"Order Model"` — one of the four repro variants named in the original #76606 bug
   report — from a generic not-found message to an immediate, field-named format-guidance
   message. My own trace agrees with the verifier's: this is a net improvement, not a
   regression, and the logic is correct. **But this is "missing regression coverage for a
   claimed fix"** (this repo's review bar explicitly lists that as blocking-worthy) — the guard
   is part of what this commit claims to fix for one of the documented repro variants, and nothing
   in the suite exercises it. This is a one-method addition (call
   `attach_base_worksheet(type:"logicalModel", path:"Order Model")`, assert it throws
   `PairingException` with a message containing `"path"`/the expected-format text and NOT the
   `"no logical model named"` wording).

Requesting one addition before merge: a regression test for the new `PairingException` guard
on a bare (no `/`) model-name path, e.g. `path:"Order Model"`. Everything else in this PR is
correct, generic, and scoped — no other blocking issues found.

(Side note, not part of this review's blocking findings: CI's `build` check was still
`pending`/in-progress and the `claude-review` bot check failed with what looks like an
unrelated tool/infra error (`is_error:true` after 1 turn, `total_cost_usd: 0` — no actual
review content produced), not a substantive finding about this diff.)

NeverStopFMH:
## Review round 2 (reviewer-76606-r2)

Independently verified the new commit `a4c1bd4ab7` (test-only, 33 insertions, 0 deletions, no production-code change) closing round 1's one requested gap:

- New test `attachBaseWorksheetRefusesLogicalModelMissingSeparator` calls `attach_base_worksheet(type:"logicalModel", path:"Order Model")` and asserts it throws `PairingException` with the format-guidance message ("'path' must be \"<datasource>/<model>\"") and NOT the generic "no logical model named" message. Hand-traced against the guard (`sepIdx <= 0 || sepIdx == trimmedPath.length() - 1`): for `"Order Model"`, `sepIdx == -1`, guard fires before any repository probe — matches the assertion exactly.
- Ran the full `ViewsheetAssemblyAgentControllerTest` suite myself: **93 tests, 0 failures, 0 errors — BUILD SUCCESS.**
- Control experiment: temporarily disabled just the guard (`if(sepIdx <= 0 || ...)` → `if(false)`) in production code, re-ran only the new test method — **it failed** (`expected: <true> but was: <false>`), confirming the test is a real regression guard, not decorative. Restored the file, confirmed `git status` clean, re-ran full suite — back to 93/93 green.
- Confirmed round-1-approved parts of the diff (the production-code fix, the earlier `attachBaseWorksheetLogicalModelProbesEntryWithPrefixAndSourceProperties` test) are byte-for-byte unchanged by this commit.
- CI: `build` was pending on this fresh push; polled until it passed (~12m42s). `claude-review` still fails — this is the pre-existing, known-broken infra check identified in round 1 (misconfigured `ANTHROPIC_API_KEY` in that job), unrelated to code content, not treated as a finding. `Semgrep OSS`, `Semgrep Scan`, `license/cla` all pass. `mergeable: MERGEABLE`.

No blocking issues found. This round closes clean — squash-merging.

(Formal `gh pr review --approve` unavailable in this environment: same-account restriction, as round 1 documented.)

PR #5199, opened 2026-09-14
Bug #76510: fix Scale-to-Screen overlap on initial embed load

larryliang-inetsoft:
## Summary

Fixes Redmine bug #76510: a viewsheet embedded via the `inetsoft-viewer` web component (`scale="true"`) with Scale-to-Screen enabled renders with overlapping/misplaced components on initial load. Toggling fullscreen or refreshing the page fixes the layout.

**Root cause:** `ViewerAppComponent.openViewsheet()` computes
```ts
const waitResize = this.scaleToScreen && this.inPortal && !this.fitToWidth ? 100 : 0;
```
For the web-component embed path, `inPortal` is always `false`, so `waitResize` was always `0` and `openViewsheet0()` (which calls `setAppSize()`, reading `viewerRoot.nativeElement.offsetWidth/offsetHeight`) ran synchronously on bootstrap. This races the host page's own layout resolution for the freshly-upgraded custom element, so the initial `OpenViewsheetEvent` carries a wrong (sometimes zero-height) size. With `scaleToScreen=true`, the server computes its initial layout scale from that bad size, producing the overlap. A later resize (fullscreen toggle, or a full page refresh restarting `openViewsheet`) re-reads the by-then-settled dimensions and self-corrects — which is why toggling fullscreen or refreshing "fixes" it.

**Fix:** widen the existing (already portal-proven) `waitResize` mechanism to also cover the embed path by dropping the `inPortal` condition:
```ts
const waitResize = this.scaleToScreen && !this.fitToWidth ? 100 : 0;
```
This is a minimal, one-line change. Portal behavior is unchanged: `inPortal` was always `true` for the portal case, so the condition's truth value there is unaffected.

Also removes leftover `console.log("BUG76510-TRACE ...")` debug instrumentation added during the investigation (in `viewer-app.component.ts` and `embed-viewer.component.ts`) — pure logging, no functional changes, now fully removed.

## Test plan
- [x] Added a regression test (`viewer-app.component.interaction.tl.spec.ts`, Group 27) asserting `openViewsheet()` delays the `/events/open` STOMP send by 100ms when `scaleToScreen && !fitToWidth`, regardless of `inPortal`, and sends immediately when `fitToWidth=true`.
- [x] `npm run lint` — 0 errors (pre-existing warnings only, unrelated to this change).
- [x] Scoped test run (`ng run portal:test-tl --include=".../viewer-app.component.interaction.tl.spec.ts"`) — 77/77 tests passed, including the 2 new ones.


fredaan (commented):
## Review

Verified the fix against `ViewerAppComponent.openViewsheet()` / `ngOnInit()` / `processSetViewsheetInfoCommand()`. The logic itself is correct, but the **PR description is stale** and no longer matches the shipped diff — worth fixing before merge so history stays accurate.

**What the description says the fix is:**
```ts
const waitResize = this.scaleToScreen && !this.fitToWidth ? 100 : 0;
```
(dropping only the `inPortal` condition)

**What's actually in the diff (commit `7ab9012c`, "round 2"):**
```ts
const waitResize = !this.inPortal || (this.scaleToScreen && !this.fitToWidth) ? 100 : 0;
```

These are materially different. I confirmed why round 2 was necessary: `this.scaleToScreen`/`this.fitToWidth` are only populated either by the portal's route resolver (`viewDataService.data`, read in `ngOnInit`) or by the server's `processSetViewsheetInfoCommand` response — which arrives *after* the very `OpenViewsheetEvent` that `openViewsheet()` is about to send. For the embed/web-component path (and the standalone `/app/viewer/**` route, and Composer's `linkView` tabs), neither source has run yet at that point, so both fields are still `false`. Round 1's `scaleToScreen && !fitToWidth` therefore always evaluated to `0` for those hosts — a no-op, as the round-2 commit message itself explains. The round-2 condition (`!inPortal || ...`) correctly fixes this by making *any* non-portal host wait unconditionally, since we can't know in advance whether it'll turn out to need scale-to-screen handling.

That's the right call given the constraint, but it's a bigger behavior change than the description advertises: **every non-portal viewsheet open (embed, standalone viewer route, Composer linkView) now always pays a 100ms delay before the open event is sent, even when scale-to-screen isn't in play at all** — not just the scale-to-screen embeds the description focuses on. Please update the PR description to reflect the round-2 rationale (the round-2 commit message already has a good explanation to lift from) so reviewers are evaluating the actual shipped behavior, including that universal-wait trade-off for non-portal hosts.

Code and regression tests otherwise look correct and I didn't find functional bugs — portal's conditional wait is provably unchanged (`!inPortal` is `false` there, reducing to the original expression), and the round-2 test fixes (modeling real pre-server-response defaults, advancing fake timers by 0ms before asserting "not yet called") are sound.

fredaan (approved):

PR #5200, opened 2026-09-14
feat(admin): add recycle bin browse/restore/purge area (Redmine #76604)

lizhepu:
## Summary
- Adds a new community-tier `inetsoft.web.admin.ai.recyclebin` package (`AdminRecycleBinController` + `RecycleBinService`/`RecycleBinChangePlanService`/`RecycleBinChangesetApplyService` + DTOs), giving admin-chat browse/restore/purge over whatever's already in StyleBI's recycle bin (populated today only by EM/Composer/Portal deletes — admin-chat's own `apply_viewsheet_changes` delete still bypasses the recycle bin entirely, unchanged, by deliberate scope decision).
- Delegates directly to the existing `RecycleUtils`/`RecycleBin`/`AssetRepository` primitives; never forwards to `RepositoryRecycleBinController` itself.
- Placed in the community module (not enterprise) to match where the underlying capability actually lives and is already available.

## Test plan
- `mvn -pl community/core -am compile` — clean, zero errors.
- No new Java unit tests added this pass — flagged in the companion plugin-side PR's build report as needing a live/in-container StyleBI instance to validate `RecycleUtils` static-mock behavior with confidence; not written speculatively.
- No live StyleBI server exercised — restore/purge dispatch and rollback re-recycle paths are traced against source and compile clean only.

See `docs/teams/2026-09-14-viewsheet-parity-76604/track-a-recycle-bin/` in the `stylebi-wiz` repo for the full design/reconcile/build record.


lizhepu:
## Review round — Java side (stylebi#5200)

**Method**: read the design/reconcile/build docs and `plugin/admin/CLAUDE.md`'s Recycle Bin section first (this PR is reviewed as the companion to stylebi-wiz#2474). For each new file, stated the invariant it claims to hold, then tried to construct a scenario that breaks it (other call sites of the same primitive, sibling branches — e.g. does the same fix show up on both the `preview`-time and `apply`-time collision check, restore vs. purge, sheet vs. WS-folder vs. repository-folder — boundary/shape variants, repeat/ordering, config variants).

**The one item this track most needed verified — restore-collision refusal (`03-reconcile.md`'s settled call)**:
`RecycleBinChangePlanService.resolveOne` (preview) throws loud, naming the occupied path, when `collides && !overwrite` (lines ~527–533) — confirmed this is NOT only a preview-time check: `RecycleBinChangesetApplyService.applyRestore` re-runs the identical `wouldCollide` check fresh at apply time (lines ~900–907, explicitly commented "never trust anything computed at preview"), so a destination that became occupied *between* preview and apply is still caught, not silently allowed through on a stale plan. And because `risk` is part of the plan-hash input, any drift in collision state between preview and apply changes the hash and trips `PlanHashMismatchException` (409) *before* the `acknowledgeIrreversibleDelete` check is even reached — so there's no path where a plan reviewed as low-risk sneaks through as a destructive overwrite without a fresh confirmation. Tried the sibling case (destination freed up between preview and apply, not just occupied) — also handled correctly (falls back to `RISK_LOW`, no spurious ack requirement).

**Purge unconditionally requires `acknowledgeIrreversibleDelete`**: confirmed via `hasIrreversible` in `apply()` — a plan-wide `RISK_HIGH` scan, not a per-verb special case, so it also catches the "restore + overwrite + real collision" combination without a second flag (per the reconcile decision to reuse the same flag name).

**Package placement**: verified independently, not by trusting the PR description — this repo (`inetsoft-technology/stylebi`, checked out at the PR's own head `dc1ad214d`) has no `enterprise/` directory at all alongside `core/`; the new package sits at `core/src/main/java/inetsoft/web/admin/ai/recyclebin`, i.e. genuinely community-tier by construction, not merely by claim.

**`apply_viewsheet_changes`'s existing delete semantics**: diff is 100% new files under the new `recyclebin` package; zero touches to `ViewsheetApiService`/`ViewsheetChangePlanService`/`ViewsheetChangesetApplyService` or anything else pre-existing. Confirmed via `gh pr diff --name-only`.

**The one piece of genuinely new (not-delegated-to-an-existing-primitive) logic — purging a repository (dashboard) folder via `RepletRegistry.removeFolder(entry.getPath())`** (rather than calling an existing `RecycleUtils` method, since none exists for purge): traced this against the actual `RecycleUtils.moveRepositoryFolderToRecycleBin` source (fetched from this repo at the PR's base ref) — recycling a repository folder is implemented as `registry.changeFolder(originalPath, trashPath, ...)`, i.e. the folder's registry label is renamed to live AT the trash path, carrying the "My Dashboard" prefix convention along consistently for user-owned folders. So `registry.removeFolder(entry.getPath())` is dispatching against the label's actual current location, not a mistranslated path. This matches `04-build.md`'s own disclosed caveat that this one dispatch was "traced against source, not verified against a live server" — I did the source-tracing verification myself rather than taking that at face value; the live-server gap remains open and disclosed, not resolved by this round.

**Minor, non-blocking observation** (confidence well under the reporting bar, noting only for completeness): `RecycleBinChangesetApplyService.rollbackRestore`'s WS-folder branch calls the `stripMyDashboard` helper on `entry.getOriginalPath()` before use, but tracing `RecycleUtils.restoreWSFolder`'s own source shows a WS-folder's `originalPath` never carries that prefix in the first place (only sheets/viewsheets do) — so the call is a harmless no-op there, not a functional bug. Not reporting this as a finding.

**Compilation — actually run, not taken from `04-build.md`'s self-report**: used the pre-existing dedicated worktree at `E:\inetsoft\wizard-wt-76604-recycle-bin\community` (confirmed HEAD `dc1ad214d` exactly matches this PR's head sha via `gh api .../pulls/5200 --jq .head.sha`), NOT the shared `E:\inetsoft\wizard` checkout. Deleted the already-built `core/target/classes/inetsoft/web/admin/ai/recyclebin` output first to force a real recompile (an untouched incremental build had skipped these files as "up to date" from a prior run), then ran `./mvnw -pl core -am compile -DskipTests -o`: **BUILD SUCCESS**, 12 new source files (plus dependents) compiled clean, zero errors.

**No new Java unit tests in this PR** — disclosed plainly in the PR body/build report, not hidden. Not treating "no tests" as "overfit tests": there's nothing here narrowed to a specific reported literal: the design's own review-time finding (destination-collision-without-overwrite refusal) is implemented as a general collision check reused identically at both preview and apply time, not an if-branch keyed to a specific path/entry name.

**Findings: none at blocker/important (≥80 confidence) severity.**

**Conclusion: no blocker/important findings, recommend approve.** Same account-restriction note as the companion plugin PR: this CLI session is authenticated as `lizhepu`, the same account that opened this PR, so GitHub refuses a formal self-approve — recording the review result as a comment instead.

PR #5205, opened 2026-09-14
Fix silently-swallowed rename failures in org-lifecycle dashboard registry copy

larryliang-inetsoft:
## Summary

`AbstractEditableAuthenticationProvider.copyDataSpace()` called `DataSpace.rename()` and discarded its boolean return value. A rename that failed (e.g. `BlobStorage.rename()`'s hard 10-second timeout tripping under load) was indistinguishable from success, so `copyOrganizationInternal()` went on to clear the source org's caches/properties as if the whole org had migrated — leaving stale registry files behind.

This caused an intermittent CI failure in `DashboardRegistryOrgLifecycleTest` (`rename_copyOrganizationInternal_dashboardFilesRelocatedByDataSpaceRename_contentNotRewritten`, assertion at line 274: "old admin registry path must no longer exist ... expected: <false> but was: <true>"), observed on an unrelated PR's CI run. A manual rerun of that same job then hung for ~39 minutes in the same test area before being cancelled by CI's timeout — a second, distinct symptom in the same code path, consistent with contention on the executor the rename funnels through.

## Root cause (two compounding issues, same call path)

- **Defect A** — `copyDataSpace()` discarded `dataspace.rename()`'s return value. `DataSpace.rename()` silently returns `false` on `FileNotFoundException`/`IOException` (only logging a WARN), and `BlobStorage.rename()` bounds the actual rename to a hard 10s `cluster.submit(...).get()`. Under CI-load contention on the shared single-thread rename executor, this timeout can trip; nobody downstream ever checked the result.
- **Defect B** — `getOrgScopedPaths()` returns a flat list of every directory placeholder and descendant file, but `DataSpace.rename()` already recurses through a directory's descendants on its own. So renaming an ancestor directory silently makes every subsequent rename call for its already-moved descendants redundant — extra load on the same contended executor that Defect A's timeout budget depends on.

## Fix

- Capture `rename()`'s boolean result; on `false`, log ERROR with path/newPath and accumulate the failure. After the loop, if any rename failed under `replace=true`, throw rather than letting `copyOrganizationInternal()` proceed as if the move fully succeeded.
- De-duplicate the scoped path list before renaming: skip any path that is a descendant of another path already in the list, since `DataSpace.rename()` already recurses through directories.
- Confirmed `FileAuthenticationProvider` (the production provider) does not override `copyDataSpace()`/`copyOrganization()`, so this applies uniformly to production, not just the test's `StubProvider`.

Scoped to `copyDataSpace()`'s call site only — `DataSpace.rename()`'s own exception/return contract is unchanged (it has ~28 other call sites file-wide; a contract change there is out of scope for this fix).

A third mitigation (retrying `BlobStorage.rename()` once on timeout) was drafted and then dropped after review found it could race the still-in-flight original rename task and, combined with the fail-loud change above, misreport an actually-successful-but-slow rename as a hard failure — worse than the original bug. Not included here.

A fourth, lower-confidence lead (a possible non-volatile static-field race in `ConfigurationContext.getSpringBean()` that might explain the 39-minute hang specifically) is documented but intentionally not pursued in this PR — it needs a reproduced hang/thread dump to confirm before touching that code.

## Test plan

- [x] Added `copyDataSpace_replaceTrue_renameReturnsFalse_throwsAndDoesNotSwallowFailure` — regression test for the fail-loud behavior.
- [x] Added `copyDataSpace_replaceTrue_nestedPerUserRegistry_eachPathRenamedExactlyOnce` — regression test for the de-duplication, using a nested per-user dashboard registry fixture.
- [x] Updated the pre-existing `copyDataSpace_replaceTrue_renamesAllScopedPaths` test to stub `rename()` → `true` (needed once the return value is checked; this was a real gap the fix exposed).
- [x] `DashboardRegistryOrgLifecycleTest`: 4 run, 0 failures, 1 skipped (pre-existing, unrelated `@Disabled` scenario, untouched).
- [x] `AbstractEditableAuthenticationProviderStaticDepTest`: 32 run, 0 failures.
- [x] `AbstractEditableAuthenticationProviderTest`: 39 run, 0 failures, 6 skipped (pre-existing `@Disabled`s, untouched).
- [x] Independent code review pass (correctness, scope discipline, path-prefix edge cases) — no issues found in the code as merged here.


jshobe-inetsoft (approved):
## Manual review (replacing the failed `claude-review` bot run)

The automated `claude-review` check on this PR failed due to a CI/quota infrastructure error in the bot itself (`is_error:true`, 1 turn, $0 cost — its Claude invocation never actually ran), not because it found a blocking issue. This is a manual review performed in its place.

### Summary of the change

`copyDataSpace()` in `AbstractEditableAuthenticationProvider` previously discarded the boolean return value of `DataSpace.rename()`. A failed rename (e.g. `BlobStorage.rename()`'s hard 10s `cluster.submit(...).get()` timing out under load) was indistinguishable from success, so `copyOrganizationInternal()` proceeded to clear the source org's caches/properties as if the whole org had migrated, leaving stale registry files behind. This PR:

1. **Fail loud (Defect A)** — captures each `rename()` result, logs an ERROR with path/newPath on failure, and throws `RuntimeException` after the loop if any rename failed under `replace=true`, instead of letting the caller proceed as if the migration fully succeeded.
2. **De-duplicate scoped paths (Defect B)** — `getOrgScopedPaths()` returns a flat list including both a directory placeholder and every descendant file/dir individually, but `DataSpace.rename()`/`renameRecursively()` already recurses through a directory's own descendants. `removeRedundantDescendantPaths()` sorts by string length (guaranteeing ancestors are evaluated before descendants) and drops any path that is a descendant (`path.startsWith(ancestor + "/")`, correctly avoiding false positives like `portal/fromOrg` vs `portal/fromOrg2`) of a path already kept.

### Verification performed

- Read `DataSpace.rename()`/`renameRecursively()` and `getOrgScopedPaths()` directly: confirmed the PR's description of both the recursion behavior and the flat-list shape returned by `getOrgScopedPaths()` is accurate.
- Confirmed **Defect B's dedup is not just an optimization — it's required for Defect A's correctness**: without it, an already-recursively-renamed descendant path would be renamed again, hit `FileNotFoundException` in `renameRecursively`, return `false`, and — post-Defect-A — be misreported as a real failure on every successful org copy that has nested per-user paths. The PR handles this dependency correctly by shipping both fixes together, and the ordering-independence of the dedup (sort-by-length before filtering) means it's correct regardless of the order `getOrgScopedPaths()` returns paths in.
- Confirmed `FileAuthenticationProvider` (production) does not override `copyDataSpace()`/`copyOrganization()` — grepped for overrides, found none — so the fix applies uniformly, not just to the test's `StubProvider`.
- Confirmed `copyDataSpace()` is called early in `copyOrganizationInternal()`, before the `replace`-branch's destructive cleanup (`clearScopedProperties`, `dashboardRegistryManager.clear()`, identity/permission updates). A thrown exception here aborts before those steps run, so the source org's state isn't torn down on a failed migration — this is the right place to fail fast.
- Confirmed `copyOrganization`'s public interface methods declare no checked exceptions, so adding an unchecked `RuntimeException` doesn't require a signature change; `RuntimeException` is also the exception type already used elsewhere in this same file (3 other call sites), so this is stylistically consistent.
- Checked all other call sites of `copyOrganization` (`IdentityService`, `UserTreeService`) — none currently catch exceptions around this call, so a failure will now propagate as an uncaught exception (presumably surfacing as a 500 at the REST boundary) instead of silently succeeding. This is the intended behavior change, not an accidental regression.
- Checked the other Spring-context-backed org-lifecycle test suites that also extend `AbstractEditableAuthenticationProvider` via `StubProvider` (`DataCycleManagerOrgLifecycleTest`, `AbstractEditableAuthenticationProviderTest`, `DashboardRegistryOrgLifecycleTest`, `OrgLifecyclePortalBrandingTest`, `OrgLifecycleThemeOrchestrationTest`) for the classic Mockito gotcha here — an unstubbed mocked `rename()` defaults to `false`, which would now throw where it previously didn't. Only `AbstractEditableAuthenticationProviderStaticDepTest` uses a fully mocked `DataSpace` via `mockStatic`, and the PR correctly updates its one affected pre-existing test (`copyDataSpace_replaceTrue_renamesAllScopedPaths`) to stub `rename()` → `true`. The other suites use a real/Spring-backed `DataSpace`, so they're unaffected. The `build` check result (SUCCESS, run after the latest push) corroborates this.
- New regression tests (`...renameReturnsFalse_throwsAndDoesNotSwallowFailure`, `...nestedPerUserRegistry_eachPathRenamedExactlyOnce`) exercise both fixes directly and are well-targeted at the two defects described.
- The `InvocationTargetException` unwrapping added to the test's reflection helper is correct and necessary so `assertThrows` sees the real `RuntimeException` rather than a reflection wrapper.
- Imports are all covered by the existing `java.util.*` wildcard import; no missing imports.

### Non-blocking observations (not required for approval)

- The same redundant-recursive-call pattern (Defect B) also technically exists on the `copy()`/`replace=false` path (`DataSpace.copy()`/`copyRecursively()` has the identical recursion structure, and `BlobStorage.copy()` goes through the same `cluster.submit(...).get(10s)` pattern as `rename()`). The PR deliberately scopes both the fail-loud and de-dup changes to the `replace=true` branch only, which matches the title/observed bug (this is a *rename* failure PR) and the "Scoped to copyDataSpace()'s call site only" statement in the description. Worth a future follow-up if the copy path ever shows the same executor-contention symptom, but not a defect in this PR.
- The dropped "retry rename once on timeout" mitigation and the un-pursued 39-minute-hang lead are both reasonably explained and correctly left out of scope.

### Checklist against required approval conditions

- No blocking issues found in code review. ✅
- No failing checks other than `claude-review` (Semgrep Scan, build, Semgrep OSS, license/cla all SUCCESS). ✅
- No merge conflicts (`mergeable: MERGEABLE`; the `BLOCKED` merge state is solely due to the missing required review, which this review supplies). ✅

**Verdict: Approve.** This is a well-scoped, correctly-reasoned fix with solid regression test coverage. Not auto-merging per review policy.

PR #5194, opened 2026-09-14
feat(wiz): real set_column_labels implementation (VTB-011, Redmine #76574)

lizhepu:
## Summary

Replaces the deliberate `UnsupportedOperationException` (HTTP 501) stub that unconditionally
refused `set_column_labels` on a bound column with a real implementation.

- **Table**: `ColumnRefModel.alias` is set directly (an existing, already-wired field —
  `VSTableBindingFactory.updateTableAssembly` already round-trips it). Includes a port of
  `ComposerVSTableService.changeColumnTitle`'s `FormatInfo`/column-width rekey so a pre-existing
  per-column format follows the rename instead of orphaning under the old name.
- **Crosstab**: needs a rendered `VSTableLens` to resolve a column to its header `TableDataPath`,
  which `TableBindingMutator` (model-only) doesn't have — it now only validates and resolves the
  label down to a shelf position, handing that off to `TableBindingService` (which has
  `RuntimeViewsheet`/`ViewsheetSandbox` access) to do the actual `FormatInfo` write. Extracted the
  shared `SetTableHeaderAliasHandler` (from `ComposerVSTableService`'s own private `setAlias`, now
  delegating to it — mechanical, native Composer behavior unchanged) and promoted
  `ClearTableHeaderAliasHandler.matchAgg`/`matchDim` to package-visible for the reverse (name ->
  `TableDataPath`) resolution.
- New `entries: [{shelf, column, index, label}]` disambiguates a column bound more than once (a
  Year/Quarter drill), alongside the existing `labels` map for the common case.
- `get_table_binding`'s `columnLabels` stops sourcing from the dead `name2Labels` field; each
  shelf's `FieldRef` now carries a `label`, and `columnLabels` is populated only for unambiguous
  columns (a `columnLabelsNote` when a column is bound more than once).

## Design doc

`docs/teams/2026-09-11-bugs-vtb-p3/bug-76574-vtb011/01-design.md` (stylebi-wiz repo) — full
rationale, file:line citations, and confirmed decisions (§7a).

## Refresh/repaint finding (design's highest-risk open question)

The Crosstab `FormatInfo` write happens on the *live* assembly before
`bindingModelService.setBinding()` clones it — since that clone (via `bfactory.updateAssembly`)
only touches the binding-model-derived parts of the assembly's info, not `FormatInfo`, the
existing refresh (`assemblyInfoHandler.apply()` -> `coreLifecycleService.execute`/
[...]

PR #5206, opened 2026-09-14
fix(wiz): port syncHighlight into VTB-011's Table rename rekey

lizhepu:
## Summary

Follow-up to the already-merged VTB-011 PR #5194 (`set_column_labels` real implementation,
Redmine #76574), addressing a gap found by a retroactive Java review.

- Design §7a item 2 (user-confirmed) required porting `ComposerVSTableService.changeColumnTitle`'s
  full Table-branch rekey "so a pre-existing per-column format/highlight/width entry follows the
  rename instead of silently orphaning under the old name." The merged implementation
  (`TableBindingService.rekeyTableFormatAndWidth`) only ported the `FormatInfo` and column-width
  rekeys — it never touched `TableDataVSAssemblyInfo.getHighlightAttr()`. `syncHighlight`, the
  native private method `changeColumnTitle` calls for exactly this, was never ported.
- **Effect**: a Table column with an existing per-column highlight silently lost it whenever
  `set_column_labels` renamed that column — the highlight stayed keyed under the old display name,
  which no longer matched anything.
- **Fix**: port `ComposerVSTableService.syncHighlight`'s exact rekey logic (confirmed via `git
  show` against the pre-refactor source) into a new `syncHighlight` helper, called from the
  renamed `rekeyTableFormatWidthAndHighlight` (was `rekeyTableFormatAndWidth` — renamed since it
  now does three rekeys, not two).
- Also fixes a related test-coverage gap: `setColumnLabelsRenamesATableColumnAndRekeysItsFormat`
  mocked `TableDataVSAssemblyInfo` to avoid an NPE on the width-rekey call but never asserted the
  call happened. Captures the mock as a named variable and adds
  `verify(mockedTableDataVSAssemblyInfo).updateColumnWidthNames("REGION", "Sales Region")`.
- Adds `setColumnLabelsRenamesATableColumnAndRekeysItsHighlight`, a regression test that seeds a
  `TableHighlightAttr` entry under the old display name and asserts it moves to the new one (and
  is gone from the old one) after the rename.

## Test plan

- [x] `./mvnw -pl core test -o -Dtest=TableBindingServiceTest,TableBindingMutatorTest,SetTableHeaderAliasHandlerTest`
      — 121 tests, 0 failures.
- [x] Fail-before/pass-after discipline: temporarily commented out the new `syncHighlight(...)`
      call — the new highlight regression test failed as expected; restored it — the test passes.
- [x] Broader regression: `./mvnw -pl core test -o -Dtest="inetsoft.web.binding.**,inetsoft.web.wiz.**"`
      — 3753 tests, 0 failures, 1 skipped (pre-existing, unrelated), matching this feature's
      established baseline.

**Not merging** — needs review first.


lizhepu (commented):
Verified against the actual diff (`gh pr diff 5206`) and against the pre-refactor original,
`ComposerVSTableService.syncHighlight` (`git show 64512b1d6:...ComposerVSTableService.java` in
the `stylebi`/`community` history) — line-by-line, not just by description.

**1. Ported `syncHighlight` logic** — confirmed a faithful, correctly batch-generalized port:
same `hattr == null` guard, same map-snapshot-then-mutate (`new ArrayList<>(map.keySet())`
before removing keys, avoiding `ConcurrentModificationException`), same
`path.clone(String[])`/`getHighlight`/`map.remove`/`setHighlight` sequence. The batch version
folds the original's `Objects.equals(oname, tpaths[0])` single-name check into the existing
`renamedTo(renames, pathArr[0])` helper (already used by the `FormatInfo` rekey loop above it in
the same method) — a single pass over a path snapshot checked against all renames, which is
actually a cleaner generalization than a naive per-rename loop, and preserves the "old-name
lookups captured before any mutation" property that avoids rename-chain collisions (verified this
matters: paths are snapshotted from the pre-mutation map, so renaming A→B and B→C in one call
can't clobber each other).

One deliberate, harmless deviation from the original: the ported version drops the original's
separate `TableDataVSAssemblyInfo info = ...; if(info == null) return;` guard, going straight to
`table.getTableDataVSAssemblyInfo().getHighlightAttr()`. This is consistent with the
pre-existing, unguarded `updateColumnWidthNames` call three lines above it in the same method
(same assumption already established before this PR) — not a new regression.

**2. Method rename** — `rekeyTableFormatAndWidth` → `rekeyTableFormatWidthAndHighlight` applied
consistently: the one call site in `setColumnLabels`, both javadocs. Only remaining reference to
the old name is in the new test's own docstring, explicitly framed as history ("as originally
named") — not a stale/incorrect reference.

**3. New regression test** (`setColumnLabelsRenamesATableColumnAndRekeysItsHighlight`) — verified
both seed gotchas against the actual source:
`TextHighlight.isEmpty()` does return `getName() == null` (confirmed), and
`HighlightGroup.validate()` does drop an entry when `isEmpty() && isConditionEmpty()` both hold
(confirmed) — so the `setName("h1")` call is load-bearing, exactly as claimed. Also confirmed
`TableDataPath.equals()` compares by value (level/type/row/col/path contents, not identity), so
the test's independently-constructed `newPath` correctly matches the map key produced by
`path.clone(...)` inside `syncHighlight`, and `TableHighlightAttr.setHighlight` stores the
`HighlightGroup` reference as-is (`hlmap.put(tpath, hg)`, no defensive copy) — so `assertSame`
genuinely proves the *same* instance moved, not a re-created lookalike, and does so by asserting
absence under the old path too.

**4. Test-coverage-gap fix** — `TableDataVSAssemblyInfo` mock is now a named
`mockedTableDataVSAssemblyInfo` with a real
`verify(mockedTableDataVSAssemblyInfo).updateColumnWidthNames("REGION", "Sales Region")` —
confirmed, no more silently-accepting anonymous mock.

**5. Fail-before/pass-after** — single-commit PR, can't be independently re-run from PR history
alone; the test's own assertions (identity + old-key absence) are self-evidently tied to the
fix, consistent with this batch's usual caveat.

**6. Scope** — diff touches exactly two files (`TableBindingService.java`,
`TableBindingServiceTest.java`), 47+/7- and 59+/2-. No drive-by changes found.

**7. Base branch** — `stylebi-wiz-main-rebase`, confirmed.

Aside: the `claude-review` bot check on this PR errored out after 1s (same as it did on the
original #5194 per `05-review-r1-java.md`) — a tooling issue, not a finding against this diff.

No issues found. (Can't self-approve as PR author; posting as a comment review instead.)
Ready to merge pending the user's own final go-ahead.

PR #5207, opened 2026-09-14
Fix #76627 WBS-045: surface namedGroup in read_worksheet_model's groups

fredaan:
## Bug

Redmine #76627 / WBS-045: `read_worksheet_model`'s `aggregates.groups[]` entries never surface which `namedGroup` (if any) governs a group-by field. A caller applies `set_group_aggregate` with a `namedGroup` option, then reads the worksheet back to confirm what was applied — and the named group is simply absent from the response, indistinguishable from never having been set.

## Root cause

`WorksheetReadService#readAggregates` (`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetReadService.java`, ~L790-831) builds each `GroupModel` from a `GroupRef` without calling `gr.getNamedGroupAssembly()` — a getter that already exists (`GroupRef.java:350-352`) — and the `GroupModel` record itself (`WorksheetModel.java`) had no field to hold it.

The write path already stores the value correctly: `WorksheetMutationSupport#applyAggregateInfo` calls `gr.setNamedGroupAssembly(...)` on the exact same `GroupRef` object instance that `readAggregates` later reads. This is a pure read-side omission — structurally identical to the `timeSeries` gap fixed in `2e071d982` (`WorksheetReadServiceTest#readsTimeSeriesFlagOnDateGroupedColumn`), which this fix mirrors.

## Fix

- `WorksheetModel.GroupModel`: add a nullable `String namedGroup` field. The record already carries `@JsonInclude(JsonInclude.Include.NON_NULL)`, so a `null` value is simply omitted from JSON — no other change needed for backward compatibility.
- `WorksheetReadService#readAggregates`: pass `gr.getNamedGroupAssembly()` through to the new `GroupModel` field.
- `WorksheetReadServiceTest`: add `readsNamedGroupOnGroupedColumn` (asserts the named group name round-trips) and `reportsNullNamedGroupWhenUnset` (asserts a plain group still reports `null`), mirroring `readsTimeSeriesFlagOnDateGroupedColumn`'s structure.

## Testing

```
./mvnw -o -pl core -am test -Dtest=WorksheetReadServiceTest -Dsurefire.failIfNoSpecifiedTests=false
```
`Tests run: 48, Failures: 0, Errors: 0, Skipped: 0` (46 pre-existing + 2 new).

Confirmed `GroupModel`'s only production construction site is `WorksheetReadService.java:813` (verified by grep, and by the full `core` module test-compile succeeding with no other callsite breaking on the new arity).


fredaan (commented):
## Review — PR #5207 (Fix #76627 WBS-045: surface `namedGroup` in `read_worksheet_model`'s groups)

**Note:** I am running as `fredaan`, the PR's own author — GitHub refuses `--approve` from the PR author, so this is posted as `--comment`. A different account (human or otherwise) needs to `--approve` and merge.

### 1. Diff matches the confirmed, narrow fix — PASS
Exactly three files touched, matching the confirmed root cause precisely:
- `WorksheetModel.java:241` — adds nullable `String namedGroup` to the `GroupModel` record, with Javadoc in the same style as the existing `field`/`dateLevel`/`timeSeries` params (and matching precedent commit `2e071d982`'s own style for `timeSeries`).
- `WorksheetReadService.java:813` — passes `gr.getNamedGroupAssembly()` through at the one construction site.
- `WorksheetReadServiceTest.java` — two new regression tests.

`WorksheetMutationSupport.java` (write path) is untouched, as expected — it was already correct.

### 2. Backward compatibility — PASS
Confirmed both `AggregateModel` and the nested `GroupModel` record carry `@JsonInclude(JsonInclude.Include.NON_NULL)` (read directly from the fix branch, not just the fix doc's claim). A `null` `namedGroup` — the common, ungrouped case — is omitted from JSON rather than serialized as `"namedGroup":null`.

### 3. Regression tests — real, not tautological — PASS
Read both new tests directly:
- `readsNamedGroupOnGroupedColumn` builds a real `DefaultNamedGroupAssembly` via the test file's existing `namedGroup(...)` helper and applies it through the actual production mutator (`WorksheetMutationSupport.applyAggregateInfo` with a `GroupSpec` carrying `namedGroup`), then asserts `group.namedGroup()` equals the applied name — not a hand-constructed fake `GroupModel`.
- `reportsNullNamedGroupWhenUnset` asserts a plain group-by column with no named group still reports `namedGroup() == null`, guarding against the fix always populating something.

Both round-trip through the real `GroupRef`/mutator path, mirroring `readsTimeSeriesFlagOnDateGroupedColumn`'s structure as intended.

Also confirmed the `GroupSpec(field, dateLevel, timeSeries, namedGroup)` 4-arg constructor used by the new test already existed pre-fix (checked `git show HEAD^:.../WorksheetMutationSupport.java`) — this PR isn't quietly depending on new write-path surface.

### 4. No other `GroupModel` construction site — PASS (independently verified)
Ran my own grep, not just trusting the fix doc's claim:
```
grep -rn "GroupModel(" core/src/main/java/inetsoft/web/wiz core/src/test/java/inetsoft/web/wiz
```
Only two real hits for `WorksheetModel.AggregateModel.GroupModel`: the record declaration (`WorksheetModel.java:241`) and the single construction site (`WorksheetReadService.java:813`). (Grep also turns up several unrelated `*GroupModel` classes — `EditGroupPaneModel`/`RotationRadioGroupModel`/etc. — different types entirely, not a match.) The 4-arg change is source-compatible with no other caller.

### 5. Base branch — PASS
`baseRefName` is `stylebi-wiz-main-rebase`, confirmed via `gh pr view`. Correct integration branch for this fork.

### 6. Code-quality pass — PASS
Javadoc addition matches the existing `field`/`dateLevel`/`timeSeries` doc style and precedent commit `2e071d982`'s own diff for `timeSeries`. No other style issues found.

### 7. CI status
- `Semgrep OSS` — pass
- `Semgrep Scan` — pass
- `license/cla` — pass
- `build` — pending at review time (not blocking, per task instructions)
- `claude-review` (automated Anthropic review action) — **fail**, but the failure is infra-shaped, not code-shaped: the job log shows `ANTHROPIC_API_KEY` empty in its env dump and the action exits after "No buffered inline comments" / a token-revoke step with no actual review content posted. This reads as an action/secrets configuration issue in the workflow, not a finding about this diff. Flagging for awareness; not treating it as a code-correctness blocker.

### Verdict
Clean. Narrow, correctly-scoped fix; backward-compatible; real regression tests exercising the production write path; no other construction site broken; correct base branch. No blocking issues found.

Nelson-cao (approved):

PR #5208, opened 2026-09-14
Bug #76607: fix logical-model calc field permanently unbindable

Nelson-cao:
## Summary
- `add_calc_field` on a Logical-Model-backed viewsheet stores the new field keyed by the model's own name, but the chart-scoped binding tree fans a model out into one `TableAssembly` per entity and looks calc fields up keyed by the entity's name — storage key and lookup key never matched, so the field was permanently unbindable to any assembly.
- `BaseTreeModelBuilder.appendColumnNodes` now also looks calc fields up by the model's own name, restricted to the call path that produces a properly-tagged `TABLE` node (so the field folds into the model's own bindable-fields group).

## Evidence
Full investigation, root-cause, and live-verification record: https://github.com/inetsoft-technology/stylebi-wiz/tree/docs/bug-76607-deep-debug/docs/teams/2026-09-14-bug-76607-calc-field-logical-model

## Test plan
- [x] Live-verified through the plugin's own MCP tools against a running StyleBI instance: `add_calc_field` → `list_bindable_fields` now shows the field under the model's group → `set_chart_shelf` succeeds (previously refused) → `get_viewsheet_image` renders the correct discount-adjusted values.
- [x] Affected unit suites green: `BindableFieldsServiceTest` (23), `BindableColumnsTest` (22), `CalcFieldAgentServiceTest` (14) — 59 tests, 0 failures.
- [ ] Known residual cosmetic issue (pre-existing, documented in the archive): the field also still appears a second time in an untagged fallback group in the same listing — harmless to binding/rendering, follow-up recommended.


Nelson-cao:
# Bug #76607 · Review round 1 (reviewer-r1)

PR: https://github.com/inetsoft-technology/stylebi/pull/5208
Head reviewed: `4295c7310` (branch `fix/76607-calc-field-logical-model-unbindable`, on top of
`e18855111` = `origin/stylebi-wiz-main-rebase`)
Diff scope: `core/src/main/java/inetsoft/analytic/composition/event/BaseTreeModelBuilder.java`
only (37 lines added, 0 removed).

## What was checked

Read `00-charter.md`, `00-map.md`, `02-root-cause.md`, `04-fix.md`, `05-verify.md` as settled
ground. Read the full diff and the surrounding un-diffed code it depends on:
`BaseTreeModelBuilder.appendColumnNodes`/`addLMEntryLevel`/`addCurrentQueryNode`/
`addAggregateNode`, `CubeTreeModelBuilder.getModel`/`getBuilder`, `VSTreeHandler.getChartTreeModel`/
`checkBaseWSChanged`, `BindableFieldsService` (`modelOf`/`fold`/`collect`/`gather`), `VSUtil.
appendCalcFields` (both overloads), `VSEventUtil.createPseudoAssemblies`, `Viewsheet.getWorksheet`/
`createTableAssembly`, and `AbstractDataRef.equals`/`CalculateRef.equals`. No live StyleBI access
(read-only teammate per the dispatch) — everything below is a static-analysis trace, called out
explicitly wherever it hasn't been live-confirmed.

## Overfit checklist (per review-workflow.md)

- **Keys on a repro literal?** No. The added code branches on `baseEntry.isLogicModel()` and
  `!addLMEntryLevel`, not on "Order Model", the calc field's name, or any other literal from the
  repro.
- **Special-case `if` excluding the reported input?** No — this is an addition of a new lookup, not
  an exclusion.
- **Fix at the call site while the broken invariant lives further down, breaking other callers?**
  Partially — see Finding 2 below. The restriction to `!addLMEntryLevel` is deliberate and (see
  Finding 1) turns out to be *justified* by where the call is actually reached from, not just
  reasoned about in the abstract. But there's a second, narrower way the same "wrong caller, right
  invariant" pattern still bites: see Finding 2.
- **Regression test narrow enough that a weaker fix would still pass?** There is no automated
  regression test at all for the changed method (see Finding 3) — narrower than "narrow", this is
  the empty case of that checklist item.
- **Same bug, other site?** `05-verify.md` already swept `VSUtil.appendCalcFields(Collection,
  TableAssembly, ...)` callers and found the two others (`VSEventUtil.java:2583,3703`) are inside
  the *unscoped* tree builder, which never fans a Logical Model into per-entity assemblies the same
  way — independently re-confirmed by reading that code (`VSEventUtil.appendColumnNodes` at
  `VSEventUtil.java:2565-2588` builds `list` from `assembly.getColumnSelection`, one call per
  already-correctly-named table/entity node in that tree shape). Agree with 05-verify.md: not the
  same defect, correctly left unmodified.

## Finding 1 (informational, not blocking) — source of the residual cosmetic duplicate, identified

`04-fix.md`/`05-verify.md` both flag a known residual: "Net Revenue" still shows a second time in
an untagged `Dimensions` fallback group, confirmed pre-existing but not traced to its source.
Traced it by reading (not live-verified — flagging that explicitly):

`CubeTreeModelBuilder.getModel()` (`CubeTreeModelBuilder.java:379-404`) calls, in order:
`addBaseQueries(assemblies)` → `addLMEntryLevel(assemblies, root)` → `addCurrentQueryNode(assemblies,
root)`, all three against the *same* `assemblies` list. For a Logical-Model base attached directly
(no separate worksheet — `Viewsheet.isDirectSource()` true), `Viewsheet.getWorksheet()` /
`createTableAssembly` (`Viewsheet.java:323-459`) synthesizes exactly **one** placeholder
`BoundTableAssembly`, named `VSUtil.getTableName(entry.getName())` — i.e. the *model's own name*
(`Viewsheet.java:441-445`) — and that's the sole assembly `addBaseQueries` puts in `baseWS.
getAssemblies()`.

`addLMEntryLevel` runs first and iterates that list *before* `addCurrentQueryNode` has replaced it
with per-entity pseudo-assemblies (`VSEventUtil.createPseudoAssemblies`, which only runs inside
`addCurrentQueryNode`'s own loop body, called after `addLMEntryLevel` returns). So
`addLMEntryLevel`'s call to `appendColumnNodes(assembly, root, true, true)` — the `addLMEntryLevel
=true` path this PR deliberately excludes — receives `assembly` = **the model-named placeholder**,
not an entity. `assembly.getName()` there equals the model's own name, which *is* the storage key
`add_calc_field` used. So the **pre-existing, unmodified** call at line 444
(`VSUtil.appendCalcFields(list, assembly, vs, ...)`, which does `vs.getCalcFields(assembly.
getName())`) already finds the calc field on this path — by coincidence of the placeholder's name,
not because of anything this PR touches. Because `addLMEntryLevel=true` forces `aname=""`
(`appendColumnNodes` line 479), the resulting Dimensions/Measures folders attach directly under
root with no `TABLE`-typed ancestor, so `BindableFieldsService.collect()`'s untagged fallback
(`node.label()`) is what names the group — reproducing exactly the "Dimensions" duplicate the fix
docs describe.

This matches 04-fix.md's own account almost exactly except for one correction: it isn't that
`addLMEntryLevel`'s loop "receives an empty assemblies list" — it receives a *non-empty* one-element
list (the model placeholder), which is what makes the duplicate a correct-looking sighting of the
real field rather than a garbled one. Confirms the duplicate is pre-existing and unrelated to this
diff, as 04-fix.md claims. Recommend closing the open question in `04-fix.md`/`99-retro.md` with
this trace, and, if someone wants to actually remove the duplicate later, doing it by giving
`addLMEntryLevel`'s placeholder-assembly call the same "is this the model itself, not an entity"
guard this PR already added for the other path — not by touching `BindableFieldsService`.

## Finding 2 (important) — calc field name colliding with a real column can silently vanish again

The new code's per-entity guard is:

```java
if(calc.isDcRuntime() || list.contains(calc) ||
   !lmModelCalcFieldsAdded.add(modelName + "." + calc.getName()))
{
   continue;
}
```

`list` at this point holds the entity's own physical (and already-entity-keyed-calc) columns.
`list.contains(calc)` calls `equals` down to `AbstractDataRef.equals` (`AbstractDataRef.java:303-
318`), which compares by `hashCode0()`/`getName0()` only — i.e. by bare name, not by which table it
belongs to. `CalculateRef` never overrides plain `equals(Object)` (only `equals(Object, boolean)`,
`CalculateRef.java:89-104`), so this is the comparison actually used.

Consequence: if the model-level calc field's name happens to equal an existing column's name in a
given entity (plausible for common ERM attribute names — `Name`, `ID`, `Amount`, `Date`, `Total`,
etc., which frequently recur across entities of the same model), `list.contains(calc)` is `true` for
that entity, so the whole condition short-circuits to `continue` *before* `lmModelCalcFieldsAdded.
add(...)` is ever called (Java `||` short-circuit — the dedup set is only touched on the branch that
reaches it). That's not just "skip this entity" — because the field was never marked added, the
*next* entity is still eligible to add it, so if just one entity's own column set is collision-free,
the field is still surfaced (just possibly under a different entity node than the user expects —
harmless, since `modelOf()`/`fold()` group by the model name stamped on every entity's `TABLE` node
regardless of which one, confirmed by reading `applyTableNodeProperties`, `BaseTreeModelBuilder.java:
262-267`).

But **if the calc field's name collides with an existing column in every entity of the model**, the
field is silently dropped from the tree everywhere, with `ok:true` and no warning at any layer —
reproducing bug #76607's exact symptom (permanently unbindable, no error) for a specific, plausible
naming choice, and *this* PR's own new code is what introduces the silent-drop path (the pre-fix
code never looked up model-keyed calc fields at all, so it couldn't collide this way; now it does,
and the collision guard has no visible-failure fallback).

This was not live-tested this round (no live access). Recommend: name a calc field the same as an
existing attribute shared by the model's entities (e.g. many sample ERM models have a `Name` or `ID`
column on more than one entity) and re-run the charter's repro. If confirmed, the guard needs to
distinguish "this name is already a real column" from "this model-calc-field was already added by an
earlier iteration" rather than using one boolean for both — e.g. skip only when the *dedup set*
already contains the key, and treat a same-named physical column as a distinct problem to surface
(rename hint / warning) rather than a silent skip.

## Finding 3 (important) — the fix's own invariant doesn't hold when an assembly's source has
drifted from the current base model

Per `02-root-cause.md`/`04-fix.md`, the stated fix objective is "a viewsheet-scoped calc field on a
Logical Model source [becomes] bindable" generally, not just for the one traced repro. Tracing where
`!addLMEntryLevel` is actually reachable from:

`VSTreeHandler.checkBaseWSChanged` (`VSTreeHandler.java:169-187`):

```java
if(bentry != null && bentry.isLogicModel() && sinfo != null &&
   !bentry.getName().equals(sinfo.getSource()))
{
   builder.setSourceInfo(null);
   builder.setAggregateInfo(null);
}
```

and `BaseTreeModelBuilder.addCurrentQueryNode(List, Node)` (`BaseTreeModelBuilder.java:168-171`):

```java
if(sinfo == null || sinfo.getType() != XSourceInfo.ASSET) {
   return;
}
```

So whenever an assembly's own stored `SourceInfo.getSource()` no longer matches the *current* base
model's name (still `isLogicModel()`, just a different or renamed model than what the assembly was
bound to) — e.g. the viewsheet's base model was swapped via `attach_base_worksheet` after the chart
already existed, or the assembly's source was set from something other than the current model
somehow — `addCurrentQueryNode(List, Node)` returns immediately and never reaches the per-entity
loop, `addAggregateNode`/`addNormalQueryNode`, or `appendColumnNodes(..., false)` at all for that
tree build. Only `addLMEntryLevel` still runs (it never checks `sinfo`), which is exactly the path
this PR excludes. In that state, the fix's new lookup never fires, and a model-level calc field is
again reachable only through the untagged fallback grouping — the original bug, for a re-based
chart.

This is **not** introduced or worsened by this diff — `addLMEntryLevel` already ran unconditionally
before this patch, and `checkBaseWSChanged` predates it — and it does not affect the reported repro,
since `add_assembly` auto-binds a new assembly's `SourceInfo` to the viewsheet's base name at
creation time (`WizVsService.java:4279-4281`, confirmed by reading), so `sinfo` starts in sync. It's
narrower and rarer than Finding 2 (requires an existing chart plus a later base-model change, not
just a plain calc-field add). But it means the fix's own stated invariant ("makes the field bindable
to any assembly on that model") has a live counterexample the moment the base model is swapped out
from under an existing chart — a sequence a plugin-driven session can produce (attach a different
Logical Model to the same viewsheet later). Not live-tested this round. Recommend a live check
(create a chart, add a calc field so it's bindable, then swap the viewsheet's base model via
`attach_base_worksheet` to a *different* Logical Model, then swap back or check `list_bindable_fields`
against the original chart) before deciding whether this needs its own fix or just its own tracked
gap.

## Finding 4 (important) — no automated regression test for the changed method

`05-verify.md` itself records: "No existing unit test targets `BaseTreeModelBuilder`/
`CubeTreeModelBuilder` directly... the live verification above is this run's only direct coverage of
the actual changed method." Confirmed by search — this diff adds no test file and no test method.
The only coverage is one manual live click-through session that has since been torn down. None of
Findings 1-3 above, nor even the basic "second calc field, second entity" case `02-root-cause.md`
itself live-tested, would be caught by any CI run today; a future refactor of `appendColumnNodes`
could silently reopen #76607 with nothing to catch it. Recommend at minimum one unit test
instantiating `BaseTreeModelBuilder`/`CubeTreeModelBuilder` (or the smallest fixture that reaches
`appendColumnNodes` with `baseEntry.isLogicModel()==true`) asserting a model-keyed calc field is
surfaced via the `addCurrentQueryNode` path and not duplicated when called twice (dedup), before this
is considered fully verified.

## What I looked at and did NOT find a problem

- **Multiple entities / "second calc field, second entity" (explicitly asked)**: not a problem.
  `modelOf()`/`fold()` (`BindableFieldsService.java:123-157`) group by the `table` property stamped
  by `applyTableNodeProperties` onto *every* entity's `TABLE` node for a Logical Model
  (`BaseTreeModelBuilder.java:264-267`), regardless of which entity's node physically holds the added
  calc field ref. Which entity gets picked (first one whose `list.contains(calc)` and dedup-set both
  pass) doesn't change the reported grouping.
- **Two charts on the same model racing to add the same field before either's tree refreshes**: not
  a problem. `lmModelCalcFieldsAdded` is an instance field of one `CubeTreeModelBuilder`/
  `BaseTreeModelBuilder`, and `CubeTreeModelBuilder.getBuilder()` (`CubeTreeModelBuilder.java:54-58`)
  constructs a brand-new instance per tree-build request — confirmed by reading, no static/shared
  state, so no cross-request staleness is possible.
  - **Recomputed builder deduping twice across the same request**: I did not find a way for
    `appendColumnNodes` to be invoked twice for the identical `(model, entity)` pair within one
    request, so `lmModelCalcFieldsAdded`'s dedup never actually gets exercised across more than one
    call in current call graphs I traced — the dedup is defensive rather than currently load-bearing,
    which is fine but worth noting it isn't proven necessary by a test either (folds into Finding 4).
- **Crosstab / calc-table sharing the same code path (explicitly asked, not live-tested per 05-
  verify.md)**: structurally confirmed by reading — the new code sits *before* the `info instanceof
  TableVSAssemblyInfo` branch (`appendColumnNodes`, `BaseTreeModelBuilder.java:482`) that is the only
  thing distinguishing chart layout from table/crosstab/calc-table layout, so it is not gated by
  assembly type at all. No reason to expect a crosstab/calc-table on the same Logical Model to behave
  differently; still recommend one live check before fully closing, as 05-verify.md itself flags.
- **Ordinary (non-Logical-Model) worksheet regression risk**: agree with 04-fix.md/05-verify.md — the
  new block is gated on `baseEntry.isLogicModel()`, unreachable for any non-model table by
  construction.
- **Simpler/more general fix the root-cause doc itself suggests but the diff doesn't take**: no —
  `02-root-cause.md`'s own "Scope of the fix" section already argues against wiring up the
  existing correctly-keyed `addAggregateNode`/`addLMEntryLevel=true` branch (section 4 of that doc),
  since it solves a different sub-problem (reconciling bound fields, not surfacing new ones). Having
  now traced where `addLMEntryLevel`'s own `appendColumnNodes(..., true, true)` call actually gets its
  `assembly` argument from (Finding 1), I don't see a more general fix that avoids the two-call-site
  split without also touching `addLMEntryLevel`'s "which assembly names it" logic, which is a larger,
  riskier change than this PR's scope. Agree the current scoping is a reasonable line, and both the
  "which caller does the check restrict" open question and the "why does the duplicate survive"
  open question this review was asked to chase down both turn out to have the same underlying
  explanation (Finding 1).

## Verdict

**Changes requested.** No blockers — the primary reported symptom is genuinely fixed and the fix's
own scoping logic is sound as far as it goes. But Findings 2-4 are important: a plausible field-name
collision can silently reproduce the exact bug being fixed, the fix's own stated invariant has a
live (if narrow) counterexample once a base model is swapped, and there is zero automated coverage
of any of this. Recommend the fixer address Finding 2 (or confirm via live test that it doesn't
reproduce and explain why not) and Finding 4 (at least one unit test) before this merges; Finding 3
can be a tracked follow-up if a live check shows it's out of scope for how this plugin is actually
driven. Finding 1 is offered as the answer to the open "where does the duplicate come from" question
and does not block.

Nelson-cao:
## Round 1 fix applied — approach changed, not just patched

Thanks for the thorough round 1 review. Finding 2 (name-collision silent drop) was real, and a live discussion with the user surfaced a deeper problem with the original approach's *design*: attributing a model-level calc field to one arbitrary entity's tree node is semantically wrong (it doesn't belong to any single entity) and would show up mis-attributed in the native Composer's own binding tree too.

**The fix moved entirely to the read side.** `BaseTreeModelBuilder.java` is now fully reverted — it turns out it already places a Logical Model's calc field correctly, in an untagged Dimensions/Measures folder directly under the tree root (its actual, correct home). `BindableFieldsService.collect()` just had no way to recognize that folder as belonging to the model, so it fell into the generic "no table ancestor" fallback and read as a separate, unbindable pseudo-table named "Dimensions". The fix folds that untagged folder into the model's own group instead.

This also resolves Finding 2 by construction rather than by patching around it: the untagged folder's own field list only ever contains calc fields (no ordinary columns pass through that code path at all), so there's nothing for a calc field's name to collide with.

Finding 3 (base model swapped under an existing chart) is unaffected by this change either way — still a narrow, pre-existing, out-of-scope case, tracked in the archive rather than fixed here.

Finding 4 (no automated test): added `BindableFieldsServiceTest.foldsAnUntaggedRootFolderIntoTheModelForAScopedCallOnALogicModelBase`, exercising the new fold path directly.

Also live-verified (found mid-review) that this fix covers a second calc field added *after* the chart already has bindings — a shape the previous write-side attempt would NOT have covered without a separate change to `addAggregateNode`; this read-side fix handles it with no extra code.

Full writeup: docs/teams/2026-09-14-bug-76607-calc-field-logical-model/04-fix.md and 05-verify.md in stylebi-wiz (branch docs/bug-76607-deep-debug).

Nelson-cao:
## Round 2 review (redesigned, read-side-only fix)

This is round 2. Round 1 reviewed a different, larger diff touching
`BaseTreeModelBuilder.java` (write side) and found real problems (name-collision
silent-drop, no tests, model-level field misattributed to one entity). The fix was
completely redesigned in response — this diff touches **only**
`BindableFieldsService.java` and its test; `BaseTreeModelBuilder.java` has zero
changes (verified myself: `git diff origin/stylebi-wiz-main-rebase...HEAD --stat`).

**Verdict: no blockers.**

- The fix keys on a general property (`baseEntry.isLogicModel()`), not any repro
  literal.
- Round 1's name-collision risk is structurally impossible in this design — I
  traced *both* reachable call sites that produce the "untagged folder" shape
  (`appendColumnNodes`'s and `addAggregateNode`'s `addLMEntryLevel=true` branches,
  not just the one the fix doc cites) and confirmed both are exhaustively
  calc-field-only; no ordinary column can ever reach either.
- No double-counting between the entity-keyed and model-keyed calc-field lookups —
  the two key spaces are disjoint by construction (that disjointness is literally
  bug #76607's own root cause).
- Unscoped `list_bindable_fields()` (`assembly == null`) is unaffected: `modelName`
  provably stays `null` on that path, and the unscoped tree has a different node
  shape that never reaches the new branch regardless.
- Crosstab live-verified in `05-verify.md`; calc-table not live-tested but
  structurally identical, same as round 1 left it.

**Finding 1 (important, non-blocking)** — traced *why* the "chart already has
bindings" scenario works, since this was left unexplained in `05-verify.md`. It's
not a property of this fix: `CubeTreeModelBuilder.isValidAggregateInfo()` /
`fixAggregateInfo()` is a pre-existing, general (not calc-field-specific) mechanism
that auto-rebuilds a Logical-Model chart's `AggregateInfo` to include every column
of the model (bound or not) essentially any time the chart isn't bound to literally
every column the model exposes — which is also why this rebuild **mutates the
chart's live `VSAssemblyInfo` as a side effect of what looks like a read-only
listing call** (`VSBindingTreeService.getBinding` passes the live object, no clone).
This is real, pre-existing, and untouched by this PR — not a regression to fix
here — but it was untraced anywhere in the archive and is surprising enough to
deserve its own follow-up ticket independent of #76607.

**Finding 2 (important, non-blocking)** — the new test
(`foldsAnUntaggedRootFolderIntoTheModelForAScopedCallOnALogicModelBase`) orders its
tree nodes entity-before-folder, but the real tree
(`CubeTreeModelBuilder.getModel()`) builds the untagged folder *before* any entity
node. `fold()`'s linear scan should be symmetric either way, and I don't have a
concrete counterexample — but the test currently gives no protection against a
future `fold()` change that assumes the entity always creates the group first.
Recommend adding (or flipping) a case with the folder node ordered first.

Full findings: `docs/teams/2026-09-14-bug-76607-calc-field-logical-model/06-review-r2.md`
in the stylebi-wiz repo.

Nelson-cao:
## Round 2 fix — addressed Finding 2, Finding 1 tracked as follow-up

Thanks — round 2 confirmed the redesign closes round 1's collision risk by construction (verified for both reachable call sites, not just one) and traced through why the "chart already has bindings" case is covered without a separate `addAggregateNode` change (it rests on a pre-existing, general `CubeTreeModelBuilder.isValidAggregateInfo()`/`fixAggregateInfo()` mechanism, unrelated to this PR).

**Finding 2 (test ordering)**: added `foldsCorrectlyWhenTheUntaggedFolderArrivesBeforeTheEntityNode`, mirroring the case already there but with production ordering (untagged folder before the entity node — `addLMEntryLevel` runs before `addCurrentQueryNode` in `CubeTreeModelBuilder.getModel()`). Passes — `fold()` is genuinely order-symmetric, not accidentally correct in the one order already tested.

**Finding 1 (list_bindable_fields silently mutating live chart state as a side effect)**: real and worth its own investigation, but pre-existing and unrelated to this PR's own change — not fixing it here. Tracking as a follow-up in the archive (`docs/teams/2026-09-14-bug-76607-calc-field-logical-model/99-retro.md`, stylebi-wiz).

No blockers from either round. Ready to merge as far as this run's own review is concerned.

fredaan (approved):

PR #5210, opened 2026-09-14
fix(wiz): correct pageIndex off-by-one at y=0 in print layout (VAR-002, Redmine #76629)

Zxm-happy:
## Summary

- `VSLayoutService.getPageNumber` (both the `(int, int)` and `(VSAssemblyLayout, Dimension)` overloads) computes the wiz-surfaced `pageIndex` field with a 1-indexed ceiling-division formula, `Math.ceil(y / (float) pageHeight)`. That formula is correct everywhere except exactly `y = 0`, where `Math.ceil(0 / h) == 0` instead of the correct first-page value `1` — every other object in the same physical first page (no matter how small a positive `y`) already reads `1` correctly.
- This is a **`pageIndex`-field-reporting fix, not a print/render pagination fix**. The actual PDF/print pagination pipeline (`VsToReportConverter.getPageNumber`) uses a separate, integer-truncating-division method that does not share this degenerate case, and is only ever consumed via relative comparisons, never surfaced as an absolute page number — confirmed unaffected.
- Fix: both overloads now special-case `y <= 0` (and `pos.y <= 0`) back to page `1`, leaving the already-correct/tested exact-multiple-of-`pageHeight` boundary (e.g. `y = pageHeight` still reads page `1`, a deliberate boundary-inclusive convention) untouched.

## Root cause

`inetsoft.web.composer.vs.controller.VSLayoutService.getPageNumber` is called from `LayoutReadService.toObjectModel` to compute the `pageIndex` field returned by the wiz `get_layout` tool. The ceiling-division formula it uses degenerates only at `y = 0`.

## Test plan

- [x] Added `LayoutReadServiceTest.getReportsPageOneForAnObjectAtTheExactPageOrigin` — an object at layout position `y=0` now reads `pageIndex() == 1`.
- [x] Added `VSLayoutServiceTest.getPageNumberIntOverloadHandlesTheOriginBoundary` and `getPageNumberLayoutOverloadHandlesTheOriginBoundary` — direct unit coverage of both `getPageNumber` overloads (this file previously had zero coverage of either), covering `y=0`, `y=1`, `y=pageHeight` (boundary-inclusive, unchanged), `y=pageHeight+1`, and `pageHeight=0`.
- [x] `mvnw.cmd -o -pl core test -Dtest=LayoutReadServiceTest` — 11/11 passed.
- [x] `mvnw.cmd -o -pl core test -Dtest=VSLayoutServiceTest` — 9/9 passed (7 pre-existing + 2 new), no regressions in the pre-existing tests that exercise the classic composer's print-layout auto-arrange/push-down assist (which also calls `getPageNumber` internally).


Zxm-happy:
## Review round 1 -- VAR-002 (Redmine #76629, stylebi PR #5210)

**Head commit reviewed:** `857f50457` (`fix(wiz): correct pageIndex off-by-one at y=0 in print layout (VAR-002, Redmine #76629)`)

### Diff against the confirmed root cause

- `VSLayoutService.getPageNumber(int, int)` and `getPageNumber(VSAssemblyLayout, Dimension)` both changed from `(int) Math.ceil(y / (float) pageHeight)` to `y <= 0 ? 1 : (int) Math.ceil(...)`. This matches the diagnosis's proposed fix verbatim, in both overloads.

### Scenarios walked (adversarial pass)

1. **Keys on a literal from the repro?** No. `y <= 0` is a general boundary condition covering the entire degenerate-and-negative domain, not a special case for one reported y value.
2. **`y = pageHeight` exact-multiple boundary left untouched, and is that a genuine non-bug?** Re-derived by hand: `Math.ceil(pageHeight / (float) pageHeight) = Math.ceil(1.0) = 1`, same as before the fix -- this is a well-defined ceiling of a positive quotient, not the `Math.ceil(0)` degenerate case. Confirmed via the new `VSLayoutServiceTest` assertions too (`y=pageHeight -> 1`, documented as boundary-inclusive convention). No defect here; correctly out of scope.
3. **Secondary exposure (classic Composer push-down/auto-arrange, `VSLayoutService.java` internal callers ~585/598/666/671/763/798-799/870, reached via `VSLayoutControllerService`):** PR description says only "no regressions in the pre-existing tests that exercise [that path]" -- it does not claim to have fixed or tested that path's y=0 behavior. Checked: none of `VSLayoutServiceTest`'s 7 pre-existing test names (`bottomTabsSingleChildFillsContentBand`, `bottomTabsMultipleChildrenFlushWithTabBar`, `bottomTabsShortChildFlushWithTabBar`, `scriptSetBottomTabsPositionedCorrectly`, `scriptOverrideToNonBottomTabsPositionedCorrectly`, `nonBottomTabsTopNotShifted`, `findViewsheetLayoutReturnsEmptyForUnconfiguredPrintLayout`) exercise an object at y=0, so "no regression" is an accurate, non-overclaiming statement. This matches the diagnosis's own scoping (flagged, not independently confirmed, correctly left as a follow-up).
4. **Same defect pattern elsewhere in the repo?** Grepped all `Math.ceil(...height...)` call sites under `core/src/main/java`. The only page-number-style absolute computation is the two overloads already fixed; every other hit (`CoordinateHelper`, `FontInfo`, `TablePaintable`, `VSTableLens`, `PainterElementDef`, `AFontMetrics`, `Bounds`, `DefaultAxis`) is unrelated geometry/font math, not a page-index computation. No other site of "same bug, other site."
5. **Rendering/print pipeline (`VsToReportConverter.getPageNumber`) untouched, correctly:** confirmed this is a separate int-truncating-division method, only ever consumed via relative comparisons, not an absolute page index -- out of scope and correctly left alone.

### Test verification (independently rerun, not just trusted)

```
cd E:/inetsoft/stylebi-community-worktrees/bug-76629-var-002
./mvnw.cmd -o -pl core test -Dtest=VSLayoutServiceTest,LayoutReadServiceTest
```
Result: `VSLayoutServiceTest` 9/9 passed, `LayoutReadServiceTest` 11/11 passed -- BUILD SUCCESS. Matches the fixer's report exactly (7 pre-existing + 2 new in VSLayoutServiceTest; 10 pre-existing + 1 new in LayoutReadServiceTest).

### Regression test quality

Both new tests assert the exact origin boundary (`y=0`/`pos.y=0` -> `1`) plus the adjacent `y=1`, the `y=pageHeight` boundary, `y=pageHeight+1`, and the `pageHeight=0` error case -- not just the single repro point, so the tests would catch a narrower/still-broken fix (e.g. one that special-cased only `y==0` via `==` instead of `<=`, since negative y is also covered by `y <= 0` even though no explicit negative-y test was added -- minor, non-blocking observation, not required given int y is never observed negative in this code path per the diagnosis).

### CI status

- `build`: pass
- `Semgrep Scan` / `Semgrep OSS`: pass
- `license/cla`: pass
- `claude-review`: **fail**, but inspected the check-run annotations directly -- it is an infrastructure failure of the automated review workflow itself ("Claude execution failed: result is_error:true", "Claude result reported subtype success with is_error:true"), not a posted code-review finding. No review comments or code annotations were produced by it. Treated as non-blocking (consistent with a known infra-failure pattern on a sibling PR in this batch).
- Mergeable: yes, no conflicts.

### Verdict

**No blocking findings.** The fix is minimal, keyed on the general boundary condition rather than the repro literal, correctly leaves the exact-multiple boundary and the rendering pipeline untouched, does not overclaim coverage of the out-of-scope classic-Composer push-down path, and has direct regression coverage in both call-site overloads plus the wiz-surfaced `LayoutReadService` integration test. Test results independently reproduced.

Posted as a PR comment rather than a `gh pr review --approve` because the reviewing identity and the PR author share the same GitHub login (`Zxm-happy`) -- GitHub rejects self-approval. Merging directly per the batch's auto-merge authorization.

PR #5211, opened 2026-09-14
feat(wiz/worksheet): expose MV incremental-refresh conditions via set_mv_conditions (Redmine #76626)

fredaan:
## Root cause

The agent-bridge (`inetsoft.web.wiz.worksheet`: `EditRequest`/`WorksheetMutationSupport`/`WorksheetEditService`/`WorksheetAgentController`) had no op or field that read or wrote a worksheet table's MV incremental-refresh condition fields on `TableAssembly`:

- `getMVUpdatePreConditionList`/`setMVUpdatePreConditionList`
- `getMVUpdatePostConditionList`/`setMVUpdatePostConditionList`
- `getMVDeletePreConditionList`/`setMVDeletePreConditionList`
- `getMVDeletePostConditionList`/`setMVDeletePostConditionList`
- `isMVForceAppendUpdates`/`setMVForceAppendUpdates`

These fields already exist on `TableAssembly` (`community/core/src/main/java/inetsoft/uql/asset/TableAssembly.java:181-232`) and are already wired for the human Composer UI via `AssemblyConditionDialogService`/`MVConditionPaneModel`, but no `mcp__composer-chat__*` tool could reach them — Redmine #76626.

## Fix

Adds a single combined `/edit` op, **`set_mv_conditions`**, carrying all five fields at once (mirrors `MVConditionPaneModel`'s own shape), built directly against `TableAssembly` the same way `set_conditions`/`set_post_conditions` already are — **not** through `AssemblyConditionDialogService`, which is untouched and out of scope.

- `EditRequest.java`: five new trailing fields — `mvUpdatePreConditions`, `mvUpdatePostConditions`, `mvDeletePreConditions`, `mvDeletePostConditions` (`List<WorksheetMutationSupport.ConditionNode>`), `mvForceAppendUpdates` (`Boolean`) — appended at the end of the canonical constructor, plus one new compatibility constructor for old callers, following this file's existing per-addition pattern (no field inserted mid-list).
- `WorksheetMutationSupport.java`: extracts `setConditions`' condition-tree-building loop into a shared `buildConditionList()` helper and adds `setMVConditions()` reusing it for all four MV lists. Each condition-list parameter is independent: `null` leaves that `TableAssembly` field untouched; a non-null (even empty) list replaces it — matching `setConditions`' own null/empty convention. `mvForceAppendUpdates == null` leaves the flag unchanged.
- `WorksheetEditService.java`: new `Editor.setMVConditions()`, following the same `requireTable`/`requireFilterable`/`requireConditionFields` precondition pattern as `setConditions`/`setPostConditions`.
- `WorksheetAgentController.java`: new `"set_mv_conditions"` case in the `/edit` op switch.
- `WorksheetModel.java` / `WorksheetReadService.java` (read side): four new `FilterModel` list fields plus `mvForceAppendUpdates` on `TableModel`, populated from the same five `TableAssembly` getters, so `read_worksheet_model` can surface them once the plugin side maps them.

## Contract (for the companion plugin-side PR in `stylebi-wiz`)

- Op: `set_mv_conditions`
- Request fields: `mvUpdatePreConditions`, `mvUpdatePostConditions`, `mvDeletePreConditions`, `mvDeletePostConditions` (condition trees, same shape as `conditions`), `mvForceAppendUpdates` (boolean)
- Null semantics: each condition-list field independent, `null` = leave untouched, non-null (even empty) = replace; `mvForceAppendUpdates == null` = leave unchanged
- Read side: `WorksheetModel.TableModel` gains `mvUpdatePreConditions`/`mvUpdatePostConditions`/`mvDeletePreConditions`/`mvDeletePostConditions` (`List<FilterModel>`, empty when unset) and `mvForceAppendUpdates` (boolean)

## Tests

Round-trips all five fields through the real mutator onto `TableAssembly`, confirms the null-vs-empty-list "leave untouched" semantics, confirms the embedded-table guard, and confirms the read side surfaces what the write side sets (including the empty/unset case):

- `WorksheetEditServiceMutatorsTest`: `setMVConditionsRejectsEmbeddedTable`, `setMVConditionsWritesAllFiveFieldsOntoTheTableAssembly`, `setMVConditionsNullLeavesThatFieldUntouched`
- `WorksheetAgentControllerTest`: `editDispatchesSetMVConditions`
- `WorksheetReadServiceTest`: `readsMVIncrementalRefreshConditions`, `mvIncrementalRefreshConditionsAreEmptyWhenUnset`

`./mvnw -pl core test -Dtest="inetsoft.web.wiz.worksheet.*Test"` — 572 tests, 0 failures, 0 errors (full `inetsoft.web.wiz.worksheet` package, no regressions).

## Out of scope

`AssemblyConditionDialogService`/`AssemblyConditionDialogModel`/`MVConditionPaneModel` (the browser-UI's own path) — untouched, per the diagnosis. The plugin/TypeScript side (`plugin/composer/src/tools/worksheetTools.ts` in `stylebi-wiz`) is a separate PR building against this exact contract.


fredaan:
## Review — bug #76626, StyleBI-side fix

Reviewed against `docs/teams/2026-09-14-bug-76626/bug-76626/01-diagnosis.md` (confirmed root cause) and `02-refute.md` (HOLDS). Read the actual diff (not the fixer's own writeup) plus the surrounding source on `origin/fix-76626-review` for context not shown in the diff.

**Verdict: Approve.** No blocking issues found.

### What I checked

1. **Reaches the real accessor/mutator pairs, not a narrower keying** — `WorksheetMutationSupport.setMVConditions()` calls `t.setMVUpdatePreConditionList/setMVUpdatePostConditionList/setMVDeletePreConditionList/setMVDeletePostConditionList/setMVForceAppendUpdates` directly on `TableAssembly` (the exact pairs at `TableAssembly.java:181-232`), not keyed on a literal table name or fixture. Confirmed by reading `WorksheetMutationSupport.java` on the branch directly.

2. **`WorksheetPreviewService`'s snapshot/restore stays correct** — this PR does not touch `WorksheetPreviewService.java`. Read it directly on the branch: `snapshotConditions`/`restoreConditions` already clone/restore all four MV condition lists generically via a null-safe `cloneConds()` helper (`wrapper != null ? wrapper.clone() : null`), with no truthiness/emptiness assumption. It will keep working correctly now that these lists can be genuinely non-empty — no change needed there, confirmed not broken.

3. **Null-vs-empty semantics** — verified `setMVConditions()`'s guards are `if(x != null) { t.setX(buildConditionList(...)); }`, not `if(x != null && !x.isEmpty())`. An explicit empty list is non-null, so it reaches `buildConditionList`, which returns `null` for an empty node list — i.e. it still clears the field, it just doesn't short-circuit the way `setConditions()`'s guard does. Same net effect, so the "empty list clears" contract holds; the specific bug class the task asked me to check for (truthiness check silently treating empty-as-no-op) is not present.

4. **`EditRequest.java` field placement** — counted the record's field list directly (79 fields total after this PR); the 5 new fields (`mvUpdatePreConditions`, `mvUpdatePostConditions`, `mvDeletePreConditions`, `mvDeletePostConditions`, `mvForceAppendUpdates`) are appended after `extraProperties`, the prior last field — genuinely at the end, not inserted mid-list. New compatibility constructor added following the file's own established per-addition pattern, defaulting all five to `null`. (I initially misread one test's null-count as off-by-one from a tool-output artifact; re-verified byte-for-byte with `git show`/`grep -c` against the actual PR branch — it's correct, 79 args in, 79 fields in the canonical constructor. The `build` CI job — 22m55s, includes compile+test — passing is independent confirmation this all actually compiles.)

5. **`buildConditionList()` refactor is behavior-preserving** — read the extracted method's full body: it's the exact same loop `setConditions()` used to run inline, now returning `cl.isEmpty() ? null : cl` instead of calling the setter directly; `setConditions()` now does `t.setXConditionList(buildConditionList(...))` — identical result for every input. No change to junction-insertion or null/empty node handling.

6. **Read side / null-vs-empty convention** — `WorksheetReadService.readConditions()` already null-guards (`wrapper == null || wrapper.isEmpty() -> emptyList()`), reused unchanged for the 4 new MV fields; `mvForceAppendUpdates` is `boolean` (primitive) on `WorksheetModel.TableModel`, not `Boolean`, so there's no way for a `null` to leak through on the read side.

7. **Test quality** — spot-checked `setMVConditionsNullLeavesThatFieldUntouched` and `setMVConditionsRejectsEmbeddedTable` directly: the null-semantics test genuinely round-trips all five fields, then re-calls with only one field set and asserts the other three lists' sizes *and* the flag are unaffected — this is a real assertion of the "null leaves untouched" contract, not just happy-path plumbing. The embedded-table test confirms the `PairingException` fires before any mutation, for every input shape (including a non-null empty list on one param), and that no field changed.

8. **`requireConditionFields`/`requireFilterable`/pre-post pairing** — `setMVConditions()` calls `requireConditionFields(t, updatePre, false)`/`(t, updatePost, true)`/`(t, deletePre, false)`/`(t, deletePost, true)`, matching the exact `pre=false`/`post=true` convention `setConditions`/`setPostConditions` already use, and the same `requireFilterable` embedded/snapshot-embedded guard. Consistent error handling with the sibling ops.

### Non-blocking notes

- The `claude-review` CI check on this PR is failing, but it's an infra/config issue, not a code finding: the job log shows `is_error:true` after 379ms with `total_cost_usd: 0` — the automated reviewer bot didn't actually run (looks like a missing/empty `ANTHROPIC_API_KEY` in that workflow), not a real review result. Worth a separate look at the repo's CI secrets, unrelated to this PR's content.
- No `.prompt` files or LLM prompt/system strings in this diff (confirmed via the file list) — the "prompt PR needs human merge" gate does not apply here.

Not merging this myself per the review workflow — leaving the merge decision to the team lead.

Nelson-cao (approved):

PR #5212, opened 2026-09-14
fix(wiz): enforce locked flag on move/resize/resize_title/move_from_container (VAR-001, Redmine #76629)

Zxm-happy:
## Summary

- `ViewsheetEditService`'s `move()`, `resize()`, `resizeTitle()`, and `moveFromContainer()` never checked `LockableVSAssembly.islocked()` before delegating to `ComposerObjectService`, unlike `arrange()` (align/distribute), which already filters out locked assemblies (see its existing `:723-736`).
- The human Composer UI only enforces the lock by withholding the draggable/resizable jQuery-UI handles client-side (`editable-object-container.component.ts`'s `isLocked()`), so it never routes through `ViewsheetEditService` and was already unaffected by this gap. The wiz agent's `edit` tool bypasses that UI layer and could move/resize a locked `Image`/`Shape` even though its `locked` flag stayed `true`.
- Adds a shared `requireUnlocked()` guard (mirrors the check `arrange()` already applies) and calls it from all four ops, throwing `IllegalArgumentException` naming the assembly and op when it is locked.

## Scope

- Change is confined to `web/wiz/viewsheet/ViewsheetEditService.java` (the wiz agent bridge layer). `ComposerObjectService` (shared with the manual Composer UI's WebSocket controller, `ComposerObjectController`) is untouched — confirmed no file under `web/composer/` references `ViewsheetEditService`, so this has zero effect on manual Composer usage.
- Other `EDIT_OPS` (add/remove/rename/copy/cut/paste/set_z_index/group*/ungroup) are intentionally left alone, per `LockableVSAssembly`'s own Javadoc scoping the lock semantic to "cannot be drag".
- Note: `resize_title`'s own `requireTitled()` already rejects every `LockableVSAssembly` type (`Image`/`Shape` are never `TitledVSAssembly`), so in practice no real assembly can hit the new check via `resize_title` today — it's added for consistency/defense-in-depth per the fix proposal's four-site scope, and is covered by a test that forces the combination onto a mock.

## Test plan

- [x] `ViewsheetEditServiceTest`: added regression tests — locked assembly refuses move/resize/resize_title/move_from_container (asserting no downstream `ComposerObjectService` interaction), unlocking allows the ops again, and a non-lockable type (Text) is unaffected.
- [x] `./mvnw.cmd -pl core test -Dtest=ViewsheetEditServiceTest` — 73/73 passing, including existing `align`/`distribute` locked-assembly tests (unchanged).


Zxm-happy:
## Review

Re-read the diagnosis, refute, and fix reports, then independently re-verified against the diff.

**Change**: `ViewsheetEditService.move()`/`resize()`/`resizeTitle()`/`moveFromContainer()` now call a new `requireUnlocked()` guard before delegating to `ComposerObjectService`, mirroring the check `arrange()` already had for align/distribute. Placement is correct in all four (before any mutating call, right after `requireExisting`/`requireTitled`), style matches the file's existing `requireXxx` convention, and the guard is a no-op for non-`LockableVSAssembly` types (Text etc.) — confirmed via a dedicated test.

**Judgment call on the `resizeTitle()` guard being unreachable via real input**: independently re-verified the disjointness claim myself, not just trusting the fix report. `LockableVSAssembly` is implemented only by `ShapeVSAssembly`/`ImageVSAssembly`, and I grepped the *entire* `stylebi-enterprise` checkout (community + enterprise + connectors, not just the community submodule the fix report checked) for any other implementer — none found. `TitledVSAssembly` implementers (Chart/Table-family/CurrentSelection/TimeSlider/Calendar) are a disjoint set. So the two interfaces genuinely never intersect anywhere in this codebase today, and the `resizeTitle()` check is dead code for any currently-existing assembly type.

Landing on **(a): fine to keep as-is.** It's a one-line, zero-behavior-cost defensive check that keeps all four drag-family ops consistent (same guard, same call pattern) rather than leaving one of four silently unguarded if a future `VSAssembly` subtype ever implements both interfaces. The test that exercises it is explicit that it's forcing an unreachable combination via a composed mock, with a comment explaining why — so it won't be mistaken for a real repro path by a future reader. This is a defensible exception to "don't validate what can't happen": the cost here is effectively zero and the alternative (3-of-4 sites guarded, 1 silently not) is a worse trap than the unreachable branch. Not a blocking finding.

**Other checks**: scope is correctly confined to the wiz bridge layer (`ComposerObjectService`, shared with the manual Composer UI, is untouched — confirmed no `web/composer/` file references `ViewsheetEditService`). The other `EDIT_OPS` left out of scope (copy/cut/paste/rename/remove/set_z_index/group*/ungroup) don't call `moveObject`/`resizeObject` and aren't "drag" per `LockableVSAssembly`'s own Javadoc scope, so leaving them alone is correct minimal-scope, not a missed sibling site. No unused imports (all types were already imported), no regressions to the existing align/distribute locked tests, and the new tests assert both the failure (with `verifyNoInteractions`) and that unlocking restores the ops.

**CI**: `build` passing, Semgrep clean, CLA signed, no conflicts (`mergeable: MERGEABLE`). The `claude-review` bot check failed with "Claude encountered an error after 0s" — an infra hiccup on the bot's own side, not a substantive finding (no review comments were posted), and this repo has no branch protection requiring it.

**Verdict: approved. Merging directly (approve blocked by shared reviewer/author login).**

PR #5213, opened 2026-09-14
fix(wiz): thread caller-supplied text through layout-object add/move_resize (VAR-003, Redmine #76629)

Zxm-happy:
## Summary

`LayoutMutationService.addObjects` created every new layout-only text object via `VSLayoutService.createVSAssembly`, which hardcodes `setValue("text")`/`setTextValue("text")` — no field in the per-object request map was ever read for content, so there was no way to set what a layout-only text object (e.g. a print-layout confidentiality label) actually displays. `moveResize` had no way to update that content afterward either, only position/size.

## Root cause

This is a genuine gap, not a wiring oversight: the REST endpoint/request shape passes `objects` as `List<Map<String, Object>>` (arbitrary keys), but `LayoutMutationService`'s own reads only ever looked at `name`/`x`/`y`/`width`/`height`/`type`.

An existing, non-hazardous precedent for mutating a layout-only text object's `TextVSAssemblyInfo` directly (no `apply()`) already exists at `VSTextService.changeText`'s print-layout overload (`core/src/main/java/inetsoft/web/composer/vs/objects/controller/VSTextService.java:79-110`) — it can't be called wholesale from `LayoutMutationService` (STOMP/dispatcher-dependent, would double-checkpoint against `LayoutSessionService.mutateLayout`'s single-checkpoint-per-call invariant), but it confirms the inline `setTextValue`/`setValue` mutation added here is a safe, precedented pattern in this codebase.

## Fix

- `addObjects`: after creating a new layout-only assembly, if the caller supplied `text` and the created assembly is a `TextVSAssembly`, apply it via `setValue`/`setTextValue` instead of leaving the hardcoded literal. Reject `text` against a non-text type (`image`/`pagebreak`) with a named `IllegalArgumentException`.
- `moveResize`: extend the existing `instanceof VSEditableAssemblyLayout` branch (which already syncs position/size onto the embedded `VSAssemblyInfo`) to also apply an optional `text` field when that info is a `TextVSAssemblyInfo`. Reject `text` against an object whose info is not a `TextVSAssemblyInfo` (an existing-assembly placement, or a non-text layout-only object) with a named error instead of silently dropping it.

## Test plan

- [x] Added `addWithTextSetsTheCallersTextOnTheCreatedTextAssembly`, `addWithTextRejectsANonTextType`, `moveResizeWithTextUpdatesAnExistingLayoutOnlyTextObject`, `moveResizeWithTextRejectsANonTextObject` to `LayoutMutationServiceTest`
- [x] `cd core && ../mvnw.cmd test -Dtest=LayoutMutationServiceTest` — 15/15 passing (11 pre-existing + 4 new)

## Companion PR

wiz plugin side (schema/forwarding): https://github.com/inetsoft-technology/stylebi-wiz/pull/new/fix/76629-var-003 (opening next, will cross-link once created)


Zxm-happy:
Companion wiz PR now open: inetsoft-technology/stylebi-wiz#2487

Zxm-happy:
## Round 1 review — bug-VAR-003 (Redmine #76629), stylebi/community half

**Head commit reviewed**: `eb3a6fd93f9133482dfda39b84d7f430d759fb52`

Reviewed together with the companion wiz PR (inetsoft-technology/stylebi-wiz#2487) as one bug per the diagnosis/refute/fix docs — the feature doesn't work with only one side.

### What was checked

- `addObjects`/`moveResize` text-application logic read in full against `01-diagnosis.md`/`02-refute.md`/`03-fix.md`.
- Whether the fix keys on the repro literal (`"ConfidentialLabel"`) — it does not; both methods dispatch on `instanceof TextVSAssembly`/`TextVSAssemblyInfo`, generic.
- Whether `text` is rejected on a non-text target in both `addObjects` and `moveResize`, with a real error rather than a silent drop (the VAR-004 defect class).
- Wire-shape composition against the wiz-side PR: `object.text` (string) forwarded by the wiz tool matches `object.get("text")` read here exactly — same key, same nesting inside the `objects` array entries. Confirmed correct.
- Persistence path (in-`mutateLayout`-lambda mutation, no `apply()`) — matches the class's own Hazard-1 documentation and the existing position/size pattern.
- Tried to build a counter-example around every other call site of `addObjects`'s `existAssembly` branch and `moveResize`'s two branches.

### Finding (blocking — important)

**`addObjects` silently drops `text` when the target is an existing real assembly, instead of rejecting it.**

`core/src/main/java/inetsoft/web/wiz/viewsheet/LayoutMutationService.java`, `addObjects` (~line 290-308):

```java
VSAssembly assembly = masterVs.getAssembly(name);
boolean existAssembly = assembly != null;

if(!existAssembly) {
   event.setType(parseAssetType(object.get("type"), name));
   assembly = vsLayoutService.createVSAssembly(event, layout, masterVs, name);

   if(object.containsKey("text")) {
      ...
   }
}
```

The entire `text` read/apply block lives inside `if(!existAssembly)`. When `op:"add"` is used to place an **existing** viewsheet assembly onto the layout (the wiz tool's own documented usage: *"Adding an assembly already on the viewsheet needs nothing else"* — `type` is optional and ignored in this path since `parseAssetType` is only called under `!existAssembly` too), there is no check at all for `object.containsKey("text")`. A caller supplying `text` alongside an existing assembly's name — including the entirely plausible case of re-placing an *existing* `TextVSAssembly` and also trying to update its text in the same call, with `type:"text"` matching its real type so the wiz-side client guard doesn't block it — has that `text` silently ignored. The assembly is placed, no error is raised, and the result looks successful but the text is simply not applied.

This is the same defect class the finding was asked to check `moveResize` against, and `moveResize` correctly handles it (the `else if(object.containsKey("text")) throw ...` branch at ~line 257-260 for the non-`VSEditableAssemblyLayout` case). `addObjects` needs the mirror: when `existAssembly` is true and `object.containsKey("text")`, throw a named `IllegalArgumentException` (e.g. *"'text' only applies when creating a new layout-only text object via add, not when placing an existing assembly"*), rather than silently dropping it.

**Regression test needed**: `LayoutMutationServiceTest` — `add` with `text` against an **existing** assembly name (real assembly already on the viewsheet, e.g. reusing `"Table1"` or an existing text assembly) must throw, not silently place the object with `text` dropped.

### Not a finding, verified clean

- `moveResize`'s two text-rejection branches (non-text-info editable object, and non-editable/existing-assembly placement) are both correct and symmetric — no gap there.
- Wire-shape composition between the two repos is correct.
- The `text` handling does not key on the repro's specific name/id.

Do not merge yet. The fix belongs in this repo (`addObjects`, stylebi/community side) — no change needed on the wiz PR for this finding.

(Note: the `claude-review` GitHub check currently shows `fail` on this PR, but the run log shows it failing during checkout/cleanup with no review content produced — looks like CI/workflow infra noise, not a real code finding. Worth someone confirming separately, but not treated as a review finding here.)

Zxm-happy:
Fixed the blocking finding in 5032bb804.

`addObjects` now rejects `text` when `existAssembly == true` (placing an existing viewsheet assembly onto the layout via `op:"add"`), mirroring `moveResize`'s existing `else if(object.containsKey("text")) throw ...` pattern for its own non-editable/existing-assembly case — same tone, names the assembly and why `text` doesn't apply there:

```java
else if(object.containsKey("text")) {
   throw new IllegalArgumentException(
      "edit_layout_objects add: \"" + name + "\" is an existing viewsheet " +
      "assembly -- \"text\" only applies when creating a new layout-only text " +
      "object.");
}
```

Added `addWithTextRejectsAnExistingAssemblyPlacement` to `LayoutMutationServiceTest`, which places `"Table1"` (a real assembly already on the master viewsheet — `existAssembly == true`) via `add` with `text` set, and asserts the call throws with `"Table1"` in the message and the existing object's layout position is left untouched.

`cd core && ../mvnw.cmd test -Dtest=LayoutMutationServiceTest` — all 16 tests green (15 prior + 1 new), no regressions.

Zxm-happy:
## Review round 2 — approved

Verified the round-1 fix (commit `5032bb804`):

- The new `else if(object.containsKey("text")) { throw ... }` in `addObjects` is a true `else` of `if(!existAssembly)` — not an independent `if`. Confirmed the three cases compose correctly:
  - `existAssembly == false`, no `text`: creates the object normally, unaffected.
  - `existAssembly == false`, `text` present: existing inner check (type-guard + set) still runs.
  - `existAssembly == true`, no `text`: falls through untouched to the existing "place an existing assembly" code below — the ordinary, already-working case is not affected by this change.
  - `existAssembly == true`, `text` present: now throws, matching `moveResize`'s established pattern.
- Ran `LayoutMutationServiceTest` locally: `Tests run: 16, Failures: 0, Errors: 0, Skipped: 0` — independently confirmed green.
- The new regression test (`addWithTextRejectsAnExistingAssemblyPlacement`) dispatches generically off `existAssembly` resolved via `masterVs.getAssembly(name) != null`, not off the `"Table1"` literal — it isn't overfit to one assembly name.
- Re-checked round 1's other clean findings (widened `move_resize` empty-patch guard, `moveResize`'s two rejection branches, wire-shape composition with wiz#2487) — untouched by this fix round, still correct.
- CI: `build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla` all pass. `claude-review` fails the same way as round 1 (`ANTHROPIC_API_KEY` empty in the Action env, no review content produced) — infra noise, not a code finding.

Merging both stylebi#5213 and the companion stylebi-wiz#2487 now that this side is fixed.

PR #5214, opened 2026-09-14
feat(admin): add repository asset export/import admin-chat area, community tier (Redmine #76604)

lizhepu:
## Summary
- Adds a new community-tier `inetsoft.web.admin.ai.repository` package (`AdminAssetExportController`/`Service`, `AdminAssetImportController`/`ApplyService`, `RepositoryImportChangePlanService`, `RepositoryAssetSelectionResolver`, plus DTOs) — wraps the existing `ExportAssetService`/`ImportAssetService`/`DeployService` (all `community/core`, no changes to those classes).
- Export is one synchronous MCP tool call (resolve → filter → dependent-asset computation → create → download) — no preview/apply, matching `backup_storage`'s read-only precedent. Internally split into 2 Java endpoints (JSON metadata + a separate binary download stream) since `wizClient.downloadToFile` can't carry both, but this stays one plugin tool call.
- Import goes through the plugin's full stage/preview/apply/`planHash`/`taskToken` contract (per the reconciled design decision — this plugin already closed a confirm-then-swap gap for Schedule Tasks via `taskToken`, and import shouldn't reopen it with an acknowledgement-only shortcut).
- Folder-to-viewsheet/worksheet expansion uses `AssetRepository.getAllEntries` (a genuine `community/core` primitive) rather than anything in `enterprise` — `community/core` has no dependency on `enterprise` at all (confirmed via `pom.xml`), so `AdminViewsheetController`/`ViewsheetApiService` were never reachable here regardless of design-doc assumptions.

## Known, disclosed limitations (see `docs/teams/2026-09-14-viewsheet-parity-76604/track-d-import-export/04-build.md` in `stylebi-wiz` for full detail)
- `willOverwrite`/`willCreate` is genuinely new classification logic (built from `AssetRepository.getAssetEntry` existence checks) since the underlying `ExportedAssetsModel.overwriting()` is a single global flag, not a per-target-path diff.
- When `targetFolderPath` retargets an import, `willOverwrite`/`willCreate` check each asset's ORIGINAL path, not the retargeted one — `DeployManagerService`'s retargeting computation is private with no reusable entry point, and this cut was constrained to not touch `DeployService` itself.
- `currentLastModifiedTime` on `willOverwrite` entries is always `null` (no direct timestamp-by-entry lookup exists) — doesn't affect the overwrite/create classification or the `planHash`.

## Test plan
- `./mvnw -pl community/core compile -DskipTests` — clean, `BUILD SUCCESS`.
- No live StyleBI server exercised — the full multipart→stage→preview→apply chain, folder-recursion against a real populated repository, and cluster-cache behavior across nodes are disclosed as unverified.


lizhepu:
## Review — Track D Java side (round 1)

No blocker/important findings. Verified independently in the `wizard-wt-76604-import-export` worktree (matches this PR's exact commit `e12175441` / outer `47f3de93`):

- **Compiles clean**: `./mvnw -pl community/core clean compile -DskipTests` (full clean build, not incremental) — `BUILD SUCCESS`, only pre-existing deprecation warnings.
- **Disclosure #1 confirmed**: `RepositoryAssetSelectionResolver` uses only `AssetRepository.getAllEntries` — grepped the whole new package for any `enterprise`/`AdminViewsheetController`/`ViewsheetApiService` reference; none found. `community/core`'s `pom.xml` genuinely has no `enterprise` dependency.
- **Disclosure #2 confirmed**: `RepositoryImportChangePlanService.alreadyExists` does a real per-asset `assetRepository.getAssetEntry(entry) != null` check for each `VIEWSHEET`/`WORKSHEET` entity — not reusing `ExportedAssetsModel.overwriting()`.
- **Disclosure #3 confirmed as genuinely disclosed**: the original-vs-retargeted-path limitation is documented in this class's own javadoc AND surfaced through to the plugin's tool description/CLAUDE.md (verified in the companion PR #2488) — not a silently hidden gap.
- **Disclosure #4/#5 confirmed**: `AdminAssetImportApplyService.apply` re-resolves the plan from scratch, compares `planHash`, verifies `TaskAuditToken`, and independently requires `acknowledgeOverwrite` whenever `plan.acknowledgeOverwriteRequired()` — a real server-side anti-drift check, not client-trust.
- **`stagingToken`/`importId` wiring checked**: the controller-generated UUID passed into `setJarFile`/`getJarFileInfo` round-trips correctly — `DeployService.getJarFileInfo` sets `.importId(importId)` on the returned model (line 248), so the plugin's `stagingToken: info.importId` aliasing is sound.

No Java unit tests were added for this new package (disclosed, with reasoning); compile-only verification is what's actually usable given no live server access, and that has now been independently re-run rather than just taken on faith.

Not approving via `--approve` (same GitHub account opened this PR — self-approval restriction). No changes requested.

PR #5216, opened 2026-09-14
Fix: edit_date_range_column cascade into downstream mirror AggregateInfo (WBS-047, Redmine #76627)

fredaan:
## Bug

Redmine #76627 (WBS-047): renaming a date-range column via `edit_date_range_column`
(e.g. changing `QuarterOfYear(Order Date)` to `MonthOfYear(Order Date)`) silently
**drops a downstream mirror's grouping entirely** — a mirror table that grouped by
that date-range column loses its group from `AggregateInfo` the instant the source
column's option is changed, with no error and no warning.

## Root cause (two compounding gaps, per the debug-workflow diagnosis)

1. `WorksheetEditService.editDateRangeColumn` renames the date-range column on the
   target table only. Unlike the Composer's own `rename_column` path
   (`RenameColumnController.renameColumn` → `renameTableColumn`), it never cascaded
   the rename into a downstream mirror's `AggregateInfo`.
2. `AggregateInfo.validate(ColumnSelection)` has no repair path for "same source
   column, only the date option changed" — its only recovery does an exact-string
   match on the stale encoded name, which can never match post-rename, so it falls
   through to **deleting** the group entry on the next column-selection refresh.

The native Composer UI's own "Value Range" dialog
(`ValueRangeService.editOldColumn`) shares this identical gap (confirmed by the
diagnosis) — left untouched here as a follow-up to keep this fix minimal; see
"Scoped down" below.

## Fix

`WorksheetEditService.editDateRangeColumn` now snapshots the column's pre-rename
state and, after applying the rename, invokes the existing, already-tested
`RenameColumnController.renameTableColumn(ws, table, old, new)` cascade — the same
machinery the Composer UI's `rename_column` path already relies on to propagate a
column rename into every downstream `ComposedTableAssembly`/`MirrorTableAssembly`'s
column selection and `AggregateInfo`. No new cascade machinery was written; this
reuses the proven path.

## Test

Added `WorksheetEditServiceMutatorsTest#editDateRangeColumnCascadesIntoDownstreamMirrorAggregateInfo`:
builds a table with a date-range column, mirrors it, groups by the mirror's copy of
that column, changes the source column's date option, and asserts the mirror's
group survives (group count stays 1, not 0) and reflects the NEW option/name.

Confirmed via real (non-mocked) `AssetQuerySandbox`-backed execution:
- **Red** against the pre-fix code path: group count goes 1 → 0 (reproducing the
  reported bug exactly).
- **Green** with the fix applied.

```
cd core
./mvnw -o -pl core -am test -Dtest=WorksheetEditServiceMutatorsTest -Dsurefire.failIfNoSpecifiedTests=false
```
All 247 tests in `WorksheetEditServiceMutatorsTest` pass (no regressions).

## Scoped down

- `ValueRangeService.editOldColumn` (the native Composer "Value Range" dialog) has
  the identical gap per the diagnosis, but is **not** touched in this PR — kept out
  to keep this fix minimal and reviewable. Follow-up opportunity.
- The `AggregateInfo.validate()` repair-capable fallback (recognize "same unwrapped
  source column, different `DateRangeRef` option" and reattach instead of deleting)
  proposed as a belt-and-suspenders backstop was **not** implemented — the cascade
  fix above closes the actual reported bug on its own, and the diagnosis flagged
  this fallback as secondary/optional with a specific correctness gotcha (must match
  on the unwrapped base attribute's name, not entity-qualified equality) that
  deserves its own careful pass if pursued.


fredaan (commented):
# Review: PR #5216 — edit_date_range_column cascade into downstream mirror AggregateInfo (WBS-047, Redmine #76627)

**Verdict: clean, no blockers.** (Posted as `--comment`, not `--approve` — see note at the end.)

## 1. Diff matches confirmed root cause / fix direction — yes

The change is exactly what `03-fix.md` describes: in `WorksheetEditService.editDateRangeColumn`,
clone the pre-rename `ColumnRef` (`originalColumnRef`) before `dateRef` is mutated in place, apply
the rename as before, then call the existing `RenameColumnController.renameTableColumn(ws, t,
originalColumnRef, newColumnRef)` — the same worksheet-wide cascade the plain `rename_column` path
already relies on. No new cascade logic was written; the diff is additive (one new import, one
snapshot line, one guarded call). It does not touch any other branch of `editDateRangeColumn`
(validation, option parsing, hash-invalidation comment/logic above it are untouched).

## 2. Cascade guard condition — correct

```java
if(originalColumnRef != null && !newName.equals(currentName) && ref instanceof ColumnRef newColumnRef) {
   RenameColumnController.renameTableColumn(ws, t, originalColumnRef, newColumnRef);
}
```
`!newName.equals(currentName)` is evaluated against `currentName` captured *before* mutation and
`newName` computed from the new option — so a same-name / no-op option change correctly skips the
cascade (and skips the unnecessary query rebuild it triggers). The `ref instanceof ColumnRef`
checks guard against the (currently unreachable, per the method's own earlier `baseRef == null`
check) case where `ref` isn't a `ColumnRef` wrapper.

## 3. Regression test is real, not tautological — verified

Read the full test body (diff) and independently re-ran it:

```
./mvnw -o -pl core -am test -Dtest=WorksheetEditServiceMutatorsTest#editDateRangeColumnCascadesIntoDownstreamMirrorAggregateInfo -Dsurefire.failIfNoSpecifiedTests=false -q
```
→ exit code 0 (passes), confirming the fixer's claimed green state independently (I did not
re-toggle red, since that requires editing source, which is out of scope for a review-only pass —
but the fixer's own red/green log in `03-fix.md` is consistent with what the code does).

The assertions are not count-only:
```java
assertEquals(1, mirror.getAggregateInfo().getGroupCount(), ...);          // not 0
DataRef groupRef = mirror.getAggregateInfo().getGroup(0).getDataRef();
assertEquals("MonthOfYear(orderDate)", groupRef.getAttribute(), ...);     // reflects the NEW option
```
This rules out the "stale-but-present group survives" false-positive scenario named in the task —
the test would fail if the fix cascaded a rename but left the group pointing at the old
`QuarterOfYear(orderDate)` name.

## 4. Real (non-mocked) AssetQuerySandbox — confirmed

```java
when(rws.getAssetQuerySandbox()).thenReturn(new AssetQuerySandbox(ws, null, new VariableTable()));
```
This is a real `AssetQuerySandbox`, not a Mockito mock — matches `02-refute.md`'s own finding that
`WorksheetEventUtil.refreshColumnSelection` silently no-ops when the sandbox is null/unmocked, which
is exactly the harness gap that let this bug go undetected by every pre-existing test in this file.
Test also seeds real embedded rows with headers matching the declared columns, with a documented
reason (avoid the real sandbox's column reconciliation renaming `total` to a positional
placeholder) — a sign the author understood the real-sandbox wiring's side effects, not just
copy-pasted it.

## 5. Scope-down decisions — reasonable, and disclosed

Both deferrals (`ValueRangeService.editOldColumn` native-UI gap left unfixed;
`AggregateInfo.validate()` repair-capable fallback not implemented) are:
- Justified in `03-fix.md` with concrete reasons (different test infra needed for the UI path;
  the validate() fallback is a broader, more invasive change with a real correctness gotcha —
  entity-qualified vs. name-based matching — that deserves its own pass).
- **Also disclosed in the PR body itself** under "Scoped down" (not just the internal fix doc), so
  a human merger isn't surprised later. This satisfies the check.

## 6. Nested mirrors (mirror-of-a-mirror) — already handled, not a new edge case

Read `RenameColumnController.renameTableColumn` (lines 267–328) directly: it walks
`ws.getAssemblies()`, and for any `ComposedTableAssembly` whose `getTableNames()` includes the
renamed table, it recurses — `renameTableColumn(ws, composedTable, column, ncolumn)` — passing the
just-updated composed table back through the same walk, so a second-level mirror that lists the
first-level mirror as one of its own tables is caught on the next iteration of the same loop. This
is pre-existing, already-exercised machinery (247 passing tests in this file, unrelated to this
PR's own new test) — reusing it means nested mirrors are handled "for free," not a gap this PR
introduces or needs to test separately.

## 7. CI status

- `Semgrep OSS` / `Semgrep Scan` / `license/cla`: pass.
- `build`: pending at review time — not blocking per task instructions, but should be green before
  merge.
- `claude-review`: **fail** — but the annotations show this is an *infrastructure* failure of the
  automated review bot itself (`"Action failed with error: Claude execution failed: result
  is_error:true"`, `"Claude result reported subtype success with is_error:true (run did not
  complete successfully)"`), not a content-based rejection of the diff. No actual review findings
  are attached. Flagging for visibility, not treating as a blocker — consistent with this being a
  known recurring infra issue this batch has hit before.

## Note on approval

The PR author (`fredaan`) is the same account this review is running as, so `gh pr review
--approve` is refused by GitHub. Posted as `--comment` instead. **A human or a different account
needs to `--approve` and merge** — recommended action is approve-and-merge once the `build` check
goes green; no code-level blockers found.

jshobe-inetsoft (approved):
## Manual review (PR #5216)

The automated `claude-review` check's most recent run failed with an infrastructure/quota error in the CI bot (`is_error:true`, 1 turn, $0 cost — the bot's own Claude invocation never actually ran) rather than a content-based rejection. This is a manual review to take its place.

### Summary

`WorksheetEditService.editDateRangeColumn` renamed a date-range column on its own table but never cascaded that rename into downstream mirrors, so `AggregateInfo.validate()` — which only does an exact-string match on the stale encoded name — silently dropped the mirror's group on the next column-selection refresh. The fix snapshots the pre-rename `ColumnRef` and, after applying the rename, calls the existing `RenameColumnController.renameTableColumn(ws, table, old, new)` cascade — the same machinery the Composer's own `rename_column` path (`RenameColumnController.renameColumn`, lines 232–258) already uses for the identical pattern (clone the old column, apply the change, cascade with old/new). No new cascade logic was introduced.

### Correctness verification

- **Cascade target confirmed**: Traced `renameTableColumn` → `renameTableColumn0` (`RenameColumnController.java:441`), which calls `renameAggregateInfo(ws, mirror, mirror.getAggregateInfo(), column, newMirrorCol)` directly — this is exactly the repair path missing before this fix, and it is pre-existing, already-exercised code (also updates condition lists, sort info, join operators, and column selections, consistent with the plain rename path).
- **Clone captures pre-mutation state correctly**: `ColumnRef.clone()` (`ColumnRef.java:821`) deep-clones the wrapped `DataRef` (`col2.ref = (DataRef) ref.clone()`). `DateRangeRef` has no `clone()` override, so it falls back to a shallow field copy — but since `name` (String) and `dateOption` (int) are copied by value at clone time (before `dateRef.setDateOption`/`setName` mutate the live object), `originalColumnRef` correctly reflects the pre-rename name/option rather than aliasing the live, soon-to-be-mutated object.
- **Guard logic is correct**: `!newName.equals(currentName)` compares the pre-mutation `currentName` against the freshly computed `newName`, so a no-op option change (same encoded name) correctly skips the cascade and its query-rebuild cost. The `ref instanceof ColumnRef` checks are consistent with the method's existing type-narrowing style used elsewhere in the same method.
- **`ws` field scope confirmed**: `ws` is the `Editor` inner class's own worksheet field (`Worksheet ws = rws.getWorksheet()`), the same worksheet `t` (`requireTable(table)`) was resolved from — so `renameTableColumn(ws, t, ...)` operates on the correct model.
- **Import addition is consistent with existing architecture**: `WorksheetEditService` already imports from `inetsoft.web.composer.ws.*` (e.g. `WorksheetControllerService`), so pulling in `RenameColumnController` doesn't introduce a new package coupling.

### Test verification

- `WorksheetEditServiceMutatorsTest#editDateRangeColumnCascadesIntoDownstreamMirrorAggregateInfo` builds a table with a date-range column, mirrors it, groups by the mirror's copy of that column, changes the source's date option, and asserts both that the group **count** survives (1, not 0) and that the surviving group's `DataRef` reflects the **new** name (`MonthOfYear(orderDate)`, not the stale `QuarterOfYear(orderDate)`) — this rules out a false-positive "stale group merely survives" test.
- Wires a real (non-mocked) `AssetQuerySandbox` onto the mocked `RuntimeWorksheet`, which is necessary because `WorksheetEventUtil.refreshColumnSelection`'s post-mutation sweep silently no-ops without a real sandbox — this is why the bug went undetected by the file's other 247 pre-existing tests.
- `build` check (full validation build, which runs this test) is green, confirming the test compiles and passes in CI, not just locally.

### Scope and disclosure

The PR body explicitly discloses two related gaps left out of scope: `ValueRangeService.editOldColumn` (native Composer "Value Range" dialog has the identical bug) and a more invasive `AggregateInfo.validate()` repair-capable fallback. Both are reasonable to defer — the cascade fix closes the reported bug on its own, and the fallback would need its own careful pass (entity-qualified vs. unwrapped-attribute matching, per the PR's own analysis). No hidden scope creep.

### Other checks

- No security concerns — the change only manipulates internal `DataRef`/`AggregateInfo` model objects derived from already-validated column names; no new user input surface or injection vector.
- No backward-incompatibility: the cascade only fires when a rename actually changes the encoded name (guarded), and reuses a cascade path already relied upon elsewhere, so existing single-table (no mirror) behavior is unchanged.
- CI: `build` (Validation Build) SUCCESS, `Semgrep Scan`/`Semgrep OSS` SUCCESS, `license/cla` SUCCESS. `claude-review` FAILURE is an infrastructure error in the bot itself (no actual findings attached), not a content-based rejection — not treated as blocking per this review's instructions.

### Verdict

**Approve.** No blocking issues found; all non-`claude-review` checks pass; no merge conflicts.

PR #5217, opened 2026-09-14
fix(wbs-046): auto-alias unaliased aggregate outputs so re-aggregation fails loud (#76627)

fredaan:
## Bug

Redmine #76627 / WBS-046: `set_group_aggregate` called a second time on the same worksheet table (no mirror), referencing the first call's aggregate output by its raw column name, silently resolved back to the raw, pre-aggregation column instead of failing loud — a plausible-but-wrong result.

This is distinct from the already-fixed sibling defect (`184341cb0` / `036945949`), which covers the case where the first call gave the output an **explicit** alias. WBS-046 is the common case: the first call's aggregate has **no alias** (e.g. `Sum(QUANTITY)` naturally keeps displaying as `QUANTITY` unless the caller bothers to alias it). Since `WorksheetMutationSupport.applyAggregateInfo`'s "First aggregate on this column" branch only called `colRef.setAlias(...)` when `spec.alias() != null`, nothing was ever tracked in `AGGREGATE_OUTPUT_ALIASES` for this case, so the existing `clearAggregateAliases` fix had nothing to clear on a later call.

## Root cause

`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetMutationSupport.java`'s `applyAggregateInfo`, "First aggregate on this column" branch: an un-aliased aggregate's output and its pre-aggregation raw source share the exact same `ColumnRef` with no marker distinguishing "currently an active aggregate output" from "a plain raw column" once no alias was assigned.

## Fix

Auto-generate and track an alias (the same `"_1"`/`"_2"` suffix convention the secondary-aggregate path already uses) whenever the column had no alias at all before this call. This makes the existing `clearAggregateAliases` mechanism always have something to clear, so a caller that reads the updated model between calls (as intended) and references the aggregate by its now-displayed output name on a second, un-mirrored call now hits the same fail-loud `PairingException` path the explicitly-aliased case already gets. Skipped when the column already carries an alias (e.g. from `rename_column`) so a deliberate rename is never overwritten.

## Honest caveat — this fix is partial

I built a real (non-mocked) `AssetQuerySandbox` end-to-end test reproducing the **literal** WBS-046 repro shape: first call aggregates `Sum(C)` with no alias, second call (no mirror) reuses the **raw** name `"C"` directly (not any alias) with `NthLargest(C, n=2)`, grouped differently. Against an in-memory `EmbeddedTableAssembly`, this **already computes correctly** (differentiated per group), both before and after this fix — it does **not** reproduce the "collapses every group to the same grand-total constant" symptom the original live 2026-09-12 finding reported against a JDBC-backed StyleBI datasource (`Examples/Orders`).

Since an embedded table's aggregate is always evaluated in Java post-process and never pushed into a SQL `GROUP BY` (unlike a JDBC-backed table, where `PreAssetQuery.mergeGroupBy()` merges the grouping into the generated SQL), this is evidence — not proof — that the live "collapse to grand constant" mechanism is a **separate, still-open defect** in the SQL GROUP BY merge path, not something this alias-tracking fix touches. I did not have live browser/MCP access in this role to reproduce against a real JDBC datasource, and building a full JDBC-integration test harness was beyond this task's scope. See `bug-wbs046/03-fix.md` in this worktree for the full writeup.

This fix still lands real, verified value: it turns the alias-name-collision trigger condition into the same fail-loud behavior the already-fixed explicit-alias case gets, for any caller that chains off the model's own displayed output name — which is the expected, documented agent workflow (read the model between edits).

## Tests

- `WorksheetEditServiceMutatorsTest#reAggregatingSameTableWithAutoAliasedOutputNowFailsLoudOnChaining` (new) — structural regression: confirms the auto-generated alias, and that chaining on it now throws `PairingException`.
- `WorksheetMutationSupportUnaliasedReaggregateEndToEndTest` (new file) — real `AssetQuerySandbox` end-to-end test for the literal WBS-046 repro shape (see caveat above).
- Full `inetsoft.web.wiz.worksheet` package (`WorksheetEditServiceMutatorsTest` 247/247, `WorksheetAgentControllerTest` 165/165, `WorksheetEditServiceTest`, `WorksheetPreviewServiceTest`, `WorksheetReadServiceTest`, `WorksheetScriptControllerTest`, `WorksheetScriptServiceTest`) — all green, no regressions.
- `ChartGroupAggregatePoisonEndToEndTest` and `WsMergeServiceTest` (related alias/merge mechanisms) — all green.


fredaan (commented):
Reviewed the diff, the fix writeup (`bug-wbs046/03-fix.md`), and the sibling diagnosis/refute docs. Independently verified rather than taken on faith:

**1. Auto-alias scope is correct.** The new branch is `else if(colRef.getAlias() == null)` — it only fires when the column has no alias at all yet, so a column already renamed via `rename_column` (or already explicitly aliased) is untouched. Confirmed the `"_1"/"_2"` suffix convention isn't invented: the identical pattern (`base + "_1"`, then `base + "_" + (++suffix)` looped against `containsColumnNamed`) already exists ~60 lines below in the pre-existing secondary-aggregate path (`WorksheetMutationSupport.java:933-936`), unchanged by this PR.

**2. The new structural test is genuinely red→green, not tautological.** `reAggregatingSameTableWithAutoAliasedOutputNowFailsLoudOnChaining` asserts both halves: (a) `columnByAttribute(t, "amount").getAlias()` equals the auto-generated `"amount_1"` after the first call, and (b) a second call chaining on `"amount_1"` throws `PairingException` whose message contains `"amount_1"`. It also checks a third call reusing the raw `"amount"` name still succeeds, so the fix doesn't strand the column. Ran it myself: 247/247 in `WorksheetEditServiceMutatorsTest`.

**3. "Why a separate test file" claim checks out.** Confirmed `@WizAgentTestSupport` (used by `WorksheetEditServiceMutatorsTest`) wires only `BaseTestConfiguration` + `SwapperTestConfiguration`. `AssetDataCache`'s only `@Bean` definition is in `IntegrationTestConfiguration` (`assetDataCache(...)`, line ~466), which `WizAgentTestSupport` does not include — so the claimed `NoSuchBeanDefinitionException` is exactly what you'd expect from adding a real-query test there. `ChartGroupAggregatePoisonEndToEndTest` genuinely exists (`core/src/test/java/inetsoft/report/composition/execution/`) and genuinely does use `@ContextConfiguration(classes = {BaseTestConfiguration.class, SwapperTestConfiguration.class, IntegrationTestConfiguration.class}, ...)` — the exact config the new end-to-end test copies. Not an excuse; the new file's context config is real and necessary. `WorksheetEditServiceMutatorsTest` has 243 `@Test` methods, so "widening would affect ~90 other tests" undercounts if anything, not oversells.

**4. The "in-memory can't exercise SQL-merge" claim is correct — verified independently, not just taken on the fixer's word.** Traced `EmbeddedQuery.isSourceMergeable()` → returns `false` unconditionally (`EmbeddedQuery.java:166`). `PreAssetQuery.isAggregateInfoMergeable()` short-circuits to `false` whenever `isSourceMergeable()` is `false`, and `mergeGroupBy()`'s `if(!isAggregateInfoMergeable())` branch is the one that does NOT build `GROUP BY` into the `UniformSQL` — it just collects columns for Java-side post-process instead. So an `EmbeddedTableAssembly` (used by the new end-to-end test) structurally never reaches the `UniformSQL`-building loop at the bottom of `mergeGroupBy()` that a JDBC-backed table does. The fixer's "partial, not full" framing is accurate, not overly pessimistic and not understating anything — the in-memory test genuinely cannot see the code path the live JDBC finding implicates.

**5. PR body discloses the partial-fix status clearly.** It has a dedicated "Honest caveat — this fix is partial" section stating plainly that the fix does not reproduce/resolve the live "collapses to grand constant" JDBC symptom, names the suspect code (`PreAssetQuery.mergeGroupBy()`), and points to `bug-wbs046/03-fix.md` for the full writeup. A human merging this would not be misled.

**6. Independently reran the regression tests** (not just trusted the writeup) from the worktree, git-dir/work-tree bound per this batch's convention:
```
./mvnw -o -pl core -am test -Dtest=WorksheetEditServiceMutatorsTest,WorksheetMutationSupportUnaliasedReaggregateEndToEndTest -Dsurefire.failIfNoSpecifiedTests=false
```
Result: `Tests run: 248, Failures: 0, Errors: 0, Skipped: 0` (247 + 1), including the new end-to-end test's printed per-group values `{X=20.0, Y=5.0}` — matches the PR's claimed output exactly, confirming the "does not reproduce the live symptom" claim is real and not fabricated.

**7. CI:** `build` is still pending (not blocking per this run's instructions). `claude-review` failed, but the log shows an infra/automation problem unrelated to this PR's content — `ANTHROPIC_API_KEY` empty in that job's env, `is_error:true` on the very first turn before any review content could be produced. Not a code-quality signal against this PR. `Semgrep OSS`, `Semgrep Scan`, and `license/cla` all pass.

**Assessment: this is a real, independently-verified improvement, and it is honestly scoped.** The auto-alias change is safe (correctly skips already-aliased columns), reuses an existing naming convention rather than inventing one, is covered by a genuine red→green structural test, and the "why a separate file"/"can't exercise SQL-merge" claims both hold up under independent source-level verification, not just narrative trust. The PR body's disclosure is clear enough that a future reader will not mistake this for a full fix of WBS-046.

**Recommendation: merge as-is.** This closes the alias-tracking gap for the aliased-chaining case and leaves a precise, well-documented pointer (`PreAssetQuery.mergeGroupBy()`, the `ColumnRef` dual-identity theory) for the still-open JDBC "collapses to grand constant" defect, which should stay open as a separate follow-up (WBS-046 should not be closed by this PR alone).

(Posting as `--comment`, not `--approve`: I'm the PR author (`fredaan`) in this session, and GitHub refuses self-approval.)

fredaan:
## Live re-verification on a real JDBC-backed table (2026-09-14)

Verified this bug's core mechanism live, using this repo's own plugin MCP tools against a real `ORDERS` table (`ORDERS1`, JDBC-backed via the shared Examples datasource) — not the in-memory `EmbeddedTableAssembly` this PR's own test used.

**Repro**: `set_group_aggregate(groups=[CUSTOMER_ID,EMPLOYEE_ID], aggregates=[Sum(DISCOUNT)])` (no alias), then `set_group_aggregate(groups=[CUSTOMER_ID], aggregates=[Max(DISCOUNT)])` on the same table, no mirror — coarsening the grouping without ever referencing the first call's own (unaliased) output.

**Result**: `CUSTOMER_ID=1` returned `0.25`, matching neither of the first stage's own per-employee sums (`0` and `0.35`). Independently confirmed via a raw, filtered preview: `0.25` is that customer's own raw, pre-aggregation `DISCOUNT` maximum across all their individual orders — the second call silently resolved `DISCOUNT` against the pre-aggregation source, bypassing the first stage's output entirely.

**This confirms the core defect this PR partially fixes is still real and reproducible on a live JDBC-backed table** in exactly the "unaliased, no prior-alias reference" scenario this PR's own auto-alias mitigation doesn't cover (the mitigation only helps once a SECOND call references the auto-generated alias by name; a fresh `set_group_aggregate` call with a new grouping, never naming that alias, as done here, still silently resolves to raw).

One nuance worth noting: unlike the original 2026-09-12 report (every group collapsing to the same grand-total constant), this run produced genuinely differentiated per-`CUSTOMER_ID` values (`0.25, 0.25, 0.1, 0.15, 0.05, ...`) — each one that customer's own raw-row max, not one shared value across all groups. This matches the milder manifestation this PR's own fix writeup found via its in-memory test — but since this run reproduced the SAME milder shape on a real JDBC table (not just in-memory), the fixer's speculation that JDBC-backed execution specifically produces the more severe "collapse to one constant" shape doesn't hold as stated. What actually distinguishes the two severities (formula choice? dimension-collapse shape? something else) remains unsettled — full writeup in `bug-wbs046/04-live-verify-orders.md` in the batch archive.

Doesn't change this PR's own merge-worthiness (the auto-alias fix is still correct and real, just narrower in scope than the full defect) — flagging for whoever picks up the still-open "core symptom" follow-up.

fredaan:
## MAJOR CORRECTION — recommend holding this PR, likely closing it (not merging)

After live-testing this bug's premise against the native Composer UI (at the user's request), the underlying "bug" this PR fixes appears to be a misdiagnosis, not a real defect.

**Decisive test**: in the native Composer's own Group and Aggregate dialog, on a real `ORDERS` table — apply `[CUSTOMER_ID, EMPLOYEE_ID]` grouped `Sum(DISCOUNT)`, then **reopen the same dialog** and change it to `[CUSTOMER_ID]` grouped `Max(DISCOUNT)`. **The native UI recomputes directly from the table's own original data**, not from the first application's result — confirmed by the user directly in StyleBI itself, not just via this plugin's tools.

This means `set_group_aggregate` (the tool name itself: "**set** grouping and aggregate definitions") is a **replace**, not a **chain** — exactly matching the native dialog. There is no "stage 2 built on stage 1" concept anywhere in the product: `read_worksheet_model` reports exactly one `aggregates` object per table, not a pipeline.

Under this corrected model:
- The already-existing alias-fail-loud behavior (`184341cb0`/`036945949`, predating this batch) was never a bug — referencing a name from a now-discarded prior definition correctly fails, because it genuinely isn't in the table's own underlying columns.
- The originally-reported "silently resolves to the pre-aggregation source" behavior is, under this model, simply **correct**: a field name that happens to match a real raw/source column *is* that raw column, and resolving to it is right, not a silent fallthrough.
- The documented "workaround" (`add_mirror` first, then re-aggregate the mirror) was never a workaround for a bug — it's the *only* correct way to build a genuine two-stage aggregation, since a mirror's own raw columns are rebuilt from its source's *output*, which is what actually creates a parent→child relationship between two stages.

**This PR's own auto-alias change is now suspect for a concrete reason, not just "fixes a non-bug"**: it would make a legitimate case fail. If a caller's second `set_group_aggregate` call happens to name a field that's a genuine raw/source column (exactly this session's `DISCOUNT` test), that *should* resolve to the raw column — correct, intended behavior. This PR's auto-alias would turn that correct resolution into a `PairingException`, breaking working functionality to "fix" a misdiagnosed defect.

**Recommend**: do not merge as-is; hold for the human to decide, likely close. Full writeup: `bug-wbs046/05-major-correction-not-a-bug.md` in the batch archive (supersedes this PR's own `03-fix.md`/`04-live-verify-orders.md`).

One thing NOT settled either way: the ORIGINAL report's most severe claim (every group collapsing to the exact same grand constant) was never actually reproduced by anyone in this batch, including this correction's own live test (which got differentiated, individually-correct-under-this-model values). That could be a genuine misread by the original reporter, or could point to something else entirely — not resolved, flagging rather than guessing.

fredaan:
Closing — not a bug. Confirmed via a decisive live test against the native Composer UI (reopening the Group and Aggregate dialog with a new grouping recomputes from the table's own raw data, not from the prior application's result) that `set_group_aggregate` is a replace/redefine operation, not a chainable pipeline — exactly matching this PR's own auto-alias change's target behavior, except that behavior turns out to be correct as-is, not a defect. This PR's auto-alias mitigation would have broken a legitimate case (a field name that happens to match a real raw/source column resolving to that column, which is correct). Full writeup: `bug-wbs046/05-major-correction-not-a-bug.md` in `docs/teams/2026-09-14-bugs-76627-cond-agg-expr/` (branch `docs/76627-batch-archive`).

PR #5209, opened 2026-09-14
Catalog permission.andcondition property (bug 76600 Gap 2)

lizhepu:
## Summary
Redmine #76600, Gap 2 of 3 gaps found in the admin-chat plugin's Permissions area coverage (Gap 1 is the substantive fix, in the companion `stylebi-enterprise` PR; Gap 3 is docs-only, in the companion `stylebi-wiz` PR).

- The org-level "Access requires BOTH User and Role permission" (AND) vs. EITHER (OR) switch is a plain `SreeEnv` boolean property (`permission.andCondition` in source, normalized to `permission.andcondition` on read/write by `PropertiesEngine.computePropertyNameCase`) — not a bespoke Permissions-area concept.
- It was already reachable through the admin-chat plugin's existing Properties-area uncatalogued-property path (`list_properties`/`get_property`/`preview_changes`/`apply_changes`), but showed up unrecognized/high-risk with no description, since it had no catalog entry.
- Adds one entry to `admin-property-catalog.json` (boolean, `risk: "high"` — this switch changes org-wide permission-evaluation semantics, the same class of consequence as other high-risk entries in this catalog, not a low-risk operational toggle).

**Known caveat (not fixed here — inherent to how every never-set catalogued property already behaves):** on a deployment that has never explicitly touched this property, `get_property` returns `exists:"unknown"`, not the true effective default (`false`/EITHER) — the default lives only in the `SreeEnv.getProperty(..., false, true)` call site and is never persisted.

## Companion PRs
- `stylebi-enterprise` (Gap 1, the real work item — widens the Permissions allowlist for Security Actions capability types + adds a discovery endpoint): https://github.com/inetsoft-technology/stylebi-enterprise/pull/816
- `stylebi-wiz` (TS mirror of Gap 1 + Gap 3 docs): https://github.com/inetsoft-technology/stylebi-wiz/pull/2483

## Test plan
- [x] `./mvnw -pl community/core -Pcommunity -Dtest=AdminPropertyCatalogTest test` — 34/34 pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5218, opened 2026-09-14
Bug #76606 (follow-up): populate logicalModel entry's "type" property

NeverStopFMH:
## Summary

Follow-up to #5198. That fix populated `prefix`/`source` on the hand-built
`logicalModel` `AssetEntry` so `resolveDataSourceEntry` could resolve the
model at all — but live re-testing after merge found it incomplete: once
resolution succeeds, `Viewsheet.getWorksheet` builds a `SourceInfo` from
the same entry's `"type"` property via `Integer.parseInt(...)`, which was
never set, crashing with `NumberFormatException` the first time a real
open (e.g. `create_viewsheet(type:"logicalModel", ...)`) was actually
attempted past the initial resolution.

## Fix

One line: `entry.setProperty("type", XSourceInfo.MODEL + "")`, matching
the convention `AbstractAssetEngine.getQueryEntries` already uses at both
its `isDataSource()` and `isLogicModel()` branches when it builds a real,
tree-walk-produced logical-model-sourced entry.

## Test plan
- [x] `ViewsheetAssemblyAgentControllerTest` (93/93) — asserts the probed
      entry's `"type"` property.
- [x] Live-verified via `mcp__composer-chat__create_viewsheet(type:
      "logicalModel", path:"Examples/Orders/Order Model")` against a real
      StyleBI deployment — confirmed the crash before the fix (exact
      `NumberFormatException`), confirmed a working, chart-bindable
      viewsheet after.

Redmine #76606.


NeverStopFMH:
## Review round 1 — PR #5218

**Head commit reviewed:** `04e2a563c` (`Bug #76606 (follow-up): populate the "type" property on the hand-built logicalModel AssetEntry`)

**Note on formal approval:** `gh pr review --approve` was attempted and rejected by GitHub ("Can not approve your own pull request") — this environment's `gh` identity authored the PR. Posting the verdict as a plain comment instead, consistent with this repo's established pattern for the same limitation (see #5197/#5198's own review rounds).

### What was checked

The invariant this diff claims to hold: *every hand-built `AssetEntry` that reaches `Viewsheet.getWorksheet`'s QTL path (i.e. `wentry.isDirectSource()`) must carry a `"type"` property parseable by `Integer.parseInt`, because `Viewsheet.getWorksheet` (`Viewsheet.java:329-331`) reads it unconditionally with no null/format guard.*

Walked this against the scenarios that would break it if the fix were narrower than it looks:

- **Sibling branch (`physicalTable`)** — `isDirectSource()` (`Viewsheet.java:1418-1421`) covers exactly `isQuery() || isPhysicalTable() || isLogicModel()`, so the physicalTable branch of `resolveDataSourceEntry`/`resolvePhysicalTableEntry` reaches the identical `getWorksheet` code path and could plausibly have the same missing-property bug. Traced it: `resolvePhysicalTableEntry`'s returned leaf entry (`ViewsheetAssemblyAgentController.java:1758`, `found`) is **not** hand-built — it comes from `rep.getEntries()`'s real tree walk (`findPhysicalTableEntry` → `AbstractAssetEngine.getPhysicalEntries`/`createPhysicalEntry`), which does set `"type"` to `XSourceInfo.PHYSICAL_TABLE` on every real entry it produces (`AbstractAssetEngine.java:777`, `:1055`). Confirmed by the method's own javadoc (`:1741-1746`), which explicitly says these properties are populated by the repository's JDBC-metadata walk, not hand-built. So the physicalTable branch is not affected by this defect class — no finding there.
- **`worksheet` branch** — `Viewsheet.reloadBaseWorksheet` (`:215-219`) shows `wentry.isWorksheet()` takes a completely different path (`rep.getSheet`), never reaching `getWorksheet`'s `Integer.parseInt`. Unaffected, confirmed by reading, not just assumed.
- **Other entry points** — `resolveDataSourceEntry` is called from both `createViewsheet` (`:245`) and `attachBaseWorksheet` (`:1831`); the fix and its regression test live in the shared method, so both entry points get the fix and the coverage. No caller left behind.
- **Convention consistency** — `AssetEventUtil.getAttributesBySource` (`AssetEventUtil.java:1248-1253`) hand-builds an equivalent `LOGIC_MODEL` entry for a different purpose and already sets `"type"`; this fix brings `ViewsheetAssemblyAgentController` in line with that existing convention rather than inventing a new one.
- **Same-bug-elsewhere sweep** — grepped every hand-built `new AssetEntry(..., AssetEntry.Type.LOGIC_MODEL, ...)` in production code (`VSEventUtil`, `AssetEventUtil`, `LocalDependencyHandler` x2, `LogicalModelService`, and this file). Only this file's `resolveDataSourceEntry` omitted `"type"`; the others already set it or don't reach the QTL path. No other site found.

Counter-example attempted and not found: could not construct an entry-type or call-path variant under `resolveDataSourceEntry`/`resolvePhysicalTableEntry` that still reaches `Viewsheet.getWorksheet` without a `"type"` property after this fix.

**CI note (not a finding):** the `build` check fails on this PR, but on two tests in an unrelated module (`AdminPropertiesControllerTest.distinguishesARealUnsetPropertyFromAnInventedOneOnlyByGuidance`, `.reportsExistenceAsUnknownWhenUncataloguedAndUnset`) that this diff's files never touch. The branch is currently 9 commits behind `origin/stylebi-wiz-main-rebase` (including #5209, which touched the adjacent admin-property-catalog area) — consistent with a pre-existing/environment issue rather than something this diff caused. Will need a rebase before merge regardless, but that's a merge-readiness matter, not a correctness finding on this diff. `claude-review`'s failure is the already-documented missing-`ANTHROPIC_API_KEY` CI noise from prior rounds in this repo.

### Findings

None.

### Verdict: **approved**

fredaan (approved):

PR #5219, opened 2026-09-14
Fix create_viewsheet's browser-follow-along mechanism (4 bugs)

NeverStopFMH:
## Summary

Stacks on #5218. Live-testing #76606/#76617's fixes (once #5218 also
landed) surfaced that `create_viewsheet`'s "the browser will visually
follow along" promise (its own tool docstring) was broken end to end for
**any** call, regardless of source type — four independent bugs stacked
on top of each other, found by actually watching the browser while
driving the tool live rather than only checking the MCP tool's own
response:

1. **`composer-main.component.ts`** — `editAsset.subscribe`'s
   `command.viewsheet` branch called `this.openViewsheet(command.assetId)`
   without passing `command.runtimeId` (the parallel worksheet branch two
   lines below already passed it correctly) — the browser's own
   `RuntimeViewsheetRef` never bound to the already-minted runtime, so
   every WebSocket call needing it (`getBinding`, `populateObjectTree`,
   `refreshViewsheet`) NPE'd on a null id (`Ouch! Argument cannot be
   null: key`, from Ignite's `AffinityKey`).
2. **`SheetOpenService.createViewsheet`** — opened the new runtime as the
   **agent's own** principal instead of the **acting session's (browser's)**
   principal — the same class of bug `openBaseWorksheet`'s own
   `rvs.getUser()` already guards against (see its own comment) — so the
   browser's later attach died on `"Invalid user found"` (two principals
   for the same user differing only by session).
3. **`SheetOpenService.createViewsheet`** — told the browser to open
   `dataSource`'s own identifier (a `LOGIC_MODEL`/`DATA_SOURCE`/`WORKSHEET`
   entry) instead of the newly-minted runtime's own real,
   `TEMPORARY_SCOPE` `Type.VIEWSHEET` entry (`"Untitled-N"`, which
   `openTemporaryViewsheet` already assigns internally) — so the
   browser's own re-open always found a non-viewsheet-typed entry and
   silently no-opped (`VSLifecycleService.openViewsheet`'s own
   `entry.isViewsheet()` guard).
4. **`SheetAgentBroadcastService.broadcastSave`** — sent
   `SetWorksheetInfoCommand` unconditionally for every runtime type, but
   `viewsheet-pane.component.ts` only listens for
   `SetViewsheetInfoCommand.assemblyInfo.name` — a viewsheet's own save
   (unlike a worksheet's) could never update its tab title, even once
   opened correctly.

An earlier attempt at bug 3 reordered `VSLifecycleService.openViewsheet`'s
own entry-validation ahead of its `runtimeViewsheetId` reattach check —
reverted in favor of fixing bug 3 at the actual source (the wrong
identifier being sent in the first place), since `VSLifecycleService` is
shared with the normal Composer UI's own viewsheet-open flow and this
avoids touching it at all.

## Test plan
- [x] `SheetOpenServiceTest` (38/38) — new assertions confirm the browser
      principal is used and the real temporary entry (not dataSource's)
      is sent; `createViewsheetAcceptsExplicitDataSourceFromAViewsheetSession`
      updated for the now-unconditional `getSheetForPairing` call.
- [x] `SheetAgentBroadcastServiceTest` (10/10) — new test confirms
      `SetViewsheetInfoCommand` (not `SetWorksheetInfoCommand`) is sent
      for a `RuntimeViewsheet`, independently reverted-and-reproduced to
      confirm it fails without the fix.
- [x] `ViewsheetAssemblyAgentControllerTest` (93/93) — unaffected,
      confirmed no regression.
- [x] Live-verified against a real StyleBI deployment, full loop: create
      a viewsheet from a logical model and from a physical table via
      `create_viewsheet`, watch the browser tab open with the correct
      auto-generated name and no errors, add a chart and bind real
      columns, confirm it renders in the browser (not just via
      `get_viewsheet_image`), save with a real name, confirm the tab
      title updates to it.

Redmine #76606, #76617 (filed as a follow-up on the same investigation
rather than new bugs, since both are consequences of the same
`create_viewsheet` feature those two bugs already track).


NeverStopFMH:
## Review round 1 — PR #5219

**Head commit reviewed:** `dda053e34` (`Fix create_viewsheet's browser-follow-along mechanism (4 independent bugs)`)

Note: this PR stacks on #5218 — the diff GitHub shows includes #5218's own commit (`04e2a563c`, the `XSourceInfo.MODEL` "type"-property fix). That commit was already reviewed and approved on #5218's own thread; this round only reviews what's new here: `SheetAgentBroadcastService.java`, `SheetOpenService.java`, `composer-main.component.ts`, and their tests.

**Note on formal review:** `gh pr review -r` was attempted and rejected ("Can not request changes on your own pull request" — same account authored the PR). Posting as a plain comment instead, same fallback used on #5218 and in this repo's prior rounds (#5197/#5198).

### What was checked, per change

**1. `SheetAgentBroadcastService.broadcastSave` — type-matched tab-label command.**
Invariant: the tab-label command sent matches the runtime's actual type, because each client component only listens for its own type's command.
- Confirmed `RuntimeSheet` has exactly two concrete subclasses (`RuntimeViewsheet`, `RuntimeWorksheet`) — the `instanceof RuntimeViewsheet`/`else` split is exhaustive, no third sheet type silently falls into the wrong branch.
- Confirmed the worksheet sibling path is untouched (same `SetWorksheetInfoCommand` code, just moved into the `else`).
- Verified the claim end-to-end rather than trusting the PR description: `viewsheet-pane.component.ts:798-799` (Composer's own pane, not `viewer-app.component.ts`, which is a different, unrelated app context) does `processSetViewsheetInfoCommand(command) { this.vs.label = command.assemblyInfo["name"]; }`, and the dispatch mechanism (`command-processor.ts:84`, `"process" + message.type`) confirms this method is actually wired to receive `SetViewsheetInfoCommand` messages, not dead code.
- Swept for other unconditional `SetWorksheetInfoCommand`/`SetViewsheetInfoCommand` sends in `inetsoft.web.wiz` — this is the only site. No same-bug-elsewhere finding.
- No finding.

**2 & 3. `SheetOpenService.createViewsheet` — browser principal + real temp entry.**

Invariant claimed (per the PR's own comment at `SheetOpenService.java:266-269`): mirrors `openBaseWorksheet`'s `rvs.getUser()` pattern to open the new runtime as the browser's principal, not the agent's.

Walked the sibling of the branch the repro took — specifically, the **explicit-`dataSource`** call pattern (`create_viewsheet(type:"logicalModel", ...)` or `type:"physicalTable"`), as opposed to the **default-to-acting-worksheet** pattern the PR's own new test coverage exercises (`createViewsheetDefaultsToActingWorksheetWhenNoDataSourceGiven`) — and found a real regression:

- **`SheetOpenService.java:270-271`**: `runtimeAccess.getSheetForPairing(actingSession.sheetType(), actingSession.runtimeId(), user)` is now called **unconditionally**, before the `dataSource == null` check — moved out of the branch that actually needs it (the old code only called the equivalent `getSheetForPairing` inside the `dataSource == null && actingSession.sheetType() == WORKSHEET` defaulting branch).
- `SheetRuntimeAccess.getSheetForPairing` (`SheetRuntimeAccess.java:87-98`, delegating to `getWorksheetForPairing`/`getViewsheetForPairing`, `:197-241`) **never returns `null` on a stale runtime — it throws `PairingException(SESSION_EXPIRED, ...)`** when `worksheetAccessor.getSheetDirect(runtimeId)`/`viewsheetAccessor.getSheetDirect(runtimeId)` returns null (own javadoc at `:85`: "`@throws PairingException` if the runtime is not found"). The `actingSheet == null ? ... : ...` fallbacks at `:281` and `:298` are therefore effectively dead code in production — they can only be reached if `getSheetForPairing`'s contract changes.
- `actingSession` (a `JoinSession`, resolved via `sheetSessions.resolve`, checked earlier at `:239-244`) and the underlying `RuntimeSheet` (looked up separately via `getSheetForPairing`) have **independent lifecycles**: `SheetSessionService` keeps its own in-memory map with a fixed 30-minute TTL (`SheetSessionService.java:39`) and its own `evictExpired` sweep, entirely separate from the runtime engine's own cache (Ignite-backed `RuntimeSheetCache`, referenced elsewhere in this codebase for its own, independent eviction). A paired session can legitimately still be within its 30-minute TTL (and have a live browser socket — the `actingSession.socketSessionId() == null` guard at `:260-264` only checks the session's own socket metadata, not the runtime) while its own underlying `RuntimeSheet` has already been evicted from the runtime cache.
- **Concrete failure scenario**: an agent has a live paired session to a worksheet or viewsheet whose *runtime* has since been evicted (idle timeout / memory pressure / node restart) while the pairing session itself (separate 30-min-TTL store) is still valid. Before this PR, `create_viewsheet(type:"logicalModel", path:...)` — an explicit-`dataSource` call — succeeded regardless, because the code never touched the acting runtime at all in that branch. After this PR, the same call now throws `PairingException("... runtime not found or expired ...")` and fails outright, purely because of an unrelated, no-longer-needed side effect (fetching the browser's principal) — the new viewsheet's own data source doesn't depend on the acting runtime at all.
- Contrast with the pattern this claims to mirror: `openBaseWorksheet`'s dependency on `viewsheetSessions.resolve` (`SheetOpenService.java:129-131`) is **intrinsic** — it cannot determine the base entry without the live runtime — and it also uses the null-tolerant `resolve` (not the throwing `getSheetForPairing`), degrading to a clean `IllegalArgumentException` instead of a raw `PairingException`. `createViewsheet`'s new dependency is *not* intrinsic when `dataSource` is explicit; it only exists to fetch a `Principal`.
- **Test coverage gap, matching the "regression test asserts only the exact repro input" overfit smell**: `createViewsheetService`'s fixture (around `SheetOpenServiceTest.java:850-870`) now stubs `getSheetForPairing` to **unconditionally return a non-null mock** for every `actingType` — it never exercises what happens when `getSheetForPairing` throws. `createViewsheetAcceptsExplicitDataSourceFromAViewsheetSession` was updated to assert the call now happens (`verify(runtimeAccess).getSheetForPairing(...)`), but never asserts what happens if that call fails for a still-valid pairing session with an expired runtime — the exact scenario this changes.

**Severity: Important.** This is a real regression on a previously-working call pattern (explicit `dataSource`, not just the reported "default to acting worksheet" repro), it's silent until the specific timing window occurs, and neither the code's own defensive-looking null checks nor the updated tests actually cover it.

**4. `composer-main.component.ts` — pass `command.runtimeId` through to `openViewsheet`.**
- Confirmed the sibling worksheet branch (`:2976`, `openWorksheet(command.assetId, false, command.runtimeId)`) already followed this exact convention — the fix brings the viewsheet branch to parity, not a novel pattern.
- Confirmed two other existing call sites (`:473`, `:3138`) use the identical `(assetId, false, runtimeId, ...)` shape.
- Checked the dedup guard just above (`:2948-2949`, matching an existing open tab by `s.id === command.assetId`): with the fix, `command.assetId` is the fresh runtime's own `"Untitled-N"` temp identifier, so this correctly never matches a pre-existing tab for a brand-new create — an incidental improvement over the old behavior (which could have accidentally matched an already-open tab on `dataSource`'s own identifier), not a regression.
- No finding.

### Findings

| # | File | Line(s) | Severity | Summary |
|---|---|---|---|---|
| 1 | `core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java` | 270-271 (interacting with 281, 298) | Important | `getSheetForPairing` is now called unconditionally and throws `PairingException` on an expired/evicted acting runtime, regressing the explicit-`dataSource` call path (which previously never depended on the acting runtime being alive) into a hard failure for a real, plausible timing window (paired session still within its independent 30-min TTL, underlying runtime already evicted). Not covered by the updated tests, which only stub the success case. |

NeverStopFMH:
## Review round 2 — PR #5219

**Head commit reviewed:** `c09fe7b50`
**Range since round 1:** `dda053e34..c09fe7b50` (one new commit: "Fix #76606 PR review finding: tolerate an expired acting runtime for explicit-dataSource create_viewsheet")

**Note on formal approval:** `gh pr review --approve` was attempted and rejected ("Can not approve your own pull request" — same account authored the PR). Posting as a plain comment instead, same fallback as round 1 and #5218.

### Status of round 1's finding

**Finding #1 (`SheetOpenService.java:270-271`, unconditional `getSheetForPairing` regressing the explicit-`dataSource` path): FIXED.**

Verified, not just read:
- The fix wraps the fetch in `try/catch(PairingException e)`. When `dataSource != null` (explicit) and `e.getKind() == SESSION_EXPIRED`, it falls back to `actingSheet = null` (→ `browserUser = user`, the agent's own principal — the exact pre-#5219 behavior for this path) instead of propagating. Any other case (`dataSource == null`, or any other `Kind`, e.g. `USER_MISMATCH`) still rethrows.
- **Independently redid the revert/reapply check** rather than trusting the diff by eye: reverted just `SheetOpenService.java` to the pre-fix commit (`dda053e34`) and re-ran the new `createViewsheetToleratesExpiredActingRuntimeWhenDataSourceIsExplicit` test — it fails exactly as predicted (`PairingException: Viewsheet runtime not found or expired`, thrown from `SheetOpenService.java:270`, uncaught). Reapplied the fix (`c09fe7b50`) and reran the full class: `SheetOpenServiceTest`, 40/40 (was 38/38 before this round; +2 new tests).
- Walked the sibling this fix itself introduces, per this workflow's own standard (a fix can introduce a new defect):
  - `dataSource == null` + `SESSION_EXPIRED` → still rethrows, unchanged from pre-#5219 behavior for the defaulting branch. Confirmed by the new `createViewsheetStillFailsWhenDefaultingAndActingRuntimeIsExpired` test, which asserts the `PairingException` still propagates.
  - `dataSource != null` + `USER_MISMATCH` (a real security-relevant failure, distinct `Kind`) → still rethrows, not silently swallowed. Confirmed by reading the condition (`e.getKind() != PairingException.Kind.SESSION_EXPIRED`) — only `SESSION_EXPIRED` is tolerated.
  - `getWorksheetForPairing`/`getViewsheetForPairing`'s "wrong runtime type at this id" branch also throws `Kind.SESSION_EXPIRED` (same as "not found") — both collapse into the same tolerant fallback here. Not a new problem: the public `PairingException` API gives no finer-grained signal to distinguish them, and both are equally benign to fall back on (proceed as the agent's own principal) rather than fail the whole call.
  - The `LOG.warn(...)` call inside `SheetRuntimeAccess.getWorksheetForPairing`/`getViewsheetForPairing` still fires before the exception is thrown, so the underlying condition remains observable in logs even though `SheetOpenService` no longer propagates it as a hard failure for this one path.
- No new defect found in the fix itself.

### New findings this round

None.

### Verdict: **approved**

fredaan (approved):

PR #5224, opened 2026-09-14
Fix latent client-side collision guard + regression test (bug #76636)

lizhepu:
## Summary
- Bug #76636 was diagnosed against `composer-main.component.ts`'s `editAsset.subscribe` handler: the server used to send the new viewsheet's *base worksheet* id as `command.assetId` (colliding with the already-open worksheet tab's own `Sheet.id`) and the client dropped `command.runtimeId` when calling `openViewsheet`. Both symptoms matched the reported "Failed to open viewsheet: <worksheet runtimeId>" error and the "no new tab appears" behavior.
- `runtimeId` threading was already fixed by #5219, and #5219 also stopped the server from ever sending a colliding `assetId` for `create_viewsheet`. Reviewer feedback on round 1 of this PR (test-only at the time) correctly pointed out that our added regression test used a hardcoded non-colliding `assetId`/`Sheet.id` pair, so it gave **zero coverage** for the collision/stray-socket-event half of the original bug — it would not catch a regression of that half.
- Investigating that finding surfaced a real, previously-undetected gap: **the client handler itself never had a defensive type check.** Driving a genuine collision (an `assetId` equal to an already-open `Worksheet`'s own `id`, i.e. the pre-#5219 server shape) directly into the current handler still force-casts that worksheet tab into a `Viewsheet` and fires `/events/vs/bindingtree/gettreemodel`/`/events/vs/refresh` on its socket — exactly the original bug's symptom. #5219 only changed the *server* to stop sending a colliding id; it never closed this off client-side, so the bug is latent, not eliminated, and would resurface if the server's contract ever regressed.
- **This PR now includes a real fix**, not just a test: `editAsset`'s own `sheets.find` and `openSheet`'s `sheets.findIndex` (`composer-main.component.ts`) now also require `sheet.type` to match, so a `Worksheet` tab can never be mistaken for a `Viewsheet` (or vice versa) by id collision alone, regardless of what `assetId` the server sends.
- Checked all other `OpenComposerAssetCommand` builder sites (`SheetOpenService.openBaseWorksheet`/`createWorksheet`, `ComposerController.editViewsheet`/`editWorksheet`) — none share this wrong-semantics pattern.

## Test plan
- [x] New collision-case test constructed to fail against the pre-fix predicate (verified by temporarily reverting the production change) and pass with the `sheet.type` guard in place.
- [x] `npx ng run portal:test-tl --include "**/composer-main.component.editasset-viewsheet-handoff.tl.spec.ts"` — both tests (non-colliding + colliding) pass
- [x] `npx ng run portal:test-tl --include "**/composer-main.component*.tl.spec.ts"` — all 5 files / 81 tests pass (no regressions)


lizhepu:
## Review (round 1) — comment only, same GitHub account as PR author (account-lane self-approval block)

**Verdict: not clean yet — one important finding + two failing checks.**

Independently re-verified before reviewing the content:
- `gh pr diff 5224` is genuinely test-only: one new file, no production code touched.
- `git merge-base --is-ancestor 7f45a3ade HEAD` / `...origin/stylebi-wiz-main-rebase` both return true — commit 7f45a3ade (PR #5219) really is an ancestor of both the fix branch and the target base, confirming the claimed root-cause fix already landed.
- Read current source directly: `SheetOpenService.java:317-352` and `composer-main.component.ts:2965` both show the fixed shape described in the fix doc.
- Re-ran the suite myself (didn't just trust the write-up): `npx ng run portal:test-tl --include "**/composer-main.component*.tl.spec.ts"` → 5 test files, 80 tests, all passed.

**Important finding:** the new test's own claimed coverage overstates what it protects. It hardcodes a non-colliding `assetId`/`ws.id` pair, so `sheets.find(s => s.id === command.assetId ...)` never matches by construction — `expect(ws.socketConnection.sendEvent).not.toHaveBeenCalled()` is trivially true regardless of whether the collision-lookup logic is correct or reintroduced. This test *does* correctly guard the `runtimeId`-drop half of the original bug (that assertion is independent of the collision setup), but it provides **no regression coverage for the collision/stray-socket-event half**, contrary to the PR description's claim that "if a future change reintroduced either bug ... this test would fail." Since that invariant is really a server-side guarantee, a client-only Angular test may not be able to assert it at all — but the PR description should be corrected to not claim coverage it doesn't have (either drop the collision-half claim, or note the limitation explicitly and point at the Java side as where that regression would need to be caught).

**Non-code blockers:**
- `build` check FAILING — `AdminPropertiesControllerTest` (Java, admin properties) — unrelated to this PR's single Angular spec file, looks pre-existing/flaky, but needs confirming before merge.
- `claude-review` check FAILING — the bot's own Claude invocation errored (`is_error:true`, 1 turn, no review content) — looks like an infra failure, not a real review outcome.
- Semgrep + CLA are green. `mergeable: MERGEABLE`, `mergeStateStatus: UNSTABLE` (no conflicts, just the two failing checks above).

Full findings written to `docs/teams/2026-09-14-bug-76636/bug-76636-viewsheet-handoff/05-review-r1.md` in stylebi-wiz.

lizhepu:
# Bug 76636-A: PR stylebi#5224 review (round 2)

## Verdict: CLEAN on code content. NOT mergeable right now -- two failing checks (both
pre-existing/infra, not caused by this diff). Posted as a PR comment only (account-lane
self-approval block, same as round 1).

Account-lane note: gh pr view 5224 --json author -> lizhepu; gh auth status -> active
account lizhepu (same). Per the account-lane self-approval block, no approve/merge from this
review regardless of verdict -- this is a comment-only round.

## What changed since round 1

Round 1 found the original regression test assetId/ws.id pair was hardcoded non-colliding,
so it gave zero coverage for the collision half of the bug. The fixer (06-fix-r1.md)
investigated, found this was a real latent client-side defect (not just a test gap), and landed a
genuine 2-line production fix plus a second collision-case test. Verifying that claim, end to end,
is this round job.

## What I checked (independently, not taken on the fixer or round-1 word alone)

### 1. Diff actually contains both claimed production changes, at the claimed locations

gh pr diff 5224 confirms exactly two files changed:
- New test file (127 lines): adds a second it(...) block for the collision case, keeps the
  original non-colliding test.
- composer-main.component.ts, two hunks:
  - editAsset sheets.find (approx 2948-2950): s.type equals "viewsheet" AND s.id equals command.assetId
  - openSheet sheets.findIndex (approx 1781-1783): sheet.type equals type AND sheet.id equals id

Both are exactly the one-line add-the-type-discriminator pattern claimed, nothing more.

### 2. Does the guard break any legitimate same-type reopen?

Read openSheet full body and all three callers (composer-main.component.ts:1363
onSheetReload -- passes sheet.type for the sheet being reloaded; :1768 openWorksheet --
hardcodes type: worksheet; :1774 openViewsheet -- hardcodes type: viewsheet). Every
caller passes a type that genuinely describes what it wants to open/match, so requiring
sheet.type equals type only narrows the match to the caller own intended sheet kind -- it cannot
reject a legitimate same-type reopen (worksheet-to-worksheet or viewsheet-to-viewsheet), since in
that case sheet.type and type are the same value by construction. Also confirmed Sheet.type is a
real discriminated field set once in each subclass constructor (Worksheet ctor sets
worksheet, Viewsheet ctor sets viewsheet, sheet.ts:22 union type worksheet, viewsheet, wiz)
-- not something that could be undefined/stale for an already-open tab.
The focusedSheet early-return branch (checks this.focusedSheet and this.focusedSheet.id equals id
and index at least 0) is unaffected in the legitimate case (index still at least 0 when types
match) and correctly stops matching in the collision case (index becomes -1 when types differ, so
the early return no longer fires and a real new tab gets created) -- no new bug introduced.

### 3. Is the second fix (openSheet findIndex) really necessary, not scope creep?

Verified this directly rather than trusting the fixer or round-1 account. In the fix worktree
(E:\inetsoft\wizard-wt-76636-viewsheet-handoff\community), I surgically reverted only the
openSheet hunk (restored the plain sheet.id equals id check with no type check) while leaving
editAsset fix in place, and reran the collision test:

  comp.sheets.length: expected 2, received 1
  (the sendEvent assertion still passed -- bug (a) alone is fixed by fix number 1)

This confirms fix number 2 is independently load-bearing: openViewsheet really does delegate into
openSheet, whose own lookup has the identical missing-discriminator gap, and re-matches the
worksheet tab a second time -- exactly the mechanism 02-refute.md (point 3) already traced for
bug (b). Not scope creep; both call sites share the one root defect pattern and both had to be
fixed for the new collision test (which asserts on both sendEvent and sheets.length) to pass.
Reverted this probe change afterward (git checkout on the file); worktree confirmed clean and
back to the PR actual committed state before re-running the full suite.

### 4. Independently verified test fails against pre-fix code, passes with the fix

Reverted both production hunks (git apply -R on the diff between the fix commit and its
parent) and ran just the collision test: it failed, with ws.socketConnection.sendEvent called
twice -- once with RefreshBindingTreeEvent (destination /events/vs/bindingtree/gettreemodel) and
once with VSRefreshEvent (destination /events/vs/refresh) -- exactly the stray-event symptom from
bug (a), and exactly what the fixer reported. Restored the fix (git apply forward) and reran the
full suite to confirm no residual state.

### 5. Re-ran the full suite myself

  cd E:\inetsoft\wizard-wt-76636-viewsheet-handoff\community\web
  npx ng run portal:test-tl --include composer-main.component pattern for tl.spec.ts files
  -> Test Files 5 passed (5), Tests 81 passed (81)

Matches the fixer claim exactly (80 previous plus 1 new). Ran this twice (once after the fix
number 2 probe, once after the pre-fix-code probe), both clean, worktree left at the exact
committed state (git status --short empty) both times.

### 6. PR title/body accuracy

Title: Fix latent client-side collision guard plus regression test (bug number 76636). Body
clearly states: original test-only framing, round-1 finding, the real latent-defect discovery,
both fix locations, and the fail-pre-fix / pass-post-fix verification method. A future reader
would not be confused about why an add-a-test PR now touches production code -- the body explains
the reason (PR 5219 closed the server-side trigger, never the client-side gap) up front. No
correction needed here (this addresses round 1 separate finding about the old body overclaim,
which no longer applies since the description was rewritten).

### 7. Unconditional prompt-merge gate

Still not applicable -- only composer-main.component.ts (production) and one new test spec file
touched; nothing under wiz-services/src/agents/prompts/ or any prompt-doc.

## Blockers: failing checks (non-code)

gh pr checks 5224:
- build: FAILURE. Read the actual Maven log (gh run view 34841600805 --log-failed):
  inetsoft-core module, AdminPropertiesControllerTest --
  distinguishesARealUnsetPropertyFromAnInventedOneOnlyByGuidance and
  reportsExistenceAsUnknownWhenUncataloguedAndUnset. Identical two test names to round 1 report
  on the previous run of this same PR -- confirms this is the same pre-existing/flaky failure,
  not something newly introduced by the openSheet second fix or anything else added since round 1.
  Entirely unrelated module (Java admin-properties controller vs. this PR Angular-only diff).
- claude-review: FAILURE. Read the actual run log (gh run view 34841600809 --log):
  message reads Internal error, directory mismatch for directory tsconfig.json, is_error true,
  1 turn, no review content produced -- an infra/action failure in the bot itself, not a real
  review verdict. Same characterization as round 1 (this looks like a recurring flake in this
  repo Claude Code Review action, not something the fixer can address from this PR).
- Semgrep (both) and CLA: green.
- mergeable: MERGEABLE -- no conflicts.

Both failing checks are, on direct inspection of their logs, pre-existing/infra and unrelated to
this PR content -- but per the review gate they still block merge as-is. This is not something a
further code change to this PR would fix; it needs either a rerun/infra fix or an explicit human
call that these are known-unrelated flakes before merging past them.

## Summary

Code content: clean. Both production fixes verified correct, necessary (each independently, by
reverting only one at a time), and narrowly scoped -- no regression to legitimate same-type
reopen paths, all three openSheet callers checked. The new collision test is genuine (fails
pre-fix, passes post-fix, both independently reproduced this round, not just re-trusted from
06-fix-r1.md). Full suite (81/81) reproduced twice. PR title/body accurately reflects the
test-only to fix-plus-test scope change.

Not clean to merge yet: build and claude-review checks are both red. Both are, on direct log
inspection, pre-existing/unrelated to this diff (same two Java test names as round 1; same bot
infra error as round 1) -- but that determination is being surfaced here, not acted on
unilaterally. Posting as a PR comment only (same GitHub account as the PR author). This PR is
ready for the user merge decision once the two check failures are resolved or explicitly waived
as known flakes -- no further code-side rework identified.

jshobe-inetsoft (changes requested):
## Manual Code Review — PR #5224

(Posted manually because the automated `claude-review` check failed due to a CI-bot infrastructure/quota error — `is_error:true`, 1 turn, $0 cost — not because it found a blocking issue.)

### Summary of the change
- `composer-main.component.ts`: `openSheet`'s `sheets.findIndex` and the `editAsset` command handler's `sheets.find` now additionally require `sheet.type === type` (or `"viewsheet"`) before matching a tab by `id`. This closes a latent client-side gap: even after #5219 fixed the *server* to stop sending a colliding `assetId` for `create_viewsheet`, the client handler itself had no defensive type check and would still force-cast an unrelated same-id `Worksheet` tab into a `Viewsheet` and fire `/events/vs/bindingtree/gettreemodel` / `/events/vs/refresh` on its socket if the server ever regressed to the old (colliding) id shape.
- Adds `composer-main.component.editasset-viewsheet-handoff.tl.spec.ts` with two cases: (1) non-colliding `assetId` opens a new viewsheet tab with the server-provided `runtimeId`; (2) a genuine id collision with an open `Worksheet` tab does not force-cast/fire on that tab's socket and still opens the new viewsheet correctly.

### Code review findings
- **Correctness**: The `type` values passed into `openSheet` (`"worksheet"`, `"viewsheet"`, `"wiz"`) are the same string-literal discriminator used pervasively elsewhere in this file (`sheet.type === "viewsheet"` / `"worksheet"` / `"wiz"` appear ~20+ times), so the added guard is consistent with the existing type-narrowing idiom rather than introducing a new pattern. Every call site of `openSheet`/`openWorksheet`/`openViewsheet` passes a `type` that matches the sheet being created, so the extra predicate term can only narrow (never break) existing correct matches — I didn't find a call site where a legitimate match relies on `type` being absent from the predicate.
- **Regression risk**: Low. The change is a pure narrowing of two lookup predicates to also require the discriminator field that's already set correctly on every sheet object; it doesn't touch the widening/creation paths, so it cannot itself cause a "duplicate tab" regression, only prevent a cross-type mismatch.
- **Test coverage**: The new collision-case test is well-targeted — the PR description states it was verified to fail against the pre-fix predicate and pass with the guard. Good regression coverage for a previously completely-untested code path (`editAsset.subscribe`'s `command.viewsheet===true` branch).
- **Scope**: Minimal — two files, one small production diff (guard clause), matches the "Minimal scope" guideline. No unrelated refactoring.
- No security, error-handling, or backward-compatibility concerns identified; this is a client-only defensive-programming fix with no API/contract change.

**No blocking issues found in the code itself.**

### CI status — blocking issue
The `build` check is currently **FAILING** (independent of `claude-review`): https://github.com/inetsoft-technology/stylebi/actions/runs/34841600805/job/103967629843

```
[ERROR] Failures:
[ERROR]   AdminPropertiesControllerTest.distinguishesARealUnsetPropertyFromAnInventedOneOnlyByGuidance:312 expected: <unknown> but was: <confirmed>
[ERROR]   AdminPropertiesControllerTest.reportsExistenceAsUnknownWhenUncataloguedAndUnset:277 expected: <false> but was: <true>
```

This failure is in `inetsoft-core`'s `AdminPropertiesControllerTest`, entirely unrelated to this PR's diff (which only touches `composer-main.component.ts` and a new Angular `.tl.spec.ts` file under `web/projects/portal/...`). I confirmed it is **not a regression introduced by this PR** — the identical two assertions fail, byte-for-byte, on other unrelated open PRs against the same `stylebi-wiz-main-rebase` base branch (e.g. #5223, #5225, #5226, #5221), so it's a pre-existing, base-branch-wide broken/flaky test.

Worth noting separately: because the Maven reactor fails fast on `inetsoft-core`, the build never reaches the `inetsoft-web` module, so CI has **not actually executed** the new `.tl.spec.ts` regression test added by this PR — the passing results in the PR description's test plan are from local runs only. That's not this PR's fault, just a gap to be aware of until the base-branch test issue is fixed.

### Verdict
Per the review policy, a failing check outside of `claude-review` requires **Changes Requested** regardless of root-cause attribution, so this review requests changes — **not because of any problem with this PR's code**, but because the `build` check is red. Recommend: get the pre-existing `AdminPropertiesControllerTest` failure on `stylebi-wiz-main-rebase` fixed (separately from this PR), then re-run/merge. No changes are requested to this PR's own diff.

lizhepu:
Closing without merging — the reported root cause (bug 76636-A) is already fixed and merged: PR #5219 (commit 7f45a3ade, merged 2026-09-14) fixed both halves of the assetId/runtimeId handoff bug independently confirmed against the current source in this branch:

- `SheetOpenService.createViewsheet` now sends the new viewsheet's own asset entry id (`newVsEntry.toIdentifier()`), not the base worksheet's colliding id.
- `composer-main.component.ts`'s `editAsset` handler now threads `command.runtimeId` through `openViewsheet(...)`.

This PR's remaining content (a `sheet.type` defensive guard added in review round 1, plus its regression test) closes a *latent* gap — the client handler still has no independent defense against a colliding assetId, should one ever be sent again by a future regression — rather than anything required to resolve the reported symptom. After discussion, we're treating that as optional hardening rather than part of this bug fix, and choosing not to carry it forward here to keep this fix scoped to what #5219 already covers.

No code from this PR is being merged. Sub-bug 76636-A requires no further action.

PR #5226, opened 2026-09-14
fix(VCA-005): reconnect the null-sentinel mapping value to the existing delete path

fredaan:
## Summary
- A categorical-color pin made through `shareColors:true` could not be removed through any documented path — every write (rebind, `reset_visual_frame`, unbind+rebind) merges the viewsheet's existing pins forward and can only add/overwrite, never shrink. That merge-forward design (`ChartAestheticMutator.carryShareColorState`/`mergePins`) is deliberate and untouched by this fix.
- The backend already implements a genuine single-key delete: `Tool.getColorFromHexString` special-cases the literal string `CoreTool.NULL` (`"null"`) and returns Java `null`, and `Viewsheet.setDimensionColors` already filters `null`-valued entries when rebuilding a column's pins. The only thing blocking it was `VisualFrameAliases.normalizeColor` treating `"null"` as invalid hex and throwing before the sentinel could reach `getColorFromHexString`.
- `VisualFrameAliases.categoricalColor`'s mapping loop now passes the `CoreTool.NULL` sentinel through unchanged; every other value is still hex-validated exactly as before.

## Test plan
- [x] `VisualFrameAliasesTest` — new coverage: a `"null"`-string mapping value passes through as the sentinel (unshared and shared), and an actual malformed value (`"notacolor"`) still throws.
- [x] `ViewsheetTest` — new coverage: `Viewsheet.setDimensionColors` genuinely drops a `null`-sentinel-mapped key while a sibling key in the same write survives.
- [x] `ChartAestheticMutatorTest` — full suite re-run, unaffected.
- [ ] Live verification against a StyleBI instance (tester phase, not part of this PR).

Companion plugin-side PR (deliberate sentinel forwarding + documentation): inetsoft-technology/stylebi-wiz, branch `fix/76647-vca005-color-pin`.


fredaan:
## Review round 1 — VCA-005 Java fix (PR #5226)

**Head commit reviewed:** aa4343da5

**What I checked:** Read 01-diagnosis.md and 02-refute.md in full (both rounds), confirmed the
worktree's copies match `origin/main` byte-for-byte (md5 identical after line-ending
normalization — no staleness). Read the full diff. Ran the targeted tests myself:
`./mvnw.cmd -pl core -am -Dtest=VisualFrameAliasesTest,ViewsheetTest -Dsurefire.failIfNoSpecifiedTests=false test`
→ 75/75 pass (4 ViewsheetTest, 71 VisualFrameAliasesTest, including all 4 new regression tests).

**CI `build` failure investigated, not this PR's fault:** the two failing tests
(`AdminPropertiesControllerTest.distinguishesARealUnsetPropertyFromAnInventedOneOnlyByGuidance`,
`...reportsExistenceAsUnknownWhenUncataloguedAndUnset`) are in an unrelated module (admin
properties, not chart aesthetics) and fail with the identical assertion messages on sibling PR
#5224 (a different bug in this same batch, against the same `stylebi-wiz-main-rebase` base,
touching unrelated files) — a pre-existing, base-branch-wide failure, not something this diff
caused.

**Invariant claimed:** the `CoreTool.NULL` bypass in `categoricalColor()` is scoped to exactly
the literal string `"null"`; every other mapping value is still hex-validated exactly as before.
Confirmed directly in the diff — the ternary is `CoreTool.NULL.equals(value) ? CoreTool.NULL :
normalizeColor(value)`, an exact-match guard, not a broadened regex or a loosened check. The new
`anInvalidNonHexMappingValueStillThrows` test confirms a real malformed value is still refused.

**Scenarios walked beyond the reported repro:**
- Case-sensitivity (`"NULL"`, `"Null"`) — falls through to `normalizeColor`, still throws. Correct
  (the sentinel is one exact literal, not a general escape hatch).
- A JSON-literal `null` from `entry.getValue()` (not the string `"null"`) — `String.valueOf(null)`
  also produces `"null"`, so it's treated identically. Consistent, not a gap.
- Unshared/local (`shareColors:false`) delete via the same sentinel — not in the bug's original
  repro (which required `shareColors:true`), but reachable through the same `categoricalColor()`
  code path since the bypass sits before the `useGlobal` branch. I traced this by hand since it's
  a real "sibling of the branch the repro took": `assignMappedColors` →
  `CategoricalColorFrame.setColor(option, null)` (`CategoricalColorFrame.java:262-276`) removes the
  key from `cmap` rather than storing a null Color, and the read-back path
  (`CategoricalColorModel(wrapper)` constructor, `CategoricalColorModel.java:68`) rebuilds
  `colorMaps[]` from `wrapper.getStaticValues()` (i.e. `cmap.keySet()`), which no longer has the
  deleted key — so the local/unshared case also genuinely deletes, with no null-Color ever reaching
  rendering or XML serialization (`CategoricalColorFrameWrapper.writeXML` also null-checks and reads
  from `frame.getStaticValues()`, not the stray `wrapper.dimensionColors` field that
  `assignMappedColors` incidentally sets with a null value). Verified safe — not a regression — but
  flagging that this is now a real, working, additional capability the tests never actually
  exercise (see finding below).
- Double-delete (calling the sentinel twice, or deleting the last remaining pin on a column) —
  traced through `mergePins`/`Viewsheet.setDimensionColors`: idempotent, no crash, correctly leaves
  the column with zero pins if it was the last one.

**Finding (important) — the actual merge-forward layer has zero coverage of the null-sentinel.**
The two new test files each cover one end of the pipe in isolation:
`VisualFrameAliasesTest` calls `VisualFrameAliases.create()` directly (bypasses
`ChartAestheticMutator` entirely), and `ViewsheetTest` calls `Viewsheet.setDimensionColors`
directly (bypasses `create()`/`mergePins` entirely). Neither exercises
`ChartAestheticMutator.carryShareColorState`/`mergePins` — the layer that is actually the heart of
the bug (claim 2 in the diagnosis) and the layer whose overwrite-then-filter interaction is what
makes the *single-call* delete-while-other-pins-survive behavior work at all. This is exactly the
scenario the original bug report cared about (delete one pinned value, in one call, while a
sibling column's other pins survive) and it is untested end-to-end.

`ChartAestheticMutatorTest.java` already has the ideal template two tests away —
`aSharedMappingOverwritesAPinItNames` (line 1341) and `aSharedMappingAddsToTheColumnsPinsInsteadOfReplacingThem`
(line 1314) set up an existing `globalColorMaps` pin via `setGlobalColorMaps`, then call
`ChartAestheticMutator.setFrame(..., spec("type","categorical","shareColors",true,"mapping",...))`
and assert on `frame.getGlobalColorMaps()`. Suggest one more test in the same style: seed two
existing pins (e.g. East/West), call `setFrame` with `mapping: Map.of("East", CoreTool.NULL)`,
and assert `frame.getGlobalColorMaps()` has exactly one entry left (West), proving the delete
survives the merge with a coexisting pin — the actual, single-call fix for the bug as reported.

I traced this by hand through `carryShareColorState`/`mergePins`/`applyColorsToViewsheet` and
believe it is correct, but a fix for a bug whose root cause is explicitly the merge mechanism
should have a regression test that runs the merge, not just its two neighbors.

No other findings. Verdict: **changes requested** (one important finding).

fredaan:
Addressed — pushed 436cd29f2 adding `ChartAestheticMutatorTest.aSharedMappingsNullSentinelSurvivesTheMergeAlongsideACoexistingPin`.

One correction worth flagging: the suggested assertion (that `frame.getGlobalColorMaps()` already reflects the delete right after `ChartAestheticMutator.setFrame()`) doesn't hold — `ChartAestheticMutatorTest` operates purely on the in-memory `ChartBindingModel`/`CategoricalColorModel` and never touches a real `Viewsheet`. `VSChartBindingFactory.applyColorsToViewsheet` (where `Tool.getColorFromHexString` turns the sentinel into Java `null`) and `Viewsheet.setDimensionColors` (where the null-filter actually drops the key) are a separate class this test class never reaches. Ran it as originally suggested first and got `expected: <1> but was: <2>` — confirmed that's a layer-boundary issue, not a defect in the fix.

Corrected the assertion to match what this layer is actually responsible for: the merge carries the sentinel-valued pin through *unchanged* for the named key (`mergePins` has no null-awareness — keys only on `getOption()`) while leaving the coexisting pin untouched. The real delete stays covered end-to-end by `ViewsheetTest.setDimensionColorsGenuinelyDropsANullSentinelEntry`, which already exercises the actual `Viewsheet`. Re-ran `ChartAestheticMutatorTest`/`VisualFrameAliasesTest`/`ViewsheetTest` together: 177/177 green.

Detail recorded in `docs/teams/2026-09-14-bugs-76647-chart-aesthetic/bug-vca-005/03-fix.md` (stylebi-wiz PR #2509).

fredaan:
## Review round 2 — VCA-005 Java fix (PR #5226)

**SHA range reviewed:** aa4343da5..436cd29f2 (1 commit: the merge-layer regression test)

**What was checked:** the round-1 finding's fix. Read the diff of
`ChartAestheticMutatorTest.aSharedMappingsNullSentinelSurvivesTheMergeAlongsideACoexistingPin` in
full, and re-ran all three targeted test classes myself (not just trusted the reported number):
`./mvnw.cmd -pl core -am -Dtest=VisualFrameAliasesTest,ViewsheetTest,ChartAestheticMutatorTest -Dsurefire.failIfNoSpecifiedTests=false test`
-> **177/177 pass** (102 ChartAestheticMutatorTest, 71 VisualFrameAliasesTest, 4 ViewsheetTest).

**On the pushback about my suggested assertion:** agreed, and the correction is right. This test
class (`ChartAestheticMutatorTest`) operates purely on the in-memory `ChartBindingModel` — it
never touches a real `Viewsheet`, so `VSChartBindingFactory.applyColorsToViewsheet` and
`Viewsheet.setDimensionColors` (where the actual null-filter/delete happens, already proven by
`ViewsheetTest.setDimensionColorsGenuinelyDropsANullSentinelEntry`) are simply unreachable from
this layer. My original suggestion (asserting the merge result is already down to 1 entry) was
asking this class to prove something outside its boundary. The corrected assertion — the merge
carries the sentinel through unchanged for the named key (`East: CoreTool.NULL`) while leaving the
coexisting pin (`West: "#F28E2C"`) untouched — is exactly what this layer is actually responsible
for, and it's a real, non-trivial assertion: it would fail if `mergePins` filtered/dropped
null-colored pins itself (it doesn't — confirmed in round 1, `mergePins` never inspects
`getColor()`), and it would fail if the sentinel got corrupted into something else during the
merge. Combined with `ViewsheetTest`'s proof that the storage layer genuinely deletes a
null-valued entry, the two tests together now cover the full pipe end to end, matching what I
asked for.

No new findings. Verdict: **approved**.

Squash-merging #5226, then #2509 (its documentation/schema counterpart, already approved in round
1) once #5226 is confirmed merged.

PR #5229, opened 2026-09-14
fix(76647-vca002): degenerate LinearSizeFrame domain collapses to a fixed neutral size

fredaan:
## Summary

Redmine #76647, bug VCA-002 (P1): a chart's `size` aesthetic bound to a genuine numeric measure whose value is identical across every row (a degenerate `linear`-frame domain, min == max) was reported to render marks at visibly different sizes rather than a uniform size.

**This PR fixes one confirmed, independent defect found during diagnosis -- it does NOT confirm or fully resolve the originally reported symptom (per-mark divergent sizes).** Please read the "What this does and does not fix" section below before treating VCA-002 as closed.

## What this does and does not fix

**Confirmed and fixed**: `LinearScale.init()` unconditionally zero-collapses and then axis-widens ("nice tick") a degenerate (min == max) domain, with no carve-out for a size frame's own value-scale -- unlike the existing treemap-only `Scale.ZERO` carve-out in `VSFrameVisitor`. This makes `LinearSizeFrame.getSize()` compute a wrong-but-*uniform* ratio against the widened range (e.g. `35/40` of the size span) instead of a fixed neutral size, whenever every bound value is the same. This is real, independently demonstrated, and fixed here.

**Not confirmed, not fixed here**: the originally reported symptom is *divergent* per-mark sizes (different marks rendering at different sizes), not a uniform-but-wrong size. The uniform-wrong-ratio defect above is mathematically incapable of producing per-mark divergence on its own (every mark reads the same value through the same shared scale object). Diagnosis + adversarial refutation (two full rounds) traced several candidate mechanisms for the divergence -- a stale `CategoricalSizeFrame` surviving a rebind, a stale cached `Scale`, `PointElement.sortData()`'s by-size sort/`SortedDataSet` row translation, and a `VSFrameVisitor.getVisualDataSet` `SubDataSet`/`UnionDataSet` dataset-view mismatch -- and source-refuted or substantially narrowed all of them without finding a replacement mechanism. The open question remains **unresolved**; it needs live JDWP-attach or instrumented-logging verification (or a `/deep-debug-workflow` escalation) to settle, which was out of scope for this pass. Full writeup: `docs/teams/2026-09-14-bugs-76647-chart-aesthetic/bug-vca-002/` in the `stylebi-wiz` repo (`01-diagnosis.md`, `02-refute.md`, `03-fix.md`).

## Changes

- `LinearScale.java`: records whether the raw data domain was degenerate (`min == max`) before the existing zero-collapse/tick-widening mutates it, exposed via a new `isDegenerateDomain()` getter. Purely additive -- does not change `getMin()`/`getMax()`/`getTicks()`/`isValid()` or any other consumer's existing behavior.
- `LinearSizeFrame.java`: `getSize(Object val)` now returns a fixed midpoint size (`getSmallest()`/`getLargest()` average) when the scale's domain is degenerate, instead of computing the axis-widened ratio. Scoped to `LinearSizeFrame` only -- no other `LinearScale` consumer (axis, color/shape/texture frames) is affected.

## Test plan

- [x] New `LinearSizeFrameTest` (3 cases: non-zero degenerate domain, zero degenerate domain, non-degenerate domain unchanged) -- all pass with the fix.
- [x] Verified as a real regression test: reverted the fix, reran -- degenerate-domain cases fail with the pre-fix wrong values (`26.375` / `1.0` instead of the expected `15.5` midpoint); non-degenerate case still passes. Reapplied the fix, reran -- all green.
- [x] Full `inetsoft.graph.aesthetic.*`, `inetsoft.graph.scale.*`, `inetsoft.report.composition.graph.*` test packages (19 classes) pass with the fix, no regressions.


fredaan:
## Round 1 — CHANGES REQUESTED

(Posted as a comment, not a formal review — this PR was opened under the same account driving this review, and GitHub refuses self-review requests-changes.)

Head commit reviewed: `46fd152fa48db20b9cb74b7a796b3cea0134794d`.

### What was checked
- Full diagnosis/refute/fix record (`docs/teams/2026-09-14-bugs-76647-chart-aesthetic/bug-vca-002/` in `stylebi-wiz`), the PR diff, and the real source in the `community` worktree.
- **Independently re-verified the regression test is real, not vacuous**: reverted `LinearScale.java`/`LinearSizeFrame.java` to `46fd152fa^` (keeping the new test), ran it — the two degenerate-domain cases failed with the exact claimed values (`expected: <15.5> but was: <26.375>` / `<1.0>`). Restored the actual fix, reran — all 3 pass, byte-identical to the commit.
- Ran the claimed `inetsoft.graph.aesthetic.*` / `inetsoft.graph.scale.*` / `inetsoft.report.composition.graph.*` packages myself: 96 tests / 19 classes, all green.
- Verified Part 2's honesty-check source citations directly (`GraphGenerator.java:3304`'s `getRTFieldGroups().size() > 1` gate, `AbstractChartInfo.java:1474`'s `!isMultiStyles()` early return) — accurate. PR description correctly does not overclaim resolution of the reported per-mark divergent-size symptom.
- Confirmed `isDegenerateDomain()` is additive/inert for every other `LinearScale` consumer (grepped the whole `community` tree — only `LinearScale.java` and `LinearSizeFrame.java` reference it).

### Finding (important) — size legend swatches collapse to identical size for a degenerate domain

`LinearSizeFrame.getSize(Object val)`'s new branch (`core/src/main/java/inetsoft/graph/aesthetic/LinearSizeFrame.java:69-74`) ignores the `val` argument entirely and returns the fixed midpoint whenever `scale.isDegenerateDomain()` is true — correct for the per-mark case, but this same method is also the one `SizeLegendItem.paintSymbol` calls (`core/src/main/java/inetsoft/graph/guide/legend/SizeLegendItem.java:53`, `frame.getSize(getValue())`) once per legend tick item, where the tick values are synthetic values spanning the scale's still-axis-widened range (this PR didn't touch tick/value generation). `LinearSizeFrame.isMultiItem()` unconditionally forces the legend to show multiple items regardless of domain degeneracy.

I verified this empirically with a throwaway probe (not committed) against the actual fixed code: for a 3-row all-35.0 dataset, `getSize(0.0)`, `getSize(10.0)`, `getSize(40.0)` — three plausible legend tick values — all returned the identical `15.5`. So a size legend for exactly this bug's own chart shape (a degenerate linear size domain) will show several differently-labeled tick items (e.g. "0"/"10"/"20"/"30"/"40") whose swatches all render at the identical size, once this fix ships — a new, concrete, demonstrable defect in the confirmed fix's own implementation, not a restatement of the accepted "symptom not fully resolved" limitation.

Full writeup with repro steps: `docs/teams/2026-09-14-bugs-76647-chart-aesthetic/bug-vca-002/04-review-r1.md` (stylebi-wiz repo).

No other findings — the core degenerate-collapse mechanism, its per-mark test coverage, and the PR's honesty about scope are all independently verified correct.

fredaan:
Fixed in 2eb946277 -- good catch, thanks.

`LinearScale` now records the degenerate domain's raw single value (`getDegenerateValue()`, alongside the existing `isDegenerateDomain()`), and `LinearSizeFrame.getSize(Object)` only collapses to the neutral size when the value being sized actually equals it. A real per-mark read always matches (the data really is that one value); a synthetic legend tick (as in `SizeLegendItem.paintSymbol`, spanning the scale's still axis-widened range) normally doesn't, so the size legend keeps showing a meaningful range of swatch sizes, same as before this fix.

Added `degenerateDomainDoesNotCollapseSizeForSyntheticLegendTickValues` to `LinearSizeFrameTest`, which asserts synthetic tick values (0.0, 40.0) on a degenerate 35.0 domain still produce distinct, increasing sizes while the real bound value (35) still collapses to neutral. Verified it would have caught the round-1 finding: reverted to the unscoped branch, reran, and the new test failed with exactly the reported symptom (`expected: not equal but was: <15.5>`). Reapplied, all green. Full `aesthetic`/`scale`/`composition.graph` packages (19 classes, 97 tests) still pass with no regressions.

(Also cleaned up a stray compiled `LinearSizeFrameLegendProbeTest.class` left in `core/target/test-classes` from the throwaway probe -- its source was already gone, but the stale `.class` was still getting picked up by surefire's package-glob test runs and failing against the new, narrower behavior. Not part of the repo, just a local build artifact in the shared worktree.)

Ready for round 2.

fredaan:
## Round 2 — APPROVED

(Posted as a comment, not a formal review — self-review restriction, same as round 1.)

Head commit reviewed: `2eb946277` (range since round 1: `46fd152fa..2eb946277`).

### Fixer's change
`LinearScale` now also records the degenerate domain's raw single value (`getDegenerateValue()`), and `LinearSizeFrame.getSize(Object)` only collapses to the neutral midpoint when the value actually being sized equals it — not merely whenever the scale is flagged degenerate. Added `degenerateDomainDoesNotCollapseSizeForSyntheticLegendTickValues`.

### Independent verification
- Traced `Scale.map()` → `LinearScale.mapValue()` to confirm the compared value is a raw, untransformed numeric reading (not something distorted by the axis-widened range) — a real per-mark value always matches `getDegenerateValue()` by construction; a synthetic legend tick normally won't.
- Grepped the whole `community` tree: `getDegenerateValue()`/`degenerateValue` referenced only in `LinearScale.java` and `LinearSizeFrame.java` — still additive/inert for every other consumer.
- **Independently re-verified the fixer's "would have caught it" claim**: swapped the round-1 (unscoped) source back in while keeping the round-2 test — the new test failed with `expected: not equal but was: <15.5>`, reproducing my round-1 finding exactly. Restored the actual commit (byte-identical), reran — all 4 pass.
- Reran `inetsoft.graph.aesthetic.*`/`inetsoft.graph.scale.*`/`inetsoft.report.composition.graph.*`: 97 tests / 19 classes, all green.
- Walked the other `SizeFrame`/`getSize` call sites again for the new equality condition (`StackedMeasuresSizeFrame`, `LinearSizeFrameWrapper`'s brush override, `TreemapElement`, `ParaboxPointGeometry`) — all inherit the corrected behavior automatically, no new gap.
- **CI**: `build` failed, but independently confirmed via the job log that the only 2 failures in the full 9937-test run are `inetsoft.web.admin.ai.AdminPropertiesControllerTest.distinguishesARealUnsetPropertyFromAnInventedOneOnlyByGuidance` / `...reportsExistenceAsUnknownWhenUncataloguedAndUnset` — nothing under `inetsoft.graph.*`, and this PR touches only `LinearScale.java`/`LinearSizeFrame.java`/`LinearSizeFrameTest.java`. Confirmed as the same pre-existing baseline flake shared across sibling PRs #5223/#5224/#5225/#5226/#5227/#5228 in this batch. Not a blocker.

No new findings. Full writeup: `docs/teams/2026-09-14-bugs-76647-chart-aesthetic/bug-vca-002/04-review-r1.md` (stylebi-wiz repo, "Round 2" section).

**Verdict: approved. Merging.**

PR #5231, opened 2026-09-14
fix(wiz #76637): push a binding-tree refresh from attach_base_worksheet

fredaan:
## Summary

- `attach_base_worksheet` (`ViewsheetAssemblyAgentController.attachBaseWorksheet`) already calls `broadcast.broadcastRefresh(...)`, but that call only repaints visible top-level `VSAssembly` canvases (`SheetAgentBroadcastService.broadcastViewsheetRefresh`'s own doc comment says so). It never touches the Data panel/asset tree, so the reporter's newly attached worksheet only showed up after reopening the viewsheet or hitting Composer's own refresh action.
- Added `SheetAgentBroadcastService.broadcastBindingTreeRefresh(RuntimeSheet, String runtimeId, Principal)`, which builds the tree the same way `VSBindingTreeControllerService.getBinding` does (no assembly selected — the same no-assembly-context `ComposerBindingTree` itself uses) and pushes it as a `RefreshBindingTreeCommand` via the existing unsolicited, runtime-id-keyed `sendToBrowser`/`sendCommand` addressing `broadcastRefresh` already uses.
- `attachBaseWorksheet` now calls this alongside its existing `broadcastRefresh` call.
- On the client this reaches `VSPane.processRefreshBindingTreeCommand` (`viewsheet-pane.component.ts`), which writes into the same `VSBindingTreeService` singleton the persistent Toolbox tree (`ComposerBindingTree`/`BindingTreeComponent`) reads from — already wired, no Angular change needed.

## Scope limitation (accepted, not a defect)

`processRefreshBindingTreeCommand` only applies the refresh when `this.vs?.isFocused` — i.e. only for the currently-focused tab. An agent-paired viewsheet tab that the human has switched away from will still need a manual refresh on refocus, same as before this fix. This is a known, documented constraint of the fix's coverage, not something this PR also attempts to fix.

## Test plan

- [x] Added `SheetAgentBroadcastServiceTest`: `broadcastBindingTreeRefreshSendsTheTreeVsBindingTreeServiceBuilds`, `broadcastBindingTreeRefreshSendsNothingWhenNoTreeCanBeBuilt`, `broadcastBindingTreeRefreshSkippedWithoutASocketSession`.
- [x] Added `ViewsheetAssemblyAgentControllerTest.attachBaseWorksheetAlsoBroadcastsABindingTreeRefresh`, verifying `attachBaseWorksheet` calls the new broadcast.
- [x] `./mvnw test -pl core -Dtest=SheetAgentBroadcastServiceTest,ViewsheetAssemblyAgentControllerTest,SheetOpenServiceTest` — all pass (147 tests, 0 failures).
- [x] Ran the full `inetsoft.web.wiz.**` package suite for collateral-regression check — the three touched classes are still 100% green inside that run; the suite's pre-existing, unrelated `NoClassDefFound` flakiness in `WorksheetEditServiceMutatorsTest`/`WorksheetReadServiceTest`/etc. (worksheet-domain classes untouched by this change) is not new.


fredaan:
## Reviewer pass (bugs-76637)

Verified the diff against the confirmed root cause/mechanism from diagnosis+refutation (`docs/teams/2026-09-14-bugs-76637/bug-76637/01-diagnosis.md`, `02-refute.md`):

- `SheetAgentBroadcastService.broadcastBindingTreeRefresh` builds the tree via `vsBindingTreeService.getBinding(runtimeId, null, false, principal)` -- the exact call shape the diagnosis specified (mirroring `VSBindingTreeControllerService.getBinding`) -- and sends it through the class's existing shared `sendCommand` helper, so it inherits the same `RUNTIME_ID_ATTR`/no-`inetsoftClientId` addressing every other broadcast in this class already uses. Not a narrower, symptom-specific shortcut.
- `attachBaseWorksheet` calls it immediately after the pre-existing `broadcastRefresh(...)` call, reusing the controller's existing `broadcast` field -- no new dependency added to the controller itself.
- The stated deviation (new method added to `SheetAgentBroadcastService` rather than injecting `VSBindingTreeService` directly into the controller, to keep the constructor-parameter blast radius to this service's own 10 test call sites instead of the controller's ~13) is sound and doesn't add a bad layering dependency -- this service already owns "build + push a browser notification," including building richer domain objects (`VSObjectModel`) for a similar purpose.
- Correctly scoped: does not touch `CapturingCommandDispatcher`/`ModifyCalculateFieldService` (the wider, separate architectural gap the refutation round 1 found for `sessions.mutate()`-wrapped writes) -- confirmed via the full changed-file list, only `SheetAgentBroadcastService.java`, `ViewsheetAssemblyAgentController.java`, and their 3 test files are touched.
- The `isFocused` scope limitation (`VSPane.processRefreshBindingTreeCommand` only applies while the tab is focused) is correctly left alone -- no Angular files touched -- and is documented in the new method's own javadoc rather than glossed over.
- New/updated tests assert real content, not just "was called": `broadcastBindingTreeRefreshSendsTheTreeVsBindingTreeServiceBuilds` captures the actual `RefreshBindingTreeCommand` payload and asserts `assertSame(tree, command.getTreeModel())` plus the real `convertAndSendToUser` destination/topic/user.
- Failure handling matches the "stale tree must not fail a mutation that already succeeded" claim: the new method's own try/catch logs+swallows build/send failures, and it runs after the pre-existing `broadcastRefresh` call, so it can't roll back or fail an already-successful mutation.

No blocking issues found.

**CI note:** `build` (Validation Build) failed on `AdminPropertiesControllerTest.distinguishesARealUnsetPropertyFromAnInventedOneOnlyByGuidance` / `reportsExistenceAsUnknownWhenUncataloguedAndUnset` -- confirmed this is a pre-existing, repo-wide failure unrelated to this diff (the identical two assertions fail identically on other unrelated open PRs, e.g. `fix/76636-viewsheet-handoff`'s latest run). This PR never touches admin/properties code. `claude-review` also failed, but only due to a bot infra error ("Claude encountered an error after 0s"), not an actual review finding. Proceeding to merge on that basis.

Self-review note: this PR was opened under the same account running this review, so a formal `--approve` review is blocked by GitHub; posting findings as a plain comment instead, per this workflow's standing merge authorization.

PR #5233, opened 2026-09-14
fix(tabular): don't explode a single JSON object's own scalar-array fields (WBT-001)

fredaan:
## Summary
- Bug WBT-001 (Redmine #76648): `add_table`'s endpoint form against a single-object REST/JSON response (e.g. GitHub's `Repository` endpoint) returned N rows instead of 1 when the response has a top-level scalar-array field (e.g. `topics`). Every sibling field was duplicated across the N rows.
- Root cause: `ExpandedJsonTable.expandMap` (`connectors/inetsoft-tabular-util`) explodes any top-level list-valued field whenever the current recursion level is still under the query's `expandLevels` budget. For a bare JSON object at the document root (no enclosing list envelope), that budget is still fully available at level 0, so a scalar-array field belonging to the single record gets treated as a record collection and explodes. A paginated response's own per-record array fields already avoid this correctly, because the outer envelope list has already spent the expansion level by the time `expandMap` examines them.
- Fix: `loadStreamed` now flags a document root that is a single object (not a bare list) with a `DOCUMENT_ROOT_LEVEL` sentinel. `expandMap`'s list-expansion check requires the list to contain at least one record (object) to expand at that sentinel level; a list of scalars at the document root is instead joined/stringified into one value, exactly like a record's own scalar-array field already behaves once inside a paginated envelope. A root object's own list-of-records field (e.g. a `{"lastpage":true,"values":[...]}`-style wrapped pagination response) and the `LookupData` always-expand path are unaffected.

## Test plan
- [x] New regression test `testSingleObjectRootWithScalarArrayDoesNotExplode` in `ExpandedJsonTableTest` mirrors the exact GitHub `Repository`/`topics` shape from the bug report — asserts 1 row with `topics` stringified.
- [x] Existing `testExpandLevelsWithMap`, `lookupDataIsExpanded`, `testExpandLevelsWithArray`, `testExpandDeepArray`, and the rest of `ExpandedJsonTableTest` pass unchanged (9/9), proving the paginated/list-of-records and lookup-expansion paths are not regressed.
- [x] `inetsoft-rest`'s `DocumentParserTest`/`NoSchemaDocumentParserTest` (which also exercise `ExpandedJsonTable` for XML parsing) pass unchanged.
- [ ] Live re-verification via the plugin's own `add_table`/`preview_worksheet_data` MCP tools against GitHub `Repository` (owner `YingWise`, repo `StyleBI-GitHub-Test`) — blocked in this session by Chrome browser-automation connectivity; unit-level coverage above reproduces the exact reported shape and confirms the fix.

Redmine #76648 (WBT-001)


fredaan:
**Review round 1: CHANGES REQUESTED** (posted as a comment, not a formal review — GitHub blocks `gh pr review` on a PR authored by the same account)

# WBT-001 PR review - round 1

PR: https://github.com/inetsoft-technology/stylebi/pull/5233
Head commit reviewed: 2c1735d30de60fef1a732d1045ef6eec35874b59 (full diff since PR open - round 1)

## What was checked

- Read 01-diagnosis.md, 02-refute.md, 03-fix.md before the diff, per this repos review
  convention: checked the diff against the confirmed mechanism (expandLevels-budget consumption
  tied to whether the document root is itself a list envelope), not the reported symptom
  (topics specifically) or the originally-theorized object-vs-scalar branch.
- Confirmed the diff does not key on any repro literal (topics, GitHub, repo name) - the new
  DOCUMENT_ROOT_LEVEL sentinel and containsRecord() gate are shape-based, not field/connector-named.
- Independently re-ran ExpandedJsonTableTest at the PR head (mvnw.cmd -pl connectors/inetsoft-tabular-util
  -o test -Dtest=ExpandedJsonTableTest): 9/9 green, confirming the fixs own claim.
- Read expandMap/processJson/expandList in full (not just the diff hunks) to trace how the new
  DOCUMENT_ROOT_LEVEL = -1 sentinel propagates through recursion, per this repos "walk the invariant,
  not just the diff" convention.
- Verified the LookupData claim directly against the diff: atDocumentRoot is only ever set to true
  when lvl == DOCUMENT_ROOT_LEVEL, and the list instanceof LookupData branch (line ~312-318)
  unconditionally clears it before the eligibility check, and separately resets lvl = 0 itself - so a
  LookupData list never sees the document-root gate regardless of where it appears. This claim holds.
- Verified testExpandLevelsWithMap (root {"lastpage":..,"values":[...]}, values a record list
  directly on the root) and lookupDataIsExpanded (same shape, one LookupData element) genuinely
  exercise the containsRecord() gate correctly (not a workaround narrow to the fixture): both have their
  qualifying list field directly on the document root, which is exactly what the gate discriminates on.

## Alternate scenario walked (and the counter-example built)

Per review-workflows requirement to walk the sibling of the branch the repro took: the new
DOCUMENT_ROOT_LEVEL sentinel is only tested/guarded for a list field that sits directly on the
document root (topics, values). I walked the case one hop deeper: a document root object whose
non-list nested map field itself has a scalar-array field (e.g. GitHub-shaped
{"name":..., "owner": {"login":..., "aliases": [...]}} - a nested object one level under the root
carrying its own array).

Built and ran the counter-example (temporary JUnit test in ExpandedJsonTableTest, added, run, and
reverted - no artifact left in the repo; git status on E:\inetsoft\stylebi\community is clean):

    { "name": "Repo", "owner": { "login": "YingWise", "aliases": ["a", "b"] } }

at expandLevels = 1.

- At the PR head (2c1735d30): owner.aliases explodes into 2 rows (rowCount=3: header + 2
  data rows, one per alias) - the same defect class this PR exists to fix, reintroduced one level deeper.
- At the parent commit (1abda8a6e, pre-fix): owner.aliases correctly stays a single stringified
  value (rowCount=2: header + 1 data row, owner.aliases = ["a","b"]) - this is not a
  pre-existing, already-broken case; the PR change introduces it.

Root cause of the regression: expandMap nested-map recursion (ExpandedJsonTable.java:295,
expandMap(ridx, cnt, row, mapnames.get(i), map, expLvls, lvl + 1)) passes lvl + 1, where lvl is a
plain copy of the level parameter of the method - it is not adjusted for DOCUMENT_ROOT_LEVEL the way
the list-handling loop is (ExpandedJsonTable.java:306-310 explicitly maps DOCUMENT_ROOT_LEVEL to 0
before use). So when level == DOCUMENT_ROOT_LEVEL (-1), the nested map is recursed into with
level = 0 instead of the pre-fix value of 1. Every list field found one level inside that nested map
is then evaluated against expLvls at level = 0 instead of 1 - one full level cheaper than before
across the entire subtree - losing the atDocumentRoot/containsRecord gate entirely for anything past
the immediate root object, and, worse, letting a scalar array that is not on the document root explode
at a budget where it correctly would not have before this PR.

This is exactly the "sibling of the branch the repro took" case that the diagnosis and fix scope note
called out as needing evaluation (03-fix.md explicitly says other AbstractRestQuery connectors/shapes
need checking) but neither the fix nor its one regression test covers it. It is not hypothetical: any
REST/JSON connector shape with a nested (non-array) object field that itself carries an array - e.g. a
"data": { "tags": [...] } single-entity response - hits this at expandLevels = 1, the default for
every connector inheriting AbstractRestQuery expanded=true/expandTop=true (per the diagnosis own
"other call sites" section).

## Findings

- [blocking / important] connectors/inetsoft-tabular-util/src/main/java/inetsoft/uql/util/ExpandedJsonTable.java:295
  (nested-map recursion in expandMap): DOCUMENT_ROOT_LEVEL is not normalized to 0 before being used
  as the base for lvl + 1 here (unlike the list-handling loop at lines 306-310), so any list field
  reachable via one hop through a non-list nested map under the document root is now evaluated one level
  cheaper than before this PR, and can incorrectly explode where it previously (correctly) did not.
  Demonstrated with a reverted, throwaway JUnit test (see above) - not merely inferred. Suggested
  direction: normalize lvl the same way the list loop does (if(lvl == DOCUMENT_ROOT_LEVEL) { lvl = 0; })
  before computing lvl + 1 for the nested-map recursion, so the root object own fields keep the
  special gate while everything reached through it recurses at the same levels as before this PR.
  A regression test for this exact shape (root object -> nested non-list map -> nested scalar array)
  should be added alongside the fix.

No other blocking findings. The LookupData non-impact claim holds under direct inspection. The existing
testExpandLevelsWithMap/lookupDataIsExpanded tests are not narrow-workaround passes - they exercise the
actual discriminator (list field directly on the document root) the gate is built around.

## Non-blocking process notes (not gating this round verdict on their own)

- CI build check is FAILURE, but the failure (AdminPropertiesControllerTest, module inetsoft-core)
  is unrelated to this diff (diff only touches inetsoft-tabular-util) and is a pre-existing stale-fixture
  issue that has since been fixed on the base branch by PR #5243 (merged 2026-09-15, after this PR last
  CI run). This branch is behind base (mergeStateStatus: UNSTABLE, no conflicts) - it needs updating from
  stylebi-wiz-main-rebase to get a fresh, accurate build result once the code finding above is fixed.
- CI claude-review check is also FAILURE, but its own log shows an empty ANTHROPIC_API_KEY and an
  infra-level is_error:true before any review content was produced - not a real review finding, an
  action/infra problem out of scope for this PR.
- Diff does not touch wiz-services/src/agents/prompts/ or any prompt-affecting path - confirmed, this is
  pure Java in community/connectors/inetsoft-tabular-util.
- Minor style nit (non-blocking): the new test commented-out debug line
  (//inetsoft.report.internal.Util.printTable(actual);) matches the file own pre-existing convention
  (every other test in this file has the same commented print line), so not treated as a CLAUDE.md
  "unnecessary comment" violation.

## Verdict

Changes requested. One important-severity finding (regression at ExpandedJsonTable.java:295),
demonstrated with a reverted throwaway test, not a nitpick.

fredaan:
**Review round 2: APPROVED** (posted as a comment, not a formal GitHub review approval — GitHub blocks `gh pr review` on a PR authored by the same account as the reviewer)

# WBT-001 PR review - round 2

PR: https://github.com/inetsoft-technology/stylebi/pull/5233
Commit range reviewed: a13bb6d8f (round 1 head) .. 9a0a166bd (round 2 head, current PR head)

## Status of round-1 finding

**Fixed.** Round-1 finding was: connectors/inetsoft-tabular-util/src/main/java/inetsoft/uql/util/ExpandedJsonTable.java:295
(nested-map recursion in expandMap) did not normalize DOCUMENT_ROOT_LEVEL before computing lvl + 1,
so a scalar array nested one hop inside a non-list map field under the document root incorrectly
exploded.

Re-derived independently rather than trusting the commit message or 03-fix.md narrative:

- Read the actual round-2 diff (git compare a13bb6d8f...9a0a166bd): the only functional change is
  `int lvl = level;` -> `int lvl = level == DOCUMENT_ROOT_LEVEL ? 0 : level;` at
  ExpandedJsonTable.java:287 (the maps-loop, immediately before the maps.get(i) recursion at line 295),
  plus one new regression test (testSingleObjectRootWithNestedMapScalarArrayDoesNotExplode) mirroring
  my round-1 counter-example almost exactly (root -> owner map -> aliases scalar array).
- Confirmed this is the correct fix location and shape: it makes the nested-map recursion path use the
  same DOCUMENT_ROOT_LEVEL -> 0 normalization the list-handling loop already had (lines 306-310 in the
  round-1 diff), so recursing from the document root into a nested map now passes lvl + 1 = 1 again
  (matching pre-round-1 / production semantics), instead of the round-1 bug's -1 + 1 = 0.
- Re-ran the exact reverted throwaway test from round 1 (root.owner.aliases, expandLevels=1) against the
  current PR head (9a0a166bd), inserted and reverted the same way as round 1 (no artifact left; git status
  on E:\inetsoft\stylebi\community is clean after revert): now correctly gives 1 data row, with
  owner.aliases stringified as ["a","b"]. Confirmed via ExpandedJsonTableTest run: 11/11 tests pass
  (9 original + round 2's own new test + my reinserted scratch test).
- Went one hop further than the round-2 regression test itself covers, to check the fix generalizes rather
  than only patching the exact fixture reviewed in round 1: root -> owner (map) -> profile (map) ->
  aliases (scalar array), two hops deep, expandLevels = 1. Also reverted throwaway test. Result: 1 data
  row, owner.profile.aliases stringified correctly - not exploded. This confirms the normalization at the
  first hop from the document-root sentinel is sufficient for the whole recursive subtree (every level
  after the first hop already uses ordinary, non-sentinel level arithmetic identical to pre-round-1
  behavior), not merely a fix narrowed to the one scenario the round-2 test asserts.
- Re-confirmed the LookupData claim still holds after this change: the normalization only touches the
  maps-loop's lvl before recursing into a plain nested Map; the list-loop's independent atDocumentRoot
  handling and the LookupData-specific lvl = 0 resets (both map- and list-side) are untouched by this
  diff. lookupDataIsExpanded still passes.

No further scenarios found where the invariant (a scalar-only array anywhere in the document, regardless
of how many nested-map hops separate it from the root, should not spuriously multiply rows purely because
of where it sits relative to the document root) breaks. I tried to build a counter-example at two hops
deep and could not - it passed.

## CI

- build: now SUCCESS (was FAILURE in round 1). Confirmed this resolved cleanly rather than assumed: the
  branch head (9a0a166bd) descends from ac0f63bf6 / 9703bee1f, i.e. it now includes PR #5243's
  AdminPropertiesControllerTest stale-fixture fix that round 1 identified as the actual cause of the prior
  build failure. The new build run (2026-09-15T02:33-02:55) is a fresh, real green result, not a stale
  or skipped check.
- claude-review: still FAILURE, confirmed same failure mode as round 1 (not a new problem): the job log
  shows the identical empty ANTHROPIC_API_KEY and is_error:true before any review content is produced.
  Infra-only, out of scope for this PR, unchanged from round 1's assessment.
- mergeStateStatus is UNSTABLE (mergeable: MERGEABLE, no conflicts) - this reflects the still-failing
  claude-review check, not a merge conflict.
- Diff (both commits) still touches only
  connectors/inetsoft-tabular-util/src/main/java/inetsoft/uql/util/ExpandedJsonTable.java and its test
  file - confirmed no prompt-affecting path (wiz-services/src/agents/prompts/ or similar) is touched.

## New findings this round

None.

## Verdict

**Approved.** Round-1 finding is genuinely closed (re-derived and independently re-verified, not just
re-read), the fix generalizes beyond the one scenario its own regression test covers, and no new issues
were introduced. build is green, no merge conflicts. claude-review remains a known, unrelated infra
failure. Not merging - per instruction this PR is reserved for the user to merge personally.

Nelson-cao (approved):

PR #5234, opened 2026-09-14
Modern visualization: dark values for the selection and focus affordances

FigurePaper:
Two selection affordances drew the light brand orange `#E58A2A` on a dark card. Separate
defects in separate mechanisms; neither is a regression. Base `epic-74519`, community-only.

All ΔE figures below are measured in OKLab against the **re-tuned** `Modern Dark` head-8
(`#4FA5FF #FF8367 #49447D #2DEEC6 #AE41F5 #FFCB82 #FE3290 #9FEB28`), which this branch
sits on after being rebased past the palette re-tune (#5188). An earlier revision of this
description carried figures taken against the pre-re-tune palette; they have been
re-measured, and the notes below record which arguments the new numbers weaken. 0.106 is
the design set's separation line.

## 1. The active-border token had no dark value

`--inet-viz-active-border` was the one token whose dark-block entry re-pointed at the
*light* primary accent. It now resolves to `var(--inet-primary-color-dark)` — `#C96F12`,
the brand accent's existing deep step, not a new colour. `.bd-selected-assembly` moves
onto that token, which is the semantically correct one: it binds to `isFocused()`, and
`_viz-tokens.scss` defines `active` as exactly that, "outline, no fill".

Reachability was half the fix. The `.focus-assembly` overlay is a preceding sibling of
the wrapper carrying `viz-modern`/`viz-dark`, so no per-assembly dark token reached it.
The body-level `viz-shell-dark` would, but follows the org flag rather than the
assembly's mark — wrong palette for a `MODERN_DARK` assembly in a light org. The overlay
now binds both classes from the model, as its sibling already does.

Deeper rather than lighter is counterintuitive but measured: every `Modern Dark` member
is high-lightness by design, so lightness is the axis that separates.

| | contrast on `#252428` | ΔE to nearest `Modern Dark` |
|---|---|---|
| `#E58A2A` today | 5.88:1 | 0.079 — slot 2 `#FF8367` |
| `#C96F12` | 4.23:1 | 0.130 — slot 2 `#FF8367` |

The orange it replaces is **under** the 0.106 line; `#C96F12` clears it. Note the trade
the table makes explicit: contrast on the card drops 5.88:1 → 4.23:1, bought in exchange
for that separation, and still clear of the 3:1 non-text floor.

Two consumers change beyond the dark cards above, both deliberately. A **gate-off**
assembly in a **dark org** carries no viz class of its own, so it inherits the token from
the body-level `viz-shell-dark` scope and its ring changes too — an improvement on its
white card, 2.62:1 → 3.64:1. And `.vs-combo-box-trigger:focus`
(`_bootstrap-override.scss:1318`) is the token's other consumer and changes in dark:
same affordance, same collision.

## 2. Chart mark selection is canvas-painted and never saw the dark pass

`.viz-modern .chart-object-canvas` named `--inet-primary-color` directly with no dark
counterpart. Chart selections are drawn on a `<canvas>`: `drawRegions` reads that rule's
computed `color` as its fill and `border-color` as its stroke, so no custom property the
dark scope redefines was reaching it. Fix 1 could not have touched this.

Dark now strokes `--inet-viz-selected-text`, so a selected mark and a selected title
agree. The family's ink rather than its border, deliberately — this stroke lands on top
of the marks:

| dark stroke candidate | contrast on `#252428` | ΔE to nearest `Modern Dark` |
|---|---|---|
| `#E58A2A` today | 5.88:1 | 0.079 — slot 2 `#FF8367` |
| `--inet-viz-selected-border` `#2DD4BF` | 8.28:1 | 0.072 — slot 4 `#2DEEC6` |
| `--inet-viz-selected-text` `#A8EEF5` | 11.91:1 | 0.114 — slot 4 `#2DEEC6` |

The family's border is the palette's own teal to within 0.072 — under the line, so a
selection stroked with it would not read apart from the data underneath it. The ink is
the only candidate that clears the line, and it is the highest contrast of the three.
(Against the pre-re-tune palette the border measured 0.000, an exact match for a series
colour; the argument is the same one, less starkly.)

The fill is `--inet-viz-selected-fill-dark`, the same ink at the 28%
`--inet-focus-ring-color` uses in light. It is a literal `rgba()` rather than a
`color-mix()` of the ink because `ctx.fillStyle` rejects the `oklab()`/`color()` form
`getComputedStyle` returns for a mix, and drops it without raising.

Light is untouched: the selected family's light border is 1.43:1 on a white plot.

## Testing

CI `build` green on the rebased branch. Compiled-stylesheet assertions confirm the token
chain resolves, that the dark canvas rule still follows the light one on source order,
and that light mode is untouched — exactly two `--inet-viz-active-border:
var(--inet-primary-color)` declarations remain. Scoped suites green: the three
`vs-object-container` TL files (98 tests) and the `chart-tool` / `vs-data-tip` /
`vs-object-container` unit files (42 tests).

The manual dark matrix — the release gate, since nothing automated here looks at a
rendered pixel — has been run and passed on every row. One caveat recorded in the design:
it ran before the rebase, so the separation it confirmed by eye was against the
pre-re-tune palette. The re-measurement above is arithmetic, not a second look.


FigurePaper:
**Verdict: changes requested.** Posted as a comment rather than a formal review — GitHub rejects `REQUEST_CHANGES` on a self-authored PR ("Can not request changes on your own pull request").

---

# Code review — PR #5234

Two colour defects in two unrelated mechanisms, both real. I verified the substance independently (compiled `global.scss`, traced the `closest()` call sites, re-derived the contrast figures, ran the scoped suites) and **found no correctness or security problems**. The changes do what the description says they do.

Requesting changes for process reasons rather than code ones: two checks are red and the branch is behind its base. Details and the code-level items below.

## Verification I ran

| Check | Result |
|---|---|
| `sass` compile of `projects/portal/src/global.scss` | clean |
| Light-mode invariance | exactly 2 `--inet-viz-active-border: var(--inet-primary-color)` declarations remain (`:root`, `.viz-modern`) — the dark block is the only one re-pointed |
| `--inet-primary-color-dark` | emits `#C96F12` at `:root`, via `$primary-dark` → `$shell-primary-hover` |
| Rule order | `.viz-dark .chart-object-canvas` emits at line 13762, `.viz-modern .chart-object-canvas` at 13757 — equal specificity (0-2-0), dark wins on source order |
| `portal:test-tl` scoped to `vs-object-container.component.*.tl.spec.ts` | 3 files / 98 tests passed |
| `ng test portal` for `chart-tool*`, `vs-data-tip.directive`, `vs-object-container.component` | 5 files / 42 tests passed |
| Contrast figures in the description | `#C96F12` on `#252428` = 4.24:1, `#E58A2A` = 5.88:1, `#A8EEF5` = 11.91:1 — all three match |

Two regression risks I chased and cleared:

**The new `viz-modern`/`viz-dark` on `.focus-assembly` is genuinely inert.** Every compiled rule whose selector can match that element itself (`.viz-modern`, `.viz-shell`, `[data-viz-theme=v2]`, `.viz-dark`, `.viz-shell-dark`, the three `.viz-density-*` matrices) declares only custom properties — no paint. The div has no children, so the descendant rules cannot reach anything.

**The classes now make the overlay answer `closest(".viz-modern")`, and that turns out not to matter.** The overlay hosts both `VSDataTip` and `VSPopComponent`, and both directives call `GuiTool.isVizModernElement(this.elementRef.nativeElement)` — at `vs-data-tip.directive.ts:248` and `vs-pop-component.directive.ts:298`. Both calls sit inside `if(this.miniToolbar)`, and neither host on this div sets `miniToolbar`, so the changed return value is never read. Worth stating explicitly because it is the one non-obvious way a class binding on an empty div can change behaviour, and the plan's "the div has no children" argument does not cover it.

Also confirmed: `_viz-tokens.scss` and `_imports.scss` are both imported only from `portal/src/global.scss`, tokens first, so `.bd-selected-assembly` can never resolve `--inet-viz-active-border` in a bundle that lacks it. Had that not held, the `border` shorthand would have gone invalid-at-computed-value-time and the outline would have disappeared entirely rather than changing colour.

## Blocking: failing checks

Neither failure is caused by this PR, but both must be green before merge.

**`build` — 3 test failures in `inetsoft-core`, inherited from the base branch.**

```
VSChartBindingColorPalettesTest.anUnresolvableMarkStillReturnsEveryPalette:111  expected: <9> but was: <0>
VSChartBindingColorPalettesTest.nullVsIdFlagsTheNineHiddenNamesAndOmitsNothing:76  expected: <9> but was: <0>
VSChartPaletteDefaultsTest.cssHeadMatchesTheJavaFallback:295  NullPointerException: "css" is null
```

All three tests were introduced by `0761ce5596` ("Chart palettes: re-tune Modern and Modern Dark, add Contrast, scope the picker per chart", #5188), which landed on `epic-74519` after this branch was cut. None of them exists at this PR's head, and this PR changes no Java at all. The branch is 1 behind / 7 ahead. Please rebase onto current `epic-74519` and re-run — if they still fail there, that is a base-branch problem to fix separately, but it needs establishing before this merges rather than after.

**`claude-review` — infrastructure, not a finding.** The run errored after 422 ms on turn 1 with `is_error: true` and `ANTHROPIC_API_KEY:` empty in the job env. It produced no review. Re-run it.

`Semgrep Scan`, `Semgrep OSS` and `license/cla` pass. `mergeable` is `MERGEABLE` — there are no conflicts; `BLOCKED` is the failing checks plus `REVIEW_REQUIRED`.

## Code-level items

### 1. The dark fill hard-codes a value the line below it reads from a token

`_themeable.scss:1462-1463`

```scss
.viz-dark .chart-object-canvas {
  color: rgba(168, 238, 245, 0.28);
  border-color: var(--inet-viz-selected-text);
}
```

`rgba(168, 238, 245, …)` is `#A8EEF5`, which is exactly what `--inet-viz-selected-text` resolves to in this scope. So the fill and the stroke are the same colour by coincidence of transcription rather than by construction. Re-tune `--inet-viz-selected-text-dark`, or let a customer override it through `theme-variables.css` — the seam the `-dark` block in `_viz-tokens.scss` exists to provide — and the stroke follows while the fill silently does not. The light rule four lines up has no such split: both its fill and its stroke are tokens, and its fill token `--inet-focus-ring-color` is `rgba(229, 138, 42, 0.28)`, the same 28%.

Suggest declaring the paired literal next to its siblings in the customer-overridable dark block and consuming it with `var()`, which keeps the pair in one place and matches how light does it:

```scss
// _viz-tokens.scss, beside --inet-viz-selected-text-dark
--inet-viz-selected-fill-dark: rgba(168, 238, 245, 0.28);
```

### 2. The comment for that line gives a reason the adjacent line disproves

"The fill stays a literal wash because canvas silently drops a computed value it cannot parse."

As written this reads as ruling out `var()`, which cannot be right — `getComputedStyle().color` substitutes `var()` and returns a plain `rgb()`/`rgba()` string, which is exactly what `drawRegions` relies on for `border-color` on the very next line, and for `color: var(--inet-focus-ring-color)` in the light rule. If the real concern is `color-mix()` or relative colour syntax resolving to `oklab()` / `color(srgb …)`, which `ctx.fillStyle` can indeed reject silently, say that specifically. The current wording will send the next reader down the wrong path.

### 3. `.viz-dark .chart-object-canvas` depends on source order with nothing pinning it

It ties with `.viz-modern .chart-object-canvas` on specificity and wins only by appearing later. `_viz-tokens.scss:167-169` states that convention explicitly for its own dark block ("This block follows the `.viz-modern` blocks so its selector wins on source order"); the new rule in `_themeable.scss` does not. A future reorder, or hoisting the rule into a partial that is imported earlier, reverts dark charts to the light accent with no error anywhere. One clause in the comment that is already there would pin it.

### 4. The documentation covers only half of what ships, and one of its claims is now false

The design doc, the plan and the roadmap Done row all describe fix 1 only — `grep` for "canvas" in `2026-09-11-dark-active-border-design.md` returns nothing. `CLAUDE.md` makes `chart-card-roadmap.md` the authoritative entry point for this track, so shipping fix 2 unrecorded leaves the entry point understating what landed.

More than an omission, the same PR adds this to "Still undecided":

"2026-09-11: still open, and now with one fewer consumer to migrate. The dark active-border work deliberately did *not* point the assembly focus outline at this family…"

But `.viz-dark .chart-object-canvas` points a selection affordance straight at `--inet-viz-selected-text` and hard-codes a second copy of its value. Net, this PR **adds** consumers of the teal family rather than removing one. Whoever picks up the retirement question will be reading a note that is wrong about the state this PR left. Either record fix 2 in the Done row and correct that annotation, or split fix 2 into its own PR with its own design note.

### 5. The named release gate has not been run, and the matrix is missing the case the design flags as a regression

Credit where due: the design says so out loud rather than burying it — "**The manual dark matrix below has NOT been run.** It is the release gate… It is recorded as outstanding rather than quietly dropped." The plan says "**Do not mark the plan complete without it**", and all 33 of its checkboxes are committed unticked. Both changes are pure colour changes; no automated check in either fix looks at a rendered pixel, and the Testing section of the description lists only the two Vitest suites.

That is fine for a review but not for a merge. Either run it and record the outcome in the design's Verification section, as Task 4 Step 6 directs, or say in the PR that the gate is deliberately deferred and who owns it.

One gap in the matrix itself while it is still being edited: the design identifies "the one axis on which this change is a regression" — at nine or more series both palettes splice the legacy tail, whose slot 31 `#CC6600` sits ΔE **0.019** from `#C96F12`, against 0.052 for today's `#E58A2A`. Task 4 Step 2 enumerates assembly types but never fixes a series count, so the matrix as written cannot observe the one case the design predicts will look wrong. Add a ≥9-series chart row.

### 6. The description understates the blast radius

"Light is untouched" is accurate, and the light-mode verification backs it. But two things change beyond the dark cards the description discusses, both correctly documented in the plan and neither mentioned here:

- A **gate-off** (non-modern) assembly in a **dark org** carries no viz class of its own, so it inherits `--inet-viz-active-border` from the body-level `viz-shell-dark` scope and its focus ring changes to `#C96F12` too. The plan argues this is an improvement (2.62:1 → 3.64:1 on its white card) and I agree, but it is a visible change to legacy assemblies.
- `.vs-combo-box-trigger:focus` (`_bootstrap-override.scss:1318`) is the token's other consumer and changes in dark as well.

Also worth noting in the body, since it is the trade the whole change rests on: contrast on the dark card **drops** from 5.88:1 to 4.24:1. Still clear of the 3:1 non-text floor, and bought in exchange for palette separation — but a reader of the description alone would take "deeper rather than lighter" as a pure win.

## Not blocking

- The template binding ships with no regression guard, which the design records as a deliberate trade — all four `vs-object-container` spec files go through `makeComponent()` and none renders the template, so guarding a one-attribute change means standing up a render harness. I read those files and the claim is accurate. Reasonable call, though the project does have TL infrastructure (`render()` + MSW) that could carry it cheaply if this area keeps changing.
- No committed assertion guards either stylesheet change; the check scripts live in a scratch directory by design. There is no precedent in the repo for compiled-CSS assertions in a spec, so this is consistent rather than a lapse.
- `core/src/main/resources/inetsoft/web/resources/css/layout.css:26` declares `--inet-primary-color-dark: #D25F11`, a different value from the portal's `#C96F12`. Pre-existing and on the login/error path rather than this one, so out of scope here, but it means "the brand accent's existing deep step" is not a single value across the product.

## Summary

The engineering is sound: the diagnosis is right in both cases, the reachability half of fix 1 is a real bug that a token-only change would have left inert, the measurements hold up, light mode is provably untouched, and nothing I could find regresses. Items 1-3 are small and mechanical. Items 4-6 are about the record and the release gate, and item 5 is the one that should actually decide whether this merges — both changes are colour, and nothing automated has looked at a pixel yet.

Next steps: rebase onto `epic-74519` and get `build` green, re-run `claude-review`, address items 1-4, and either run the dark matrix or state the waiver.

FigurePaper:
Rebased onto `epic-74519` (clean — the base commit touched none of this branch's files) and addressed items 1-4 in two commits. **Both are local at the time of writing and not yet pushed, so the diff above is still the pre-rebase state**; this comment goes up first so the measurement correction below is on the record either way.

What the two commits do, briefly, since the diff does not show them yet:

- **1.** The canvas fill becomes `--inet-viz-selected-fill-dark`, declared beside its siblings in the customer-overridable dark block, so fill and stroke are both overridable and both in one place.
- **2.** The comment now names the real constraint — `ctx.fillStyle` rejects the `oklab()`/`color()` form `getComputedStyle` returns for a `color-mix()` — rather than implying `var()` was the problem.
- **3.** The rule's source-order dependency is stated in the comment.
- **4.** The canvas fix gets a "The second defect" section in the design, items 5 and 6 in "The change", its own row in the roadmap's Done table, and a note in the plan recording that it was found mid-execution rather than retrofitted as a task.

Then what the diff will not show even once it lands:

**The rebase invalidated every ΔE figure in these documents, including two the argument leans on.** The palette re-tune replaced all eight `Modern Dark` head colours, so the numbers in the PR body above were measured against a palette this branch no longer carries. Re-measured in OKLab against the re-tuned head-8:

| | quoted above | re-tuned |
|---|---|---|
| `#C96F12` (active-border-dark) | 0.155 | **0.130** |
| `#E58A2A` (what it replaces) | 0.119 | **0.079** — under the 0.106 line |
| `--inet-viz-selected-border` `#2DD4BF` | 0.000, "same colour as slot 7" | **0.072** — slot 4 `#2DEEC6` |
| `--inet-viz-selected-text` `#A8EEF5` | 0.126 | **0.114** |

Both conclusions hold and the contrast figures were all correct. But two arguments weaken, and are now recorded as weaker rather than left reading as settled: the selected border is no longer an *exact* palette match, only under the separation line; and the neutral `#94A3B8` goes from 0.060 to **0.121**, so it clears the line too and loses to `#C96F12` on margin rather than decisively. The legacy-tail collision at 0.019 is unchanged. The `_themeable.scss` comment carried the same stale "exact match to slot 7" claim and is fixed with it.

**Correcting what I said about the `build` failure.** I called it a base-branch problem to fix separately. Sharper: both classes pass *in isolation* on the rebased branch — `VSChartPaletteDefaultsTest` 25/0/0, `VSChartBindingColorPalettesTest` 4/0/0. They fail only inside the full 5741-test run, and the shape of it (`ColorPalettes.getPalette` returning null, hidden-name sets coming back empty) points at shared static state rather than at the palette code being wrong. So it is suite pollution on the base, and the rebase alone may not clear CI.

**Deliberately not resolved:**

- **The matrix pre-dates the rebase.** It has been run and passed on every row, and that is now recorded in the design's Verification section with the plan's checkboxes ticked. But it ran against the old palette, so what it confirmed by eye was separation against colours the branch no longer carries. The re-measurement above is arithmetic, not a second look at a rendered pixel. Re-running the chart rows is cheap and has not been done.
- **Item 6, the PR description.** Still says only "Light is untouched". The gate-off-in-a-dark-org change, the `.vs-combo-box-trigger:focus` change, and the 5.88:1 → 4.23:1 contrast trade are all documented in the design and the plan but not in the body above. Left for whoever edits the description.

One thing worth pulling out of item 4, because it binds whoever picks up the teal-family retirement: both documents claimed this work left that family with one fewer consumer to migrate. It leaves it with two more — the canvas fix strokes `--inet-viz-selected-text` and adds `--inet-viz-selected-fill-dark` beside it. The active/selected split the sentence was defending did hold, which is how the error survived.

FigurePaper:
**Verdict: changes requested — for one red check, not for the code.** Posted as a comment because GitHub rejects `REQUEST_CHANGES` on a self-authored PR.

To be plain about which of the three approve criteria fails, because only one does:

| Criterion | Status |
|---|---|
| No blocking issues in the code review | **met** — nothing blocking; four non-blocking notes below |
| No failing checks | **not met** — `claude-review` is red |
| No conflicts preventing merge | **met** — `MERGEABLE` |

`claude-review` failed in 37s with `ANTHROPIC_API_KEY:` empty in the job env and `is_error: true` after a single turn at 520ms. It produced no review at all. That is a repository secret problem, not something reachable from this branch — no commit here can turn it green. If the check is not going to be fixed, this needs an admin override or the check marked non-required rather than another push.

**`build` passed, in 22m.** That reverses what I said on the first pass, twice over, and the second correction matters more than the first. I had reported the three `VSChartPaletteDefaults` / `VSChartBindingColorPalettes` failures as inherited from base commit `0761ce5596`, then refined that to suite pollution on the base after both classes passed in isolation locally. Neither held: the full suite is green on the rebased branch. The honest read is that the earlier failure was transient or was fixed on the base in between, and my "the rebase alone may not clear CI" prediction was wrong. It cleared it.

---

Second pass, at `8f1b4f7ee2` — rebased onto current `epic-74519`, with the four items from the first review addressed.

## The four items

All four are resolved, and I re-verified each against the compiled output rather than the diff:

- **1.** The fill is `var(--inet-viz-selected-fill-dark)`, declared at `:root` (compiled line 383) beside its siblings. No literal left in the rule.
- **2.** The comment no longer implies `var()` is the obstacle; it names `color-mix()` resolution as the reason the alpha is pre-applied.
- **3.** The source-order dependency is stated in the rule's comment. Compiled order confirms it still holds: `.chart-object-canvas` 13753, `.viz-modern` 13758, `.viz-dark` 13763.
- **4.** The canvas fix now has a design section, two entries in "The change", a roadmap Done row, and a plan note. The "one fewer consumer to migrate" claim is corrected in both places.

Re-verified: `global.scss` compiles; exactly 2 light `--inet-viz-active-border` declarations remain; 98 TL tests and 42 unit tests green.

**The cross-cases are right**, which is the part of this scoping that usually breaks and which I checked properly this time rather than assuming:

- A `MODERN_DARK` assembly in a **light** org — wrapper carries `.viz-dark`, so `--inet-viz-selected-text` resolves through the assembly's own scope, and the fill token is `:root`-level and always resolvable. Neither depends on the body class.
- A `MODERN_LIGHT` assembly in a **dark** org — the wrapper has no `.viz-dark`, so the dark rule does not match and the light values apply, even though the body's `viz-shell-dark` would have supplied a dark `--inet-viz-selected-text`. It is never read, because the rule that would read it does not match.

That is the failure the `_directives.scss` tooltip comment describes, avoided here by construction.

## Findings

Nothing blocking. Four things worth recording, in descending order of how much they matter.

### 1. The fill and the ink can still diverge — the fix narrows this rather than closing it

I want to be precise, because my own first-round wording overstated what this achieves. `--inet-viz-selected-fill-dark` is an independent `:root` token. A customer theme that overrides `--inet-viz-selected-text-dark` alone moves the stroke and leaves the wash at the stock cyan.

What the change does buy is real: the two now sit two lines apart in the same block, so a **re-tune** touches both, and the value is discoverable and overridable at all rather than buried in a rule. It is also exactly the structure light already has — `--inet-primary-color` and `--inet-focus-ring-color` are equally independent — so this is parity, not a new seam.

The alternative that would close it is `color-mix()`, which is the thing the comment rules out. So this is the best available shape, not a compromise. But "keeps the pair in sync" would be the wrong thing to claim for it, and the docs should not grow that claim later.

### 2. The block's header comment is now inaccurate in two small ways

`_viz-tokens.scss:55-56` introduces the block as *"customer-overridable dark state colors, mirroring the -modern seam; consumed only inside `.viz-dark` below, so gate-off / light-modern are inert."* Both halves have drifted:

- **"consumed only inside `.viz-dark` below"** — `--inet-viz-selected-fill-dark` is consumed in `_themeable.scss`, not in the `.viz-dark` block below in this file. The *intent* survives (the consuming rule is `.viz-dark`-scoped, so gate-off and light-modern are still inert) but someone tracing consumers within the file will conclude the token is dead.
- **"mirroring the -modern seam"** — there is no `--inet-viz-selected-fill-modern`; light reaches for `--inet-focus-ring-color` instead. The block advertises a symmetry this token does not have.

One clause fixes both. Cheap, and the comment is load-bearing for anyone deciding where a new token goes.

### 3. The canvas-parsing claim is load-bearing and unverified in-repo

*"ctx.fillStyle rejects the oklab()/color() form getComputedStyle returns for a color-mix(), and drops it without error"* is the entire justification for the pre-multiplied literal. It is plausible and matches how `--inet-focus-ring-color` is authored, but nothing in the repo exercises it, and the TL suite runs in a DOM shim that would not reproduce canvas colour parsing either way. Browser support for CSS Color 4 in canvas has also moved. Not worth a test on its own; worth knowing it is an assertion rather than a measured result if anyone revisits the fill.

### 4. Naming: `-fill-dark` sits next to `-bg-dark`, and they are different things

`--inet-viz-selected-bg-dark` is the opaque selected-cell background `#0E3B44`; `--inet-viz-selected-fill-dark` is a 28% wash for canvas. Distinct roles, adjacent names, two lines apart. Not worth churning now, but a reader reaching for "the selected background in dark" has two plausible hits.

## Still open from the first review

**The PR description is now wrong, not merely incomplete.** The rebase re-based every ΔE figure, and the body still carries the pre-rebase numbers in two tables:

| in the description | correct against the rebased base |
|---|---|
| `#C96F12` — 0.155 | **0.130** |
| `#E58A2A` — 0.119 | **0.079**, which is *under* the 0.106 line |
| `#2DD4BF` — 0.000, "the same colour as slot 7" | **0.072**, slot 4 `#2DEEC6` — not an exact match |
| `#A8EEF5` — 0.126 | **0.114** |

The docs in the diff have all of this right and record which arguments weaken as a result. The description does not, and it is what a squash-merge turns into the commit message. Worth a few minutes before merge.

Also still unmentioned there, and correct in the design: the gate-off-assembly-in-a-dark-org change, the `.vs-combo-box-trigger:focus` change, and that contrast on the dark card **drops** 5.88:1 → 4.23:1 in exchange for palette separation.

**No committed regression guard** for either stylesheet change or the template binding. Unchanged, deliberate, and recorded in the design. Still reasonable.

## Security, compatibility, regressions

No security surface: CSS custom properties and two boolean class bindings from a server-supplied model, no user input, no injection path. No API or model changes, so nothing to break for callers. The behavioural changes are the two the design already names — gate-off assemblies in dark orgs and the combo-box focus border — both documented, both improvements on contrast. The `viz-modern`/`viz-dark` bindings on the focus overlay remain inert: every compiled rule that can match that element declares only custom properties, and both `GuiTool.isVizModernElement` call sites are behind `if(this.miniToolbar)`, which neither host on that div sets.

FigurePaper:
Two follow-ups to the review above.

**The description has been corrected.** All four stale ΔE figures now carry the values measured against the re-tuned `Modern Dark` head-8 this branch actually sits on — 0.079 / 0.130 for the active border, 0.072 / 0.114 for the stroke candidates — each with the slot it is nearest to, and the head-8 itself is quoted at the top so the numbers can be checked without digging. The `#2DD4BF` row no longer claims to be "the same colour as slot 7"; that was true of the old palette and is not true of this one, and the old figure is kept as a labelled aside rather than deleted, since the argument it supported is the same one.

While there I also filled the gaps the review flagged: the contrast trade (5.88:1 → 4.23:1, bought for separation, still clear of the 3:1 non-text floor), the gate-off-assembly-in-a-dark-org and `.vs-combo-box-trigger:focus` changes, the new `--inet-viz-selected-fill-dark` and why it is a literal rather than a `color-mix()`, and a Testing section that says what was actually verified rather than carrying test counts from before the rebase.

**`claude-review` being red can be disregarded.** It is not reachable from this branch:

- The job env shows `ANTHROPIC_API_KEY:` empty. The run errors on turn 1 at ~0.5s with `is_error: true` and produces no review at all — it never gets far enough to look at a diff.
- It failed identically on both heads, before and after the rebase (37s each, `520ms` and `422ms` to the error). Two unrelated trees, same failure, same duration profile.
- Nothing in this PR touches CI configuration, workflows, or secrets — the diff is two SCSS files, one template attribute pair, and documentation.

So it is a missing repository secret, not a finding. Restoring it, overriding the check, or de-requiring it are the options; another push will not change the outcome.

With that check set aside, the position is: `build` green, both Semgrep checks and CLA green, `MERGEABLE` with no conflicts, and nothing blocking in the code review — four non-blocking notes, of which the most substantive is that making the fill a token narrows rather than closes the fill/ink divergence under a partial theme override. The remaining gate is `REVIEW_REQUIRED`, which needs a human approval; GitHub will not take one from the author, and I cannot approve a PR I have been committing to.

jshobe-inetsoft (approved):

PR #5235, opened 2026-09-14
Bug #76576, don't dispose a runtime viewsheet mid-export

jkidacki-inetsoft:
## Summary

Follow-up to PR #5152. That PR fixed a null-`VSTableLens` NPE in the HTML
exporters and serialized the clone/swap in
`ExportControllerService.writeViewsheetExport()` with `synchronized(rvs)`,
but didn't stop a *different* thread from disposing the same runtime
viewsheet out from under an in-progress export.

## Root Cause

Two overlapping HTML export requests for the same viewsheet resolve to the
same `RuntimeViewsheet` instance. `ExportControllerService.exportViewsheet()`
never set the `__EXPORTING__` property that
`ViewsheetEngine.closeViewsheet()` checks before disposing a
`RuntimeViewsheet` (a guard `VSExportService` already implements correctly).
As a result, when one export's `finally` block unconditionally closed the
runtime viewsheet after finishing, a second export still inside the
`synchronized(rvs)` clone/export block in `writeViewsheetExport()` could
observe a null `Viewsheet` and throw:

```
java.lang.NullPointerException: Cannot invoke "inetsoft.uql.viewsheet.Viewsheet.clone()" because "originalViewsheet" is null
	at inetsoft.web.composer.vs.controller.ExportControllerService.writeViewsheetExport(ExportControllerService.java:222)
```

## Fix

- Set `rvs.setProperty("__EXPORTING__", "true")` before the real export
  work in `ExportControllerService.exportViewsheet()`, and clear it in a
  `finally` — matching the pattern already used by
  `VSExportService.exportViewsheet()`.
- Moved the actual export call, output stream close, and result
  construction into the `try` block so a thrown exception from the export
  itself is now caught and the cleanup `finally` (including the
  `__EXPORTING__` clear and the existing `_CLOSE_AFTER_EXPORT_` close
  check) actually runs. Previously that `try` only wrapped the
  non-throwing `return new ViewsheetExportResult(...)` call, so an
  exception from the export bypassed cleanup entirely.

## Test Plan

- `mvnw clean test-compile -pl core -am` — compiles cleanly.
- `mvnw test -pl core -am -Dtest=VSExportServiceTest,AbstractVSExporterLabelTest` — 43/43 pass.
- Manual repro from the Redmine issue (export HTML from Portal twice in a
  row before the first completes) no longer throws.

Fixes Redmine #76576.


jshobe-inetsoft (changes requested):
## Manual Review (replacing failed `Claude Code Review / claude-review` bot run)

The automated `claude-review` check errored out in <1s (known signature of the review bot hitting its usage quota — it never actually reviewed this PR). I performed a manual review of the diff, verified the fix logic against the existing, working reference implementation (`VSExportService`), compiled the branch, and ran the tests cited in the PR description.

### Summary of the change

`ExportControllerService.exportViewsheet()` now:
1. Sets `rvs.setProperty("__EXPORTING__", "true")` before doing the real export work, matching the guard `ViewsheetEngine.closeViewsheet()` already checks (and that `VSExportService.exportViewsheet()` already sets) before disposing a `RuntimeViewsheet`.
2. Moves the actual export call, output-stream close, and result construction inside the `try` block, so an exception thrown from the export path is now caught/logged and — critically — the `finally` block (which clears `__EXPORTING__` and performs the deferred `_CLOSE_AFTER_EXPORT_` close) now always runs. Previously the `try` only wrapped the non-throwing `return new ViewsheetExportResult(...)`, so a thrown exception bypassed cleanup entirely.
3. Drops a dead `rvs != null &&` check in the finally condition — `rvs` is dereferenced unconditionally earlier in the method (`rvs.getViewsheet()...`), so it can never be null at that point; this is a safe, no-op cleanup, not a behavior change.

### Correctness

- I traced the fix against `VSExportService.exportViewsheet()`/`doExportViewsheet()`, which the PR explicitly models this change on, and against the `__EXPORTING__`/`_CLOSE_AFTER_EXPORT_` handshake in `ViewsheetEngine.closeViewsheet()` (lines ~386-430). The new code reproduces that pattern faithfully: set the flag before the guarded work, clear it in `finally`, then check `_CLOSE_AFTER_EXPORT_` (deferred by a concurrent close request) to actually close afterward.
- Verified the specific reported repro path: an already-open viewsheet exported twice concurrently uses `matchesAssetIdFormat = false` (set in `ExportController.java`, since no fresh `AssetEntry`-based `openViewsheet()` call happens for an already-open runtime viewsheet), so the self-close branch in the `finally` doesn't fire for either overlapping export — only a genuine third-party close request (which now correctly gets deferred via `_CLOSE_AFTER_EXPORT_`) would trigger a close while an export is active. This matches the intended fix.
- Compiled the branch locally (`mvnw -pl core -am test-compile`) — clean, no errors.
- Ran the tests the PR cites: `VSExportServiceTest` (35 tests) and `AbstractVSExporterLabelTest` (8 tests) — all 43 pass locally, confirming no regression in the sibling export path this change was modeled on.

### Minor observations (non-blocking)

1. **Residual race with 3+ concurrent actors**: since `__EXPORTING__` is a single boolean property rather than a reference count, if two overlapping exports of the same `RuntimeViewsheet` overlap with the *first* export both finishing early and independently satisfying its own `matchesAssetIdFormat`/`_CLOSE_AFTER_EXPORT_` close condition while a second export's `synchronized(rvs)` clone/export block is still in flight, the flag could be cleared prematurely. This is a pre-existing limitation of the same pattern in `VSExportService`, not something introduced or worsened by this PR, and is out of scope for a minimal, targeted bug fix. Worth a follow-up ticket if it's ever observed in practice, but not blocking here.
2. **No new automated regression test** for the concurrency scenario itself (only the pre-existing `VSExportServiceTest`/`AbstractVSExporterLabelTest` suites were re-run, plus a manual repro per the PR description). Concurrency bugs like this are notoriously hard to assert deterministically in a unit test, so the omission is understandable, but a stress/integration test (e.g., two overlapping export calls against a shared mock `RuntimeViewsheet` asserting `closeViewsheet` is deferred) would harden this against future regressions.
3. Resource cleanup: if `writeViewsheetExport()` throws, the `DeferredFileOutputStream`/`BinaryTransfer` created earlier in the method are not explicitly closed/cleaned up in the catch/finally. This is pre-existing behavior (identical before this PR, since the same exception previously propagated before ever reaching the `try` block) and is not a regression — just flagging as a possible future cleanup.

None of the above are blocking; the change is a small, well-targeted, low-risk fix that mirrors an established, working pattern elsewhere in the codebase, is backward compatible (no public API/signature changes), and does not appear to break any existing tests.

### CI status

- `claude-review`: FAILURE — excluded per review policy (bot quota exhaustion, no actual review performed).
- `build` (Validation Build): **FAILURE** (cancelled after exceeding the 45-minute job time limit — `The job has exceeded the maximum execution time of 45m0s`). This is a real, non-excluded failing check blocking merge (`mergeStateStatus: BLOCKED`), independent of whether it's an infrastructure timeout rather than a code defect.
- `Semgrep Scan` / `Semgrep OSS`: PASS.
- `license/cla`: PASS.

### Decision

Per the review policy, since the `build` check is failing (excluding only `claude-review`), this PR cannot be approved even though the code change itself looks correct and low-risk. **Requesting changes** — please re-run/investigate the `build` job (it appears to have timed out rather than failed on a real compile/test error; a re-run may be sufficient) so it completes successfully before this is merged.

jshobe-inetsoft (approved):

PR #5236, opened 2026-09-14
Keep the modern table style when converting a crosstab to a freehand table

FigurePaper:
## Problem

With modern visualization and dark mode enabled, converting a crosstab to a freehand
table produced unreadable dark body text on the dark card. The crosstab it came from,
and a freehand table created from scratch, both rendered correctly — only the converted
table was wrong.

Two other symptoms came from the same cause: the converted table lost its header band
and picked up light-gray gridlines.

## Root cause

`VSLayoutTool.syncCellFormat` swaps the converted table's style from `"Default Style"`
to `inetsoft.report.style.CrosstabStyle` to mimic the crosstab's alignment.

The modern structure palette is overlaid only onto the shipped Default Style
(`DataVSAQuery.getViewTableLens` requires `DEFAULT_STYLE.equals(sname)`), so that swap
silently converts a marked table out of the modern interior. `CrosstabStyle` is
transparent with `Color.lightGray` gridlines and no foreground, so the table falls back
to the shipped dark body text — over the dark card the conversion copied across from
the crosstab.

## Fix

A marked table keeps its style and takes the explicit alignment branch a few lines
below, which is already what any non-default-styled crosstab uses. It writes the same
centered-header / right-aligned-data layout `CrosstabStyle` supplied, and `V_TOP` is
`VSFormat`'s unspecified default (`ALIGN = H_LEFT | V_TOP`), so alignment is visually
unchanged. Unmarked (legacy) conversions are untouched.

## Testing

New `ConvertToFreehandStyleTest` pins all three behaviors: a marked conversion keeps
Default Style, a marked conversion still aligns cells like CrosstabStyle, and an
unmarked conversion still takes CrosstabStyle. Written first and confirmed failing
before the fix.

Related suites pass: `DataVSAQueryDarkStructureTest`, `VSTableStructureDefaultsTest`,
`SeedChromeDefaultsTest`, `XTableStyleGroupSubtotalOrderTest`, `TableStyleTest` —
260 tests, 0 failures.

Manually verified in the app by the reporter.


FigurePaper:
## Code review

Two files, +107/-1: a one-condition guard in `VSLayoutTool.syncCellFormat` plus a new
`ConvertToFreehandStyleTest`. I traced the conversion path end to end and ran the tests locally.

### Correctness — verified

The root-cause narrative in the description holds up against the code:

* `DataVSAQuery.getViewTableLens` (line 133-139) overlays the modern structure palette only when
  `VizContext.of(tinfo).modern && TableDataVSAssemblyInfo.DEFAULT_STYLE.equals(sname)`. The guard
  added here uses the same `VizContext.of(info).modern` predicate, so the converter's decision and
  the renderer's decision are now driven by the same signal instead of disagreeing.
* `CrosstabStyle.Style extends Transparent` with `Color.lightGray` row/column borders and no
  foreground — which accounts for all three reported symptoms (dark body text falling back to the
  shipped default over the copied dark card, the lost header band, the light-gray gridlines).
* `VizContext.of(VSAssemblyInfo)` is null-safe and reduces to `info.getVizMark() != null`, so the
  guard cannot NPE on an unmarked or freshly built info.

**The mark is present when the guard reads it.** This was the one thing that could have made the fix
a no-op in the app while still passing a unit test that sets the mark by hand.
`ComposerVSTableService` calls `freehandInfo.copyInfo(info, false)` at step 5, before
`syncCellFormat` at step 6; `CalcTableVSAssemblyInfo.copyInfo` → `super.copyInfo` →
`VSAssemblyInfo.copyViewInfo` copies `vizMark` unconditionally (line 623-625), with no early return
on the path. So the freehand carries the crosstab's mark by the time the guard runs.

**Alignment equivalence holds.** `CrosstabStyle.getAlignment` returns `H_CENTER` when
`r < headerRowCount || c < headerColCount` and `H_RIGHT` otherwise; the fallback branch computes
`header` with the identical expression and writes `H_CENTER | V_TOP` / `H_RIGHT | V_TOP`.
`VSFormat.ALIGN` is `H_LEFT | V_TOP`, so the added `V_TOP` bit matches what an unspecified cell
already resolved to. The claim of no visual alignment change checks out.

### Regressions and backward compatibility

* Unmarked assemblies take the original branch unchanged — no behavior change for any saved
  viewsheet, and no XML or method-signature change.
* A marked crosstab whose style is *not* `Default Style` already took the explicit-alignment branch
  before this PR. The change routes marked + `Default Style` into that same long-shipped branch
  rather than inventing a new path, which is the low-risk shape for this fix.
* `crosstabStyle` is a local with exactly one reader (`if(crosstab && !crosstabStyle)`), so no other
  logic keys off the swap.
* Nothing else in the repo depends on the conversion producing `CrosstabStyle` — the only other
  mentions are `VSDialogService`'s style list and `TableStyleTest`'s registry list, neither affected.

### Tests

Verified rather than taken on faith:

* `ConvertToFreehandStyleTest`: 3 tests, 0 failures.
* Reverting only the production line and re-running makes it fail the way a real regression test
  should — `markedConversionKeepsDefaultStyle` (`expected: <Default Style> but was:
  <inetsoft.report.style.CrosstabStyle>`) and `markedConversionAlignsCellsLikeCrosstabStyle`
  (`expected: <2> but was: <0>`), while `unmarkedConversionStillTakesCrosstabStyle` passes in both
  directions. The test genuinely pins the fix and the legacy path.
* The related suites named in the description reproduce exactly: `DataVSAQueryDarkStructureTest`,
  `VSTableStructureDefaultsTest`, `SeedChromeDefaultsTest`, `XTableStyleGroupSubtotalOrderTest`,
  `TableStyleTest`, `ConvertToFreehandStyleTest` — **260 tests, 0 failures**.

No existing test is broken by the change.

### Security

No security surface: no user input, no serialization or permission change, no new I/O. The added
read is an in-memory enum check.

### Non-blocking observations

1. **`objfmt` is null-checked above and dereferenced below.** Lines 831-834 guard
   `if(objfmt != null)` before reading `objAlignment`, but the alignment branch at line 894
   dereferences `objfmt.getUserDefinedFormat()` unguarded. `FormatInfo`'s constructor seeds
   `OBJECTPATH` and `parseContents` adds to that map rather than clearing it, so it should never be
   null in practice — but this PR makes that branch the path taken by the common marked
   `Default Style` conversion instead of a minority case, so the asymmetry is worth tidying while
   you are here. (Related: the `if(objfmt == null && path != null && path.getType() == path.DETAIL)`
   block just above is dead — `path` was built with `TableDataPath.OBJECT`, so the type test can
   never be true. Pre-existing, no action needed.)

2. **String literal vs constant.** The edited line still compares against the `"Default Style"`
   literal while the new test correctly uses `TableDataVSAssemblyInfo.DEFAULT_STYLE`. Swapping the
   literal for the constant on the line you are already touching would remove the divergence.
   Pre-existing, and I appreciate the minimal-scope restraint if that was deliberate.

3. **Tighter alignment assertion.** `assertEquals(H_CENTER, header & H_CENTER, ...)` would also pass
   if additional horizontal bits were set. `assertEquals(H_CENTER | V_TOP, header)` pins the exact
   value the branch writes.

4. **Style nit.** The test imports `inetsoft.report.TableLens` but then fully qualifies
   `inetsoft.report.StyleConstants` inline; an import would read better.

5. **Already-converted dashboards.** A freehand converted before this fix has
   `inetsoft.report.style.CrosstabStyle` persisted and will still render with the bug; `Modernize`
   seeds chrome but does not rewrite table style, so it will not repair them. On an epic branch that
   is dev-only data and almost certainly fine — worth a conscious "no migration needed" call before
   this reaches a release branch. Similarly, `VizModernizeUtil.revert` clears the mark but leaves the
   explicit cell alignment written here, so a reverted conversion keeps centered headers and
   right-aligned data without `CrosstabStyle`'s gridlines. Cosmetic, and arguably the better outcome.

None of these block merging.

### Verdict

The change is correct, minimal, well targeted at the actual root cause, and properly covered by a
test that I confirmed fails without it. I found no blocking issues in the code.

### Merge readiness

| Item | State |
| --- | --- |
| Blocking issues in this review | None |
| `build` (Validation Build) | pass — 19m14s |
| `license/cla` | pass |
| `claude-review` | **fail — infrastructure, not code** |
| Conflicts | None — `mergeable: MERGEABLE` |

The red `claude-review` check is not a finding against this diff. The job errored 509 ms in, on
turn 1, with `total_cost_usd: 0` and an empty `modelUsage` — the review never ran at all.
`ANTHROPIC_API_KEY` is empty in the step environment, and the bot's own comment reads "Claude
encountered an error after 1s." Nothing in the change causes it, and re-running the job (or fixing
the workflow credential) should clear it.

Posted as a comment rather than a formal review because GitHub does not accept a self-review from
the PR author. From the code's side this is ready to merge once a second pair of eyes signs off and
`claude-review` is green.

jshobe-inetsoft (approved):

PR #5237, opened 2026-09-14
Bug #76329, fix cloud-secret Security Token orphaned in vault on upgrade

jkidacki-inetsoft:
## Summary

Follow-up to PR #5145 (already merged), which added OAuth 2.0 support to `SalesforceDataSource`/`RemedyforceDataSource`/`SalesforceReportsDataSource`. A later review round on the sibling `v1.1.x` PR (#5144) found that the earlier fix for the legacy Security Token migration (commit `1593324`) only covered data sources whose credential was stored **locally**. It missed data sources configured with "Use Secret ID" (cloud vault credential).

## Root Cause

Before #5145, these connectors used `CredentialType.PASSWORD_SECURITY_TOKEN`, so a cloud-secret-backed data source stored its security token inside the vault secret's own JSON, under the `CloudSecurityTokenCredential` schema (`"security_token"` key).

After #5145, `getCredentialType()` returns `PASSWORD_OAUTH2_WITH_FLAGS`, backed by `CloudPasswordAndOAuth2WithFlagCredentialsGrant`, which has no `securityToken` field and never reads `"security_token"` from the vault JSON. `securityToken` is now a plain field on `SalesforceDataSource`, populated only from local XML — so for a cloud-secret-backed data source (whose local XML never carried the token to begin with), the value is permanently orphaned in the vault after upgrade. The `securityToken` UI field was also hidden whenever "Use Secret ID" was selected (`visibleMethod = "useCredentialForPassword"`, which requires a *local* credential), so there was no way to view or re-enter it either.

## Fix

- `SalesforceDataSource.parseContents()` now falls back to reading the raw vault secret JSON (via a new protected, test-overridable `loadCloudSecurityToken()` seam) for the legacy `"security_token"` key when neither the new top-level nor the legacy-local XML location has it.
- Added `SalesforceDataSource.showSecurityToken()` (`!isOauth()`) and switched the `securityToken` view row in `RemedyforceDataSource`/`SalesforceReportsDataSource` to use it instead of `useCredentialForPassword()`, since the token is no longer part of the credential and should stay visible in Legacy Password mode regardless of "Use Secret ID".

## Test Plan

- [x] `RemedyforceDataSourceTest.legacySecurityTokenIsMigratedFromCloudSecret` — round-trips a fake vault secret JSON through `parseXML()` and asserts the security token is recovered
- [x] `RemedyforceDataSourceTest.securityTokenStaysVisibleInLegacyModeRegardlessOfCredentialSource` — visibility regression test
- [x] Existing `legacySecurityTokenIsMigratedFromPasswordCredentialNode` (local-XML case) still passes
- [x] `./mvnw test -pl connectors/inetsoft-rest` — all Salesforce/Remedyforce tests pass

Fixes Redmine #76329 (cloud-secret gap in the earlier fix).


jshobe-inetsoft (approved):
## Manual Review (replacing failed automated `claude-review` check)

The automated "Claude Code Review / claude-review" check failed after ~1s with no output, which is the known signature of the review bot hitting its usage quota rather than performing an actual review. This is a manual review performed in its place.

### Summary of the change

Follow-up to #5145: `SalesforceDataSource`/`RemedyforceDataSource`/`SalesforceReportsDataSource` switched from `CredentialType.PASSWORD_SECURITY_TOKEN` to `PASSWORD_OAUTH2_WITH_FLAGS`. The earlier legacy-token migration (`1593324`) only handled the security token stored in the local `<PasswordCredential>` XML node; it missed the case where the data source used a cloud "Use Secret ID" credential, which stored the token inside the vault secret's own JSON (`security_token` key, from the old `CloudSecurityTokenCredential` schema). This PR:

- Adds a fallback in `SalesforceDataSource.parseContents()` that reads the raw vault secret JSON (via a new `protected` seam, `loadCloudSecurityToken()`) for the legacy `security_token` key, only when neither the new top-level nor legacy-local XML location has it, and only when `!isOauth()` and the credential is a `CloudCredential` with a non-empty id.
- Adds `showSecurityToken()` (`!isOauth()`) and rewires the `securityToken` view row in `RemedyforceDataSource`/`SalesforceReportsDataSource` to use it instead of `useCredentialForPassword()`, since the field is no longer part of the credential and should stay visible/editable in Legacy Password mode regardless of "Use Secret ID".
- Adds test coverage in `RemedyforceDataSourceTest` for the cloud-secret migration path, the OAuth-mode negative case (vault must not be consulted), and the visibility regression.

### Correctness

- Verified there are only two concrete subclasses of the abstract `SalesforceDataSource` (`RemedyforceDataSource`, `SalesforceReportsDataSource`); both `@View1("securityToken", ...)` declarations were updated consistently to `showSecurityToken`. No missed call site.
- The vault fallback is correctly gated: `token == null` (both new and legacy-XML locations empty) AND `!isOauth()` AND credential is a `CloudCredential` with a non-empty id. OAuth-mode data sources never had a token to migrate and are explicitly excluded (covered by `oauthModeNeverConsultsTheVaultForALegacySecurityToken`), which also avoids an unnecessary billed vault call.
- Confirmed via `TabularDataSource.parseContents()`/credential parsing order that `getCredential()` is already populated (including its id) from XML by the time `super.parseContents(root)` returns and the new fallback branch runs, so the cast/id check is safe and not reading stale state.
- `node.get("security_token") != null` followed by `.textValue()` correctly handles a present-but-non-textual or JSON-null value (`textValue()` returns `null` rather than throwing), so no NPE risk there.
- Ran the full `RemedyforceDataSourceTest` suite locally against the PR branch (`a1e9e9e8a`, includes the second "don't consult the vault in OAuth mode" commit): all 4 tests pass (0 failures/errors), including the pre-existing local-XML migration test, confirming no regression.

- The vault read is wrapped in `catch(Exception ignore)`, deliberately leaving the token unset rather than failing data source parsing — reasonable, and consistent with `CloudCredential.fetchCredential()`'s general "don't blow up the whole data source over a secrets-manager hiccup" philosophy elsewhere in the codebase.
- Minor, non-blocking nit: unlike `CloudCredential.fetchCredential()`, which logs at `error` level when a failure isn't a `SecretsUnavailableException`, this new fallback silently swallows every exception with no logging at all. That could make a genuine, persistent migration failure (e.g. malformed vault JSON) invisible until someone notices the token never migrated. Consider at least a `debug`/`warn` log line.

### Minor documentation nits (non-blocking)

- `loadCloudSecurityToken()`'s Javadoc calls it a "Package-visible seam," but the method is `protected`, not package-private — the comment and modifier don't quite agree (protected is actually necessary here, since the anonymous test subclasses are in the same package as `RemedyforceDataSource`, not `SalesforceDataSource`).
- `showSecurityToken()`'s Javadoc links to "the class-level comment on `{@link #getSecurityToken()}}`", but that comment is a plain `//` line comment above the getter, not a Javadoc block — the `{@link}` itself still resolves fine, just the "class-level comment" description is slightly imprecise.

- Local-XML migration path (existing `legacySecurityTokenIsMigratedFromPasswordCredentialNode` test) is untouched and still passes.
- No changes to the credential type, wire format, or any other data source's behavior; the diff is scoped to the three Salesforce-family classes plus one test file.
- One behavioral note for awareness rather than a blocker: the vault fallback re-reads the vault secret on every `parseContents()` call as long as the migrated token hasn't been persisted back into local XML (which only happens if/when an admin re-saves the data source). This mirrors the existing `CloudCredential.fetchCredential()` pattern (also re-fetched on every parse/cache-refresh), so it's not a new class of cost, but it does mean the "orphaned" token isn't fixed once-and-for-all until the data source is explicitly re-saved. Worth keeping in mind for support/cost expectations, not a reason to block.

### Security

- No new secret-handling risk: the credential id used to query the vault comes from the already-parsed, admin-configured `CloudCredential`, the same id already used by the standard `CloudCredential.fetchCredential()` path elsewhere. No new attack surface introduced.

### Tests

- All 4 tests in `RemedyforceDataSourceTest` pass locally (`legacySecurityTokenIsMigratedFromPasswordCredentialNode`, `legacySecurityTokenIsMigratedFromCloudSecret`, `oauthModeNeverConsultsTheVaultForALegacySecurityToken`, `securityTokenStaysVisibleInLegacyModeRegardlessOfCredentialSource`).
- `connectors/inetsoft-rest` module builds cleanly with the change (`mvn install -DskipTests` succeeds).
- No existing test appears broken by this change.

### CI / mergeability

- All required checks are green except `claude-review` (excluded per review policy — it errored out at the usage-quota signature, not a real failure).
- `mergeStateStatus: BLOCKED` is solely due to the branch-protection "review required" gate (no reviews yet), which this review satisfies.

### Verdict

PR #5238, opened 2026-09-14
Bug #76488/#76485: guard the second assembly lookup in aggInfoInvalid()

larryliang-inetsoft:
## Claim boundary (please read first)

This is a **structural fix derived from code analysis, not a fix confirmed against a live reproduction** of Bug #76488 or Bug #76485. Neither bug has ever been reproduced — multiple live attempts across three investigation rounds (including one against the reporter's own `vs_calc.zip` asset) all failed to trigger the symptom. What this PR closes is a real, independently-demonstrable defect sitting directly on the established failure path. It should **not** be described as resolving either Redmine issue; both stay In Progress.

## Background: the established failure path

The Composer binding tree's client-side `_loadingTree` flag has exactly one clearer: receipt of a `RefreshBindingTreeCommand`. The entire server chain that produces one —

`VSBindingTreeController` → `VSBindingTreeControllerService` → `VSBindingTreeService` → `VSTreeHandler` → `CubeTreeModelBuilder`

— has **no exception handling anywhere**, and there is no STOMP-level global exception handler behind it (`GlobalExceptionHandler` only covers `@RequestMapping`, not `@MessageMapping`). So any uncaught exception on this path silently drops the response: no command, no error, no timeout, and the spinner is stranded forever. That much was confirmed independently by several prior passes.

## The defect this PR fixes

`CubeTreeModelBuilder.getCubeTreeModel()` calls `processor.aggInfoInvalid(info)` whenever `isValidAggregateInfo()` finds the crosstab's stored `AggregateInfo` out of step with the source table's column selection (`CubeTreeModelBuilder.java:134-136`, `281-345`).

Both `Processor` implementations in `VSTreeHandler` then re-resolve the assembly **by name, a second time** and dereference the result with no null check:

- `VSTreeHandler.java:294-296` (table/crosstab path) — `vs.getAssembly(name)` → `assembly.getAbsoluteName()`
- `VSTreeHandler.java:123-124` (chart path) — `vs.getAssembly(name)` → `chart.setVSAssemblyInfo(info)`

Each outer method already guards its *own* first lookup (`cass == null` in `getTableTreeModel()`, `chart == null` in `getChartTreeModel()` — the former added by #5059). This deeper second lookup, inside the `Processor` callback, was left unguarded — a time-of-check-to-time-of-use gap. A viewsheet disposed or reset between the two lookups produces an NPE out of the whole unguarded chain rather than a response.

Returning early instead leaves `getCubeTreeModel()` free to finish and return a tree, so the client still receives its `RefreshBindingTreeCommand` and clears the loading state. Note this is a meaningfully different outcome from #5059's guard: that one converts an NPE into a `null` return (which still produces a response, since `VSBindingTreeControllerService` sends `RefreshBindingTreeCommand(null)` unconditionally); this one lets the request complete normally.

## Why the deeper site, specifically

This was the surface flagged as unexamined by the prior round's own notes: nobody had traced `CubeTreeModelBuilder.getCubeTreeModel()` or the `Processor` callbacks it invokes. Tracing them end to end also settled two other open questions, neither of which turned out to be the mechanism:

- **Write-lock starvation — refuted.** `VSRefreshService.refreshViewsheet()` releases its write lock in a `finally` block (`VSRefreshService.java:197-212`) and re-throws, so a later reader can never be blocked forever by that path.
- **Async-refresh exception swallowing — real, but not a hang.** `VSRefreshController` never awaits the `Future` from `refreshViewsheetAsync`, so exceptions inside the async refresh are silently discarded. That is a genuine gap, but it happens *after* the lock is released, so it can only leave inconsistent data behind — it cannot itself strand the response. Left alone here; worth its own ticket.

## Changes

- `core/src/main/java/inetsoft/web/binding/handler/VSTreeHandler.java` — null guard on the second assembly lookup in both `Processor.aggInfoInvalid()` implementations (crosstab/table and chart), mirroring the guard pattern each outer method already uses.
- `core/src/test/java/inetsoft/web/binding/handler/VSTreeHandlerAggInfoInvalidTest.java` — new regression test driving `isValidAggregateInfo()` to `false` (non-empty `AggregateInfo` whose bound-ref count doesn't match the table's column count) with the assembly resolving on the first lookup and gone on the second, which is the interleaving a concurrent refresh produces.

## Verification

- `./mvnw -pl core test -Dtest=VSTreeHandlerAggInfoInvalidTest,VSTreeHandlerTest` → **4/4 pass** (2 new + the 2 pre-existing from #5059, no regressions).
- Test confirmed load-bearing: with the crosstab-side guard temporarily removed, the new test fails with exactly the predicted `NullPointerException: Cannot invoke "inetsoft.uql.viewsheet.VSAssembly.getAbsoluteName()" because "assembly" is null`; the control case (assembly stays resolvable) passes either way. Guard restored, all 4 green again.
- No live/end-to-end verification — see the claim boundary above.

Full analysis: `docs/teams/2026-09-13-bug-76488-76485-round2/04-root-cause-and-fix.md`.


jshobe-inetsoft (approved):
## Manual Review (standing in for the failed `Claude Code Review / claude-review` bot check)

The automated `claude-review` check failed after ~36s, which is the known signature of the review bot hitting its usage quota rather than performing an actual review. This is a manual review performed in its place.

### Summary of the change

Adds a null guard on the *second* (in-callback) lookup of the target assembly inside both `CubeTreeModelBuilder.Processor.aggInfoInvalid()` implementations in `VSTreeHandler` (chart path at `getChartTreeModel()` and table/crosstab path at `getTableTreeModel()`). Each outer method already guards its *first* lookup (`chart == null` / `cass == null`), but a second, unguarded `vs.getAssembly(name)` re-resolution inside the `Processor` callback could return `null` if the viewsheet was concurrently disposed/reset between the two lookups (e.g. a concurrent refresh), producing an unguarded NPE that silently dropped the STOMP response (no exception handler exists anywhere on this call chain), stranding the Composer binding-tree loading spinner client-side.

The PR is explicit that this is a structural fix derived from code tracing, not a confirmed reproduction of Bug #76488/#76485 — a refreshingly honest claim boundary that I verified against the code.

### Verification performed

- Read the full diff and the surrounding code in `VSTreeHandler.java` and `CubeTreeModelBuilder.java` (`getCubeTreeModel()`) to confirm that returning early from `aggInfoInvalid()` leaves `getCubeTreeModel()` free to finish building and returning a tree using state captured earlier in the builder — it does not depend on the second lookup succeeding. This matches the PR's claim.
- Built `core` from the PR branch (`./mvnw -pl core -am -DskipTests install`) — succeeds, confirming syntactic correctness (braces/imports/generics all fine).
- Ran `VSTreeHandlerAggInfoInvalidTest` and `VSTreeHandlerTest` — **4/4 pass**, no regressions.
- Confirmed the new test is load-bearing: with the crosstab/table-path guard temporarily removed, `getTableTreeModel_doesNotThrow_whenAssemblyDisappearsBeforeAggInfoInvalid` fails with exactly the predicted `NullPointerException: Cannot invoke "inetsoft.uql.viewsheet.VSAssembly.getAbsoluteName()" because "assembly" is null`, while the control test (`...AssemblyRemainsResolvable`) still passes. Restored the guard afterward.

### Findings

**No blocking issues.**

Minor, non-blocking observations for a possible follow-up (not required for this PR):

1. **Asymmetric test coverage.** The new test file only exercises the table/crosstab-path guard (`getTableTreeModel`'s `Processor.aggInfoInvalid`); the chart-path guard (`getChartTreeModel`'s `Processor.aggInfoInvalid`) added in the same diff has no dedicated regression test, only the same manual reasoning by analogy. The fix itself is trivial and mirrors an already-verified pattern, so this isn't blocking, but a companion test for the chart path would make the coverage symmetric with the fix.
2. **No debug logging on the new guards.** The pre-existing outer guards (`cass == null` in `getTableTreeModel`, added by #5059) log at DEBUG with a `BUG76488-76485-TRACE` tag before returning. The two new inner guards added here don't log anything, which could make a live recurrence of the underlying race harder to distinguish from the outer-guard case in production logs. Purely a debuggability nit, not a correctness issue.

### Other review criteria

- **Correctness/completeness:** The two added guards are consistent with each other and with the existing outer-guard pattern in the same class. No half-applied changes.
- **Error handling/edge cases:** This *is* an edge-case fix (TOCTOU race on assembly resolution); the guard is the correct minimal response — skip the update and let the tree still be returned rather than propagating an NPE.
- **Readability/maintainability:** Clear, well-commented (each guard explains why the null can occur and what the early return means for the caller).
- **Security:** No security-relevant surface touched; purely a defensive null check on internally-resolved objects.
- **Regressions:** None found. The early-return path matches the pre-existing `box.isEmpty()` early-return immediately above it in both methods, so this isn't a new code shape — it's applying an existing local pattern one call deeper.
- **Backward compatibility:** No public API/signature changes; behavior only changes in the previously-crashing edge case.
- **Test breakage:** None — all 4 relevant tests (2 new + 2 pre-existing) pass; no existing tests were modified.

### CI/merge status

- All checks pass except `claude-review`, which is being disregarded per the known quota-failure signature (fails in ~36s with no real review performed).
- `mergeStateStatus: BLOCKED` is solely due to the branch-protection requirement for a review (`reviewDecision: REVIEW_REQUIRED`, zero prior reviews) — this review satisfies that.

### Decision: **Approve**

All three approval conditions are satisfied: no blocking issues in the code, no failing checks other than the disregarded `claude-review`, and no merge conflicts.

PR #3, opened 2024-11-20
Name as "Open Source"

swaker854:
change name

PR #5244, opened 2026-09-15
fix(wiz): let $(ComponentName) dynamic references through PropertyPath's visible gate and BindableColumns' existence checks (#76641)

Nelson-cao:
## Summary

Companion to `inetsoft-technology/stylebi-wiz` PR (plugin/composer side) for Redmine #76641 — StyleBI's own Composer UI lets a dialog property's "Variable" button, or a chart/crosstab/table shelf field, be pointed at `$(FormComponentName)`. The wiz agent tool path had three independent gates that rejected this string before it ever reached the existing, already-native `DynamicValue`/`VSUtil.isVariableValue` resolution mechanism:

1. **`PropertyPath.java`** — `requireAllowedValue()`'s `CONSTRAINED_STRINGS` gate hard-rejected any value for `visible` outside its fixed enum (`true/show/false/hide/...`). Added a narrow `DYNAMIC_CAPABLE_CONSTRAINED_STRINGS` bypass, scoped to `visible` only — deliberately NOT extended to `trendLineType`/`position` (the other two `CONSTRAINED_STRINGS` entries), which are not `DynamicValue`-backed and would silently store an unresolvable value.
2. **`BindableColumns.require()`** — checked a shelf field's `column` against the assembly's already-established source table's real columns, rejecting `$(...)` as a typo. Bypassed the same way.
3. **`BindableColumns.hasAll()`/`requireHasAll()`** — a second, independent pair of checks in the same class, reached only when establishing a source on a brand-new/unsourced assembly (a different code path from #2, found via live testing a crosstab rather than a chart). Same bypass, with an added guard: a solo dynamic-reference field must not by itself manufacture a table match during inference — only literal fields decide.

Each fix duplicates `VSUtil.isVariableValue`'s one-line predicate locally (`isDynamicVariableReference`) rather than importing `VSUtil` directly — `VSUtil`'s static initializer reaches for a live Spring `ApplicationContext`, which crashes in these classes' existing Spring-free unit test files (`PropertyPathTest`/`BindableColumnsTest`, no Spring bootstrap). Confirmed live before choosing this over adding Spring bootstrap to ~70 existing zero-dependency tests.

## Test plan

- [x] `PropertyPathTest`: 51/51 pass (47 pre-existing unaffected + 4 new).
- [x] `BindableColumnsTest`: 32/32 pass (20 pre-existing unaffected + 12 new, across two fix rounds).
- [x] Regression guards included in both: the ORIGINAL cases these classes exist for (a real nonexistent/wrong-table literal column, `trendLineType`/`position` still rejecting `$(...)`) still throw exactly as before.
- [x] Live-verified against a real running StyleBI instance (via the `stylebi-wiz` plugin, hot-swapping each compiled `.class` and restarting between rounds): `visible`/`enabled` on `set_assembly_properties`, a chart Y-axis measure swap on an already-sourced chart (`set_chart_shelf`), a crosstab row-dimension swap on a brand-new/unsourced assembly (`set_table_fields` — the case that caught fix #3), and a chart color aesthetic (`set_aesthetic_field`) — all accepted, round-tripped, and rendered with the reference correctly resolved to the driving component's current value.

Full process record, including the live-verification transcript that found gaps #2 and #3 (not caught by source reading alone): `stylebi-wiz` repo, `docs/teams/2026-09-14-bug-76641-form-variable-binding/`.


Nelson-cao:
## Review round 1

Head reviewed: `0b9b73cf034406f7021ee559ecb62930c8a14eda` (repo `inetsoft-technology/stylebi`, base `stylebi-wiz-main-rebase`).

### What was checked

- Re-read `PropertyPath.java` and `BindableColumns.java` in full at the reviewed head, against the diff.
- Ran `PropertyPathTest` + `BindableColumnsTest` myself from a clean working tree: **83/83 pass** (51 + 32), matching the PR's claimed counts.
- Traced `trendLineType`/`position` independently rather than trusting the PR's own Javadoc claim: `ChartLinePaneModel`→`PlotDescriptor.trendline` and `LegendFormatDialogModel`→`LegendsDescriptor.layout` are both plain `int` fields with no `DynamicValue`/`executeDynamicValue` path anywhere — confirmed correct to exclude them from the bypass.
- Traced every caller of `BindableColumns.require`/`hasAll`/`requireHasAll`: `BindingAgentController.resolveSourceTable()` is the single shared choke point for `chart/shelf`, `chart/single-shelf`, `chart/aesthetic-field`, `table/fields`, and `table/field/add` (5 endpoints) — all get every fix automatically. `CalcTableService.requireBindableColumn()` (backing `set_cell_binding`) also calls `require()` directly and is fixed for free at the Java layer. No caller missed.
- Adversarial case from the dispatch: two candidate tables, one dynamic field + one literal field present in BOTH tables. Hand-traced `hasAll`'s actual logic — the dynamic field is unconditionally skipped and contributes nothing to which candidates pass, so the literal field alone still produces two candidates and still throws "exists in more than one table," exactly as pre-fix. The bypass cannot let one table win by iteration order because it never participates in the decision. (No shipped test covers this exact combination, but the code path is unambiguous by inspection.)
- Non-ASCII: `isDynamicVariableReference` is a plain `startsWith`/`endsWith` check, no regex, no case-folding — `"$(中文名)"` behaves identically to an ASCII name.
- Regression guard: the ORIGINAL cases each gate exists for (nonexistent column, wrong-table column, `visible`/`trendLineType`/`position` domain rejection) are still tested and still throw.
- The "duplicate `VSUtil.isVariableValue` three times instead of importing it" decision: verified `VSUtil.scriptDeps` is a class-level static field whose constructor reaches Spring, so *any* static call into `VSUtil` (even to an unrelated method) forces the same crash in these Spring-free test files — there is no narrower import that avoids it. The predicate is 2 lines, stable, and cited back to the real method in every copy. Reasonable call, not filed as a finding.

### Findings on this diff

None at reviewable confidence.

### Swept for the same pattern elsewhere

`WizAutoBindingService.derivedFieldConfigsWithin()`/`selectBindColumns()` do the same shape of check (a field's column vs. a `Set<String>` of real worksheet columns, no dynamic-reference awareness) during chart-type-change auto-binding — a different subsystem entirely, not touched by this PR. **Not confirmed as an actual defect**: `deriveFieldConfigs` sources its list via `collectFlatBinding`, which per its own comments "prefers RT refs" (already-resolved runtime values) — if that means a dynamic-reference field reports its resolved value rather than the literal `"$(...)"` string at this point, this concern evaporates. Flagging as a candidate follow-up, not a blocker — this PR's own scope (the three gates it names) is correctly bounded.

### Verdict: approved

Full findings/verification detail: `06-review-java-r1.md` in the reviewed worktree.

fredaan (approved):

PR #5247, opened 2026-09-15
Surface the column-limit warning through the agent worksheet preview endpoint

fredaan:
## Summary

- Redmine #76648 (WBT-002): `GET /api/wiz/v1/agent/worksheet/{token}/preview` (the plugin's `preview_worksheet_data` tool) silently returned fewer columns than the worksheet model reports whenever the executed table's column count exceeds the organization's column limit (`Util.MAX_COLUMN_COUNT`, default 200, `Util.getOrganizationMaxColumn()`). `AssetQuerySandbox.getColumnLimitTableLens` already raises a warning for this exact condition via `Tool.addUserMessage(Util.getColumnLimitMessage())`, but that pending message is only ever read by the interactive Designer UI's own request/response cycle — this REST agent endpoint had no equivalent mechanism to capture and forward it, so the truncation was completely silent.
- `WorksheetPreviewService.preview()` now clears any stale pending message before executing (guards against a pooled thread carrying over an unrelated request's message) and, after building the result rows, reads back `CoreTool.getUserMessage()` — capturing the column-limit warning whenever it fired. Return type changes from a bare `List<Map<String, Object>>` to a new `PreviewResult(rows, warnings)` record; `WorksheetAgentController.preview()` now returns that record directly, so the endpoint's JSON body becomes `{ rows, warnings }` instead of a bare array.
- The organization column limit itself (`Util.MAX_COLUMN_COUNT` / `getColumnLimitTableLens`) is unchanged and still applies to every other caller exactly as before (~30 call sites: CSV import, chart/crosstab binding, SQL query limits, etc.) — this PR only makes the existing warning visible on this one REST path, which previously had no way to surface it. This endpoint has exactly one caller in this codebase (the controller itself), so the shape change is fully contained.

## Test plan

- [x] `WorksheetPreviewServiceTest`: all 8 existing tests updated for the new `PreviewResult` return type; added 3 new tests — `surfacesPendingUserMessageAsWarning` (simulates the real `CoreTool.addUserMessage` call `getColumnLimitTableLens` makes), `warningsAreEmptyWhenNothingWasRaised`, `doesNotLeakAnUnrelatedPriorMessageFromTheSamePooledThread`. 19/19 passing.
- [x] `WorksheetAgentControllerTest`: both `preview` tests updated for the new return type. 183/183 passing.
- [x] `mvnw -pl core -o compile` clean.
- [x] Companion `stylebi-wiz` PR relays `{ rows, warnings }` through `wiz-services` and surfaces it (with an explanatory note) in the plugin's `preview_worksheet_data` tool response.


fredaan:
## Review round 1 (posted as comment -- self-authored PR, `gh pr review --approve` is rejected by GitHub) -- head `e96e7cc0f`

Reviewed together with the companion PR (stylebi-wiz #2529) as two halves of one fix for
WBT-002 (Redmine #76648).

**Checked:**
- `WorksheetPreviewService.preview()` clears any pending message (`CoreTool.clearUserMessage()`)
  BEFORE `box.getTableLens(...)` and captures it (`captureWarnings()` -> `CoreTool.getUserMessage()`)
  AFTER building rows -- verified this ordering by reading the full current file, not just the
  diff hunks. A stale message from an earlier request on a pooled thread cannot leak into this
  call's result.
- `CoreTool.getUserMessage()` itself reduces ALL queued messages via `UserMessage::merge`
  (`CoreTool.java:3573-3579`), so a join hitting the cap on more than one sub-table in the same
  execution is not silently dropped -- pre-existing behavior this fix correctly relies on.
- The 3 new tests actually assert what the PR description claims, not just exist:
  `doesNotLeakAnUnrelatedPriorMessageFromTheSamePooledThread` queues a message BEFORE calling
  `preview()` and asserts it's not attributed to this call -- a real assertion of the
  clear-then-capture order, not a renamed existing test.
- Breaking-change claim verified independently: grepped all of `community` for callers of
  `WorksheetPreviewService.preview(...)`/`previewService.preview(...)` -- only
  `WorksheetAgentController.preview()` calls it in production code (also updated here); the
  only other hit is a doc-comment describing a deliberately out-of-scope future integration, not
  a real caller. Shape change is fully contained.
- Cross-repo shape consistency: this record's `{rows, warnings}` matches wiz-services'
  `WorksheetPreviewResult` field-for-field (checked in the companion PR).
- Alternate-scenario sweep: grepped for every other `getTableLens(..., RUNTIME_MODE)` call site
  in `community/core/.../web/wiz`. Found one this PR does not cover --
  `RawDataService.writeWorksheetTableCsvStream` (CSV export under
  `/api/wiz/profiling/rawdata/export/...`, the data-annotation profiler) has the identical
  silent-truncation exposure with no warning capture. It's a different subsystem, not reachable
  through the agent/plugin surface this bug is about, so it's out of scope here -- flagging as a
  follow-up for whoever next touches CSV export, not a finding against this diff.

**CI:** `build`, `Semgrep OSS`, `Semgrep Scan`, `license/cla` all pass. `claude-review` shows
`failure`, but the job log shows it errored in 0s (empty `ANTHROPIC_API_KEY`) before producing
any review content -- an infra failure, not a content finding.

**No blocking findings. Approve-worthy.** Per the batch policy for this bug, this PR will not be
merged by this review -- merging is reserved for the user.

Nelson-cao (approved):

PR #5248, opened 2026-09-15
Bug #76663, mark non-anchor calc-table cells merged-into their anchor

lizhepu:
## Summary
- `get_calc_layout`/`get_cell_binding` (`CalcTableService.readLayout`/`readCell`) reported a merged-away, non-anchor calc-table cell's binding as its own -- but `TableLayoutHandler.mergeCells`/`CellInfoHandler.spreadValue` overwrite that cell's stored binding with a clone of the anchor's own (post-merge) binding, restoring only the cell's pre-merge `cellName`. A caller reading the layout right after a merge saw a cell's old name paired with the anchor's data, with nothing on the cell's own entry indicating it had been merged away.
- Fixes this the same way the Composer's own design-time table model already does for a human editor (`CalcTableLayout.buildBaseInfo`, which blanks a merged-away cell's display text and points it at the anchor via `baseInfo` so it's never independently selectable): a non-anchor cell now reports `mergedInto: {row, col}` pointing at its anchor and `binding: null`, instead of describing the anchor's cloned binding as its own.
- Adds two regression tests to `CalcTableServiceTest` covering both `readLayout` (the grid read) and `readCell` (the single-cell read), verifying the non-anchor cell reports `mergedInto`/`binding: null` and that `getCellBindingInfo` is never even called for it.

## Root cause
Only the anchor cell of a merged span carries a non-null `Dimension` from `TableLayout.getSpan`; every other cell in the span carries none, so nothing on a non-anchor cell's own entry ever indicated it was part of a merge. `readLayout`/`readCell` described every cell's stored `TableCellBinding` the same way regardless, exposing the anchor-cloned, stale-named intermediate state that `spreadValue` writes into non-anchor cells at merge time -- an internal representation detail the Composer's own UI already goes out of its way to hide from a human, but was never redacted on this cell-addressed read path.

## Test plan
- [x] `./mvnw -pl core -Dtest=CalcTableServiceTest test` -- 52/52 pass (50 existing + 2 new)
- [x] Live-reproduced against a local StyleBI instance via the stylebi-wiz plugin's `composer-chat` MCP tools before this fix (merge then get_calc_layout showed the corrupted state exactly as filed); not yet re-verified live after the fix (would need a running local stack rebuilt from this branch)

Redmine: #76663


lizhepu:
## Round 1 review — head commit `90890afc0`

(Note: GitHub refuses a formal \`--approve\` review from this account because it's also the PR author — \`gh pr review --approve\` failed with "Can not approve your own pull request". Posting the review as a comment instead; verdict below is APPROVE, no blocker/important findings.)

Reviewed against `01`–`04` in the debug-workflow's investigation record (charter, diagnosis, refute, live-confirmation, fix note) and this repo's review-workflow adversarial checklist.

**Checked:**
- **Mechanism, not symptom**: `mergedIntoMap` keys off `TableLayout.getSpan`/anchor-rectangle geometry, the actual place StyleBI records a merge — not a literal coordinate or column name. Confirmed by reading `MatrixOperation.setSpan` (`report/MatrixOperation.java:231-286`) myself: a merge/re-merge always leaves exactly one non-null span per region, at the union rectangle's own top-left, with any absorbed sub-anchor's span explicitly nulled — so the anchor-scan approach holds for re-merge/merge-of-merges, not just a single fresh merge.
- **1xN/Nx1 spans**: `setSpan` only nulls a span when `width==1 && height==1` (or a non-positive dimension) — a row-only or column-only merge keeps a normal non-null `Dimension`, so the rectangle math in `mergedIntoMap` is unaffected. The `(span.width <= 1 && span.height <= 1)` guard in `mergedIntoMap` is therefore defensive/unreachable in practice, not a functional issue.
- **Other call sites**: grepped `core/src/main` for `CalcTableService` — its only consumer is `BindingAgentController` (`calcLayout`/`calcCell` endpoints), which is exactly `readLayout`/`readCell`. No other place in `web/wiz/binding/` (`CalcCellVocabulary`, `FieldRefFactory`, etc.) independently walks `TableLayout` cells the same naive way, so there's no sibling "same bug, other site" left unfixed.
- **Blast radius**: confirmed independently (not just trusting the diagnosis) — `plugin/composer/src/tools/calcTools.ts`'s `get_calc_layout`/`get_cell_binding` tools are bare passthroughs with no schema validation on the response, and every place that reads a cell's `binding` field afterward (`boundCellsInRange`, `survivingBinding`, `set_calc_cell_script`'s current-binding read) does so via `?? {}`, which absorbs `null` safely — no unguarded `.binding.<field>` access exists on the plugin side that would break on `binding: null`.
- **Scope (readCell too)**: extending the same fix to `get_cell_binding` is correct scope, not overreach — it's literally the same defect (same stored, corrupted `TableCellBinding`) read through the single-cell path instead of the grid path; leaving it unfixed there would just reopen the bug via a different tool.
- **Analogy to the Composer's own UI handling**: verified `CalcTableLayout`'s span-scan (`web/binding/model/table/CalcTableLayout.java:175-210`) does the same anchor/non-anchor walk to blank a merged-away cell's display text for a human editor — confirming this mirrors existing, working precedent in this codebase. (Minor, non-blocking: the new code's javadoc and the PR's own fix note call this method `buildBaseInfo`; it's actually named `initBaseInfo`. Doesn't affect correctness, just a citation slip worth fixing in a follow-up if anyone touches this comment again.)
- **Tests**: ran `./mvnw -pl core -Dtest=CalcTableServiceTest test` myself from a fresh worktree off this branch — **52/52 pass**, matching the PR description. Both new tests exercise the actual mechanism (a stubbed `getSpan(0,0)`, not a hardcoded coordinate literal) and assert `getCellBindingInfo` is never called for the merged-away cell, which would catch a regression to the old "describe every cell's stored binding" behavior.

**Not a blocker, worth a follow-up if anyone's in this file again**: test coverage is a single 1-row/2-col merge in each test; a multi-cell span (e.g. 2x2) or a merge-of-merges isn't directly exercised by a test, only by source-reading (`02-refute.md`). Given `mergedIntoMap`'s rectangle-scan is straightforward and the geometry primitives it depends on (`getSpan`/`setSpan`) are pre-existing, well-established code, I don't think this rises to changes-requested, but it's a gap worth closing later.

No blocker/important findings. Verdict: **APPROVE**.

PR #5250, opened 2026-09-15
Bug #76497: provide GlobalSubmitService on the Portal viewer's edit route

larryliang-inetsoft:
## Summary

Bug #76497 was reopened after PR #5055 fixed a real, separate server-side NPE in `VSFilterRecommendationFactory.createDataRefs()` — the reporter confirmed the Wizard's Filter recommendation preview (Range Slider / Selection List) still went blank when editing an existing chart from the Portal viewer's pencil icon, even though the same recommendation worked fine when built fresh inside Composer's own "New Dashboard" wizard flow.

Root cause: `VSRangeSlider` and `VSSelection` both require `GlobalSubmitService` as a hard (non-`@Optional()`) constructor dependency. `GlobalSubmitService` has no `providedIn: "root"` and is provided only by `ComposerMainComponent` and `ViewerAppComponent`. `viewer-edit.routes.ts`'s `edit` route is a routing *sibling* of the normal viewer route (which hosts `ViewerAppComponent`), not a descendant — so neither provider was ever reachable there, and Angular threw `NG0201` (`NullInjectorError`) the instant the wizard's preview switched to Range Slider or Selection List, leaving the canvas blank. `VSChart` has no such dependency, which is why Chart rendered fine in the same session.

This is a distinct, second defect from the one PR #5055 fixed — a frontend Angular dependency-injection wiring gap, not a server-side NPE — that happened to produce an outwardly identical symptom.

## Fix

Add `GlobalSubmitService` to `viewerEditRoutes`'s `providers:` array in `viewer/viewer-edit/viewer-edit.routes.ts`, mirroring how it's already provided in `ComposerMainComponent` and `ViewerAppComponent`. This also covers Calendar and Selection Tree, which share the same gap via the same missing provider (both were previously untested live).

## Test plan

- [x] Added `viewer-edit.routes.spec.ts`, a TestBed-level regression test that configures the real `viewerEditRoutes[0].providers` array and asserts `GlobalSubmitService` resolves. Confirmed **red** before the fix (`NG0201: No provider found for `_GlobalSubmitService``), **green** after.
- [x] Re-ran the affected suites (`range-slider`, `selection`, `calendar`, `viewer-edit` specs) — 34 tests, no regressions.
- [x] Live-reproduced the original bug report end-to-end against a running server before diagnosis began (portal pencil-edit → Wizard → Filter → Range Slider/Selection List → blank canvas + `NG0201` in console).
- [ ] The exact live-browser sequence was not re-run against a rebuilt frontend after the fix (this environment's running container has no accessible static-asset mount for hot-patching, and a full image rebuild wasn't done). A maintainer merging this should ideally do one live click-through, given the prior fix for this same bug report also looked complete under its own test and turned out not to be.

Full investigation archive (three independent parallel investigations converging on the identical root cause): `docs/teams/2026-09-14-bug-76497/` in the enterprise repo.


jshobe-inetsoft (approved):
## Manual Code Review (replacing failed automated Claude Code Review)

The automated **Claude Code Review / claude-review** check failed after ~30s with no findings posted ("No buffered inline comments") — this matches the known signature of the review bot hitting its subscription usage quota rather than performing an actual review. This is a manual review performed in its place.

### Summary of the change

Adds `GlobalSubmitService` to the `providers:` array of `viewerEditRoutes[0]` in `viewer/viewer-edit/viewer-edit.routes.ts`, and adds a new regression spec (`viewer-edit.routes.spec.ts`) that boots a `TestBed` with the route's real provider array and asserts `GlobalSubmitService` resolves.

### Root cause verification

I independently verified the root-cause analysis in the PR description against the code:

- `GlobalSubmitService` (`vsobjects/util/global-submit.service.ts`) is a plain `@Injectable()` with **no** `providedIn: "root"`, so it must be explicitly provided somewhere in the injector tree.
- `VSRangeSlider` (`vs-range-slider.component.ts:193`) and `VSSelection` (`vs-selection.component.ts:309`) both inject it as a **required** constructor parameter (no `@Optional()`), unlike other optional deps in the same constructors (e.g. `selectionMobileService` on `VSSelection` *is* `@Optional()`).
- It is provided today only at `viewer-app.component.ts:299` (`viewer-app` component, used by the "view" route via `ViewerViewComponent`) and at `composer-main.component.ts:253` (`ComposerMainComponent`).
- `viewer-edit.routes.ts`'s `edit` route renders `ViewerEditComponent`, which embeds `<vs-wizard>` directly — it never instantiates `viewer-app`, so `GlobalSubmitService` was never reachable from that injector subtree. This matches the PR's explanation precisely.
- I confirmed there is no other route/component in the codebase that renders `<vs-wizard>` and was missing this provider (`grep` for `VsWizardComponent`/`<vs-wizard` usage turns up only `composer-main` and `viewer-edit`), so the fix appears complete — no other reachable code path has the same gap.

### Testing performed

- Checked out the PR branch locally and ran the new spec directly (`ng test portal --include=".../viewer-edit.routes.spec.ts"`): **passes** (1/1).
- Ran the broader affected suites (`selection`, `range-slider`, `calendar` under `vsobjects/objects/`): **33/33 pass**, no regressions observed.
- Verified `package.json`/`package-lock.json` are unchanged by this PR (no new dependency surface).

### Review findings

- **Correctness**: The fix is correct and directly addresses the confirmed root cause (a DI wiring gap, not the previously-fixed server-side NPE from PR #5055). It follows the exact existing pattern used by `ComposerMainComponent`/`ViewerAppComponent` rather than introducing a new pattern (e.g., it deliberately does *not* make `GlobalSubmitService` a root/singleton service, which is right — that would risk leaking state such as `hasUnapplyData`/`globalSubmitStates` across independently-scoped viewer/composer/wizard instances).
- **Scope**: Minimal and targeted — a two-line source change plus a new unit test. No unrelated refactoring.
- **Test quality**: The new spec is a real regression test (asserts against `viewerEditRoutes[0].providers` directly, i.e. it exercises the actual production route config, not a re-declared copy), and the PR author confirmed it was red before the fix and green after. It also carries a clear comment explaining the bug and pointing to the root-cause trace.
- **Readability/maintainability**: Import ordering and code style match the surrounding file conventions. The regression spec's comment is a genuine aid for future maintainers who might otherwise reintroduce this gap.
- **Error handling/edge cases**: N/A — this is a DI provider registration fix, not logic with error paths. `GlobalSubmitService`'s own internal state handling is unchanged.
- **Security**: No security-relevant surface touched (no new user input handling, no auth/permission logic, no serialization changes).
- **Backward compatibility**: None broken — this purely adds a provider to a route-scoped injector; it cannot affect any other route or any external API/consumer.
- **Regression risk**: Very low. Adding a provider to a route's DI array only affects resolution within that route's component subtree and cannot shadow or conflict with providers in unrelated routes (Angular route-level providers are scoped to that route's injector). Confirmed no double-registration/broken singleton concerns since `ViewerEditComponent` does not also instantiate `viewer-app` (which separately provides the same service for the "view" route).
- **Known gap called out honestly by the author**: The PR's own test plan notes the live end-to-end click-through was not re-verified after the fix in this environment. This is a reasonable, disclosed limitation for a one-line DI fix backed by a real TestBed regression test that reproduces the exact `NG0201` failure mode; it does not block merging given the strength of the automated verification.

No blocking issues were found.

### CI / mergeability status

- All checks pass except **Claude Code Review / claude-review**, which errored out before performing any review (quota/error signature described above) — excluded from this decision per review policy.
- `mergeable: MERGEABLE`, no merge conflicts. (`mergeStateStatus: BLOCKED` is solely due to the branch-protection "review required" gate, which this review satisfies.)

### Verdict

**Approve.** No blocking issues, all non-excluded checks pass, no merge conflicts.

PR #5251, opened 2026-09-15
Bug #76660: refuse a non-static shape/size frame on unbound measures

fredaan:
## Summary
- `ChartAestheticMutator.setFrame`'s field-less per-measure broadcast branch guarded `color` and `line`/`texture` (VCA-006) against a non-static frame, but not `shape` or `size` — despite `VSShapeFrameStrategy`/`VSSizeFrameStrategy.createCombinedFrame` having the byte-identical unconditional cast to `StaticShapeFrame`/`StaticSizeFrame`. A categorical `shape`/`size` frame on a field-less channel wrote straight onto every measure, `set_visual_frame` answered `ok`, and the next render crashed with a `ClassCastException` — shape's cast fires only on a point (or point-line radar) chart; size's fires on essentially any chart type except stock and Map+Polygon.
- Adds `requireStaticShapeOrSizeOnMeasures`, mirroring `requireStaticLineOrTextureOnMeasures`, called alongside the existing color/line-texture guards. Unlike line/texture (which share shape's field and both point their refusal at "bind shape"), shape and size are each independently members of `AestheticChannels.FIELD_CHANNELS`, so the refusal for each points the caller at binding that channel's own field.
- Diagnosis (independently reproduced the crash via a throwaway render-path test) and adversarial refutation both survive in `docs/teams/2026-09-15-bug-76660-vca-shape-size-guard/bug-76660/` on the companion `stylebi-wiz` repo.

## Test plan
- [x] `./mvnw.cmd -pl core -Dtest=ChartAestheticMutatorTest test` — 112 tests, 0 failures, 0 errors
- [x] `./mvnw.cmd -pl core -Dtest='inetsoft.web.wiz.binding.*Test' test` — 667 tests, 0 failures; 11 pre-existing errors in `ChartBindingMutatorTest`/`SecondaryYPostProcessingPipelineTest`/`CalcTableServiceTest` (`NoClassDefFoundError` on `AggregateFormula`, a Spring-context teardown/test-order issue when the whole package runs together). Deterministic across two runs with this diff present, and `ChartBindingMutatorTest` alone passes in isolation. The diff here touches only `ChartAestheticMutator.java`/`ChartAestheticMutatorTest.java`, neither of which has any path to `AggregateFormula` or Spring context lifecycle, so this reads as pre-existing package-sweep test-isolation flakiness rather than something this change introduced — flagging for visibility rather than asserting it's confirmed pre-existing on `stylebi-wiz-main-rebase` without this diff, which I did not separately verify.


fredaan:
## Review round 1 (independent reviewer pass)

Head commit reviewed: 3162679b8.

**Bug-76660-specific check (diff vs. confirmed root cause):** independently re-verified against
01-diagnosis.md/02-refute.md, not just re-stated:
- Guard is general (dispatches on channel + frame type), not keyed on any repro literal.
- No special-casing that excludes only the reported input.
- Guard placement covers both the single-measure and broadcast-to-all branches.
- Re-grepped `createCombinedFrame`: confirmed `VSFrameVisitor.java:237` is the single call site
  shared by all five per-measure strategies.
- Re-grepped every `(StaticShapeFrame)`/`(StaticSizeFrame)` cast in `core/src/main/java`: no other
  unguarded cast reachable from this write path; `BarVO.java:1072` is confirmed
  `instanceof`-guarded already.
- Re-read `resetFrame`/`resetted()` myself: cannot manufacture the crash state from a static
  frame (no static-to-categorical case exists); confirmed out of scope for this PR, as the
  diagnosis claimed.
- Re-verified the wording-asymmetry claim against `AestheticChannels.FIELD_CHANNELS` as it
  actually reads today (`color, shape, size, text`) — matches the per-channel refusal wording.
- Re-checked test validity: `VisualFrameAliases.create` for a categorical size spec does not
  itself throw before reaching the new guard, so the size regression tests genuinely exercise the
  new guard rather than failing upstream of it for an unrelated reason.

No divergence found between the diff and the confirmed root cause.

**Generic pass:** scope stays to the two files the diagnosis/fix scoped to; no dead code, no
unrelated refactors; regression coverage (6 new + 1 replaced test) matches the diagnosis's
recommended list; no security-relevant surface.

**Build verification (self-run against the PR branch, not just the fixer's report):**
`./mvnw.cmd -pl core -Dtest=ChartAestheticMutatorTest test` → Tests run: 112, Failures: 0,
Errors: 0.

**CI status at review time:** `build` (Validation Build) still pending; `claude-review` shows
`fail`, but its own PR comment is a bot infra error ("Claude encountered an error after 0s"), not
a content finding — not treated as a blocking code-quality issue, but still a non-green check.
Semgrep (both) and license/cla pass. Mergeable, no conflicts.

**Findings:** none (blocker or important).

**Verdict:** No blocking code-review findings — the fix is independently confirmed to correct the
actual write-side guard gap, not a symptom-only patch, and my own build run matches the fixer's
report. Not marking this approved this round because (1) the `build` check is still pending on
GitHub and `claude-review` shows a `fail`, and (2) this is a self-authored PR (I am both author
and reviewer here) so I do not have merge/approve rights regardless — handing back to the lead
for the actual approve/merge action.

Full review: `docs/teams/2026-09-15-bug-76660-vca-shape-size-guard/bug-76660/05-review-r1.md` in
`stylebi-wiz`.

PR #5252, opened 2026-09-15
Bug #76666, give VSRefreshController.refreshViewsheet an explicit-id path for session-less callers

Nelson-cao:
## Problem

`ModifyCalculateFieldService`'s post-write refresh (fired on essentially every
`add_calc_field`/`edit_calc_field`/`remove_calc_field` call) asks
`VSRefreshController.refreshViewsheet(event, principal, dispatcher, linkUri)` to refresh the
viewsheet. That method derives its runtime id from `RuntimeViewsheetRef`, a
WebSocket-session-scoped bean populated only from a native STOMP header on a real browser
session. A session-less caller -- the wiz agent's HTTP-driven calc field mutations -- never has
that session, so `getRuntimeId()` returns null. The null reaches
`RuntimeSheetCache.getAffinityKey` unguarded and throws
`NullPointerException("Ouch! Argument cannot be null: key")`. A prior fix (VBS-003) wrapped this
in a try/catch, so the calc field write still commits, but the NPE fires on every call and is
only visible as a swallowed WARN in the server log -- the same underlying null id that #4971
(PCB-006) and #5186 (#76615) already fixed at two other call sites of the same
`RuntimeViewsheetRef` invariant.

## Fix

`VSRefreshController` gets an explicit-runtime-id overload of `refreshViewsheet`, mirroring the
sibling methods `refreshVsAssembly`/`refreshVsAssemblyView` that already take their id explicitly
from the event payload instead of `RuntimeViewsheetRef`. The STOMP-mapped entry point now
delegates to it using the session-derived id, for the native browser flow.
`ModifyCalculateFieldService` calls the new explicit-id overload with the id it already holds as
its own `@ClusterProxyKey` parameter -- the same shape as PR #4971's fix a few lines earlier in
the same method.


Nelson-cao:
Reviewed over two rounds and independently verified. Round 1 raised two important findings, both applied in `d704a78f0` and `01cea305b`; round 2 confirmed them fixed with no blockers.

Verification was re-run independently rather than accepted from the fix: 7/7 unit tests, 92/92 across seven regression classes, `mvnw -pl core clean compile` clean, and CI `build` green. The null guard's own test was checked for vacuity by removing the guard and confirming the test then fails as "expected NullPointerException, nothing was thrown" — the correct failure mode, so the test pins the guard rather than some downstream behaviour standing in for it.

Two things recorded rather than waved past:

- This fix narrows the shared invariant, it does not close it. Five further latent instances of the same shape were found during review — all confirmed unreachable today — and are filed as Redmine #76674 rather than left implicit.
- The red `claude-review` check is red across the repository (empty `ANTHROPIC_API_KEY` in the workflow env; same failure on #5251, #5249, #5248). It is a CI configuration fault, not a finding against this change.

The defect itself was reproduced live today against both a Logical-Model-direct viewsheet and a plain physical table, on a freshly built server, confirming it is still present on current main and is not tied to the binding shape the original report suggested. The fix is verified structurally; it has not been run on a live server.

Nelson-cao:
Housekeeping: the automated review posted at 06:12 UTC describes a version of this branch that no longer exists.

That review covers 9 changed files, including a `requireNonNull` guard in `RuntimeSheetCache.getAffinityKey` and rewired call sites in `TouchAssetService` and `VSChartRefreshService`. Those commits were reverted — the branch was reset to `01cea305b` and force-pushed — so none of that is in this PR. The 06:17 UTC review is the one that describes what is actually here: 5 files, `VSRefreshController` gaining an explicit-id overload with the null guard, and `ModifyCalculateFieldService` plus `ComposerViewsheetService.checkMV` using it.

The revert was a deliberate scoping decision, not a response to a defect in that work. Its analysis is preserved and the wider direction it pointed at — a guard at the shared affinity-key funnel, and the service-calls-controller layering inversion that generates this defect class — is recorded on Redmine #76674 rather than lost.

Worth noting alongside the passing `claude-review` check: that reviewer states it could not run the build or tests in its sandbox and relied on static call-graph tracing, citing this PR's own comment for the test numbers. It is not a substitute for the independent verification reported earlier; the CI `build` check passing on `01cea305b` is.

fredaan (approved):

PR #5254, opened 2026-09-15
Bug #76139: add wiz bridge endpoint for datasource folder rename

Zxm-happy:
## Summary
- The wiz portal has no way to rename a datasource folder. The native rename capability already exists (`DataSourceController.renameDatasourceFolder` -> `DataSourceBrowserService.renameFolder`), but no wiz-facing bridge endpoint exists on `WizDatabaseController` to reach it — this is the missing piece behind the wiz-services/portal PR for the same bug (companion PR TBD, `inetsoft-technology/stylebi-wiz`, branch `fix/76139-datasource-folder-edit`).
- Adds `POST /api/wiz/datasources/folders/rename` (`WizFolderRenameRequest{path, name}` -> `WizFolderSaveResult`), mirroring `createFolder`'s shape: validates the new name is a single path segment, checks WRITE on the folder, checks for a sibling name collision (skipped when the name is unchanged, since that's a no-op rename), then delegates to `DataSourceBrowserService.renameFolder`.
- **Finding that changed the implementation from the initial proposal**: the permission gate is this endpoint's own explicit responsibility, not something it can rely on `DataSourceBrowserService.renameFolder` to enforce. That method's own internal `securityEngine.checkPermission(...)` call at line ~404 never inspects the boolean it returns (dead code), and the `XEngine.updateDataSourceFolder` call beneath it has its own `checkPermission(ResourceType, String, ResourceAction)` hardcoded to `return true; // NO-OP, just for remote`. So nothing downstream of this endpoint enforces anything on a rename today — the new endpoint's own `requireFolderPermission(path, WRITE, principal)` check is the only real gate, matching what the native controller's `@Secured(DATA_SOURCE_FOLDER, WRITE)` annotation already enforces in practice (the deeper `ADMIN` check some earlier analysis assumed was live turns out to be inert).

## Test plan
- [x] `WizDatabaseControllerSecurityTest` — 8 new mock-based tests: success, permission denial, duplicate-name rejection, no-op-rename skips the duplicate check, name-format validation, plus the existing `mappingsMatchTheFrozenContract` test updated for the new route. All 28 tests in this class pass.
- [x] `WizDatabaseControllerRealRegistryTest` — left unchanged. Attempted to add a real-registry regression test proving a nested datasource survives a folder rename (the diagnosis's own falsifiable claim), but `DataSourceBrowserService.renameFolder`'s underlying `XEngine.updateDataSourceFolder`/`cutDataSourceFolder` call chain reaches several *different* static Spring-singleton lookups (`XRepository.getRepository()`, then `DataSourceRegistry.getRegistry()`, cascading further into `AdditionalConnectionDataSource`/metadata-cache cleanup) that this test harness does not wire up, and registering enough delegate beans to satisfy them broke Spring's eager singleton bean lifecycle for the whole test class. Reverted that addition rather than ship a broken/flaky test; a full real-registry rename test is still a gap and needs either a heavier Spring-Boot-test-style context or a live-server manual check.
- [x] `mvn -pl core -Dtest=WizDatabaseControllerSecurityTest,WizDatabaseControllerRealRegistryTest test` -> 32/32 pass.
- [x] `mvn -pl core -Dtest=inetsoft.web.wiz.**` -> all `WizDatabaseController*` test classes (46 tests) pass; the batch also surfaced ~230 failures in unrelated `Worksheet*` test classes (`NoClassDefFoundError` on `ExpressionRef`/`AggregateFormula`/etc.), which reproduce independent of this change and look like pre-existing classloading fragility in that large fork, not something this PR touches.


Zxm-happy:
Round 1 review (bug 76139 / AA-006). SHA reviewed: eaf619879af91af163a7f0f823a85f4396f0c7a8 (full history, first round).

Full writeup: docs/teams/2026-09-15-bugs-76139-datasource-folder-edit/bug-76139/05-review-r1.md in stylebi-wiz.

**Code review: clean.** Diagnosis-to-code-reach confirmed (the new renameFolder bridge endpoint calls the existing DataSourceBrowserService.renameFolder/DependencyTransformer rename mechanism, not a from-scratch reimplementation). Independently re-verified the permission-gate deviation by reading DataSourceBrowserService.renameFolder and XEngine.checkPermission/updateDataSourceFolder myself: the service's internal ADMIN checkPermission call at line 404-405 discards its return value (confirmed dead code), and the deeper XEngine.checkPermission(ResourceType,String,ResourceAction) at line 593 is a hardcoded NO-OP returning true (only implementation of that signature in the tree). WRITE is the correct bar for the new bridge endpoint's own requireFolderPermission check, matching what the native REST controller's @Secured(WRITE) already enforces in production, and matching the sibling deleteDatasources endpoint's own established pattern of gating explicitly at the bridge layer. No counter-examples found walking boundary/repetition/sibling-branch scenarios (see full writeup).

Prompt-content gate: clean, diff touches only 3 files under core/.../web/wiz, no prompt-facing code.

**Merge-readiness: blocked, not on code.** CI's `build` check is currently failing. I reproduced it myself directly on this PR's branch (not trusted from CI alone): `inetsoft.uql.ConditionTest$RegressionTests#toSqlConditionSilentlyDefaultsOnAnUnmatchedName` fails deterministically with a java.sql.Date equality mismatch. Confirmed unrelated to this diff — the failing test lives in a completely different package (inetsoft.uql date/SQL-condition handling) that this PR's diff does not touch at all, and the test's own code comment marks it a pre-existing "permanent characterization" tied to a different bug (PC-006 follow-on). Per this repo's mandatory build-verification rule, a red build blocks merge regardless of relatedness, so this needs to be resolved (either fixed/quarantined separately, or this PR rebased once resolved) before merge, even though the code in this diff itself is sound.

Requesting changes on that basis only — no code changes needed in this PR itself.

PR #5256, opened 2026-09-15
Fix #76671: guard IdentityThemeService.updateTheme against null oldId

fredaan:
## Summary

Community-side half of the fix for Redmine bug #76671 (organization theme handling). See the paired stylebi-enterprise PR for the primary fix (createOrganization/updateOrganization theme persistence).

`IdentityThemeService.updateTheme(oldId, id, fn)` used `Tool.equals(theme.getOrgID(), oldId)` to detect an org-rename and migrate a theme's `orgID`. Since `Tool.equals(null, null) == true`, a create-time call (`oldId == null`) matched every global (`orgID == null`) `CustomTheme` in the system, silently re-scoping it to the new identity's id via `setOrgID(id)` and then throwing an NPE on `getJarPath().replace(oldId, id)` (since `oldId` was null).

`oldId == null` means "this is a create, not a rename" and must never be treated as matching a global theme's null `orgID`. This was reachable via `SecurityApiService.createOrganization`'s broken `updateTheme(null, name, CustomTheme::getUsers)` call (fixed in the paired enterprise PR to use `assignTheme` instead), but this PR closes the underlying landmine in the shared utility directly so no future caller can trigger it the same way.

## Test plan

- [x] `IdentityThemeServiceTest.updateTheme_oldIdNull_doesNotMutateGlobalThemeOrgId` added — confirmed it fails against pre-fix code (NPE) and passes after the fix.
- [x] `./mvnw -q -pl core -am -Dtest=IdentityThemeServiceTest -Dsurefire.failIfNoSpecifiedTests=false test` — 7/7 passing (6 baseline + 1 new).


fredaan:
## Review round 1 (adversarial pass, informed by diagnosis/refutation) — CHANGES REQUESTED

Head commit reviewed: `260682a2`.

### Findings 2a/2b in this PR: mechanism check passed

Verified `updateTheme`'s `oldId != null && ...` guard against the diagnosis's/refutation's
stated mechanism (`Tool.equals(null, null) == true` previously let a create-time call with
`oldId == null` silently re-scope any global theme's `orgID` and then NPE on
`getJarPath().replace(oldId, id)`). The guard correctly closes this. New test
(`updateTheme_oldIdNull_doesNotMutateGlobalThemeOrgId`) exercises it directly. No objection to
this file's change in isolation.

### Blocker finding, cross-cutting with sibling enterprise PR #826

This PR's fix is correct on its own terms, but the paired enterprise PR (#826) routes
`createOrganization`/`updateOrganization` theme assignment through `IdentityThemeService.
assignTheme()`/this file's `updateTheme()` — helpers built for user/group/role identities, which
have no per-identity "selected theme" pointer. Organizations do have one
(`CustomThemesManager.setOrgSelectedTheme`/`getOrgSelectedTheme`, read by
`CustomThemesImpl.getUserTheme`'s org-level fallback), and it's never set by either
`assignTheme` or `updateTheme`. Full detail and evidence trail in `05-review-r1.md` (findings 1
and 3) — this comment is the required per-PR record; the actionable fix likely needs a change on
the enterprise side (or a new org-aware variant here) rather than in this file. Flagging here
too since this file's helpers are the shared mechanism both organization call sites route
through.

### Build/CI

`build` check on this PR is `pending` as of this review (checked twice, several minutes apart).
Per this repo's own StyleBI-side merge gate, this must resolve to green before merge regardless
of review outcome.

### Note on approval

This session authored this PR end-to-end (debugger → refuter → fixer → this review all ran under
the same account), so even a clean round could not be formally approved by this session — that
constraint doesn't change the verdict here since real blocking findings exist regardless.

Full findings: `docs/teams/2026-09-15-bugs-76671/bug-76671/05-review-r1.md` in stylebi-wiz.

fredaan:
## Review round 2 — code clean, build gate still not satisfied

Head commit reviewed: `e0f5c5ea`.

### Central claim re-traced independently: confirmed correct

Round 2's fix (06-fix-r2.md) reconciled round 1's own internal contradiction (Finding 1 section
said the runtime consumer checks by name; Finding 3 section said the real write path uses id) by
tracing `CustomThemesImpl.getUserTheme`'s `orgName` variable back to its actual assignment. I
re-traced this myself from source rather than trusting the fix doc's citation: `provider.
getOrganizationIDs()`'s own interface javadoc (`AuthenticationProvider.java`) says "Get a list of
all organization ids in the system," and every call site across the codebase names its loop
variable accordingly; `principal.getOrgId()`'s own javadoc in `XPrincipal.java` says "the
organization id." Both operands of the `orgID.equals(org)` comparison that populates `orgName`
are independently confirmed id-valued — the fix's reconciliation holds up. Full trace in
`08-review-r2.md`.

### This file's change (`getTheme(IdentityID, fn)` → thin wrapper + new `getTheme(String, fn)`)

Correct and minimal. Confirmed every other call site of `getTheme(...)` in both repos
(user/group/role, in `SecurityApiService`, `RoleController`, `UserTreeService`) still uses the
`IdentityID` overload unchanged — those identity types are legitimately name-keyed throughout
(their `assignTheme` writes are name-keyed too), so this fix correctly scopes the id-keyed
change to organizations only rather than over-generalizing. No objection.

### Not a code finding, but the round's overall verdict is still blocked

Zero code-correctness findings against this PR's diff. The overall round-2 verdict across both
PRs is still not "approved" — the StyleBI build gate isn't satisfied yet: `build` on this PR
remains `pending` (third consecutive observation across both review rounds), long enough now
that it looks more like a stuck check than one still legitimately queued. Per this repo's own
build-gate rule, that's treated as a blocker independent of code review outcome. Full detail:
`docs/teams/2026-09-15-bugs-76671/bug-76671/08-review-r2.md` in stylebi-wiz.

### Note on approval

Same as round 1: this session authored this PR end-to-end, so even a fully clean/green round
could not be formally approved by this session. Not attempting `gh pr review --approve`/
`gh pr merge`.

Nelson-cao (approved):

PR #5257, opened 2026-09-15
Fix #76676: mergeRowGroup/mergeColGroup ignored in calc-table SUMMARY aggregation scope

lizhepu:
## Summary

- `LayoutTool.createCalcBindingExpression` built a SUMMARY calc-table cell's aggregate qualifier chain purely from the cell's native `rowGroup`/`colGroup` ancestor chain (via `buildGroupList`). `mergeRowGroup`/`mergeColGroup` were faithfully stored and round-tripped, but never consulted anywhere in the value-computation path -- the only non-passthrough reader in the codebase was `RuntimeCalcTableLens.postProcess()`, which governs visual cell-merge span, not the computed aggregate value.
- This meant a SUMMARY cell placed inside a row that already has its own nearest native group cursor active could not "escape" to a named coarser ancestor group via `mergeRowGroup` -- it silently fell back to its own row's finest scope instead. Affects both #76676 (row axis, VCT-012) and sibling #76675 (column axis, VCT-013) via the same shared code path.
- Fix: truncate the shared `groups`/`gnames` ancestor-chain arrays at the named `mergeRowGroup`/`mergeColGroup` cell before either the `crossTabSupported` branch or the plain-aggregate branch consumes them, so both plain and crosstab-backed calc cells are covered. If the named group exists elsewhere in the layout but isn't actually an ancestor of the cell, fail loud with a `MessageException` naming the cell/property/value instead of silently using the native chain.

## Test plan

- [x] `LayoutToolTest` (4 new tests): row-axis truncation, column-axis truncation (#76675), the crosstab-supported branch, and the fail-loud non-ancestor case.
- [x] `./mvnw -pl core -am test -Dtest=LayoutToolTest,RuntimeCalcTableLensTest,CalcTableLensTest,CalcGroupSelectorTest,CalcRefTest,CalcTableScopeTest` -- 110 tests, all pass, no regressions in adjacent calc-table/formula machinery.

Full root-cause writeup and diagnosis/refutation trail: `docs/teams/2026-09-15-bug-76676/` in the stylebi-wiz repo (00-charter.md, bug-76676/01-diagnosis.md, 02-refute.md, 03-fix.md).


lizhepu:
Closing without merging — this PR's diagnosis was wrong, not just its implementation.

`mergeRowGroup`/`mergeColGroup` are a real, separate StyleBI feature (visual "merge expanded cells"
span control, gated entirely behind the `mergeCells` checkbox in the Composer UI — confirmed by
reading `calc-option-pane.component.html`, `[disable]="!cellBinding.mergeCells"`). They were never
meant to control aggregation scope. The field that actually does that is `rowGroup`/`colGroup`,
consumed by `LayoutTool.applyDefaultGroups()`/`CalcTableLens` — completely separate from anything
this PR touched.

Sibling bug #76675 (column-axis version of the same reported symptom) went through the identical
wrong turn — a diagnosis and one refutation round both concluded `mergeColGroup` should drive
scope — before a human's own live Composer UI debugging caught that the real, working mechanism is
`colGroup` set directly, via a different dialog panel that happens to reuse the same
`_#(Row Group)`/`_#(Column Group)` UI label. #76675's actual fix (already merged: #5261 here,
#2544 in stylebi-wiz) exposes real `rowGroup`/`colGroup` parameters on the plugin's
`set_cell_binding` tool — the plugin never exposed the fields that already, correctly, control
scope; no StyleBI core-engine change was needed.

This PR instead added a new core-engine mechanism (`LayoutTool.truncateAtMergeGroup`) making
`mergeRowGroup`/`mergeColGroup` also affect aggregation scope. That's not a missing capability —
it's new, incorrect semantics for a field pair that was never supposed to touch computed values,
and it would have shipped alongside (and in conflict with) the correct `rowGroup`/`colGroup`
mechanism #76675 just finished exposing.

Full writeup: `docs/teams/2026-09-15-bug-76676/bug-76676/09-reassessment.md` in stylebi-wiz.
Follow-up: verify whether #76675's merged fix already lets a caller set `rowGroup` on a
`grouping:"summary"` cell (not just `grouping:"group"`) — if so #76676 is already resolved as a
side effect and needs no separate change; if SUMMARY cells are excluded, that's the actual
remaining gap, and it belongs in the plugin's tool schema/validation, not here.

PR #5259, opened 2026-09-15
Bug #76672 (mv): a per-item throw before mutating was misreported rollback-failed

Nelson-cao:
## Summary

`MvChangesetApplyService.apply()`'s per-item `catch(Throwable)` unconditionally added every throwing entry to `unknownStateFailures`, forcing `STATUS_ROLLBACK_FAILED` even when the throwing entry's own mutating call (`createMV`/`setDataCycle`) was never invoked -- e.g. when the analysis-freshness re-check `applyCreate`/`applySetCycle` each perform throws because the cached analysis was evicted between the plan's own `resolve()` and this entry's turn in the loop, while an earlier entry's rollback completed cleanly and verifiably.

- Threaded a per-iteration `AtomicBoolean mutationEntered` into `applyCreate`/`applySetCycle`, set only immediately before each method's own mutating call(s). The loop's `catch(Throwable)` now only adds to `unknownStateFailures` when the flag is true.
- `applySetCycle` additionally gained an explicit `analysisResult.getStatus()` call as its own first statement (mirroring `applyCreate`'s existing one): the live analysis-freshness lookup for this verb previously lived one call deeper, inside `MVSupportService.setDataCycle` itself, past the point `mutationEntered` would already be `true` under a naive flag placement.
- `applyDelete` is unchanged -- its only statement (`dispose`) is the mutation itself, with no separate pre-mutation check to gate on.
- `catch(Throwable e)` itself is unchanged (required: `AdminMvGateway.createMV`/`MVSupportService.createMV` declare `throws Throwable`).

## Test plan

- [x] `MvChangesetApplyServiceTest` (new): pre-mutation throw in `applyCreate` -> `STATUS_ROLLED_BACK`, no rollback failures
- [x] `MvChangesetApplyServiceTest` (new): pre-mutation throw in `applySetCycle` (the round-2-revised path) -> `STATUS_ROLLED_BACK`
- [x] `MvChangesetApplyServiceTest` (new): a throw after the mutating call has started (`createMV`) still forces `STATUS_ROLLBACK_FAILED`
- [x] `../mvnw.cmd -o -Dtest=MvChangesetApplyServiceTest test` from `core/` -- 3/3 passing


Nelson-cao:
Found and confirmed valid — thanks for catching this. `applyDelete` never threaded `mutationEntered`, so a throw from `mvGateway.dispose(...)` was silently excluded from `unknownStateFailures`, a regression versus the pre-fix conservative behavior for this irreversible verb. Fixing now in a follow-up commit on this branch.

Nelson-cao:
Follow-up pushed: commit 55f8de1da threads mutationEntered into applyDelete, setting it true immediately before mvGateway.dispose(...) (the method's first statement and the mutation itself), and updates the call site to pass it through. Added a regression test (deleteThrowStillForcesRollbackFailedEvenThoughRollbackIsClean) that asserts STATUS_ROLLBACK_FAILED when dispose() throws alongside a create entry that rolls back cleanly. All 4 tests in MvChangesetApplyServiceTest pass.

fredaan (approved):

Nelson-cao:
## CI build failure analysis (two runs)

Investigated both `build` check failures on this PR before the base-branch merge. Neither is caused by this PR's diff.

**Run 1** (`OrganizationContextHolderThreadingTest.twoThreads_haveIsolatedState`): the failure came from a background `ThreadPool$WorkerThread`'s `cleanUp()` calling `LicenseManager.getInstance()` while the Spring context was tearing down (`ShutdownException`), not from the test's own assertions — a classic async-background-thread-vs-context-teardown race, unrelated to `OrganizationContextHolder` (a plain `ThreadLocal` wrapper) or anything this PR touches.

**Run 2** (`OrgLifecycleRepletRegistryIntegrationTest.delete_updateRepletRegistry_removalIsInMemoryOnly_notPersistedToDisk`): this test's own code comment (on `detachFileWatch()`) documents it as a known, load-timing-sensitive intermittent failure: *"the event for a fixture's own save() can land at an arbitrary later point... intermittently, and only when the JVM is busy enough to delay delivery, i.e. when the whole security package runs in one surefire JVM."* This is a pre-existing, self-acknowledged flaky area, predating this PR by ~2 months (added 2026-07-24, commit `8ab67c912d`).

**Checked for an indirect relationship anyway**, since both failures are timing/load-sensitive and this PR adds new tests to the same `core` module/surefire fork: all three new/changed test files in this bug-family (`StoredAssetChangesetApplyServiceTest`, `RecycleBinChangesetApplyServiceTest`, `MvChangesetApplyServiceTest`) use `@ExtendWith(MockitoExtension.class)` only — no Spring context (`@ContextConfiguration`/`SpringExtension`), and `RecycleBinChangesetApplyServiceTest` uses `mockStatic(RecycleUtils.class)` for the one call site that would otherwise touch a real `RepletRegistry`. None of the three touch `DataSpace` file-watching, `RepletRegistry`, `ThreadPool`, or `LicenseManager` in any real (non-mocked) way. The only plausible indirect link is generic test-suite load (more test methods in the same JVM fork shifting timing) — the same marginal effect any PR adding a handful of fast unit tests would have, not a functional relationship to this diff's logic.

Merged the base branch in and pushed (`ab0a04099`) to get a fresh CI run; watching for the result.

Nelson-cao:
## Fixed one of the two flaky-test CI failures directly; explaining why the other one is not a safe fix

Per the request to unblock merge, I looked at actually fixing both root causes, not just re-running CI.

### Fixed: `ThreadPool.cleanUp()` swallowing `ShutdownException` (commit `e829494a3`)

Root cause: a background worker thread's `cleanUp()` (runs on its own 30s timer, independent of the Spring context lifecycle) calls `LicenseManager.getInstance()`, which throws `ShutdownException` once the Spring context has torn down. This is a genuine, if narrow, shutdown-ordering gap — the same class of race the codebase already handles elsewhere with `catch(ShutdownException ignore)` (`GroupedThread`, `XSwapper`, `LogbackInitializer`). Applied the identical pattern here. Verified it compiles and `OrganizationContextHolderThreadingTest` passes.

### Investigated but deliberately NOT fixing: `OrgLifecycleRepletRegistryIntegrationTest`

The "obvious" fix — make `IdentityService.updateRepletRegistry()` call `registry.save()` so the folder removal persists — would be **wrong**, not just cosmetic. Traced the real production call order for org deletion (`IdentityService.java` around line 613-614):

```
removeOrgScopedDataSpaceElements(oOrg);   // physically deletes the org's DataSpace files, INCLUDING repository.xml
updateRepletRegistry(orgID, null);         // removes folders in-memory only
```

By the time `updateRepletRegistry()` runs, `repository.xml` is **already deleted**. Adding `save()` there would **resurrect** the just-deleted file with a folder-stripped registry — a real regression, not a fix. The test's own comment already says this precisely: the in-memory-only removal "only appears correct in production because the DataSpace-level path deletion... already removed... the physical file a step earlier in the real call chain." That's not a bug description, it's an accurate architecture note.

Ran the test in isolation 3x — 100% pass every time. It only fails "when the whole security package runs in one surefire JVM" (the test's own words), i.e. it's a load/timing-dependent async race in the shared `BlobStorage` single-thread event executor across the whole test package, not something reproducible or safely fixable by touching this one test or this one production method. A real fix would mean exposing a synchronous drain/await hook on `BlobStorage`'s event executor — a change to widely-used core infrastructure, out of scope and risk profile for this bug-fix PR.

Leaving this one as a known, pre-existing, load-dependent flaky test (predates this PR by ~2 months). If it fires again on a re-run, that's expected and unrelated to this diff.

Nelson-cao:
## build now passes ✅

After the base-branch merge (\`ab0a04099\`) and the \`ThreadPool.cleanUp()\` fix (\`e829494a3\`), the latest \`build\` check passed (21m15s): https://github.com/inetsoft-technology/stylebi/actions/runs/34949933691

Both previously-failing tests ran clean in this run:
\`\`\`
Tests run: 10, Failures: 0, Errors: 0 -- OrganizationContextHolderThreadingTest
Tests run: 3, Failures: 0, Errors: 0 -- OrgLifecycleRepletRegistryIntegrationTest
\`\`\`

Also confirmed locally: ran the full \`core\` module test suite (10044 tests) on this branch — \`BUILD SUCCESS\`, 0 failures, 0 errors, both tests above pass there too.

\`claude-review\` still shows \`fail\` — that's the same infra-level Action execution error (\`is_error:true\`) from before, unrelated to the diff or code content; not a code-review finding to address.

This PR should be ready to merge (pending human review sign-off).

fredaan (approved):

PR #5260, opened 2026-09-15
fix(wiz): warn on runQuery('ws:<token>:<path>') scope mismatch (bug #76661)

NeverStopFMH:
## Summary
- `runQuery('ws:<token>:<path>')` addressing is scope-sensitive by design (`AssetEntry.createAssetEntry`): a literal `"global"` token resolves `GLOBAL_SCOPE`, anything else (e.g. a username) resolves that user's `USER_SCOPE`. Getting the token wrong resolves to an `AssetEntry` that legitimately has nothing at it, and the failure was completely silent — `ReportWorksheetProcessor.execute()` only logs a WARN and returns `null`; `XUtil.runQuery()` never null-checks that and wraps it into an empty `XTableArray` with no exception anywhere.
- Fixing the shared runtime directly would also change behavior for real Viewer/Composer sessions and scheduled/exported classic Reports unrelated to this bug (`ReportJavaScriptEngine` runs the same `runQuery` with `reportScript=true`, which rethrows). Instead, this adds a static pre-flight check confined to `ScriptExecuteService.execute()` — the service backing only the wiz plugin's own `/api/wiz/v1/agent/script/*` endpoints, gated behind `SheetAgentFeature`, never reached by a real end-user session.
- When a literal `runQuery('ws:...')` call addresses an empty scope but the same-named worksheet exists at the other scope (checked via the real `AssetRepository.containsEntry()`), the result's `summary` gains an explanation plus the corrected `ws:` literal. Genuinely-missing-everywhere and correctly-scoped calls are untouched; `value`/`ok` are never altered.

## Test plan
- [x] 3 new unit tests in `ScriptExecuteServiceTest` (hit, correctly-scoped control, missing-everywhere control) — 17/17 passing
- [x] Live end-to-end via the wiz plugin's own `run_script_live` against a real viewsheet: wrong scope shows the note (value unchanged, still `0`); corrected `ws:global:<name>` returns the real row count with a clean summary; no UI-visible difference in either case (`get_viewsheet_image` confirmed)

Related: Redmine #76661, companion PR inetsoft-technology/stylebi-wiz#2542


NeverStopFMH:
Addressed the one cheap, worthwhile item: added a one-line javadoc note on `firstWsScopeMismatch` that it only reports the first mismatch found in a script (no behavior change).

Left the other three as-is, per your own framing of them as minor/non-blocking:
- Performance (up to 2 `containsEntry()` calls per literal `ws:` runQuery) -- you already judged this acceptable given the low-volume wiz-agent-only path.
- Quote-style re-wrapping (`'...'` regardless of original quoting) -- cosmetic, harmless.
- `getOrgId()`/`getOrgID()` nullability -- not a new risk, matches existing `AssetEntry.createAssetEntry` call sites elsewhere in the codebase.

NeverStopFMH:
Fixed both issues from the second review pass:

1. **Wording bug**: the GLOBAL→USER branch now says `Use "ws:<user>:<path>" to reach it.` instead of the misleading `Save the worksheet at user scope...` (the sibling already exists — confirmed via `containsEntry()` — so "save" was wrong). Added a 4th unit test (`runLiveAppendsWsScopeMismatchNoteForTheReverseDirectionWithCorrectFixWording`) asserting the exact message for this direction, since the prior 3 tests only ever exercised the other one.

2. **NPE gap**: switched to `XPrincipal.getIdentityID()` (resolved once, null-checked before use, instead of re-deriving from `getName()` at two separate unguarded call sites), and wrapped the whole method body in a top-level try/catch as a backstop — matches the "never throws" contract the javadoc already claimed.

Left the 4-segment `ws:token:path:table` note as a comment only (not a behavior fix), per your own "rare enough in practice" framing.

Verified: 18/18 unit tests passing (`ScriptExecuteServiceTest`), rebased onto the latest `stylebi-wiz-main-rebase`.

fredaan (approved):

PR #5263, opened 2026-09-15
Add provider update verb, preserving chain index (bug 76686)

Zxm-happy:
## Summary

Server-side (`community` submodule) half of wiz bug 76686. `preview_provider_changes`/`apply_provider_changes` only supported `create`/`delete`/`duplicate` for authentication/authorization provider chains — editing a single field (LDAP bind password, host, port) on an existing provider required delete+create, which always re-lands the provider at the **end** of the chain (`addAuthenticationProvider` has no insert-at-index form), silently changing provider resolution order for multi-provider chains.

`AuthenticationProviderService.editAuthenticationProvider`/`AuthorizationProviderService.editAuthorizationProvider` already exist, are index-preserving (`providerList.set(i, ...)`), and are EM's own real "Edit" action's backing calls — but had zero self-lockout preflight, and this area's plan/apply services never wired a path to them.

- **`ProviderChangeRequest`**: adds `VERB_UPDATE`.
- **`ProviderChangePlanService`**: new `resolveUpdate` dispatch for `chain="authentication"` only — `chain="authorization"` is refused outright, since the only authorization provider type this area creates (FILE) has no editable fields, and the spec DTO (`ProviderLdapSpec`) is LDAP-configuration-only. A non-LDAP current authentication provider (FILE) is likewise refused, for the same reason. `providerType`/`newName` are refused on update (an update cannot change type or rename in this cut).
  - `mergePartialLdapSpec`: overlays whatever fields the caller's partial spec sent onto the current provider's LDAP config, carrying every omitted field over unchanged.
  - `requireLdapCredentialCrossValidation`: cross-field `useCredential` validation against the *merged* model (not the raw partial spec) — mirrors `DataSourceChangePlanService.requireCredentialCrossValidation`'s precedent, refusing e.g. a `password`-only patch against a `useCredential=true` provider.
  - `requireAuthenticationEditPreflight`: the one genuinely new preflight — generalizes the existing delete preflight's simulate-and-check pattern (`providerHasSysAdmins`/`callerRetainsSysAdmin` against a simulated post-change chain) from "remove" to "replace at the same index with the proposed provider," refusing loud if the edit would strip either the deployment's or the calling session's own system-administrator access. No new authorization-chain preflight is needed (`AuthorizationChain.getPermission`'s resolution has no length-dependent branch, and edit never changes chain length).
- **`AuthenticationProviderService`**: new `buildProviderForPreflightSimulation` — a thin public passthrough to the existing (private) `getProviderFromModel`, so the preflight can construct a throwaway provider instance to evaluate `getRoles()`/`getRoleMembers()` against, torn down immediately after use, never added to the live chain.
- **`ProviderChangesetApplyService`**: new `applyUpdateAuthentication`, re-deriving the merge/cross-validation/preflight fresh at apply time (never trusting anything captured at preview) before calling `editAuthenticationProvider`; new `Undo.UPDATED_AUTHENTICATION` kind, rolled back via `editAuthenticationProvider(name, before, user)` — restoring the provider at its **original** chain index, a strict improvement over delete's own append-only rollback.
  - **Correctness fix found during implementation**: the captured "before" model (from `getAuthenticationProvider`) always carries `oldName(name)` set. If passed as-is to `editAuthenticationProvider` during rollback, `replacePlaceholderWithPassword` would see a non-null `oldName` and — since the provider being rolled back is *still live* in the chain at that point (unlike delete's rollback, where it's already gone) — would silently substitute the **current** (just-rotated) password into the "restored" model instead of leaving the original alone. Rollback now clears `oldName` before calling `editAuthenticationProvider`, converting this into the same well-understood "literal password fails loud, needs manual intervention" contract `rollbackDeletedAuthentication`'s own javadoc already documents, instead of a silent wrong-password substitution.

## Test plan

- [x] `./mvnw -pl core test -Dtest=ProviderChangePlanServiceTest,ProviderChangesetApplyServiceTest` — 76/76 pass (8 new resolve-level tests, 3 new apply-level tests)
- [x] `./mvnw -pl core test -Dtest=AuthenticationProviderServiceTest,AdminProviderControllerTest` — 23/23 pass (no regressions in adjacent/collateral areas)

This is the server-side half only. A companion wiz-side PR (`plugin/admin` — `providerTools.ts`/`normalize.ts` wiring the new verb through to the MCP tool surface) and a follow-up submodule-bump PR in `stylebi-enterprise` are still to come.


Zxm-happy:
## Review — bug 76686 provider `update` verb (community half)

*(Posted as a comment rather than a formal review — GitHub disallows a review with a decision state on one's own PR. Treat the "Important finding" below as **request-changes**-equivalent: please address/acknowledge before merge.)*

Reviewed against `docs/teams/2026-09-15-bug-76686-provider-update-verb/bug-76686/01-diagnosis.md` (incl. Revision round 1) and `02-refute.md` (incl. the round-1 recheck) — the confirmed, adversarially-checked design — plus a direct, independent read of every changed file.

### Correctness — confirmed against the diagnosis, no issues found

- `resolveUpdate`'s pipeline (merge → `requireLdapCredentialCrossValidation` against the *merged* model → `requireAuthenticationEditPreflight`) matches section 4 as amended by Revision round 1, in the right order.
- **The refusal of `chain="authorization"` (a deliberate deviation from the diagnosis's literal dispatch-generically text)** — verified independently by reading `AuthorizationProviderModel.java`: for a FILE-type authorization provider the model carries only `providerName`/`providerType`/a `customProviderModel` field that is populated for CUSTOM only. There is genuinely no editable field for FILE. This is correct, well-justified scope-narrowing, not an unjustified gap.
- **The rollback password-substitution fix** — traced `rollbackUpdatedAuthentication` and `replacePlaceholderWithPassword` by hand. Without clearing `oldName`, rollback would substitute the *current* (post-rotation) password into the "restored" model, because — unlike delete's rollback, where the provider is already gone — the provider being rolled back here is still live in the chain. Clearing `oldName` before the rollback's `editAuthenticationProvider` call is the correct fix.
- `requireAuthenticationEditPreflight` tears down the throwaway provider in a `finally` block (no leak), and runs `providerHasSysAdmins`/`callerRetainsSysAdmin` against the simulated post-edit list, not the live one.
- `applyUpdateAuthentication` re-derives everything (before-model, merge, cross-validation, preflight) fresh against live state — never trusts anything from preview, matching `applyCreateAuthentication`'s existing discipline. `applyOne`'s assumption that `chain="authorization"` can never reach it for `update` is safe, since `apply()` calls `planService.resolve()` (which refuses that combination) before the loop ever runs.

### Important finding — preview now performs a live LDAP bind, contradicting the diagnosis's own risk framing

`01-diagnosis.md` section 4a states the preflight's provider construction is "checkParameters() only, not a live bind... bounded and already-accepted risk, not new risk," equating it with what `create`/`duplicate` already accept. I read `LdapAuthenticationProvider.checkParameters()` (`core/src/main/java/inetsoft/sree/security/ldap/LdapAuthenticationProvider.java:509-562`) directly: it calls `loadCredential()`, `createContext()`, and `testContext(context)` — a genuine live LDAP connection/bind against the real server, with dedicated `CommunicationException`/`AuthenticationException`/`NamingException` handling. This is a live bind, not a local validation.

That matters because `resolveCreate` (create's own preview path) never calls `getProviderFromModel` at all — it only does field-presence validation (`requireLdapSpec`); the live bind for `create` happens only at **apply** time. The new `requireAuthenticationEditPreflight` calls `buildProviderForPreflightSimulation` → `getProviderFromModel` at **preview** time. So `update` is the only verb whose `preview_provider_changes` performs live external I/O against the LDAP server, using the proposed (possibly just-rotated) credentials:

- A transient LDAP outage/slow network will now fail or hang `preview_provider_changes` for an `update` entry even when the merge/cross-validation/self-lockout logic is entirely correct — a materially different failure mode than every other verb in this area.
- None of the new unit tests exercise this real behavior (`buildProviderForPreflightSimulation` is mocked away in every test), so it is untested in practice.
- This may well be a reasonable, even desirable, design choice (catching a bad rotated password/host before commit) — but it should be an explicit, documented decision, not something that slipped through mischaracterized as "no new risk." Please update the diagnosis's/PR's characterization and add a line to the `update` verb's docs (`providerTools.ts` prose, when that PR lands) disclosing that a preview of an LDAP `update` performs a live connection test against the proposed configuration.

Not a functional bug — the code does what it does correctly — but the risk framing was wrong and neither round of adversarial review (diagnosis or refute) caught it, so this needs to be acknowledged/documented before merge, not silently shipped.

### Build gate

Ran an independent, from-scratch clean build myself (not reusing the fixer's cached `target/`):
- `./mvnw -o -pl core -am clean test-compile` → **BUILD SUCCESS** (6962 main + 1098 test source files compiled cleanly, ~2m24s).
- `./mvnw -o -pl core clean test -Dtest=ProviderChangePlanServiceTest,ProviderChangesetApplyServiceTest` → **BUILD SUCCESS**, Tests run: 76, Failures: 0, Errors: 0.

CI's own `build` check was still `pending` after 17+ minutes at review time (unrelated `claude-review` check failure is an SDK/infra error in the bot itself — empty result, `is_error:true`, 0 turns of actual review — not a content finding). Given my own independent clean build+test run is at least as rigorous as CI's, I consider the build gate substantively satisfied, but the fixer/lead should still confirm CI's `build` check lands green before actual merge.

Zxm-happy:
## Round 2 recheck — Important finding closed

Confirmed commit \`f94687e10\` fully closes the round-1 "Important finding" (preview's live LDAP bind mischaracterized as no-new-risk):

- \`requireAuthenticationEditPreflight\`'s javadoc now states plainly that this preflight performs a real live LDAP bind at preview time via \`checkParameters()\`'s \`createContext()\`/\`testContext()\`, contrasts it correctly with \`create\` (which never calls \`getProviderFromModel\` at preview, only at apply), names the consequence (transient outage/slow network can fail or hang \`update\`'s own preview), and explicitly frames it as an accepted, intentional trade-off rather than a defect. Accurate and complete — matches what I verified independently in round 1.
- The addendum in \`03-fix-community.md\` (wiz-repo docs, not part of this PR's diff) correctly redirects future readers away from the stale \`01-diagnosis.md\` section 4a characterization.
- Doc-only change confirmed: no logic or test changes in this commit, consistent with the commit message.

Also confirmed the companion wiz-side disclosure (PR #2551, commit \`093230919\`, \`stylebi-wiz\` repo): \`providerTools.ts\`'s \`preview_provider_changes\` prose now tells the calling LLM about the live-bind behavior right after the self-lockout warning, correctly contrasts it with every other verb, and gives actionable guidance (attribute a preview failure to connectivity, not necessarily to the proposed change). Ran the scoped tests myself independently: \`providerNormalize.test.ts\` + \`providerTools.test.ts\` → **119/119 passing**, including the new disclosure-text assertion — matches the reported count exactly.

**Verdict: both docs changes fully close the round-1 finding. No further code, test, or doc changes needed on either side from my end.** Community PR's fresh CI \`build\` run (for this doc commit) was still \`pending\` at the time of this recheck — not blocking given the from-scratch clean build+test I ran independently in round 1 already covers this module's compile/test correctness, and this commit is documentation-only (javadoc comment changes cannot affect compilation). Recommend a final glance at the CI result once it lands purely as routine hygiene before the merge click.

PR #5264, opened 2026-09-15
fix(76691): skip Office lock files (~$*) during ServerFile catalog enumeration

fredaan:
## Summary

`ServerFileCatalog.collect`'s enumeration whitelist (`ServerFileUtil.isText`/`isExcel`) is
suffix-only, so a Microsoft Office lock file such as `~$Excel Data.xlsx` (created automatically
whenever the real workbook is open elsewhere) matches the `.xlsx` check and is handed to
`readSheetNames`, which throws because it isn't a real workbook. That throw is never caught in
`collect`'s per-child loop, so it aborts enumeration of the *entire* directory — every other file
in the folder (however healthy) gets dropped, and the datasource is reported `enumerable:false`.

- Add `ServerFileUtil.isOfficeLockFile` (`~$` filename-prefix check).
- Exclude such files from the enumeration whitelist in both `ServerFileCatalog.collect` and
  `ServerFileUtil.getFileList` (the class javadoc states these two file sets must stay in sync).
- A genuinely corrupt/unopenable workbook (no `~$` prefix) still aborts the whole catalog —
  the existing contract test `anUnopenableWorkbookThrows_ratherThanBeingSkippedFromTheCatalog`
  is left unmodified and still passes.
- New regression test `officeLockFileIsSkipped_otherFilesStillEnumerate`.

Redmine #76691 (WBT-008).

## Test plan
- [x] `./mvnw -pl connectors/inetsoft-serverfile -Dtest=ServerFileCatalogFailureTest test` — 5/5 passed (4 pre-existing + new lock-file test), C4 contract test unmodified and still passing.
- [x] `./mvnw -pl connectors/inetsoft-serverfile test` (full module) — 35/35 passed, BUILD SUCCESS.
- [ ] CI — not yet observed on this PR; please confirm it runs green before merging.


fredaan:
## Review (round 1) — WBT-008 / Redmine #76691

Read `01-diagnosis.md`/`02-refute.md`/`03-fix.md`/`04-verify.md` plus the diff and the live files at the PR head commit (`9b3954010`) directly, not just the archived write-ups.

**Correctness against the confirmed cause (not just the symptom):**
- `ServerFileUtil.isOfficeLockFile` (`getName().startsWith("~$")`) is a general filename-prefix check, not keyed to the specific repro literal `~$Excel Data.xlsx` — confirmed by reading the method body myself.
- It's a filename-prefix predicate, structurally unable to also match a legitimately-named corrupt workbook, so the C4 contract test (`anUnopenableWorkbookThrows_ratherThanBeingSkippedFromTheCatalog`) is left untouched and still enforces "a real unopenable workbook still aborts the whole catalog."
- Both enumeration call sites are patched (`ServerFileCatalog.collect` and `ServerFileUtil.getFileList`), matching the class javadoc's stated invariant that these two file sets must stay in sync. Confirmed the `getFileList` branch is load-bearing (placed before the `isExcel` check, so it actually changes the result — not a no-op).
- New regression test `officeLockFileIsSkipped_otherFilesStillEnumerate` sits right after the C4 test and asserts the general behavior (one healthy file enumerated, lock file silently skipped).

**Scoping decision I independently checked:** `ServerFileUtil.getFirstFile` is left untouched. I read it directly — it IS a directory walk (recurses via `listFiles()`, sorts by name, returns first match), so the refuter's correction of the diagnosis's "not a directory walk" claim is accurate. Since `~` sorts after every ASCII letter/digit, a `~$`-prefixed file is only ever `getFirstFile`'s pick when it's the *sole* candidate in a folder — a narrower, different-call-path scenario (feeds `getColumnDefinition`, not `list_tabular_targets`/`ServerFileCatalog`) than what WBT-008 reports. I agree this is reasonable to leave out of this PR. Non-blocking suggestion: file it as a known follow-up so it isn't lost.

No blocking findings.

**Merge gate (checked live just now):**
- `build` (Validation Build): **IN_PROGRESS** — not yet resolved.
- `claude-review`: **FAILURE**, but the only comment is "Claude encountered an error after 0s" with zero review content — reads as an infra/job crash, not a substantive finding, though still worth re-triggering/confirming before merge.
- Semgrep Scan / Semgrep OSS / license-cla: all green. Mergeable: MERGEABLE (no conflicts).

Per this repo's convention, I am not approving or merging this PR myself (my own team authored it end-to-end this run). Do not merge until `build` finishes green and `claude-review` is resolved or confirmed to be environmental noise.

jshobe-inetsoft (approved):

PR #5265, opened 2026-09-15
Fix #76690 WBT-004: customLookups jsonPath silently no-ops on scalar

fredaan:
## Summary
- `RestJsonQueryRunner.doLookups()` evaluated a `customLookups[i].jsonPath` against the parent row via Jayway JsonPath, but only handled a resolved `Map` or `Array` — a resolved scalar (e.g. `jsonPath:"$.id"`) fell through both branches with no `else`, no exception, and no log, leaving the row completely unmodified while `add_table` still reported `ok:true`.
- Adds a fail-loud `else` branch: when the transformed result is neither a map, an array, nor `null`, throw `IllegalStateException` naming the offending level and its `jsonPath`. `null` is left alone (already handled upstream by `PathNotFoundException`/`JsonTransformer`'s null-leaf behavior).
- Matches the pre-existing "not a map" exception convention already used by `LookupService` for the sibling `EndpointJsonQuery` lookup path (`IllegalStateException`), applied here to the `customLookups`/plain-`RestJsonQuery` path that never had an equivalent check.
- Diagnosed and refuted end-to-end in `docs/teams/2026-09-15-bug-76690-wbt-004-005-006/bug-wbt-004/` (in the `stylebi-wiz` repo) via a real JUnit test driving `doLookups()` with the actual production setters. Per the refuter's correction, implements only the "fail loud" fix direction (the alternative "forgiving scalar-as-id" option was shown not to fix the actual reported repro, since it only applies when `key` is blank).

## Test plan
- [x] Added `RestJsonQueryRunnerCustomLookupJsonPathTest` (`connectors/inetsoft-rest/src/test/java/inetsoft/uql/rest/`): asserts `jsonPath:"$.id", key:"id"` now throws a clear, field-named `IllegalStateException` and leaves the row unmutated; asserts `jsonPath:"$", key:"id"` (the already-working form) is unaffected and still adds the `lookup1` property.
- [x] `./mvnw -q -pl connectors/inetsoft-rest -Dtest=RestJsonQueryRunnerCustomLookupJsonPathTest test` — 2/2 passing.
- [x] Broader regression check: `./mvnw -q -pl connectors/inetsoft-rest -Dtest="*RestJson*,*Lookup*,ExpandedJsonTest" test` — 21 tests, 0 failures, 0 errors.


fredaan (commented):
## Review R1 — WBT-004 customLookups jsonPath fail-loud fix

**Verdict: clean — 0 blockers, 1 important (non-blocking) observation.**

**Does this key on the confirmed cause, not just the reported symptom?** Yes. Verified against
the live source (not just the fix description): the new `else if(transformedJson != null)` branch
in `RestJsonQueryRunner.doLookups()` sits immediately after the existing `isMap`/`isArray` checks —
the actual gate the diagnosis/refute identified — and fires for *any* scalar jsonPath result, not a
`$.id`-specific special case. Confirmed only fix option 2 ("fail loud") was implemented, matching
the refuter's correction that option 1 (forgiving, key-blank scoped) would not have fixed the
literal reported repro (`key:"id"`, non-blank) — no key-blank branch was added anywhere. Confirmed
the regression test drives the exact reported shape (`jsonPath:"$.id", key:"id"`) through the real
production setters, and confirmed the `EndpointJsonQuery` overload in `LookupService` (the stricter,
pre-existing check for the catalog/named-endpoint path) is genuinely a separate call path,
correctly left untouched and out of scope.

**Exception convention**: `IllegalStateException` matches the pre-existing "resolved value is not a
map" convention already used in `LookupService.java` (lines 37, 63) for the sibling path. Read
`runStream()`'s blanket `catch(Exception ex) { logException(ex); }` directly — this new exception is
handled exactly the way `LookupService`'s existing throw already is (logged, not surfaced as an
`add_table`-time rejection). Established behavior in this file, not invented by this PR.

**One important, non-blocking observation**: the "fail loud" here is loud only in the *server log*
(via `runStream()`'s catch), not anywhere in the `add_table`/`preview_worksheet_data` tool response.
The charter's stated symptom was explicitly "no error anywhere in the *response*" — this fix adds a
log entry but the response itself still won't tell an LLM-driven caller why the table came back
empty/truncated. Not a new regression (it mirrors the exact precedent this fix deliberately follows
from `LookupService`), so not blocking, but worth a follow-up thought on surfacing backend lookup
errors into the tool response itself.

**Verification I ran myself** (not just re-reading the fixer's claim): re-ran
`./mvnw -q -pl connectors/inetsoft-rest -Dtest="*RestJson*,*Lookup*,ExpandedJsonTest" test` in the
PR's own worktree at HEAD `769904207` — exit 0, surefire reports confirm 21/21 tests passing
(0 failures, 0 errors), matching the fixer's claimed result. The `build` CI check was still
in-progress at review time; the `claude-review` check shows FAILURE but its content is just an
infra error ("Claude encountered an error after 0s"), not a substantive finding.

No `wiz-services/src/agents/prompts/` or other prompt files touched — prompt-merge gate doesn't
apply. Code style/indentation matches the surrounding file; error message is field-named and
actionable.

Full findings: `docs/teams/2026-09-15-bug-76690-wbt-004-005-006/bug-wbt-004/04-review-r1.md` in the
`stylebi-wiz` repo.

jshobe-inetsoft (approved):
## Manual Review (replacing the failed automated `Claude Code Review / claude-review` check)

The automated `claude-review` GitHub Action check failed after ~1s with no output, the known signature of the review bot hitting its usage quota rather than performing an actual review. This is a manual review performed in its place.

### Summary of the change

`RestJsonQueryRunner.doLookups()` evaluates `customLookups[i].jsonPath` against the parent row via Jayway JsonPath and previously only handled a resolved `Map` (`doLookupOnEntity`) or `Array` (per-element `doLookupOnEntity`). A resolved scalar (e.g. `jsonPath:"$.id"`) fell through both `if`/`else if` branches with no `else`, silently leaving the row unmodified while the caller still reported success. This PR adds a fail-loud `else if(transformedJson != null)` branch that throws `IllegalStateException` naming the offending `customLookups[depth]` level and its `jsonPath`.

### Correctness

- **Fix logic is sound.** I traced `RestJsonQuery.getLookupQueries()` and confirmed `query.getLookupDepth()` in the loop body always corresponds 1:1 to the `customLookups[depth]` index actually being evaluated at that recursion level (`getLookupQueries()` returns a singleton list keyed by `lookupDepth`), so the exception message correctly identifies the failing level.
- **`null` is correctly excluded** from the new branch. `JsonTransformer` configures Jayway with `Option.DEFAULT_PATH_LEAF_TO_NULL`, so a legitimately-missing leaf returns `null` rather than throwing `PathNotFoundException`; the existing `catch(PathNotFoundException)` plus the new `transformedJson != null` guard together preserve that "leave the row alone" behavior for genuinely absent values, while only the previously-unhandled non-null scalar case now throws.
- **Consistent with existing convention.** `LookupService` already throws `IllegalStateException("Only lookups on json objects are supported.")` for the analogous "not a map" case on the `EndpointJsonQuery` lookup path; this change brings the plain `RestJsonQuery`/`customLookups` path in line with that pattern rather than inventing a new error-handling style.
- **Array-of-scalars path is unaffected.** The array branch (`doLookupOnEntity` per element, including the `jsonEntity instanceof String` case used for scalar array entries) is untouched, so this fix doesn't regress the deliberately-supported "array of raw ids" case — it only closes the gap for a *top-level* scalar result.

### Testing

I checked out the PR branch (`connectors/inetsoft-rest` module) and verified locally rather than relying solely on the PR description:
- `mvn -pl connectors/inetsoft-rest -am -DskipTests install` — builds clean.
- `mvn -pl connectors/inetsoft-rest -Dtest=RestJsonQueryRunnerCustomLookupJsonPathTest test` — 2/2 pass (scalar path throws with the expected message and leaves the row unmutated; `"$"` row path still adds `lookup1` as before).
- Broader regression pass: `mvn -pl connectors/inetsoft-rest -Dtest="*RestJson*,*Lookup*,ExpandedJsonTest" test` — 23 tests, 0 failures, 0 errors (covers `LookupServiceTest`, `EndpointJsonQueryLookupTest`, `JsonLookupIntegrationTest`, `XMLLookupIntegrationTest`, etc.).
- CI's own "build" workflow (full validation build) also passed.

No existing test is broken by this change.

### Code quality / readability

- The new branch is small, additive, and matches the surrounding indentation/style of the file. Minimal-scope, as it should be for a bug fix.
- Error message is clear and actionable: it names the level, the offending `jsonPath`, and suggests the correct fix (`"$"` + `key`).
- Test class follows the module's existing conventions (JUnit 5, AGPL header, `connectors/inetsoft-rest/src/test/java/...` layout) matching sibling test files.

1. **Backward compatibility trade-off (inherent to the fix, not a defect):** any existing/production `customLookups` configuration that happens to have a scalar `jsonPath` (previously silently a no-op) will now throw and abort the query instead of quietly completing with a missing lookup column. This is the intended, correct "fail loud" behavior per the bug report and matches the established convention elsewhere in this file, but it is a behavior change for any pre-existing misconfiguration relying on the old silent no-op — worth a mention in release notes if such configs are known to exist in the wild.
2. **Pre-existing, unrelated issue noticed while testing (not introduced by this PR):** `RestJsonQueryRunner.runLookupQuery()` has a broad `catch(NullPointerException ex)` that logs and swallows the exception, so a lookup that fails for an unrelated NPE reason still results in a `lookupN` property being set (to an empty list) rather than surfacing an error. This is out of scope for this PR (not touched by the diff) and shouldn't block it, but could be a good follow-up cleanup given the "fail loud" spirit of this fix.
3. The PR targets `stylebi-wiz-main-rebase` rather than `main`. That branch is currently a superset of `main` (151 commits ahead, 0 behind), so this isn't a merge-conflict concern, just flagging it in case it wasn't the intended base.

### Checks / mergeability

- `build` (Validation Build): **pass**
- `Semgrep Scan` / `Semgrep OSS`: **pass**
- `license/cla`: **pass**
- `claude-review`: **fail** (quota-exhaustion signature — ignored per review policy; this manual review supersedes it)
- Mergeable: **yes**, no merge conflicts.

### Verdict

No blocking issues found; all non-`claude-review` checks pass; no merge conflicts. **Approving.**

PR #5266, opened 2026-09-15
Fix Rest.XML xpath namespace resolution and swallowed load error (#76690 WBT-005)

fredaan:
## Summary

Redmine #76690, WBT-005: a Rest.XML `add_table` `xpath` using the response document's real
namespace prefix (e.g. `/wb:countries/wb:country` against a document declaring
`xmlns:wb="..."`) was rejected with a bare "returned no columns... see the server log"
message, while the naive unprefixed guess (`/countries/country`) was silently accepted and
returned one fabricated `{"Column": null}` row.

Root cause: `AbstractXMLStreamTransformer` builds its XSLT stylesheet by substituting the
caller's xpath as literal text into an `xsl:template match="..."` attribute, but the generated
stylesheet only ever declared `xmlns:xsl`/`xmlns:is` - never anything from the actual response
document. So a correctly-prefixed xpath had an unbound prefix in the stylesheet's own scope and
failed to *compile* (`TransformerConfigurationException`), while an unprefixed one compiled but
structurally matched zero real elements (XSLT 1.0 match-pattern rule: an unprefixed name test
only matches elements in no namespace). The compile exception was also swallowed three
normal-return layers deep (`RestXMLQueryRunner.runStream()`'s catch -> `QueryRunner.run()`'s
zero-size-to-null collapse -> `TabularHandler.execute()`'s null-table branch), so
`wizLoadColumnsError` never got set and the rejection carried no cause detail at all.

- `AbstractXMLStreamTransformer` now peeks the response body's root-element namespace bindings
  (a cheap StAX scan of just the root start element) and injects them onto the generated
  stylesheet's `$namespaces` placeholder, added to all three affected resources
  (`basic-xpath.xslt`, `iteration-xpath.xslt`, `paginated-xpath.xslt` - confirmed to share the
  identical defect). A correctly-prefixed xpath now resolves against the real document.
- `RestXMLQueryRunner.runStream()` now sets `query.setProperty("wizLoadColumnsError", ...)`
  directly inside its catch block, bypassing the intermediate null-collapsing layers, so the
  actual cause (e.g. an unresolvable namespace prefix) survives to the `add_table` rejection
  message instead of an unattributed "no columns" failure.

**Documented, explicitly out-of-scope limitation**: a response using a *default*
(unprefixed, `xmlns="..."`) namespace is not covered - XPath/XSLT 1.0 node-test rules never
resolve an unprefixed name test against any in-scope namespace, default or not, so callers of
such documents still need the `local-name()` wildcard idiom even after this fix.

Full diagnosis/refute/fix trail: `docs/teams/2026-09-15-bug-76690-wbt-004-005-006/bug-wbt-005/`
in the `stylebi-wiz` repo.

## Test plan

- [x] New `connectors/inetsoft-rest/src/test/java/inetsoft/uql/rest/xml/RestXmlNamespaceXpathTest.java`
      (4 cases against a World-Bank-shaped fixture): prefixed xpath now matches real data,
      unprefixed xpath still matches nothing (unchanged, spec-correct), `local-name()` wildcard
      idiom still works, and an unresolvable prefix now surfaces a non-null `wizLoadColumnsError`.
- [x] `./mvnw -q -o -pl connectors/inetsoft-rest clean && ./mvnw -q -o -pl connectors/inetsoft-rest -Dtest="inetsoft.uql.rest.xml.**" test -Dsurefire.failIfNoSpecifiedTests=false` - all green (new test 4/4, every other test in the package 0 failures/errors).
- [x] Re-ran the existing (unrelated, currently `@Disabled`) `XSLTTest.java` to confirm no regression, then restored `@Disabled`.


fredaan (commented):
# Review R1 - StyleBI PR #5266 (WBT-005, Redmine #76690)

## Verdict

**Keys on the confirmed cause.** The diff matches 01-diagnosis.md's revised (post-refute)
root cause exactly: (1) `AbstractXMLStreamTransformer`/`basic-xpath.xslt`/`iteration-xpath.xslt`/
`paginated-xpath.xslt` now inject the response document's own root namespace bindings onto the
generated stylesheet instead of leaving it `xsl`/`is`-only, and (2)
`RestXMLQueryRunner.runStream()`'s catch now sets `wizLoadColumnsError` directly, bypassing the
two normal-return collapse points (`QueryRunner.run()`, `TabularHandler.execute()`) the refuter
traced. No scope drift onto the reported symptom alone - the fix targets the XSLT
namespace-resolution mechanism and the error-swallowing chain identified in diagnosis/refute, not
just "make the World Bank URL work."

**2 Important findings** (below), both in the same new method
(`injectNamespaceDeclarations`/`peekRootNamespaces`), no blockers. CI `build` check is still
pending as of this review - see CI/local-test section; do not merge on this review alone until
it goes green (or accept my local substitute run below as the interim signal, per the dispatch).

---

## Important (80-89)

### 1. `injectNamespaceDeclarations` only escapes `"`, not `&` or `<`, in a namespace URI taken from untrusted response content

`AbstractXMLStreamTransformer.java`, new method `injectNamespaceDeclarations` (around line 82-92
of the diff):

```java
final String declarations = namespaces.entrySet().stream()
   .map(e -> "xmlns:" + e.getKey() + "=\"" + e.getValue().replace("\"", "&quot;") + "\"")
   .collect(Collectors.joining(" "));
```

This is a hand-rolled XML-attribute-value serializer, deliberately bypassing
`XSLTParamTransformer`'s existing (safe) HTML-attribute-encoding param mechanism - a reasonable
call for *why* (that mechanism would over-encode literal `xmlns:prefix="uri"` syntax), but the
replacement only escapes `"`. A namespace URI containing a literal `&` (legal XML-namespace-URI
content, and not unusual for a query-string-shaped URI, e.g. `http://example.org/ns?a=1&b=2`) is
copied through unescaped, producing a stylesheet with a bare `&` in an attribute value - which is
not well-formed XML. A namespace URI containing a literal `<` would be worse: it breaks out of
the attribute value entirely and splices arbitrary markup into the `xsl:stylesheet` root's start
tag / document.

Two distinct consequences, worth calling out separately:

- **Correctness**: an `&`-bearing namespace URI turns a would-have-worked case into a stylesheet
  parse failure. Given Part 2 is now in place, this fails loud (caught by
  `RestXMLQueryRunner.runStream()`, surfaced via `wizLoadColumnsError`) rather than silently
  corrupting data - so it degrades no worse than the pre-fix behavior for a *correctly*-namespaced
  document, but it does mean a legitimately correct, `&`-bearing namespace URI still won't work
  after this fix, undocumented as a limitation (unlike the default-namespace gap, which *is*
  explicitly documented).
- **Security-adjacent**: this method now splices attacker-controlled content (the external REST
  server's own response, per this repo's own stated threat model - "parsing untrusted response
  content") into a stylesheet document that already has Xalan Java-extension resolution enabled
  (`xmlns:is="xalan://inetsoft.uql.rest.xml.xslt.TransformFunctions"` is the existing, trusted use
  of that mechanism). A response root that declares e.g. `xmlns:wb="xalan://some.Class"` (instead
  of a real URL) gets that URI copied verbatim into the compiled stylesheet's namespace scope. I
  was not able to construct a full exploit chain in the time available - my reading is that Xalan
  only resolves a namespace as a Java-extension binding when it's referenced in a *function-call*
  QName position in an evaluated XPath expression, not when it's merely bound and used as an
  element name-test's namespace (which is the only way the caller's own `$xpath` match pattern
  uses it) - so I believe ordinary use is not exploitable via this path alone. But combined with
  the missing `<` escaping above (which allows breaking out of the attribute value and injecting
  literal sibling markup, not just a namespace binous string), I'd treat "attacker-controlled
  response content now contributes literal text into a stylesheet with extension functions
  enabled" as something that deserves an explicit security sign-off rather than being assumed safe
  by construction - this is a genuinely new trust boundary this PR introduces (previously the
  stylesheet's only namespaces were `xsl`/`is`, both author-controlled).

Suggested fix: escape `&` first, then `"`, then `<` (standard XML attribute-value escaping order),
in `injectNamespaceDeclarations`. Optionally also reject/skip namespace URIs using non-`http(s)`
schemes (in particular anything matching `xalan:` or other extension-mechanism scheme prefixes)
before injecting them, as defense in depth, since there is no legitimate reason a response
document's namespace URI needs to resolve as a Java-extension binding.

### 2. (Related, same root cause as #1) No test exercises a namespace URI containing `&`, `<`, or an extension-scheme-shaped value

The four new `RestXmlNamespaceXpathTest` cases all use a single, simple, bare `http://` URI
(`http://www.worldbank.org`) with no special characters. Given finding #1, a regression test with
a URI containing `&` (asserting it still resolves correctly, once escaping is fixed) would have
caught this before merge. Not asking for an adversarial/security-fuzz test here - just one
additional case with a URI shaped like a real-world query-string namespace (e.g.
`http://example.org/ns?a=1&b=2`) once #1 is fixed.

---

## Nits (not blocking, confidence < 80, FYI only)

- `peekRootNamespaces` hardcodes the StAX property name string `"javax.xml.stream.isSupportingExternalEntities"`
  instead of referencing the `XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES` constant (it does
  use the constant for `SUPPORT_DTD` on the line above). Purely stylistic - the string is correct.
- `transform(InputStream)` now unconditionally does `input.readAllBytes()` before the StAX peek,
  fully buffering every response body in memory where the prior code streamed it directly into
  the transformer. For a very large paginated/iteration response this is a real (if likely
  acceptable) change in peak-memory behavior. Not flagging as Important since I have no evidence
  this module's responses are ever large enough for it to matter in practice, and the alternative
  (peek via a mark/reset-capable stream wrapper) is meaningfully more complex for a benefit I
  can't demonstrate is needed.

---

## Checklist from the dispatch

**Part 1 - namespace injection**

- StAX peek is cheap/safe: confirmed. `peekRootNamespaces` stops at the first `START_ELEMENT`
  (explicit `break` after collecting that element's namespace declarations) - it does not walk the
  rest of the document. DTD support (`XMLInputFactory.SUPPORT_DTD`) and external entities
  (`"javax.xml.stream.isSupportingExternalEntities"`) are both explicitly set `false` on the
  factory used for this peek - genuinely disabled, not just claimed in the fix doc's prose.
- Literal, unescaped substitution reasoning: sound in intent (why it bypasses
  `XSLTParamTransformer`'s HTML-attribute-encoding), but the escaping it does instead is
  incomplete - see Important #1 above. A namespace URI containing `&` is not handled; this is a
  real, not merely theoretical, gap given namespace URIs are free-form strings, not restricted to
  a `&`-free subset.
- All three XSLT resources changed consistently: confirmed by direct diff read -
  `basic-xpath.xslt`, `iteration-xpath.xslt`, and `paginated-xpath.xslt` each replace their old
  "add additional namespace mappings from source document" dead comment with the same
  `$namespaces` placeholder on the `xsl:stylesheet` root, plus an explanatory comment.
- Transformer-caching change (`transformerNamespaces` field, widened rebuild condition): correct.
  `Map<String,String>.equals()` (both sides are `LinkedHashMap`/created via `new LinkedHashMap<>()`
  in `peekRootNamespaces`) compares by entry-set content, not reference, so "namespaces changed
  since last built" is detected reliably regardless of insertion order. No pathological
  performance concern: the peek itself is cheap (stops at first element), and a full
  transformer rebuild only happens when the map actually differs (or on the first call, where
  `transformer == null` already forces it independently of the new condition).
- Documented default-namespace limitation: accurately described, and the regression test
  (`unprefixedXpathStillMatchesNothingInNamespacedDocument`) explicitly asserts the pre-existing
  one-fabricated-row behavior is unchanged (still exactly one row), not fixed - matches what the
  fix doc claims.

**Part 2 - wizLoadColumnsError surfacing**

- `RestXMLQueryRunner.runStream()`'s catch now calls `query.setProperty("wizLoadColumnsError",
  detail)` where `detail` is `ex.getMessage()` falling back to `ex.getClass().getSimpleName()` if
  the message is null/blank - a clean, caller-facing string, not `ex.toString()`/a raw stack
  trace. Good.
- Regression risk for other exception types reaching this same catch: checked directly by reading
  `WorksheetAgentController.java` (the only two-guard structure that reads this property, lines
  977-985 in the current worktree state) - `loadError` is read **only inside** the
  `if(columns == null || columns.getAttributeCount() == 0)` branch. Setting `wizLoadColumnsError`
  more often (for every exception this catch now handles, not just namespace-prefix failures)
  cannot cause a false rejection for a query that actually loaded columns successfully, since the
  guard that decides "reject or not" is independent of this property - it only supplies detail
  text once rejection was already going to happen for an unrelated, already-correct reason. No
  regression found.
- The "not re-verified with a live web-layer test" gap: I read
  `WorksheetAgentController.java:979-983` directly (not just trusting the fix doc's prose). The
  read side is exactly as simple as the fix doc assumes - a direct `query.getProperty(...)` read
  used only to build a message string. I also traced the intermediate write path
  (`TabularTableAssembly.java:188` clears the property to `null` before
  `loadOutputColumns(vtable)`; line 208 is the pre-existing catch that would also set it on a
  thrown exception) and confirmed no other code path resets the property between the new
  `RestXMLQueryRunner` write and the `WorksheetAgentController` read. This is a low-risk gap, not
  a blocker - agree with the fix doc's own assessment.

**Regression test quality**

- All four new `RestXmlNamespaceXpathTest` cases use real production classes end-to-end - verified
  `TestHttpHandler`/`TestHttpResponse`/`RequestResponse` are real, pre-existing test-support
  classes in `inetsoft.test` (not fabricated for this PR), reused the same way
  `XMLLookupIntegrationTest` in the same module already uses them.
- `unresolvablePrefixSurfacesAsWizLoadColumnsError` (case 4) does meaningfully distinguish "Part 2
  fixed" from "was already working by coincidence": it calls `RestXMLQueryRunner.run()` directly
  (not through the outer `TabularHandler`/`AbstractRestRuntime` layers), and without Part 2's
  `query.setProperty(...)` call in the catch block, `wizLoadColumnsError` would stay `null` for
  this case exactly as it did pre-fix (per the refuter's traced chain) - the assertion would fail
  without Part 2, regardless of Part 1.
- Stale-`.class`-file note in the fix doc (2 false-alarm failures from a since-deleted
  `RefuterNamespaceXsltRealClassesTest`, resolved by `mvn clean`): plausible and consistent with
  what I saw in my own run below (I ran `clean` first and saw no such failures). Not a
  `.gitignore`/build-hygiene issue worth a nit - `target/` is already excluded from version
  control; this is an ordinary stale-incremental-build artifact, not a repo hygiene gap.

**Other checks**

- Prompt-merge gate: confirmed not applicable. `gh pr diff 5266` touches only
  `connectors/inetsoft-rest/src/main/java/...`, `connectors/inetsoft-rest/src/main/resources/*.xslt`,
  and one new test file under `connectors/inetsoft-rest/src/test/java/...` - nothing under
  `wiz-services/src/agents/prompts/` or any `stylebi-wiz` path at all (this PR is entirely inside
  the `stylebi` repo's `community` submodule).
- Code style/sizing: methods are reasonably sized and named
  (`peekRootNamespaces`/`injectNamespaceDeclarations` each do one clearly-scoped thing); follows
  existing file conventions.

## CI / local-test status

- `gh pr checks 5266`: `build` (Validation Build) is still **pending** at review time;
  `claude-review` shows **fail**, but its own log
  (`https://github.com/inetsoft-technology/stylebi/actions/runs/34953459887/job/104329741748`)
  shows this is an infra/execution failure of the automated review action itself (empty
  `ANTHROPIC_API_KEY` in that job's env, "Claude result reported subtype success with
  is_error:true"), not a posted finding about this diff - no actionable content to react to there.
  `Semgrep OSS`, `Semgrep Scan`, and `license/cla` all pass.
- Ran the fixer's own command locally as an interim substitute, in the same worktree
  (`E:\inetsoft\stylebi\community\.worktrees\76690-wbt-005-xpath-namespace`, confirmed clean via
  `git status --short` and at the exact same commit as the PR's head,
  `d0faa1876b4fb6890f945a9bbf72fa5280c4eeb3`, confirmed via `gh pr view --json headRefOid`):
  ```
  ./mvnw -q -o -pl connectors/inetsoft-rest clean
  ./mvnw -q -o -pl connectors/inetsoft-rest -Dtest="inetsoft.uql.rest.xml.**" test -Dsurefire.failIfNoSpecifiedTests=false
  ```
  Result: all green, matching the fix doc's own report - `RestXmlNamespaceXpathTest`: 4/4 pass;
  every other test in the `inetsoft.uql.rest.xml` package (`XMLPagedIterationIteratorStrategyTest`,
  `XMLPagedPageCountIteratorStrategyTest`, `XMLLookupIntegrationTest`, `DocumentParserTest`,
  `MapNodeTest`, `NoSchemaDocumentParserTest`, `XSLTTest` (skipped, `@Disabled`),
  `XTypeNodeFinderTest`): 0 failures, 0 errors (per `target/surefire-reports/*.txt`). Some noisy
  `SreeEnv`/`PropertiesEngine` NPE stack traces appeared on stderr during the run but did not
  correspond to any failed test (confirmed via the surefire report `.txt` summaries, not just exit
  code) - looks like benign Spring-context logging noise in this test harness, not a real failure.
- **Do not treat this PR as merge-ready on this local substitute alone** once the real `build` CI
  check finishes - if it comes back red, that supersedes my local run (my worktree, JDK/Maven
  version, and local `.m2` cache could differ from CI in ways a local run wouldn't catch).

## Findings posted

Posted via `gh pr review 5266 --repo inetsoft-technology/stylebi --comment` (not `--approve`, per
the self-review/self-merge restriction on this batch's own agents).

fredaan (commented):
# Review R2 - StyleBI PR #5266 (WBT-005, Redmine #76690)

## Verdict

**Clean. Both round-1 findings genuinely fixed, verified against the actual diff and my own
independent test run - no new blocker/important findings.** Commit `5797c476e` (on top of
`d0faa1876`), confirmed as the PR's current `headRefOid`, worktree clean, `mergeable: MERGEABLE`.

---

## Finding 1 (escaping) - verified fixed

Read the actual diff, not just the fix doc's prose. New `escapeAttributeValue`:

```java
private static String escapeAttributeValue(String value) {
   return value.replace("&", "&amp;").replace("\"", "&quot;").replace("<", "&lt;");
}
```

Order is correct: `&` is replaced **first**. If `<` or `"` were escaped before `&`, the entities
those replacements introduce (`&lt;`, `&quot;`) would themselves contain a literal `&` that the
subsequent `&`->`&amp;` pass would then corrupt into `&amp;lt;`/`&amp;quot;`. Escaping `&` first
means every literal `&` in the *original* value is converted before any entity text exists, and
the later `"`/`<` replacements can't introduce new `&` characters, so nothing gets double-escaped.
This is the textbook-correct order for sequential-replace-based XML attribute escaping, and the
fixer's own comment states the reasoning accurately.

Confirmed via my own local test run (not just trusting the fix doc): re-ran
`RestXmlNamespaceXpathTest` after a `mvn clean` - all 5 cases pass, including the new
`namespaceUriWithAmpersandIsCorrectlyEscapedWhenInjected`, which uses a fixture with
`xmlns:ex="http://example.org/ns?a=1&amp;b=2"` (XML-escaped on the wire, i.e. the *parsed* value
StAX hands back is the raw string containing a literal `&`) and asserts the query actually resolves
and loads the one real row - not merely that it doesn't crash. This is the right kind of test:
it would fail under the old `.replace("\"", "&quot;")`-only escaping (malformed stylesheet ->
compile exception -> zero rows), and passes now.

## Finding 1b (security-adjacent, http(s)-only allowlist) - verified fixed at the right point

`isHttpUri(uri)` is checked **inside the `peekRootNamespaces()` collection loop**, before a
binding is ever added to the `namespaces` map:

```java
if(prefix != null && !prefix.isEmpty() && isHttpUri(uri)) {
   namespaces.put(prefix, uri);
}
```

This is the correct placement the dispatch asked me to confirm - a non-`http(s)` URI (in
particular anything shaped like `xalan://...`) is never collected in the first place, so it never
reaches `injectNamespaceDeclarations` at all. It isn't filtered "after injection where it'd be too
late" - it's excluded before it's even a candidate. `isHttpUri` itself is a simple, correct
case-insensitive (`Locale.ROOT`, avoiding locale-dependent case folding) prefix check against
`"http://"`/`"https://"`. This closes the round-1 concern I raised (a response could otherwise get
its own arbitrary-scheme namespace URI spliced into a stylesheet that already has Xalan
Java-extension resolution enabled via `xmlns:is="xalan://..."`) - regardless of whether that was
ever actually exploitable (I was not able to build a full exploit chain in round 1 either), the
allowlist removes the question rather than relying on an argument about Xalan's resolution rules,
which is the right way to close an "I couldn't prove it's exploitable but couldn't prove it isn't
either" gap.

## Finding 2 (test coverage) - verified adequate; agree with the fixer's judgment call on the omitted case

The new `namespaceUriWithAmpersandIsCorrectlyEscapedWhenInjected` test is exactly what round 1
asked for - a query-string-shaped namespace URI, asserting correct round-trip (not just "doesn't
crash"). Confirmed it's wired through a real fixture and real production classes, same harness
pattern as the other four cases (refactored `runQuery`/`createRunner` to take an optional fixture
parameter, the existing four cases unchanged).

**On the omitted `xalan:`-scheme test**: I agree with the fixer's call not to add one. My own
round-1 ask was framed as "an explicit security sign-off... or a test" for the allowlist -
the allowlist itself is a three-line, directly-readable prefix check with no branching complexity
to hide a bug in, and there's no demonstrated exploit path it's guarding against (both the fixer
and I independently concluded Xalan only resolves an extension binding in a function-call
position, not a plain node-test namespace). A test asserting "a `xalan:` URI is silently skipped"
would mostly be re-asserting `isHttpUri`'s own literal logic back at itself. Not worth blocking on.

---

## New consideration this round (not blocking - a scope/documentation note, not a defect)

The `http(s)`-only allowlist is good defense in depth, but it also narrows this fix's own
applicability in a way that isn't currently documented anywhere (unlike the existing, explicitly
documented default-namespace limitation): a Rest.XML datasource whose root uses a **non-http(s)**
namespace scheme - `urn:` is the common real-world case (SOAP/enterprise XML feeds routinely use
`urn:schemas-...`-style namespace URIs, not URLs) - still gets the original pre-fix behavior for a
correctly-prefixed xpath: an unresolved-namespace-prefix compile failure, now at least surfaced via
`wizLoadColumnsError` (Part 2 still applies), but the namespace itself is silently never injected,
indistinguishable from "we couldn't parse this as XML at all." This isn't a regression - it's
already better than before this whole PR (loud, named error instead of nothing) - but it does mean
"only prefixed documents get first-class namespace support" (the fix doc's own phrase for the
default-namespace case) is now actually "only prefixed documents **using an http(s) namespace
URI**" - worth a one-line addition to the fix's documented-limitations note for whoever picks this
up next, not a change I'd send back for. Not counted as a blocker/important finding.

---

## Re-verification performed (not just trusting the fix doc)

- Read the actual diff via `gh pr diff 5266 --repo inetsoft-technology/stylebi` (not the fix doc's
  prose) for every changed line in `AbstractXMLStreamTransformer.java` and
  `RestXmlNamespaceXpathTest.java`.
- Confirmed worktree state before testing:
  `git --git-dir=.../worktrees/76690-wbt-005-xpath-namespace status --short` -> clean;
  worktree `HEAD` = `5797c476e323d0fc69b4359e2dc16e2ff415e09f`, matching
  `gh pr view 5266 --json headRefOid` exactly; `mergeable: MERGEABLE`.
- Ran, myself, after a fresh `clean`:
  ```
  ./mvnw -q -o -pl connectors/inetsoft-rest clean
  ./mvnw -q -o -pl connectors/inetsoft-rest -Dtest="inetsoft.uql.rest.xml.**" test -Dsurefire.failIfNoSpecifiedTests=false
  ```
  Result (from `target/surefire-reports/*.txt`, not just console exit code):
  `RestXmlNamespaceXpathTest`: **5/5 pass** (0 failures, 0 errors), including the new
  ampersand-escaping case. Every other test in the `inetsoft.uql.rest.xml` package: 0
  failures, 0 errors across all 9 test classes.

## CI status

`gh pr checks 5266`:
- `build` (Validation Build): still **pending**.
- `claude-review`: **fail** - same pre-existing infra failure as round 1 (`Run Claude Code Review`
  step itself fails with exit 1; not a posted finding about the diff).
- `Semgrep Scan`: **fail**, newly so this round (passed in round 1) - but checked the job's own
  step breakdown (`gh run view 34955122358`): the `Run Semgrep` step itself **passed** (no findings
  reported); only the subsequent `Upload SARIF to GitHub Code Scanning` step failed, with no
  visible error text in its own log (job completes "Uploading results" then runs post-job cleanup
  normally). This reads as a CI/GitHub-code-scanning-API infra issue (e.g. permissions or
  processing-wait timeout for that upload step), not a Semgrep-reported code finding about this
  diff - there is nothing in the run's annotations or step log naming an actual security finding.
- `license/cla`: pass.

**Do not merge on this review alone until `build` goes green** - my local run above is an interim
substitute, not a replacement, per the same caveat as round 1 (different JDK/Maven/cache state
than CI could hide something CI would catch).

## Findings posted

Posted via `gh pr review 5266 --repo inetsoft-technology/stylebi --comment` (not `--approve`, per
the self-review/self-merge restriction on this batch's own agents).

jshobe-inetsoft (changes requested):
## Manual review (replacing the failed automated Claude Code Review)

The `Claude Code Review / claude-review` check failed in ~1 second with no turns/cost — the
known signature of the review bot hitting its usage quota rather than performing an actual
review. This is a full manual review performed in its place.

### Summary of the change

Fixes Redmine #76690 (WBT-005): a Rest.XML `add_table` `xpath` using the response document's
real namespace prefix (e.g. `/wb:countries/wb:country`) was rejected, while the incorrect
unprefixed guess silently returned one fabricated null-column row. Root cause and fix are well
diagnosed in the PR description:
- `AbstractXMLStreamTransformer` now peeks the response body's root-element namespace bindings
  via a cheap StAX scan and injects them into the generated stylesheet's `$namespaces`
  placeholder (added consistently to all three affected templates: `basic-xpath.xslt`,
  `iteration-xpath.xslt`, `paginated-xpath.xslt`).
- `RestXMLQueryRunner.runStream()` now sets `wizLoadColumnsError` directly in its catch block so
  the real cause survives past the intermediate null-collapsing layers instead of producing an
  unattributed "no columns" failure.

### Correctness / completeness

- Verified all three `.xslt` resources that share the transformer code path were updated
  consistently — no template was missed.
- Verified the escaping helper (`escapeAttributeValue`) escapes `&` before `"` and `<`, which is
  the correct order to avoid double-escaping and to prevent a namespace URI from breaking out of
  the generated attribute (the second commit on this branch, "Fix XML-escaping gap ... (review
  r1)", already addresses an earlier gap here — good sign that this got a real review pass).
- `isHttpUri()` filtering non-`http(s)` schemes before injecting is sensible defense-in-depth
  against colliding with the stylesheet's own `xmlns:is="xalan://..."` extension binding.
- Built `connectors/inetsoft-rest` offline and ran
  `-Dtest="inetsoft.uql.rest.xml.**"`: all pre-existing tests in the package are unaffected
  (same pass/skip counts as before), and the new `RestXmlNamespaceXpathTest` passes 5/5
  (prefixed xpath resolves, unprefixed still matches nothing per spec, `local-name()` wildcard
  still works, `&`-containing namespace URI round-trips correctly, and an unresolvable prefix now
  surfaces a non-null `wizLoadColumnsError`). This matches the PR description's test plan.

### Potential issues / things worth a second look (non-blocking, for awareness)

1. **Full in-memory buffering of the response body.** `transform(InputStream)` now does
   `input.readAllBytes()` up front (to peek namespaces) and re-wraps the bytes in a new
   `ByteArrayInputStream` for the actual transform, whereas previously the original stream was
   passed straight through to `runXSLTTransformer`. For an `AbstractXMLStreamTransformer`, whose
   whole premise is streaming, this forces the entire response into memory even for very large
   paginated/iterated XML payloads. Probably an acceptable trade-off to fix the bug, but worth
   confirming this connector isn't used against endpoints that can return very large bodies
   where this would meaningfully increase memory pressure.
2. **Reserved-prefix collision not handled.** If a response document happens to declare a
   namespace under the prefix `is` (or `xalan`/`xsl` for `iteration-xpath.xslt`), the injected
   `xmlns:is="..."` would duplicate the stylesheet's own hardcoded `xmlns:is="xalan://..."`
   attribute on the same element, which is an XML well-formedness violation (duplicate attribute)
   and would fail stylesheet compilation. This is an edge case (uncommon prefix choice), and the
   failure mode is at least no longer silent — it will surface via the new `wizLoadColumnsError`
   plumbing — so it isn't a regression, just an unhandled corner case that could be worth a guard
   or an explicit comment/test.
3. **Exception messages now surface to `wizLoadColumnsError`.** This is the intended fix, but it
   does mean raw exception text (potentially including endpoint/parsing details) that previously
   only went to the server log is now also returned via the query property to whatever UI
   surfaces `add_table` errors. Given the intent of the fix is exactly to surface actionable
   detail to the person configuring the data source, this seems like a reasonable and deliberate
   trade-off, not a vulnerability — just flagging it as a slightly larger information surface than
   before.

None of the above are blocking; they're the kind of thing worth a mental note rather than a
change request.

- All touched methods (`createXSLTTransformer`, `createXSLTSource`) are `private`, so there's no
  API surface change for subclasses/callers.
- The documented out-of-scope limitation (default/unprefixed `xmlns="..."` namespaces still
  require the `local-name()` idiom) is accurate per XPath/XSLT 1.0 semantics and is clearly
  called out rather than silently left as a surprise.
- No test cases were broken; the new test class adds targeted regression coverage and the
  existing `XSLTTest` (already `@Disabled`) was manually re-run per the PR description and
  restored to its prior state.

### CI / mergeability status (per review criteria)

- **`Semgrep Scan` check is currently FAILING** on this PR (separate from the excluded
  `claude-review` check). Looking at the job, the `Run Semgrep` step itself succeeded; the
  failure is in the subsequent `Upload SARIF to GitHub Code Scanning` step, which shows no
  explicit error in its log — this looks like a CI/infra hiccup (e.g. code-scanning upload
  issue) rather than a Semgrep finding in this diff. That said, per the review's decision rule,
  any non-`claude-review` failing check blocks approval regardless of root cause.
- `build` (Validation Build) is passing.
- `license/cla` is passing.
- No merge conflicts — GitHub reports `MERGEABLE`.

### Decision

Requesting changes **solely because of the failing `Semgrep Scan` check** (excluding
`claude-review`, which is disregarded per instructions). The code itself has no blocking issues
found in this review — please re-run/investigate the `Semgrep Scan` job (it appears to be a SARIF
upload/infra failure rather than an actual finding in this diff) and this can be approved once
checks are green.

fredaan:
## Semgrep Scan resolved — re-run, not a code issue

Investigated the `Semgrep Scan` failure flagged in the manual review above.

**Root cause**: the `Run Semgrep` step itself completed successfully with **0 findings** (confirmed
in the job log: `Ran 65 rules on 9 files: 0 findings`). The failure was entirely in the subsequent
`Upload SARIF to GitHub Code Scanning` step, which failed with no error text in its own log. Cross-checked
via the Code Scanning API: this commit's SARIF was never registered as an analysis at all (unlike
every other recent PR's Semgrep run, which uploaded cleanly), and several *other*, unrelated PRs
around the same time hit the identical upload-step failure — confirming this was a transient,
repo-wide GitHub Code Scanning processing hiccup, not anything caused by this diff.

**Fix**: re-ran the failed job (`gh run rerun 34955122358 --failed`). It passed on retry (`Semgrep
Scan: pass`, 46s) with no code changes needed.

Current status: `build` ✅, `Semgrep Scan` ✅, `Semgrep OSS` ✅, `license/cla` ✅. Only
`claude-review` still shows failing — per the reviews already on this PR, that's the automated
bot's own infra error (0 turns, $0 cost, erroring in under a second), not a posted finding, and is
disregarded by the same reasoning both prior manual reviews applied.

This PR should now be clear for merge from a CI standpoint.

Nelson-cao (approved):

PR #5267, opened 2026-09-15
fix(76690): stop suffix from corrupting an existing query string (WBT-006)

fredaan:
## Summary

Redmine #76690, WBT-006: `queryParams.suffix` corrupted the request whenever the base URL already carried a query string (e.g. `https://api.worldbank.org/v2/country?format=xml`). `URLCreator.combineDataSourceAndQueryUrlFragments()` chose its join strategy purely from the base URL's trailing character (`/` or `?`), so any base URL that already had a complete query string not ending in `?` fell into the path-append branch, splicing a stray `/` into the middle of the existing query string regardless of `suffix`'s content — corrupting `format=xml` into `format=xml/` (or worse) and triggering the upstream API's own "requested response format was not found" error. This reproduced even for `suffix:""`.

## Fix

Detect an already-present query string (`url.contains("?") && !url.endsWith("?")`) ahead of the existing branches:
- blank `suffix` → leave the URL unchanged (nothing to append)
- non-blank `suffix` → strip any leading `?`/`&` and `&`-join it onto the existing query string

The pre-existing branches for a base URL ending in bare `?` (`.../country?` + `suffix`) and a base URL ending in `/` + a `/`-prefixed suffix are untouched and still pass their existing tests.

## Test plan
- [x] Added 4 regression tests to `URLCreatorTest`: the three reported `suffix` values against `https://api.worldbank.org/v2/country?format=xml` (all now produce clean, non-corrupted URLs), plus a control case confirming a base URL with no existing query string is unaffected.
- [x] `./mvnw -o -pl connectors/inetsoft-rest -Dtest=URLCreatorTest test` → `Tests run: 9, Failures: 0, Errors: 0` (5 existing + 4 new), `BUILD SUCCESS`.


fredaan (commented):
## Review R1 — WBT-006 (`queryParams.suffix` corrupts existing query string)

**Verdict: clean — no blockers, no important findings.**

Checked against the confirmed diagnosis + refuter's correction (`docs/teams/2026-09-15-bug-76690-wbt-004-005-006/bug-wbt-006/{01-diagnosis,02-refute}.md` in the stylebi-wiz repo):

1. **Root cause targeting** — the new `else if(url.contains("?") && !url.endsWith("?"))` branch is inserted after the `RestJsonQuery`-only `ignoreBaseUrl` branch and before the pre-existing `/`-trailing (line-74) and bare-`?`-trailing (line-77) branches, matching the diagnosis's confirmed location and ordering intent.
2. **Refuter's correction implemented, not just the diagnosis's original proposal** — `suffix.replaceFirst("^[?&]+", "")` strips leading `?`/`&` before `&`-joining, exactly per the refuter's §5. The `+` quantifier (strips *all* leading occurrences, not just one) is a deliberate, reasonable choice — there's no realistic suffix value with an intentional repeated leading separator.
3. **Verified the three reported URLs + control cases myself**, at the actual PR head commit (`8f267281a`, fetched via `pull/5267/head` and cross-checked against `gh pr view --json headRefOid`), not just by reading the fix doc's claims:
   - `suffix:""` → unchanged `.../country?format=xml`.
   - `suffix:"&per_page=5"` → clean `&`-join, no `&&`.
   - `suffix:"?format=xml&per_page=5"` → `.../country?format=xml&format=xml&per_page=5` — confirmed the duplicate `format=xml` comes from that suffix's own content, not a joining defect.
   - Control (no existing query string) → unaffected, still old path-append behavior. This is the key "scoped correctly" test and it passes.
4. **Ran `URLCreatorTest` locally** at the PR head commit: `Tests run: 9, Failures: 0, Errors: 0` — `BUILD SUCCESS` (5 existing + 4 new), independently reproduced, not just trusting the PR description's numbers.
5. **Scope gate**: confirmed via `gh pr view --json files` — only `URLCreator.java` + `URLCreatorTest.java` touched, both in `connectors/inetsoft-rest`. Nothing under `wiz-services/src/agents/prompts/`.
6. **CI status**: `Semgrep OSS`/`Semgrep Scan`/`license/cla` pass. `build` is still pending as of this review — my own local test run above is the interim substitute. `claude-review` shows `fail`, but the job log shows it errored out after 373ms with `total_cost_usd: 0` and 0 turns (`is_error:true`, no review content produced) — this reads as an infra/tooling failure of the bot itself, not an actual review finding against the code.

**One non-blocking note for the record**: the new branch's condition is checked before the line-74 (`/`-trailing) branch, so a base URL that both ends in `/` and already contains a `?` earlier (e.g. `.../search?q=1/`) would now take the new branch instead of line-74. I checked — no existing test (before or after this PR) actually exercises the line-74 branch at all, so this doesn't break any real test, and the URL shape (existing query string *then* a trailing `/`) is unusual enough that I don't consider it a blocker. Flagging only because it's a genuine, currently-untested behavior change for that specific shape.

jshobe-inetsoft (approved):
## Manual Code Review — PR #5267 (fix(76690): stop `suffix` from corrupting an existing query string, WBT-006)

*Posted in place of the "Claude Code Review / claude-review" automated check, which errored out after ~0.4s with `total_cost_usd: 0` / 0 turns (a subscription-quota infra failure of the bot, not an actual review of this code).*

### Summary of the change

`URLCreator.combineDataSourceAndQueryUrlFragments()` picked its join strategy for `suffix` purely from the trailing character of the base URL (`/` or `?`). A base URL that already carried a *complete* query string not ending in `?` (e.g. `https://api.worldbank.org/v2/country?format=xml`) fell through to the "append as path" branch, which spliced a `/` (or the raw suffix) into the middle of the existing query string — corrupting `format=xml` into `format=xml/` even when `suffix` was empty.

The fix inserts a new branch, checked before the existing `/`-trailing and bare-`?`-trailing branches:
```java
else if(url.contains("?") && !url.endsWith("?")) {
   if(!suffix.isBlank()) {
      url += "&" + suffix.replaceFirst("^[?&]+", "");
   }
}
```
Blank suffix → URL left untouched. Non-blank suffix → leading `?`/`&` stripped, then `&`-joined onto the existing query string.

### Correctness

I traced all five branches by hand against the four new tests and the pre-existing five tests:
- `suffix:""` against a URL with an existing query string → unchanged. Correct, fixes the reported corruption.
- `suffix:"&per_page=5"` → clean `&`-join, no `&&`. Correct.
- `suffix:"?format=xml&per_page=5"` → `...?format=xml&format=xml&per_page=5` (duplicate `format=xml`). This is a faithful join, not a defect — the duplication comes from the suffix's own content, and the test documents the behavior rather than hiding it.
- Control case (no existing query string) → unaffected, falls through to the untouched pre-existing branches.

The new branch is placed correctly relative to the pre-existing bare-`?` branch (`url.endsWith("?")` is mutually exclusive with the new condition, so no interference) and the `ignoreBaseUrl` branch (checked first, unaffected).

### Edge case / non-blocking observation

`suffix` is a generic, free-text "URL Suffix" field used elsewhere in this codebase purely as a **path** fragment (e.g. `HubSpotQuery` checks `endpoint.getSuffix().startsWith("/companies")`). The new branch assumes that if the base URL already contains a complete query string, `suffix` must be *query* content, and joins it with `&` (only stripping a leading `?`/`&`, not `/`). If a base URL happened to already contain a query string (e.g. an API key baked into a custom REST datasource URL) **and** `suffix` were a path fragment like `/companies`, the result would now be `...?apikey=x&/companies` instead of the previous (already-broken) `...?apikey=x/companies`. 

This is not a regression in the sense of breaking previously-correct behavior — that combination was already producing a corrupted URL before this PR, just via a different code path/failure mode — and I found no built-in connector that hardcodes a base URL containing `?`. It only affects the free-text "Custom URL" case, the same case this PR is fixing. Flagging for awareness only; not a blocker.

### Tests

- 4 new regression tests added to `URLCreatorTest`, covering blank/`&`-prefixed/`?`-prefixed suffix against a query-string base URL, plus a control case with no existing query string.
- All 5 pre-existing `URLCreatorTest` cases remain structurally untouched and are not affected by branch ordering (verified by manual trace; local `mvn test` for this module could not be run in this sandbox due to a stale/mismatched offline `.m2` cache for `inetsoft-core`, unrelated to this PR — the PR's own CI `build` check is green, and it reports `Tests run: 9, Failures: 0, Errors: 0`).
- No other production call sites of `combineDataSourceAndQueryUrlFragments()`/`URLCreator` exist to regress.

### Style / maintainability

- Matches surrounding code style (bracing, `else if` chain, no space after `if`).
- The inline comment explains *why* the branch exists, not just what it does — good.
- Minimal, scoped diff: only `URLCreator.java` + its test file touched.

### Security

- Pure string manipulation on an admin/user-supplied REST datasource configuration (base URL + suffix), same trust boundary as before. No new injection surface; the regex (`^[?&]+`) is a fixed, non-backtracking anchor pattern — no ReDoS concern.

### Backward compatibility

- No public API/signature changes (package-private method, private-scoped change). No persisted-asset format changes. The only behavior change is to the previously-corrupting code path itself, which is the intended fix.

### Checks / mergeability

- All CI checks pass except `claude-review`, which is being disregarded per the known bot-quota failure signature (errors in <1s, 0 turns, `$0` cost).

### Verdict

**No blocking issues.** Approving.

PR #5268, opened 2026-09-15
Fix WBT-007: NPE-message leak and residual broken table on failed add_table

fredaan:
## Summary

Redmine #76691, WBT-007. Two independent server-side defects in the tabular
`add_table` write path, both diagnosed and independently re-verified in a prior
debugger/refuter round of the debug workflow (see
`docs/teams/2026-09-15-bugs-76691-wbt-007-008-009/bug-wbt-007/` in the
`stylebi-wiz` companion repo for the full write-up).

**1. Raw NPE-message leak — `ServerFileQuery.loadOutputColumns()`**

`getColumns()` was called a second time and `.length` indexed on it directly.
`SelectableTabularQuery.getColumns()` documents returning `null` when the source
can't determine its columns — a legitimate, expected outcome for e.g. a genuinely
empty/headerless Excel sheet — but `loadOutputColumns()` never null-checked it, so
this degraded into an uncaught `NullPointerException`. That NPE's raw message
(JEP-358's `"...because the return value of \"...getColumns()\" is null"` text)
then leaked verbatim into the caller-visible `PairingException` via the existing
`wizLoadColumnsError` mechanism (stylebi #5137), producing a confusing,
implementation-detail error for what is really just "this sheet has no columns to
read."

Fixed by reusing the already-assigned `columnDef` from the first `getColumns()`
call, null-checking it, and throwing a clean, named `IOException` instead. This
also avoids a real double file-read on the failure path: a `null` result from
`getColumns()` is not cached by `SelectableTabularQuery`, so the removed second
call would have re-run the entire Excel-header-read I/O again before NPEing.

**2. Residual broken table — `WorksheetAgentController.addQueryParamsTable` /
`addTabularTable`**

Both methods' `applyOnRuntime` lambdas called `ws.addAssembly(assembly)` *before*
probing the query for columns via `loadColumnSelection`/`getColumnSelection`.
`WorksheetEditService.applyOnRuntime` has no rollback for a mutation lambda that
throws after partially mutating the live `RuntimeWorksheet` (an existing,
documented contract — see the comment at `WorksheetEditService.java:3017`), so a
query that genuinely produces zero columns left a broken, zero-column
`TabularTableAssembly` in the live worksheet even though the overall `add_table`
call reported failure — visible to a subsequent `read_worksheet_model` call.

Fixed by reordering each lambda to probe → check-empty → throw-if-empty → *then*
`ws.addAssembly`, mirroring `WorksheetTableService.buildTabularTable`'s existing
correct order for the identical operation. This ordering is already proven safe
in this exact codebase: `buildTabularTable` calls `loadColumnSelection`/
`getColumnSelection` on a `TabularTableAssembly` constructed with a worksheet
reference but not yet added via `ws.addAssembly`, in production, today.

This second defect is not connector-specific — `addTabularTable`'s REST-JSON
(`endpoint`/`suffix`) form has the identical shape and gets the identical fix
alongside `addQueryParamsTable`'s connector-agnostic (`queryParams`) form. It
affects any tabular connector reaching a genuine zero-column outcome, not only
`ServerFileQuery`/empty-Excel. `addBoundTable`/`addLogicalModelTable` do not call
`loadColumnSelection` (their columns come from metadata, not a query probe), so
they don't have this defect shape and are untouched.

## Test plan

- [x] `ServerFileQueryEmptySheetTest` (new): `ServerFileQuery.loadOutputColumns()`
  against a genuinely empty Excel sheet now throws a clean `IOException`, not
  `NullPointerException`. Verified failing before the fix, passing after.
- [x] `WorksheetAgentControllerTest#addTabularTableNoColumnsDoesNotLeaveResidualAssembly`
  and `#addQueryParamsTableNoColumnsDoesNotLeaveResidualAssembly` (new): a
  zero-column probe result leaves the worksheet with no residual assembly.
  Verified failing before the fix (assembly count 1, not 0), passing after.
- [x] `./mvnw -pl connectors/inetsoft-serverfile test` — 35/35 passing.
- [x] `./mvnw -pl core -Dtest='inetsoft.web.wiz.worksheet.WorksheetAgentControllerTest,inetsoft.web.wiz.worksheet.WorksheetEditServiceTest,inetsoft.web.wiz.worksheet.WorksheetEditServiceMutatorsTest,inetsoft.web.wiz.service.WorksheetTableService*' test` — 586/586 passing, no regressions in either affected method's existing coverage.


fredaan:
## Reviewer verdict (round 1) — code review clean; build check is currently RED, not pending

### Code review

Read `01-diagnosis.md` → `02-refute.md` → `03-fix.md` → `04-verify.md` alongside `gh pr diff 5268`
and independently re-traced the diff against the two confirmed root causes.

1. **`ServerFileQuery.loadOutputColumns()` fix is general, not repro-specific.** It null-checks the
   already-fetched `columnDef` and throws a plain `IOException` naming no sheet/file/datasource —
   applies to any empty/headerless source, not just `Sheet3`/`Excel Data.xlsx`.
2. **The reorder actually eliminates the residual-assembly window.** I read `positionBelowExisting`
   (`WorksheetAgentController.java:1181-1198`) directly: it only reads `ws.getAssemblies()` (existing
   assemblies) and mutates the *new, not-yet-added* `assembly` object's own pixel offset — it never
   touches `ws` itself. So there is no earlier statement between `positionBelowExisting` and the new
   `ws.addAssembly(assembly)` placement that could still leave residue; the reorder is a full fix, not
   a partial one.
3. **Third call site check.** Re-read `addBoundTable` (:533), `addLogicalModelTable` (:626), and
   `addDatasourceScopedNamedGroup` (:1011) myself — none call `loadColumnSelection`/probe a query
   inside the mutation lambda, so none have this shape; confirms the diagnosis/refuter's own
   conclusion, not just taking it on faith. **One related, out-of-scope finding**: `createEmbeddedTable`
   (`WorksheetAgentController.java:~2283-2330`, the CSV/Excel *embedded*-table import path, unrelated
   to the tabular-query path this bug is about) has the same add-before-confirm shape —
   `ws.addAssembly(assembly)` runs, then `AssetEventUtil.initColumnSelection(rws, assembly)` is called
   in a try/catch that rethrows as `PairingException` on failure, with no rollback of the just-added
   assembly. Structurally the same defect class, but a different assembly type/failure mode and not
   part of this bug's confirmed root cause — flagging as a **non-blocking follow-up** (worth its own
   ticket), not a reason to hold this PR.
4. **`IOException` message reaches the caller cleanly.** Confirmed by reading
   `TabularTableAssembly.loadColumnSelection`'s catch block (`catch(Exception ex)` →
   `ex.getMessage()` → `wizLoadColumnsError`) — it catches `Exception` generically, so the new
   `IOException` is caught the same way the old NPE was, and `singleLine()` only strips newlines (the
   new message has none), so nothing mangles it before it reaches the `PairingException` wrapper.

No blocking issues found in the diff. Regression tests are real (fixer/tester both independently
demonstrated fail-before/pass-after on `WorksheetAgentControllerTest`'s two new tests, and
`ServerFileQueryEmptySheetTest`'s `assertInstanceOf(IOException.class, ...)` cannot pass pre-fix).
Scope is minimal — only the two confirmed root-cause files plus their tests; `addBoundTable`/
`addLogicalModelTable` correctly untouched.

### CI / merge gate

Per this workflow's rule, a StyleBI-side PR must not be merged until its `build` check is actually
green. As of this review, `gh pr view 5268 --json statusCheckRollup` shows:

- Semgrep Scan / Semgrep OSS / claude-review / license/cla — all SUCCESS
- **`build` — FAILURE** (completed, not pending)

The failure (`gh run view` on the build job) is a single test error, unrelated to any file this PR
touches:

```
DashboardRegistryOrgLifecycleTest.rename_copyOrganizationInternal_dashboardFilesRelocatedByDataSpaceRename_contentNotRewritten
  NullPointer: Cannot invoke "inetsoft.sree.web.dashboard.VSDashboard.getViewsheet()" because "dashboard" is null
```

This diff only touches `ServerFileQuery.java`, `WorksheetAgentController.java`, and their two test
files — nothing related to dashboard registry/org-lifecycle. The four other PRs from this same recent
commit history (#5264, #5265, #5266, #5267) all built green, which points at this being a flaky/
order-dependent test rather than something this PR's change caused — but per policy that's a
hypothesis for a human to confirm by re-running the `build` workflow, not something I'm treating as
already cleared. **This PR should not be merged until `build` is re-run and comes back green.**

Merging is a human decision per this repo's self-review restriction (this run's own agents authored
this PR end to end) — I am not approving or merging.

jshobe-inetsoft (changes requested):
Claude's automated review found no correctness/security/backward-compatibility issues in the code changes themselves. However, the `build` check is currently a genuine FAILURE (not pending, not a claude-review-only error):

```
DashboardRegistryOrgLifecycleTest.rename_copyOrganizationInternal_dashboardFilesRelocatedByDataSpaceRename_contentNotRewritten
  NullPointer: Cannot invoke "inetsoft.sree.web.dashboard.VSDashboard.getViewsheet()" because "dashboard" is null
```

This test failure is unrelated to the files this PR touches (`ServerFileQuery.java`, `WorksheetAgentController.java`, and their tests), and the sibling PRs from the same commit history (#5264, #5265, #5266, #5267) all built green, which points at a flaky/order-dependent test rather than something this PR's change caused. Per repo policy, a failing `build` check blocks merge regardless of cause until it's confirmed green — please re-run the `build` workflow (or push a rebase) and confirm it passes before this is merged.

Nelson-cao (approved):

PR #5269, opened 2026-09-15
Bug #76653, add regression test for scheduler DST handling

jkidacki-inetsoft:
## Summary

Redmine Bug #76653 was filed against the closed-source InetSoft Studio product: when the scheduler's own default time zone does not observe DST (e.g. server run with `-Duser.timezone=UTC`) but a scheduled task's configured time zone does, the computed fire time was off by exactly the DST delta (1 hour for US zones), because `TimeCondition`'s scheduling computation converted the task's time into the server's zone using `TimeZone.getRawOffset()` (standard-time offset only, DST-blind).

I checked whether this reproduces here, and it does not. StyleBI's `TimeCondition.getRetryTime()` no longer has this defect - since the Bug #70179 fix (commit `508e32d`), it builds its `Calendar` directly in the task's own configured `TimeZone` (`Calendar.getInstance(getTimeZone())`) instead of converting into the server's zone via a manually computed offset. Because the `Calendar`/`TimeZone` machinery itself applies the correct DST rule for that zone when converting fields to millis, there's no separate raw-offset step left to be DST-blind.

I confirmed this by writing the same DST regression scenario used to verify the fix on the closed-source side (server zone = GMT with no DST, task zone = America/New_York with DST) against this repo's current `main` with **no production code changes**, and it passed immediately.

This PR adds that test as a permanent regression guard, since the existing `TimeConditionTest` suite doesn't otherwise cover this scenario.

## Root Cause

N/A for this repo - see above. (`getHourOffset()`/`getTimeZoneOffset()` still use `getRawOffset()`, but those are only used for the legacy `ajax=true` XML serialization path in `writeAttributes()`, which is a separate, narrower concern from what #76653 reported and is intentionally out of scope for this PR.)

## Fix

No production code change. Added `TimeConditionDstOffsetTest` in `core/src/test/java/inetsoft/sree/schedule/`.

## Test Plan

- `./mvnw -pl core test -Dtest=TimeConditionDstOffsetTest` - passes.
- `./mvnw -pl core test -Dtest=TimeConditionTest,TimeConditionDstOffsetTest` - full existing suite plus the new test, all green (86 run, 0 failures, 5 pre-existing `@Disabled` known-bug tests skipped).

Related to Redmine #76653 (closed-source InetSoft Studio).


jshobe-inetsoft (approved):
Claude's automated review found no functional, security, or backward-compatibility issues. Test-only change, all checks green, mergeable.

PR #5270, opened 2026-09-15
Bug #76687: add run-tasks/stop-tasks endpoints under /api/wiz/v1/admin/schedule

lizhepu:
## Root cause

admin-chat's schedule-task tools (`preview_schedule_task_changes`/`apply_schedule_task_changes`) only support `verb: "create"|"delete"` -- there is no way to trigger a task's next run on demand or stop a currently-executing one. StyleBI already has the underlying primitives (`ScheduleService.runScheduledTask`/`stopScheduledTask`, used today by `EMScheduleController`'s `/api/em/schedule/run-tasks`/`stop-tasks`), but admin-chat's Bearer-JWT auth (`wizClient`) cannot reach `EMScheduleController`: `WizServiceAuthenticationFilter`'s `isWizRequest` guard only recognizes `/api/wiz/**` paths or a `wiz_auth` cookie, so a request to `/api/em/schedule/run-tasks` never even gets its Authorization header inspected and fails closed. Reaching the equivalent enterprise Public API (`ScheduleApiController`) would require a second, unrelated api-token credential type. The fix is a new, correctly-authenticated front door under the existing `/api/wiz/v1/admin/schedule/**` namespace, delegating to the same `ScheduleService` primitives -- no new business logic.

## Changes

- `AdminScheduleController`: two new `@PostMapping`s, `run-tasks`/`stop-tasks`, same `@Secured`/`requireSiteAdmin` shape as the controller's existing 4 routes. Request body reuses `TaskListModel` (`{taskNames: string[]}`), the same DTO `EMScheduleController` already uses.
- `AdminScheduleGateway`: thin `runTask`/`stopTask` wrapper methods delegating to `ScheduleService.runScheduledTask`/`stopScheduledTask`, keeping the controller's existing "always go through the gateway" shape.
- New `ScheduleActionOutcome`/`ScheduleActionResult`: a typed per-task result (`{taskName, status, error}`) instead of mirroring `EMScheduleController`'s untyped, `Catalog`-localized `MessageException` string. `scheduler-not-running` and `task-disabled` are checked up front (a new `ScheduleService.isSchedulerReady()` plus the existing task/enabled lookup) so `status` is a field-named, locale-independent discriminator a caller can branch on, not a message it would have to pattern-match.
- `stop` deliberately has no precondition on the task's current run status: `ScheduleTaskStatus.lastRunStatus` is a `Catalog`-localized string (`catalog.getString("Running")`), not a stable enum, so a server-side check here would silently stop matching on non-English-locale deployments. An idle-task stop is whatever `ScheduleService`/`stopNow` itself reports.
- `ScheduleService.stopScheduledTask`: org-qualify the `getScheduleTask` lookup to match `runScheduledTask`'s own org-qualified lookup (an unqualified lookup resolves against the calling thread's ambient org rather than the principal's own org) -- a minor consistency fix found during review, folded in since it's the same method.

## Tests

`core/src/test/java/inetsoft/web/admin/ai/schedule/{AdminScheduleControllerTest,AdminScheduleGatewayTest}.java` -- 12 new cases covering the new routes' delegation/forbidden-gate, and the started/stopped/scheduler-not-running/task-disabled/failed outcomes.

```
cd core && mvnw.cmd -pl core -o -Dtest=AdminScheduleControllerTest,AdminScheduleGatewayTest,EMScheduleControllerTest test
# Tests run: 53, Failures: 0, Errors: 0, Skipped: 0 -- BUILD SUCCESS
```

Also confirmed a full offline `compile` of the `core` module succeeds with no new warnings/errors.

## Companion PR

Plugin-side wiring (`plugin/admin`) that consumes these two new endpoints is a separate PR against `inetsoft-technology/stylebi-wiz`.

Part of a `/debug-workflow` batch for Redmine bug 76687. Not merged -- holding for review and explicit go-ahead.


lizhepu:
**Review verdict: APPROVE (no blocking findings).** Posted as a comment, not a formal GitHub review, because this PR and the reviewing account are the same GitHub identity (`lizhepu`) and GitHub rejects self-approval (`Can not approve your own pull request`).

Checked against docs/teams/2026-09-15-bug-76687's diagnosis/refutation/fix writeups and the diff directly:

- New `run-tasks`/`stop-tasks` routes sit under the same `/api/wiz/v1/admin/schedule/**` base path as the existing 4 routes — not a repeat of the diagnosed unreachability bug.
- `@Secured`/`requireSiteAdmin` matches the existing 4 routes byte-for-byte — no looser check on this admin-sensitive surface.
- The up-front scheduler-ready/task-disabled precheck in `AdminScheduleGateway` handles the check-then-act race safely: a race between precheck and the actual call falls into the generic `"failed"` bucket via the underlying method's own throw, not a crash or a false "started"/"stopped".
- `ScheduleService.isSchedulerReady()` is purely additive — no existing caller touched.
- `TaskListModel` reuse matches `EMScheduleController`'s existing shape and the plugin PR's described request body.
- `stopScheduledTask`'s org-qualification fix is a no-op for its two existing callers (`EMScheduleController`, portal `ScheduleController` — both resolve org off the same request-bound principal already) while fixing the new gateway caller's real bug; corroborated by the pre-existing `EMScheduleControllerTest` suite still passing unmodified (53/53).
- 12 new tests cover all 5 claimed outcomes (started/stopped/scheduler-not-running/task-disabled/failed) plus the forbidden gate.

Full notes: `docs/teams/2026-09-15-bug-76687/bug-76687/04-review-stylebi-r1.md`.

**Not merging** — held for the lead per this batch's instructions, and a formal GitHub approve/merge from this seat isn't possible anyway due to the self-authorship restriction above; the lead or a different account will need to record the formal approval if the repo's branch protection requires one.

PR #5274, opened 2026-09-15
Bug #76658, add the impersonated identity to the tabular query cache key

jshobe-inetsoft:
## Summary

A REST data source configured with `authType=KERBEROS` and `constrainedDelegation=true` issues its HTTP request while impersonating an identity derived from the calling user, and the remote service applies its own per-user authorization. The tabular query cache key did not capture that identity, so one user's result was served to every other user from the process-wide cache.

## Root Cause

`KerberosAuthenticator.authenticateRequest()` resolves the identity to impersonate from `ThreadContext.getContextPrincipal()`: for `impersonationType=PRINCIPAL` it is the principal's name, and for `PROPERTY` it is one of the principal's properties. The credential is then obtained via S4U2 (`ExtendedGSSCredential.impersonate`), so two different users running the identical query legitimately receive different data from the endpoint.

The key built by `XHandler.getQueryKey` and the override in `TabularHandler.getQueryKey` is composed of the organization id, the data source full name, the query XML, the query variable values and the resolved tabular data source variable values. None of those vary with the impersonated identity. The cache read (`XHandler.isQueryCached`), the write-back (`XSessionManager.DataCacheResult.getCacheKey`) and the eviction (`XHandler.removeQueryCache`) all derive from that one method, and the `DataCache` is a field of the `XSessionManager` singleton, so the entry is shared process wide. Whichever user ran the query first populated the cache, and everyone else was served that user's data until the entry expired (`query.cache.timeout`, default 600000 ms).

Not affected, and confirmed independently:

- `impersonationType=STATIC` — a fixed identity, so the result is not user dependent.
- Any other `authType` (NONE/BASIC/OAUTH/TWO_STEP) — credentials are data-source level.
- Per-user additional connections — `TabularHandler.execute()` resolves the data source before the cache check and `XDataSource.getFullName()` returns the child connection's own name, so those already produce distinct keys.
- JDBC — `JDBCHandler` has its own key rooted at `JDBCDataSource.getIdentity()`, and VPM injects `_USER_`/`_ROLES_`/`_GROUPS_`.
- `AssetDataCache` — its `DataKey` already includes `box.getUser()`.

`KerberosAuthenticator` is the only place in the whole `connectors/` tree where a tabular query's *result* can depend on the calling principal; the only other principal read is `AbstractRestRuntime:44`, which propagates the thread context.

## Fix

Added `TabularDataSource.getCacheDiscriminator()`, which returns `null` by default, and appended it to the key in `TabularHandler.getQueryKey` when it is non-null. Core cannot inspect `AbstractRestDataSource` because connectors load in a separate classloader, hence the hook. Every existing data source produces a byte-identical key, so no existing cache behaviour changes.

`AbstractRestDataSource` overrides it to return the impersonated identity, but only when constrained delegation is enabled and the impersonation type is `PRINCIPAL` or `PROPERTY`. The parts of the discriminator are length-prefixed so that a null, an empty string and a value containing the separator cannot collide. The calling principal's name is included alongside the resolved identity, because the identity can be null (the principal does not carry the property) and because for `PROPERTY` the property name is read before the data source variables are substituted, so a `$(var)` property name would not resolve at key time; including the principal keeps the key user-distinct in both cases.

The identity derivation moved out of `KerberosAuthenticator` into `AbstractRestDataSource.getImpersonatedIdentity()` so that the cache key and the identity actually impersonated cannot drift apart. It is read from the thread context rather than from the user the key is built for, because the thread context principal is what gets impersonated: `AbstractRestRuntime.runQuery` captures it into a `GroupedThread` when it submits the request, on the same thread that computed the key. Should the two differ on an eviction, the entry is not removed early and expires on its own, which is preferable to a key that does not match the result stored under it. That trade-off is documented in the method.

As a side effect, a null context principal now yields a null identity instead of an NPE, and a non-`SRPrincipal` principal under `PROPERTY` yields null instead of a `ClassCastException`.

## Test Plan

Automated:

- `./mvnw clean install -pl core,connectors/inetsoft-rest -am -DskipTests` — BUILD SUCCESS.
- `./mvnw test -pl core,connectors/inetsoft-rest` — core 4921 tests and inetsoft-rest 120 tests, 0 failures, 0 errors.
- New regression tests. `TabularHandlerTest` gains three: the discriminator distinguishes users; a null discriminator leaves the key unchanged (asserted as a prefix, so the pre-fix key is provably intact); users resolving to the same identity still share a key. `KerberosCacheDiscriminatorTest` is new, with eleven covering non-Kerberos auth, Kerberos without constrained delegation, STATIC, PRINCIPAL, PROPERTY including a missing property value and a principal that carries no properties, discriminator stability for one user, and `getImpersonatedIdentity()` per impersonation type plus the null-principal and unset-type cases.
- Both new test classes were checked by mutation: reverting `getCacheDiscriminator()` to return null fails 3 of the 11 REST tests, and disabling the append in `TabularHandler` fails 2 of the 5 core tests — in each case exactly the tests that assert the fixed behaviour.

Manual, against a running server with a REST JSON data source pointed at a local endpoint that counts every HTTP request:

- Create query: 1 request (metadata).
- Live Data View (1st): 1 request, 3 rows returned.
- Meta -> Live Data View (2nd): 0 requests, same rows, served from cache.
- Meta -> Live Data View (3rd): 0 requests, same rows, cache still valid.

The zero-request repeats confirm the key is unchanged for a non-Kerberos source. Also verified that the data source dialog renders the same five fields with Kerberos selected (Service Principal Name, Use Constrained Delegation, Configuration Service Name, Impersonation Type, Principal to Impersonate) and no new ones, that the persisted XML gains no new elements, and that the server log has no ERROR lines.

Not verified: the Kerberos leg itself. Confirming that two users with `constrainedDelegation=true` and `impersonationType=PRINCIPAL` or `PROPERTY` now get separate cache entries requires a KDC with S4U2 constrained delegation, which is not available in the local test environment; a container is not a substitute, since S4U2Self/S4U2Proxy is an Active Directory feature that MIT Kerberos does not exercise on this path. That property rests on the analysis: `getQueryKey` and `KerberosAuthenticator` now call the same `getImpersonatedIdentity()` on the same thread, so the key cannot disagree with the identity impersonated.

Fixes Redmine #76658.


jshobe-inetsoft:
Thanks for the review. Responses to the three items, in order.

**1. `GSSManager.createName(null, ...)` — confirmed, it throws a checked `GSSException`, so the null path is caught.**

I ran the exact call against the JDK's JGSS provider on the build JDK (Temurin 21.0.11):

```
GSSManager.getInstance().createName((String) null, GSSName.NT_USER_NAME)
→ sun.security.jgss.GSSExceptionImpl: Cannot import null name
     at sun.security.jgss.GSSNameImpl.<init>(GSSNameImpl.java:165)
     at sun.security.jgss.GSSManagerImpl.createName(GSSManagerImpl.java:110)
```

`GSSExceptionImpl` extends `GSSException`, which is checked. Thrown inside the `PrivilegedExceptionAction`, `Subject.doAs` wraps it in a `PrivilegedActionException`, and both types are already in the existing catch clause. So a null impersonated identity reaches `LOG.error("Failed to authenticate to the KDC", e)`, no credentials provider is set on the context, and the request fails exactly the way any other KDC failure does — no unchecked exception escapes from the deeper call site. That matches the claim in the PR description: the pre-fix behaviour on this path was an uncaught `NullPointerException` (or `ClassCastException` under `PROPERTY`), and it is now a logged, contained failure.

The one real cost is that `lc.login()` runs before the failure is detected, which is wasted work on an already-failing path rather than a correctness issue. An early `if(impersonate == null)` guard would skip that login and give a log line naming the actual cause instead of a generic KDC failure, but since the outcome is already correct I left it out to keep the change minimal on a security-sensitive path.

**2. `getCacheDiscriminator()` reading `ThreadContext` rather than the `user` parameter — intentional, keeping it.**

You have both the mechanism and its consequence right. Core cannot inspect `AbstractRestDataSource` because connectors load in a separate classloader, so the discriminator has to be produced by the data source; and the identity that is actually impersonated is the thread context principal. Deriving the key from anything other than the same source that `KerberosAuthenticator` impersonates is precisely what allows the key and the identity to drift apart, which is the defect this PR exists to make structurally impossible. On the eviction path, a mismatch means the entry is not removed early and expires on `query.cache.timeout` instead — a stale-but-not-evicted entry, never a wrong result served to the wrong user. The Javadoc on `getCacheDiscriminator()` documents that tradeoff for whoever touches the dispatch path next.

**3. Duplication between `getImpersonatedIdentity()` and `getCacheDiscriminator()` — leaving as is.**

Agreed it is a readability nit rather than a correctness one, and per the repo's bug-fix guidance on minimal scope I would rather not refactor shared derivation logic on this path inside a fix PR.

No code changes in response to this review.

jkidacki-inetsoft (approved):

PR #5275, opened 2026-09-15
Chart palettes: give a target band the companion of its measure's colour

FigurePaper:
Second slice of the chart palette redesign, after #5188. Adds the **companion
colour** — a soft partner for each categorical palette slot, same hue, receding
toward the surface — and spends it on its first consumer: a chart's target band
stops being an unrelated pale green and becomes the soft version of the colour
the measure it sits on is drawn in.

Base: `epic-74519`. Community-only.

## What a user sees

A target band on a modern chart is born the soft version of its measure's
colour instead of `#e4f2e9`. The Edit Target dialog shows that colour, the
author can change it, and the chart draws whatever the dialog shows.

Charts coloured by a dimension keep the classic green: one band spans many
series colours there, so no single measure colour exists to soften. Classic
(non-modern) charts are untouched, and so is every target that already exists.

## How it is built

| | |
|---|---|
| `OKLab` | sRGB↔OKLab/OKLCH, gamut-mapped by reducing chroma |
| `CategoricalColorFrame.getCompanionColor` | the derivation rule |
| `Modern-soft` / `Modern Dark-soft` | authored companions, not selectable |
| `VSChartPaletteDefaults.companionColor` | authored first, derived second |
| `ChartPropertyService.seedCompanionBandFill` | the one consumer |

Companions are **derived by rule**, with the authored palettes as an override —
so a deployment's own palette gets companions too, rather than the feature
working only for the sets we ship. A drift guard asserts the authored hexes
still equal the rule's output.

The companion is a **seeded default, not a render-time substitution**. The
render path is not touched by this PR. The dialog exposes the band fill as an
editable swatch, so substituting a different colour at render would make that
swatch disagree with the canvas — the design-time WYSIWYG break we already hit
with `VSTitleChromeDefaults`. Seeding also means no provenance flag, no new
persisted state, and no gate: the target is born holding the companion and the
author owns it from then on.

An earlier revision of this branch did substitute at render, and carried a
`bandFillUserSet` flag, an XML attribute and a `GraphGenerator` gate to make it
safe. All of that is gone. The design doc records why, in *The rework*.

## Notes for review

- **`OKLab` is the only new subsystem** and is pure maths — no dependency, no
  state. The matrices are the published Ottosson constants; the tests pin them
  against reference vectors and against our own authored colour set.
- **The `-soft` suffix is reserved.** It is filtered inside `getPaletteNames`
  so the reservation lives in one place; resolution by name stays unfiltered,
  because filtering that would break rendering. A deployment palette genuinely
  named `X-soft` would vanish from its own picker — accepted, and noted in the
  design.
- **The companion lookup is memoized** on the same CSS-timestamp stamp as the
  base palettes and guarded like `getPaletteSafely`, so a malformed customer
  `format.css` degrades rather than NPE-ing a chart render.
- The design and plan documents are included; the source handoff they derive
  from was never committed and is recovered in the design's *Corrections*.
- The diffstat reads 2,782 insertions, but 1,800 of those are the two
  documents. The code is roughly 980 lines including tests.

## Verification

- `core` suite green; the companion/palette/target sweep is 376 tests, 0 failures.
- Community `clean install -DskipTests`, all 46 modules: BUILD SUCCESS. The
  `-Pcommunity,enterprise` build fails on the pre-existing
  `License.Builder.formLicensed` skew, unrelated to this work and the same skew
  #5188 hit.
- `palette-dialog.spec.ts` 9/9; neither `-soft` name appears in the dropdown.
- Manually verified in the browser: single-band, dimension-bound, and multi-band
  targets, plus author-set colours surviving a reload.


FigurePaper:
On the unclamped lightness in `getCompanionColor` — leaving it as-is.

Only the light-mode branch can exceed L=1: the dark branch requires L >= 0.50 before subtracting 0.26, so it cannot reach below 0. Across the 14,007 sRGB colours that do push light-mode L above 1, clamping to [0,1] gives byte-identical output — the chroma bisection converges to zero either way and the channel conversion clamps to [0,255].

The genuine rough edge is a different one: a near-white series colour gets a near-white companion. A clamp would not change that, since it needs the rule to reverse direction rather than saturate.

The two inline comments are both addressed in code.

FigurePaper:
Added a light-end case: the rule lifted lightness with no ceiling, so any base above L≈0.82 companioned to white and rendered invisible. Mirrors the existing anchor case for the darkest member, and generalizes the hand-tuning already applied to Amber.

jshobe-inetsoft (approved):

PR #5277, opened 2026-09-15
Bug #76625, stop publishing the calc-field measure clone into the shared sandbox

ngovus-inetsoft:
## Problem

`AbstractSelectionVSAQuery.checkAndRunCalcFieldMeasureQuery()` clones the viewsheet so it can add a temp crosstab and clear selections, then installed that clone with `box.setViewsheet(clone, false)`, restoring it in a `finally`. `setViewsheet()` writes `ViewsheetSandbox.vs` — a single shared field — so the clone was visible to **every** thread.

Nothing serializes two top-level requests on one sandbox: `VSSelectionService.applySelection` holds only a shared **read** lock, and the STOMP inbound channel is a thread pool with no per-session ordering. Two overlapping requests could therefore each capture the other's clone as the "original" to restore.

This is not theoretical — a two-thread probe driven through the old code path reproduced the leak (thread B's clone became visible to thread A). And because `setViewsheet()` also moves action listeners, an interleaved restore could strand the sandbox on a throwaway clone for the rest of the session, with the real worksheet's `metarep` listener permanently dropped.

The same idiom in the export paths produced null table lenses (NPE) and apparent hangs — bug #76576, whose latest fix landed in 6bb8988e32.

## Fix

Replace the swap with `ViewsheetSandbox.createTemporaryBox()` / `releaseTemporaryBox()`.

`createTemporaryBox()` returns a **shallow copy** of the sandbox — the same idiom as the existing `getMVDisabledBox()` — that differs only in which viewsheet it reports. Id, mode, user, asset entry, variable table, MV state, the lock, `metarep`, `wbox` and the caches all stay shared, so the nested query still locks the real sandbox and still executes through the original `AssetQuerySandbox` over the original worksheet, exactly as before. The clone, the `resetSelection()` loop and `refreshRuntimeConditionList()` are unchanged.

The functional diff is 6 lines of semantics; the rest is the new helper and its Javadoc.

## Rejected alternatives (recorded so they aren't retried)

- **A lock.** The nested query releases the sandbox lock mid-call in `VSAQuery.getDataWithoutSandboxLock()`, whose contract is that callers hold no monitor a sandbox-lock holder can need. A thread parked in `restoreLocks()` while holding a swap lock deadlocks against a thread holding the write lock and waiting for that lock. #76576's `synchronized(rvs)` works only because exports don't run under the write lock.
- **A thread-local viewsheet override.** Implemented and abandoned. Making `getViewsheet()` thread-dependent splits the brain against everything sandbox-global keyed off it: `qmgrs` + `shrink()` (which cancels QueryManagers whose assembly is absent), the `bmap` child-sandbox cache, `metarep.metamap`, and the separate worker pools in `AssetDataCache`, `JoinQuery` and `ConcatenatedQuery` that each hand-propagate their own chosen thread state.

## Caller constraints (in the Javadoc)

A shallow copy shares object *references* while non-final fields are snapshots, and the sandbox's collaborators are bound to the original object. Three limits, none of which affect this call site:

1. **Scriptables for temp-only assemblies don't resolve.** `ViewsheetScope` is bound to the real sandbox, so its fallback `box.getViewsheet().containsAssembly(name)` misses the temp crosstab. Harmless here because the crosstab's column, formula and group values are all already-resolved literals (`getFormula()` returns `formulaValue.getRuntimeValue(true)`), so `executeDynamicValue()` never needs the scriptable.
2. **`disposed` is a snapshot.** Keep the copy short-lived.
3. **Events fired by the clone are handled against the real sheet.** Safe here because the crosstab is added and removed with `fireEvent=false`.

## Testing

- New `ViewsheetSandboxTemporaryBoxTest` (5 tests): the shared sandbox never reports the clone; two concurrent temp boxes stay independent; the copy shares id / `AssetQuerySandbox` / entry / mode / user / lock; listeners land on the clone and the real sheet's are untouched; `releaseTemporaryBox(null)` is a no-op.
- Full `core` suite green.

## Not verified

**No end-to-end check of measure values.** Standing up a viewsheet with a calc-field selection measure needs an imported asset bundle — none of the existing test zips has one, and building one is the disproportionate scaffolding this repo also declined for the sibling bug #76614. The unit tests pin the isolation contract, not the numbers. Since this change preserves the query logic verbatim, the risk is low, but please confirm measure bars against a real viewsheet — **including one inside an embedded viewsheet**, since `createTemporaryBox()` deliberately drops the `updateRootSandboxMap()` step `setViewsheet()` performed (that step mutates the *parent* viewsheet, which is the shared-state bug being fixed).


jshobe-inetsoft (approved):

PR #5280, opened 2026-09-16
Bug #76691 (WBT-009): disclose add_table's resolved assembly name

fredaan:
## Summary

`add_table` can silently create its new assembly under a different name than the caller
requested. All four datasource-backed forms (bound table, logical-model entity, endpoint/suffix,
queryParams) resolve name collisions through `AssetUtil.getNextName` (e.g. a request for
`"ORDERS"` lands as `"ORDERS1"`), but `WorksheetAgentController.edit` returned `void`, so there
was no wire response for a caller to read the resolved name from -- the only way to discover the
substitution was a later `read_worksheet_model` call, or a confusing "table not found" from some
other tool trying to use the name it originally asked for.

This changes `edit`'s response contract **for `add_table` only**:

- New `EditResponse(String assemblyName)` record.
- `edit`: `void` -> `ResponseEntity<EditResponse>`. Returns `200` with the resolved assembly
  name for `add_table`; every other op (all ~30 others sharing this endpoint) keeps getting
  `204 No Content`, matching today's empty-body behavior exactly.
- `addBoundTable`, `addLogicalModelTable`, `addTabularTable`, `addQueryParamsTable` (all four
  `add_table` branches) now report their resolved name via a new out-param threaded through a
  private 4-arg `editOp` overload. The existing public 3-arg `editOp` (the one other in-process
  caller, `WorksheetScriptService`) is now a thin delegator passing `null` for that out-param, so
  that call site is unaffected.

Companion plugin-side PR (reads this new field and surfaces it to the caller as a `note`):
inetsoft-technology/stylebi-wiz#TBD (opening immediately after this one) -- **that PR's new
behavior is inert until this one merges and ships**, since it depends on this response field
existing.

Redmine #76691 / WBT-009.

## Test plan

- [x] `./mvnw -pl core test -Dtest=WorksheetAgentControllerTest` -- 185/185 pass, including 2 new
      regression tests (`addBoundTableEditResponseDisclosesResolvedNameOnCollision`,
      `addQueryParamsTableEditResponseDisclosesResolvedNameOnCollision`)
- [x] Reverted the fix, confirmed both new tests fail (`NoSuchMethodError` against the old `void
      edit` signature), restored, confirmed 185/185 pass again
- [x] Sibling suites unaffected: `WorksheetScriptServiceTest` 31/31 (the other in-process `editOp`
      caller), `WorksheetEditServiceTest` 26/26


fredaan:
# WBT-009 Review r1

Adversarial review of both PRs, against the CONFIRMED cause (not just the reported symptom),
per the review-workflow adversarial pass + the lead's four extra checks.

- Java/`stylebi` PR #5280 (base `stylebi-wiz-main-rebase`, commit `cff81742f`):
  https://github.com/inetsoft-technology/stylebi/pull/5280
- Plugin/`stylebi-wiz` PR #2561 (base `main`, commit `593da0ce9`):
  https://github.com/inetsoft-technology/stylebi-wiz/pull/2561

## Diff review

Read `gh pr diff 5280` (376 lines) and `gh pr diff 2561` (81 lines) in full, plus the actual
`WorksheetAgentController.java` and `WorksheetScriptService.java` file contents at the PR head
commit (not just the diff hunks), and the full `add_table` tool handler in `worksheetTools.ts`
at the PR head commit.

### 1. Does the new 4-arg editOp leave the other ~26 non-add_table ops' dispatch byte-for-byte unchanged?

Confirmed by reading the full private `editOp` method body (lines 310-545 of the PR-head file,
not just the 4 call-site diff hunks). Every other op branch (`add_named_group` in both its
datasource-scoped and plain forms, `set_variable_values`, `convert_to_embedded`,
`edit_sql_query`, `update_mirror`, `auto_layout`, `refresh_data`, `insert_column`,
`reorder_concat_subtables`, `add_variable`, and the final generic
`editService.apply(sessionToken, user, editor -> dispatch(editor, req))` fallback that catches
everything else) is untouched. None of them pass `resolvedNameHolder` through, so it stays null
for all of them, and `edit()` returns `ResponseEntity.noContent().build()` for every one. Only
the four `add_table` branches (`addQueryParamsTable`, `addTabularTable`, `addLogicalModelTable`,
`addBoundTable`) were touched, each by exactly one added parameter and one call-site edit. No
other line in `editOp` changed. Confirmed by reading the method myself, not just trusted from
the diff hunks.

### 2. Is WorksheetScriptService's call site to the 3-arg editOp really unaffected?

Read `WorksheetScriptService.java` directly: line 137 calls
`worksheetController.editOp(session.sessionToken(), req, agent)`, the untouched public 3-arg
overload, which the PR turns into a one-line delegator (`editOp(sessionToken, req, user, null)`).
Confirmed unaffected. `WorksheetScriptServiceTest` 31/31 passing (per `04-verify.md`) corroborates
this independently at the test level too.

### 3. Does ResponseEntity.noContent() (204) preserve identical wire behavior to the old bare void for every existing caller?

Checked for any consumer of this endpoint besides `plugin/composer`'s `wizClient`: zero hits for
`agent/worksheet` under `community/web-src` (StyleBI's own Angular frontend), and the only hits
under `wiz-services/src` are an unrelated doc-string import path substring match, not an actual
HTTP call. `02-refute.md` already established none of `worksheetTools.ts`'s 18 call sites to this
same `/edit` endpoint read the response body today. This is a status-code change (plain 200 from
a void Spring MVC handler to explicit 204) for every non-`add_table` op, not byte-for-byte
identical, but since no in-repo caller (plugin or StyleBI's own frontend) inspects this endpoint
status code today, it is a safe, unobserved change in practice. Worth keeping in mind if some
future/external consumer starts checking status codes here, but not a blocking finding given
today's actual callers.

### 4. Does the plugin's new note collide with the existing WBS-019 row-count note?

Read the full `add_table` tool handler in `worksheetTools.ts`, not just the new diff hunk. The
two notes are structurally mutually exclusive, not just coincidentally non-colliding:
- The WBS-019 row-count note only fires inside `if (!a.datasource) { ... return ... }` (the
  EMBEDDED path), which returns early before the function ever reaches the `renamedNote`-bearing
  final `return` statement.
- `renamedNote` can only be truthy if `editResponse.assemblyName` is set, and on the Java side
  `resolvedNameHolder` is only populated by the four datasource-backed branches
  (`addBoundTable`/`addLogicalModelTable`/`addTabularTable`/`addQueryParamsTable`), all of which
  require `req.datasource()` to be non-blank to be dispatched at all (confirmed in `editOp`'s own
  dispatch guards). The EMBEDDED (no-datasource) path never reaches any of the four branches, so
  `assemblyName` is never set for it.

Both gate conditions (`!a.datasource` vs. a datasource-backed branch having populated the
resolved name) are mutually exclusive by construction, and the EMBEDDED branch's early return
makes it doubly impossible in practice. No collision, no silent-drop scenario. This resolves the
review concern cleanly.

### 5. Is resolvedNameHolder[0] = assemblyName really set unconditionally in all four branches?

Confirmed by reading all four branches directly in the PR-head file (`addBoundTable`, around
L622-625; `addLogicalModelTable`, around L720-723; `addTabularTable`, around L900-904;
`addQueryParamsTable`, around L1001-1005): each sets `resolvedNameHolder[0] = assemblyName`
(guarded only by `resolvedNameHolder != null`) immediately after computing `assemblyName` via
`AssetUtil.getNextName`, with no conditional on whether an actual rename occurred. This is safe:
the TS side (`worksheetTools.ts`) itself diffs `editResponse.assemblyName !== a.table`, so an
unconditional set (same value as requested when there was no collision) correctly produces no
note. Confirmed by reading, not assumed.

## Test coverage

Both sides have real regression tests that assert on the HTTP response body itself
(`response.getBody().assemblyName()` on the Java side; `note` content on the plugin side), not
just internal worksheet state -- matching this project's tool-misuse-is-a-plugin-gap standard.
Fail-before/pass-after was already independently reproduced twice (fixer + tester, per `03-fix.md`
/ `04-verify.md`); this review's own reading of the full (not just diffed) source corroborates the
same conclusions rather than contradicting them.

## Scope

No unrelated refactors, no dead code, no unnecessary comments (the added Javadoc/comments explain
non-obvious behavior -- why the 3-arg overload still exists, why the EMBEDDED note and rename
note can't collide -- not restating what the code does). No security concerns (no new input trust
boundary; the resolved name is a server-computed string, not attacker-controlled beyond what
`AssetUtil.getNextName` already produced pre-fix). No CLAUDE.md violations found.

## PR state (checked live, this session)

- #5280: OPEN, MERGEABLE. Checks: `claude-review` SUCCESS, `Semgrep Scan` SUCCESS, `Semgrep OSS`
  SUCCESS, `license/cla` SUCCESS, `build` (Validation Build): IN_PROGRESS -- not green. Per
  this workflow's rule, must NOT be merged until `build` completes and passes.
- #2561: OPEN, MERGEABLE. Only check: `claude-review` SUCCESS. No blocking check, but by design
  (per its own PR description and `03-fix.md`) it is inert/dependent on #5280 landing first --
  do not merge independently of #5280.

## Verdict

No blocking (important/critical) findings on either PR. Both diffs match the confirmed root
cause exactly, with no scope drift, no regression to the other ~26 `editOp` ops, no note
collision, and real response-body-asserting regression tests on both sides.

- #5280: clean on content; blocked from merge only by its own `build` check still being
  IN_PROGRESS -- re-check before merging, do not merge on this review alone.
- #2561: clean on content and CI; blocked from merge only by its documented dependency on #5280
  shipping first, not by anything in this review.

This review does not merge or approve either PR (self-authored-PR review restriction) -- posted
as PR comments only; a human must approve/merge.

Nelson-cao (approved):

PR #5281, opened 2026-09-16
fix(portal-session-pairing): identity-keyed cross-sheet-follow lookup (PSP-028)

larryliang-inetsoft:
## Summary
- `SheetSessionService.setCrossSheetFollow`/`syncToCurrentFocus` looked up a directly-established (D10) session by exact `socketSessionId` equality, but `openEstablishedDirectly` always mints that session with `socketSessionId == null` (it is never paired to one browser tab) -- `socketSessionId.equals(null)` is always false, so the toggle could never find its own session (threw `INVALID_ARGUMENT` unconditionally) and the cross-sheet retarget could never fire either. Both now resolve the caller's own authenticated STOMP-frame `Principal` into an `ownerIdentity` and match on that instead, via a shared `findByIdentityAndFlag` helper (mirrors PSP-027's own `sendToComposerByIdentity` identity-resolution pattern).
- Added `GET /api/wiz/pairing/cross-sheet-follow` so each Angular app's own separate `CrossSheetFollowService` instance (the portal shell and Composer app are distinct bootstraps, each with its own `providedIn: "root"` instance) learns the server's current state on construction -- otherwise enabling the toggle in the portal shell could never reach the Composer app's own instance, whose `reportCurrentFocus` is what actually matters for sending focus reports. Mirrors `WizService`'s existing constructor-time-GET convention for `/api/wiz/pairing/feature`.
- `CrossSheetFollowService.setEnabled` no longer flips its local `enabled` flag until the server confirms success (reads the `CrossSheetFollowResponse` on `/user/commands/wiz/pairing/cross-sheet-follow`, mirroring `ConnectToClaudeComponent`'s own subscribe-then-send pattern for the mint reply) -- matches PSP-025's established "a toggle must never look like it succeeded when it didn't" principle.
- Removed the now-dead `findEstablishedDirectlyBySocket`/`findCrossSheetFollowBySocket` and their stale javadoc (one incorrectly claimed a cross-sheet-follow session "is only ever attached by `syncToCurrentFocus` itself", which was never true of the actual implementation).

## Root cause
`SheetSessionService.openEstablishedDirectly` (the only entry point for a D10 directly-established session) always constructs its `JoinSession` with `socketSessionId = null`, and nothing in the class ever assigns it a non-null value afterward. The two cross-sheet-follow lookups keyed on an exact `socketSessionId` match, which could therefore never succeed for the one session type the feature exists to serve.

## Test plan
- [x] `SheetSessionServiceTest`, `SheetPairingControllerTest` -- 88/88 pass (`./mvnw test -pl core -Dtest=SheetSessionServiceTest,SheetPairingControllerTest`)
- [x] Broader `inetsoft.web.wiz.pairing.**` package -- 187/187 pass
- [x] Deliberate-break-verified: reverted `setCrossSheetFollow` to the original socket-equality shape and reran -- 6 new/changed assertions failed as expected across both test files; restored the fix and reran clean
- [x] `inetsoft.web.wiz.pairing.**,inetsoft.web.wiz.viewsheet.**,inetsoft.web.wiz.binding.**,inetsoft.web.wiz.script.**` -- 1940 tests, 1 failure + 37 errors, all in unrelated worksheet/chart-binding/calc-field classes, matching PSP-027's own documented pre-existing batch-order-dependent static-init flake baseline exactly (not touched by this change)
- [ ] Live browser re-verify -- not performed in this session (no browser-automation tool available); see the fix write-up (`stylebi-wiz` repo, `docs/teams/2026-09-15-test-portal-session-pairing/case-PSP-028/02-fix.md`, committed locally there per that repo's own divergence note) for exact steps


larryliang-inetsoft:
Adversarial review pass complete. The server-side fix and its security story are solid — but there's one real blocking client-side race that this PR's own new code introduces, and it reproduces the exact class of bug this PR exists to fix (a toggle that silently shows the wrong state).

## Verified correct (no action needed)

- **Principal trust story**: `crossSheetFollowViaSocket`/`currentFocusViaSocket` take `Principal owner` resolved by Spring's STOMP messaging infra from the authenticated handshake session (the same mechanism the pre-existing, unmodified `mintViaSocket` already relies on for `ownerKey(owner)`). `WebSocketConfig` registers no custom `HandshakeHandler`, so `Principal` comes from the servlet container's `HttpServletRequest.getUserPrincipal()` at handshake time, not anything in the client's message payload — not spoofable.
- **`findByIdentityAndFlag`**: correctly scoped — `findEstablishedDirectly` still gates on `JoinSession::establishedDirectly`, so a pane-scoped session can never match (charter assertion 11 intact). `ownerIdentity` equality is between two identity-key strings both produced by the same `ownerKey()`/`IdentityID.convertToKey()` path, so exact `.equals` is the right comparison here (this is different from `syncToCurrentFocus`'s separate `PairingUtil.sameLogicalUser` check, which is comparing a stored identity string against a live `Principal` from a different subsystem — that one is untouched by this PR and still correct). "First match wins" for multiple sessions per identity is documented as an accepted, inherited limitation, same tradeoff PSP-027 accepted.
- **New `GET /api/wiz/pairing/cross-sheet-follow`**: takes only `Principal owner`, no identity/user param from the client — cannot be used to probe another user's toggle state. Defaults `false` for both an unauthenticated caller and "no live session," consistent with the STOMP sibling's treatment of the same conditions.
- **No stale call sites**: grepped for every other caller of `setCrossSheetFollow`/`syncToCurrentFocus`/the removed `findEstablishedDirectlyBySocket`/`findCrossSheetFollowBySocket` — none exist outside this file and its own tests.
- **Deliberate-break re-verified myself**: reverted `findByIdentityAndFlag` to match on `s.socketSessionId()` instead of `s.ownerIdentity()` and reran `SheetSessionServiceTest`/`SheetPairingControllerTest` — 9 failures (session service) + 3 failures (controller), confirming the regression tests actually catch a naive/wrong reimplementation, not just "does it compile." Restored the real fix and confirmed clean again.
- **Test counts confirmed independently**: `SheetSessionServiceTest` 67/67, `SheetPairingControllerTest` 21/21 (88/88 total). Full `inetsoft.web.wiz.pairing.**` package: 187/187 (67+21+4+4+5+13+24+33+11+2+3, matches exactly). The broader run pulls in unrelated `binding`/`script` packages with the same pre-existing batch-order flake signature (`ChartBindingMutatorTest`, `CalcFieldServiceTest`, `ScriptReadServiceTest`, `ScriptEditServiceTest`, `SecondaryYPostProcessingPipelineTest`) already documented by PSP-027 — none of it is in the pairing package.
- **Clean compile**: `mvnw compile -pl core -o -q` succeeds.

## Blocking issue

**`cross-sheet-follow.service.ts` `setEnabled()` is not safe against overlapping/rapid calls** (`web/projects/portal/src/app/composer/gui/wiz/services/cross-sheet-follow.service.ts:118-146`)

`StompClientChannel.subscribe` (`web/projects/shared/stomp/stomp-client-channel.ts:56-83`) multiplexes every subscriber to a destination through one shared `Subject` per topic — it is not a per-call, single-reply channel. Every call to `setEnabled` adds a fresh subscription to `/user/commands/wiz/pairing/cross-sheet-follow` with no request-correlation id in the payload/reply, and `portal-agent-notice.component.html:31-32` has no `[disabled]` guard on the checkbox while a request is in flight (confirmed — `onCrossSheetFollowChange` calls `setEnabled` directly, `portal-agent-notice.component.ts:91-92`).

Concrete failure: user toggles on, then off again before the first server reply returns. Two live subscriptions now exist on the same destination. The first reply (answering "on") fires **both** callbacks (each only checks `body.ok`, with no way to tell which request it answers) — both set `this.enabled = true` and both unsubscribe. The second reply (answering "off") has no subscriber left and is silently dropped. The client is now stuck showing "enabled" — the opposite of the user's last action and the server's actual final state — until a full page reload re-seeds it from the constructor's GET.

This isn't inherited from the `ConnectToClaudeComponent` pattern this code says it mirrors — that flow has a `loading`-gated button preventing overlapping mints in the first place. This service has no equivalent guard, so the race is new, introduced by this PR. Given the entire point of this fix (and its own PSP-025 precedent citation) is "a toggle must never look like it succeeded when it didn't," a new path that leaves the toggle silently wrong until reload is a regression of that same principle, not a nitpick.

Fix: either gate re-entrant `setEnabled` calls while one is pending (disable the checkbox in `portal-agent-notice.component.ts`/`.html` while in flight, mirroring `ConnectToClaudeComponent`'s `loading` flag), or correlate the request/reply (e.g. unsubscribe-then-resubscribe fresh per call, or a request id echoed in the reply) so a stale reply can't be applied to the wrong request.

(The GitHub `claude-review` bot on this PR independently found and described the same issue in detail — I verified its claim by reading `StompClientChannel.subscribe` and the component template/controller myself; it holds.)

## Non-blocking notes

- No `.spec.ts` exists for `CrossSheetFollowService` itself, and `portal-agent-notice.component.spec.ts` mocks the service via `useValue`, so neither the constructor's new `HttpClient` seed nor `setEnabled`'s reply handling (including the race above) is exercised by any test today. Worth a follow-up test once the race is fixed.
- Minor: the client-side race above would be a good thing to add a regression test for, alongside the fix.

Everything else in this PR — the server-side identity-keyed lookup, the new status endpoint, the removal of dead socket-keyed code, the javadoc corrections, and the Java-side test coverage — is correct and ready. Once the client-side race above is addressed, this should be good to merge.

PR #5283, opened 2026-09-16
fix(wiz): refuse selection ops on a column-less assembly, fix deselect vacuous-truth gap

Nelson-cao:
## Problem

`SelectionRuntimeService`'s `setSelection`/`clearSelection`/`selectSubtree` never checked
whether a `SelectionList`/`SelectionTree` assembly actually had a column bound.
`SelectionListVSAssembly.getConditionList()`/`getSelection()` short-circuit to null/false
when the column is unbound, so on an unbound assembly every one of these endpoints reported
success while writing no filter at all -- `set_selection` would say `"selected 1 value(s)"`
immediately followed by `get_condition` showing `conditionCount:0`.

Separately, `deselectTargets`'s shortest-prefix search treated an empty (or simply
non-overlapping) `remaining` selection as vacuous proof that a requested path was safe to
target, manufacturing a deselect target -- and a nonzero `"deselected"` count -- for values
that were never actually selected. This produced the `deselected:1` / `clearedCount:0`
contradiction from the bug report, and misfires independently of the column-binding issue
on any mismatched deselect value against a perfectly healthy, non-empty selection.

## Fix

- `setSelection`/`clearSelection`/`selectSubtree` now throw `IllegalArgumentException` naming
  the assembly when a `SelectionListVSAssembly` has no `getDataRef()` or a
  `SelectionTreeVSAssembly` has no non-null `getDataRefs()`, before any of the downstream code
  runs. `TimeSliderVSAssembly` is untouched -- its column binding is a different mechanism.
- `deselectTargets` now skips any requested path that `current` never actually overlapped
  with, and `setSelection`'s deselect branch reports the count of values that genuinely
  matched `current`, not the raw requested count.

## Test plan

- [x] `SelectionRuntimeServiceTest`: 67/67 passing (58 baseline + 9 new), via
      `./mvnw -q -pl core -am -Dtest=SelectionRuntimeServiceTest test`
- [x] Full `inetsoft.web.wiz.viewsheet` package: 953/953 passing, no regressions


Nelson-cao:
## Review (bug-76701 debug-workflow)

Base branch confirmed corrected (`stylebi-wiz-main-rebase`, 2 files, +203/-21). Diff read against
the confirmed diagnosis/refute conclusion, not just the reported symptom — see
`docs/teams/2026-09-16-bug-76701-selection-filter-freeze/bug-76701/05-review-r1.md` (stylebi-wiz
repo) for full detail. Summary:

**What's right:** `requireBoundColumn` keys on the real mechanism (`getDataRef()`/`getDataRefs()`
null), covers both ID-mode and non-ID-mode trees (both use `getDataRefs()`), correctly leaves
`TimeSliderVSAssembly` alone, and `deselectTargets`'s `everSelected` fix generalizes to the
broader non-empty-`current` case, not just the one live repro's fully-empty case. Tests assert
real behavior, not tautologies; `Calendar` exclusion is correctly out of scope.

**Two blocking findings — both independently verified against source here, originally flagged by
this PR's own `claude-review` check:**

1. **`selectSubtree` (~line 380):** `requireBoundColumn` runs *before* the existing
   `instanceof SelectionTreeVSAssembly` check. Calling `selectSubtree` on a column-less
   `SelectionListVSAssembly` (wrong tool for this assembly type) now reports "no column bound"
   instead of the correct, pre-existing "subtrees only exist on a selection tree" — misdirects the
   caller toward binding a column when the real problem is calling the wrong tool. Fix: reorder,
   type check first. Confirmed this doesn't break `refusesSelectSubtreeOnAColumnlessTree` (fixture
   is already a tree).

2. **`setSelection` (~line 150):** `requireBoundColumn` fires unconditionally, even for a
   legitimate sortOrder-only or singleSelect-only call (no `values`/`deselect`) — neither of which
   touches the actually-broken `getConditionList()`/`getSelection()` path. This blocks a
   plausible legitimate call with a misleading message about select/deselect/clear the caller
   never asked to do. Fix: gate on `values != null || hasDeselect`. Confirmed this doesn't break
   any of the 5 new guard tests (all pass `values` explicitly). Note: no existing test would catch
   this regression today (fixtures default to a bound column), so the fix-round should add one.

**Non-blocking:** `deselected`'s count (~line 280) recomputes `everSelected` for the whole
`deselect` list a second time, duplicating what `deselectTargets` already computed per-path
internally. Same result, just redundant — worth a follow-up, not a fix-round on its own.

**Merge gate:** `Semgrep Scan`/`Semgrep OSS`/`claude-review`/`license/cla` all pass, no conflicts
(`mergeable: MERGEABLE`). `build` (Validation Build) was still `in_progress` after 20+ minutes of
polling at review time — not yet green, which alone would block merge for this StyleBI-side PR
regardless of the two findings above.

Not approving via `gh pr review` since this session's own identity authored this PR (self-approve
is rejected) — posting as a plain comment instead. Would approve once both findings are fixed,
re-verified, and `build` is green.

Nelson-cao:
## Review round 2 (bug-76701 debug-workflow)

Verified both round-1 blocking findings against the actual diff (`git diff 9a265a4fd..effae1539`),
not the fix writeup alone.

**Finding 1 (selectSubtree ordering) — fixed, verified.** `requireBoundColumn` now runs after the
`instanceof SelectionTreeVSAssembly` check. Confirmed nothing between the old and new position
depended on `requireBoundColumn` having run first. Confirmed the existing
`refusesSelectSubtreeOnAColumnlessTree` test is unaffected (its fixture is already a real tree, so
the type check passes through and the bound-column check still fires second).

**Finding 2 (setSelection guard scope) — fixed, verified.** Guard now gated on
`values != null || hasDeselect` — correctly fires for a deselect-only call, correctly skips a
sortOrder-only or singleSelect-only call. The two new tests
(`allowsASortOrderOnlyCallOnAColumnlessSelectionList`,
`allowsASingleSelectOnlyCallOnAColumnlessSelectionList`) assert `assertDoesNotThrow` on a
column-less fixture and use values that differ from the fixture defaults (`"desc"` vs `SORT_ASC`,
`true` vs `false`), so they actually exercise the sort/single-select code path behind the guard,
not just a no-op call — real regression coverage, not a placeholder.

**Non-blocking nit #4 (duplicated `everSelected` computation):** left alone, as this review
suggested in round 1. Still fine with that — not re-raising it.

**No new issues found in this round's delta** (2 files, +37/-2, scoped to exactly the two
findings).

**Merge gate:** `Semgrep Scan`/`Semgrep OSS`/`claude-review` all pass on `effae1539`. `build`
(Validation Build) is still in progress — this PR cannot merge until it's green, independent of
this review's verdict.

Posting as a plain comment (not `--approve`) for the same reason as round 1 — this session's own
identity authored the PR. Would approve once `build` is confirmed green.

fredaan (approved):

PR #5286, opened 2026-09-16
fix(76726): exclude password from schedule-task plan hash

lizhepu:
## Summary

- `ScheduleXmlProjection.EXCLUDED_ATTRIBUTES` strips `lastModified`/`path`/`editable`/`removable`
  before a schedule task's XML serialization feeds `ScheduleChangePlanService.hash`, but never
  `password`.
- `ServerPathInfo#writeXML`/`EmailInfo#writeXML` re-encrypt the stored `saveToServerFilePaths`/
  email-zip password via `Tool#encryptPassword` on every serialization, and the underlying JCE
  cipher generates a fresh random IV per call, so the `password="..."` ciphertext substring differs
  on every call even though the task itself never changed.
- Net effect: any schedule task carrying a saved server-path or email-zip password got a different
  `planHash` on every `preview_schedule_task_changes`/`apply_schedule_task_changes` call, so a
  delete plan for that task always came back `status: "conflict"`, even on the very first
  preview -> apply attempt.
- Fix: add `"password"` to `EXCLUDED_ATTRIBUTES` — the same idiomatic attribute-name strip already
  used for the other four entries.
- Edge case (deliberate, not a regression): a delete plan's hash can no longer distinguish "task
  unchanged" from "task's password changed between preview and apply." There is no update/edit verb
  in this area (only create/delete), and `ScheduleChangesetApplyService#applyDelete` verifies
  success via `getScheduleTask(taskId) == null`, never by comparing against the captured password
  substring — so this has no effect on delete's actual behavior.

Root cause diagnosed, independently refuted, and fix scope confirmed in
`docs/teams/2026-09-16-bug-76726/` on the `stylebi-wiz` repo.

## Test plan

- [x] `ScheduleXmlProjectionTest#projectionIsStableAcrossPasswordReencryption` — two tasks with the
      identical plaintext password but a genuinely varying `Tool.encryptPassword` return value per
      call now project identically; verified failing on pre-fix code, passing after.
- [x] `ScheduleChangePlanServiceTest#hashIsStableForDeleteWhenPasswordIsReencryptedBetweenCalls` —
      two `resolve()` calls for the same delete plan, with a genuinely varying encrypted-password
      value per call, now produce the same `planHash`; verified failing on pre-fix code, passing
      after.
- [x] Full `inetsoft.web.admin.ai.schedule` package test suite
      (`AdminScheduleControllerTest`, `AdminScheduleGatewayTest`, `ScheduleChangePlanServiceTest`,
      `ScheduleChangesetApplyServiceTest`, `ScheduleXmlProjectionTest`) passes with no regressions.


lizhepu:
## Review round 1 — verdict: approve (posted as comment; GitHub blocks self-approval on this account)

Verified independently (not just re-reading the diagnosis chain):

- **Diff generalizes correctly**: `EXCLUDED_ATTRIBUTES` gets `"password"` — a generic attribute-name
  regex strip via the existing `buildPatterns()` mechanism, not scoped to any repro-specific
  literal (task id/name).
- **Covers both password paths**: independently read `ServerPathInfo.writeXML` and
  `EmailInfo.writeXML` in `community` — both emit the identical `password="..."` attribute name,
  and a repo-wide grep of `sree/schedule` found no other class emitting a `password` attribute for
  an unrelated purpose. One `EXCLUDED_ATTRIBUTES` entry correctly covers both call sites.
- **Regression tests actually test the bug**: reverted the one-line fix in the PR's own worktree,
  reran both new tests — both failed with the expected differing-hash / differing-ciphertext shape.
  Restored the fix, reran — both pass, full `ScheduleXmlProjectionTest`/`ScheduleChangePlanServiceTest`
  suites green.
- **Edge case re-checked**: `ScheduleChangesetApplyService.applyDelete` verifies success via
  `getScheduleTask(taskId) == null`, never by diffing the captured password substring — stripping
  `password` from the hash has no effect on delete's actual behavior.
- No unrelated/drive-by changes.

**Non-blocking nitpick**: the pre-existing class-level javadoc on `ScheduleXmlProjection` (untouched
by this diff, ~lines 34-38) claims excluded attribute names "do not recur ... anywhere in a
condition or action sub-element" — that's now false for `password`, which is exactly why this
one-line fix covers both `ServerPathInfo` and `EmailInfo`. Worth a follow-up doc correction, not
worth blocking this PR over.

Full findings: `docs/teams/2026-09-16-bug-76726/bug-76726/05-review-r1.md` in stylebi-wiz.

Not merging — human confirmation required per workflow convention.

lizhepu:
## Round 2 review (sentinel-value refinement, commit e62d13b12)

**Process note**: this round's dedicated fixer/reviewer subagents both hit session rate limits; the lead finished the code and this review directly — flagging that up front since it's not a fully independent second party the way round 1 was.

**Verdict: Approve**, with 2 non-blocking findings:

1. The javadoc's claim "both writers only ever emit `password` when non-blank" is airtight for `ServerPathInfo`'s plain username/password branch (the one this bug/these tests actually exercise), but not for the secret-store-credential branches in `ServerPathInfo`/`EmailInfo` (gated on key-presence or a different field's non-blankness, not the password value itself) — worth narrowing the doc claim in a follow-up, doesn't affect correctness since the regex normalizes an empty value the same as a non-empty one.
2. The "mirrors ProviderProjection" framing is conceptually right (same problem: a secret that must never enter a hash, but whose presence is meaningful) but mechanically different — Provider builds its projection by hand and never touches real ciphertext; Schedule's fix is a post-hoc regex rewrite of the real `writeXML()` output. Worth a lighter word than "mirrors" in the commit message if this gets amended again, not blocking.

Independently re-verified: regex correctness (execution, not just reasoning), fail-before/pass-after on all 3 new/changed assertions (2 AssertionFailedError + 1 NoSuchFieldError without the change), and the full 5-class sweep (93/93, no regressions). Not merging — awaiting the user's own compile/test confirmation.

PR #5295, opened 2026-09-16
fix(76689): four sort/chartType/secondaryY bugs in the wiz binding read/write path

NeverStopFMH:
## Summary
Four related defects found and root-caused during Redmine #76689's live-verification pass (a batch of 13 findings from a VS-CHART-SHELF test run against the composer-chat MCP tool surface), all in `inetsoft.web.wiz.binding`:

- **VCS-012** — `BindingReadService`'s `sorts` map was keyed by bare column name only; the same column bound to two different chart shelves (e.g. a date grouped by month on `x` and by year on `group`) collapsed to one entry, silently hiding one shelf's sort state. Fixed by requalifying with `[shelf]` only on an actual cross-shelf collision.
- **VCS-004** — `BindingReadService.withTypes()` (called for every measure on a multi-style chart, to report its own chartType) rebuilt `FieldRef` via a backward-compat constructor that predates `label`/`secondaryY` and silently dropped both. Fixed by calling the canonical constructor and threading both through.
- **VCS-014** — `set_chart_field_sort`/`set_chart_field_ranking`'s `sortByField`/`measure` accepted a bare column name that matched 2+ differently-aggregated bound measures (e.g. both `Sum(Total)` and `Average(Total)`), silently resolving to whichever was bound first. Fixed with a new ambiguity check in `ChartBindingMutator.setSort`/`setRanking` — an already-qualified form is unaffected; a genuinely ambiguous bare name is refused, naming the qualified alternatives.
- **VCS-005** — `ChartBindingMutator.setShelf` rebuilds a chart's x/y field list from scratch on every write (`toChartRef` never sets `chartType`, by design), so any prior `set_chart_type` was silently lost on the next shelf edit — including the ordinary case of adding one more field to an already-typed shelf. Confirmed live to be specific to this write path: dragging a new field onto an already-typed shelf in the native Composer UI does not reset existing measures' types. Fixed by capturing each shelf's non-auto measures before the rebuild and restoring them by (column, aggregate[, secondaryY]) identity afterward.

Each fix has regression tests confirmed to fail without the fix (verified by temporarily reverting each) and pass with it. All four were also re-verified against a live, rebuilt StyleBI instance through the wiz plugin's own MCP tools (not just unit tests).

## Known remaining gap (not fixed here, out of scope)
VCS-005 has one further edge case: two measures sharing the exact same `chart/type` write-path label (same column+aggregate, differing only by `secondaryY`) still can't be disambiguated by `set_chart_type`'s own `field` argument — the ambiguity lives in the shared, non-wiz `ChangeChartTypeEvent`/`ChangeChartTypeProcessor`/`ChangeChartTypeService` classes (`inetsoft.web.binding`/`inetsoft.report.internal.graph`), also used by the native Composer UI's own retype flow. Deliberately not attempted here given the larger blast radius; the shelf-rewrite fix above degrades safely for this case (skips restoring rather than misapplying).

## Test plan
- [x] `ChartBindingMutatorTest` (36/36 passing, 8 new)
- [x] `BindingReadServiceTest` (19/19 passing, 2 new)
- [x] Each fix's regression test confirmed to fail without the fix, pass with it
- [x] All four re-verified live against a rebuilt StyleBI instance via the wiz plugin's own tools


NeverStopFMH:
Addressed both findings from the automated review in `15e46e8ba`:

1. **`requireUnambiguousMeasure`'s qualified-form fast path** — now counts qualified matches (not just bare-column matches) and refuses when 2+ refs stringify to the identical qualified name (the secondaryY-collision shape), since there's no further qualified string available to disambiguate them. New test: `rejectsAnAlreadyQualifiedSortByFieldThatIsItselfAmbiguous` (confirmed to fail without the fix).
2. **`sameMeasure`'s case-sensitive comparison** — switched to a small `equalsIgnoreCaseOrBothNull` helper, consistent with `requireDimension`/`requireUnambiguousMeasure`'s existing case-insensitive matching elsewhere in this class.

The third point (scope limited to x/y/group, not the single-value shelves) was confirmation-only — left as-is, since `requireDimension`/`readShelf` already scope `sortByField`/`measure` resolution to those same three shelves today.

PR #5299, opened 2026-09-16
Fix CheckBox selection revert after an interleaved viewsheet export (#76699 VFO-017 mechanism 2)

Nelson-cao:
## Summary
- `RuntimeViewsheet.setViewsheet(Viewsheet)` unconditionally forces `ViewsheetSandbox.resetRuntime()` (no way to pass `resetRuntime=false`). `VSExportService`'s whole-viewsheet "current view" export -- `get_viewsheet_image`'s fallback for a CheckBox, which cannot render in isolation -- calls it to swap the live `Viewsheet` for a clone and back, even though it deliberately passes `resetRuntime=false` to the sandbox-level `setViewsheet()` call it also makes right next to it.
- `resetRuntime()` clears `ViewsheetSandbox.parametersApplied`, the flag that normally makes `applyParameterToInput()` a no-op after its first run. Once cleared, the very next `VSInputService.refreshVS0 -> reset(clist)` cascade (`initing=true`) re-fires `applyParameterToInput()`, which reapplies a stale `VariableTable` snapshot over a selection change that had just landed correctly -- in the same request. Same class of bug as #74220 (undo/redo), just triggered by the export swap instead.
- Fix: call `ViewsheetSandbox.markParametersApplied()` right after the swap-restore completes, the same idiom `RuntimeViewsheet.restoreCheckpoint0()` already uses for #74220.
- Root-caused live with a debugger attached to a running StyleBI JVM (object-identity-correlated logpoints, then a stack-logging breakpoint that caught the exact revert call chain) -- see `docs/teams/2026-09-16-bug-76699-vfo017-checkbox-clear-lag/07-mechanism2-root-cause.md` in the stylebi-wiz repo. This is mechanism 2 of Redmine #76699 / VFO-017; mechanism 1 (the `CheckBox.value` `cellValue` snapshot) was already fixed and merged in #5289.

**Update (review round 1):** the reviewer found that the risk this PR originally described as hypothetical ("a third, unaudited caller could reopen this class of bug") was already live: `ExportControllerService.writeViewsheetExport`'s "current view" branch (the Composer toolbar's manual "Export" action) is a near-verbatim copy of the same swap-restore shape, unpatched. Applied the identical one-line fix there too. Also upgraded both regression tests to use a real, minimal `AbstractVSExporter` subclass instead of a bare mock, so `AbstractVSExporter.export()`'s own internal `rvs.setViewsheet` swap-restore (a second, independent `resetRuntime()` trigger) is genuinely exercised, not just the direct one. See `docs/teams/2026-09-16-bug-76699-vfo017-checkbox-clear-lag/06-review-mechanism2-r1.md` for the full review and `04-fix-mechanism2.md` for the updated fix writeup.

While auditing other `rvs.setViewsheet(...)` call sites for the same pattern: `VSBindingService.java:1940` already carries an equivalent `markParametersApplied()` fix (pre-existing, unrelated). `WizVsService.java:2495` (an exception-triggered rollback to the pre-mutation viewsheet during chart-mutation handling) has the same unprotected `resetRuntime()` shape but in a materially different context -- **flagged, not fixed in this PR**, since it needs auditing that rollback path's own semantics rather than just copying the one-line pattern. All other `rvs.setViewsheet(...)` sites found are permanent content transitions (new/replaced viewsheet content), not this bug's "temporary swap back to identical content" shape.

## Test plan
- [x] `VSExportServiceParametersAppliedRegressionTest` and `ExportControllerServiceParametersAppliedRegressionTest` each drive the real, private `writeViewsheetExport(...)` method via reflection with `current=true`, against a real (non-mocked) `Viewsheet`/`CheckBoxVSAssembly`/`ViewsheetSandbox`/`RuntimeViewsheet` and a real (non-mocked) `AbstractVSExporter` subclass. Both confirmed RED without their respective fix and GREEN with it.
- [x] Full affected-surface regression sweep, all green:
  - `inetsoft.web.viewsheet.service.*Test,inetsoft.report.composition.execution.*Test,inetsoft.report.composition.*Test,inetsoft.web.composer.vs.controller.*Test` -> 224/224 (1 pre-existing unrelated skip)
  - `inetsoft.web.wiz.viewsheet.*Test,inetsoft.web.wiz.script.*Test,inetsoft.report.io.viewsheet.*Test,inetsoft.web.binding.service.*Test` -> 1132/1132


Nelson-cao:
## Independent review + verification summary

Two rounds each of independent code review and independent test verification, run via `/deep-debug-workflow` for Redmine #76699 (VFO-017).

### Round 1
- **Review**: found the PR's own risk note ("a third, unaudited caller could reopen this") was actually a live, unpatched instance — `ExportControllerService.writeViewsheetExport`'s "current view" branch (Composer toolbar's manual Export action) has the identical unprotected swap-restore shape. Also flagged the original regression test's `VSExporter` mock never actually exercised `AbstractVSExporter.export()`'s own internal `resetRuntime()` trigger.
- **Verify**: confirmed the original fix's diff, red/green test behavior, and both regression sweeps (207/207, 1126/1126) exactly as claimed.
- Fixed in commit `c73c9235b`: identical `markParametersApplied()` fix applied to `ExportControllerService`, both regression tests rewritten to use a real `AbstractVSExporter` subclass (`NoOpVSExporter`) instead of a mock, plus a full audit of every other `rvs.setViewsheet(...)` call site in the tree (`VSBindingService.java:1940` already protected; `WizVsService.java:2495` shares the shape but in a materially different context — an exception-rollback path restoring a genuinely different pre-mutation state, not a temporary swap-back — flagged as a separate follow-up rather than folded into this PR).

### Round 2
- **Review**: confirmed both round-1 fixes are placed correctly (verified independently by reading the actual control flow, not re-trusting the round-1 account), confirmed the `WizVsService.java:2495` scoping call is defensible, no new overfit smell in `NoOpVSExporter`. No blockers.
- **Verify**: re-ran both regression tests from scratch in both directions (red/green), confirmed byte-identical restoration after each toggle, instrumented trace probes to independently confirm `AbstractVSExporter.export()`'s internal swap-restore genuinely fires during the test. One nuance noted for the record: the test's pass/fail outcome doesn't actually hinge on that internal trigger (the outer `markParametersApplied()` call re-arms the latch unconditionally after `export()` returns regardless), so it's correct-and-worth-keeping for production-path fidelity rather than closing a differential coverage gap in the strict sense — doesn't change the merge recommendation. Reproduced both regression sweeps exactly (224/224, 1132/1132).

**Net result: no blocking findings after two rounds.** `build`/Semgrep CI green; the automated `claude-review` bot check shows `fail` on the latest commit but the comment body is a bare "Claude encountered an error after 0s" — an infra failure, not a review finding (confirmed via the API).

Full archive: `stylebi-wiz/docs/teams/2026-09-16-bug-76699-vfo017-checkbox-clear-lag/` (`06-review-mechanism2-r1.md`, `06-review-mechanism2-r2.md`, `05-verify-mechanism2.md`, `05-verify-mechanism2-r2.md`).

Holding on merge pending explicit go-ahead.

Nelson-cao:
## CI's own automated review (round 1, commit 0faa20f9d) — for completeness

The `claude-review` GitHub Action check *did* run successfully once, on the original commit, before later erroring out on the round-2 commit (see above). It independently confirmed the fix's correctness and raised two non-blocking notes that are still open (neither was addressed by the round-1 human-equivalent review's fix pass, since that pass focused on the `ExportControllerService` duplicate and test-mock gap instead):

1. **Comment accuracy** (`VSExportService.java`, around the fix): the inline comment attributes the `resetRuntime()`/flag-clearing to the *restore* call (`rvs.setViewsheet(originalViewsheet)`), but `ViewsheetSandbox.setViewsheet` short-circuits on reference equality, so that particular call is actually a no-op — the real clearing happens at the earlier **forward** swap (`rvs.setViewsheet(cviewsheet)`). Doesn't affect correctness (the flag is still `false` by the time `finally` runs either way), but the comment could mislead a future reader tracing the call chain. Cosmetic — not required for merge, but cheap to fix now while the diff is fresh.
2. **Embedded/nested viewsheets**: `getSandboxes()` clears `parametersApplied` on the outer sandbox *and* every embedded-viewsheet sandbox. This fix (and the pre-existing `#74220` fix it mirrors) only re-arms the outer one. A CheckBox living inside an embedded `Viewsheet` assembly could in principle still exhibit this exact bug. Same inherited-not-introduced characterization as the `WizVsService.java:2495` finding from the independent reviews above — reasonable as a follow-up rather than a blocker, but worth tracking alongside it rather than letting it quietly not exist anywhere.

Neither is blocking, per both the bot's own review and the two independent human-equivalent rounds above, but flagging both explicitly here so they don't get lost if this PR merges as-is.

fredaan (approved):

PR #5305, opened 2026-09-16
fix(76730-wbs049): let convert_to_embedded accept JOIN/MIRROR/CONCAT tables

Zxm-happy:
## Summary

- `WorksheetAgentController.convertToEmbedded()` guarded on `instanceof BoundTableAssembly`, a sibling of (not a supertype of) the composite table types, so `convert_to_embedded` rejected JOIN/MIRROR/CONCAT tables with `Not a bound table: <name>` even though the delegate `AssetEventUtil.convertEmbeddedTable()` and the native Composer UI "Convert To Embedded Table" action both accept any `TableAssembly` with no such restriction.
- Broadens the guard to `instanceof TableAssembly` (and the cast alongside it), matching what the delegate method's own parameter type requires.
- Also catches `EmptyTableToEmbeddedException` (thrown when the resolved query result has zero columns — e.g. a join whose column selection was never refreshed before conversion) and rethrows it as a clean, field-named `PairingException`, matching how the native UI path (`ConvertEmbeddedService`) already handles the identical exception, instead of letting a raw internal exception with a Java object `toString()` leak to the LLM caller.

Root cause investigation and refutation: `docs/teams/2026-09-16-bugs-worksheet-join-namedgroup/bug-76730-wbs049/01-diagnosis.md` and `02-refute.md` in the `wiz` monorepo (Redmine #76730, WBS-049).

## Test plan

- [x] New `WorksheetAgentControllerConvertToEmbeddedTest` (separate class, needs `IntegrationTestConfiguration` for the `AssetDataCache` bean that real query execution resolves through):
  - `convertToEmbeddedConvertsJoinTableWithRealJoinedData` — real (non-mocked) `AssetQuerySandbox`, real INNER JOIN of two embedded tables, asserts actual joined cell values.
  - `convertToEmbeddedConvertsMirrorTableWithRealMirroredData` — real `MirrorTableAssembly`, asserts actual mirrored cell values.
  - `convertToEmbeddedStillRejectsNonTableAssembly` — guards against a future regression back to "any assembly name works".
  - `convertToEmbeddedThrowsPairingExceptionForEmptyJoinResult` — exercises the new `EmptyTableToEmbeddedException` → `PairingException` translation.
  - Command: `./mvnw.cmd -pl core -am -Dtest=WorksheetAgentControllerConvertToEmbeddedTest test` — 4/4 pass.
- [x] Full `WorksheetAgentControllerTest` (the file this change touches) — 185/185 pass, no regressions.
- [x] `./mvnw.cmd -pl core -am test-compile` — clean compile.


jshobe-inetsoft (approved):

PR #5307, opened 2026-09-16
Fix #76720: gate MV create background dispatch on noData

Nelson-cao:
## Summary
- `MvChangesetApplyService.applyCreate` passed the request's `runInBackground` through to `createMV` unconditionally, even when `noData:true` (the documented "fast, no-background-job" default). On a scheduler-less deployment this routed registration through `MVSupportService.createMV0`'s async `remoteCreatePool.submit(...)`, racing `applyCreate`'s own `existsInOrg` check that runs immediately afterward -- a real MV creation could be reported as `STATUS_FAILED`/`"mv not found after create"` moments before the MV became visible.
- `background` is now `false` whenever `noData` is `true`, matching `plugin/admin`'s own `mvTools.ts` contract (`runInBackground` only meaningful when `noData:false`).
- `MvChangePlanService.resolveCreate`'s preview string is updated the same way so it doesn't describe a `runInBackground` value `apply` no longer actually uses.

## Scope boundary (left alone, deliberately)
Explicit `noData:false` with no scheduler node still races `existsInOrg` via the same `remoteCreatePool` path -- a separate, opt-in case (real materialization requested) that forcing synchronous would regress by blocking `apply()` under `APPLY_LOCK` for the materialization duration. Tracked as a residual known risk, not part of this fix.

## Test plan
- [x] `MvChangesetApplyServiceTest`: 4 new regression tests via `ArgumentCaptor<Boolean>` verifying the `background` argument passed to `mvGateway.createMV(...)`:
  - defaults omitted -> `background == false`
  - explicit `noData:true, runInBackground:true` -> `background == false`
  - control case `noData:false, runInBackground:true` -> `background == true`
  - `noData:false, runInBackground:false` -> `background == false`
- [x] `mvn test -pl core -am -Dtest=MvChangesetApplyServiceTest` -- 8/8 tests pass, BUILD SUCCESS


Nelson-cao:
## Review (debug-workflow, bug #76720)

Reviewed the diff directly against the confirmed root cause (see `docs/teams/2026-09-16-bug-76720-mv-create-verify-race/bug-76720/{01-diagnosis,02-refute,03-fix,05-review-r1}.md` in the `stylebi-wiz` repo for the full trail). Note: can't post this as an "Approve" review since the PR was opened under the same account — posting as a comment instead. Verdict: **approve on content, no blockers.**

- The fix (`background = !noData && (runInBackground==null||runInBackground)`) gates on the general `noData` condition, not on any literal from the reported repro — fixes the actual contract violation (`plugin/admin/src/tools/mvTools.ts` already documents `runInBackground` as only meaningful when `noData:false`; the Java side wasn't enforcing it).
- Independently re-checked `MVService.java` (not just trusted the diagnosis): confirmed zero `existsInOrg` calls anywhere in that file — `create()`/`create0()` use a `createMVMap`/`CreateMVResponse.complete` poll pattern instead, so it's genuinely unaffected by both the bug and this fix.
- The bundled `MvChangePlanService.resolveCreate` change mirrors the same gate for the preview string only — confirmed display-only, doesn't feed the plan-hash comparison.
- The 4 new tests capture the real `background` argument via `ArgumentCaptor<Boolean>` through the actual `apply()` path — tests 1–2 would fail against the pre-fix code (old code always passed `background=runInBackground`), so this is real regression coverage, not decorative.
- The documented residual case (`noData:false` + no scheduler node still races) is a reasonable, explicitly-flagged scope boundary — forcing it synchronous would block `apply()` under `APPLY_LOCK` for real materialization duration, a regression per the plugin's own tool docs.

**Two procedural notes for whoever merges:**
1. `build` (Validation Build) was still `IN_PROGRESS` at review time — please confirm it's green before merging, per this repo's own StyleBI-side merge discipline.
2. `claude-review`'s FAILURE conclusion is a bot infra error (`is_error:true`, 0 turns, $0 cost — it never actually analyzed the diff), not a real finding against this PR's content.

Not merging this PR — left for human merge per the current debug-workflow run's instructions.

Nelson-cao:
## Live verification (debug-workflow, bug #76720)

Ran the charter's exact repro 3 times against a real running StyleBI server (local, hot-patched with this fix's compiled classes) via the plugin's own `admin-chat` MCP tools (`analyze_mv` → `preview_mv_changes` → `apply_mv_changes` → `get_mv_status`), targeting `Examples/Sales Summary` at the documented defaults (`noData`/`runInBackground` both omitted):

| Round | `apply_mv_changes` status | Entry status | Error |
|---|---|---|---|
| 1 | `applied` | `verified` | none |
| 2 | `applied` | `verified` | none |
| 3 | `applied` | `verified` | none |

3/3 clean — zero recurrence of the original `"rolled-back"` / `"mv not found after create"` misreport. `preview_mv_changes`'s own `proposedValue` (`runInBackground=false` for the default input) confirmed the server was actually running the patched code before any apply was made. Cleaned up after every round; final `get_mv_status` confirmed no residue.

Full write-up: `docs/teams/2026-09-16-bug-76720-mv-create-verify-race/bug-76720/04-verify.md` (stylebi-wiz repo).

Still not merging — left for human merge, and `build` (Validation Build) should still be confirmed green first.

jshobe-inetsoft (approved):

PR #5308, opened 2026-09-16
fix(76694): add atomic update verb for admin-chat license key replacement

NeverStopFMH:
## Summary
- `LicenseKeySettingsService` gains `replaceServerKey(oldKey, newKey)`, mirroring `addServerKey`/`removeServerKey`'s cluster-broadcast/auth-cache-reset side effects but calling `LicenseManager.replaceLicense` directly -- the same atomic (single in-memory swap, single persisted write) primitive Enterprise Manager's own "Edit License Key" dialog already resolves to.
- `LicenseChangeRequest`/`LicenseChangePlanService`/`LicenseChangesetApplyService` gain a real `update` verb: validates `oldKey` is installed, `newKey` differs and isn't already installed and resolves to a valid license type, re-verifies all of that fresh at apply time, and rolls back by calling `replaceServerKey` in reverse (exact, no claiming-node-drift advisory unlike `remove`'s own rollback).
- Closes a real, narrow gap in the existing remove+add workaround: each of `addServerKey`/`removeServerKey` persists independently via its own `saveLicenses()`, so a process crash between the two calls can leave a deployment with zero installed keys on disk -- `replaceLicense` never has that window.
- Companion `stylebi-wiz` plugin PR (inetsoft-technology/stylebi-wiz#2605) wires the new verb through `normalizeLicenseVerb`/`licensingTools.ts`.

## Test plan
- [x] `LicenseKeySettingsServiceTest`, `LicenseChangePlanServiceTest`, `LicenseChangesetApplyServiceTest` — 49/49 passing (`./mvnw -pl community/core -am test`).
- [x] Live end-to-end via the real `admin-chat` MCP tools against a local deployment built from this branch: swapped the deployment's only installed license key for a locally key-generator-minted test key with `verb: "update"`, confirmed via `list_license_keys`, then swapped back.
- [x] Live-verified all validation refusal paths this change adds: `oldKey` not installed, `newKey == oldKey`, `newKey` unparseable, `newKey` already installed as a different key.

Redmine: https://issues.inetsoft.com/issues/76694


NeverStopFMH:
## Round 1 review — head commit `254f9bb4e9f05a77b54130d5efe876488002a101`

### What was checked

Read all four production files in full (`LicenseChangePlanService.java`, `LicenseChangeRequest.java`, `LicenseChangesetApplyService.java`, `LicenseKeySettingsService.java`) plus the three test files, alongside the untouched sibling methods (`resolveAdd`/`resolveRemove`, `applyAdd`/`applyRemove`, `rollbackAdded`/`rollbackRemoved`, `addServerKey`/`removeServerKey`) to compare conventions. Also read `EnterpriseLicenseStrategy.addLicense`/`removeLicense`/`replaceLicense`/`replaceClaimedLicense`/`claimLicense` in `enterprise/src/main/java/inetsoft/enterprise/license/EnterpriseLicenseStrategy.java` in full to verify the atomicity and claiming-node-routing claims rather than trust the PR description/comments.

Confirmed, not just asserted:
- **Atomicity claim is true.** `replaceLicense`'s "this node holds the claim" fast path (`replaceClaimedLicense`) does exactly one `installedLicenses` swap + one `saveLicenses()` call. The remote branch looks up the node claiming **`oldKey` specifically** (`getClaimedClusterLicenses().get(oldKey)`) and exchanges a `ReplaceLicenseMessage`, vs. `addLicense`'s remote branch which picks **any** node with `getValue() == null` (an unclaimed slot). This genuinely confirms the "exact node vs. any node" distinction the PR's comments claim, so the "no claiming-node-drift advisory for update" reasoning in `rollbackUpdated`'s doc comment holds.
- **Adversarial same-key-across-entries plans are already refused.** Traced `resolve()`'s shared `seenKeys` set: both `key` and `newKey` for every entry (regardless of verb) are registered in the same set via `requireUnseen`. I built and traced `update(A to B)` + `remove(B)`, `update(A to B)` + `add(A)`, and the reverse orderings — all throw "duplicate entry for key" before any resolution logic runs, because the literal key strings collide across entries. I could not construct a plan where a resolved `update` entry contradicts a sibling `add`/`remove` entry on the same literal key string.
- **De-licensing count is correct in both places.** `resolve()` never increments `addCount`/`removeCount` for `update`; `apply()`'s independent fresh re-derivation (lines ~112-119) also only counts `VERB_ADD`/`VERB_REMOVE`, so an `update`-only plan can never trip `acknowledgeDelicensing`, and a `remove`+`update` plan's warning is driven only by the `remove`.
- **Rollback field mapping is not swapped.** `Undo.updated(oldKey, newKey)` gives `key=oldKey`, `newKey=newKey`; `applyUpdate` calls `replaceServerKey(oldKey, newKey)`; `rollbackUpdated` calls `replaceServerKey(currentKey=undo.newKey, originalKey=undo.key)` — genuinely symmetric, verified by reading both call sites side by side.
- **Check ordering in `resolveUpdate`** (installed -> differ -> collision -> valid-parse) mirrors `resolveRemove`/`resolveAdd`'s own precedent and surfaces the more actionable message first for the "two problems at once" case (oldKey not installed and newKey invalid simultaneously -> caller sees "not currently installed", which is correct since there is nothing to validate the replacement against once the source key doesn't exist).

### New findings

**Critical (92) — `applyUpdate` uses the raw, untrimmed `newKey` at apply time instead of the resolved/trimmed value, breaking the plan/apply consistency guarantee the whole hash-gated design exists to provide.**
- File: `core/src/main/java/inetsoft/web/admin/ai/licensing/LicenseChangesetApplyService.java`, `applyOne` (the new `else` branch, ~line 223-226): `applyUpdate(txId, task, key, original.getNewKey(), ...)`.
- `key` here is `change.property()` — the trimmed `oldKey` produced by `LicenseChangePlanService.requireKey`/`resolveUpdate` during `resolve()`. But `original.getNewKey()` is the raw `LicenseChangeRequest.newKey` field (plain getter/setter, no trim — confirmed in `LicenseChangeRequest.java`), never the trimmed value `requireNewKey` computed during resolution. Contrast with `add`/`remove`, where the single key used at apply time (`change.property()`) is always the resolved/trimmed value — there is no equivalent raw fallback for those verbs.
- Concrete failing scenario: caller submits `update(key="K1", newKey="K2\n")` (a trailing newline is a very plausible artifact of an LLM/agent pasting a license blob — exactly the class of input CLAUDE.md's tool-robustness section exists for). `resolve()`/preview trims `"K2\n"` to `"K2"`, validates and parses `"K2"` successfully, and returns a plan whose `proposedValue` reflects the clean `"K2"` license. The operator approves that plan. At `apply()`, `applyUpdate` receives the raw `"K2\n"`: it re-checks `isInstalled("K2\n")` (false — no whitespace-containing key is installed) and re-parses `licenseManager.parseLicense("K2\n")`. Two outcomes are both bad:
  1. If key parsing is strict about the literal string (checksum/signature over the exact bytes), this now resolves to `LicenseType.INVALID`, and the update fails with the misleading message "newKey no longer resolves to a valid license at apply time (concurrent change)" — even though nothing changed concurrently; the plan was valid and got refused only because of the untrimmed carry-through.
  2. If parsing tolerates the whitespace internally, `licenseKeySettingsService.replaceServerKey("K1", "K2\n")` installs a license whose `key()` may retain the literal `"K2\n"` string (since `EnterpriseLicenseStrategy.replaceLicense`/`parseLicense` build the `License` from exactly the string handed to them) — a different literal from the canonical `"K2"` the operator reviewed and that got audited/hashed into the plan, and one that later `isInstalled("K2")`/removal-by-key lookups would no longer match.
- No test in `LicenseChangePlanServiceTest`/`LicenseChangesetApplyServiceTest` exercises a `newKey` with incidental whitespace, so this passed the author's "49/49 tests" and live smoke-test claims without being caught.
- Fix: pass the resolved/trimmed `newKey` through to apply time — either add a raw-value field to `PlanChange`/`ResolvedEntry` the way `property` already carries the trimmed `oldKey`, or at minimum call `original.getNewKey().trim()` in `applyOne` (less good, since it re-derives rather than reuses the already-validated value, but at least matches what `resolve()` validated).

**Important (82) — No test exercises `applyUpdate` itself throwing mid-apply, unlike the dedicated coverage `applyAdd`/`applyRemove` each get.**
- File: `core/src/test/java/inetsoft/web/admin/ai/licensing/LicenseChangesetApplyServiceTest.java`.
- This file has, for both `add` and `remove`: (a) a test where the mutating call itself throws (`midApplyFailureRollsBackThePreviouslyAppliedAddButReportsRollbackFailedForTheUnknownState`, throwing on `addServerKey("K2")` — K2 being the failing entry itself) and (b) a Bug-76567 variant where the throw fires strictly before the mutating call (`unknownStateBeforeMutationDoesNotForceRollbackFailedWhenRollbackIsClean`/`...ForRemoveWhenRollbackIsClean`), to prove the `mutationEntered` classification is correct in both directions.
- The only new throw-related test for `update`, `rollbackOfUpdateRevertsViaReplaceServerKeyWithNoAdvisory`, injects the throw on `addServerKey("K3")` — a different, later, add-verb entry in the same plan. It proves `rollbackUpdated` is invoked correctly when a later item fails, but it never makes `replaceServerKey` itself throw, so `applyUpdate`'s own `mutationEntered.set(true)` -> throw -> `unknownStateMutationEntered` path, and the "throw strictly before `replaceServerKey`" path (e.g. from `findInstalled`/`isInstalled`/`parseLicense` inside `applyUpdate`), are unverified. The code mirrors `applyAdd`'s already-proven structure closely, so risk is likely low, but this is a real, checkable gap against this file's own established per-verb coverage convention, and the setup-method doc comment (line ~54-59) documents "mid-apply failures are injected by having a mocked addServerKey/removeServerKey call throw" without ever mentioning `replaceServerKey`, confirming the omission wasn't a deliberate equivalent-coverage decision.

**Minor (55) — `resolveAcceptsValidUpdate` only asserts non-null, not correctness, of `currentValue`/`proposedValue`.**
- File: `core/src/test/java/inetsoft/web/admin/ai/licensing/LicenseChangePlanServiceTest.java`, ~line 429-444.
- `assertNotNull(change.currentValue()); assertNotNull(change.proposedValue());` would still pass if a future refactor accidentally swapped which projection (`oldProjection`/`newProjection`) went into `currentValue` vs. `proposedValue` in `resolveUpdate`. Current implementation is correct (verified by reading `resolveUpdate`); this is a latent-regression risk in the test, not a live bug.

### Verdict: changes requested (blocking on the Critical finding above)

NeverStopFMH:
Addressed round 1's findings in ab16281ee:

**Finding 1 (critical) — fixed.** `LicenseChangesetApplyService.applyOne` was passing the raw, untrimmed `original.getNewKey()` to `applyUpdate` instead of a trimmed value. Confirmed there's no existing resolved slot for `newKey` the way `key`/`change.property()` already carries the trimmed old key through — `PlanChange.proposedValue()` holds a `LicenseKeyProjection.canonical()` string (`key=...;type=...;valid=...;description=...`), not the raw new key literal, so it can't be reused directly as the value passed to `replaceServerKey`. Went with the simplest option from the proposal: `applyOne` now trims via `LicenseChangePlanService.requireNewKey("apply." + key, original.getNewKey())`, the same validator `resolve()` already uses — this is genuinely idempotent (only a null/blank check + trim, no other validation `requireNewKey` performs is skipped) and by the time `applyOne` runs, `resolveUpdate` for that same index has already succeeded once via the fresh top-of-`apply` `resolve()` call, so `requireNewKey` is guaranteed not to throw here. Added a regression test (`appliesAnUpdateTrimmingWhitespaceFromNewKeyAtApplyTime`) that sends a whitespace-padded `newKey` and asserts `replaceServerKey` is called with the trimmed value.

**Finding 2 (important) — fixed.** Added the two missing update-verb tests mirroring the existing add/remove pairs: `midApplyFailureAtUpdatesOwnMutatingCallReportsRollbackFailedForTheUnknownState` (throw AT `replaceServerKey`, after `mutationEntered` is set) and `unknownStateBeforeUpdatesOwnMutatingCallDoesNotForceRollbackFailedWhenRollbackIsClean` (throw via `parseLicense`'s own re-verification inside `applyUpdate`, strictly before `mutationEntered`).

**Finding 3 (minor) — fixed.** `resolveAcceptsValidUpdate` now asserts `currentValue`/`proposedValue` contain the expected `key=`/`type=` content instead of merely non-null.

Ran `./mvnw -q -o -pl community/core -am test -Dtest=LicenseChangePlanServiceTest,LicenseChangesetApplyServiceTest,LicenseKeySettingsServiceTest` — 52 tests, 0 failures, 0 errors.

No pushback — all three findings held up against the current source.

NeverStopFMH:
## Round 2 review (verdict: approved — no `gh pr review` since the authenticated account is the PR author; posting as a comment instead)

**Head commit reviewed:** `ab16281ee1344823444fdb9d3412d52b8a146b81`
**Range reviewed since round 1:** `254f9bb4e..ab16281ee` (single commit: "fix(76694): trim newKey at apply time, add update-verb rollback tests (round-1 review)")

### What was checked
- Read the full current `LicenseChangesetApplyService.applyOne`/`applyUpdate`/`rollbackUpdated` and `LicenseChangePlanService.requireNewKey`/`requireKey`/`resolveUpdate`/`hash`.
- Traced exactly how `key` and `newKey` reach the mutating calls at apply time, including where each is (re-)trimmed.
- Ran the full targeted suite: `./mvnw -q -o -pl community/core -am test -Dtest=LicenseChangePlanServiceTest,LicenseChangesetApplyServiceTest,LicenseKeySettingsServiceTest` → **26 + 23 + 3 = 52 tests, 0 failures, 0 errors** (surefire reports confirmed directly, not just console output). Matches the fixer's claim.
- Diff stat confirms only the 3 files reported were touched (7 lines in the service, 8 in one test file, 96 new lines in the other) — nothing else slipped in.

### Round 1 findings

**1. Critical — raw `original.getNewKey()` reaching `applyUpdate` untrimmed: FIXED.**
`applyOne` now does `String newKey = LicenseChangePlanService.requireNewKey("apply." + key, original.getNewKey());` and passes that into `applyUpdate`. Verified:
- `requireNewKey` is a pure null/blank-check + `.trim()`, no side effects, fully idempotent.
- Critically, `original` (`originals.get(i)` from `req.getChanges()`) is the *exact same* raw request object that `planService.resolve(req)` — called at the very top of `apply()` — already ran through `requireNewKey` inside `resolveUpdate`, in the same single-threaded `apply()` invocation, with no intervening mutation possible. So the second call in `applyOne` is not a "re-validate against a stale preview" situation at all — it's a second call on the identical immutable input already proven valid earlier in the *same* call. There is no scenario in which the second call throws while the first (inside `resolve()`) didn't; a hash mismatch would already have been caught earlier if `req.getChanges()` differed from what was hashed.
- Tried to build a counter-example: "preview with clean `newKey`, apply with whitespace-padded `newKey`" — traced through `hash()`: it hashes `PlanChange.proposedValue` (a canonical projection built from `resolved = licenseManager.parseLicense(trimmed-newKey)`), not the raw string, so a whitespace-only difference produces the *same* hash both times. Confirmed this exact case is what `appliesAnUpdateTrimmingWhitespaceFromNewKeyAtApplyTime` exercises, and it passes.
- Checked `applyUpdate`/`rollbackUpdated` for any other unfiltered use of `original.getNewKey()` or similar raw fields reaching `replaceServerKey`/`isInstalled`/`parseLicense` — none found; every call site inside `applyUpdate` uses the `newKey` parameter, which is now always the trimmed value.
- No remaining path where the raw value leaks through. Closed.

**2. Important — missing throw-mid-`applyUpdate` test coverage: FIXED, and verified rigorous.**
Read both new tests in full and traced exact call counts:
- `midApplyFailureAtUpdatesOwnMutatingCallReportsRollbackFailedForTheUnknownState` stubs `replaceServerKey("K1","K2")` to throw. Confirmed via code reading that `mutationEntered.set(true)` in `applyUpdate` runs immediately before `replaceServerKey`, so this throw is genuinely "at the mutating call, after mutationEntered." Assertions correctly check `STATUS_ROLLBACK_FAILED`, the `"state unknown"` rollback-failure entry for K1, and that K3's own add still rolled back cleanly.
- `unknownStateBeforeUpdatesOwnMutatingCallDoesNotForceRollbackFailedWhenRollbackIsClean` stubs `parseLicense("K2")` with `.thenReturn(valid).thenReturn(valid).thenThrow(...)`. Traced the actual call count: (1) the test's own `planService.resolve(...)` call to compute `hash`, (2) `apply()`'s own fresh `planService.resolve(req)` at the top, (3) `applyUpdate`'s own fresh re-verification `licenseManager.parseLicense(newKey)` — which happens strictly before `mutationEntered.set(true)`. The three-call stub sequence lines up exactly with this, so the throw really does land at call #3, before mutation. Assertions correctly check `STATUS_ROLLED_BACK`, null `rollbackFailures`, and `replaceServerKey` never invoked.
- Both tests follow the file's existing conventions (same `installed`/`valid()`/`add()`/`update()`/`applyRequest()` helpers, same `mockAudit()` try-with-resources pattern, same assertion style as the sibling add-verb tests they're explicitly modeled on). No overfit-to-repro-literal smell — they assert on the actual status/rollback-failure semantics, not incidental values.

**3. Minor — `resolveAcceptsValidUpdate` content-checked assertions: FIXED and confirmed effective.**
`LicenseKeyProjection.canonical()` is `"key=" + key + ";type=" + type + ";valid=" + valid + ";description=" + description`. The test asserts `currentValue` contains `"key=K1"` and `proposedValue` contains `"key=K2"` (both `type=CPU`, so `type` alone wouldn't catch a swap, but `key=` does). Confirmed: if `resolveUpdate` swapped `oldProjection`/`newProjection` when constructing `PlanChange`, `currentValue` would contain `"key=K2"` instead of `"key=K1"` and the test would fail. This is a real, load-bearing check, not vacuous.

### Sibling-asymmetry question (round 2's specific ask)
Investigated whether `key` (validated once, via the plan's own `change.property()`) and `newKey` (now re-validated a second time via a fresh `requireNewKey` call) behave differently for a whitespace-only difference between preview and apply. They do not, for two reasons: (a) `key` *is* re-validated at apply time too — via the same `planService.resolve(req)` call at the top of `apply()`, which reruns `requireKey`/`.trim()` on the current request's raw field — it just doesn't need a *second* explicit call because the trimmed result is retained in `PlanChange.property()`; `newKey` has no equivalent retained slot in the shared `PlanChange` record (confirmed by reading `PlanChange.java` — no `newKey` field), which is exactly why `applyOne` has to re-derive it independently. (b) Both the old-key lookup (`byKey.get(trimmedOldKey)`) and the new-key parse (`parseLicense(trimmedNewKey)`) happen only after trimming in both the preview and apply code paths, and the plan hash is computed over post-trim canonical projections, not raw strings — so a whitespace-only difference in either field is silently absorbed identically on both sides. No asymmetry, no bug.

### New findings this round
None. Diff is minimal and precisely scoped to the reported gap; no unrelated changes, no unused code, no new abstractions.

### Verdict: **approved**

jshobe-inetsoft (approved):

PR #5309, opened 2026-09-16
Bug #76700, fix set_selection_source for Logical-Model selection columns

jkidacki-inetsoft:
## Summary

- `set_selection_source` returned `ok:true` for a Logical-Model-sourced SelectionList/SelectionTree/TimeSlider/Calendar column, but `selectedColumn` always read back `null` (`get_assembly_properties`) and the rendered selection list stayed empty — confirmed live against Examples/Orders' "Order Model".
- Root cause: `SelectionBindingService.columnRef()` split a folded Logical-Model column name (`"Customer:Region"`) into `entity="Customer"`/`attribute="Region"` when building the persisted `OutputColumnRefModel`. The property dialog's own read-back tree (`VSOutputService.getSelectionTablesTree`) represents that same column with `entity` left `null` and the whole compound string as `attribute` — the shape every logical-model column entry uses server-side. The mismatch meant the read-back match (`SelectionDialogService.findSelectedOutputColumnRefModel`) never found the column back.
- Fix: stop splitting — keep the full column string as `attribute` (entity stays `null`), matching exactly what a human's own interactive selection binding already produces through the same tree.

## Root Cause

`SelectionBindingService.columnRef()` built `AttributeRef(entity, attribute)` by splitting on `:`, but the read-back path never did the reverse split — it always represents a logical-model column as `entity=null`, `attribute="Customer:Region"` (the full compound string). The non-cube comparison branch in `findSelectedOutputColumnRefModel` only compares bare `attribute` strings, so `"Region" != "Customer:Region"` and the match always failed.

## Fix

`columnRef()` no longer splits the column name at all — the whole string becomes `attribute`, `entity` is never set. This applies uniformly to all four call sites that share this helper (SelectionList, SelectionTree, TimeSlider, Calendar) via `columnRef()`/`columnRefs()`.

## Test Plan

- [x] Unit: `SelectionBindingServiceTest` — 14/14 pass (12 pre-existing physical-table cases unaffected, 2 new regression tests: single-column `SelectionListVSAssembly` path and the array-form `columnRefs()` path used by SelectionTree)
- [x] Live, against a from-source server with Examples/Orders' "Order Model": bound a SelectionList to `Order Model`/`Customer:Region` on both a direct-source Logical-Model viewsheet and a worksheet-based viewsheet with an explicit `Customer` entity table — `get_assembly_properties` now resolves `selectedColumn` correctly, and `get_viewsheet_image` renders real values ("USA East", "USA West") instead of an empty list.
- [x] Code review (code-reviewer agent): no issues found at confidence ≥ 80.

Fixes Redmine #76700.


jshobe-inetsoft (approved):

PR #5310, opened 2026-09-16
Fix Bug #76732 (VCC-001): refuse a hidden axis title's region properties instead of NPEing

fredaan:
## Summary

- `ChartArea` only builds a `TitleArea` for an axis whose title is currently visible (its constructor skips building `xTitleArea`/`x2TitleArea`/`yTitleArea`/`y2TitleArea` when the axis's `TitleDescriptor.isVisible()` is false, since `GraphGenerator`'s title-spec builder sets a null label in that case). `RegionPropertyDialogService.getTitleFormatDialogModel` (`core/src/main/java/inetsoft/web/composer/vs/dialog/RegionPropertyDialogService.java:261`) dereferenced that null `TitleArea` unconditionally, so `list_chart_region_properties`/`set_chart_region_properties` with `region:"title"` on a hidden axis title crashed with a raw NPE instead of a clear, named error.
- `ChartRegionPropertyService.set()` shares the exact same crash site through its own `readModel()` prefetch before writing, so the write path was affected identically, not via a separate defect.
- Fix: `ChartRegionPropertyService.requireExistingTarget` — the single point upstream of both `list()` and `set()` — now refuses a `title` region whose target axis title is hidden, with a named-field error pointing at `set_chart_element_visibility`/`list_chart_elements`, before either path can reach the crashing dereference.

## Test plan

- Added `ChartRegionPropertyServiceTest`: `refusesToListAHiddenTitleRegionInsteadOfNpeing`, `refusesToWriteAHiddenTitleRegionInsteadOfNpeing`, `stillListsATitleRegionWhenVisible` (companion success case; existing title tests already cover the write-success path).
- Ran `./mvnw -q -pl core -Dtest=ChartRegionPropertyServiceTest test -DskipITs -o` — all 48 tests pass (0 failures, 0 errors).


fredaan:
## Review round 1 — changes requested

Head reviewed: `bd7d8d19e` (single commit).

**What I checked:** full diff, the diagnosis/refutation/fix reports, `ChartRegionPropertyService`'s
`list`/`set`/`requireExistingTarget`/new `requireVisibleTitle`, `ChartRegionResolver.canonical`/
`requireAxis`, `ChartRegionHandler.getTitleDescriptor`/`getTitleArea`, and `ChartFormatConstants`.
Confirmed both `list()` and `set()` (via its own `readModel` prefetch) go through the new guard, the
"chart" whole-title case is refused earlier so `requireVisibleTitle` only ever sees `x/x2/y/y2`, and
the guard's canonicalization correctly handles mixed case and long-form aliases (`top_x_axis`, etc).
Read the 3 new/changed test cases in full — the harness uses a **real** `ChartRegionHandler`, so the
hidden/visible fixtures genuinely exercise the real dispatch rather than a mock that always agrees
with the input; no nested-stubbing violation found in the committed code; `verifyNoInteractions`
confirms the guard fires before any call to `RegionPropertyDialogService`.

**Verified myself (not just the fix report's claims):**
- `./mvnw -q -pl core -Dtest=ChartRegionPropertyServiceTest -DfailIfNoTests=false test -DskipITs -o`
  → 48/48 pass (surefire report checked directly).
- `./mvnw -q -pl core -am compile -o` → clean, exit 0 (confirms the new constructor param doesn't
  break any other caller; grepped and only one other `new ChartRegionPropertyService(...)` exists,
  in the test file — production wiring is `@Autowired`/`@Component`, no update needed elsewhere).
- CI `build` check is still pending at review time; not relied on for this verdict per this
  workflow's rule (self-run verification above is what this verdict rests on).

**Finding (important, 80-89): guard coverage gap.**
`core/src/main/java/inetsoft/web/composer/vs/dialog/RegionPropertyDialogController.java:164-174`
(`GET /api/composer/vs/title-format-dialog-model/{objectId}`) — the **native StyleBI Composer's own
REST endpoint** — calls `RegionPropertyDialogService.getTitleFormatDialogModel` directly, bypassing
`ChartRegionPropertyService` (and the new `requireVisibleTitle` guard) entirely. A hidden axis title
requested through this endpoint still hits the identical unguarded NPE at
`RegionPropertyDialogService.java:261` this bug report is about. The diagnosis itself named a second,
narrower fix option — a null check directly in `getTitleFormatDialogModel` around line 261 — as an
alternative to the upstream guard, calling it "cheaper but duplicated"; it didn't note that this
alternative is also the only one of the two that closes *every* caller, including this controller.
Suggested: add that null check too (in addition to, not instead of, the existing
`ChartRegionPropertyService` guard, whose tool-specific error message is more actionable for the
plugin's own callers) so the underlying defect is closed at its root for every caller, not only the
wiz-agent one.

No other findings — fix scope, test correctness, and the build are otherwise sound.

**Verdict:** changes requested.

fredaan:
## Review round 2 — approved (zero findings)

Head reviewed: `d7e14e316` (range since round 1: `bd7d8d19e..d7e14e316`).

**Round-1 finding status: fixed.** The gap I found (native Composer's own REST endpoint,
`RegionPropertyDialogController.getTitleFormatDialogModel`, bypassing the wiz-only guard) is closed
by moving the null check into `RegionPropertyDialogService.getTitleFormatDialogModel` itself —
verified the check sits exactly between the `titleArea` assignment and its dereference, and re-grepped
every caller of `getTitleFormatDialogModel`/`getTitleArea` fresh (not reusing the round-1 list): same
3 real call sites as before, but the fix now lives in the one method upstream of all of them, so
the controller and the wiz service are both covered. Checked the exception convention against
`TableViewPropertyDialogService.java:170-197` — matches (`IllegalArgumentException`, plain message,
no wiz-tool-specific language, correctly different from the wiz-side guard's message).

**Test quality check:** `RegionPropertyDialogServiceTest` spies a **real** service instance (verified
constructor wiring matches exactly) and stubs only `getChartArea()` (the one thing a plain unit test
can't build without a live sandbox/graph layout) — both tests exercise the real method body, not a
tautology. Confirmed `setTitleFormatDialogModel` is still correctly untouched (no parallel defect,
re-read lines 275-303).

**Verified myself, not just the fix report's claims:**
- `./mvnw -q -pl core -Dtest=RegionPropertyDialogServiceTest,ChartRegionPropertyServiceTest -DfailIfNoTests=false test -DskipITs -o`
  → 2/2 and 48/48 pass (surefire reports checked directly).
- `./mvnw -q -pl core -am compile -o` → clean, exit 0.
- Worktree confirmed clean and at the right commit via explicit `--git-dir`/`--work-tree` (a bare
  `cd`+`git status` in this shared worktree setup hits a known auto-discovery quirk and reports
  unrelated files — not a real issue with this branch).
- CI `build` is still pending; not relied on for this verdict.

Walked mixed-case/long-form axis targets, all four axis types, the whole-chart-title case, and
double-guard interaction between the wiz-side and root-level checks — found no counter-example.

**No new findings. Verdict: approved.**

Attempted `gh pr review --approve` — blocked by GitHub's self-review restriction (same account as
PR author), so recording the approval here instead. **Human merge needed.**

jshobe-inetsoft (approved):

PR #5311, opened 2026-09-16
Chart palettes: brushing recedes each series to its own companion colour

FigurePaper:
Stacked on #5275 — open against `feature-chart-companion-colors` so the diff here is only this slice. Retarget to `epic-74519` once #5275 merges.

Implements ENGINE §2 of the chart palette redesign. Design: `docs/superpowers/specs/lookfeel/2026-09-15-brushing-companion-colors-design.md`.

## What changes

Brushing a chart used to paint every unselected mark one flat grey and every selected mark one flat red, whatever the chart's palette. The moment you brushed, you could no longer tell the series apart — which is the one thing brushing exists to help you do.

On a chart carrying a modern `VizMark`, an unselected mark now recedes to a soft companion of its **own** series colour, and a selected mark keeps that colour. Series identity survives brushing in both states, and red leaves the marked path.

## What does not change

- An unmarked chart, a report chart (`VizContext.LEGACY`), and any deployment whose `format.css` declares `.brush-dim-color` or `.brush-highlight-color` are pixel-identical to before. The two CSS properties are a tuned pair, so declaring either one drops the whole chart back to the flat path rather than producing a hybrid nobody designed.
- Maps keep flat brushing. `MapGenerator` and `VGraphPair` are untouched — see Deferred.

## How

`CompanionBrushColorFrame` wraps the chart's real colour frame and returns the companion of whatever that frame would have answered. It is keyed on **colour**, not series index: `companionOf` walks the active palette for a matching slot and derives from the colour itself when there is none. That is what lets a colour-unbound or measure-coloured chart resolve an answer, and why no call site needs an index threaded through it.

The same frame answers "is this mark brushed?", replacing colour-identity comparisons that stop working once colours vary.

## Commits

1. **Design and the recovered palette handoff** — also commits the external handoff, which was never in the repo and had come close to being lost twice.
2. **Implementation plan.**
3. **`BrushedMarks` + `BrushingComparator`** — the only commit with no dependency on #5275; it rebases cleanly onto `epic-74519` on its own and was verified building and passing in isolation.
4. **Lift `companionOf` to `VSChartPaletteDefaults`** — pure refactor; the companion mechanism now has three consumers rather than one.
5. **`CompanionBrushColorFrame`** — unwired, so no chart renders differently.
6. **Wire it in** — the behaviour change.

## Worth a reviewer's attention

Wrapping the colour frame moves the brush predicate one level down, and **four** callers looked for it at the top level. Each failed silently rather than loudly:

| caller | symptom if missed |
|---|---|
| `addDensityContour` point selector | contour selects zero points |
| `isApplyColor` | custom shapes stop being tinted |
| `ExcelChartHelper` guard | *inverse* — a legacy target chart has no frame to ask, so it reported every mark unbrushed and could change an exported workbook |
| `RelationGeometry.isApplicable` | a brushed tree rendered its root unbrushed, or took its child's hue |

The Excel guard keeps its flat colour comparison as a fallback for exactly the legacy case, where brushing colours are still uniform and the comparison is still correct. `createBrushingTargetColorFrame` takes its base as a parameter rather than reading the generator's colour frame, which its second caller reaches after that field is replaced.

## Testing

Full `core` suite green. Unit coverage for the frame, the predicate, the resolver, and the drift guard.

The gate and the CSS fallback have **no unit test** — both are decided by a boolean on `GraphGenerator`, and reaching it from a unit test means standing up a full chart info, descriptor and dataset. Their coverage is the manual pass, which is recorded in the design's §6 along with a cheaper route if someone wants that coverage later.

Manually verified in light and dark: multi-series bar and line, colour-unbound chart, tree/network with a node colour binding, brushing target dashboards, maps, an unmarked dashboard, and a deployment with `.brush-dim-color` set.

## Deferred, with triggers

Recorded in the design's §7:

- **Maps** — `MapGenerator` applies `getDarkColor()` to the brush colour, which has no per-series meaning yet. Taking this up requires converting `VGraphPair:825` in the same change; it is correct only while map colours stay uniform.
- **Density contour colour** — `DensityFormVO` resolves by contour band index, not by data row, so the per-row decorator has nothing to attach to. Its point selector was fixed; its colour stays flat.
- **Connector lines on a relation chart** — brushing the topmost node also brushes the lines to its children, in legacy too. Edges resolve through a different frame; observed, not changed.
- **ENGINE §3 palette overflow** — manual verification fired this trigger. A chart past slot 8 draws from the legacy tail, whose lightest entries have no room to recede. Measurements in `docs/superpowers/specs/lookfeel/engine-3-palette-overflow-trigger.md`. The light-end fix for that went into #5275, where the rule lives.


FigurePaper:
Thanks — both points actioned.

**Duplicated CSS-gate computation** — fixed. Extracted `companionBrushing(VizContext)` beside `isBrushingSource`; both constructors call it, so the CSS-override rule is stated once.

**Shared `acolor` across per-measure iterations** — verified rather than reasoned. Tested a multi-style bar chart with two measures, each with its own colour binding, as a brushing target: each measure's dimmed layer recedes to a companion of its own colour, so the composite is not carrying stale per-measure state. Bar is the demanding case, since line, point, radar, candle, stock and boxplot deep-clone the composite per element at `GraphGenerator:4675` and would have hidden a fault. The sharing model itself predates this change and is unaltered by it.

Good catch that the manual list was too vague to act on — `isMultiAesthetic()` is `isMultiStyles()`, so a chart coloured by a dimension never reaches that path. The design's manual pass now names the setup and both outcomes.

**Linear palette scan** — agreed, leaving it. Memoized per frame instance and the head is eight slots.

jshobe-inetsoft (approved):
Approving based on Claude's completed automated review, which found no correctness, security, or backward-compatibility issues (two minor non-blocking notes only). All checks pass and there are no merge conflicts.

PR #5312, opened 2026-09-16
fix(76728): log DependencyTool.getDependencies lookup failures instead of silently swallowing them

fredaan:
## Summary

Redmine #76728: `apply_viewsheet_changes` force-delete on a viewsheet with a
real dependent came back with the change entry's `advisory` field `null`,
even though the immediately-prior unforced `preview_viewsheet_changes`
correctly refused and named the dependent. Real operator harm: a dependent
viewsheet silently broke after the delete, and since viewsheet delete has no
recycle-bin path, the deleted asset couldn't be restored.

Root cause: both the viewsheet and data-source apply paths derive their
`advisory` text from `DependencyTool.getDependencies()`'s return value,
treating an empty list as "confirmed no dependents". `getDependencies()`
wrapped its lookup (`DependencyStorageService.getWithOrg`) in
`catch(Exception ignored)` with **no logging at all** — unlike sibling
methods (`remove`/`rename`) in `DependencyStorageService`, which do log
failures. Any transient exception during the lookup (e.g. a per-org
`KeyValueStorage` not yet loaded) silently became an empty list,
indistinguishable from a genuinely dependency-free asset.

## Fix

- `core/src/main/java/inetsoft/uql/asset/sync/DependencyTool.java`: the
  swallowed exception in `getDependencies()` is now logged with
  `LOG.warn`, naming the entry id, so a recurrence is diagnosable from logs.
- The return contract (empty list on failure) is intentionally left
  unchanged — `getDependencies()` has other callers (schedule-task
  dependents, etc.) this fix must not affect.
- Scope note: the viewsheet/data-source apply-time callers
  (`ViewsheetChangesetApplyService`/`DataSourceChangesetApplyService`) that
  would ideally distinguish "confirmed zero dependents" from "lookup
  failed" for the advisory text live in the `enterprise` module, which is
  currently mid an unrelated, in-progress, uncommitted migration into this
  `community` submodule and is out of reach from this PR (those files don't
  exist yet on this branch). That caller-side distinction is a follow-up
  once the migration lands.

## Test plan

- [x] Added `DependencyToolGetDependenciesLookupFailureTest` — forces
      `DependencyStorageService.getWithOrg` to throw and asserts a WARN
      naming the entry id is logged, while the empty-list return contract
      is preserved. Confirmed it fails on the pre-fix code (0 warnings
      logged) and passes after the fix.
- [x] `./mvnw -pl core -am test -Dtest=DependencyToolGetDependenciesLookupFailureTest` — 1/1 pass
- [x] Broader regression sweep, no failures:
      `./mvnw -pl core test -Dtest="inetsoft.uql.asset.sync.*Test,inetsoft.uql.asset.CheckSheetRemoveableStaleStoreTest,inetsoft.uql.asset.LocalDependencyHandlerTest"`


jshobe-inetsoft (approved):

PR #5314, opened 2026-09-16
Bug #76737, fix missing agent-connected indicator on agent-created sheets

jkidacki-inetsoft:
## Summary

- The Composer tab-bar "agent connected" indicator never appeared on a viewsheet/worksheet the wiz AI-agent plugin creates itself (`open_base_worksheet`, `create_viewsheet`, `create_worksheet`), even though the agent session was genuinely live and functional against that runtime.
- Root cause: a structural ordering bug, not a timing coincidence — the notification was published to a channel with no subscriber, deterministically, for every brand-new runtime.
- Fix: carry the "agent active" state directly on the command that opens the tab, instead of racing a second, separately-timed push.

## Root Cause

`SheetAgentBroadcastService.sendAgentActive()` publishes `SetAgentActiveCommand` to `CommandDispatcher.COMMANDS_TOPIC` — a per-runtime channel the browser only has a subscriber for once it has already opened that specific runtime (i.e. once a pane component for that `runtimeId` exists with its own STOMP subscription).

In `SheetOpenService.java`, all three "attach an agent to a brand-new runtime" methods (`openBaseWorksheet`, `createViewsheet`, `createWorksheet`) called `broadcast.sendAgentActive(session)` **before** `broadcast.sendToComposer(..., OpenComposerAssetCommand)` — and it's that `OpenComposerAssetCommand` that causes the browser to open/subscribe to the new runtime in the first place. So the notification was sent to a channel nobody was listening to yet, and was silently lost every time.

Compare `SheetJoinService.join()` (the normal human-pairs-a-code flow, unmodified by this fix): it only ever attaches to a runtime the human's browser already has open, so the channel is already subscribed and the identical `sendAgentActive` call works correctly there.

## Fix

- `OpenComposerAssetCommand` (Java + TypeScript) gains two new fields: `agentActive: boolean` and `agentOwnerIdentity: string`.
- The three `SheetOpenService` call sites now set these fields when building the command, instead of calling `sendAgentActive` separately (removed those calls and their try/catch blocks).
- On the Angular client, `composer-main.component.ts`'s `editAsset` subscription handler now calls a new `applyAgentActive(command)` helper right after `openViewsheet`/`openWorksheet` — since those methods synchronously create and push the new sheet's model object, the helper looks it up by `runtimeId` and sets `agentConnected`/`agentOwnerIdentity` immediately, with no race window at all.

## Test Plan

- [x] Java: `SheetOpenServiceTest` — 37/37 pass, updated to assert `OpenComposerAssetCommand.agentActive()`/`agentOwnerIdentity()` instead of a separate `sendAgentActive` call for all three entry points.
- [x] Frontend: new test group in `composer-main.component.interaction.tl.spec.ts` drives a real RxJS `Subject` through the mocked `composerClient.editAsset` to exercise the actual subscription → `openViewsheet`/`openWorksheet` → `applyAgentActive` path; 44/44 pass in that file.
- [x] `tsc --noEmit` clean on both `tsconfig.app.json` and `tsconfig.spec-tl.json`.
- [x] Live, against a from-source server: created a new viewsheet via `create_viewsheet` — tab-bar agent-connected indicator now shows immediately, confirmed visually by the reporter (it did not show before this fix, and did after rebuilding both the Java server and the Angular frontend).
- [x] Code review (code-reviewer agent): no issues found at confidence ≥ 80.

Note: `create_worksheet`'s new tab currently doesn't open in the browser at all due to a separate, already-filed defect (#76738 — null `assetId` crash), unrelated to this fix and not fixed here; the underlying `agentActive` mechanism for that path is identical and covered by the same passing unit tests.

Fixes Redmine #76737.


jshobe-inetsoft (approved):

jshobe-inetsoft (changes requested):
Resolve the merge conflicts.

jshobe-inetsoft (approved):
# Review: stylebi PR #5314 — Bug #76737, fix missing agent-connected indicator on agent-created sheets

Base branch: `stylebi-wiz-main-rebase` · Head: `bug-76737` (single commit `d8958ec`)

## Verdict

**Ready to merge** (as of the current head commit `d8958ec`). No blocking issues found.

Supporting facts:
- No merge conflicts: `gh pr view` reports `mergeable: MERGEABLE`. The prior "Resolve the merge conflicts" review round (requesting changes) appears to have been addressed by the new commit — the PR head is not showing as `CONFLICTING`/`DIRTY`.
- `mergeStateStatus` is `UNSTABLE` only because the `claude-review` GitHub Action check failed. Per the review criteria given for this task, `claude-review` is explicitly excluded from consideration since this manual review supersedes it. Inspecting that run shows it is an **infrastructure failure**, not a code finding: `Action failed with error: Claude execution failed: result is_error:true` — the automated reviewer itself errored out before producing a review, it did not report a code problem.
- All other checks pass: Semgrep Scan, Semgrep OSS, Validation Build (`build`), and `license/cla` are all green.
- The fix is narrowly scoped, logically sound, backed by both Java unit tests and new Angular TL tests that exercise the actual regression path, and does not touch unrelated code.

Two non-blocking items are called out below for the author's awareness; neither should hold up the merge.

---

## Summary of the change

`SheetOpenService`'s three tool entry points that mint a brand-new runtime and attach an agent session to it (`open_base_worksheet`, `create_viewsheet`, `create_worksheet`) used to call `SheetAgentBroadcastService.sendAgentActive()` *before* sending `OpenComposerAssetCommand`. `sendAgentActive` publishes `SetAgentActiveCommand` on the per-runtime `CommandDispatcher.COMMANDS_TOPIC`, which the browser only subscribes to once it has actually opened that runtime — and it's `OpenComposerAssetCommand` that causes the browser to open it. So the notification was always sent before anyone could be listening, and was silently dropped.

The fix carries the state (`agentActive: boolean`, `agentOwnerIdentity: string`) directly on `OpenComposerAssetCommand` itself, which is guaranteed to be delivered (it rides the always-subscribed `/user/composer-client` channel via `sendToComposer`). On the Angular side, `composer-main.component.ts`'s `editAsset` handler now calls a new `applyAgentActive()` right after `openViewsheet`/`openWorksheet`, which synchronously create the `Sheet` model object and push it onto `this.sheets` — so the lookup-by-`runtimeId` inside `applyAgentActive` cannot race the model's own creation.

`SheetJoinService.join()` (the human-pairs-a-code flow) is untouched, correctly, since it always attaches to a runtime the browser already has open (already subscribed).

## Correctness review

- **Root cause analysis verified against source.** Read `SheetAgentBroadcastService.sendToBrowser` (used by `sendAgentActive`, filtered by `CommandDispatcher.RUNTIME_ID_ATTR` on the per-runtime topic) vs. `sendToComposer` (addressed to `ComposerClientService.COMMANDS_TOPIC`/`/user/composer-client`, which `composer-main.component.ts` subscribes to unconditionally in `initComposerClient()`). The PR's stated mechanism is accurate — `sendToComposer` is the only one of the two guaranteed to have a subscriber when a runtime is brand-new.
- **No leftover call sites.** Confirmed `sendAgentActive` is only still called from `SheetJoinService.join()` (unmodified, and correctly so) — it is not dead code, and no site was missed.
- **No race in the Angular fix.** `openViewsheet`/`openWorksheet` → `openSheet()` synchronously `push`es the new `Viewsheet`/`Worksheet` onto `this.sheets` before returning; `applyAgentActive(command)` runs immediately afterward in the same call stack/tick and finds the sheet by `runtimeId`, which was set from the same `command.runtimeId` just used to open it. Verified `Worksheet`/`Viewsheet` don't shadow `Sheet.agentConnected`/`agentOwnerIdentity`, and `VSPane`/`WSPane`'s `ngOnInit` doesn't reset those fields — so the optimistic client-side set survives pane construction.
- **No accidental turn-on for human-driven opens.** `ComposerController.java` has three other `OpenComposerAssetCommand.builder()` call sites (the ordinary "Edit Data"/"Edit Table" human flows) that don't set `agentActive`; it correctly defaults to `false` via `@Value.Default`, and `applyAgentActive` no-ops on a falsy value. Covered by the third new TL test.
- **Error-handling change is not a regression.** The two removed `try/catch` blocks wrapped only the now-deleted `sendAgentActive` calls. The `sendToComposer(...)` call that now also carries the agent-active fields was already unguarded before this PR (same throw-on-failure behavior it always had) — so overall resilience is unchanged, and the change actually removes one extra point of failure (one broadcast call instead of two) rather than adding one.
- **Java/TS type parity.** `agentActive(): boolean` defaults to `false` via `@Value.Default` (matches the existing `wsWizard()` pattern); `agentOwnerIdentity()` is `@Nullable` (matches `parentId()`). The TypeScript `OpenComposerAssetCommand` interface mirrors both as optional fields. Semantically consistent with the value `SetAgentActiveCommand.ownerIdentity()` always carried.
- **Tests are substantive, not just refactored to pass.** The Java test rewrites replace `verify(broadcast).sendAgentActive(...)` assertions with `verify(broadcast).sendToComposer(...)` + assertions on `sent.agentActive()`/`sent.agentOwnerIdentity()`, exercising the actual new code path (all three entry points). The removed `tabBarNotifyFailureDoesNot...` tests are correctly deleted since there is no longer a separate best-effort call to fail. The new Angular TL tests drive a real `rxjs.Subject` through the mocked `composerClient.editAsset` to exercise the real `subscribe` → `openViewsheet`/`openWorksheet` → `applyAgentActive` path end-to-end (not just unit-testing `applyAgentActive` in isolation), including the negative case (command without `agentActive`). This is a meaningfully better regression guard than testing the helper alone.

## Non-blocking observations

1. **Stale doc comment left inconsistent by this PR.** `web/projects/portal/src/app/composer/data/sheet.ts` (not touched by this diff) documents `agentConnected` as: *"driven by SetAgentActiveCommand... never set optimistically client-side."* That is no longer true — `applyAgentActive()` now sets it directly from `OpenComposerAssetCommand`, without any `SetAgentActiveCommand` round-trip, which is exactly the point of the fix. Worth a one-line follow-up to that comment so a future reader isn't misled into thinking `applyAgentActive`'s approach violates a documented invariant.
2. **PR description references Bug #76738 (null `assetId` crash) as still open for `create_worksheet`**, but the `SheetOpenService.java` code on this base branch already resolves the new worksheet's `AssetEntry` via `worksheetService.getWorksheet(runtimeId, browserUser).getEntry()` before building the command — `assetId` cannot be null on that path as currently written. This isn't code introduced by this PR (the comments/logic predate it), so it's not a functional issue in the diff, but the description's characterization of that entry point's current state looks out of date and may cause confusion for whoever picks up #76738 next.

## Security / backward compatibility

- No new security surface: `sendToComposer` addressing (session-scoped, same trust boundary as the pre-existing `ComposerController` "Edit Data" sends) is unchanged by this PR; only two extra fields ride the same already-established command/channel.
- No backward-incompatible change: additive optional fields on both the Java command (`@Value.Default`/`@Nullable`) and the TS interface (`?:`), so any other producer/consumer of `OpenComposerAssetCommand` that doesn't know about the new fields continues to work exactly as before.

PR #5315, opened 2026-09-16
Bug #76709 (recyclebin): gate applyRestore's mutation flag on RecycleUtils' own pre-mutation lookups

ngovus-inetsoft:
Follow-up to #5258 (bug #76672), which threaded a per-item `AtomicBoolean mutationEntered` through `RecycleBinChangesetApplyService` so that a throw occurring *before* an item was ever mutated is reported as a clean `rolled-back` instead of `rollback-failed` ("server may be partially changed, do not retry"). That fix covered `applyPurge`'s three branches but not its sibling verb.

## The defect

`applyRestore` set `mutationEntered.set(true)` immediately after computing `beforeProjection`, before dispatching into `RecycleUtils.restoreSheet` / `restoreWSFolder` / `restoreRepositoryFolder`. Each of those opens with its own source lookup that throws strictly before that method's own first mutating statement:

| Branch | Pre-mutation throw points | First mutating call |
|---|---|---|
| `restoreSheet` | `getSheetEntry`, then the type-qualified re-read | `validatePath` |
| `restoreWSFolder` | `getAssetEntry`, then `getFolderAssetEntry` | `validatePath` |
| `restoreRepositoryFolder` | `getRegistry` | `checkParentFolderExist` |

So a restore whose entry lost its backing asset to a concurrent purge between preview and apply was reported as an unrecoverable failure needing manual escalation, even though nothing was touched.

## The fix

Adds `RecycleBinService.requireRestorableSource` — the source-side twin of the existing `wouldCollide`, which already exists to mirror `RecycleUtils`' internal checks so callers can refuse *before* the real mutating restore. `applyRestore` calls it and only then sets `mutationEntered`.

Two things worth flagging for review:

- **The gate cannot be pushed deeper into `RecycleUtils`.** `validatePath` and `checkParentFolderExist` — the very next statements in those methods — already mutate (`repository.addFolder` / `registry.addFolder`), so they can't be part of a pre-mutation gate. This is also why `RecycleUtils` is left untouched, consistent with #76672's own decision to descope changes to that file.
- **`RecycleUtils` uses two different scope discriminators, and the mirror follows each exactly.** The sheet path keys off `SUtil.isMyDashboard(path)` (`restoreSheet` prepends the prefix itself when the entry was user-scoped); the worksheet-folder path keys off `getOriginalUser() != null`, because a recycled folder's trash path is a bare `"Recycle Bin/<uuid>"` with no prefix (`moveAssetFolderToRecycleBin` → `getRecycleBinPath`). Keying the folder branch off the path instead would refuse every user-scoped folder restore outright. The mirror also reproduces `restoreSheet`'s second, type-qualified re-read, so a My Dashboards dashboard resolving only through the global `VIEWSHEET_SNAPSHOT` fallback isn't waved through the gate and then misreported.

The repository-folder branch overlaps a `getRegistry` call `wouldCollide` already makes a few lines earlier on this path; it is kept so the method's contract holds for all three entry kinds rather than silently no-opping for one, and the comment says so.

## Tests

- `RecycleBinChangesetApplyServiceTest` (+3): two gate cases, plus a guard proving a throw out of the restore *itself* still yields `rollback-failed` — the gate was narrowed, not removed.
- New `RecycleBinServiceTest` (10): pins the mirror to what `RecycleUtils` actually does, including the user-scoped folder and snapshot-re-read cases above.

Verified red-then-green: with the gate line removed, the two new apply tests fail with `expected: <rolled-back> but was: <rollback-failed>`; with the scope discriminator and re-read mirroring reverted, the two corresponding service tests fail. All green with the fix — 30/30 across `RecycleBinServiceTest`, `RecycleBinChangesetApplyServiceTest`, `RepositoryRecycleBinControllerTest`, `RecycleBinOrgLifecycleTest`.

## Out of scope

Found while tracing, deliberately not fixed here and worth its own ticket: when a restore or purge mutation *succeeds* but `verified` comes back false, the loop breaks with `failed = true` but adds nothing to `undoable` or `unknownStateFailures` — so a batch whose asset may be half-moved is reported as a clean `rolled-back`. Same misreporting family, opposite and more dangerous direction.


jshobe-inetsoft (approved):

PR #5316, opened 2026-09-16
Bug #76738, fix create_worksheet's new tab never opening in the browser

jkidacki-inetsoft:
## Summary

- `create_worksheet`'s new worksheet runtime was genuinely functional server-side, but its tab never opened in the browser's Composer window at all.
- Live debugging with the reporter (iteratively adding diagnostic logging, reading server logs, and testing against a real from-source server) uncovered two layered defects, both fixed here.

## Root Cause

**Layer 1 — client-side crash:** `create_worksheet`'s `OpenComposerAssetCommand` carries `assetId: null` (an unsaved worksheet has no real asset path yet). `createAssetEntry(null)` returns `null` without throwing, but `ComposerRecentService.addRecentlyViewed(entry)` then dereferenced `entry.path` with no null-check, inside a loop over the existing recently-viewed list. If that list was non-empty (true for any session with prior activity), it threw, and the exception propagated out of the `editAsset` STOMP subscription callback in `composer-main.component.ts` — silently skipping the `openWorksheet(...)` call right after it.

**Layer 2 — server-side deserialization failure (found after fixing layer 1):** once the crash was fixed, the tab appeared but a new error surfaced: `POST /api/ws/open 400 (HttpMessageNotReadableException)`. `OpenWorksheetEvent.id()` is a non-`@Nullable` Immutables field, so a JSON payload with `"id": null` fails to even deserialize server-side.

Rather than patch nullability through the whole downstream chain (which also dereferences the entry unconditionally in several places), the actual fix is one level up: `WorksheetEngine.openTemporaryWorksheet` already falls back to generating a real `(TEMPORARY_SCOPE, "Untitled-N")` `AssetEntry` internally when handed a `null` entry — exactly mirroring `openTemporaryViewsheet`'s own fallback for viewsheets. `SheetOpenService.createViewsheet` already fetches this real temp entry back and uses it as `assetId`; `createWorksheet` never did the equivalent, and instead sent a literal `null` on the (incorrect) assumption that a freshly-minted worksheet has no identity at all.

## Fix

1. `composer-recent.service.ts` — `addRecentlyViewed` now returns early on a `null` entry, mirroring the existing null-list guard one line above.
2. `SheetOpenService.createWorksheet` — fetches the runtime's own real entry (`worksheetService.getWorksheet(runtimeId, user).getEntry()`) and uses its identifier as `assetId`, instead of `null`. This eliminates the null at the source rather than threading nullability through every downstream consumer.

## Test Plan

- [x] Unit: new `composer-recent.service.spec.ts` (3 tests) — null entry doesn't throw, real entry still adds normally, dedup-by-path still works.
- [x] Unit: `SheetOpenServiceTest` — 41/41 pass, new `pushesTheRuntimesOwnRealTempEntryAsAssetIdNotNull` test asserts the broadcast command's `assetId` is the real temp entry identifier, never `null`.
- [x] Full `test:portal` suite: 378/378 pass, no regressions.
- [x] `tsc --noEmit` clean.
- [x] Live, against a from-source server (restarted + rebuilt with the fix): `create_worksheet` now opens a working worksheet tab with no client-side crash and no server-side error, confirmed directly by the reporter.
- [x] Code review (code-reviewer agent): approved, no changes requested — independently confirmed via `EngineConfiguration.java` that `worksheetService`/`viewsheetService` are the same Spring-wired singleton (by design, so runtime sessions share one cache), verifying the fix's core assumption.

Fixes Redmine #76738.


jshobe-inetsoft (approved):

PR #5317, opened 2026-09-16
Bug #76705: apply theme/locale/member changes on an org rename

ngovus-inetsoft:
## Problem

`IdentityService.setOrganizationInfo` had an early-return branch that fired whenever an organization's display name changed but its id did not — a plain rename. It wrote the new name, persisted, and returned, **before** reaching the method's own member sync (`updateOrganizationMembers`), locale reverse-lookup, and theme sync (`updateCustomThemeOrganization`).

The EM sends all of those fields in a **single** request: `edit-identity-view.component.ts` `updateModel()` fills `name`, `theme`, `roles`, `members`, then org-only `id`/`locale`, and `apply()` emits that one model, which `users-settings-page.component.ts` POSTs once. So "rename this organization **and** give it a different theme" is a normal one-click EM operation, and everything except the rename was silently dropped — no error, no warning. The public REST path (`SecurityApiService.updateOrganization`) builds the same shape and hit the same branch.

Found during the AI verification pass on #76671, which confirmed it as real but deferred it as out of scope there.

## Fix

Delete the early return, so an id-preserving rename takes the same path as every other same-id edit:

- **The same-id branch now applies the new display name.** That branch persists `fromOrg` — the *stored* organization, which still carries the old name — so without this the rename itself would be lost.
- **The org-id migration block is re-gated on the organization id** instead of `!Tool.equals(fromOrg, newOrg)`. `Organization.equals` compares name *and* id, so an id-preserving rename looks "changed" to it, and without this change a rename would newly trigger `dashboardRegistryManager.migrateRegistry`, a replet-registry `shutdown()`, and `updateOrgScopedDataSpace` — machinery it never reached before. All three are id-migration operations and are no-ops when the id is unchanged (`updateOrgScopedDataSpace` rewrites paths via `path.replace(oorg.getId(), norg.getId())`; `migrateRegistry` short-circuits on its `changeId` flag).

`syncIdentity` stays unreachable for a same-id rename, which is intended: for an organization it routes into `eprovider.copyOrganization(...)`, a full org copy that must not run on a rename-in-place. Rename permission/grant rewriting is already handled by the caller (`UserTreeService.editOrganization` → `handleOrgNameChangePermittedIdentities` + `setIdentityPermissions`).

## Tests

Three tests added to `IdentityServiceTest`:

- `setOrganizationInfo_renameWithThemeChange_appliesThemeAndPointer` — asserts the theme is persisted **and** `CustomThemesManager.setOrgSelectedTheme` is called (the pointer `CustomThemesImpl.getUserTheme` actually gates on; list membership alone would not serve the theme).
- `setOrganizationInfo_renameWithLocaleChange_appliesLocale` — asserts the label→code reverse lookup lands on the stored org.
- `setOrganizationInfo_renameOnly_stillRenamesAndSkipsIdMigration` — regression guard: a name-only change still renames, and touches neither registry manager nor the dataspace.

Verified the first two **fail** without the source change (`expected: <theme-1> but was: <null>`, `expected: <en_US> but was: <>`) and pass with it; the third passes either way, as intended.

`IdentityServiceTest` 17/17 green, plus `DashboardRegistryOrgLifecycleTest`, `OrgLifecycleDataSpaceIntegrationTest`, `OrgLifecycleScopedPropertiesIntegrationTest` and `PermissionMatrixOrgLifecycleTest` (the existing org-id-change coverage the early return used to shield) all still green.

## Noted, not fixed here

Pre-existing and separate from this defect — via the public REST API, a partial update that omits the member lists sends an empty list (`EntityModel.members()` defaults to `[]`) and `updateOrganizationMembers` deletes the users absent from it. That already applies to every non-rename organization update today; this change makes renames behave the same way rather than introducing a new class of risk. Worth its own issue if that API is meant to be PATCH-like. Same shape for `locale()` defaulting to `""`.

Redmine: #76705


jshobe-inetsoft (approved):
Approving based on Claude's completed automated review, which found the fix correct and precisely scoped, with no blocking issues (one non-blocking follow-up noted). All checks pass and there are no merge conflicts.

PR #5318, opened 2026-09-17
fix(76721): add Maven-coordinate upload path to admin-chat plugin endpoint

Nelson-cao:
## Summary

`upload_driver_or_plugin`'s only backing StyleBI endpoint (`POST /api/wiz/v1/admin/plugins/upload`) accepts a local file only. The EM UI's Maven-coordinate resolve path (`POST /api/em/upload/maven`) is not mapped under `/api/wiz/**`, so an admin-chat bearer-token request can never reach it, regardless of what the plugin tool sends.

- Add `AdminPluginService.uploadMaven(String gav, Principal)`, reusing the existing injected `UploadService.add(String)` and `requireUploadDriversPermission` check (no new Spring wiring).
- Add `POST /api/wiz/v1/admin/plugins/upload/maven` in `AdminPluginController`, mirroring the existing local-file upload endpoint's structure, taking `{ gav }` (reusing `MavenUploadRequest`).
- Response is `{ uploadId, fileNames: string[] }` — a list, not a singular `fileName`, since a resolved Maven GAV's transitive dependencies can produce more than one jar (`UploadService.add(String)` already returns `UploadFilesResponse.files(): List<String>`).
- An unresolvable GAV (`FileNotFoundException` from `UploadService.add`) now surfaces as a clear, field-named 400 via a new `@ExceptionHandler`, instead of a generic wrapped error.

This is the StyleBI-side half of bug #76721; a companion plugin-side change (extending the `upload_driver_or_plugin` MCP tool with a `mavenCoord` alias) lands separately in `stylebi-wiz`.

**Holding for human review — do not merge/auto-merge.**

## Test plan
- [x] `./mvnw -q -pl core -am -Dtest="AdminPluginServiceTest" -Dsurefire.failIfNoSpecifiedTests=false -DfailIfNoTests=false test` — 23/23 pass (20 existing + 3 new `uploadMaven` tests)


Nelson-cao:
Finding #1 (open across both prior reviews) fixed in 5bc8811f9: `AdminPluginService.uploadMaven` now validates `gav` (non-null, non-blank, exactly 3 `:`-separated segments) before calling `uploadService.add(gav)`, throwing a field-named `IllegalArgumentException` that `AdminPluginController`'s existing `handleIllegalArgument` handler already maps to a clean 400 — matching the same shape/message convention as `upload()`'s file-extension check. Confirmed independently (read `MavenClientService.resolve` and `AdminPluginController`'s three exception handlers directly) that a null/blank/malformed gav previously threw an uncaught NPE/ArrayIndexOutOfBoundsException.

Added `uploadMavenRejectsAMalformedGavWithoutCallingUploadService` (6 bad-input cases: null, empty, blank, 0/1/3 colons) to `AdminPluginServiceTest`, asserting the exception and that `uploadService` is never invoked.

Re-ran scoped tests fresh: `AdminPluginServiceTest` 24/24 pass, `AdminPluginControllerTest` 14/14 pass (no regression from the added validation).

Finding #3 (minor, non-blocking, controller-wide `FileNotFoundException` handler message) left as-is per the review's own note that it's "not a live bug today" — no other endpoint in this controller currently routes through that overload.

Nelson-cao:
## Round 2 review (this is this PR's first review round actually posted to GitHub — an earlier internal pass by `reviewer-76721` found no findings but was never posted; several follow-up commits landed since, some from the CI `claude-review` bot, applied directly rather than through a dispatched fixer)

**Verdict: changes requested** (posted as a comment, not a formal review — `gh pr review` is rejected as self-review since this PR is authored under the same account running this review).

**Head commit reviewed:** `5bc8811f9` — full diff `origin/stylebi-wiz-main-rebase...HEAD` (touches `AdminPluginController.java`, `AdminPluginService.java`, `AdminPluginServiceTest.java`, `AdminPluginControllerTest.java`).

### What was checked
- Re-read the full accumulated diff against the diagnosis/refutation docs, not just the latest commit.
- Independently re-ran the targeted suites (not trusting prior reports): `AdminPluginServiceTest` **24/24 pass**, `AdminPluginControllerTest` **14/14 pass** (both via `./mvnw -q -pl core -am -Dtest=... test` from `community/`).
- Specifically audited `5bc8811f9`'s new `uploadMaven` gav guard (`gav == null || gav.isBlank() || gav.split(":").length != 3`) against boundary/shape variants beyond the 6 tested inputs — wrote and ran a throwaway `String.split` probe to confirm actual Java semantics rather than assume them.
- Confirmed the controller-layer tests added in `4f1555fef` (`uploadMavenDelegatesToService`, `uploadMavenThrowsForbiddenForNonSiteAdminBeforeReachingService`, `handleFileNotFoundReturnsFailedStatusWithMessage`) correctly mirror the file's existing per-endpoint pattern.

### Status of prior findings (from the CI `claude-review` bot rounds, per `07-review-followup-5318.md`)
- Finding #1 ("`uploadMaven` doesn't validate `gav` shape") — **partially fixed, see new finding below**: a length-of-3 check was added, but it doesn't reject blank segments, which is the same class of defect one level down.
- Finding #2 (controller-layer test coverage gap) — **fixed and confirmed**: `AdminPluginControllerTest` now has the forbidden/delegates/handler tests, all passing.
- Finding #3 (controller-wide `FileNotFoundException` handler message could mislead for an unrelated future caller) — **left as-is, acceptable**: confirmed still true today (no other `FileNotFoundException`-throwing `UploadService.add` overload is reachable through this controller), left as a comment for whoever extends this controller next. Not blocking.

### New finding — Important (≈82/100): `uploadMaven`'s gav guard admits a blank segment

`AdminPluginService.java:109`: `gav.split(":").length != 3` only checks the segment **count**, not that each segment is non-blank. Verified directly (`"a::c".split(":")` and `":b:c".split(":")` both yield length-3 arrays in Java, e.g. `["a", "", "c"]` / `["", "b", "c"]` — confirmed by compiling and running a standalone probe, not just reasoning about it). So a gav like `"org.postgresql::42.7.3"` or `":postgresql:42.7.3"` **passes this guard** and flows unchanged into `MavenClientService.resolve` (untouched by this PR), which does `args.put("module", coords[1])` (or `"group"`) with an empty string and calls `Grape.resolve(args, args)`. That is very likely to throw an exception type none of the three `@ExceptionHandler`s here catch (`IllegalArgumentException`/`FileNotFoundException`/`SecurityException`), surfacing as a generic 500 instead of the clean, field-named 400 this guard's own commit message promises to guarantee.

The regression test added for this guard, `uploadMavenRejectsAMalformedGavWithoutCallingUploadService` (6 cases: `null`, `""`, `"  "`, 1-segment, 2-segment, 4-segment), does not cover this shape at all — it only varies segment *count*, never blankness of a present segment. So the test's implied claim ("rejects a malformed gav") is narrower than reality.

Note this is directly the sibling of a case this same fix already got right elsewhere: PR #2613's `requireMavenCoord` (plugin side) already validates `parts.some(part => part === "")` after splitting — the identical scenario, already handled correctly on the TypeScript side. The Java guard just needs the same check added, e.g. `Arrays.stream(gav.split(":", -1)).anyMatch(String::isBlank)` (use limit `-1` so a trailing-empty segment is caught explicitly too, rather than relying on the default trailing-empty-drop behavior of `split(String)`), plus 2 more test cases (`"org.postgresql::42.7.3"`, `":postgresql:42.7.3"`).

### Build CI status
`build` (full Maven validation) was still `pending` at review time — not a finding, just noting it should be confirmed green before merge regardless of this review's outcome.

**This PR is held for an explicit human merge decision regardless of review outcome — I have not merged, approved-and-auto-merged, or enabled auto-merge, and won't.**

Nelson-cao:
Fixed the blank-segment gap from the round-2 review (`05-review-r2.md`).

Root cause: `gav.split(":")` (no limit, defaults to 0) drops *trailing* empty strings but keeps leading/interior ones, so `"a::c"` → `["a","","c"]` and `":b:c"` → `["","b","c"]` both had length 3 and slipped past the `length != 3` guard, reaching `MavenClientService.resolve`/`Grape.resolve` with an empty group or artifact — an uncaught exception instead of the clean 400 the guard is meant to guarantee.

Fix (`AdminPluginService.java`): split with limit `-1` (keeps trailing empties too) and reject if any of the 3 segments is blank, in addition to the length check.

Added 3 regression cases to `uploadMavenRejectsAMalformedGavWithoutCallingUploadService`: `"org.postgresql::42.7.3"`, `":postgresql:42.7.3"`, `"org.postgresql:postgresql:"` — each asserts the `IllegalArgumentException` and `verifyNoInteractions(uploadService)`.

Re-ran `AdminPluginServiceTest` + `AdminPluginControllerTest` (`./mvnw -pl core -am -Dtest=... test` from `community/`): 24/24 and 14/14 pass, including the new cases.

Held for human merge decision, not merged/auto-merge.

Nelson-cao:
## Round 3 review — `AdminPluginService.uploadMaven` gav validation

**Head SHA reviewed:** `257bd5903` (community submodule) — SHA range since round 2: `5bc8811f9..257bd5903`.

**Note:** `gh pr review --approve` fails on this PR with "Can not approve your own pull request" (same account owns the PR, as in round 2) — posting the verdict as a comment instead, per the workflow's documented fallback.

### What was checked

1. **Round-2 finding status: FIXED.** `AdminPluginService.java:114-119` now does:
   ```java
   String[] gavParts = gav.split(":", -1);
   if(gavParts.length != 3 || Arrays.stream(gavParts).anyMatch(String::isBlank)) { ... }
   ```
   - `split(":", -1)` correctly keeps trailing empty strings (confirmed against `String.split` javadoc semantics: a negative limit applies the pattern as many times as possible and does not discard trailing empty strings — the only reason the previous no-arg `split(":")` call under-counted `"a:b:"`-shaped input in the first place).
   - Walked every malformed shape asked for: `"a::c"` → `["a","","c"]`, length 3, blank segment caught. `":b:c"` → `["","b","c"]`, blank segment caught. `"a:b:"` → `["a","b",""]`, blank segment caught. `"a:b:c:"` → `["a","b","c",""]`, length 4, rejected by the length check. A genuinely valid `"org.postgresql:postgresql:42.7.3"` → 3 non-blank parts, accepted.
   - Tried to break it further: a 4-segment classifier-style gav (`"group:artifact:1.0:sources"`) is correctly rejected — out of scope for this endpoint's `group:artifact:version` contract, not a regression. The one edge case that does NOT get caught by `String::isBlank` is a segment made of only U+00A0 (non-breaking space) or similar codepoints `Character.isWhitespace` doesn't recognize (NBSP is `isSpaceChar` but not `isWhitespace` in Java) — e.g. `"a:\u00A0:c"` would pass the guard. Checked impact: `UploadService.add(gav)` passes the raw string straight to `MavenClientService.resolve(gav)`, which re-splits and attempts Maven resolution — a bogus artifact id (NBSP or otherwise) still fails there with the same "not found" outcome as any other unresolvable coordinate, not a 500 or silent success. No behavioral regression, so not raising this as a blocking finding (sub-80 confidence, cosmetic at most).
2. **Test coverage confirmed real.** `AdminPluginServiceTest.uploadMavenRejectsAMalformedGavWithoutCallingUploadService` (lines 174-190) now includes `"org.postgresql::42.7.3"`, `":postgresql:42.7.3"`, and `"org.postgresql:postgresql:"` alongside the pre-existing shape checks. Mentally reverted to the round-1/round-2 guard (`gav.split(":").length != 3`, no per-segment blank check): the no-arg `split(":")` still returns length-3 arrays for `"org.postgresql::42.7.3"` and `":postgresql:42.7.3"` (interior/leading empties are kept by default split, only trailing ones are dropped) — so both would have silently passed the old guard. The new tests would fail against that old code, confirming they're not tautological.
3. **Same-pattern sweep.** Grepped the community tree for other `split(":")` gav/coordinate parsing. `MavenClientService.java:90` (`gav.split(":")`, no validation) is the pre-existing internal callee this guard protects against — already known and explicitly called out as shared/out-of-scope in the round-2 fix commit message (also reached by the UI's own `/api/em/upload/maven`, unaffected by this change either way). No other new instance of this pattern introduced by this diff.
4. **Tests re-run fresh** (not trusting the fixer's numbers): `mvn -pl core -am test -Dtest=AdminPluginServiceTest,AdminPluginControllerTest` → `AdminPluginServiceTest`: 24/24 passed. `AdminPluginControllerTest`: 14/14 passed.
5. **CI `build` check:** still `pending` at review time (other checks — Semgrep OSS, Semgrep Scan, claude-review, license/cla — all `pass`). Not blocking this verdict since the scoped local Maven run covers the changed code directly; flagging for whoever makes the merge decision to confirm `build` goes green before merge.

### Verdict: **Approved** (posted as comment, not a GitHub review approval, due to the self-review restriction noted above)

No findings this round. **Not merging or enabling auto-merge** — this is held for a human merge decision, and the `build` check should be confirmed green first.

fredaan (approved):

PR #5320, opened 2026-09-17
fix(wiz): create_worksheet must open its runtime as the browser's principal, not the agent's (Redmine #76636)

lizhepu:
## Summary

- Reopened bug #76636's "browser doesn't follow along" finding is a second, sibling instance of a bug class already fixed once for `createViewsheet`.
- `SheetOpenService.createWorksheet` (`core/src/main/java/inetsoft/web/wiz/viewsheet/SheetOpenService.java`) opened its new worksheet runtime with the raw agent `Principal` instead of the connected browser session's principal, unlike `openBaseWorksheet` (`rvs.getUser()`) and `createViewsheet` (`actingSheet`/`browserUser`, fixed in commit `7f45a3ade` / PR #5219). `createWorksheet` was added later (`a9469ece3` / PR #5002) and was never given the same fix, so PR #5219 never touched it.
- `WorksheetEngine`/`ViewsheetEngine`'s ownership check (`RuntimeSheet.matches` → `SRPrincipal.equals`/`ClientInfo.equals`) requires full `Principal` equality including session id. A runtime minted under the agent's own JWT-rebuilt principal is never equal to the browser's live-session principal, so the browser's later attach to it (directly, or to a viewsheet created from it via `create_viewsheet`) fails with `InvalidUserException` ("Invalid user found: ... instead of ..."). Live-reproduced this run, end to end, against a local StyleBI instance.

## Fix

`createWorksheet` now resolves the acting session's `RuntimeSheet` via `runtimeAccess.getSheetForPairing(actingSession.sheetType(), actingSession.runtimeId(), user)` and opens the new runtime as `actingSheet.getUser()` when available, mirroring `createViewsheet`'s pattern.

For the `PairingException` handling, `createWorksheet` differs from `createViewsheet` in one respect worth calling out: `createViewsheet` tolerates `SESSION_EXPIRED` only when an explicit `dataSource` is given (because defaulting `dataSource` from the acting worksheet genuinely needs the acting runtime to be alive) and fails loud when defaulting. `createWorksheet` has no `dataSource` concept at all — the new worksheet is always blank — so `actingSheet` is used here *only* to resolve the browser principal, never for content. There is no "genuinely needs it" branch to fail loud on, so `createWorksheet` unconditionally tolerates `SESSION_EXPIRED` (falls back to the agent's own principal) and still propagates every other `PairingException` kind (bad argument, user mismatch, feature disabled, rate limited, internal error), since those name a real problem with the request itself.

**Rebase reconciliation with concurrently-merged Bug #76738 (PR #5316):** while this fix was in flight, an unrelated fix for Bug #76738 landed on `stylebi-wiz-main-rebase`, touching the same method — it fetches the runtime's own `AssetEntry` right after minting it (`worksheetService.getWorksheet(runtimeId, user).getEntry()`) so `OpenComposerAssetCommand.assetId` carries a real identifier instead of `null`. Rebased onto that commit and updated the fetch to use `browserUser` instead of the raw agent `user` — the runtime is now registered under `browserUser`, so fetching it back with the agent's principal would hit this same bug's own mechanism at that call site (it would only have worked by accident, via the `pairedAgent` bypass flag `getSheetForPairing` incidentally sets on the agent principal while resolving the acting session). This matches `createViewsheet`'s own already-established pattern (`viewsheetService.getViewsheet(runtimeId, browserUser)`).

## Tests

`SheetOpenServiceTest`'s `createWorksheet` block previously had no assertion that could distinguish "opened with the agent's principal" from "opened with the browser/acting-session's principal" — the one relevant assertion used a loose `any(Principal.class)` matcher. Added:

- `createWorksheetOpensTheRuntimeAsTheActingSessionsBrowserPrincipalNotTheAgents` — stubs the acting session's `RuntimeSheet` with a distinct browser principal and asserts `openTemporaryWorksheet` is called with *that* principal, not the agent's.
- `createWorksheetToleratesExpiredActingRuntimeAndFallsBackToTheAgentPrincipal` — `getSheetForPairing` throws `PairingException(SESSION_EXPIRED)`; asserts the create still succeeds, falling back to the agent's own principal. Also asserts `getSheetForPairing` was actually invoked (review round 1 caught that without this, the test passed identically against the pre-fix method too, since it never calls `getSheetForPairing` at all).
- `createWorksheetPropagatesANonSessionExpiredPairingException` — `getSheetForPairing` throws `PairingException(USER_MISMATCH)`; asserts it propagates and no runtime is opened.
- Extended Bug #76738's own `AssetEntry`-fetch test with a captor-based assertion that the fetch now runs under the browser principal, not the agent's.

Also updated the `createWorksheetService` test harness to stub `runtimeAccess.getSheetForPairing` with a distinct browser principal (previously unstubbed, so it silently returned `null` and every existing test passed by accident regardless of which principal the fix used).

`cd core && ./mvnw -o -pl core test -Dtest=SheetOpenServiceTest -Dsurefire.failIfNoSpecifiedTests=false` → 44/44 pass.

## Test plan
- [x] `SheetOpenServiceTest` (44/44 pass, unit level)
- [ ] Live browser re-verification (tester phase, out of scope for this PR per the debug-workflow's fixer role)


lizhepu:
## Adversarial review (independent pass, bug 76636 reopened)

Read the diagnosis/refute/fix archive first, then reviewed this diff against the confirmed mechanism (agent-vs-browser Principal identity feeding `RuntimeSheet.matches`), not just the reported symptom.

**Core fix: correct.** `createWorksheet` now resolves the acting session's browser principal via `runtimeAccess.getSheetForPairing(...)` exactly like `createViewsheet` already does (PR #5219), instead of opening the new runtime under the raw agent principal. This corrects the general rule, not just this call site's symptom. Grepped the whole `inetsoft.web.wiz` package for every other method that mints a new runtime *and* reuses an acting session's live socket (`sheetSessions.open` + `OpenComposerAssetCommand`) — only `openBaseWorksheet`/`createViewsheet`/`createWorksheet` do both, and all three are now correct. Other runtime-minting call sites (`WorksheetTableService.openProbeWorksheet`, `WizAutoBindingService`, `WizDashboardService`, `WizVsService`, `LayoutSessionService.resolveClone`) never reuse an acting browser socket, so they're not instances of this bug class.

**Finding 1 (worth a follow-up, not blocking):** the `SESSION_EXPIRED` fallback (`actingSheet = null` → falls back to the agent's own principal) still calls `sheetSessions.open(...)` with the acting session's live socket and still pushes `OpenComposerAssetCommand` — i.e. in that narrow window (JoinSession still valid, underlying runtime cache separately expired) the runtime can end up agent-owned again while a browser socket is re-attached to it, reproducing this exact bug shape. This isn't new to this PR — `createViewsheet`'s already-merged fallback (#5219) has the identical shape when `dataSource` is explicit — so this PR is consistent with existing precedent rather than introducing a fresh regression, but neither PR calls out or closes this gap. Suggest a follow-up ticket covering both methods.

**Finding 2 (test-quality nitpick):** verified empirically by running the new tests against the actual pre-fix source (isolated worktree at the PR's merge-base, `965b15568`, with only `SheetOpenServiceTest.java` overlaid): 41/43 pass, 2 fail.
- `createWorksheetOpensTheRuntimeAsTheActingSessionsBrowserPrincipalNotTheAgents` — fails pre-fix. Genuine regression test.
- `createWorksheetPropagatesANonSessionExpiredPairingException` — fails pre-fix. Genuine regression test.
- `createWorksheetToleratesExpiredActingRuntimeAndFallsBackToTheAgentPrincipal` — **passes pre-fix too** (pre-fix code never calls `getSheetForPairing`, so the stub is never exercised and the assertion is trivially true either way). It documents intended behavior but provides no regression protection.

**CI**: Semgrep OSS/Scan, claude-review, license/cla all pass; `build` was still `pending` at review time — please confirm it's green before merging.

**Verdict: clean, recommend merge pending user confirmation** (and pending the `build` check). Full writeup: `docs/teams/2026-09-17-bug-76636-reopened/bug-76636-viewsheet-handoff/05-review-r1.md` in stylebi-wiz.

lizhepu:
## Review round 2 (post-rebase reconciliation)

Re-reviewed after the rebase onto `stylebi-wiz-main-rebase` (now includes the concurrently-merged Bug #76738 fix) and the reconciliation commit.

- Confirmed the reconciliation is correct and general: `createWorksheet`'s `getWorksheet(runtimeId, ...)` fetch (added by #76738) now uses the resolved browser principal, not the raw agent principal — matching `createViewsheet`'s own established `getViewsheet(runtimeId, browserUser)` pattern. Independently reverted just that one line and confirmed the new regression test (`pushesTheRuntimesOwnRealTempEntryAsAssetIdNotNull`'s captor assertion) fails appropriately, then restored it and confirmed 44/44 green.
- Round 1's non-genuine `SESSION_EXPIRED`-tolerance test is fixed — confirmed the added `verify(runtimeAccess).getSheetForPairing(...))` makes it a real regression test now.
- Re-confirmed (fresh grep) no other call site in `inetsoft.web.wiz` shares this bug's shape — still exactly the three methods in `SheetOpenService.java`, all now correct.
- `build`/`claude-review`/Semgrep/CLA all green as of this comment.

**Recommend merge, pending the user's own confirmation** (not merging this myself, per this repo's standing convention).

Full record: `docs/teams/2026-09-17-bug-76636-reopened/bug-76636-viewsheet-handoff/07-verify-r2.md` and `08-review-r2.md`.

PR #5321, opened 2026-09-17
fix(76742-vts-004): support rename/move in WizTableStyleController.update

Zxm-happy:
## Summary
- `update_table_style`'s backend `UpdateTableStyleRequest` had no `name`/`folder` fields, so renaming or moving a Table Style through the wiz plugin required delete+recreate.
- `WizTableStyleController.update()` now accepts optional `name`/`folder`, validates them (no `LibManager.SEPARATOR`, mirroring `create`'s guard), duplicate-checks the resolved destination full name (`TableStyleLogicalLibrary.rename`/`LibManager.renameTableStyle` do no such check themselves), checks WRITE on the new full name, and — when the folder actually changes — checks WRITE on the destination folder too (mirroring `ChangeAssetController.changeTableStyle`'s `TABLE_STYLE_LIBRARY`/`TABLE_STYLE` resource-type split).
- Critically, after calling `lib.renameTableStyle(oldFullName, newFullName, styleId)` (which renames the *stored* entry), the local clone's name is also updated via `style.setName(newFullName)` before the method's pre-existing unconditional `lib.setTableStyle(style.getID(), style)` tail runs — without this, that tail silently persists the stale-named clone and undoes the rename on every single call.

## Test plan
- [x] Added regression coverage to `WizTableStyleControllerTest.java`: rename-only, folder-move-only, rename+folder together, duplicate-name-at-destination rejected, separator-in-name/folder rejected, and two tests specifically covering the "rename gets silently reverted by the tail write" regression (one within a single call, one across two sequential calls).
- [x] `mvn -pl core -o test -Dtest=WizTableStyleControllerTest` — `Tests run: 20, Failures: 0, Errors: 0, Skipped: 0`, `BUILD SUCCESS`.

Paired wiz-side PR (adds `name`/`folder` to `update_table_style`'s MCP tool schema): see companion PR in `inetsoft-technology/wiz`.


Zxm-happy:
## Review — VTS-004 (round 1)

Reviewed the actual diff and the full `update()` method (not just the hunk) via `git show`, checking against the refuter's confirmed findings in the wiz-side team doc (`docs/teams/2026-09-17-bugs-76742/bug-VTS-004/02-refute.md`), which established that the diagnosis's original ordering claim was wrong: the local `style` clone's stale name would silently undo the rename via the tail's unconditional `lib.setTableStyle(...)` call unless the clone's name is also updated.

**Ordering, verified**: confirmed the fix does exactly that — inside `if(!newFullName.equals(oldFullName))`: duplicate check → permission check (name, then folder if it changed) → `lib.renameTableStyle(...)` → `style.setName(newFullName)` on the local clone — all *before* the pre-existing unconditional format-merge + `lib.setTableStyle(...)` tail. Tried to construct a path where the rename runs but `setName` is skipped; found none — they're the last two statements of the same conditional block, and the only way to skip both is a true no-op (`newFullName == oldFullName`), which is correct.

**Duplicate-name check**: confirmed it runs before the rename call, and a dedicated test asserts `renameTableStyle`/`setTableStyle` are never invoked when the destination name collides.

**Destination-folder WRITE check**: confirmed it's implemented (`requireFolderWritePermission`) and gated on `!Objects.equals(newInternalFolder, oldInternalFolder)` — only fires when the folder actually changes, so existing format-only callers (who never enter the name/folder block at all) see no regression.

**Regression tests**: both required tests exist and assert what's claimed (single rename+format call persists the new name; two sequential calls — rename then format-only — the name survives both, with the second test explicitly simulating `renameTableStyle`'s real in-place mutation between calls). I tried a third sequence (rename, then a folder-only move with no format) looking for a gap the tests don't cover — traced through the code and couldn't construct a failure: both name and folder changes share the same code path, which always recomputes `newFullName` from a fresh `getTableStyle(styleId)` fetch rather than cached state, so it's structurally identical to the tested format-only-second-call case. Worth adding as an explicit test for completeness, but not a functional gap — non-blocking nit.

**Scope check**: confirmed via `gh pr diff --name-only` — only `WizTableStyleController.java` + its test file. No changes to `ChangeAssetController.java`, `RenameAssetController.java`, `LibManager.java`, or `TableStyleLogicalLibrary.java`, as required.

**Build verification — ran independently, not taken on faith**: `mvn -pl core -o test -Dtest=WizTableStyleControllerTest` in a clean worktree on this exact branch. Exit code checked directly: `0`. `Tests run: 20, Failures: 0, Errors: 0, Skipped: 0` / `BUILD SUCCESS`, matching the file's 20 `@Test` methods.

**No blockers, no important findings** — one non-blocking test-coverage nit (the third sequence above) and one non-blocking input-robustness nit shared with the wiz-side PR (explicit `null` for an omitted field slipping past `!== undefined` guards — no data corruption, just a wasted round-trip; unlikely for an LLM caller to hit).

**Verdict: ready to merge once the `build` CI check finishes green** (it was pending at review time — checked directly, not assumed from the clean review).

PR #5323, opened 2026-09-17
Fix #76711: refresh_data silently serves stale rows for tabular queries

Nelson-cao:
## Problem

`refresh_data` on an already-built tabular (e.g. SERVER_FILE) table assembly is unreliable and silent: it returns `{"ok":true}` whether or not it actually re-executed the query. `WorksheetAgentController.refreshData` cleared `AssetQuerySandbox`'s table-lens cache and `AssetDataCache`, but never invalidated `XSessionManager`'s deeper query-result cache, which `BoundQuery`/`TabularBoundQuery` execution reads from with effectively no time-based expiry. Once that cache was populated once for a given assembly, every later `refresh_data` call kept serving the original rows no matter how many times the underlying file changed, and a brand-new `add_table` against the same file always worked because it computes a different cache key (the assembly name is baked into the key via `TabularBoundQuery`'s `XQuery` name).

## Fix

Set `__refresh_report__` on the variable table around the reload in `refreshData`'s single-table branch, matching `WSQueryService.runQuery` (the UI's own "Run Query" action this method already mirrors elsewhere). This flag makes `XSessionManager.DataCacheResult.visitCache` skip the stale cache entry entirely, forcing a real re-execution and overwriting it with fresh data.

## Test plan

- Added `refreshDataSetsRefreshReportFlagAroundReload` to `WorksheetAgentControllerTest`, asserting the flag is set during the reload and removed afterward.
- `WorksheetAgentControllerTest`: 186/186 pass.
- Full `core`-module test suite (`mvn -pl core -am test`): 633/633 pass, 0 failures/errors.


Nelson-cao:
## Review update — round-1 fix does not reliably fix the bug (confirmed live)

This run's own `reviewer-76711` seat reviewed the first commit clean (correctness/root-cause fidelity checked out), but that missed a real race condition that GitHub's own `claude-review` bot found on the very first push, and which I've now **independently confirmed live** against the actual running service:

`box.getVariableTable().put("__refresh_report__", "true")` / `finally { remove(...) }` wraps `RenderWaitSupport.awaitOrRetry(...)`, which runs the reload on a **separate virtual thread** and does **not cancel** it on timeout (2s) — the work keeps running in the background per `RenderWaitSupport`'s own doc comment. On any reload slower than 2s, the `finally` clears the flag while the background thread is still executing, so `XSessionManager`'s read of `__refresh_report__` can see it already cleared and silently fall back to the stale cache — reproducing this exact bug.

Live repro (via `mcp__composer-chat__*`, against the actual rebuilt+restarted service running this fix): edited the probe file, called `refresh_data` twice in a row on the same table, both returned `{"ok":true}` but served stale data both times; a fresh `add_table` against the same file read current data immediately.

Sending back for a round-2 fix (in progress) — not merging this PR as-is. Full record: `docs/teams/2026-09-17-bug-76711-refresh-data-tabular-silent/bug-76711/04-review-r1.md` in stylebi-wiz.

Nelson-cao:
## Round 2: fixed the async-lifetime race from round 1

Closes the race a previous commit missed and `claude-review`/live testing caught (see the comment above): `refreshData` now synchronously calls `BoundQuery.clearQueryCache(vars)` (mirroring `WSQueryService.runQuery`'s own `AssetQuery.createAssetQuery(...)` + `clearQueryCache(...)` pattern) **before** the bounded async reload is even submitted, instead of relying on a `__refresh_report__` flag whose removal could race a still-running background thread. Verified this path has no Spring-context dependency (contrary to an old in-code comment's claim) by reading the actual call chain end to end.

Also extends the "refresh all tables" branch (`table` omitted) with the same cache invalidation — review Finding 2, previously unfixed.

New regression test (`refreshDataClearsBoundQueryCacheBeforeSlowReload`) simulates a reload slower than the timeout and asserts the cache is cleared before the reload even starts, via call-order tracking — this is the actual guard for the race; round 1's test never exercised real async timing so it couldn't catch this.

`core` module: 10942 tests, 0 failures, 0 errors (full suite, not just the targeted file).

Next: asking the user to rebuild + restart the live service, then re-running the same live `composer-chat` repro that caught round 1's regression, before considering this ready to merge.

Nelson-cao:
## Round 2 also fixed nothing — traced live with a debugger, found the actual reason

Live-tested round 2 (commit 21cc7a40c) against the rebuilt service — still reproduces identically. Attached IDEA's debugger to the running process and single-stepped through the actual execution.

**`BoundQuery.clearQueryCache()` — the mechanism round 2's fix relies on, and the same one `WSQueryService.runQuery` uses — is a silent no-op in this deployment.** `XSessionManager.removeQueryCacheData` guards its real work with `if(service instanceof XEngine)`; in this deployment `service` is a JDK dynamic proxy (`$Proxy224`), and `XEngine` is a concrete class, not an interface — a dynamic proxy can never pass that check. Confirmed live: execution jumps straight over the `if` body, `removeQueryCache` is never invoked.

Round 2 kept `__refresh_report__` as a "belt and suspenders" second layer, reasoning it no longer needed to win its timing race since the cache would already be cleared synchronously first — but since that synchronous clear is a no-op here, the flag is still the *only* actual mechanism, and round 2 never changed its lifetime bug (still cleared in the caller's `finally` as soon as `awaitOrRetry` returns, racing the still-running background reload exactly as in round 1).

Full trace: `docs/teams/2026-09-17-bug-76711-refresh-data-tabular-silent/bug-76711/05-idea-debug-r2.md` in stylebi-wiz.

Round 3: fix the flag's actual lifetime (move its removal into the reload callable itself, so it's cleared exactly when the real work finishes, not when the caller gives up waiting) rather than relying on the broken `clearQueryCache` path — in progress.

Nelson-cao:
## Round 3 also reproduces — found the REAL root cause this time (mode mismatch, not a caching mechanism bug)

Round 3 (commit 15165c91b) correctly fixed the `__refresh_report__` lifetime race — verified live, step by step: `refreshColumnSelection`/`loadTableData` both complete without throwing, and the flag is removed at exactly the right point. **And the bug still reproduces**, on a brand-new clean table, with zero debugger interference on the read side.

Root cause: `refreshData`'s single-table branch computes `int mode = WorksheetEventUtil.getMode(table)`, which for an ordinary table built via `add_table` (confirmed live: `isRuntime()=false`, `isLiveData()=false`) resolves to `DESIGN_MODE`. But `WorksheetPreviewService` — the actual implementation behind `preview_worksheet_data` — reads with a **hardcoded** `AssetQuerySandbox.RUNTIME_MODE` (`WorksheetPreviewService.java:95`), regardless of the table's own mode flags.

**`refresh_data` has been correctly and thoroughly invalidating a `DESIGN_MODE` cache entry that `preview_worksheet_data` never reads. The `RUNTIME_MODE` entry it actually reads has never been touched by any of the three fix rounds.** This is why the bug reproduced identically no matter how the invalidation mechanism itself was implemented — the mechanism was never the problem past round 1's initial diagnosis; which mode it targeted was.

Full trace: `docs/teams/2026-09-17-bug-76711-refresh-data-tabular-silent/bug-76711/06-idea-debug-r3.md` in stylebi-wiz.

Round 4: make `refreshData` also invalidate/reload the `RUNTIME_MODE` slot specifically — in progress.

PR #5324, opened 2026-09-17
fix(76742-vhl-005): validate webLink's hyperlink:<field>/=<expr> prefixes at write time

Zxm-happy:
## Summary

Redmine #76742, sub-defect VHL-005: `AssemblyHyperlinkService.apply()` accepted a `webLink` value
shaped as `hyperlink:<field>` or `=<expr>` with no validation, silently persisting a nonexistent
field name or a syntactically broken expression (`ok:true` either way). Both are live,
render/click-time mechanisms, not cosmetic — `Hyperlink.java` parses the `hyperlink:` prefix at
click time, and `ViewsheetSandbox.executeDynamicValue` runs a `=<expr>` value's script at view
refresh, catching a failure only as a `hyperlinkScriptFailed` viewer message one render downstream
— a check the plugin tool itself never sees.

- **`hyperlink:<field>`**: now validated against the assembly's actual runtime data columns — the
  same unfiltered set `VSDataSet.getHyperlink0` (chart) and `Hyperlink.getColumnValue` (table/
  crosstab, including its base-table fallback) read at click time — reached via the
  `RuntimeViewsheet`/`ViewsheetSandbox` already in hand inside `set()`'s `sessions.mutate` callback.
  Deliberately **not** `HyperlinkDialogService.getFields()`/`getAvailableRefs` (what
  `parseParamList` already validates a `paramList` `field` entry against): that list is scoped for
  a different, narrower purpose (restricting a *parameter* to dimensions already fixed when the
  link fires) and would cause false rejections of field names that resolve correctly at click time
  — e.g. a same-axis dimension positioned after the target, or a column reachable only through the
  base-table fallback.
- **`=<expr>`**: now run synchronously through the same script-execution path
  `ViewsheetSandbox.executeDynamicValue` uses at view refresh (`ViewsheetScope.execute`), so a
  broken expression is refused at write time with a field-named error instead of silently
  persisting and only failing later.

Companion wiz-side PR (Piece 1, documentation): inetsoft-technology/stylebi-wiz#2625

## Test plan

- [x] `./mvnw -pl core test -Dtest=AssemblyHyperlinkServiceTest` — 66 passed, 0 failed, 0 errors.


Zxm-happy:
## Review (VHL-005, Piece 2a/2b)

Read the actual diff (`gh pr diff`), independently re-traced every load-bearing claim in the fetched PR-head source (isolated worktree, not the shared checkout) rather than trusting the fix log's prose. Findings:

**Correctness — all confirmed:**
1. Field-mode validation (`requireValidWebLinkField`/`runtimeColumnNames`/`collectTableColumns`) genuinely reads the runtime dataset/table-lens column set, not `HyperlinkDialogModel.getFields()`. Confirmed no reference to `model.getFields()` anywhere in the new code. The regression test `acceptsAFieldTheNarrowParamListFieldSetDoesNotCarryButTheRuntimeDatasetDoes` is a real discriminating test — it sets `model.getFields()` to only `REVENUE` while the mocked dataset carries `STATE, REVENUE`, and asserts `hyperlink:STATE` is accepted; this would fail against the old, refuted `model.getFields()`-based approach.
2. Chart path: traced `ViewsheetSandbox.getData(name)` → `DataMap.NORMAL` (`= BRUSH | ZOOM`) → `VSAQuery.createVSAQuery` setting `qry.setBrush(true)` → `ChartVSAQuery.getData()`'s early `!brush` null-return branch is skipped → falls through to `new VSDataSet(data, refs)`. Confirmed exactly as claimed by reading the actual method bodies.
3. Table/crosstab descent (`collectTableColumns`/`TableFilter.getTable()`): terminates the same way `Hyperlink.getColumnValue`'s own base-table fallback loop does; no infinite-loop risk; deliberately a superset of the real fallback (biased toward not rejecting valid fields), consistent with stated intent.
4. Expression validation (`requireValidWebLinkExpression`) only reacts to a thrown exception from `scope.execute(...)`, never inspects the return value — a valid-but-falsy expression is not wrongly rejected.
5. Ran `AssemblyHyperlinkServiceTest` myself in a fresh worktree (not piped through `tail`): `Tests run: 66, Failures: 0, Errors: 0`. Matches the fix log's claim.
6. Diff scope confirmed clean — only `AssemblyHyperlinkService.java` and its test touched; no changes to `Hyperlink.java`, `ViewsheetSandbox.java`, `HyperlinkDialogService.java`, or `VSDataSet.java`.

**Important (non-blocking) finding:** `runtimeColumnNames`/`requireValidWebLinkExpression` don't catch an exception thrown *during* the data fetch (`ViewsheetSandbox.getData`/`getVSTableLens` both declare `throws Exception`) — only a clean `null`/empty result is treated as "skip validation." `ViewsheetSessionService.mutate()` only specially handles `CommandErrorException`, so a genuinely broken chart query (unrelated to the hyperlink value itself) would now abort the whole `set_hyperlink` write with an opaque error, where before this change no data was ever computed for `webLink` and the write always succeeded. Recommend treating a thrown exception from these two calls the same as "no data yet" (skip, don't refuse) — consistent with this PR's own stated philosophy. Not a blocker for this fix; suggest a fast follow-up.

**Nit (non-blocking):** `requireValidWebLinkExpression` executes against the *live* `VSAScriptable`, so a side-effecting expression runs once extra at write time in addition to its normal render-time execution — same call, same scriptable, just one more invocation. Not a new capability, just worth a one-line doc mention eventually.

**Blocker — this PR's own CI `build` check is currently red** (`gh pr checks 5324`), not just pending. Per this workflow's standing rule, a clean review (and a passing local test run) is not a substitute for that. I dug into the failure: `inetsoft-core` (the module this diff lives in) built and tested SUCCESS, including the 66/66 passing `AssemblyHyperlinkServiceTest`. The failure is downstream, in the `inetsoft-server` module's `process-aot` goal — a `ConflictingBeanDefinitionException` between `inetsoft.web.admin.viewsheet.ViewsheetService` and `inetsoft.web.admin.sheet.vs.ViewsheetService`. This looks unrelated to this PR's diff (core-module-only, no Spring bean/config touch, and the fetched worktree only contains one `ViewsheetService` class) — likely a base-branch (`stylebi-wiz-main-rebase`) or build-environment issue rather than something introduced here. Regardless of cause, **this must not be merged until the `build` check is actually green.**

**Verdict: not ready to merge.** Core fix logic is sound; blocked purely on the red CI `build` check.

PR #5326, opened 2026-09-17
fix(themes): add jarUploaded provenance field to CustomTheme (THM-005)

lizhepu:
## Summary

Redmine #76740, THM-005 (plugin/admin Themes area): `download_theme_jar` fabricates a valid jar for themes that never had a real jar uploaded, because `CustomTheme` has no field distinguishing "a jar file was actually uploaded" from "a jar-shaped CSS-only artifact was synthesized." The existing server-side guard in `AdminThemeController.downloadThemeJar` (`theme.getJarPath() == null`) is dead code — `jarPath` gets set unconditionally on every `addTheme`/`updateTheme` call, jar-uploaded or not.

This PR adds the new persisted `jarUploaded` bit and its read-side model counterpart. It does **not** wire the field into `ThemeService`/`AdminThemeController` — that logic lives in `wizard/enterprise`, outside this repo.

## Changes

- `core/src/main/java/inetsoft/sree/portal/CustomTheme.java`
  - `private boolean jarUploaded;` field (default `false`), mirroring the existing `emDark` boolean field's pattern exactly.
  - `isJarUploaded()` / `setJarUploaded(boolean)` accessors.
  - Wired into XML `writeXML`/`parseXML`, `equals`/`hashCode`/`toString`, and `clone()` (which copies fields one-by-one, not generically, so this needed its own explicit line).
- `core/src/main/java/inetsoft/web/admin/presentation/model/CustomThemeModel.java`
  - `@Nullable Boolean hasJar();` — a new read-side accessor on this `@Value.Immutable` interface, distinct from the existing write-only `jar()` upload-payload field (`{name, content}`, input-only). The generated `Builder` picks up a `hasJar(Boolean)` setter automatically.

## Two-repo fix

**This is one half of a two-repo fix.** The other half — populating `hasJar`/`jarUploaded` in `ThemeService.getTheme()`/`addTheme()`/`updateTheme()` and switching `AdminThemeController.downloadThemeJar`'s guard from `theme.getJarPath() == null` to `!theme.isJarUploaded()` — lives in `wizard/enterprise` and will be a separate PR that depends on this one merging first (and `wizard`'s submodule pointer being bumped to include it).

**Do not merge — the user will review and merge this personally.**

## Test plan

- [x] Confirmed diff is scoped to exactly these two files (`git status`).
- [x] Checked existing tests referencing `CustomTheme`/`CustomThemeModel` (`CustomThemesManagerTest.java` and others) — none assert an exhaustive field list, exact XML string, or equals/hashCode contract that a new default-`false` field would break.
- [ ] Module-scoped Maven compile of `core` was attempted but is not usable in this worktree: `build-tools/antlr2-maven-plugin`'s target dir has a stale/broken plugin descriptor (pre-existing environment state, unrelated to this change), which fails both offline and online `mvn -pl core -am compile`. Not fixed here as it's outside this fix's scope.
- No new automated test per this fix's stated convention for this change; the diagnosis recommends regression tests (`AdminThemeControllerTest`, `ThemeServiceTest`, XML round-trip) that belong with the enterprise-side PR since the guard logic and test infra for it live there.


lizhepu:
## Review round 1

Checked `CustomTheme.java` against the diagnosis's checklist (all read directly from the PR head commit, not taken on the fix docs' word):

- **XML serialize/deserialize**: `writeXML` writes `jarUploaded` unconditionally (mirrors `emDark`); `parseXML` reads via `"true".equals(...)`, defaulting to `false` when the element is absent — correct for old-format XML ("no jar" is the right default). Confirmed exact.
- **equals/hashCode/toString**: `jarUploaded` included alongside `jarPath`/`emDark` in all three, consistent with existing convention.
- **clone()**: this method copies fields one-by-one (not generically) — `clone.jarUploaded = jarUploaded;` was added next to `clone.jarPath = jarPath;`. Read the full `clone()` body; no other field-by-field construction site was missed.
- **CustomThemeModel.java**: `@Nullable Boolean hasJar()` added next to the existing write-only `jar()` field, with no change to the `from(...)` builder overloads — correctly left for the enterprise-side PR.

Tried to build a counter-example (missing XML element on old data, double-clone, a theme with `jarPath` set but `jarUploaded` never set) and couldn't break any of it — the change is additive, defaults safely, and isn't read anywhere yet in this repo.

**No findings. Approving.**

PR #5328, opened 2026-09-17
Add DATABASE authentication provider support to admin-chat (bug 76710/76716)

NeverStopFMH:
## Summary
- `resolveCreate`/`resolveUpdate`/`requireDeletableAuthenticationType` previously excluded `providerType=DATABASE` unconditionally, citing a license gate this service did not itself enforce — that rationale went stale once bug #76359 (already merged) added the real gate, `AuthenticationProviderService.checkProviderTypeLicensed`, to `addAuthenticationProvider`/`editAuthenticationProvider` directly. This reuses that gate (via a new public `requireProviderTypeLicensed` wrapper) rather than re-deriving a parallel check.
- Adds `ProviderDatabaseSpec` (mirroring `DatabaseAuthenticationProviderModel` field-for-field) and a new `databaseSpec` field on `ProviderChangeRequest`. `resolveUpdate` dispatches on the CURRENT provider's real type, never on which of `spec`/`databaseSpec` the caller sent.
- `requireDeletableAuthenticationType` now also allows DATABASE, since delete's own rollback-recreate risk is already covered by the same reused license gate.
- Adds `GET /api/wiz/v1/admin/providers/multi-tenant` (`AdminProviderController#getMultiTenantStatus`), reading `SUtil.isMultiTenant()` directly — needed because the generic properties-catalog path is genuinely ambiguous for this specific, uncatalogued, never-explicitly-set property.

Companion `stylebi-wiz` PR (the admin-chat plugin side that consumes this): inetsoft-technology/stylebi-wiz#2629

## Test plan
- [x] `community/core` full recompile: `BUILD SUCCESS`
- [x] `ProviderChangePlanServiceTest` (58), `ProviderChangesetApplyServiceTest` (32), `AdminProviderControllerTest` (15), `AuthenticationProviderServiceTest` (10) — 115/115 passing
- [x] Live-verified end to end against a real StyleBI instance in both multi-tenant and non-multi-tenant modes: create/update/delete a DATABASE provider, real JDBC connection, real directory queries (users/groups/roles)


NeverStopFMH:
Follow-up: the review on the companion `stylebi-wiz` PR (#2629) found that `requireDatabaseCredentialMode`/`requireDatabaseCredentialCrossValidation`'s server-side counterparts here had the identical gap — unconditionally requiring `secretId` or `user`+`password` regardless of `requiresLogin`. Confirmed live against the real EM dialog (`database-provider-view.component.html`'s `@if` wraps the ENTIRE credential block on `requiresLogin`, `.component.ts` clears all three validators when unchecked) and applied the same fix here in `0d1581193`, with 3 new tests. `requiresLogin=false` now exempts the credential-mode check on both create and update; an omitted value still defaults to `true` and requires credentials exactly as before.

fredaan (approved):

PR #5332, opened 2026-09-17
Fix #76711: refresh_data silently serves stale rows for tabular queries

Nelson-cao:
## Summary

`refresh_data` on a tabular (SERVER_FILE) worksheet table could report `{"ok":true}` while
still serving pre-refresh rows, indefinitely, with no error. Root cause spanned five distinct
layers, each confirmed via live IDEA-attached debugging against a running StyleBI instance
(not just unit tests):

1. **Async flag lifetime**: `__refresh_report__` (the only mechanism that bypasses
   `XSessionManager`'s query-result cache) was cleared by the caller's own timeout `finally`
   while the reload kept running in the background on a separate virtual thread — clearing the
   flag before the work it guards had actually finished.
2. **Dead cache-clear call**: `XSessionManager.removeQueryCacheData`'s `instanceof XEngine`
   guard never passes in this deployment, since `service` is a JDK dynamic proxy — filed
   separately as #76760, not fixed here since correctness no longer depends on it.
3. **Mode-keyed cache-slot mismatch**: every real reader of this data
   (`preview_worksheet_data`, CSV export, sample probes) reads under a hardcoded
   `RUNTIME_MODE`, but an ordinary `add_table`'d table's own mode is `DESIGN_MODE` — the
   original invalidation never touched the `RUNTIME_MODE` slot in
   `AssetQuerySandbox.tmap`/`AssetDataCache`.
4. **Query-shape mismatch**: `XSessionManager.dataCache`'s key includes the query's own
   `maxrows`/`timeout`. The row-cap hint was only removed when `mode == RUNTIME_MODE`
   (mirroring `WSQueryService.runQuery`), so a `DESIGN_MODE` table's reload kept writing into
   the wrong (row-capped) cache entry no matter how thoroughly the outer layers were
   invalidated.
5. **No forced RUNTIME_MODE execution**: invalidating the outer `tmap`/`AssetDataCache` slots
   only clears the shell — nothing in `refresh_data` ever forced a real query execution under
   `RUNTIME_MODE`, so the deep `XSessionManager.dataCache` entry `RUNTIME_MODE` readers actually
   consult was never repopulated. Fixed by forcing a real `RUNTIME_MODE` read inside the
   `__refresh_report__`-protected window, in both the single-table branch and the "refresh all
   tables" branch (which had the identical gap, confirmed live).

## Test plan

- [x] `core` module unit tests: 192/192 pass in `WorksheetAgentControllerTest`, including 3 new
      regression tests for the row-cap-hint removal and forced `RUNTIME_MODE` read (single-table
      and bulk) — each confirmed to fail against the pre-fix code via revert-and-retest.
- [x] Live verification against a running StyleBI instance via `mcp__composer-chat__*` only (no
      backend/DB shortcut): repeated edit-file → `refresh_data` → `preview_worksheet_data`
      cycles, single-table and bulk, each correctly reflecting the latest file content across a
      genuine service restart.


core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
```diff
@@ -3318,18 +3366,120 @@ private void refreshData(String sessionToken, EditRequest req, Principal user)
...
+
+                  try {
+                     clearBoundQueryCache(box, table, mode);
+                  }
+                  catch(Exception e) {
+                     LOG.warn("Failed to clear the bound query cache for table: " +
+                        table.getName(), e);
+                  }
+
+                  box.getVariableTable().put("__refresh_report__", "true");
```
Nelson-cao, line 3489: **Critical: the bulk ("refresh all tables") branch never forces a RUNTIME_MODE execution for a table whose own mode is already RUNTIME_MODE**

This is the exact gap round 6's PR description says it closes for "both the single-table branch and the refresh all tables branch," but the bulk branch's implementation doesn't actually cover it.

Walk the per-table loop:
```java
box.getVariableTable().put("__refresh_report__", "true");
try {
   if(mode != AssetQuerySandbox.RUNTIME_MODE) {
      box.getTableLens(table.getName(), AssetQuerySandbox.RUNTIME_MODE);
   }
}
finally {
   box.getVariableTable().remove("__refresh_report__");
}
```

Unlike the single-table branch — which unconditionally calls `WorksheetEventUtil.loadTableData(rws, req.table(), true, true)` under the table's *own* mode (so a table whose `getMode()` already returns `RUNTIME_MODE` gets a genuine forced RUNTIME_MODE execution from that call alone, and the extra `getTableLens(RUNTIME_MODE)` is correctly skipped as redundant) — the bulk branch has **no equivalent unconditional reload at all**. The *only* thing that can force a real query execution in this loop is the guarded `getTableLens` call, and that's precisely what's skipped when `mode == RUNTIME_MODE`.

`resetTableLens`/`assetDataCache.remove` (a few lines above) only invalidate the outer `AssetQuerySandbox.tmap`/`AssetDataCache` slots — per this same PR's own round-4/round-6 comments, that's lazy invalidation only and does not itself run a query. So for a table already in RUNTIME_MODE, this loop: invalidates the outer caches, runs the (mostly no-op in this deployment, per the round-2/round-3 comments) `clearBoundQueryCache`, and then sets and immediately clears `__refresh_report__` around nothing. No query ever actually executes inside the `__refresh_report__`-protected window for this table. The next real RUNTIME_MODE reader (e.g. `preview_worksheet_data`) will hit a genuine `tmap` cache miss and re-execute — but by then `__refresh_report__` is long gone, so that re-execution's own `XSessionManager.dataCache` lookup can still return the stale entry. That's bug #76711, reproduced for this specific table/mode combination.

None of the new tests (`refreshDataAllTablesAlsoForcesARuntimeModeExecution` included) construct a table whose own mode is RUNTIME_MODE, so this gap isn't caught.

Suggested fix: mirror the single-table branch's structure — call `box.getTableLens(table.getName(), mode)` (the table's own mode) unconditionally inside the `__refresh_report__` window, then additionally force `RUNTIME_MODE` only when `mode != RUNTIME_MODE`, same as the single-table branch already does.

core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java
```diff
@@ -3277,34 +3288,71 @@ private void refreshData(String sessionToken, EditRequest req, Principal user)
...
+               // XSessionManager cache entry via a fresh AssetQuery.createAssetQuery(...) +
+               // BoundQuery.clearQueryCache(...); do it here too, best-effort. NOTE: confirmed
+               // live (see bug #76711 archive, 05-idea-debug-r2.md) that this call is a silent
+               // no-op in deployments where XSessionManager's underlying service is a JDK dynamic
+               // proxy rather than a concrete XEngine -- removeQueryCacheData's `instanceof
+               // XEngine` guard can never pass a proxy, so removeQueryCache is never actually
+               // invoked. Do NOT rely on this call alone to invalidate the cache; the
+               // __refresh_report__ flag below is the mechanism this method's correctness
+               // actually depends on.
+               clearBoundQueryCache(box, table, mode);
```
Nelson-cao: **Important: `clearBoundQueryCache` is called unguarded here, contradicting its own "best-effort" comment**

The comment right above this call says:

> WSQueryService.runQuery also drops a BoundQuery/TabularBoundQuery's own XSessionManager cache entry via a fresh AssetQuery.createAssetQuery(...) + BoundQuery.clearQueryCache(...); do it here too, **best-effort**. [...] Do NOT rely on this call alone to invalidate the cache; the `__refresh_report__` flag below is the mechanism this method's correctness actually depends on.

But the call itself isn't wrapped in try/catch here, so if `AssetQuery.createAssetQuery(...)` throws for any reason (bad datasource state, a transient resolution failure, etc.) the exception propagates straight out of `refreshData` and fails the entire single-table refresh — even though, by the comment's own admission, this call is optional and the method's correctness doesn't depend on it succeeding.

Compare with the bulk branch a bit further down, which correctly treats the identical call as best-effort:
```java
try {
   clearBoundQueryCache(box, table, mode);
}
catch(Exception e) {
   LOG.warn("Failed to clear the bound query cache for table: " + table.getName(), e);
}
```

Suggest wrapping the single-table branch's call the same way, so a failure here (which is explicitly documented as non-essential, and confirmed live to already be a silent no-op in this deployment's dynamic-proxy `XSessionManager` case) can't turn into a hard failure of the whole `refresh_data` request.

Nelson-cao:
## Review

Reviewed `WorksheetAgentController.refreshData`/`clearBoundQueryCache` and the new tests in `WorksheetAgentControllerTest`.

**Scoping/compile check**: the `mode` variable is correctly scoped — the pre-computed `final int mode` (line 3268) is declared *inside* the `if(req.table() != null && !req.table().isBlank())` block, so it does not conflict with the separately-declared `int mode` inside the `else` (bulk) branch's for-loop; these are sibling blocks, not nested, so no shadowing/compile error. The `-1` sentinel and its `mode >= 0` guard at the forced-read call site are consistent.

**`RenderWaitSupport.awaitOrRetry` exception handling**: verified against `RenderWaitSupport`'s actual implementation — the only way `awaitOrRetry` can throw while the reload is *still running in the background* is the `TimeoutException -> RenderNotReadyException` path. Every other exception path (the callable's own exception via `ExecutionException`, or a `submit()`-time rejection) means the callable has either already finished (its own `finally` already removed `__refresh_report__`) or never started at all. So the `catch(RenderNotReadyException)` vs `catch(Exception)` split correctly preserves the invariant the comments claim.

**Two issues found, posted inline:**
1. **Critical** — the bulk ("refresh all tables") branch never forces a real RUNTIME_MODE query execution for a table whose *own* mode is already RUNTIME_MODE (the `getTableLens(RUNTIME_MODE)` call is skipped exactly in that case, and there's no unconditional reload equivalent to the single-table branch's `loadTableData` call to fall back on). This reproduces bug #76711 for that specific table/mode combination, contradicting the PR's stated goal of fixing "both the single-table branch and the refresh all tables branch." Not covered by any of the new tests.
2. **Important** — `clearBoundQueryCache(box, table, mode)` in the single-table branch is called unguarded, but its own comment describes it as "best-effort" / non-essential to correctness. The bulk branch correctly wraps the identical call in try/catch + `LOG.warn`; the single-table branch should do the same.

Everything else — the row-cap-hint removal becoming unconditional, the RUNTIME_MODE cache-slot invalidation, the `__refresh_report__` flag lifetime fix, and the 3 new regression tests for the covered cases — looks correct and non-trivial (each test's own docstring explains which prior round's code it would fail against, and that's borne out by re-reading the diff).

Not mergeable as-is until finding 1 (bulk-branch RUNTIME_MODE gap) is addressed — it's a real recurrence of the bug this PR exists to fix, just for a narrower table-mode combination. Finding 2 is a robustness/regression-risk issue worth fixing in the same pass.

Nelson-cao:
Both findings from the review above are fixed and pushed (commit 669d50351):

1. **Critical (bulk branch RUNTIME_MODE gap)**: the forced `getTableLens(RUNTIME_MODE)` read in the bulk ("refresh all tables") branch is now called unconditionally, not only when `mode != RUNTIME_MODE`. That guard was correct in the single-table branch (where `loadTableData` already executes the query under the table's own mode first) but wrong here, since the bulk branch has no equivalent unconditional execution to fall back on for a table whose own mode is already RUNTIME_MODE.

2. **Important (unguarded clearBoundQueryCache)**: the single-table branch's call is now wrapped in try/catch + `LOG.warn`, matching the bulk branch's existing handling and its own "best-effort" comment.

Added a new regression test (`refreshDataAllTablesForcesRuntimeModeExecutionEvenWhenAlreadyRuntimeMode`) that reproduces finding 1's exact scenario — confirmed to fail against the pre-fix code via revert-and-retest. Full suite: 193/193 pass.

Nelson-cao:
Re-verified commit 669d50351 directly against the source (not just the description):

- Finding 1 (bulk-branch RUNTIME_MODE gap): confirmed. The forced `box.getTableLens(table.getName(), AssetQuerySandbox.RUNTIME_MODE)` call in the bulk branch (WorksheetAgentController.java) is now unconditional, no more `if(mode != RUNTIME_MODE)` guard around it. New test `refreshDataAllTablesForcesRuntimeModeExecutionEvenWhenAlreadyRuntimeMode` sets `table.setRuntime(true)` (verified this makes `WorksheetEventUtil.getMode` return RUNTIME_MODE), drives the bulk (no-table-name) refresh_data path, and asserts `verify(box).getTableLens("Table1", RUNTIME_MODE)` — this genuinely would have failed against the prior guarded version.
- Finding 2 (unguarded clearBoundQueryCache): confirmed. The single-table branch's call is now wrapped in try/catch + LOG.warn, matching the bulk branch's existing handling.

Both findings resolved. No new issues spotted in the diff. Mergeable as far as my review goes.

PR #5333, opened 2026-09-17
fix(76764-vts001): register missing tableStyle alias for calctable

Zxm-happy:
## Summary

`PropertyAliases.calcTable()` never registered a `tableStyle` short-name alias, unlike `table()` and `crosstab()`, which both map it to `tableViewGeneralPaneModel.tableStylePaneModel.tableStyle`. `CalcTablePropertyDialogModel`/`CalcTablePropertyDialogService` already read and write that exact path unconditionally for calctable assemblies (`getCalcTablePropertyDialogModel`/`setCalcTablePropertyModel`), so the short name was simply omitted, not intentionally excluded — no `DEAD_FIELDS` entry, no gating comment, unlike deliberate omissions elsewhere in this file.

Effect: `set_assembly_properties({assembly: "<calctable>", properties: {tableStyle: "..."}})` was refused with `IllegalArgumentException: "tableStyle" is not a property of calctable. Known names: ...` even though the identical raw dotted path succeeded and was genuinely applied.

## Fix

One line in `calcTable()`, matching the existing `table()`/`crosstab()` pattern and position:

```java
aliases.put("tableStyle", "tableViewGeneralPaneModel.tableStylePaneModel.tableStyle");
```

No plugin-side (TypeScript) change is needed — `plugin/composer/src/tools/propertyTools.ts`'s `set_assembly_properties` does no client-side name validation and passes properties straight through to the backend.

## Test plan

- [x] Added a regression assertion in `PropertyAliasesTest.newTableAndCrosstabAliasesRoundTrip()` asserting `PropertyAliases.resolveForWrite("calctable", "tableStyle")` resolves to the same path table/crosstab use.
- [x] `./mvnw test -pl core -Dtest=PropertyAliasesTest` — 46 tests run, 0 failures, 0 errors.


jshobe-inetsoft (approved):
## Code Review

**Summary:** Registers the missing `tableStyle` short-name alias for `calctable` in `PropertyAliases.calcTable()`, matching the existing pattern in `table()` and `crosstab()`.

### Correctness
- Verified `table()` (line 797) and `crosstab()` (line 827) both map `tableStyle` → `tableViewGeneralPaneModel.tableStylePaneModel.tableStyle`. The new line in `calcTable()` (line 1029) uses the identical path and is inserted in the same relative position (immediately after the `title()` call), consistent with the existing style.
- Verified `"calctable"` is the correct registered type key (`register(registry, "calctable", CalcTablePropertyDialogModel.class, calcTable())`), matching what the new test uses.
- Verified in `CalcTablePropertyDialogService`: `getCalcTablePropertyDialogModel` populates `tableStylePaneModel.setTableStyle(...)` (read) and `setCalcTablePropertyModel` applies `calcTableAssemblyInfo.setTableStyleValue(tableStylePaneModel.getTableStyle())` (write) unconditionally — no gating condition — so the alias is safe to expose unconditionally, exactly like `table`/`crosstab`.
- Confirmed `calctable` is not present in `DEAD_FIELDS`, so there's no prior deliberate exclusion being overridden.

### Regression / Backward Compatibility risk
- None found. This is a pure addition to a `Map<String,String>` populated by a private builder method; it does not alter any existing alias, remove any entry, or change resolution order for other keys. Old behavior (refusal with `IllegalArgumentException`) is only replaced for this one previously-missing key; no caller could have depended on the refusal.

### Tests
- The regression assertion was added to the existing `newTableAndCrosstabAliasesRoundTrip()` test rather than a new dedicated test method. This is a minor style/naming nit (the method name no longer fully reflects its contents) but not a functional problem, and the PR description states `PropertyAliasesTest` passes (46 tests, 0 failures).
- The added assertion directly targets the fixed behavior (`resolveForWrite("calctable", "tableStyle")`) and is sufficient to catch a regression.

### Code Quality / Style
- Minimal, single-line change scoped exactly to the defect. No unrelated refactoring.
- No error handling, security, or readability concerns — this is static alias-table registration data, not runtime logic.

### CI / Mergeability
- All checks pass except `claude-review`, which failed with "Claude encountered an error after 0s" (a known automation/quota failure, not a real finding) — this manual review supersedes it.

### Verdict

PR #5334, opened 2026-09-17
fix(76751): let DateLevels.normalize accept a $(ComponentName) dynamic value

lizhepu:
## Summary
- Companion Java-backend fix for Redmine #76751 / stylebi-wiz PR #2638. That plugin PR lets `set_chart_shelf`/`set_table_fields`/`set_aesthetic_field` forward a chart field's `dateLevel` as a `$(ComponentName)` reference to a Form component's live value (StyleBI's existing dynamic-value convention, already applied to `column` and to a measure's `aggregate` formula).
- `DateLevels.normalize()` had no branch recognizing that shape and threw `IllegalArgumentException('$(ComponentName)' is not a date level...)` before the value ever reached `VSDimensionRef`'s already dynamic-value-aware `setDateLevelValue`. This adds a bypass: a well-formed `$(...)`/`=script` dynamic value returns unchanged; every other input is still validated exactly as before.
- Confirmed the callers (`FieldRefFactory.toChartRef`, `TableBindingMutator.dimensions()`/`matches()`) both write through `BDimensionRefModel.setDateLevel` (a bare field) into `VSDimensionRef.setDateLevelValue`, which is already dynamic-value-aware -- no caller changes were needed beyond `DateLevels.normalize()` itself.

## Implementation note
- The natural idiom (`VSUtil.isDynamicValue`) could not be called directly: `VSUtil`'s static initializer reaches Spring's `ApplicationContext` via `ConfigurationContext.getSpringBean`, which plain-JUnit callers of `DateLevels` (its existing test class) don't have. Importing it broke all 32 pre-existing `DateLevels` tests with `NoClassDefFoundError`. The fix instead reimplements `VSUtil.isDynamicValue`'s exact one-line predicate locally in `DateLevels`, keeping it a dependency-free utility as before.

## Test plan
- [x] `FieldRefFactoryTest` (34/34 pass, including 2 new regression tests: dynamic reference passes through unchanged, a malformed `$(...)`-shaped garbage string still throws)
- [x] `TableBindingMutatorTest` (82/82 pass, unaffected)

Companion to stylebi-wiz PR #2638 -- that PR's `dateLevel` support is inert without this fix; recommend landing this one first or close together.


jshobe-inetsoft (approved):
## Code Review

**Summary:** `DateLevels.normalize()` gains a bypass so a well-formed `$(ComponentName)`/`=script` dynamic value passes through unresolved instead of throwing `IllegalArgumentException`, letting a chart/table dimension's `dateLevel` carry a Form-component reference the same way `column` and aggregate `formula` already can. The new `DateLevels.isDynamicValue()` predicate is also used to reject a dynamic `dateLevel` explicitly in `CalcTableService.applyDateGroup` (where `OrderModel.option` is a plain `int` with no `DynamicValue` counterpart to hold it), and it replaces a duplicated private predicate that `TableBindingMutator` previously reimplemented for `percentageBy`/`stringBoolean`.

### Correctness
- Verified `VSDimensionRef.dlevelValue` is a genuine `DynamicValue` (`core/src/main/java/inetsoft/uql/viewsheet/VSDimensionRef.java:67`) and `setDateLevelValue` just sets the raw `DValue` (line 571-576) — resolution happens later via `getRuntimeValue(true)` in `getDateLevel()`. This confirms the PR's premise: the receiving setter is already dynamic-value-aware, so `DateLevels.normalize()` only needed to stop rejecting the shape before it got there.
- `DateLevels.isDynamicValue()`'s predicate (`(val.startsWith("$(") && val.endsWith(")")) || val.startsWith("=")`) is an exact mirror of `VSUtil.isVariableValue`/`isScriptValue`/`isDynamicValue` (`core/src/main/java/inetsoft/uql/viewsheet/internal/VSUtil.java:291-330`), so no semantic drift between the two copies.
- The calc-table path (`CalcTableService.applyDateGroup`) correctly special-cases the dynamic-value shape before `Integer.parseInt`, producing a clear, field-named `IllegalArgumentException` instead of leaking a confusing `NumberFormatException`. This is the right call since `OrderModel.option` genuinely has nowhere to stash an unresolved dynamic value.
- `TableBindingMutator.matches()` (used to preserve sort/ranking across a rebind) compares `previousLevel`/`incomingLevel` with plain `Objects.equals`, which continues to work correctly for a dynamic value stored verbatim on both sides — no special-casing needed there, and none was added.
- The refactor commit correctly removes the now-dead private `TableBindingMutator.isDynamicReference` and repoints both of its call sites (`percentageBy`, `stringBoolean`) at the new package-visible `DateLevels.isDynamicValue`; confirmed no other reference to the removed method remains anywhere in the tree.
- Confirmed `DateLevels` and its three callers (`FieldRefFactory`, `TableBindingMutator`, `CalcTableService`) are all in the same package (`inetsoft.web.wiz.binding`), so the package-private visibility on `isDynamicValue` is sufficient — no unnecessary API widening.

### Testing
- Ran the full affected test classes against the PR branch: `FieldRefFactoryTest` (34/34), `CalcTableServiceTest` (60/60, includes the 3 new cases), `TableBindingMutatorTest` (82/82) — all green, matching the PR's stated test plan.
- New tests cover the three meaningful cases: a well-formed dynamic reference passes through (`FieldRefFactoryTest.passesADynamicComponentReferenceThrough`), a malformed `$(...)`-shaped string is still rejected (`stillRefusesAGarbageValueShapedLikeAReference`), and the calc-table path rejects a dynamic value with a message naming both the field and the offending value (`CalcTableServiceTest.refusesADynamicDateLevelOnACalcTableCellWithACleanError`), plus a non-regression check that a plain garbage literal (`"fortnight"`) is still refused there too.
- No existing test appears to rely on `normalize()` throwing for `$(...)`/`=...` input, and no test relies on the now-removed private `TableBindingMutator.isDynamicReference` (it was private, so nothing outside the class could).

### Design / maintainability
- Reimplementing `VSUtil.isDynamicValue`'s predicate locally, rather than calling it, is well-justified and documented in both the code comment and the PR description: `VSUtil`'s static initializer reaches Spring's `ApplicationContext` via `ConfigurationContext.getSpringBean`, which the plain-JUnit `DateLevels`/`FieldRefFactory` tests don't have available. This avoids turning 32+ existing lightweight unit tests into Spring-context tests. The tradeoff (a second copy of a one-line predicate to keep in sync) is small and is now de-duplicated to a single package-shared location instead of two.
- Comments throughout are thorough and explain *why*, not just *what* — consistent with the surrounding file's existing documentation style.
- Error messages are specific and actionable (field name + offending value + guidance), consistent with the rest of `DateLevels`.

### Security / regressions / compatibility
- No security concerns: this only changes validation of an already-trusted internal string field; it doesn't affect script/expression evaluation itself (that happens later, in the existing dynamic-value resolution path, unchanged by this PR).
- No backward-incompatibility: previously-valid inputs (names, numbers, `-1` sentinel, `null`/blank) are handled identically; the only behavior change is that a previously-rejected input shape (`$(...)`/`=...`) is now accepted for chart/table dimensions and explicitly (and clearly) rejected for calc-table cells rather than falling through to a less helpful error.
- No regression risk identified in the three call sites reviewed; all existing tests for `DateLevels`, `FieldRefFactory`, `CalcTableService`, and `TableBindingMutator` pass unchanged.

### CI / mergeability
- `build`, `Semgrep OSS`, `Semgrep Scan`, and `license/cla` checks are all passing.
- `claude-review` (automated Claude Code Review) failed immediately ("Claude encountered an error after 0s"), consistent with a transient quota/usage-limit issue on the automation rather than a real finding on this PR. This manual review supersedes it.
- PR is `MERGEABLE` with no merge conflicts against its base branch (`stylebi-wiz-main-rebase`).

### Verdict
No blocking issues found. All non-excluded checks pass, and the PR has no merge conflicts. **Approving.**

PR #5335, opened 2026-09-17
fix(76764-vts003): refuse a dangling tableStyle write on table/crosstab/calctable

Zxm-happy:
## Summary

`set_assembly_properties` silently accepted a `tableStyle` value that matched no real table
style. `PropertyPath`'s `CONSTRAINED_STRINGS` gate does not cover `tableStyle` (its domain is
dynamic and per-library, unlike the closed enums that gate handles), so an unresolvable name/ID
was written through unchanged. At render time, `DataVSAQuery`/`VSUtil.getTableStyle` resolve that
to `null` and silently fall back to CSS-only formatting -- no error anywhere, and the bogus value
is echoed back on read, indistinguishable from a table that never had a style set.

`AssemblyPropertyService.set()` now validates any `*.tableStylePaneModel.tableStyle` write (covers
`table`/`crosstab`/`calctable` uniformly, since all three share `TableStylePaneModel`) against the
model's own, already-populated `styleTree` before applying it. The match accepts **either**
`TreeNodeModel.data()` (the style's internal ID -- what the interactive Composer UI itself
writes/matches) **or** `label()` (the folder-stripped display name -- what this plugin's own
`tableStyleTools.ts` documents as the calling convention), since both are genuinely resolvable at
render time via `LibManager.getTableStyle` (exact-ID lookup, then fuzzy name lookup). A validator
that accepted only one field would falsely refuse the other's real, currently-legitimate calling
convention.

No change to `PropertyPath.java`/`PropertyAliases.java` or any wiz-side (TypeScript) file --
the plugin already surfaces the Java error text verbatim to the caller, so a client-side preflight
duplicate is unnecessary.

Note: VTS-001 (a sibling bug from this same batch, adding a `calctable` `tableStyle` short alias
in `PropertyAliases.java`) may be landing concurrently. Different file, no expected conflict, but
flagging so both are visible to the reviewer.

## Test plan

- [x] `./mvnw test -pl core -Dtest=AssemblyPropertyServiceTest -DfailIfNoTests=true -q` --
      65 tests run (60 pre-existing + 5 new), 0 failures/errors/skipped, exit code 0.
- [x] New tests cover: a dangling name refused on `table` and `calctable`; a write matching a real
      style by `label()` (name) succeeds; a write matching a real style by `data()` (ID) succeeds;
      clearing `tableStyle` to `""` is still allowed.


Zxm-happy:
## Review round 1 -- no blocking issues found

Re-verified the diagnosis/refutation/fix claims independently against current source (not taken on the write-up's word), plus one direct experiment. Re-ran `AssemblyPropertyServiceTest` in a fresh, isolated worktree (detached at `refs/pull/5335/head`): `Tests run: 65, Failures: 0, Errors: 0, Skipped: 0`, real exit code `0`.

1. **`data()`-or-`label()` matching**: read `VSDialogService.getChildNodes` directly. Folder nodes never set `leaf(true)` (`TreeNodeModel.leaf()` defaults to `false`), so `styleTreeHasStyle`'s `node.leaf() && (...)` guard structurally excludes every folder from the match -- folder-spoofing a real style name is not possible. Both `data()` (ID, the interactive Composer UI's convention) and `label()` (folder-stripped name, this plugin's own `tableStyleTools.ts` convention) are genuinely legitimate today; matching either is correct.

2. **Empty/blank "clear the style" allowance**: wrote a temporary debug test writing a whitespace-only value (`"   "`) via the raw path and printed the stored result -- `STORED_VALUE=[]`. Not a bypass: `PropertyPath.coerce()` (`PropertyPath.java:233`) trims every String-typed write at the shared layer, independent of this PR's check, so whitespace-only behaves identically to an explicit empty write. No dirty untrimmed value can reach storage.

3. **`calctable` reachability independent of VTS-001**: the check keys off the *resolved* path suffix, and `PropertyAliases.resolveForWrite` passes an unaliased dotted key through unchanged (`PropertyAliases.java:345-364`), so the raw path already reaches `calctable` today with zero dependency on VTS-001's short-alias addition. `refusesADanglingTableStyleNameOnACalcTable` confirms this and passes.

4. **Test-harness wiring** (`serviceWithTable`/`serviceWithCalcTable`): structurally identical to every existing sibling (`gauge`, `selectionList`, `calendar`, `chart`) in the same harness method -- no deviation. The harness mocks the whole `*PropertyDialogService` (as every other test in this file does), so it doesn't exercise the real `VSDialogService.getChildNodes` construction path directly -- but I independently read that method and confirmed the canned `sampleStyleTree()` matches its real output shape exactly. Pre-existing test-file convention, not a new gap.

5. **`SheetPropertyService.set()`**: grepped directly -- no `tableStyle` reference anywhere in that file. Genuinely unreachable/out of scope, as claimed.

Note: `claude-review` shows as failed, but its run log shows an infra error (`is_error:true` after 1 turn / 359ms, no review content produced) -- not a real review verdict either way.

`build` is still pending as of this review -- holding off on merge until it goes green.

Full write-up: `docs/teams/2026-09-17-bugs-76764/bug-VTS-003/04-review-r1.md` in the wiz monorepo.

jshobe-inetsoft (approved):
## Code Review — bug/76764-vts-003: refuse a dangling `tableStyle` write

### Summary
This PR closes a real gap: `AssemblyPropertyService.set()` writes an arbitrary `tableStyle`
string through unvalidated because `PropertyPath`'s `CONSTRAINED_STRINGS` gate only covers
closed-enum string fields, not `tableStyle`'s dynamic, per-library domain. At render time
`VSUtil.getTableStyle`/`LibManager.getTableStyle` resolve an unknown value to `null` and
silently fall back to CSS-only formatting — the write reports success, and the bogus value is
echoed back on read, indistinguishable from "no style set". The fix adds a validator that
checks any `*.tableStylePaneModel.tableStyle` write against the model's own already-populated
`styleTree` before applying it.

### Verification performed
- Traced the suffix check (`entry.getValue().endsWith(".tableStylePaneModel.tableStyle")`)
  against `PropertyAliases`: `table()` and `crosstab()` both alias `tableStyle` to
  `tableViewGeneralPaneModel.tableStylePaneModel.tableStyle`; `calctable()` has no friendly
  alias for it but the raw-dotted-path escape hatch (`PropertyAliases.resolve`) reaches the same
  path, and the PR's own `refusesADanglingTableStyleNameOnACalcTable` test exercises exactly that
  route. All three assembly types are genuinely covered.
- Confirmed `TableViewPropertyDialogService`, `CrosstabPropertyDialogService`, and
  `CalcTablePropertyDialogService` all populate `styleTree` via the same
  `VSDialogService.getStyleTree`, so the validator's domain is sourced consistently across types.
- Confirmed the match semantics (`TreeNodeModel.leaf()` + `data()`/`label()`) line up with
  `LibManager.TableStyleLogicalLibrary.getByName`'s own exact-then-fuzzy-suffix (`~name`) lookup,
  so a value accepted by the validator is genuinely resolvable at render time, and vice versa.
- Confirmed `TableStylePaneModel` is a plain mutable POJO, so the "same reference regardless of
  prior patch entries" claim in the javadoc holds, and reflows through `PropertyPath.get`/`set`
  correctly (no NPE risk from `children()` defaulting to an empty list, not null).
- Confirmed `canonicalShowType` (called just before the new check) only rewrites `showType`
  fields and passes `tableStyle` through untouched, so the validated value and the persisted
  value are the same object — except for the whitespace case noted below.
- Confirmed the check only fires when `tableStyle` is actually part of the patch, so pre-existing
  dangling values already stored on assemblies that aren't part of this write are left alone (no
  new blocking behavior for untouched data).
- Test wiring (`serviceWithTable`/`serviceWithCalcTable`, new mocks spliced into the
  `AssemblyPropertyService` constructor call) matches the existing pattern used for every other
  assembly type in this test file, and the mock positions line up with the constructor's actual
  parameter order.
- Ran the stated test command mentally against the diff; the four new tests (dangling name
  refused on table/calctable, accepted by label, accepted by id, empty clears) match the four
  behaviors the javadoc claims.

### Minor, non-blocking finding
`requireKnownTableStyle` computes `String text = String.valueOf(value).trim()` and validates
against the *trimmed* text, but then writes the original, untrimmed `value` via
`PropertyPath.set(model, entry.getValue(), value)`. Two edge cases follow from this mismatch:

- A value with incidental leading/trailing whitespace around a real style name/id (e.g.
  `" Default Style"`) passes validation (matches after trim) but the untrimmed string is what
  gets persisted. `VSUtil.getTableStyle`/`LibManager` do not trim, so this would still resolve to
  `null` at render time — the exact silent-fallback failure mode this PR sets out to close, just
  for a whitespace-padded value instead of a wrong one.
- A value that is only whitespace (e.g. `"   "`) trims to empty and is treated as a legitimate
  "clear the style" request (skipped, not validated), but the untrimmed value is what's written.
  `VSUtil.getTableStyle` only treats a zero-length string as "no style"; a 3-character
  whitespace string is not zero-length, so it is not treated as cleared and still resolves to
  `null` at render.

In practice this requires a caller to submit a value with stray whitespace, which is unlikely
for an agent that round-trips names verbatim from `list_table_styles`/`get_assembly_properties`,
so I'm not treating it as blocking. Suggest either validating against the untrimmed `value` (to
keep the checked and persisted values identical) or trimming `value` itself before both the
check and the write, in a follow-up.

### Other things checked, no issues found
- No backward-incompatibility: this only tightens acceptance on writes going through the
  agent-facing `set_assembly_properties` pipeline; the interactive Composer UI's own submit flow
  is untouched.
- No security concern: the style-folder permission filtering already applied in
  `VSDialogService.getStyleTree` (via `SecurityEngine.checkPermission(...READ)`) means a caller
  cannot use this endpoint to probe for the existence of styles in folders they can't read —
  arguably a plus, not a regression.
- Error message correctly references real, existing tool names (`list_table_styles`,
  `get_assembly_properties`) used consistently elsewhere in this package.
- No test regressions expected; the changed test file only adds new tests and new mock wiring
  that mirrors existing patterns for the other assembly types.

### CI / mergeability
- `claude-review` check failure: this is the automated Claude Code Review job, which failed
  after 359ms/1 turn/$0 cost — the known signature of the review bot exhausting its shared
  subscription quota rather than a real finding. This manual review supersedes it, per this PR's
  review process.
- `Semgrep Scan`, `Semgrep OSS`, `build` (Validation Build), and `license/cla` all pass.

### Verdict: Approve
No blocking issues, no failing checks other than the excluded `claude-review` quota artifact, and
no merge conflicts.

jshobe-inetsoft (approved):

PR #5336, opened 2026-09-17
feat(wiz): add Script Library structural CRUD for admin-chat (Redmine #76719 Gap 1)

NeverStopFMH:
## Summary
- New `inetsoft.web.admin.ai.scriptlibrary` package (Controller/Service/ChangePlanService/ChangesetApplyService), the same pattern Viewsheets/RecycleBin already use, wrapping `LibManager` directly rather than proxying `RepositoryScriptController` (session-cookie/CSRF EM SPA endpoint) or composer-chat's own `inetsoft.web.wiz.controller.ScriptLibraryController` (wrong permission model — `READ`/`WRITE` for a business user, not `ADMIN` for a site administrator).
- Supports `create`/`rename`/`update` (description only)/`delete`, all with a real, working rollback — a script's own text/description is captured before mutating and replayed on rollback. `delete` is still always classified `RISK_HIGH` and requires `acknowledgeIrreversibleDelete` unconditionally (matching Viewsheets'/data sources'/MV's own "delete is always high risk" convention), since anything calling the script by name breaks for as long as the deletion stands, even though the content itself is recoverable.
- The dependency check (`force`) mirrors composer-chat's `ScriptLibraryController.delete` exactly, replicated here since this area calls `LibManager` directly.
- Community-tier, not enterprise-gated — `LibManager` is itself community-tier, matching RecycleBin/MV/repository export-import's own placement.
- Companion PR on `stylebi-wiz` adds the matching admin-chat MCP tools: https://github.com/inetsoft-technology/stylebi-wiz/pull/2641

## Test plan
- [x] `mvnw -pl core compile` — clean compile of the whole `core` module
- [x] Live-tested end to end against a running local instance (`http://localhost:8080/sree`):
  - `list_script_library`/`get_script_library_entry` read correctly
  - `create` → preview → apply → `verified`, Tier-2 backup taken
  - `rename` → old name gone, new name resolves with identical content
  - `update` → description-only change, body text untouched
  - `delete` dependency check refuses loud naming a real dependent; `force:true` proceeds with advisory
  - Plan-drift protection: apply re-validates the whole plan and refuses atomically (zero side effects) when an entry has drifted since preview
  - Cross-verified against composer-chat's `read_script_library_function` — confirms both surfaces share the same underlying `LibManager` asset


NeverStopFMH:
## Round 1 review

Head commit reviewed: `8b3e71c459c6ae5d797fd7abd7af62fb01160103`

Build: `mvnw.cmd -pl core compile -o -q` from `E:/inetsoft/stylebi-github/stylebi/community` — clean, exit 0, no output (offline, using the populated `~/.m2` cache). Confirms the PR description's compile claim independently.

Also read the companion PR `inetsoft-technology/stylebi-wiz#2641` in full (`plugin/admin/src/tools/scriptLibraryTools.ts`, `plugin/admin/src/tools/normalize.ts`) and cross-checked every wire field (`verb`/`name`/`newName`/`description`/`text`/`force`/`task`/`changes`/`planHash`/`reviewOutcome`/`acknowledgeIrreversibleDelete`) and every endpoint path (`GET /v1/admin/script-library`, `GET /v1/admin/script-library/entry`, `POST /v1/admin/script-library/preview`, `POST /v1/admin/script-library/apply`) against the Java side — all match exactly, including per-verb "unused field" refusal parity between `normalize.ts`'s `readScriptLibraryChange` and `ScriptLibraryChangePlanService`'s `requireUnused` calls.

### What I checked, per invariant

**1. Talks to `LibManager` directly, never proxies `RepositoryScriptController`/composer-chat's `ScriptLibraryController`.** Confirmed — `ScriptLibraryService`/`ScriptLibraryChangesetApplyService` only import `LibManager`/`LibManagerProvider`, no controller class. Tried to construct a counter-example where some path falls back to a controller call (e.g. via `DependencyHandler`) — could not; `dependents()` uses `DependencyTransformer.getDependencies` directly, matching composer-chat's own private helper byte-for-byte (same `AssetEntry(COMPONENT_SCOPE, Type.SCRIPT, name, null)`).

**2. Permission is `ADMIN` on `ResourceType.SCRIPT`, matching `RepositoryScriptController`, at every mutation path.** **False — see Finding 1 below.** `rename`/`update`/`delete` all call `scriptLibraryService.requirePermission(name, user, ResourceAction.ADMIN)` in `ScriptLibraryChangePlanService`; `create` never calls it, at either preview (`resolveCreate`) or apply (`applyCreate`) time. Composer-chat's own `ScriptLibraryController.create()` — the precedent for "check permission even before the resource exists" — does call `requirePermission(principal, name, WRITE)`. Confirmed via `SecurityEngine.checkPermission` (lines ~869-895) that `ResourceType.SCRIPT` + `ADMIN` on a **non-existent** name still resolves meaningfully (falls back through `getPermission(SCRIPT, name) == null` to the parent `SCRIPT_LIBRARY` resource's own grant) — so this isn't dead code being skipped, it's a real check being skipped only for `create`.

**3. `delete` unconditionally `RISK_HIGH`, fully reversible; rollback correctness for mixed plans.** Walked a hand-built mixed plan (create A → rename A→B → update B.description → delete C) through `applyOne`/rollback by hand: if the delete of C fails at apply time, `rollback()` iterates `undoable` in reverse — undo-update(B) restores B's captured pre-rename description, undo-rename(B→A) moves back and restores C's... no, restores A's permission grant, undo-create(A) removes A. Each step's captured state (`Undo.rename`/`Undo.update`/`Undo.create` all capture values at the point THAT step ran, post-earlier-mutations) is internally consistent — tried to break this with a name collision mid-sequence (rename target already used by a later step) and could not, because `claim()` scopes keys to the change's own `name` field, and each entry is validated against the LIVE store, not a simulated intermediate state (this matches every other area in this package — chaining a create/rename with a same-plan follow-up addressed by the NEW name would fail at `resolve()` time with a `MissingResourceException`, since preview never simulates intermediate application; this is a pre-existing, package-wide limitation, not something unique to this PR, so not flagged as a new finding). Rename's own permission move (`getPermission`/`removePermission`/`setPermission`) IS null-checked on both apply and rollback (`if(permission != null)`) — actually stricter than the `RepositoryScriptController.setScriptModel` precedent it's modeled on, which moves unconditionally with no null check. Not a defect (an improvement), noting it since the task asked to verify this specifically.

**4. Dependency check mirrors composer-chat's `ScriptLibraryController.delete` exactly.** Confirmed line-for-line identical `AssetEntry`/`DependencyTransformer` construction between `ScriptLibraryService.dependents()` and the composer-chat controller's own inline check.

**5. Every apply-time check re-derived fresh.** Confirmed for existence/collision on create/rename and dependents on delete (`applyCreate`/`applyRename`/`applyDelete` each re-query `LibManager`/`dependents()` rather than trusting the plan). Permission is checked ONLY at preview/resolve time, never re-checked at apply — same as every other verb in this package (not a new gap), except create never checks it at either time (Finding 1).

**6. Plan-hash folds `text` into a SHA-256 digest.** Confirmed (`ScriptLibraryChangePlanService.digest`/`project`). Tried "digest collision hides a real drift" scenario — not exploitable for drift-detection purposes (a SHA-256 preimage/collision attack is out of scope for a plan-integrity hash of this kind, consistent with every other area's own hash design).

**7. Community-tier, no enterprise imports.** Confirmed — package only imports `inetsoft.report.*`, `inetsoft.sree.security.*`, `inetsoft.uql.asset.*`, `inetsoft.util.*`, `inetsoft.web.admin.ai.*`/`inetsoft.web.security.*`, all community-tier.

**Concurrency (`APPLY_LOCK`).** Confirmed structurally identical to `RecycleBinChangesetApplyService`'s own lock — same static `ReentrantLock`, same `lock()`-before-resolve/`finally unlock()` critical section covering resolve+hash-check+mutate. No divergence found.

**Repeated `apply` with the same `planHash`.** Traced: the second call's `planService.resolve()` re-derives the hash from current (already-mutated) state, which will differ from the stale `req.getPlanHash()`, so it throws `PlanHashMismatchException` → HTTP 409 with the fresh plan attached — same as every other area. No confusing-message issue found.

**Boundary/shape (CJK names, empty description, `newName == name`).** `requireNonBlank` trims both `name`/`newName` before the equality check, so the "must differ" check is airtight against leading/trailing-whitespace-only differences. Did not find any ASCII-only or byte-length validation anywhere in the reviewed code (`canonical()`/`digest()` explicitly handle `null` and UTF-8) — nothing in this PR's own code imposes or violates a CJK/Unicode constraint; whether `LibManager`'s underlying storage handles it is outside this PR's diff.

### New findings

**Finding 1 (Medium) — `create` never checks `ADMIN` permission on the target script name, unlike every other verb, contradicting the PR's own stated invariant.**
`core/src/main/java/inetsoft/web/admin/ai/scriptlibrary/ScriptLibraryChangePlanService.java`, `resolveCreate` (lines 114-134), and `ScriptLibraryChangesetApplyService.applyCreate` (lines 203-238): neither calls `scriptLibraryService.requirePermission(...)`, while `resolveRename`/`resolveUpdate`/`resolveDelete` all call `requirePermission(name, user, ResourceAction.ADMIN)` (lines 153, 185, 204). The class javadoc on `ScriptLibraryService` states "an admin-chat caller never sees or touches a script it could not already administer" — that's false for `create`. `SecurityEngine.checkPermission(user, SCRIPT, nonExistentName, ADMIN)` (verified by reading `SecurityEngine.java` lines 845-895) does NOT trivially return true for a nonexistent resource — it falls through to the resource's parent (`SCRIPT_LIBRARY`) grant, so this is a real, meaningful check being skipped, not dead code. Composer-chat's own `ScriptLibraryController.create()` calls `requirePermission(principal, name, WRITE)` for exactly this reason, confirming the precedent this PR claims to follow does check create. Concretely: a site administrator whose org/role has been explicitly denied `ADMIN` on the Script Library root (a plausible delegated-admin scenario mentioned nowhere as out of scope) can still create arbitrary new scripts through this tool, even though they cannot rename/update/delete anything. Fix: add `scriptLibraryService.requirePermission(name, user, ResourceAction.ADMIN)` in `resolveCreate` (mirroring the other three verbs).

**Finding 2 (Low-Medium) — zero Java-side tests for the whole new package.**
`find core/src/test -iname "*ScriptLibrary*"` only turns up the pre-existing, unrelated `core/src/test/java/inetsoft/web/wiz/controller/ScriptLibraryControllerTest.java` (composer-chat's own controller). There is no test file anywhere under `core/src/test/java/inetsoft/web/admin/ai/scriptlibrary/` for any of the 4 service/controller classes this PR adds — no coverage of resolve/apply/rollback for any verb, the permission gap above, the plan-hash digest, or the rename permission-move sequence. The companion `stylebi-wiz` PR has real unit tests for its own normalization layer (`scriptLibraryNormalize.test.ts`), but nothing on the StyleBI side exercises the actual `LibManager`-backed logic this PR adds. Per the review checklist, this is itself a finding, not merely a nice-to-have.

### Verdict

Changes requested — Finding 1 is a real, source-confirmed inconsistency against this PR's own stated design invariant (not a hypothetical), and Finding 2 is a genuine test-coverage gap for a brand-new package with real rollback/concurrency logic.

NeverStopFMH:
Addressed both round-1 findings in 70a45b83f.

**Finding 1 (permission gap on create) — fixed.**

Added `scriptLibraryService.requirePermission(name, user, ResourceAction.ADMIN)` to `resolveCreate` in `ScriptLibraryChangePlanService.java`, positioned right after the name is claimed (matching the existing verbs' ordering: check the name, claim it, then gate permission before doing anything more expensive/informative). Regression-tested in `ScriptLibraryChangePlanServiceTest#resolveCreateRefusesWhenCallerLacksAdminPermission` — confirmed by temporarily reverting the fix that this test fails without it (`AssertionFailedError: Expected java.lang.SecurityException to be thrown, but nothing was thrown`) and passes with it.

**Pushback on the "defense-in-depth in applyCreate" half of Finding 1.** I did not add a second permission check to `applyCreate`. Traced the actual call chain: `ScriptLibraryChangesetApplyService.apply()` calls `planService.resolve(req, user)` as its very first line — with the live apply-time user and current server state — before doing anything else, and *that* resolve is what's hash-checked and then executed. So the new `resolveCreate` permission check is already re-derived fresh at apply time, automatically, via this call. I also checked whether `applyRename`/`applyUpdate`/`applyDelete` re-check permission a second time themselves (the "never trust what was computed at preview" pattern the finding cites) and they do not — none of the four `apply*` methods in this file calls `requirePermission`/`checkPermission` directly. I extended the check to every sibling `*ChangesetApplyService.java` under `core/src/main/java/inetsoft/web/admin/ai/` (recyclebin, viewsheet, datasource, shapes, dashboard, mv, licensing, providers, schedule, presentation, cluster, schedulerstatus) — none of them re-check permission inside their own apply-time verb methods either; all of them rely on the same "apply() re-resolves the plan fresh" mechanism. So a second call in `applyCreate` would be both redundant and inconsistent with the actual convention in this codebase (as opposed to the convention the finding described, which turned out not to exist here). Happy to add it anyway if you still want literal per-verb defense-in-depth as a deliberate new precedent, but as written it wouldn't change behavior.

**Finding 2 (no Java tests) — added, findings on the "no comparable tests exist" hypothesis.**

Checked first: comparable Java unit tests absolutely DO exist for every sibling admin-ai area (`recyclebin`, `viewsheet`, `shapes`, `datasource`, `dashboard`, `providers`, `schedule`, `identities`, `mv`, `licensing`, `presentation`, `permissions`, `cluster`, `schedulerstatus`, `file` all have `*ChangePlanServiceTest`/`*ChangesetApplyServiceTest` pairs under `core/src/test/java/inetsoft/web/admin/ai/`) — so this area's total absence of tests was the outlier, not the norm; I did not invent an untested pattern from scratch.

Added `ScriptLibraryChangePlanServiceTest.java` (16 tests) and `ScriptLibraryChangesetApplyServiceTest.java` (12 tests), modeled on `ShapeChangePlanServiceTest`/`ShapeChangesetApplyServiceTest` for structure (in-memory fake collaborator + Mockito) and on `ScriptLibraryControllerTest` (the composer-chat companion controller this whole area is modeled on) for how to mock `LibManager`/`LibManagerProvider`/`SecurityEngine`/`DependencyTransformer` specifically. Coverage: per-verb `PlanChange`/risk classification (create/rename/update always LOW, delete always HIGH) for all four verbs, the Finding-1 permission-gap regression test described above, the dependency-check refuse/force-allow gate on delete (both at preview and carried through to apply), request-shape validation, plan-hash-mismatch/reviewOutcome/acknowledgeIrreversibleDelete apply-time gates, a successful apply for every verb (including the rename permission-grant move and the update-never-touches-text guarantee), and two rollback scenarios (a verified create undone when a later change fails verification without throwing, and a rollback-failed case where the undo itself throws).

One incidental finding surfaced while writing these tests, noted here rather than silently worked around: for `rename`/`update`/`delete`, a caller denied `ADMIN` never actually reaches those verbs' own explicit `requirePermission` call — `requireEntry` → `getEntry` → `isVisible` already gates on the identical `ResourceAction.ADMIN` check first, and `getEntry` deliberately folds "does not exist" and "not visible" into one `MissingResourceException`. So a permission denial on those three verbs surfaces as "no script library entry named ..." rather than a raw `SecurityException`, and the later explicit `requirePermission` call in each of those three methods is effectively unreachable when denied (it can only ever pass, since it re-checks the same thing `isVisible` already gated on). This matches `RecycleBinService.getEntry`'s documented "not found or not permitted" posture and isn't a security hole (the caller is still correctly refused, just with a different exception type/message than `create`'s), so I left it as-is and just wrote the tests to assert the real behavior — flagging it here in case the difference in exception type ever surprises a caller trying to distinguish "not found" from "not permitted" for rename/update/delete specifically.

**Compile/test results:**
- `mvnw.cmd -pl core compile -o -q` — clean
- `mvnw.cmd -pl core test-compile -o -q` — clean (full core test tree, not just this package)
- `mvnw.cmd -pl core test -o -q -Dtest=ScriptLibraryChangePlanServiceTest,ScriptLibraryChangesetApplyServiceTest,ScriptLibraryControllerTest` — 28/28 pass

Commit: 70a45b83f, pushed to `feature/76719-admin-script-library`.

NeverStopFMH:
## Round 2 review

**Head commit reviewed:** `70a45b83f` (range since round 1: `8b3e71c45..70a45b83f`)

### What I checked

- **Finding 1 fix, read in context**: `resolveCreate` in `ScriptLibraryChangePlanService.java` now calls `scriptLibraryService.requirePermission(name, user, ResourceAction.ADMIN)` at line 121, right after the duplicate-claim check and before the `exists()`/collision check — so a denied caller never learns whether the name is taken. Same `ResourceAction.ADMIN`/`ResourceType.SCRIPT` pair `resolveRename`/`resolveUpdate`/`resolveDelete` already use. The whole production diff since round 1 is this single added line — confirmed via `git diff 8b3e71c45..70a45b83f -- core/.../scriptlibrary/`.

- **Independently verified the `applyCreate` pushback** (did not just trust it): `ScriptLibraryChangesetApplyService.apply()`'s first statement inside the lock is `planService.resolve(req, user)` — this re-runs `resolveCreate` (now with the permission check) on every `apply()` call, not just preview, so the fix is enforced at apply time too. I independently checked 4 sibling `*ChangesetApplyService.java` files (Shapes, RecycleBin, Viewsheet, DataSource — more than the 2 requested), not just trusting the fixer's count: each calls its own plan service's `resolve`/`resolveEntries` as apply's first step, and a `grep -n "requirePermission\|checkPermission\b"` across all 6 sibling apply-service files (adding Dashboard, MV) returned **zero matches** — none of them re-check permission inside their own apply-verb methods. This corroborates the fixer's claim. **Agree with the pushback**: no redundant check needed in `applyCreate`.

- **Actively probed the regression test for theater, by actually doing it**: commented out line 121 (`scriptLibraryService.requirePermission(...)` in `resolveCreate`), ran `mvnw.cmd test -o -q -Dtest=ScriptLibraryChangePlanServiceTest` from `core/`. Result: **exactly one failure** — `resolveCreateRefusesWhenCallerLacksAdminPermission` (`Expected java.lang.SecurityException to be thrown, but nothing was thrown`), all other 15 tests in the class still passed. Reverted the comment-out (`git diff` on the file is now clean) and re-ran both new test classes: **28/28 pass, 0 failures** — matches the fixer's claim exactly (16 + 12).

- **Checked the tests for non-trivial assertions**, not just "no exception thrown": `resolveRenameProducesARiskLowChangeAndRequiresPermission` asserts `checkPermission` was called exactly `times(2)` (once via `requireEntry`→`isVisible`, once via the explicit call) — verifies the *shape* of the check, not just its outcome. `rollsBackAnEarlierVerifiedCreateWhenALaterDeleteFailsVerification` asserts real pre/post state on the in-memory `LibManager` fake: the rolled-back create's script is gone (`assertFalse(scripts.containsKey("newScript"))`) AND the pre-existing script that failed verification is still present (`assertTrue(scripts.containsKey("keep"))`) — a genuine before/after check, not a bare no-throw assertion. `appliesARenameAndMovesAnExistingPermissionGrant` verifies the permission grant was actually moved via `verify(securityEngine).removePermission(...)`/`setPermission(...)`. No theater found.

- **Ran `mvnw.cmd -pl core compile -o -q`** from `community/` — clean, no output/errors.

- **Assessed the exception-type asymmetry with evidence, not intuition**: traced both paths.
  - `ScriptLibraryService.requirePermission` throws a bare, unqualified `SecurityException` in a file with no `inetsoft.sree.security.SecurityException` import — this resolves to **`java.lang.SecurityException`**.
  - `AdminExceptionHandler` (`@ControllerAdvice(basePackages = {"inetsoft.web.admin", ...})`, which covers `AdminScriptLibraryController`'s package) has an explicit `@ExceptionHandler(java.lang.SecurityException.class)` → `handleAccessDenied` → **HTTP 403**, `GenericError` body, sanitized message ("Access was denied...").
  - rename/update/delete's permission-denied path never reaches `requirePermission` at all in that case — `requireEntry`→`getEntry`→`isVisible` (same `ADMIN` check) throws `MissingResourceException` first, folding "not found" and "not visible" together → caught by the same `AdminExceptionHandler.handleMissingResource` → **HTTP 404**, `MissingResourceError` body, deliberately ambiguous message ("...may not exist... or you may not have permission to see it").
  - This is a **materially different HTTP status code and response shape** for the identical "denied ADMIN" condition, not merely a different exception class name that maps the same way — correcting the "purely cosmetic" framing.
  - Checked the caller-facing MCP layer (`plugin/admin/src/tools/scriptLibraryTools.ts`, companion `stylebi-wiz` repo): `apply_script_library_changes` only special-cases `HTTP_409`; any other status (403/404/500) is simply rethrown untouched — so this asymmetry is **not** papered over at the MCP tool layer either; it reaches the calling LLM agent as-is.
  - That said, this is **not a security hole** (both paths correctly refuse the action) and appears structurally hard to fully eliminate: `create`'s target doesn't exist yet, so there's no entry to "fold" a not-found/not-visible ambiguity with the way the other three verbs can — the two verbs are asymmetric by nature, not just by an oversight in this fix. Severity: low.

### Round-1 findings status

1. **Finding 1 (missing ADMIN check on create): FIXED**, verified live via the comment-out/rerun probe above.
2. **Finding 2 (zero Java tests): FIXED**, 28 substantive tests added and independently confirmed passing (`mvnw.cmd test -o -q -Dtest=ScriptLibraryChangePlanServiceTest,ScriptLibraryChangesetApplyServiceTest,ScriptLibraryControllerTest` → 16+12+4, all pass).

### New finding (this round)

- **Low severity, non-blocking-for-security**: the create-vs-rename/update/delete exception-type asymmetry the fixer disclosed is real, not cosmetic — it produces a different HTTP status (403 vs 404) and response body shape for the identical permission-denied condition, and this is visible unchanged all the way to the admin-chat MCP tool caller. Recommend either documenting this explicitly in `AdminScriptLibraryController`'s javadoc (it currently isn't mentioned there) or leaving it as an accepted, disclosed limitation — no code change is strictly required given the structural asymmetry between create and the other three verbs, but the "purely cosmetic" characterization should be corrected to "real but low-severity and largely inherent."

Both round-1 findings are conclusively fixed and well-tested. The only open item is the new low-severity finding above, which I'm flagging per the review protocol rather than silently accepting — happy to see this land as "approved" once it's either documented or explicitly accepted as a known limitation in a short follow-up comment/commit.

NeverStopFMH:
Investigated the exception-type asymmetry (create -> 403 via `SecurityException` vs rename/update/delete -> 404 via `MissingResourceException`) with the RecycleBin precedent as instructed. Conclusion: **by design, not a defect — no code change.**

**RecycleBin's own rationale** (`RecycleBinService.java`):
- `getEntry` javadoc: "a caller must not be able to distinguish 'does not exist' from 'exists, but you cannot see it'"
- `requireEntry` javadoc: throws `MissingResourceException` "if `path` is absent from the recycle bin, or present but not visible to `user`"

**Does that rationale transfer to Script Library?** Yes, and it isn't a RecycleBin-only pattern — `ScriptLibraryService.getEntry`/`requireEntry` explicitly cite it as their own precedent ("same posture `RecycleBinService.getEntry` documents for its own area"), and the plugin surface built on top already commits to it for reads, undisputed: `get_script_library_entry`'s tool description states plainly that `found:false` is "true both when no script exists by that name AND when one exists but is not visible to the caller." Nobody is proposing to change that read contract, and rename/update/delete's gate is the identical check (`isVisible` = ADMIN on `ResourceType.SCRIPT`) reached through `requireEntry`.

**Per-entry ownership/scoping**: Script Library entries carry none of RecycleBin's per-owner scoping (RecycleBin's `isVisible` checks ADMIN on `SECURITY_USER` for a recycled item's *original owner* when it's user-scoped). Script Library entries are a flat, non-identity-scoped namespace (`LibManager`) gated by a plain per-name `ADMIN` grant on `ResourceType.SCRIPT` — the same resource/action pair `RepositoryScriptController`'s own EM settings-page editor already uses. So the privacy payoff here isn't "hide which user has what," it's the more general "don't disclose whether a permission-gated named resource exists to a caller who isn't permitted to see it" — which is exactly what a 404-that-could-mean-either-thing accomplishes, and exactly what `create`'s own fix accomplishes by a *different* mechanism: `resolveCreate` calls `requirePermission` **before** `exists()` specifically so a denied caller learns nothing about whether the name is taken (see the round-1 regression test `resolveCreateRefusesWhenCallerLacksAdminPermission`, which asserts `lib.getScript` is never invoked when permission fails).

So both branches protect the same fact (existence, from an unauthorized caller) via different plumbing that both verbs' own shapes force: `create` has no existing resource whose existence needs hiding *after* the check, so ordering suffices; `rename`/`update`/`delete` operate on a name that may or may not already exist, so the fold-into-404 is what hides it. `AdminExceptionHandler.handleCheckedAccessDenied`/`handleAccessDenied` additionally sanitize *every* `SecurityException` message to a generic i18n string regardless of area, specifically so no permission/role/principal detail leaks — so `create`'s 403 is exactly as uninformative about the *specific* denial as rename/update/delete's ambiguous 404 is about existence. Neither branch discloses more than the other; they just guard different facts because there's a different fact to guard in each case.

This exact question was also already explicitly raised and resolved during round 1 — see `ScriptLibraryChangePlanServiceTest.java`'s own javadoc on `resolveRenameRefusesWhenCallerLacksAdminPermission`/`resolveDeleteRequiresAdminPermission`: "documented, matching `RecycleBinService.getEntry`'s own posture ... Verified real behavior, not a bug."

I also checked the premise about the companion `stylebi-wiz` plugin promising a uniform error contract across verbs — I don't find such a promise in `plugin/admin/src/tools/scriptLibraryTools.ts`; it special-cases only `HTTP_409` (plan-hash drift, expected control flow), and otherwise both a 403 and a 404 simply propagate as tool errors, each already carrying an explanatory (if intentionally non-specific) message for the calling agent.

**No code change made.** Happy to revisit if there's a concrete scenario where the 404 path actually discloses something the 403 path doesn't (or vice versa) that I'm missing.

NeverStopFMH:
## Round 3 review

**Head reviewed:** `70a45b83f` (unchanged since round 2 — confirmed via `gh pr view --json headRefOid`).

### Independent verification of the round-2 fixer's "by design" pushback

I re-read every cited artifact directly rather than trusting the citation:

- **`RecycleBinService.getEntry`/`requireEntry`** (lines 80-103): javadoc confirms the "must not distinguish does-not-exist from exists-but-not-visible" posture, folding both into `Optional.empty()`/`MissingResourceException`.
- **`ScriptLibraryService.getEntry`/`requireEntry`** (lines 81-104): javadoc explicitly cites `RecycleBinService.getEntry` as its own precedent for the identical fold. Confirmed `exists()` (line 110-112) is a *separate* method used only by `create`'s collision check, deliberately checking existence "regardless of visibility."
- **`AdminExceptionHandler`**: confirmed `handleAccessDenied`/`handleCheckedAccessDenied` (lines 149-167) both route through the private `accessDenied()` helper (line 174), which returns a **generic, localized, non-parameterized** message (`Catalog.getString("http.error.unauthorized")`) for `SecurityException` (both checked and unchecked variants) — no principal/name/permission detail. I also checked `handleMissingResource` (line 58) → `new MissingResourceError(e)`, and confirmed `MissingResourceError` (extends `ApiError`) surfaces `e.getMessage()` **verbatim** — i.e. the crafted-vague text from `requireEntry`'s own message ("...it may not exist, may already have been deleted/renamed, or you may not have permission to see it"), not a further-sanitized generic string. One correction to the fixer's framing: `AdminExceptionHandler`'s two access-denied handlers sanitize only the `SecurityException` path; the `MissingResourceException` path was never routed through a *sanitizing* handler in the first place — it was already non-disclosing by construction (the message text itself refuses to commit to a reason), and I additionally verified `AdminScriptLibraryController` declares its own local `@ExceptionHandler(MissingResourceException.class)` (lines 137-142) that also just echoes `ex.getMessage()` — same non-disclosing content, different wrapper shape (`Map` vs `MissingResourceError`), so no behavioral difference from what the global advice would have done.
- **`ScriptLibraryChangePlanServiceTest`**: read the actual test bodies, not just line numbers. `resolveCreateRefusesWhenCallerLacksAdminPermission` (line 189) asserts `SecurityException` AND `verify(lib, never()).getScript("newScript")` — i.e. it is test-enforced, not just argued in prose, that a denied `create` caller never reaches the existence check. `resolveRenameRefusesWhenCallerLacksAdminPermission` (line 230, javadoc 221-229) and `resolveDeleteRequiresAdminPermission` (line 325, javadoc 323-324) both assert `MissingResourceException` for the identical lacks-ADMIN condition, and their javadocs already state "Verified real behavior, not a bug."

### The constructed scenario

Per the task: an unauthorized caller (no ADMIN on resource `SCRIPT:X`) tries `create` with a name `X` that already exists (visible to them or not).

- `resolveCreate` (`ScriptLibraryChangePlanService.java` lines 114-135) calls `requirePermission` **before** `exists()`. An unauthorized caller throws `SecurityException` → sanitized 403 unconditionally, and — confirmed by the `never()` verification in the round-1 regression test — never executes the `exists()` branch at all. So the 403 leaks nothing about whether `X` was taken.
- Compare `rename`/`update`/`delete`: an unauthorized caller gets `MissingResourceException` → 404 with the same vague wording **whether or not** the name exists (`resolveRenameFailsWithMissingResourceExceptionWhenNameDoesNotExist` for the true-absent case, `resolveRenameRefusesWhenCallerLacksAdminPermission` for the exists-but-denied case — identical exception type and message shape in both).

So, per-operation, existence is never observable to an unauthorized caller in either family — the privacy property holds on both sides via different mechanisms, exactly as claimed.

**Residual asymmetry (not a leak):** the HTTP *status code* itself still differs across verbs for the identical "lacks ADMIN" condition — `create` → 403, `rename`/`update`/`delete` → 404. This is a real, observable inconsistency (a client can't rely on a uniform status code for "permission denied" across this API surface), but it doesn't disclose anything an attacker couldn't otherwise learn, is consistent within each verb, and mirrors the same fold-vs-order-based-privacy split already established by `RecycleBinService` elsewhere in the codebase. I consider this cosmetic/API-consistency, not a security defect.

### Verdict on the exception-asymmetry finding: **withdrawn**

I agree with the fixer's round-2 pushback after independently re-deriving the evidence rather than accepting the citations at face value. No further action requested on this finding.

### Broader sweep (since HEAD is unchanged, this is the substance of round 3)

- Re-ran `mvnw.cmd -pl core compile -o -q` from `community/` — clean, no output, exit 0.
- Re-ran `mvnw.cmd -pl core test -o -q -Dtest=ScriptLibraryChangePlanServiceTest,ScriptLibraryChangesetApplyServiceTest,ScriptLibraryControllerTest` from `community/` — all green: `ScriptLibraryChangePlanServiceTest` 16/16, `ScriptLibraryChangesetApplyServiceTest` 12/12, `ScriptLibraryControllerTest` (pre-existing composer-chat controller, unrelated to this PR) 4/4. 0 failures, 0 errors.
- Confirmed the PR's actual file-level diff (`AdminScriptLibraryController`, `ScriptLibraryService`, `ScriptLibraryChangePlanService`, `ScriptLibraryChangesetApplyService`, DTOs, and their tests — 15 files, all new/additive) is unchanged since round 2.
- One very minor, non-blocking observation: `resolveUpdate` shares the exact same `requireEntry` → `requirePermission` code path as `resolveRename`/`resolveDelete`, but unlike those two there is no dedicated `resolveUpdateRefusesWhenCallerLacksAdminPermission` regression test (only `resolveUpdateProducesARiskLowChange`/`resolveUpdateRequiresADescription`). Given the path is byte-identical to the already-tested rename/delete cases, I don't consider this worth blocking on — flagging only for optional completeness, not as a finding requiring a fix.
- No other new findings.

**Overall verdict: approved.**

jshobe-inetsoft (approved):

PR #5337, opened 2026-09-17
fix(76741,VSC-008): give viewsheet condition junctions an independent level

fredaan:
## Summary

Fixes Redmine #76741 / VSC-008: a viewsheet condition's `level` field, which is supposed to let a caller override AND/OR grouping (e.g. force `(A OR B) AND C` instead of default AND-before-OR precedence), silently had no effect at all.

**Root cause** (two parts):
1. `ConditionVocabulary.junction()` hardcoded every junction's level to `0`, regardless of the caller's input.
2. `HierarchyList.validate()` — invoked by every real `ConditionGroup` construction path in the codebase (crosstab post-conditions, ranking conditions, highlight conditions, order-info conditions, plain table/chart row filtering) — unconditionally re-derives every condition's level from its neighboring junctions' levels. So even the condition's own `level` (which *was* passed through) got silently overwritten back to 0 by every junction being hardcoded to 0.

No exception was thrown; conditions silently evaluated with flat/AND-before-OR precedence no matter what `level` the caller set.

**Fix** — give the junction an independently-settable level rather than deriving it purely from its two flanking conditions, since a levels-only formula can't distinguish a two-independent-groups boundary junction (e.g. `(A OR B) AND (C OR D)`) from an in-group one — both flanking conditions sit at the same level in the ambiguous case, so no pairwise function of just those two levels can ever produce the correct answer for both shapes.

## Changes

- `ConditionVocabulary.Clause` (`community/core/src/main/java/inetsoft/web/wiz/viewsheet/ConditionVocabulary.java`) gains a new, nullable `junctionLevel` field, appended at the end of the record with a compatible 7-arg auxiliary constructor (matching this repo's own `EditRequest.java` precedent for backward-compatible record field additions), so the ~15 existing call sites keep compiling unchanged.
- `ConditionVocabulary.junction()` now uses `clause.junctionLevel()` when given, otherwise defaults to `Math.min(clause.level(), nextClause.level())` — the **minimum**, not maximum, of the two flanking levels. (An earlier draft of this fix used `Math.max`, which was caught and corrected during review since it collapses the exact motivating example back to flat/level-0 — see the diagnosis trail linked below.)
- `ViewsheetAssemblyAgentController.ConditionClause` (the record Jackson binds a `set_condition` request body's `conditions` array into) gains the same `junctionLevel` field, threaded through `toClause()`'s call into `ConditionVocabulary.Clause` — without this, the field would never reach `ConditionVocabulary` at all.
- Regression tests added to `ConditionVocabularyTest.java`:
  - `defaultJunctionLevelIsMinOfFlankingLevelsNotMax` — the higher-value test, since most real callers will hit the default-inference path, not the explicit one; asserts the withdrawn `Math.max` bug doesn't regress.
  - `explicitJunctionLevelProducesTwoIndependentGroups` — builds `(A OR B) AND (C OR D)` through the real `ConditionVocabulary` → `ConditionList`/`ConditionItem`/`JunctionOperator` → `ConditionGroup` pipeline (including `HierarchyList.validate()`) and asserts the resulting evaluation matches the correct 4-input truth table.

## Diagnosis trail

This fix implements the diagnosis in the `stylebi-wiz` repo, which went through two full rounds of independent adversarial review (debugger → refuter → revision → recheck) before implementation:
- `docs/teams/2026-09-17-bug-76741-vsc-remaining/bug-vsc008/01-diagnosis.md` (see especially the "Revision round 1" section for the corrected, final fix spec)
- `docs/teams/2026-09-17-bug-76741-vsc-remaining/bug-vsc008/02-refute.md` (final recheck: "Clean — no blocking problems found in Revision round 1. Ready for the fixer phase.")

## Test plan

- [x] `./mvnw test -pl core -Dtest=ConditionVocabularyTest` — 52/52 pass (50 pre-existing + 2 new)
- [x] `./mvnw test -pl core -Dtest=ParameterDiscoveryServiceTest` — 5/5 pass (unrelated, pre-existing content in this branch's history — confirmed unaffected)
- [x] `./mvnw compile test-compile -pl core -am` — full module compiles clean, confirming the ~15 existing `ConditionVocabulary.Clause` call sites (2 production, ~13 test) are unaffected by the record's new field


fredaan:
## Review round 1 — PR #5337 (bug 76741 / VSC-008)

**Head commit reviewed:** `909db6032` (single commit, `fix/76741-vsc008` → `stylebi-wiz-main-rebase`)

**What was checked:**
- Diff read against `01-diagnosis.md`'s final, corrected spec ("Revision round 1") and `02-refute.md`'s final recheck, not the superseded Tier 1 text earlier in the same diagnosis file.
- Confirmed the default formula is `Math.min(clause.level(), nextClause.level())`, not the withdrawn `Math.max`.
- Confirmed `junctionLevel` threads end-to-end: `ViewsheetAssemblyAgentController.ConditionClause` → `toClause()` → `ConditionVocabulary.Clause` → `toConditionList()`'s call site (now passing `clause`+`clauses.get(i+1)`) → `junction()`.
- Confirmed the auxiliary 7-arg `Clause` constructor delegates to the 8-arg canonical one with `junctionLevel = null`, matching the `EditRequest.java` precedent the refuter verified.
- Ran `./mvnw test -pl core -Dtest=ConditionVocabularyTest -q` myself (not trusting the fixer's report): **52/52 pass, 0 failures/errors** (50 pre-existing + 2 new).
- Ran `./mvnw compile test-compile -pl core -am -q` myself: **exit 0**, confirming the ~15 existing `ConditionVocabulary.Clause` call sites (2 production, ~13 test) still compile.
- Walked the sibling scenarios per the review discipline: other call sites of `junction()`/`Clause` (only one production call site, confirmed by grep + successful compile), the read-path sibling of the write-path this PR touches, boundary shapes (single clause, no-junction).

**New finding this round:**

### Critical (confidence ~92) — `describe()` never surfaces a junction's own level, so `get_condition` → `set_condition` silently drops an explicit `junctionLevel` and reintroduces the exact bug this PR fixes

`ConditionVocabulary.describe()` (`core/src/main/java/inetsoft/web/wiz/viewsheet/ConditionVocabulary.java:117-141`) builds each returned clause map from a `ConditionModel`'s own `level` only (`clause.put("level", condition.getLevel())`, line ~131) and gets the `junction` token from `junctionAfter()` (line 446), which reads only `JunctionOperatorModel.getType()` — never `.getLevel()`. So the junction's own level (the new field this PR introduces sensitivity to) is **never exposed** in `get_condition`'s output, and there is no `junctionLevel` key anywhere in `describe()`'s return map.

Consequence: an agent builds `(A OR B) AND (C OR D)` via `set_condition` with an explicit `junctionLevel: 0` on the joining `and` clause (exactly this PR's own new test, `explicitJunctionLevelProducesTwoIndependentGroups`). It then calls `get_condition` to inspect or edit the condition — an ordinary, expected workflow; this file already has a named regression test guarding exactly this class of interaction for plain `level` (`aNonZeroLevelSurvivesAGetConditionThenSetConditionRoundTrip`, line ~579). The described output has all four conditions at `level: 1` and no `junctionLevel` anywhere. If that output is fed back into `set_condition` unchanged (the destructive-round-trip pattern that test file's own docstring calls out as the recorded failure mode this vocabulary has been bitten by before), `junction()`'s default path fires: `Math.min(1, 1) = 1` for the joining `and` — not the required `0`. This silently collapses back to the flat/misgroup evaluation this PR exists to fix, with no error, on the exact shape (two independent groups) that motivated Tier 1 over Tier 2 in the diagnosis.

This is the read-path sibling of the write-path this PR tested — `AssemblyConditionService.read()` (line 71) calls `describe()` directly, and `AssemblyConditionService.set()` (line 88) calls `toConditionList()` directly, so this is reachable through the real `get_condition`/`set_condition` MCP round trip, not just a theoretical API gap.

**Suggested fix:** have `describe()` also read the following junction's level (mirroring `junctionAfter()`'s existing lookup of the following `JunctionOperatorModel`) and include it as a `junctionLevel` key on the clause map, and add a regression test extending the existing round-trip test pattern to the two-groups shape (build via explicit `junctionLevel`, `describe()`, replay into `toConditionList()`, assert the junction level and/or truth-table evaluation still holds).

**No other findings ≥ 80 confidence.** I actively tried to break this on: other call sites (none beyond the one, confirmed by compile), boundary shapes (single-clause list, no second-to-last transition), the mixed-operator same-depth-run validation the diagnosis flagged as optional defense-in-depth (Tier 1 item 4, explicitly deferred by the diagnosis, not required for this PR) — no other defect found.

**CI status — blocking, not yet clear to merge:** as of this review round, the `build` (Validation Build) check is still `IN_PROGRESS`. `claude-review` shows `FAILURE` but completed in ~0s (a tooling hiccup per the dispatch note, not a real review output) — not treated as a blocker here, but flagging if it later produces real content. **This PR must not be merged until `build` completes successfully**, independent of the finding above.

fredaan:
Fixed in b0ed01f01. Thanks for catching this — confirmed the reachable read-modify-write gap.

`ConditionVocabulary.describe()` now also reads the following junction's own level (mirroring `junctionAfter()`'s existing lookup of the following `JunctionOperatorModel`, using `.getLevel()` instead of `.getType()`) and includes it as a `junctionLevel` key on each clause map — so `get_condition`'s output now carries the same information `set_condition` accepts.

Added `anExplicitJunctionLevelSurvivesAGetConditionThenSetConditionRoundTrip`, extending the existing `aNonZeroLevelSurvivesAGetConditionThenSetConditionRoundTrip` pattern to the two-groups shape: builds `(A OR B) AND (C OR D)` with an explicit `junctionLevel: 0` on the joining `and`, calls `describe()`, asserts the `junctionLevel` key is `0` (not absent), replays the described output straight back into `toConditionList()`, and asserts the joining junction's level is still `0` after the round trip (not flattened to the default-inferred `1`).

Test results:
- `./mvnw test -pl core -Dtest=ConditionVocabularyTest` — 53/53 pass (52 prior + 1 new)
- `./mvnw compile test-compile -pl core -am` — exit 0, no other call sites affected

fredaan:
## Review round 2 — PR #5337 (bug 76741 / VSC-008)

**Commit range reviewed:** `909db6032..b0ed01f01` (new commit `b0ed01f01`, "surface junctionLevel through get_condition's describe()")

**Status of round 1's finding:** Fixed. Re-read the actual diff (not the fixer's summary):

- `ConditionVocabulary.describe()` now calls a new `junctionLevelAfter(conditionList, i)` helper
  and puts it in the returned clause map as `"junctionLevel"`. Confirmed `junctionLevelAfter()`
  mirrors `junctionAfter()`'s existing lookup of the following `JunctionOperatorModel` but returns
  `junction.getLevel()` (an `Integer`) instead of deriving the `and`/`or` token — correct, minimal,
  and consistent with the rest of the file's style.
- Confirmed the new test `anExplicitJunctionLevelSurvivesAGetConditionThenSetConditionRoundTrip`
  actually exercises the round trip I flagged: builds `(A OR B) AND (C OR D)` with an explicit
  `junctionLevel: 0` on the joining clause, calls `describe()`, asserts the map at index 1 (clause
  `b`, whose junction is the joining "and") carries `junctionLevel: 0`, then rebuilds `Clause`
  objects entirely from the described map (including the new `junctionLevel` key) and re-runs
  `toConditionList()`, asserting the joining junction is still level `0` — not the `1` the default
  `Math.min(1,1)` formula would silently produce. Traced this by hand against the described-index
  math (`describe()` only emits one map entry per condition, so `described.get(1)` is clause `b` at
  conditionList index 2, and `junctionLevelAfter(conditionList, 2)` correctly reads
  `conditionList[3]`, the explicit-level-0 `and` junction) — indexing is correct. This test would
  fail on the pre-fix code (`junctionLevel` key wouldn't exist in the map, so the replayed clause's
  `junctionLevel` would be `null`, and `Math.min(1, 1) = 1` would fire instead of `0`) — confirmed
  by tracing the pre-fix `describe()` code path, which never wrote a `junctionLevel` key at all.

**Independently re-ran the tests myself** (not the fixer's report), per the note that `-am` breaks
on this specific `-Dtest` invocation against an unrelated upstream reactor module — ran `-pl core`
alone as instructed, matching every prior invocation in this task:

```
./mvnw test -pl core -Dtest=ConditionVocabularyTest -q
```
Surefire report: **53 tests run, 0 failures, 0 errors** (52 previous + 1 new).

Also reran the full compile (this one does tolerate `-am`):
```
./mvnw compile test-compile -pl core -am -q
```
**Exit 0** — no regression on the ~15 existing `Clause` call sites.

**No new findings ≥ 80 confidence.** The fix is minimal, scoped exactly to the reported gap, and
does not touch anything else.

**CI status — still blocking:** `build` (Validation Build) is **IN_PROGRESS** again against the new
commit as of this round. `claude-review` again shows `FAILURE` completing in ~0s — same tooling
hiccup pattern as round 1, not real review content. **Still not clear to merge until `build`
completes successfully.**

**Verdict: clean, pending CI.** No code-review blockers remain; approval is withheld only on the
`build` check, which is not yet green.

jshobe-inetsoft (approved):

jshobe-inetsoft (approved):

PR #5338, opened 2026-09-17
fix(76674): guard null runtime ids at the cluster-affinity funnel

ngovus-inetsoft:
> **Base:** this targets `stylebi-wiz-main-rebase`, not `main`. The prior fixes it builds on — #4971's proxy swap in `ModifyCalculateFieldService` and #5252's explicit-id `refreshViewsheet` overload — are not on `main` yet, so the change does not apply there.

Follow-up to #5252 (bug #76666). Closes out a defect class that has now been fixed one call site at a time four separate times.

## The shape

A caller takes a viewsheet runtime id from `RuntimeViewsheetRef` — a Spring STOMP-message-scoped bean, populated only from a native header on a real browser WebSocket session — and feeds it into a `@ClusterProxyKey`-routed cluster-affinity call. A caller with no live STOMP session gets `null`, that null reaches `RuntimeSheetCache.getAffinityKey`, and Ignite's `AffinityKey` constructor throws `NullPointerException("Ouch! Argument cannot be null: key")`. Ignite ships no sources jar, so that message is opaque until someone decompiles `ignite-core`.

Prior fixes: #4971 (PCB-006), #5186 (#76615), #5252 (#76666, which also folded in `ComposerViewsheetService.checkMV`).

Rather than a fifth site-by-site patch, this closes the pattern at both ends.

## 1. The funnel guard

`RuntimeSheetCache.getAffinityKey` now rejects a null id with a named exception instead of letting it reach Ignite.

`new AffinityKey` appears **exactly once** in the repository (community and enterprise), and every affinity path funnels through this method — `containsKey`/`get`/`put`/`putSheet`/`remove`/`clear`/`flush`/`isLocal`, plus `WorksheetEngine.affinityCall` and `affinityCallAsync`. So this covers all 240+ `@ClusterProxyKey` methods at once rather than one path, and replaces a message that costs a decompile to interpret.

**Not a behaviour change.** `getOriginalId` already null-guarded and returned null, so anything reaching the new check previously threw from `AffinityKey`'s constructor one line later. The `Map`-facing entry points all gate on `key instanceof String`, so `get(null)` / `containsKey(null)` / `remove(null)` never reach the guard and behave exactly as before. `isLocal(String)` still short-circuits on `cluster.isLocalCall()` before it.

## 2. The five remaining call sites

Each already held a live id and simply wasn't threading it through.

| Site | Fix |
|---|---|
| `VSChartRefreshService.refreshChart` | explicit-id `refreshViewsheet` overload |
| `TouchAssetService.processRefreshEvent` | explicit-id `refreshViewsheet` overload |
| `ChangeGeographicService` | controller hop → `VSBindingTreeControllerServiceProxy` |
| `ConvertChartRefService` | controller hop → `VSBindingTreeControllerServiceProxy` |
| `ConvertTableRefControllerService` | controller hop → `VSBindingTreeControllerServiceProxy` |

For the first two, `runtimeId` was already in lexical scope — the private helper just didn't take it.

For the last three, the durable fix is dropping the controller hop rather than adding an explicit-id overload to the controller. A service invoking a `@MessageMapping` controller as an ordinary Java method is the layering inversion that *generates* this defect class: the controller is entitled to read message-scoped state because it executes inside a message, and a service calling it directly smuggles that entitlement into a stack with none. This is what #4971 did for `ModifyCalculateFieldService`. `VSBindingTreeController` keeps its single STOMP-only method.

`ConvertChartRefService`'s `RuntimeViewsheetRef` injection turned out to be already dead — never read anywhere in the class — and comes out with the controller.

A repo-wide grep for injected `*Controller` fields outside controller classes confirms this list is complete for this shape. The three remaining hits (`AssemblyConditionDialogService`, `ExpressionDialogService`, `WorksheetScriptService`) inject controllers whose methods take explicit arguments and touch neither `RuntimeViewsheetRef` nor `@ClusterProxyKey` — a layering smell without this defect.

## 3. A loading mask this fix chain silently dropped

Found in review. The explicit-id `refreshViewsheet` overload lacked the `@LoadingMask(value = true, watchdogTimeout = 0)` that the STOMP-mapped 4-arg method carries, so **every site that moved between them lost its busy indicator** — #5252's `ModifyCalculateFieldService` and `ComposerViewsheetService.checkMV`, as well as `VSChartRefreshService` here. Annotating the overload repairs all three.

Safe against double-masking: the 4-arg reaches it by a plain `this` call, which bypasses the Spring AOP proxy, so the STOMP path still fires exactly once.

Two near-misses, recorded so they aren't re-derived:
- `TouchAssetService` was never affected — its event sets `autoRefresh(true)`, and `EventAspect.clearLoadingMask` returns early on that before applying a mask.
- `sendUserMessage` advice is **not** lost at these sites. Its pointcut is `@MessageMapping && within(inetsoft.web.viewsheet..*)`, which the outer controller still matches, so it wraps the nested call. `VSBindingTreeController` sits in `inetsoft.web.binding.controller`, outside that pointcut, and carries no `@LoadingMask` — which is why the three `getBinding` sites lose no advice at all.

## Testing

27 tests green across the affected classes, including the 18 pre-existing ones on these paths.

Every regression test asserts **both** that the correct id reached the routed call and that the session-deriving overload was *not* used. Asserting only that nothing was thrown passes against unfixed code too, because these call sites wrap the call in a `try`/`catch` that warns and continues — the trap that let this defect class survive its own tests repeatedly.

The `@LoadingMask` coverage is by reflection rather than behaviour, deliberately: the advice is applied by Spring AOP at the proxy boundary, so no unit test invoking the method can observe it. That property is exactly why three sites lost the mask with every test still green.

## Not in scope

Replacing proxy re-entry with a direct service call at sites already on the right node. It is entangled with converting `refreshViewsheetAsync` to synchronous, which also affects the native browser STOMP path and needs its own evaluation. Worth noting while evaluating it: nothing awaits that `Future` today, so a failure arising *inside* the async job is not caught by the caller's `try`/`catch` at all — only the synchronous `getAffinityKey` throw ever was.

## Needs human verification

The changed paths are all interactive and were verified by unit tests only, not in a running server. Worth exercising in a browser:
- an **embedded**-viewsheet chart refresh (the `hasEmbeddedAssembly()` branch), where the restored loading mask should now appear
- an idle viewsheet long enough for `touchAsset`'s auto-refresh
- binding editor: change a column's geographic role, convert a chart ref, convert a table ref — each must still repopulate the binding tree

A companion `claude/cluster.md` update in the enterprise repo is being held back for a separate PR.


jshobe-inetsoft (approved):
# Review: stylebi PR #5338 — "guard null runtime ids at the cluster-affinity funnel"

Repo: `inetsoft-technology/stylebi` (community)
Branch: `bug-76674` → base `stylebi-wiz-main-rebase` (not `main` — see note below)
Files changed: 14 (+862 / −26), 7 production files + 7 new/updated test files

## Verdict

**Ready to merge.** No blocking issues found. This is a well-scoped, well-tested fix that I independently verified against the actual base-branch source (not `main` — see note below), including tracing every modified call site, the funnel guard's interaction with all `Map`-facing entry points of `RuntimeSheetCache`, the `@LoadingMask`/Spring-AOP self-invocation reasoning, and an independent repo-wide grep for the "controller injected into a non-controller service" pattern this PR closes out — I found no additional unfixed site.

Non-blocking notes below (base-branch target, a failed automated check that is infra noise rather than a real finding, and two minor style observations) — none of them should hold up the merge.

---

## Verification method

Because this PR's base is `stylebi-wiz-main-rebase`, not `main`, I fetched that ref (`git fetch origin stylebi-wiz-main-rebase`) and read all "before" state from it rather than from the local `main` checkout — the local checkout is missing PR #4971 and #5252, which this PR builds on, so diffing against it would show spurious context. I also created a temporary worktree of that base ref to do a repo-wide grep for the defect pattern (removed afterward, no lasting changes to the repo).

## Correctness of the core fix

**1. The funnel guard (`RuntimeSheetCache.getAffinityKey`).** Confirmed by reading the full file at the base ref:
- Every one of the `Map`-interface entry points (`get`, `containsKey`, `remove`) that takes `Object key` gates on `key instanceof String id` before ever calling `getAffinityKey`, so `get(null)` etc. still short-circuit to "not found" exactly as before — the guard is unreachable from those paths, matching the PR's claim.
- The typed-`String` entry points (`put`, `putSheet`, `putAll`, `clear`, `flush`, `updateAccessTime`, `isLocal(String)`) would, prior to this change, pass a null straight into `new AffinityKey<>(id, getOriginalId(id))` and blow up inside Ignite's constructor with the opaque `"Ouch! Argument cannot be null: key"` message. After this change they hit the named `Objects.requireNonNull` one line earlier instead. Same failure, better message — genuinely not a behavior change, as claimed.
- `WorksheetEngine.affinityCall`/`affinityCallAsync` (lines 1266–1301) call `amap.getAffinityKey(rid)` directly, confirming the "covers 240+ `@ClusterProxyKey` methods" claim is architecturally accurate — this is in fact the single chokepoint.
- `getOriginalId` and the `tempIdPattern` it uses are `static`, so the tests mocking `RuntimeSheetCache` with `Mockito.mock(RuntimeSheetCache.class, CALLS_REAL_METHODS)` and calling the real `getAffinityKey` exercise real logic without needing to stand up a `Cluster`/Ignite cache — that test technique is sound.

**2. The five call-site fixes.** I read each production file at the base ref in full (not just the diff hunks) and confirmed in every case that the `@ClusterProxyKey`-scoped id the fix now threads through is the *same* id the enclosing method already used to look up the `RuntimeViewsheet`/state it's operating on — i.e., these aren't just "pass something non-null," they're passing the demonstrably correct id:
- `ChangeGeographicService.changeGeographic`, `ConvertChartRefService.convertChartRef`, `ConvertTableRefControllerService.convertTableRef` — all three call `viewsheetService.getViewsheet(id, principal)` earlier in the same method using the exact `id` now passed to `vsBindingTreeService.getBinding(id, ...)`. Dropping the `VSBindingTreeController` hop in favor of calling `VSBindingTreeControllerServiceProxy` directly is a real layering fix, not a workaround: `VSBindingTreeController.getBinding` (unchanged, still the sole STOMP endpoint) does nothing but call this same proxy with `runtimeViewsheetRef.getRuntimeId()` — so the fix is byte-for-byte equivalent to what the controller would have done, minus the null.
- `TouchAssetService.processRefreshEvent` — `rs = worksheetService.getSheet(runtimeId, principal)` is fetched by `runtimeId` at the top of `touchAsset`; the `oid`-walking loop only touches ancestor sheets, it never reassigns `rvs`, so passing the original `runtimeId` into the refresh is correct.
- `VSChartRefreshService.refreshChart` — `VSChartStateInfo.createChartState(event, viewsheetService, principal, runtimeId)` (confirmed via `VSChartControllerService`) fetches the `RuntimeViewsheet` by that same `runtimeId`, so `chartState.getRuntimeViewsheet()` and the `runtimeId` now passed to `refreshViewsheet` are guaranteed to agree.

**3. `ConvertChartRefService`'s dropped `RuntimeViewsheetRef` field.** Grepped the base-ref file: the field is assigned in the constructor and declared, but never read anywhere else in the class. Genuinely dead; safe to remove.

**4. The `@LoadingMask` restoration on `VSRefreshController`'s 5-arg overload.** Read the full base-ref file. The 4-arg STOMP-mapped method (already `@LoadingMask`-annotated) invokes the 5-arg overload via a bare `refreshViewsheet(this.runtimeViewsheetRef.getRuntimeId(), ...)` call — a same-class self-invocation. Spring AOP's default proxy-based interception does not intercept self-invocations, so adding `@LoadingMask` to the 5-arg overload does not cause the aspect to fire twice on that path — the PR's "no double-masking" claim is correct standard Spring-AOP behavior, not hand-waving.
   - Verified `TouchAssetService`'s refresh event sets `.autoRefresh(true)`, and `EventAspect.clearLoadingMask` (read in full) returns `pjp.proceed()` immediately when the `VSRefreshEvent` argument has `autoRefresh()==true`, before ever touching the mask machinery — so the new annotation is genuinely inert for that call site, exactly as the PR states.
   - Confirmed `ComposerViewsheetService` calls the 5-arg overload from three sites (including `checkMV`), and both `TouchAssetService`/`VSChartRefreshService` call it too — so this one annotation addition is the complete fix for all three previously-silently-broken loading-mask regressions the PR describes, without needing to touch those other files.

**5. Completeness of "no other sites left."** I independently grepped the full base-ref source tree for `@Service` classes with a constructor parameter typed `*Controller*` (not `*ControllerService*`/`*ControllerServiceProxy*`) — the shape this defect class takes. Beyond the five sites this PR fixes, I found `ChangeChartTypeService` (already using `VSBindingTreeControllerServiceProxy` directly with the correct explicit id — apparently already fixed/written correctly), `ModifyCalculateFieldService`/`ComposerViewsheetService` (already fixed by #4971/#5252, not touched by this PR), and the three "safe" sites the PR names (`AssemblyConditionDialogService`, `ExpressionDialogService`, `WorksheetScriptService`), plus two wiz-layer services (`ViewsheetEditService`, `LayoutMutationService`) that already inject `*ControllerServiceProxy`/plain-`Service` types rather than raw controllers. Spot-checking the three "safe" sites confirmed their controller-typed usages are either static utility calls, locally-`new`'d helper objects, or take an already-resolved `RuntimeWorksheet`/`RuntimeViewsheet` as an explicit argument — none touch `RuntimeViewsheetRef` or route through `@ClusterProxyKey`. I found no unfixed instance of this defect class.

## Tests

- Verified every new test's constructor-argument list against the actual (post-fix) production constructor signatures at the base ref — all match in type and order (`ChangeGeographicServiceTest`, `ConvertChartRefServiceTest`, `ConvertTableRefControllerServiceTest`, `TouchAssetServiceTest`, `VSChartRefreshServiceTest`).
- Each test's core assertion is a `verify(..., never())` that the *null-deriving* overload/path was **not** used, in addition to asserting the correct id was used — this is the right assertion shape for this defect class, since (per the PR body, and confirmed by reading the surrounding `try`/`catch` in these methods) a test that only checks "nothing threw" would pass against the unfixed code too.
- `RuntimeSheetCacheTest` pins the exact guard message and explicitly asserts the message does **not** contain `"Ouch!"` (i.e., that Ignite's own constructor was never reached) — a good regression trap against someone "fixing" this by reordering the guard later.
- CI's "Validation Build" check passed in ~23 minutes (a full build, not `-DskipTests`), with no test-failure annotations — consistent with the PR's claim of "27 tests green."

## Other CI checks

- `build`: SUCCESS. `Semgrep`/`Semgrep OSS`: SUCCESS. `license/cla`: SUCCESS. `mergeable`: MERGEABLE (no conflicts).
- `claude-review` (the automated Claude Code Review GitHub Action): **FAILURE**, but the run log shows `Claude execution failed: result is_error:true` / `Process completed with exit code 1` — i.e., the automated review tool itself errored out before producing a review, not a finding about this code. This is infra noise, not a legitimate blocking signal, and this manual review supersedes it per the project's stated purpose for `review-and-merge-prs`.

## Security

No new attack surface. The change only adds a null-check and re-routes internal calls to pass an id that was already validated/derived upstream (via `@ClusterProxyKey`, which is itself populated from the caller's session context before these methods run) — it does not introduce any new way for an untrusted caller to influence which id is used.

## Backward compatibility

All changed method/constructor signatures are on internal Spring-managed beans (`@Service`/`@Controller` classes wired purely by Spring DI within this module); nothing here is a public/serialized API. No external callers are plausible. No compatibility concern.

## Non-blocking observations

1. **Base branch.** The PR explicitly targets `stylebi-wiz-main-rebase` rather than `main`, because it depends on unmerged prior fixes (#4971, #5252) that live there. This is called out prominently in the PR description, so it's not a surprise to a reviewer, but whoever merges should double check that `stylebi-wiz-main-rebase` is the intended long-term integration branch for this line of work (and that this fix eventually lands on `main` once the base branches converge) — not something to block this PR on, just a heads-up for downstream tracking.
2. **Comment duplication.** The explanatory comment above the `vsBindingTreeService.getBinding(id, ...)` call is duplicated nearly verbatim across `ChangeGeographicService`, `ConvertChartRefService`, and `ConvertTableRefControllerService`. This is defensible — the comment aids per-call-site readability and this is exactly the kind of subtle regression class the PR is trying to prevent from recurring — but a shared doc comment on `VSBindingTreeControllerServiceProxy.getBinding` (or a short pointer to it) would reduce triplication if this pattern recurs again.
3. **Enterprise `claude/cluster.md` follow-up.** The PR description notes a companion doc update is being held for a separate PR in the enterprise repo. Worth confirming that follow-up actually happens (not verifiable from this repo).

PR #5322, opened 2026-09-17
feat(wiz-agent): add collect_parameters/set_parameters endpoints for viewsheet variables

fredaan:
## Summary
- Add `GET`/`POST /api/wiz/v1/agent/viewsheet/{sessionToken}/parameters` endpoints backing two new composer-chat MCP tools (`collect_parameters`/`set_parameters`), so an agent can discover and set variables a connected viewsheet's source query/tree references, even when there is no on-canvas input assembly bound to them.
- `ParameterDiscoveryService`: shared discovery logic wrapping `ParameterService.getPromptParameters`, plus exclusion of variables already bound to an on-canvas input assembly.
- `ParameterCollectionService` (read) / `ParameterValueService` (write, reusing the existing `VSCollectParametersServiceProxy.collectParameters` — the same production path StyleBI's own "Parameters" prompt dialog uses).
- `ParameterModel` record for the JSON response shape.

Base branch is `feature-admin-ai-community-migration` rather than `main`: the entire `inetsoft.web.wiz.viewsheet` agent-pairing surface this change extends currently lives only on that migration branch, not yet on `origin/main`.

This is the Java half of a two-repo feature; the paired `stylebi-wiz` composer-chat plugin PR is inetsoft-technology/stylebi-wiz#2622.

Design doc (in the `stylebi-wiz` repo): `docs/superpowers/plans/2026-09-17-viewsheet-collect-set-parameters.md` (Tasks 1-3 for this PR's scope).

## Test plan
- [x] `./mvnw test -pl core -Dtest=ParameterDiscoveryServiceTest` — 5/5 passing (pure mapping/validation logic; live discovery needs a real `RuntimeViewsheet`/`ViewsheetSandbox`, not constructible outside a Spring context)
- [x] `./mvnw compile -pl core -am` — BUILD SUCCESS
- [x] Independent code review (subagent): no issues at confidence ≥80, including tracing the write path through `AssetQuerySandbox.refreshVariableTable` to confirm `set_parameters` merges into the existing variable table rather than replacing it
- [ ] Live verification against a running StyleBI instance (deferred until this PR and the paired `stylebi-wiz` PR both merge)

PR #5339, opened 2026-09-17
Bug #76739, add set_viewsheet_data_source/convert_data_source_to_worksheet backend support

jkidacki-inetsoft:
## Summary

Redmine #76739 found three Composer Viewsheet Options dialog features with zero coverage in the wiz plugin's property tooling: Data Source select/clear, the Prompt for Parameters "Customize" list, and "Convert Source to Worksheet". This PR adds the backend support for all three.

## Root Cause

- `enabledParameters`/`disabledParameters` (a plain `String[]` pair on `ViewsheetParametersDialogModel`, genuinely applied by `ViewsheetSettingsService.setViewsheetParameterInfo`) had no short-name alias in `PropertyAliases`.
- `selectDataSourceDialogModel.dataSource` needs a resolved, permission-checked `AssetEntry` rather than a scalar JSON leaf, so it could not be a plain alias — nothing on the agent surface could replace or clear an already-set base (`attach_base_worksheet` deliberately refuses to touch one).
- The "Convert Source to Worksheet" action goes through a completely different service method (`ViewsheetPropertyDialogService.convertLogicModelToWorksheet`) that nothing on the wiz surface called.

## Fix

- Added `enabledParameters`/`disabledParameters` aliases to `PropertyAliases.viewsheet()`.
- Added `SheetPropertyService.setDataSource` + a new `set-data-source` endpoint on `ViewsheetAssemblyAgentController`, reusing the existing `resolveDataSourceEntry` helper (shared with `attach_base_worksheet`/`create_viewsheet`) for asset resolution, and the existing `set_viewsheet_properties` apply path (`dialogService.setViewsheetInfo`) for the actual write — so dependency-refresh/sandbox-reset side effects run exactly as they would from a human editing the dialog. Unlike `attach_base_worksheet`, this always replaces an existing base (or clears it) rather than refusing.
- Added `SheetPropertyService.convertDataSourceToWorksheet` + a new `convert-data-source-to-worksheet` endpoint, calling `ViewsheetPropertyDialogService.convertLogicModelToWorksheet` directly (matching how `getViewsheetInfo`/`setViewsheetInfo` are already called in this class).

Three additional defects were found and fixed via live testing against a from-source server:

1. **Silent parameter drop**: setting `enabledParameters`/`disabledParameters` to a name the viewsheet's query doesn't actually declare reported `ok:true` and then silently vanished on the next read (the write is unconditional; the read re-derives the arrays by filtering against the live query variables). Now refused with a clear message naming the invalid entry and the known parameters.
2. **Opaque 500 on refusal**: `convertLogicModelToWorksheet`'s "not a logical model" refusal (and any checked exception from `setDataSource`'s underlying `Viewsheet.update` call) is a plain `Exception`, a type neither this controller's local `@ExceptionHandler(PairingException.class)` nor the global `WizControllerErrorHandler` recognizes — it reached the client as an opaque, unhelpful 500. Both new endpoints now rewrap into `PairingException`, matching how `attach_base_worksheet`/`create_viewsheet` already handle `resolveDataSourceEntry` failures.
3. **Stale Toolbox pane**: after a successful data-source change, the Composer's Toolbox/Data panel tree did not refresh — the identical gap `attach_base_worksheet` already hit and fixed (bug #76637). `setDataSource` now pushes the same `broadcastBindingTreeRefresh`/`broadcastViewsheetInfoRefresh` calls after a successful write.

## Test Plan

- `PropertyAliasesTest`, `SheetPropertyServiceTest`, `ViewsheetAssemblyAgentControllerTest`: 171/171 pass (`./mvnw -pl core test -Dtest=PropertyAliasesTest,SheetPropertyServiceTest,ViewsheetAssemblyAgentControllerTest`).
- `./mvnw -pl core install -am -DskipTests` builds clean.
- Live-verified against a from-source server: set/replace/clear a viewsheet's data source, confirmed the Toolbox pane refreshes without a manual reload, confirmed `convert_data_source_to_worksheet`'s refusal surfaces cleanly for a non-logical-model source, confirmed an unrecognized `enabledParameters` name is refused rather than silently dropped.

Plugin PR (adds the corresponding MCP tools): https://github.com/inetsoft-technology/stylebi-wiz/pull/2645

Fixes Redmine #76739.


jkidacki-inetsoft:
Follow-up fix pushed, found via further live testing with a real worksheet that has query parameters (the earlier live test used an unparameterized worksheet, which only exercised the refusal path):

**Bug**: patching only one of `enabledParameters`/`disabledParameters` left the moved name listed in **both** arrays on the next read, since the untouched field still carried it from the pre-patch state. The Composer's own "Customize" dialog can never produce this — it always submits both columns as one already-partitioned pair; the underlying `ViewsheetSettingsService.setViewsheetParameterInfo` has no mutual-exclusivity guarantee of its own.

**Fix**: `SheetPropertyService.set()` now prunes the overlap from whichever side the patch didn't touch after the ordinary per-key writes apply, and refuses outright if the caller explicitly puts the same name in both lists in one patch (a genuine self-contradiction, not an ordering question).

5 new tests added (176/176 total pass). Live-verified: disabling a real parameter via a single-field patch no longer leaves it in `enabledParameters` too, and an explicit `{enabledParameters:["X"], disabledParameters:["X"]}` patch is now refused.

jshobe-inetsoft (changes requested):
# Review: stylebi PR #5339 — Bug #76739, set_viewsheet_data_source / convert_data_source_to_worksheet / enabledParameters-disabledParameters aliases

Base branch: `stylebi-wiz-main-rebase` (a fast-moving, periodically-rebased wiz-agent integration branch — not `main`). Head: `7bd372ac6` (2 commits on top of `ea5123db8`).

## Verdict

**REQUEST_CHANGES** — blocking on the failing `build` check only. The reviewed logic itself is sound, well-tested, and consistent with the surrounding codebase; I found no correctness, security, or regression defects in the new code.

Blocking issues:

1. **`build` check fails to compile, and this PR cannot merge until it is fixed.** Root cause (confirmed from the raw Maven log, not just the annotation summary): the two new `ViewsheetAssemblyAgentControllerTest.controllerWith(...)` helper overloads added by this PR (`core/src/test/java/inetsoft/web/wiz/viewsheet/ViewsheetAssemblyAgentControllerTest.java:998` and `:934`, i.e. the 2-arg and 3-arg overloads) construct `new ViewsheetAssemblyAgentController(...)` with 31 positional arguments. At CI build time, the actual constructor on the base branch required 33 arguments — `ParameterCollectionService` and `ParameterValueService` (inserted after `InputValueService`) plus a trailing `ViewsheetPropertyDialogService`, all added by an **already-merged, unrelated PR** (`dcabca8c4`, "feat(wiz-agent): add collect_parameters/set_parameters endpoints for viewsheet variables", #5322) that landed on `stylebi-wiz-main-rebase` *after* this PR's branch point (`ea5123db8` is not an ancestor of the current base tip). The base has moved further still since (it now also carries `HierarchyDimensionService`/`FormTableRowService` params). This is a **stale-branch / semantic-merge-conflict problem, not a defect in this PR's own added code** — none of the new `SheetPropertyService`/`PropertyAliases`/`ViewsheetAssemblyAgentController` production logic in this diff is implicated. Confirmed by the exact javac diagnostic:
   ```
   ViewsheetAssemblyAgentControllerTest.java:[1030,14] / [1069,14]
   constructor ViewsheetAssemblyAgentController ... cannot be applied to given types;
   reason: actual and formal argument lists differ in length
   ```
   **Action needed before merge:** rebase `bug-76739` onto the current `stylebi-wiz-main-rebase` tip and update both `controllerWith` overloads (and verify no other call site) to supply mocks for the now-required constructor parameters. Given how quickly this base branch accrues new constructor dependencies, this will very likely need to be redone again at merge time if more wiz-agent PRs land first — that is an operational risk on this integration branch generally, not something specific to fix in this diff beyond the rebase itself.

No other blocking issues found.

Non-blocking suggestions:

2. **Double session resolution in the new `set-data-source` endpoint.** `ViewsheetAssemblyAgentController.setDataSource` calls `sessions.resolve(sessionToken, user)` once at the top (to get `rvs` for the post-write broadcasts), and then `SheetPropertyService.setDataSource` independently calls `sessions.mutate(sessionToken, user, ...)`, which internally re-resolves the same session/runtime via `requireSession`/`getSheetForPairing`. Functionally harmless (both resolve to the same runtime, and a session that disappears between the two calls will simply surface as a rewrapped `PairingException`), but it's two lookups where `attachBaseWorksheet` needs only one. Not worth blocking on, but worth a look if this pattern is copy-pasted for future endpoints.

3. **`requireKnownParameterNames`'s "known" set is `getViewsheetParameterInfo`-dependent on the sandbox being initialized.** I traced `ViewsheetSettingsService.getViewsheetParameterInfo` (`core/src/main/java/inetsoft/web/composer/vs/dialog/ViewsheetSettingsService.java:42`): it only calls `VSEventUtil.refreshParameters(...)` when `rvs.getViewsheetSandbox()` is present; if the sandbox hasn't been initialized yet, `varNames` is empty and the "known" parameter set the new validation checks against would be empty even for a query that genuinely declares variables — producing a false "declares no variables at all" refusal for what would otherwise be a legitimate `set_viewsheet_data_source` → immediately-following `enabledParameters` customization sequence. This is a pre-existing characteristic of `getViewsheetParameterInfo` (not introduced by this PR), and it mirrors what the human dialog itself would show in the same state, so it's not a regression — just a sharp edge worth being aware of if agent callers hit spurious refusals right after attaching a new data source.

## Findings by review dimension

### Completeness and correctness

- I verified every factual claim in the PR description against the actual source, not just against the PR's own comments:
  - `PropertyAliases.viewsheet()` (`core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyAliases.java`): `enabledParameters`/`disabledParameters` aliases correctly map to `vsOptionsPane.viewsheetParametersDialogModel.enabledParameters`/`disabledParameters`. Correct — `ViewsheetSettingsService.setViewsheetParameterInfo` (line 99) genuinely writes `disabledVariable`/`orderedVariables` from these two fields unconditionally.
  - The "silent parameter drop" defect is real: `getViewsheetParameterInfo` (same file, line 42) filters both arrays against `varNames` derived from `VSEventUtil.refreshParameters(...)` — a name the query doesn't declare is silently dropped on the next read. `requireKnownParameterNames` in `SheetPropertyService` correctly guards this by validating the raw patch value against the pre-patch (correctly-filtered) `enabledParameters`/`disabledParameters` of the freshly-read model, before any write is applied.
  - `reconcileParameterLists` correctly prunes the untouched side's overlap when only one of the two arrays is patched, and correctly refuses (without writing) when both are patched and still overlap. All 3 code paths (touch-one-side, touch-both-no-overlap, touch-both-with-overlap) are exercised by tests and match the described semantics.
  - `SheetPropertyService.setDataSource` sets `model.vsOptionsPane().getSelectDataSourceDialogModel().setDataSource(entry)` and delegates to the shared `dialogService.setViewsheetInfo(...)` apply path. I read `ViewsheetPropertyDialogService.setViewsheetInfo` (`core/src/main/java/inetsoft/web/composer/vs/dialog/ViewsheetPropertyDialogService.java:263`) directly: it correctly handles both the non-null branch (`viewsheet.setBaseEntry(datasource)`, `viewsheet.update(...)`, `updateBoundAssemblies`, sandbox `resetRuntime()`, `VSDependencyChangedCommand`) and the null/"clear" branch (`clearOldBinding`, `setBaseEntry(null)`, sandbox `resetRuntime()`). The PR's claim that this reuses "exactly" the dialog's own apply path, including dependency-refresh/sandbox-reset side effects, is accurate.
  - Unlike `attachBaseWorksheet` (which refuses when `vs.getBaseEntry() != null`), `setDataSource` never checks the existing base and always replaces/clears it — matches the PR's stated intent and is exercised by `setDataSourceReplacesAnAlreadyBasedViewsheet`.
  - `convertDataSourceToWorksheet` calls `dialogService.convertLogicModelToWorksheet(rvs.getID(), user)` directly and returns `path`/`hasMaterializedViews` — a thin, faithful wrapper; it does not rebind the viewsheet, which is correctly documented as intentional (matching the dialog's own two-step UX).
  - The `PairingException` rewrapping in both new controller endpoints (and in `setDataSourceOrRewrap`) is justified: `convertLogicModelToWorksheet` and `Viewsheet.update(...)` (invoked transitively from `setViewsheetInfo`'s data-source branch) are declared to throw plain `Exception`, which neither the controller's local `@ExceptionHandler(PairingException.class)` nor `WizControllerErrorHandler` would otherwise turn into a clean 4xx — confirmed by reading the controller's existing `@ExceptionHandler` and comparing to the identical, pre-existing rewrap pattern already used by `attachBaseWorksheet`/`createViewsheet`.
  - The Toolbox/Data-panel refresh gap and its fix (`broadcastDataSourceChanged` calling `broadcastBindingTreeRefresh`/`broadcastViewsheetInfoRefresh`) mirrors the already-shipped fix for the identical gap in `attachBaseWorksheet` (bug #76637); reusing that same pair of calls is the right, minimal fix and is exercised for both the clear and set branches, and confirmed *not* to fire when the request is refused before any write (`setDataSourceDoesNotBroadcastWhenTheRequestIsRefused`).
- I did not find any half-implemented piece of the three stated features (Data Source select/clear, parameter Customize list, Convert Source to Worksheet) — all three have both production code and dedicated tests.

### Clean code / best practices

- The change follows the file's existing conventions closely: same request-validation-before-mutation ordering as `attachBaseWorksheet`, same `PairingException` rewrap idiom, same `resolveDataSourceEntry` reuse, same broadcast reuse. This is not a from-scratch pattern; it is a disciplined extension of the existing one.
- Javadoc is unusually dense and includes concrete "confirmed live" evidence with dates for each defect fixed — good practice for auditability, if verbose. Not a code-quality concern.
- Minor: `SheetPropertyService.namesOf` is a small, well-scoped private helper that cleanly unifies `String[]` (post-read model) and `Collection<?>` (raw JSON patch) inputs, with a clear error for anything else. No issues.

### Error handling and edge cases

- Field-combination validation (`clear` combined with any source field, or no fields at all) is checked before any repository/service call, matching the existing `resolveDataSourceEntry` philosophy of "cheap validation before expensive/fallible lookups" (there's even an explicit prior regression test elsewhere in this file guarding that exact ordering for `physicalTable`).
- `namesOf` rejects non-array/non-collection JSON leaves for the parameter lists with a clear message, rather than silently misinterpreting a bare string.
- Self-contradictory patches (same name in both `enabledParameters` and `disabledParameters` in one request) are explicitly refused rather than resolved arbitrarily — correct, since silently picking a side would be surprising.
- See non-blocking note #3 above regarding the sandbox-not-yet-initialized edge case in the new parameter-name validation.

### Readability and maintainability

- Good. Method names (`reconcileParameterLists`, `requireKnownParameterNames`, `namesOf`, `broadcastDataSourceChanged`, `setDataSourceOrRewrap`) are self-describing and each carries a javadoc explaining *why*, not just *what*, including references to the specific live-testing session that surfaced each defect. This matches the file's established style.

### Security

- No new attack surface: `set_viewsheet_data_source` reuses the same `resolveDataSourceEntry` permission-checked resolution (`ResourceAction.READ` / `getSheet(..., true, ...)`) already used by `attach_base_worksheet`/`create_viewsheet`; no new unauthenticated or unauthorized path is introduced. `requireEnabled()` (the feature-flag gate) is called at the top of both new endpoints, consistent with every other endpoint in this controller.
- Exception messages passed back to the caller include `e.getMessage()` from underlying service exceptions, which could in principle leak internal detail — but this is an existing, already-established pattern in this file (`attachBaseWorksheet`, `createViewsheet`), not something newly introduced by this PR.

### Regressions / backward incompatibility

- The two new aliases (`enabledParameters`/`disabledParameters`) and the two new endpoints are purely additive; no existing alias, endpoint, or method signature is removed or changed in an incompatible way for external callers.
- The one intentional behavior change — an `enabledParameters`/`disabledParameters` write naming an unknown parameter now fails loudly instead of silently succeeding-then-vanishing — is a deliberate bug fix (the aliases themselves are new in this PR, so there is no prior working caller of this exact path via aliases; and I found no pre-existing test elsewhere in the suite exercising the raw dotted-path escape hatch for these two fields with an out-of-set name, so no existing passing test is broken by this behavior change).
- No changes to any other alias, dialog model, or shared service method signature that could ripple into unrelated callers.

### Tests

- New tests are thorough and specific: `PropertyAliasesTest` gets both a positive alias-resolution test and a deliberate negative-space test proving the data-source field is *not* aliased. `SheetPropertyServiceTest` and `ViewsheetAssemblyAgentControllerTest` cover the mutual-exclusivity reconciliation (both directions, overlap-only pruning, explicit-conflict refusal, no-conflict pass-through), the unknown-parameter refusal (both `enabledParameters` and `disabledParameters` paths, plus the "no parameters at all" message variant), the clear/set broadcast behavior (including the "no broadcast on refusal" case), field-combination refusals, and the plain-`Exception`-to-`PairingException` rewrap for both new endpoints.
- I found no existing test elsewhere in the module whose expectations this diff invalidates (searched for other test usages of `enabledParameters`/`disabledParameters` under `viewsheetParametersDialogModel` — none found).
- The only test-related problem is the build-breaking constructor-arity mismatch described in blocking issue #1, which is a base-branch integration problem rather than a flaw in the tests' own logic — the tests were written correctly for the constructor signature that existed at this PR's branch point.

lizhepu:
Rebased `bug-76739` onto the current `stylebi-wiz-main-rebase` tip and fixed the build-blocking constructor-arity mismatch flagged in the review above, per your "Action needed before merge" note.

The base branch had moved further than expected — the `ViewsheetAssemblyAgentController` constructor now takes **37 params**, not 33 (picked up `HierarchyDimensionService`/`FormTableRowService`/`ColumnOptionService` in addition to `ParameterCollectionService`/`ParameterValueService`/`ViewsheetPropertyDialogService`). Rebase itself was clean, no conflicts.

Added one small commit (`2de13a2b8`, on top of the original two commits — nothing squashed/rewritten) updating just the two `controllerWith(...)` test helper overloads in `ViewsheetAssemblyAgentControllerTest.java` to supply mocks for the newly-required params, following the exact pattern already used elsewhere in the same file. No production code touched. Diff: +14/-2, one file.

Verified: clean build (`mvnw.cmd -pl core -am ... install`) succeeds, and this PR's own 3 test classes (`PropertyAliasesTest`, `SheetPropertyServiceTest`, `ViewsheetAssemblyAgentControllerTest`) all pass — 200/200.

Done as part of debug-workflow's VSD-005 investigation (Redmine #76853) — the plugin tool this PR's backend supports (`set_viewsheet_data_source`) was found failing with an opaque 404 because this PR never merged. Left unmerged for you/jkidacki to take it from here.

PR #5340, opened 2026-09-17
Bug #76770: add a working write path for chart target lines

ngovus-inetsoft:
Redmine #76770 (VCG-007).

## The issue

A chart's fixed-value target (goal) lines are **entirely unreachable** through the agent tool surface — not degraded, not partial, absent.

`chartAdvancedPaneModel.chartTargetLinesPaneModel.chartTargets` is a `TargetInfo[]`. `PropertyPath.coerce` builds arrays of primitives, `String` and enums and nothing else, so any JSON object array is refused (`"expects TargetInfo, which '{…}' is not."`), and nothing under that pane is aliased, so the raw dotted path — the only other way in — hits the same wall. The refusal is clean (no partial write); there is simply nothing behind it.

"Add a goal line to a chart" is an ordinary business ask, and the Composer has a full Targets tab for it. The static audit already recorded this as open/undecided: `docs/parity/L7-assembly-viewsheet-properties-hyperlinks/L7-parity-report.md:757` in `stylebi-wiz`.

## Why a dedicated service rather than teaching `coerce` about beans

Generic bean-array coercion would trade a loud failure for a silent one. Every corner of a hand-written `TargetInfo` produces a target that saves, lists back, survives a reload — and draws nothing:

| Trap | Result |
|---|---|
| `updateTargetCommonInfo` dereferences `measure` + three `ColorInfo`s unguarded | NPE inside the commit |
| `updateTarget` dispatches on `tabFlag` | anything outside 0/1/2 is a total no-op |
| `addTarget` fires only when `index == -1` | `index >= 0` on an empty chart is built and discarded |
| a non-numeric constant | strategy with zero boundaries — no geometry |
| `Average`/`Min`/`Max`/`Median`/`Sum` | read as *formulas* — a data-derived line reported as fixed |
| `supportsTarget` never consulted on write | a target on a pie is stored and never drawn |

So the agent-facing surface is scalars and every bean is built server-side, from the pane's own `newTargetInfo` template.

## What this adds

- **`ChartTargetLineService`** — `list` / `add` / `remove`, driving the same `ChartPropertyDialogService` the Composer does, one `sessions.mutate` per call (one undo checkpoint). Each trap above is a named refusal.
- **Three routes** on `ViewsheetAssemblyAgentController` under `…/chart/target-lines`.
- **`PropertyAliases`** — the target-lines pane is refused for write with a message naming the failure modes and the tools that do work, instead of leaving the generic coercion error as the last word.

Scope is deliberately one kind of target: a fixed-value **line**. Band and statistics targets are listed and removable (an agent needs their indexes) but not creatable.

## Adjacent bug fixed here

`AssemblyPropertyService.set` now clears `changed` on existing targets before writing. Every chart property patch — a title, a tooltip, `showValues` — reads the whole dialog model and writes it back, and `updateAllTargets` re-derives every target whose flag is set. `getTargetInfo` never clears it and its default is `true`, so the round trip **collapses a statistics target's several labels into one** (they are comma-joined on read and re-escaped as one string on write) and re-applies a band's default fill as an explicit user colour. `updateAllTargets` skips unchanged entries outright, so this is exactly a no-op for the write the caller asked for.

The Composer's own pane has the identical hole (`chart-target-lines-pane.component.ts` sets `changed = true` on add and edit and never `false`) — left for a separate fix.

## Notes for the reviewer

- `remove` addresses the descriptor **positionally**, because `removeDeletedTargets` does (`cDescp.getTarget(n)` is `targets.get(n)`). The `TargetInfo.index` field usually agrees but not always — `SyncChartHandler` clones only the still-bound targets and keeps their original index, leaving gaps — and sending the index field there deletes the wrong target while reporting success.
- `remove`'s post-write count check returns early when the dispatcher captured errors. A refused write does not throw inside the lambda (`editObjectProperty` dispatches an ERROR and returns `false`; `ViewsheetSessionService` converts it afterwards), so checking the count first would report a bogus partial apply for a concurrent-edit conflict.
- The measure is always copied out of `availableFields`, never rebuilt from its name: `isDateField`/`isTimeField` decide whether the constant is parsed as a number or a date.

## Testing

`./mvnw test -pl core -Dtest='inetsoft.web.wiz.viewsheet.**'` — **1058/1058 pass.**

- `ChartTargetLineServiceTest` (33) — the write shape (`index == -1`, `changed == true`, `tabFlag == 0`, everything non-null), every refusal, positional removal with gapped index fields, one-model-one-mutation.
- `ChartPropertyServiceTargetRoundTripTest` (11) — characterizes the **existing** `ChartPropertyService` behaviour the above depends on: the statistics-label collapse, the silent drop, the unguarded dereferences, the `tabFlag` no-op. Written against the old code deliberately, so the `changed` bookkeeping is not later "simplified" away.
- `PropertyAliasesTest` / `ViewsheetAssemblyAgentControllerTest` extended for the refusal and the routes.

Not yet done: the live end-to-end run through the plugin's own tools against a running instance.

## Companion PR

The MCP tool layer (`list_chart_target_lines` / `add_chart_target_line` / `remove_chart_target_line`) is a separate PR in `stylebi-wiz`; it depends on the routes added here.


jshobe-inetsoft (changes requested):
> **Changes requested.** The review below found no blocking issues in the code, but this PR cannot be merged as it stands: the branch has merge conflicts with its base. Once that is resolved a later review pass will pick it up.

---

# Review: stylebi PR #5340 — "Bug #76770: add a working write path for chart target lines"

Base branch: `stylebi-wiz-main-rebase` (the in-progress "wiz" AI-agent tool-surface feature line).
Files changed: 8 (4 main + 4 test), +1721/-0.

## Verdict

**Ready to merge.** No blocking issues found. I traced every factual claim in the PR
description (the `ChartPropertyService` traps, the `changed`-flag round-trip loss, the
positional-vs-index-field removal semantics, `PropertyPath.coerce`'s array-of-beans gap) against
the actual source in `core/src/main/java/inetsoft/web/viewsheet/service/ChartPropertyService.java`,
`ChartDescriptor.java`, `GraphTarget.java`, `TargetInfo.java`, `ChartTargetLinesPaneModel.java`,
`PropertyAliases.java`, `PropertyPath.java`, and `ChartPropertyDialogService.java`, and all of
them check out exactly as described. The new code's method signatures, constructor wiring, and
functional-interface usage (`ViewsheetSessionService.Mutation`, `CapturingCommandDispatcher`)
match the real APIs. Test coverage is thorough (33 + 11 + extensions to two existing suites) and
exercises the specific traps the PR documents. A handful of non-blocking suggestions are listed
below.

---

## What the PR does

- New `ChartTargetLineService` (`list`/`add`/`remove`) backing three new routes on
  `ViewsheetAssemblyAgentController` (`GET`/`POST /api/wiz/v1/agent/viewsheet/{token}/chart/target-lines`,
  `POST .../target-lines/delete`), giving the AI-agent tool surface a working path to chart
  target (goal) lines — previously entirely unreachable because `chartTargets` is a
  `TargetInfo[]` and `PropertyPath.coerce` only builds arrays of primitives/String/enum.
- `PropertyAliases.chartWriteRefusal` now explicitly refuses any write to
  `chartTargetLinesPaneModel` (under any spelling) with a message pointing at the new tools,
  instead of falling through to the generic `coerce` failure.
- `AssemblyPropertyService.set` now clears `TargetInfo.changed` on a chart's *existing* targets
  before every property write, fixing a pre-existing data-loss bug: any unrelated chart-property
  patch (title, tooltip, `showValues`, …) round-trips the whole `ChartPropertyDialogModel`, and
  `ChartPropertyService.updateAllTargets` re-derives every target whose `changed` flag is (by
  default) `true` — which silently collapses a statistics target's multiple labels into one and
  re-applies a band's default fill as an explicit user color.

## Verification performed

- Read the full diff (`gh pr diff 5340`).
- Read `ChartPropertyService.java` in full for the methods the PR's Javadoc makes claims about:
  `getTargetInfo`, `updateAllTargets`, `updateTarget`/`updateTargetCommonInfo`,
  `updateLineTarget`/`updateBandTarget`/`updateStatTarget`, `updateStatLabel`, `isFormulaSupported`,
  `removeDeletedTargets`, `getMeasures`, `supportsTarget`, `NO_TARGET_STYLES`. Confirmed each trap
  cited in the class Javadoc (unguarded `measure`/color dereferences, `tabFlag` dispatch doing
  nothing for unknown values, `addTarget` only firing on `index == -1`, formula-name shadowing,
  `supportsTarget` not being consulted on write) is real.
- Read `ChartDescriptor.addTarget/removeTarget/getTarget` and `GraphTarget.equals`/`hashCode` to
  confirm the "removal is equals-based, addressed here by position" claim, and confirmed
  `removeDeletedTargets`'s sort-descending-then-remove-by-position algorithm matches the service's
  positional (not `TargetInfo.index`-field) removal.
- Confirmed `ChartPropertyDialogService.setChartPropertyModel` calls `updateAllTargets` before
  `removeDeletedTargets`, matching the "never both in one call" comment (the service always sends
  either an all-unchanged-plus-one-new array with an empty `deletedIndexList`, or an
  all-unchanged-survivors array with a non-empty `deletedIndexList`, never both a new target and a
  deletion in the same call).
- Confirmed `PropertyPath.coerce`'s array branch recurses on the component type and has no
  bean-array case, so a `TargetInfo[]` genuinely bottoms out at "expects TargetInfo, which '{…}' is
  not." — the exact message quoted in the PR description.
- Confirmed `TargetInfo`, `MeasureInfo`, `ColorInfo`, `ChartTargetLinesPaneModel` field/getter/setter
  shapes match every access in the new service (no phantom methods).
- Confirmed `AssemblyPropertyService.typeOf` normalizes chart's simple class name to `"chart"`
  (lowercase), matching the `"chart".equals(type)` guard added around `markTargetsUnchanged`.
- Confirmed `PropertyAliases.writeRefusal`'s call chain (`viewsheetWriteRefusal` →
  `hierarchyDimensionsWriteRefusal` → `chartWriteRefusal` → `deadFieldWriteRefusal`) and that
  `resolveForWrite` checks both the raw input key and the resolved path, so the new
  `TARGET_LINES_PANE` substring check in `chartWriteRefusal` catches the field under an alias, a
  raw full path, or the raw path's own short name (`chartTargetLinesPaneModel.chartTargets`), and
  does not falsely catch sibling panes (`chartPlotOptionsPaneModel`, `chartLinePaneModel`) —
  matches the two new `PropertyAliasesTest` cases.
- Confirmed the `Mutation`/`CapturingCommandDispatcher` types and `ChartPropertyDialogService`
  method signatures (`getChartPropertyDialogModel(runtimeId, objectId, principal)`,
  `setChartPropertyModel(runtimeId, objectId, value, linkUri, principal, dispatcher)`) used by the
  new service match the real declarations exactly (arg count, arg order, types).
- Attempted a local Maven compile of the PR branch (`core` module) to double-check for
  compile-time issues beyond manual signature-matching; it failed on an unrelated, pre-existing
  `build-tools` plugin-descriptor problem in this sandboxed checkout, not on anything touched by
  this PR, so I fell back to the manual cross-check above (which is comprehensive: every method,
  field, and constant this PR references was located and read in the real source).
- Read `claude/ai-assistant.md` (the closest existing feature doc) and confirmed it documents a
  *different* system (the chat-panel LLM integration), not the `ViewsheetAssemblyAgentController`
  "wiz" agent-tool surface this PR extends. Grepped all `claude/*.md` for `wiz`/`AgentController`;
  none document this REST surface in depth (see non-blocking note below).
- Read `claude/chart.md` for chart-engine background; it has no target-line coverage to
  contradict or corroborate, no conflicts found.

## Detailed findings

### Correctness

- The `add()`/`remove()` write shape (`index == -1`, `changed == true`, `tabFlag == LINE_TARGET`
  for the new entry; `changed == false` for every survivor) is exactly what `updateAllTargets`
  requires to append/no-op correctly, and is asserted directly in
  `ChartTargetLineServiceTest.appendsOneChangedLineTargetAndLeavesTheRestUnchanged`.
- `remove()`'s positional addressing (`deletedIndexList` built from loop position `i`, not
  `TargetInfo.getIndex()`) is correct given `ChartTargetLinesPaneModel.chartTargets` is always
  built by walking `ChartDescriptor`'s targets in order (`ChartPropertyDialogService.java:249-250`,
  `ChartPropertyService.getTargetInfoList`), so position `i` in the pane's array and position `i`
  in `cDescp` agree at read time, and nothing mutates `cDescp` between the read and the
  `removeDeletedTargets` call inside the same `setChartPropertyModel` invocation.
- `resolveMeasure`'s decision to copy an existing `MeasureInfo` out of `availableFields` rather
  than construct one from the name is correct and necessary: `updateTargetCommonInfo` sets
  `target.setDateField/setTimeField` from `targetInfo.getMeasure()`'s flags, and those flags
  control whether `updateLineTarget`'s constant is later parsed as a date vs. a number
  (`TargetParameterWrapper.setDateField/setTimeField`, not read in this PR's diff but present in
  `ChartPropertyService.updateLineTarget`/`updateBandTarget`).
- `requireFixedValue`'s two refusal cases (formula-name literal match, then numeric-parse +
  `Double.isFinite` check) correctly pre-empt both silent failure modes described
  (`isFormulaSupported` shadowing, and `TargetStrategyWrapper` producing a boundary-less strategy
  on unparseable input) — including the `"1e400"` → `Infinity` edge case, caught by
  `Double.isFinite`.
- `describeChartTypes`/`typeName`'s `NO_TARGET_STYLES` int→name switch has exactly the 17 cases
  needed to name every entry in `ChartPropertyService.NO_TARGET_STYLES` (verified 1:1 by reading
  both lists side by side) — no silent "chart type N" fallback will fire for a chart actually
  reachable through this refusal path.
- `AssemblyPropertyService.markTargetsUnchanged` is inserted after all patch application and
  before `writeModel`, so it always operates on the pane exactly as freshly read (the target-lines
  pane itself can no longer be part of `resolved` given the new `PropertyAliases` refusal, so
  there's no ordering conflict with a patch that also tries to touch targets).

### Completeness

- Scope is intentionally narrow (fixed-value line targets only; bands/statistics are
  list/remove-only) and this is clearly documented in both the class Javadoc and the PR
  description, with a stated rationale (their own required shapes carry the same class of
  hand-built-bean trap and nothing has asked for them yet). This is a reasonable, well-labeled
  scope cut, not a hidden gap.
- The identical `changed`-flag bug in the ordinary Composer UI path
  (`chart-target-lines-pane.component.ts`, per the PR description) is explicitly called out as
  **not** fixed by this PR ("left for a separate fix"). That's a reasonable scope boundary for a
  focused bug-fix PR, not a blocking omission here — but see the non-blocking note below.

- Every refusal path (`requireChart`, `requirePane`, `requireFixedValue`, `resolveMeasure`,
  `normalizeLineStyle`, `normalizeColor`, `requireIndexes`, out-of-range index, chart-type doesn't
  support targets) throws before any mutation is attempted, and each is covered by a corresponding
  `assertNoWrite` assertion in `ChartTargetLineServiceTest`.
- `remove()`'s deliberate ordering — checking `dispatcher.getErrors()` before doing the
  post-write count-based sanity check — is correct and well-reasoned: `editObjectProperty`-style
  composer services report failure via an `ERROR` `MessageCommand` rather than a thrown exception,
  and `ViewsheetSessionService` only converts that to a thrown `CommandErrorException` after the
  mutation lambda returns, so checking the count first would misreport a concurrent-edit conflict
  as "removed only some of what was asked."
- Empty `deletedIndexList`/`Integer[0]` is handled safely by `ChartPropertyService.removeDeletedTargets`
  (`Arrays.sort` on an empty array, loop body never executes).

### Security

- No new authn/authz surface: all three routes gate on the existing `requireEnabled()` check (used
  consistently by every other route in the controller) and resolve through the existing
  `ViewsheetSessionService.resolve`/`mutate`, which is the same session/permission boundary every
  other agent-tool route already relies on.
- All user-supplied strings (`measure`, `value`, `lineStyle`, `lineColor`) are validated against
  fixed regexes/allow-lists before use; none are interpolated into SQL, scripts, or shell commands.
  The regexes (`NUMERIC`, `HEX_COLOR`) are simple, bounded patterns with no catastrophic-backtracking
  shape.

- `AssemblyPropertyService.markTargetsUnchanged` changes behavior for every existing chart
  property write through the wiz agent surface (previously: writing e.g. a title also silently
  mangled existing statistics-target labels; now: it doesn't). This is a pure bug fix — nothing in
  the new code path *adds* the ability to edit an existing target's fields through
  `set_assembly_properties` (that was already impossible; `chartTargets` couldn't be coerced), so
  no previously-working capability is removed.
- `PropertyAliases`'s new refusal converts what was previously a generic `coerce`-failure
  `IllegalArgumentException` (for any raw-path write attempt at `chartTargetLinesPaneModel.*`) into
  a differently-worded, more specific `IllegalArgumentException`. Both are exceptions of the same
  type with a message a caller wasn't parsing for content (per the existing `PropertyAliasesTest`
  pattern, callers are expected to read the message, not match on it), so this is not a breaking
  change to any documented contract.
- Test-file changes to `ViewsheetAssemblyAgentControllerTest` are purely mechanical — a new
  `mock(ChartTargetLineService.class)` argument inserted into every existing
  `controllerWith(...)` helper overload at the correct constructor position — matching the new
  constructor parameter added to `ViewsheetAssemblyAgentController`. I checked several of these
  call sites against the new constructor's parameter list and the ordering is consistent
  throughout.

### Test coverage

- `ChartPropertyServiceTargetRoundTripTest` is a well-motivated addition: it pins the *existing*
  `ChartPropertyService` behavior this PR's `changed=false` bookkeeping depends on, specifically so
  a future "simplification" of that bookkeeping fails loudly rather than silently reintroducing the
  label-collapse bug. Good practice.
- `ChartTargetLineServiceTest` (33 cases per the PR description) covers: the write shape, template
  field completeness (no-NPE), measure resolution (single/ambiguous/`all`/date-refusal/unknown),
  value validation (formula-name shadowing, non-numeric, boundary values), line style/color
  validation, one-mutation-one-checkpoint, removal (position-vs-index-field, out-of-range,
  empty-list), and `list()`'s read-only/no-write guarantee and band/statistics reporting.
- `PropertyAliasesTest` additions cover every spelling of the target-lines path being refused, and
  a negative case confirming the refusal doesn't over-match sibling panes.
- `ViewsheetAssemblyAgentControllerTest` additions confirm the three new routes forward every field
  correctly to the service layer.
- I did not find any existing test that this PR's changes would break: the `markTargetsUnchanged`
  behavior change only affects the `"chart"` branch of `AssemblyPropertyService.set`, and the new
  `ChartPropertyServiceTargetRoundTripTest` explicitly re-confirms `ChartPropertyService`'s own
  pre-existing (unchanged) behavior rather than asserting anything different about it.

## Non-blocking suggestions

1. **Follow-up tracking for the Composer-side twin bug.** The PR explicitly identifies that
   `chart-target-lines-pane.component.ts` has the identical `changed`-flag data-loss bug and defers
   it. Worth confirming a follow-up Redmine issue actually exists/gets filed so it doesn't quietly
   fall off — the PR body reads as aware of this but doesn't cite a tracking number.
2. **No edit-in-place for an existing line target's value/color/style** — only add and remove are
   exposed. This is a reasonable, clearly-documented scope cut for a first pass, not a defect, but
   likely the next natural ask once this ships.
3. **`normalizeColor` only accepts 6-digit `#RRGGBB`.** `Tool.getColorFromHexString` (used
   downstream) also supports an 8-digit alpha-hex form; not required here since `alpha` is handled
   separately by the template, but worth a one-line note if a future caller expects 8-digit input
   to work.
4. **No dedicated `claude/*.md` doc for the `ViewsheetAssemblyAgentController` "wiz" agent-tool
   surface.** Per CLAUDE.md's own instruction to consult feature docs before working an area, this
   fast-growing controller (60+ routes per `git show --stat` on the base branch) has no reference
   doc analogous to `chart.md`/`angular-chart.md`. This PR doesn't regress anything by not adding
   one (none of the ~50 prior routes in this controller are documented there either), but it's
   worth flagging given how large this surface has become.

jshobe-inetsoft (approved):
# Review: stylebi PR #5340 — "Bug #76770: add a working write path for chart target lines"

Base branch: `stylebi-wiz-main-rebase`. Head commit at review time: `f89b4acc` (previously reviewed
commit was `60641b2a`). Files changed: 8 (4 main + 4 test), +1728/-2.

## Verdict

**Ready to merge.** No blocking issues.

This is a re-review after new commits. The only thing that actually changed since the prior
"changes requested" review is that the branch was **rebased onto a newer base tip to resolve the
merge conflict** that blocked merging last time (`mergeStateStatus` is now `CLEAN`, CI is green:
`build`, `Semgrep OSS`, `Semgrep Scan`, `license/cla` all pass). I diffed the previously-reviewed
commit (`60641b2a`) against the current head (`f89b4acc`) restricted to this PR's own files and
confirmed:

- `ChartTargetLineService.java` (the 737-line core of the PR), `ChartTargetLineServiceTest.java`,
  and `ChartPropertyServiceTargetRoundTripTest.java` are **byte-for-byte identical** to what was
  reviewed before.
- `AssemblyPropertyService.java`, `PropertyAliases.java`, `ViewsheetAssemblyAgentController.java`,
  and their tests show a diff between the two commits, but that diff is **base-branch drift**
  absorbed by the rebase (unrelated fixes for bugs #76803/#76886/#76809 that landed on
  `stylebi-wiz-main-rebase` from other merged PRs in the meantime) — none of it is this PR's own
  content. The actual PR diff against the *current* base (`gh pr diff 5340`) shows exactly the
  same target-lines-only changes as before: `markTargetsUnchanged` in `AssemblyPropertyService`,
  the `chartTargetLinesPaneModel` write refusal in `PropertyAliases`, and the three new routes +
  constructor wiring in `ViewsheetAssemblyAgentController`. The rebase reapplied cleanly with no
  missing hunks, no leftover conflict markers, and no corruption.

So the new commits are correct and complete: they resolve exactly the blocker from the last
review (merge conflicts) and change nothing else.

I independently re-verified (not just re-read the PR's own claims) the load-bearing facts against
the actual current source, since a full re-review was requested:

- `TargetInfo.changed` defaults to `true` (`core/.../adhoc/model/property/TargetInfo.java:401`)
  and `ChartPropertyService.getTargetInfo` never calls `setChanged`, confirming the round-trip
  bug `markTargetsUnchanged` fixes is real.
- `ChartPropertyService.updateAllTargets` (`core/.../viewsheet/service/ChartPropertyService.java:605`)
  skips any `TargetInfo` with `changed == false`, only calls `addTarget` when `index == -1`, and
  silently drops an `index >= 0` on an empty descriptor — all matching the PR's stated traps.
- `ChartPropertyDialogService.setChartPropertyModel` dereferences
  `advancePane.getChartTargetLinesPaneModel().getChartTargets()` unguarded
  (`core/.../composer/vs/dialog/ChartPropertyDialogService.java:481-484`), which is exactly what
  `ChartTargetLineService.requirePane()`'s null-check is defending against.
- `ChartDescriptor.removeTarget` is `ArrayList.remove(Object)` (equals-based) against a
  `GraphTarget` whose `equals()` compares every field
  (`core/.../report/composition/graph/GraphTarget.java:750`), confirming the "positional, not
  index-field" removal reasoning; the only theoretical edge case (two field-identical targets at
  different positions) is behaviorally indistinguishable after removal, so it isn't exploitable.
- `PropertyPath.get()` returns `null` on any null intermediate rather than throwing, so
  `AssemblyPropertyService.markTargetsUnchanged` cannot NPE on a chart model missing the pane.
- `PropertyAliases.resolveForWrite` checks the write refusal against both the raw key and the
  resolved path, so the new `chartTargetLinesPaneModel` refusal (matched via `String.contains`)
  cannot be bypassed by an alternate spelling — mirrors the precedent already set for
  `hierarchyPropertyPaneModel.dimensions` (bug #76771) in the same file.
- The controller wiring is mechanically correct: the new `ChartTargetLineService` constructor
  parameter is inserted at the same position in every one of the ~15 test-helper
  `controllerWith(...)` overloads, and the field/constructor assignment in
  `ViewsheetAssemblyAgentController` matches.

## Non-blocking observations (carried over / newly noted, none of them block merge)

1. **`removeDeletedTargets` duplicate-value edge case** (pre-existing utility, not new code): if
   two targets are field-identical (same value/label/color/etc.) but at different positions,
   `ChartDescriptor.removeTarget`'s equals-based `ArrayList.remove` could remove the first
   matching instance rather than the one at the requested position. The PR's own comment
   acknowledges this ("distinct positions make a mis-hit impossible in practice") and the outcome
   is behaviorally identical either way since the two targets are indistinguishable. Not worth
   blocking on, and out of scope for this PR (the bug, if it is one, is in
   `ChartDescriptor`/`GraphTarget`, not in the new service).
2. **Map charts**: `add()` doesn't refuse a target bound to a geographic measure on a map chart
   (`GraphGenerator.addTarget` silently drops it at render time) — the PR deliberately reports
   this via a `note` field in `list()`'s `mapInfo` block rather than refusing on write, since a
   non-geo measure on the same map is valid. This is a reasonable, documented trade-off, not an
   oversight.
3. **Composer parity gap**: the PR notes the Composer's own `chart-target-lines-pane.component.ts`
   has the identical `changed` round-trip bug and is deliberately left unfixed here ("a separate
   fix"). That's a reasonable scope boundary for a backend-agent-tool PR, but it means the
   Composer UI itself is still lossy for this exact scenario — worth a follow-up ticket if one
   doesn't already exist, referenced from the PR description.
4. Minor style: `ChartTargetLineService.add()`'s Javadoc for `measure` says `"all"` binds "the
   chart's only measure" when blank/null but that's actually described in the method body logic
   (auto-bind only when exactly one real measure exists, else refuse) — the Javadoc phrasing
   ("or null/blank to bind the chart's only measure") reads as if null/blank *is* how you request
   `"all"`, when actually blank auto-binds the sole measure and `"all"` is the separate explicit
   opt-in for the unbound line. Worth a one-line Javadoc clarification, but the behavior itself
   (and its test coverage) is correct and the message text at runtime is already clear.

## Test coverage

`ChartTargetLineServiceTest` (unchanged, 27 tests) and `ChartPropertyServiceTargetRoundTripTest`
(unchanged, 9 tests) are thorough, exercise every named trap (NPE on missing bean fields, silent
no-op on bad tabFlag/index, statistics-label collapse, positional vs. index-field removal,
one-mutation-per-call), and are characterization tests against the *existing* `ChartPropertyService`
rather than assertions about the new service, which is the right approach for pinning
undocumented legacy behavior. `PropertyAliasesTest` and `ViewsheetAssemblyAgentControllerTest`
additions cover the refusal message and the three new routes' request-forwarding respectively.
No existing test is broken by this change (CI `build` is green).

fredaan (changes requested):
# Review: PR #5340 — "Bug #76770: add a working write path for chart target lines"

Repo: `inetsoft-technology/stylebi`, base: `stylebi-wiz-main-rebase`, head: `bug-76770`

## Verdict

**Not ready to merge — one blocking issue, mechanical rather than substantive.**

- **Blocking: the branch is out of date and no longer merges cleanly (`mergeable: CONFLICTING` /
  `mergeStateStatus: DIRTY`).** A `git merge-tree` dry run against the current tip of
  `stylebi-wiz-main-rebase` (`8eb3b9d4d`, "Bug #76932 …" #5524) shows a genuine conflict in
  `core/src/test/java/inetsoft/web/wiz/viewsheet/PropertyAliasesTest.java`: a commit already
  merged to the base branch appended a different new test at the same anchor point (end of the
  class) that this PR also appends tests to. `AssemblyPropertyService.java`,
  `ViewsheetAssemblyAgentControllerTest.java`, etc. all auto-merge fine — only this one file
  conflicts, and it's a pure both-sides-appended-near-the-same-line conflict, not a semantic
  disagreement. **Action needed:** rebase `bug-76770` onto the current `stylebi-wiz-main-rebase`
  tip and re-resolve the trivial conflict (keep both blocks) before merge.
- All CI checks that ran are green (`build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla`). The
  disabled `Claude Code Review` workflow isn't present in this check list at all, so there's
  nothing to ignore there.
- No blocking issues found in the code itself — see below. Everything else noted is a non-blocking
  suggestion.

Once rebased, this is good to merge on its merits.

---

## What the PR does

Adds the only working write path for a chart's fixed-value target (goal) lines through the wiz
agent surface: `ChartTargetLineService` (`list` / `add` / `remove`), three new routes on
`ViewsheetAssemblyAgentController`, a `PropertyAliases` refusal that names the new tools, and a
real adjacent-bug fix in `AssemblyPropertyService.set` (clearing `TargetInfo.changed` on
pre-existing targets before every chart property write, to stop an unrelated write from silently
corrupting existing target lines).

## Verification against the actual base-branch source

I did not just read the diff — I pulled the referenced base-branch files
(`core/src/main/java/inetsoft/web/viewsheet/service/ChartPropertyService.java`,
`ChartDescriptor.java`, `SyncChartHandler.java`, `GraphGenerator.java`, `ChartPropertyDialogService.java`,
`ChartTargetLinesPaneModel.java`, `TargetInfo.java`, `MeasureInfo.java`, `ColorInfo.java`,
`PropertyPath.java`, `PropertyAliases.java`, `ViewsheetSessionService.java`,
`CapturingCommandDispatcher.java`) from `origin/stylebi-wiz-main-rebase` and checked every load-bearing
claim the PR description and code comments make. All of them check out:

- `PropertyPath.coerce` really has no path to build a `TargetInfo` (no `String` constructor on the
  class, so it falls through to the generic `"expects TargetInfo, which '{…}' is not."` refusal) —
  confirmed in `PropertyPath.java`.
- `TargetInfo.changed` really defaults to `true` (`private boolean changed = true;`) and
  `getTargetInfo` (the read side) never sets it — confirmed in `TargetInfo.java` /
  `ChartPropertyService.java`. This makes the `AssemblyPropertyService.markTargetsUnchanged` fix
  a real, previously-live bug, not a hypothetical one.
- `updateAllTargets`'s three traps are all real: only `index == -1` reaches `addTarget`; an
  `index >= 0` on an empty chart builds a `GraphTarget` and discards it (`cDescp.getTargetCount()
  == 0` branch creates the object but the `addTarget` call is still gated on `index == -1`); an
  unrecognized `tabFlag` falls through `updateTarget`'s if/else chain silently.
- `updateTargetCommonInfo` dereferences `getLineColor().getColor()`, `getFillAboveColor()`,
  `getFillBelowColor()`, and `getMeasure().getName()` with no null guards, and runs
  `Integer.parseInt(targetInfo.getAlpha())` — an NPE waiting to happen for any hand-built bean,
  exactly as the class doc claims.
- `ChartDescriptor.getTarget(int)` really is `targets.get(idx)` (list position), and
  `removeDeletedTargets` addresses targets that way — so the service's decision to send *list
  position*, not `TargetInfo.getIndex()`, into `deletedIndexList` is correct.
- `SyncChartHandler.syncChartProperties` really clones only targets whose binding still exists
  and keeps each clone's original `index` field, which is exactly how position and the `index`
  field can drift apart — confirmed the mechanism the PR cites as the reason `remove()` must be
  positional.
- `GraphGenerator.addTarget` really returns early for `GraphTypes.isGeo(chartType)` — confirms the
  map-chart caveat surfaced in `list()`'s `note` field.
- The read side really builds `availableFields`/target list with
  `appliedDateComparison || hasDynamic(vsChartInfo)` — confirmed in `ChartPropertyDialogService`,
  which is why `resolveMeasure` must use the pane's own `availableFields` rather than rebuild a
  `MeasureInfo` from a name.
- `CapturingCommandDispatcher.withCapturingDispatcher` really inspects `dispatcher.getErrors()`
  *after* the mutation lambda returns and throws `CommandErrorException` there — confirming
  `remove()`'s early-return-on-error reasoning (returning before the post-write count check is
  correct, not a bug).
- All constructor call sites for `ViewsheetAssemblyAgentController` (18 in the base branch test
  file) are updated with the new `ChartTargetLineService` parameter, and it's the only file in the
  base branch that constructs the controller directly — no other caller was missed.
- The new `POST …/chart/target-lines/delete` route matches the existing convention already used
  for `bookmarks/delete`, `form-table/delete-rows`, `hierarchy/dimensions/delete`,
  `highlights/delete` in the same controller — not a new pattern.

I did not find any place where the PR's description misrepresents the underlying behavior, which
is unusually thorough for a PR of this size.

## Code quality / design

- The choice to build every `TargetInfo` from the pane's own `newTargetInfo` template
  (`fromPrototype`) rather than constructing one from scratch is the right call given the
  unguarded-dereference traps documented above.
- Refusing writes on chart types that don't support targets, on date/time measures, on ambiguous
  multi-measure charts with no `measure` argument, and on the five formula-name strings are all
  real edge cases with real consequences (silent no-op or silently wrong result) if left
  unhandled — good defensive coverage.
- Error messages are unusually actionable (name the tool to use instead, name the valid values,
  name the chart type). This matches the existing house style in this package.
- Test coverage is substantial and split sensibly: `ChartPropertyServiceTargetRoundTripTest`
  characterizes the *existing*, unfixed `ChartPropertyService` behavior the new service depends
  on (so a future "simplification" of the `changed`-flag bookkeeping will fail a test that
  explains why it's there), and `ChartTargetLineServiceTest` covers the new service's own
  contract. This is good practice for a fix that depends on subtle, undocumented behavior of code
  it doesn't own.

## Non-blocking suggestions

1. **Cosmetic message-accuracy mismatch in `requireFixedValue`.** It refuses a value using
   `formula.equalsIgnoreCase(text)` against the five formula names, but
   `ChartPropertyService.isFormulaSupported` does a case-*sensitive* `equals`. So a lowercase
   value like `"max"` is refused here with "is one of the five names the Composer reads as a
   formula" — which isn't quite true for that exact casing (the underlying code would not treat
   lowercase `"max"` as a formula; it would instead hit the "no boundaries" numeric-parse failure).
   Not a functional bug — `"max"` isn't numeric either way, so it's still correctly refused — just
   a slightly inaccurate reason in that one edge case. Test
   `refusesAValueThatIsNotAFixedNumber` doesn't check message content, so nothing is asserting the
   wrong thing; this is purely a "the message could be more precise" nit.
2. **`remove()` does an extra read-back round trip** (`chartService.getChartPropertyDialogModel`
   again after the write) purely to verify the removal count. This is explicitly reasoned about in
   the code comments as intentional and cheap for a non-hot path — flagging only so a future
   optimization pass doesn't remove it without understanding why it's there (a `ChartDescriptor`
   target-removal is equals-based on every field of a `GraphTarget`, so a silent mis-hit is exactly
   the failure mode this guards against).
3. **Documentation maintenance convention.** `claude/wiz-agent-viewsheet-binding.md` says: "when
   fixing a bug, update the matching category under 'Testing by Business Area'". This PR fixes a
   real, previously-live bug (`TargetInfo.changed` defaulting `true` and never being cleared,
   silently corrupting statistics-target labels and band fill colors on every unrelated chart
   property write) that fits squarely under the existing "Property write validation
   (`AssemblyPropertyService` / `PropertyAliases`)" category in that doc, but the doc wasn't
   updated with an entry for #76770. Not blocking, but worth a follow-up per the repo's own stated
   convention.
4. The PR description itself flags "the live end-to-end run through the plugin's own tools against
   a running instance" as **not yet done** — worth confirming before/soon after merge, since this
   is exactly the kind of change (bean construction feeding a rendering pipeline) where a
   unit-test-clean but visually-wrong result is possible. Not a reason to block given how
   thoroughly the round trip is characterized against the real `ChartPropertyService`, but worth
   flagging since the author flagged it themselves.

## Security / regressions / backward compatibility

- No injection surface: values are constrained by an explicit numeric regex, a hex-color regex, and
  a whitelist map for line styles; measure binding only ever copies an existing `MeasureInfo` out
  of the model's own `availableFields`, never constructs one from an arbitrary client string.
- `AssemblyPropertyService.markTargetsUnchanged` is scoped to `"chart".equals(type)` only, touches
  only the `changed` flag (no other field), and only when the resolved path actually holds a
  `TargetInfo[]` — low blast radius, and it only *removes* a latent corruption path rather than
  changing any other observable behavior of unrelated chart writes.
- No existing REST contract changes; the three new routes are additive. The one constructor
  signature change (`ViewsheetAssemblyAgentController`) is Spring-DI-internal and its only direct
  instantiator (the test file) is updated in the same PR.


---

**Merge blocker (current state):** GitHub reports this PR as `CONFLICTING` with `stylebi-wiz-main-rebase`. Please rebase and resolve the conflict in `PropertyAliasesTest.java`.

PR #5342, opened 2026-09-17
Bug #76768, SelectionList measure + SelectionTree ID-hierarchy dynamic/literal binding

jkidacki-inetsoft:
## Summary

`set_selection_source` (the composer-chat plugin's tool for binding a SelectionList/SelectionTree/RangeSlider/Calendar to a table/column) had two gaps:

1. SelectionList's `measure` (an optional bar-chart aggregate column) couldn't accept `"$(ComponentName)"` — StyleBI's own `DynamicValue` convention for pointing a value at a Form component's live value, already wired up for other fields in #76641.
2. SelectionTree had no way at all — not even with literal column names — to bind its ID-hierarchy mode (`SelectionTreeVSAssemblyInfo.ID`, an arbitrary-depth tree from one flat, self-referencing table via parentId/id/label columns), distinct from the existing fixed-hierarchy-levels `columns` array.

## Root Cause / Fix

**`measure`**: `SelectionBaseVSAssemblyInfo.measureValue` was already a `DynamicValue` and the REST path (`SelectionBindingService.setSource` → `SelectionMeasurePaneModel.setMeasure`) forwarded the string untouched — a pure plugin-side validation/wiring gap (see the paired stylebi-wiz PR). No server-side change was needed for this half beyond documentation.

**SelectionTree ID-hierarchy mode**: `SelectionBindingService.setSource` gained three new params, `parentIdColumn`/`idColumn`/`labelColumn`, required together (all-or-nothing) and mutually exclusive with `columns`. Each is resolved into an `OutputColumnRefModel` (reusing the existing `resolveColumns`/`columnRef` helpers) and forwarded via `SelectionTreePaneModel`'s existing `setParentId`/`setId`/`setLabel` + `setParentIdRef`/`setIdRef`/`setLabelRef` — machinery the interactive Composer UI already exercises.

Initially these three were designed to also accept `"$(ComponentName)"`, mirroring `measure`. A code review caught that this doesn't actually work: `SelectionTreeVSAQuery2.refreshSelectionValue0` locates the id/parentId/label columns by matching `assembly.getDataRefs()[i].getName()` against the *resolved* value of the `DynamicValue` — and a dynamic field's ref is never added to `dataRefs` in the first place (there's no real column to add), so it can never match at runtime. This is structurally different from `measure`'s dynamic value, which is resolved through a separate per-row aggregation path that never touches `dataRefs`. So `parentIdColumn`/`idColumn`/`labelColumn` are literal-column-only; a `"$(...)"`-shaped value is refused outright (`refuseIfDynamic`), both here and client-side in the plugin, with an explanatory error.

## Test Plan

- `SelectionBindingServiceTest`/`BindingAgentControllerTest` (41 tests) pass, including new coverage for ID-mode binding, dynamic-reference refusal, partial-params refusal, `columns`+ID-mode mutual exclusion, unknown-column refusal, and non-tree-assembly refusal.
- Full `community/core` module compiles and the full multi-module reactor build succeeds.
- Live-verified against a running server (paired with the stylebi-wiz plugin from the companion PR): `measure` round-trips a `"$(ComponentName)"` reference correctly, and the ID-hierarchy mode correctly renders a real self-referencing table's parent/child structure (a control test with the existing COLUMN-mode binding on the same data rendered correctly, isolating an earlier apparent-empty result to needing real parent/child pairs, not a defect).
- Two independent code-review passes (one that found and one that verified the fix for the dynamic-reference runtime bug above).

Companion plugin PR: https://github.com/inetsoft-technology/stylebi-wiz/pull/2647

Fixes Redmine #76768.


jshobe-inetsoft (changes requested):
# Review: PR #5342 — `SelectionList` measure + `SelectionTree` ID-hierarchy dynamic/literal binding

Repo: `inetsoft-technology/stylebi` · Base: `stylebi-wiz-main-rebase` · Head: `bug-76768`
Files: `core/src/main/java/inetsoft/web/wiz/binding/BindingAgentController.java`,
`core/src/main/java/inetsoft/web/wiz/binding/SelectionBindingService.java`,
`core/src/test/java/inetsoft/web/wiz/binding/SelectionBindingServiceTest.java`

## Verdict

**Request changes.** One blocking correctness issue:

1. **Case-sensitivity regression in the new `SelectionTree` ID-hierarchy path** — `SelectionBindingService.setSource()` writes the caller-supplied `parentIdColumn`/`idColumn`/`labelColumn` strings *as given* into `SelectionTreePaneModel.setParentId/setId/setLabel` (which flow into `SelectionTreeVSAssemblyInfo`'s `DynamicValue`-backed `parentIDValue`/`idValue`/`labelValue`), while the *canonicalized* column name (resolved case-insensitively, same as every other field this service resolves) is what gets written into `dataRefs` via `parentIdRef`/`idRef`/`labelRef`. At runtime, `SelectionTreeVSAQuery2.refreshSelectionValue0` matches `dataRefs[i].getName()` against `assembly.getID()`/`getParentID()`/`getLabel()` with `Tool.equals(a, b)` — which defaults to **case-sensitive** comparison (`CoreTool.equals(Object, Object)` → `equals(obj, obj2, true)`). If a caller passes a column name in a different case than the DB's actual column (e.g. `idColumn: "state_id"` for a real column `STATE_ID`), `set_selection_source` reports success with no error, but the tree silently renders empty (`idColumn` case: `idIndex == -1` short-circuits `refreshSelectionValue0` — the exact failure mode the PR's own `refuseIfDynamic`/javadoc set out to prevent) or silently drops the hierarchy/labels (`parentIdColumn`/`labelColumn` case).

   This is not hypothetical edge-case paranoia: this exact file already fixed the identical class of bug once, for `additionalTables`, as Bug #76747 (see `resolveAdditionalTables`'s doc comment, and the `resolvesAMixedCaseAdditionalTableToItsCanonicalName` test) — "an entry that does not match a real worksheet table-assembly name (including a correctly-spelled one in the wrong case) ... silently skips ... with no error surfaced back to the caller." `table`, `additionalTables`, and `columns` are all resolved case-insensitively and then the *canonical* value is what gets persisted. The three new ID-hierarchy fields are the only ones in this service that resolve case-insensitively (via `resolveIdModeColumnRef` → `resolveColumns`) but then persist the *raw, un-canonicalized* input instead of the resolved value. No test exercises a case-mismatched literal column name for ID mode, so this wasn't caught.

   **Suggested fix**: in the `idMode` branch, call `pane.setParentId(parentIdRef.getAttribute())`, `pane.setId(idRef.getAttribute())`, `pane.setLabel(labelRef.getAttribute())` (or otherwise thread the resolved `BindableField.column()` value through) instead of the raw `parentIdColumn`/`idColumn`/`labelColumn` parameters — mirroring how `resolveAdditionalTables`/`resolveTable` already do this for `table`/`additionalTables`. The `result.put("parentIdColumn", parentIdColumn)` triple a few lines down should use the same canonicalized values for consistency with `result.put("table", resolvedTable)` and the `columns` response field, which already report canonical names.

No other blocking issues found. CI is otherwise green (`build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla` all SUCCESS); only `claude-review` shows FAILURE, which per review instructions is excluded from the merge-readiness bar since this manual review takes its place. PR is `MERGEABLE` with no merge conflicts.

---

## Detailed Findings

### Correctness

- **[Blocking]** Case-sensitivity mismatch between `dataRefs` (canonical) and `parentId`/`id`/`label` `DynamicValue`s (raw input) in ID-hierarchy mode — see Verdict above. This is the one issue I'd want fixed before merge.

- **[Verified, not a bug]** The PR's central design claim — that `parentIdColumn`/`idColumn`/`labelColumn` cannot accept a `"$(ComponentName)"` dynamic reference the way `measure` can — checks out against the actual runtime code. I traced `SelectionTreeVSAQuery2.refreshSelectionValue0` (community core): it resolves `assembly.getID()`/`getParentID()`/`getLabel()` (which evaluate the `DynamicValue`'s *runtime* value — for `"$(Foo)"` that's Foo's live value, not a column name) and matches that string, by name, against `assembly.getDataRefs()[i].getName()`. Since `setAssemblyInfoDataRefs`'s ID-mode branch only adds a `ColumnRef` to `dataRefs` for a literal `parentIdRef`/`idRef`/`labelRef` (there is no ref to add for a dynamic value — no real column exists), a dynamic reference could never find a matching index at runtime, exactly as documented. Refusing it outright (`refuseIfDynamic`) rather than accepting a value that silently renders broken is the right call, and matches the class's stated philosophy elsewhere (`requireRepoint`, `resolveAdditionalTable`).

- **[Verified, not a bug]** Argument ordering across `BindingAgentController.SelectionSourceRequest` (record), `BindingAgentController.setSelectionSource()`'s call, `SelectionBindingService.setSource()`'s signature, and every updated call site in the test file was checked positionally and is consistent (`... measure, parentIdColumn, idColumn, labelColumn, force ...` throughout). No parameter-swap bug.

- **[Verified, not a bug]** Exception-safety: if `resolveIdModeColumnRef` throws partway through the `SelectionTreeVSAssembly` idMode branch (e.g. an unknown `idColumn`), the local `model`/`pane` DTO mutations made so far are simply discarded — `selectionTreeService.setSelectionTreePropertyModel(...)` (the actual write) is only called after all three refs resolve successfully, so no partial state is persisted. This matches the pre-existing behavior of every other validation failure in this class (e.g. an unknown `measure`... actually `columns`) and is unchanged by this PR.

- **[Verified, not a bug]** `SelectionTreePropertyDialogService.setAssemblyInfoDataRefs`'s ID-mode branch only reads `model.getParentIdRef()/getIdRef()/getLabelRef()` (not `getSelectedColumns()`), and its COLUMN-mode branch only reads `getSelectedColumns()` — so there's no cross-contamination between a tree's previous COLUMN-mode `selectedColumns` and a new ID-mode bind, or vice versa, even though the new code never explicitly clears the field it isn't using.

- **[Verified, not a bug]** `refusesIdModeColumnsOnANonTreeAssembly` — passing ID-mode params to a `SelectionListVSAssembly` correctly falls through to the pre-existing "columns is empty" arity check (since ID-mode fields never populate `columns`), producing a message that names "selection list" as the test expects. The error text ("needs exactly 1 column in 'columns', got 0") doesn't explicitly say "ID-hierarchy mode isn't valid here", which is a little indirect for a caller debugging blind, but it's accurate and covered by a passing test — not a correctness bug, just a minor clarity nit (see Non-blocking below).

### Clean Code / Maintainability

- **[Non-blocking]** `refuseIfDynamic` reimplements dynamic-value detection (`value.trim().startsWith("$(")`) instead of reusing the existing canonical helper `VSUtil.isVariableValue(String)` (`inetsoft.uql.viewsheet.internal.VSUtil`, used elsewhere in the codebase for exactly this check: `val != null && val.startsWith("$(") && val.endsWith(")")`). The new check is looser (no `endsWith(")")` requirement), which is harmless here since it only widens what gets *refused* (a real column name would never start with `"$("`), but it's needless duplication of logic that already has a canonical home. Consider calling `VSUtil.isVariableValue(value.trim())`.

- **[Non-blocking]** `resolveIdModeColumnRef` calls `refuseIfDynamic` a second time even though `setSource` already validates all three fields up front before entering `sessions.mutate(...)` (the comment at the call site even says so). This is defensible defensive programming for the private helper's own contract, but it is redundant in the only path that calls it today; worth a one-line comment noting it's intentionally belt-and-suspenders, or dropping it if the maintainers prefer the up-front check to be the single source of truth.

- **[Non-blocking]** The success response for ID-hierarchy mode includes both the type-specific `parentIdColumn`/`idColumn`/`labelColumn` keys *and* the generic `columns: []` key (since `resolvedColumns` is empty in `idMode`). Not wrong, but a caller parsing the response has to know to ignore `columns` in this mode. Once the case-sensitivity fix above is applied, these three values should also be the canonicalized names, matching how `table` already reports `resolvedTable` rather than the raw input.

### Readability

- Javadoc is unusually thorough and directly traces through the runtime query-matching code to justify the literal-only restriction — this is exactly the kind of documentation that will save the next person a debugging session. No issues.
- Method decomposition (`requireIdModeAllOrNothing`, `resolveIdModeColumnRef`, `refuseIfDynamic`) is clean and each piece is independently testable, which the test file exploits well.

- All-or-nothing validation for `parentIdColumn`/`idColumn`/`labelColumn` (`requireIdModeAllOrNothing`) correctly names exactly which fields are present/missing rather than a generic error. Good.
- Mutual exclusion with `columns` is validated before any session mutation. Good.
- Unknown-column and dynamic-value refusals both happen before `sessions.mutate(...)` runs for the dynamic-value case, and inside it (but before any write) for the unknown-column case via `resolveColumns`'s existing behavior — both fail cleanly with no partial write, as discussed above.
- Not covered (edge case, not exercised by tests): what happens when `idMode` is requested against `TimeSliderVSAssembly`/`CalendarVSAssembly` — traced manually above; it falls through to the existing arity check with `resolvedColumns` empty and produces a reasonable, if generic, error. Fine, just untested; not blocking.

### Security

- No new attack surface: the three new fields are plain `String` request-body fields resolved against the same case-insensitive `BindableField` allow-list (`list_bindable_fields`) already used for `columns`/`table`/`additionalTables`; there's no SQL/script construction from raw input, and the existing session/pairing/ownership checks (`ViewsheetSessionService.mutate`, unchanged by this PR) still gate the whole endpoint. `refuseIfDynamic` rejecting a `"$(...)"`-shaped literal is a safety measure, not a security control, but it doesn't introduce any injection risk either way (DynamicValue resolution for a live component reference isn't script evaluation of caller-controlled code).

- `SelectionSourceRequest` is a Java record; the three new fields are appended at the end and are nullable/optional, so existing callers that don't send them continue to work (Jackson record deserialization defaults absent fields to `null`), and the existing `measure`/`columns`-only flows for `SelectionList`/`TimeSlider`/`Calendar`/`SelectionTree`-COLUMN-mode are unchanged apart from being moved into an explicit `else` branch — verified line-by-line against the pre-PR version, behavior is identical for the non-idMode path.
- The pre-existing `SelectionTreeVSAQuery`/COLUMN-mode path is untouched by this PR (mode is now set explicitly rather than unconditionally to `COLUMN`, but the resulting value for every existing caller is the same `COLUMN` constant as before).
- No other call sites of `SelectionBindingService.setSource(...)` exist in the tree besides `BindingAgentController` and the test file (checked via `git grep`), so no other caller was missed.

### Tests

- Good coverage for the new happy path, dynamic-reference refusal (naming the specific offending field), partial-params refusal, `columns`+ID-mode mutual exclusion, unknown-column refusal (reusing the existing "names what's available" convention), and non-tree-assembly refusal. All existing tests were mechanically updated for the new parameter positions and still assert the same behavior.
- **Gap**: no test exercises a case-mismatched literal column name for `parentIdColumn`/`idColumn`/`labelColumn` (e.g. `idColumn: "state"` against a table field named `"STATE"`) — this is exactly the scenario that would have caught the blocking issue above. Given `resolvesAMixedCaseAdditionalTableToItsCanonicalName` already exists as a template for this pattern in this same test file, adding an analogous `resolvesAMixedCaseIdModeColumnsToTheirCanonicalNames` test (asserting `pane.getId()`/`pane.getParentId()`/`pane.getLabel()` equal the canonical DB casing, not the caller's input) would both catch the current bug and prevent a silent recurrence.

jkidacki-inetsoft:
Rebased onto the current `stylebi-wiz-main-rebase` (which had picked up #5425, an independently-merged PR wiring the same `parentIdColumn`/`idColumn`/`labelColumn` ID-hierarchy fields) and addressed the review.

**Merge conflict resolution:** #5425 landed the same underlying feature while this PR was open, but — unlike this PR's version — it already persisted the *canonicalized* `BindableField.column()` value into `pane.setParentId`/`setId`/`setLabel` instead of the raw caller input. That's exactly what @jshobe-inetsoft's review flagged as this PR's blocking case-sensitivity bug. I kept #5425's correct canonicalization and layered this PR's actual unique contributions on top:
- `measure`'s `"$(ComponentName)"` dynamic-value support + its javadoc (server-side no-op, already worked — this PR only documents and validates it).
- The explicit `refuseIfDynamic` check (and its detailed javadoc explaining *why* it can't work for id-mode fields, tracing `SelectionTreeVSAQuery2.refreshSelectionValue0`), so a dynamic id-mode value gets a clear, specific error instead of a generic "not a column" one.
- `requireIdModeAllOrNothing`'s more detailed partial-params error message (names exactly which fields are given vs. missing).
- Removed the now-dead `validateIdModeFields`/`isBlank` helpers #5425 had added, since `requireIdModeAllOrNothing` supersedes them.

**Review fix applied:** `pane.setParentId(parentIdRef.getAttribute())` etc. (and the `result.put(...)` response fields) now use the resolved, canonical column name — matching how `table`/`additionalTables` already report canonical names. Added `resolvesMixedCaseIdModeColumnsToTheirCanonicalNames`, mirroring the existing `resolvesAMixedCaseAdditionalTableToItsCanonicalName` test, to cover exactly the case-mismatch scenario the review called out.

**Non-blocking suggestion evaluated and reverted:** I initially switched `refuseIfDynamic` to the suggested canonical `VSUtil.isVariableValue`, but that broke every id-mode unit test — `VSUtil`'s static initializer reaches into a live Spring `ApplicationContext`, which `SelectionBindingServiceTest`'s plain-Mockito setup doesn't provide (`NoClassDefFoundError`/`ExceptionInInitializerError` instead of a clean `IllegalArgumentException`). Kept the local `startsWith("$(")`/`endsWith(")")` check instead, with a comment explaining why.

Verified: `SelectionBindingServiceTest` (now 12 tests, including the new mixed-case regression test) and `BindingAgentControllerTest` pass; `core` module compiles clean.

jshobe-inetsoft (approved):
# Review: PR #5342 — `SelectionList` measure + `SelectionTree` ID-hierarchy dynamic/literal binding (re-review of new commits)

Repo: `inetsoft-technology/stylebi` · Base: `stylebi-wiz-main-rebase` · Head: `466d5007e3d6bf5967632ca8956fe2c761f41e12` (`bug-76768`)
Files changed since base: `core/src/main/java/inetsoft/web/wiz/binding/SelectionBindingService.java`, `core/src/test/java/inetsoft/web/wiz/binding/SelectionBindingServiceTest.java`
(`BindingAgentController.java`'s parameter plumbing is unchanged in this diff — it was already merged into the base via the independently-landed PR #5425, per the author's rebase note.)

## Verdict

**Ready to merge.** The one blocking issue from my prior review — the case-sensitivity regression in the new `SelectionTree` ID-hierarchy path — is fixed correctly in the current head, and is now covered by a targeted regression test. I found no other blocking issues. Only non-blocking polish suggestions remain (see below).

CI is green (`build`/Validation Build, `Semgrep Scan`, `Semgrep OSS`, `license/cla` all SUCCESS); PR is `MERGEABLE` with no conflicts.

---

## Verification of the fix for my earlier blocking finding

**Original finding:** `setSource()` was writing the caller-supplied `parentIdColumn`/`idColumn`/`labelColumn` strings *as given* into the `DynamicValue`-backed `parentId`/`id`/`label` fields, while the *canonicalized* name (resolved case-insensitively) went only into `dataRefs`. Since `SelectionTreeVSAQuery2.refreshSelectionValue0` matches `dataRefs[i].getName()` against `assembly.getID()`/`getParentID()`/`getLabel()` with a case-sensitive `Tool.equals`, a case-mismatched literal (e.g. `idColumn: "state_id"` for a real `STATE_ID` column) would report success but silently render an empty or broken tree.

**Current code** (`SelectionBindingService.java`, idMode branch):

```java
OutputColumnRefModel parentIdRef = resolveIdModeColumnRef(
   tables, assemblyName, resolvedTable, "parentIdColumn", parentIdColumn);
OutputColumnRefModel idRef = resolveIdModeColumnRef(
   tables, assemblyName, resolvedTable, "idColumn", idColumn);
OutputColumnRefModel labelRef = resolveIdModeColumnRef(
   tables, assemblyName, resolvedTable, "labelColumn", labelColumn);

pane.setMode(SelectionTreeVSAssemblyInfo.ID);
pane.setParentId(parentIdRef.getAttribute());
pane.setId(idRef.getAttribute());
pane.setLabel(labelRef.getAttribute());
pane.setParentIdRef(parentIdRef);
pane.setIdRef(idRef);
pane.setLabelRef(labelRef);
...
result.put("parentIdColumn", parentIdRef.getAttribute());
result.put("idColumn", idRef.getAttribute());
result.put("labelColumn", labelRef.getAttribute());
```

`resolveIdModeColumnRef` resolves the value case-insensitively via the existing `resolveColumns` helper and returns an `OutputColumnRefModel` built from the matched (canonical) `BindableField`, via `columnRef()`. So `parentIdRef.getAttribute()`/`idRef.getAttribute()`/`labelRef.getAttribute()` are now the **same canonical string** that ends up in `dataRefs` (via `SelectionTreePropertyDialogService.setAssemblyInfoDataRefs` → `createColumnRefFromModel`, which builds the `ColumnRef`'s name from the same `attribute`). Both the `DynamicValue` strings and `dataRefs` now agree, so the case-sensitive match in `refreshSelectionValue0` succeeds regardless of the caller's input casing. I traced this end-to-end:

- `resolveColumns` (unchanged helper): matches `columns.get(i)` against `BindableField.column()` with `equalsIgnoreCase`, returns the resolved (DB-canonical) `BindableField`.
- `columnRef()`: sets `attribute`/`name` from `field.column()` (canonical), not the raw input.
- `SelectionTreeVSAQuery2.refreshSelectionValue0` (`core/src/main/java/inetsoft/report/composition/execution/SelectionTreeVSAQuery2.java:77-88`): `Tool.equals(refs[i].getName(), assembly.getID())` etc. `Tool extends CoreTool`, and `CoreTool.equals(Object, Object)` defaults `sensitive=true` (case-sensitive) — confirmed by reading `CoreTool.java:218-225`.

**Test coverage:** `resolvesMixedCaseIdModeColumnsToTheirCanonicalNames` (new) passes `"city"`/`"state"`/`"order_date"` against a mock table whose fields are `CITY`/`STATE`/`ORDER_DATE`, and asserts `pane.getParentId()`/`getId()`/`getLabel()` **and** `getParentIdRef().getAttribute()`/etc. **and** the response map all come back as the canonical uppercase names. This is exactly the regression scenario, modeled directly on the existing `resolvesAMixedCaseAdditionalTableToItsCanonicalName` test for the analogous `additionalTables` bug (#76747). I confirmed the mock fixture (`harness()`) actually has uppercase field names, so this test would have failed against the pre-fix code and passes against the current code — it's a real, valid regression test, not a tautology.

This closes the finding. No other correctness issue found in the diff.

## Author's response — assessed

The PR comment accurately describes what happened: PR #5425 landed the same `parentIdColumn`/`idColumn`/`labelColumn` wiring independently while this PR was open, and in doing so *also* fixed the case-sensitivity bug (it already persisted `BindableField.column()`, not raw input). The rebase kept that canonicalization and layered this PR's own contributions (dynamic-value refusal + its javadoc, the `requireIdModeAllOrNothing` all-or-nothing error message, dead-code removal) on top. I verified this by reading the diff's removed/base lines (which already computed and used the canonical `BindableField`) versus the added lines (which restructure around `OutputColumnRefModel`s returned from a new helper but preserve the same canonicalization) — the description holds up.

The revert of `VSUtil.isVariableValue` in favor of a local `startsWith("$(")/endsWith(")")` check is a reasonable, well-justified engineering trade-off (avoiding pulling `VSUtil`'s heavier static dependencies into a plain-Mockito unit test), and is called out with an inline comment explaining why — good practice, not a red flag.

## Other findings (unchanged from before / newly reviewed)

### Non-blocking

- **Duplicated `setSelectionTreePropertyModel(...)` call.** The refactor moved this call from a single site after the `if(idMode) {...} else {...}` block into *both* branches individually (compare removed lines at the end of the old `else` block vs. the two new call sites inside each branch). Functionally equivalent — it still runs exactly once per invocation — but it's now duplicated code where a single call after the if/else would do. Minor DRY nit; not worth blocking on.
- **`result.put("columns", columnNames)` still runs unconditionally**, including in ID-hierarchy mode where `resolvedColumns` is empty, so a successful ID-mode response contains both the type-specific `parentIdColumn`/`idColumn`/`labelColumn` keys and an empty `columns: []`. Harmless, but a caller has to know to ignore `columns` in this mode. (Flagged in my prior review too; not addressed, still low-priority.)
- **`resolveIdModeColumnRef` calls `refuseIfDynamic` a second time** even though `setSource` already validates all three fields up front before `sessions.mutate(...)` runs. Defensible belt-and-suspenders for the private helper's own contract, but redundant on the only path that calls it today. (Also flagged previously; author chose not to change it — fine either way.)
- **No dedicated unit test exercises `measure` accepting a `"$(ComponentName)"` value** (e.g. asserting `pane.getSelectionMeasurePaneModel().getMeasure()` equals the dynamic string unchanged). Since this half of the PR is a pure documentation change over already-existing, untouched code (`pane.getSelectionMeasurePaneModel().setMeasure(measure)` forwards any string, dynamic or not, and was not modified by this PR), a test isn't strictly required to justify merging, but it would guard against a future accidental regression (e.g. someone later adding validation to that setter).

### Verified, not bugs

- Argument ordering across `BindingAgentController.SelectionSourceRequest` → its call site → `SelectionBindingService.setSource()`'s signature is consistent end-to-end (checked positionally): `assembly, table, columns, additionalTables, measure, parentIdColumn, idColumn, labelColumn, force` throughout. `BindingAgentController.java` wasn't touched by this diff, but I re-verified it directly against the current PR head since the previous review's check predates the rebase.
- The dynamic-value refusal for `parentIdColumn`/`idColumn`/`labelColumn` is correct and necessary, confirmed independently by reading `SelectionTreeVSAQuery2.refreshSelectionValue0` and `SelectionTreePropertyDialogService.setAssemblyInfoDataRefs`: a dynamic value never gets an entry added to `dataRefs`, so it could never be matched at runtime regardless of the case-sensitivity fix. Refusing it loudly is the right design.
- `measureValue` on `SelectionBaseVSAssemblyInfo` is confirmed to be a `DynamicValue` field (`core/src/main/java/inetsoft/uql/viewsheet/internal/SelectionBaseVSAssemblyInfo.java:1015`), supporting the PR's claim that `measure`'s dynamic-reference support required no server-side behavior change.
- Backward compatibility: the new fields are appended, optional/nullable record components; existing callers that omit them are unaffected. The non-idMode path is untouched aside from being made an explicit `else` branch.

### Not covered by CLAUDE.md's `claude/*.md` docs

This change is entirely in the server-side `wiz`/composer-chat REST binding layer (`inetsoft.web.wiz.binding`), not the Angular selection UI. `claude/angular-selection.md`'s "MODE.COLUMN vs MODE.ID" section describes a different, UI-side selection-propagation concept (click behavior in the rendered tree), not `SelectionTreeVSAssemblyInfo.ID` (the data-binding ID-hierarchy mode this PR adds); I confirmed there's no overlap or contradiction. No other `claude/*.md` document covers this wiz-binding-controller area.

## Tests

- `SelectionBindingServiceTest` now has 12 id-mode-related and pre-existing tests updated for the new parameter order; all reviewed and each assertion traces correctly to the corresponding source behavior (verified manually, not just by running — I don't have a way to execute the Maven build in this review, but CI's `build` check is green).
- The new `resolvesMixedCaseIdModeColumnsToTheirCanonicalNames` test is a real, valid regression test for the fixed bug (see above).
- No test is broken by this change; all updated call sites use the current 11-parameter `setSource` signature consistently.

PR #5343, opened 2026-09-17
Bug #76761, serialize chart flyover with table/crosstab flyover locks

jkidacki-inetsoft:
## Summary

When a Chart and a Table/Crosstab in the same viewsheet are both configured with the same Flyover View target, rapidly moving the mouse between them could leave the viewsheet stuck showing a loading spinner indefinitely.

## Root Cause

Bug #76532 fixed the analogous table-to-table race by having `BaseTableFlyoverService.applyFlyovers()` serialize concurrent flyover requests that touch the same target assembly, using per-name lock objects obtained from `ViewsheetSandbox.getFlyoverLock(name)`. `VSChartFlyoverService` (the chart-side flyover path) performs the same save/mutate/execute/restore sequence on a shared target's condition state (`PreRuntimeConditionList` save/restore, `TipConditionList` mutation, `executeView`/`execute`), but never acquired any of these locks. When a Chart and a Table/Crosstab shared the same Flyover View target, their flyover requests could interleave that sequence and corrupt the shared target's condition state, leaving the viewsheet's loading cycle for that target never resolving.

## Fix

Add the same per-target `flyoverLock` serialization to `VSChartFlyoverService.processFlyover()`, mirroring `BaseTableFlyoverService.applyFlyovers()`/`applyFlyoversLocked()`:
- Compute the same lock name set (source table name + flyover target view names, sorted via `TreeSet` for deterministic acquisition order).
- Recursively acquire `ViewsheetSandbox.getFlyoverLock(name)` monitors for each name before running the flyover apply/execute/restore logic.

Unlike the table-side fix, no `unlockAll()`/`restoreLocks()` dance around lock acquisition is needed here, since `VSChartFlyoverService`'s call chain never holds the sandbox's own read/write lock at the point it enters these per-target locks (verified by tracing `eventHandler()` → `VSChartControllerService.processEvent()` → `VSChartStateInfo.createChartState()`).

## Test Plan

- Reproduced with the attached "Risk Dashboard V3.zip" asset (Chart + Crosstab sharing a Flyover View target); rapid mouse movement between them no longer gets the viewsheet stuck loading.
- `./mvnw clean test-compile -pl core -am` — builds cleanly.
- `./mvnw test -pl core -am` — full `core` module test suite passes, no regressions.
- Code reviewed by a code-reviewer subagent: lock acquisition order, deadlock-freedom (no sandbox lock held while entering a flyover-lock monitor), and behavioral equivalence of the extracted method all verified against the codebase.

Fixes Redmine #76761.


jshobe-inetsoft (approved):
# Review: stylebi PR #5343 — "Bug #76761, serialize chart flyover with table/crosstab flyover locks"

Repo: `inetsoft-technology/stylebi` (community). Branch `bug-76761` → `main`. 1 file changed, +74/-4:
`core/src/main/java/inetsoft/web/viewsheet/controller/chart/VSChartFlyoverService.java`

## Verdict

- CI: `build`, `Semgrep Scan`, `Semgrep OSS`, and `license/cla` all pass. `claude-review` fails, but per the review instructions for this task that automated check is excluded — this manual review takes its place.
- Merge conflicts: none (`mergeable: MERGEABLE`). GitHub's `mergeStateStatus: BLOCKED` appears to stem from branch-protection requiring an approving review (and/or the excluded `claude-review` check) rather than from any conflict or code defect.
- No blocking correctness, safety, or regression issues were found in the diff after tracing the full lock-acquisition call chain (details below).

Non-blocking suggestions are listed at the end; none of them should hold up the merge.

---

## What the change does

Bug #76532 previously fixed a race where two *table/crosstab* flyover requests hitting the same Flyover-View target could interleave their save/mutate/execute/restore sequence on the target's condition state, by having `BaseTableFlyoverService` serialize per-target via monitor objects obtained from `ViewsheetSandbox.getFlyoverLock(name)`.

`VSChartFlyoverService.processFlyover()` (the chart-side flyover path) runs the analogous save/mutate/execute/restore sequence but never acquired those locks, so a **Chart** flyover and a **Table/Crosstab** flyover pointed at the same target could still interleave and corrupt the target's condition state, leaving the viewsheet's loading cycle stuck.

This PR adds the same per-target locking to the chart side:
- `processFlyover()` now computes the identical lock-name set as `BaseTableFlyoverService.applyFlyovers()` (source table assembly name + flyover target absolute names, collected into a `TreeSet<String>` for deterministic acquisition order), resolves each to a lock object via `box.getFlyoverLock(name)`, and delegates to a new `processFlyoverLocked()`.
- `processFlyoverLocked()` recursively enters `synchronized` blocks on each lock (same recursive-index pattern as `BaseTableFlyoverService.applyFlyoversLocked()`), then calls a new `doProcessFlyover()`, which contains the original method body unchanged (apart from the early-return for empty `views` being hoisted above the now-deferred `preList`/`tassembly`/`vs` setup).

## Correctness / deadlock-safety analysis

The PR description's central technical claim is that, unlike the table-side fix, no `unlockAll()`/`restoreLocks()` dance is needed around lock acquisition here because the chart-side call chain never holds the sandbox's own read/write lock at the point it enters the per-target `flyoverLock` monitors. I traced this independently rather than taking it on faith, since getting it wrong reproduces exactly the deadlock class bug #76532's "round 1" fix had to correct (a thread blocked *entering* a `flyoverLock` monitor while still holding the sandbox lock, racing against another thread that holds that monitor and is blocked upgrading to the sandbox write lock):

- `VSChartFlyoverService.eventHandler()` → `processEvent()` → `VSChartControllerService.VSChartStateInfo.createChartState()`: this acquires `box.lockRead()` but releases it in a `finally` block *before* `createChartState()` returns. So by the time `eventHandler`'s lambda runs, no sandbox lock is held.
- `eventHandler()` then calls `box.getVGraphPair(...)` before `processFlyover()`. For a top-level (non-embedded, no `.` in name) chart, this method only takes its own private per-chart `graphLock` (a `ReentrantLock`, unrelated to the sandbox read/write lock) and releases it before returning; it never touches the sandbox's own lock in that path. (For the embedded/sub-sandbox case it explicitly does `unlockAll()`/`restoreLocks()` around the recursive call, which is also lock-neutral by the time it returns.)
- So `processFlyover()` (and therefore `processFlyoverLocked()`'s entry into the `synchronized` blocks) genuinely runs with no sandbox lock held by the current thread, confirming the PR's claim.
- Inside `doProcessFlyover()` (the extracted, unmodified original body), the only sandbox-lock usage is the pre-existing narrow `box.lockRead()`/`unlockRead()` pair wrapping just `applyCondition()`, which is *not* held across the `box.executeView()`/`execute()` (→ `coreLifecycleService.execute()`) calls that can reach `ViewsheetSandbox.lockWrite()`. This is unchanged from the pre-PR code, so the PR doesn't introduce a new read-lock-held-across-write-upgrade hazard either.
- Deadlock-freedom across the two services: both `BaseTableFlyoverService` and `VSChartFlyoverService` build their lock sets from the *same* string key space (`AbstractTableAssembly.getName()` for the source table, `VSAssembly.getAbsoluteName()` for flyover targets) and always acquire them in ascending lexical order via `TreeSet` iteration. Two threads acquiring arbitrary (possibly different) subsets of a shared total order can't form a cycle, which is what makes cross-serialization between chart and table/crosstab flyovers safe without a global lock.

I also checked parameter-order alignment across `processFlyover()` → `processFlyoverLocked()` → `doProcessFlyover()`, since several adjacent parameters share type `String` (`linkUri`, `name`, `conds`) and a transposition wouldn't be caught by the compiler. All three call sites match their target signatures exactly.

No other code paths in the repo reference `getFlyoverLock`/`flyoverLocks` besides these two services, so there's no other producer that could hold the sandbox lock while entering one of these monitors.

## Regressions / behavioral equivalence

`doProcessFlyover()` is a verbatim extraction of the old `processFlyover()` body; the only functional change is that the `views == null || views.length == 0` early return now happens before the (side-effect-free) `tassembly`/`preList` lookups, which is a harmless reordering. I did not find any behavioral difference for the single-source-of-truth case (no cross-service contention) — the new locks are uncontended there and add negligible overhead (`ConcurrentHashMap.computeIfAbsent` + entering unheld monitors).

## Backward compatibility

All modified/added methods (`processFlyover`, `processFlyoverLocked`, `doProcessFlyover`) are `private`; no public/protected API, event, or wire-format changes. No test or other file references these methods directly (verified via search), so nothing outside this file can be broken by the refactor.

## Test coverage

The PR adds no new automated test — verification is a manual repro (attached asset) plus the existing `core` test suite passing. This matches the precedent set by the table-side fix (bug #76532 / PR #5147), which also shipped without a new automated regression test for the same class of race. Race conditions on `synchronized` monitors are inherently hard to pin down deterministically in a unit test, so this is understandable, but flagged as a non-blocking gap below.

## Non-blocking suggestions

1. **No regression test.** Consider a targeted test (e.g., using two threads/latches to force interleaving of a chart flyover and a table flyover on a shared target and asserting the shared target's condition state ends up consistent) if there's an appetite for it, though I recognize this wasn't required for the analogous prior fix either.
2. **Long parameter lists.** `processFlyoverLocked` (12 params) and `doProcessFlyover` (10 params) are hard to scan and reorder-prone (see the same-type-`String` observation above). This mirrors the existing style in `BaseTableFlyoverService`, so it's consistent with the codebase, but a small `FlyoverContext`-style holder (as is already done via `VSChartStateInfo` for some of these fields) could reduce the risk of a future edit silently swapping two `String` args.
3. **Redundant parameter threading.** `chartState` is threaded through `processFlyoverLocked`/`doProcessFlyover` alongside `box` and `chartAssembly`, both of which are also retrievable from `chartState` itself (`chartState.getViewsheetSandbox()`, `chartState.getAssembly()`). Not a bug, just a minor readability/DRY nit inherited from mirroring the table-side structure where the equivalent fields aren't all bundled into one object.
4. It may be worth a short follow-up comment/doc note (e.g. in `claude/angular-chart.md`'s flyover section, which currently doesn't mention the locking) pointing at `ViewsheetSandbox.getFlyoverLock()` now being used from two call sites, to help the next person who touches either service find the other.

PR #5344, opened 2026-09-17
feat(wiz): add General Settings area for admin-chat (Redmine #76712)

ngovus-inetsoft:
Redmine #76712.

## Problem

Enterprise Manager's **Settings > General** page is one combined page backed by a single endpoint (`GeneralSettingsPageController`, gated `EM_COMPONENT`/`settings/general`) aggregating seven sub-models, each with its own service. Of those seven, only **licenseKey** was genuinely covered by admin-chat (by the Licensing area), and **storage** only half-covered — `backup_storage` wraps `DataSpaceSettingsService.doBackup()` alone, leaving the storage-type configuration unreachable.

The remaining five — **localization, mv (server-wide defaults), cache, email, performance** — had *no read and no write*.

## What this adds

`inetsoft.web.admin.ai.general`, exposing six sub-models over the standard triple plus two bare actions:

| Endpoint | Purpose |
|---|---|
| `GET /api/wiz/v1/admin/general/settings` | read one sub-model or all six |
| `POST /api/wiz/v1/admin/general/preview` | resolve a hashed plan, no mutation |
| `POST /api/wiz/v1/admin/general/apply` | apply a reviewed plan, all-or-nothing |
| `POST /api/wiz/v1/admin/general/cache/cleanup` | clear cached data (irreversible) |
| `POST /api/wiz/v1/admin/general/localization/reload` | reload locale resources |

It wraps the sub-services directly rather than `GeneralSettingsPageController`, whose own `@Secured` gate is checkPermission-mediated and therefore inert for a wiz-tagged caller.

**Site administrators only** — deliberately narrower than the EM page, which has a second branch letting a non-site-admin edit `mvSettingsModel` alone on a multi-tenant deployment. An org admin gets 403 for this whole area. Stated in the class javadoc rather than left to be discovered.

## Decisions worth reviewing

- **Email credentials** (`smtpPassword`, `smtpClientSecret`, `smtpAccessToken`, `smtpRefreshToken`) are masked on read and refused on write — exactly the four `mail.*` names `AdminPropertyCatalog.ENCRYPTED_CREDENTIALS` already withholds. That catalog tells an agent, as instruction text, that admin-chat will not write them; a sibling verb that did would make it a false statement (the shape of #76170). `smtpTokenUri` is deliberately *not* secret: read with `getPassword` but written with `setProperty`, so by the catalog's classify-by-the-writer rule it is ordinary.
- **Email writes are refused outright under cloud secrets with `SASL_XOAUTH2`/`GOOGLE_AUTH`.** `setModel` rewrites the three OAuth credentials on every write under those auth types, and under cloud secrets they are read *resolved* but stored *literally* — so any change, even one touching only `smtpHost`, would replace three secrets-manager references with plaintext, irreversibly. Rollback cannot repair it. `NONE`/`SMTP_AUTH` stay writable.
- **Two read-back simulations.** `cache.directory` reproduces the server's `$(sree.home)` rewrite (including its `contains`-then-`substring` quirk — behaving better than the server would resurrect the false-mismatch rollback of #76729), and `localization.locales` reproduces `getModel`'s `Arrays.sort`. Without these, a correct write is reported as a failed verification and rolled back.
- **Storage scope does not imply non-compensable here**, unlike the presentation area. Every writable sub-model's `setModel` takes a complete model object, so writing the captured pre-apply model back is a real inverse. That is why there is no blanket `acknowledgeIrreversibleUpdate` — an acknowledgement every plan can satisfy trains a caller to pass it unread, which would hollow out the one place it means something (`clean_up_cache`).
- **A failed `mv` write reports unknown state, not a clean rollback.** `MVSettingsService.setModel` updates the data-cycle manager before `mvManager`, while the read-back reports `mvManager` — so a throw in between leaves the deployment changed while the read still shows the old value.
- **`localization.locales` removal needs `acknowledgeLocaleRemoval`.** The list is a whole-list replace, so a short list deletes. An earlier revision refused every short list while telling the caller to "send the full list without it and say so in the task description" — which was the very thing it refused, with no override anywhere, making locale removal impossible through this area. The flag makes the documented action reachable while keeping the accident loud.
- **`dataSpace` is read-only** (`DataSpaceSettingsService` has no `setModel`; storage types come from `inetsoft.yaml` at boot) but is still returned by every read, because "which storage backend is this on?" is useful context and a silently missing sub-model reads as a bug.

## Community, not enterprise

Every service behind this lives in `community/core` with no enterprise layer, matching the community-tier shape `AdminClusterController` and `AdminPresentationController` already establish. None of these tools is enterprise-gated.

## Non-area change

`LocalizationSettingsService.reloadLocales()` widens from package-private to `public` so the area can expose it. Behaviour and the existing same-package caller are untouched.

## Testing

**105 tests**, including `GeneralSettingsCoverageTest`, which walks each model's own accessors so a sub-model gaining a field fails the build until someone decides explicitly whether it is settable.

Also **verified end to end against a running server** through the admin-chat MCP tools — not just mocks: login, read with credential masking, preview → apply → read-back → revert, every refusal path, and acknowledged locale removal (`before: [en_US]` → `after: []`, verified).


jshobe-inetsoft (approved):
# Review: PR #5344 — feat(wiz): add General Settings area for admin-chat (Redmine #76712)

## Verdict

**Ready to merge.** I found no blocking issues.

- No correctness, security, or regression defects in the new code.
- The failing `build` check is **not caused by this PR** — see "Build failure analysis" below; it is a transient Maven Central `500 Internal Server Error` resolving an unrelated transitive dependency (`jakarta.inject:jakarta.inject-api:pom:2.0.1`) during the very first reactor module (`inetsoft-core`), before any of this PR's own compilation units are reached. The PR touches no `pom.xml` files.
- `mergeable: MERGEABLE` per `gh pr view`, no merge conflicts.
- Semgrep and Semgrep OSS checks pass; CLA check passes.

Two non-blocking suggestions are listed below for the author's consideration; neither should hold up the merge.

---

## Build failure analysis (explicitly requested)

The `build` job fails at the dependency-resolution stage of the Maven reactor, before compiling any module this PR touches:

```
[ERROR] Failed to execute goal on project inetsoft-core: Could not resolve dependencies for
project com.inetsoft.stylebi:inetsoft-core:jar:1.2.0-SNAPSHOT: Failed to collect dependencies at
jakarta.platform:jakarta.jakartaee-api:jar:10.0.0 -> jakarta.platform:jakarta.jakartaee-web-api:jar:10.0.0
-> jakarta.inject:jakarta.inject-api:jar:2.0.1: Failed to read artifact descriptor for
jakarta.inject:jakarta.inject-api:jar:2.0.1: The following artifacts could not be resolved:
jakarta.inject:jakarta.inject-api:pom:2.0.1 (absent): Could not transfer artifact
jakarta.inject:jakarta.inject-api:pom:2.0.1 from/to central (https://repo.maven.apache.org/maven2):
status code: 500, reason phrase: Internal Server Error (500)
```

- This PR adds no dependencies and changes no `pom.xml` anywhere in the diff (20 files, all `.java`).
- `jakarta.inject-api` is a pre-existing transitive dependency of `jakarta.jakartaee-api`, unrelated to anything in this changeset.
- The failure is a `500` from Maven Central itself (an upstream availability blip), not a checksum/version/resolution logic problem that code changes could trigger.
- This is infrastructure flakiness external to the PR. It should be resolved by re-running the workflow; it is not evidence of a defect in this change.

---

## Scope and context

This PR adds a new admin-chat ("wiz") REST area, `inetsoft.web.admin.ai.general`, that exposes six of the seven sub-models behind Enterprise Manager's "Settings > General" page (`localization`, `mv`, `cache`, `email`, `performance`, `dataSpace`) to the AI admin-chat tool surface, following the same read/preview/apply pattern already established by sibling areas (`cluster`, `presentation`, `licensing`, etc.) in `inetsoft.web.admin.ai`.

There is no `claude/*.md` document in this repository covering the admin-chat/"wiz" tool-surface architecture specifically (`claude/ai-assistant.md` covers a different, unrelated AI-assistant/chart feature). I instead reviewed this PR against its closest sibling area, `inetsoft.web.admin.ai.presentation` (fetched from the PR's actual base branch, `stylebi-wiz-main-rebase`, since that infrastructure does not exist on `main`), to check for architectural consistency, and against the real service/model classes the new code wraps (`LocalizationSettingsService`, `MVSettingsService`, `CacheSettingsService`, `EmailSettingsService`, `PerformanceSettingsService`, `DataSpaceSettingsService`, and their model interfaces) to verify every behavioral claim in the PR description and code comments against the actual implementation.

## Correctness verification performed

I checked the following non-obvious claims made in the code against the real service implementations on the base branch, and all of them hold:

- **`CacheSettingsService.getModel()`** does perform the `contains`-then-`substring(home.length())` rewrite of `directory` that `simulateCacheReadback` reproduces (including the documented "quirk" of substringing from index 0 rather than from the match position) — verified in `CacheSettingsService.java`.
- **`LocalizationSettingsService.getModel()`** does `Arrays.sort` the locale keys while `setModel()` preserves caller order — verified, matches `simulateLocalizationReadback`.
- **`QueryCacheSettings.parseLimit`/`parseTimeout`** do clamp to `[0, Integer.MAX_VALUE]` and `[0, Long.MAX_VALUE]` respectively — verified, matches `requireInRangeCacheSettings`'s bounds and rationale exactly.
- **`EmailSettingsService.setModel`** writes `smtpUser`/`smtpClientId`/OAuth fields conditionally per `SMTPAuthType`, and writes `smtpAuthUri`/`smtpTokenUri`/`smtpOAuthScopes`/`smtpOAuthFlags` only under `SASL_XOAUTH2` (not `GOOGLE_AUTH`) — verified against the actual `if`/`else if` branches; `EMAIL_AUTH_TYPE_FIELDS` in `GeneralChangePlanService` matches exactly.
- **`EmailSettingsService.getModel`** reads `smtpTokenUri` via `SreeEnv.getPassword` but `setModel` writes it via plain `SreeEnv.setProperty` — verified, matches the documented rationale for excluding it from `EMAIL_SECRET_FIELDS`.
- **`MVSettingsService.setModel`** does call `dataCycleManager.setDefaultCycle()` + `.save()` before `mvManager.setDefaultCycle()`, while `getModel()` reports `mvManager`'s value — verified, matches `GeneralSubModel.readBackObservesEveryWrite() == false` for `MV` and the "unknown state" handling path in `GeneralChangesetApplyService`.
- **`EMAIL_SECRET_FIELDS`** (`smtpPassword`, `smtpClientSecret`, `smtpAccessToken`, `smtpRefreshToken`) exactly matches the four `mail.*` names covered by `AdminPropertyCatalog.ENCRYPTED_CREDENTIALS`, as claimed.
- **`LocalizationSettingsService.reloadLocales()`** visibility change (package-private → `public`) is the entire diff to that file; behavior and the existing same-package caller (`GeneralSettingsPageController`) are untouched.
- **`PlanChange`**'s shared record shape (`property, orgId, currentValue, proposedValue, risk, snapshotScope, recognized, description`) is constructed with correct argument order and types.
- Test count claim (105 tests) is accurate: 18 + 54 + 12 + 4 + 4 + 13 = 105 across the six new test classes.
- `GeneralSettingsCoverageTest` genuinely does what the PR claims: it reflects over every model accessor and asserts each field is either settable or in an explicit `EXPECTED_REFUSED` set, so a field added to any of these six models later will fail the build until someone classifies it. This is a good regression guard.
- `GeneralChangesetApplyServiceTest` has explicit tests for the tricky rollback paths described in the PR body: `failedMvWriteReportsUnknownStateRatherThanACleanRollback`, `storageScopeSubModelRollsBackInsteadOfReportingFailure`, `failedRollbackIsReported`, `performanceOutcomeCarriesTheCacheAdvisory`.

I did not find any place where the code's behavior diverges from what its own comments/javadoc claim, which is unusual and a point in the PR's favor — the documentation is exceptionally precise and appears to have been written against the actual implementation rather than aspirationally.

## Clean code / best practices

- Follows the established sibling-area shape closely (`AdminPresentationController` was used as the direct point of comparison): same controller structure, same `requireSiteAdmin`/`AdminAiCallerGuard.requireBearerAuthenticatedRequest()` gate, same `@ExceptionHandler` set for `IllegalArgumentException`/`PlanHashMismatchException`/`TaskTokenMismatchException`, same hash/plan/apply separation.
- `GeneralJson` is a plain (non-enum) utility class specifically to avoid the Java enum static-initialization-order hazard when `GeneralSubModel`'s enum constants call `GeneralJson.fieldNames()` from their own constructors — a real, non-obvious correctness concern, and it's called out explicitly in a comment rather than left implicit.
- Field-name introspection (`GeneralJson.fieldNames`) is derived mechanically from the model interface via reflection rather than hand-maintained, which prevents drift between the model and the catalog as new fields are added — reinforced by `GeneralSettingsCoverageTest`.
- Validation error messages are consistently field-labeled (`changes[N].spec.fieldName: ...`) and explain *why* a value is refused (what the underlying service does with it), which is good for both human operators and the LLM caller consuming these as tool-call errors.

## Error handling and edge cases

- Per-sub-model read failures are isolated (`readErrors` map) so one broken sub-model (e.g., `PerformanceSettingsService.getModel()` throwing on an unparseable `query.runtime.timeout`) doesn't take down the whole `GET` response — correct and matches the stated intent.
- The apply path distinguishes three post-failure states rather than collapsing them into one: (1) confirmed rolled back, (2) confirmed nothing moved so no rollback needed, (3) genuinely unknown state (read failed, or the sub-model's writer touches state the reader can't observe — `MV`). This tri-state handling is more careful than a naive "did the write throw" check and is directly exercised by tests.
- The cloud-secrets clobber guard (`requireNoCloudSecretClobber`) refuses an entire class of otherwise-silent, irreversible data loss (replacing secrets-manager references with plaintext) rather than attempting a partial workaround — a reasonable and conservative choice given rollback genuinely cannot undo it (the pre-apply snapshot already holds resolved plaintext).
- `acknowledgeLocaleRemoval` and `acknowledgeIrreversibleAction` are narrowly scoped (only meaningful/required exactly when their specific hazard applies) rather than a single blanket flag that a caller could learn to pass unconditionally — a deliberate and well-reasoned design choice, explained in the javadoc.

## Security

- Write access is gated by `AdminAiCallerGuard.requireBearerAuthenticatedRequest()` (existing, shared infra, unchanged by this PR) plus `OrganizationManager.isSiteAdmin()`, consistent with sibling areas — no new attack surface introduced by the gating mechanism itself.
- Credential fields (`smtpPassword`, `smtpClientSecret`, `smtpAccessToken`, `smtpRefreshToken`) are masked on every read path (including the "read all six" path) and refused outright on write, matching `AdminPropertyCatalog`'s existing guidance for the same properties on the generic-properties tool — this PR does not create a bypass of that existing guidance.
- `cache.cleanUpCache` (the one genuinely irreversible, destructive action in this area) requires an explicit `acknowledgeIrreversibleAction: true` and is audited on both success and failure.
- Every mutating action is a `POST`, explicitly noted as intentional given the CSRF exemption on `/api/wiz/**` (mutating `GET`s would be a real CSRF risk under that exemption; the EM page's own `GET`-based cache cleanup is not replicated).

## Backward incompatibility / regressions

- `LocalizationSettingsService.reloadLocales()` widening from package-private to `public` is additive and cannot break any existing caller.
- No existing endpoint, model, or service method signature is modified. All new files are additive (`357+54+37+61+36+45+905+119+350+39+160+124+222` lines added, `0` deleted, aside from the 2-line visibility change).
- No test files are modified or deleted; all new tests are additive.

I do not see any way this PR breaks existing tests or existing behavior.

## Non-blocking suggestions

1. **`AdminPropertyCatalog` is not updated with a cross-reference note.** The presentation area's precedent (`AdminPresentationController` for `share.slackUrl`/`googleChatUrl`) has an explicit "the other door to the same value" note in `AdminPropertyCatalog`'s javadoc, cross-referencing the dedicated sub-model surface. This PR does not add an equivalent note for the six `general` sub-models. Since `AdminPropertyCatalog`'s own javadoc states "uncatalogued properties remain settable" via the raw generic-properties tool, a caller could still reach `locale.available`, `mail.smtp.host`, `replet.cache.directory`, `mv.ondemand`, etc. directly through that other tool, bypassing this area's careful validation (whole-list locale checks, auth-type field matching, cache-path simulation, MV cycle validation). This is a pre-existing condition (the raw properties path already allowed writing any `SreeEnv` property before this PR existed) and this PR does not worsen it — if anything it adds a safer alternative path — but adding the cross-reference note would bring this area to full parity with the presentation-area precedent and reduce the chance an LLM caller picks the less-safe tool. Not a regression, and reasonable to defer to a follow-up.
2. **Naming overlap between `general.mv` and the existing `inetsoft.web.admin.ai.mv` area.** Both use "mv" but for materially different things (deployment-wide MV defaults vs. per-viewsheet MV analyze/create/apply). The PR javadoc explicitly calls this out ("these two share only the letters MV"), which mitigates confusion for a human reviewer, but it's worth the author double-checking that the wiz-services broker's tool catalog gives these two areas sufficiently distinct tool names/descriptions so an LLM caller doesn't conflate them at the point of tool selection. This is outside the diff under review (tool registration presumably lives in the broker, not in this PR) and is purely a suggestion for follow-up verification.

## Testing

105 new tests across 6 test classes, all additive, with good coverage of the documented edge cases (MV unknown-state rollback, storage-scope rollback, cloud-secrets clobber refusal, whole-locale-list refusal/acknowledgment, auth-type field mismatch, cache-directory/locale-order read-back simulation, and a reflection-driven "every field is settable or deliberately refused" coverage guard). I did not find gaps in the test suite relative to the logic in the production code.

---

**Approved, not merged.** The `build` check is failing (a transient maven central 500 while resolving an unrelated transitive dependency), so this PR is left unmerged for now. No action needed from you if the cause is external; a later run will merge it once the branch is green.

PR #5345, opened 2026-09-17
Chart palettes: give the measure→colour path three house ramps

FigurePaper:
Fourth slice of the chart palette redesign, after #5188, #5275 and #5311. Base: `epic-74519`. Community-only.

Implements ENGINE §4 of the palette handoff. Design: `docs/superpowers/specs/lookfeel/2026-09-16-chart-ramps-design.md`.

## What a user sees

Binding a **measure** to colour has been untouched by the whole redesign so far. The re-tune gated the *categorical* picker on the chart's mark and re-tuned its series colours; a measure binding still seeded `BluesColorFrame`, a 2010-era ColorBrewer ramp, on every chart regardless of mark.

A modern-marked chart now gets **Teal**, whether it was built in the composer or through the object wizard. Its picker offers three house ramps — **Amber** and **Teal** under Single Hue, **Variance** under Diverging — and stops offering the fifteen ColorBrewer ramps and Heat they succeed. Multi-hue keeps all twelve, because nothing in the house set replaces them, and Custom stays for brand matching.

A classic chart is unchanged in both picker and seed, and may still choose the new ramps: retirements are gated on the mark, additions are not.

## The ramps are derived, not transplanted

The handoff's stops were tuned against the light canvas `#F8F7F4` alone. Amber's low stop `#FCEFE0`, Teal's `#E2F4F2` and Variance's midpoint `#EFEDE7` are all tints of it — each *meant* to recede into the surface, and each the brightest thing on screen against the dark surface `#252428`.

So the stops are re-derived by rule: hue and relative chroma from the source, lightness remapped into a band that clears both surfaces. The rule lives in test sources, because nothing derives a ramp at runtime — the frames ship literal hexes and a drift guard re-derives and compares.

**Variance inverts.** Its midpoint means *at target*, which the source spells as dissolving into the canvas — something one hex cannot do for two surfaces. Lightness goes flat and **chroma carries magnitude** from an achromatic `868686` centre, with hue carrying direction. "At target" reads as a grey mark rather than an absent one. That is a real departure from the source's intent, recorded in the design for the designer to ratify.

## Worth a reviewer's attention

| | |
|---|---|
| **The threading is 34 call sites, not 8** | `ChangeChartTypeProcessor.process()` reaches the seed only via `GraphUtil.fixVisualFrames` / `fixVisualFrames0` — names a grep for `fixVisualFrame(` never matches. Threading only the literal seeds would have left every `process()` call on Blues with the whole suite green. |
| **The wizard held the mark and discarded it** | `VSWizardBindingHandler.getTempChart` took the marked `ChartVSAssembly` off the temporary info and returned only its `ChartInfo`. Every recommender seed read legacy regardless of the host's mark. |
| **Only a transition may re-seed a ramp** | `seedChromeDefaults` runs on creation, Modernize, Revert **and every bookmark restore**. The categorical branch is non-destructive there; a linear re-seed replaces the frame an author picked. `VizContext.transition` scopes it to the two transitions. |
| **The `changed` flag cannot be that guard** | It persists and round-trips, and four sibling panes set it — but `linear-color-pane` never did, and `updateVisualFrameWrapper0` is a no-op none of the 30 linear factories override. Every chart saved before this slice reads unchanged even where the author chose the ramp. |
| **Hide, never remove** | A chart already on a hidden ramp keeps it offered and selected, so nothing repaints on a no-op OK. |

## Testing

Full `core` suite: **5862 tests, 0 failures, 0 errors, 69 skipped.** Scoped picker TL spec: 16. Aesthetic unit specs: 44 across 7 files. `clean install -pl core -am -DskipTests`: BUILD SUCCESS.

Manually verified: a modern chart born on Teal in composer and wizard; both surfaces legible; a classic chart unchanged; the picker's exact row counts per mark; and a modern chart already on Spectral keeping it offered, selected and unrepainted.

The two highest-risk wiring lines were proved by mutation rather than assertion — reverting each turns exactly one test red.

## Deferred

`DensityForm`'s contour default; `VSChartInfo:1048`'s `ChartDcProcessor` construction (latent, traced); a house multi-hue ramp; `GradientColorFrame`'s OKLCH interpolation; designer ratification of the derived stops, especially Variance's inversion. Full list with triggers in the design's §7.


jshobe-inetsoft (approved):
# Review: stylebi PR #5345 — "Chart palettes: give the measure→colour path three house ramps"

Base: `epic-74519` (part of the chart-palette-redesign epic). 67 files changed, +4066/-121.

## Verdict

**Ready to merge.** No blocking issues found. All CI checks are green (Semgrep Scan, Semgrep OSS,
Validation Build, CLA — all `SUCCESS`), the PR is `MERGEABLE` with no conflicts, and there is no
"Claude Code Review / claude-review" check to exclude. Test coverage is unusually thorough
(unit tests for the colour-math derivation, the seeding/threading logic, the wizard path, and
Testing-Library specs for the picker UI), and the change is backward-compatible: every existing
public method signature that needed a new `VizContext` parameter was preserved via an additive
legacy overload rather than modified in place.

A few non-blocking follow-ups are listed below — none of them are regressions introduced by this
PR's own scope, and most are explicitly disclosed by the author in the PR description itself.

---

## What the PR does

Extends the ongoing chart-palette redesign to the *measure→colour* binding (the linear/continuous
colour ramp), which previous slices in the epic (#5188, #5275, #5311) had left on the 2010-era
`BluesColorFrame` regardless of chart mark. This PR:

- Adds three new house ramps: `Amber`/`Teal` (sequential) and `Variance` (diverging), each as a
  `ColorFrame` + `ColorFrameWrapper` + client `ColorModel`, wired into the existing
  `VisualFrameWrapper.createWrapper()` switch and `ColorFrameModelFactory` registry.
- Threads a `VizContext` parameter through ~30 call sites (`GraphUtil.fixVisualFrame(s)`,
  `ChangeChartTypeProcessor`, `ChangeChartDataProcessor`, `ChartDcProcessor`, `ChartDndHandler`,
  the wizard recommender filters, etc.) so a modern-marked chart's *fresh* measure→colour binding
  seeds `TealColorFrame` instead of `BluesColorFrame`, while every un-threaded/legacy caller keeps
  the old Blues default unchanged.
- Adds `VizContext.transition` / `VizContext.ofTransition(mark)`, and gates a *second*,
  narrower reseed (replacing an *existing* linear frame that is the other mark's exact default
  class) to run only on Modernize/Revert — not on ordinary state/bookmark restore, where it would
  otherwise clobber an author's deliberate choice.
- Hides the 15 ColorBrewer single-hue/diverging ramps and Heat from a modern chart's linear-colour
  picker (server: `VSChartPaletteDefaults.hiddenLinearFrames` + new
  `GET /api/composer/chart/hiddenlinearframes`; client: `linear-color-pane.component.ts` filtering),
  with "hide, never remove" semantics — a chart already on a hidden ramp keeps it offered/selected.
- Adds a from-first-principles colour-derivation rule (`ChartRampDerivation`, test-only) that
  re-derives each ramp's stops from fixed source hues, and a `HouseRampTest` that pins the shipped,
  literal hex tables to that rule so a future edit can't silently drift the authored colours.

## Findings

### Non-blocking

1. **`visual-dropdown-pane.component.html` was not updated to pass `[vsId]`/`[assemblyName]` to
   `<linear-color-pane>`.** `color-field-mc.component.html` (the live path) was updated; this
   sibling template (`web/projects/portal/src/app/binding/editor/chart/aesthetic/visual-dropdown-pane.component.html`)
   still instantiates `linear-color-pane` with neither input. Without them,
   `LinearColorPane.getHiddenFrames()` calls the endpoint with empty params, and
   `VSChartBindingController.pickerContext()` falls back to `VizContext.ofGate()` — the *org-wide*
   modern-chrome gate rather than the specific chart's own mark. I confirmed via
   `grep -rln "visual-dropdown-pane"` that this component's selector (`visual-dropdown-pane`) is not
   referenced anywhere else in the codebase, i.e. it currently appears to be dead/unwired code, so
   this has no live user-facing effect today — but it's a latent inconsistency if that component is
   ever revived.

2. **`VSChartInfo.update()`'s date-comparison runtime path (~line 1048) still constructs
   `ChartDcProcessor` via the legacy 2-arg constructor**, i.e. `VizContext.LEGACY`, because
   `VSChartInfo` has no reach back to its owning `ChartVSAssemblyInfo`'s mark. In principle this
   means a modern-marked, date-comparison-enabled chart whose colour aesthetic is freshly bound to a
   measure during a runtime refresh could seed `BluesColorFrame` instead of `TealColorFrame`,
   diverging from the composer-binding paths that do thread `VizContext.of(...)`. This is explicitly
   disclosed by the author in the PR description's "Deferred" section ("`VSChartInfo:1048`'s
   `ChartDcProcessor` construction (latent, traced)"), so it's a known, accepted limitation rather
   than a silent gap — but worth flagging since date-comparison charts are a mainstream feature, not
   a corner case like contour maps.

3. **`ChangeChartProcessor.fixColorField()`** (`core/src/main/java/inetsoft/report/internal/graph/ChangeChartProcessor.java:215`)
   still unconditionally does `newFrame = new BluesColorFrame();` for a measure colour binding, with
   no `VizContext` parameter added. I confirmed via `git grep -n "fixColorField("` that this method
   has **no callers anywhere** in `core/src/main/java`, so it's dead code and poses no live-behavior
   risk. Worth a follow-up cleanup/removal, but not blocking.

4. **Design tradeoff worth recording (not a bug):** `ChartVSAssemblyInfo.isOtherMarkSeededLinearFrame`
   cannot distinguish "still holding the auto-seeded default" from "the author deliberately reselected
   the exact same ramp class the default would have used." On Revert, a modern chart where the author
   *deliberately* re-picked `TealColorFrame` (now a legitimate, offered choice on a modern chart) will
   be silently replaced with `BluesColorFrame`, same as an untouched default would be — there is no way
   to tell the two apart, and the code's own comments acknowledge this explicitly ("outside a
   transition there is nothing to distinguish the two and the caller must not guess" — implying that,
   inside a transition, it does guess). This mirrors pre-existing behavior for the categorical
   palette re-seed (which has always unconditionally rewritten default colours on every
   `seedChromeDefaults` call), so it isn't a new class of risk introduced by this PR, and the test
   suite explicitly covers and documents the adjacent "deliberate non-default ramp survives Revert"
   case (`ChartVSAssemblyInfoSeedTest.anAuthorsSpectralRampSurvivesRevert`). Flagging only so it's on
   record as a known edge case, not because it needs to block this PR.

### Verified as sound (no issues found)

- **Backward compatibility:** every pre-existing public method whose behavior needed to vary by
  `VizContext` (`GraphUtil.fixVisualFrames[0]`, `fixVisualFrame`, `ChangeChartTypeProcessor`,
  `ChangeChartDataProcessor`, `ChartDcProcessor`, `ChartDndHandler.dropToAesthetic/dropToChartView`,
  `ChartCombinationUtil.getChartInfos`) keeps its old signature as a "legacy" overload that resolves
  to `VizContext.LEGACY` — i.e. any caller not touched by this PR (including enterprise code or
  scripts outside this diff) gets byte-for-byte the same `BluesColorFrame` behavior as before.
- **Persistence/round-trip:** the three new frame classes are registered in
  `VisualFrameWrapper.createWrapper()`'s switch using the correct simple-class-name key convention
  (matching existing entries like `"BipolarColorFrame"`), and `HouseRampTest.eachFrameSurvivesTheWrapperRoundTrip`
  exercises XML serialize→parse for all three, so old viewsheets/assets are unaffected and new ones
  round-trip correctly. (There is also an existing `Class.forName` fallback in `createWrapper` if a
  switch case is ever missing.)
- **Hide-never-remove correctness:** both the Java (`VSChartPaletteDefaults.hiddenLinearFrames`) and
  TypeScript (`LinearColorPane.visible()`/`heatVisible`) sides keep a currently-selected-but-hidden
  ramp in the offered list and pre-selected; this is exercised by dedicated TL specs (`"should keep a
  hidden ramp offered and selected..."`, `"...Heat row rendered and checked..."`).
- **New REST endpoint** (`GET /api/composer/chart/hiddenlinearframes`) exactly mirrors the existing
  `getColorPalettes` endpoint's parameter/annotation shape and reuses the same `pickerContext()`
  helper (with its existing graceful fallback to `VizContext.ofGate()` on missing params or
  resolution failure), so it introduces no new attack surface or security-relevant divergence from
  an already-reviewed pattern. Semgrep (both scans) passed.
- **Node-color / RelationChartInfo aesthetics**: `ChartVSAssemblyInfo.seedColorPalette`'s reseed
  loop iterates `VSChartInfo.getAestheticRefs(runtime)`, which (like the pre-existing categorical
  branch it sits beside) does not cover `RelationChartInfo.getNodeColorField()`. This is a
  pre-existing gap in the categorical seeding too, not something this PR regresses.
- **Test coverage** is unusually deep for a UI/default-value change: colour-math derivation tests
  (`ChartRampDerivationTest`, `HouseRampTest`) that pin authored hex literals to a from-first-principles
  rule; seeding-threading tests at three layers (`SeededLinearFrameTest`,
  `ChartVSAssemblyInfoSeedTest`, `VizModernizeUtilTest`, `WizardSeededLinearFrameTest`) that
  specifically exercise the "34 call sites, not 8" and "only a transition may reseed" invariants the
  PR description calls out as the highest-risk parts of the change; and 7 new/expanded TL spec groups
  for the picker UI. The PR description states these were validated by reverting each of the two
  highest-risk wiring lines and confirming exactly one test goes red — a good sign of real (not just
  incidental) coverage.
- No existing test asserts a fixed count of linear/categorical ramps that this change would break,
  and no other spec file references the modified `singleHueModels`/`multiHueModels`/`divergingModels`
  arrays or the `getColorPalettes`/`hiddenPaletteNames` categorical machinery this PR leaves alone.

## Scope note

Reviewed via `gh pr diff 5345` against the local checkout, cross-checked several points against the
actual PR branch tip (`origin/feature-chart-ramps`, base `origin/epic-74519`) since the local
`community` checkout was on `main`. Read `claude/chart.md` and `claude/angular-chart.md` (aesthetic
frame / `VisualFrame` taxonomy and `AestheticPane` sections) and `claude/chart-date-comparison.md`
(`ChartDcProcessor` / `VSChartInfo` wiring) for architectural context, per the repo's CLAUDE.md
instruction to consult the relevant `claude/*.md` references before reviewing.

PR #5346, opened 2026-09-17
fix(76668): keep a $(ComponentName) shelf field through a chart type change

ngovus-inetsoft:
Follow-up from the #76641 round-1 review (PR #5244). That review flagged an **unconfirmed** sibling of the defect it fixed, and asked for the trace to settle it either way.

**It is a live defect.** The RT-ref escape hatch does not apply.

## Trace

`WizVsService.collectFlatBinding`'s "prefers RT refs" comment describes only `collectChartSlots` — the separate `slots` map. The `dimensions`/`measures` that become the field list come from `info.getBindingRefs(false)` (`false` = **design** refs), and `WizFieldInfoFactory` sets `field` from `dim.getGroupColumnValue()` / `agg.getColumnValue()`, both of which return `DynamicValue.getDValue()`. `collectCrosstabFlatBinding` reads design headers too. So `field` arrives as the literal `"$(RadioButton2)"`.

The impact is worse than the issue anticipated. `changeType` calls the **unfiltered** `deriveFieldConfigs(targetAssembly)` and forwards the result as an explicit list, which makes `sourceAssembly` null — so the tolerant `derivedFieldConfigsWithin` guard never runs and the list goes straight into the **throwing** `selectBindColumns`:

```
Unknown field(s) in fieldConfigs: [$(RadioButton2)]. Available worksheet columns: [...]
```

A 400 on every type change of a chart carrying a #76641 dynamic reference — not the soft "bind every visible column" fallback.

## Why a bypass alone wasn't enough

`selectBindColumns` also *filters* the worksheet columns by config key. A `$(...)` key matches no column, so adding only the `isDynamicVariableReference` skip would have traded the hard failure for **silent field loss** — the other bad outcome the issue named. The downstream pipeline (`buildEntryFromColumn` → `updateTemporaryFields`) materializes fields from `ColumnRef`s only, so the field needs a real column to survive, and its dynamic binding restored afterward.

## Changes

- **Resolve** — `deriveFieldConfigs` rewrites a `$(...)` field to the column it currently resolves to, reusing the existing `VSDimensionRef.getName()` / `VSAggregateRef.getName()` accessors, so it survives column selection.
- **Restore** — `restoreChartDynamicColumnValues` puts the literal back on the rebuilt refs so the chart keeps **following** the component rather than freezing to a static column. Also wired into the crosstab path. Runs **strictly after** `applyFieldConfigs` and `applyAxisTitlesFromFieldConfigs`, both of which look their config up by the ref's current column value — there is a dedicated ordering regression test.
- **Tolerant gates** — `isDynamicVariableReference` skips in both `derivedFieldConfigsWithin` and `selectBindColumns` for residue a cold RVS cannot resolve, plus a guard so an all-dynamic list falls back to every visible column instead of binding nothing.

The helper duplicates `VSUtil.isVariableValue` exactly as `BindableColumns.isDynamicVariableReference` does, and for the same documented reason (`VSUtil`'s static initializer reaches for a live Spring `ApplicationContext`). The stashed literal rides on a `@JsonIgnore` field, so the wiz API's field echo is unchanged.

## Testing

`Tests run: 328, Failures: 0` across `WizAutoBindingService*`, `WizVsService*` and the pre-existing `BindableColumnsTest` (32). 12 are new in `WizAutoBindingServiceDynamicColumnRefTest`, covering: the reported 400, a genuinely unknown field still throwing alongside a dynamic one, the all-dynamic fallback, both existence gates, resolution (including the cold-RVS no-op and the untouched ordinary field), and the restore on chart and crosstab plus its ordering.

**Not verified end-to-end** — the evidence is code-trace plus unit tests; no live `/viewsheet/changeType` run.

## Noted, not fixed

`BindableColumns` compares column names case-**insensitively** (`TreeSet<>(String.CASE_INSENSITIVE_ORDER)`) while these two `WizAutoBindingService` methods use case-sensitive sets. Pre-existing inconsistency, left alone per minimal scope.

Base is `stylebi-wiz-main-rebase` rather than `main`: the wiz subsystem does not exist on `main`.


jshobe-inetsoft (approved):
# Review: stylebi PR #5346 — fix(76668): keep a $(ComponentName) shelf field through a chart type change

Base branch: `stylebi-wiz-main-rebase` (the wiz subsystem does not exist on `main`, per PR description). Head: `bug-76668`. Files touched: `FieldInfo.java` (+21), `WizAutoBindingService.java` (+246/-3), new test `WizAutoBindingServiceDynamicColumnRefTest.java` (+331).

## Verdict

**Ready to merge.** No blocking issues found. The fix is correct, well-targeted, and verified — I rebuilt the PR head in an isolated worktree, compiled the `core` module, and ran the full `WizAutoBindingService*` / `WizVsService*` / `BindableColumnsTest` suite: **328 tests, 0 failures**, matching the PR description exactly. The 12 new tests pass and meaningfully cover the reported crash, the tolerant fallback, and the restore/ordering behavior.

Two non-blocking gaps are worth a follow-up ticket (see "Non-blocking findings" below) — neither causes a crash, data loss, or regression versus current `main`/base-branch behavior; they narrow, rather than eliminate, the class of bug the PR title targets.

CI: all checks green (Semgrep Scan, Semgrep OSS, build, license/cla). PR is `MERGEABLE`, no conflicts.

---

## Summary of the change

Chart/crosstab shelf fields can be dynamically bound to a Form component's live value (`"$(RadioButton2)"`, from #76641). `WizVsService#collectFlatBinding` reads **design** refs, whose value for such a field is the literal string, not a column name. Previously:

- `changeType()`'s fallback path derives an authoritative field list via `deriveFieldConfigs(targetAssembly)` and forwards it as an explicit list, which sends it straight into the throwing `selectBindColumns` → a 400 on every type change of a chart carrying a dynamic reference.
- A bare "tolerate the literal" fix would have caused the literal to match no worksheet column and be silently dropped from the rebuild instead.

The PR fixes this by: (1) resolving a `$(...)` design value to the column it currently resolves to (`resolveDynamicFieldConfigs`/`dynamicColumnResolutions`) so the field survives column selection, (2) stashing the original literal on a new `@JsonIgnore` `FieldInfo.dynamicColumnValue` field, (3) restoring the literal onto the rebuilt ref after all lookups that key off the resolved column name have run (`restoreChartDynamicColumnValues` / `restoreDynamicColumnValues`, called for both chart and crosstab paths), and (4) adding tolerant gates (`isDynamicVariableReference`) in `derivedFieldConfigsWithin` and `selectBindColumns` so a still-unresolved literal (cold RVS) doesn't fail or discard the whole list, with an all-dynamic fallback to "bind every visible column."

## Correctness verification

I traced the claims against the actual source rather than taking the PR description at face value:

- `VSDimensionRef.getGroupColumnValue()` returns the raw design value (`groupValue.getDValue()`); `getName()` returns the runtime value only when it's a `String`, else falls back to the design value. This is exactly the "prefers RT refs only in `getName()`" mechanism the PR relies on. Confirmed in `core/src/main/java/inetsoft/uql/viewsheet/VSDimensionRef.java:443-529`.
- `VSAggregateRef.getName()` returns `rval + ""` when `ref`/`rval` are null — i.e., literally the string `"null"` — which is exactly what `addDynamicResolution`'s `!"null".equals(runtimeName)` guard defends against (`VSAggregateRef.java:170-185`). This is a subtle, correctly-reasoned guard, not a stray defensive check.
- `applyFieldConfig`, `applyAxisTitle`, and `applyAggregateFormulasTo` all key their `configMap` lookups off `WizardRecommenderUtil.getChartRefFieldName(ref)` / `dim.getGroupColumnValue()` / `agg.getColumnValue()` — i.e., the ref's *current* (resolved) column value at the time they run. `restoreChartDynamicColumnValues`/`restoreDynamicColumnValues` are called strictly after each of these in both the chart branch (`autoBindingInternal` line ~349) and the crosstab branch (end of `applyCrosstabAggregateFormulas`), so the ordering claim in the javadocs is accurate and matches the dedicated ordering regression test.
- `WizFieldInfoFactory.baseDimensionFieldInfo`/`createMeasureFieldInfo`/`createChartMeasureFieldInfo` do set `field` from `dim.getGroupColumnValue()` / `agg.getColumnValue()` (design values), confirming the root-cause trace in the PR description.
- `isDynamicVariableReference`'s `"$(" ... ")"` check is an exact duplicate of `VSUtil.isVariableValue` and `BindableColumns.isDynamicVariableReference`, for the documented reason (`VSUtil`'s static initializer needs a live Spring `ApplicationContext`). This mirrors an existing, already-accepted pattern in the same package.
- `collectCrosstabFlatBinding` reads exactly `getDesignRowHeaders()`/`getDesignColHeaders()`/`getDesignAggregates()`, matching `bindingRefsOf`'s crosstab branch one-to-one — no gap there.

Build/test verification performed:
- `./mvnw -o -pl core -am test-compile -DskipTests` on the PR head — clean compile, no warnings.
- `./mvnw -o -pl core -Dtest=WizAutoBindingServiceDynamicColumnRefTest test` — 12/12 pass.
- `./mvnw -o -pl core -Dtest="WizAutoBindingService*,WizVsService*,BindableColumnsTest" test` — 328/328 pass, 0 failures/errors, matching the PR's own reported run.

## Non-blocking findings

**1. `bindingRefsOf` does not cover every ref `WizVsService#collectChartFlatBinding` walks.**

`bindingRefsOf` (new, in `WizAutoBindingService.java`) walks `info.getBindingRefs(false)` plus color/shape/size/text aesthetics plus `getAggregateAestheticRefs(false)`. But `collectChartFlatBinding` (which is what actually produces the `dimensions`/`measures` list `deriveFieldConfigs` resolves) additionally walks:
- the chart's `pathField`,
- `CandleVSChartInfo`'s `highField`/`lowField`/`closeField`/`openField` (confirmed: `CandleVSChartInfo` does not override `getBindingRefs`, and these fields are stored separately, not folded into X/Y — see `CandleVSChartInfo.java`),
- `RelationVSChartInfo`'s `getNodeColorField()`/`getNodeSizeField()` aesthetics (source/target *are* covered, since `RelationVSChartInfo.getBindingRefs` folds them in — but node aesthetics are not).

The `bindingRefsOf` javadoc acknowledges refs `collectFlatBinding` reads but it misses will "simply not be resolved, and the tolerant gates would let it through unbound rather than fail" — true as far as it goes, but the practical consequence is stronger than that phrasing suggests: an unresolved dynamic literal on one of these three slots doesn't get folded into `resolved`, so it is never rewritten to a real column name, so it never survives `selectBindColumns`'s `configMap.containsKey(c.getDisplayName())` filter (unless it's the *only* config, triggering the all-dynamic fallback). In a mixed binding (e.g. an ordinary X dimension plus a dynamically-bound Candle high field), the dynamic Candle field would be silently dropped from the rebuild — precisely the "other bad outcome" the PR's own description says it set out to avoid, just for a narrower set of slots than the reported case. No crash results, and GanttVSChartInfo/RelationVSChartInfo/VSMapInfo source-target/geo refs are all correctly covered via their `getBindingRefs` overrides — only Candle H/L/C/O, `pathField`, and Relation node aesthetics are missed.

**2. The wizard's temp chart never gets its dynamic literal restored, which permanently defeats the `changeType` fast path for a chart/crosstab carrying a dynamic reference.**

`restoreChartDynamicColumnValues`/`restoreDynamicColumnValues` are called on the **rendered** assembly (`chartAsm`/`primaryAssembly`, and the crosstab info) but never on `tempInfo.getTempChart()`. `applyFieldConfigsToTempChart(rvs, configMap)` (called right after, at `autoBindingInternal` line ~374, and again via `recordFieldConfigsOnTempChart` in the `changeType` fallback) applies ranking/sort/format settings to the temp chart's refs but does not touch their column values — those refs were built from `entries`/`bindColumns` (real worksheet `ColumnRef`s), so they permanently carry the *resolved* column name (e.g. `"Households"`), never the `"$(RadioButton2)"` literal.

`tempChartDescribes()` (used by `changeType` to decide whether the cached recommendation model — the fast path — is still valid) compares `BindingFieldSettings.columnKeys(tempChart)` against `BindingFieldSettings.columnKeys(target)`, both keyed on raw `getGroupColumnValue()`/`getColumnValue()`. Since the target (the real, saved assembly) *does* have its literal restored while the temp chart does not, these key sets will permanently disagree for any chart/crosstab bound to a dynamic reference — `staleForTarget` will evaluate `true` on every subsequent `changeType` call, forcing the expensive full-rebuild-via-`autoBindingInternal` fallback path indefinitely instead of ever reusing the cached-model fast path.

This is not a data-loss or correctness bug — `deriveFieldConfigs` always reads from the real target assembly (which is correct), so the field, its dynamic binding, and its ranking/sort/date-level settings are recomputed correctly on every call via the authoritative `fieldConfigs` list. It is a silent performance regression specific to the exact feature this PR fixes (every type change on a chart with a Form-component-bound field becomes a full recommender re-run instead of the lean reslot), and it isn't covered by any of the 12 new tests or mentioned in the PR's "Noted, not fixed" section. Likely fix: also call `restoreChartDynamicColumnValues`/`restoreDynamicColumnValues` (or an equivalent) on the temp chart's refs in `applyFieldConfigsToTempChart`.

**3. Minor / style, no action needed:**
- `isDynamicVariableReference` is now duplicated a third time (`VSUtil.isVariableValue`, `BindableColumns.isDynamicVariableReference`, and this one). Each duplication is individually justified (avoiding `VSUtil`'s Spring-context-touching static initializer in a plain unit test), and it follows an existing precedent in the same package, so this is not worth blocking on — but a shared, dependency-free helper (e.g. a small package-private utility class) would remove the risk of the three copies drifting apart over time.
- `resolveDynamicFieldConfigs`'s `configMap` is still built via `Collectors.toMap(SimpleFieldInfo::getField, f -> f, (a, b) -> a)` (pre-existing code, unchanged by this PR). If two distinct dynamic references happened to resolve to the same runtime column simultaneously, one config (and its `dynamicColumnValue`) would be silently dropped from the map, and `restoreDynamicColumnValues` could restore the wrong literal onto one of the refs. This is an edge case of an existing, pre-PR limitation of the single-column-keyed `configMap` (it already collapses any two refs bound to the same real column name), not something this PR introduces or meaningfully worsens.

## Other review areas

- **Backward compatibility:** `FieldInfo.dynamicColumnValue` is `@JsonIgnore` on both accessors, so it is never serialized to or deserialized from the wiz API — no change to the wire contract, confirmed correct and intentional.
- **Security:** No injection or trust-boundary concerns; the `"$(" ... ")"` check is a pure string predicate over server-side state (chart ref values), not user-supplied text evaluated as code.
- **Error handling:** The `resolveDynamicFieldConfigs`/`dynamicColumnResolutions` path is appropriately best-effort (a cold-RVS/never-resolved literal is left alone rather than thrown on), consistent with the rest of the file's tolerant-gate style.
- **Readability/maintainability:** Javadocs are unusually extensive but consistent with the file's existing documentation density (see `applyAxisTitlesFromFieldConfigs`, `BindingFieldSettings`, etc.), and they correctly describe the mechanism (verified above), so they add real value here rather than being noise.
- **Test breakage:** None. Full targeted suite (328 tests across `WizAutoBindingService*`, `WizVsService*`, `BindableColumnsTest`) passes on the PR head with no modifications.

PR #5347, opened 2026-09-17
Bug #76733 (WBT-012): apply the column limit on the agent worksheet preview path

ngovus-inetsoft:
Redmine [#76733](https://issues.inetsoft.com/issues/76733), finding `WBT-012` (filed jointly with `WBT-011` — see below, that half needs no code here).

## The bug

`preview_worksheet_data` returned an unboundedly wide row — 300+ columns observed live off a GitHub *Repository Events* endpoint, reconfirmed against a second 262-column table — with `warnings: []`, no cap and no indication anything unusual had happened.

## Root cause

`AssetQuerySandbox.executeQuery` only wraps a result in `getColumnLimitTableLens` when `table.isLiveData()` is true:

```java
if(table.isLiveData()) {
   base = getColumnLimitTableLens(base, table instanceof UnpivotTableAssembly);
}
```

But `isLiveData()` is a **display-mode** flag. `WorksheetEditService.setTableMode` maps `"detail"` → `liveData=true` and `"full"` → `liveData=false`, and `TableAssemblyInfo.live` defaults to `false` — so every assembly `add_table` creates reads back as `mode:"full"` and is never capped. `add_join`'s result happens to land in `"detail"`, which is the only reason `WBT-002`'s JOIN repro got the warning at all. Confirmed directly off the live worksheet model: five plain `add_table`-built tables all `"full"`, the joined one `"detail"`.

`WorksheetPreviewService`, meanwhile, deliberately executes in `RUNTIME_MODE` *regardless* of that flag (see its class javadoc on why `LIVE_MODE` must not be used there). So the agent preview path already ignores `isLiveData()` for execution while still inheriting it as a gate on the cap. That mismatch is the bug, and it is specific to this path.

## The fix

`WorksheetPreviewService.preview` re-applies the cap itself, mirroring `getColumnLimitTableLens` exactly (`Util.getOrganizationMaxColumn` + `ColumnMapFilter` + `Util.getColumnLimitMessage`, which the existing `captureWarnings()` already relays into `PreviewResult.warnings`).

Two alternatives were considered and rejected:

- **Dropping the `isLiveData()` gate in `AssetQuerySandbox`** — one-line diff, but it changes behavior for every sandbox consumer, including the interactive composer's full/metadata modes. Far wider blast radius than this bug warrants.
- **Defaulting `add_table`'s assembly to `"detail"`/`liveData=true`** — changes the assembly's persisted display mode and whatever else keys off it.

Notes on the implementation:

- **Idempotent.** A `mode:"detail"` table is already capped by the sandbox, so `getColCount()` is within the limit by the time `preview` sees it — the block is a no-op and the warning is never raised twice.
- **Inside the existing `try`**, not before it, so a lens whose `getColCount()` throws still surfaces as the `PairingException` the `catch` produces rather than a bare crash (an earlier placement outside it broke `throwsPairingExceptionWhenGetColCountThrows`).
- **The sandbox's cached lens is untouched** — the wrap happens outside the cache, so the cache keeps the full-width lens.

## Tests

Three added to `WorksheetPreviewServiceTest`: cap-and-warn over the limit, no-op at/under it, and no double-warn when the sandbox already capped. These are reachable precisely because that class mocks `AssetQuerySandbox` — the boundary the existing `WBT-002` test notes it cannot reach past.

`./mvnw test -pl core -Dtest=WorksheetPreviewServiceTest` → 22 tests, 0 failures.

## On `WBT-011`, the other half of #76733

No code here, deliberately. Traced into source: the finding's own guess ("presumably a different controller method for the `endpoint` form") is wrong — `addTabularTable` (endpoint) and `addQueryParamsTable` (queryParams/ServerFile) have byte-identical add-before-probe lambdas, and **#5268 already reorders both** (its own message: *"addTabularTable's REST-JSON (endpoint/suffix) form has the identical defect shape and gets the identical fix"*), with a matching `addTabularTableNoColumnsDoesNotLeaveResidualAssembly` test. It reproduced live only because #5268 is still open. Landing #5268 is the fix.

The compounding missing-`note` half could not be reproduced against current code — both halves of `WBT-009`'s fix are present and do cover the endpoint form, and the live observation is dated the same day those PRs merged. Flagged for one live re-verification instead of speculative code.


jshobe-inetsoft (approved):
# Review: stylebi PR #5347 — Bug #76733 (WBT-012): apply the column limit on the agent worksheet preview path

## Verdict

**Ready to merge.** No blocking issues found. Non-blocking suggestions are listed below.

- All CI checks are green (Semgrep, Semgrep OSS, Validation Build, CLA).
- I independently rebuilt `WorksheetPreviewServiceTest` on the PR branch (`./mvnw -o test -pl core -Dtest=WorksheetPreviewServiceTest`): **22/22 tests pass**, matching the PR description's own reported run.

---

## Summary of the change

`WorksheetPreviewService.preview()` (the REST/agent worksheet-preview endpoint, `core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetPreviewService.java`) returned every column of the executed `TableLens`, unbounded. `AssetQuerySandbox.executeQuery` only applies its own column cap (`getColumnLimitTableLens`) when `TableAssembly.isLiveData()` is `true` — a display-mode flag (`mode:"detail"`) that defaults to `false` and is `false` for the common `add_table`-created assembly (`mode:"full"`). Because this endpoint always executes in `RUNTIME_MODE` regardless of that flag, `"full"`-mode tables reached the caller with no cap and no warning.

The fix re-applies the identical capping logic (`Util.getOrganizationMaxColumn()` + identity `int[] map` + `ColumnMapFilter` wrap + `Util.getColumnLimitMessage()` via `CoreTool.addUserMessage`) inside `preview()`, placed inside the existing `try` block (before the header/row-extraction code), plus three new unit tests.

## Verification performed

- Read the full diff (`gh pr diff 5347`).
- Read `WorksheetPreviewService.java` and `WorksheetPreviewServiceTest.java` in full at the PR head commit (checked out via `git fetch origin pull/5347/head`).
- Compared the new block against `AssetQuerySandbox.getColumnLimitTableLens` (lines ~1045-1060) and confirmed the logic is an exact behavioral mirror (same threshold check, same identity-map construction, same message helper, same filter class). The unused `unpivot` parameter on the original method is dead code there too, so nothing is lost by not carrying it over.
- Confirmed `Tool.addUserMessage`/`CoreTool.addUserMessage` are the same static method (`Tool extends CoreTool`), so the message level (`ConfirmException.INFO`) and dedup semantics (`CoreTool.existUserMessage` dedupes identical message+assembly text) are unchanged from the pre-existing sandbox path.
- Confirmed `ColumnMapFilter.getObject`/`moreRows`/`getColCount` delegate to the wrapped lens with the identity map, so header-row reading (`getObject(0, col)`) and pagination (`moreRows(row)`) in the rest of `preview()` are unaffected by the wrap.
- Traced `TableAssemblyInfo.live` defaulting to `false` and `TableModeService` toggling it via `setLiveData(true/false)` for `"detail"`/`"full"` modes — confirms the PR's stated root cause (the PR description names `WorksheetEditService.setTableMode`; the actual class is `TableModeService`, a cosmetic inaccuracy in the PR prose only, not in code).
- Confirmed the `throwsPairingExceptionWhenGetColCountThrows` test still passes with the new code placed earlier in the same `try`: the new block's own `lens.getColCount()` call throws first now, still caught by the same `catch(RuntimeException e)` and wrapped into the same `PairingException` message — no behavior change to that test's assertion.
- Rebuilt and ran the full test class in an isolated git worktree on the PR branch: **22 passed, 0 failed, 0 errors**.
- Confirmed the identity-map-plus-`ColumnMapFilter` idiom this PR duplicates is already repeated verbatim across a dozen+ call sites in this codebase (`AssetQuerySandbox`, `WorksheetOpenAssetService`, `PostProcessor`, `TableVSAQuery`, etc.), so the duplication here is consistent with existing conventions rather than a new pattern.
- Checked `claude/worksheet-query.md` (the relevant feature reference doc): step 9 of `AssetQuerySandbox.getTableLens` is listed as "Apply normalizer, column limit, text size limit" — consistent with this PR's understanding of where the cap normally happens and why the agent preview path (which bypasses that flow's gate) needed its own application of it. No dedicated `claude/*.md` doc exists yet for the `inetsoft.web.wiz` agent-pairing package itself.

## Findings

### Blocking
None.

### Non-blocking suggestions

1. **Logic duplication.** The cap-and-wrap block (build `Util.getOrganizationMaxColumn()`, identity `int[]`, `ColumnMapFilter`, `addUserMessage`) is now duplicated between `AssetQuerySandbox.getColumnLimitTableLens` and `WorksheetPreviewService.preview`. A small shared helper (e.g. a package-visible static on `Util`, or exposing `AssetQuerySandbox`'s private method) would remove the duplication and guarantee the two stay in sync if the capping algorithm ever changes. Given this exact duplication pattern is already pervasive elsewhere in the codebase (a dozen+ sites), this is a stylistic nit, not something this PR introduced as a new problem, and not worth blocking on.
2. **PR description names the wrong class.** The write-up attributes the `"detail"`/`"full"` → `liveData` mapping to `WorksheetEditService.setTableMode`; the method that actually does this is `TableModeService` (`core/src/main/java/inetsoft/web/composer/ws/TableModeService.java`). Purely descriptive — doesn't affect the code — but worth a one-line correction for anyone using the PR description as a reference later.
3. **No test at the exact boundary `max + 1`.** The two "over the limit" tests use `max + 5` and `max` (at the limit); an explicit `max + 1` case is implicitly covered by the `> maxCols` condition being a plain off-by-one comparison shared with the already-tested `AssetQuerySandbox` logic, so this is a very low-value gap, not worth requesting a change for.

## Coverage against the requested review dimensions

- **Completeness/correctness:** Fix directly addresses the described root cause; logic verified to exactly mirror the existing, already-trusted sandbox implementation; tests pass.
- **Clean code / best practices:** Mostly good; only nit is the duplicated algorithm (see above), which matches existing codebase style.
- **Error handling / edge cases:** The cap check is correctly placed inside the existing `try`/`catch(RuntimeException)` so a `getColCount()` failure still surfaces as `PairingException` rather than crashing; verified against the existing `throwsPairingExceptionWhenGetColCountThrows` test, which still passes.
- **Readability/maintainability:** Comments are thorough and explain *why*, including the reasoning for two rejected alternative fixes (documented in the PR description) and the two-source-of-truth relationship in the updated `captureWarnings()` javadoc.
- **Security:** No new attack surface; if anything this closes an unbounded-data-return gap (data volume, not injection) on an agent-facing endpoint.
- **Regressions:** None found. The cached lens in `AssetQuerySandbox` (`tmap`) is untouched — the wrap happens only on the copy already returned to this caller, so other consumers of the cache are unaffected. Confirmed idempotent for already-`"detail"`-mode tables via the third new test and via code inspection of `CoreTool.existUserMessage`'s dedup behavior.
- **Backward incompatibility:** The response shape for over-limit `"full"`-mode tables changes (now capped, now carries a warning) — this is the intended fix, not an unintended break, and mirrors behavior already present for `"detail"`-mode tables and throughout the rest of the product.
- **Broken tests:** None. Full `WorksheetPreviewServiceTest` suite (22 tests, including all pre-existing ones) passes on the PR branch, verified independently in this review.

PR #5349, opened 2026-09-17
fix(76757): let a constrained String property take an "=expression" dynamic value

ngovus-inetsoft:
## Problem

`PropertyPath`'s closed-value gate already lets a `"$(ComponentName)"` variable through unresolved for `visible` (#76641) — the underlying field is a `DynamicValue` that `ViewsheetSandbox.executeDynamicValue` resolves at render time, so the gate must not judge it.

But the predicate tested only `VSUtil.isVariableValue`'s half of the convention. The other half — a leading `=`, `VSUtil.isScriptValue` — was refused with `'visible' accepts only [false, hide, hide on print and export, show, true]`.

That blocked the only way to express a *derived* property. `"$(Name)"` substitutes one component's value verbatim; it cannot branch on it. The motivating case: a gauge showing a customer count, and a chart whose title should read `">20"` or `"<20"` depending on that count.

`DynamicValue` is one mechanism with two halves, discriminated by prefix — and StyleBI's own Composer says so, offering **Variable** and **Expression** side by side on this very field (`dynamic-combo-box.component.ts`'s `updateType` classifies by these same two prefixes).

## Change

`isDynamicVariableReference` → `isDynamicValue`, mirroring `VSUtil.isDynamicValue` (both halves) rather than just `isVariableValue`.

Still reimplemented locally rather than called, for the unchanged reason already in its javadoc: `VSUtil`'s static initializer reaches for a live Spring `ApplicationContext` that this file's plain `PropertyPathTest` does not have.

`trendLineType` and `position` stay out of `DYNAMIC_CAPABLE_CONSTRAINED_STRINGS` for the unchanged reason too — neither is `DynamicValue`-backed, so an `"=..."` there would be stored and never resolved by anything: the silent-success failure this whole gate exists to prevent. There is now a test pinning that.

## Tests

`PropertyPathTest` — 53 pass, 2 new:
- `visible` accepts `"=Gauge1 > 20"` and stores it unresolved.
- `trendLineType` still refuses one, naming the property.

## Notes for the reviewer

- **Not live-verified yet.** This is the tool/plumbing half. Whether StyleBI actually re-renders the property when the expression's inputs change is a separate question that needs a live session — `VSUtil.getDynamicValueDependeds` routes a script value through `getReferencedAssets`, so the dependency *should* register, but that was traced, not run.
- Paired with **stylebi-wiz#2650**, which wires the same `=expression` half into the composer plugin (`set_assembly_properties`, and shelf `aggregate`/`dateLevel`). Neither PR depends on the other to compile; without this one, an expression on `visible` specifically is still refused server-side.
- `set_format`'s colours are **not** in scope here — #76754 / stylebi-wiz#2644 already landed those, plugin-side only. I verified no server change was needed there: with `colorType` null, `FormatPainterService.setUserFormat` calls `setForegroundValue(model.getColor())`, and `VSFormat.setForegroundValue`'s javadoc is literally "expression or RGB number", so the dynamic string reaches the `DynamicValue` intact.


jshobe-inetsoft (approved):
# Review: stylebi PR #5349 — fix(76757): let a constrained String property take an "=expression" dynamic value

**Base:** `stylebi-wiz-main-rebase` (feature branch, not `main`) · **Head:** `bug-76757`
**Files:** `core/src/main/java/inetsoft/web/wiz/viewsheet/PropertyPath.java` (+17/-13), `core/src/test/java/inetsoft/web/wiz/viewsheet/PropertyPathTest.java` (+43/-0)

## Verdict

- CI is green: `build`, `Semgrep Scan`, `Semgrep OSS`, `license/cla` all pass. No `Claude Code Review / claude-review` check is present to exclude.
- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN` — no merge conflicts.
- The core claim of the PR (that `visible` is `DynamicValue`-backed and resolved via `ViewsheetSandbox.executeDynamicValue` regardless of which of the two `DynamicValue` conventions — `$(...)` variable or `=...` script — is used) was independently verified against `VSUtil.isVariableValue`/`isScriptValue`/`isDynamicValue`, `DynamicValue.getRValue()`, `VSAssemblyInfo.visibleValue`/`getViewDynamicValues()`, and the Composer's `dynamic-combo-box.component.ts:updateType()`. The new local `isDynamicValue` predicate is logically identical to `VSUtil.isVariableValue(text) || VSUtil.isScriptValue(text)` (Java `&&`/`||` precedence makes the unparenthesized form correct).
- One **non-blocking** documentation-staleness issue is noted below; it does not affect correctness, tests, or the build, so it does not block this PR, but it should be cleaned up (either here or in a fast follow).

---

## Findings

### Non-blocking

1. **Stale cross-reference in a sibling file's javadoc.** `core/src/main/java/inetsoft/web/wiz/binding/BindableColumns.java` has its own independent, narrower `isDynamicVariableReference(String)` (intentionally `$(...)`-only, since a bound column name can't be a script expression). Its javadoc says:

   > "... for the same reason `{@code PropertyPath.isDynamicVariableReference}` does ..."

   This PR renames `PropertyPath`'s method from `isDynamicVariableReference` to `isDynamicValue`, so that cross-reference in `BindableColumns.java` now points to a method name that no longer exists on `PropertyPath`. It's a comment-only staleness issue (nothing fails to compile, since it's prose in a javadoc, not an `@link`/`@see` tag that the compiler would check), but it will mislead a future reader who goes looking for `PropertyPath.isDynamicVariableReference`. Suggest updating that one line to say `PropertyPath.isDynamicValue`.

2. **Style nit:** the new boolean expression relies on `&&` binding tighter than `||` without parentheses:
   ```java
   return text != null &&
      (text.startsWith("$(") && text.endsWith(")") || text.startsWith("="));
   ```
   This is correct as written (mirrors `isVariableValue(text) || isScriptValue(text)` exactly), but an explicit `(text.startsWith("$(") && text.endsWith(")")) || text.startsWith("=")` would make the grouping unambiguous at a glance rather than relying on precedence. Purely cosmetic.

3. **Explicitly out of scope / already covered by the PR description**, listed here only for completeness of the review rather than as a new finding: the PR itself flags that end-to-end re-rendering when the expression's inputs change was traced through `VSUtil.getDynamicValueDependeds`/`getReferencedAssets` but not run live, and that the composer-plugin half that actually emits `=expression` values lives in a separate repo/PR (`stylebi-wiz#2650`). Neither is a defect in what's being merged here — this PR is scoped to the server-side gate, and the unit tests correctly pin both the newly-allowed case (`visible` + `=expression`) and the still-refused case (`trendLineType` + `=expression`, proving the gate discriminates by whether the property is `DynamicValue`-backed, not by the value's shape).

### Verified as correct (no issues)

- **`isDynamicValue` correctness:** confirmed byte-for-byte equivalent to `VSUtil.isVariableValue(String) || VSUtil.isScriptValue(String)` by reading `VSUtil.java`. The `PropertyPathTest`'s stated reason for re-implementing rather than calling `VSUtil` (its static initializer reaches a live Spring `ApplicationContext` via `ConfigurationContext.getSpringBean`, which throws outside a running server) is a pre-existing constraint in this file, unchanged by this PR, and is the same rationale already used for the pre-existing `$(...)`-only version.
- **Scope of `DYNAMIC_CAPABLE_CONSTRAINED_STRINGS`:** still `Set.of("visible")` only — `trendLineType` and `position` remain excluded, and a new test (`refusesAnExpressionForAPropertyThatIsNotDynamicValueBacked`) pins that an `=expression` on `trendLineType` still throws, naming the property. This correctly preserves the fix from the prior bug (#76412/76754-era silent-success defect) that this same gate exists to prevent.
- **`VSAssemblyInfo.visibleValue`** is a `DynamicValue` added to `getViewDynamicValues()`/`getDynamicValues()`, which is what `ViewsheetSandbox.executeDynamicValues()`/`executeDynamicValue()` walks at render time — confirming an `=expression` stored into `visible` really is resolved at render time and not just stored inert.
- **Composer UI parity:** `dynamic-combo-box.component.ts:updateType()` classifies a bound value into `ComboMode.VARIABLE` (`$(...)`) or `ComboMode.EXPRESSION` (leading `=`) using the identical two prefixes — supporting the PR's claim that "Variable" and "Expression" are the two recognized halves of the same convention in the existing UI, not something novel being introduced by this change.
- **Backward compatibility:** this is a pure widening of what's accepted (a previously-rejected `=expression` value on `visible` is now accepted); no previously-accepted value becomes rejected. No existing caller can regress.
- **Test coverage:** the diff's stated "53 pass, 2 new" matches — the file has exactly 53 `@Test` methods after the change. The two new tests are well-isolated (separate fixture classes `StringVisible`/`ChartLine`, no naming collisions with other fixtures in the file) and each assert both the resulting stored value and (for the negative case) that the exception message names the offending property, consistent with the rest of the file's conventions.
- **Error handling:** the negative test correctly exercises the existing `requireAllowedValue` rejection path (case-insensitive enum-token matching against `CONSTRAINED_STRINGS`), which still fires for `trendLineType` since it's outside `DYNAMIC_CAPABLE_CONSTRAINED_STRINGS`; the exception message and its tokens are unchanged for that path.
- **Security:** allowing `=expression` through this AI/"wiz" property-setting tool does not introduce a new capability into StyleBI — a Composer user can already type the identical `=expression` string into the same `visible` field via the UI's "Expression" button, and the expression executes through the same Rhino script-engine sandbox as any other viewsheet script (see `DynamicValue`/`ViewsheetSandbox.executeDynamicValue`). This PR only lets the plugin reach a pathway that was already reachable by a human user through the same authenticated session; it does not bypass any existing authorization or sandboxing boundary.

## Summary

The change is small, precisely scoped, well-reasoned, and backed by tests that pin both the new positive case and a deliberately-preserved negative case. All factual claims in the PR description about `VSUtil`, `DynamicValue`, `VSAssemblyInfo`, and the Composer UI's Variable/Expression convention were independently verified against the source and hold up. The only issue found — a stale method-name cross-reference in a sibling file's javadoc — is cosmetic and does not block merging.

PR #5350, opened 2026-09-17
fix(wiz): keep an agent-paired runtime's socket session current, not just filled once

larryliang-inetsoft:
## Summary
- `ViewsheetSessionService`/`WorksheetEditService`/`ScriptEditService`'s `applySocketSession` only wrote the browser's socket session id onto a runtime when the field was still `null` — a fill-once guard.
- Once set (by the human's normal open flow, or by an earlier agent mint), it could never be corrected again — even after the browser's WebSocket reconnected (idle timeout, proxy hiccup) and made that id dead.
- `SheetAgentBroadcastService.broadcastRefresh` finds a non-null id and has no way to know it's stale, so every subsequent agent-driven edit silently stopped reaching the browser: `add`/`set_table_source`/`set_table_fields`/`refresh_viewsheet` all reported success while nothing visibly changed in the Composer tab.
- Re-pairing with a fresh code didn't help either: the new `JoinSession`'s own `socketSessionId` was correctly captured live at that mint, but the fill-once guard still refused to overwrite the runtime's already-non-null (stale) value with it.
- **Confirmed live**: a paired viewsheet's agent edits stopped reaching the browser entirely. Only the human's own manual Refresh — which goes through `CoreLifecycleService`'s *unconditional* `rvs.setSocketSessionId(dispatcher.getSessionId())` — fixed it, after which agent-driven pushes (confirmed with a plain `set_title`) worked again with no other change.

## Fix
Always take the current `JoinSession`'s `socketSessionId`/`socketUserName` when non-null, rather than only filling an empty field. A `JoinSession`'s values are captured fresh at mint time from a live STOMP request, so they are always at least as trustworthy as whatever the runtime already holds.

## Test plan
- [x] Added a regression test to each of the three services' test classes: mocks a stale-but-non-null runtime socket alongside a session carrying a fresh one, asserts the fresh value is written through (`setSocketSessionId("fresh-live-socket")`).
- [x] `mvn test -pl community/core -Dtest=ViewsheetSessionServiceTest,WorksheetEditServiceTest,ScriptEditServiceTest` — 52/52 passing.
- [x] Confirmed a broader `inetsoft.web.wiz.**.*Test` run has pre-existing, unrelated `NoClassDefFoundError` failures on the unmodified baseline too (verified via `git stash`) — not caused by this change.


lrry05x (changes requested):
## Review

I read the diff, the three modified `applySocketSession` methods and their new tests, and traced
`JoinSession`'s `socketSessionId`/`socketUserName` lifecycle through `SheetSessionService`,
`SheetPairingController`, and `CoreLifecycleService`/`VSLifecycleControllerService` to check the
"always take the session's value" claim against how/when that value actually gets refreshed.

### Blocking concern: the fix's "self-heal" claim doesn't hold for the same-token case, and can regress the very recovery path described in the PR

`JoinSession.socketSessionId()`/`socketUserName()` are captured **once**, at `mint()` time, from
whatever STOMP session is live *then* (`SheetPairingController.mintViaSocket` →
`accessor.getSessionId()`). Nothing subsequently refreshes them for the lifetime of that token:

- `SheetSessionService.resolve()` (community/core/.../pairing/SheetSessionService.java:147-160) only
  bumps the TTL; it rebuilds the `JoinSession` by copying `s.socketSessionId()`/`s.socketUserName()`
  through unchanged.
- `retarget()`/`popFocus()`/`setFollowFocus()` do the same (copy `s.socketSessionId()` through).
- `socketClosed()` (SheetSessionService.java:187-206) explicitly does **not** touch whole-sheet
  (`editorContext == null`) sessions — the doc comment says so directly: "Whole-sheet sessions are
  deliberately left untouched here -- they keep today's TTL-only lifetime." So a whole-sheet
  session's `socketSessionId` is frozen from mint until the token expires or a *new* mint happens.

That means for a single pairing token reused across many agent calls (the normal case — an agent
doesn't re-pair between every tool call), `session.socketSessionId()` is exactly as stale after a
WebSocket reconnect as `rvs.getSocketSessionId()` already is. Overwriting one stale value with
another stale (but identical) value is a no-op — it does not "self-heal" the case the PR's own
narrative leads with (reconnect happens, no re-pair). The fix only actually helps the *re-pair*
case ("Re-pairing with a fresh code didn't help either" in the description) — there, a brand-new
mint really does capture a fresher id, and the old fill-once guard on the runtime refused it. That
part is correctly fixed.

More importantly, this unconditional overwrite can **regress the exact recovery path described in
the PR**: the description says a human's manual Refresh — going through
`CoreLifecycleService`'s/`VSLifecycleControllerService`'s *unconditional*
`rvs.setSocketSessionId(dispatcher.getSessionId())` — is what fixes a stale runtime today, "after
which agent-driven pushes worked again." Under the **old** fill-only-if-null code, that manual fix
sticks, because every later `applySocketSession` call sees a non-null field and leaves it alone.
Under the **new** unconditional-overwrite code, the very next agent call on the same (un-re-paired)
session token will call `applySocketSession(rvs, session)` with the *same frozen, still-stale*
`session.socketSessionId()` from mint time, and stamp it right back over the fresh value the human's
Refresh just installed — silently re-breaking the broadcast again, this time caused by the agent's
own next call rather than by staying silently broken. That is a worse failure mode than before: it
turns a one-time staleness into something that flips back to broken after every human-side recovery,
until the token is re-paired.

I don't think this needs a redesign necessarily, but it needs to be checked before merging: was the
fix actually live-verified against "browser reconnects (or human manually refreshes) *without*
re-pairing, then one more agent edit" — as opposed to only "re-pair with a fresh code, then edit"?
The PR's live-verification note only describes the broken state and (implicitly) the re-pair fix; it
does not describe re-testing the reconnect-without-repair scenario against the new code. If that
scenario wasn't retested, please do so, or reconsider deriving the runtime's socket target from
something that's actually kept live (e.g. `SheetSessionService.findBySocketAndRuntime`-style lookup
keyed by the current STOMP accessor) rather than a value frozen at a specific token's mint time.

### `socketUserName` overwrite
Same source, same staleness profile as `socketSessionId` — I don't see a materially different risk
here beyond what's already covered above (it's derived from the same mint-time snapshot and
overwritten unconditionally in the same call). No separate concern beyond the general one above.

### Tests
The three new regression tests (`ViewsheetSessionServiceTest`, `WorksheetEditServiceTest`,
`ScriptEditServiceTest`) do exercise the real `applySocketSession` path through `mutate`/`apply`, not
just re-assert mock wiring: with `rvs.getSocketSessionId()` stubbed non-null
("stale-dead-socket") and the resolved `JoinSession` carrying a different non-null value
("fresh-live-socket"), the old guard (`session.socketSessionId() != null &&
rvs.getSocketSessionId() == null`) would never call `setSocketSessionId`, so `verify(rvs)
.setSocketSessionId("fresh-live-socket")` would fail against the pre-fix code and passes now — a
real regression test for the change as written. They don't, and can't, cover the frozen-token
scenario above since that requires two calls across a simulated reconnect/refresh, not one.

### Code quality / consistency
The doc comments follow this file's existing convention (verbose, WHY-focused, cross-referencing
sibling methods) and are consistent with the rest of `ViewsheetSessionService`/`WorksheetEditService`/
`ScriptEditService`. No dead code, no unrelated refactors — the diff is scoped tightly to the three
`applySocketSession` bodies plus their tests.

### CI / mergeability
`Semgrep`, `Semgrep OSS`, and `license/cla` pass. The `build` check is still pending/in progress as of
this review. `mergeable: MERGEABLE` (no conflicts) per the PR API.

### Verdict
Requesting changes — not because the fix direction is wrong (the re-pair case genuinely needed
fixing), but because the PR's framing overstates what it fixes (no actual self-heal on a bare
reconnect without re-pairing) and I can construct a plausible regression of the described manual-
refresh recovery path that the added tests don't cover. Please confirm with a live re-test of
"reconnect/manual-refresh survives one more agent edit without re-pairing" before merging, or adjust
the design so a fresher, already-live runtime value can't be clobbered by an older frozen session
snapshot.

larryliang-inetsoft:
Good catch — confirmed and fixed in 9618ef73b.

You're right that `JoinSession.socketSessionId()`/`socketUserName()` are frozen at mint time forever (`SheetSessionService.resolve()` just copies the same value forward on every call, never re-derives it). The previous commit's unconditional overwrite in `applySocketSession` was correct for "re-pair with a fresh code" but would indeed re-stomp a human's own later recovery (a manual Refresh's unconditional `rvs.setSocketSessionId(dispatcher.getSessionId())`) on the very next agent call under the same un-re-paired token.

Redesigned: `applySocketSession` in all three services is back to fill-only-if-null (safe for every repeat call). The unconditional overwrite now lives ONLY in `SheetJoinService.join()`, applied exactly once per pairing from the grant's own values — the one moment they're provably fresh (captured from the live STOMP mint request happening right at that call). This fixes the re-pair case without ever touching (and therefore never undoing) whatever a human action sets afterward.

Flipped the three service-level tests to instead assert the case the previous version regressed (a stale-but-non-null runtime socket survives a same-token agent mutation untouched), and added two new tests to `SheetJoinServiceTest` for the join-time stamp itself.

`mvn test -pl community/core -Dtest=ViewsheetSessionServiceTest,WorksheetEditServiceTest,ScriptEditServiceTest,SheetJoinServiceTest` — 78/78 passing.

lrry05x (approved):
## Second pass on 9618ef73b (join-time redesign)

Verified the redesign directly against my previous finding, not just re-reading the description.

### 1. Does the join-time-only stamp close the gap without reopening another hole?

Yes. I re-traced the same chain as before:

- `SheetJoinService.join()`'s new step 3c stamps `runtimeSheet` (resolved once at 3b via
  `getRuntimeSheetDirect`, and confirmed — by reading `SheetRuntimeAccess.getSheetForPairing`,
  which calls `getRuntimeSheetDirect` internally — to be the same underlying runtime object that
  `resolve()`/`mutate()`/`read()` later operate on, not a different lookup) unconditionally, but
  **exactly once**, from `grant.socketSessionId()`/`socketUserName()` — values captured moments
  earlier from the live STOMP mint request that produced this exact join call.
- The three `applySocketSession` methods are back to fill-only-if-null, so a runtime value that's
  since been correctly refreshed by something else authoritative (a human's manual Refresh through
  `CoreLifecycleService`/`VSLifecycleControllerService`'s unconditional stamp) is never clobbered by
  a later agent call replaying the same token's frozen `JoinSession.socketSessionId()`.

This is the right fix: the "provably fresher" property only actually holds at the instant of
join/mint, not on every later resolve — which is exactly what the code now reflects. The
regression I flagged (an agent's own next call undoing a human's manual recovery) is gone: with
fill-only-if-null restored, a non-null, human-healed value on the runtime is left alone.

### 2. Any new concurrent-session/ordering concern from stamping at join() rather than per-call?

No new concern beyond what already existed pre-PR. `join()` still only writes once per successful
pairing, same as the original (pre-bug) design's intent, just now unconditional at that one moment
instead of conditional. This is actually a narrower window for cross-session clobbering than the
now-reverted per-call-unconditional version — it can only clobber at the moment of a *new* join,
which `SheetSessionService.findOpen()` already discourages duplicating for the same owner+sheetType.
No change needed here.

### 3. Do the new/flipped tests exercise the real path and would fail if either half were missing?

Ran both test classes directly (`mvn test -pl core -Dtest=ViewsheetSessionServiceTest,
WorksheetEditServiceTest,ScriptEditServiceTest,SheetJoinServiceTest`, offline, against
9618ef73b) — confirmed 6+38+8+26 = **78/78 passing**, matching the PR description.

- The three flipped `mutate...`/`apply...`/`applyOnRuntimeIfChanged...Never...` tests stub the
  runtime's *current* socket as `"human-healed-live-socket"` and the token's own frozen value as a
  *different* `"stale-frozen-at-mint"`, then assert `verify(rvs/rws, never()).setSocketSessionId(any())`.
  These would fail against the previous (now-reverted) unconditional-overwrite code — I confirmed
  this by re-reading that code path in the prior commit, which had no guard at all — so they do
  guard specifically against the regression I raised, not just re-assert existing mock wiring.
- `joinDoesNotStampWhenTheRuntimeIsNotFound` correctly exercises the new null-guard in `join()`
  (would NPE without it).
- `joinStampsTheRuntimesSocketSessionUnconditionallyFromTheFreshGrant` proves the stamp fires and
  uses the grant's fresh values. One minor, non-blocking gap: as written (relying on Mockito's
  default `null` return rather than stubbing a pre-existing non-null value, to avoid an
  unnecessary-stubbing failure under strict stubbing, per the test's own comment), it doesn't
  distinguish "unconditional overwrite" from "fill-only-if-null" at the test level — that property
  is true by direct inspection of `join()`'s code (no read-then-compare, just an unconditional
  write), just not independently pinned by this specific test. Could be closed with
  `lenient().when(rws.getSocketSessionId()).thenReturn("stale-dead-socket")` if you want the test
  itself to prove it, but I don't consider this blocking given the source is unambiguous.

### Other notes (non-blocking)
- None of the three service test files have an explicit "fills the socket when the runtime's field
  actually is null" positive-path test asserting `setSocketSessionId` **is** called — that behavior
  predates this PR and is only exercised incidentally by other tests (e.g. defaults on unstubbed
  mocks), not newly introduced or worsened here, so not something this PR needs to add.
- Doc comments again match this file's existing verbose, WHY-focused convention and now correctly
  describe the actual, narrower invariant (frozen-at-mint, one-time authoritative stamp at join).

### Verdict
Approving. The redesign directly and correctly addresses the regression I raised, tests genuinely
exercise both halves of the fix, and I independently reran the four affected test classes (78/78
green) against the actual new commit rather than trusting the description alone. Leaving the merge
decision to the user, per standing instructions — this only updates my review verdict.

PR #5352, opened 2026-09-18
fix(76748): calendar period comparison drops days between non-adjacent boundary dates

Nelson-cao:
## Summary

Redmine #76748: a double-calendar period comparison (`rangeComparison:true`) with a period defined by 2+ non-adjacent dates meant to represent a contiguous range (e.g. a month's start/end) silently dropped every day strictly between the boundary dates from the query result.

## Root cause

`CalendarVSAssembly.getConditionList(DataRef[])`'s period-comparison branch reused `createConditionList()`, which is also the single-calendar OR'd-day-selection builder — it only merges two date-cells into one contiguous range when they are calendar-adjacent (`start == endDay`); otherwise each cell becomes its own OR'd single-day condition. For a period comparison, a period's date-cell array is meant to be one `[start,end)` window, not a set of independently OR'd days, so non-adjacent boundary dates (the common case — a month's first and last day) collapsed to two single-day conditions instead of the full range.

## Fix

Added `createPeriodRangeConditionList()`, used only by the period-comparison branch, which takes the min/max across a period's date-cell array as literal range endpoints (mirroring the existing non-period `dual`/range branch's min/max construction). `createConditionList()` itself is untouched for its other caller (single-calendar OR'd independent day-selections), which still only merges calendar-adjacent cells.

`core/src/main/java/inetsoft/uql/viewsheet/CalendarVSAssembly.java`

## Test plan

- [x] Added `CalendarVSAssemblyTest` with 3 cases:
  - period comparison with non-adjacent boundary dates now builds one contiguous range per period (the regression test — fails against the pre-fix code with 15 condition items instead of 7)
  - single-calendar non-adjacent days still stay OR'd as independent days (unchanged, anti-regression)
  - single-calendar adjacent days still merge into one range (unchanged, anti-regression)
- [x] `./mvnw -pl core test -Dtest=CalendarVSAssemblyTest` — all pass
- [x] Full `-pl core -am clean install` (guards against stale lazy-init proxies from the new methods) — succeeds

Root cause and fix were verified against a real, live StyleBI instance (screenshotted reproduction), not just static reading — see `docs/teams/2026-09-17-bugs-76748-calendar-range-comparison/bug-76748/01-diagnosis.md` in the `stylebi-wiz` repo for the full investigation trail.

Note: per this batch's instructions, this PR is intentionally **not** auto-merged.


Nelson-cao:
## Round 1 — review-workflow adversarial review

**Verdict: CHANGES REQUESTED** (posted as a plain comment, not a formal review — GitHub rejects `gh pr review -r`/`--approve` against your own PR)

**Head commit reviewed:** `f070b531819e43fc295b9d963d8baf3d89485707`

### What was checked

**Root-cause fidelity (this batch's extra check, against `01-diagnosis.md`/`02-refute.md`/`03-fix.md`):**
- Confirmed the diff's `createPeriodRangeConditionList` is exactly the confirmed root cause fix: the period branch of `CalendarVSAssembly.getConditionList(DataRef[])` no longer reuses `createConditionList()` (the adjacency-merge builder), and instead takes literal min/max across a period's date-cell array, mirroring the existing `dual`/non-period range branch's construction (same `level=0` scheme as that branch).
- Verified `createConditionList()` itself is byte-for-byte behavior-preserving for its one remaining caller (single-calendar OR'd-days, line 752): the diff only extracts the per-cell parsing into `parseDateRange()`; the adjacency check, level assignment (`length==1?0:1`), and loop structure are untouched. Confirmed by reading the full extracted PR-branch file, not just the diff hunks.
- Grepped the whole `core` module for the same "adjacency-merge" pattern (`start == endDay`) and for other callers of `createConditionList` — only the one file, only the two known call sites. No other site still carries the diagnosed defect.
- Checked the diagnosis's flagged consumer, `RangeCondition.createRangeConditions` (fixed 4-item-stride walk assuming one `(greater,AND,less)` triple per period) — confirmed the new method always emits exactly one triple per period via a single `appendCondition` call, so the shape this consumer assumes is actually *more* reliably satisfied post-fix than pre-fix (pre-fix, a non-adjacent multi-date period could have emitted >1 triple, silently breaking this consumer's stride assumption in a case nobody had noticed). Not a regression.
- The clamp in `appendCondition`/`getLimitRange()` ("Finding B", dropped by diagnosis/refutation) is reused unchanged and unconditionally by the new method — confirmed by reading, not just trusting the fixer's claim.
- Does not key on any literal from the bug report's own repro dates — the fix is a general min/max-over-array rule, not a special case for April/May 2023.

**Independent test execution:**
- Ran `CalendarVSAssemblyTest` myself against the actual PR head in the fixer's own worktree (`E:\inetsoft\stylebi-fix-76748`, confirmed `HEAD == f070b5318`, clean tree) — 3/3 pass, independently reproducing the fixer's and tester's reported result.
- Actively tried to break the "generalizes to >2 dates" claim (`03-fix.md`'s claim, re: `CalendarDisplayService.validateDateCount` not capping a period at 2 dates): added a scratch test with **3 non-adjacent, deliberately out-of-order date cells per period** (period 1 array order: day20, day1, day10). Ran it against the PR's actual code — passed, correctly computing true min/max regardless of array order or count. Reverted the scratch edit before finishing (worktree left clean, matching PR head).

### Findings

**Important (not blocker) — missing regression coverage for >2 dates per period, an overfit-smell per review-workflow's own criteria**

`core/src/test/java/inetsoft/uql/viewsheet/CalendarVSAssemblyTest.java` — the only period-comparison regression test (`periodComparisonBuildsAContiguousRangePerPeriodInsteadOfOringTheBoundaryDates`) uses exactly 2 dates per period. With exactly 2 elements, "first/last element" and "true min/max across the array" produce identical results, so this test suite **would also pass against a narrower, still-buggy implementation** that used `dates[index]`/`dates[index+length-1]` instead of a real min/max scan — even though `CalendarDisplayService.validateDateCount` allows (and the fixer's own account claims to support) periods with more than 2 date cells.

I independently confirmed via the scratch probe above that the *actual shipped code* is not narrower — it does the true min/max scan and handles out-of-order, >2-date periods correctly. So this is a coverage gap, not a live defect. But per review-workflow's explicit overfit-smell list ("the regression test asserts only the exact repro input, so it would still pass against a narrower, still-broken version of the change"), this is exactly that shape and should be closed before this is the final round: add a 3+-date-per-period test case (ideally with cells out of array order) to `CalendarVSAssemblyTest`, so a future refactor that regresses to first/last can't sneak past CI silently.

(Independently, the concurrent verify pass in `04-verify.md` flagged the identical gap by static reading; this finding is corroborated by two independent methods — static reading and an executed counter-example.)

### Verdict

**Changes requested** — one Important finding (test-coverage gap only; the fix logic itself is verified correct by independent execution, including a case not covered by any existing test).

Per this batch's instructions: PR is not being merged as part of this review loop regardless of verdict.

Nelson-cao:
## Review Round 2 — commit `fb2f4a419`

**Verdict: APPROVED**

### What was checked

- **Round-1 finding status: RESOLVED.** Round 1 flagged that the only period-comparison
  test used exactly 2 dates per period, where "first/last array element" and "true
  min/max scan" give identical results — so the test couldn't distinguish a genuine
  min/max implementation from a narrower, still-buggy first/last-element one.

  Commit `fb2f4a419` adds `periodComparisonScansTheTrueMinAndMaxAcrossOutOfOrderPeriodDates`:
  3 dates per period, deliberately out of array order (period 1: Apr 15, Apr 1, Apr 8 —
  true min/max both away from the first/last slots). I read `createPeriodRangeConditionList`
  directly (`CalendarVSAssembly.java:134-158`) and confirmed it does a real running
  min/max scan over every cell (`range[0] < startDay` / `range[1] > endDay`), not an
  indexed first/last read — so this test is checking real production behavior, not just
  asserting against itself.

  I independently re-ran the suite against the actual PR head (didn't just trust the
  fixer's report): **4/4 pass** on `fb2f4a419` (worktree `HEAD` and PR `headRefOid` both
  confirmed to be `fb2f4a419e477bd0e076441601b12a767b4dc16a`).

  Also checked for a narrower-but-still-passing alternative implementation the fixer
  might have missed — e.g. "sort the array, take first/last" would also produce a
  *correct* result here (sorting fixes the min/max problem for scalar dates), so it's
  not a false-negative risk. I didn't find a realistic narrow-but-wrong implementation
  that both plausibly resembles the original bug and would still pass this new test.
  The specific narrow case the fixer targeted (first/last array *position* instead of
  true min/max) is genuinely closed.

- **New diff reviewed on its own terms**: it's a pure test addition (27 lines, one new
  `@Test` method, no production code touched). Verified:
  - Date-token month-index convention (`d2023-3-1` → April 1, 0-indexed month) matches
    the existing test 1's established convention — the in-code comments ("April 15,
    April 1, April 8" / "May 1, May 8, May 15") correctly correspond to the literal
    tokens (`d2023-3-*` = April, `d2023-4-*` = May).
  - `getConditionList(DataRef[])` splits `dates` at `dates.length / 2` for period mode
    (`CalendarVSAssembly.java:697`), so 6 date cells split into two size-3 sub-arrays as
    the test assumes — verified by reading, not assumed.
  - `ConditionItem` count/index scheme (0/2 for period 1's greater/less, 4/6 for period
    2's) is fixed at 3 items per period (`appendCondition` always emits exactly one
    greater/AND/less triple) regardless of how many source date cells fed the min/max —
    so the indices hold for 3-date periods the same as the existing 2-date test.
  - Expected values are internally consistent with the date math: min(Apr15,Apr1,Apr8)
    = Apr1, ceil(max) = Apr16; min(May1,May8,May15) = May1, ceil(max) = May16. No typos.

### Independent test execution

```
cd E:\inetsoft\stylebi-fix-76748
git fetch && git log --oneline -3   # HEAD = fb2f4a419, matches PR headRefOid
./mvnw -q -pl core -Dtest=CalendarVSAssemblyTest -Dsurefire.failIfNoSpecifiedTests=false -Dcheckstyle.skip=true -Dlicense.skip=true test
```
Result: `Tests run: 4, Failures: 0, Errors: 0, Skipped: 0`.

### Summary

Round-1's Important finding (test-coverage gap for >2-date periods / out-of-order cells)
is fixed, not just claimed-fixed. No new findings on the round-2 diff. This is likely the
last round before the batch stops for human merge — **not merging this PR**, per this
batch's explicit instruction; that remains a separate human-authorized action.

PR #5354, opened 2026-09-18
Add editConcatenation to change an existing concatenation's type/distinct in place

Nelson-cao:
## Summary

Bug #76743 (WBS-052): no agent tool could change an existing worksheet concatenation (UNION/INTERSECT/MINUS) assembly's `concatType` or duplicate-row (`concatDistinct`) setting in place — the only workaround was deleting and rebuilding the whole assembly.

- Adds `Editor.editConcatenation(String concatName, String opType, Boolean distinct)` in `WorksheetEditService.java`, wired into `WorksheetAgentController`'s dispatch switch as `"edit_concatenation"`. Reuses `ConcatenatedTableAssembly.setOperator(int, TableAssemblyOperator)`, the same already-proven-stateless primitive `addConcatSubtable` uses.
- `opType == null` preserves each pair's existing operation rather than defaulting to UNION (`parseConcatType(null)` defaults to UNION, which is correct for create but wrong for a distinct-only edit).
- The new value(s) are applied uniformly across every adjacent pair — matching `addConcatenation`'s own creation-time convention. This is a deliberate, disclosed scope decision: it flattens any pre-existing mixed per-pair state (`A UNION B MINUS C`, a legal assembly per `WorksheetReadService`'s own doc comment) rather than retargeting a single connection the way the native Composer "Edit Concatenation" dialog's default mode can. Documented in the method's own doc comment and in `readConcatType`'s doc comment.
- Companion read-side fix: adds `readConcatDistinct`/`TableModel.concatDistinct` (mirroring `readConcatType`'s exact pattern — `null` when pairs disagree, since `Boolean` has no "MIXED" sentinel), so the new write is round-trip-verifiable through `read_worksheet_model`.

## Test plan

- [x] `./mvnw -o -pl core -Dtest=WorksheetEditServiceMutatorsTest -Dsurefire.failIfNoSpecifiedTests=false test` — 257 tests, all pass (5 new: type change, distinct-only preserves operation, applies to every pair on a 3-table concat, not-found throws, both-null throws)
- [x] `./mvnw -o -pl core -Dtest=WorksheetReadServiceTest -Dsurefire.failIfNoSpecifiedTests=false test` — 54 tests, all pass (2 new: `concatDistinct` reported when pairs agree, `null` when they disagree; plus updated the existing null-fields/serialization tests to also cover `concatDistinct`)


Nelson-cao:
Round 1 review: **no blocking issues** (cannot use `gh pr review --approve` — GitHub rejects self-approval since the PR author and reviewing identity are the same account here). Verified against source (not just the fix report): the `opType == null` guard correctly avoids `parseConcatType(null)`'s UNION default by preserving each pair's existing operation via `getKeyOperator()` (confirmed this never returns null, per `TableAssemblyOperator.java:396-416`). Refuter's required disclosure is present in the Java doc comment (WARNING block on `editConcatenation`, plus updated `readConcatType` doc). All 5 diagnosed test cases present with real assertions in `WorksheetEditServiceMutatorsTest`, plus companion `concatDistinct` read-side tests. Diff scope is tight (dispatch case, one new method, read-side companion, tests only) with the single `TableModel` construction site correctly updated for the record's positional field order. CI green, mergeable, clean.

PR #5355, opened 2026-09-18
fix(wiz): reject blank/null text in create/update_script_library_function (#76765 SSL-003)

Zxm-happy:
## Summary
- `ScriptLibraryController.create()` validated `name` via `isBlank()` but only null-coalesced `text` (no blank check), letting an empty function body be persisted silently.
- `GraalJavaScriptEngine.installLibraryFunctions()` permanently skips installing any library function whose source is empty, with no log — so the function becomes unreachable and calling it throws `ReferenceError`, indistinguishable from the function never having existed, and nothing at create time indicates a problem.
- `update()` has the identical gap (`text` only null-coalesced, no blank check), letting an update blank out an already-referenced function's text with the same silent-failure outcome.

## Scope note
SSL-003 as filed only named `create()`. Diagnosis + an adversarial refute (CONFIRMED) found the identical missing-validation pattern in the sibling `update()` method, so this PR widens the fix to cover both — fixing only `create()` would leave `update_script_library_function(text:"")` reachable with the same silent-degradation bug, likely to resurface as a "new" bug on the next test pass.

## Changes
- `core/src/main/java/inetsoft/web/wiz/controller/ScriptLibraryController.java`: `create()` and `update()` now throw `IllegalArgumentException` when `text` is `null` or blank, mirroring the existing `name` validation style.
- `core/src/test/java/inetsoft/web/wiz/controller/ScriptLibraryControllerTest.java`: added `create_rejectsBlankText`, `create_rejectsNullText`, `update_rejectsBlankText`.

## Test plan
- [x] `mvn -pl core -am -Dtest=ScriptLibraryControllerTest -Dsurefire.failIfNoSpecifiedTests=false test` — 7 tests run, 0 failures, 0 errors (real, non-piped exit code 0).


Zxm-happy:
Review (formal --approve blocked: reviewer and PR author share the same GitHub identity in this environment, so GitHub rejects self-approval -- recording findings as a comment instead).

Reviewed diagnosis → refute (CONFIRMED) → fix → lead's independent verify trail, then the PR diff directly.

- Root cause fix matches the CONFIRMED diagnosis: create() and update() now reject null/blank text via isBlank() (not isEmpty() or a hardcoded "" check) -- the create_rejectsBlankText test uses whitespace-only "   " specifically to exercise isBlank() semantics, not just empty string.
- Scope widening to update() is explicitly documented in the PR's "Scope note" section, consistent with the diagnosis/refute trail's argument (same missing-validation pattern in the sibling method; leaving it unfixed would resurface as a "new" bug).
- All three new tests assert `verify(fixture.lib, never()).setScript(...)` in addition to the thrown exception -- confirms no side effect happened, not just that an exception type matched.
- Diff scope is tight: 2 files only (Controller +10/-1, Test +35), nothing unrelated.
- Independently re-ran `mvn -pl core -am compile -q` in the worktree (non-piped, exit 0) as a broader build sanity check beyond the lead's targeted ScriptLibraryControllerTest run (7/7 pass, exit 0).
- All CI checks green (build, Semgrep Scan, Semgrep OSS, license/cla).

Non-blocking nitpick: create() line ~125 still has `request.text() == null ? "" : request.text()` in the setScript call, which is now dead code (guaranteed non-blank by the new check above it) -- update()'s equivalent was simplified to `request.text()` but create()'s wasn't. Not worth a fix round on its own.

No blockers. Proceeding to squash-merge.

PR #5356, opened 2026-09-18
feat(wiz): add Auto Save Recycle Bin restore/delete for admin-chat (Redmine #76719 Gap 2)

NeverStopFMH:
## Summary
- New `inetsoft.web.admin.ai.autosave` package (Controller/Service/ChangePlanService/ChangesetApplyService) wrapping `AutoSaveUtils`/`inetsoft.web.AutoSaveService` directly — never proxying `AutoSaveController` (session-cookie/CSRF EM SPA endpoint under `/api/em/content/repository/autosave/*`).
- No existing "list every draft" primitive exists to wrap, so `listEntries` parses `AutoSaveUtils.getAutoSavedFiles`'s raw storage keys directly, de-escaping the embedded path with the same regex `ContentRepositoryTreeService`'s own (private) `denormalizeAssetName` uses for the identical use case.
- `restore` mirrors `RecycleBinService`'s own collision-dependent risk split; its apply is a real two-step sequence mirroring `AutoSaveController.restoreAutoSaveAssets` exactly (create the live sheet via the real `AutoSaveServiceProxy` primitive, then delete the source draft as a separate step), both captured for rollback.
- `delete` is always `RISK_HIGH` unconditionally (matching Script Library's own convention) but, like Script Library, is fully compensable — the draft's raw bytes are captured before mutating and replayed on rollback.
- Community-tier, not enterprise-gated.
- Companion PR on `stylebi-wiz`: https://github.com/inetsoft-technology/stylebi-wiz/pull/2656

## Test plan
- [x] `mvnw -pl core compile` (full clean compile) — clean
- [x] Live-tested end to end against a running local instance using real browser-crash-simulated drafts:
  - `list_autosave_entries`/`get_autosave_entry` parse storage-key fields correctly
  - `restore` to the default (original) path and to an explicit `assetName`
  - Destination-collision refuse (no `overwrite`) and overwrite-allow (with the destroyed-asset advisory correctly surfaced)
  - `delete` — unconditional high-risk gating, apply-time verification
  - **A genuine post-mutation rollback** (not just a pre-flight refusal): two `restore` entries targeting the same new path in one plan — the first succeeds, the second's fresh apply-time collision check fails against what the first just created, producing a real `"rolled-back"` result. Confirmed the newly-created asset was removed AND the first draft's bytes were replayed back into the auto-save bucket.
  - Multi-entry `delete` in one plan


NeverStopFMH:
## Round 1 review — commit `0326823a692e75bdbdc96f91a95e78126b0c6301`

**Scope checked:** all 12 new files under `core/src/main/java/inetsoft/web/admin/ai/autosave/`, against precedent (`inetsoft.web.admin.ai.recyclebin.*`), the wrapped primitives (`AutoSaveUtils`, `inetsoft.web.AutoSaveService.restoreAutoSaveAssets`, `AutoSaveController`, `ContentRepositoryTreeService.denormalizeAssetName`), and the companion TS client (`stylebi-wiz` PR #2656, `plugin/admin/src/tools/autoSaveRecycleBinTools.ts` + `normalize.ts`).

### Build / tests
- `mvnw.cmd -pl core compile -o -q` from the repo root: **clean**, no output. Verified via `core/target/classes/inetsoft/web/admin/ai/autosave/*.class` (14 class files present, including the `Undo`/`UndoKind` inner classes).
- **No Java unit tests exist** for `inetsoft.web.admin.ai.autosave` (`core/src/test/java/inetsoft/web/admin/ai/autosave/` does not exist), while the immediate sibling `inetsoft.web.admin.ai.recyclebin` this PR claims to mirror has `RecycleBinServiceTest.java` + `RecycleBinChangesetApplyServiceTest.java`. Same gap class as the Script Library PR referenced in the task — should be added before merge, at minimum for `parse()`/`denormalizeAssetName` (the genuinely new parsing logic) and the apply/rollback state machine.
- Community-tier placement confirmed: no import in the new package resolves outside `inetsoft.web`/`sree`/`uql`/`util`, `java.*`, Spring, SLF4J, Jackson.

### Per-invariant verification

**1. `listEntries`/`parse` field mapping.** Compared `parse()`'s `Tool.split(id,'^')` index mapping against `AutoSaveUtils.getUserAutoSaveFiles`'s own convention (`attrs[0]`=scope, `[1]`=type, `[2]`=owner, `[3]`=path) — matches. Malformed-key guard (`attrs.length <= 3` -> `null`) is *stricter* than `AutoSaveUtils.createAssetEntry`'s own guard (`paths.length < 3`, which then unconditionally indexes `paths[3]` — a latent AIOOBE in the existing primitive for a length-3 input), and `listEntries` correctly skips a `null` parse rather than propagating a failure. No counter-example found for a crash on a malformed key.

**2. `denormalizeAssetName` replication.** Byte-for-byte identical regex (`_(34|42|44|47|58|59|60|62|63|92|124)_`) and loop body to `ContentRepositoryTreeService`'s private method. Confirmed.

**3/7. Collision check + apply-time re-derivation.** `wouldCollide`'s `AssetEntry.Type` selection is derived from the *draft's own* type consistently in both `wouldCollide` and the actual `restoreAutoSaveAssets`/rollback `AssetEntry` construction — tried to construct a worksheet/viewsheet type-mismatch false-collision and could not, since worksheets and viewsheets are separately namespaced and every code path keys off the same source-of-truth (`entry.type()`). `applyRestore`/`applyDelete` both re-call `requireEntry`/`wouldCollide` fresh rather than trusting the resolved plan — confirmed.

**Rollback ordering / duplicate-id / lock shape.** Reverse iteration (`for(int i = undoable.size()-1; i>=0; i--)`) matches `RecycleBinChangesetApplyService`'s own shape. A duplicate `id` within one plan is refused at `resolve()` via `seenKeys`, so the "same id twice" scenario never reaches apply. `APPLY_LOCK` is a `static ReentrantLock`, same shape/critical-section coverage as the recycle-bin sibling.

### New findings

**Finding 1 (HIGH) — `rollbackOne` silently reports success for a non-compensable restore rollback; the advisory it computes is unreachable dead code.**

`AutoSaveRecycleBinChangesetApplyService.rollbackOne` (diff lines ~952-993):
```
boolean verified = true;
String advisory = null;
if(undo.kind == UndoKind.RESTORE) {
   if(undo.compensable) {
      ... verified = !assetRepository.containsEntry(created);
   }
   else {
      advisory = "the asset that overwrite:true destroyed at ... was not restored -- there is no live inverse for it";
   }
   restoreBytes(undo, user, failures);
}
...
if(!verified) {
   failures.add(new RollbackFailure(undo.key, "... " + (advisory == null ? "" : " (" + advisory + ")")));
}
```
When `!undo.compensable` (a restore that collided via `overwrite:true`), `verified` is **never reassigned** — it stays `true` from initialization. `advisory` is populated in exactly this branch, but the *only* place `advisory` is ever read is inside `if(!verified)`, which is unreachable whenever `advisory != null`. The result: this rollback entry is written to the audit log as `STATUS_VERIFIED` and never appears in `rollbackFailures` — the caller is told the changeset "rolled back" cleanly, with zero indication that this particular entry left the destroyed-pre-existing-asset caveat unaddressed (the source draft bytes do get replayed via `restoreBytes`, but the live sheet that was created/overwritten is deliberately left in place — correct behavior, but silently so). This is exactly the scenario task item #6 asked to be verified, and it fails: no clear advisory surfaces, plain silent success is reported instead.

**Finding 2 (HIGH) — `applyRestore` never queues an `Undo` record when the delete-half of the two-step sequence fails, orphaning the just-created live sheet.**

`AutoSaveRecycleBinChangesetApplyService.applyRestore` (diff lines ~835-892) gates `undoable.add(Undo.restore(...))` behind `if(verified)`, where `verified = !autoSaveRecycleBinService.exists(id, user)` — i.e. verified means "the *source draft* is gone," which is the outcome of the *second* mutating step (`AutoSaveUtils.deleteAutoSaveFile`). But `AutoSaveUtils.deleteAutoSaveFile`'s own internal delete call is wrapped in a try/catch that **swallows any exception and only logs at DEBUG** (`AutoSaveUtils.java` lines 254-267 / 620-631). So: `restoreAutoSaveAssets` succeeds (live sheet genuinely created) -> `deleteAutoSaveFile` silently fails on some transient storage error -> `verified=false` -> status `STATUS_FAILED` -> **no `Undo` record is queued for this entry at all** -> `apply()` sees `STATUS_FAILED` and enters rollback -> rollback iterates only what's in `undoable`, so the live sheet this entry just created is never removed. The caller receives `"rolled-back"` (if other entries had nothing to roll back) or `"rollback-failed"`, in neither case learning that a live sheet now exists at `assetName` outside of any tracked undo. This directly contradicts the PR's own stated design ("both captured for rollback regardless of which half needs undoing" — this class's own javadoc, and the `stylebi-wiz` companion PR's tool description "a restore's own rollback removes the newly-created live sheet"). The precedent this pattern was borrowed from (`RecycleBinChangesetApplyService.applyRestore`'s identical `if(verified) { undoable.add(...) }` gate) is safe there only because Recycle Bin's restore is a *single* atomic primitive (`RecycleUtils.restoreSheet`) where "still recycled" genuinely means "nothing happened" — that assumption does not transfer to a *two-separate-step* primitive, which is exactly the shape this PR itself calls out as the novel part of this design (task point 4). Suggested fix: capture whether `restored==true` independently of the second step's outcome, and always queue the create-half's `Undo` once the live sheet is confirmed created, regardless of whether the delete-half also succeeded.

**Finding 3 (MEDIUM) — `parse()`'s owner-null detection omits `"anonymous"`, silently hiding anonymous/guest-session drafts from every admin.**

`AutoSaveRecycleBinService.parse` (diff line ~1340):
```
String owner = "_NULL_".equals(attrs[2]) || attrs[2].isEmpty() ? null : attrs[2];
```
`AutoSaveUtils` itself treats `"anonymous"` as an equivalent "no real owner" sentinel in two other places in the *same class*: `deleteUserAutoSaveFiles` (`fileUser == null || "anonymous".equals(fileUser) || Tool.equals(fileUser, "_NULL_")`) and `migrateAutoSaveFiles` (`user = "anonymous".equals(user) ? "_NULL_" : user`). If a guest/anonymous session's crash-recovery draft carries `attrs[2] == "anonymous"` (plausible: `Principal.getName()` for an unauthenticated session), `parse()` sets `owner = "anonymous"` (non-null), which then routes through `isVisible()`'s per-owner check (`securityProvider.checkPermission(user, SECURITY_USER, "anonymous", ADMIN)`) — essentially never true for any real admin — so the entry is silently filtered out of `listEntries()` and `getEntry()`/`requireEntry()` report not-found. This is precisely the null/`_NULL_`-vs-real-owner boundary the task asked to be tested, and turns up a third value (`"anonymous"`) the code doesn't handle the way the rest of the codebase's own `AutoSaveUtils` does — defeating admin-mediated recovery for exactly the session class (anonymous/guest) least likely to have another recovery path.

**Finding 4 (LOW/MEDIUM) — `SCOPE_GLOBAL`/`SCOPE_USER` binary collapses `TEMPORARY_SCOPE` into `SCOPE_GLOBAL`, losing a real distinction the restore default depends on.**

`parse()`'s scope mapping only special-cases `AssetRepository.USER_SCOPE` (4); everything else — including `TEMPORARY_SCOPE` (8), which is the scope used for a brand-new, never-before-saved sheet (`SheetOpenService.java`'s own comments confirm `(TEMPORARY_SCOPE, Type.VIEWSHEET, "Untitled-N")` is the real shape for this case) — is reported as `"global"`, identical to a real, previously-saved GLOBAL_SCOPE asset mid-edit. `restore`'s `assetName` defaults to `entry.path()` when omitted (the "put it back where it was" default) — for a `TEMPORARY_SCOPE` entry that default is a generated placeholder like `"Untitled-1"`, not a real prior repository location, which is a materially different situation from restoring a real global asset to its actual saved path. The caller (an LLM) has no way to distinguish these two cases from the projection alone. `RecycleBinService.project`'s identical-looking two-state collapse (which this PR's javadoc cites as precedent) is safe there because a Recycle Bin entry can only ever have been `GLOBAL_SCOPE` or `USER_SCOPE` (a recycled asset was, by definition, already saved) — `TEMPORARY_SCOPE` cannot appear in that area at all, so the precedent doesn't actually transfer to Auto Save Recycle Bin's genuinely three-way scope space. Not functionally unsafe (the apply-time collision check still protects against an actual destination collision), but a real information-quality gap for the plan description/audit trail.

**Finding 5 (LOW) — `mutationEntered` is set before `restoreAutoSaveAssets` can report a well-understood, non-mutating "collision since preview" outcome, which can misclassify that outcome as "unknown state."**

In `applyRestore`, `mutationEntered.set(true)` runs immediately before `autoSaveServiceProxy.restoreAutoSaveAssets(...)`; if that call returns `false` (its own internal, race-safe collision recheck — a well-defined "definitely did not mutate" outcome, not a partial-mutation scenario), the method throws `IllegalArgumentException("assetName: ... was created since preview ...")`. Because `mutationEntered` was already `true`, `apply()`'s catch block adds this to `unknownStateFailures` ("state unknown: apply did not return a verifiable outcome"), and the changeset is reported `rollback-failed` even though nothing was mutated and the source draft is completely intact. Narrow (only reachable if the apply-time `wouldCollide` check and the proxy's own internal recheck disagree within the same call, i.e. an external actor creates the collision between this class's own fresh check and the proxy call), but worth tightening — e.g. only flag `unknownStateFailures` for an exception thrown *after* `deleteAutoSaveFile` is entered, or special-case this specific, self-thrown `IllegalArgumentException`.

Findings 1 and 2 are real correctness defects that can leave the server in a state the caller is told is clean when it isn't (an untracked live sheet, or a silently-unsurfaced non-compensable-rollback caveat) — both are exactly the classes of gap the task's own review methodology was designed to surface, not hypothetical edge cases. Finding 3 defeats the feature for anonymous/guest sessions. Recommend fixing 1-3 before merge; 4-5 and the missing test coverage can be addressed in the same pass or a fast follow-up, but should not be deferred silently.

NeverStopFMH:
Addressed all 5 findings from the round-1 review. New commit: 49fb4fa9d.

**Finding 1 (HIGH, advisory discarded on non-compensable rollback) — fixed, judgment call.**
`rollbackOne` left `verified == true` for a non-compensable restore rollback (correct — removing the newly-created live sheet would only make things worse), but the `advisory` it computed had no channel to reach the caller since the only consumer (`if(!verified)`) never ran.

Judgment call: rather than extending the shared `RollbackFailure` record (bigger, cross-area change) or folding the advisory into `RollbackFailure.error` (would misreport a clean, understood outcome as a "failure"), I grepped `.advisory` usage across `inetsoft.web.admin.ai.*` and found an established, repeated pattern for exactly this situation: `IdentityChangesetApplyService`, `ProviderChangesetApplyService`, and `LicenseChangesetApplyService` all thread a `Map<String, String> advisories` (key -> disclosure) through their own rollback methods, then merge each entry's rollback-time advisory into its ORIGINAL apply-time outcome via a `mergeAdvisory` helper after rollback completes, combining with `" | "` if the outcome already carried its own advisory. I replicated that exact shape here (`rollback`/`rollbackOne` now take an `advisories` map; `apply()` builds `finalResults` via a new `mergeAdvisory`). This is the smallest change that's consistent with how this codebase already solves "an advisory that isn't a failure" in three other areas — no new mechanism invented, and `RollbackFailure.java` was NOT touched.

**Finding 2 (HIGH, orphaned live sheet on cleanup failure) — fixed.**
`applyRestore` gated queuing the rollback `Undo` on `!autoSaveRecycleBinService.exists(id, user)` alone, which only tests whether the best-effort draft cleanup (`AutoSaveUtils.deleteAutoSaveFile`, confirmed it swallows its own exceptions and logs at debug) succeeded — not whether the live sheet was actually created. Fixed by verifying the live sheet's own existence via `assetRepository.containsEntry(...)` independently of the draft-cleanup signal, and queuing the `Undo` whenever the live sheet was actually created, regardless of cleanup outcome. When the live sheet was created but cleanup failed, the outcome now reports `STATUS_VERIFIED` (the actual product-visible result: a live asset now exists) with an advisory that a stale draft copy may remain.

**Finding 3 (MEDIUM, "anonymous" owner sentinel not recognized) — fixed.**
Confirmed both call sites in `AutoSaveUtils` (`deleteUserAutoSaveFiles` and the org-scoped listing helper) treat `"anonymous"` as equivalent to `"_NULL_"` before any owner comparison. `AutoSaveRecycleBinService.parse` now recognizes it the same way, so an anonymous/guest-session draft's owner is treated as absent rather than a bogus real-looking identity that made `isVisible` fail closed for every admin.

**Finding 4 (LOW/MEDIUM, SCOPE_GLOBAL/SCOPE_USER binary) — judgment call, no behavior change.**
Went with the smaller option: tightened `AutoSaveRecycleBinEntryProjection.SCOPE_GLOBAL`'s javadoc to be explicit that it really means "not `SCOPE_USER`" (covers `TEMPORARY_SCOPE` and any other non-user `AssetRepository` scope constant), not a precise match on `AssetRepository.GLOBAL_SCOPE`. Reasoning: this field is informational only and not used for any access-control decision in this area (confirmed by reading every call site), so the added API surface of a third `SCOPE_TEMPORARY` value wasn't justified by a concrete caller need — a doc fix is enough to stop the field from being misread as more precise than it is.

**Finding 5 (LOW, mutationEntered set too early) — fixed, plus a guard test.**
Moved `mutationEntered.set(true)` to after a *confirmed* `true` return from `restoreAutoSaveAssets`, so a clean `false` return (the primitive's own internal duplicate check refusing strictly before any `repository.setSheet` call) is no longer misreported as "unknown state". To avoid losing real coverage, the call is wrapped in its own try/catch: if `restoreAutoSaveAssets` itself throws (state genuinely unknown, could have partially mutated), `mutationEntered` is still set true before the exception propagates. Added a regression test for both the benign-refusal case and a guard test proving a genuine mid-call exception still triggers the "unknown state" rollback-failure path.

**Tests.** This package had zero test coverage before this PR (also flagged by the reviewer). Added `core/src/test/java/inetsoft/web/admin/ai/autosave/AutoSaveRecycleBinChangesetApplyServiceTest.java` (4 tests: findings 1, 2, and the finding-5 pair) and `AutoSaveRecycleBinServiceTest.java` (3 tests: finding 3 plus regression guards for the existing `_NULL_` sentinel and a real owner identity). Followed `RecycleBinChangesetApplyServiceTest`'s own structural precedent (Mockito + `MockedStatic` for `Tool`/`Audit`/`AutoSaveUtils`, `@Tag("core")`), matching the pattern already used for `AutoSaveUtils` static mocking in `AutoSaveControllerTest`/`VSEventUtilAutoSaveTest`.

**Verification.**
- `mvnw.cmd -pl core compile -o -q` — clean, no errors.
- `mvnw.cmd -pl core test-compile -o -q` — clean.
- `mvnw.cmd -pl core test -o -Dtest=AutoSaveRecycleBinChangesetApplyServiceTest,AutoSaveRecycleBinServiceTest` — `Tests run: 7, Failures: 0, Errors: 0`.

**Scope.** All changes are confined to the `autosave` package plus its new test directory. `RollbackFailure.java` was NOT touched (finding 1 was resolved by reusing the existing per-area `mergeAdvisory`/advisories-map pattern instead).

NeverStopFMH:
## Round 2 review

**Head commit reviewed:** `49fb4fa9d` (confirmed matches PR's current `headRefOid`)
**Range since round 1:** `0326823a6..49fb4fa9d` (5 files, +546/-18: `AutoSaveRecycleBinChangesetApplyService.java`, `AutoSaveRecycleBinEntryProjection.java`, `AutoSaveRecycleBinService.java`, and two new test files)

### What I checked
- Read the full round-1→round-2 diff directly (not just the fixer's summary), plus the complete current `AutoSaveRecycleBinChangesetApplyService.java`, `AutoSaveRecycleBinService.java` and `AutoSaveRecycleBinEntryProjection.java`.
- Traced Finding 1's data flow by hand: `rollbackOne` puts the non-compensable-restore advisory into `advisories.put(undo.key, advisory)`; `apply()` then builds `finalResults = results.stream().map(o -> mergeAdvisory(o, rollbackAdvisories.get(o.property())))` and returns `finalResults` (not `results`) in both the `ROLLED_BACK` and `ROLLBACK_FAILED` branches. `undo.key`/`o.property()` are the same plan-change key, so the merge lands on the right entry.
- Independently opened `IdentityChangesetApplyService.java`, `ProviderChangesetApplyService.java`, `LicenseChangesetApplyService.java` and confirmed all three already have their own `rollbackAdvisories` map + identical `mergeAdvisory` shape (e.g. Identity's is at lines ~154-163, 785). This is a genuine structural match, not a superficially-similar invention.
- Traced Finding 2: `applyRestore` now computes `liveSheetCreated = assetRepository.containsEntry(created)` independently of `draftCleaned = !autoSaveRecycleBinService.exists(id, user)`; `Undo` is queued under `if(liveSheetCreated)`, and a failed cleanup now surfaces as `cleanupAdvisory` instead of masking the outcome/rollback tracking. Checked the `AssetEntry` construction (`GLOBAL_SCOPE`, type, assetName, actingUser) used for this check is identical to `wouldCollide`'s and the pre-existing rollback check's construction — confirmed this predates round 2 (present unchanged in `0326823a6`, just resequenced earlier in the method), so no new "wrong AssetEntry" gap was introduced.
- Checked Finding 3's `"anonymous"` handling against both real `AutoSaveUtils` call sites (`deleteUserAutoSaveFiles` line ~289, and the org-scoped listing helper) — same lowercase literal, same `.equals()` semantics; the new `parse()` check matches exactly.
- Checked Finding 4: confirmed `scope`/`SCOPE_GLOBAL`/`SCOPE_USER` are used only in `project()`'s diff string and the read-model result (`GetAutoSaveRecycleBinEntryResult`) — never in `isVisible` or any access-control path. I could not construct a concrete problem scenario; treating this as accepted judgment call, not open.
- Checked Finding 5: `mutationEntered.set(true)` only happens (a) inside the `catch(Exception e)` around `restoreAutoSaveAssets`, or (b) after the `if(!restored) throw` guard — so a clean `false` return never sets it, an exception always does. Matches the claim.

### Test probes (actually executed, not just read)
Ran `mvnw.cmd -pl core compile -o -q` (clean) and `mvnw.cmd -pl core test -o -q -Dtest=AutoSaveRecycleBinChangesetApplyServiceTest,AutoSaveRecycleBinServiceTest` — **7 tests, 0 failures** (4 + 3), matching the fixer's claim, on this exact HEAD.

Then, on top of that clean HEAD, I made two temporary one-line reverts (each reverted, recompiled/retested, then restored — `git diff` on the file is byte-identical to HEAD afterward):
1. **Finding 2 probe:** changed the Undo-queuing gate from `if(liveSheetCreated)` back to `if(draftCleaned)` (the original bug's gate). Result: `applyRestoreQueuesUndoWhenLiveSheetCreatedEvenIfDraftCleanupFails` **failed** — `Wanted but not invoked: assetRepository.removeSheet(...)`, i.e. the Undo was never queued, exactly reproducing the original orphaning bug. Reverted the probe; suite is clean again (4/4).
2. **Finding 1 probe:** changed the two rollback-path `return` statements to build the result from `results` instead of `finalResults` (skipping the advisory merge). Result: `rollbackOfNonCompensableRestoreSurfacesDestroyedAssetAdvisoryOnTheOutcome` **failed** at the `assertTrue(outcome1.advisory().contains("no live inverse"))` line — `expected: <true> but was: <false>`, exactly reproducing the original silently-dropped-advisory bug. Reverted the probe; suite is clean again (4/4), and `git diff` on the file confirms zero net change.

Both tests assert real state (advisory string content via `.contains(...)`, and `verify(assetRepository).removeSheet(...)` / `verify(..., never())`), not just "no exception thrown" — confirmed genuine regression coverage, not theater.

### Status of the 5 round-1 findings
1. **HIGH (rollback advisory silently dropped)** — **Fixed**, verified by code trace + test probe above.
2. **HIGH (Undo gated on wrong signal, orphaning live sheet)** — **Fixed**, verified by code trace + test probe above.
3. **MEDIUM ("anonymous" owner sentinel not recognized)** — **Fixed**, verified against both real `AutoSaveUtils` call sites + dedicated test.
4. **LOW/MEDIUM (`TEMPORARY_SCOPE` collapsed into "global")** — **Accepted as judgment call**: confirmed via grep that scope is informational-only (never used for access control), javadoc now documents this precisely. No concrete problem scenario found.
5. **LOW (`mutationEntered` set before internal collision re-check)** — **Fixed**, verified by code trace; also covered by a dedicated regression test plus a "still-throws" counter-test guarding against over-narrowing.

### New findings from the broader sweep
None. I read the complete diff (all 5 files, not just the parts the fixer's summary called out) — the `AutoSaveRecycleBinEntryProjection.java` change is javadoc-only, and the `AutoSaveRecycleBinService.java` change is exactly the one-line Finding-3 fix, nothing else slipped in.

### Verdict: **approved**

All 5 round-1 findings are resolved (4 fixed-and-verified, 1 accepted judgment call with no counter-scenario found), the two new test files assert real behavior and were confirmed to actually catch their target regressions by direct probe, and the build/tests are green on this exact HEAD.

PR #5357, opened 2026-09-18
fix(wiz): prevent script library function from self-listing as its own dependent (#76765 SSL-002)

Zxm-happy:
## Summary
- Redmine #76765 SSL-002: deleting a Script Library function that had ever been updated could refuse the delete with a "used by" message that listed the function itself as one of its own referencers, alongside genuine referencers.
- Root cause: `LocalDependencyHandler.updateScriptDependencies(String, AssetEntry, boolean, boolean)` scans a script's text with a declaration-blind tokenizer (`FunctionIterator`/`ScriptIterator`) that matches any `identifier(` shape as a reference, with no way to tell a function's own declaration header apart from a real call site. Updating a function's own body necessarily still contains `function <name>(`, so the function's own header was recorded as a reference to itself, and a genuine self-loop dependency edge was persisted to `DependencyStorageService` (not a display-time artifact).
- Fix: skip recording (add path) when the matched identifier resolves to the same `AssetEntry` as the entry being scanned, in the shared `LocalDependencyHandler` method itself — this covers both callers of the method: `ScriptLibraryController.update()` and StyleBI's own Composer script-editor save path, `OpenScriptController.saveScript` → `ScriptService.updateScriptDependencies`.

## Test plan
- [x] Added `updateScriptDependencies_doesNotRecordAFunctionAsItsOwnDependent` and `updateScriptDependencies_stillRecordsAGenuineReferenceToAnotherFunction` to `core/src/test/java/inetsoft/uql/asset/LocalDependencyHandlerTest.java` — the second asserts genuine external dependents are still recorded, so the fix can't be satisfied by a guard that also breaks real dependency tracking.
- [x] `mvn -pl core test -Dtest=LocalDependencyHandlerTest` — `Tests run: 5, Failures: 0, Errors: 0, Skipped: 0`


Zxm-happy:
Review findings (posted as a plain comment — self-approve via `gh pr review --approve` is rejected since PR author and this GitHub account are the same identity, same workaround as SSL-003).

Reviewed against the confirmed diagnosis/refute trail for SSL-002.

Verified independently (not just trusted the docs):
- The guard `if(scriptEntry.equals(entry)) return;` sits inside the shared `LocalDependencyHandler.updateScriptDependencies(String, AssetEntry, boolean, boolean)` listener, before `id`/`addScriptDependencyToFile`/`removeScriptDependencyToFile` — i.e. before the write, not a narrower fix in `ScriptLibraryController`. Grepped all `updateScriptDependencies(` call sites in the worktree: both real callers (`ScriptLibraryController.update()` and `OpenScriptController.saveScript` -> `ScriptService.updateScriptDependencies`) funnel into this exact method, so both are covered by one fix.
- Read `AssetEntry`'s constructor and `equals()` directly: both `scriptEntry` (built in the listener from `token.val`) and `entry` (the controller's own `AssetEntry`) use `AssetRepository.COMPONENT_SCOPE`, and the constructor only assigns `this.user` when `scope == USER_SCOPE` — so `user` is never populated on either side regardless of what's passed in. `equals()` compares `path/user/orgID/type/scope`; all five line up when the scanned identifier textually matches the entry's own name. Not a subtly-wrong comparison (no reversed variables, no identifier-string-vs-equals() mixup).
- Ran `LocalDependencyHandlerTest` myself, non-piped: `mvn -pl core -Dtest=LocalDependencyHandlerTest test`, real exit code 0 (checked separately from any pipe), 5/5 tests pass.
- Test 1 (`...doesNotRecordAFunctionAsItsOwnDependent`) asserts `verify(dependencyStorageService, never()).put(eq(selfEntry.toIdentifier()), any())` — a real assertion that the write never happened, not just "no exception."
- Test 2 (`...stillRecordsAGenuineReferenceToAnotherFunction`) asserts the helper function's dependents list still contains the caller's entry — proves the guard can't be satisfied by a version that also breaks real dependency recording. Good regression coverage on both sides of the fix.
- The `XPrincipal` mock (instead of plain `Principal`) is a reasonable test-setup necessity, not a hack masking a prod defect: `OrganizationManager.getCurrentOrgID()` genuinely hard-casts the thread principal to `XPrincipal` in production too, so any real caller already needs an `XPrincipal` on the thread context; the test mock matches production reality rather than working around a bug.
- The two out-of-scope observations (possibly-reversed old/new arg order in `OpenScriptController.java:68`, dead `functions` Vector) are correctly left untouched — neither affects SSL-002's mechanism or fix correctness.
- Diff is tight: only `LocalDependencyHandler.java` (+5) and its test file (+81) — no unrelated changes.

CI: build/Semgrep Scan/Semgrep OSS/license-cla all SUCCESS, PR is MERGEABLE.

No blockers. Proceeding to squash-merge.

PR #5358, opened 2026-09-18
fix(wiz): add rename endpoint for Script Library functions (#76765 SSL-005)

Zxm-happy:
## Summary
- Redmine #76765, SSL-005: the wiz Script Library CRUD surface had no way to rename a function.
- `LibManager.renameScript()` alone only relocates the registry key and patches the derived signature (`ScriptLogicalLibrary.renameEntry`) — it never rewrites the function's own declaration text, which is what `GraalJavaScriptEngine.installLibraryFunctions` actually binds as the JS global. Left alone, a naive rename would leave every caller (already rewritten by `RenameTransformHandler`, asynchronously, to call the new name) broken with a `ReferenceError` — worse than not renaming at all.
- New `POST /api/wiz/v1/script-library/{name}/rename` endpoint: captures the old text, calls `renameScript`, then rewrites the captured text with the existing `Util.renameScriptDepended` utility (the same one already used for callers' scripts) and persists it under the new name via `setScript`. Also adds a uniqueness check on `newName`, since nothing in the `rename()` call chain otherwise checks it (would silently clobber an existing function of that name).
- Note: `Util.renameScriptDepended` is a dot/bracket-bounded substring replace, not an identifier-exact-match rename — documented as a known pre-existing limitation in the code comment and test, not something this fix newly introduces or is expected to solve.
- Pairs with the wiz-side `rename_script_library_function` tool: inetsoft-technology/stylebi-wiz#(see linked PR).

## Test plan
- [x] `ScriptLibraryControllerTest` (11 tests, was 7) — `mvn -o test -Dtest=ScriptLibraryControllerTest`, exit code 0, `Tests run: 11, Failures: 0, Errors: 0`.
- [x] New regression tests: rejects rename onto an existing name (and confirms the existing entry isn't clobbered), rejects rename of a non-existent old name, rejects a blank new name, and confirms the persisted text after rename actually starts with `function <newName>(` (the load-bearing regression this fix is about).


Zxm-happy:
Reviewed per docs/teams/2026-09-18-bugs-76765-script-library/bug-ssl-005/05-review-r1.md.

Verified the load-bearing sequence directly in the diff:
- `oldText` is captured via `lib.getScript(name)` BEFORE `lib.renameScript(name, newName)` is called (getting the old key's text would return null after the rename).
- `Util.renameScriptDepended(name, newName, oldText)` rewrites the captured text, and the rewritten text is persisted under the NEW name via `lib.setScript(newName, newText)` — not the stale old-name-declaring text.
- Uniqueness check (`lib.getScript(newName) != null`) happens before any mutation (before `renameScript` is called), so an occupied `newName` is rejected without touching the old entry's registry slot.
- Regression test `rename_rewritesTheFunctionsOwnDeclarationText` asserts `verify(fixture.lib).setScript("newFn", "function newFn(x) { return x * 2; }")` — this fails under the naive (pre-fix) implementation that only calls `renameScript`, so it actually exercises the fix, not just "no exception thrown".
- Javadoc and test comment describe `Util.renameScriptDepended` accurately as a dot/bracket-bounded substring replace, not identifier-exact-match — matches the refuter's requested correction.
- `build`/Semgrep/CLA checks all green, no conflicts.

No blockers. Approving via comment since `gh pr review --approve` rejects self-approval on this account (same workaround as SSL-002/003). Squash-merging now.

PR #5361, opened 2026-09-18
fix(wiz): SSL-001 resync warning on Script Library CRUD (Redmine #76765)

Zxm-happy:
## Summary
- Redmine #76765 SSL-001: a Script Library function is installed as a global JS binding only once, when a viewsheet session's script runtime (`GraalJavaScriptEnv`/GraalJS `Context`) is first built (`GraalJavaScriptEngine.installLibraryFunctions()`). Nothing rebuilds that binding set for an already-open session when `LibManager` changes, so `execute_script`/`run_script_live` against a session opened before a create/update/delete kept returning stale results (`ReferenceError` for a brand-new function, the old body for an updated one, a still-callable orphan for a deleted one) with no indication of why.
- `ScriptLibraryController.create`/`update`/`delete` now return a `resyncWarning` string telling the caller that already-open sessions need to resync (e.g. `refresh_viewsheet`) to see the change. `read` is unaffected (`resyncWarning: null`).
- `delete()`'s return type changes from `void` to a small `DeleteScriptLibraryFunctionResult(String resyncWarning)` record — its only caller is wiz's own `plugin/composer/src/tools/scriptLibraryTools.ts` (see companion wiz-repo commit), which has been updated to read and pass through the field.

## Why this direction, not an automatic runtime fix
The diagnosis's direction (a) (auto-invalidate/rebuild the affected session's script scope when `LibManager` changes) looked feasible via a per-instance `LibManager` listener on `GraalJavaScriptEnv` calling its existing `reset()` (mirrors the established `LibScriptable` pattern, and `reset()`/`exec()` already share a lock so it's race-safe). However, further reading found `with(__scope__){...}`-wrapped script execution means an *implicit global* a script creates on first use (e.g. an accumulator set in `onLoad` and read by a later button script) lives in the GraalJS Context directly, not in `GraalJavaScriptEnv`'s `vars` map that `reset()` preserves — so an automatic reset triggered by an unrelated user's library edit could silently wipe that state for every concurrently open session in the org. This is exactly the class of bug `ViewsheetSandbox.resetRuntime()` was historically hardened against (it deliberately does not clear `scope`, per its `bug1429296619632` comment). Without live verification of how much this matters in practice, this PR takes the safer, smaller-blast-radius option: surface the caveat in the response instead of silently rebuilding a shared, high-traffic script runtime.

## Known open items (not resolved by this PR — flagged for the post-merge live verification pass)
- The addendum contradiction (SSL-001 server logs showing `get_viewsheet_image`'s export path returning a fresh `ReferenceError` while `execute_script` on the same session simultaneously returned a stale result) is **not explained**. A separate refute pass showed the export and interactive paths in fact resolve to the same `RuntimeViewsheet`/`ViewsheetSandbox`/`ViewsheetScope`/Context, which makes the original observation harder to explain, not easier — this PR does not attempt to resolve that.
- The exact mechanism by which `refresh_viewsheet`/`save_viewsheet` actually resync a session's script bindings today was not pinned down from source in this investigation. The `resyncWarning` text names `refresh_viewsheet` as an example based on triage's behavioral observation, not on a mechanism confirmed in this PR.
- This PR does not fix the underlying staleness — it only makes it visible in the response.

Full analysis: `docs/teams/2026-09-18-bugs-76765-script-library/bug-ssl-001/03-fix.md` in the `wiz` repo (diagnosis/refute/fix records).

## Test plan
- [x] `./mvnw -q -pl core test -Dtest=ScriptLibraryControllerTest -DfailIfNoTests=true` — 11/11 pass (4 new: `create_returnsAResyncWarning`, `update_returnsAResyncWarning`, `delete_returnsAResyncWarning`, `read_doesNotReturnAResyncWarning`).
- [ ] Live verification (browser/MCP) deferred to a post-merge pass, per this repo's established pattern for this workstream.


Zxm-happy:
Reviewed (adversarial pass, review-workflow.md). No blocking issues.

Confirmed:
- Diff scope is limited to `ScriptLibraryController.java` response-body shape (new `resyncWarning` field on `ScriptLibraryFunctionDetail`, `delete()`'s return type changed `void` -> `DeleteScriptLibraryFunctionResult`). No script-runtime file (`GraalJavaScriptEngine`/`GraalJavaScriptEnv`/`ViewsheetSandbox`/`ViewsheetScope`) touched -- matches the stated decision to go with direction (b) after finding the implicit-global-loss risk in (a).
- `delete()`'s return-type change: verified via the class header comment and a repo-wide search that this is a wiz-only, JWT-authenticated `/api/wiz/v1/script-library` endpoint, not proxied by StyleBI's internal Composer SPA controllers, and its only real caller is wiz's `plugin/composer/src/tools/scriptLibraryTools.ts` (already updated in companion PR #2660). No other caller found.
- Warning text (`RESYNC_WARNING`) says "may not see this change ... until they resync (e.g. via refresh_viewsheet)" -- appropriately hedged as an example, not a guaranteed fix, consistent with the resync-trigger mechanism never having been traced in source.
- `read_doesNotReturnAResyncWarning` test exists and asserts `null`; the three CRUD tests assert the field is non-null.
- PR description discloses the open items (addendum contradiction unresolved, resync mechanism unconfirmed, staleness itself not fixed) rather than overclaiming a complete fix.
- Build/Semgrep/CLA checks all green, no conflicts.
- Commit message in English, no unrelated changes, no secrets.

Self-approve isn't available since the PR author and this review identity are the same GitHub account -- recording findings as a comment, then squash-merging directly (same workaround used for SSL-002/003/004/005).

PR #5367, opened 2026-09-18
fix(wiz): add live worksheet-table CSV export endpoint (Redmine #76753, WBS-057)

NeverStopFMH:
## Summary

Redmine #76753, sub-finding WBS-057: no `mcp__composer-chat__*` tool could export a worksheet
table's live data. The only existing worksheet-table CSV export
(`RawDataController.exportWorksheetTableToCsv` / `RawDataService.writeWorksheetTableCsvStream`)
resolves a **persisted asset by identifier**, not a live agent session — so it silently returns
stale data for any worksheet edited since its last save, and has no callable identifier at all
for a worksheet that has never been saved (a `TEMPORARY_SCOPE` asset until `save_worksheet`).

This PR adds a genuine session-scoped alternative, modeled on the existing `/preview` endpoint
(which already resolves `sessionToken` → live `RuntimeWorksheet` → `AssetQuerySandbox` →
`TableLens`):

- `RawDataService.writeLiveWorksheetTableCsvStream(RuntimeWorksheet, table, OutputStream)` — takes
  a live `RuntimeWorksheet` instead of a persisted `wsId`, finds the target `TableAssembly` by
  `getAbsoluteName()`, calls `box.getTableLens(name, AssetQuerySandbox.RUNTIME_MODE)`, and reuses
  the existing `writeCsvContent`/`writeCsvCell` CSV-writing helpers unchanged.
- `WorksheetAgentController`: new `GET /api/wiz/v1/agent/worksheet/{sessionToken}/export?table=X`,
  next to `/preview`, using the same `requireWholeSheetSession` + `editService.resolve` auth/scope
  gate `/preview` already uses. `WorksheetAgentController` did not previously have a
  `RawDataService` dependency — added as a new constructor parameter (mechanical DI change; every
  test call site building this controller updated accordingly).

## Design decisions worth calling out explicitly

1. **No `VIEWSHEET_TOOLBAR_ACTION`-style permission check on the new endpoint.**
   `ViewsheetAssemblyAgentController#export` has an extra
   `securityEngine.checkPermission(user, ResourceType.VIEWSHEET_TOOLBAR_ACTION, "Export", ...)`
   gate that this worksheet-side endpoint does not carry over. This is **deliberate, not an
   oversight**: there is no worksheet-side equivalent permission primitive to reuse (grepping the
   codebase for a `WORKSHEET_TOOLBAR_ACTION` resource type turns up nothing). The new endpoint
   instead inherits the same `requireWholeSheetSession`/`editService.resolve` auth/scope every
   other worksheet-agent endpoint (including `/preview`) already uses — no weaker exposure than the
   existing worksheet tool surface, just not identical to the viewsheet side's stricter posture.
   Documented in the new endpoint's own javadoc so a future reviewer diffing the two `/export`
   endpoints doesn't mistake this for a gap to close.

2. **"Auto-save then reuse the existing `RawDataController` endpoint" was considered and
   rejected.** It would force an unwanted persistence side effect (an unsaved worksheet has no
   name to save under without inventing one; an already-saved worksheet would be silently
   overwritten as a side effect of "just export the data"), and it's not even clear the existing
   `save_worksheet` response carries a usable asset identifier in the shape `RawDataController`
   expects. The session-scoped design here avoids all of that by keying off `sessionToken` → live
   `RuntimeWorksheet`, the same way every other worksheet tool already does.

## Companion PR

TypeScript side (new `export_worksheet_table` tool calling this endpoint):
inetsoft-technology/stylebi-wiz#2666 (branch `fix/76753-wbs057`)

## Test plan

- [x] `RawDataServiceTest` — new tests for `writeLiveWorksheetTableCsvStream`: happy path (CSV
      streamed from a mocked live sandbox), sandbox-absent, table-not-found (fails loudly, names
      the table, never touches the sandbox), lens-null.
- [x] `WorksheetAgentControllerTest` — new tests for the `/export` endpoint: streams CSV to a
      mocked `HttpServletResponse` with `text/csv` content type and `Content-Disposition:
      attachment`, and fails loudly (propagates the table-naming exception) on a bad table name.
      Every existing `new WorksheetAgentController(...)` call site (7 in this file, plus
      `WorksheetAgentControllerConvertToEmbeddedTest` and `WorksheetScriptControllerTest`) updated
      for the new constructor parameter.
- [x] Ran targeted: `./mvnw test -pl core -Dtest=RawDataServiceTest,WorksheetAgentControllerTest,WorksheetAgentControllerConvertToEmbeddedTest,WorksheetScriptControllerTest,WorksheetPreviewServiceTest,ViewsheetAssemblyAgentControllerTest` — 338 tests, all passing, no compilation errors.
- [ ] No live StyleBI session was available this pass — not live-verified end-to-end.


NeverStopFMH:
Companion TypeScript PR (new `export_worksheet_table` tool calling this endpoint): https://github.com/inetsoft-technology/stylebi-wiz/pull/2666

PR #5368, opened 2026-09-18
feat(wiz): add schedule task export/import (Redmine #76719 Gap 3)

NeverStopFMH:
## Summary
- Closes the third and final gap Bug #76719 documented: no way to export/import Schedule Task definitions through admin-chat, mirroring the Repository asset export/import area's own stage/preview/apply/planHash/taskToken discipline.
- New `inetsoft.web.admin.ai.schedule` classes: `ScheduleTaskTransferService` (export + in-memory staging cache), `ScheduleTaskImportChangePlanService` (preview), `ScheduleTaskImportChangesetApplyService` (apply/rollback, reusing `ScheduleTaskService.sanitizeConditions`/`sanitizeAction` — the same sanitization `AdminScheduleGateway`'s own `create` verb already applies), `AdminScheduleTransferController` (REST endpoints), plus the request/result DTOs.
- Import is fully compensable: a create's rollback removes the task; an overwrite's rollback restores the captured original task exactly.

## Bug found and fixed during live testing
`ScheduleTaskTransferService`'s staging cache holds one `ScheduleTask` instance per `stagingToken` for its whole 30-minute lifetime, shared across every preview/apply call against that token. The apply path was mutating that shared instance in place (linkURI rewrite + sanitization) instead of cloning it first — live-confirmed: a `preview` call issued *after* an `apply` against the same `stagingToken` showed the apply's own linkURI rewrite already baked into `proposedValue`, which a read-only preview must never do. Fixed by cloning the staged task before mutating in `ScheduleTaskImportChangesetApplyService.applyOne`.

## Test plan
- [x] `mvnw -pl core compile` — clean
- [x] Live-tested via admin-chat MCP tools only (no side-channel DB/backend access), end to end:
  - export a real task to XML, stage/parse it back
  - collision refusal without `overwrite`, successful overwrite with field-level verification via `get_schedule_task`
  - brand-new-task import path (`risk: low`, no signoff required)
  - resolve-time drift rejection (both the field-named `IllegalArgumentException` refusal and the outer `planHash`/`taskToken` conflict check) — confirmed nothing is applied on either path
  - invalid XML / non-XML input refused loud, naming the field
  - unknown `stagingToken` / unstaged `taskId` return structured not-found errors
- [x] Regression-verified the staging-cache mutation bug fix by re-previewing the same `stagingToken` after an apply and confirming `proposedValue` no longer drifts


NeverStopFMH:
## Review round 1

**Head commit reviewed:** `5ea2230ff437cab8969c6d2687e2281814052c2a`

**What was checked:** Full diff (`gh pr diff 5368`), all 14 new files under `core/src/main/java/inetsoft/web/admin/ai/schedule/`, and the sibling/precedent implementations this PR claims to mirror (`ScheduleChangePlanService`, `ScheduleChangesetApplyService`, `AdminScheduleGateway`, `ScheduleTaskService`, `AdminAssetImportApplyService`, `ImportTaskController.updateTaskInfo`, `SheetSessionService`). For each changed piece of logic, named the invariant it claims and tried to break it against sibling branches (create vs overwrite, resolve vs apply), other entry points (the 4 REST endpoints), boundary/shape variants (empty batch, malformed XML, duplicate taskId, no-owner task), repetition/ordering (double-stage, replayed apply), and concurrency (apply racing preview). Also assessed test coverage against the surrounding package's own convention, and the single-node in-memory staging cache's cluster-deployment implications.

**Previous findings:** N/A (round 1).

### Findings

**1 — BLOCKING:** The PR's headline "bug found and fixed" (staging-cache mutation bleeding from apply into a subsequent preview) is **not actually fixed** for any task that has actions. `ScheduleTaskImportChangesetApplyService.applyOne` (core/src/main/java/inetsoft/web/admin/ai/schedule/ScheduleTaskImportChangesetApplyService.java:220) clones the staged `ScheduleTask` before mutating it, but `ScheduleTask.clone()` (core/src/main/java/inetsoft/sree/schedule/ScheduleTask.java:1159-1174) only does `Vector.clone()` on `conds`/`acts` — a shallow copy. The `ScheduleAction`/`ScheduleCondition` *objects themselves* are still the same shared instances referenced by the cached entry in `ScheduleTaskTransferService`'s staging map. `updateTaskLinkUri` (line 266-272, via `AbstractAction.setLinkURI`) and `sanitizeAction` (called at line 242) both mutate the action object's fields directly — so the exact same "apply's rewrite leaks into a subsequent preview's `proposedValue`" symptom the PR claims is fixed still reproduces, just one level down (action fields instead of task fields), for any staged task with a `ViewsheetAction`/similar. The codebase's own existing idiom for a true deep copy — `ScheduleTask.copyScheduleTask` (ScheduleTask.java:1228-1246, "Creates a deep copy for this task") — explicitly re-clones every action via an XML round-trip after `.clone()`, precisely because `.clone()` alone isn't deep enough; that method is `protected static` and unreachable from this new package, but the fix needs to replicate its effect. Compounding factor: `applyOne`'s mutation runs under `APPLY_LOCK`, but `resolve()`/`preview` take no lock at all, so a concurrent preview can race unsynchronized with an in-flight apply's in-place field write, not just observe stale-but-consistent state.

**2 — IMPORTANT:** `ScheduleTaskTransferService.stage()` (lines 190-200) wraps the outer XML-document parse in a clean `IllegalArgumentException`, but the per-`<Task>` element loop calls `task.parseXML(element)` with no try/catch. `ScheduleTask.parseXML` does unguarded numeric parsing (`Long.parseLong` on `lastModified`/`startDate`/`endDate`, ScheduleTask.java:1417-1433), so a single malformed attribute in one `<Task>` element throws a raw, unwrapped `NumberFormatException` that isn't one of `AdminScheduleTransferController`'s handled exception types — it surfaces as an opaque 500 instead of a field-named 400, inconsistent with this same class's own adjacent handling and the plugin's fail-loud convention. It also fails the entire multi-task batch with no signal about which task/attribute was at fault.

**3 — IMPORTANT:** This PR adds zero unit tests. Every functionally-equivalent class in the *same* package (`ScheduleChangePlanServiceTest`, `ScheduleChangesetApplyServiceTest`, `ScheduleConfigChangePlanServiceTest`, `ScheduleFolderChangePlanServiceTest`, `ScheduleFolderChangesetApplyServiceTest`, `ScheduleXmlProjectionTest`, `AdminScheduleControllerTest`, `AdminScheduleGatewayTest`, `AdminScheduleFolderGatewayTest`) has a test, and this `*ChangePlanServiceTest`/`*ChangesetApplyServiceTest`/`*ControllerTest`-per-class pattern holds across essentially every other `inetsoft.web.admin.ai` subpackage. The one precedent with no tests is the cross-package repository-import area this PR cites as its architectural mirror, so the choice matches that one precedent but breaks the much stronger local convention around it. Given that Finding 1 is exactly the class of aliasing bug a mock-based unit test (stage → apply → assert the cached object's fields are untouched; or apply → preview → assert `proposedValue` unchanged) would have caught mechanically before it reached the PR description as "fixed," this is a real gap, not just a nice-to-have.

**4 — MINOR:** `ScheduleTaskImportChangePlanService.java:35-41`'s javadoc claims the sibling `ScheduleChangePlanService` needs "an 8-field digest layered on top of the server's own plan hash." Reading that sibling directly, it uses the same plain `TaskAuditToken.issue(planHash, task)` this PR uses — no extra digest exists in that file, and no other file in `inetsoft.web.admin.ai` mentions an "8-field" anything. Not a functional defect, but the stated rationale references something that doesn't exist in the file it names.

**5 — NIT:** `ScheduleTaskStageRequest`/`ScheduleTaskImportPlanRequest`/`ScheduleTaskImportChangeRequest` javadocs document routes under `/api/wiz/v1/admin/schedule/import/...`, but `AdminScheduleTransferController` actually maps `/api/wiz/v1/admin/schedule/transfer/...`. Cosmetic doc drift.

### Invariants checked, including attempted counter-examples that did NOT break

- Duplicate taskId across `changes[]` in one plan request: `resolveOne`'s `seenKeys` check refuses cleanly — holds.
- Multi-entry batch, partial failure mid-apply: rollback undoes only the verified/applied entries, in reverse — holds (correct all-or-nothing).
- Replay of the same apply twice: create-verb replay is refused by fresh re-resolve ("already exists"); overwrite-verb replay re-applies deterministically — holds.
- Empty task list / empty changes list on export()/resolve(): both refuse cleanly — holds.
- Task with no owner on CREATE-undo rollback path: null-safe ternary on `getOwner()` — holds.
- Cluster/multi-node: the staging map is single-node in-memory, but this is disclosed in `ScheduleTaskTransferService`'s own `Staged` javadoc, which is *more* documentation than the cited precedent `SheetSessionService` carries for the same limitation (no class-level javadoc on scope at all there) — not a newly-introduced silent regression, so not raised as a blocking finding; a minor non-blocking suggestion is that the "unknown or expired" staging-token error could hint at node-routing as a possible cause in clustered deployments, since it's currently indistinguishable from genuine expiry.

NeverStopFMH:
Round-1 review triage, applied in 658da228b:

**Finding 1 (BLOCKING) — accepted, fixed.** Confirmed by reading the actual code: `ScheduleTask.clone()` does `Vector.clone()` on `conds`/`acts`, which copies the Vector container but not the `ScheduleCondition`/`ScheduleAction` elements inside it — so `updateTaskLinkUri`'s `AbstractAction.setLinkURI(...)` call (and `ScheduleTaskService.sanitizeConditions`/`sanitizeAction`, which also mutate `TimeCondition`/`ViewsheetAction` fields in place) were still corrupting the object shared with `ScheduleTaskTransferService`'s staging cache.

Fix: `applyOne` now deep-copies the staged task via an XML round-trip (`ScheduleTask.writeXML` → `parseXML`) before any mutation, isolating both actions AND conditions from the cached instance. I did **not** reuse `ScheduleTask.copyScheduleTask`/`copyScheduleAction` as suggested, for two reasons: (1) it's `protected static`, unreachable from `inetsoft.web.admin.ai.schedule` without widening its accessibility, and (2) more importantly, `copyScheduleTask` only round-trips **actions** after a shallow `clone()` — it does **not** deep-copy conditions, so it would not have actually closed the gap for `sanitizeConditions`'s in-place `TimeCondition` mutations. The XML round-trip I added instead reuses the exact same `writeXML`/`parseXML` pair `ScheduleTaskTransferService.stage()` already uses to build a `ScheduleTask` from an uploaded file, and the same per-object round-trip idiom `copyScheduleAction` itself uses — just applied to the whole task, so conditions are covered too. Added `ScheduleTaskImportChangesetApplyServiceTest`, which I verified fails against the pre-fix code (`expected: <http://source-server/> but was: <http://different-server/>`) and passes with the fix.

On the flagged missing lock around `resolve()`/preview racing a concurrent in-flight apply: I'm treating this as a **separate, out-of-scope concern** rather than fixing it here. `resolve()` is a read against the SAME cached staged object `applyOne` now deep-copies before mutating, so with the Finding-1 fix, a concurrent preview can no longer observe a torn/mutated write from an in-flight apply — the specific hazard the finding was tracing) is closed as a side effect. A *general* lock around preview to serialize it with apply would be a broader concurrency-model change (this area's own preview is stateless/read-only by design, matching every other area's own preview endpoint, none of which take `APPLY_LOCK`) that I don't think belongs in a "minimal fix" PR-review pass. Flagging for a maintainer/follow-up ticket if this is judged worth generalizing across all mutating areas at once, rather than singling this one out.

**Finding 2 (IMPORTANT) — accepted, fixed.** Confirmed `ScheduleTaskTransferService.stage()`'s per-`<Task>` `task.parseXML(element)` call had no try/catch, and `ScheduleTask.parseXML` does unguarded `Long.parseLong` on `lastModified`/`startDate`/`endDate`. Wrapped the call, converting any parse failure into a field-named `IllegalArgumentException` naming the offending task (by `name` attribute, falling back to element index if blank). Added `ScheduleTaskTransferServiceTest` covering both a malformed `<Task>` attribute and the pre-existing outer-document-parse-failure path.

**Finding 3 (IMPORTANT) — accepted, fixed.** Added `ScheduleTaskImportChangesetApplyServiceTest` (Finding-1 regression, described above) and `ScheduleTaskTransferServiceTest` (Finding-2 regression), following the existing package's Mockito/MockedStatic conventions (`ScheduleChangesetApplyServiceTest`/`ScheduleChangePlanServiceTest`).

**Finding 4 (MINOR) — accepted, fixed.** Confirmed by reading `ScheduleChangePlanService.java`: there is no "8-field digest" anywhere in it or the codebase — that phrase doesn't appear outside the one javadoc that claimed it. `ScheduleChangePlanService` and `ScheduleTaskImportChangePlanService` both just SHA-256-hash their own `PlanChange` list via `TaskAuditToken`, no extra digest layer. Corrected `ScheduleTaskImportChangePlanService`'s javadoc to state the real reason no extra digest is needed here: `ScheduleXmlProjection.project()` already folds the ENTIRE staged task into the hash via a full `writeXML()` serialization, so there's no field gap for `TaskAuditToken` to plug.

**Finding 5 (NIT) — accepted, fixed, expanded.** Fixed the 2 confirmed route-javadoc mismatches (`ScheduleTaskStageRequest`, `ScheduleTaskImportPlanRequest`). `ScheduleTaskImportChangeRequest` doesn't actually carry a route javadoc (nothing to fix there). While auditing all javadoc route strings in the package I found 4 more with the same `/schedule/export` or `/schedule/import/...` vs. actual `/schedule/transfer/...` mismatch that the finding didn't call out: `ScheduleTaskExportRequest`, `ScheduleTaskExportResult`, `ScheduleTaskStagingResult`, `ScheduleTaskImportApplyRequest`. Fixed all of them for consistency, same minimal javadoc-only change.

Compile (`mvnw -pl core compile`) and the full `inetsoft.web.admin.ai.schedule`/`inetsoft.web.admin.schedule`/`inetsoft.sree.schedule` test packages (633 tests) are green after these changes.

NeverStopFMH:
## Review round 2

**Head commit reviewed:** `658da228b632e783a6970a6af577cf2c5d443c7a`
**Range reviewed since round 1:** `5ea2230ff437cab8969c6d2687e2281814052c2a..658da228b632e783a6970a6af577cf2c5d443c7a`

**What was checked:** Re-verified each round-1 finding's fix against the actual current source (not the fixer's claims). For Finding 1 specifically: traced every field mutated by `sanitizeAction`/`sanitizeConditions`/`updateTaskLinkUri` against `ScheduleTask.writeXML()`/`parseXML()` to confirm the new `deepCopy()` XML-round-trip genuinely reconstructs independent objects; independently re-verified (via a throwaway probe, not part of this PR) the fixer's claim that `ScheduleTask.copyScheduleTask()` deep-copies actions but not conditions — confirmed true, so `deepCopy()` is a real improvement over reusing that method. Confirmed `resolveOne` only reads the staged task (via `ScheduleXmlProjection.project()`, itself just `writeXML` + normalization) and never mutates it, so it needs no deep copy. Re-checked the `APPLY_LOCK`/preview-concurrency point: `APPLY_LOCK` already serializes apply-vs-apply; since the cached object is now never mutated anywhere in the apply path, the originally-reported hazard (preview showing an apply's in-place mutation) is now structurally impossible rather than just less likely — reasonable scoping decision not to also lock preview. Compiled (`./mvnw -pl core compile`) and ran the two new test files directly — both pass and are non-tautological.

### Status of round 1 findings

- **Finding 1 (blocking) — fixed.** `deepCopy()`'s XML round-trip is a genuine deep copy for every field the apply path touches. One new low-severity item surfaced during verification (see below).
- **Finding 2 (important) — fixed.** Malformed `<Task>` element now throws a clean `IllegalArgumentException` naming the task (by name or index), correctly mapped to 400 by the controller's existing exception handler. Minor nuance: the message names *which task* failed but not the specific bad attribute (it appends the wrapped exception's raw message) — a partial, not full, field-name, but a real improvement over a raw `NumberFormatException` escaping; not re-raising as a separate finding, noting for the record.
- **Finding 3 (important) — partially fixed.** Both new test files are real and non-tautological. But `ScheduleTaskImportChangesetApplyServiceTest` has only one test method and it exercises only the action/linkURI mutation path — not the condition-mutation path (`sanitizeConditions`), because the mocked `scheduleTaskService` has no stubbing so `sanitizeConditions` is a silent no-op in the test. Verified concretely: this test would still pass if `deepCopy()` were reverted to `copyScheduleTask()` — which is exactly the regression the fixer's own stated rationale for choosing `deepCopy()` over `copyScheduleTask()` was meant to guard against. The one thing that justified this implementation choice is the one thing the regression test doesn't check.
- **Finding 4 (minor) — fixed.** Corrected javadoc verified accurate against `ScheduleChangePlanService`/`TaskAuditToken`/`ScheduleXmlProjection.project()` — no "8-field digest" exists anywhere in the codebase; the new text correctly describes the real reason (full-task XML projection folded into the hash).
- **Finding 5 (nit) — fixed.** Grepped the whole package; all 6 DTOs now say `/transfer/...`, no remaining `/import/...` references.

### New findings (round 2)

**6 — IMPORTANT (test-coverage residual of Finding 3):** Add a test case stubbing `scheduleTaskService.sanitizeConditions` to actually mutate a `TimeCondition` field on the `staged` argument it's given, then assert the *original cached* task in the staging map is unaffected. Without this, a future refactor that silently swaps `deepCopy()` back to a shallower copy would not be caught by the current test suite, even though that's precisely the bug class this whole PR is about.

**7 — MINOR:** `ScheduleTask.writeXML()` does not `Tool.escape()` the `locale` attribute, unlike every other string attribute it writes (`name`, `path`, `description`, `timeZone` all are). This is a pre-existing latent bug in `writeXML` itself, but it's newly *reachable* through this PR's `deepCopy()` call site (the previous `.clone()` never invoked `writeXML`/`parseXML` at all). A staged task whose locale string contained an XML-special character would make `deepCopy()`'s re-parse throw a `SAXException` during apply. Low likelihood in practice (locale strings are machine-generated, `en_US`-shaped), not blocking, but worth a one-line `Tool.escape()` fix while this code path is already open, or an explicit acknowledgment if judged out of scope.

### Verdict: changes requested (non-blocking)

Both new findings are addressable in a small follow-up; nothing here reopens the closed blocking issue.

NeverStopFMH:
## Review round 3

**Head commit reviewed:** `31243c3e61fa0c513e137ed791d66aa6a8a58c33`
**Range reviewed since round 2:** `658da228b632e783a6970a6af577cf2c5d443c7a..31243c3e61fa0c513e137ed791d66aa6a8a58c33`

**What was checked:** Independently verified both round-2 findings' fixes rather than trusting the fixer's report. For Finding 6, traced the object flow end-to-end: confirmed `sanitizeConditions`'s `doAnswer` stub in the new test mutates the exact `ScheduleTask` instance `applyOne` actually passes it (the `deepCopy()` result), and that the test's post-apply assertion reads back the same cached object instance `ScheduleTaskTransferService.requireStagedTask` has held since `stage()` (no read-side copy exists, so this is a faithful check). Did not stop at reading the code — reproduced the fixer's fail-before/pass-after claim myself: temporarily swapped `deepCopy()` for a shallow `task.clone()`, ran the test, confirmed both the new condition-isolation test AND the existing linkURI test fail (proving a shallow clone reopens both the original bug and the new one), then reverted the change exactly and re-confirmed `git status`/`git diff` show no residual edit and both tests pass again. Checked the mock call-ordering the fixer adjusted (`scheduleManager.getScheduleTask` stub sequence) against the real number/order of calls in `resolve()`+`applyOne()` — confirmed it mirrors the real 3-call sequence rather than shortcutting past a check. For Finding 7, confirmed the `Tool.escape(locale)` fix matches the neighboring attributes' escaping style, confirmed the read side (`Tool.getAttribute`) already entity-decodes regardless of write-side escaping so there's no double-decode risk, and grepped all tests for any literal-unescaped-`locale=` assertion that this change could break — none found.

Beyond re-verifying round 2: ran a fresh, not-yet-tried concurrency scenario — two concurrent `apply()` calls against the identical stagingToken+taskId. `APPLY_LOCK` is a single static lock held for the whole `apply()` method body (not scoped per token), so this fully serializes at the method's entry; there is no window for a second call's `requireStagedTask`+`deepCopy()` to observe torn state. Also checked cross-task dependency within one batch: each `applyOne()` commits to the real `ScheduleManager` state sequentially before the next batch entry runs, so a later entry that depends on an earlier one's committed taskId sees real state, not a staging-cache artifact — same sequential-apply/reverse-rollback model already exercised by the sibling `ScheduleChangesetApplyServiceTest`. Ran the full compile (`./mvnw -pl core compile`, clean) and the full relevant test scope (`inetsoft.web.admin.ai.schedule.*Test`, `inetsoft.sree.schedule.*Test`, `inetsoft.web.admin.schedule.*Test` — 634 tests, 0 failures/errors) myself rather than trusting the reported run, and confirmed no stray leftover test artifacts from earlier rounds' throwaway probes remain. Diffed all 3 of this PR's commits (`5ea2230ff`, `658da228b`, `31243c3e6`) via `git show --stat` to confirm the unrelated pre-existing local `defaults.properties` change never got swept into any of them.

### Status of round 2 findings

- **Finding 6 — fixed**, independently re-verified by breaking and re-fixing the invariant directly (not just reading).
- **Finding 7 — fixed**, verified consistent with neighboring escaping style and no test lock-in broken.

### New findings this round

None.

### Full-PR sweep

- Compile: clean.
- Tests: 634 run, 0 failures, 0 errors, across all 3 commits combined.
- Commit hygiene: confirmed clean — no unrelated file in any of the 3 commits.
- All previously-accepted scoping decisions (Finding 2's partial field-naming, the no-lock-on-preview call) re-confirmed as not reopened by this round's diff.
- CI: Semgrep and CLA checks pass; the `build` Action check is still pending as of this review — not a code defect, just not yet finished running.

### Verdict: approved

Zero findings across this round, and both items carried over from round 2 are independently confirmed fixed. All 3 rounds are now clean. Approving as of `31243c3e61fa0c513e137ed791d66aa6a8a58c33`.

(Note: `gh pr review --approve` is expected to fail here since this session's gh identity is the PR's own author — GitHub's self-review restriction — so this verdict is recorded via PR comment instead, per the same accommodation used in rounds 1-2.)

PR #5369, opened 2026-09-18
fix(76759-vcg001): don't NPE opening a chart's Properties dialog after it's renamed

lizhepu:
## Summary
- `ChartPropertyDialogService.getChartPropertyDialogModel0` looked up a chart by `objectId` and immediately called `.getVSAssemblyInfo()` with no null check. If `objectId` doesn't resolve (e.g. the wiz plugin renamed the chart via `set_assembly_properties(name:...)` after a caller's reference to it was taken — Redmine #76759 VCG-001), this NPEs as an unhandled 500 in the live Composer UI instead of degrading gracefully.
- Fix mirrors the existing `TextPropertyDialogService.getTextPropertyDialogModel` convention for the identical "assembly not found" case: return an empty/default model rather than proceeding into a chain of unguarded calls on a null assembly.

## Scope note
This is the defensive half of VCG-001 — confirmed safe and correct regardless of root cause by an earlier adversarial refute pass (`docs/teams/2026-09-17-bug-76759/track-b-vcg001/02-refute.md` in the sibling `stylebi-wiz` repo), which also found that neither of the two originally-suspected client-side staleness scenarios has an obvious source-level gap — the deeper "what exact sequence reproduces this" question is still open pending a live repro (blocked on a Chrome extension outage this run hit), and isn't addressed by this fix. This fix makes the failure mode safe (no crash) regardless of how it's later found to be triggered.

## Test plan
- Compiled and verified locally: `ChartPropertyDialogService.class` regenerated cleanly via `./mvnw -pl core -am compile -o` after the change.
- No new tests, per this repo's established convention for this bug's other fixes.


lizhepu:
**Review verdict: CHANGES REQUESTED — real finding below.** (Posting as a plain comment, not a formal review: this session's `gh` account matches the PR author, so `gh pr review --request-changes` is rejected by GitHub, as on the other PRs in this bug's review loop.)

## What I checked

**1. Early-return placement is correct.** Read the full `getChartPropertyDialogModel0` method, not just the diff hunk. The `chartAssembly == null` check sits immediately after `vs.getAssembly(objectId)` and before the first dereference (`chartAssembly.getVSAssemblyInfo()` on the next line). No other line in the method touches `chartAssembly`/`chartAssemblyInfo`/`vsChartInfo` before that point, so the guard is not bypassable within this method.

**2. No other caller depends on the NPE.** `getChartPropertyDialogModel0` is only invoked from this class's own `@ClusterProxyMethod` wrapper (`getChartPropertyDialogModel`, line 81 via lambda). `ChartPropertyDialogController.getChartPropertyDialogModel` calls the proxy, not `...Model0` directly. Grepped the whole `core` tree — no other call site.

## Real finding: the fix moves the crash to the frontend instead of eliminating it (confidence ~90)

`ChartPropertyDialogModel` has **three** fields with no lazy-init getter: `chartAdvancedPaneModel`, `chartLinePaneModel`, `vsAssemblyScriptPaneModel` (contrast `chartGeneralPaneModel`/`hierarchyPropertyPaneModel`, which self-initialize on get). `new ChartPropertyDialogModel()` in the new early-return leaves all three `null`, and Jackson serializes that as JSON `null`.

`ChartPropertyDialog.getScripts()` (`web/projects/portal/src/app/vsobjects/dialog/graph/chart-property-dialog.component.ts:116`) does:
```ts
return [this.model.vsAssemblyScriptPaneModel.expression];
```
— no optional chaining. And the base `PropertyDialog.ngOnInit()` (`.../composer/dialog/vs/property-dialog.component.ts:58`) calls `this.getScripts()` **unconditionally, immediately on dialog construction**, before the user does anything (not gated behind the script tab or Apply/OK).

So for the exact scenario this PR targets (stale `objectId` after rename), opening the Properties dialog will now throw a client-side `TypeError: Cannot read properties of null (reading 'expression')` in `ngOnInit()` instead of the server returning a 500. That's a different bad outcome, not a safe degrade — the dialog still fails to open, just via an uncaught JS exception instead of an HTTP error.

**Why Text's case doesn't have this problem:** `TextPropertyDialogModel.getClickableScriptPaneModel()` *does* lazy-init a default (`ClickableScriptPaneModel.builder().scriptExpression("").onClickExpression("").scriptEnabled(false).build()`) when null, so `TextPropertyDialog.getScripts()` (`.../composer/dialog/vs/text-property-dialog.component.ts:90`) reading `this.model.clickableScriptPaneModel.scriptExpression` can never see a null `clickableScriptPaneModel`. The chart model's `vsAssemblyScriptPaneModel` has no equivalent lazy default — the two "empty model" shapes are not actually equivalent, despite the code comment's claim that this mirrors Text's convention.

**Suggested fix:** in the `chartAssembly == null` branch, either populate `vsAssemblyScriptPaneModel` with an empty default (matching the shape `VSAssemblyScriptPaneModel.builder()` would otherwise produce further down in this same method) before returning, or add a lazy-init getter to `ChartPropertyDialogModel.getVsAssemblyScriptPaneModel()` mirroring `getChartGeneralPaneModel()`/`getHierarchyPropertyPaneModel()`. The latter is more robust since it also covers `chartAdvancedPaneModel`/`chartLinePaneModel`, which aren't dereferenced unguarded on the immediate-open path today but are dereferenced without `?.` in `closing()` (line ~130) if a user somehow gets the dialog open and clicks OK/Apply.

Not fixing this myself per instructions — flagging back to the lead.

lizhepu:
Round 1 fix applied: added lazy-init getters to the three fields the review identified as unsafe.

**Verified the finding against the real source** (`core/src/main/java/inetsoft/web/composer/model/vs/ChartPropertyDialogModel.java`): `chartAdvancedPaneModel`, `chartLinePaneModel`, `vsAssemblyScriptPaneModel` had plain field getters with no null-check, unlike `chartGeneralPaneModel`/`hierarchyPropertyPaneModel` which already lazy-init. Confirmed `PropertyDialog.ngOnInit()` (`property-dialog.component.ts:58`) unconditionally calls `getScripts()`, which for chart (`chart-property-dialog.component.ts:116`) reads `this.model.vsAssemblyScriptPaneModel.expression` with no `?.` — so `new ChartPropertyDialogModel()` would indeed crash the dialog open with a frontend `TypeError` instead of a backend 500. Also confirmed `closing()` (same .ts file, line ~119) copies `chartAdvancedPaneModel`/`chartLinePaneModel` by reference into the outgoing payload; those aren't property-dereferenced in `closing()` itself, but making them always non-null via lazy-init is still correct and matches what the reviewer asked for.

**Fix**: added the same null-check-then-construct-and-assign pattern used by `getChartGeneralPaneModel()`/`getHierarchyPropertyPaneModel()` to all three getters:
- `getChartAdvancedPaneModel()` / `getChartLinePaneModel()` — both types have public no-arg constructors, so `new ChartAdvancedPaneModel()` / `new ChartLinePaneModel()` directly, mirroring the existing pattern.
- `getVsAssemblyScriptPaneModel()` — `VSAssemblyScriptPaneModel` is an Immutables-generated abstract type with required `expression`/`scriptEnabled` properties, so the default is built via `VSAssemblyScriptPaneModel.builder().scriptEnabled(false).expression("").build()`, matching the exact builder usage already in `ChartPropertyDialogService` (line ~303-306) for the normal (non-degraded) path.

**Compile verified**:
```
./mvnw -pl core -am compile -o
...
[INFO] BUILD SUCCESS
```

No new tests added, matching this PR's established convention. Pushed to `fix/76759-vcg001-chart-rename-npe` at commit 5cb64f3d4.

lizhepu:
## Round 2 review

**Head commit:** `5cb64f3d4dc8b3ad4af413b950466a3b83c9541e`
**Range since round 1:** `337333df5` (round-1 subject) → `5cb64f3d4` (new fix commit)

### What I checked
1. Read the current `ChartPropertyDialogModel.java` in full and diffed the three new getters against the pre-existing `getChartGeneralPaneModel()`/`getHierarchyPropertyPaneModel()`.
2. Read `VSAssemblyScriptPaneModel.java` to check the builder call against its actual Immutables contract.
3. Ran `./mvnw -pl core -am compile -o` myself in this worktree.
4. Re-read `ChartPropertyDialogService.getChartPropertyDialogModel0` (round-1 fix) and `chart-property-dialog.component.ts`'s `closing()`/`ngOnInit()`/`getScripts()` to re-check the fixer's pushback.
5. Walked the VCG-001 scenario end to end against the combined diff, including how the nested models actually reach the frontend via Jackson serialization.

### Round-1 finding status: RESOLVED
`getChartAdvancedPaneModel()`, `getChartLinePaneModel()`, and `getVsAssemblyScriptPaneModel()` (`ChartPropertyDialogModel.java:38-91`) now use the identical null-check → construct → **assign to field** → return shape as `getChartGeneralPaneModel()` (line 26) and `getHierarchyPropertyPaneModel()` (line 62) — the assignment is present in all three (not just returning a throwaway instance), so repeated calls are cached correctly, consistent with the existing convention.

`VSAssemblyScriptPaneModel` is an `@Value.Immutable` abstract class with exactly two abstract accessors, `expression()` (String) and `scriptEnabled()` (boolean) — no other required fields, no defaults declared. `VSAssemblyScriptPaneModel.builder().scriptEnabled(false).expression("").build()` sets both, so this isn't the "missing required field only throws at `build()` time" trap — it's genuinely complete. `ChartAdvancedPaneModel`/`ChartLinePaneModel` both have public no-arg constructors, so the plain `new X()` calls are valid too.

Compiled it myself: `BUILD SUCCESS`, `inetsoft-core` recompiled exactly the 2 changed source files (not a stale no-op).

### Re-examined the `closing()` pushback
Confirmed accurate as stated: `closing()` (`chart-property-dialog.component.ts:128-134`) copies `chartAdvancedPaneModel`, `chartLinePaneModel`, and `vsAssemblyScriptPaneModel` into the outgoing payload by reference with no property dereference on any of the three in that method. The lazy-init fix is still the right (and now necessary) call regardless, for two reasons found while re-checking end-to-end:
- `getScripts()` (called from `ngOnInit()` via the base class) does dereference `vsAssemblyScriptPaneModel.expression` unconditionally — this is the actual crash site round 1 flagged, and it's fixed.
- `chart-property-dialog.component.html` binds `model.chartAdvancedPaneModel`/`model.chartLinePaneModel` straight into `<chart-advanced-pane>`/`<chart-line-pane>` with no `?.`/`@if` guard, so a null there would just push the crash one level down into those child components instead. Lazy-init at the model root is the correct fix, not a narrower one scoped to just `vsAssemblyScriptPaneModel`.

### End-to-end scenario walkthrough
Renamed chart, stale `objectId`, Properties dialog opened:
- Backend: `chartAssembly == null` → returns bare `new ChartPropertyDialogModel()`.
- Jackson serialization calls the bean getters to build the response JSON; since `ChartGeneralPaneModel`'s own nested getters (`getGeneralPropPaneModel()`, etc.) are already lazy-init and get invoked during serialization, the whole tree — including the three newly-fixed fields — serializes as populated-but-empty objects, not JSON `null`. This is the same mechanism that already made `chartGeneralPaneModel`/`hierarchyPropertyPaneModel` safe before this PR.
- Frontend `ngOnInit()` → `getScripts()` reads `vsAssemblyScriptPaneModel.expression` → `""`, no crash.
- Dialog renders: General/Advanced/Hierarchy/Script tabs all bind to non-null-but-empty models, so it opens showing blank/default fields rather than crashing.
- If the user clicks OK, `closing()` also reads `chartGeneralPaneModel.generalPropPaneModel.basicGeneralPaneModel.name` (line 142, pre-existing code, unrelated to this PR's diff) — also non-null via the same cascading-lazy-init-at-serialization mechanism, so no crash there either.

**Non-blocking note:** the dialog opens and behaves without crashing, but it will silently show a blank/default Properties dialog (empty name, default chart type, etc.) for an assembly that no longer resolves, with no indication to the user that this is a stale reference rather than a real empty chart. Not a regression from this fix and arguably out of scope for a targeted NPE fix, but worth a follow-up ticket if there's an appetite for a clearer "this object no longer exists" affordance instead of a blank dialog.

### Fresh pass over the full diff
Nothing else found across both commits (`337333df5`, `5cb64f3d4`). The service-layer null check and the three lazy-init getters are the entire change; both are minimal and consistent with existing conventions in the file.

### Verdict: **APPROVE**

(Posting as a PR comment, not `gh pr review`, since I'm authenticated as the same GitHub user as the PR author — a formal review would be self-review-blocked.)

lizhepu:
## Round 3 review

**Head commit:** `f532a7a59c68ce40a4235042cbdb920bf27a28ee`
**Range since round 2:** `5cb64f3d4` (round-2 approved head) → `f532a7a59` (new write-guard commit)

### What I checked
1. Read the full current body of `setChartPropertyModel` (`ChartPropertyDialogService.java:340-524`), not just the diff hunk, including everything after the new guard through the method's existing `return null;` at line 523.
2. Confirmed the method signature (`public Void setChartPropertyModel(...)`, line 342) accepts a `null` return — it already returns `null` at its normal exit path (line 523), so the new early `return null;` (line 361) is consistent with the existing convention, not a new pattern.
3. Checked every use of `chartAssembly`/`assemblyInfo`/`vsChartInfo`/`chartDescriptor`/`viewsheet` between the guard and the method's end — all are only reached after the guard's early return, so nothing downstream depends on `chartAssembly` being non-null in a way the reordering could break. `chartAssembly` itself is only read again at line 519 (`vsBindingService.createModel(chartAssembly)`), well past the point a null would already have exited.
4. Reproduced the compile myself: `./mvnw -pl core -am compile -o` — `BUILD SUCCESS`, no errors.
5. Spot-checked `TextPropertyDialogService.setTextPropertyDialogModel` (lines 189-206): confirmed accurate — `vs.getAssembly(objectId)` (line 200) is dereferenced immediately at line 201 (`textAssembly.getVSAssemblyInfo()`) with no null check, the identical unguarded pattern. Correctly scoped out of this PR.
6. Fresh full-diff pass across all three commits (`337333df5`, `5cb64f3d4`, `f532a7a59`) — service-layer read guard, the 3 lazy-init getters (verified all 5 pane-model getters in `ChartPropertyDialogModel.java` are now lazy-init: `chartGeneralPaneModel`/`hierarchyPropertyPaneModel` pre-existing, `chartAdvancedPaneModel`/`chartLinePaneModel`/`vsAssemblyScriptPaneModel` added round 2), and now the write-side guard. Nothing else found.

### New write-guard commit: correct
Placement, null-check condition, and `return null;` are all right. No regression risk — the guarded path is a strict subset of the previous unguarded path (an early exit added before the first dereference, nothing removed or reordered).

### Degrade-to-no-op vs. throwing: right call, with one caveat already priced in
Traced what the Apply/OK flow actually does with the return value: this RPC is invoked via `@MessageMapping("/composer/vs/chart-property-dialog-model/{objectId}")` (`ChartPropertyDialogController.java:94`, `public void`, STOMP fire-and-forget) — the controller doesn't read or forward `setChartPropertyModel`'s return value at all. On the frontend, `vs-chart-action-handler.ts:190-210` closes the dialog and calls `sendEvent()` as soon as the dialog's own OK button fires — it does **not** wait for a server ack before closing. So a silent no-op and a would-be exception are behaviorally identical to the user today: the dialog already closes optimistically regardless of what the server does with the write. Returning `null` doesn't introduce a new "user thinks it saved but it didn't" gap beyond what already exists for every stale-reference case (including the read-side blank dialog round 2 already flagged as a non-blocking UX note). No code path surfaces a distinct error either way, so throwing would only add a stack trace to the log with no different user-facing outcome. No blocker; consistent with the read side's degrade decision as the commit message states.

### Verdict: **APPROVE**

All three commits together fully close VCG-001's crash-prevention scope: opening the Properties dialog with a stale `objectId` no longer NPEs (commit 1), the resulting blank model no longer NPEs the frontend on render (commit 2), and clicking Apply/OK on that blank dialog no longer NPEs the write path either (commit 3). The still-open "why does `objectId` go stale in the first place" root-cause question remains explicitly out of scope for this PR, as does the same-shaped gap in `TextPropertyDialogService` (confirmed real, flagged for a separate ticket).

(Posting as a PR comment, not `gh pr review`, since I'm authenticated as the same GitHub user as the PR author — a formal review would be self-review-blocked, per round 2's note.)

PR #5370, opened 2026-09-18
Fix #76788 WBS-067: readJoins() drops native-UI merge/cross joins with null operator fields

Nelson-cao:
## Problem

`MergeJoinService.concatenateTable`/`.insertJoinTable` and
`CrossJoinService.concatenateTableCross` (the native Composer UI's "Add Merge
Join" and keyless cross join actions) build a `TableAssemblyOperator.Operator`
by calling `setOperation()` only -- they never call
`setLeftTable()`/`setRightTable()` on it, unlike the plugin's own
`addCrossJoin()`/`addMergeJoin()` paths and `InnerJoinService`, which always
set both fields.

`WorksheetReadService.readJoins()` silently drops any operator whose
left/right table is null, so `read_worksheet_model()` reported `joins: []`
for these native-UI-built assemblies even though they are fully populated,
valid joins.

## Fix

`readJoins()` now falls back to the assembly's own `getTableNames()[0]`/`[1]`
when an operator's left/right table fields are both null -- but only when the
assembly has exactly 2 source tables. With exactly 2 tables there is only one
possible `(ltable, rtable)` pair, so the fallback can never mismatch a
non-adjacent edge. This gate matters because an ordinary 3+-table join built
through `InnerJoinService` can hold operators keyed by non-adjacent table
pairs (e.g. tables `[A, B, C]` with stored pairs `(A,B)` and `(A,C)`, not
`(A,B)`/`(B,C)`) -- an unconditional positional guess would silently drop the
`(A,C)` edge, which is a real regression, not just under-reporting.

**Accepted residual gap:** a native-UI-built merge/cross join with 3+ sources
whose operators also have null left/right fields is still under-reported
after this fix. No evidence this shape is reachable through the current
native UI actions; documented narrowing, not an oversight.

## Tests

Added to `WorksheetReadServiceTest`:
- `readJoinsFallsBackToTableNamesForNativeUiMergeJoinWithNullOperatorFields` --
  2-source `MergeJoinTableAssembly` built the way `MergeJoinService` actually
  builds it (operator fields null); asserts the fallback engages.
- `readJoinsReportsBothEdgesForThreeTableJoinWithNonAdjacentPairing` -- 3-table
  `RelationalJoinTableAssembly` with non-adjacent `(A,B)`/`(A,C)` pairing, both
  fully populated; asserts both edges are still reported (must behave
  identically before and after this fix, pinning the 2-source-only gate).

`mvn -pl core -am -Dtest=WorksheetReadServiceTest test` -- 57/57 pass.
Full `inetsoft.web.wiz.worksheet.*Test` package also re-run clean, no
regressions.


Nelson-cao:
Adversarial review complete (see internal record: docs/teams/2026-09-18-bugs-76788-worksheet-join-editing/bug-wbs-067/04-review-r1.md in stylebi-wiz).

**Verdict: clean, no blocking findings.** Could not post an approving review — reviewer's `gh` identity matches the PR author.

Checked (fresh clone, `pull/5370/head`, read the live file rather than trusting the fix author's diff excerpt):
- Boundary is exact: `tnames.length == 2` (not `>=2`), overwrite condition is `leftTable == null && rightTable == null && canFallbackByPosition` (AND, not OR) — character-for-character matching the confirmed *revised* design, not the originally-proposed-and-rejected one.
- Iteration mechanism unchanged (`getOperators()` Enumeration) — the rejected `getOperatorCount()`/`getOperator(int)` approach was not reintroduced.
- Both regression tests build the exact operator shapes described (2-source `MergeJoinTableAssembly` with null fields; 3-table non-adjacent `(A,C)` pairing) — not lookalikes.
- `WorksheetEditService.editJoin` (WBS-064's own fix site) correctly left untouched.
- Ran the tests myself, independent of the fix author's report: `mvn -pl core -am -Dtest=WorksheetReadServiceTest test` → 57/57 pass, both new test methods confirmed present by name in the surefire XML.

CI: Semgrep/CLA pass; `build` was still pending at review time — not treated as a blocker given the independent local verification above.

Ready to merge pending the `build` check going green.

PR #5371, opened 2026-09-18
Fix #76788 (WBS-064): populate left/right table names on native-UI cross joins

Nelson-cao:
## Problem

Creating a cross join through the Composer's native UI (not the plugin) left the
resulting join's operator with `leftTable`/`rightTable` both `null`, because
`CrossJoinService.concatenateTableCross` built its `TableAssemblyOperator.Operator`
with `setOperation` only, in both the fresh two-table join branch and the
append-to-existing-join branch. Since `WorksheetReadService.readJoins` and
`WorksheetEditService.editJoin` both read those fields directly (not the
`(ltable, rtable)` map key the operator is actually stored under), the assembly
reported `joins: []` and `edit_join` refused with "operator with no left/right
table name recorded".

## Fix

Set `op.setLeftTable`/`setRightTable` to the same table-name pair each existing
`setOperator(...)` call already uses, in both branches of
`concatenateTableCross`.

A sibling code path with the identical defect shape
(`InnerJoinService.joinSourceAndTargetTables`'s no-auto-match append branch) was
investigated and found to already self-heal via `editExistingJoinTable`'s
"tables no longer joined" recovery loop, so it needs no production change — only
a regression test to guard that self-healing.

## Tests

- `CrossJoinServiceTest` (new) — covers both the fresh-join and append branches
  of `concatenateTableCross`.
- `InnerJoinServiceSelfHealingTest` (new) — locks in that dragging an unmatched
  table onto an existing join self-heals to a fully-fielded operator, for both a
  keyed existing join and a keyless-CROSS existing join.

Refs Redmine #76788, WBS-064.


Nelson-cao:
Adversarial review complete (see internal record: docs/teams/2026-09-18-bugs-76788-worksheet-join-editing/bug-wbs-064/04-review-r1.md in stylebi-wiz).

**Verdict: clean, no blocking findings.** Could not post an approving review — reviewer's `gh` identity matches the PR author.

Checked:
- Diff scope: exactly `CrossJoinService.java` (+6/-0) + two new test files. `InnerJoinService.java` correctly untouched per the diagnosis/refutation's settled scope decision (a sibling defect there was found to self-heal, needs only a regression test).
- Both `newtbl` and `!newtbl` branches' new `setLeftTable`/`setRightTable` calls verified to use the exact same table-name-pair expressions as each branch's pre-existing `setOperator(...)` call — no subtly-wrong pairing.
- `InnerJoinServiceSelfHealingTest`'s replication of the private `joinSourceAndTargetTables` bad-edge-injection logic verified line-by-line faithful to the real source.
- Ran the tests myself (not just trusting the fix author's report): 4/4 pass with the fix; reverted the fix via `sed` and confirmed both `CrossJoinServiceTest` cases fail as expected without it (load-bearing, not vacuous); full `inetsoft.web.composer.ws.joins` package green, no regressions.
- CI: Semgrep/CLA pass; `build` was still pending at review time — not treated as a blocker given the independent local verification above against the exact PR commit.

Ready to merge pending the `build` check going green.

PR #5373, opened 2026-09-18
fix(76788-wbs065): reject assembly names that collide with an unrelated existing assembly

Nelson-cao:
## Problem

`add_join`/`add_cross_join`/`add_merge_join`/`add_concatenation`/`add_mirror`/`add_rotate`/
`add_unpivot` (via `placeAssembly`), the `joinPaths` `addJoin` overload, `add_named_group`
(datasource mode), `import_csv_table`/`import_excel_table`, and `add_sql_query` all build a
new assembly and hand it to `Worksheet.addAssembly()` without checking whether the
caller-supplied name already identifies a different, existing assembly.
`Worksheet.addAssembly()` silently evicts and replaces any assembly already registered
under that name, so a same-named-but-otherwise-unrelated `add_join` call (etc.) destroys the
original assembly with no exception and no mention in the response.

WBS-053 (#76744) added `requireNoSelfReferencingSource`, but that guard only rejects a name
that collides with one of the *new* assembly's own declared sources — it does not catch the
general case of colliding with an unrelated, pre-existing assembly, which is what Redmine
#76788 reports.

## Fix

Adds `WorksheetEditService.Editor.requireNoNameCollision(assembly, ws)`: an exact-match,
case-sensitive check (matching `Worksheet.addAssembly`'s own `getAssembly` lookup) for an
existing assembly already registered under the new assembly's name. Wired into:

- `placeAssembly` (covers `add_join`/`add_cross_join`/`add_merge_join`/`add_concatenation`/
  `add_mirror`/`add_rotate`/`add_unpivot`), after `requireNoSelfReferencingSource` so the
  latter's more specific self-reference message still wins when both apply
- the `joinPaths` `addJoin` overload, which bypasses `placeAssembly`
- `addDatasourceScopedNamedGroup` (backing `add_named_group`'s datasource branch)
- `createEmbeddedTable` (backing `import_csv_table`/`import_excel_table`)
- `addSqlQuery` (backing `add_sql_query`) — inserted immediately after constructing the
  assembly, before the `JDBCQuery`/`UniformSQL` construction, the up-to-10s synchronous SQL
  parse wait, and the real JDBC column-selection query, so a doomed call fails fast

## Test plan

- [x] `WorksheetEditServiceTest`: 4 new tests (two-table `add_join`, `joinPaths` overload,
      `add_merge_join`, and a negative non-colliding `add_merge_join` case) — 42/42 pass
- [x] `WorksheetAgentControllerTest`: 3 new tests (`add_named_group` datasource mode,
      `import_csv_table`, `add_sql_query` — the last also asserts
      `QueryManagerService.getColumnSelection` is never invoked) — 200/200 pass


Nelson-cao:
Review (reviewer-wbs065): clean, no blocking issues found.

Verified against the confirmed diagnosis/refutation design (`docs/teams/2026-09-18-bugs-76788-worksheet-join-editing/bug-wbs-065/{01-diagnosis,02-refute}.md` in stylebi-wiz):
- `requireNoNameCollision` is exact-match, case-sensitive (`ws.getAssembly(name) != null`) — no `containsAssemblyIgnoreCase` regression.
- Ordering in `placeAssembly` and `addJoin(joinPaths)` puts `requireNoNameCollision` after `requireNoSelfReferencingSource`, so self-reference keeps its more specific message.
- All 3 `WorksheetAgentController` insertion points (`addDatasourceScopedNamedGroup`, `createEmbeddedTable`, `addSqlQuery`) are at the correct spot per the confirmed design, including `addSqlQuery`'s fast-fail placement before the 10s SQL parse wait and the JDBC column-selection round-trip.
- `replaceTable`/`replaceEmbeddedTable` independently confirmed to be a mutually-exclusive, legitimate in-place-mutation path (never constructs a new assembly, never calls `ws.addAssembly`) — correctly left untouched.
- All 7 claimed regression tests present and substantive, including `verifyNoInteractions(queryManagerService)` proving the `add_sql_query` fast-fail.

Build/test confirmation: CI is now all green (`build` passed 21m7s) and mergeable. Independently reran the two targeted test classes in a fresh, isolated checkout — `WorksheetEditServiceTest`: 42/42, `WorksheetAgentControllerTest`: 200/200, matching the PR description's numbers exactly.

Ready to merge (squash) as far as this review is concerned.

PR #5374, opened 2026-09-18
fix(text-property-dialog): guard write path against a stale objectId

lizhepu:
## Summary
- `TextPropertyDialogService.setTextPropertyDialogModel` dereferenced `vs.getAssembly(objectId)` with no null check, unlike this same class's own read path (`getTextPropertyDialogModel`), which already guards the identical lookup.
- Found while fixing the chart-specific sibling of this exact gap for Redmine #76759 VCG-001 (`ChartPropertyDialogService.setChartPropertyModel`, see stylebi PR #5369) — confirmed the same read-guarded/write-unguarded asymmetry exists here too.
- Degrades to a no-op (nothing to apply the patch to) rather than NPEing, mirroring both this class's own read-side guard and the chart fix's write-side guard.

## Not part of #76759
This is a Text-assembly-specific fix, unrelated to the chart bug that surfaced it — opened as its own PR rather than bundled into #5369, since #5369 is scoped to chart property dialogs.

## Test plan
- Compiled and verified locally: `./mvnw -pl core -am compile -o`, `TextPropertyDialogService.class` regenerated cleanly.
- No new tests, per this repo's established convention for this class of fix.


lizhepu:
Review verdict: **Approve** (posting as a comment, not a formal review — this session's `gh` account is the PR author, so GitHub rejects `gh pr review --approve`).

Checked:
- **Full method read**: `setTextPropertyDialogModel` (TextPropertyDialogService.java:189-310). The `textAssembly == null` guard sits immediately after the cast, before the only dereference (`Tool.clone(textAssembly.getVSAssemblyInfo())`). `textAssemblyInfo` is declared but only assigned in the non-null branch — the null branch's `return null` exits the method first, so this is definite-assignment-safe (confirmed by clean compile, not just inspection). Nothing between the guard and the rest of the method has an independent side effect that needs to run in the null case.
- **Return type**: method signature is `public Void setTextPropertyDialogModel(...)`, and it already has an unconditional `return null;` at its normal exit (line 309), so the new early `return null;` is consistent with the existing pattern — same as the chart PR's equivalent.
- **Compile**: reproduced locally, `./mvnw -pl build-tools/antlr2-maven-plugin install -o` then `./mvnw -pl core -am compile -o` → BUILD SUCCESS.
- **Caller chain**: `TextPropertyDialogController.setTextPropertyDialogModel` (line 78-89) is a `@MessageMapping` STOMP handler returning `void`, and it discards the proxy call's return value entirely — identical shape to the chart PR's caller. No new silent-failure UX gap: the caller was already ignoring the return value before this fix: 400/NPE turns into an equally-silent no-op, but a no-op is strictly better than the current NPE.
- **Same-bug-other-site spot check**: checked ~10 sibling `*PropertyDialogService` classes (Image, Line, Gauge, Rectangle, Crosstab, TableView, SelectionList, Calendar, Submit, Tab) for the same "read-guarded, write-unguarded" asymmetry that this PR and #5369 fixed. None have that specific asymmetry. Worth flagging separately (not a blocker for this PR): Image/Line/Gauge's *both* get and set methods dereference `vs.getAssembly(objectId)` unguarded (e.g. `ImagePropertyDialogService.java:75,242`, `LinePropertyDialogService.java:57`, `GaugePropertyDialogService.java:74`) — a broader instance of the same underlying bug class that neither this PR nor #5369 was scoped to touch.

No issues found. Scope is correctly narrow and matches the described chart-fix pattern.

PR #5378, opened 2026-09-18
fix(76792): normalize global dashboard oname before registry lookup on rename

lizhepu:
## Summary
- `RepositoryDashboardService.setSettings()` looked up `oldDashboard` via the raw, un-suffixed `oname`, but `DashboardRegistry` keys GLOBAL-scope dashboards with a `__GLOBAL` suffix — so renaming a global dashboard NPE'd on the unchecked `((VSDashboard) oldDashboard).getViewsheet()` call, misreported by the changeset-apply layer as `rollback-failed` even though nothing was actually mutated.
- Apply the same `fixDashboardName()` normalization already used elsewhere in this method to `oldName` before the initial `registry.getDashboard(oldName)` lookup, and drop the now-redundant second `fixDashboardName(oldName, owner)` call further down.
- Add a defense-in-depth `instanceof` check before the cast, matching the existing check later in the same method.

Per-user (owner-scoped) dashboard rename was not affected — `fixDashboardName` is a no-op when `owner != null`.

**Bonus fix surfaced during independent review** (see `docs/teams/2026-09-18-bug-76792-dashboard-rename-npe/bug-76792/02-refute.md` point 3): moving the normalization earlier also fixes the `dependencyHandler.updateDashboardDependencies(owner, oldName, false)` call a few lines below it. `LocalDependencyHandler.updateDashboardDependencies` does its own raw-name registry lookup internally and was silently no-op'ing on the un-suffixed global name before this fix (masked because the NPE aborted the method one line later anyway); it now runs correctly, symmetric with the "add" side of the same method which already passed the fixed name.

Fixes #76792.

## Test plan
- [x] `./mvnw -o compile -pl core -DskipTests` — compiles clean
- [x] Root cause independently re-diagnosed and adversarially refuted in a `/debug-workflow` run — see `docs/teams/2026-09-18-bug-76792-dashboard-rename-npe/` (`bug-76792/01-diagnosis.md`, `02-refute.md`, verdict SURVIVED)
- [ ] Live re-verification via admin-chat's `preview_dashboard_changes`/`apply_dashboard_changes` on a GLOBAL-scope dashboard rename (per bug repro)


jshobe-inetsoft (approved):
# Review: PR #5378 — `fix(76792): normalize global dashboard oname before registry lookup on rename`

Repo: `inetsoft-technology/stylebi` (community)
File changed: `core/src/main/java/inetsoft/web/admin/content/repository/RepositoryDashboardService.java` (+3/-3)
Base: `stylebi-wiz-main-rebase` · Head: `fix/76792-dashboard-rename-global-npe`

## Verdict

**Ready to merge.** No blocking issues found. CI is green (build, Semgrep, Semgrep OSS, CLA all `SUCCESS`), the PR is `MERGEABLE` with no conflicts, and independent source analysis confirms the root-cause diagnosis and fix are correct with no regressions in the owner-scoped (per-user) dashboard path.

Non-blocking suggestions (listed below) are worth considering but do not need to block the merge.

---

## Change Summary

`RepositoryDashboardService.setSettings()` previously looked up the dashboard being renamed with the raw `model.oname()`:

```java
String oldName = model.oname();
Dashboard oldDashboard = registry.getDashboard(oldName);
dependencyHandler.updateDashboardDependencies(owner, oldName, false);
ViewsheetEntry oldEntry = ((VSDashboard) oldDashboard).getViewsheet();
...
oldName = fixDashboardName(oldName, owner);   // fixed too late to help the lookups above
```

`DashboardRegistry` (global scope) always stores dashboards keyed with a `__GLOBAL` suffix (`DashboardRegistry.parseXML()`, line ~202: `if(isGlobal() && !name.endsWith("__GLOBAL")) { name = name + "__GLOBAL"; }`), and `getDashboard()` is a plain `dashboardsMap.get(name)` — no internal normalization. So if a caller supplies the raw (un-suffixed) name for a global-scope rename, `registry.getDashboard(oldName)` returns `null`, and the unchecked cast/dereference `((VSDashboard) oldDashboard).getViewsheet()` throws an NPE (a raw cast of `null` doesn't throw `ClassCastException`; the NPE comes from calling `.getViewsheet()` on the null result).

The fix moves the existing `fixDashboardName()` normalization (already used elsewhere in this method, e.g. for `name`) up to before the initial lookup, adds an `instanceof` guard before the cast (mirroring the existing `instanceof VSDashboard` check later in the same method, line ~195), and drops the now-redundant second `fixDashboardName(oldName, owner)` call.

## Correctness Verification

- Confirmed `fixDashboardName()` is a no-op whenever `owner != null`, so per-user dashboard rename behavior is provably unchanged by this diff (verified by reading the method at the bottom of the file).
- Confirmed `DashboardRegistry.getDashboard()` does no name normalization — verified in `core/src/main/java/inetsoft/sree/web/dashboard/DashboardRegistry.java`, matching the PR's stated root cause.
- Confirmed the "bonus fix" claim: `LocalDependencyHandler.updateDashboardDependencies(IdentityID, String, boolean)` (core/src/main/java/inetsoft/uql/asset/LocalDependencyHandler.java:1934) does its own raw `getRegistry(user).getDashboard(name)` lookup with no internal suffix-fixing. Before this patch, the `dependencyHandler.updateDashboardDependencies(owner, oldName, false)` call a few lines above the crash site used the un-suffixed `oldName` for global dashboards, so it silently no-op'd (masked because the NPE aborted the method one line later anyway). After this patch it receives the already-fixed name and behaves symmetrically with the "add" side call at the bottom of the method (`dependencyHandler.updateDashboardDependencies(owner, name, true)`), which already used the fixed `name`.
- Removing the second `fixDashboardName(oldName, owner)` call is safe: the function is idempotent (it checks `!dashboard.endsWith("__GLOBAL")` before appending), so applying it once earlier is equivalent to applying it twice.
- Traced all subsequent uses of `oldName` in the method (duplicate-name check, `renamed` flag computation, `RenameInfo`/`okey`/`nkey` construction for the rename-transform queue, action-record object naming) — all of them already expected the *fixed* (suffixed) form in the pre-existing code, so moving the fix earlier does not change their inputs.
- The added `instanceof VSDashboard` guard is a genuine (not just cosmetic) improvement: it also protects against the case where `oldDashboard` is legitimately absent from the registry (e.g. stale client state, a race with a concurrent delete) regardless of scope — previously *any* such case would NPE, not just the global-oname case. `PortletDashboard` is the only other historical `Dashboard` implementation found in the codebase, and it is explicitly ignored/dropped by `DashboardRegistry.parseXML()`, so `VSDashboard` is effectively the only concrete type that can come back from `getDashboard()` today; the guard is low-risk, defense-in-depth consistent with the existing identical check later in the same method.

## Context Worth Noting (non-blocking)

- **The standard EM Angular UI already works around this bug today.** `repository-dashboard-settings-page.component.ts` (`changeModel()`) pre-appends `__GLOBAL` to both `model.name` and `model.oname` client-side before POSTing when the dashboard is global-scoped, so the community EM UI would not currently trigger this NPE in normal use. The real exposure is any other caller of `POST /api/em/content/repository/dashboard` that supplies a raw, un-suffixed `oname` for a global dashboard — the PR references an "admin-chat" `apply_dashboard_changes` tool as the trigger, but no such tool (or the `docs/teams/2026-09-18-bug-76792-dashboard-rename-npe/` analysis directory the PR body cites) exists anywhere in this repository (community or enterprise) at this commit. This doesn't make the fix wrong — pushing normalization into the service layer instead of relying on the frontend to pre-normalize is the more robust place for it — but reviewers should be aware the PR's "independently re-diagnosed and adversarially refuted" claim and its live-repro test-plan item (`- [ ] Live re-verification via admin-chat's preview_dashboard_changes/apply_dashboard_changes...`) are unchecked/unverifiable from what's in the diff, so this review's confidence is based on static analysis of the Java call graph, not an end-to-end repro.
- **No test was added.** `RepositoryDashboardService` has no existing unit test coverage in the repo (`find … -iname "*RepositoryDashboard*Test*"` returns nothing), so this PR is consistent with existing practice, but a small unit test for `setSettings()` covering a global-scope rename would meaningfully guard against regressing this exact NPE in the future.
- **Base branch is `stylebi-wiz-main-rebase`**, not `main` — a divergent development branch carrying additional unmerged commits. This appears intentional (matches the target branch's own recent history/conventions) but is worth a sanity check that this is the correct integration target before merging, since it is unusual relative to the branching guidance in the root `CLAUDE.md`.

## Other Review Areas

- **Clean code / readability**: The change is minimal and idiomatic for the surrounding code (mirrors the existing `instanceof VSDashboard` pattern used later in the method). No scope creep.
- **Security**: No new attack surface. The `oldName` used for the `AssetEntry`/registry lookups was already attacker-influenced (`model.oname()` from the request body) before this change; normalizing it earlier does not weaken the existing `ResourceAction.ADMIN` permission check performed above (lines 130–135, unchanged).
- **Backward compatibility**: No API/model/schema changes; purely an internal ordering/null-safety fix. No serialization format changes.
- **Regressions**: None identified. All downstream uses of `oldName` were traced and confirmed to already expect the fixed/suffixed form.

PR #5380, opened 2026-09-18
fix(76786): add_join extends an existing named join in place (WBS-063)

Nelson-cao:
## Summary
- `add_join`'s two-table overload always nested an already-existing named join as one of two sources of a brand-new assembly, regardless of which side the existing join was on. Chaining `add_join` calls onto a join built earlier produced a growing chain of wrapper assemblies instead of one combined view.
- Now, whenever exactly one of `leftTable`/`rightTable` already resolves to a `RelationalJoinTableAssembly`, that join is extended in place (the other, plain side becomes a new member of the SAME assembly) instead of nesting. Symmetric by design — it doesn't matter which side the existing join is on. If both sides are already-existing joins, or neither is, behavior is unchanged (nest / create new).
- Extending reuses `InnerJoinService.editExistingJoinTable`, seeded with the join's own current operators (`getOperatorsOfJoinTable`) so its original edges survive rather than being reconnected as `CROSS_JOIN`s. The new operator is wired against whichever DIRECT member of the join actually owns the given key column (new `resolveMemberOwningColumn` helper, fails loud on zero or multiple matches) rather than the join's own composite name — `concatenateTable`'s `isContain` has no self-check, so passing the composite's own name would add the join as a member of itself. A recursive-join guard (`JoinUtil.tableContainsSubtable`) refuses re-joining a table already a direct member. A `name` argument different from the existing join's current name renames it first (propagating any collision) before extending.
- `add_cross_join`, `add_merge_join`, and the multi-table `joinPaths` overload have the same underlying gap but are out of scope here (flagged as follow-up in code comments).

## Test plan
- [x] `WorksheetEditServiceMutatorsTest` — 8 new tests added covering: extend when existing join is leftTable, extend when existing join is rightTable (symmetry), same-name no-op, different-name rename (+ old name fails "not found"), recursive-join refusal, missing-column refusal, ambiguous-column refusal, both-sides-existing-joins regression (still nests), neither-side-existing-join regression (still creates new). All 266 tests in the file pass.
- [x] `WorksheetEditServiceTest` (existing `joinPaths` overload tests) — 38 tests pass, confirming no regression to the untouched multi-table overload.


Nelson-cao:
Round 1 review, head commit `7fc522038304` (branch `fix/76786-join-nest-extend`).

Full record: `docs/teams/2026-09-18-bug-76786-wbs063-join-extend/bug-76786/05-review-r1.md` in stylebi-wiz.

Checked against `01-diagnosis.md`/`02-refute.md`'s CONFIRMED cause and the lead's symmetric-design decision, not just the reported symptom.

**Confirmed correct:** symmetric extend (either `leftTable` or `rightTable` may be the existing join, matching the lead's decision, not the diagnosis's asymmetric sketch); recursive-join guard fires strictly before any mutation; an operator's table name is never resolved to the existing join's own composite name; the `getOperatorsOfJoinTable` seeding is genuinely necessary (verified directly against `InnerJoinService.editExistingJoinTable`'s CROSS_JOIN-reconnect fallback at `:744-777`) and is actually exercised by a test, not just claimed; `joinPaths`/`addCrossJoin`/`addMergeJoin` are untouched.

**Blocking finding:** the new extend-in-place branch calls `renameTable(existingJoin.getName(), name)` BEFORE validating `leftKeys.size() == rightKeys.size()` and before `resolveMemberOwningColumn`'s missing/ambiguous-column checks. `WorksheetEditService.apply()` has no rollback on exception, so a call that supplies a different `name` together with mismatched-length key lists (or a key that doesn't resolve to exactly one direct member) renames the existing join and STILL throws — leaving the worksheet with the join silently renamed but not extended. None of the 9 added tests catch this because every failure-path test reuses the existing join's own name (no rename triggered). Please reorder: validate/build the operator(s) fully before calling `renameTable`, and add a regression test combining a rename with each failure mode.

**CI note:** the "build" check failed on this head commit with a GitHub Packages dependency-resolution error inside the `antlr2-maven-plugin` build-tools module (`org.eclipse.sisu:org.eclipse.sisu.plexus:jar:0.3.5 (absent)` etc.), before any test runs — unrelated to this diff's `core` source change, and the same-shaped registry failure is not specific to this PR (other same-day PRs against the same base show their own, different build outcomes). Reran the failed job to check for transience; not treating it as a code-correctness signal, since the blocking finding above already withholds approval regardless.

**Verify gate:** `bug-76786/04-verify.md` does not exist yet in stylebi-wiz's docs. Holding on that gate too, independent of this review.

Verdict: **request changes**. Please push a fix for the ordering issue above; will re-review.

Nelson-cao:
Round 2 review, head commit `db473a35a0b1` (branch `fix/76786-join-nest-extend`).

Full record: `docs/teams/2026-09-18-bug-76786-wbs063-join-extend/bug-76786/05-review-r1.md` (Round 2 section) in stylebi-wiz.

Confirmed directly in the diff: the rename block now runs strictly LAST in the extend-in-place branch of `addJoin`, after the recursive-join guard, the `leftKeys`/`rightKeys` length check, and BOTH `resolveMemberOwningColumn` calls (missing- and ambiguous-column paths) have fully succeeded. There is no remaining throwing statement between the rename and the end of the branch other than `editExistingJoinTable`'s own try/catch, unchanged from round 1.

Read all 4 new tests in full (not just their names) — `addJoinExtendingInPlaceRecursiveJoinFailureDoesNotRename`, `...MissingKeyColumnFailureDoesNotRename`, `...AmbiguousKeyColumnFailureDoesNotRename`, `...MismatchedKeyListLengthFailureDoesNotRename` — each genuinely combines a rename request with its named failure mode and asserts both the specific exception AND that the old name still resolves while the new name doesn't. No shortcuts.

Ran the build myself rather than trusting the reported numbers: `./mvnw -o -pl core -Dtest=WorksheetEditServiceMutatorsTest,WorksheetEditServiceTest -Dsurefire.failIfNoSpecifiedTests=false test` from `E:\inetsoft\stylebi-76786-join-extend\community` → **Tests run: 308, Failures: 0, Errors: 0 — BUILD SUCCESS**. (CI's own "build" check was still in progress at review time; not treating it as a signal either way — the local run above is.)

Verdict: **approve**. Round 1's finding is resolved. Not merging yet — per the lead's gate, `04-verify.md` still needs to reflect this round-2 fix specifically before merge.

PR #5381, opened 2026-09-18
Bug #76797, preserve RUNNINGTOTAL resetLevel when breakBy defaults from unset

Zxm-happy:
## Root cause

`VSUtil.updateCalculateInfo()` (`core/src/main/java/inetsoft/uql/viewsheet/internal/VSUtil.java:1443-1482`) clears a `RunningTotalCalc`'s `resetLevel` back to `NONE` (-1) whenever `breakBy` changes within the same call — including the case where `breakBy` was simply omitted from the write request and gets defaulted for the first time to the `ROW_INNER`/`COLUMN_INNER` sentinel (`"0"`/`"1"`).

This makes `resetLevel` a silent no-op for any crosstab RUNNINGTOTAL calculator whose `calculateInfo` leaves `breakBy` unset — the natural shape of a request that lets `breakBy` default to the row dimension the running total is being computed over — regardless of whether that row dimension is date-typed. Setting `resetLevel` via `set_table_fields`/the Composer's Running Total dialog with `breakBy` omitted therefore has no effect at all, either in `get_table_binding`'s readback or in the rendered values.

Confirmed both by source trace (`VSUtil.updateCalculate` → `VSCrosstabBindingFactory.updateAssembly()`, called unconditionally on every crosstab binding write) and by live composer-chat evidence across 5 variants independently varying breakBy-omitted-vs-explicit, date-typed-vs-plain-string row dimension, and row-only-vs-row+col shelf shape — resetLevel survives if and only if breakBy was explicitly present in the request.

## Fix

Capture whether the original `breakBy` (before this method's own defaulting logic runs) was empty/unset. Only clear `resetLevel` when `breakBy` was previously a real/explicit value and is now changing to something different — the case this logic is legitimately guarding against (an already-configured running total gets re-pointed at a different breakBy, so its previously-chosen reset granularity may no longer make sense). A transition from empty/unset to a freshly-computed default sentinel is no longer treated as a "user changed breakBy" event.

The pre-existing early-return guard (an explicitly-supplied breakBy that already matches the shelf-implied sentinel short-circuits before the resetLevel-clearing check ever runs) is untouched.

## Tests

New `core/src/test/java/inetsoft/uql/viewsheet/internal/VSUtilTest.java` (no test previously exercised `VSUtil.updateCalculateInfo()`/`updateCalculate()` at all):
- `resetLevelIsPreservedWhenBreakByDefaultsToRowInner` / `...ColumnInner` — the direct regression: breakBy omitted, resetLevel explicitly set, must survive the defaulting-to-sentinel transition. Fails pre-fix (`expected: <0> but was: <-1>`), passes post-fix.
- `resetLevelIsPreservedWhenBreakByAlreadyMatchesRowInner` / `...ColumnInner` — guards the pre-existing early-return-guard path (breakBy already explicit and matching the shelf-implied sentinel) against a future regression; passes both pre- and post-fix, confirming the fix doesn't disturb the already-working case.

An initially-proposed third test case ("breakBy explicitly re-pointed from one real dimension to a different real dimension should still clear resetLevel") was dropped after a live check: on the `set_table_fields`/`updateAssembly()` write path, the aggregate's calculator is always rebuilt fresh from the incoming request, so there is no persisted "old" breakBy in scope to diff against, and `resetLevel:1` was empirically confirmed to survive a real-dimension-to-real-dimension `breakBy` change unchanged on this path. That scenario doesn't reproduce here; a test for it would need to target the DND path (`VSCrosstabDndService.java`) instead, which does carry an old calculator into the call, but that's left out of scope for this fix.

## Verification

```
mvn -pl core test -Dtest=VSUtilTest,CompoundGrowthCalcTest,RunningTotalCalcTest,RunningTotalColumnTest -DfailIfNoTests=true
```
All green: `VSUtilTest` (4/4), plus existing `CompoundGrowthCalcTest` (1/1), `RunningTotalCalcTest` (2/2), `RunningTotalColumnTest` (15/15) — no regressions in related calculator tests.


jshobe-inetsoft (approved):
# Review: PR #5381 — Bug #76797, preserve RUNNINGTOTAL resetLevel when breakBy defaults from unset

Repo: `inetsoft-technology/stylebi` (community). Branch `fix/76797-runningtotal-resetlevel-breakby-default` → `main`.
Files changed: `core/src/main/java/inetsoft/uql/viewsheet/internal/VSUtil.java` (+4/-2), new test `core/src/test/java/inetsoft/uql/viewsheet/internal/VSUtilTest.java` (+92).

## Verdict

- CI: all reported checks pass (Semgrep Scan, Semgrep OSS, Validation Build, license/cla), excluding the disabled "Claude Code Review / claude-review" check, which is out of scope per project convention.
- Merge state: `mergeable: MERGEABLE`. The GitHub `mergeStateStatus` shows `BLOCKED` only because the branch protection rule requires a review (`reviewDecision: REVIEW_REQUIRED`, zero reviews so far) — not because of merge conflicts. That is expected/normal and not a code defect.
- No test cases are broken by this change (see analysis below).

Non-blocking suggestions are listed at the end; none of them need to hold up the merge.

---

## Analysis

### Root cause and fix correctness

I traced the write path: `VSCrosstabBindingFactory.updateAssembly()` (`core/src/main/java/inetsoft/web/binding/service/VSCrosstabBindingFactory.java:286`) calls `VSUtil.updateCalculate(crossInfo, null)` unconditionally on every crosstab binding write. The per-aggregate `RunningTotalCalc` is rebuilt fresh from the incoming request via `RunningTotalCalcInfo.toCalculator0()` (`core/src/main/java/inetsoft/web/binding/model/graph/calc/RunningTotalCalcInfo.java:43-50`), which sets `resetLevel` and `breakBy` directly from the request JSON — so a request that supplies `resetLevel` but omits `breakBy` correctly produces a calculator with the user's chosen `resetLevel` and an empty `breakBy`. `VSUtil.updateCalculateInfo()` then defaults `breakBy` to the `ROW_INNER`/`COLUMN_INNER` sentinel, and the pre-fix code treated that unset→sentinel transition the same as any other change, unconditionally zeroing `resetLevel`. This confirms the diagnosed root cause.

I manually traced all three code paths inside `updateCalculateInfo` (row-add, column-add, and the generic/aggregate-add `else if(!findColumn)` branch that the request-write path hits with `dropType == null`) against the new `breakByWasUnset` guard:

- **Unset → sentinel (the bug):** `breakByWasUnset` is `true`, captured *before* any mutation of `rcalc`. Since the local `breakBy` variable is never reassigned (only the `rcalc` object is mutated), the final `if` correctly skips the "breakBy changed" clause, and none of the OR clauses that check `ROW_INNER.equals(breakBy)`/`COLUMN_INNER.equals(breakBy)` against the *original* (empty) value fire either. `resetLevel` is preserved. ✓ matches both new tests.
- **Already-explicit sentinel (`breakBy == ROW_INNER`/`COLUMN_INNER` before the call):** the pre-existing early-return guard (`if(AbstractCalc.ROW_INNER.equals(breakBy) && hasRow) return;` etc.) fires before the resetLevel-clearing check is ever reached, exactly as before this diff. Untouched, still correct. ✓ matches both new tests.
- **Genuine real-dimension breakBy change** (e.g. `breakBy` explicitly set to a real column name that no longer resolves, so `findColumn` is false and the method reassigns it to `ROW_INNER`/`COLUMN_INNER`): `breakByWasUnset` is `false`, so behavior is identical to pre-fix — `resetLevel` is still cleared. Not exercised by a new test, but reasoned through manually (see suggestion below).
- **Rows/columns changed while breakBy is already `ROW_INNER`/`COLUMN_INNER`:** unaffected by the new guard (those OR clauses don't depend on `breakByWasUnset`), so behavior is unchanged.

I also checked the other call sites of the same method: `VSCrosstabDndService` (drag-and-drop add-column flow, `core/src/main/java/inetsoft/web/binding/dnd/controller/VSCrosstabDndService.java:115,286`) calls the very same `VSUtil.updateCalculateInfo`, so the fix applies uniformly there too — this is desired, not a regression, since the "unset defaults to sentinel is not a user-initiated breakBy change" reasoning applies regardless of which caller triggers the defaulting. Note that column-*removal* (`dndTotree`) goes through a separate, private `updateCalculateInfo` overload defined inside `VSCrosstabDndService.java` itself (different signature, not touched by this diff), so this fix does not affect the removal path at all. The PR description's claim that a genuine real-to-real `breakBy` re-point "doesn't reproduce on this [request-write] path" checks out: the calculator is always rebuilt fresh from the request with no persisted "old" value to diff against, so the `!Tool.equals(breakBy, rcalc.getBreakBy())` clause can only ever be triggered by this method's own internal defaulting logic on that path — which is exactly the case the fix now correctly excludes only for the unset→default transition.

### Code quality

- The change is minimal and scoped exactly to the described defect; no unrelated refactoring.
- `Tool.isEmptyString(breakBy)` matches the idiom already used a few lines below in the same method for `PercentCalc`'s `percentage` field — consistent style.
- `inetsoft.util.Tool` is already available via the existing `import inetsoft.util.*;` wildcard import, and `Tool.equals`/`Tool.isEmptyString` are already used elsewhere in this file, so the diff is self-contained and compiles.
- The updated comment above the `if` clearly explains the new exception ("but not just defaulted from unset to its first-time sentinel within this same call"), which aids maintainability.
- No error handling concerns: this is pure boolean/string logic with no I/O, exceptions, or nullability risk beyond what `Tool.isEmptyString`/`Tool.equals` already handle (both are null-safe).
- No security concerns: no user input is used unsafely, no injection surface, no serialization changes.

### Tests

- New `VSUtilTest.java` follows the exact boilerplate/conventions of sibling tests in the module (e.g. `RunningTotalColumnTest.java`) — `@ExtendWith(SpringExtension.class)`, `@ContextConfiguration`, `@SreeHome`, `@Tag("core")` — so it should integrate cleanly with the existing test infrastructure and tagging used for selective test runs.
- The four tests directly cover the two behavior classes fixed/preserved by this diff: unset→row-sentinel, unset→column-sentinel, already-row-sentinel, already-column-sentinel. The PR author verified locally that `VSUtilTest`, `CompoundGrowthCalcTest`, `RunningTotalCalcTest`, and `RunningTotalColumnTest` all pass; I did not re-run the Maven build in this review, but nothing in the diff appears capable of breaking those or any other existing test — no existing test in the repo exercises `VSUtil.updateCalculateInfo`/`updateCalculate` prior to this PR (confirmed via search), so there is no prior test whose expectations this change could invalidate.
- Backward compatibility: this only changes when `resetLevel` is *cleared*; it never causes a previously-preserved `resetLevel` to be cleared. Existing viewsheets/bookmarks with a persisted `resetLevel` on a `RunningTotalCalc` are unaffected unless they hit exactly the buggy path (breakBy omitted + resetLevel set), where behavior now matches user intent instead of silently discarding it. No serialization/format change either (no XML/JSON schema touched).

### Non-blocking suggestions

1. **Missing unit test for the "genuine real-dimension breakBy change still clears resetLevel" case.** The PR description explains why this doesn't reproduce via the `VSCrosstabBindingFactory` write path, but at the `updateCalculateInfo` unit level it's easy to construct directly (e.g. `calc.setBreakBy("SomeRealDim"); calc.setResetLevel(YEAR); updateCalculateInfo(true, false, false, calc, null);` with `findColumn=false` for a dimension no longer present) and would guard against a future over-broadening of the `breakByWasUnset` exception silently suppressing legitimate resetLevel clears. Would strengthen confidence but is not required for correctness of the current fix.
2. Tests 3 and 4 (`resetLevelIsPreservedWhenBreakByAlreadyMatchesRowInner`/`ColumnInner`) both terminate at the same pre-existing early-return guard and don't exercise the new `breakByWasUnset` line at all — they're valid regression guards for the untouched guard, but a reviewer looking for coverage of the *new* code specifically should rely on tests 1 and 2.
3. Consider (optionally, future work, not this PR) adding an explicit test at the `VSCrosstabDndService`/DND call-site level, since that path shares the same underlying method and drop-type branches (`ROW_HEADERS`/`COLUMN_HEADERS`) that aren't directly exercised by the new tests (which only exercise the `dropType == null` branch).

PR #5382, opened 2026-09-18
feat(wiz): add cross-worksheet add_mirror form (Redmine #76801)

Zxm-happy:
## Summary
- `add_mirror` could only reference an assembly already in the CURRENT worksheet (`WorksheetEditService.Editor.addMirror` uses the 3-arg, `entry=null` `MirrorTableAssembly` constructor — an inner mirror). The native Composer UI's drag-and-drop already builds a real outer mirror (`AssetEventUtil.createMirrorAssembly` with a non-null `AssetEntry`) when a saved worksheet asset is dropped onto another worksheet's canvas, but that mechanism (`WorksheetOpenAssetService`) is interactive/`CommandDispatcher`-shaped and unreachable from the headless wiz agent path.
- Adds `path`/`scope` fields to `EditRequest` and a new `add_mirror` branch in `WorksheetAgentController` that resolves a saved worksheet asset by path, copies its outer dependency assemblies via `AssetUtil.copyOuterAssemblies`, and wraps the result in a real outer `MirrorTableAssembly` via `AssetEventUtil.createMirrorAssembly` — matching `update_mirror`'s own `applyOnRuntime` + `AssetRepository` precedent (not a new bypass-dispatch shape; reuses the pattern this codebase already established for exactly this class of need).
- Today's same-worksheet `add_mirror(name, source)` form is untouched — the new branch only fires when `path` is present.
- `set_mirror_auto_update` needed no code change: it already works on any `MirrorTableAssembly` with a non-null `AssetEntry`, closing the loop its own docstring already promised for a cross-worksheet mirror. Confirmed by test.

## Test plan
- [x] `mvn -pl core test -Dtest=WorksheetAgentControllerTest -o` — 202/202 passed (includes 6 new tests: same-worksheet regression, new cross-worksheet happy path, missing-name refusal, unknown-worksheet refusal, `set_mirror_auto_update` toggling a cross-worksheet mirror)
- [x] `mvn -pl core test -Dtest=WorksheetAgentControllerConvertToEmbeddedTest -o` — 5/5 passed (also constructs `EditRequest`)
- [x] `mvn -pl core test -Dtest=ViewsheetEditServiceTest -o` — 73/73 passed (also constructs `EditRequest` positionally)

Companion wiz-side PR (plugin/composer `add_mirror` tool extension + `add_table` refusal fix) tracked separately against `stylebi-wiz`'s `main`.

Redmine #76801 (WBS-071).


jshobe-inetsoft (changes requested):
# Review: stylebi PR #5382 — feat(wiz): add cross-worksheet add_mirror form (Redmine #76801)

Base branch: `stylebi-wiz-main-rebase` (a "wiz" AI-agent feature line, not yet merged to `main`).
Files changed: `EditRequest.java` (+68/-2), `WorksheetAgentController.java` (+74), `WorksheetAgentControllerTest.java` (+232).

## Verdict

**Request changes — one blocking issue.**

- **Blocking:** The new `addWorksheetMirror` path in `WorksheetAgentController` registers the newly-created outer mirror with `ws.addAssembly(mirror)` without first calling `WorksheetEditService.Editor.requireNoNameCollision(mirror, ws)`. `Worksheet.addAssembly` silently evicts and replaces any existing assembly that already has the requested name. Every other assembly-creation path in this exact class (and in `Editor.placeAssembly`) calls this guard — it is even documented as package-private specifically so `WorksheetAgentController` can reuse it for direct-construction paths like this one — but the new method skips it, so `add_mirror` with a `path`/`scope` and a `name` that collides with an existing table/assembly in the current worksheet will silently destroy that assembly with no error. See "Findings" below for full detail and file/line evidence.

CI (Semgrep Scan, Semgrep OSS, build) all report SUCCESS and the PR is `MERGEABLE`, so the only reason to hold this PR is the correctness issue above.

---

## Findings

### 1. Blocking: missing `requireNoNameCollision` guard before `ws.addAssembly(mirror)`

`core/src/main/java/inetsoft/web/wiz/worksheet/WorksheetAgentController.java`, new `addWorksheetMirror` method:

```java
editService.applyOnRuntime(sessionToken, user, rws -> {
   Worksheet ws = rws.getWorksheet();
   WSAssembly[] created = AssetUtil.copyOuterAssemblies(
      assetRepository, entry, user, ws, new Point(25, 25));
   WSAssembly source = created[created.length - 1];
   WSAssembly mirror = AssetEventUtil.createMirrorAssembly(rws, source, name, entry);
   mirror.setPixelOffset(new Point(25, 25));
   AssetEventUtil.adjustAssemblyPosition(mirror, ws);
   ws.addAssembly(mirror);
   return null;
});
```

`Worksheet.addAssembly` (`core/src/main/java/inetsoft/uql/asset/Worksheet.java`) silently replaces any existing assembly with the same name — there is no exception, no log warning visible to the caller, just quiet data loss.

This exact hazard already has a dedicated, well-documented guard in the same package:

```java
// WorksheetEditService.java
/**
 * Refuses an assembly name that already identifies a different, existing assembly in the
 * worksheet. {@link Worksheet#addAssembly} silently evicts and replaces whatever already
 * has that name ...
 *
 * <p>Package-private (not {@code private}) so {@link WorksheetAgentController} can reuse
 * the same check for the assembly-creation paths it builds directly against
 * {@code Worksheet}/{@code RuntimeWorksheet} rather than through this {@code Editor}.
 */
static void requireNoNameCollision(WSAssembly assembly, Worksheet ws) throws PairingException { ... }
```

And `WorksheetAgentController` already reuses it at three other direct-construction call sites (confirmed in the base branch, pre-existing, unrelated to this PR): lines ~1224, ~2478, ~4315, each following the same idiom —

```java
WorksheetEditService.Editor.requireStorableName(name, "...");
... build the assembly ...
WorksheetEditService.Editor.requireNoNameCollision(assembly, ws);
... position ...
ws.addAssembly(assembly);
```

`addWorksheetMirror` calls `requireStorableName` (good — matches the idiom) but omits the `requireNoNameCollision` step before `ws.addAssembly(mirror)`. Since this is a headless, AI-agent-facing controller, a colliding `name` is not a far-fetched input — an agent could easily reuse a name from a prior op, or from a table already present in the worksheet being edited, and the result is silent, permanent, uncommunicated destruction of the existing assembly (and everything that depends on it) instead of a clear `PairingException`.

None of the 5 new tests in this PR exercise a name collision, so this gap is untested as well as unguarded.

**Suggested fix:** add `WorksheetEditService.Editor.requireNoNameCollision(mirror, ws);` right after `mirror` is constructed and before `ws.addAssembly(mirror)`, matching the pattern at the other three call sites, plus a regression test asserting a `PairingException` (not silent eviction) when `name` already exists in the target worksheet.

(Lower-severity, same family: `Editor.placeAssembly` also calls `requireNoSelfReferencingSource`, which the new method omits too. In practice this narrower check mainly matters when a caller-supplied source name is echoed into the new assembly's own name, which is less directly reachable here since the outer-copied dependency names are auto-generated by `copyOuterAssemblies`'s `prefix + i` scheme rather than caller-supplied — so this omission is consistent with how the controller's *other* existing direct-construction methods already behave, none of which call it either. Not blocking on its own, but worth a look together with the fix above.)

### 2. Non-blocking: new mirror doesn't get `setLastModified` set at creation

The reference implementation this PR explicitly follows (`WorksheetOpenAssetService.createNewMirror`) does:

```java
nassembly = (MirrorAssembly) AssetEventUtil.createMirrorAssembly(rws, assembly, name, entry);
nassembly.setLastModified(src.getLastModified());
```

`addWorksheetMirror` never sets `lastModified` on the new mirror, so `MirrorTableAssemblyInfo`'s "modified" display (used in the Composer's mirror-table properties/description dialog) will show `new Date(0)` (1970) until the mirror is later refreshed via `update_mirror` (which does correctly set `lastModified` — `MirrorAssemblyImpl.updateMirror`). This is cosmetic only — it does not affect data correctness or the mirror's actual freshness/update behavior — but it is a small, easily-fixed deviation from the documented reference implementation this PR says it's mirroring. Consider adding `mirror.setLastModified(...)` (via the resolved source worksheet's own `getLastModified()`) for parity.

### 3. Non-blocking: minor doc/description nits
- The new `addWorksheetMirror` javadoc cites `WorksheetOpenAssetService#dropWorksheet` as "the same construction the native Composer UI performs," but the actual mirror-construction logic lives in the private `createNewMirror` helper that `dropWorksheet` calls into. The javadoc's entry-point reference is still accurate as a pointer for readers, just slightly imprecise about where the logic itself lives.
- The PR description's test-plan bullet claims "6 new tests" but lists/diff shows 5 new `@Test` methods. Trivial, doesn't affect the review.

### 4. What's correct / verified
- **EditRequest record field ordering**: this is a positional record with 81 components; I mechanically compared the new compatibility constructor's parameter list (and its delegating `this(...)` call) against the canonical field declaration order and confirmed they match exactly, including where `path`/`scope` are appended and defaulted to `null, null`. No off-by-one.
- **Backward compatibility**: the new fields are appended at the end of the record, and a new compatibility constructor is added mirroring the pattern already used for every prior field addition to this record (there are ~11 such constructors already). Existing callers (`WorksheetAgentControllerConvertToEmbeddedTest`, `ViewsheetEditServiceTest`) are untouched by this diff and the CI build (which compiles and runs tests) is green, confirming no breakage.
- **Dispatch guard placement**: the new `if("add_mirror".equals(req.op()) && req.path() != null && !req.path().isBlank())` branch in `editOp` is placed before the generic `dispatch` fallthrough and is scoped correctly — it does not intercept the existing same-worksheet `add_mirror(name, source)` form (verified: a request with `source` and no `path` still falls through to `editService.apply(...)` → `dispatch`). This is also covered by a dedicated regression test (`editAddMirrorWithSourceStillResolvesSameWorksheetAssembly`).
- **Permission enforcement**: the pre-check calls `assetRepository.getSheet(entry, user, true, AssetContent.ALL, false)` with `permission=true`, exactly matching the equivalent, more heavily-commented precedent in `ViewsheetAssemblyAgentController.resolveDataSourceEntry`'s `worksheet` branch (which explicitly warns against using a `permission=false` probe for this exact reason — an agent should not be able to confirm existence of an asset it can't read). `AssetUtil.copyOuterAssemblies`'s own internal `getSheet(..., false, ...)` call is unchecked by design (existing code, unrelated to this PR) — the upfront permission-checked probe is what supplies real access control here, and it does so correctly.
- **Column/data correctness**: `WSAssembly source = created[created.length - 1]` correctly matches the native `createNewMirror`'s own indexing convention for picking the primary/outer-most copied assembly out of `copyOuterAssemblies`'s dependency-sorted result array.
- **Scope/org resolution**: `"user".equalsIgnoreCase(req.scope())` → `USER_SCOPE` else `GLOBAL_SCOPE`, and `uname.orgID` for the entry's org — both consistent with the established `resolveDataSourceEntry` pattern used on the viewsheet side for the equivalent `attach_base_worksheet` op.
- **No apparent security issue**: path traversal / identifier-injection is guarded via `WizUtil.requireNoCaret`, matching the existing `attach_base_worksheet` precedent; no SQL/script injection surface is introduced.
- Read `claude/asset-framework.md` and `claude/worksheet-query.md` for AssetEntry/AssetRepository/dependency-tracking background; no dedicated `claude/*.md` document currently covers the `wiz` AI-agent worksheet-editing surface itself (this feature line predates/parallels those docs), so review relied directly on reading the surrounding source and the establishe­d in-file precedents (`updateMirror`, `resolveDataSourceEntry`, `Editor.placeAssembly`, the other three `requireNoNameCollision` call sites).

PR #5383, opened 2026-09-18
Fix org properties read scope for non-caller organizations (Bug #76798)

fredaan:
## Summary

- `SecurityService.readOrganizationProperties(orgId)` re-resolved each already-found,
  prefix-stripped property name via `SreeEnv.getProperty(name, false, true)` ->
  `PropertiesEngine.useAvailableOrgProperty`, which derives "current org" from
  `OrganizationManager.getCurrentOrgID()` (the calling principal's own org) instead of the
  `orgId` parameter actually being read. Reading an organization other than the caller's
  own (e.g. an admin in `host-org` reading org `ai-verify-fresh-org` via
  `get_identity_organization`) therefore silently returned an empty `properties` list,
  even though the value was persisted correctly at the org-prefixed key
  (`inetsoft.org.<orgId>.<name>`) — confirmed via a raw `get_property` lookup in the
  original report.
- The sibling write path, `applyOrganizationProperties`, already wraps its writes in
  `OrganizationManager.runInOrgScope(orgId, ...)`. This PR wraps the read the same way.
- Adds a regression test, `SecurityServiceTest.getOrganization_callerInDifferentOrg_stillReadsBackProperties`,
  that simulates a caller in one org reading a different org's properties and fails
  without the fix (`expected: <1> but was: <0>`), passes with it.

Redmine: #76798

## Test plan

- [x] `./mvnw.cmd -pl core -am compile -DskipTests` — clean build
- [x] `./mvnw.cmd -pl core test -Dtest=SecurityServiceTest` — 65/65 pass with the fix
- [x] Confirmed the new test fails before the fix (reverted `SecurityService.java` only,
      reran, saw the exact pre-fix symptom) and passes after reapplying it


fredaan:
## Review findings (bug #76798)

The production fix itself is correct and general: `readOrganizationProperties` now wraps its whole scan in `OrganizationManager.runInOrgScope(orgId, ...)`, which directly targets the confirmed root cause (`OrganizationManager.getCurrentOrgID()` resolving to the caller's own org instead of the target `orgId` during the `SreeEnv.getProperty(name, false, true)` re-lookup). No literal-org special-casing. The `try/catch(Exception e) { throw new RuntimeException(e); }` wrap has a real, load-bearing precedent in `RecycleBin.migrateStorageData` (`core/src/main/java/inetsoft/web/RecycleBin.java:367`) — `readOrganizationProperties`'s only caller chain (`getOrganizations()`'s `.stream().map(...)`, `SecurityService.java:774`) is a `Function` that cannot propagate a checked exception, so the wrap is required, not decorative. Independently ran `./mvnw.cmd -pl core test -Dtest=SecurityServiceTest` myself in the fix worktree: 65/65 pass. The new regression test (`getOrganization_callerInDifferentOrg_stillReadsBackProperties`) is a genuine regression test, not a tautology — its `SreeEnv.getProperty` stub only returns the value when the simulated current-org matches the target org, which only happens post-fix.

### Blocker: same defect exists at a second, untouched call site

`UserTreeService.getOrganizationModel` (`core/src/main/java/inetsoft/web/admin/security/user/UserTreeService.java:1044-1059`) has the identical unscoped `SreeEnv.getProperty(propName, false, true)` re-lookup for a caller-supplied `orgID`, with no `runInOrgScope` wrap anywhere in the method. It's reachable via a real HTTP endpoint with an arbitrary org: `OrganizationController.getOrganization` (`OrganizationController.java:75-92`, `GET /api/em/security/providers/{provider}/organization/{organization}/`). An EM sysadmin viewing another org's settings through the EM UI (rather than admin-chat) would hit the exact same silent-empty-properties symptom this PR fixes for the wiz surface.

Notably, this PR's own new Javadoc on `readOrganizationProperties` (pre-existing text from an earlier commit, not part of this diff) already names `UserTreeService.getOrganizationModel` as having "its own identical scan" — so the duplicate was already known in-repo.

Not necessarily in scope for this PR, but should be filed as a follow-up bug so the EM-UI path gets the same fix, since it's the same user-visible defect left unfixed in a sibling method.

### Non-blocking

The `build` (Validation Build) CI check on this PR is still pending/in-progress as of this review — should resolve green before merge.

Full review notes: `docs/teams/2026-09-18-bug-76798-org-properties-read-scope/bug-76798/04-review-r1.md` (stylebi-wiz repo).

PR #5384, opened 2026-09-18
fix(asset): rename_table no longer corrupts a table's own aggregate (bug 76796)

fredaan:
## Root cause

`Worksheet.renameAssembly` (`core/src/main/java/inetsoft/uql/asset/Worksheet.java:459-492`) calls
`renameDepended(oname, nname)` on every assembly in the worksheet, including the one that was just
renamed (it's still in the `assemblies` list, and its own `setName(nname)` already ran). For a
table added from a logical-model entity (`add_table` with a `logicalModel`), each column's
`AttributeRef` entity is set to the logical-model entity name
(`WorksheetAgentController.addLogicalModelTable`), which — because both the entity name and the
assembly's default worksheet name are derived from the same `entityName` local variable — is
guaranteed to equal the table's own initial worksheet-assembly name too. A pre-existing
rename-cascade mechanism (`AggregateInfo.renameDepended`/`renameEntity`, added in commit
`28a5ede3b` to keep a *different* table's aggregate pointed at a renamed source it depends on,
e.g. a join/mirror) can't distinguish "this entity denotes the assembly's own name" from "this
entity denotes an unrelated logical-model source that happens to share that string" — so it fires
on the renamed table's own aggregate too, rewriting the `AttributeRef` entity to the *new*
worksheet name in place. That entity no longer resolves against the logical model at all, and the
next column-selection refresh's `AggregateInfo.validate()` silently drops the now-unresolvable
group/aggregate — matching the reported symptom exactly (`rename_table` returns `{ok:true}`, but
the aggregate and its underlying column both disappear).

## Fix

`AbstractTableAssembly.renameAggregateInfo` (`core/src/main/java/inetsoft/uql/asset/AbstractTableAssembly.java:1707`)
now skips `ginfo.renameDepended(oname, nname)` when the assembly being processed is the one being
renamed (`getName() == nname`, reliable since `setName(nname)` already ran and worksheet assembly
names are unique). The legitimate cross-table case (a join/mirror's own aggregate pointing at some
*other*, renamed source table) is untouched, since that assembly's own name never equals `nname`.

**Composed tables (joins/mirrors/concatenated tables):** `ComposedTableAssembly.renameAggregateInfo`
calls `super.renameAggregateInfo(...)` first (inheriting this same guard) and then runs its own
extra `ColumnRef.renameColumn` loop over the composed table's group/aggregate columns. Traced how
every composed-table type builds those columns (`MirrorTableAssembly.updateColumnSelection`,
`ConcatenatedTableAssembly.getDefaultColumnSelection`, `CompositeColumnHelper`) — in every case the
entity is set to a *child/source* table's name via `AssetUtil.getOuterAttribute(childName, ...)`,
never to the composed table's own name (a composed table can't be its own child). So the self-rename
collision this bug is about structurally cannot occur for `ComposedTableAssembly`'s own extra loop,
and no second guard is needed there — the single guard in the base class fully covers both cases.

## Tests

Added two tests in `core/src/test/java/inetsoft/uql/asset/WorksheetTest.java`:

- `renameAssemblyDoesNotRequalifyItsOwnAggregateRefOnSelfRename` — builds a `BoundTableAssembly`
  with a `SourceInfo.MODEL` source and a column whose `AttributeRef` entity equals the assembly's
  own name (mirroring `add_table`'s logical-model construction), applies an `AggregateInfo`
  sharing that same `ColumnRef` object (mirroring `WorksheetMutationSupport.applyAggregateInfo`'s
  no-clone sharing), renames the assembly, and asserts the aggregate survives with its entity
  unchanged.
  - **Confirmed failing before the fix**: `expected: <Product> but was: <ReturnTotalSummary>`.
  - **Passes after the fix.**
- `renameAssemblyStillRequalifiesAnotherTablesAggregateThatDependsOnTheRenamedTable` — a
  `MirrorTableAssembly`'s aggregate over a renamed *source* table must still be correctly
  requalified to the new name (the legitimate case commit `28a5ede3b` was for). Passed both before
  and after the fix, confirming no regression.

Ran both scoped (`./mvnw -o -pl core -Dtest=WorksheetTest#... test`) and the full
`AggregateInfoTest` + `WorksheetEditServiceMutatorsTest` suites (286 tests total) — all green,
no regressions.


fredaan:
Independent review (reviewer seat, debug-workflow bug 76796): review is clean, no blocking issues found. (`gh pr review --approve` failed with "Can not approve your own pull request" — this run's gh identity matches the PR author, so approval/merge is blocked on the self-review restriction, not on any code issue. Escalating to a human for the actual approve+merge.)

- Fix guard (`Tool.equals(getName(), nname)`) is a general self-identity check, not keyed on any repro literal.
- Sits at the diagnosed root (`AbstractTableAssembly.renameAggregateInfo`, the per-assembly dispatch point in `Worksheet.renameAssembly`'s loop), not a downstream symptom-suppressor — `AggregateInfo.renameEntity`/`ColumnRef.renameAttributeRef` (the legitimate cross-table cascade) are untouched.
- Spot-checked `ComposedTableAssembly.renameAggregateInfo`'s own separate `ColumnRef.renameColumn` cascade myself (not taken on faith): `MirrorTableAssembly`, `ConcatenatedTableAssembly`, and `CompositeColumnHelper` all derive column entities from a distinct child/source table's name, never the composed table's own name — structurally immune to this bug's self-collision, so the base-class guard (inherited via `super`) fully covers it.
- Regression tests target the actual corrupted field (entity string on the shared `ColumnRef`), not a shallower proxy; a companion test proves the legitimate cross-table cascade still works post-fix.
- Independently reran the fixer's scoped tests and a broader regression sweep in my own worktree (PR head `0d16f0d39`): `WorksheetTest` (2/2), `AggregateInfoTest`+`WorksheetEditServiceMutatorsTest` (286/286) — all `BUILD SUCCESS`. GitHub's own CI `build` check was still queued at review time; treating the build as confirmed per independently-run scoped verification.

jshobe-inetsoft (approved):
# Review: PR #5384 — `fix(asset): rename_table no longer corrupts a table's own aggregate (bug 76796)`

Repo: `inetsoft-technology/stylebi` (community). Base: `stylebi-wiz-main-rebase`. Head: `fix/redmine-76796-rename-aggregate-drop`.

## Verdict

- CI is green: `Semgrep Scan`, `build` (full validation build, which runs the test suite), and `Semgrep OSS` all report `SUCCESS`.
- `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE` — no merge conflicts.
- The fix is minimal, correctly scoped, and I independently traced the collision it closes and confirmed it cannot regress the legitimate cross-table rename-cascade case (see analysis below). Test coverage exercises both the fixed bug and the pre-existing legitimate case side by side.

Non-blocking suggestions are listed at the end.

---

## Change summary

`AbstractTableAssembly.renameAggregateInfo(oname, nname)` (`core/src/main/java/inetsoft/uql/asset/AbstractTableAssembly.java:1707`) is called from `renameDepended`, which `Worksheet.renameAssembly` (`core/src/main/java/inetsoft/uql/asset/Worksheet.java:483-486`) invokes on *every* assembly in the worksheet — including the assembly that was just renamed itself (`setName(nname)` already ran before the loop). The fix adds a guard:

```java
if(Tool.equals(getName(), nname)) {
   return;
}
```

before delegating to `ginfo.renameDepended(oname, nname)`, so the cascade no longer fires on the renamed assembly's own `AggregateInfo`.

## Root-cause verification

I traced the mechanism independently rather than taking the PR description at face value:

- `AggregateInfo.renameDepended` (`core/src/main/java/inetsoft/uql/asset/AggregateInfo.java:754`) calls `renameEntity(...)` on every group/aggregate ref, which bottoms out in `ColumnRef.renameAttributeRef` (`core/src/main/java/inetsoft/uql/asset/ColumnRef.java:101`): `if(Tool.equals(oname, ref.getEntity())) { ...rebuild with entity = nname... }`.
- For a table added from a logical-model entity, `WorksheetAgentController.addLogicalModelTable` sets the column's `AttributeRef` entity to the logical-model entity name, which — per the PR's description of that code path — is also used as the table's default worksheet-assembly name. When that table is renamed, `oname`/`nname` in the `Worksheet.renameAssembly` loop equal the table's own old/new name, so `entity == oname` is coincidentally true for the table's *own* aggregate, and the rewrite corrupts the ref to point at a name that no longer resolves against the logical model.
- I confirmed the originating mechanism (`AggregateInfo.renameDepended`/`renameEntity`) was added by commit `28a5ede3b` ("`AggregateInfo.renameDepended requalifies group/aggregate entities`"), which exists in this repo's history and is intended for a different case: a *composed* table's (mirror/join) group/aggregate entity is qualified by a **child worksheet assembly's name** (via `AssetUtil.getOuterAttribute(childName, ...)`, confirmed in `MirrorTableAssembly`/`ConcatenatedTableAssembly`), which legitimately needs to be requalified when that child is renamed.

## Why the guard is safe (no regression, no lost functionality)

I checked each way the guard could theoretically break something and did not find one:

1. **Uniqueness invariant.** `Worksheet.renameAssembly` rejects the rename up front if `nname` is already used by another assembly (`containsAssemblyIgnoreCase(nname)`), and assembly names share one namespace across all assembly types (tables, named-group assemblies, etc. all live in the same `assemblies` list). So after the rename, `getName() == nname` is true for exactly one assembly in the loop — the one that was renamed — never for any other assembly whose entity/named-group reference might legitimately equal `nname`. The guard therefore can't accidentally skip a legitimate cross-assembly case.
2. **The guard short-circuits the whole `ginfo.renameDepended` call, not just the entity rewrite** — this also skips `GroupRef.renameDepended(oname, nname)`, which updates a *different*, unrelated field (`assembly`, a named-group-assembly reference used for custom grouping). By the same uniqueness argument, that field can never legitimately equal `oname` (the renamed table's own old name) on the renamed table's own `GroupRef`, since no other assembly could have held that exact name while the table itself held it. So skipping it here loses nothing.
3. **Composed tables are structurally immune to the same collision** — `ComposedTableAssembly.renameAggregateInfo` (`core/src/main/java/inetsoft/uql/asset/ComposedTableAssembly.java:434-453`) calls `super.renameAggregateInfo` (inheriting the guard) and then runs its own extra `ColumnRef.renameColumn` loop directly. I checked `MirrorTableAssembly`, `ConcatenatedTableAssembly`, and `CompositeColumnHelper` — in every case the composed table's group/aggregate columns are qualified by a **child** assembly's name, never the composed table's own name (a table can't be its own child, and names are unique), so the self-rename collision cannot occur there and no second guard is needed. The PR's second test exercises exactly this path (mirror depending on a renamed source) and confirms it still requalifies correctly.
4. **Condition lists (`preconds`/`postconds`/`topns`/MV conditions) and `renameExpressionDependeds` are untouched by this PR and don't need the same fix** — I checked `ConditionUtil.renameConditionListWrapper` → `AssetCondition.renameDepended`, which only rewrites `SubQueryValue` and `UserVariable` references (different assembly names, not the "entity string matches this table's own name" pattern), so they aren't exposed to the same collision class.
5. **`RenameColumnController.renameAggregateInfo`** is an unrelated private method (in `core/src/main/java/inetsoft/web/composer/ws/RenameColumnController.java`) for per-column rename, not assembly rename — confirmed it's a different code path, not affected by this change.
6. `AbstractTableAssembly.renameAggregateInfo` has exactly one call site (`renameDepended`, itself only called from `Worksheet.renameAssembly`), so the blast radius of the change is fully enclosed by the analysis above.

## Test coverage

`core/src/test/java/inetsoft/uql/asset/WorksheetTest.java` adds two tests:

- `renameAssemblyDoesNotRequalifyItsOwnAggregateRefOnSelfRename` — reproduces the bug directly (logical-model-style `BoundTableAssembly` whose column entity equals its own name, aggregate sharing the same `ColumnRef` object as the column selection per `WorksheetMutationSupport#applyAggregateInfo`'s no-clone sharing), asserts the aggregate and its entity survive a rename. PR states this fails before the fix and passes after — I did not re-run the suite myself, but the assertions match the traced bug mechanism exactly, and CI's `build` check (which runs the full test suite) is green on this diff.
- `renameAssemblyStillRequalifiesAnotherTablesAggregateThatDependsOnTheRenamedTable` — companion regression test for the legitimate cross-table case (`28a5ede3b`'s original scenario) via a `MirrorTableAssembly` aggregate over a renamed source table. Good practice: pairing the fix test with an explicit non-regression test for the case the guard must *not* affect.

Constructors/APIs used in the new tests (`BoundTableAssembly(ws, name)`, `SourceInfo(int, String, String)`, `MirrorTableAssembly(ws, name, WSAssembly)`, `TestWorksheets.tableWithColumns`) all match their actual signatures in the codebase.

I did not find any existing test that this change would break — the guard only changes behavior for the previously-buggy self-rename case; the cross-table cascade path (what other passing tests would plausibly exercise) is untouched.

## Other review dimensions

- **Security**: none applicable — pure in-memory rename-cascade logic, no I/O, no injection surface.
- **Backward compatibility**: the change only prevents an erroneous mutation from happening; it does not change any public API, serialization format, or persisted state shape. No backward-incompatibility concerns.
- **Readability/maintainability**: the guard is short and the accompanying comment clearly explains *why* (not just *what*), consistent with the codebase's existing commenting style in this area (e.g. the similar detailed comments already present in `ColumnRef.java`).
- **Code style**: `Tool.equals(...)` (null-safe) is used rather than `.equals()`/`==`, consistent with codebase conventions and correctly null-safe despite the code comment's informal phrasing ("`getName() == nname`") in the PR description.

## Non-blocking suggestions

1. The inline comment in `AbstractTableAssembly.java` cites a specific commit hash (`28a5ede3b`) for context. This is helpful today but will rot if that commit is ever squashed/rebased away in history rewrites. Referencing the Redmine bug/PR number instead (or in addition) would be more durable. Not worth blocking on.
2. The PR targets `stylebi-wiz-main-rebase` rather than `main`. This appears to be an intentional integration branch for a related series of wizard/agent-driven worksheet-mutation fixes (several similarly-themed PRs — #5392, #5388, #5380 — are already merged there), so this is very likely correct, but it's worth a sanity check by the author/maintainers that this is the intended target and not an accidental base-branch mix-up, since it does mean this fix won't reach `main` until that branch itself merges forward.
3. Consider (optionally, future work, not this PR) whether the same "entity happens to equal the assembly's own name" coincidence could affect other rename-cascade consumers outside `AggregateInfo` if new ones are added later (e.g. sort refs, ranking conditions in `ComposedTableAssembly`) — today they're all either guarded transitively via this fix (composed-table extra loops only ever reference child names) or use a different, unaffected rename mechanism, but this class of bug (convention-based entity qualification colliding with default naming) is easy to reintroduce in new code and might be worth a short note in `claude/erm.md` or `claude/asset-framework.md` for future maintainers. Not required for this PR to merge.

